PR #2003, opened 2018-11-19
22684-testNoUnusedTemporaryVariablesLeft-uses-all-CompiledMethod-instances

pavel-krivanek:
https://pharo.fogbugz.com/f/cases/22684/testNoUnusedTemporaryVariablesLeft-uses-all-CompiledMethod-instances

use all installed localMethods instead of all compiled method instances


MarcusDenker (approved):
ok!

PR #2004, opened 2018-11-19
Fix the layout of Glamour tabulator rows and columns in the presence …

hamamo:
…of fixed-size panes.

Note: resizing with splitter still does not work correctly, there's more work to be done there.


hamamo:
Test failure is most likely caused by slightly different formatting on MacOS. Will make the test more robust.

hamamo:
Apparently, the screen size of the MacOS VM is so small that the window opened by the test cannot accomodate the panes. When setting a fixed size, this should hopefully work.

pavel-krivanek (approved):

PR #2006, opened 2018-11-21
22689-support-reification-value-and-operation-on-LiteralArray

MarcusDenker:
support reification #value and #operation on LiteralArray
  https://pharo.fogbugz.com/f/cases/22689/support-reification-value-and-operation-on-LiteralArray


MarcusDenker:
will redo

PR #2009, opened 2018-11-22
22690-32-vs-64-bits-and-large-integer-hash

MarcusDenker:
32 vs 64 bits and large integer hash
https://pharo.fogbugz.com/f/cases/22690/32-vs-64-bits-and-large-integer-hash


pavel-krivanek (approved):
I checked the related discussion. I do not think it an ideal solution but good enough

PR #2010, opened 2018-11-22
22691-Adopt-file-names-following-branch-tag-convention

estebanlm:
https://pharo.manuscript.com/f/cases/22691/Adopt-file-names-following-branch-tag-convention

using branch approach to calculate version when not directly tagged.

Then, it will be something like this:
- tagged versions will be called: Pharo7.0.0-rc2* (for example)
- non tagger versions will be called: Pharo7.0-SNAPSHOT* (for example)

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -8,6 +8,20 @@ def isWindows(){
     return env.NODE_LABELS.toLowerCase().contains('windows')
 }
 
+// Answers if we are in a development branch (we assume is a development branch if it starts with "Pharo")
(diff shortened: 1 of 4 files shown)


estebanlm:
in fact t this is ready. 
I'm only hesitating about the name of the development branch. 
Since we are not going to have "release" branch, having just "dev-blah" seems to me is not so good. So I was thinking on calling each branch `Pharo7.0`, `Pharo8.0`, etc.

estebanlm:
ok, I will go for "PharoEtc" nomination :)

PR #2012, opened 2018-11-26
22695-Make-value-for-work-method-after

MarcusDenker:
Make #value for work method #after
https://pharo.fogbugz.com/f/cases/22695/Make-value-for-work-method-after


MarcusDenker:
This PR fixed #value for #after on MethodNode.

The strategy is to not wrap in a block, but instead use the method wrapper functionality that we need for primitive method anyway.

- Now #after  on method wraps method instead of doing block wrapping

- wrapper always saves return value in a temp

TODO: ensure: wrapping, which will be added later back

PR #1980, opened 2018-11-13
22658-Better-manage-encoding-of-environment-variables

guillep:
Enhance management of encoding of environment.
 - common API manages encoding automatically by the system
 - *nix provides raw* and *encoding:* versions that allow getting the raw bytes and to provide an encoding

Issue: https://pharo.fogbugz.com/f/cases/22658/Better-manage-encoding-of-environment-variables

PR #2015, opened 2018-11-27
22698-OSEnvironment-is-missing-setEnvvalue-method

jecisc:
Add missing abstract method to OSEnvironment

Fixes #22698

### DIFF
--- a/src/System-OSEnvironments/OSEnvironment.class.st
+++ b/src/System-OSEnvironments/OSEnvironment.class.st
@@ -212,6 +212,13 @@ OSEnvironment >> removeKey: key [
 ]
 
 { #category : #accessing }
+OSEnvironment >> setEnv: nameString value: valueString [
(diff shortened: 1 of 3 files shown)


jecisc:
https://pharo.fogbugz.com/f/cases/22698/OSEnvironment-is-missing-setEnv-value-method

akgrant43 (approved):
Approved, except they call the platform specific set environment routine, not getenv() (see the comment) :-)

jecisc:
@akgrant43 Updated.

akgrant43:
All good, thanks!

guillep:
Hi, i don't think this is a good change. I'd deprecate all usages of getenv and setenv in favor of at and atput. See #1980

jecisc:
Ok, I was not aware of how this class should be used. I only saw that the methods were on all subclasses but not the superclass.

PR #2028, opened 2018-11-30
22687-Update-Tonel

tesonep:


jecisc:
I'll test when the build will success and I will have an image to test. Relaunching for the second time.

pavel-krivanek (approved):

PR #2029, opened 2018-11-30
22718-argumentNames-should-not-use-propertyValueAt

MarcusDenker:
argumentNames should not use propertyValueAt
https://pharo.fogbugz.com/f/cases/22718/argumentNames-should-not-use-propertyValueAt


MarcusDenker:
now this will continue to fail as re-runs are not done against the latest code

MarcusDenker:
I merge it because is is just renaming a call

PR #2031, opened 2018-11-30
22676-Cannot-extract-method-with-a-name-matching-the-one-in-a-superclass-being-a-subclassResponsibility

carolahp:
changed the validation for launching an #alreadyDefined error. Now the selector name is searched directly in the class instead of its  hierarchy


dionisiydk:
I think it is wrong change. Idea of this precondition is to ensure safeness of refactoring. When extracted method overrides existing method from superclass it can break functionality.
Current error is signal or notification to user that his change is not safe. Then user can proceed refactoring if everything is correct.

tesonep:
Fixed in Calypso.

PR #2032, opened 2018-11-30
22714 leftover halts in the system

dupriezt:
Removed a bunch of leftover `halt` in the image.
Left those that looked legitimate (for example in methods testing that `self halt` does indeed raise a Halt).
Replaced those that were used as debugging infrastructure (ex: `debug ifTrue: [self halt]`) with a new method: Object>>openDebugger (and its flavour Object>>openDebugger:). This new method just does `self halt`, but it clarifies that the halt it replaced was legitimate and not some leftover debugging code.

### DIFF
--- a/src/Kernel/Object.class.st
+++ b/src/Kernel/Object.class.st
@@ -1584,6 +1584,12 @@ Object >> openDebugger [
 	self halt.
 ]
 
+{ #category : #'as yet unclassified' }
+Object >> openDebugger: aString [
+	"Used in debugging infrastructure that should stay in production, instead of leaving a halt in the code. This way it is not flagged as debugging code left in method"
+	self halt: aString.
+]
+
 { #category : #'message performing' }
 Object >> perform: aSymbol [ 
 	"Send the unary selector, aSymbol, to the receiver.
--- a/src/Kernel/Object.class.st
+++ b/src/Kernel/Object.class.st
@@ -1578,6 +1578,12 @@ Object >> okToChange [
 	^ true
 ]
 
+{ #category : #'as yet unclassified' }
+Object >> openDebugger [
+	"Used in debugging infrastructure that should stay in production, instead of leaving a halt in the code. This way it is not flagged as debugging code left in method"
+	self halt.
+]
+
 { #category : #'message performing' }
 Object >> perform: aSymbol [ 
 	"Send the unary selector, aSymbol, to the receiver.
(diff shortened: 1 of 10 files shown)


MarcusDenker:

src/Kernel/Object.class.st
```diff
@@ -1578,6 +1578,18 @@ Object >> okToChange [
 	^ true
 ]
 
+{ #category : #'as yet unclassified' }
```
pavel-krivanek: Uncategorized methods, causes tests failures

dupriezt: I just recategorised this method, as well as two others that were not categorised. I am not sure why the recategorisation of okToChange does not appear in the commit though. https://github.com/pharo-project/pharo/pull/2032/commits/73630e39810a83f2957dcfc95c24ed77e2a60f43

dupriezt: I tried to reload my branch in a fresh image to understand why the recategorisation of okToChange does not appear in the commit, but upon loading it, I get tons of `Instance of RBBlockNode did not understand #isParseError` errors that lock the image...

src/Kernel/Object.class.st
```diff
@@ -1578,6 +1578,18 @@ Object >> okToChange [
 	^ true
 ]
 
+{ #category : #debugging }
+Object >> openDebugger [
```
pavel-krivanek, line 1582: Actually, I'm not really sure about this. The kernel should be independent of things like debugger as much as possible. The system raises an exception and it's the responsibility of tools what they will do. The opening of the debugger is only one option, in a production application, it may be something else. Then this message could be misleading.

dupriezt: The reasoning behind this method is that some code in the image, for example DelayBasicScheduler>>runBackendLoopAtTimingPriority, DelayBasicScheduler>>schedule: contain code like
```"When two debuggers appear, step/proceed through this higher priority one first."
debug ifTrue: [ self halt ].```
This looks like some debugging code left in the image on purpose as a framework to ease debugging later. So a proxy method to call halt (which is the only thing Object>>openDebugger does) allows this method to specify that this "halt" is intended to stay in production.

dupriezt: I did that with @MarcusDenker during a sprint, so maybe he can contribute to the discussion.

Ducasse: You should really discuss with Guille. But as a KEY GENERAL principle we do not want to have dependencies to tools from the Kernel. This is an insane idea! 
We spent days removing them. Can't we think in terms of layers?

Ducasse:
Please split the changes so that we can integrate the simple ones.

PR #2034, opened 2018-12-01
22724-Setting-to-disable-job-interruption

jecisc:
Add a setting to be able to disable progress bar interruption.

Fixes https://pharo.fogbugz.com/f/cases/22724/Setting-to-disable-job-interruption


MarcusDenker:

src/Morphic-Base/JobProgressBarMorph.class.st
```diff
@@ -1,22 +1,61 @@
 "
 A JobProgressBarMorph is a progress bar with a title and a icon to stop update.
 
-JobProgressBarMorph new openInWorld
+		JobProgressBarMorph new openInWorld.
+
+I am often used by the Job class in pharo and not directly.
```
pavel-krivanek:
```suggestion
I am often used by the Job class in Pharo and not directly.
```

src/Morphic-Base/JobProgressBarMorph.class.st
```diff
@@ -1,22 +1,61 @@
...
+
+It is possible to disable the icon to prevent the user of stopping the task via the #isInterruptable setting.
```
pavel-krivanek:
```suggestion
It is possible to disable the icon to prevent the user from stopping the task via the #isInterruptable setting.
```

jecisc:
The error is related. It seems that JobProgressBarMorph class is not initialized at loading...

jecisc:
I rebased to P8. I'll need to check with someone why my class is not initialized.

jecisc:
Hum... Maybe I do not need the initialization method.

tesonep (approved):

PR #1999, opened 2018-11-16
22679-Metacello-attributes-should-be-based-on-SystemVersion-instead-of-hardcoding

jecisc:
Do not hardcode platform attributes of Metacello.

PR #2035, opened 2018-12-02
Revert "22679-Metacello-attributes-should-be-based-on-SystemVersion-instead-of-hardcoding"

estebanlm:
Reverts pharo-project/pharo#1999

### DIFF
--- a/src/System-Support/SmalltalkImage.class.st
+++ b/src/System-Support/SmalltalkImage.class.st
@@ -1128,14 +1128,9 @@ SmalltalkImage >> memoryHogs [
 
 { #category : #miscellaneous }
 SmalltalkImage >> metacelloPlatformAttributes [
-	"Returns the tags for the conditional platform loading in Metacello. Pay attention the order is important: from most  to least general."
-
-	| attributes systemeVersion |
-	attributes := OrderedCollection withAll: #(#squeakCommon #pharo).
-	systemVersion := SystemVersion current.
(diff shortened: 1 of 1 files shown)


MarcusDenker:
yes, green!

jecisc:
Do you have a copy of the failing build? Because the check was green at the time we integrated this PR. I would like to understand where is the problem. 
Maybe we use the Metacello attributes before setting the system version? But in that case, why the check was green?

MarcusDenker:
every CI test run now fails with

Errors in script loaded from /builds/workspace/uest_and_branch_Pipeline_PR-2032/32/bootstrap/scripts/bootstrap.st'
[32] MessageNotUnderstood: EPUndeclaredBinding>>markWrite
[32] EPUndeclaredBinding(Object)>>doesNotUnderstand: #markWrite


We can not integrate anymore as all tests of all PRs are read. Only this one now is green again

MarcusDenker:
my theory is that it leads to an Undeclared in the bootstrap somehow. And that triggers a bug with the bootstrap that makes the bootstrap fail

MarcusDenker:
it actually triggers a bug in the bootstrap: the bootstrap is broken when there are undeclared as the are managed somehow with subclasses of Association (EPUndeclaredBinding), but this should be a subclass of LiteralVariable instead.

But I do not even know where that class comes from, the I can not fix it. (it is not on the image)

estebanlm:
anyway, I will integrate :)

tesonep:
To fix it, it should be an extension method in Metacello, not a method in the kernel. It is adding a new dependency with SystemVersion and this is not handled.

PR #2036, opened 2018-12-02
Fixed menubar consistency

philippeback:
* no more moving labels
* use the world menu highlight color for the menu
* add spaces for lining up labels


jecisc:
I think that we should still call #menubarItemSelectionFillStyleFor: to allow customization. 

What we can change its is value in the theme. But I wonder if the menubar should take the color of the context menu, or if the context menu should take the color of the menu bar because we choose the color by looking at other GUI projects since the goal of the menubar is to be closer to standard applications.

philippeback:
World menu is using the menu colors settings. I do not like theme defaults for menus, that is why I change that in the settings.

No issue with using #menubarItemSelectionFillStyleFor: which in turn will call the appropriate thing.

FWIW, PragmaMenuBuilder uses the value I used.

MarcusDenker:

PR #2037, opened 2018-12-02
22726 pharo icon in menus is larger than other icons 20 pix instead of 16 making its row height different

philippeback:
No more mismatched icon and line height in menus.

image


MarcusDenker:

jecisc:
This PR has code not related to the issue but to another issue.

philippeback:
Well, it is an issue on top of the menubar one, the actual commit is for the 16@16 instead of the 20@20.
The top bar has to be fixed anyway and having the menu with other colors than the world menu for selection is really disturbing. 
What do you advise me to do for this thing? I actually created a branch from an issue in iceberg and commited. So, maybe it branched from the branch instead of from master indeed.

PR #2039, opened 2018-12-03
22727-NonInteractiveTranscript-is-broken-with-files

svenvc:
Fix the issue with NonInteractiveTranscript when using files as output
Add some elementary unit tests in package Transcript-NonInteractive-Tests

https://pharo.manuscript.com/f/cases/22727/NonInteractiveTranscript-is-broken-with-files


jecisc:
Thank you for the fix!

The failing test is not related.

pavel-krivanek (approved):

PR #2040, opened 2018-12-04
Improving comment of class DebugSession

dupriezt:


Ducasse:
Thanks.
Could you put Pharo code instead of Smalltalk code :).
And we push it into the system

dupriezt:
*"Could you put Pharo code instead of Smalltalk code :)."*
@Ducasse I don't understand.

estebanlm:
s/Smalltalk/Pharo/g

:)

estebanlm (changes requested):
What Stef asks is to replace "Smalltalk" with "Pharo".
Then it will be ok :P

dupriezt:
Ah ok. It is done now.
I also fixed a few typos while I was at it.

PR #2042, opened 2018-12-05
Adding BreakpointBrowser to Pharo

dupriezt:
BreakpointBrowser is accessible via the "Debugging" menu of the image and shows a list of the methods that contain halts or breakpoints in the image as a Calypso method browser.

### DIFF
--- a/src/BreakpointBrowser/BreakpointBrowser.class.st
+++ b/src/BreakpointBrowser/BreakpointBrowser.class.st
@@ -1,0 +1,83 @@
+"
+My class method ""open"" opens a Calypso method browser showing all the methods of the image that have halts or breakpoints in them. The different flavours of halts and breakpoints are also covered (break once, break on condition, haltIf: haltOnCount:, haltOnce...).
+
+By default, I do not show methods from my own package, nor methods registered on my ""blacklist"" class method. Blacklisted methods are meant to be methods from the core of Pharo that include halts for legitimate reasons (testing for example) and would probably pollute the experience of users if they were shown in the BreakpointBrowser.
+If you want to see all methods containing halts and breakpoints without any filtering, you can call my ""openUnfiltered"" class method. 
+"
+Class {
+	#name : #BreakpointBrowser,
+	#superclass : #Object,
+	#category : #BreakpointBrowser
+}
+
+{ #category : #filtering }
+BreakpointBrowser class >> blacklist [
+	"These methods will not be shown by BreakpointBrowser by default. They contain halts but are meant to contain them (for example, they can be testing that executing self halt does indeed raise a halt)
+	Format: Array of {class name symbol. method name symbol} arrays"
+	^ { 
+	{#ClyClassWithHalts. #methodWithHalts2}.
+	{#ClyClassWithHalts. #methodWithHalts1}.
+	{#ClyClassWithProblemMethods. #methodWithHalt}.
+	{#ClyClassWithProblemMethods. #methodWithHalt2}.
+	{#GTSUnitExampleFailingTest. #testWithHalt}.
+	{#Object. #openDebugger}.
+	{#Object. #openDebugger:}.
+	{#ObjectTest. #b}.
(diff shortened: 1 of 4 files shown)


src/BreakpointBrowser/BreakpointBrowser.class.st
```diff
@@ -0,0 +1,83 @@
...
+	{#ObjectTest. #testHaltOnce}.
+	{#ObjectTest. #testHaltOnCount}.
+	{#UndefinedObjectTest. #testHaltIfNil}.
+	{#UndefinedObjectTest. #testIfNil}.
+	{#UndefinedObjectTest. #testIfNilIfNotNil}.
+	{#UndefinedObjectTest. #testIfNotNilIfNil}.
+	{#CodeSimulationTests. #methodWithHalt}.
+	{#RBRefactoryTestDataApp. #cruft}.
+	{#RBSmalllintTestObject. #codeCruftLeftInMethods}.
+	{#RBDummyRefactoryTestDataApp. #cruft}.
```
dionisiydk, line 37: Did you consider to mark those methods with special pragma instead of hardcoding such list?
Like add \<testingHalt\> into method which should not be shown as method with real halt.

src/BreakpointBrowser/BreakpointBrowser.class.st
```diff
@@ -0,0 +1,83 @@
...
+		aMethod containsHalt]].
+]
+
+{ #category : #filtering }
+BreakpointBrowser class >> open [
+	<script>
+	"Opens a method browser showing all methods containing breakpoints (and its flavours like break once), halts (and its flavours like haltOnce), inactive breakpoints (and flavours) and inactive halts (and flavours). Filter out such methods that belong to the HaltManager package."
+	| b |
+	b := ClyQueryBrowser browseMethods: (self filterOutMethodsFromBlacklist: (self filterOutMethodsFromThisPackage: (self methodsToDisplay))) withTitle: 'Breakpoint Browser'.
+	b switchResultTo: (ClySortedQueryResult using: SortMethodsByPackageFunction new )
```
dionisiydk, line 72: I feel that all this code can be part of current calypso plugin for breakpoints.
This method can be defined on ClyQueryBrowser:
```Smalltalk
ClyQueryBrowser openBreakpoints
```
The filtering logic can be moved to ClyActiveBreakpoints query as parameter "all halts or only real halts" with "only real halts" as default value. Then #openBreakpoints can be defined as:
```Smalltalk
b := ClyQueryBrowser openOn: ClyActiveBreakpoints new.
b switchResultTo: (ClySortedQueryResult using: SortMethodsByPackageFunction new).
```

dupriezt: I did that (and added the pragma you mentioned above named: <haltOrBreakpointForTesting>).
https://github.com/dupriezt/Calypso/tree/breakpointQueryImprovement

But there seems to be a caching issue going on. When loading a fresh image and running `ClyQueryBrowser browseBreakpointsExcludingHaltsForTesting.`, it shows 37 methods (expected). When running `ClyQueryBrowser browseBreakpoints.` after it, it still shows 37 methods even though it should show the one method that has the pragma in addition.
When loading a fresh image and running `ClyQueryBrowser browseBreakpoints.` first, it shows the expected 38 methods.

dionisiydk: Let's discus details here https://github.com/pharo-ide/Calypso/pull/383

dionisiydk: I explained it in Calypso PR

pavel-krivanek:
for Pharo 8

dupriezt:
This feature is being implemented into Calypso, so this pull request can be closed.

dionisiydk:
But now you can mark example methods with halts with pragmas

чт, 20 дек. 2018 г., 15:00 dupriezt notifications@github.com:

PR #2045, opened 2018-12-05
Fix for issue 22729

bouraqadi:
Fix dropping .st and .changes file correctly handled and does not raise exception.
Added regression tests


src/Tool-ExternalBrowser/ExternalBrowserTest.class.st
```diff
@@ -0,0 +1,85 @@
+Class {
+	#name : #ExternalBrowserTest,
+	#superclass : #TestCase,
+	#instVars : [
+		'initialWindows',
+		'file'
+	],
+	#category : #'Tool-ExternalBrowser-Test'
```
pavel-krivanek, line 8: The test should be in a standalone package

PR #2056, opened 2018-12-10
issue22748-method-template

bencoman:
Issue #22748 - distinguish between instance-side and class-side in method template

### DIFF
--- a/src/OpalCompiler-Core/Behavior.extension.st
+++ b/src/OpalCompiler-Core/Behavior.extension.st
@@ -110,12 +110,12 @@ Behavior >> sourceCodeTemplateFor: aClassOrMetaclass [
 	"Answer an expression to be edited and evaluated in order to define 
 	methods in this class or trait."
 
-		^ aClassOrMetaclass isMeta 
+		^ aClassOrMetaclass isMeta
 				ifTrue: 
 [ 
 'messageSelectorAndArgumentNames
-	"comment stating purpose of factory/class-side message"
-	"scope: class-variables  &  factory/class-instance-variables"	
+	"comment stating purpose of class-side message"
+	"scope: class-variables  &  class-instance-variables"	
 			
 	| temporary variable names |
 	statements' 
--- a/src/OpalCompiler-Core/Behavior.extension.st
+++ b/src/OpalCompiler-Core/Behavior.extension.st
@@ -104,3 +104,29 @@ Behavior >> sourceCodeTemplate [
 	| temporary variable names |
 	statements'
 ]
+
+{ #category : #'*OpalCompiler-Core' }
+Behavior >> sourceCodeTemplateFor: aClassOrMetaclass [ 
+	"Answer an expression to be edited and evaluated in order to define 
+	methods in this class or trait."
+
+		^ aClassOrMetaclass isMeta 
+				ifTrue: 
+[ 
+'messageSelectorAndArgumentNames
+	"comment stating purpose of factory/class-side message"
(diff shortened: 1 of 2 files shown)


bencoman:
Introduces Behaviour>>sourceCodeTemplateFor: aClassOrMetaclass. 
Now I'm not sure what to do with the old method Behaviour>>sourceCodeTemplate method which has
only a single sender RubTextAreaExamples-class>>sourceCodeTemplateWrappedTo400
which has no senders.

src/OpalCompiler-Core/Behavior.extension.st
```diff
@@ -104,3 +104,29 @@ Behavior >> sourceCodeTemplate [
 	| temporary variable names |
 	statements'
 ]
+
+{ #category : #'*OpalCompiler-Core' }
+Behavior >> sourceCodeTemplateFor: aClassOrMetaclass [ 
+	"Answer an expression to be edited and evaluated in order to define 
+	methods in this class or trait."
+
+		^ aClassOrMetaclass isMeta
```
pavel-krivanek: it should use isClassSide, or better, use polymorphism instead of if

src/OpalCompiler-Core/Behavior.extension.st
```diff
@@ -104,3 +104,29 @@ Behavior >> sourceCodeTemplate [
...
+{ #category : #'*OpalCompiler-Core' }
+Behavior >> sourceCodeTemplateFor: aClassOrMetaclass [ 
+	"Answer an expression to be edited and evaluated in order to define 
+	methods in this class or trait."
+
+		^ aClassOrMetaclass isMeta 
+				ifTrue: 
+[ 
+'messageSelectorAndArgumentNames
+	"comment stating purpose of factory/class-side message"
```
pavel-krivanek: why factory? The comments should be different for traits

Ducasse:
Hi Ben 
can you do a pass on Pavel comments like that we can integrate it. 

Stef

src/OpalCompiler-Core/Behavior.extension.st
bencoman: I did try polymorphism first, but it felt awkward piling another extra method onto Object.  Whether its class-side or instance-side, both seem to be a functions of Behaviour, so in this case I went with the conditional.
I'll update it to use isClassSide.

src/OpalCompiler-Core/Behavior.extension.st
bencoman: > why factory?

The idea that "classes are objects" is foreign to most newcomers.  Early on I read that Smalltalk comes with the Factory Design Pattern "built-in" and that really had an impact on me.  Framing the unfamiliar against the familiar provide better learning outcomes.  But I don't see this mentioned much and think we under promote this.  
This was just a slight attempt to promote that.  But its a personal view, so if you don't like it, I can delete "factory"  from  it.

> The comments should be different for traits

Can you suggest some code to test for this, plus associated text?

astares: Yes - please delete the "factory"

src/OpalCompiler-Core/Behavior.extension.st
bencoman: I've just tested it with isClass, and that doesn't work.
I've added OCSourceTemplates>>testsourceCodeTemplateFor to help demonstrate this.

Ducasse:
Thanks.

PR #2057, opened 2018-12-10
22750-Add-GTPlayground-classopenUrl

svenvc:
Add GTPlayground class>>#openUrl:

https://pharo.fogbugz.com/f/cases/22750/Add-GTPlayground-class-openUrl


MarcusDenker (approved):
ok

PR #2064, opened 2018-12-12
[Do not integrate] Update bootstrap to pharo7

carolahp:


bencoman:
Can you comment on why its "do not integrate".

carolahp:
> Can you comment on why its "do not integrate".

This is an intermediate step. It is necessary to create definitive versions of the related repositories (in progress). Besides, I prefer to wait for the Pharo7 release.

Ducasse:
What should we do on this one? we are now in P8

tesonep:
The process is still using Pharo6.1, it should be passed to use Pharo7 (as it is the stable).
@carolahp maybe it will be better to open a new PR with the modification to take all the changes. 
Also, we need to generate a release of Ring2 with the required support.

PR #2065, opened 2018-12-13
22761 Add class rule to point to untagged/uncategorized classes

astares:
Add rule to point to uncategorized methods 

https://pharo.fogbugz.com/f/cases/22761

### DIFF
--- a/src/Renraku/ReUncategorizedClassRule.class.st
+++ b/src/Renraku/ReUncategorizedClassRule.class.st
@@ -1,0 +1,50 @@
+"
+Check if the class is uncategorized within its package - if yes it should critisized so it will be tagged
+with either a proper tag or a tag called ""Base"" as then the class belongs to the base part of the package.
+"
+Class {
+	#name : #ReUncategorizedClassRule,
+	#superclass : #ReAbstractRule,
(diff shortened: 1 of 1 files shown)


src/Renraku/ReUncategorizedClassRule.class.st
```diff
@@ -0,0 +1,50 @@
...
+ReUncategorizedClassRule class >> uniqueIdentifierName [
+	"This number should be unique and should change only when the rule completely change semantics"
+	
+	^'UncategorizedClassRule'
+]
+
+{ #category : #running }
+ReUncategorizedClassRule >> basicCheck: aClass [
+
+	^(aClass category = aClass package name)
```
pavel-krivanek, line 30: I think this is too naive implementation. For small packages, there should be no reason to force use categories. The rule should fail only if the package has already some categories but the particular class is not categorized.

I think it should wait on Pharo 8 since it is not a bugfix

Ducasse:
I do not really understand the purpose of this rule. We should be able to have package without tags.
In addition class category should not be used anymore.

Ducasse:
I will close this issue.

PR #2066, opened 2018-12-13
22762 Cleanup TextEditorTest

astares:
- reduce double code by using setUp/tearDown
- tag and comment the class

https://pharo.fogbugz.com/f/cases/22762


MarcusDenker (approved):
ok

PR #2071, opened 2018-12-13
22766-Transcript-window-is-leaking

pavel-krivanek:
unsubscribe Transcript window during its closing

https://pharo.fogbugz.com/f/cases/22766/Transcript-window-is-leaking


MarcusDenker (approved):
ok

PR #2072, opened 2018-12-13
22768 Categorize classes in OSWindow-SDL2-Base

astares:
tagging only to not be shown as uncategorized - no change in behavior, safe to integrate

https://pharo.fogbugz.com/f/cases/22768/

### DIFF
--- a/src/OSWindow-SDL2/OSSDL2WindowHandle.class.st
+++ b/src/OSWindow-SDL2/OSSDL2WindowHandle.class.st
@@ -14,7 +14,7 @@ Class {
 	#pools : [
 		'SDL2Constants'
 	],
-	#category : #'OSWindow-SDL2'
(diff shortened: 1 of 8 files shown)


src/OSWindow-SDL2/OSSDL2WindowHandle.class.st
```diff
@@ -241,7 +241,7 @@ OSSDL2WindowHandle >> newGenericRenderer [
 		yourself
 ]
 
-{ #category : #'*OSWindow-SDL2-OpenGL' }
+{ #category : #'as yet unclassified' }
```
pavel-krivanek: uncategorized method

astares: fixed

pavel-krivanek (approved):

PR #2075, opened 2018-12-14
22772-chooseExistingFileReferenceextensionspathpreview-returning-wrong-FileReference

guillep:
Fix issue https://pharo.fogbugz.com/f/cases/22772/chooseExistingFileReference-extensions-path-preview-returning-wrong-FileReference.

The solution is to ask the file dialog to return a path and not a file name.


svenvc:
Duh, obvious. You only have to know. Thx for the fix!

guillep:
> Duh, obvious. You only have to know. Thx for the fix!

Yeh, It was not the most obvious thing, that's why I've missed it the first time :/

PR #2076, opened 2018-12-14
22777 ManifestMonticello and ManifestRingMonticello should be tagged with "Manifest"

astares:
- categorize /  tag the two classes with a "Manifest" class tag

https://pharo.fogbugz.com/f/cases/22777


MarcusDenker:
failure not related

MarcusDenker (approved):
ok

PR #2082, opened 2018-12-16
22785 Refactor FFICompilerPlugin

astares:
- Refactor FFICompilerPlugin>>ffiCalloutSelectors to be available from class side so it could be reused in other methods like CompiledMethod>>isFFICall from Calypso later
- Use FFI instead of "Ffi" in method names
- Separate into UFFI and NativeBoost selectors
- refactor initialization
- use proper categorization (like "private - initialization" for initialization methods)
- rework class comment

https://pharo.fogbugz.com/f/cases/22785/

### DIFF
--- a/src/UnifiedFFI/FFICompilerPlugin.class.st
+++ b/src/UnifiedFFI/FFICompilerPlugin.class.st
@@ -84,7 +84,7 @@ FFICompilerPlugin class >> defaultUFFICalloutSelectors [
 FFICompilerPlugin class >> ffiCalloutSelectors [
 	<script: 'self ffiCalloutSelectors inspect'>
 	
-	^FFICalloutSelectors ifNil: [ FFICalloutSelectors := self class initializeFFICalloutSelectors ]
+	^FFICalloutSelectors ifNil: [ self class initializeFFICalloutSelectors ]
 ]
 
 { #category : #'class initialization' }
--- a/src/UnifiedFFI/FFICompilerPlugin.class.st
+++ b/src/UnifiedFFI/FFICompilerPlugin.class.st
@@ -154,10 +154,10 @@ FFICompilerPlugin class >> uninstall [
 	self recompileSenders
 ]
 
-{ #category : #api }
-FFICompilerPlugin >> initializeFFICalloutSelectors [
+{ #category : #'private - accessing' }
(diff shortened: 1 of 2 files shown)


Ducasse:
Hi torsten

It would be important to deprecate the method selectors you change names. 

S.

dionisiydk:
Plugin was introduced in Pharo 7. I wonder that we need these deprecations

src/UnifiedFFI/FFICompilerPlugin.class.st
```diff
@@ -41,40 +41,75 @@ Class {
...
+	self addFFICalloutSelector: aPragma methodSelector.
+	self recompileSendersOf: aPragma method
+]
+
+{ #category : #'private - defaults' }
+FFICompilerPlugin class >> defaultFFICalloutSelectors [
+	"FFI uses UFFI and NB for the time being"
+	
+	^ self defaultUFFICalloutSelectors 
+		join: self defaultNativeBoostCalloutSelectors
```
dionisiydk: I wonder that #join really works for two arrays. But it make strange order. 
I think it is better to use #, message because the purpose of join: is different even if it works here.

src/UnifiedFFI/FFICompilerPlugin.class.st
```diff
@@ -41,40 +41,75 @@ Class {
...
+		 ffiCall:module:options:)
 ]
 
 { #category : #accessing }
-FFICompilerPlugin class >> defaultFfiCalloutSelectors [
-	^ #(ffiCall: #ffiCall:module: #ffiCall:options: #ffiCall:module:options: nbCall: #nbCall:module: #nbCall:options: #nbCall:module:options:)
+FFICompilerPlugin class >> ffiCalloutSelectors [
+	<script: 'self ffiCalloutSelectors inspect'>
+	
+	^FFICalloutSelectors ifNil: [ FFICalloutSelectors := self class initializeFFICalloutSelectors ]
```
dionisiydk: It does not looks nice. I would not expect anything from initialize method.

astares:
@dionisiydk Ok - should be fine now. 

For the deprecated methods I'm not sure - if someone already is using it it is OK. Not a big problem to have them.

Second review welcomed.

MarcusDenker (approved):
ok

PR #2083, opened 2018-12-16
22788  AthensCairoSurface class should not use nbCall:module: anymore

astares:
- use ffiCall:module: from UFFI instead of old NativeBoost nbCall:module:

https://pharo.fogbugz.com/f/cases/22788


MarcusDenker (approved):
ok

PR #2084, opened 2018-12-16
22760 RBRefactoryClassChange should define #isAbstract with true

astares:
- implement #isAbstract
- adjust comment

https://pharo.fogbugz.com/f/cases/22760/


MarcusDenker (approved):
ok

PR #2087, opened 2018-12-16
22792 Cleanup MessageTallyTest

astares:
- simple cleanups like use assert:equals

https://pharo.fogbugz.com/f/cases/22792/Cleanup-MessageTallyTest


MarcusDenker (approved):
ok

PR #2089, opened 2018-12-16
22794 Cleanup testExtractAllKinds

astares:
- use assert:equals:

https://pharo.fogbugz.com/f/cases/22794/Cleanup-testExtractAllKinds


MarcusDenker (approved):
ok

PR #2090, opened 2018-12-16
22795 Lint cleanup in Bitmap

astares:
- category change to satisfy lint and follow object protocol

https://pharo.fogbugz.com/f/cases/22795/Lint-cleanup-in-Bitmap


MarcusDenker (approved):
category change only, ok

PR #2091, opened 2018-12-16
22796 Fix testResettingTheSelection

astares:
- do not use shouldnt:raise: in  test method #testResettingTheSelection

https://pharo.fogbugz.com/f/cases/22796/Fix-testResettingTheSelection


MarcusDenker (approved):
ok

PR #2092, opened 2018-12-16
22797 deprecated file does not exist should be moved to deprecated70 package

astares:
https://pharo.fogbugz.com/f/cases/22797/Deprecated-FileDoesNotExist-should-be-moved-to-Deprecated70-package


MarcusDenker:
error not related

PR #2094, opened 2018-12-18
22803 GlamourPresentationModel should also be categorized

astares:
- just tag the class so we have no uncategorized left

https://pharo.fogbugz.com/f/cases/22803/GlamourPresentationModel-should-also-be-categorized


MarcusDenker:
error not related

PR #2096, opened 2018-12-18
22805-FTTableMorph-has-too-big-minimal-height

dionisiydk:
Set minHeight to be approximately size of one textual row


dionisiydk:
https://pharo.fogbugz.com/f/cases/22805/FTTableMorph-has-too-big-minimal-height

PR #2098, opened 2018-12-18
22211 Find better category for #openInWorld than initialize and unify category on subclasses

astares:
- Unify into #display method category

https://pharo.fogbugz.com/f/cases/22211/Find-better-category-for-openInWorld-than-initialize-and-unify-category-on-subclasses


MarcusDenker (approved):
ok

PR #2103, opened 2018-12-19
22812 Clean the ImageCleaner

astares:
Small cleanups - primarily putting
 - examplePackages
 - helpPackages
 - testPackages
on class side so it can be reused in ReleaseTests and other later

https://pharo.fogbugz.com/f/cases/22812/Clean-the-ImageCleaner


MarcusDenker (approved):
ok

PR #2104, opened 2018-12-19
22813 Integer should return isAbstract with true

astares:
- implement #isAbstract and add more tests

https://pharo.fogbugz.com/f/cases/22813/Integer-should-return-isAbstract-with-true


astares:
@MarcusDenker green

MarcusDenker (approved):
looks good to me

PR #2111, opened 2018-12-19
22821 Cleanup Exception class

astares:
- removed unused variables
- remove unnecessary spaces in comments
- let comments equally start uppercase

https://pharo.fogbugz.com/f/cases/22821/Cleanup-Exception-class

### DIFF
--- a/src/Debugging-Core/Exception.extension.st
+++ b/src/Debugging-Core/Exception.extension.st
@@ -2,7 +2,7 @@ Extension { #name : #Exception }
 
 { #category : #'*Debugging-Core' }
 Exception >> retry [
-	"Abort an exception handler and re-evaluate its protected block."
+	"Abort an exception handler and re-evaluate its protected block."
 
 	handlerContext restart
 ]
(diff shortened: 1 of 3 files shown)


gcotelli (approved):

MarcusDenker:
Exception tests are now failing.

astares:
The removed temp looked unnecessary but is required

```Smalltalk
resumeUnchecked: resumptionValue
	"Return resumptionValue as the value of #signal, unless this was called after an #outer message, then return resumptionValue as the value of #outer."

	| ctxt |
	outerContext 
		ifNil: [
			signalContext return: resumptionValue ] 
		ifNotNil: [
			ctxt := outerContext.
			outerContext := ctxt tempAt: 1. "prevOuterContext in #outer"
			ctxt return: resumptionValue ]
```

I will restore the temp and add a comment that it is used/accessed in #resumeUnchecked:
Exception tests are green again

Nonetheless we can continue with the PR due to the other small cleanups as soon as CI is green

MarcusDenker (approved):
ok

PR #2116, opened 2018-12-21
21876 Re-evaluating of class definition with slots cleans the class instance variables

guillep:
Fix https://pharo.fogbugz.com/f/cases/21876
Use #copyClassSlotsFromExistingClass to copy class side inst var slots


tesonep (approved):
Looks nice!

PR #2118, opened 2018-12-21
22828-MethodFinder-Enhancement

Ducasse:
fixed the printOn: so that we can print it.


jecisc:
There is a conflict now

MarcusDenker:
error not related

PR #2121, opened 2018-12-22
22833-PluggableTextMorph-and-co-should-be-removed

plantec:
added RubTextField>>#wrapFlag: to ensure that iceberg is still working with Rubric
To be fixed later (wraflag should be removed)


Ducasse:
Merci Alain :)

Ducasse:
Alain I will integrate your changes.

PR #2125, opened 2018-12-27
22826-senders-of-asNonTranslucentColor-should

Ducasse:
asNonTranslucentColor -> beOpaque


Ducasse:
https://pharo.fogbugz.com/f/cases/22826/senders-of-asNonTranslucentColor-should-used-beOpaque-instead

jecisc (approved):

PR #2131, opened 2018-12-27
22841-Fix-printString-related-bugs

Ducasse:
Fixed https://pharo.fogbugz.com/f/cases/22841/Fix-printString-related-bugs


Ducasse:
I worked in the previous branch :(
Now I think that both should be integrated.

Ducasse:
The broken printOn related are fixed

PR #2133, opened 2018-12-29
Bottom21

Ducasse:
Retrying for the fifth times (refork, reclone....).

PR #2142, opened 2018-12-30
Revert "Bottom21"

pavel-krivanek:
Reverts pharo-project/pharo#2133


pavel-krivanek:
I will revert them at once

Ducasse:
You can keep this too because they are inside the Rubric package and have no impact on others

Ducasse:
So should I redo them?

PR #2143, opened 2018-12-30
Revert "22848-Patch-to-avoid-breaking-iceberg"

pavel-krivanek:
Reverts pharo-project/pharo#2132


pavel-krivanek:
I will revert them at once

Ducasse:
This is the same for this one. We should keep the Rubric new package and this change can be kept.
Let me know and I can redo it.

PR #2149, opened 2018-12-31
22833-PluggableTextMorph-and-co-should-be-removed

plantec:
fixe the main issues reported by steph

### DIFF
--- a/src/Rubric/RubAbstractTextArea.class.st
+++ b/src/Rubric/RubAbstractTextArea.class.st
@@ -109,7 +109,7 @@ Class {
 
 { #category : #settings }
 RubAbstractTextArea class >> backgroundColor [
-	^ BackgroundColor ifNil: [ BackgroundColor := self theme backgroundColor ] 
+	^ BackgroundColor ifNil: [ self theme backgroundColor ]
 ]
 
 { #category : #settings }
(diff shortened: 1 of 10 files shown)


src/Rubric/RubAbstractTextArea.class.st
```diff
@@ -109,7 +109,7 @@ Class {
 
 { #category : #settings }
 RubAbstractTextArea class >> backgroundColor [
-	^ BackgroundColor ifNil: [ BackgroundColor := Color white] 
+	^ BackgroundColor ifNil: [ BackgroundColor := self theme backgroundColor ]
```
jecisc: Hi,

I don’t know if it’s done somewhere, but when the theme change, this should also be updated I think? Because if the defensive strategy is executed in the light theme, then we change to the dark theme then we need the background to not be white :)

Ducasse: Yes in my code it was different. I'm chatting with Alain.

Ducasse: What I did it to always prefer the theme except when the user set sometimes. 
RubAbstractTextArea class >> backgroundColor [
  ^ BackgroundColor ifNil: [ self theme backgroundColor ]
]

guillep:
I've moved the base branch and the milestone to Pharo8, according to issue https://pharo.fogbugz.com/f/cases/22833/

Ducasse:
Cyril may be we should close this one and check again, for the next one.

jecisc:
Yes. We should redo it with smaller PR in P8.

estebanlm:
conflicting now

Ducasse:
We will revisit it with Cyril soon.

estebanlm:
then please reopen the PR when you are ready :)

PR #2160, opened 2019-01-02
22859-PrettyPrinter-eats-space-on-large-selectors

Ducasse:
just make sure that a character and not a block is returned.


Ducasse:
https://pharo.fogbugz.com/f/cases/22859/PrettyPrinter-eats-space-on-large-selectors

At the end it was not a bug but leads me to do one in the new version or mixing something.

Ducasse:
The failing test is not related. It is a GLM UI test

PR #2162, opened 2019-01-02
22860-Re-introduce-new-rubric-classes

jecisc:
RÃ©introduce RubPluggableTextFieldMorph

Fixes https://pharo.fogbugz.com/f/cases/22860/Re-introduce-new-rubric-classes


jecisc:
For now I only added back RubPluggableTextFieldMorph.

I don't know if we should re-introduce the others in P7.

PR #2168, opened 2019-01-03
22863-Improve-Stability-of-Glamour-Tests

tesonep:
Adding a superclass to the Glamour tests to handle the waiting for UI redrawn in a more tidy way.

Glamour Tests are using directly World >> #doOneCycle.
This produces a problem when executing in the CI.
As the tests in the CI are run in another process different than the UI process, if we do #doOneCycle there is the risk of race conditions.

This will improve the stability of the build in the CI.

Issue: https://pharo.manuscript.com/f/cases/22863/Improve-Stability-of-Glamour-Tests


Ducasse:
Thanks Pablo this is a good change. Removing false positive will make us more agile.

PR #2173, opened 2019-01-04
22865-Metacello-Baselines-using-the-copy-functionality-over-a-p

gcotelli:
https://pharo.manuscript.com/f/cases/22865/


estebanlm:
damn. 
I just realised. 
I would have left also a comment explaining why that method is added. 
Time to time we have the "cleaning fever" and we remove methods that seems of no use :(

PR #2177, opened 2019-01-07
22866-Integrate-Alains-changes-to-Rubric

jecisc:
Redo changes of Alain keeping only the ones related to Pharo 7.

Fixes https://pharo.fogbugz.com/f/cases/22866/Integrate-Alains-changes-to-Rubric


jecisc:
Do not integrate. We have failing tests. I'll see with Alain if we should still push that in P7.

jecisc:
Too complicated for P7.

PR #2179, opened 2019-01-08
22871-MC-makes-dirty-more-packages-than-expected

pavel-krivanek:
https://pharo.fogbugz.com/f/cases/22871/MC-makes-dirty-more-packages-than-expected

fix package resolving from a category


pavel-krivanek:
in the CI artefact, the Calypso-SystemQueries is no more marked as dirty

jecisc:
Failure not related.

PR #2180, opened 2019-01-08
22224-ZipArchiveMember-still-uses-deprecated-RWBinaryOrTextStream-and-MultiByteBinaryOrTextStream

guillep:
Remove dependencies from Compression to old Streams
 - use regular streams on strings and bytearrays
 - clean users
 - remove unused and obscure "test" methods that log into files


svenvc:
Looks good, much simpler

guillep:
I've moved the issue to be integrated in Pharo8, I don't think this is relevant enough for pharo7.

PR #2195, opened 2019-01-08
22831-Methods-compiled-in-classes-from--UnpackagedPackage-does-no-have-package

tesonep:
An unpackaged method should return the default package.
Issue: https://pharo.manuscript.com/f/cases/22831/Methods-compiled-in-classes-from-_UnpackagedPackage-does-no-have-package


dionisiydk:
It looks like workaround and can have side effects. 
What is the real reason of this issue? Why _UnpackagedPackage does not behaves like a normal package.
Problem with this fix is that following invariant will be broken:
```
aMethod package includesSelector: aMethod selector ofClass: aMethod origin 
```
(and we should really improve this package API in Pharo8)

guillep:
Yes, let's do it to Pharo8. Calypso has a lot of issues and workaround managing unpackaged code anyways and the unpackaged package support must be revised.

Also this issue is not so critical, working on an unpackaged package is a borderline usecase.

I close this issue as we discussed with Pablo

PR #2198, opened 2019-01-08
Fix Pharo home URL in WebBrowser example method

sroccaserra:
`WebBrowser example` method was opening a browser on
<http://www.pharo.org>, (which currently displays a directory with a
`springer.png` file through Apache by the way).

It now opens a browser on <http://pharo.org>, which is the URL shown in
the github description.

### DIFF
--- a/src/WebBrowser-Core/WebBrowser.class.st
+++ b/src/WebBrowser-Core/WebBrowser.class.st
@@ -10,7 +10,7 @@ Class {
 { #category : #examples }
 WebBrowser class >> example [
 
-	self openOn: 'http://www.pharo.org'
+	self openOn: 'http://pharo.org'
 ]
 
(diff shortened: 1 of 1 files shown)


jecisc:
I'm not sure we should change the URL in the image. I would prefer that we fix the website.

guillep:
I'd do whatever works now. I think this is a good cheap solution. The server is broken since already some time so this is the most efficient way to solve it.

astares:
Back when I wrote the WebBrowser package it was working. Any glue what we changed on the server meanwhile? Should'nt we nonetheless fix the server? Because www.pharo.org would also be a common 
URL people may try...

estebanlm:
yes.
what changed is the server itself. 
I think @demarey can do something?

PR #2204, opened 2019-01-09
Removing leftover halts

dupriezt:
Or adding the <haltOrBreakpointForTesting> to methods that are meant to contain halts in production. This way they won't appear in the breakpoint browser that got integrated into Pharo 7 with the new version of Calypso.

### DIFF
--- a/src/Kernel-Tests/ProtoObjectTest.class.st
+++ b/src/Kernel-Tests/ProtoObjectTest.class.st
@@ -17,32 +17,53 @@ ProtoObjectTest >> testFlag [
 
 { #category : #'tests - testing' }
 ProtoObjectTest >> testIfNil [
-	| object block |
+	| object block reachabilityTest|
+	reachabilityTest := false.
 	object := ProtoObject new.
-	object ifNil: [ self halt ].
-	self assert: (object ifNil: [ nil ]) == object.	"Now the same without inlining."
-	block := [ self halt ].
+	object ifNil: [reachabilityTest := true ].
+	self assert: (object ifNil: [ nil ]) == object.
+	self assert: reachabilityTest equals: false.
+	"Now the same test without inlining."
+	reachabilityTest := false.
+	block := [ reachabilityTest := true].
 	object ifNil: block.
 	block := [ nil ].
-	self assert: (object ifNil: block) == object
+	self assert: (object ifNil: block) == object.
+	self assert: reachabilityTest equals: false.
 ]
 
 { #category : #'tests - testing' }
 ProtoObjectTest >> testIfNilIfNotNil [
 
-	| object returnValue block |
+	| object returnValue block notReached1 reached1 notReached2 reached2 reached3 notReached3 reached4 notReached4 |
+	notReached1 := false. reached1 := false. notReached2 := false. reached2 := false.
+	notReached3 := false. reached3 := false. notReached4 := false. reached4 := false.
 	object := ProtoObject new.
 	returnValue := Object new.
-	self should: [ object ifNil: [ self error ] ifNotNil: [ self halt ] ] raise: Halt.
-	self should: [ object ifNil: [ self error ] ifNotNil: [ :o | self halt ] ] raise: Halt.
+
+	object ifNil: [ notReached1 := true ] ifNotNil: [ reached1 := true ].
+	self assert: notReached1 equals: false.
+	self assert: reached1 equals: true.
+
+	object ifNil: [ notReached2 := true ] ifNotNil: [ :o | reached2 := true ].
+	self assert: notReached2 equals: false.
+	self assert: reached2 equals: true.
+
 	self assert: (object ifNil: [ false ] ifNotNil: [ :o | o == object ]).
 	self assert: (object ifNil: [ nil ] ifNotNil: [ returnValue ]) == returnValue.
 	self assert: (object ifNil: [ nil ] ifNotNil: [ :o | returnValue ]) == returnValue.
+	
 	"Now the same without inlining."
-	block := [ self halt ].
-	self should: [ object ifNil: [ self error ] ifNotNil: block ] raise: Halt.
-	block := [ :o | self halt ].
-	self should: [ object ifNil: [ self error ] ifNotNil: block ] raise: Halt.
+	block := [ reached3 := true ].
+	object ifNil: [ notReached3 := true ] ifNotNil: block.
+	self assert: notReached3 equals: false.
+	self assert: reached3 equals: true.
+
+	block := [ :o | reached4 := true ].
+	object ifNil: [ notReached4 := true ] ifNotNil: block.
+	self assert: notReached4 equals: false.
+	self assert: reached4 equals: true.
+	
 	block := [ :o | o == object ].
 	self assert: (object ifNil: [ false ] ifNotNil: block).
 	block := [ returnValue ].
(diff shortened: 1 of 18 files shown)


tesonep:
@dupriezt there is an issue in fogbugz for this? It should be integrated in Pharo7? or Pharo8?

src/Kernel-Tests/ProtoObjectTest.class.st
```diff
@@ -78,19 +106,38 @@ ProtoObjectTest >> testIfNotNil [
...
+	block := [ reached := true ].
+	object ifNotNil: block ifNil: [ notReached := true ].
+	self assert: reached equals: true.
+	self assert: notReached equals: false.
+	
+	reached := false. notReached := false.	
+	block := [ :o | reached := true ].
+	object ifNotNil: block ifNil: [ notReached := true ].
+	self assert: reached equals: true.
+	self assert: notReached equals: false.
```
guillep, line 139: Is really all this repeated code needed?

src/Manifest-Resources-Tests/MFClassA.class.st
```diff
@@ -10,5 +10,5 @@ Class {
 { #category : #'as yet unclassified' }
 MFClassA >> method [
 	|foo|
-	self halt.
+	1+1.
```
guillep: Why not just leaving an empty method returning `self`?

src/Manifest-Resources-Tests/MFClassB.class.st
```diff
@@ -15,5 +15,5 @@ MFClassB >> method2 [
 { #category : #'as yet unclassified' }
 MFClassB >> method3 [
 	|foo|
-	self halt.
+	1+1.
```
guillep: Why not just leaving an empty method returning `self`?

jecisc:
Since it make the Breakpoint browser more usable I think it night be good for P7.

guillep:
Let's wait until all tests run. I don't know if this will have an effect on them or not.

dupriezt:
The two tests that fail are:
- unix-32 / Tests-unix-32 / testSizeAndSpan – Unix32.Glamour.Tests.Morphic.GLMTabulatorMorphicTest
- windows-32 / Tests-windows-32 / testAliasCompositions – Windows32.Tests.Traits.TraitCompositionTest

I tried the first one on my linux, and it passes...
What should I do?

src/Manifest-Resources-Tests/MFClassA.class.st
dupriezt: I don't know. Do you want me to commit an empty method instead?

src/Manifest-Resources-Tests/MFClassB.class.st
dupriezt: I don't know. Do you want me to commit an empty method instead?

src/Kernel-Tests/ProtoObjectTest.class.st
dupriezt: It's not testing exactly the same thing each time, so... maybe?
I don't know :-)

jecisc:
They are unrelated. I can relaunch it later.

tesonep:
This PR should be carefully addressed. It is changing the semantics of the tests.

The tests are not well written, but they are using the halts instead of asserts or failures. 

For example, the proto object test is testing that the blocks are not executed in if the receiver is not nil, and it is testing it with and without parameter in the block. 

A test should not be modified without understanding it. Unless we have tests for the tests.

I rather prefer a false positive in a tool than losing a test

dupriezt:
I'm not an expert, but it seems the intention of these tests on ProtoObject was just to check that the correct branch of the ifNil (and flavours) is taken each time. Which is preserved by using booleans to track which branch was taken instead of exceptions.
Do you think I should re-write these tests to use another kind of exceptions than halts and errors to check which branches are taken?

tesonep:
@dupriezt I have re-checked and you are right!

src/Manifest-Resources-Tests/MFClassA.class.st
guillep: An empty method would indeed be better.
Moreover, putting a message saying why that method is empty even better.

src/Manifest-Resources-Tests/MFClassB.class.st
guillep: An empty method would indeed be better.
Moreover, putting a message saying why that method is empty even better.

src/Manifest-Resources-Tests/MFClassA.class.st
guillep: Imagine in a couple of months the person that will read that 1+1 method that does not return and think it's a bug.

src/Manifest-Resources-Tests/MFClassB.class.st
guillep: Imagine in a couple of months the person that will read that 1+1 method that does not return and think it's a bug.

guillep:
I'd be ok to integrate this. But we should make an issue to enhance it in Pharo8 I think.

For example, notReached := true  can be just `self fail`

src/Manifest-Resources-Tests/MFClassA.class.st
dupriezt: Done 1227d92

src/Manifest-Resources-Tests/MFClassB.class.st
dupriezt: Done 1227d92

dupriezt:
I removed the dummy code from the methods Guille pointed at earlier in this discussion. These methods are now empty.

guillep:
There are 4 tests failing in all platforms that look related to this issue

image

guillep (changes requested):
We need a fogbugz issue for this PR, and the branch has to be named correspondingly, otherwise the CI server cannot correctly send the corresponding issues. This has been said a week ago.

Also, the PR has broken a couple of tests, probably related to the fact of leaving unused temporary variables.

estebanlm:
yup
I think is too late for Pharo7, I will move this to Pharo8. 
Good news is that with that you do not need a fogbugz issue but a GitHub issue ;)

dupriezt:
Thanks for the list of failing tests @guillep I looked into them:
- testCriticsOf fails because I removed the unused variables from MFClassA>>method, MFClassB>>method2 and MFClassB>>method3 and this test was checking that these methods were indeed triggering a "variable not read or written" critics rule. I fixed that in my last commit, and added a comment to these methods.
- testIsToDo fails because... I'm not really sure why. But adding a line of code to MFClassB>>method2 makes it pass so I did that and added a comment to the method.
- same thing for testToDoOf
- testNoUnusedTemporaryVariablesLeft fails because of the way it is written. It checks that there are not temporary variables unused in the system. To do that it builds the list of all methods with unused temporaries, remove from it a list of exceptions (method, method2 and method3 from above) from it and checks that the resulting list is empty. So when I removed the unused variables from these methods, the removeAll: fails because it tries to remove elements that are not in the list and that's what makes the test fail. It now passes with my last commit that put back the unused variables in method, method2 and method3, but I think there's maybe a design problem with the way this test is written.

dupriezt:
I made a new pull request with the same code but based on a github issue and with the correct name as requested: #2255
Closing this one.

PR #2215, opened 2019-01-10
22886 Zinc should allow custom server string

astares:
Allow to customize the server string for deployment scenarios relying on security. Here the server should not tell the 
outside world about its version or technology and not expose such confidential data to an attacker.

    ZnConstants defaultServerString: 'SecureServer'

### DIFF
--- a/src/Zinc-HTTP/ZnConstants.class.st
+++ b/src/Zinc-HTTP/ZnConstants.class.st
@@ -21,6 +21,7 @@ ZnConstants class >> defaultHTTPVersion [
 
 { #category : #accessing }
 ZnConstants class >> defaultServerString [ 
+	
 	^DefaultServerString ifNil: [ DefaultServerString := self frameworkNameAndVersion , ' (' , self systemVersion , ')' ]
 
 ]
--- a/src/Zinc-HTTP/ZnConstants.class.st
+++ b/src/Zinc-HTTP/ZnConstants.class.st
@@ -8,6 +8,7 @@ Class {
 	#superclass : #Object,
 	#classVars : [
 		'DefaultMaximumEntitySize',
+		'DefaultServerString',
 		'HTTPStatusCodes'
 	],
 	#category : #'Zinc-HTTP-Support'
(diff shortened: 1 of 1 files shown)


astares:
Bug is in https://pharo.fogbugz.com/f/cases/22886/

astares:
To verify

"Start a server"
ZnServer startDefaultOn: 8080.

"Check the response"
(ZnEasy get: 'http://localhost:8080') 

"and click on >>Headers<< in the inspector"

estebanlm:
hi @svenvc 
can you check at this?

svenvc:
This was my reply in the issue tracker:

I understand the issue, but I do not agree with the fact that this is a must fix bug for Pharo 7, far from it.

It is relatively easy to overwrite the default Server header on a per server instance granularity today. And that is what you want, not a global default. Consider:

```
ZnServer startDefaultOn: 8080.

ZnServer default delegate in: [ :delegate |
	ZnServer default delegate: (ZnValueDelegate with: [ :request | | response | 
		response := delegate handleRequest: request.
		response headers at: #Server put: 'Google'.
		response ]) ].

(ZnClient new get: ZnServer default localUrl; response) headers at: #Server.
```

The above wraps any existing delegate, lets the original delegate do its work, and then modifies the response.

I also think very few people will ever need to make this change. Like you said, it is part of production hardening.

For me, Enhancement, Would be nice, Pharo 8

astares:
I would add it as it is a non-breaking change (the name is now only cached in a class var. the class var can additionally be set from the outside) 

But if P7 is already closed I have to wait another year :(

svenvc:
Torsten,

Fear not, I will make a change to Zn to improve the situation with configurations, parameters, settings and options. I think/hope I have a good design for a solution that both simplifies things and adds more control. 

Sven

astares:
Thanks. I stay patient. Let's close the PR

svenvc:
This will be handled upstream, see https://github.com/svenvc/zinc/issues/27

PR #2235, opened 2019-01-13
22894-Unecessary-dialogs-during-refactorings

dionisiydk:
It updates Calypso: v0.15.1 which fixes the issue

### DIFF
--- a/src/BaselineOfIDE/BaselineOfIDE.class.st
+++ b/src/BaselineOfIDE/BaselineOfIDE.class.st
@@ -845,7 +845,7 @@ BaselineOfIDE >> loadCalypso [
 
 	Metacello new
 		baseline: 'Calypso';
-		repository: 'github://pharo-ide/Calypso:v0.15.1/src';
+		repository: 'github://pharo-ide/Calypso:v0.15.2/src';
 		onConflict: [ :err | err useIncoming ];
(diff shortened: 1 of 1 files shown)


dionisiydk:
https://pharo.manuscript.com/f/cases/22894/Unecessary-dialogs-during-refactorings

dionisiydk:
VM crash is unrelated.
The fix was really trivial: it removes explicit first item selection from initiaization of scope drop list. Drop list itself selects first item when new list is specified. So extra #selectItem: was redundant and it leaded to double selection handler.

guillep:
@jecisc since you're the one invoved in the calypso issue, do you think this is for Pharo7?

jecisc:
Definitively.

You do not want to have two times the same warnings each time you rename a method. That's something we do often.

dionisiydk:
Notice it is one line change

пн, 14 янв. 2019 г., 9:48 CyrilFerlicot notifications@github.com:

dionisiydk:
I updated version to 0.15.2 to include another fix (also one line) for broken setUp method tab

dionisiydk:
And version 0.15.3 includes fix for class remove command:
https://pharo.manuscript.com/f/cases/22895/Cannot-remove-a-class-that-has-spurious-references-to-it

MarcusDenker:
error not related.

PR #2238, opened 2019-01-14
Restore #getEnv:

StephanEggermont:
that used to be public API

The new impl handles the characters outside ASCII. For the strings that were valid before the new impl will return exactly the same. The change is for the environment values using characters with codepoint > 127. For example accented leters or japanese characters
We got to this, because the launcher is used by users with usernames in french, spanish or japanese

### DIFF
--- a/src/System-OSEnvironments/OSEnvironment.class.st
+++ b/src/System-OSEnvironments/OSEnvironment.class.st
@@ -6,19 +6,9 @@ Get access using:
 
 Low level API
 
-- getEnv: aVariableName
-Gets the value of an environment variable called `aVariableName`
-It is the system reponsibility to manage the encoding.
-Rationale: A common denominator for all platforms providing an already decoded string, because windows does not (compared to *nix systems) provide a encoded byte representation of the value. Windows has instead its own wide string representation.
-
-- getEnvRaw: anEncodedVariableName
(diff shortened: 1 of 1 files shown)


svenvc:
Are you sure ?
This looks strange...

src/System-OSEnvironments/OSEnvironment.class.st
```diff
@@ -187,6 +177,14 @@ OSEnvironment >> do: aBlock [
 	^self valuesDo: aBlock
 ]
 
+{ #category : #accessing }
+OSEnvironment >> getEnv: aKey [
```
dionisiydk, line 181: Should not it be done with deprecation?

StephanEggermont: See 22658

estebanlm:
still... deprecation?

guillep:
I think readding it with a proper deprecation is the right thing to do.
I'll create an issue for this in fogbugz and introduce the deprecation if you don't mind.

guillep:
I've remade the PR with an attached fogbugz issue, including better class comments and the corresponding deprecation.

https://github.com/pharo-project/pharo/pull/2252/files

Please let's continue the conversation there

PR #2243, opened 2019-01-15
2242-Sync-Pharo-7-and-8

### LINKED ISSUE #2242: Sync Pharo 7 and 8
I think we should do a sync of Pharo 7 and 8 to ensure new contribution do not conflict.

jecisc:
Merge branch 'Pharo7.0' into 2242-Sync-Pharo-7-and-8

Fix #2242


jecisc:
Oh, apparently it was done not long ago. So I'm not sure this is so useful.

PR #2245, opened 2019-01-15
2244-NECPreferences-set-values-at-two-places

### LINKED ISSUE #2237: NECPreferences should not access setting for menu color
We currently have that in NEC:

```Smalltalk
NECPreferences class>>backgroundColor 
	^backgroundColor ifNil: [ 
		backgroundColor := (Smalltalk ui theme settings menuColor ifNil: [ Color white ]) ]
```

I think we should have: 

```Smalltalk
backgroundColor 
	^ backgroundColor ifNil: [ backgroundColor := self theme menuColor ]
```

Because menu color cannot be nil, we should not reference Smalltalk for theme in morphs and we should not directly access settings in NEC.

### LINKED ISSUE #2244: NECPreferences set values at two places
In the NECPreferences class, we have an initialize method to initialize all the values but we also have defensive strategies for multiple items. 

One time we also have different values in both of them.

We should unify and keep only one mechanism.

jecisc:
Clean NECPreferences

- Remove defensive strategy when it's already initialized
- Fix value of backgroundColor to use theme value instead of theme settings value
- Correct clash for #useEnterToAccept value
- Format 

Fixes #2244
Fixes #2237


Ducasse:
Cyril could you introduce a method named theme that is Smalltalk ui theme
I would really like to remove as many as possible of these TERRIBLE Smalltalk ui BAD idea from the system

jecisc:
@Ducasse I was thinking about it. At the beginning I also did it but in the end I saw that it was called only one time at one place so I was wondering if it was necessary. 

But I can do it. I'll add it on my todolist.

jecisc:
@Ducasse Done.

Ducasse:
Cool I have the impression that a bug somewhere is breaking the build :(

estebanlm:
honestly, I think a method `#theme` is as bad as `Smalltalk ui` etc. 
(I think same about `#iconNamed`, btw).
Reason is we will populate `Object` with this methods and I do not like it. I would prefer to have just **one** method to rule this... which IMO should be `environment` (where `environment` is usually `Smalltalk`, but user can change the environment as desired).

Then, you can have: 

```Smalltalk
self environment classNamed: aName.
self environment theme backgroundColor.
self environment icons iconNamed: #ok.
```

(I know, not the place to discuss it, but we need to discuss it at a point :) )

estebanlm:
(I didn't reopened the PR because of that, btw... it was just to force the rebuild and test ;) )

PR #2255, opened 2019-01-16
2254 leftover halts

dupriezt:
I made another pull request on this (with the same code) to have it based on a github issue and with a matching name.
Original pull request (with all the discussions): #2204

### DIFF
--- a/src/Debugger-Tests/FilterTest.class.st
+++ b/src/Debugger-Tests/FilterTest.class.st
@@ -18,7 +18,7 @@ Class {
 { #category : #running }
 FilterTest >> setUp [
 	super setUp.
-	context := [ (Set with: 1 with: 2) collect: [ :e | e * 2 ]. self halt ] asContext.
+	context := [ (Set with: 1 with: 2) collect: [ :e | e * 2 ]] asContext.
 	process := Process 
 		forContext: context 
 		priority: Processor userInterruptPriority.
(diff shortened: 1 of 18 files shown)


dupriezt:
On the tests that were failing (I fixed them with 3b8fbed):
- testCriticsOf was failing because I removed the unused variables from MFClassA>>method, MFClassB>>method2 and MFClassB>>method3 and this test was checking that these methods were indeed triggering a "variable not read or written" critics rule. I fixed that in my last commit, and added a comment to these methods.
- testIsToDo was failing because... I'm not really sure why. But adding a line of code to MFClassB>>method2 makes it pass so I did that and added a comment to the method.
- same thing for testToDoOf
- testNoUnusedTemporaryVariablesLeft was failing because of the way it is written. It checks that there are not temporary variables unused in the system. To do that it builds the list of all methods with unused temporaries, remove from it a list of exceptions (method, method2 and method3 from above) from it and checks that the resulting list is empty. So when I removed the unused variables from these methods, the removeAll: fails because it tries to remove elements that are not in the list and that's what makes the test fail. It now passes with my last commit that put back the unused variables in method, method2 and method3, but I think there's maybe a design problem with the way this test is written.

guillep:
Arghhh this was created with a github issue instead of a fogbugz issue...

dupriezt:
> Arghhh this was created with a github issue instead of a fogbugz issue...

Because Esteban said it was too late for pharo 7...
Thanks @guillep

PR #2257, opened 2019-01-16
Fixes Issue #2256 Recover parameter names, remove unneeded temps in Character

### LINKED ISSUE #2256: Character>>#printOn: has decompiled source
the method `Character>>#printOn:` has only the decompiled code so the real argument name and temporaries are lost. 

arg1 = aStream
tmp1 = name

StephanEggermont:
Sources file seems to have been missing a long time ago (recovered from Pharo 4)

### DIFF
--- a/src/Kernel/Character.class.st
+++ b/src/Kernel/Character.class.st
@@ -711,27 +711,16 @@ Character >> lowercase [
 ]
 
 { #category : #printing }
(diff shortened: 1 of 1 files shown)


guillep (approved):
Just cosmethic issues. I've checked them, they look ok.

PR #2267, opened 2019-01-17
22887-TonelWritercreateDefaultOrganizationFrom-can-return-string

estebanlm:
update iceberg to 1.5.5

https://pharo.manuscript.com/f/cases/22887/TonelWriter-createDefaultOrganizationFrom-can-return-string


pavel-krivanek:
I think that the Tonel version should be updated in pharo/bootstrap/src/BaselineOfPharoBootstrapProcess.package/BaselineOfPharoBootstrapProcess.class/instance/baseline..st  too

Issue #2270, opened 2019-01-18
Context>>methodReturnReceiver duplicates the code of Context>>home


dupriezt:

MarcusDenker:
merged

MarcusDenker:
hmm... how does this work, I guess we need to close?

MarcusDenker:
fixed

PR #2276, opened 2019-01-18
 Context>>methodReturnContext now uses Context>>home instead of repli…

dupriezt:
Solves #2270


MarcusDenker (approved):
ok

MarcusDenker:
errors are not related

PR #2280, opened 2019-01-21
2303-Fix deprecation comments

StephanEggermont:
The preview is only needed when provided (and currently not implemented)


Ducasse:
Hi stephan 
can you open a bug entry else we will have to do it?
Tx

StephanEggermont:
Fixes #2303

estebanlm:
damn it
it was opened against P7

PR #2281, opened 2019-01-21
2294-Halt-on-count-logic-fix

### LINKED ISSUE #2294: Halt on count logic fix
Halt onCount: now only halt once when the count is reached.
Halt fromCount: will always halt after the count is reached.

StevenCostiou:
Halt onCount: now only halt once when the count is reached.
Halt fromCount: will always halt after the count is reached.

Fixes #2294


bencoman:
There seem to be a lot of commits unrelated to the PR title.  Is this intended? I don't know the impact of merging them all.

StevenCostiou:
Nope. I will fix it.

Ducasse:
Hi Steven 

This is cool! Now 
Could you add some method comments? For haltAfterCount: and haltAtCount:?
I would like that we ask some native english to help us for these names.

I do not really understand what means haltFromCount: and haltCount: 
Is it after you reach a number you halt all the times? (I would rename the method haltAfterCount:) 

What is haltCount:? Does it stop a number of times and stop? Or does stop only after that number has been reached?

Stef

StevenCostiou:
I will add comments.

The original method was haltOnCount: or Halt onCount: but it was halting every time after reaching the count number.  It has been renamed haltFromCount; haltAfterCount would also work.

The new haltOnCount only halts once after the count is reached. It could be renamed haltOnceOnCount.

I would also welcome help from native speakers :)

Ducasse:
Ok waiting for your push

StevenCostiou:
I commented, but i did not change the naming. It fixes the logic, although it also depends on the naming. Comments are welcome.

Ducasse:
Ok steven I will integrate it and after we should really rename these methods. 
Can you add it to your hacking todo?

StevenCostiou:
Yes, I cannot decide the renaming alone ;)

PR #2307, opened 2019-01-23
[IGNORE]2306-Remove-Deprecated70

### LINKED ISSUE #2306: Remove Deprecated70

jecisc:
Remove Deprecated 70.

Fixes #2306


Ducasse:
Cyril I think that we should not remove deprecated70.
Some methods should be kept to be part of migrator (a package that people can use to help the migration of the code). There are probably some methods that we do not want to keep. I would prefer to do it in P8 that in P7. And I'm not sure how to handle this well. 
I collected already some methods in Migrator on SmalltalkHub.

jecisc:
We can keep the PR and integrate it when the work on Migrator will be finished.

astares:
@Ducasse @jecisc @estebanlm @pavel-krivanek 
We should remove Deprecated70 as we do on any new release cycle. This would also reveal trouble
that we havent thought about yet.
Developers of migrator can always load the Deprecated70 packages back if they need it, no?

Anyhow:
At least we should introduce a Deprecated80 as quick as possible.  

This would be helpful for issues like https://github.com/pharo-project/pharo/issues/2387

Ducasse:
Let us make sure that we take the time to think how to proceed and to avoid pain. 
Tx

Ducasse:
Next week we should discuss how we can handle that nicely. The use case scenario is not obvious for me (basically I imagine that a user that want to migrate from p5 to p7 would like to have p7 and p6 deprecation, while the one that only migrate from p6 to p7 only p7 deprecations). 
I will start to publish the deprecated method in migrator.

Ducasse:
I imagine that a user that want to migrate from p5 to p7 would like to have p7 and p6 deprecations, 
while the one that only migrate from p6 to p7 wants only p7 deprecations and so on.

I can have a 
	migratorFromP6 includes deprecation for P6 and P7
	migratorFromP7  includes deprecation for P7

and each time a new version arrives I add the deprecations of the new version making sense to each of them. 

But thinking about it, I will simply do one package for each pharo version and 
the guy wanted to migrate from p6 to p8
will simply load migratorFromP6, migratorFromP7 and migratorFromP8 
like that a same package can be use in different scenario (migrating from p6 to p7 or from p6 to p9)

What do you think.

Ducasse:
I created MigratorFromPharo60To70 and MigratorFromPharo50To60 in https://github.com/Ducasse/Migrator so that we can deprecate the initial effort in SmalltalkHub/PharoExtras (now marked as such).

Now I "just" packaged deprecated packages. I created several packages that can be reloaded but we should pay attention because loading MigratorFromPharo50To60 in Pharo 80 may break Pharo80 (this is the case with Spotter). Based on such packages we can see if we want to produce MigratorFromPharo50To80

jecisc:
What is the status of this PR? Should we still hold?

jecisc:
Ping

jecisc:
Still waiting, don't stale

PR #2314, opened 2019-01-23
2309-Add-assertIsEmpty

### LINKED ISSUE #2309: Add #assertEmpty:
I would like to introduce an #assertIsEmpty: because when a test fails on the CI because a collection is not empty, it's hard to debug without seeing the collection.

jecisc:
Introduce #assertEmpty: and #denyEmpty: + use them

The commit also contains some cleanings such as replacing some `assert: something = true` by `assert:` etc.

Fixes #2309

### DIFF
--- a/src/AST-Tests-Core/RBParserTest.class.st
+++ b/src/AST-Tests-Core/RBParserTest.class.st
@@ -304,18 +304,12 @@ RBParserTest >> testIsA [
 	| nodes types |
 	nodes := Bag new.
 	types := Set new.
-	#(#(#isAssignment 1) #(#isBlock 1) #(#isCascade 1) #(#isLiteralNode 2) #(#isMessage 3) #(#isMethod 1) #(#isReturn 1) #(#isSequence 2) #(#isValue 15) #(#isVariable 7) #(#isUsed 10) #(#isDirectlyUsed 9) #(#hasParentheses 1) #(#isBinary 0) #(#isPrimitive 0) #(#isImmediateNode 10) #(#isWrite 1) #(#isRead 3)) 
-		do: 
-			[:each | 
-			each last timesRepeat: [nodes add: each first].
-			types add: each first].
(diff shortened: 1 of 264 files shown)


Ducasse:
I relaunched the CI.

jecisc:
@svenvc My proposal impacts Zinc tests. If this is approved and pass checks (when the CI will be back), do you agree to integrate those changes to Zinc with a compatibility package that would be loaded in Pharo < 8? I can do the PR to do that.

astares:
I'm not convinced we should have it. See my notes in #2309

svenvc:
Hi Cyril,

I understand the rationale, but I think the effect on Zn will be more negative than positive (as I try to maintain compatibility over more Pharo versions), this is also not worth the trouble of an extra compatibility package. The gain is too small, this is more of a style difference to me.

I hope you understand,

Sven

jecisc:
@svenvc Ok

Ducasse:
So what do we do? I would add it because this is nice and after people can use it or not.

jecisc:
I am waiting for the CI to work so that I can check I did not introduce any regression.

Ducasse:
Argh Cyril the package GTSpotter-EventRecorder was removed! Can you resubmit your changes?
Tx

astares:
If you put objections aside then why not: 
 - add the new methods #assertIsEmpty: only in this PR as the subject says. 
 - change the tests that should really use it in separate PR(s)

Ducasse:
Cyril can you check because you have conflict with the removed packages?

jecisc:
I'll do as Astares said to split in 2 PR because I agree it's a really big PR and if a problem arises, it will be easier to repair if we have two PR.

jecisc:
First PR: https://github.com/pharo-project/pharo/pull/2400
Second PR: https://github.com/pharo-project/pharo/pull/2401

PR #2322, opened 2019-01-24
2321-The-dark-age-is-coming

### LINKED ISSUE #2321: The dark age is coming
As a result of switching of themes with every release, we should change the default theme to the dark one again

pavel-krivanek:
Switch to the dark theme. 
Fixes #2321

### DIFF
--- a/src/BaselineOfIDE/BaselineOfIDE.class.st
+++ b/src/BaselineOfIDE/BaselineOfIDE.class.st
@@ -90,7 +90,7 @@ BaselineOfIDE >> additionalInitialization [
 	
 	RubAbstractTextArea highlightMessageSend: true.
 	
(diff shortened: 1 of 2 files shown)


astares:
Ohhh - no - not again a dark age... 

So only Pharo 9 will be bright again. ;)

Ducasse:
I would like to have the blueish one esteban is using.

astares:
@Ducasse Blueish one? Any pointers or code to load?

pavel-krivanek:
https://github.com/estebanlm/themes

astares:
Thanks for the pointer

estebanlm:
I think my dark metal theme (that I love) is too much for regular usage :)

Issue #2275, opened 2019-01-18
Athens test3,4 do not work


Ducasse:
AthensTextRenderTest class >> test3

psteeve:
@Ducasse this should be closed i think.

MarcusDenker:
PR was merged: close

PR #2326, opened 2019-01-25
Athens test3, 4 do work.

570gina:
We changed the Text to a String. Because  the glyph renderer was expecting a string.
Also the form is now accessed by asForm.


tesonep:
It fixes #2275

Ducasse:
Thanks Chia!

PR #2329, opened 2019-01-25
2164-GrowlMorph-has-duplication

### LINKED ISSUE #2164: GrowlMorph has duplication
Look at the constructors in the class side. They all reconstruct everything. We should have the behavior in one constructor and the others should call this one.

pavel-krivanek:
clean GrowlMorph constructors

fixes #2164


Ducasse:
I relaunched the build

PR #2339, opened 2019-01-26
2338 GT Extensions should also be visible on empty directories

### LINKED ISSUE #2338: GT Extensions should also be visible on empty directories
If you use 
   System -> Startup -> Version Preference Folder
   System -> Startup -> General Preference Folder  
an inspector with GT Extension button is shown. One can use the buttons to create a file (like startup.st) 

But the directories might be empty. Then it is not possible as the GT extensions are not shown.

Same issue if you use
```Smalltalk
   FileSystem workingDirectory inspect
```
where you see the GT extension in the inspector
while 
```Smalltalk
   (FileSystem workingDirectory / 'foo') ensureCreateDirectory 
```
does not show them (assuming foo is a new directory)

astares:
Fixes #2338


astares:
Maybe we should also backport this to Pharo 7

estebanlm:
thanks :)
@astares, if you put 
Fixes (or closes) and the issue number (as I modified your comment), issue will be  automatically referenced there (no need to go and put a comment). 
In addition, issue will be automatically closed when merged.

PR #2344, opened 2019-01-26
2342-memory-leak-through-GTSnippets-instance-cache

j-brant:
making the snippets cache use a WeakIdentityKeyDictionary (2342)


MarcusDenker (approved):
ok

Ducasse:
Marcus apparently jenkins is dead so I'm waiting to integrate because I do not know if it will work.

Issue #2345, opened 2019-01-26
envversion.sh causes out-of-repository bootstrap to fail


akgrant43:
Recent changes to envversion.sh assume that the bootstrap is being run from within the repository, which is not required (BOOTSTRAP_REPOSITORY can be set to point to the repository).

The solution is to change directory to BOOTSTRAP_REPOSITORY before issuing the git commands, and then change back.

PR to be submitted soon.

Ducasse:
Tx!!

akgrant43:
PR: https://github.com/pharo-project/pharo/pull/2346

akgrant43:
Merged.

PR #2346, opened 2019-01-26
Issue 2345 envversion.sh causes out-of-repository bootstrap to fail

akgrant43:
Recent changes to envversion.sh assume that the bootstrap is being run
from within the repository, which is not required (BOOTSTRAP_REPOSITORY
can be set to point to the repository).

The solution is to change directory to BOOTSTRAP_REPOSITORY before
issuing the git commands, and then change back.

Issue: https://github.com/pharo-project/pharo/issues/2345


akgrant43:
The CI failure is due to the wrong Iceberg version, see #2347

akgrant43:
Thanks, Esteban!

estebanlm:
yeah, but no :)
seems to be breaking the CI. 
I will revert until we dig more :P

akgrant43:
Hi Esteban,

Can you point me to the log where the error is occurring?

Thanks,
Alistair

akgrant43:
OK, I think the problem is that transform_32_into_64.sh isn't calling envvars.sh.

akgrant43:
I've created https://github.com/pharo-project/pharo/pull/2368 to test the hypothesis.

Issue #2347, opened 2019-01-26
BaselineOfIDE references non-existant Iceberg 1.5.5


akgrant43:
BaselineOfIDE currently tries to load Iceberg 1.5.5, which doesn't exist.

Use V1.5.6 (previous available version is 1.5.3).  See https://github.com/pharo-vcs/iceberg/releases

PR to be submitted.

akgrant43:
Merged

PR #2348, opened 2019-01-26
2347 BaselineOfIDE references non-existant Iceberg 1.5.5

akgrant43:
BaselineOfIDE currently tries to load Iceberg 1.5.5, which doesn't
exist.

Use V1.5.6 (previous available version is 1.5.3). See
https://github.com/pharo-vcs/iceberg/releases

Issue: https://github.com/pharo-project/pharo/issues/2347


akgrant43:
CI failures appear to be unrelated: the fact that it built and got to run the tests got it past the issue.

Ducasse:
So esteban can we harvest fixes now?

akgrant43:
Thanks, Esteban!

PR #2364, opened 2019-01-27
2363 Add ProperManifestCategorizationTest as new release test

### LINKED ISSUE #2363: Add ProperManifestCategorizationTest as new release test
If a package provides a PackageManifest subclass then this manifest class should be packaged 
with 'Manifest' tag similar to all other manifests.

We cleaned the package manifests in https://pharo.fogbugz.com/f/cases/22436/PackageManifest-subclasses-should-be-packaged-with-Manifest-tag and subcases. 

Now we should add a new release test that keeps that quality level: Package manifests should not be untagged but tagged in a class category 'Manifest'.

astares:
Fixes #2363

### DIFF
--- a/src/ReleaseTests/ProperManifestCategorizationTest.class.st
+++ b/src/ReleaseTests/ProperManifestCategorizationTest.class.st
@@ -1,0 +1,23 @@
+"
+Check for proper Manifest categorization
+"
+Class {
+	#name : #ProperManifestCategorizationTest,
+	#superclass : #TestCase,
+	#category : #'ReleaseTests-Categorization'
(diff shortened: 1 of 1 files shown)


jecisc:
Can we wait for https://github.com/pharo-project/pharo/pull/2314 and use #assertEmpty: please? 

Release tests are the typical tests where this is the most useful.

astares:
There is no need to wait for 2314 ... can we move forward please?

astares:
2314 is closed, can we move on?

estebanlm:
checking the crash

PR #2368, opened 2019-01-27
transform_32_into_64.sh: call envvars.sh

akgrant43:
envvars.sh is required to set up environment variables such as
BOOTSTRAP_REPOSITORY.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -154,7 +154,7 @@ def bootstrapImage(){
         stage ("Convert Image - 32->64") {
           dir("conversion") {
             shell "cp ../bootstrap-cache/*.zip ."
-            shell "bash ../bootstrap/scripts/transform_32_into_64.sh"
+            shell "BUILD_NUMBER=${BUILD_NUMBER} BOOTSTRAP_ARCH=${architecture} bash ../bootstrap/scripts/transform_32_into_64.sh"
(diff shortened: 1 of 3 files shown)


pavel-krivanek:
causes the same CI failure

akgrant43:
Slightly different, I think.  Before it was BOOTSTRAP_REPOSITORY that was missing, now it's BOOTSTRAP_ARCH.

I'll try and track down why it isn't still defined (it obviously was earlier in the pipeline).

akgrant43:
I think this now solves the issue.  There are 4 existing test failures, however I don't think they're related (the tests wouldn't even get to run if there was a problem with this PR).

demarey:
I will merge this PR because builds are failing mainly due to `BOOTSTRAP_REPOSITORY: unbound variable`. This PR looks ok.
@akgrant thanks for the fix. By the way, what does mean +x in `${BUILD_NUMBER+x}`? I do not know this notation

demarey:
@estebanlm you were faster than me ;)

akgrant43:
@demary - re "+x", take a look at https://stackoverflow.com/questions/3601515/how-to-check-if-a-variable-is-set-in-bash.  It basically changes the check from "is empty" to "is set".  The two are typically regarded as the same, but are really different.  Most scripts don't care about the difference, but "set -o nounset" in the transform script causes an "is empty" test to fail if the variable isn't set.

Issue #2333, opened 2019-01-25
Menu in ToolDockingBar no longer works


StephanEggermont:
|w menuBar selectMenu|
w := StandardWindow new.
	selectMenu := w newMenu.	
	selectMenu addToggle: 'Close' target: w  selector: #close.
	menuBar := w newToolDockingBar.
	menuBar
		add: 'Select' translated
		subMenu: selectMenu.
w 	
		addMorph: menuBar
			fullFrame: ((0 @ 0 corner: 1 @ 0) asLayoutFrame bottomOffset: menuBar minExtent y+10).
w openInWorld

Used to work in Pharo 6 and early 7's. The menu doesn't show a triangle and doesn't open

StephanEggermont:
Worked in build 1095, broken in build 1096. Reopened https://pharo.fogbugz.com/f/cases/22084/Menubar-root-items-without-children-should-be-clickable

StephanEggermont:
See Issue #2331

Issue #2331, opened 2019-01-25
DockingBarMorph>>add: icon: help: subMenu: , does not consider submenu in the method implementation


VincentBlondeau:

VincentBlondeau:
I cannot commit the fix yet:

```
add: wordingString icon: aForm help: helpString subMenu: aMenuMorph 
	"Append the given submenu with the given label."
	
	self add: wordingString icon: aForm help: helpString subMenu: aMenuMorph action: nil keyText: nil
```

jecisc:
Seems fine! I can do the PR if you want.

MarcusDenker:
I close this now with the two PR merged

PR #2376, opened 2019-01-28
Ensure subMenu is opened (backport)

StephanEggermont:
Solves Issue #2331, #2333. Backport for Pharo 7. See 22084 in Fogbugz


estebanlm:
why this backport? why this is a critical fix?

StephanEggermont:
It breaks client code

estebanlm:
why? AFAIK, this never worked before (or it does not works since a lot of time).
and which client code?

StephanEggermont:
It worked in Pharo 6. I have code that needs to work in Pharo 7 so I need the fix backported

StephanEggermont:
unrelated failures

estebanlm:
now that I remember, there was a reason why this was removed. 
@jecisc you were working there, do you have an observation to do ?

jecisc (approved):
The PR seems fine for me. 

I introduce this problem while removing duplication. I forgot one parameter in this method.

jecisc:
Before each menu constructor was duplication the logic. I changed it to build in one method and have the other ones calling this one.

Issue #2383, opened 2019-01-28
Cleanup DoubleLinkedList packages


astares:
Pharo 8.0.0
Build information: Pharo-8.0.0+build.33.sha.a896d6da6b2617aa96aed6d3614402d518d8da4 (64 Bit)

in package Collections-DoubleLinkedList and Collections-DoubleLinkedList-Tests

- lint laments that the test package does violate a rule about using the result of an #add method
  but this is fine - so we should ban that rule for the test class
- so we also need a proper manifest
- class comments are missing
- tag the classes on test and core package equally as well as the manifest

astares:
Fix is in https://github.com/pharo-project/pharo/pull/2384

astares:
still on todo list

juliendelplanque:
Bump, I just used the DoubleLinkedList and this issue is super important.

juliendelplanque:
Can some admin reopen it? @MarcusDenker ?

juliendelplanque:
I see that PR #6022 addressed this issue, maybe it can be closed?

@astares can you confirm?

astares:
For the records: #6022 fixed the issue

PR #2384, opened 2019-01-28
2383 Cleanup DoubleLinkedList packages

astares:
Fixes issue https://github.com/pharo-project/pharo/issues/2383

### DIFF
--- a/src/Collections-DoubleLinkedList-Tests/DoubleLinkedListTests.class.st
+++ b/src/Collections-DoubleLinkedList-Tests/DoubleLinkedListTests.class.st
@@ -1,7 +1,10 @@
+"
+Unit tests for DoubleLinkedList
+"
 Class {
 	#name : #DoubleLinkedListTests,
 	#superclass : #TestCase,
-	#category : #'Collections-DoubleLinkedList-Tests'
+	#category : #'Collections-DoubleLinkedList-Tests-Base'
 }
 
 { #category : #testing }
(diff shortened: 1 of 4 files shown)


Ducasse:
Hi torsten

Why do you add a group? We should pay attention not to over clean. 
I have a lot of things that I think that we should improve and adding tag to test package is for me too much. 

Stef

astares:
@Ducasse 
No - it is not overclean. It is consistent as we have tags for others too. Look at packages with many classes and many tests - we tag the classes and their tests similar so they will be easily found. This should basically also be true for packages with not so many classes/tags. Because otherwise the question would be: where do we draw the line.  

Also as soon as we have a manifest we have it in tag "Manifest" - not tagging the others would leave them as "Uncategorized". 

Believe me: if you do things more consistent you as well as others (especially newbees) would feel right at home and more comfortable. That's why P7 already looks more consistent and clean compared to any other previous version.

Additionally: in the mid-term we should think also about merging LinkedList support into regular collection packages.  

Rembering the https://en.wikipedia.org/wiki/Broken_windows_theory

Ducasse:
Tx for this PR. Now we will not integrate it for the reason mentioned above.

guillep:
Ok, I've followed the discussion.
This PR, besides the new `Base` tag, 
 - adds a manifest
 - and blacklists a lint rule

Making other issues gone.
I think it's good enough to go in.

guillep:
hi @astares I'm ok with integrating this.
Could you fix the merge issues? I've tried on my side, but I cannot push the merge commit to your PR. So I'd have to do a new PR.

Tell me so I know if I do the new PR or not.

Thanks

guillep:
Duplicate of https://github.com/pharo-project/pharo/pull/6022. The PR has been redone

PR #2398, opened 2019-01-29
2397-Use-theme-instead-of-referencing-Smalltalk-in-FastTable

### LINKED ISSUE #2397: Use theme instead of referencing Smalltalk in FastTable

jecisc:
FastTable: Do not reference Smalltalk to acces theme in morphs.

Fixes  #2397


Ducasse:
Not related!

Issue #2405, opened 2019-01-29
NewValueHolder should use hook to get its value


guillep:
When announcing a value change

```
valueChanged
	
	self announcer announce: (ValueChanged newValue: value)
```

Using the value in the variable directly does not let subclasses override that behaviour.
This creates issues in the SelectionValueHolder and breaks all Selection events in Spec

Ducasse:
ahhhh nice

guillep:
PR: https://github.com/pharo-project/pharo/pull/2408

guillep:
Integrated

PR #2406, opened 2019-01-29
Fix issue #2405

guillep:
NewValueHolder should use the #value hook so subclasses can override


guillep:
I'll remake the PR from a Pharo8 image.

PR #2408, opened 2019-01-29
2405-NewValueHolder-should-use-hook-to-get-its-value

guillep:
Use #value accessor so subclasses can hook on it


jecisc (approved):

jecisc:
If the checks pass, it's good.

Ducasse:
Breaking tests unrelated.

PR #2417, opened 2019-01-30
add g prefix to keep code compatible with git describes

guillep:
Fix for issue #2379.

The code that parses the version info expects a git-describe like version info. We need to produce a version info with that form so we get the correct commit hash and iceberg finds the right commit later on.


guillep:
BTW, just notice that I could not test nor run the bootstrap because I'm on windows.

pavel-krivanek (approved):
looks good, let us see if the CI will complain

PR #2420, opened 2019-01-30
2419-Use-PharoShortcut-in-FastTable-instead-of-hardcoding-select-all-shortcut

### LINKED ISSUE #2419: Use PharoShortcut in FastTable instead of hardcoding select all shortcut

jecisc:
Use PharoShortcut instead of hardcoded shortcut for selectAll 

Fixes #2419


Ducasse:
Cyril would it be possible to introduce a shortcut instance variable that is initialized to 
PharoShortcuts current 
and avoid to directly use PharoShortcuts current ?

jecisc:
Included in #2451

PR #2377, opened 2019-01-31
Integrate Clap

cdlm:
I think BaselineOfIDE still has to be modified so that Clap actually gets included in new images.

### DIFF
--- a/src/BaselineOfClap/BaselineOfClap.class.st
+++ b/src/BaselineOfClap/BaselineOfClap.class.st
@@ -1,0 +1,21 @@
+Class {
+	#name : #BaselineOfClap,
+	#superclass : #BaselineOf,
+	#category : #BaselineOfClap
+}
+
+{ #category : #baseline }
+BaselineOfClap >> baseline: spec [
+	<baseline>
+	spec for: #common do: [ spec
+		package: 'Clap-Core';
+		package: 'Clap-CommandLine' with: [ spec requires: #('Clap-Core') ];
+		package: 'Clap-Tests' with: [ spec requires: #('Clap-Core' 'Clap-Examples') ];
+		package: 'Clap-Examples' with: [ spec requires: #('Clap-CommandLine') ];
+		package: 'Clap-Commands-Pharo' with: [ spec requires: #('Clap-CommandLine') ];
+
+		group: 'default' with: #(core development);
+		group: 'core' with: #('Clap-Core' 'Clap-CommandLine');
+		group: 'pharo' with: #('Clap-Commands-Pharo');
+		group: 'development' with: #('Clap-Tests' 'Clap-Examples' 'Clap-Commands-Pharo') ]
+]
--- a/src/BaselineOfClap/package.st
+++ b/src/BaselineOfClap/package.st
@@ -1,0 +1,1 @@
+Package { #name : #BaselineOfClap }
--- a/src/BaselineOfIDE/BaselineOfIDE.class.st
+++ b/src/BaselineOfIDE/BaselineOfIDE.class.st
@@ -134,6 +134,8 @@ BaselineOfIDE >> baseline: spec [
 			spec repository: repository
 		].
 
+		spec baseline: 'Clap' with: [ spec repository: repository ].
+		
 		spec package: 'ReleaseTests'. 
 		spec package: 'Refactoring-Critics'.
 		spec package: 'NautilusRefactoring'.
@@ -801,7 +803,8 @@ spec group: #default with: #(
 	'BasicTools'	"dependency"
 	'Athens'
 	'Flashback'
-
+	'Clap'
+	
 	'Mandatory-Group'
 
 	'Kernel-Tests-Group'
--- a/src/Clap-CommandLine/ClapCommandLineHandler.class.st
+++ b/src/Clap-CommandLine/ClapCommandLineHandler.class.st
@@ -1,0 +1,38 @@
+"
+A thin layer to hook Clap into the existing command-line system
+"
+Class {
+	#name : #ClapCommandLineHandler,
+	#superclass : #CommandLineHandler,
+	#category : #'Clap-CommandLine'
+}
+
+{ #category : #accessing }
+ClapCommandLineHandler class >> commandName [
+	^ 'clap'
+]
+
+{ #category : #activation }
+ClapCommandLineHandler >> activate [
+	[
+		ClapContext withPragmaCommands beObeyingExits
+			activateWith: self arguments
+			onExit: [ :exit | self handleExit: exit ]
+	]
+	forkAt: Processor userSchedulingPriority
+	named: 'Clap commandline handler process'
+]
+
+{ #category : #activation }
+ClapCommandLineHandler >> handleExit: exit [
+	Smalltalk isInteractive
+		ifFalse: [ ^ exit pass ].
+
+	exit isSuccess
+		ifFalse: [ ^ exit resignalAs: (Error new messageText: exit messageText) ]
+]
+
+{ #category : #activation }
+ClapCommandLineHandler >> noneMatched [
+	self exitFailure: 'Unknown command'
+]
--- a/src/Clap-CommandLine/ClapCommandLineHandler.class.st
+++ b/src/Clap-CommandLine/ClapCommandLineHandler.class.st
@@ -1,5 +1,5 @@
 "
-A thin layer to hook Clap into the existing command-line system
+I hook Clap into the existing command-line handlers system.
 "
 Class {
 	#name : #ClapCommandLineHandler,
--- a/src/Clap-CommandLine/package.st
+++ b/src/Clap-CommandLine/package.st
@@ -1,0 +1,1 @@
+Package { #name : #'Clap-CommandLine' }
--- a/src/Clap-Commands-Pharo/ClapCodeEvaluator.class.st
+++ b/src/Clap-Commands-Pharo/ClapCodeEvaluator.class.st
@@ -1,0 +1,68 @@
+Class {
+	#name : #ClapCodeEvaluator,
+	#superclass : #Object,
+	#instVars : [
+		'arguments'
+	],
+	#category : #'Clap-Commands-Pharo'
+}
+
+{ #category : #'command line' }
+ClapCodeEvaluator class >> evaluate [
+	<commandline>
+	^ (ClapCommand withName: 'evaluate')
+		aliases: #('eval');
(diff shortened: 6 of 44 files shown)


cdlm:
I think BaselineOfIDE still has to be modified so that Clap actually gets included in new images.

Ducasse:
Thanks Damien. I think that we should ask Pablo or Guille since Clap should probably be in the bootstrap.

Ducasse:
I added them as reviewers. I do not know the impact of this but we will see.

guillep:
@Ducasse For a very first iteration we cannot just add Clap to BaselineOfIDE?
Then we can iterate on it: I'd suggest @cdlm to try bootstrapping and checking BaselineOfPharoBootstrap, where the packages used to bootstrap are declared, and PharoBootstrapInitialization where the image is initialized.

svenvc:
I hope this won't interfere or break normal command line stuff, that would be a disaster.

cdlm:
@svenvc For the moment there's a normal `clap` command-line handler which works as a prefix for all Clap commands and leaves the existing stuff untouched. But I do hope we manage to deprecate it in some finite number of releases…

Ducasse:
@cdlm Damien did you follow what guille proposed? I really want to get clap integrated and create new command line handler. 
@svenvc we should really get  better way to define command line. And we need help. We will try to avoid impacting people but not at all cost.

svenvc:
Where can I read about CLAP ?
What problem does it try to solve ?

akgrant43:
> Where can I read about CLAP ?
> What problem does it try to solve ?

https://github.com/cdlm/clap-st

svenvc:
Thanks for the pointer.
Sure, clap looks nice. I am all for improvements.
But please implement all complex real world use cases before talking about replacing anything.

cdlm:
There's also a WIP booklet: https://github.com/SquareBracketAssociates/Booklet-ClapCommandLine

Ducasse:
Hi sven 
I really agree with you. I started to write some command lines with Clap and we need more users. 
The old API is not really good - I faced it in Pillar. So if you give a try let us know.

svenvc:
Let's integrate it with the prefix and then try to use it for real

cdlm:
Had to re-do the PR without `TClapBase` after removing it with `git rm`…

cdlm:
Bootstrap produces an image with Clap loaded but with a half-dozen `PrimitiveFailed: primitive #'bad argument' in File class failed` at boot time

akgrant43:
Hi Damien,

> Bootstrap produces an image with Clap loaded but with a half-dozen `PrimitiveFailed: primitive #'bad argument' in File class failed` at boot time

Can you clarify what you mean by 'at boot time'?  Does this happen if I download the image and try to use it, or was it during the bootstrap process (I couldn't find it in the log)?

Thanks,
Alistair

cdlm:
@akgrant43 I tried bootstrapping an image from my working copy of this PR, by running `bootstrap/scripts/bootstrap.sh`. The image builds, but when I launch it I get the errors mentioned.

I don't know where to download an image since CI fails for an unrelated reason.

akgrant43:
Hi Damien,

Thanks for the clarification.  I'll have a go at reproducing this, although it might take a couple of days.

Cheers,
Alistair

akgrant43:
Hi Damien,

>I don't know where to download an image since CI fails for an unrelated reason.

The image can be downloaded by clicking on the "Details" for the CI build, and then "Artifacts":

https://ci.inria.fr/pharo-ci-jenkins2/blue/organizations/jenkins/Test%20pending%20pull%20request%20and%20branch%20Pipeline/detail/PR-2377/3/artifacts

The first three artifacts are the sources and 32 and 64 bit zip files.

If I download the 64 bit image and use the latest VM it starts without any problems.  I can run the File related automated tests without anything like what you're describing.

Can you provide the output of pharo --version?


Cheers,
Alistair.

cdlm:
I saw later that it's only the windows build which fails, sorry.

Here's the VM that was used to do the bootstrap (I think). But then it's downloaded automatically by the bootstrap script, so it should be correct.
```
5.0 5.0.201806281256 Mac OS X built on Jun 28 2018 13:06:06 UTC Compiler: 4.2.1 Compatible Apple LLVM 7.3.0 (clang-703.0.31) [Production Spur VM]
CoInterpreter VMMaker.oscog-eem.2401 uuid: 29232e0e-c9e3-41d8-ae75-519db862e02c Jun 28 2018
StackToRegisterMappingCogit VMMaker.oscog-eem.2401 uuid: 29232e0e-c9e3-41d8-ae75-519db862e02c Jun 28 2018
VM: 201806281256 https://github.com/OpenSmalltalk/opensmalltalk-vm.git
Date: Thu Jun 28 14:56:30 2018 CommitHash: a8a1dc1
Plugins: 201806281256 https://github.com/OpenSmalltalk/opensmalltalk-vm.git
```

akgrant43:
Hi Damien,

That's a really old VM and doesn't have the required primitives for Pharo 8.  You should have one dated 5 Jan 2019 or later.

The CI build used:

Target VM: Date: Mon Feb 18 09:42:23 2019 CommitHash: f9ae4a14

Cheers,
Alistaiir

cdlm:
urgh 😑
when do we fix that vm/image versioning hassle (and how…) ?

akgrant43:
Not a solution, but at least makes the issue more explicit:

There's already a check in the image for the VM date.  I think I'll put forward a PR that updates the date.

At least the error will be meaningful (something like VM is too old).

cdlm:
Images could start with a `#!/usr/bin/env pharo-image-runner` shebang, which would proceed to spawning the matching VM and sources file…

akgrant43:
This could be used to ensure the VM is correct for the image type, i.e. 32 bit vs 64 bit, cog vs stack, etc., but it doesn't really handle versions / releases within those groups (at the moment).

akgrant43 (changes requested):
Hi Damien,

Running the hello example results in:

```
$ ./pharo Pharo.image clap hello --shout you
Renamed to #implicitMeaning:efaultMeaning: called from ClapCodeEvaluator class>>#evaluate has been deprecated.
Renamed to #implicitMeaning:efaultMeaning: called from ClapCommand class>>#forHelp has been deprecated.
HELLO, YOU!
```

Can we get rid of the deprecation warnings before merging?

Thanks,
Alistair

akgrant43 (approved):
Deprecation messages are gone.

I tried this with a couple of my applications.  Only simple command lines, but it worked well.  I'm looking forward to having this in the image. :-)

akgrant43:
Actually, I'll have to postpone the approval - there are a couple of automated test failures that look related:

- Unix32.Clap.Tests.Integration.ClapDocumentationTest
- MacOSX32.Clap.Tests.Integration.ClapDocumentationTest

akgrant43:
Hi Damien,

ClapDocumentationTest>>testHelloShortUsage hasn't been updated to reflect the --language option that was added.

Cheers,
Alistair

cdlm:
Sorry about that, should be fixed. I'm trying a bootstrap now… *edit* …and it seems to be working.

akgrant43:
Hi Damien,

It looks like this is triggering a number of test failures:

- ClapPharoVersion is causing #testHashMethodNeedsToBeInComparingProtocol to fail.
- ClapMatchesTest is causing #testNoUnusedTemporaryVariablesLeft to fail.
- Several of the test #setUp methods don't call super setUp.

Cheers,
Alistair

cdlm:
> * ClapPharoVersion is causing #testHashMethodNeedsToBeInComparingProtocol to fail.

That's because I use `#perform:`, which isn't really… satisfying… but easy to remedy by adding `#meaning:` blocks in the command spec.

> * ClapMatchesTest is causing #testNoUnusedTemporaryVariablesLeft to fail.
> * Several of the test #setUp methods don't call super setUp.

Easy to fix, just not sure I'll have enough focus time this weekend…

akgrant43:
> > * ClapPharoVersion is causing #testHashMethodNeedsToBeInComparingProtocol to fail.
> 
> That's because I use `#perform:`, which isn't really… satisfying… but easy to remedy by adding `#meaning:` blocks in the command spec.

This just needs ClapPharoVersion>>hash to be moved to the 'comparing' protocol (from 'formats').

Ducasse:
Thank you both. I really want to stress test more Clap because we should have a really clean way to create command line :). 

BTW I do not really these rules that raises issues instead of fixing them. We will have to do something about this because it slows us. 

@Alistair damien is travelling on the far west of france this week-end.

akgrant43:
> BTW I do not really these rules that raises issues instead of fixing them. We will have to do something about this because it slows us.

I think the real issue here is all the noise created by all the other unrelated test failures.  If anyone could run the test suite and be confident that any failure was due to changes they made they would look much more closely at them.  The multiple repeating failures mean that people tend to miss the real issues.

That actually happened here.  I guess that one or more of the #testHashMethodNeedsToBeInComparingProtocol, #testNoUnusedTemporaryVariablesLeft and #setUp test failures were present when I reported the deprecated message sends.  Which has meant that Damien has had to go back and do multiple fixes (sorry about that, Damien).  This probably wouldn't have happened if the unrelated test failures weren't there.

This obviously isn't an easy problem to solve - several people have put quite a bit of effort in and it still persists.

Cheers,
Alistair

cdlm:
I don't see what you report in the CI build logs from Jenkins, and when I push I get a mail from Travis with an unrelated issue (my account doesn't have credentials)… are you running those tests from the test runner by hand ?

akgrant43:
Hi Damien,

I don't have CI credentials either.  Clicking on Details on this page takes me to Jenkins, I then select "Tests" from the top menu bar (it has a 12 next to it indicating 12 failures).

On the Tests page in the 9 existing failures are the three tests that I mentioned above (the #setUp test is  #testAndMakeSureSuperSetupIsCalledAsFirstMessageInSetupMethodsOfTestCases).

I then ran them by hand to confirm that it is Clap related classes that are triggering the failures.

HTH,
Alistair

cdlm:
Thanks!

The test code isfixed. The `hash` redefinition requires a small redesign, working on that…

akgrant43 (approved):
- Deprecation messages and related test failures have all been resolved.
- Remaining test failures are pre-existing and unrelated.
- Examples work out-of-the-box


Ready to be merged...

cdlm:
🔔

Ducasse:
I was checking and this is good that this is in the baseline of ide!

PR #2431, opened 2019-02-01
2393 integrate file attributes plugin

### LINKED ISSUE #2393: #isSymlink broken and additional file attributes
- #isSymlink returns false for valid symlinks on Linux (and the code doesn't look right for MacOS).
- stat() on linux & MacOS returns additional attributes that aren't currently accessible from within the image.
- Provide access to symlink attributes (lstat())

PR: https://github.com/pharo-project/pharo/pull/2431

Related fogbugz issues:

- https://pharo.fogbugz.com/f/cases/18279/isSymlink-seems-to-be-broken-on-Linux
- https://pharo.fogbugz.com/f/cases/21368/Integrate-FileAttributesPlugin
- https://pharo.fogbugz.com/f/cases/22368/File-permissions-and-ownership

Mailing list discussions:

- http://forum.world.st/FileSystem-file-attributes-and-isSymlink-patch-td4956517.html
- http://lists.squeakfoundation.org/pipermail/vm-dev/2017-July/025605.html

The fix uses a new plugin in the VM, FileAttributesPlugin:

```smalltalk
MCHttpRepository
    location: 'http://smalltalkhub.com/mc/Alistair/FileAttributesPlugin/main'
    user: ''
    password: ''
```

More details of the changes:

1. #isSymlink now works properly on Linux (and it should also work on 
  MacOS and BSD).
2. The list of file attributes available from DiskDirectoryEntry now is:
    #accessTime (new)
    #changeTime (new)
    #deviceId (new)
    #gid (new)
    #inode (new)
    #isBlock (new)
    #isCharacter (new)
    #isExecutable (new)
    #isFIFO (new)
    #isRegular (new)
    #isSocket (new)
    #isSymlink (works)
    #numberOfHardLinks (new)
    #targetFile (new)
    #uid (new)

    #creationTime
    #exists
    #isDirectory
    #isFile
    #isReadable
    #isWritable
    #modificationTime
    #permissions
    #size
3. FileReference>>exists is faster than before (well, at least on my 
  linux laptop).  This is useful as it is called quite often.
4. It is possible to retrieve symbolic link attributes, e.g. all the 
  attributes listed above plus the target path.

As implied above, the changes are all backward compatible (except 
the broken #isSymlink), although a couple deserve mention:

1. Obviously #isSymlink now answers correctly (previously it would only 
  answer correctly for a broken link).
2. Requesting any of the attributes listed above (except #isSymlink) 
  will return the value of the target path.
3. The attributes of a symbolic link can be retrieved using 
  FileReference>>symlinkEntry.

Overall, performance is slightly better than before.  Code that accesses a
single attribute through FileReference will be faster than before (it used 
to retrieve all supported elements and throw the unused ones away).  Code
that needs to access multiple attributes and uses FileSystemDirectoryEntry
will access the new attributes without any additional performance cost.

I've been using this code continuously since May 2017 - initially on 32 bit linux and more recently on 64 bit linux.  It's had some testing on Windows and MacOS.

akgrant43:
Resolves #2393 

- #isSymlink returns false for valid symlinks on Linux (and the code doesn't look right for MacOS).
- stat() on linux & MacOS returns additional attributes that aren't currently accessible from within the image.
- Provide access to symlink attributes (lstat())


akgrant43:
The CI has 6 failed tests, however none appear related and all have appeared in other recent PR builds.

jecisc:
The Unix build failed so I'll relaunch. 

But I confirm the failing tests on Windows and OSX are unrelated.

akgrant43:
Thanks, Cyril.

Ducasse:
Thanks a lot Alistair. I read all the code and it is consequent. 
I'm super happy to see the file library getting improved.
Thanks a lot again.

Issue #2395, opened 2019-01-29
Non-ASCII class and author names break SourceFileArray>>#getPreambleFrom:at:


svenvc:
People are reporting that they get strange encoding errors when filing out code while using a non-ASCII class or author name.

Here is some initial analysis:

This is one manifestation of a problem that has been present for quite a while.

I'll start by describing what I did, what went well and where/how this fails, some generic points, and two conceptual solutions (that need further verification).

Like you, I created a new subclass:

````
Object subclass: #ClasseFrançaise
	instanceVariableNames: ''
	classVariableNames: ''
	package: '_UnpackagedPackage'
````

With comment:

````
I am ClasseFrançaise.

Try:

	ClasseFrançaise new élève.
	ClasseFrançaise new euro.
````

And two methods (in the 'test' protocol):

````
élève
	^ 'élève'

euro
	^ '€'
````

I added the euro sign (because that is encoded in UTF-8 with 3 bytes, not 2 like ç).
Like you said, the system can cope with such class and method names and seems to function fine.

Looking at the .changes file, the correct source code was appended:

````
----SNAPSHOT----2019-01-26T23:36:18.548555+01:00 work.image priorSource: 339848!

Object subclass: #ClasseFrançaise
       instanceVariableNames: ''
       classVariableNames: ''
       package: '_UnpackagedPackage'!
!ClasseFrançaise commentStamp: 'SvenVanCaekenberghe 1/27/2019 12:25' prior: 0!
I am ClasseFrançaise.!
!ClasseFrançaise methodsFor: 'test' stamp: 'SvenVanCaekenberghe 1/27/2019 12:26'!
élève
       ^ 'élève'! !
!ClasseFrançaise commentStamp: 'SvenVanCaekenberghe 1/27/2019 12:27' prior: 33898360!
I am ClasseFrançaise.

Try:

       ClasseFrançaise new élève.
       ClasseFrançaise new euro.
!
!ClasseFrançaise methodsFor: 'test' stamp: 'SvenVanCaekenberghe 1/27/2019 12:27'!
euro
	^ '€'! !
````

Doing a file out (or otherwise saving the source code) fails. The reason is an incorrect manipulation of this source file while looking for what is called the method preamble, in SourcFileArray>>#getPreambleFrom:at: position

An programmatic way to invoke the same error is by doing

````
(ClasseFrançaise>>#élève) timeStamp.
(ClasseFrançaise>>#élève) author.
````

Both fail with the same error.

[...]

PR #2433, opened 2019-02-01
2395-Non-ASCII-class-and-author-names-break-SourceFileArraygetPreambleFromat

svenvc:
Reimplement SourceFileArray>>#getPreambleFrom:at: to delegate to new SourceFile>>#getPreambleAt:
New implementation of SourceFile>>#getPreambleAt: to work on byte level and deal with non-ascii characters (classname or author) in the preamble itself

Fix for https://github.com/pharo-project/pharo/issues/2395


Ducasse:
The failed tests do not look related. I relaunched the build for osx

PR #2436, opened 2019-02-03
OpalCompiler-Core spelling in class comment

bencoman:
While reading source code on github I noticed these few spelling errors.  
I was *encouraged* to fix them since it was so very easy to do it in place by clicking the <Edit> icon.
While general policy is to require an Issue to go with each contribution, 
maybe this be stretched for documentation or at least simple spelling errors 
like these.  It could encourage others to do similar small fixes.

One thing to distinguish is that Github PRs provide a place to discuss contributions which the previous mcz process didn't.


MarcusDenker (approved):
ok

PR #2454, opened 2019-02-07
Fix Year>>firstThursday

### LINKED ISSUE #2453: Year>>firstThursday sometimes returns a day from the preceding year.
Example: `(Year year: 1988) firstThursday year` yields `1987`.

(I have a bug fix with test that I will submit soon.)

khinsen:
Ensure that it never returns a date in the… preceding year.
Test case provided.
Fixes #2453.

### DIFF
--- a/src/Kernel-Chronology-Extras/Year.class.st
+++ b/src/Kernel-Chronology-Extras/Year.class.st
@@ -83,9 +83,9 @@ Year >> daysInYear [
 
 { #category : #accessing }
 Year >> firstThursday [
-	"Return the years first Thursday.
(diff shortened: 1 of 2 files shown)


khinsen:
Note: my signed contributor agreement is in the mail.

svenvc:
Assuming nothing else broke, I agree with this change

Ducasse:
I love when fix are proposed with tests!!!! Excellent!
Tx

khinsen:
What's wrong with continuous integration? It says that one check failed, but when I click on "Details" it shows no problem at all.

Ducasse:
I can see the details and may be you do not have access. The failed tests are not related.

PR #2465, opened 2019-02-08
2464-Remove-GLMDashboard

### LINKED ISSUE #2464: Remove GLMDashboard
GLMDashboard is not documented and is not used in the image.

I propose to remove it. If users want it, they should use the bigger version of Glamour.

jecisc:
Remove GLMDashboard

GLMDashboard is not documented and is not used in the image.
I propose to remove it. If users want it, they should use the bigger version of Glamour.

Fixes #2464


Ducasse:
BrokenTests unrelated

Issue #2441, opened 2019-02-05
Less Smalltalk ui theme - Cleaning Glamour and GT (Part 1)


Ducasse:
AntiSingleton Campaign!
Choose your camp designer. Be pro or lame.

Ducasse:
May be we should introduce 
DisplayableObject which as a theme variable and object using a theme should really subclass from it because avoiding this Smalltalk ui theme plague.

cdlm:
Is there a critic rule for this? Other instances that could be flagged by such a rule?

Ducasse:
This is strange because I thought that I did a PR.

Ducasse:
I issued a new PR.

juliendelplanque:
@MarcusDenker @Ducasse this issue has been fixed by #2473  but was not closed.

PR #2473, opened 2019-02-08
2441-Less-Smalltalk-ui-theme-Part-1

Ducasse:
Cleaning in the hope of a better world but having serious doubt. Because we introduced this bad code!.
https://github.com/pharo-project/pharo/issues/2441


Ducasse:
Broken tests unrelated.

Ducasse:
less shit in the system!

PR #2487, opened 2019-02-11
2486-Class-builder-should-call-initializeSlot-on-instance-migration

### LINKED ISSUE #2486: Class builder should call initializeSlot on instance migration
When migrating instances, if we change the definition of a class adding a new slot that requires initialization, we should initialize that slot in the new instance before doing the value migration.

For example, if we add a valueHolder slot, we should initialize that slot with a value holder before migrating the value from the normal instance variable to the value holder inside the new instance variable.

guillep:
Fix #2486
- fix copyObject:to: to initialize slots on class migration
- add tests for slot initialization and migration


MarcusDenker (approved):
looks good

MarcusDenker:
failing tests not related

PR #2490, opened 2019-02-11
2489-DatePrintFormatTester-should-be-removed

### LINKED ISSUE #2489: DatePrintFormatTester should be removed
It has a lot of duplication of asserters and is used only for one test. 

We can just remove it and do the testing directly in the test. This removes a lot of useless code and make the test more direct.

jecisc:
Remove DatePrintFormatTester that contains a lot of duplicated code and is used in only one test.

Fixes #2489


Ducasse:
Broken tests are unrelated.

PR #2491, opened 2019-02-11
2488-ZipArchive-should-use-FileReferences-and-not-path-strings

### LINKED ISSUE #2488: ZipArchive should use FileReferences and not path strings
ZipArchive and co uses path strings evreywhere. It prevents us to use a memory file system for tests by example. FileReference should be use instead

demarey:
fixes #2488

### DIFF
--- a/src/Compression-Tests/ZipArchiveTest.class.st
+++ b/src/Compression-Tests/ZipArchiveTest.class.st
@@ -44,7 +44,7 @@ ZipArchiveTest >> tearDown [
 { #category : #tests }
 ZipArchiveTest >> testAddNonExistentFile [
 	
-	self should: [ zip addFile: 'it_would_be_crazy_if_this_file_existed.ext' ] raise: FileDoesNotExistException.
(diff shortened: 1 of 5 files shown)


svenvc:
There is still 2 place where you can send #pathString directly to a FileReference instead of going via #path ...

Ducasse:
Christophe can you address the point raised by sven and we integrate your changes.

Ducasse:
Christophe can you update your PR so that we integrate it?
Tx

demarey:
@svenvc I do so because pathString does not have the same behaviour on FileReference and on Path. On a FileReference, pathString first resolves the file reference, so it is impossible to get relative pathes.
```smalltalk
'a' asFileReference path pathString "'a'"
'a' asFileReference pathString "'/Users/demarey/Documents/Pharo/images/Pharo 8.0 - 64bit - ziparchive PR/a'"
```

svenvc:
But they are the same file, no ?

Anyway, if you are sure and say that it has to be like that, then OK for me.

demarey:
I cannot change to use absolute pathes because the places where I use `fileRef path pathString` are places where we specify the local name of the file in the archive (not the file to compress). It has to be a relative path.

PR #2497, opened 2019-02-12
More reasonable timespan arithmetic

### LINKED ISSUE #2458: Timespan subclasses induce unexpected behavior
Before reading on, consider the expression `(Year year: 2019) intersection: (Week year: 2019 week: 1)`. What would you expect its value to be?

My expectation was: the part of the first week of 2019 that's actually in 2019 (given that the first day is in 2018, a consequence of the ISO definition of a week). But its actual value is... `Year year: 2019`!

This is a consequence of `Timespan>>intersection:` returning an object of `self class`, and `Year` forcing all its instances to be one year from January to December. A meaningful result of my expression would have to be of class `Timespan`, not `Year`.

Unfortunately, `Timespan` has two kinds of subclasses: `Year`, `Month` etc. are special *values* of timespans, whereas `Schedule` is a timespan with additional semantics. This doesn't look like a particularly clean object model to me. You'd probably want the intersection of a schedule with a timespan to be a schedule, but you don't want the intersection of a year with anything else to be a year.

There are other operations that suffer from similar problems. For example. `(Year year: 2019) + 1 day` is again `Year year: 2019`, whereas any other timespan would be shifted by one day. But perhaps the clearest indicator of the problem is the class method `starting:duration:` that all these classes override in order to explicitly ignore its second argument and modify the first one.

khinsen:
- Subclasses of fixed duration (Year, Month, Week, Date) behave like plain Timespans under arithmetic (+ - intersection:) (see `testSubclassArithmetic` for illustration)
- `Timespan >> intersection:` does not return results shortened by `clockPrecision` (see `testIntersectionWithSelf` to understand why this is more reasonable)

Fixes #2458

### DIFF
--- a/src/Kernel-Chronology-Extras/Month.class.st
+++ b/src/Kernel-Chronology-Extras/Month.class.st
@@ -199,3 +199,9 @@ Month >> printOn: aStream [
 
 	aStream nextPutAll: self monthName, ' ', self year printString.
 ]
+
+{ #category : #private }
+Month >> species [
+
+	^Timespan
+]
--- a/src/Kernel-Chronology-Extras/Timespan.extension.st
+++ b/src/Kernel-Chronology-Extras/Timespan.extension.st
@@ -183,10 +183,11 @@ Timespan >> intersection: aTimespan [
 
 	 | aBegin anEnd |
 	 aBegin := self start max: aTimespan start.
-	 anEnd := self end min: aTimespan end.
-	 anEnd < aBegin ifTrue: [^nil].
+	"Use start + duration rather than end, because the latter subtracts clockPrecision."
+	 anEnd := (self start + self duration) min: (aTimespan start + aTimespan duration).
+	 anEnd <= aBegin ifTrue: [^nil].
 
-	 ^ self class starting: aBegin ending: anEnd.
+	 ^ self species starting: aBegin ending: anEnd.
 
 ]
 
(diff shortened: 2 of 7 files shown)


Ducasse:
Hi konrad

Thanks for looking into this. Last time I was working on a simple bill generator I found many places where the logic was not good in the date and time. I added some methods more but it felt always clunky. I wanted to have period from 1 march to 31 march and it was strange to get it. 
I was thinking at some point that I would like to be able to have other time oriented libraries for more
advanced time modeling. I do not know if you know Chronos (not maintained) and Aconcagua. 
Stef

Ducasse:
Change looks (from far) nice. I will let other have a look.

khinsen:
Hi Stef, thanks for looking into this! @svenc mentioned Chronos, of which I even found a downloadable version, but being both large and abandoned doesn't really encourage me to look at it in any more detail. Aconcagua looks interesting for other applications (in physics/chemistry I have to deal with plenty of units), but it doesn't seem to handle dates, only durations.

As I said in #2458, I believe that any modern programming system should have support for dates and times at the ISO 6801 level. Anything more (as in Chronos) can get a chore to maintain, but anything less is disappointing. For now I remain optimistic that the Pharo date/time classes can be fixed to get there.

astares:
If we ever have the time to rework this then we should support also other calendar systems
(beside Gregorian) for instance Hijri calendar:

https://www.joda.org/joda-time/cal_islamic.html

Ducasse:
I would love to get some new time libraries with super nice tests. Anyway let us integrate this. 
Can you point me to the chronos code? Does it contain tests? I would like to try to publish it on github for archival.

Ducasse:
If you find other glitches please send PRs. Tx

khinsen:
The latest Chronos source code I could find is [release B1.196](http://map.squeak.org/package/46b5731f-5112-4157-94d3-d40c4ed8304d/autoversion/8) for Squeak.

PR #2505, opened 2019-02-12
[DoNotIntegrate]2504-FTTable-does-not-hide-headers-dynamically

### LINKED ISSUE #2504: FTTable does not hide headers dynamically
In an already open fast table, if it was showing a header and we set it to hide it, the header is still shown.

guillep:
Fix #2504
 - add also headerRow to be able to test it from Spec adapter tests


guillep:
PR redone in https://github.com/pharo-project/pharo/pull/2506

PR #2506, opened 2019-02-12
2504-FTTable-does-not-hide-headers-dynamically

### LINKED ISSUE #2504: FTTable does not hide headers dynamically

guillep:
Fix #2504
 - reset headerRow variable in case no column headers should be shown
 - add headerRow getter for testing in spec


Ducasse:
Broken tests are unrelated and changes look ok. I integrate.

PR #2508, opened 2019-02-13
Fix `Week>>year:week:` to return a correct ISO week (starting on Monday)

### LINKED ISSUE #2455: Week>>year:week: doesn't do what its documentation says
The comment in the method says:

    Return the ISO week for the given week number and the given year.
    Week 1 contains the year's first Thursday (weekday = 5)

However, ISO8601 specifies that weeks start on a Monday, whereas Pharo generally adopts the convention that weeks start on Sunday.

I can see two correction strategies:

 1) Change the documentation of `Week>>year:week:`. Add a new method `Week>>year:isoWeek:` that returns a Week starting on Monday.
 2) Change the behavior of `Week>>year:week:` to reflect the documentation.

This comes down to deciding how important backwards compatibility is. Nowadays week numbers always refer to the ISO6801 definition (not sure that was always the case), so the current behavior of `Year>>year:week:` is not very useful, but nevertheless existing code may rely on it.

I'd be happy to provide a patch for either approach.

khinsen:
Test included.
Fixes #2455

### DIFF
--- a/src/Kernel-Chronology-Extras/Week.class.st
+++ b/src/Kernel-Chronology-Extras/Week.class.st
@@ -69,9 +69,13 @@ Week class >> week: weekNumber [
 
 { #category : #'instance creation' }
 Week class >> year: aYear week: weekNumber [
-	"Return the ISO week for the given week number and the given year.
-	Week 1 contains the year's first Thursday (weekday = 5)"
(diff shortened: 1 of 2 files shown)


Ducasse:
Fix and tests look good. Broken tests are unrelated

Issue #2511, opened 2019-02-13
Tag in Slot-Examples should be uppercase


astares:
There is a class tag called "base" in package "Slot-Examples". It should be uppercase "Base"

astares:
fixed in https://github.com/pharo-project/pharo/pull/2512

MarcusDenker:
merged

PR #2512, opened 2019-02-13
2511 Tag in Slot-Examples should be uppercase

astares:
solving https://github.com/pharo-project/pharo/issues/2511
- use "Base" instead of "base" for class tag


MarcusDenker (approved):
ok

MarcusDenker:
error not rested

Issue #2513, opened 2019-02-13
Tag in MenuRegistration package should be uppercase


astares:
The tag "example" in package MenuRegistration should be uppercase "Example"

astares:
Fix in https://github.com/pharo-project/pharo/pull/2514

MarcusDenker:
merged

PR #2514, opened 2019-02-13
2513 Tag in MenuRegistration package should be uppercase

astares:
https://github.com/pharo-project/pharo/issues/2513


MarcusDenker (approved):
ok

PR #2520, opened 2019-02-13
2205-OCASTTranslator-class-initialize-should-not-invoke-super

MarcusDenker:
OCASTTranslator class initialize should not invoke super

fixes 2205


MarcusDenker:
error not related

PR #2521, opened 2019-02-13
2189-superNodeClass-selfNodeClass-and-thisContextNodeClass-are-not-used-

### LINKED ISSUE #2189: superNodeClass, selfNodeClass and thisContextNodeClass are not used.
This is probably normal since the creation logic is in the RBVariable but may be we should remove such methods.

MarcusDenker:
removes the methods. fixes #2189


MarcusDenker:
error not related

PR #2522, opened 2019-02-13
2518-DiskStoreentryFromNodepathfor-doesnt-use-existing-attributes

### LINKED ISSUE #2518: DiskStore>>entryFromNode:path:for: doesn't use existing attributes
In Pharo 8 DiskStore>>entryFromNode:path:for: doesn't pass on the attribute information that it has, causing the information to be unnecessarily retrieved a second time when the entry is used.

An example of this is sending FileReference>>entries to a directory.

See also http://forum.world.st/File-problems-in-Pharo-8-tp5095111p5095176.html

PR #2522

akgrant43:
DiskStore>>entryFromNode:path:for: doesn't pass on the attribute information that it has, causing the information to be unnecessarily retrieved a second time when the entry is used.

An example of this is sending FileReference>>entries to a directory.

See also http://forum.world.st/File-problems-in-Pharo-8-tp5095111p5095176.html

- Initialise DiskDirectoryEntry with the stat attributes
- Add automated test to ensure the attributes are loaded

Fixes #2518

### DIFF
--- a/src/FileSystem-Core/DiskDirectoryEntry.class.st
+++ b/src/FileSystem-Core/DiskDirectoryEntry.class.st
@@ -1,3 +1,28 @@
+"
+DiskDirectoryEntry holds attributes for a single file / directory on the disk.
+
+The actual attributes are held in two arrays which are returned directly from the #fileAttributes:mask: primitive.  See statAttributes and accessAttributes, below.
+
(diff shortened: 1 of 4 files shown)


akgrant43:
Failing CI tests are all existing and unrelated.  I think this is OK to merge.

Ducasse:
Tx Alistair. I confirm... unrelated and boring red tests :)

akgrant43:
Thanks, Stef!

Issue #2525, opened 2019-02-13
DiskStore>>delete: UTF8 encodes twice


akgrant43:
When deleting files (as opposed to directories) DiskStore>>delete: UTF8 encodes the file name twice, wasting time and memory.

PR to follow...

akgrant43:
PR: #2526

PR #2526, opened 2019-02-13
2525-DiskStoredelete-UTF8-encodes-twice

akgrant43:
When deleting files (as opposed to directories) DiskStore>>delete: UTF8 encodes the file name twice, wasting time and memory.

Fixes: https://github.com/pharo-project/pharo/issues/2525

### DIFF
--- a/src/FileSystem-Disk/DiskStore.class.st
+++ b/src/FileSystem-Disk/DiskStore.class.st
@@ -268,17 +268,17 @@ DiskStore >> defaultWorkingDirectory [
 { #category : #public }
 DiskStore >> delete: path [
 	| pathString encodedPathString |
-	
+
 	((self exists: path) or: [ self isSymlink: path ])
 		ifFalse: [ ^ FileDoesNotExistException signalWith: path ].
-		
+
 	pathString := self stringFromPath: path.
-	encodedPathString := File encodePathString: pathString.
-	
-	(self isDirectory: path)
-		ifTrue: [ File deleteDirectory: encodedPathString ]
-		ifFalse: [ 
-			(File named: pathString) delete ]
+
+	(self isDirectory: path) ifTrue: 
+		[ encodedPathString := File encodePathString: pathString.
+		File deleteDirectory: encodedPathString ]
+	ifFalse: 
+		[ File deleteFile: pathString ]
 ]
 
 { #category : #accessing }
--- a/src/FileSystem-Disk/DiskStore.class.st
+++ b/src/FileSystem-Disk/DiskStore.class.st
@@ -267,7 +267,7 @@ DiskStore >> defaultWorkingDirectory [
 
 { #category : #public }
 DiskStore >> delete: path [
-	| pathString encodedPathString |
+	| pathString |
 
 	((self exists: path) or: [ self isSymlink: path ])
 		ifFalse: [ ^ FileDoesNotExistException signalWith: path ].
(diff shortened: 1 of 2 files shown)


akgrant43:
It turns out there were a few inefficiencies:

- A File object was created purely to call the delete primitive.
- The path string was UTF8 encoded twice.
- The #exists method that was called opened and closed the file to check if the file exists.

Improving all of the above and testing on my 6 year old Dell XPS13 by creating 10000 empty files in a RAM disk reduced the time to delete the files from around 0.68 seconds down to 0.14 seconds.  See the script below.

```smalltalk
| root dir fileCount files times |

root := '/dev/shm' asFileReference.
dir := root / 'Issue2525'.
dir ensureCreateDirectory.
fileCount := 10000.
files := Array new: fileCount.

times := Array streamContents: [ :stream |
	10 timesRepeat: [ 
		Smalltalk garbageCollect.
		"Create the files"
		1 to: fileCount do: [ :i | | file |
			file := dir / ('f', i asString).
			file ensureCreateFile.
			files at: i put: file ].
		"Delete the files"
		stream nextPut: ([ 1 to: fileCount do: [ :i | (files at: i) delete ] ] timeToRun) ] ].

times
```

akgrant43:
The CI failed before finishing, however:

- the failure, "The class #NuclearSiteTestingClassTestis removed twice", has appeared in multiple previous runs.
- 36593 tests passed
- the 9 failures have all appeared previously.

So I think this is OK to merge.

Ducasse:
Hi alistair

Thanks for your time :). I have the impression that we do not need the variable encodedPathString in delete:

Ducasse:
Else for me just browsing the code, the changes are good. 
On which OS are you? Unix? Because Pablo has to improve the speed of file operation on windows because we are really slow. So any changes improving file operations are good and appreciated.

akgrant43:
> Hi alistair
> 
> Thanks for your time :). I have the impression that we do not need the variable encodedPathString in delete:

You're correct, the variable isn't really required.  I'll refactor and update the PR.

akgrant43:
> On which OS are you? Unix? Because Pablo has to improve the speed of file operation on windows because we are really slow. So any changes improving file operations are good and appreciated.

Linux (Ubuntu 16.04).

This should improve windows as well, although I haven't tested it.  Actually, it may even be a bigger percentage improvement since Windows seems to be slower for opening and closing files, which this used to do (as part of the #exists test).

But don't expect miracles.  While the percentage improvement is quite good, this is a small part of overall operations, so the absolute reduction in time isn't that large.

Ducasse:
Tx! :). I imagine that we will have to bench and study the wonderful windows API and see what we can do. Tonel is already improving the situation for code but we should make progress on this front.

Ducasse:
Broken tests are unrelated. Now I do not get why two builds are failing. Annoying infrastructure.

akgrant43:
I'm planning to try and reproduce the VM crash next, although it doesn't
seem consistent, unfortunately.

Ducasse:
I will let esteban decide because I'm away on vacation so if I merge and break something they will be in the mess. Now if you think that the changes are safe we can got with them. Alpha is alpha.

akgrant43:
Stef, thanks for restarting the CI.

This run was more successful - no VM crash, 36598 passed, only 4 failures, all of which have appeared previously.

So it is probably OK to merge now.

akgrant43:
Thanks, Stef.

PR #2528, opened 2019-02-14
Add trait TSingleton and modify existing singleton classes to use it

khinsen:
Following a comment by @MarcusDenker on the pharo-users mailing list, I packaged the singleton trait discussed there with modifications to existing singleton classes for inclusion in Pharo 8. I believe this is useful both as a kind of code cleanup operation and to provide a simple example for the use of traits.

I modified only singleton classes that followed the standard pattern. The remaining ones have different behavior in the `uniqueInstance`, `new`, or `reset` methods. A particularly frequent variant stores the unique instance in a class variable rather than an instance variable of the class. I have no idea why this choice was made, so I preferred not to touch them.

### DIFF
--- a/src/Glamour-Morphic-Brick/GLMBrickHeightTraverser.class.st
+++ b/src/Glamour-Morphic-Brick/GLMBrickHeightTraverser.class.st
@@ -1,21 +1,15 @@
 Class {
 	#name : #GLMBrickHeightTraverser,
 	#superclass : #GLMBrickLayoutTraverser,
-	#classInstVars : [
-		'uniqueInstance'
-	],
+	#traits : 'TSingleton',
+	#classTraits : 'TSingleton classTrait',
 	#category : #'Glamour-Morphic-Brick-Layouts-Utils'
 }
 
 { #category : #cleanup }
 GLMBrickHeightTraverser class >> cleanUp [
+	self reset
 
-	uniqueInstance := nil
-]
-
-{ #category : #accessing }
-GLMBrickHeightTraverser class >> uniqueInstance [
-	^ uniqueInstance ifNil: [ uniqueInstance := self new ]
 ]
 
 { #category : #private }
--- a/src/Glamour-Morphic-Brick/GLMBrickLayouter.class.st
+++ b/src/Glamour-Morphic-Brick/GLMBrickLayouter.class.st
@@ -31,9 +31,8 @@ rootWidthNode
 Class {
 	#name : #GLMBrickLayouter,
 	#superclass : #Object,
-	#classInstVars : [
-		'uniqueInstance'
-	],
+	#traits : 'TSingleton',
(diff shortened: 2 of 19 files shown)


svenvc:
I am not necessarily against this but I understood that the goal was to make the kernel of pharo free of traits use for reasons related to the bootstrap process. I might be wrong.

Ducasse:
Sven is right. We should really be careful about this aspect. The system should be layered and we removed Traits from the core in P7. 
In addition I would prefer to have such behavior encapsulated into a first class variables and we should do the same (reifying the behavior of certain class variables such as LRUCache) to many other caches.
So I would close this PR and think carefully about the impact. 
Now we can introduce TSingleton (even if I hate Singleton).

noha:
I'm cautious here, too. But I think if it should be layered there is need for a description of the layers as a fundation for such decision. If the kernel needs to be traits free then TSingleton cannot be used in any class below this layer. But it is still useful for everything above. As long as the layers are not defined this is hard to do. And yes I avoid singletons where I can, too, but that is not the point

astares:
It might be useful for unifying on the layers "above" if you check the existing but different implementions of #uniqueInstance in a regular P7/P8 image.

BTW:
The #uniqueInstance should do a check on #isAbstract so the trait can be applied in a common abstract superclass.

Ducasse:
Hi norbert

the guys from the bootstrap have tests to check dependencies. Now they are busy. So I would split the issue in 
- introducing the Trait definition
- then using it in GLM and related
- Re*

I checked a bit GLMDirection and there is no need of a singleton, just some class vars would have worked.

noha:
I like this plan. But I though GLM will be removed from the image so it is mabe not necesary to change them. Or do I understand that wrong?

PharoProject:
As a first attitude will be deprecated, so I figure is ok to apply it to it.

By the way I would like a good comment in TSingleton explaining that in general it is an anti-pattern and it needs to be used carefully.

estebanlm:
“Attitude”?
That happens when you answer mails with you phone :P

Ducasse:
So do I close it and we get another one? I guess so but let me know.

khinsen:
I'd be happy to submit a new version, just for the GLM stuff, but not before next week.

khinsen:
I have prepared a modified version of this patch in a [new branch](https://github.com/khinsen/pharo/commits/singleton-trait-2). However, when I check out a  fresh Pharo image and load this branch, I get lots of error messages whose cause isn't clear to me, and I have no time to dig into this right now.

If someone more knowledgeable wants to take over, please feel free to start from this new branch.

MarcusDenker:
I added it to the things to look at at the next sprint.

Ducasse:
I close this PR then

MarcusDenker:
I made a PR with just the Trait: https://github.com/pharo-project/pharo/issues/3203

Issue #2531, opened 2019-02-14
[Removing global action] Introduce DisplayableObject 


Ducasse:
To remove many of the Smalltalk ui icons .... and we should introduce
- DisplayableObject for objects that are not Model (this bug entry)
- Model should be a subclass of DisplayableObject

Ducasse:
For example Paragraph is using theme but it is not a model and does not need an icon. 
So it could be a DisplayableObject, same for Borderedxxx

Ducasse:
BorderStyle should be a subclass of DisplayableObject

astares:
DisplayableObject is an unusual name.

I would suggest to name it "DisplayObject" as you will find in most other frameworks. (VisualSmalltalk, Adobe, ...)

pavel-krivanek:
We already have the DisplayObject with a different meaning. DisplayableObject looks a little bit confusing to me. I would propose UIObject, UIAwareObject or something like that.

Ducasse:
OK we will discuss it and we will see. UIObject or UIAwareObject is not nice.

PR #2533, opened 2019-02-14
2531-Removing-global-action-Introduce-DisplayableObject-

Ducasse:
Introducing DisplayableObject as a first step in the battle against the super super ugly: Smalltalk ui evil.
#2531
https://github.com/pharo-project/pharo/issues/2531

### DIFF
--- a/src/System-Model/DisplayableObject.class.st
+++ b/src/System-Model/DisplayableObject.class.st
@@ -1,0 +1,23 @@
+"
+I'm a simple object superclass for objects that need to access information to be better displayed. 
+In particular I offer to access to a ui theme and icon. 
+
+For now I do not hold state but in the future it may make sense that I get an instance variable for the default theme and the icon manager. 
+"
+Class {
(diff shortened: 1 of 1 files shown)


Ducasse:
Broken tests are unrelated

astares:
DisplayableObject is an unusual name.

I would suggest to name it "DisplayObject" as you will find in most other frameworks. (VisualSmalltalk, Adobe, ...)

Ducasse:
well displayObject is not that nice. 
We will fix the name later. For now we need to clean all the mess around theme.

astares:
"Nice" is subjective. An objectively look would reveal:

DisplayObject is more common (405000 Google hits)
DisplayableObject is less common (12300 Google hits)

However ... you seem to flog your own pull requests into the image ignoring any comments.

It would be really preferable if you would behave like a regular contributor where PR's 
are always merged by another person than the one who contributed.

PR #2537, opened 2019-02-14
2536-Add-table-sorting-on-column-in-FastTable

### LINKED ISSUE #2536: Add table sorting on column in FastTable
I'll soon do a PR to introduce sorting while clicking on a column header in FastTable. 

It will have examples, documentation (class comments) and tests.

jecisc:
Add FastTable sorting. See examples.

Fixes #2536


Ducasse:
Cyril you have two related tests breaking
windows-32 / Tests-windows-32 / testSortDataSource – Windows32.Morphic.Widgets.FastTable.Tests.Sorting.FTDefaultColumnSortingStrategyTest
<1s
osx-32 / Tests-osx-32 / testSortDataSource – MacOSX32.Morphic.Widgets.FastTable.Tests.Sorting.FTDefaultColumnSortingStrategyTest

dionisiydk:
Hi @jecisc. Thanks for this effort.

I have few proposals to change your design:

**1. I think SortFunction should be used in first place instead of hiding it under strategy>>dataSource communication. It will simplify a logic and allow more features.**

Currently SortFunction is created inside this method:
```Smalltalk
FTDescendingSortingState >> executeSortingOn: aDataSource [
	aDataSource sortWith: [ :entity | self context property value: entity ] descending
```
And users specify properties for column sorting:
```Smalltalk
(FTColumn id: 'Class name')
	sortableOnProperty: #printString;
	yourself
```
where:
```
FTColumn >> sortableOnProperty: aProperty [
	self sortingStrategy: (FTDefaultColumnSortingStrategy column: self property: aProperty)
```
We can directly use instance of SortFunction here and have all nice composition algebra available for sorting. So for users it will be:
```Smalltalk
(FTColumn id: 'Class name')
	sortableUsing: #printString ascending;
	yourself
```
And  sorting state will be simple:
```Smalltalk
FTDescendingSortingState >> executeSortingOn: aDataSource [
	aDataSource sortWith: self context sortFunction
```
As example of sort composition imagine list of methods. We can sort it by selector and then by class:
```Smalltalk
(FTColumn id: 'Class name')
	sortableUsing: #sorted: #selector ascending, (#origin collatedBy: #name);
	yourself
```
To play with sort functions in live use following code:
```Smalltalk
Point methods, Number methods sorted: #selector ascending, (#origin collatedBy: #name). 
Point methods, Number methods sorted: #selector ascending, (#origin collatedBy: #name) reversed.
```
**2. Having SortFunction in mind do we really need strategy and state?**

Column can be configured directly with an instance of sort function. We will only need introduce  NullSortFunction to represent no sorting column state.
Sort function itself can tell us about #isSortingEnable or #sortingIcon.

So we can have simpler code like:
```Smalltalk
FTColumn >> sortDataSource: aDataSource [
 	aDataSource sortElements: sortFunction.
        sortFunction := sortFunction reversed.
]
```
Notice that simple #reversed operation will toggle sort direction of column. 
I would also imagine more advanced logic to take into account previous sorting of table:
```Smalltalk
FTColumn >> sortDataSource: aDataSource [
         sortFunction := sortFunction, self table lastSortFunction.
 	aDataSource sortElements: sortFunction.
        sortFunction := sortFunction reversed.
]
```
**3. It would be better to return new data source instance instead of ask data source to sort elements.**

First it will follow implementation of filtering. So it can be something like that
```Smalltalk
FTColumn >> sortTable: aTableMorph [
        sortFunction := sortFunction, aTableMorph lastSortFunction.
 	aTableMorph sortRowsBy: sortFunction.
        sortFunction := sortFunction reversed.
]
FTTableMorph>>sortRowsBy: sortFunction
    lastSortFunction := sortFunction.
    self dataSource: (self dataSource newDataSourceSortedBy: sortFunction).
````
I think this approach will simplify implementation of data sources.  
The idea with #sortElements: forces to maintain some state inside data sources. In FTSimpleDataSource example your were forced to introduce unsortedElements. So every implementation is forced to do similar. But this management can be extracted to table itself (there is already initialDataSource which can be reused for reseting).
So FTSimpleDataSource it would be:
```Smalltalk
newDataSourceSortedBy: aSortFunction
	| newElements |
	newElements := elements sorted: aSortFunction
	^ self class new elements: newElements
```
And this design will suite very well with Calypso data sources. It will not require changes in their implementation. In Calypso data sources are part of browser history and "mutable" data source instances will complicates this logic a lot.

Ducasse:
Thanks denis for this great feedback!!!!

jecisc:
Hi,

The idea of strategies was to let the user create new kind of strategies that can be customized for their need and also for some custom things they can have for advanced model/cells.

I did not know we could call #reverse on a SortFunction. So I agree it might be better to let the user give a sort function. Also with the strategy, we can also support both and let the user choose how he wants to do it. 

I'll take a look at your idea on point 3.

Thanks for the feedback.

jecisc:
For the part 3, I do not see how you can unsort the table with this implementation.

dionisiydk:
I do not understand why you need these two kind of strategy: property based and function based?.
SortFunctionSortingStrategy covers all these cases.

But more general question. Can you explain why strategy and state are needed? What they can do which sortFunction itself can't? Would be nice to have example because I see that this functionality can be built using plain sort functions.

dionisiydk:
For part 3. Can you explain what unsort command are you trying implement? To be able reset sorting at all?

dionisiydk:
Back to my question about strategy/state vs sortFunction. I think I understand why single sort function is not enough. If we would modify it like in my example then there will be no way to reset it to original value. We need to restore current column setting when user clicks on another column. Right?
Without these extra objects we would need to ask a data source if it is sorted or not. Probably it is  better to avoid it because it would complicate data source implementation.

But your solution with two hierarchies for state and strategy looks too complex for me. It can be done with two variables: initialSortFunction and activeSortFunction. They can be defined directly in column. Maybe extracting them into strategy is a good idea. But such strategy will not assume any subclassing.

jecisc:
The reason I used strategy is that in some cases the user might not want to use sort functions.

Some examples
- If I have a huge huge collection taking a lot of time to sort, maybe I already have a cache with the sorted elements somewhere. In that case, I might want a new kind of sorting strategy just setting my cache as the elements of the data source.
- If I display the content of a Postgres database for example, maybe for performance reasons I want to get the sorting done by Postgres and not use a Pharo sort function. 
- If I display some data that are not textual and do not implement comparison operators, maybe I can have a different way to sort them. In that case, I can create a new strategy for that. 

For the reset, in general, most sortable table implementation I saw had this behavior by default:
- One click, sort ascending
- Second click, sort descending
- Third click, unsort and go back to original state

jecisc:
Hum... Build is failing because I used a stateful Trait but this package is loaded before stateful traits?

I'll look at it monday.

dionisiydk:
> Some examples
> 
> * If I have a huge huge collection taking a lot of time to sort, maybe I already have a cache with the sorted elements somewhere. In that case, I might want a new kind of sorting strategy just setting my cache as the elements of the data source.
> * If I display the content of a Postgres database for example, maybe for performance reasons I want to get the sorting done by Postgres and not use a Pharo sort function.

Ok. But then DataSource>>sortElements: has incorrect assumption about sortFunction. 

From the other side sortFunction is quite good object to be used in those examples. It will not perform actual sorting but it will have required information about sorting operation (direction and property). 
In case of DB the data source can ask sort function for sorting expression.  
In case of cache sort function can be used as a key for such cache.

> * If I display some data that are not textual and do not implement comparison operators, maybe I can have a different way to sort them. In that case, I can create a new strategy for that.

This is perfect case for sort function. User will implement special subclass which will be responsible for this logic.

dionisiydk:
> Hum... Build is failing because I used a stateful Trait but this package is loaded before stateful traits?

Is it really make sense to use trait here? This trait can be just a superclass for those strategies.
Why you want to keep this property based strategy? It is sub case of sort function.

dionisiydk:
> For the reset, in general, most sortable table implementation I saw had this behavior by default:
> 
> * One click, sort ascending
> * Second click, sort descending
> * Third click, unsort and go back to original state

Interesting. I never saw it. Finder in MacOS and Explorer in Windows don't have "third click" behaviour.
I think it should be optional. 

But it would be better to extract this feature from data source (into table maybe). Because otherwise every data source would be required to implement it.  
Also it requires extra logic in data source filtering. It should take into account "sort mode". 
For example FTSimpleDataSource in this PR does not support it correctly. If you will filter sorted table then "third click" function will not work.

Ducasse:
Hi guys 

I love this discussion. This is really great reading it and learning from it. 
Thanks you. 
Stef

Ducasse:
Broken tests are unrelated

Issue #2550, opened 2019-02-15
Cleanup Keymapping-Core


astares:
- tag uncategorized classes
- add <scipt> tags to enable/disable debugging on KMLog

astares:
Fixed in https://github.com/pharo-project/pharo/pull/2551

MarcusDenker:
merged pr

PR #2551, opened 2019-02-15
2550 Cleanup Keymapping-Core

astares:
https://github.com/pharo-project/pharo/issues/2550

* tag uncategorized classes
* add  tags to enable/disable debugging on KMLog


MarcusDenker (approved):
tag changes only

Issue #2562, opened 2019-02-15
Remove unused cache from TermInfoCharacter


astares:

astares:
Fixed in https://github.com/pharo-project/pharo/pull/2563

PR #2563, opened 2019-02-15
2562 Remove unused cache from TermInfoCharacter

astares:
https://github.com/pharo-project/pharo/issues/2562


MarcusDenker (approved):
ok

Ducasse:
Same remark. For trivial fix please integrate them. It just produces noise and I think that people should focus their energy on reviewing more challenging impacting changes.

Issue #2564, opened 2019-02-15
HandMorph cleanups required


astares:
HandMorph allUnreferencedInstanceVariables 

returns "#(#keyboardListeners #eventSource #lastSystemEvent)"

Do we still need them?

astares:
fix in https://github.com/pharo-project/pharo/pull/2565

jecisc:
They cannot be removed or your image will crash. 

I am already working on removing unused instance variable (https://github.com/pharo-project/pharo/pull/2468, https://github.com/pharo-project/pharo/pull/2503, https://github.com/pharo-project/pharo/issues/2494) and I already know most of the ivar we should not remove. If you can way a little I'll finish my cleanup, add a way to mark which unused var should be kept and add a release test.

astares:
Already fixed a few today.

No - my image is working fine after removing them. Did I miss something?

jecisc:
![screenshot 2019-02-15 at 11 17 56](https://user-images.githubusercontent.com/9519971/52850289-6333f580-3113-11e9-9167-c738b8a482cd.png)

Yes, it does make the image crash in some cases. I don't know this part of the system well enough to know how we get those events making the image crash, but it happens.

astares:
Maybe this is on Mac due to the migrations of ivar order and event processing when loading the code or applying the change.

I can work with the image without problems on Linux and Windows. If the image is freshly built it should work anyhow. Can you try a CI built fresh image of the PR on Mac?

astares:
@jecisc Can you try again on Mac?

hilaire:
I just tested on Ubuntu/P8 64 bits, no crash. 
Just an error once right after the changes.

```
Point(Object)>>doesNotUnderstand: #hand
	Receiver: (284@108)
	Arguments and temporary variables: 
		aMessage: 	hand
		exception: 	MessageNotUnderstood: Point>>hand
		resumeValue: 	nil
	Receiver's instance variables: 
		x: 	284
		y: 	108


MouseOverHandler>>handleAsMouseOver:
	Receiver: a MouseOverHandler
	Arguments and temporary variables: 
		anEvent: 	(284@108)
	Receiver's instance variables: 
		mouseOverMorphs: 	an Array(a RubEditingArea(773896448) a TransformWithLayoutMorp...etc...
		enteredMorphs: 	a WriteStream
		overMorphs: 	a WriteStream
[...]

PR #2565, opened 2019-02-15
2564 HandMorph cleanups required

astares:
https://github.com/pharo-project/pharo/issues/2564


jecisc (changes requested):
Some of those are indirectly used. If you remove them you wight wake the image unusable in some cases.

jecisc:
@astares It makes two time in a row that all three platform fails.

Ducasse:
I added the tag Need more work so that I do not read 20 times the same issue while new commits should be made.

Issue #2571, opened 2019-02-16
RBPatternWrapperBlockNode missing #sentMessages


j-brant:
`RBProgramNode` and subclasses have `#sentMessages` methods that return all of the messages sent by the AST node and its children. This is used to optimize searching for patterns in the lint rules (and other places). It can be used to check that a method sends all of the `#sentMessages` before parsing the method to search for the pattern. The `RBPatternWrapperBlockNode` is missing an implementation of the `#sentMessages` method, so searching for a pattern like `` `@rcvr yourself `{:node | node isUsed}`` requires parsing the whole image when it should only need to parse the methods that send `#yourself`

PR #2572, opened 2019-02-16
2571-RBPatternWrapperBlockNode-missing-sentMessages

j-brant:
adding RBPatternWrapperBlockNode>>sentMessages
fixes case #2571


Ducasse:
Thanks John! And with tests!

Issue #2573, opened 2019-02-16
SortedCollection slow when sorting collections with many equal elements


j-brant:
Using a SortedCollection to sort collections with many equal values can be slow. The quick sort implementation degenerates to the n^2 case. For example, on my machine sorting 20,000 equal integers takes about 1 second (`Time millisecondsToRun: [(Array new: 20000 withAll: 1) asSortedCollection]`), but sorting 20,000 different integers takes only a couple milliseconds (`Time millisecondsToRun: [(1 to: 20000) asSortedCollection]`).

PR #2574, opened 2019-02-16
2573-SortedCollection-slow-when-sorting-collections-with-many-equal-elements

j-brant:
change the quick sort implementation to skip over equal elements before recursively sorting partitions -- changes the running time of "sorting" a collection of equal values from O(n^2) to be O(n)
fixes case #2573


Ducasse:
Thanks! The broken tests are unrelated.

svenvc:
Great catch ! Thank you

Issue #2459, opened 2019-02-08
Many VM crashes during Pharo checks


jecisc:
I have a PR I already relaunched 5 times and each time a different OS-specific check fail because there is a VM crash.

This make it really hard to validate PRs.

Ducasse:
Let me know if I close it or not.

akgrant43:
> Let me know if I close it or not.

The changes I've made won't solve the problem - if we're lucky they'll help narrow down what introduced it.

So... this will probably be open for a while yet.

PR #2650, opened 2019-02-19
Log the Target VM date and CommitHash during bootstrap

akgrant43:
At the moment it is difficult to be sure exactly which VM is being used
to bootstrap Pharo.  This will assist in tracking down VM issues during
the bootstrap process.

In support of #2459


akgrant43:
Thanks, Stef.

PR #2654, opened 2019-02-19
Use the stable VM instead of Latest for bootstrap

akgrant43:
In support of #2459

As described in #2459, the bootstrap process is failing quite often with
VM crashes.  This will allow us to get an idea of whether it is due to a
recent change in the VM, or something else (which may or may not be part
of the VM).


jecisc:
I relaunched the CI but the change seems fine.

akgrant43:
Hi Cyril,

Thanks for the relaunch.

I think this should still be merged so that we can get a sense of the behaviour over multiple days and with slightly difference versions of the image (from the various PRs).

Cheers,
Alistair

jecisc (approved):
Build passed. Test failures not related.

akgrant43:
Thanks, Cyril & Stef.

PR #2658, opened 2019-02-19
2657-Duplication-in-RB-variable-related-classes

### LINKED ISSUE #2657: Duplication in RB variable related classes
RBProtectVariableTransformation, RBPullUpVariableTransformation, and RBPushDownVariableTransformation all have some duplicated methods and are the only subclasses of RBCompositeVariableTransformation using one of its variables. 

Maybe they should have a common superclass, the variable should be pushed down and the methods pushed up.

Methods:
- #storeOn:
- #variable:class:classVariable:

jecisc:
Remove duplication in variable transformations.

Fixes #2657

### DIFF
--- a/src/Refactoring2-Transformations/RBCompositeVariableTransformation.class.st
+++ b/src/Refactoring2-Transformations/RBCompositeVariableTransformation.class.st
@@ -3,74 +3,10 @@ Class {
 	#superclass : #RBCompositeTransformation,
 	#instVars : [
 		'variableName',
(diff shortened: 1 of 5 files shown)


MarcusDenker:
RBTransformationTest is failing.

Ducasse:
I added the tag Need more work so that I do not read 20 times the same issue while new commits should be made.

MarcusDenker:
I think we should close PRs with problems after one Month

PR #2667, opened 2019-02-20
2666-Show-slot-class-definition-by-default

### LINKED ISSUE #2666: Show slot class definition by default
I discussed with some people at RMoD and we are multiple to think it's time to enable by default the show slot definition setting by default in the browser.

If nobody as an argument against it, I'll do a PR.

jecisc:
Show slot definition by default. 

Fixes #2666


jecisc:
Tests failures are related.

Apparently, we have multiple tests depending on the default definition.

jecisc:
Needs more work

PR #2690, opened 2019-02-22
22595-Median-is-not-calculated-correctly-for-even-sizes

ghost:
Fix case 22595: Median is not calculated correctly for even sizes


Ducasse:
Thanks for the catch! and the tests!

Ducasse:
Could you remove the middle check for a number?

SortedCollection >> median [
	"Return the middle element, or as close as we can get."

	|  middle |
	middle := (self size + 1) // 2.
	^ (self size even
		ifTrue: [ ((self at: middle) + (self at: middle + 1)) / 2 ]
		ifFalse: [ self at: middle ]
]

Issue #2578, opened 2019-02-17
Test result notification uses wrong plural forms


rillig:
I have added a second test case to `PDFColorTest` and made it fail intentionally, just to get my hands wet with testing.

![image](https://user-images.githubusercontent.com/3233724/52912821-2c96e000-32b7-11e9-9fa4-326466062114.png)

The yellow notification says "1 failures", which is grammatically wrong. It should be "1 failure" instead.

rillig:
Here is my suggestion for `TestResult.printOn`:

```smalltalk
printOn: aStream
	| putN sep |
	sep := ''.
	putN := [ :n :singular :plural |
		n = 0 ifFalse: [
			aStream
				nextPutAll: sep;
				nextPutAll: n printString;
				nextPutAll: ' ';
				nextPutAll: (n = 1 ifTrue: singular ifFalse: plural).
			sep := ', '
		]
	].

	putN value: self runCount value: 'run' value: 'run'.
	putN value: self expectedPassCount value: 'pass' value: 'passes'.
	putN value: self skippedCount value: 'skipped' value: 'skipped'.
	putN value: self expectedDefectCount value: 'expected failure' value: 'expected failures'.
	putN value: self unexpectedFailureCount value: 'failure' value: 'failures'.
	putN value: self unexpectedErrorCount value: 'error' value: 'errors'.
	putN value: self unexpectedPassCount value: 'unexpected pass' value: 'unexpected passes'.
	sep = '' ifTrue: [ aStream nextPutAll: 'no tests run' ].
```

I tested it interactively, but as a Pharo beginner, I cannot estimate what influence the changed formatting of this message has on the whole system.

svenvc:
Did you see String>>#asPluralBasedOn: ?
Check its senders as well.
That will simplify your code.

rillig:
> Did you see String>>#asPluralBasedOn: ?

I think the above code has too many exceptional cases to make `asPluralBasedOn` useful in this case. For example, there is `run` and `run` (which presumably is an abbreviation of `test has been run` and `tests have been run`), and there is `pass` and `passes`. Therefore I prefer to leave the code as it is right now.

Thanks for the hint anyway, I would not have thought that such a function existed at all since most other programming languages also don't have it. Furthermore it is useless in an internationalized context where [the rules are much more complicated](https://www.unicode.org/cldr/charts/34/supplemental/language_plural_rules.html).

bencoman:
[...]

PR #2695, opened 2019-02-22
2578-Test-result-notification-uses-wrong-plural-forms

svenvc:
Improve TestResult>>#printOn: to better take singular/plural forms into account

https://github.com/pharo-project/pharo/issues/2578

### DIFF
--- a/src/SUnit-Core/TestResult.class.st
+++ b/src/SUnit-Core/TestResult.class.st
@@ -321,21 +321,25 @@ TestResult >> passedCount [
 
 { #category : #printing }
 TestResult >> printOn: aStream [
+	self runCount isZero ifTrue: [ ^ aStream nextPutAll: 'No tests ran' ].
 	aStream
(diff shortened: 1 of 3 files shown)


Ducasse:
This is strange because we have many broken and related tests. 
Can you have a look?

Ducasse:
Ok we will have to update the tests because they checked the exact string.

svenvc:
Right, there are a couple of tests that literally test the print string, duh. 
I should have checked those, not clever.
I will try to fix this as soon as possible.

svenvc:
I fixed the tests, but they remain literal string tests

svenvc:
Failed tests does not seem to be related (existed before)

svenvc:
Fixes 2578

Ducasse:
Thanks sven!!!. 
We will have a new pass on the testing framework soon and a tool with new features!

Issue #2624, opened 2019-02-18
GTDebugger inconsistency of "Create" option


jecisc:
Migrated case from Manuscript.
	
Original case: https://pharo.fogbugz.com/f/cases/20052
Status: Work Needed
Project: GTools
Original Author: CyrilFerlicot
Date: 11 May 2017 9:22:48 pm
	
Description:
	
In the GTDebugger when we have a "MNU" a create button is at the left of the "Proceed" button. When we have a "SubclassResponsibility", the "Create" button is at the right of "Through". 

This behaviour is inconsistent and often I miss the "create" button in the subclass responsibility case.

PR #2703, opened 2019-02-22
Moved the "create" button of the debugger when a SubclassResponsibili…

570gina:
…ty error is raised to the left of the other debug action buttons, to be at the same place as the "create" button when a MessageNotUnderstood exception is raised
Solves #2624


jecisc (approved):
Great thank you!

I was always looking for the button :)

Ducasse:
Tx!

PR #2705, opened 2019-02-22
2704-Improve-as-when-the-class-is-already-of-the-right-kind

### LINKED ISSUE #2704: Improve #as: when the class is already of the right kind
Currently, #as: creates a new object all the time.

Maybe we could add `self class == aSimilarClass ifTrue: [ ^ self ]`

jecisc:
Do not create a new instance in #as: if we already have the right kind.

+ add a test.

Fixes #2704


Ducasse:
Broken tests are unrelated

PR #2706, opened 2019-02-22
2701-Remove-StreamCollectionwrite

### LINKED ISSUE #2701: Remove [Stream|Collection]>>#write:
Did you know that both Stream and Collection respond to #write: (as an alias to #<< and #add: respectively) ? I sure did not (and I did quite a bit of stream hacking).

This selector is not used for these 2 receiver types in the standard image (except for one unit test), I never saw it being used in actual code.

These are just aliases that add nothing but cognitive load.

They make the API's of two important hierarchies (and especially the interface for non-inheriting classes) heavier than needed for no functional benefit.

So I propose to remove [Stream|Collection]>>#write:

svenvc:
Remove Stream and Collection  #write: (as an alias to #<< and #add: respectively) 

This selector is not used for these 2 receiver types in the standard image (except for one unit test), I never saw it being used in actual code.

These are just aliases that add nothing but cognitive load.

They make the API's of two important hierarchies (and especially the interface for non-inheriting classes) heavier than needed for no functional benefit.

Fixes #2701

### DIFF
--- a/src/Collections-Abstract/Collection.class.st
+++ b/src/Collections-Abstract/Collection.class.st
@@ -1290,11 +1290,6 @@ Collection >> union: aCollection [
 	^ self species withAll: set asArray
 ]
 
-{ #category : #'filter streaming' }
-Collection >> write: anObject [ 
-	^ self add: anObject
-]
-
 { #category : #enumerating }
 Collection >> | aCollection [
 	^ self union: aCollection
--- a/src/Collections-Streams/Stream.class.st
+++ b/src/Collections-Streams/Stream.class.st
@@ -224,11 +224,3 @@ Stream >> upToEnd [
 		elements add: self next ].
 	^elements
 ]
-
-{ #category : #accessing }
-Stream >> write: anObject [
-	"Like <<, but returns the object putOn the receiver."
-
-	anObject putOn: self.
-	^ anObject.
-]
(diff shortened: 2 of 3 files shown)


jecisc:
Maybe we could use a deprecation warning with a rewrite rule in case external projects are using it.

svenvc:
I really doubt this is a commonly used selector.
Hence a deprecation rule would be a bit overkill.
On the other hand, if someone proposes it, we could add it.

svenvc:
Failed test does not seem to be related (existed before)

jecisc:
I will do next week because the cost is so small, if it helps even some people to migrate it’s still useful.

svenvc:
OK, great, thanks.

estebanlm:
@sven: the proper way to refer to an issue is to put: 

Fixes #2701

This will automatically create the reference (you do not have to go there and put a link to here), with the additional benefit that merging the PR will automatically close the issue.

estebanlm:
(but is you who has to do it :P)

estebanlm:
Now... not to be pretentious but, wouldn't the correct removal process be to deprecate it? (and then remove it next version?)

Ducasse:
Yes we should deprecate it :) but sven was in "cleaning rage mode" like I'm :)

Issue #2709, opened 2019-02-23
Rename KmGlobalDispatcher to KMGlobalDispatcher


Ducasse:

Ducasse:
PR submitted

PR #2710, opened 2019-02-23
2659-Update-and-rename-KmGlobalDispatcherclassst-to-KMGlobalDispatcherc

Ducasse:
Fixes: #2709 Apparently there is no bug entry so I created a new one https://github.com/pharo-project/pharo/issues/2709


Ducasse:
Broken tests are unrelated.

PR #2713, opened 2019-02-23
2712-deprecate-write-methods-so-that-people-see-that-we-are-serious-about-backward-compatibility

### LINKED ISSUE #2712: deprecate write methods so that people see that we are serious about backward compatibility
in it was not obvious that WE REALLY CARE ABOUT BACKWARD COMPATIBILITY but not to the point to die killed by immobilism!

Ducasse:
Fixes #2712  https://github.com/pharo-project/pharo/issues/2712


Ducasse:
I will try to recommit it to learn how to deal with such situation.

PR #2719, opened 2019-02-24
2718-Rubric-should-not-depend-on-TextEditor

Ducasse:
Fixing dependencies from Rubric to editor


Ducasse:
This is strange because the reference is just within the same package instead of having it between different packages and I did not remove anything.

jecisc (approved):

jecisc:
Test failures not related.

Ducasse:
Thanks cyril. I could not see why it was crashing before. Super strange.

PR #2721, opened 2019-02-25
2587-Cannot-remove-a-Trait

tesonep:
When Calypso removes a Trait it uses the refactoring browser. As RB is not prepared for stateful traits I have to extend it. 
Also I have improved the detection of subclasses.

### DIFF
--- a/src/Kernel/Class.class.st
+++ b/src/Kernel/Class.class.st
@@ -665,7 +665,7 @@ Class >> hasSharedPools [
 
 { #category : #'accessing class hierarchy' }
 Class >> hasSubclasses [
-	^subclasses notNil
+	^ self subclasses isEmpty not
 ]
 
 { #category : #'subclass creation - immediate' }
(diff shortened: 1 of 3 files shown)


src/Kernel/Class.class.st
```diff
@@ -665,7 +665,7 @@ Class >> hasSharedPools [
 
 { #category : #'accessing class hierarchy' }
 Class >> hasSubclasses [
-	^subclasses notNil
+	^ self subclasses isEmpty not
```
jecisc, line 668:
```suggestion
	^ self subclasses isNotEmpty
```

jecisc (changes requested):
I suggested a change.

MarcusDenker:
I propose to merge this now and improve later

MarcusDenker:
(because I do not find how to accept that change even after assigning the issue to me)

PR #2723, opened 2019-02-25
2722-Symbol-should-not-implement-pragmas

### LINKED ISSUE #2722: Symbol should not implement #pragmas
Currently, GTInspector is extending Symbol to add a method #pragmas. This is wrong in my opinion. We should update the code calling this method or at least rename it #gtPragmas because it does not make much sense to be able to ask "pragmas" to a symbol.

jecisc:
Remove Symbol>>pragmas

Fixes #2722

### DIFF
--- a/src/GT-InspectorExtensions-Core/Symbol.extension.st
+++ b/src/GT-InspectorExtensions-Core/Symbol.extension.st
@@ -20,10 +20,3 @@ Symbol >> gtInspectorSendersIn: composite [
 		showOnly: 30;
 		withSmalltalkSearch 
 ]
-
-{ #category : #'*GT-InspectorExtensions-Core' }
-Symbol >> pragmas [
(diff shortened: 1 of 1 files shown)


Ducasse:
Should deprecate it and may be we can move a part of it to SystemNavigation? 
I do not like the name #foo pragmas, it should have been at least #foo methodWithPragmas but it does not read well. May SystemNavigation current methodWithPragma: #foo.

jecisc:
This does not even return CompiledMethod but Rings methods.

MarcusDenker:
ah, the fact that it returns methods makes it really odd... we should check how many users it has

jecisc:
@MarcusDenker I have the impression it has none.

guillep (approved):
Looks ok, there seem to be no users of this

PR #2724, opened 2019-02-25
2590-We-can-get-two-times-the-mouseEntermouseLeave-event-when-subscribing-to-the-eventHandler-of-a-Morph

### LINKED ISSUE #2590: We can get two times the mouseEnter/mouseLeave event when subscribing to the eventHandler of a Morph
	
Original case: https://pharo.fogbugz.com/f/cases/22705
Status: Work Needed
Project: Morphic
Date: 28 November 2018 2:19:44 pm
	
Description:
	
StringMorph new
  contents: 'test;
  on: #mouseEnter send: #mouseEnter:from: to: self;
  on: #mouseLeave send: #mouseLeave:from: to: self;
  on: #click send: #mouseClick:from: to: self;
  yourself.

With this piece of code, if you add the methods to the class and log each time an event happens, you will see that we have two enter and leave event each time.

The problem comes from Morph>>handleMouseEnter/Leave:

jecisc:
Try to fix double event announcement.

I am not 100% sure but since we are in alpha of Pharo 8 maybe we can try it?

Fixes #2590


jecisc:
I am waiting for reviews...

jecisc:
Ping

guillep (changes requested):
Any chances to write a test about it? Simulating an event shouldn't be too difficult.

Ducasse:
I tried to follow the code but I failed to see it.

MarcusDenker (approved):
ok, lets try

PR #2726, opened 2019-02-25
2725-When-doing-a-FileOut-propose-to-open-the-folder-containing-the-fileout

### LINKED ISSUE #2725: When doing a FileOut, propose to open the folder containing the fileout
When I do a fileout, often the goal is to use/send it. I find logical to propose to the user to open the location after the export. 

I'll propose a PR.

jecisc:
Propose to open location of fileout after export.

Fixes #2725

### DIFF
--- a/src/CodeExport/CodeExporter.class.st
+++ b/src/CodeExport/CodeExporter.class.st
@@ -21,12 +21,20 @@ CodeExporter class >> writeSourceCodeFrom: aStream baseName: baseName isSt: stOr
 CodeExporter class >> writeSourceCodeFrom: aStream toFileReference: aFileReference [
 	"Write the supplied changes to aFileReference"
 
-	aFileReference 
+	aFileReference
 		ensureDelete;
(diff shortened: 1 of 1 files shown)


akgrant43 (approved):
Nice!  In some ways I'd prefer that an internal file browser be opened, but that's subjective, and certainly not all the time.  Maybe hold down the shift key to open the internal browser?  That's just a suggestion and doesn't stop this from being integrated now.

VM crash is unrelated, failing tests are existing issues and unrelated, enough tests have run that the VM crash can be ignored.

jecisc:
I would like to have a replacement of Growl that would be closer to native notifications where we could have a title, text, image, icon, buttons... like that we could have 2 or 3 buttons. 
But as usual... this needs time.

Ducasse:
Cyril I do not see how this is related to growl.

jecisc:
Growl currently only allows 1 title, 1 textual content and 1 action on click because it is a subclass of TextMorph. (Or StringMorph or something like this).

PR #2732, opened 2019-02-27
2731-Rescue-NewRubric-clients-to-remove-PluggableTextMorph-and-friends

### LINKED ISSUE #2731: Rescue NewRubric clients (to remove PluggableTextMorph and friends)
Once #2729 is integrated we can focus on the clients of the new rubric version.

Ducasse:
Fixes #2731 Update the clients of Rub new version


Ducasse:
Argh I do not really know how to handle this conflict ;(
I wanted to check if everything is more or less working so that Spec dev are less impacted but I will have to check and I do not have time until tuesday or thursday.

guillep:
I've fixed the conflict. Let's wait for the tests to run.

Ducasse:
So 99 tests are breaking in Spec :)
What should I do?
Push and we fix after?

guillep:
If you give me until tomorrow I can try to fix it (just in case everything becomes unusable...)

Ducasse:
This is what I thought. And alain told me that we can pair program thursday or friday evening :)
So we can start to massively clean.

src/Spec-MorphicAdapters/MorphicTextInputFieldAdapter.class.st
```diff
@@ -29,8 +29,7 @@ MorphicTextInputFieldAdapter >> adapt: aModel [
 { #category : #factory }
 MorphicTextInputFieldAdapter >> buildWidget [
 	| plu |
-	plu := PluggableTextFieldMorph new
-		convertTo: String;
+	plu := RubPluggableTextFieldMorph new
```
guillep, line 32: The old PluggableTextFieldMorph implemented `maxLength:` and Rub's does not...

guillep: The same with `hideScrollBarsIndefinitely`

guillep: It also does not have a getter for `ghostText` (which we require for testing that we set the right thing)

guillep: We also are using `widget setTextAndAccept: aString`

guillep (changes requested):
Those are all the things that would be missing at least from a Spec perspective. We should at least know what are the replacements for them (or if we should not use some method for example).

Also, we have found a bug in:
```smalltalk
RubScrolledTextMorph >> canDiscardEdits
	^ (hasUnacceptedEdits and: [askBeforeDiscardingEdits]) not

RubScrolledTextMorph >> resetState
	hasEditingConflicts := nil.
	hasUnacceptedEdits := nil.
	self changed

RubScrolledTextMorph >> hasEditingConflicts
	^ hasEditingConflicts ifNil: [ hasEditingConflicts := false ] 
```
It seems that the first method should not have a direct access to `hasEditingConflicts ` but access it through the accessor. Otherwise it may happen that we find a `nil` there and a `mustBeBoolean`.

src/Spec-MorphicAdapters/MorphicTextInputFieldAdapter.class.st
guillep: Then... there is another more fundamental problem with this change, and it is the fact that it may kill Spec's compatibility with Pharo7.

Hmm well, looking at it, we have in Pharo7
```smalltalk 
canDiscardEdits
	^self hasUnacceptedEdits not
```
So if we can add the missing methods as extension methods in a compatibility package, that may be enough.

jecisc:
@guillep For info: When Alain tried to integrate his changes into Pharo, I got Spec text presenter working with this code for the adaptor:

```Smalltalk
buildWidget
	| newWidget |
	newWidget := (RubTextFieldMorph on: self)
		getTextSelector: #getText;
		setTextSelector: #accept:notifying:;
		getSelectionSelector: #readSelection;
		menuProvider: self selector: #codePaneMenu:shifted:;
		setSelectionSelector: #setSelection:;
		entryCompletion: self entryCompletion;
		ghostText: self placeholder;
		encrypted: self encrypted;
		enabled: self enabled;
		askBeforeDiscardingEdits: self askBeforeDiscardingEdits;
		autoAccept: self autoAccept;
		vResizing: #spaceFill;
		hResizing: #spaceFill;
		setBalloonText: self help;
		dragEnabled: self dragEnabled;
		dropEnabled: self dropEnabled;
		yourself.
	self setEditingModeFor: newWidget withBehavior: self behavior.
	^ newWidget
```

guillep:
> @guillep For info: When Alain tried to integrate his changes into Pharo, I got Spec text presenter working with this code for the adaptor:
> 
> ```smalltalk
> buildWidget
> 	| newWidget |
> 	newWidget := (RubTextFieldMorph on: self)
> 		getTextSelector: #getText;
> 		setTextSelector: #accept:notifying:;
> 		getSelectionSelector: #readSelection;
> 		menuProvider: self selector: #codePaneMenu:shifted:;
> 		setSelectionSelector: #setSelection:;
> 		entryCompletion: self entryCompletion;
> 		ghostText: self placeholder;
> 		encrypted: self encrypted;
> 		enabled: self enabled;
> 		askBeforeDiscardingEdits: self askBeforeDiscardingEdits;
> 		autoAccept: self autoAccept;
> 		vResizing: #spaceFill;
> 		hResizing: #spaceFill;
> 		setBalloonText: self help;
> 		dragEnabled: self dragEnabled;
> 		dropEnabled: self dropEnabled;
> 		yourself.
> 	self setEditingModeFor: newWidget withBehavior: self behavior.
> 	^ newWidget
> ```

Ok, I could live if you tell me to remove the call to `hideScrollBarsIndefinitely`. Also I'm sure there will be a way to access the `ghostText` and to set a text and accept it.

But what about `maxLength:`? Do we drop that feature?

Ducasse:
What is the next step? I'm lost :)

Ducasse:
I will try to check it with Alain tomorrow evening.

Ducasse:
I checked maxlength: and it looks like a hack. The only use is in TextMorphForFieldView>>keystroke.

keyStroke: evt
	"Don't allow editing keys if the edit view is disabled.
	Do the key if max length has not been reached.
	Don't allow tabs." 
	(self localHandleKeystroke: evt) ifTrue: [^self].
	(self maxLength isNil or: [self text size < self maxLength or: [
			self editor selectionInterval size > 0 or: [self isEditEvent: evt]]])
		ifTrue: [self doAutoAccept: evt].
	super keyStroke: evt

Ducasse:
Guille do you have a test using ghostText: because I did not find one and I wanted to use it to write a test for ghostText.

Ducasse:
In RubScrolledTextMorph both hasEditingConflicts and hasUnacceptedEdits are set to nil and used correctluy (i.e. using accessors). Now I do not understand why Calypso does NOT show me 
canDiscardEdits when I look for reference to hasUnacceptedEdits! grgrgr

Now in RubPluggableTextMorph it is not the case anymore since the two variables are initialized with booleans. 
 
initialize
	super initialize.
	hasUnacceptedEdits := false.
	hasEditingConflicts := false.
	askBeforeDiscardingEdits := true

Ducasse:
Ok I understand that a change was pushed to fix 

RubScrolledTextMorph >> canDiscardEdits
	^ (hasUnacceptedEdits and: [askBeforeDiscardingEdits]) not
into 
RubScrolledTextMorph >> canDiscardEdits
	^ (self hasUnacceptedEdits and: [askBeforeDiscardingEdits]) not

Ducasse:
I still suggest that hasEditingConflicts and hasUnaccepedEdits should be initialized to false and avoid lazy initialize in RubScrolledTextMorph like that this is the same logic in subclass. And we cando the same for hasValidText (to be initialized to true).

Ducasse:
In the new branch I created with Guillermo named recoverRub2Client in my repo I initialized the three variables to their respective booleans and clean the protocols (initialize release and aasYet unclassified). 

Next step:
- maxLength: usage.

Ducasse:
testGhostText
	presenter placeholder: 'This is a cool ghost text'.
	self
		assert: presenter placeholder
		equals: 'This is a cool ghost text'

So apparently you do something magic because my code ghostText is rewritten to this.

Ducasse:
Ok so this is placeholder in Spec.

Ducasse:
May be I'm tired but I could not understand why 

MorphicTextInputFieldAdapter >> placeholder
	^ self model placeholder

and not 
       ^ self widget ghostText

jecisc:
In general, methods of the adapters are here to access presenter value, not widget values.

guillep:
> May be I'm tired but I could not understand why
> 
> MorphicTextInputFieldAdapter >> placeholder
> ^ self model placeholder
> 
> and not
> ^ self widget ghostText

This is purely a convenience. PluggableXXXMorphs require a model for the get/set[XXX]selector(s).
So we are using the adapters as the model of the pluggable morphs. So those methods are used only by the morph and should not be used elsewhere... We are trying to avoid them whenever we can (for example fast table announces its changes with announcements so we don't need that mechanism).

In any case, I think that the name `model` is not good in the context of the adapter. We should use either `widget` or `presenter` to make it more explicit...

MarcusDenker:
Now we have test failing in the release

PR #2741, opened 2019-03-01
2740-do-not-zap-method-dict-when-making-classes-obsolete

### LINKED ISSUE #2740: do not zap method dict when making classes obsolete
There is an ugly hack in the system: when we obsolete classes, we clean out the method dict.
But just on the instance side, not the class side.

This was added to be able to "shrink" an image with lots of obsoletes. 

Does this make sense? 
- not uniform: we do it for instance side, but not class side
- space saving: when the obsolete class is GCed, the method will go away, too
- would allow to execute method on obsolete classes (this is needed for the class side methods already)

==> remove the #canZapMethodDictionary infrastructure.

MarcusDenker:
do not zap method dict when making classes obsolete

fixes #2740

### DIFF
--- a/src/Kernel-Tests-Extended/CompiledMethodTest.class.st
+++ b/src/Kernel-Tests-Extended/CompiledMethodTest.class.st
@@ -342,7 +342,7 @@ CompiledMethodTest >> testIsInstalled [
 	method := (self class)>>#returnTrue.
 	self assert: method isInstalled.
 
-	"now make an orphaned method by just deleting the class."
+	
 
 	Smalltalk removeClassNamed: #TUTU.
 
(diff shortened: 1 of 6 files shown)


MarcusDenker:
errors are related. need to fix tests

MarcusDenker:
testIsInstalled 
 testObsoleteClassRemovesMethodsClassSide 
 testObsoleteClassRemovesMethods

Ducasse:
I was wondering why we do not zap on class side?
Because an obsolete class can still execute method?
I liked the idea that we do not really on gc to get rid of data that we know should be removed. 
With a nice method comment it would be ok for me.

MarcusDenker:
We had an issue with meta-link deinstallation that would just work if the methods are not removed.. but quite hard to solve else.

MarcusDenker:
Errors are not related: all spec failures, fail for other PR, too

pavel-krivanek (approved):

PR #2754, opened 2019-03-06
2753-FileReference-version-number-issues

### LINKED ISSUE #2753: FileReference version number issues
FileReference>>versionNumber has a number of issues:

- '/dev/shm/foo.bar.txt' asFileReference ensureCreateFile; nextVersion -> "File @ /dev/shm/foo.1.txt"
  incorrectly changing the base file name.
- '/dev/shm/foo.bar' asFileReference nextVersion will cause a primitiveFailure if there is a directory /dev/shm/foo.
- The method is quite inefficient as it requests a collection of all children of the parent directory, which can be quite large.

There is also an issue that associated methods:

- #nextVersion 
- #lastFileFor:extension:
- #nextNameFor:extension:

all do their own parsing of file names, resulting in duplicated code and possible inconsistencies.

Refactor the code to provide a single implementation of parsing the file name for the version number and fix the three issues listed above.

PR to be supplied.

akgrant43:
FileReference>>versionNumber has a number of issues:

- '/dev/shm/foo.bar.txt' asFileReference ensureCreateFile; nextVersion -> "File @ /dev/shm/foo.1.txt" incorrectly changing the base file name.
- '/dev/shm/foo.bar' asFileReference nextVersion will cause a primitiveFailure if there is a directory /dev/shm/foo.
- The method is quite inefficient as it requests a collection of all children of the parent directory, which can be quite large.

There is also an issue that associated methods:

- #nextVersion
- #lastFileFor:extension:
- #nextNameFor:extension:

all do their own parsing of file names, resulting in duplicated code and possible inconsistencies.

- Refactor the code to provide a single implementation of parsing the file name for the version number and fix the three issues listed above.
- Extend automated tests to cover the issues listed above.

Fixes: #2753

### DIFF
--- a/src/FileSystem-Core/FileReference.class.st
+++ b/src/FileSystem-Core/FileReference.class.st
@@ -357,15 +357,15 @@ FileReference >> lastFileFor: baseFileName extension: extension [
 
 	"FileSystem workingDirectory lastFileFor: 'games' extension: 'ston'"
 
-	| files |
-	files := self childrenMatching: baseFileName , '.*.' , extension.
-	files ifEmpty: [ ^ self error: 'No file with number pattern' ].
-	"splitting twice but for now a simple '*.*.*' does not work since it selects: name.45.677.mcz"
-	files := files select: [ :each | (each basename splitOn: $.) size = 3].
-	^ (files
-		asSortedCollection:
-			[ :a :b | ((a basename splitOn: $.) at: 2) asNumber < ((b basename splitOn: $.) at: 2) asNumber ])
-		last
+	| version |
+
+	version := self versionNumberFor: baseFileName extension: extension.
+	^(version
+		ifNil: [ self error: 'No file with number pattern' ]
+		ifNotNil: [ version = 0
+			ifTrue: [ baseFileName, '.', extension ]
+			ifFalse: [ baseFileName , '.' , version asString , '.' , extension ] ]) asFileReference 
+
 ]
 
 { #category : #accessing }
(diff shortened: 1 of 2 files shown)


akgrant43:
CI failing tests are pre-existing and appear to be unrelated.

svenvc:
I read the code on github and with the tests this looks like a massive improvement.

akgrant43:
Thanks, Sven! :-)

Ducasse:
Thanks Alistair. I really rely on the #lastFileFor:extension:
#nextNameFor:extension: because nextVersion logic was too simplistics. This is cool that you improved my code :).

src/FileSystem-Core/FileReference.class.st
```diff
@@ -580,6 +575,26 @@ FileReference >> uid: uid gid: gid [
...
 
+{ #category : #versions }
+FileReference >> versionNumberFor: basename extension: extension [
+	"Answer the latest (largest) version number for the specified file.
+	0 = basename.extension exists, but nothing later.
+	nil = no file exists"
+
+	| regex maxVersion |
+
+	regex := (basename, '.(\d+).', extension) asRegex.
```
dionisiydk, line 586: It adds implicit dependency on regex engine. Do we know if it is correct or not? At what time regex is loaded? I suspect that it would be broken dependency.

dionisiydk: Another notice: can we be sure that baseline and extension parts of regex string are safe to be in regex expression? I think not. And it requires escaping which we don't have.

So I think it would be better to avoid regex with special helper method. It will also improve performance because regex in Pharo is highly unoptimized.

akgrant43: Hi Dennis,

> It adds implicit dependency on regex engine. Do we know if it is correct or not? At what time regex is loaded? I suspect that it would be broken dependency.

Regex is loaded immediately after FileSystem-Memory.

I thought there were automated tests for the dependencies, and I don't remember anything being flagged.


> Another notice: can we be sure that baseline and extension parts of regex string are safe to be in regex expression? I think not. And it requires escaping which we don't have.

Can you provide an example of an unsafe file name?


> So I think it would be better to avoid regex with special helper method. It will also improve performance because regex in Pharo is highly unoptimized.

It would be nicer to optimise Regex - but since I'm not volunteering I'm not pushing the point :-)

Thanks,
Alistair

dionisiydk: > > It adds implicit dependency on regex engine. Do we know if it is correct or not? At what time regex is loaded? I suspect that it would be broken dependency.
> 
> Regex is loaded immediately after FileSystem-Memory.
> 
> I thought there were automated tests for the dependencies, and I don't remember anything being flagged.

I am sure we don't have tests for such kind of dependency. It is just a message send, not a class reference. 

> > Another notice: can we be sure that baseline and extension parts of regex string are safe to be in regex expression? I think not. And it requires escaping which we don't have.
> 
> Can you provide an example of an unsafe file name?

I don't know really. But #asRegex could easily fail on arbitrary string. But maybe "file name format" by itself is restricted enough to be safe for regex syntax.

akgrant43: > I am sure we don't have tests for such kind of dependency. It is just a message send, not a class reference.

D'oh, of course.

> I don't know really. But #asRegex could easily fail on arbitrary string. But maybe "file name format" by itself is restricted enough to be safe for regex syntax.

It's finally clicked for me what you are talking about. :-)
Linux file names don't have any restriction (apart from / and null) so they could easily create a bad regex expression.

I'll re-write this as you suggest (although it will be after I've finished looking at ZnMimePart).

Thanks,
Alistair

dionisiydk: I decided to play with it and done: https://github.com/pharo-project/pharo/pull/2848

PR #2756, opened 2019-03-06
Fix incorrect cascade in DomainError>>signalFrom:to:

cdlm:
An easy one… instead of delegating to the canonical `signal:from:to:`, it was signalling first (`signal:from:`) and then cascading `to:`, which is not implemented on class-side.


svenvc:
cascades can be dangerous, good catch

MarcusDenker:
failing tests seem to be all related to spec

MarcusDenker (approved):
ok

PR #2758, opened 2019-03-06
2751-rename-RBBLintRuleTests-to-RBBLintRuleTestData-and-add-comment

### LINKED ISSUE #2751: rename RBBLintRuleTests to RBBLintRuleTestData and add comment
comment should explain that this code is test data for RB and should thus not be "fixed"

MarcusDenker:
add *Data prefix as *Test makes people want to clean these classes. I added a comment, too.

fixes #2751


MarcusDenker:
seems there are references by name to the classes from tests outside of the RB tests --> need to fix.

(lots of failing tests are related to spec, though)

MarcusDenker:
errors not related

PR #2760, opened 2019-03-06
2759-Spec-Deprecated-should-be-removed-from-Pharo-8

### LINKED ISSUE #2759: Spec-Deprecated should be removed from Pharo 8
This package should be removed. To avoid any conflict with the updated Spec I'll do a PR to remove it.

jecisc:
Remove Spec-Deprecated.

Fixes #2759


Ducasse:
Are you sure that you want to remove the deprecated? Should we publish it into migrator?

jecisc:
This package contains the deprecated stuff from Pharo 7. Not the deprecation from Spec 2.

It would fit nicely in a transition package from Pharo 6 to 7.

Ducasse:
Ok.

Issue #2763, opened 2019-03-06
Bag doesn't support #asDictionary, Walkback DNU #key


macta:
I was surprised to find that a Bag can’t convert to a dictionary (even though it has a mapping between values and occurences) - e.g.

Bag new 
	addAll: 'aabbbbcddd’; 
	asDictionary 

Gives an error - Dnu #key

It looks to me like Bag is inheriting a bad version of #associationsDo: 

I think it should support this, or throw an exception and perhaps guide the user to #valuesAndCounts.

MarcusDenker:
merged PR

PR #2764, opened 2019-03-06
Provide Bag compatibility with asDictionary and don't throw a walkback error

macta:
There is still some discussion about this on the mailing list - as would you expect the inverse -

 { #foo->2. #bar->3 } asDictionary asBag.

to give a bag the same (currently it gives stats on the values).

This is what I expected to work when I hit this - however could be convinced that asDictionary throws an exception (and tells you about #valuesAndCounts)

### DIFF
--- a/src/Collections-Tests/BagTest.class.st
+++ b/src/Collections-Tests/BagTest.class.st
@@ -369,6 +369,22 @@ BagTest >> testAsBag [
 ]
 
 { #category : #'basic tests' }
+BagTest >> testAsDictionary [
+
+	| aBag aDictonary |
+	aBag := Bag new.	
+	aBag add:'a' withOccurrences: 4.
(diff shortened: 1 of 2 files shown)


Ducasse:
Thanks Tim I thought the same key with objects and occurrences.

macta:
This would fix #2763

macta:
The mailing list discussion seems a bit split on this one - personally I expected asDictionary to work and was surprised it didn't however others advocate that it shouldn't. A better exception would have helped me (as the current one is very deep and rather mysterious). 

I (and others) can't see the harm of asDictionary particularly as Exercism and other types of kata's end up using a Bag and provide tests that compare against Dictionaries - so we end up fighting against wider usage.

I don't mind either way - this was good practice.

MarcusDenker (approved):
I like it

PR #2767, opened 2019-03-07
2766-RubScrolledTextMorphcanDiscardEdits-need-to-use-accessor-as-var-can-be-nil

### LINKED ISSUE #2766: RubScrolledTextMorph>>#canDiscardEdits: need to use accessor as var can be nil
hasUnacceptedEdits can be nil. it needs to use the accessor

MarcusDenker:
use accessor
fixes #2766


MarcusDenker:
It is GREEN!!!!

PR #2768, opened 2019-03-07
2765-Feature-Add-disableDuring-in-Epicea

### LINKED ISSUE #2765: Feature: Add #disableDuring: in Epicea
For some cases, we want to disable Epicea the time we are doing something and turn it on again after (if it was originally on).

Example:
- While generating a Famix Metamodel
- While generating a visitor
- While exporting/importing a lot of files in Seaside
- While compiling methods inside a test
- Etc

With Benoît, we propose to introduce a method to Epicea `disableDuring:` taking a block as a parameter.

jecisc:
Add the possibility to disable Epicea during the execution of a block.

Fixes #2765

### DIFF
--- a/src/Epicea-Tests/EpDisabledIntegrationTest.class.st
+++ b/src/Epicea-Tests/EpDisabledIntegrationTest.class.st
@@ -17,6 +17,18 @@ EpDisabledIntegrationTest >> testDisableDuring [
 ]
 
 { #category : #tests }
+EpDisabledIntegrationTest >> testDisableDuringWithInteruption [
+	monitor enable.
+	self assert: monitor isEnabled.
+	[ monitor
+		disableDuring: [ self deny: monitor isEnabled.
+			self error ] ]
+		on: Error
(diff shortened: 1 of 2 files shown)


akgrant43 (changes requested):
Hi Cyril,

Shouldn't this be wrapped in a block with ensure: [self enabled: currentStatus]?

Cheers,
Alistair

jecisc:
Good catch! ;)

jecisc:
@akgrant43 Corrected. Thank you.

akgrant43:
Hi Cyril,

Thanks, that looks better.

While waiting for the CI to run (and sorry I didn't think of this earlier): What do you think of adding a class side #disableDuring: that operates on the current instance?

Cheers,
Alistair

jecisc:
Done. I was hesitant to do that while doing the PR. Since you also think it would be good, I did it :)

akgrant43 (approved):
CI failing tests appear to be unrelated (this only adds new methods, so can't really interfere).

Ready for merging...

PR #2770, opened 2019-03-07
2769-add-Slot-example-write-once-slot

### LINKED ISSUE #2769: add Slot example: write once slot
Simple example of a slot that can be writing only once.

MarcusDenker:
fixes  #2769


Ducasse:
Marcus the WriteOnce tests are yellow.

MarcusDenker:
error not related

Issue #2776, opened 2019-03-08
Hitting an undefined class gives a debugger with misleading message "#new was sent to nil"


macta:
When you save a method that references an as yet undefined class - there is now an option to save is as undeclared (the same applies when using the Opal compiler - it can be configured to do this automatically as well).

This is a great enhancement, that lets you code "in the now" without being forced down another path you aren't ready for.

However - when you execute the code and hit that undefined class (perhaps due to something like: MyUndefinedClass new" the debugger shows the messsage "#new was sent to nil".

While this might be technically correct - its not very helpful and I'm sure we can do better. In my mind, this is an undefined class (not the same as nil), and I wonder if it is stored in the bytecode as such (or could be) so that we could present the more informative message "#MyNewClass is Undefined".

Everything else is then perfect, as we have already fixed the debugger so that create will let you create a new class an continue.

So its just the message - and then it really makes TDD awesome! (its still pretty awesome now - in Exercism, when you run a test, you can click Create, to get a new class, continue, then get a debugger and click Create again to define a new method - its very fluid).

Tim

bencoman:
On Fri, 8 Mar 2019 at 17:41, Tim Mackinnon <notifications@github.com> wrote:

> When you save a method that references an as yet undefined class - there
> is now an option to save is as undeclared (the same applies when using the
> Opal compiler - it can be configured to do this automatically as well).
>
> This is a great enhancement, that lets you code "in the now" without being
> forced down another path you aren't ready for.
>
> However - when you execute the code and hit that undefined class (perhaps
> due to something like: MyUndefinedClass new" the debugger shows the
> messsage "#new was sent to nil".
>
> While this might be technically correct - its not very helpful and I'm
> sure we can do better. In my mind, this is an undefined class (not the same
> as nil), and I wonder if it is stored in the bytecode as such (or could be)
>
I don't have the answer, but for some insight, in the Browser for the
containing method, right-click > Extra > Inspect, then review the tabs...
[AST] tab expanded to the leaf node shows...
"RBVariableNode(MyUndefinedClass )"
[Ir] tab shows... "pushLiteralVariable: MyUndefinedClass"
[Bytecode] tab shows... "17 <40> pushLit: MyUndefinedClass"
[Raw] Tab shows... literal1 "#MyUndefinedClass->nil"

So its considered a variable with a value of nil, which makes sense since
"Leave variable undeclared" was selected when the method was saved

cheers -ben

[...]

PR #2783, opened 2019-03-10
Adds a new exception for a missing class error (ClassNotUnderstood) and…

macta:
… handles this in the debugger

This finishes the work that was done in Pharo 6 to support creating a missing class in the debugger. However in fixing #2776 to show a better message, a better solution came about by properly signalling a new ClassNotUnderstood exception, and then using that in the pre debugger to create the class (this latter part seems much cleaner).

I have tested this manually (but written a unit test for the detection of the missing variable node) as this is tricky to automate. You need to save a method with a MissingClass and choose the new save as undeclared menu item. Then executing the method will trigger the code in this change (and let you create a new class in the debugger).

Some questions for the reviewer:

1. Is the message I am using (generated by ClassNoUnderstood) a good one; e.g. the title bar now shows something like: #MissingClassYouEncountered is missing, and does not understand #new
1. #doesNotUnderstand has been overriden in UndefinedObject to support this behaviour (I think this is the cleanest way to do this), but is it safe to do this (I think so) - and is there a better way to not repeat the #dnu signalling code (copied from object and changed to use a different class. I was wondering if that signalling code should be extracted to a new method in object so it can be called from subclasses - but then I'm nervous adding new methods to Object?

### DIFF
--- a/src/DebuggerActions/DoesNotUnderstandDebugAction.class.st
+++ b/src/DebuggerActions/DoesNotUnderstandDebugAction.class.st
@@ -51,22 +51,14 @@ DoesNotUnderstandDebugAction >> closeWindow [
 ]
 
 { #category : #private }
-DoesNotUnderstandDebugAction >> createMissingClassIn: aContext [
-	| senderContext variableNode previousContext errorMsgNode |
-	
-	variableNode := nil.
-	
-	[ senderContext := aContext sender.
-	errorMsgNode := senderContext method sourceNodeExecutedForPC: senderContext pc.
-	variableNode := self findUndeclaredVariableIn: errorMsgNode ]
-		on: Error
-		do: [ ^self ].
+DoesNotUnderstandDebugAction >> createMissingClassWith: variableNode in: aContext [
+	|  previousContext  |
 		
 	OCUndeclaredVariableWarning new
 		node: variableNode;
 		defineClass: variableNode name.
 		
-	previousContext := aContext stack second.
+	previousContext := aContext sender.
 	
 	self closeWindow.
 	
@@ -91,11 +83,13 @@ DoesNotUnderstandDebugAction >> executeAction [
 	MessageNotUnderstood exception. Create a stub for the method that was
 	missing and proceed into it, or create a class if it was missing instead"
 	
-	| msg msgCategory chosenClass |
+	| msg msgCategory chosenClass exception |
 	
 	msg := self interruptedContext tempAt: 1.
-	(msg lookupClass == UndefinedObject ) ifTrue: [ 
-		self createMissingClassIn: self interruptedContext ].
+	exception := self interruptedContext tempAt: 2.
+	
+	(exception class == ClassNotUnderstood) ifTrue: [ 
+		self createMissingClassWith: exception variableNode in: self interruptedContext ].
 	
 	chosenClass := self 
 		askForSuperclassOf: self interruptedContext receiver class
@@ -108,14 +102,6 @@ DoesNotUnderstandDebugAction >> executeAction [
 		inClass: chosenClass 
 		forContext: self interruptedContext.
 	self debugger selectTopContext
-]
-
-{ #category : #private }
-DoesNotUnderstandDebugAction >> findUndeclaredVariableIn: errorMsgNode [
-	^ errorMsgNode allChildren
-		detect: [ :n | n isVariable and: [ n isUndeclared ] ]
-		ifNone: [ errorMsgNode parent allChildren
-				detect: [ :n | n isVariable and: [ n isUndeclared ] ] ]
 ]
 
 { #category : #accessing }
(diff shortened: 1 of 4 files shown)


macta:
If approved, this fixes #2776

Ducasse:
Thanks Tim.
We should check because we are working on UndefinedClasses to be pushed into Pharo80.

macta:
Actually - with UndefinedClasses there may still be some additional work to do - as I notice that when you have one of those, in the debugger - my changes (and those in Pharo 7) don't know how to detect this and so you don't get the option to Re-Create the undefined class - instead it just shows a DNU or the message sent to it. If someone can explain how UndefinedClasses work, we could modify the defaultAction of the pre-debugger to also let you re-create an undefined class or potentially another button to not reference the undefined class anymore.

macta:
BTW - It's very neat how powerful all of this stuff is - I'm frankly amazed (and it takes me back to Lugano when you guys said you were going to get all this stuff visible, so people could flexibly play around with language design)

Ducasse:
Sure. I'm not saying that UndefinedClasses are solving your point. I'm just saying that we should sync with theo and guille because they told me that Undefined class should be pushed soon. 
So we will have probably to do another iteration.

macta:
Ok -  @guillep can you keep me updated on the UndefinedClasses work and when appropriate I can see if I need to update anything on this PR to support them (and I may need some background info to do that. It should be something very tiny if we just want to report errors on them better, and/or allow a create button to bring them back)

Ducasse:
We can also push you PR and let the Undefiner handle it :)

Ducasse:
UndefinedObject >> findUndeclaredVariableIn: ast [
 	"Walk the ast of the current statment and

=> statement

Ducasse:
Failing loading due to  
was 
] Loaded -> Glamour-Tests-Morphic-tonel.1 --- tonel:///builds/workspace/uest_and_branch_Pipeline_PR-2783/32/src --- cache
[32] Loaded -> Glamour-Rubric-Presentations-tonel.1 --- tonel:///builds/workspace/uest_and_branch_Pipeline_PR-2783/32/src --- cache
[32] Loaded -> Glamour-FastTable-tonel.1 --- tonel:///builds/workspace/uest_and_branch_Pipeline_PR-2783/32/src --- cache
[32] Loaded -> Glamour-SpecIntegration-tonel.1 --- tonel:///builds/workspace/uest_and_branch_Pipeline_PR-2783/32/src --- cache
[32] Loaded -> GT-Inspector-tonel.1 --- tonel:///builds/workspace/uest_and_branch_Pipeline_PR-2783/32/src --- cacheError: This package depends on the following classes:
[32]   WindowModel
[32]   ComposableModel
[32] You must resolve these dependencies before you will be able to load these definitions: 
[32]   ComposableModel>>#gtInspectorModelNestingIn:
[32]   ComposableModel>>#gtInspectorPreviewIn:
[32]   WindowModel>>#gtInspectorModelNestingIn:
[32] 
[32] MCPackageLoader(Object)>>notify:

bencoman:
On Sun, 10 Mar 2019 at 19:48, Tim Mackinnon <notifications@github.com>
wrote:

>
>    1. Is the message I am using (generated by ClassNoUnderstood) a good
>    one; e.g. the title bar now shows something like:
>    #MissingClassYouEncountered is missing, and does not understand #new
>
> I haven't dug deeply, but I think you can't distinguish between Classes
and ClassVariables - both start with a capital letter.
So maybe it should be UnknownClassOrVariable.  Also, (nothing I can justify
but...)  I feel "NotUnderstood" should be reserved for messages/methods.

macta:
@bencoman - not sure I understand your comment about the capital letter, if it's in reference to the message  '#MissingClassYouEncountered is missing, and does not understand #new', the #Missing... could any kind of variable (be it class or class variable - and to be clear, it will display the actual name prefixed with a #), so I think the message is ok - but if you chose to send #new to a PasswordID (as in a class variable), then I guess its up to you whether you want to create a class (provided) or go and manually create a new variable like before.

I guess this does raise the interesting thought that - there are really 2 options: Create Class, Define Variable. This PR just offers the former, but technically you could also do the latter, however in practice I think its actually rare (not many things really use class variables).

But I guess we can see if it crops up a lot, and if it does offer the 2nd option. I don't think we can really easily infer it (or to do so it quite a bit of work to understand how the variable is used). So I'd stick with the simpler first case which comes up a lot.

guillep:
Hi, sorry I arrived late to the conversation.

@macta I'll send you a mail with our progress on Undefined classes in the last weeks. We are planning to make a PR soon.

Regarding this PR, I'd have vouched to name the exception `ClassNotFound` instead of `ClassNotUnderstood`. What do you think?
Anyways, it's ok as it is, we can iterate it :)

macta:
@guillep yes, I wasn't sure on naming either - ClassNotUnderstood came from the fact I subclassed MessageNotUnderstood, as its a type of not understood concept - however with your's and ben's feedback it strikes me that VariableNotDeclared (or maybe GlobalNotDeclared) is actually the more intent revealing name. As Ben pointed out - while its typically a missing class, it is possible its a class variable that is missing and so we could ultimately offer some king of create action for that scenario too (but small steps I guess).

I'm happy to change it - I guess I push a change to my branch and then it can also be reviewed and then merged and if we are happy we can then close it?

PR #2787, opened 2019-03-10
2790-standalone-Tool-Registry

### LINKED ISSUE #2786: Improve usability of the minimal image
We need some refactorings to be able to clearly use TelePharo on top of the minimal image

### LINKED ISSUE #2790: standalone Tool-Registry
The images that have no GUI but still have tools like the minimal image with the TelePharo need the ToolRegistry that is now part of Tool-Base package together with other UI dependent code. We need to make it as a standalone package

pavel-krivanek:
ToolRegistry in standalone Tool-Registry package

fixes #2790


Ducasse:
Hi pavel 

this is cool. Now could we avoid 
        Smalltalk tools whenToolRegistered: [ worldState defaultWorldMenu ].
 	Smalltalk tools whenToolUnregistered: [ worldState defaultWorldMenu ].

We should not use Smalltalk and Smalltalk tools directly. 
This is evil.

pavel-krivanek:
I agree but what is the current alternative?

Ducasse:
May be add an instance variable and initialize it so that you have only one place?

noha:
This is super. Fixes a big problem with TelePharo

PR #2796, opened 2019-03-12
2782-Athens-does-not-have-a-way-to-set-and-get-the-fill-rule

akevalion:
Added fill rule primitives and variables to Athens. Check the example
AthensSurfaceExamples class >> #drawSetFillRule


MarcusDenker:
This seems PR seems to contain a lot of changes that it should not

MarcusDenker:
a newer PR has been submitted

PR #2804, opened 2019-03-12
2803-parseErrorNode-should-fill-value-for-the-case-of-expecting-a-

### LINKED ISSUE #2803: parseErrorNode: should fill value for the case of expecting a )

MarcusDenker:
fixes #2803


MarcusDenker:
error not related

PR #2807, opened 2019-03-13
2806-test-and-fix-for-PropertySlot-remove-propertySlot-when-instances-are-present

MarcusDenker:
add test and fix that allows removal of propertyslots when instances are present.


MarcusDenker:
error not related

PR #2812, opened 2019-03-13
2811-Test-for-metalink-on-slot-that-gets-assigned-

### LINKED ISSUE #2811: Test for metalink on slot that gets assigned.
add a test for metalling on a slot that gets assigned to. It should get triggered

MarcusDenker:
fixes #2811


MarcusDenker:
failing test not related

PR #2825, opened 2019-03-14
2824-unify-and-cleanup-compactAll-OrderedCollection

MarcusDenker:
OrderedCollection had #compactInstances: add #compactAll and call it in cleanUp: just like HashedCollection
remove Behaviour >>compress (not called, can be called manually if needed (methodDict rehash_
#cleanUpForRelease
-- remove explicit call to #compactInstances, as this is now done with the #cleanup
-- add a final #rehash of all HashedCollections at the end


MarcusDenker:
errors not related

Ducasse:
Looks good to me.

PR #2827, opened 2019-03-14
2826-remove-ReDeadMethod-and-Test

### LINKED ISSUE #2826: remove ReDeadMethod and Test
ReDeadMethod is not really good: the idea was to warn specially if methods that are unsent do themselves send selectors that do not exist.
In the end not that good and the current version does not even implement it

MarcusDenker:
fixes #2826


MarcusDenker:
error not related

PR #2830, opened 2019-03-14
2829-RPackagepackageManifestOrNil-should-be-directly-in-RPackage-Core

### LINKED ISSUE #2829: RPackage>>#packageManifestOrNil should be directly in RPackage-Core
RPackage>>#packageManifestOrNil should be directly in RPackage-Core because it does not make any new dependency on the Manifest builder and it will enable to use TelePharo on the minimal image without Manifest-Core package

pavel-krivanek:
move ownership of packageManifestOrNil

fixes #2829


MarcusDenker (approved):
looks good

MarcusDenker:
failing tests not related

PR #2837, opened 2019-03-15
2836-removeUnreferencedKeys-is-only-there-for-the-Undeclared-mechanism-no-need-to-implement-on-other-Disctionaries

### LINKED ISSUE #2836: #removeUnreferencedKeys is only there for the Undeclared mechanism, no need to implement on other Disctionaries
removeUnreferencedKeys is a private method that just makes sense for #cleanOutUndeclared, as such we only need it on Dictionary.

MarcusDenker:
removeUnreferencedKeys is a private method that just makes sense for #cleanOutUndeclared, as such we only need it on Dictionary.

fixes #2836


MarcusDenker:
when looking at the diff:

BIN -190 Bytes (99%) src/Collections-Unordered/Dictionary.class.st
Binary file not shown.

But the merge seems to work fine

PR #2840, opened 2019-03-15
2839-Clean-AthensTest

Ducasse:
Renamed the class to convey that it is not a TestCase subclass


Ducasse:
Broken tests are unrelated

PR #2848, opened 2019-03-16
2753-betterFileVersionExtraction

dionisiydk:
It replaces regex parsing of filename pattern (basename.2.ext) with helper method


dionisiydk:
It improves #2753

akgrant43 (approved):
Thanks, Dennis!

For a discussion of why this change is necessary, see #2754.

CI test failures are unrelated.

Approved.

PR #2857, opened 2019-03-17
2852-CodeCompletion-should-not-be-managed-via-common-tools-but-via-a-TextArea-registration

### LINKED ISSUE #2852: CodeCompletion should not be managed via common tools but via a TextArea registration
May be it should be via the text editor and not the textArea. But since the area is actually handling the keystroke logic it should also provide a way to have access to the completion engine.

Ducasse:
Fix #2852 adding registration and unregistration to PharoCommonTools!
next step 
TextMorph and removing transcript traces.


Ducasse:
Broken tests are unrelated.

Ducasse:
Need some other reviewer....

Ducasse:
Alain did a code review and this is ok so I push it.

PR #2875, opened 2019-03-18
2847-Implement-method-to-remove-class-side-slot

### LINKED ISSUE #2847: Implement method to remove class side slot
From fogbugz: https://pharo.fogbugz.com/f/cases/22411/Implement-method-to-remove-class-side-slot

We can add slots to instance and class side of classes using "aClass addSlot: aSlot". But it does not work for slot removal. "aClass classSide removeSlot: aSlot" returns DNU because method #removeSlot: is defined in Class.
It should be moved to ClassDescription, close to #addSlot:

--

I renamed title because I found that addSlot: is just subclass responsibility in ClassDescription.

To add class side slots there is instance side method addClassSlot: (defined in Class).
But there is no removeClassSlot:. 
So we should implement missing methods.

MarcusDenker:
fixes #2847


Ducasse:
change looks ok

MarcusDenker:
failing tests are not related

PR #2878, opened 2019-03-18
2877-whichSelectorsAccess-and-slots

### LINKED ISSUE #2877: whichSelectorsAccess: and slots
whichSelectorsAccess: should use readsSlot: and writesSlot:

This will be slow, but we can speed that up by checking for  the special case of InstanceVariableSlot in readsSlot:  and writesSlot: (to be done later)

MarcusDenker:
fixes #2877

### DIFF
--- a/src/Kernel/ClassDescription.class.st
+++ b/src/Kernel/ClassDescription.class.st
@@ -1370,11 +1370,12 @@ ClassDescription >> whichCategoryIncludesSelector: aSelector [
 ]
 
 { #category : #'testing method dictionary' }
-ClassDescription >> whichSelectorsAccess: instVarName [ 
+ClassDescription >> whichSelectorsAccess: instVarName [
 	"Answer a set of selectors whose methods access the argument, 
 	instVarName, as a named instance variable."
 
 	| slot |
+	(self hasSlotNamed: instVarName) ifFalse: [ ^#() ].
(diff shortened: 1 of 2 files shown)


MarcusDenker:
failing test no related

src/Kernel/ClassDescription.class.st
```diff
@@ -1369,6 +1369,19 @@ ClassDescription >> whichCategoryIncludesSelector: aSelector [
...
+ClassDescription >> whichSelectorsAccess: instVarName [
+	"Answer a set of selectors whose methods access the argument, 
+	instVarName, as a named instance variable."
+
+	| slot |
+	(self hasSlotNamed: instVarName) ifFalse: [ ^#() ].
+	slot := self slotNamed: instVarName.
+	^ self selectors select:  [:sel | | method |
+		method := self compiledMethodAt: sel.
+		(method readsSlot: slot) or: [method writesSlot: slot]]
```
dionisiydk, line 1382: We already have polymorphic optimized methods in slot classes: #isReadIn: and #isWrittenIn: . #readsSlot is general method which is used by Slot class implementation. And readsField: is used by InstanceVariableSlot implementation.
So here method should use:
```Smalltalk
(slot isReadsIn: method) or: [slot isWrittenIn: method]
```
And it would be nice to add isReferencedIn: method for such cases.

PR #2888, opened 2019-03-18
UndefinedObject should not implement at:put: (or inherit it). This is…

### LINKED ISSUE #2795: sending at:put: to an undefined object (nil) give confusing error message
If you forget to initialise a variable you thought was a Dictionary - you get a confusing error message: "Error: only integers should be used as indices"

This is a consequence of having at:put: defined on Object (which is a bit nasty)

Should UndefinedObject at least override this and signal something a bit more obvious? This is a common easily done thing and we don't support it very well.

I just did it in some code, and was scratching my head initially until I read the stack properly and realised it was something much simpler than I thought I had done.

macta:
… the minimal fix

Lots of discussion on the ML about this - but no real decisions. The bare minimum is that nil should throw an error, but a future fix could be to remove at:put: from object.

fixes #2795

### DIFF
--- a/src/Kernel/UndefinedObject.class.st
+++ b/src/Kernel/UndefinedObject.class.st
@@ -50,6 +50,11 @@ UndefinedObject >> asSetElement [
 	^ SetElement withNil
 ]
 
+{ #category : #accessing }
+UndefinedObject >> at: index put: anObject [ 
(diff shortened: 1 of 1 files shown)


MarcusDenker (approved):
ok

PR #2893, opened 2019-03-18
2776-Hitting-an-undefined-class-gives-a-debugger-with-misleading-message-new-was-sent-to-nil-2

macta:
This re-applies the fix, which re-works what was done in Pharo 6 to let the debugger create a missing class, the code is cleaner and crucially it gives a better title in the debugger that you've hit a missing Class/Variable.

see: https://github.com/pharo-project/pharo/pull/2783

This applies the feedback from @guillep to use a better name for the exception raised to identify a missing variable/class that the debugger can then detec.


macta:
This fixes #2776

MarcusDenker (approved):
looks good

PR #2897, opened 2019-03-19
2880-speedup-writesSlot-and-readsSlot-for-the-case-of-Ivar

MarcusDenker:
speed up #readsSlot and #writesSlot: by delegating to the Slot. The InstanceVariableSlot as an optimzied implementation

### DIFF
--- a/src/Slot-Core/Slot.class.st
+++ b/src/Slot-Core/Slot.class.st
@@ -229,7 +229,12 @@ Slot >> isLocal [
 
 { #category : #testing }
 Slot >> isReadIn: aCompiledCode [
-	^aCompiledCode readsSlot: self
+	| nodes |
+	nodes := aCompiledCode ast allChildren.
+	nodes := nodes select: #isVariable.
+	nodes := nodes select: #isInstance.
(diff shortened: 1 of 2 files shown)


MarcusDenker:
fixes #2897

MarcusDenker:
error not related

src/Slot-Core/Slot.class.st
```diff
@@ -256,7 +257,8 @@ Slot >> isVisible [
 
 { #category : #testing }
 Slot >> isWrittenIn: aCompiledCode [
-	^aCompiledCode writesSlot: self
+	^aCompiledCode ast instanceVariableWriteNodes
+		anySatisfy: [ :node | node binding slot == self ]
```
dionisiydk, line 261: Maybe we can introduce two methods for AST-nodes:
- aNode readsSlot: aSlot
- aNode writesSlot: aSlot

It will do recursion over children as default implementation on RBProgramNode. 
RBVariableNode will do "binding slot == aSlot"
RBAssignmentNode will provide special implementation for writesSlot:

And I think it would be pretty useful methods

PR #2905, opened 2019-03-19
2901-ClassBuilder-fillFor-should-not-use-classVariablesString

### LINKED ISSUE #2901: ClassBuilder: #fillFor: should not use #classVariablesString
#fillFor: should not use #classVariablesString.. it seems we can just copy the object over, no need to go via string conversion

MarcusDenker:
fixes #2901


MarcusDenker:
errors not related

Issue #2880, opened 2019-03-18
speedup writesSlot: and readsSlot: for the case of Ivar


MarcusDenker:
writesSlot: and readsSlot: are really slow as they are a query on the AST (which might even not be there yet).

For the case of ivars, we can make it much faster by fallen back on writesField: and readsField:

dionisiydk:
Hi @MarcusDenker. 
I commented on closed issue. But let's continue here:
We already have polymorphic methods in slot classes: #isReadIn: and #isWrittenIn: . They delegate logic to method:
- #readsSlot: is used by main Slot class implementation.
- #readsField: is used by InstanceVariableSlot implementation. 

So with current design it is supposed to ask a slot about usage instead of method.
Would be nice to have both ways available. But it would require better names to distinguish primitive API from high level API (readsField: and readsSlot: are now both primitive API)

dionisiydk:
Maybe current #readSlot: code should be inlined into Slot class. Then method #readSlot: can become general by delegation to given slot.

Also readSlot: code with ast analysis deserves love. Too many message chains. We should ask nodes to do the job.

MarcusDenker:
I think if we move the code of readSlot: to isReadIn: it will do the speedup: isReadIn: is overridden for indexed vars to use the fast code.

yes, the queries are ugly.

MarcusDenker:
https://github.com/pharo-project/pharo/pull/2897

MarcusDenker:
I added https://github.com/pharo-project/pharo/issues/2898 for the simplification

MarcusDenker:
I merged this version. More to come in another Issue

PR #2919, opened 2019-03-19
2918-accessesSlot: should use #isAccessedIn: method

### LINKED ISSUE #2918: accessesSlot: should use #isAccessedIn: method
continue the changes from #2880

- add #isAccessedIn: for Slot (with fast version for instance vars)
- use in accessesSlot:
- use in whichSelectorsAccess:
- whichSelectorsRead: and whichSelectorsStoreInto: changed to call the tests on the slot

Now we need to unify with the isReferencedIn: that is used on Globals and in Calypso

MarcusDenker:
fixes #2918

continue the changes from #2880

add #isAccessedIn: for Slot (with fast version for instance vars)
use in accessesSlot:
use in whichSelectorsAccess:
whichSelectorsRead: and whichSelectorsStoreInto: changed to call the tests on the slot
Now we need to unify with the isReferencedIn: that is used on Globals and in Calypso

### DIFF
--- a/src/AST-Core/RBProgramNode.class.st
+++ b/src/AST-Core/RBProgramNode.class.st
@@ -357,6 +357,11 @@ RBProgramNode >> hashForCollection: aCollection [
 ]
 
 { #category : #querying }
+RBProgramNode >> instanceVariableNodes [
+		^self variableNodes select: [:each | each isInstance]
+]
+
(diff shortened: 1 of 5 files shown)


MarcusDenker:
arg, wrongly name PR...

PR #2923, opened 2019-03-19
2690-22595-Median-is-not-calculated-correctly-for-even-sizes

Ducasse:
Fix: #2690 Added some tests. Changed logic about testing if elements are integer. We can only compute a mean on objects that can be added.


MarcusDenker (approved):
ok

Issue #2782, opened 2019-03-10
Athens does not have a way to set and get the fill rule


akevalion:
Steps to reproduce the issue
1 First read this https://www.cairographics.org/manual/cairo-cairo-t.html#cairo-set-fill-rule.
2 Download pharo7 with PharoLauncher
3 execute the next piece of code that creates a donut in athens
```Smalltalk
| surface |
surface := AthensCairoSurface extent: 300@300.
surface drawDuring: [ :canvas | | path |
	canvas pathTransform
		translateBy: 50@50.
	"donut"
	"canvas setFillRuleEvenOdd."
	path := canvas createPath: [ :builder |  
		builder
			absolute;
			moveTo: 100 @0;
			arcCenterX: 0
				centerY: 0
				radius: 50
				startAngle: 0 
				endAngle: Float pi * 2;
			moveTo: 75@0;
			arcCenterX: 0
				centerY: 0
				radius: 30
				startAngle: 0
				endAngle: Float pi * 2.
			 ].
	canvas setPaint: Color blue.
	canvas drawShape: path.
	 ].
surface asForm asMorph openInWindow.
```
4 The result is:
<img src="https://user-images.githubusercontent.com/10532890/54080348-6bbfbc00-42cc-11e9-9404-7d77c4db8cd0.png" width="300"/>

Expected Result: 
<img src="https://user-images.githubusercontent.com/10532890/54080356-8a25b780-42cc-11e9-9da1-53acaa19a7e4.png" width="300"/>

Athens needs new methods and variables for the following classes
1.
```Smalltalk
AthensCairoPathBuilder >> setFillRuleEvenOdd
	context setFillRuleEvenOdd
```
2.
```Smalltalk
AtthensCairoCanvas >> setFillRuleEvenOdd
	self setFillRule: CAIRO_FILL_RULE_EVEN_ODD
```
3.
```Smalltalk
AtthensCairoCanvas >> setFillRule: aCairoFillRule
	^ self ffiCall: #( void cairo_set_fill_rule ( self, cairo_fill_rule_t aCairoFillRule ) )
```
4.
```Smalltalk
AthensCairoDefinitions class >> initialize_cairo_fill_rule_t
	CAIRO_FILL_RULE_WINDING := 0.
	CAIRO_FILL_RULE_EVEN_ODD := 1.
```
5.
```Smalltalk
AthensCairoDefinitions class>> initialize_types
[...]

PR #2924, opened 2019-03-19
2782-Athens-does-not-have-a-way-to-set-and-get-the-fill-rule

akevalion:
Fix: #2782
Added to athens: set fill rule and new example


MarcusDenker:
failing tests not related

PR #2926, opened 2019-03-20
2925-Integrate-latest-version-of-Hiedra

tinchodias:
Last weeks I have invested time in Hiedra (source at: https://github.com/tinchodias/hiedra), that is already in Pharo since 6.0 but needed some extra love. 

* Adds support for FastTable, and it's optimized for it (keeps support for Spec's TreePresenter).
* Better tests: more coverage and more understandable.
* Added examples of use.
* Improved comments in methods and classes.
* More understandable code.

API changed so this commit includes the upgrade of the only user: Epicea.


Ducasse:
Thanks a **LOT** martin

PR #2950, opened 2019-03-21
2864-CompiledMethodcommentsIn-should-be-enhanced

### LINKED ISSUE #2864: CompiledMethod>>#commentsIn: should be enhanced
The source code is quite hard to understand and I think it should be possible to do it directly using a regular expression. Or maybe even directly using system's parser.

MarcusDenker:
fixes #2864

### DIFF
--- a/src/AST-Core/CompiledMethod.extension.st
+++ b/src/AST-Core/CompiledMethod.extension.st
@@ -7,6 +7,20 @@ CompiledMethod >> ast [
 ]
 
 { #category : #'*AST-Core' }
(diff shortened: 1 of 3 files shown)


MarcusDenker:
The PR changes #comments to use the AST instead of doing the parsing by hand.

Reasons why this seems to be good:

1) the parsing by hand was incomplete
2) I can not think of any use case where one needs to do this without having a compiler installed
3) speed should not be an issue for any user. It will either be interactive (reads single methods) or batch mode (e.g. when analysing comments).

PR #2956, opened 2019-03-21
2955-speedup-isAccessedIn-in-BooleanSlot-and-PropertySlot

### LINKED ISSUE #2955: speedup isAccessedIn: in BooleanSlot and PropertySlot
we can check if the base slot is there. if not ---> we know we are not

MarcusDenker:
fixes #2955

### DIFF
--- a/src/Slot-Examples/BooleanSlot.class.st
+++ b/src/Slot-Examples/BooleanSlot.class.st
@@ -76,6 +76,24 @@ BooleanSlot >> installingIn: aClass [
 	offset := self calculateOffset: aClass.
 ]
 
+{ #category : #testing }
+BooleanSlot >> isAccessedIn: aMethod [
+	(baseSlot isAccessedIn: aMethod) ifFalse: [ ^false ].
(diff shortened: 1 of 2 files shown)


src/Slot-Examples/BooleanSlot.class.st
```diff
@@ -76,6 +76,24 @@ BooleanSlot >> installingIn: aClass [
 	offset := self calculateOffset: aClass.
 ]
 
+{ #category : #testing }
+BooleanSlot >> isAccessedIn: aMethod [
+	(baseSlot isAccessedIn: aMethod) ifFalse: [ ^false ].
+	^super isAccessedIn: aMethod
```
dionisiydk, line 82: I see it is same methods in PropertySlot.
What about extracting common superclass? like DerivedSlot?

MarcusDenker: Yes, I was thinking about this... I will add a note to my TODO

PR #2957, opened 2019-03-21
2922-the-class-builder-talks-about-sharedVariables-while-every-else-they-are-called-Class-Variables

### LINKED ISSUE #2922: the class builder talks about sharedVariables, while every else they are called Class Variables
this is very confusing. 

--> rename the ivar and accessor to classVariable

MarcusDenker:
fixes #2922

### DIFF
--- a/src/Hermes/HEInstaller.class.st
+++ b/src/Hermes/HEInstaller.class.st
@@ -67,7 +67,7 @@ HEInstaller >> build: aHEClass [
 		superclass: superclass;
 		layoutClass: layoutType;
 		slots: (self asSlots: aHEClass instanceVariables);
-		sharedVariables:  (self asClassVariables:aHEClass classVariables);
+		classVariables:  (self asClassVariables:aHEClass classVariables);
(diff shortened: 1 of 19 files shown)


Ducasse:
So that we get fresher feedabck

guillep (changes requested):
This change impacts the bootstrap. The bootstrap uses the class builder to create classes, and during bootstrap interpretation automatic deprecations are not handled by the AST interpreter.
Maybe @carolahp can help @MarcusDenker with this? It should not be difficult to fix.

Also @carolahp, this is a clear example of bootstrap bug :)

Ducasse:
I review it and it looks ok now we should check the bootstrap expert

MarcusDenker:
I will let run the tests again, because it might have been using exactly the broken image of last weekend.

guillep:
Hi, this PR requires changes in the bootstrap (I am not guessing, I checked the logs and I know the process). Closing and Reopening will not solve it automatically.

Issue #2967, opened 2019-03-21
[ICEBERG]The merge preview is not well sized to click on the button


VincentBlondeau:
Sorry about opening bugs reports but....
![image](https://user-images.githubusercontent.com/4706329/54777835-7acf3400-4c13-11e9-90af-0f3697927460.png)

VincentBlondeau:
Fixed

bouraqadi:
This issue does still exist in Pharo 
![icebergMergeButtonHidden](https://user-images.githubusercontent.com/435343/154310048-ee9f5c06-1dc1-48a1-ada1-a2328292f611.jpg)

VincentBlondeau:
indeed: the button should stick to the right side:
<img width="362" alt="CleanShot 2022-03-03 at 09 44 17@2x" src="https://user-images.githubusercontent.com/4706329/156528795-faee82fb-09f8-4a5c-9ecd-1d325a956219.png">

MarcusDenker:
This was fixed

Issue #2966, opened 2019-03-21
[Iceberg]Is there an option for rebasing?


VincentBlondeau:
When the Pharo8.0 branch is ahead of you and you want to catch the train. Nice feature to have to keep up with the fast progresses that Pharo is making

VincentBlondeau:
Not fixed nor integrated in the image

guillep:
I think this is an issue for the iceberg issue tracker, and more than an issue is a feature request :).
Let's keep the discussion in Iceberg's issue tracker

VincentBlondeau:
Then it is reopen in https://github.com/pharo-vcs/iceberg/issues/1361

Issue #2965, opened 2019-03-21
[ICEBERG]Message when openning an branch for an issue should be improved


VincentBlondeau:
The pic talks by itself
![image](https://user-images.githubusercontent.com/4706329/54777226-25465780-4c12-11e9-9aae-afe99157dd03.png)

jecisc:
Are you on the right remote?

MarcusDenker:
the "remote" field is a drop down (not visible). You need to select the pharo repository there.

Yes, that needs to be simplified and improved.

estebanlm:
yes, I don't know why you do not see the dropdown, but this is not an error :)
and @MarcusDenker, I don't know how this could be simplified more (pick remote, type issue number) other than querying for the open issues, which would take time.

VincentBlondeau:
Maybe adding a message that says: "you may want to check the other remotes?"

astares:
What about using "INVALID-ISSUE-FOR-REMOTE" instead of "INVALID_ISSUE" so one could at least have an idea it might be due to the selected remote...

VincentBlondeau:
Good idea! 
Can the issue be reopenned? please?

MarcusDenker:
Just got a mail from another new commuter (in german) who did not know what to do at this step. It is not clear that this is a drop=down menu.

guillep:
Moved to https://github.com/pharo-vcs/iceberg/issues/1364

Issue #2964, opened 2019-03-21
[Tools][Rubric]Cannot open the Profiler....


VincentBlondeau:
I just missed a button and clicked unfortunately on the profiler menu item. And guess what: I received an exception (that I cannot fuel out because it not working too: See #2963).
![ur5Yh7UPhu](https://user-images.githubusercontent.com/4706329/54776280-0e066a80-4c10-11e9-9114-9eee66d3dd54.gif)

StephanEggermont:
Seems connected to #2980. There is probably no test for the missing styler though

StephanEggermont:
Caused by #2789

plantec:
I've dug a little bit. How a styles is plugged is different from how it is implemented in PluggableTextMorph. Rubric uses EditingMode and to edit smalltalk from a Rubric morph one to send #beForSmalltalkCode to it. 
ok, it works but after a new problem arise with the #interactionModel that was introduced when we wanted to improve Nautilus.
The question is: does #interactionModel still make sense  in Pharo 8 ?

StephanEggermont:
Are there other users that need it? Would it be difficult to modify Spec?

MarcusDenker:
this is now fixed: one can open it

PR #2969, opened 2019-03-21
Rewrite with TDD the updateTasks to avoid race conditions

VincentBlondeau:
BTW: we are not able to use the commit function integrated in Pharo8.0 correctly (we are probably to dumb for this, we will need a tutorial... jk). So we decided with Stephan Eggermont to file out the modifications we made see in the Fileout folder ;) 

PS: You could maybe try to load it in another image to successfully integrate our changes. Though, you will need to change the right baseline to add the new package we created for morphic tests


VincentBlondeau:
Sorry for the cr, I use windows

jecisc:
Line endings are crlf instead of lf.

StephanEggermont:
the change to updateTasks is missing

StephanEggermont:
See Issue #2974

Ducasse:
Guys I have no idea what you are talking about. 
What is the updateTasks?

Stef

Ducasse:
Can you explain what did not work exactly?
I use iceberg daily and it works. So I would like to understand.

StephanEggermont:
We started with a Pharo 8 branch where we had already made a fix and a PR. We only noticed that when trying to commit, and then tried to fix the situation. #2953 #2954 #2961 #2963 #2964 #2965 #2966 #2967 and then we were confused and just did a fileout. Somehow it also misses the last change to actually use the new updateTask form which is found in the test code.

We noticed there might be race conditions in CI with the taskbar, and tried to find out where they might be coming from. The stack trace lead us to updateTasks in TaskbarMorph, which we found difficult to understand. We decided to rewrite it TDD

StephanEggermont:
Keep open

VincentBlondeau:
ping

VincentBlondeau:
test failure not related:
windows-32 / Tests-windows-32 / testisZipArchive – Windows32.Compression.Tests.Archives.ZipArchiveTest

VincentBlondeau:
ping

MarcusDenker (approved):
ok

Issue #2972, opened 2019-03-22
MailAddressTokenizer doesn't handle non-ascii characters


akgrant43:
RFC 6531 allows mail addresses to include characters outside the normal ascii character set.

The soon to be attached PR extends MailAddressTokenizer to handle these characters (codePoint > 127) and adds automated tests.

akgrant43:
PR: #2973

PR #2973, opened 2019-03-22
2972-MailAddressTokenizer-doesnt-handle-non-ascii-characters

akgrant43:
RFC 6531 allows mail addresses to include characters outside the normal ascii character set.

Extend MailAddressTokenizer to handle these characters (codePoint > 127) and add automated tests.

Thanks to Levente Uzonyi for identifying the issue, appropriate RFC and changes to CSNonAtom in Squeak.

Fixes: #2972


akgrant43:
CI crashed, but:

- The VM crash and failed tests are all pre-existing.
- This change should be platform independent.
- It is in a class that doesn't get much use.

So it should be safe to merge.

VincentBlondeau (approved):

akgrant43:
Thanks Marcus & Vincent!

PR #2988, opened 2019-03-22
2987-provide-standard-class-comment-for-tests

### LINKED ISSUE #2987: provide standard class comment for tests
Tests without comments are now shown with a ! in the Browser

We now add lots of comments of the kind 

"this is a test class".

==> add this comment as the default comment in TestCase class

MarcusDenker:
fixes #2987


MarcusDenker:
we really need to get better at green test. This PR makes two tests fail and we merged it because the tests are broken anyway and nobody checks anymore

PR #2999, opened 2019-03-25
feature/sunit-test-pragma

zecke:
Allow to mark a test with a pragma

Make it possible to mark a test with a pragma and get it treated as test. This allows to name a test as "selector_state_expectedOutcome" as proposed  in some style guides.

### DIFF
--- a/src/Ring-Deprecated-Core-Kernel/RGMethodDefinition.class.st
+++ b/src/Ring-Deprecated-Core-Kernel/RGMethodDefinition.class.st
@@ -446,7 +446,7 @@ RGMethodDefinition >> isTestMethod [
 		ifFalse: [ ^ false ].
 	selectorString := self selector asString.
 	"unary selectors starting with 'should' are supposed to be treated as test methods too"
-	((selectorString beginsWith: 'test') or: [ selectorString beginsWith: 'should' ])
+	(((selectorString beginsWith: 'test') or: [ selectorString beginsWith: 'should' ]) or: [ (self pragmaAt: #test) isNotNil ])
 		ifFalse: [ ^ false ].
 	"Is the receiver a TestCase test method?"
 	self methodClass isTestCase
--- a/src/SUnit-Core/CompiledMethod.extension.st
+++ b/src/SUnit-Core/CompiledMethod.extension.st
@@ -30,7 +30,7 @@ CompiledMethod >> isTestMethod [
 	self numArgs isZero
 		ifFalse: [ ^ false ].
 	"unary selectors starting with 'should' are supposed to be treated as test methods too"
-	((self selector beginsWith: 'test') or: [ self selector beginsWith: 'should' ])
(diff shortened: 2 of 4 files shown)


MarcusDenker:
I like the idea

Ducasse:
I would like to see how we can properly extend SUnit because the more options we add the more burden for SUnit maintainer. We are about to clean SUnit and to check the exact semantics of <skip> self skip.... and others. So I would not add yet another way to declaring test. To me it looks cosmetic. And it means that the other tools will have to be checked and updated.

Ducasse:
And why 

testIsTrueShouldReturnTrue is not sufficient?

juliendelplanque:
Hello,

I am one of the people that will do what @Ducasse talks about.

We have the plan to create a SUnit visitor which will provide an API for visiting tests in a package (we already have a prototype, but we need to be careful when modifying SUnit because it is one of the most important framework of Pharo).
Seeing what you had to do in order to integrate your proposal, it is clear that duplicated code to visit tests is bad...

If we want your proposal to get integrated, I propose to do it after the integration of SUnit visitor (soon, maybe I'll do it during Pharo Days hacking sessions next week). With the support for visitor, we will be able to create a new visitor that takes into account your feature and it will be way cleaner than spreading the way to decide if a method is a test or not across the image.

So, right now, I'd prefer to reject this PR.

Cheers,

Julien

dionisiydk:
Hello.

The question is do we want to define tests using pragma? If yes then why not integrate it? We can always iterate to improve code and fix code duplication.

BWT there is duplication of isTestMethod logic in Calypso (there is PR from @zecke). I did it to support tests from traits: you can run test for each users when testcase trait is selected (try it for example with TEmptySequenceableTest). Have it in mind too: traits can supply tests.

Ducasse:
For me I do not want to have test with pragmas because it is not needed. 
I do not see the value besides making things more complex to handle. 
This is already the mess with self skip and <skip> not event sure that it is working. 
Let us keep the system simple.

astares:
Java also supports unit tests by annotating a method so one does not have to stick to the initial unit testing rule to provide a method with testXXX signature pattern. 

On one side the pragmas allow to be more free on the name of the test method. Also one could deactivate a test by removing the pragma but keeping the method name intact. 

But in bigger projects and large unit classes with many methods (utilities and test methods mixed)
not using testXXX a pragma can quickly lead to confusion as it disrupts the tests / methods. 

Using an additional pragma would mean one can not distinguish the test methods directly from the method signature. If not indicated by an icon one always have to look at the method body to check for the pragma. 

Also using regular textXXX forces people more to think about having a good wording for the "XXX" part in conjunction with the "test" prefix. So the "test" in the selector should already be very intention revealing on what to test which is often helpful.

Otherwise we might quickly end up with:

    foo
       <test>
       self assert: ...

methods.

I'm with Stef here - the simpler the better. I never understood what the additional complexity of annotations in JUnit really brings.

But maybe more people can comment on their preferences.

zecke:
@estebanlm Is there a design process for Pharo8 features?

PR #3008, opened 2019-03-25
Adding <debuggerCompleteToSender> pragma to some assert methods so th…

### LINKED ISSUE #3006: When a test fails, open the debugger on the test instead of on Object>>#assert:
When executing a single test, if it fails, the debugger opens on the Object>>#assert: method.
I think it would be more useful if the debugger opened directly on the failing test method.

Kind of like how the debugger opens on the method that sent #halt, instead of the Object>>#halt method.

To do this, we can add the pragma <debuggerCompleteToSender> to the relevant assert methods.

Images:
Currently, the debugger opens like this:
image
With the pragma, the debugger opens directly on my failing test method:
image

dupriezt:
…at the debugger opens directly on the failing test method when a test fails.

Fixes #3006

### DIFF
--- a/src/Kernel/Object.class.st
+++ b/src/Kernel/Object.class.st
@@ -365,6 +365,7 @@ Object >> asString [
 Object >> assert: aBlock [
 	"Throw an assertion error if aBlock does not evaluates to true.
 	We check for true explicitly to make the assertion fail for non booleans"
+	<debuggerCompleteToSender>
 	self assert: aBlock description: 'Assertion failed'.
 ]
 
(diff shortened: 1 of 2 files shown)


MarcusDenker:
I will reopen to run with the latest tests...

akgrant43 (changes requested):
Hi @dupriezt 

I like this change, but there are multiple methods where it also should be added (actually I'm not sure how this ever worked for Object>>assert: since #assert: calls #assert:description:, which is missing the pragma, thus the debugger opens in #assert:description:):

- Object>>assert:description:
- All the methods in TestAsserter asserting protocol

Cheers,
Alistair

akevalion:
I have added the the pragma <debuggerCompleteToSender> check the PR https://github.com/pharo-project/pharo/pull/3108

MarcusDenker:
doing things in parallel is often quite confusing. It is better to review the other and push to have it integrated *or* create one shared PR.

dupriezt:
Closing this since PR #3116 encompasses its changes.

PR #3015, opened 2019-03-26
3013-RegressionDropListMorph-content-is-now-selectable

### LINKED ISSUE #3013: Regression: DropListMorph content is now selectable
With recent Rubric changes, DropListMorph changed to use rubric instead of pluggable text morph. This causes a regression because now the text is selectable. 

This cause multiple problems such as it is now harder to open the list or RSOD if you select the text and change for a shorter option.

jecisc:
DropListMorph text area should be locked so it is not selectable. 

Fixes #3013


Ducasse:
I could have review it but now I'm dropping out from the integrator team for lack of common vision.
Thanks for your time. Ask esteban to review and integrate PR.

Ducasse:
Tx a lot cyril!

PR #3020, opened 2019-03-26
3019-GT-EventRecorder-is-not-used-anymore

### LINKED ISSUE #3019: GT-EventRecorder is not used anymore
I propose to remove it since it is dead code.

jecisc:
Remove GT-EventRecorder.

Fixes #3019


astares:
Not used anymore is relative - it means you do not use it.

I would propose to deal with it like with other places: if we want to get rid of something we should move it to Deprecatedxx packages to help people (who might potentially used it in the past) can have the chance to think about transitions. 

So we should not REMOVE but DEPRECATE first and then REMOVE. No?

pavel-krivanek:
There is a border where you need to remove. Else you would have the deprecated Nautilus in Pharo 7, for example. Most of the old GT code is supposed to be removed from Pharo

Ducasse:
GT recorder is not really related to Glamour. 
I will ask benoit if this is GTrecorder that he uses for Smart Test.
It would be nice to package it so that people can use.

Ducasse:
I would remove the dependency to glamour or gt such as the settings and keep it.

Ducasse:
I checked with Benoit and this is a nice package to collect information about the system. 
I think that the first action is to decouple it from Glamour and to add its own baseline. 
I would keep it inside Pharo once cleaned. Later we can version it in as a separate project.

Issue #3009, opened 2019-03-25
CharacterSetComplement size causes GT inspector to crash


akgrant43:
CharacterSetComplement>>size is currently implemented as #shouldNotImplement, which causes GT inspector to crash (amongst other problems).

Given that files can already be larger than the maximum character value (2 ** 30 - 1) it doesn't make sense to artificially restrict iteration over the character set.

Soon to be attached PR:

- Updates CharacterSetComplement>>size to return the correct value.
- Reimplements CharacterSetComplement>>do: to do the expected thing.
- Adds CharacterSetComplement>>gtInspectorItemsIn: to view the absent list (which will be a manageable size)
- Updates the class and package comments to provide more information

akgrant43:
PR: #3028

PR #3028, opened 2019-03-27
3009-CharacterSetComplement-size-causes-GT-inspector-to-crash

akgrant43:
CharacterSetComplement>>size is currently implemented as #shouldNotImplement, which causes GT inspector to crash (amongst other problems).

Given that files can already be larger than the maximum character value (2 ** 30 - 1) it doesn't make sense to artificially restrict iteration over the character set.

This commit:

- Updates CharacterSetComplement>>size to return the correct value.
- Reimplements CharacterSetComplement>>do: to do the expected thing.
- Adds CharacterSetComplement>>gtInspectorItemsIn: to view the absent list (which will be a manageable size)
- Updates the class and package comments to provide more information

Fixes: #3009

### DIFF
--- a/src/Collections-Support/CharacterSet.class.st
+++ b/src/Collections-Support/CharacterSet.class.st
@@ -1,5 +1,11 @@
 "
-A set of characters.  Lookups for inclusion are very fast.
+A set of characters: 
+
+- Lookups for inclusion are very fast.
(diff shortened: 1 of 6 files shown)


akgrant43:
All tests pass.  A review would be nice, but otherwise ready to merge.

Ducasse:
To me your changes looks good but I'm not expert in character and character set. But this is why I'm reviewing PRs :) to learn by helping others.

PR #3034, opened 2019-03-27
3033-TextMorphEditView-should-be-packaged-with-PluggableTextMorph

### LINKED ISSUE #3033: TextMorphEditView should be packaged with PluggableTextMorph
Because it is its only user. 

When PluggableTextMorph will be removed, it should also be removed.

jecisc:
Repackage TextMorphForEditView next to PluggableTextMorph.

Fixes #3033


Ducasse:
Please wait before integrating it because I 'm trying to push the other issue.

PR #3039, opened 2019-03-28
3038-Simplify-Slots-that-use-base-slot

### LINKED ISSUE #3038: Simplify Slots that use base slot
no shared hierarchy yet, but as a first step a simplification:

-> no need for #layOutChanged: as #installingIn: is send when the class is fully created
-> offset calculation simplifies.

TODO:
- shared Hierarchy
- tests for removing with instances present
- add a slot higher in the hierarchy: how to move base slot up?

MarcusDenker:
fixes #3038


MarcusDenker:
I need to redo that with a new image

PR #3050, opened 2019-03-29
2895-remove-SmalltalkEditor

### LINKED ISSUE #2895: remove SmalltalkEditor
Now that Rubric is sync back with text/smalltalkEditor, the old editors should be removed.
Let us start with SmalltalkEditor.

estebanlm:
remove dependent extensions

Fixes #2895


estebanlm:
just retaking the PR from Alain and Stef

src/Glamour-Morphic-Renderer/GLMMorphicSmalltalkCodeRenderer.class.st
```diff
@@ -87,7 +87,7 @@ GLMMorphicSmalltalkCodeRenderer >> textModelFor: aPresentation [
 { #category : #rendering }
 GLMMorphicSmalltalkCodeRenderer >> textMorphForModel: aTextModel [
 	| morph  |
-	morph := GLMPluggableTextMorph new.
+	morph := RubPluggableTextMorph new.
 	morph styler workspace: aTextModel.
```
StephanEggermont, line 91: So where is the styler? Does that depend on the DNU?

Ducasse: We will integrate this and you can open another issue. We cannot continue to have large changes in the air.

bencoman: > So where is the styler? Does that depend on the DNU?

@StephanEggermont, was this a side question? or a concern about this change?

StephanEggermont: There have been other issues where the styler is missing, breaking functionality, so this might break something more in the image. I expressed my concern already on the ML

Issue #3021, opened 2019-03-26
Morphic / Checkbox enable / disable 


570gina:
checkboxButtonSelectedDisabledFillStyleFor: aCheckboxButton
	"Return the selected disabled checkbox button fillStyle for the given button."	
	^self checkboxButtonDisabledFillStyleFor: aCheckboxButton

A disable checked checkbox is rendering an unchecked checkbox instead.

pavel-krivanek:
https://github.com/pharo-spec/Spec/issues/233

PR #3054, opened 2019-03-29
A disable checked checkbox can be rendering a checked checkbox.

570gina:
solve #3021


570gina:
t

570gina:
efrgthnybrvb
ntbybyn

MarcusDenker:
"Note: I am not a very intelligent bot, I can only react to new comments. Please add a comment for me if you update the body or title."

MarcusDenker:
Only a new comment will get the bot to remove the flag.

570gina:
The checkbox can show checked state when disabled:)

Ducasse:
Hi gina

did you add a test? Because it would be great. 
We can do that in another issue. 

Stef

PR #3056, opened 2019-03-29
Change stack implementation

Alisu:
Doesn't inherit from LinkedList anymore.
Inherits from Object and use an OrderedCollection.
Fix issue 2863


Ducasse:
you have as yet unclassified. 
Then why not using the message last instead of using lastIndex?
If last is not implemented in OrderedCollection it would be better to add it tha lastIndex
Finally why do you use cull:?
In my implementation (check http://github.com/Ducasse/Containers-Stack I do not use cull:). 

Stef

Alisu:
Clients use linked list API for stacks causing compatibility problems.

PR #3071, opened 2019-03-29
2141-Spotter-should-use-the-current-theme-instead-of-creating-its-own-theme-classes

### LINKED ISSUE #2141: Spotter should use the current theme instead of creating its own theme classes
It is hard currently to create a theme that includes Spotter because the Spotter has its own theme classes. 

Spotter should probably use the background and font colors that are directly in the theme instead of creating its own.

estebanlm:
Fixes #2141


MarcusDenker (approved):
looks good

PR #3076, opened 2019-03-29
Lint cleanup / ban for DiskFileSystemTest>>testEqual

### LINKED ISSUE #3075: Lint cleanup / ban for DiskFileSystemTest>>testEqual
System suggests to use assert:equals: in DiskFileSystemTest>>testEqual
but the reason of the test is to really check for #= comparision.

So we should ban this check on the method by adding it to the manifest.

astares:
fixes #3075


MarcusDenker (approved):
ok

PR #3087, opened 2019-03-29
Abstract collections should redefine #isAbstract

### LINKED ISSUE #3086: Abstract collections should redefine #isAbstract
All the collections in "Collections-Abstract" package currently return "false" when asked for #isAbstract.

```Smalltalk
{Collection isAbstract.
SequenceableCollection isAbstract.
HashedCollection isAbstract.
ArrayedCollection isAbstract }
```
which is wrong.

So we should define #isAbstract to return true.

astares:
fixes #3086


MarcusDenker:
failing tests no related

Issue #3007, opened 2019-03-25
Use <expectedFailure> instead of #expectedFailures


MarcusDenker:
the pragma <expectedFailure> is much better than using the  #expectedFailures method as this way we know when browsing the code that the test is not a test.

==> clean it up but putting the pragma into the tests everywhere but Fuel (there we should check upstream).

SUnitTest>>#expectedFailures can be deleted

MarcusDenker:
Moved from https://pharo.fogbugz.com/f/cases/4711/Use-expectedFailure

<blockquote><img src="/favicon_fogbugz.ico" width="48" align="right"><div><strong><a href="https://pharo.fogbugz.com/f/cases/4711/Use-expectedFailure">FogBugz</a></strong></div></blockquote>

MarcusDenker:
merged PR

PR #3088, opened 2019-03-29
Use <expectedFailure> instead of #expectedFailures.

JurajKubelka:
Changed all but FUEL packages.

issue #3007.


dionisiydk (approved):
And as a present for this we will have these tests grouped in special protocol in the browser

bencoman:
+1. The locality of a pragma is great.  And @dionisiydk having a group for them even better.

MarcusDenker (approved):
looks good. if the test run is ok it can be merge

PR #3096, opened 2019-03-29
2855 autocompletion of a method parameter when typing in a method should be higher

### LINKED ISSUE #2855: autocompletion of a method parameter when typing in a method should be higher
when faced with a method like:
```
executeShould: aBlock inScopeOf: anExceptionalEvent withDescriptionContaining: aString
	^[aBlock value.
 	false] on:  <cursor here>
		do: [:ex | ex return: (ex description includesSubstring: aString) ]
```

And your cursor is after the on: - if you start to type "anE" - the exception parameter is not your first item in the list (its the third, after some letter matching in some un-related classes). I would argue that matching method parameters should always be first in the list - as other wise you always have to scroll down the list, or type more, to easily autocomplete the most commonly used variables?

image

bergel:
Pair programmed with Alejandro Infante. We fixed https://github.com/pharo-project/pharo/issues/2855

### DIFF
--- a/src/NECompletion/NECModel.class.st
+++ b/src/NECompletion/NECModel.class.st
@@ -10,7 +10,7 @@ Class {
 		'narrowString',
 		'entries'
 	],
-	#category : #'NECompletion-Model'
+	#category : 'NECompletion-Model'
 }
 
 { #category : #'instance creation' }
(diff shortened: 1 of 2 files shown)


macta:
This is awesome guys, Would be great if this one might be considered for a future Pharo7 update (assuming we do one at some point).

Ducasse:
Thx. I checked the code and it looks ok. Did not try. I plan to rename such classes.

Ducasse:
GLMRubricTextMorphicTest has some red tests but not on my machine.

Ducasse:
I checked and it is working :)

PR #3097, opened 2019-03-29
2959-Epicea-doesnt-reload-events-in-the-correct-order

tinchodias:
Fixed "lost changes detector" and added a test. Also refactored a bit the already present tests.


src/EpiceaBrowsers-Tests/EpLostChangesDetectorTest.class.st
```diff
@@ -5,7 +5,8 @@ Class {
 	#name : #EpLostChangesDetectorTest,
 	#superclass : #EpEnabledIntegrationTest,
 	#instVars : [
-		'detector'
+		'detector',
+		'detectedLostChanges'
```
jecisc: Shouldn't it be a temporary variable?

tinchodias:
Done, thanks!

PR #3100, opened 2019-03-29
[test] Added a test to check that we break the limit of  directoryDes…

hogoww:
Added a test to check that we break the limit of  directoryDescriptors consistently.

### DIFF
--- a/src/FileSystem-Tests-Core/FileSystemTest.class.st
+++ b/src/FileSystem-Tests-Core/FileSystemTest.class.st
@@ -679,6 +679,23 @@ FileSystemTest >> testRootIsNotAFile [
 	self deny: (filesystem isFile: Path root)
 ]
 
+{ #category : #'tests-streams-compatibility' }
+FileSystemTest >> testTooManyDirectoryDescriptorOpened [
+	| opened testSize |
+
+	self halt.
+	testSize := 1025.
+	opened := OrderedCollection new:testSize.
+
+	self should: 
+		[1 to: testSize do: [ :x | opened add: ((File primOpendir: '/etc') at:3)]] 
+	raise: Error.
+	
+	opened do:[:each |
+    File primClosedir: each. ]
+
+]
+
 { #category : #tests }
 FileSystemTest >> testWorking [
 	self assert: filesystem workingDirectory fileSystem = filesystem.


Ducasse:
Ok the test breaks and I imagine that this is normal since you are helping to spot the problems with the file primitive. Can you confirm it?

hogoww:
Indeed. There was a problem with Icebergs tests that probably opened too many directories, which we had trouble spotting. 
This seems quite weak though. I'm thinking of testing that Directory Descriptors are well closed, and check if they need to be garbage collected to let another one be used.

src/FileSystem-Tests-Core/FileSystemTest.class.st
```diff
@@ -679,6 +679,23 @@ FileSystemTest >> testRootIsNotAFile [
 	self deny: (filesystem isFile: Path root)
 ]
 
+{ #category : #'tests-streams-compatibility' }
+FileSystemTest >> testTooManyDirectoryDescriptorOpened [
+	| opened testSize |
+
+	self halt.
+	testSize := 1025.
+	opened := OrderedCollection new:testSize.
```
guillep, line 688: Here giving the test size is not necessary :)

src/FileSystem-Tests-Core/FileSystemTest.class.st
```diff
@@ -679,6 +679,23 @@ FileSystemTest >> testRootIsNotAFile [
...
+
+	self should: 
+		[1 to: testSize do: [ :x | opened add: ((File primOpendir: '/etc') at:3)]]
```
guillep, line 691: you can use `x timesRepeat: []` instead of `#to:do:`

src/FileSystem-Tests-Core/FileSystemTest.class.st
```diff
@@ -679,6 +679,23 @@ FileSystemTest >> testRootIsNotAFile [
...
+	testSize := 1025.
+	opened := OrderedCollection new:testSize.
+
+	self should: 
+		[1 to: testSize do: [ :x | opened add: ((File primOpendir: '/etc') at:3)]] 
+	raise: Error.
+	
+	opened do:[:each |
+    File primClosedir: each. ]
+
```
guillep, line 696: This method requires some nice formatting...

src/FileSystem-Tests-Core/FileSystemTest.class.st
```diff
@@ -679,6 +679,23 @@ FileSystemTest >> testRootIsNotAFile [
 	self deny: (filesystem isFile: Path root)
 ]
 
+{ #category : #'tests-streams-compatibility' }
+FileSystemTest >> testTooManyDirectoryDescriptorOpened [
+	| opened testSize |
+
+	self halt.
```
guillep, line 686: This is a leftover I imagine.

guillep (changes requested):
I've made some comments on the code.

Also, we have seen the main problem comes from the fact of opening empty directories.
The following tests reproduces another facet of the problem: opening an empty directory at some point fails and the image becomes unusable. There is however no way to close it because the returned handle is nil. Thus, this test is kind of tricky because we cannot recover from the broken state... 

```smalltalk
empty := FileLocator workingDirectory / 'empty'.
empty ensureCreateDirectory.

1 to: 1000 do: [ :i |
	"The index is important to understand when it stopped"
	self assert: (File primOpendir: empty fullName) isNil ]
```

src/FileSystem-Tests-Core/FileSystemTest.class.st
hogoww: agreed

src/FileSystem-Tests-Core/FileSystemTest.class.st
hogoww: It's just to be able to tests several sizes more easily change it, given that the number of open Directory Descriptor varies on each OS. But i'll remove it.

src/FileSystem-Tests-Core/FileSystemTest.class.st
hogoww: Too much crashes made me forget what version i was committing :)

src/FileSystem-Tests-Core/FileSystemTest.class.st
hogoww: Yeup

Ducasse:
So do we close it?

src/FileSystem-Tests-Core/FileSystemTest.class.st
guillep: I meant in the collection's initial size :).

guillep:
Ouki. Comment from the mailing list:

> The main problem is that the only way to test it properly is to open a file, close it and ask your system if the file is still open (which is OS dependent e.g., calling lsof on unix/osx, or the equivalent on windows).
>   The second better test we can do is to try to open as many files as the limit and check that we cannot open more files or that we do not segfault :) but it's not super fun when debugging.

akgrant43:
Hi Everyone,

Sorry, but I'm having trouble figuring out what the driver for this PR was in the first place.  I'm wondering if it might also be fixed by:

https://github.com/OpenSmalltalk/opensmalltalk-vm/commit/a838346b1a67712cc28298534dafbd0c26ea34fb

Thanks,
Alistair

PharoProject:
Hi alistair, that's it. I'd like to build a regression test for it but it
seems overcomplicated.

bencoman:
Can we have a primitive to test for an open file?
cheers -ben

hogoww:
@bencoman In this specific case, it wouldn't help since we get back a nil handler. But it could be interesting to have.
@guillep Looks like the test should be a primitive, that would check after opening a directory if the system count was modified in the right way. 
Hence, if the handler returned is nil, it should be the same
             else the handler points to something, and it should be minus one.
I don't think there's a way to test it from pharo.

akgrant43:
> Can we have a primitive to test for an open file?

How do you expect this would be implemented?

> Looks like the test should be a primitive, that would check after opening a directory if the system count was modified in the right way.

I don't think this can be reliable - files can be opened and closed by libraries within the process outside of the test's control.

My understanding is that the goal is to prevent regressions in FileAttributesPlugin>>primitiveOpendir: where a directory was left open if it was empty.

Guille, do you think the following would be reasonable on Unix (using a horrible mix of smalltalk and bash pseudo code):


| emptyDir openDir myPid |

myPid := LibC uniqueInstance currentProcessId.
"Create a random directory name"
emptyDir := FileLocator tmp / aRandomLongFilename.
"Ensure that it doesn't already exist"
self deny: emptyDir exists.
"Ensure that the process doesn't have the directory open (Unix allows a handle to be kept open after the file / directory is deleted)"
self assert: (bash 'lsof -p myPid | grep emptyDir') isEmpty.
"Create the directory"
emptyDir ensureCreateDirectory.
"Ensure that the create directory primitive didn't leave it open"
self assert: (bash 'lsof -p myPid | grep emptyDir') isEmpty.
"Open the directory.  Since it is empty, the handle should be closed immediately"
openDir := File primOpenDir: emptyDir.
"Check that nothing was returned and that the directory handle is closed"
self assert: openDir isNil.
self assert: (bash 'lsof -p myPid | grep $emptyDir') isEmpty.

I think I can write this such that it will run in a normal image.

Cheers,
Alistair

PR #3108, opened 2019-03-29
3006-When-a-test-fails-open-the-debugger-on-the-test-instead-of-on-Objectassert

akevalion:
I have added <debuggerCompleteToSender> pragma to TestAsserter methods in the protocol #asserting

PR #3116, opened 2019-04-01
3006-When-a-test-fails-open-the-debugger-on-the-test-instead-of-on-Objectassert

akevalion:
Fix: https://github.com/pharo-project/pharo/issues/3006
Added  debuggerCompleteToSender pragma to TestAsserter  and some objects like Process, Object, Exception, Context and BlockClosure

### DIFF
--- a/src/Kernel/BlockClosure.class.st
+++ b/src/Kernel/BlockClosure.class.st
@@ -144,7 +144,7 @@ BlockClosure >> cull: anArg [
 	"([:x | x + 12] cull: 3)
 	>>> 15
 	"
-	
+	<debuggerCompleteToSender>
 	^numArgs = 0 
 		ifTrue: [self value]
 		ifFalse: [self value: anArg]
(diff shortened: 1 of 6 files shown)


akevalion:
I have created a new PR because the last one https://github.com/pharo-project/pharo/pull/3108 has conflicts

akgrant43 (changes requested):
This PR seems to have the same issues as #3008 - it adds the pragma to Object>>assert:, but not to Object>>assert:description:, so the debugger will open in #assert:description: instead of the caller of #assert:.

Can you also close #3008 and #3108 with a link back to this PR so that it is clear what should be reviewed at the moment.

akevalion:
Hi @akgrant43 I have test it. In test classes when you use `Object>>#assert:` this calls `TestAsserter>>#assert:description:` For that reason the debugger opens directly on the failing test method. However I have commit the change for `Object>>#assert:description`

guillep (approved):
Looks ok!

akgrant43 (approved):
> Hi @akgrant43 I have test it. In test classes when you use `Object>>#assert:` this calls `TestAsserter>>#assert:description:` For that reason the debugger opens directly on the failing test method. However I have commit the change for `Object>>#assert:description`

Hmm, that's not what I see in Pharo 8.0.  In the attached image you can see that the stack shows:

UndefinedObject(Object)  assert:description:
UndefinedObject(Object)  assert:
UndefinedObject  DoIt

TestAsserter isn't part of the stack (this can be confirmed by opening the debugger, the comments in TestAsserter>>assert:description: and Object>>assert:description: are different).

Regardless, I agree with Guille, the PR looks good.

Thanks,
Alistair


Screenshot from 2019-04-02 19-25-24

akevalion:
Hi, do I need to change something to Integrate it on pharo8

guillep:
Let's integrate it and we can enhance it further if needed :).
Thanks Milton!!

PR #3030, opened 2019-03-27
Integrating UndefinedClass, ClassDefinition and modified classBuilder…

Alisu:
… to support classDefinition.

The objective is to create a class from its definition, the steps are:
-Add a ClassDefinition model et parser.
-Modify class builder/installer to create a class from previous model.
-Use several strategies in case of missing superclass, default raise Error, another is to use UndefinedClass.
-Add UndefinedClass (stubs for missing classes)

PR #3120, opened 2019-04-01
Reparing pull request #3030

Alisu:
Repair initialize in shift classbuilder.
Clean some tests.
Recover sugsuggestion.


MarcusDenker:
hmmm, I think the revert of #3030 did not remove the files in git. That creates the conflict.

Issue #3122, opened 2019-04-01
Adding debugger test


dupriezt:
New test for step over

PR #3123, opened 2019-04-01
New debugger test for stepOver

dupriezt:
Fixes: #3122


Ducasse:
I love these tests!!!

Issue #3128, opened 2019-04-02
Rubric's text field should limit typing when maxLenght is set


guillep:
If Rubric's maxlength is set as in PR https://github.com/pharo-project/pharo/pull/3129, it should also limit input from typing.

guillep:
I put @Ducasse and @plantec in CC so they are aware of this.

Ducasse:
Thanks Guille.

PR #3129, opened 2019-04-02
3119-Failing-test-on-CI-MorphicTextInputFieldAdapterTesttestMaxLengthIsSetInWidget

### LINKED ISSUE #3119: Failing test on CI: MorphicTextInputFieldAdapterTest>>testMaxLengthIsSetInWidget
MorphicTextInputFieldAdapterTest>>testMaxLengthIsSetInWidget fails on the CI

guillep:
Fix #3119
 - make Rubric text field accept max length
 - make rubric truncate when max length is set
 - this fixes (the existing) spec tests (and should help getting the CI green)

However, rubric does not yet prevent typing when maxLength is set (see issue https://github.com/pharo-project/pharo/issues/3128)

### DIFF
--- a/src/Rubric/RubTextFieldMorph.class.st
+++ b/src/Rubric/RubTextFieldMorph.class.st
@@ -7,7 +7,8 @@ Class {
 	#instVars : [
 		'hasValidText',
 		'acceptOnCR',
-		'entryCompletion'
+		'entryCompletion',
+		'maxLength'
 	],
 	#category : #'Rubric-Editing-Widgets'
 }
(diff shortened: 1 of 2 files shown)


guillep:
Tests are green!

Ducasse:
Thanks guille. 
I will discuss with Alain next time we pair program. 
Because we took a different design for maxLength to avoid forcing to specify it. 
Anyway this is cool to have this test green. We also added some tests for the widget.

Ducasse:
What we did was that maxLength was initialized to nil 
because we do not know what is the maximum and we dealt with that. 
Because if we set maxlength to zero then the user is forced to set a maxLength to be able to do something.

plantec:
I've just pushed a fix  that makes a text area check maxLength with keyStroke:
I've added 2 tests but have to test also if a text area remains ok even with a copy paste.

PR #3133, opened 2019-04-03
[fix] Weird semantic on #propertyAt:ifPresent:IfAbsent: in RBProgramNode

hogoww:
AbsentBlock has to be valued manually by the caller (or it won't be evaluated).  Changed #propertyAt:ifAbsent: to avoid creating a useless new block object.

### DIFF
--- a/src/AST-Core/RBProgramNode.class.st
+++ b/src/AST-Core/RBProgramNode.class.st
@@ -683,7 +683,7 @@ RBProgramNode >> propertyAt: aKey ifAbsent: aBlock [
 	"Answer the property value associated with aKey or, if aKey isn't found, answer the result of evaluating aBlock."
 	
 	^ properties 
-		ifNil: [ aBlock value ]
+		ifNil: aBlock
 		ifNotNil: [ properties at: aKey ifAbsent: aBlock ]
 ]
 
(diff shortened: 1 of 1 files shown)


src/AST-Core/RBProgramNode.class.st
```diff
@@ -698,7 +698,7 @@ RBProgramNode >> propertyAt: aKey ifAbsentPut: aBlock [
 RBProgramNode >> propertyAt: aKey ifPresent: aPresentBlock ifAbsent: anAbsentBlock [
 	"Answer the property value associated with aKey or, if aKey is found, answer the result of evaluating aPresentBlock, else evaluates anAbsentBlock."
 
-	^ properties ifNil: [ anAbsentBlock ] ifNotNil: [ properties at: aKey ifPresent: aPresentBlock ifAbsent: anAbsentBlock ]
```
guillep, line 701: Good catch!

Any chance we could have a test for this?

hogoww:
Done cf #3134. (couldn't do it in the same pull request because of a problem with branches)

guillep:
Ok, so I'll close it in favor of https://github.com/pharo-project/pharo/pull/3134 :)

Thanks!

PR #3134, opened 2019-04-03
[test] added a regression test for RBProgramNode >> #propertyAt:ifPresent:ifAbsent:. 

hogoww:
As requested in previous pull request. (#3133) (didn't do it in the same one because of problems with branches).


guillep:
Test failure not related. Thanks!

PR #3136, opened 2019-04-04
Pharo8.0-ffi

zecke:
Add testcase for the internal structure of the read FFIExternalStructure

Any idea how to make this more than a change detector? I want to asses that "a" is a valid object and can be used for future FFI. I tried to call another callback from within the callback but that seemed to deadlock.

### DIFF
--- a/src/FreeType/FFIExternalStructureReferenceHandle.extension.st
+++ b/src/FreeType/FFIExternalStructureReferenceHandle.extension.st
@@ -1,6 +1,0 @@
-Extension { #name : #FFIExternalStructureReferenceHandle }
-
-{ #category : #'*FreeType' }
-FFIExternalStructureReferenceHandle >> asExternalAddress [
-	^ handle + startOffset
-]
--- a/src/UnifiedFFI-Tests/FFICallbackParametersTests.class.st
+++ b/src/UnifiedFFI-Tests/FFICallbackParametersTests.class.st
@@ -290,6 +290,37 @@ FFICallbackParametersTests >> testPassing4IntStructureInTheStack [
 ]
 
 { #category : #running }
+FFICallbackParametersTests >> testPassingPassableStruct [
+	"Tests that a large struct passed by value can be passed to another callback."
+			
+	| param handleWasExternalAddress |
+
+	handleWasExternalAddress := false.
+	callback := FFICallback 
+		signature: #(int (FFITestStructure4Int64Structure a))
+		block: [ :a |
+			self assert: a x equals: 2. 
+			self assert: a y equals: 3. 
+			self assert: a w equals: 5. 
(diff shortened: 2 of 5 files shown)


estebanlm:
yes, that's more or less how I was expecting it to work :)
I mean: no aliens in the side.

src/FreeType/ExternalAddress.extension.st
```diff
@@ -0,0 +1,6 @@
+Extension { #name : #ExternalAddress }
+
+{ #category : #'*FreeType' }
```
tesonep: Should be moved to UnifiedFFI package

tesonep (changes requested):

src/FreeType/ExternalAddress.extension.st
zecke: FFIExternalStructureReferenceHandle>>#asExternalAddress is inside the FreeType package (https://github.com/pharo-project/pharo/blob/Pharo8.0/src/FreeType/FFIExternalStructureReferenceHandle.extension.st). Should both move to UnifiedFFI?

guillep: Could be nice...

zecke:
ping to prevent auto close.

zecke:
keep open for now.

src/FreeType/ExternalAddress.extension.st
estebanlm: Hi, @zecke 
can you perform the recategorisation, so I can integrate this?
Thanks!
Esteban

zecke: Done. I moved both into UnifiedFFI

MarcusDenker:
I it seems now the image crashes when we use iceberg (directly on cloning a project). This might be a side effect of this PR?

estebanlm:
yes. We need to revert.

estebanlm:
@zecke somehow this change crashes the cloning of projects using HTTPS (maybe other cases too). 
Probably an invalid #free call somewhere, but it needs to be dug. 

Here the way to reproduce the crash (provided by @astares): 

- Open Iceberg
- click on "Add"
- select "Clone from GitHub"
- use "astares" as Owner name and "pharo" as Project name, select "HTTPS" as Protocol, click on "Accept"

Pharo tries to clone the repo - but then the Pharo process just dies.

On Ubuntu the consol tells:

 free(): invalid pointer
 ./pharo-ui: line 11:  3604 Aborted                 (core dumped) "$DIR"/"pharo-vm/pharo" "$@"

jecisc:
A fetch is enough to crash the image for me.

astares:
Details described here: https://github.com/pharo-project/pharo/issues/3461

Issue #3112, opened 2019-03-29
openInEnvironment: does not seem to exist anymore but it is sent


Ducasse:
Denis do you have an idea how we can rewrite the users to use Calypso/....

dionisiydk:
Let's list them:

1. ChangeSorterApplication>>#openChangeSetBrowser. 
2. SystemNavigation>>#browseAllUsersOfTrait:. 
3. SystemNavigation>>#browseClassesWithNamesContaining:caseSensitive:
4. FinderPragmaNode>>#browse. 
5. NautilusRefactoring>>#openEnvironment:for:.

dionisiydk:
And let's remember what Nautilus does in that cases.
It opened full browser where package list was collected for given classes. And when package was selected the class pane showed only classes from this list and from selected package.

For the record it did not work for FinderPragmaNode.

dionisiydk:
Calypso does not support RB environments. There is no way to open FullBrowser on set of classes and have same semantics like in Nautilus. 
I can imagine simple way to show packages from given classes. But restrict the class pane will require much more work.

But listed users can be supported differently. The QueryBrowser can be opened on any Calypso queries. It can show classes, class comments and methods in same list. Particularly it is used to query sources for string.
So Calypso way would be something like this:
```Smalltalk
ClyQueryBrowser openOn: (ClyConstantQuery returning: givenClasses)
```
It allows to fix 1, 2, 3. And I think it is generally better way to show classes to users. With Nautilus we saw packages in first place and I always found it a very strange approach.

The case with FinderPragmaNode (4) is just about senders. And fix is simple:
```Smalltalk
Smalltalk tools messageList browseSendersOf: self item
```
And NautilusRefactoring case needs to be removed. It is used to open browser for any kind of refactorings. But it just does not work for all refactoring cases:
```Smalltalk
NautilusRefactoring>>refactoringOptions: aRefactoring
...
        setOption: #openBrowser
	toUse: [ :ref :env | self openEnvironment: env  for: ref]
``` 
Calypso commands already use this #openBrowser option in the places which need it and in the way which is appropriate for them. We can bravely remove this line and method.

dionisiydk:
And to have a fix with QueryBrowser we need to add new API to not have a dependency issue:
```Smalltalk
Smalltalk tools messageList browseClasses: givenClasses
```
But it needs to be added to Calypso first.

dionisiydk:
[...]

PR #3137, opened 2019-04-04
3112-openInEnvironment-does-not-seem-to-exist-anymore-but-it-is-sent

dionisiydk:
Fix for #3112: calypso way to show list of classes.
For details read issue comments


dionisiydk:
It uses new API #browseClasses: which is not implemented yet. But changes can be integrated because affected code do not work anyway.
New API will be integrated with new version of Calypso (PR is there). And this code will start work.

MarcusDenker (approved):
ok

PR #3142, opened 2019-04-06
Make size_t work inside a nested structure

zecke:
Add the missing selectors to forward the get/set to the bytearray at the right offset.

### DIFF
--- a/src/UnifiedFFI-Tests/FFIExternalStructureFieldParserTests.class.st
+++ b/src/UnifiedFFI-Tests/FFIExternalStructureFieldParserTests.class.st
@@ -30,7 +30,7 @@ FFIExternalStructureFieldParserTests >> testParseFieldsStructure [
 		structure: FFITestStructure.
 		
 	self assert: fieldSpec notNil. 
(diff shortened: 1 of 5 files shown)


src/UnifiedFFI-Tests/FFIExternalStructureFieldParserTests.class.st
```diff
@@ -30,13 +30,14 @@ FFIExternalStructureFieldParserTests >> testParseFieldsStructure [
 		structure: FFITestStructure.
 		
 	self assert: fieldSpec notNil. 
-	self assert: fieldSpec fieldNames equals: #(byte short long float double int64 ulong).
+	self assert: fieldSpec fieldNames equals: #(byte short long float double int64 ulong size).
```
tesonep: why not size_t to follow the same naming than C/C++?

zecke: Changed to have the name match the type.

guillep (approved):
Look ok to me!

PR #3154, opened 2019-04-10
3153-Slots-Cleanups--some-first-ComposableSlot-

### LINKED ISSUE #3153: Slots: Cleanups + some first ComposableSlot
- #resolveSlot should raise SlotNotFound instead of error
- Very first ComposedSlot + some tests (more to be done)
- inline #initializeProperties
- remove #changingIn:, not send by the new class builder
- add DerivedSlot superclass for PropertySlot, BooleanSlot...
- Simplify WeakSlot
- add WeakSlotComposable (will replace WeakSlot later)
- refactor tests: one class per Slot Example

MarcusDenker:
- #resolveSlot should raise SlotNotFound instead of error
- Very first ComposedSlot + some tests (more to be done)
- inline #initializeProperties
- remove #changingIn:, not send by the new class builder
- add DerivedSlot superclass for PropertySlot, BooleabSlot...
- Simplify WeakSlot
- add WeakSlotComposable (will replace WeakSlot later)
- refactor tests: one class per Slot Example

fixes #3153


MarcusDenker:
error not related

src/Slot-Core/ComposedSlot.class.st
```diff
@@ -0,0 +1,60 @@
+"
+Slots can be composed. A composition always has one ""real"" slot that knows how to store values and one ore more SlotDecorators. The decorators are chained and delegate reading/writing to the front of the compostion until the finally use the one storage slot to store or read the value.
+"
+Class {
+	#name : #ComposedSlot,
```
guillep: Why the comment talks about `SlotDecorators` and this one is called `ComposedSlot`? Looking at the implementation `SlotDecorator` looks better..

src/Slot-Core/TemporaryVariable.class.st
```diff
@@ -23,19 +23,10 @@ Class {
...
-		
-]
-
-{ #category : #'class initialization' }
-TemporaryVariable class >> initializeProperties [
-	Properties := WeakIdentityKeyDictionary new.
+	Properties := Properties
+		ifNil: [ WeakIdentityKeyDictionary new ]
+		ifNotNil: [ (WeakIdentityKeyDictionary newFrom: Properties) rehash ].
+	PersistentTemps ifNil: [ PersistentTemps := Set new ]
```
guillep, line 29: I've seen this code three times already in this PR... wouldn't it be worth it to externalize this maybe?

src/Slot-Examples/DerivedSlot.class.st
```diff
@@ -0,0 +1,29 @@
+"
+A derived slot is a virtual slot that stores its value in a hidden base slot.
+"
+Class {
+	#name : #DerivedSlot,
```
guillep, line 5: It's not super clear the difference between a `DerivedSlot` and a `ComposedSlot`/`SlotDecorator`...

src/Slot-Examples/BooleanSlot.class.st
```diff
@@ -106,7 +87,7 @@ BooleanSlot >> todo [
 	'
     - adding a Booleanslot in between --> offsets need to be updated?
 	 - removing BooleanSlot in the middle	
-	 - adding booleanslot higher in the hierarchy: move baseselot (same as in TODO for DictionarySlot)
+	 - adding booleanslot higher in the hierarchy: move baseselot (same as in TODO for PropertySlot)
```
guillep: Having TODOs as methods is no so nice, at least not if they are only understandable by you...

If they are to stay here I'd clarify it a bit more.
For example, it's not clear wether booleanslot = Booleanslot or BooleanSlot. Same with baseslot...

src/Slot-Tests/ComputedSlotTest.class.st
```diff
@@ -37,28 +35,26 @@ ComputedSlotTest >> testReadComputedSlotReflective [
 { #category : #tests }
 ComputedSlotTest >> testReadComputedSlotReflectiveNoArg [
 	| slot object |
-	aClass := self make: [ :builder | builder slots: {(#slot1 => ComputedSlot with: [ 3 + 2 ] )} ].
+	slot := #slot1 => ComputedSlot with: [:obj | 3 + 2 ].
```
guillep: Shouldn't this test use no args in the block? otherwise it's the same as `testReadComputedSlotReflective`

guillep (changes requested):
I've left some questions

src/Slot-Core/TemporaryVariable.class.st
MarcusDenker: We could do a Trait, but we ca nit use traits in the kernel...

src/Slot-Examples/DerivedSlot.class.st
MarcusDenker: Yes, there is the question if I should do all the composition stuff outside the image for now... it is exploratory and will change as the examples are implemented.

src/Slot-Core/ComposedSlot.class.st
MarcusDenker: removed for now, SlotDecorator is really better

src/Slot-Examples/DerivedSlot.class.st
MarcusDenker: I will keep DerivedSlot as a superclass for now as there is code shared, I hope to find a better name as part of the composition work.

src/Slot-Examples/BooleanSlot.class.st
MarcusDenker: removed the TODO, will open issue tracker entry (I have now a plan for how to fix it...)

src/Slot-Tests/ComputedSlotTest.class.st
MarcusDenker: fixed

PR #3155, opened 2019-04-10
Adding flags for Smalltalk reference. Refactor some direct reference …

Alisu:
…to instVar.

NECCompletion had a lot of them and it could be nice to be able to pass another Environment as argument.

### DIFF
--- a/src/NECompletion/NECContext.class.st
+++ b/src/NECompletion/NECContext.class.st
@@ -282,9 +282,6 @@ NECContext >> globals: anObject [
 NECContext >> guessArgument: aSHRange [ 
 	| name |
 	name := self sourceOf: aSHRange.
-	(name = 'html') ifTrue: [
-		 globals at: #WAHtmlCanvas ifPresent: [ :canvasClass | ^canvasClass].
-		 globals at: #WARenderCanvas ifPresent: [ :canvasClass | ^canvasClass]].
 	^ NECInstVarTypeGuesser getClassFromTypeSuggestingName: name
 ]
 
--- a/src/NECompletion/NECContext.class.st
+++ b/src/NECompletion/NECContext.class.st
@@ -407,7 +407,7 @@ NECContext >> initialize [
 	ranges := OrderedCollection new.
 	completionIndex := 0.
 	completionToken := nil.
-	globals := Smalltalk.
+	globals := self class environment.
 ]
 
 { #category : #'private-compute-index' }
--- a/src/NECompletion/NECContext.class.st
+++ b/src/NECompletion/NECContext.class.st
@@ -37,7 +37,8 @@ Class {
 		'model',
 		'controller',
 		'variables',
-		'selectors'
+		'selectors',
+		'globals'
 	],
 	#category : #'NECompletion-Model'
 }
@@ -267,13 +268,23 @@ NECContext >> findSourceRangeFor: aNumber [
 										ifFalse: [^ nil]]]]]
 ]
 
+{ #category : #accessing }
+NECContext >> globals [
+	^ globals
+]
+
+{ #category : #accessing }
+NECContext >> globals: anObject [
+	globals := anObject
+]
+
 { #category : #'private-type-guessing' }
 NECContext >> guessArgument: aSHRange [ 
 	| name |
 	name := self sourceOf: aSHRange.
 	(name = 'html') ifTrue: [
-		 Smalltalk at: #WAHtmlCanvas ifPresent: [ :canvasClass | ^canvasClass].
-		 Smalltalk at: #WARenderCanvas ifPresent: [ :canvasClass | ^canvasClass]].
+		 globals at: #WAHtmlCanvas ifPresent: [ :canvasClass | ^canvasClass].
+		 globals at: #WARenderCanvas ifPresent: [ :canvasClass | ^canvasClass]].
 	^ NECInstVarTypeGuesser getClassFromTypeSuggestingName: name
 ]
 
@@ -305,7 +316,7 @@ NECContext >> guessClassVarClass: aSHRange [
 { #category : #'private-type-guessing' }
 NECContext >> guessGlobal: aNumber [ 
 	| aClass |
-	aClass := Smalltalk 
+	aClass := globals
 		at: (self sourceOf: (ranges at: aNumber)) asSymbol
 		ifAbsent: [ ^ nil ].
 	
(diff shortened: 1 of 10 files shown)


src/NECompletion/NECEntry.class.st
```diff
@@ -34,6 +34,7 @@ NECEntry >> browseWith: anECContext [
 	| guessType |
 	guessType := self guessTypeWith: anECContext.
 	guessType ifNil: [^ false].
+	self flag: 'Reference to Smalltalk tools'.
```
jecisc:
```suggestion
	self flag: #SmalltalkReference. "Reference to Smalltalk tools"
```

src/NECompletion/NECPreferences.class.st
```diff
@@ -285,6 +285,7 @@ NECPreferences class >> spaceAfterCompletion: anObject [
 
 { #category : #accessing }
 NECPreferences class >> theme [
+self flag:#SmalltalkReference.
```
jecisc:
```suggestion
	self flag: #SmalltalkReference.
```

src/NECompletion/NECUntypedModel.class.st
```diff
@@ -93,6 +105,7 @@ NECUntypedModel >> initializeInstVars [
 NECUntypedModel >> initializeSelectors [
 	self resetSelectors.
 	narrowString size < 2  ifTrue: [ ^ self ].
+	self flag:#DoWeWantAllSelectors.
```
jecisc:
```suggestion
	self flag: #DoWeWantAllSelectors.
```

src/NECompletion/NOCCompletionTable.class.st
```diff
@@ -167,6 +167,7 @@ NOCCompletionTable >> maxNumberOfEntries [
 
 { #category : #protocol }
 NOCCompletionTable >> quickFillWithCategoryOf: cls [
+	self flag:#SmalltalkReference.
```
jecisc:
```suggestion
	self flag: #SmalltalkReference.
```

jecisc (changes requested):
I added suggestions to follow common formatting in Pharo + a proposition to improve coherence in flags.

Alisu:
Ok thanks jecisc, i am going to change that.

src/NECompletion/NECContext.class.st
```diff
@@ -395,7 +406,8 @@ NECContext >> initialize [
 	selectors := true.
 	ranges := OrderedCollection new.
 	completionIndex := 0.
-	completionToken := nil
+	completionToken := nil.
+	globals := Smalltalk.
```
jecisc: I think it is better to use `self class environment` instead of Smalltalk directly.

Someone told me that one time, like that we could have other environments than Smalltalk in the future.

jecisc: Maybe you can ask confirmation to Guille.

src/NECompletion/NECEntry.class.st
```diff
@@ -34,6 +34,7 @@ NECEntry >> browseWith: anECContext [
 	| guessType |
 	guessType := self guessTypeWith: anECContext.
 	guessType ifNil: [^ false].
+	#SmalltalkReference. "Reference to Smalltalk tools"
```
jecisc: The flag is missing, only the parameter is here.

src/NECompletion/NECContext.class.st
guillep: Yes

guillep:
@jecisc can you check that the new changes are ok?

src/NECompletion/NECContext.class.st
```diff
@@ -267,13 +268,23 @@ NECContext >> findSourceRangeFor: aNumber [
...
+{ #category : #accessing }
+NECContext >> globals: anObject [
+	globals := anObject
+]
+
 { #category : #'private-type-guessing' }
 NECContext >> guessArgument: aSHRange [ 
 	| name |
 	name := self sourceOf: aSHRange.
 	(name = 'html') ifTrue: [
```
jecisc: It's not a problem with this PR but... Why do we have Seaside related code directly in Pharo?! Maybe a hook is missing in the code completion for other frameworks to plug their behavior?

src/NECompletion/NECVarTypeGuesser.class.st
```diff
@@ -18,6 +18,7 @@ Class {
 
 { #category : #private }
 NECVarTypeGuesser class >> classFrom: className [
+
```
jecisc, line 21: No flag here?

jecisc (commented):
Seems fine for me. I added a comment at a place where I think it is missing a flag.

src/NECompletion/NECVarTypeGuesser.class.st
Alisu: I was planning on removing this one but forgot about it in the end. It needs to be pushed in the instance side.

src/NECompletion/NECContext.class.st
guillep: 🗡 👯‍♂️ 🍡

jecisc:
Please do not remove that without adding a hook to get the same behavior if you work on it :( 

It must be horrible to code in Seaside if the code completion propose methods that are not from the canvas in rendering methods :'(

src/NECompletion/NECContext.class.st
```diff
@@ -267,13 +268,23 @@ NECContext >> findSourceRangeFor: aNumber [
...
-		 Smalltalk at: #WAHtmlCanvas ifPresent: [ :canvasClass | ^canvasClass].
-		 Smalltalk at: #WARenderCanvas ifPresent: [ :canvasClass | ^canvasClass]].
+		 globals at: #WAHtmlCanvas ifPresent: [ :canvasClass | ^canvasClass].
+		 globals at: #WARenderCanvas ifPresent: [ :canvasClass | ^canvasClass]].
```
tesonep: These should be completed removed, both lines

tesonep (changes requested):

src/NECompletion/NECContext.class.st
jecisc: If you remove them, please provide a hook to get back the behavior. Just removing it without any equivalent would reduce a lot the productivity in Seaside.

tesonep: that is outside the PR, if you want to do it, your are always welcome. 
If we have to keep the hack, it has to be in Seaside, not in the image.

guillep (approved):
To me this is a good change. Test failure not related (Zodiac)

tesonep (approved):

guillep:
Thanks Theo!

PR #3157, opened 2019-04-10
[Partial Fix] Experiment to test FreeType Problem

tesonep:
After checking the problem with Guille, we have the hypothesis of the source of the problem.
We have seen that accessing Free Type is not thread-safe.
Basically, the FTFace holds a structure that is filled up with the glyph and its information. 
As this structure is shared by the whole Font, only one request to a glyph can be executed at the time.

Also, we have seen that there many accesses to the glyphs outside the UI process. 
This problem started to appear as we starting to use Calypso (but it is not limited to it), as Calypso uses lazy tabs. The creation of these tabs is done outside the UI process.

This is only a hack to test our hypothesis. 
To fix it correctly we will have to rewrite the drawing of the glyphs and synchronize the access to the glyph data information. Also, we want to do it in a way that does not penalize the processing in the UI process. 

I have only patched the code that is used by the rendering of morphic, other renderings like Athens or any other user using FT should be correctly rewritten.

### DIFF
--- a/src/FreeType-Tests/FreeTypeCacheTest.class.st
+++ b/src/FreeType-Tests/FreeTypeCacheTest.class.st
@@ -113,6 +113,47 @@ FreeTypeCacheTest >> testFreeTypeCacheEntry [
 ]
 
 { #category : #tests }
+FreeTypeCacheTest >> testGlyphAccessIsThreadSafe [
+
+	"This test is weak because it depends in the rendering of Morphic and Rubrik"
+	"This test breaks 95% of times"	
+
+	self timeLimit: 10 seconds.
+	
+	10 timesRepeat: [  
+		 | sem text canvases blocky |
+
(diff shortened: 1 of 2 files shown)


tesonep:
Once we are sure that synchronizing the access to FT fixes the problem, we will do a real fix not a dirty hack like this.

tesonep:
I have added a test that validates the error is not happening.
In a stock image the test if failing 95% of the time, so if you run it 10 times. It will only give a false possitive 1 every 10^14 times. 

So we think it is pretty good.

tesonep:
This quick and dirty solution works, although we need a better implementation of the font usage.

pavel-krivanek (approved):

MarcusDenker:
I will merge this so we can see if the hypothesis is true. (tests look ok)

Issue #2975, opened 2019-03-22
Problem when you read an old non utf8 mcz


Alisu:
When you try to read an mcz in iso-latina the #contentsForMember: member method is bugged.
It works with a String but got a Stream instead.

VincentBlondeau:
Could you give an example that can be easily reproduced?

Alisu:
When i try to load this for example.
repository := MCHttpRepository
    location: 'http://www.squeaksource.com/MutationTesting'
    user: ''
    password: ''.

versionInfo := repository versionInfoFromVersionNamed: 'MutationTesting'.
version := repository versionWithInfo: versionInfo.

version snapshot

guillep:
Yes, this problem is there in Pharo 7 and Pharo8. We are working on a solution.

PR #3159, opened 2019-04-10
2975-Problem-when-you-read-an-mcz

guillep:
Fix for #2975

contentStreamFromEncoding: should return a steam and not a string.

Fix for Pharo8. Should be backported for Pharo7.


guillep:
Failing tests not related

PR #3160, opened 2019-04-10
2975-Problem-when-you-read-an-mcz-pharo7

### LINKED ISSUE #2975: Problem when you read an old non utf8 mcz

guillep:
Fix #2975

contentStreamFromEncoding: should return a steam and not a string.
Fix for Pharo7


guillep:
Failing tests not related
image

PR #3164, opened 2019-04-11
MailMessage can not send mails with attachment in Pharo7

SabineMa:
This is my first pull request I hope I did it correct...
Question: will this go to Pharo8 automatically or do I have to put it there, too?

PR #3166, opened 2019-04-11
3163

SabineMa:
adding the same changes from https://github.com/pharo-project/pharo/pull/3164 to Pharo8, too...


SabineMa:
this was not correct, I close it!

PR #3171, opened 2019-04-11
3170-PreDebugger-should-work-with-Spec2

### LINKED ISSUE #3170: PreDebugger should work with Spec2
The PreDebugger is currently breaking in Spec 2 which makes it hard to develop it. I'll migrate it to Spec2.

jecisc:
Make GTSpecPreDebugger work with Spec 2.

WARNING: This will break the possibility to open the full debugger by clicking on the stack for now. This will be fixed by the next Spec2 integration. The full debugger can still be opened by clicking on the "Debug" button. 
I propose the change because it'll make it easier to work on Spec2 like this (currently we get an emergency debugger at each error...) and will already be done for the next Spec2 integration.

Fixes #3170

### DIFF
--- a/src/GT-Debugger/GTSpecPreDebugWindow.class.st
+++ b/src/GT-Debugger/GTSpecPreDebugWindow.class.st
@@ -9,17 +9,14 @@ Class {
 	#superclass : #DynamicComposablePresenter,
 	#instVars : [
 		'debugger',
-		'title',
 		'message'
 	],
 	#category : #'GT-Debugger-UI'
 }
 
 { #category : #specs }
-GTSpecPreDebugWindow class >> spec [
-	<spec: #default>
-		^ SpecLayout composed
-			yourself
+GTSpecPreDebugWindow class >> defaultSpec [
+	^ SpecLayout composed
 ]
 
 { #category : #'actions lookup' }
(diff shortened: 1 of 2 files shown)


src/GT-Debugger/GTSpecPreDebugWindow.class.st
```diff
@@ -179,26 +163,24 @@ GTSpecPreDebugWindow >> initializeCodePane [
...
-			ifNotNil:  [ self rebuildWidget ]	 ]
 
+	debugger whenChangedDo: [ :aDebugger | aDebugger ifNotNil: [ self rebuildWidget ] ]
 ]
 
 { #category : #'initialization widgets' }
 GTSpecPreDebugWindow >> initializeStackPane [
 	self stackPane
-		displayBlock: [ :aContext | self columnsFor: aContext ];
+		addColumn: (StringTableColumn title: 'Class' evaluated: [ :aContext | self debugger formatStackClassColumnForContext: aContext ]);
```
estebanlm, line 174: this will have a problem because this methods answer a Text and they will be editable in new Table Presenter

src/GT-Debugger/GTSpecPreDebugWindow.class.st
```diff
@@ -179,26 +163,24 @@ GTSpecPreDebugWindow >> initializeCodePane [
...
+		addColumn: (StringTableColumn title: 'Method' evaluated: [ :aContext | self debugger formatStackMethodColumnForContext: aContext ]);
+		addColumn: (StringTableColumn title: 'Extra' evaluated: [ :aContext | self debugger formatStackExtraColumnForContext: aContext ]);
 		items: self filteredStack;
 		whenSelectionChangedDo: [ :selection | 
-			[ :aContext | 
-			"Set the selection before, as debugAction removes the link with the debugger. "
-			self debugger stackPresentation selection: aContext.
-			self openFullDebugger ] cull: selection selectedItem ]
+			selection selectedItem
```
estebanlm, line 179: I would prefer to extract this into a `openFullDebuggerOnContext: selection selectedItem` method

src/GT-Debugger/GTSpecPreDebugWindow.class.st
jecisc: Is there a kind of column managing text or should we create it/modify the string one?

src/GT-Debugger/GTSpecPreDebugWindow.class.st
jecisc: Ok with me.

Ducasse: So cyril? is it ready for integration?

jecisc:
I need to do 2 little changes. I’ll do it next week. But it’s not as urgent as before because I fixed in Soec what was breaking the debugger.

PR #3183, opened 2019-04-14
ClassParser: methods polymorphic to RB-nodes for Calypso integration

dionisiydk:
- parent field is added to CD nodes
- methods polymorphic to RB-nodes which are required for Calypso integration (isLiteralVariable, isVariable, isGlobalVariable, etc)

### DIFF
--- a/src/ClassParser-Tests/CDBehaviorParserTest.class.st
+++ b/src/ClassParser-Tests/CDBehaviorParserTest.class.st
@@ -116,3 +116,25 @@ CDBehaviorParserTest >> testClassDefFromLegacyStringHasSlots [
 	self assert: classDefinition slotNodes first name equals: self firstInstanceVariableName.
 	self assert: classDefinition slotNodes second name equals: self secondInstanceVariableName.
 ]
+
+{ #category : #tests }
+CDBehaviorParserTest >> testClassNameNodeHaveParentReference [
+
+	| nameNode |
+	nameNode := classDefinition classNameNode.
+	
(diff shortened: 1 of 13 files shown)


guillep:
Hi Denis, this change is at least breaking the three following tests:

image

dionisiydk:
Hi @guillep.
Thanks for notice (always skip failure reports). I fixed tests. 

What do you think to rename more messages for testing the type of class in definition? (in another PR).
I feel it a bit weird to read "aClassDefinitionNode isBytes". I think it would be better to have a pattern like "definesTypeNameClass": definesImmediateClass, definesBytesClass, definesNormalClass and so on.

Ducasse:
Hi denis

Indeed we should improve the API. 
We should aslo remove some duplicated code I was forced to introduce. 
Stef

guillep:
> What do you think to rename more messages for testing the type of class in definition? (in another PR).
> I feel it a bit weird to read "aClassDefinitionNode isBytes". I think it would be better to have a pattern like "definesTypeNameClass": definesImmediateClass, definesBytesClass, definesNormalClass and so on.

Yes maybe that could be good? Do you want to add that to this PR or to add a new one?

@Ducasse have you reviewed the changes?

dionisiydk:
I will do rename in another PR

вт, 16 апр. 2019 г., 14:07 Guille Polito <notifications@github.com>:

guillep (approved):
Ok!

@Alisu maybe you want to merge also these changes into ClassInfrastructure?

PR #3185, opened 2019-04-15
3184-KeyboardKey-class-should-have-a-method-to-get-an-instance-from-its-name

juliendelplanque:
- Added #named: method on class-side of KeyboardKey.
- Fixed name of KeyboardKey: were all capitalized which clash for 'a' and 'A' for example.
- Added test class for KeyboardKey: KeyboardKeyTest.
- Modified the baseline to make it loaded in images.

### DIFF
--- a/src/System-Platforms/KeyboardKey.class.st
+++ b/src/System-Platforms/KeyboardKey.class.st
@@ -675,6 +675,11 @@ KeyboardKey >> printOn: aStream [
 		nextPutAll: ')'
 ]
 
+{ #category : #accessing }
+KeyboardKey >> value [
+	^ value
+]
+
 { #category : #initialization }
 KeyboardKey >> withValue: aValue [
 	self withValue: aValue andName: self class unknownKeyName
--- a/src/System-Platforms/KeyboardKey.class.st
+++ b/src/System-Platforms/KeyboardKey.class.st
@@ -604,6 +604,16 @@ KeyboardKey class >> macOSVirtualKeyTable [
 	^MacosVirtualKeyTable
 ]
 
+{ #category : #'instance creation' }
+KeyboardKey class >> named: aString [
+	"Returns the instance of myself having aString as name.
+	
+	 (KeyboardKey named: 'SHIFT_L') >>> (KeyboardKey value: 65505)
+	"
+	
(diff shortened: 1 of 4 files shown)


juliendelplanque:
@guillep it might be interesting that you do a review.

src/System-Platforms/KeyboardKey.class.st
```diff
@@ -257,7 +257,7 @@ KeyboardKey class >> initializeKeyTable [
 		16r007c bar "XK_bar"
 		16r007d braceright "XK_braceright"
 		16r007e asciitilde "XK_asciitilde")
-			pairsDo: [ :keyCode :keyname | KeyTable at: keyCode put: (self basicNew withValue: keyCode andName: keyname asUppercase) ].
+			pairsDo: [ :keyCode :keyname | KeyTable at: keyCode put: (self basicNew withValue: keyCode andName: keyname) ].
```
guillep: I don't think this is right... I see that in the key table we have both keys:
```
XK_A
...
XK_a
````

But they are not really mapped in the platforms. The thing is that there is no such an "uppercase A" key. There is only an "A" key.

juliendelplanque: Since both are defined shouldn't we have the possibility to instantiate an object for both?

guillep: I'm not sure, it does not look right at first sight. We should try with a real case first being conclusive. Also, if you see the mappings in `initializeMacOSVirtualKeyTable`, `initializeUnixVirtualKeyTable` and `initializeWindowsVirtualKeyTable` it is evident that even if we have an "uppercase A" we will never receive such an event because it is not mapped...

You can try the `KeyPrinterMorph new openInWindow`. Even with Capslock activated, I have always a (lowercase) A key code = 65.

guillep: Maybe the right thing would be to remove the wrong one?

juliendelplanque: Ok, I'm convinced, I will remove the wrong one.

juliendelplanque:
I think broken tests are not related.

guillep (approved):
Good

PR #3189, opened 2019-04-16
Kernel tests should be correctly identified

tesonep:
When the kernel tests run, the errors and results should be correctly identified.


guillep (approved):
Good!

PR #3190, opened 2019-04-16
3188-Some-classes-have-an-users-property

### LINKED ISSUE #3188: Some classes have an "users" property
SymbolTest has an "users" property which is specific to trait whereas SymbolTest is a class (at least not a trait).

Edit: it seems a lot of classes (if not all) using trait gain an extra property "users" set at nil.
The only class using trait I found without it is the one i created from ClassDefinition.

tesonep:
- Handling correctly the slots coming from the Trait itself.
- Adding tests.
Fixes #3188

### DIFF
--- a/src/TraitsV2-Tests/T2TraitTest.class.st
+++ b/src/TraitsV2-Tests/T2TraitTest.class.st
@@ -46,6 +46,20 @@ T2TraitTest >> createT2 [
 ]
 
 { #category : #tests }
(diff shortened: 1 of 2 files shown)


guillep (approved):
Good!

Alisu:
testTraitHaveUsersInstanceVariable is failing both for Unix and Win is it normal ?

PR #3204, opened 2019-04-18
3203-Add-trait-TSingleton-

### LINKED ISSUE #3203: Add trait TSingleton
This is the TSingleton fron https://github.com/pharo-project/pharo/pull/2528
(just the class, no code changed to use it)

MarcusDenker:
fixes #3203

### DIFF
--- a/src/Kernel-Traits/TSingleton.trait.st
+++ b/src/Kernel-Traits/TSingleton.trait.st
@@ -1,0 +1,26 @@
+"
+I am a trait that makes a class into a singleton whose unique instance is accessed via the method TSingleton>>uniqueInstance
+"
+Trait {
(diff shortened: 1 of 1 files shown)


guillep (changes requested):
Why no adding also the users?

MarcusDenker:
Then we could just use the original PR: https://github.com/pharo-project/pharo/pull/2528

guillep:
But why adding the `TSingleton` if it has no users?

I understood from the other PR that we had to check the TSingleton users so we do not integrate dependencies to traits in the kernel. That does not mean to have no users at all, but to intelligently pick them.

Otherwise having it there has not real value.
People could just load it as an external library.

What am I missing?

PR #3205, opened 2019-04-18
3198-WaitfreeQeuenextOrNilSuchThat-memory-leak

### LINKED ISSUE #3198: WaitfreeQeue>>nextOrNilSuchThat: memory leak
WaitfreeQeue>>nextOrNilSuchThat: currently never removes zombie queue items, i.e. an AtomicQueueItem where the object points back to the dummy entry.  This means the list of items grows until #nextOrNil is called enough times to empty the queue, potentially growing the queue indefinitely and wasting memory and CPU cycles.

- Update the class comments to describe how thread-safety is achieved and the constraints imposed as part of thread-safety. (read this if the opening paragraph doesn't make sense)
- Modify #nextOrNilSuchThat: to remove zombie list items (unless it is the last item in the list).
- Add automated tests.

Additional details on tracking the issue down and figuring out the appropriate behaviour at: https://github.com/feenkcom/gtoolkit/issues/244

akgrant43:
WaitfreeQueue>>nextOrNilSuchThat: currently never removes zombie queue items, i.e. an AtomicQueueItem where the object points back to the dummy entry.  This means the list of items potentially grows indefinitely unless #nextOrNil is called enough times to empty the queue, wasting memory and CPU cycles.

- Modify #nextOrNilSuchThat: to remove zombie list items (unless it is the last item in the list).
- Update the class comments to describe how thread-safety is achieved and the constraints imposed as part of thread-safety.
- Add automated tests.

Additional details on tracking the issue down and figuring out the appropriate behaviour at: https://github.com/feenkcom/gtoolkit/issues/244

Fixes: #3198


guillep:
Hi @akgrant43, why changing the semantics of `#nextOrNilSuchThat:`? The original method comment states clearly that it is "skipping (but not removing) any intermediate objects". So I see no conflict there.

In case you need the intermediate objects to be removed, I'd say that it would just be better to add a new method that "pops elements until a block" for example.

Or is there something else to take into account?

akgrant43:
Hi Guille,

Thanks for your review.

> Hi @akgrant43, why changing the semantics of `#nextOrNilSuchThat:`? The original method comment states clearly that it is "skipping (but not removing) any intermediate objects". So I see no conflict there.
> 
> In case you need the intermediate objects to be removed, I'd say that it would just be better to add a new method that "pops elements until a block" for example.
> 
> Or is there something else to take into account?

But I'm not removing the intermediate objects.  What I've changed is that the zombie AtomicQueueItem that is left in the linked list in the current implementation will be removed unless it is the last item in the linked list (don't forget that an AtomicQueueItem holds the object in the list).  So there's no change in semantics.

Cheers,
Alistair

Ducasse:
I love to see all these new tests!

src/Collections-Atomic/WaitfreeQueue.class.st
```diff
@@ -139,6 +159,29 @@ WaitfreeQueue >> isEmpty [
 	^ false
 ]
 
+{ #category : #accessing }
+WaitfreeQueue >> itemCount [
+	"Answer the number of items in the queue including zombies. 
+	If queue is currently in the middle of extraction by other process, give a (very) approximate answer.
+	
+	This method having a non-deterministic results, because of nature of concurrent shared queue.
```
bjorn3:
```suggestion
	This method has a non-deterministic results, because of nature of concurrent shared queue.
```

src/Collections-Atomic/WaitfreeQueue.class.st
```diff
@@ -139,6 +159,29 @@ WaitfreeQueue >> isEmpty [
...
+	| item count |
+
+	item := dummy.
+
+	count := 0.
+	[ item := item next. item == dummy ifFalse: [ count := count + 1 ]. item isCircular ] whileFalse.
+
+	count = 0 ifTrue: [
+		"if queue is locked by other process, check the tail and give a (very) approximate answer."
+		^ tail object == dummy ifTrue: [ 0 ] ifFalse: [ 1]
```
bjorn3:
```suggestion
		^ tail object == dummy ifTrue: [ 0 ] ifFalse: [ 1 ]
```

src/Collections-Atomic/WaitfreeQueue.class.st
akgrant43: Done.

src/Collections-Atomic/WaitfreeQueue.class.st
akgrant43: Done.  Thanks for your feedback!

akgrant43:
I've merged in the latest Pharo 8.0 branch.

It would still be nice to get this in.

Just to clarify the change in behaviour:

I'm using a few different terms specifically:

- an item (instance of AtomicQueueItem) is what is used to form the linked list in the queue.
- an item can either be holding an entry (object) or nothing, in which case it's referred to as a zombie item.

The issue is that #nextOrNilSuchThat: doesn't remove the zombie items from the linked list, resulting in a memory leak.

As the comment says it is "skipping (but not removing) any intermediate objects".  These are AtomicQueueItems that hold on to an actual object.

The only change is to remove the zombie entries (which are null entries anyway).

akgrant43:
This PR still has no conflicts, and the CI failures appear to be unrelated.

MarcusDenker:
I will close and reopen to run in the latest code

MarcusDenker (approved):
code looks good

src/Collections-Tests/WaitfreeQueueTest.class.st
```diff
@@ -0,0 +1,216 @@
...
+]
+
+{ #category : #'helper methods' }
+WaitfreeQueueTest >> queueFromCollection: aCollection [
+	"Answer a WaitfreeQueue initialised from the supplied collection of items"
+
+	| queue |
+
+	queue := WaitfreeQueue new.
+	(queue instVarNamed: #dummy) next: aCollection first.
```
dionisiydk: Can we avoid instVarNamed? 
It breaks encapsulation and code navigation. No way to find those methods using var references

dionisiydk: In this case it should method of queue class instead of test class.

src/Collections-Tests/WaitfreeQueueTest.class.st
```diff
@@ -0,0 +1,216 @@
...
+
+{ #category : #tests }
+WaitfreeQueueTest >> testZombieFirst [
+	"Check that a zombie item at the start of the queue is correctly removed"
+
+	| items queue real |
+
+	items := self newItems: 2.
+	queue := self queueFromCollection: items.
+	items first object: (queue instVarNamed: #dummy).
```
dionisiydk: It could be item method #beZombie or queue method #makeZombieAtItem:

src/Collections-Tests/WaitfreeQueueTest.class.st
akgrant43: Hi Denis,

> Can we avoid instVarNamed?
> It breaks encapsulation and code navigation. No way to find those methods using var references

I'm interested in your thoughts here.  There's two rules that both make sense:

1. Don't access internal state using `#instVarNamed:` for the reasons you outline above, as well as being slow.
2. Don't expose internal state that other objects shouldn't have access to (the dummy variable in this case).

I try to balance this by not using `#instVarNamed:` in production code, but using it for setting up internal state in associated tests (which are intimately tied to the tested class).

I can add a #dummy access method, but it will the be breaking the second rule.

dionisiydk: I would say let's be honest. In Smalltalk we can't really avoid exposed internal state. Because we need to examine the state in tests, we need to examine the state during navigation, debugging and scripting. Trying to do it through reflection does not change those needs. It just complicates our life. 
But it does not mean that we should design our code using message chains through accessors. We all know best practice, right?  

And here we also can avoid it by moving method to queue itself:
```Smalltalk
WaitfreeQueue withAll: aCollection
```
instead of 
```Smalltalk
WaitfreeQueueTest >> queueFromCollection: aCollection
```

Ducasse:
May be we should extent SUnit to be able to define some tests in the class.

src/Collections-Tests/WaitfreeQueueTest.class.st
akgrant43: Hi Denis,

Beautifully put, obvious now that you've said it, and I'm feeling embarrassed that you had to say it.

I've pushed an update.

Thanks!
Alistair

src/Collections-Atomic/WaitfreeQueue.class.st
```diff
@@ -295,3 +365,13 @@ WaitfreeQueue >> size [
 
 	^ count
 ]
+
+{ #category : #initialization }
+WaitfreeQueue >> withAll: aCollection [
```
dionisiydk: little note:
It would be better to name instance side method as #nextPutAll: to be closer to #nextPut: .

src/Collections-Atomic/WaitfreeQueue.class.st
```diff
@@ -295,3 +365,13 @@ WaitfreeQueue >> size [
...
+	"Initialise the receiver with the supplied items"
+
+	dummy next: aCollection first.
+	1 to: aCollection size - 1 do: [ :i |
+		(aCollection at: i) next: (aCollection at: i+1) ].
+	tail := aCollection last.
```
dionisiydk: And could you explain why it is not implemented using #nextPut:?
I see that comment points to initialization of queue. Is it difficult to have it as a common #nextPutAll: method? 
For initialization only I would propose name #initializeWithAll:.

src/Collections-Atomic/WaitfreeQueue.class.st
akgrant43: > It would be better to name instance side method as #nextPutAll: to be closer to #nextPut: .

But the semantics are different:

- `#nextPut:` takes a value (entry) which is added to the queue in a thread-safe manner.
- `#withAll:` takes a collection of items (AtomicQueueItem) and adds them to the queue in a non-thread-safe manner.

It would of course be possible to write `#nextPutAll:`, but there hasn't been a need for it yet.

src/Collections-Atomic/WaitfreeQueue.class.st
akgrant43: > And could you explain why it is not implemented using #nextPut:?
> I see that comment points to initialization of queue. Is it difficult to have it as a common #nextPutAll: method?

For testing I want to be able to set up the items in a pre-determined manner, e.g. be able to add zombie items.  That wouldn't be possible with `#nextPutAll:`.


> For initialization only I would propose name #initializeWithAll:.

Good suggestion, I'll change the name.

src/Collections-Tests/WaitfreeQueueTest.class.st
akgrant43: > It could be item method #beZombie or queue method #makeZombieAtItem:

An item can't make itself a zombie as it doesn't know what the dummy item is (the queue uses object identity to test).  It could be passed the dummy item, but that seems like extending the interface too far.

I've replaced the remaining uses of `#instVarNamed:`

akgrant43:
Marcus, thanks for removing the stale label :-)

I believe all the feedback to date has been addressed (and github says there are no merge conflicts).

Ducasse:
Rekicked in the build process.

jecisc:
If tests passed and feedback was already given and taken into account maybe we can merge?

akgrant43:
Thanks to everyone for their feedback and Marcus for merging!

PR #3210, opened 2019-04-19
2261-DarkTheme-depends-on-Shout

### LINKED ISSUE #2261: DarkTheme depends on Shout!
shStyleTable

	^SHRBTextStyler darkStyleTable

and 
UITheme class>> 
...

self class environment at: #SHPreferences ifPresent: [ :shPreferences |
		shPreferences setStyleTable: aUITheme shStyleTable ].
...

There is something wrong in terms of cyclic references.

estebanlm:
removing dependency with Shout

Fixes #2261

### DIFF
--- a/src/Polymorph-Widgets/PharoDarkTheme.class.st
+++ b/src/Polymorph-Widgets/PharoDarkTheme.class.st
@@ -359,9 +359,9 @@ PharoDarkTheme >> selectionTextColor [
 ]
 
 { #category : #defaults }
-PharoDarkTheme >> shStyleTable [
+PharoDarkTheme >> shStyleTableName [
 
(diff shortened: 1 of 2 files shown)


Ducasse:
esteban 

Smalltalk classNamed: -> Smalltalk globals classNamed: 
even better self class environment classNamed: 

S.

Ducasse:
I integrate it but the logic of the UITheme is bullshit because it should have a themer or an object  responsible for the style and this object should be able to be set or not. Now like that with a false reference

estebanlm:
I know is not top. 
Sonner than later we need to think in a way to handle themes. 
With @jecisc we are playing with some "stylesheet-like" design for Spec2, that should be the starting point.

PR #3215, opened 2019-04-23
Exporting MCZ from the BaselineOfMetacello

tesonep:
During the bootstrap some packages are loaded using Monticello.
It should use the list of packages to import Metacello from the Baseline of Metacello.

This is important as Monticello does not support stateful traits. 
The number of packages exported with it should be minimized to the only required.
Also, as Gofer is used by Metacello, it tries to generate an entry in the package-cache, that cannot be done in this scenario. So, the cache should be disabled also.


estebanlm (approved):
yeah, failures are not related.
rest seems good.

PR #3223, opened 2019-04-23
3222-RBChange-hierarchy-should-avoid-to-point-directly-to-RBRefactoryChangeManager-changeFactory

### LINKED ISSUE #3222: RBChange hierarchy should avoid to point directly to RBRefactoryChangeManager changeFactory
It is better to introduce an instance variable in the root of the change hierarchy and to initialize to RBRefactoryChangeManager changeFactory once for all. 

We should chase and kill GLOBALS!

Ducasse:
Fixes: #3222
Finished to hardcode RBRefactoryChangeManager everywhere. 
Finished to hardcode Smalltalk globals.


Ducasse:
I will check testRefactoring – Windows32.Refactoring2.Transformations.Tests.RBRemoveProtocolTransformationTest

PR #3226, opened 2019-04-24
3225-RGPackage-should-be-RGPackageDefinition

### LINKED ISSUE #3225: RGPackage should be RGPackageDefinition
There is a big mess in Ring naming. Most of the classes have the suffix -Definition however some of them like RGPackage not. In the Ring2 I named all the classes naturally (like RGClass) but in case of the package, I needed to use RGPackageDefinition to have the Ring2 loadable next to the original Ring. 

We should fix it simultaneously in both versions. This will break some compatibility with external users of the old Ring. Unfortunately, we cannot load Ring2 and put into its package class a constructor that will simply redirect the instantiation to the old Ring because both use the same method (`named:`)

pavel-krivanek:
remame RGPackage to RGPackageDefinition because of the namin inconsistency and collision with the new Ring version

fixes #3225


Ducasse:
So pavel what should I do? Is it ok to still introduce this PR and handle the bootstrap part later?

pavel-krivanek:
the bootstrap part does not need any update so it should be ready to merge

pavel-krivanek:
...it will need to be changed as soon as it will switch to Pharo 8

PR #3230, opened 2019-04-25
3217-copy-class-do-not-copy-class-side-instanceVariableNames

### LINKED ISSUE #3217: copy class do not copy class side instanceVariableNames
copy a class where there are instanceVariableNames in class side
and you wil see that the new class do not have any instanceVariableNames in class side.

MarcusDenker:
fixes  #3217


Ducasse:
Marcus
Was test broken?
Because else it means that we should add a test to cover this behavior else it may break unnoticed.

MarcusDenker:
I added a test

MarcusDenker:
error not related

jecisc (approved):
Seems fine!

PR #3247, opened 2019-04-26
[WIP]3246-Compiling-a-class-does-not-reset-the-cache-of-compiled-method-using-it

### LINKED ISSUE #3246: Compiling a class does not reset the cache of compiled method using it
If you write a method using an undeclared class and then create the class, the AST of the compiled method will not be reset and will still reference the undeclared.

With Theo, we will propose a PR to correct this.

jecisc:
When changing undefine classes to a defined one, reset the AST cache of the compiled methods using it.

Result of a Sprint with Theo.

Fixes #3246

### DIFF
--- a/src/AST-Core/ASTCache.class.st
+++ b/src/AST-Core/ASTCache.class.st
@@ -45,6 +45,11 @@ ASTCache class >> initialize [
 	SessionManager default registerSystemClassNamed: self name
 ]
 
+{ #category : #accessing }
+ASTCache class >> remove: aCompiledMethod [
+	default removeKey: aCompiledMethod
+]
+
 { #category : #'initialization-release' }
 ASTCache class >> reset [
 	default reset.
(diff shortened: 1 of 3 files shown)


dionisiydk:
What is interesting is to pass it through benchmarks because it could impact speed of code loading significantly

jecisc:
I think it should not because the extra work will only be executed in the case of undeclared classes.

MarcusDenker:
#usingMethods seems to not work for the bootstrap, it seems that uses Ring internally which is not loaded?

dionisiydk:
It could be that code loading will generate undeclared in the middle of
process then they will become resolved.

Maybe it is more clever and it first creates all classes

пт, 26 апр. 2019 г., 13:01 CyrilFerlicot <notifications@github.com>:

jecisc:
Do not close. I started to check if we could fix the problem with undeclared before ring is loaded. It will just take some time.

PR #3254, opened 2019-04-26
3175-RBChange-should-work-on-environment-and-not-Smalltalk-globals

### LINKED ISSUE #3175: RB*Change should work on environment and not Smalltalk globals
and their new environment variable could be initialized to Smalltalk globals for now.

Alisu:
Remove smalltalks references in this case 'self class environment' seems the best choice.
Fixes #3175 .

### DIFF
--- a/src/NautilusRefactoring/RBAddMethodChange.extension.st
+++ b/src/NautilusRefactoring/RBAddMethodChange.extension.st
@@ -29,9 +29,8 @@ RBAddMethodChange >> accept: aText notifying: aController [
 RBAddMethodChange >> oldVersionTextToDisplay [
 	| class |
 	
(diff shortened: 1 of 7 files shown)


VincentBlondeau (changes requested):
Conflicted and maybe some tests should be added to ensure the behavior inside several environments

Ducasse:
Theo I like your introduction of a method classNamed: now I prefer my introduction of onSystemDictionary because the systemDictionary is not necessary the one in which the class RBRefactoryChange resides. So could do a new PR which 
- introduce classNamed: but using onSystemDictionary and 
- use classNamed: instead of the ugly at: that I did

Tx
stef

PR #3257, opened 2019-04-26
[WIP]3256-Remove-Halt-from-Kernel

### LINKED ISSUE #3256: Remove Halt from Kernel
I propose as a first step to move Halt from Kernel and put it in Debugging-Core.

jecisc:
Extract Halt in Debugging-Core to reduce Kernel size.

Fixes #3256


Ducasse:
I do not really understand because none of the jobs worked but the system reports 96 broken tests.

Ducasse:
Trying to relaunch the builds

jecisc:
Errors seem related. I'll need to check

jecisc:
I'll close this one and start a new PR. Here it removes Halt from Kernel but not from the bootstrap. We should remove Halt from the bootstrap since Halt relies on the compiler that is not in the image at the bootstrap level.

Issue #3250, opened 2019-04-26
RBTemporaryNode shouldn't be visited as ArgumentsNode


hogoww:
For a sequences nodes, temporaries declarations are visited as arguments.
They shouldn't. 
PR in progress to add an intermediary hook, while retaining old behavior.

MarcusDenker:
fixed

PR #3261, opened 2019-04-27
[fix) issue #3250

hogoww:
I hope i didn't forget to commit something (had a few computer crashes while working on this :))


hogoww:
added fix for #3262 as well.

PR #3270, opened 2019-04-28
Fixes: #3245 to ensure that we cannot create an empty protocol.

### LINKED ISSUE #3245: We should have a check on protocol names
It is possible outside Pharo to edit some tonel code and to have "empty" protocols (protocols that are just empty strings).

Currently, Pharo will load the code but Calypso cannot display it. 

We discussed with Steph and we think that the solution is to check while loading the code if a protocol is valid. If not, the protocol should be set to as yet classified.

Ducasse:
Added a test and when a category is empty turn it into a 'as yet unclassified'.
Fixes: #3245


Ducasse:
Broken tests are unrelated.

MarcusDenker (approved):
Ok!

PR #3273, opened 2019-04-29
3272-Slot-Demo-HistorySlot

### LINKED ISSUE #3272: Slot Demo: HistorySlot
add a demo: History slot.

It saves 5 old values.

MarcusDenker:
fixes #3272

Implementation of HistorySlot and tests

### DIFF
--- a/src/Slot-Tests/HistorySlotTest.class.st
+++ b/src/Slot-Tests/HistorySlotTest.class.st
@@ -1,0 +1,79 @@
+Class {
+	#name : #HistorySlotTest,
+	#superclass : #SlotSilentTest,
+	#category : #'Slot-Tests-Examples'
+}
+
+{ #category : #tests }
+HistorySlotTest >> testAddOneElement [
+	| slot instance |
+	slot := #slot1 => HistorySlot.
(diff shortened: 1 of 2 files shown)


MarcusDenker:
failing tests not related

src/Slot-Tests/HistorySlotTest.class.st
```diff
@@ -0,0 +1,79 @@
...
+	slot := #slot1 => HistorySlot.
+	aClass := self make: [ :builder | builder slots: {slot} ].
+
+	self compileAccessorsFor: slot.
+
+	instance := aClass new.
+	instance slot1: 25.
+	instance slot1: 35.
+	self assert: instance slot1 equals: 35.
+	self assert: (slot readHistory: instance) size equals: 3
```
bencoman: #size seems a softer condition than an array of the historic values, and also demonstrate how it works.
Is it #(25 35 35) or #(35 35 25) ?

MarcusDenker: yes, I made it better... especially as the first value is nil, which is not at all that obvious

bencoman:
nice. thx.

PR #3277, opened 2019-04-30
3276-adding-initializeSlots-may-break-instance-initialization

### LINKED ISSUE #3276: adding initializeSlots: may break instance initialization
`initializeSlots:` is being added to class initialize when we define slots who `wantsInitalization`.

The problem is that it is being added as last action of the initialize method and it should be the first (because the initialisation chain could be giving value to the variables that will be reseted when calling initializeSlots).

Additionally, `initializeSlots:` is being added without taking into account that it may have been added before in the initialisation chain (making the slots to be initialised multiple times).

estebanlm:
traverse initialize in class to verify initializeSlots is there.

Fixes #3276

### DIFF
--- a/src/Slot-Core/Slot.class.st
+++ b/src/Slot-Core/Slot.class.st
@@ -378,14 +378,16 @@ Slot >> scope: aScope [
 { #category : #private }
 Slot >> sendsInitializeSlots: aClass [
 		
-	(aClass includesSelector: #initialize) ifFalse: [ 
-		^ self sendsInitializeSlots: aClass superclass ].
+	aClass isTrait ifFalse: [ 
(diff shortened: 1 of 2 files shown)


MarcusDenker (approved):
looks good

MarcusDenker:
Failing tests... it seems the generated code is bidding a . somehow?

receiver of "super" is nil
Stacktrace
UndefinedObject>>doesNotUnderstand: #super
SlotTestsClassA>>initialize
SlotTestsClassA class(Behavior)>>new
HistorySlotTest>>testAddTwoElements

estebanlm:
now is ok.
@MarcusDenker, can you re-review?

MarcusDenker (approved):
ok1

Issue #3287, opened 2019-05-02
aTrait isClass should be false


Alisu:
aTrait isClass anwser true.

myroslavarm:
aTrait isClass should (and does) give false, whereas Trait isClass should (and does) give true. I think this issue should be closed.

dionisiydk:
Ok. Better confusion example:
```Smalltalk
ClyTraitRoot1 isKindOf: Class. "==> true"
ClyTraitRoot1 isClass "==> will be false"
``` 
What explanation will be for this logic? It would be nice to put it in the comment for #isClass method if you will decide to perform this change.

jecisc:
This seems to have changed

```
TEmpty isClass "false"
```

PR #3288, opened 2019-05-02
3287-Redefine isClass in Trait

Alisu:
Should correct issue #3287

### DIFF
--- a/src/TraitsV2/Trait.class.st
+++ b/src/TraitsV2/Trait.class.st
@@ -192,6 +192,12 @@ Trait >> isBaseTrait [
 ]
 
 { #category : #testing }
+Trait >> isClass [
+	
+	^ false
+]
+
+{ #category : #testing }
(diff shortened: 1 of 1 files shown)


MarcusDenker (dismissed):
Yes, that looks like a good change. no tests break with this

dionisiydk:
Looking at senders of #isClass it will affect some existing behaviour.
For example CompiledMethod>>#referencedClasses will not return referenced traits. Maybe it is expected change but some tools could rely on it.

Generally considering that Trait is now subclass of Class I feel it confusing that isClass will return false while "Trait inheritsFrom: Class" will be true

tesonep (approved):
I like the change, if there is need for changes they should be added as tests. 
The method in CompiledMethod should use #isBehavior if that is required.

astares (approved):
OK

tesonep (approved):

PR #3290, opened 2019-05-03
3289-beForText-is-deprecated-but-there-are-two-senders

### LINKED ISSUE #3289: #beForText is deprecated but there are two senders
... one is in createNotifierPaneWidgets of the debugger.

This means that one gets an infinite loop as soon as the debugger tries to open

MarcusDenker:
fixed #3289

### DIFF
--- a/src/GT-Debugger/GTSpecPreDebugWindow.class.st
+++ b/src/GT-Debugger/GTSpecPreDebugWindow.class.st
@@ -95,9 +95,6 @@ GTSpecPreDebugWindow >> createButtonWidgetsForActions: aCollection [
 GTSpecPreDebugWindow >> createNotifierPaneWidgets [
 
 	self instantiatePresenters: self buildNotifierPaneWidgetsSpec.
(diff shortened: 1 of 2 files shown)


StephanEggermont (changes requested):
Uhm, not calling the deprecated method is not going to result in the correct widget being used?

jecisc:
Before TextPreserter managed Text and Code. Now we have two different presenters. 

So if we instantiate a text presenter, the behavior will be correct. If we instantiated a Code presenter, it will not be correct. 

In case we want a code presenter which can sometimes present text (it happens for example in the changes sorter), esteban is working on the disabling of syntax highlighting for the CodePresenter.

MarcusDenker:
failing tests not related

Ducasse:
Indeed it looks good. Tx Cyril.

StephanEggermont:
So yes, the correct widgets are being used. The pre-debugger doesn't show code, and the behavioral change in TextPresenter is that it now shows text by default, not code.

PR #3293, opened 2019-05-03
add example: LazySlot

MarcusDenker:
manual PR as iceberg is broken without fixing #3289


MarcusDenker:
failing tests not related

PR #3298, opened 2019-05-05
Use Calypso 0.16.3

dionisiydk:
- double click on tab header to expand it to full window
- fixes for TelePharo in pharo 8


MarcusDenker (approved):
ok. no failing tests

PR #3313, opened 2019-05-07
3312-KMKeymapnotifyPartialMatchTo-leads-to-an-error

### LINKED ISSUE #3312: KMKeymap>>#notifyPartialMatchTo: leads to an error
```
notifyPartialMatchTo: aListener
	| listeners |
	aListener isCollection ifTrue: [ listeners := aListener ] ifFalse: [ { aListener } ].
	listeners do: [ :l | l partialMatch ].
```

If aListener is not a collection, then we will have an error.

estebanlm:
Fixes #3312

### DIFF
--- a/src/Keymapping-Core/KMKeymap.class.st
+++ b/src/Keymapping-Core/KMKeymap.class.st
@@ -142,8 +142,11 @@ KMKeymap >> notifyCompleteMatchTo: aListener buffer: aBuffer [
 { #category : #matching }
 KMKeymap >> notifyPartialMatchTo: aListener [
 	| listeners |
-	aListener isCollection ifTrue: [ listeners := aListener ] ifFalse: [ { aListener } ].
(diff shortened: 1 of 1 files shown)


Ducasse:
Esteban can you avoid using do: aSymbol in the image. 
We should not promote such practice. It will make the analyse of code a lot more complex
and it is slower. 

Finally did you add a test? If we do not add test each we fix something then we can be sure that we will not make progress on coverage and the system will break again in the future.

Ducasse:
BTW I found the use of isCollection super bad design. I know it was like that still it is bad.

estebanlm:
ok, changed.
I didn't do the `#isCollection` part, I just wanted an easy fix to continue working on key bindings for Gtk (we cannot use the same KMDispatcher, but we can reuse most of the infrastructure).

PR #3317, opened 2019-05-08
3316-Slots-more-tests-improve-comments-improve-Initialized-slots

### LINKED ISSUE #3316: Slots: more tests, improve comments, improve Initialized slots
- ClassBuilder: do not call layoutChanged: as we call #installingIn: already
- Typos
- add tests for ProcessLocalSlot
- more tests for ExampleSlotWithDefaultValue
- comment AbstractInitializedSlot
- support blocks for Initialized Slots
- LazySlot: assign default in first read
- Improved class comment Slot
- ExampleSlotWithDefaultValue renamed to InitializedSlot

MarcusDenker:
- ClassBuilder: do not call layoutChanged: as we call #installingIn: already
- Typos
- add tests for ProcessLocalSlot
- more tests for ExampleSlotWithDefaultValue
- comment AbstractInitializedSlot
- support blocks for Initialized Slots
- LazySlot: assign default in first read
- Improved class comment Slot
- ExampleSlotWithDefaultValue renamed to InitializedSlot

fixes #3316


src/Slot-Core/Slot.class.st
```diff
@@ -1,16 +1,20 @@
...
 I'm a meta-object for accessing a slot in an Object. 
 
-I define a protocol to read (#read:) and to write (#write:to:) values. 
+I define a reflective protocol to read (#read:) and to write (#write:to:) values. 
+
+	For efficiency, subclasses (e.g. InstanceVariableSlot) override #emitStore: and #emitValue: to not call #read: and/or write:to:
 
 For customizing a subclass can override the meta-object-protocol methods. See subclasses for examples. If your subclass holds state, the following must be overridden:
-- Printing: #definitionString must return the source code string to reconstruct an instance. This will be displayed in the class definition. By default, #definitionString is the printString, so you can also override #printOn: instead.
+- Printing: #printOn: must print the source code string to reconstruct an instance. This will be displayed in the class definition.
```
juliendelplanque, line 9: In the case we do not use #definitionString anymore, I propose to use #storeOn: for this purpose.

Indeed, #printOn: is not expected to print executable code where #storeOn: is expected to do that.

src/Slot-Core/Slot.class.st
```diff
@@ -1,16 +1,20 @@
...
 For customizing a subclass can override the meta-object-protocol methods. See subclasses for examples. If your subclass holds state, the following must be overridden:
-- Printing: #definitionString must return the source code string to reconstruct an instance. This will be displayed in the class definition. By default, #definitionString is the printString, so you can also override #printOn: instead.
+- Printing: #printOn: must print the source code string to reconstruct an instance. This will be displayed in the class definition. 
 - Equality: #= and #hash must be customized taking account of additional state.
 
+
 Vocabulary:
-- variable: named accessor for a Slot
+- Field: space occupied in an object, used to hold values accessed via Slots
 - Slot: class-side meta-object, mapping of names to values using a MOP to fields
```
juliendelplanque, line 15: What is MOP ?

juliendelplanque (changes requested):
Seems nice, my main point is that we should use #storeOn: instead of #printOn: to replace #definitionString.

src/Slot-Core/Slot.class.st
MarcusDenker: Meta Object Protocol.

src/Slot-Core/Slot.class.st
MarcusDenker: Not that easy... the problem is that slots might have additional state that is not part of the definition... e.g. the offset of IndexedSlot and subclasses.

MarcusDenker: I will add it to the backlog... maybe the #printDefinitionOn: idea is not bad after all: this then allows both storeOn: and printOn: to have a good default *and* to be changed indpendently

src/Slot-Core/Slot.class.st
MarcusDenker: I will improve the comment (added to backlog)

PR #3318, opened 2019-05-09
2598-isSpecial-is-not-an-explicit-name-for-a-method

### LINKED ISSUE #2598: #isSpecial is not an explicit name for a method
	
Original case: https://pharo.fogbugz.com/f/cases/22586
Status: Work Needed
Project: Usability
Date: 16 October 2018 7:51:58 am
	
Description:
	
While browsing the implementation of Association I found a method #isSpecial. 

The name is not revealing. Maybe this method (and its other implementations) can have a better name? Because what does it mean that it is special or not?

MarcusDenker:
- simplfy #hasSameDefinitonAs:
- Improve comment Slot 
- describe owningClass and definingClass 
- isSpecial -> useFullDefinition
- improvement of #usesSpecialSlot and #usesSpecialClassVariables is in the backlog.

fixes #2598

### DIFF
--- a/src/Collections-Support/Association.class.st
+++ b/src/Collections-Support/Association.class.st
@@ -40,13 +40,6 @@ Association >> isSelfEvaluating [
 	^ self class == Association and: [self key isSelfEvaluating and: [self value isSelfEvaluating]]
 ]
 
-{ #category : #testing }
-Association >> isSpecial [
-	"Implemented here so we can use simple Assocications as Bindings instead of LiteralVariable
-	subclasses"
-	^false
-]
-
 { #category : #accessing }
 Association >> key: aKey value: anObject [ 
 	"Store the arguments as the variables of the receiver."
@@ -79,6 +72,13 @@ Association >> storeOn: aStream [
 	aStream nextPutAll: '->'.
 	value storeOn: aStream.
 	aStream nextPut: $)
+]
+
+{ #category : #testing }
+Association >> useFullDefinition [
+	"Implemented here so we can use simple Assocications as Bindings instead of LiteralVariable
+	subclasses"
+	^false
 ]
 
 { #category : #evaluating }
--- a/src/GT-SpotterExtensions-Core/Behavior.extension.st
+++ b/src/GT-SpotterExtensions-Core/Behavior.extension.st
@@ -69,8 +69,8 @@ Behavior >> spotterTraitUsersFor: aStep [
 Behavior >> spotterUsedSlotsFor: aStep [
 	<spotterOrder: 70>
 	aStep listProcessor
-			title: 'Special Slots';
-			allCandidates: [ self slots select: [ :slot | slot isSpecial ] ];
+			title: 'Full Defintion Slots';
+			allCandidates: [ self slots select: [ :slot | slot useFullDefinition ] ];
 			itemName: [ :item | item definitionString ];
 			filter: GTFilterSubstring
 ]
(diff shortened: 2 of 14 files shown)


src/GT-SpotterExtensions-Core/Behavior.extension.st
```diff
@@ -69,8 +69,8 @@ Behavior >> spotterTraitUsersFor: aStep [
 Behavior >> spotterUsedSlotsFor: aStep [
 	<spotterOrder: 70>
 	aStep listProcessor
-			title: 'Special Slots';
-			allCandidates: [ self slots select: [ :slot | slot isSpecial ] ];
+			title: 'Full Defintion Slots';
```
juliendelplanque:
```suggestion
			title: 'Full Definition Slots';
```

src/Slot-Core/Slot.class.st
```diff
@@ -12,7 +12,13 @@ For customizing a subclass can override the meta-object-protocol methods. See su
 
 Vocabulary:
 - Field: space occupied in an object, used to hold values accessed via Slots
-- Slot: class-side meta-object, mapping of names to values using a MOP to fields
+- Slot: class-side meta-object, implements readind and writing.
```
juliendelplanque:
```suggestion
- Slot: class-side meta-object, implements reading and writing.
```

src/Slot-Core/Slot.class.st
```diff
@@ -201,11 +206,11 @@ Slot >> hasProperty: aKey [
...
 Slot >> hasSameDefinitionAs: otherSlot [
 
-	"similar to #= but does not take a slot index into count"
-	self == otherSlot
-		ifTrue: [ ^ true ].
-	^ (self species == otherSlot species) 
-			and: [ name = otherSlot name ]
+	"equal definition. Slots an have additional state that is not part of the definitoon
+	(e.g. the index in IndexSlot).
+	This method then is overriden to not take thtat state into account"
```
juliendelplanque:
```suggestion
	This method then is overriden to not take that state into account"
```

MarcusDenker:
failing tests not related

juliendelplanque (changes requested):

src/Collections-Support/Association.class.st
```diff
@@ -81,6 +74,13 @@ Association >> storeOn: aStream [
 	aStream nextPut: $)
 ]
 
+{ #category : #testing }
+Association >> useFullDefinition [
```
dionisiydk, line 78: One notes Marcus.
Considering other true/false methods it should begin with "uses" instead of use.

But generally I would try to avoid this method at all. Seems it is used only for class definition logic and it can be simple printStringForClassDefinition with friends without introducing that "type" message

MarcusDenker: printStringForClassDefinition would be a good solution of we would not use this in addition to check if a class needs to print a slot definition...

MarcusDenker: maybe "needsFullDefinition" would be better

MarcusDenker: committed #3320 which changes it to #needsFullDefinition

dionisiydk:
чт, 9 мая 2019 г., 14:59 Marcus Denker <notifications@github.com>:

> *@MarcusDenker* commented on this pull request.
> ------------------------------
>
> In src/Collections-Support/Association.class.st
> <https://github.com/pharo-project/pharo/pull/3318#discussion_r282497959>:
>
> > @@ -81,6 +74,13 @@ Association >> storeOn: aStream [
>  	aStream nextPut: $)
>  ]
>
> +{ #category : #testing }
> +Association >> useFullDefinition [
>
> committed #3320 <https://github.com/pharo-project/pharo/issues/3320>
> which changes it to #needsFullDefinition
>
Yes, it is better. Thanks

PR #3321, opened 2019-05-09
3320-Improve-HistorySlot-and-other-small-fixes

### LINKED ISSUE #3320: Improve HistorySlot and other small fixes
- add two tests for HistorySlot related to adding  slots to existing instances (green!)
- add #rawRead: like the Spec Slot to start to use the same vocabulary
- use #rawRead: instead of super read:
- add #emitValue: for speed. (no emitStore yet)
- simplify usesSpecialVariables 
- simplify definitionForNautilus
- useFullDefinitions --> needsFullDefinition

MarcusDenker:
- add two tests for HistorySlot related to adding  slots to existing instances (green!)
- add #rawRead: like the Spec Slot to start to use the same vocabulary for these wrappers
- use #raweRead instead of super read:
- add #emitValue: for speed. (no emitStore yet)
- simplify usesSpecialVariables 
- simplify definitionForNautilus
- useFullDefinitions --> needsFullDefinition

fixes #3320

### DIFF
--- a/src/Collections-Support/Association.class.st
+++ b/src/Collections-Support/Association.class.st
@@ -56,6 +56,13 @@ Association >> literalEqual: otherLiteral [
 	^self == otherLiteral
 ]
 
+{ #category : #testing }
+Association >> needsFullDefinition [
+	"Implemented here so we can use simple Assocications as Bindings instead of LiteralVariable
+	subclasses"
+	^false
+]
+
 { #category : #printing }
 Association >> printOn: aStream [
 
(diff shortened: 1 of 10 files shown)


src/Collections-Support/Association.class.st
```diff
@@ -56,6 +56,13 @@ Association >> literalEqual: otherLiteral [
 	^self == otherLiteral
 ]
 
+{ #category : #testing }
+Association >> needsFullDefinition [
+	"Implemented here so we can use simple Assocications as Bindings instead of LiteralVariable
```
juliendelplanque:
```suggestion
	"Implemented here so we can use simple Associations as Bindings instead of LiteralVariable
```

src/Slot-Core/LiteralVariable.class.st
```diff
@@ -127,6 +127,13 @@ LiteralVariable >> name: aString [
 	self key: aString asSymbol
 ]
 
+{ #category : #testing }
+LiteralVariable >> needsFullDefinition [
+	"only ClassVariable can used a simplified definition"
```
juliendelplanque:
```suggestion
	"only ClassVariable can use a simplified definition"
```

juliendelplanque (changes requested):
Some typos to fix.

PR #3339, opened 2019-05-13
3338-Slots-isSpecial-cleanup-part-2

### LINKED ISSUE #3338: Slots: #isSpecial cleanup part 2
- rename #usesSpecialVariables-> #needsSlotClassDefinition
- rename #usesSpecialSlot -> slotsNeedFullDefinition
- rename #usesSpecialClassVariables -> classVariablesNeedFullDefinition

in addition:

- simplify #=>
- remove dead code: #asValidInstVarName

MarcusDenker:
rename #usesSpecialVariables-> #needsSlotClassDefinition
rename #usesSpecialSlot -> slotsNeedFullDefinition
rename #usesSpecialClassVariables -> classVariablesNeedFullDefinition
in addition:

simplify #=>
remove dead code: #asValidInstVarName

fixes #3338

### DIFF
--- a/src/Slot-Core/Slot.class.st
+++ b/src/Slot-Core/Slot.class.st
@@ -286,6 +286,12 @@ Slot >> name: aSymbol [
 	name := aSymbol
 ]
 
+{ #category : #accessing }
+Slot >> named: aSymbol [
+	"to be polymorhic with slot class"
+	name := aSymbol
+]
+
 { #category : #testing }
 Slot >> needsFullDefinition [
 	"I am more than just a backward compatible ivar slot, I need a full definition "
--- a/src/Slot-Core/Slot.class.st
+++ b/src/Slot-Core/Slot.class.st
@@ -289,7 +289,7 @@ Slot >> name: aSymbol [
 { #category : #accessing }
 Slot >> named: aSymbol [
 	"to be polymorhic with slot class"
-	name := aSymbol
+	self name: aSymbol
 ]
 
 { #category : #testing }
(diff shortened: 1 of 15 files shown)


MarcusDenker:
failing tests are related

MarcusDenker:
errors not related

src/Slot-Core/Slot.class.st
```diff
@@ -286,6 +286,12 @@ Slot >> name: aSymbol [
 	name := aSymbol
 ]
 
+{ #category : #accessing }
+Slot >> named: aSymbol [
+	"to be polymorhic with slot class"
+	name := aSymbol
```
juliendelplanque: I think:
```suggestion
	self name: aSymbol
```

Would be better to avoid code duplication.

juliendelplanque (changes requested):
Except for my remark, it seems fine to me.

src/Slot-Core/Slot.class.st
```diff
@@ -286,6 +286,12 @@ Slot >> name: aSymbol [
 	name := aSymbol
 ]
 
+{ #category : #accessing }
+Slot >> named: aSymbol [
+	"to be polymorhic with slot class"
+	self name: aSymbol
+]
```
dionisiydk, line 293: I understand the intention and maybe it is not important. But according to other similar methods (like reversed, sorted, with ~ed suffix)  I would expect new instance in that case instead of modification of receiver. 
And I can imagine the use case for this. For example rename refactoring could ask existing slot to produce new one with new name. So the #named: message will return a new copy with given name.

MarcusDenker: yes, I have it in the backlog... with the composition I need to clean this up anyway.

PR #3348, opened 2019-05-14
2279-We-should-revisit-definitionForNautilus-

### LINKED ISSUE #2279: We should revisit definitionForNautilus
We can ask a class its definition and definition for nautilus. 
The definition returns old format
 'Object subclass: #Point
	instanceVariableNames: ''x y''
	classVariableNames: ''''
	poolDictionaries: ''''
	category: ''Kernel-BasicObjects'''

and definitionForNautilus

'Object subclass: #Point
	instanceVariableNames: ''x y''
	classVariableNames: ''''
	package: ''Kernel-BasicObjects'''

which is a lot better. 

Now it would be good make sure that changersorter and other tools can load the new version 
and rename definitionForNautilus into definition and definition in oldDefinition

estebanlm:
do not use #definitionForNautilus anymore (instead, use just definition).

Fixes #2279


MarcusDenker:
lots of failing tests.

I tried this once and gave up...

MarcusDenker:
#3339 touches definitionForNautilus, too... I should have merged it, but I am waiting for a review.

MarcusDenker:
but it looks already quite good: mostly RB tests failing, my problem was Monticello back then

MarcusDenker (dismissed):
failing tests need to be fixed

Ducasse:
Oops we still have related test failing.

estebanlm:
resuming this. 
I will try to fix it today, but this is going to be just a middle-term solution. In a next issue, we need to change all the #oldDefinition sends with a real ClassDefinition object, that can compare by entity and not a single string comparison as it is now.

estebanlm:
ok now, errors not related :)

MarcusDenker (approved):
ok

PR #3357, opened 2019-05-16
3355-Strange-logic-in-RBFactoryChangeManager

### LINKED ISSUE #3355: Strange logic in RBFactoryChangeManager
RBFactoryChangeManager is a singleton but 
it creates a factory instance (which is essentially stateless) 

changeFactory
	^ RBRefactoryChangeFactory new

Ducasse:
fixes: #3355
Just add an instance var and initialize it.


Ducasse:
BrokenTests are unrelated.

PR #3359, opened 2019-05-16
3356-In-RBFactoryChange-all-methods-should-use-changefactory-instance-variable-instead-of-accessing-the-singleton

### LINKED ISSUE #3356: In RBFactoryChange, all methods should use changefactory instance variable instead of accessing the singleton
In RBFactoryChange, all methods should use changefactory instance variable instead of accessing the singleton. Looks obvious but it was apparently not.

Ducasse:
Fixes: #3356: 
- remove usage of class and use its initialized instance instead
- clean class hardcoding in related tests (all green).


Ducasse:
Broken tests are unrelated.

Ducasse:
I do not know how to solve the conflicts and I cannot get a new image. 
May be just doing a fetch can work ... will try later

PR #3363, opened 2019-05-16
3361-Process-should-not-depend-on-Halt

### LINKED ISSUE #3361: Process should not depend on Halt
The second step to repackage Halt. Kernel classes should not depend on Halt. Currently, `Process` depends on it.

jecisc:
Improve Process>>complete: to not add direct dependency on specific exception

Fixes #3361

### DIFF
--- a/src/Kernel/Exception.class.st
+++ b/src/Kernel/Exception.class.st
@@ -150,6 +150,15 @@ Exception >> freezeUpTo:  aContext [
 	signalContext := signalContext copyTo: aContext
 ]
 
+{ #category : #private }
+Exception >> handleProcessCompletionIfNeeded: aProcess with: aContext [
+	"I am a hook during process completion in case one of my subclass wants to plug a special behavior.
+	
+	I should return nil if I do nothing."
+
+	^ nil
+]
+
 { #category : #testing }
 Exception >> isNested [
 	"Determine whether the current exception handler is within the scope of another handler for the same exception."
--- a/src/Kernel/Exception.class.st
+++ b/src/Kernel/Exception.class.st
@@ -97,6 +97,15 @@ Exception class >> signalIn: context [
 	^ self new signalIn: context
 ]
 
+{ #category : #private }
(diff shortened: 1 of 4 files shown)


src/Kernel/Process.class.st
```diff
@@ -145,23 +145,20 @@ Process >> calleeOf: aContext [
...
 	suspendedContext := pair first.
 	error := pair second.
-	error class = Halt ifTrue: [ 
-		aContext methodNode 
-			pragmaNamed: #debuggerCompleteToSender 
-			ifPresent: [ ^ self completeTo: aContext sender ] ].
-	error ifNotNil: [^ error signalerContext].
-	^ suspendedContext
+	error ifNil: [ ^ suspendedContext ].
+	(error handleProcessCompletionIfNeeded: self with: aContext) ifNotNil: [ :result | ^ result ].
```
dionisiydk: it's nice. But let's improve it to avoid nil result logic.
By default #handle method will return "self signallerContext" and Halt will do extra logic.

dionisiydk: And maybe method name can be simpler and follow sender name #complete:
```Smalltalk
error completeProcess:  self at: aContext
```

jecisc:
It's indeed nicer.

src/Kernel/Exception.class.st
```diff
@@ -97,6 +97,15 @@ Exception class >> signalIn: context [
 	^ self new signalIn: context
 ]
 
+{ #category : #private }
+Exception >> completeProcess: aProcess with: aContext [
+	"I am a hook during process completion in case one of my subclass wants to plug a special behavior.
+	
+	By default I retrun my signaler context"
```
dionisiydk:
```suggestion
	By default I return my signaler context"
```

dionisiydk (approved):

bencoman:
Just from a static code review, I like it a lot. Very neat.

MarcusDenker (approved):
ok

PR #3365, opened 2019-05-16
3362-RBClassModelFactory-rbClass-should-be-pushed-to-instance-side

### LINKED ISSUE #3362: RBClassModelFactory rbClass should be pushed to instance side
so that we can avoid hardcoding everywhere the RBClassModelFactory.
The same should be done for rbNamespace, rbMethod and rbMetaclass.

In addition RBClassModelFactory should be renamed RBModelFactory.

Ducasse:
Fixes: #3362 
- Introducing RBEntity.
- Making sure that we can use an instance. 
- removing hardcoded class.
- Cleaning tests. 
- Extracting parser class.
- Making sure initialize is invoked.


Ducasse:
I will check since some related tests are failing.

PR #3370, opened 2019-05-17
3368-Move-Halt-tests-out-of-Kernel-tests

### LINKED ISSUE #3368: Move Halt tests out of Kernel tests
Next step to remove Halt from Kernel is to move the tests out of the Kernel tests and move them with debugging packages.

jecisc:
Extract Halt tests to new package Debugging-Utils-Tests.

When all dependencies to Halt from the bootstrap will be removed, I'll propose to move Halt to a package Debugging-Utils.

Fixes #3368


Ducasse:
Broken tests are unrelated

PR #3372, opened 2019-05-17
3369-Context-should-not-depend-on-Halt

### LINKED ISSUE #3369: Context should not depend on Halt
I removed the direct dependencies on Halt in the Kernel, but we still have (at least) one in the packages of the Bootstrap. I'll propose a PR to remove this dependency. 

Long term goal is to remove Halt from the bootstrap since it relies on a compiler that is not present at that time. (And debugging utilities should probably not be in the bootstrap)

jecisc:
Remove dependency between Debugging-Core and Halt. 

Fixes #3369


jecisc:
Tests are passing. Now it just needs a review.

VincentBlondeau (approved):

MarcusDenker (approved):

PR #3374, opened 2019-05-17
3373-Integrate-SUnit-visitor

### LINKED ISSUE #3373: Integrate SUnit visitor
SUnit-Visitor has been implemented some times ago and it allows to reuse visit logic of SUnit.

After a discussion with Steph, Guille and Damien we think it should be integrated.

Actual repository: https://github.com/juliendelplanque/SUnit-Visitor

juliendelplanque:
Added visitor entries in baseline.

Fix #3373


juliendelplanque:
@cdlm @guillep You might be interested in reviewing.

juliendelplanque:
Failing because no space left on CI's disk.
Needs to be restarted once this problem is fixed.

jecisc (changes requested):
Dependencies tests are failing. (You should also maybe merge Pharo8.0 in your branch to fix the announcements tests)

juliendelplanque:
Just merged Pharo8. Looking at dependencies problem.

juliendelplanque:
Not sure how to solve the dependencies problem. Any hints?

Ducasse:
Yes I would like to sit with guille and pablo to see because I also do not get why when I extracted a baseline I got dependencies violation that I could not find with the package analyzer.

Ducasse:
Many for the dependencies it would be something to do for this PR and the one we did with cyril to extract the recentActivity. 
What we could do is to integrate them both one after the other and fix all the dependencies in one shot?
What do you think?

tesonep:
@juliendelplanque these days we can check it. Don't let the impulse die!

juliendelplanque:
Yeah we should I have been a bit busy these last week.

estebanlm:
whoa. Why all this changes?
Maybe worths a merge with Pharo8.0 before?

Ducasse:
We are trying with guille to integrate this change and the problem is that the SystemDependenciesTest are super painful and I do not think that there are really useful. (we get an broken test and a package not loaded instead of a broken build).

jecisc:
I did a cleaner merge in this new PR: https://github.com/pharo-project/pharo/pull/3859

jecisc:
@Ducasse This one can be closed.

PR #3378, opened 2019-05-19
3377-RBRefactoryChangeManager-changeFactory-should-be-nil

### LINKED ISSUE #3377: RBRefactoryChangeManager changeFactory should be nil
I probably broke it. 
There was no test covering it. So I will start adding a test. 
And this logic around ChangeManager and Factory **is** strange. No

Ducasse:
Fixes #3377 but via a side effect. Please help me by reading the issue.


Ducasse:
Broken unrelated tests

PR #3382, opened 2019-05-21
3381-MorphhandlesMouseMove-does--not-works-if-using-an-event-handler

### LINKED ISSUE #3381: Morph>>#handlesMouseMove: does  not works if using an event handler
for some reason `Morph>>#handlesMouseMove:` has commented the verification of event handling (and its commented implementation is buggy).

estebanlm:
Fixes #3381

### DIFF
--- a/src/Morphic-Core/Morph.class.st
+++ b/src/Morphic-Core/Morph.class.st
@@ -3156,7 +3156,7 @@ Morph >> handlesMouseDown: evt [
 
 { #category : #'events-processing' }
 Morph >> handlesMouseMove: anEvent [
-	"Rules say that by default a morph gets #mouseMove if and only if:
+	"Rules say that by default a morph gets #mouseMove if
 		* the hand is not dragging anything,
 			+ and some button is down,
 			+ and the receiver is the current mouse focus."
--- a/src/Morphic-Core/Morph.class.st
+++ b/src/Morphic-Core/Morph.class.st
@@ -3156,7 +3156,7 @@ Morph >> handlesMouseDown: evt [
 
 { #category : #'events-processing' }
 Morph >> handlesMouseMove: anEvent [
-	"Rules say that by default a morph gets #mouseMove if
+	"Rules say that by default a morph gets #mouseMove if and only if:
 		* the hand is not dragging anything,
 			+ and some button is down,
 			+ and the receiver is the current mouse focus."
--- a/src/Morphic-Core/Morph.class.st
+++ b/src/Morphic-Core/Morph.class.st
@@ -3156,12 +3156,14 @@ Morph >> handlesMouseDown: evt [
 
 { #category : #'events-processing' }
 Morph >> handlesMouseMove: anEvent [
-	"Rules say that by default a morph gets #mouseMove iff
+	"Rules say that by default a morph gets #mouseMove if
 		* the hand is not dragging anything,
 			+ and some button is down,
 			+ and the receiver is the current mouse focus."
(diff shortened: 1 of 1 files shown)


akgrant43 (commented):
Hi Esteban,

I'm not sure the "iff" is a typo, "iff" stands for "if, and only if".

Cheers,
Alistair

estebanlm:
it has to be a typo, you cannot expect to write a comment using a special notation and pretend people to understand it :)
if is "if and only if", better to be explicit. 
Also, I do not think it has sense (if is enough there, I think)

estebanlm:
mais voilà :)

akgrant43:
> it has to be a typo, you cannot expect to write a comment using a special notation and pretend people to understand it :)

I know it, so everyone should :-)

> if is "if and only if", better to be explicit.

Agreed.

> Also, I do not think it has sense (if is enough there, I think)

After your change I agree that it shouldn't be there.

Maybe the comment could be:

```
A morph receives #mouseMove if:

- it has an event handler which handles the event, OR
- the hand is not dragging anything,
-- and some button is down,
-- and the receiver is the current mouse focus.
```

akgrant43:
Sorry, our replies crossed paths.

Isn't the "if and only if" incorrect now because it can also be if there is an eventHandler and it accepts the event?

akgrant43:
@estebanlm 
P.S. Sorry about my original comment.  I think it's caused more confusion than help.

estebanlm:
@akgrant43 ok I reverted it ;)

guillep:
Hi @estebanlm seemingly this PR is the cause of #4458 and #5638.
I'd like to know if this should be reverted (until a proper solution is put in place) or not.

Actually, more picky: will reverting this break something else? Because the PR and issue comments are not explicit about that...

If this does not break anything, we should at least revert for Pharo8, as this breaks halos.

Thoughts?

eliohernanph:
Hi @estebanlm, I also support that this should be looked at sooner if possible. I myself was trying to learn Pharo (Smalltalk actually) and more likely one would read Pharo by Example 5 (or maybe 8). Unfortunately, in Chapter 2 (page 13) introduces halos and one will easily get put off that it is not working. This might seem innocent but someone might start questioning the quality of the project. I started looking at Squeak because of this issue but Squeak does not look as pleasant as Pharo. 

I would like to add that I really appreciate all your work!

estebanlm:
However, this is not the cause that is breaking the halos.
And this is an important feature (the ability to attach to events of a morph without needing to subclass it for that), that is used inside the image and also in other projects.
The problem with the halos is somewhere else (probably in some miss-use of this feature), and needs to be fixed were the problem is, not removing a feature that clashes with it.

Ducasse:
@eliohernanph what is the problem with halos because I can get them in P8 and they are working. So I do not really see your point.

Ducasse:
And you can use Pharo without halos :). Believe me.

Ducasse:
@estebanlm this is true that thomas and steven asked you a while ago why you needed this change because at that time it broke the halo rotation.

estebanlm:
I remember the discussion, but I do not remember exactly the context :P

eliohernanph:
@Ducasse, try to move the mouse close to Move icon, it jumps. Note that I just re-installed P8 just now and I can still observe the problem. 

RE: And you can use Pharo without halos :). Believe me.
I agree. But the use case I am planning for needs it :(. I am hoping to create an interactive human machine interface.

PR #3384, opened 2019-05-21
3360-RecentMessageList-does-not-seem-to-receive-events

### LINKED ISSUE #3360: RecentMessageList does not seem to receive events

Ducasse:
Fixes: #3360 adding new baseline


Ducasse:
I do not get how the analysis identifies that Morphic depends on RecentActivity in test  testExternalMorphicDependencies.

in the test self externalDependendiesOf: BaselineOfMorphic allPackageNames
seems to mean that RecentActivity is a dependency of Morphic and I have no idea why. 
I used the dependency analyser and checked one by one the morphic packages and I did not find 
the dependency.

Ducasse:
When I looked at what are the packages depending on RecentActivity I do not see any Morphic. 
So I really do not get from where this dependency comes from. 

For the testUndeclared test, I do not get it since on my machine the test passes.

tesonep:
I have merge Pharo8.0 in this branch, so we can check the problem.

Issue #2865, opened 2019-03-18
The Find dialog is hanging the image


pavel-krivanek:
The find dialog ends in infinite recursion and the image starts to be unresponsive to user interrupt very soon then.

1. open Playground
2. pres Ctrl+F to open the Find dialog
3. start to write someting

StephanEggermont:
Caused by #2789

pavel-krivanek:
the stack:
```1
RubTextFieldArea(RubAbstractTextArea)>>privateReplaceFrom:to:with:
2
RubTextFieldArea>>privateReplaceFrom:to:with:
3
RubTextEditor>>zapSelectionWith:
4
RubTextEditor>>replace:with:and:
5
RubTextEditor>>replaceSelectionWith:
6
[ aStrinbeEditableWhile:ifFalse:self deselect ] ] ] in RubTextFieldArea(RubAbstractTextArea)>>updateTextWith:
7
BlockClosure>>ensure:
8
RubTextFieldArea(RubAbstractTextArea)>>beEditableWhile:
9
[ self  beEditableWhile:ifFalse:self deselect ] ] ] in RubTextFieldArea(RubAbstractTextArea)>>updateTextWith:
10
RubTextFieldArea(RubAbstractTextArea)>>handleEdit:
11
RubTextFieldArea(RubAbstractTextArea)>>updateTextWith:
12
RubTextScrollPane>>updateTextWith:
13
RubPluggableTextFieldMorph(RubScrolledTextMorph)>>updateTextWith:
14
RubPluggableTextFieldMorph>>update:
15
MorphChanged>>deliverTo:
16
RubPluggableTextFieldMorph(Morph)>>handleUpdate:
17
MessageSend>>value:
18
MessageSend>>cull:
19
MessageSend>>cull:cull:
20
[ action cull: anAnnouncement cull: announcer ] in AnnouncementSubscription>>deliver:
21
BlockClosure>>on:do:
22
BlockClosure>>on:fork:
23
AnnouncementSubscription>>deliver:
24
[ "Ensure delivery to remaining announcements" subscription deliver: anAnnouncement ] in SubscriptionRegistry>>deliver:to:startingAt:
25
BlockClosure>>ifCurtailed:
26
SubscriptionRegistry>>deliver:to:startingAt:
27
SubscriptionRegistry>>deliver:to:
28
SubscriptionRegistry>>deliver:
29
Announcer>>announce:
[8:39] Pavel:
[...]

PR #3401, opened 2019-05-24
2865-The-Find-dialog-is-hanging-the-image

pavel-krivanek:
use just annoucement for findPolicy change

fixex #2865

well, this PR has a lot of own issues. The find dialog buttons now do not work correctly etc. But at least, we will be able to use it without killing of our images.


Ducasse:
Tx pavel!

MarcusDenker (approved):
ok

Issue #3064, opened 2019-03-29
Each assert like method in TestRunner should have the dual deny method


juliendelplanque:
Some assert methods do not have their dual deny method.
It should be the case.
With @GDayne we looked and summarise what should be done:

| Assert method  |  Dual deny method |
|---|---|
| `#assert:`  | `#deny:`  |
| `#assert:closeTo:`  | /  |
| `#assert:description:`  | `#denyDescription:`  |
| `#assert:description:resumable` | `#deny:description:resumable` |
| `#assert:equals:` | `#deny:equals:` |
| `#assert:identicalTo:` | `#deny:identicalTo:` |
| `#assertCollection:equals:` | `#denyCollection:equals:` |
| `#assertCollection:hasSameElements:` |  / |
| `#assertCollection:includesAll:` | / |
| `#assertEmpty:` | `#denyEmpty:` |

"/" means the dual does not exist.

Our proposal is to add these missing methods:

- `#deny:closeTo:`
- ~`#denyCollection:hasSameElements:`~ In fact this one is useless, seem comment below.
- `#denyCollection:includesAll:`

Additionally, we propose to add these two methods which can be useful:

- `#assert:closeTo:precision:`
- `#deny:closeTo:precision:`

juliendelplanque:
While implementing this, we realised that in fact adding `#denyCollection:hasSameElements:` makes no sense because it is juste checking that the two collections are different. So we will not add this one.

juliendelplanque:
Bump.

MarcusDenker:
all done

PR #3403, opened 2019-05-24
3064-Each-assert-like-method-in-TestRunner-should-have-the-dual-deny-method

oliveiraallex:
Created new methods as described in https://github.com/pharo-project/pharo/issues/3064


MarcusDenker:
reopen to run with new code

Issue #3376, opened 2019-05-19
Reducing the amount of hardcoded reference to RBParser!!!


Ducasse:
Let us start relying less on classes and more on messages and API!

PR #3408, opened 2019-05-24
3376-Reducing-the-amount-of-hardcoded-reference-to-RBParser

guillep:
Partial fix for #3376.
Removing direct references to RBParser. Extracting them to helper methods in tests.


Ducasse:
Thanks Guille!
I feel lighter :)

PR #3409, opened 2019-05-24
Remove convertInterval:for: of tests of Refactoring

Lin777:
Remove unnecessary and redundant method convertInterval:for:, this only return the interval, put inline the interval in all senders of this method

### DIFF
--- a/src/Refactoring-Tests-Core/RBExtractMethodTest.class.st
+++ b/src/Refactoring-Tests-Core/RBExtractMethodTest.class.st
@@ -7,16 +7,14 @@ Class {
 { #category : #'failure tests' }
 RBExtractMethodTest >> testBadInterval [
 	self
-		shouldFail: (RBExtractMethodRefactoring 
-				extract: (self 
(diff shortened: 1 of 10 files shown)


Ducasse:
I do not get why this is better to inline the definition of conbertInterval:for: 
to me this is regression.

Lin777:
The method is this:
```
convertInterval: anInterval for: aString 
	"Convert the interval to ignore differences in end of line conventions."

	^anInterval
```
I think it was unnecessary, but maybe it was created for some future work. The next one I will create an Issue before, to know if this change is necessary or not

Ducasse:
Ok I was confused reading the code.

MarcusDenker:
the build now has test failing

MarcusDenker:
testInlineComponentIntoCascadedMessage :RBInlineMethodFromComponentTest
testInlineEmptyComponentMethod – RBInlineMethodFromComponentTest

both doesNotUnderstand: #convertInterval:for:

Issue #3416, opened 2019-05-26
Tell, don't ask Rubric, reference to shoutStyler


StephanEggermont:
RubSmalltalkCodeMode>>shoutStyler
	^ self textArea decoratorNamed: #shoutStyler

Ducasse:
Was integrated

PR #3417, opened 2019-05-26
Use the accessor in the textArea for the styler (Rubric)

StephanEggermont:
Fixes issue #3416


StephanEggermont:
failures unrelated

PR #3422, opened 2019-05-27
3420-ChangeSorter-title-window-should-not-be-untitledWindow

### LINKED ISSUE #3420: ChangeSorter title window should not be untitledWindow
I do not know how the system invoke title because this class defines a title method

title
	^ String streamContents: [ :stream | stream << self class title << ' on: ' << self model currentChangeSet name ]

Ducasse:
Fixes: #3420 Better logic, class comment, fixing refresh glitches.

### DIFF
--- a/src/Spec-Tools/ChangeSorterPresenter.class.st
+++ b/src/Spec-Tools/ChangeSorterPresenter.class.st
@@ -1,8 +1,10 @@
 "
-A ChangeSorterApplication is spec version of the dual sorter.
+A change sorter is a little tool to manage changes. 
+Developer can give name to change containers and all the change made during this container is active are collected in the container. 
 
-ChangeSorterApplication 
-	new openWithSpec
+The DualChangeSorter tool supports manipulation of changes between containers and fileout of the changes.
+
+ChangeSorterPresenter new openWithSpec
 "
 Class {
 	#name : #ChangeSorterPresenter,
@@ -32,6 +34,11 @@ ChangeSorterPresenter class >> defaultSpec [
 				add: #methodsListPresenter;
 				add: #textPresenter;
 				yourself)
+]
+
+{ #category : #specs }
+ChangeSorterPresenter class >> title [
+	^ 'Change Sorter'
 ]
 
 { #category : #'menu - change set' }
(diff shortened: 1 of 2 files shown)


src/Spec-Tools/ChangeSorterPresenter.class.st
```diff
@@ -347,6 +354,12 @@ ChangeSorterPresenter >> forgetMessage [
 		self setSelectedClassIndex: self selectedClassIndex ]
 ]
 
+{ #category : #api }
+ChangeSorterPresenter >> initialExtent [
```
jecisc, line 358: We are deprecating this way to set the initial extent of the window. 

The way we will try to enforce is to add a method `initializeWindow:` taking a window presenter as a parameter and to set it to this presenter. The goal is to remove window state from all composable presenters.

jecisc (changes requested):

src/Spec-Tools/ChangeSorterPresenter.class.st
Ducasse: Well how can I know it?
This is not explained anywhere! 
What we should do then is to change all the initialExtent to use initialWindow: 
now I could not understand why the previous initialWindow: method was not doing a super call.

Then I do not understand why self title: 'HJKJL' does not update the window title. 
So this is confusing.

Ducasse: How should I write initializeWindow: 
because I do not get it. When I read the following code I see that the initialExtent is invoked.

So again Spec20 is not correctly communicating with me. 





initializeWindow: aWindowPresenter
	"override this to set window values before opening. 
	 You may want to add a menu, a toolbar or a statusbar"
	
	aWindowPresenter 
		title: self title;
		initialExtent: self initialExtent;
		windowIcon: self windowIcon;
		askOkToClose: self askOkToClose;
		aboutText: self aboutText

Ducasse: then if I should write

```
initializeWindow: aWindowPresenter

    super initializeWindow: aWindowPresenter.
    aWindowPresenter initialExtent: 500@ 350.
```
instead of 
```
initialExtent
    ^ 500@ 350 
```
I do not really understand the gain because the framework puts the burden on 
its user. So to me this is pretty bad design. 
For example I can forget to call super.

Ducasse: Cyril I will integrate it now and we will see.

PR #3425, opened 2019-05-27
3424-MetaLinks-reification-of-index-for-ivars

MarcusDenker:
add #index as a refication for ivar read and assignment


MarcusDenker:
failure not related

Issue #3410, opened 2019-05-24
[Trait] Impossible to use from a classTrait a method that is also defined in a subclass of Behavior even when following the doc


VincentBlondeau:
Following of #3389 

So I try what is written in the doc, and it happens that it fails. I expect that the Trait MyTrait takes the lead on the methods that are defined on the class side.

Example to reproduce the issue:
```
Trait
	named: #MyTrait
	uses: {}
	category: 'TestTrait'.

(Smalltalk globals at: #MyTrait) classTrait compile: 'obsolete2 ^true' classified: 'testing'.
(Smalltalk globals at: #Behavior)  compile: 'obsolete2 ^false' classified: 'testing'.
(Smalltalk globals at: #TraitedClass)  compile: 'obsolete2 ^false' classified: 'testing'.


Object subclass: #MyTraitUser
	uses: (Smalltalk globals at: #MyTrait) 
	instanceVariableNames: ''
	classVariableNames: ''
	package: 'TestTrait'.

self assert: (Smalltalk globals at: #MyTraitUser) class obsolete2
```

VincentBlondeau:
With the previous commit, this assertion is true:
```
self assert: ((Smalltalk globals at: #MyTraitUser) class >> #obsolete) sourceCode
 = 'obsolete ^true'.
```
But sill not this one...
`self assert: (Smalltalk globals at: #MyTraitUser) class obsolete2`

VincentBlondeau:
Actually the good assertion is `self assert: (Smalltalk globals at: #MyTraitUser) obsolete2` and not `self assert: (Smalltalk globals at: #MyTraitUser) class obsolete2`

VincentBlondeau:
Tests still should be added related to #3426 .

Backport have to be done too

VincentBlondeau:
There is still an issue when there is a trait that is already used in the superclass. The subclass does not contain the right method. Test will be added in a PR

VincentBlondeau:
#3906 should be integrated too

PR #3426, opened 2019-05-27
3410-[Trait] Impossible to use from a classTrait a method that is also defined in a subclass of Behavior even when following the doc

VincentBlondeau:
Tests will be added later


VincentBlondeau:
@bot : See #3410

MarcusDenker:
I will close and reopen so it builds with the newest code base

MarcusDenker (approved):

MarcusDenker (approved):
checked with Pablo: we can integrated this for now, but should not forget to improve it later

PR #3429, opened 2019-05-27
3428-Use-more-reasonable-error-when-writing-to-read-only-changes-file

### LINKED ISSUE #3428: Use more reasonable error when writing to read-only changes file
When you open on Windows the image twice, the second image has no write access to the changes file. When you file-in something, you got an error that is confusing for newcomers.

`Instance of ZnCharacterReadStream did not understand #flush`

pavel-krivanek:
safer sources file flush

fixes #3428


MarcusDenker (approved):
ok

PR #3433, opened 2019-05-28
Remove deprecation as #tempNames is used in the build

### LINKED ISSUE #3432: RBMethodNode>>tempnames is used in the build
So should not be deprecated

StephanEggermont:
Resolves #3432


MarcusDenker:
I will do a new one where we do the call from CompiledMethod

PR #3436, opened 2019-05-29
3432-RBMethodNodetempnames-is-used-in-the-build

MarcusDenker:
- use argumentNames and temporaryNames from the ast instead of the deprecated tempNames
- we use #ast instead of #parseTree to fill the AST Cache


MarcusDenker:
error not related

PR #3438, opened 2019-05-29
[Backport]3437-Adding-an-instance-class-variable-remove-the-traits-of-a-class

### LINKED ISSUE #3437: Add test for: Adding an instance class variable remove the traits of a class
If you have a class with traits and use #addClassSlot: on it (or #addSlot: on its instance), it will remove the traits of the class.

jecisc:
Backport fix in class slot addition.

Fixes #3437

### DIFF
--- a/src/Shift-ClassBuilder/ShiftClassBuilder.class.st
+++ b/src/Shift-ClassBuilder/ShiftClassBuilder.class.st
@@ -297,12 +297,12 @@ ShiftClassBuilder >> fillFor: aClass [
 		superclass: aClass superclass;
 		name: aClass getName;
 		layoutClass: aClass classLayout class;
-		slots: aClass localSlots;
-		classSlots: aClass class localSlots;
-		sharedVariables: aClass classVariables;
+		slots: aClass localSlots ;
+		sharedVariablesFromString: aClass classVariablesString;
 		sharedPools: aClass sharedPoolsString;
 		category: aClass category;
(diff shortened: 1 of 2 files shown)


MarcusDenker:
you need to look at fillFor:, too... it was missing setting something in Pharo8

jecisc:
Oh, ok. Just this already fix the problem we had, but maybe there is another one?

MarcusDenker:
ah, the fix there might have been a side effect of removing copyClassSlotsFromExistingClass...

jecisc:
So, what would be the clean solution? Do you have an idea?

jecisc:
I reverted the change on fillFor: since it can cause problem.

MarcusDenker (approved):
Ok

jecisc:
Faling tests are already failing for me in a clean image.

MarcusDenker:
So do we merge this? It is for Pharo7

jecisc:
@tesonep Is that Ok for you?

MarcusDenker:
So what do we do?

MarcusDenker:
If nobody objects, this will be merged end of the week.

jecisc:
We can ask people to review during the sprint

jecisc:
Oups, I closed it by accident because I cleaned my branched and this one had more than 1 month.

MarcusDenker:
I finally merged it.

PR #3454, opened 2019-06-05
2332-groupedBy-does-not-work-for-SortedCollection

MarcusDenker:
add comment


MarcusDenker:
implement #groupedBy: to retain sortBlock for SortedCollectiob
simplify sortBlock (new creates size 10 by default)
turned the example in the issue into a simple test

PR #3455, opened 2019-06-05
2188-Reflectivity-uses-a-Symbol-while-RBVariableNode-a-string-for-SelfNode-name

### LINKED ISSUE #2188: Reflectivity uses a Symbol while RBVariableNode a string for SelfNode name
Two points: 

- Reflectivity uses a Symbol while RBVariableNode a string for SelfNode nameI do not know if this is a problem but it looks like having all the time the same is better. 

- Then I was wondering why RBSelfNode new would not work instead of RBSelfNode named: 'self'
because it means that we can have RBSelfNode named: 'selfandsomethingelse'

generatePrimitiveWrapper
	| wrappedMethod send wrapperMethod assignmentNode |
	OCASTSemanticCleaner clean: ast.
	ast compilationContext 
		semanticAnalyzerClass: RFSemanticAnalyzer;
		astTranslatorClass: RFASTTranslator.
	ast doSemanticAnalysis. "force semantic analysis"
	wrappedMethod := ast generate: compiledMethod trailer.
	
	send := RBMessageNode
		receiver: (RBSelfNode named: #self)
		selector:  #rFwithArgs:executeMethod:
		arguments: {RBArrayNode statements: ast arguments . (RFLiteralVariableNode value: wrappedMethod)}.

MarcusDenker:
- implement #new for RBSelfNode, RBThisContextNode and RBSuperNode
- change all users to use this instead of #named: 

fixes #2188

### DIFF
--- a/src/AST-Core/RBSelfNode.class.st
+++ b/src/AST-Core/RBSelfNode.class.st
@@ -7,6 +7,13 @@ Class {
 	#category : #'AST-Core-Nodes'
 }
 
+{ #category : #'instance creation' }
+RBSelfNode class >> new [
+	^super new 
+		named: 'self' start: nil;
+		yourself.
+]
+
 { #category : #visiting }
 RBSelfNode >> acceptVisitor: aProgramNodeVisitor [
 	^ aProgramNodeVisitor visitSelfNode: self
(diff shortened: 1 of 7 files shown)


MarcusDenker:
Tests are green. It just needs a review

PR #3456, opened 2019-06-05
2175-CharacterisSpecial-is-not-complete

### LINKED ISSUE #2175: Character>>#isSpecial is not complete
the method Character>>#isSpecial defines following possible characters allowed in binary symbols: 

+-/\*~<>=@,%|&?!

This list is not complete because the RBScanner class>>#initializeClassificationTable adds several other possible binary characters:

± · × ÷

It needs to be synchronized.

MarcusDenker:
fix #isSpecial to be in sync with the RBScanner. fixes #2175

### DIFF
--- a/src/Kernel/Character.class.st
+++ b/src/Kernel/Character.class.st
@@ -686,7 +686,11 @@ Character >> isSpacingCombiningMark [
 Character >> isSpecial [
 	"Answer whether the receiver is one of the special characters"
 
-	^'+-/\*~<>=@,%|&?!' includes: self
+	('+-/\*~<>=@,%|&?!' includes: self) ifTrue: [ ^true ].
+	"To be in sync with RBScanner>>#initializeClassificationTable, we
+	check for plus-or-minus, centered dot, times, divide"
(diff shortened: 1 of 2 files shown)


MarcusDenker:
failing test not related

src/Kernel/Character.class.st
```diff
@@ -686,7 +686,11 @@ Character >> isSpacingCombiningMark [
 Character >> isSpecial [
 	"Answer whether the receiver is one of the special characters"
 
-	^'+-/\*~<>=@,%|&?!' includes: self
+	('+-/\*~<>=@,%|&?!' includes: self) ifTrue: [ ^true ].
```
jecisc: Can't we add a method #specialCharacterCodes returning all of those and make RBScanner use it?

It would avoid duplicating the information.

MarcusDenker: yes, I was thinking about that, too...

pavel-krivanek (approved):

Ducasse:
Excellent!
I love to see our libraries getting better and better

PR #3457, opened 2019-06-05
3354-Running-crititz-on-equality-and-hash-function-gives-a-message

### LINKED ISSUE #3354: Running crititz on equality and hash function gives a message
Hello, 

When I use the extra -> generate equal and hash this code is made 

```
= anObject
	"Answer whether the receiver and anObject represent the same object."

	self == anObject
		ifTrue: [ ^ true ].
	self class = anObject class
		ifFalse: [ ^ false ].
	^ minute = anObject minute and: [ hour = anObject hour ]
```

but when running critiz gives this message: 

afbeelding

this one is also tracked here : https://github.com/exercism/pharo-smalltalk/issues/341

where we concluded it is a pharo problem not a exercism problem

MarcusDenker:
remove RBIfTrueReturnsRule as it tells people to make code more comples (nested boolean expressions are much harder to unserstand for humans than multiple early returns!) fixes #3354


MarcusDenker:
error not related

PR #3469, opened 2019-06-06
3468-Slots-use-variable-name-instead-of-Object-new-as-temp

### LINKED ISSUE #3468: Slots: use variable name instead of Object new as temp
the #emitStore: methods of some slots need a space to save a value from the stack. We use "Object new" now, but this is not needed: if we use the same name all the time, we can safely remove it (it is never used in a way that we need a dedicated temp for every use)

We need to use a name, though, that can not be a valid name for a temp. I suggest 0slotTempForStackManipulation

MarcusDenker:
the #emitStore: methods of some slots need a space to save a value from the stack. We use "Object new" now, but this is not needed: if we use the same name all the time, we can safely remove it (it is never used in a way that we need a dedicated temp for every use)

We need to use a name, though, that can not be a valid name for a temp. I suggest 0slotTempForStackManipulation

fixes #3468

### DIFF
--- a/src/Spec-Core/SpecObservableSlot.class.st
+++ b/src/Spec-Core/SpecObservableSlot.class.st
@@ -9,7 +9,7 @@ SpecObservableSlot >> emitStore: aMethodBuilder [
 	"generate bytecode for 'varName value: <stackTop>'"
 
 	| temp |
-	temp := Object new.	"we need a unique Object as a temp name"
+	temp := '0slotTempForStackManipulation'.
 	"We pop the value from the stack into a temp to push it back in the right order"
 	aMethodBuilder addTemp: temp.
 	aMethodBuilder storeTemp: temp.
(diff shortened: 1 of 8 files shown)


MarcusDenker:
error not related

guillep (commented):
I don't think these changes are good.
As the comment say in the old versions, we need unique objects **per compilation** to guarantee there are no conflicts between different temps. Because in case slots are used many times in the same method, this is a recipee for disaster...

@jecisc I think this is related to your problem with Spec. Can you try reverting this?

jecisc:
I don’t think so because the problem is with ObservableSlot that is still using Object new.
I’ll take a look next week because I’m at euratech

src/Spec-Core/SpecObservableSlot.class.st
```diff
@@ -9,7 +9,7 @@ SpecObservableSlot >> emitStore: aMethodBuilder [
 	"generate bytecode for 'varName value: <stackTop>'"
 
 	| temp |
-	temp := Object new.	"we need a unique Object as a temp name"
+	temp := '0slotTempForStackManipulation'.
```
guillep, line 12: @jecisc what about this?

Issue #3472, opened 2019-06-07
Revert #copyUpTo: and #copyUpThrough: are identical in SequencableCollection


tesonep:
#2568 
The methods are different, this should be reverted!!!
Please the comments or the method code when assuming there is a duplication.

PR #3473, opened 2019-06-07
Revert "2568-copyUpTo-and-copyUpThrough-are-identical-in-SequencableCollection"

tesonep:
Reverts pharo-project/pharo#3453


MarcusDenker:
why?

tesonep:
#3472 
The original change was wrong, the methods are doing the same. In one is returning an element more in the collection.

copyUpTo: anElement 
	"Answer all elements up to but not including anObject. If there
	is no such object, answer a copy of the receiver."

copyUpThrough: anElement 
	"Answer all elements up to and including anObject. If there
	is no such object, answer a copy of the receiver."

PR #3477, opened 2019-06-07
Refactoring callbacks

tesonep:
- Extract the callout API to a new object
- Separate the callbacks in backend (the real user of Alien) and the execution in Pharo.
- Adding support for NULL and nil with the correct casting.


MarcusDenker:
testUndeclared failing:

 Found undeclared references: #(#thunk)

tesonep:
Yes, I have seen that. I will fix the callbacks of libgit because they are
accessing directly the callbacks. It has to be refactor also the use of FFi
in libgit, because it is breaking the encapsulation in many places.

PR #3480, opened 2019-06-09
3479-Gofer-UI-seems-unused

### LINKED ISSUE #3479: Gofer-UI seems unused
Running the dependency analyzer on it does not give any dependency and there is nothing in the package to add a menu entry to open some of its content.

I have the impression we can deprecate it and remove it in Pharo 9.

jecisc:
Deprecate Gofer-UI since this package seems unused. 

If someone know of one of its usage please comment to let us know how this is used and in that case we can provide better comments.

Fixes #3479


MarcusDenker (approved):
ok

PR #3486, opened 2019-06-11
3478-Some-slots-cannot-be-assigned-in-a-block

### LINKED ISSUE #3478: Some slots cannot be assigned in a block
Here is a test to show the problem:

```Smalltalk
testIsPossibleToSetSlotInBlock
	| slot writerString |
	slot := #slot1 => ProcessLocalSlot.
	aClass := self make: [ :builder | builder slots: {slot} ].

	writerString := String streamContents: [ :str |
		str 
			nextPutAll: slot name;
			nextPutAll: ': anObject';
			cr;tab;
			nextPutAll: 'true ifTrue: [ ';
			nextPutAll: slot name;
			nextPutAll: ':= anObject ]'.
		].
	aClass compile: writerString classified: 'accessing'.
```

@MarcusDenker It can interest you.

MarcusDenker:
do not forget to pop the value when storing in the temp in #emitStore:

fixes #3478


jecisc (approved):
This fixes my problem!

PR #3494, opened 2019-06-13
3493-ClassOrganizationcopyFrom-should-not-use-instVarNamed

### LINKED ISSUE #3493: ClassOrganization>>#copyFrom: should not use instVarNamed:
instVarNamed: is very slow... we do have an accessor for the remoteString, let's use it

MarcusDenker:
- use accessor #commentRemoteString in #copyFrom:
- rewrite #hasComment to use #commentRemoteString, too (

fixes #3493


Ducasse:
change look good.

PR #3498, opened 2019-06-13
3497-instVarIndexForifAbsent-is-wrong

### LINKED ISSUE #3497: instVarIndexFor:ifAbsent: is wrong
instVarIndexFor:ifAbsent: gives *every* visible slot an offset by iterating over all slots.

This is wrong on many levels:

1) we can have hidden instance variables with offset.
2) if there is a slot with no index in the middle, it will return a different offset than the offset of the slot

Instead of iterating and counting, this method should just ask the slot for the index and execute the failblock for non-indexed vars.

MarcusDenker:
implement instVarIndexFor:ifAbsent: not by interating over all slots and counting, but instead by just asking the slot for it's index.

fixes #3497


Ducasse:
Changes look ok.

PR #3510, opened 2019-06-13
3509-Slots-do-not-use-Association-to-reference-objects-in-compiled-code-but-AdditionalBinding

### LINKED ISSUE #3509: Slots: do not use Association to reference objects in compiled code but AdditionalBinding
We have AdditionalBinding already, we use them when referencing binding from the overlay dict in the compiler.

We can use them, too, for Slots instead of Associations

MarcusDenker:
fix all places where we use an association as a binding and instead use AdditionalBinding

fixes #3509

### DIFF
--- a/src/OpalCompiler-Core/OCASTTranslator.class.st
+++ b/src/OpalCompiler-Core/OCASTTranslator.class.st
@@ -634,7 +634,7 @@ OCASTTranslator >> visitNode: aNode [
 { #category : #'visitor-double dispatching' }
 OCASTTranslator >> visitParseErrorNode: anErrorNode [  
 	methodBuilder 
-		pushLiteralVariable: #error -> anErrorNode asSyntaxErrorNotification;
+		pushLiteralVariable: (AdditionalBinding key: #error value: anErrorNode asSyntaxErrorNotification);
 		send: #signal.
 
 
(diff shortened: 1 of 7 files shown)


MarcusDenker:
CI was green, Benchmarks are red but hard to say why

tesonep:
Hi Marcus, there are regressions in the execution of benchmarks:

Adding An Instance Variable To A Class With A Subclass: got from 1883 ms to 2213.
Create a class using a trait: from 1937 to 2821
Create a Trait that uses other traits: from 673 to 911

MarcusDenker:
it is hard to see how this change impacts the benchmark. I just replace Associations with it's subclass...

MarcusDenker:
and it is just for Slots and Reflectivity, I do not think the code even executes for the benchmark

dionisiydk:
It also changes how LiteralVariables are encoded in methods.
What if there are some method scanning code during recompilation which lookup literals of exactly Assosiation class and when it fails it goes to slow path using AST

MarcusDenker:
Hmm but the changed in  LiteralVariable are only for the reflective read case, or not?

MarcusDenker:
So I checked it again and the code path modified is only related to Reflectivity Reifications and fall-back code generation to global read if we do it reflectively. (call #read )

PR #3512, opened 2019-06-14
3511-browseAllAccessesTofrom-needs-to-use-slots-to-check-isAccessedIn

### LINKED ISSUE #3511: #browseAllAccessesTo:from: needs to use slots to check isAccessedIn:
#browseAllAccessesTo:from: needs to use slots to check isAccessedIn:, else it would only correctly update for instance variable slots

MarcusDenker:
#browseAllAccessesTo:from: needs to use slots to check isAccessedIn:, else it would only correctly update for instance variable slots

fixes #3511

### DIFF
--- a/src/Tool-Base/SystemNavigation.extension.st
+++ b/src/Tool-Base/SystemNavigation.extension.st
@@ -7,12 +7,14 @@ SystemNavigation >> browseAllAccessesTo: instVarName from: aClass [
 
 	"self new browseAllAccessesTo: 'contents' from: Collection."
 
-	| coll |
+	| coll slot |
 	coll := self allAccessesTo: instVarName from: aClass.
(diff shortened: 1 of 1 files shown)


src/Tool-Base/SystemNavigation.extension.st
```diff
@@ -7,12 +7,14 @@ SystemNavigation >> browseAllAccessesTo: instVarName from: aClass [
 
 	"self new browseAllAccessesTo: 'contents' from: Collection."
 
-	| coll |
+	| coll slot |
 	coll := self allAccessesTo: instVarName from: aClass.
```
juliendelplanque, line 11: Can't slots be used here as well?

juliendelplanque (approved):
Except for the comment, looks good to me.

src/Tool-Base/SystemNavigation.extension.st
MarcusDenker: yes, that is the next step :-)

PR #3523, opened 2019-06-14
Categorizes uncategorized classes in Athens-Core

### LINKED ISSUE #3519: Categorizes uncategorized classes in Athens-Core
we should get rid of "Uncategorized" in Athens-Core package

astares:
Fix #3519


MarcusDenker (approved):
ok

PR #3529, opened 2019-06-14
Save item in the toolbar

### LINKED ISSUE #3522: Save item in the toolbar
The "Save" menu item is now the last one, this is odd.

Screen Shot 2019-06-14 at 10 45 42 AM

astares:
- fix save items order and separate save items from quit menu item 
- Fix #3522


bergel (approved):
Looks like an easy catch. Thanks Astares

PR #3544, opened 2019-06-14
Correctly deprecate RGPackage and introduce Deprecated80 package

### LINKED ISSUE #3542: Correctly deprecate RGPackage and introduce Deprecated80 package
- fix issue described in http://lists.pharo.org/pipermail/pharo-dev_lists.pharo.org/2019-June/275467.html
- we should keep a deprecated class RGPackage as subclass of RGPackageDefinition in a package Deprecated80

astares:
Fix #3542

### DIFF
--- a/src/BaselineOfPharo/BaselineOfPharo.class.st
+++ b/src/BaselineOfPharo/BaselineOfPharo.class.st
@@ -15,6 +15,7 @@ BaselineOfPharo >> baseline: spec [
 		
 		spec baseline: 'IDE' with: [spec repository: repository].
 		spec package: 'Deprecated70'.
+		spec package: 'Deprecated80'.
 		spec package: 'Gofer-UI'.
 	]
 ]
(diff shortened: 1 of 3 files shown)


MarcusDenker (approved):
ok

MarcusDenker:
I really don't like RGPackageDefinition. Why not just "Package"?

MarcusDenker:
up, it's RGPackage... ok

jecisc:
This change causes a problem in Moose and all projects using Ring2 because in Ring2 there is a class RGPackage. 

Now using Ring2 conflict with Pharo.

astares:
Mmmhh ... but not having a RGPackage in Pharo (as before) and correctly deprecate for migration would mean to not be able to load projects with Manifests (see original bug description).

One option could be to throw it out of Pharo 8 in full (so that Moose / Ring2) works and then anyone lamenting that his package does not loading on Pharo 8 point to
 - either renaming RGPackage into RGPackageDefinition in its source
 - or use Pharo 7 to migrate the code (although P7 has no RGPackageDefinition class, but one can create a stub there)

pavel-krivanek:
The old Ring was marked as deprecated in Pharo 7 already so we should not be bound by its inconsistent naming conventions

dionisiydk:
Hi.

Problem with bugs from critiques should be fixed on the level of parser of
those manifest rules. So it will aware about RGPackage renaming

сб, 13 июл. 2019 г., 12:25 Astares <notifications@github.com>:

pavel-krivanek:
I will try to solve it with Ring2 integration

PR #3548, opened 2019-06-17
3547-deprecate-instanceVariables

### LINKED ISSUE #3547: deprecate #instanceVariables
The idea behind #instanceVariables was to have a nicer name for #instVarNames... but it is not good: it returns names, not variables! 

-> deprecate now for #instVarNames

MarcusDenker:
deprecate #instanceVariables

fixes #3547


MarcusDenker:
error not related, green on two platforms

PR #3554, opened 2019-06-17
3553-Direct-usage-of-global-variables-World-and-ActiveWorld-is-wrong

### LINKED ISSUE #3553: Direct usage of global variables World and ActiveWorld is wrong
Since World is kind of static and ActiveWorld is the "moveable" target, the indistinct access people does right not on one or the other is incorrect. 
Also, in a context of several worlds being executed at once (like it would be with other backends) this usage is incorrect.

estebanlm:
Fix #3553


estebanlm:
this is not really something no-one can review, @Marcus. 
They are 141 classes that changed :P

PR #3564, opened 2019-06-18
3563-fix-freetype-convertString-in-64bit-zero--NULL

### LINKED ISSUE #3563: fix freetype convertString:... (in 64bit, zero ~= NULL)
Now that we have fixed the mismatch between 0 and NULL in UFFI, we need to fix: 
```Smalltalk
convertString: utf8String 
    len: strlen 
    ofFont: aScaledFont 
    toGlyphs: glyphs 
    numGlyphs: numGlyphs 
    x: x 
    y: y
```

The call is passing zeroes instead NULL and that's making the image crash in certain circumstances.

estebanlm:
Fixes #3563


estebanlm:
the #beDefault method is not from this PR but it will be needed (and it does not harms).
Let's keep it there :)

MarcusDenker:
reopen to run with fixed code

tesonep:
The errors are not related. Fixed later!

PR #3586, opened 2019-06-20
3584-SystemNavigation-use-methodReference

### LINKED ISSUE #3584: SystemNavigation: use #methodReference
we can now use #methodReference to create the wrappers

MarcusDenker:
use #methodReference to create the wrappers

fixes #3584


Ducasse:
marcus what is important also is to add new tests to really cover the changes.

PR #3592, opened 2019-06-20
3591-DisplayScreenboundingBox-references-invalid-clippingBox

### LINKED ISSUE #3591: DisplayScreen>>#boundingBox references invalid clippingBox
Probably as a remain of ancient behaviours, `DisplayScreen>>#boundingBox` uses a clippingBox that is always equal to the `ActiveWorld>>#bounds` value. 

Except in the context of an extra world, where this value is always zero (and hence wrong). 

Removing the clipping will enhance behaviour of extra worlds.

estebanlm:
remove clippingBox since is not used

Fixes #3591

### DIFF
--- a/src/Graphics-Display Objects/DisplayScreen.class.st
+++ b/src/Graphics-Display Objects/DisplayScreen.class.st
@@ -14,7 +14,6 @@ Class {
 	#name : #DisplayScreen,
 	#superclass : #Form,
 	#instVars : [
-		'clippingBox',
 		'extraRegions'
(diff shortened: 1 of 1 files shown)


MarcusDenker:
"Found undeclared references: #(#clippingBox)"

MarcusDenker:
failing test seems not related: testGetForm – Windows32.Zinc.Tests.ZnClientTests

MarcusDenker (approved):

PR #3601, opened 2019-06-21
3600-Iceberg-is-sometimes-unable-to-commit-classes

### LINKED ISSUE #3600: Cannot change class that have an inst var of the same name as a class inst var
I have some cases where Iceberg is not able to commit classes. I have no idea of what is the origin of the bug but I found a fix I'll propose to Pharo.

jecisc:
Fix bug in MCClassDefinition

In some cases we receive a string of variables (expl: 'var1 var2 var3') and we need to transform it into a collection of variable names. 

Fixes #3600


MarcusDenker:
the change sorts instance variables, that is normally not a good idea

jecisc:
This fix is wrong

PR #3603, opened 2019-06-21
3602-Little-extensions-related-to-size-for-RBProgramNode-and-MethodNode

### LINKED ISSUE #3602: Little extensions related to size for RBProgramNode and MethodNode

Ducasse:
Fixes: #3602 with tests because we are real and cool developers! Not like all the guys that do not write tests.


Ducasse:
Guys you are super cool to improve the systems and write tests to cover the new features. I love you.

MarcusDenker:
the Variable Node that is the argument knows it's stop position in the source:

arguments last stop

maybe that is better than calculating it yourself

PR #3611, opened 2019-06-24
3610-Move-extensions-methods-from-Spec-to-Morphic

### LINKED ISSUE #3610: Move extensions methods from Spec to Morphic
Some extensions methods are implemented on Spec but should be in Morphic. 

Some of them are only used in Morphic itself. Some others would better fit in their original class as public API.

jecisc:
Move some methods from Spec to Morphic.

Methods of FTBasicItem can be useful for other parts of the system. I think we can move them directly in the original class.

Method on LayoutFrame are not used by Spec at all, so move them to the package needing them.

Fixes #3610


MarcusDenker (approved):

MarcusDenker:
CI failure not related

PR #3615, opened 2019-06-25
Self, super and thisContext nodes are initialised with their #start v…

### LINKED ISSUE #3608: Decompilation may signal error
Evaluating the below snippet will signal an error in Pharo 8 because the compiled method does not contain any source code (neither in the source files nor in the method trailer). Sending `#ast` or `#sourceCode` will send `#formattedCode` after decompilation, which initiates BlueInk. But BlueInk can't handle AST's without source annotations.
```
(Object class compiler
	source: 'foo self param1: self context';
	compile) ast
```

dupriezt:
…ariable being 0 instead of nil

Fixes: #3608

### DIFF
--- a/src/AST-Core/RBSelfNode.class.st
+++ b/src/AST-Core/RBSelfNode.class.st
@@ -10,7 +10,7 @@ Class {
 { #category : #'instance creation' }
 RBSelfNode class >> new [
 	^super new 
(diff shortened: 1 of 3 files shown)


MarcusDenker (approved):
ok

PR #3643, opened 2019-06-26
3641-remove-OCompletion

MarcusDenker:
remove Ocompletion


myroslavarm (approved):
looks good, removes OCompletion and doesn't break anything else

Issue #3642, opened 2019-06-26
Improve hex byte printing/parsing


svenvc:
There are multiple places in the image where printing/parsing of byte and byte arrays to/from a standard hex representation are implemented.

Cleanup the base implementation to improve code sharing.

svenvc:
https://github.com/pharo-project/pharo/pull/3645

Ducasse:
Better add Fixes: #3645 in your commit it will close automatically the issues on PR integration.

PR #3645, opened 2019-06-26
Improve hex byte printing/parsing

svenvc:
https://github.com/pharo-project/pharo/issues/3642

Add new shared code infrastructure  Integer>>#print[Lowercase]HexByteOn: and Integer class>>#readHexByteFrom: with supporting Character>>#hexDigitValue

Reuse new shared code infrastructure in ByteArray and UUID reading and printing

### DIFF
--- a/src/Collections-Native/ByteArray.class.st
+++ b/src/Collections-Native/ByteArray.class.st
@@ -144,28 +144,12 @@ ByteArray >> hash [
 
 { #category : #converting }
 ByteArray >> hex [
-	"Answer a hexa decimal representation of the receiver"
+	"Answer a lowercase hexadecimal String representation of the receiver"
 		
-	"An alternate implementation was 
-		| result stream |
-		result := String new: self size * 2.
-		stream := result writeStream.
-		1 to: self size do: [ :ix | |each|
-			each := self at: ix.
-			stream
-				nextPut: ('0123456789ABCDEF' at: each // 16 + 1);
-				nextPut: ('0123456789ABCDEF' at: each \\ 16 + 1)].
-  	  ^ result"
-	
-	| string v index map |
-	map := '0123456789abcdef'.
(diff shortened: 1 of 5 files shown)


Ducasse:
Thanks I will integrate it.
What would be super nice is to have some executable comments to illustrate.

juliendelplanque:
Nice, thanks for fixing this issue @svenvc !

ErikOnBike:
Two questions to learn (I highly regard Sven, so please do not see this is 'bad critics'):
In Integer class the following is present `^ (highNibble bitShift: 4) + lowNibble` in `#readHexByteFrom:`. Is there a reason to choose the `+` instead of the `bitOr:`, since the latter is more congruent with the rest of the implementation like `#printHexByteOn:`?
And is there a reason (like performance) to choose `#printHexByteOn:` and `#printLowercaseHexByteOn:` with almost identical implementation?

Regards,
Erik

svenvc:
Erik,

In this case there is indeed no difference between #+ and #bitOr: - somehow #+ feels more natural/simpler to me (there is also no such things as integer arithmetic overflow in Pharo).

There are no performance differences between #printHexByteOn: and #printLowercaseHexByteOn: - they are functionally different though: the first one uses 0123456789ABCDEF as digits, while the second uses 01234556789abcdef - while printing, a choice must be made, while parsing both must be accepted. Assuming that was your question.

Sven

ErikOnBike:
Hi Sven,

Thx for such a quick response. 

Interesting. I had the feeling `bitOr:` was more logical. The #printHexOn: also uses bit manipulation. But both are good. 

Wrt the two print methods, my question was about whether two very similar implementations were chosen and not have for example a (third) single implementation and have both print methods use this third one. It now feels like duplication, but there might be a reason (like performance) to have two similar implementations. That made me wonder. 

Cheers,
Erik

PR #3647, opened 2019-06-27
3646-simplify-allSelectorsInProtocol

### LINKED ISSUE #3646: simplify #allSelectorsInProtocol:
allSelectorsInProtocol: can be written using flatCollectAsSet: and should not sort (that is UI concern)

MarcusDenker:
allSelectorsInProtocol: can be written using flatCollectAsSet: and should not sort (that is UI concern)

fixed #3646


myroslavarm (approved):
looks good!

bencoman:
Approve. Its a fairly shallow senders graph and reviewing all paths, the return values 
are ultimately sent `#do:` so there seems no impact changing from an Array output to a Set output.

PR #3649, opened 2019-06-27
3648-classThatDefinesInstanceVariable-is-taking-names-not-variables

MarcusDenker:
#classThatDefinesInstanceVariable: takes as an arg a name. This should be explicit and follow all the other API about instVarNames.

- add classThatDefinesInstVarNamed: and implement it using the slot API instead of searching names
- rewriting deprecation of classThatDefinesInstanceVariable:
- rewrite users


MarcusDenker:
ci error not related

Ducasse:
Can you add a test for the new method?

Issue #3654, opened 2019-06-27
SystemNavigation: browseAllStoresInto:from: and cleanup


MarcusDenker:
browseAllStoresInto:from: should be implemented like browseAllAccessesTo:from:

in addition, delete some dead code that is unused and makes no sense as a public API

MarcusDenker:
merged

PR #3655, opened 2019-06-27
3654-SystemNavigation-browseAllStoresIntofrom-and-cleanup

MarcusDenker:
browseAllStoresInto:from: should be implemented like browseAllAccessesTo:from:

in addition, delete some dead code that is unused and makes no sense as a public API

fixe #3654


Ducasse:
The changes look ok. Now it was a bit difficult to follow since I got the impression that another changes was included too.

PR #3665, opened 2019-06-27
3664-add-minimal-code-to-have-the-new-code-completion

myroslavarm:
adding test code completion


MarcusDenker:
Ci error not related.

MarcusDenker (approved):
ok

MarcusDenker:
This is an "intermediate" version of the new code completion: it actually does not yet use the AST for getting completion, but the old "Model" classes are reused. (the next step is to replace those).

(with this version we could do not use the special parser nor the byte code based Typing, which is one of the main goals of the project)

PR #3682, opened 2019-06-28
3680-Fuel-serialization-should-null-external-addresses

tesonep:
The addresses should be null before serialization and handled correctly after the deserialization.


MarcusDenker (approved):
ok

PR #3685, opened 2019-06-28
3083-TAssertable-should-be-sync-with-TestAsserter

### LINKED ISSUE #3083: TAssertable should be sync with TestAsserter?
It seems out of sync.

juliendelplanque:
Moved class-side #assert:description: to TAssertable class-side.

Fix #3083 

Need @tesonep to fix the bootstrap of minimal kernel accordingly to our discussion IRL.


jecisc:
The image is not building with those changes.

jecisc:
Is that related to what needs to be done in the bootstrap?

juliendelplanque:
@jecisc I think so

PR #3699, opened 2019-06-29
3698-Improve-LastSpacePosition-comment-in-class-String

### LINKED ISSUE #3698: Improve LastSpacePosition comment in class String
We can add some executable comments and make the comment clearer

Ducasse:
Fixes: #3698 turn comments in executable comments


Ducasse:
Strange I just change comments. So I do not see why it would break.

PR #3703, opened 2019-06-29
3702-SystemDictionary-add-a-cache-for-allBehaviors

### LINKED ISSUE #3702: SystemDictionary: add a cache for allBehaviors
I think if we cache allBehaviors instead of classNames, everything would be much faster.

(first benchmark shows that then it is faster to use "#allBehaviors do:" instead of "allBehaviorsDo:", which means in turn we can then use advanced collection operations (e.g. flatCollect:) to simplify things a lot (and speed up even more)

MarcusDenker:
fixes #3702

### DIFF
--- a/src/System-Support/SystemDictionary.class.st
+++ b/src/System-Support/SystemDictionary.class.st
@@ -24,7 +24,8 @@ Class {
 	#superclass : #IdentityDictionary,
 	#instVars : [
 		'cachedClassNames',
-		'cachedNonClassNames'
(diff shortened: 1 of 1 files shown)


Ducasse:
Not related broken test

Ducasse:
But when we add a class should we not flush this cache?

MarcusDenker:
its is: #flushClassNameCache is called on every change of the SystemDictionary

PR #3715, opened 2019-06-30
Cleanup System-Localization-Tests

### LINKED ISSUE #3714: Cleanup System-Localization-Tests
- use assert:equals: and such

astares:
Fix #3714

### DIFF
--- a/src/System-Localization-Tests/LocaleIDTest.class.st
+++ b/src/System-Localization-Tests/LocaleIDTest.class.st
@@ -22,7 +22,7 @@ LocaleIDTest >> germanLocaleID [
 { #category : #tests }
 LocaleIDTest >> testComparision [
 
-	self deny: self germanLocaleID = self frenchLocaleID.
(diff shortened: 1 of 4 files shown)


src/System-Localization-Tests/LocaleIDTest.class.st
```diff
@@ -23,7 +23,7 @@ LocaleIDTest >> germanLocaleID [
 LocaleIDTest >> testComparision [
 
 	self deny: self germanLocaleID = self frenchLocaleID.
```
jecisc: Why not use #deny:equals: here?

jecisc:
```suggestion
	self deny: self germanLocaleID equals: self frenchLocaleID.
```

astares: This is a deny:

jecisc (approved):

PR #3731, opened 2019-07-01
3730-SystemNavigation-allSendersOf-and-allReferencesTo-simplification

### LINKED ISSUE #3730: SystemNavigation allSendersOf: and allReferencesTo: simplification
allReferencesTo: and allSendersOf: (and allCallsOn:) so all the same

- rewrite to use #allBehaviors and flatCollect
- make allSendersOf: call allReferencesTo: 

unifying the tree methods is then future work

MarcusDenker:
fixes #3730
allReferencesTo: and allSendersOf: (and allCallsOn:) so all the same

- rewrite to use #allBehaviors and flatCollect
- make allSendersOf: call allReferencesTo: (as allCallsOn: does this already)

unifying the three methods is then future work

### DIFF
--- a/src/System-Support/SystemNavigation.class.st
+++ b/src/System-Support/SystemNavigation.class.st
@@ -194,20 +194,12 @@ SystemNavigation >> allPrimitiveMethods [
 ]
 
 { #category : #query }
-SystemNavigation >> allReferencesTo: aLiteral [ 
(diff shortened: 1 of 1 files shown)


MarcusDenker:
[#adsasdf senders] bench

before:  "'16.206 per second'"
after:  "'18.714 per second'"

PR #3736, opened 2019-07-02
Backport of fix for diamond problem in traits with slots.

tesonep:
This fix handles the validation of slots that came from different traits but they are defined in a single trait.

As they are defined once they should not be a conflict.


MarcusDenker (approved):

MarcusDenker:
CI failures are not related

PR #3740, opened 2019-07-02
3394-OverridesDeprecatedMethodRule-should-not-activate-on-deprecated-method

### LINKED ISSUE #3394: OverridesDeprecatedMethodRule should not activate on deprecated method
When a deprecated method is overridden, the rule should not detect a positive result if the detected method is also deprecated because it might be because we want two different rewrite rule depending on the class. 

Example:
image

jecisc:
Do not raise OverridesDeprecatedMethodRule if the method is also deprecated.

Fixes #3394


MarcusDenker (approved):
ok

PR #3749, opened 2019-07-03
3748-Compiler-add-the-concept-of-a-productionEnvironment-enable-for-CompiledMethod

### LINKED ISSUE #3748: Compiler: add the concept of a productionEnvironment, enable for CompiledMethod
When compiling, there might be the case that the result of the compiler (the compiledMethod) should not be an instance of the class CompiledMethod from the environment 
in which we compile, but from some other environment.

(e.g. imagine you have multiple environments and want to cross cross compile)

This PR adds the concept of a productionEnvironment and enables it for CompiledMethod.
(other use cases might make sense, too, for example all the literals in the CompiledMethod, this is future work but can be done similarly by handing the context into the Scanner)

This is our test for CompiledMethod:

testCompileWithProductionEnvironment
	| environment method return | 
	environment := {#CompiledMethod -> MyCompiledMethodForTesting} asDictionary.
	method := Smalltalk compiler 
					productionEnvironment: environment;
					compile: 'tt ^1+2'.
	
	self assert: method class equals: MyCompiledMethodForTesting.			
	return := method valueWithReceiver: nil arguments: #().
	self assert: return equals: 3.

As one can see, the produces compiled method of the Compiler is an instance of our class.

MarcusDenker:
fixes #3748

see the issue for more information

### DIFF
--- a/src/OpalCompiler-Core/IRBytecodeGenerator.class.st
+++ b/src/OpalCompiler-Core/IRBytecodeGenerator.class.st
@@ -75,7 +75,8 @@ Class {
 		'forceLongForm',
 		'primNumber',
 		'encoderClass',
-		'inBlock'
+		'inBlock',
+		'compilationContext'
 	],
 	#category : #'OpalCompiler-Core-Bytecode'
 }
@@ -163,6 +164,11 @@ IRBytecodeGenerator >> closureFrom: fromSeqId to: toSeqId copyNumCopiedValues: n
 	encoder genPushClosureCopyNumCopiedValues: numCopied numArgs: numArgs2 jumpSize: distance. 
 ]
 
+{ #category : #accessing }
+IRBytecodeGenerator >> compilationContext: aCompilationContext [
+	compilationContext := aCompilationContext
+]
+
 { #category : #results }
 IRBytecodeGenerator >> compiledBlockWith: trailer [
 	| lits header cb |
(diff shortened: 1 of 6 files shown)


src/OpalCompiler-Core/IRBytecodeGenerator.class.st
```diff
@@ -203,8 +209,10 @@ IRBytecodeGenerator >> compiledMethodWith: trailer [
 
 { #category : #results }
 IRBytecodeGenerator >> compiledMethodWith: trailer header: header literals: lits [
-	| cm |
-	cm := trailer createMethod: self bytecodes size class: CompiledMethod header: header.
+	| cm cmClass|
+	"we look up the class in the productionEnvironment"
+	cmClass := compilationContext productionEnvironment at: #CompiledMethod ifAbsent: [ CompiledMethod ].
```
dionisiydk, line 214: Like here. Why "compilationContext environment at: #CompiledMethod" is not enough?

dionisiydk: I think I see. 
For example in TelePharo the remote playground doIt's are compiled locally using remote environment for globals. But method is created as local object using local CompiledMethod class. 
At the end method is sent to remote side for execution.

In other scenario we would want to use different method class. For example we could copy CompiledMethod and modify its logic for experiment. Then we will need to use it to test compilation

MarcusDenker: yes, exactly.

dionisiydk: Now I wonder what other objects you are going to retrieve from this productionEnvironment?
Or in other words why simple variable #compilerMethodClass is not enough?

MarcusDenker: The idea is that this might be extended to all the Literal objects that are part of the compiledMethod that is created: Symbol, String, Array, ....

PR #3751, opened 2019-07-03
3739-Rules-override-wrong-method

### LINKED ISSUE #3739: Rules override wrong method
ReAbstractRule seems to be designed so that we override #basicCheck: 

Problem is that some rules override #check:forCritiquesDo: instead and duplicate the code from its superclass.

jecisc:
Override #basicCheck: instead of #check:forCritiquesDo: when possible to avoid duplication.

Fixes #3739


MarcusDenker (approved):
looks good

MarcusDenker:
failing tests not related

PR #3757, opened 2019-07-04
Improving the OSWindow backend for the world

tesonep:
It fixes:
  - The clipboard
  - The CMD/CTRL/ALT + Point
  - Improves mouse and key events.

### DIFF
--- a/src/Multilingual-Encodings/Unicode.class.st
+++ b/src/Multilingual-Encodings/Unicode.class.st
@@ -880,7 +880,5 @@ Unicode class >> ucsTable [
 { #category : #'instance creation' }
 Unicode class >> value: code [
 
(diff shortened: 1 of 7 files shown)


estebanlm (approved):

Ducasse:
self currentWorld worldState worldRenderer
could be 
self worldRenderer
No?

estebanlm:
you mean add a method in Object? I would not pollute it so much. 
maybe we could do instead?
```
self currentWorld worldRenderer
```

estebanlm:
but this is orthogonal to this change :)

Ducasse:
No!!! How can you think that I would like to add a method in Object!
I mean in this class OSWindowClipboard!

Ducasse:
There is no need to duplicate chains of method call.

PR #3760, opened 2019-07-04
3572-SystemNavigationallReferencesToPool-should-be-implemented-on-SharedPool

MarcusDenker:
SharedPool class should be the place to implement allReferencesToPool:
-> just call #usingMethods
-> deprecate the method on SystemNavigation


Ducasse:
We should pay attention to deprecate API. But this is ok for this one. 
And we should be able to access this method from the tools such calypso.

PR #3768, opened 2019-07-04
Lint fix for readEvalPrint

### LINKED ISSUE #3767: Lint fix for readEvalPrint
#readEvalPrint should use streaming instead of ,

astares:
Fix #3767

### DIFF
--- a/src/EmergencyEvaluator/Transcripter.class.st
+++ b/src/EmergencyEvaluator/Transcripter.class.st
@@ -223,7 +223,7 @@ Transcripter >> readEvalPrint [
 				ifTrue: [okToRevert
 						ifTrue: [(self confirm: 'Revert: ' , RecentMessageList uniqueInstance lastEntry asString , ' ?')
 								ifTrue: [RecentMessageList uniqueInstance revertLastMethodSubmission.
-									self cr; show: 'reverted: ' , RecentMessageList uniqueInstance lastEntry asString.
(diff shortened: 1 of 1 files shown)


MarcusDenker (approved):

src/EmergencyEvaluator/Transcripter.class.st
```diff
@@ -223,7 +223,7 @@ Transcripter >> readEvalPrint [
 				ifTrue: [okToRevert
 						ifTrue: [(self confirm: 'Revert: ' , RecentMessageList uniqueInstance lastEntry asString , ' ?')
 								ifTrue: [RecentMessageList uniqueInstance revertLastMethodSubmission.
-									self cr; show: 'reverted: ' , RecentMessageList uniqueInstance lastEntry asString.
+									self cr; show: 'reverted: '; nextPutAll: RecentMessageList uniqueInstance lastEntry asString.
```
nicolas-cellier-aka-nice, line 226: Hmm, it should be the other way around: `nextPutAll:` first, then `show:`, or twice `nextPutAll:` followed by `flush`

astares: Ooops .... yes. It was already merged.

I roll back to the original version of the method and better ban the lint rule. We follow up on this in https://github.com/pharo-project/pharo/issues/3799

PR #3781, opened 2019-07-05
3734-add-usingMethod-for-Behaviour

MarcusDenker:
- add #usingMethods in Behavior 
- rewrite allCallsOn to use it
- improve #usingMethods of SharedPoll to return both the references to the Pool as well as all the vars
-let later factored out as  #methodsAccessingPoolVariables


Ducasse:
marcus to me PoolUsers were classes. So may be we should 
- add classUsingPools returning classes
- change the test testPoolUsers to testMethodsUsingPools

MarcusDenker:
poolUsers still are the classes, usingMethods now includes all methods that reference the pool vars and the pool class itself.

MarcusDenker:
Ah, the test name is wrong: I overrode #testPoolUsers, should have it's own name instead. 

Will fix in a new PR

PR #3800, opened 2019-07-05
Fix mistake from PR3768

### LINKED ISSUE #3799: Fix mistake from PR3768
I did a mistake in https://github.com/pharo-project/pharo/pull/3768 which was already merged.

It's better to revert to the original method version of readEvalPrint.But then we still have the lint error on this method.

With the "," it is easier to read than with a wrong order of #nextPutAll: and show:
- so it is better to ban the rule on this method

astares:
- revert to original version of readEvalPrint
- ban the rule so "," which is more readable can be used
- due to the additional manifest we tag the classes in a class category

Fix #3799


MarcusDenker (approved):

astares:
@MarcusDenker Thanks!

PR #3804, opened 2019-07-06
2943-ToolRegistry-should-drop-support-for-completion

### LINKED ISSUE #2943: ToolRegistry should drop support for completion
- once SmalltalkEditor is removed we will have one less users. 
- we should either remove the handling in TextMorph (see keyStroke:) or add support like we did in RubTextEditor (i.e. by adding an instance variable and passing the completer)

MarcusDenker:
remove #codeCompletion from tools. Side effect is that TextMorph does not support code completion anymore, but that is no problem as we use the Rubric Editor for all code editing

fixes #2943


Ducasse:
This logic was so bad. People were just hacking like mad without thinking. We could have got a simple null object and completion but no conditional and a bit of meta behavior. I'm happy that we threw away.

PR #3807, opened 2019-07-07
3111-Why-do-we-have-in-Object--isCodeCompletionAllowed

MarcusDenker:
- NEController>> codeCompletionAround: textMorph: keyStroke:
  -- simplify
 -- call  on the editor, not the model

- remove isCodeCompletionAllowed and isCodeCompletionAllowed:  everywhere but the editor and editing mode classes

In a next PR we then can continue simplifying


MarcusDenker:
(this might enable code completion in all editors for now, the next PR will restrict it to those where we are  
in code editing mode)

Issue #3822, opened 2019-07-08
return empty model for sequence in test completion


myroslavarm:
this issue for reference : https://github.com/myroslavarm/Experimental-Completion/issues/53

MarcusDenker:
merged

PR #3823, opened 2019-07-08
3822-return-empty-model-for-sequence-in-test-completion

myroslavarm:
adding a fix for sequence


myroslavarm:
issue : https://github.com/pharo-project/pharo/issues/3822

MarcusDenker (approved):

PR #3824, opened 2019-07-08
Don't use an instance variable name as a method argument name.

jgfoster:
Although Pharo appears to allow a method argument name to duplicate that of an instance variable, it is bad style and doesn't port well to other dialects.

### DIFF
--- a/src/FFI-Kernel/ExternalType.class.st
+++ b/src/FFI-Kernel/ExternalType.class.st
@@ -413,14 +413,14 @@ ExternalType >> asPointerType [
 ]
 
 { #category : #converting }
-ExternalType >> asPointerType: pointerSize [
+ExternalType >> asPointerType: anotherPointerSize [
 	"convert the receiver into a pointer type"
 	| type |
 	type := self asPointerType.
-	^type pointerSize = pointerSize
+	^type pointerSize = anotherPointerSize
 		ifTrue: [type]
 		ifFalse:
-			[type copy pointerSize: pointerSize; yourself]
+			[type copy pointerSize: anotherPointerSize; yourself]
 ]
 
 { #category : #accessing }


jecisc (approved):

dionisiydk:
Actually Pharo do not allow it. In interactive mode the browser gives an error in such cases.
But when you need to load code it needs to be loaded. That's why such code could exist (and we need to clean it like you do). Probably it was a result of some refactoring like inst var rename.

Also nice lint rule exists and highlights such cases in the editor.

MarcusDenker (approved):

MarcusDenker:
CI failure not related

astares:
Good catch @jgfoster - and I guess it is not Pharo's fault but this glitch was introduced in Squeak:

1. I did a little research: 

    This method is from "FFI" (the old Squeak foreign function interface) which was originally 
    developed in Squeak and later used in Pharo too. 
     Even when we now have our new "Unified FFI" (UFFI) still a small part of the old FFI 
     (the "FFI-Kernel" and "FFI-Pools") which shared among Squeak, Cuis, Pharo 
      is still part of the Pharo base image.

      Sources of this original FFI are hosted on http://source.squeak.org/FFI.html and the method
      ExternalType>>#asPointerType: pointerSize was introduced by Eliot on 16 February 2016

      I guess because Eliot is primarily using Squeak as development environment where such 
      lint violations are not shown in the browser.       
       
```
Name: FFI-Kernel-eem.31
Author: eem
Time: 16 February 2016, 1:23:41.695311 pm
UUID: f061547b-db79-4d65-86d2-67880645312f
Ancestors: FFI-Kernel-eem.30

Add support for 8-byte pointers.  Add shortPointerAt:[put:] and longPointerAt:[put:] toByteArray and change pointerAt:put: to base its size on ExternalAddress wordSize.  cache the wordSize in a class inst var of ExternalAddress.  Make ExternalAddress update its word size on startUp: (also in the postscript of this version).

Allow ExternalStructure subclasses to implement pointerSIze on the class size and initialize ExternalTypes based on this.  If pointerSIze is non-nil the field definition code will use either shortPointerAt:[put:] or longPointerAt:[put:] as appropriate.

The issue here is that pointer size is baked into structure sizes and we have no convenient way of changing the pointer size, since it exists not just in the compiledSpecd but in the accessor methods and implicitly in any accessor methods after a pointer acessor method.
```

To contribute this fix back also to the original FFI version and the Squeak community in the http://source.squeak.org/FFI.html repo I backported it in: FFI-Kernel-TorstenBergmann.64.mcz there
in case we have to resynch or other later.

2. Yes - @dionisiydk is right. Pharo allows this to be able to load old code and meanwhile Pharo  
    tools like Calypso show such bad code:
 
image

   We put a lot of effort into shaping such better tools and cleaning - but this only goes step by step...

jecisc:
To not see those variables coming back I did this PR: https://github.com/pharo-project/pharo/pull/3839

bencoman:
@astares Thanks for making the effort to sync this upstream

PR #3825, opened 2019-07-08
remove period from end of method names

jgfoster:
While Pharo appears to allow a period at the end of a method name, it doesn't seem to be intentional and doesn't port well to other dialects.


jecisc (approved):
Maybe we should also add a release test to ensure we do not add another in the future. But it should be done in another PR. This one is fine. I’ll add it to my todo list aswell as a test to not have shadowed variables (related to the other PR of James)

MarcusDenker (approved):

PR #3826, opened 2019-07-08
Make Symbol references in literal Array explicit

jgfoster:
While Pharo recognizes punctuation characters (like + and -) to be symbols most of the image makes the array element explicit with the # symbol. This makes it easier to read in other dialects.

### DIFF
--- a/src/Collections-Streams/Stream.class.st
+++ b/src/Collections-Streams/Stream.class.st
@@ -132,7 +132,7 @@ Stream >> next: anInteger [
 
 { #category : #accessing }
 Stream >> next: anInteger put: anObject [ 
-	<compilerOptions: #(+ optionInlineTimesRepeat)>
+	<compilerOptions: #(#+ optionInlineTimesRepeat)>
 	
 	"Make anObject be the next anInteger number of objects accessible by the 
 	receiver. Answer anObject."
--- a/src/Kernel/NonBooleanReceiver.class.st
+++ b/src/Kernel/NonBooleanReceiver.class.st
@@ -9,9 +9,9 @@ Some constructs are optimized in the compiler :
 So you cannot by default use them on non boolean objects.
 	
 If you really need to use optimized constructs, you can enable Opal compiler and do one of the following :
-		- recompile your method with the pragma : <compilerOptions: #(+ optIlineNone)>
+		- recompile your method with the pragma : <compilerOptions: #(#+ optIlineNone)>
 		- recompile your class with the method : MyClass class>>compiler 
-			^ super compiler options: #(+ optIlineNone)
+			^ super compiler options: #(#+ optIlineNone)
 		- call from this method by Object>>#mustBeBooleanInMagic:""
 "
 Class {
--- a/src/Kernel/Object.class.st
+++ b/src/Kernel/Object.class.st
@@ -1512,9 +1512,9 @@ Object >> mustBeBooleanIn: context [
 	So you cannot by default use them on non boolean objects."
 	
 	"If you really need to use optimized constructs, you can enable Opal compiler and do one of the following :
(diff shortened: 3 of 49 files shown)


MarcusDenker:
ci error not related

estebanlm:
I do not like this change.
I don’t know why this is easier for other dialects but the intention on the use of array literals to declare ffi calls follows two guidelines:

- make the declaration as close to its C declaration as possible.
- make it more human readable

Both of this intentions are dismissed with this change.

PLEASE DO NOT INTEGRATE

jgfoster:
@estebanlm , I understand the issue with FFI calls and was hesitant about it (I included the comment "this change does reduce readability"). How do you feel about the other, non-FFI, changes? With the FFI changes removed would you feel comfortable with the remainder?

noha:
I don‘t understand that change. Why is it important to make a difference for + and -. What are the semantics elsewhere not to define a list of symbols with #() ?

jgfoster:
@noha, Technically, #() does not define an Array of Symbol instances, but an Array of literals. And, technically, Symbols are introduced with the pound/sharp/hash symbol (#). Other literals include instances of Character, String, Number, Boolean, UndefinedObject, and Array. Most parsers will treat names (other than true, false, and nil) as Symbols. Some (including Pharo) will go further and treat punctuation as Symbols. But others, such as GemStone, treat +/- as the beginning of a numeric literal. This gives trouble porting code from Pharo. Does that explain the changed?

Other than a concern for readability of FFI (which I accept), is there a reason not to make the remaining changes?

Ducasse:
Can you show examples?

Ducasse:
I saw the changes and indeed it breaks the C feel. 
The integration with C is important for us. 
Could GS have an importer that pretty prints such arrays. Because the syntax of Pharo always allows
#( (1 2)) 
-> #( #(1 2))

The mooc videos show this.

jgfoster:
In GemStone, the following does not compile:
`#( "unexpected token ->" + )`

jgfoster:
I'll submit this again without the change to the FFI packages.

estebanlm:
@jgfoster thanks. 
I'm sorry I was not available last days (I was on the road, in fact I sent my last comment from my phone). 
In particular, I do not want to affect readability and the fact that most of the time, to create an FFI function is just copy&paste the C declaration.

For the others, I would accept it if that would make easier the work of importing pharo into Gemstone :)

estebanlm:
@MarcusDenker since we are here... we could think in a way to avoid the usage of #+ and #- for option style things?

MarcusDenker:
Igor invented this form and I liked it... now if we have to use #+ it looses all it's charm. 

Do people have ideas how to specify options nicely in some other way?

In general, I have to say I do not believe in the "we have to be compatible to other smalltalks"... it just means stopping to innovate.

e.g. what do we do if we want to start to use https://github.com/guillep/parserMOP ?

<blockquote><div> GitHub</div><div><strong><a href="https://github.com/guillep/parserMOP">guillep/parserMOP</a></strong></div><div>Experiments with Parser MOP on Pharo. Contribute to guillep/parserMOP development by creating an account on GitHub.</div></blockquote>

jgfoster:
> In general, I have to say I do not believe in the "we have to be compatible to other smalltalks"... it just means stopping to innovate.

Are there open PRs that you believe will block innovation? Let's discuss it there.

jgfoster:
@estebanlm,

> I'm sorry I was not available last days (I was on the road, in fact I sent my last comment from my phone).

Thank you very much for the follow-up. It helps soften the tone of the original comment. 

> to create an FFI function is just copy&paste the C declaration

Could you add your comments to #3845? I believe it satisfies your request.

PR #3827, opened 2019-07-08
Use accessors rather than direct references for Pragma.selector

jgfoster:
GemStone doesn't have 'selector' as an instance variable and it would be easier for me to port the class if I have to change only two methods instead of five. But if there is a reason to prefer direct variable accesses, that's fine as well. (Note that I'm not arguing against direct access in general, just in this case!)


jecisc (approved):
Looks fine to me.

PR #3829, opened 2019-07-08
3577-GLMRubricExample-presentation-has-unnecessary-tempvar-seg

brackendev:
Remove unnecessary tempvar seg from GLMSimpleRubricExample


astares:
@brackendev Thanks!

PR #3830, opened 2019-07-08
3679-Regression-with-drag-and-drop

### LINKED ISSUE #3679: Regression with drag and drop
PasteUpMorph>>dropFiles:

Looks wrong in P7 and P8.
A Write stream is created while a Read stream is expected.
Result: drag and drop of a sketch file does not work anymore in DrGeo.
Should it be a binary read stream or just a read stream?
With a read stream it works fine for DrGeo, not sure about other.

See https://github.com/pharo-project/pharo/issues/2634 too

[PasteUpMorph-dropFiles.st.txt](https://github.com/pharo-project/pharo/files/3337844/PasteUpMorph-dropFiles.st.txt)

tomooda:
handles file drop with a FileStream instead of a binary write stream. fixes https://github.com/pharo-project/pharo/issues/3679

### DIFF
--- a/src/Morphic-Base/DefaultExternalDropHandler.extension.st
+++ b/src/Morphic-Base/DefaultExternalDropHandler.extension.st
@@ -1,16 +1,17 @@
 Extension { #name : #DefaultExternalDropHandler }
 
 { #category : #'*Morphic-Base' }
-DefaultExternalDropHandler >> handle: dropStream in: pasteUp dropEvent: anEvent [ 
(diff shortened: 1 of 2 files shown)


tomooda:
The default handler will just inspect the FileReference object instead of open a FileList because the FileList class needs a lot of fixes to get it working.

svenvc:
Hi @tomooda 

This looks good: elegant & simple. I haven't actually tried it, but I assume it works.

So: +1

Thx,

Sven

MarcusDenker (approved):

PR #3832, opened 2019-07-09
LanguageEnvironment and Locale[ID]

jgfoster:
LanguageEnvironment is present in a minimal (headless) image, but Locale and LocaleID are not. That seems like an oversight since a headless image could still interact with languages, but that is another issue. For now, I'd like to avoid a load error if the LanguageEnvironment package is reloaded.


src/Multilingual-Languages/LanguageEnvironment.class.st
```diff
@@ -49,7 +49,10 @@ LanguageEnvironment class >> clearDefault [
 { #category : #accessing }
 LanguageEnvironment class >> currentPlatform [
 
-	^ Locale currentPlatform languageEnvironment.
+	| localeClass |
+	"A minimal (headless) image has LanguageEnvironment but not Locale!"
+	localeClass := SmalltalkImage globals at: #Locale ifAbsent: [^self].
```
jecisc:
```suggestion
	localeClass := self environment at: #Locale ifAbsent: [^self].
```

src/Multilingual-Languages/LanguageEnvironment.class.st
```diff
@@ -85,12 +88,14 @@ LanguageEnvironment class >> fileNameConverterClass [
 LanguageEnvironment class >> initKnownEnvironments [
 	"LanguageEnvironment initKnownEnvironments"
 
-	| known |
+	| known localeIdClass |
+	"A minimal (headless) image has LanguageEnvironment but not LocaleID!"
+	localeIdClass := SmalltalkImage globals at: #LocaleID ifAbsent: [^self].
```
jecisc:
```suggestion
	localeIdClass := self environment at: #LocaleID ifAbsent: [^self].
```

src/Multilingual-Languages/LanguageEnvironment.class.st
```diff
@@ -173,9 +178,12 @@ LanguageEnvironment class >> localeChanged [
 
 { #category : #accessing }
 LanguageEnvironment class >> localeID: localeID [
+	| localeIdClass |
+	"A minimal (headless) image has LanguageEnvironment but not LocaleID!"
+	localeIdClass := SmalltalkImage globals at: #LocaleID ifAbsent: [^self].
```
jecisc:
```suggestion
	localeIdClass := self environment at: #LocaleID ifAbsent: [^self].
```

jecisc (changes requested):
It is not possible to access globals with `SmalltalkImage globals`. Instead, we should ask the environment of the class.

noha:
To be honest I'm not sure this means we should remove LanguageEnvironment from the minimal image.

jgfoster:
@jecisc, Thanks for the comments. I've made the corrections requested.

jgfoster:
@noha, Was there something said above that suggested that LanguageEnvironment should be removed?

noha:
@jgfoster Yes. The fact that Locale and LanguageEnvironment appear dependent on each other. And the fact that language settings and minimal do not play well together in my mind

bencoman:
Moved comment to #3844

jgfoster:
@bencoman, Could we move the discussion of the "minimal" image to #3844? I'd like to limit the discussion here to the question of whether it is okay to fix a compile error? That is, the code is broken now since it references missing globals; with my PR at least it compiles. Is there any reason not to merge this PR? What objection does anyone have to this now?

noha:
> The term "minimal" is open to interpretation. There is:

It is not really because http://files.pharo.org/image/80/latest-minimal.zip



<blockquote></blockquote>

noha:
Sorry James for hijacking your pull request :)

bencoman:
> localeClass := self environment at: #Locale ifAbsent: [^self].

I wonder the impact on senders of `#currentPlatform` that they get back the class and not an instance.
The two senders follow up with `class systemConverterClass` and `class filenameConverterClass` which would end up with DNUs.  I think it better to either "fail early", or return a default instance - I'm not sure which would be better but maybe the latter.  In the Full-Pharo7 image I've got open `LanguageEnvironment currentPlatform` returns `a Latin1Environment` and that seems like a reasonable default.

bencoman:
> localeIdClass := self environment at: #LocaleID ifAbsent: [^self].

Same again.  Returning the class is going to cause a failure later on that may be harder to track back to this point.  For this one I think better to fail explicitly here with [self error: '....' ].  Could you move the repeated code to its own method (so there is just one place that #3844 may need to consider later).

jgfoster:
> Could you move the repeated code to its own method

@bencoman, yes that is a good suggestion. Let me know if you have any other thoughts.

src/Multilingual-Languages/LanguageEnvironment.class.st
```diff
@@ -85,12 +88,13 @@ LanguageEnvironment class >> fileNameConverterClass [
 LanguageEnvironment class >> initKnownEnvironments [
 	"LanguageEnvironment initKnownEnvironments"
 
-	| known |
+	| known localeIdClass |
+	localeIdClass := self localeIdClass.
```
bencoman, line 94: Now there is `localeIdClass` method, storing in a variable is a bit redundant. 
Really a minor thing and I approve this PR with or without changing that.

bencoman (approved):

bencoman:
Looks like some cleanup needed on OSX server. Error is...
   Pharo7.0-PR-32bit-bf9d0ab.image:  write error (disk full?).  Continue? (y/n/^C)

src/Multilingual-Languages/LanguageEnvironment.class.st
jgfoster: I don't mind changing it but since it is used inside a loop it seemed appropriate to cache it. When not used in a loop I didn't cache it.

bencoman: That is a fair policy. Leave it as is.

bencoman:
@jecisc @noho, anything else needed for this?

jecisc (approved):
Until we clean the dependencies of this part of the system it's good enough IMO.

jecisc:
Tests passed.

PR #3834, opened 2019-07-09
3811-simplify-and-speedup-unreferencedInstanceVariables

### LINKED ISSUE #3811: simplify and speedup unreferencedInstanceVariables
#unreferencedInstanceVariables is now written as:

	^ self instVarNames reject: [:ivn | 
		self withAllSubclasses anySatisfy: [:class | 
			(class whichSelectorsAccess: ivn) notEmpty]]

as variables are now objects that know if they are used or not, we can do it simpler and faster
by just asking them:

unreferencedInstanceVariables
	"Return the instance variables defined in this class which are not referenced"

	^ self slots reject: [:slot | slot isReferenced]

MarcusDenker:
fixes #3811

### DIFF
--- a/src/Kernel/Behavior.class.st
+++ b/src/Kernel/Behavior.class.st
@@ -1838,9 +1838,7 @@ Behavior >> ultimateSourceCodeAt: selector ifAbsent: aBlock [
 Behavior >> unreferencedInstanceVariables [
 	"Return a list of the instance variables defined in the receiver which are not referenced in the receiver or any of its subclasses."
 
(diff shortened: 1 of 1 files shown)


MarcusDenker:
The speedup will come when we improve #isReferenced for Slot, right now it does the same as the old implementation. But is works for all kinds of slots, not just indexed ivar slots.

jecisc (approved):

jecisc:
Test failure not related.

PR #3835, opened 2019-07-09
3802-make-code-completion-experiment-the-default-for-feedback

myroslavarm:
adding test completion as default

### DIFF
--- a/src/NECompletion/NECController.class.st
+++ b/src/NECompletion/NECController.class.st
@@ -78,8 +78,8 @@ NECController class >> initialize [
 
 { #category : #'tools registry' }
 NECController class >> register [
-
-	RubSmalltalkEditor completionEngine: self.
+	"just for testing: we register the subclass here to test it"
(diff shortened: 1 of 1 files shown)


jecisc (changes requested):
There is a merge conflict.

myroslavarm:
@jecisc yes, i saw, should be fixed now

MarcusDenker (approved):
OK!

astares:
@MarcusDenker 
@myroslavarm 
@pavel-krivanek 
@jecisc 

Can we please roll back this PR #3835 and better put the switch to the TestCompletionController into 
the method "BaselineOfIDE>>additionalInitialization" method which according to what I see is the place where such configuration setups are currently done.

So we would have them all on a single place....

MarcusDenker:
this is just temporary anyway... and reverting would not change anything: the status before this change was that the Completion controller was registered using this method. Yes, can be improved, but that is another TODO.

PR #3837, opened 2019-07-09
3836-Read-only-image-mode

### LINKED ISSUE #3836: Read-only image mode
For running multiple Pharo instances from one image file and for running Pharo from read-only file systems, we need to have the ability to force the read-only mode for the image.

pavel-krivanek:
fixes #3836 

- Introduce ReadOnlyFileException 
- the platform-dependent filesystem Stores have read-only variants (strategies)
- BasicCommandLineHandler accepts new image option: `--readOnlyMode`. Internally, this option is ignored because its value must be known before the command line handler startup is executed. So the real processing of this image argument is done in DiskStore class >> wantsReadOnly. The FileStore read-only or writable strategy is determined using this message
- changes file processing was improved so it now works better with changes file while the writing to it is disabled. The new source code is installed directly into the methods
- Ombu and GTPlayground and the session manager are able to work in read-only mode
- the old streams and File are checking read-only mode (quite ugly, the old mess is even messier now)
- PharoDebug.log and std i/o placeholder files are not created in read-only mode

The read-only mode is just image-based mechanism so it does not prevent the image to modify files for real, the VM still can do it.

### DIFF
--- a/src/DeprecatedFileStream/FileStream.class.st
+++ b/src/DeprecatedFileStream/FileStream.class.st
@@ -291,6 +291,10 @@ FileStream class >> flushAndVoidStdioFiles [
 FileStream class >> forceNewFileNamed: fileName [
  	"Create a new file with the given name, and answer a stream opened for writing on that file. If the file already exists, delete it without asking before creating the new file."
 
+	DiskStore wantsReadOnly ifTrue: [ (CannotDeleteFileException new
+		messageText: 'Could not delete the old version of file ' , fileName) signal.
+		^ NullStream new ].
+
 	^self concreteStream forceNewFileNamed: fileName
 ]
 
--- a/src/DeprecatedFileStream/FileStream.class.st
+++ b/src/DeprecatedFileStream/FileStream.class.st
@@ -291,7 +291,7 @@ FileStream class >> flushAndVoidStdioFiles [
 FileStream class >> forceNewFileNamed: fileName [
  	"Create a new file with the given name, and answer a stream opened for writing on that file. If the file already exists, delete it without asking before creating the new file."
 
-	FileSystemStore wantsReadOnly ifTrue: [ ((CannotDeleteFileException fileName: fileName)
+	SessionManager default currentSession isReadOnlyAccessMode ifTrue: [ ((CannotDeleteFileException fileName: fileName)
 		messageText: 'Could not delete the old version of file ' , fileName) signal.
 		^ NullStream new ].
 
--- a/src/DeprecatedFileStream/FileStream.class.st
+++ b/src/DeprecatedFileStream/FileStream.class.st
@@ -291,7 +291,7 @@ FileStream class >> flushAndVoidStdioFiles [
 FileStream class >> forceNewFileNamed: fileName [
  	"Create a new file with the given name, and answer a stream opened for writing on that file. If the file already exists, delete it without asking before creating the new file."
 
-	DiskStore wantsReadOnly ifTrue: [ ((CannotDeleteFileException fileName: fileName)
+	FileSystemStore wantsReadOnly ifTrue: [ ((CannotDeleteFileException fileName: fileName)
 		messageText: 'Could not delete the old version of file ' , fileName) signal.
 		^ NullStream new ].
 
--- a/src/DeprecatedFileStream/FileStream.class.st
+++ b/src/DeprecatedFileStream/FileStream.class.st
@@ -291,7 +291,7 @@ FileStream class >> flushAndVoidStdioFiles [
 FileStream class >> forceNewFileNamed: fileName [
  	"Create a new file with the given name, and answer a stream opened for writing on that file. If the file already exists, delete it without asking before creating the new file."
 
-	DiskStore wantsReadOnly ifTrue: [ (CannotDeleteFileException new
+	DiskStore wantsReadOnly ifTrue: [ ((CannotDeleteFileException fileName: fileName)
 		messageText: 'Could not delete the old version of file ' , fileName) signal.
 		^ NullStream new ].
 
--- a/src/DeprecatedFileStream/MultiByteFileStream.class.st
+++ b/src/DeprecatedFileStream/MultiByteFileStream.class.st
@@ -561,6 +561,11 @@ MultiByteFileStream >> nextPutAll: aCollection [
 { #category : #'open/close' }
 MultiByteFileStream >> open: fileName forWrite: writeMode [ 
 	| result |
+
+	(writeMode and: [DiskStore wantsReadOnly]) 
+		ifTrue: [ (CannotDeleteFileException new
+			messageText: 'Attempt to open file ' , fileName, ' as wrtitable on read-only filesystem') signal. ].
+
 	result := super open: fileName forWrite: writeMode.
 	result ifNotNil: [
 			converter ifNil: [self converter: UTF8TextConverter new].
--- a/src/DeprecatedFileStream/MultiByteFileStream.class.st
+++ b/src/DeprecatedFileStream/MultiByteFileStream.class.st
@@ -563,7 +563,7 @@ MultiByteFileStream >> open: fileName forWrite: writeMode [
 	| result |
 
 	(writeMode and: [DiskStore wantsReadOnly]) 
-		ifTrue: [ (CannotDeleteFileException new
+		ifTrue: [ ((CannotDeleteFileException fileName: fileName)
 			messageText: 'Attempt to open file ' , fileName, ' as wrtitable on read-only filesystem') signal. ].
 
 	result := super open: fileName forWrite: writeMode.
--- a/src/DeprecatedFileStream/MultiByteFileStream.class.st
+++ b/src/DeprecatedFileStream/MultiByteFileStream.class.st
@@ -562,7 +562,7 @@ MultiByteFileStream >> nextPutAll: aCollection [
 MultiByteFileStream >> open: fileName forWrite: writeMode [ 
 	| result |
 
-	(writeMode and: [FileSystemStore wantsReadOnly]) 
+	(writeMode and: [SessionManager default currentSession isReadOnlyAccessMode]) 
 		ifTrue: [ ((CannotDeleteFileException fileName: fileName)
 			messageText: 'Attempt to open file ' , fileName, ' as wrtitable on read-only filesystem') signal. ].
 
--- a/src/DeprecatedFileStream/MultiByteFileStream.class.st
+++ b/src/DeprecatedFileStream/MultiByteFileStream.class.st
@@ -562,7 +562,7 @@ MultiByteFileStream >> nextPutAll: aCollection [
 MultiByteFileStream >> open: fileName forWrite: writeMode [ 
 	| result |
 
-	(writeMode and: [DiskStore wantsReadOnly]) 
+	(writeMode and: [FileSystemStore wantsReadOnly]) 
 		ifTrue: [ ((CannotDeleteFileException fileName: fileName)
 			messageText: 'Attempt to open file ' , fileName, ' as wrtitable on read-only filesystem') signal. ].
 
--- a/src/DeprecatedFileStream/StandardFileStream.class.st
+++ b/src/DeprecatedFileStream/StandardFileStream.class.st
@@ -82,6 +82,10 @@ StandardFileStream class >> forceNewFileNamed: fileName [
 	asking before creating the new file."
 	| dir  fullName f |
 	
+	DiskStore wantsReadOnly ifTrue: [ (CannotDeleteFileException new
+		messageText: 'Could not delete the old version of file ' , fileName) signal.
+		^ NullStream new ].
+	
 	fullName := self fullName: fileName.
 	(self isAFileNamed: fullName)
 		ifFalse: [f := self new open: fullName forWrite: true.
@@ -534,6 +538,8 @@ StandardFileStream >> open: fileName forWrite: writeMode [
 	"Open the file with the given name. If writeMode is true, allow writing, otherwise open the file in read-only mode."
 	| f |
 	f := fileName asVmPathName.
+	
+	(writeMode and: [ FileSystem disk isWritable not ]) ifTrue: [ ^ nil ].
 
 	fileID := StandardFileStream retryWithGC:[self primOpen: f writable: writeMode] 
 					until:[:id| id notNil] 
--- a/src/DeprecatedFileStream/StandardFileStream.class.st
+++ b/src/DeprecatedFileStream/StandardFileStream.class.st
@@ -82,7 +82,7 @@ StandardFileStream class >> forceNewFileNamed: fileName [
 	asking before creating the new file."
 	| dir  fullName f |
 	
-	FileSystemStore wantsReadOnly ifTrue: [ ((CannotDeleteFileException fileName: fileName)
+	SessionManager default currentSession isReadOnlyAccessMode ifTrue: [ ((CannotDeleteFileException fileName: fileName)
 		messageText: 'Could not delete the old version of file ' , fileName) signal.
 		^ NullStream new ].
 	
--- a/src/DeprecatedFileStream/StandardFileStream.class.st
+++ b/src/DeprecatedFileStream/StandardFileStream.class.st
@@ -82,7 +82,7 @@ StandardFileStream class >> forceNewFileNamed: fileName [
 	asking before creating the new file."
 	| dir  fullName f |
 	
-	DiskStore wantsReadOnly ifTrue: [ (CannotDeleteFileException new
+	DiskStore wantsReadOnly ifTrue: [ ((CannotDeleteFileException fileName: fileName)
 		messageText: 'Could not delete the old version of file ' , fileName) signal.
 		^ NullStream new ].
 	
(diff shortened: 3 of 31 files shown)


pavel-krivanek:
It is not only for the read-only FileSystem but for cases when you want the image to behave as it would be on a read-only system while it is not. 

The `FileSystem isWritable` would be access to a global variable. You should be able to control more filesystem with different settings.

src/FileSystem-Disk/DiskStore.class.st
```diff
@@ -79,12 +85,46 @@ DiskStore class >> maxFileNameLength [
 	self subclassResponsibility 
 ]
 
+{ #category : #public }
+DiskStore class >> readOnlyVariant [
+
+	"return a class variant for read-only store. It should be properly defined in my subclasses"
+
+	^ self
```
jecisc: Shouldn't it be a subclass responsibility or an error `NoReadOnlyStoreAvailableForPlatform`?

pavel-krivanek: I was thinking about subclass responsibility but I wanted to keep compatibility with possible external subclasses.

jecisc: Well, I don't think it will break compatibility because if someone has his own store, then the code to use it will not use your read-only option. 

Then, if the user wants to use the read-only option, he should be notified that his store is not able to do that.

src/DeprecatedFileStream/FileStream.class.st
```diff
@@ -291,6 +291,10 @@ FileStream class >> flushAndVoidStdioFiles [
 FileStream class >> forceNewFileNamed: fileName [
  	"Create a new file with the given name, and answer a stream opened for writing on that file. If the file already exists, delete it without asking before creating the new file."
 
+	DiskStore wantsReadOnly ifTrue: [ (CannotDeleteFileException new
```
gcotelli:
```suggestion
	DiskStore wantsReadOnly ifTrue: [ ((CannotDeleteFileException fileName: fileName)
```

src/DeprecatedFileStream/MultiByteFileStream.class.st
```diff
@@ -561,6 +561,11 @@ MultiByteFileStream >> nextPutAll: aCollection [
 { #category : #'open/close' }
 MultiByteFileStream >> open: fileName forWrite: writeMode [ 
 	| result |
+
+	(writeMode and: [DiskStore wantsReadOnly]) 
+		ifTrue: [ (CannotDeleteFileException new
```
gcotelli:
```suggestion
		ifTrue: [ ((CannotDeleteFileException fileName: fileName)
```

src/DeprecatedFileStream/StandardFileStream.class.st
```diff
@@ -82,6 +82,10 @@ StandardFileStream class >> forceNewFileNamed: fileName [
 	asking before creating the new file."
 	| dir  fullName f |
 	
+	DiskStore wantsReadOnly ifTrue: [ (CannotDeleteFileException new
```
gcotelli:
```suggestion
	DiskStore wantsReadOnly ifTrue: [ ((CannotDeleteFileException fileName: fileName)
```

src/FileSystem-Disk/ReadOnlyMacStore.class.st
```diff
@@ -0,0 +1,102 @@
...
+ReadOnlyMacStore class >> writableVariant [
+
+	^ MacStore
+]
+
+{ #category : #public }
+ReadOnlyMacStore >> basicOpen: aPath writable: aBoolean [
+
+	aBoolean ifTrue: [ 
+		(ReadOnlyFileException new
```
gcotelli: This can be replaced with something like:
```
ReadOnlyFileException signal: 'Attempt to open file ' , aPath pathString, ' as writable on a read-only file system'.
```
By default sending signal: to an exception class with set the messageText

gcotelli (changes requested):
You can use the following rewrite expressions to improve exception signaling:
Convert:
```
(ReadOnlyFileException new messageText: `@messageText) signal
```
to:
```
ReadOnlyFileException signal: `@messageText
```

src/FileSystem-Disk/DiskStore.class.st
pavel-krivanek: ``
```suggestion
	^ self subclassResponsibility
```

dionisiydk:
> It looks horrible.

Sorry, this comment is shown as general statement. I only attached it to the File method with #respondsTo: logic.  
(I used phone and it seems something works different in that case)

pavel-krivanek:
yes, it is horrible, I know. But have you a better idea of how to solve this dependency issue?

src/FileSystem-Disk/ReadOnlyMacStore.class.st
pavel-krivanek: done

src/FileSystem-Disk/ReadOnlyMacStore.class.st
```diff
@@ -0,0 +1,95 @@
...
+		ReadOnlyFileException signal: 'Attempt to open file ', aPath pathString, ' as writable on a read-only file system'.
+		^ self ].
+		
+	^ super basicOpen: aPath writable: aBoolean. 
+]
+
+{ #category : #public }
+ReadOnlyMacStore >> createDirectory: path [
+
+	ReadOnlyFileException signal: 'Attempt crate directory ', path pathString, ' on a read-only file system'.
```
gcotelli, line 35:
```suggestion
	ReadOnlyFileException signal: 'Attempt to create directory ', path pathString, ' on a read-only file system'.
```

src/FileSystem-Disk/ReadOnlyUnixStore.class.st
```diff
@@ -0,0 +1,95 @@
...
+		ReadOnlyFileException signal: 'Attempt to open file ', aPath pathString, ' as writable on a read-only file system'.
+		^ self ].
+		
+	^ super basicOpen: aPath writable: aBoolean. 
+]
+
+{ #category : #public }
+ReadOnlyUnixStore >> createDirectory: path [
+
+	ReadOnlyFileException signal: 'Attempt crate directory ', path pathString, ' on a read-only file system'.
```
gcotelli, line 35:
```suggestion
	ReadOnlyFileException signal: 'Attempt to create directory ', path pathString, ' on a read-only file system'.
```

src/FileSystem-Disk/ReadOnlyWindowsStore.class.st
```diff
@@ -0,0 +1,95 @@
...
+		ReadOnlyFileException signal: 'Attempt to open file ' , aPath pathString, ' as writable on a read-only file system'.
+		^ self ].
+		
+	^ super basicOpen: aPath writable: aBoolean. 
+]
+
+{ #category : #public }
+ReadOnlyWindowsStore >> createDirectory: path [
+
+	ReadOnlyFileException signal: 'Attempt crate directory ', path pathString, ' on a read-only file system'.
```
gcotelli, line 35:
```suggestion
	ReadOnlyFileException signal: 'Attempt to create directory ', path pathString, ' on a read-only file system'.
```

gcotelli:
I have the feeling that all the `ReadOnly*Store` subclasses may be replaced by a generic ReadOnlyStore wrapping the underlying store,  delegating the read-only methods and raising the exceptions for the others.

pavel-krivanek:
Hmm, it is possible but it has other drawbacks. We would need to keep the API in sync when we would extend the abstract store superclasses and it would not work as soon as some platform  store will want to use own extended functionality implemented in new methods

pavel-krivanek:
only benchmark failures

dionisiydk:
> yes, it is horrible, I know. But have you a better idea of how to solve this dependency issue?

Why it needs to be bound to DiskStore? Can it be just a store option? (base class is FileSystemStore)

dionisiydk:
And I agree with @gcotelli . It could be ReadOnlyStore , a wrapper over existing store.
For your argument it's still can ask a store for #readOnlyVariant. But it will be common implementation. And for very special store implementation it could override this method for specific read only option.

src/FileSystem-Disk/DiskStore.class.st
```diff
@@ -79,12 +85,46 @@ DiskStore class >> maxFileNameLength [
...
+{ #category : #current }
+DiskStore class >> resetAsReadOnly [
+
+	"Reset the disk store as read-only. Change indentity of the old store to utilize it in the old users"
+
+	| old |
+	
+	old := CurrentFS.
+	CurrentFS := FileSystem store: self activeClass readOnlyVariant createDefault.
+	old becomeForward: CurrentFS.
```
dionisiydk: I don't think that this functionality really requires special #become: feature from the language.
We don't need to replace existing FileSystem instance if it exist. we can just ask it to become readonly:
```Smalltalk
self currentFileSystem beReadOnly
```
And #beReadOnly will do the work:
```Smalltalk
FileSystem>>beReadOnly
   store := store asReadOnlyStore
```

src/FileSystem-Disk/DiskStore.class.st
```diff
@@ -79,12 +85,46 @@ DiskStore class >> maxFileNameLength [
...
+DiskStore class >> resetAsWritable [
+
+	"Reset the disk store as writable. Change indentity of the old store to utilize it in the old users"
+
+	| old |
+	
+	old := CurrentFS.
+	CurrentFS := FileSystem store: self activeClass writableVariant createDefault.
+	old becomeForward: CurrentFS.
+	^ CurrentFS
```
dionisiydk: And here it would be 
```Smalltalk
self currentFileSystem beWritable
```
And 
```Smalltalk
FileSystem>>beWritable
   store := store asWritableStore
```

src/FileSystem-Disk/DiskStore.class.st
```diff
@@ -99,6 +139,31 @@ DiskStore class >> startUp: resuming [
...
+	| arguments | 
+	
+	arguments := Array
+		streamContents: [ :str | 
+			| arg i |
+			i := 0.
+			[ i > 998 or: [ (arg := Smalltalk argumentAt: i) isNil ] ]
+				whileFalse: [ 
+					str nextPut: arg.
+					i := i + 1 ] ].
```
dionisiydk: I wonder we do not have nice API to access startup options?
It definitely should be extracted to some right place

dionisiydk:
> And I agree with @gcotelli . It could be ReadOnlyStore , a wrapper over existing store.

Really it's not a feeling, it smells. ReadOnly..Store classes duplicate all methods

src/GT-Playground/GTPlayBook.class.st
```diff
@@ -26,6 +26,7 @@ Class {
 GTPlayBook class >> cacheDirectory [
 	" lazy initialize the cache and always ensure its existence - especially while the image is running "
 	cacheDirectory ifNil: [ cacheDirectory := self defaultCacheDirectory ].
+	cacheDirectory fileSystem isWritable ifFalse: [ ^ cacheDirectory ].
```
dionisiydk, line 29: Why you do not use existing fileReference API isReadable/isWritable?
We should not expose internals of fileReference.

dionisiydk: They of course require some changes to support "forced readOnly file systems"

dionisiydk: Following and improving real FileReference API will benefit us to fix installation to really read only OS directories

src/GT-Playground/GTPlayPageFilePersistence.class.st
```diff
@@ -6,7 +6,12 @@ Class {
 
 { #category : #private }
 GTPlayPageFilePersistence >> deleteFromFileSystem [
-	self fileReference ensureDelete
+
+	| aFileReference |
+	
+	aFileReference := self fileReference.
+	(aFileReference notNil and: [ aFileReference fileSystem isWritable ]) ifTrue: [ 
+		aFileReference ensureDelete ]
```
dionisiydk, line 14: Also an idea (maybe not for this work). 
It would be really nice to avoid such checks in all places where fileReferences are using.
Specially configured file reference could be a deaf object and ignore all write accesses. So users like this one will only need to check location of cache in one place and replace it with something like
```Smalltalk
cacheDirectory := cacheDirectory asIgnoredAnyWrites.
```

src/GT-Playground/GTPlayPageFilePersistence.class.st
```diff
@@ -21,6 +26,11 @@ GTPlayPageFilePersistence >> fileName [
 
 { #category : #'accessing-dynamic' }
 GTPlayPageFilePersistence >> fileReference [
+
+	self fileDirectory ifNil: [ 
+		"the directory was not created (read-only filesystem etc.)"
+		^ nil ].
```
dionisiydk, line 32: And special object would avoid such horrible nils

pavel-krivanek:
> Why it needs to be bound to DiskStore? Can it be just a store option? (base class is FileSystemStore)

OK, I moved the wantsReadOnly to the FileSystemStore and use it directly

src/GT-Playground/GTPlayPageFilePersistence.class.st
pavel-krivanek: Maybe but I would solve it in a different PR because this may have strange consequences when data may be lost without any notice

src/GT-Playground/GTPlayBook.class.st
pavel-krivanek: The problem in some usages of the FileReference is that isWritable on has different semantics. It requires an existing referenced file while we need to check if we will be able to create such file/directory. Of course, a proper method in FileReference API would be better (canBeCreated?)

src/GT-Playground/GTPlayPageFilePersistence.class.st
pavel-krivanek: A special file reference would be a significant overkill because such then you need a special write stream that does nothing for real too etc. At least, renamed this method to fileReferenceOrNil to explicitly mention the possible special case.

src/FileSystem-Disk/DiskStore.class.st
pavel-krivanek: > And I agree with @gcotelli . It could be ReadOnlyStore , a wrapper over existing store.
> For your argument it's still can ask a store for #readOnlyVariant. But it will be common implementation. And for very special store implementation it could override this method for specific read only option.

I tried to implement it. We have two options. One is to use MNU mechanism which I would like to avoid if it is not really necessary. The second option is to implement the methods explicitly with the resending to internal store. Which means to resend not only DiskStore methods (57) but FileSysstemStore methods (71). That would be horrible mess, definitelly worse than in case of separate read-only store classes. And you would mix two design patterns together. The optimal solution would be a Trait usage but we cannot use it

src/GT-Playground/GTPlayBook.class.st
dionisiydk: Yes for #canBeCreated

src/System-SessionManager/SessionAccessModeResolver.class.st
```diff
@@ -0,0 +1,49 @@
...
+]
+
+{ #category : #'class initialization' }
+SessionAccessModeResolver class >> resolve [ 
+
+	"check if the image was started in the read-only mode"
+
+	| arguments mode | 
+	
+	arguments := Smalltalk argumentsStartingAtIndex: 0.
```
dionisiydk, line 28: I am not against your approach.
But I assumed it will be something simpler directly in session creation method:
```Smalltalk
 SessionManager>>newSession
	| aWorkingSession |
	aWorkingSession := WorkingSession new.
	aWorkingSession manager: self.
	aWorkingSession readPropertiesFrom: (self environment argumentsStartingAtIndex: 0).
	^ aWorkingSession
```

pavel-krivanek: I think that as part of the startup list it will be more clear and more fitting into the current startup architecture and with a standalone class, we can extend the responsibility of it. In the end, the access mode may be a standalone class(es)

dionisiydk: Ok. 
But my idea was also introduce nice API to query startup options.
With this approach it still not exists.
So the possible "next StartupOptionResolver" will again do the anySatisfy loops with arguments.

src/System-SessionManager/SessionAccessModeResolver.class.st
```diff
@@ -0,0 +1,49 @@
...
+	mode := #readWrite.
+
+	(arguments anySatisfy: [ :each | each = '--readWriteAccessMode' ])
+		ifTrue: [ mode := #readWrite ].
+	(arguments anySatisfy: [ :each | each = '--readOnlyAccessMode' ])
+		ifTrue: [ mode := #readOnly ].
+	(arguments anySatisfy: [ :each | each = '--writeOnlyAccessMode' ])
+		ifTrue: [ mode := #writeOnly ].
+	(arguments anySatisfy: [ :each | each = '--disabledAccessMode' ])
+		ifTrue: [ mode := #disabled ].
```
dionisiydk, line 39: I don't want to be annoying :). But is not it too premature implementation of those modes? Here we need only one.

If if we will have many of them it could be better to reify real objects instead of symbols. So we can put behaviour to them

pavel-krivanek: we have two so currently, it is only two more ;-) When I was writing the resolver, I just wanted to make it complete. I had the first-class access mode in mind but it looked not important to me

dionisiydk (approved):
Let's proceed. We can always improve later

PR #3839, opened 2019-07-09
3838-Add-release-test-to-ensure-we-do-not-have-shadow-variable-in-the-image

### LINKED ISSUE #3838: Add release test to ensure we do not have shadow variable in the image
To ensure what James removed is not introduced again.

jecisc:
Add release test + remove shadowed variables.

This PR will not be broken until Tonel is fixed :(

Fixes #3838

### DIFF
--- a/src/Clap-Core/ClapContext.class.st
+++ b/src/Clap-Core/ClapContext.class.st
@@ -194,8 +194,8 @@ ClapContext >> initialize [
 { #category : #activation }
 ClapContext >> interpret: argumentsSequence [
 	^ self arguments: argumentsSequence;
-		ifMatch: [ :match |
-			match value: self.
+		ifMatch: [ :matched |
+			matched value: self.
 			self exitSuccess ]
 		ifMismatch: [ self noneMatched ]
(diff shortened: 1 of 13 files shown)


jecisc:
This PR requires the integration of a Tonel version with this fix: https://github.com/pharo-vcs/tonel/pull/79

MarcusDenker:
The problem is that we now have to wait for the already fix to show up in Pharo8.

jecisc:
Yes, I know :( But I still did the PR otherwise I would have forgotten it.

MarcusDenker:
could you split this into two: the test only and all the cleanups?

The test is so small then that it could be part of the issue, not the PR.

(PRs need to be actionable, else they just use up space here...)

MarcusDenker:
This shows again that we can not afford to rely on code in the image where trivial fix takes weeks to do

jecisc:
Here is the first step: https://github.com/pharo-project/pharo/pull/4193

When it will be integrated we can merge Pharo8 in this PR and get only the test.

PR #3843, opened 2019-07-09
Make a few binary selectors in literal Array into explicit Symbols

jgfoster:
This pull request modifies a few literal arrays in the "minimal" (headless) image so that they have explicit Symbol references for single-character punctuation (primarily +/-). This change allows GemStone's VM to parse this Pharo source.


bencoman:
Hi James, I see recently you make a few changes for Gemstone compatibility, which I support and think is exciting, but this one seems to imply inconsistent handling of elements within an array literal.  It has always seemed to me that the hash  '#' is mathematically-distributed-through-the-brackets to "all" elements.  It seems VisualWorks 8.3 is the same where inspecting `#(- a b)` gives three elements:
- #- 
- #a 
- #b
Does Gemstone not do this?  if not, should it? My concern is how many future places this may need to corrected in Pharo and also user's application code.  Can Gemstone consider adapting here?
(P.S. I don't want to discourage your contributions, just asking the question and curious what your constraints are.)

jgfoster:
@bencoman, 

> It has always seemed to me that the hash '#' is mathematically-distributed-through-the-brackets to "all" elements. 

The rule is not quite so simple as that. Consider #(true 42 $a). Does that give three symbols, #'true', #'42', and #'$a'? No, it gives three non-Symbol literals, a Boolean, a SmallInteger, and a Character.

> Does Gemstone not do this? if not, should it? ... Can Gemstone consider adapting here?

I appreciate Pharo's role as an innovator and GemStone has, in many cases, adopted Pharo's syntax to simplify portability (we even allow the underbar (_) for assignment!). My situation is this is a private project and I'm trying to move as quickly as possible to _discover_ what issues exist. 

I could, each time I find a roadblock, take the weeks necessary to modify GemStone and seek internal approval for that change and wait for it to be built into the product. Or I could take a couple hours to modify a few Pharo methods to further my exploration and discover the next issue. (Perhaps I should do this on a fork of Pharo, but I'd really like to keep up and not take the time to constantly reapply my changes.)

> My concern is how many future places this may need to corrected in Pharo and also user's application code.

I invite the Pharo team to view my proposed changes as limited to the places where I made them. With these changes I'm able to compile 1530 classes and ~25,000 methods into GemStone. There are many (thousands?) of other places in Pharo code that I'm not changing. Would they need to change eventually? Probably not. Think of what I'm doing as an _exploration_ with various possible outcomes, including the following:

* I abandon the project and nothing more is asked of Pharo;
* I eventually demonstrate the ability to run a "minimal" image and those who want to use GemStone can "downgrade" their code to conform to GemStone's syntax; or
* The project is such an overwhelming success that GemStone's management is convinced to completely adopt every new Pharo feature so that everything is 100% portable.

Naturally, we hope for the last outcome, but without knowing what issues will arise it is difficult to get there. My approach has been to "work around" the issues in the simplest way possible (without major compromise). As we learn more we can refine the process and modifications to GemStone are a likely part of that refinement.

> P.S. I don't want to discourage your contributions, just asking the question and curious what your constraints are.

Thanks. I appreciate that very much. I hope the above explanation helps.

astares:
@jgfoster I guess anyone is appreciating that you value Pharo as good enough to consider it for your Gemstone integration experiment. We all agree that it could be of value for both sides - Gemstone and Pharo.

But why should we change Pharo already as you go along when you still check if it is possible? You are still in the phase of experimenting, no?

I already proposed very easy ways for doing custom experiments (as one can read in https://github.com/pharo-project/pharo/pull/3845) with a local effect on your side

- for instances you follow Norberts approach from https://github.com/noha/pharo-minimal
  you just have to fork this project, create a special dir for "gemstone" and apply your patches 
    
- you can also just fork Pharo in git (which you already did) and boostrap and work in your own 
  fork without any effect on Pharo yet - but keeping up with Pharo's movement

It does not make sense to steer Pharo into a specific different when you still evaluate feasability. I hope you can understand this point of view.

If you need guidance I'm pretty sure the overall Pharo community will support. The more we understand goals and problems on your side the better we can help.

<blockquote><div> GitHub</div><div><strong><a href="https://github.com/noha/pharo-minimal">noha/pharo-minimal</a></strong></div><div>A set of scripts and resources to produce a pharo minimal headless image - noha/pharo-minimal</div></blockquote>

jgfoster:
@astares, 

> It does not make sense to steer Pharo into a specific different

Are you objecting to anything specific in this PR, or just suggesting that this change doesn't further your goals and that I play somewhere else?

> you can also just fork Pharo in git (which you already did) and boostrap and work in your own
fork without any effect on Pharo yet

Yes, that looks like it would be faster.

astares:
For PR #3845 I clearly I was clearly objecting due to the introduction of incompatibilities in UFFI calls. But as I said others should comment as well. Pharo is not me alone. There are many people doing work, different interests and ideas.

For this PR #3843 I can only say that out of personal preference 
```Smalltalk
<compilerOptions: #(+ optionInlineTimesRepeat)>
``` 
looks easier to read than
```Smalltalk
<compilerOptions: #(#+ optionInlineTimesRepeat)>
``` 

If you want to move fast in a private project as you said and dont want to wait for Gemstone changing or getting headwind from Pharo side I personally would first finish the experimental stage to clarify the technical feasability and use a custom private setup for the time being. 

If all technical feasability is clarified I would propose all necessary changes and put the additional weight of a working gemstone integration into the discussion and see if there is enough interest to move changes into the base Pharo or keep it separately in a custom setup. 

But at current stage as you said yourself it is private, experimental and not clear if there will be show-stoppers or other. So I think it is too early to propose and decide. But again: this is only my personal opinion and does not necessarily reflect the community thinking. If you want other opinions too it would be good to discuss this on the Pharo mailinglist.

jgfoster:
> If you want other opinions too it would be good to discuss this on the Pharo mailinglist.

Thanks, but _no_. This has been more than enough discussion.

bencoman:
> I hope the above explanation helps.

yes, that was what I was looking to understand.

> > you can also just fork Pharo in git (which you already did) and boostrap and work in your own
fork without any effect on Pharo yet

> Yes, that looks like it would be faster.

In the git world, this is really a valid workflow.  Certainly you eliminate friction but on the flip side you risk building up technical debt.  Not to give free reign on changes**,  but while I don't think this is a great PR, with the FFI changes reverted I think it is actually low impact and the benefit for both sides for James to continue working on Trunk outweighs the negative.  So I propose this proceeds with the request that:
* a ticket is opened marked somehow for Pharo8-Release to review this
* in parallel James follows up with Gemstone to see if they can parse the original code prior to Pharo8-Release.
Is that amenable to both sides @jgfoster? @astares?

** Hopefully even with some pushbacks we can still resolve things quicker than a corporate body.
Lets work together.

jgfoster:
It looks like most of these are are <compilerOptions> and <metaLinkOptions>. I'll take some time away from working on porting primitives and see if I can filter these general cases. I may reopen this with a smaller list.

PR #3845, opened 2019-07-09
Use string literal to define a few FFI calls 

jgfoster:
See existing code/comment in FFIFunctionParser>>parseNamedFunction: for the pre-existing support for this approach. This small change avoids inconsistencies in literal Array parsing in other dialects for the "minimal" (headless) image while preserving the readability of function definitions.


astares:
@jgfoster Can you please ellaborate a little bit more on what you are trying to solve/introduce
and what inconsistency you mean specifically? 
To me this PR would introduce an inconsistency (instead of removing one) into the standard image as we would use two different ways to do the UFFI calls after applying it.

**To explain:**
Yes - according to the comment of FFIFunctionParser>>parseNamedFunction: it is possible to use
 a) literal array representation
 b) string representation

but the first one is the most common and default. Dont know if and where the second one is ever used. 

The fact is:
- all UFFI calls within the standard image (be it in WinPlatform that you want to change now, Athens, FreeType2 bindings and many other) follow the literal array representation 
- same literal array representation is also used in projects like Iceberg (see LibGit binding) that are externally managed in GitHub but used as part of the Pharo image build after Bootstrap
- many external projects and packages (be it on GitHub or in Catalog) also use this presentation
   when it comes to the UFFI calls
- it is used in the docu, tests and examples of UFFI itself 

To be honest I'm currently not aware anyone directly using the second form.

So now applying the proposed changes would mean to introduce an inconsistency into the standard
image as WinPlatform would use the string form while any other citizen in town would use the literal array representation for UFFI calls. So this does not make much sense.

So with the informations given so far combined with the current situation I would vote for not including the PR. Maybe you can give more details and other can comment too.

jgfoster:
@astares I'm trying to load the Pharo "minimal" (headless) image into GemStone (see my ESUG 2017 [presentation](https://programminggems.files.wordpress.com/2018/08/pharo-in-gemstone-james-foster-2017-09-esug.pdf)). Unfortunately, the parsing of literal arrays is different in GemStone so that approach doesn't work (and modifying the Pharo literal arrays was even more of a problem). Since this involves 1530 classes and 26,233 methods, I'd like to find a way to simplify the process so that I can dream of finishing in my lifetime (years instead of decades). With the PRs I've submitted yesterday and today I can compile all but five methods (instance var names in just two classes!) in GemStone. Yes, there are other ways this could be done (without modifying a few Pharo methods), but they would involve substantially more effort and would not move us closer to our goal (please talk to Stef if you don't see any value in this project).

Ducasse:
James why not specializing the prettyprinter to emit code that can be loaded in GS. 
Because you will have problem with Slots and others too. No?
BTW I would like to see your work in a nice article :)
I did that approach long time ago when I wanted to load Squeak code in VW.

jgfoster:
@Ducasse I'd very much like to write up what I'm doing in an article, but I'll probably need some guidance (or even a collaborator)! As to the prettyprinter, could you give me some sample code that would adjust the array literals? Thanks!

MarcusDenker:
"pretty printing" slots would be hard... for some it might be possible (e.g. inlining the code of the read operation at all the reads of the variable), but for others it might not work ever. E.g. if you have code that relies of the fact that variables defined in classes are objects, not strings...

My personal goal is to improve Pharo in a way that we can do things that are just impossible to do if you do not have these improvements...

bencoman:
> With the PRs I've submitted yesterday and today I can compile all but five methods (instance var names in just two classes!) in GemStone. 

Very cool!!!  A temporary workaround would be to manually edit this exported file.
Presuming you are scripting the Pharo-export/Gemstone-import, you could md5 fingerprint the exported file to be alert for changes, and if unchanged copy your `WinPlatform.class.st(gemstone)` over the top of it.

jgfoster:
@MarcusDenker,

> My personal goal is to improve Pharo in a way that we can do things that are just impossible to do if you do not have these improvements...

What are "these improvements"? In what is my proposed change incompatible?

astares:
@jgfoster Please do not argue with things like "my project is important so it need to get in". For sure Pharo as another Gemstone hosted environment is an interesting project - at least for a part of the community or Gemstone customers.  

But lets discuss the changes really independent from this fact to be more objective and focus on the PR and code changes proposed.

I have already written that these changes would introduce inconsistencies to the (standard) image and all other places UFFI calls are used.

We already invested and still invest a lot in Pharo to provide a **clean** and **consistent** environment even shaping the next step beyond Smalltalk. This was and still is our primary goal after forking from Squeak. We would like to move forward with Pharo and should not do steps backward - also not in the quality we accept. That's why I still vote for not including it.

This PR might save you some minutes for the Gemstone integration project. But I agree with the others and think that this will end up in more problems over time - also more and more conflicts. Other UFFI calls might be added again using the standard literal array notation, other topics might be comin in the area of Slots, Traits, ...
 
So the basic question is what is the solution to let Pharo and the community focus on its original and future goals but also help you with having a "minimal Gemstone integrateable Pharo image".

**Currently I see several options you might want to evaluate:**

Since Pharo 7 the community is using Bootstrap to build a (default) minimal image and the standard Pharo IDE image.

 1. When you want to use the already prebuilt minimal image of Pharo (Pharo-minimal.image) then you
     might want to explorer the path to modify it the way you need and save it as a custom
     "Gemstone-Pharo- minimal.image".
	
    So while the mini image is built up in Bootstrap process you can further adopt, stripe or load 
    add-ons at will to create a custom one based on that.
	
	This link might give you some infos how to do that:
	https://despotadesdibujau.github.io/southsmalltalks/how-to-deploy-in-pharo.html
		
 2. Norbert a few weeks back also published an interesting project he uses to build own mini headless
     images using scripting and patches.
     
     The project is on GitHub (https://github.com/noha/pharo-minimal) and Norbert automized the
     creation of minimal images for Seaside or a remotely controllable TelePharo image. 
    
     Basically these are shell scripts downloading the minimal image and applying patches to it or loading 
     other code on top.
	
      You may want to have a look and follow a similar approach to build a custom image based on the 
      minimal one where the proposed changes from this PR or other necessary code changes could be 
      easily applied.
	
	Would be easy to add a "gemstone" folder with a "patches-gemstone" including the *.st files 
        including the changes on WinPlatform.

 3. The FFI calls changed here are primarily used for getting environment variables on Windows 
     environment.
     If you know that the Gemstone and image is not run on Windows you can also follow the approach 
     to throw classes like WinPlatform and subclasses out of the minimal image. This can also be easily
      scripted / applied as a step after 
	downloading the minimal image.
 
If you do not want to base directly on the prebuilt image but more on the source code level
you can also follow this path:
 
 4. Pharo is now using git for SCM and is hosted on GitHub. You already have your own fork at 
     https://github.com/jgfoster/pharo
     So what you can do is to play and host your necessary changes in your forked GitHub repo.
	
    You could for instance have a branch "Pharo8-Gemstone" and apply the necessary changes for 
    WinPlatform. Then build a custom minimal image using this branch. If you need latest additions 
    of regular Pharo development you just need to pull the changes in using git and synch. 

Approach number 2 is easiest. Approach number 5 might be easiest in managing the changes necessary
as git can show you easily when a merge conflict appears and other.
	
	





<blockquote><div>southsmalltalks</div><div><strong><a href="https://despotadesdibujau.github.io/southsmalltalks/how-to-deploy-in-pharo.html">Build your application from the Pharo-minimal image.</a></strong></div></blockquote>
<blockquote><div> GitHub</div><div><strong><a href="https://github.com/noha/pharo-minimal">noha/pharo-minimal</a></strong></div><div>A set of scripts and resources to produce a pharo minimal headless image - noha/pharo-minimal</div></blockquote>
<blockquote><div> GitHub</div><div><strong><a href="https://github.com/jgfoster/pharo">jgfoster/pharo</a></strong></div><div>The Sources for Pharo. Contribute to jgfoster/pharo development by creating an account on GitHub.</div></blockquote>

jgfoster:
@astares, I believe you have mischaracterized my answer to your initial question with your "quote" and I find it difficult to reply rationally because of your admonition. Nonetheless, I will try...

> an interesting project - at least for ... Gemstone customers

I don't know of any GemStone customers that are interested in this project. This was suggested by Pharo leaders.

> We already invested and still invest a lot in Pharo

Since you are speaking in the first person plural, would you be willing to give your real name? I find anonymity to make internet discussions more difficult and your GitHub profile is rather thin; I might take less offense if I knew who you are!

> We would like to move forward with Pharo and should not do steps backward - also not in the quality we accept.

Please recognize that there is a [cost](https://en.wikipedia.org/wiki/Perfect_is_the_enemy_of_good) to perfection. Would you regard #3824, #3825, #3828, and #3832 to be improvements? 

I understand that if the Pharo maintainers do not accept my pull requests then I'll need to follow a different approach (if I continue at all). But you should understand that as well. While that will avoid "steps backward," it might also deter some steps forward. (Please understand that this is not a threat of "If I don't get my way I'll leave," but an invitation to consider how to manage the delicate balance of maintaining quality while encouraging volunteers to participate in an open source project.) 

> this will end up in more problems over time - also more and more conflicts

How will this change end up with "more problems" a year from now than a week from now? Would you support a change that disallows string literals for FFI definitions?

> let Pharo and the community focus on its ... goals but also help you

I'm only doing this because I thought it was of interest to "Pharo and the community." I resent the implication that I'm asking you to help me. 

> **Currently I see several options you might want to evaluate**

Thank you for the suggestions. Working with my own fork and not contributing back to Pharo seems like the easiest.

jecisc:
Personally, I don't see the problem with having a string, even if at other places we do not use a string, as long as it's in the specification of UFFI that the parameter can be a string or an array. And as James said, it is written explicitly in the parser of UFFI.

Also, I don't see any difference between the two. Sometimes simple changes make sense like using #assert:equals: instead of #assert: and #= because the first one will print a better failure message. But in this case, what difference will it make?

@jgfoster For some fix you propose that might not be integrated into Pharo, the best solution might indeed be to start with a fork. But I hope it will not discourage you to contribute back to Pharo because multiple of your PR improved Pharo IMO.

Some of them did not improve or worsen Pharo, such as the one on Pragma>>#selector, but some of them definitely improved Pharo's code IMO. I have in mind PR #3824 and #3825 as an example. It even pushed me to do one more step after your PR to fix other places and add a non-regression test (#3839).

Also, your project seems really interesting and I hope you'll get some interesting results :)

jgfoster:
@astares, 

> I was clearly objecting due to the introduction of _incompatibilities_ in UFFI calls. ([emphasis added](https://github.com/pharo-project/pharo/pull/3843))

Do you really mean "[incompatible](https://www.lexico.com/en/definition/incompatible)" ("so different in nature as to be incapable of coexisting")? In software we generally use that term to apply to API changes that break existing code and it carries a bad connotation (I assume you aren't intentionally misapplying a pejorative term as a rhetorical device). 

I'd prefer to describe the proposal as "an alternative form that is equally or more readable, documented in the API, but not presently in general use"? Then we could discuss how much "improvement" is required to overcome consistency (or, to use a pejorative term as a rhetorical device, _inertia_ ("this is how we've always done it")).

I accept that _consistency_ has some value, and that opinions can differ as to the respective magnitudes.

astares:
Rather use "inconsistencies" - as I originally wrote.
Put quibbles aside. Better focus on what you want to achieve... Audere est facere.

estebanlm:
@jgfoster I want your experiment to succeed. 
But I think you are taking a longer path than needed. 
You know that since 7.0 we are bootstrapping Pharo. But maybe you are far from our day to day and you are not aware that that means that **you** can bootstrap your own version of Pharo. 
Absolutely nothing forbids you to do a clone of Pharo and work on that clone (you can build your own minimal images from there).
It seems to me that this will even speed your work since you do not need to do a PR on each workaround you need to do. 
The instructions to bootstrap from your machine are here:
https://github.com/pharo-project/pharo/blob/Pharo8.0/README.md#bootstrapping-pharo-from-sources
You will see is easy :)

I really want your experiment to succeed, please let me know if this approach would work for you.

<blockquote><div> GitHub</div><div><strong><a href="https://github.com/pharo-project/pharo">pharo-project/pharo</a></strong></div><div>The Sources for Pharo. Contribute to pharo-project/pharo development by creating an account on GitHub.</div></blockquote>

bencoman:
btw, do I infer this change is to work around spreading '#' through the FFI calls that you reverted in #3843 ?

As I mentioned on #3843, I think there is benefit to both sides to facilitate James continuing to work on Trunk which to me outweighs the negatives of this particular issue if the following contingencies can be put in place to review it before Pharo 8 Release: 
* a ticket is opened marked somehow to review this prior to Pharo8-Release 
* a comment in placed FFI calls saying "This FFI syntax is not standard Pharo and a temporary workaround to facilitate being parsed by Gemstone.  Refer to <above-ticket>"
* in parallel James follows up with Gemstone to see if they might parse the original code prior to Pharo8-Release.

The benefit for Pharo is that Jame's further improvements are more likely to end up in Pharo.
The benefit for James is less technical debt.

Now how workable this approach is depends on how many such contentious changes come along, but as James seems close to achieving what he needs, so lets be optimistic that wont be many. 
Lets work together and keep the option of James working in a fork in background until it becomes unworkable.

bencoman:
OSX server needs some clean up. CI error...
    Pharo7.0-PR-32bit-ae4039a.image:  write error (disk full?).  Continue? (y/n/^C)

jgfoster:
@estebanlm 
> I want your experiment to succeed.

Thanks! The encouragement helps.

> But I think you are taking a longer path than needed. ... [You should] do a clone of Pharo and work on that clone ... It seems to me that this will even speed your work since you do not need to do a PR on each workaround you need to do.

Yes, I'm already working on a clone. As to submitting back my changes, I have a strong aversion to long-running branches. In my experience is it a bad engineering practice to work on out-of-date code (I'm a strong believer in continuous integration). So, I'm trying to stay as close to the current Pharo8.0 as possible. 

@bencoman 
> btw, do I infer this change is to work around spreading '#' through the FFI calls that you reverted in #3843 ?

Yes. I agree that the first attempt was needlessly ugly and I think this addresses the readability issue. But I recognize that others prefer consistent use of the Array syntax over the String syntax, so spent the day modifying my export code so that it rewrites array literals (it wasn't exactly forward progress, but I'm back on track mapping primitives). This gives me something in GemStone that isn't identical to Pharo, but close enough for now. Thanks for your patience with the discussion.

bencoman:
I think we were approaching a point of merging this, but I understand enthusiasm is proportional to velocity and this your quickest way to move forward.  I've created #3894 for future contentious issues :)

jgfoster:
@bencoman, I do appreciate you working with me on this and I believe my concerns were being heard and considered. It turns out that solving the larger one (compilerOptions and metaLinkOptions) solved this one as well. I saw the new issue and hope we can keep the need for it minimized!

bencoman:
@estebanlm, Just curious... would using a string for the function call signature have any performance impact?

jgfoster:
@estebanlm,

 > The instructions to bootstrap from your machine are [here](https://github.com/pharo-project/pharo/blob/Pharo8.0/README.md#bootstrapping-pharo-from-sources). You will see is easy :)

I followed those instructions and it is easy to build a Pharo7.0 image. Now, how do I build a Pharo8.0 image?

estebanlm:
Sorry, I'm on holidays, hence the delayed answers :)
@bencoman It shouldn't, but I cannot be sure since I never tried, nor I know anyone that ever used it, it was "just there" :P

PR #3849, opened 2019-07-10
3847-TestRunner-code-coverage-does-not-need-to-use-method-reference-wrapping

### LINKED ISSUE #3847: TestRunner code coverage does not need to use method reference wrapping
TestRunner code coverage does not need to use method reference wrapping, it can use use the methods directly

MarcusDenker:
remove methodReference usage... (yes, these methods can be simplified further...(

fixes #3847

### DIFF
--- a/src/SUnit-UI/TestRunner.class.st
+++ b/src/SUnit-UI/TestRunner.class.st
@@ -94,11 +94,11 @@ TestRunner >> addMethodsUnderTestIn: listOfPackages to: methods [
 	listOfPackages
 		reject: #isNil
 		thenDo: [:package | 
-			package methodReferences
+			package methods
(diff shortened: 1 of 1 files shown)


jecisc:
I tried to run some coverage in my image but when I do Pharo does not answer anymore. :(

MarcusDenker:
is this change the reason?

jecisc:
No, it happens also without this change. But because of that I cannot test using the TestCoverage :(

Either Pharo does not answer as soon as I click on "Run Coverage" or when I select a package to cover.

Are you able to launch a test coverage selecting AST-Core-* tags for example?

Ducasse:
I would remove coverage from the SUnit.

PR #3852, opened 2019-07-10
2312-ondo-contexts-added-on-step-through-not-removed

### LINKED ISSUE #2312: #on:do: contexts added on step through not removed
When using step through the debugger adds an ` #on:do:` context to catch possible unhandled exceptions. Those contexts are however not removed from the stack when the debugger is opened.

For example, in the debugger below I used `Restart` and one `Step through` action five times. This adds five ` #on:do:`  contexts on the stack.

screen shot 2019-01-23 at 16 49 12

Using `Step through` in longer debugging sessions can lead to some confusion due to all the added contexts.

MarcusDenker:
uncomment a commented line that removes the on:do: handler.

fixes #2312


MarcusDenker:
failing test not related

PR #3862, opened 2019-07-10
3861-test-and-fix-for-slotUsers

### LINKED ISSUE #3861: test and fix for #slotUsers
slotUsers was missing a test (and broken due to that :-)

MarcusDenker:
fix #slotUsers, add test

fixes #3861

### DIFF
--- a/src/Slot-Tests/SlotTest.class.st
+++ b/src/Slot-Tests/SlotTest.class.st
@@ -97,3 +97,8 @@ SlotTest >> testisUsed [
 	self assert: IndexedSlot isUsed. "subclasses are users"
 	self assert: ProcessLocalSlot isUsed. "references count as uses, too"
 ]
+
+{ #category : #'tests - misc' }
+SlotTest >> testslotUsers [
+	self assert: ToOneRelationSlot slotUsers first equals: SlotExampleMovie
+]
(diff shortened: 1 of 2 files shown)


src/Slot-Tests/SlotTest.class.st
```diff
@@ -97,3 +97,8 @@ SlotTest >> testisUsed [
 	self assert: IndexedSlot isUsed. "subclasses are users"
 	self assert: ProcessLocalSlot isUsed. "references count as uses, too"
 ]
+
+{ #category : #'tests - misc' }
+SlotTest >> testslotUsers [
+	self assert: ToOneRelationSlot slotUsers first equals: SlotExampleMovie
```
jecisc, line 103:
```suggestion
	self assert: (ToOneRelationSlot slotUsers includes: SlotExampleMovie).
	self deny: (ToOneRelationSlot slotUsers includes: SlotExamplePerson)
```

jecisc (changes requested):
I would not use #first for the test because maybe one day there will be other users and the first will not be SlotExampleMovie. 

I proposed a change. Another solution would be to create a MockSlot with only one or two mock users.

src/Slot-Tests/SlotTest.class.st
MarcusDenker: I will do another PR with the improved test

PR #3874, opened 2019-07-11
3873-add-inspector-for-slot-using-classes

### LINKED ISSUE #3873: add inspector for slot using classes
When you inspect the class that defines a slot, it would be nice if it would show an inspector pane that shows all the using classes (all classes that use this kind of slot)

MarcusDenker:
fixes #3873


MarcusDenker:
after this is added, inspect SpecObservableSlot

PR #3883, opened 2019-07-11
3880-Remove-the-stack-pre-debuger

### LINKED ISSUE #3880: Remove the stack pre debuger
See: http://forum.world.st/Changes-proposal-on-Pre-debugger-td5101059.html

jecisc:
Open bre debugger only for warnings. Open full debugger for other errors.

This PR also remove useless debug actions and "open full debugger" setting.

Fixes #3880


MarcusDenker:
We tried to make the "use full debugger" as the default, but people convinced us in the past that this was not a good idea.

MarcusDenker:
e.g. the pre-debugger is very minimal: it might work, while the fill debugger might crash the image

jecisc:
For debugger crashes, I think we have the Emergency Evaluator showing the stack already.

jecisc:
@MarcusDenker We talked with Thomas and Steven. 

One thing we could have is fallback debuggers.

Currently, when your debugger crash, the emergency evaluator opens. 

What we could do is to register a collection of debuggers with order and if the first one crash, we fall back to the second one until the last that would be the emergency evaluator. 

Like that we could have for example:
Spec debugger -> Glamour debugger -> Scriptable debugger -> Emergency evaluator

PR #3890, opened 2019-07-11
Deprecated #addSelector:withMethod:notifying: methods should be in De…

### LINKED ISSUE #3889: Deprecated #addSelector:withMethod:notifying: methods should be in Deprecated package
Move deprecated methods

   Behavior>>#addSelector:withMethod:notifying: into
   TraitedClass>>#addSelector:withMethod:notifying: into

into Deprecated80 package so we can later easily remove them

Side note:
I guess they were forgotten during P7 development because a similar method #addAndClassifySelector:withMethod:inProtocol:notifying: 
is in Deprecated70 already

astares:
…precated package

Fix #3889

Move deprecated methods

Behavior>>#addSelector:withMethod:notifying: into
TraitedClass>>#addSelector:withMethod:notifying: into

into Deprecated80 package so we can later easily remove them


MarcusDenker:
The PR adds a method to SHRBTextStyler, that is strange

jecisc:
The problem is not that it adds a method. The problem is that the method is not removed from the original package.

astares:
@MarcusDenker 
I guess this is because I used the *extension functionality in Calypso - and it looks like this is pushing it down to the class instead of keeping the extension on the trait.

MarcusDenker:
I removed the method from with the git interface

MarcusDenker:
but that seems to break tonel loading?

PR #3904, opened 2019-07-12
Fix Fully #3410- [Trait] Impossible to use from a classTrait a method that is also defined in a subclass of Behavior even when following the doc

### LINKED ISSUE #3410: [Trait] Impossible to use from a classTrait a method that is also defined in a subclass of Behavior even when following the doc

VincentBlondeau:
Fix #3410
- Add tests
- Fix new issue with inheritance

### DIFF
--- a/src/TraitsV2-Tests/T2OverloadingOfMethodsInTraitedClassTest.class.st
+++ b/src/TraitsV2-Tests/T2OverloadingOfMethodsInTraitedClassTest.class.st
@@ -1,0 +1,78 @@
+Class {
+	#name : #T2OverloadingOfMethodsInTraitedClassTest,
+	#superclass : #T2AbstractTest,
+	#category : #'TraitsV2-Tests-Tests'
+}
+
+{ #category : #tests }
+T2OverloadingOfMethodsInTraitedClassTest >> createT1 [
+	| t1 |
+	t1 := self newTrait: #T1 with: '' asSlotCollection.
+	t1 classTrait
+		compile:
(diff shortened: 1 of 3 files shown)


MarcusDenker:
Error
Subclasses of TestCase should have #tearDown methods in one of the following protocols #(#running)
Stacktrace
ProperMethodCategorizationTest(TestAsserter)>>assert:description:resumable:
ProperMethodCategorizationTest(TestAsserter)>>assert:description:
ProperMethodCategorizationTest>>assureAll:areCategorizedInProtocols:whenSubclassOf:
ProperMethodCategorizationTest>>testTearDownMethodInSUnitTestsNeedsToBeInRunningProtocol

MarcusDenker:
test now look ok!

tesonep (approved):

PR #3907, opened 2019-07-12
3446-String-has-many-methods-about-searching-in-the-accessing-category fix #3446

### LINKED ISSUE #3446: String has many methods about searching in the 'accessing' category
String has many methods about searching in the 'accessing' category. It would be nice to put them in a category that is clearer ("searching"? not sure...).

majella67:
Merge 3d32bd4ec2cefeb931367203d8468be4418958fc


jecisc (changes requested):
#allRangesOfRegexMatches: and #allRegexMatches: changes should be reverted. Those methods are part of the Regex-Core package and not the String package.

jecisc (approved):

PR #3943, opened 2019-07-14
3942-speedup-isReferenced-for-ClassVariables

### LINKED ISSUE #3942: speedup #isReferenced for ClassVariables
we should implement a dedicated version of isReferenced for some LiteralVariable classes, this way we can return false for those that are never referenced and optimised the case of Class Variables

MarcusDenker:
fixes #3942


MarcusDenker:
will redo with a fresh image

PR #3952, opened 2019-07-15
Use #includesBehavior: instead of testing manually for class identity…

gcotelli:
… or inheritance

### DIFF
--- a/src/Announcements-Core/Announcement.class.st
+++ b/src/Announcements-Core/Announcement.class.st
@@ -37,8 +37,7 @@ Announcement class >> asAnnouncement [
 Announcement class >> handlesAnnouncement: anAnnouncement [
 	"The receiver acts as a filter to determine whether subscribers who used the receiver as signaling tag (event identifier class or symbol) should receive incoming announcement. In particular, registering to a superclass will receive the announcements from all subclasses."
 
-
-	^ anAnnouncement class == self or: [ anAnnouncement class inheritsFrom: self ]
+	^ anAnnouncement class includesBehavior: self
 ]
 
 { #category : #converting }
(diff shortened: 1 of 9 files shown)


jecisc:
Not for this PR but maybe we could have a lint rule to detect such cases and add a release test checking the rule is not violated in the image.

jecisc (approved):

MarcusDenker:
there are failing tests. they might be related

MarcusDenker:
testSingleTraitCreation
testAnnouncement

gcotelli:
I will take a look at the failing tests

gcotelli:
@MarcusDenker @jecisc I reviewed the failing tests, the problem it's with the `isKindOf:` change, the previous version considered a Trait as a valid argument instead of a class. In this specific case the problem is in `SystemDictionary>>#fillCaches` because with the new code traits are not included anymore in the `classNames` cache so `traitNames` is not responding with the proper trait names.

I can see two options here:
1. Revert the `isKindOf:` change , and in that case add a comment explaining that it can work with classes or traits. But it will remain inconsistent with the `includesBehavior: ` semantics IMHO, because traits are treated different.
2. Change `SystemDictionary>>#traitNames` to respond with the proper result and this will fix the failing tests.

I will wait for your input on how to proceed.

MarcusDenker:
I would do first 1). Because this way we already have a lot of of the way.

Then we do 2) in as a new issue. Because it is a bit more complex and we need to think about it a bit.
(the nice thing is that we actually do not need the name cache in dictionary anymore...)

gcotelli:
@MarcusDenker I've reverted this specific change and created #3958 for discussion.

MarcusDenker:
good! I will merge when the tests show that everything is good.

PR #3953, opened 2019-07-15
More minor cleanup for portability

jgfoster:
Remove duplicate periods; avoid GemStone compile errors. This brings us to 1640 classes and 27,850 methods that are loading (but not necessarily running!) in GemStone.

### DIFF
--- a/src/Kernel-Tests/BlockClosureTest.class.st
+++ b/src/Kernel-Tests/BlockClosureTest.class.st
@@ -150,11 +150,15 @@ BlockClosureTest >> testNew [
 
 { #category : #tests }
 BlockClosureTest >> testNoArguments [
+	| block1 block2 |
+	"avoid compile error in GemStone"
+	block1 := [:arg | 1 + 2].
+	block2 := [:arg1 :arg2 | 1 + 2].
 	[10
-		timesRepeat: [:arg | 1 + 2]]
+		timesRepeat: block1]
 		ifError: [:err | self deny: err = 'This block requires 1 arguments.'].
 	[10
-		timesRepeat: [:arg1 :arg2 | 1 + 2]]
+		timesRepeat: block2]
 		ifError: [:err  | self deny: err = 'This block requires 2 arguments.'] 
 ]
 
--- a/src/Kernel-Tests/NumberTest.class.st
+++ b/src/Kernel-Tests/NumberTest.class.st
@@ -142,8 +142,10 @@ NumberTest >> testRaisedToIntegerWithFloats [
 { #category : #tests }
 NumberTest >> testReadFrom [
 	
-	self assert: 1.0e-14	= (Number readFrom: '1.0e-14').
-	self assert: 2r1e26	= (Number readFrom: '2r1e26').
+	self assert: 1.0e-14    = (Number readFrom: '1.0e-14').
+	"Use a literal number format that compiles in GemStone.
+	 Note that the test still checks the same value."
+	self assert: 16r4000000 = (Number readFrom: '2r1e26').
 	self should: [Number readFrom: 'foo'] raise: Error
 ]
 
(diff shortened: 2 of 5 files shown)


src/Kernel-Tests/BlockClosureTest.class.st
```diff
@@ -150,11 +150,15 @@ BlockClosureTest >> testNew [
 
 { #category : #tests }
 BlockClosureTest >> testNoArguments [
+	| block1 block2 |
+	"avoid compile error in GemStone"
+	block1 := [:arg | 1 + 2].
+	block2 := [:arg1 :arg2 | 1 + 2].
 	[10
-		timesRepeat: [:arg | 1 + 2]]
+		timesRepeat: block1]
```
jgfoster, line 158: GemStone has a compiler optimization that recognizes literal blocks passed with the #timesRepeat: message. By placing the block in a variable, we hide the true nature from the compiler. The Pharo test runs the same and gets the expected error at runtime.

src/Kernel-Tests/NumberTest.class.st
```diff
@@ -142,8 +142,10 @@ NumberTest >> testRaisedToIntegerWithFloats [
 { #category : #tests }
 NumberTest >> testReadFrom [
 	
-	self assert: 1.0e-14	= (Number readFrom: '1.0e-14').
-	self assert: 2r1e26	= (Number readFrom: '2r1e26').
+	self assert: 1.0e-14    = (Number readFrom: '1.0e-14').
+	"Use a literal number format that compiles in GemStone.
+	 Note that the test still checks the same value."
+	self assert: 16r4000000 = (Number readFrom: '2r1e26').
 	self should: [Number readFrom: 'foo'] raise: Error
```
jgfoster, line 149: GemStone doesn't allow radix 2 floating point numbers (2r1e26), so I've changed the numeric literal to 16r4000000. The test works the same since we are testing the Number class >> readFrom: method. With this change the test still tests the same thing in Pharo and at least compiles in GemStone.

src/Kernel-Tests/TestDelayBasicSchedulerMicrosecondTicker.class.st
```diff
@@ -243,8 +243,8 @@ TestDelayBasicSchedulerMicrosecondTicker >> testPerformanceInlined_ifNotNil [
 { #category : #tests }
 TestDelayBasicSchedulerMicrosecondTicker >> testPerformanceInlined_ifTrue [
 	"For optimal performance, the outer conditionals of #timingPriorityHandleEvent should be inlined"
-	|codeSample|
-	codeSample := [ nil ifTrue: [ 0 ]].
+	| codeSample x |
+	codeSample := [ x ifTrue: [ 0 ]].
 	self assert: codeSample sourceNode body statements first isInlined
```
jgfoster, line 248: GemStone's compiler has a special check for a literal receiver for an #ifTrue: message. This change works the same in Pharo but uses a variable rather than a literal. It still checks for the statement being inlined.

src/Kernel-Tests/BlockClosureTest.class.st
MarcusDenker: we have the same optimisation, but it is turned off by default, we should turn it on, too

src/Kernel-Tests/NumberTest.class.st
MarcusDenker: ok

src/Kernel-Tests/TestDelayBasicSchedulerMicrosecondTicker.class.st
MarcusDenker: we could even remove these methods... #timingPriorityHandleEvent does not exist anymore, the test should have been written to test the real methods. Like this it makes not much sense.

(removing them would be another PR)

MarcusDenker (approved):
OK

PR #3960, opened 2019-07-16
3959-update-completion-with-the-latest-version

myroslavarm:
adding code for an updated completion


MarcusDenker:
Ci did not find any problem (good). We should download the CI build artefact and play with it a little to ensure that it is usable (as the new completion is on by default)

MarcusDenker:
we tested the version manually, seems to work ok (system is usable)

PR #3967, opened 2019-07-16
Use ifNil: ifNotNil: and ifNil:ifNotNil: instead of isNil and ifTrue/…

gcotelli:
…ifFalse combinations


jecisc:
For some of them we can improve further because sometimes a temporary variable is created and then we test it is not nil. We could just use the argument of the #ifNotNil: block. 
But that might be done in another PR :)

jecisc (approved):
Reading the code this seems good. If tests pass we can merge :)

gcotelli:
@jecisc I will do a first pass with this rewrites doing a PR per package group, and in a second pass we can detect such cases and improve them more

PR #3970, opened 2019-07-17
ifNil usage on Debugger-Model

gcotelli:

### DIFF
--- a/src/Debugger-Model/DebugSession.class.st
+++ b/src/Debugger-Model/DebugSession.class.st
@@ -265,16 +265,18 @@ DebugSession >> name: aString [
 
 { #category : #context }
 DebugSession >> pcRangeForContext: aContext [
+
 	"Answer the indices in the source code for the method corresponding to 
 	aContext's program counter value."
-	(aContext isNil) ifTrue: [ 
-		self terminate.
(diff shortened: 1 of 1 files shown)


src/Debugger-Model/DebugSession.class.st
```diff
@@ -265,16 +265,18 @@ DebugSession >> name: aString [
...
-		self terminate.
-		 ] ifFalse: [  
-	(aContext isDead) ifTrue:
-		[^1 to: 0].
-	^aContext debuggerMap
-		rangeForPC: aContext pc
-		contextIsActiveContext: (self isLatestContext: aContext) ]
+
+	aContext
+		ifNil: [ self terminate ]
```
MarcusDenker, line 273: Nice how this refactoring shows a bug: in the ifNil: case, the method returns a DebugSession, not a range

MarcusDenker: (we should merge this and add a new issue. Rule 1 of tiny changes: *only* do the tiny change, everything else should be added to the backlog)

MarcusDenker: added #3979

MarcusDenker (approved):

PR #3975, opened 2019-07-17
ifNil usage on M* packages

gcotelli:


jecisc (approved):
Seems fine :)

PR #3977, opened 2019-07-17
ClassBinding for Metaclass

MarcusDenker:
metaclasses use Association as a class binding. We should make a subclass of LiteralVariable so that we can implement a compatible API

### DIFF
--- a/src/ClassParser/CDClassDefinitionNode.class.st
+++ b/src/ClassParser/CDClassDefinitionNode.class.st
@@ -23,7 +23,7 @@ CDClassDefinitionNode class >> on: aRBMessageNode [
 { #category : #accessing }
 CDClassDefinitionNode >> binding [ 
 	Transcript show: '.'.
-	^ self class environment associationAt: self className ifAbsent: [nil -> self].
+	^ self class environment associationAt: self className ifAbsent: [ClassBinding key: nil value: self].
 ]
 
 { #category : #testing }
--- a/src/ClassParser/CDClassDefinitionNode.class.st
+++ b/src/ClassParser/CDClassDefinitionNode.class.st
@@ -23,7 +23,7 @@ CDClassDefinitionNode class >> on: aRBMessageNode [
 { #category : #accessing }
 CDClassDefinitionNode >> binding [ 
 	Transcript show: '.'.
-	^ self class environment associationAt: self className ifAbsent: [ClassBinding key: nil value: self].
+	^ self class environment associationAt: self className ifAbsent: [LiteralVariable key: nil value: self].
 ]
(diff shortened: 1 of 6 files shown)


src/ClassParser/CDClassDefinitionNode.class.st
```diff
@@ -23,7 +23,7 @@ CDClassDefinitionNode class >> on: aRBMessageNode [
 { #category : #accessing }
 CDClassDefinitionNode >> binding [ 
 	Transcript show: '.'.
-	^ self class environment associationAt: self className ifAbsent: [nil -> self].
+	^ self class environment associationAt: self className ifAbsent: [ClassBinding key: nil value: self].
```
tesonep: It is not better to have a constructor with only the value, so we don't have to pass the nil key?

src/ClassParser/CDClassNameNode.class.st
```diff
@@ -10,7 +10,7 @@ Class {
 { #category : #accessing }
 CDClassNameNode >> binding [
 	| classBinding |
-	classBinding := self existingBindingIfAbsent: [className -> nil].
+	classBinding := self existingBindingIfAbsent: [ClassBinding key: className value: nil].
```
tesonep: And a constructor for only the name, and it is not better to use the name of what the key is? Like className?
At least for the external API

src/Slot-Core/ClassBinding.class.st
```diff
@@ -0,0 +1,8 @@
+"
```
tesonep: Why a new class? If there is no different behavior so long we can use LiteralVariable. 
Or better where are the places we have to refactor to take advantage of adding the new class.

MarcusDenker:
Hmm... 
-> the API question: you are right, we should not use the Association API. I will add that to the backlog.
-> LiteralVariable vs. it's own class: indeed... I did that already for the case where we reference metalinks (that is, meta objects defined by links) in a method.

I will change the PR to use LiteralVariable for now, but the API use I will add as a new TODO item

Ducasse:
Could you also rename LIteralVariable into LiteralVariableAssociation (if this is a subclass of Association).

MarcusDenker:
The thing is that it is only an Association for implementation reasons.

e.g. there is a subclass "ClassVariable" which plays *exactly* the role of what a Slot is for instance variables: the "first class variable" object. It is just an Association so that we re-use an existing instance and not add more instances for each variable.

This is not perfect, but I would like to not call the variables "ClassVariableAssociation" as the are supposed to play mainly the role of a Variable, not that of an Association.

PR #3980, opened 2019-07-17
ifNil usage on Polymorph package

gcotelli:


jecisc (approved):
Changes seem fine. We can merge if tests pass.

PR #3981, opened 2019-07-17
ifNil usage on packages starting with R

gcotelli:


gcotelli:
I will review the failing tests

MarcusDenker (approved):

PR #3994, opened 2019-07-17
Hint GH language parser on .st files

gcotelli:
Hint the GitHub language parser, so it take .st files as Smalltalk code and report right the language percentages


jecisc (approved):

dionisiydk:
Can we force it from Iceberg?

PR #3995, opened 2019-07-17
Use indirect access for classes that might not exist

jgfoster:
MailMessage, MailAddressParser, and ImageReadWriter are referenced from the "minimal image" but don't exist. This PR modifies the reference to be indirect so that at least we can compile the code.


bencoman (approved):
LGTM

src/Network-MIME/MIMEDocument.class.st
```diff
@@ -277,7 +277,8 @@ MIMEDocument >> parts [
 	[ parseStream atEnd ] whileFalse: 
 		[ messages add: msgStream upToEnd.
 		msgStream skipThisLine ].
-	^ messages collect: [ :e | MailMessage from: e ]
+	"MailMessage is not present in a minimal image"
+	^ messages collect: [ :e | (self environment at: #MailMessage) from: e ]
```
dionisiydk: It looks like you are trying to fix dependency issue here.

I don't know details of MIMEDocument. But the method #parts looks like required part of it and it strictly depends on MailMessage. Thus we should fix dependency to have this class in minimal image or remove MIMEDocument from it. 
Otherwise you just hide the issue.

src/Network-Protocols/SMTPClient.class.st
```diff
@@ -159,7 +159,8 @@ SMTPClient >> mailFrom: fromAddress [
 	" MAIL <SP> FROM:<reverse-path> <CRLF>"
 
 	| address |
-	address := (MailAddressParser addressesIn: fromAddress) first.
+	"MailAddressParser is not in the minimal image"
+	address := ((self environment at: #MailAddressParser) addressesIn: fromAddress) first.
```
dionisiydk: Same here. SMTP requires MailAddressParser. So if SMTP is in image then this parser should be here too.

dionisiydk (changes requested):

src/Network-MIME/MIMEDocument.class.st
jgfoster: Actually, no, I'm not trying to fix a dependency. I'm trying to fix a compile error (that is currently hidden). Is it necessary to resolve that more difficult dependency issue now, or can we let this PR fix the compile error and #3998 deal with the dependency at such time as someone who knows more than I do has time to address it?

src/Network-Protocols/SMTPClient.class.st
jgfoster: See #3999.

src/Zinc-HTTP/ZnEasy.class.st
```diff
@@ -86,7 +86,8 @@ ZnEasy class >> getImageOfType: mimeType fromUrl: urlObject [
 		enforceAcceptContentType: true;
 		get.
 	"ImageReadWriter does automatic type detection"
-	^ ImageReadWriter formFromStream: client entity readStream
+	"ImageReadWriter is not in minimal image"
+	^ (self environment at: #ImageReadWriter) formFromStream: client entity readStream
```
jgfoster: See #4000.

bencoman:
I think its appropriate to log separate tickets for followup work (as James has done) and merge this to unblock his work evaluating a Gemstone port of Pharo.  Consider Jame's work a "dependency identification" phase.  They can easily gathered later using Finder>SearchSource... "self environments at: #"

dionisiydk (approved):
I am fine with that. 
But I am curious: Gemstone does not allow code with undefined variables?

MarcusDenker:
I have personally come to the conclusion that the Smalltalker's fear of Undeclared variables is a bit odd... shouldn't we late-bind class access not the same as we late bind messages?

MarcusDenker:
There are lots of SystemDependenciesTest failing now (maybe because these dependencies are now not there anymore). One example:

Given Collections do not match. Got left := #(#'Athens-Morphic' #'Fonts-Chooser' #RecentSubmissions #'Refactoring-Critics' #'Refactoring-Environment' #Shout #'Spec-Core' #'Tool-Diff' #'Tool-FileList' #'Tool-Profilers'). instead of right :=#(#'Athens-Morphic' #'Fonts-Chooser' #'Network-Mail' #RecentSubmissions #'Refactoring-Critics' #'Refactoring-Environment' #Shout #'Spec-Core' #'Tool-Diff' #'Tool-FileList' #'Tool-Profilers'). Collection size does not match: left=10 vs. right=11 Got 1 additional element(s) in the right collection: #(#'Network-Mail')
Stacktrace
SystemDependenciesTest(TestAsserter)>>assertCollection:equals:
SystemDependenciesTest>>testExternalMorphicDependencies

MarcusDenker:
failing tests:

testExternalDisplayDependencies
testExternalMetacelloDependencies
testExternalMonticelloDependencies
testExternalMorphicCoreDependencies
testExternalMorphicDependencies
testExternalSUnitDependencies
testExternalUFFIDependencies
testExternalUIDependencies

Ducasse:
The problem is that using environment at: #X is not a nice way to address the problem.
It will just makes the dependencies more difficult to identify. 
I'm really against such changes.

src/Zinc-HTTP/ZnEasy.class.st
guillep: Hi, probably @svenvc would like to see this change.

I think a better solution would be to put this method into a separate package? something like Zinc-HTTP-Images?

guillep (commented):
I would prefer to see how we can move these methods to another package too.
The reality is that this style of coding just hides the problem, and then it will most probably not be fixed  for a super long time.

I can check these methods later and see what I can help with.

src/Zinc-HTTP/ZnEasy.class.st
jgfoster: Feel free to submit a PR doing something else.

jgfoster:
Note that this does not "hide" a problem, but exposes a problem that was previously undiscovered.

jgfoster:
I'm thinking that the best short-term solution is for me to include the missing classes. I'll submit a new PR for that.

bencoman:
@jgfoster, an alternative approach I believe will have broader support is make the method an extension of the package that is home for the missing-class. 

For example, `ImageReadWriter` is in package **"Graphics-Files"**, so try this...
1. In class `ZnEasy`, create a temporary protocol "temp"
2. Drag method `#getImageOfType:fromUrl:`  into "temp" protocol.
3. Right-click on "temp" protocol then "Convert to extension" and choose package **"Graphics-Files"**
4. Commit, push, submit PR.

@guillep, @Ducasse, would that be satisfactory?

guillep:
> @jgfoster, an alternative approach I believe will have broader support is make the method an extension of the package that is home for the missing-class.
> 
> For example, `ImageReadWriter` is in package **"Graphics-Files"**, so try this...
> 
> 1. In class `ZnEasy`, create a temporary protocol "temp"
> 2. Drag method `#getImageOfType:fromUrl:`  into "temp" protocol.
> 3. Right-click on "temp" protocol then "Convert to extension" and choose package **"Graphics-Files"**
> 4. Commit, push, submit PR.
> 
> @guillep, @Ducasse, would that be satisfactory?

Yes, an approach using extension methods would be more appropriate IMHO

jgfoster:
"Commit should have been implemented in IceUnknownCommit."

Screen Shot 2019-07-23 at 8 28 38 AM

bencoman (approved):
From a static review, LGTM.

bencoman:
@jgfoster, Is that Iceberg error still an Issue for you or did you get past it?
If you can detail steps to reproduce it, recording that in another issue would be useful.

jgfoster:
@dionisiydk 
> But I am curious: Gemstone does not allow code with undefined variables?

Unfortunately, no. GemStone expects identifiers to exist at the time they are referenced. 

@MarcusDenker 
> I have personally come to the conclusion that the Smalltalker's fear of Undeclared variables is a bit odd... shouldn't we late-bind class access not the same as we late bind messages?

Fair enough, though note that there is a degree of late binding already in that we bind to an Association that holds a value that will be used. The value can be changed after the code is compiled. So in a sense we do have late binding to the value.

@MarcusDenker 
> There are lots of SystemDependenciesTest failing now

Were these failures introduced by this PR? If not, should they be addressed in a new issue?

@bencoman 
> Is that Iceberg error still an Issue for you or did you get past it?
If you can detail steps to reproduce it, recording that in another issue would be useful.

See #4088.

jgfoster:
I've redone this using extension methods. Any other comments or suggestions?

MarcusDenker:
SystemDependenciesTest is failing still on the CI. They need to be updated (the removed dependencies need to be removed from the list of dependencies in the tests)

jecisc:
@jgfoster The dependency tests are based on a tool called DependencyAnalyzer. This tool is able to find most dependencies bu,t not all of them.

When you are removing direct accesses to classes, the tool will not find the dependencies anymore. This is why Stephane said, «It will just makes the dependencies more difficult to identify.»

So, to be integrated the PR need to update the dependency tests to remove the dependencies that are not found anymore. That will make it harder for us to clean those dependencies later but since you opened issues about them I guess it's fine. We have the information somewhere.

bencoman:
> That will make it harder for us to clean those dependencies later  
@jecisc, is that still a concern after the dependent-method has converted to an extension from the package of the referenced class ?  Or is that then considered to have been cleaned and not need later action?

jgfoster:
I believe I've fixed the failing SystemDependenciesTest. Anything else?

guillep (approved):
Good to me too

Ducasse:
Thx

PR #4008, opened 2019-07-18
4007-Propagate-correctly-changes-in-a-trait-when-used-in-an-anonymous-class

tesonep:
When a trait is used in an anonymous class changing the trait should not install in the system environment.


jecisc:
Would be awesome to also backport this one if possible :)

MarcusDenker (approved):

PR #4012, opened 2019-07-18
CodeCompletionImprove

MarcusDenker:
- improve variable completion: do only show vars
- methodnode: allow completion of parametor
- Model: Cleanup + create model (and reparse) more often


Ducasse:
The GTevent tests were broken and I do not if this is due to this PR.

MarcusDenker:
I do not see the GTEvent test failure, so I guess it was a radom failure

PR #4013, opened 2019-07-18
Use Integer literal to avoid Float in GemStone

jgfoster:
GemStone treats 1e3 as a SmallDouble (Float) while Pharo treats it as an Integer. Here we make it explicitly an Integer as 1000.


jecisc (approved):
Code is equivalent, should be fine to merge.

bencoman (approved):

PR #4022, opened 2019-07-20
4020-OCLiteralSet-can-be-replaced-by-PluggableSet

### LINKED ISSUE #4020: OCLiteralSet can be replaced by PluggableSet
In the compiler, we have a dedicated subclass of Set to check identity using #literalEquals.

This actually can be replaced by a PluggableSet configured correctly to use #literalEquals:

  PluggableSet new  equalBlock: [ :a :b | a literalEqual: b]

(Compiler tests are green).

Why is this good? It removes on subclass of Set. Subclassing basic collection classes can be sub-optimal as they assume too much about the implementation details of the superclass.

Note: as #literalEquals: is a Kernel method, we could add a class side instance creation method to PluggableSet in the style of #integerSet

### LINKED ISSUE #4021: PluggableSet class>#integerSet should be in instance creation category
PluggableSet class>#integerSet  is categorised as "accessing", this is not nice as is is constructor. It should be in something like "instance creation" category.

(This is a very trivial issue perfectly do be done by a beginner)

MarcusDenker:
In the compiler, we have a dedicated subclass of Set to check identity using #literalEquals.

This actually can be replaced by a PluggableSet configured correctly to use #literalEquals:

PluggableSet new equalBlock: [ :a :b | a literalEqual: b]

fixes #4020 
fixes #4021

(Compiler tests are green).

Why is this good? It removes on subclass of Set. Subclassing basic collection classes can be sub-optimal as they assume too much about the implementation details of the superclass.

Note: as #literalEquals: is a Kernel method, we could add a class side instance creation method to PluggableSet in the style of #integerSet


src/Collections-Unordered/PluggableSet.class.st
```diff
@@ -45,11 +45,17 @@ Class {
...
-{ #category : #accessing }
+{ #category : #'instance creation' }
 PluggableSet class >> integerSet [
 	^self new hashBlock: [:integer | integer hash \\ 1064164 * 1009]
 ]
 
+{ #category : #'instance creation' }
+PluggableSet class >> literalSet [
+	"A Set using equality as defined for the Literals in CompiledMethods"
+	^self new equalBlock: [ :a :b | a literalEqual: b]
```
gcotelli, line 56: Are we sure the default hash it's Ok for every object answering true to `literalEqual:` ? For example `Association` implementation uses `==` but the default hash only considers the key in the association.

MarcusDenker: I checked OCLiteralSet and it uses the default hash... which had the same problem... and I asked myself the same...

astares:
@MarcusDenker Just to make sure: the bot closed it - is this still valid?

MarcusDenker:
Yes, I wanted to check the equal vs hash (that problem we do have already now, so the change would not make things worse)

MarcusDenker:
I opened it

astares:
keep open

PR #4027, opened 2019-07-22
moveCodeCompletionAroundToEditor

MarcusDenker:
This is a preparation to get rid of the uniqueInstace in the completion engine: the editor should handle instantiation.

To prepare this, we move the method where the uniqueInstance is created to the editor. This implifies the code a bit already, too


myroslavarm (approved):
looks good!

PR #4032, opened 2019-07-22
4029-CodeCompletion-bug-fix-for-visitBlockNode

MarcusDenker:
- add fallBack #completionToken to RBProgramNode
- add #visitBlockNode: to the NodeProducer
- for arguments: only show 0-arg selectors as possible completion


myroslavarm (approved):
this is a good fix!

PR #4033, opened 2019-07-22
4019-Measure-the-use-and-usefulness-of-literal-string-sharing-in-literal-frame

MarcusDenker:
experiment: try to not share String literals that are equal in a compiledmethod


Ducasse:
marcus could you document what is literalEqual: because now with the context I think that I understand but in 3 months reading this will be totally obscure. So adding comment to this method explaining what it controls would be good and needed.

MarcusDenker:
yes, I will do that

PR #4047, opened 2019-07-23
Isolate references to Scheduler class variable in Delay

jgfoster:
In GemStone we cannot use the same Scheduler across multiple VMs, so by isolating the references we make it easier to substitute a local Scheduler.

### DIFF
--- a/src/Kernel/Delay.class.st
+++ b/src/Kernel/Delay.class.st
@@ -34,13 +34,13 @@ Class {
 { #category : #testing }
 Delay class >> anyActive [
 	"Return true if there is any delay currently active"
-		^Scheduler anyActive
+		^self scheduler anyActive
 
 ]
 
 { #category : #settings }
 Delay class >> delaySchedulerClass [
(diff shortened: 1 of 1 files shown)


jecisc (approved):
It just add a getter and setter for a class var. 

Seems all good!

bencoman:
Hi James,  The delay scheduling is a special interest of mine.  I'd like to chat about how this will work on Gemstone. In the past I've had some ideas about using external semaphores rather than the Cog-VM-internal-semaphores and wonder if they might be applicable.  
Buzz me on https://discordapp.com/invite/QewZMZa (my timezone is GMT+8)


<blockquote><div> Discord</div><div><strong><a href="https://discordapp.com/invite/QewZMZa">Join the Pharo Discord Server!</a></strong></div><div>Check out the Pharo community on Discord - hang out with 1,422 other members and enjoy free voice and text chat.</div></blockquote>

bencoman (approved):

Issue #4045, opened 2019-07-22
RBSequenceNode: #stop does not take temp definitions into account


MarcusDenker:
We have found that #stop for SquenceNode is wrong if there are no statements *but* temp definitions. e.g. method | temp |
the stop of the temp is correct, just RBSequenceNode returns 0.

A possible fix:

```Smalltalk
stop
	| tempDefinitionStop |
	tempDefinitionStop := (self temporaries isNotEmpty) 
		ifTrue: [ self temporaries last stop  ]
	 	ifFalse: [ 0 ]. 
	
	^(periods isEmpty ifTrue: [tempDefinitionStop] ifFalse: [periods last]) 
		max: (statements isEmpty ifTrue: [tempDefinitionStop] ifFalse: [statements last stop])
```

MarcusDenker:
strongly I can not push

stop

	^{  
	temporaries isEmpty ifTrue: [0] ifFalse: [ self temporaries last stop]. 
	periods isEmpty ifTrue: [0] ifFalse: [periods last].
	statements isEmpty ifTrue: [0] ifFalse: [statements last stop]
	} max

and for a new RBSequenceNodeTest:

testStop
	| ast |
	ast := RBParser parseMethod: 'method | temp |'.
	self assert: ast body stop equals: ast body temporaries last stop.

MarcusDenker:
merged

PR #4049, opened 2019-07-23
4045-RBSequenceNode-stop-does-not-take-temp-definitions-into-account

MarcusDenker:
test + fix for #4045

### DIFF
--- a/src/AST-Core-Tests/RBSequenceNodeTest.class.st
+++ b/src/AST-Core-Tests/RBSequenceNodeTest.class.st
@@ -1,0 +1,12 @@
+Class {
+	#name : #RBSequenceNodeTest,
+	#superclass : #TestCase,
+	#category : #'AST-Core-Tests-Nodes'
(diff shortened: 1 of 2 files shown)


myroslavarm (approved):
this is a cool fix!

estebanlm:
late to the party but... why 
```Smalltalk
temporaries isEmpty ifTrue: [0] ifFalse: [ self temporaries last stop]. 
periods isEmpty ifTrue: [0] ifFalse: [periods last].
statements isEmpty ifTrue: [0] ifFalse: [statements last stop]
```
instead 
```Smalltalk
temporaries ifNotNil: [ self temporaries last stop ] ifNil: [ 0 ]. 
... etc
```
?

myroslavarm:
@estebanlm it was the original syntax and we were more preoccupied with fixing the bug than optimizing the code i guess. but you are right, could be simplified further

PR #4050, opened 2019-07-23
RubSmalltalkEditor #keystroke Simplify more

MarcusDenker:
- #isCompletionEnabled should check if CompletionEngine is not nil
- we can simplify #keystroke: even more by moving more code into #completionAround:keyStroke:
- trivial cleanup in TextEditor>>#keystroke:


myroslavarm (approved):
good

PR #4051, opened 2019-07-23
4030-CodeCompletion-DNU-when-completing-sends-to-Class-Variables

### LINKED ISSUE #4030: CodeCompletion: DNU when completing sends to Class Variables
There is a bug when completing sends to class variables
(e.g. take the ones in SmalltalkImage as an example).

Problem seems to be that the typer (class TestTypingVisitor) does not do the right thing
in visitGlobalNode:

MarcusDenker:
fix #4030

### DIFF
--- a/src/NECompletion/TestTypingVisitor.class.st
+++ b/src/NECompletion/TestTypingVisitor.class.st
@@ -54,8 +54,10 @@ TestTypingVisitor >> visitBlockNode: aBlockNode [
 ]
 
 { #category : #visiting }
-TestTypingVisitor >> visitGlobalNode: aGlobalNode [ 
-	aGlobalNode propertyAt: #type put: (Smalltalk globals at: aGlobalNode name) class.
+TestTypingVisitor >> visitGlobalNode: aGlobalNode [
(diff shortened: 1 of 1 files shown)


src/NECompletion/TestTypingVisitor.class.st
```diff
@@ -54,8 +54,10 @@ TestTypingVisitor >> visitBlockNode: aBlockNode [
 ]
 
 { #category : #visiting }
-TestTypingVisitor >> visitGlobalNode: aGlobalNode [ 
-	aGlobalNode propertyAt: #type put: (Smalltalk globals at: aGlobalNode name) class.
+TestTypingVisitor >> visitGlobalNode: aGlobalNode [
+	aGlobalNode
+		propertyAt: #type
+		put: aGlobalNode binding assoc read class
```
jecisc, line 60: Maybe we could have a method on global node to avoid hurting Demeter too much in a next PR .

MarcusDenker: yes, I make a note in the backlog

PR #4052, opened 2019-07-23
Speedup code loading for big packages

guillep:
Several little improvements that improves the loading of big packages (e.g., loading VMMaker down from ~120 secs to ~50 secs)

- Accelerate pragma collector by not iterating all methods in the class
- Make Monticello to only listen to events coming not from Monticello
- Cache sort-key, for big packages containing lots of methods
- Remove weak cache that adds pressure on the GC for seemingly nothing

### DIFF
--- a/src/Monticello/MCMethodDefinition.class.st
+++ b/src/Monticello/MCMethodDefinition.class.st
@@ -38,19 +38,14 @@ MCMethodDefinition class >> cachedDefinitions [
 ]
 
 { #category : #'instance creation' }
-MCMethodDefinition class >> className: classString
-classIsMeta: metaBoolean
-selector: selectorString
-category: catString
-timeStamp: timeString
-source: sourceString [
-	^ self instanceLike:
-		(self new initializeWithClassName: classString
-					classIsMeta: metaBoolean
-					selector: selectorString
-					category: catString
-					timeStamp: timeString
-					source: sourceString)
+MCMethodDefinition class >> className: classString classIsMeta: metaBoolean selector: selectorString category: catString timeStamp: timeString source: sourceString [
+	^ self new
+		initializeWithClassName: classString
+		classIsMeta: metaBoolean
+		selector: selectorString
+		category: catString
+		timeStamp: timeString
+		source: sourceString
 ]
 
 { #category : #'instance creation' }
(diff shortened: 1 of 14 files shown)


src/Monticello/MCMethodDefinition.class.st
```diff
@@ -367,7 +352,12 @@ MCMethodDefinition >> shortSummaryPrefix [
 
 { #category : #comparing }
 MCMethodDefinition >> sortKey [
-	^ self className, '.', (self classIsMeta ifTrue: ['meta'] ifFalse: ['nonmeta']), '.', self selector
+	^ sortKey
+		ifNil: [ sortKey := self className , '.'
+				,
+					(self classIsMeta
+						ifTrue: [ 'meta' ]
+						ifFalse: [ 'nonmeta' ]) , '.' , self selector ]
```
jecisc, line 360: Maybe we could even have:

```Smalltalk
self className, (self classIsMeta ifTrue: ['.meta.'] ifFalse: ['.nonmeta.']), self selector
```

to avoid useless concatenations.

jecisc (approved):
Reading the code it seems to make sense. 

Let's wait to see what the CI is saying :)

jecisc:
Build error seems weird:

```
Loaded -> System-Identification-tonel.1 --- tonel:///builds/workspace/uest_and_branch_Pipeline_PR-4052/32/src --- cache
	Loaded -> Tool-Base-tonel.1 --- tonel:///builds/workspace/uest_and_branch_Pipeline_PR-4052/32/src --- cache
Finished atomic load'Starting postload action'
Sending interrupt signal to process
/builds/workspace/uest_and_branch_Pipeline_PR-4052/32/bootstrap-cache/vmtarget/pharo: line 11: 31616 Terminated              "$DIR"/"pharo-vm/pharo" --nodisplay "$@"
script returned exit code 143
```

I'll relaunch it.

guillep:
I was a timeout, probably my changes affect something during monticello bootstrap. I'm checking it locally, no need.

pavel-krivanek:
Nice. It is interesting that the cache actually slowed things down significantly.

pavel-krivanek (approved):

guillep:
Actually, it seems it is a space optimization, meant for the case when we will have many (similar) mc snapshots alive at the same time. Funny thing is that the cache was not avoiding creating the objects, it was creating the objects and discarding them if it found a similar one in the cache already...

jecisc (approved):

jecisc:
PR is green and seems ready to merge.

Ducasse:
This is nice. I read all the code. What I see is that I would love to have more comments. 
Monticello was never commented as many other packages. 
This does not help revisiting the decision after the fact.

PR #4053, opened 2019-07-23
nodeForOffsetAndTests

myroslavarm:
adding a method as an alternative to bestNodeFor and tests for it


MarcusDenker (approved):
CI looks good, it is not yet hooked up so we can safely merge

PR #4056, opened 2019-07-23
4055-Code-Completion-temp-vars-have-methods-with-colons-as-suggestions

myroslavarm:
filtering the results


MarcusDenker (approved):
ok. We added general cleanup of that method to the backlog

PR #4059, opened 2019-07-23
adding-nodeForOffset-instead-of-bestNodeFor

myroslavarm:
replacing bestNodeFor and interval


MarcusDenker:
we should test the build artefact of this one a little

MarcusDenker:
I played with code completion with this change enable and it looks that it works. I will merge

PR #4075, opened 2019-07-24
2311-Strange-behaviour-in-DebugSessionpcRangeForContext-

### LINKED ISSUE #2311: Strange behaviour in DebugSession>>#pcRangeForContext: (?)
The current implementation of `DebugSession>>#pcRangeForContext:` is:

```
pcRangeForContext: aContext
	"Answer the indices in the source code for the method corresponding to 
	aContext's program counter value."
	(aContext isNil) ifTrue: [ 
		self terminate.
		 ] ifFalse: [  
	(aContext isDead) ifTrue:
		[^1 to: 0].
	^aContext debuggerMap
		rangeForPC: aContext pc
		contextIsActiveContext: (self isLatestContext: aContext) ]
```

So if the parameter `aContext` is nil the current session is terminated. Calling `terminate` will kill the current interrupted process making the debugger not really usable any more.

Normally `pcRangeForContext:` should just be an accessor method computing the pc range.

Looking at the senders of `pcRangeForContext:` it should be possible to remove the call to `terminate`

### LINKED ISSUE #3979: pcRangeForContext: does not return a range if nil
pcRangeForContext: does not return a range if nil

MarcusDenker:
return the same value 1 to: 0 in case context being nil as it was already for dead contexts. 

fixes #2311 
fixes #3979

### DIFF
--- a/src/Debugger-Model/DebugSession.class.st
+++ b/src/Debugger-Model/DebugSession.class.st
@@ -265,18 +265,14 @@ DebugSession >> name: aString [
 
 { #category : #context }
 DebugSession >> pcRangeForContext: aContext [
-
 	"Answer the indices in the source code for the method corresponding to 
 	aContext's program counter value."
 
-	aContext
-		ifNil: [ self terminate ]
-		ifNotNil: [ aContext isDead
-				ifTrue: [ ^ 1 to: 0 ].
-			^ aContext debuggerMap
-				rangeForPC: aContext pc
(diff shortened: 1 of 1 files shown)


MarcusDenker:
it would be nice to get a review for this one...

Ducasse:
I read it but I do not understand. May be thomas can have a look.

MarcusDenker:
still relevant

PR #4079, opened 2019-07-24
2341-memory-leak-through-NECController-uniqueInstance

MarcusDenker:
remove ivar

### DIFF
--- a/src/NECompletion/NECController.class.st
+++ b/src/NECompletion/NECController.class.st
@@ -38,6 +38,11 @@ NECController class >> contextClass: aClass [
 	ContextClass := aClass
 ]
 
+{ #category : #'class initialization' }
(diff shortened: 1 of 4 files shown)


MarcusDenker:
(somehow I could not commit the class ivar removal, might be a Iceberg issue, to be checked later)

MarcusDenker:
tested it in the build artefact a little and code completion works ok with this change

Ducasse:
I read everything but I do not have enough context. 

Stef

MarcusDenker:
The idea is to do a new instance (per editor) instead of using #uniqueInstance on the class of the CodeCompletion controller.

I just hooked everything into the already prepared setup in the editor and it seems to work

I will merge so we test it for real

PR #4086, opened 2019-07-24
3955-remove-testPerformanceInlined-ifNotNil-and-testPerformanceInlined-ifTrue

MarcusDenker:
runBackendLoopAtTimingPriority
- comment out debug checks
- use #== instead of notNil
- ban the notNil rule for the class
- remove the two tests that checked for optimzied coed


Ducasse:
Marcus are you sure that the isNil -> ~= nil does not change the scheduler code?
I imagine but I prefer to ask.

MarcusDenker:
"something notNil" should be the same as "something ~= nil", just without the message send.
(I think, but my brain is not working at 39 degrees anymore ;-)

PR #4091, opened 2019-07-25
simplify some lazy read accessors

MarcusDenker:
simplify some lazy read accessors of the kind

CatalogSourceURL ifNil: [ CatalogSourceURL := self defaultCatalogSourceURL ].
	^CatalogSourceURL


jecisc:
@MarcusDenker Did you had a script to find them? If yes, can we make it a rule and add a release test checking there is no violation to this rule?

MarcusDenker:
No, this was my "what do I do at 1 am when it is still nearly 30 degrees"... I stopped because yes, this should be a script...

Issue #4112, opened 2019-07-26
TestRunner is using instVarNamed: instead of an accesor


GDayne:
The TestRunner is changing the variables in a TestResult with 
```
instVarNamed: 'errors'
instVarNamed: 'failures'
```
It is beacause the TestResult does not have accesors.
To solve this problem we can add the accesors and work with that in TestRunner

PR #4114, opened 2019-07-26
4112-TestRunner-is-using-instVarNamed-instead-of-an-accesor

CafeKrem:
fix issue #4112 .
instVarNamed:  replace by setter


CafeKrem:
we run the test (on OSX, LINUX) and we saw that the broken CI tests are not related to the changes

MarcusDenker (approved):
ok

PR #4120, opened 2019-07-26
4119-Code-Completion-method-names-are-not-updated

myroslavarm:
adding methods on the node to update the AST

### DIFF
--- a/src/AST-Core/RBMethodNode.class.st
+++ b/src/AST-Core/RBMethodNode.class.st
@@ -228,6 +228,11 @@ RBMethodNode >> compilationContext: aCompilationContext [
 	compilationContext := aCompilationContext.
 ]
 
+{ #category : #completion }
+RBMethodNode >> completionToken [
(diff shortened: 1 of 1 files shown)


MarcusDenker (approved):
ok

src/AST-Core/RBMethodNode.class.st
```diff
@@ -476,6 +481,12 @@ RBMethodNode >> methodOrBlockNode [
 	"^ self"
 ]
 
+{ #category : #completion }
+RBMethodNode >> narrowWith: aString [
```
dionisiydk, line 485: Can you explain the meaning of narrowWith:?

MarcusDenker: it is a bad name: better would be #updateNode: or something like that. But the whole idea to update the AST I do not like. Instead we should make sure to re-parse. I added an issue to the code completion project issue tracker for that.

PR #4121, opened 2019-07-26
2190-RPackagecategoryName-should-deprecated-and-just-use-name

### LINKED ISSUE #2190: RPackage>>categoryName should deprecated and just use name
Let us do another pass and clean...

RPackage >> categoryName
    self deprecated rewriting....
    ^ self name

So clients should call name directly.

apierr:
Fixes #2190

### DIFF
--- a/src/Deprecated80/RPackage.extension.st
+++ b/src/Deprecated80/RPackage.extension.st
@@ -3,6 +3,7 @@ Extension { #name : #RPackage }
 { #category : #'*Deprecated80' }
 RPackage >> categoryName [
 	self
-		deprecated: 'This method will be removed. Clients should call name directly'.
+		deprecated: 'This method will be removed. Clients should call name directly'
+		transformWith: '`@receiver categoryName' -> '`@receiver name'.
 	^ self name
 ]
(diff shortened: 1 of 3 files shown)


src/Deprecated80/RPackage.extension.st
```diff
@@ -0,0 +1,8 @@
+Extension { #name : #RPackage }
+
+{ #category : #'*Deprecated80' }
+RPackage >> categoryName [
+	self
+		deprecated: 'This method will be removed. Clients should call name directly'.
```
jecisc: We could have an automatic rewrite for this one.

jecisc (changes requested):

src/Deprecated80/RPackage.extension.st
MarcusDenker: yes, that would be good.

jecisc:
```
transformWith: '`@receiver categoryName' -> '`@receiver name'
```

MarcusDenker:
Bootstrap crashes with:

[32] MessageNotUnderstood: RPackage>>categoryName
[32] RPackage(Object)>>doesNotUnderstand: #categoryName
[32] RPackageOrganizer>>validateCanBeAddedPackage:

MarcusDenker:
I think we need to fix al the callers that are needed during the image build process.

First one to fix is the on evalidateCanBeAddedPackage:, then running the tests again will show if that is enough

MarcusDenker (approved):
ok

PR #4124, opened 2019-07-26
2846-MCWorkingCopyBrowserTest-should-be-in-a-test-package

### LINKED ISSUE #2846: MCWorkingCopyBrowserTest should be in a test package
From FogBugz: https://pharo.fogbugz.com/f/cases/22779/MCWorkingCopyBrowserTest-should-be-in-a-test-package

MCWorkingCopyBrowserTest should be packaged into an own "MonticelloGUI-Tests" package.

Currently it is bundled in the GUI package - instead of having an own test package

ADDITIONALLY:
- Also a #tearDown should be implemented
- the ivar "b" should be named "browser"
- class should be commented

GDayne:
Fixed #2846 
moved the test Class to a new package
changed the instance variable name (b to browser)
added the package to the baseline


CafeKrem:
we don't need to create a tearDown, the browser will be garbage collect

GDayne:
CI failing in windows-32, Not related with this

MarcusDenker (approved):
ok

PR #4133, opened 2019-07-26
4025-some-callers-of-compiledMethodAtifPresentifAbsent-do-not-need-ifAbsent-part

### LINKED ISSUE #4025: some callers of compiledMethodAt:ifPresent:ifAbsent: do not need ifAbsent: part
some senders of compiledMethodAt:ifPresent:ifAbsent: use a empty block or nil as the ifAbsent: argument.

Now that we have  #compiledMethodAt:ifPresent:, these callers can just use that method (that is, one can delete the "ifAbsent: nil" or "ifAbsent: []".

(This is a very simple issue, nice to do for a beginner)

carlotxra:
fixes #4025

Use compiledMethodAt:ifPresent where possible instead of compiledMethodAt:ifPresent:ifAbsent:


Ducasse:
Thanks!

PR #4134, opened 2019-07-27
FileTest cleanup

jgfoster:
Cleanup comment for `File`.
Refactor `FileTest` for code critic issues and to use block rather than symbol as argument to `#ifNotNil:`.
Remove duplicate assignment.


Ducasse:
Thanks James!

PR #4138, opened 2019-07-27
4136-Replace-removal-and-remove-protocols-by-removing

### LINKED ISSUE #4136: Replace #removal and #remove protocols by #removing
We have 2 protocols remove and removal in the image. They should be renamed removing. 

We should also add a release test to ensure we do not add them once again in the Pharo image later.

The goal is to reduce the number of protocols used in the default image and make the protocol proposal more usable.

jecisc:
Updated removing related protocols and introduce non regression test.

Fixes #4136


Ducasse:
strnage none of the build run.

Issue #4140, opened 2019-07-28
Pharo contains deprecated methods that are not in a deprecated package


jecisc:
I think we should ensure that deprecated methods are in deprecated packages to make their removal easier. 

I propose to correct it for all current deprecated methods and then add a release test. 

We can find them like this:

```Smalltalk
Smalltalk allClasses select: [ :c | c methods anySatisfy: [ :m | m isDeprecated and: [ m package isDeprecated not ] ] ]
```

Note that it might take some time to correct the problem since some of them are in Iceberg.

PharoProject:
I don't think this is a good move, because it complexifies a lot the
dependencies and the baselines.

Removal is also not easier, is just the same in my opinion. We could do a
script as the one you showed removing the methods.

I think this idea of having a deprecated package is a remnant of the old
build system using mutation of images and mc packages.

Le dim. 28 juil. 2019 à 22:09, CyrilFerlicot <notifications@github.com> a
écrit :

> I think we should ensure that deprecated methods are in deprecated
> packages to make their removal easier.
>
> I propose to correct it for all current deprecated methods and then add a
> release test.
>
> We can find them like this:
>
> Smalltalk allClasses select: [ :c | c methods anySatisfy: [ :m | m isDeprecated and: [ m package isDeprecated not ] ] ]
>
> Note that it might take some time to correct the problem since some of
> them are in Iceberg.
>

jecisc:
Well, at least something needs to be done because currently at each release we only remove the DeprecatedX0 package and some methods in Pharo 8 are deprecated since multiple versions and were never removed.

I liked the idea to move them in DeprecatedX0 because since it's the last package loaded it force us to remove the invocations on the deprecated methods.

PharoProject:
let's remove old deprecated methods then :)

Le dim. 28 juil. 2019 à 22:23, CyrilFerlicot <notifications@github.com> a
écrit :

> Well, at least something needs to be done because currently at each
> release we only remove the DeprecatedX0 package and some methods in Pharo 8
> are deprecated since multiple versions and were never removed.
>
> I liked the idea to move them in DeprecatedX0 because since it's the last
> package loaded it force us to remove the invocations on the deprecated
> methods.

jecisc:
[...]

PR #4141, opened 2019-07-28
Deprecate-some-class-creation-methods-using-categories

jecisc:
Deprecate Class>>#subclass:instanceVariableNames:classVariableNames:category: and Class>>#subclass:instanceVariableNames:classVariableNames:poolDictionaries:category:


guillep:
The build is broken because there are users of that method.

Still, please, let's not put stuff in the deprecated package anymore...

guillep (changes requested):
The build is broken because there are users of that method.

Still, please, let's not put stuff in the deprecated package anymore...

Ducasse:
Guille for me such a core method should be deprecated with a transform rule so that we can easily import old code.

Ducasse:
And this are the typical methods that I would like to add to the migrator package.
Now all the broken code in Morphic and GT can be just removed. Code not working before should be removed.

jecisc:
See discussion at for the Deprecated package: https://github.com/pharo-project/pharo/issues/4140#issuecomment-518162948

For the broken build, it's weird because I thought I fixed all users. I'll check when I'll have the time.

Ducasse:
To see if it can compile.

jecisc:
Apparently, the bug occurs while loading source files during the bootstrap.

jecisc:
I do not find the file `bootstrap-cache/st-cache/Monticello.st` is it generated? Or stored somewhere else?

jecisc:
Ok, apparently it's generated here: https://github.com/jecisc/pharo/blob/Deprecate-some-class-creation-methods-using-categories/bootstrap/src/Pharo30Bootstrap.package/PBBootstrap.class/instance/exportPackages.withBlacklistedClasses.withBlacklistedMethods.usingInitializeScript.intoFile..st

<blockquote><div> GitHub</div><div><strong><a href="https://github.com/jecisc/pharo">jecisc/pharo</a></strong></div><div>Contribute to jecisc/pharo development by creating an account on GitHub.</div></blockquote>

jecisc:
@guillep On which version of Pharo is the bootstrap launched?

The problem is that the bootstrap generates a script that uses the old API. If the bootstrap runs on Pharo 7 then I think we should backport the change to Pharo 7 before we proceed with this PR.

jecisc:
Hum... That's not so simple because all class creation patterns do not have an equivalent with #package: instead of #category:...

I'll close this PR and we should do a pass on class creation methods

guillep:
> @guillep On which version of Pharo is the bootstrap launched?
> 
> The problem is that the bootstrap generates a script that uses the old API. If the bootstrap runs on Pharo 7 then I think we should backport the change to Pharo 7 before we proceed with this PR.

On Pharo 6.1 :) But that does not matter: if you're bootstrapping pharo8 commit xxx, you'll end up having the version of monticello present in pharo8 commit xxx.

guillep:
In other words, if you break monticello, it will not bootstrap :P

jecisc:
@guillep The problem here is that during the bootstrap, the code that is exported use the API with `category:` instead of `package:`.

So, if we plan to remove later this API, we need to update the exporter. I did another PR as a first step toward the goal of updating the ST code writer.

jecisc:
What I mean is that it's not Monticello that is broken, but the MCStWriter generate this kind of code:

```
Boolean subclass: #False
	instanceVariableNames: ''
	classVariableNames: ''
	poolDictionaries: ''
	category: 'Kernel-Objects'!
```

Instead of :

```Smalltalk
Boolean subclass: #False
	instanceVariableNames: ''
	classVariableNames: ''
	poolDictionaries: ''
	package: 'Kernel-Objects'!
```

PR #4142, opened 2019-07-28
Deprecate-unused-methods-in-obsolete-proctocol

jecisc:
Mark methods in obsolete protocol as deprecated.


Ducasse:
Thanks for cleaning this! I think that we can remove such methods.

PR #4143, opened 2019-07-28
Update-protocol-of-GTSpotterStep-isEmpty

jecisc:
Move GTSpotterStep>>isEmpty to *Deprecated80 to remove zz-deprecated protocol.

### DIFF
--- a/src/Deprecated80/GTSpotterStep.extension.st
+++ b/src/Deprecated80/GTSpotterStep.extension.st
@@ -1,0 +1,7 @@
+Extension { #name : #GTSpotterStep }
+
+{ #category : #'*Deprecated80' }
+GTSpotterStep >> isEmpty [
+	self deprecated: 'Use hasCandidates instead'.
+	^ self hasCandidates not
+]
--- a/src/GT-Spotter/GTSpotterStep.class.st
+++ b/src/GT-Spotter/GTSpotterStep.class.st
@@ -255,12 +255,6 @@ GTSpotterStep >> isCategorySelected: aProcessor [
 	^ self hasSelected and: [ self selected processor == aProcessor ]
 ]
 
-{ #category : #'zz-deprecated' }
-GTSpotterStep >> isEmpty [
-	self deprecated: 'Use hasCandidates instead'.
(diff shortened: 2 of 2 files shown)


guillep (changes requested):
Please, let's not put stuff in the deprecated package anymore...

Ducasse:
Guille you mean that we remove it?
Ok for me.

guillep:
> Guille you mean that we remove it?

I don't know, when was this method originally deprecated? Is this public API and there are any callers?

My understanding is that @jecisc wants to make sure that deprecations are effectively removed after a release cycle (like, when the pharo8 cycle starts, we can remove pharo7 deprecations).

My only point is that moving the deprecations to a single `DeprecatedXX` package does not solve that issue per se. We can do the same with a simple script (`methods select: #isDeprecated thenDo: #removeFromSystem`).

I understand however that having a package with deprecations from old versions can ease migration (thus the https://github.com/Ducasse/Migrator package, no?).

Still, **IF** we move deprecations to the Deprecated package (which I don't see the value due if we have migrator), the same commit/PR has to ensure that the deprecated class/method/trait is not used, otherwise, since the deprecated package is loaded at the end, it will break the build due to an unfulfillable dependency (happened in another PR with the class building messages).

Finally, being I'm less conservative I could also argue that deprecations could only apply yo public API methods. Internally used methods do not really require it... But sometimes (like this one) we cannot know because it is a super common message... :)

<blockquote><div> GitHub</div><div><strong><a href="https://github.com/Ducasse/Migrator">Ducasse/Migrator</a></strong></div><div>A list of extensions with automatic transformations to ease migration. - Ducasse/Migrator</div></blockquote>

Ducasse:
I see.
What I would like 
- one package with public API deprecated method => Pushed in Migrator after
- fixing all users in Pharo (this is why I like that cyril changed the tests to not use category anymore -- I did the same internally to the class first and got distracted).
- remove broken not working old deprecated method. 
-

jecisc:
See discussion at: https://github.com/pharo-project/pharo/issues/4140#issuecomment-518162948

PR #4144, opened 2019-07-28
2799-rename-NuclearSiteTestingClassTest

### LINKED ISSUE #2799: rename NuclearSiteTestingClassTest
Some test create classes with the substring NuclearSiteTestingClassTest

--> find a better name that communicates that a) this class is created by a test and b) by which package

carlotxra:
Fixes #2799

Renamed package and class names created via the tests.


Ducasse:
This is a good idea. Could you rename Programatically into Programmatically? 
And we integrate it. 
Tx

PR #4145, opened 2019-07-28
3690-TestCasesimulateKeyStroke-and-TestCasesimulateKeyStrokes-should-not-exists-there

### LINKED ISSUE #3690: TestCase>>#simulateKeyStroke: and TestCase>>#simulateKeyStrokes: should not exists there
These methods are specific to Morphic testing.

Maybe extracting them to a trait to apply on classes requiring it.

carlotxra:
Fixes #3690

The added level of indirection is not adding anything. Remove test support methods from TestCase and inline calls to use ActiveHand directly.


Ducasse:
Hi 
We should not directly access ActiveHand like that. 
In general we should not directly access such kind of global and better access them via a message. 
Check the bad aspect of StandardFonts for example. 
So I close this ticket but thanks for your contribution. 
If you want to help Pharo why not syncing with the core team because we could pair-improve Pharo :)

Ducasse:
This is still the case.

PR #4146, opened 2019-07-29
Separate Network-Tests-Mail into new package

jgfoster:
I'd like to be able to load Network-Tests into a minimal image without these tests since they test things that aren't in the minimal image (MailAddressParser, MailAddressTokenizer, and MailMessage).  This appears to do it but I've not created a new package before so anticipate issues. Advice and guidance is welcome! Thanks.


bencoman:
I guess you do the file rename from the command line, since the class-category was not updated.
Class {
	#name : #MailAddressParserTest,
	#superclass : #TestCase,
	#category : #'Network-Tests-RFC822'
}

jgfoster:
Actually, I didn't rename the file. Should the category be #'Network-Tests-Mail-RFC822'?

bencoman:
Oh, I saw this in the File Changed tab...
image

and wondered how the class(/file) changed packages(/directory) without the "Category:" of the class being changed to match the package (as seen inside the file) .  Any ideas?  

> 'Network-Tests-Mail-RFC822'

It didn't occur to me, but yes, it might be worth not throwing that info away.

Ducasse:
Thanks James. I like this change!

PR #4035, opened 2019-07-22
4026-migration-of-CriticBrowser-Spec-to-Spec-2-

CafeKrem:
migration of criticBrowser

PR #4148, opened 2019-07-29
4026 migration of critic browser spec to spec 2 

CafeKrem:


CafeKrem:
Cyril reviews it. for more information see PR #4035

CafeKrem:
the CI is broken but I don't understand why.
Can someone explain to me?

VincentBlondeau:
That the error:
```
Loaded -> GT-SUnitDebugger-tonel.1 --- tonel:///builds/workspace/uest_and_branch_Pipeline_PR-4148/32/src --- cacheError: This package depends on the following classes:

  CriticWorkingConfiguration

  CriticWorkingConfiguration>>#reset
```

Ducasse:
Clement est ce que tu peux regarder car j'aimerai integrer ton travail.

Ducasse:
OK I will check that the build tell us.

CafeKrem:
it's already the same error, I think it's because I forgot to add to group: method of the baseline

CafeKrem:
so the CI break on 
window , unix 
and on osx , the error is  a 'Bus error'

if I did something someone can tell what is my mistake ?

Ducasse:
Relaunching the build

CafeKrem:
same CI break with the same error

Ducasse:
Ouch I will try to integrate another PR to see.

guillep:
Now there are just broken tests, that seem unrelated.

@CafeKrem are you writing tests about your new UIs?

CafeKrem:
yes

src/BaselineOfIDE/BaselineOfIDE.class.st
```diff
@@ -231,6 +231,10 @@ BaselineOfIDE >> baseline: spec [
 		spec package: 'Glamour-FastTable'.
 		spec package: 'Glamour-SpecIntegration'.
 	
+		spec package: 'Tool-CriticBrowser'.
+		spec package: 'Tool-CriticBrowser-TestEnvironment'.
+		spec package: 'Tool-CriticBrowser-Tests'.
+
```
guillep, line 235: Could we put this in a separate baseline?

src/BaselineOfIDE/BaselineOfIDE.class.st
```diff
@@ -645,9 +650,13 @@ spec group: 'QualityAssistant-Core-Group' with: #(
 	'Manifest-Core-Group' "dependency"
 	'Glamour-Core-Group'	"dependency"
 	'DependencyAnalyser-Core-Group' "dependency"
-	'GT-Core-Group'	"dependency"
-	'Renraku'		
+	'Renraku'	
 	'Tool-CriticBrowser'
+	'Tool-CriticBrowser-TestEnvironment'
+	'Tool-CriticBrowser-Tests'
```
guillep: Splitting this in a separate baseline will simplify this duplication (and make this BaselineOfIDE smaller too :))

src/Spec2-Core/SpAbstractTreeSingleSelectionMode.class.st
```diff
@@ -20,7 +20,7 @@ SpAbstractTreeSingleSelectionMode class >> on: aPresenter [
 
 { #category : #selection }
 SpAbstractTreeSingleSelectionMode >> clearSelection [
-
+	selection ifEmpty: [ ^ self ].
```
guillep, line 23: Q: is this really needed? Is this to avoid an event when re-assigning the selection? If so, why is this not responsibility of the observable slot (if it is an observable slot)?

src/Spec2-Core/SpActionBarPresenter.class.st
```diff
@@ -15,11 +15,12 @@ SpActionBarPresenter class >> adapterName [
 
 { #category : #adding }
 SpActionBarPresenter >> add: aButtonPresenter [
-	
-	(items 
-		at: #start
-		ifAbsentPut: [ OrderedCollection new ] ) 
-		add: aButtonPresenter
+	self startItems add: aButtonPresenter
```
guillep, line 18: what is `startItems`?

src/Spec2-Core/SpActionBarPresenter.class.st
```diff
@@ -15,11 +15,12 @@ SpActionBarPresenter class >> adapterName [
...
+]
+
+{ #category : #adding }
+SpActionBarPresenter >> addAll: aCollection [
+	self startItems addAll: aCollection
```
guillep, line 23: what is `startItems`?

src/Spec2-Core/SpActionBarPresenter.class.st
```diff
@@ -43,3 +44,8 @@ SpActionBarPresenter >> items [
 
 	^ items value
 ]
+
+{ #category : #adding }
+SpActionBarPresenter >> startItems [
+	^ items at: #start ifAbsentPut: [ OrderedCollection new ]
```
guillep, line 50: what is `#start` in here?

src/Spec2-Core/SpTreeTablePresenter.class.st
```diff
@@ -183,6 +183,13 @@ SpTreeTablePresenter >> itemAtPath: anArray ifAbsent: aBlock [
 		do: [ aBlock value ]
 ]
 
+{ #category : #api }
+SpTreeTablePresenter >> refreshTree [
+	"I refresh the content of the tree. I can be useful for example if the children of the tree changed."
+
+	self withAdapterDo: [ :adptr | adptr refreshTree ]
```
guillep, line 190: This is not good. The presenter should not talk to an adapter, it should send an event. This will otherwise break compatibility with other adapters.

src/Spec2-Core/SpTreeTablePresenter.class.st
```diff
@@ -183,6 +183,13 @@ SpTreeTablePresenter >> itemAtPath: anArray ifAbsent: aBlock [
...
+]
+
 { #category : #initialization }
 SpTreeTablePresenter >> registerEvents [
 	self whenMenuChangedDo: [ self withAdapterDo: #updateMenu ]
```
guillep, line 195: This is not good. The presenter should not talk to an adapter, it should send an event. This will otherwise break compatibility with other adapters.

src/Tool-CriticBrowser-Tests/CBAbstractTest.class.st
```diff
@@ -0,0 +1,71 @@
...
+}
+
+{ #category : #testing }
+CBAbstractTest class >> isAbstract [ 
+	^ self = CBAbstractTest 
+]
+
+{ #category : #setUp }
+CBAbstractTest >> criticBrowserClass [
+	^ CBCriticBrowser
```
guillep: Are there different implementations of this `criticsBrowserClass`? I imagine this is to give a possibility of redefine this class in subclasses to reuse the test? If that's the case, I don't think that really works in the long term. Imagine that tomorrow I have a new criticsBrowserClass `MyCriticBrowser` that requires to be instantiated with other than `openOnRules:environment:`. I cannot reuse the test at all, I have to rewrite it.

What pays more in the long term is to have a factory method

`newCriticBrowser...`

that subclasses can redefine and implement differently if needed.

src/Tool-CriticBrowser-Tests/CBCriticBrowserTest.class.st
```diff
@@ -0,0 +1,94 @@
+Class {
+	#name : #CBCriticBrowserTest,
+	#superclass : #CBAbstractTest,
+	#category : #'Tool-CriticBrowser-Tests'
+}
```
guillep, line 5: Good!!!! :D

src/Tool-CriticBrowser-Tests/CBCriticsCacheTest.class.st
```diff
@@ -0,0 +1,195 @@
+Class {
+	#name : #CBCriticsCacheTest,
+	#superclass : #TestCase,
+	#instVars : [
+		'criticCache',
+		'rule'
+	],
+	#category : #'Tool-CriticBrowser-Tests'
+}
```
guillep, line 9: Good!

src/Tool-CriticBrowser/CBCriticBrowser.class.st
```diff
@@ -0,0 +1,698 @@
...
+		'removeTestCase',
+		'noteBookPresenter',
+		'criticCache',
+		'noteBookPageCriticErrorPresenter',
+		'noteBookPageRuleCommentPresenter',
+		'criticActionBar',
+		'ruleActionBar'
+	],
+	#category : #'Tool-CriticBrowser-Base'
+}
```
guillep, line 100: This class looks a bit fat for a presenter, no? It looks it implements the UI concerns but also system event handling and lots of decisions on the critics... Wouldn't it be good to split that into a model class?

src/Tool-CriticBrowser/CBCriticBrowser.class.st
```diff
@@ -0,0 +1,698 @@
...
+{ #category : #'system Annoucements-Handle' }
+CBCriticBrowser >> onWindowClosed [
+	self unregisterFromAnnouncements.
+	criticCache cacheChange
+		ifTrue: [ (UIManager default
+				confirm:
+					'Do you want log all wrong violations in the Manifests
+before closing the Critics Browser ?')
+				ifTrue: [ criticCache logInManifest ] ]
+]
```
guillep, line 590: Why do we have this and also

```smalltalk
CBCriticBrowser >> initializeWindow: aWindowPresenter [
	aWindowPresenter
		initialExtent: 900 @ 700;
		title: 'Critic Browser';
		whenClosedDo: [ self unregisterFromAnnouncements ]
]
```
?

src/Tool-CriticBrowser/CBCriticBrowser.class.st
```diff
@@ -0,0 +1,698 @@
...
+	 
+	self unregisterFromAnnouncements.
+	SystemAnnouncer uniqueInstance weak
+		when: ClassAdded send: #handleClassAdded: to: self;
+		when: ClassModifiedClassDefinition send: #handleClassModified: to: self;
+		when: ClassRemoved send: #handleClassRemoved: to: self;
+		when: MethodAdded send: #handleMethodAdded: to: self;
+		when: MethodModified send: #handleMethodModified: to: self;
+		when: MethodRemoved send: #handleMethodRemoved: to: self.
+	self announcer when: WindowClosed send: #onWindowClosed to: self" Unmatched " "in comment".
```
guillep, line 628: It looks like the `" Unmatched " "in comment"` is a leftover?

src/Tool-CriticBrowser/CBCriticBrowser.class.st
```diff
@@ -0,0 +1,698 @@
...
+{ #category : #'system annoucements' }
+CBCriticBrowser >> unregisterFromAnnouncements [
+
+	SystemAnnouncer uniqueInstance unsubscribe: self
+]
+
+{ #category : #utilities }
+CBCriticBrowser >> updateTree [
+"	self needRebuild: false.
+	self buildWithSpec"
```
guillep, line 697: Is this method needed at all? It's empty with commented code...

src/Tool-CriticBrowser/CBLogRuleAction.class.st
```diff
@@ -0,0 +1,29 @@
...
+
+{ #category : #accessing }
+CBLogRuleAction class >> groupPriority [ 
+	^ 20
+]
+
+{ #category : #testing }
+CBLogRuleAction class >> runOn: aCriticBrowser [
+	self flag: #pharoFixMe.	"In the long term the UI manager should not be used in Spec application."
+	(UIManager default
```
guillep, line 20: This only works in Spec morphic. Shouldn't this be delegated to the application?

src/Tool-CriticBrowser/CBReapplyRuleCommand.class.st
```diff
@@ -0,0 +1,24 @@
...
+	#superclass : #CBRuleCommand,
+	#category : #'Tool-CriticBrowser-Commands'
+}
+
+{ #category : #hooks }
+CBReapplyRuleCommand >> execute [
+	self selectedRule isComposite
+		ifTrue: [ self selectedRule rules
+				do: [ :each | self context reapplyRule: each ] ]
+		ifFalse: [ self context reapplyRule: self selectedRule ]
```
guillep: Why the if? Cannot we just delegate in the rule? Then a composite rule will apply each of the subrules, simple rules will just apply themselves. This will simplify the code and allow rules to be composite in more than one level (your code does not work with more than one level of composition :))

src/Tool-CriticBrowser/CBResetRuleAction.class.st
```diff
@@ -0,0 +1,32 @@
...
+
+{ #category : #accessing }
+CBResetRuleAction class >> groupPriority [ 
+	^ 1 
+]
+
+{ #category : #testing }
+CBResetRuleAction class >> runOn: aCriticBrowser [
+	self flag: #pharoFixMe.	"In the long term the UI manager should not be used in Spec application."
+	(UIManager default
```
guillep, line 20: Again, shouldn't this delegate to the app? otherwise it will never work in Gtk or other backedns

src/Tool-CriticBrowser/CBSelectPackageBrowser.class.st
```diff
@@ -0,0 +1,174 @@
...
+		add:
+			(SpBoxLayout newHorizontal
+				add: #addButton;
+				add: #addAllButton;
+				add: #removeButton;
+				add: #removeAllButton;
+				add: #nextButton;
+				yourself)
+			withConstraints: [ :contraint | contraint height: 25 ];
+		yourself
```
guillep, line 40: This method has a lot of things in common with the one in the superclass. Can't it be refactored to use super?

src/Tool-CriticBrowser/CBSelectPackageBrowser.class.st
```diff
@@ -0,0 +1,174 @@
...
+{ #category : #private }
+CBSelectPackageBrowser >> removeAllItems [
+	toSelectPresenter items: (self sort: toSelectPresenter items withAll: selectedPresenter items).
+	selectedPresenter items: #()
+]
+
+{ #category : #private }
+CBSelectPackageBrowser >> removeItems [
+	selectedPresenter selectedItems
+		ifNotNil: [ :items |
```
guillep: Isn't `selectedPresenter` a list? Then selectedItems should never be nil (it's either a collection with elements or empty), then the if is not necessary, is it?

src/Tool-CriticBrowser/CBUnBanRuleCommand.class.st
```diff
@@ -0,0 +1,23 @@
...
+Class {
+	#name : #CBUnBanRuleCommand,
+	#superclass : #CBRuleCommand,
+	#category : #'Tool-CriticBrowser-Commands'
+}
+
+{ #category : #hooks }
+CBUnBanRuleCommand >> execute [
+	self selectedRule isComposite
+		ifTrue: [ self selectedRule rules do: [ :r | self context allPackages do: [ :package | self context removeFalsePositiveRule: r forPackage: package ] ] ]
```
guillep: Same as before. Couldn't you apply a normal composite here?

guillep (changes requested):
Some comments while I was having lunch :)

src/Tool-CriticBrowser/CBResetRuleAction.class.st
CafeKrem: https://github.com/pharo-spec/Spec/issues/502

src/Spec2-Core/SpAbstractTreeSingleSelectionMode.class.st
CafeKrem: it's a @jecisc commit

src/Spec2-Core/SpActionBarPresenter.class.st
CafeKrem: it's a @jecisc commit

src/Spec2-Core/SpActionBarPresenter.class.st
CafeKrem: it's a @jecisc commit

src/Tool-CriticBrowser/CBReapplyRuleCommand.class.st
CafeKrem: it's a good idea fix in progress

src/Tool-CriticBrowser/CBSelectPackageBrowser.class.st
CafeKrem: in my case, I think that using refactoring is a bad idea because in CBSelectPackageBrowser defaultSpec I insert 2 widgets between 2 group of widget.
in case of adding at the start or at the end of the specDefault we must use refactoring

src/Tool-CriticBrowser/CBSelectBrowser.class.st
```diff
@@ -0,0 +1,134 @@
...
+				yourself);
+		add:
+			(SpBoxLayout newHorizontal
+				add: #addButton;
+				add: #addAllButton;
+				add: #removeButton;
+				add: #removeAllButton;
+				add: #nextButton;
+				yourself)
+			withConstraints: [ :contraint | contraint height: 25 ];
```
estebanlm, line 45: this approach is incorrect.
To create action bars as this forces the size of the component and then is hard to scale later. 
There is already a presenter made to contain action buttons: `SpActionBarPresenter`, which receive buttons as desired: 
```Smalltalk
actionBar := self newActionBar
		add: (self newButton
			label: 'Add'; 
			icon: (self iconNamed: #add);
			help: 'Add a new element to the list.';
			action: [ self doAdd ];
			yourself);
		yourself
```

and the advantage is that we can style it later.

estebanlm (changes requested):
please replace the "ad hoc" action bars with the correct presenter.

src/Tool-CriticBrowser/CBSelectBrowser.class.st
estebanlm: and YES, I know documentation is still lacking ;)

src/Spec2-Core/SpActionBarPresenter.class.st
estebanlm: action bars can have elements at `start` or at `end` of the bar. 
my original implementation had this in a dictionary and I guess @jecisc moved that into separated variables/accessors.

( btw, in the future action bars can also have `center` position, just not implemented yet)

src/Tool-CriticBrowser/CBSelectBrowser.class.st
CafeKrem: I'm working on it Thanks you! it will be in the next commit

src/Spec2-Core/SpAbstractTreeSingleSelectionMode.class.st
jecisc: It is indeed to not launch multiple time events when re-assigning the selection. 

Maybe it would be better to put the behaviour in the slot but I wanted to get soon in a state where this PR can be integrated. ("Better is the enemy of good")

I'll add an issue to the Spec repo.

jecisc: https://github.com/pharo-spec/Spec/issues/676

src/Spec2-Core/SpActionBarPresenter.class.st
jecisc: Esteban answered.

src/Spec2-Core/SpActionBarPresenter.class.st
jecisc: Esteban answered.

src/Spec2-Core/SpTreeTablePresenter.class.st
jecisc: This way is currently used in Spec. If it is not good we should remove it everywhere and we should have another system to do that.

src/Tool-CriticBrowser/CBLogRuleAction.class.st
jecisc: It's the plan. An issue is opened but we have no time currently to do it.

CafeKrem:
please, can someone integrate it into Pharo?

pavel-krivanek:
restarting CI check...

VincentBlondeau:
Conflict to solve

estebanlm:
this needs a lot of work and I think it needs to be integrated with newtools project instead. 
I will add in my TODO recover this work on P9.

PR #4151, opened 2019-07-30
First pass on compiler tests for full blocks

guillep:
Integrating full blocks for real requires that we do a pass on the compiler tests.
The compiler tests are right now using the default compiler, thus they cannot test the full blocks.

This PR introduces:
 - more compiler tests
 - refactored compiler tests to configure the bytecode backend and optimizations
 - extending SUnit and Parametrized tests to allow combination/multiplication of test matrixes

### DIFF
--- a/src/Kernel/Process.class.st
+++ b/src/Kernel/Process.class.st
@@ -40,14 +40,6 @@ Class {
 }
 
 { #category : #'process specific' }
-Process class >> psKeysSema [
-	"Isolate handling of class variable"
-
-	PSKeysSema ifNil: [ PSKeysSema := Semaphore forMutualExclusion ].
-	^PSKeysSema
-]
-
-{ #category : #'process specific' }
 Process class >> allocatePSKey: aPSVariable [
 
 	"Add a new process-specific key. 
@@ -98,6 +90,14 @@ Process class >> forContext: aContext priority: anInteger [
 	newProcess priority: anInteger.
 	Processor activeProcess installEnvIntoForked: newProcess.
 	^newProcess
+]
+
+{ #category : #'process specific' }
+Process class >> psKeysSema [
+	"Isolate handling of class variable"
+
+	PSKeysSema ifNil: [ PSKeysSema := Semaphore forMutualExclusion ].
+	^PSKeysSema
 ]
 
 { #category : #'process specific' }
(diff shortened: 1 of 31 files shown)


guillep:
In the middle, I've found a mismatch between the optimized and non-optimized versions of whileTrue/False

```
b1 := [ 4 < 3 ].
b2 := [  ].
b1 whileTrue: b2.
  => "the receiver block"

[ 4 < 3 ] whileTrue: [  ].
  => nil
```

Which can be also seen in
```
OpalCompiler new
	evaluate: '[ 4 < 3 ] whileTrue: [  ]'.
  => "nil"
	
OpalCompiler new
	options: #( -optionInlineWhile );
	evaluate: '[ 4 < 3 ] whileTrue: [  ]'.
  => "the receiver block"
```

I propose that the `whileTrue:` and `whileFalse:` methods do return nil to behave as the optimized version.

src/Kernel/Process.class.st
```diff
@@ -39,14 +39,6 @@ Class {
...
 }
 
-{ #category : #'process specific' }
-Process class >> psKeysSema [
-	"Isolate handling of class variable"
-
-	PSKeysSema ifNil: [ PSKeysSema := Semaphore forMutualExclusion ].
-	^PSKeysSema
-]
-
```
guillep, line 49: This is super strange, I don't understand why this method got moved down...
Actually, my commit is sorting it. So somebody commited non-sorted methods?

Ducasse: I saw this code in recent PRs but I do not remember which one.

guillep: https://github.com/pharo-project/pharo/pull/4106/files

Ducasse:
Guille do we wait that all the tests are green? Because there some broken tests.

guillep:
Yes, I have some little stuff to do here yet. Please do not integrate until tests are green :)

Ducasse:
So we integrate it?

guillep:
I'd say yes. There is a failure in the build but due to the following test that fails from time to time.
```
0xffe8fc30 M [] in FLContextSerializationTest>testDoIt 0x8deb498: a(n) FLContextSerializationTest
```

PR #4153, opened 2019-07-30
4152-Cleanups-in-UIManager

### LINKED ISSUE #4152: Cleanups in UIManager
- no dots at end of methods
- unify comments to start uppercase and end with dot
- add missing return
- unify formatting of similar methods

astares:
Cleanups in UIManager

- no dots at end of methods
- unify comments to start uppercase and end with dot
- add missing return
- unify formatting of similar methods

Fix #4152


Ducasse:
Thanks torsten. 
For the record I add always a space after ^ 
so...

We will integrate soon an alternate pretty printer with better default.
I'm not sure yet that we can pretty print everything but I will try.

PR #4163, opened 2019-08-02
Adding a variation of trait sequence composition that has preference for one of the members.

tesonep:
This preferred sequence is valid as it implements the same semantics than removing the methods duplicated before:

Let's think T2 defines #m1 and #m2.

(T1 + T2) withPrecedenceOf: T2 ==> 
(T1 - {#m1. #m2}) + T2


Ducasse:
I see. Thanks for the explanation.

PR #4167, opened 2019-08-02
Add submenu, and new shortcuts

### LINKED ISSUE #3737: Calypso menus should have a "Refactoring" sub menu
Currently, refactorings are inlined in the menus. It would be better for ergonomy to have a submenu "Refactorings" instead.

Lin777:
Fixes #3737 

Add a submenu named: <em> More refactorings </em> for the less used refactorings, so leave the following refactorings in the main menu according to the type:

 **- Method:** Rename, Classify methods, and Remove.
  **- Class:** Rename, generate accessors, Remove.

Any suggestion for the modification in the menu structure, the order of the submenu, the content of the submenu, the addition or modification of shortcuts is very welcome.

Link video after add submenu: https://www.youtube.com/watch?v=AuycJsdfvCI


astares:
Calling the menu item "More refactorings" implies that the original menu is all about refactorings which  is not the case.

Just call the menu "Refactorings" only as it is known from other IDE's like Eclipse and where such functionality is available from similar context menu.

Ducasse:
Tx. BTW in Pharo 80 I get an error when I rename a class (I will check and confirm with the latest version).

Issue #4088, opened 2019-07-24
Commit should have been implemented in IceUnknownCommit


jgfoster:
In Pharo build 453.sha.43f21722, select any package and attempt to commit:

![image](https://user-images.githubusercontent.com/1577872/61814994-207ef380-adfe-11e9-9889-1309311b1e45.png)

bencoman:
Hi James,

Rather than "any package", can you specify the steps to reproduce one
concrete example in a fresh image.
This will help co-ordinate multiple people working on the same issue.

Also can you test that sequence in latest build 561?

cheers -ben

jgfoster:
The problem still exists in build 561 with AST-Core (and every other package that I tried). I'm not sure what other steps I can provide: in a brand new image, open a System Browser, right click on any package, select Commit, and see walkback. If Commit works for others then what is wrong with my setup? What is IceUnknownCommit?

My workaround is to manually edit files in Git, so it isn't a significant burden.

jgfoster:
It appears that this happens if the image is launched from bootstrap-cache.

PR #4171, opened 2019-08-02
Rename test classes

jgfoster:
This doesn't affect my work, but it seems like it is inconsistent.

### DIFF
--- a/src/Network-Kernel/Socket.class.st
+++ b/src/Network-Kernel/Socket.class.st
@@ -1702,7 +1702,7 @@ Socket >> unregister [
 
 { #category : #waiting }
 Socket >> waitForAcceptFor: timeout [
-	"Wait and accept an incoming connection. Return nil if it falis"
+	"Wait and accept an incoming connection. Return nil if it fails"
 	self waitForConnectionFor: timeout ifTimedOut: [^ nil].
 	^ self isConnected
 		ifTrue:[self accept]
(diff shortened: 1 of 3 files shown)


astares (approved):
OK, thanks!

gcotelli (approved):

jecisc:
I have the impression there is something weird with this PR because the files were not renamed. I have the impression they should have been renamed.

jgfoster:
Yes, renaming the files makes sense. I'll submit something on this soon.

jecisc:
@jgfoster Are you working with a file editor or Pharo itself?
If you are working with a file editor, is there a reason to not work in Pharo? (I ask to know if there is something we might improve in Pharo to reduce barrier to entrance)

jgfoster:
@jecisc Thanks for fixing the file names. And, yes, I'm using a file editor. I admit that I'm not doing things the conventional way, but I've run into issues trying to edit code in Pharo. Using PharoLauncher, when I try to launch the image in the checkout directory, nothing happens; also when I double-click on the image nothing happens. When I use Bootstrap to build images, and launch with PharoLauncher, then I can see and run code but attempting to commit gives an error (#4088). So I've just been editing files then re-running Bootstrap.

jecisc:
If you wish, maybe we can sit together at esug if you are comming this year and check your troubles?

jgfoster:
Yes, that might be helpful. I plan to be there for Sunday's Camp Smalltalk. Thanks!

jecisc:
I plan to be there Sunday afternoon

PR #4181, opened 2019-08-03
Collection-test-cases-should-end-with-Test

jecisc:
Collection test cases should end with Test instead of Tests.


jecisc:
Test failure unrelated.

PR #4183, opened 2019-08-03
STON-test-cases-should-end-with-Test

jecisc:
STON test cases should end with Test instead of Tests.


jecisc:
This might interest you @svenvc

PR #4184, opened 2019-08-03
Zinc-and-Zodiac-test-cases-should-end-with-Test

jecisc:
Zinc and Zodiac test cases should end with Test instead of Tests.


jecisc:
This might interest you @svenvc

PR #4187, opened 2019-08-03
Test-cases-should-end-with-Test-instead-of-Tests

jecisc:
This PR update the last test cases ending with Tests to end with Test instead. All the others are done in other PRs.


jecisc:
When all the PRs around Tests -> Test will be merged, I'll check all Test cases that do not end with 'Test' and change them. Once done, I'll add a release test.

PR #4191, opened 2019-08-05
OSEnvironmentsTest in a new package

jgfoster:
The 'Tests' package gets an error when loaded into a minimal image. This moves the offending test into a new package.


guillep (approved):

jgfoster:
I've merged this into #4195 (it got too complicated keeping each of the repackaged classes in separate PRs.

PR #4195, opened 2019-08-05
Repackage some tests

jgfoster:
Replace #4191 with more repackaging of tests. With these changes we are now able to add the following to a minimal image: 
* Tests,
* RPackge-Tests, and
* Monticello-Tests.

The things moved to different (new or existing) packages were:
* MCSnapshotBrowserTest,
* MCMczInstallerTest,
* OSEnvironmentTest, and
* EventManagerTest.

The file bootstrap/scripts/emulation/shrink.st has been modified so that the various test packages are not all on one line. This simplifies diffs and merges when making changes.

The various BaselineOf* files have been updated so that Bootstrap runs.


Ducasse:
This is cool. I like to see all these better packaging changes. Thanks a lot.

PR #4197, opened 2019-08-05
4196-Code-Completion-Add-sorting-strategies-to-Settings

myroslavarm:
some changes to add a Sorter to Settings

### DIFF
--- a/src/NECompletion/NECPreferences.class.st
+++ b/src/NECompletion/NECPreferences.class.st
@@ -36,6 +36,11 @@ NECPreferences class >> availablePopupShortcuts [
 	^ {Character space shift . Character tab asKeyCombination}
 ]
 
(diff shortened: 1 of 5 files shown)


myroslavarm:
reverse alphabetic is essentially for testing -- as far as i was able to see, with only one sorting strategy the option to choose does not appear in the settings. once we have a better sorting strategy (for example, ML-based), then we'll get rid of reverse alphabetic as it does not serve any logical purpose. otherwise everything should be pretty straight-forward

Ducasse:
Miroslava I will merge it but I would really like if after you can rename the classes because TestSorter is a strange name for a class related to sorting completion.

myroslavarm:
@Ducasse yes, i think one of the next things to do will be to clean up / rename things

PR #4205, opened 2019-08-06
Do-not-flush-all-caches-when-a-class-is-removed

jecisc:
When we remove a class from the system we do not need to flush all the SystemDictionary caches.

This will speed up a little the removal of big packages from the system.

### DIFF
--- a/src/System-Support/SystemDictionary.class.st
+++ b/src/System-Support/SystemDictionary.class.st
@@ -206,21 +206,20 @@ SystemDictionary >> fillCaches [
 
 { #category : #'class and trait names' }
 SystemDictionary >> flushClassNameCache [
-	"Smalltalk flushClassNameCache"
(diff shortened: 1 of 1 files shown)


src/System-Support/SystemDictionary.class.st
```diff
@@ -325,14 +324,19 @@ SystemDictionary >> removeClassNamed: aName [
 	self at: aName asSymbol ifPresent: [:oldClass | oldClass removeFromSystem]
 ]
 
+{ #category : #removing }
+SystemDictionary >> removeFromCaches: aClassName [
```
gcotelli: Rename `aClassName` to `aKey`, because it's  also trying to remove it from `cachedNonClassNames`, so it looks like it's not always a class name.

MarcusDenker (approved):
code looks good

PR #4212, opened 2019-08-06
Fixing Issue 4211: Equality + hash definition made easy

bouraqadi:
Add support for equality as defined in [https://github.com/bouraqadi/PharoMisc/tree/master/Equals](https://github.com/bouraqadi/PharoMisc/tree/master/Equals )

### DIFF
--- a/src/BaselineOfEquals/BaselineOfEquals.class.st
+++ b/src/BaselineOfEquals/BaselineOfEquals.class.st
@@ -1,0 +1,13 @@
+Class {
+	#name : #BaselineOfEquals,
+	#superclass : #BaselineOf,
+	#category : #BaselineOfEquals
+}
+
+{ #category : #baselines }
+BaselineOfEquals >> baseline: spec [
+	<baseline>
+
+	spec for: #'common' do: [
+		spec package: #Equals ]
+]
--- a/src/BaselineOfEquals/package.st
+++ b/src/BaselineOfEquals/package.st
@@ -1,0 +1,1 @@
+Package { #name : #BaselineOfEquals }
--- a/src/Equals/Apple.class.st
+++ b/src/Equals/Apple.class.st
@@ -1,0 +1,25 @@
+Class {
+	#name : #Apple,
+	#superclass : #Fruit,
+	#instVars : [
+		'color'
+	],
+	#category : #'Equals-Examples'
+}
+
+{ #category : #'instance creation' }
+Apple class >> color: aColor [
+	^self new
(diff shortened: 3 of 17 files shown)


astares:
It would be better to keep this as an external loadable package.

Reasons:
 - it is not properly packaged (Tests are not separated in an own package), no need to add Apples and Bananas ;)
 - there default implementation for #valuesToCompareForEquality without a responsibility to implement may result in side effects as by defaults all ivars are used, an explicit implementation is better
 - UI behavior (reference to GrowlMorph) is mixed into more core like classes like Comparisons making it hard to use in headless environments
 - the trait for equals is not really needed (there already is a "Extra"->"Generate equal and hash" one can use)

Sorry - from my perspective I would downvote on the inclusion of this PR into the standard image.

Ducasse:
Hi Noury 
I like the idea because it is much better than the simple bitXor: patterns from K. Beck book. 
Now could you 
- repackage the package: split tests and core.
- avoid to use the growl morph. 
There is a typo in the class comment (implmentation -> implementation). 

@Torsten 
- Does the generate = and hash correctly generate the hash?
- I did not get what you mean by "there default implementation for #valuesToCompareForEquality without a responsibility to implement may result in side effects as by defaults all ivars are used, an explicit implementation is better" this traits is based on a convention. Now if a developer blindly uses without understanding the way to use it, then I would say that this is the fault of the silly dev 
and not the solution proposed by Noury.

bouraqadi:
Will do the change and resubmit a pull request.

Noury

MarcusDenker:
I close this for now then

dionisiydk:
Hi @bouraqadi .
Few points: 
- Can you prefix your classes? Current names for examples could easily conflict with domain names. 
- I suggest to make TEquality to be an abstract trait and introduce TReflectiveEquality with complete implementation based on reflection. It will remove methods from the Object. 
Users will use this trait instead of primitive one if they agree with reflection approach (accept the performance penalty). 
But for the rest of people the TEquality will force to implement #valuesToCompareForEquality. And IDE will assist with that.
- Reflection logic can be simplified  by using slots
- It would be much better to not return a collection for comparison because it generates a garbage and it can be a very bad side effect and completely unexpected. Imagine a big collection where every item will produce a garbage for every hash and equal calls. 
My proposal is to replace #valuesToCompareForEquality with #equalityValuesDo: 
- And please use equality word instead of comparison because existing #compare methods in the image are about order Comparison is not just equality.

PR #4216, opened 2019-08-06
Don't return an OrderedCollection from UIManager class>>default

jgfoster:
Better to return nil. Also, fix a few typos in comments.


MarcusDenker (approved):
ok

PR #4218, opened 2019-08-06
ReTestClassNameShouldEndWithTest-should-also-accept-TestCase

jecisc:
ReTestClassNameShouldEndWithTest should accept test classes ending with 'Test' but also 'TestCase'.


jecisc:
Currently, the rule check that test case ends with `Test`. But in the system, we have almost a hundred test case ending with `TestCase`. Most of them are abstract classes and IMO it's not a bad thing. Thus, in this PR I changed the rule to also accept `TestCase` as a suffix.

Ducasse:
agreed

PR #4224, opened 2019-08-07
4223-Lint-clean-RubCharacterBlockScanner-cr

### LINKED ISSUE #4223: Lint clean RubCharacterBlockScanner >>cr
Already fixed CharacterBlockScanner in issue #4221 

Same issue in RubCharacterBlockScanner which seems to be a copied method

astares:
Lint clean RubCharacterBlockScanner >>cr 

Fix #4223

### DIFF
--- a/src/Rubric/RubCharacterBlockScanner.class.st
+++ b/src/Rubric/RubCharacterBlockScanner.class.st
@@ -91,7 +91,7 @@ RubCharacterBlockScanner >> cr [
 	CharacterBlocks, sometimes including the carriage return (cursor is at 
 	the end) and sometimes not (cursor is in the middle of the text)."
 
-	((characterIndex ~= nil
+	((characterIndex notNil
 		and: [characterIndex > text size])
 			or: [(line last = text size)
 				and: [(destY + line lineHeight) < characterPoint y]])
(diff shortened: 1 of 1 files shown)


MarcusDenker (approved):

src/Rubric/RubCharacterBlockScanner.class.st
```diff
@@ -100,9 +100,9 @@ RubCharacterBlockScanner >> cr [
 				baselineY := line lineHeight.
 				lastCharacter := nil.
 				characterPoint := (nextLeftMargin ifNil: [leftMargin]) @ destY.
-				(lastIndex < text size and: [(text at: lastIndex) = Character cr and: [(text at: lastIndex+1) = Character lf]])
-					ifTrue: [lastIndex := lastIndex + 2]
-					ifFalse: [lastIndex := lastIndex + 1].
+				lastIndex := (lastIndex < text size and: [(text at: lastIndex) = Character cr and: [(text at: lastIndex+1) = Character lf]])
+					ifTrue: [ lastIndex + 2 ]
+					ifFalse: [ lastIndex + 1 ].
```
nicolas-cellier-aka-nice, line 105: you could as well factor `lastIndex := lastIndex + (... ifTrue: [2] ifFalse: [1]).`

src/Rubric/RubCharacterBlockScanner.class.st
```diff
@@ -91,7 +91,7 @@ RubCharacterBlockScanner >> cr [
 	CharacterBlocks, sometimes including the carriage return (cursor is at 
 	the end) and sometimes not (cursor is in the middle of the text)."
 
-	((characterIndex ~= nil
+	((characterIndex notNil
```
nicolas-cellier-aka-nice, line 94: I don't think that difference is really measurable nowadays (more over, the frequency of cr is probably about 1/20 in typical Smalltalk code), but the purpose of this kind of optimization was to make the full Smalltalk character rendering inner loop feasible on slow machines.

In doubt, one could use `ifNotNil:` to ensure same level of optimization.

Or we could profile performance of rendering as part of non regression tests...

PR #4228, opened 2019-08-07
4227-Use-ifNilifNotNil-in-Dictionaryadd

### LINKED ISSUE #4227: Use ifNil:ifNotNil: in Dictionary>>add:
- we should use ifNil:ifNotNil: in Dictionary #add: method

astares:
Use ifNil:ifNotNil: in Dictionary>>add: 

Fix #4227

SIDE NOTE: This is a method that is often used. From what I measured using #bench there is no noticable speed difference. But it would be good if this could be checked from reviewer side too.


MarcusDenker:
Strange that the tone file for Dictionary ist deemed to be binary... I will checked the file on disk, I checked the file manually

MarcusDenker:
from the compiler side,

 == nil ifTrue:ifFalse: 

compiled to the same code as #4227 

so this change is good.

PR #4231, opened 2019-08-07
4229-Lint-clean-BehaviorallInstVarNames

### LINKED ISSUE #4229: Lint clean Behavior>>allInstVarNames
- move assignment outside
- use ifNil:ifNotNil:

astares:
Lint clean Behavior>>allInstVarNames

Fix #4229

### DIFF
--- a/src/Kernel/Behavior.class.st
+++ b/src/Kernel/Behavior.class.st
@@ -131,9 +131,9 @@ Behavior >> allInstVarNames [
 	accessed by the interpreter."
 
 	| vars |
-	self superclass == nil
-		ifTrue: [vars := self slots collect: #name]	"Guarantee a copy is answered."
-		ifFalse: [vars := self superclass allInstVarNames , (self slots collect: #name)].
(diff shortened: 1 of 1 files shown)


Ducasse:
Would it not be better that way

vars := self superclass 
			ifNil: [ self slots collect: #name ]	"Guarantee a copy is answered."
			ifNotNil: [ self superclass allInstVarNames , (self slots collect: #name) ].

vars := self superclass 
			ifNil: [ self slots collect: [:each | each name ]	"Guarantee a copy is answered."
			ifNotNil: [ :sup | sup allInstVarNames , (self slots collect: [:each | each name ]) ].


In addition, I think that we should stop using xxx: #foo in core system because such way of doing things is really slower.

S.

MarcusDenker:
as this PR just does the ifNil: conversion, I will merge it and out a second pass on allInstVarNames on my TODO. (I personally think it should call #instVarNames, this will make it much simpler).

PR #4244, opened 2019-08-07
4238-Calypso-does-not-show-protocols-of-traits

### LINKED ISSUE #4238: Calypso does not show protocols of traits
I have a class using a trait with methods in a 'testing' protocol.

If I have at least one local method in the 'testing' protocol, then I see the methods of the traits in it. But if I have no local method in it, then the protocol does not appear in the list of protocols.

jecisc:
In calypso, also show protocols when only methods from traits are inside.

Fixes #4238


jecisc:
@dionisiydk This PR might interest you.

dionisiydk (approved):

PR #4245, opened 2019-08-07
Remove-duplicated-method-in-TraitedClass

jecisc:
Remove deplicated method from TraitedClass


jecisc:
The method is the same in ClassDescription

Ducasse:
Tx

PR #4246, opened 2019-08-07
Clarify why we write the ZipArchiveMember header after the data

jgfoster:
The previous implementation was to write a header, write the data, then "refresh" the header. The refresh method was largely a duplicate of the first write method, and included the following:
```
self flag: 'WTF! What is the purpose of this method?! We just wrote the headers. Why write them again?'.
```

To answer that question: the header must appear first in the file, but must include some information that we don't know until after we write the data (including the compressed data size and CRC). So the first write of the header appears to be just a way of reserving space. 

This refactoring removes the duplicate code and makes it explicit that we are just skipping some number of bytes to leave room for the header that is now only written once. A variable is also renamed to be more explicit.


Ducasse:
strange failures in the build. Close and restart to check.

Ducasse:
James can you have a look because I have the impression that your changes are breaking the MCZInstaller.

Error

Attempt to set the position of a PositionableStream out of bounds
Stacktrace

RWBinaryOrTextStream(Object)>>error:
RWBinaryOrTextStream(PositionableStream)>>positionError
RWBinaryOrTextStream(PositionableStream)>>position:
RWBinaryOrTextStream(WriteStream)>>position:
ZipStringMember(ZipArchiveMember)>>writeTo:
[ :member |
member writeTo: stream.
member endRead ] in ZipArchive>>writeTo:
OrderedCollection>>do:
ZipArchive>>writeTo:
MCMczWriter>>flush
MCMczWriter class>>fileOut:on:
MCMczInstallerTest>>testInstallFromStream
MCMczInstallerTest(TestCase)>>performTest

jgfoster:
Yes, that seems related. I'll investigate...

jgfoster:
I think I've addressed the problem I introduced. I'm not sure about the GC errors but they seem (to me) to be unrelated.

MarcusDenker:
tests are now good

PR #4247, opened 2019-08-07
Repackage PositionableStream class >> with:do: 

jgfoster:
This method is referenced from `GZipReadStream class>>unzip:to:` but is provided by a testing package. Should I rewrite the code to not use the missing method? Or move the method to the minimal image? Or something else?

With this PR I propose to move it from `Fuel-Tests-Core` to `Collections-Streams` since it might be used elsewhere. I'm happy to take another approach, but at present the minimal image is broken so something should be done.


astares (approved):
OK

jecisc (approved):

PR #4248, opened 2019-08-08
The WorldState should keep the value of the window size.

tesonep:
This is useful to store the value in the image, so the OSWindow can have it. 
And it does not depend on a primitive to access it.

### DIFF
--- a/src/OSWindow-Core/OSWorldRenderer.class.st
+++ b/src/OSWindow-Core/OSWorldRenderer.class.st
@@ -57,7 +57,7 @@ OSWorldRenderer >> checkForNewScreenSize [
 	display := Form extent: self actualScreenSize depth: 32.
 	self osWindowRenderer form: display.
 
-	world worldState storedExtent: self actualScreenSize.
+	world worldState realWindowExtent: self actualScreenSize.
 
 	world restoreMorphicDisplay.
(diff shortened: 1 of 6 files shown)


src/OSWindow-Core/OSWorldRenderer.class.st
```diff
@@ -126,21 +128,25 @@ OSWorldRenderer >> displayWorldState: aWorldState ofWorld: aWorld submorphs: sub
 { #category : #initialization }
 OSWorldRenderer >> doActivate [
 
-	| attributes |
+	| attributes initialExtent |
+	
+	initialExtent := world worldState realWindowExtent ifNil: [1500@1000].
```
dionisiydk, line 133: Would be nice to define World method #realWindowExtent with default value being there instead of in the renderer.

dionisiydk: And maybe different name for world method like #preferredExtentForStartup to be more explicit that it is not an actual current window extent

dionisiydk (approved):
This is nice.
Do we already have opened worlds/windows in the image which are restored on startup?

Ducasse:
I merge and you can continue discussing.

PR #4254, opened 2019-08-08
Pre-load MCWorkingCopy cache 

jgfoster:
Instead of using recursion (with a risk of stack overflow) this populates the cache in reverse order.


Ducasse:
tx james! I like your comments

MarcusDenker:
failing tests are related to the change I think

jgfoster:
Yes, there was a problem with a variable name. Thanks!

jecisc (approved):
This is fine. I think we can merge it.

PR #4256, opened 2019-08-08
Fix refactoring preview when the change is empty

### LINKED ISSUE #4198: Refactoring preview is broken when the change is empty
For example, when we create accessors. 
If accessors already exist then the preview just raises an error.

Lin777:
Fixes #4198


Ducasse:
The changes look good now I would like to have the integration checking.

Ducasse:
The broken test is unrelated.

PR #4262, opened 2019-08-09
Fixed problems with ZnMimeType hash when constructed using fromString:

### LINKED ISSUE #4255: ZnMimeType hash have problems in Pharo 7.0.3/8.0.0
Code to reproduce the problem:
```smalltalk
| mime1 mime2 |
mime1 := (ZnMimeType fromString: 'application/json;q=1').
mime2 := (ZnMimeType applicationJson parameterAt: 'q' put: '1').
{ mime1 = mime2 . mime1 hash = mime2 hash }

--> #(true false)
```

This problem is fixed (or never introduced I don't know) in the main Zinc Repository. The problem is that `ZnMimeType>>fromString:` uses a plain `Dictionary` for the parameters instead of `SmallDictionary` like the parameters method that lazy initializes the variable.

In Sven's Zinc repository `SmallDictionary` is always used and the hash computation is consistent. https://github.com/svenvc/zinc/blob/master/repository/Zinc-Resource-Meta-Core.package/ZnMimeType.class/class/fromString..st

**System info:**
```
Pharo 7.0.3
Build information: Pharo-7.0.3+build.158.sha.0903ade8a6c96633f07e0a7f1baa9a5d48cfdf55 (64 Bit)
```

I don't know how external projects included in Pharo are supposed to be updated. We must modify code in Pharo directly? Or there's some way to bring the updates sven makes in his repo?

gcotelli:
Fixes #4255 in Pharo 8
- Added a test case for the problem
- Changed `ZnMimeType>>#fromString:` to use the same kind of dictionary used when lazy initialized
- Additionally fixed some uses of `assert:` and `=` to use `assert:equals:`

### DIFF
--- a/src/Zinc-Resource-Meta-Core/ZnMimeType.class.st
+++ b/src/Zinc-Resource-Meta-Core/ZnMimeType.class.st
@@ -135,7 +135,7 @@ ZnMimeType class >> fromString: aString [
 	sub := aString copyFrom: main size + 2 to: endOfSub.
 	endOfSub = aString size ifTrue: [ ^ self main: main sub: sub ].
 	parts := (aString copyFrom: endOfSub + 1 to: aString size) substrings: ';'.
-	parameters := Dictionary new.
+	parameters := SmallDictionary new.
 	parts do: [ :each | 
 		parameters
 			at: (each copyUpTo: $=) trimBoth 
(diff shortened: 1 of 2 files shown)


Ducasse:
Changes look good!

PR #4264, opened 2019-08-09
[Backport] P7 - Fixed problems with ZnMimeType hash when constructed using fromString: 

### LINKED ISSUE #4255: ZnMimeType hash have problems in Pharo 7.0.3/8.0.0

gcotelli:
Fixes #4255 in Pharo 7
- Added a test case for the problem
- Changed `ZnMimeType>>#fromString:` to use the same kind of dictionary used when lazy initialized

### DIFF
--- a/src/Zinc-Resource-Meta-Core/ZnMimeType.class.st
+++ b/src/Zinc-Resource-Meta-Core/ZnMimeType.class.st
@@ -135,7 +135,7 @@ ZnMimeType class >> fromString: aString [
 	sub := aString copyFrom: main size + 2 to: endOfSub.
 	endOfSub = aString size ifTrue: [ ^ self main: main sub: sub ].
 	parts := (aString copyFrom: endOfSub + 1 to: aString size) substrings: ';'.
-	parameters := Dictionary new.
+	parameters := SmallDictionary new.
 	parts do: [ :each | 
 		parameters
 			at: (each copyUpTo: $=) trimBoth 
--- a/src/Zinc-Resource-Meta-Tests/ZnMimeTypeTests.class.st
+++ b/src/Zinc-Resource-Meta-Tests/ZnMimeTypeTests.class.st
@@ -38,6 +38,20 @@ ZnMimeTypeTests >> testCharset [
 	self assert: mimeType charSet = 'utf-8'
 ]
 
+{ #category : #tests }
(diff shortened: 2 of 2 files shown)


gcotelli:
Failing tests seems to be unrelated.

gcotelli:
Just for reference the same changes were already merged in Pharo 8 : #4262

gcotelli:
This is still valid, waiting for merge.

gcotelli:
This is still valid, waiting for merge.

CafeKrem (approved):
It seems good

svenvc:
It is already part of Zinc itself.

CafeKrem:
Ok sorry I don't know but why this PR exist ?

CafeKrem:
may be we can close it ?

Ducasse:
Sven just that I'm sure. 
should we integrate a more version of Zinc in P7 having this fix.

svenvc:
I don't think so.

Most people who deeply care about HTTP load the latest version of Zn anyway.

This is a really small change.

Ducasse:
Ok so I will close it because we need to focus on P8.

PR #4277, opened 2019-08-12
4275-allInstVarNames-should-call-instVarNames

MarcusDenker:
simplify #allInstaVarnames  by re-using #instVarNames


jecisc (approved):

Ducasse:
Why the code is not like the following:

self superclass
			ifNil: [ self instVarNames ]
			ifNotNil: [: sup | sup allInstVarNames , self instVarNames ]

?

MarcusDenker:
ah, yes, that is even better

PR #4278, opened 2019-08-12
Revert-Spec1-changes

jecisc:
Revert changes to done to Spec 1 during Pharo 8.


MarcusDenker:
bootstrap seems to fail

jecisc:
I'm on it ;) 

I expect other errors but since it's messy to launch all Pharo tests locally I'll wait for the CI.

jecisc:
I'm working on the new error

Ducasse:
Hi cyril now the revert of iceberg is integrated so it may help you.

Ducasse:
cyril we have 75 failing tests so may be this is better to fix some of them before integrating else we will not see if we break more. What do you think?

jecisc:
I agree.

jecisc:
I reverted some changes to NautilusRefactoring and DependencyAnalyzer that are not yet migrated but still used Spec 2 methods.

jecisc:
I'll take care of the last failures tomorrow

PR #4288, opened 2019-08-12
Quick-migration-of-Epicea-to-Spec2

jecisc:
Do a quick migration of Epicea to Spec 2. 

The code is not as good as it could be but it is not worst than before and will allow us to revert Spec "1.5" to Spec 1.

### DIFF
--- a/src/EpiceaBrowsers/EpDashboardPresenter.class.st
+++ b/src/EpiceaBrowsers/EpDashboardPresenter.class.st
@@ -6,7 +6,7 @@ self open
 "
 Class {
 	#name : #EpDashboardPresenter,
-	#superclass : #ComposablePresenter,
+	#superclass : #SpPresenter,
 	#instVars : [
 		'monitor',
 		'createLogButtonPresenter',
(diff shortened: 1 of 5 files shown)


tinchodias:
Ehhh! "not worse than before", hah. It will be good to see it in Spec2. Good.

jecisc:
@tinchodias I mean that moving Epicea from Spec "1.5" to 2 like this does not make it use the new Spec 2 ways. Not that the Spec 1 version was bad ;)

src/EpiceaBrowsers/EpDashboardPresenter.class.st
```diff
@@ -17,6 +17,20 @@ Class {
 	#category : #'EpiceaBrowsers-UI-Dashboard'
 }
 
+{ #category : #specs }
+EpDashboardPresenter class >> defaultSpec [
+	^ SpBoxLayout newVertical
+		add: #currentSessionInfoPresenter;
+		add:
```
estebanlm, line 24: Hi, 

Sorry for arriving late (I'm having a huge queue of issues to review and slowly walking them). 
The usage of `#add:withConstraints:` to force a height is bad. 
Instead,  we use `#add:expand:fill:padding:`. In this example the code will be: 

```Smalltalk
EpDashboardPresenter class >> defaultSpec [
 	^ SpBoxLayout newVertical
 		add: #currentSessionInfoPresenter;
 		add: (SpBoxLayout newHorizontal
 				add: #enableMonitorButtonPresenter;
 				add: #createLogButtonPresenter;
 				add: #openSettingsButtonPresenter;
 				yourself)
			expand: false
			fill: false
			padding: 0;
		yourself
```
Please, can you fix this?

jecisc: Done in https://github.com/pharo-project/pharo/pull/4474

PR #4295, opened 2019-08-12
 Delete warnings of references between classes that will be remove

### LINKED ISSUE #4294: Removing classes should not warn about references in classes to remove
If I want to remove a class A and a class B with B referencing A, currently, Calypso will warn me that A is still referenced before applying the refactoring.

IMO this should not happen because the class referencing A is part of the classes to remove. 

https://github.com/Lin777/Enjoliveur/issues/41

Lin777:
Fixes #4294


src/Calypso-Browser/Collection.extension.st
```diff
@@ -0,0 +1,14 @@
+Extension { #name : #Collection }
+
+{ #category : #'*Calypso-Browser' }
+Collection >> nonIntersect: aCollection [
```
jecisc: I think this is equivalent to #difference:

jecisc (changes requested):
I think your #nonIntersect: is equivalent to difference: 
If you can confirm it, then replace it please :)

src/Calypso-Browser/Collection.extension.st
Lin777: Yes, they have the same function. I will update the PR to use the #difference: function

src/Calypso-SystemTools-Core/ClyBrowserMorph.extension.st
```diff
@@ -26,20 +26,38 @@ ClyBrowserMorph >> confirmEmptySystemQuery: aQuery [
...
+ClyBrowserMorph >> confirmEmptySystemQuery: aQuery excluding: classes [
+	
+	| showResult result |
+	
+	result := aQuery execute.
+	(result isEmpty or: [ ((result items collect: [:each | each origin]) difference: classes ) isEmpty]) ifTrue: [ ^true ].
+	
+	showResult := UIManager default confirm: 'There are '
+		, result size asString, ' ', aQuery description, '. Show them?'.
+	showResult ifTrue: [self spawnQueryBrowserOn: aQuery].
```
dionisiydk, line 39: I think we need to show only external references instead of all of them. And for this we can introduce ClyExternalClassReferencesQuery. In that case this method will not be needed.

Otherwise the method needs to be renamed at least because system query (what is in method name) is not only about methods. Queries can return other objects (classes, comments)

src/Calypso-SystemTools-Core/ClyBrowserMorph.extension.st
```diff
@@ -26,20 +26,38 @@ ClyBrowserMorph >> confirmEmptySystemQuery: aQuery [
...
+	^false
+]
+
 { #category : #'*Calypso-SystemTools-Core' }
 ClyBrowserMorph >> confirmUnusedClasses: classes [
 	
 	| refQuery noUsers answer subclasses users |
+	
 	refQuery := ClyClassReferencesQuery toAny: classes from: self systemScope.
```
dionisiydk, line 48: So here it can be simple 
```Smalltalk
refQuery := ClyExternalClassReferencesQuery...
noUsers := self confirmEmptySystemQuery: refQuery
```

dionisiydk (changes requested):

src/Calypso-SystemTools-Core/ClyBrowserMorph.extension.st
```diff
@@ -26,20 +26,38 @@ ClyBrowserMorph >> confirmEmptySystemQuery: aQuery [
...
+	
 	refQuery := ClyClassReferencesQuery toAny: classes from: self systemScope.	
-	noUsers := self confirmEmptySystemQuery: refQuery.
+	noUsers := self confirmEmptySystemQuery: refQuery excluding: classes.
 	
-	(classes anySatisfy: [ :each | each hasSubclasses ]) ifTrue: [
+	(classes anySatisfy: [ :each | |subclassesOfClass|
+		subclassesOfClass := each subclasses.
+		subclassesOfClass notEmpty and: [ (subclassesOfClass difference: classes) notEmpty ] ]) ifTrue: [
 		noUsers := false.
```
dionisiydk, line 53: Maybe better would be to collect all "external" subclasses and show only them. It can be like:
```
subclasses := (classes flatCollect: [:each | each subclasses]) copyWithoutAll: classes.
subclasses ifNotEmpty: [ "show them" ]
```
And same for trait users

Lin777: Ok, I'll update it

Ducasse:
Thanks you all for the reviews and enhancements. this is great.

PR #4297, opened 2019-08-13
Deprecate #llCallsOnIn:

MarcusDenker:
allCallsOnIn: is not really useful: it is just called from SystemNavivation with itself as a parameter. Use #allCallsOn instead

### DIFF
--- a/src/System-Support/Behavior.extension.st
+++ b/src/System-Support/Behavior.extension.st
@@ -18,10 +18,8 @@ Behavior >> allCallsOn: aSymbol [
 { #category : #'*System-Support' }
 Behavior >> allCallsOnIn: aSystemNavigation [
 	"Answer a SortedCollection of all the methods that refer to me by name or as part of an association in a global dict."
-
-	^ (aSystemNavigation 
-		allReferencesTo: (self environment associationAt: self instanceSide name)), 
-		(aSystemNavigation allCallsOn: self instanceSide name)	
+	self deprecated: 'use allCallsOn'.
+	^ self allCallsOn
 ]
 
 { #category : #'*System-Support' }
--- a/src/Tool-Base/SystemNavigation.extension.st
+++ b/src/Tool-Base/SystemNavigation.extension.st
@@ -25,7 +25,7 @@ SystemNavigation >> browseAllCallsOnClass: aClass [
 	aClass. For example, SystemNavigation new browseAllCallsOnClass: Object."
 
 	^ self
-		browseMessageList: (aClass allCallsOnIn: self)
+		browseMessageList: aClass allCallsOn
 		name: 'Users of class ' , aClass instanceSide name
 		autoSelect: aClass instanceSide name
 ]


Ducasse:
The problem is that we cannot scope anymore calls. I do not see the added value of this change.

MarcusDenker:
the problem is that the scoping never really worked (and is not used at all anymore)

guillep:
@MarcusDenker could you think on a better and more general aproach to enhance system queries? These enhancements are nice, but I'd like to see how they can fit in a more general enhancement vision.

dionisiydk:
Hi guys.

Calypso queries support scoping and composition. And they can replace all methods of systemNavigator.

Notice those queries do not depend on browser part (separate packages and baseline groups). They are installed on minimal servers in case of TelePharo.

I remember I shown DSL idea some of you to have smart language over all of it.
Something like:
```Smalltalk
Smalltalk query
    sendersOf: #do:;
    list.

Smalltalk query
    sendersOf: #do:;
    openInBrowser.

Smalltalk query
    sendersOf: #do:;
    from: (ClyClassScope of: Point);
    list.

Smalltalk query
    sendersOf: #do:;
    fromClasses: {Point};
    list

Point query
    sendersOf: #do:;
    list.

Point query
    sendersOf: #do:;
    referencesTo: Point; "merge two queries"
    list
```

If you like this idea I can prepare PR with proposal.

Ducasse:
@marcus. When designing a library this is normal that there is not sender. To me this is not a good criteria. 
I think that it would be good to focus on changes that are more impacting (like fixing Reflectivity because we spent some days with steven to fix the context stackj mapping while we do not understand anything in IR). If we only focus on micro enhancements pharo will never change. And for this particular change I do not see that it is worth the energy that I spend on writing this mail. Reducing possibilities looks strange to me.

@Denis What I would like to see also is how this is polymorphic to with RBEnvironment and Query. 
We want to stop to have three different ways to do the same. 
So personally I do not care of the RBEnvironment but it works for refactorings. 
Now I do not want to have to maintain broekn system (refactorings do not work for Slot and traits for example but they will not magically work). 
Now denis to reply to you question frankly I do not know because your design often uses too many classes and not enough objects. You can compare the design of Commander2 and Commander to see what I mean. And I really like Commander2 because it needs far less classes.
So here is what I suggest: Give a try and we should assess it its design before committing to it.

dionisiydk:
Hi @Stef

чт, 15 авг. 2019 г., 8:59 StéphaneDucasse <notifications@github.com>:

> @marcus <https://github.com/marcus>. When designing a library this is
> normal that there is not sender.
>
It is not a normal at least by one important criteria:  no sender means no
test describing it. I am sure if there will be a test @marcus will not
touch it.

stef:
hey @dionisiydk you should tag the right person, not random persons totally unrelated to this project. or just leave of the @ if you don't want to bother other people but still want to stick to the nick... thx.

dionisiydk:
> hey @dionisiydk you should tag the right person, not random persons totally unrelated to this project. or just leave of the @ if you don't want to bother other people but still want to stick to the nick... thx.

@stef sorry about that. Phone integration seems does not work correctly

dionisiydk:
чт, 15 авг. 2019 г., 8:59 StéphaneDucasse

> @denis <https://github.com/denis> What I would like to see also is how
> this is polymorphic to with RBEnvironment and Query.
>
They are not polymorphic. I think RBEnvironment is more like scopes in
Calypso. There are no reified queries in RB, queries are just set of
methods in environment classes. So if you want composition of queries you
need to create a method which will call others. In calypso you can compose
existing queries with builtin composition mechanizm (simply q1 , q2).
In RB you can compose environments similar to the Calypso scopes
composition.

If you think about using RBEnvironment instead of scopes for calypso
queries then it will require some refactoring and integration of features
which scopes provide. I don't know how difficult it would be.
But technically Calypso scopes are much lighter objects in terms of what
information you can retrieve from it. There are only enumeration methods
like methodsDo:, classesDo depending on the type of scope. Anything complex
is supposed to be reified as queries by design.

> We want to stop to have three different ways to do the same.
> So personally I do not care of the RBEnvironment but it works for
> refactorings.
>
I think moving refactorings to calypso queries will not be a much of work.
And it will bring (with some extra effort) remote refactorings to
TelePharo. But it's better to do it on top of nice query API.
So my question was more about what you think about proposed quering API?

> Now I do not want to have to maintain broekn system (refactorings do not
> work for Slot and traits for example but they will not magically work).
>
Now denis to reply to you question frankly I do not know because your
> design often uses too many classes and not enough objects.
>
Often? Maybe you should look at other projects which I did in Lille. For
the browser and commands I prefer to extend them in declarative way. And
classes here are used similar to exceptions and announcements.

> You can compare the design of Commander2 and Commander to see what I mean.
> And I really like Commander2 because it needs far less classes.
>
 I did and I prefer my design. But it is different story

Ducasse:
> They are not polymorphic. I think RBEnvironment is more like scopes in Calypso. There are no reified queries in RB, queries are just set of methods in environment classes. So if you want composition of queries you need to create a method which will call others. In calypso you can compose existing queries with builtin composition mechanizm (simply q1 , q2). In RB you can compose environments similar to the Calypso scopes composition. If you think about using RBEnvironment instead of scopes for calypso queries then it will require some refactoring and integration of features which scopes provide. I don't know how difficult it would be. But technically Calypso scopes are much lighter objects in terms of what information you can retrieve from it. 

I do not know. I would like to have some analysis and better reuse of existing things. 
You tend to never reuse and create new things and after we have three different ways to do the same. I do not care about RBenvironment. I'm looking for reduce the load of Pharo. 
To me a query can return an RBenvironment or 
We will have to rewrite Calypso in Spec (and people here are not happy that they have to do it),
else we will have to throw away calypso when we pass on Brick :( 

There are only enumeration methods like methodsDo:, classesDo depending on the type of scope. Anything complex is supposed to be reified as queries by design.
> We want to stop to have three different ways to do the same. So personally I do not care of the RBEnvironment but it works for refactorings.
> I think moving refactorings to calypso queries will not be a much of work. And it will bring (with some extra effort) remote refactorings to TelePharo. But it's better to do it on top of nice query API. So my question was more about what you think about proposed quering API?

Ok now if the calypso queries are full of class static methods and class annotations and are in the same vein than the Commander design then this is a no go. This is not by accident that we worked on Commander2. Everybody in the team was unconfortable with the design of Commander. It took me some months to really understand what I do not like and now I know. 

About the query I do not know. It looks cool but again I need more time to make a real assessment. 

> Now I do not want to have to maintain broekn system (refactorings do not work for Slot and traits for example but they will not magically work).
> Now denis to reply to you question frankly I do not know because your
> design often uses too many classes and not enough objects.
> Often? Maybe you should look at other projects which I did in Lille. For the browser and commands I prefer to extend them in declarative way. And classes here are used similar to exceptions and announcements.
> You can compare the design of Commander2 and Commander to see what I mean. And I really like Commander2 because it needs far less classes.
> I did and I prefer my design. But it is different story
> […](#)

Ducasse:
So please package your query idea so that we can have a look. I cannot just guess.

bencoman:
> It took me some months to really understand what I do not like and now I know.
> class static methods and class annotations

@Ducasse Its not clear to me what these are.  Perhaps myself and others could make the same mistake.   It might be good to expand on "why" some "Pharo coding style" document.

PR #4310, opened 2019-08-13
Use stack instead of recursion in MCScanner>>next

jgfoster:
This avoids a stack overflow when loading Metacello into GemStone.
I've also added a number of comments (which accounts for some of the extra length).

### DIFF
--- a/src/Monticello/MCScanner.class.st
+++ b/src/Monticello/MCScanner.class.st
@@ -45,7 +45,7 @@ MCScanner >> next [
 		"For an array, start a new level on the stack"
 		c = $( ifTrue: [ stream next. stack push: Stack new ] ifFalse: [
 		"At the end of an array, so add it to the previous level"
-		c = $) ifTrue: [ | x | x := stack pop. stack top push: x asArray reverse ] ifFalse: [ 
+		c = $) ifTrue: [ | x | stream next. x := stack pop. stack top push: x asArray reverse ] ifFalse: [ 
 		"Unexpected token"
 		self error: 'Unknown token type' ]]]].
 		"Keep looping while we are in an array"
--- a/src/Monticello/MCScanner.class.st
+++ b/src/Monticello/MCScanner.class.st
@@ -28,8 +28,7 @@ MCScanner class >> scanTokens: aString [
 { #category : #actions }
 MCScanner >> next [
 	"This would be slightly simpler using recursion but it risks a stack overflow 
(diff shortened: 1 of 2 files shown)


src/System-Hashing/MD5NonPrimitive.class.st
```diff
@@ -25,7 +25,7 @@ Class {
 
 { #category : #'class initialization' }
 MD5NonPrimitive class >> initialize [
-	"MD5 initialize"
+	"MD5NonPrimitive initialize"
 	"Obscure fact: those magic hex numbers that are hard to type in correctly are
```
jgfoster, line 29: This isn't related to the other edit, but was just a comment I stumbled across that seems wrong.

src/Monticello/MCScanner.class.st
```diff
@@ -27,14 +27,32 @@ MCScanner class >> scanTokens: aString [
...
-	c = $' ifTrue: [^ self nextString].
-	c = $( ifTrue: [^ self nextArray].
-	c isAlphaNumeric ifTrue: [^ self nextSymbol].
-	self error: 'Unknown token type'.	
+	"This would be slightly simpler using recursion but it risks a stack overflow 
+	 for some packages on some platforms, so we implement it using a local stack. 
+	 The performance is essentially equivalent (<5% difference)."
+
+	| stack |
+	stack := OrderedCollection with: OrderedCollection new.
```
gcotelli: Why not using `Stack` as the stack?

jgfoster: Because I'm used to GemStone that doesn't have it. I've submitted an update. Thanks!

Ducasse:
James the following tests are failing and they look related to your changes.
unix-32 / Tests-unix-32 / test4 – Unix32.Monticello.Tests.Base.MCScannerTest
<1s
unix-32 / Tests-unix-32 / test5 – Unix32.Monticello.Tests.Base.MCScannerTest

jgfoster:
Those tests certainly would seem related to my changes, but I'm not able to see the failures. It seems to me that as of commit 4d27c0e the Unix32 tests passed.

Screen Shot 2019-08-18 at 10 44 26 AM

jgfoster:
I'm able to see the failures now. I'll investigate.

jgfoster:
MCScannerTest now passes.

MarcusDenker (approved):

PR #4312, opened 2019-08-13
2540-self-theme-settings-fadedBackgroundWindows-is-overzealous-violation-of-LOD

### LINKED ISSUE #2540: self theme settings fadedBackgroundWindows is overzealous violation of LOD
It is a bit insane to have such kind of logic. 
theme should expose self theme settings fadedBackgroundWindows and not showing the fact that this is a setting!!!

jecisc:
Show some respect to Demeter by implementing #fadedBackgroundWindows on theme.

Fixes #2540

### DIFF
--- a/src/Morphic-Base/HaloMorph.class.st
+++ b/src/Morphic-Base/HaloMorph.class.st
@@ -963,13 +963,10 @@ HaloMorph >> handlerForBlueButtonDown: anEvent [
 
 { #category : #initialization }
 HaloMorph >> initialize [
-	"initialize the state of the receiver"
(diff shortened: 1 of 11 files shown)


Ducasse:
Well there is a fundamental BAD IDEA AND FUCKING BAD DESIGN in this Singleton crap Smalltalk ui theme, StandardFonts, Smalltalk tools, ... as a community we were lame not to produce objects encapsulating toolSuite, iconManager access from root widget, theme access in root object (may be ThemedMorph).
So thank cyril.

PR #4325, opened 2019-08-14
4324-activateCursor-method-is-missing-on-WorldMorph

### LINKED ISSUE #4324: activateCursor: method is missing on WorldMorph
The #activateCursor: method was removed (during refactoring of cursor ownership) 
but is sent within #activateInCursorOwner: which is looking like that in P8 now:
```Smalltalk
   activateInCursorOwner: aCursorOwner
	^ aCursorOwner activateCursor: self
```
There is no #activateCursor: implementation in the image.

When one sets a new Cursor using Cursor (class)>>#currentCursor: then the cursor is 
not set - instead the system crashes into the emergency debugger.

image

One can reproduce this by loading the "VistaCursor" project for Windows which worked in P7 
but not in P8. Note that it has to be a Windows machine as the cursor switch does not happen on Mac and Linux as "VistaCursor" only works on Windows.

astares:
activateCursor: method is missing on WorldMorph

Fix #4324

### DIFF
--- a/src/Graphics-Display Objects/Cursor.class.st
+++ b/src/Graphics-Display Objects/Cursor.class.st
@@ -1035,7 +1035,7 @@ Cursor class >> write [
 
 { #category : #primitives }
 Cursor >> activateInCursorOwner: aCursorOwner [
-	^ aCursorOwner activateCursor: self
+	^ self beCursor
 ]
 
 { #category : #converting }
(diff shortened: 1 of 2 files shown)


astares:
@MarcusDenker 

To verify the PR:

- load "VistaCursor" goodie which is for Windows in a regular Pharo 8 image -> emergency
debugger
- then try the same with the code change of the PR

MarcusDenker (approved):

PR #4330, opened 2019-08-16
code-completion-refactoring-changes

myroslavarm:
renaming classes and fixing temp var name in initEntries

### DIFF
--- a/src/NECompletion/ReverseAlphabeticSorter.class.st
+++ b/src/NECompletion/ReverseAlphabeticSorter.class.st
@@ -2,19 +2,19 @@
 Reverse alphabetic sorting
 "
 Class {
-	#name : #TestReverseAlphabeticSorter,
-	#superclass : #TestSorter,
+	#name : #ReverseAlphabeticSorter,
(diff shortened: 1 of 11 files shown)


MarcusDenker (approved):
looks good

src/NECompletion/ReverseAlphabeticSorter.class.st
```diff
@@ -2,19 +2,19 @@
 Reverse alphabetic sorting
 "
 Class {
-	#name : #TestReverseAlphabeticSorter,
-	#superclass : #TestSorter,
+	#name : #ReverseAlphabeticSorter,
+	#superclass : #CompletionSorter,
 	#category : #'NECompletion-New'
 }
```
dionisiydk, line 8: Not related to this particular changes but I am curious did you see SortFunction's in Pharo?
It looks like it can be used here instead of new sorters hierarchy. Check class comment and subclasses.

MarcusDenker: Ah, interesting... I put it on the TODO to check that

PR #4332, opened 2019-08-16
code-completion-context-tests

myroslavarm:
adding tests that test both old and new completion.
for new completion 10 out of 21 tests are green but the rest fail so they are skipped until we fix or delete them on second iteration


MarcusDenker:
SystemDependenciesTest>>testExternalBasicToolsDependencies is failing with this... 

Given Collections do not match! additions : #(#'NECompletion-Tests') missing: #()
Stacktrace

MarcusDenker (approved):
lets add it and then do another iteration

PR #4338, opened 2019-08-17
Ensure-all-class-creation-methods-have-package-equivalent-instead-of-category

jecisc:
- Ensure all class creation methods have a version using #package: instead of #category:
- Ensure all class creation methods call the right method
- Put in deprecated protocol all versions with #category: (Not yet full deprecated since they are used a lot)

### DIFF
--- a/src/Kernel/Class.class.st
+++ b/src/Kernel/Class.class.st
@@ -507,32 +507,37 @@ Class >> environment: anEnvironment [
 ]
 
 { #category : #'subclass creation - weak' }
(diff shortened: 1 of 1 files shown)


jecisc:
Next step to remove #category: is to make sure MCClassDefinition>>#printDefinitionOn: is updated.

jecisc:
Tests passed!

Ducasse:
This is cool. I tried to fix the problem in RB tests but so far I failed.

PR #4339, opened 2019-08-17
Exctract-Colors-from-Bootstrap

jecisc:
Extract Colors from PharoBootstrap to Display.

There is one last dependency that needs to be cleaned, it is for the command line handler.


Ducasse:
Thanks cyril this is strange since the build breaks.

jecisc:
I think it might be because the command line is using colors :( I'll check later. If it's the case I'll do a first PR with the cleaning of dependencies but without the removal of Color from the bootstrap and work on it later.

Ducasse:
Oki.

PR #4340, opened 2019-08-17
Remove-dependency-between-FileSystem-Core-and-Deprecated-file-package

jecisc:
FileSystem-Core should not depend on DepercatedFileStream


Ducasse:
Cyril this is strange I checked and there are no senders of onHandle: do you think that it makes sense to keep it.

Ducasse:
Alistair what do you think?

jecisc:
I don’t know but since we have a deprecated package and it’s only one method I guess we can just let it there until we remove it in Pharo 9 or 10.

PR #4341, opened 2019-08-17
Rubric-should-not-depend-on-deprecated-streams

jecisc:
Update rubric to not depend on deprecated file streams.


jecisc:
Failure not related.

PR #4343, opened 2019-08-17
Fonts-should-not-depend-on-deprecated-file-streams

jecisc:
Fonts should not depend on deprecated file streams.


jecisc:
I am not really sure this is the right replacement. I would appreciate the review of someone knowing the new file implementation.

jecisc:
Error seems unrelated.

Ducasse:
It looks ok for me. I will ask guillep to have a look.

PR #4344, opened 2019-08-17
Graphics-primitives-should-not-depend-on-deprecated-streams

jecisc:
Graphics-Primitives should not depend on DeprecatedFileStream


Ducasse:
Cyril I do not really get this one because bitmap is invoking newFromStream: (may be not the one of StandardFileStream). So this is not clear that we are not breaking something.

PR #4345, opened 2019-08-17
Remove-PNGReader-test-missing-resources

jecisc:
Remove test that needs an unavailable resource.

There is a link but the resource seems unavailable now. So this is just dead code now.


Ducasse:
May be we could archive the following suite http://www.schaik.com/pngsuite/ and check how good our pngreader is. 


<blockquote><div><strong><a href="http://www.schaik.com/pngsuite/">PngSuite - the official set of PNG test images</a></strong></div><div>The PngSuite consists of a two hundred 
    PNG images in all possible formats to support the development of PNG 
    viewing and conversion software</div></blockquote>

Ducasse:
I created a separated issue.

PR #4348, opened 2019-08-17
StartupPreferences-should-not-depend-on-deprecated-streams

jecisc:
Startup preferences should not depend on deprecated streams.


svenvc:
Yes, and better/more OO

PR #4359, opened 2019-08-19
4306-Add-comment-why-name-cache-in-SystemDictionary-is-sorted

MarcusDenker:
add comment to explain why we sort the name cache


MarcusDenker:
Just a comment and tests are green

PR #4360, opened 2019-08-19
Fixing change detection of traits

tesonep:
Improving the detection of changes when compiling a Trait. It was triggering a change in the hierarchy and there was no change in the Metaclass only it was using the default one instead of the MetaclassForTraits


tesonep:
Also fixes #4327

MarcusDenker:
all tests are green

MarcusDenker (approved):

PR #4361, opened 2019-08-19
Cleanup some classes of Refactoring packages

Lin777:
Change of:
```
isEmpty not => isNotEmpty
isEmpty ifTrue: [] => ifEmpty: []
(detect: [] ifNone: [nil] ) notNil => anySatisfy: []
```

### DIFF
--- a/src/Refactoring-Changes/RBRefactoryChangeManager.class.st
+++ b/src/Refactoring-Changes/RBRefactoryChangeManager.class.st
@@ -122,12 +122,12 @@ RBRefactoryChangeManager >> disconnectFromChanges [
 
 { #category : #testing }
 RBRefactoryChangeManager >> hasRedoableOperations [
-	^ redo isEmpty not
+	^ redo isNotEmpty
 ]
 
 { #category : #testing }
 RBRefactoryChangeManager >> hasUndoableOperations [
-	^ undo isEmpty not
+	^ undo isNotEmpty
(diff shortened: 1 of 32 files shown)


MarcusDenker (approved):

Ducasse:
I have an important point about the change ifNotEmpty: 
ifNotEmpty: is costly, it uses cull:. 
So this means that this is not fully equivalent. 
I would be cautious not to apply this blindly. 

ifNotNil: ifNotNilBlock
	"Evaluate the block, unless I'm == nil (q.v.). If the receiver is not nil, pass it as argument to the block."

	"(2 ifNotNil: [ :o | o + 3]) >>> 5"
	"(nil ifNotNil: [:o | o +3 ]) >>> nil"

	^ ifNotNilBlock cull: self

Ducasse:
For example 	

self superclass notNil ifTrue: [self superclass removeSubclass: self].
self superclass ifNotNil: [self superclass removeSubclass: self].

is not good because it should at least be
self superclass ifNotNil: [:sup| sup removeSubclass: self].

Lin777:
Oh, I didn't know that difference; I thought they behaved the same way. I will be attentive if this change triggers an error. 

Thanks so much for the explanation :D

Ducasse:
Once I really understand the implementation and optimisation I will write an email to the mailing-list.

PR #4366, opened 2019-08-20
4365-Reintroduce-separator

### LINKED ISSUE #4365: Reintroduce separator
See discussion here 

http://lists.pharo.org/pipermail/pharo-dev_lists.pharo.org/2019-July/275670.html

- remove the unnecessary separator after "Quit"
- reintroduce separator after "Save and Quit"

Using the additonal divider between the three "safe my data" options and the "just leave" is intentional and makes sense: from the point of view of grouping, visually and from clicking behavior.

Menu will look like this afterwards:

 !--------------------
  ! SETTINGS
  !--------------------
  ! SAVE
  ! SAVE AS
  ! SAVE AND QUIT
  !--------------------
  ! QUIT
  !--------------------

astares:
Reintroduce separator

Fix #4365


gcotelli (approved):
LGTM

Issue #4149, opened 2019-07-29
instVarNames is not including instance variables coming from Stateful traits


gcotelli:
For example, `TextPresenter` is using `TSpecHaveWrappingScrollBars` that includes a `wrapScrollBars` slot, but 
`TextPresenter superclass instSize` >>	40
`TextPresenter instSize` >> 42 "Seems to be considered here, but:"
`TextPresenter instVarNames` -> #(#scrollValue)

This is breaking Fuel serialization because `ClassDescritpion>>instanceVariableNamesDo:` uses `instSize` to iterate over the variables returned by `instVarName`, so you get a `SubscriptOutOfBounds` error.

jecisc:
Seems linked to https://github.com/pharo-project/pharo/issues/3546 and https://github.com/pharo-project/pharo/issues/3520

MarcusDenker:
I adde a PR that we might consider: https://github.com/pharo-project/pharo/pull/4367

The idea is that #instVarnames should return #slots, not #localSlots.

MarcusDenker:
This is now fixed

Issue #3546, opened 2019-06-16
Behavior>>instSize includes slots from Traits, but ClassDescription>>instVarNames doesn't


akgrant43:
`ClassDescription>>instanceVariableNamesDo:` will generate a SubscriptOutOfBounds error when the receiver includes a Trait with slots, e.g.:

```
Trait named: #TestTrait2
	uses: { }
	slots: { #slot1 }
	category: 'AKG-instSize'.

Object subclass: #HasTrait1
	uses: TestTrait2
	instanceVariableNames: 'iVar'
	classVariableNames: ''
	package: 'AKG-instSize'.

{ HasTrait1 instSize. HasTrait1 instVarNames. }.
" #(2 #(#iVar))"

HasTrait1 instanceVariableNamesDo: [ :x | ].
"==> SubscriptOutOfBounds: 2"
```

As can be seen from the above, `#instSize` includes the slot (`#slot1`), while `#instVarNames` does not.

This was triggered by trying to Fuel out a class with this type of definition, so I assume that `#instVarNames` should be modified to include the defined slots.

MarcusDenker:
We need to be careful: Slots is  can be "virtual", that is: there can be a slot, but it does not increase the size of the object.

MarcusDenker:
I think the solution is to provide an explicit "indexedInstVarsDo:" method (and #indexedInstVarNames) for low level users that just care about indexed ivars (and includes hidden ones).

Then indeed have #instVarNames return all visible slot names. 

In general, I think this is were we start to see a case for some "view" concept: Do I want to look at the instance at the low level (memory layout) or see the high level view?

MarcusDenker:
this seems to be related: instVarNames explicitly just answers the local ones:

instVarNames
	"Answer an Array of the receiver's instance variable names."

	^self localSlots collect: [ :each | each name ]

MarcusDenker:
PR merged

Issue #3520, opened 2019-06-14
Cannot see variables of stateful traits in Calypso


jecisc:
In Moose we use a lot of stateful traits but it's hard to work with them because they do not appear in the variables pane of Calypso.

MarcusDenker:
seems to be the same as #3546

MarcusDenker:
this has been merged

PR #4367, opened 2019-08-20
4149-instVarNames-is-not-including-instance-variables-coming-from-Stateful-traits

MarcusDenker:
This is an idea to #4149, #3546 ,  #3520

the idea is that #instVarNames should return not only the local slots.


MarcusDenker:
The testDefinitionOfTraitUsingTrait was testing exactly the old behaviour and is failing now

MarcusDenker:
ah, no, that was the wrong test to fix: it is testing that the MC class definition does not show the non-local slots (which it should indeed not).

I will do another pass

MarcusDenker:
tests are green!

MarcusDenker:
reviewed with max (as it is related to Fuel, too).

PR #4371, opened 2019-08-20
4370-Trait-composition-slots-method-contains-duplication

### LINKED ISSUE #4370: Trait composition `slots` method contains duplication
`ArrayTest class traitComposition slots`

If you execute this snippet you'll get a collection with a lot of duplication. This is because it will not remove duplication between the slots present in multiple traits of the trait composition.

jecisc:
TaSequence should not duplicate slots.

Fixes #4370


src/TraitsV2-Tests/T2AbstractTest.class.st
```diff
@@ -62,8 +62,8 @@ T2AbstractTest >> setUp [
 
 { #category : #running }
 T2AbstractTest >> tearDown [
-	createdClasses reverse do:#removeFromSystem.
+	createdClasses reverse do: #removeFromSystem.
```
juliendelplanque: `#reverseDo:` is normally better from a performance point of view since it does not create intermediate collection.

```suggestion
	createdClasses reverseDo: #removeFromSystem.
```

src/Collections-Sequenceable/OrderedCollection.class.st
```diff
@@ -685,6 +685,23 @@ OrderedCollection >> removeAt: index [
 	^removed
 ]
 
+{ #category : #removing }
+OrderedCollection >> removeDuplicates [
```
juliendelplanque, line 689: This method should be moved to `SequenceableCollection` so it becomes available for any of these collections.

juliendelplanque, line 689: Also, a unit test would be good.

src/Collections-Sequenceable/OrderedCollection.class.st
```diff
@@ -685,6 +685,23 @@ OrderedCollection >> removeAt: index [
...
+	| iterator |
+	"Remove the copies of elements, but keep the same order"
+	
+	self ifEmpty: [ ^ self ].
+	iterator := 1.
+	[ iterator <= self size ]
+		whileTrue: [ | each newIndex |
+			each := self at: iterator.
```
juliendelplanque: Here there is a possible optimisation (together with the code suggestion under) that allow to get better performance according to a benchmark I will attach in my main comment.
```suggestion
			each := self at: iterator.
			newIndex := iterator + 1.
```

src/Collections-Sequenceable/OrderedCollection.class.st
```diff
@@ -685,6 +685,23 @@ OrderedCollection >> removeAt: index [
...
+OrderedCollection >> removeDuplicates [
+	| iterator |
+	"Remove the copies of elements, but keep the same order"
+	
+	self ifEmpty: [ ^ self ].
+	iterator := 1.
+	[ iterator <= self size ]
+		whileTrue: [ | each newIndex |
+			each := self at: iterator.			
+			[ newIndex := (self indexOf: each startingAt: iterator+1).
```
juliendelplanque: Second part of the optimisation.
```suggestion
			[ newIndex := (self indexOf: each startingAt: newIndex).
```

juliendelplanque (approved):
I propose an optimisation for `OrderedCollection>>removeDuplicates`.

Here are the benchmarks I did to evaluate the gain:

```Smalltalk
col := ((1 to: 10e2) collect: [ :x | 10e1 atRandom ]) asOrderedCollection.

timesNormal := OrderedCollection new.
10e2 timesRepeat: [
	colForBench := col copy.
	timesNormal add: [ colForBench removeDuplicates ] timeToRunWithoutGC ].

timesOptimized := OrderedCollection new.
10e2 timesRepeat: [
	colForBench := col copy.
	timesOptimized add: [ colForBench removeDuplicatesOptimized ] timeToRunWithoutGC ].

timesNormal average asFloat.
 "3.41"
timesOptimized average asFloat
"0.011"
```

I think the optimisation is worth it.

src/Collections-Sequenceable/OrderedCollection.class.st
juliendelplanque: If this is done, we need to do specific implementation in case of non-growable collections.

jecisc: I just moved this function from somewhere else in the image to get this PR working.

Further improvements should be done in a new PR (But I agree with you)

MarcusDenker:
I merge this now, we can then do the improvements in a follow up PR

PR #4373, opened 2019-08-20
A formatter functionality according to the CommonLisp's one.

massimo-nocentini:
We want to implement a message that mimic the formatting functionalities provided by the CommonLisp's `format` function. In particular, we'll follow the approach of Seibel (https://github.com/gigamonkey/python-format) and validate it against http://www.gigamonkeys.com/book/a-few-format-recipes.html and the reference http://www.lispworks.com/documentation/lw50/CLHS/Body/22_c.htm.

We open this PR immediately to allow discussion about our effort and tracking either missing or desired feature provided by the original `format`.

### DIFF
--- a/src/Collections-Strings/String.class.st
+++ b/src/Collections-Strings/String.class.st
@@ -843,6 +843,11 @@ String >> charactersExactlyMatching: aString [
 	^ count
 ]
 
+{ #category : #formatting }
+String >> clFormat: arguments [
+	self shouldBeImplemented 
+]
+
 { #category : #comparing }
 String >> compare: aString [ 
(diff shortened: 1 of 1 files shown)


Ducasse:
Massimo the first thing to do is to provide an implementation and tests and discuss it  on the mailing-list.

massimo-nocentini:
Okay Stef, next week I'll be at ESUG and I'll hope to meet there to talk about this stuff.

MarcusDenker:
Yes, PRs with no content for discussion are not good: we want to move fast with PRs and never have more then around 20 max "in flight".
For everything else, there is the issue tracker.

PR #4375, opened 2019-08-20
Traits-tests-should-work-with-slots-definitions

jecisc:
Make trait tests pass when slot definition is shown.


MarcusDenker (approved):
ok, test passed, too

PR #4378, opened 2019-08-21
4282-Backport-Pharo7-trying-to-add-an-ivar-to-UndefinedObject-crashes-the-VM-4017

MarcusDenker:
backport to warn when trying to add ivar to UndefinedObject


MarcusDenker:
as this is in Pharo8 since a week --> merge

PR #4382, opened 2019-08-21
CodeGenWeakClassVar

MarcusDenker:
add code generation for WeakClassVariable. There are already tests for this, see #testReadWriteWeakClassVarCompiled


bencoman:
What is the philosophy behind a WeakClassVariable?  How would it work / help?

jecisc:
@bencoman The idea is to have an equivalent of the WeabkSlot for class variables. 

With it, we could declare that a Slot is Weak and we would not need to create a WeakArray of size one to hold weak variables.

PR #4384, opened 2019-08-21
Improve-WeakValueDictionary

jecisc:
Improve WeakValueDictionary in three way:
- Document the internal mecanism to explain why we need a wrapper around the values
- Fix bug in #at:ifPresent: (with test)
- Rename SetElement into CollectionElementWrapper since this class is use by other collections than Set


dionisiydk:
Hi @jecisc.

Why SetElement is a bad name?
It was introduced specifically to allow a Set to include nil. It has zero relation to weak structures.

IMHO proposal rename feels bad because
```Smalltalk
anObject asCollectionElementWrapper == anObject "==> true"
``` 
It is not logical from the words meaning. I would expect a wrapper instance as a result.

In contrast
```Smalltalk
anObject asSetElement == anObject "==> true"
``` 
looks fine because any object can be an element of a Set. 
With your proposal we will say that any object is a wrapper for collection element. It sounds weird.

jecisc:
Hi,

I proposed to rename it because SetElement is not only used by sets but also by other collections such as weak collections. So, IMO, it is not a good name. (I caused me trouble when I was looking at an issue with WeakValueDictionary).

CollectionElementWrapper is not a name that I really like but I did not found anything better yet. I am open to suggestions about that. CollectionElement would be better?

jecisc:
> It has zero relation to weak structures.

@dionisiydk It is used to manage internal behavior of WeakValueDictionary

dionisiydk:
> > It has zero relation to weak structures.
> 
> @dionisiydk It is used to manage internal behavior of WeakValueDictionary

Right

dionisiydk:
> Hi,
> 
> I proposed to rename it because SetElement is not only used by sets but also by other collections such as weak collections. So, IMO, it is not a good name. (I caused me trouble when I was looking at an issue with WeakValueDictionary).
> 
> CollectionElementWrapper is not a name that I really like but I did not found anything better yet. I am open to suggestions about that. CollectionElement would be better?

Looks better to me

jecisc:
I’ll update it tomorrow.

Thank you for the review! I really appreciate it :)

dionisiydk:
Thank you too @jecisc . You are doing very good job

jecisc:
I did new commit to the branch but they do not appear... I'll open a new PR.

jecisc:
@dionisiydk https://github.com/pharo-project/pharo/pull/4396

PR #4385, opened 2019-08-21
Remove-duplication-in-Dictionary

jecisc:
Remove duplicated code in Dictionary


MarcusDenker:
It is really strange that git thinks that Dictionary is Binary...

MarcusDenker:
there is a conflict now

jecisc:
@MarcusDenker Removing the non ascii character worked!

MarcusDenker:
at:put: is used a lot... a benchmark would be interesting

src/Collections-Unordered/Dictionary.class.st
```diff
@@ -320,16 +320,11 @@ Dictionary >> at: key ifPresent: aPresentBlock ifAbsentPut: anAbsentBlock [
...
 	"Set the value at key to be anObject.  If key is not found, create a
 	new entry for key and set is value to anObject. Answer anObject."
 
-	| index assoc |
-	index := self findElementOrNil: key.
-	assoc := array at: index.
-	assoc
-		ifNil: [self atNewIndex: index put: (Association key: key value: anObject)]
-		ifNotNil: [assoc value: anObject].
+	self add: (Association key: key value: anObject).
```
juliendelplanque, line 327: Out of curiosity, why not using :
```suggestion
	self add: key -> anObject.
```

So there is no direct dependency on the Association class?

juliendelplanque (approved):
Looks nice.

src/Collections-Unordered/Dictionary.class.st
VincentBlondeau: #-> does reference Association. So it will not change anything. But another issue it that -> is made to be used by the user library.

VincentBlondeau, line 327: ` . ^anObject`
is missing at the end.

VincentBlondeau (changes requested):

MarcusDenker:
i still wonder about performance: #at:put: needs to be fast...

at:put: now:
	assoc
		ifNil: [self atNewIndex: index put: (Association key: key value: anObject)]
		ifNotNil: [assoc value: anObject].

vs. add:
	element
		ifNil: [ self atNewIndex: index put: anAssociation ]
		ifNotNil: [ element value: anAssociation value ].

means that if we call at:put: on an existing entry, we right now do *not* create an association object, 
but with the new code we would do that.

Ducasse:
We should bench it. And for now I would not change.

PR #4395, opened 2019-08-22
Iceberg-Git-thinks-Dictionary-is-a-binary-file

### LINKED ISSUE #4393: Iceberg: Git thinks Dictionary is a binary file
BIN -158 Bytes (99%) src/Collections-Unordered/Dictionary.class.st 

This means GitHub can not show Difs

jecisc:
Remove non ascii character in Dictionary.

Fixes #4393


jecisc:
This PR remove a non ascii character and I hope it will be enough to fix the problem. If after the merge it's still not enough I'll look a little more into the issue.

PR #4396, opened 2019-08-22
Improve-WeakValueDictionary

jecisc:


jecisc:
Improve WeakValueDictionary in three-way:
* Document the internal mechanism to explain why we need a wrapper around the values
* Fix bug in #at:ifPresent: (with test)
* Rename SetElement into CollectionElement since this class is used by other collections than Set


Follows https://github.com/pharo-project/pharo/pull/4384

MarcusDenker (approved):

Issue #4394, opened 2019-08-22
Mac OS primitive broken for NetNameResolver class>>#localHostAddress


bouraqadi:
On Mac NetNameResolver class>>#localHostAddress answers 127.0.0.1 which is the loopback address instead of the actual IP address on the network. The comment of the methods states the following
```
<todo>
"On Mac the primitive is buggy and can return an empty IP address. 
Use a standard value in that case "
```

PR #4398, opened 2019-08-22
Fix for  issue #4394. Hack to be used until the VM primitive is fixed.

bouraqadi:
Fix for NetNameResolver localHostAddress on Mac

### DIFF
--- a/src/Network-Kernel/NetNameResolver.class.st
+++ b/src/Network-Kernel/NetNameResolver.class.st
@@ -166,11 +166,10 @@ NetNameResolver class >> localHostAddress [
 
 { #category : #lookups }
 NetNameResolver class >> localHostAddressHackForMac [
-
 	"This is a hack. The best solution should be done by fixing the primitive on Mac"
 	| ipString |
-	ipString := Smalltalk os resultOfCommand: 'curl ipecho.net/plain'.
-	ipString ifEmpty: [ ^self loopBackAddress ].
+	ipString := Smalltalk os resultOfCommand: 'ipconfig getifaddr en0'. 
+	ipString ifEmpty:	[^self loopBackAddress].
 	^SocketAddress fromDottedString: ipString
 
 ]
--- a/src/Network-Kernel/NetNameResolver.class.st
+++ b/src/Network-Kernel/NetNameResolver.class.st
@@ -158,10 +158,21 @@ NetNameResolver class >> localHostAddress [
 	"On Mac the primitive is buggy and can return an empty IP address. Use a standard value in that case"
 		
 	| primAddress |
-
+	Smalltalk os isMacOS ifTrue: [ ^self localHostAddressHackForMac ].
 	self initializeNetwork.
 	[ primAddress := self primLocalAddress ] on: PrimitiveFailed do: [ :err | primAddress := #[0 0 0 0] ].
 	^ (primAddress = #[0 0 0 0] ifTrue: [ #[127 0 0 1] ] ifFalse: [ primAddress ]) asSocketAddress
+]
+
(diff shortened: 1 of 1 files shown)


svenvc:
First remark: we can perfectly do HTTP in Pharo ;-)

    ZnClient new beOneShot; enforceHttpSuccess: true; get: 'http://ipecho.net/plain'.

Second remark: using a web service call to another server does not strike me as a good implementation of the system functionality 'what is my own IP address'. What if the internet is not reachable ? What if the service is down ? What about the speed ?

Furthermore, the answer seems wrong: it gives the external IP, not the internal one assigned to the machine. All computers on our internal office network will get the same answer, while they all have a different internal IP address assigned.

bouraqadi:
Thanks Sven. That's a good catch.
Any hint about how to get the internal IP?

Noury

svenvc:
This is really much more complicated than it would seem: on most machines there are multiple network interfaces (and virtual interfaces), so there is not one single answer.

The problem is also that any solution needs implementations on the 3 main platforms (macOS, windows, linux).

bouraqadi:
There seems to be a VM plugin doing the job for Linux and Windows. 
We need a solution, only for Mac.

Noury

bouraqadi:
Found this, that works on my Mac and gives me the IP string.
What do you think?

Noury

Smalltalk os resultOfCommand: 'ipconfig getifaddr en0'

svenvc:
In terms of result, on macOS, this is probably OK.

However, there can be multiple interfaces, en0 is just one name, so it won't be correct in all cases, I am afraid.

    networksetup -listallhardwareports

noha:
It is only ok when your Mac has one network interface. A few years back when there was cable and wifi wifi would have been en1 and not be found. There is usually one primary network interface and that is the one that as an IP address in the same net as the default route

astares:
Why should it work only on Mac? Do the other platform not count anymore?

We should check if we could solve it by wrapping "getifaddrs()" function using UFFI

bouraqadi:
Thanks everyone for the feedback.

Just to remind the context.
NetNameResolver localhostAddress seems to work on all platforms, except on Mac.
The comment of the method says that its a bug in the socket plugin version for Mac OS.
This is why the focus here is Mac

This is a method of NetNameResolver that answers only 1 address.
Definitely, we should consider changing the API and improving it.
But, meanwhile, it would be good to have a quick fix.
Hence my proposal.

I don't know what is the default interface on Mac OS.
I took en0.
Any hint on this?
> wrapping "getifaddrs()" function using UFFI
> 

Yes, this would be better. 
But, does it worth it to spend more time on this hack?

bouraqadi:
I did a new pull request with the same issue number.

Noury

MarcusDenker:
so this can be closed?

bouraqadi:
Hi,

This issue is closed but the commits with the fix were not merged.

Noury

Ducasse:
Hi guys

can you have a look and that we conclude?

S.

MarcusDenker:
But why keep *this* pull request open if there is another one? You clearly stated "I did a new pull request with the same issue number".

bouraqadi:
Sorry, got lost between the two. One appears under: pull requests, and the other as issue...
https://github.com/pharo-project/pharo/pull/4398 <https://github.com/pharo-project/pharo/pull/4398>
and
https://github.com/pharo-project/pharo/issues/4394 <https://github.com/pharo-project/pharo/issues/4394>

Anyway, the point is that we have a workaround for a Mac VM bug that was not integrated.

Noury

Ducasse:
Marcus we simply did not see it.

PR #4400, opened 2019-08-22
WeakValueDictionary-return-association-with-collectionElements-instead-of-nil

jecisc:
Fix bug in WeakValueDictionary when #assotiationAt: was returning the wrong kind of association.

```Smalltalk
testReturnedAssociationsAreRight
	| dictionary |
	dictionary := WeakValueDictionary new.
	dictionary add: (GlobalVariable key: 'test' value: nil).
	dictionary add: (GlobalVariable key: 'test2' value: 1).
	self assert: (dictionary associationAt: 'test') value isNil.
	self assert: (dictionary associationAt: 'test2') value equals: 1.
	self assert: ((dictionary associationAt: 'test') isKindOf: GlobalVariable)
```

### DIFF
--- a/src/Collections-Tests/WeakValueDictionaryTest.class.st
+++ b/src/Collections-Tests/WeakValueDictionaryTest.class.st
@@ -132,6 +132,17 @@ WeakValueDictionaryTest >> testKeysAndValuesDoWithGarbageCollectedValue [
 ]
 
 { #category : #tests }
+WeakValueDictionaryTest >> testReturnedAssociationsAreRight [
+	| dictionary |
+	dictionary := self classToBeTested new.
+	dictionary add: (GlobalVariable key: 'test' value: nil).
+	dictionary add: (GlobalVariable key: 'test2' value: 1).
+	self assert: (dictionary associationAt: 'test') value isNil.
+	self assert: (dictionary associationAt: 'test2') value equals: 1.
+	self assert: ((dictionary associationAt: 'test') isKindOf: GlobalVariable)
+]
+
+{ #category : #tests }
 WeakValueDictionaryTest >> testSizeWithGarbageCollectedValue [
 
 	self
--- a/src/Collections-Tests/WeakValueDictionaryTest.class.st
+++ b/src/Collections-Tests/WeakValueDictionaryTest.class.st
@@ -135,11 +135,10 @@ WeakValueDictionaryTest >> testKeysAndValuesDoWithGarbageCollectedValue [
 WeakValueDictionaryTest >> testReturnedAssociationsAreRight [
 	| dictionary |
 	dictionary := self classToBeTested new.
-	dictionary add: (GlobalVariable key: 'test' value: nil).
-	dictionary add: (GlobalVariable key: 'test2' value: 1).
+	dictionary at: 'test' put: nil.
+	dictionary at: 'test2' put: 1.
 	self assert: (dictionary associationAt: 'test') value isNil.
-	self assert: (dictionary associationAt: 'test2') value equals: 1.
-	self assert: ((dictionary associationAt: 'test') isKindOf: GlobalVariable)
+	self assert: (dictionary associationAt: 'test2') value equals: 1
 ]
 
 { #category : #tests }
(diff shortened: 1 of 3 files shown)


jecisc:
The test was failing because the value for 'test' was a CollectionElement and not nil and the association kind was WeakValueAssociation and not GlobalVariable.

src/Collections-Tests/WeakValueDictionaryTest.class.st
```diff
@@ -131,6 +131,17 @@ WeakValueDictionaryTest >> testKeysAndValuesDoWithGarbageCollectedValue [
...
 
+{ #category : #tests }
+WeakValueDictionaryTest >> testReturnedAssociationsAreRight [
+	| dictionary |
+	dictionary := self classToBeTested new.
+	dictionary add: (GlobalVariable key: 'test' value: nil).
+	dictionary add: (GlobalVariable key: 'test2' value: 1).
+	self assert: (dictionary associationAt: 'test') value isNil.
+	self assert: (dictionary associationAt: 'test2') value equals: 1.
+	self assert: ((dictionary associationAt: 'test') isKindOf: GlobalVariable)
```
dionisiydk: To be a binding registry (like Undeclared) the another invariant is expected by system:
```Smalltalk
self assert: (dictionary associationAt: 'test') == (dictionary associationAt: 'test') 
```
But it will fail with this PR.

For example undeclared variable references will stop working: check the #readsRef: method.

src/Collections-Weak/WeakValueDictionary.class.st
```diff
@@ -16,10 +16,32 @@ Class {
...
+	"Set the value at key to be anObject.  If key is not found, create a new
+	entry for key and set is value to anObject. Answer anObject."
+
+	| index |
+	index := self findElementOrNil: anAssociation key.
+	(array at: index)
+		ifNil: [ self
+				atNewIndex: index
+				put:
+					((WeakValueAssociation key: anAssociation key value: anAssociation value asCollectionElement)
```
dionisiydk: Let's introduce #asWeakValueAssociation. So if it is already weak we will not create a copy.

src/Collections-Weak/WeakValueDictionary.class.st
```diff
@@ -16,10 +16,32 @@ Class {
...
 ]
 
+{ #category : #adding }
+WeakValueDictionary >> associationAt: key ifAbsent: aBlock [
+	"Answer the association with the given key.
+	If the key is not found, return the result of evaluating aBlock."
+
+	^ (array at: (self findElementOrNil: key))
+		ifNil: [ aBlock value ]
+		ifNotNil: [ :assoc | assoc asOriginalAssociation ]
```
dionisiydk: I do not like the idea that dictionary should take care about original association.
We store #key and #value. The association is an implementation detail or suitable transfer object to deliver a pair to the dictionary. And I think it is better to keep it like that. Extra responsibility will complicate logic. 

You are trying to make Undeclared structure to be a weak. We can modify users to use normal  accessing methods of the dictionary (#at:/#at:put:) instead of #add:  and #associationsAt.
The #add: here was probably for optimization to avoid extra wrapper objects (associations) to store bindings. But I really doubt it is important here because Undeclared is normally empty. 

What you think to try another approach? I believe it will be simpler

dionisiydk (changes requested):

src/Collections-Weak/WeakValueDictionary.class.st
jecisc: @dionisiydk Thinking more about it, indeed this should not be in WeakValueDictionary to keep the association. We should fix the problem with the value been nil, but not the problem of keeping the right association.

For Undeclared we should maybe create another class for it. I don't know if using #at: will be enough because I guess the system uses the kind of association to perform things. I will talk more with Marcus.

src/Collections-Weak/WeakValueDictionary.class.st
jecisc: Indeed!

jecisc:
@dionisiydk I updated the PR to only fix the #isNil bug.

dionisiydk (approved):

PR #4405, opened 2019-08-23
Roassal3Extentions

akevalion:
Simple extension for 

- Rectangle, floatCenter, because #center in Rectangle returns the Integer Point and not the Float Point
- Announcer(accessing subscribers by class), Useful to know the subscriber for the announcer.

from Roassal3 package

### DIFF
--- a/src/Announcements-Core/Announcer.class.st
+++ b/src/Announcements-Core/Announcer.class.st
@@ -29,6 +29,12 @@ Announcer >> basicSubscribe: subscription [
 ]
 
 { #category : #testing }
+Announcer >> handleSubscriberClass: eventClass [
+	self subscriptions ifNil: [ ^ false ].
+	^ self subscriptions handleSubscriberClass: eventClass
+]
+
+{ #category : #testing }
 Announcer >> hasSubscriber: anObject [
 
 	registry subscriptionsOf: anObject do: [:each | ^ true].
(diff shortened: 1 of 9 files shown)


src/Announcements-Core/SubscriptionRegistry.class.st
```diff
@@ -52,6 +52,14 @@ SubscriptionRegistry >> deliver: anAnnouncement to: subs startingAt: startIndex
 				self deliver: anAnnouncement to: subs startingAt: index + 1 ]]
 ]
 
+{ #category : #testing }
+SubscriptionRegistry >> handleSubscriberClass: eventClass [
+	"Return true if the receiver has a callback subscripbed for the event class"
+	^ subscriptions anySatisfy: [ :sub | 
+		(sub subscriber class == eventClass) or: 
+			[ sub subscriber class includesBehavior: eventClass ] ]
```
dionisiydk, line 60: It should be at least #handles... with ending s as a common convention for such methods.
But generally name does not sound good in English. It should be something like
```Smalltalk
announcer isSubscribedToAnnouncementClass:  eventClass
``` 
and 
```Smalltalk
subscriptionRegistry includesSubscriptionsToAnnouncementClass:  eventClass
```

src/Announcements-Core/Announcer.class.st
```diff
@@ -88,6 +94,12 @@ Announcer >> subscriptions [
 	^ registry
 ]
 
+{ #category : #accessing }
+Announcer >> subscriptionsForClass: subscriberClass [
+	"Return the list of subscription for a given class"
+	^ self subscriptions subscriptionsForClass: subscriberClass
```
dionisiydk, line 100: And here the better name would be:
```Smalltalk
announcer subscribersKindOf:  subscriberClass
```
and 
```Smalltalk
subscriptionRegistry subscribersKindOf:  subscriberClass
```

dionisiydk:
@Ducasse Would be nice to review new API to core classes before merging.

PR #4409, opened 2019-08-23
4406-fix-class-def-parser-for-simple-definitions

### LINKED ISSUE #4406: fix class def parser for simple definitions
the parseSlot method is missing this:

aRBMessageNode isLiteralNode ifTrue: [ | slot |
		slot := CDSlotNode
			node: aRBMessageNode
			name: aRBMessageNode value
			typeName: aRBMessageNode value
			start: aRBMessageNode start
			stop: aRBMessageNode stop.
		classDefinition addSlot: slot.
		^ self
	].

And we need to test complex Slots.. (e.g. #var => LazySlot default: 5) but that is another issue

MarcusDenker:
support simplified slot defintions (using #symbol). fixes #4406

### DIFF
--- a/src/ClassParser/CDClassDefinitionParser.class.st
+++ b/src/ClassParser/CDClassDefinitionParser.class.st
@@ -199,7 +199,6 @@ CDClassDefinitionParser >> parseSlotNode: aRBMessageNode [
 			^ self ].
 
 	"If we are here we did not recognize the slot syntax"
-	self halt
 ]
 
 { #category : #parsing }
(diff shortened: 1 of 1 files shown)


dionisiydk (approved):
For future improvements:
This method looks like it should go to the message node itself:
```Smalltalk
aRBMessageNode asSlotClassDefinitionNode
```

dionisiydk:
Hey,  we need a test for this case.
It should be easy to add as there are other tests for this parser

MarcusDenker:
I wil add tests later in another PR

Ducasse:
What does it mean "aRBMessageNode isLiteralNode" ?

MarcusDenker:
It is not a message. I just did a quick workaround to unblock people that wanted to work, I did that on the train with bad internet.
More work is needed.

PR #4413, opened 2019-08-24
Class Variables with Default value

MarcusDenker:
InitializedClassVariable and LazyClassVariable, with tests

#name => LazyClassVariable default: 5.

can now be used to create a class var with default value


dionisiydk (commented):
Hi Marcus.

I think it would be nice to hide those classes from the user:
```Smalltalk
var => ClassVariable default: 5
```
It will produce InitializedClassVariable.
And for lazy:
```Smalltalk
var => ClassVariable lazy default: 5
```

src/Slot-Examples/AbstractInitializedClassVariable.class.st
```diff
@@ -0,0 +1,40 @@
...
+		'default',
+		'isBlock'
+	],
+	#category : #'Slot-Examples-ClassVariables'
+}
+
+{ #category : #accessing }
+AbstractInitializedClassVariable >> default: anObject [
+	default := anObject.
+	isBlock := false
```
dionisiydk, line 21: isBlock variable looks redundant.
If there is anyway isBlock ifTrue logic then just ask "default isBlock"

MarcusDenker: I for now did it in a way that you can specify a block as the default, but not evaluate it.
AbstractInitializedClassVariable default: []
vs
AbstractInitializedClassVariable defaultBlock: []

in the second case the block is evaluated, in the first not

I was trying to do it by always wrapping in a block, but that is not nice either.

MarcusDenker:
The problem with

var => ClassVariable default: 5

is that there will be other ClassVariables, too. And I wanted to follow the style of Slots for now...

MarcusDenker:
(I want in the end a unified way:

#myVar => Observable + Lazy default: 5

should work for both slots and class vars... )

src/Slot-Examples/AbstractInitializedClassVariable.class.st
MarcusDenker: But I think it is possible to simplify: just have default:, and if people want a block they have to do [[]] 

I will try that...

jecisc: Can’t default be 

default: anObject
    self defaultBlock: [ anObject ]

?

Or the other way around. I am on my phone and it’s not easy to read the code.

MarcusDenker:
I made a PR with a version like Denis suggested: https://github.com/pharo-project/pharo/pull/4420

PR #4415, opened 2019-08-24
4414-slot-class-defintion-Class-Variables-with-parameters-are-not-printed-correctly

### LINKED ISSUE #4414: slot class defintion: Class Variables with parameters are not printed correctly
When trying to use a class variable with parameter:

Object subclass: #MyTestClass
	slots: {  }
	classVariables: { #LazyVar => LazyClassVariable default: 5 }
	package: 'Spec2-Core-Widgets-Tab'

after accept, it prints as 

Object subclass: #MyTestClass
	slots: {  }
	classVariables: { #LazyVar => LazyClassVariable}
	package: 'Spec2-Core-Widgets-Tab'

MarcusDenker:
fix #definitionString for LiteralVariable to use the printString as overriden by the subclasses. fixes #4414


MarcusDenker:
somehow this is the wrong PR

PR #4416, opened 2019-08-24
4414-slot-class-defintion-Class-Variables-with-parameters-are-not-printed-correctly

MarcusDenker:
slot class defintion: Class Variables with parameters are not printed correctly


MarcusDenker:
I make something wrong again

PR #4420, opened 2019-08-25
Simplify Variables with default blocks

MarcusDenker:
- add code generation for Class Variable wiit default
- simplify #default: for block: 
-- remove #defaultBlock
-- if the default is a block, do a #cull:

### DIFF
--- a/src/Slot-Examples/InitializedClassVariable.class.st
+++ b/src/Slot-Examples/InitializedClassVariable.class.st
@@ -16,6 +16,6 @@ Class {
 { #category : #'class building' }
 InitializedClassVariable >> installingIn: aClass [
 	default isBlock 
-		ifTrue: [ self write: (default cull: aClass)]
-		ifFalse: [ self write: default copy]
+		ifTrue: [ self write: (default cull: aClass) ]
+		ifFalse: [ self write: default ]
 ]
--- a/src/Slot-Examples/InitializedClassVariable.class.st
+++ b/src/Slot-Examples/InitializedClassVariable.class.st
@@ -15,7 +15,7 @@ Class {
 
 { #category : #'class building' }
 InitializedClassVariable >> installingIn: aClass [
-	isBlock 
+	default isBlock 
 		ifTrue: [ self write: (default copy cull: aClass)]
-		ifFalse: [ self write: default]
+		ifFalse: [ self write: default copy]
 ]
--- a/src/Slot-Examples/InitializedClassVariable.class.st
+++ b/src/Slot-Examples/InitializedClassVariable.class.st
@@ -16,6 +16,6 @@ Class {
 { #category : #'class building' }
 InitializedClassVariable >> installingIn: aClass [
 	default isBlock 
-		ifTrue: [ self write: (default copy cull: aClass)]
+		ifTrue: [ self write: (default cull: aClass)]
 		ifFalse: [ self write: default copy]
 ]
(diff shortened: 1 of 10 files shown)


src/Slot-Examples/InitializedClassVariable.class.st
```diff
@@ -15,7 +15,7 @@ Class {
 
 { #category : #'class building' }
 InitializedClassVariable >> installingIn: aClass [
-	isBlock 
-		ifTrue: [ self write: (default copy cull: aClass)]
-		ifFalse: [ self write: default]
+	default isBlock 
+		ifTrue: [ self write: (default cull: aClass)]
+		ifFalse: [ self write: default copy]
```
dionisiydk: Why the #copy is required here?

src/Slot-Examples/InitializedSlot.class.st
```diff
@@ -15,9 +15,9 @@ Class {
 
 { #category : #initialization }
 InitializedSlot >> initialize: anObject [
-	isBlock 
-		ifTrue: [ self write: (default copy cull: anObject) to: anObject ]
-		ifFalse: [ self write: default to: anObject ]
+	default isBlock 
+		ifTrue: [ self write: (default cull: anObject) to: anObject ]
+		ifFalse: [ self write: default copy to: anObject ]
```
dionisiydk: And #copy is duplicated in few places

src/Slot-Examples/LazySlot.class.st
```diff
@@ -26,12 +26,14 @@ LazySlot >> emitValue: aMethodBuilder [
 		pushLiteral: nil;
 		send: #==;
 		jumpAheadTo: #target if: false;
-		popTop;
-		pushLiteral: default.
-	isBlock ifTrue: [ 
+		popTop.
+	default isBlock
```
dionisiydk: Having so many "default isBlock ifTrue/ifFalse" statements it looks like we need to introduce object here. Like VariableInitializer with subclasses ConstantVariableInitializer and BlockVariableInitializer. And replace the #default var with #initialzer var.

It could also allow to avoid some subclasses of ClassVariable. I wonder if we can move #initializer variable directly to the base class ClassVariable which will be NullVariableInitializer by default.

src/Slot-Tests/LazyClassVariableTest.class.st
```diff
@@ -39,7 +39,7 @@ LazyClassVariableTest >> testLazyClassVarReadReflective [
 { #category : #tests }
 LazyClassVariableTest >> testLazyClassVarReadReflectiveBlock [
 	| classVar |
-	classVar := #ClassVar => LazyClassVariable defaultBlock: [5].
+	classVar := #ClassVar => LazyClassVariable default: [5].
 	aClass := self make: [ :builder | builder sharedVariables: {classVar}].
```
dionisiydk, line 43: Side question. I noticed that class builder uses name sharedVariable. Do we have a plan to rename ClassVariable to SharedVariable?

MarcusDenker: We use the term "shared variable" *only* in the API if the class builder.

Thus tried to do the opposite and change the API of the classbuilder, but it broke the bootstrap. Need to look at it again.

src/Slot-Examples/InitializedClassVariable.class.st
MarcusDenker: maybe it is not needed... we could make the default read only instead. I just wanted to not have people change the default if it is a string...

src/Slot-Examples/LazySlot.class.st
MarcusDenker: I will think about that

src/Slot-Examples/InitializedSlot.class.st
MarcusDenker: I will remove the copy

PR #4423, opened 2019-08-25
4088-Commit-should-have-been-implemented-in-IceUnknownCommit

### LINKED ISSUE #4088: Commit should have been implemented in IceUnknownCommit

jecisc:
Inform the user of error while trying to commit with a non operational git repository.

A Camp smalltalk fix with James Foster ;)

Fixes #4088

### DIFF
--- a/src/Calypso-SystemPlugins-Monticello-Browser/ClyCommitMCPackageCommand.class.st
+++ b/src/Calypso-SystemPlugins-Monticello-Browser/ClyCommitMCPackageCommand.class.st
@@ -43,19 +43,27 @@ ClyCommitMCPackageCommand >> defaultMenuItemName [
 { #category : #execution }
 ClyCommitMCPackageCommand >> execute [
 	| repoBrowser commitBrowser repos targetRepo |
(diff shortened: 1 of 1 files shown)


VincentBlondeau (approved):
targetRepo isMissing or: [ targetRepo isCodeMissing or: [ targetRepo isDetached or: [ targetRepo hasUnbornProject ] ] ]
Could be extracted to a method in the targetRepo class and message can be adapted accordingly too.

VincentBlondeau:

Ducasse:
may be we should commit this to iceberg repo?

tesonep:
Yes, it should go in iceberg and maybe using polymorphism for handling the messages instead of a chained condition (only the nil, we should not add messages to nil)

jecisc:
I agree it should be improved and probably moved. I did this PR to at least fix it until we have time to further improve it because it currently causes trouble to newcomers to get an exception.

MarcusDenker:
Changes to Iceberg tend to take very long till they end up in Pharo. So having a workaround is good

Issue #4429, opened 2019-08-26
Extract method does not check for duplicated instances of the refactored code


Lin777:
This change to improve extract method refactoring

Ducasse:
C00L :)
May be we should add an option at the end of any refactporing menu sequence

Lin777:
I was thinking the same thing but I didn't know exactly in which menu I would put it, because it could be in the class, method, variable or source code browser. And as I saw it very repetitive I decided to put the undo last refactoring command where it is now.

Ducasse:
Hi evelyn 
I think that we talked about a different issue than the title. 
Could you check? Because I was talking about the undo and you too. 
Do you know if extract method is looking for duplicated instances?

PR #4430, opened 2019-08-26
Add missing find and replace refactoring

### LINKED ISSUE #4428: Missing find and replace refactoring
Consists of selecting a method and looking for occurrences of this in other methods

Lin777:
Link video: https://www.youtube.com/watch?v=yxOn3tdh1wU

Fix #4428, #4429

### DIFF
--- a/src/Refactoring-Core/RBExtractMethodRefactoring.class.st
+++ b/src/Refactoring-Core/RBExtractMethodRefactoring.class.st
@@ -17,7 +17,8 @@ Class {
 		'parameters',
 		'needsReturn',
 		'selectorNewMethod',
-		'parameterMap'
+		'parameterMap',
+		'newExtractedSelector'
 	],
 	#category : #'Refactoring-Core-Refactorings'
 }
@@ -122,6 +123,11 @@ RBExtractMethodRefactoring >> checkTemporaries [
 ]
 
 { #category : #transforming }
+RBExtractMethodRefactoring >> compiledMethod [
+	^ class realClass >> selector
+]
+
+{ #category : #transforming }
 RBExtractMethodRefactoring >> createTemporariesInExtractedMethodFor: assigned [ 
 	assigned do: [:each | extractedParseTree body addTemporaryNamed: each]
 ]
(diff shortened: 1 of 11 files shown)


src/Refactoring-Help/Collection.extension.st
```diff
@@ -0,0 +1,13 @@
+Extension { #name : #Collection }
+
+{ #category : #'*Refactoring-Help' }
+Collection >> collectAll: aBlock [
```
dionisiydk, line 4: There is flatCollect for this feature

Ducasse:
This is cool that we can apply extract refactoring in subclasses.

src/Refactoring-Core/RBExtractMethodRefactoring.class.st
```diff
@@ -121,6 +121,22 @@ RBExtractMethodRefactoring >> checkTemporaries [
 				yourself
 ]
 
+{ #category : #transforming }
+RBExtractMethodRefactoring >> compiledMethod [
+	^ self compiledMethodIn: class realClass
+]
+
+{ #category : #transforming }
+RBExtractMethodRefactoring >> compiledMethodIn: aClass [
```
dionisiydk, line 130: There is a method for this:
   aClass lookupSelector: aSelector

Lin777:
Thanks for the revisions @dionisiydk, I will correct them in a new PR

dionisiydk:
Hi @Lin777 .
I know that it was not trivial task. But code looks quite complex. And there are just two tests.

I think many logic can be simplified if it will be moved from refactoring to AST nodes. And it can be covered by nice unit tests.
Like converting method to pattern node:
```Smalltalk
aMethod ast asPatternNode
```
It will return new ast-nodes recursively replacing original variable nodes with RBPatternVariableNode's. And it will not require generating sources for RB-rewrite expression.

That's just suggestion for improvement. 
Good job anyway.

src/SystemCommands-MessageCommands/SycFindAndReplaceMethodCommand.class.st
```diff
@@ -0,0 +1,59 @@
+"
+I am a command to find and replace selector in a range of methods, with a call to itself
+"
+Class {
+	#name : #SycFindAndReplaceMethodCommand,
+	#superclass : #SycChangeMessageSignatureCommand,
```
dionisiydk, line 6: It does not look like a message related command because it is based on ast of the method. I think it should be a kind of SycMethodCommand (it's based on multiple methods but you can require single method in your class)

Lin777:
I understand, it was a really complex task but I'll try to fix those details by following your advice. Thank you very much for the reviews :).

PR #4434, opened 2019-08-28
4433-Pragmakeyword-should-be-deprecated-via-deprecation-mechanism

### LINKED ISSUE #4433: Pragma>>#keyword should be deprecated via deprecation mechanism
According to the method's comment, this method should be deprecated. It is not.

juliendelplanque:
- Deprecated pragma methods that needed to be.
- Migrated code to not use deprecated pragma API.

Fixes #4433


juliendelplanque:
Ok so now apparently the failing tests existed before this PR.

Should be ready for integration.

PR #4436, opened 2019-08-28
[CI CLEANING] Fixing current test failures

VincentBlondeau:
Due to the integration of some PRs that were failing (including this one: https://github.com/pharo-project/pharo/pull/4430), we now have some tests that are failing in the main branch of Pharo.

This PR aims to solve them.

But one thing that I do not understand is why some PR are accepted despite the tools says that they are breaking some tests? Maybe I do not understand something in the design of the workflow...

### DIFF
--- a/src/Refactoring-Changes/AbstractTool.extension.st
+++ b/src/Refactoring-Changes/AbstractTool.extension.st
@@ -1,0 +1,17 @@
+Extension { #name : #AbstractTool }
+
+{ #category : #'*Refactoring-Changes' }
+AbstractTool class >> undoLastRefactoring [
+	| manager |
+	manager := RBRefactoryChangeManager instance.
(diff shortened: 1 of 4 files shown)


badetitou (approved):

badetitou (approved):

bencoman:
I guess its due to frequent "randomly" occurring errors - which leads to judgement calls like "those errors look unrelated to my changes" - which risks being wrong and letting new problems slip in, which further undermines the integrity of the CI reports.   One way to get some credibility back into CI reports would be for failed tests to retested to see if passes the second time.

VincentBlondeau:
@bencoman Yes they should. We could also annotate them with a pragma that sayswhen was the last fail, so we can track them.

PR #4438, opened 2019-08-28
4435-Cannot-browse-trait-in-class-description

dionisiydk:
- Fix trait definition parsing when uses section is empty
- tests for trait creation


dionisiydk:
it fixes #4435

PR #4440, opened 2019-08-29
4439-Epicea-Should-FileOut-to-Image-Directory

### LINKED ISSUE #4439: Epicea Should FileOut to Image Directory
Currently they end up mixed in with the session files, which doesn't seem to make a lot of sense.

seandenigris:
Fix #4439

### DIFF
--- a/src/EpiceaBrowsers/EpLogBrowserPresenter.class.st
+++ b/src/EpiceaBrowsers/EpLogBrowserPresenter.class.st
@@ -324,11 +324,10 @@ EpLogBrowserPresenter >> eventsMenuActions [
 { #category : #'menu - operations' }
 EpLogBrowserPresenter >> fileOutSelection [
 
-	| entries outputDirectory |
+	| entries |
 	entries := self selectedEntryItems collect: [ :each | each entry ].
-	outputDirectory := log store directory.
 	
-	EpOmbuExporter askFileNameAndFileOut: entries in: outputDirectory.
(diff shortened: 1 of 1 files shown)


src/EpiceaBrowsers/EpLogBrowserPresenter.class.st
```diff
@@ -324,11 +324,10 @@ EpLogBrowserPresenter >> eventsMenuActions [
 { #category : #'menu - operations' }
 EpLogBrowserPresenter >> fileOutSelection [
 
-	| entries outputDirectory |
+	| entries |
 	entries := self selectedEntryItems collect: [ :each | each entry ].
-	outputDirectory := log store directory.
 	
-	EpOmbuExporter askFileNameAndFileOut: entries in: outputDirectory.
+	EpOmbuExporter askFileNameAndFileOut: entries in: FileLocator imageDirectory.
```
guillep: Why image directory and not working directory? In most cases it will be the same, but if the image is deployed in a separate directory, let's say read-only, working directory is better (and more user friendly IMO).

guillep (changes requested):

src/EpiceaBrowsers/EpLogBrowserPresenter.class.st
seandenigris: Done. Thanks for the explanation. I wasn't aware of the distinction.

Ducasse:
It looks good to me.

tinchodias:
nice

guillep:
thanks Sean!

PR #4441, opened 2019-08-29
[Calypso] optimize the analysis of overridden methods

dionisiydk:
Overridden state analysis is optimized by using  #allImplementorsCache:
- only when browser opens the first time on the big hierarchy there will be a little initial delay (200 millseconds on my machine) to build the cache of all implementors in the image.
- for little hierarchies the cache is not built (until user selects the Object for example)
- having the cache the browser performance is fast enough to not notice any analysis at all


dionisiydk:
And for benchmarks I use following code:
```Smalltalk
plugin := ClyNavigationEnvironment currentImage getPlugin: ClyInheritanceAnalysisEnvironmentPlugin.

[ Smalltalk allClassesDo: [ :c | c methodsDo: [ :m | plugin isMethodOverridden: m ]]] timeToRun. 

"before PR: 0:00:00:03.143" 
"with PR:  0:00:00:00.236"
```

PR solves the performance issue with opening browser of Object and other big hierarchy roots.

dionisiydk:
Seems there is an issue with Iceberg in latest image. Because some code was not committed

Issue #4161, opened 2019-08-01
Less Transcript hardcoded references


Ducasse:
(SystemNavigation default allGlobalRefsOn: #Transcript) 
	reject: [ :each | each package name = #Kernel ]

>>> 118 cases and many are not good.

Ducasse:
methodWithTranscript
	Transcript show: 'something'

testTranscriptPrinting
	self runSimulated: [ self methodWithTranscript ]

To me, this is not a real test. No explicit assertion

astares:
@VincentBlondeau @Ducasse Still relevant?

VincentBlondeau:
@astares I think they should still be some of them

Ducasse:
Yes it would be good to reduce their number. I will tag it as easy so that people are not afraid 
to start contributing.

GeraldineGalindo:
Good night, on branch #9302 I made some refactorings based on previous code reviewing PR https://github.com/pharo-project/pharo/pull/4445, in the beginning, the Transcript cases were 32, in this change are 23, I will be attentive to any observation, or suggestion, thanks!!

guillep:
Still 36 of them, some should be super easy to fix.

PR #4445, opened 2019-08-30
4161 less transcript hardcoded references

VincentBlondeau:
Related to issue: #4161.

- Deprecating Object >> *log* for *trace*
- Improve a bit some tests and printing patterns
84 references still to kill.

### DIFF
--- a/src/Athens-Examples/VGTigerDemo.class.st
+++ b/src/Athens-Examples/VGTigerDemo.class.st
@@ -2101,54 +2101,42 @@ VGTigerDemo >> initialize [
 { #category : #run }
 VGTigerDemo >> runDemo [
 	| paths extent rotation scale translation rotationSpeed time frames zoom |
-	
 	paths := self convertPathData2.
-
 	rotationSpeed := 1.
 	rotation := 0.
 	extent := surface extent.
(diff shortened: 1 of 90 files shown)


Ducasse:
Wooo! this is cool.
I will review that later but this is cool

Ducasse:
The build does not work because of bootstrap. 
I do not know how to proceed but this is a good case for me to learn. 
Could you split this issue with the change that are not in Object and related.

VincentBlondeau:
The issue is the integration of the zip changes. Reverted by  https://github.com/pharo-project/pharo/pull/4449. If you trigger the build again, it should go a bit further

Ducasse:
ouch. Thanks!!!

dionisiydk:
Hi @VincentBlondeau 
Why #trace name is better than #log?

Ducasse:
We should discuss it on the mailing-list.
I cleaned the beacon package and I would like to introduce it. 
So we should check that there is not API clash.
Denis log does not work because numbers understand it. So it can introduce super strange bugs.

VincentBlondeau:
Where should KMLog class initialization should be called from?

PR #4402, opened 2019-08-23
4401-ZipArchive-should-not-convert-FileReference-to-paths

demarey:
Always use a file reference as input for the read stream.
Fixes #4401

PR #4449, opened 2019-08-30
Revert "4401-ZipArchive-should-not-convert-FileReference-to-paths"

VincentBlondeau:
Reverts pharo-project/pharo#4402 because it should not have been integrated yet....
No review + no test passing


Ducasse:
Thanks for the revert. 
For the record I reviewed it. And I thought that I checked the tests. So were they broken?

VincentBlondeau:
Yes at bootstrap level

PR #4460, opened 2019-08-31
WeakIdentitySet

dionisiydk:
It implements WeakIdentitySet.
Tests and overrides are based on IdentitySet implementation


Ducasse:
Hi denis

WeakIdentitySetTest >> testGrowWithNil [
		"This test covers that grow take into account that nil are wrapped elements of sets"
		| set |
		set := IdentitySet new.
Why this is not 
                set := WeakIdentitySet new.

I'm tired so may be I did not understand. 

Stef

dionisiydk:
@Ducasse Thanks for review.
I improved tests. (#classToBeTested was not really used in WeakSetTest)

MarcusDenker (approved):

PR #4464, opened 2019-09-01
Improving performance of athens

tesonep:
- Athens should cache the latest used stroke paints and paints, as the conversion of Forms to Athens surfaces is not cheap. 
It requires the execution of a transformation for all the pixels in the form / surface.


Ducasse:
We have one test failing: 
     - Subclasses of Object should have #= methods in protocol comparing

tesonep:
Yes, I tried to put the comparison as extension methods. So, I moved to the "comparing" protocol

PR #4465, opened 2019-09-01
[DO NOT MERGE] Add instanceVariablesRawTypes to ClassDescription to support Live Typing

hernanwilkinson:
This is only to bootstrap a 64 image with support for Live Typing. 
I run the bootstrap locally but only generates de 32 bit version, I need the 64 bit version that I'll get from CI


tesonep:
We should not close it yet!

PR #4469, opened 2019-09-02
Fixes 2936

kasperosterbye:
The test for isSelector is not done in the test class for RBScanner, but there is an extensive suite of tests in RBConditionTest (Should I raise an issue to have them moved???). I reincluded the two outcommented tests in testCheckInvalidMethodName

My brief view on this bug is that selectors are *not* part of the smalltalk grammar, and as such there was no good mechanism and tokens to use. The current version failed for that reason, trying to get by using what was there. I added a new method (and its helper), to check for validity of selectors. It is a small state machine, see documentation in helper class.

The new scanner for isSelector: is based on [this statemachine](http://www.plantuml.com/plantuml/png/RO-n2i8m48RtUugRXRn0XzBg98wEuX2R4mlNHE8LwTkRL4Xxa2cNx_z2_ki-OgFC4mF8A4nu9QP1GeJRuOx6j7WSlOqBDlXOXy9xVhzipy5Joco-m0vfeoQuP9n2cjFp8PygRJ7_qo259m2ide8REliBBTyGNQa16p9LGUUw1C1_uThk9TBrEl9kdlW7).

### DIFF
--- a/src/AST-Core/RBScanner.class.st
+++ b/src/AST-Core/RBScanner.class.st
@@ -95,15 +95,12 @@ RBScanner class >> initializeUnderscore [
 ]
 
 { #category : #testing }
-RBScanner class >> isSelector: aSymbol [ 
-	| scanner token |
+RBScanner class >> isSelector: aSymbol [
+	| scanner |
 	scanner := self basicNew.
 	scanner on: (ReadStream on: aSymbol asString).
 	scanner step.
-	token := scanner scanAnySymbol.
-	token isLiteralToken ifFalse: [^false].
-	token value isEmpty ifTrue: [^false].
-	^scanner atEnd
+	^ scanner isSelector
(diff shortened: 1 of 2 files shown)


kasperosterbye:
It is not quite clear to me, as this is my first pharo PR, if the error is that my code is wrong, or that something else is happening,.

VincentBlondeau:
Thanks @kasperosterbye ! The fix is well explained with the schema and tests are present to check that value is correct. The CI failure is unrelated to your changes. So it should be merged soon!

The modifications on the tests are good in RBConditionTest for now, and could be moved afterward but I do not think it is need right now.

Thanks for your first contribution!

VincentBlondeau (approved):

Ducasse:
Thanks look nice to me.

PR #4473, opened 2019-09-02
4401-bis-ZipArchive-should-not-convert-FileReference-to-paths

### LINKED ISSUE #4401: ZipArchive should not convert FileReference to paths
Indeed, it prevents to use a memory store (e.g. for test purposes)

demarey:
Fixes #4401 ZipArchive should not convert FileReference to paths.


jecisc:
Failure not related

Issue #2523, opened 2019-02-13
Use NativeFileSystemBrowser instead of just NativeBrowser


astares:
As there could be different native browser types (file browser, web browser, ...)  can we please rename the new selectors like  #openInNativeBrowser into  #openInNativeFileSystemBrowser as this would be more clear.

dionisiydk:
What about OSFileBrowser? It is shorter and simpler

13 февр. 2019 г. 4:00 PM пользователь "Astares" <notifications@github.com>
написал:

As there could be different native browser types (file browser, web
browser, ...) can we please rename the new selectors like
#openInNativeBrowser into #openInNativeFileSystemBrowser as this would be
more clear.

—
You are receiving this because you are subscribed to this thread.
Reply to this email directly, view it on GitHub
<https://github.com/pharo-project/pharo/issues/2523>, or mute the thread

astares:
Why so shy on longer names?

It is not a file browser as it displays directories as well. I would prefer meaningful names - so I think "FileSystemBrowser" is more appropriate.

dionisiydk:
Directories are files too (in linux at least).
And we already use this File Browser name for our internal tool (world
menu/tools/)

13 февр. 2019 г. 4:47 PM пользователь "Astares" <notifications@github.com>
написал:

Why so shy on longer names?

It is not a file browser as it displays directories as well. I would prefer
meaningful names - so I think "FileSystemBrowser" is more appropriate.

—
You are receiving this because you commented.

Reply to this email directly, view it on GitHub
<https://github.com/pharo-project/pharo/issues/2523#issuecomment-463274522>,
or mute the thread

astares:
Still valid

hilaire:
So what is your final say:
openInOSFileBrowser or openInNativeFileSystemBrowser

astares:
Whatever fits best - but native Browser is not enough as there is also a Web browser

hilaire:
go for openInOSFileBrowser then. I will make a PR

PR #4475, opened 2019-09-02
Use-platform-visitor-in-native-browser

jecisc:
Chirstophe added platform visitor in Pharo8. This PR takes advantage of it in NativeBrowser.


pavel-krivanek (approved):
works ;-)

astares:
Can we still please rename into NativeFileBrowser?

MarcusDenker:
So what is the status? Do we merge and then rename? or is the idea that this PR will be updated? or replaced?

In general, it is much easier to merge everything that is a good step and do the next step in next PR

astares:
Side note: the renaming is handled in https://github.com/pharo-project/pharo/issues/2523

MarcusDenker:
There is now a conflict

MarcusDenker:
So this is why I asked about merging... to me it was clear that if we merge the second step before this then we would get a conflict.

jecisc:
Fixed.

jecisc:
Can we remove the Need More Work tag, please? This is working and pass the tests.

jecisc:
Ping

PR #4484, opened 2019-09-03
writeScanOn-is-dead-code

jecisc:
TextAttribute>>#writeScanOn: seems to be dead code.

I remove it because it is missleading when we try to understand how Text works.


Ducasse:
This was for the generation of color and other in code. 
We removed this support because code should only contain code. 
For comment we could discuss it but...

PR #4493, opened 2019-09-04
fix/blockprinting

MarcusDenker:
fix block printing for full blocks, with test


src/Kernel/CompiledBlock.class.st
```diff
@@ -22,6 +22,11 @@ CompiledBlock >> hasMethodReturn [
 		 self encoderClass methodReturnBytecodes includes: byte]
 ]
 
+{ #category : #'as yet unclassified' }
```
jecisc:
```suggestion
{ #category : #testing }
```

jecisc (changes requested):

PR #4494, opened 2019-09-04
Add-simple-undo-redo-without-wrapper

jecisc:
For the FileDialog of Pharo 8 I want to use an undo/redo system. I checked HistoryIterator but I find it too complicated for my needs.

In this PR I introduce a new history iterator. You configure it with an undo and redo block and then you can directly register objects you navigate into. When and undo or redo will happen, the objects will be used as parameter of the valuables to execute the action.

For example check test class and class comment of ConfigurableHistoryIterator.

### DIFF
--- a/src/System-History-Tests/ConfigurableHistoryIteratorTest.class.st
+++ b/src/System-History-Tests/ConfigurableHistoryIteratorTest.class.st
@@ -1,0 +1,146 @@
+Class {
+	#name : #ConfigurableHistoryIteratorTest,
+	#superclass : #TestCase,
+	#instVars : [
+		'historyList',
+		'saved'
+	],
+	#category : #'System-History-Tests'
+}
+
+{ #category : #running }
+ConfigurableHistoryIteratorTest >> setUp [
+	super setUp.
+	historyList := ConfigurableHistoryIterator undo: [ :integer | saved := integer ] redo: [ :integer | saved := integer asString ]
+]
+
+{ #category : #tests }
+ConfigurableHistoryIteratorTest >> testHasNext [
+	historyList register: 3.
+
+	self deny: historyList hasNext.
+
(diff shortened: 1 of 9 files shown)


Ducasse:
Ok cyril but if we do not improve existing we will have a part using x, a part y .... and duplicated. If the one in pharo is bad let us fix it.

jecisc:
In the current HistoryIterator, from what I saw, you need to register each item with redo and undo actions. 

To register it you need to create for the redo and the undo an array with 3 elements:
- The receiver of the action
- The selector to perform on it
- The arguments of the action

We end up with code like this:

```Smalltalk
addDeleteSelectionUndoRecord
	| undoText redoText |
	undoText := self selection.
	redoText := self nullText.
	self
		redoArray:	{textArea. #undoTypeIn:interval:.	{redoText.	(self selectionInterval)}}
		undoArray:	{textArea. #redoTypeIn:interval:.	{undoText.	(self selectionInterval first to: self selectionInterval first - 1)}}
```

This is needed for complex stuff when the undo/redo actions will be different and depend on the context. 

In my case, I know that for any undo or redo action I'll do the same. So I just need to store context objects. I want to use it for the previous/next in the FileDialog. With this alternative implementation I could just do:

```Smalltalk
ConfigurableHistoryIterator action: [ :fileReference | self currentDirectory: fileReference ]
```

When the user would click on an element I could do:

```
history register: aFileReference
```

And for the buttons, the actions would be `history undo` and `history redo`. 

So, the goal of this PR is to offer a simpler alternative for simple cases and to keep the more powerful version for non-simple cases.

jecisc:
I think I see ways to improve the HistoryIterator to make it simpler (I think it would be better to use blocks instead of MessageSend), but it will not resolve what this PR is resolving.

PR #4500, opened 2019-09-04
4499-Reducing-the-amount-of-hardcoded-reference-to-RBParser-Part2-

Ducasse:
fixes: 4499 
Factoring out RBParser in many places.


kasperosterbye:
Great - I know I might have gotten into problems with this one at some point in time as I sometimes override parserClass.

PR #4502, opened 2019-09-05
Fixes issue 3068

kasperosterbye:
It seems that the is a near solution already in the discussion of the issue. 
There is a method #firstComment, which should return the first comment, while #comments should return a list of them all.
I changed the method to return an ordered collection of comments in the method.

I changed the test method to depend less on the comments in other parts of the system, and noted where it does. Obviously I made the test check if two comments can be extracted as well.


MarcusDenker (approved):
good!

PR #4503, opened 2019-09-05
Validating that the FFI callout forwards objects

tesonep:
Adding a test to validate that the FFI callout primitive is forwarding the objects.


MarcusDenker:
error not related

MarcusDenker (approved):

PR #4504, opened 2019-09-05
2372-Add-scanTokens-to-RBScanner

### LINKED ISSUE #2372: Add scanTokens: to RBScanner
Something probably like that

	

[:c| c isSeparator] split: 'Object subclass: #NameOfSubclass
	instanceVariableNames: ''''
	classVariableNames: ''''
	package: ''UndefinedClasses-Experiment'''
>>> an OrderedCollection('Object' 'subclass:' '#NameOfSubclass' '' 'instanceVariableNames:' '''''' '' 'classVariableNames:' '''''' '' 'package:' '''UndefinedClasses-Experiment''')

The following does not work

(RBParser new parseLiterals: ('Object subclass: #NameOfSubclass
	instanceVariableNames: ''''
	classVariableNames: ''''
	package: ''UndefinedClasses-Experiment''' ))

(RBScanner new on: ('Object subclass: #NameOfSubclass
	instanceVariableNames: ''''
	classVariableNames: ''''
	package: ''UndefinedClasses-Experiment''' readStream))

	scanLiteral

kasperosterbye:
Fixes: #2372

I added a "scanTokens:" to RBScanner. I do not believe there should be a helper method in RBParser as the parser should focus on AST, but I am still new to all this.
The method was nearly there. I build on the #contents method of the scanner. Since the issue mentions a similarity to the #split method, I extract the token values from the scanner, and ignores the token types.
I added a test method as well, which check that white space and comments are removed.

### DIFF
--- a/src/AST-Core-Tests/RBScannerTest.class.st
+++ b/src/AST-Core-Tests/RBScannerTest.class.st
@@ -92,6 +92,20 @@ RBScannerTest >> testNextWithTwoDoubleQuotesInCommentGetError [
 		raise: SyntaxErrorNotification
 ]
 
+{ #category : #tests }
+RBScannerTest >> testScanTokens [
+	| inp exp |
+	inp := 'Object subclass: #NameOfSubclass'.
+	exp := {'Object'.
+	'subclass:'.
+	#NameOfSubclass asString}.
+	self assert: (RBScanner scanTokens: inp) equals: exp.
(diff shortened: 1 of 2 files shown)


MarcusDenker (approved):
looks good

MarcusDenker:
Yes, the scanner is about tokes, the parser about the AST. E.g. one could build a parser that does not use tokens... so separating the two concepts is good.

PR #4507, opened 2019-09-05
4506-line-end-conversion-does-not-work-for-old-ByteTextConverters

### LINKED ISSUE #4506: line end conversion does not work for old ByteTextConverters
While these have been replaced by the new ZnCharacterEncoders, it might be nice to fix (some of?) the bugs that exist in the old TextConverters, for educational reasons and/or potential backport to Pharo 6.1 (... and maybe as a cautionary tale of the dangers of mixing too many responsibilities)

The issues are:
- In installLineEndConvention:, the latin1Map is replaced by nonAsciiMap, meaning all non-ascii characters should be translated. This only works for utf8 - in the byte converters there are usually many entries with byte encoding equivalent to unicode code points. 
When the optimized nextPutByteString: searches for next index to convert, for most non-ascii, it will then encounter false hits, for which the latin1Encodings does not contain a set of bytes to translate to, because they shouldn't.
- In the "slow" fallback for WideStrings, unicodeToByte: is used. This method precedes the latin1Encoding optimalization / lineEndConvension being added to Converter responsibilities, and thus assumes each code point corresponds to at most one byte. This is not true when cr can be converted to crlf.

How line end convention is supposed to work when decoding is anyone's guess, the above applies to the encoding process.

Another minor bug:
-  installLineEndConvention uses convertFromSystemString: when the correct call would be convertToSystemString:. Luckily, cr and lf are the same in all encodings, but it highlights the importance of method comments that state what one means by "system" string ;)

Rydier:
Fixes #4506 , contains tests for line end conversion issues in the old ByteTextConverters:  .

Also updated class comment of TextConverter to better explain why it's no longer recommended for active use.

Couldn't find the "development" branch as described on 
https://github.com/pharo-project/pharo/wiki/Contribute-a-fix-to-Pharo
but saw there were pull requests being merged on Pharo8, so hopefully this is the right way...

### DIFF
--- a/src/Multilingual-TextConversion/Character.extension.st
+++ b/src/Multilingual-TextConversion/Character.extension.st
@@ -4,7 +4,7 @@ Extension { #name : #Character }
 Character >> macRomanToUnicode [
 	"Convert the receiver from MacRoman Unicode."
 
-	^MacRomanTextConverter new unicodeToByte: self
+	MacRomanTextConverter new unicodeToBytes: self do: [ :encodedChar | ^encodedChar ]
 ]
 
 { #category : #'*Multilingual-TextConversion' }
--- a/src/Multilingual-Tests/ByteTextConverterTest.class.st
+++ b/src/Multilingual-Tests/ByteTextConverterTest.class.st
@@ -36,6 +36,61 @@ ByteTextConverterTest >> testLatin2ToUnicodeConversion [
 ]
 
 { #category : #testing }
+ByteTextConverterTest >> testLineEndConversion [
+	"Regression test for line end conversion.
+	Many byte encodings have overlap in the non-ascii range (and thus 0 entries in the 128-255 range of latin1Map, meaning they shouldn't be converted.
+	Previously a full nonAsciiMap with 1 as entries for the entire 128-255 range was installed.
+	This only ever really worked for utf-8 - for byte text converters it would lead to exceptions on any non-ascii character that didn't have to be converted."
+	| cp1252Bytes internalString converter encodedBytes  |
+        
+	"æøå all map to the same code points in cp1252 as in unicode."
+	cp1252Bytes := #[16rE6 16r0D 16r0A 16rF8 16r0D 16r0A 16rE5] .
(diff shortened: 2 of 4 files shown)


CafeKrem:
your test don't succeed under window and OSX.
the linux test bash script failed 

OmFileStoreReadingError: Reading entry from File @ /builds/workspace/uest_and_branch_Pipeline_PR-4507/Tests-unix-32/pharo-local/ombu-sessions/Pharo7.0-PR-32bit-f46e34b.eb5sixbd62rsq27d9nukeeqy.ombu@3276802: Illegal leading byte for utf-8 encoding

src/Multilingual-TextConversion/Character.extension.st
```diff
@@ -4,7 +4,7 @@ Extension { #name : #Character }
 Character >> macRomanToUnicode [
 	"Convert the receiver from MacRoman Unicode."
 
-	^MacRomanTextConverter new unicodeToByte: self
+	MacRomanTextConverter new unicodeToBytes: self do: [ :encodedChar | ^encodedChar ]
```
CafeKrem, line 7: why did you use a return statement , '^encodedChar' , in lambda function use to iterate on collection ?

CafeKrem: can you change the selector because in unicodeToBytes: do: .
do: is very strange in the implementation you only iterate on a collection in one case.

Rydier: Well, if Ombu is broken in pharo8 and can’t correctly encode methods containing non-ascii as utf8, so they fail to read from disk, that’s hardly the fault of the test itself, is it?

I use a return in the block because do: return self. For macroman with no line end convension installed, it’s always 1 - 1, so returning the first element iterated is correct.

 We iterate over all bytes the Unicode maps to in all the cases, hence do:.
It’s just in two out of three cases, “all bytes” will be a single one.

Rydier:
Thanks!
I'll redo a bit and resubmit :)

PR #4511, opened 2019-09-06
Deprecation Transform: add #flag at method start

MarcusDenker:
When a deprecation transforms code, it is not clear why the method changed. This PR adds a #flag call at the method start

### DIFF
--- a/src/Kernel/Deprecation.class.st
+++ b/src/Kernel/Deprecation.class.st
@@ -228,7 +228,7 @@ Deprecation >> transformWarningNode [
 		arguments:
 			{(RBLiteralValueNode
 				value:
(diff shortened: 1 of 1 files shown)


jecisc:
Hum... Maybe we should have a setting for this.

Personally, I love the automatic transformations but I would not like to commit methods with such a flag. And it will make me lose a lot of time to remove all the flag each time.

jecisc:
Maybe an alternative would be to have a browser recording all automatic transformations in the current session and to open it (if it's not already opened) when a transformation happens.

Ducasse:
Yes I agree with Cyril.

MarcusDenker:
Ok

PR #4512, opened 2019-09-06
Update licence

juliendelplanque:
To set "Pharo Contributors" in copyright.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,21 +1,72 @@
-The MIT License (MIT)
+Licensed under the MIT License with parts under the Apache License.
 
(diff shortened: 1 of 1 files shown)


MarcusDenker:
We should make sure that it is in sync with the in-image version

juliendelplanque:
Sure, one step at a time, I just fixed the easy part bi GitHub web app.

juliendelplanque:
This should be sync during the bootstrap from source ideally by the way.

juliendelplanque:
Hum, the licence in image seems to be more complete, I will copy it here.

MarcusDenker (approved):

PR #4532, opened 2019-09-09
Some additional tests about Continuation objects

massimo-nocentini:
The following commits add more test cases about the use of Continuation objects.

According to the book [The Seasoned Schemer](https://mitpress.mit.edu/books/seasoned-schemer-second-edition), we implemented the functions that Friedman and Felleisen used for teaching continuations, porting them from Scheme to Smalltalk.


Ducasse:
This is nice to have more tests on Continuation. 
- Now we are trying to remove methods from Object so can you package ~~> as an extension of the package of continuation (it will be helpful with the following step). 
- Could you improve the comment of the method ~~>
- What would be great is to create a separate package for Continuation and related but this should be for another PR.

massimo-nocentini:
As requested, the last commit provides a comment for Object>>#~~>. Sorry for the typo in the commit message.

Can I also classify it as an extension of the package 'Kernel-Methods' where the Continuation class lies?

Ducasse:
Ok if you can at least package it with the package it will help.

massimo-nocentini:
Before classifying it I would like to propose to leave the message `#~~>` in the Object class because it allows objects to be "promoted" in a chain, as the message `#->` promotes them in key-value pairs.

I know that `ValueLink` objects are raw building blocks for lists and usually they are not manipulated directly but through the enclosing container; however having the possibility to create a raw chain I think is a sufficient reason to leave `#~~>` there (it should be a base for introducing Lisp-like `Cons` cells also).

Let me know if you prefer to have it as an extension anyway and I'll do.

Ducasse:
I think that Value is so central that we can have it in Object but I will let others comment too.

MarcusDenker:
removed stale bot. Will have a look

pavel-krivanek:
warning, this PR is against Pharo 7

estebanlm:
let's move this to P9 :)

estebanlm:
... and of course it will need some conflict resolution :( 
I wonder why adding some tests causes conflicts, but who knows.

estebanlm:
mmm, @massimo-nocentini can you provide a PR pointing to Pharo9.0 ?
Thanks!

estebanlm:
(or to Pharo8.0, I will move it to Pharo9.0 once provided)

massimo-nocentini:
Replaced by https://github.com/pharo-project/pharo/pull/5491.

PR #4534, opened 2019-09-09
4533-Provide-a-trimAlltrimAll-to-complete-the-trim-messages

### LINKED ISSUE #4533: Provide a trimAll/trimAll: to complete the trim messages
We have
 - trimLeft
 - trimRight
 - trimBoth

but no way to replace overall separators or character provided in a block expression.

But this is something one often needs in parsing or data reading.

We should therefore provide additionally
 - trimAll
 - trimAll: aBlock

astares:
Provide a trimAll/trimAll: to complete the trim messages

Fix #4533

### DIFF
--- a/src/Collections-Strings/String.class.st
+++ b/src/Collections-Strings/String.class.st
@@ -2430,6 +2430,24 @@ String >> trim [
 ]
 
 { #category : #copying }
+String >> trimAll [
+	"Trim all separators outside and within the receiving string."
+	
+	"   ' Hello  World ' trimAll >>> 'HelloWorld'    "
+	
+	^ self trimAll: [:each | each isSeparator ] 
+]
+
+{ #category : #copying }
+String >> trimAll: aBlock [
+	"Trim all characters satisfying the condition given in aBlock."
+
+	^self select: [ :each | 
+		(aBlock value: each) not ]
+	 
+]
+
+{ #category : #copying }
 String >> trimBoth [
 	"Trim separators from both sides of the receiving string."
 	
--- a/src/Collections-Tests/StringTest.class.st
+++ b/src/Collections-Tests/StringTest.class.st
@@ -1833,6 +1833,22 @@ StringTest >> testTranslateToUppercase [
 ]
 
 { #category : #'tests - converting' }
+StringTest >> testTrimAll [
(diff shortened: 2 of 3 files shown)


astares:
Test failure seems unrelated

juliendelplanque (commented):
I was checking how `#trimmed` is implemented and, there is actually `#trimBoth` and `#trimBoth:` messages that implement what you propose in this PR.

So I think it is not needed.
But need to be documented more. :-)

astares:
@juliendelplanque  No. Just have a closer look.
#trimmed and #trimmedBoth are the same - combining the trimming left and right.

But #trimAll removes all separators - also the ones in the middle as it is stated in its comment and test. 

```Smalltalk
' He     l l O Wordl' trimBoth.  "'He     l l O Wordl'"
' He     l l O Wordl' trimmed.   "'He     l l O Wordl'"
' He     l l O Wordl' trimAll    "'HellOWordl'"
```

noha:
But trimming is usually left and/or right according to https://en.wikipedia.org/wiki/Trimming_(computer_programming). In every more complex case I would rather use a variant of #copyWithout: that utilizes a block. Your example can be done using 

`' Hello World !' copyWithoutAll: Character separators `

<blockquote><div><strong><a href="https://en.wikipedia.org/wiki/Trimming_(computer_programming)">Trimming (computer programming) - Wikipedia</a></strong></div></blockquote>

juliendelplanque:
Then, what is the difference with `#reject:` ?

juliendelplanque:
I agree with @noha btw, to me trimming should only touch left and/or right side of the string not its middle.

noha:
Of course #reject: is the block variant

astares:
Ok, we close it

https://twitter.com/awbjs/status/1172273315727101952?s=20

<blockquote><div>Twitter</div><div><strong><a href="https://twitter.com/awbjs/status/1172273315727101952">Allen Wirfs-Brock on Twitter</a></strong></div><div>“Reminded today that trimStart and trimEnd are in ES2019—a little later doing some organizing of my long term storage I found this. I guess somethings just keep recurring forever. (Looks like a bug in the comment on example line 4, something else that never changes)”</div></blockquote>

PR #4536, opened 2019-09-09
4535-Is-there-a-reason-new-completion-does-not-use-RBProgramNodeVisitor

### LINKED ISSUE #4535: Is there a reason new completion does not use RBProgramNodeVisitor?
The new completion has a visitor inheriting from Object. 

Is there a reason not to inherit from RBProgramNodeVisitor? Because currently, it seems to duplicate its logic.

jecisc:
Make CompletionProducer inherit from RBProgramNodeVisitor + rename it CompletionProducerVisitor

Fixes #4535

### DIFF
--- a/src/NECompletion/CompletionModel.class.st
+++ b/src/NECompletion/CompletionModel.class.st
@@ -91,7 +91,7 @@ CompletionModel >> hasMessage [
 { #category : #accessing }
 CompletionModel >> initEntries [
 	| producer suggestionsList |
-	producer := CompletionProducer new.
+	producer := CompletionProducerVisitor new.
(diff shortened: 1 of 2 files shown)


jecisc:
I also add the missing method #visitSlotInitializationNode:. I returned `#()`, maybe this is not right but I don't know what it should return.

MarcusDenker (approved):
ok!

MarcusDenker:
I tested the PR, works

Ducasse:
Hi cyril
For now we are still in flux with the design of certain aspects of the class definition. 
I will adapt it and revisit everything once we know for slots and also based on the client needs.

PR #4543, opened 2019-09-10
4542-update-rfEnsure

### LINKED ISSUE #4542: update rfEnsure:
rfEnsure: is there to avoid meta recursion links when installing meta links on all methods.

ensure: was updated for full block closures, we need to have rfEnsure:  to be in sync.

MarcusDenker:
update #rfEnsure: fixes #4542


MarcusDenker:
errors not related

PR #4553, opened 2019-09-11
4550-Remove-class-refactoring-should-not-warn-user-if-the-class-reference-itself

### LINKED ISSUE #4550: Remove class refactoring should not warn user if the class reference itself
Currently, if a class reference itself and you try to remove it, you get a warning saying the class is referenced. 

In case the class reference itself, it should not warn the user.

MarcusDenker:
fix #4550 

fix is  to exclude both class and instance side


jecisc (approved):
Test failure not related!

PR #4562, opened 2019-09-11
4561-Ring-definitions-with-slots-use-category

### LINKED ISSUE #4561: Ring definitions with slots use category:
The Ring needed to follow the buggy definitions of classes where, for classes with slots, the `category:` keyword was still used. It is fixed in Pharo so Ring has to adopt it

pavel-krivanek:
fixes #4561
package: keyword in Ring definitions


jecisc (approved):
Great!

MarcusDenker (approved):

PR #4564, opened 2019-09-11
4563-reflectivity-update-rfValue

### LINKED ISSUE #4563: reflectivity: update #rfValue
-> update rfValue for full blocks

MarcusDenker:
just add a copy with methods where metalinks are turned off. they are used in the code path that activates links, thus having the copy allows us to annotate the orginals without running into strange loops.

fixes #4563


MarcusDenker:
failing tests not related

PR #4566, opened 2019-09-11
2165-GrowlMorphdefaultBorderColor-is-not-take-into-account

### LINKED ISSUE #2165: GrowlMorph>>defaultBorderColor is not take into account
I think this method is not taken into account (so the theme cannot customize the color) because we explicitly set the border style in the initialize.

hilaire:
Fix #2165
GrowlMorph>>defaultBorderColor is not take into account


Ducasse:
Thanks Hilaire.

PR #4580, opened 2019-09-13
2404-Add-rule-and-release-test-to-not-use-Smalltalk-ui-theme-in-subclasses-of-Morph

### LINKED ISSUE #2404: Add rule and release test to not use `Smalltalk ui theme` in subclasses of Morph
Morph implements in the instance and the class side a method #theme. Thus, its subclasses should never call `Smalltalk ui theme`. 

We should correct current usage, add a rule and a release test about that.

hilaire:
Fixes #2404
Add rule  to not use `Smalltalk ui theme` in subclasses of Morph and Model.
Correct the misusages.

### DIFF
--- a/src/BlueInk-Extras/BISettingPreviewer.class.st
+++ b/src/BlueInk-Extras/BISettingPreviewer.class.st
@@ -47,10 +47,10 @@ BISettingPreviewer class >> defaultSpec [
 
 { #category : #settings }
 BISettingPreviewer class >> dialogOpenBIPreviewer [
-	^ Smalltalk ui theme
+	^ self theme
 		newRowIn: self currentWorld
 		for:
-			{(Smalltalk ui theme buttonLabelForText: 'Open Blue Ink Setting Previewer ').
+			{(self theme buttonLabelForText: 'Open Blue Ink Setting Previewer ').
 			self buildOpenBIInspectorButton}
 ]
 
@@ -169,7 +169,7 @@ BISettingPreviewer >> displayNodeFor: aNode [
 		addAllMorphs:
 			{nodeMorphLeft.
 			nodeMorphRight}; 
-		color: Smalltalk ui theme lightBackgroundColor;
+		color: self theme lightBackgroundColor;
 		hResizing: #shrinkWrap;
 		yourself
 ]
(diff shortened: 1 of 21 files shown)


Ducasse:
Pay attention hilaire you commit contains also your old changes (MemoryStream....

hilaire:
Ok. Somethings I did'nt understand because I carefully check the code to commit.

hilaire:
Any ideas what's wrong when I push the pull request?

Ducasse:
May be it show you the diff between the image and your last pushed commit. 
Hence you do not see it. 
What I do is the following

I have a bottom commit (the one that I create just after downloading an image)
and I checkout this branch (so that I go back undoing all my previous changes)
and only then I start to work by creating a new branch for a bug entry 
when I'm done (and I commit the bug and push) I got back to this bottom commit. 

Like that I can produce multiple independant changes.

hilaire:
Ok. I will  try to understand because I am confused and the doc does not align with I see on the UI.

jecisc:
The flow to do multiples fixes is:
- Checkout Pharo8.0
- Create a branch
- Commit, push and do a PR
- Checkout Pharo8.0 <== This is the missing step, you need to go back to the Pharo 8.0 branch before starting a new one, else the new one will have the commits of the previous branch as ancestor
- Create a branch
- Commit, push and do a PR
- Once again

hilaire:
Questions:
**Checkout Pharo8.0**  From where? Guess: Checkout branch menu entries. Then pharo-project repo? I tried, it announce Pharo8.0 already exist. So wrong guess.

**Repair** The doc mentions repair operation. I don't see it in my image menu. It is likely out of sync though. My iceberg menu is different to the one in the doc

Capture

Ducasse:
If you checkout Pharo 80 you will lose your commit. 
First you should commit them. So that after you can checkout a branch (and it will revert your changes) and then you can reload or not your changes.

hilaire:
No change to commit.
CO of P8 branch does not work

Capture2
Capture

Ducasse:
I do not know. If you know we can do a screen sharing session so that I show you what I'm doing.

hilaire:
Why not, I can start from a fresh image but I need a hand to recover my fixes. When I look at Iceberg I have no clue how to do that.But let continue on the ml.

hilaire:
Guys I have a long commit for this issue... I guess you can repoen the issue to let it be inserted...

PR #4585, opened 2019-09-13
4583-digitLength-should-be-renamed

### LINKED ISSUE #4583: digitLength should be renamed
Guess, what is the result of `333 digitLength`? It is `2`, of course ;-) You are supposed to use decimalDigitLength
We should rename this method to something less confusing, e.g. bytesCount.
While it is so confusing, it would be better to remove it without deprecation but it is used quite a lot so we shall use auto-deprecation

pavel-krivanek:
fixes #4583
rename Integer messages:
digitLength -> bytesLength
digitAt: -> bytesAt:
digitAt:put: -> bytesAt:put:
digitCompare: -> bytesCompare:


src/Deprecated80/Integer.extension.st
```diff
@@ -0,0 +1,46 @@
...
+		deprecated: 'Please use #byteAt:put: instead'
+		transformWith:
+			'`@receiver digitAt: `@statements1 put: `@statements2'
+		   -> '`@receiver byteAt: `@statements1 put: `@statements2'.
+
+	^ self byteAt: index put: value 
+]
+
+{ #category : #'*Deprecated80' }
+Integer >> digitCompare: arg [
```
dionisiydk, line 28: I did not know about this method. It seems it is same as #threeWayCompareTo: which was introduced for sorting. So I think we just need to use #threeWayCompareTo and do not keep a new one. But please alter the implementation to use optimized version from #digitCompare:.

dionisiydk: And It will move threeWayCompare to kernel

pavel-krivanek: The main difference is that it calls a primitive by default ;-)

dionisiydk: Yes. I mean let's change threeWayCompare to use digitCompare code. It will work faster

Ducasse:
Oops there is a conflict.

src/Deprecated80/Integer.extension.st
pavel-krivanek: I opened an issue for that: https://github.com/pharo-project/pharo/issues/4605

pavel-krivanek:
only usual osx-32 failure now

PR #4586, opened 2019-09-13
Removal of Smalltalk globals in Gofer-Tests

Alisu:
Add an environment instance variable initialized to Smalltalk global in setUp, setter/getter.

### DIFF
--- a/src/Gofer-Tests/GoferOperationTest.class.st
+++ b/src/Gofer-Tests/GoferOperationTest.class.st
@@ -4,6 +4,9 @@ SUnit tests for Gofer operations
 Class {
 	#name : #GoferOperationTest,
 	#superclass : #GoferTest,
+	#instVars : [
+		'environment'
+	],
 	#category : #'Gofer-Tests-Tests'
 }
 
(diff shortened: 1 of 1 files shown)


src/Gofer-Tests/GoferOperationTest.class.st
```diff
@@ -37,9 +50,15 @@ GoferOperationTest >> hasVersion: aString [
 { #category : #running }
 GoferOperationTest >> setUp [
 	super setUp.
+	self setUpEnvironment.
 	gofer repository: self monticelloRepository
 ]
 
+{ #category : #running }
+GoferOperationTest >> setUpEnvironment [
+	environment:= Smalltalk globals.
```
jecisc:
```suggestion
	environment := self class environment.
```

jecisc: Can't we inline this if it is call only once in initialize?

jecisc (changes requested):

jecisc (approved):

PR #4594, opened 2019-09-13
2845-Error-in-CollectionflatCollect-b-as-String

### LINKED ISSUE #2845: Error in Collection>>flatCollect: b as: String
From FogBugz: https://pharo.fogbugz.com/f/cases/21690/Error-in-Collection-flatCollect-b-as-String

This code seems to be misbehaving in Pharo 6.1:

#(foo bar baz) flatCollect: #yourself as: String

I had expected the result to be 'foobarbaz' but I see an error:

ShouldNotImplement: #add: should not have been implemented in ByteString

kasperosterbye:
Fixes #2845

The reason for the error  is that the method `#flatCollect:as:` inserts elements into the collection type given as second argument. String is immutable, so that is not possible.

### DIFF
--- a/src/Collections-Abstract/Collection.class.st
+++ b/src/Collections-Abstract/Collection.class.st
@@ -695,11 +695,11 @@ Collection >> flatCollect: aBlock as: aCollectionClass [
 	"Evaluate aBlock for each of the receiver's elements and answer the
 	list of all resulting values flatten one level. Assumes that aBlock returns some kind
 	of collection for each element. Equivalent to the lisp's mapcan"
+
 	| col |
-	col := aCollectionClass new: self size. 	
-	self do: [ :each |
-		col addAll: (aBlock value: each) ].
-	^col
+	col := OrderedCollection new: self size.
+	self do: [ :each | col addAll: (aBlock value: each) ].
+	^ aCollectionClass withAll: col 
 ]
 
 { #category : #enumerating }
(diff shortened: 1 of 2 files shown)


kasperosterbye:
Fixes #2845 Error in Collection>>flatCollect: b as: String

Analysis:
The same error is also possible to provoke using:
'#(#(1 2) #(4 3)) flatCollect: #yourself as: Array'

The reason is that the method `#flatCollect:as:` inserts elements into the collection type given as second argument. String and Array are both immutable, so that is not possible. 

The solution is to flatten the elements into a temporary OrderedCollection, and then return an instance of second argument build from that temporary OrderedCollection.

I am a bit uncertain if there are any unforseen effects of using OrderedCollection, so I submit this PR to see how the tests go.

Collection >> flatCollect: aBlock as: aCollectionClass
	"Evaluate aBlock for each of the receiver's elements and answer the
	list of all resulting values flatten one level. Assumes that aBlock returns some kind
	of collection for each element. Equivalent to the lisp's mapcan"

	| col |
	col := OrderedCollection new: self size.
	self do: [ :each | col addAll: (aBlock value: each) ].
	^ aCollectionClass withAll: col

MarcusDenker (approved):

MarcusDenker:
from the perspective of the tests: it seems to not introduce any problems

PR #4596, opened 2019-09-14
Revert "4537-Integrate-MiniBeacon"

Ducasse:
Reverts pharo-project/pharo#4588


Ducasse:
I really do not understand why the baselineOfBaseLibraries was not loaded and executed.

tesonep:
It has worked

PR #4600, opened 2019-09-15
2404-Morph-Theme

### LINKED ISSUE #2404: Add rule and release test to not use `Smalltalk ui theme` in subclasses of Morph

hilaire:
Fixes #2404


Ducasse:
Ok continue and after we will integrate it :)

jecisc (approved):

Ducasse:
Hello cyril we are working with hilaire and he will continue to commit.

Ducasse:
Thanks Hilaire this is supercool.

Issue #2335, opened 2019-01-25
MemoryFileWriteStream did not understand #next:putAll:startingAt:


gcotelli:
Trying to use the memory file system to write a binary stream I got a DNU.
This only happen when you have a huge ByteArray, I don' know exactly the size triggering the problem but it's reproducible using:
```smalltalk
| m serializer bytes |

serializer := FLSerializer newDefault.
serializer analyzer when: [ :object | object class isObsolete ] substituteBy: [ :object | nil ].
serializer
	addPostMaterializationAction: [ :materialization | 
		Smalltalk tools debugger
			openOn: ( Processor activeProcess newDebugSessionNamed: 'External stack' startedAt: materialization root )
				withFullView: true
	].

bytes := ByteArray
		streamContents:
			[ :binaryStream | serializer serialize: thisContext on: binaryStream ].
		
m := FileSystem memory.
m createDirectory: 'logs'.
(m / 'logs' / 'stack.fuel')
	binaryWriteStreamDo: [ :stream | stream nextPutAll: bytes].
```
System Report:
```
Image
-----
/home/gcotelli/Pharo/Willow/Pharo.image
Pharo7.0.1
Build information: Pharo-7.0.1+build.142.sha.b8825954d0dcff883bf4f47ea5c516c9cfc2fe4b (64 Bit)
Unnamed

Virtual Machine
---------------
/home/gcotelli/Pharo/Willow/pharo-vm/lib/pharo/5.0-201901051900/pharo
CoInterpreter VMMaker.oscog-eem.2504 uuid: a00b0fad-c04c-47a6-8a11-5dbff110ac11 Jan  5 2019
StackToRegisterMappingCogit VMMaker.oscog-eem.2504 uuid: a00b0fad-c04c-47a6-8a11-5dbff110ac11 Jan  5 2019
VM: 201901051900 https://github.com/OpenSmalltalk/opensmalltalk-vm.git Date: Sat Jan 5 20:00:11 2019 CommitHash: 7a3c6b64 Plugins: 201901051900 https://github.com/OpenSmalltalk/opensmalltalk-vm.git

Unix built on Jan  5 2019 19:12:23 Compiler: 4.8.4
[...]

PR #4601, opened 2019-09-15
2335-MemoryFileWriteStream

hilaire:
Fixes for #2335-MemoryFileWriteStream-did-not-understand-nextputAllstartingAt
Hopefully without unwished commits.

### DIFF
--- a/src/FileSystem-Memory/MemoryFileWriteStream.class.st
+++ b/src/FileSystem-Memory/MemoryFileWriteStream.class.st
@@ -3,7 +3,7 @@ A file write stream - but within memory
 "
 Class {
 	#name : #MemoryFileWriteStream,
-	#superclass : #Object,
+	#superclass : #Stream,
 	#instVars : [
 		'file',
 		'stream'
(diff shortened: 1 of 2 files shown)


hilaire:
Waiting for review

Ducasse:
I do not understand why next is cancelled. Even in memory we can move one by one, no?

hilaire:
Well, I did not look at checking the behavior of next because at the first place MemoryFileWriteStream was a subclass of an Object and next was not implemented. Making it a subclass of Stream fix the error, but I can't say then if the inherited next is meaningful or working. So I prefere to stick it as shouldnotiomplement.
What should we do?

svenvc:
I also do not understand: this looks totally wrong to me.

I've said this before, a simple WriteStream on a ByteArray is a perfect 'memory file write stream'. Why do we even need something else ? I would like to understand that first.

hilaire:
So is the idea more like MemoryFileWriteStream is not needed and should be deprecated?

estebanlm:
Mmm. I guess @hilaire (and @svenvc) are right, the uses of `MemoryFileWriteStream` can be done as a `WriteStream` on a `ByteArray`.

PR #4621, opened 2019-09-16
4620-Simplify-FloatArray-handling

### LINKED ISSUE #4620: Simplify FloatArray handling
- we should use #withIndexDo: in #asFloatArray and #dot: 
- write a test case for dot product
- use assert:equals: in #testArithmeticCoercion

astares:
Simplify FloatArray handling

Fix #4620

### DIFF
--- a/src/Collections-Abstract/SequenceableCollection.class.st
+++ b/src/Collections-Abstract/SequenceableCollection.class.st
@@ -228,7 +228,7 @@ SequenceableCollection >> asFloatArray [
 
 	| floatArray |
 	floatArray := FloatArray new: self size.
-	1 to: self size do:[:i| floatArray at: i put: (self at: i) asFloat ].
+	self withIndexDo: [:el :i| floatArray at: i put: el asFloat ].
 	^floatArray
 ]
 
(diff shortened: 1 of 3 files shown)


src/Collections-Abstract/SequenceableCollection.class.st
```diff
@@ -228,7 +228,7 @@ SequenceableCollection >> asFloatArray [
 
 	| floatArray |
 	floatArray := FloatArray new: self size.
-	1 to: self size do:[:i| floatArray at: i put: (self at: i) asFloat ].
+	self withIndexDo: [:el :i| floatArray at: i put: el asFloat ].
```
jecisc, line 231: I am not sure about this change because using #to:do: is faster. So it might be used for performance reasons.

estebanlm:
@jecisc is right. `FloatArray` is general used to perform some math computations that require speed, and I guess `to:do:` there has sense. 
... but, in other sense, having the test is cool. 
What can we do @astares ?

estebanlm:
In any case, this will be for Pharo9.0

Ducasse:
I opened https://github.com/pharo-project/pharo/issues/5704 and we can close this issue.

PR #4630, opened 2019-09-16
4629-ClassDefinitionParser-should-allow-global-message-node-receivers-for-slots

### LINKED ISSUE #4629: ClassDefinitionParser should allow global message node receivers for slots
When a slot is given not using the current syntax:

```Smalltalk
  { #slotName => MySlot parameter: 2  }
```
but just using a class side instance creation method like:

```Smalltalk
   { MySlot named: #'slotName' parameter: 2 }
```

as described in https://lists.pharo.org/pipermail/pharo-dev_lists.pharo.org/2019-September/276248.html

then the class definition sends an invalid message and the context menu in Calypso
is not shown anymore.

image

astares:
ClassDefinitionParser should allow global message node receivers for slots

fix #4629

### DIFF
--- a/src/ClassParser/CDClassDefinitionParser.class.st
+++ b/src/ClassParser/CDClassDefinitionParser.class.st
@@ -208,6 +208,10 @@ CDClassDefinitionParser >> parseSlotNode: aRBMessageNode [
 			classDefinition addSlot: slot.
 			^ self ].
 
+	"When a slot is defined by just using a slot class (without => syntax) we just return"
+	(aRBMessageNode receiver isGlobal) ifTrue: [
+		 ^self ].
+
 	"when a slot is just 'inst' => InstanceVariableSlot default: 5."
 	aRBMessageNode receiver selector = '=>'
 		ifTrue: [ "If we are here we did not recognize the slot syntax"


dionisiydk:
We need a test

astares:
As you see from the discussion on the ML this is not yet decided which path Pharo will go with slots in the class definition. First Marcus will work on the slot compositions and then we will see which way we go with the parser.

Currently the CDCParser is unfinished but already integrated. It leaves  the context menu broken when a slot object definition is done using a class message - which makes it unusable to work with Calypso. 

For now this PR just avoids the debug context menu - so yes it is a workaround for the time being.
A test makes sense after having a full syntax decision.

dionisiydk:
Hi @astares .

I disagree with you. Whatever syntax we will choose the class parser should not fail on such expressions.

Currently the only user of class parser is "Calypso commands machinery". But idea is to use it for class creation. CMD+s on class creation tab in the browser will parse definition written by user and it should not fail on incorrect syntax.

So I think we need the test covering slot definition part starting with global name

astares:
@dionisiydk 

As I said: I agree that we need a test - but this PR should go in now as currently the context menu is broken in latest image and Calypso unusable (or even write a sample to provide a test)

Opened a separate issue for the test addition: https://github.com/pharo-project/pharo/issues/4636

PR #4631, opened 2019-09-16
2523-Use-NativeFileSystemBrowser-instead-of-just-NativeBrowser

### LINKED ISSUE #2523: Use NativeFileSystemBrowser instead of just NativeBrowser

hilaire:
Fixes #2523 
Use NativeFileSystemBrowser instead of just NativeBrowser

### DIFF
--- a/src/Deprecated80/OSPlatform.extension.st
+++ b/src/Deprecated80/OSPlatform.extension.st
@@ -1,0 +1,8 @@
+Extension { #name : #OSPlatform }
+
+{ #category : #'*Deprecated80' }
+OSPlatform >> openNativeBrowserOn: aFileReference [ 
+	self deprecated: 'Use message openOSFileBrowserOn:'.
+	^ self openOSFileBrowserOn: aFileReference
+
+]
--- a/src/Deprecated80/OSPlatform.extension.st
+++ b/src/Deprecated80/OSPlatform.extension.st
@@ -2,7 +2,7 @@ Extension { #name : #OSPlatform }
 
 { #category : #'*Deprecated80' }
 OSPlatform >> openNativeBrowserOn: aFileReference [ 
-	self deprecated: 'Use message openOSFileBrowserOn:'.
+	self deprecated: 'Use message openOSFileBrowserOn:' transformWith: '`@receiver openNativeBrowserOn: `@statements1' -> '`@receiver v `@statements1'.
 	^ self openOSFileBrowserOn: aFileReference
 
 ]
--- a/src/Native-Browser/AbstractFileReference.extension.st
+++ b/src/Native-Browser/AbstractFileReference.extension.st
@@ -1,6 +1,13 @@
 Extension { #name : #AbstractFileReference }
 
 { #category : #'*Native-Browser' }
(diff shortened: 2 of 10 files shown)


astares (approved):

jecisc (changes requested):
This code is used in some projects. It would be better to use deprecation with transformation rules to not break them.

astares:
@jecisc 
Do we really need the deprecation? To my knowledge the #openNativeBrowserOn was introduced in P8 and not earlier. 

Because it would mean to deprecate within a release iteration (P8 in this case) methods that were introduced within the same iteration...

jecisc:
@astares It is used in Iceberg so without deprecation, it will break.

astares:
@jecisc OK 

@hilaire can you please change the PR to
 - add a deprecation messages to #openNativeBrowserOn: and #openInNativeBrowser
    (or ideally a transformation rule so callers are changed automatically)
 - dispatch in the method to the new #openOSFileBrowserOn and #openInOSFileBrowser
 - move all the old methods to Deprecated80 package

src/Deprecated80/OSPlatform.extension.st
```diff
@@ -0,0 +1,8 @@
+Extension { #name : #OSPlatform }
+
+{ #category : #'*Deprecated80' }
+OSPlatform >> openNativeBrowserOn: aFileReference [ 
+	self deprecated: 'Use message openOSFileBrowserOn:'.
```
jecisc:
```suggestion
	self deprecated: 'Use message openOSFileBrowserOn:' transformWith: '`@receiver openNativeBrowserOn: `@statements1' -> '`@receiver v `@statements1'.
```

src/Native-Browser/AbstractFileReference.extension.st
```diff
@@ -2,5 +2,12 @@ Extension { #name : #AbstractFileReference }
 
 { #category : #'*Native-Browser' }
 AbstractFileReference >> openInNativeBrowser [
-	^ OSPlatform current openNativeBrowserOn: self
+	self deprecated: 'Use message openInOSFileBrowser'.
```
jecisc:
```suggestion
	self deprecated: 'Use message openOSFileBrowserOn:' transformWith: '`@receiver openNativeBrowserOn: `@statements1' -> '`@receiver v `@statements1'.
```

jecisc (changes requested):
I did suggestions to use rewrite rules for deprecation. They can be accepted directly via Github's web application.

src/Deprecated80/OSPlatform.extension.st
hilaire: Great. I did not know about transformation rule.

hilaire:
I accepted your suggestions. Are we done with it?

MarcusDenker:
errors not related

MarcusDenker (approved):

PR #4642, opened 2019-09-18
4641-OCASTClosureAnalyzer-does-not-update-binding-of-temp-var-definitions

### LINKED ISSUE #4641: OCASTClosureAnalyzer does not update binding of temp var definitions
the binding of temp var definitions is not updated, it should in this case be an instance of
OCVectorTempVariable

ast := Smalltalk compiler parse: 
'
m04_two_temps_one_temp_in_remote_vector

| t1 t2 |

t1 := 1.
[ t2 := t1 +  1.0 ] value.

^t2'.
ast doSemanticAnalysis.
var := ast temporaries second binding class

MarcusDenker:
factor out binding re-lookup and fix as #lookupAndFixBinding: and call it at the end of blocks and method on the temp defintion nodes

fixes #4641

### DIFF
--- a/src/OpalCompiler-Core/OCASTClosureAnalyzer.class.st
+++ b/src/OpalCompiler-Core/OCASTClosureAnalyzer.class.st
@@ -25,6 +25,14 @@ OCASTClosureAnalyzer >> doRemotes [
 	
 ]
 
+{ #category : #variables }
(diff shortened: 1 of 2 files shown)


dionisiydk (commented):
What about a test?

MarcusDenker:
error not related

PR #4645, opened 2019-09-18
2174-Add-withoutDuplicates-in-Collections

### LINKED ISSUE #2174: Add #withoutDuplicates in Collections?
Sometimes I need to remove the duplicated entries of a collection, but I want to keep the same kind of collection, so #asSet is not what I need. Currently, I have things like `anArray asSet asArray` in my code. 

Maybe we could have a method #removeDuplicates in Collections? Or do we already have something like that?

hilaire:
Fixes #2174  
A first naive implementation with tests.

### DIFF
--- a/src/Collections-Abstract/Collection.class.st
+++ b/src/Collections-Abstract/Collection.class.st
@@ -466,6 +466,15 @@ Collection >> copyWithoutAll: aCollection [
 	^ self reject: [:each | aCollection includes: each]
 ]
 
+{ #category : #copying }
+Collection >> copyWithoutDuplicates [
+	"Answer a copy of the receiver twithout any duplicated elements"
+	"#(2 3 4 4  5 6) copyWithoutDuplicates  >>> #(2 3 4 5 6)"
+	"#('do' 'la' 'si' 'do' 'la') copyWithoutDuplicates >>> #('la' 'do' 'si')"
+	"#(#do #la #si #do #la) copyWithoutDuplicates >>> #(#la #do #si)"
+	^ self asSet asArray
+]
+
 { #category : #enumerating }
 Collection >> count: aBlock [ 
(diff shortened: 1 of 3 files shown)


Ducasse:
"Answer a copy of the receiver twithout any duplicated elements
-> 
"Answer a copy of the receiver without any duplicated elements

and we should resolve the conflict

hilaire:
I can't fetch the branch anymore :-( Seem to be too distant from my working image.. The image stall
Log from the terminal
```
Instance of OCASTClosureAnalyzer did not understand #lookupAndFixBinding:
OCASTClosureAnalyzer(Object)>>doesNotUnderstand: #lookupAndFixBinding:
OCASTClosureAnalyzer>>visitVariableNode:
RBTemporaryNode(RBVariableNode)>>acceptVisitor:
OCASTClosureAnalyzer(RBProgramNodeVisitor)>>visitNode:
OCASTClosureAnalyzer(RBProgramNodeVisitor)>>visitAssignmentNode:
RBAssignmentNode>>acceptVisitor:
OCASTClosureAnalyzer(RBProgramNodeVisitor)>>visitNode:
```

Ducasse:
This looks like an error introduced by marcus cleaning....
I think that this is good so that marcus can fix it. It is not related to iceberg.

MarcusDenker:
I think that you can not merge the change I did back then in a live image... in general, we do not support updating the image by loading the Iceberg diff... it works for small changes, but not for any change where the order is important or where one needs to transform existing objects.

hilaire:
@MarcusDenker which are my options to work again on that specific branch?

MarcusDenker:
I always keep the image that I used to create a PR till it gets integrated... not nice but it works.

hilaire:
Okay it makes sense. 
So there are no option but hand recovering with the diff?

PR #4651, opened 2019-09-19
2607-reset-methods-should-be-in-initialization-protocol-and-a-test-should-ensure-we-do-not-have-regression

hilaire:
Merge 987c7671182b5323c2ce7ef8533fa0cab69cd872


hilaire:
Is this pull request ok? Because Pharo repeatedly crashed so I have to install a new one, the repo was out of sync. 
While I resync some unwished commit was inserted, I try do delete from github this commit (EpLogBrowserPresenter.class.st) and I need to know if it is fine.
Is the "Delete EpLogBrowserPresenter.class.st" commit clearly a diff delete and not a file delete from repo?
Thanks

MarcusDenker:
the PR deletes class EpLogBrowserPresenter. Which is strange, as if this should be done it would be better done it it's own PR.

hilaire:
I don't want to delete this file but delete its diff. When I pushed, Pharo did a pull where this diff on EoLogBrowserPresenter shows up.

MarcusDenker:
it might be that EpLogBrowserPresenter calls a deprecated method and gets transformed automatically (the transcript should have a log about it).

These auto-transformed methods can be just committed, even as part of another PR, I think.

hilaire:
Hum, yes. And about my questioning?

MarcusDenker:
This PR still deletes a whole class. The easiest is to just redo it... there are now merge conflicts in addition.

estebanlm:
:( 
Please, submit a new PR.

PR #4655, opened 2019-09-19
Cannot-rewrite-code-with-empty-replacement

jecisc:
Allows one to write a rule to remove a node from the AST. 

In order to do that I introduced a RBEmptyNode in the AST to use when the rewrite rule contains nothing instead of an empty RBSequenceNode.

### DIFF
--- a/src/AST-Core-Tests/RBParseTreeRewriterTest.class.st
+++ b/src/AST-Core-Tests/RBParseTreeRewriterTest.class.st
@@ -17,6 +17,15 @@ RBParseTreeRewriterTest >> compare: anObject to: anotherObject [
 ]
 
 { #category : #helpers }
+RBParseTreeRewriterTest >> methodWithNilAssignment [
+	| temp |
+	temp := 1.
+	temp := 1 + temp.
+	temp := nil.
+	
+]
+
+{ #category : #helpers }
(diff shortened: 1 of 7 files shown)


jecisc:
Tests are related :(

jecisc:
Tests passed but do not integrate it yet please. We need to think about it with Marcus to see if we can do better.

Ducasse:
This is cool 
Now I wonder why do we need 

{ #category : #accessing }
	RBEmptyNode >> allDefinedVariables [
		^ self temporaryNames asOrderedCollection
			addAll: super allDefinedVariables;
			yourself
	]
	
	{ #category : #accessing }
	RBEmptyNode >> allTemporaryVariables [
		^ self temporaryNames asOrderedCollection
			addAll: super allTemporaryVariables;
			yourself
	]
	
	{ #category : #accessing }
	RBEmptyNode >> children [
		^ OrderedCollection withAll: self temporaries
	]


And why you need to change the formatter?

jecisc:
Well, this is why I said we should think a little more about it before integrating. To make tests pass, I needed to duplicate a lot of the SequenceNode code in the EmptyNode. So I think there is something wrong there. I need to think a little more about it to find a better solution.

Ducasse:
OKI

PR #4656, opened 2019-09-19
Opal-clean-lookUpVar

MarcusDenker:
- rename lookUpVar:inContext: to #lookupAndReadVar:inContext
- remove all the  lookUpVar:inContext methods that where not needed
- factor out caclulation of index into #indexFromIR (for temps)


Ducasse:
broken test does not look related now may be marcus should have a look.

MarcusDenker:
error is not related

PR #4657, opened 2019-09-19
Remove-nil-initialization-from-initialize-methods

jecisc:
Remove nil assignations from initialize methods + add rule + add release test.


Ducasse:
Broken tests are related to Zinc... so we can merge

PR #4667, opened 2019-09-20
4652-Slot-definition

Ducasse:
all test green with ast nodes :)


Ducasse:
All the tests passed.

PR #4669, opened 2019-09-20
OpalTempReadSimplify-Step3

MarcusDenker:
next step:

- re-categorize methods a bit
- use #streamContents to simplify two methods
- instead of lookupAndReadVar: name inContext: aContext I added a method to return the context where the variable is defined: lookupDefiningContextFor: name startingFrom: aContext
- this way we can fix the write methods, too, to store in the correct context. 
- this means that we do not need indexFromIRFromScope: anymore, as we now always use the offset of the defintion context


Ducasse:
Broken tests are unrelated

PR #4673, opened 2019-09-20
Review-Adding the concept of testingEnvironment.

Alisu:
A lot of tests checked if a class is present in the environment.
This instance variable is to keep this environment to do those kind of checks.

The commits following are refactors in order to take into account this IV.


src/SUnit-Core/TestCase.class.st
```diff
@@ -10,7 +10,8 @@ Class {
 	#superclass : #TestAsserter,
 	#instVars : [
 		'testSelector',
-		'expectedFails'
+		'expectedFails',
+		'aTestEnvironment'
```
dionisiydk: Why the name begins with a?

Alisu: The name without the a is taken by another slot in a subclass.
I will certainly change the name later.

jecisc: It could be "testingEnvironment”

Ducasse:
I think that we should pay attention that if a subclass of an external project defines the same attribute it will not be able to load. 
So I would introduce a new subclass of TestCase, add the instance variable there and migrate our tests using Smalltalk globals to be a subclass of this class.

jecisc:
Maybe we could have a trait instead of a subclass? Because if we use the environnement in a test case that is already specialized, we will have troubles

Alisu:
Not sure i understood well. Should i finish to push the commits ? 

Edit: I pushed them all. I have to do the same work for the Smalltalk at:

Ducasse:
Hi theo
could you check this test unix-32 / Tests-unix-32 / testExternalBasicToolsDependencies – Unix32.System.DependenciesTests.SystemDependenciesTest because it looks like there is a problem related to your changes.

Alisu:
They are fixed now. There is still tests failing in OSX.

PR #4682, opened 2019-09-20
2874-clampLowHigh

### LINKED ISSUE #2874: Replace #min:max with #clampLow:High:
Moved from https://pharo.fogbugz.com/f/cases/21758/Replace-min-max-with-clampLow-High

> On Sat, 21 Apr 2018, Ben Coman wrote:
>> On 21 April 2018 at 03:51, Hilaire <hilaire@drgeo.eu> wrote:
>>      Hi,
>>
>>      Out of curiosity.
>>      I always found the #min:max: confusing and lost in its expressiveness.
>>      One should write:
>>
>>          10 min: 48 max: 12
>>
>>      to expect 12.
>>      but logically one (at least me) may want to express it as:
>>
>>          10 min: 12 max: 48
>>
>>      Then when reading its source code, it is even more confusing:
>>
>>      min: aMin max: aMax
>>          ^ (self min: aMin) max: aMax
>>
>>      Are not the argument names inversed in their meaning, if any?
>>
>>
>> I would agree.  I see most use by Color like...
>>  
>>    Color>>adjustBrightness: brightness
>>        "Adjust the relative brightness of this color. (lowest value is 0.005 so that hue information is not lost)"
>>
>>        ^ self class
>>                h: self hue
>>                s: self saturation
>>                v: (self brightness + brightness min: 1.0 max: 0.005)
>>                alpha: self alpha
>>
>> Trying to read that twists my brain.  
>>
>>
>> I can understand the intent from the implementation 
>> min: aMin max: aMax 
>> ^ (self min: aMin) max: aMax
>>
>> but that message might more properly be  #min:thenMax:
>> However something like    
>>    (self brightness + brightness boundedBy: 0.005 and: 1.0)
>>    (self brightness + brightness boundedMin: 0.005 max: 1.0)
>> seems more intention revealing.
>>
>>
>> Altering  #min:max  semantics would make awful portability,
>> but perhaps it should forward to a new method to make it clear the other is preferred.
>>
>> Would the Squeak community be amenable to a similar change?
>> I'd be happy to contribute the changes to the Squeak Inbox.

On 21 April 2018 at 17:56, Levente Uzonyi <leves@caesar.elte.hu> wrote:
>
> Squeak has #clampLow:high: for this reason.
>
> Levente

hilaire:
Fixes #2874
- implement clampLow:high:,
- redirect min:max: to it
- add a test


Ducasse:
Could we improve the method comment:

Magnitude >> clampLow: minValue high: maxValue [
	"Answer the value within the interval [minValue ; maxValue]; if the value is contained into the interval then return it, else return the adequate interval value."

Ducasse:
I read the translation in french about clamp (fixer, attacher) and I was thinking that clampLow:high: is not a super nice name. 
We could really take the time to think about a better. 
May be we can ask in the ML for ideas.
I would add between 
and I would prefer beBetween: and: 
Because clampLow:high: does not convey the intention of the method (at least when I read it, I do not understand it).

Ducasse:
" 10 clampLow: 12 high: 20 >>> 12"
should be rewritten  "(10 clampLow: 12 high: 20) >>> 12"

Ducasse:
Hi hilaire

I was thinking about it and I did not like the solution. 
Changing min:max: to introduce a new method with its own problems could be better. 
I think that it would be great that the new method is not depend on the fact that min is smaller than max.

Because I understand the original point: it was that with low: and high: people would pass "correct" i.e., argument in order smaller and larger to get a positive interval. 
But this only works well when we pass number now if we pass x and y. 
We have not idea that the result will not be the one we expect. 

I would like to have a real discussion about it. 
And I would let min:max: like it is but adding a real method comment and examples.

Then I would define beBetween: that works in all cases for real.
What do you think?

Here are the scenario.

```
(5  min: 12 max: 20) 
>> 20
>>> this is a problem 
(5 clampLow: 12 high: 20) 
>> 12
(5 beBetween: 12 high: 20)
>> 12 

(5  min: 20 max: 12)
>> 12
(5 clampLow: 20 high: 12)
>> 20 ????
(5 beBetween: 20 high: 12)
>> 12

(100  min: 20 max: 12) 
>> 20
(100 clampLow: 20 high: 12)
>> 20

(100  min: 12 max: 20) 
>> 20
(100 clampLow: 12 high: 20) 
>> 20
```

Ducasse:
min:max: comment should really say that it is min:thenMax:
and may be we should rename it like. 

The solution would be 
- min:max: -> renamed min:thenMax:
add 
- beBetween:and: 

with different semantics.

hilaire:
Agree, I don't like much the clampLow:hight: message. The intention was the keep come compatibility with Squeak, but we likely passed this point.

hilaire:
May be to keep some compatibility issues, we could just keep min:max: as it is, renamed its parameters, then add a proper comment plus test as you suggested it.
As least the user will have proper information.

Now thinking about it, beBetween:and: when self is outside the interval, the message name is not meaningful about which result we expect.

10 beBetween: 5 and: 8
=> 5 or 8 ?
10 beBetween: 8 and: 5
=> 5 or 8 ?

Or something like beBetweenLow:andHigh: but your idea is the parameters can be reversed without harm, and it will not work with this name.

It fell like antinomism.

We could just live with min:max:

Ducasse:
I will discuss with pablo and guille and let you know.

Ducasse:
After discussing with guile and pablo I think that we can
- deprecate min:max: into min:thenMax:
- then we can introduce add beBetween:and: with a correct semantics (not having the trouble of mix:max:)

hilaire:
What is the expect result in these scenari

10 beBetween: 5 and: 8
=> 5 or 8 ?
10 beBetween: 8 and: 5
=> 5 or 8 ?

Ducasse:
Like clipping 8 because 8 is the closest. 
May be we should discuss on the mailing-list if it is worth now
to me clampLow:High: does not bring anything and it is more confusing than min:thenMax:

Stef

hilaire:
I don't understand the conflict

hilaire:
I am not sure either about replacing `min:max:` by `min:thenMax:`
- it will brake a lot of packages, it is use in many places
- the intend of the message `min:thenMax:` is only slightly better (it tells about what it is doing but not about its meaning), indeed, the parameters are still swapped in meaning (knots in your brain).
- a `min:max:` message with proper comments and examples will already help a lot the developer

Ducasse:
Hilaire usually for central message we deprecated them over multiple version. 
So we can really introduce min:thenMax: and fix all the senders of min:max: to use mix:thenMax: and then we deprecate with transformation min:Max: and people get two years to automatically migrate. 
I think that this is more than effort.

hilaire:
@Ducasse ok understood. What do you think about the other points I mentioned?

Ducasse:
Sorry which other point.

hilaire:
in my previous comment

Ducasse:
I do not understand how to read the conflicts else the code is good for me.

Ducasse:
I fixed the conflicts and waiting for the build to rerun before integrating the changes.

PR #4689, opened 2019-09-23
4686-code-completion-not-working-properly-on-inspectors

### LINKED ISSUE #4686: code completion not working properly on inspectors
if you open an inspector and start typing for example : 

```Smalltalk
42 factorial
```

It will raise a DNU

MarcusDenker:
CompletionContext should check the editor if we are in a workspace or not. fixes #4686

(the setter of CompletionContext will be cleaned up more when we remove the old completion)


MarcusDenker:
errors are related

PR #4690, opened 2019-09-23
3434-Random--nextInt-should-fail-early-when-argument-is-FloatInfinity

kasperosterbye:
Fixes 3434

Added a comment that the expression "anInteger asFloat isInfinite" is just a way to check if we are outside the range of float.


MarcusDenker (approved):
comment added, ok!

PR #4691, opened 2019-09-23
Context tempNamed: use Debuggermap directly

MarcusDenker:
tempNamed: and #tempNamed:put: where implemented in terms of #namedTempAt:, which means that they call the debugger map twice.
We can instead use #tempNamed:in: directly


MarcusDenker:
one slave fails (unrelated), tests green

PR #4696, opened 2019-09-23
Opal: DefiningContext should be looked up by scope

MarcusDenker:
lookupDefiningContextFor: startingFrom:  was using variable names. It is better to instead walk up contexts till the scope is the definition scope of the variable (independend of the name).


MarcusDenker:
failing test not related

gcotelli (approved):
LGTM

PR #4700, opened 2019-09-23
4698 reflectivity object centric features not migrated from p7 to p8

### LINKED ISSUE #4698: Reflectivity object-centric features not migrated from P7 to P8
The per-object reflectivity is in P7 but not in p8.

StevenCostiou:
Merging P7 Reflectivity object-centric with P8 Reflectivity
Fixes #4698


StevenCostiou:
@MarcusDenker Can you have a look?

MarcusDenker:
There are two tests failing (easy to fix):

A test class (subclass of TestCase) should have it's name ending with 'Test' or 'TestCase' Violations: - MetaLinkObjectAPITests - MetaLinkTargetResolverTests - LinkInstallerTests

ProperManifestCategorizationTest>>testProperManifestCategorization

MarcusDenker (approved):
Ok, we should merge this (thus both Pharo7 and 8 are in sync) and then we can so the next iteration (I have an idea why links on temps does not work).

PR #4707, opened 2019-09-24
4704-reduce-direct-calls-of-debuggerMap-rangeForPCcontextIsActiveContext

### LINKED ISSUE #4704: reduce direct calls of #debuggerMap: rangeForPC:contextIsActiveContext:
rangeForPC:contextIsActiveContext: should be implemented as  #rangeForPC:contextIsActive:  on Context 

This way we do not need to call #debuggerMap from the ProcessBrowser

MarcusDenker:
- Implement #rangeForPC:contextIsActive:  in CompiledCode, calling #debuggerMap
- implement versions on Context: pcRange, pcRangeContextIsActive: 
- push up #debuggerMap to CompiledCode
- fix senders

fixes #4704

### DIFF
--- a/src/Debugging-Core/Context.extension.st
+++ b/src/Debugging-Core/Context.extension.st
@@ -56,6 +56,16 @@ Context >> namedTempAt: index put: aValue [
 ]
 
 { #category : #'*Debugging-Core' }
+Context >> pcRange [
+	^self pcRangeContextIsActive: false
+]
+
+{ #category : #'*Debugging-Core' }
+Context >> pcRangeContextIsActive: aBoolean [
+	^self method rangeForPC: pc contextIsActive: aBoolean
+]
+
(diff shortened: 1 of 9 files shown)


MarcusDenker:
@StevenCostiou please review

src/Debugging-Core/Context.extension.st
```diff
@@ -55,6 +55,23 @@ Context >> namedTempAt: index put: aValue [
 	^self debuggerMap namedTempAt: index put: aValue in: self
 ]
 
+{ #category : #'*Debugging-Core' }
+Context >> pcRangeContextIsActive: contextIsActive [
```
StevenCostiou, line 59: Method naming seems wrong, seems its supposed to return a boolean.
However the next commit contains a fix, so it should be accepted as is.

StevenCostiou (approved):

Issue #4706, opened 2019-09-24
convenient method to instantiate a Form from a base64 encoded String


demarey:
It looks like I'm not the only one who never remember how to get a Form out of an picture encoded with base64.
I propose to add a convenient method on the Form class.

PR #4708, opened 2019-09-24
4706 convenient method to instantiate a form from a base64 encoded string

demarey:


demarey:
See issue https://github.com/pharo-project/pharo/issues/4706

Ducasse:
Could we have a little test?
Else yes definitively needed.

demarey:
Sure, Thanks for the reminder.
I added one test and also put an extra comment to explain how to get the encoded String.
There is also a PR here: https://github.com/pharo-open-documentation/pharo-wiki/pull/242

MarcusDenker (approved):

PR #4714, opened 2019-09-24
4701-Review-asInteger-

### LINKED ISSUE #4701: Review #asInteger
As reported by Frank Lesser [LSW] on Discord today in channel #development:

```Smalltalk
(1/2) asInteger = 0
``` 

caused by Number >>#asInteger - not conforming to ANSI

kasperosterbye:
Fixes #4701

Changed the implementation to match the ansi standard.
Added a test case to match the standard and the issue raised.

### DIFF
--- a/src/Kernel-Tests/MagnitudeTest.class.st
+++ b/src/Kernel-Tests/MagnitudeTest.class.st
@@ -1,0 +1,8 @@
+"
+A MagnitudeTest is a test class for testing the behavior of Magnitude
+"
+Class {
+	#name : #MagnitudeTest,
+	#superclass : #TestCase,
+	#category : #'Kernel-Tests-Numbers'
+}
--- a/src/Kernel-Tests/NumberTest.class.st
+++ b/src/Kernel-Tests/NumberTest.class.st
@@ -7,6 +7,13 @@ Class {
 	#category : #'Kernel-Tests-Numbers'
 }
 
+{ #category : #test }
+NumberTest >> testAsInteger [
+	self assert: 0.5 asInteger equals: 1.
+	self assert: (1/2) asInteger equals: 1
+	
+]
+
 { #category : #tests }
 NumberTest >> testFractionPart [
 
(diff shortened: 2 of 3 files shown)


svenvc:
Looks OK.

I would add a bit more test cases, like 0 0.0 1/3 3/4 0.3 0.7 -1.4 -1.5 -1.75

Ducasse:
I think that there is an impact on the bootstrap. I restarted the build.

kasperosterbye:
I have looked at this a bit more. I believe the right thing is to let it remain truncated rather than rounded. The reason is that Squeak is also doing truncated, and it appears to be used  for a lot of geometry calculations which computes middle lines. All of those are likely to go one off if we change this. I have reverted the implementation back, but did add a few tests to `rounded` while I was at it. I also added a comment in `asInteger` reflecting that we are not ansi compliant.

kasperosterbye:
@svenvc I added some extra tests to `rounded`. Actually I do not believe we need more tests for `asInteger` because that test only need to ensure we are using truncate/rounded. 
What I am trying to say, is that correct rounding is not the "subject under test". 
I am relatively new to Pharo, so I might have overlooked some principle.

frank-lesser:
looking at senders of asInteger changing to ANSI has quite a few side effects ( e.g Float parsing ) - for instance in Float >> #asApproximateFractionAtOrder:  asInteger has to be changed to truncated.
or (Float pi asApproximateFractionAtOrder: 3) = (355/113) answers false.  also it may afect Number parsing - so be warned this is not a trivial change

Ducasse:
Hi kasper
Yes I think that this is more reasonable. I like the tests :).

Stef

src/Kernel-Tests/MagnitudeTest.class.st
```diff
@@ -0,0 +1,8 @@
+"
+A MagnitudeTest is a test class for testing the behavior of Magnitude
+"
+Class {
+	#name : #MagnitudeTest,
```
nicolas-cellier-aka-nice, line 5: I see a new class without any new method... YAGNI or WIP?

kasperosterbye: You seem to be better with abbreviations than I do :-).
However, you are right, and I opened issue #4730 to that matter, and hope to have a fix sometime today.

Issue #4719, opened 2019-09-25
Cannot load Pharo with a metacello script


dupriezt:
Running the following in a pharo 8 image throws a SubscriptOutOfBound exception.
```Smalltalk
Metacello new
	baseline: 'Pharo';
	repository: 'github://pharo-project/pharo:Pharo8.0/src';
	load.
```
The exception happens in MetacelloPlatform>>extractTypeFromDescription:
```Smalltalk
extractTypeFromDescription: description
  description == nil
    ifTrue: [ ^ nil ].
  ((description beginsWith: '/') or: [ description second = $: ])
  ...
```
*description* is an empty string (not nil), so `description second` throws the aforementioned exception.

dupriezt:
The issue comes from BaselineOfPharo defining sub-baseline using `self repositoryPackageURL` (BaseLibraries and IDE). Loading pharo with the metacello script works when these sub-baselines are commented. It also works when these sub-baselines explicitly define the address of the repository: for example `spec baseline: 'BaseLibraries' with: [spec repository: 'github://pharo-project/pharo:Pharo8.0/]`.
```Smalltalk
BaselineOfPharo>>#baseline: spec
	<baseline>
	| repository |
	repository := self packageRepositoryURL.

	spec for: #common do: [
		spec postLoadDoIt: #'postload:package:'.
		spec baseline: 'BaseLibraries' with: [spec repository: repository].
		
		spec baseline: 'IDE' with: [spec repository: repository].
                ...
```

dupriezt:
I set up a simpler repository with a sub-baseline to reproduce the issue quicker: https://github.com/dupriezt/testBsl (instead of having to wait for every package of Pharo to be checked out everytime)
Loading script:
```Smalltalk
Metacello new
    baseline: 'TestBsl';
    repository: 'github://dupriezt/TestBsl';
    load.
```

<blockquote><img src="https://avatars0.githubusercontent.com/u/32486709?s=400&v=4" width="48" align="right"><div><img src="https://github.githubassets.com/favicon.ico" height="14"> GitHub</div><div><strong><a href="https://github.com/dupriezt/testBsl">dupriezt/testBsl</a></strong></div><div>Contribute to dupriezt/testBsl development by creating an account on GitHub.</div></blockquote>

dupriezt:
The issue is that BaselineOf>>#packageRepositoryURL does not find the url of the repo when it is loaded as a sub-baseline.
[...]

PR #4722, opened 2019-09-25
BaselineOf>>#packageRepositoryURL tries to find the repo url by looki…

dupriezt:
…ng at  iceberg repos  if it cannot find it otherwise

### DIFF
--- a/src/Metacello-PharoExtensions/BaselineOf.extension.st
+++ b/src/Metacello-PharoExtensions/BaselineOf.extension.st
@@ -32,9 +32,11 @@ BaselineOf >> packageRepositoryURL [
 		ifNil: [ '' ].
 	url ifNotEmpty: [ ^ url ].
 	
-	"The following is in case the url of the repository could not be determined. It looks at iceberg repositories in the image to see if any contains the baseline we are trying to load. If there is one, we return its repository url"
-	packageName := self class package name.
-	^ 'tonel://' , (IceRepository registry detect: [ :repo | repo workingCopy packageNames includes: packageName ] ifFound: [ :repo | repo location pathString , '/' , repo subdirectory ] ifNone: [ '' ])
+	self class environment at: #IceRepository ifPresent: [ :class |
+		"The following is in case the url of the repository could not be determined. It looks at iceberg repositories in the image to see if any contains the baseline we are trying to load. If there is one, we return its repository url"
+		packageName := self class package name.
+		^ 'tonel://' , (class registry detect: [ :repo | repo workingCopy packageNames includes: packageName ] ifFound: [ :repo | repo location pathString , '/' , repo subdirectory ] ifNone: [ '' ])
+	].
 
 
 	
--- a/src/Metacello-PharoExtensions/BaselineOf.extension.st
+++ b/src/Metacello-PharoExtensions/BaselineOf.extension.st
@@ -26,9 +26,15 @@ BaselineOf >> packageRepositoryURL [
 	" Tries to determine a repository URL from which the baseline is being loaded. Useful for 
 	refering other baselines in the same repository. "
 
-	^ self packageRepository
+	| url packageName |
+	url := self packageRepository
 		ifNotNil: [ :aRepository | aRepository description ]
-		ifNil: [ '' ]
+		ifNil: [ '' ].
+	url ifNotEmpty: [ ^ url ].
+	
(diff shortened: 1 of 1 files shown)


dupriezt:
Solves #4719

jecisc (approved):
I added this in Moose for multiple months and it works fine to manage a project with multiple baselines referencing themselves.

Ducasse:
Sounds good. Now I do not get how the current situation is working since we load the baseline after each commit. So why is it working there and not with your scenario. May be the loading is different.

jecisc:
It is working when you load with a Tonel or FileTree template but not a github:// or gitlab:// or bitbucket:// repository template.

src/Metacello-PharoExtensions/BaselineOf.extension.st
```diff
@@ -26,9 +26,17 @@ BaselineOf >> packageRepositoryURL [
 	" Tries to determine a repository URL from which the baseline is being loaded. Useful for 
 	refering other baselines in the same repository. "
 
-	^ self packageRepository
+	| url packageName |
+	url := self packageRepository
```
estebanlm, line 30: This can be solved a lot easier as: 

```Smalltalk
self packageRepository ifNotNil: [ :aRepository | ^ aRepository description ]
```

src/Metacello-PharoExtensions/BaselineOf.extension.st
```diff
@@ -26,9 +26,17 @@ BaselineOf >> packageRepositoryURL [
...
 		ifNotNil: [ :aRepository | aRepository description ]
-		ifNil: [ '' ]
+		ifNil: [ '' ].
+	url ifNotEmpty: [ ^ url ].
+	
+	self class environment at: #IceRepository ifPresent: [ :class |
+		"The following is in case the url of the repository could not be determined. It looks at iceberg repositories in the image to see if any contains the baseline we are trying to load. If there is one, we return its repository url"
+		packageName := self class package name.
```
estebanlm, line 37: I would extract this to a method.

estebanlm (changes requested):
I don't understand why we want this, can you elaborate?

estebanlm:
And in any case: Pharo9.0 :)

jecisc:
@estebanlm It's in case you have a project containing multiple baselines referencing themselves. 

For example in Moose you have BaselineOfMoose referencing BaselineOfFamix referencing BaselineOfFame. The three of them are in the same repository.

In that case, when declaring the dependency you need to give the repository template that invocked the loading of the baseline. For example, if you load BaselineOfMoose via `github://moosetechnology:v1.0.0/src`, then the repository for the Famix dependency need to be `github://moosetechnology:v1.0.0/src`. You cannot hardcode the version as you do when you declare external dependencies.

Here comes the method #packageRepositoryURL. This method gives you the right repository to give to internal dependencies. It is currently working if you already cloned the project and load the clone by referencing the folder (like the Pharo Jenkins is doing), but it does not work if you install a project directly from github/gitlab/bitbucket. 

This PR improve this by looking in the Iceberg registry to find the right repository URL in the project is registered there.

jecisc:
Ping

guillep:
This issue is stalled since december with requested changes.
I'll close it.

Please, feel free to re-open if the issues are going to be addressed.

jecisc:
This is really useful to me and multiple people came to ask me how I did it to do it themself. So I think it is important to have it. 

The two request changes are nothing major, just details to make the code more readable. I think a ping would be enough instead of closing the PR with the risk of losing a change that is already working.

guillep:
> This is really useful to me and multiple people came to ask me how I did it to do it themself. So I think it is important to have it.

That's fair. Are you willing to push the issue then?
Because if I base myself on what is in this issue:
 - open in september
 - review in december
 - ping in february

It's more than six months, and pinging does not get the job done.
Let's at least agree on that.

> The two request changes are nothing major, just details to make the code more readable

I'll leave outside of the discussion that I think this solution is not a solution but a hack.
This is the kind of code that when we find it 1 year later we say "Why is this here?".
It introduces a brittle dependency against Iceberg that the dependency analyser could fail to find
(`self class environment at: #IceRepository`) and it depends on a non-public API of iceberg that may be subject to change.

Now, I could accept this, but we need to acknowledge this is not a good solution, even if this "works". Do we?

> I think a ping would be enough instead of closing the PR with the risk of losing a change that is already working.

History shows pinging is not enough. Thomas has his own agenda, and I understand that.
So somebody has to take responsibility for this, that's all.

PR #4723, opened 2019-09-25
4670-Deprecation-transfrom-make-sure-the-fallback-is-signal

### LINKED ISSUE #4670: Deprecation transfrom: make sure the fallback is #signal
in #transform we have a number of ^self. 

They should be ^self signal, so that in this case a normal deprecation is signalled

MarcusDenker:
fixes #4670. no #singal in the case for the #isDoit, as we really want to ingore it there


jecisc:
This creates a lot of error on Spec tests.

We should check why.

MarcusDenker:
after re-running the tests, all tests are green

PR #4724, opened 2019-09-25
Improve-stability-of-Docking-bar-morph

jecisc:
I have projects adding elements to the MenuBar in Pharo. In the CI of the projects it often happens that it crash in DockingBarMorph>>updatePosition because between the start and end of the method the owner was removed. 

This commit store the owner to not end up with a nil during the update.


jecisc:
Error not related.

PR #4726, opened 2019-09-25
ZnBufferedReadWriteStream should handle better the buffer.

tesonep:
Creating a custom Buffered Read/Write Stream that shares the buffer for writing and reading.

In Windows, this is really important when using disks with physical segments of 4k.

### DIFF
--- a/src/FileSystem-Memory/MemoryFileWriteStream.class.st
+++ b/src/FileSystem-Memory/MemoryFileWriteStream.class.st
@@ -50,6 +50,12 @@ MemoryFileWriteStream >> isBinary [
 ]
 
 { #category : #writing }
+MemoryFileWriteStream >> next: anInteger putAll: aCollection startingAt: anInteger3 [ 
+	
+	stream next: anInteger putAll: aCollection startingAt: anInteger3 
+]
+
+{ #category : #writing }
 MemoryFileWriteStream >> nextPut: aCollection [
 	^ self stream nextPut: aCollection
 ]
@@ -71,6 +77,12 @@ MemoryFileWriteStream >> position: anInteger [
 	create an error notification."
 
 	stream position: anInteger
+]
+
+{ #category : #reading }
+MemoryFileWriteStream >> readInto: aCollection startingAt: anInteger count: anInteger3 [ 
+
+	^ stream readInto: aCollection startingAt: anInteger count: anInteger3 
 ]
 
 { #category : #positioning }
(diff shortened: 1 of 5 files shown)


tesonep:
This fix improves the speed of loading code in around 20 times.

jecisc:
This might interest @svenvc

Ducasse:
Pablo you got some SDL 20 mixed inside (just a couple of methods).

Ducasse:
I wonder why we have code from MemoryFileStream in this PR.

MemoryFileWriteStream >> next: anInteger putAll: aCollection startingAt: anInteger3 [ 

 	stream next: anInteger putAll: aCollection startingAt: anInteger3 
 ]

tesonep:
> I wonder why we have code from MemoryFileStream in this PR.
> 
> MemoryFileWriteStream >> next: anInteger putAll: aCollection startingAt: anInteger3 [
> 
> ```
> stream next: anInteger putAll: aCollection startingAt: anInteger3 
> ```
> 
> ]

The API of MemoryFileWriteStream was missing this method to be polymorphic with FileWriteStream

tesonep:
More insight about the change. 
Basically in new machines with 4kb physical sectors solid hard drives,  Windows formats the drives using 4kb sectors and emulating 512 bytes sectors.  (https://docs.microsoft.com/es-es/windows/win32/w8cookbook/advanced-format--4k--disk-compatibility-update?redirectedfrom=MSDN) 

Basically it handles the writing and reading of small sectors accessing the disk.
This is really bad in the changes file as we are reading and writing the same file all the time. 
Moreover, the changes file was using buffered streams for reading and writing, but the integration of them is through the disk, so any combination of reading and writing operations flush the buffers forcing accesses to the disk, therefore, slowing the process.

In a stock Pharo 8 image, the time to compile methods in a machine using emulated 512 bytes segments takes a lot of time: 

Object subclass: #Prb
	instanceVariableNames: ''
	classVariableNames: ''
	package: 'Prueba'.
	
[20 timesRepeat: [ Prb compile: 'm2 
	^ 42']] timeToRun. 

"0:00:00:23.194"

This problem is really clear when loading big projects, as compiling any class or method access the changes file. Moreover, this also improves the times in a machine without 4kb physical sector disks.

When using the new solution that shares the buffer and improves the flushing handle the times get better:

[20 timesRepeat: [ Prb compile: 'm2 
	^ 42']] timeToRun. 
 "0:00:00:00.323"



<blockquote><div><strong><a href="https://docs.microsoft.com/en-us/windows/win32/w8cookbook/advanced-format--4k--disk-compatibility-update">Advanced format (4K) disk compatibility update - Windows applications</a></strong></div><div>Advanced format (4K) disk compatibility update</div></blockquote>

jecisc:
The failing test seems unrelated.

PR #4737, opened 2019-09-26
fix-necontroller-can-be-nil

estebanlm:
controller can be nil (not always we want that). 
Also, controller could not come from morphic, so introduce this hidden dependency will not be beneficial.
demeters law!


MarcusDenker (approved):
ok

MarcusDenker (changes requested):
Ups, test failing:

SubContextTest>>testReceiverArgument

estebanlm:
wow. 
how's that possible? :P

MarcusDenker (approved):

PR #4739, opened 2019-09-26
Porting a new implementation of ZnBufferedReadWriteStream

tesonep:
More insight about the change.
Basically in new machines with 4kb physical sectors solid hard drives, Windows formats the drives using 4kb sectors and emulating 512 bytes sectors. (https://docs.microsoft.com/es-es/windows/win32/w8cookbook/advanced-format--4k--disk-compatibility-update?redirectedfrom=MSDN)

Basically it handles the writing and reading of small sectors accessing the disk.
This is really bad in the changes file as we are reading and writing the same file all the time.
Moreover, the changes file was using buffered streams for reading and writing, but the integration of them is through the disk, so any combination of reading and writing operations flush the buffers forcing accesses to the disk, therefore, slowing the process.

In a stock Pharo 8 image, the time to compile methods in a machine using emulated 512 bytes segments takes a lot of time:

Object subclass: #Prb
instanceVariableNames: ''
classVariableNames: ''
package: 'Prueba'.

[20 timesRepeat: [ Prb compile: 'm2
^ 42']] timeToRun.

"0:00:00:23.194"

This problem is really clear when loading big projects, as compiling any class or method access the changes file. Moreover, this also improves the times in a machine without 4kb physical sector disks.

When using the new solution that shares the buffer and improves the flushing handle the times get better:

[20 timesRepeat: [ Prb compile: 'm2
^ 42']] timeToRun.
"0:00:00:00.323"

### DIFF
--- a/src/FileSystem-Memory/MemoryFileWriteStream.class.st
+++ b/src/FileSystem-Memory/MemoryFileWriteStream.class.st
@@ -47,6 +47,12 @@ MemoryFileWriteStream >> isBinary [
 ]
 
 { #category : #writing }
+MemoryFileWriteStream >> next: numberOfElements putAll: aCollection startingAt: startingAt [ 
+
+	^ stream next: numberOfElements putAll: aCollection startingAt: startingAt 
+]
+
+{ #category : #writing }
 MemoryFileWriteStream >> nextPut: aCollection [
 	^ self stream nextPut: aCollection
 ]
@@ -59,6 +65,12 @@ MemoryFileWriteStream >> nextPutAll: aCollection [
 { #category : #positioning }
 MemoryFileWriteStream >> position [
 	^ self stream position
+]
+
+{ #category : #position }
+MemoryFileWriteStream >> position: aPosition [ 
+	
+	stream position: aPosition
 ]
 
 { #category : #positioning }
--- a/src/Tests/SourceFileArrayTest.class.st
+++ b/src/Tests/SourceFileArrayTest.class.st
@@ -375,10 +375,13 @@ SourceFileArrayTest >> testWriteToBufferedStream [
 	| fs array |
 	fs := FileSystem memory.
 	array := SourceFileArray new.
-	array
-		changesFileStream: (ZnBufferedReadWriteStream on: (fs / 'changes.chunk') writeStream).
-	array
-		sourcesFileStream: (ZnBufferedReadWriteStream on: (fs / 'sources.chunk') writeStream).
+	
+	array changesFileStream: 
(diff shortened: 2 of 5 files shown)


svenvc:
Hi,

I think I understand what you are trying to do, and if it makes so much performance difference, it must certainly be done. I do consider this class a bit outside Zn, because it is actually quite specific for accessing the changes file. In Zn I try to separate reading and writing, and simplify as much as possible buffer management, as it is very tricky,

This is very ambitious (to get right). I think you have to add some comments describing what it actually means to share a buffer between a read and a write stream, in terms of semantics of the different operations.

Sven

tesonep:
Hi Sven, 
   the name is an inheritance because we have already a class named like that. This class was wrapping two separate buffered streams, one for reading and the other for writing. This class is only used by the source files, maybe we can move this class next to the source file and do not use it elsewhere.  However, I am not sure if there are other users reading and writing the same file that can get an improvement with this.

Thanks for your checking.

tesonep:
As I have checked the given class is used only in:

- SourceFiles
- OMStoreFile
- ZnFileSystemUtils 

But I am almost sure the last 2 users can be removed without problem as they are not using the file for reading and writing at the same time.

Ducasse:
Indeed it may be good to 
- change the wrong users
- package it close to its unique then user.

svenvc:
Yes I agree

tesonep:
In Pharo 7 I will let the modification as it is.
I will improve the implementation as Stef said in Pharo 8. 
The tests that are broken are also broken in a clean image.
@estebanlm can we include this in a new version?

Ducasse:
We got this 3 tests broken 
testSizeAndSpan – Unix32.Glamour.Tests.Morphic.GLMTabulatorMorphicTest
testDropChangeFile – Unix32.Tool.ExternalBrowser.Test.ExternalBrowserTest
testDropStFile – Unix32.Tool.ExternalBrowser.Test.ExternalBrowserTest

StephanEggermont:
Outside the image, there are more cases where simultaneous and buffered read and write access is needed to a file. The exact buffering and flushing semantics are very difficult to get right, indeed. They not only depend on the domain needs, but also on the underlying file system used. https://danluu.com/file-consistency/ (and several more articles on his site)

<blockquote><div><strong><a href="https://danluu.com/file-consistency/">Files are hard</a></strong></div></blockquote>

Ducasse:
Hi pablo so should we integrate this?

Ducasse:
broken tests are not related so I commit.

Ducasse:
Should I revert since I did not see that it was for P7.

guillep:
I don't think so. Broken tests were not related to this change.
Still, this is not available in pharo7 yet, a new release is needed (via a tag).

PR #4742, opened 2019-09-27
4741-remove-DebuggerMethodMapOpal

### LINKED ISSUE #4741: Deprecated DebuggerMethodMapOpal
DebuggerMethodMapOpal is just a tiny facade over the AST... it was there to be compatible to the old compiler, but now that we unloaded the old compiler we do not need it anymore.

after merging https://github.com/pharo-project/pharo/pull/4740 (wich removes one API already) we can just inline the methods into the caller.

This will simplify things *and* speed up the code as we will not have to create a DebuggerMethodMapOpal instance for every call.

MarcusDenker:
remove DebuggerMethodMapOpal, fix senders.

fixes #4741


MarcusDenker:
This removes the mapper class, I will do another PR to bring back a deprecated version

MarcusDenker:
I will do a new one with a deprecated version, I close this

PR #4746, opened 2019-09-27
2863-Stack-should-not-inherit-from-LinkedList

### LINKED ISSUE #2863: Stack should not inherit from LinkedList
From: https://pharo.fogbugz.com/f/cases/22450/Stack-should-not-inherit-from-LinkedList

Because the fact it is implemented using a LinkedList is only a technical detail.

A client of Stack should not have access to all the LinkedList API, else it means what it needs is a linked list not a stack.

juliendelplanque:
Introduced a NewStack object that inherits from Collection and provide the minimal Stack API.

This NewStack uses a strategy design pattern for its internal storage.
I created LinkedList strategy and OrderedCollection strategy.

Stack class is now deprecated. This is the best way to avoid breaking everything.

The idea is that once we remove the deprecated Stack (next major release). We can rename NewStack as Stack, create a deprecated NewStack subclass and get back to a well named Stack.

Fixes #2863

### DIFF
--- a/src/Collections-Stack/NewStack.class.st
+++ b/src/Collections-Stack/NewStack.class.st
@@ -80,6 +80,11 @@ NewStack >> remove: oldObject ifAbsent: anExceptionBlock [
 	self shouldNotImplement
 ]
 
+{ #category : #removing }
+NewStack >> removeAll [
+	self strategy reset
+]
+
 { #category : #accessing }
 NewStack >> strategy [
 	^ strategy
--- a/src/Collections-Stack/NewStack.class.st
+++ b/src/Collections-Stack/NewStack.class.st
@@ -1,0 +1,92 @@
+"
+I am a parametrizable stack. That is to say, I provide the interface of a stack but depending on my #strategy, I can use a LinkedList, an OrderedCollection, etc for implementation details.
+"
+Class {
+	#name : #NewStack,
+	#superclass : #Collection,
+	#instVars : [
+		'strategy'
+	],
+	#category : #'Collections-Stack-Base'
+}
+
+{ #category : #factory }
+NewStack class >> usingLinkedList [
+ 	"Returns an instance of Stack using a LinkedList under the hood."
+ 	^ self usingStrategy: LinkedListStackStrategy new
+]
+
+{ #category : #factory }
+NewStack class >> usingOrderedCollection [
+ 	"Returns an instance of Stack using a OrderedCollection under the hood."
+ 	^ self usingStrategy: OrderedCollectionStackStrategy new
+]
+
+{ #category : #'instance creation' }
+NewStack class >> usingStrategy: aStackStrategy [
+ 	"Given a aStackStrategy, returns an instance of Stack using this strategy under the hood."
+ 	^ self new
+ 		initializeWith: aStackStrategy;
+ 		yourself
+]
+
+{ #category : #adding }
+NewStack >> add: newObject [
+ 	"Include newObject as one of the receiver's elements. Answer newObject. 
+ 	ArrayedCollections cannot respond to this message."
+
+  	^ self push: newObject
+]
+
+{ #category : #enumerating }
+NewStack >> do: aBlock [
+ 	"Evaluate aBlock with each of the receiver's elements as the argument."
+
+  	self strategy do: aBlock
+]
+
+{ #category : #initialization }
+NewStack >> initialize [
+	super initialize.
+	strategy := LinkedListStackStrategy new
+]
+
+{ #category : #initialization }
+NewStack >> initializeWith: aStrategy [
+	strategy := aStrategy
+]
+
+{ #category : #removing }
+NewStack >> pop [
+	" Remove the object at the top of the stack and return it.
+
+  	 (Stack new push: 1; push: 2; pop) >>> 2
+	"
+ 	^ self strategy pop
+]
+
+{ #category : #adding }
+NewStack >> push: newObject [
+ 	"Push newObject on the top of the stack."
+ 	self strategy push: newObject.
+ 	^ newObject
+]
+
+{ #category : #removing }
+NewStack >> remove: oldObject ifAbsent: anExceptionBlock [
+	"A stack should not implement that.
+	 Use a linked list if you really need this feature.
+	"
+	self shouldNotImplement
+]
+
+{ #category : #accessing }
+NewStack >> strategy [
+	^ strategy
+]
+
+{ #category : #accessing }
+NewStack >> top [
+ 	"Returns the object at the top of the stack without removing it."
+ 	^ self strategy top
+]
--- a/src/Collections-Tests/AbstractStackStrategyTest.class.st
+++ b/src/Collections-Tests/AbstractStackStrategyTest.class.st
@@ -1,0 +1,95 @@
+Class {
+	#name : #AbstractStackStrategyTest,
+	#superclass : #TestCase,
+	#instVars : [
+		'strategy'
+	],
+	#category : #'Collections-Tests-Stack'
+}
+
+{ #category : #testing }
+AbstractStackStrategyTest class >> isAbstract [
+ 	^ self = AbstractStackStrategyTest
+]
+
+{ #category : #running }
+AbstractStackStrategyTest >> setUp [
+ 	super setUp.
+ 	strategy := self stackStrategyClass new
+]
+
+{ #category : #accessing }
+AbstractStackStrategyTest >> stackStrategyClass [
+ 	^ self subclassResponsibility
+]
+
+{ #category : #tests }
+AbstractStackStrategyTest >> testEmptyError [
+ 	| aStack |
+ 	aStack := strategy.
+ 	self should: [ aStack top ] raise: Error.
+ 	self should: [ aStack pop ] raise: Error.
+ 	aStack push: 'element'.
+ 	aStack top.
+ 	aStack pop.	"The stack is empty again due to previous pop"
+ 	self should: [ aStack top ] raise: Error.
+ 	self should: [ aStack pop ] raise: Error
+]
+
+{ #category : #tests }
+AbstractStackStrategyTest >> testPop [
+
+  	| res elem aStack |
+ 	aStack := strategy.
+ 	elem := 'anElement'.	
+ 	self assert: aStack size equals: 0.
+
+  	aStack push: 'a'.
+ 	aStack push: elem.
+ 	res := aStack pop.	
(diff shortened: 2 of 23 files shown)


juliendelplanque:
The test failing is one of those randomly failing.

src/Collections-Stack/NewStack.class.st
```diff
@@ -0,0 +1,97 @@
...
+ 		yourself
+]
+
+{ #category : #adding }
+NewStack >> add: newObject [
+ 	"Include newObject as one of the receiver's elements. Answer newObject. 
+ 	ArrayedCollections cannot respond to this message."
+
+  	^ self push: newObject
+]
```
kasperosterbye, line 39: Why do you have an add: method?

src/Collections-Stack/NewStack.class.st
```diff
@@ -0,0 +1,97 @@
...
+ 	^ newObject
+]
+
+{ #category : #removing }
+NewStack >> remove: oldObject ifAbsent: anExceptionBlock [
+	"A stack should not implement that.
+	 Use a linked list if you really need this feature.
+	"
+	self shouldNotImplement
+]
```
kasperosterbye, line 81: Why include it just to give an error message?

src/Collections-Stack/NewStack.class.st
```diff
@@ -0,0 +1,97 @@
...
+
+{ #category : #removing }
+NewStack >> removeAll [
+	self strategy reset
+]
```
kasperosterbye, line 86: popAll?

src/Collections-Tests/AbstractStackStrategyTest.class.st
```diff
@@ -0,0 +1,111 @@
...
+{ #category : #tests }
+AbstractStackStrategyTest >> testReset [
+	| stack |
+	stack := strategy.
+	
+	stack push: 1.
+	stack push: 2.
+	stack push: 3.
+	
+	self assert: stack size equals: 3.
```
kasperosterbye, line 77: Technically I think this was already tested in testPush, so no need to do it again I think.

src/Deprecated80/Stack.class.st
```diff
@@ -1,12 +1,22 @@
 "
-I implement a simple Stack. #push: adds a new object of any kind on top of the stack. #pop returns the first element and remove it from the stack. #top answer the first element of the stack without removing it.
+This class is deprecated because while exposing the API of a Stack, it was also inheriting form LinkedList which allowed people to use LinkedList API.
+
+This was wrong because if one needs something else than #pop, #push, #top and #size, then it means that what one is looking for is a LinkedList.
+Thus, one should explicitely use it.
+
+Use NewStack instead.
```
kasperosterbye, line 7: Should you name it `PureStack`?

src/Collections-Stack/NewStack.class.st
```diff
@@ -0,0 +1,97 @@
...
+NewStack >> do: aBlock [
+ 	"Evaluate aBlock with each of the receiver's elements as the argument."
+
+  	self strategy do: aBlock
+]
+
+{ #category : #initialization }
+NewStack >> initialize [
+	super initialize.
+	strategy := LinkedListStackStrategy new
```
kasperosterbye, line 51: I am not sure LinkedList is the best strategy here, it has quite a object creation overhead over OrderedCollection. Have you done any analysis to check if this is the better choice, or did you just "inherit it" from the old stack.

Ducasse:
Do you think that people outside the core are using Stack?
Personally I have my own implementation in CT

Ducasse:
Then I have a question. Did you get a case where we need to have something else than an orderedCollection for a Stack? To me it looks overkill for a stack to have such complex strategy stuff.

src/Graphics-Files/LzwGifDecoder.class.st
```diff
@@ -121,13 +121,13 @@ LzwGifDecoder >> handleCode: anInteger withPreviousCode: prevInteger on: aWriteS
 				the prefix index values and take the suffix each
 				time"
 				ifFalse: [ 
-					searchStack := Stack new.
+					searchStack := NewStack new.
```
kasperosterbye, line 124: I have not checked these other parts of the system. I presume their tests are green after your rewrites.

kasperosterbye (commented):
My only overall comment is if NewStack is a bit too nonsense. I would suggest `PureStack`.
Also, I suggest to use the OrderdCollection as the default one.

juliendelplanque:
@Ducasse 

> Do you think that people outside the core are using Stack?

For sure, we use them a lot in Moose community for symbol resolution of AST for example.

> Personally I have my own implementation in CT

Does it uses more things than the API of NewStack ?

> Then I have a question. Did you get a case where we need to have something else than an orderedCollection for a Stack? To me it looks overkill for a stack to have such complex strategy stuff.

Yes, I talked about that with @MarcusDenker , we could have a Stack for which the growth is limited and would use an array in its strategy. Apparently, the compiler could profit of such stack.
So we have 3 possible strategies already. I think it makes the strategy design pattern worth it.

src/Collections-Stack/NewStack.class.st
juliendelplanque: Since the original stack was inheriting from LinkedList, I thought that it was a good idea to make the default strategy a linked list. But we can change that if needed.

src/Deprecated80/Stack.class.st
juliendelplanque: Well, it is not really pure as it is made of a strategy. I do not like the `New` prefix but the idea is to come back to `Stack` name in Pharo 9.

src/Collections-Stack/NewStack.class.st
juliendelplanque: We can add such methods next to it.
But as Stack inherits from Collection, it requires to implement `#removeAll`.

src/Collections-Stack/NewStack.class.st
juliendelplanque: Because it inherits from `Collection` still.

src/Graphics-Files/LzwGifDecoder.class.st
juliendelplanque: They are. :-)

src/Collections-Stack/NewStack.class.st
juliendelplanque: Because of inheritance from `Collection`.

Ducasse:
Stack should not inherit from Collection. 
Collection is too large. We should think in terms of API.

Ducasse:
Julien why don't we take my Stack implementation?
Simplicity has a value.

https://github.com/Ducasse/Containers-Stack/blob/master/src/Containers-Stack/CTStack.class.st

may be we can add a couple of iterators. 



<blockquote><div> GitHub</div><div><strong><a href="https://github.com/Ducasse/Containers-Stack">Ducasse/Containers-Stack</a></strong></div><div>A dead stupid stack implementation, but one fully working :) - Ducasse/Containers-Stack</div></blockquote>

Ducasse:
I do not think that we need the strategy and we should not inherit from Collection. 
So....

juliendelplanque:
For the inheritance from Collection, I don't know. There are parts of the API that can be useful with a stack. The iterators for example.

To solve the problem of "too big API" of Collection, we should not avoid inherit from it but clean Collection API... A Stack IS a Collection. Inheritance makes sense.

A way to reduce `Collection` API is to reify Iterators... As I started to do there: https://github.com/juliendelplanque/Iterators

<blockquote><div> GitHub</div><div><strong><a href="https://github.com/juliendelplanque/Iterators">juliendelplanque/Iterators</a></strong></div><div>Implements Iterators for Pharo's Collections. Contribute to juliendelplanque/Iterators development by creating an account on GitHub.</div></blockquote>

Ducasse:
Sorry but the argument Stack is a collection so it should inherit from collection is not valid. 
Let us not get trapped into this game. 
Let us be focus on what a stack should do. 

I will add do: collect: select: ... to my Stack implementation. 

We will not change Collection tomorrow for the sake of having a nicer Stack class. 
We should focus on impact and ROI.

Ducasse:
Julien I think that having internal and external control based iterators is a good solution. 
As a dev I can decide if I want to use an internal "iterator" or an explicit one like your library. 
We have the best of two worlds.

kasperosterbye:
I am somewhat scared about this. The issue with the stack revealing its implementation is but one small example of the strangeness of the smalltalk collection hierarchy. 

The Cook paper (http://citeseerx.ist.psu.edu/viewdoc/download?doi=10.1.1.24.7900&rep=rep1&type=pdf) gives a good overview of the problems as they were in 1992. I am not sure it has gotten cleaned up since then. But what has happened since then is that a lot of more code has written, a lot of it depending on its idiosyncrasies.
To see the most obvious examples of wrong design, you can merely look at all senders of  `shouldNotImplement` present in the collection hierarchy.

juliendelplanque:
> Sorry but the argument Stack is a collection so it should inherit from collection is not valid.
> Let us not get trapped into this game.
> Let us be focus on what a stack should do.
> I will add do: collect: select: ... to my Stack implementation.

I do not get why it is not valid.

The fact that we need to add such methods to the Stack is an argument to make it directly inherit from `Collection`.
Additionally to `do:`, `collect:`, `select:` and similar, you need to add:
- `#ifEmpty:`, `#ifNotEmpty:` and so one.
- `#allSatisfy:`, `#noneSatisfy:` and so one.
- conversion methods to convert Stack into another kind of collection.
- copy methods.
- probably more stuff.

Not inheriting from `Collection` will create a lot of code duplication.
Thus, even if this does not make the world perfect, inheriting from `Collection` is a good tradeoff.

We do not want to keep Collection and Stack APIs in sync by hand.
There is no tooling available to do that in Pharo as far as I know.

> We will not change Collection tomorrow for the sake of having a nicer Stack class.

I did not mean that. I think we need to simplify `Collection` API in general, not just in Stack context.

Ducasse:
No time to discuss and type: it is a question of API.

svenvc:
Indeed, IMO, a stack is a very simple data structure with a specific access protocol (push:, pop, isEmpty). Iteration (and everything that comes with it), has no place in it (except maybe in rare cases, it should not be made impossible to look at the elements).

API/protocol is much more important than inheritance/typing.

Ducasse:
Yes this is my point, else we can use an orderedCollection. 
Now we can really replace the current stack because this is a bad idea to inherit from linked-list.

PR #4751, opened 2019-09-27
Fix #4697. It defines the comments by using  fileName asFileReference…

### LINKED ISSUE #4697: BMPReadWriterTest references old API FileStream
BMPReadWriterTest has methods with input data having a comment referencing old FileStream API. Instructions should be updated to use the new stream API.
example:
```Smalltalk 
bmpData16bit
	"Created via:
		(Base64MimeConverter mimeEncode: 
			(FileStream readOnlyFileNamed: 'bmptest16b.bmp') binary)
				contents
	"
	^ 'Qk24AAAAAAAAADYAAAAoAAAACAAAAAgAAAABABAAAAAAAIIAAADDDgAAww4AAAAAAAAAAAAA
4APgA+AD4AMfAB8AHwAfAOAD4APgA+ADHwAfAB8AHwDgA+AD/3//f/9//38fAB8A4APgA/9/
/3//f/9/HwAfAAAAAAD/f/9//3//fwB8AHwAAAAA/3//f/9//38AfAB8AAAAAAAAAAAAfAB8
AHwAfAAAAAAAAAAAAHwAfAB8AHwAAA==' base64Decoded
```

sbragagnolo:
… binaryReadStream upToEnd base64Encoded . Still, there should be a better documentaiton than this.

### DIFF
--- a/src/Graphics-Tests/BMPReadWriterTest.class.st
+++ b/src/Graphics-Tests/BMPReadWriterTest.class.st
@@ -10,9 +10,7 @@ Class {
 { #category : #data }
 BMPReadWriterTest >> bmpData16bit [
 	"Created via:
-		(Base64MimeConverter mimeEncode: 
-			(FileStream readOnlyFileNamed: 'bmptest16b.bmp') binary)
-				contents
+		'bmptest16b.bmp' asFileReference binaryReadStream upToEnd base64Encoded 
 	"
 	^ 'Qk24AAAAAAAAADYAAAAoAAAACAAAAAgAAAABABAAAAAAAIIAAADDDgAAww4AAAAAAAAAAAAA
 4APgA+AD4AMfAB8AHwAfAOAD4APgA+ADHwAfAB8AHwDgA+AD/3//f/9//38fAB8A4APgA/9/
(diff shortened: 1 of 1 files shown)


MarcusDenker (approved):

akgrant43:
These are only comments, so it doesn't actually break anything, but:

```
'bmptest16b.bmp' asFileReference binaryReadStream upToEnd base64Encoded
```

is poor as it leaves the stream open.

```
'bmptest16b.bmp' asFileReference binaryReadStreamDo: [ :stream |
    stream upToEnd base64Encoded ]
```

will close the stream (code untested)

MarcusDenker:
ups... as it is merged already the best is to open a new issue tracker entry, then we do another pass

svenvc:
Actually I would implement a #binaryContents method on FileReference, then we would be able to write

    'bmptest16b.bmp' asFileReference binaryContents base64Encoded

with as implementation for #binaryContents

    self binaryReadStreamDo: [ :stream | stream upToEnd ]

although much faster implementations are possible if the size is known up front.

akgrant43:
Hi Sven,

Modifying ZnBufferedReadStream with:

```
upToEnd
	"Read elements until the stream is atEnd and return them as a collection."

	| streamSize result |

	"If the stream knows its size we can reduce overhead by allocating a buffer of the correct size"
	streamSize := [ self size ]
		on: Error
		do: [ 0 ].
	streamSize > 0 ifTrue:
		[ result := self next: (streamSize - self position) ].
	"Some streams may only have an estimate, so if not at the end, get the rest"
	self atEnd ifFalse:
		[ result := self collectionSpecies streamContents: [ :out | 
			result ifNotNil: [ out nextPutAll: result ].
			[ self atEnd ] whileFalse: [ 
				position > limit
					ifTrue: [ self nextBuffer ].	
				out next: limit - position + 1 putAll: buffer startingAt: position.
				position := limit + 1 ] ] ].
	^result
```

Takes advantage of streams that know their size, while still working with streams that don't (and streams that only have an estimate of their size).

akgrant43:
I'd be happy to submit a PR :-)

PR #4752, opened 2019-09-27
Making the ReadWrite buffered stream to be used only by SourceFile

tesonep:
- Moving the classes next to SourceFile
- Removing other users


svenvc:
Reading the code online, this looks OK

PR #4753, opened 2019-09-27
4514-Move-stable-slots-out-of-Example-package

### LINKED ISSUE #4514: Move stable slots out of Example package
Because people fear to use them as they are tagged as example.

Slots *ARE* working, we should promote them.

juliendelplanque:
Moved stable slots out of example group.
Adapted baseline.
Fixed dependencies.

Fixes #4514


juliendelplanque:
Crash seems to be random.

myroslavarm (approved):
looks ok to me

MarcusDenker (approved):

PR #4754, opened 2019-09-27
WIP (not yet ready for merging): first pass at Web platform support, via the SqueakJS VM

ccrraaiigg:
first pass at Web platform support (via the SqueakJS VM)

### DIFF
--- a/src/FileSystem-Path/AbsolutePath.class.st
+++ b/src/FileSystem-Path/AbsolutePath.class.st
@@ -27,7 +27,11 @@ AbsolutePath >> isAbsolute [
 
 { #category : #testing }
 AbsolutePath >> isRoot [
-	^ self size = 0
+	^self size = 0 or: [
+		JS
+			ifConnected: [
+				self size = 1 and: [(self at: 1) = 'SqueakJS']]
+			ifNotConnected: [false]]
 ]
 
 { #category : #printing }
--- a/src/Kernel/Object.class.st
+++ b/src/Kernel/Object.class.st
@@ -824,6 +824,7 @@ Object >> doesNotUnderstand: aMessage [
 	"Testing: (3 activeProcess)"
 		
 	| exception resumeValue |
+	JS ifConnected: [JS debugger].
 	(exception := MessageNotUnderstood new)
 		message: aMessage;
 		receiver: self.
--- a/src/System-BasicCommandLineHandler/BasicCommandLineHandler.class.st
+++ b/src/System-BasicCommandLineHandler/BasicCommandLineHandler.class.st
@@ -77,6 +77,8 @@ BasicCommandLineHandler class >> startUp: isImageStarting [
 	"only handle when lauching a new image"
 	isImageStarting ifFalse: [ ^ self ].
 
+	JS ifConnected: [ ^self ].
+	
 	Smalltalk session 
 		addDeferredStartupAction: [ self new activate ]
 ]
--- a/src/System-Support/VirtualMachine.class.st
+++ b/src/System-Support/VirtualMachine.class.st
@@ -372,6 +372,7 @@ VirtualMachine >> isRunningCogit [
 	"Returns true if we're running on the Cog JIT
 	 (vmParameterAt: 46 is the size of the machine code zone)"
 
+	JS ifConnected: [ ^false ].
 	^[(self parameterAt: 46) > 0] on: Error do:[:ex| ex return: false]
 ]
 
@@ -648,7 +649,7 @@ VirtualMachine >> parameterAt: parameterIndex [
 	71	the vmProxyMinorVersion (the interpreterProxy VM_MINOR_VERSION)"
 
 	<primitive: 254>
-	self primitiveFailed
+	^nil
 ]
 
 { #category : #parameters }
--- a/src/FileSystem-Core/WebResolver.class.st
+++ b/src/FileSystem-Core/WebResolver.class.st
@@ -1,0 +1,42 @@
+Class {
+	#name : #WebResolver,
+	#superclass : #PlatformResolver,
+	#category : #'FileSystem-Core-Base-Resolver'
+}
+
+{ #category : #accessing }
+WebResolver class >> platformName [
+	^ 'Web'
+]
+
+{ #category : #origins }
+WebResolver >> cache [
+	"Operating Systems often define standard locations for a personal cache directory. The cache directory is a user-specific non-essential (cached) place where data should be written."
+	^'/SqueakJS'
+]
+
+{ #category : #origins }
+WebResolver >> desktop [
+	^'/SqueakJS'
+]
+
+{ #category : #origins }
+WebResolver >> documents [
+	^'/SqueakJS'
+]
+
+{ #category : #origins }
+WebResolver >> home [
+	^'/SqueakJS'
+]
+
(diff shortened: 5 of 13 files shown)


MarcusDenker:
There seems to be some problem in the bootstrap:

 UndefinedObject(Object)>doesNotUnderstand: ifConnected:

as an endless loop

src/FileSystem-Path/AbsolutePath.class.st
```diff
@@ -27,7 +27,11 @@ AbsolutePath >> isAbsolute [
 
 { #category : #testing }
 AbsolutePath >> isRoot [
-	^ self size = 0
+	^self size = 0 or: [
+		JS
```
dionisiydk, line 31: That does not looks good. I think we should not have knowledge of JS environment in core classes.
Why "self size = 0" is not right condition for JS?

src/Kernel/Object.class.st
```diff
@@ -824,6 +824,7 @@ Object >> doesNotUnderstand: aMessage [
 	"Testing: (3 activeProcess)"
 		
 	| exception resumeValue |
+	JS ifConnected: [JS debugger].
```
dionisiydk, line 827: Why in case of JS we should forbid MessageNotUnderstood exceptions?
It will break any code like "[...] on: MessageNotUnderstood do: [...]" even if it is not a good practice.

dionisiydk: To be able inject native JS debugger you should implement a debugger.
Check debugger API by "Smalltalk tools debugger" senders

src/System-BasicCommandLineHandler/BasicCommandLineHandler.class.st
```diff
@@ -77,6 +77,8 @@ BasicCommandLineHandler class >> startUp: isImageStarting [
 	"only handle when lauching a new image"
 	isImageStarting ifFalse: [ ^ self ].
 
+	JS ifConnected: [ ^self ].
```
dionisiydk, line 80: You should implement different "default" command line handler for JS environment. (if special startup logic is really required).

src/System-Support/VirtualMachine.class.st
```diff
@@ -372,6 +372,7 @@ VirtualMachine >> isRunningCogit [
 	"Returns true if we're running on the Cog JIT
 	 (vmParameterAt: 46 is the size of the machine code zone)"
 
+	JS ifConnected: [ ^false ].
```
dionisiydk, line 375: Why JS VM could not setup correct value into parameter 46?

dionisiydk (changes requested):

noha:
Why is this a pull request? It seems like a bunch of hacks you usually do in your private copy, it touches central classes and introduces JS classes which cannot be part of the bootstrap. And the pull request is for pharo7. So this should live as a branch in the developer repo and surely not be integrated

src/FileSystem-Path/AbsolutePath.class.st
ccrraaiigg: If you want the system to come up on the Web VM, then there will be some decisions that must be delegated to the current host platform. To avoid mentioning any Web-related objects directly in the core classes, there should be additional obligatory protocol in the platform classes.

These changes are just meant to illustrate the codepoints where there must be additional collaboration with the host platform.

src/Kernel/Object.class.st
ccrraaiigg: This is not forbidding MessageNotUnderstood exceptions, only adding a JS-level halt as a side-effect along the way (e.g., something that will make the Chrome DevTools stop, in a resumable way). "<object>.debugger" is a common JS idiom; another would be to throw a particular JS exception.

src/System-BasicCommandLineHandler/BasicCommandLineHandler.class.st
ccrraaiigg: Agreed. The intent here is just to decline starting this facility if the host platform is Web. There should be some delegation to the host platform here.

src/System-Support/VirtualMachine.class.st
ccrraaiigg: Currently, the JS VM (when running a non-Cog image) doesn't have that many VM parameters. This can be faked, sure.

ccrraaiigg:
Oh, good morning to you too, Norbert. ;)  This is a pull request because Stéphane asked for a pull request of the changes I have made so far to get Pharo7 to start on the SqueakJS VM (which it indeed does). Yes, it's hacky, because there needs to be additional collaboration between the core classes and the platform classes, to support the new Web platform classes (WebPlatform, WebResolver, and WebStore).

I don't presume to decide for you what those core-classes changes should be, only to illustrate the hacks-in-progress that motivate them. It is through the review process of this pull request that we decide what those changes should be.

I did this work with Pharo7 because it is more stable, and I took the PharoLauncher's not working at all as a bad sign about starting from Pharo-latest. I leave it to interested folks to adapt whatever we decide to Pharo8 and beyond at this time.

Sheesh, you're welcome. I guess if you want to run Pharo in a web browser you can just go to https://caffeine.js.org/pharo and forget about all this coordinated development stuff.

<blockquote><div><strong><a href="https://caffeine.js.org/pharo">Caffeine :: livecode the web</a></strong></div></blockquote>

src/FileSystem-Path/AbsolutePath.class.st
dionisiydk: That's fine @ccrraaiigg . But still why isRoot needs modification? Do you have concrete example what was not working considering aPath? 
We should refactor FileSystem to address the found problem. So that WebFileSystem will be the only required part

src/Kernel/Object.class.st
dionisiydk: But #doesNotUnderstood:  is not the place where debugger is opening. 
MessageNotUnderstood exception is not different than any other error like NotFound for example.

Why JS.debugger can't be called during UnhandledException processing? And it was my proposal to implement own debugger and register it in the tools. It will just call JS level halt.

noha:
Thanks for explaining your rationale. In this case it is good to label the pull request with WIP or DONOTMERGE or something like that. Pull requests are meant to be merged and it should be obvious to the reviewer that this is just a draft.

src/FileSystem-Path/AbsolutePath.class.st
ccrraaiigg: Agreed. As for the root name size: at the moment, the filesystem that the SqueakJS VM creates from IndexedDB has a root hardwired with the name "/SqueakJS" instead of just "/". This is changeable.

src/Kernel/Object.class.st
ccrraaiigg: Ah, okay. Those both make sense. I do have a particular interest in halting in JS for unhandled MessageNotUnderstood exceptions, though.

ccrraaiigg:
Fair enough; thanks. I didn't get any hints about that during the sprint. I just assumed that pull-requests were the only medium for discussion of proposed changes, regardless of how polished they are.

Ducasse:
Craig what would be nice is that you split this PR into smaller ones and that we can integrate/improve them until they get in. I will close this PR because it will not be integrated as it is. 
S.

PR #4759, opened 2019-09-27
4459-Why-the-debugger-select-all-the-occurrences-of-a-message-

### LINKED ISSUE #4459: Why the debugger select all the occurrences of a message
Because the selection makes more difficult to spot the actual sender.

MarcusDenker:
fixes #4459

### DIFF
--- a/src/AST-Core/RBMessageNode.class.st
+++ b/src/AST-Core/RBMessageNode.class.st
@@ -115,12 +115,6 @@ RBMessageNode >> copyInContext: aDictionary [
 ]
 
 { #category : #accessing }
-RBMessageNode >> debugHighlightStart [
-
-	^ self  keywordsPositions first
-]
-
-{ #category : #accessing }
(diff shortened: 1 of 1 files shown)


dionisiydk:
It does not looks good.
Try debug second message in cascade

MarcusDenker:
we could just detect if we are in a cascade:

debugHighlightStart
	^ parent isCascade
		ifTrue: [ self keywordsPositions first ]
		ifFalse: [ super debugHighlightStart ]

I will make a new PR

dionisiydk:
But in case of cascade it will have original issue. No?

пн, 30 сент. 2019 г., 13:21 Marcus Denker <notifications@github.com>:

MarcusDenker:
Yes, but there it makes sense to just highlight the selector...  yes, it will then show the other substrings, but it will be an improvement.

We need to have a way to highlight without the editor hightlighting similar strings in addition...

PR #4771, opened 2019-09-30
fix all senders #namedTempAt:

### LINKED ISSUE #4768: do not use #namedTempAt: and #namedTempAt:put:
#namedTempAt: and ##namedTempAt:put: are quite strange APIs: the offset it a virtual offset in all vars that a user could type and expect a value when in the debugger. 

I checked and all users actually do have the name they want to read.

We can thus rewrite all the clients to use #tempNamed: , it even will be faster, as we do not
need to calculate that array of virtual names...

MarcusDenker:
- change all senders of #namedTempAt: to use #tempNamed: instead
- search all source and found some comments to fix for that, too

- some rewrites in Spec due to running the tests

fixes #4768

### DIFF
--- a/src/Debugging-Core/Context.extension.st
+++ b/src/Debugging-Core/Context.extension.st
@@ -428,10 +428,11 @@ Context >> tempsAndValues [
 	"Return a string of the temporary variabls and their current values"
 	
 	^ String streamContents: [ :aStream |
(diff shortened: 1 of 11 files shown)


myroslavarm (approved):
looks good!

PR #4774, opened 2019-09-30
4769-debuggerMap-move-implementation-to-Context

### LINKED ISSUE #4769: #debuggerMap: move implementation to Context
DebuggerMethodMapOpal has methods that should be implemented on context (our "Map" is the AST).

-> move implementation to the methods on Context that now call #debuggerMap.

This is independent of the question if/when to deprecate (there is another issue for that)

MarcusDenker:
- move all the implementation of  DebuggerMethodMapOpal to Context
- add some comments

fixes #4769


MarcusDenker:
there are some typos in the comments and they need to be improved, but good as a first step

PR #4775, opened 2019-09-30
Deprecation should be optional and should be able to be disabled

tesonep:
Or just being disabable


jecisc:
By "Deprecation" you mean "Transformation"?

MarcusDenker (approved):
Yes, makes sense to be able to turn off rewrites. code looks ok (but did not test)

tesonep:
Yes, the idea is to generate the warning and do not perform the transformation in a deprecation.

PR #4776, opened 2019-09-30
Another Smalltalk globals removal

Alisu:
This time in OpalCompiler and Monticello.


MarcusDenker:
Breaks bootstrap

MarcusDenker:
I would make sampler PRs. e.g. base rule: *never* mix a change in the compiler with any other change. Compiler changes are hard to get right.

Alisu:
Indeed, i think i will revert the changes in the compiler.

MarcusDenker:
I would not do any PR for this that changes two different packages at the same time.

CafeKrem (approved):

Alisu:
Failure

PR #4777, opened 2019-09-30
2312-ondo-contexts-added-on-step-through-not-removed

MarcusDenker:
better comment


myroslavarm (changes requested):
this makes sense although the second comment is a bit confusing:
"Error was *not* raised we need to it remove inserted above contexts then return signaler context"
was it supposed to be "we need to remove" without "it"?

myroslavarm (approved):
good!

MarcusDenker:
Let's try it, we can always revert

PR #4784, opened 2019-10-01
Better support for literals in UFFI callouts

### LINKED ISSUE #4695: Enhance UFFI literals
UFFI literals support only numbers so far (e.g., integers)
And the only types that can be specified are ints, not floats, strings or user specified data types.

guillep:
Fix #4695: Support for more literals (such as floats, strings, bytearrays).

- base function parser on RB parser to be able to parse any literal object and not only integers
- introduce types for literal objects (e.g. `float 2.71` is now a valid argument) 
- split function parsing and resolution
- have two resolution modes
  - *strict* does not accept literals without types
  - *non strict* for backwards compatibility (and default for now), works as before but is dangerous and can introduce marshalling bugs (as before)
- make management of class variables explicit (before they were resolved as types and was not so coherent...)


MarcusDenker:
failing test: testSetUpMethodInSUnitTestsNeedsToBeInRunningProtocol

Ducasse:
This is cool to see all these enh!

Ducasse:
what is the strict mode?

guillep:
> failing test: testSetUpMethodInSUnitTestsNeedsToBeInRunningProtocol

I'll fix it, that is easy :)

guillep:
> what is the strict mode?

The strict mode is a mode for uFFI that will reject with an exception callouts that do not present type information. For example, let's say you have
```smalltalk
myMethod
  self ffiCall: #( void function( 42 )  )
```
The non-strict version of the parser will accept the 42 and assume it's an uint32. However, that type guessing may not be correct, since `function` could be defined otherwise.

The strict mode (optional for now) will force users to declare a type for the `42` literal. Meaning they should chose a type and declare it as follows

```smalltalk
myMethod
  self ffiCall: #( void function( **int32** 42 )  )
```

To specify the strict mode, there are several ways:

 - per callout: `self ffiCall: #( ... ) options: #( + optStrict )
 - per library, shared to all its callouts: `MyLibrary>>options ^  #( + optStrict ) `

PR #4803, opened 2019-10-03
4211 easy support for equality and hash with nice default behavior

bouraqadi:
Equality tests are in a separate package now


Ducasse:
Tx noury!

estebanlm:
This is nice to have but it will need to be for Pharo9.0

estebanlm (changes requested):
I think classes like "Fruit" and "Apple" need to have a prefix, they are too generic to add them to base image just as it.

bouraqadi:
I just did a new commit to with the requested changes

tesonep:
I have a small change, can we take out the extension methods on Object and ProtoObject and put them in the class side of the trait? is it possible or I am just not seeing a part?

bouraqadi:
Pablo, `allInstVarNamesForEqualityComparison` makes recursive calls to the superclass. This means that all classes should understand it. Moving it to the trait, would break recursing, and make only classes which use the trait understand this message.

Moreover, I'd like at some point to get rid of the trait. We need a better default `=` implmentation on `Object`. Unfortunately, we cannot do the change currently. It freezes the whole image. Hopefully, after the PhD of Carolina, we'll have some tools that would help with this regard.

estebanlm (approved):

estebanlm:
so... what we do with this?
merge? is... old :P

Ducasse:
well... why not.

estebanlm:
well, because this is a feature and according our rules, we are in freeze status :)
Nevertheless, in this case I am willing to make an exception because this is very old and we already made it wait for one complete release without integrating it...

demarey:
I rebased this PR on top of Pharo10 branch + added some comments + turn examples into tests

PR #4832, opened 2019-10-04
Fix #2853. When removing a class, also log its contained methods

### LINKED ISSUE #2853: Epicea: Revert class deletion does not reveret method deletion
Today I did a wrong shortcut and ended up deleting a package. I wanted to use Epicea to revert that, but it only reverted the classes deletion. The classes then contained no method :(

tinchodias:
I propose this fix/workaround to tackle this issue.

### DIFF
--- a/src/Epicea-Tests/EpCodeChangeIntegrationTest.class.st
+++ b/src/Epicea-Tests/EpCodeChangeIntegrationTest.class.st
@@ -125,6 +125,22 @@ EpCodeChangeIntegrationTest >> testClassRemoval [
 ]
 
 { #category : #tests }
+EpCodeChangeIntegrationTest >> testClassRemovalWithMethods [
+
+	| events category |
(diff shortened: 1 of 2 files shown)


tinchodias:
Included a test, but you can also evaluate this code and see the result in the Code Changes browser:
~~~
clazz := Object subclass: #Clazz
	instanceVariableNames: ''
	classVariableNames: ''
	poolDictionaries: ''
	package: 'Martin'.

1 to: 7 do: [:index |
	clazz compile: 'inst', index asString.
	clazz classSide compile: 'class', index asString ].
clazz removeFromSystem.
~~~

tinchodias:
Epicea supports nesting events (e.g. in refactorings), so it could be an improvement to this fix to nest al method removals inside the class removal. This can be reported as another issue.

MarcusDenker (approved):

PR #4835, opened 2019-10-05
Move the #errorNotKeyed message from Collection to Bag

### LINKED ISSUE #4808: Collection>>#errorNotKeyed is only sent from Bag class
So it should not be implemented in `Collection`.

leonardoce:
The #errorNotKeyed message is sent only by the Bag class, so it shouldn't belong to che Collection class.

Fixes: #4808

### DIFF
--- a/src/Collections-Abstract/Collection.class.st
+++ b/src/Collections-Abstract/Collection.class.st
@@ -653,13 +653,6 @@ Collection >> errorNotFound: anObject [
 ]
 
 { #category : #private }
-Collection >> errorNotKeyed [
-
-	self error: ('Instances of {1} do not respond to keyed accessing messages.' format: {self class name})
-
-]
-
-{ #category : #private }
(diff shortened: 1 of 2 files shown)


Ducasse:
Thanks!!!
To me it means that Bag should not have these stupid at: and at:put: methods. 
We should remove them. They make no sense. If we start to cancel all the method that are inherited from object we are doomed. And at: is not defined on Collection. So let us remove the error and the at: and at:put:. 
Can you change your PR accordingly?

S.

jecisc (approved):

leonardoce:
Agree with @Ducasse. I removed the methods and modified my commit accordingly.
Thank @jecisc and @Ducasse

Ducasse:
Thanks Leonard. This is cool.

Ducasse:
I like the idea that we are improving one step at a time.

PR #4840, opened 2019-10-05
Commands for Iceberg to browse repository with Calypso

dionisiydk:
It extends Iceberg browsers with a commands to browse repository with Calypso (based on new Ring2) :
- on selected repository

iceRepoBrowser1

- on selected package

iceRepoBrowser2

- on selected commit in repository history.

iceRepoBrowser3

Commands are defined in existing Calypso plugin which provide Iceberg integration. Now it also extends the Iceberg itself.


dionisiydk:
Now I do not understand the issue with dependency. The modified Calypso package already referenced IceRepository. Are Iceberg commands loaded in another step? (after Calypso).
 
So it needs to be redone differently. What the proper way? Can Iceberg UI depends on Calypso?

pavel-krivanek (approved):
the build artifact works well

Ducasse:
Thanks you guys!!

PR #4842, opened 2019-10-06
Remove `bounds` ivar from AthensBalloonPath

### LINKED ISSUE #4549: AthensBalloonPath has unused ivar bounds
bounds is not used (only assigned one value) --> can be removed

leonardoce:
That ivar was not used so this patch removes it.

Fixes: #4549


MarcusDenker (approved):
ok

PR #4845, opened 2019-10-07
4837-Typing-Point-in-Transcript-raises-a-DNU-on-completion

### LINKED ISSUE #4837: Typing Point in Transcript raises a DNU on completion
Hello myroslava

When I type a class name in the transcript I get an error in CompletionController. 
Because Transcript does not understand selectedClassOrMetaClass
I imagine that you will have the same in old workspace.

MarcusDenker:
fixes #4837

### DIFF
--- a/src/Transcript-Core/ThreadSafeTranscript.class.st
+++ b/src/Transcript-Core/ThreadSafeTranscript.class.st
@@ -194,6 +194,11 @@ ThreadSafeTranscript >> printOn: aStream [
 	aStream nextPutAll: 'Transcript'
 ]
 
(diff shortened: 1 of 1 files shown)


Ducasse:
It would be nice to package this as extensions of the completion engine?

MarcusDenker:
We need to clean this up. There is a huge mess in these methods: #selectedBehavior #selectedClass, #selectedClassOrMetaClass. 

One of these is needed to evaluate, too. See #modelCurrentSelectedClass in the editor...

I propose to merge this as cleaning this up is another thing. (I put it on the backlog)

PR #4848, opened 2019-10-07
add-shortcut-printer

estebanlm:
adding a generic shortcut printer (needed to properly print menus for different platforms and backends).


Ducasse:
esteban this is cool. Now having some tests would be nice too. 
Because one day we will break it.

estebanlm:
not finished :)

Ducasse:
:) but going to bed now :)

estebanlm:
now is good to integration.

Ducasse:
You see 
ToggleMenuItemShortcut >> color [

 	^ color ifNil: [ Smalltalk ui theme ghostTextColor ]
 ]

I think that this guy should either be a subclass of displayableObject. 
We should stop adding Smalltalk ui theme and other like that in the code.

estebanlm:
you ask it, you have it

Ducasse:
Ok cool now I do not understand why I still see 

ToggleMenuItemShortcut >> color [

 	^ color ifNil: [ Smalltalk ui theme ghostTextColor ]
 ]
It should probably be something like [self theme ghostTextColor] no?

estebanlm:
is just that my network was failing and the push didn't arrive until a bit later.

Ducasse:
oki

PR #4851, opened 2019-10-07
use #lookupVar: instead of #lookupVarForDeclaration:

MarcusDenker:
No idea why we have #lookupVar: and #lookupVarForDeclaration. Opal tests are green without it


tesonep:
@MarcusDenker this fix is not allowing the playground to evaluate code using a block with temporaries or parameters

MarcusDenker:
Ah... I will revert and check

PR #4854, opened 2019-10-07
Add new #printOneLineString and related methods

### LINKED ISSUE #4849: Introduce Object>>printOneLineString
The idea is that we should be able to have printOn: but that replaces cr and lf by space
and probably limited to a certain amount of character. 

printOneLineString
    should invoke printOnOneLine: 
    on a special Stream class (that interprets cr and lf as space).

By default printOnOneLine: should delegate to printOn: so that we get for free all 
the printOn:

leonardoce:
It's nice to have a #printOneLineString that is just like #printString
but avoids printing CRs and LFs. This patch implements such idea adding
a method named #printOnOneLine:, working on a Stream, and a
FilteringStream.

The FilteringStream is used to replace CRs and LFs with a space, but can
be used to filter elements going into a Stream via a generic block.

Fixes: #4849

### DIFF
--- a/src/Collections-Streams/FilteringStream.class.st
+++ b/src/Collections-Streams/FilteringStream.class.st
@@ -1,0 +1,44 @@
+"
+I am a basic filtering stream that will wrap every object put in the stream in a block, before sending it to my decoree.
+
+"
+Class {
+	#name : #FilteringStream,
+	#superclass : #DecoratorStream,
+	#instVars : [
+		'filter'
+	],
+	#category : #'Collections-Streams-Base'
+}
+
+{ #category : #'instance creation' }
+FilteringStream class >> on: aStream [
+	self error: 'FilteringStreams are created with #on:with:'
+]
+
+{ #category : #'instance creation' }
+FilteringStream class >> on: aStream with: aBlock [
+	^ self basicNew
+		on: aStream;
+		filter: aBlock;
+		yourself.
+]
+
+{ #category : #accessing }
+FilteringStream >> filter [
+	^ filter
+]
+
+{ #category : #accessing }
+FilteringStream >> filter: anObject [
+	filter := anObject
+]
+
+{ #category : #accessing }
+FilteringStream >> nextPut: anObject [ 
+	"Insert the argument, anObject, as the next object accessible by the 
+	receiver. Answer anObject."
+	
+	super nextPut: (filter value: anObject).
+	^anObject 
+]
--- a/src/Collections-Streams/FilteringStreamTest.class.st
+++ b/src/Collections-Streams/FilteringStreamTest.class.st
@@ -1,0 +1,26 @@
+"
+This is the unit test for the class FilteringStream
+"
+Class {
+	#name : #FilteringStreamTest,
+	#superclass : #TestCase,
+	#category : #'Collections-Streams-Tests'
+}
+
+{ #category : #tests }
+FilteringStreamTest >> testWithCharacters [
+	|f|
+	f := FilteringStream on: (String new writeStream) with: [:x | x].
+	f nextPut: $a.
+	self assert: f contents size equals: 1. 
+	self assert: (f contents at:1) equals: $a.
+]
+
+{ #category : #tests }
+FilteringStreamTest >> testWithNumbers [
+	|f|
+	f := FilteringStream on: (WriteStream with: LinkedList new) with: [:x | x + 10].
+	f nextPut: 23.
+	self assert: f contents size equals: 1. 
+	self assert: (f contents at:1) equals: 33.
+]
--- a/src/Kernel-Tests/ObjectTest.class.st
+++ b/src/Kernel-Tests/ObjectTest.class.st
@@ -103,6 +103,31 @@ ObjectTest >> testPrintLimitedString [
 ]
 
 { #category : #'tests-printing' }
+ObjectTest >> testPrintOnOneLine [
+	|ws|
+	ws := String new writeStream.
+	'this is a test' printOnOneLine: ws.
+	self assert: ws contents equals: '''this is a test'''.
+
+]
+
+{ #category : #'tests-printing' }
+ObjectTest >> testPrintOnOneLineString [
+	self 
+		assert: ('this is a test' printOnOneLineString) 
+		equals: '''this is a test'''.
+	
+	self 
+		assert: ('this is a', Character cr asString, 'test') printOnOneLineString 
+		equals: '''this is a test'''.
+	
+	self 
+		assert: ('this is a', Character lf asString, 'test') printOnOneLineString 
+		equals: '''this is a test'''.
+
+]
+
+{ #category : #'tests-printing' }
 ObjectTest >> testPrintString [
 
 	self assert: Object new printString equals: 'an Object'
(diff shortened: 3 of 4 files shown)


Ducasse:
Nice :) You were fast. And with tests. Excellent. I could not do it better :)
BTW I think that your tests do not cover the case for lf (or I did not see it) because Pharo internally is using cr (unfortunately).

Ducasse:
Rereading the code I have some suggestion: 

- first 
=========================================================
Object >> printOnOneLine: aStream [
 	"Append to the argument, aStream, a sequence of characters that  
 	identifies the receiver, while replacing cr and lf with one space"

 	|filter spaces|
 	spaces := Set with: Character cr with: Character lf.
 	filter := FilteringStream on: aStream with: [ :c | 
 		(spaces includes: c) ifTrue:Character space ifFalse:c ].
 	self printOn: filter.
 ]

I would write this way because ifTrue:ifFalse: requires blocks. 

Object >> printOnOneLine: aStream [
 	"Append to the argument, aStream, a sequence of characters that  
 	identifies the receiver, while replacing cr and lf with one space"

 	| filter spaces |
 	spaces := { Character cr . Character lf}. 
 	filter := FilteringStream on: aStream with: [ :c | 
 		(spaces includes: c) ifTrue: [ Character space ] ifFalse: [c] ].
 	self printOn: filter.
 ]

2 
====================================
With your implementation subclasses are forced to do a super printOnOneLine:

I would better do it like

printOnOneLineString [
 	"Answer a String whose characters are a description of the receiver, 
 	avoiding printing cr and lf"

 	| filter spaces |
 	spaces := { Character cr . Character lf}. 
 	filter := FilteringStream on: aStream with: [ :c | 
 		(spaces includes: c) ifTrue: [ Character space ] ifFalse: [c] ].
 	self printOnOneLine: filter.
        ^ filter contents


Object >> printOnOneLine: aStream [
 	"Append to the argument, aStream, a sequence of characters that  
 	identifies the receiver, while replacing cr and lf with one space"

       self printOn: aStream

This way the extended can only subclass printOn: and if needed printOnOneLine:
but he is not forced to do a super call.

What do you think?

leonardoce:
You too were really fast and helpful with your review! Thank you for that.

> [...]
> BTW I think that your tests do not cover the case for lf (or I did not see it) because Pharo internally is using cr (unfortunately).

Oh, I didn't realize that. I'll cover also lf in the tests and modify my commit accordingly.

> Rereading the code I have some suggestion:
>     * first
> [...] I would write this way because ifTrue:ifFalse: requires blocks. [...]

Totally agree with that, I'll modify my patch.

> # 2
> 
> [...] 
> Object >> printOnOneLine: aStream [
> "Append to the argument, aStream, a sequence of characters that
> identifies the receiver, while replacing cr and lf with one space"
> 
> ```
>    self printOn: aStream
> ```
> 
> This way the extended can only subclass printOn: and if needed printOnOneLine:
> but he is not forced to do a super call.

I'm surely wrong, but at first glance with your proposal `#printOnOneLine:` and `#printOn:` have exactly the same behavior, and that means that `#printOnOneLine:` wouldn't replace cr and lf with spaces.

What I misunderstood? :) Thanks!

Ducasse:
My idea is that printOnOneLineString is the template method and printOnOneLineOn: is a hook (whose default is just printOn: with the eating new line stream). 

printOnOneLineOn: will not have the same behavior since it is using a different stream than printOn:. But what is beautiful with this design is that we can still fully reuse the logic of the printOn: for free (we just changed the behavior of the stream but reuse the logic of printing). 

This is super nice.

Ducasse:
Is it better?
Julien proposed the idea and I love it, so nice.

leonardoce:
Now I understand! That's a good idea.
I'll modify my patch including your considerations.

Ducasse:
This looks good to me.

Ducasse:
I will give a chance to others to have a look.

leonardoce:
Thank you @Ducasse!

Ducasse:
BTW I love your comments!
Everybody should learn from them!. 
I entered other issues related to printString and printOn: thinking that you could be interested.

leonardoce:
@Ducasse, you are too kind to me. I'm reading the issues you created.

MarcusDenker:
The build fails after this was integrated.

MarcusDenker:
I think we need to revert this for now

MarcusDenker:
I create a revert PR. I do not yet understand it... after this PR was merged, the main build failed, with something related to printing. But it looks strange.


'Errors in script loaded from /builds/workspace/est_and_branch_Pipeline_Pharo8.0/32/bootstrap/scripts/bootstrap.st'
[32] MessageNotUnderstood: receiver of "definition" is nil
[32] UndefinedObject(Object)>>doesNotUnderstand: #definition
[32] PBSpurClassLoader(PBClassLoader)>>executeDefinitionFor:
[32] PBSpurClassLoader(PBClassLoader)>>createBehaviorFromDefinition:
[32] PBImageBuilderSpur5032bit(PBImageBuilder50)>>createBehaviorFromDefinition:
[32] [ ^ super createBehaviorFromDefinition: aRFiDefinition ] in PBImageBuilderSpur5032bit(PBImageBuilderSpur50)>>createBehaviorFromDefinition: in Block: [ ^ super createBehaviorFromDefinition: aRFiDefini...etc...
[32] BlockClosure>>on:do:
[32] PBImageBuilderSpur5032bit(PBImageBuilderSpur50)>>createBehaviorFromDefinition:
[32] EPASTInterpreter>>loadClass:
[32] EPASTClassLoaderVisitor>>visitVariableNode:
[32] RBVariableNode>>acceptVisitor:
[32] EPASTInterpreter>>visitClassDefinition:
[32] EPASTClassDefinition>>acceptVisitor:
[32] EPASTInterpreter(ASTInterpreter)>>interpret:
[32] EPASTInterpreter>>evaluateAST:
[32] EPASTEvaluator>>evaluateAST:
[32] EPASTEvaluator>>evaluateCode:withTemps:
[32] EPASTEvaluator>>evaluateCode:
[32] PBSpurClassLoader(PBClassLoader)>>executeDefinitionFor:
[32] PBSpurClassLoader(PBClassLoader)>>createBehaviorFromDefinition:

MarcusDenker:
[32] *** Warning: Deprecation: The method Object>>#name called from PBClassLoader>>#createBehaviorFromDefinition: has been deprecated.
[32] Implement your own domain representation of an object, or use #asString or #printString instead.

MarcusDenker:
I think the reason is that the Test case is part of the Streams package, so the boo trap loads it but the superclass (TestCase) is not there.

The bootstrap works again after reverting.

Ducasse:
What I do not understand is why the builds worked in the first place. I will have a look

Ducasse:
I'm confused. Now the status says merged but I would like to get this issue still open. 
Because it is not integrated.

Ducasse:
@tesonep I would really like to sit with you to 
- understand why the validation did not bark before
- understand how I can reopen this issue :)
- fix it :)

Ducasse:
I will package this behavior in a little package that can loaded separatedly from the bootstrap.

Issue #4843, opened 2019-10-07
Semantic Analysis: add a temp variable for the Temp Vector Copied Var


MarcusDenker:
Copied temps are copies of "normal" temp variable that are used to copy a value from the definition context to the use (if the use is just a read).

As such there is a "originalVar" 

The temp vector is made accessible in sub contexts as a copied var, but the "originalVar"  is
nil

--> we should have a temp for the temp vector. This will make lots of things easier (e.g. finding the definition scope for any temp vector copied var will be just "originalVar scope").

PR #4857, opened 2019-10-07
Simplify TempVector Offset from IR

MarcusDenker:
simplify getting the index from the temp vector from the IR
  - make sure that we ask the right scope
  - simplify the code on the IR Level

### DIFF
--- a/src/OpalCompiler-Core/IRMethod.class.st
+++ b/src/OpalCompiler-Core/IRMethod.class.st
@@ -399,6 +399,5 @@ IRMethod >> tempMap [
 
 { #category : #enumerating }
 IRMethod >> tempVectorNamed: aName [
-	
-	^self firstInstructionMatching: [:ir | ir isTempVector and:  [ir name = aName ] ]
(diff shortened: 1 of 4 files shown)


MarcusDenker:
The idea behind this change: #indexInTempVectorFromIR: was always asking the method scope. This means that # tempVectorNamed: there was implemented to search *the whole* IR tree for the temp vector.

Now we instead get first the definition scope of the vector (using #tempVectorDefinitionScopeForTempStoringIt) and then ask *that* scope, There we can be 100% sure that a temp vector is defined in the IR sequence.

This in turn then allows us to radically simplify tempVectorNamed: to just call it on the sequence.

Lots of checking code can be removed (e.g. hasTempVector:, #blockHasTempVector: and so on).

The step after this is to simplify tempVectorDefinitionScopeForTempStoringIt, for that we have issue https://github.com/pharo-project/pharo/issues/4843 

(instead of calling #tempVectorDefinitionScopeForTempStoringIt we then can just do "originalVar scope")

PR #4858, opened 2019-10-08
Opal: compiler shoud not copy AST for pluging

MarcusDenker:
the compiler does not need to copy the AST before calling plugins as they can do that if needed. Speeds up compilation by 3-5%


MarcusDenker:
recompile of whole image is 45 seconds before, 42 seconds after

PR #4864, opened 2019-10-08
Add a browse menu item

### LINKED ISSUE #4839: We need to have a Browse menu item
that given any object open the class of the object. 
I'm often doing inspect + self class.

leonardoce:
The new menu item will evaluate the current selection and then open its
class. It can be used in the system browser and also in the Playground.

Fixes: #4839


leonardoce:
I hope to have understood correctly the request by @Ducasse.
In this moment, I haven't bound any shortcut to the new action.

This is the look of the contextual menu of the playground after the patch:

Screenshot from 2019-10-08 22-18-25

And this is what can be seen from the browser:

Screenshot from 2019-10-08 22-18-50

Ducasse:
This is cool. 
Thanks I will review it tomorrow.

estebanlm:
there is already a "browse it", under "code search" submenu (along with "senders" and "implementors".
Why do we need this new one?

Ducasse:
I extended super simply browse to handle 
Point browse
10@10 browse
No need for a new entry and now we can select an expression or self!!!!

PR #4865, opened 2019-10-09
Improve #printString implementations

### LINKED ISSUE #4855: printString reimplementations look suspicious
We should revisit the printString implementations because it looks like some of them are simply printOn: disguised. 

In addition some of them promote bad pattern:

MalGraphEdge >> printString

	^ self from printString, ' -> ', self to printString

Why do we create two substreams instead of sharing the printString via printOn:?
Here printString each time create a stream then .... it is worse , create spurious copies...
Super super ugly. 

In general printOn: in general should not call printString, better use self print: str
So may be we can open another issue for this issue.

leonardoce:
On MalGraphEdge, MalHitsNode and MalWeightedEdge, the #printString
implementation was calling #printString on their ivars, causing the
creation of too many temporary Streams and spurious copies.

This patch reimplement those methods in a more performance-aware way,
while adding some unit tests that were running also before the
implementation to be modified, avoiding changing the former behavior of
the methods.

Fixes: #4855

### DIFF
--- a/src/Moose-Algos-Graph/MalHitsNode.class.st
+++ b/src/Moose-Algos-Graph/MalHitsNode.class.st
@@ -35,13 +35,13 @@ MalHitsNode >> label [
 ]
 
 { #category : #printing }
-MalHitsNode >> printString [
+MalHitsNode >> printOn: aStream [
+	aStream nextPut: $(.
+	self model printOn: aStream.
+	aStream space.
+	self auth printOn: aStream showingDecimalPlaces: 2.
+	aStream space.
+	self hub printOn: aStream showingDecimalPlaces: 2.
+	aStream nextPutAll: ')'.
 
-	^ 	'(',
-		self model printString,
-		' ',
-		(self auth printShowingDecimalPlaces: 2),
-		' ',
-		(self hub printShowingDecimalPlaces: 2),
-		') '
 ]
--- a/src/Moose-Algos-Graph/MalGraphEdge.class.st
+++ b/src/Moose-Algos-Graph/MalGraphEdge.class.st
@@ -60,12 +60,6 @@ MalGraphEdge >> printOn: aStream [
 ]
 
 { #category : #accessing }
-MalGraphEdge >> printString [
-
-	^ self from printString, ' -> ', self to printString
-]
-
-{ #category : #accessing }
 MalGraphEdge >> to [
 	^ to
 ]
--- a/src/Moose-Algos-Graph/MalGraphEdgeTest.class.st
+++ b/src/Moose-Algos-Graph/MalGraphEdgeTest.class.st
@@ -1,0 +1,12 @@
+Class {
+	#name : #MalGraphEdgeTest,
+	#superclass : #TestCase,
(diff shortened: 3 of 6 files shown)


VincentBlondeau:
Good job. But could you implement the printOn: aStream method instead of printString ? (and then the super class printString can be used instead of subclassing it)

leonardoce:
> Good job. But could you implement the printOn: aStream method instead of printString ? (and then the super class printString can be used instead of subclassing it)

Yes, it's a good idea. Let me modify the patch according to your suggestion.

leonardoce:
@VincentBlondeau I modified my patch considering your advices. Thanks for them!

src/Moose-Algos-Graph/MalHitsNode.class.st
```diff
@@ -35,13 +35,13 @@ MalHitsNode >> label [
...
-		self model printString,
-		' ',
-		(self auth printShowingDecimalPlaces: 2),
-		' ',
-		(self hub printShowingDecimalPlaces: 2),
-		') '
+MalHitsNode >> printOn: aStream [
+	aStream nextPut: $(.
+	self model printOn: aStream.
+	aStream nextPut: Character space.
```
VincentBlondeau: aStream space.

src/Moose-Algos-Graph/MalHitsNode.class.st
```diff
@@ -35,13 +35,13 @@ MalHitsNode >> label [
...
+	self auth printOn: aStream showingDecimalPlaces: 2.
+	aStream nextPut: Character space.
```
VincentBlondeau: Same here

src/Moose-Algos-Graph/MalHitsNode.class.st
```diff
@@ -35,13 +35,13 @@ MalHitsNode >> label [
...
+	self hub printOn: aStream showingDecimalPlaces: 2.
+	aStream nextPutAll: ') '.
```
VincentBlondeau: No space at the end of the printing

VincentBlondeau (changes requested):
Thanks for the changes! Still some to do and it is ok for me !

src/Moose-Algos-Graph/MalHitsNode.class.st
leonardoce: This is going to change the former behavior.
I'm removing the space at the end (from tests too) because I also think it's better to avoid it.
Let me know if this is ok with you.

leonardoce:
@VincentBlondeau I've set you as co-author of my commit. If that's not ok with you, please feel free to remove the line from the commit message. Thanks!

src/Moose-Algos-Graph/MalHitsNode.class.st
VincentBlondeau: The printOn: should be used only for development purposes, so it is ok to change. Clients should have to add a space to fix the issue. But they can also put a CR if they want

VincentBlondeau (approved):

PR #4867, opened 2019-10-09
[DO NOT MERGE] Adding an additional instance variable to hold an additional methodDict

tesonep:
This is an experiment we are doing with Theo.


estebanlm:
ok, reopen when need it again :)

PR #4877, opened 2019-10-10
Integrate-ObservableSlot-as-part-of-VariablesLibrary

jecisc:
Introduce ObservableSlot as part of VariableLibrary. 

This slot is a copy of the spec SpObservableSlot and will replace this one.

This slot comes with tests and I improved the documentation that was started in Spec.


jecisc:
Tests passed \0/

PR #4882, opened 2019-10-10
4881-RBImplementedNotSentRule-is-disabled

### LINKED ISSUE #4881: RBImplementedNotSentRule is disabled
RBImplementedNotSentRule was disabled. It is quite useful, if it is too slow we should speed it up more.

MarcusDenker:
- enable RBImplementedNotSentRule
- document #excludedRuleNames
- remove unsent message
- improve rule a little

fixes #4881

### DIFF
--- a/src/GeneralRules/RBImplementedNotSentRule.class.st
+++ b/src/GeneralRules/RBImplementedNotSentRule.class.st
@@ -35,17 +35,6 @@ RBImplementedNotSentRule class >> allMessages [
 	^ allMessages
 ]
 
-{ #category : #accessing }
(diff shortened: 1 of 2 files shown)


Ducasse (commented):
Smalltalk globals -> self class environment?

MarcusDenker:
for now it would work with "self class environment".  But that is the environment the tool is in, while what we conceptional want is "please analyse this environment", something that we do not model right now.

For me then using "Smalltalk globals" does not feel that bad  because it makes clear: "analyse the global environment, if you want something else there is still work todo", while "self class environment" uses explicitly the environment the tool is installed in as if this is what we want. And it is not.

PR #4885, opened 2019-10-10
4883-FileSystemResolver-should-let-the-user-add-its-own-origins

jecisc:
Use pragma to collect the supported origins of the file system resolvers. 

This allows one to add its own origins to the resolvers.


jecisc:
Tests passed.

MarcusDenker (approved):

PR #4890, opened 2019-10-11
Add-command-to-browse-rule-class

jecisc:
Add a command to Calypso to browse a critic class in the extra menu.


MarcusDenker (approved):

dionisiydk (commented):
It should be a subclass of ClyCritiqueCommand. It provides required variables and initialization. So you will need just an execution method.

And I would rename it to ClyBrowseCritiqueRuleCommand

PR #4893, opened 2019-10-11
4843-Semantic-Analysis-add-a-temp-variable-for-the-Temp-Vector-Copied-Var

### LINKED ISSUE #4843: Semantic Analysis: add a temp variable for the Temp Vector Copied Var

MarcusDenker:
this change makes sure that the copied vars of the temp vector are standard copied vars. This allows us to easily find their definiton scope (this will be used in a second step) and in addition simplify the code a bit

fixes #4843

### DIFF
--- a/src/OpalCompiler-Core/OCASTClosureAnalyzer.class.st
+++ b/src/OpalCompiler-Core/OCASTClosureAnalyzer.class.st
@@ -68,6 +68,8 @@ OCASTClosureAnalyzer >> visitVariableNode: aVariableNode [
 	aVariableNode adaptToSemanticNode.
 	aVariableNode isTemp ifFalse: [^self].
 	var := self lookupAndFixBinding: aVariableNode.
-	var isTempVectorTemp ifTrue: [scope addCopyingTempToAllScopesUpToDefVectorNamed: var vectorName].
+	var isTempVectorTemp ifTrue: [ | vectorVar |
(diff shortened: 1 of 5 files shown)


MarcusDenker:
Bootrap failure:

[32] *** Warning: Deprecation: The method Object>>#name called from PBClassLoader>>#createBehaviorFromDefinition: has been deprecated.
[32] Implement your own domain representation of an object, or use #asString or #printString instead.
[32] 'Errors in script loaded from /builds/workspace/uest_and_branch_Pipeline_PR-4893/32/bootstrap/scripts/bootstrap.st'
[32] MessageNotUnderstood: receiver of "definition" is nil
[32] UndefinedObject(Object)>>doesNotUnderstand: #definition
[32] PBSpurClassLoader(PBClassLoader)>>executeDefinitionFor:
[32] PBSpurClassLoader(PBClassLoader)>>createBehaviorFromDefinition:


No idea if this PR is the reason...

PR #4901, opened 2019-10-14
4817-Epicea-Monticello-save-event-only-adds-noise-to-log

### LINKED ISSUE #4817: Epicea: "Monticello save" event only adds noise to log
At least when Metacello loads a baseline, for example see "MC load"s here:

Screen Shot 2019-10-03 at 19 43 08

tinchodias:
Fix #4817. Do not log MC save anymore.


tinchodias:
Post-load, monitor restart is required:
```
EpMonitor restart.
```

MarcusDenker (approved):

PR #4908, opened 2019-10-14
4875-Recover-lost-changes-does-not-apply-changes-in-right-order

### LINKED ISSUE #4875: Recover lost changes does not apply changes in right order
After an image crash, if you **click yes on "recover lost changes" dialog**, then you select changes and click accept => changes are not applied in the right order.
Epicea tries to declare methods before their class.

Simple case to reproduce:
Create a class Foo, then create #a, #b, #c, #d, #e methods. Make the image crash (on OS X full window -> sleep -> wake up).
=> errors saying that class Foo is not found while trying to declare methods.

What is strange is that if you say no to the recover lost changes window and you use 'code changes' tool from the menu, changes are applied in the right order from the tool.

@tinchodias an idea of what happens?

tinchodias:
Fixes #4875. Backporting PR #3097 to P7.0.

Fixed "lost changes detector" and added a test. Also refactored a bit the already present tests.

### DIFF
--- a/src/EpiceaBrowsers-Tests/EpLostChangesDetectorTest.class.st
+++ b/src/EpiceaBrowsers-Tests/EpLostChangesDetectorTest.class.st
@@ -12,25 +12,40 @@ Class {
 
 { #category : #tests }
 EpLostChangesDetectorTest >> testDetectInEmptyLog [
-
 	detector := EpLostChangesDetector newWithLog: monitor log.
 	self deny: detector hasLostChanges.
 	self assert: detector lostChanges isEmpty.
 
-	self assert: monitor log entries isEmpty. "Just to be sure of the assumed precondition"
-
+	self assert: monitor log entries isEmpty
 ]
 
 { #category : #tests }
 EpLostChangesDetectorTest >> testDetectNoChange [
-
 	classFactory newClass.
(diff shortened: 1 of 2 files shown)


tinchodias:
Steps to review it in the image:

1. In a fresh Pharo 7, evaluate in a workspace:
```
class := Object subclass: #WillBeSaved.
1 to: 9 do: [:index |
	class compile: 'm ', index asString ].
```
2. Save image.
3. Evaluate in another workspace:
```
class := Object subclass: #WillBeLost.
1 to: 9 do: [:index |
	class compile: 'm ', index asString ].
```
4. Quit WITHOUT save.
5. Open the image and click on "Recover lost changes"
6. You will see the list of changes in the wrong order: the oldest first... 
7. Select all changes and click on apply changes, Epicea will tell you to recover the wrong version of method `m`: "m 1", while it should propose "m 9".

Now, install this PR and repeat the steps... you should get the right behavior.

tinchodias:
Additionally, there are the `EpLostChangesDetectorTest` tests to run.

tinchodias:
The CI error seems unrelated: " MessageNotUnderstood: receiver of "width" is nil"

jecisc:
I restarted

PR #4915, opened 2019-10-15
Add new #printOneLineString and related methods

tesonep:
It's nice to have a #printOneLineString that is just like #printString
but avoids printing CRs and LFs. This patch implements such idea adding
a method named #printOnOneLine:, working on a Stream, and a
FilteringStream.

The FilteringStream is used to replace CRs and LFs with a space, but can
be used to filter elements going into a Stream via a generic block.

Original PR #4854 
Issue: #4849

### DIFF
--- a/src/Collections-Streams/FilteringStream.class.st
+++ b/src/Collections-Streams/FilteringStream.class.st
@@ -1,0 +1,44 @@
+"
+I am a basic filtering stream that will wrap every object put in the stream in a block, before sending it to my decoree.
+
+"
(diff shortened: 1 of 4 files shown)


estebanlm:
ci is failing on image build.

Ducasse:
Yes there is a problem in the bootstrap but I could not understand how to fix it.

Ducasse:
We should probably package that in a separate package.
Left to do to everyone with energy.

PR #4916, opened 2019-10-15
CompletionSmallCleanUp

MarcusDenker:
Small cleanup
- remove unsent messages
- RBReturnNode>>#completionToken seems wrong. Remove it to return empty string instead
- categorize CompletionModel


Ducasse:
Cache font tests as usual

PR #4918, opened 2019-10-16
AbstractTool DeadCode remove

MarcusDenker:
renove dead code (unsent method) in AbstractTool


Ducasse:
Yes more cleaning.
I will continue my cleaning phase too.

PR #4921, opened 2019-10-16
CleanNECContextTest

MarcusDenker:
Another cleanup of the tests: remove tests that test the old system, add explanation to the new test class


src/NECompletion-Tests/CompletionContextTest.class.st
```diff
@@ -1,26 +1,18 @@
...
+{ #category : #testing }
+CompletionContextTest class >> shouldInheritSelectors [
+	^ true
 ]
 
 { #category : #private }
 CompletionContextTest >> createContextFor: aString at: anInteger [ 
-	^ NECContext
-		controller: NECController new
+	^ CompletionContext
```
myroslavarm, line 14: if you are running these tests for the new CompletionContext then SubContextTest class with the tests overrides should be deleted...

MarcusDenker: Yes, I renamed the old CompletionContext to NECContextTest and SubContextTest to CompletionContextTest

This way everything is named after what it tests, and SubContextTest is removed

PR #4922, opened 2019-10-16
Add-argumentNamed-to-Pragma

jecisc:
Add a new way to access arguments of pragma.

The new way is to use #argumentNamed:. 

I also added some tests on pragmas.


juliendelplanque (approved):

pavel-krivanek (approved):

Ducasse:
do we get an idea if it is slower? if this is the case if we can add this to the method comment?

jecisc:
I did it

PR #4924, opened 2019-10-16
3111-Why-do-we-have-in-Object--isCodeCompletionAllowed

MarcusDenker:
remove the last #isCodeCompletionAllowed implementors


myroslavarm (approved):
you should also immediately link the original issue this references, especially if it's as old as this one and one has to go look for it

PR #4925, opened 2019-10-17
Completion Cleanup:  use all Literals

MarcusDenker:
- rename ivar clazz to class (as there is no reading accessor, no clash with the #class message of Object)
- remove #methodNames: (inline in the one sender)
- completion of Literals: do not stop at 10, we use all literals when completing  var definitions, and it works

### DIFF
--- a/src/NECompletion/CompletionProducerVisitor.class.st
+++ b/src/NECompletion/CompletionProducerVisitor.class.st
@@ -58,7 +58,7 @@ CompletionProducerVisitor >> visitLiteralNode: aRBLiteralValueNode [
 	| value |
 	
 	value := aRBLiteralValueNode value.
-	(value isKindOf: Symbol) ifFalse: [ ^#() ].
+	value isSymbol ifFalse: [ ^ #() ].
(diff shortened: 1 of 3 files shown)


src/NECompletion/CompletionProducerVisitor.class.st
```diff
@@ -61,16 +55,12 @@ CompletionProducerVisitor >> visitLiteralArrayNode: aRBLiteralArrayNode [
 
 { #category : #visiting }
 CompletionProducerVisitor >> visitLiteralNode: aRBLiteralValueNode [
-	| results |
-	(aRBLiteralValueNode value isKindOf: Symbol) ifFalse: [ ^#() ].
+	| value |
+	
+	value := aRBLiteralValueNode value.
+	(value isKindOf: Symbol) ifFalse: [ ^#() ].
```
jecisc:
```suggestion
	value isSymbol ifFalse: [ ^ #() ].
```

jecisc (commented):
I suggested a small cleaning

PR #4928, opened 2019-10-17
enhance KMShorcutPrinter

estebanlm:
add special printing for Enter and Space because they are used and they can not be printed.

### DIFF
--- a/src/Keymapping-Core/KMShortcutPrinter.class.st
+++ b/src/Keymapping-Core/KMShortcutPrinter.class.st
@@ -48,6 +48,22 @@ KMShortcutPrinter class >> toString: aShortcut [
 		visit
 ]
 
+{ #category : #visiting }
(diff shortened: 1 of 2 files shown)


src/Keymapping-Core/KMShortcutPrinter.class.st
```diff
@@ -48,6 +48,22 @@ KMShortcutPrinter class >> toString: aShortcut [
 		visit
 ]
 
+{ #category : #visiting }
+KMShortcutPrinter >> genericSpecialCharacter: char [ 
+
+	char = Character cr ifTrue: [ ^ 'Enter' ].
```
jecisc, line 54: Shouldn't we manage lf also? In case in the future the line return of Pharo change?

estebanlm: nope. 
this is not "return line" but "return the char"

PR #4935, opened 2019-10-18
CompletionNavigationKeys

MarcusDenker:
There is an option to turn on a small integrated browser in completion. "Use navigation keys for extended completion functionality".

This PR adds the needed methods to show a stand-in there. Which means that this now just needs to be extended to show useful things for all the cases.


myroslavarm (approved):
yeah this is a good step to start with this

Issue #4872, opened 2019-10-09
Browse it in code search is broken...


Ducasse:
'Browse it' should be renamed 'Browse full class'
so that we keep 'Browse it' for something similar to 'Inspect it'

myroslavarm:
should be closed too?

PR #4947, opened 2019-10-21
Rename 'Browse It' to 'Browse full class'

PlasmaIntec:
https://github.com/pharo-project/pharo/issues/4872


Ducasse:
We should probably rename the method browseIt into browseFullClass too.

PR #4954, opened 2019-10-21
fix-old-workspace

estebanlm:
remove one of the duplicated implementations of #fontToUse


jecisc (approved):
Maybe we should have a smoke test?

PR #4955, opened 2019-10-21
4953-moduleName-needs-to-be-properly-deprecated

### LINKED ISSUE #4953: moduleName needs to be properly deprecated
The FFI messages moduleName were changed to libraryName (which is good) without deprecation (which is not so good). The old message is still used (Mars-Gtk etc.).

pavel-krivanek:
add moduleName deprecation

fixes #4953

### DIFF
--- a/src/Deprecated80/FFICalloutMethodBuilder.extension.st
+++ b/src/Deprecated80/FFICalloutMethodBuilder.extension.st
@@ -1,0 +1,10 @@
+Extension { #name : #FFICalloutMethodBuilder }
+
+{ #category : #'*Deprecated80' }
+FFICalloutMethodBuilder >> moduleName [
+	self
+		deprecated: 'Please use #libraryName instead'
+		transformWith: '`@receiver moduleName' -> '`@receiver libraryName'.
+
+	^self moduleName
+]
--- a/src/Deprecated80/FFILibrary.extension.st
+++ b/src/Deprecated80/FFILibrary.extension.st
@@ -1,0 +1,10 @@
+Extension { #name : #FFILibrary }
+
+{ #category : #'*Deprecated80' }
+FFILibrary >> moduleName [
+	self
+		deprecated: 'Please use #libraryName instead'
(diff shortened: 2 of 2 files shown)


pavel-krivanek:
unrelated test failures
```
windows-32 / Tests-windows-32 / testLinkOnTempVarForObject – Windows32.Reflectivity.Tests.LinkInstallerTest
<1s
osx-32 / Tests-osx-32 / testLinkOnTempVarForObject – MacOSX32.Reflectivity.Tests.LinkInstallerTest
```

MarcusDenker (approved):

src/Deprecated80/FFICalloutMethodBuilder.extension.st
```diff
@@ -0,0 +1,10 @@
+Extension { #name : #FFICalloutMethodBuilder }
+
+{ #category : #'*Deprecated80' }
+FFICalloutMethodBuilder >> moduleName [
+	self
+		deprecated: 'Please use #libraryName instead'
+		transformWith: '`@receiver moduleName' -> '`@receiver libraryName'.
+
+	^self moduleName
```
guillep, line 9: This is an infinite recursion :/

src/Deprecated80/FFILibrary.extension.st
```diff
@@ -0,0 +1,10 @@
+Extension { #name : #FFILibrary }
+
+{ #category : #'*Deprecated80' }
+FFILibrary >> moduleName [
+	self
+		deprecated: 'Please use #libraryName instead'
+		transformWith: '`@receiver moduleName' -> '`@receiver libraryName'.
+
+	^self moduleName
```
guillep, line 9: This too...

guillep (commented):
Are the infinite recursions ok? If we disable the automatic rewritings nothing will work...

Besides, this method is supposed to be abstract and redefined in subclasses, so nobody will ever call it actually.  That's why I did not deprecate it in my changes, but I'm not against the deprecation on itself.

pavel-krivanek:
damn

PR #4963, opened 2019-10-21
move-asStringYMDHM-to-base-package

estebanlm:
moving this method from iceberg to base package. 

This is because more packages are now using it (like Dr Test)


jecisc:
I opened an issue to do the change also in Iceberg: https://github.com/pharo-vcs/iceberg/issues/1322

When this PR will be integrated I hope it will be done fast because we will have a dirty version of Pharo.

jecisc (approved):

estebanlm:
tests errors not related

PR #4968, opened 2019-10-22
when focus is lost cursor should be turned off

estebanlm:
Removing focus when window lose focus and restoring it when window regains it.
Until now, cycle continued drawing the cursor, consuming CPU and making things very confusing.

This is just for the SDL2 backend and hence it will work just for the headless vm, but since this is the future... :)


Ducasse:
Broken tests are unrelated

Issue #4950, opened 2019-10-21
Add a new test for withUnixLineEndings


Ducasse:
withUnixLineEndings deserves a nicer test and it can be inspired from #4912 PR

Tests already cover it, but this is another way to review its behavior:

original :='asa' , String cr , 'asa'.
self assert: (original at: 4) = Character cr.
newString := original withUnixLineEndings.
self assert: (original at: 4) = Character cr.

olekscode:
I will try to improve that test

olekscode:
I replaced that test with a single assertion which compares two strings and should cover all cases:
```Smalltalk
testWithUnixLineEndings
	| stringWithVariousEndings stringWithUnixEndings |
	
	stringWithVariousEndings := String streamContents: [ :stream |
		stream
			lf;
			nextPutAll: 'lorem';
			cr; cr;
			nextPutAll: 'ipsum';
			cr; lf;
			nextPutAll: 'dolor sit amet';
			crlf ].
	
	stringWithUnixEndings := String streamContents: [ :stream |
		stream
			lf;
			nextPutAll: 'lorem';
			lf; lf;
			nextPutAll: 'ipsum';
			lf;
			nextPutAll: 'dolor sit amet';
			lf ].
	
	self assert: stringWithVariousEndings withUnixLineEndings equals: stringWithUnixEndings.
```

MarcusDenker:
PR https://github.com/pharo-project/pharo/pull/5811 has been merged that fixes this

PR #4970, opened 2019-10-23
Proposed fix for #4950.

WillHensel:
Added new test for withUnixLineEndings.

I chose to modify the existing test method with the new test rather than creating a new test method. I'm happy to create a new test method if that is desired.


kasperosterbye:
Looks good to me.

jgfoster:
The only changes here are to comments and tests. Looks good to me.

PR #4971, opened 2019-10-23
fasttable-not-homogeneous-rows

estebanlm:
add not homogenous container morph and capabilities to FTTableMorph

Now, passing #beRowNotHomogeneous will make the table to ask the datasource for the row height to draw.


estebanlm:
errors not related

PR #4975, opened 2019-10-24
4974-Remove-RBShouldntRaiseErrorRule

### LINKED ISSUE #4974: Remove ReShouldntRaiseErrorRule
This rule is not good for multiple reasons:
- When doing a smoke test you *want* to give the information of what exactly you want to test
- Using an assertion instead of nothing give better description when it fails
- Failing in a #shouldnt:raise: produce an assertion failure and not an error, which are two different things

I propose to remove this rule.

jecisc:
Remove RBShouldntRaiseErrorRule

This rule is not good for multiple reasons:
- When doing a smoke test you *want* to give the information of what exactly you want to test
- Using an assertion instead of nothing give better description when it fails
- Failing in a #shouldnt:raise: produce an assertion failure and not an error, which are two different things

I propose to remove this rule.

Fixes #4974


Ducasse:
Cyril can you explain what this rule is doing?

Ducasse:
The point behind the rule is that we should not capture all the error
When we write

self shouldnt: [bbb] raise: Error / Exception

We can hide any other exception, hiding super nasty bug. 
So people should not use the exception Error or Exception as argument of raise:

The rationale is badly explained but this is an important rule.

jecisc:
It creates a critic if we use the assertion #shouldnt:raise: on Error or Exception in a test.

I get **a lot** of false positive to this rule because when you do a smoke test on examples you cannot know in advance which error you can get. But you want to ensure your example is still working a minimum.

This week I found a lot of tests without any assertion. So now I don't know what to do because I don't know if the assertions were removed because features were removed, or if it's a smoke test. If it's a smoke test I have no idea what is tested.

Also, I see how we can hide an error with a #should:raise:, but how can we hide an error with a #shouldnt:raise:? This is just an assertion ensuring that one error do not get to a certain abstraction layer.

pavel-krivanek:
I think that the rule is ok. What we may should do is to introduce a special explicit message for the smoke tests like `shouldntRaiseAnyError` or something like that.

tesonep:
We have to improve the rule or the explanation, but not removing it

tesonep:
Also, Smoke Tests are not good. Having Smoke tests does not guarantee that something is not doing what intended nor broken

jecisc:
No test is perfect, it’s not for that they are not good...
Not enough yes but not good no.

If we remove all tests that do not ensure that everything is ok, we will not have much test in Pharo in the end.

hogoww:
I agree with Cyril in the sense that sometimes i test that other things are still fine, like recompiling a method that I described in a string. Those errors are *not* handled by me, but I want to know if something changes in the compiler that makes my test fail, or if I did a breaking change that does not break at my program level, but at the compilation one. (at least that's my use case.

I also agree i dislike removing the #shouldnt:raise: message, because you don't know why a statement is left there, and this is bad for reading code later on. 
(I think that's how most people will remove the warning, rather than introduce a new Error type)

Maybe not removing it will still let it, but people who use it take their responsibility in the fact that they are using something dangerous, that may hide bugs in the very thing that is suppose to show them.

For the few tests I have using this mecanism that raises this rule, I personally let it as it is, with the critic. I feel it's an okay compromise.

estebanlm:
I cannot decide here, I need the opinion of the experts.

@Ducasse, @MarcusDenker, what do you think? Do we proceed to remove, do we keep it?

guillep:
This issue is stalled since october.
I agree with Pavel, Stef and Pablo in that the PR is wrong also.
Shouldnt raise should not be used for smoke tests because it hides exceptions.
Smoke tests can be simply written without assertions.

If you're expecting a specific exception, test for it.

I'll close it.

PR #4976, opened 2019-10-24
Remove-RBMissingTranslationsInMenusRule

jecisc:
Remove RBMissingTranslationsInMenusRule.

I propose to remove RBMissingTranslationsInMenusRule because of multiple reasons:
- This rule has far too many false positive. Each time you use an APIÂ with #title or #label you end up getting this rule
- It should be the responsibility of the menu to du the translation and not the client if the translation should be done in anycase
- The rule was removed from the Critic Browser by default...

I also clean a little the code of the critic browser removing some dead code.


Ducasse:
I agree!

Ducasse:
I wanted to integrate it but I could not understand 

{ #category : #'*Manifest-Core' }
 RBCompositeLintRule class >> allGoodRules [
 	| rules |
 	rules := self rules: (self rulesClassifiedFor: RBLintRule) name: 'All Rules'.
 	self removedRules do: [ :r | rules removeRule: r ].
 	rules removeRule: (RBCompositeLintRule new name: 'Spelling').
 	rules removeRule: RBMissingTranslationsInMenusRule new.
 	^ rules
 ]

 { #category : #'*Manifest-Core' }
 RBCompositeLintRule class >> removedRules [
 	^ {(RBCompositeLintRule new name: 'Spelling').  
 		RBMissingTranslationsInMenusRule new.
 		RBAssignmentWithoutEffectRule new.
 		RBUtilityMethodsRule new. }


 	^ self rules: (self rulesClassifiedFor: RBLintRule) name: 'All Rules'

jecisc:
The CriticBrowser was removing this rule and the “Spelling” rules from its results.

I removed that because this rule is removed and there is no “Spelling” rule anymore. 

And the removedRules is dead code.

Issue #4866, opened 2019-10-09
ClyFullBrowser class>>#worldMenuOn: checks if nautilus is in the image


juliendelplanque:
It should not anymore because Nautilus has been removed.

The code checking that can be removed.

jecisc:
@MarcusDenker Maybe we can tag it as easy for beginners.

MarcusDenker:
has been merged

PR #4980, opened 2019-10-24
4866-ClyFullBrowser-classworldMenuOn-checks-if-nautilus-is-in-the-image

myroslavarm:
#4866 removed the ifTrue check for Nautilus as it's not in the system anymore

### DIFF
--- a/src/Calypso-SystemTools-FullBrowser/ClyFullBrowser.class.st
+++ b/src/Calypso-SystemTools-FullBrowser/ClyFullBrowser.class.st
@@ -146,19 +146,7 @@ ClyFullBrowser class >> taskbarIconName [
 { #category : #'world menu' }
 ClyFullBrowser class >> worldMenuOn: aBuilder [ 
 	<worldMenu> 
-	| isDefaultBrowser target targetName |
-	(self environment hasClassNamed: #Nautilus) ifFalse: [^self].
-	
-	isDefaultBrowser := Smalltalk tools browser = self.
-	(isDefaultBrowser and: [ self environment hasClassNamed: #Nautilus ])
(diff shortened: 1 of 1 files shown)


jecisc (commented):
There is still a check for Nautilus just above. 
`self environment hasClassNamed: #Nautilus`

myroslavarm:
@jecisc removing that line makes calypso pop out of tools into the main "world" context menu which i'm guessing is far from ideal. i left the ifFalse checks because logically it makes sense that when it does not find nautilus (which it doesn't) that it does some other behaviour

dionisiydk:
@myroslavarm I also think all Nautilus part needs to be removed. If there is a side effect from it then it needs to be fixed

src/Calypso-SystemTools-FullBrowser/ClyFullBrowser.class.st
```diff
@@ -146,19 +146,7 @@ ClyFullBrowser class >> taskbarIconName [
...
-	(isDefaultBrowser and: [ self environment hasClassNamed: #Nautilus ])
-		ifFalse: [ 
-			target := self.
-			targetName := 'Calypso' ].
-	(aBuilder item: (targetName, ' browser') asSymbol)	
-		order: 0.5;
-		help: targetName, ' browser';
-		icon: (self iconNamed: #nautilus);
-		action: [ target open ]
+	^ self
```
jecisc: Then now the method can be totally removed :)

jecisc (changes requested):

dionisiydk (commented):
Ah, exactly. It was just to override default browser menu

PR #4981, opened 2019-10-24
Collection comments

Kavikick:
Added executable comments to a few methods of the Collections Class


jgfoster:
Changes are limited to (executable) comments and I don't understand the built failure. Looks good to me.

Issue #4949, opened 2019-10-21
browseClassFromIt is a bad method name => browseFullClass


Ducasse:
We should 
   - rename RubSmalltalkEditor>>browseClassFromIt
   - rename browseIt (all the implementations) into browseFullClass
   It means that  RubSmalltalkEditor>>browseIt can be removed. 

Should update the senders. 

RubEditingMode class >> extendedSearchMenuOn: aBuilder [
 		label: 'Code search...' translated;
 		withSeparatorAfter;
 		with: [ 
 					(aBuilder item: #'Browse it' translated)
 					(aBuilder item: #'Browse full class' translated)
 						keyText: 'b';
 						selector: #browseIt.
 					(aBuilder item: #'Senders of it' translated)

Ducasse:
This PR broke Pharo. Since it is integrated I closed it and we will fix this in #5001

PR #4982, opened 2019-10-24
Proposed fix for #4949. 

riggja:
Rename browseClassFromIt to browseFullClass.


jgfoster:
I believe that this addresses the request in the associated issue. Looks good to me.

PR #4983, opened 2019-10-25
executable comments

JP-WWU:
Added a few executable comments to String and Boolean methods.


jgfoster:
The changes are exclusively the addition of (executable) comments and I don't understand the build/test failure. Looks good to me.

jecisc:
Test failure is unrelated. 

It's a know random failure.

Ducasse (commented):
Pay attention. 
The best is to have on executable comments wrapped in a single comment.

So can you convert

"Answer some indication of how alike the receiver is to the argument,  0 is no match, twice aString size is best score (but see example with 7).  Case is ignored. This method is used to help find mistyped variable names in methods.
 	('abc' alike: 'abc') >>> 7.
 	('action' alike: 'actions') >>> 7.
 	('action' alike: 'caption') >>> 5.
 	('action' alike: 'name') >>> 0.
 	"
 		
into 

"Answer some indication of how alike the receiver is to the argument,  0 is no match, twice aString size is best score (but see example with 7).  Case is ignored. This method is used to help find mistyped variable names in methods."
 	"('abc' alike: 'abc') >>> 7"
 	"('action' alike: 'actions') >>> 7"
 	"('action' alike: 'caption') >>> 5"
 	"('action' alike: 'name') >>> 0"

jgfoster:
@ducasse, Thank you for the style guidance. We actually debated this internally during the development exercise. The reason I came to prefer the executable line without quotes is because the cursor can be placed anywhere and a 'print-it' command will select the _entire_ line and print it. With each line in its own comment, you have to manually select the inside of the line. We certainly want to follow the Pharo convention, but I'd like to learn the benefit of individual comments. Thanks!

Ducasse:
This is easier for us to identify the executable comments when we turned them into tests.

PR #4987, opened 2019-10-25
4986-Improve-messageBrowser-title-customization

### LINKED ISSUE #4986: Improve messageBrowser title customization
RecentMessageList does 

open
	"opens the system window for recent messages"

	^ MessageBrowser new
		topologicSort: false;
		sortingBlock: [ :a :b | a timeStamp > b timeStamp ];
		messages: methodReferenceList;
		title: 'Recently accepted methods' translated;
		open
but the messagebrowser title cannot be customized

title
	^ self class title , ' [' , messageList numberOfElements printString , ']'

Ducasse:
Fixes: #4986 now message list title can be changed by users.

### DIFF
--- a/src/Spec2-Tools/MessageBrowser.class.st
+++ b/src/Spec2-Tools/MessageBrowser.class.st
@@ -550,7 +550,7 @@ MessageBrowser >> textModel: aModel [
 
 { #category : #private }
 MessageBrowser >> title [
-	^ title ifNil: [ title := self defaultTitle ]
+	^ title ifNil: [ self defaultTitle ]
 
 ]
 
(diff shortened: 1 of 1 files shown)


src/Spec2-Tools/MessageBrowser.class.st
```diff
@@ -544,7 +550,14 @@ MessageBrowser >> textModel: aModel [
 
 { #category : #private }
 MessageBrowser >> title [
-	^ self class title , ' [' , messageList numberOfElements printString , ']'
+	^ title ifNil: [ title := self defaultTitle ]
```
jecisc:
```suggestion
	^ title ifNil: [ self defaultTitle ]
```

jecisc: We should not store the default title since it's used for upadate

jecisc (changes requested):

PR #4989, opened 2019-10-25
4967-Base-image-contains-invalid-source-pointers

tesonep:
Add a clean up of the pragma collectors before of the condensing of the sources and the generation of the image.


MarcusDenker:
I personally would like to explore to simplify the whole thing and get rid of the Plugin: why not just do the thing it does (adding a method property for ffi methods) just in the compiler? All the machinery that is now there so that things are automatically recompiled if new ffi selectors are added... is this needed?

MarcusDenker:
I close this as the real solution is to remove the Pragma magic from the plugin

PR #4991, opened 2019-10-25
4990-less-squeak

### LINKED ISSUE #4990: less squeak

Ducasse:
fixes: #4990 removing hardcoded squeak references.

### DIFF
--- a/src/Collections-Strings/String.class.st
+++ b/src/Collections-Strings/String.class.st
@@ -2676,6 +2676,13 @@ String >> withSeparatorsCompacted [
 	"'test  test		 ' withSeparatorsCompacted"
 ]
 
+{ #category : #formatting }
+String >> withSqueakLineEndings [
+	"Answer a new instance where all occurrences of CRLF and LF are substituted with CR."
+	
+	^ self withInternalLineEndings
+]
+
 { #category : #'platform conventions' }
 String >> withUnixLineEndings [
 	"Answer a new instance where all occurrences of CRLF and LF are substituted with LF."
--- a/src/Collections-Strings/String.class.st
+++ b/src/Collections-Strings/String.class.st
@@ -2581,6 +2581,13 @@ String >> withCRs [
 ]
 
 { #category : #'platform conventions' }
+String >> withInternalLineEndings [
+	"Answer a new instance where all occurrences of CRLF and LF are substituted with CR. Pharo internally uses CR for carriage return."
+
+	^ self withLineEndings: String cr
+]
+
+{ #category : #'platform conventions' }
 String >> withInternetLineEndings [
 	"change line endings from CR's and LF's to CRLF's.  This is probably in prepration for sending a string over the Internet"
 	
@@ -2667,13 +2674,6 @@ String >> withSeparatorsCompacted [
 	"' test ' withSeparatorsCompacted"
 	"' test test' withSeparatorsCompacted"
 	"'test  test		 ' withSeparatorsCompacted"
-]
-
-{ #category : #'platform conventions' }
-String >> withSqueakLineEndings [
-	"Answer a new instance where all occurrences of CRLF and LF are substituted with CR."
-
-	^ self withLineEndings: String cr
 ]
 
 { #category : #'platform conventions' }
--- a/src/Compression-Tests/ZipArchiveTest.class.st
+++ b/src/Compression-Tests/ZipArchiveTest.class.st
@@ -251,13 +251,13 @@ ZipArchiveTest >> testFilePermissions [
 ]
 
 { #category : #tests }
-ZipArchiveTest >> testParseSqueakZipTimestamp [
+ZipArchiveTest >> testParsePharoZipTimestamp [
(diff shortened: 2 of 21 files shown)


svenvc:
yessss !

src/Deprecated80/DateAndTime.extension.st
```diff
@@ -0,0 +1,10 @@
+Extension { #name : #DateAndTime }
+
+{ #category : #'*Deprecated80' }
+DateAndTime class >> fromSqueakTime: anInteger [
+	"Answer an instance of the receiver with the time represented by anInteger.
+	Squeak time is in the local timezone with an epoch of 1 Jan. 1901"
+	
+	self deprecated: 'Use #fromInternalTime: instead.' transformWith: '`@receiver fromSqueakTime: `@arg' -> '`@receiver fromInternalTime: `@arg'.
+	^ self basicPharoToIso
```
gcotelli:
```suggestion
	^ self fromInternalTime: anInteger
```

Ducasse:
Thanks Gabriel. An unfortunate copy and paste.

gcotelli (changes requested):
I've suggested a code change in one of the deprecated methods

gcotelli (approved):

src/Kernel/Character.class.st
```diff
@@ -425,7 +425,7 @@ Character >> basicIdentityHash [
 ]
 
 { #category : #converting }
-Character >> basicSqueakToIso [
+Character >> basicPharoToIso [
```
nicolas-cellier-aka-nice, line 428: What is the use case for such a method?
In Squeak, there is no `basicSqueakToIso`, just this (stamped 2005):

    squeakToIso
        ^self "no longer needed"

It's been kept around just for backward compatibility with oooold packages (probably from the macRoman encoding era).
Renaming cruft Pharo instead of Squeak does not make a better world, it's still cruft ;)
On the contrary, such stamp might give the feeling that it's a validated Pharo method, and delay more in-depth examination!

If very old Squeak packages were really to be loaded into Pharo (which was never a goal, was it?),
then there should better be a Squeak-compatibility package or something like that.

I didn't verify usage in Pharo, but it's a good candidate for paste-bin IMO, please review!

src/Collections-Strings/String.class.st
```diff
@@ -2580,6 +2580,13 @@ String >> withCRs [
 	^ self collect: [ :c | c = $\ ifTrue: [ Character cr ] ifFalse: [ c ]].
 ]
 
+{ #category : #'platform conventions' }
+String >> withInternalLineEndings [
+	"Answer a new instance where all occurrences of CRLF and LF are substituted with CR. Pharo internally uses CR for carriage return."
+
+	^ self withLineEndings: String cr
+]
+
```
nicolas-cellier-aka-nice, line 2589: BTW, wasn't there a project to bring Character lf as the internal lineEnding like was done in Cuis?

src/Deprecated80/Character.extension.st
```diff
@@ -0,0 +1,8 @@
+Extension { #name : #Character }
+
+{ #category : #'*Deprecated80' }
+Character >> basicSqueakToIso [
+	
+	self deprecated: 'Use #basicPharoToIso instead.' transformWith: '`@receiver basicSqueakToIso' -> '`@receiver basicPharoToIso'.
+	^ self basicPharoToIso
+]
```
nicolas-cellier-aka-nice, line 8: Like said above, IMO it's more than deprecated, it's putrefacted!

nicolas-cellier-aka-nice: ~~above~~ below

src/Compression-Tests/ZipArchiveTest.class.st
```diff
@@ -251,13 +251,13 @@ ZipArchiveTest >> testFilePermissions [
 ]
 
 { #category : #tests }
-ZipArchiveTest >> testParseSqueakZipTimestamp [
+ZipArchiveTest >> testParsePharoZipTimestamp [
 	"Issue 6054. We're testing a private method. Eventually, the zip timestamp handling should be refactored; maybe a ZipTimestamp subclass of DosTimestamp"
 
-	| squeakDosTimestamp dateAndTime |
+	| dosTimestamp dateAndTime |
 	"This was the error-causing timestamp from http://code.google.com/p/pharo/issues/detail?id=6054#c1 . It was the 'package' file inside Balloon-CamilloBruni.85.mcz"
```
nicolas-cellier-aka-nice, line 258: While at it, this link is dead. It would be good to update it, or explain the pitfalls.

src/Kernel/Character.class.st
jecisc: I checked the usage and there is 0 sender :)
So we can just deprecate it without non-deprecated replacement!

src/Collections-Strings/String.class.st
jecisc: It would be awesome indeed :)

src/Kernel/Character.class.st
jecisc: Done in: https://github.com/pharo-project/pharo/pull/4997

PR #4993, opened 2019-10-25
make block arg and temp def use visitNode: to style comment in shout

### LINKED ISSUE #4884: Comments in temporaries declaration are not the right color
If you add comment into the temporaries declaration it does not appear the right color. 

For example, you can check the method `AdditionalMethodState>>#pragmas`

maenu:
fixes #4884


MarcusDenker (approved):
looks good. tested the image with the change it it works correctly

PR #4994, opened 2019-10-25
4458 rotation halo is broken

### LINKED ISSUE #4458: Rotation Halo is broken
When I pass the mouse on the rotation halo I get degrees sent to nil. 
Because angleOffset is nil

dupriezt:
Fixes #4458 
And adds a test for it.

This is the quick way of fixing the bug: removing the fact that a Morph receives MouseMove events if he explicitly registered for them, regardless of whether the requirements are met (see method Morph>>#handlesMouseMove:).

This bypass feature has been added by #3381 , but it doesn't give a concrete case of why it was added.

If this feature must stay after all, fixing this bug could be done by refactoring the HaloMorph class, to create dedicated subclasses for each type of morph handle and implement their methods #mouseMove: and #mouseDown: in a way that removes the need for the explicit event registration happening in the methods creating the handles (for example HaloMorph>>#addRotateHandle:

### DIFF
--- a/src/Morphic-Core/Morph.class.st
+++ b/src/Morphic-Core/Morph.class.st
@@ -3167,9 +3167,9 @@ Morph >> handlesMouseMove: anEvent [
 			+ and some button is down,
 			+ and the receiver is the current mouse focus."
 			
-	self eventHandler ifNotNil: [ :handler | 
+	"self eventHandler ifNotNil: [ :handler | 
 		(handler handlesMouseMove: anEvent) ifTrue: [ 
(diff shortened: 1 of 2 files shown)


estebanlm (changes requested):
Test is cool, but solution is bad. 
Removing event handler will break a lot of other things, it is there for something :)

estebanlm:
please, submit again with new fix :)

Ducasse:
Esteban we will have to do something about it because it was super difficult to find it.

PR #4997, opened 2019-10-25
Remove-basicSqueakToIso-completly

jecisc:
The method as no sender. No need to create a #basicPharoToIso.


Ducasse:
Cyril going from our internal representation to ISO makes sense to me this is why I did not remove it in the first place.

PR #5006, opened 2019-10-27
Remove-Deprecated70

jecisc:
Pharo 8 development started since months now. Is it time to remove Deprecated70?


Ducasse:
What would be good is to package deprecated 70 as part of migrator. 
What we should make sure it that the deprecated methods are not used internally.

jecisc:
I checked and no method is used internally. (At least the dependency analyzer did not find anything)

And tests are passing.

MarcusDenker (approved):

astares:
Cool !!!

PR #5008, opened 2019-10-27
Boolean-method

ChiExplores:
Executable comments  to Boolean  eqv: method

### DIFF
--- a/src/Kernel/Boolean.class.st
+++ b/src/Kernel/Boolean.class.st
@@ -82,15 +82,14 @@ Boolean >> deepCopy [
 ]
 
 { #category : #'logical operations' }
-Boolean >> eqv: aBoolean [ 
+Boolean >> eqv: aBoolean [
 	"Answer true if the receiver is equivalent to aBoolean."
 
-	^self == aBoolean
-	
 	"false eqv: true >>> false"
 	"(1 == 1) eqv: true >>> true"
 	"( 1 > 2)  eqv: false >>> true"
 
+	^ self == aBoolean
 ]
 
 { #category : #controlling }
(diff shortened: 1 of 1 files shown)


jecisc:
Hi! 
Thank you. 

Can the executable comment be betweent the comment and the code please?

MarcusDenker (approved):

src/Kernel/Boolean.class.st
```diff
@@ -82,10 +82,14 @@ Boolean >> deepCopy [
 ]
 
 { #category : #'logical operations' }
-Boolean >> eqv: aBoolean [ 
+Boolean >> eqv: aBoolean [
 	"Answer true if the receiver is equivalent to aBoolean."
 
-	^self == aBoolean
+	"false eqv: true >>> false"
+	"(1 == 1) eqv: true >>> true"
```
nicolas-cellier-aka-nice, line 89: Though, a better example would be with an equality test `(1 = 1)` rather than an identity test `(1 == 1)`.
The fact that equal SmallInteger share same identity is an implementation detail.

ChiExplores: ahh I see, I should have used Array new or something? I guess I wanted to make sure that whatever the user decides to compare, that it is a deep copy or referencing the same memory space and using == is a safe bet even with SmallIntegers.

nicolas-cellier-aka-nice: No, no need for Array new, just use `(1=1)` and it will be perfect.

It happens that `(1 == 1)` is currently true with most VM, but that's an implementation detail, there's been some experimental VM (written in js I think) where this was false. In case of 1, it's not even pointing to same memory, but rather the value is encoded directly into the Object Oriented Pointer (oop), this is known as an _immediate value_ in VM parlance.

PR #5009, opened 2019-10-27
Do-a-pass-on-some-flags

jecisc:
Do a pass on flags

- Remove some useless flags
- Update some flags to #pharoTodo and #pharoFixMe in prevision of the FlagBrowser integration
- Fix some flags. For example fix multiple flags in RPackage.


Ducasse:
There gofer tests broken but I do not know if there are related to your changes.

jecisc:
It is related, I see why. I’ll need to fix this.

Ducasse:
Cool thanks!

PR #5019, opened 2019-10-27
5018-RPackageSetcollectFromAllPackages-and-collectFromAllPackageswith-can-be-replaced-by-flatCollect

### LINKED ISSUE #5018: RPackageSet>>collectFromAllPackages: and collectFromAllPackages:with: can be replaced by #flatCollect:
RPackageSet>>collectFromAllPackages: and RPackageSet>>collectFromAllPackages:with: seems to be replaceable by #flatCollect: and I think it would be more efficient seen the implementations.

jecisc:
Use #flatCollect: instead of #collectFromAllPackages: since it is more optimized.

Fixes #5018

### DIFF
--- a/src/Deprecated80/RPackageSet.extension.st
+++ b/src/Deprecated80/RPackageSet.extension.st
@@ -1,6 +1,20 @@
 Extension { #name : #RPackageSet }
 
 { #category : #'*Deprecated80' }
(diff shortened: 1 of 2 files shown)


MarcusDenker (approved):
ok

PR #5020, opened 2019-10-27
Deprecated-TextMorphForEditView

jecisc:
Since PluggableTextMorph is deprecated, TextMorphForEditView should also be deprecated.

Deprecated this class, its subclass and the features of the image only used by this class.


jecisc:
The failure seems related.

Ducasse:
YES!!!

jecisc:
I fixed the problem raised by the CI. Let's see if it finds a new problem.

Ducasse:
Let us check again

MarcusDenker:
tests look ok

estebanlm:
recheck

MarcusDenker:
I changed the PR to be for pharo9

guillep:
Should this go now to `Deprecated90` instead of `Deprecated80`?

Otherwise I think the change is good.

guillep:

Ducasse:
One of the problem is that we have still some users of PluggableListMorph. 
Developing Pharo is a guerre eternelle.

guillep:
Ok, I've talked to @jecisc , he'll take care of it later.
I'll close the PR, that does not kill the branch, so a future PR could be done with the same branch.

PR #5025, opened 2019-10-28
RBExplicitRequirementMethodsRule-shouldnt-fail-on-abstract-class

jecisc:
By definition an abstract class will not implement all the logic to work in standalone. Thus, RBExplicitRequirementMethodsRule should not fail on abstract classes.


jecisc:
I also removed some dead code + categorized a method.

MarcusDenker (approved):

PR #5027, opened 2019-10-29
Dictionary #value  should use #streamContents:

MarcusDenker:
Dictionary and SmallDictionary #values can use new:streamContents: (#keys is doing that already).

Speed is not impacted as the implemention does exactly the same,


svenvc:
Yes, better.

But I still need to supply a patch to restore the old behaviour in #new:streamContents: where if the requested size and actual size at the end are the same, you get the internal buffer back and not a copy (and hence not waste memory).

PR #5028, opened 2019-10-29
#hasAbstractMethods: use #anySatisfy:

MarcusDenker:
#hasAbstractMethods can use #anySatisfy:


dionisiydk:
Previously this method was heavily used in the browser to mark abstract classes.
So it was called for every class in the class list and #methodsDo: was an optimization to avoid copy of methodDict values. Maybe it still makes sense to keep it.

(Now days the logic to mark abstract classes is moved to one of Calypso plugins and it is not only about abstract methods)

MarcusDenker:
ahh... true... I really wish anySatisfy: would be clever enough to optimise it's senders automatically...

What I will do: close this and later do a new version where I add a comment that explains the optimisation.

PR #5035, opened 2019-10-29
Executable Comments in Number

3dCameron:
Added executable comments to various methods in Number. In the case of Modulo, I removed examples from the method description and replaced them with their equivalent executable comments.

### DIFF
--- a/src/Kernel/Number.class.st
+++ b/src/Kernel/Number.class.st
@@ -196,7 +196,11 @@ Number >> @ y [
 { #category : #arithmetic }
 Number >> \\ aNumber [ 
 	"modulo. Remainder defined in terms of //. Answer a Number with the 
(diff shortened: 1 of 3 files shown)


jecisc (approved):
Thanks!

It seems good to me.

MarcusDenker (approved):

src/Kernel/Number.class.st
```diff
@@ -707,19 +738,30 @@ Number >> roundUpTo: aNumber [
 { #category : #'truncation and round off' }
 Number >> rounded [
 	"Answer the integer nearest the receiver."
+	
+	"1.4 rounded >>> 1"
+	"1.5 rounded >>> 2"
+	"2 rounded >>> 2"
+	"-1.5 rounded >>> -2"
```
nicolas-cellier-aka-nice, line 745: Just thoughts for a future PR:
the comment should tell that this method always round AWAY from zero in case of tie

Then provide an example of tie with even integer part `2.5 rounded >>> 3`
That is because some language might use other convention (round tie to nearest even).

PR #5036, opened 2019-10-30
3113-we-should-rename-NEC-and-NOC-Controller-into-CompletionEngine

### LINKED ISSUE #3113: we should rename NEC and NOC Controller into CompletionEngine
Name name names....

MarcusDenker:
- rename CompletionController to CompletionEngine
- remove NECController

fixes #3113


myroslavarm (commented):
great!

Issue #5031, opened 2019-10-29
Optimize #new:streamContents:


svenvc:
We lost the optimisation where the following would only allocate the properly sized collection once:

````
Array new: 10 streamContents: [ :out |
	10 timesRepeat: [ out nextPut: 99 atRandom ] ]. 
````

Basically, at the end of the block, if the specified size equals the actual size, #originalContents can be returned, skipping the need for a copy (since the buffer was internal, is copied fully and would become garbage anyway).

svenvc:
https://github.com/pharo-project/pharo/pull/5037

PR #5037, opened 2019-10-30
5031-Optimize-newstreamContents

svenvc:
https://github.com/pharo-project/pharo/issues/5031

Restore the optimization to SequenceableCollection>>#new:streamContents: where, if the size written equals the internal buffer collection, possibly the predicted one, no copy is made.

Add method comments and executable examples to all methods in the class side protocol streaming of SequenceableCollection.

### DIFF
--- a/src/Collections-Abstract/SequenceableCollection.class.st
+++ b/src/Collections-Abstract/SequenceableCollection.class.st
@@ -9,6 +9,10 @@ Class {
 
 { #category : #'stream creation' }
 SequenceableCollection class >> << blockWithArg [
+	"An alias to #streamContents:"
+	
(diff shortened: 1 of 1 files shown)


Ducasse:
I should say that I ***LOVE*** the comments. 
I will integrate it. Just a question do we have tests for this new behavior?

svenvc:
Thanks.

It is a bit hard/ugly to test, here is one test of the 'new' (optimized) behavior:
````
| array result |
array := nil.
result := Array new: 10 streamContents: [ :out | 
	array := out originalContents.
	10 timesRepeat: [ out nextPut: 99 atRandom ] ].
self assert: result == array
````
But I could not find a test class to add this to.

PR #5038, opened 2019-10-30
Code completion: Remove Producer

MarcusDenker:
- Instead of CompletionProducerVisitor, add methods #completionEntries to the AST as extension methods. Much simpler to understand
- rename setLabel... to not have "set"
- remove #workspace from ComplionEngine (dead code)


Ducasse:
just a remark from a design perspective. 
You can have completionEntries as an extension method on nodes and it invokes a visitor.

PR #5041, opened 2019-11-01
FileSystem>>copyFileAndDelete:to: don't raise a FileDoesNotExistException if the destination doesn't exist

akgrant43:
When renaming files Pharo falls back to copy-and-delete if the rename primitive fails (e.g. because the files are on different OS file systems).

If the copy fails the destination is (appropriately) deleted:

	[self copy: sourcePath toReference: destination] 
		on: Error 
		do: [ :error |
			destination delete.
			error signal].

However with the current implementation, if the destination doesn't exist, e.g. because the destination's directory doesn't exist, or doesn't have write access, the delete operation will raise a FileDoesNotExistException and mask the underlying cause of the copy failure.

Modify FileSystem>>copyFileAndDelete:to: so that it doesn't raise an exception if the destination doesn't exist.

### DIFF
--- a/src/FileSystem-Core/FileSystem.class.st
+++ b/src/FileSystem-Core/FileSystem.class.st
@@ -252,7 +252,7 @@ FileSystem >> copyFileAndDelete: sourcePath to: destination [
 	[self copy: sourcePath toReference: destination] 
 		on: Error 
 		do: [ :error |
-			destination delete.
+			destination ensureDelete.
 			error signal].
 	self delete: sourcePath.
 	^destination


akgrant43:
The following code triggers the issue:

```
| from toDir |

from := (FileLocator home / '.bashrc') resolve.
self
	assert: (from exists and: [ from isReadable ])
	description: 'Choose a file that you have read access to'.

toDir := '/root' asFileReference.
self
	assert: (toDir isDirectory and: [ toDir isWritable not ])
	description: 'to should be a directory without write access'.

to := toDir / 'this.shouldnt.exist'.
self
	assert: to exists not 
	description: 'Choose a file that doesn''t exist in toDir'.

"The following should resignal the exception raised by File>>openForWrite:, not a FileDoesNotExistException when trying to delete the destination file in FileSystem>>copyFileAndDelete:to:"
from copy moveTo: to.
```

Note that creating an automated test is difficult as the underlying copy code also raises a `FileDoesNotExistException` if it can't write to the destination (instead of a more appropriate directory does not exist, or no permissions type exception).  But this should be addressed in a separate issue.

Ducasse:
Thanks Alistair. It is a non working day today.

myroslavarm (approved):
cool!

bencoman (approved):
LGTM

akgrant43:
Thanks Stef, Myroslava, Ben & Marcus!

PR #5042, opened 2019-11-01
Spec based find & replace dialog

pavel-krivanek:
Replaces the current completely broken Rubric find & replace dialog with a new one that uses Spec.

It does not introduce dependency between Spec and Rubric because it is an optional package that uses (new) service registration.

### DIFF
--- a/src/BaselineOfIDE/BaselineOfIDE.class.st
+++ b/src/BaselineOfIDE/BaselineOfIDE.class.st
@@ -68,6 +68,9 @@ BaselineOfIDE >> additionalInitialization [
 	RBProgramNode formatterClass: BIConfigurableFormatter.
 
 	Color flushCache.
+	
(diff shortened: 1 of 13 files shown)


Ducasse:
{ #category : #accessing }
 SpRubFindReplaceService class >> newFor: anObject [

 	^ super new
 ]

=>

{ #category : #accessing }
 SpRubFindReplaceService class >> newFor: anObject [

 	^ self new
 ]

Idem here
DummyFindReplaceService class >> newFor: anObject [

 	^ super new
 ]

Ducasse:
I see you introduce a registration in rub and then use it!

pavel-krivanek:
Pharo_15j1cx2JhE

pavel-krivanek:
Pharo_UrncAMNGXF

PR #5051, opened 2019-11-04
4850-removeDuplicates-fails-on-Arrays

### LINKED ISSUE #4850: #removeDuplicates fails on Arrays
`#(1 2 3 4 3) removeDuplicates ` fails because array does not understand #removeAt:.

kasperosterbye:
fixes #4850

remove duplicates was badly implemented in the wrong class. Better to stick with #asSet which has a clear semantics and is efficient.

### DIFF
--- a/src/Collections-Sequenceable/OrderedCollection.class.st
+++ b/src/Collections-Sequenceable/OrderedCollection.class.st
@@ -687,20 +687,8 @@ OrderedCollection >> removeAt: index [
 
 { #category : #removing }
 OrderedCollection >> removeDuplicates [
-	| iterator |
-	"Remove the copies of elements, but keep the same order"
-	
-	self ifEmpty: [ ^ self ].
-	iterator := 1.
-	[ iterator <= self size ]
-		whileTrue: [ | each newIndex |
-			each := self at: iterator.
-			newIndex := iterator + 1.
-			[ newIndex := (self indexOf: each startingAt: newIndex).
-			newIndex > 0 ]
-				whileTrue: [ self removeAt: newIndex ].
-			iterator := iterator + 1.
-	 ]
+	self deprecated: 'use asSet instead' on: '2019-11--04' asDate in: 'Pharo 8.0'.
+	^self asSet asOrderedCollection 
 ]
 
 { #category : #removing }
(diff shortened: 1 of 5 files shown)


jecisc:
I am not fan of this change. 

Sometime I just want to *remove the duplicates* and not change the kind of collection which has a high cost for big collections. 

I would prefer to have a `self asSet as: self species` on Collection and override this one in some other collections to be more efficient.

jecisc:
Also, #removeDuplicates is much more intention revealing than #asSet.

kasperosterbye:
As I wrote in the issue #4850, there are no uses of this method outside a few I changed. Those places the result was used as merely for iteration.

The method name #asSet indicates that order is lost, while #removeDuplicates do not indicate what happens. #removeDuplicatesKeepingTheFirst, or #removeDuplicatesKeepingLast are both better names. But the semantics could be part of the comment.

The problem with implementing #removeDuplicates is exactly which of two (or more) to remove if ordering matter. By forcing `asSet asArray` you force the user to to contemplate the lack of order.

I am not against what you suggest, and a relatively efficient "keep first occurrence" algorithm exist . 

The really huge difference is that the current implementation actually **removes** duplicates from the receiver, whereas #asSet makes a copy.

The highest point in which the a true "removeDuplicates" (which do not copy) seems to be OrderedCollection as one cannot shrink all sequenceable collection (short of #becoming: them).

If we want to keep the `removeDuplicates` as a destructive method, this one does a better job:

```Smalltalk
removeDuplicates
	| elements skipped idx elem |
	"Remove all but first instance of elements, and keep the same order"
	self ifEmpty: [ ^ self ].
	elements := Set new.
	skipped := 0.
	idx := 1.
	[ idx + skipped <= self size ]
		whileTrue: [ elem := self at: idx + skipped.
			(elements includes: elem)
				ifTrue: [ skipped := skipped + 1 ]
				ifFalse: [ elements add: elem.
					self at: idx put: elem.
					idx := idx + 1 ] ].
	self removeLast: skipped
```

The same algorithm can be used for a `copyWithoutDuplicates` in SequenceableCollection.

kasperosterbye:
Actually, this one seems to work for for all collections and retaining O(N) performance:

```Smalltalk
Collection>>copyWithoutDuplicates
	| elements copy |
	"Return a copy with same iteration order, but with all but first occurance of an element removed"
	self ifEmpty: [ ^ self ].
	elements := Set new.
	copy := OrderedCollection new.
	self
		do: [ :elem | 
			(elements includes: elem)
				ifFalse: [ elements add: elem.
					copy add: elem ] ].
	^ self species withAll: copy
```

kasperosterbye:
Discussion died on this one. As it was not accepted I close it.

PR #5054, opened 2019-11-05
2283-WorldState-refers-to-missing-class-DelayScheduler

### LINKED ISSUE #2283: WorldState refers to missing class DelayScheduler
WorldState>>innerCyclePause:

MarcusDenker:
fix the reference to the non existing class. This PR does not fix anyting else, that is another thing to do. fixes #2283

### DIFF
--- a/src/Morphic-Core/WorldState.class.st
+++ b/src/Morphic-Core/WorldState.class.st
@@ -666,7 +666,7 @@ WorldState >> interCyclePause: milliSecs [
         		ifFalse: [ 0 ]
         		ifTrue: [ lastCycleTime + milliSecs - Time millisecondClockValue ].
  
-      		self flag: 'Issue 14754 - wait2>millisecs is only True for clock rollover. Remove it once DelayScheduler based on microsecondClock - Ben Coman 19.01.2015'.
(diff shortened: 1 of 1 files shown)


src/Morphic-Core/WorldState.class.st
```diff
@@ -666,7 +666,7 @@ WorldState >> interCyclePause: milliSecs [
         		ifFalse: [ 0 ]
         		ifTrue: [ lastCycleTime + milliSecs - Time millisecondClockValue ].
  
-      		self flag: 'Issue 14754 - wait2>millisecs is only True for clock rollover. Remove it once DelayScheduler based on microsecondClock - Ben Coman 19.01.2015'.
+      		self flag: 'Issue 14754 - wait2>millisecs is only True for clock rollover. Remove it once display scheduler is based on microsecondClock - Ben Coman 19.01.2015'.
```
jecisc:
```suggestion
      		self flag: 'Issue 14754 - wait2>millisecs is only True for clock rollover. Remove it once delay scheduler is based on microsecondClock - Ben Coman 19.01.2015'.
```

PR #5055, opened 2019-11-05
Fix-protocol-and-remove-useless-method

jecisc:
Fix protocol of Behavior>>#instSize and remove useless check that Object has no instance variables.


MarcusDenker (approved):
Looks good, I checked all implementors of veryDeepCopyWith: and none relies on ivar offsets

PR #5058, opened 2019-11-05
type-for-the-entries-is-not-a-good-name

myroslavarm:
changed type to kind

### DIFF
--- a/src/NECompletion/CompletionModel.class.st
+++ b/src/NECompletion/CompletionModel.class.st
@@ -146,9 +146,7 @@ CompletionModel >> sorter: aSorter [
 
 { #category : #accessing }
 CompletionModel >> title [
-	| kind |
 	[
-	kind := node receiver propertyAt: #type ifAbsent: [ ^ '' ].
-	^ kind name
+	^ node receiver propertyAt: #type ifPresent: #name ifAbsent: ''
 	] on: Error do: [ :e | ^ '' ]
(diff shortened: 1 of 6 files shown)


src/NECompletion/CompletionModel.class.st
```diff
@@ -146,9 +146,9 @@ CompletionModel >> sorter: aSorter [
 
 { #category : #accessing }
 CompletionModel >> title [
-	| type |
+	| kind |
 	[
-	type := node receiver propertyAt: #type ifAbsent: [ ^ '' ].
-	^ type name
+	kind := node receiver propertyAt: #type ifAbsent: [ ^ '' ].
```
jecisc: You could also write it

`^ node receiver propertyAt: #type ifPresent: #name ifAbsent: ''`

MarcusDenker (changes requested):
exactly here #type is a good name... the idea was that we use #type when we talk about types, but *not* for the method that returns the symbol for choosing the highlight color.

This is the method #type on the Entries hierarchy, which should be named maybe very explicitly #hightlightSymbol or something like that.

myroslavarm:
@MarcusDenker ok yeah sorry, fixed the method now - should be fine. And also i did update #title according to Cyril's suggestion, too.

MarcusDenker (approved):
Yes! Looks good.

PR #5059, opened 2019-11-06
2334-Regarding-Encoder-class

### LINKED ISSUE #2334: Regarding Encoder class
Hi!

It seems that `Encoder` only method (class side) could be pulled down to `BytecodeEncoder`. I am missing something?

MarcusDenker:
Encoder is not referenced and just has two methods on the class side
- move methods down
- remove Encoder

fixes #2334


pavel-krivanek (approved):
Fine, Deprecation seems to have no point here. And we one completely generic class name in the system is gone ;-)

PR #5061, opened 2019-11-06
remove CrLfExchangeTable

### LINKED ISSUE #4913: CrLfExchangeTable is not used anymore after #4912
BTW, now this method has no senders in the image:

String class >>
Once https://github.com/pharo-project/pharo/pull/4912 is merged

```
crLfExchangeTable

	^CrLfExchangeTable
```
which might be removed together with:

```
String class >>
newCrLfExchangeTable
	"a table for exchanging cr with lf and vica versa"
	^ Character allByteCharacters collect: [ :each |
		each = Character cr 
			ifTrue: [ Character lf ]
			ifFalse: [ 
				each = Character lf
					ifTrue: [ Character cr ]
					ifFalse:[ each ]]]
```

which is only used by #initialize and a test.

MarcusDenker:
Manually remove the var in the text file (how awkward...). fixes #4913

### DIFF
--- a/src/Collections-Strings/String.class.st
+++ b/src/Collections-Strings/String.class.st
@@ -22,7 +22,6 @@ Class {
 		'CSSeparators',
 		'CaseInsensitiveOrder',
 		'CaseSensitiveOrder',
-		'CrLfExchangeTable',
 		'LowercasingTable',
 		'Tokenish',
 		'TypeTable',


pavel-krivanek (approved):

MarcusDenker:
testInstSize is failing, this method did not exist in my image so I will do another PR after merging

PR #5064, opened 2019-11-06
5063-testInstSize-fails

### LINKED ISSUE #5063: testInstSize fails
testInstSize in Behavior is failing on the CI

-> check why and fix the test

MarcusDenker:
fixes #5063


jecisc:
This is weird because the test was green on my machine.

PR #5066, opened 2019-11-06
4232-Review-ReflectivityExamples-class-packaging

### LINKED ISSUE #4232: Review ReflectivityExamples class packaging
ReflectivityExamples class is currently packaged in "Reflectivity-Tests". 
But it is not a test - it is an example class used in tests and other places

So would ReflectivityExamples class not be better packaged into existing package "Reflectivity-Examples" ?

Comments?

MarcusDenker:
- copy the two examples to the CoverageDemo class

This way it does not need to reference any class from the tests and is self-contained

fixes #4232


pavel-krivanek (approved):
nice

PR #5067, opened 2019-11-06
5062-includesKey-does-not-work-after-garbage-collect-in-WeakValueDictionary

### LINKED ISSUE #5062: #includesKey: does not work after garbage collect in WeakValueDictionary
After the garbage collect, you obtain an empty array by asking #keys but the #includesKey: still answer True when asked a collected key. This is due to implementation inherited from Dictionary.

The following test added in WeakValueDictionaryTest show the problem:

testIncludesKey
	self
		deny:
			(self dictionaryWithGarbageCollectedValue
				includesKey: self keyWithGarbageCollectedValue)

jecisc:
WeakValueDictionary>>#includesKey: should return false if the value was GCed. 

Fix the bug + add a test (thanks Guillaume).

Fixes #5062

### DIFF
--- a/src/Collections-Weak/WeakValueDictionary.class.st
+++ b/src/Collections-Weak/WeakValueDictionary.class.st
@@ -77,7 +77,9 @@ WeakValueDictionary >> includesKey: key [
 
 	"Here we need to check that we have a key and also that the value of the key was not garbarge collected."
 
-	^ (array at: (self scanFor: key)) ifNil: [ false ] ifNotNil: [ :value | value value isNotNil ]
+	^ (array at: (self scanFor: key))
+		ifNil: [ false ]
+		ifNotNil: [ :value | value value isNotNil ]
 ]
 
 { #category : #private }
--- a/src/Collections-Weak/WeakValueDictionary.class.st
+++ b/src/Collections-Weak/WeakValueDictionary.class.st
@@ -73,9 +73,7 @@ WeakValueDictionary >> at: key put: anObject [
 
 { #category : #testing }
 WeakValueDictionary >> includesKey: key [
-	"Answer whether the receiver has a key equal to the argument, key."
-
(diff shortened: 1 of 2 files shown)


src/Collections-Weak/WeakValueDictionary.class.st
```diff
@@ -71,6 +71,15 @@ WeakValueDictionary >> at: key put: anObject [
 	^ anObject
 ]
 
+{ #category : #testing }
+WeakValueDictionary >> includesKey: key [
+	"Answer whether the receiver has a key equal to the argument and also that the value associated to this key was not garbage collected."
+
+	^ (array at: (self scanFor: key))
+		ifNil: [ false ]
+		ifNotNil: [ :value | value value isNotNil ]
```
dionisiydk, line 80: unrelated comment:
Do we really decide to move to #isNotNil methods instead of #notNil? For me the consistency is to use the only version and I doubt that it is possible to deprecate #notNil. Then we will always have both versions in the image.

dionisiydk (approved):

MarcusDenker (approved):

src/Collections-Weak/WeakValueDictionary.class.st
jecisc: Personally, I always use #isNotNil for consistency with #isNil. I prefer to have the same prefix in all cases. I do not find consistent to have `notNil` and `isNil`.

dionisiydk: I understand the original intention of #isNotNil. And I would also prefer this kind of consistency for testing methods where all of them are starting with #is prefix. 
But the reality is that we can't remove notNil and notEmpty (do you agree?). And then I would prefer couple of exceptions from the convention either than duplicated methods.

PR #5068, opened 2019-11-06
Clean Compiler external dependencies

pavel-krivanek:
Cleans external dependencies in the compiler packages. As in case of kernel packages, FileSystem-Core dependency left


pavel-krivanek:
tests are ok (one unrelated failure on mac)

Ducasse:
This is just super cool!!!

PR #5072, opened 2019-11-07
MCClassTraitDefinition-should-take-category-into-account-for-equality

jecisc:
MCClassTraitDefinition should take category into account for equality. 

I got a bug this week because it considered that an instance with a nil category was the same that an instance with a category.


jecisc:
This can cause bugs in the loading of a project and I think it should be backported to Pharo 7. I'll do a PR this afternoon.

PR #5075, opened 2019-11-07
Backport-5072-MCClassTraitDefinition-should-take-category-into-account-for-equality

jecisc:
This is fixing a bug while loading code.

In Pharo 7, the loading of code is using a cache for MCDefinitions. 
When you create a new definition, it checks if one equivalent was already created, if yes, it return this one.

The equality of MCClassTraitDefinition does not take the category into account and I got bug while loading code because the MCClassTraitDefinition was wrong because it found one with another category.


jecisc:
Backport of https://github.com/pharo-project/pharo/pull/5072

jecisc:
Any news on this? I still get the bug in Pharo 7 and it's pretty annoying :(

myroslavarm (approved):

Ducasse:
The broken tests seems unrelated.

Issue #5077, opened 2019-11-07
Add byteAt: as alias to digitAt: to Integer in Pharo 7 to support forward compatibility with Pharo 8


svenvc:
Add byteAt: as alias to digitAt: to Integer in Pharo 7 to support forward compatibility with Pharo 8

svenvc:
Correct PR is https://github.com/pharo-project/pharo/pull/5079 (against Pharo 7)

Wrong PR is https://github.com/pharo-project/pharo/pull/5078 (against Pharo 8)

MarcusDenker:
merged

PR #5078, opened 2019-11-07
5077-Add-byteAt-as-alias-to-digitAt-to-Integer-in-Pharo-7-to-support-forward-compatibility-with-Pharo-8

svenvc:
Add byteAt: as alias to digitAt: to Integer in Pharo 7 to support forward compatibility with Pharo 8

https://github.com/pharo-project/pharo/issues/5077


svenvc:
Oeps, wrong branch, this PR should be deleted

PR #5082, opened 2019-11-07
Compiler API: turn forSyntaxHighlighting into an option

MarcusDenker:
- change the #forSyntaxHighlighting state to use a compiler option instead of an ivar 
- this allows us to use the compiler facade in both syntax highlighting and code completion for parsing, makign the code much easier to understand and remove direct references to RBParser.


MarcusDenker:
Hmm.. there is a problem with class definitions. Will check

MarcusDenker:
I changed it to for now just to the option conversion. Need to chang #parse to not do doit-wrapping (if you check the users this makes a lot of sense)

PR #5092, opened 2019-11-07
Clean Display packages group dependencies

pavel-krivanek:
clean Display packages group dependencies


Ducasse:
It feels like we will finally have a minipharo and super nice packages. I feel like working on miniimage and cleaning :)

Ducasse:
The broken test is a glyt font cache test

PR #5095, opened 2019-11-07
5087-Fuel-should-be-out-of-Morphic-baseline-

### LINKED ISSUE #5087: Fuel should be out of Morphic baseline.
Fuel should probably be in an extra tools baseline but I do not get why it would make sense to have it in Morphic.

pavel-krivanek:
standalone BaselineOfFuel

fixes #5087


pavel-krivanek:
one unrelated test failure

jecisc (approved):
Fuel is still in Morphic but now in its standalone baseline. It's a really nice first step to remove it from Morphic.

Ducasse:
This is really cool to get all these improvements.

PR #5101, opened 2019-11-08
add-method-to-identify-aio-interrupt-vm

estebanlm:
Prepare the image to react to the new AIO interrupt VM. 
IdleTime now can be bigger (we are trying with 1min, but it can be even larger). 
In presence of a non-AIO Interrupt VM, we behave as usual.


Ducasse:
This changes break 99 tests so it should be something simple.

estebanlm:
mmm... I will let the automatic selection of idle time for later. 
I will make another PR to add the selectors, since they are needed to adapt Spec2+GTK

PR #5105, opened 2019-11-08
5104-The-Through-button-in-the-debugger-should-be-named-Step

### LINKED ISSUE #5104: The Through button in the debugger should be named Step
The Through button in the debugger should be named simply Step. In most cases, it is the kind of stepping the users want and this name is more friendly and understandable to newcomers.

pavel-krivanek:
rename Through button to Step in the debugger

fixes #5104

### DIFF
--- a/src/Debugger-Actions/StepIntoDebugAction.class.st
+++ b/src/Debugger-Actions/StepIntoDebugAction.class.st
@@ -23,7 +23,7 @@ StepIntoDebugAction >> defaultKeymap [
 { #category : #accessing }
 StepIntoDebugAction >> defaultLabel [
 
-	^ 'Into'
+	^ 'Step into'
 ]
 
 { #category : #accessing }
--- a/src/Debugger-Actions/StepOverDebugAction.class.st
+++ b/src/Debugger-Actions/StepOverDebugAction.class.st
@@ -23,7 +23,7 @@ StepOverDebugAction >> defaultKeymap [
 { #category : #accessing }
 StepOverDebugAction >> defaultLabel [
 
-	^ 'Over'
+	^ 'Step over'
 ]
 
 { #category : #accessing }
--- a/src/Debugger-Actions/StepThroughDebugAction.class.st
+++ b/src/Debugger-Actions/StepThroughDebugAction.class.st
@@ -23,7 +23,7 @@ StepThroughDebugAction >> defaultKeymap [
 { #category : #accessing }
 StepThroughDebugAction >> defaultLabel [
 
-	^ 'Through'
+	^ 'Step'
 ]
 
 { #category : #accessing }
--- a/src/Debugger-Actions/StepThroughDebugAction.class.st
+++ b/src/Debugger-Actions/StepThroughDebugAction.class.st
@@ -23,7 +23,7 @@ StepThroughDebugAction >> defaultKeymap [
 { #category : #accessing }
 StepThroughDebugAction >> defaultLabel [
 
-	^ 'Step'
+	^ 'Step Through'
 ]
 
 { #category : #accessing }


myroslavarm (commented):
i actually think the opposite: "step" by itself doesn't really convey any meaning. usually in other IDEs it's just "step into", "step out" and "step through" and in pharo we just got rid of the word "step" everywhere. so while the icon next to "through" helps, i think renaming it to simply "step" might actually be more confusing. especially as it kind of defeats the purpose of naming everything in the same convention, because now you'll have two prepositions and a verb

akgrant43:
> i actually think the opposite: "step" by itself doesn't really convey any meaning...

+1

Ducasse:
For me I do not understand what through is doing compare to step into or step over.
I think that this is totally confusing. I never use through because of that.
Which is super sad.

pavel-krivanek:
What about this... it is similar to VW naming
Pharo_N8d5jKYmVf

estebanlm:
I think `Step Through` would maybe fit :)
In any case, this can be decided on Pharo9 (so we can rollback if it does not fits)

guillep:
I like indeed `step through`. @pavel-krivanek if you're ok with that too, I think we have an agreement.

I'll wait for input for a couple of days :)

StevenCostiou:
I agree with Myroslava, why not removing "step" everywhere once and for all (like the current gt debugger) to have meaning?

In Pharo 9 we'll have "step to return", "step out" and probably some "step blabla" in addition.
In the end it is obvious each of those actions does step the code, and the only concern is "how".

I would go for "Into", "Over", "Through", "Out", "Return", etc.
Maybe we need a better name than "through".

What we should have is, when you go over the button, display highlight in the debugger source code pane so show where that action is supposed to take you.
This way, we'd have names with meaning and visual feedback to understand that meaning (but that is a story for the Spec debugger...).

guillep:
> In the end it is obvious each of those actions does step the code, and the only concern is "how".

Let's not project ourselves and our expertise into others ^^.
I imagine for people learning, it's not so obvious.
Yes, it's a bit more verbose for us, but for other people I think it is necessary.

Just out of curiosity I checked how the debugger in IntelliJ does it.
image
image

- They have buttons without labels
- Each button has a tooltip with full name + shortcut
- There is below in the status bar a longer explanation

Ducasse:
I'm may be not a newbie anymore but I have super difficulties learning abstract icons so this is why I hated so much when there were no text or deficient fly by help (the problem with fly by help is that I have to still go over the icons). 
Having the text and the icon is so much more powerful.
Please do not break my productivity.
S.

myroslavarm:
to clarify: i didn't mean we should get rid of "step", i just said that's what we had in pharo. however, getting rid of a clarifying preposition like "through" and just leaving "step" in its place breaks convention imo. with or without "step" (but yes, including "step" makes sense, to be closer to how it's in other IDEs), i would say into/over/through should be left alone

StevenCostiou:
So what about having short labels: "into", "over"... with icons, tooltip, explanation and code highlighting/decorations?

Ducasse:
It would work for me the text is key because I can read it. 
Which reminds me that I did not send the icons to lusy.

StevenCostiou:
> to clarify: i didn't mean we should get rid of "step", i just said that's what we had in pharo. however, getting rid of a clarifying preposition like "through" and just leaving "step" in its place breaks convention imo. with or without "step" (but yes, including "step" makes sense, to be closer to how it's in other IDEs), i would say into/over/through should be left alone

We should get rid of step. Takes useless place in the GUI and once you get accustomed you don't need it.
It do not feel, from practice, that it makes the tool easier to understand.
I never wonder what "step through" means, just "what's the difference between in and through".
Having helpers and explanation looks like a good start (and easy to implement).

myroslavarm:
> We should get rid of step. Takes useless place in the GUI and once you get accustomed you don't need it.
> It do not feel, from practice, that it makes the tool easier to understand.
> I never wonder what "step through" means, just "what's the difference between in and through".
> Having helpers and explanation looks like a good start (and easy to implement).

that's fair enough too

guillep:
> Having helpers and explanation looks like a good start (and easy to implement).

But that's not the point of this PR.
So you suggest to close this PR?

> It do not feel, from practice, that it makes the tool easier to understand.
> I never wonder what "step through" means, just "what's the difference between in and through".

I personally feel that it would have an impact with students and people learning.

> We should get rid of step. Takes useless place in the GUI and once you get accustomed you don't need it.

So I understand that what bothers you is actually a screen space problem. Am I right?
If that's the thing, I personally think that shrinking the buttons is not a solution but a workaround :).
I understand the problem, but I would need to read more books on good UI design (I've read a couple) to understand how to provide a real solution to the fact of having **tons** of actions and few space.

Anyhow. @pavel-krivanek if you follow the discussion, give us some feedback.

StevenCostiou:
> > Having helpers and explanation looks like a good start (and easy to implement).
> 
> But that's not the point of this PR.
> So you suggest to close this PR?

No, I suggest removing step and putting it in the helper bubble as a start.
This is just a suggestion. 

> > It do not feel, from practice, that it makes the tool easier to understand.
> > I never wonder what "step through" means, just "what's the difference between in and through".
> 
> I personally feel that it would have an impact with students and people learning.

That is possible, in fact I do not have data to back-up my point.

> > We should get rid of step. Takes useless place in the GUI and once you get accustomed you don't need it.
> 
> So I understand that what bothers you is actually a screen space problem. Am I right?

No, that's part of the ui problem.
The problem to me is that there is no additional value in the meaning of the button.
Of course it steps. What do standard debuggers do apart stepping?
Users have to document themselves a bit before using a tool. Putting this constraint of being "begginner-friendly" seems unrelated. For beggingers, we need something else, like tool-tips or tutorials the first time you open the debugger (for example, not suggesting).

> If that's the thing, I personally think that shrinking the buttons is not a solution but a workaround :).

It's not, but it's a side-effect of something I find not bringing additional value. 
Maybe I am wrong about the value.

> I understand the problem, but I would need to read more books on good UI design (I've read a couple) to understand how to provide a real solution to the fact of having **tons** of actions and few space.

Please share references ;)

> Anyhow. @pavel-krivanek if you follow the discussion, give us some feedback.

+1

guillep:
> No, I suggest removing step

My point, if you check the changes (I encourage you to do it :)), In the current state there is no "Step", and Pavel proposes to add "Step" ;).
So "removing step" == not integrating the PR. 

> and putting it in the helper bubble as a start.
> This is just a suggestion.

Well, my point is that the PR originally was not about adding bubbles.
So somebody should make responsible of doing that, and not necessarily Pavel, because that was not the point of the PR.

> You need to document yourself a bit before using a tool. Putting this constraint of being "begginner-friendly" seems unrelated.

I'm sorry, you're right. It's me that never reads the manuals :P.
I buy a printer and plug it xD. I buy a machine to make pasta and explore it :).
But maybe others do it.

> not bringing additional value.

**for you** :P

> Please share references ;)

I have a copy of [Dont make me think](https://www.amazon.com/Dont-Make-Think-Revisited-Usability/dp/0321965515) at home, if you want.
This one looks interesting https://www.amazon.com/dp/0134777603/ref=emc_b_5_i

<blockquote><div><strong><a href="https://www.amazon.com/dp/0134777603/ref=emc_b_5_i">Sorry! Something went wrong!</a></strong></div></blockquote>

StevenCostiou:
> > No, I suggest removing step
> 
> My point, if you check the changes (I encourage you to do it :)), In the current state there is no "Step", and Pavel proposes to add "Step" ;).
> So "removing step" == not integrating the PR.
> 
> > and putting it in the helper bubble as a start.
> > This is just a suggestion.
> 
> Well, my point is that the PR originally was not about adding bubbles.
> So somebody should make responsible of doing that, and not necessarily Pavel, because that was not the point of the PR.
> 
I see, well then I vote for integrating the PR if it matters (my vote not the PR).

> > You need to document yourself a bit before using a tool. Putting this constraint of being "begginner-friendly" seems unrelated.
> 
> I'm sorry, you're right. It's me that never reads the manuals :P.
> I buy a printer and plug it xD. I buy a machine to make pasta and explore it :).
> But maybe others do it.

I think I miswrote: 
- it was not directed to you
- it is not the right level of dealing with new users (still in my opinion) => don't put constraints on the tool but add exploratory features/tools 

> 
> > not bringing additional value.
> 
> **for you** :P
>
> > Please share references ;)
> 
> I have a copy of [Dont make me think](https://www.amazon.com/Dont-Make-Think-Revisited-Usability/dp/0321965515) at home, if you want.
> This one looks interesting https://www.amazon.com/dp/0134777603/ref=emc_b_5_i
> 
Thanks.

<blockquote><div><strong><a href="https://www.amazon.com/dp/0134777603/ref=emc_b_5_i">Sorry! Something went wrong!</a></strong></div></blockquote>

guillep:
> I think I miswrote:
> 
> * it was not directed to you

Me neither, so I did not take it personally.
I just wanted to emphasize the fact that we make assumptions based on my/our own backgrounds. We all do. And me in particular, so I was self-correcting myself in public, just that :)

guillep:
> > I think I miswrote:
> > 
> > * it was not directed to you
> 
> Me neither, so I did not take it personally.
> I just wanted to emphasize the fact that we make assumptions based on my/our own backgrounds. We all do. And me in particular, so I was self-correcting myself in public, just that :)

pavel-krivanek:
Sorry, I missed this discussion. 
So... From my perspective, the problem is that probably in every other debugger that the newcomers have seen in other environments, there is just "step into" and "step over" while the meaning of "step over" is the same as our "step through". So "step through" is for them something alien and strange and not something what they would suppose to use in the first place. So I tried to simply make from "step into" and "step over" the special cases.

Does it make sense to you?

pavel-krivanek:
We could consider to rename "step through" to "step over" and change "step over" to something else but then we would confuse the current users and it is hard to find short name with correct meaning.

guillep:
> but then we would confuse the current users and it is hard to find short name with correct meaning.

Yes, I'd avoid that :).

Yes, step through is kind of alien, but I don't think calling it just "step" will make it less alien. If something, it will just confuse more people "what's the difference between step and step over? or step into?" :)

Now, I don't think we will arrive to a 100% consensus in this. This is the kind of change where everybody has a different feeling :).

So I'll cut it with what I've heard:
 - simple to understand for newcomers (even new debugger users) => prefix with "Step"
 - use "Step through" instead of step
 - I'll be very happy if somebody makes a follow-up PR with good tooltips :)

guillep (approved):

pavel-krivanek:
Thanks Guille

guillep:
Just a little follow up, the buttons seem to have already their tooltip.

image
image
image

Maybe we need to use some other mechanism?

guillep:
Thanks to everybody for the discussion :)

PR #5106, opened 2019-11-08
4337-tests-of-RB-should-not-use-category-based-class-definition-but-package

### LINKED ISSUE #4337: tests of RB should not use category based class definition but package:
testAddClassInteractively
	| change |
	change := changeFactory addClassDefinition: 'TestCase subclass: #' , self class name , '
	instanceVariableNames: ''instVar''
	classVariableNames: ''ClassVar''
	poolDictionaries: ''PoolDict''
	category: ''' , self class category , '''' for: self.

MarcusDenker:
fixes #4337


Ducasse:
agreed These string manipulation of class definitions are annoying

PR #5107, opened 2019-11-08
5102-FastTable-Horizontal-Scrolling-not-implemented

danil-osipchuk:
A simple implementation of horizontal scrolling for FTTableMorph, to enable one has to set trialHSB:=true during initialization, otherwise current behavior is unchanged.
When enabled FTTableContainerMorph  allocates space to columns (#calculateColumnWidths) starting (#columnOrderOfWidthDistribution) from startColumnIndex ( instance var added with this commit, when set to 0 used as a flag to activate current/default behavior )  corresponding to the horizontal bar position triggered by  #adjustToHorizontalScrollBarValue: 
Only columns which manage to get space are being drawn  (#exposedColumnsRange: )

FTTableMorph newTrialHorizontalScrollBar                  <-- existing example
FTTableMorph>>#initialize                                                     <--  add trialHSB := true.  to enable it everywhere

### DIFF
--- a/src/Morphic-Widgets-FastTable/FTExamples.class.st
+++ b/src/Morphic-Widgets-FastTable/FTExamples.class.st
@@ -807,7 +807,7 @@ FTExamples class >> exampleTableHorizontalScroll [
 
 	<example>
 	| table |
-	table := FTTableMorph newTrialHorizontalScrollBar
+	table := FTTableMorph newWithHorizontalScrollBar
 		extent: 650 @ 500;
 		addColumn: ((FTColumn id: '#') width: 40);
 		addColumn: ((FTColumn id: 'Name') width: 350);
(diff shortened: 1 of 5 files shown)


estebanlm (approved):

estebanlm:
but... I would change `trialHSB` with `horizontalScrollBar` (since is not a trial anymore)

danil-osipchuk:
trialHSB was used as a flag to enable initialization of a (existing) horizontalScrollBar with ScrollBar instance. I've nuked trialHSB and am using horizontalScrollBar itself for that role to have it disabled for now by default.
If we are to decide to enable it permanently all is needed is to remove  'horizontalScrollBar ifNotNil:' conditions in FTTableMorph

estebanlm:
errors not related

PR #5108, opened 2019-11-08
bind on collections

pavel-krivanek:
introduce bind: on collections

example:
#(1 2 3) bind: [ :first :second :third | first + second + third ]


svenvc:
Cool ! Looks good.

What is the change in TraitedClass >> isLocalMethodsProtocol: ?

pavel-krivanek:
that is automatic deprecation rewrite not directly related to the subject of the PR

Ducasse:
Pavel what is happening if I have more or less items. 
May be the comment should be clearer. 
"Extract from the receiver items the arguments that should be passed to the block argument."

pavel-krivanek:
it raises a classic error about not matching arguments count (which is the simple error with text description so it cannot be properly tested...)

pavel-krivanek:
I introduced the ArgumentsCountMismatch error and improved the method comment

svenvc:
Yeah, beautiful.

pavel-krivanek:
the dependency test failure is unrelated: `KeyNotFound: key 'SmartSuggestions' not found in Dictionary`
It is a side-effect of another PR

kasperosterbye:
This one is hard to disagree upon. Very nice!

My only minor concern is the name `bind`. It is a good name, but so is `bindIn:` which ties back to `Object>>in:`. In python some call it "unpacking".

jecisc (approved):

svenvc:
@kasperosterbye I understand your point, but bind is a good name by itself

In Common Lisp this is called destructuring-bind (although it is a bit more powerful), Pavel's example becomes
````
(destructuring-bind (first second third) 
     '(1 2 3) 
  (+ first second third))
````
http://www.lispworks.com/documentation/HyperSpec/Body/m_destru.htm


<blockquote><div><strong><a href="http://www.lispworks.com/documentation/HyperSpec/Body/m_destru.htm">CLHS: Macro DESTRUCTURING-BIND</a></strong></div></blockquote>

kasperosterbye:
Cool - no matter the name it is nice to have an inverted syntax for `valueWithArguments:`.

Ducasse:
I wonder why we have this conflict? I was about to integrate it but I prefer to wait. 
I like the New exception. :)

pavel-krivanek:
The deprecation message fix (non-related to the subject of this PR) was already integrated (with different formatting). Fixed

PR #5111, opened 2019-11-08
5110-remove-smartSuggestions-classes

Ducasse:
Cleaning all references and removed all the classes


Ducasse:
I will fix the external dependencies once the packages are removed.

PR #5113, opened 2019-11-09
5112-Float-negativeZero-should-be-negative

### LINKED ISSUE #5112: Float negativeZero should be negative
Snippit from Discord/Pharo/development:

image

kasperosterbye:
Fixes #5112
Overrides the implementation of positive and negative to use the signBit for floats. This allows Float negativeZero to be negative.

### DIFF
--- a/src/Kernel-Tests/FloatTest.class.st
+++ b/src/Kernel-Tests/FloatTest.class.st
@@ -361,6 +361,16 @@ FloatTest >> testNaNisLiteral [
 	self deny: Float nan isLiteral description: 'there is no literal representation of NaN'
 ]
 
+{ #category : #test }
+FloatTest >> testNegative [
+
+	self assert: 0.0 negative not.
+	self assert: -1.0 negative.
+	self assert: Float fmax negated negative.
+	self assert: Float negativeInfinity negative.
+	
+]
+
 { #category : #'tests - zero behavior' }
 FloatTest >> testNegativeZeroAbs [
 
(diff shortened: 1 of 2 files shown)


jecisc:
Hi @nicolas-cellier-aka-nice .
I think you know more about the number implementation than most people. Maybe this can iterest you.

nicolas-cellier-aka-nice:
That's what I first thought some years ago:

    myself expect: Float negativeZero isNegative.
    myself expect: Float negativeZero isZero.

(as the name tells)
But...
Look at the meaning of negative and positive currently:

    0 positive -> true.
    0 strictlyPositive -> false.
    0 negative -> false.
    0 strictlyNegative -> MessageNotUnderstood.

You see a dyssymetry, `positive` means `positiveOrNull`, while `negative` means `strictlyNegative`.

With that in mind, we can't have `Float negativeZero striclyNegative`, because it's essentially zero.

Or you have to introduce `striclyNegative` to distinguish the case, but then "_positive_" `0.0 negative -> true` and `-0.0 positive -> true`...
You then might have to rename positive as positiveOrNull and negative as negativeOrNull in order to mitigate the surprise.

Unless you do that, this change is not a good change, whatever myself (and some others) did expect from the naming...

kasperosterbye:
I was wrong to post both the issue and the PR.

According to wikipedia (which is normally OK om math and computer science), negative is LESS than zero. Float negativeZero is not less than zero.

Also according to wikipedia, positive is LARGER than 0. However, the comment of the postitive test states the in Smalltalk zero is positive. So I close this one.

PR #5115, opened 2019-11-09
Some-cleanings-on-protocols

jecisc:
Some cleanings in protocols

- rename 'instance-creation' into 'instance creation'
- add non regression test on the previous cleaning
- remove useless prefixes on some protocols to reduce the number of protocols used in the image
- Improve ProtocolConventionsTest>>#assertProtocols:areNotUsedInsteadOf: to take traits and class sides into account
- rename 'instance initalization' into 'instance initialization'
- remove some dead code
- Unify 'filein/out' protocols to have the same form + add non regression test.
- Categorize methods in 'not categorized' protocol
- Improve classification of some methods


Ducasse:
Thanks cyril
I have a little tool to help doing this in the Automatic Categorizer.

PR #5122, opened 2019-11-12
Backport - Fix nextVersion for cases where an entry with basename only exists in…

noha:
… the directory. In that case the calculation of the offset raises a primitive failure. #nextVersion has been rewritten in pharo8 so this is only for pharo7 backport


noha:
test failure seem to be unrelated

pavel-krivanek (approved):

PR #5132, opened 2019-11-13
Rules-are-failing-on-bytearrays

jecisc:
Fix two problems with Global Variable Usage rule.

- The rule crashed on ByteArray
- The rule considered Traits as globals


MarcusDenker:
This PR breaks the Global Variable Usage rule and make *a lot* of tests fail

Normally the failing test should have been catched by the CI. 

I will revert this

MarcusDenker:
After reverting the tests are green, I merged the revert

PR #5135, opened 2019-11-13
Completion: #completionEntries creates entries

MarcusDenker:
CompletionEntry: we do the wrapping too late. when we instantiate it we just have symbols, so we do not know what they
where originally (variable, selectors...)

Thus: create entries in the #completionEntries method 

TODO after this: we need to create one kind of  entries for selectors, other subclass for selectors


myroslavarm (approved):
looks good! regarding the TODO: do you mean method/message nodes?

MarcusDenker:
I think we need one subclass for all variables, one for message sends and one for selectors

PR #5139, opened 2019-11-14
5117-Division-by-zero-for-Float-should-be-infinity

### LINKED ISSUE #5117: Division by zero for Float should be infinity
`2.0 / 0.0` should be `Float infinity` according to the IEEE754. But it raises division by zero.

Should `2 / 0` also yield `Float infinity`? (as there is no other infinity)

Same goes for `0.0 raisedTo: -1.0`, which should also be Float infinity.

On the other hand `0 log` do properly yield `Float negativeInfinity`.

kasperosterbye:
Fixes #5117. Allow division by zero for float as per IEEE754


kasperosterbye:
This idea, while correct in principle, should not be done as a PR without further discussion. I close it and leave numbers undisturbed until someone explicitly ask me to work on them.

PR #5143, opened 2019-11-15
Completion: More Cleanups

MarcusDenker:
Some more cleanups:
- more renames controller -> engine
- uniyfy editor and model ivars: we can just use the model of the editor, they will always be the same
- remove parameter in #openMenuFor:


MarcusDenker:
failing tests not related

PR #5146, opened 2019-11-15
Completion: Move #commonSuperclass:

MarcusDenker:
The extension method commonSuperclass: of the completion framwork makes a lot of sense as a general method,
-> Move to Class
-> add test

### DIFF
--- a/src/Kernel-Tests-Extended/ClassTest.class.st
+++ b/src/Kernel-Tests-Extended/ClassTest.class.st
@@ -149,9 +149,9 @@ ClassTest >> testClassSide [
 
 { #category : #tests }
 ClassTest >> testCommonSuperclass [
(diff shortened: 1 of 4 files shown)


src/Kernel-Tests-Extended/ClassTest.class.st
```diff
@@ -147,6 +147,13 @@ ClassTest >> testClassSide [
 	self assert: Point class classSide equals: Point class. 
 ]
 
+{ #category : #tests }
+ClassTest >> testCommonSuperclass [
```
jecisc:
```suggestion
ClassTest >> testCommonSuperclassWith [
```

jecisc (approved):
I suggested renaming the test so that we have the little icon next to the method in the browser. 

Except this, this PR seems good to me!

jecisc (approved):
Good!

PR #5148, opened 2019-11-15
Cleanup: #copiedMethodsFromSuperclass

MarcusDenker:
- copiedMethodsFromSuperclass and friends are not used, we now have  rule for this instead 
Two small things:
- #hasAbsractMethods: explain why not #anySatisfy
- better comment in #contains


jecisc (approved):
Good!

PR #5150, opened 2019-11-15
#parseTree needs to use compiler

MarcusDenker:
#parseTree should use the compiler. This way compiler settings of the class are taken into account


Ducasse:
Looks much nicer :)

PR #5152, opened 2019-11-16
CompilerNilCompiler

MarcusDenker:
add a #compiler method to UndefinedObject. We use nil when compiling and parsing expressions, if we have this method, we can easily use the compiler that the class defines for syntax highlight and code completion (without needing to add an if to check for nil)


Ducasse:
I did not understand the comment of the method. 
Can you extend it?
Because I see Smalltalk compiler and I was wondering how can it be nil then.
There is also a typo in the example. 

Add also this "We use nil when compiling and parsing expressions, if we have this method, we can easily use the compiler that the class defines for syntax highlight and code completion (without needing to add an if to check for nil)"

PR #5153, opened 2019-11-17
HandMorph cleanups required: remove three ivars

### LINKED ISSUE #2564: HandMorph cleanups required

MarcusDenker:
remove unreferenced keyboardListeners eventSource lastSystemEvent

fixes #2564

### DIFF
--- a/src/Morphic-Core/HandMorph.class.st
+++ b/src/Morphic-Core/HandMorph.class.st
@@ -11,7 +11,6 @@ Class {
 		'keyboardFocus',
 		'eventListeners',
 		'mouseListeners',
-		'keyboardListeners',
 		'mouseClickState',
 		'mouseOverHandler',
 		'lastMouseEvent',
@@ -27,8 +26,6 @@ Class {
 		'lastEventBuffer',
 		'lastKeyScanCode',
 		'combinedChar',
-		'eventSource',
-		'lastSystemEvent',
 		'captureBlock',
 		'recentModifiers'
 	],


MarcusDenker:
before merging, resulting mac artefact should be manually tested (see issue tracker entry)

Ducasse:
BTW you will have to explain to me why if one slave fails and the other pass we should not integrate because I do not get then the rationale of running on three in parallel.
I did not get the UI showing the results.

MarcusDenker:
I tested the mac image: works!

MarcusDenker:
For slave: if a slave is *down* (that is, the CI setup can not start a virtual slave for that platform), then the *whole* CI process does not work. For example, the merge will not trigger a new build.

And I am not sure if the CI, when it waits for a build, shows the failing tests of the other builds correctly (e.g, last week a change was merged in that situation that broke lots of tests).

My rule is: if the CI setup is broken, I stop to merge.

MarcusDenker:
But this means: "the CI setup is broken", not "one test run fails on one slave". These are two different things.

MarcusDenker:
e.g. if the virtual disk is full on a slave (happens sometimes) it just fails. No problem to merge a if the change is not related to specially that platform in this case.

Ducasse:
Ok you will show me how I can see that the slave is down. Because I confused it with a Vm crashed and one execution did not happen.

jecisc:
I just tried and the image did not crash. 

The weird thing is that my startup preferences only loaded half of the stuff and this without error... I don't know if it comes from this fix or from a missing information in the bootstraped image.

MarcusDenker:
I suggest we try and merge...

jecisc:
The current Pharo 8 works fine for me for now.

PR #5156, opened 2019-11-18
5155-Compiler-callPlugins-should-be-done-after-parse-not-before-compile

### LINKED ISSUE #5155: Compiler: #callPlugins should be done after parse, not before compile
Right now we call #callPlugins as part of compile.

this has the problem that if you ask for the AST of a method and then compile it without the involvement of the compiler facade, the AST is not transformed.

A solution is to do it after parse. This in addition makes sure that the AST that we use for compiling is the same as we see when asking for #ast, which should make mapping AST->text  (and BC->text) work, which is now broken.

MarcusDenker:
move #callPlugins to parsing. fixes #5155

### DIFF
--- a/src/OpalCompiler-Core/OpalCompiler.class.st
+++ b/src/OpalCompiler-Core/OpalCompiler.class.st
@@ -250,7 +250,7 @@ OpalCompiler >> callPlugins [
 	| plugins  |
 	plugins := compilationContext astTransformPlugins ifEmpty: [ ^self ].
 	plugins sort: [:a :b | a priority > b priority]. "priority 0 is sorted last"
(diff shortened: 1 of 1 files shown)


MarcusDenker:
SpCodeAdapterTest>>testTextWithStyle fails.

MarcusDenker:
I have no idea why this fails.

MarcusDenker:
ah... it's the change to visit the methodNode in the callPlugin... that is no good (as we want to keep "ast" to be returned the expression if we parse expressions).

MarcusDenker:
Discussed with Manuel and we found this is not a good idea as it is unclear what plugins can do to mess up code highlighting and general mapping.

And for his problem we found a better solution

PR #5161, opened 2019-11-18
Replace-assert-equals-from-packages-starting-with-G

jecisc:
Replace assert = in packages starting by G.

Replace:

- deny = by deny: equals:


astares:
@jecisc Is there a transformation script you use? We should apply this also to some external packages...

PR #5165, opened 2019-11-19
new-vm-identifying-methods

estebanlm:
introduce methods to indentify new properties of the VM.
adapt dely ticker to take same time as the idleTime (otherwise it has no sense, the tick will always happen even when there is no need for it).


estebanlm:
damn, I made the branch in wrong point

bencoman:
I guess you'll want to make the same change to DelayMicrosecondTicker.  I presume its still the active one.  Check with World > System > Process Browser and look at the name of the priority 80 process.

PR #5167, opened 2019-11-19
introduce methods to indentify new properties of the VM

estebanlm:
adapt dely ticker to take same time as the idleTime (otherwise it has no sense, the tick will always happen even when there is no need for it).

(second attempt)


bencoman:
I guess you'll want to make the same change to DelayMicrosecondTicker. I presume its still the active one. Check with World > System > Process Browser and look at the name of the priority 80 process.

estebanlm:
now this is ready to merge

MarcusDenker (approved):

PR #5172, opened 2019-11-19
Debugger selector

dupriezt:
This pull requests adds multiple things.

## DebugSessionStorage
**Description**
When a DebugSession is created (typically when an exception is signaled but not caught, before it is debugged), `DebugSessionStorage` stores it. To avoid cluttering the image with tons of processes (because storing a DebugSession means keeping the process it is debugging), `DebugSessionStorage` has a (configurable) limit on how many DebugSessions it can hold. When the limit is exceeded, the oldest DebugSessions are terminated (along with the process they debug) to make room. The exception is if a debugger is currently open on a DebugSession. A DebugSession currently opened in a debugger is not terminated, and does not count towards the storage limit. It will go back to counting towards the limit if the debugger in which it was opened is closed.

**The Point**
This allows to reopen previously closed DebugSessions, and to do so with a different debugger. The ability to open DebugSessions with a different debugger already exists in some fashion, but it is internal to the GTDebugger (you can see it at play if you click "Bytecode" in the top right corner of the GTDebugger, it will switch the debugger to the GTBytecodeDebugger), so it is not easy to support new non-GT debuggers.

**Additional Detail on DebugSessions life-cycle**
To maintain a DebugSession alive, a debugger has to call `#keepAlive:` on it (with itself as argument). To release the DebugSession, it has to call `#stopKeepingAlive:` (again, with itself as argument).

## DebuggerSelector
**The Point**
The idea of `DebuggerSelector` is to have a generic system for choosing "the most suitable" debugger to open a given DebugSession. A system like this is already in place in some fashion, but it is internal to the GTDebugger (you can see it at play when you debug a test: the debugger looks a bit different and shows test-specific information, because it's actually the GTTestDebugger that was opened), so it is not easy to add new non-GT debuggers.

**Short Description**
`DebuggerSelector` finds all the debugger classes of the system, ask them how they rank the DebugSession to be opened, and opens the DebugSession with the highest-ranking debugger. If this opening fails (because the debugger signals an Error), then `DebuggerSelector` tries to open the DebugSession with the second-highest ranking debugger and so on...

**What is a Debugger class?**
A debugger class is a class that implements all the methods from the AbstractDebugger class that are marked as SubclassResponsibility. For now, that is:
- `iAmADebugger`: This method simply needs to contain the `<iAmADebugger>` pragma, and is just there as a marker.
- `rankDebugSession: aDebugSession`: This method returns the rank of this debugger on the provided DebugSession. If the debugger absolutely cannot open the DebugSession, it should signal the `DebugSessionCannotBeOpenedByThisDebugger` exception. For example, a debugger that is specialised to debug test executions will return a high number if the DebugSession is that of a test execution, and signal the exception if it is not.
- `openOn: aDebugSession withFullView: aBool andNotification: aString`: This method should open the provided DebugSession in that debugger. It is a good place to call `#keepAlive:` on the DebugSession to ensure it does not get terminated by `DebugSessionStorage` while the debugger is open on it.
An important point is that debugger classes do not need to inherit from the `AbstractDebugger` class to be considered as debugger classes. They only need to implement the three (class-side) methods detailed above.

**More Details on the Debugger Selection Algorithm**
The algorithm outlined above gives the main idea. In addition to that:
- Debugger classes that signal the `DebugSessionCannotBeOpenedByThisDebugger` exception when asked to rank the DebugSession cannot be selected
- `DebuggerSelector` has a `mainDebuggerClass` variable. If a debugger class is in this variable, `DebuggerSelector` will always attempt to open the DebugSession with it first, and only if that signals an Error will it fall back on the default algorithm. A use for this is to manually set a specific debugger for images in production mode, for example one that only logs the exception in a file but does not open a graphical debugger allowing the user to modify code.

## User Interface
This pull request includes a user interface to show the list of all debugger in the image, show all the stored DebugSession, and interact with them.
DebuggersAndDebugSessionList
The left list shows all the debuggers in the image, along with their rank when it does not depend on the DebugSession. The buttons under it respectively allow to browse the selected debugger class, and set/unset it as main debugger.
The right list shows all the stored DebugSession, along with the number of debuggers open on them ("Refs" column) and the time at which they were stored (from most to least recent). The field in the top right allows to change the storage limit. The buttons under the list respectively allow to:
- Debug the selected DebugSession with the debugger the image would normally choose (using `DebuggerSelector`)
- Debug the selected DebugSession with the selected debugger
- Inspect the selected DebugSession
- Terminate (and remove from the storage) the selected DebugSession
- Terminate (and remove from storage) all the DebugSessions

This UI is available in the world menu.
DebugggersListAndDebugSessionsListWorldMenuItem

## Other Changes
- `MorphicUIManager`: when an uncaught exception has to be debugged, it was calling the debugger of the image (typically the GT debugger) to open the corresponding DebugSession. In this pull request, it tells `DebugSessionStorage` to store the DebugSession, and it tells `DebuggerSelector` to select the most suitable debugger and to open the DebugSession with it.
- `GT Debugger`: this pull requests adds the required methods for the GT debugger (and its SUnit and Bytecode variants) to be recognised as debugger classes.


dupriezt:
Here is the pull request, @StevenCostiou .
@MarcusDenker

Ducasse:
This looks cool :)
Did you copy this great description in a class or package comment somewhere?

Ducasse:
I will do a real code review when I have a moment :)

dupriezt:
> This looks cool :)
> Did you copy this great description in a class or package comment somewhere?

Done

estebanlm:
I think this will be integrated in NewTools

PR #5183, opened 2019-11-20
Replace-assert-equals-from-packages-starting-with-S

jecisc:
Replace assert = in packages starting by S.

Replace:

- deny = by deny: equals:


svenvc:
These were done upstream for STON, but they should still be applied to Pharo 8 as in this PR

jecisc:
Cool, thanks @svenvc !

MarcusDenker (approved):

PR #5186, opened 2019-11-20
Replace-assert-equals-from-packages-starting-with-Z

jecisc:
Replace assert = in packages starting by Z.

Replace:

- deny = by deny: equals:


svenvc:
Go ahead for Zinc/Zodiac, I will sync upstream later

MarcusDenker (approved):

PR #5194, opened 2019-11-21
Simplify and speed up  #classVarNames and #allClassVarNames

MarcusDenker:
- do not sort #classVarNames
-  #allClassVarNames  does not need to be a Set as the class vars are unique over the hierarchy
- implement #allClassVarNames simpler in the style of #allInstVarNames
- remove a #asOrderedCollection from the code completion as it is not needed anymore


MarcusDenker:
[SmalltalkImage allClassVarNames] bench 

Before:
124457.417 per second

After:
820793.283 per second

>6 times faster

pavel-krivanek (approved):
looks ok

MarcusDenker:
sadly there are tests failing

testDefinitionString
testClassSelected
testNormalKernelClasses

due to expecting sorted class variables...

svenvc:
sort/convert before testing should help

MarcusDenker:
Now i changed MC to not sort... but maybe this is not good as the class vars are stored in a dictionary which means that the order is not guaranteed... 

So not yet a good solution.

I think I might break this into multiple parts: first just not use Set and simplify (but still sort), then have a look at sorting...

gcotelli:
Keep in mind that the order of the class vars do not change randomly when written in Tonel format. Because this causes unexpected changes when commiting code.

MarcusDenker:
Yes... I will close this PR and do the "class vars are not a set" simplification...

PR #5195, opened 2019-11-21
ProtoObject: #mustBeBooleanDeOptimize

MarcusDenker:
if we move #mustBeBoolean to ProtoObject, the mechanism to get #ifTrue: sends will work for subclasses of ProtoObject. As this is used often for proxies, having that mechanism in ProtoObject makes sense, I think.


noha (approved):

MarcusDenker:
aargh... committed to the wrong PR

MarcusDenker:
so how can I now remove that PR?

MarcusDenker:
ah, ok, reverting worked.

PR #5196, opened 2019-11-21
TestExecutionEnvironmentWithoutCleaning

pavel-krivanek:
introduce  TestExecutionEnvironmentWithoutCleaning and TestExecutionEnvironmentWithWorldStepping

### DIFF
--- a/src/SUnit-Core/DefaultExecutionEnvironment.extension.st
+++ b/src/SUnit-Core/DefaultExecutionEnvironment.extension.st
@@ -4,7 +4,7 @@ Extension { #name : #DefaultExecutionEnvironment }
 DefaultExecutionEnvironment >> runTestCase: aTestCase [
 
 	| testEnv |
-	testEnv := TestExecutionEnvironment new.
+	testEnv := aTestCase executionEnvironmentClass new.
 	
 	testEnv beActiveDuring: [
 		testEnv runTestCase: aTestCase]
--- a/src/SUnit-Core/DefaultExecutionEnvironment.extension.st
+++ b/src/SUnit-Core/DefaultExecutionEnvironment.extension.st
@@ -9,9 +9,3 @@ DefaultExecutionEnvironment >> runTestCase: aTestCase [
 	testEnv beActiveDuring: [
 		testEnv runTestCase: aTestCase]
 ]
-
-{ #category : #'*SUnit-Core' }
-DefaultExecutionEnvironment >> runTestsBy: aBlock [
-
-	TestExecutionEnvironment new beActiveDuring: aBlock
-]
--- a/src/SUnit-Core/TestExecutionEnvironmentWithoutCleaning.class.st
+++ b/src/SUnit-Core/TestExecutionEnvironmentWithoutCleaning.class.st
@@ -1,0 +1,25 @@
+"
+The standard TestExecutionEnvironment kills all the processes forked during the test execution when the test stops without errors or with a time-out exception. That makes debugging of tests that use side-processes very hard because if such a process causes an error.
+
+This execution environment does not clean the process so you need to take responsibility for them. 
+
+-------------------------------------------------------------------
+  Use this testing environment only during debugging of the tests
+-------------------------------------------------------------------
+"
+Class {
+	#name : #TestExecutionEnvironmentWithoutCleaning,
+	#superclass : #TestExecutionEnvironment,
+	#category : #'SUnit-Core-Kernel'
+}
+
+{ #category : #controlling }
+TestExecutionEnvironmentWithoutCleaning >> checkForkedProcesses [
+]
+
+{ #category : #controlling }
+TestExecutionEnvironmentWithoutCleaning >> prepareForNewProcess: aProcess [
+
+	forkedProcesses add: aProcess.
+
+]
--- a/src/SUnit-Core/TestSuite.class.st
+++ b/src/SUnit-Core/TestSuite.class.st
@@ -144,7 +144,7 @@ TestSuite >> run: aResult [
 { #category : #running }
 TestSuite >> runTestsInTestingEnvironments: aResult [
 	self setUp.
-	[ self tests
+	[ self allSubtests 
 		do: [ :aTestCase | 
 			CurrentExecutionEnvironment 
 				useTestingDuring: [
--- a/src/SUnit-Core/TestSuite.class.st
+++ b/src/SUnit-Core/TestSuite.class.st
@@ -40,6 +40,13 @@ TestSuite >> addTests: aCollection [
 	aCollection do: [:eachTest | self addTest: eachTest]
 ]
 
+{ #category : #accessing }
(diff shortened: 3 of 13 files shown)


pavel-krivanek:
Introduce special testing environments that do not clean the processes created during the test execution and allow Morphic UI to be responsive during testing.

The motivation for it is to be able to make debugging of tests that use external processes or write something to Transcript easier. 

These execution environments are supposed to be used only during the debugging of the tests.. See the class comments for details.

src/SUnit-Core/DefaultExecutionEnvironment.extension.st
```diff
@@ -4,7 +4,7 @@ Extension { #name : #DefaultExecutionEnvironment }
 DefaultExecutionEnvironment >> runTestCase: aTestCase [
 
 	| testEnv |
-	testEnv := TestExecutionEnvironment new.
+	testEnv := aTestCase executionEnvironmentClass new.
```
dionisiydk, line 7: I like the idea to have specific behavior for execution of given test class. 
But this redirection to test is not enough. It works only for single test but full test suite will ignore it:
```Smalltalk
TestSuite>>run: aResult
   TestExecutionEnvironment new beActiveDuring: [...]
```
Tests here will not be running under defined execution environment. All of them will be managed by same environment instance. And it will be inconsistent to single test run.

I would better redirect to test case the configuration of execution environment. So it will be always performed in same way for given test.
But it would require refactoring of TestExecutionEnvironment to reify such configuration which will allow to disable the process management or inject the extra logic (like UI loop).
I imagine this as a special kind of test resources which will live for single test by registering in execution environment. 
But maybe different concept will be better for this purpose.

dionisiydk: As continue to my idea I would like to be able to avoid code in tests like:
```Smalltalk
file := 'myTestFile' asFileReference.
[file writeStreamDo: [:s | ...].
...
self assert: ...
] ensure: [file ensureDelete].
```
Or
```Smalltalk
server := MyWebserver new.
[server start.
...
self assert: ...
] ensure: [server stop]
```
Currently we are managing such cases by extracting "instantiation" into setUp, storing "resource" in instance variable, and extracting "cleanup" into tearDown. But it is always quite boring work especially when it is just a single test. Sometimes we are even introducing separate test class to correctly manage such resources.
It can be simplified like:
```Smalltalk
self addResource: [file := 'myTestFile' asFileReference] releaseBy: [file ensureDelete]
file writeStreamDo: [:s | ...].
...
self assert:...
```
and 
```Smalltalk
self addResource: [server := MyWebserver new.] releaseBy: [server stop].
server start.
...
self assert:...
```
From this perspective the current process management is automatic registration of such per-test resources where every forked process is a resource.

I do not ask you to implement it. I just want to inspire your thoughts related to this PR work.

src/SUnit-Core/TestExecutionEnvironmentWithoutCleaning.class.st
```diff
@@ -0,0 +1,25 @@
...
+
+This execution environment does not clean the process so you need to take responsibility for them. 
+
+-------------------------------------------------------------------
+  Use this testing environment only during debugging of the tests
+-------------------------------------------------------------------
+"
+Class {
+	#name : #TestExecutionEnvironmentWithoutCleaning,
+	#superclass : #TestExecutionEnvironment,
```
dionisiydk, line 12: I do not understand why you decided to introduce new class after adding #wantsCleanProcesses flag in the base class.

pavel-krivanek: ahh, wantsCleanProcesses  is a leftover from the original experiments. I'm removing it

src/SUnit-Core/TestSuite.class.st
```diff
@@ -127,19 +127,22 @@ TestSuite >> run [
...
-			self changed: each ] ]
-		ensure: [ self tearDown ]
+		do: [ :aTestCase | 
+			| testEnv |
+			testEnv := aTestCase executionEnvironmentClass new.
+			testEnv beActiveDuring: [
+				aTestCase run: aResult.
+				self announceTest: aTestCase.
+				self changed: aTestCase ] ] 
+	] ensure: [ self tearDown ]
```
dionisiydk, line 155: The single environment per suite was an optimization to avoid overhead for every test during a big test suite.
For example each environment organizes a watch dog process to control the test running time.

pavel-krivanek: I see, maybe we should check if all the tests in the suite have the same environment class and split them only if it is not the case...

pavel-krivanek:
This is an interesting failure:
 ```
Found unknown processes: 

a Process in [ self schedule. "It is critical that the following has no suspension point so that the test and the wait primitive are atomic. In addition, if the delay is no longer being waited on while entering the way we know that it is expired because the delay has already fired." beingWaitedOn ifTrue: [ delaySemaphore wait ] ifFalse: [ expired := true ] ] in DelayWaitTimeout>>wait, 

a Process in [ "waiting new test case" watchDogSemaphore wait. "waiting while test completes" [ timeIsGone := watchDogSemaphore waitTimeoutMSecs: maxTimeForTest asMilliSeconds. testCompleted ] whileFalse: [ "this subloop allows to dynamically change time limit and restart watch dog" timeIsGone ifTrue: [ "The main purpose of following condition is to ignore timeout when test is under debug. Test process is suspended only when it is debugged" testProcess isSuspended ifFalse: [ testProcess signalException: TestTookTooMuchTime new ] ] ] ] in TestExecutionEnvironment>>watchDogLoopFor:, a Process in [ delaySemaphore wait ] in Delay>>wait
```

That makes sense because the running of the test execution environment watchdog generates them. But then I do not understand why this test was not failing before. 
@dionisiydk Do you have some explanation? It looks like the environment was not active for our tests in console HDTestReport until now

Ducasse:
A side remark:
this is cool to see new ideas and ways to make things simpler. 
Now we should pay attention that companies may have a lot and lot of tests like 100,000 of them. So this is important that running these tests are not more prohibitive that they are now.

dionisiydk:
@pavel-krivanek Sorry, but I do not like this change. It is too complex. I can't say that I understand it.
I believe the right approach is to keep single class for test environment but allow test cases to configure it. It would be much simpler than proposed logic.

estebanlm:
For Pharo9.0 now :)

guillep:
Hi @pavel-krivanek can we resume the state of this PR?
Last activity is from december.

dionisiydk:
Hi Guille.

I am working on it.

вс, 15 мар. 2020 г., 13:29 Guille Polito <notifications@github.com>:

guillep:
I'll close this PR. This does not close the original branch of course.
So as soon as you have something else, please open a new PR.

Thanks

PR #5197, opened 2019-11-21
Improve-AssertEquals-Rule

jecisc:
Improve assert:equals: rule to avoid assert: =

It now matches more cases:

- assert: = with assert: equals:
- deny = with deny: equals:
- assert: == with assert: identicalTo:
- deny: == with deny: identicalTo:


astares:
The class name does not fit to the deny: messages.

Why not create a rule class per message?

Ducasse:
Failure is not related. Thanks Cyril. 
I would also like a rule that says that we should not refer to Smalltalk directly.

PR #5198, opened 2019-11-21
Replace-assert-equals-from-non-test-methods

jecisc:
Replace assert = in all non test methods from TestCase subclasses.

Replace:

- deny = by deny: equals:


Ducasse:
Strange nothing worked

MarcusDenker:
failing tests not related

PR #5203, opened 2019-11-22
Cleanup #allClassVarNames

MarcusDenker:
- classVarNames does not need to be a set as the variables are unique in the hierarchy
- simplify #allClassVarNames to be implemented in the style of #allInstVarNames
- remove an asOrderedCollection in the code copletion code that is not needed anymore


Ducasse:
Happy  that we revisit all the corners.

PR #5204, opened 2019-11-22
#flatCollect: implement using #streamContents:

MarcusDenker:
#flatCollect: should be implemented using streamcontents.

### DIFF
--- a/src/Collections-Abstract/Collection.class.st
+++ b/src/Collections-Abstract/Collection.class.st
@@ -712,9 +712,10 @@ Collection >> flatCollect: aBlock as: aCollectionClass [
 	of collection for each element. Equivalent to the lisp's mapcan"
 
 	| col |
-	col := OrderedCollection new: self size.
-	self do: [ :each | col addAll: (aBlock value: each) ].
-	^ aCollectionClass withAll: col 
+	col := OrderedCollection
+		new: self size
+		streamContents: [ :str | self do: [ :each | str nextPutAll: (aBlock value: each) ] ].
+	^ aCollectionClass withAll: col
 ]
 
 { #category : #enumerating }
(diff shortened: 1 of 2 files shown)


Ducasse:
I have a stupid question but I dare asking it. 

when we do 
col := OrderedCollection
 		new: self size
 		streamContents: [ :str | self do: [ :each | str nextPutAll: (aBlock value: each) ] ]

The stream is preallocated but what will happen if the contents we will add via the nextPutAll: is larger? I imagine that the internal stream collection is copied into a larger one.

Ducasse:
BTW how much faster is this new implementation?
It would be nice to give an indication in the method comments to educate the masses.

MarcusDenker:
For 

new:streamContents:

the stream is initialised with a collection of size. If this estimate was not correct, we will create a larger collection, no problem there. The good thing is that it is done only once (we use the stream to avoid constantly having to make the collection larget, which is slow).

Sven added a really nice optimisation to new:streamContents: so that *if* the size estimate is correct, we do not need to copy the internal collection that the stream used.

MarcusDenker:
benchmarks I did not yet do. I wanted to do that before deciding if the change is worth it

svenvc:
I would change #flatCollect:as: to NOT use an intermediary collection

#withAll: is defined in terms of #addAll: so it would be possible to do it directly: instanciate the target collection, loop over the receiver and #addAll: each element

Agreed ?

MarcusDenker:
it sadly does not work for String... as it does not allow #add:

MarcusDenker:
I think I will make version if this PR with *only* the change to flatCollect:, as the code there is the same as we have in streamContents:

svenvc:
Yes, I was wrong, it won't work for Array either, sorry for the noise

PR #5206, opened 2019-11-22
UseNecEntrySubclasses

MarcusDenker:
instantiate subclasses of NECEntry so we know what a completion entry stands for (so we can highlight it and make the browser work)


myroslavarm (approved):
this is really cool refactoring!

PR #5207, opened 2019-11-22
[donotintegrate] attempt to use improved string comparison primitive 

hogoww:
see what ci says

### DIFF
--- a/src/Collections-Strings/ByteString.class.st
+++ b/src/Collections-Strings/ByteString.class.st
@@ -17,7 +17,7 @@ ByteString class >> compare: string1 with: string2 collated: order [
 	"Return 1, 2 or 3, if string1 is <, =, or > string2, with the collating order of characters given by the order array."
 
 	| len1 len2 c1 c2 |
-	<primitive: 'primitiveCompareString' module: 'MiscPrimitivePlugin'>
+	<primitive:158>
 	
 	<var: #string1 declareC: 'unsigned char *string1'>
 	<var: #string2 declareC: 'unsigned char *string2'>


hogoww:
Interestingly enough, using the 158 primitive (compareString primitive) seems to work without more integration than changing the primitive.
We'll have to do some benchmarks on Moose too

jecisc:
This is really weird all tests are passing I think because the two primitives returns different values.

```Smalltalk
a := 'Test'.
b := 'Test'.

a compareWith: b. "0"
ByteString compare: a with: b collated: String newAsciiOrder. "2"

b := 'AAAA'.

a compareWith: b. "19"
ByteString compare: a with: b collated: String newAsciiOrder. "3"
b := 'ZZZZ'.

a compareWith: b. "-38"
ByteString compare: a with: b collated: String newAsciiOrder. "1"
```

And checking the code of #< & co, it feels really weird all tests are green.

```Smalltalk
< aString 
	"Answer whether the receiver sorts before aString.
	The collation order is simple ascii (with case differences)."

	^ (self compare: self with: aString collated: AsciiOrder) = 1
```

Well, at least we know the new primitive is present in Pharo 8 and also Pharo 7! (I just checked).

jecisc:
Here is a little bench comparing two strings of the same size but different elements

image

VincentBlondeau:
@jecisc maybe some tests are missing?

Ducasse:
Cyril you just changed compareWith: to call the primitive?
Because this is indeed really strange that nothing break I will look at the sender of compareWith:

Ducasse:
Ok I saw what you did.

Ducasse:
May be one trick is to change all the users of compare:with:collated: par threeWayCompareTo:
then change compare:with:collated: to use the new primitive and then to revert the changes.

jecisc:
It’s not that simple. 

Currently, comparison is a class side method #compare:with:collared: with two versions.
A version on ByteString it comparing two ByteStrings and a version on String used when one of the two is not a bytestring. 

In the new version, you have a pronostics working instance side. So you need to add a method #compareWith: using the primitive 158. This primitive will use a default order. If you want another order you need to also add a method #compareWhit:order:.

Then the senders of #compare:with:collated: needs to be updated. To use #comparewith:(order:) and to manage positive, negative or 0 as return values instead of 1, 2 or 3..

I have an image with some of the changes. I plan to do a PR with all of them next week since I already started to do it. 

Also for info, it is normal that we have positive, negative and 0 instead of -1, 0 and 1 as announced in Clement blog post because it is faster this way. So Sophie and Clement choose the version that is the more efficient.

Ducasse:
I did the following experience

threeWayCompareTo: aString
	"Do a three-way comparison between the receiver and anotherObject, returning
	-1 if self < anotherObject
	0 if self = anotherObject
	1 if self > anotherObject
	This assumes a total order in accordance with the mathematical law of trichotomy.
	See also:  http://en.wikipedia.org/wiki/Three-way_comparison"
	<primitive: 158>
	" ^ (self compare: self with: aString collated: AsciiOrder) - 2 "
	
	
testStrings
	| a b c d |
	a := 'xc'.
	b := 'xd'.
	c := 'xd'.
	d := 'xa'.
	self assert: (b threeWayCompareTo: c) isZero.
	self assert: (a threeWayCompareTo: b) equals: -1.
	self assert: (b threeWayCompareTo: a) equals: 1.
	self assert: (a threeWayCompareTo: d) equals: 1.
	self assert: (d threeWayCompareTo: a) equals: -1
	
	self assert: (a threeWayCompareTo: d) equals: 1.
		got 2 instead of 1

I have to dig further to understand. 
I was surprised that I could also put a halt into String <<  < aString without breaking the system

hogoww:
I'm an idiot. 
Tickled me all week end but I couldn't put my finger on it.
I changed the primitive on the class side, so it only used the fail code....
I do not have this branch anymore, but i'll do another pr.
Thankfully Cyril has a real benchmark !

hogoww:
See  #5219 for the follow up. This one won't be changed anyway

PR #5209, opened 2019-11-23
Make SlotClassDefinition the default

MarcusDenker:
- enable the slot class definitin form by default

I want to for one see if all tests are green, in addtion I think it might be a good idea to have  this on for the next release

### DIFF
--- a/src/Slot-Core/Slot.class.st
+++ b/src/Slot-Core/Slot.class.st
@@ -81,7 +81,7 @@ Slot class >> named: aSymbol [
 
 { #category : #settings }
 Slot class >> showSlotClassDefinition [
-	^slotClassDefinition ifNil: [ slotClassDefinition := false ]
+	^slotClassDefinition ifNil: [ slotClassDefinition := true ]
 ]
 
 { #category : #settings }


jecisc:
I agree! During Pharo 7 and 8 development I think I fixed some failing tests around that, but IIRC there should still be failures.

jecisc:
I agree! During Pharo 7 and 8 development I think I fixed some failing tests around that, but IIRC there should still be failures.

Ducasse:
For the record I'm working on a class definition parser but it will really not be for now.

Ducasse:
There are broken tests.
In fact while thinking about it. I would not produce Pharo 80 with it because we will change the syntax again. I think that what we could do is to make sure that everything works with it on but do not expose users to it. In addition it means that suddenly all the books will be outdated. I can change all the books but I would prefer to wait for the new syntax.

Ducasse:
I was thinking about this change and I would not do it. we can enable it in a preference to test and use it but I would like to avoid to change two times syntax. 
Else it really means that we will have to change everything twice.

MarcusDenker:
yes, true... I close this and do a PR with just the failing tests fixed

PR #5219, opened 2019-11-25
primitiveString change

hogoww:
follow up of #5207.
Attempt to use the CI to do the change.
Fails locally


hogoww:
Thought string comparison wasn't (surprisingly) needed in the bootstrap. Will check with carolina again when she's at the office.

Ducasse:
just a question because I did not know. The primitive is checking the number of arguments?
Because I see that it is used in methods with different args. 

{ #category : #comparing }
 String >> compareWith: aString [
 	<primitive:158>
 	^ self primitiveFail  
 ]

 { #category : #comparing }
 String >> compareWith: aString collated: order [
 	<primitive:158>
 	^ self primitiveFail  
 ]

hogoww:
Yeup. if order is not passed it's set as nil, and the primitive is slightly optimised if i recall correctly.

hogoww:
Will check with Caro during the Sprint on friday

src/Collections-Strings/String.class.st
```diff
@@ -883,7 +895,7 @@ String >> compare: string1 with: string2 collated: order [
 	"('a000' asWideString sameAs: (ByteArray with: 97 with: 0 with: 0 with: 0) asString) >>> false"
 	
 	(string1 isByteString and: [string2 isByteString]) ifTrue: [
-		^ ByteString compare: string1 with: string2 collated: order].
+		^ string1 compareWith: string2 collated: order].
```
guillep: Doing this change froze everything.

Also, can we blindly do this? I understand that the old primitive returns values 1, 2 and 3 for smaller, equal, bigger. And the second one returns negative, 0 and positive for smaller, equal, bigger. But there is no transformation here between these...

For example, to make this work I had to use instead

```smalltalk
(string1 compareWith: string2 collated: order) sign + 2
```

Am I missing something?

guillep (changes requested):
Looking at the CI I'd say that the bootstrap relies on string comparison at some point. For sure the dependency is not structural, but probably related to a reflective call to the class builder (that's why the failure is seen in the AST interpreter). I'm sure @carolahp can help in debugging this with here debugger.

src/Collections-Strings/String.class.st
hogoww: Not at all, I just wanted to go incrementally. 
Also, I wasn't aware of the #sign selector, which makes the change very readable. Thanks :)

Also, we are planning to pair program on the issue next week, try out her tools !

guillep: Sure. My point is: how did you do this change without exploding your image? because as soon as you do this change, nothing else works :)

hogoww: There's not only Image modification :D
emacs String.class.st 
*whistle*

hogoww:
Worked on it with @carolahp 

She was able to see first that the problem was the primitive itself that was failing.
Then, she discovered that was because the primitive wasn't registered in the Simulator's primitive table of the VM, and therefore failed because the corresponding function was unknown.
But it was registered in the VM primitive table.

So, the simulator for the VM just needs to be updated to a version knowing about the primitive.
Hence, the Bootstrap's simulator should be updated =]

bencoman:
The synchronisation of Simulator and VM primitive tables is something that would eb good to know more about.
Could you do a post to the mail list?

estebanlm:
For Pharo9.0

hogoww:
ci fail seems unrelated

Error while sending email: javax.mail.SendFailedException: Invalid Addresses;
  nested exception is:
	com.sun.mail.smtp.SMTPAddressFailedException: 452 Too many recipients received this hour

Guess i'll retry later

hogoww:
Actually related, but I do not know how to proceed;
On linux build
32] + popd
[32] + ./pharo Pharo.image --no-default-preferences /builds/workspace/uest_and_branch_Pipeline_PR-5219/32/bootstrap/scripts/prepare_image.st --save --quit
[32] Could not open the Pharo image file `Pharo.image'.


Locally on a mac:
I gave permission to execute from anywhere

+ popd
+ ./pharo Pharo.image --no-default-preferences /Users/missegarciagenereuse/Desktop/pierre/pharo/bootstrap/scripts/prepare_image.st --save --quit
/BuildRoot/Library/Caches/com.apple.xbs/Sources/AppleFSCompression/AppleFSCompression-96.200.3/Common/ChunkCompression.cpp:49: Error: unsupported compressor 8
/BuildRoot/Library/Caches/com.apple.xbs/Sources/AppleFSCompression/AppleFSCompression-96.200.3/Libraries/CompressData/CompressData.c:353: Error: Unknown compression scheme encountered for file '/System/Library/CoreServices/CoreTypes.bundle/Contents/Resources/Exceptions.plist'
/BuildRoot/Library/Caches/com.apple.xbs/Sources/AppleFSCompression/AppleFSCompression-96.200.3/Common/ChunkCompression.cpp:49: Error: unsupported compressor 8
/BuildRoot/Library/Caches/com.apple.xbs/Sources/AppleFSCompression/AppleFSCompression-96.200.3/Libraries/CompressData/CompressData.c:353: Error: Unknown compression scheme encountered for file '/System/Library/CoreServices/CoreTypes.bundle/Contents/Library/AppExceptions.bundle/Exceptions.plist'
/BuildRoot/Library/Caches/com.apple.xbs/Sources/AppleFSCompression/AppleFSCompression-96.200.3/Common/ChunkCompression.cpp:49: Error: unsupported compressor 8
/BuildRoot/Library/Caches/com.apple.xbs/Sources/AppleFSCompression/AppleFSCompression-96.200.3/Libraries/CompressData/CompressData.c:353: Error: Unknown compression scheme encountered for file '/System/Library/CoreServices/CoreTypes.bundle/Contents/Resources/UnknownFSObjectIcon.icns'
/BuildRoot/Library/Caches/com.apple.xbs/Sources/AppleFSCompression/AppleFSCompression-96.200.3/Common/ChunkCompression.cpp:49: Error: unsupported compressor 8
/BuildRoot/Library/Caches/com.apple.xbs/Sources/AppleFSCompression/AppleFSCompression-96.200.3/Libraries/CompressData/CompressData.c:353: Error: Unknown compression scheme encountered for file '/System/Library/PrivateFrameworks/FinderKit.framework/Resources/Base.lproj/SidebarView.nib'
/BuildRoot/Library/Caches/com.apple.xbs/Sources/AppleFSCompression/AppleFSCompression-96.200.3/Common/ChunkCompression.cpp:49: Error: unsupported compressor 7
/BuildRoot/Library/Caches/com.apple.xbs/Sources/AppleFSCompression/AppleFSCompression-96.200.3/Libraries/CompressData/CompressData.c:353: Error: Unknown compression scheme encountered for file '/System/Library/PrivateFrameworks/FinderKit.framework/Resources/Base.lproj/ColumnView.nib'
/BuildRoot/Library/Caches/com.apple.xbs/Sources/AppleFSCompression/AppleFSCompression-96.200.3/Common/ChunkCompression.cpp:49: Error: unsupported compressor 8
/BuildRoot/Library/Caches/com.apple.xbs/Sources/AppleFSCompression/AppleFSCompression-96.200.3/Libraries/CompressData/CompressData.c:353: Error: Unknown compression scheme encountered for file '/System/Library/PrivateFrameworks/FinderKit.framework/Resources/Base.lproj/ArrangeByMenu.nib'

Any idea?

tesonep:
It looks like a problem with OSX. What version of the VM is? What version of OSX are you using? Did you try replacing the VM with the latest headless and see?

hogoww:
the macOs problem, yes, a problem with compression from a popd
On linux I don't know.

Currently the VM is just downloaded from the bootstrap's script.
But since i'm doing critical modifications on the image, that impact the bootstrap, I don't think that's coming from the vm

hogoww:
added support file (including the vm used by the image... I thought it was using a default one, sorry :/)
No change. But I think there's a path problem in my changes.

hogoww:
it was a stupid path problem, my bad. 
I'll finally be able to work on the real issue now that the storage of the image is fixed.

hogoww:
Current state: stuck in MonticelloBootstrap, a package name is uncapitalized seemingly out of nowhere. 
I'm adding debug information to understand why

hogoww:
Works locally on macOs mojave something !
(In time for Christmas too :D)
There's just 2 hacks that should be removed.

hogoww:
Unix's segfault on previous build may have been random, i'll look further tomorow.
macOs has the same error twice, i started looking already.

Regardless, more important people will now have to see how they want to proceed to integrate this ! ( if )

hogoww:
Tests run locally without errors on MacOS 32  Mojave 10.14.6.

Capture d’écran 2019-12-26 à 09 10 19

I don't know if the CI failures may or may not be related.

I spend a lot of time on this, I'm done for this one till I get back to work.
I'll still answer/correct if I can when asked.

Makefile~
```diff
@@ -0,0 +1,149 @@
+BOOTSTRAP_ARCH = 32
```
guillep, line 1: This temporary file shouldn't be here, should it?

Makefile
```diff
@@ -3,7 +3,7 @@ BUILD_NUMBER = 1
 
 ifeq ($(BOOTSTRAP_ARCH),32)
 	SEEK=24
-else
+else
```
guillep, line 6: Why this change?

bootstrap/scripts/2-download.sh
```diff
@@ -14,8 +14,12 @@ SCRIPTS="$(cd "$(dirname "${BASH_SOURCE[0]}")" ; pwd -P)"
 mkdir -p "${BOOTSTRAP_CACHE}" #required to generate hermes files
 
 ${BOOTSTRAP_REPOSITORY}/bootstrap/scripts/getPharoVM.sh 70
-wget --progress=dot:mega https://github.com/carolahp/PharoBootstrap/releases/download/v1.7.0/bootstrapImage.zip
-unzip bootstrapImage.zip
+wget --progress=dot:mega https://github.com/hogoww/BootstrapImage/archive/master.zip
```
guillep, line 17: Please, stop having different versions of these things everywhere :(
The bootstrap image should not change repository every 2 commits.

`https://github.com/hogoww/BootstrapImage`

bootstrap/scripts/2-download.sh~
```diff
@@ -0,0 +1,44 @@
+#!/usr/bin/env bash
```
guillep, line 1: This is also a temporary file that shouldn't be here.

src/Collections-Strings/String.class.st
```diff
@@ -883,7 +895,8 @@ String >> compare: string1 with: string2 collated: order [
 	"('a000' asWideString sameAs: (ByteArray with: 97 with: 0 with: 0 with: 0) asString) >>> false"
 	
 	(string1 isByteString and: [string2 isByteString]) ifTrue: [
-		^ ByteString compare: string1 with: string2 collated: order].
+		^ (string1 compareWith: string2 collated: order) sign + 2
```
guillep, line 898: Could we add a comment on why the "+2"? :)

src/Monticello/MonticelloBootstrap.class.st
```diff
@@ -60,7 +60,10 @@ MonticelloBootstrap >> bootstrapVersionsSatisfying: aBlock [
 
 	^ RPackageOrganizer default packages
 		select: [ :package | package name ~= RPackage defaultPackageName and: [ aBlock value: package ] ]
-		thenCollect: [ :rpackage | self fileNameForPackageNamed: rpackage name ]
+		thenCollect: [ :rpackage | "Hack, I don't know where this was added the wrong way"
+		rpackage name = #'zinc-resource-meta-core'
```
guillep, line 64: Why is this "hack" here?

src/Monticello/MonticelloBootstrap.class.st
```diff
@@ -85,8 +88,8 @@ MonticelloBootstrap >> fileNameForPackageNamed: name [
 
 	| versionInfos |
 	versionInfos := self localRepository directory files
-		select: [ :fileName | (fileName basename copyUpToLast: $-) = name ].
-	
+		     select: [ :fileName | (fileName basename copyUpToLast: $-) = name ].
```
guillep, line 91: Could we get rid of these changes?

src/Monticello/MonticelloBootstrap.class.st~
```diff
@@ -0,0 +1,170 @@
+"
```
guillep, line 1: This is a temporary file. Can we get rid of it?

guillep (changes requested):
This is not ready :). Too many silly mistakes easy to repair.
How can we make it move forward?

hogoww:
Actually, besides the garbage files, this is already usable. See my answers to the comments for the rest.
We have to work on this with pablo, to do a clean PR, and investigate the hack i had to do.
We already started, but could not go all the way through

Makefile~
hogoww: Nope, garbage file. I was so happy it was working I just stopped instantly after I did :D

Makefile
hogoww: Fail, should be removed. (of course)

bootstrap/scripts/2-download.sh
hogoww: Agreed. I did not know how put it as a ressource on Github, so I did what I had to do to make it work.
I do not have access to other's people github account to give them privilege of the image's bootstrap. Maybe it should have it's own repository, that'd be a lot easier to track changes too.
I did that over christmas break, alone, with zero knowledge of the boostrap. 
Point was to get it done, and see that it worked.

bootstrap/scripts/2-download.sh~
hogoww: yeup

src/Collections-Strings/String.class.st
hogoww: Of course, I'll do that on the reworked PR I'm doing with pablo

src/Monticello/MonticelloBootstrap.class.st
hogoww: Of course, it's just formating.

src/Monticello/MonticelloBootstrap.class.st~
hogoww: yeup

src/Monticello/MonticelloBootstrap.class.st
hogoww: Because we have to check if the name of the package is in lower cases instead of having upper cases at the start of the words.
This allowed it to work.
Now, is it because the new string primitive behaves in a different way ?
I don't know, this has to be investigated.
Having most pharo tests passing though, helps think that it may not be that, but then, why does it currently work? 

Btw pablo was thinking that this was because someone has used as a protocol name 
'*zinc-resource-meta-core'
instead of:
'*Zinc-Resource-Meta-Core'

hogoww:
This is currently stale, but I actually am not forgetting about it. Not a week has passed without me asking to someone what I had to do to have it accepted :D

Ducasse:
So first streamline it.

hogoww:
It was considered as not ready because of the messy commit history, hence the 
"We have to work on this with pablo, to do a clean PR, and investigate the hack i had to do.
We already started, but could not go all the way through"

guillep:
So should I close this one because you're working on another branch?

Of course closing the PR does not kill your branch, so no work should be lost.

hogoww:
I just wanted to keep this for documentation purpose, which it achieves by being closed too.
As you wish. For me this forces us to not forgetting about this, but I could not care less one way or the other.

guillep:
You can keep a link in the app you use to manage your TODOs :).

PR #5220, opened 2019-11-25
Completion: Fix emebdded Browser

MarcusDenker:
There is a setting to turn on an embedded mini browser in code completion. This PR makes it work for Globals and message sends (the most used use case). Some more work should be done for selectors, other kinds of vars. But it is usable.


myroslavarm (approved):
looks good!

PR #5221, opened 2019-11-25
Add class deprecation through meta-links

maenu:
I added a mechanism to use transformations for replacing deprecated classes that have a one-to-one replacement. It works by adding pragma to the class-side `#isDeprecated` method, e.g.
```
SomeDeprecatedClass class >> #isDeprecated 
    <transformTo: 'ReplacementClass'>
    ^ true
```
Once the pragma is added, all methods referencing the deprecated transformable class are instrumented. Before the reference to a deprecated class a meta-link is inserted that replaces the reference with the replacement expression (it uses the RB transformation language, hence it can be any expression, not just another class reference).

To make it work, I had to adapt two meta-link related methods in Calypso, as it may happen that a meta-link is uninstalled on a method that does not have a methodClass anymore. I could not reliably reproduce this though, but it seems to be happening when Calypso is refreshing system queries.

### DIFF
--- a/src/Calypso-SystemPlugins-Reflectivity-Queries/MetalinkChanged.extension.st
+++ b/src/Calypso-SystemPlugins-Reflectivity-Queries/MetalinkChanged.extension.st
@@ -13,14 +13,15 @@ MetalinkChanged >> affectsMethods [
 
 { #category : #'*Calypso-SystemPlugins-Reflectivity-Queries' }
 MetalinkChanged >> affectsMethodsDefinedInClass: aClass [
-
-	^link methods anySatisfy: [ :each | each origin == aClass]
+	^ link methods
+		anySatisfy: [ :each | each methodClass isNotNil and: [ each origin == aClass ] ]
 ]
 
 { #category : #'*Calypso-SystemPlugins-Reflectivity-Queries' }
 MetalinkChanged >> affectsMethodsDefinedInPackage: aPackage [
-
-	^link methods anySatisfy: [ :each | each package == aPackage]
+	^ link methods
+		anySatisfy:
+			[ :each | each methodClass isNotNil and: [ each package == aPackage ] ]
 ]
 
 { #category : #'*Calypso-SystemPlugins-Reflectivity-Queries' }
--- a/src/ClassDeprecation-Tests/ClassDeprecationClassTest.class.st
+++ b/src/ClassDeprecation-Tests/ClassDeprecationClassTest.class.st
@@ -1,0 +1,71 @@
+"
+I am a copy of class ClassDeprecationMetaLinkTest. This comment is copied from there, and might not be entirely accurate
+
+This class contains tests
+"
+Class {
+	#name : #ClassDeprecationClassTest,
+	#superclass : #ClassDeprecationTest,
+	#category : #'ClassDeprecation-Tests'
+}
+
+{ #category : #running }
+ClassDeprecationClassTest >> setUp [
+	super setUp.
+	referencingClass compile: referencingOriginalSource
+]
+
+{ #category : #tests }
+ClassDeprecationClassTest >> testDeprecationTransformationExpression [
+	self
+		assert: deprecatedClass deprecationTransformationExpression
+		equals: replacementClass name
+]
+
+{ #category : #tests }
+ClassDeprecationClassTest >> testInstallDeprecationTransformMetaLinks [
+	| globalNodes |
+	globalNodes := self globalNodes: referencingClass >> #references.
+	self deny: globalNodes first hasDeprecationTransformMetaLink.
+	self deny: globalNodes second hasDeprecationTransformMetaLink.
+	self deny: globalNodes third hasDeprecationTransformMetaLink.
+	deprecatedClass installDeprecationTransformMetaLinks.
+	globalNodes := self globalNodes: referencingClass >> #references.
+	self assert: globalNodes first hasDeprecationTransformMetaLink.
+	self assert: globalNodes second hasDeprecationTransformMetaLink.
+	self deny: globalNodes third hasDeprecationTransformMetaLink
+]
+
+{ #category : #tests }
+ClassDeprecationClassTest >> testIsDeprecatedTransformable [
+	self deny: referencingClass isDeprecatedTransformable.
+	self assert: deprecatedClass isDeprecatedTransformable.
+	deprecatedClass class
+		compile:
+			'isDeprecated
+	<transformTo: ''' , replacementClass name
+				,
+					'''>
+	^ false'.
+	self deny: deprecatedClass isDeprecatedTransformable.
+	deprecatedClass class
+		compile:
+			'isDeprecated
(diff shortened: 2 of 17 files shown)


Ducasse:
I was thinking that we should keep that for Pharo 90.
Any concerns?

MarcusDenker:
I would add it, the code is not called if it is not used.

maenu:
It also has a system setting, which is by default disabled. So it will not do anything unless enabled by the user.

astares:
Yes - I would add it too

dionisiydk:
Hi @maenu.

Very impressive work. 

Did you see existing support for deprecating classes? There are Class>>#deprecationRefactorings and Class>>applyDeprecation. And there is a browser command "Apply deprecation" for deprecated classes.
They serve different purpose to transform everything in system. But maybe something can be shared with your work.

When we implemented it with @pavel-krivanek we also thought about automatic transformations in future but with different approach. Idea was to introduce first class DeprecatedBinding which will implement value reading in exactly same way how current method deprecation is done. Basically it would reuse full logic of method transformation. I think it is much simpler than your approach but we did not try to implement it.

maenu:
@dionisiydk Yep, I saw the refactorings. This PR only addresses the migration of class references, which I think is part of the "external" complement to the "internal" refactoring. Maybe there should be a refactoring "Replace with new and deprecate" which would create a copy of the class, set the superclass of old class to the new class, moves the old class to a deprecated package, and adds the `#isDeprecated` with the transformation to the old class.

I also played with using a `DeprecatedBinding` to reuse exactly the same machinery as for method deprecation with a compiler plugin. The issue is though, that the mappings from bytecode to AST nodes is broken that way. I.e. `sourceNodeForPC:` will not work during execution, as the AST it tries to map to is rebuilt using the parser only, which does not include the binding yet. With meta-links we can be sure that the correct AST is kept around, otherwise we would need to ensure to keep the special binding AST around, or change the way bytecode is mapped to AST nodes, or support parser plugins. All of the alternatives I explored were not as clean and pluggable as this meta-link approach.

dionisiydk:
@maenu I see. I expected such kind of problems with this approach (DeprecatedBinding).
It would be nice to report separate issues to cover them if you can recover it from your experiments.
Normally transforming AST on fly from the variable access should work.

dionisiydk:
Hi @maenu.

> With meta-links we can be sure that the correct AST is kept around, otherwise we would need to ensure to keep the special binding AST around, or change the way bytecode is mapped to AST nodes, or support parser plugins. All of the alternatives I explored were not as clean and pluggable as this meta-link approach.

Your comment encouraged me to try my idea because I do not see why metalinks should give more robust solution. It could be only because of bugs. But in my experiment I did not see any issue.

So please, look at my PR #5262. It is only proof of concept but it shows the simplicity of this approach. I am interesting to see what issues I do not notice

maenu:
@dionisiydk Set a breakpoint in `Deprecation>>#transform` and check the source node it matched from the PC. It is some parent of the class reference, not the reference itself. This messes with the debugger and syntax highlighting. The class reference node also does not have the deprecation binding. The replacing works in a way, but it is imprecise.

dionisiydk:
I will check but how it affects the code transformation during deprecation.

As I said the transformation works correctly with my code.

пт, 29 нояб. 2019 г., 8:03 maenu <notifications@github.com>:

maenu:
Yes, the transformation works. But it starts the matcher not on the global node, but on some parent. This differs from the method deprecation, which transform on the actual message send node only.

dionisiydk:
> Yes, the transformation works. But it starts the matcher not on the global node, but on some parent. This differs from the method deprecation, which transform on the actual message send node only.

Following assertion is true in my PR:
```Smalltalk
>>readDeprecatedValue
   self assert: thisContext sender sourceNodeExecuted class = RBGlobalNode.
```
So I do not see that something is broken. Do you have concrete code for deprecation user which will show the problem?

dionisiydk:
Notice I am using Pharo 8.

maenu:
Check the binding of the node, it is not the deprecated one. I managed to trick it with more nested structures, e.g. using the class  in an argument or in a return statement. I also used Pharo 8 for my tests.

dionisiydk:
Hm. In my tests I can't get the wrong state:
```Smalltalk
readDeprecatedValue
| node |
node := thisContext sender sourceNodeExecuted.
self assert: node class = RBGlobalNode.
self assert: node binding assoc class = DeprecatedGlobalVariable.
```
All these assertions pass in my tests with various expressions where deprecated class was a receiver, an argument or a return value.

maenu:
Now I understand better what happens in your PR. I did not read your PR correctly and somehow assumed you make changes on the AST level, which you do not, instead you change the binding. Changing the binding is indeed quite elegant, but I think there are also some corner cases to cover. E.g. when I load you code and run
```
DeprecatedClassExample binding primitiveChangeClassTo: DeprecatedGlobalVariable new.
nil deprecationUserExample.
```
no transformation happens. After recompiling `#deprecationUserExample` it works though, not sure why this was necessary. The question is also who changes the global bindings other than the deprecation, and can there be situations where you would want the old binding?

I cannot really tell which approach is more robust and predictable. I favor the meta-link + system announcement approach because I understand it better, it is tested, and it is pretty light-weight as it by itself does not make deep changes in the system, instead it plugs into a few entry points that already exist, i.e. meta-links and announcements. But I am no neutral judge in this situation, so probably somebody else should decide on which way to go. One further step could be to test the same scenarios I do in this PR with the binding variant in #5262.

dionisiydk:
> no transformation happens. After recompiling `#deprecationUserExample` it works though, not sure why this was necessary. 

I just implemented incorrect deprecation method:
```Smalltalk
readDeprecatedValue
	self
		deprecated: 'Please use Object instead'
		transformWith: 'DeprecatedClassExample' -> 'Object'.
	^self
```
Normally the code after deprecation should implement new logic. So it should be "^Object" at the end for this example.

dionisiydk:
> The question is also who changes the global bindings other than the deprecation, and can there be situations where you would want the old binding?

It can be similar "tracking code" which you did for metalinks. It will alter the type of binding instead of installing/deinstalling deprecation metalinks. All problems you solved here (in your PR) are relevant to my approach.

> I cannot really tell which approach is more robust and predictable. 

It's only the question of current bugs in these parts of system. Normally they should be equivalent.
But currently I think we have some issues with debugging methods with active metalinks (stepping over/into nodes with "invisible metalinks logic", selecting current execution node)  

>I favor the meta-link + system announcement approach because I understand it better, it is tested, and it is pretty light-weight as it by itself does not make deep changes in the system, instead it plugs into a few entry points that already exist, i.e. meta-links and announcements. But I am no neutral judge in this situation, so probably somebody else should decide on which way to go. One further step could be to test the same scenarios I do in this PR with the binding variant in #5262.

Yes, here is where your PR wins :). I don't know if I will finish my demo PR.

MarcusDenker (approved):

PR #5224, opened 2019-11-25
5212-Executable-comment-checker-problem

### LINKED ISSUE #5212: Executable comment checker problem
I do not get why Bag >> includes: executable is reported as an error.
Same for 
Screenshot 2019-11-24 at 09 46 26

Ducasse:
Fixes: #5212 Some executable comments... .more to come.

### DIFF
--- a/src/Collections-Strings/String.class.st
+++ b/src/Collections-Strings/String.class.st
@@ -481,8 +481,8 @@ String >> alike: aString [
 
 { #category : #'finding/searching' }
 String >> allRangesOfSubstring: aSubstring [
-	"'Ab cd ef Ab cd' allRangesOfSubstring: 'cd' >>> {(4 to: 5). (13 to: 14)}"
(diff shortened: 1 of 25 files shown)


hogoww:
I worked on the drtest plugin this week end for that, and I wonder something. (will continue friday)
Do you think we should declare this comment wrongly written?
1 between:2 and:3 >>> false.
Because in this we send the message #>>> to 3 instead of the result.
(I did a checker that tries to '''''parse it''''' when it first fail . This works for half of the current drTest failures , so about 50.)
So either we authorise it, with the problems that come with, or we rewrite them if they are wrong.

Ducasse:
Hi pierre
I fixed everything. We should use plain syntax.
I did not know you were working on it.

Ducasse:
I'm now working on #5225 to avoid errors.

hogoww:
Oh, you fixed the drtest pluggin as well?

Ducasse:
Borken tests are related to blueInk removal and I will try to work on Enlumineur introduction this evening.

PR #5229, opened 2019-11-26
Completion: EntryCleanup

MarcusDenker:
- NECSpecialVariableEntry instead of special classes for self and super (not yet used)
- use subclasses NECEntry 
- some more cleanups


MarcusDenker:
failing tests seem not related

Ducasse:
Marcus can you consider to avoid using Smalltalk everywhere. This is a global variable. 
This is a plague. A real one.

MarcusDenker:
This is can be done in a second PR (I am doing this very incremental to not disturb people as the code is executed when coding)

PR #5244, opened 2019-11-27
[shortToReview] remove call to basicNew in Object >> #->

hogoww:
If that is, in fact, normal, it should have a comment saying why. 
(Like if it's used in the boostrap? Although I feel this is probably too high level for that)

There's other references to basicNew in Object, but I don't feel as comfortable removing them :p

### DIFF
--- a/src/Kernel/Object.class.st
+++ b/src/Kernel/Object.class.st
@@ -248,7 +248,7 @@ Object >> -> anObject [
 	
 	"Note that key and value can be ANY object."
 	
-	^Association basicNew key: self value: anObject
+	^Association key: self value: anObject
 ]
 
 { #category : #comparing }


MarcusDenker (approved):
yes, good cleanup

hogoww:
> you remove the basicNew message send. now you need to remove the comment
> "the code of Association>>#key:value: is inline here for speed"

Sorry, I thought the whole code was available her, which it is not..
I have no idea how I was able to miss that comment ... >_<

@MarcusDenker @guillep  What do you, think? remove the comment or revert this?
@guillep I don't think I've done those bench right.

I'd tend to revert it if it's used in the kernel much, given that the performance varie quite a lot.
[ Object new ->  Object new ] bench. 
 " without inlined code '24977448.110 per second'"
 "'18418246.452 per second'"
 "'18783934.613 per second'"
 "'19108582.850 per second'"
[ Object new ->  Object new ] bench. 
" with inlined code     29069318.473 per second'"
 "'31253506.696 per second'"
 "'31563000.600 per second'"

(first run for each was done a while before the following, maybe the memory was crowded after that.
after fresh GC run, performances seems similar.
[ Object new ->  Object new ] bench." without"
 "'30353560.688 per second'"

[ Object new ->  Object new ] bench. 
 "'31752747.551 per second'" "with"

But if it's really for performance, I'd keep it that way for convenience, and use this version in the kernel / base collections, (or an special method inlined by the compiler...):
[ Association key: Object new value: Object new  ] bench.
 "'31132840.032 per second'"
"'32016940.236 per second'"
"'36153477.809 per second'"
"'32302426.259 per second'"


[ Association basicNew key: Object new value: Object new  ] bench. 
"'33608275.490 per second'"
"'32630870.078 per second'"
"'32510358.457 per second'"

guillep:
I think the performance should not be a problem. 18M associations per second are too many associations. Those needing performance could use `Association key: Object new value: Object new` directly.

guillep:
What happened with the comment?

hogoww:
I did not remove the comment : "the code of Association>>#key:value: is inline here for speed" from the method.
Therefore you currently have a comment that matches nothing in the Object >> #-> method !

Issue #5249, opened 2019-11-27
Spotter is broken 


astares:
Open Spotter in recent image gives a DNU since the brick iVar is nil

![image](https://user-images.githubusercontent.com/5980033/69704113-23af2680-10f3-11ea-84b6-867ba3f934e6.png)

MarcusDenker:
fixed by merging https://github.com/pharo-project/pharo/pull/5251

astares:
@MarcusDenker and @VincentBlondeau: Thanks!

PR #5251, opened 2019-11-27
Fix spotter openning bug

VincentBlondeau:
As the settings button is added only when #GTEventRecorderSettings is there. Since it has been deleted, the same check should be done. 

Also cleaning the baseline.


astares (approved):
Baseline change not related to Spotter - but is a cleanup.

The change works.

astares:
Fixes #5249

MarcusDenker (approved):

Ducasse:
Thanks Vincent. Another dependency. I'm amazed how badly modular is our system.

PR #5254, opened 2019-11-27
5246--AST-Core-Tests-package-disappeared

### LINKED ISSUE #5246: AST-Core-Tests package disappeared
We lost all tests for the AST.

pavel-krivanek:
revert loading of AST-Core-Tests

fixes #5246


MarcusDenker:
31 tests fail

jecisc:
This is not that simple:

```Smalltalk
{ #category : #accessing }
RBParserTest >> exampleClasses [
	<sampleInstance>
	^ Array
		with: self parserClass
		with: RBScanner
		with: RBProgramNode
		with: BIConfigurableFormatter
]
```

Ducasse:
I really hate when people do not respect dependencies. 
You see this method is terrible because it creates a dependencies from the AST core to its users. When will we start to consider that dependencies are important if we want to have a modular system. I did not even thought that they was a dependency and of course no baseline was reflecting this.

Ducasse:
Thanks Pavel I owe you some beers :)

PR #5255, opened 2019-11-27
In bootstrap: Preparing heap before writing the image to shrink its size

carolahp:
Fixes the problem in which all bootstrapped images are generated weighting 30 MB aprox.


pavel-krivanek (approved):
The bootstrapped image size is 575 KiB

jecisc:
It would be great to add a comment explaining why it is there :)

PR #5257, opened 2019-11-28
Deprecate-removed-projects

jecisc:
Deprecate BlueInk and GTEnventRecorder.

### DIFF
--- a/src/BaselineOfBlueInk/BaselineOfBlueInk.class.st
+++ b/src/BaselineOfBlueInk/BaselineOfBlueInk.class.st
@@ -1,0 +1,21 @@
+Class {
+	#name : #BaselineOfBlueInk,
+	#superclass : #BaselineOf,
+	#category : #BaselineOfBlueInk
+}
+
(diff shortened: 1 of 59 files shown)


Ducasse:
So you should also fix the dependencies.
And we will have to remove again the dependencies in P9.
And you can remove also the tests of BlueInk.

jecisc:
Hi,

The two projects are loaded last in the image and it builds well. There is no dependence on the image to those two projects.

All the dependency tests are green.

For the removal, everything is in a deprecated section in the BaselineOfPharo. It is as simple as to remove this section since it is loaded last and in general, I take care of it:
- https://github.com/pharo-project/pharo/pull/5006
- https://github.com/pharo-project/pharo/pull/2760
- https://github.com/pharo-project/pharo/pull/2307

VincentBlondeau (approved):
Built image with the deprecated packages. The load of the PR is not possible on an old image though. 

Thanks Cyril!

estebanlm:
this is already handled

Issue #5260, opened 2019-11-28
[CI]Why the tests are not run on Pharo x64?


VincentBlondeau:
Is there any reason the tests are not run on the x64 version of Pharo?
Since both the x64 and x32 are released why they should be skipped?

MarcusDenker:
because nobody did it.

MarcusDenker:
it's really that simple: everything obvious that should be done is not done because nobody did it

VincentBlondeau:
I was expecting a more complex answer... I'll do a PR then!

MarcusDenker:
was done

PR #5261, opened 2019-11-28
Fixes-5260 - Add x64 tests run in the CI

VincentBlondeau:
See #5260 Add x64 tests run

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -126,7 +126,7 @@ Build Url: ${env.BUILD_URL}
 def bootstrapImage(){
    cleanWs()
   def builders = [:]
-  def architectures = ['32']//, '64']
+  def architectures = ['32' , '64']
   for (arch in architectures) {
       // Need to bind the label variable before the closure - can't do 'for (label in labels)'
       def architecture = arch


svenvc:
Sometimes even the simplest things might work, let's try.

VincentBlondeau:
It is not working since I do not have right accesses on the repo. Someone of the project should do it. Or integrate with eyes closed ;)

See https://support.cloudbees.com/hc/en-us/articles/227246367-Why-Jenkinsfile-changes-Are-Not-Reflected-in-PR-Build

<blockquote><div>CloudBees Support</div><div><strong><a href="http://support.cloudbees.com/hc/en-us/articles/227246367-Why-Jenkinsfile-changes-Are-Not-Reflected-in-PR-Build">Why Jenkinsfile changes Are Not Reflected in PR Build</a></strong></div><div>Issue


When a user modify the Jenkinsfile and opens a PR, the build does not take the changes into account


Console logs shows one of the following depending on the GitHub Branch Source plugin ve...</div></blockquote>

guillep:
I'll try to give you access :)

guillep:
Bof, I've tried to change jenkins settings, but apparently it cannot be done that way.
I've further seen https://support.cloudbees.com/hc/en-us/articles/227246367-Why-Jenkinsfile-changes-Are-Not-Reflected-in-PR-Build. So this can only be added by giving push requests on the repo...

Two things:
 - I'll try to take ownership of this PR if this does not bother you, like that the JenkinsFile changes are taken into account...
 - What if we only run tests for 64 bits? It should be the default platform for most people (outside still existing 32bit arm machines). Because otherwise this will multiply the test-running by 2 demanding twice as slaves/resources, making the build slower and less reliable...

What do you think?
I invoke also @estebanlm @tesonep so they can give their opinion.

<blockquote><div>CloudBees Support</div><div><strong><a href="http://support.cloudbees.com/hc/en-us/articles/227246367-Why-Jenkinsfile-changes-Are-Not-Reflected-in-PR-Build">Why Jenkinsfile changes Are Not Reflected in PR Build</a></strong></div><div>Issue


When a user modify the Jenkinsfile and opens a PR, the build does not take the changes into account


Console logs shows one of the following depending on the GitHub Branch Source plugin ve...</div></blockquote>

VincentBlondeau:
You can take ownership of PR. I just care about running the x64 tests. And I thank that running the x32 ones could be nice to compare if the x64 fails.

guillep:
Work continued in #5268 . Let's continue the discussion there.

PR #5262, opened 2019-11-28
[Do not integrate] DeprecatedGlobalVariable experiment

dionisiydk:
It is a proof of idea to represent deprecated classes with special first class variable: DeprecatedGlobalVariable. 
It compiles the variable read as a message send #readDeprecatedValue to the class. So existing method deprecation logic can be reused here:
- by default it calls simple #deprecated: message which shows the warning
- concrete classes can transform senders using rewrite expression

The PR does not implement the hook to use new variable instead of existing GlobalVariable binding.
So for testing load PR and evaluate following code:
```Smalltalk
DeprecatedClassExample binding primitiveChangeClassTo: DeprecatedGlobalVariable new.
```
PR includes example of deprecated class user. To see transformation evaluate:
```Smalltalk
nil deprecationUserExample.
```
Try also remove the method DeprecatedClassExample class>>#readDeprecatedValue to see the simple warning.

For real usage the following logic needs to be implemented:
- compiler plugin to alter the class binding when #isDeprecated method is modified(created/removed). 
- support full package deprecation where bindings of all classes becomes deprecated. Package deprecation is also based on #isDeprecated method (for Manifest class). So the compiler plugin can be responsible for this logic also.
- improve existing Deprecated class to take into account the class deprecation scenario. Now the warning tells that method #readDeprecatedValue is deprecated but it should talk about class. Probably it would require two separate subclasses: MethodDeprecated and ClassDeprecated

### DIFF
--- a/src/Announcements-Core/DeprecatedClassExample.class.st
+++ b/src/Announcements-Core/DeprecatedClassExample.class.st
@@ -1,0 +1,20 @@
+Class {
+	#name : #DeprecatedClassExample,
+	#superclass : #Object,
+	#category : #'Announcements-Core'
+}
+
+{ #category : #testing }
+DeprecatedClassExample class >> isDeprecated [ 
+	^true
+]
+
+{ #category : #deprecation }
+DeprecatedClassExample class >> readDeprecatedValue [
+
+	self
+		deprecated: 'Please use #crTrace678 instead'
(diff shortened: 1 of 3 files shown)


estebanlm:
please reopen again if still relevant

PR #5264, opened 2019-11-29
[FeedbackNeeded] test color report (drTest/Calypso)

hogoww:
Started to change colors used by DrTest and Calypso, to be less flashy.
Not ready, want to change some more stuff with Julien a bit later today.

commit number 3 is lacking a part of its message:  'and tweeked color to be less flashy'
overs are explicit


hogoww:
SimpleButtonMorph >> doButtonAction was as far as I went to find Calypso's colors. 
But I feel they are fine in comparison, being so small and without text, so i don't know if i'll investigate further.

src/Calypso-SystemPlugins-SUnit-Browser/ClyTestCommand.class.st
```diff
@@ -31,14 +31,15 @@ ClyTestCommand >> defaultMenuIcon [
...
+	theme := Smalltalk ui theme.
 	testResult hasPassed
-		ifTrue: [ color := Color green ].
+		ifTrue: [ color := theme successBackgroundColor ].
 	testResult hasFailures
-		ifTrue: [ color := Color yellow ].
+		ifTrue: [ color := theme warningBackgroundColor ].
 	testResult hasErrors
-		ifTrue: [ color := Color red ].
+		ifTrue: [ color := theme dangerBackgroundColor].
```
dionisiydk: I suggest to follow SUnit vocabulary: greenTestColor, yellowTestColor and redTestColor. This is well established naming.

dionisiydk (changes requested):

src/Calypso-SystemPlugins-SUnit-Browser/ClyTestCommand.class.st
hogoww: I like this. We wouldn't impact other users.
But should the other users keep those colors? 
It's a real question I'm wondering, I don't know about everything that's affected, but if it's as unreadable as miniDrTest, it might make sense to still change them.

hogoww:
Ok, so now we have default color described in Sunit. gonna try to find "better" colors for tests.

@dionisiydk Didn't find the color for unit tests in Calypso yet. Could you give me an hint as where I should look please?

hogoww:
Attempted to have better coloring for DrTest/Calypso test. Would love some feedback to be able to know what would be nicer.

Important note: Spec2 doesn't support coloring for buttons label's, so we're stuck with white text in dark mode for now.

src/SUnit-Core/TestResult.class.st
```diff
@@ -17,6 +17,41 @@ Class {
 	#category : #'SUnit-Core-Kernel'
 }
 
+{ #category : #'as yet unclassified' }
+TestResult class >> defaultColorBackGroundForErrorTest [
+	^ Smalltalk ui theme testColorError
```
juliendelplanque: Maybe extract `Smalltalk ui` in a separated method to avoid duplicating it everywhere?

src/SUnit-Core/TestResult.class.st
```diff
@@ -17,6 +17,41 @@ Class {
...
+]
+
+{ #category : #'as yet unclassified' }
+TestResult class >> defaultColorBackGroundForFailureTest [
+	^ Smalltalk ui theme testColorFailure
```
juliendelplanque: Same here

src/SUnit-Core/TestResult.class.st
```diff
@@ -17,6 +17,41 @@ Class {
...
+]
+
+{ #category : #'as yet unclassified' }
+TestResult class >> defaultColorBackGroundForPassingTest [
+	^ Smalltalk ui theme testColorPassing
```
juliendelplanque: Same here.

juliendelplanque (changes requested):
Seems nice, little detail to discuss, see related comments.

estebanlm:
1) I need a screenshot to evaluate the before/after of the changes
2) I like to remove hardcoded colours
3) This will be for Pharo9.0 now :)

hogoww:
For the dark theme:
Personnal notes first:
- I cannot read anything on the current success/failure, error is okay.
- My screen is probably too bright, but on it some colors are just blinding. Got to hide my drtest under another window 1
- We added support for default font color, but Spec does not support it yet (soon !). Still, this prevents us to do better, by replacing the font color and have a nice color, with readable text. 
(also, my color vision is not the best..)

success new
success_dark_new
success old
dark_success_old


failure new
Screenshot 2019-12-25 at 17 21 12

failure old
dark_failure_old


error new
error_dark_old
error old
dark_error_old

hogoww:
light theme
Only failure was unusable. 
I'm not sure why text is white and don't remember if I once knew. 
This is making it worse though.
iirc the text color follows the theme (white text dark theme / black text white theme...)

Success new
light_success_new
success old 
light_success_old

failure new
light_failure_new
failure old
light_failure_old

error new
error_failure_new
error old
light_error_old

Ducasse:
I usually cannot read comments in the black theme and the red is also far too strong I cannot read text in it.

Ducasse:
So take care of it is really important.

hogoww:
@Ducasse "I usually cannot read comments in the black" Alright, but this is unrelated to this PR, unless I misunderstood. 
"the red is also far too strong I cannot read text in it." Do you mean the Dark theme new red ?

juliendelplanque (approved):

Ducasse:
Julien we have a conflict.

hogoww:
Should be fine now

hogoww:
Failure seems unrelated :)

MarcusDenker (approved):

PR #5267, opened 2019-11-29
3437-Add-test-for-Adding-an-instance-class-variable-remove-the-traits-of-a-class

### LINKED ISSUE #3437: Add test for: Adding an instance class variable remove the traits of a class

MarcusDenker:
- add test for adding a class side slot (class instance variable) and check that it gets added. 
- Check especially that the class does not lose the trait.
- add another test just testing adding slots, not class side slots
- autimattic deprecation rewrite

fixes #3437

### DIFF
--- a/src/TraitsV2-Tests/T2TraitWithComplexSlotsTest.class.st
+++ b/src/TraitsV2-Tests/T2TraitWithComplexSlotsTest.class.st
@@ -66,6 +66,31 @@ T2TraitWithComplexSlotsTest >> testTraitWithComplexSlot [
 ]
 
 { #category : #'instance creation' }
+T2TraitWithComplexSlotsTest >> testTraitWithComplexSlotAddClassSlot [
+	| t1 c1 |
+	self createSlotClass.
+	t1 := self newTrait: #T1 with: { #ivar }.
+	c1 := self newClass: #C1 with: { #aSlot } uses: { t1 }.
+	self assert: (c1 traitComposition includesTrait: t1).
+	c1 addClassSlot:  #ivar2 => InstanceVariableSlot.
+	self assert: c1 instVarNames equals: #( aSlot ivar ).
+	self assert: c1 class instVarNames equals: #( ivar2 ).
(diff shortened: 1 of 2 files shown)


src/TraitsV2-Tests/T2TraitWithComplexSlotsTest.class.st
```diff
@@ -65,6 +65,31 @@ T2TraitWithComplexSlotsTest >> testTraitWithComplexSlot [
...
+	| t1 c1 |
+	self createSlotClass.
+	t1 := self newTrait: #T1 with: { #ivar }.
+	c1 := self newClass: #C1 with: { #aSlot } uses: { t1 }.
+	self assert: (c1 traitComposition includesTrait: t1).
+	c1 addSlot: #ivar2 => InstanceVariableSlot.
+	self assert: c1 instVarNames equals: #( aSlot ivar2 ivar ).
+	self assert: (c1 traitComposition includesTrait: t1)
+]
+
```
guillep, line 92: Why are these tests repeating so many things? It kind of hides the purpose of the test (it would be good to name the test after its purpose too...)

guillep (changes requested):
I think the tests should better show the intention following the discussion in #3437
"adding a instance/class slot does not remove traits"

src/TraitsV2-Tests/T2TraitWithComplexSlotsTest.class.st
MarcusDenker: Thinking about it again: 
1) both tests test different things: one adds a slot, one adds a class slot. This is clear from the name of the tests
2) I test for both tests the same as you have to test the same: the added slot has to be there, the existing slots have to be still there *and* the trait needs to be there.

So the tests test adding a slot / adding a class slot in a class that has a Trait that itself has a slot and checks that everything is ok.

I prefer that to adding a test that is explicitly testing for one bug.

PR #5268, opened 2019-11-29
Add 64 bits to the list of architectures for test runs

guillep:
Redo of #5261 , fixing #5260

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -223,7 +223,17 @@ try{
 
     //Testing step
     def testers = [:]
-    def architectures = ['32', '64']
+    // We run the whole process in 64 bits all the time. 
+    // The 32 bits process is only run when a PR is integrated
+  
+    def architectures
+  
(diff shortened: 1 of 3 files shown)


VincentBlondeau:
It seems that the fix is a bit more complex. The x64 files needs to be stashed correctly too.

VincentBlondeau:
L 129 should be edited too. @guillep  I let you do it :)

guillep:
Yes sir, let's see how this goes. Before bootstrapping in 64 bits used to crash the VM.

VincentBlondeau:
Ok. The configuration change is working!
And, there are issues on the x64 version :)

VincentBlondeau (approved):
Change is working but tests are failing. Test should be solved in another PR

estebanlm:
Moving to P9 because no time for the release (P9 is starting now anyway so you can continue working)

PR #5283, opened 2019-12-02
4454-TraitedMetaClassisRejectedMethod-rejects-too-many-methods

### LINKED ISSUE #4454: TraitedMetaClass>>isRejectedMethod: rejects too many methods
I need to override two methods on the class side: `#ffiLibrary` and `ffiLibraryName`. I do this with a trait. When I compile the class I lose those two overrides due to the logic in TraitedMetaClass>>isRejectedMethod:.

MarcusDenker:
- change #isRejectedMethod: to never reject methods that are in the trait composition
- inline the comments into the code (easier to understand)

Traits tests are green

fixes #4454 (I hope)


VincentBlondeau (changes requested):
Could we add a test in a subclass of T2AbstractTest so we are sure that the bug we will raise again?
Thanks!

MarcusDenker:
yes! we should really add a test for this

VincentBlondeau (approved):
Thanks Marcus!

MarcusDenker:
all tests green

PR #5286, opened 2019-12-02
OrderedClassPool

MarcusDenker:
The class pool is not ordered. This leads to the sad state of affairs that we ordere the #classVarNames when we access them...

This is an experiment to see what breakes if we just make them OrderedDisctionaries. The next step then is to remove the #sort from #classVarNames...


MarcusDenker:
method missing: 
Error: declareVariable:from: not found in OrderedDictionary

MarcusDenker:
need to check at the bootstrap love...

PR #5288, opened 2019-12-02
5222-Metacello-does-not-produce-loading-log-anymore

### LINKED ISSUE #5222: Metacello does not produce loading log anymore
In my Pharo 8's CIs, I do not have the loading logs anymore.

I have the logs about packages been fetch but not about the packages been loaded, the undeclared variables, the shadow variables...

pavel-krivanek:
replace Metacello logging to Transcript with a specialized SystemNotification

fixes #5222


svenvc:
Structured notifications would be better still ;-)

Will these still be logged on the Transcript ?

pavel-krivanek:
yes, the system notifications are by default printed to Transcript.

estebanlm (approved):

PR #5289, opened 2019-12-02
Adding halt to the list of captured exceptions when single stepping

### LINKED ISSUE #5280: stepping on halt is breaking debugging experience
In P7 stepping on halt was just going over it and now it blocks completely and open another debugger and break everything. 
This is a serious regression in terms of debugging experience.

StevenCostiou:
- Halt was missing from the catched exceptions, was it removed on pur…pose? (present in P5, removed in P6)

- UnhandledError cannot be catched, as it is not a subclass of Error
#fixes #5280

### DIFF
--- a/src/Kernel/Context.class.st
+++ b/src/Kernel/Context.class.st
@@ -1674,7 +1674,6 @@ Context >> runUntilErrorOrReturnFrom: aSender [
 	context := aSender insertSender: (Context
 		contextOn: Error, Halt do: [:ex |
 			error ifNil: [
-				"this is ugly but it fixes the side-effects of not sending an Unhandled error on Halt"
 				error := ex.
 				topContext := thisContext.
 				ex resumeUnchecked: here jump ]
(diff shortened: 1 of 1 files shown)


src/Kernel/Context.class.st
```diff
@@ -1672,10 +1672,10 @@ Context >> runUntilErrorOrReturnFrom: aSender [
 	"Insert ensure and exception handler contexts under aSender"
 	error := nil.
 	context := aSender insertSender: (Context
-		contextOn: Error do: [:ex |
+		contextOn: Error, Halt do: [:ex |
 			error ifNil: [
 				"this is ugly but it fixes the side-effects of not sending an Unhandled error on Halt"
```
estebanlm: if you remove the `isKindOf:`, this comment still has sense? it applies?

tesonep (approved):

src/Kernel/Context.class.st
StevenCostiou: Hmmm indeed, that comment could be removed.

MarcusDenker (approved):

Issue #5291, opened 2019-12-02
SourceFileBufferedReadWriteStream does not handle well #refreshBufferFrom:


VincentBlondeau:
When writing in a deferFlushDuring mode (to compile methods), when the buffer is full #refreshBufferFrom: will write the contents of it even if the chunk is not complete. After a #refreshBufferFrom: the source file can be in an instable state and then cannot be read by epicea.

I suggest to flush only complete strings (the argument of the nextPutAll: for instance) and not flushing in the middle of it.

VincentBlondeau:
But it will not work if the string to write is larger than the buffer... I then suggest to disable the gathering of the timestamp from Epicea in the change log when in deferFlushDuring: block.

VincentBlondeau:
Weak linked with #5292

tesonep:
@VincentBlondeau can you check if the fix I have proposed for #5292 keeps the same problem? It should be fixed by the same code.

MarcusDenker:
merged

PR #5292, opened 2019-12-02
[Perfs] x10 on method compiling with #deferFlushDuring:

VincentBlondeau:
Changes:
- Avoid flushing at the end of each chunk.
- Tools should not read the logfile when the deferFlushing is active because it can read incomplete chunks.

Bench:
Before:

> SourceFiles deferFlushDuring: [  [EphemeronTest compile: 'test ^true' ] bench]  "'82.668 per second'"

After:
>SourceFiles deferFlushDuring: [  [EphemeronTest compile: 'test ^true' ] bench]  "'1066.573 per second'"

### DIFF
--- a/src/System-Sources/CompiledMethod.extension.st
+++ b/src/System-Sources/CompiledMethod.extension.st
@@ -20,10 +20,9 @@ CompiledMethod >> putSource: source withPreamble: preambleBlock [
 	SourceFiles 
 		writeSource: source 
 		preamble: (String streamContents: preambleBlock)
-		onSuccess: [ :newSourcePointer | 
+		onSuccess: [ :newSourcePointer :theChangesFileStream |
 			"Method chunk needs a final ! !"
-			SourceFiles changesWriteStreamDo: [ :file | 
-				(ChunkWriteStream on: file) nextPut: ' ' ].
+			(SourceChunkWriteStream on: theChangesFileStream) nextPut: ' '.
 			"Update with new source pointer"
 			self setSourcePointer: newSourcePointer ]
 		onFail: [ 
--- a/src/System-Sources/SourceChunkWriteStream.class.st
+++ b/src/System-Sources/SourceChunkWriteStream.class.st
@@ -1,0 +1,33 @@
+"
+I am a special ChunkWriteStream used to write on the streams without the flushing part.
+If the flushing part can be changed in my super class then my existence must be terminated.
+"
+Class {
+	#name : #SourceChunkWriteStream,
+	#superclass : #ChunkWriteStream,
+	#category : #'System-Sources-Utilities'
+}
+
+{ #category : #accessing }
+SourceChunkWriteStream >> nextPut: aString [
+	"Append the argument, aString, to the receiver, doubling embedded ! terminators and adding a extra one"
+
+	| string start bangIndex |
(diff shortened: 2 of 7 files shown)


tesonep:
Hi Vincent, it looks like the generated source file in the bootstrap does not include the sources.

svenvc:
I saw this excessive flushing before as well, it is obviously too much when writing lots of bits, but it is probably needed when writing only one.

Of course this is again one of those 'let's not touch it because it might be needed', so let's try.

VincentBlondeau:
@tesonep it seems to be more an issue with the statefull traits and Spec

tesonep:
The problem is that during the installation of a stateful trait the source code of the methods are used (if it requires to recompile, in this case, it is using an slot) so there is a problem with the generation of the changes file

VincentBlondeau:
That's a weird behavior... So either, this mechanism should be changed or the file flushed before compiling the trait. I'll prefer the first but the second sounds easier

tesonep:
Hi Vincent, I would check the nature of the problem reproducing the bootstrap step by step. I like to integrate your improvement. 😄

VincentBlondeau:
So that the code to reproduce the bug:
```
compiler := Smalltalk globals at: #FBDDecompiler.
[
	Smalltalk globals removeKey: #FBDDecompiler.
	SourceFiles deferFlushDuring: [ 
	CmTWithBasicNameAndDescription compile: 'test2 ^42'.
	
	CmAbstractCommand subclass: #CmCommand
	uses: CmTWithBasicNameAndDescription + CmTDecorable
	instanceVariableNames: 'context'
	classVariableNames: ''
	package: 'Commander2-Commands'  ].
]ensure: [ Smalltalk globals at: #FBDDecompiler put: compiler ].
CmCommand new test2
```

In the bootstrap, there is no decompiler so the sources cannot be retrieved from either the source code or the decompiler (through OpalCompiler>>decompileMethod:).

TaAbstratComposition>> #sourceCodeAt: should be implemented through another way to get the data from the other already existing method.

tesonep:
Hi @VincentBlondeau, I have checked the problem and the problem is that the source code of a method is not accessible until the stream is flushed.
 
If you check SourceFileArray >>#readStreamAtFileIndex: index atPosition: position ifPresent: presentBlock ifAbsent: absentBlock the code is read from the file directly (using a set of read-only streams directly). I can change this method to read using the w/r stream if they are dirty, although I don't know the impact in speed that will have (as the read-only streams allow Pharo to share the access to the underlying file). 

I will try this modification.

Also, I have a doubt, to me, it is not important but maybe someone cares. As if there is an error in loading a metacello package the .changes file is not correctly saved.

VincentBlondeau:
Or you can force the flush of the stream since we not using the statefull traits often. However, I still think the compilation of the traits should not be based on the change files to retrieve the source code in case we have an image in read only mode.

tesonep:
Yes, the problem it requires the source code (or a decompiled version) to recompile the usage of instance variables or if there is an alias or deep alias.

tesonep:
#5291 is fixed also with my modification

VincentBlondeau:
Good for me!

MarcusDenker:
is it good to merge?

Issue #5189, opened 2019-11-20
Finder should give the class of the exact result


VincentBlondeau:
Example: 'ababa' .  $a . $z . 'zbzbz' 
![image](https://user-images.githubusercontent.com/4706329/69257273-19ed6680-0bbb-11ea-8d79-85dfec1c70dd.png)

VincentBlondeau:
There is a star missing before the name of the class

kasperosterbye:
Odd. For me this do not even give a result, but fails
<img width="410" alt="image" src="https://user-images.githubusercontent.com/17723745/69974515-b0a81480-1525-11ea-9ee4-a31d40564625.png">

It fails because of the two-argument-keyword. 
Also this example fails: `'abcdef'. 2. 4. 'bcd'`.

If fails also in Pharo 7.

MarcusDenker:
merged PR

PR #5293, opened 2019-12-02
5290-Finder-example-search-break-with-two-argument-keyword-selectors

### LINKED ISSUE #5290: Finder example search break with two argument keyword selectors.
If I ask the finder to solve `'abcdef'. 2. 4. 'bcd'`, or `'ababa' . $a . $z . 'zbzbz'` (from issue #5189), it finds the method, but gives a primitive failure when trying to expand the result.
image

The problem lies in the:
```Smalltalk
FinderExampleMethodNode>>receiver
	| index rec |
	index := self item findString: self selector.
	rec := self item copyFrom: 1 to: index - 1.
	^ Smalltalk compiler evaluate: rec
```
The selector is not found in the first line because it is a two keyword selector.

kasperosterbye:
Fixes #5290 and #5189
Use the compiler instead of string manipulation - it is more robust

### DIFF
--- a/src/Tool-Finder/FinderExampleMethodNode.class.st
+++ b/src/Tool-Finder/FinderExampleMethodNode.class.st
@@ -17,10 +17,8 @@ FinderExampleMethodNode >> childNodeClassFromItem: anItem [
 
 { #category : #private }
 FinderExampleMethodNode >> receiver [
-	| index rec |
-	index := self item findString: self selector.
-	rec := self item copyFrom: 1 to: index - 1.
-	^ Smalltalk compiler evaluate: rec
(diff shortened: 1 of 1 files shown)


MarcusDenker (approved):
I do not like that it references the RBParser class directly, but for now it has to as the API of "Smalltalk compiler" is not easy to use in this case (and we do not have yet "Smalltalk parser").

So for now it is good, I will use this as the next TODO to improve the compiler API.

MarcusDenker:
I merge it and add the direct reference to RBParser to my backlog (to improve the API and fix it then)

PR #5294, opened 2019-12-02
5250-someClass-basicNew-printString-fails-for-around-10000-classes

### LINKED ISSUE #5250: someClass basicNew printString fails for around 10.000 classes
Is that a problem?

```Smalltalk
Object allSubclasses select: [ :c |  
    [(c basicNew) printString. false ] 
        on: Exception 
        do: [ true ] ].
```
The problem has an easy fix, changing the `#printStringLimitedTo:using:` method in 'Object' to have an exception handling part , perhaps returning 'an uninitialized nameOfClass` (or something better please).

Perhaps also some of the other implementations of `printString` need to be fixed, but not sure.

kasperosterbye:
Fixes #5250
Introduce a robustness to the general #printString. All non-failing printString calls continue to work. The fix is in particularly nice when debugging initialization code.

### DIFF
--- a/src/Kernel/Object.class.st
+++ b/src/Kernel/Object.class.st
@@ -1645,7 +1645,9 @@ Object >> printString [
 	"Answer a String whose characters are a description of the receiver. 
 	If you want to print without a character limit, use fullPrintString."
 
-	^ self printStringLimitedTo: 50000
+	[^ self printStringLimitedTo: 50000]
+	on: Exception
+	do: [ ^ 'an unPrintable/unInitialized ', self class name ]
 ]
 
 { #category : #printing }


svenvc:
You are essentially swallowing exceptions now.

There is some support in tools already.

Try inspecting 
    ZnMimeType new
The inspector says "error printing" and lower, "Error printing. Try self printString to debug".

I don't know if this is a good idea or not....

kasperosterbye:
True, I do swallow the exceptions. It was my experience that I could not run the debugger beyond a certain point because it got stuck on sending messages to nil (uninitialized instance variables) in classes I do not own. 

Yes, GT inspector do the same kind of error handling as I propose for the general case. 
If I try to follow the debugger into `OrderedCollection new`, I end up in `#setCollection:`, with self as 'error printing' as you say. If I then click on `self`, I get 'message not understood', as opposed to 'error printing'.

It is a simple thing, and I can live with just doing it in my image for my convenience. I cannot see any other way to do it.

kasperosterbye:
Closing this one as reviewer did not like it.

PR #5298, opened 2019-12-03
5282-MessageNotUnderstood-TextfindLastOccurrenceOfStringstartingAt-

### LINKED ISSUE #5282: MessageNotUnderstood: Text>>findLastOccurrenceOfString:startingAt:
How to reproduce

- Open Transcript
- In a method type
--  ()
-- then remove the last )
-- adding space or removing space produces 

MessageNotUnderstood: Text>>findLastOccurrenceOfString:startingAt:
MessageNotUnderstood: Text>>findLastOccurrenceOfString:startingAt:

MarcusDenker:
make sure the parser is handed a string, not a text. fixes #5282

### DIFF
--- a/src/Shout/SHRBTextStyler.class.st
+++ b/src/Shout/SHRBTextStyler.class.st
@@ -972,7 +972,7 @@ SHRBTextStyler >> pixelHeight [
 SHRBTextStyler >> privateStyle: aText [
 	| ast |
 	ast := classOrMetaClass compiler
-		source: aText;
+		source: aText asString;
 		noPattern: self isForWorkspace ;
 		options:  #(+ optionParseErrors + optionSkipSemanticWarnings);
 		requestor: workspace;


kasperosterbye:
I tend to think the error is in `SHRBTextStyler>>#privateStyle:` as it is the method which puts a Text into the Compiler using `source:` which expect a String (indicated by its argument aString). 

That is, it is the method:
```Smalltalk
privateStyle: aText
	| ast |
	ast := classOrMetaClass compiler
		source: aText asString; "<------- add asString here"
		noPattern: self isForWorkspace ;
		options:  #(+ optionParseErrors + optionSkipSemanticWarnings);
		requestor: workspace;
		parse.				
	ast doSemanticAnalysis.
	^ self style: aText ast: ast
```
However, the PR by Marcus solves the problem flat out.

Ducasse:
Yes good catch

Issue #5300, opened 2019-12-04
Traits: categories are not updated when rebuilding classes


MarcusDenker:
When e.g. removing a trait, the categories of the class side of the class retain the class organiser entries (but the methods are removed).

"
	t1 := self createT1.
	t2 := self createT2.
	c1 := self newClass: #C1 with: 'g h' asSlotCollection uses: t1 + t2.
	c1 := self newClass: #C1 with: 'g h' asSlotCollection uses: {}.		

	c1 organization allMethodSelectors do: [:each | self assert: (c1 includesSelector: each) ].
	c1 class organization allMethodSelectors do: [:each | 
		self assert: (c1 class includesSelector: each) ]
"

PR #5301, opened 2019-12-04
5300-Traits-categories-are-not-updated-when-rebuilding-classes

MarcusDenker:
a test and fix for #5300: categories not upated when Traits are removed from classes.


MarcusDenker:
tests green, benchmark step fails (we should remove the benchmarks for now from the CI...)

tesonep (approved):

PR #5304, opened 2019-12-04
fix/ffilibrary-should-return-itself

estebanlm:
FFILibrary needs to return self when calling #ffiLibrary

### DIFF
--- a/src/UnifiedFFI/FFILibrary.class.st
+++ b/src/UnifiedFFI/FFILibrary.class.st
@@ -51,6 +51,12 @@ FFILibrary >> calloutAPIClass [
 	^ FFICalloutAPI calloutAPIClass 
 ]
 
+{ #category : #accessing }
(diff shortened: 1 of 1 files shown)


akgrant43:
Hi Esteban,

This looks sensible, but was there a particular failure that triggered the realisation?

Thanks,
Alistair

MarcusDenker (approved):

estebanlm:
@akgrant43 yes there was. 
In the cases were you use the library as an entry point for a function (take `LibC` as an example). The call to `#ffiLibrary` was actually answering `ffiLibraryName asFFILibrary` which is always an `FFIUnknownLibrary`.
The consequence is that we cannot change `#calloutAPI` or `#runner`, and obviously this is a bug :)

akgrant43:
@estebanlm Thanks for the explanation!

Ducasse:
Can we get a better method comment?

PR #5318, opened 2019-12-05
5314-Code-completion-issue-in-methods-with-syntax-error

### LINKED ISSUE #5314: Code completion issue in methods with syntax error
Let's have a method starting with an error in method name (it was started as copy of call in existing code). When the code completion starts, it immediately does a mess as if the method would be accepted. It switches the cursor position and opens the completion menu on the wrong place.

```smalltalk
registerSharedAction: action as: action memberName from: aGroup

    curr
```
D8NfK43UQP

MarcusDenker:
turn off shadowed variable warning when #optionSkipSemanticWarnings is enabled. fixes #5314


pavel-krivanek (approved):
works :-)

Issue #4847, opened 2019-10-07
No text annotation for background color 


kasperosterbye:
It is possible to set font, color, emphasis most other aspects of text rendering using `addAttribute:` on instances of class `Text`. However, it seems not possible to set the background color of a text.

My usage scenarie is this: In Github `markdown`, it is common to use `inline code blocks` using ticks. I would like to be able to render this in-image.

kasperosterbye:
This was fixed by PR #5326

PR #5326, opened 2019-12-06
TextBackgroundColorAttribute

kasperosterbye:
Adds a TextBackgroundColor with support in Rubrics RubAbstractTextArea

It works on my machine, and want to see if the CI finds errors outside what I can do.

Example: 
'this is a text' asText addAttribute: (TextBackgroundColor color: Color veryLightGray lighter) from: 6 to: 9

### DIFF
--- a/src/Rubric/RubAbstractTextArea.class.st
+++ b/src/Rubric/RubAbstractTextArea.class.st
@@ -572,7 +572,8 @@ RubAbstractTextArea >> completionPostAction [
 RubAbstractTextArea >> compose [
 
 	self prepareParagraphToCompose.
-	self paragraph compose 
+	self paragraph compose.
+	self markBackgroundColors
 ]
 
 { #category : #composing }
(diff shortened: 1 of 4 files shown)


jecisc:
Thank you Kasper!

I needed that some month ago but I had no time to implement it :) I'll maybe be able to do what I wanted to do before :)

kasperosterbye:
I need it to render the github inline codeblock `like this one`, in the pharo image. Now I can move on to figuring a way to do tables.... ( I have the github markdown parser in place).

jecisc:
I often add insecable spaces without noticing so I want to add a metalink in Rubric to highlight them. But Rubric seems to use Text to do the syntax highlighting.

kasperosterbye:
I do not know what the "continuous-integration/jenkins/pr-merge" errors are - are they serious?

kasperosterbye:
@jecisc I wanted specifically to extend Rubric to understand Text with an attribute for background color. This allow me to use Text to specify what to render with a background color.

For your problem, can you extend SHRBTextStyler to add background color to those spaces?, then with this extension, they should me marked.

jecisc:
@kasperosterbye You have a dependency issue.

```
  RubTextSegmentMorph
  TextBackgroundColorSegmentMorph
  TextBackgroundColorSegmentMorph>>#color:from:to:
  TextBackgroundColorSegmentMorph>>#color:
```

So you are using Rubric in Text, but Text does not depend on rubric.

kasperosterbye:
@jecisc Thanks, I can fix that one (I think).

kasperosterbye:
@jecisc I believe I fixed the dependency. However, it fails for unix32. Not sure if it is my code. I cannot see any unix specific things in what I did.

kasperosterbye:
fixes #4847

kasperosterbye:
I have no idea why this fails for a 32bit unix I am afraid.

MarcusDenker:
the failure om linux is not related (this crash happens randomly)

MarcusDenker:
now the tests are green but the benchmark fails. As I already mentioned multiple times: we *really* need to remove the benchmark step. Right now it *gurantees* that the CI ends with a failure.

MarcusDenker (approved):
ok

PR #5338, opened 2019-12-08
[tests][cleanup] converted test in comment to them properly,

hogoww:
slight comment modification too.

Sidenote, #withSeparatorsCompacted selector feels wrong to me, since it's only treating spaces

### DIFF
--- a/src/Collections-Strings/String.class.st
+++ b/src/Collections-Strings/String.class.st
@@ -2649,26 +2649,25 @@ String >> withNoLineLongerThan: aNumber [
 
 { #category : #converting }
 String >> withSeparatorsCompacted [
-	"Return a copy of the receiver with each sequence of whitespace (separator) 
-	characters replaced by a single space character"
+    "Returns a copy of the receiver with each sequence of whitespace (separator) 
+    characters replaced by a single space character"
+    "' test ' withSeparatorsCompacted >>> ' test '"
+    "' test  test' withSeparatorsCompacted >>> ' test test'"
+    "'test  test      ' withSeparatorsCompacted >>> 'test test '"
 
-	self isEmpty ifTrue: [ ^ self ].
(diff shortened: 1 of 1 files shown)


src/Collections-Strings/String.class.st
```diff
@@ -2649,26 +2649,25 @@ String >> withNoLineLongerThan: aNumber [
...
-	"' test ' withSeparatorsCompacted"
-	"' test test' withSeparatorsCompacted"
-	"'test  test		 ' withSeparatorsCompacted"
+    "Returns a copy of the receiver with each sequence of whitespace (separator) 
+    characters replaced by a single space character"
+    "' test ' withSeparatorsCompacted >>> ' test '"
+    "' test  test' withSeparatorsCompacted >>> ' test test'"
+    "'test  test      ' withSeparatorsCompacted >>> 'test test '"
+
+    self isEmpty ifTrue: [ ^ self ].
```
jecisc, line 2658:
```suggestion
    self ifEmpty: [ ^ self ].
```

jecisc (commented):
I suggested a small improvement.

src/Collections-Strings/String.class.st
hogoww: Often in the kernel classes, those kind of shortcuts are omitted. I assume because of some inlining. I'll wait for Marcus to approve this as well =)

MarcusDenker: ifEmpty is a bit odd... some people feel that it goes to far and do not use it...  we I think never added a rule to enforce using it due to that.

MarcusDenker: for inlining: isEmpty IifTrue: is a message send + inlined ifTrue:, so it indeed is faster.

MarcusDenker (approved):

PR #5341, opened 2019-12-08
5339-Double-clicking-on-each-should-not-select-the-characte--only-each

### LINKED ISSUE #5339: Double clicking on :each should not select the characte : only each.
while doubleclicking on do: should. 

Like that we can inspect and manipulate the argument of a block inside a debugger.

Ducasse:
fixes: #5339, let us continue to fix all the boring glitches that annoy us and we will be a lot more productive!


kasperosterbye:
Works like a charm.

PR #5342, opened 2019-12-09
5309-Rules-display-wrong-or-not-all-codelines

### LINKED ISSUE #5309: Rules display wrong or not all codelines
- Create a class Foo
- provide a #foo method with a message send to a **non-existing** #bar method
- by default the line with the bar message send is marked in red (which is intended and OK for a simple method)

But 
 - it does not work if the message name appears in the comment, because then the 
   name in the comment is marked
 - it does not work if the message send appears several times in the same method, because then only the first one is marked

image

kasperosterbye:
Fixes #5309

Finds multiple errors and report them by position instead of search.


MarcusDenker (approved):
ok

PR #5345, opened 2019-12-09
5344-DrTests-Allow-one-to-prioritise-tree-view

### LINKED ISSUE #5344: [DrTests] Allow one to prioritise tree view
For now the pragma do not allow to prioritise the tree view in the droplist.
I propose to change these pragmas so the order is deterministic.

juliendelplanque:
Now tree view are ordered when they are declared via the pragma.
This allows to have a deterministic order in the drop-list which was not the case until now.

Added test to ensure tree views are well ordered.

Fix #5344

### DIFF
--- a/src/Deprecated80/DTCommentToTest.extension.st
+++ b/src/Deprecated80/DTCommentToTest.extension.st
@@ -1,0 +1,7 @@
+Extension { #name : #DTCommentToTest }
+
+{ #category : #'*Deprecated80' }
+DTCommentToTest >> oldPragmaForResultTrees [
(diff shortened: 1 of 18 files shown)


MarcusDenker:
Failing test: 

Got 'for tests' instead of 'test result'.
Stacktrace
DTMockPluginTest(TestAsserter)>>assert:description:resumable:
DTMockPluginTest(TestAsserter)>>assert:description:
DTMockPluginTest(TestAsserter)>>assert:equals:
DTMockPluginTest>>testResultTreeViews

juliendelplanque:
I will fix this. This is a mistake of mine.

PR #5358, opened 2019-12-11
5343-Inconsistency-in-the-rules-relating-to-checking-the-code-of-methods

### LINKED ISSUE #5343: Inconsistency in the rules relating to checking the code of methods
In looking at #5309, I found other inconsistencies in the rules relating to checking the **code** (as opposed to other aspects) of methods - I have found two issues relating to misbehaving methods:

1. Rule only looks for existence of violations, not all occurences.
2. Rule do not report location of violation, or report it incorrectly.

Any use of `basicCheck:` will cause both of these issue - the worst example I found is `RBMethodSourceContainsLinefeedsRule`, as it actually could tell you were the LineFeeds were, but do not.

The `basicCheck:` should not be marked as subclassResponsibility, as it is better to implement `check:forCritiquesDo:` when critiquing the code of a method.

*Notice: I have not looked at the rules which relate to classes, the organization of methods etc. Only those rules which looks at the code of a method.*

The following two expressions guestimates which rules to check:
```Smalltalk
(ReAbstractRule allSubclasses select: [ :rule | 
	(rule lookupSelector: #check:forCritiquesDo: ) sourceCode beginsWith: 'check: aMethod' ]) 
		collect: [:rule | rule lookupSelector: #check:forCritiquesDo:].
```

```Smalltalk
(ReAbstractRule allSubclasses 
	select: [ :rule | (rule lookupSelector: #basicCheck: ) sourceCode beginsWith: 'basicCheck: aMethod' ]) 
		collect: [:rule | rule lookupSelector: #basicCheck: ].
```

kasperosterbye:
Fixes #5343

This changes many of the code smell rules to indicate where in the source of a method  the problem is, and not merely if it is present.


Ducasse:
Thanks a lot!

MarcusDenker:
strange error:


Syntax Error on line 2: 'Unmatched '' in string literal.'
=========================================================
1: basicName: anObject
2: 	self deprecated: 'Use #name: instead.' transformWith: '`@receiver basic

kasperosterbye:
Indeed a strange error. In particular as I did not touch those classes or methods

MarcusDenker:
I restarted the test run

MarcusDenker:
Better: now the only relevant failing test is 

SendsDeprecatedMethodToGlobalRuleTest>>testBasicCheck3

ReExceptionProperty>>rule
Stacktrace
ReExceptionProperty(Object)>>doesNotUnderstand: #rule
[ :eprop | eprop rule class = SendsDeprecatedMethodToGlobalRule ] in SendsDeprecatedMethodToGlobalRuleTest>>testBasicCheck3
[ :item | (aBlock value: item) ifTrue: [ ^ false ] ] in OrderedCollection(Collection)>>noneSatisfy:
OrderedCollection>>do:
OrderedCollection(Collection)>>noneSatisfy:
SendsDeprecatedMethodToGlobalRuleTest>>testBasicCheck3

kasperosterbye:
It is not in my code, but I might have broken something - investigating...
Wrong - it was in my code, it just manifested itself in an interesting manner.

kasperosterbye:
This time it is in the `GoferApiTest>>#testRepositoryErrors` causing trouble.
Its success or failure depends on the stability of the server: `http://pharo-project.org/page-that-will-never-ever-exist`. I too experience non-responsiveness from that server from time to time.

One way to fix it is to use this URL instead:
`http://google.com/pharo_official_language_for_android`

And we can then hope one day the test will indeed fail because the page will exist, in which case I will buy Stef and Marcus a good bottle of wine.

MarcusDenker (approved):

PR #5359, opened 2019-12-11
Revert "[Perfs] x10 on method compiling with #deferFlushDuring:"

MarcusDenker:
Reverts pharo-project/pharo#5292


VincentBlondeau:
Ok. Good for me since it breaks some projects. But it is fair to integrate this one too https://github.com/pharo-project/pharo/pull/5257 since the consequences are similar.

VincentBlondeau (approved):

PR #5365, opened 2019-12-12
DebugSession modification for DebuggerSelector

dupriezt:
## Disclaimer
This pull request adds code to DebugSession that is necessary for DebuggerSelector to work.
**This code is not called by the standard image and has no impact on its current behaviour, so it is safe to merge before P8's release**
DebuggerSelector will be added to NewTools, but it requires some changes on the DebugSession class that cannot be made via extension methods because they include adding instance variables.

## For the details
This code adds a reference counting mechanism to DebugSession, so that debuggers can manually keep them alive. DebugSession with no debuggers keeping them alive can be terminated by DebuggerSelector.
This PR also includes an observer pattern on DebugSession, so that objects can be notified when a debugger wants to keep alive or stop keeping alive a DebugSession.

### DIFF
--- a/src/Debugger-Model/DebugSession.class.st
+++ b/src/Debugger-Model/DebugSession.class.st
@@ -26,7 +26,9 @@ Class {
 		'name',
 		'interruptedContext',
 		'interruptedProcess',
-		'errorWasInUIProcess'
+		'errorWasInUIProcess',
+		'objectsThatWantToKeepMeAlive',
+		'subscribersToKeepAliveEvents'
 	],
 	#classVars : [
 		'LogDebuggerStackToFile'
@@ -100,6 +102,11 @@ DebugSession class >> named: aString on: aProcess startedAt: aContext [
 		detectUIProcess
 ]
 
+{ #category : #keepAlive }
+DebugSession >> addSubscriberToKeepAliveEvents: aNewSubscriber [
+	self subscribersToKeepAliveEvents add: aNewSubscriber
+]
+
 { #category : #'debugging actions' }
 DebugSession >> clear [
 	"If after resuming the process the user does plan to reuse this session with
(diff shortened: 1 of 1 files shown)


tesonep:
Hi, @dupriezt two questions:

1. Can it be moved to Pharo 9, the Pharo 9 development will start really soon / already started?
2. Is reference counting the best solution to it? Cannot it be resolved with the solutions we already have in the image (Weak structures, Weak registries, etc)? This is not a trivial change as problems with the release of debugging sessions could result in the leak of all the objects referred by the debugged process.

dupriezt:
@tesonep 
>     1. Can it be moved to Pharo 9, the Pharo 9 development will start really soon / already started?

I guess it could. But I don't see Pharo 9 in the pharo-launcher for the moment, so having it in Pharo 8 would be more convenient. Also, as I pointed out, this code doesn't do anything in pharo 8 because it is not called, so it doesn't cost anything.

>     2. Is reference counting the best solution to it? Cannot it be resolved with the solutions we already have in the image (Weak structures, Weak registries, etc)? This is not a trivial change as problems with the release of debugging sessions could result in the leak of all the objects referred by the debugged process.

For the record, it's not exactly reference-counting.

I don't know if it's the best solution, so I'll just describe the current system, and the improvement I want to make.

In the current system, there are three actors:
- the debugged processes
- the DebugSessions on these processes
- the unique debugger (GT-Debugger)

Processes are not garbage-collected.
When a process is being debugged, a DebugSession is created for it, which is given to an instance of debugger. When the debugger window is closed, it terminates the process, and both the debugger and the DebugSession get garbage collected later.

In the system I'm building (DebuggerSelector and DebugSessionStorage), I have multiple debuggers, and I want to keep the DebugSessions alive even when a debugger is closed, so that they can be opened with another debugger. To prevent memory leaks, DebugSessionStorage only keeps the last N DebugSessions. For the other DebugSessions, DebugSessionStorage stops referencing them so they get garbage-collected, and terminates the corresponding process.
An exception to this system (and the reason for the code in this PR) is that a DebugSession that is currently open in a debugger should not be removed from the storage and its process be terminated. To this end, the contract is that when a debugger opens, it calls #keepAlive: on the DebugSession is opens, and when it closes, it calls #stopKeepingAlive:. The debugger passes itself as argument both times. So every DebugSessions (and therefore DebugSessionStorage) can now which DebugSessions are currently opened in debuggers, which allows DebugSessionStorage not to remove+terminate those that are opened.

Ducasse:
Ok after discussion with Pablo and Thomas we can merge because nothing is sent. It just makes sure that we will be able to load newtools in Pharo 80.

PR #5366, opened 2019-12-12
Demeter's law: fix two cases

MarcusDenker:
Fix two cases of Demeter's law violoations
- the Rub Editor should know if it is for scripting or not
- OCLiteralVariable should have a #read method

### DIFF
--- a/src/OpalCompiler-Core/OCLiteralVariable.class.st
+++ b/src/OpalCompiler-Core/OCLiteralVariable.class.st
@@ -91,6 +91,11 @@ OCLiteralVariable >> name [
 	^ assoc name
 ]
 
+{ #category : #reading }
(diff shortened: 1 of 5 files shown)


src/OpalCompiler-Core/OCLiteralVariable.class.st
```diff
@@ -91,6 +91,11 @@ OCLiteralVariable >> name [
 	^ assoc name
 ]
 
+{ #category : #reading }
+OCLiteralVariable >> read [
+	^assoc read
```
hogoww: ^ assoc
Little formating

hogoww (commented):
Formatting is just for the principle, but whatever.
The stuff that change looks good anyway!

Issue #5369, opened 2019-12-13
stepToSendOrReturn causes an error on a dead context


j-brant:
If a context is dead sending `#stepToSendOrReturn` causes an error: 
```
| semaphore context |
semaphore := Semaphore new.
[ context := thisContext.
	semaphore signal ] fork.
semaphore wait.
context stepToSendOrReturn
```
The problem is that the `#willSend` message is sent before the `#isDead` message and `#willSend` assumes a non-dead context.

PR #5370, opened 2019-12-13
5369-stepToSendOrReturn-causes-an-error-on-a-dead-context

j-brant:
fix and test for sending stepToSendOrReturn to a dead context -- issue #5369

### DIFF
--- a/src/Debugging-Core/Context.extension.st
+++ b/src/Debugging-Core/Context.extension.st
@@ -336,9 +336,8 @@ Context >> stepToSendOrReturn [
 	returning a value to the receiver (that is, until switching contexts)."
 
 	| context |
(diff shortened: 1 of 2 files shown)


Ducasse:
Tx. Looks ok to me. I will let a chance to other to know that changes and integrate it.

Ducasse (commented):
Hi John

when discussing with Guille, we are thinking that a simpler way to create a dead context that may be more reliable is to just create a new context and set the PC to nil. 

The forking and signalling of semaphore depend on the fact that when the test process resumes
the second context has already return.

S. and G.

j-brant:
I changed the code so that it doesn't need to run the new process: 
```
deadContext
	| process context |
	process := [  ] newProcess.
	context := process suspendedContext.
	context pc: nil.
	^ context
```

Ducasse:
Thanks this is cool because we are fighting with flaky tests.

PR #5374, opened 2019-12-13
Revert "Revert "[Perfs] x10 on method compiling with #deferFlushDuring

VincentBlondeau:
This reverts the revert of the initial PR #5292.
Modifications has to be made so traits compilation is handled on better way.


estebanlm:
error not related

PR #5375, opened 2019-12-13
Adding the conversion of the image to sista bytecode

tesonep:


Ducasse:
This test is failing

testTallyInstructions
	"Depends on compiler settings... If everything is compiled to the V3PlusClosures bytecode set, then the result is 27.
	 If everything is compiled to the SistaV1 bytecode set, then the result is 26. If execution is mixed between method
	compiled with V3PlusClosures and SistaV1, result is then different and untested."
	
	"We test class name and not the class because there are multiple versions of the encoders depending on the compiler used."
	| expectedResult |
	expectedResult := (aBlockContext method encoderClass = EncoderForSistaV1) ifTrue: [26] ifFalse: [27].
	self assert: (Context tallyInstructions: aBlockContext ) size equals: expectedResult

Ducasse:
Pablo I would move this for P9

estebanlm:
But is just one test. 
Seems "doable" :)

Ducasse:
OKi but I will be able to say that I told you :) :)

tesonep:
It is not only one test, we require a new version of fuel to support
methods with sista, as the header in sista is 64 bits and not 32.

I vote to move it to Pharo 8.1 or whatever is called the next version of
pharo 8

estebanlm:
ok :)

guillep:
Hi, complicated issue seemingly :).

I understand we are missing
 - `testTallyInstructions` This is strange because the image should be entirely in sista bytecodes...
 - an update in Fuel. Does it have to do with https://github.com/pharo-project/pharo/pull/5871? Or an additional update is needed?

Ducasse:
What is needed to make progress on this issue?

Ducasse:
@guillep 
@tesonep 
```
tallyInstructions: aBlock
	"This method uses the simulator to count the number of occurrences of
	each of the Smalltalk instructions executed during evaluation of aBlock.
	Results appear in order of the byteCode set."
	| tallies |
	tallies := Bag new.
	thisContext sender
		runSimulated: aBlock
		contextAtEachStep:
			[:current | tallies add: current nextByte].
	^tallies sortedElements
```

So if this test is not working when we set the sista bytecode do you think that it means that the simulator is not fully working?

MarcusDenker:
I think we do not need this anymore?

PR #5382, opened 2019-12-14
execution counters sometimes produce errors when uninstalling

### LINKED ISSUE #5381: ExecutionCounters sometimes produce errors when uninstalling
There seems to be missing methods.
- install a counter in a method through the browser
- use it
- remove the counter: sometimes it fails (DNU)

StevenCostiou:
Added tests and fixing bugs in exec counter.
Basically missing method and erroneous uninstall.
fixes #5381


Ducasse:
Thanks thanks thanks. 
Yes more tests.
Yes more tests.
Yes more tests.
Yes more tests.

PR #5384, opened 2019-12-16
Improving the FreeTypeCache mutex.

tesonep:
Extending the mutex to the access also.


estebanlm (approved):

Ducasse:
Strange the build did not pass.

Issue #5372, opened 2019-12-13
Completion failure when using super in a trait


pavel-krivanek:
When in method, that belongs to a trait, you start to use message call in form `super message` then during the writing of the `message` selector, the completion fails with error `#allSelectors was sent to nil`

![Pharo_sNUQO0WguH](https://user-images.githubusercontent.com/572494/70781256-2b72fa00-1d86-11ea-869a-8f7515afb680.png)

MarcusDenker:
Has been merged in Pharo8, I checked and the fix is in Pharo9, too

PR #5385, opened 2019-12-16
5372-Completion-failure-when-using-super-in-a-trait

MarcusDenker:
fix for #5372


tesonep:
The test error looks unrelated, but as it is the code completion I am not sure.

MarcusDenker:
It does not fail on my machine with the fix applied on the mac...

MarcusDenker:
(lets try and merge...)

PR #5390, opened 2019-12-17
rollback-bad-comment-when-copying-class

estebanlm:
remove bad comment copy. 

This is bad, very bad, because it introduces confusion producing a lot of comments that are a plain copy from previous class, and wrong! (I already seen some around)


Ducasse:
I agree :) This is fun.

MarcusDenker (approved):

PR #5394, opened 2019-12-18
Adding missing pragmas to methods with halts for test and dev purpose

StevenCostiou:
Adding <haltOrBreakpointForDevelopment> and <haltOrBreakpointForTest>  pragmas to methods with halts for testing or debugging purpose.

Not having these pragmas produces tool interference when trying to know what are the halts in the base system.

### DIFF
--- a/src/Calypso-SystemPlugins-Reflectivity-Queries-Tests/ClyActiveBreakpointsQueryTest.class.st
+++ b/src/Calypso-SystemPlugins-Reflectivity-Queries-Tests/ClyActiveBreakpointsQueryTest.class.st
@@ -24,14 +24,16 @@ ClyActiveBreakpointsQueryTest >> installBreakpointInto: aMethod [
 ClyActiveBreakpointsQueryTest >> setUp [
 	super setUp.
 	
-	installedBreakpoints := OrderedCollection new
+	installedBreakpoints := OrderedCollection new.
+	ClyClassWithHalts compile: 'methodWithHalts1 self halt'.
+	ClyClassWithHalts compile: 'methodWithHalts2 self halt'.
 ]
 
 { #category : #running }
 ClyActiveBreakpointsQueryTest >> tearDown [
-	installedBreakpoints do: [ :each | each remove ].
-	installedBreakpoints removeAll.
-	
+	ClyClassWithHalts removeSelector: #methodWithHalts1.
+	ClyClassWithHalts removeSelector: #methodWithHalts2.
+	Breakpoint removeFromMethod: (ClyClassWithBreakpoints >> #methodWithBreakpoints).
 	super tearDown
 ]
 
@@ -53,6 +55,7 @@ ClyActiveBreakpointsQueryTest >> testCheckIfEmpty [
 
 { #category : #tests }
 ClyActiveBreakpointsQueryTest >> testComparison [
+
 	self assert: ClyActiveBreakpointsQuery new equals: ClyActiveBreakpointsQuery new.
 
 	self assert: ClyActiveBreakpointsQuery newIncludingHaltsForTesting equals: ClyActiveBreakpointsQuery newIncludingHaltsForTesting.
--- a/src/UnifiedFFI/FFIFunctionParser.class.st
+++ b/src/UnifiedFFI/FFIFunctionParser.class.st
@@ -289,8 +289,8 @@ FFIFunctionParser >> returnType: aType [
 
 { #category : #accessing }
 FFIFunctionParser >> setOn: aStreamSource [
-
-	| functionDeclarationArray |
+	<haltOrBreakpointForDevelopment>
+	| functionDeclarationArray | 
 	functionDeclarationArray := aStreamSource isString ifTrue: [ | node |
 			node := RBParser parseExpression: '#(', aStreamSource, ')'.
 			node isLiteralNode ifFalse: [ self halt ].
--- a/src/UnifiedFFI/FFIFunctionParser.class.st
+++ b/src/UnifiedFFI/FFIFunctionParser.class.st
@@ -289,11 +289,10 @@ FFIFunctionParser >> returnType: aType [
 
 { #category : #accessing }
 FFIFunctionParser >> setOn: aStreamSource [
-	<haltOrBreakpointForDevelopment>
 	| functionDeclarationArray | 
 	functionDeclarationArray := aStreamSource isString ifTrue: [ | node |
 			node := RBParser parseExpression: '#(', aStreamSource, ')'.
-			node isLiteralNode ifFalse: [ self halt ].
+			node isLiteralNode ifFalse: [FFIUnsupportedFunctionDeclarationError signalFor: node ]. 
 			node evaluate
 		] ifFalse: [ aStreamSource ].
 			
--- a/src/Calypso-SystemPlugins-Critic-Queries-Tests/ClyClassWithProblemMethods.class.st
+++ b/src/Calypso-SystemPlugins-Critic-Queries-Tests/ClyClassWithProblemMethods.class.st
@@ -9,13 +9,14 @@ Class {
 
 { #category : #'method examples' }
 ClyClassWithProblemMethods >> methodWithHalt [
-	<haltOrBreakpointForTesting>
+	<haltOrBreakpointForTesting> 
 	self halt.
 ]
 
 { #category : #'method examples' }
 ClyClassWithProblemMethods >> methodWithHalt2 [
-	self halt.
+	<haltOrBreakpointForTesting> 
+	self halt
 ]
 
 { #category : #'method examples' }
--- a/src/Calypso-SystemPlugins-Reflectivity-Queries-Tests/ClyClassWithHalts.class.st
+++ b/src/Calypso-SystemPlugins-Reflectivity-Queries-Tests/ClyClassWithHalts.class.st
@@ -9,11 +9,13 @@ Class {
 
 { #category : #methods }
 ClyClassWithHalts >> methodWithHalts1 [
+	<haltOrBreakpointForTesting> 
 	self halt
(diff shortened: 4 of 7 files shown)


Ducasse:
Steven can you check because some tests are broken 

 MacOSX32.Calypso.SystemPlugins.Reflectivity.Queries.Tests.Breakpoints.ClyActiveBreakpointsQueryTest

StevenCostiou:
I updated. The pragma is not compatible with Calypso tests, so instead I generate halts and remove them after tests, so that there are no junk halts left in the code (which should be avoided in general...).

src/UnifiedFFI/FFIFunctionParser.class.st
```diff
@@ -289,8 +289,8 @@ FFIFunctionParser >> returnType: aType [
 
 { #category : #accessing }
 FFIFunctionParser >> setOn: aStreamSource [
-
-	| functionDeclarationArray |
+	<haltOrBreakpointForDevelopment>
+	| functionDeclarationArray | 
 	functionDeclarationArray := aStreamSource isString ifTrue: [ | node |
 			node := RBParser parseExpression: '#(', aStreamSource, ')'.
 			node isLiteralNode ifFalse: [ self halt ].
```
dupriezt: This looks like a debugging halt left behind, not like a halt set there on purpose.
Instead of adding \<haltOrBreakpointForDevelopment\> to paperwall the issue, I would replace the halt with an error like:
```Smalltalk
node isLiteralNode ifFalse: [ self error: 'A literal node is expected' ].
```

The PR looks good to me otherwise.

dupriezt (changes requested):

src/UnifiedFFI/FFIFunctionParser.class.st
StevenCostiou: I agree. Why not an assertion instead?

However, why not keeping the <haltOrBreakpointForDevelopment> in the dev versions?
That means that it should not be there in production images.

guillep: Oups, this seems a leftover from a refactoring I did.
I'd prefer a non assertion exception. Either something like what @dupriezt proposes or maybe another kind of exception like `FFIUnsupportedFunctionDeclarationError` ?

dupriezt: Could be an assertion too. I don't have a preference.

> However, why not keeping the in the dev versions?

We have separate dev versions and production versions for pharo?

StevenCostiou: Well yes.
Or, at least, there are the release and the dev.
Ok for the FFIUnsuporterd error.
For the pragma, either nobody ever leaves a halt in the code and it gets rejected every time, or we need the pragma so that it does not mess with the tools.

guillep (approved):
Good!

src/Calypso-SystemPlugins-Reflectivity-Queries-Tests/ClyActiveBreakpointsQueryTest.class.st
```diff
@@ -24,14 +24,16 @@ ClyActiveBreakpointsQueryTest >> installBreakpointInto: aMethod [
 ClyActiveBreakpointsQueryTest >> setUp [
 	super setUp.
 	
-	installedBreakpoints := OrderedCollection new
+	installedBreakpoints := OrderedCollection new.
+	ClyClassWithHalts compile: 'methodWithHalts1 self halt'.
```
dionisiydk, line 28: You perform code changes during tests in live environment. I think it is very bad practice and it should be avoided. If you want to "emulate methods" you need to create classes in your own test environment. This way it will not affect the live system state.

dionisiydk (changes requested):

dionisiydk:
> I updated. The pragma is not compatible with Calypso tests

There is already support for this pragma. But tests were not adopted. Try following change:

```Smalltalk
ClyActiveBreakpointsQueryTest>>createQuery
	^(ClyActiveBreakpointsQuery from: ClyClassScope of: ClyClassWithHalts in: environment)
		excludesHaltsForTesting: false.
```

>, so instead I generate halts and remove them after tests, so that there are no junk halts left in the code (which should be avoided in general...).

I do not see why it should be avoided. It is a good to have examples in image. And here is a live example of methods with halt, how they looks and behaves in the browser

dionisiydk:
Can you also check the critique "halt is left in the code"? It needs to take into account this pragma. Otherwise browser highlights problems in such testing methods

src/Calypso-SystemPlugins-Reflectivity-Queries-Tests/ClyActiveBreakpointsQueryTest.class.st
StevenCostiou: That is done for a lot of thing that is meta. 
For instance testing metalinks implies recompiling.
To me, code changes already happen during live tests.
Can you specify why you think it is wrong?

> > I updated. The pragma is not compatible with Calypso tests
> 
> There is already support for this pragma. But tests were not adopted. Try following change:
> 
> ```smalltalk
> ClyActiveBreakpointsQueryTest>>createQuery
> 	^(ClyActiveBreakpointsQuery from: ClyClassScope of: ClyClassWithHalts in: environment)
> 		excludesHaltsForTesting: false.
> ```

Ah, thanks. I will use that if we give up on the dynamic method creation, but all the Calypso model is hard to grasp so I did not see that (nor took the time to find it I admit).

> > , so instead I generate halts and remove them after tests, so that there are no junk halts left in the code (which should be avoided in general...).
> 
> I do not see why it should be avoided. It is a good to have examples in image. And here is a live example of methods with halt, how they looks and behaves in the browser

Well, Calypso tests or framework are not the place to demonstrate methods with halts, it is rather in the documentation of the system.
And how they work in the browser, as users can just put a halt, they can see it live.

The problem is that rogue halts (without the pragma) are going to halt if executed on the base system (which seems wrong?) and additionaly, it is interfering with debugging tools that we are building (and I don't see why it should).

What I mean is that halts without a specification that they are for tests or example should be avoided in the base image. Maybe we could use a pragma <example> that tools can filter?

StevenCostiou:
> Can you also check the critique "halt is left in the code"? It needs to take into account this pragma. Otherwise browser highlights problems in such testing methods

I will check

src/Calypso-SystemPlugins-Reflectivity-Queries-Tests/ClyActiveBreakpointsQueryTest.class.st
dionisiydk: > That is done for a lot of thing that is meta.
> For instance testing metalinks implies recompiling.
> To me, code changes already happen during live tests.
> Can you specify why you think it is wrong?

Running those tests will mark code to be dirty. After image restart it will show lost changes. 
It is about side effect which tests produce. It is good practice to avoid it

dionisiydk: > Well, Calypso tests or framework are not the place to demonstrate methods with halts, it is rather in the documentation of the system.

I agree. It would be good to have separate reusable package with such examples.

> And how they work in the browser, as users can just put a halt, they can see it live.

Examples demonstrate what is possible in the system. For "just put a halt" you need to know about it. Especially when you are from another language you have no idea about "self halt".
Documentation is a good place of course but it is not a live thing but the real code is.

dionisiydk: > ```smalltalk
> ```smalltalk
> ClyActiveBreakpointsQueryTest>>createQuery
> 	^(ClyActiveBreakpointsQuery from: ClyClassScope of: ClyClassWithHalts in: environment)
> 		excludesHaltsForTesting: false.
> ```
> 
> 
> 
> Ah, thanks. I will use that if we give up on the dynamic method creation, but all the Calypso model is hard to grasp so I did not see that (nor took the time to find it I admit).

Sure it can be difficult to find features. But I'm curious here it is about a class under broken test. Did you try debug test before making conclusion?

StevenCostiou: > > Well, Calypso tests or framework are not the place to demonstrate methods with halts, it is rather in the documentation of the system.
> 
> I agree. It would be good to have separate reusable package with such examples.
> 
> > And how they work in the browser, as users can just put a halt, they can see it live.
> 
> Examples demonstrate what is possible in the system. For "just put a halt" you need to know about it. Especially when you are from another language you have no idea about "self halt".
> Documentation is a good place of course but it is not a live thing but the real code is.

True, that is why those users will look at the documentation or put breakpoints as in mainstream languages by clicking on the code :p

StevenCostiou: > > That is done for a lot of thing that is meta.
> > For instance testing metalinks implies recompiling.
> > To me, code changes already happen during live tests.
> > Can you specify why you think it is wrong?
> 
> Running those tests will mark code to be dirty. After image restart it will show lost changes.
> It is about side effect which tests produce. It is good practice to avoid it

Hmm true. But that is the only way to test some things (again, metalinks).
I will try your solution with the pragma and the test configuration, but its impossible to do anything for all other code that does the same thing.

guillep:
Hi @StevenCostiou can you check the state of this?

guillep:

guillep:
I'll close this PR. This does not close the original branch of course.
So as soon as you have something else, please open a new PR.

Thanks

Issue #3604, opened 2019-06-21
Rename package should rename Manifest


jecisc:
Currently, we have Manifest classes to hold metadatas about a package. It would be great if the rename package refactoring could also rename the manifest.

PR #5402, opened 2019-12-20
3604 rename package should rename manifest

jordanmontt:
Fixes issue #3604. Added "renameManifest" method on "RPackage" classs. And the "renameManifest" method is called when the package is renamed.

### DIFF
--- a/src/RPackage-Core/RPackage.class.st
+++ b/src/RPackage-Core/RPackage.class.st
@@ -1423,7 +1423,7 @@ RPackage >> renameTo: aSymbol [
 				do: [ :each | SystemOrganizer default removeCategory: each ] ].
 	self renameTagsPrefixedWith: oldName to: newName.
 	self renameExtensionsPrefixedWith: oldName to: newName.
-	self renameManifest: newName.
+	self renameManifest: newName. 
 	self organizer
 		basicUnregisterPackageNamed: oldName;
 		basicRegisterPackage: self.
--- a/src/RPackage-Core/RPackage.class.st
+++ b/src/RPackage-Core/RPackage.class.st
@@ -1374,6 +1374,14 @@ RPackage >> renameExtensionsPrefixedWith: oldName to: newName [
 	
 ]
 
+{ #category : #register }
+RPackage >> renameManifest: aPackageName [
+	| newName packageNameOnlyLetters |
+	packageNameOnlyLetters := aPackageName onlyLetters.
+	newName := 'Manifest', packageNameOnlyLetters.
+	self classes detect:[:e| e isManifest] ifFound: [:e| e rename: newName]
+]
+
 { #category : #private }
 RPackage >> renameTagsPrefixedWith: oldName to: newName [
 	| oldPrefix newPrefix |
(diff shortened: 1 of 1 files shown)


astares:
This one is still having the issues described in https://github.com/pharo-project/pharo/issues/5725
as #onlyLetters is used again.

We should centralize the name building of the manifest classes.

Therefore it would be better
1. we wait with this PR and first merge the PR #5732
2. then this PR should be adopted to use the "TheManifestBuilder class >> manifestClassNameFor: aPackageName" method from PR #5732

guillep:
Hi @astares , I see https://github.com/pharo-project/pharo/pull/5732 has been already integrated.
What are the next steps to do?

astares:
@guillep 

The **new and correct/"safe" logic** to create a manifest class name from a package class name is now implemented in TheManifestBuilder class >> manifestClassNameFor:
This is also considering alphanumeric and we have a release test to assure we do not introduce new cases. This is all in package "Manifest-Core" which also defines things like #manifestClassPrefix and other. So this package as it is now cares on all things around Manifests.

This PR #5402 (as it was done before) still provides the **old logic** to create a manifest name as you can see in provided RPackage>>#renameManifest:
This is in package "RPackage-Core" - caring on packages.


So from the logical point of view the #renameManifest in "RPackage-Core" should
call and use the implementation in TheManifestBuilder class >> manifestClassNameFor: so we do not have a second (wrong) implementation of the logic.

But that would make "RPackage-Core" dependent on package "Manifest-Core", which might **not** be correct. Because from the image it looks like "RPackage-Core" is part of the bootstrap while "Manifest-Core" is (currently) not.

So "who calls whom" has to be answered if we want to introduce new dependencies or not. Another option is that we move the logic to  "RPackage-Core" and let "Manifest-Core" depend on it. 

But then we would have Manifest related logic in the "RPackage-Core" package while one would naturally expect it in "Manifest-Core".

astares:
@guillep 
From the dependency analyzer it shows that  "Manifest-Core" is dependent on "RPackage-Core" which looks a little bit strange in the first place (as for the package renaming we need infos about manifest classes) but makes sense in terms of not having the packages bloated and a slim kernel/package package (see below).

Also the "Manifest-Core" is not really just core functionality for manifests - it provides (beside the builder utility class) also rules - but this is another story and should be solved later by slicing this package (and extract the rules into an own).

**I guess we should think from the bootstrap process upwards:**

The first and real essential "manifest" support is in the "Kernel" package by providing 
a PackageManifest class (which is superclass to all manifests). This is also necessary to be already in the
Kernel package so the Kernel package itself is able provide an own manifest (ManifestKernel).

Having this in mind I think we should 
  Step A: move #manifestClassNameFor: #manifestClassPrefix and #manifestTag to 
              PackageManifest class (but as extensions by "RPackage-Core" on "Kernel" package to keep the kernel slim)

  Step B:  Then we would be clean and are able to call #manifestClassNameFor: from #renameManifest: 
      (in "RPackage-Core") and also from TheManifestBuilder class >> manifestClassNameFor:
      (in  "Manifest-Core").

What do you think?

guillep:
Looks ok to me. I thought this PR was in better shape, it seems there is still a lot to be done 🤣

PR #5405, opened 2019-12-22
2019-12-22-2

### LINKED ISSUE #5404: Better comments for Mutex
We should say that Mutex (also named recursionLock) do not deadlock on nested critical sections.

Ducasse:
fixes: #5404 better class comment and critical:


hogoww (approved):
Good improvement !

Ducasse:
Only cmments!

PR #5409, opened 2019-12-22
5408-About-Processor-yield--Object--yield

### LINKED ISSUE #5408: About Processor yield => Object >> yield
Could we introduce yield in Object to avoid to have to write 
Processor yield?

This looks so dated to have a global variable and refer to it everywhere looks like a bad idea 
and promotes bad design (Smalltalk tools, Smalltalk ui theme....).

Ducasse:
Fixes: #5408 add yield to Object, so that we can rewrite all the Processor direct references.


Ducasse:
I will redo it in a clean image

PR #5410, opened 2019-12-22
5408-About-Processor-yield--Object--yield

Ducasse:
Fixes: #5408 add yield to Object, so that we can rewrite all the Processor direct references.


Ducasse:
I will redo it in clean image

Issue #5406, opened 2019-12-22
trace and trace: display too many quotes


Ducasse:
I suggest to changer their definition as follows: 

```
trace
	self trace: self
```
instead of 

```
trace
	self trace: self printString
```

and 

``` 
trace: aString
	"Log the argument. Use self trace: instead of Transcript show: "
	
	Transcript show: aString asString
```
instead of 

```
trace: aString
	"Log the argument. Use self trace: instead of Transcript show: "
	
	Transcript show: aString
```

which in addition is more robust.

Ducasse:
same for crTrace: and crTrace

bencoman:
Looking at the following snapshot from the github comments, I see the first commit #5414 has only one commit and #5415 has the 52 commits, so what was the action between the two?  What were you trying to achieve? Because if its not easy to do, the perhaps others hit the same problem.
![image](https://user-images.githubusercontent.com/1713447/71328069-e8661300-254b-11ea-90bf-4a2b4382596a.png)

Ducasse:
My image was dirty so I captured all the changes I was working on with enlumineur enhancements. I redid it.

PR #5411, opened 2019-12-22
5406-trace-and-trace-display-too-many-quotes

Ducasse:
Fixes: #5406 more robust trace: that automatically convert to string and in addition do not display spurious single quotes


Ducasse:
I will redo it in a clean image

PR #5412, opened 2019-12-22
About-Processor-yield--Object--yield-2Version

Ducasse:
fixes: #5409 new version from a clean image

### DIFF
--- a/src/Kernel/Object.class.st
+++ b/src/Kernel/Object.class.st
@@ -2102,7 +2102,7 @@ Object >> writeSlotNamed: aName value: anObject [
 ]
 
 { #category : #'process termination handling' }
-Object >> yield [ 
+Object >> yieldProcessor [
 	"Suspend the active process and give back control to the scheduler. It gives a chance to other pending processes to be scheduled."
 	
 	Processor yield
--- a/src/Kernel/Object.class.st
+++ b/src/Kernel/Object.class.st
@@ -2101,6 +2101,13 @@ Object >> writeSlotNamed: aName value: anObject [
 	^(self class slotNamed: aName) write: anObject to: self
 ]
 
+{ #category : #'process termination handling' }
+Object >> yield [ 
+	"Suspend the active process and give back control to the scheduler. It gives a chance to other pending processes to be scheduled."
+	
+	Processor yield
+]
+
 { #category : #accessing }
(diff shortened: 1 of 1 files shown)


Ducasse:
BTW these fixes are important for the books we are producing

src/Kernel/Object.class.st
```diff
@@ -2101,6 +2101,13 @@ Object >> writeSlotNamed: aName value: anObject [
 	^(self class slotNamed: aName) write: anObject to: self
 ]
 
+{ #category : #'process termination handling' }
+Object >> yield [
```
bencoman: I like the idea in principal, but adding methods to Object potentially constrains and clashes with application domain.
For example, a financial application might want to implement #yield...
https://www.investopedia.com/terms/y/yield.asp

And there are many others...
https://www.merriam-webster.com/dictionary/yield

Can we make this method more specific, i.e. #yieldProcessor ?
```suggestion
Object >> yieldProcessor [
``` 

btw, new github feature, you should be able to integrate my suggestion with a single click...
https://help.github.com/en/github/collaborating-with-issues-and-pull-requests/incorporating-feedback-in-your-pull-request

bencoman (approved):
LGTM

hogoww:
Should probably rename it yieldProcess

guillep:
I have conflicting feelings about this PR. I know this PR reduces the amount of references to global state, and that it also produces a hook for users to override. But I also see:

1) it's hiding a dependency that before was clear => you had the ugly global there. Now the dependency analyser cannot catch it easily.

2) adding stuff in Object just makes me noise, but this are my past experiences talking, you know. I was the one giving `Object>>name` its final hit :)

3) introducing this without changing the users of `Processor>>yield` makes it somehow incomplete... What would be a plan to do the rest and how can we plan for it? I wouldn't want to introduce something and then we will forget about it :).

Thoughts?

Ducasse:
I know. Now in Object or not you have the dependency :)
Now as soon as this is integrated I will fix all the reference like a nuts :).

guillep:
> I know. Now in Object or not you have the dependency :)

I know you know. I just prefer that the discussion is out loud, for people that do not know.

> Now as soon as this is integrated I will fix all the reference like a nuts :).

deal

PR #5413, opened 2019-12-22
About-Processor-yield--Object--yield-2Version

Ducasse:
fixes: #5409 new version from a clean image


Ducasse:
Ok I did it from a P8 image and it is not good

PR #5414, opened 2019-12-22
5406-Trace-2ndVersion

### LINKED ISSUE #5406: trace and trace: display too many quotes

Ducasse:
Fixes: #5406 more robust methods and nicer display (less spurious quotes).

### DIFF
--- a/src/Transcript-Core/Object.extension.st
+++ b/src/Transcript-Core/Object.extension.st
@@ -2,14 +2,15 @@ Extension { #name : #Object }
 
 { #category : #'*Transcript-Core' }
 Object >> crTrace [
-	self crTrace: self printString
+	self crTrace: self
 ]
 
 { #category : #'*Transcript-Core' }
 Object >> crTrace: aString [
 	"Log the argument. Use self crTrace: instead of Transcript cr; show: "
 	
-	Transcript cr; show: aString.
+	Transcript cr.
+	self trace: aString
 ]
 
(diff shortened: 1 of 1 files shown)


src/Transcript-Core/Object.extension.st
```diff
@@ -29,31 +30,33 @@ Object >> logExit [
...
-	self trace: self printString
+	self trace: self
 ]
 
 { #category : #'*Transcript-Core' }
 Object >> trace: aString [
 	"Log the argument. Use self trace: instead of Transcript show: "
 	
-	Transcript show: aString
+	Transcript show: aString asString
```
svenvc: the argument should be called anObject instead of aString since you are sending #asString to it

Ducasse: Yes! Thanks for the feedback

Ducasse:
BTW these fixes are important for the books we are producing

tesonep (approved):

PR #5415, opened 2019-12-22
5406-Trace-2ndVersion

Ducasse:
Fixes: #5406 more robust methods and nicer display (less spurious quotes).


bencoman:
Seems like a good idea.  Although 52 commits seems erroneous, like this was done in an old image that was updated.  Can you advise the git actions taken in your Image to get it to this point?

Ducasse:
done from P8 so I close the PR.

PR #5422, opened 2019-12-25
5420-Menus-not-accessible-when-SHIFT-is-pressed

### LINKED ISSUE #5420: Menus not accessible when SHIFT is pressed
Scenario:
  - hold down shift key
  - click on a menu item in world menu (like "System Browser")

=> the action is not performed

So basically that means menues are not accessible / usable when SHIFT is pressed (Windows)

tesonep:
The mouse buttons should only try to handle the editting of the label if the menu item is editable.
Fixes #5420


astares (approved):
Fine - thanks!

PR #5424, opened 2019-12-25
[cleanup] remove empty test

hogoww:
Empty test


hogoww:
Probably should be added as a rule,
when a test has an empty body?

PR #5429, opened 2020-01-01
#5428 resync enlumineur

### LINKED ISSUE #5428: resync enlumineur with master dev
We did a first pass to clean enlumineur and optimize it a bit.

Ducasse:
fixes: #5428 resync enlumineur


Ducasse:
I do not get why the RBMethodTest is not loaded before.... I did not change anything from the package structure. Now probably that the tests are not loaded in the same order. 
What a pain this dependency hell.

PR #5431, opened 2020-01-02
2854-Pharo-7-Saving-the-settings-breaks-local-directory

### LINKED ISSUE #2854: Pharo-7 Saving the settings breaks local directory
I don't think this problem is expressed clearly enough in #2646 - simply saving the settings file in an image will cause all future images to use/share the iceberg-local directory for the image who's settings were saved. 

This causes terrible and confusing issues when using iceberg, as you don't actually get a clean copy of your repo, and loading things like master - can leave you many pull requests behind when you load a baseline (and hence, you will inadvertently have the wrong code).

Bajger:
fixes #2854

- PragmaSetting>>chooseFileDirectory returned string path before this fix instead of file reference instance. This caused problem, when FileLocator>>resolve tried to resolve local directory and file reference was expected. 
- fix will prevent from entering invalid directory directly by typing, so only valid directory setting is accepted (SystemResolver class>> userLocalDirectory:)
- test was added

### DIFF
--- a/src/FileSystem-Core/SystemResolver.class.st
+++ b/src/FileSystem-Core/SystemResolver.class.st
@@ -35,8 +35,10 @@ SystemResolver class >> userLocalDirectory [
 ]
 
 { #category : #accessing }
-SystemResolver class >> userLocalDirectory: aFileReference [
-	UserLocalDirectory := aFileReference
+SystemResolver class >> userLocalDirectory: aFileReferenceOrPathName [
+	|aDirReference| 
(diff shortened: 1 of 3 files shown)


Bajger:
New failure of ZnClientTest>>testGetLargeWideStringHTML (connection timeout) seems to be unrelated to changes for this issue. Please review, if Zinc client might be dependent on setting of local directory. It seems to be random problem, but I might be wrong.

Issue #5443, opened 2020-01-06
Reading out of range Float exponents is inconsistent


svenvc:
It was noted that the following two have different results:

    Float readFrom: '7.5e333'. => Float infinity
    Float readFrom: '7.5e3333'. => Error

Turns out NumberParser>>#readExponent has a bogus and unnecessary test at the end.

More specifically: Float emin and emax define the allowed exponent interval in binary, not decimal. Even so, trying to use a too large exponent gets automatically resolved to infinity when the float instance is created. Hence the test can be removed.

nicolas-cellier-aka-nice:
It might be related to this little game:
https://codegolf.stackexchange.com/questions/7197/crash-your-favorite-compiler/11429#11429

<blockquote><img src="https://cdn.sstatic.net/Sites/codegolf/img/apple-touch-icon@2.png?v=ed60a1645c65" width="48" align="right"><div>Code Golf Stack Exchange</div><div><strong><a href="https://codegolf.stackexchange.com/questions/7197/crash-your-favorite-compiler/11429">Crash your favorite compiler</a></strong></div><div>Write a perfectly legal code in a decent language of your choice whose compiling will either crash the compiler or send it into an infinite loop (infinite compile time).

Restrictions:
Use a stand...</div></blockquote>

svenvc:
Yes, we probably need lots of resource safeguards while parsing, this is not so easy to do well, especially in a language that allows very large structures easily.

> On 6 Jan 2020, at 15:53, Nicolas Cellier <notifications@github.com> wrote:
> 
> It might be related to this little game:
> https://codegolf.stackexchange.com/questions/7197/crash-your-favorite-compiler/11429#11429
> 
> —
> You are receiving this because you authored the thread.
> Reply to this email directly, view it on GitHub, or unsubscribe.
> 

--
Sven Van Caekenberghe - mailto:sven@beta9.be
Beta Nine - software engineering - http://www.beta9.be

PR #5444, opened 2020-01-06
5443-Reading-out-of-range-Float-exponents-is-inconsistent

svenvc:
Remove a bogus test from NumberParser>>#readExponent that resulted in inconsistent behavior
Added FloatText>>#testReadingTooLargeExponents to cover the issue

https://github.com/pharo-project/pharo/issues/5443

### DIFF
--- a/src/Kernel-Tests/FloatTest.class.st
+++ b/src/Kernel-Tests/FloatTest.class.st
@@ -407,6 +407,19 @@ FloatTest >> testReadFromManyDigits [
 	self assert: (Number readFrom: s2) equals: 1 
 ]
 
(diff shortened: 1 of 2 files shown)


hogoww (approved):

hogoww:
@Ducasse You merged it in P8, shouldn't it be in p9 ? (I the compared branch wasn't right)

MarcusDenker:
I checked with pablo: as the impact of the change is likely not negative, we suggest to keep it in Pharo8. Then we do a new PR to have it in Pharo9, too.

I close the revert PR and then do a PR for Pharo9

PR #5445, opened 2020-01-06
5392-Proposed-improvement--performance-of-atRandom-in-class-Bag

svenvc:
Add a faster Bag>>#atRandom (due to faster e numeration) overwriting the generic one from Collection

https://github.com/pharo-project/pharo/issues/5392

### DIFF
--- a/src/Random-Core/Bag.extension.st
+++ b/src/Random-Core/Bag.extension.st
@@ -1,0 +1,18 @@
+Extension { #name : #Bag }
+
+{ #category : #'*Random-Core' }
+Bag >> atRandom: aGenerator [
+	"Answer a random element of the receiver. Uses aGenerator which
+    should be kept by the user in a variable and used every time. Use
+    this instead of #atRandom for better uniformity of random numbers because 
+	 only you use the generator. Causes an error if self has no elements."
+	
+	| rand index |
+	self emptyCheck.
+	rand := aGenerator nextInt: self size.
+	"overwritten to use a faster enumeration"
+	index := 0.
+	self doWithOccurrences: [ :key :count | 
+		index := index + count.
+		rand <= index ifTrue: [ ^ key ] ]
+]
--- a/src/Kernel-Tests/FloatTest.class.st
+++ b/src/Kernel-Tests/FloatTest.class.st
@@ -407,6 +407,19 @@ FloatTest >> testReadFromManyDigits [
 	self assert: (Number readFrom: s2) equals: 1 
 ]
 
+{ #category : #'tests - conversion' }
+FloatTest >> testReadingTooLargeExponents [
+	self assert: (Float readFrom: '7.5e333') equals: Float infinity.
+	self assert: (Float readFrom: '7.5e-333') equals: Float zero.
+	self assert: (Float readFrom: '-7.5e333') equals: Float infinity negated.
+	self assert: (Float readFrom: '-7.5e-333') equals: Float zero.
+
+	self assert: (Float readFrom: '7.5e3333') equals: Float infinity.
+	self assert: (Float readFrom: '7.5e-3333') equals: Float zero.
+	self assert: (Float readFrom: '-7.5e3333') equals: Float infinity negated.
+	self assert: (Float readFrom: '-7.5e-3333') equals: Float zero.
+]
+
 { #category : #'tests - NaN behavior' }
 FloatTest >> testReciprocal [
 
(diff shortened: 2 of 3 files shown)


PharoProject:
Hmm, this got too much included.

How does one reset an image after one PR to the next ?

Because from the Iceberg commit standpoint, only one change was included.

Ducasse:
Usually when you repair your image from a detached head you create a bottom branch representing the state before you do a change, then you do a change and submit a PR and after you can checkout your bottom branch to go back in the situation prior to your change and do another change.

svenvc:
Thx, I will try that next time, I hope I will remember.

> On 6 Jan 2020, at 20:29, StéphaneDucasse <notifications@github.com> wrote:
> 
> Usually when you repair your image from a detached head you create a bottom branch representing the state before you do a change, then you do a change and submit a PR and after you can checkout your bottom branch to go back in the situation prior to your change and do another change.
> 
> —
> You are receiving this because you authored the thread.
> Reply to this email directly, view it on GitHub, or unsubscribe.
> 

--
Sven Van Caekenberghe - mailto:sven@beta9.be
Beta Nine - software engineering - http://www.beta9.be

Ducasse:
I made a little iceberg scripter that set the image for me. 
I can share it with you if you want. 
For now it configure and perform the step to be ready to make a contribution (I was fedup to do it by hand all the time). But you will still have to checkout your bottom (automatically created) branch. This is in Cozy (menu prepare). Now I have to check it is works when people do not use shared repositories.

astares:
@PharoProject Can you please use a better GitHub name (real name or if you want to stay anonymous then another avatar name) - because using @PharoProject  misleads other users as if your account is the official pharo project itself.

Ducasse:
+ 1 :)

estebanlm:
mmm... that account is created to add some bots and that. 
I guess it was @MarcusDenker commenting and he forgot to switch users?

MarcusDenker:
No, it was not me

kasperosterbye:
This looks good to me. I do not think it can be done better in the current implementation of Bag. Perhaps if the bag kept some more info (like its size), one could do better.

kasperosterbye:
If you want to add it while you are at it, this one is an efficient implementation of the same thing for Set.

```smalltalk
atRandom: aGenerator
	"Answer a random element of the receiver. Uses aGenerator which
    should be kept by the user in a variable and used every time. Use
    this instead of #atRandom for better uniformity of random numbers because 
	only you use the generator. Causes an error if self has no elements."
	| index |

	self emptyCheck.
	index := aGenerator nextInt: array size.
	[ (array at: index) isNil ] whileTrue: [ index := index \\ array size + 1 ].
	^ array at: index.
```

hogoww (approved):
Can we cherry pick the relevant commit?

svenvc:
What do you want to cherry pick ?

There is only one method changed in one file.

Maybe for Pharo 9 ?

hogoww:
Just to avoid having the float failled commit, which was already merged (not that it is THAT important :p)

hogoww:
@Ducasse You merged it in P8, shouldn't it be in p9 ? (I the compared branch wasn't right)

MarcusDenker:
I will add a revert PR

src/Random-Core/Bag.extension.st
```diff
@@ -0,0 +1,18 @@
...
+	 only you use the generator. Causes an error if self has no elements."
+	
+	| rand index |
+	self emptyCheck.
+	rand := aGenerator nextInt: self size.
+	"overwritten to use a faster enumeration"
+	index := 0.
+	self doWithOccurrences: [ :key :count | 
+		index := index + count.
+		rand <= index ifTrue: [ ^ key ] ]
```
nicolas-cellier-aka-nice, line 17: could be done without `index`:

    (rand := rand - count) <= 0 ifTrue: [ ^key ]

kasperosterbye: Very good @nicolas-cellier-aka-nice . I might suggest using `index` instead of `rand` for the variable. It hurts my eyes to reassign `rand`.

MarcusDenker: I will first add the original suggestion as a new PR so we get it in Pharo9, them I open a new issue as the #atRandom:  in Collection should be improved the same.

MarcusDenker: I did a PR: https://github.com/pharo-project/pharo/pull/5663

PR #5446, opened 2020-01-07
Fix rotation, scaling and growing halo events

ericbrandwein:
It was processing the messages to rotate and scale even if the mouseDown
event wasn't sent.

To reproduce:
1. Open a World Menu;
2. Do Alt+Shift+Click on it;
3. Move the mouse over the rotation HaloMorph (the blue one on the lower left);
4. A Debugger should appear with the message "#degrees was sent to nil".

In the case of the scaling and growing, it was simply doing them without the user having to click on them; just moving the mouse over them would trigger the action.

### DIFF
--- a/src/Morphic-Base/HaloMorph.class.st
+++ b/src/Morphic-Base/HaloMorph.class.st
@@ -805,6 +805,7 @@ HaloMorph >> doGrow: evt with: growHandle [
 	"Called while the mouse is down in the grow handle"
 
 	| newExtent extentToUse scale |
+	growingOrRotating ifFalse: [ ^self ].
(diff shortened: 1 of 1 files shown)


Ducasse:
Thanks a lot for your contribution!!!

Ducasse:
There was another fix because the author tried to understand why the code got broken. 
Because it was working before some event changes.

Ducasse:
I was wondering where the growOrRotation was initialized beside in the startGrow.

MarcusDenker (approved):

MarcusDenker:
growOrRotation is initialised in the #initialize method

PR #5454, opened 2020-01-07
Fixing FFI Error in Cairo

tesonep:
Fixing a problem with constant propagation in FFI in Cairo

### DIFF
--- a/src/Athens-Cairo/AthensCairoSurface.class.st
+++ b/src/Athens-Cairo/AthensCairoSurface.class.st
@@ -342,9 +342,12 @@ AthensCairoSurface class >> primImage: aFormat width: aWidth height: aHeight [
 
 { #category : #private }
 AthensCairoSurface class >> primImageFromData: data width: width height: height pitch: stride [
+	
+	"CAIRO_FORMAT_ARGB32 -> 0"
+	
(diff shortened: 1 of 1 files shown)


src/Athens-Cairo/AthensCairoSurface.class.st
```diff
@@ -342,9 +342,12 @@ AthensCairoSurface class >> primImage: aFormat width: aWidth height: aHeight [
 
 { #category : #private }
 AthensCairoSurface class >> primImageFromData: data width: width height: height pitch: stride [
+	
+	"CAIRO_FORMAT_ARGB32 -> 0"
+	
 	^ self ffiCall: #(AthensCairoSurface cairo_image_surface_create_for_data (
 		void *data,
-		CAIRO_FORMAT_ARGB32,
+		0,
```
estebanlm, line 350: why this was an error?

tesonep: Because it is not working by an error of FFI, it is not resolving the constant.

estebanlm: but before it was working... question is, what happened in the middle that this constant solving got broken?. 
I am not asking the question to question your change (sic), but to know what happened :P)

PR #5455, opened 2020-01-07
Made metalinks installable on ast nodes by selecting code when browsing methods in Calypso

StevenCostiou:
I wanted to add helpers to easily install metalinks in ast nodes of methods. 
The idea is to select an ast node in a method in Calypso, then right click and select a metalink instance to install on that node.
Basically it opens a window with a list of existing metalinks that are not breakpoints, watchpoints or execution counters.

However, debugging actions were all accessible from the "source code" menu, while I find it more interesting to have them in a "debugging" menu.

So:
- I removed all debugging actions from the source code menu
- I added a debugging menu with all debugging actions
- I renamed classes that represent menus or groups of debugging commands but that contained  "metalink" in their name, which is illogical as they are debugging actions and not metalink actions. I replaced "metalink" by "debugging" in their name and added new commands for metalinks.
- From the debugging menu, we can now install and uninstall metalinks.


src/Calypso-SystemPlugins-Reflectivity-Browser-Tests/ClyInstallMetaLinkPresenterTest.class.st
```diff
@@ -0,0 +1,142 @@
...
+	metalink uninstall.
+	metalink2 uninstall.
+	super tearDown
+]
+
+{ #category : #tests }
+ClyInstallMetaLinkPresenterTest >> testInstallSelectedMetalink [
+	|presenter list|
+	presenter := ClyMetaLinkInstallationPresenter onNode: self nodeInRealMethod forInstallation:  true.
+	list := presenter instVarNamed: #list.
```
dionisiydk: Can we not use reflection to access variable? 
Otherwise variable references will never show this code.

dionisiydk (changes requested):

src/Calypso-SystemTools-Core/SycOpenDebuggingMenuCommand.extension.st
```diff
@@ -0,0 +1,15 @@
+Extension { #name : #SycOpenDebuggingMenuCommand }
+
+{ #category : #'*Calypso-SystemTools-Core' }
+SycOpenDebuggingMenuCommand class >> methodEditorShortcutActivation [
+	<classAnnotation>
+	
+	^CmdShortcutActivation by: $t meta for: ClySourceCodeContext
```
dionisiydk: It overrides "Source code" menu: SycOpenSourceCodeMenuCommand class>>#methodEditorShortcutActivation

dionisiydk (changes requested):

src/Calypso-SystemPlugins-Reflectivity-Browser-Tests/ClyInstallMetaLinkPresenterTest.class.st
StevenCostiou: I do not understand.
Which variable do you mean?

src/Calypso-SystemTools-Core/SycOpenDebuggingMenuCommand.extension.st
StevenCostiou: Ah yes! Thank you!
But now I need to find an available shortcut :/
Is there a way to know which is free?

StevenCostiou:
Just to show what it does:

Screenshot 2020-01-07 at 21 42 35

Then clicking open a simple UI to select the metalink to install:

Screenshot 2020-01-07 at 21 42 49

src/Calypso-SystemTools-Core/SycOpenDebuggingMenuCommand.extension.st
dionisiydk: It will give you Commander based shortcuts:
```Smalltalk
CmdShortcutActivation registeredInstances groupedBy: #keyCombination
```
And try to select only commands for ClySourceCodeContext.
But there are also some old style text editor shortcuts. I don't know how to see them.

src/Calypso-SystemPlugins-Reflectivity-Browser-Tests/ClyInstallMetaLinkPresenterTest.class.st
dionisiydk: I mean instVarNamed: #list.

dionisiydk: It can be simple:
```Smalltalk
presenter list clickItem: 1
```

dionisiydk: which made me think that #list is not really meaningful name for variable

src/Calypso-SystemTools-Core/SycOpenDebuggingMenuCommand.extension.st
StevenCostiou: We could use meta+d, the only user is ClyDebugTestCommand. 
What is this command for?

src/Calypso-SystemPlugins-Reflectivity-Browser/ClyMetaLinkInstallationPresenter.class.st
```diff
@@ -0,0 +1,123 @@
...
+{ #category : #opening }
+ClyMetaLinkInstallationPresenter class >> openUninstallerOnNode: aSourceNode [
+	(self onNode: aSourceNode forInstallation: false) openWithSpec
+]
+
+{ #category : #accessing }
+ClyMetaLinkInstallationPresenter >> allRelevantMetaLinks [
+	| baseLinks |
+	baseLinks := self isForMetaLinkInstallation
+		             ifTrue: [ MetaLink allInstances ]
```
dionisiydk: This is not nice logic. It needs to be explicit registry of what can be installed instead of #allInstances call.

src/Calypso-SystemPlugins-Reflectivity-Browser-Tests/ClyInstallMetaLinkPresenterTest.class.st
StevenCostiou: Ah, yes, I understood "can we not...?" as "why not using", so it was not clear to me.
This #list is a dumb spec list presenter and is the only widget with a toolbar in the owner presenter.
I did not want to put an accessor just for the test, as it is not meant to be accessed.

src/Calypso-SystemPlugins-Reflectivity-Browser/ClyMetaLinkInstallationPresenter.class.st
```diff
@@ -0,0 +1,123 @@
...
+		             ifFalse: [ sourceNode links ].
+	^ baseLinks reject: [ :link | 
+		  link metaObject == Break or: [ 
+			  { 
+				  Watchpoint.
+				  ExecutionCounter } includes: link metaObject class ] ]
```
dionisiydk, line 60: So the idea here is to remove from the list Break, Watchpoint and ExecutionCounter?

src/Calypso-SystemPlugins-Reflectivity-Browser/ClyMetaLinkInstallationPresenter.class.st
dionisiydk: Imagine somebody will run all metalink tests - there will be a lot of metalinks in memory (GC will not immediately remove them).

StevenCostiou: This is temporary, I should have put a comment.
The idea would be to add an explicit option at the metalink level, so that only the ones that are "installable through the menu" would appear.
But we need to think about how to do that exactly.
I will add a comment, but this should not block: having that now is better than not having that.

src/Calypso-SystemPlugins-Reflectivity-Browser/ClyMetaLinkInstallationPresenter.class.st
StevenCostiou: Yes, because those are not metalinks but tools using them.
We should not interfere.
The best would be an full configuration UI that would allow, for each selected node, to configure all possible options, we are discussing that with @dupriezt.
I imagine something similar (but not the same) as intelliJ or Eclipse breakpoint window.

src/Calypso-SystemPlugins-Reflectivity-Browser-Tests/ClyInstallMetaLinkPresenterTest.class.st
dionisiydk: The problem is that such approach breaks tools:
- code navigation. User will browse this class and analyze #list variable. User will not find these tests using var references.
- refactoring. User will rename the #list variable. Tests will become broken because only variable is renamed, not a #list symbol

And most important it directly breaks incapsulation.   
If you need access to the variable from the outer of class you need an accessor. There is no other way.

src/Calypso-SystemTools-Core/SycOpenDebuggingMenuCommand.extension.st
dionisiydk: I guess it is to debug a test :)

src/Calypso-SystemPlugins-Reflectivity-Browser/ClyMetaLinkInstallationPresenter.class.st
dionisiydk: ok

src/Calypso-SystemTools-Core/SycOpenDebuggingMenuCommand.extension.st
StevenCostiou: But where is it used?
If there was an example I could see if there are possible conflicts.

src/Calypso-SystemPlugins-Reflectivity-Browser/ClyMetaLinkInstallationPresenter.class.st
StevenCostiou: I added a comment.
I will also add an issueif this PR is merged.

src/Calypso-SystemPlugins-Reflectivity-Browser-Tests/ClyInstallMetaLinkPresenterTest.class.st
StevenCostiou: I'm not convinced encapsulation is that important when writing such UI tests, because it exposes my state only for tests and now to users it's part of the API while in reality it should not (and since reflection is there...).

But you are right it breaks all tools.

So I added an accessor, but then as you suggested I also changed the name (to #metalinkList) because now it needs to really mean something from a user point of view.
I realize now it should maybe be named metalinkListPresenter because it is not a list and it should be as an extension in the test package.

src/Calypso-SystemTools-Core/SycOpenDebuggingMenuCommand.extension.st
dionisiydk: It is accessible on selected test method

Ducasse:
BTW steven when you do your PR make sure that it is for Pharo 90. 
You can edit the laucnher (check the files in Documents/pharo/source and create a new entry in the template. I checked this morning and everything should work.

src/Calypso-SystemPlugins-Reflectivity-Browser/ClyRemoveWatchpointCommand.class.st
```diff
@@ -19,7 +19,7 @@ ClyRemoveWatchpointCommand class >> contextMenuOrder [
 ]
 
 { #category : #execution }
-ClyRemoveWatchpointCommand >> metalinkManagerClass [
+ClyRemoveWatchpointCommand >> handlerClass [
```
dionisiydk: What this handler is about? What handlerClass handles?
Original name was about idea that these kind of metalinks are managed by system: system maintains registries of breakpoints, watchpoints and counters. I thought that they could be refactored to have common management code.

dionisiydk: I just think about better name because now we have quite strange abstract method ClyDebuggingCommand>>handlerClass

StevenCostiou: To me they are not metalinks, they are tools using metalinks.
For instance you always call the `Breakpoint` api and not methods from the metalink it holds.
So you it's not managing metalinks. Tools are using them, but then you do not need to know that in the commands, only that you delegate work to those "tools" that "handle" your command action.
To me `handleClass`was more appropriate.
We can find a better name.

dionisiydk: I agree with you. Maybe #debuggingToolClass?
And to solve code duplication in those tools we could introduce DebuggingTool as a common superclass. Now they have a lot of equal code

StevenCostiou: Yes that's good.
I updated.

src/Calypso-SystemPlugins-Reflectivity-Browser-Tests/ClyInstallMetaLinkPresenterTest.class.st
StevenCostiou: Renamed to metalinkListPresenter

src/Calypso-SystemTools-Core/SycOpenDebuggingMenuCommand.extension.st
StevenCostiou: Ok, to me there is no conflict, unless we are supposed to be able to open submenus from the methods list?

dionisiydk (approved):

PR #5456, opened 2020-01-07
Improving metalink printing and visualization in Calypso

StevenCostiou:
Added icon styler to indicate installed metalinks in Calypso.
Clicking in the gutter icon will inspect the set of links installed on the highlighted node.

Screenshot 2020-01-07 at 21 41 24

Enhanced printString behavior.
The link defined as:
```Smalltalk
MetaLink new
	metaObject: [ :o :l :n | o crTrace ];
	selector: #value:value:value:;
	arguments: #(#object #link #node);
	control: #instead;
	options: #(#optionInlineMetaObject #optionInlineCondition #optionWeakAfter).
```

will print as:
`Link instead [ :o :l :n | o crTrace ] value: #object value: #link value: #node `

### DIFF
--- a/src/Reflectivity-Tools/MetaLinkIconStyler.class.st
+++ b/src/Reflectivity-Tools/MetaLinkIconStyler.class.st
@@ -1,0 +1,38 @@
+"
+I'm in charge to style an ast when there are metalink in the ast
+"
+Class {
+	#name : #MetaLinkIconStyler,
+	#superclass : #IconStyler,
+	#category : #'Reflectivity-Tools-Breakpoints'
+}
+
+{ #category : #defaults }
+MetaLinkIconStyler >> highlightColor [
+	^(Color orange alpha: 0.1)
+]
+
+{ #category : #defaults }
+MetaLinkIconStyler >> iconBlock: aNode [
+	^ [ aNode links inspect ]
+]
+
+{ #category : #defaults }
+MetaLinkIconStyler >> iconFor: aNode [
+	^ self iconNamed: #smallDebugIcon 
+]
+
+{ #category : #defaults }
+MetaLinkIconStyler >> iconLabel: aNode [
+	^ 'Metalinks...'
+]
+
+{ #category : #testing }
+MetaLinkIconStyler >> shouldStyleNode: aNode [
+	^ aNode hasLinks and: [ 
+		  aNode links anySatisfy: [ :link | 
+			  (link metaObject == Break or: [ 
+				   { 
+					   Watchpoint.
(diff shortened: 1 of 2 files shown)


src/Reflectivity-Tools/MetaLinkIconStyler.class.st
```diff
@@ -0,0 +1,38 @@
...
+]
+
+{ #category : #testing }
+MetaLinkIconStyler >> shouldStyleNode: aNode [
+	^ aNode hasLinks and: [ 
+		  aNode links anySatisfy: [ :link | 
+			  (link metaObject == Break or: [ 
+				   { 
+					   Watchpoint.
+					   ExecutionCounter } includes: link metaObject class ]) not ] ]
```
dionisiydk, line 37: it needs to be refactored.
If we will introduce kind of DebuggingTool hierarchy for Break, Watchpoint and others we will be able to remove parallel hierarchies for them. 
In this case we could have just one DebuggingIconStyler. It will check known tools installed on the node and it will style "unknown" metalinks in some default way like here.

dionisiydk (approved):
refactoring can be done in separate PR.

dionisiydk:
But you need to retarget PR to Pharo 9

src/Reflectivity-Tools/MetaLinkIconStyler.class.st
StevenCostiou: I agree.
I will add an issue after the PR is merged.

src/Reflectivity/MetaLink.class.st
```diff
@@ -477,16 +477,29 @@ MetaLink >> permaLinkFor: aClassOrObject option: option instanceSpecific: instan
 	^ permaLink
 ]
 
-{ #category : #'link api' }
+{ #category : #printing }
 MetaLink >> printString [
```
dionisiydk, line 481: I just noticed that it is #printString method instead of #printOn: .  Normally classes implement #printOn: with a stream as argument.

dionisiydk (changes requested):

src/Reflectivity/MetaLink.class.st
StevenCostiou: Yes, I do not know why it was done like this @MarcusDenker 
It is not part of the PR (I only changed the method and its protocol).
Can you add an issue?

MarcusDenker: yes, should be fixed.

MarcusDenker (approved):

astares:
Nice addition!

Issue #5459, opened 2020-01-08
does addSelfReturn needs to return a VariableNode


hogoww:
@MarcusDenker 
Currently we have: 
```
addSelfReturn
	| node |
	self lastIsReturn
		ifTrue: [ ^ self statements last ].
	node := RBReturnNode value: (RBVariableNode named: 'self').
	^ self addNode: node
```
Could we replace the VariableNode by a SelfNode directly, or may it mess up with the Semantic Analysis?

MarcusDenker:
Should be save to change that, yes. Should not have an impact on Semantic Analysis (it would be done again there but that is no problem)

PR #5460, opened 2020-01-08
[improvement] now returns a selfNode directly rather than wait for th…

hogoww:
…e semantic analysis

follow up of #5459

Edit: Also renamed RBReturnNode >> value: aValueNode argument to anExpression.
(for future research)

might need to move to P9, i don't have anything ready for a quick fix on this, sorry  >_<


hogoww:
Forgot that I had this PR open. should have done a different one, but it's small enough to pass anyway.

@MarcusDenker I don't know if anExpression is the best name as replacement (see second commit), But I think aValueNode is missleading for people trying to learn the API 
(although I'm pretty sure that when doing AST manipulation, you already know the syntax, and therefore are aware that this is wrong...)

MarcusDenker (approved):

PR #5463, opened 2020-01-10
5097-Creating-a-method-from-debugger-goes-into-infinite-loop-when-typing-something-in-the-class-list

### LINKED ISSUE #5097: Creating a method from debugger goes into infinite loop when typing something in the class list
Not sure this belong here, might be for spec2, didn't find a similar issue.

When creating a method from the 'create' button of the debugger, if you enter a string in the 'research' bar under the class names, it goes into an infinite loop. Sometimes breakable ( seems to be with a small number of character)

(tested on the last p8 version of the launcher)

estebanlm:
kind of a hack, but I do not find a better way to do it.

Fixes #5097


estebanlm:
meh, I don't understand why it fails :)

tesonep (approved):

PR #5465, opened 2020-01-10
5464-Bug-Slot-template-add-variable-without-hash-Crash-on-accept

### LINKED ISSUE #5464: Bug Slot template: add variable without #hash: Crash on accept
when accepting 

Object subclass: #NameOfSubclass
	slots: {hello}
	classVariables: {}
	package: 'Athens-Cairo-Tests'

the system asks if it is ok to add the unknown variable hello. But that is the wrong question to ask
(and  then is done in class UndefinedObject -> not good).

Solution: The new class definition parser can check this for real.

Intermediate solution:

in compileANewClassFrom:notifying:startingFrom: wrap the compilation in a execution handler like that:

evaluate] on: OCUndeclaredVariableWarning do: [ self inform: 'Undeclared Variable in Class Definition' . ^nil ].

MarcusDenker:
fixes #5464


MarcusDenker:
Why my change picks up Hilaire's change I have no idea...

MarcusDenker:
Ah... it should be done for pharo9. I close this

PR #5468, opened 2020-01-11
Unification of simple form of "isAbstract" method.

kamk:
Among three most common implementations:
1. self = SomeClass
2. self == SomeClass
3. self name = #SomeClass

second one is the most fastest.

See original FogBugz [case #22771](https://pharo.fogbugz.com/f/cases/22771/isAbstract-for-classes-should-be-unified).

### DIFF
--- a/src/AST-Core-Tests/RBParseTreeTest.class.st
+++ b/src/AST-Core-Tests/RBParseTreeTest.class.st
@@ -7,7 +7,7 @@ Class {
 { #category : #testing }
 RBParseTreeTest class >> isAbstract [ 
 
-	^self name = #RBParseTreeTest
+	^self == RBParseTreeTest
 ]
 
 { #category : #helpers }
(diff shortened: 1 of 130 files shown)


Ducasse:
Thanks for your submission. 
Did you check if class override = because in Object = is defined as ==.
Another point could you issue the PR for P9 because P8 is in beta and we do not integrate changes except for the urgent and important bug fixes.

kamk:
Okay, I will create new PR for Pharo-9.0 branch.
I didn't find any override of = at class side.

Ducasse:
Ok so before doing it it would be nice to really measure be difference between == and = because using = gives another method that we can redefine in case we want to play with spies and other tooling. While we cannot redefine easily ==.

kamk:
I see. Maybe it should be noticed in ProtoObject>>#== comment that this primitive should not be used directly (except in Object's "=" comparison method).

Ducasse:
This is not systematic, sometimes you overrides = and still need to use ==.

PR #5469, opened 2020-01-11
Unification of simple form of "isAbstract" method.

kamk:
Among three most common implementations:
1. self = SomeClass
2. self == SomeClass
3. self name = #SomeClass

second one is the most fastest.

See original FogBugz [case #22771](https://pharo.fogbugz.com/f/cases/22771/isAbstract-for-classes-should-be-unified).

### DIFF
--- a/src/AST-Core-Tests/RBParseTreeTest.class.st
+++ b/src/AST-Core-Tests/RBParseTreeTest.class.st
@@ -7,7 +7,7 @@ Class {
 { #category : #testing }
 RBParseTreeTest class >> isAbstract [ 
 
-	^self name = #RBParseTreeTest
+	^self == RBParseTreeTest
 ]
 
 { #category : #helpers }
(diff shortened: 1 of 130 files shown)


dionisiydk:
We had discussion in mailing list in past: [Pharo-dev] how implement isAbstract?. 
I did not remember if we had any agreement. The discussion seems was moved to another question why we need isAbstract.

Ducasse:
Thanks denis. I think that in addition it can also give an indirection in case of proxies.

Ducasse:
I think that we should close all the fogbugz because they are old now.

kamk:
I see, I've found the [discussion here](http://forum.world.st/how-implement-isAbstract-td5097729.html#a5097747).

Ducasse:
Thanks for putting the reference. I reread all the thread and I like the remark of guille.
I like that people mark the class as abstract while relying on subclassResponsibility.

PR #5470, opened 2020-01-11
Fix Exception #pass semantics

dionisiydk:
Exception should reset #handlerContext during #pass operation because #pass  should be equivalent to no handler scenario. 
It solves the bug with UnhandledError:
```Smalltalk
 [ [ MyTestError signal ]
             on: UnhandledError
             do: [ :e | self halt ] ]
         on: MyTestError
         do: [ :z | z pass ].
```
The fix is influenced by Cuis implementation but it avoids keeping all processed handlers in collection. 
In addition #resignalAs: and couple of Exception tests are added from Cuis 
(#resignalAs: should unwinds the stack up to signaler context according to ANSI)

### DIFF
--- a/src/Kernel-Tests/ExceptionTest.class.st
+++ b/src/Kernel-Tests/ExceptionTest.class.st
@@ -392,6 +392,7 @@ ExceptionTest >> testUnhandledErrorWhenNoHandlers [
 	executed := false.
 	semaphore := Semaphore new.
 	[[ 1/0 ] on: UnhandledError do: [ :e | executed := true ]. semaphore signal] fork.
+	"we should not wait forever in case of wrong behaviour (general test timeout is disabled in #runCaseManaged)"
 	semaphore wait: 2 seconds.
 	self assert: executed
 ]
--- a/src/Kernel-Tests/ExceptionTest.class.st
+++ b/src/Kernel-Tests/ExceptionTest.class.st
@@ -379,9 +379,10 @@ ExceptionTest >> testUnhandledErrorWhenHandlerPassesOriginalException [
 	semaphore := Semaphore new.
 	[ [ 
 			[ 1/0 ] on: UnhandledError do: [ :e | executed := true ]
-		] on: ZeroDivide do: [:z | z pass]
-		. semaphore signal] fork.
-	"we should not wait forever in case of wrong behaviour (general test timeout is disabled in #runCaseManaged)"
+		] on: ZeroDivide do: [:z | z pass].
+		semaphore signal] fork. "fork allows to avoid any handlers from the SUnit"
+	"we should not wait forever in case of wrong behaviour 
+	(general test timeout is disabled in #runCaseManaged)"
 	semaphore wait: 2 seconds.
 	self assert: executed
 ]
@@ -391,8 +392,27 @@ ExceptionTest >> testUnhandledErrorWhenNoHandlers [
 	| executed semaphore |
 	executed := false.
 	semaphore := Semaphore new.
-	[[ 1/0 ] on: UnhandledError do: [ :e | executed := true ]. semaphore signal] fork.
-	"we should not wait forever in case of wrong behaviour (general test timeout is disabled in #runCaseManaged)"
+	[
+		[ 1/0 ] on: UnhandledError do: [ :e | executed := true ]. 
+		semaphore signal] fork. "fork allows to avoid any handlers from the SUnit"
+	"we should not wait forever in case of wrong behaviour 
+	(general test timeout is disabled in #runCaseManaged)"
+	semaphore wait: 2 seconds.
+	self assert: executed
+]
+
+{ #category : #'tests - handling' }
+ExceptionTest >> testUnhandledErrorWhenTwoHandlersPassOriginalException [
+	| executed semaphore |
+	executed := false.
+	semaphore := Semaphore new.
+	[ [ [  
+				[ 1/0 ] on: UnhandledError do: [ :e | executed := true ]
+			] on: ZeroDivide do: [:z | z pass]
+		] on: ZeroDivide do: [:z | z pass].
+		semaphore signal] fork. "fork allows to avoid any handlers from the SUnit"
+	"we should not wait forever in case of wrong behaviour 
+	(general test timeout is disabled in #runCaseManaged)"
 	semaphore wait: 2 seconds.
 	self assert: executed
 ]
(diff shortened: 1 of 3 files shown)


dionisiydk:
Fixes #5472

dionisiydk:
I am not sure about Pharo 8. But it fixes quite critical issue - the root problem with debugger (some of scenarios).
From the other side we lived with this from Pharo 4. So please decide (@estebanlm  ?). 

P.S. I am going to do another changes related to exceptions (and therefore to debugger).

Ducasse:
Super thanks Denis. My impression is that we keep it for P9 and we can always backport it in P8 once we see and understand the effect. We should release P8.

noha:
@dionisiydk Why is pass supposed to be the no handler case? Do you have any pointers to that? I always thought it means pass onto the next handler

dionisiydk:
@noha here is a comment from @martinmcclure (I had a chat with him about this problem):

> A pass in the outermost handler (no handler to pass *to*) is clearly
> defined by ANSI to behave as though the handler was not there at all. So
> the MyTestError should be sent #defaultAction (and that happens) but
> when the defaultAction signals UnhandledError, the search fails to find
> the handler.
> 
> ANSI says that #defaultAction is executed in the signaling environment,
> so the handler for UnhandledError *should* be found.

src/Kernel-Tests/ExceptionTest.class.st
```diff
@@ -320,12 +352,45 @@ ExceptionTest >> testSimpleResume [
...
+	| executed semaphore |
+	executed := false.
+	semaphore := Semaphore new.
+	[ [ 
+			[ 1/0 ] on: UnhandledError do: [ :e | executed := true ]
+		] on: ZeroDivide do: [:z | z pass]
+		. semaphore signal] fork.
+	semaphore wait: 2 seconds.
+	self assert: executed
+]
```
dupriezt: I'm looking at the explicit 2 second timeout for the semaphore.
If I understand correctly, this test has no default timeout because it's not running in the test execution environment (see change to `runCaseManaged` method). So you have to write an explicit timeout in the test to prevent it from hanging.
If I'm right, I think there should be a comment explaining why there is an explicit timeout on the semaphore. Otherwise, future readers could mistake it as a code smell.

src/Kernel-Tests/ExceptionTest.class.st
```diff
@@ -320,12 +352,45 @@ ExceptionTest >> testSimpleResume [
...
+
+{ #category : #'tests - handling' }
+ExceptionTest >> testUnhandledErrorWhenNoHandlers [
+	| executed semaphore |
+	executed := false.
+	semaphore := Semaphore new.
+	[[ 1/0 ] on: UnhandledError do: [ :e | executed := true ]. semaphore signal] fork.
+	semaphore wait: 2 seconds.
+	self assert: executed
```
dupriezt: I'm looking at the explicit 2 second timeout for the semaphore.
If I understand correctly, this test has no default timeout because it's not running in the test execution environment (see change to `runCaseManaged` method). So you have to write an explicit timeout in the test to prevent it from hanging.
If I'm right, I think there should be a comment explaining why there is an explicit timeout on the semaphore. Otherwise, future readers could mistake it as a code smell.

src/Kernel-Tests/ExceptionTest.class.st
```diff
@@ -320,12 +352,45 @@ ExceptionTest >> testSimpleResume [
...
+{ #category : #'tests - handling' }
+ExceptionTest >> testUnhandledErrorWhenHandlerPassesOriginalException [
+	| executed semaphore |
+	executed := false.
+	semaphore := Semaphore new.
+	[ [ 
+			[ 1/0 ] on: UnhandledError do: [ :e | executed := true ]
+		] on: ZeroDivide do: [:z | z pass]
+		. semaphore signal] fork.
+	semaphore wait: 2 seconds.
```
dionisiydk, line 386:
```suggestion
	"we should not wait forever in case of wrong behaviour (general test timeout is disabled in #runCaseManaged)"
	semaphore wait: 2 seconds.
```

src/Kernel-Tests/ExceptionTest.class.st
```diff
@@ -320,12 +352,45 @@ ExceptionTest >> testSimpleResume [
...
+	self assert: executed
+]
+
+{ #category : #'tests - handling' }
+ExceptionTest >> testUnhandledErrorWhenNoHandlers [
+	| executed semaphore |
+	executed := false.
+	semaphore := Semaphore new.
+	[[ 1/0 ] on: UnhandledError do: [ :e | executed := true ]. semaphore signal] fork.
+	semaphore wait: 2 seconds.
```
dionisiydk, line 416:
```suggestion
	"we should not wait forever in case of wrong behaviour (general test timeout is disabled in #runCaseManaged)"
	semaphore wait: 2 seconds.
```

dionisiydk (commented):
is this comment fine?

dupriezt:
> is this comment fine?

Yes

dupriezt:
Something that could be worth testing is whether a succession of 2 pass works.
I tried to load your PR in a Pharo 8, but I got an emergency evaluator :( (it's probably because of some loading issue, not your PR itself).
I don't have more time today unfortunately.

dionisiydk:
> Something that could be worth testing is whether a succession of 2 pass works.
> I tried to load your PR in a Pharo 8, but I got an emergency evaluator :( (it's probably because of some loading issue, not your PR itself).
> I don't have more time today unfortunately.

Can you show example here? (I do not understand what you mean)

src/Kernel-Tests/ExceptionTest.class.st
```diff
@@ -320,12 +352,67 @@ ExceptionTest >> testSimpleResume [
...
+		[ 1/0 ] on: UnhandledError do: [ :e | executed := true ]. 
+		semaphore signal] fork. "fork allows to avoid any handlers from the SUnit"
+	"we should not wait forever in case of wrong behaviour 
+	(general test timeout is disabled in #runCaseManaged)"
+	semaphore wait: 2 seconds.
+	self assert: executed
+]
+
+{ #category : #'tests - handling' }
+ExceptionTest >> testUnhandledErrorWhenTwoHandlersPassOriginalException [
```
dionisiydk, line 405: @dupriezt Here is the test for two handlers passing an error. Hopefully it is what you was asking

dupriezt: Yes, that was it.

dionisiydk:
it seems I addressed all feedback. Can it be merged?

PR #5471, opened 2020-01-11
Fixes pharo-project#3430: ifError should accept error instance instead of its properties

kamk:
ifError: block accepts anError instance as a first argument


src/Kernel/BlockClosure.class.st
```diff
@@ -336,13 +336,13 @@ BlockClosure >> ifError: errorHandlerBlock [
...
 		[1 whatsUpDoc] ifError: [:err :rcvr | 'huh?'].
 		[1 / 0] ifError: [:err :rcvr |
-			'ZeroDivide' = err
+			ZeroDivide = err class
 				ifTrue: [Float infinity]
 				ifFalse: [self error: err]]
 "
 
 	^ self on: Error do: [:ex |
-		errorHandlerBlock cull: ex description cull: ex receiver]
```
dionisiydk, line 345: I would also remove second parameter. It is incompatible change anyway. So it should not add more problems.
(my idea was to make ifError consistent to on:do:)

dionisiydk (approved):

src/Kernel/BlockClosure.class.st
kamk: Attempt in the commit kamk/pharo@415dfdefea0d5da0a49f7da62dff382687248ef7.

dionisiydk (approved):

dionisiydk:
benchmarks step is completely broken. Can we disable it? (@MarcusDenker ?)

tesonep:
@dionisiydk I have disabled

dionisiydk:
Great, thanks

пн, 13 янв. 2020 г., 16:23 Pablo Tesone <notifications@github.com>:

dionisiydk:
The failure looks unrelated. CI needs to be run again

MarcusDenker (approved):
I like this change. But as it is not backward compatible, I will wait for more comments before merging

Bajger:
Hi! There are about 33 senders of #ifError: message in the system. Do you expect that this message is widely used in other user's code? Could we use deprecation mechanism, when #ifError: is called with old-way arguments? (Method name #ifError: is same, but argument block would need to be tested appropriately and deprecation called if block argument count is more than 1,and if first (and only) parameter is error instance?)

kamk:
I guess it would require some handler's code analysis if first argument is treated as some kind of String. I have added deprecation notice if more than one argument is passed.

guillep:
> I guess it would require some handler's code analysis if first argument is treated as some kind of String. I have added deprecation notice if more than one argument is passed.

The more """organized""" way to do this would be:
 - add a new `ifErrorDo:` with the new semantics.
 - deprecate `ifError:` to redirect people to the new one

Then eventually we could remove the old one.
What do you think?
That would at least remove part of the "fear component" that incompatibility brings :)

guillep:
News on this front? I think the deprecation is important (and easy!) for this case...

guillep:
I see no activity here.
I'll close it. Please reopen if necessary.

Bajger:
Hi Guille, I'll try to dispatch wrong argument class to new #onErrorDo: method (one would will be on Object with deprecation message, other implementor would be on Error class). We could still preserve original #ifError: method on BlockClosure. I'll add adequate tests. (wrong argument count and wrong argument type). How can I reopen issue again? Only someone from core devs can do that?

guillep:
Thanks for the follow up! I'll reopen it.

> We could still preserve original #ifError: method on BlockClosure

I'm ok with deprecating `ifError:` too ^^. My point is that many external libraries may use this method, and we are just breaking them when it when we could make it much more nicely...

guillep:
@Bajger I see the PR was issued by another user. Give it a try, but maybe it's just easier to do a new PR, as you feel it better.

Bajger:
Can I just clone Kamil's repository with his commit in PR? I guess it's doable, right? But I would need to commit to his repo & PR...

kamk:
Uff, I screwed up something when trying to merge upstream of current Pharo 9.0 so I hopefully have reversed it. Now I have ready `ifErrorDo:` logic so should I commit and push it to my repo?

guillep:
You should be able to do it with some easy git / iceberg manipulations.
For example, if you go up and click on the PRd branch you can get the repository and branch that were issued.

image

Then you may:
 - add that repository url as remote
 - checkout the branch
 - push it to your repository

If you want to do it with iceberg, you need to do some additional manipulations (that are necessary just for contributing pharo, and not for other projects):
 - add that repository url as remote
 - checkout the branch, but selecting in the checkout preview to "Not load any packages" (there is a drop down)
 - merge latest Pharo90 branch into the branch (proceed in the warning)
 - then you will have a branch that has these changes + merged with latest pharo changes.

Bajger:
Hi Kamil, proposed solution from Guille of using new #ifErrorDo: and deprecating #ifError: seems to be more convenient, so I agree. Tests just needs to be updated accordingly (even using testing raising deprecation, if ifError: is sent). You can do new PR, if you have issues with continuing on existing PR.

guillep:
Wow thank you both for the discussion and the effort! I'm going to pass on the PR changes fast and let's wait for the CI :)

Bajger:
Thank you Kamil!

src/Kernel/BlockClosure.class.st
```diff
@@ -341,10 +341,27 @@ BlockClosure >> ifError: errorHandlerBlock [
...
 
+{ #category : #evaluating }
+BlockClosure >> ifErrorDo: errorHandlerBlock [
+	"Evaluate the block represented by the receiver, and normally return it's value.  If an error occurs, the errorHandlerBlock is evaluated, and it's value is instead returned.  The errorHandlerBlock must accept zero or one parameter (the error object)."
+	"Examples:
+		[1 whatsUpDoc] ifErrorDo: [:err | 'huh?'].
+		[1 / 0] ifErrorDo: [:err |
+			ZeroDivide = err class
+				ifTrue: [Float infinity]
+				ifFalse: [self error: err]]
```
Bajger, line 358: Eventually, there could be executable examples created from this comment by using #>>> method.

src/Morphic-Base/Form.extension.st
```diff
@@ -12,13 +12,13 @@ Form >> asMorph [
...
-	^ FloodFillTolerance ifNil: [FloodFillTolerance := 0.0]
+Form >> floodFillTolerance [
+	^ self class floodFillTolerance
 ]
 
 { #category : #'*Morphic-Base' }
-Form >> floodFillTolerance [
-	^ self class floodFillTolerance
+Form class >> floodFillTolerance [
+	^ FloodFillTolerance ifNil: [FloodFillTolerance := 0.0]
```
Bajger, line 21: Are these changes relevant to #ifError: refactoring? Or it comes from different commit...

Bajger (approved):
There are some changes that probably not relevant, but they might be from different commit. But it is mergeable, so problem shouldn't occur, right?

src/Morphic-Base/Form.extension.st
kamk: No, it's mystery to me how did this change get there.

guillep: It's just the order that got changed. Probably just a change in the tonal format. I think its no biggy :)

guillep (approved):
Good! Thanks both!

astares:
Why "if" instead of #onErrorDo: ? It dispatches to #on:do: so we should keep the "on"

guillep:
I think this is just because historically the deprecated method was called `ifError:`. I'm not against renaming it as `onErrorDo:`

kamk:
It could also be defined as call to `ifErrorDo:` if this is kind of redudancy is okay.

PR #5473, opened 2020-01-12
5434-Default-values-in-settings-should-not-be-stored

### LINKED ISSUE #5434: Default values in settings should not be stored
When saving any modified value in the settings, all the default values are changed to the one in the image. 
This is really bad specially for the localDirectory settings. 
For example: Epicea stores all the ombu files in the same directory, does not matter if they are from different images, making it impossible to use.

The same, at least, with Iceberg and the playground cache.

tesonep:
Fixes #5434 

- Adding default values to all settings.
- So, we can take advantage of not saving the settings if they have a default value.


Ducasse:
Thanks Pablo. It means that we should have a way to check that future settings should have a default value. 

This is strange there are 20 broken tests. (no time to really try to understand).

tesonep:
I have added a test to validate it and some missing ones!

PR #5482, opened 2020-01-14
5476-Code-completion-failure-in-in-place-editing

MarcusDenker:
version for Pharo9 first (to be backported)


Ducasse:
Marcus this test does not pass CompletionContextTest

MarcusDenker:
No problems when I run it here...

MarcusDenker:
For the second run the test is green

Issue #5481, opened 2020-01-14
Metacello logs now prints lots of new lines


jecisc:
Since yesterday's fix, I can now see the logs from Metacello again but between each line I get a new line containing "[0m".

See screenshot

![Screenshot 2020-01-14 at 10 57 14](https://user-images.githubusercontent.com/9519971/72333783-ad737d00-36bc-11ea-86b3-77241b0691d4.png)

jecisc:
Done

PR #5484, opened 2020-01-14
5481-Metacello-logs-now-prints-lots-of-new-lines

tesonep:
Improving the logging of system notifications.
Removing extra escape characters and cr.
Also not closing Stderr


tesonep:
Fixes #5481

PR #5487, opened 2020-01-15
switch theme to white (Pharo 8.0 is a white themed release)

estebanlm:


astares:
To explain more deeply the issue:

Pharo releases are alternating regarding the UI theme because there is no common sense which one people prefer. Therefore one time we use a dark theme - one time the light theme. 

For Pharo 7 we used the dark theme - so for Pharo 8 it is time to use the light theme again.

guillep:
Hi guys, Pharo 7 **was** white themed, Pharo 6 was dark themed,
 => Pharo 8 became dark themed at the beginning of the cycle to be able to debug possible theme glitches.
 => Pharo 9 should be white themed...

Is this change good?

Ducasse:
We can keep it black we will redo the videos for the mooc in white anyway.

estebanlm:
in fact Guille is right. P8 was dark, not white.
Should we revert?

Ducasse:
Go ahead. 
I would have been nice to a different one.

PR #5488, opened 2020-01-15
5486-Updating-values-in-a-dictionary-is-cumbersome-and-inefficient

### LINKED ISSUE #5486: Updating values in a dictionary is cumbersome and inefficient
As an example, the following code from `Bag`.

```smalltalk
add: newObject withOccurrences: anInteger 
	"Add newObject anInteger times to the receiver. Answer newObject."

	contents at: newObject put: (contents at: newObject ifAbsent: [0]) + anInteger.
	^ newObject
```

It should not be necessary to lookup the key twice, and while it is a one liner it is an idiom rather than an abstraction. I will propose we include a method in` Dictionary`: 

```smalltalk
aDict at: key update: [:val | val + 7] initial: [0].
```

kasperosterbye:
Fixes #5486, update methods for Dictionary


kasperosterbye:
Skipping this for Pharo 8.

kasperosterbye:
I had confused myself regarding issue and PR in connection with the transition from 8.0 to 9.0. The PR still works I think.

kasperosterbye:
As always I feel really alienated by:

image

and then:
image

I mean, why does a change to `Dictionary` fail to build on unix and not on windows or mac???

src/Collections-Unordered/Dictionary.class.st
```diff
@@ -333,6 +333,21 @@ Dictionary >> at: key put: anObject [
...
 
+{ #category : #accessing }
+Dictionary >> at: key update: updateBlock [ 
+	"I am used to update the value at a given key, or if the key does not exist, to throw an error"
+	self at: key update: updateBlock initial: [ self errorKeyNotFound: key ].
+]
+
+{ #category : #accessing }
+Dictionary >> at: key update: updateBlock initial: initObjectOrValue [
+	| index |
```
juliendelplanque, line 349: Maybe you should state here that the fact that you use #findElementOrNil: instead of the

```Smalltalk
self at: key put: (self at: key)
```

is because this method is optimised.

juliendelplanque (changes requested):
See my comment, else it is good.

src/Collections-Unordered/Dictionary.class.st
kasperosterbye: Comments is always a good thing. I will see if I am able to get it done. Git and the transition to 9.0 has given me all kinds of problems (most likely because I do not understand fully what goes on).

juliendelplanque: Maybe for a comment you can do it from GitHub web app directly?
But you need to be careful. :-)

Ducasse:
Moved to P9

MarcusDenker (dismissed):
ok

MarcusDenker:
There is a test failing that checks that Dictionary has the same API as SmallDictionary and OrderedDictionary

#testDictionaryPublicProtocolCompatibility

kasperosterbye:
I'll take a look - sometime later today I think.

kasperosterbye:
Hmm, this time it is one of those errors I am not able to understand.

MarcusDenker:
it looks like a random crash to me

MarcusDenker:
But the problem is that your PR seems to be faulty: it is *huge* and seems to undo lots of things we did in Pharo9

kasperosterbye:
OK - skip it. I will try to make an other one at some other time. I am not sure exactly what happened because I it was originally for pharo 8, then it is for 9 now. I am not sure how to handle that one anyways.

kasperosterbye:
It should not be more than 6 methods in Dictionary, SmallDictionary and OrderedDictionary, and similar on the test side. Let me try to make it from scratch again (aka. copy methods into a cleanly cloned Pharo9 branch).

Issue #5496, opened 2020-01-15
RB* and GR* should be renamed as Re* ... Rule


egvillalobos:
The rules migrated from RB and GR to Renraku should be renamed as Re rules so that there is no confusion and regularity is a plus.

PR #5497, opened 2020-01-15
RB* and GR* should be renamed as Re* ... Rule - fixes: #5496

egvillalobos:
| pkgPrefix newClassPrefix env model |
pkgPrefix := 'General'.
newClassPrefix := 'Re'.
env := RBBrowserEnvironment new 
			forPackageNames: (RPackage organizer packageNames select: [ : pkgName | (pkgName beginsWith: pkgPrefix) ]).
model := (RBClassModelFactory rbNamespace onEnvironment: env) name: 'MyModel'; yourself.
RBClassRegexRefactoring new
	model: model;
	renameClasses; 					"Here I just want a rename no copy"
	replace: '^GR(.*)$' with: newClassPrefix,'$1';
  execute.


Ducasse:
This is strange, Esteban we will have to check because I do not get why this refactoring broke so many tests. There should be one simple mistake.

Ducasse:
Ok we should have a look at the problem.

estebanlm:
please put a real name to the PR, otherwise is hard to keep tracking

Ducasse:
We have this test failing. 
 testRule – Unix32.Renraku.Test.ReCritiqueTest
Can you have a look? Thanks

Ducasse:
We will get there...

Ducasse:
Tests are ok (we should check latter the setting ones). 
Thanks for your support.

PR #5502, opened 2020-01-15
Allow concatenation of CompositeAnnotationContext subclasses too

### LINKED ISSUE #5501: Allow concatenation of CompositeAnnotationContext subclasses too
The #, concatenation is implemented with a hard reference on CompositeAnnotationContext

```Smalltalk
, anAnnotationContext
	^ CompositeAnnotationContext with: (parts copyWith: anAnnotationContext asAnnotationContext)
```
which is also shown as a lint problem.

So concatenation is not possible when one (for whatever reason) needs to subclasses this class. We should better use "self class" here.

```Smalltalk
, anAnnotationContext
	^ self class with: (parts copyWith: anAnnotationContext asAnnotationContext)
```

astares:
Fix #5501

### DIFF
--- a/src/ClassAnnotation/CompositeAnnotationContext.class.st
+++ b/src/ClassAnnotation/CompositeAnnotationContext.class.st
@@ -31,7 +31,7 @@ CompositeAnnotationContext class >> with: contextCollection [
 
 { #category : #converting }
 CompositeAnnotationContext >> , anAnnotationContext [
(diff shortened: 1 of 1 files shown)


astares:
@dionisiydk if possible please review

dionisiydk (approved):

astares:
Thanks

MarcusDenker (approved):

PR #5510, opened 2020-01-15
Unreadable "red" marking color collide with "red" unimplemented messages text

### LINKED ISSUE #5508: Unreadable "red" marking color collide with "red" unimplemented messages text
Evaluate

```Smalltalk
(DebugAction>>#asMenuRegistrationIn:) browse
```
in Pharo 9 to see the following situation in the browser.

The red color is too strong to see the text of the unimplemented method (displayed in red as well)

image

Maybe we should use more a pink like color for the text background

astares:
Use pink instead of red for background marking 

Fixes #5508

### DIFF
--- a/src/Polymorph-Widgets/PharoDarkTheme.class.st
+++ b/src/Polymorph-Widgets/PharoDarkTheme.class.st
@@ -270,6 +270,12 @@ PharoDarkTheme >> initialize [
 	self settings preferGradientFill: false.
 ]
 
+{ #category : #testing }
+PharoDarkTheme >> isLightTheme [
+
+	^false
+]
+
 { #category : #'accessing colors' }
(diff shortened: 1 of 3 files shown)


astares:
image

astares:
in dark theme it looks like before with the red color:

image

Ducasse:
To me the logic of the changes is not good. 
It introduces a dependency that is disguised. 
The solution is to add color in the ReAbstractCritique and that the theme set such color.
ReAbstractCritique should NOT query state on the theme. 

This is the architecture of the Settings:
- any component should be configurable and use its own personal data (color for example)
- any component that want to expose it configuration should describe it using setting 
- and the setting infra will be able to change them.

I will have spend my life fighting with the impact of the proposed changes. 
So we should NOT integrate such logic. 

A critique should not depend on Smalltalk ui theme!!!

PR #5535, opened 2020-01-21
Switch to the light theme

pavel-krivanek:
Following the theme switching policy (to be sure that none of the default themes is abandoned), let's switch to the ligth theme in Pharo 9.0.
Switching  to the dark theme (if it is your favourite one) is literally SINGLE click operation.


Ducasse:
Yes Excellent idea!

PR #5540, opened 2020-01-21
5478-Inspector-on-a-ByteSymbol-should-get-the-senders-of-this-symbol-on-a-lazy-way

### LINKED ISSUE #5478: Inspector on a ByteSymbol should get the senders of this symbol on a lazy way
One doesn't want to look for all the senders of it in a huge image...

image

The main issue is that the tab contents is searched before the opening of the window.

And even if the tab is loaded lazily, a confirmation message for the loading will be welcomed.

Ducasse:
Fixes: #5478 removed unnecessaryt tags that can cost a lot and blow up your debugging experience. 

senders/implementors in symbols
allInstances in classes


Ducasse:
I'm thinking that the senders and implementors should be extensions a package that we load when we want to navigate the system.

MarcusDenker:
I found a way to seriously speedup the Selector senders and implementors

I will close this and do two new PRs: one to remove the allInstances tab, one to add the speedup (we can then see if it is fast enough)

Issue #5542, opened 2020-01-22
Unnecessary double checks of infinity


SergeStinckwich:
I'm having a look how floats are printed.
Why we check two times if a Float is infinite ?
A first time in:

```Smalltalk
Float>>printOn: stream base: base
...
self isInfinite
ifTrue: [
stream nextPutAll: 'Float infinity'.
^ self sign = -1 ifTrue: [ stream nextPutAll: ' negated' ] ].
...
```

and a second time in:

```Smalltalk
Float>>absPrintExactlyOn: aStream base: base
...
self isInfinite ifTrue: [
aStream nextPutAll: 'Float infinity'.
^ self].
...
```

the same in ``Float>>absPrintInexactlyOn: aStream base: base``

```Smalltalk
...
self isInfinite ifTrue: [aStream nextPutAll: 'Float infinity'. ^ self].
...
```
Apparently the last two methods are never called when number is infinite.

I remove the second checks and apparently nothing breaks (all number tests are still green).

MarcusDenker:
merged

PR #5546, opened 2020-01-22
Fix issue #5542: Unnecessary double checks of infinity

SergeStinckwich:
Remove unnecessary double checks of infinity in Float class

### DIFF
--- a/src/Kernel/Float.class.st
+++ b/src/Kernel/Float.class.st
@@ -297,9 +297,6 @@ Float >> absPrintExactlyOn: aStream base: base [
 	by using exact integer arithmetic."
 
 	| fBase significand exp baseExpEstimate be be1 r s mPlus mMinus scale roundingIncludesLimits d tc1 tc2 fixedFormat decPointCount |
-	self isInfinite ifTrue: [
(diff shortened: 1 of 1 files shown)


MarcusDenker:
I close and reopen to run against the latest code base of Pharo9

MarcusDenker (approved):
i checked the code. The only reason why this might be problematic would be if the methods that now have the infinite check removed would be called directly. But it seems that this makes no sense.

So I think it is safe to merge

SergeStinckwich:
Yes you are right.

In the comment, this is written: "Print my value on a stream in the given base.  Assumes that my value is strictly positive; negative numbers, zero, and NaNs have already been handled elsewhere."

Why we would consider infinite numbers differently?

Maybe comments should be update like : "Print my value on a stream in the given base.  Assumes that my value is strictly positive; negative numbers, zero, NaNs and infinite numbers have already been handled elsewhere."

MarcusDenker:
I will merge an open a new issue to improve the comment!

MarcusDenker:
I added https://github.com/pharo-project/pharo/issues/5723 to update the comment

PR #5552, opened 2020-01-22
5549-String-assumes-specific-class-hierarchy

### LINKED ISSUE #5549: String assumes specific class hierarchy
I want to make an other subclass of String (slices). In that process I have found a number of methods in String which use `self class` instead of `self species` preventing me from making a seamless implementation of string slices.

Also, the method alike make assumptions on class hierarchy through assumptions on which superclass to hit on `super at:`. 

Finally, the `asByteArray` is implemented for `WideString` in String, and overwritten in `ByteString`, instead of being declared abstract in `String`.

As soon as I get my git stuff set up for Pharo 9, I expect to make a PR which fixes this (which I did already in a local image).

kasperosterbye:
Fixes #5549


MarcusDenker (approved):
Looks good: changes are: class replaced by #species, some super -> self and adds #asByteArray

PR #5554, opened 2020-01-22
fix-predebug-error

### LINKED ISSUE #5545: Debugger crashes for any Warning - emergency evaluator opens instead
In latest images of Pharo 8 and 9 (including the "main" ones in Pharo Launcher), whenever debugger should open for Warning, it does not and emergency evaluator opens instead.

In Pharo 8, it happens since build `1121.sha.bbcdf97` (commit bbcdf97 ). 
Build `1120.sha.c115bf4` (commit c115bf4 ) is ok.
On Windows, Linux and Mac.

Steps to reproduce:
1. Get clean latest Pharo 8 image.
2. Run `self notify: 'hi'` or anything else that should signal any warning.

estebanlm:
Fixes  #5545

### DIFF
--- a/src/GT-Debugger/GTSpecPreDebugActionsPresenter.class.st
+++ b/src/GT-Debugger/GTSpecPreDebugActionsPresenter.class.st
@@ -6,6 +6,8 @@ I'll collect the actions via pragmas and create button presenters for them.
 Class {
 	#name : #GTSpecPreDebugActionsPresenter,
 	#superclass : #SpPresenter,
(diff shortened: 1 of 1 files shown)


Ducasse:
Thanks Esteban. 
Do you have an idea what was the cause of the missing trait?

estebanlm:
when you rename the trait (which was what I did), package is not being marked as dirty, hence not shown in commit window. So I missed it :(

Ducasse:
Oops did you enter a bug because we should fix this.

PR #5569, opened 2020-01-24
5562-Executing-All-Instances-in-the-startup-affects-loading-times

### LINKED ISSUE #5562: Executing All Instances in the startup affects loading times.
There are three operations that should be improved:
- Nulling of ExternalAddresses.
- Updating OmSessionStore instances
- Resetting Hiedra HiColumnController

These operations use #allInstances, with a big image traversing the whole heap takes a lot of time.

tesonep:
Removing the use of allInstances during the startup of the image. In a 4GB image the startup is reduced in a 50%

Fixes #5562


estebanlm (approved):

tinchodias:
Oops. Two cases were my fault!

PR #5577, opened 2020-01-24
removeBalloonCanvas

pavel-krivanek:
an experiment to remove BalloonCanvas (unify it with the FormCanvas) to obtain cleaner Canvas API.

(seems to work already)


Ducasse:
Excellent. Now just a question do you think that Pharo can work only with athens/cairo?

pavel-krivanek:
I hope so, that's why I want to make the Canvas API more transparent... and experiment with DPI independency

Ducasse:
This is great. 
I see that sometimes we have sent to undefined methods :(
I do not get why we do not run the lint rules. I will try to relaunch this idea and process.

pavel-krivanek:
single test failure - not related

tesonep (approved):

PR #5581, opened 2020-01-27
feature/add-libraryfinders

estebanlm:
add some helpers to find libraries according to platforms. This should easy the lookup of libraries specially on linux, but also others.


estebanlm:
wait. Some wrong classes went in

PR #5582, opened 2020-01-27
feature/add-libraryfinder

estebanlm:
add some helpers to find libraries according to platforms. This should easy the lookup of libraries specially on linux, but also others.

### DIFF
--- a/src/Athens-Cairo/CairoLibrary.class.st
+++ b/src/Athens-Cairo/CairoLibrary.class.st
@@ -20,14 +20,7 @@ CairoLibrary >> unix32LibraryName [
 	Packages such as snapcraft can also change the path.
 	Look in a number of known locations, or LD_LIBRARY_PATH."
 
-	(#('/usr/lib/i386-linux-gnu' '/usr/lib32' '/usr/lib'),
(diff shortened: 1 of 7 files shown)


Ducasse:
A question why the basePaths are only customisable via subclasses and overriding?
I was thinking that clients could also specify their base paths (so turning the basePaths in an instance variable that can also be set).

estebanlm:
Because I didn't think it has sense: This is a tool to find libraries in the standard directories of a platform (specially annoying in linux, since different distributions put libraries sin different places), not anywhere.
I do not understand who would want to use them outside this purpose?

Ducasse:
Oki

Ducasse:
BTW why do you return all the times newly created ordered collections?
If you call 10 times basePaths you will create 10 collections while one is enough.

Ducasse:
Esteban did you see my last remark?

estebanlm:
yes I saw it. 
Just no time :)
Originally I thought this helpers as a "one time shot", hence caching the arrays wouldn't have much sense... but I can add them, in case someone wants to keep the finder.

Ducasse:
Ok my sole point is that if the method is only invoked once we do not care.
Now if it is regularly invoked then it can be worth.

PR #5585, opened 2020-01-28
Fix rotten DebuggerTest

dionisiydk:
Fix the simulation of process in DebuggerTest to correctly setup the top context to behave like a real process (created by #newProcess method).
It fixes rotten tests which did not run given process block while tests were passing successfully.
For example put #here trace into the process block in #testStepOverHalt: 
```Smalltalk
StepOverTest>>testStepOverHalt
	"Stepping over a self halt should not raise an unhandled exception"
	self settingUpSessionAndProcessAndContextForBlock: [ #here trace. self halt. ].
	self shouldnt: [[ session interruptedProcess isTerminated ] whileFalse: [ session stepOver. ]] raise: Exception.
```
This test will not write anything into Transcript which proves that halt is not executed.
Here the test machinery is fixed and separate test covers the state of simulated process.

The problem was that process methods (and maybe even VM) relies on particular form of top context and it should not be the context of given process block (receiver of #newProcess). 
For example check the implementation of Process>>#isTerminated. It detects the termination when the top context is executed (pc > startPc).

### DIFF
--- a/src/Debugger-Tests/AssignmentAndLiteralDebuggerTest.class.st
+++ b/src/Debugger-Tests/AssignmentAndLiteralDebuggerTest.class.st
@@ -20,7 +20,7 @@ AssignmentAndLiteralDebuggerTest >> testInterruptedContextEqualsSuspendedContext
 
 { #category : #tests }
 AssignmentAndLiteralDebuggerTest >> testNewDebugSession [
-	self assert: context size equals: 0.
+	self assert: context size equals: 1.
 	self assert: session interruptedContext equals: context.
 ]
 
--- a/src/Debugger-Tests/DebuggerTest.class.st
+++ b/src/Debugger-Tests/DebuggerTest.class.st
@@ -34,9 +34,13 @@ DebuggerTest >> setUp [
 
 { #category : #utilities }
 DebuggerTest >> settingUpSessionAndProcessAndContextForBlock: aBlock [
-	context := aBlock asContext.
-	process := Process 
-	    forContext: context 
-	    priority: Processor userInterruptPriority.
+	process := Process
+		forContext: 
+			[aBlock value.
+			Processor terminateActive] asContext
+		priority: Processor userInterruptPriority.
+	[process step closure == aBlock] whileFalse.
+	process step.
+	context := process suspendedContext.
 	session:= process newDebugSessionNamed: 'test session' startedAt: context.
 ]
--- a/src/Debugger-Tests/DebuggerTest.class.st
+++ b/src/Debugger-Tests/DebuggerTest.class.st
@@ -34,11 +34,7 @@ DebuggerTest >> setUp [
 
 { #category : #utilities }
 DebuggerTest >> settingUpSessionAndProcessAndContextForBlock: aBlock [
-	process := Process
-		forContext: 
-			[aBlock value.
-			Processor terminateActive] asContext
-		priority: Processor userInterruptPriority.
+	process := aBlock newProcess.
 	[process step closure == aBlock] whileFalse.
 	process step.
 	context := process suspendedContext.
--- a/src/Debugger-Tests/StepIntoTest.class.st
+++ b/src/Debugger-Tests/StepIntoTest.class.st
@@ -25,20 +25,6 @@ StepIntoTest >> stepA1 [
 ]
 
 { #category : #tests }
-StepIntoTest >> testDetectionOfExecutionEnd [
-	self settingUpSessionAndProcessAndContextForBlock: [  ].
-	session stepInto.
-	"These ways do not detect that the execution ended"
-	self assert: session interruptedContext isDead not.
-	self assert: (session isContextPostMortem: (session interruptedContext)) not.
-	self assert: session interruptedProcess suspendedContext isNil not.
-	self assert: session interruptedProcess suspendedContext isDead not.
-	"This way seems to detect that an execution ended"
-	self assert: session interruptedProcess isTerminated.
-
-]
-
-{ #category : #tests }
 StepIntoTest >> testStepIntoDeadContextShouldRaiseException [
 	self settingUpSessionAndProcessAndContextForBlock: [  ].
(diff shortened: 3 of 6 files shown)


src/Debugger-Tests/AssignmentAndLiteralDebuggerTest.class.st
```diff
@@ -20,7 +20,7 @@ AssignmentAndLiteralDebuggerTest >> testInterruptedContextEqualsSuspendedContext
 
 { #category : #tests }
 AssignmentAndLiteralDebuggerTest >> testNewDebugSession [
-	self assert: context size equals: 0.
+	self assert: context size equals: 1.
```
dionisiydk, line 23: simulated process must have at least two context: the top one and the given process block (what #newProcess does).

src/Debugger-Tests/DebuggerTest.class.st
```diff
@@ -34,9 +34,9 @@ DebuggerTest >> setUp [
 
 { #category : #utilities }
 DebuggerTest >> settingUpSessionAndProcessAndContextForBlock: aBlock [
-	context := aBlock asContext.
-	process := Process 
-	    forContext: context 
-	    priority: Processor userInterruptPriority.
+	process := aBlock newProcess.
+	[process step closure == aBlock] whileFalse.
+	process step.
```
dionisiydk, line 39: The separate step here is to really activate the block - to set the context pc to the next expected instruction. So the following stepInto will do exactly what would be expected in the debugger.
It allowed to remove from many tests the first stepInto to prepare the process block.

src/Debugger-Tests/StepIntoTest.class.st
```diff
@@ -24,20 +24,6 @@ StepIntoTest >> stepA1 [
 	^42
 ]
 
-{ #category : #tests }
-StepIntoTest >> testDetectionOfExecutionEnd [
```
dionisiydk, line 28: This test does not make any sense because stepInto at the end of process block does not terminates it. Try to do it in the debugger on any doIt.
With correctly simulated process this test does not work. But it is the expected behaviour.

src/Debugger-Tests/StepOverTest.class.st
```diff
@@ -70,13 +70,25 @@ StepOverTest >> testErrorSignalledDuringStepOverShouldBeCaught [
 	self shouldnt: [[ session interruptedProcess isTerminated ] whileFalse: [ session stepOver. ]] raise: Error.
 ]
 
+{ #category : #tests }
+StepOverTest >> testSimulatedProcessIsInCorrectState [
```
dionisiydk, line 74: This test covers the problem with incorrectly simulated process.

Ducasse:
Thanks denis we will have to stop running like crazy and look at your fixes.

dupriezt:
I tried to load Denis' branch in an image to have a look, but I got an emergency evaluator when doing that:
image

It doesn't seem related to Denis' code in particular. It looks like a code loading problem.

What's even stranger is that the `ReImplementedNotSentRule` class does have a `#methodRemoved:` method. (I checked in a fresh pharo 9 image)

dupriezt:
Loading it in a pharo 8 image worked. (I was loading it in a pharo 9 image)

dupriezt:
The test `StepOverTest>>#testStepOverHalt` now opens a debugger when run, which is an issue.
image

Here's the code:
```Smalltalk
testStepOverHalt
	"Stepping over a self halt should not raise an unhandled exception"
	self settingUpSessionAndProcessAndContextForBlock: [ self halt. ].
	self shouldnt: [[ session interruptedProcess isTerminated ] whileFalse: [ session stepOver. ]] raise: Exception.
```
Somehow, the Halt signaled by self halt is not caught by the `#shouldnt:raise:`.

dionisiydk:
The problem is with iceberg. When you load PR it really loads my version instead of merging it into the image.
My fix was done in the Pharo version where step over halt in debugger had such issue. 
(Then it was fixed and now it works correctly in Pharo 8 release).

To prove what I said try manually debug "self halt" using Step Over button (in the image with my PR).  It will spawn extra debugger like you see in the test. But PR does not include any debugger changes. It is purely about tests.

StevenCostiou:
> The problem is with iceberg. When you load PR it really loads my version instead of merging it into the image.
> My fix was done in the Pharo version where step over halt in debugger had such issue.
> (Then it was fixed and now it works correctly in Pharo 8 release).
> 
> To prove what I said try manually debug "self halt" using Step Over button (in the image with my PR). It will spawn extra debugger like you see in the test. But PR does not include any debugger changes. It is purely about tests.

I was actually trying to load your code because the changes you made seem right to me. But should I load it in P8 or P9?

dionisiydk:
should work everywhere.
But I did it in old Pharo 8 where step over halt was broken where it was
interesting to understand why test does not fail

пт, 31 янв. 2020 г. в 10:07, Steven Costiou <notifications@github.com>:

dupriezt (approved):
Looks good. The issue with `StepOverTest>>#testStepOverHalt` spawning a debugger should disappear when this PR is integrated into the new Pharo that had this bug (stepping over a halt opens a new debugger) fixed.

StevenCostiou (approved):

StevenCostiou:
> Looks good. The issue with `StepOverTest>>#testStepOverHalt` spawning a debugger should disappear when this PR is integrated into the new Pharo that had this bug (stepping over a halt opens a new debugger) fixed.

I agree.

dionisiydk:
Could you open an Iceberg bug? I think it is quite critical issue

пт, 31 янв. 2020 г. в 10:26, Steven Costiou <notifications@github.com>:

Ducasse:
So can I integrate it?
I thought so.

StevenCostiou:
> So can I integrate it?
> I thought so.

Yes you can integrate it.
The iceberg problem is the loading problem that Thomas encountered and it is separate.
@dionisiydk can you confirm?

dionisiydk:
> @dionisiydk can you confirm?

Yes, merge it please :)

PR #5588, opened 2020-01-29
5387-RBSimpleFormatter-contains-deadcode

### LINKED ISSUE #5387: RBSimpleFormatter contains deadcode
```
formattedSourceFor: aNode
	^ lookaheadCode
		at: aNode
		ifAbsentPut: [ 
			self class new
				indent: 1;
				format: aNode ]
```

is not invoked. 
So it means that the lookaheadCode is always empty.

MarcusDenker:
remove some leftover code that was not removed when the simple formatter was created from the configurable formatter. fixes #5387

### DIFF
--- a/src/AST-Core/RBSimpleFormatter.class.st
+++ b/src/AST-Core/RBSimpleFormatter.class.st
@@ -1,6 +1,7 @@
 "
 I'm a simple formatter that prints nodes (without using source code).
- I'm useful when AST are programmatically assembled. 
+I'm useful when AST are programmatically assembled. 
+
 I could be improved, but the goal is to get the information is a more or less ok form.
 
 "
@@ -10,7 +11,6 @@ Class {
 	#instVars : [
 		'codeStream',
 		'indent',
-		'lookaheadCode',
 		'originalSource',
 		'lineStart'
 	],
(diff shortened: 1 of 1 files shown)


MarcusDenker:
Did the tests work? The first time they showed a failing build. I restarted to check if that is happening again

Ducasse:
Ok I missed it. 
This user interface is really not good.

MarcusDenker:
If the build would be green more often we could configure to not allow non-green merges... it would improve things a lot

Ducasse:
I do not get why the benchmarks are still executed. 
I do not understand the log either.
+ git log HEAD^1 -1 --format=%H
fatal: Not a git repository (or any parent up to mount point /builds)
Stopping at filesystem boundary (GIT_DISCOVERY_ACROSS_FILESYSTEM not set).
script returned exit code 128

Ducasse:
May be we should try this. Because this is really boring to have to click and check all the "green" not green things.

Ducasse:
Do we revert to check?

MarcusDenker:
The build did have some problems but it seems to have uploaded an image with the change, so the complete build failure seems to not have happen again.

We can wait and see what the next PRs do... if they can use the build it should be fine

Ducasse:
yes I saw that too.

PR #5589, opened 2020-01-29
5584-enabling-pharo9x-in-pharo90

### LINKED ISSUE #5584: enabling pharo9.x in pharo90
SmalltalkImage>>#metacelloPlatformAttributes

Ducasse:
Fixes: #5584 like that we can start differentiating 8.0 from 9.0!

### DIFF
--- a/src/BaselineOfSpec2/BaselineOfSpec2.class.st
+++ b/src/BaselineOfSpec2/BaselineOfSpec2.class.st
@@ -51,7 +51,7 @@ BaselineOfSpec2 >> baseline: spec [
 				package: 'Spec2-Morphic-Backend-Tests' with: [ spec requires: #('ParametrizedTests' 'Spec2-Adapters-Morphic') ];
 				package: 'Spec2-Tests' with: [ spec requires: #('Spec2-Examples' 'ParametrizedTests') ] ].
 	spec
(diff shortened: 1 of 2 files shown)


MarcusDenker:
Build failure:

  MessageBrowser
  MessageBrowser>>#example

MCMultiPackageLoader(Object)>>notify:
MCMultiPackageLoader(MCPackageLoader)>>warnAboutDependencies
MCMultiPackageLoader(MCPackageLoader)>>validate
MCMultiPackageLoader(MCPackageLoader)>>loadWithNameLike:
MCVersionLoader>>basicLoadWithNameLike:

Ducasse:
What a shit.
We change one number and we get such kind of error.

Ducasse:
I have NO idea why this is happening. 
I have no idea why the example on MessageBrowser would be impacted especially since it is ridiculously stupid

```
example
	self browseSendersOf: #flag:
```
How come this could be a dependency. 
The bug is somewhere else. :(

Now this change is blocking me so I will go an code in Pharo 80 since I cannot load the packages in Pharo 90. What

Ducasse:
Pablo do you have an idea about this problem? 
Because we cannot express baselines for Pharo 9.0.

PR #5590, opened 2020-01-29
Calypso should use SpNotebookMorph instead TabManager

estebanlm:
Since tab manager handles tabs by forking process and that's suboptimal.
SpNotebookMorph will load tabs lazily.


Ducasse:
why we should not integrate :)?

dionisiydk:
Hi Esteban.

I am curious, what is the goal? To remove old TabManagerMorph?
I am asking because I would expect here the full Spec2 reimplementation of tab component. So why this intermediate step? (especially that you are now mixing morph and spec in the implementation)

estebanlm:
> why we should not integrate :)?

because is not finished :)

> I am curious, what is the goal? To remove old TabManagerMorph?

Yes, because old `TabManagerMorph` spawns a process for each tab and this is affecting systems with heavy charge. 
And yes, this is an intermediate solution since we are going to replace the UI with a Spec2 UI... but that, as you know, is a lot of work and we need a solution for big images now :)
(This, of course, is not "the solution", is just one of the many details we want to tweak to enhance the Pharo experience).

estebanlm:
it can be integrated now :)

estebanlm:
bump

Ducasse:
Esteban why do we name Morph with Sp?

estebanlm:
because this is a Morph made to be used with Spec2, hence the name (yes, I know Calypso does not uses Spec2, but I used it in a couple of cases).

estebanlm:
And yes... it could be extracted/generalised...but is not made yet :)

PR #5593, opened 2020-01-30
CompletionEngine-and-Monticello-backport

### LINKED ISSUE #5364: CompletionEngine and Monticello saving
image

Open Monticello browser and try to save a new version of a package (here a config is used) by giving a comment to the mcz. While typing a version comment you get an error as seen in the screen.

=> System ends up in a problem of completion engine

MarcusDenker:
fix #5364 in Pharo8


MarcusDenker:
hmm... CI does not want to run. But it is the same as we added to Pharo9 already, so I suggest we merge.

PR #5596, opened 2020-01-30
5595-BIConfigurableFormatter-should-not-use-start-and-stop-to-print-comments

### LINKED ISSUE #5595: BIConfigurableFormatter should not use start and stop to print comments
should use

basicFormatCommentFor: aComment

	codeStream nextPutAll: aComment contents

and then we can reduce the referenced to originalSource (the goal is to never use the original source when pretty printing)

MarcusDenker:
do not use originalSource for printing comments. fixes #5595


MarcusDenker:
failure not related

hogoww (approved):
Same as #5594 , looks good to me

PR #5598, opened 2020-01-30
5597-EFFormatter-should-not-use-start-and-stop-to-print-comments

### LINKED ISSUE #5597: EFFormatter should not use start and stop to print comments
same as https://github.com/pharo-project/pharo/pull/5596 but for EFFormatter

MarcusDenker:
fix #5597: do not use orginalSource to print comments. This way we can print comments even when we do not have source originally


hogoww (approved):
Same as #5594 , looks good to me

PR #5604, opened 2020-01-31
[improvement] fix 5603 overlayCompilerbug

hogoww:
I have no idea where I put my feet on this one, but this seems to fix the overlay compiler.
It does not break additional tests on OpalCompiler.
We'll see what the ci says

### DIFF
--- a/src/OpalCompiler-Core/OCExtraBindingScope.class.st
+++ b/src/OpalCompiler-Core/OCExtraBindingScope.class.st
@@ -51,7 +51,7 @@ OCExtraBindingScope >> lookupVar: name [
 				assoc: assoc;
 				scope: self;
 				yourself ].
-	^ nil 
+	^ outerScope lookupVar: name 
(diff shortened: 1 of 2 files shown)


hogoww:
It does not seems to work.
This is still weird because what I changed is a message send to super, which DIRECTLY hits a subclassresponsibility, even if the problem is not coming from that part of the code

MarcusDenker:
Here is a test that shows that we need to 

^ outerScope lookupVar: name

at the end, not return nil.

testArrayEvaluateWithBindingsReadGlobal
	| result |
	result := Smalltalk compiler
		bindings: {(#a -> 3)};
		evaluate: 'OrderedCollection'.
	self assert: result equals: OrderedCollection

hogoww:
Thanks To @MarcusDenker which found the fix !

MarcusDenker (approved):
ok!

PR #5608, opened 2020-01-31
5485-LOC-should-not-count-line-with-space-or-tab-only

olekscode:
Git add wrote tests for CompiledMethod >> linesOfCode and improved the comment


Ducasse:
Oleksandr I have the impression that you introduced some unwanted dependencies

olekscode:
This is very strange.
Because I just wrote tests and fixed a comment

Ducasse:
Indeed it looks ok now :)

PR #5610, opened 2020-01-31
Revert "5392-Proposed-improvement--performance-of-atRandom-in-class-Bag"

MarcusDenker:
Reverts pharo-project/pharo#5445


MarcusDenker:
I discussed with Pablo: we decided to not revert the pharo8 changes but will keep them (they should not have a bad impact).

I close this PR and see how to re-open the original one so we get it into pharo9

PR #5611, opened 2020-01-31
Removing deprecated methods

Alisu:
We have to remove the methods deprecated in Pharo 8 and prepare the infrastructure for Pharo 9


MarcusDenker:
There is something bad happening with this PR: it un=does a prior PR, see the changes to 

RBSimpleFormatter

MarcusDenker:
The PR has to be redone, it contains *a lot* of changes that undo changed already in Pharo8

PR #5614, opened 2020-01-31
Ast tester

jecisc:


hogoww:
(thanks cyril for doing it for me)
Format the tester so they are more easily usable.
This will fail, if someone manipulates code that wasn't semantically analysed (and therefore variables nodes won't be differenciated)

Ducasse:
This is strange since we get

 For more information please see
[64] https://github.com/OpenSmalltalk/opensmalltalk-vm/releases/tag/r3732#linux
[64] KeyNotFound: key #aParameter not found in Dictionary
[64] Dictionary>>errorKeyNotFound:
[64] aBlockClosure(no source code or decompiler available) in Dictionary>>at: in Block: aBlockClosure(no source code or decompiler availab...etc...
[64] Dictionary>>at:ifAbsent:
[64] Dictionary>>at:
[64] IRPushClosureCopy>>indexForVarNamed:
[64] IRTranslatorV2>>visitPushTemp:
[64] IRPushTemp>>accept:
[64] IRTranslatorV2(IRVisitor)>>visitNode:
[64] IRTranslatorV2>>visitInstruction:
[64] aBlockClosure(no source code or decompiler available) in IRTranslatorV2(IRVisitor)>>visitSequence: in Block: aBlockClosure(no source code or decompiler availab...etc...
[64] OrderedCollection>>do:
[64] IRSequence>>do:
[64] IRTranslatorV2(IRVisitor)>>visitSequence:
[64] IRTranslatorV2>>visitSequence:
[64] IRSequence>>accept:
[64] IRTranslatorV2(IRVisitor)>>visitNode:
[64] aBlockClosure(no source code or decompiler available) in IRTr


<blockquote><div> GitHub</div><div><strong><a href="https://github.com/OpenSmalltalk/opensmalltalk-vm">OpenSmalltalk/opensmalltalk-vm</a></strong></div><div>Cross-platform virtual machine for Squeak, Pharo, Cuis, and Newspeak. - OpenSmalltalk/opensmalltalk-vm</div></blockquote>

hogoww:
@Ducasse Would you mind relaunching the CI? Since Cyril did it for me i cannot. 
It's possible that the bootstrap uses code that is not semantically analysed

Ducasse:
Thanks for the suggestion.

guillep:
Well, the problem is not in the bootstrap itself I'd say. Following the logs:

1. The bootstrap successfully finishes
  => It generates a minimal image without sources
2. Several packages (including the compiler) are installed using Hermes
  => we have the compiler image, still without sources
3. Then Monticello is being loaded using the compiler
  => compilation happens by calling `compile:classified:withStamp:notifying:logSource:`
        see the line `Model class(ClassDescription)>>compile:classified:withStamp:notifying:logSource:`
  => here it fails ❌


> This will fail, if someone manipulates code that wasn't semantically analysed (and therefore variables nodes won't be differenciated)

Well this is maybe true, but you're changing the compiler's behaviour and there is no single test about this change in the PR. How can we know the impact of this? Have you run the compiler tests? (I'd assume yes...). What this failure tells me so far is that the changes introduce a bug in the compiler, because the bootstrap does not use a special version of the compiler whatsoever.

src/OpalCompiler-Core/RBVariableNode.extension.st
```diff
@@ -11,39 +11,17 @@ RBVariableNode >> binding: aSemVar [
...
-{ #category : #'*opalcompiler-core' }
-RBVariableNode >> isGlobal [
-	^self binding isGlobal
-]
-
-{ #category : #'*opalcompiler-core' }
-RBVariableNode >> isInstance [
-	^self binding isInstance
-]
-
```
guillep, line 34: This removal is not a "formatting" as said in the description of the PR.

src/OpalCompiler-Core/RBVariableNode.extension.st
```diff
@@ -11,39 +11,17 @@ RBVariableNode >> binding: aSemVar [
 	self propertyAt: #binding put: aSemVar.
 ]
 
-{ #category : #'*opalcompiler-core' }
-RBVariableNode >> isArg [
-
-	^self binding isArg
-]
-
```
guillep, line 19: This removal is not a "formatting" as said in the description of the PR.

guillep (changes requested):
I'd suggest tests are written for this :/ It's not clear to me that running the semantic analysis will replace all variable nodes as this PR is expecting...

hogoww:
This PR was done quickly at the end of the sprint. Just wanted to see if that made sense.
It does not. I'll fix it correctly this time.

hogoww:
@MarcusDenker Should the tester be in the OpalCompiler package or the AST package? 
Currently they are spread around between the two, it does not look great from an API point of view, does it?

guillep:
So what's the status of this?
Could we have an issue description?
A better PR title?

Looking at the changes this is not a "tester". These look like a cleanup of the *testing* methods. Am I right?

What do we do with the broken compiler?

hogoww:
I've been waiting for Marcus to answer one question before doing a new PR, so i guess we can close this one.

guillep:
Ok, I'll close it.
Do you think the discussion in this PR could have value?
What do you think about opening an issue explaining better what was the original issue about?

I was really missing a description, and given that I was not there personally when you did the change, I'm really missing context :). Imagine people that do not understand how the compiler works.

hogoww:
I don't think it is.

The thing is just that using the message isAKindOfVariable is relying on the compiler when it really should not.
Therefore, you can have "message send to nil" pop while having a hard time understanding why

Ducasse:
Where? I do not get it. can you give an example?

hogoww:
Of course, try executing
t := RBTemporaryNode new name: 'pouet' ; yourself.
t isArg

This does not work
(disregard the naming, we already talked about it with marcus, that's not good)
(this should not work from at least pharo 7, I think).

guillep:
The thing is that that node is supposed to be created by the semantic analysis.
Since you're creating it by hand, you're creating it wrong.
There is no correct temporary node without binding.

Now, this would also mean that `RBTemporaryNode`, `RBInstVarNode` and friends, do not belong to the AST, they belong to the compiler.

But you see @hogoww , the discussion has value, we should store this knowledge in an issue.
Otherwise we will lose it.

hogoww:
The discussion up to the previous point had no value* 
What came since then does !

' The thing is that that node is supposed to be created by the semantic analysis.' 
Yeup, the problem is that the compiler is not the only one to create nodes, and still need to be able to test their type.
My current solution (not pushed) is to let the RBVariableNode testers to use the binding, and overide them in the temp/arg (...) subclasses to not rely on the compiler.
That way, when the variables are specialized, they answer correctly, while still being valid before that is the case

hogoww:
Other solution would be to put a fake binding on creating a node, but i don't like this.

guillep:
> but i don't like this.

Could you expand on this? It would be super useful to know why you don't like it.

hogoww:
There could be an object that only knows the API used by the property binding (which would be the lesser evil to me) 
- memory overhead (especially since, if i recall correctly, all the ASTs of the methods in the system are kept in memory) .
- a middle dispatch, which is not useful so time overhead (very mitigated by the inline cache, and probably not that much used anyway, but it exists !)

The tester override that I describe earlier on the other hand are about a constant memory overhead (method for the testers) , and no time overhead

There's probably over solutions, but that's the first two that come to mind.
As for the not liking part, besides the performances, I don't like to have a kind of proxy for a behavior that is quite straightforwardly done otherwise. But that's personal taste.

guillep:
> There could be an object that only knows the API used by the property binding (which would be the lesser evil to me)
> 
> * memory overhead (especially since, if i recall correctly, all the ASTs of the methods in the system are kept in memory) .
> * a middle dispatch, which is not useful so time overhead (very mitigated by the inline cache, and probably not that much used anyway, but it exists !)
> 
> The tester override that I describe earlier on the other hand are about a constant memory overhead (method for the testers) , and no time overhead

Well, this could be the case, but we need to back up these kind of assumptions with benchmarks :).
My experience tells me that in this case, performance (speed or memory wise) is not an issue.
Do you have maybe a problematic case?

I see more problems regarding the dependencies and maintenance in the long term.
I see that lots of compiler details have crippled into the AST.
And this means many usages of the AST do now depend on the compiler, and people have not realised it.

They **will** realise when the compiler changes.

And it will not be pretty, because it will be a ripple change: the compiler breaks, the AST breaks because the compiler breaks, your app breaks because the AST breaks.
That's why respecting boundaries is important.

What I'd like to have => less dependencies.
Make a clear difference between *AST* and *semantically annotated (thus Compiler dependant) AST*.
And users should somehow understand when they are using (and they depend on) the one or the other. Maybe one way is to have different APIs.

hogoww:
"Do you have maybe a problematic case?"
I don't think so, but even without benchmarks, I don't see the use for it because algorithmically it make sense, and low level details wise, also.
Benchmarks could actually give the wrong result, if the memory involved is further apart on one attempt. 
(not that I think it will be the case, but my point being that benchmarks are not the one argument when measuring performances).
I compare these two approaches as one which is heavy, and the other cheap.
Why not use the cheap one, which is not bad ?
I may miss something, of course, but I'd need arguments to make me change my mind =)

"And this means many usages of the AST do now depend on the compiler, and people have not realised it." 
Actually we talk regularly about this with Marcus, I think he told me he's planning to work on that.
I think we all agree that the AST needs a rework, especially interface wise.
If you want a bad one : 
RBReturnNode >> removeNode: aNode
    self parent removeNode: self

" What I'd like to have => less dependencies. "
could not agree more.

guillep:
> "Do you have maybe a problematic case?"
> I don't think so, but even without benchmarks, I don't see the use for it because algorithmically it make sense, and low level details wise, also.
> Benchmarks could actually give the wrong result, if the memory involved is further apart on one attempt.
> (not that I think it will be the case, but my point being that benchmarks are not the one argument when measuring performances).

No, I don't agree. What other argument is there?
If your benchmarks are wrong (e.g., you're benchmarking height by asking people the age), don't blame benchmarking, blame the one who wrote the benchmark.

> I compare these two approaches as one which is heavy, and the other cheap.
> Why not use the cheap one, which is not bad ?

But you use words that can be misleading. Heavy and cheap are relative and ambiguous.
So we need to be more precise, and we need to account for a far more complete set of qualities.
Software is not only good because it is fast ;)

 - what solution is easier to change and why?
 - what solution reduces/increases dependencies ans why?
 - what solution is easier to test and why?
 - what solution is faster/less memory consuming and why?

A super strong building without water or electricity is not a good building :).
Or with water but no toilets.

> I may miss something, of course, but I'd need arguments to make me change my mind =)

Well, when you try to learn and understand slang, let's say the least of the problems you have are machine performance, speed/memory ;). You have more problems on repeated code, misleading semantics, inconsistencies, lack of documentation, lack of hooks, lack of tests. And this happens to everybody on every domain :).

In any case, I feel my arguments are never enough for you, which is a pity because I feel you don't trust me, and I am on your side...

hogoww:
> > "Do you have maybe a problematic case?"
> > I don't think so, but even without benchmarks, I don't see the use for it because algorithmically it make sense, and low level details wise, also.
> > Benchmarks could actually give the wrong result, if the memory involved is further apart on one attempt.
> > (not that I think it will be the case, but my point being that benchmarks are not the one argument when measuring performances).
> 
> No, I don't agree. What other argument is there?
> If your benchmarks are wrong (e.g., you're benchmarking height by asking people the age), don't blame benchmarking, blame the one who wrote the benchmark.
Benchmarking relies on underlying system,  and has a potentiality to be influenced in the wrong way.
That does not mean that they don't have their, use, just that there are not the only thing to consider, as you say it yourself with your arguments in the next lines.

> 
> > I compare these two approaches as one which is heavy, and the other cheap.
> > Why not use the cheap one, which is not bad ?
> 
> But you use words that can be misleading. Heavy and cheap are relative and ambiguous.
Agreed. I abstract words when I don't consider them important. In this particular case, they were a summary of my previous statement, which was precise. 
Theses words by themselves are useless, because as you said it, they are way too general.
> So we need to be more precise, and we need to account for a far more complete set of qualities.
> Software is not only good because it is fast ;)
Agreed. 
Thank you for pointing out theses parts, which allow me to focus on the problem better.
> 
> * what solution is easier to change and why?
My solution of overriding is less modular than an having a kind of object design for the purpose of understanding how to test, but modularity has to stop somewhere.
Also, this depends on what the binding is used for besides testing.
There might be several classes to create to have a little separation of concerns.
Unless there is a real use, overriding the methods in the subclasses is significantly easier to do.
> * what solution reduces/increases dependencies ans why?
I was actually waiting for marcus's answer for that.
But this is not the right question. 
The right question is: "where should the tester methods be put?" Compiler or AST?
And that I don't have the answer to.
As long as this question is not answered, yours won't be, because dependencies are not a consideration as long as there is nothing is nowhere.
> * what solution is easier to test and why?
I think they are both equivalent.
Very slightly more complicated for the object proxy because in one case you have an extra object.
But IMO this is negligible.
> * what solution is faster/less memory consuming and why?
I think I already answered that. (otherwise tell me what is missing)
Important note, we are discussing 2 ideas of solving a problem. I'm not considering at all other possible solutions. Of course.
> 
> A super strong building without water or electricity is not a good building :).
> Or with water but no toilets.
> 
> > I may miss something, of course, but I'd need arguments to make me change my mind =)
>

PR #5616, opened 2020-01-31
[feedback needed] attempt to improve class comment for RemoteString

hogoww:
I don't think that's enough. I'd love some feedback on this

### DIFF
--- a/src/System-Sources/RemoteString.class.st
+++ b/src/System-Sources/RemoteString.class.st
@@ -1,7 +1,7 @@
 "
 My instances provide an external file reference to a piece of text.  It may be the sourceCode of a method, or the class comments of a class.
 
-The 'changes' file or 'file-in' file usually has a chunk that is just the source string of a method:
+The 'changes' file or 'file-in' file usually have a chunk that is just the source string of a method:
 
 max: aNumber
 	^ self > aNumber ifTrue: [self] ifFalse: [aNumber]!
--- a/src/System-Sources/RemoteString.class.st
+++ b/src/System-Sources/RemoteString.class.st
@@ -1,18 +1,19 @@
 "
 My instances provide an external file reference to a piece of text.  It may be the sourceCode of a method, or the class comments of a class.
 
-The changes file or file-in file usually has a chunk that is just the source string of a method:
+The 'changes' file or 'file-in' file usually has a chunk that is just the source string of a method:
 
 max: aNumber
 	^ self > aNumber ifTrue: [self] ifFalse: [aNumber]!
 
-I can return either a String or a Text.  Some a chunk is followed by a second chunk (beginning with ]style[) containing style information.  The encoding is like this:
+I can return either a String or a Text. A chunck of code is followed by a second chunk (beginning with ]style[) containing style information. The encoding is like this:
 
 max: aNumber
 	^ self > aNumber ifTrue: [self] ifFalse: [aNumber]!
 ]style[(14 50 312)f1,f1b,f1LInteger +;i!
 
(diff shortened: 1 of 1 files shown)


MarcusDenker (approved):

src/System-Sources/RemoteString.class.st
```diff
@@ -1,18 +1,19 @@
...
 
-I can return either a String or a Text.  Some a chunk is followed by a second chunk (beginning with ]style[) containing style information.  The encoding is like this:
+I can return either a String or a Text. A chunck of code is followed by a second chunk (beginning with ]style[) containing style information. The encoding is like this:
 
 max: aNumber
 	^ self > aNumber ifTrue: [self] ifFalse: [aNumber]!
 ]style[(14 50 312)f1,f1b,f1LInteger +;i!
 
-Allowed TextAttributes are TextFontChange, TextEmphasis, TextColor, TextDoIt, TextKern, TextLink, TextURL.  TextFontReference and TextAnchor are not supported.
+Allowed TextAttributes are TextFontChange, TextEmphasis, TextColor, TextDoIt, TextKern, TextLink, TextURL.
```
myroslavarm, line 19: I would say "TextAttributes that are allowed are: ..."

src/System-Sources/RemoteString.class.st
```diff
@@ -1,18 +1,19 @@
 "
 My instances provide an external file reference to a piece of text.  It may be the sourceCode of a method, or the class comments of a class.
 
-The changes file or file-in file usually has a chunk that is just the source string of a method:
+The 'changes' file or 'file-in' file usually has a chunk that is just the source string of a method:
```
myroslavarm: I'm not sure whether the changes and file-in files are meant to be different here. If they are, i would say "changes and file-in files usually have..."; if they aren't, i would say "the changes (or file-in) file usually has..."

src/System-Sources/RemoteString.class.st
```diff
@@ -1,18 +1,19 @@
...
 
 max: aNumber
 	^ self > aNumber ifTrue: [self] ifFalse: [aNumber]!
 
-I can return either a String or a Text.  Some a chunk is followed by a second chunk (beginning with ]style[) containing style information.  The encoding is like this:
+I can return either a String or a Text. A chunck of code is followed by a second chunk (beginning with ]style[) containing style information. The encoding is like this:
```
guillep: Not all chunks have style. Only some of them.

src/System-Sources/RemoteString.class.st
guillep: They -> have ;)

guillep (changes requested):
Hi @hogoww could you fix the couple of typos? Otherwise I think it is good :)

hogoww:
Don't know why i cannot answer you on the conversation.
The typo is fixed.
The other one i'm not sure I understand the initial text. so i'm having trouble correcting it.
I'll see tomorrow with a rested mind.

src/System-Sources/RemoteString.class.st
guillep: What this "wanted to mean" (:P) was that in a source file (sources, changes, fileouts) we store chunks of code. Some chunks are styled (they have parts in bold, or font color...). When a chunk is styled, it is followed by a second chunk (prefixed with something like ]sytle[), and specifying the style.

hogoww:
I actually used what you wrote because i think this was waaaaaaaaaay better, I hope you don't mind.
Thanks ! =)

guillep:
Not at all :P

I wrote that with a baby that was drooling on my arm and moving around, so if that was good enough...

hogoww:
Guess he inspired you :D

guillep (approved):
Good enough. I'll integrate

guillep:
Thanks :)

PR #5619, opened 2020-02-01
Update Spec 0.2.1

estebanlm:
merge Spec 0.2.1

fix dependency problems

### DIFF
--- a/src/BaselineOfMorphic/BaselineOfMorphic.class.st
+++ b/src/BaselineOfMorphic/BaselineOfMorphic.class.st
@@ -124,7 +124,6 @@ BaselineOfMorphic >> baseline: spec [
 		spec package: 'Morphic-Widgets-TickList'.
 		
 		spec package: 'System-Changes-FileServices'.
-		spec package: 'System-OSEnvironments'.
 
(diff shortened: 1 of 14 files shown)


estebanlm:
I fixed the problems coming from Spec2 but now there are: 

1) Spec1 dependency problem (shouldn't we remove Spec1?)
2) testExternalUFFIDependencies
3) testExternalMorphicCoreDependencies 

I do not think this is related to my merge, but certainly those problems should not be there.

Any idea, @tesonep, @MarcusDenker ?

Ducasse:
So esteban I should I merge?

estebanlm:
if you want :)
I can still do a new PR to fix the remaining issues... but I wonder why I'm having them.

tesonep:
This merge was integrated somewhere else....

estebanlm:
who?

tesonep:
I see it in the history, it is your commit but it cannot be said how it got
there. There is no merge, maybe it have been pushed in other pr or
directly, both by error. But we have to think how to avoid it. Although, it
is not easy.

This is already fixed, and also we can always revert commits, so it is not
a real issue.

PR #5621, opened 2020-02-01
5617-LazyListMorph-missing-theme-hook

### LINKED ISSUE #5617: LazyListMorph missing theme hook
The color of the row should call the them for both selected and unselected row

colorForRow: row 
	"Answer the color for the row text."
	
	^ (self isRowSelected: row )
		ifTrue: [ self theme selectedItemListTextColor  ]
		ifFalse: [ self theme listTextColor ]

Ducasse:
Fixes: #5617 make sure that we can control the color of non/selected list item.

### DIFF
--- a/src/Polymorph-Widgets/UITheme.class.st
+++ b/src/Polymorph-Widgets/UITheme.class.st
@@ -5322,6 +5322,13 @@ UITheme >> selectedFillStyleFor: aMorph height: anInteger [
 ]
 
 { #category : #'accessing colors' }
+UITheme >> selectedItemListTextColor [
+	"Answer the color for the selected item in a list text."
+
+	^ Color black
+]
+
+{ #category : #'accessing colors' }
 UITheme >> selectionBarColor [
 
 	^ self settings selectionBarColor
--- a/src/Polymorph-Widgets/UITheme.class.st
+++ b/src/Polymorph-Widgets/UITheme.class.st
@@ -5325,7 +5325,7 @@ UITheme >> selectedFillStyleFor: aMorph height: anInteger [
 UITheme >> selectedItemListTextColor [
 	"Answer the color for the selected item in a list text."
 
-	^ Color black
+	^ self textColor
 ]
 
 { #category : #'accessing colors' }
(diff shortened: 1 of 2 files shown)


Ducasse:
Si vous pouvez jeter un oeil pour regarder si cela ne casse pas vos themes
Moi j'en ai besoin pour les miens.

src/Polymorph-Widgets/UITheme.class.st
```diff
@@ -5321,6 +5321,13 @@ UITheme >> selectedFillStyleFor: aMorph height: anInteger [
 
 ]
 
+{ #category : #'accessing colors' }
+UITheme >> selectedItemListTextColor [
+	"Answer the color for the selected item in a list text."
+
+	^ Color black
```
estebanlm: default selected color needs to be "textColor", not black.

estebanlm (changes requested):

src/Polymorph-Widgets/UITheme.class.st
Ducasse: Ok I will check if it is working with my style. My problem is that I could not see the text of certain items. 
Thanks for the feedback

Ducasse: This does not work for me. 


Screenshot 2020-02-01 at 11 33 58

Ducasse: This is the only place in the system it does not work. 
I do not understand why this is happening. In particular since

In my theme I redefined textColor this way. 

	"This is only for all the list and the text except that the selection color can hide it on certain list."
	^ Color black

Ducasse: I'm investigating (but my brain is fucked by the flu) still I spent a lot of time on this issue in the past. There is something strange.

Ducasse: Now I'm totally confused :( So probably should go to bed. ;(

Ducasse: Ok apparently lazyListMorph is used in dropBox. Not the finder tree.

Ducasse: So for the dropbox what you propose is working.

Ducasse:
Now I have a problem with the tree color or something like that.

Ducasse:
I do not get why the second change I did does not show up. 
So boring.

Ducasse:
The change was done!

PR #5607, opened 2020-01-31
[improvement] Kasper's implementation of atRandom for bags. 

hogoww:
See PR #5445
@kasperosterbye =]

PR #5630, opened 2020-02-05
5392-atRandom-Pharo9

### LINKED ISSUE #5392: Proposed improvement : performance of #atRandom: in class Bag
Whenever using Bag instances with a LARGE number of objects (like millions), #atRandom becomes less and less useable since it almost crawls to a halt. The problem resides in the fact that the Bag class inherits #atRandom: (which is called by #atRandom) from Collection and the current implementation of Collection>>#atRandom: uses #do: to iterate over EVERY item (virtual) of the bag. A faster method would be to use the occurrences of every object in the bag to iterate only over the keys in the dictionary instead of iterating over all the "virtual" items contained in the bag.  

I have included a fix (see attached file).

Here are some example timings to show the performance difference:

Tests WITHOUT #atRandom: override!

Very good case : evaluated #atRandom 200 times in 17.27 seconds for 3000000 elements!
Average case : evaluated #atRandom 200 times in 193.38 seconds for 3000000 elements!
Worst case : evaluated #atRandom 200 times in 283.19 seconds for 3000000 elements!

Tests WITHOUT #atRandom: override!

Very good case : evaluated #atRandom 200 times in 16.26 seconds for 3000000 elements!
Average case : evaluated #atRandom 200 times in 185.52 seconds for 3000000 elements!
Worst case : evaluated #atRandom 200 times in 295.37 seconds for 3000000 elements!

--------------

Tests WITH #atRandom: override!

Very good case : evaluated #atRandom 200 times in 12.76 seconds for 3000000 elements!
Average case : evaluated #atRandom 200 times in 37.89 seconds for 3000000 elements!
Worst case : evaluated #atRandom 200 times in 44.79 seconds for 3000000 elements!

Tests WITH #atRandom: override!

Very good case : evaluated #atRandom 200 times in 12.61 seconds for 3000000 elements!
Average case : evaluated #atRandom 200 times in 37.93 seconds for 3000000 elements!
Worst case : evaluated #atRandom 200 times in 44.73 seconds for 3000000 elements!
[Bag-atRandom.zip](https://github.com/pharo-project/pharo/files/3977404/Bag-atRandom.zip)

MarcusDenker:
fixes #5392 for Pharo9. The method is the same as the one in Pharo8, I will open a new issue for further suggested improvements

### DIFF
--- a/src/Random-Core/Bag.extension.st
+++ b/src/Random-Core/Bag.extension.st
@@ -1,0 +1,18 @@
+Extension { #name : #Bag }
+
+{ #category : #'*Random-Core' }
+Bag >> atRandom: aGenerator [
+	"Answer a random element of the receiver. Uses aGenerator which
+	should be kept by the user in a variable and used every time. Use
+	this instead of #atRandom for better uniformity of random numbers because 
+	only you use the generator. Causes an error if self has no elements."
+
+	| rand index |
+ 	self emptyCheck.
+ 	rand := aGenerator nextInt: self size.
+ 	"overwritten to use a faster enumeration"
+ 	index := 0.
+ 	self doWithOccurrences: [ :key :count | 
+ 		index := index + count.
+ 		rand <= index ifTrue: [ ^ key ] ]
+]


MarcusDenker:
Ups. We have already https://github.com/pharo-project/pharo/pull/5607

MarcusDenker:
No, lets *first* get it in sync with Pharo8. Then we can improve

MarcusDenker:
(the other one was for Set, this is for Bag)

MarcusDenker:
failure not related

PR #5636, opened 2020-02-05
5633-Speedup-implementors-and-senders-do-not-wrap-in-Ring1-entity

### LINKED ISSUE #5633: Speedup implementors and senders: do not wrap in Ring1 entity
senders is not very fast, part of the reason is that we wrap every single result in a RGMethod.

This is done by #methodReference. 

We should try to remove it. What happens if we just return compiledMethods? In the past they did not know their class nor selector, so a wrapper was needed to have that information.

I suggest to remove the #methodReference sends and then check if the tools can deal with raw methods. One thing to check is especially when code is updated: will the tools show that?

MarcusDenker:
do not wrap senders and implementors in Ring1 entities. Fixes #5633

### DIFF
--- a/src/Kernel-Tests-Extended/BehaviorTest.extension.st
+++ b/src/Kernel-Tests-Extended/BehaviorTest.extension.st
@@ -5,15 +5,15 @@ BehaviorTest >> testAllReferencesTo [
 	| result |
 	result := SystemNavigation new allReferencesTo: Point binding.
 	result do: [ :each | self assert: (each compiledMethod hasLiteral: Point binding) ].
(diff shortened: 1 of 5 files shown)


MarcusDenker:
[#ifTrue: senders ] bench  

Before:  "'4.726 per second'"
After: "'14.152 per second'"

The more senders found, the larger the difference.

MarcusDenker:
Failing tests:

BehaviorTest>>testAllReferencesTo
SystemNavigationTest>>testAllSendersOfReturnsExistingMethods
SystemNavigationTest>>testIsMessageSentInSystemAnswersMethodReference

MarcusDenker:
failing tests are not related

PR #5640, opened 2020-02-06
5478-Inspector-on-a-ByteSymbol-should-get-the-senders-of-this-symbol-on-a-lazy-way

### LINKED ISSUE #5478: Inspector on a ByteSymbol should get the senders of this symbol on a lazy way

MarcusDenker:
Speedup "senders" and "implementors" tab. 
- It is already faster due to not wrapping in Ring Definitions
- It was getting the lists twice, I changed that to only one time
- it was *rendering* the code for the method, not the name --> much faster without that.

It might be still too slow on large images, but we should check that before we remove.

fixes #5478

### DIFF
--- a/src/GT-InspectorExtensions-Core/Symbol.extension.st
+++ b/src/GT-InspectorExtensions-Core/Symbol.extension.st
@@ -6,7 +6,7 @@ Symbol >> gtInspectorImplementorsIn: composite [
 	composite list 
 		title: 'Implementors';
 		display: [ self systemNavigation allImplementorsOf: self ];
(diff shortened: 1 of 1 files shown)


MarcusDenker:
Bootstrap failes with a DNU for 

RGBehavior #announceEnvironmentChanged 

which does not exist in either Pharo 8 nor Pharo9

MarcusDenker:
Bootstrap got fixed, error not related

PR #5641, opened 2020-02-06
RemoveAllInstancesInspector

MarcusDenker:
#allInstances is too slow on large images and can potentially lead to *very* large lists to be rendered. The best is to remove this tab


tesonep (approved):

jecisc:
Maybe we should have in the Spec inspector the possibility to declare a tab like potentially slow and when the user clicks on it he could have the possibility to stop the creation of the tab?

PR #5644, opened 2020-02-06
5551-Inspecting-a-CompiledMethod-should-allow-update

### LINKED ISSUE #5551: Inspecting a CompiledMethod should allow update?
It is possible to see the source code of a CompiledMethod in the inspector. One use scenario I have found is 
```smalltalk
("complicated collect/select of compiled methods") inspect
```
which gives me in the inspector a list of methods which have a specific problem I queried for.
Now, to fix these problems, I need to go and find the method in the browser, rather than editing it directly in the inspector where it is showing.

One example I looked at was to find all the methods in String which sends `class` to self:
```smalltalk
String methods select: [ :m | m ast selfMessages includes: #class ]
```
some of these should be changed to `self species`, other should not. But it would be convenient to do it in the inspector where I got the list of methods, and looked at the source code.

MarcusDenker:
Make a source editing view the default for CompiledMethod, keep the old style for CompiledBlock. Fixes #5551

### DIFF
--- a/src/GT-InspectorExtensions-Core/CompiledBlock.extension.st
+++ b/src/GT-InspectorExtensions-Core/CompiledBlock.extension.st
@@ -4,3 +4,15 @@ Extension { #name : #CompiledBlock }
 CompiledBlock >> gtDisplayOn: stream [
 	self printOn: stream
 ]
+
+{ #category : #'*GT-InspectorExtensions-Core' }
+CompiledBlock >> gtInspectorSourceIn: composite [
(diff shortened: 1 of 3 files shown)


tesonep:
@MarcusDenker this can lead to losing code. If we are inspecting an older version of the method. 
Maybe it should check that the method inspecting is the one that is installed in the class. This is at least, I am not sure if this is enough.

MarcusDenker:
I close it for now

PR #5657, opened 2020-02-10
5401-AST-tests-should-not-depend-on-the-formatter

### LINKED ISSUE #5401: AST tests should not depend on the formatter
`testReplaceAssignment` and `testReplaceBlock` use the formatter to assert that they did a correct rewriting. However, this is a bit fragile because as soon as the formatter setting changes the tests are broken.

It would be simpler to depend on an AST dump or a storeOn:

MarcusDenker:
do not use pretty printer in RBProgramNodeTest. Fixes #5401


MarcusDenker:
failure not related

PR #5659, opened 2020-02-10
5650-Make-unclosable-in-SystemWindow-does-not-work-revert-PR-5552

### LINKED ISSUE #5650: "Make unclosable" in SystemWindow does not work (revert PR #5552)
and it is my fault :-/

It turns out that PR #5552 was not completely sound.

My analysis of the problem is the following:

In SystemWindow class there is a method `windowMenuOn:`, which has the following code:
```smalltalk
	closableLabel := aBuilder model mustNotClose
		ifFalse: [ #'Make unclosable' ]
		ifTrue: [ #'Make closable' ].
	(aBuilder item: closableLabel)
		action: [ aBuilder model perform: closableLabel asCamelCase uncapitalized ].
```

The method `asCamelCase` were one of the methods in #5552 which was changed from using `self class` to `self species`. It turns out the the species method for symbols is their corresponding string class. For example `ByteSymbol` implements species as:

```smalltalk
species
	"Answer the preferred class for reconstructing the receiver."
	^ByteString
```

My guess is that this is done because symbols are immutable, and do not allow the `at:put:` used in `replaceFrom:to: with: startingAt:` called from `copyFrom:to:`. And symbols need to be able to be copied.

I believe the closable code is ugly as it is, and should be fixed in any case.

kasperosterbye:
Fixes #5650
While the root course of the problem is deeper, this was ugly code anyways.

### DIFF
--- a/src/Morphic-Widgets-Windows/SystemWindow.class.st
+++ b/src/Morphic-Widgets-Windows/SystemWindow.class.st
@@ -455,7 +455,7 @@ SystemWindow class >> windowMenuOn: aBuilder [
 		ifFalse: [ #'Make unclosable' ]
 		ifTrue: [ #'Make closable' ].
 	(aBuilder item: closableLabel)
-		action: [ "aBuilder model perform: closableLabel asCamelCase uncapitalized "
+		action: [ 
 			(closableLabel = #'Make unclosable')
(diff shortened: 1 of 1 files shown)


MarcusDenker:
error not related

MarcusDenker (approved):

PR #5660, opened 2020-02-10
redo-5464-Bug-Slot-template-add-variable-without-hash-Crash-on-accept

### LINKED ISSUE #5464: Bug Slot template: add variable without #hash: Crash on accept

MarcusDenker:
fixes #5464 by checking for the OCUndeclaredVariableWarning and warning in that case

### DIFF
--- a/src/Calypso-SystemQueries/ClySystemEnvironment.class.st
+++ b/src/Calypso-SystemQueries/ClySystemEnvironment.class.st
@@ -140,16 +140,17 @@ Is this really what you want to do?') asText
 						makeBoldFrom: 1
 						to: newClassName size))
 				ifFalse: [ ^ nil ] ].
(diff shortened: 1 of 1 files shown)


MarcusDenker:
failing tests not related

dionisiydk (approved):

PR #5661, opened 2020-02-10
Slots: Unify Api Layout and Scope

MarcusDenker:
on a  class we ask for #slotNamed, then on the Layput it is #resolveSlot:, to then be #atName: on the Layout.

As a frist step, we unify the API of Layout and Scope to both use #resolveSlot:, this can be done without deprecation as it is not a public API.

In a later step, we might unify everthing to be #slotNamed:


MarcusDenker:
error not related

PR #5663, opened 2020-02-11
5631-Improve-atRandom-

### LINKED ISSUE #5631: Improve #atRandom:
After merging https://github.com/pharo-project/pharo/pull/5630 

1) could be done without index:

(rand := rand - count) <= 0 ifTrue: [ ^key ]

2) After that, we can use "index" for the var that is now named rand.

3) Check OrderedCollection>>#rand: so that both are in sync (both regarding to empty check and use of just one temp)

MarcusDenker:
- remove the #errorEmptyCollection in #atRandom on Collection, as we do an emptyCheck in the first line
- simplify Bag>>#atRandom: as described in the issue tracker entry

We could simplidy Collection>>#atRadom more if we would have withIndexDo: on Collection (and I really wonder why we do not have that)

fixes #5631

### DIFF
--- a/src/Random-Core/Bag.extension.st
+++ b/src/Random-Core/Bag.extension.st
@@ -7,12 +7,10 @@ Bag >> atRandom: aGenerator [
 	this instead of #atRandom for better uniformity of random numbers because 
 	only you use the generator. Causes an error if self has no elements."
 
-	| rand index |
+	| index |
  	self emptyCheck.
- 	rand := aGenerator nextInt: self size.
+ 	index := aGenerator nextInt: self size.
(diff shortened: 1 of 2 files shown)


svenvc:
OK

Not all collections are indexable, though they are all iterable in some deterministic order.

MarcusDenker:
I added two followups to this:

1)  withIndexDo:  should be in Collection:    https://github.com/pharo-project/pharo/pull/5670 ( PR in need of review)

2) Collection>>#atRandom: can use #withIndexDo:  https://github.com/pharo-project/pharo/issues/5684

PR #5669, opened 2020-02-12
Completion: Sorter should be Instancevar in context

MarcusDenker:
- make the sorter a invar of the completion context. This reduces instantiation, good for sorters that prepare data.
in addition it allows tests to be written that change the sorter.

- make sure that the sorter gets an OrderedCollection handed in

### DIFF
--- a/src/NECompletion/CompletionContext.class.st
+++ b/src/NECompletion/CompletionContext.class.st
@@ -13,7 +13,8 @@ Class {
 		'isWorkspace',
 		'class',
 		'entries',
-		'editor'
+		'editor',
+		'sorter'
 	],
 	#classVars : [
 		'SorterClass'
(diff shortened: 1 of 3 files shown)


myroslavarm (approved):
this is a really good change!

src/NECompletion/CompletionContext.class.st
```diff
@@ -108,10 +110,7 @@ CompletionContext >> hasMessage [
 
 { #category : #entries }
 CompletionContext >> initEntries [
-	| suggestionsList sorter |
-	sorter := self class sorterClass new context: self.
-	suggestionsList := sorter sortCompletionList: (node completionEntries: position).
-	^ suggestionsList
+	^ sorter sortCompletionList: (node completionEntries: position) asOrderedCollection
```
dionisiydk, line 113: Did you think to avoid sorter classes at all? 
I think they can be easily replaced by SortFunction instances. In this case the code would look like:
```Smalltalk
(node completionEntries) sorted: sortFunction
``` 
For simple alphabetic order we would have
```Smalltalk
SortFunction default
```
And of reversed order:
```
SortFunction default reversed
```

MarcusDenker: The idea was that people should be able to create more sorters, and those showing up in the Settings automatically

myroslavarm: @dionisiydk This implementation was specifically created so it would be extendable and could include more sophisticated sorting strategies all part of this one tool.

PR #5670, opened 2020-02-13
unify implementation of #doWithIndex: and withIndexDo

MarcusDenker:
we have both doWithIndex: and withIndexDo:. This PR does *not* try to solve that problem. Instead it just unifies the implementation
- doWithIndex: was implemented on HashedCollection, the other not --> implement withIndexDo: on Collection.
- implement doWithIndex: just in Collection to forward to #withIndexDo:


MarcusDenker:
failing tests not related

Issue #5687, opened 2020-02-13
Adopt WelcomeHelp size to show all text


astares:
Currently the size of the welcome help window is too small to show the full text.
So one has to use the scrollbar or resize the window:

![image](https://user-images.githubusercontent.com/5980033/74472840-d9904280-4ea2-11ea-9cc0-6b16f4b070d8.png)

What to do:
The size of the welcome help should be adopted to show the full text with hints to first time users

MarcusDenker:
merged

PR #5688, opened 2020-02-13
Adopt WelcomeHelp size to show all text

astares:
Fix for #5687


MarcusDenker (approved):
I think this is good, yes...

Issue #5693, opened 2020-02-14
DrTests: in test coverage, show class name in drTestName, not just selector


StephanEggermont:

StephanEggermont:
Trivial

PR #5694, opened 2020-02-14
Also show the class name.

StephanEggermont:
And don't use name because that doesn't work if methodClass return nil

Solves Issue #5693


src/DrTests-TestCoverage/CompiledMethod.extension.st
```diff
@@ -2,5 +2,5 @@ Extension { #name : #CompiledMethod }
 
 { #category : #'*DrTests-TestCoverage' }
 CompiledMethod >> drTestsName [
-	^ self selector asString
+	^ self methodClass asString, '>>#', self selector asString
```
dionisiydk, line 5:
```suggestion
	^ self printString
```

Issue #5695, opened 2020-02-14
[Port to 8] DrTests: in test coverage, show class name in drTestName, not just selector


StephanEggermont:
trivial change

MarcusDenker:
merged

PR #5696, opened 2020-02-14
Also show the class name

StephanEggermont:
Solves Issue #5695


MarcusDenker:
this has been already merged on Pharo9. There was a sugestion for an improvement (to just us printString I think), which we can do in Pharo9.

PR #5707, opened 2020-02-18
WIP: ContextSimplifyExecutedPC

MarcusDenker:
Experiment to use the same logic in #executedPC as we use in  #pcRangeContextIsActive:

- use #endPC for dead context as in #pcRangeContextIsActive:

### DIFF
--- a/src/Debugging-Core/Context.extension.st
+++ b/src/Debugging-Core/Context.extension.st
@@ -61,16 +61,12 @@ Context >> namedTempAt: index put: aValue [
 Context >> pcRangeContextIsActive: contextIsActive [
  	"return the debug highlight for aPC"		 
 	| thePC |
-
-  	"make sure we have some usable value (can happen for contexts in the ProcessBrowser"		
+	"make sure we have some usable value (can happen for contexts in the ProcessBrowser"		
  	thePC := pc ifNil: [self method endPC].		
-
-  	"When on the top of the stack the pc is pointing to right instruction, but deeper in the stack		
- 	the pc was already advanced one bytecode, so we need to go back this one bytecode, which		
- 	can consist of multiple bytes. But on IR, we record the *last* bytecode offset as the offset of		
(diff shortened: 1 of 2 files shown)


guillep:
Can we not push experimental code to the main branch?

MarcusDenker:
I want to check if the tests are green, I added "WIP" to the title.

MarcusDenker:
(but in general I prefer to do small changes that then everyone uses and we see fast if they are wrong instead of doing 20 of these just in my own image and then later merge them all at once...)

guillep:
Sure, but we should avoid integrating immature things.
Moreover, this PR does not add a test for the change, so it's difficult to see its effect.

MarcusDenker:
The nice thing is that we have already tests... and what I want is to check that nothing changes. So writing a test that fails before but is green is kind of hard...

But in the end: yes, I should setup another CI for this.

Ducasse:
may be we could have a ci for the branches under dev.

estebanlm:
there is already a CI for that.
1) all PRs will build an image artefact.
2) all branches under pharo-project/pharo will be built.

The only difference is *where* you get the image if you want to work taking that into account.
(you take it from "artefacts" on Jenkins)

Ducasse:
I do not get it. But if during 3 months we work on PR, it will always be in the PRs list and we will see it constantly.

estebanlm:
Yes... and sometimes is what you want. 
You may prefix them with WIP marking that is not finish (which reminds me I need to make mergeable the one on calypso). 
If you do not want to appear as PR, you work in a branch (under pharo-project/pharo), which is the point 2 of what I described (of course, not many people can do that, just the most trusted developers). 

In both cases you will have an artefact to work on as a result.

Ducasse:
Ok I did not know the story about the branch

MarcusDenker:
Nice thing: it's green and I will do a cleanup PR later based on this. I close this.

PR #5708, opened 2020-02-18
enhance-externaladdress-compare

estebanlm:
simplifying comparisson by skipping one #isKindOf: when we know we are comparing two ExternalAddress.
Benchmark show an improvement of x3 in speed of lookup by address.


estebanlm:
errors not related

PR #5710, opened 2020-02-20
Context: #sourceNode simplify by introducing #scope

MarcusDenker:
#sourceNode on Context was returning the node of an optimized block if the context was executing this optimized block. This is not good as what you naively expect is that "thisContext sourceNode" returns the ast node of the block or method that created the context.  

It turns out that allt he users that expected the optimized block actually are interested in the *scope* of the block.

- add a method #scope
- change all senders of #sourceNode scope to #scope
- simplify #soureNode. (just looking at the code tells us that this is how it should be)

sourceNode
	"Return the source node that created the method or the block of this context"

	^ closureOrNil
		ifNil: [ self method sourceNode ]
		ifNotNil: [ closureOrNil sourceNode ]

### DIFF
--- a/src/Debugging-Core/Context.extension.st
+++ b/src/Debugging-Core/Context.extension.st
@@ -22,7 +22,7 @@ Context >> callPrimitive: primNumber [
 { #category : #'*Debugging-Core' }
 Context >> lookupSymbol: aSymbol [
 	| var |
-	var := self scope lookupVar: aSymbol.
+	var := self astScope lookupVar: aSymbol.
 	"Local variables"
 	var isTemp ifTrue: [^ self tempNamed: aSymbol].
 	"Instance variables"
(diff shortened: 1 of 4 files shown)


MarcusDenker:
failing test not related (network timeout)

dionisiydk:
Hi Marcus.
Maybe more meaningful name can be use like astScope.

dionisiydk (approved):
I thought you removed sourceNodeExecuted. Is it in another PR or am I wrong?

MarcusDenker:
#sourceNodeExecuted we still need.... the first step I did was to use it less, there is another PR to simplify the #executedPC method.

But the #sourceNodeExecuted we need to use to get the astScope, as if we are in an optimised block we need to get first to the ast node that executes and then up to the block.

PR #5718, opened 2020-02-20
Fix: exporting only non empty RGPackages to Tonel

carolahp:
The export was failing when trying to export an empty RGPackage, this happened when the Extensions package is empty. Now only non-empty RGPackages are exported


MarcusDenker:
failing teste not related

myroslavarm (approved):
imho for readibility purposes might be better to have `isNotEmpty` instead of `isEmpty not`
but otherwise this is great :)

MarcusDenker (approved):
ok

PR #5727, opened 2020-02-21
ManifestDeprecated needs to be ManifestDeprecated80

### LINKED ISSUE #5726: ManifestDeprecated needs to be ManifestDeprecated80
to be in synch with package name

astares:
Fix #5726


MarcusDenker:
which reminds me that we should remove that package :-)

MarcusDenker (approved):

PR #5729, opened 2020-02-21
ManifestSpec2CommanderTests should be ManifestSpec2Commander2Tests

### LINKED ISSUE #5728: ManifestSpec2CommanderTests should be ManifestSpec2Commander2Tests
as it is in package "Spec2-Commander2-Tests-Manifest" and should stay in synch with package name

This is the last case to fix

astares:
as it is in package "Spec2-Commander2-Tests-Manifest" and should stay in synch with package name

Fix #5728

### DIFF
--- a/src/Spec2-Commander2-Tests/ManifestSpec2Commander2Tests.class.st
+++ b/src/Spec2-Commander2-Tests/ManifestSpec2Commander2Tests.class.st
@@ -6,7 +6,7 @@ Spec is a framework in Pharo for describing user interfaces. It allows for the c
 I contain test about the integration of Commander 2 in Spec.
 "
 Class {
-	#name : #ManifestSpec2CommanderTests,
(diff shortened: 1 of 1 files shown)


astares:
This normally is a change for the Spec2 repo - and I asked @estebanlm to do it on Spec2 side
so we do not have to wait for a reintegration/resynch of Spec2 packages.

astares:
@estebanlm did the similar change on Spec2 now:

https://github.com/pharo-spec/Spec/commit/43627d4c92238cd1552450e9ff860ff031efd30a

@MarcusDenker 
So we can savely apply this PR to the regular Pharo 9 image already

astares:
mhhh ... anyone able to restart CI for that one. Crash seems unrelated - I doubt it is from the renaming.

MarcusDenker (approved):

PR #5731, opened 2020-02-21
5724-Improve-AdditionalMethodStaterefersToLiteral

### LINKED ISSUE #5724: Improve AdditionalMethodState>>refersToLiteral:
In https://github.com/pharo-project/pharo/issues/5154 Denis wrote:

I found another ugly method AdditionalMethodState>>refersToLiteral:

`(propertyOrPragma isVariableBinding
			ifTrue: [propertyOrPragma key == literal
					or: [propertyOrPragma value == literal
					or: [propertyOrPragma value isArray
						and: [propertyOrPragma value hasLiteral: literal]]]]
			ifFalse: [propertyOrPragma hasLiteral: literal]) ifTrue:
			`

			
We can define refersToLiteral: in Pragma and VariableBinding (or what is the property here?). And the code will be very simple:

`1 to: self basicSize do: [:i |
		| propertyOrPragma "<Association|Pragma>" |
		propertyOrPragma := self basicAt: i.
                (propertyOrPragma refersToLiteral: literal) ifTrue: [^true]`

MarcusDenker:
implement #hasLiteral: in lookup key, use it to simplify method #refersToLiteral:. Fixes #5724

### DIFF
--- a/src/Collections-Support/LookupKey.class.st
+++ b/src/Collections-Support/LookupKey.class.st
@@ -32,6 +32,13 @@ LookupKey >> = aLookupKey [
 		ifFalse: [^false]
 ]
 
+{ #category : #testing }
+LookupKey >> hasLiteral: literal [
+	^ self key == literal
+		or: [ self value == literal
+				or: [ self value isArray and: [ self value hasLiteral: literal ] ] ]
+]
+
 { #category : #comparing }
 LookupKey >> hash [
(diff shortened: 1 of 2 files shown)


MarcusDenker:
this was not a good change: semantics are a bit odd: for Bindings, we want to only look into them deeply if they are in a property. Else not. I will revert it and put a comment

PR #5732, opened 2020-02-22
Renaming packages leads to out of synch manifest names

### LINKED ISSUE #5725: Renaming packages leads to out of synch manifest names
The name of Manifest classes are created based on the package name. See #createManifestNamed: 
method implemented in TheManifestBuilder which is implemented like

```Smalltalk
createManifestNamed: packageName
	|tag|
	tag := self class manifestTag.
	manifestClass := PackageManifest subclass: (tag, packageName onlyLetters) asSymbol
			instanceVariableNames: ''
			classVariableNames: ''
			poolDictionaries: ''
			package: packageName.
	manifestClass 
		tagWith: tag;
		comment: self class manifestClassComment.
	^ manifestClass
```

now when you rename a package the class name of the manifest class is not renamed.
Also the above method only uses the letters of the package names - but we have package
names like "Spec" and "Spec2" which then might collide with the manifest class name

This can lead to 
 - the package name not being in synch with the manifest name anymore
 - collissions in manifest names as "Spec" and "Spec2" would both lead to a ManifestOfSpec
and other side effects

astares:
- proper categorization of "accessing methods" on class side
- refactor so that we have a proper #manifestClassNameFor: aPackageName on class side instead of hiding the name
  building within a method
- use #isAlphaNumeric instead of #onlyLetters as we have packages like "Spec" and "Spec2" which now would
   have different manifest names (instead of a bad collision)
- add a release test that checks that package name and manifest names are in synch

- Fixes #5725

### DIFF
--- a/src/Manifest-Core/TheManifestBuilder.class.st
+++ b/src/Manifest-Core/TheManifestBuilder.class.st
@@ -13,27 +13,27 @@ Class {
 	#category : #'Manifest-Core-Base'
 }
 
-{ #category : #utilities }
+{ #category : #accessing }
 TheManifestBuilder class >> allManifestClasses [
 
 	^ PackageManifest subclasses
 ]
 
-{ #category : #utilities }
+{ #category : #'accessing - tags' }
 TheManifestBuilder class >> falsePositiveBeginningTag [
 	"the string that identifies uniquely the beginning of a selector who give  the set of false positive for a rule"
 	
 	^ 'rule'
 ]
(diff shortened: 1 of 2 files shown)


astares:
These two:
https://github.com/pharo-project/pharo/pull/5736
https://github.com/pharo-project/pharo/pull/5734
need to be merged first

MarcusDenker (approved):
looks good

PR #5738, opened 2020-02-22
stepGlobal on Transcript slightly more "correct".

David5i6:
stepGlobal on Transcript slightly more "correct".

### DIFF
--- a/src/Transcript-Core/ThreadSafeTranscript.class.st
+++ b/src/Transcript-Core/ThreadSafeTranscript.class.st
@@ -240,14 +240,13 @@ ThreadSafeTranscript >> stepGlobal [
 		stepContents := ''.
 		self changed: #clearText.
 	].
-	deferredEndEntry ifTrue: 
-	[	deferredEndEntry := false.  	
-		self critical: 
-		[	stepContents := stream contents.
-			stream resetContents.	
-		].
-		self changed: #appendEntry.	
+	deferredEndEntry ifFalse: [ ^self]. 
+	deferredEndEntry := false.  	
+	self critical: 
+	[	stepContents := stream contents.
+		stream resetContents.	
 	].
+	self changed: #appendEntry.	
 	
 ]
 


David5i6:
Edited description

Ducasse:
deferredEndEntry ifFalse: [ ^self]. 
 	deferredEndEntry := false.  	

I do not really understand this logic

David5i6:
Hi, just followed the steps on a critique that says:

> 

> Transforms single branch conditionals with multi-statement bodies into a sequence of statements guarded by a conditional return.
> For example
> 
> [[[ 
> foo 
> 	statements. 
> 	condition ifTrue: [ statement1. statement2 ]
> ]]]
> 
> is transformed into 
> 
> [[[  
> foo
> 	statements.
> 	condition ifFalse: [^self].
> 	statement1.
> 	statement2.
> ]]]

imagen


The change was made automatically.
What it does is returning if the boolean value is False, otherwise, set it to false (to not execute that bloc again).

svenvc:
This is just a transformation, the code is totally equivalent, there is no functional change

Ducasse:
I prefer the flow of the existing version.

David5i6:
I know is totally equivalent, but ... if there is a tool included in the Pharo Image that sugest the replacement of those cases with a guard clause, I think the code should follow that rule, otherwise the rule will be removed for being useles. Or perhaps there's something I've missed...

Ducasse:
It is not that simple :)
Else we would have fix it. 

Sometimes it is better to have a guard at the beginning of the method since it makes the logic

if.....^ self
....
...
...

Now in the present case the benefit is not that obvious. 

```
deferredClear ifNil: [ deferredClear := false ].
deferredEndEntry ifNil: [ deferredEndEntry := false ].
stepContents ifNil: [ stepContents := '' ].
	
deferredClear ifTrue: 
	[ deferredClear := false.  
	stepContents := ''.
	self changed: #clearText. ].

deferredEndEntry ifFalse: [ ^self]. 
deferredEndEntry := false.  	
self critical: 
	[ stepContents := stream contents.
	stream resetContents.].
self changed: #appendEntry.	
```
is more clumsy than 
```
deferredClear ifNil: [ deferredClear := false ].
deferredEndEntry ifNil: [ deferredEndEntry := false ].
stepContents ifNil: [ stepContents := '' ].
	
deferredClear ifTrue: 
	[ deferredClear := false.  
	stepContents := ''.
	self changed: #clearText ].

deferredEndEntry ifTrue: 
	[ deferredEndEntry := false.  	
	self critical: 
		[stepContents := stream contents.
		stream resetContents ].
		self changed: #appendEntry ].
```

David5i6:
Yes I agree. I let this Issue as closed.

PR #5741, opened 2020-02-23
Compiler deprecate #methodMap

MarcusDenker:
Deprecate #debuggerMap: we can call these methods directly on the Context. Out "Map" is then natural meta object of a method: The AST after name analysis. It has all the information, we do not contruct any map. 

- deprecate #debuggerMap with a description what to do 
- remove the DebuggerMethodMapOpal class which just provided a backward compatible API.


src/Debugging-Core/CompiledCode.extension.st
```diff
@@ -43,7 +43,7 @@ CompiledCode >> abstractBytecodeMessagesFrom: startpc to: endpc do: aBlock [
 
 { #category : #'*Debugging-Core' }
 CompiledCode >> debuggerMap [
-	^self compilerClass debuggerMethodMapForMethod: self.
+	^self deprecated: 'call the methods on Context and directly!'
```
guillep, line 46: For somebody falling into this method, it's not clear what are the methods to use instead.

src/Debugging-Core/InstructionStream.extension.st
```diff
@@ -17,7 +17,7 @@ InstructionStream >> atEnd [
 
 { #category : #'*Debugging-Core' }
 InstructionStream >> debuggerMap [
-	^self method debuggerMap
+	^self deprecated: 'call the methods on Context and directly!'
```
guillep, line 20: Same same

guillep (changes requested):
The deprecation comment is confusing...

Also it is not clear to me why this needs deprecation, and not just a direct removal :).
Is the debugger map a public concern? Or is it a private detail of the debugger?

I understand that we don't syntactically differentiate public from private methods, but that difference exist nevertheless.

guillep:

MarcusDenker:
So I once just blankly removed that method as you suggest, but that broke the CI testrunner (the travis one).

I would like to deprecate it for that at least for a little so people have it easier if if brakes. But I would not keep it for a whole release, just some week or so and then remove.

You are right that the information is not enough, I will update it with a list of methods to use.

MarcusDenker:
methods are:

tempNamed: name in: aContext -> aContext tempNamed: name

tempNamed: name in: aContext put: aValue ->  aContext tempNamed: name put: aValue

tempNamesForContext: aContext ->  aContext tempNames 

rangeForPC: aPC contextIsActiveContext: aBoolean  -> someContext pcRangeContextIsActive: aBoolean

I will update the PR with this

guillep:
> So I once just blankly removed that method as you suggest, but that broke the CI testrunner (the travis one).

Maybe there were senders? There should not be anymore, right?

> I would like to deprecate it for that at least for a little so people have it easier if if brakes. But I would not keep it for a whole release, just some week or so and then remove.

Ok, I'll open an issue for that as soon as this is integrated.

> You are right that the information is not enough, I will update it with a list of methods to use.

As soon as you push that I'll merge.

MarcusDenker:
I will close this PR and do a new one where I will do the deprecation on the level of the DebuggerMethodMapOpal methods.

PR #5742, opened 2020-02-23
Compiler: remove old deprecated Compiler API

MarcusDenker:
finally remove the old API. all the method where deprecated since a while, but we have to remove then eventually... lets do that now


Ducasse:
Marcus do you think that we should add the deprecated method to Migrator?

PR #5760, opened 2020-02-24
5759 deletion of rg behavior from rg environment using calypso browser fails

carolahp:
Introducing the usage of environment in Refactoring classes to allow the deletion of Ring Classes and Ring Methods from the Calypso browser.


carolahp:
The solution fails when eliminating a class with subclass. Can not integrate yet

pavel-krivanek:
the Ring changes look ok

src/SystemCommands-RefactoringSupport/SycRefactoringPreview.class.st
```diff
@@ -257,9 +257,12 @@ SycRefactoringPreview >> updateChanges [
 SycRefactoringPreview >> updateTablePresenter [
 	| aCompositeChange |
 	aCompositeChange := RBCompositeRefactoryChange new.
+	aCompositeChange onSystemDictionary: 
+		(changes at:1 ifAbsent: [ self error: 'No changes to apply' ]) model environment.
```
guillep: This is strange.
Why not `changes ifEmpty:`?

Why the validation should be here?
I have the impression that before this worked even without changes, and now you added a validation so if there are changes, this fails.
Is that ok?

src/Ring-Core/RGEnvironmentAnnouncer.class.st
```diff
@@ -116,7 +116,7 @@ RGEnvironmentAnnouncer >> methodAdded: aMethod [
 RGEnvironmentAnnouncer >> methodRemoved: aMethod [ 
 
 	"TODO:"
```
guillep: Can re remove the TODO comment?
It does not say what is there to TODO, so it's pretty useless :)

guillep: Also maybe your change is what was there to TODO? :)

guillep (commented):
@carolahp can you check the comments I made?
Thanks

src/SystemCommands-RefactoringSupport/SycRefactoringPreview.class.st
```diff
@@ -256,10 +256,13 @@ SycRefactoringPreview >> updateChanges [
 { #category : #update }
 SycRefactoringPreview >> updateTablePresenter [
 	| aCompositeChange |
+	
+	changes ifEmpty: [ self error: 'No changes to apply' ].
+	
 	aCompositeChange := RBCompositeRefactoryChange new.
-	aCompositeChange onSystemDictionary: 
-		(changes at:1 ifAbsent: [ self error: 'No changes to apply' ]) model environment. 
+	aCompositeChange onSystemDictionary: (changes at:1) model environment.
```
guillep: the `at:1` could be replaced by `#anyOne`.

It would be nice to add a comment explaining why.
I understand now that the RBCompositeRefactoryChange requires an environment to work. And I assume it is not available? So you get it from one of the changes? (and this assumes all changes are on the same environment?

Did I get it right?

guillep (changes requested):

src/Ring-Core/RGEnvironmentAnnouncer.class.st
guillep: Thanks!

src/SystemCommands-RefactoringSupport/SycRefactoringPreview.class.st
carolahp: Exactly, I was writting a reply for your previous comment but you were faster haha.
Now I removed the error signaling in the last version of the method, because if there are no changes, the CompositeChange won't need an environment anyway

carolahp: > the `at:1` could be replaced by `#anyOne`.

Done ;)

guillep (approved):
ok. I think these are good enough now.

Thanks!!

PR #5761, opened 2020-02-25
Update Clap

cdlm:
Follow API changes so that people following the booklet don't hit walls

### DIFF
--- a/src/BaselineOfClap/BaselineOfClap.class.st
+++ b/src/BaselineOfClap/BaselineOfClap.class.st
@@ -8,14 +8,21 @@ Class {
 BaselineOfClap >> baseline: spec [
 	<baseline>
 	spec for: #common do: [ spec
+		baseline: 'Okay' with: [ spec repository: 'github://cdlm/okay-st' ];
+
 		package: 'Clap-Core';
 		package: 'Clap-CommandLine' with: [ spec requires: #('Clap-Core') ];
-		package: 'Clap-Tests' with: [ spec requires: #('Clap-Core' 'Clap-Examples') ];
-		package: 'Clap-Examples' with: [ spec requires: #('Clap-CommandLine') ];
 		package: 'Clap-Commands-Pharo' with: [ spec requires: #('Clap-CommandLine') ];
+		package: 'Clap-Examples' with: [ spec requires: #('Clap-CommandLine') ];
+		package: 'Clap-Examples-Booklet' with: [ spec requires: #('Clap-CommandLine') ];
+		package: 'Clap-Tests' with: [ spec requires: #('Clap-Core' 'Clap-Examples') ];
+		package: 'Clap-Okay-Tests' with: [ spec requires: #('Clap-Core' 'Okay') ];
 
 		group: 'default' with: #(core development);
 		group: 'core' with: #('Clap-Core' 'Clap-CommandLine');
 		group: 'pharo' with: #('Clap-Commands-Pharo');
-		group: 'development' with: #('Clap-Tests' 'Clap-Examples' 'Clap-Commands-Pharo') ]
+		group: 'examples' with: #('Clap-Examples');
+		group: 'tests' with: #('Clap-Tests');
+		group: 'development' with: #(core pharo examples tests);
+		group: 'development-full' with: #(development 'Clap-Examples-Booklet' 'Clap-Okay-Tests') ]
 ]
--- a/src/Clap-Core/ClapContext.class.st
+++ b/src/Clap-Core/ClapContext.class.st
@@ -16,28 +16,40 @@ Class {
 		'stdio',
 		'match',
 		'leftovers',
-		'session'
+		'session',
+		'exitException'
 	],
 	#category : #'Clap-Core-Activation'
 }
 
 { #category : #accessing }
 ClapContext class >> defaultRoot [
-	^ (ClapRoot withName: 'clap')
+	^ (ClapRoot id: #clap)
 		description: 'Entry point for commands implemented with Clap';
 		add: ClapFlag forHelp;
 		add: ClapCommand forHelp;
+
+		add: ((ClapFlag id: #save)
+			description: 'Save the image after the command terminates');
+		add: ((ClapFlag id: #noSave)
+			description: 'Do not save the image even if the command is stateful');
+		add: ((ClapFlag id: #rename)
+			description: 'Rename the image as <NAME> before running the command';
+			add: (ClapPositional id: #NAME));
+
+		add: ((ClapFlag id: #keepAlive)
+			description: 'Keep the image running after the command terminates');
+
 		meaning: [ :args |
-			args atName: 'help' ifFound: [ :help |
+			args at: #helpFlag ifPresent: [ :help |
 				help value; exitSuccess ].
 			args validateAll ];
 		yourself
-	"setup default flags, omnipresent subcommands & basic meanings here"
 ]
 
 { #category : #accessing }
 ClapContext class >> pragmaCommands [
-	^ (PragmaCollector filter: [:prg | prg selector = 'commandline']) reset
+	^ (PragmaCollector filter: [:prg | prg selector = #commandline]) reset
 		collect: [ :pragma |
 			| theClass theSelector |
 			theClass := pragma method methodClass.
@@ -56,23 +68,6 @@ ClapContext class >> withAll: commandCandidates [
 { #category : #'instance creation' }
 ClapContext class >> withPragmaCommands [ 
 	^ self withAll: self pragmaCommands
-]
-
-{ #category : #activation }
-ClapContext >> activateWith: args [
-	^ self
-		activateWith: args
-		onExit: [ :exit | self handleExit: exit ]
-]
-
-{ #category : #activation }
-ClapContext >> activateWith: args onExit: exitBlock [
-	^ [ self
-			rememberSession;
-			interpret: args
-	]
-	on: Exit
-	do: exitBlock
 ]
 
 { #category : #validation }
@@ -116,6 +111,11 @@ ClapContext >> context [
 	^ self
 ]
 
+{ #category : #streams }
+ClapContext >> defaultStdio [
+	^ ClapPluggableStdio onByteArrays
+]
+
 { #category : #matching }
 ClapContext >> doMatch [
 	| args |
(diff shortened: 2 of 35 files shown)


MarcusDenker:
CI did not run, I will close and reopen to trigger it (this sometimes is needed of one changes the branch)

MarcusDenker:
Tests are failing, but only on windows (e.g. ClapPharoVersionTest).

guillep (changes requested):
Would be nice to have a changes log because this is more than just API changes...

Ducasse:
We should introduce some deprecating methods too.

cdlm:
Right… API changes:
- parameter specifications have a unique identifier used for lookups (instanciate with `id:` instead of `withName:`; I added a rewriting deprecation for that one)
- the user-visible name of a parameter is derived from the identifier, but can also be independently set
- looking up matches is now unified in a single family of methods (`at: identifierOrSpec` etc instead of `atName: aName` and `at: aSpec`; maybe a deprecation for that one would be nice as well)
- restructured the Matches hierarchy (but I think it's mostly transparent since user code should not have hard references to those classes)

Non-API changes:
- Commands are nicer to run from a workspace, because the stdio streams are captured and the image won't quit. The command-line handler explicitly configures the context so that things work as expected when invoked from an actual command line.
- includes some WIP implementations of new commands
- `ClapApplication` provides some boilerplate for implementing commands (because the command spec is on class side, and it's nicer if the bulk of the behavior is instance side)

Deprecations: why not, but IMHO it's more pressing to think about the migration from CommandLineHandler than between alpha versions of Clap when we're probably the only ones having to migrate anything. Just my ¢2…

guillep:
> Deprecations: why not, but IMHO it's more pressing to think about the migration from CommandLineHandler than between alpha versions of Clap when we're probably the only ones having to migrate anything. Just my ¢2…

Maybe, but if these are alpha versions, why integrating them?

cdlm:
Okay, call them beta or preview 😛

Ducasse:
I think that esteban used clap and it is not alpha anymore. 
What it needs is more users and this is what I tried to do when I hurt the wall of the missing documentation and now I'm retrying with Pillar.

Ducasse:
Damien can you check because the windows tests look to exhibit a pattern based probably on a cr or lf.

cdlm:
Indeed it's using `ZnNewLineWriterStream` but the tests expect `lf`. The quick and dirty fix is to change the tests to expect the platform convention, but… yuck…

Really the line end convention should be configurable on the ClapContext 🤔

guillep:
You could also fix the new line convention in the `ZnNewLineWriterStream` just for the test.

image

Or, you could use the platform line ending for your test.

image

cdlm:
Ideally `ClapContext` would be the configuration point for line end convention but trying to do that quickly becomes kludgy 😒 Probably the clean way is to not rely on `Stdio` but set the three streams individually.

So I guess platform convention it is, for now…

cdlm:
I don't understand if Jenkins has run or not :-(

cdlm:
(force-push because I started from a new image and didn't want to include 200 unrelated commits…)

src/Clap-Tests/ClapCodeEvaluatorTest.class.st
```diff
@@ -0,0 +1,38 @@
...
+		equals: 'SmallInteger' , self lineEnding
+]
+
+{ #category : #tests }
+ClapCodeEvaluatorTest >> testEvaluatePipeIn [
+	context := (ClapCodeEvaluator evaluate activationWith: #('evaluate'))
+		setStdio: (ClapPluggableStdio onByteArraysWithInput: 'Object class class class class');
+		execute.
+	
+	"self assertSuccess."
```
guillep: Is this a leftOver?

src/BaselineOfClap/BaselineOfClap.class.st
```diff
@@ -8,14 +8,21 @@ Class {
 BaselineOfClap >> baseline: spec [
 	<baseline>
 	spec for: #common do: [ spec
+		baseline: 'Okay' with: [ spec repository: 'github://cdlm/okay-st' ];
```
guillep, line 11: This is not being integrated no?
I don't see any "Okay" code. Is that ok?

Then we can discuss if we want another framework or not for testing. We have already discussed in the past about introducing mocking frameworks in the base image.

src/BaselineOfClap/BaselineOfClap.class.st
```diff
@@ -8,14 +8,21 @@ Class {
...
+
 		package: 'Clap-Core';
 		package: 'Clap-CommandLine' with: [ spec requires: #('Clap-Core') ];
-		package: 'Clap-Tests' with: [ spec requires: #('Clap-Core' 'Clap-Examples') ];
-		package: 'Clap-Examples' with: [ spec requires: #('Clap-CommandLine') ];
 		package: 'Clap-Commands-Pharo' with: [ spec requires: #('Clap-CommandLine') ];
+		package: 'Clap-Examples' with: [ spec requires: #('Clap-CommandLine') ];
+		package: 'Clap-Examples-Booklet' with: [ spec requires: #('Clap-CommandLine') ];
+		package: 'Clap-Tests' with: [ spec requires: #('Clap-Core' 'Clap-Examples') ];
+		package: 'Clap-Okay-Tests' with: [ spec requires: #('Clap-Core' 'Okay') ];
```
guillep, line 19: Same as above

src/Clap-Core/ClapContext.class.st
```diff
@@ -16,28 +16,40 @@ Class {
...
 				help value; exitSuccess ].
 			args validateAll ];
 		yourself
-	"setup default flags, omnipresent subcommands & basic meanings here"
 ]
 
 { #category : #accessing }
 ClapContext class >> pragmaCommands [
-	^ (PragmaCollector filter: [:prg | prg selector = 'commandline']) reset
+	^ (PragmaCollector filter: [:prg | prg selector = #commandline]) reset
```
guillep, line 52: Be careful, the pragma collector is super slow in large images.
I think this would be ok for now, but we should avoid using it in the mid-term.

src/Clap-Core/ClapMismatch.class.st
```diff
@@ -9,6 +9,7 @@ Class {
 
 { #category : #accessing }
 ClapMismatch >> at: aSpec [
+self flag: 'identifierOrSpec?'.
```
guillep: indentation?

guillep (changes requested):
There are two little details that could be fixed fast.
I think for the rest is good.

Test failures are unrelated, I'm working on that.

src/BaselineOfClap/BaselineOfClap.class.st
cdlm: It's not integrated, I refactored the baseline so that the default and development groups don't include either Okay or the booklet examples.

Okay is intended to be an assertion / extreme validation framework. But maybe it's better to port Andres's Assessments or use https://github.com/seandenigris/Validation-Revisited
…and Exceptions need some work to make them properly nestable too

src/BaselineOfClap/BaselineOfClap.class.st
cdlm: same as above.

src/Clap-Core/ClapContext.class.st
cdlm: How do we use pragmas then?

src/Clap-Core/ClapMismatch.class.st
cdlm: …it's a flag 😒

More seriously the problem here is that I need to check that the parameter makes sense wrt. the receiver's specification; otherwise it's a programmer error.

guillep: I'm not criticising the code, I'm saying it should be properly formatted, at least indented, that's all :)

src/BaselineOfClap/BaselineOfClap.class.st
guillep: ok

src/BaselineOfClap/BaselineOfClap.class.st
guillep: > It's not integrated, I refactored the baseline so that the default and development groups don't include either Okay or the booklet examples.

ok

src/Clap-Core/ClapContext.class.st
guillep: I don't have a proper answer for that yet.
The pragma collector does a full scan of the system, which is not good in large systems.
Moreover, several usages on the system were re-scanning everything on each method compilation.
As I said, I'm ok with this for now, but we should think an alternative in the mid term.

For example, we should be able to scope the pragma collector on a hierarchy.
That will be already something.
Another thing would be to use a pragma collector that updates itself with announcements, so paying the full-scan price only once.

Maybe @tesonep has more input.

cdlm: I think we agree that pragmas are here to stay, and even if they need work, it's ouf of the scope of this PR?

guillep: That's what I said. What is there to argue?
Let me re-quote myself:
> I think this would be ok for now, but we should avoid using it in the mid-term.

cdlm: That can be read as *Clap might have to pivot to another completely different mechanism in the future* ; of course `PragmaCollector` might go away, but I'm pretty sure there will still be some convenient API to find all methods with a given pragma, so I expect the change to be relatively simple as far as Clap is concerned.

If required, Clap specifications could be built once and installed in some static global (which is one of the reasons I don't like the meaning blocks closing over the builder, in the pillar command)

guillep: > That can be read as _Clap might have to pivot to another completely different mechanism in the future_

And that's what I meant. Where "in the future" = in the following months, very probably during Pharo 9.

What I meant is "be careful, this may change soon, but for this PR is ok".
Is that ok?

src/Clap-Tests/ClapCodeEvaluatorTest.class.st
guillep: This should be removed.

cdlm: The other tests do that assertion (that the exit status is 0); I uncommented it

guillep (approved):
Test failures are unrelated.

Thanks Damien for pushing!

guillep:
BTW, just for documentation purposes, what is the clap commitish that was integrated? In case we want to diff/merge later?

(or the tag, if any)

cdlm:
What you just merged should be identical to https://github.com/cdlm/clap-st/commit/f9be1e3ce33776cde83eaedbc7889ebbc86c2eda (latest commit to master now, and what Iceberg says in the image I used to prepare the PR). Just tagged as [pharo-pr5761](https://github.com/cdlm/clap-st/tree/pharo-pr5761).

PR #5764, opened 2020-02-25
introduce literal method concept and use it to skip some tiime-consum…

estebanlm:
…ing evaluations


MarcusDenker:
failing tests not related

MarcusDenker (approved):

PR #5770, opened 2020-02-26
Revert "5724-Improve-AdditionalMethodStaterefersToLiteral"

MarcusDenker:
Reverts pharo-project/pharo#5731

This was not a good change: semantics are a bit odd: for Bindings, we want to only look into them deeply if they are in a property. Else not. I will revert it and put a comment.

(it does not add a bug this way but it is even more confusing as before)

### DIFF
--- a/src/Collections-Support/LookupKey.class.st
+++ b/src/Collections-Support/LookupKey.class.st
@@ -32,13 +32,6 @@ LookupKey >> = aLookupKey [
 		ifFalse: [^false]
 ]
 
-{ #category : #testing }
-LookupKey >> hasLiteral: literal [
-	^ self key == literal
-		or: [ self value == literal
-				or: [ self value isArray and: [ self value hasLiteral: literal ] ] ]
-]
-
 { #category : #comparing }
 LookupKey >> hash [
 	"Hash is reimplemented because = is implemented."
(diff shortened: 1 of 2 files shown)


guillep:
Would be nice to have tests to enforce the desired semantics, which seem odd to me...

MarcusDenker:
I found that odd, too.

There are tests (for the properties case). This is how I found it. I thought "no, that code is wrong, we never want to look at a value of a VariableBinding for the literals" --> boom, test that checks for it fails.

MarcusDenker:
I will out this on "need more work" as I am tempted to get rid of the strange semantics... we should *not* look into globals if the are the the value of a method property!

guillep:
So this PR gets rid of a bug? If so, It has my approval.
I prefer to go back to a state where the code was ugly but we had less bugs, than to stay with a bug that we can avoid easily. The semantics are strange in both cases, so there is no real improvement there :).

guillep (approved):
I prefer to go back to the working version.
The code is "ugly", but it is a single method, nothing we cannot deal with.

guillep:
I could integrate it, I just need confirmation from @MarcusDenker

MarcusDenker:
yes, we should merge this revert

guillep:
Thanks Marcus!

PR #5780, opened 2020-02-27
Using Image relative font files can not work

### LINKED ISSUE #5776: Using Image relative font files can not work
Because this method uses the image path (which is including the name) and is a **file** (not a **folder**) and tries to add "Fonts" subfolder.

image

The fix is obviously

image

astares:
Fix #5776


MarcusDenker:
error not related

MarcusDenker (approved):

PR #5782, opened 2020-02-27
Fix CalypsoBrowser window grouping

stacksmith:
ClyBrowserMorph>>openAnotherBrowser now checks for caller being inside a group-windowed environment, and if so, opens all new browsers inside; otherwise, a new window is created.
* This is consistent with ClyBrowserMorph>>close (note how it tries to see if its owner is a GroupWindowMorph to correctly close the browser).
* This eliminates the need for ClyQueryBrowserMorph to override openAnotherBrowser and do roughly the same, except klugier.

as it stands, Calypso is unusable in group windows and cannot be integrated or embedded.

### DIFF
--- a/src/Calypso-Browser/ClyBrowserMorph.class.st
+++ b/src/Calypso-Browser/ClyBrowserMorph.class.st
@@ -504,7 +504,9 @@ ClyBrowserMorph >> open [
 
 { #category : #'opening/closing' }
 ClyBrowserMorph >> openAnotherBrowser: aBrowser [
-	aBrowser open
+	"If embedded into a grouped-window environment, openInWindow: groupwindow.. "
+	(self window ownerThatIsA: GroupWindowMorph) ifNil: [^ aBrowser open ].
+	aBrowser openInWindow: self window.
 ]
 
 { #category : #'opening/closing' }
--- a/src/Calypso-SystemTools-QueryBrowser/ClyQueryBrowser.class.st
+++ b/src/Calypso-SystemTools-QueryBrowser/ClyQueryBrowser.class.st
@@ -262,12 +262,6 @@ ClyQueryBrowser >> newWindowTitle [
 	^title
 ]
 
-{ #category : #'opening/closing' }
-ClyQueryBrowser >> openAnotherBrowser: aBrowser [
-	(aBrowser isKindOf: ClyQueryBrowser)
-		ifTrue: [ aBrowser openInWindow: self window]
-		ifFalse: [ aBrowser open ]
-]
 
 { #category : #initialization }
 ClyQueryBrowser >> packageNameOf: aBrowserItem [


dionisiydk:
The current behaviour is intended. And as nobody complained I hope it was accepted by people.

The idea was that the query browser should always open queries in tabbed mode. It reduces the explosion of windows during navigation workflow when many senders/implementors are queried subsequently. In that scenario the full browser request from the user (the browse button) is considered as the end of navigation and therefore it is opened in new separate window. 

If you do not like this behaviour please make it configurable instead of removing current approach completely. This way we will have more settings and all will be happy

stacksmith:
I understand.  I am trying to control the explosion of browsers as well!  I assumed that the weird embedding behaviour was accidental.

I think it's a mistake to leave it as is, because it is inconsistent and makes it impossible to keep browsers spawning into the same group.  It is also riddled with bugs.

 I think people are not complaining because they are happy to have at least some windows combining - and would be even happier with more composability.  Routinely keeping 20 browsers open in different windows is precisely why I stopped what I was doing and worked on this.

This small fix changed my life and I now work in a couple of windows.  This is so much better than sea-of-browsers, and a hell of a lot better than wearing GlamorousToolkit handcuffs.

I would hate to leave it as is, as it completely destroys my workflow.  Having a new browser for every visited class has made my life unbearable and caused me to move away from Smalltalk several times now.  I'd be sad if we lose others for 3 lines of code.  It is a lot easier to move a browser out of a group than try to organize a mess of dozens of browser windows with new ones constantly popping out.

I agree that configurability of this feature is a better in the long run, and am happy to take it on.  Would you suggest the best way to accomplish it while fitting within current and future goals of Pharo?  Is there a forward-looking strategy for dealing with flag-type configuration settings?

dionisiydk:
Maybe the best would be to keep current logic but force tabs in any condition when current grouped window was created explicitly by user.

stacksmith:
Could you confirm that my understanding of desired behaviour is correct?
**Query browsers spawn other query browsers in their own group; All other browsers spawn in new windows.**
New strategy:
- Groups keep a flag which is set only by standalone query browsers spawning a new query (legacy flag).
- All other groups - such as user groups - do not have such a flag.
This allows the spawning logic to check if the group environment should eject new full browsers (current behaviour) or keep them inside.

This should preserve current behaviour while allowing the user to group browsers and contain their progeny (and avoid having their desktop look like granpa Bogdan's house shortly before they took him away).

stacksmith:
Also looking at 'spawnedFrom' ...

Ducasse:
Hi Stacksmith 

First thanks!
Second I can tell you that we are not happy with many of the current workflow but it is often difficult to propose something better. So we are interested in any improvement. 
Third Pharo interaction and workflow does not have to be like it is now. We should reinvent it. 
Four we are working on new debugger and better experience but this is not that easy. 

For the record, this is why we are reimplementing all the tools (esteban wants to be able to code without a mouse and I understand him) and we will deprecate Glamour. 
If you want to help let us know. 

We will probably reimplement calypso too because it is not layered and cannot work in GTK and other back-ends. It was not on our roadmap (and we have something else to do but we will have to do it). 

Finally right now I'm dead after 10 hours of debugging/coding and sprinting. So I will come back to this issue when my brain let me think :)

Ducasse:
stacksmith you can also have a look how to manage setting. This is relatively simple: often define a class var with the default setting, describe it with the <setting> prag, and use an instance variable (if needed - to have instance based but default sharing for the settings) initialized from the class var value to implement the setting.

Ducasse:
I worked on a window garbage collector and it was fun but I could never put my hands on the code of the student. Should reimplement it.

stacksmith:
@Ducasse:
Thank you for jumping in.  We are completely on the same frequency - I am really excited as I was a little scared that Pharo is spinning off into a weird (and not usable by me) direction.  Yes to every point you made!

 **Interaction**.  I've abandoned Smalltalk a few times because of the terrible mess of browsers which does not work with my ADD-riddled mind.  On the other hand, Glamour approach is possibly worse - everything I want is constantly scrolling out of view!  

**Minimizing Browser Hell with a minimal effort**
My 3-line fix is 90% of what I consider a reasonable and consistent work environment.  I contain everything into a group.  No more mess!  Some problems need to be fixed for the last 10%:
- Browser state is really ambiguous. I would love to be able to get the browser state as a compact object with class/method or a list of these for a query, and be able to quickly create a browser on that...  Then tabs could just hold those instead of full instancs of browsers.
- I find myself losing browser positions!  Consider this common situation: I find a method I need to hold for later and need to look at another method in the same class.  Bam, my old method is gone.  If I use history to back up, I lose the new one.  There is no 'spawn a copy' button or key that I know of!  So I have to look at the browser and depending on what it s, click on a method or class or whatever, and create a new copy that way - and it is usually a little different...
- Tabbing/notebook infrastructure is inadequate.  I need a GroupWindowMorph that can go vertical as well as horizontal (our screens are wide!).  I need a few views for tabs - sorted, organized arbitrarily (with the ability to move them around freely!), or as a tree control by class and tag hierarchy.  Think bookmarks in a web browser.
- I really need to take and keep notes, in addition to the class comments.  
- one or more attached workspaces with all those bits of wisdom found during research and trial-and-error experimentation.

This minimalist approach is browser-agnostic and will work well, serving the role of a project/ide for most of what we need.  There are better ways, but it's hard to come up with good ones that don't force you into a rigid structure.

I have a few ideas for an improved workflow based on decades of observing myself and others in several different image-based languages (Smalltalk, Lisp, Forth) .  I would love to move this conversation to a more appropriate venue as we are getting off-topic.
Thanks.

dionisiydk:
> Could you confirm that my understanding of desired behaviour is correct?

My suggestion is:
- keep current logic as default behavior
- switch given window to new behavior only when user explicitly Create Window Group (from window menu).

So the command Create Window Group is a user desire to force tabbed mode. So no other windows should be opened from such "custom window". But until user executes this command on given window the navigation will continue follow the current spawning logic.
Does it make sense?

In addition you can add settings "always spawn full browser in new window" which is true as current default behavior. When disabled the full browser will go to current group instead of new window (Browse command from senders window will add new tab with full browser).

stacksmith:
@dionisydk:  I understand. 
Do you personally find this behaviour desirable?  I can't imagine that anyone does.  It is arbitrary and surprising.  Being spawned from a group is in fact enough of an indication that you want the result in the same group.  Spawning from a standalone window is an indication that you like standalone windows.  If there are settings, there should be a setting for not behaving in an obvious, least-surprising method - not pandering to a couple of people who got used to some arbitrary behaviour over a year or so.  We need to look into the future, and be able to explain how this stuff works to people who haven't yet developed bad habits.

Personally, I did not even bother to figure out why some windows spawn into their own groups.  It seemed kind of random, but I figured someone would fix it at some point - it's clearly a bug.  As it occupied less space than if it had behaved correctly, I did not complain - in fact I wondered why the rest of the browsers don't go into the group too!  I suspect everyone else who's ever bothered to pay attention to this feels the same.  

In short, **I respectfully disagree, and think it is just plain wrong.**  It is inconsistent with Spec, makes it impossible to embed windows and is full of bugs.

While I could implement what you suggested, it violates good design principles by giving contained objects control over their container, and adds arbitrary setting and more complexity to an already-too-complex settings system, for no reason I can personally justify.

I recommend fixing the behaviour according to **the principle of minimal surprise** that requires no extra settings.  **Restoring correct behaviour is - +3/-7 lines**, as minimal a commit as possible and accomplishes, without any additional effort:
- eliminating a special case in the query morph (flagged by the critic, btw) ;
- moving control of high-level container semantics out of contained entities;
- simplifying development of future derivative browsers;
- normalizing spawning behaviour of all browsers (and making it possible to style an IDE) [note: there are a couple of other spawning bugs lurking];
- minimizing the possibility of introducing bugs. 

If a setting is to be added, it should be one to turn on the irrational 'legacy mode' that arbitrarily groups only query browsers spawned by other query browsers while always spawning independent full browsers - a distinction I find odd and distasteful.

stacksmith:
**Another bug: Browser grouping is broken!**

See issue 5819 (click below).

This is indicative of the kind of trouble you get into for doing weird and inconsistent shit with grouping! 

I can't believe how much time I have to spend to convince you to do the right thing - considering the result is elimination of bugs, dead and/or weird code, violations in containment hierarchy, and inconsistencies and surprises in the UI.

dionisiydk:
> @dionisydk: I understand.
> Do you personally find this behaviour desirable? 

Well I implemented it. So yes it is very desirable by me.
it was my attempt to reduce the amount of windows without explicitly requiring user to create window group. Actually the tabs approach during code queries is similar to browsers in other smalltalk (Cuis maybe?) where it shows the tree of navigation. Tabs was just much simpler to implement and it has its own advantages.

Now with you PR we will get window explosion back because when you do the incidental code research using senders/implementator navigation you are not going to click "Create Window Group". This action is kind of break in user workflow. So I see that the problem with your approach is the requirement to switch to groups in advance.

Please, notice it is my opinion. And I am fine if majority users will support you.

> I can't imagine that anyone does. It is arbitrary and surprising. Being spawned from a group is in fact enough of an indication that you want the result in the same group. Spawning from a standalone window is an indication that you like standalone windows. If there are settings, there should be a setting for not behaving in an obvious, least-surprising method - not pandering to a couple of people who got used to some arbitrary behaviour over a year or so. We need to look into the future, and be able to explain how this stuff works to people who haven't yet developed bad habits.

stacksmith:
I can see that I offended you - I did not mean to.  Nothing personal - there are bugs to be fixed we must make behaviour consistent.

What we have is buggy, requires special cases, is hard to describe, makes improvements and IDEs nearly impossible, and breaks containment.  It sucks in every category imaginable.

I have a lot of respect for all the hard work you've done!  Let's not screw it up over 3 lines of code that break all rules of logic and sensibility, and cause bugs.  As I said, the entire ClyGroupWindowMorph can now be eliminated.  That is great news!

stacksmith:
I fixed the bugs, eliminated ClyWindowGroupMorph and it seems to be working great.
[Edited changelog](https://gist.github.com/stacksmith/fbddc86c9658fbcd2edc2c5e1ca3ca8c#gistcomment-3194514) if anyone wants to try.

This requires further commits (and a repair to WindowGroupMorph).  Given the friction I am experiencing here, I am not sure how to proceed.

If you think a bug-free, consistent and embeddable Calypso is not important - that is OK with me.  If lack of complaints is sufficient to not fix glaring bugs - respectfully, I am complaining.  I would love to have your help - and I feel you owe it to your users - but whatever.  Spasibo.

dionisiydk:
If there are bugs they need to be fixed. But it has no relation to the feature which you do not like.
I do not see any inconsistency in behavior. QueryBrowser just opens new queries in tabs in own window. Nothing random here. 

About ClyGroupWindowMorph. I agree it needs to be merged into superclass. At time when Calypso was separately loaded project the subclass was the only way to allow browser work in current Pharo version. Now we can improve superclass to cover missing features and get rid of this hook.

But again such changes are nothing to do with current behavior. We would be happy to merge fixes and code improvements. No complains there. But be ready for that if you change existing features.

stacksmith:
Whatever.

dionisiydk:
> If you don't think there are bugs, close this if you have the authority.

I never told there are no bugs. But removing feature where you found them is not a fix.

>  I honestly don't enjoy fighting with fools.

Sure, nobody does. In this discussion you are the only person who uses such rude language. It will not give you any credit.

> 
> However, violating containment principles is a bug. Also writing code like this (more like a crime):

You can always propose refactoring. As I said it is more than welcome. 

> 
> ```
> ClyAccrossWindowNavigationState >> restoreBrowserWindow
> 	"The ugliest thing I've ever seen!"
> 	| newBrowser |
> 	window isInWorld ifFalse: [ 
> 		"Closed browser window is always cleaned from browser. Look at ClyBrowser>>windowIsClosing"
> 		newBrowser := self createBrowser.
> 		window addMorph: newBrowser frame: (0@0 extent: 1@1).
> 		window model: newBrowser.
> 		browserState applyTo: newBrowser].
> 	windowGroup 
> 		ifNil: [
> 			window isInWorld 
> 				ifTrue: [ window comeToFront; activate]
> 			 	ifFalse: [window openAsIs ]]
> 		ifNotNil: [ 
> 			windowGroup isInWorld 
> 				ifTrue: [ windowGroup window comeToFront ]
> 				ifFalse: [ windowGroup window openAsIs ].
> ...
> ```
> 
> I have no words.

Ducasse:
@stacksmith feel free to contact me at stephane.ducasse@inria.fr. 
Because I would love to see how we can build a great system. 
You should have a look at the Autumn leaves paper on http://stephane.ducasse.free.fr/Publications.html 

S. 

<blockquote><div><strong><a href="http://stephane.ducasse.free.fr/Publications.html">Dr. Stéphane Ducasse</a></strong></div></blockquote>

stacksmith:
@Ducasse - bravo on Autumn leaves.  I am having email issues and will get in touch with you as soon as I can.  The time is right for a better browser, or something like it.

My fix above provides an option to use Calypso with something like the Eclypse model, with the option of easily pulling windows out.  The limitations of the static tab arrangement are there, and I've been working on a vertical-tab control with a temporal heat-map coloring option...

Unfortunately I can't get this obvious domain violation fix merged as you can see above - 3 *^%#ing lines!  Someone is invested in browsers jumping in your face no matter what.

bencoman:
Hi @stacksmith, Perhaps not the best way to introduce myself to you, and I understand you are frustrated, but when I see someone in their first PR already swearing at other developers and calling them fools, and then disengaging with an immature "whatever" ... it doesn't leave me with a favourable impression to interact with you.   I realize that in sharing this I risk your animosity.  I hope you consider that easiest thing for me to do was do-nothing, but I'd rather help you to achieve a good outcome and gain another co-developer.   [Please read](http://catb.org/esr/faqs/smart-questions.html#keepcool).

Consider that what is super obvious to you may not be obvious to others.  Or maybe your PR causes a problem with @dionisiydk's workflow that you don't see in your workflow.   So I see two specific concerns raised by @dionisiydk that you did not respond to.
> Now with you PR we will get window explosion back     

> I see that the problem with your approach is the requirement to switch to groups in advance.

Now I don't have a good gasp of either position.  I'm not confident that if I tried the PR I would be looking at the right thing.   @stacksmith could you perhaps attach here a GIF-screen-captures so others can "see" the issue before/after.

stacksmith:
The goal of the PR (which is just a start) is to normalize window behaviour:  If you chose to put a window into a group, it stays there, does not jump out or lose its frame or turns white or crashes the computer.  And any windows that spawn - by browsing from those windows - open in the same group.  Obviously, by creating the group, you want to keep related windows there.

When you don't group - or pull a window out of a group - it should become a standalone window.  Not go blank, or take other windows down.  Not grab other window frames and make a mess of the screen and lock it up dead.  If you spawn another window by browsing from it - you telegraphed your intention to keep windows separate - which is what most people are used to.

And you should be able to, at any time, move any windows into groups, or move them to other groups or out.  You should also be able to group groups into groups, but that's a whole other story.

And there should not be groups that work almost like normal groups - without any visual clues - that take your windows just like normal but crash later!  That is just unforgivable.  But that's just scratching the surface.

Thank you for your patient approach.  I don't have much patience for this issue, but you seem like a reasonable person so I will try to explain.   From where I am sitting, 
1) terrible and foolish bugs prevented me from being able to use the system in what is clearly intended manner - being able to compose bunches of windows into one (or a few) - to organize the mess and generally contain things inside other things.
2) Furthermore, the buggy behaviour of arbitrarily deciding on forcing some but not other windows to jump out of a carefully arranged window group lead to crashes, misdrawn windows (in calypso windows going blank, losing frames, and locking up the system - all with stock, current software.  I am sure dionisiydk has some convoluted reason to have certain browsers group themselves while others jump out - but it makes the grouping system useless - worse than useless in fact because you wind up with an even bigger mess.  Try putting browsers into a group and pulling them out yourself.
3) In order to accomplish this magic, dionisiydk created versions of grouping master frames which are only slightly different from stock windows, but hold secret data and behave differently while visually appearing the same.  An entire class full of bugs serves no purpose but to make the system behave in a complicated and frustrating way - when it works as intended and does not crash.    If you accidentally drop a regular window into one of those, you have a timebomb.
4) Removing **7 lines and adding 3 -** my original PR - makes the system **work 90% correctly:** any browser may be placed into a group, where it will spawn new browsers into the same frame; you can always pull a window out, and it will spawn separate windows after that.  So it is entirely logical, consistent and eliminates much buggy behavior.
5) Dionisiydk has hidden some other surprises that I tried to address in my changeset (I am not even bothering with a PR).  It fixes a bug in the morphic group window (that destroys windows when pulling them out of the frame).  With that fix, **the entire window grouping class of Dionisiydk can be just thrown out  and the browsers work wonderfully** - I work with 2 or 3 windows set up just right, without blocking my workspaces and transcripts.  My life is just so much better from these minor changes.  Sadly, I keep finding other places where terrible unpredictable behavior sits deep inside Calypso. 
6) Speaking of terrible code, the entire structural framework of Calypso is pretty much useless.  I haven't seen anything so Lovecraftian in years - there is not one part of it that is remotely clean.  Apparently the idea of hierarchy, encapsulation, and separating models from behaviour has never occured to the writer.  The idiocy of the windows arbitrarily opening or grouping is just scratching the surface, really.
7) There is no accounting for taste, or clarity of thinking and lack thereof - and who am I to judge.  But when solution A has convoluted impossible to describe rules , breaks any possibility of doing things any other way, and crashes the system when you try - while solution B involves **3 lines of code and largely fixes the behavior to make it consistent and work as intended with any other windows - and eliminates bugs and entire classes of badly duplicated code, I think it's a win**.
Of course if you like weird, buggy and antisocial behavior in your workflow, and you wrote the code, you can keep it.  But maybe when there is a simple alternative that normalizes behaviour and fixes bugs - that should be the default behavior.  You know - windows that work like all other windows and don't jump out covering your work or for some reason group themselves.  So that other people can enjoy the system too.  And the default behaviour should be the simplest, most obvious and most predictable one with fewest bugs and hard to explain scenarios.   Anyone who does not get that should not maintain public repositories, write code, or maybe even try to button shirts.

Anyway, that is probably enough said.  I don't need to share my code - I am reasonably happy with my fixes until I can throw this garbage out.   At this point I am familiar enough with it to make escape plans.  It sort of works, but is really difficult to make work with other workflows or compose it in ways not specifically designed for.

Ducasse:
I really think that the proposed changes are reasonable and I will integrate them.

stacksmith:
Thanks.  There is a handful of additional changes [in this changelog](https://gist.github.com/stacksmith/fbddc86c9658fbcd2edc2c5e1ca3ca8c#gistcomment-3194514).  There is an important bugfix to the groupwindow Morph which deletes windows and their contents upon removal from a group.  Once that is fixed, an entire Cly group window class is eliminated.
I've been filing it in for a couple of weeks, and it's about as solid as Calypso was without it.  Which is not very.  Calypso has a number of windowing bugs still - after a few hours of heavy use windows get wonky, start moving or sizing uncontrollably, and affecting other windows and widgets.   I get daily lockups and uncontrolled image growth (it goes to a couple of gigs in a few days with no reasonable cause - all I do is hammer onto a bunch of calypso windows with a very small body of code - I file into a clean image and start over). there are a lot of hidden places where browsers are spawned with weird windowing semantics.

Ducasse:
I will work with your changes to see how it goes.

Ducasse:
Now each time I do senders of I get a new windows so this is what we expect. 
How can we have the best of both worlds?

dionisiydk:
Hi Stef.

I explained it in my comments but it was ignored with rude words that only
fools need it.

чт, 12 мар. 2020 г., 09:16 StéphaneDucasse <notifications@github.com>:

stacksmith:
May I suggest that the default behaviour is consistent one: spawn windows like the parent window.  

This means that D. should go through all spawning code and bring it in line instead of whatever it does now.  I can do it for him if he can't but it will take a while and somewhere there is always more hidden spawning code.

If D. wants to add a setting - why not - as long as he can come up win a one-or-two word description of what to expect when it's on. I am not smart enough to name the current behaviour.

stacksmith:
Here is another thought:

When spawning browsers interactively, default to spawning into parent group if it exists or standalone if not.  If SHIFT key is pressed, use a 'special' way to spawn, which is application-dependent. 

This way we can always expect working consistent and predictable behaviour, but should the developer provide some other way to do things, holding shift gets you there.

Ducasse:
Yes this could be a good solution. 

S

Ducasse:
So did you update your PR with your proposal?

Ducasse:
Ok I will close this PR.

PR #5785, opened 2020-02-28
Refactoring - Add remove class keeping subclasses 

### LINKED ISSUE #4932: New refactoring: Remove keeping subclasses
A new refactoring that could be cool for classes is to be able to remove a class but keeping the subclasses. 

The subclasses would have the superclass of the class to remove as the new superclass.

Lin777:
This refactoring remove a class "not empty" keeping subclasses.

If the class is "not empty" (has methods and variables), any subclass is reparented to the superclass of this class, and all its methods and variables (instance and class) are push down in its subclasses. For example: if we want remove ClassA

```
Object subclass: #ClassA
	instanceVariableNames: 'foo'
	classVariableNames: ''
	package: 'Example'
```
```
ClassA >> foo
    ^ foo
```
```
ClassA subclass: #ClassB
	instanceVariableNames: ''
	classVariableNames: ''
	package: 'Example'
```
We get this after refactoring
```
Object subclass: #ClassB
	instanceVariableNames: 'foo'
	classVariableNames: ''
	package: 'Example'
```
```
ClassB >> foo
    ^ foo
```

Fixes #4932


Ducasse:
This is cool to get more refactorings. 
Did you see that we are missing convert temp to instance variable https://github.com/pharo-project/pharo/issues/5796

Lin777:
I did not see that issue #5796 , I will try do that missing refactoring

Ducasse:
This looks good to me. I like that there are tests covering the new refactoring :)!
Great job!

PR #5786, opened 2020-02-28
5437 package names little queries

### LINKED ISSUE #5437: Add Package queries
We should add two query methods to `RPackageOrganizer`:

```smalltalk
organiser selectPackages: [ :each | ... ]
```

```smalltalk
organiser selectPackageWithName: [ :name | ...  ]
```

On top of those, users could easily build more complex queries (beginsWith, endsWith, substrings, regex-based...)

daniapq:
Add query methods with testing to fix #5437

### DIFF
--- a/src/RPackage-Core/RPackageOrganizer.class.st
+++ b/src/RPackage-Core/RPackageOrganizer.class.st
@@ -476,6 +476,13 @@ RPackageOrganizer >> includesPackageNamed: aSymbol [
 	^ packages includesKey:  aSymbol asSymbol
 ]
 
+{ #category : #queries }
+RPackageOrganizer >> includingWith: aString [
+	"Return list of packages whose names includes the argument as Substring"
+	
+	^ packages select: [ :p | p name includesSubstring: aString ]
+]
+
 { #category : #initialization }
 RPackageOrganizer >> initialize [
 	super initialize.
@@ -538,6 +545,13 @@ RPackageOrganizer >> initializeMethodsFor: aBehavior [
 			 (eachProtocol methodSelectors 
 				select: [ :eachSelector | (aBehavior >> eachSelector) origin = aBehavior ])
 				do: [ :eachSelector | package addMethod: (aBehavior >> eachSelector) ] ]
+]
+
+{ #category : #queries }
+RPackageOrganizer >> matching: aPatternString [
+	"Return the list of packages whose names match the argument."
+	
+	^ self packages select: [ :p | aPatternString match: p name  ]
 ]
 
 { #category : #accessing }
--- a/src/RPackage-Core/RPackageOrganizer.class.st
+++ b/src/RPackage-Core/RPackageOrganizer.class.st
@@ -340,6 +340,13 @@ RPackageOrganizer >> defineUnpackagedClassesPackage [
 	^ self ensureExistAndRegisterPackageNamed: self packageClass defaultPackageName
 ]
 
+{ #category : #queries }
+RPackageOrganizer >> endingWith: aString [
+	"Return list of packages whose names end with the argument"
+	
+	^ self packages select: [ :p | p name endsWith: aString ]
+]
+
 { #category : #'private registration' }
 RPackageOrganizer >> ensureExistAndRegisterPackageNamed: aSymbol [
 	"A new package is now available and declared in the receiver."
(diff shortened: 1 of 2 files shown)


src/RPackage-Core/RPackageOrganizer.class.st
```diff
@@ -340,6 +340,20 @@ RPackageOrganizer >> defineUnpackagedClassesPackage [
 	^ self ensureExistAndRegisterPackageNamed: self packageClass defaultPackageName
 ]
 
+{ #category : #queries }
+RPackageOrganizer >> endingWith: aString [
```
guillep, line 344: Why not `packagesEndingWith:`?
`endingWith:` suggests (following pharo naming conventions) that it will transform the package organizer, not return a collection of packages.

guillep, line 344: Why not `packagesEndingWith:`? `endingWith:` suggests in pharo/smalltalk code conventions the method returns a new package organizer, not a collection of packages.

src/RPackage-Core/RPackageOrganizer.class.st
```diff
@@ -340,6 +340,20 @@ RPackageOrganizer >> defineUnpackagedClassesPackage [
...
+	"Return list of packages whose names end with the argument."
+	
+	^ self packages select: [ :p | p name endsWith: aString ]
+]
+
+{ #category : #queries }
+RPackageOrganizer >> endingWithDigit [
```
guillep, line 351: Why not `packagesEndingWithDigit`?

src/RPackage-Core/RPackageOrganizer.class.st
```diff
@@ -469,6 +483,13 @@ RPackageOrganizer >> includesPackageNamed: aSymbol [
 	^ packages includesKey:  aSymbol asSymbol
 ]
 
+{ #category : #queries }
+RPackageOrganizer >> includingWith: aString [
```
guillep, line 487: `includingWith:` is not good english.
Why not `packagesIncludingSubstring:`?

src/RPackage-Core/RPackageOrganizer.class.st
```diff
@@ -533,6 +554,13 @@ RPackageOrganizer >> initializeMethodsFor: aBehavior [
 				do: [ :eachSelector | package addMethod: (aBehavior >> eachSelector) ] ]
 ]
 
+{ #category : #queries }
+RPackageOrganizer >> matching: aPatternString [
```
guillep, line 558: Why not `packagesMatching:`?

src/RPackage-Core/RPackageOrganizer.class.st
```diff
@@ -888,6 +916,20 @@ RPackageOrganizer >> signalPackageExists: aPackageName [
 	RPackageConflictError signal: ('A package named {1} already exists' format: {aPackageName})
 ]
 
+{ #category : #queries }
+RPackageOrganizer >> startingWith: aString [
```
guillep, line 920: Why not `packagesStartingWith:`?

src/RPackage-Core/RPackageOrganizer.class.st
```diff
@@ -888,6 +916,20 @@ RPackageOrganizer >> signalPackageExists: aPackageName [
...
+	"Return list of packages whose names begin with the argument."
+	
+	^ self packages select: [ :p | p name beginsWith: aString ]
+]
+
+{ #category : #queries }
+RPackageOrganizer >> startingWithDigit [
```
guillep, line 927: Why not `packagesStartingWithDigit`?

guillep (changes requested):
Please, check the comments I made in the code :)

guillep:
Less important than the comments I made maybe, I'm a bit skeptical about the usefulness of these...
And some even raise a bell. For example, I'm not happy with `packagesIncludingSubstring:` because we don't check substrings in the package, we check substrings in the package name only. It's a bit misleading...

I wonder why having a single method `selectPackagesWithName: [ :name | ]` is not enough.

Ducasse:
Hi guille 
I agree with you and this is why I did not integrate it. 
I would favor selectPackages: [ :each | ] and may be one based on name.
I did not mention it before for obscure reasons. 
S.

guillep:
Ok I got it.

So, @daniapq , thanks for the contribution!
I'll close this PR, and clarify the original issue.
I think it's been good to have the conversation ^^.

PR #5787, opened 2020-02-28
add String>>#copyUpToSubstring:

pavel-krivanek:
add String>>#copyUpToSubstring: as faster and more obvious variant to `readStream upToAll:`. It is usually about 3-times faster.


dionisiydk (approved):

dionisiydk:
@pavel-krivanek you encouraged me to check #upToAll: #5831  :))

PR #5789, opened 2020-02-28
Possible-to-duplicate-of-Instance-variable-through-a-bug

tesonep:
- Adding a flag to detect if the builder is inside a remake of subclasses.

- Adding a validation to check if the slot is already in any of the subclasses.
- Adding tests


MarcusDenker (approved):
ok!

PR #5795, opened 2020-02-28
5794-Improve-comments-of-

### LINKED ISSUE #5794: Improve comments of >>>
To explain why we should pay attention when we comment keyword-based

LaurineDargaud:
fixes : #5794 improve comment


Bajger (approved):
Good catch. Anyway, all executable examples should be tested first, before published.

PR #5803, opened 2020-02-28
Removing-Deprecated80

tesonep:
- Removing Deprecated80 from the Baseline of Pharo
- Removing the references to deprecated classes from manifests.
- Adding compatibility messages to RubTextFieldArea so it can be used instead of the PluggableTextField
- Changing the old spec to remove dependency to a deprecated morph.


Ducasse:
restarting because I do not get why the build does not run.

MarcusDenker:
now it was successful.. I will merge as it fixes failing tests

astares:
@MarcusDenker 

Removing the Deprecated80 has some side effects in Iceberg:

 - https://github.com/pharo-vcs/iceberg/issues/1349

and possibly others too.

Handled in https://github.com/pharo-vcs/iceberg/pull/1351

PR #5807, opened 2020-02-28
4950-Add-a-new-test-for-withUnixLineEndings

### LINKED ISSUE #4950: Add a new test for withUnixLineEndings

olekscode:
Fixed #4950


guillep:
Hi @olekscode, thanks for the PR.

This PR is a duplicate with #5811.
It would seem that #5811 is more complete than this one (it uses a matrix to test many different combinations), so I'm leaning towards the other one :).

PR #5809, opened 2020-02-28
SequenceableCollection - Add executable comments for sorting protocol

### LINKED ISSUE #5450: Add executable comments in SequenceableCollection - sorting shuffling split protocol
There are a lot of opportunities to improve the comment using executable comments in the collection hierarchy and in particular SequencableCollection 

"#(a b c) allButFirst >>> #(b c)"

Lin777:
Only add executable comments to the sorting protocol because suffle always gives a different result and splitjoin protocol already has its executable comments

Fixes #5450

### DIFF
--- a/src/Collections-Abstract/SequenceableCollection.class.st
+++ b/src/Collections-Abstract/SequenceableCollection.class.st
@@ -1313,6 +1313,8 @@ SequenceableCollection >> isSequenceable [
 SequenceableCollection >> isSorted [
 	"Return true if the receiver is sorted by the given criterion.
 	Optimization for isSortedBy: [:a :b | a <= b]."
+	"#(1 2 3) isSorted >>> true"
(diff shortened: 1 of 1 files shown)


dionisiydk (commented):
It would be nice to add also examples with sort function like
```Smalltalk
#(1 2 3 4) sort: #even ascending

```

svenvc:
But this example makes little actual sense (sorting booleans).

What about

    #(xa xc xz xb xy) sort: #last ascending.

for sorting based on the last letter ?

dionisiydk:
Sure, your example is better. Mine was just to show the sort function here

Lin777:
Thanks for the suggestions @svenvc and @dionisiydk, I added some comments that contemplate this type of examples

dionisiydk (approved):
Great

PR #5812, opened 2020-02-28
4456 layout frame should be friendly to possible migration

### LINKED ISSUE #4456: printOn: of LayoutFrame should be friendly to possible migration
Right now 

(LayoutFrame identity
			topOffset: 12;
			bottomFraction: 0;
			bottomOffset: 15;
			leftOffset: -1;
			rightOffset: 1)	
printOn:
       a LayoutFrame ( (0@0) + (-1@12)px corner: (1@0) + (1@15)px ) 

And it would be better to print as 

(LayoutFrame identity
			topOffset: 12;
			bottomFraction: 0;
			bottomOffset: 15;
			leftOffset: -1;
			rightOffset: 1)	

So that people migrating from LayoutFrame fractions:offsets: can just inspect the object and know how to convert.

DanAparicio22:
Implement a new method to fix #4456

### DIFF
--- a/src/Morphic-Base/LayoutFrame.class.st
+++ b/src/Morphic-Base/LayoutFrame.class.st
@@ -1,7 +1,7 @@
 "
 I define a transformation frame relative to some rectangle. I'm basic data structure used for graphics.
 
-Do not use fractions: fractionsOrNil offsets: offsetsOrNil or frations: fractionsOrNil if you do not have already  the rectangles that should be passed as arguments. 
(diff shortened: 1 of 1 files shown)


Ducasse:
The proposed code can be a first step. 
Now can you use cr instead of nextPut: String cr?

DanAparicio22:
Sure, I work on it

guillep:
There is some problem with the dependencies it seems, the build is not green :(.
I'll relaunch the CI. If the problem persist I'll check the tests locally and fix them in a separate PR.

@Ducasse can you confirm that the issues you marked have been correctly addressed? ^^

PR #5815, opened 2020-02-29
5368 the debugger should not rely on the formatter

carolahp:
Removing dependency on the formatter for printing Blocks.
This solution depends on the bytecode generator and on the bytecode encoder.


carolahp:
Wrong branch!, I'm sending a new PR

PR #5816, opened 2020-02-29
Removing dependency on the formatter for printing blocks.

carolahp:
This solution depends on the bytecode encoder and on the bytecode generator.

### DIFF
--- a/src/Kernel/BlockClosure.class.st
+++ b/src/Kernel/BlockClosure.class.st
@@ -549,12 +549,7 @@ BlockClosure >> outerContext: aContext startpc: aStartpc numArgs: argCount copie
 
 { #category : #printing }
 BlockClosure >> printOn: aStream [
-
-	aStream nextPutAll: ((self method hasSourceCode
-		or: [ "There is a decompiler"
-			self class environment includesKey: #FBDDecompiler ])
-				ifTrue: [ self sourceNode formattedCode ]
-				ifFalse: [ 'aBlockClosure(no source code or decompiler available)' ])
+	aStream nextPutAll: (self method ast sourceNodeForPC: self endPC) sourceCode 
 ]
 
 { #category : #accessing }
--- a/src/Kernel/BlockClosure.class.st
+++ b/src/Kernel/BlockClosure.class.st
@@ -549,7 +549,7 @@ BlockClosure >> outerContext: aContext startpc: aStartpc numArgs: argCount copie
 
 { #category : #printing }
 BlockClosure >> printOn: aStream [
-	aStream nextPutAll: (self method ast sourceNodeForPC: self endPC) sourceCode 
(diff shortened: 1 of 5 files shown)


tesonep:
@carolahp There are test that are related:
- ReflectivityReificationTest>>testBlockValueAfter
- LazyClassVariableTest>>testLazyClassVariablePrint
- LazySlotTest>>testLazySlotPrint

Also, there are tests that validate if the blocks are correctly printed? 
Moreover, I want to know if there are tests to check if this works without the source code of the method.

carolahp:
> @carolahp There are test that are related:
> 
> * ReflectivityReificationTest>>testBlockValueAfter
> * LazyClassVariableTest>>testLazyClassVariablePrint
> * LazySlotTest>>testLazySlotPrint

They failed because the way a block is compiled at runtime during the execution of the test is different from the compilation when the block is inside the body of a method. I corrected the way the blocks were defined in the tests. Now they pass.

> 
> Also, there are tests that validate if the blocks are correctly printed?

I found only one BlockClosureTest >> #testPrintOn 

> Moreover, I want to know if there are tests to check if this works without the source code of the method.

It works!, I added one test for this case BlockClosureTest >> #testPrintOnBlockDefinedInMethodWithoutSourceCode

MarcusDenker:
Nice: it uses the source code that we have already (getting the mapping where the code of the block is via the AST like the debugger). This means the pretty printer is not called anymore

src/Kernel/BlockClosure.class.st
```diff
@@ -549,12 +549,7 @@ BlockClosure >> outerContext: aContext startpc: aStartpc numArgs: argCount copie
 
 { #category : #printing }
 BlockClosure >> printOn: aStream [
-
-	aStream nextPutAll: ((self method hasSourceCode
-		or: [ "There is a decompiler"
-			self class environment includesKey: #FBDDecompiler ])
-				ifTrue: [ self sourceNode formattedCode ]
-				ifFalse: [ 'aBlockClosure(no source code or decompiler available)' ])
+	aStream nextPutAll: (self method ast sourceNodeForPC: self endPC) sourceCode
```
MarcusDenker: we can replace
` (self method ast sourceNodeForPC: self endPC) sourceCode`

with:

`self  sourceNode sourceCode`

MarcusDenker:
I did a test to delete the sources: debugger works fine. (in that case of course we need a pretty printer, but that is what it means to have no source: you print it from decompiled bytecode)

MarcusDenker (approved):
ok! If the tests are green we can merge

MarcusDenker:
failing test is related: the tests now depend on Reflectivity due to using the MetaLink... 

Given Collections do not match! additions : #(#Reflectivity) missing: #()
Stacktrace
SystemDependenciesTest>>testExternalSUnitKernelDependencies

PR #5821, opened 2020-02-29
5656-scanTokens-is-bogus

### LINKED ISSUE #5656: scanTokens: is bogus
Why scanTokens does not invoke scanToken and return literal tokens. 

We should update the comments of this method and introduce a new one. 

scanTokens: aStringOrStream
	| scanner |
	scanner := self on: aStringOrStream readStream.
	^scanner contents collect:[ :t | t value ]

Ducasse:
fix: #5656
Add tests, extract RBParser, clean/split tests, add new methods so that we can get access to token objects!

### DIFF
--- a/src/AST-Core-Tests/RBScannerTest.class.st
+++ b/src/AST-Core-Tests/RBScannerTest.class.st
@@ -9,7 +9,7 @@ Class {
 
 { #category : #initialize }
 RBScannerTest >> buildScannerForText: source [
-	^ self parserClass on: (ReadStream on: source)
+	^ self parserClass on: source readStream
 ]
 
 { #category : #initialize }
--- a/src/AST-Core-Tests/RBScannerTest.class.st
+++ b/src/AST-Core-Tests/RBScannerTest.class.st
@@ -9,11 +9,11 @@ Class {
 
 { #category : #initialize }
 RBScannerTest >> buildScannerForText: source [
-	^ self parserClass on: (ReadStream on: source)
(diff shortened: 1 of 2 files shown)


src/AST-Core-Tests/RBScannerTest.class.st
```diff
@@ -9,7 +9,12 @@ Class {
 
 { #category : #initialize }
 RBScannerTest >> buildScannerForText: source [
-	^RBScanner on: (ReadStream on: source)
+	^ self parserClass on: (ReadStream on: source)
+]
+
+{ #category : #initialize }
+RBScannerTest >> parserClass [
```
dionisiydk: I think It should be named #scannerClass.

src/AST-Core-Tests/RBScannerTest.class.st
```diff
@@ -9,7 +9,12 @@ Class {
 
 { #category : #initialize }
 RBScannerTest >> buildScannerForText: source [
-	^RBScanner on: (ReadStream on: source)
+	^ self parserClass on: (ReadStream on: source)
```
dionisiydk:
```suggestion
	^ self parserClass on: source readStream
```
- to remove another global here

src/AST-Core-Tests/RBScannerTest.class.st
Ducasse: Yes you are rigth I was distracted.

PR #5825, opened 2020-03-02
5824-remove-class-var-that-is-still-referenced-change-is-not-saved

### LINKED ISSUE #5824: remove class var that is still referenced: change is not saved
This PR has a bad side effect: https://github.com/pharo-project/pharo/pull/5660

The idea is that it checks if a class definition has a variable (if it does, it is undeclared, which in turn asks you to define it, which in turn is done on nil, which is not good).

The solution just adds a handler for OCUndeclaredVariableWarning. But that is not good, as accepting a class definition will result in recompiling methods, which could have undeclareds.

MarcusDenker:
add check that we are indeed compiling a class defintion. fixes #5824


MarcusDenker:
I will close and reopen as the failing tests have been fixed

PR #5828, opened 2020-03-02
Fixing random failing tests

tesonep:
- Fixing random tests of Hiedra
- Better explanation for assert in NECompletion
- Bigger timeout for release test


MarcusDenker:
yes, NoUnusedTemporaryVariablesLeftTest: the problem is that in the end 

`select: [ :m | m ast temporaries anySatisfy: [ :x | x binding isUnused ] ].`

uses the AST, but we should formulate the test to be on the bytecode level instead. (even though the ideas that you can treat temps and ivars the same is nice, we are not there yet...)

PR #5829, opened 2020-03-02
Updating iceberg after deprecations in 90

tesonep:
After removing the deprecations of Pharo 80 in Pharo 90.
We have changed the old TextInputFieldAdapter to use the Rubric TextField.

It required changes in Iceberg and in Spec events.


MarcusDenker (approved):
- failing tests are not related
- changed look good

PR #5834, opened 2020-03-02
Add a hook to create the completion context

guillep:
Like that subclasses can override the creation of the completion context.


MarcusDenker (approved):

MarcusDenker:
failing tests are not related: network problems "503 Service Temporarily Unavailable"

PR #5835, opened 2020-03-03
5830-speed-up-NoUnusedTemporaryVariablesLeftTest

### LINKED ISSUE #5830: speed up NoUnusedTemporaryVariablesLeftTest
I think NoUnusedTemporaryVariablesLeftTest uses the AST level but it should be possible to do it on the bytecode level instead

MarcusDenker:
-  use #allMethodsSelect: 
-  pre-filter to only check methods on the AST level that have temps

fixes #5830

### DIFF
--- a/src/ReleaseTests/NoUnusedTemporaryVariablesLeftTest.class.st
+++ b/src/ReleaseTests/NoUnusedTemporaryVariablesLeftTest.class.st
@@ -23,8 +23,9 @@ NoUnusedTemporaryVariablesLeftTest >> tearDown [
 NoUnusedTemporaryVariablesLeftTest >> testNoUnusedTemporaryVariablesLeft [
 	"Fail if there are methods who have unused temporary variables"
 	| found validExceptions remaining |
-	found := (SystemNavigation default allBehaviors flatCollect: #localMethods)
-		select: [ :m | m ast temporaries anySatisfy: [ :x | x binding isUnused ] ].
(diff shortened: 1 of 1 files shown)


MarcusDenker:
For me it reduces the first run from 20 seconds to 10 seconds.

dionisiydk (approved):
Very nice

src/ReleaseTests/NoUnusedTemporaryVariablesLeftTest.class.st
```diff
@@ -23,8 +23,9 @@ NoUnusedTemporaryVariablesLeftTest >> tearDown [
 NoUnusedTemporaryVariablesLeftTest >> testNoUnusedTemporaryVariablesLeft [
 	"Fail if there are methods who have unused temporary variables"
 	| found validExceptions remaining |
-	found := (SystemNavigation default allBehaviors flatCollect: #localMethods)
-		select: [ :m | m ast temporaries anySatisfy: [ :x | x binding isUnused ] ].
+	found := SystemNavigation default allMethodsSelect: [ :m | 
+		((m numTemps - m numArgs) > 0) and: [
```
dionisiydk, line 27: Maybe we can add #numTempsExceptArgs or with some better name

MarcusDenker: yes, I was thinking about it.. it is really bad that the compiledMethod and the AST have a different model of what a temp is...

PR #5844, opened 2020-03-04
Fix Alien libcName

### LINKED ISSUE #5843: Fix Alien libcName
- fix this method and use "os version"

astares:
Fix #5843


tesonep:
Why not to use directly: 
```
LibC uniqueInstance libraryName.
```

tesonep (commented):
LibC uniqueInstance libraryName.

astares:
Just repaired it (the method it provided by original Alien from SqS).

PR #5848, opened 2020-03-04
UnsafeAlien is missing nonPointerObject: implementation

### LINKED ISSUE #5847: UnsafeAlien is missing nonPointerObject: implementation
in class UnsafeAlien #nonPointerObject: setter is called from class side #forPointerTo: but not provided as a method

astares:
Fix #5847

### DIFF
--- a/src/Alien-Core/UnsafeAlien.class.st
+++ b/src/Alien-Core/UnsafeAlien.class.st
@@ -77,6 +77,12 @@ UnsafeAlien class >> initialize [
 ]
 
 { #category : #accessing }
+UnsafeAlien >> nonPointerObject: anObject [
+
+	nonPointerObject := anObject
+]
+
+{ #category : #accessing }
 UnsafeAlien >> object [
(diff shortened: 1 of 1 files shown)


astares:
Also done in http://www.squeaksource.com/Alien.html

Name: Alien-Core-tbn.103
Author: tbn
Time: 5 March 2020, 12:33:21.617091 am
UUID: 7f5ae215-69cd-344e-a76c-e19d333b1465
Ancestors: Alien-Core-nice.102

Add missing accessor UnsafeAlien>>nonPointerObject:
that is called from class side #forPointerTo:

<blockquote><div><strong><a href="http://www.squeaksource.com/Alien.html">SqueakSource</a></strong></div></blockquote>

estebanlm (approved):
is ok, but I wonder why is the use case to use Alien now?

astares:
@estebanlm The use case is to get a quality image where we dont have unimplemented calls ;)

PR #5857, opened 2020-03-05
isLiteralMethod-should-work-on-sista

estebanlm:
taking literal method bytecodes from encoder class


MarcusDenker (approved):

Ducasse:
Little question: why isLiteralMethod is still packaged in Calypso? and not in compiledMethod or related packages?

PR #5863, opened 2020-03-07
windowMenuCleanup

pavel-krivanek:
windowMenuOn: cleanup

We have two implementations of windows menu definition. One uses pragma and allows windows menus to be extended by other packages, the other is hardcoded. Unfortunately, the pragma version is not used. This PR changes it

### DIFF
--- a/src/Morphic-Widgets-Windows/SystemWindow.class.st
+++ b/src/Morphic-Widgets-Windows/SystemWindow.class.st
@@ -1326,12 +1326,17 @@ SystemWindow >> maximumExtent: aPoint [
 { #category : #api }
 SystemWindow >> menu [
 
-	| aMenuBuilder |
-	aMenuBuilder := (PragmaMenuBuilder pragmaKeyword: self discoveredMenuPragmaKeyword model: self)
-		menuSpec;
-		yourself.
-				
-	^ aMenuBuilder menu
+	| aMenuBuilder res |
+
+	aMenuBuilder := BlockMenuBuilder owner: self.
+	aMenuBuilder model: self.
+	aMenuBuilder buildingBlock: [ :aBuilder |
+		self class windowMenuOn: aMenuBuilder.
+		self class tilingMenuOn: aBuilder.
+		self class tilingMenuItemsOn: aBuilder ].
+	aMenuBuilder menuSpec.
+	res := aMenuBuilder menu.
+	^ res.
 ]
 
 { #category : #'resize/collapse' }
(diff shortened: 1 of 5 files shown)


pavel-krivanek:
with this PR, I added the window tiling menu items (we already have the shortcuts for them for a long time)

Pharo_EdpD2JE6Mx

src/Morphic-Widgets-Windows/SystemWindow.class.st
```diff
@@ -1277,10 +1277,12 @@ SystemWindow >> maximumExtent: aPoint [
...
-	| aMenu |
-	aMenu := MenuMorph new.
-	self windowMenuOn: aMenu.
-	^ aMenu
+	| aMenuBuilder |
+	aMenuBuilder := (PragmaMenuBuilder pragmaKeyword: self discoveredMenuPragmaKeyword model: self)
+		menuSpec;
+		yourself.
+				
+	^ aMenuBuilder menu
```
guillep: I invoke @tesonep here for review, because he was chasing performance problems with the pragma collector, particularly in big images.

pavel-krivanek: I see... Then we need smarter caches system for pramgas.

tesonep: Yes, this change should be redone to not use pragmas.
Directly we can call the builder method

tesonep: I think the best solution is to use the builder and remove the usage of the window pragma.

pavel-krivanek: hmm, it follows the very same pattern as the world menu. Will we change the registration mechanism for both? (Commander or so...)

pavel-krivanek: ehhh.... guess what... we do not have the other working menu builder than the `PragmaMenuBuilder`

Ducasse: Indeed based on Commander2 we wanted to build a new menu builder and remove the old one and esteban told us that this is not needed because people should use Spec2.
Well reality is in the middle. The pragmaMenuBuilder and friends suck quite a bit to my eyes and commander2 was designed (and the last chapter explains it) so that people can exactly build menu builder for their own framework.

pavel-krivanek: in this case, it is part of the Morphic-Core so we should take care of dependencies

Ducasse:
Yes now commander core is not linked to Spec at all. 
In fact all the spec2 support is managed by a decorator and the same can be done for morphic or Bloc :)

pavel-krivanek:
So, I created a menu builder that does not use pragmas. It is not the most elegant solution but integration of Commander2 into the Morphic menus building process would require such menu builder anyway... :-/

Ducasse:
So is the issue ready to integrate?
I think so

pavel-krivanek:
yes, build artifacts work well

PR #5866, opened 2020-03-09
Adding a name cache to the Package organizer.

tesonep:
Adding a name cache to the Package organizer.
The package organizer should not recalculate the names every time, as the tools access them by the collection of names


Ducasse:
Pablo who is calling the cache to remove packages?

tesonep:
The package organizer was already receiving the remove package announcement and executing basicRemovePackage: and I have added removePackageNameFromCache: in this method.

Ducasse:
Thanks for the explanation.

Ducasse:
Apparently this is a problem with the dependencyTests

Ducasse:
Pablo I will integrate it and we should fix the dependency after?

guillep:
> Pablo I will integrate it and we should fix the dependency after?

I'd say, let's not hurry :)
If we break all the tests now, other PRs will be polluted and we will start with "not related" "not related" "not related"

tesonep:
@guillep is right I will fix the PR
It will be easy to fix it now than later.

tesonep:
Also I think most of the errors are already fixed, and the run is old. I will update the PR and we can see what of them are real

tesonep:
The problem looks like a real problem with the initialization of the cache during the bootstrap. 
I am checking it.

src/RPackage-Core/RPackageOrganizer.class.st
```diff
@@ -213,6 +214,15 @@ RPackageOrganizer >> addMethod: method [
 	rPackage addMethod: method
 ]
 
+{ #category : #'package-names-cache' }
+RPackageOrganizer >> addPackageNameToCache: aPackageName [
+
+	packageNames ifNil: [ ^ self ].
```
estebanlm, line 220: why not 
```Smalltalk
self packageNames add: ... 
```
?

tesonep: Because packageNames will force the generation of the cache during the loading of code, and I wanted to generate it in the first read. Maybe I am just optimizing before needed.

guillep (approved):
Tests not related, and from an old run.

image

PR #5869, opened 2020-03-09
Added 4-value message and 4-cull message to MessageSend and WeakMessa…

David5i6:
Added 4-value message and 4-cull message to MessageSend and WeakMessageSend because the object has a 4-with perform message that is a primitive.

### DIFF
--- a/src/Kernel/MessageSend.class.st
+++ b/src/Kernel/MessageSend.class.st
@@ -115,6 +115,13 @@ MessageSend >> cull: arg1 cull: arg2 cull: arg3 [
 		ifFalse: [ self value: arg1 value: arg2 value: arg3 ]
 ]
 
+{ #category : #evaluating }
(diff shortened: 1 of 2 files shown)


Ducasse:
Hello David

While this is nice that you spent your time doing it. We do not want to have support for 4 arguments. First because a message with three arguments is already a lot so 4 is really not needed. If people really want they can use the version with an array. 
I will let other people react but this is my analysis. 
I never needed this message. 
S.

David5i6:
That makes, sense.

I also put an eye on squeack and cuis. Suprisingly Squeack has these variants while Cuis only has the array version, no 1,2 o 3 parameters.

After seeing that I liked the Cuis approximation becauses classes has few methods without losing functionality.

Ducasse:
This is true that the VM has only one primitive expecting an array of arguments. 
But forcing all the users to create an array is not that nice. 
I prefer our tradeoffs.

PR #5873, opened 2020-03-10
5872-workingDirectory-badly-resolved-if-its-path-contains-non-ascii-characters

### LINKED ISSUE #5872: workingDirectory badly resolved if its path contains non-ascii characters
To reproduce:
1. create a folder with a name containing a non-ascii character: `mkdir loïc && cd loïc`
1. get a fresh pharo image and vm: `curl get.pharo.org/64/80+vm |  bash`
1. `./pharo-ui Pharo.image eval "(FileLocator workingDirectory) asFileReference exists"` => false
1. `./pharo-ui Pharo.image eval "(FileLocator workingDirectory) asFileReference fullName"` => /private/tmp/loÃ¯c

It is because the primitive `primitiveGetCurrentWorkingDirectory` just returns a raw byte array that should be decoded with the proper Character encoding.

Running `ZnCharacterEncoder utf8 asZnCharacterEncoder decodeBytes: OSPlatform current currentWorkingDirectoryPath asByteArray`gives the expected result: /private/tmp/loïc

demarey:
Fixes #5872 workingDirectory badly resolved if its path contains non-ascii characters

### DIFF
--- a/src/System-Platforms/OSPlatform.class.st
+++ b/src/System-Platforms/OSPlatform.class.st
@@ -118,7 +118,7 @@ OSPlatform >> currentWorkingDirectoryPathWithBufferSize: aBufferSize [
 
 { #category : #accessing }
 OSPlatform >> currentWorkingDirectoryRawPath [
-	"This method calls the method getPwdViaFFI with arugement of a buffer size. 
+	"This method calls the method getPwdViaFFI with argument of a buffer size. 
 	By default it uses the defaultMaximumPathLength of each subclass as the buffer size.
 	We return a raw byte array that should be decoded with the proper encoding."
 	^ (self currentWorkingDirectoryPathWithBufferSize: self defaultMaximumPathLength) asByteArray
--- a/src/System-Platforms/OSPlatform.class.st
+++ b/src/System-Platforms/OSPlatform.class.st
@@ -126,6 +126,12 @@ OSPlatform >> currentWorkingDirectoryRawPath [
 ]
 
 { #category : #accessing }
+OSPlatform >> defaultEncoding [
+
+	^ ZnCharacterEncoder utf8
+]
+
+{ #category : #accessing }
 OSPlatform >> defaultMaximumPathLength [
 	"This method returns the default maximum path length for the getCurrentWorkingDirectory implementation"
 	 ^self subclassResponsibility 
--- a/src/System-Platforms/OSPlatform.class.st
+++ b/src/System-Platforms/OSPlatform.class.st
@@ -98,7 +98,7 @@ OSPlatform >> current [
 { #category : #accessing }
 OSPlatform >> currentWorkingDirectoryPath [
 	"We MUST decode raw bytes to support non-Ascii characters"
-	^ self currentWorkingDirectoryRawPath decodeWith: self environment defaultEncoding
+	^ self currentWorkingDirectoryRawPath decodeWith: self defaultEncoding
 ]
 
 { #category : #accessing }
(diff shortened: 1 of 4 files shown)


tesonep:
@demarey check that the modified method relies in a method that is loaded in a package that is loaded later. As this is core part of the system, we need to be aware that maybe #environment is not accessible.

Ducasse:
Christophe I let you have a look.

src/System-Platforms/OSPlatform.class.st
```diff
@@ -97,26 +97,34 @@ OSPlatform >> current [
...
 	^ self currentWorkingDirectoryPathWithBuffer: buffer
 ]
 
+{ #category : #accessing }
+OSPlatform >> currentWorkingDirectoryRawPath [
+	"This method calls the method getPwdViaFFI with arugement of a buffer size. 
+	By default it uses the defaultMaximumPathLength of each subclass as the buffer size.
+	We return a raw byte array that should be decoded with the proper encoding."
+	^ (self currentWorkingDirectoryPathWithBufferSize: self defaultMaximumPathLength) asByteArray
+
```
guillep, line 125: Is the `asByteArray` needed?

src/System-Platforms/OSPlatform.class.st
```diff
@@ -97,26 +97,34 @@ OSPlatform >> current [
 
 { #category : #accessing }
 OSPlatform >> currentWorkingDirectoryPath [
-	"This method calls the method getPwdViaFFI with arugement of a buffer size. By default it uses the defaultMaximumPathLength of each subclass as the buffer size."
-	^ self currentWorkingDirectoryPathWithBufferSize: self defaultMaximumPathLength
-
+	"We MUST decode raw bytes to support non-Ascii characters"
+	^ self currentWorkingDirectoryRawPath decodeWith: self environment defaultEncoding
```
guillep: As @tesonep says, `environment` is an extension method, so this method is creating a circular dependency. (System-Platforms <-> System-OSEnvironments)

guillep (changes requested):

guillep:
@demarey could you give us an idea of when you'll be checking this?

demarey:
@guillep 
OSPlatform>>#currentWorkingDirectoryPath uses 
- #decodeWith: packaged in Zinc-Character-Encoding-Core (loaded in Bootstrap kernel group) => ok
- #defaultEncoding packaged in System-OSEnvironments. The problem here is that System-OSEnvironments package is loaded with UFFI because it uses a lot UFFI. BaselineOfUnifiedFFI is loaded either by MorphicCore or Athens (loaded by BaselineOfIDE).
I have the feeling that we should move the #defaultEncoding definition from OSEnvironment to OSPlatfrom. WDYT?

guillep:
> * #defaultEncoding packaged in System-OSEnvironments. The problem here is that System-OSEnvironments package is loaded with UFFI because it uses a lot UFFI. BaselineOfUnifiedFFI is loaded either by MorphicCore or Athens (loaded by BaselineOfIDE).
>   I have the feeling that we should move the #defaultEncoding definition from OSEnvironment to OSPlatfrom. WDYT?

Looks good enough to me. Can we try that?

tesonep:
Really cool analysis, it will be nice to extend the dependency analyser to
know where the packages are loaded and its dependencies.

What do you think?

tesonep:
And also, I share it should be a good solution to move the method

guillep:
Thanks! Let's wait for the CI.

guillep (approved):
Good! tests are green!

PR #5882, opened 2020-03-11
5875-Some-TTF-Fonts-are-not-visible-in-Pharo-Windows

### LINKED ISSUE #5875: Some TTF Fonts are not visible in Pharo (Windows)
I installed the following TTF fonts in Windows 

    https://www.jetbrains.com/lp/mono/

and can use them from WordPad, VisualStudio, Word and other applications.

But it is not possible to find them in the Pharo font dialog.

tesonep:
Since Windows 1809, Windows allows us to install the fonts in a local directory in the user.

This values should be read from the repository in the key 
HKEY_CURRENT_USER\Software\Microsoft\Windows NT\CurrentVersion\Fonts

However, things happened and we don't have the support of the registry in the base image.
So we are trying with some well-known directories.

Fixes #5875

### DIFF
--- a/src/FreeType/FreeTypeFontProvider.class.st
+++ b/src/FreeType/FreeTypeFontProvider.class.st
@@ -309,6 +309,18 @@ FreeTypeFontProvider >> getWindowsFontFolderPath [
 ]
 
 { #category : #'file paths' }
+FreeTypeFontProvider >> getWindowsLocalPaths [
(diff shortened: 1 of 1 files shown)


estebanlm (approved):

astares:
@tesonep / @estebanlm confirm it is working, thanks!

Issue #5850, opened 2020-03-04
TonelWriter should use traceCr:


astares:
instead of logCr:

astares:
Need to be fixed here: https://github.com/pharo-vcs/tonel/issues/82

PR #5886, opened 2020-03-13
Pharo8 Backport - Fixing problem with class trait compositions with parens.

tesonep:
When Monticello was saving the traits, it was not able to handle complex trait compositions with parenthesis.
This was also affecting Iceberg and Tonel, as both of them relies in the MC model.

- Adding the node of withPrecedenceOf: in the Class parser
- The trait definition has to have the class trait composition.
- Improving the handling of parentheses.
- Updating the version of iceberg and Tonel

These complex expressions are affecting Moose.

### DIFF
--- a/src/Ring-Definitions-Monticello/MCTraitDefinition.extension.st
+++ b/src/Ring-Definitions-Monticello/MCTraitDefinition.extension.st
@@ -27,21 +27,6 @@ MCTraitDefinition >> classDefinitionString [
 ]
 
 { #category : #'*Ring-Definitions-Monticello' }
-MCTraitDefinition >> classTraitCompositionString [
-	^self traitComposition 
-		ifNil: [ '{}' ]
-		ifNotNil: [ :source| | tokens tcs |
-			tcs := ''.
-			tokens := source parseLiterals.
-			tokens do:[ :each| 
-				each first isUppercase 
-					ifTrue: [ tcs := tcs, each, ' classTrait + ' ] ].
-			tcs isEmpty 
-				ifTrue: [ '{}' ]
-				ifFalse:[ tcs copyFrom: 1 to: tcs size - 3 ] ]
-]
-
-{ #category : #'*Ring-Definitions-Monticello' }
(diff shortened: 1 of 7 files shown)


src/Ring-Definitions-Monticello/MCTraitDefinition.extension.st
```diff
@@ -26,21 +26,6 @@ MCTraitDefinition >> classDefinitionString [
...
-			tcs := ''.
-			tokens := source parseLiterals.
-			tokens do:[ :each| 
-				each first isUppercase 
-					ifTrue: [ tcs := tcs, each, ' classTrait + ' ] ].
-			tcs isEmpty 
-				ifTrue: [ '{}' ]
-				ifFalse:[ tcs copyFrom: 1 to: tcs size - 3 ] ]
-]
-
```
guillep, line 43: Yes it is good that this code goes :).
It finished its cycle

guillep:
So to also add to this:
 - This is a backport of https://github.com/pharo-project/pharo/pull/5861
 - it also fixes https://github.com/pharo-project/pharo/issues/5850 (iceberg 1.6.6?)

I'll also invoke for review @ClotildeToullec as she was involved in the issue. Could you check the changes?

@tesonep the changes look fine. The 1.6.5 -> 1.6.7 in iceberg is kind of a blackbox. I assume 1.6.6 was integrated in Pharo9 no?
Also, the CD* things seem not used.

tesonep:
@guillep Yes, the iceberg version has passed in Pharo 9. It should be working, but maybe we can have an intermediate version just for Pharo 8. 

The 1.6.6 version just removed the usage of some methods that were already deprecated in Pharo 8. 
 - crLog
 - PluggableTextField

guillep:
> @guillep Yes, the iceberg version has passed in Pharo 9. It should be working, but maybe we can have an intermediate version just for Pharo 8.

I think there is no need :)

We should discuss somehow during the week how to test the latest pharo8 image, because the launcher only has the release.

PR #5889, opened 2020-03-13
5779 - Setting Display Scale Factor Changing

### LINKED ISSUE #5779: Settings -> Display Scale Factor changing
In the Menu "Settings" in Appearance the "Display scale Factor" is alive. It changes automatically.
This issue also happens with the "Fullscreen Margin" value.

Sometimes that happens magically when entering in the Settings, other times it happens after draggin the value selector and changing the value.

It happens on MacOSX with Pharo 8 and Pharo 9 images downloaded from the launcher and in Windows 10 an pharo 9 from git.

David5i6:
fixes #5779 

An if clause is used to return if it's not dragging the slider. MacOS and Windows, seems fixed.

Also a little comment in code and reformating for better readability.

-----

Please I made this change to correct a problem with Pharo 8 and Pharo 9 on Windows 10 and MacOS (not tested on Linux), but I don't have a complete vision of Morphic and Spec, Spec2.

The problem was that when movin the mouse over the slider it changes magically it's position, but not alwas. I saw that there was no "draggin" state, but instead the style of the slider is diferent and determines the state and used it to determine if slider is dragging.

I think adding a "dragging" and "dragging:" messages to set and retrieve the dragging state should be a cleaner way so I added that in a second commit.

### DIFF
--- a/src/Morphic-Widgets-Scrolling/Slider.class.st
+++ b/src/Morphic-Widgets-Scrolling/Slider.class.st
@@ -7,7 +7,8 @@ Class {
 		'setValueSelector',
 		'sliderShadow',
 		'sliderColor',
-		'descending'
+		'descending',
+		'dragging'
 	],
 	#category : #'Morphic-Widgets-Scrolling'
 }
@@ -64,6 +65,16 @@ Slider >> descending: aBoolean [
 	self value: value
 ]
 
+{ #category : #access }
+Slider >> dragging [
+	^ dragging.
+]
+
+{ #category : #access }
+Slider >> dragging: aBoolean [
+	dragging := aBoolean.
+]
+
 { #category : #geometry }
 Slider >> extent: newExtent [
 	newExtent = bounds extent ifTrue: [^ self].
@@ -80,6 +91,7 @@ Slider >> initialize [
 	super initialize.
 	value := 0.0.
 	descending := false.
+	dragging := false.
 	self initializeSlider
 ]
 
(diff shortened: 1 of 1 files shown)


Ducasse:
Thanks this is uper cool that you spend your time fixing such kind of glitches. We are to used to jump over them. Thanks a lot. This is important because there is no little changes not worth. 
This is because people give their energy to Pharo that it improves. So thank a lot.

src/Morphic-Widgets-Scrolling/Slider.class.st
```diff
@@ -110,19 +122,20 @@ Slider >> initializeSlider [
 Slider >> mouseDownInSlider: event [ 
 
 	slider borderStyle style == #raised
-		ifTrue: [slider borderColor: #inset].
+		ifTrue: [slider borderColor: #inset. self dragging: true].
```
guillep: I don't understand why this depends on the border style. Could you add a comment on this?

src/Morphic-Widgets-Scrolling/Slider.class.st
```diff
@@ -110,19 +122,20 @@ Slider >> initializeSlider [
...
+	sliderShadow show.
+	
 ]
 
 { #category : #'other events' }
 Slider >> mouseUpInSlider: event [ 
 
 	slider borderStyle style == #inset
-		ifTrue: [slider borderColor: #raised].
+		ifTrue: [slider borderColor: #raised. self dragging: false].
```
guillep: Same same :)

guillep (changes requested):
I'd like to know, at least to keep as documentation, a couple of points.
Maybe it could make sense to add some method/class comments?

Anyways, thanks :)

David5i6:
You're right. Due that the dragging state is now reflected on draggin variable there's no need to make the check against the borderColor.

I'm going to change it. Thanks !

David5i6:
Re-reading the code, there some nonsense things. If mouse is down, It start dragging... but... ¿ It's possible to make the mouse down again without making it up ?

guillep:
Thanks, I like the changes :)

David5i6:
sorry, I detected a little very rare cases after that. I'm gouing to send a merge request again with some code guards.

guillep:
Sure, thanks for the effort!!

PR #5890, opened 2020-03-14
Cleanup world

guillep:
Several refactorings and cleanups to avoid duplications in `#displayWorldState:ofWorld:` and friends.

This PR also introduces several things like:
 - hide the fact that there is a canvas. In Form/BitBlt a canvas is always there, yes. But in athens a canvas is only available during drawing. Having accessors/setters for it makes no sense. This includes for example `worldState canvas: nil` => `worldState invalidate`
- remove some old/never-used/experimental spec1/oswindow integration. This has to be redone for spec2


Ducasse:
Guille what is the difference between OS and VM World renderer? We should probably explain it in the comments. I kind of get it: one with a canvas and the other is an external canvas?
I read all the code (not really a review). 
BTW you seem to have as yet unclassified.

guillep:
> Guille what is the difference between OS and VM World renderer? We should probably explain it in the comments.

- OS* uses OSWindow based windowing (and SDL)
- VM* uses the old mechanism where the windowing is on the VM

Yes, we should maybe rename those classes, but I'll do that separately.
That was already like that before, we can outlive it :)

> I kind of get it: one with a canvas and the other is an external canvas?

Both have canvases, but the implementation in OS window kind of hides the canvas. Because the OS window can render using Form/Display and so on, but also on Athens.
In Athens, the canvas is a temporary object you get while rendering only, something like:

```smalltalk
"I have no canvas"
myRenderer drawDuring: [ :myAthensCanvas | 
   "I got a canvas and can do whatever I want"
]
"The canvas was released"
```

So if we want to somehow exploit polymorphism, better assume the most limited.

> I read all the code (not really a review).

Now that I've explained how to do a review in github you'll have no choice MWAHAHA.

> BTW you seem to have as yet unclassified.

Ouch, will check. Thanks :)

guillep:
Tests not related (and logged in https://github.com/pharo-project/pharo/issues/5895)

tesonep (approved):

PR #5897, opened 2020-03-15
5896 detected error in extreme rare cases for slider

David5i6:
fixes #5896 

There where error on some extremely rare cases.

Also a bug when clicking outside the slider handler is corrected.

Code is documented (Perhaps too verbose).

PR #5911, opened 2020-03-17
Fix slider broken tests

guillep:
Issue #5897 broke a couple of spec1 tests.
I integrated a bit carelessly because the change looked simple.
So here's a fix.

 - the text can be empty (handle that case)
 - the value can be a fraction: convert it to float before transforming it to string => otherwise asNumber will fail later


guillep:
Tests fixed. There are still the HTTPs tests, which are unrelated to this issue.

image

PR #5916, opened 2020-03-17
Remove ActiveHand and ActiveEvent

### LINKED ISSUE #5908: Should remove accesses to `ActiveHand` global
The global assumes there is a single hand, while there may be many of them.
And to this to work properly, there is code like this that updates the global in a loop:

```smalltalk
self extraWorldList do: [:world |
	ActiveHand := world activeHand.
	world doOneCycle.
].

(self currentWorld isNotNil and: [(self extraWorldList includes: self currentWorld) not]) ifTrue: [
	self currentWorld doOneCycle.
	ActiveHand := self currentWorld activeHand
]
```

Notice also that the code below may have a bug too, because it may happen that the cycle in the second block executes with a wrong hand. In my image we have 50 users of the class variable. We should study case by case.

guillep:
Fixes #5908.

Mostly, delegate the obtention to the active hand to the active world.
This raises a couple of issues:

 - some non-morph objects try to access the world => they should get it as argument
 - some morph objects try to get their world before being installed in the system

A summary of changes:
 - Remove global cursor usages. `Cursor X isCurrent` requires access to a world
 - `ActiveHand` => `self activeHand`, or `morph activeHand`
 - take care of objects trying to get keyboard focus before shown/installed in the world.
 - `ActiveEvent` => `aMorph activeHand lastEvent`

My image behaves pretty well after these changes.
We should see a bootstrapped image.

### DIFF
--- a/src/Athens-Morphic/MenuMorph.extension.st
+++ b/src/Athens-Morphic/MenuMorph.extension.st
@@ -5,7 +5,7 @@ MenuMorph >> drawOnAthensCanvas: anAthensCanvas [
 	"Draw the menu. Add keyboard-focus feedback if appropriate"
 
 	super drawOnAthensCanvas: anAthensCanvas.
-	(ActiveHand notNil and: [ ActiveHand keyboardFocus == self and: [ self rootMenu hasProperty: #hasUsedKeyboard ] ])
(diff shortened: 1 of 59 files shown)


guillep:
Only one failing tests, looks like a fleaky one

image

PR #5925, opened 2020-03-17
5924 scroll bars dont drag with mouse

### LINKED ISSUE #5924: ScrollBars don't drag with mouse
Scrollbars don't drag when using mouse.
This bug is related with my fix for the slider.

I'm going to make a pullrequest with this fix.

Also the useShadow option is checked in code but UITheme alwas return false.
I made a little change in UITheme to let users customize this behaivor.

David5i6:
fixes #5924 

The scrollBars don't drag with mouse as a result of an previous fix over the slider.
Also UITheme now lets users to set the behaivor of the shadow of the scrollbar handler.


guillep:
I'm going to be pushy :)

Would you dare trying to add a test for these cases too?
Because we were bitten by these bugs, a test could help us detecting these things earlier in the future.

Thanks!

David5i6:
Indeed I wanted to do a test, but not sure about how to test response to UI events. (I have no more than a few hours of pharo/smalltalk experience) but I'm looking source for examples. I will add a new comment when add them or when resign.

guillep:
Just, maybe as a starting point, check that morphs implement several methods to simulate events:

image

You could use those or add new ones that help you with this test :)

David5i6:
Thanks !

David5i6:
Added tests for slider and scrollbar.
Added some new "simulate" methods for helping in creating those and future tests.
Revised scrollbar.
All 4 tests passed and slider text checked to fail in original versión prior to correction.
Also Scrollbar test check to fail prio to corrections.

David5i6:
By error moved "simulate" method to protocol "simulation" instead of "evetn testing", corrected in the last commit.

Now it seems everything is ok. :-)

guillep:
Wow, this is great!

Thanks!

I'll check it first thing tomorrow morning (now family is asking for me).

guillep:
We are getting an exception at bootstrap

```
[64] *** EXCEPTION ***

[64] MessageNotUnderstood: ThemeSettings>> #useScrollbarThumbShadow:
```

It's strange in here it is sent to a `ThemeSettings`
https://github.com/pharo-project/pharo/pull/5925/files#diff-23715ef0c842a2cd75d12a8802ea68c2R202

But that method is defined in `UITheme`, not `ThemeSettings`
https://github.com/pharo-project/pharo/pull/5925/files#diff-23715ef0c842a2cd75d12a8802ea68c2R6000

<blockquote><div> GitHub</div><div><strong><a href="https://github.com/pharo-project/pharo/pull/5925">5924 scroll bars dont drag with mouse by David5i6 · Pull Request #5925 · pharo-project/pharo</a></strong></div><div>fixes #5924
The scrollBars don't drag with mouse as a result of an previous fix over the slider.
Also UITheme now lets users to set the behaivor of the shadow of the scrollbar handler.</div></blockquote>
<blockquote><div> GitHub</div><div><strong><a href="https://github.com/pharo-project/pharo/pull/5925">5924 scroll bars dont drag with mouse by David5i6 · Pull Request #5925 · pharo-project/pharo</a></strong></div><div>fixes #5924
The scrollBars don't drag with mouse as a result of an previous fix over the slider.
Also UITheme now lets users to set the behaivor of the shadow of the scrollbar handler.</div></blockquote>

David5i6:
There is something very strange on the fullStack... Seem like it is called:

 #useScrollbarThumbShadow:
PharoLightTheme

guillep:
When the image is created from the source code, it installs the `PharoLightTheme` (subclass of `UITheme`)  using `PharoLightTheme newDefault`. And that is calling `useScrollbarThumbShadow:`on ThemeSettings, but `useScrollbarThumbShadow:` is defined in `UITheme`.

```
ThemeSettings(Object)>>doesNotUnderstand: #useScrollbarThumbShadow:
PharoLightTheme class(UITheme class)>>newDefaultSettings
PharoLightTheme class(UITheme class)>>defaultSettings
PharoLightTheme(UITheme)>>defaultSettings
PharoLightTheme(UITheme)>>initialize
PharoLightTheme>>initialize
PharoLightTheme class(Behavior)>>new
PharoLightTheme class(UITheme class)>>newDefault
...
```

Shouldn't be `useScrollbarThumbShadow:` be moved to `ThemeSettings`?

David5i6:
Done !

David5i6:
Humm, now the VM sems to crash...

\sqAllocateMemorySegmentOfSizeAboveAllocatedSizeInto mmap: Cannot allocate memory

But I think that's not related with this change.

guillep:
> \sqAllocateMemorySegmentOfSizeAboveAllocatedSizeInto mmap: Cannot allocate memory
> 
> But I think that's not related with this change.

That's a funny one. Yes, I agree it's not related.

But there are a couple of broken tests (if you click in the tests tab in the top right)

image

guillep:
I'm checking if those are a side effect of what I was doing (probably not the slider ones but the others ^^).

src/Morphic-Core/Morph.class.st
```diff
@@ -5743,6 +5785,95 @@ Morph >> simulateKeyStrokes: aString [
...
+Morph >> simulateMouseMove: aPointStart to: aPointEnd [
+"Simulates a mouse moving from a given point to a destinationpoint"
+	| event |
+	event := MouseMoveEvent  new 
+				setType: #mouseMove  
+				startPoint: aPointStart 
+				endPoint: aPointEnd 
+				trail: nil
+				buttons: 0
+				hand: ActiveHand
```
guillep: this one should be `self activeHand`

guillep (changes requested):
I think this is the problem with `testUndeclared – Windows64.ReleaseTests.Release.ReleaseTest`

guillep:
I just downloaded a fresh pharo 9, and I don't have the failing tests...

src/Morphic-Core/Morph.class.st
David5i6: Hummm, is curious, that runs on my image, and the existing "simulateKeyStroke" also uses ActiveHand instead of self activeHand.

I will try to change both, but don't undertand why I don't get an error on my image.

guillep: I changed that this week... sorry for the noise.

David5i6: ok.

guillep:
I'm investigating why the CI still breaks.
I've loaded your PR and I'm checking it locally.

David5i6:
Thanks !

David5i6:
Great job @guillep  !!!
assertSomeValueAfter is great !

I'm going to push a commit with a comment and another with a change of an assert:equal: for a false value replaced by deny:

guillep:
Well it turns out that simulating the tests using the active hand is not the best. Since the CI is not actually in graphical mode, that is something we cannot control :).

In any case, I hope you're ok with my changes:
 - split big tests in small tests with one assertion per test (this was super useful to get the right feedback from the CI, to know **what** assertion was failing)
 - there were some style things (formatting code, assert/deny instead of assert:equals: boolean)
 - I made the slider delegate to the internal morph the simulation of mousedowns to avoid repetitions (since this was also defined exactly the same in Morph)

Of course, I couldn't have done this without your initial tests and comments :) Understanding the thing to was kind of straightforward, so I applied the refactoring almost blindly ^^

The test failures seem unrelated again.
We should kill those fleaky tests...

guillep (approved):

David5i6:
Everything is ok and learned some smalltalk from your refactorings and hints  ;-)

guillep:
Thanks again!

Issue #5923, opened 2020-03-17
Refactor Theme to simplify its usage


guillep:
The `UITheme` class implements a lot of factory methods for morphs, like:
```smalltalk
^self theme
		newFuzzyLabelIn: self
		for: nil
		label: aString
		offset: 1
		alpha: 0.5
		getEnabled: nil
```

Now, most (if not all) of these methods receive a first "themableMorph" argument.
It looks however that that argument is not used.
My theory is further backed up by the fact that there is a conveinence `TEasilyThemed` trait that sends `self` as first argument always. And that trait is used by objects that are not morphs and that do not share a morph/themable interface).

As an example of how this issue could be resolved, a straightforward (though long :)) refactor (that leaves the system working at each step) could be:
 - make `UITheme` use `TEasilyThemed` => like that it will automatically have the easy API
 - change all users to use the simple API that avoids the extra argument
 - all users of `TEasilyThemed` now do not need that anymore, they can use the theme and avoid the trait
 - flatten the `TEasilyThemed` trait in the `UITheme` class
 - inline each of the old methods (with extra arg) into the new methods (without extra arg)
 - deprecate old methods

sic-f:
@guillep wouldn't it be better to retain `TEasilyThemed` because there are several classes that would contain the same APIs, hence the use of `TEasilyThemed`?

<img width="610" alt="Screenshot_15_9_20__1_07_PM" src="https://user-images.githubusercontent.com/17798988/93168327-2ef04300-f755-11ea-8ea6-d2194881be63.png">

And then just rename `UITheme>>#newFuzzyLabelIn:for:label:offset:alpha:getEnabled:` to the easy API `#newFuzzyLabelFor:label:offset:alpha:getEnabled:` with the same body:

```Smalltalk
Rename UITheme>>#newFuzzyLabelIn:for:label:offset:alpha:getEnabled: to

(UITheme)
newFuzzyLabelFor: aModel label: aString offset: offset alpha: alpha getEnabled: enabledSel

    ^(FuzzyLabelMorph contents: aString font: self labelFont)
	offset: offset;
	alpha: alpha;
	model: aModel;
	getEnabledSelector: enabledSel	
```

and then call that instead in `TEasilyThemed>>#newFuzzyLabelFor:label:offset:alpha:getEnabled:` like so:

```smalltalk
(TEasilyThemed)
newFuzzyLabelFor: aModel label: aString offset: offset alpha: alpha getEnabled: enabledSel

[...]

Issue #5917, opened 2020-03-17
Should remove accesses to `currentWorld` and `ActiveWorld`


guillep:
My image has 265 senders of `currentWorld`.
And many of them are not morphs.
This introduces super messy dependencies.

We should make sure that
- objects that need access to a world, are given it by argument
- morphs should usually be able to replace `self currentWorld` => `self world`
- we should be careful with morphs accessing the world before being installed in one

guillep:
Just as a side note, related to this , most users of the theme use `self currentWorld` as first argument, but this is actually not needed :) I've just killed ~60 senders of `currentWorld` d'un coup.

Related issue: https://github.com/pharo-project/pharo/issues/5923

guillep:
Here is a script to get the big families:

```smalltalk
#currentWorld senders
	collect: [ :e | | node |
		node := e methodNode allChildren detect: [ :c | 
			c isMessage and: [ c selector = #currentWorld ] ].
		node parent isMessage
			ifTrue: [ node parent selector ]
			ifFalse: [ node parent asString ] ]
	as: Bag
```

Ducasse:
whouahhh and what are they when they are not Morph.

Ducasse:
This flu and virus is a real pity because I would love to pair program with you on that :)
This looks like fighting the plague :).

PR #5926, opened 2020-03-17
Reduce users of current world

guillep:
Partial fix of #5917 harvesting easily dropping fruit.

Getting down to 178 references.
- About half of the removed ones are non-necessary usages of currentWorld for theming (see #5923).
- Others were simple refactorings of repeated accessors to `displayScaleFactor`
- Others were `currentWorld` => `world`

On the 178 remaining references, I've identified several families. They should be easy to solve in isolation.

### DIFF
--- a/src/BlueInk-Extras/BISettingPreviewer.class.st
+++ b/src/BlueInk-Extras/BISettingPreviewer.class.st
@@ -48,7 +48,7 @@ BISettingPreviewer class >> defaultSpec [
 { #category : #settings }
 BISettingPreviewer class >> dialogOpenBIPreviewer [
 	^ self theme
-		newRowIn: self currentWorld
+		newRowIn: self
 		for:
 			{(self theme buttonLabelForText: 'Open Blue Ink Setting Previewer ').
 			self buildOpenBIInspectorButton}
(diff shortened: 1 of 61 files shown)


Ducasse:
Two little related broken tests
testIntoWorldCollapseOutOfWorld and CircleMorph.

Ducasse:
I do not really understand the impact of 

removing 	self currentWorld activeHand addEventListener: self.
and 

```
		autoAccept: autoAccept;
 		setTextWith: self initialContents.
 	self currentWorld activeHand addEventListener: self.
 	editor announcer
 		when: MorphLostFocus send: #whenEditorLostFocus: to: self;
 		when: RubTextAccepted send: #whenTextAccepted: to: self;
 @@ -229,7 +228,6 @@ RubFloatingEditorBuilder >> openEditorWithContents: aText thenDo: aBlock [

 { #category : #'announcement handling' }
 RubFloatingEditorBuilder >> whenEditorDeleted: anAnnouncement [
 	anAnnouncement morph activeHand removeEventListener: self.
 	anAnnouncement morph announcer unsubscribe: self.
 	editor := nil
 ]
```

guillep:
> I do not really understand the impact of
> 
> removing self currentWorld activeHand addEventListener: self.

Me neither. What I noticed is:
1 - removing that did not break it
2 - that class it not very used

What that line does is to redirect ALL events to that morph.
Regardless the focus, regardless the morph composition.
Which I don't think is needed for that morph either (it's actually a super corner case).

If it breaks something, we can re-add that line.

estebanlm:
I guess it may be used for modal views?

guillep:
> I guess it may be used for modal views?

Maybe, dunno...
If that was the case, there are other more elegant ways to do it (like subscribing to FocusLost :P) than listening to ALL events

PR #5931, opened 2020-03-18
Fix #5930: Write the error name and description in the body of the error too

### LINKED ISSUE #5930: Jenkins Failing tests do not show the failing exception/assertion
Right now, a failing test in jenkins shows only the textual stack trace.

In https://github.com/pharo-project/pharo/pull/5916 for example, I had the following report

image
...
image

The stack trace gives for example the idea of a time out and where it happened, but the name/printstring of the exception that lead to the failure is not there.

guillep:
Fix #5930

Write the error name and description in the body of the error too


guillep:
Ok, this is working. I had the same test failing, but now it prints the exception in the report

image

PR #5932, opened 2020-03-18
fix-double-selection-on-calypso

estebanlm:
fixes a problem when user selects two methods by cmd+click on them


guillep (approved):
Test failures not related

PR #5933, opened 2020-03-18
Fix and small enhance of ExternalAddress>>readArrayOf:until:

massimo-nocentini:
The additional `count` keeps track of the number of objects of C type `externalType` has been fetched. This could be useful, for instance, when fetching an array of floats where the one-arg block should not be sufficient to halt the fetching loop given the float alone.


guillep (changes requested):
Hi @massimo-nocentini, it would be nice to add a test for it.
Would you give it a try? It does not look very difficult.

guillep (approved):
Great! thanks for the effort!

PR #5934, opened 2020-03-18
Cleanup Morphic Render Loop

### LINKED ISSUE #5892: Morphic Render loop is duplicated in several places
Usually there should be a single render loop like (just an example):
```
[WorldMorph doOneCycle] repeat
```

It is common to implement features such as modal windows by starting a new cycle within an existing cycle. However, although such a thing should be scarse, there are many places repeating such a logic. I see some strange ones, some that I could understand (like the tests). We should review all of them in a one-by-one basis.

- senders of `#doOneCycle`

image

- senders of `#doOneSubCycle`

image

guillep:
Fix #5892 

- extract the render loop in MorphicRenderLoop
- use it in all places
- after many removals, it is only used in
   - a couple of tests (that test render, we should see in the future how to enhance them maybe)
   - the main render loop
   - modal window support
   - menu support

I've also done several cleanups:
 - opening a morph in a world now has a template method
 - SystemWindows had all the opening repeated => made them conform to Morph
 - Unify MenuMorph opening mechanisms (there were two almost copy-pasted)
 - modal opening is a concern for windows => move it to the window class, and removed conditional code.
 - have modal support in a single place

### DIFF
--- a/src/Calypso-Browser/ClyBrowserSearchDialogWindow.class.st
+++ b/src/Calypso-Browser/ClyBrowserSearchDialogWindow.class.st
@@ -102,7 +102,7 @@ ClyBrowserSearchDialogWindow >> openOn: aQuery withTitle: dialogTitle thenDo: re
 	self showItemsOf: aQuery.
 	self title: dialogTitle.
 
-	browser openModal: self.
+	self openModal.
 
 	^["Following two lines should prevent special condition 
 	when filter background process modifies current table data source.
--- a/src/Calypso-Browser/ClyBrowserSearchDialogWindow.class.st
+++ b/src/Calypso-Browser/ClyBrowserSearchDialogWindow.class.st
@@ -102,7 +102,7 @@ ClyBrowserSearchDialogWindow >> openOn: aQuery withTitle: dialogTitle thenDo: re
 	self showItemsOf: aQuery.
 	self title: dialogTitle.
 
-	self openModal.
+	browser window openModal: self.
 
 	^["Following two lines should prevent special condition 
 	when filter background process modifies current table data source.
--- a/src/FormCanvas-Core/FormCanvas.class.st
+++ b/src/FormCanvas-Core/FormCanvas.class.st
@@ -166,7 +166,7 @@ FormCanvas >> drawOval: r color: c borderWidth: borderWidth borderColor: borderC
 		fill: c
 		borderWidth: borderWidth
 		borderColor: borderColor
-		transform: transform.
+		transform: self transform.
 ]
 
 { #category : #'drawing-polygons' }
@@ -185,7 +185,7 @@ FormCanvas >> drawPolygon: vertices fillStyle: aFillStyle borderWidth: borderWid
 		fill: aFillStyle
 		borderWidth: borderWidth
 		borderColor: borderColor
-		transform: transform.
+		transform: self transform.
 ]
 
 { #category : #'private - balloon engine' }
@@ -197,7 +197,7 @@ FormCanvas >> drawRectangle: r color: c borderWidth: borderWidth borderColor: bo
 		fill: c
 		borderWidth: borderWidth
 		borderColor: borderColor
-		transform: transform.
+		transform: self transform.
 ]
 
 { #category : #'drawing-text' }
@@ -275,7 +275,7 @@ FormCanvas >> ensuredEngine [
 		engine deferred: false.
 		engine].
 	engine colorTransform: colorTransform.
-	engine edgeTransform: (transform ifNil: [MatrixTransform2x3 identity]).
+	engine edgeTransform: self transform.
 	^engine
 ]
 
@@ -706,6 +706,12 @@ FormCanvas >> stencil: stencilForm at: aPoint sourceRect: sourceRect color: aCol
 	port stencil: stencilForm
 		at: aPoint + origin
 		sourceRect: sourceRect.
+]
+
+{ #category : #'private - balloon engine' }
+FormCanvas >> transform [
+
+	^ (transform ifNil: [AthensAffineTransform new])
 ]
 
 { #category : #'drawing-support' }
--- a/src/FormCanvas-Core/FormCanvas.class.st
+++ b/src/FormCanvas-Core/FormCanvas.class.st
@@ -711,7 +711,7 @@ FormCanvas >> stencil: stencilForm at: aPoint sourceRect: sourceRect color: aCol
 { #category : #'private - balloon engine' }
 FormCanvas >> transform [
 
-	^ (transform ifNil: [AthensAffineTransform new])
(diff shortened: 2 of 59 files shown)


src/Calypso-Browser/ClyBrowserSearchDialogWindow.class.st
```diff
@@ -102,7 +102,7 @@ ClyBrowserSearchDialogWindow >> openOn: aQuery withTitle: dialogTitle thenDo: re
 	self showItemsOf: aQuery.
 	self title: dialogTitle.
 
-	browser openModal: self.
+	self openModal.
```
dionisiydk: I can't load full PR right now. How it changes the behavior? 
Previously simple #openModal behaves in the way that you can lost it by clicking on the initial browser.

guillep: Hi @dionisiydk.
> How it changes the behavior?
> Previously simple #openModal behaves in the way that you can lost it by clicking on the initial browser.

Right now what will happen is that the window will open.
And the caller method will wait until it is closed.
Clicking outside the window does not close the window, and does not make the modality return.

At least that happens for most windows. A window could close itself when it loses focus. We don't have power there :).

The main impact here is that the browser will continue answering to events.
I can fix that in this PR, no problem, but I'd need you to tell me how to get the browser window => because modality is a window concern.

dionisiydk:
```Smalltalk
browser window
```

Ducasse:
guille this code looks strange to me

```
SystemWindow >> modalLockTo: aSystemWindow [
 	"Lock the receiver as a modal owner of the given window."

 	aSystemWindow setProperty: #modalOwner toValue: self.
 	self setProperty: #modalChild toValue: aSystemWindow.
...
```

or does it means that a SystemWindow is playing also a kind of nullPattern. 
Or it means that a window can be a modal of another window - probably ok too.

guillep:
> modalLockTo:

The second. Notice also that I did not add that method.
I moved it from Polymorph-Something to the main SystemWindow package.

See:
https://github.com/pharo-project/pharo/pull/5934/files#diff-b29eac7d5acc25d2772bebdc71570116L960

<blockquote><div> GitHub</div><div><strong><a href="https://github.com/pharo-project/pharo/pull/5934">Cleanup Morphic Render Loop by guillep · Pull Request #5934 · pharo-project/pharo</a></strong></div><div>Fix #5892

extract the render loop in MorphicRenderLoop
use it in all places
after many removals, it is only used in

a couple of tests (that test render, we should see in the future how to enhance...</div></blockquote>

Ducasse:
I love to see all the Morphic cleaning.
I remember spending time cleaning the event low level array representation of event creeping into handMorph and these changes were lost :(

This eventBuf is terrible. 

processEventsFromQueue: anEventQueue

	"Process user input events from the local input devices."

	| evt evtBuf type hadAny |

	ActiveEvent
		ifNotNil: [ "Meaning that we were invoked from within an event response.
		Make sure z-order is up to date" self mouseOverHandler processMouseOver: lastMouseEvent ].
	hadAny := false.
	[ anEventQueue isNotNil and: [ ( evtBuf := anEventQueue nextEvent ) isNotNil ] ]
		whileTrue: [ evt := nil.	"for unknown event types"
			type := evtBuf first.
			type = EventTypeMouse
				ifTrue: [ self world beCursorOwner.
					recentModifiers := evtBuf sixth.
					evt := self generateMouseEvent: evtBuf
					].
			type = EventTypeKeyboard
				ifTrue: [ recentModifiers := evtBuf fifth.
					evt := self generateKeyboardEvent: evtBuf
					].
			type = EventTypeDragDropFiles
				ifTrue: [ evt := self generateDropFilesEvent: evtBuf ].
			type = EventTypeWindow
				ifTrue: [ evt := self generateWindowEvent: evtBuf ].	"All other events are ignored"
			( type ~= EventTypeDragDropFiles and: [ evt isNil ] )
				ifTrue: [ ^ self ].
			evt
				ifNotNil: [ "Finally, handle it"
					self handleEvent: evt.
					hadAny := true.	"For better user feedback, return immediately after a mouse event has been processed."
					( evt isMouse and: [ evt isMouseWheel not ] )
						ifTrue: [ ^ self ]
					]
			].	"note: if we come here we didn't have any mouse events"
	mouseClickState notNil
		ifTrue:
			[ "No mouse events during this cycle. Make sure click states time out accordingly" mouseClickState handleEvent: lastMouseEvent asMouseMove from: self ].
	hadAny
		ifFalse:
			[ "No pending events. Make sure z-order is up to date" self mouseOverHandler processMouseOver: lastMouseEvent ]

Ducasse:
Hi guille

Yes I know that this method was there before. 

S.

guillep:
Ah I just saw my quote did not quite work.

> Or it means that a window can be a modal of another window - probably ok too.

^^ This.

A window in modal will be modal to another window. One is the child the other the parent.
And they now each other through properties...

Ducasse:
I was reading a bit the code of InputEventFetcher and related and it improved
in the past the code was full of eventBuf first second third....
Igor told me that with SDL2 and OSWindows we could get rid of all the bad code. 
So I hope :)

Ducasse:
So I think that we can merge it.

Ducasse:
no? Did you fix the point raised by denis?

src/FormCanvas-Core/FormCanvas.class.st
```diff
@@ -708,6 +708,12 @@ FormCanvas >> stencil: stencilForm at: aPoint sourceRect: sourceRect color: aCol
 		sourceRect: sourceRect.
 ]
 
+{ #category : #'private - balloon engine' }
+FormCanvas >> transform [
+
+	^ (transform ifNil: [AthensAffineTransform new])
```
guillep: Ah this change is not good, I should revert it!

guillep:
> no? Did you fix the point raised by denis?

Not yet. I was reading rotten green tests and playing with the little monster :P
I'll do it as soon as I integrate Denis changes and I fix the thing I just noticed

src/FormCanvas-Core/FormCanvas.class.st
guillep: Fixed

src/Calypso-Browser/ClyBrowserSearchDialogWindow.class.st
guillep: Fixed :) Thanks for the review.

guillep:
I treated all issues.
Let's wait for the CI.

guillep:
There are related broken tests. I'm working on them.

guillep:
OSX build failed, seems a CI hiccup.
All tests are green now.

tesonep (approved):

PR #5942, opened 2020-03-19
Remove direct references to World

guillep:
Hide `World` global behind `self currentWorld` as we may have many worlds.


guillep:
PR is green :)

Waiting for a review.

tesonep (approved):

PR #5943, opened 2020-03-19
Moving debugging actions creation mechanism to GT-Debugger package

StevenCostiou:
Moving debugging actions creation mechanism from Debugger-Model package to GT-Debugger package.

Only the GTDebugger is using this now, and NewTools debugger will not.
It should not be in core packages.


guillep (approved):
Good.

Test failures unrelated.

guillep:
Thanks!

PR #5961, opened 2020-03-22
5960-ENH-add-first-simple-hoook-in-RBParser

### LINKED ISSUE #5960: ENH add first simple hoook in RBParser
The idea is that scanner next should be done in a separate method and the two users should just do self scannerNext. 
Like that subclasses do not have to do copy and paste. 
So basic design principle. So useful

Ducasse:
fixes: #5960 introduce a simple hook. the real solution could be to have a strategy token providers and that we can plug what we want.


guillep (approved):
Good!

PR #5963, opened 2020-03-23
Fix #5958

### LINKED ISSUE #5958: Drag and drop a method leads to DNU shiftPressed sent to nil
**Describe the bug**
Drag and drop a method leads to DNU shiftPressed sent to nil
Dragging a method to a different protocol. 

Screenshot 2020-03-22 at 10 59 37

**Version information**
 - Pharo 90
 - Commit info Pharo 9.0.0
Build information: Pharo-9.0.0+build.173.sha.c97646c61233fdcf1c691bfef84b97133865fedb (64 Bit)

**Additional context**
```
withPassenger: anObject from: source hand: dragHand
	| hand |
	
	hand := dragHand.
	
	^ self new
		passenger: anObject;
		source: source;
		"If the client hasn't provided a hand use the currently active hand"
		dragHand: hand;
		shouldCopy: hand shiftPressed;
                                     ^^^^^^^^^^^^^^^^^^^^^^
		position: source positionInWorld;
		yourself

Guille this may be related to some hand cleaning?

guillep:
Transfer morph needs a hand.
However, a hand is never available globally.

- Since a hand is always required, use the hand of the source (The transfer morph may not have been installed in the world yet).
- Remove the #withPassenger: variant since it is not used and creates an incorrect TransferMorph.

### DIFF
--- a/src/Morphic-Base/TransferMorph.class.st
+++ b/src/Morphic-Base/TransferMorph.class.st
@@ -43,27 +43,17 @@ TransferMorph class >> initialize [
 ]
 
 { #category : #'instance creation' }
-TransferMorph class >> withPassenger: anObject [ 
-	^ self withPassenger: anObject from: nil
-]
-
-{ #category : #'instance creation' }
-TransferMorph class >> withPassenger: anObject from: source [ 
-	^self withPassenger: anObject from: source hand: nil
+TransferMorph class >> withPassenger: anObject from: source [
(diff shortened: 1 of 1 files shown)


tesonep (approved):

guillep:
No breaking tests.
There was a VM crash on linux, which seems to be a new random problem.
I'll open an issue for that.

PR #5969, opened 2020-03-23
DebuggerMethodMapOpal-Cleanup1

MarcusDenker:
This is the first step for cleaning up DebuggerMethodMapOpal: we deprecate the methods with a clear description what to use instead.
Normally they are not API methods, only used by tools like the debugger.

After this is int the image, we can be sure that the clients all changed and *then* we can clean up more easily.


guillep (approved):
Good!

Unrelated test failure

PR #5973, opened 2020-03-23
Factorial using 2-partition algorithm

### LINKED ISSUE #5972: Factorial enhancement
While factorial implementation is very simple and surprisingly fast, there are some problems.

First is not the fastest implementation by far. (perhaps thats not the most important).

It's high recursive, so while trying to benchmark with **TimeProfiler** some scenarios in wich I have Taylor polinomies the resulting tree is hard to navigate, and need too much time to generate.

For a very extreme case:

```smalltalk

[100000 factorial] timeToRun

```

I was mading some tests and using a non-recursive aproximation the **TimeProfiles** works great.

Original factorial last for more than a minute, while non recursive version finishes in few seconds.

Also if doing some small improvements grouping elements together in pairs time for large number is almost the half of original version. ( also less GC and memory usage).

I'm going to make a pull request with my chage because I think it can improve performance and made code more easy to profile, but feel free to discard it if that does not fit into pharo philosofy.

David5i6:
fixes #5972 

Implementation of factorial using 2 partition algorithm

When using **TimeProfiler** the original recursive implementation became  so slow while it runs fast with non recursive versions.

Also this algorithm is almost twice fast than the original without usage of prime numbers.

### DIFF
--- a/src/Kernel-Tests/IntegerTest.class.st
+++ b/src/Kernel-Tests/IntegerTest.class.st
@@ -384,8 +384,14 @@ IntegerTest >> testEven [
 
 { #category : #'tests - bitLogic' }
 IntegerTest >> testFactorial [	
-	"Compare fast implementarion with existing one for the first 1000 positive integers plus zero"
-	0 to: 1000 do: [ :idx | self assert: idx factorial equals: idx slowfactorial]
+	"Check some know values"
+	self assert: 0 factorial equals: 1.
+	self assert: 1 factorial equals: 1.
+	self assert: 2 factorial equals: 2.
+	self assert: 3 factorial equals: 6.
+	self assert: 4 factorial equals: 24.
+	self assert: 5 factorial equals: 120.
+	self assert: 6 factorial equals: 720
 ]
 
 { #category : #'tests - printing' }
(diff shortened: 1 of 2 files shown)


guillep:
Without much to say, I wonder what the people around polymath say.
@SergeStinckwich , @olekscode , if you have 10 minutes to give an opinion during these days, it would be great :)

Ducasse:
```
^ self < 0
 				ifTrue: [ self error: 'Not valid for negative integers' ]
 				ifFalse: [ self > 0
 						ifTrue: [ ^ self ]
 						ifFalse: [ ^ 1 ] ] ].
```

this code has too many returns :)

```
^ self < 0
 				ifTrue: [ self error: 'Not valid for negative integers' ]
 				ifFalse: [ self > 0
 						ifTrue: [ self ]
 						ifFalse: [  1 ] ] ].
```

Ducasse:
can you rename factorial as slow factorial and add a comment stating that we know that is is slow but that we keep it for the fun.

David5i6:
I added some tests, comments and moved original to slowfactorial.

I've seen other comments that point to let original as is, and put new as fastfactorial....

It's only a matter of little refactoring to rename them.

src/Kernel-Tests/IntegerTest.class.st
```diff
@@ -1101,6 +1107,18 @@ IntegerTest >> testRounding [
 	self assert: (5 round: 2) equals: 5
 ]
 
+{ #category : #'tests - bitLogic' }
+IntegerTest >> testSlowFactorial [
+
+	"Check some know values"
```
olekscode: "Check some **known** values"

src/Kernel-Tests/IntegerTest.class.st
```diff
@@ -382,6 +382,12 @@ IntegerTest >> testEven [
 	
 ]
 
+{ #category : #'tests - bitLogic' }
+IntegerTest >> testFactorial [
```
olekscode: I think it is better to confider `factorial` as the default implementation. So instead of asserting if `factorial` is equal to `slowFactorial`, I think it is better to have a normal test here, and then in `testSlowFactorial`, we can assert whether `slowFactorial` is equal to `factorial`. This way we would reverse the dependency and it will be `slowFactorial` that depends on `factorial` and not the other way around. It is good because `slowFactorial` could be removed or renamed in the future.

Ducasse:
could you rename slowfactorial as slowFactorial?

pavel-krivanek:
it would be more newcomers friendly if it would be `factorialSlow` because then these methods would be next to each other...

David5i6:
After reading comments I've made some small changes. Renamed *slowfactorial" to "slowFactorial".
Tests of factorial does not have dependency of slowFactorial.
Also a small (unnoticeable) improvement in slowFactorial changing order of the ifs.

guillep:
I think this is good enough. We can always improve later.

Thanks @David5i6 !

PR #5978, opened 2020-03-24
make Fonts-Infrastructure independent on FreeType

pavel-krivanek:
motivation: Fonts-Infrastructure is the dependency of Spec2-Core so it cleans Spec dependencies


pavel-krivanek:
I will need to check manually the artifacts if they display fonts correctly

tesonep (approved):
The changes looks nice.

tesonep:
Let's wait for the output of the bootstrap process

pavel-krivanek:
the resultant image seems to work ok

PR #5979, opened 2020-03-24
Make Glamour Bricks implement morphic event interface

### LINKED ISSUE #5971: activeHand DNU error while trying to open the "bindings" window in playground
You get an activeHand DNU error while trying to open the "bindings" window in playground (it also happens with the list of saved playgrounds).

**To Reproduce**
1. open playground.
2. click on "bindings" window.

**Expected behavior**
To not have a DNU.

guillep:
Fix #5971

Glamour bricks do handle events in a funny way.
Instead of redefining the hooks defined by morphic, they listen to all events in the system and have their own way to decide whether an event applies to them or not.

This PR makes GLM*Brick handle normal events instead of listen the hand.


tesonep (approved):

guillep:
Tests failures are unrelated

PR #5980, opened 2020-03-24
Clean cyclic dependencies between Keymapping packages

pavel-krivanek:
Clean cyclic dependencies between Keymapping packages.  They are, of course, still logically dependent on each other but now they do not require atomic loading


pavel-krivanek:
unrelated `MacOSX64.Kernel.Tests.Processes.LocalRecursionStopperTest` failure

PR #5981, opened 2020-03-24
Menu morph should open on the active hand of the opening world

### LINKED ISSUE #5975: activeHand DNU when executing cmd+t in editor
**Describe the bug**
I have an activeHand DNU when executing cmd+t in editor.

**To Reproduce**
1. Open browser on any method.
2. Press cmd+t 

**Expected behavior**
To not have a DNU :)

guillep:
Fix #5975

Use the active hand of the same world we are spawning into.
Menu morph does not have an active hand, since it has not been open in a world yet.

### DIFF
--- a/src/Morphic-Base/MenuMorph.class.st
+++ b/src/Morphic-Base/MenuMorph.class.st
@@ -914,7 +914,7 @@ MenuMorph >> invokeModal [
 	menu invokeModal."
 	
   ^ self
-		invokeAt: self activeHand position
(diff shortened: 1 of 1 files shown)


tesonep (approved):
Maybe we need to pass the world where we want to open it, but that is a bigger refactoring.

guillep:
> Maybe we need to pass the world where we want to open it, but that is a bigger refactoring.

The problem are those that open a menu and are not morphs and do not know a world O_o

PR #5982, opened 2020-03-24
Restore deferredUpdateMode in case of nested render loops

### LINKED ISSUE #5976: DNU in VMWorldRenderer>>#updateDamage:
**To Reproduce**
1. Open Calypso, go to Text-Core package
2. browse dependencies
3. open left tree

Pharo_4tbhRU2OWA

alternativelly, it opens emergency evaluator

### LINKED ISSUE #5977: progress bars show only red box of death
to reproduce:
```smalltalk
(1 to: 100) do: [ :i | 2 milliSeconds wait  ] displayingProgress: 'Title'
```

It is strange that it does not try to display any error text inside...

Pharo_Pr9GU2vQm7

guillep:
Fix #5976.
NOT - Fix #5977 probably also.

In case of nested render loops, the deferredUpdateMode should be kept and restored correctly.

@pavel-krivanek could you check this fixes both issues?

### DIFF
--- a/src/Morphic-Core/VMWorldRenderer.class.st
+++ b/src/Morphic-Core/VMWorldRenderer.class.st
@@ -97,6 +97,8 @@ VMWorldRenderer >> deferUpdates: aValue [
 { #category : #operations }
 VMWorldRenderer >> deferUpdatesDuring: aBlock [
 
+	| previousDeferredUpdateMode |
+	previousDeferredUpdateMode := deferredUpdateMode.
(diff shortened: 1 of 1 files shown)


pavel-krivanek:
Thanks, it fixes the dependency analyzer but not the progress bars

guillep:
Thanks @pavel-krivanek . I'll ask you then more input on the other issue if you have any, because i cannot reproduce it with just that expression.

PR #5984, opened 2020-03-24
Increase hiedra test time limit

### LINKED ISSUE #5938: Fix Hiedra Fleaky test
We have from time to time these fleaky tests in the CI:

image

They make noise in the run.

guillep:
Fix #5938
Increase time limit to avoid random failures in the CI


tesonep (approved):
Yes... If this continue maybe we have to remove it, it is a bad bad bad test.

PR #5986, opened 2020-03-24
Add specific pharo8 template

guillep:
Added template with a default Pharo8.0 label for issues on the stable version.


guillep:
I've seen in the last couple of days that adding issue templates was working.
This adds a new one for Pharo8 issues that automatically adds the Pharo8.0 label.

PR #5992, opened 2020-03-25
Implementing a new iteration on top of spotter.

tesonep:
In Highlights it includes: 
 
- A set of processors implemented using generators and composable filters, so they don't need to traverse the whole image on a query.
- Introduction of a query object representing complex queries that can be done (like # operators and splitting in words).
- Adding the concept of SptEntry to avoid adding extension methods to all the classes in the system.
- Removing the use of pragma based processors in the main spotter to enhance the experience as looking up pragmas is too heavy.
- Adding a spinner to show activity in the query, and a help message to provide help to the new users.
- Better handling of the system settings and more ordered elements, showing a more concise and useful view. 
- The processors' results are collected iteratively, so it has a better response to the user.
- Extensible to support the use of indexes.


dionisiydk:
Could you comment here concrete changes about new query object? (in review). It is difficult to see these interesting parts.

tesonep:
Hi @dionisiydk, it just adding an object to represent the query entered by the user. 
For example in spotter is possible to put #Classes to scope the search into classes, or adding #Packages to add packages. What I have done is to add an object that parses this query string and generates a nice object you can ask stuff. So, I have taken the logic that was in the processor and put it in the query. 

It just parses the string and identifies the elements, then these elements are used or not by the processor to resolve the results to show. 
The SptQuery does not execute the query itself or handle the results, but it just parses the query string.

tesonep:
One of the objectives was to minimize the distribution of string operations all over the implementation. 
Still, we need to parse the string, but it is better to do it in a single place.

Ducasse:
Pablo the build failed so I kicked in again.

tesonep:
Thanks... what strange error. Let's see if that happens again.

pavel-krivanek:
one issue that is probably not related to your code: in `GLMScrollPaneBrick>>#unsubscribeListeners`, the 'currentWorld' should be used, else we can get this issue:

TnTHhk9mbb

pavel-krivanek:
Question: would it be possible now to implement:
- option to open Spotter search result in a standalone Calypso window (for example, if you press Ctrl+M twice)
- search strings in methods, search in image source codes

pavel-krivanek:
on Window, Ctrl+M works but Ctrl+N adds "N" at the end of the string (but seems to perform search anyway)

pavel-krivanek:
maybe older issue: if you display code preview (Ctrl+P) and try to use context menu on some code (RMB), it ends on `Instance of GLMPhlowPharoMethodBrick did not understand #menu`

pavel-krivanek:
There are dependency issues, The GT-Spotter-Processor is depending on GT-Spotter-UI, Morphic-Base and Morphic-Widgets-Windows. 
GT-Spotter is depending on GT-Spotter-UI and GT-Spotter-Processor (cyclic dependencies)

pavel-krivanek:
Please, check Critic Browser results for the package ` GT-Spotter-Processor`, there are some minor issues like inconsistent classification etc.

tesonep:
Thanks, I will check the issues!

tesonep:
> Question: would it be possible now to implement:
> 
> * option to open Spotter search result in a standalone Calypso window (for example, if you press Ctrl+M twice)
> * search strings in methods, search in image source codes

Yes, why not... I will check how expensive it could be.

guillep:
@pavel-krivanek Could you check if the concerns you raised were fixed in Pablo's commits?

pavel-krivanek:
With the returning back in the spotter navigation, I still get an error, this time: `#removeEventListener: was sent to nil`

tesonep:
Is it ok to integrate @guillep @pavel-krivanek ?

pavel-krivanek (approved):
ok for me

PR #6007, opened 2020-03-27
[cleanup] changed "message"  by "method" in the template method

hogoww:
This might be controversial, but i'm going for it !

messageSelectorAndArgumentNames
   "comment stating purpose of instance-side message"
(...)

In the current template we say we define a message, whereas as far as I understand, we're defining a method.

I therefore replaced it with
methodSelectorAndArgumentNames
   "comment stating purpose of instance-side method"

Let me know if you think this is right or not =)

### DIFF
--- a/src/Kernel/Behavior.class.st
+++ b/src/Kernel/Behavior.class.st
@@ -1587,8 +1587,8 @@ Behavior >> sourceCodeTemplate [
 	"Answer an expression to be edited and evaluated in order to define 
 	methods in this class or trait."
 
-	^ 'messageSelectorAndArgumentNames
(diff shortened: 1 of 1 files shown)


pavel-krivanek (approved):
The name "method" makes sense to me here

Ducasse:
I agree with pavel we are editing a method :)

Ducasse:
So I will close it :) but thanks Pierre.

hogoww:
wat.
We all agree then, don't we?
Currently in the image, the messages is using "message"
I changed it to have "method"
No..?

MarcusDenker:
Yes, it should be merged, not closed?

PR #6015, opened 2020-03-27
Use introduced ifExist:(ifAbsent:)

jecisc:
Add #ifExists:ifDoesNotExists: on files to simplify code.


dionisiydk:
Just a note about method name.
Maybe ifExists:ifAbsent: would be a better name? And is ~s really needed here. I just think about other similar method like at:ifPresent:ifAbsent:.

jecisc:
Maybe #ifExists:ifAbsent: can be a good alternative. I have no strong opinion on this point.

For the "s", since I'm not native English speaker I don't know but the version with the s seems more common https://books.google.com/ngrams/graph?content=if+exists%2Cif+exist&year_start=1800&year_end=2008&corpus=15&smoothing=3&share=&direct_url=t1%3B%2Cif%20exists%3B%2Cc0%3B.t1%3B%2Cif%20exist%3B%2Cc0#t1%3B%2Cif%20exists%3B%2Cc0%3B.t1%3B%2Cif%20exist%3B%2Cc0

pavel-krivanek:
+1 for ifExists:ifAbsent:, if you check the current methods with similar name, all use ifExists:

jecisc:
Done!

src/FileSystem-Core/AbstractFileReference.class.st
```diff
@@ -530,6 +530,34 @@ AbstractFileReference >> humanReadableSize [
...
 
+{ #category : #delegated }
+AbstractFileReference >> ifAbsent: aBlock [
+	"Return self, or evaluate the block if I do not exist"
+
+	^ self exists ifTrue: [ self ] ifFalse: [ aBlock cull: self ]
+]
+
+{ #category : #delegated }
+AbstractFileReference >> ifAbsent: anotherBlock ifExists: aBlock [
```
dionisiydk: I would probably do not add another combination of words, but ok

dionisiydk (approved):

src/FileSystem-Core/AbstractFileReference.class.st
jecisc: It made sense when we were copying the pattern for "ifTrue:ifFalse", "ifNil:ifNotNil:", "ifEmpty:ifNotEmpty:" but I agree this one seems weird. I'll remove it!

jecisc:
The build fails but there is no error message :(

Ducasse:
Cyril I rekicked the build.

Ducasse:
I do not understand why this PR would fail but for now the build fails.

Ducasse:
Ok now I understand. 
What would be good is to separate the introduction of the new methods from their usage.

guillep (approved):
Good!

jecisc:
I'll separate it

jecisc:
https://github.com/pharo-project/pharo/pull/6044

guillep:
Wait. I don't understand. Why creating #6044? Should I integrate #6044 first?
This PR and #6044 overlap? It seems so!
 => Doing these kind of moves without communicating well is not good

I'd say that the split, if desired, is done in future changes.
I personally like self contained issues, otherwise we lose track.

jecisc:
If you read the previous comments you can see that there is a failure in the build and Steph proposed to split the PR to first integrate the introduction of the new methods before using it in the system.

guillep:
> This PR and #6044 overlap? It seems so!

Sure, but still this does not answer my question. Why do they still overlap then?

Anyways, I've checked the log. It seems this change is creating an infinite loop after loading morphic:

```
[64] NewUndeclaredWarning: NonInteractiveUIManager>>currentWorld (World is Undeclared)
[64] MetacelloNotification: Loaded -> Morphic-Core-tonel.1 --- tonel:///builds/workspace/uest_and_branch_Pipeline_PR-6015/64/src --- cache
[64] Sending interrupt signal to process
[64] /builds/workspace/uest_and_branch_Pipeline_PR-6015/64/bootstrap-cache/vmtarget/pharo: line 11: 30620 Terminated              "$DIR"/"pharo-vm/pharo" --nodisplay "$@"
Cancelling nested steps due to timeout
```

I recommend to try to reproduce the build locally and check what the infinite loop is.

jecisc:
This PR now only includes the new invocations of the introduced methods.

Issue #6012, opened 2020-03-27
Click outside of completion pop-up should close it


pavel-krivanek:
**To Reproduce**
1. start writing some code that opens completion pop-up with some existing options
2. press arrow down key to move focus to pop-up (for the first item, it will not have any visual effect which we may consider as a bug)
3. click outside of the pop-up window with mouse

**Expected behavior**
The pop-up should be closed

pavel-krivanek:
It is actually much worse....
To prevent closing of the window by mouse click outside it is enough to just move the cursor over the pop-up for a moment.
Moreover, it remembers this state for the same editor window so even if you close it by Esc and then you write some other code that opens the pop-up, the pop-up cannot be closed by clicking outside of it even if you do not enter it with the mouse or press an arrow key

pavel-krivanek:
hmm, it depends if you click on the related editor or not...

guillep:
Just to add more info.
This was working on build 162. Stopped working with my hand changes in 163.

In Build 162, when we click outside the completion menu, the stack that closes the menu looks like:

```
SmallInteger(Object)>>halt
CompletionEngine>>closeMenu
MessageSend>>value
MessageSend>>cull:
MessageSend>>cull:cull:
[action cull: anAnnouncement cull: announcer] in AnnouncementSubscription>>deliver: in Block: [action cull: anAnnouncement cull: announcer]
BlockClosure>>on:do:
BlockClosure>>on:fork:
AnnouncementSubscription>>deliver:
[ subscription deliver: anAnnouncement ] in SubscriptionRegistry>>deliver:to:startingAt: in Block: [ subscription deliver: anAnnouncement ]
BlockClosure>>ifCurtailed:
SubscriptionRegistry>>deliver:to:startingAt:
SubscriptionRegistry>>deliver:to:
SubscriptionRegistry>>deliver:
Announcer>>announce:
RubEditingArea(Morph)>>doAnnounce:
RubEditingArea(Morph)>>announceKeyboardFocusChange:
RubEditingArea(Morph)>>keyboardFocusChange:
RubEditingArea(RubAbstractTextArea)>>keyboardFocusChange:
HandMorph>>newKeyboardFocus:
RubTextScrollPane(Morph)>>takeKeyboardFocus
NECMenuMorph>>mouseDown:
NECMenuMorph(Morph)>>handleMouseDown:
MouseButtonEvent>>sentTo:
NECMenuMorph(Morph)>>handleEvent:
[...]

PR #6017, opened 2020-03-27
6012-Click-outside-of-completion-pop-up-should-close-it

pavel-krivanek:
introduce new annoucement MorphClosePopups and use it for explicit closing of NECMenuMorph when clicking  on Rubric text area

fixex #6012


pavel-krivanek:
test failures not related to this PR

myroslavarm (approved):
wow i don't remember this bug before, so seems like it's fairly recent. anyway this is super cool to have it fixed

PR #6026, opened 2020-03-27
6002-Generate-accessors-on-inst-var-does-not-work

carolahp:
Setting the environment for the existent change model instead of instantiating a new model

### DIFF
--- a/src/SystemCommands-RefactoringSupport/SycRefactoringPreview.class.st
+++ b/src/SystemCommands-RefactoringSupport/SycRefactoringPreview.class.st
@@ -128,7 +128,7 @@ SycRefactoringPreview >> generateChanges [
 	changes := command asRefactorings.
 	rbEnvironment := self activeRBEnvironment.
 	changes do: [ :each | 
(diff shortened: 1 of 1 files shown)


pavel-krivanek:
it does not seem to fix the issue

carolahp:
> it does not seem to fix the issue

It works for me in both cases: generating variables from the context menu of the class and of the instance variable. I'm using a fresh Pharo9 image downloaded this morning. 

Here's a screenshot of generating accessors with the fix I propose in the PR:
Screen Shot 2020-03-27 at 14 51 51

Here's the same image with the fix commented:
Screen Shot 2020-03-27 at 14 57 57


Could you please give more details on how are you testing?

pavel-krivanek:
I just tried to apply the method change to my image, let's wait on the build artifact (it failed so I restarted it)

pavel-krivanek:
sorry, I realized that there was a copy&paste error, it really works, thank you... let's wait for the test results

carolahp:
That's great that it worked for you, thanks.
Let's wait for the tests

pavel-krivanek (approved):

Ducasse:
thanks you both!

PR #6031, opened 2020-03-27
Add this context tester on the AST

hogoww:
I don't know whether that make sense or not.
I assume if it did, it would have been put in already.
But since I'm working on this topic, I wanted to offer it, for API consistency.

What do you all think about it? :)


hogoww:
mixed several PR, redo : 6033

PR #6032, opened 2020-03-27
prevent isInstance/isGlobal to throw an error on semantically analyses nodes

hogoww:
This allows to not rely on binding for instanceVariable/global nodes.
This has a minimal constant memory cost,
No time cost.
This should not impact users that were working before. (unless there's a bug in the semantic analysis)
I did not do tests, because I want to wait for people's feelings about this first, given that there is several ways to fix this issue.
Let me know :)

### DIFF
--- a/src/AST-Core/RBArgumentNode.class.st
+++ b/src/AST-Core/RBArgumentNode.class.st
@@ -30,3 +30,13 @@ RBArgumentNode >> isBlockVar [
 
 	^ (self whoDefines: self name) isBlock
 ]
+
+{ #category : #testing }
+RBArgumentNode >> isGlobal [
+	^ false
+]
+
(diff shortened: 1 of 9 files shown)


dionisiydk:
It would be nice if we could improve all those names because it does not make any sense when we ask other non variable nodes about those questions:
```Smalltalk
aMessageNode isGlobal
aCascadeNode isInstance.
```
I propose to rename them (with deprecation):
- isGlobal -> isGlobalVariable
- isInstance -> isInstanceVariable
- isTemp -> isTempVariable
- maybe there are other methods

hogoww:
redone in #6034 to remove missed commit
I'll answer there Denis, if you could put your comment there too

PR #6033, opened 2020-03-27
[Opinion required] Add this context tester on the AST

hogoww:
I don't know whether that make sense or not.
I assume if it did, it would have been put in already.
But since I'm working on this topic, I wanted to offer it, for API consistency.

What do you all think about it? :)
(redo of previous PR, several commit got mixed in it)

### DIFF
--- a/src/AST-Core/RBProgramNode.class.st
+++ b/src/AST-Core/RBProgramNode.class.st
@@ -527,6 +527,11 @@ RBProgramNode >> isTemp [
 ]
 
 { #category : #testing }
+RBProgramNode >> isThisContext [
+	^ false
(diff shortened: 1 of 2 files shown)


StevenCostiou:
Hi, do you have an example of use?
I imagine it could be used by tools that look over the AST...?

hogoww:
I don't.
When I need to treat thisContextNodes, it's usually with other stuff, and I therefore use a visitor.

guillep (approved):
I think consistency is good.
And cheap.

guillep:
Thanks @hogoww and @StevenCostiou for the review!

PR #6034, opened 2020-03-27
[opinion required]prevent isInstance/isGlobal to throw an error on semantically analyses nodes

hogoww:
This allows to not rely on binding for instanceVariable/global nodes.
This has a minimal constant memory cost,
No time cost.
This should not impact users that were working before. (unless there's a bug in the semantic analysis)
I did not do tests, because I want to wait for people's feelings about this first, given that there is several ways to fix this issue.
Let me know :)

Edit:
I missed the fact that they were missing on RBProgramNode. Added them in a new commit.


dionisiydk:
It would be nice if we could improve all those names because it does not make any sense when we ask other non variable nodes about those questions:
```Smalltalk
aMessageNode isGlobal
aCascadeNode isInstance.
```
I propose to rename them (with deprecation):
- isGlobal -> isGlobalVariable
- isInstance -> isInstanceVariable
- isTemp -> isTempVariable
- maybe there are other methods

hogoww:
Actually most of the AST tester have not good names.
They also don't always follow the same naming  convention making it hard to use at times.
But they are users to deal with, as for a lot of other problems of the AST.

dionisiydk:
I think with deprecation we can easily improve this part

hogoww:
I missed the fact that they were missing on RBProgramNode.
Added them for consistency again.

MarcusDenker:
This would make the current scheme a little bit more robust, for this case:
- you did do name analysis, RBVariableNode instances where changed to the specific (e.g. RBGlobalNode)
- then you remove the semantic information (binding all nil)

For the case that no semantic analysis in done in the first place, it still leads to errors (as the fallback is implemented to forward to the binding)

What I do not like is to use add isGlobal  and friends to the RBProgramNode... the selectors are very specific and make not much sense outside of already knowing that it is a variableNode. It makes more sense if we change and use selectors like #isGlobalVariable

MarcusDenker:
I think we should do the following:

-> implement the is* methods on RBVariableNode forwarding to "self variable"
-> inside #variable, do a nil check and raise an error explaining that the method can only be called on analysed ASTs

I am tempted to add all the is*Variable methods to RBProgramNode, too, as we see very often the pattern
aNode isVariable and: [anNode isGlobalVariable] and similar. Now that the is* methods contains the -Variable at the end, it makes sense...

I close this and continue to refactor step by step.

PR #6035, opened 2020-03-27
#isArg add comment to tell users to not use it

hogoww:
isArg and isTemp have the same convention but they express different things.
there is both isArg that is the Compiler's view on the AST (and therefore relies on compiler stuff)
and isArgument which expresses the AST point of view.

isTemp is (as of this morning) not relying on the compiler.

So naming is still bad, but at least there is a little bit of explanation.

Also removes an  (I'm pretty sure) erroneous user of isArg


src/OpalCompiler-Core/RBVariableNode.extension.st
```diff
@@ -13,7 +13,8 @@ RBVariableNode >> binding: aSemVar [
 
 { #category : #'*opalcompiler-core' }
 RBVariableNode >> isArg [
-
+	"This is a compiler specific function, requiring a binding.
+	If you're working on the AST, you should use probably isArgument"
```
guillep, line 17: I've just checked at a non-semantically analysed AST. I did

```smalltalk
method parseTree.
```
And I got

image

If you look, it does not use argument nodes. So actually `#isArg` and `#isArgument` are both compiler dependent. None of them is valid without a semantic analysis. The first one requires a binding, the second one requires the replacement of the node...

src/GeneralRules/ReAssignmentOnBlockArgumentRule.class.st
```diff
@@ -32,7 +32,7 @@ ReAssignmentOnBlockArgumentRule class >> uniqueIdentifierName [
 ReAssignmentOnBlockArgumentRule >> check: aMethod forCritiquesDo: aCriticBlock [
 	| problemAssignments|
 	problemAssignments := aMethod ast assignmentNodes 
-		select: [ :assignment | assignment variable isArg].
+		select: [ :assignment | assignment variable isArgument ].
```
guillep: Does this change anything?

guillep (commented):
For the future we should check if the semantic analysis is a concern of the compiler or not, because many tools seem to rely on it.

src/OpalCompiler-Core/RBVariableNode.extension.st
hogoww: On principle yes. But the compiler is not the only manipulating AST, and therefore let's call them "semantically analyses nodes".

src/GeneralRules/ReAssignmentOnBlockArgumentRule.class.st
hogoww: Changes that if an ASTs is installed which has no binding, this wont throw an error.
I'm not sure if ASTs can be installed by someone else but the compiler, but I think we should still go with the right semantic method when we can..

hogoww:
Agreed, that often comes in our discussions with Marcus, the later one being this morning

dionisiydk:
I always had impression that changing the class of nodes during semantic analysis is a big smell.

Alternative would be to just delegate all semantics dependant questions to binding variable. Then we will not have such problems like here. There will be always same AST structure but we will inject specific bindings during semantics analysis. That would be a big change of course and probably it would have own drawbacks but it's good to think about alternatives.
From the other side AST-visitors can be kept compatible quite easily with extra delegation to bindings. And while most of ast users are visitors it could be not so hard change.

Ducasse:
Yes guille was proposing the same: a little delegation.

hogoww:
Removed conflict of updating a selector, which is not relevant anymore, as it the guilty class was removed.

this is now just a comment (cf first commit) about usage of #isArg.

MarcusDenker:
I was looking at this finally.. sorry for not doing it before.

For #isArg, we should get rid of isArg before and replace it with #isArgument (no real change but less methods doing the same is good). They are both the result of the semantic analysis.

The new Variable hierarchy now is in sync with what we want: args are not temps and both are LocalVariables. 
That should make things much easier.

I will as a fist step deprecate it on the level of the Variable hierarchy and introduce isArgument there. Then we can fix the AST level in another step.

Issue #5571, opened 2020-01-24
testDefaultImageDoesNotStoreAnySetting failing in Pharo9 on CI (all plattforms)


MarcusDenker:
Error
Given Collections do not match! additions : an Array(a StoredSetting('#useSecondarySelection#secondarySelectionTextColor') a StoredSetting('#Rubric#backgroundColor') a StoredSetting('#''Code Completion''#backgroundColor') a ClassStoredSetting('#appearance#uiThemeClass') a StoredSetting('#LineNumbers#lineNumbersTextColor') a StoredSetting('#menu#menuColor') a StoredSetting('#LineNumbers#lineNumbersBackgroundColor') a StoredSetting('#Rubric#textColor')) missing: #()
Stacktrace
SystemSettingsPersistenceTest(TestAsserter)>>assert:description:resumable:
SystemSettingsPersistenceTest(TestAsserter)>>assert:description:
SystemSettingsPersistenceTest(TestAsserter)>>assertCollection:hasSameElements:
SystemSettingsPersistenceTest>>testDefaultImageDoesNotStoreAnySetting

MarcusDenker:
Interestingly it is green when I run it locally with the latest Pharo9

tesonep:
To reproduce it should be run without preferences, I am uploading a PR for it

PR #6038, opened 2020-03-27
6036-Test-should-not-rely-on-random-formatter

Ducasse:
do not hard BI in default settings.

### DIFF
--- a/src/AST-Core-Tests/RBFormatterTest.class.st
+++ b/src/AST-Core-Tests/RBFormatterTest.class.st
@@ -78,7 +78,7 @@ RBFormatterTest >> testParseError [
 	"parse error nodes should have the faulty code"
 	inputSource := ')'.
    errorNode := self parseFaultyExpression: inputSource.
-	self assert: errorNode source equals: errorNode formattedCode
+	self assert: errorNode source equals: (self formatterClass format: errorNode)
 
 ]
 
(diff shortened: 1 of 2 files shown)


guillep (approved):
Good!

guillep:
There is related broken test, failing in the three platforms equally:

```
SystemSettingsPersistenceTest >> testDefaultImageDoesNotStoreAnySetting
```

Ducasse:
I tried to understand but I have no idea about how this test is working. 
I do not even understand where these settings are stored and sorry but these tests are not self documented.

Ducasse:
And I do not get why there are 10 elements while apparently one only is related to BIformatter.
Seriously this is totally arcane to me.

guillep:
Probably related to

https://github.com/pharo-project/pharo/pull/5572
https://github.com/pharo-project/pharo/issues/5571
https://github.com/pharo-project/pharo/pull/5473
https://github.com/pharo-project/pharo/issues/5434

Settings that make this test fail make images in the pharo launcher behave strangely.
Probably those tests deserve better comments, yes.
I don't know the details.

Ducasse:
I know that this is linked with the idea that we should not save setting when they did not change or something like that but after 30 min I could not understand the factory and the tests. :(

Ducasse:
I think that I will need pablo to have a look because I'm lost there.

tesonep:
I have changed back the default formatter in the settings to BIConfigurableFormatter
As it is initialized by BaselineOfIDE

Ducasse:
ah but I want to be able to unload BIConfigurableFormatter so I will check.

Ducasse:
I'm confused. I put the simple formatter because the Configurable to be optional and we should unload it.

Ducasse:
So should I merge my changes then later we address the problem of the reference to the configurationFormatter?

Issue #5945, opened 2020-03-20
Endless creation of UI windows


myroslavarm:
In the latest Pharo 9 since at least yesterday afternoon if you try to change a method protocol and you get a window to select the name you want, you start getting the same window for every click you do, whereas it's for 'ok', 'cancel', or any click anywhere in the image. And no way to get rid of them!

![image](https://user-images.githubusercontent.com/24677542/77168253-cf6fee00-6ab7-11ea-8f54-329abd634a01.png)

I assume this is Spec related @guillep

guillep:
> I assume this is Spec related @guillep

Wrong assumption, this has nothing to do with Spec :)

I cannot reproduce it. Can you give more details?

What I did:
 - took latest image
 - I created a new package
 - I created a class in it
 - I created a method in it (as yet unclassified)
 - I clicked on the protocol => cmd + R 

Then clicking on OK, cancel or Browse works correctly.

I then repeated the same but from the button in the bottom of the window.

![image](https://user-images.githubusercontent.com/708322/77170217-01368400-6abb-11ea-971d-ff2deb551e96.png)

It worked again.

I then tried to change the method protocol by doing "Classify methods" from the method.
It worked again.

I then tried to rename the protocol by doing. Right click -> rename. On the protocol (note, without shortcut).
Worked again.

I need more details...

myroslavarm:
I always change the protocol by clicking on:
![image](https://user-images.githubusercontent.com/24677542/77171152-51faac80-6abc-11ea-99f6-4174036a7d29.png)
and when the selection pops up, i choose a name from the list, and the UI window starts multiplying after every click after. It is the same even as i created a new package, class, method and tried to classify a newly created method.

P.S. i don't know if it might influence something, but i'm using windows :)

myroslavarm:
I don't think this was happening yesterday in the early afternoon,
but in the evening it was for sure there.

guillep:
What commit is your image built ?

Please, issue reports should at least contain:
 - pharo version (+ commit)
 - platform
 - how to reproduce
 - extra information that could be useful

:(

myroslavarm:
pharo 9 64 bit, build: 171, commit: 1b24cb3
windows 10

everything about reproducing is already in there

guillep:
I'm on the same commit

Pharo 9.0.0
Build information: Pharo-9.0.0+build.171.sha.1b24cb39f45f8659fa6c435d7ffc52144c8e7ad2 (64 Bit)

[...]

PR #6039, opened 2020-03-27
Queue pending mouse events

### LINKED ISSUE #6003: change on menu activation on list
May be this is not a bug. 
Now I noticed that in Epicea if I select a list of items and select one action, the action is only happening on the items currently getting the action (because when I select the menu item it deselect the list).
I experimented a bit and to have the action on the selected items I have to keep shift or option pressed. 

And I wonder if this is not related to the event change.

guillep:
Fix #6003

This PR should fix #6003 without having regressions in #5945 

 - refactor processEventsFromQueue:, split obtention of events in nextEventFrom:
 - add pendingEventQueue and queuePendingEvent:
 - make nextEventFrom: give priority to pending events
 - make MouseClickState queue pending mouseUp for next cycle

# Details

This solution restores the order of events (MouseDown, Click, MouseUp) while, not blocking MouseUp if a click starts a modal loop.
Changing the order of events like I did in #5964 is not a good solution because the event order makes sense for clients, and there are clients handling both clicks and mouseUps.
Also altering the order just moves the problem to mouseUp: If somebody tries to use a modal window on mouseUp => Boom same problem but with clicks.

This solution ensures that:
1) the order of events is the right one
2) no events get "blocked", so if an artificial event (e.g., click, double click) is created it can be queued and it will be managed in the next event cycle (from the same loop of from an inner modal loop).

## Background: How Mouse Events Work

The Hand is the one managing/dispatching the events.
When the Sensor (running in a higher priority thread) finds events, it puts them into a queue.
Eventually, the hand dequeues events and treats them.

Now, the mouse events have two main flavours: mouseDown and mouseUp
There is no click, nor double click event, they are simulated by the hand morph.

When a morph is interested on a click it has to
1. subscribe to mouse down first
2. when it receives a mouse down, it has to ask the hand "Hey! I would like clicks!"
Usually done by some code like this:

```smalltalk
anEvent hand waitForClicksOrDrag: h
  event: (anEvent transformedBy: tfm)
  selectors: { nil. nil. nil. #dragTarget:. }
  threshold: 5
```

When that happens, the hand changes its state, and stores a "MouseClickState object" that implements a kind of state machine to detect clicks and double clicks.

After the mouseDown, the morph receives a mouseUp (that should be transformed into a click).
So the hand, before sending the event to the morph, it makes it pass through the state machine which has some code like this:

```smalltalk
self click.    
aHand handleEvent: firstClickUp.
```

Where
1. click  will send a click event to the morph and
2. handleEvent: will send the mouseUp

## More details on the problem

One problem happens when a morph decides to BLOCK on a modal window on the click- :smile:
That means that mouseUp event will stay blocked in the stack, because the click does not return from the modal until modality finishes.
This means `handleEvent:` is never sent

In my fix the other day, I exchanged the order.
```smalltalk
aHand handleEvent: firstClickUp.
self click.
```

That solved that first problem. But that alters the order of the events.
So morphs used to receive MouseDown, Click, MouseUp now receive a MouseDown, MouseUp, then Click. This makes FastTable event logic loose the multiple selection.

### DIFF
--- a/src/Morphic-Base/Form.extension.st
+++ b/src/Morphic-Base/Form.extension.st
@@ -12,13 +12,13 @@ Form >> asMorph [
 ]
 
 { #category : #'*Morphic-Base' }
-Form >> floodFillTolerance [
-	^ self class floodFillTolerance
+Form class >> floodFillTolerance [
+	^ FloodFillTolerance ifNil: [FloodFillTolerance := 0.0]
 ]
 
 { #category : #'*Morphic-Base' }
-Form class >> floodFillTolerance [
-	^ FloodFillTolerance ifNil: [FloodFillTolerance := 0.0]
+Form >> floodFillTolerance [
+	^ self class floodFillTolerance
 ]
 
 { #category : #'*Morphic-Base' }
--- a/src/Morphic-Base/MouseClickState.class.st
+++ b/src/Morphic-Base/MouseClickState.class.st
@@ -120,12 +120,17 @@ MouseClickState >> handleEvent: evt from: aHand [
 				self click.
 				aHand resetClickState.
 				^true].
-			"Otherwise transfer to #firstClickUp"
-			firstClickUp := evt copy.
+
+			"Change the state to #firstClickUp.
+			Then queue the mouseUp event for later processing.
+			We will handle a click now."
 			clickState := #firstClickUp.
+			firstClickUp := evt copy.
 			"If timedOut or the client's not interested in dbl clicks get outta here"
-			aHand handleEvent: firstClickUp.
+			aHand queuePendingEvent: firstClickUp.
+			
 			self click.
+			
 			^false].
 		isDrag ifTrue:["drag start"
 			self doubleClickTimeout. "***"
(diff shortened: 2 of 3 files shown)


myroslavarm:
Hm is this also related to what Pavel was fixing regarding completion window not closing in https://github.com/pharo-project/pharo/pull/6017? There's so many UI bugs popping up :D

myroslavarm (approved):

guillep:
It seems there is a broken tests introduced:

```
testNoUnusedTemporaryVariablesLeft – Windows64.ReleaseTests.CleanCode.NoUnusedTemporaryVariablesLeftTest
```

The other ones seem randomish

```
testPackageEnvironment – MacOSX64.Refactoring.Tests.Environment.RBBrowserEnvironmentTest
testExpressionInIncompleteParanthesis – Windows64.NECompletion.Tests.CompletionContextTest
testAllExamples – Windows64.Hiedra.Tests.Model.HiExamplesTest
```

guillep:
Nope. This problem was introduced in build 178.
Pavel's problem exists already in build 172 (quick check).

Ducasse:
Broken tests are unrelated.

PR #6045, opened 2020-03-27
6040-TonelWriterwriteSnapshot-includes-tag-name-in-package-name-when-theres-no-OrganizationDefinition

carolahp:
Adding tags to MCClass definition.
Adding a MockClass whose category includes a tag

### DIFF
--- a/src/Monticello-Tests/MCSnapshotResourceWithTags.class.st
+++ b/src/Monticello-Tests/MCSnapshotResourceWithTags.class.st
@@ -1,0 +1,10 @@
+Class {
+	#name : #MCSnapshotResourceWithTags,
+	#superclass : #MCSnapshotResource,
+	#category : #'Monticello-Tests-Resources'
+}
+
+{ #category : #accessing }
+MCSnapshotResourceWithTags class >> mockPackageName [
+	^ 'MonticelloMocks-Tags'
+]
--- a/src/Monticello-Tests/MCTestCase.class.st
+++ b/src/Monticello-Tests/MCTestCase.class.st
@@ -155,6 +155,12 @@ MCTestCase >> mockSnapshot [
 ]
 
 { #category : #mocks }
+MCTestCase >> mockSnapshotWithTag [
+
+	^ MCSnapshotResourceWithTags current snapshot
+]
+
+{ #category : #mocks }
 MCTestCase >> mockToken: aSymbol [
 	^ MCMockDefinition token: aSymbol
 ]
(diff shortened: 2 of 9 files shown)


carolahp:
Failures are due to the introduction of a new mock class MCMockClassJ.
Tests should pass considering the PR in the Tonel repository: https://github.com/pharo-vcs/tonel/pull/85

tesonep:
Hi @carolahp , does it require to have a change in Tonel?

carolahp:
> Hi @carolahp , does it require to have a change in Tonel?

Hello @tesonep 
Yes, I have made a PR in the Tonel repo https://github.com/pharo-vcs/tonel/pull/85

tesonep:
I have merged the PR in tonel, I will release a new Iceberg and it will have this change, as soon I finish with the release I am doing.

tesonep:
There are some tests that are broken in the Monticello package. 
Can you test them?

carolahp:
@tesonep 
I checked the tests and found the cause.

Before my changes, no class in the MonticeloMocks package belonged to a tag, but I set the tag "Tag" in one class of the package and now the result of asking the snapshot all its categories includes the category "MonticeloMocks-Tag".

The failing tests are MonticelloGUI tests and they fail because the snapshot browser is showing separately each category (i.e. package name -  tag) instead of only the package name, as shown in the screenshot. 

Is this the desired behavior?

Screen Shot 2020-03-30 at 13 07 53

If it is then I would just need to modify the tests to expect the category "MonticeloMocks-Tag"

tesonep:
> @tesonep
> I checked the tests and found the cause.
> 
> Before my changes, no class in the MonticeloMocks package belonged to a tag, but I set the tag "Tag" in one class of the package and now the result of asking the snapshot all its categories includes the category "MonticeloMocks-Tag".
> 
> The failing tests are MonticelloGUI tests and they fail because the snapshot browser is showing separately each category (i.e. package name - tag) instead of only the package name, as shown in the screenshot.
> 
> Is this the desired behavior?
> 
> Screen Shot 2020-03-30 at 13 07 53
> 
> If it is then I would just need to modify the tests to expect the category "MonticeloMocks-Tag"

I don't know. But it looks like ok for me.

guillep:
I say we just make the tests green. Having two separate categories or as a single category is the same to me at this point. So maybe we need to change the tests.

carolahp:
> I say we just make the tests green. Having two separate categories or as a single category is the same to me at this point. So maybe we need to change the tests.

I just changed the tests here and also in tonel, I just did a PR there https://github.com/pharo-vcs/tonel/pull/88

guillep:
Seems there is no interest for this PR from the author. I'll close. Reopen if needed.

PR #6050, opened 2020-03-30
2146-Hardcode-class-rewriter-and-searcher-in-RB-class

### LINKED ISSUE #2146: RB classes should not hardcode class rewriter and searcher....
We should reduce the number of references to `RBParseTreeRewriter` and `RBParseTreeSearcher`

LaurineDargaud:
fix: #2146
From 62 to 23 references to RBParseTreeRewriter

### DIFF
--- a/src/AST-Core-Tests/RBParseTreeSearcherTest.class.st
+++ b/src/AST-Core-Tests/RBParseTreeSearcherTest.class.st
@@ -10,21 +10,21 @@ Class {
 	#category : #'AST-Core-Tests-Matching'
 }
 
+{ #category : #'tests - to de refined' }
+RBParseTreeSearcherTest >> getSearcher [
+	^ RBParseTreeSearcher new
+]
+
 { #category : #helpers }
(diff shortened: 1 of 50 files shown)


pavel-krivanek:
test failures seem to be related

Ducasse:
LaurineDargaud can you check a bit the tests that are broken.

Ducasse:
LaurineDargaud can you check a bit the tests that are broken.

Ducasse:
getSearcher is an ugly name the one before was much better. 
Can you use the previous one? parseTreeSearcher
get is not idiomatic of Pharo. It feels super strange.


In addition what you can do 
introduce a method called 
```
parseTreeRewriterClass
  ^ RBParseTreeRewriter 
```
and 
``` 
parseTreeRewriter
    ^ self parseTreeRewriterClass new
```

and use them in the hieararchy.

LaurineDargaud:
Alright, I'll open another PR with good names at top-level in classes hierarchy
I close this one

PR #6051, opened 2020-03-30
ReflectivityExamples: do not do compileTemporaryMethods in #initialize

MarcusDenker:
ReflectivityExamples removes and adds two methods in #initialize. This is not really needed as the tests that use it already treat it correctly.

The side effect was that executing the tests added these methods, leading to uncommited code.


MarcusDenker:
@hogoww I found the problem you mentioned last week (that the test package was dirty after running the tests)

tesonep:
Hi @MarcusDenker there tests for it? The existing tests validate these changes?

MarcusDenker:
This is in the test. Do we now write test for the tests?

tesonep:
Ok, cool I didn't get it. Sorry for the noise

PR #6054, opened 2020-03-30
Cleanup: namedTempAt: and namedTempAt:put:

MarcusDenker:
Next step in the methodMap cleanup: remove the method related to "namedTempAt:". 

the debugger methodmap has two kinds of offsets for temps: the real one (tempAt:) and a "virtual" offset for all real temps that are accessible by name.
This makes the whole thing quite difficult to understand and is better removed. There are no senders in the image, the methods where private to the debugger.

### DIFF
--- a/src/Debugging-Core/Context.extension.st
+++ b/src/Debugging-Core/Context.extension.st
@@ -42,22 +42,6 @@ Context >> methodReturnConstant: value [
 ]
 
 { #category : #'*Debugging-Core' }
-Context >> namedTempAt: index [
-	"Answer the value of the temp at index in the receiver's sequence of tempNames."
-	"NOTE: this list of temp names is completely virtual! It omits temp vectors but 
-	adds every variable that could be acccessed from a source perspective but might not be"
-	^self tempNamed: (self tempNames at: index)
-]
-
-{ #category : #'*Debugging-Core' }
-Context >> namedTempAt: index put: aValue [
-	"Set the value of the temp at index in the receiver's sequence of tempNames"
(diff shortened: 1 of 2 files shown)


MarcusDenker:
I think we can remove the methods on context, they were just there to implement tempNamedAt: and make no sense else.

myroslavarm (approved):
looks good to me

Ducasse:
I would love to understand this part of the system. Do we have a way to access the temps from the context without such methods?

Ducasse:
I mean the temps are not managed on context? I thought that they were.

guillep:
yes but not by name, just by index.

Le mar. 31 mars 2020 à 21:51, StéphaneDucasse <notifications@github.com> a
écrit :

MarcusDenker:
there is both namedTempAt: and tempAt: and tempNamed:

tempNamed: used to be implemented in terms of namedTempAt:, which is similar to  tempAt:, but it was operating on a purely virtual list of "temps that have a real name that you could access in this context in the debugger". Which means this list omits e.g. temp offsets that store the temp vector, but adds temps that are actually not stored in this context but an outer context.

This is all quite complex. 

By removing namedTempAt: (which we do not use anymore), things get at least a little bit simpler:  

- we have  tempAt: to access the real temp that is really there in this context
- we have  tempNamed: to to access temps by name which does allow to access even temps that are from an outer context.

MarcusDenker:
we found that it has users, will check with Esteban before merging.

MarcusDenker:
I close this an do a version where I just remove the already deprecated methods

PR #6057, opened 2020-03-31
Fix minimal Pharo on Window

pavel-krivanek:
The minimal Pharo is not able to do nor load anything useful on Windows becaues its start-up code is dependent of UnifiedFFI and fails immediatelly. 

Use low-level FFI call for WideCharToMultiByte


tesonep:
I have added some tests, because this was not tested at all before.
@pavel-krivanek we have only one way of the transformation using FFI kernel, it should not be done with the other side of the transformation?

PR #6073, opened 2020-03-31
Cleanup HashAndEqualsTestCase

### LINKED ISSUE #6072: Cleanup HashAndEqualsTestCase
- should implement #isAbstract to return true (as prototypes is initialized as empty collection running the tests makes no sense)
- modify comment to say it is abstract
- the method comments say something about "TextAttribute" which I guess is a leftover from refactoring functionality into this generic superclass for tests

Also: if you put a breakpoint or halt into  HashAndEqualsTestCase>>testEquality and run
the tests in a subclass like TextAlignmentTest you will see **that currently the tests are not inherited and therefore not even be run** (which makes no sense at all)

astares:
- fix #6072
-- provide implementation for #isAbstract to return true
-- this way now #testEquality and #testHash are really run on subclasses
-- adjust class comment
-- use #assert: instead of #should:
-- fix method comments

### DIFF
--- a/src/SUnit-Core/HashAndEqualsTestCase.class.st
+++ b/src/SUnit-Core/HashAndEqualsTestCase.class.st
@@ -1,5 +1,5 @@
 "
-I am a simple TestCase that tests for correct operation of #hash and #=.
+I am an abstract TestCase superclass that tests for correct operation of #hash and #=.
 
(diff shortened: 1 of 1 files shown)


guillep (approved):
Good! Test failures not related!

PR #6084, opened 2020-03-31
Improve UITestCase

### LINKED ISSUE #6083: Improve UITestCase
Class comment says "All instances of UITestCase are skipped in headless mode" which is not correct
because tests of subclasses are skipped.
So basically UITestCase is an abstract test superclass.

We should
1. implement a class side **#isAbstract** method 

```Smalltalk
isAbstract
     ^self == UITestCase 
```

2. Change the **class comment** to 

```
This is an abstract superclass for all tests requiring a user interface (UI). The tests are skipped when Pharo runs in headless mode.
```

3. Possibly even **rename the class** into AbstractUITestCase 

Step 1 and 2 should be done, step 3 is optional but would make it more clear

astares:
- rename the class UITestCase into AbstractUITestCase
- implement #isAbstract to return true
- tag the classes in SUnit-Support-UITesting and SUnit-Support-UITesting-Tests
- provide a better class comment
- provide a "deprecated" class UITestCase in Deprecated90 package for migration possibility (and to avoid loading problems)
- the deprecated class includes a class comment to now use AbstractUITestCase in the future

Fix #6083

### DIFF
--- a/src/Deprecated90/UITestCase.class.st
+++ b/src/Deprecated90/UITestCase.class.st
@@ -1,0 +1,8 @@
+"
+Please use AbstractUITestCase now
+"
+Class {
+	#name : #UITestCase,
+	#superclass : #AbstractUITestCase,
+	#category : #Deprecated90
+}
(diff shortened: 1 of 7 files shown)


guillep (approved):
Good!

Failure tests are fleaky tests. I'll open an issue for them separately...

Thanks @astares !

Issue #6058, opened 2020-03-31
RubShoutStyleDecorator DNU #click:for:controller


astares:
1.  Open latest image: Pharo 9.0.0 Build information: Pharo-9.0.0+build.202.sha.2799b9114f70ce3931612334575cd42509315882 (64 Bit)
2. in the Pharo **Welcome Screen** click on the link text called "regular network connection"

![image](https://user-images.githubusercontent.com/5980033/78006741-a9f7a580-733d-11ea-97a8-fcd7d7e9d579.png)

astares:
This is a side effect of https://github.com/pharo-project/pharo/pull/5992/files

The integrated changes and removal of displayCatalogProjectsInSpotter: which is used
in the welcome text left #fastNetworkDefaults and #slowNetworkDefaults in a broken state.
When clicked it leads to the given error.

The issue will automatically get fixed when 
https://github.com/pharo-project/pharo/issues/6059
is solved by integration of PR https://github.com/pharo-project/pharo/pull/6095

So we can close this one here already

<blockquote><img src="https://avatars0.githubusercontent.com/u/355455?s=400&v=4" width="48" align="right"><div><img src="https://github.githubassets.com/favicons/favicon.svg" height="14"> GitHub</div><div><strong><a href="https://github.com/pharo-project/pharo/pull/5992">Implementing a new iteration on top of spotter. by tesonep · Pull Request #5992 · pharo-project/pharo</a></strong></div><div>In Highlights it includes:

A set of processors implemented using generators and composable filters, so they don&#39;t need to traverse the whole image on a query.
Introduction of a query object re...</div></blockquote>

PR #6095, opened 2020-04-01
CatalogSettings displayCatalogProjectsInSpotter:

### LINKED ISSUE #6059: CatalogSettings displayCatalogProjectsInSpotter: missing
Looks like the recent pass on Spotter removed the ability to load
catalog entries from spotter.

Also the settings method #displayCatalogProjectsInSpotter was removed (instead of deprecated)
which breaks startup scripts that include:

```Smalltalk
CatalogSettings displayCatalogProjectsInSpotter: true
```

astares:
- use the new implementation of the settings to show catalog entries in spotter (fix broken code in #fastNetworkDefaults and #slowNetworkDefaults)
- provide old setting method #displayCatalogProjectsInSpotter: as a deprecated method in Deprecated90 package for compatibility / migration

Fix #6059

### DIFF
--- a/src/Deprecated90/CatalogSettings.extension.st
+++ b/src/Deprecated90/CatalogSettings.extension.st
@@ -1,0 +1,9 @@
+Extension { #name : #CatalogSettings }
+
+{ #category : #'*Deprecated90' }
+CatalogSettings class >> displayCatalogProjectsInSpotter: aBoolean [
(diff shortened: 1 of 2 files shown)


astares:
This PR will also solve another issue:   https://github.com/pharo-project/pharo/issues/6058

PR #6108, opened 2020-04-01
Endless loop in TickDialogWindow openModal

### LINKED ISSUE #6106: Endless loop in TickDialogWindow openModal
Scenario:

open the Settings browser, go to "Appearance" -> "IconSet" and click on the "Fetch from remote" button
Pharo will stop reacting and loop endless due to indirect recursive call because
TickDialogWindow is calling itself endlessly as seen from the stack presented by user interrupting:

image

astares:
Fix #6106


guillep (approved):
Good! Thanks!

PR #6114, opened 2020-04-01
Rework init logic in DamageRecorder

### LINKED ISSUE #6113: Rework init logic in DamageRecorder
The class DamageRecorder overwrites the new methods

```Smalltalk
new

	^ super new reset
```
and provides a reset like

```Smalltalk
reset
	"Clear the damage list."

	invalidRects := OrderedCollection new: 15.
	totalRepaint := false
```

This looks like a strange pattern. It would be better to have an instance side #initialize method.
Such an #initialize method is usually expected to see what and which initialization happens - this makes overwriting the "new" unnecessary.

astares:
- remove class side #new as #initialize is provided (which is called anyway
- removal of unncessary dots

Fix #6113

### DIFF
--- a/src/Morphic-Core/DamageRecorder.class.st
+++ b/src/Morphic-Core/DamageRecorder.class.st
@@ -11,19 +11,18 @@ Class {
 	#category : #'Morphic-Core-Support'
 }
 
-{ #category : #'instance creation' }
-DamageRecorder class >> new [
-
-	^ super new reset
-
-]
-
 { #category : #recording }
 DamageRecorder >> doFullRepaint [
 	"Record that a full redisplay is needed. No further damage rectangles will be recorded until after the next reset."
 
-	^ totalRepaint := true.
+	^ totalRepaint := true
 
+]
+
+{ #category : #initialization }
+DamageRecorder >> initialize [
+	super initialize.
+	self reset
 ]
 
 { #category : #recording }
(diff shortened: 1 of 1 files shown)


astares:
Unrelated failing of osx-64 / Tests-osx-64 / testExpressionInIncompleteParanthesis – MacOSX64.NECompletion.Tests.CompletionContextTest
```
Error
Improper store into indexable object
Stacktrace
Error
Improper store into indexable object
ByteString(Object)>>error:
ByteString(Object)>>errorImproperStore
ByteString>>at:put:
WriteStream>>nextPut:
[ :aStream |
aStream
nextPut: 'Expecting to find #sin 
```

Ducasse:
I was wondering how much time per second this method is used. because we could also remove the super initialize to have exactly the same behavior.

guillep:
None. There is a single instance per world. Created at bootstrap time. (from what I see in the code :P)

astares:
@guillep: a single instance is not correct, Basically we have three instances: DamageRecorder is instantiated in HandMorph and WorldState- We have a single hand, a world with a world state and as the MenuMorph (taskbar) instantiates one as well we have 3 instances afterall. Each with a damage recorder.

But I agree - we only have a few of them. Overhead is minimal.

**Side note:**
As we (usually can be sure) that we have a damage recorder in the world state I wonder why 
the nil test in #invalidate of WorldState is required:

```Smalltalk
invalidate
	self worldRenderer invalidate.
	damageRecorder 
		ifNil: [damageRecorder := DamageRecorder new]
		ifNotNil: [damageRecorder doFullRepaint]
```
maybe here a few cycles could be saved.

guillep (approved):

guillep:
Thanks!!

PR #6127, opened 2020-04-02
Skip Epicea fleaky test

### LINKED ISSUE #5938: Fix Hiedra Fleaky test

guillep:
Fix #5938.

testAllExamples fails in the CI randomly in different oss even with an increased 30 seconds time limit.


tinchodias:
Epicea no, Hiedra ;).
I'd blame some examples that process and render a random graph model, and then show the Athens surface into the column of a Spec table.

PR #6131, opened 2020-04-02
Introduce UnhandledException hierarchy

dionisiydk:
- Introduce UnhandledException as a root for ultimate exceptions like Halt and UnhandledError.

The ultimate meaning here is that such exceptions execute the UI action to interrupt the process execution as the result of not being handled. So they do not raise any other extra exception. And in interactive mode they trigger the UIManager to open a debugger.

Now with common hierarchy for Halt and UnhandledError we can simplify many places where we now write UnhandledError, Halt.

- raise an UnhandledError as default implementation for #defaultAction.

There are already several duplicated #defaultAction's in the hierarchy (removed here). It is better to have it as default implementation with following description: 
Even if not all exceptions are errors the absence of a handler for signalled exception is considered as an error by default. Therefore #defaultAction for Exception raises an error, an UnhandledError. 

- UnhandledError delegates its own default action to underlying exception using #unhandledErrorAction.

For example It allows to have specific debugger for Warning while it still processed through same UnhandledError machinery.

- Warning is processed through UnhandledError logic (see previous point)
- TestFailure is processed through UnhandledError logic

Last two changes fix Step Through in the debugger over "Warning signal" and "TestCase new assert: false" (or any assert in the test").

### DIFF
--- a/src/Flashback-Decompiler/CannotDecompileNativeBoostCalls.class.st
+++ b/src/Flashback-Decompiler/CannotDecompileNativeBoostCalls.class.st
@@ -6,8 +6,3 @@ Class {
 	#superclass : #Exception,
 	#category : #'Flashback-Decompiler-Exceptions'
 }
-
-{ #category : #accessing }
-CannotDecompileNativeBoostCalls >> defaultAction [
-	UnhandledError signalForException: self
-]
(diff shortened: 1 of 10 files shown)


Ducasse:
Nice idea!

dionisiydk:
Any suggestions for better names and comments are highly welcome

пт, 3 апр. 2020 г., 17:10 StéphaneDucasse <notifications@github.com>:

dionisiydk:
unrelated fleaky rubric test

pavel-krivanek (approved):

PR #6133, opened 2020-04-03
IRBuilder: #addTemps: simplify

MarcusDenker:
simplify #addTemps by using #withIndexDo:


MarcusDenker:
failing tests do not look related to this change

MarcusDenker:
yes, it is related. I will have a look. closing for now

MarcusDenker:
dumb me: it skips of course incrementing the index if the key is already there... so withIndexDo: can not be used here.

PR #6134, opened 2020-04-03
Cleanup: #vectorOffset

MarcusDenker:
Small cleanup in preparation for the offset caching PR to be done later
- remove unused accessor #vectorOffset and the one test that called it


guillep (approved):
Good! no other senders!

PR #6142, opened 2020-04-06
do-not-generate-initialize-method-for-slots-if-it-already-call-super

jecisc:
Do not generate useless initialize methods via slot usage.

When we use a slot, we ensure the #initialize method call #initializeSlots. But this mecanism can create useless initialize methods in case the superclass already call #initializeSlot and the subclass has nothing more to initialize. 
This change the position of #ensureInitializeMethodExists to not generate useless code.

+ Update tests

### DIFF
--- a/src/Slot-Core/Slot.class.st
+++ b/src/Slot-Core/Slot.class.st
@@ -116,23 +116,6 @@ Slot >> = other [
 			and: [ name = other name ]
 ]
 
-{ #category : #private }
-Slot >> addSlotInitToInitialize: aClass [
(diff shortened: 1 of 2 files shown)


MarcusDenker:
failing tests are related

jecisc:
This is weird because those tests are green in my image.

jecisc:
Updated. Let's see what the CI says.

MarcusDenker:
failing test is not related

MarcusDenker (approved):
looks good

PR #6143, opened 2020-04-06
use-ifExists-in-more-places

jecisc:
Use #ifExists:(ifAbsent:) at more places.


jecisc:
Apparently the problem I got in my previous PR comes from this subset of changes.

jecisc:
I'll remove some candidates for the problem and see what the CI says.

jecisc:
Apparently, the problem happens when we update UITheme. I wonder why.

Anyway, the tests passed here.

guillep (approved):
Ok! Thanks!

PR #6154, opened 2020-04-08
backportIfErrorDo

pavel-krivanek:
backport ifErrorDo: from Pharo 9.0 to Pharo 8.0 to allow us to write code working in both versions


astares:
Why "if" instead of  #onErrorDo: ?  It dispatches to #on:do: so we should keep the "on"

pavel-krivanek:
see https://github.com/pharo-project/pharo/pull/5471
Here is not the right place to discuss this change.

guillep:
Let's wait for the resolution of #6158 before. Like that we do not have to do two backports :)

guillep:
This PR is dating a bit. I'll close. Reopen if needed.

PR #6158, opened 2020-04-08
ifErrorDo: should be onErrorDo: and deprecate with transform

### LINKED ISSUE #6157: ifErrorDo: should be onErrorDo: and deprecate with transform
The method #ifError: was deprecated and #ifErrorDo: should be used now

The problem:
  1. #ifErrorDo: is not a good name - it should have been called #onErrorDo:
      as it dispatches  #on: Error do: aBlock

   2. the deprecated method #ifError: have not been moved to deprecated package

   3. The depreaction was not done with a transformation rule

   4. there were no tests added

astares:
1. Rename #ifErrorDo: into #onErrorDo:
2. use #onErrorDo: in deprecated methods like #ifError: and  #ifErrorDo:
     and move these to Deprecated90 package
3. have a transformation rule for the deprecations in  #ifError: and  #ifErrorDo:
4. Add a test #testOnErrorDo

Fix #6157


src/Deprecated90/BlockClosure.extension.st
```diff
@@ -0,0 +1,15 @@
+Extension { #name : #BlockClosure }
+
+{ #category : #'*Deprecated90' }
+BlockClosure >> ifError: errorHandlerBlock [
+
+	self deprecated: 'Use #onErrorDo: instead.' transformWith: '`@receiver ifError: `@arg' -> '`@receiver onErrorDo: `@arg'.
```
dionisiydk: the deprecation should more tricky here which is probably the reason why rewrite was not implemented in first PR.
Originally ifError could accept two arguments (using cull). I do not remember exactly. Check Pharo 8.

I think for simplicity we can do rewrite only for zero argument block. Otherwise use deprecation message.

dionisiydk (changes requested):

astares:
@dionisiydk Can you please check again? Thanks in advance!

src/Deprecated90/BlockClosure.extension.st
```diff
@@ -0,0 +1,21 @@
+Extension { #name : #BlockClosure }
+
+{ #category : #'*Deprecated90' }
+BlockClosure >> ifError: errorHandlerBlock [
+
+	errorHandlerBlock numArgs = 0 
+		ifTrue: [ self deprecated: 'Use #onErrorDo: instead.' transformWith: '`@receiver ifError: `@arg' -> '`@receiver onErrorDo: `@arg' ]
+		ifFalse: [ self deprecated: 'Use #onErrorDo: instead.' ].
+		
+	^ self onErrorDo: errorHandlerBlock
```
dionisiydk: You need to wrap it like:
```Smalltalk
self onErrorDo: [:err | errorHandlerBlock cull: err receiver cull: err description]
```
But I did not check in code what exact #cull:cull: it should be to be compatible.

dionisiydk (changes requested):

src/Deprecated90/BlockClosure.extension.st
astares: Why wrap? We just dispatch further on. Do you have a problematic case?

dionisiydk: That is original method:
```Smalltalk
ifError: errorHandlerBlock
   ^ self on: Error do: [:ex |
		errorHandlerBlock cull: ex description cull: ex receiver]
```
The issue is about eliminating this strange error unboxing so that the handler block will accept a normal error instance.

dionisiydk: what I am trying to say is that deprecated method should continue work with this strange two parameters

guillep: Yes, the deprecated method should continue working as before :)

guillep:
ping?

astares:
@guillep pong!

guillep:
I've made the deprecated methods work as before to keep compatibility. Let's wait for the CI

guillep (approved):
The overall changes look ok. Let's wait for the CI.

dionisiydk (approved):

PR #6162, opened 2020-04-08
Fixing the installation of a different world.

tesonep:
This method is for backward compatibility and should not be used as it will be replaced.
New tools should not use it, and old should be migrated.
Once, we have the complete change we will provide a deprecation and a replacement.
This is to keep working the tools that are using it, as now there is no replacement yet.


astares (approved):

Ducasse:
I integrate it.

astares:
Thanks!

PR #6167, opened 2020-04-09
Cleanup: #lookUpForDeclaration: in the Compiler

MarcusDenker:
#lookupVarForDeclaration: was was duplicating the lookupVar: logic just to skip the OCRequestorScope lookup when in interactive use.

Instead of re-implemeting the lookup, the new version uses a dynamic variable LookupForDeclaration that is set to true in #lookupVarForDeclaration: and then checked in OCRequestorScope>>#lookupVar:

This way we can reuse the implementation of #lookupVar: instead of copying it.

### DIFF
--- a/src/OpalCompiler-Core/OCRequestorScope.class.st
+++ b/src/OpalCompiler-Core/OCRequestorScope.class.st
@@ -39,6 +39,9 @@ OCRequestorScope >> findVariable: aBlock ifNone: aNotFound [
 { #category : #lookup }
 OCRequestorScope >> lookupVar: name [
 
+	"do not create bindings if we lookup for declaration"
+	LookupForDeclaration value ifTrue: [ ^ outerScope lookupVar: name ].
+
 	name = 'self' ifTrue: [ ^ outerScope lookupVar: name ].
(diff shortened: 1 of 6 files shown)


src/OpalCompiler-Core/OCRequestorScope.class.st
```diff
@@ -39,6 +39,9 @@ OCRequestorScope >> findVariable: aBlock ifNone: aNotFound [
 { #category : #lookup }
 OCRequestorScope >> lookupVar: name [
 
+	"do not create bindings if we lookup for declaration"
+	LookupForDeclaration value ifTrue: [ ^ outerScope lookupVar: name ].
```
guillep, line 43: I don't understand how this avoids to create bindings. This is not clear in the code...

guillep (changes requested):
Is this change really good? I think using these kind of dynamic variables for this purpose is at least tricky. It makes the code more difficult to read and understand... It generates a dynamic dependency between two pieces of code that look seemingly unrelated...

And also it complicates debugging (because it depends on the dynamic scope that the debugger does not properly inherit...)

MarcusDenker:
I will keep the current solution. The dynamic variable is too much magic.

PR #6168, opened 2020-04-09
CompilerCleanup-hasIncompleteIdentifier

MarcusDenker:
This PR documents #hasIncompleteIdentifier and does some small fixes for #findVariable:ifNone: 

- add a default method to OCAbstract scope, removes two methods that were implemetend the same in the subclasses
- make sure to evaluate the ifNone block if the outerScope is nil
- comments

I am tempted to get rid of findVariable: ifNone: and instead:
 -  implement #hasBindingThatbeginsWith: on the scope hierarchy. 
 - use the existing (fast) hasBindingThatBeginsWith: when checking Class and global scope

But as a first step, this PR is good to integrate


MarcusDenker:
I will update this PR to use #hasBindingThatbeginsWith:

guillep (approved):
I really wonder if this has no impact in the tools at all... Specially seeing that we remove `hasBindingThatBeginsWith:` in all tools. Was that used for something in particular.

This `hasIncompleteIdentifier` seems only used by the syntax highlighter, no?

I wonder the usefulness of this. This requires in the worst case a full scan of the system to check if a symbol is **almost** defined and put it in *italics*...

I'd even go further and remove entirely the support. This would simplify many things for an almost useless features that creates problems in large images...

guillep:
I think the change is ok by now, but maybe removing this does pay in the long term.

Ducasse:
Do you have an idea how we could get the italic because to me this is super cool to know that what I'm typing exist but it not complete

guillep:
> Do you have an idea how we could get the italic because to me this is super cool to know that what I'm typing exist but it not complete

Maybe with the indexes Pablo and Esteban wrote.

In general I don't think people understand italics means *incomplete*. They may understand that red is **wrong**, because that's commonly used everywhere. but I don't think normal people see a difference between red and italics+red :).

The question is: is the cost of having search indexes worth the difference between *red* and *italics+red*?

Ducasse:
To me this is super cool because this is a super nice information I use it all the time. 
Now I do not get why this is red + italic because in my book for kids it was not like that

Red=Error
blue + italics is much better. 

It is boring when things get broken :(

PR #6178, opened 2020-04-14
Fix 6160 Cleanup two tests in BlockClosureTest

### LINKED ISSUE #6160: Cleanup two tests in BlockClosureTest
the tests
- testSupplyAnswerOfFillInTheBlank
- testSupplyAnswerOfFillInTheBlankUsingDefaultAnswer

should:
 - not use assertions instead of #2192  
 - be in category 'tests' not 'testing'

VincentBlondeau:
Fix #6160


VincentBlondeau:
Indeed, you are not a very intelligent bot :)

astares (approved):
OK

Ducasse:
Thanks vincent. 
What would be nice is it package the supply method with the UImanager or in a separate package. I super tired but I found super strange to have a dependency to the UIManager from the BlockClosure tests. We can do that in another issue.

PR #6179, opened 2020-04-14
Fix Warning defaultAction

dionisiydk:
I lost Warning>>#defaultAction in #6131  . Here is the fix

### DIFF
--- a/src/Kernel-Tests/ExceptionTest.class.st
+++ b/src/Kernel-Tests/ExceptionTest.class.st
@@ -38,6 +38,20 @@ ExceptionTest >> runCaseManaged [
 	^DefaultExecutionEnvironment beActiveDuring: [ self runCase]
 ]
 
+{ #category : #private }
+ExceptionTest >> runWithNoHandlers: aBlock [
+	"Executing the given block directly would go through all handlers of SUnit machinery.
+	Here we simulate the clean environment with no outer handlers for possible block errors"
+	| synchSemaphore result |
+	synchSemaphore := Semaphore new. 
+	[result := aBlock ensure: [synchSemaphore signal]] forkNamed: 'Test process'.
+	"we should not wait forever in case of wrong behaviour 
+	(general test timeout is disabled in #runCaseManaged)"
+	(synchSemaphore wait: 2 seconds)
+		ifTrue: [ self error: 'Timeout for block execution' ].
+	^result
+]
+
 { #category : #'tests - exceptiontester' }
 ExceptionTest >> testDoubleOuterPass [
 	self assertSuccess: (ExceptionTester new runTest: #doubleOuterPassTest ) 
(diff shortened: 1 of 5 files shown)


Ducasse:
Hi denis 
could you add some tests?
S.

dionisiydk:
> Hi denis
> could you add some tests?

yes, they were in progress. And now committed

guillep:
Sorry for coming late, I understand the issue and I think this PR is good to integrate for compatibility with #6131 . However, I wonder what is the impact of moving `defaultAction` to `unhandledErrorAction`, especially for libraries subclassing exceptions and redefining `defaultAction`.

Is (was) `defaultAction` meant to be executed when the error was unhandled only? Then maybe we can simplify this even further and turn back `unhandledErrorAction` into `defaultAction` (keeping compatibility) and change the caller. Something like:

```
UndefinedObject >> handleSignal: exception [
	"When no more handler (on:do:) context left in sender chain this gets called.  Return from signal with default action."

	^ exception resumeUnchecked: exception unhandledErrorAction
]

Exception >> unhandledErrorAction [
     ^ self raiseUnhandledError
]

UnhandledError >> unhandledErrorAction [
    ^ innerException defaultAction
]
```

@dionisiydk  What do you think? Something like this would keep `defaultAction` as the main extension point (just as before!) and make all exceptions throw unhandled errors...

guillep (approved):
Meanwhile, this is ok.

dionisiydk:
Hi @guillep 

>  However, I wonder what is the impact of moving `defaultAction` to `unhandledErrorAction`, especially for libraries subclassing exceptions and redefining `defaultAction`.

My change was pure refactoring (except that I lost Warning part which is fixed here). So it should be no impact on any code.
Considering what you are asking there were two goals:
- avoid code duplication of UnhandledError signalling and "debugger fallback" in exception hierarchy
- allow specific "debugger fallback" for specific exceptions (Error vs Warning)

1. The first is solved by two parts:

- extract UnhandledError signalling into reusable method #raiseUnhandledError and calling it as default Exception logic for #defaultAction. See my method comment that by default we consider that no handler is an error condition.

- make existing exceptions with "debugger fallback" to use #raiseUnhandledError (explicitly or relying on introduced inherited logic)
 
Notice that #defaultAction itself is not about error from the point of view of signal processing (#handleSignal:). For example Notifications (DynamicVariables) do some logic "silently" and resume a signaller.  But as default implementation in Exception we now process it as an error. 
(Previously it was #subclassResponsibility. And subclasses with missing implementation were in fact broken)

2.  The second part is done using extra delegation in UnhandledError>>#defaultAction into underlying exception:
```Smalltalk
defaultAction
	^ exception unhandledErrorAction
```
Here I was needed to introduce this new method #unhandledErrorAction. It only participates in processing of UnhandledError, what to do with concrete exception in "ultimate scenario" when completely nothing helps to find any handler.

dionisiydk:
> Is (was) `defaultAction` meant to be executed when the error was unhandled only? 

yes

> Then maybe we can simplify this even further and turn back `unhandledErrorAction` into `defaultAction` (keeping compatibility) and change the caller. Something like:
> 
> ```
> UndefinedObject >> handleSignal: exception [
> 	"When no more handler (on:do:) context left in sender chain this gets called.  Return from signal with default action."
> 
> 	^ exception resumeUnchecked: exception unhandledErrorAction
> ]
> 
> Exception >> unhandledErrorAction [
>      ^ self raiseUnhandledError
> ]
> 
> UnhandledError >> unhandledErrorAction [
>     ^ innerException defaultAction
> ]
> ```
> 
> @dionisiydk What do you think? Something like this would keep `defaultAction` as the main extension point (just as before!) and make all exceptions throw unhandled errors...

I do not think it is a good idea because UnhandledError is a tool to catch "really" unhandled errors (where we can be sure that there will be a "debugger fallback" at the end).
If all exceptions would be processed through UnhandledError machinery we would not be able to distinguish notifications-like signals which never "break execution". 

The main use case for this tool is StepOver and StepThrough commands in the debugger. 
UnhandledError handler is injected into current context to stop execution when another "debugger fallback" is detected. And the causing error is shown. When we step over notification the debugger should not break at signaler.

Our current and old issues with debugger are related to broken UnhandledError logic (#5470) and  new sources of fallback (Halt, Warning, ..) created in past. Here (#6131) I introduced UnhandledException as superclass for such special exceptions and I removed redundant cases like Warning and TestFailure (I think Halt deserves to be special for a good reason).

PR #6185, opened 2020-04-15
Try fix #6021 

### LINKED ISSUE #6021: TaskBar is not appearing by default
**Describe the bug**
The TaskBar is not appearing

**To Reproduce**
1. Open a new Pharo9 Image
2. The bar is not there

**Expected behavior**
To have the taskBar at the bottom of the world.

**Screenshots**
Expected:

image

Actual:

image

**Version information (please complete the following information, you can get it from Help->about):**
 - OS: any
 - Version: Pharo 9.0

**Additional context**
If you do ```TaskbarMorph reset``` it appears back.
Looks like an initialization problem. Maybe the initialization of the task bar during the bootstrap is not in the right place

guillep:
Explicitly set a taskbar to the world at initialization time, instead of depending on the chaotic order of class side `initialize` methods and their global dependencies.

This is cleaner than having global dependencies.

Let's wait until the CI finishes to see if the resulting image has the taskbar.

### DIFF
--- a/src/BaselineOfIDE/BaselineOfIDE.class.st
+++ b/src/BaselineOfIDE/BaselineOfIDE.class.st
@@ -49,7 +49,6 @@ BaselineOfIDE >> additionalInitialization [
 
 	PolymorphSystemSettings desktopColor:  Color veryVeryLightGray lighter.
 	SourceCodeFonts setSourceCodeFonts: 10.
-	LogicalFontManager unload.
 	FreeTypeSystemSettings loadFt2Library: true.
 	FreeTypeSettings current monitorType: #LCD.
 	FreeTypeSettings current glyphContrast: 55.
@@ -85,9 +84,7 @@ BaselineOfIDE >> additionalInitialization [
 	SDL_Event initialize.
 	
 	HiRulerBuilderTest initialize.
-	
-	LogicalFontManager current addFontProvider: FreeTypeFontProvider current.
-	
+		
 	3 timesRepeat: [
 		Smalltalk garbageCollect.
 		Undeclared removeUnreferencedKeys.].
--- a/src/BaselineOfMorphic/BaselineOfMorphic.class.st
+++ b/src/BaselineOfMorphic/BaselineOfMorphic.class.st
@@ -317,8 +317,6 @@ BaselineOfMorphic >> postload: loader package: packageSpec [
 	ZnConstants ZnMimeType ZnNetworkingUtils ZnServer ZnSingleThreadedServer
 	) do: [:each | (Smalltalk globals at: each) initialize. ].
 
-	LogicalFontManager current addFontProvider: FreeTypeFontProvider current.
-
 	BalloonMorph setBalloonColorTo: Color yellow.
 
 	UIManager default terminateUIProcess.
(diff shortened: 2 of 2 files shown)


guillep:
Turns out that taskbar support for the world is all implemented as extension methods.
The world belongs to `BaselineOfMorphicCore`. The taskbar belongs to `BaselineOfMorphic` which is loaded **after**. So we cannot initialize the taskbar at the initialization time of the world because the taskbar is not available.

So, second attempt: explicitly initialize the taskbar upon loading of `BaselineOfMorphicCore`

guillep:
Second attempt does not work either...

pavel-krivanek:
It it initialized properly, I guess, so it makes sense that this fix does not work. It is strange that it started to fail because of such fonts change. Isn't the problem in initialization of the RealEstateAgent or so? Because the position of progress bars is broken too...

guillep:
> It it initialized properly, I guess, so it makes sense that this fix does not work. It is strange that it started to fail because of such fonts change. Isn't the problem in initialization of the RealEstateAgent or so? Because the position of progress bars is broken too...

Maybe it's a lot of side effects that add up... I'm trying to bootstrap locally and logging to see.

guillep:
Ok, my last try seems to do it :).
Removing both the unload and the second initialization seem to do the trick. Still I'm not sure why. What I know is that the font code is a labyrinth.

Also, Should we add a release test to check there is a taskbar?

Ducasse:
About the font code: been there and yes it is still a dark place.

Ducasse:
The idea of the test may be good.

pavel-krivanek:
What about reset (`TaskbarMorph reset`) after the font manager reinitialization?

guillep:
> What about reset (`TaskbarMorph reset`) after the font manager reinitialization?

I wonder why the reinitialization (the send to `#unload`) is required at all.

If we look at the CI output of this PR, we also see

```
TestFailure
Assertion failed
LogicalFontManagerTest(TestAsserter)>>assert:description:resumable:
LogicalFontManagerTest(TestAsserter)>>assert:description:
LogicalFontManagerTest(Object)>>assert:
LogicalFontManagerTest>>testHasCurrentFreeTypeFontProvider
```

This means the `LogicalFontManager` and the `FreeTypeFontProvider` are not in sync. This is probably related to the code I changed. But I wonder why, because I did not force the creation of any new `FreeTypeFontProvider`. Looking a bit further shows this code in `BaselineOfMorphic`:

. ].

```smalltalk
	LogicalFontManager current addFontProvider: FreeTypeFontProvider current.

...

	#(#AbstractFont #BalloonMorph #CP1253TextConverter #Clipboard #EditorFindReplaceDialogWindow #EmbeddedFreeTypeFontInstaller #FLCompiledMethodCluster #FLLargeIdentityHashedCollection #FLPlatform

#FreeTypeFontProvider

#GB2312 #GradientFillStyle #GreekEnvironment #ISO88597TextConverter #IconicButton #JISX0208 #JapaneseEnvironment #KOI8RTextConverter #KMLog #KSX1001 #KoreanEnvironment #LogicalFont #LucidaGrandeRegular #OSEnvironment #RussianEnvironment #ScrollBar #SimplifiedChineseEnvironment "#SourceCodeProRegular #SourceSansProRegular" #StrikeFont #SystemSettingsPersistence #TabMorph #TaskbarMorph #Text #TextAction #TextConstants #TextEditor #TextStyle) do: [:each |
		 (Smalltalk globals at: each) initialize. ].
```

so we are initializing the `FreeTypeFontProvider` class AFTER having initialized the `LogicalFontManager` with it. `FreeTypeFontProvider` class side initialize of course does this:

``` smalltalk
FreeTypeFontProvider >> initialize [
initialize
	"
	self initialize
	"
	"ensure that other classes have also been initialized by forcefully initializing them now.
	It then does not matter which order they are initialized in during the package load"

...	
	current := nil.
	self current. "this creates an instance of me, and updates from the system"
]
```

So it recreates the instance...

pavel-krivanek:
there is one adding of a font provider in `BaselineOfIDE>>#additionalInitialization`. Maybe that's the reason of non-synced state

guillep:
> there is one adding of a font provider in `BaselineOfIDE>>#additionalInitialization`. Maybe that's the reason of non-synced state

Ah I just realized that you may have not read the latest code in this PR. I removed that extra thing, and that seems to work. Do you mind reviewing?

pavel-krivanek:
ah, ok. Now I slightly rember that I did the mess with `unload` etc. because of cyclic dependencies in the intialization of these classes...

pavel-krivanek (approved):
works well now, thanks!

PR #6195, opened 2020-04-18
[review needed] Repair arg renaming refactoring

hogoww:
The first commit shows the problem.
I'm offering some renaming the next three, and differentiation of the #isTempNodeSelected.
I'm arguing that we're reasoning from an AST point of view, and that nodes should be differentiated.
(We had the same problem with the compiler, making #isTemp "complicated" to modify correctly).

Also, 3 refactoring depend on #isTempNodeSelected.
1 of them is actually using #isTempOrArgNodeSelected.
2 of them actually want the right semantic that exists for #isTempNodeSelected.
(unless I'm mistaken)

On the testing part, I'm not sure of what to test.
I'd tend to test SycRenameArgOrTempCommand >> canBeExecutedInContext:
But this requires the creation of a SourceCodeContext which I'm not too sure about.
Also this is required for every refactor commands.
@guillep What do you think?

I'll complete this PR later

### DIFF
--- a/src/Refactoring-Core/RBRenameArgumentOrTemporaryRefactoring.class.st
+++ b/src/Refactoring-Core/RBRenameArgumentOrTemporaryRefactoring.class.st
@@ -7,7 +7,7 @@ The variable declaration an all references in this method are renamed.
 My precondition verify that the new name is a valid variable name and not an existing instance or a class variable name
 "
 Class {
-	#name : #RBRenameTemporaryRefactoring,
+	#name : #RBRenameArgumentOrTemporaryRefactoring,
 	#superclass : #RBMethodRefactoring,
 	#instVars : [
 		'selector',
@@ -20,7 +20,7 @@ Class {
 }
 
 { #category : #'instance creation' }
-RBRenameTemporaryRefactoring class >> model: aRBSmalltalk renameTemporaryFrom: anInterval to: newName in: aClass selector: aSelector [ 
+RBRenameArgumentOrTemporaryRefactoring class >> model: aRBSmalltalk renameTemporaryFrom: anInterval to: newName in: aClass selector: aSelector [ 
 	^(self new)
 		model: aRBSmalltalk;
 		class: aClass
@@ -31,7 +31,7 @@ RBRenameTemporaryRefactoring class >> model: aRBSmalltalk renameTemporaryFrom: a
 ]
 
 { #category : #'instance creation' }
-RBRenameTemporaryRefactoring class >> renameTemporaryFrom: anInterval to: newName in: aClass selector: aSelector [ 
+RBRenameArgumentOrTemporaryRefactoring class >> renameTemporaryFrom: anInterval to: newName in: aClass selector: aSelector [ 
 	^self new
 		class: aClass
 		selector: aSelector
@@ -40,7 +40,7 @@ RBRenameTemporaryRefactoring class >> renameTemporaryFrom: anInterval to: newNam
 ]
 
 { #category : #initialization }
-RBRenameTemporaryRefactoring >> class: aClass selector: aSelector interval: anInterval newName: aString [ 
+RBRenameArgumentOrTemporaryRefactoring >> class: aClass selector: aSelector interval: anInterval newName: aString [ 
 	class := self classObjectFor: aClass.
 	selector := aSelector.
 	interval := anInterval.
@@ -48,7 +48,7 @@ RBRenameTemporaryRefactoring >> class: aClass selector: aSelector interval: anIn
 ]
 
 { #category : #preconditions }
-RBRenameTemporaryRefactoring >> preconditions [
+RBRenameArgumentOrTemporaryRefactoring >> preconditions [
 	^ (RBCondition
 		withBlock: [ | methodSource |
 			interval first > interval last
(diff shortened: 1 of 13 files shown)


src/Refactoring-Core/RBRenameArgumentOrTemporaryRefactoring.class.st
```diff
@@ -7,7 +7,7 @@ The variable declaration an all references in this method are renamed.
 My precondition verify that the new name is a valid variable name and not an existing instance or a class variable name
 "
 Class {
-	#name : #RBRenameTemporaryRefactoring,
+	#name : #RBRenameArgumentOrTemporaryRefactoring,
```
dionisiydk, line 10: I still feel that this new distinguish between temp and arg is wrong. And I believe this complex names prove that. 
What is wrong and where it is not convenient to have just temps variables? 
Arguments are also temp vars. But not all temp vars are arguments. Why this sentence is bad?

hogoww: I'm gonna try to make an argument, we'll see if that makes sense :p
Also, this argument is to the best of my knowledge, so if you find holes in them, please  help me fill those gaps =)

For me, a lot of tools are working on the AST (such as this one).
Therefore, when I interact with a tool, I have the AST's semantics in mind;
The AST differentiate temporaries and arguments (which I personally like).

In the compiler, you have a compiler view, which does not differentiate temporaries and args.
When interacting with the AST, it changes view, it should ask #isArgOrTemp to have the same behavior as on its nodes for #isTemp.
When you know one view, and someone else's comes into play without warning, what do you do? 
You make semantics errors.

As I said, in the context of the sourceCodeRefactor, you have currently 3 users of #isTempNodeSelected.
"1 of them is actually using #isTempOrArgNodeSelected.
2 of them actually want the right semantic that exists for #isTempNodeSelected."
What I meant by that is that semantically, there is 1 that is appliable on both temps and args, and two only on temps.
Meaning that without differentiation, there is 2 extra refactors offered for argument nodes that cannot be used.
I think it's a semantic mistake. 

>Arguments are also temp vars. But not all temp vars are arguments. Why this sentence is bad?
(I think that) if they had a common super class, that showed that, I would agree with this.
And I don't mean VariableNodes because instanceVariables are definitely not temps.

Maybe a simple differentiation between something like #isTemp on one side, and #isTemporaryNode/#isArgumentNode on the other would be enough ? 
I think it's error prone, but this would at least separate the different concepts, the different views.

Let me know what you think !

Ducasse:
@dionisiydk you do not want to have the same refactoring on args and temp. 
tmp could be move to smaller scope. 
For arg you want to rename, remove. 
So having 
isTemp
isArg
and isTempOrArg can make sense.
No?

dionisiydk:
My complain was not about the PR particularly. This is the right change. I agree that we need to know temp vars which are not arguments. 
But I really do not like naming "tempOrArg" everywhere. And this idea that on compiler OC level it will have different meaning is bad (notice that OCArgumentVariable is subclass of OCTempVariable).

Could we have something better? I know it is easy to complain and not propose anything. But do you really ok with this state of "variables model"?

Now, writing this I got an idea for different names. What you think to name all kind of temps a "method variables"? Temps and arguments will be separate subclasses. 
So we will have #isMethodVariable and "isTemp and isArg" to have extra distinguish like now.
MethodVariable will follow its MethodScope name which I think will be really consistent to other vars: GlobalVar -> global scope, InstanceVar -> instance/class scope.

And for this PR we will have simple nice names: RBRenameMethodVariable, isMethodVariableSelected and so on.

guillep:
Why not calling them local variables?

where isLocal = isArgument or isTemporary ?

hogoww:
Sorry @dionisiydk I was way off :').
I like it, i'll do it in a separate PR though, and I'll try to do it progressively to not break the compiler.

dionisiydk:
> Why not calling them local variables?
> 
> where isLocal = isArgument or isTemporary ?

Let's decide what is better: #isLocalVariable or isMethodVariable?

Personally I prefer to not use #local name because it conflict with remote scenarios. For example in Seamless there are remote variables.

dionisiydk:
> Sorry @dionisiydk I was way off :').
> I like it, i'll do it in a separate PR though, and I'll try to do it progressively to not break the compiler.

good.

guillep:
Well, "local variable" is a kind of well used terminology to talk about variables local to a function/method... So if somebody has to win, I'd vote to give the name to the most common and practical scenario :)

Ducasse:
I like is isLocalVariable
Denis 
local variable is a well know concept so may be you should call your variables remoteVariable.

dionisiydk:
> Well, "local variable" is a kind of well used terminology to talk about variables local to a function/method... So if somebody has to win, I'd vote to give the name to the most common and practical scenario :)

Ok. I agree. This is better.

dionisiydk:
> I like is isLocalVariable
> Denis
> local variable is a well know concept so may be you should call your variables remoteVariable.

Yes, they are named as remote variables. In case of "distributed method" it will include remote and local variables where local would mean network location, not a semantical scope.
It is of course very special thing and we can ignore it here. But after Seamless work I always have such doubts about local word when it is not about the network. #localMethods for example has same issue if you would start think about distributed environment.

hogoww:
I added tests, this should be okay to use. 
The isLocal part has to be added in a different PR.
Let me know if that's okay

PR #6197, opened 2020-04-19
Fix Process>#isSuspended

dionisiydk:
#isSuspended method returns incorrect process state. For example:
```Smalltalk
[state := Processor activeProcess isSuspeded] fork.
```
returns true in state variable.
PR describes problem cases with tests and fixes them with Cuis version.

### DIFF
--- a/src/Kernel-Tests-Extended/ProcessTest.class.st
+++ b/src/Kernel-Tests-Extended/ProcessTest.class.st
@@ -183,8 +183,48 @@ ProcessTest >> testInjectingMultipleExceptionHandlersIntoRunningProcess [
 ]
 
 { #category : #tests }
+ProcessTest >> testIsNotSuspendedWhenItIsActiveProcess [
+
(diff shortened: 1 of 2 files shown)


src/Kernel-Tests-Extended/ProcessTest.class.st
```diff
@@ -233,7 +273,7 @@ ProcessTest >> testIsTerminatingForcedTermination [
 	self assert: terminator identicalTo: process ].
 
 	[ process isTerminated ] whileFalse: [ 50 milliSeconds asDelay wait ].
-	self assert: process isSuspended.
+	self deny: process isSuspended.
```
dionisiydk, line 276: Notice that terminated process is not suspended.

Ducasse:
Ohhhhhh excellent. Thanks a lot denis.

guillep (approved):
Yes!

PR #6198, opened 2020-04-21
IRBuilder-fill-TempOffsets

MarcusDenker:
The OCTempVariables used to cache the index after the first access. The index itself was taken from the IR.

We can do it differently: every time a temp is created on the IR level, we can backfill the offset in the OCTempVariables.

This PR:

- IRBuilder now fills the index of semantic variables on #addTemp:
- This allows to not search for the offsets in the IR when accesing temps reflectively (#readFromLocalContext: and similar)

Fixes:
- some improvements to #lookupVar: (so we can call it with non-string var names)
- map the node correctly when generating inlined blocks and full blocks

Cleanups:
- remove #tempVectorNamed: #indexFromIRForVarNamed: #indexFromIR and #indexInTempVectorFromIR:

Accessing the index of OCTempVariables continues to do a nil check. This allows us to call it even if the IR has not yet
been generated.

### DIFF
--- a/src/OpalCompiler-Core/IRBuilder.class.st
+++ b/src/OpalCompiler-Core/IRBuilder.class.st
@@ -97,9 +97,10 @@ IRBuilder >> addTemp: tempKey [
 	tempMap := self currentScope tempMap.
 	
 	"adding the same var multiple times reuses the same offset"
-	(tempMap includesKey: tempKey) ifTrue: [ ^ self ].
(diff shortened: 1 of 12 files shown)


MarcusDenker:
Now that all the other compiler changes are in (and debugged) I would like to merge this to see if it really works (I did do some tests to check that all the temps do get mapped, but one never nows...)

PR #6199, opened 2020-04-21
Merge Sindarin BCOffsetCache

MarcusDenker:
This PR integrates the SindarinBytecodeToASTCache as OCBytecodeToASTCache

=> API changed to require a RBMethodNode, not the compiledMethod
=> improved the case when asking for start pc for methods with primitives

This is another step towards making the IR an transient datastructure that is not cached

### DIFF
--- a/src/OpalCompiler-Core/OCBytecodeToASTCache.class.st
+++ b/src/OpalCompiler-Core/OCBytecodeToASTCache.class.st
@@ -1,0 +1,95 @@
+"
+I cache a mapping between possible bytecode offsets and the AST nodes they correspond to for a given compiled method.
+
+Instanciate me using my class method generateForCompiledMethod: and give me as parameter a compiled method.
+
+Use me through the node access API method nodeForPC: and give me a program counter as parameter.
+
+I store: 
+- firstBcOffset: The first bytecode pc. If you try to access a pc below that first pc, I return the method node.
+- lastBcOffset: The last bytecode pc. If you try to access a pc after this last pc, I return the node associated with the last pc.
+- bcToASTMap: A map associating each possible pc between firstBcOffset and lastBcOffset and the corresponding ast node.
+- the methode node.
+"
+Class {
+	#name : #OCBytecodeToASTCache,
+	#superclass : #Object,
+	#instVars : [
+		'firstBcOffset',
+		'lastBcOffset',
+		'bcToASTMap',
+		'methodNode'
+	],
+	#category : #'OpalCompiler-Core-Bytecode'
+}
+
+{ #category : #initialization }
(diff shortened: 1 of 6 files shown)


src/OpalCompiler-Core/OCBytecodeToASTCache.class.st
```diff
@@ -0,0 +1,95 @@
...
+{ #category : #initialization }
+OCBytecodeToASTCache >> generateForMethodNode: aNode [
+	| methodIR currentBcOffset |
+	methodNode := aNode.
+	methodIR := methodNode ir.
+	bcToASTMap := Dictionary new.
+	firstBcOffset := methodNode compiledMethod initialPC.
+	
+	"if there is a primitive, we need to record the offset"
+	(methodIR irPrimitive num ~= 0) ifTrue: [ bcToASTMap at: firstBcOffset put: methodNode].
```
StevenCostiou, line 68: This case with primitives is not covered by a test. 
Could you add one?

Could you give more details about that case? I would like to document it in the compiler booklet.

src/OpalCompiler-Core/RBMethodNode.extension.st
```diff
@@ -167,7 +167,8 @@ RBMethodNode >> rewriteTempsForContext: aContext [
 
 { #category : #'*OpalCompiler-Core' }
 RBMethodNode >> sourceNodeForPC: anInteger [
-	^(self ir instructionForPC: anInteger) sourceNode
+	bcToASTCache ifNil: [ bcToASTCache := OCBytecodeToASTCache generateForMethodNode: self].
```
StevenCostiou: If we add a lazy accessor to bcToASTCache, we could just have 
`^ self bcToASTCache nodeForPC: anInteger`.

What do you think?

StevenCostiou (changes requested):

src/OpalCompiler-Core/OCBytecodeToASTCache.class.st
MarcusDenker: There is #testPrimitiveMethodSourceNodeAtInitialPC which failed without this.

So the thing is that primitives are modelled with a bytecode, but there irPrimitive is not part of the other IR instructions, but instead the IRMethod has an ivar that references is.

This is something we could actually change.

MarcusDenker: (but not as part of this PR)

src/OpalCompiler-Core/RBMethodNode.extension.st
MarcusDenker: Yes!

StevenCostiou (approved):

PR #6207, opened 2020-04-23
refactor-spanNewSessionFrom

estebanlm:
this refactor allows the extension of #spanNewSession by subclasses maybe this has to be cleaned in the future instead? 
anyway while GTDebugger is still around, this is needed.


MarcusDenker (approved):

Ducasse:
Esteban is it not spawnNewSession because span looks strange to me

tesonep (approved):

MarcusDenker:
this breaks debugging tests

PR #6209, opened 2020-04-23
ReNodeRewriteRule subclasses might change the AST

### LINKED ISSUE #6201: super sends mess up parent chain in AST
**Describe the bug**
I've found source code that produces an AST in which a node has the wrong parent.

**To Reproduce**
Evaluate the code below. For some reason the problem does not always occur. It should show up on the block node (where I've placed the halt). If you can't find the problem when evaluating, browse the method and you should see an error in the critics: "MessageNotUnderstood: receiver of 'methodNode' is nil".

Comparing the parents in this situation shows that the parent of the block node has `self` instead of `super` in the generated source.

```smalltalk
| block |
Object subclass: #Foo
	instanceVariableNames: ''
	classVariableNames: ''
	package: 'Hacks'.
(Smalltalk at: #Foo) compile: 'addKeyToSequence: aString
	super at: aString ifAbsent: [Smalltalk addLast: aString]'.
block := [ :node |
	node children do: [ :child |
		self assert: child parent == node.
		self assert: child methodNode notNil.
		self haltIf: [ child class = RBBlockNode ].
		child nodesDo: block ] ].
((Smalltalk at: #Foo) >> #addKeyToSequence:) ast nodesDo: block
```

**Expected behavior**
Asking the block node from the above AST for `#methodNode` should return the top node of the AST. It answers nil.

**Screenshots**
Screenshot 2020-04-22 at 13 49 34

**Version information:**
 - OS: macOS
 - Version: Pharo 9 build: 245, commit: 7879085 (also happens in Pharo 8)

MarcusDenker:
-> in #nodeCritiquesOf: switch around the check: run all checks on one method after the other. This is possible as we cache the AST.
-> for Rewrite rules, make a copy of the AST for each rule run. Coypying the AST is faster then re=parsing.

fixes #6201  ( I hope)

### DIFF
--- a/src/Renraku/ReAbstractRule.class.st
+++ b/src/Renraku/ReAbstractRule.class.st
@@ -225,6 +225,11 @@ ReAbstractRule >> isComposite [
 	^ false
 ]
 
+{ #category : #testing }
+ReAbstractRule >> isRewriteRule [
+	^false
+]
+
 { #category : #accessing }
 ReAbstractRule >> name [
 	"Answer a human readable name of the rule."
--- a/src/Renraku/ReCriticEngine.class.st
+++ b/src/Renraku/ReCriticEngine.class.st
@@ -116,7 +116,7 @@ ReCriticEngine >> nodeCritiquesOf: aMethod [
 			| ast |
 			ast := aMethod ast.
 			"for rewrite rules, we run every rule on a copy of the ast"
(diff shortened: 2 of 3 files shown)


MarcusDenker:
mac ci has some unrelated problem, but all tests are green on the two other platforms

PR #6211, opened 2020-04-24
6104-MethodNode-should-offer-a-nice-way-to-access-primitive-number

### LINKED ISSUE #6104: MethodNode should offer a nice way to access primitive number
**Describe the problem**
Right now to access the primitive number of a primitive in a method node
we should do 

```
self pragmas first argumentAt: 'primitive:'
```

And this is not really user friendly. 
The solution should be to propose

```
self primitiveNumber 
```

MarcusDenker:
add a small helper method to return the primitive number if the AST defines a primitive by pragma. 

Add a test to explain what it does.

fixes #6104

### DIFF
--- a/src/AST-Core-Tests/RBMethodNodeTest.class.st
+++ b/src/AST-Core-Tests/RBMethodNodeTest.class.st
@@ -169,7 +169,6 @@ RBMethodNodeTest >> testPrimitiveModuleErrorIsPrimitive [
 	
 ]
 
-
 { #category : #'tests - primitives' }
 RBMethodNodeTest >> testPrimitiveModuleIsPrimitive [
 	"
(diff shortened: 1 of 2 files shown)


Ducasse:
Thanks marcus it was on my todo. Now I was wondering how to handle primitive that do not have a primitive number so may be we should have a primitive tag?

Ducasse:
I started to write that kind of code :(
```
(aMessageNode methodNode pragmas first argumentAt: 'primitive:') value 
```
for a new book and I thought that it is not good.

MarcusDenker:
No, the thing is that they do have a primitiveNumber, too.

Named primitives are all num 117

Ducasse:
Yes I know that. 
Then I was wondering how to access the tag associated with it

screenDepth
	<primitive: #primitiveScreenDepth >
	^ self renderer form depth
or 

isYoung: oop
	"Primitive. Answer whether the object currently resides in youngSpace."
	<primitive: 'primitiveIsYoung'>
	^self primitiveFailed

profileSemaphore: aSemaphore
	"Primitive. Install the semaphore to be used for profiling, 
	or nil if no semaphore should be used.
	See also primitiveProfileStart."
	<primitive: 'primitiveProfileSemaphore'>
	^self primitiveFailed

MarcusDenker:
I think what you want is a model of primitives on the AST level (that do not care about implementation details)

There a pritive would just have the information that is there on the text level: a number, a tag.

While what we do on the IR level is something else: that primitive model is a more of *compiled* primitive, where e.g. named primitives have a number even though this is not in the AST.

MarcusDenker:
I will close this for now and think a bit more... using the low level IR model of primitives seems to be not good here.

Issue #6172, opened 2020-04-13
Syntax highlighter doesn't work with capitalised symbols 


VincentBlondeau:
**Describe the bug**
When your write # followed by a capitalized letter A-Z in the code editor (like #World), the syntax highlighter doesn't update the syntax coloring of the method anymore.

**To Reproduce**
Go see an example here:
`(BaselineOfMorphicCore >> #postload:package:) browse`

Live example:
![LjRUXxUa04](https://user-images.githubusercontent.com/4706329/79122026-fd82de00-7d96-11ea-832e-76f184ab143f.gif)

**Expected behavior**
You know

**Screenshots**
![LjRUXxUa04](https://user-images.githubusercontent.com/4706329/79122026-fd82de00-7d96-11ea-832e-76f184ab143f.gif)

**Version information:**
 - OS: WIN10
 - Version: Pharo9.0.0 build: 233, commit: 989264b

**Expected development cost**
Base number:
- Someone that know the framework: x= 30 mins.
- Me: x= 180 mins
*Estimated Total Development cost* : x * 2 + 10

**Additional context**

MarcusDenker:
this has been merged

PR #6212, opened 2020-04-24
6172-Syntax-highlighter-doesnt-work-with-capitalised-symbols

tesonep:
Fixes issue #6172
The global variables maybe do not point to a class, it is the example of World.


Ducasse:
So cool to have more tests!

Issue #6186, opened 2020-04-15
System end up in a bad state when adding a class in a trait composition instead of a trait


jecisc:
**Describe the bug**
While writing a test I misstyped and set up a Class as a trait composition of another class. 

After that, the system react pretty badly. I cannot remove the trait composition. I cannot remove the class. I cannot remove the package...

**To Reproduce**

```Smalltalk
	| class trait |
	class := Object
		subclass: #ChanelMethodOnlyCallingSuperFake
		instanceVariableNames: ''
		classVariableNames: ''
		package: self class package name.
	trait := Object
		subclass: #TChanelMethodOnlyCallingSuperFake
		instanceVariableNames: ''
		classVariableNames: ''
		package: self class package name..

	class setTraitComposition: trait.
```

**Expected behavior**
I would prefer to get an explicit error during the compilation saying that it is not possible to use a class as a trait composition.

**Version information:**
 - Version: Pharo 8

Ducasse:
Thanks. I cannot see in your example that you added a class as a composition. 
```
Object subclass: #2643
    uses: Point
...
```
Is it it?

jecisc:
`class setTraitComposition: trait`

Ducasse:
Yes I read  it but I do not get it. 
So we will see.

Ducasse:
Could you have a reproducible case? Because else we have to try and see.

Ducasse:
Ah I see trait is a class!

tesonep:
I will fix it. 
Do we need to add a validation also if you put your cat in the microwave? :)

tesonep:
Fixed in #6214

MarcusDenker:
this was fixed

PR #6214, opened 2020-04-24
6186-System-end-up-in-a-bad-state-when-adding-a-class-in-a-trait-composition-instead-of-a-trait

tesonep:
Adding validation that the members of a trait composition are traits.
Fixes issue 6186


tesonep:
Fixes issue #6186

MarcusDenker (approved):

PR #6216, opened 2020-04-24
Fix #5854. SycMethodNameEditor fails with DNU in some cases

### LINKED ISSUE #5854: Add argument in extract method throws exception
When doing an extract method from the source editor

image

In the extract method dialog

image

If we add an argument with a unary selector and click ok, we have an index out of bounds

image

guillep:
Fix #5854

This PR enhances the UI so invalid states do not happen (disabling buttons in those cases)

- Rename MethodNameEditorTest => SycMethodNameEditorTest to follow co…
- add accessors to SycMethodNameEditor
- disable parameter edition input and button if no there is no selection
- many tests


Ducasse:
Oh this is cool! I love these tests. So cool feeling.

PR #6219, opened 2020-04-24
Fix #5853: some enhancements to inline senders refactoring

### LINKED ISSUE #5853: Inline senders Refactoring is strange
1. It is not clear what it does: does it inline the methods used within the selected method? Or does it inline the selected method in its senders? It would make sense that it is the second one. For the 
other, we can do manually `inline method`.

2. For some reason, when using the refactoring, it does not work but it is not explained why. So I cannot take any decision.

image

guillep:
Fix #5853

- better error message when not applicable
- make it work from the text editor and the method list

### DIFF
--- a/src/Refactoring-Core/RBInlineAllSendersRefactoring.class.st
+++ b/src/Refactoring-Core/RBInlineAllSendersRefactoring.class.st
@@ -49,7 +49,7 @@ RBInlineAllSendersRefactoring class >> sendersOf: aSelector in: aClass [
 { #category : #transforming }
 RBInlineAllSendersRefactoring >> checkInlinedMethods [
 	numberReplaced = 0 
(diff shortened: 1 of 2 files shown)


Ducasse:
I think that unix-64 / Tests-unix-64 / testExample – Unix64.Spec.Tests.Examples.ListSelectionPresenterTest is not related to your change so I integrate it.

guillep:
> I think that unix-64 / Tests-unix-64 / testExample – Unix64.Spec.Tests.Examples.ListSelectionPresenterTest is not related to your change so I integrate it.

Yes, I also think it is not related.
I tried to see how to test this but this exceeded my timebox for the issue (although I strongly believe we should slowly but steadily start testing the browser and all tools...)

PR #6220, opened 2020-04-24
6171-Enhance-Collection-with-rollingAvegra

### LINKED ISSUE #6171: Enhance Collection with rollingAvegra
**Expected behavior**
Like that we can have rolling average :)

**Expected development cost**
Nearly none just turn the following code into a PR. 

```
Collection >> runningMeans: width
     "This returns an array of running means as if the receiver
      were broken into overlapping segments 'width' long and the
      mean of each calculated.  This uses an adaptation of
      Welford's algorithm for stably updating the mean; it is
      important to maintain a current MEAN not a current SUM.
      This has been tested against 7 other algorithms.  It was
      the most accurate of the faster ones.  The result is an
      Array no matter what kind of sequence the receiver is.
      If the receiver is a tree (like a virtual concatenation)
      or a singly or doubly linked list you should convert the
      receiver to an Array first.  Note that there is no
      explicit check that width is an Integer or is in range;
      none is needed because those checks happen anyway."

    "(#(1 1 2 2 3 3) asOrderedCollection runningMeans: 2) >>> {1. (3/2). 2. (5/2). 3}".

     |result mean resultIndex|
     result := Array new: self size - width + 1.
     mean := 0.
     1 to: width do: [:i | mean := (self at: i) + mean].
     mean := mean / width.
     resultIndex := 1.
     result at: resultIndex put: mean.
     width + 1 to: self size do: [:i |
       mean := ((self at: i) - (self at: resultIndex)) / width + mean.
       resultIndex := resultIndex + 1.
       result at: resultIndex put: mean].
     ^result
```

```
OrderedCollection >> testRunningMeans

	| result col |
	col := #(1 1 2 2 3 3) asOrderedCollection.
	result := col runningMeans: 2. 

	self assert: (result = {1. (3/2). 2. (5/2). 3}).
	
	self assert: (result class = Array).
	
	self assert: result size <= (col size). "Running means with 1 has little interest ?"
	
	self should: [ col runningMeans: 7 ] raise: SubscriptOutOfBounds.
	self  should: [ col runningMeans: -2 ] raise: SubscriptOutOfBounds.
	self  should: [ col runningMeans: 1.3 ] raise: Error withExceptionDo: [ :anException | self assert: anException messageText equals: 'primitive #basicNew: in Array class failed’ ]

```

```
testRunningMeansDeviation
OrderedCollection >> 
	| result1 result2 col1 col2 |
	
	col1 := #(0.3s1 1s1 0.1s1 0.3s1 1s1 0.1s1 0.3s1 1s1 0.1s1).
	result1 := col1 runningMeans: 2.
	
	col2 := #(0.3 1 0.1 0.3 1 0.1 0.3 1 0.1). 
	result2 := col2 runningMeans: 2.
	
	self assert: result1 first equals: result1 fourth.
	
	"presence of a rounding error"
	self deny: result2 first equals: result2 fourth.
	self assert: result2 first closeTo: result2 fourth.
```

olekscode:
Fixed #6171

### DIFF
--- a/src/Collections-Abstract/Collection.class.st
+++ b/src/Collections-Abstract/Collection.class.st
@@ -319,7 +319,7 @@ Collection >> asSortedCollection [
 
 { #category : #converting }
 Collection >> asSortedCollection: aSortBlock [ 
-	"Answer a SortedCollection whose elements are the elements of the receiver. The sort order is defined by the argument, aSortBlock. Note that this is better to use #sorted: if you don't really need a SortedCollection, but a sorted collection!!"
+	"Answer a SortedCollection whose elements are the elements of the receiver. The sort order is defined by the argument, aSortBlock. Note that it is better to use #sorted: if you don't really need a SortedCollection, but a sorted collection!!"
 
 	| aSortedCollection |
 	aSortedCollection := SortedCollection new: self size.
--- a/src/Collections-Arithmetic-Tests/CollectionArithmeticTest.class.st
+++ b/src/Collections-Arithmetic-Tests/CollectionArithmeticTest.class.st
@@ -55,3 +55,65 @@ CollectionArithmeticTest >> testAverageWithEmptyOrderedCollection [
 CollectionArithmeticTest >> testAverageWithEmptySet [
 	self should: [ Set new average ] raise: CollectionIsEmpty
 ]
+
+{ #category : #tests }
+CollectionArithmeticTest >> testRunningAverage [
+	| result collection |
+	collection := #(1 1 2 2 3 3).
+	result := collection runningAverage: 2. 
+
+	self assert: result equals: {1 . (3/2) . 2 . (5/2) . 3}.
+
+]
+
+{ #category : #tests }
+CollectionArithmeticTest >> testRunningAverageSubscriptOutOfBounds [
+	| collection |
+	collection := #(1 1 2 2 3 3).
+
+	self should: [ collection runningAverage: 7 ] raise: SubscriptOutOfBounds.
+	self  should: [ collection runningAverage: -2 ] raise: SubscriptOutOfBounds.
+
+]
+
+{ #category : #tests }
+CollectionArithmeticTest >> testRunningAverageWithFullSubsetSizeIsSameAsAverage [
+	| collection |
+	collection := #(1 1 2 2 3 3).
+	
+	self
+		assert: (collection runningAverage: collection size) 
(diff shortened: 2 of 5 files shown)


svenvc:
I like it, especially the tests.

But your implementation generates new arrays for every step, that is a lot of garbage, right ?

What if instead of copying, you create only one, reusing it by copying elements into it ?

Ducasse:
Could we have 

```
^ self running: #average of: aSubsetSize
```

=> 
```
^ self running: [:each | each average] of: aSubsetSize
```

olekscode:
OK, I will try to improve it.
@Ducasse is it a bad practice to use selector instead of a block?

Ducasse:
For now we try to avoid in the core.

Ducasse:
Did you bench the diff on large or super large arrays?

olekscode:
I have replaced symbolic selectors with blocks and updated the PR:
```Smalltalk
^ self running: [:each | each average] of: aSubsetSize
```

olekscode:
@svenvc I profiled the `runningAverage`, it takes approximately 3.5 seconds to compute the average of triples running over 20 million numbers:
```Smalltalk
| array |
array := (1 to: 20000000) asArray.
array runningAverage: 3.
```
Of that only 70 milliseconds is taken by `copyFrom:to:`. So it doesn't seem like a big deal (but maybe I'm wrong, maybe garbage collection takes time later and I'm looking at a wrong thing in a Profiler - I don't know).

Also, I am not sure what would be a good way to avoid creating a new collection for every subset.
I could do something like this:

```Smalltalk
subset := self species ofSize: aSubsetSize.
	
result := (1 to: self size - aSubsetSize + 1) collect: [ :i |
    (i to: i + aSubsetSize - 1) collect: [ :j | self at: j ] into: subset.
    aBlock value: subset ].
```
But then the block would always be evaluated on a reference to the same array, which can result in unexpected behaviour. For example:
```Smalltalk
#(1 2 3 4 5) running: [ :each | each ] of: 2.
 "#(#(4 5) #(4 5) #(4 5) #(4 5))"
``` 
And for some reason, this alternative implementation is almost 2 times slower.
Maybe I'm doing it wrong, so please let me know if you have a better solution

Ducasse:
Pay attention you have mixed code from another PR in this PR (about intersection). 
I think that this is ok since we integrated the PR of Pablo on intersection but :)

PR #6222, opened 2020-04-24
heapSortExample should have an example tag #6204

### LINKED ISSUE #6204: heapSortExample should have an example tag
Heap >> heapSortExample should get a <sampleInstance>

LaurineDargaud:
Fix: #6204 
heapExample and headSortExample are now examples

### DIFF
--- a/src/Collections-Sequenceable/Heap.class.st
+++ b/src/Collections-Sequenceable/Heap.class.st
@@ -68,7 +68,7 @@ Heap class >> defaultSortBlock [
 ]
 
 { #category : #examples }
-Heap class >> heapExample [	
+Heap class >> heapSimpleBenchmark [
 	<sampleInstance>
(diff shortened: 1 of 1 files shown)


Ducasse:
You should use <sampleInstance> and do not add inspect at the end. 
because this instance can be used from tests and you do not want the pop up of the inspector
now this is the tool that should take care of spawning an inspector or not. not us.

tesonep:
@LaurineDargaud Maybe we should remove the test method 
HeapTest>>testExamples
It is not adding anything, it is just a smoke test.

tesonep:
Also it is failing because of the rename of the examples to benchmark.
But I think we should remove it.

Ducasse:
Hi pablo
I will have a look because this case is a bit messy. So I will do another PR.

Issue #5952, opened 2020-03-21
PharoEnvironmentHelp outdated


dvtomas:
There's a lot of outdated information in the PharoEnvironmentHelp class comment (propagated to #wikiStyleHelp). A review would be nice to avoid confusing users.

oliveiraallex:
https://github.com/pharo-project/pharo/pull/6225

CafeKrem:
why this issue is still open when there is a PR merged ?  
is there some work to do? it's a bit confusing

MarcusDenker:
if there is a "fixes #number" in the PR, the issue gets closed automatically. If not, it has to be closed manually. 

I will close this.

PR #6225, opened 2020-04-24
#5952-PharoEnvironmentHelp-outdated

oliveiraallex:
Updated Pharo Enviroment Help description


myroslavarm (approved):
these are good updates! (although i created an issue about shortcuts (https://github.com/pharo-project/pharo/issues/6229) because it's weird that there is a separate tool in Pharo for browsing shortcuts -- KeymapBrowser -- and it does not work). another thing is that you should link your commit/PR with the issue you are referencing, so you could actually browse the PR reference in the issue and the issue reference in the PR: you can do this by typing # and issue number

myroslavarm:
the last thing i mentioned is just for going to the issue and seeing that there's an existing pull request fixing it, because otherwise it is not clear

PR #6227, opened 2020-04-24
ReBasicTreeRule-Cleanup

MarcusDenker:
- change FloatReferencesRule to be a ReBasicTreeRule
- fix tests to use changed API
- delete deprecated ReBasicTreeRule
- small cleanup: categorize ReTestCaseShouldNotUseInitialize


src/GeneralRules/ReTestCaseShouldNotUseInitialize.class.st
```diff
@@ -7,7 +7,7 @@ Initialize should only use variables defined in #instanceVariablesToKeep.
 Class {
 	#name : #ReTestCaseShouldNotUseInitialize,
```
myroslavarm: it would be nice if you could rename it to end with Rule, i.e. `ReTestCaseShouldNotUseInitializeRule`. i made a PR renaming all the other rules but i'm concerned that me renaming this one in my commit could break merge compatibility maybe?

MarcusDenker: I will commit a new version of the PR without this change (it was not related to the other changes)

myroslavarm: ok cool, so i renamed and recategorised it in my commit

myroslavarm (approved):

PR #6228, opened 2020-04-24
5494-Rename-rules-to-end-up-with-Rule

### LINKED ISSUE #5494: Rename rules to end up with Rule!!!
ReTestClassNameShouldEndWithTest is a rule should it should terminate with Rule.
We should make also a rule for that!!!!

myroslavarm:
fixes #5494

### DIFF
--- a/src/GeneralRules-Tests/ReGlobalVariableRuleTest.class.st
+++ b/src/GeneralRules-Tests/ReGlobalVariableRuleTest.class.st
@@ -7,7 +7,7 @@ Class {
 { #category : #tests }
 ReGlobalVariableRuleTest >> applyingCritiques [
 	^ self class methods
-		select: [ :m | (ReGlobalVariablesUsage new check: m) isNotEmpty ]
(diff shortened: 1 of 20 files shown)


MarcusDenker:
ReGlobalVariablesUsageTest
ReTestCaseShouldNotUseInitializeTest

are failing (I guess they reference renamed rules by name)

myroslavarm:
@MarcusDenker it's weird because they don't reference rules by names in the code (and I renamed everything automatically so the references should be updated in any case), but it seems the problem is with the tests names themselves? Let's see if my last commit fixes the fail.

MarcusDenker (approved):

astares:
@myroslavarm Would be good to add a Release test (see package with this name) to check that included rules end up with "Rule" in the name.

So we keep the quality status and make sure future rules added later to the image follow the same pattern.

myroslavarm:
@astares yeah i thought about it but wasn't sure how to check if something is a rule... i will look into this once i have more time on my hands in a couple of weeks!

astares:
@myroslavarm Thx!

PR #6230, opened 2020-04-24
Rewrote algorithm of String>>#skipAnySubstring:start:

### LINKED ISSUE #6191: findBetweenSubstrings: freezes the image
**Describe the bug**
Running the following piece of code freezes the image 
```
'text' findBetweenSubstrings: #('te' 'x')
```
whereas
```
'text' findBetweenSubstrings: #('te' 't')
>>> a Collection('x')
```

**Version information:**
 - Version: Pharo9

dupriezt:
- Rewrote algorithm of String>>#skipAnySubstring:start: to remove bug,  use more explicit variable names, add comment and improve the structure.
- Added test method for String>>#skipAnySubstring:start

Fix #6191

### DIFF
--- a/src/Collections-Strings/String.class.st
+++ b/src/Collections-Strings/String.class.st
@@ -2203,26 +2203,28 @@ String >> sameAs: aString [
 
 { #category : #accessing }
 String >> skipAnySubstring: delimiters startingAt: start [ 
-	"Answer the index of the last character within the receiver, starting at start, that does NOT match one of the delimiters. delimiters is a Array of substrings (Characters also allowed).  If the receiver is all delimiters, answer size + 1."
-
-	| any this ind ii |
-	ii := start-1.
-	[(ii := ii + 1) <= self size] whileTrue: [ "look for char that does not match"
-		any := false.
-		delimiters do: [:delim |
-			delim isCharacter 
-				ifTrue: [(self at: ii) == delim ifTrue: [any := true]]
-				ifFalse: ["a substring"
-					delim size > (self size - ii + 1) ifFalse: "Here's where the one-off error was."
-						[ind := 0.
-						this := true.
-						delim do: [:dd | 
-							dd == (self at: ii+ind) ifFalse: [this := false].
-							ind := ind + 1].
-						this ifTrue: [ii := ii + delim size - 1.  any := true]]
-							ifTrue: [any := false] "if the delim is too big, it can't match"]].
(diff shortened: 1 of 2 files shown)


Ducasse:
Oh this is cool!. Thanks.

Ducasse:
Just a question. Did you bench it a bit to see if we are in the same area in terms of complexity?

dupriezt:
> Just a question. Did you bench it a bit to see if we are in the same area in terms of complexity?

I don't know if there's a specific method for benchmarking performance in pharo.
Here's what I did:
I ran the following code in a time profiler on an original pharo 9 image, and on one with my change, and looked at the time spent inside the #skipAnySubstring:startingAt: method. I did 10 measures for each case.
**The code:**
`1000000 timesRepeat: ['ababc' skipAnySubstring: 'ab' startingAt: 1]`
**Original Method**
Measures (ms): `305 286 313 302 290 276 267 283 283 283`
Average (ms): `288.8`
**Modified Method**
Measures (ms): `1037 1016 1093 1056 1099 1120 1076 1124 1119 1073`
Average (ms): `1081.3`
**Increase in Time: 374%**

So what should I do from there? This PR fixes an infinite loop problem in the original method, makes it so the method tries to skip the largest substring it finds (instead of the first one it finds) and rewrites the algorithm to "look nicer" (using #collect:, #beginsWith:, #max... instead of doing it imperative-style). It's probably possible to rewrite the fixed algorithm in a more imperative style.

*Note: Some tests were failing because they called the method directly with a string instead of with an array containing the string. I fixed that, and am now waiting on the next result of the CI to see if some tests remain broken.*

dupriezt:
The broken tests that remain fail because they take too much time

Ducasse:
Hi Thomas

bench and timeToRun are the way to bench and you have smark for benching. 
Usually we do a run multiple thousand times. 
It would be good to see if we can use this more imperative style. 

S

guillep:
You don't want to bench. You want to profile :).
I tried profiling it a bit, but it becomes more evident with larger strings in addition to more iterations.

One main difference between the original algorithm and the new one is that the new one creates copies of substrings using `last:` to use `beginsWith:`. Maybe it coud be implemented alternatively using `findAnySubstring:startingAt:` or `findString:startingAt:` and keeping track of the current search index?

guillep:
Seems there is no interest for this PR from the author. I'll close. Reopen if needed.

PR #6235, opened 2020-04-27
ReAssignmentOnBlockArgumentRule-Remove

MarcusDenker:
ReAssignmentOnBlockArgumentRuleTest checks that an argument of a block is not assigned to.

This rule made sense in older versions where one could assign to arguments. This is not possible anymore: in interactive mode, this code is rejected. In non-interactive mode it clearly raises an compiler error on the transcript.

I think we can remvoe thos tule (there is no rule that checks for assigning to method args, either)


Ducasse:
Indeed we cannot be in such state.

PR #6237, opened 2020-04-27
DeadCodeCleanup: Rules

MarcusDenker:
Small dead code cleanup in Renraku and GeneralRules

- remove old deprecated methods
- remove unsent private methods


src/Renraku/RBTransformationRule.extension.st
```diff
@@ -12,13 +12,3 @@ RBTransformationRule >> check: aMethod forCritiquesDo: aCriticBlock [
 					tree: rewriteRule tree)  ]
 		
 ]
-
```
Ducasse: Marcus I think that this one is nice to keep.

Ducasse: Talking about suggestedCode:

MarcusDenker: My problem is that I deleted the image already and I still do no know how to load this PR now and update it...

Ducasse:
Check your remote and it should contain all your branches. 
And you can checkout that branch. 
And this is a nice XP to do.

Ducasse:
Thanks!

PR #6242, opened 2020-04-27
Improve Tests OCBytecodeToASTCache

MarcusDenker:
OCBytecodeToASTCacheTest had some tests that tested against the old implementation. As we replaced this with the cache, we need to rewite the tests to compare something else (e.g which kind of node)

### DIFF
--- a/src/OpalCompiler-Tests/OCBytecodeToASTCacheTest.class.st
+++ b/src/OpalCompiler-Tests/OCBytecodeToASTCacheTest.class.st
@@ -36,9 +36,10 @@ OCBytecodeToASTCacheTest >> testFirstBCOffsetTest [
 OCBytecodeToASTCacheTest >> testHigherThanLastBCOffsetAccessTest [
 	| pc |
 	pc := cache lastBcOffset + 5.
+	"if we are beyond the last bc, we map to the whole method"
 	self
 		assert: (cache nodeForPC: pc)
-		identicalTo: (compiledMethod sourceNodeForPC: pc)
+		identicalTo: compiledMethod ast
 ]
 
 { #category : #tests }
@@ -60,21 +61,17 @@ OCBytecodeToASTCacheTest >> testLowerThanFirstBCOffsetAccessTest [
 { #category : #tests }
 OCBytecodeToASTCacheTest >> testNodeForBCOffsetRangeTest [
 	"As we associate each node to each possible bytecode offset that can refer to it,
-	we have to check that associations are consistent between a range and a node"
+	we have to check that associations are all pointing to a node"
 
 	| pcRange |
 	pcRange := 0 to: cache lastBcOffset.
-	pcRange do: [ :pc | 
-		self
-			assert: (cache nodeForPC: pc)
-			identicalTo: (compiledMethod sourceNodeForPC: pc) ]
+	pcRange do: [ :pc | self assert: ((cache nodeForPC: pc) isKindOf: RBProgramNode)]
 ]
 
 { #category : #tests }
 OCBytecodeToASTCacheTest >> testNodeForBCOffsetTest [
 	| pc |
+	"there should be an assignment at pc 51"
 	pc := 51.
-	self
-		assert: (cache nodeForPC: pc)
-		identicalTo: (compiledMethod sourceNodeForPC: pc)
+	self assert: (cache nodeForPC: pc) isAssignment
 ]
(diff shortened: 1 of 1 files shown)


src/OpalCompiler-Tests/OCBytecodeToASTCacheTest.class.st
```diff
@@ -60,21 +61,17 @@ OCBytecodeToASTCacheTest >> testLowerThanFirstBCOffsetAccessTest [
...
-	we have to check that associations are consistent between a range and a node"
+	we have to check that associations are all pointing to a node"
 
 	| pcRange |
 	pcRange := 0 to: cache lastBcOffset.
-	pcRange do: [ :pc | 
-		self
-			assert: (cache nodeForPC: pc)
-			identicalTo: (compiledMethod sourceNodeForPC: pc) ]
+	pcRange do: [ :pc | self assert: ((cache nodeForPC: pc) isKindOf: RBProgramNode)]
```
StevenCostiou, line 68: I agree in the end that is better, however we are only testing that it maps to a node but not the correct one.
I would do a brute check: for each pc, check that it is the correct node.
Maybe we can generate it?

src/OpalCompiler-Tests/OCBytecodeToASTCacheTest.class.st
```diff
@@ -60,21 +61,17 @@ OCBytecodeToASTCacheTest >> testLowerThanFirstBCOffsetAccessTest [
...
 
 { #category : #tests }
 OCBytecodeToASTCacheTest >> testNodeForBCOffsetTest [
 	| pc |
+	"there should be an assignment at pc 51"
 	pc := 51.
-	self
-		assert: (cache nodeForPC: pc)
-		identicalTo: (compiledMethod sourceNodeForPC: pc)
+	self assert: (cache nodeForPC: pc) isAssignment
```
StevenCostiou: I would add additional checks. I put that example here thinking that it refers to the assignment `i := 5` of `helper12` method (but same idea if that's the other assignment).

```Smalltalk
| pc node |
"there should be an assignment at pc 51"	
pc := 51.
node := (cache nodeForPC: pc).
self assert: node isAssignment.
self assert: node variable equals: i. "find the i var node in the method"
self assert: node variable equals: 5. "put a literal node here to check for equality"
```

MarcusDenker: I change it to: 	

self assert: (cache nodeForPC: pc) sourceCode equals: 'i := i - 1'.

src/OpalCompiler-Tests/OCBytecodeToASTCacheTest.class.st
```diff
@@ -73,5 +73,5 @@ OCBytecodeToASTCacheTest >> testNodeForBCOffsetTest [
 	| pc |
 	"there should be an assignment at pc 51"
 	pc := 51.
-	self assert: (cache nodeForPC: pc) isAssignment
+	self assert: (cache nodeForPC: pc) sourceCode equals: 'i := i - 1'.
```
StevenCostiou, line 76: I would also check:
- node class
- node start
- node stop
This way we're sure it is the correct one, as source code strings can be printed from different nodes.

src/OpalCompiler-Tests/OCBytecodeToASTCacheTest.class.st
StevenCostiou: I generated a test for this method: https://github.com/StevenCostiou/OpalASTMappingTestGenerator

MarcusDenker: I add this to the backlog, I would like to merge the current state

src/OpalCompiler-Tests/OCBytecodeToASTCacheTest.class.st
MarcusDenker: There is just one expression "i := i - 1" in th method under test. there is no way that we print that string and it is not the node we are looking for

StevenCostiou: But I'm concerned by the fact that we may change the test method one day, and introduce such subtility that would make the test pass while it should not.
As we're testing the node mapping, I think we need to always make sure that we precisely identify the node we test.

MarcusDenker: ok, added stop and start check

MarcusDenker:
I integrate this, next steps will be another PR

PR #6245, opened 2020-04-28
Make all users of manifests use #uniqueIdentifierName to construct selectors

### LINKED ISSUE #6244: The methods for banning critic rules use the rule name instead of the unique identifier
**Describe the bug**
The methods for banning critic rules use the rule name instead of the unique identifier. That has no effect in the system browser because the method that checks for bans also uses the rule class name. However, the ReSmallLintChecker uses `#uniqueIdentifierName` to check for banned rules and will thus not exclude matches when the rule class name is different from `#uniqueIdentifierName`.
Vice versa, banning a rule from the critic browser will not ban it from the system browser.

**To Reproduce**
1. Evaluate this:
```smalltalk
Object subclass: #Foo
	instanceVariableNames: ''
	classVariableNames: ''
	package: 'Hacks'.
(((Smalltalk at: #Foo) compile: 'foo nil isKindOf: UndefinedObject'; yourself) >> #foo)  browse
```
This will open a browser on a new method with a bad message send.
2. Ban the rule on the method level, the warning in the browser will disappear
3. Open the Critic browser
4. Select the "Hacks" package, click "Add" then "Next"
5. Under "Coding Idiom Violation" select "Sends 'questionable' message, click "Add" then "Next"
6. You will now see one failed rule for the method just created.

**Expected behavior**
All uses of the manifests should agree on which rules are banned.

**Version information:**
 - OS: macOS
 - Version: Pharo-9.0.0+build.265.sha.c82e3cc26b844b585dccb56874523a70d8608c78 (64 Bit)

I'll provide a fix for the manifest builder. I will open a separate issue for updating manifests, as all bans that use class names instead of `#uniqueIdentifierName` need to be updated:
```smalltalk
ReAbstractRule allSubclasses select: [ :class |
	class uniqueIdentifierName ~= class name ]
```

theseion:
Fixes #6244

### DIFF
--- a/src/Renraku/TheManifestBuilder.extension.st
+++ b/src/Renraku/TheManifestBuilder.extension.st
@@ -2,25 +2,27 @@ Extension { #name : #TheManifestBuilder }
 
 { #category : #'*renraku' }
 TheManifestBuilder >> banRule: aRule for: anEntity [
-	| name number |
-	  name := aRule class name.
+	| ruleId number |
+	ruleId := aRule class uniqueIdentifierName.
 	number := aRule class identifierMinorVersionNumber.
 
-	(self hasFalsePositiveOf: name version: number) ifFalse: [ 
-		self installFalsePositiveOf: name version: number ].
+	(self hasFalsePositiveOf: ruleId version: number) ifFalse: [ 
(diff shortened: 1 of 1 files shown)


MarcusDenker (approved):
Ok!

PR #6252, opened 2020-04-28
Semantic Analysis: put Escape Analysis In one place

MarcusDenker:
This PR moves all code related to analysing and marking read and escape status of a variable to two methods:
#analyseEscapingRead: and #analyseEscapingWrite:

These are called from #lookupVariableForRead: and lookupVariableForWrite:


Ducasse:
It looks better to read.

PR #6255, opened 2020-04-29
remove OCKeyedSet

MarcusDenker:
- use a standard Dictionary for copied vars
- remove OCKeyedSet (private internal class)

the copiedVars2 name will be fixed in a second step


MarcusDenker:
one slave fails but that is not related to the change.

Ducasse:
I do not know this class and what it was used to do. 
Now reading the code I was wondering why the ivs are called copiedVars2 and not copiedVars.

MarcusDenker:
copiedVars2: you can not rename it just like that. I want to rename it by committing a PR that changes the dead code.

OCKeyedSet is a set that has the api of a dictionary. Very odd, it was supposed to be remove a long time ago.

MarcusDenker:
I merge it and then do the copiedVars2 rename in a second step

PR #6269, opened 2020-04-30
TextTest-trivial-clean

MarcusDenker:
- simplify TextTest examples
- simplify MetacellloVersionValidator>>#visited


MarcusDenker:
hmm.. image is in bad state

PR #6274, opened 2020-05-01
Small cleanups in BorderedMorph

### LINKED ISSUE #6273: Small cleanups in BorderedMorph
- #borderInitialize should be better named "initializeBorder" and put in a "private - initialization"
  protocol

additionally
- defaultBorderColor and defaultBorderWidth should be in a "defaults" category
- comment in exampleGradient is not necessary as we have an icon for the example to click
- exampleGradient should use "self" instead of class name
- exampleGradient  could use cascading

astares:
Fix #6273


Ducasse:
looks good to me.

PR #6279, opened 2020-05-01
Better TestExecutionEnvironment

dionisiydk:
PR slightly improves the implementation and logic of TestExecutionEnvironment regarding the monitoring of processes forked during tests. 

It extracts the process management functionality into separate class ProcessMonitorTestService which can be  disabled independently from the execution environment.

See ProcessMonitorTestService comment for more details.


src/SUnit-Core/ProcessMonitorTestService.class.st
```diff
@@ -0,0 +1,459 @@
+"
+I monitor all processes forked during tests and track all exceptions signaled from them.
```
dionisiydk, line 2: Here the main description for process monitoring logic

src/Kernel-Tests-Extended/ProcessTest.class.st
```diff
@@ -182,11 +182,51 @@ ProcessTest >> testInjectingMultipleExceptionHandlersIntoRunningProcess [
 	self assert: lastHandler
 ]
 
+{ #category : #tests }
+ProcessTest >> testIsNotSuspendedWhenItIsActiveProcess [
```
dionisiydk, line 186: Hm, it's strange. The fix for isSuspended was already integrated here #6197

dionisiydk:
I will reset garbage commits

PR #6280, opened 2020-05-01
Better TestExecutionEnvironment

dionisiydk:
PR slightly improves the implementation and logic of TestExecutionEnvironment regarding the monitoring of processes forked during tests.

It extracts the process management functionality into separate class ProcessMonitorTestService which can be disabled independently from the execution environment.

See ProcessMonitorTestService comment for full details. Here a little summary on functional improvements:
- in interactive mode (under debugger) system returns full control to the user. 
  - no processes are terminated when test is complete (proceed or close debugger)
  - any new errors are normally passed and new debuggers are shown like in normal execution 
- ProcessMonitorTestService can be disabled independently from full TestExecutionEnvironment (keep the test timeout but disable any process logic)
  - there is global settings (System/SUnit)
  - disable/enable in test (or setUp):
```Smalltalk
   self executionProcessMonitor disable
   self executionProcessMonitor enable
```
- Configurable process cleanup (termination) at the end of test
  - there is global settings (System/SUnit/ProcessMonitotTestService)
  - disable/enable in test (or setUp):
```Smalltalk
   self executionProcessMonitor disableProcessesTermination
   self executionProcessMonitor terminateProcessesAfterTest
```
- New logic to fail the test which left running processes (disabled by default for now)
  - new exception TestLeftRunnningProcesses is signaled at the end of test
  - there is global settings (System/SUnit/ProcessMonitotTestService)
  - disable/enable in test (or setUp):
```Smalltalk
   self executionProcessMonitor failTestLeavingProcesses
   self executionProcessMonitor allowTestToLeaveProcesses
```
- TestExecutionEnvironment contains collection of registered services (TestExecutionService)
   - all services are notified about various events during test execution:
      - signaled exception
      -  new process
      - completed test  
  - process monitoring is implemented as such service (ProcessMonitorTestService)
  - services can be enabled/disabled globally by default (or in settings) or they can be enabled/disabled manually during tests:
```Smalltalk
self executionEnvironment enableService: YourServiceClass.
self executionEnvironment disableService: YourServiceClass
```
- handy methods are added to TestCase to access current execution environment and process monitor:
```Smalltalk
aTestCase executionEnvironment.
aTestcase executionProcessMonitor.
```


src/SUnit-Core/ProcessMonitorTestService.class.st
```diff
@@ -0,0 +1,459 @@
+"
+I monitor all processes forked during tests and track all exceptions signaled from them.
+
```
dionisiydk, line 3: Here is a main description of process monitoring during tests

dionisiydk:
Tests are related. I will check. Maybe I did bad cleanup of my commits

Ducasse:
Thanks denis. 
This is important that we control flaky tests and this is also why we want to introduce TaskIt to avoid to have dangling process everywhere.

dionisiydk:
Now the PR is ready to merge.

I skipped tests which discovered the issue with Process completion in bootstrap image. The separate PR #6745 records this bug and allow further investigation. It proves that current PR changes are not responsible for this problem.

MarcusDenker (approved):
ok

PR #6283, opened 2020-05-02
Morphics #closestPointTo: has no senders - should be deprecated

### LINKED ISSUE #6282: Morphics #closestPointTo: has no senders - should be deprecated
There are three methods called #closestPointTo: in 

- BorderedMorph
- EllipseMorph
- PolygonMorph

that have NO SENDERS. 

I think we should remove them by deprecating them first.

astares:
- deprecate all #closestPointTo: methods
- fix #6282

### DIFF
--- a/src/Deprecated90/BorderedMorph.extension.st
+++ b/src/Deprecated90/BorderedMorph.extension.st
@@ -1,0 +1,15 @@
+Extension { #name : #BorderedMorph }
+
+{ #category : #'*Deprecated90' }
+BorderedMorph >> closestPointTo: aPoint [
+	"account for round corners. Still has a couple of glitches at upper left and right corners"
+	| pt |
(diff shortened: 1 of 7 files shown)


Ducasse:
I was thinking about it such methods are API methods (and of course without tests) and I would not deprecate it. I thought that I added a conversation about it but I probably forgot to press the button.

bencoman:
I feel this falls under the wide feature set of Morphic.  It seems to offer some value with three implementations that would be very useful to anyone user that needed it - but then also I can't say I've used it, although I could imagine using it in particular projects.   

I think there is more efficiency in deprecating whole subsystems that have been replaced by new functionality rather than deprecating individual methods.  On the other hand, we shouldn't discourage tiny incremental cleanups, so I'm on the fence (sorry in the thats no help, but may flesh out other thoughts - but then also bikeshedding this is not efficient. :)

Ducasse:
Let us keep it in the API

PR #6285, opened 2020-05-02
PolygonMorph #closestSegmentTo: should be deprecated

### LINKED ISSUE #6284: PolygonMorph #closestSegmentTo: should be deprecated
Method has NO SENDERS and is Morphic legacy. I propose to remove it by deprecating it

astares:
Fix #6284

- deprecation message added
- moved to deprecation package


Ducasse:
We keep it in the API

PR #6294, opened 2020-05-02
6293 opal ast mapping cache produces errors and freezes image when creating accessors method from the debugger after a dnu

### LINKED ISSUE #6293: Opal AST mapping cache produces errors and freezes image when creating accessors method from the debugger after a DNU
**Describe the bug**
The bytecode->AST cache is producing an error when creating an accessor from a DNU.

**To Reproduce**
Create this class:
```Smalltalk
Object subclass: #Toto
  instanceVariableNames: 'field'
  classVariableNames: ''
  package: 'Toto'.
```
Do this in a playground: `Toto new field`.

=> DNU => create method, select any protocol => boom.

**Version information:**
All versions from the integration of https://github.com/pharo-project/pharo/pull/6242

**Additional context**
This is because the mapping fails to recognize the bytecode for primitives in quick methods returning an inst var.

StevenCostiou:
Fixes #6293 
Quick methods primitive bytecodes were not taken into account in the ast mapping cache.
Also improved and fixed tests.


Ducasse:
I do not get why the builds are not run.

Ducasse:
Ok now it was building

PR #6304, opened 2020-05-04
Rules checking Tests should be in package SUnit-Rules

### LINKED ISSUE #6303: Rules checking Tests should be in package SUnit-Rules
1. Classes like
    - ReTestClassNameShouldEndWithTestRule
    - ReTestClassNameShouldNotEndWithTests
    - ReTestClassNotInPackageWithTestEndingNameRule
   who are checking unit tests should be moved to "SUnit-Rules" package instead of 
   general Renraku package.

2. These rules should also not be in "Optimizations" group of the Linter but in "SUnit" group.
    We can share a 

3. Classes like ShouldSendSuperSetUpAsFirstMessage and ShouldSendSuperTearDownAsLastMessage should also be prefixed with "Re" for rules engine

4. also move ReTestBasedTestCase and subclasses into a new package "SUnit-Rules-Tests"

5. Make ReTestBasedTestCase an abstract class (it has no tests) by implementing #isAbstract

astares:
- move ReTestClassNameShouldEndWithTestRule into "SUnit-Rules" package
- move ReTestClassNameShouldNotEndWithTests into "SUnit-Rules" package
- move ReTestClassNotInPackageWithTestEndingNameRule into "SUnit-Rules" package
- add a trait ReTSUnitGroupedRule to have a single implementation of #group and use this traiit
- give ShouldSendSuperSetUpAsFirstMessage also the rules engine "Re" prefix
- give ReShouldTransformedIntoAssertRule also the rules engine "Re" prefix

Fix #6303

### DIFF
--- a/src/ReleaseTests/ProperlyImplementedSUnitClassesTest.class.st
+++ b/src/ReleaseTests/ProperlyImplementedSUnitClassesTest.class.st
@@ -18,7 +18,7 @@ ProperlyImplementedSUnitClassesTest >> testAndMakeSureSuperSetupIsCalledAsFirstM
 			each
 				compiledMethodAt: #setUp
 				ifPresent: [ :method | 
-					(ShouldSendSuperSetUpAsFirstMessage superSetUpNotCalledFirstIn: method)
+					(ReShouldSendSuperSetUpAsFirstMessage superSetUpNotCalledFirstIn: method)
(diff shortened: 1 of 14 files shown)


MarcusDenker (approved):
looks good. Traits use might be something someone could not like, but we need to actually use them, so I think it's ok.

PR #6305, opened 2020-05-05
fix-keyUp-keyDown-eventHandler

estebanlm:
Morph is missing correct handling for keyUp/keyDown events with the morphic event handler.
This PR fixes that problem :)


estebanlm:
wait a bit, I just discovered there are more cases

estebanlm:
ok, done (I think :) )

jecisc (approved):
It reverts a chang I did before but the regression test I introduced is still green. It might be caused by another improvement made. 

This seems ok to me :)

PR #6312, opened 2020-05-05
Remove wrong detection of a lost change after save, wait and quit

### LINKED ISSUE #5308: Epicea: Do not consider "snapshot done" as a lost change
Steps to reproduce:
1. open an image
2. do a code change (e.g. create a class)
3. save image (DO NOT QUIT)
4. save image AND QUIT
5. open image again and see a dialog window proposing the recovery of lost changes (but no change is lost)

tinchodias:
Log EpSnapshotDone in shutDown: instead of SystemAnnouncer's SnapshotDone.

The problem was that the EpMonitor, as a subscriber to SnapshotDone, was notified after the image is dumped to disk and so adding a new EpSnapshotDone to the log.

Fixes #5308

### DIFF
--- a/src/Epicea/EpMonitor.class.st
+++ b/src/Epicea/EpMonitor.class.st
@@ -51,6 +51,11 @@ EpMonitor class >> hasCurrent [
 	^ current isNotNil
 ]
 
+{ #category : #'class initialization' }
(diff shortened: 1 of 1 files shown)


Ducasse:
Thanks martin.

PR #6314, opened 2020-05-06
optionalFindThePig

pavel-krivanek:
The CPU Watcher can be dangerous because may suspend processes that it expects to be too much CPU demaning but they may be needed for proper image function. Make this "pig finding" optoinal and make a setting for it


MarcusDenker (approved):
ok

PR #6315, opened 2020-05-06
DebugSession: removing "keep-alive" dead-code 

StevenCostiou:
This code was inserted for compatibility with tools that are not in the image, and is not going to be put in it in the future.

In addition, this code provoked bugs while running tests, and in some cases prevents the tear down method to execute when a debugger is opened then closed during a test (e.g. if failure).

### DIFF
--- a/src/Debugger-Model/DebugSession.class.st
+++ b/src/Debugger-Model/DebugSession.class.st
@@ -26,9 +26,7 @@ Class {
 		'name',
 		'interruptedContext',
 		'interruptedProcess',
-		'errorWasInUIProcess',
-		'objectsThatWantToKeepMeAlive',
-		'subscribersToKeepAliveEvents'
+		'errorWasInUIProcess'
 	],
 	#classVars : [
 		'LogDebuggerStackToFile'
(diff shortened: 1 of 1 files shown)


StevenCostiou:
@estebanlm this might have an impact to you. Are you calling `#stopKeepingAlive:` or similar in gtk? (all those calls should be removed after this is merged)

StevenCostiou:
It is also the source of this bug: after closing debuggers processes are never garbage collected and it ends up in severe ui glitches. In the end one is forced to kill all processes manually and refresh the image.
Please consider this PR @Ducasse @estebanlm
[EDIT] actually it is the usage of that code by debuggers in NewTools that, it seems, provokes the processes bug. After removing all calls to that I do not have the problem again.

Screenshot 2020-05-12 at 22 12 37

cdlm (approved):
Seems fine… but I don't understand how to review that in the image without Iceberg completely confusing me about what's actually loaded 🤷‍♂️

MarcusDenker (approved):
ok

PR #6317, opened 2020-05-07
SUnitDependencies

pavel-krivanek:
cleanup external dependencies


Ducasse:
Hello Pavel 
I got completely confused by this PR. I could not really understand the visiting reflectivity stuff. 
S

pavel-krivanek:
these methods make dependency of `Reflectivity` on `AST-Core-Traits`. We do not really need `AST-Core-Traits` soon, it is required only for `Shout`. So I moved these simple extension methods directly to the  `AST-Core-Traits` package.

PR #6322, opened 2020-05-08
Check matches of ReGlobalVariablesUsageRule against list of known globals

### LINKED ISSUE #6321: Ignore known globals when checking RBGlobalVariablesUsageRule
**Describe the request**
RBGlobalVariablesUsageRule treats known globals like `Smalltalk` and `Processor` as normal global variables. While `Smalltalk` could be replaced with `SmalltalkImage current` (which actually uses the global variable `Smalltalk`), there is no replacement for `Processor`.

I suggest to treat the following globals as known: 
- ActiveEvent
- ActiveHand
- ActiveWorld
- Display
- Processor
- Sensor
- Smalltalk
- SourceFiles
- SystemOrganization
- Transcript
- Undeclared
- World

**Expected behavior**
RBGlobalVariablesUsageRule should ignore matches of known globals.

**Expected development cost**
10 Minutes

theseion:
Fixes #6321


MarcusDenker:
ReGlobalVariablesUsageRuleTest has failing tests that seem related

theseion:
Thanks Marcus. I've updated the test case. The CI checks appear to be fine now.

MarcusDenker:
Hmm... CI still looks strange. I close and re-open to merge with latest

MarcusDenker:
There are still tests failing in ReGlobalVariableRuleTest

MarcusDenker:
ReGlobalVariableRuleTest can just be removed 
- ReGlobalVariablesUsageRuleTest test the rule already
- the test here test just what you changed

theseion:
Thanks Marcus, I'll look at it right now.

theseion:
Looks good to me. Failures are all related to Zinc.

MarcusDenker:
I still think it is good to merge this. yes, we need rules for most of these Globals that tells users to not use them. 

For now, the rule as it is in the image tells me multiple times per day that I am not allowed to use "Smalltalk image" or "Smalltalk compiler" and that makes me mad.

MarcusDenker:
and because we waited, of course there is now a conflict

theseion:
I'll update the PR to merge into the latest changes.

theseion:
Merged with tip of Pharo9.0. I hope this does the trick.

theseion:
Good to go.

PR #6324, opened 2020-05-08
merge-spec2-0-6

estebanlm:
New Features:
===

General
---
- #eventHandler low level protocol added to presenters. Now you can directly listen: focusReceived, focusLost, keyUp, keyDown, mouseUp, mouseDown, mouseEnter, mouseLeave and mouseMove.
- <meta+t> will show the context menu of presenters, if defined.
- styles of presenters can be added and removed in runtime.
- added #defaultKeyboardFocus to complement focus state (when I want to transmit the default focus to one of my children instead of the first one... but I do not want to rewrite the full #focusOrder)
- To all text presenters (text area, text field, code area, etc): autoAccept property was removed, all text is by default "autoAccept" now.
	- whenSubmitDo:/whenResetDo: mapping <meta+s> and <meta+l> events were added to replace the usage of acceptBlock:, etc.. 
- added a general mechanism to allow i18n.

Presenters
---
- SpTextPresenter
	- add "editable" property to use when I need to show text but not modifiable.
	- adds natural width/height propagation concept (to auto-expand text areas when needed, instead following scrollbar size).
	- cursorPosition answers row@column position (adds to cursorPositionIndex which answers position from start of text).
- SpCodePresenter
	- now can add user defined contextMenu in addition to code default menu (you can still override it completely if you want).
	- Now commands work properly (it was not the case).
	- It has a new #interactiveModel api to give the presenter an object who will handle all Pharo related stuff.
- SpMenuItemPresenter
	- name and icon now are modifiable on runtime.
- SpToolBarPresenter
	- added SpToolBarToggleButtonPresenter
	- added SpToolBarMenuButtonPresenter
- SpModalWindowPresenter
	- close on backdrop click is not false by default (the regular desktop behaviour)

Layouts
---
- SpBoxLayout now has alignment properties (hAlignCenter, hAlignStart, hAlignEnd, vAlignCenter, vAlignStart, vAlignEnd) that will handle how the contained presenters will be aligned in the layout.

New presenters:
===

- SpPopoverPresenter, to show popovers relative to a presenter (or located elsewhere in screen).
- SpWorldPresenter (Morphic specific) restored and working properly (this is used to make a presenter to "take over the world". 

Fixes: 
===

- Toolbar now does not has a weird effect when hovering
- MenuItems now display shortcut properly
- Number presenter answers zero to value when content is empty.
- Style now correctly calculates properties (there was a bug where nested defining properties where replacing higher level ones, instead merging them).
- Scrolling and selection now decoupled (you can select without scrolling, which was not possible)

(And a lot more tweaks and fixes that makes this version a lot more stable)


estebanlm:
wow, how is this even possible?

Ducasse:
Esteban the EF UI packages should be unloaded from the image. The code never worked. 
So the UI Tests are not working and this is normal too.

estebanlm:
mmm... so what I do? We wait?

estebanlm:
ok, errors not related :)
However, @MarcusDenker... I'm getting a constant failure on the reported tests. Do we have an idea what is happening?

MarcusDenker:
I think they are network problems of the slave. And a file system oddity.

It does not look like problems introduces by the PR

Ducasse:
So esteban we can integrate no?

PR #6328, opened 2020-05-09
Refactor monstrous Process>>#terminate method

dionisiydk:
Split the logic on:
- termination from the process itself 
- termination from another processes

### DIFF
--- a/src/Kernel/Process.class.st
+++ b/src/Kernel/Process.class.st
@@ -314,8 +314,7 @@ Process >> isTerminated [
 
 { #category : #testing }
 Process >> isTerminating [
-	"lazy initialization is a fallback only for processes that existed before this addition"
-	^ terminating ifNil: [ false ]
+	^ terminating
 ]
 
 { #category : #printing }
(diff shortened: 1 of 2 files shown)


src/Kernel/Process.class.st
```diff
@@ -678,51 +677,58 @@ Process >> suspendingList [
...
-			of the suspendedContext to nil (#isDead ---> true)."
-			(suspendedContext notNil and: [
-				suspendedContext isBottomContext ]) ifTrue: [
-					suspendedContext pc: nil ] ] ]
+		ifTrue: [ self terminateFromYourself ]
+		ifFalse: [ self terminateFromAnotherProcess]
+]
+
+{ #category : #'changing process state' }
+Process >> terminateFromAnotherProcess [
```
dionisiydk: Here it's a code copied from the original #terminate method. It's still huge but it is a work for another time

dionisiydk:
Hm. The fails bootstrap at 4-build.sh with infinite recursion over:
```Smalltalk
CompiledMethod(Object)>>doesNotUnderstand: #5879 
MessageNotUnderstood: CompiledMethod>> #reflectiveMethod
```
I doubt that my changes are broke because locally all works fine. And ProcessTest and other related tests are green. I will try to do this changes in smaller steps.

@guillep Do you have idea how splitting #terminate method (with simple extract method refactoring) could break bootstrap?

dionisiydk:
Ok. I see the issue:
```Smalltalk

UndefinedObject>>doesNotUnderstand: #mustBeBooleanDeOptimizeIn:
UndefinedObject(ProtoObject)>>mustBeBoolean
Process>>terminateFromYourself
```
I changed #isTerminating to not be lazy. There was a comment:
```Smalltalk
"protocol: testing"
isTerminating
	"lazy initialization is a fallback only for processes that existed before this addition"
	^ terminating ifNil: [ false ]
```
I thought that it was needed in the very old time. Now with bootstrap we should not have old process instances but bootstrap failed due to nil #terminating. How is it possible?
There is #initialize method: 
```Smalltalk
Process>>initialize
	super initialize.
	
	terminating := false
```

dionisiydk:
CI looks dead. Last error is segmentation fault on Windows 64

dionisiydk:
I checked another time that new methods are correctly extracted from original #terminate implementation. 
Please merge it.

PR #6336, opened 2020-05-11
SimplifyOCASTTranslater-by-tempVarNamesWithoutArguments

MarcusDenker:
This PR adds #tempVarNamesWithoutArguments on the semantic scope for methods and blocks. This then allows us to simplify some code on the OCASTTranslator

In addition, make sure to use #tempVectorVarNames

### DIFF
--- a/src/OpalCompiler-Core/OCAbstractMethodScope.class.st
+++ b/src/OpalCompiler-Core/OCAbstractMethodScope.class.st
@@ -251,7 +251,7 @@ OCAbstractMethodScope >> tempVarNames [
 { #category : #'temp vars' }
 OCAbstractMethodScope >> tempVarNamesWithoutArguments [
 
-	^ tempVars values select: [ :each | each isArg not ] thenCollect: [ :each | each name ]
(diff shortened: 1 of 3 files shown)


MarcusDenker:
This is now open longer than a week.

src/OpalCompiler-Core/OCAbstractMethodScope.class.st
```diff
@@ -248,6 +248,12 @@ OCAbstractMethodScope >> tempVarNames [
 	^ tempVars keys
 ]
 
+{ #category : #'temp vars' }
+OCAbstractMethodScope >> tempVarNamesWithoutArguments [
+
+	^ tempVars values select: [ :each | each isArg not ] thenCollect: [ :each | each name ]
```
jecisc:
```suggestion
	^ tempVars values reject: [ :each | each isArg ] thenCollect: [ :each | each name ]
```

jecisc (commented):
I commented with a little suggestion (nothing serious but new version seems more readable)

PR #6339, opened 2020-05-11
[DO NOT MERGE] Adding livetyping

tesonep:
This PR is to be able to generate the images needed to try the livetyping extensions of Hernan Wilkinson in Pharo, and to develop new tools around the collected information


Ducasse:
Pablo can we close this PR?

PR #6343, opened 2020-05-12
6342-SubscriptoutOfBounds-while-unzipping-with-ZipArchive

bocasti:
Hack the #uncompressDataTo: to bypass SubscriptOutOfBound error

### DIFF
--- a/src/Compression/ZipFileMember.class.st
+++ b/src/Compression/ZipFileMember.class.st
@@ -193,7 +193,7 @@ ZipFileMember >> uncompressDataTo: aStream [
 		| chunkSize |
 		chunkSize := 32768 min: readDataRemaining.
 		buffer := decoder next: chunkSize into: buffer startingAt: 1.
-		aStream next: chunkSize putAll: buffer startingAt: 1.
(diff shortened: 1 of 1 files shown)


src/Compression/ZipFileMember.class.st
```diff
@@ -193,7 +193,7 @@ ZipFileMember >> uncompressDataTo: aStream [
 		| chunkSize |
 		chunkSize := 32768 min: readDataRemaining.
 		buffer := decoder next: chunkSize into: buffer startingAt: 1.
-		aStream next: chunkSize putAll: buffer startingAt: 1.
+		aStream next: (chunkSize min: buffer size) putAll: buffer startingAt: 1.
```
bocasti, line 196: In the case I experimented, I ended up with chunkSize of 32768 and buffer size of 20242
This lead to a SubscriptOutOfBounds error.
This is a very naive hack, but maybe someone with more knowledge may know how much sense does make that change

Ducasse:
Thanks for this fix. I will rerun the tests.

PR #6344, opened 2020-05-13
enable-full-block-closures

tesonep:
Adding the conversion to Sista Bytecode including Full Block Closures


Ducasse:
Pablo why do we need the recompiledMethod version? just out of curiousity.

tesonep:
Basically the implementation of recompile: and the version in the decompiler were using the most low-level API to add a method in the class. This API goes and directly modify the method dictionary. 
The recompile: method was not working with Traits, as the change in the trait was not propagated to its users. 
The traits need to have a hook to propagate the changes. Maybe, we can do it in a another way. Just I don't get to that.

I have found it, through a test of Calypso that checks to look for users of a trait and checking it has found the methods. And it was founded the method because they were different. One was compiled with V3 and the other with Sista. 

This is done because all the bootstrap is done in V3 and then all the methods are recompiled in Sista + Full block closures.

Ducasse:
thanks for the explanation!

MarcusDenker (approved):
Ok!

guillep (approved):

PR #6347, opened 2020-05-13
Compiler: #notify:at: Cleanup

MarcusDenker:
- #notifty:at: is just used in OCSemanticError Hierarchy, remove it from OCSemanticWarning
- OCReadOnlyVMWarning is not needed anymore: all VMs support read only objects
- add #requestor to OCSemanticError and simplify notify:at: there


MarcusDenker:

PR #6348, opened 2020-05-13
DeprecateObject-notify-at

MarcusDenker:
Object>>#notify:at: just calles #notify. It is not send anywhere and makes no sense. 

The PR adds a transforming deprecation


MarcusDenker:

dionisiydk (approved):

PR #6349, opened 2020-05-13
Cleanup: OCAbortCompilation

MarcusDenker:
OCAbortCompilation was just an error subclass without any additional behavior. In addition, it is raised after a nil check at a place where the value is not nil (it might be an empty string). 

I tested the SyntaxErrorDebugger and I could never trigger the nil case.

This PR removes the nil check as it seems to be not needed and removes OCAbortCompilation.


MarcusDenker:

dionisiydk:
I am not sure that it is a right cleanup:

SyntaxErrorNotificaton is processed through current UIManager. There are many subclasses. At least DummyUIManager looks suspicious. It does #return at the end. Will it get the ifNil condition ?.
Or the notification can be intercepted by user code (on:do:) where it can be resumed with nil. 
I guess in that case the compilation should not continue and the abort fallback is a good approach.

MarcusDenker:
Ok, I close this.

PR #6352, opened 2020-05-13
6351-SyntaxErrorDebugger-messes-up-display-state-when-closed

### LINKED ISSUE #6351: SyntaxErrorDebugger messes up display state when closed
SyntaxErrorDebugger, if we close it, messes up the state of the morphic rendering.

The reason is that it suspends the morphic process and creates a debugger, but when closing it does not terminate the debugger so the rendering is left in a bad state.

It used to be that #release was called to do that, but it is not anymore.

MarcusDenker:
SyntaxErrorDebugger, if we close it, messes up the state of the morphic rendering.

The reason is that it suspends the morphic process and creates a debugger, but when closing it does not terminate the debugger so the rendering is left in a bad state.

It used to be that #release was called to do that, but it is not anymore.

Fix: call it in #closeWindow.

In addition, this PR contains two trivial cleanups. fixes #6351

### DIFF
--- a/src/Tools/SyntaxErrorDebugger.class.st
+++ b/src/Tools/SyntaxErrorDebugger.class.st
@@ -126,7 +126,7 @@ SyntaxErrorDebugger >> checkForUnprintableCharacters: aString [
 
 { #category : #other }
 SyntaxErrorDebugger >> closeWindow [
-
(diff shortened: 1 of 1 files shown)


MarcusDenker:

dionisiydk (approved):

PR #6354, opened 2020-05-14
Variable breakpoints: break on variable (inst/temps/args) accesses (read/write/both)

StevenCostiou:
- Implemented and added breakpoints for variables accesses (read, write, both)
- Fixed bugs in Reflectivity + tests
- Improved a bit Breakpoint + tests
- Added meta-information (context + executed node) to Break signals
- Added a simple observer pattern specific to breakpoints to notify of breakpoint (un)installation, and hits

Note that for now a variable breakpoint breaks **after** the variable has been accessed. It means, for a write, that when it breaks it is after the new value has been written into the variable. This will become customizable in the future.

VariableBreakpoints


svenvc:
This looks great ! And I want it.

MarcusDenker:
some load order problem:


MetacelloNotification: Loaded -> Reflectivity-Examples-tonel.1 --- tonel:///builds/workspace/uest_and_branch_Pipeline_PR-6354/64/src --- cache
[64] NewUndeclaredWarning: VariableBreakpointTest>>setUp (DummyBreakpointObserver is Undeclared)
[64] MetacelloNotification: Loaded -> Reflectivity-VariableBreakpoint-tonel.1 --- tonel:///builds/workspace/uest_and_branch_Pipeline_PR-6354/64/src --- cache
[64]   SycSingleClassCommand
[64]   ClyDebuggingCommand
[64]   SycVariableCommand

MarcusDenker (commented):
Somehow it does not load on the CI due to missing classes

src/Reflectivity-VariableBreakpoint-Extensions/ClyBreakOnVarAccessCommand.class.st
```diff
@@ -0,0 +1,25 @@
+"
+I break on all accesses to selected target variable in the source code
+"
+Class {
+	#name : #ClyBreakOnVarAccessCommand,
```
dionisiydk: This is very confusing names: ClyBreakOnAccessCommand vs ClyBreakOnVarAccessCommand. And same for subclasses and superclasses.
What the difference between them? 
I am sure we can avoid these parallel hierarchies.

src/Reflectivity-VariableBreakpoint-Extensions/ClyBreakOnVarAccessCommand.class.st
```diff
@@ -0,0 +1,25 @@
...
+	#superclass : #ClyVariableBreakpointCommand,
+	#category : #'Reflectivity-VariableBreakpoint-Extensions'
+}
+
+{ #category : #testing }
+ClyBreakOnVarAccessCommand class >> canBeExecutedInContext: aSourceCodeContext [
+	^ (super canBeExecutedInContext: aSourceCodeContext)
+		and: [ aSourceCodeContext selectedSourceNode isArgument not ]
```
dionisiydk: Why it does not work for arguments?

src/Reflectivity-VariableBreakpoint-Extensions/ClyBreakOnVarAccessCommand.class.st
dionisiydk: I did not look deeply in the code. But my first idea to solve this is to ask variables to install breakpoints instead of doing it directly in the command

src/Reflectivity-VariableBreakpoint-Extensions/ClyBreakOnAccessCommand.class.st
```diff
@@ -0,0 +1,20 @@
...
+	#name : #ClyBreakOnAccessCommand,
+	#superclass : #ClyBreakOnVariableCommand,
+	#category : #'Reflectivity-VariableBreakpoint-Extensions'
+}
+
+{ #category : #execution }
+ClyBreakOnAccessCommand >> breakOn: variable [
+	VariableBreakpoint
+		watchVariable: variable name
+		in: variable definingClass
```
dionisiydk: Here it can be simple 
```Smalltalk
variable breakOnAccess
```

dionisiydk:
But I should start review with the words that it is super cool features :)

src/Reflectivity-VariableBreakpoint-Extensions/ClyBreakOnVarAccessCommand.class.st
StevenCostiou: It is because of how I define "access", for me it is write + read, and as we cannot write in args, then only the read option is available. 
It could be changed, by I feel it is not worth the effort.

src/Reflectivity-VariableBreakpoint-Extensions/ClyBreakOnVarAccessCommand.class.st
dionisiydk: For example check other SycVariableCommand's. All of them are available in variables list and in source code (when variable name is selected).

StevenCostiou: The problem (to me) is that it is very hard to understand and build commands with all that subclassing. Maybe I got something wrong and it can be simplified.

Difference is:
- ClyBreakOnAccessCommand => takes a class (or an object) and break on *all* variable acccesses
- ClyBreakOnVarAccessCommand => takes a particular variable

Also if I understood right, it is needed to have three hierarchies because they don't apply to the same contexts. I found no way of having the menus right in those three places (source code, on the class and on the variable view).

src/Reflectivity-VariableBreakpoint-Extensions/ClyBreakOnAccessCommand.class.st
StevenCostiou: It would work, but it also needs an object-centric API and all the variations read/write/access. This needs some brainstorm - maybe it's simple in the end but I do not wish to do it now.

I suggest to leave it as is and work on this problem separately through another PR.

StevenCostiou:
> But I should start review with the words that it is super cool features :)

Thanks :)

src/Reflectivity-VariableBreakpoint-Extensions/ClyBreakOnVarAccessCommand.class.st
dionisiydk: I see. I think it deserves the comment.

From the other side it can be confusing to the users of UI:
For the most cases there will be a "break on access". Then when occasionally I will select a variable (without thinking that it is an argument) that would be quite unexpected to not find this command. 
I would really keep this command for all variables even if "break on read" would do the same.

StevenCostiou: But then users will be confused because they won't understand the difference between "access" and "read". I feel we can always make an unperfect choice, and users will have to learn.
But the point is good, so what could be done:
- add in the name of the menu item "arg" or "inst" or "temp" to be explicit
- mention the subtility in the popover help

dionisiydk: > But then users will be confused because they won't understand the difference between "access" and "read".

Personally I do not see the confusion here. Users need to understand the difference between these two commands independently from the argument type.
The fact that argument variables are not writable anywhere in the method should not break this understanding IMHO. Meaning of "access" word here is not changed (from the user point of view).

dionisiydk: Also arguments can be writable. There is a settings in the compiler. it can be useful for code migration from old versions of Pharo or Squeak

StevenCostiou:
> some load order problem:
> 
> MetacelloNotification: Loaded -> Reflectivity-Examples-tonel.1 --- tonel:///builds/workspace/uest_and_branch_Pipeline_PR-6354/64/src --- cache
> [64] NewUndeclaredWarning: VariableBreakpointTest>>setUp (DummyBreakpointObserver is Undeclared)
> [64] MetacelloNotification: Loaded -> Reflectivity-VariableBreakpoint-tonel.1 --- tonel:///builds/workspace/uest_and_branch_Pipeline_PR-6354/64/src --- cache
> [64] Error: This package depends on the following classes:
> [64] SycSingleClassCommand
> [64] ClyDebuggingCommand
> [64] SycVariableCommand

So it is because it depends on `SystemCommands`. I do not know how to solve this.
I added my tool in two packages in Reflectivity baseline. First package is ok, it only depends on RF. Second contains the calypso commands, but then in which baseline should I add this package?

src/Reflectivity-VariableBreakpoint-Extensions/ClyVariableBreakpointCommand.class.st
```diff
@@ -0,0 +1,48 @@
...
+Class {
+	#name : #ClyVariableBreakpointCommand,
+	#superclass : #ClyDebuggingCommand,
+	#category : #'Reflectivity-VariableBreakpoint-Extensions'
+}
+
+{ #category : #testing }
+ClyVariableBreakpointCommand class >> canBeExecutedInContext: aSourceCodeContext [
+	^ self isAbstract not
+		and: [ aSourceCodeContext class = ClyMethodSourceCodeContext
```
dionisiydk: What you are trying to disable here?
Why it is not suitable for any source code context?

StevenCostiou:
> > some load order problem:
> > MetacelloNotification: Loaded -> Reflectivity-Examples-tonel.1 --- tonel:///builds/workspace/uest_and_branch_Pipeline_PR-6354/64/src --- cache
> > [64] NewUndeclaredWarning: VariableBreakpointTest>>setUp (DummyBreakpointObserver is Undeclared)
> > [64] MetacelloNotification: Loaded -> Reflectivity-VariableBreakpoint-tonel.1 --- tonel:///builds/workspace/uest_and_branch_Pipeline_PR-6354/64/src --- cache
> > [64] Error: This package depends on the following classes:
> > [64] SycSingleClassCommand
> > [64] ClyDebuggingCommand
> > [64] SycVariableCommand
> 
> So it is because it depends on `SystemCommands`. I do not know how to solve this.
> I added my tool in two packages in Reflectivity baseline. First package is ok, it only depends on RF. Second contains the calypso commands, but then in which baseline should I add this package?

Ok it seems my classes should actually be in `Calypso-SystemPlugins-Reflectivity-Browser`

src/Reflectivity-VariableBreakpoint-Extensions/ClyVariableBreakpointCommand.class.st
StevenCostiou: I will check again, but to me it was buggy: the source code context are not always consistent.

StevenCostiou: Yes, so because sometimes it is a `ClyMethodContextOfFullBrowser` (when clicking on a variable in the var view). I really do not understand it, how to use commands in Calypso is cryptic. If I do not filter like this, it tries to call the source node on that context and it fails.

MarcusDenker:
Two tests are failing that seem related:

ProperlyImplementedSUnitClassesTest>>testAndMakeSureSuperSetupIsCalledAsFirstMessageInSetupMethodsOfTestCases

	-> an OrderedCollection(VariableBreakpointTest>>#setUp) should have been empty

and: 

ReleaseTest>>testManifestNamesAccordingToPackageNames

src/Reflectivity-VariableBreakpoint-Extensions/ClyVariableBreakpointCommand.class.st
dionisiydk: It looks like a bug. I will look when you merge

src/Reflectivity-VariableBreakpoint-Extensions/ClyBreakOnVarAccessCommand.class.st
dionisiydk: Hm. I see following subclasses:
- ClyBreakOnAccessCommand <-- ClyBreakOnVarCommand <-- SycVariableCommand - which are commands for the list of variables
- ClyBreakOnVarAccessCommand <-- ClyVariableBreakpointCommand <-- ClyDebuggingCommand <-- SycSourceCodeCommand - which are commands for the given astNode in given method

I have impression that PR/commit is corrupted. I found two files ClyVariableBreakpointCommand.class.st in different packages (not extension). Also there is  SycVariableBreakpointCommand. 
It looks like you did some repackaging and rename and then Iceberg commit became wrong.

StevenCostiou: Indeed I renamed a package. I see it also make other tests failing related to the manifest.
I will look at that tomorrow morning, this is strange :/

src/Reflectivity-VariableBreakpoint-Extensions/ClyBreakOnVarAccessCommand.class.st
StevenCostiou: Thanks, I'll give you a reproducible case. It is possible I am using it the wrong way.

src/Reflectivity-VariableBreakpoint-Extensions/ClyVariableBreakpointCommand.class.st
dionisiydk: Not a bug. You should use #lastSelectedSourceNode instead of #selectedSourceNode. But I understand the confusion. 
#selectedSourceNode is internal method of ClySourceNodeContext. But the general API to access objects from context always follow multiple selection idea. When command needs just one object it should be explicit first or last selected object. 
For example in method context you also can get lastSelectedSourceNode which is an ast of last selected method. And it is a reason why you see the method context here: it is compatible to source code commands.

dionisiydk: I think when you will fix it you will get another cool place for new commands - try it from class definition on variable names.

src/Reflectivity-VariableBreakpoint-Extensions/ClyBreakOnVarAccessCommand.class.st
StevenCostiou: For now I do not wish to spend more time on this, I think we should do an issue on this and brainstorm from there.
My choices can be discussed w/o any problem, I don't claim I make the best ones ;) but I will not be able to put more efforts right now and it will block the PR.

src/Reflectivity-VariableBreakpoint-Extensions/ClyVariableBreakpointCommand.class.st
StevenCostiou: It works :) So I removed the class check and use lastSelectedNode as you suggested.

src/Reflectivity-VariableBreakpoint-Extensions/ClyBreakOnVarAccessCommand.class.st
StevenCostiou: This is non-blocking, and I cannot afford more time right now to work on it. Could we do an issue to improve the command hierarchy and work from there so that it does not block the PR?

src/Reflectivity-VariableBreakpoint-Extensions/ClyBreakOnAccessCommand.class.st
StevenCostiou: The more I look at your suggestion, the more I find it better.
So I'll look at it but in the next weeks, it's not that quick (sometimes we call the API on a vaariable, sometimes on a class, sometimes on a node and a method...).
Again, I would prefer a PR to improve infrastructure so the PR can be integrated and we can take time to brainstorm without blocking the feature.

StevenCostiou:
There is still a failure but I don't understand it, and only on the macos build. Does anyone knows what it may be?
Failing tests are not mine nor related to the PR.

MarcusDenker:
The failing tests now are not related

src/Reflectivity-VariableBreakpoint-Extensions/ClyBreakOnVarAccessCommand.class.st
dionisiydk: Yes. I am always for integrate/improve cycle.

dionisiydk (approved):

PR #6358, opened 2020-05-15
FixCompilerSettings

MarcusDenker:
Fix the compiler settings in the Preference browser. Some code got lost when refactoring, restored and then simplified the code.


MarcusDenker:
It's just a setting. I will merge it

PR #6359, opened 2020-05-15
asDoitShouldStoreFormattedSource

MarcusDenker:
we have a check in generateWithSource. It is better to move that into  the #asDoit methods

### DIFF
--- a/src/OpalCompiler-Core/RBMethodNode.extension.st
+++ b/src/OpalCompiler-Core/RBMethodNode.extension.st
@@ -72,12 +72,6 @@ RBMethodNode >> generateIR [
 { #category : #'*OpalCompiler-Core' }
 RBMethodNode >> generateWithSource [
 	"Answer a CompiledMethod with source encoded in trailer."
-
-	
-	"for doits, we need to store the source pretty printed from the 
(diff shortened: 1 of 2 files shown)


guillep (approved):
Ok, just a small refactoring.

Question: is there any reason for using `formattedCode` and not just get the original source code? Reformatting the method systematically has a strange surprising effect at least on me.

MarcusDenker:
The problem is that the AST, source and bytecode need to be in sync for the debugger to work.

We do two transformations: accessing temps via the context (no other way as they might not be accessible via the temp vector / copied temps at that point). And we rewrite the last statement to a return.

ST80 did this by doing the special compile in the compiler and then rely on the decompiler to show the code. This way it is in sync, too.  But it looks the same ugly *and* even loses more information.

(But a nice aspect is that there the slow printing is done at decompile, not eval time).

I think the solution is: compile a block with the home context being the context where the doit is done. This needs some care implementation wise (for vars that are not accessed in the context we have the same problem that we need to access them via the context), but I think it is doable by introducing a special scope and a special kind of temp (that compiled to tempNamed:).

PR #6364, opened 2020-05-18
unload-enlumineur-ui

estebanlm:
enlumineur-ui and enlumineur-ui test removed because they do not work (and never worked)


estebanlm:
errors not related

estebanlm:
I need this to finish the Spec 0.6 merge.

PR #6365, opened 2020-05-18
6256-MacRomanUnicode-char-converter-methods-are-backwards

Rydier:
Reverse incorrect conversion order in these methods - document with test.


Ducasse:
Ohhhh thanks!!!

Ducasse:
Given the changes you did I do not get why some systemdependency tests are complaining. The new test does not introduce any dependency so I'm puzzled.

Ducasse:
Strange I will check within Pharo do see the dependencies because the test reports the same.

Ducasse:
Ok I integrate and we will see

PR #6371, opened 2020-05-19
CleanupOCKeyedSet2

MarcusDenker:
strangely, OCKeyedSet was not removed with the PR that removed all the users. 
- remove OCKeyedSet
- move #emitValue: from OCSpecialVariable to the self and super subclasses  and add  that "super" actually is self (it's the send that is different)


guillep (approved):
Good!

PR #6372, opened 2020-05-19
Improve Compiler Overlay

MarcusDenker:
Improve the compiler overlay code a little:

- Split the overlay creation into multiple steps, each has it's own method

- add a fifth step: existing instances should use the overlay classes, too.

### DIFF
--- a/src/OpalCompiler-Core/OpalCompiler.class.st
+++ b/src/OpalCompiler-Core/OpalCompiler.class.st
@@ -42,6 +42,17 @@ Class {
 	#category : #'OpalCompiler-Core-FrontEnd'
 }
 
+{ #category : #overlay }
+OpalCompiler class >> compilerClasses [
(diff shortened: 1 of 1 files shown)


guillep (approved):
Good, this should have no impact on non-users.

@MarcusDenker I wonder wether in the future this cannot extract this into a "PackageOverlay" .
Besides the method `overlayStep4SetImageCompiler` that is compiler dependent, all the rest of the methods should work with other packages.

MarcusDenker:
Yes, I was already thinking of what we need to do to generalise this. The main problem is that this overlay only works due to us controlling who instantiates the compiler.
(tests refer OpalCompiler, rest of the image just uses the "Smalltalk compiler", which now points to the overlay). 

There has to be a way to generalise this... maybe by rewriting in general references from the rest of the image? And we need to hook into test running to make sure that uses *not* the overlay.

Needs some exploration...

PR #6376, opened 2020-05-20
Fix Cairo bindings for PDF and SVG generation

### LINKED ISSUE #6374: Add bindings for Cairo writing PDF and SVG files
I read in [cairo documents](https://cairographics.org/manual/cairo-PDF-Surfaces.html) that creation of PDF and SVG files is quite easy: you just need to create another type of surface that is polymorphic to the one currently used to render on screen.

I also found [an example](http://zetcode.com/gfx/cairo/cairobackends/) and [another example](http://www.ie.u-ryukyu.ac.jp/~e085739/c.cairo.3.html) showing how to do it, and confirm it shouldn't be a big effort at all to create the missing bindings in Pharo.

**Is your feature request related to a problem? Please describe.**
AFAIK, the possibility to export files could be very useful for Roassal3 visualizations.

**Describe the solution you'd like**
I already did a test and it worked. I can create a PR and let community review it.

**Describe alternatives you've considered**
Not many alternatives... the `AthensCairoPDFSurface` already exists in the image but has no references to take as an example or test.
I only had to do 2 changes:
- a cairo function (cairo_surface_finish) to `AthensCairoSurface` make it write the PDF correctly.
- fix the release (finalizeResourceData:) of this class.

**Expected development cost**
A code review may be enough.

tinchodias:
Fix #6374


tinchodias:
For the record, it is possible to manually test with:
```Smalltalk
AthensCairoSurfaceExamples exampleExportTigerToSVG
AthensCairoSurfaceExamples exampleExportTigerToPDF
```
And open the generated files in working directory

tinchodias:
Creating a test to assert that a non empty file is generated and then delete it should be easy. I will do it.

tinchodias:
**Note:** 
We observed that the generated SVG has the text converted as path.  We discovered this in the some tests performed with @akevalion. 

I didn't see and mention in the cairo documentation to switch this behavior to generate [text elements](https://developer.mozilla.org/en-US/docs/Web/SVG/Element/text) instead of converting to paths.

Probably, the PDF export behaves the same way.

PR #6395, opened 2020-05-25
6278-RBProgramNodeparents-is-messy

### LINKED ISSUE #6278: RBProgramNode>>parents is messy
In the Pharo AST, it is possible to get all the children of a node with the method #allChildren. 

The equivalent for the parents is a method called #parents. The name is not coherent. 

Another problem is that parents return the receiver in the list. IMO this is really wrong because a node is to its own parent. 

I propose those changes:
- Introduce #allParents returning all the parents *without* the receiver
- Introduce #withAllParents that is an equivalent of the current #parents 
- Deprecate #parents in favour of #withAllParents
- Replace the senders of #parents in the system either with #withAllParents or #allParents

kasperosterbye:
Fixes #6278

### DIFF
--- a/src/AST-Core-Tests/RBMethodNodeTest.class.st
+++ b/src/AST-Core-Tests/RBMethodNodeTest.class.st
@@ -169,7 +169,6 @@ RBMethodNodeTest >> testPrimitiveModuleErrorIsPrimitive [
 	
 ]
 
-
 { #category : #'tests - primitives' }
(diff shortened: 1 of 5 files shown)


MarcusDenker:
failing tests not related.( I added an issue for the two test timing out: https://github.com/pharo-project/pharo/issues/6409)

MarcusDenker (approved):
good change! it is even backward compatible.

PR #6399, opened 2020-05-26
Create a common superclass for PDF and SVG surfaces and fix extent

### LINKED ISSUE #6398: Athens-Cairo surfeces to export PDF and SVG answer a wrong extent
**Describe the bug**
Trying to export a Morph as discussed in #6374, I discovered that the cairo functions used to get the width and height don't work on PDF and SVG surfaces (recently fixed/introduced in a PR).

**To Reproduce**
Create an Athens and access the extent, width or height. Example of assert that fails:
~~~
	| surface |
	surface := AthensCairoPDFSurface
		extent: 100@50
		fileName: 'fileName.pdf'.
	self assert: surface extent equals: 100@50
~~~

**Version information:**
OS: P9

**Expected development cost**
Already done in my computer.

tinchodias:
Also did a simple refactoring in AthensTiger to make the code a bit more expressive.

Fixes #6398.

### DIFF
--- a/src/Athens-Cairo-Tests/AthensCairoSVGSurfaceExportTest.class.st
+++ b/src/Athens-Cairo-Tests/AthensCairoSVGSurfaceExportTest.class.st
@@ -6,5 +6,5 @@ Class {
 
 { #category : #tests }
 AthensCairoSVGSurfaceExportTest >> athensSurfaceClass [
-	^ AthensCairoPDFSurface
+	^ AthensCairoSVGSurface
 ]
--- a/src/Athens-Cairo-Tests/AthensCairoSurfaceExportTest.class.st
+++ b/src/Athens-Cairo-Tests/AthensCairoSurfaceExportTest.class.st
@@ -64,7 +64,7 @@ AthensCairoSurfaceExportTest >> testExportBoxes [
 	Smalltalk garbageCollect.
 
 	self assert: self fileReference exists.
-	self assert: self fileReference size > 500.
+	self assert: self fileReference size > 400.
 ]
 
 { #category : #tests }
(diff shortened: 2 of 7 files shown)


tinchodias:
You can try before and after the PR the following script to export a Calypso window to PDF:
~~~smalltalk
"First, open a window"
aMorph := 1 browse window.

"Draw in a PDF"
anAthensSurface :=
	AthensCairoPDFSurface
		extent: aMorph extent
		fileName: Time microsecondClockValue asString, '.pdf'.
aMorph defer: [
	anAthensSurface drawDuring: [ :anAthensCanvas |
		anAthensCanvas pathTransform 
			translateBy: aMorph bounds origin * -1.
		aMorph fullDrawOn: anAthensCanvas asCanvasWrapper ].
	anAthensSurface finish.
	anAthensSurface := nil.
	Smalltalk garbageCollect; inform: 'Export Done' ].

"BTW, also export to Morph to other format to compare between vector based PDF vs pixel-based (and compressed) PNG"
aMorph exportAsPNGSilently.


~~~

tinchodias:
**Note**: some times the morph is broken during the export, that's why I added the defer: to the example (but didn't fix it). Didn't debug the problem and don't have plans to do it soon. 

Does somebody know the reason?

Ducasse:
This is cool because we use this for generating screenshot for books :) as we did in booktester.

Ducasse:
for the bug this is probably a thread issue.

Ducasse:
What would be good is to be able to write in a stream or fileReference so that we can have in memory tests. may be the C lib allows that because files is limited.

tinchodias:
@Ducasse yes, there is another function in the API where you pass a callback as argument instead of a filename. This one: https://cairographics.org/manual/cairo-PDF-Surfaces.html#cairo-pdf-surface-create-for-stream
Maybe somebody has time and wants to implement it.



<blockquote><div><strong><a href="https://cairographics.org/manual/cairo-PDF-Surfaces.html#cairo-pdf-surface-create-for-stream">PDF Surfaces: Cairo: A Vector Graphics Library</a></strong></div></blockquote>

Issue #5701, opened 2020-02-17
UFFI: Troubles with struct of doubles


tinchodias:
I implemented this tiny library:

```C
typedef struct cpVectD{double x,y;} cpVectD;
cpVectD newVectD(double x, double y);
double getVectDX(cpVectD point);
double getVectDY(cpVectD point);

typedef struct cpVectI{int x,y;} cpVectI;
cpVectI newVectI(int x, int y);
int getVectIX(cpVectI point);
int getVectIY(cpVectI point);
```

I followed the UFFI booklet (wonderful, many thanks) and it works with ints, but not with doubles.

Steps to reproduce:

- download P8 with: `curl https://get.pharo.org/64/80+vm | bash`.
- unzip attached file in same directory
- ensure lib in `issue` directory is compiled. For MacOs, the zip already.provides the dylib (created with `gcc -dynamiclib issuelib.c -o issuelib.dylib`).
- Open pharo, filein the provided .st
- Run tests CpLibraryTest. For CpVectI (struct of ints) it works, for CpVectD (struct of doubles) it fails. (The assert shouldn't have equals: but closeTo: but it shows the error.)

[UFFIDoublesStructIssue.zip](https://github.com/pharo-project/pharo/files/4214087/UFFIDoublesStructIssue.zip)

tinchodias:
Following a suggestion via chat, I tried "TFFI" with the same results (but not sure I did it well). What I did:
- Cloned https://github.com/pharo-project/threadedFFI-Plugin
- Executed the steps of gitlab-ci:
    - cmake .
    - make install
    - make test
(it didn't work in Mac but it did in Debian 10).
- Started the image created by the test, loaded my code and ran the tests... same.

<blockquote><img src="https://avatars3.githubusercontent.com/u/1838382?s=400&v=4" width="48" align="right"><div><img src="https://github.githubassets.com/favicon.ico" height="14"> GitHub</div><div><strong><a href="https://github.com/pharo-project/threadedFFI-Plugin">pharo-project/threadedFFI-Plugin</a></strong></div><div>A Threaded FFI plugin for Pharo. It extends the Cog VM with non blocking queue based FFI. - pharo-project/threadedFFI-Plugin</div></blockquote>

tinchodias:
In fact, it's partially fixed with TFFI: the returned struct of doubles has the right x y values in the test, but the functions don't. More exactly: the accessor #y answers the right Float, but #getY answers zero. Such method is defined as follows:
[...]

PR #6400, opened 2020-05-26
Image side changes to use the new memory accessor primitives.

ronsaldo:
This PR includes the image side changes to take advantage of the JIT compiled memory accessing primitives that introduced in the following PR in the VM: https://github.com/pharo-project/opensmalltalk-vm/pull/90

Since these changes introduce new primitives, on the primitve failure path there is support to use the old primitives with a slight performance penalty (an additional message send), but allowing the usage of a VM without the new primitives.

### DIFF
--- a/src/FFI-Kernel/ByteArray.extension.st
+++ b/src/FFI-Kernel/ByteArray.extension.st
@@ -7,47 +7,158 @@ ByteArray >> asExternalPointer [
 ]
 
 { #category : #'*FFI-Kernel' }
-ByteArray >> booleanAt: byteOffset [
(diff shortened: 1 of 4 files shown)


estebanlm (approved):

estebanlm:
tests not related.
this will make image FFI access slower (slightly, tolerable) temporary, but it will match with the new primitives for FFI in the upcoming headless VM.

tinchodias:
Might this fix #5701 ?

ronsaldo:
@tinchodias This is not related to that problem. I added a comment on the reason on the issue.

PR #6401, opened 2020-05-26
RequestorScopeHardenNonStringVarname

MarcusDenker:
We are using in some cases non-Strings/Symols as varnames when creating temps for stack manipulation in Slots (see e.g. emitStore: in ObservableSlot). We should move towards always using symbols. But until we do, we need harden the lookupVar: of requestors to work with non strings.

-> check for strings when checking upper case
-> remove #asSymbol (the cases where this is needed this would be done in the called method anyway)


demarey:
Looks good to me!
In addition, it solves a problem with compilation of ObservableSlots

PR #6417, opened 2020-05-29
6277-RBTemporaryNode-does-not-return-the-right-info-when-in-a-sequence-node

### LINKED ISSUE #6277: RBTemporaryNode does not return the right info when in a sequence node
In Pharo 8 I have this method:

```Smalltalk
method
  | test |
  test := 2.
  test.
  ^test
```

First, I was surprised that we only have a RBTemporaryNode and not a distinction between temporary declaration and usage. But I can live with it. 

The problem arises when we ask the temporaries if the are definitions or read. 

I was expecting the third RBTemporaryNode of this method to be considered as read. But in reality #isRead return false and #isDefinition return true.

The problem arises from the fact that the temporary is in the sequence node defining it. In theory, only declarations should be in this sequence node, but in reality, it's possible to have useless temporaries  like the one I presented above.

MarcusDenker:
- implement isDefinition to not use #defines but instead check if the objects is a variable definition
- format #defines better
- add test 

fixes #6277

### DIFF
--- a/src/AST-Core-Tests/RBMethodNodeTest.class.st
+++ b/src/AST-Core-Tests/RBMethodNodeTest.class.st
@@ -169,7 +169,6 @@ RBMethodNodeTest >> testPrimitiveModuleErrorIsPrimitive [
 	
 ]
 
-
 { #category : #'tests - primitives' }
 RBMethodNodeTest >> testPrimitiveModuleIsPrimitive [
 	"
--- a/src/AST-Core-Tests/RBVariableNodeTest.class.st
+++ b/src/AST-Core-Tests/RBVariableNodeTest.class.st
@@ -8,6 +8,29 @@ Class {
 }
 
 { #category : #tests }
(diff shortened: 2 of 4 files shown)


hogoww:
I don't see the #defines: format on github, is this normal?


PR looks good to me, but I have a problem with the naming.
With a C backgroud:
- a Declaration => int a;
- a Definition => int a = 2;
I don't know if this maps to other languages as well.
With that in mind, when looking at Cyril's example:

Object compile: 'myMethod
  | test |
->   test := 2.
  test.
  ^test'.

For me the second one is a definition., not the first.

Related: I added in the default visitor #visitTemporaryDeclarationNode: a while back
So there is a name inconsistency

MarcusDenker:
for defines vs declares, we did already have #defines:, I did not want to change the vocabulary. This can  be  done but that would be another issue.

hogoww:
Indeed.
Test failure looks unrelated :)

PR #6424, opened 2020-05-29
extract-return-from-conditionals-for-packages-d-to-f

jecisc:
Extract returns from conditionals for packages starting by d to f


Ducasse:
I love chanel :)

PR #6428, opened 2020-05-29
extract-return-from-conditionals-for-packages-h-to-k

jecisc:
Extract returns from conditionals for packages starting by h to k


MarcusDenker:
testTallyInstructions fails which might be related (code that it counts was changed?)

jecisc:
Indeed, the test failure is related to the PR

MarcusDenker:
"Got 28 instead of 27"
I think these are just smoke tests and you can just edit the number that it expects

MarcusDenker:
There's now a conflict. as this was auto-generated, can we close the PR for it to be redone?

PR #6432, opened 2020-05-29
[fix] remove usage of reflectivity to get ast methods by default.

hogoww:
ASTCCache was using reflectivity by default.
We removed it and are using a strategy instead.

This will allow us to work with bootstrap errors.


MarcusDenker:
Build fails:

[64] MetacelloNotification: Loaded -> Reflectivity-Tests-tonel.1 --- tonel:///builds/workspace/uest_and_branch_Pipeline_PR-6432/64/src --- cache
[64]   IconStyler
[64] You must resolve these dependencies before you will be able to load these definitions: 
[64]   CoverageRubricStyler
[64]   CoverageRubricStyler>>#addIconStyle:from:to:
[64]   MetalinkIconStyler
[64]   MetalinkIconStyler>>#shouldStyle
[64]   MetalinkIconStyler>>#iconBlock:
[64]   MetalinkIconStyler>>#iconFor:
[64]   MetalinkIconStyler>>#iconLabel:
[64]   MetalinkIconStyler>>#shouldStyleNode:
[64]

MarcusDenker (changes requested):
Somehow the build fails, no CI tests are run (I added the printout as a comment)

hogoww:
Yes, we are using the CI to test the work, it is not readyl.
We just had to wait for a long time for the CI to run because of the affluence of pull requests :p

MarcusDenker:
I looked at the code and yes, it makes sense like this

hogoww:
Issues seem unrelated to me, but I don't remember flaky stuff in Athens, so this is weird.
Anyone has an idea ? =]

MarcusDenker (approved):
OK!

PR #6434, opened 2020-05-29
extract-return-from-conditionals-for-packages-o

jecisc:
Extract returns from conditionals for packages starting by o


MarcusDenker:
There is a conflict as it edits a method that I just changed in another PR

jecisc:
Fix

MarcusDenker (approved):
ok

PR #6436, opened 2020-05-29
extract-return-from-conditionals-for-packages-p-to-r

jecisc:
xtract returns from conditionals for packages starting by p to r


MarcusDenker (approved):
ok

PR #6437, opened 2020-05-29
extract-return-from-conditionals-for-packages-s

jecisc:
Extract returns from conditionals for packages starting by s


Ducasse:
I do not get why the rb test is failing.

Ducasse:
But let us merge and see.

PR #6444, opened 2020-05-29
extract-assignations-from-conditionals-for-packages-t

jecisc:
Extract assignations from conditionals for packages starting by t


MarcusDenker (approved):
ok. test failures not related

PR #6449, opened 2020-05-29
Fix #6448 Add a way to get a FileLocator when possible instead of a FileReference 

### LINKED ISSUE #6448: Add a way to get a FileLocator when possible instead of a FileReference
When using paths with well-known origins, it would be nice to have the choice to get a file locator (late-bound reference) instead of a file reference. It can apply to known origins (e.g. home, documents, etc.)

demarey:


demarey:
When using paths with well-known origins, it would be nice to have the choice to get a file locator (late-bound reference) instead of a file reference. It can apply to known origins (e.g. home, documents, etc.)

Ducasse:
Christophe  

StringTest does not exist when reloading DeprecatedFileStream.
And it leads to a bug when reloading during the post bootstrap reloading process.

[4] + /builds/workspace/uest_and_branch_Pipeline_PR-6449/64/bootstrap-cache/vmtarget/pharo --headless Pharo9.0-PR-compiler-64bit-eebe8ee.image --no-default-preferences loadHermes DeprecatedFileStream.hermes FileSystem-Core.hermes FileSystem-Disk.hermes --save --no-fail-on-undeclared
[64] pthread_setschedparam failed: Operation not permitted
[64] This VM uses a separate heartbeat thread to update its internal clock
[64] and handle events.  For best operation, this thread should run at a
[64] higher priority, however the VM was unable to change the priority.  The
[64] effect is that heavily loaded systems may experience some latency
[64] issues.  If this occurs, please create the appropriate configuration
[64] file in /etc/security/limits.d/ as shown below:
[64] 
[64] cat <<END | sudo tee /etc/security/limits.d/pharo.conf
[64] *      hard    rtprio  2
[64] *      soft    rtprio  2
[64] END
[64] 
[64] and report to the pharo mailing list whether this improves behaviour.
[64] 
[64] You will need to log out and log back in for the limits to take effect.
[64] For more information please see
[64] https://github.com/OpenSmalltalk/opensmalltalk-vm/releases/tag/r3732#linux
[64] SystemNotification: [Hermes] Reading DeprecatedFileStream.hermes
[64] SystemNotification: [Hermes] Installing DeprecatedFileStream.hermes
[64] SystemNotification: [Hermes] Remaining Undeclared variables in the system: #(#FileLocator #FileAttributeNotSupported #BISimpleFormatter #StartupPreferencesLoader #PragmaCollector #FileSystem)
[64] SystemNotification: [Hermes] Reading FileSystem-Core.hermes
[64] SystemNotification: [Hermes] Installing FileSystem-Core.hermes
[64] [31mError: Required class named: StringTest does not exists.

demarey:
StringTest does exist but, yes probably an error with the bootstrap process.
StringTest is probably loaded after FileSystem-Test.
I will look to fix that and also add a missing method to convert a file reference to a locator.

src/FileSystem-Core/FileLocator.class.st
```diff
@@ -244,6 +244,26 @@ FileLocator class >> flushCaches [
...
+	"Returns a file locator if aPath is a reference to a supported origin or is a child of an origin.
+	Else returns nil.
+	Should not be called direcly. Prefer the use of Path or String>>#asFileLocatorOrReference " 
+	self supportedOrigins
+		detect: [ :origin | | locator |
+			locator := self perform: origin.
+			locator asPath = aPath or: [ locator contains: aPath] ]
+		ifFound: [ :origin | | locator |
+			locator := self perform: origin.
+			^ locator resolve: (aPath relativeToPath: locator asPath) ].
```
jecisc:
```suggestion
	self supportedOrigins
		do: [ :origin | 
			| locator locatorPath |
			locator := self origin: origin.
			locatorPath := locator asPath.
			(locatorPath = aPath or: [ locatorPath containsPath: aPath ]) ifTrue: [ ^ locator resolve: (aPath relativeToPath: locatorPath) ] ].
```

jecisc (changes requested):
I proposed an optimized version of this feature. A little less than twice as fast. Should have the same behaviour.

Ducasse:
I do not know when is the last time this issue was validated but we will see.

Issue #6379, opened 2020-05-20
some deprecationTransform rules are bogus


Ducasse:
Oleksandr did a nice analysis of the deprecated transform rules.

Ducasse:
![Screenshot 2020-05-20 at 12 07 20](https://user-images.githubusercontent.com/432549/82433947-7b22c380-9a92-11ea-9cf8-a1069801284e.png)

Ducasse:
The first rule is invalid because of a typo: antecedent contains an extra ”s” and hence the selector in antecedent is different from the deprecated selector, which violates the first condition of validity. The second rule seems outdated: deprecated method whenActivatedDo: has already been renamed to doubleClickAction: (the name suggested by the consequent of a Rewriter rule) but the deprecation was not removed from it. Rules 6 and 7 were also found in deprecated methods whose selectors do not correspond to antecedents, but since the selectors in consequents are also different, it is hard to interpret this inconsistency.


Violation of Condition 2. Rules 3 and 4 from Table 3 have the same antecedent and consequent, which is a violation of the second condition of validity. Both rules appear in test methods, which explains this violation. Rule 5 violates both first and second condition. Both antecedent and consequent contain selector whenEnabledChangedDo: even though antecedent should match the calls to whenEnabledChanged:. This can be explained by a typo.
Violation of Condition 3. The consequent of rule 8 is empty, which violates the third condition of validity. Perhaps, the programmer who wrote this code wanted to remove the calls to RubTextFieldArea >> wrapped:, however, this is not a valid use of the Rewriter tool.

olekscode:
I'm working on this issue

olekscode:
- I fixed the first deprecation from the table
- Second and third one are inside a test class, so they are OK
- The other ones have already been removed in Pharo 9

This table was created for Pharo 8
I will explore deprecations from Pharo 9 separately and open a new issue

olekscode:
The PR was merged, so this issue should be closed
(I'm surprised that it didn't close automatically)

PR #6451, opened 2020-05-29
Fixing #6379

olekscode:


Ducasse:
Oleksandr your process to commit is strange because we miss the title of the issues.

olekscode:
I will sit with someone and see what I am doing wrong

PR #6456, opened 2020-05-29
6378-SmallInteger-methods-are-calling-for-test-and-executable-comment

carolahp:
tests and executable comments added to SmallInteger


carolahp:
There was a bug in one executable comment. I fixed it in new pull PR:
https://github.com/pharo-project/pharo/pull/6465

PR #6458, opened 2020-05-29
refactoring-ast-to-use-strategies

tesonep:
Refactoring the VariableNodes to use an strategy instead of having subclasses that are changed with the change class primitive.


src/AST-Core/RBVariableNode.class.st
```diff
@@ -44,33 +45,93 @@ RBVariableNode class >> named: aName start: aPosition [
...
 RBVariableNode >> acceptVisitor: aProgramNodeVisitor [
-	^ aProgramNodeVisitor visitVariableNode: self
+	^ kind acceptVisitor: aProgramNodeVisitor
 ]
 
 { #category : #converting }
 RBVariableNode >> adaptToSemanticNode [
 	
-	self primitiveChangeClassTo: self binding semanticNodeClass new
+	kind := self binding semanticNodeKind new.
```
dionisiydk, line 119: Why we need another entity? Binding objects are already about those types. And the implementors of #semanticNodeKind  shows this very clearly.

dionisiydk: We can do the same visiting logic (introduced here) through OCVariable classes (which the binding here). 
I would also rename #binging variable into #variable.

MarcusDenker: Yes, I think we should move the Semantic analysis (same part of it) into the AST package.

The binding then can play exactly the role of the strategy for visiting.

Maybe we could use this PR as the first step in this direction. The refactoring it does for clients would look the same...

This means: we would merge this, and then in a second step do a PR that move the Semantic vars into the AST package, replacing the Kind classes.

jecisc:
Can't we do this in a backwards-compatible way for at least one version of Pharo please?

I see multiple classes removed and I have multiple projects using and extending the AST... I'm sure I'm not the only one.

dionisiydk:
It's sad that we did not merge it in time.
Maybe resolving conflicts will not be difficult.

MarcusDenker:
The Variable Hierarchy is now in Sync with the RBVariableKind Hierarchy suggested in this PR.

This means we can now redo all and just use the Variable subclasses to dispatch.

MarcusDenker:
I started to redo this step-by-step, but with the Variable Hierarchy instead of the Kind Hierarchy.

Issue #6412, opened 2020-05-28
RBComment should have parent pointer to node


MarcusDenker:
If you have a RBComment in hand, it is hard to find the node that it comes from.

As a simple improvement we could add a "parent" just as we have in RBProgramNodes.

-> add parent ivar
-> add accessor

The place in the Parser where this would ne set is # addCommentsTo:, it need a line

(comments do: [ :each | each parent: aNode ])

MarcusDenker:
merged

PR #6461, opened 2020-05-29
6412-RBComment-should-have-parent-pointer-to-node

oliveiraallex:
RBComment should have parent pointer to node #6412


MarcusDenker (approved):
Ok.

PR #6468, opened 2020-05-31
Fixes changes file flushing issue

### LINKED ISSUE #6063: SourceFileArray problem when only part of the method is in the internal buffer
After https://github.com/pharo-project/pharo/pull/5292 we starting to see random syntax errors when loading code. It seems that we are running into an edge case when part of the source code for a method is written to disk in the .changes file, and part of the method is still in the write buffer  of the `SourceFile` object. 

If the above happens `SourceFile>>#ensureWrittenPosition:` does not actually trigger a flush because the position of the source code of the method is before the beginning of the buffer part that is in memory. Hence, the buffer is not flushed and only part of the source code is returned. 

The best solution would be to replace `file ensureWrittenPosition: position` in `SourceFileArray>>#readStreamAtFileIndex:atPosition:ifPresent:ifAbsent:`  with `file flush`. Might decrease performance but otherwise it can be difficult to ensure that the .changes file contains the entire content.

Below is a small analysis that we did to come to this conclusion with screenshots.  Maybe there can also be other causes. Unfortunately we did no found a way to reliable reproduce the bug.

-------------------------------------------------------

Initially we were getting random Syntax errors when loading code. They were in quite different places:
Screenshot 2020-03-31 at 14 28 26

When inspecting the debug session behind the syntax error we see that it happens when installing a method (`MethodAddition`). Looking at the code from the method addition shows the right code for the method:
Screenshot 2020-03-31 at 14 39 26

So the code is the same as in the tonel file. However looking at the `compiledMethod` attribute from the `MethodAddition` object  the source code is not correct. It is the same code as in the syntax notification error:
Screenshot 2020-03-31 at 14 39 36

So the source code of the method is not correct.

Looking at the .changes file with a hex editor show something interesting. Only part of the code is present there. Actually the same part that we get when calling `sourceCode`:
Screenshot 2020-03-31 at 14 44 30

This suggests something is not flushed correctly or something is overridden. Inspecting the SourceArray shows the interesting part. The read/write buffers are dirty. The remaining part of the method is still in the internal buffer waiting to be flushed.
Screenshot 2020-03-31 at 13 10 40

So indeed first part of the method is in the changes file,  the other part in the internal buffer waiting to be flushed. 

Normally SourceFile should handle that however the optimisation in `SourceFileArray>>#readStreamAtFileIndex:atPosition:ifPresent:ifAbsent:`  seems to not capture this use case. In this method `SourceFile>>#ensureWrittenPosition:` is called to flush changes. However, this only flushes changes if the `position` attribute is inside the internal buffer. So in other words the buffer is flushed if the entire method is still in memory.

However, this is not the case here. The size of the .changes file is 28901376 bytes on disk. This is correct in `SourceFileBufferedReadWriteStream`; the buffer there hass size 28901376.  However when debugging #sourceCode for the `#timeStampParser` method the source pointer of the method is 28901147. So it's before the beginning of the stream in `SourceFileBufferedReadWriteStream`.

Screenshot 2020-03-31 at 14 57 24

`SourceFile>>#ensureWrittenPosition:` calls `SourceFileBufferedReadWriteStream>>#isPositionInBuffer:` which only checks if the position of the source code is within the limits of the buffer. In this case this does not happen as the beginning of the method is in the .changes file and only a part of it is in the buffer waiting to be flushed.
Screenshot 2020-03-31 at 15 04 44

`SourceFileArray>>#readStreamAtFileIndex:atPosition:ifPresent:ifAbsent:` does not know how long the method should be as it needs to read it first. The solution would be to flush anyway the changes as otherwise we might get an incomplete method.   

Alternatively `SourceFileArray>>#readStreamAtFileIndex:atPosition:ifPresent:ifAbsent:` could be changed to try first to read the entire method and if the end is not found flush changes and retry the read.

theseion:
SourceFileBufferedReadWriteStream>>ensureWrittenPosition: now performs an extra check to see whether the requested position has been written to the file if the buffer is dirty, since in that case the file may not contain the complete chunk that will be looked up.

Fixes #6063 for Pharo 9

### DIFF
--- a/src/System-Sources/SourceFileBufferedReadWriteStream.class.st
+++ b/src/System-Sources/SourceFileBufferedReadWriteStream.class.st
@@ -104,12 +104,16 @@ SourceFileBufferedReadWriteStream >> defaultBufferSize [
 
 { #category : #writing }
 SourceFileBufferedReadWriteStream >> ensureWrittenPosition: aPosition [
-
 	isDirty ifFalse: [ ^ false ].
-	
-	(self isPositionInBuffer: aPosition)
-		ifTrue: [ self flush. ^ true ].
-		
+
+	((self isPositionInBuffer: aPosition) or: [
+		"The position has been written to disk but
+		the buffer is dirty, so there is potentially
+		data in the buffer that we need."
+		aPosition <= bufferOffset ]) ifTrue: [
+			self flush.
+			^ true ].
+
 	^ false
 ]
 


MarcusDenker (approved):

noha:
Is this relevant for pharo8? If so, can we backport?

MarcusDenker:
There is already: https://github.com/pharo-project/pharo/pull/6467

noha:
👍 Thanks

PR #6475, opened 2020-06-03
Deprecated Slot

MarcusDenker:
This is the Depecated Slot from https://github.com/jecisc/Bazard/tree/master/src/DeprecatedSlot 

This for now just allows us to declare a slot as deprectated, raising a log entry at runtime just like a deprecated method.

The next step is to add transformation support to it, too.

### DIFF
--- a/src/VariablesLibrary/DeprecatedSlot.class.st
+++ b/src/VariablesLibrary/DeprecatedSlot.class.st
@@ -1,0 +1,76 @@
+"
+Description
+--------------------
+
+I am a slot used when the usage of an instance variable should be deprecated.
+
+Examples
+--------------------
+
+	FamixMetamodelGenerator subclass: #FamixBasicInfrastructureGenerator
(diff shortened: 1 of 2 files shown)


jecisc (approved):
A next step would also be to improve the deprecation hierarchy.

Would be nice to have for example an abstraction Deprecated class then MethodDeprecation, SlotDeprecation, ClassDeprecation (raised when we instantiate a deprecated class?).

What do you think?

MarcusDenker:
Yes! The exception hierarchy needs to be improved. I add it to the backlog

MarcusDenker:
errors not related

dionisiydk:
@MarcusDenker Is your proposal for slots refactoring implemented? According to them the deprecation can be just a specific decoration. So the developer will be able to use any kind of slot to implement deprecation, not only IndexedSlot which is superclass here.

MarcusDenker:
Not yet. That is still on the TODO. I want slots like this to be decorators that can be applied to any existing slot. This will come

PR #6477, opened 2020-06-03
emitStore-do-not-use-Object-new

MarcusDenker:
We should not use "Object new" as temp name in emitStore:, better use a string that starts with a number, this way it can not clash with other temps but is a string, which allows things to be simpler in the compiler.

### DIFF
--- a/src/Slot-Core/LiteralVariable.class.st
+++ b/src/Slot-Core/LiteralVariable.class.st
@@ -74,7 +74,7 @@ LiteralVariable >> definitionString [
 { #category : #'code generation' }
 LiteralVariable >> emitStore: aMethodBuilder [
 	| tempName |
-	tempName := Object new.
+	tempName := '0TempForStackManipulation'.
 	aMethodBuilder
 		addTemp: tempName;
 		storeTemp: tempName;
--- a/src/Slot-Core/GlobalVariable.class.st
+++ b/src/Slot-Core/GlobalVariable.class.st
@@ -17,9 +17,8 @@ GlobalVariable >> definingClass [
 
 { #category : #'code generation' }
 GlobalVariable >> emitStore: methodBuilder [
-	methodBuilder storeIntoLiteralVariable: self.
 
-
+	methodBuilder storeIntoLiteralVariable: self
 ]
 
 { #category : #'code generation' }
--- a/src/VariablesLibrary/ObservableSlot.class.st
+++ b/src/VariablesLibrary/ObservableSlot.class.st
@@ -36,7 +36,7 @@ ObservableSlot >> emitStore: aMethodBuilder [
 	"generate bytecode for 'varName value: <stackTop>'"
 
 	| temp |
-	temp := Object new.	"we need a unique Object as a temp name"
+	temp := '0slotTempForStackManipulation'.
 	"We pop the value from the stack into a temp to push it back in the right order"
 	aMethodBuilder addTemp: temp.
 	aMethodBuilder storeTemp: temp.
--- a/src/VariablesLibrary/WeakClassVariable.class.st
+++ b/src/VariablesLibrary/WeakClassVariable.class.st
@@ -19,7 +19,7 @@ Class {
 WeakClassVariable >> emitStore: aMethodBuilder [
 	"generate bytecode for 'varname at: 1 put: <stackTop>'"
 	| temp |
-	temp := '0slotTempForStackManipulation'.
+	temp := '0TempForStackManipulation'.
 	"Pop the value to store into a temp to push it back in the right order"
 	aMethodBuilder addTemp: temp.
 	aMethodBuilder storeTemp: temp.


dionisiydk:
There are many users with this pattern:
```Smalltalk
aMethodBuilder addTemp: temp.
aMethodBuilder storeTemp: temp.
aMethodBuilder popTop.	
```
What you think to introduce reusable method for this:
```Smalltalk
IRBuilder>>#storeTopIntoNewTemp: tempName
        | tempKey |
        tempMap := self currentScope tempMap.
	tempKey :=  'generatedTemp', tempMap size asString.
	tempMap at: tempKey put: tempMap size + 1.
	self cacheIndex: tempKey.
         ^tempKey
```
Temp name will be generated and returned. So users can be rewritten like:
```Smalltalk
ObservableSlot>>emitStore: aMethodBuilder
	"generate bytecode for 'varName value: <stackTop>'"
	| temp |
	temp := aMethodBuilder storeTopIntoNewTemp.

	"Push the value holder into the stack, then the value again, then send"
	aMethodBuilder pushInstVar: index.
	aMethodBuilder pushTemp: temp.
	aMethodBuilder send: #value:
```
It will make all variables with such #emitStore logic to have unique temp for them. Which means multiple technical temps in methods accessing multiple such variables.
I am not sure that single temp in that case is really desired. Can we be sure that it does not lead to problems?
But if we want shared temps between multiple variables then add parameter: aMethodBuilder storeTopIntoNewTemp: #tempName.

dionisiydk:
There are many users with this pattern:
```Smalltalk
aMethodBuilder addTemp: temp.
aMethodBuilder storeTemp: temp.
aMethodBuilder popTop.	
```
What you think to introduce reusable method for this:
```Smalltalk
IRBuilder>>#storeTopIntoNewTemp: tempName
       self addTemp: tempName; storeTemp: tempName; popTop
```
Then users can be rewritten like:
```Smalltalk
ObservableSlot>>emitStore: aMethodBuilder
	"generate bytecode for 'varName value: <stackTop>'"
	| temp |
        temp := 'GeneratedTempName'.
	aMethodBuilder storeTopIntoNewTemp: temp.

	"Push the value holder into the stack, then the value again, then send"
	aMethodBuilder pushInstVar: index.
	aMethodBuilder pushTemp: temp.
	aMethodBuilder send: #value:
```

dionisiydk:
I would also hide the name for generated temp from users. What happens when two such special variables are used in same method?
They both will ask methodBuilder to add same temp var. Is it required? 
I am curious about this part of #addTemp: 
```Smalltalk
"adding the same var multiple times reuses the same offset"
(tempMap includesKey: tempKey) ifTrue: [ ^ self cacheIndex: tempKey ].
```
Why we can't just return if variable is already in the map?

dionisiydk:
I guess the idea is that it is safe to use same technical temp shared between all variables in the method.
But here you are using different temp names for different variable types: 
- 0slotTempForStackManipulation 
- 0TempForStackManipulation.

Why it is necessary? I do not see why slots case is different than other vars.

We can have shortcuts (at least for the most of cases like 0TempForStackManipulation):
- #storeTopIntoTechTemp.
- #pushTechTemp.

And the example would be:
```Smalltalk
ObservableSlot>>emitStore: aMethodBuilder
	"generate bytecode for 'varName value: <stackTop>'"
        aMethodBuilder storeTopIntoTechTemp.

	"Push the value holder into the stack, then the value again, then send"
	aMethodBuilder pushInstVar: index.
	aMethodBuilder pushTechTemp.
	aMethodBuilder send: #value:
```
(with better names than mine of course :))

MarcusDenker:
"Why we can't just return if variable is already in the map?" 

It just returns. the additional call updates the index in the semantic variable. This is needed as it can be different one.

as for #storeTopIntoNewTemp:, one thing we need to see is that this is assembly level. too many helper methods do not fit (at least my) mental model here.. 

But I like the "storeTopIntoTechTemp" idea.. as hiding the name is nice. But as for generating multiple
vars: it would mean that 5 reads of the slot would add 5 different temps, which is not what we want.

It's a nice property of this kind of temp that we can reuse it.

dionisiydk:
> "Why we can't just return if variable is already in the map?"
> 
> It just returns. the additional call updates the index in the semantic variable. This is needed as it can be different one.

The interesting part is why it can be different? It's probably not an easy question :)

MarcusDenker:
The offset might not be different, it's just that the SemVar needs to be updated... I think this happens for copied temps? I will add it to the backlog to describe this in a comment. (took me a while to get this to work... and not updating the index exactly here was one of the things why)

MarcusDenker:
If you remove it, this test fails:

`
exampleTempNamedTempVectorInlinedLoop
"This test is for accessing temp var c within an inlined loop scope. 
Temp var c is written in the loop and closed in a block -> therefore it is in a tempvector. "
	| a b c |
	c := nil.
	a := 1.
	[ c := 42.
	b := thisContext tempNamed:'c' . a < 1 ]
		whileTrue: [ a := a + 1.
				[ c asString ] ].
	^ b

`

MarcusDenker:
The copying temp that stored the tempvector has no index in this case.

tesonep (approved):

src/Slot-Core/LiteralVariable.class.st
```diff
@@ -74,7 +74,7 @@ LiteralVariable >> definitionString [
 { #category : #'code generation' }
 LiteralVariable >> emitStore: aMethodBuilder [
 	| tempName |
-	tempName := Object new.
+	tempName := '0TempForStackManipulation'.
```
tesonep, line 77: I have a question.... in this case the names should not be unique? 
Because doing Object new each time produced different objects. 
Having an string as name, it does not produce clashes?

tesonep (changes requested):

src/Slot-Core/LiteralVariable.class.st
MarcusDenker: I don't think so:

- there is no nesting possible and thus the whole usage is just local.
- all use starts with writing into the var. Thus no confusion of old values.

So what will happen if there are multiple writes of this kind of var in the same method?
all of them will use the same synthesised var, but the use never overlaps.

At leas that is how I understand it...

MarcusDenker:
I will add the two aspects  discussed in the backlog (improve addTemp: and add helper methods for the synthesised temps)

PR #6478, opened 2020-06-03
pushOuterVectors-use-tempVectorVarNames

MarcusDenker:
simplify pushOuterVectors: by using #tempVectorVarNames

### DIFF
--- a/src/OpalCompiler-Core/RBPragmaNode.extension.st
+++ b/src/OpalCompiler-Core/RBPragmaNode.extension.st
@@ -6,7 +6,7 @@ RBPragmaNode >> asIRPrimitive [
 	| args module name spec |
 	args := (self arguments collect: [ :each | each value ]) asArray.
 	"normal methods have primitive number 0"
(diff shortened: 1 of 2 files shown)


src/OpalCompiler-Core/RBPragmaNode.extension.st
```diff
@@ -6,7 +6,7 @@ RBPragmaNode >> asIRPrimitive [
 	| args module name spec |
 	args := (self arguments collect: [ :each | each value ]) asArray.
 	"normal methods have primitive number 0"
-	self isPrimitive ifFalse: [ IRPrimitive null ].
+	self isPrimitive ifFalse: [ ^ IRPrimitive null ].
```
tesonep, line 9: Maybe I am losing something, but this is not the same behavior.

MarcusDenker: yes, the old version was a NOP. (as  IRPrimitive null has no effect).

The only client does this check itself, this is why it was never a problem.

(yes, maybe we should do a test...)

PR #6480, opened 2020-06-03
skip-testReifyTempOperationInstead

MarcusDenker:
When running the Reflectivity tests the first time, testReifyTempOperationInstead fails. This needs some more debugging, for the meantime, it is put on #skip.
(problem is added to the backlog)


dionisiydk (approved):
It reminds me to add "skipped" group for test cases in the browser

PR #6481, opened 2020-06-03
6409-Two-Zinc-tests-have-timeOut-on-the-CI

### LINKED ISSUE #6409: Two Zinc tests have timeOut on the CI
These two tests now time out on all 3 platforms. They work fine when I run them locally.

ZnEasyTest>>#testGetPng
ZnImageExampleDelegateTest>>#testDefaultImage

MarcusDenker:
Change the URL to be https: as it seems we else do not follow the redirect. fixes #6409  (I hope)


MarcusDenker:
Nope. So the timeout even happens with 10 seconds and the url changed to https://

PR #6483, opened 2020-06-04
simplify-firstPrecodeComment

MarcusDenker:
tiny cleanup in #firstPrecodeComment and #firstPrecodeCommentFor:


MarcusDenker:
We need to do a pass over comment APIs in general, so this is just some internal trivialities while reading the code without changing anything.

dionisiydk (approved):

PR #6493, opened 2020-06-05
Calypso: Class Tree Pane Should Be Collapsible

seandenigris:
Also some comment grammar fixes.

Result of [thread on dev ML](http://forum.world.st/Discoverability-tp5117839.html)

### DIFF
--- a/src/Calypso-SystemTools-FullBrowser/ClyFullBrowser.class.st
+++ b/src/Calypso-SystemTools-FullBrowser/ClyFullBrowser.class.st
@@ -476,6 +476,7 @@ ClyFullBrowser >> newClassView [
 			cell useFullIndentation.
 			self specifyClassItemNameOn: cell label for: item browserItem];
 		enableFilter;
+		enableCollapsingOfPreexpandedItems;
 		requestNavigationBy: #classSelectionChanged	
 ]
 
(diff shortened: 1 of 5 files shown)


src/Calypso-SystemTools-FullBrowser/ClyFullBrowser.class.st
```diff
@@ -789,13 +790,19 @@ ClyFullBrowser >> switchClassViewTo: aQueryResult inScope: aScope [
...
 { #category : #navigation }
 ClyFullBrowser >> switchToFlatClasses [
-	
-	| classQueries |
+	| classQueries fullQuery dataSource |
 	classQueries := OrderedCollection new.
-	packageView selection groupItemsByTypeAndDo: [ :typeClass :items | 
-		classQueries addAll: (typeClass prepareClassQueriesFrom: items in: navigationEnvironment)].
-	
-	classView showQueries: classQueries as: ClyExtensionLastSortedClasses hierarchical
```
dionisiydk: I think this code should be kept. But the widget (classView) needs to be configured what type of data source to use in the classView initialization:
```Smalltalk
classView := self newNavigationView
    ...
    enableCollapsingOfPreexpandedItems.
```
This way you will fix hierarchy mode as well.

seandenigris: Okay, I debated about adding the capability to hierarchy mode, but it seems like it's not a very wide tree - mostly (or all?) one child per parent, no? Doesn't matter to me. If you think I should add it I will. And in case you do, just to clarify, you're suggesting that I simply add that one `#enableCollapsingOfPreexpandedItems` send, correct @dionisiydk?

dionisiydk: > Okay, I debated about adding the capability to hierarchy mode, but it seems like it's not a very wide tree - mostly (or all?) one child per parent, no? Doesn't matter to me. If you think I should add it I will. 

I think for consistency classes should be always collapsable (or always not) independently of mode. 

> And in case you do, just to clarify, you're suggesting that I simply add that one `#enableCollapsingOfPreexpandedItems` send, correct

I mean that this configuration needs to be implemented in ClyQueryViewMorph. Then we will not need to touch a query API of this widget. We will just enable this behaviour in classView initialization.

seandenigris: Sounds good, but I have no idea how to translate the fix to what you suggested!

seandenigris: @dionisiydk Done. See second commit. Also, FYI, I realized you were absolutely right about extending the fix to hierarchy view - with a wide tree, the problem is the same there.

dionisiydk (approved):
Cool.

Now it would be good to show a screenshot here for other reviewers.

dionisiydk:
If there will be an issue with this feature it is easy to disable it later (one line in class view initialization)

seandenigris:
Trying to understand `OSWindowMorphicEventHandler` by seeing what its siblings are...

### Before
Screenshot 2020-06-08 21 49 43

### After
Screenshot 2020-06-08 21 48 45

p.s. @dionisiydk, are the screenshots what you had in mind?

Ducasse:
Do we have a way to get it expanded too?

seandenigris:
Yes, @Ducasse. It starts out expanded as usual. The only change is that it's now *possible* to collapse. Here's a screenshot of what you would see in the new UI by default (i.e. mirroring the "before" condition above). As you can see, it's "the same" except there are collapse/expand arrows at the parent tree nodes:

Screenshot 2020-06-12 10 48 26

seandenigris:
Anything blocking integration here? The test failures seem (to my uneducated eyes) unrelated...

PR #6497, opened 2020-06-07
[Disable Flaky Tests] Disable two Zinc flaky tests

dionisiydk:
Let's make it a convention to disable detected flaky tests with accompanying PR to enable them back.
The "disable PR" can be quickly merged to protect CI from new failures.
From the other side the "enable PR" will record the problem and track its state. In case of current Zinc issue it is always wailing on CI and therefore PR will have really good indication of problem.

### DIFF
--- a/src/Zinc-Tests/ZnEasyTest.class.st
+++ b/src/Zinc-Tests/ZnEasyTest.class.st
@@ -44,6 +44,9 @@ ZnEasyTest >> testGetJpeg [
 { #category : #testing }
 ZnEasyTest >> testGetPng [
 	| url result |
+	self flag: #flakyTest.
+	self skip.
+	
 	url := 'http://pharo.org/files/pharo.png'.
 	result := ZnEasy getPng: url.
 	self assert: result isForm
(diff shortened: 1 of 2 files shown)


src/Zinc-Tests/ZnEasyTest.class.st
```diff
@@ -44,6 +44,9 @@ ZnEasyTest >> testGetJpeg [
 { #category : #testing }
 ZnEasyTest >> testGetPng [
 	| url result |
+	self flag: #flakyTest.
+	self skip.
```
dionisiydk, line 48: The idea is to skip flaky tests and put a flag for that. So it can be easily found in the image.

dionisiydk:
CI fails due to another unrelated flaky test. I added separate PR to fix them: #6501

MarcusDenker (approved):
Pablo did find that it is indeed blocked on the level of the CI setup. So we can merge this now and when the admins fix it, we can un-skip.

I think this is a good idea, but I will wait for a second review.

MarcusDenker:
As we know the reason and that it will be fixed by the admins (and that this takes a while), I think we really can merge this.

PR #6501, opened 2020-06-07
[Flaky Test] Rely on SUnit time limit in recursion stopper tests. 

dionisiydk:
Recursion stopper tests were flaky due to very small time in assertions.
Here I removed time assertions. Tests will rely on SUnit time restriction with explicit small value for these tests.


dionisiydk:
Failed tests are unrelated (from Zinc)

MarcusDenker (approved):
ok

PR #6502, opened 2020-06-07
6495-Better-class-API

### LINKED ISSUE #6495: Better class API
Right now we have to write and duplicate code like that
```
	(self classLayout isKindOf: FixedLayout) 
```

while defining

```
hasSpecificLayout
   ^ (self classLayout isKindOf: FixedLayout) not
```

would solve the problem.

Ducasse:
Fixes: #6495 introduces new testing api + tests.

### DIFF
--- a/src/Slot-Core/Behavior.extension.st
+++ b/src/Slot-Core/Behavior.extension.st
@@ -6,7 +6,7 @@ Behavior >> classLayout [
 		ifNil: [ 
 			| superLayout scope |
 			superLayout := superclass 
-				ifNil: [ FixedLayout new slotScope: LayoutEmptyScope instance] "happend in the bootrap"
+				ifNil: [ FixedLayout new slotScope: LayoutEmptyScope instance] "happened in the bootrap"
 				ifNotNil: [:sl | sl classLayout].
 			scope := superLayout slotScope extend.
 			layout := superLayout class extending: superLayout scope: scope host: self ]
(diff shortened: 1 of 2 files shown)


src/Slot-Core/Behavior.extension.st
```diff
@@ -30,3 +30,14 @@ Behavior >> definesSlotNamed: aString [
...
 	^ self slotNames includes: aString
 ]
+
+{ #category : #'*Slot-Core' }
+Behavior >> hasSpecificLayout [
+	"Return true whether the receiver has a layout that is not a fixed one (which is the default linear layout for default classes."
+	"Point hasSpecificLayout >>> false"
+	"Array hasSpecificLayout >>> true"
+	
+	^ (self classLayout isKindOf: FixedLayout) not
```
dionisiydk, line 40: Can we have AbstractLayout>>#isSpecific (or isFixed) to avoid isKindOf:?

Ducasse: This is a good suggestion. Can you open an issue after this one is integrated.
right now I want to use this method since this code is used in the slot definition (so the kind of is already there).

Ducasse:
I prefer to have two fast and small PRs than a lengthy one. 
So I integrate and the next one will fix the implementation.

Ducasse:
If you want to fix it I created the issue. 
https://github.com/pharo-project/pharo/issues/6512

dionisiydk:
ok

PR #6511, opened 2020-06-07
6500-Cleanups

### LINKED ISSUE #6500: Cleanups
Convert some obvious collect: #foo into collect: [:each | each foo]

Ducasse:
Fixes: #6500 mainly sinmple code edits and rewrite of symbol as block use. I reviewed my changed two times so they should be ok. Went up to RGObject


MarcusDenker (approved):
ok

PR #6522, opened 2020-06-08
Unify Slots and LiteralVariable in single hierarchy

dionisiydk:
We have three separate hierarchies of variables:
- Slot subclasses
- LiteralVariable subclases
- OCAbstractVariable subclasses

There are a lot of duplication between them
- compilation (emitStore:emitValue:)
- properties
- metalinks and breakpoints
- more

Here Slots and LiteralVariable are merged under common Variable superclass.
It is a first step:
- Variable with #name instance variable and corresponding compare methods from the Slot.
- properties protocol and class var are moved to Variable (most trivial thing to do)
- LiteralVariable is adopted to be polymorphic to Associations.

In next PRs more duplications will be removed.


MarcusDenker:
nice! I will have a look!

dionisiydk:
Notice that loading it in the image will break it because of no migration support from Association to Variable. It's better to use the image from CI for review.  
(locally I have a migration script)

dionisiydk:
tests show issues. I will fix them later

MarcusDenker:
Failing tests not related.

MarcusDenker (approved):
This looks good! Very nice.

dionisiydk:
Cool. Now lets cleanup other methods. I checked Yesterday that almost all metalink methods are duplicated.

PR #6523, opened 2020-06-09
SDL Driver - Clean and Document; DO NOT INTEGRATE YET

seandenigris:
## Questions
1. How are FFIStructures like SDL_QuitEvent "automatically generated as per the method comments? That's pretty cool, @Esteban Lorenzano!!
1. For OSSDL2Driver, what is the purpose of the eventQueue? Events are (sometimes) stored there, but don't ever seem to be accessed. In fact there are not even any other references to that instVar...
1. Related to 2 above, the event is not placed in the queue if convertEvent: returns nil. Two ways that could apparently happen:
    1. The window map doesn't have a window with the ID specified by the event, but under what circumstances would an event arrive in that condition?
    1. window handleNewSDLEvent: mappedEvent returns nil, but why would a window do that? Does that signify something in the domain?

## Documentation Stubs
OSSDL2Driver is the primary “behind the scenes” object that connects Pharo to the SDL2 library (SDL). During its initialization, it starts an event loop (ultimately by sending itself a variant of eventLoopProcessXYZ). This event loop receives an SDL_Event structure from SDL and maps it to a SDL2MappedEvent subclass, using SDL_CommonEvent for “unknown” events i.e. those with no specific mapping (so in convertEvent:, how could the nil check be reached?). If the event does not specify a target window, sendEventWithoutWindow: then further converts it to anOSEvent before sending it to all windows, as well as any registered global listeners.

Why is SDL_TouchFingerDownEvent translated to OSTouchActionPointerDownEvent? What, if anything, is the distinction between “finger” and “action pointer”?

Individual gestures could register themselves and pass interesting events to their own handler, but much duplicated effort would be done e.g. keeping track of fingers. Another way, is to just register a single OSWindowGestureHandler instance, which then registers gestures via registerGesture:.

### DIFF
--- a/src/OSWindow-Core/OSWindowDriver.class.st
+++ b/src/OSWindow-Core/OSWindowDriver.class.st
@@ -1,15 +1,13 @@
 "
 I am a basic (abstract) class, which represents an OS window driver.
-The driver connects an OSWindow instances with underlaying operating system window(s) through managing OSWindowHandle(s). Driver provides an implemenation of all OSWindow functionality, starting from its creation, setting/retrieving its attributes, and finishing with event handling and/or rendering window's contents on screen.
+The driver connects OSWindow instances with underlying operating system window(s) by managing OSWindowHandle(s). I provide an implemenation of all OSWindow functionality - window creation, setting/retrieving attributes, and event handling and/or rendering its contents on the screen.
 
-The driver connects OSWindow(s) with operating system windows by providing the handle (see OSWindowHandle). The way how various OSWindow features and API are implemented is up to the concrete driver and thus considered private.
+The implementation of OSWindow features and API is up to the concrete driver and is considered private.
 
-The driver(s) responsible for initial window creation , proper setup and and managing external resources. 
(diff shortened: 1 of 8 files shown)


estebanlm:
some answers... 
1) this is just a source code generation that uses offset class variables to be sure we can update when we change platform :)
2) I have no idea. There is another event queue in the event handler used to simulate keypress, but this one I never understood why is there. Maybe we need to remove it? Maybe @ronsaldo knows better.
3) No idea either.

seandenigris:
Thanks, @estebanlm. Where is the code for the source code generation?

estebanlm:
just follow #rebuildFieldAccessors :)

ronsaldo:
I am not the person who implemented the touch screen support. I am pretty sure that this was implemented by the Thales people. This is a part that is probably not even tested/used, so it requires a proper clean up.

Ducasse:
Hello could we get some action on this PR?
Because this is not good to have PR that stay too long.

seandenigris:
I'll close it for now. I had done a lot of exploring, but by the time I got more info, I forgot everything I learned ha ha

PR #6531, opened 2020-06-09
6530-Job-progress-bar-is-under-the-bar

### LINKED ISSUE #6530: Job progress bar is under the bar

Ducasse:
Fixes: #6530 some cleans + make sure that the job bar is under the menu bar

### DIFF
--- a/src/Morphic-Base/SystemProgressMorph.class.st
+++ b/src/Morphic-Base/SystemProgressMorph.class.st
@@ -273,7 +273,8 @@ SystemProgressMorph >> initialize [
 		hResizing: #rigid;
 		vResizing: #shrinkWrap;
 		layoutInset: 10@6;
-		minWidth: 150
+		minWidth: 150.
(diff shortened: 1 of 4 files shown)


Ducasse:
Broken tests are unrelated.

src/Morphic-Base/SystemProgressMorph.class.st
```diff
@@ -273,7 +273,8 @@ SystemProgressMorph >> initialize [
 		hResizing: #rigid;
 		vResizing: #shrinkWrap;
 		layoutInset: 10@6;
-		minWidth: 150
+		minWidth: 150.
+	self position: 10@20
```
jecisc, line 277: An absolute position is not good IMO. 

Personally I change the size of my Menubar so it is still under. 

Instead, we could determine the position related to the menubar instead of hardcoding it? This would also remove the magic literals :)

PR #6534, opened 2020-06-10
fix-dependency-freetype-morphic

estebanlm:
FreeType cannot depend on Morphic, we may want Fonts without it!


estebanlm:
test error not related

MarcusDenker:
Why is Paragraph a class in Morpic? Shouldn't that be just text?

MarcusDenker (approved):
Besides the question why Paragraph is Morphic (which would be another PR), this looks good

estebanlm:
I agree... with both: should be Text, and is for another PR ;)

PR #6542, opened 2020-06-10
remove-morphic-dependency-on-fonts

estebanlm:
choosing a font is a UI operation that ends in Morphic. 
Remove that dependency.


MarcusDenker (approved):
if tests are ok --> looks good

PR #6543, opened 2020-06-10
6541-defaultTimeout-has-no-senders-The-method-for-tests-is-called-defaultTimeLimit

### LINKED ISSUE #6541: #defaultTimeout has no senders. The method for tests is called defaultTimeLimit
You can implement #defaultTimeLimit on the class side of a test. The default is 10 seconds.

But there are #defaultTimeout methods that should have been defaultTimeLimit instead.

TODO: remove all defaultTimeLimit that make no sense, change the ones in tests to be #defaultTimeLimit

It would be nice to have a pragma to set it per tests, too (but that is another TODO)

MarcusDenker:
- remove all defaultTimeLimit that make no sense, change the ones in tests to be #defaultTimeLimit

fixes #6541

### DIFF
--- a/src/Kernel-Tests/LocalRecursionStopperTest.class.st
+++ b/src/Kernel-Tests/LocalRecursionStopperTest.class.st
@@ -12,7 +12,7 @@ Class {
 }
 
 { #category : #running }
-LocalRecursionStopperTest class >> defaultTimeout [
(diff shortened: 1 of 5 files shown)


dionisiydk:
Ah, it was my fault

dionisiydk (approved):

MarcusDenker:
and it is really green...

PR #6545, opened 2020-06-10
Move Variables to Kernel

MarcusDenker:
Move all the Variables from Slots package to Kernel-Variable.


MarcusDenker:
First class variables are needed always. The current Kernel can not work without.

Maybe later we might move the WorkspaceVariable to a tools basics package, but for now it is ok here

dionisiydk (approved):

PR #6548, opened 2020-06-10
CompiledMethod-hasComment

MarcusDenker:
ChangeSet>>#methodsWithoutComments was using #firstPrecodeCommentFor: just for testing if the method has a comment.

It is much nicer to add a method that checks if a method has a method comment and use that. This way we can easily deprecate #firstPrecodeCommentFor:  later


Ducasse:
Marcus I do not know if it is related however there are two RB tests that are failing.

MarcusDenker:
all just timeouts due to a ver slow OS X slave

PR #6551, opened 2020-06-11
6506-Behavior--methodsInProtocol-invokes-subclass-only-defined-method

### LINKED ISSUE #6506: Behavior >> methodsInProtocol: invokes subclass only defined method
```
Behavior >> methodsInProtocol: aString

	^ (self organization listAtCategoryNamed: aString) 
			collect: [ :each | (self compiledMethodAt: each) ]
```
invoke methods defined in subclasses!!!

MarcusDenker:
- move #methodsInProtocol: down to ClassDescription in both Ring and Kernel
- add #allInstVarNames as subclassresponsibility in RGBehavioDefinition to fix one more rule 

fixes #6506

### DIFF
--- a/src/Ring-Definitions-Core/Behavior.extension.st
+++ b/src/Ring-Definitions-Core/Behavior.extension.st
@@ -7,13 +7,6 @@ Behavior >> methodNamed: aSelector [
 ]
 
 { #category : #'*Ring-Definitions-Core' }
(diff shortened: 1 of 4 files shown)


Ducasse:
May be the methodsInProtocols should be moved to the class and not handled as an extension.

PR #6562, opened 2020-06-11
Move #isLinux to platform classes

### LINKED ISSUE #6561: Move #isLinux to platform classes
Currently #isLinux utility method is on class side of SystemReporter - we should make it 
available on the OS platform classes similar to #isWindows or #isUnix so it could be used
independent from a loaded SystemReporter tool class.

Also we should add #isUnix and #isLinux to the API tests in #testAPI

astares:
Fix #6561


MarcusDenker:
Merge conflict

MarcusDenker (approved):
ok

PR #6567, opened 2020-06-11
Compiler-use-variable-not-assoc

MarcusDenker:
Now that LiteralVariables are not subclasses of Associations anymore, we need to update the use  of "assoc" as it is misleading. We use "variable" instead.

As this is deep in the compiler, no depreaction. (not part of the public API)


dionisiydk (approved):

dionisiydk:
tests are unrelated

PR #6573, opened 2020-06-11
Cleanup duplicated metalinks (reflectivity) protocol in variables

dionisiydk:
It cleanups duplicated metalinks (reflectivity) protocol in Slot and LiteralVariable by pushing all methods into their superclass Variable.

It also fixes a minor issue in calypso refactoring when it is applied on composite scope


MarcusDenker:
MetaLinkObjectAPITest>>testLinkObjectToSlotNamed fails which seems related

MarcusDenker (approved):
ok!

PR #6574, opened 2020-06-11
 Bugfix: object-centric metalinks were not able to distinguish object

### LINKED ISSUE #6572: Object-centric metalinks fail to recognize objects by identity
**Describe the bug**
Objects that do not compare through an identity check cannot be distinguished by Reflectivity object-centric machinery. 

**To Reproduce**
The following test shows the problem: the second set is not affected by the metalink because the machinery first looks into its registers and tries to find if the object is already there. It finds another set (the first one) and considers that the object already has an object-centric breakpoint. Therefore it does not install the link on the second set.

```Smalltalk
testLinkTargetsObjectsByIdentity
	|link ast set|
	link := MetaLink new metaObject: Halt; selector: #now.
	ast := (Set >> #add:) ast.
	
	set := Set new.
	set link: link toAST: ast.
	self should:[set add: 1] raise: Halt.
	
	set := Set new.
	set link: link toAST: ast.
	self should:[set add: 1] raise: Halt.
```

StevenCostiou:
For objects which equality is redefined (e.g. instances of Set, MetaLinks are not able to find on which object to install themselves.

Object-centric MetaLink machinery now uses WeakIdentityKeyDictionary to solve this problem.
Fixes #6572

### DIFF
--- a/src/Reflectivity/MetaLinkNodesMapper.class.st
+++ b/src/Reflectivity/MetaLinkNodesMapper.class.st
@@ -34,8 +34,8 @@ MetaLinkNodesMapper >> findNodesForObject: anObject [
 
 { #category : #initialize }
 MetaLinkNodesMapper >> initialize [
-	objectsForNodes := WeakKeyDictionary new.
-	nodesForObjects := WeakKeyDictionary new
+	objectsForNodes := WeakIdentityKeyDictionary new.
+	nodesForObjects := WeakIdentityKeyDictionary new
 ]
 
 { #category : #mapping }
--- a/src/Reflectivity-Tests/LinkInstallerTest.class.st
+++ b/src/Reflectivity-Tests/LinkInstallerTest.class.st
@@ -225,6 +225,21 @@ LinkInstallerTest >> testLinkOnTempVarForObject [
 ]
 
 { #category : #'links - installing' }
+LinkInstallerTest >> testLinkTargetsObjectsByIdentity [
+	|link ast set|
+	link := MetaLink new metaObject: Halt; selector: #now.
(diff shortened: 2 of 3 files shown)


dionisiydk (approved):

dionisiydk:
I think you also need to change MetaLinkRegistry>>#initialize

src/Reflectivity/MetaLinkNodesMapper.class.st
```diff
@@ -34,8 +34,8 @@ MetaLinkNodesMapper >> findNodesForObject: anObject [
 
 { #category : #initialize }
 MetaLinkNodesMapper >> initialize [
-	objectsForNodes := WeakKeyDictionary new.
-	nodesForObjects := WeakKeyDictionary new
+	objectsForNodes := WeakIdentityKeyDictionary new.
+	nodesForObjects := WeakIdentityKeyDictionary new
```
dionisiydk, line 38: Do you have a test which shows that weak part is really working?
I did little example in workspace and links are not collected for me when I close everything.

StevenCostiou: Indeed there is a problem. It is a separate issue, I just opened https://github.com/pharo-project/pharo/issues/6575

StevenCostiou:
> I think you also need to change MetaLinkRegistry>>#initialize

Thanks for the review :)
For the registry, for now I don't think it is necessary as we store objects that we control (metalinks, classes, etc.) that are not in the case of classes concerned by this PR (such as `Set`).

MarcusDenker (approved):

PR #6577, opened 2020-06-12
Slots-move-extensions-to-classes

MarcusDenker:
We are step by step moving the Slots package into the main System. This PR moves all the extension methods that where left to the classes that they extend. We already had moved (or directly created) many methods there, so these are actually not that many cases that needed to be fixed.


dionisiydk (approved):
you need to resolve a conflict

PR #6579, opened 2020-06-12
Deprecate-firstPrecodeCommentFor

MarcusDenker:
- transforming deprecation for #firstPrecodeCommentFor: (yes, the nil case is not taken into account, but it should be ok)
- deprecate #precodeCommentOrInheritedCommentFor: and #supermostPrecodeCommentFor: without replacement. Leftover of old tools that are not there anymore

The only other sender is fixed by another PR already in review.


guillep:
Ok!

PR #6582, opened 2020-06-12
Introduce-isValidSelector

MarcusDenker:
There are lots of references to RBScanner to check if a string or symbol is a valid selector

- add isValidSelector to String
- add test
- rewrite all senders


dionisiydk (approved):
Good change

PR #6586, opened 2020-06-13
TemporaryVariable: implement #isReferenced

MarcusDenker:
#isReferenced for TemporaryVariables need to check if the var is really accessed. This implementation does it using the AST: if there are no astNodes for this variables, it is not used.

- Implement isReferenced
- add test
- move implementation isReferenced from Variable to Slo
- add isReferenced as subclassResponsability in Variable

### DIFF
--- a/src/Kernel/Slot.class.st
+++ b/src/Kernel/Slot.class.st
@@ -232,6 +232,11 @@ Slot >> isReadIn: aCompiledCode [
 ]
 
 { #category : #testing }
+Slot >> isReferenced [
+	^self usingMethods isNotEmpty
+]
+
+{ #category : #testing }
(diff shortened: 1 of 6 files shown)


dionisiydk (approved):

dionisiydk:
Now I think it is wrong:
-  #usingMethods are methods referencing the variable. They should be in synch with isReferenced implementation.

Therefore we should fix #usingMethods which I introduced. It will fix other senders (not only isReferenced)

MarcusDenker:
:-) yes, you are right! But nevertheless this is a step in the right direction (in the sense of having the test and it is green). So we can merge it and then fix #usingMethods.

Issue #5906, opened 2020-03-16
Issue with Monticello in clean Pharo 8 image


aebutcher:
Hello, beginner here. I could not find this problem reported anywhere else, please forgive me if has already been reported.
I'm running MacOS Catalina 64bit.  Working from a clean Pharo 8 image.
When trying to save a package to a local repository, an error occurs while typing in Monticello's Log Message dialog.  It happens while typing.  If I close the error, and try to save, it does not save.
What I have been doing is typing my log message into a Playground window, copying that and pasting into the Log Message dialog, then saving, which is successful.
I have attached a screenshot of the error.
I raised this issue in the Pharo Discord MOOC server, and was advised to report it here. I don't know if this is a bug or just an issue on my computer?
![Screenshot 2020-03-16 at 11 39 28](https://user-images.githubusercontent.com/57952051/76745010-ce6c5300-677d-11ea-8c3f-83f3d4789f6f.png)

Ducasse:
Thanks for the report. This is strange we will have a look. 
Can you tell us what you did exactly? You save using the Monticello browser?

aebutcher:
Yes, I save to a local repository using Monticello, and the error seems to arise only while I am trying to type a log message.

bpieber:
I just ran into the same issue with a new image created by Pharo Launcher from template Pharo 8.0 - 64bit (stable).

crunsk:
I also got the same error message "MCSaveVersionDialog did not understand #selectedClassOrMetaClass" when I typed letters 't' and 'e'.
![example](https://user-images.githubusercontent.com/5631441/82759162-3da49a00-9df4-11ea-89f9-696c7360146d.png)
 to the save dialog.

I am running:
Windows 10 64-bit.
Pharo Launcher version: 2.0
Pharo image was created from 'Official distributions' template category Pharo 8.0 - 64bit (stable) 
After the image was created the 'Origin template' is 
Pharo 8.0 - 64bit (stable) 
(https://files.pharo.org/image/80/stable-64.zip)
![pharo_monticello_image](https://user-images.githubusercontent.com/5631441/82759088-8b6cd280-9df3-11ea-95d2-3b565f813f78.png)

<blockquote></blockquote>

PeterDNewton:
[...]

PR #6592, opened 2020-06-14
Proposed fix for: Issue #5906

PeterDNewton:
Proposes a fix for Issue #5906. It assumes that Monticello tools derived directly from MCTool are not intended as Code tools so their text areas shouldn't support code completion.

### DIFF
--- a/src/MonticelloGUI/MCCodeTool.class.st
+++ b/src/MonticelloGUI/MCCodeTool.class.st
@@ -183,3 +183,9 @@ MCCodeTool >> selectedMessageName [
 	"Answer the name of the selected message"
 	self subclassResponsibility
 ]
+
+{ #category : #'morphic ui' }
+MCCodeTool >> textMorph: aSymbol [
+	"The text areas of Code Tools should be customised for SmallTalk code"
+	^ self textMorph: aSymbol specializedFor: #beForSmalltalkScripting 
+]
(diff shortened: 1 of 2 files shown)


Ducasse:
I will relaunch the tests because some failures are strange

MarcusDenker:
I fixed something similar in the past, I need to check something, so please do not yet integrate.

(I do not think that we have to turn off code editing here, but we can fix it to work)

PeterDNewton:
Marcus, thank's for reviewing this proposed fix. For some background: currently some Monticello tools derive from the base class MCTool while others derive from MCCodeTool, my working assumption was that this is a deliberate split and not all tools should support code editing. If all tools should support code editing perhaps a different refactoring would be better.

MarcusDenker:
So I had a look... as this sounded so familiar. 

It seems that there is already a fix in Pharo8, but stable does not have it, it is in the "latest" version.

What I did was to just implement

MCSaveVersionDialog>>textMorph: aMorph
	"here the text field is not used for code editing"
 	^(super textMorph: aMorph) beForPlainText

It seems that we need to do a new release of Pharo 8 so that this fix is promoted...

MarcusDenker:
I added an Issue to make a new release: https://github.com/pharo-project/pharo/issues/6601

We need to check, too, if the latest image does not need to be more prominent in the Launcher.

PeterDNewton:
Thanks Marcus. This problem also exists in Pharo 9.0 stable release, will this change get included in 9.0 as well?

MarcusDenker:
I an honestly not re-create that in Pharo9... If I am un the Monticello "Edit Versio Name and Log Message", I can type without errors.

PR #6594, opened 2020-06-14
6593-SpWorldPresenterTest-should-not-have-side-effects-after-running

VincentBlondeau:
fix 6593


Ducasse:
Yes thanks vincent. Test side effects are annoying.

Issue #6568, opened 2020-06-11
Metaclass binding of Kernel classes are Associations, not Variables


MarcusDenker:
**Describe the bug**
All "bindings" of MetaClasses should be Variable sub-instances (e.g. GlobalVariable. 

Object class binding class   -> Association
ClyAbstractClassExample class binding class  -> LiteralVariable

I think this happens because the boo trap put Associations there.

PR #6598, opened 2020-06-14
[Experiment] Use Variable's directly instead of OCVariable wrappers

dionisiydk:
PR replaces all references of OCSlotVariable, OCLiteralVariable and OCUndeclaredVariable to use their actual variables (subclasses of Variable) which they represent (and actually wrap).
Old OC classes are deprecated to allow migration


MarcusDenker:
problem has to do with Undeclared somehow:

[64] NewUndeclaredWarning: PharoCommandLineHandler>>changePreferences (StartupPreferencesLoader is Undeclared)
[64] NewUndeclaredWarning: PharoCommandLineHandler>>changePreferences (StartupPreferencesLoader is Undeclared)
[64] Error: PharoCommandLineHandler>>#changePreferences: #StartupPreferencesLoader->nil
[64] OCASTSemanticAnalyzer(Object)>>error:
[64] OCASTSemanticAnalyzer>>visitVariableNode:

dionisiydk:
The issue is with UndeclaredVariable unification. I reverted this part of changes and build completes (tests are related but it should be easy to fix)..

MarcusDenker:
I think the Bootstrap put Undeclareds as Associations (similar to the Metaclass bindings, I added an issue for those here: https://github.com/pharo-project/pharo/issues/6568 )

dionisiydk:
> I think the Bootstrap put Undeclareds as Associations (similar to the Metaclass bindings, I added an issue for those here: #6568 )

It needs to be fixed. But the current issue is more tricky. UndefiledVariable was there for a long time and build did not fail. 
Interesting that in last saved image (I played with build artifacts) the Undeclared dict includes UndeclaredVariable instance but the problem method is compiled with association. So there are two different binding instances for same undeclared variable.
I think something is just missing in my changes. And current bootstrap will work when it fixed. (that's my hope :))

dionisiydk:
So I found and solve issue with broken build:
Current bootstrap process uses simple associations to represent vars in Undeclared registry. 
It breaks introduced logic here which explicitly return existing variable from Undeclared registry:
- compiler parts expect kind of variable and not a simple association.

Previously semantic analyzer always returned new instance OCUndeclaredVariable independently from what is registered in Undeclared dictionary. Now it is always the same registered object. So it must be a kind of variable.
Therefore the fix for new logic is to convert found registered Association instance into UndeclaredVariable (using #becomeForward).
When bootstrap will be updated we will remove this trick.

dionisiydk:
Now I need to fix tests

dionisiydk:
moved to new PR

PR #6599, opened 2020-06-15
Feature/rewrite tool basic editor

jordanmontt:
rewrite_basic_editor
This is a basic editor for create, save, load and apply custom rewrite rules. The rules can be applied to one or many classes.
Is like RewriteRuleEditor in NautilusRefactoring-Utilities package tool but with more functionalities. It uses its RewriteRuleChangeBrowser class.


StephanEggermont:
The RewriteRuleEditor does not store the rewrite rule it applies. After applying this PR the two should probably be merged. Add an ‘Apply directly’ button for that?

MarcusDenker:
ReleaseTest>>testUndeclared fails 

Found undeclared references: #(#RewriteRuleChangesBrowser)

PR #6602, opened 2020-06-15
#copyFrom: in ClassOrganization should not be extension method

MarcusDenker:
#copyFrom: in ClassOrganization is an extension method from the ClassBuilder. This is not needed, it can just be part of the Kernel package.


guillep:
Ok!

PR #6603, opened 2020-06-15
6504-slotNames-is-defined-in-ClassDescription-but-invoked-from-Behavior

### LINKED ISSUE #6504: slotNames is defined in ClassDescription but invoked from Behavior
Behavior >> definesSlotNamed: aString
	"Return true whether the receiver defines an instance variable named aString."
	
	^ self slotNames includes: aString

MarcusDenker:
- move #definesSlotNamed: to ClassDescription like al the other slot methods
- take this chance to redo the PR to move all extension methods of Slot class to the extended classes

fixes #6504


src/Collections-Strings/String.class.st
```diff
@@ -320,6 +320,11 @@ String class >> newUppercasingTable [
 	^ String withAll: (Character allByteCharacters collect: [:c | c asUppercase])
 ]
 
+{ #category : #accessing }
+String class >> pseudovariablesNames [
+	^#('self' 'true' 'false' 'nil' 'thisContext' 'super')
+]
```
guillep, line 326: Why is this in String?

src/Collections-Strings/String.class.st
```diff
@@ -526,6 +531,11 @@ String >> asCamelCase [
                        stream nextPutAll: sub capitalized]]
 ]
 
+{ #category : #variables }
+String >> asClassVariable [
+	^ self asSymbol asClassVariable
+]
+
```
guillep, line 538: Shouldn't this be an extension method?

src/Collections-Strings/String.class.st
```diff
@@ -673,6 +683,22 @@ String >> asSignedInteger [
 	^ Integer readFrom: stream
 ]
 
+{ #category : #variables }
+String >> asSlot [
+	^ self asSymbol asSlot
+]
```
guillep, line 689: Should this be an extension method?

src/Collections-Strings/String.class.st
```diff
@@ -673,6 +683,22 @@ String >> asSignedInteger [
...
+{ #category : #variables }
+String >> asSlotCollection [
+	"Parse as slots. Use space, tab and cr as separators
+	Example: 
+	  'a b' asSlotCollection --> {a => Slot. b => Slot}
+	"
+	| whitespaces |
+	whitespaces := ByteString withAll: { Character space . Character tab . Character cr }.
+	^(self substrings: whitespaces) collect: [ :substring | substring asSlot ]
+]
```
guillep, line 700: Same, Should this be an extension method?

src/Collections-Strings/String.class.st
```diff
@@ -1857,6 +1883,14 @@ String >> isPatternVariable [
 	 ^self keywords anySatisfy: [:each | each first = $`]
 ]
 
+{ #category : #testing }
+String >> isPseudovariableName [
+	"Answer true if I am a pseudo-variable name.
+	#self isPseudovariableName -> true
+	"
+	^ self class pseudovariablesNames includes: self
+]
```
guillep, line 1892: Why is this in String? The only user is Slot. This should be moved to the Slot class...

```smalltalk
aSymbol isPseudovariableName ifTrue: [ 
		^ InvalidSlotName signalFor: aSymbol ].
```

```smalltalk
(self isPseudovariableName: aSymbol) ifTrue: [ 
		^ InvalidSlotName signalFor: aSymbol ].

```

I don't see the value of this being in `String`

src/Collections-Strings/Symbol.class.st
```diff
@@ -410,12 +410,25 @@ Symbol >> = aSymbol [
 	^ super = aSymbol
 ]
 
+{ #category : #variables }
+Symbol >> => aVariable [
+	"If the slot we give as argument is not present in the image, we will get a nil. In that case we should throw an explicit error to the user saying a slot is missing."
+
+	aVariable ifNil: [ SlotNotFound signalForName: self ].
+	^ aVariable named: self
+]
```
guillep, line 419: Why is this in `Symbol`? this should be at least in an extension method.

This is not good packaging...

src/Collections-Strings/Symbol.class.st
```diff
@@ -410,12 +410,25 @@ Symbol >> = aSymbol [
...
 { #category : #converting }
 Symbol >> asAnnouncement [
 
 	^ self
 ]
 
+{ #category : #variables }
+Symbol >> asClassVariable [
+	^ ClassVariable named: self.
+]
```
guillep, line 430: Same same

src/Collections-Strings/Symbol.class.st
```diff
@@ -432,6 +445,11 @@ Symbol >> asMutator [
 	^ (self copyWith: $:) asSymbol
 ]
 
+{ #category : #variables }
+Symbol >> asSlot [
+	^ InstanceVariableSlot named: self.
+]
+
```
guillep, line 452: Same.

src/Collections-Unordered/Dictionary.class.st
```diff
@@ -391,6 +391,30 @@ Dictionary >> declare: key from: aDictionary [
...
+		self add: newGlobal.
+		].
+	(aDictionary includesKey: globalName)
+		ifTrue:  [
+			self add: ((aDictionary associationAt: globalName) primitiveChangeClassTo: ClassVariable new).
+			aDictionary removeKey: globalName]
+		ifFalse: [
+			self add: newGlobal]
+]
+
```
guillep, line 417: Again. I don't see why this should be in Dictionary. This is no general use of dictionary.
Either we need a special "variable" dictionary. Or all this logic should be in the client.

src/System-Support/SystemDictionary.class.st
```diff
@@ -170,6 +170,22 @@ SystemDictionary >> classOrTraitNamed: aString [
...
+
+	(self includesKey: key) ifTrue: [^ self].
+	(aDictionary includesKey: key)
+		ifTrue: 
+			[self add: ((aDictionary associationAt: key) primitiveChangeClassTo: GlobalVariable new).
+			aDictionary removeKey: key]
+		ifFalse: 
+			[self add: (GlobalVariable key: key)]
+]
+
```
guillep, line 188: This looks pretty similar to `Dictionary >> declareVariable: newGlobal from: aDictionary` no?

guillep (changes requested):
I think all changes related to `ClassDescription`, `Behavior`, closures are ok.

However, all the String related stuff is strange and pollutes the String API, for something not really useful.

MarcusDenker:
Thanks! The only problem now I that I can not load the PR...  changes are done in an order that is bad (it remove a method that it then needs?)

I will redo it step by step with multiple PRs taking the review into account (I let it open for that)

Ducasse:
Thank marcus. I agree with the comments of Guille. I try to avoid to have these as* on Symbol and String because it means that we cannot dispatch.

MarcusDenker:
I added everything to my backlog to be done step by step, I close this as it will soon be conflicting with other merged PR.

PR #6625, opened 2020-06-16
Added rewrite rule basic editor.

jordanmontt:
This is related to https://github.com/pharo-project/pharo/pull/6599
I was getting a weird error so I moved RewriteRuleChangeBrowser to NautilusRefactoring-Utilities package and done the PR again.
The next step is to add an ExampleBasedEditor (you can build a rule based on Pharo's code) and to integrate this tool with [MatchTool](https://github.com/jordanmontt/MatchTool-Spec2).

rewrite_basic_editor
This is a basic editor for create, save, load and apply custom rewrite rules. The rules can be applied to one or many classes.
Is like RewriteRuleEditor in NautilusRefactoring-Utilities package tool but with more functionalities. It uses its RewriteRuleChangeBrowser class.

### DIFF
--- a/src/RewriteCustomRules/CustomRuleGenerator.class.st
+++ b/src/RewriteCustomRules/CustomRuleGenerator.class.st
@@ -1,0 +1,106 @@
+"
+I am in charge of saving the RewriteRules. I receive the custom rule as plain text and save it in form of a class. 
+"
+Class {
+	#name : #CustomRuleGenerator,
+	#superclass : #Object,
+	#instVars : [
+		'input',
+		'output',
+		'searchFor',
+		'replaceWith',
+		'protocolName',
+		'ruleName'
+	],
+	#category : #'RewriteCustomRules-Core'
+}
+
+{ #category : #testing }
+CustomRuleGenerator >> areRuleAndProtocolNamesValid [
+	^ protocolName isNotNil
+		and: [ protocolName isValidGlobalName
+				and: [ ruleName isNotNil and: [ ruleName isValidGlobalName ] ] ]
+]
+
+{ #category : #defaults }
+CustomRuleGenerator >> classNotCreatedAlert [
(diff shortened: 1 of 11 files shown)


Ducasse:
Hi sebastian 
I will have a look. I think that indeed it should be package in its own package because Nautilus refactoring should be removed. 
I may integrate and after repackage it. 
S

Ducasse:
- I do not understand what are protocol? If this is a package name please call it package name. 
- can you extract the class creation in a separate method. 
- I would not use directly UIManager default to raise a popup. The domain should raise a signal and the application should do something. The rule creation has nothing to do with UI so it should be layered. So we should rethink this part. The generator should be called from the UI. And this is the UI (presenter) that should prompt the user. 
- Please add printOn: to you class since it helps debugging. 
- the default template for the expression should not hardcoded in the methods. You can have a default but we should be able to change them without recompiling the method.
- Finally and really important, you should write some tests for the domain and presenter.

jordanmontt:
>     * I do not understand what are protocol? If this is a package name please call it package name.
> 
>     * can you extract the class creation in a separate method.
> 
>     * I would not use directly UIManager default to raise a popup. The domain should raise a signal and the application should do something. The rule creation has nothing to do with UI so it should be layered. So we should rethink this part. The generator should be called from the UI. And this is the UI (presenter) that should prompt the user.
> 
>     * Please add printOn: to you class since it helps debugging.
> 
>     * the default template for the expression should not hardcoded in the methods. You can have a default but we should be able to change them without recompiling the method.
> 
>     * Finally and really important, you should write some tests for the domain and presenter.

Hi Stef,
Thank you for your feedback. I will separate the class creation from UI and change the protocol for package-name. Also, I will move RewriteRuleChangeBrowser from Nautilus. Write some test and work on the others observations as well.

Ducasse:
So I propose to close this issue and that you reopen one or multiple later.

PR #6627, opened 2020-06-16
Deprecate SmalltalkEditingState

### LINKED ISSUE #6618: Deprecate SmalltalkEditingState
the class SmalltalkEditingState is not referenced (neither directly nor in source code)

I guess we can deprecate it and move it to "Deprecated90" package

SIDE NOTE: This should be done after integrating https://github.com/pharo-project/pharo/pull/6617
which is moving the class to a differen tag

astares:
Fix #6618


MarcusDenker (approved):
ok

guillep:
Thanks!

PR #6632, opened 2020-06-18
use-OCBytecodeToASTCache-instead-of-fullBlockInstructionForPC

MarcusDenker:
This PR removes #fullBlockInstructionForPC: and instead uses the OCBytecodeToASTCache in this case just as we use it for Methods.

OCBytecodeToASTCache has been slightly changed that it can build the cache for both cases

### DIFF
--- a/src/OpalCompiler-Core/IRMethod.class.st
+++ b/src/OpalCompiler-Core/IRMethod.class.st
@@ -159,16 +159,6 @@ IRMethod >> forceLongForm [
 		
 ]
 
-{ #category : #debugging }
-IRMethod >> fullBlockInstructionForPC: aPC [
-	| pc |
-	"For a given PC, the actual instruction may start N bytes ahead."
-	pc := aPC.
-	[ pc >= 0 ] whileTrue: [ 
-		(self firstInstructionMatching: [:ir | ir bytecodeOffset = pc ]) ifNotNil: [:it |^it].
-		pc := pc - 1 ].
-]
-
 { #category : #translating }
 IRMethod >> generate [
 	^self generate: CompiledMethodTrailer empty
--- a/src/OpalCompiler-Core/OCBytecodeToASTCache.class.st
+++ b/src/OpalCompiler-Core/OCBytecodeToASTCache.class.st
@@ -58,19 +58,19 @@ OCBytecodeToASTCache >> firstBcOffset [
 
 { #category : #initialization }
 OCBytecodeToASTCache >> generateForNode: aNode [
-	| blockIR currentBcOffset |
+	| methodOrBlockIr currentBcOffset |
 	methodOrBlockNode := aNode.
-	blockIR := aNode ir.
+	methodOrBlockIr := aNode ir.
 	bcToASTMap := Dictionary new.
-	firstBcOffset := blockIR compiledMethod initialPC.
+	firstBcOffset := methodOrBlockIr compiledMethod initialPC.
 
 	"if there is a primitive, we need to record the offset"
-	(blockIR irPrimitive num ~= 0
-		or: [ blockIR compiledMethod isQuick ])
+	(methodOrBlockIr irPrimitive num ~= 0
+		or: [ methodOrBlockIr compiledMethod isQuick ])
 		ifTrue: [ bcToASTMap at: firstBcOffset put: methodOrBlockNode ]. 
 
 	currentBcOffset := firstBcOffset.
-	blockIR startSequence withAllSuccessors
+	methodOrBlockIr startSequence withAllSuccessors
 		do: [ :seq | 
 			seq
 				do: [ :ir | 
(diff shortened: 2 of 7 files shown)


MarcusDenker:
Failing tests not related. They have been introduced by a previous PR

src/OpalCompiler-Core/OCBytecodeToASTCache.class.st
```diff
@@ -57,20 +57,20 @@ OCBytecodeToASTCache >> firstBcOffset [
 ]
 
 { #category : #initialization }
-OCBytecodeToASTCache >> generateForMethodNode: aNode [
-	| methodIR currentBcOffset |
-	methodNode := aNode.
-	methodIR := methodNode ir.
+OCBytecodeToASTCache >> generateForNode: aNode [
+	| blockIR currentBcOffset |
```
StevenCostiou: `methodIR` was replaced by `blockIR`. Shouldn't it be `methodOrBlockIR`?
(long name I agree)

src/OpalCompiler-Core/IRMethod.class.st
```diff
@@ -159,16 +159,6 @@ IRMethod >> forceLongForm [
...
 
-{ #category : #debugging }
-IRMethod >> fullBlockInstructionForPC: aPC [
-	| pc |
-	"For a given PC, the actual instruction may start N bytes ahead."
-	pc := aPC.
-	[ pc >= 0 ] whileTrue: [ 
-		(self firstInstructionMatching: [:ir | ir bytecodeOffset = pc ]) ifNotNil: [:it |^it].
-		pc := pc - 1 ].
-]
```
StevenCostiou, line 170: Side note: maybe we should reuse this code in tests to validate the cache? I do not remember if we actually moved this in the cache or if it was just removed.

src/OpalCompiler-Core/OCBytecodeToASTCache.class.st
```diff
@@ -57,20 +57,20 @@ OCBytecodeToASTCache >> firstBcOffset [
...
-		or: [ methodIR compiledMethod isQuick ])
-		ifTrue: [ bcToASTMap at: firstBcOffset put: methodNode ]. 
-		
+	(blockIR irPrimitive num ~= 0
+		or: [ blockIR compiledMethod isQuick ])
+		ifTrue: [ bcToASTMap at: firstBcOffset put: methodOrBlockNode ]. 
+
 	currentBcOffset := firstBcOffset.
-	methodIR startSequence withAllSuccessors
+	blockIR startSequence withAllSuccessors
```
StevenCostiou: I am just wondering: we can seamlessly use the same algorithm for methods and full blocks because full blocks are also methods without anything specific? We do not need any specific test to validate it?

StevenCostiou (commented):
Is there a test with a fullblock where we check for the mapping?

src/OpalCompiler-Core/OCBytecodeToASTCache.class.st
MarcusDenker: The branch in #sourceNodeForPC: on block node that calls it is executed 77 times while running the Opal tests suite and >60 times when running all debugger tests.

It is covered by existing tests.

src/OpalCompiler-Core/IRMethod.class.st
MarcusDenker: the OCBytecodeToASTCache iterates over all the IR and fills the cache. So something like this could be used for a regression test... but i would prefer to not have this method as part of the API of IRMethod.

src/OpalCompiler-Core/OCBytecodeToASTCache.class.st
MarcusDenker: of course more test would be nice

src/OpalCompiler-Core/OCBytecodeToASTCache.class.st
MarcusDenker: yes!

MarcusDenker:
I added a test and fixed the name of the temp var. 

if tests are green, this is ready to merge

StevenCostiou (approved):

MarcusDenker:
strange failing test which does not fail on my machine. close and reopen to merge with latest code

MarcusDenker:
So the test fails in the image. I download it, it fails. I edit it, compile, it works.

I have no idea what the problem is

StevenCostiou:
I'm having a look, I don't think I'll go though it today, but if you recompile the method it recomputes the cache. So something is wrong with the original compiled code?

StevenCostiou:
ah, no, when recompiling the test...

StevenCostiou:
actually executing two times the test makes it pass... strange...

StevenCostiou:
Recomputing the cache a second time is what makes the test pass, so the first time it is wrong or miscomputed

StevenCostiou:
Good case for the sprint: first time it computes the cache it is the method cache that is returned (wrong) and the second it is the block cache (looks better, test pass).

StevenCostiou:
@dupriezt Possibly an interesting bug for the Swarm Experiment

MarcusDenker:
Now it works. Reason: we have to call #ir once on the mode before, else it does not know it is a full block.

This is bad and will be fixed soon (it should be tagged as part of semantic analysis instead)

PR #6633, opened 2020-06-18
6544-ReImplementedNotSentRule-has-false-positives

MarcusDenker:
This PR changes how the rule work in two ways:

- The "fast" way does not work anymore as we would need to check all nested blocks, too.
- We use #messages instead. This is slow, so we change when we update:
     - when a method is added or changed, we add the messages of the new method
     - if a method is removed, we do nothing

This means that this new scheme slowly accumulates false negatives. But every image startup will make sure the data is completely recalculated.

- I added a comment in the rule description of this and how to reset on demand.

NOTE: this is not meant to be the perfect fix! There is much more we can do (e.g. the data should be  moved to the environment, maybe even with some process updating it in the background. But that can be the next step)

### DIFF
--- a/src/Calypso-SystemPlugins-Critic-Queries-Tests/ClyClassWithProblemMethods.class.st
+++ b/src/Calypso-SystemPlugins-Critic-Queries-Tests/ClyClassWithProblemMethods.class.st
@@ -11,6 +11,7 @@ Class {
 ClyClassWithProblemMethods >> methodWithHalt [
 	<haltOrBreakpointForTesting>
 	self halt.
+	self methodWithHalt2.
 ]
 
 { #category : #'method examples' }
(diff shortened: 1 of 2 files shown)


MarcusDenker:
ClyGroupedCritiquesTest>>testFillingWithTwoSimilarCritiques is failing

MarcusDenker:
Other properties of this solution that need to be taken into account for deciding:
- the rule now is about *real* sends, not symbol use
- new/updated method used to do just reset to nil, they now call #messages on that one method (thus slightly different performance characteristics for method compilation)

MarcusDenker:
On the positive side: the old way meant that in the browser, after accept the data was always recalculated for the whole image. This change does it just for that one method.

tesonep:
Your fix is right, but I am hesitating about the whole implementation of the Rule. For this rule, we are penalizing the whole execution of the system (for example when loading a big Metacello baseline with thousands of methods). I think we should disable the rule by default or think in a better implementation for it.

MarcusDenker:
Ok, I merge this and then I do a PR to disable the rule.

Issue #5538, opened 2020-01-21
Reduce the number of allInstancesDo: and allInstances


Ducasse:

Ducasse:
allInstancesDo: should use sparingly if not!

Ducasse:
I checked them all.


onFontsChanged

	StringMorph allSubInstancesDo: [ :s | s layoutChanged ].
	TextMorph allSubInstancesDo: [ :s | s layoutChanged ].
	SystemWindow allInstancesDo: [ :w | [ w update: #relabel ] on: Error do: [ :ex | ] ].
	self currentWorld ifNotNil: [ :w | w changed ].


startUp: isImageStarting
	"Reset all instances on startup since the cached pages (athens forms) don't work in a new session."

	isImageStarting
		ifTrue: [ self allInstancesDo: #reset ]

would be nice at startup to avoid to use in addition a slow iterator. 
      isImageStarting
		ifTrue: [ self allInstancesDo: [ :each | each reset ] ]

Ducasse:
Why UIFinder class is doing this

UIFinder class >> searchedTextListMaxSize: anInteger

	self allInstancesDo: [:each | each searchedTextListMaxSize: anInteger].
	searchedTextListMaxSize := anInteger

Ducasse:
runCase
	| result |
	"Skip the test if we're in headless mode"
	self flag: 'Use skip feature once it is ready'.
	Smalltalk isHeadless
		ifTrue: [ ^ self skip ].
	SystemProgressMorph allInstances do: [  :spm | spm hide ].
	self currentWorld activeHand resetClickState.
	result := super runCase .
	SystemProgressMorph allInstances do: [  :spm | spm show ].
	^ result

PR #6635, opened 2020-06-18
PharoCommonTools-noAllInstancesDo

MarcusDenker:
This PR removes one #allInstancesDo: from the image shutdown code. (see issue #5538, which is not fixed by this, it's just one tiny step)

### DIFF
--- a/src/Tool-Registry/PharoCommonTools.class.st
+++ b/src/Tool-Registry/PharoCommonTools.class.st
@@ -104,7 +104,8 @@ PharoCommonTools class >> settingsOn: aBuilder [
 
 { #category : #'system startup' }
 PharoCommonTools class >> shutDown: aboutToQuit [
-	self allInstancesDo: #cleanUp
+	Smalltalk image tools cleanUp.
+	Smalltalk image resetTools
 ]
 
 { #category : #'world menu' }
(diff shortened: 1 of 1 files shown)


MarcusDenker:
SystemSettingsPersistenceTest>>testDefaultImageDoesNotStoreAnySetting

is failing and it is related

src/Tool-Registry/PharoCommonTools.class.st
```diff
@@ -104,7 +104,8 @@ PharoCommonTools class >> settingsOn: aBuilder [
 
 { #category : #'system startup' }
 PharoCommonTools class >> shutDown: aboutToQuit [
-	self allInstancesDo: #cleanUp
+	Smalltalk image tools cleanUp.
+	Smalltalk image resetTools
```
dionisiydk: I am curious when #image was introduced and why? I see that it is same as Smalltalk:
```Smalltalk
Smalltalk image == Smalltalk "==true"
```

dionisiydk (approved):

src/Tool-Registry/PharoCommonTools.class.st
MarcusDenker: I think the idea was to split that eventually.

I would actually like to add a reserved variable "thisImage" to talk to the image.

PR #6638, opened 2020-06-18
Use Variable's directly instead of OCVariable wrappers

dionisiydk:
1.  All references of OCSlotVariable, OCLiteralVariable and OCUndeclaredVariable are replaced by  their real variables (subclasses of Variable) which they represent (and actually wrap).
Old OC classes are deprecated to allow migration.
2. OCAbstractLocalVariable is direct subclass of Variable. Old superclass OCAbstractVariable is deprecated

It is continuation of original PR #6598. 
It can't be just loaded into the image because it changes structure of compilation related classes. For review download image from CI


MarcusDenker:
Nice! This means we can merge OCTempVariable and TemporaryVariable (and use that name, too).

In addition, we do not need the "#saveTemp:" mechanism anymore, as the temp object will exist as long as the (name analysed) AST exists.

MarcusDenker (approved):
I read the code and looked at the build artefact image.

This seems to work very well! There are some more step to be done but everything seem to work!

MarcusDenker:
These are things that can be done in a second step:

- OCAbstractLocalVariable misses #isReferenced and #usingMethods
- TemporaryVariable and OCAbstractLocalVariable  can be merged
- OCVARIABLE category cleanup
- ClyVariable hierarchy can now be merged as we now have #definingClass

dionisiydk:
Yes. I think it can be merged. 
Notice the protocol OCVariable for methods from OC variables. I will do cleanup in another PR. I would like to do some renames and deprecation there. So it is better to separate these changes.

For review here: one interesting change I did to fix tests is users of RFGlobalRead/Write: 
 https://github.com/pharo-project/pharo/pull/6638/commits/d646ad7b775faff7fc44f4250aecbfcdac29cf56. Notice "RBParser parseExpression: 'RFGlobalRead new" users. 
I think this is a good change. But please check it from your side.

dionisiydk:
> These are things that can be done in a second step:
> 
> * ClyVariable hierarchy can now be merged as we now have #definingClass

We do not have #definingClass in ClassVariable. And I think it is something missing for proper traits integration (stateful trait with class var)

MarcusDenker:
The  "RBParser parseExpression: 'RFGlobalRead new" users look good to me

dionisiydk:
It looks like CI became broken after this merge. 
Probably revert is a good idea. Let's see if it will really help.

PR #6639, opened 2020-06-19
Deprecation-transform-execute-transformedAST

MarcusDenker:
The transforming deprecations rely on the deprecated method to call the new method (the transformation is just a side-effect).

But we can do better:

- we can return from the context here, this will then not execute the rest of the deprecated method

-  as we can evaluate ASTs in a context with #evaluateForContext:, we can "run" the AST of the transfromed expression (this is the message send node that 
triggered the deprecated method to be executed which now has been rewritten)

### DIFF
--- a/src/Kernel/Deprecation.class.st
+++ b/src/Kernel/Deprecation.class.st
@@ -228,7 +228,7 @@ Deprecation >> signal [
 
 { #category : #handling }
 Deprecation >> transform [
-	| node rewriteRule aMethod |
+	| node transformedNode rewriteRule aMethod |
 	self shouldTransform ifFalse: [ ^ self signal].
 	self rewriterClass ifNil:[ ^ self signal ].
 	aMethod := self contextOfSender compiledCode method.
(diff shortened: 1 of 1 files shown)


src/Kernel/Deprecation.class.st
```diff
@@ -239,12 +239,13 @@ Deprecation >> transform [
...
 			ifFalse: [ ^ self signal ].
-		node replaceWith: rewriteRule tree. 
+		node replaceWith: (transformedNode := rewriteRule tree). 
 		Author 
 			useAuthor: 'AutoDeprecationRefactoring'
 			during: [aMethod origin compile: aMethod ast formattedCode classified: aMethod protocol].	
 		Log  ifNotNil: [:log | log add: self].
-		self logTranscript]
+		self logTranscript].
+	context return: (transformedNode evaluateForContext: context)
```
guillep, line 248: Will not this re-evaluate the receiver and arguments? That would be wrong in case they have side effects... I think doing this properly is really complicated. We should diff the receiver/arguments nodes pre-rewriting with the rewritten ones, take the ones from the stack that match, and re-evaluate the ones that do not match. All this in the proper order and taking into account that arguments can be added and removed in addition of just modified.

guillep (changes requested):

src/Kernel/Deprecation.class.st
guillep: This complexity with no tests is a no go....

guillep: And even! if the receiver/arguments do not match, nothing guarantees that a side effect in the rewritten nodes will overlap some other side effect of the previously executed ones.

guillep: I think the best is to just restart the execution from some safe point

MarcusDenker: Yes, true. This only works for the trival case (receiver and args are not expressions)

PR #6640, opened 2020-06-19
Object centric halt on call breakpoint

StevenCostiou:
Improving `Breakpoint` to support object-centric debugging.

### DIFF
--- a/src/Reflectivity-VariableBreakpoint/VariableBreakpoint.class.st
+++ b/src/Reflectivity-VariableBreakpoint/VariableBreakpoint.class.st
@@ -193,6 +193,7 @@ VariableBreakpoint >> remove [
 
 { #category : #api }
 VariableBreakpoint >> scopeTo: anInstance [
+	super scopeTo: anInstance.
 	self condition: [ :ctx | ctx receiver == targetInstance ]
 ]
 
--- a/src/Reflectivity-VariableBreakpoint/VariableBreakpoint.class.st
+++ b/src/Reflectivity-VariableBreakpoint/VariableBreakpoint.class.st
@@ -50,7 +50,7 @@ VariableBreakpoint class >> notifyBreakpointHit: aBreakpoint inContext: aContext
 ]
 
 { #category : #API }
-VariableBreakpoint class >> watchVariable: aVariableName in: aClassOrCompiledMethod [
+VariableBreakpoint class >> watchVariable: aVariableName inClass: aClassOrCompiledMethod [
 	^ (aClassOrCompiledMethod newBreakpointForVariable: aVariableName)
 		install
 ]
@@ -63,7 +63,7 @@ VariableBreakpoint class >> watchVariable: aVariableName inObject: anObject [
 ]
 
 { #category : #API }
-VariableBreakpoint class >> watchVariableReads: aVariableName in: aClassOrCompiledMethod [
+VariableBreakpoint class >> watchVariableReads: aVariableName inClass: aClassOrCompiledMethod [
 	^ (aClassOrCompiledMethod newBreakpointForVariable: aVariableName)
 		accessStrategy: #read;
 		install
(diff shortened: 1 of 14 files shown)


src/Reflectivity-VariableBreakpoint/VariableBreakpointTest.class.st
```diff
@@ -328,22 +328,22 @@ VariableBreakpointTest >> testInitialize [
 { #category : #'tests - installation' }
 VariableBreakpointTest >> testInstall [
 	
-	wp := VariableBreakpoint newWatchPointInClass: VariableBreakpointMockSubclass.	
+	wp := VariableBreakpoint newVariableBreakpointInClass: VariableBreakpointMockSubclass.
```
dionisiydk: Can it be simple #newInClass: ? It is already VariableBreakpoint. Why duplicate this meaning in the method name?
And I think #new**For**Class: sound better because the method does not perform installation while  #newInClass: sounds like it does.

src/Reflectivity/Breakpoint.class.st
```diff
@@ -17,15 +17,16 @@ Class {
...
-		'link',
-		'node',
-		'level',
-		'options'
+		'#condition',
+		'#link',
+		'#node',
+		'#level',
+		'#options',
+		'#targetInstance => WeakSlot'
```
dionisiydk, line 25: Do you have a test which explains why it should be weak?
My first thought is that it is wrong. There is a registry of breakpoints in the system. And I would expect that these registry is a weak collection. So breakpoints and their targets will be collected as a garbage when nobody references target objects (considering that target objects always reference breakpoints by design)

src/Reflectivity-VariableBreakpoint/VariableBreakpoint.class.st
```diff
@@ -95,41 +94,41 @@ VariableBreakpoint class >> watchVariableWrites: aVariableName inObject: anObjec
 
 { #category : #API }
 VariableBreakpoint class >> watchVariablesIn: aClass [
```
dionisiydk: I think it is better to have explicit name everywhere:
- watchVariables**InClass**: aClass
- watchVariablesInObjet: anObject

src/Reflectivity-VariableBreakpoint/VariableBreakpointTest.class.st
StevenCostiou: I went for `newForClass`

src/Reflectivity-VariableBreakpoint/VariableBreakpoint.class.st
StevenCostiou: Right, done.
I think it's a mistake to do it wihtin this PR, because it's not related to the object-centric breakpoint improvement.
But too late now, there will be a bit of noise when reviewing the PR ;)

src/Reflectivity/Breakpoint.class.st
StevenCostiou: Target objects are not aware of their possible instrumentation by a breakpoint.
By design, tools only are aware, and it is perfectly ok to prevent garbage collection if the tools requires it.

Breakpoint are only here to break execution of the target object and they should not interfere with garbage collection, hence the weak slot.

I think it's fine to have ghost breakpoints that do nothing because they lost their target.
We could simply check if such breakpoint exist when adding a new breakpoint and clean the registry.

dionisiydk: Ok. But test is really important for this "weak logic".

StevenCostiou: There are 2 tests, look at first commit.
It's pretty basic, but what else should I test?

StevenCostiou:
I still do not get why my PR always endup in failing with stuff that seem unrelated.
For instance, there is that in red (which is scary): `\\sqAllocateMemorySegmentOfSizeAboveAllocatedSizeInto mmap: Cannot allocate memory`

Then there is test code that is shown (why?) and if ends up with `Sending interrupt signal to process`.
I have multiple different PR ending with the same problem.

I am the only one getting this problem?

dionisiydk:
I got it many times. It's really annoying

StevenCostiou:
I'm closing and reopening to try again. Is there another way to restart the CI?

MarcusDenker (approved):
ok

MarcusDenker:
tests look good now

PR #6641, opened 2020-06-19
Renaming Watchpoint to Watch

StevenCostiou:
The current `Watchpoint` is not a breakpoint but more a `Watch` that records data from a given point in the control flow. 
All instances of `Watchpoint` has been renamed to `Watch`.

A few times it makes namings a bit odd. I did not go over all temporary variables' names in methods that may still refer to the old `Watchpoint` naming.


dionisiydk (approved):

dionisiydk:
One comment.
It's well known names in debuggers world. But Watch word can definitely conflict with some domains.
I feel that it would be better to have reflectivity prefix  RF (or another) for all related classes (RBWatch, RFBreakpoint, RFMetalink, RFPermalink, ..).

StevenCostiou:
> I feel that it would be better to have reflectivity prefix RF (or another) for all related classes (RBWatch, RFBreakpoint, RFMetalink, RFPermalink, ..).

I agree, but I would only prefix the RF tools, MetaLink should stay MetaLink as it is part of Pharo now (as all the core RF). 
Can you open a separate issue about this? I believe it should be done separately of this PR (and we could discuss what should be renamed).

dionisiydk:
#6648

MarcusDenker (approved):

PR #6642, opened 2020-06-20
TemporaryVariable-Cleanup-saveTemp

MarcusDenker:
TemporaryVariable had a (not nice) mechanism to make them persistent.  It was not yet used outside of tests. The idea was to use it to support metalinks on temp defintions.

As we will merge TemporaryVariable and OCTemp, this is not needed anymore as those will be persitet with the AST.

There are not deprecations as the mechanism was never used.


dionisiydk (approved):
looks good

PR #6646, opened 2020-06-20
Introduce Collections-Abstract-Tests

### LINKED ISSUE #6645: Introduce Collections-Abstract-Tests
Introduce "Collections-Abstract-Tests" package as a first step to break the fat "Collections-Tests" package into 
smaller packages 

So we hopefully end up with more granular

   Collections-XXX
   Collections-XXX-Tests

structure

astares:
Fix #6645


MarcusDenker:
[64] MetacelloNameNotDefinedError: project group, or package named: 'Collections-Abstract' not found when used in requires: or includes: field of package: 'Collections-Abstract-Tests' for version: baseline of BaselineOfKernelTests.

astares:
One currently can not express the dependency from the test package to the regular package the way the baselines are structured. But this is OK for now. We can fix this ones all packages are stripped and the bootstrap is more clean.

astares:
Test failures unrelated

MarcusDenker:
The problem is that the CI run does not finish on all three platforms since Friday.

We have to fix that before we can merge anything.

MarcusDenker (approved):

PR #6647, opened 2020-06-20
Introduce #readFromContext: for all variables

dionisiydk:
- introduces #readInContext: for all variables and refactors #lookupSymbol:.
As result it fixes #lookupSymbol: to work with all kind of slots (it was only working for simple inst vars).
- introduce write:inContext for all variables.

Having pair of those methods we can get read and write any kind of variable having context instance.

### DIFF
--- a/src/Debugging-Core/Context.extension.st
+++ b/src/Debugging-Core/Context.extension.st
@@ -60,7 +60,7 @@ Context >> namedTempAt: index put: aValue [
 	"Set the value of the temp at index in the receiver's sequence of tempNames"
 	"NOTE: this list of temp names is completely virtual! It omits temp vectors but 
 	adds every variable that could be acccessed from a source perspective but might not be"
-	^self tempNamed: (self tempNames at: index) put: aValue
+	self tempNamed: (self tempNames at: index) put: aValue
 ]
 
 { #category : #'*Debugging-Core' }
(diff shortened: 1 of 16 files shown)


src/Debugging-Core/Context.extension.st
```diff
@@ -22,15 +22,13 @@ Context >> callPrimitive: primNumber [
 { #category : #'*Debugging-Core' }
 Context >> lookupSymbol: aSymbol [
```
dionisiydk, line 23: I think we should rename this method to something like #valueOfSymbol:. I can do it in another PR. Suggestions for name are welcome.

MarcusDenker: yes, the name is not good... maybe valueForVariableNamed? valueForSymbol:?

MarcusDenker (approved):
Good! Sadly the CI is down...

src/Debugging-Core/Context.extension.st
dionisiydk: and we need another name to modify the value. Basically they will be generalized version of tempNamed:/tempNamed:put:

dionisiydk: I think #read/#write are good words to use there:
- aContext readVarNamed: aSymbol
- aContext write: aValue toVarNamed: aSymbol

MarcusDenker:
testReifyAssignmentTempOperationBefore is failing

MarcusDenker:
the reason is that OCTempVariable #write:inContext:  misses a return at the end.

(writing a variable used to return the value.. no idea if that is bad or good)

PR #6651, opened 2020-06-21
CompilationContext-default-optionFullBlockClosure

MarcusDenker:
optionFullBlockClosure is enabled by the build, but the default options do not reflect that.

-> make optionFullBlockClosure default


dionisiydk (approved):

MarcusDenker:
we will do that later, needs some considerations with the bootstrap

PR #6653, opened 2020-06-21
Deprecate Variable>>#variable and move other deprecated code to Deprecated90 package

dionisiydk:
PR deprecates Variable>>#variable and move other deprecated code to Deprecated90 package.

PR is based on #6647 and actual changes are in two last commits


MarcusDenker:
Looks good. same test failing as #6647  (plus a timeout), supersedes #6647

PR #6657, opened 2020-06-23
use-OCTempVariable-insteadof-TemporaryVariable

MarcusDenker:
Hook up OCTempVariable in all the places where we used TemporaryVariable. 

- This shows lots of possibilities for further cleanups, e.g. do we need temporaryVariableNamed: if we have lookupTempVar:  ? 
- lots of duplicated code in e.g. temporaryVariables
- #method is strange

But that could be a next PR


MarcusDenker:
two tests in VariableBreakpointTest failing. 

I think the Reason: TemporaryVariable used to cover arguments, too.

The new code does not return arguments when calling #temporaryVariableNamed:

dionisiydk (approved):

Issue #6659, opened 2020-06-23
PanelMorph does not implement table and crashes the image


oliveiraallex:
**Describe the bug**
PanelMorph does not implement table. You send #table to self, but that then sends it on

**To Reproduce**
Even the example in the FTCellMorph crashes the image with
```
FTCellMorph new
	addMorphBack: 'Example' asMorph;
	withTopSeparator;
	openInWindow.
```

**Screenshots**
![image](https://user-images.githubusercontent.com/39618015/85403860-5c668f80-b55e-11ea-8c0c-107dea477eed.png)

**Version information:**
 - Pharo8/9

**Expected development cost**
As you know nothing happens magically, so do you have an idea of the development effort required? 
Easy
In addition, are you willing to help because after all Pharo is yours too? 
Yes

**Additional context**
This is broken the PharoThings board inspector on Pharo8/9.

MarcusDenker:
Not a bug, as described in https://github.com/pharo-project/pharo/pull/6660

PR #6660, opened 2020-06-23
checking if  #table exist and returning nil case not #6659 

oliveiraallex:
https://github.com/pharo-project/pharo/issues/6659


estebanlm:
no, this is wrong. 
`FTCellMorph` owner should be a `FTTableRowMorph`. 
If that is not the case, somehow a cell morphs is going inside the wrong element.

PR #6662, opened 2020-06-23
CleanupIsPseudoVariable

MarcusDenker:
The Slot package has an extension method on Symbol #isPseudoVariable. As is is only used in Slot class, it is better to be moved there.

### DIFF
--- a/src/Kernel/Slot.class.st
+++ b/src/Kernel/Slot.class.st
@@ -46,13 +46,18 @@ Slot class >> checkValidName: aSymbol [
 	aSymbol startsWithDigit ifTrue: [ 
 		^ InvalidSlotName signalFor: aSymbol ].
 
-	aSymbol isPseudovariableName ifTrue: [ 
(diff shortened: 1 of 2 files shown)


src/Kernel/Slot.class.st
```diff
@@ -55,6 +55,11 @@ Slot class >> checkValidName: aSymbol [
 
 ]
 
+{ #category : #testing }
+Slot class >> isPseudovariableName: aSymbol [
```
dionisiydk, line 59: For future:
I think we should use the word "reserved". It has better meaning than pseudo or special.
And we can move related methods to ReservedVariable which we have now.

dionisiydk (approved):

src/Kernel/Slot.class.st
MarcusDenker: ok, added to the backlog (but this has true and false, too... so some rethinking needed anyway, if we remove that we can query ReservedVariable subclases instead of hardcoding)

PR #6668, opened 2020-06-24
Cleanup Debugger-Filters package

### LINKED ISSUE #6667: Cleanup Debugger-Filters package
- StackFilter is an abstract class and should therefore provide #isAbstract
- remove unnecessary dots
- tag the classes
- fix lint in kernelClassesToExclude
- ...

astares:
Fix #6667

### DIFF
--- a/src/Debugger-Filters/BlockFilter.class.st
+++ b/src/Debugger-Filters/BlockFilter.class.st
@@ -7,7 +7,7 @@ Class {
 	#instVars : [
 		'block'
 	],
-	#category : #'Debugger-Filters'
+	#category : #'Debugger-Filters-Base'
 }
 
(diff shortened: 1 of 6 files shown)


MarcusDenker:
One thing we need to think about, too: merging packages. Some of the granularity we have now makes no sense... far too many tiny packages that never make sense to be e.g. used independently

MarcusDenker:
failure not related

astares:
> One thing we need to think about, too: merging packages. Some of the granularity we have now makes no sense... far too many tiny packages that never make sense to be e.g. used independently

Agree for the Debugger-Filters (for other packages it might make sense) - but lets see how this looks after the switch to NewTools.

MarcusDenker (approved):

PR #6669, opened 2020-06-24
6532-Incomplete-state-of-method-AST

MarcusDenker:
This is a try to cache the method as part of the AST. There is a strange failure in my image, let's see if this happens on the CI, too

### DIFF
--- a/src/AST-Core/ASTCacheMissStrategy.class.st
+++ b/src/AST-Core/ASTCacheMissStrategy.class.st
@@ -6,5 +6,8 @@ Class {
 
 { #category : #accessing }
 ASTCacheMissStrategy >> getASTFor: aCompiledMethod [
-	^ aCompiledMethod parseTree doSemanticAnalysisIn: aCompiledMethod methodClass
+	| ast |
+	ast := aCompiledMethod parseTree doSemanticAnalysisIn: aCompiledMethod methodClass.
+	ast compilationContext compiledMethod: aCompiledMethod.
+	^ast
 ]
--- a/src/AST-Core/ASTCacheMissStrategy.class.st
+++ b/src/AST-Core/ASTCacheMissStrategy.class.st
@@ -6,8 +6,5 @@ Class {
 
 { #category : #accessing }
 ASTCacheMissStrategy >> getASTFor: aCompiledMethod [
-	| ast |
-	ast := aCompiledMethod parseTree doSemanticAnalysisIn: aCompiledMethod methodClass.
-	ast compilationContext compiledMethod: aCompiledMethod.
-	^ast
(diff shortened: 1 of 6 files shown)


MarcusDenker:
MustBeBooleanTest>>testWhile is failing, somehow the bc->AST mapping is out of sync

src/OpalCompiler-Core/RBMethodNode.extension.st
```diff
@@ -62,6 +64,7 @@ RBMethodNode >> generate: trailer [
 { #category : #'*OpalCompiler-Core' }
 RBMethodNode >> generateIR [
 	| ir |
+	bcToASTCache := nil.
```
dionisiydk: This part is suspicious related to failing test

src/AST-Core/ASTCacheMissStrategy.class.st
```diff
@@ -6,5 +6,8 @@ Class {
 
 { #category : #accessing }
 ASTCacheMissStrategy >> getASTFor: aCompiledMethod [
-	^ aCompiledMethod parseTree doSemanticAnalysisIn: aCompiledMethod methodClass
+	| ast |
+	ast := aCompiledMethod parseTree doSemanticAnalysisIn: aCompiledMethod methodClass.
```
dionisiydk: it means that "aCompiledMethod parseTree" will miss the compiled method. 
I think it should be set in #parseTree method. This way you will also avoid duplication in reflective method.

MarcusDenker: ah.. I will try that!

MarcusDenker:
maybe the reason is that we have to call sourceNodeForPC: on the block, it is:


sourceNodeForPC: anInteger 
	self methodNode ir.
	(self hasProperty: #ir) ifTrue: [ "FullBlockClosure"
		^ self bcToASTCache nodeForPC: anInteger].
	^ self methodNode sourceNodeForPC: anInteger

src/OpalCompiler-Core/RBMethodNode.extension.st
MarcusDenker: The reset I think is good: if we re-generate the IR, the cache is (potentially) not valid.

What I want to do after this PR is to refactor to *never* cache the IR... that should simplify things a bit

MarcusDenker:
(It is too hot and I start to mix up PRs already.. the last one was for the other strange bug :-)

MarcusDenker:
failing tests no related

MarcusDenker:
Interesting: CI run passed in 40 minutes. That seems much faster then what we saw without this PR...

PR #6671, opened 2020-06-24
ReservedVariable-usingMethods-isReferenced

MarcusDenker:
- Implement #usingMethods for ReserverdVariables. It is slow when the AST does not exist for all methods, so no test for now
- move isReferenced from Slot to Variable, it implements it in terms of #usingMethods

### DIFF
--- a/src/OpalCompiler-Core/ReservedVariable.class.st
+++ b/src/OpalCompiler-Core/ReservedVariable.class.st
@@ -48,6 +48,13 @@ ReservedVariable >> printOn: stream [
 	stream nextPutAll: self name
 ]
 
+{ #category : #queries }
+ReservedVariable >> usingMethods [
+	"first call is very slow as it creates all ASTs"
+	^SystemNavigation new allMethods select: [ : method |
+		method ast variableNodes anySatisfy: [ :varNode | varNode binding == self]]
+]
+
 { #category : #debugging }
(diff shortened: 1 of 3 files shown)


src/OpalCompiler-Core/ReservedVariable.class.st
```diff
@@ -48,6 +48,13 @@ ReservedVariable >> printOn: stream [
 	stream nextPutAll: self name
 ]
 
+{ #category : #queries }
+ReservedVariable >> usingMethods [
+	"first call is very slow as it creates all ASTs"
+	^SystemNavigation new allMethods select: [ : method |
+		method ast variableNodes anySatisfy: [ :varNode | varNode binding == self]]
```
dionisiydk, line 55: For those variables we need to implement optimized versions of readsIn:/writesIn: . Probably there is  already a way to ask method if it includes self sends, super calls or thisContext.
Then usingMethods will be quite trivial and fast.

But it is for another task. We need to clean and improve query protocol of variables.

dionisiydk (approved):

src/OpalCompiler-Core/ReservedVariable.class.st
MarcusDenker: yes! There should be bytecode scanning method (or if not we can add them easily)

PR #6681, opened 2020-06-26
Introduce Collections-Weak-Tests

### LINKED ISSUE #6680: Introduce Collections-Weak-Tests
- create a package Collections-Weak-Tests and move the weak collection test classes from Collections-Tests to it
  to further break down the bloated Collections-Tests into smaller pieces
- tag them according to the structure in the package that is tested

astares:
Fix #6680


MarcusDenker:
MetacelloNotification: Loaded -> Collections-Stack-Tests-tonel.1 --- tonel:///builds/workspace/uest_and_branch_Pipeline_PR-6681/64/src --- cache
[64]   DictionaryTest
[64] You must resolve these dependencies before you will be able to load these definitions:

astares:
Dependency problem 

We need to get https://github.com/pharo-project/pharo/pull/6726 
into standard image first

astares:
should be fine now

MarcusDenker (approved):

PR #6694, opened 2020-06-26
Symbol>>#asIcon should be in Deprecated90

### LINKED ISSUE #6693: Symbol>>#asIcon should be in Deprecated90
- move the method Symbol>>#asIcon to Deprecated90, there was enough time to convert code so we should remove it with transition to Pharo 10

astares:
Fix #6693


MarcusDenker:
failing test not related

MarcusDenker (approved):

PR #6695, opened 2020-06-26
6514-definition-with-slot-can-now-use-hasSpecificLayout

### LINKED ISSUE #6514: definition with slot can now use hasSpecificLayout
with the introduction of hasSpecificLayout now definitionWithSlots can use it.

MarcusDenker:
remove #hasSpecificLayout. The method that was supposed to use it uses now #isFixedLayout. fixes #6514


MarcusDenker:
the CI did not pass for this yet. Even for the most trivial PRs, we should wait for at least one slave running all tests

PR #6701, opened 2020-06-26
6520-Progress-bar-should-not-happen-below-the-toolbar

### LINKED ISSUE #6520: Progress bar should not happen below the toolbar.
It is annoying
Screenshot 2020-06-08 at 22 29 38

Ducasse:
fixes: #6520 at least it will work a bit better


Ducasse:
We did it with guille so I will integrate.

PR #6703, opened 2020-06-26
ReImplementedNotSentRule-Disable

### LINKED ISSUE #6696: Disable ReImplementedNotSentRule
ReImplementedNotSentRule costs too much in the way it is done now.

We should disable it for now (this way people can turn it on if they want it)

MarcusDenker:
fixes #6696


Ducasse:
This is sad but this is a useful rule.

MarcusDenker:
Yes! We need to get it back, but it needs to be implemented more clever. (data has to be hold by the environment, then maybe have a low-priority thread to update it? Or maybe this will be a use-case for TaskIt..)

Issue #6329, opened 2020-05-10
setUp template proposed by calypso could contain super call


Ducasse:
When we browse a test class calypso proposes a pane with setUp and a nice comment. 
It would be nice that the template also includes

```
setUp
     super setUp.
```

VincentBlondeau:
And the same for the tearDown method:
```
tearDown 
     super tearDown.
```

dionisiydk:
Yes, that would be nice.

But notice that currently what you see is not a template but the actual
method TestCase>>#setUp (or whatever #setUp exists in current test
hierarchy). So special logic needs to be added to override "default setup"
method with special "template sources".

вс, 10 мая 2020 г. в 19:55, StéphaneDucasse <notifications@github.com>:

dionisiydk:
вс, 10 мая 2020 г. в 20:17, Vincent Blondeau <notifications@github.com>:

> And the same for the tearDown method:
>
> tearDown
>      super tearDown.
>
> There is no behaviour in Calypso to assist with #tearDown method. Only
#setUp tab is available

Ducasse:
Yes denis I was talking about this one.

olekscode:
Let me clarify this a bit.

At the moment, Calypso does not provide any template for setUp and tearDown methods.
"setUp" and "tearDown" tabs appear only for those classes that have those methods implemented.
You want to show those tabs for every class. And in case the class does not have it's own implementation of setUp or tearDown, the tab will show the default template:
```Smalltalk
setUp
    super setUp.
```
```Smalltalk
tearDown
    super tearDown.
```
Am I correct?

olekscode:
We changed the implementation of `ClyTestSetUpEditorTool` in such a way that setUp tab is only displayed for the test classes that provide their own implementation of setUp.
Before, there was a lookup and the implementation of setUp from the nearest superclass was displayed (which in most cases was the  empty method with a comment from `TestCase`). Now we only check the child class which is being displayed and don't do the lookup.

The tearDown tab is not displayed.
However, I believe that it should be treated in the same way as setUp.
So I have created a separate issue for it: #6705

olekscode:
I think this issue can be closed because https://github.com/pharo-project/pharo/pull/6704#issuecomment-653768342 and  https://github.com/pharo-project/pharo/pull/6717

PR #6704, opened 2020-06-26
6329-setUp-template-proposed-by-calypso-could-contain-super-call

olekscode:
Fixing #6329

### DIFF
--- a/src/Calypso-SystemPlugins-SUnit-Browser/ClyTestSetUpEditorTool.class.st
+++ b/src/Calypso-SystemPlugins-SUnit-Browser/ClyTestSetUpEditorTool.class.st
@@ -30,7 +30,7 @@ ClyTestSetUpEditorTool class >> shouldBeActivatedInContext: aBrowserContext [
 	
 	(aBrowserContext lastSelectedItem hasProperty: ClyTestResultProperty) ifFalse: [ ^false ].
 	
-	(aBrowserContext lastSelectedClass lookupSelector: #setUp) ifNil: [ ^false ].
+	(aBrowserContext lastSelectedClass selectors includes: #setUp) ifFalse: [ ^false ].
 	"last condition is to not show this tool if user selects #setUp method manually"
 	aBrowserContext isMethodSelected ifFalse: [ ^true ].	
 	^aBrowserContext browser methodSelection lastSelectedItem name ~= 'setUp'
@@ -76,7 +76,7 @@ ClyTestSetUpEditorTool >> setUpModelFromContext [
 { #category : #initialization }
 ClyTestSetUpEditorTool >> setUpParametersFromModel [
 
-	editingMethod := testClass lookupSelector: #setUp.
+	editingMethod := testClass methodNamed: #setUp.
 	super setUpParametersFromModel
 ]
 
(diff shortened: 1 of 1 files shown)


MarcusDenker:
failing test not related

dionisiydk:
@olekscode One of the purposes of setUp tab was to assist with new setUp implementation. Due to simplistic approach the missing part was to provide a good template instead of raw empty TestCase>>setUp. 
Is it intentional to remove this feature?

olekscode:
@dionisiydk yes, it was intentional. What we had was a `setUp` tab that always contained the code from a superclass. In most cases, this would be an empty method from TestCase, in some cases, it would be a `setUp` method from a parent test.

Having an empty template for a method is not harmful, but also not very useful. And there was no easy fix that would improve the template by adding `super setUp` to it, because it would require adding this line to `TestCase` which does not have any superclass that implements `setUp`.

What's worse is that every time you modify the proposed "template" and try to save it, you get a pop-up asking you whether you want to add the new method to your class or to superclass (TestCase). And some people find it annoying because "of course I want to add the method to the class which I'm currently editing".

Having proper templates for `setUp` and `tearDown` would indeed be nice, but it would require a lot of added effort, because currently we do not have the mechanism for that (showing the superclass implementation might not be good, as I explained above).  And since having those two templates is not that important, we decided that it was better to remove this functionality for now and only show the `setUp` and `tearDown` tabs for classes that implement those methods themselves.

olekscode:
@MarcusDenker do you know what can be causing this test to fail?
Because indeed it seems unrelated

https://ci.inria.fr/pharo-ci-jenkins2/blue/organizations/jenkins/Test%20pending%20pull%20request%20and%20branch%20Pipeline/detail/PR-6704/1/tests

<blockquote><div><strong><a href="https://ci.inria.fr/pharo-ci-jenkins2/blue/organizations/jenkins/Test%20pending%20pull%20request%20and%20branch%20Pipeline/detail/PR-6704/1/tests">Jenkins</a></strong></div></blockquote>

src/Calypso-SystemPlugins-SUnit-Browser/ClyTestSetUpEditorTool.class.st
```diff
@@ -30,7 +30,7 @@ ClyTestSetUpEditorTool class >> shouldBeActivatedInContext: aBrowserContext [
 	
 	(aBrowserContext lastSelectedItem hasProperty: ClyTestResultProperty) ifFalse: [ ^false ].
 	
-	(aBrowserContext lastSelectedClass lookupSelector: #setUp) ifNil: [ ^false ].
+	(aBrowserContext lastSelectedClass selectors includes: #setUp) ifFalse: [ ^false ].
```
dionisiydk:
```suggestion
	(aBrowserContext lastSelectedClass includesSelector: #setUp) ifFalse: [ ^false ].
```

dionisiydk:
> Having proper templates for `setUp` and `tearDown` would indeed be nice, but it would require a lot of added effort, because currently we do not have the mechanism for that (showing the superclass implementation might not be good, as I explained above).

I do not get where this "a lot of effort required" is coming from. But challenge accepted :)). 
I created PR with improvement instead of feature removal: #6717 . You can see that it was easy to do.

MarcusDenker:
#6717 has been merged

olekscode:
Nice! 😄 
Indeed, I might have overestimated the effort that is needed.

MarcusDenker:
So this can be closed now?

Ducasse:
oleksandr should we close this issue?

olekscode:
Yes

PR #6706, opened 2020-06-26
TemporaryVariable-cleanup

MarcusDenker:
This is the first step to merge OCAbstractLocalVariable into OCTempVariable 

- copy OCAbstractLocalVariable into OCAbstractLocalVariableTOMERGE, use that as the superclass for OCTempVariable
- move OCAbstractLocalVariable to depcated and use it as the superlass there
-  merge all methods of OCAbstractLocalVariableTOMERGE down

The ivars have to moved down next and OCAbstractLocalVariableTOMERGE deleted, that is best done by editing the text files

### DIFF
--- a/src/Deprecated90/OCAbstractLocalVariable.class.st
+++ b/src/Deprecated90/OCAbstractLocalVariable.class.st
@@ -11,7 +11,7 @@ Class {
 		'scope',
 		'usage'
 	],
-	#category : #'OpalCompiler-Core-Semantics'
+	#category : #'Deprecated90-OpalCompiler-Core'
 }
 
 { #category : #converting }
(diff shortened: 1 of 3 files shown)


dionisiydk (approved):
Actually I thought to have a superclass (OCLocalVariable) to separate  temps and arguments (to follow  isLocalVariable vs isTemp/isArgument idea)

MarcusDenker:
Yes, OCLocalVariable would be nice.. so what is now OCTempVariable will be OCLocalVariable,it will  have a subclass TermporaryVariable (with subclasses for copying and tempvector, and ArgumentVariable will be another subclass (on the same level as TermporaryVariable

Issue #6705, opened 2020-06-26
Display tearDown tab for test classes in Calypso in the same way as we display the setUp tab


olekscode:
When you open a SystemBrowser, and click on a test class which provides an implementation fro the `setUp` method, Calypso displays this method in a separate `setUp` tab.
However, the same does not happen with `tearDown` (I think it should).

This issue is related to #6329

olekscode:
This implementation contains a little hack to make sure that Calypso browser tabs are not being created from an abstract superclass:
```Smalltalk
ClyTestEditorTool class >> shouldBeActivatedInContext: aBrowserContext

	"This is a hack to make sure that this method is not called for the abstract class ClyTestEditorTool but only for its subclasses"
	ClyTestEditorTool = self ifTrue: [ ^ false ].
	
	aBrowserContext isClassSelected ifFalse: [ ^false ].
	
	(aBrowserContext lastSelectedItem hasProperty: ClyTestResultProperty) ifFalse: [ ^false ].
	
	(aBrowserContext lastSelectedClass selectors includes: self selector) ifFalse: [ ^false ].
	"last condition is to not show this tool if user selects #setUp method manually"
	aBrowserContext isMethodSelected ifFalse: [ ^true ].	
	^aBrowserContext browser methodSelection lastSelectedItem name ~= 'setUp'
```

hogoww:
Hi, didn't you PR this to enable a review? =)

olekscode:
Yes, this was the PR: https://github.com/pharo-project/pharo/pull/6707

PR #6707, opened 2020-06-26
6705-Display-tearDown-tab-for-test-classes-in-Calypso-in-the-same-way-as-we-display-the-setUp-tab

olekscode:
Fixing  #6705. Added a comment describing the hack


MarcusDenker:
one slave timeout, two green

dionisiydk:
How it looks with setUp and tearDown together? One of the reason why I did not add it is a limitation of the default size of the browser. Considering that tearDown is less interesting than setUp for understanding the test I kept only this tab.

MarcusDenker:
This has now a merge conflict

MarcusDenker:
as there was another PR that was touching the same code, the best is to re-submit a new PR

PR #6709, opened 2020-06-26
fix  #2246, OSPlatform return win32 in win64

### LINKED ISSUE #2246: OSPlatform return win32 in win64
`OSPlatform current` returns an instance of Win32Platform on Windows 64bits instead of Win64Platform.

hogoww:
@bencoman 
Hope this is ok :)

### DIFF
--- a/src/System-Platforms/OSPlatform.class.st
+++ b/src/System-Platforms/OSPlatform.class.st
@@ -31,7 +31,7 @@ OSPlatform class >> currentPlatformName [
 
     |platform|
     platform := Smalltalk vm getSystemAttribute: 1001.
-    platform = 'win32' ifTrue: [ ^ 'windows_api' ].
+    platform = 'Win32' ifTrue: [ ^ 'windows_api' ].
     ^ platform
 
     "For more information refer to...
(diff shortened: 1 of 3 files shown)


MarcusDenker:
fails on windows with

FFICalloutMethodBuilder>>createFFICalloutLiteralFromSpec:
Error: External module not found


Which might be related

hogoww:
For some reason it was thinking it was on Unix.
I noticed that the comparison was with Win32 instead of win32, in the comments, so I guess this is coming from that typo.
Not sure how to reproduce locally, will see what the CI says :)

hogoww:
It seems that the 'LD_LIBRARY_PATH' is nil instead of being absent for some reason. 
No idea of why.
The entry is not present in my 

#getEnvVariable:bufferSize:ifAbsent:
Either return the absentblock value.
Throws an exception (error)
loops on a different bufferSize.
returns the buffer asString.
In no ways, should it be able to return nil, from the little I see here :/

StephanEggermont:
Please rename

bencoman:
> Hope this is ok :)

Sorry I've been offline for a bit.  Yes, looks good to me.

bencoman (approved):

hogoww:
But this is still failing, I have to look at it

hogoww:
#7634 seems to fix this as well, so i'll close this pr.
I just implemented the solution of bencoman, but I do not know why this fails.

PR #6710, opened 2020-06-26
Add #hasBadge in Spec2 for testing

guillep:
Add a testing method to Toolbar buttons for testing.
We need this to test in the upcoming 1.9 iceberg version


Ducasse:
One broken test but not related to the issue.

PR #6715, opened 2020-06-27
Reduce-Interaction-Define-Method

MarcusDenker:
When doing Test-first development, the feature to "define method" from the DNU is extremely nice. But the Question to decide for a good category *just then* breaks the flow completely. Nobody wants to decide *now*. This can be done, without harm, later.

Thus: This PR removed the dialog.

### DIFF
--- a/src/Debugger-Actions/DoesNotUnderstandDebugAction.class.st
+++ b/src/Debugger-Actions/DoesNotUnderstandDebugAction.class.st
@@ -20,16 +20,6 @@ DoesNotUnderstandDebugAction >> appliesToDebugger: aDebugger [
 ]
 
 { #category : #private }
-DoesNotUnderstandDebugAction >> askForCategoryIn: aClass default: aString [
-	| categoryName |
-	categoryName := AbstractTool requestProtocolNameFor: aClass initialAnswer: aString.
-	categoryName ifNil: [^aString].
-	^ categoryName isEmptyOrNil 
-		ifTrue: [^ aString] 
-		ifFalse: [ categoryName ]
-]
-
-{ #category : #private }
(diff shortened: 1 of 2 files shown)


MarcusDenker:
The bigger picture behind this PR is to move as many decisions as possible as late as possible. Instead of breaking flow, add a backstop later that guards. 

And example is the "unused temp". it used to be that you got a notification for ever temp, with the decision to keep or remove. Now it just annotates the AST and shows that info later, so that you can fix it in a second pass. In addition to showing it in the tool in the left gutter, wer add rules. And as the last defense, we have tests that make sure that a commit can not be merged without the CI telling you "there is an unused temp".

The same is true for categories: there is no need to break the flow, as we have a rule that shows the devloper later. And if that fails, the CI will not be green with a non-categorised method.

This "as little interaction as needed, but checks later" is a very nice principle, as it allows people to work without being interrupted by trivialities. But the we have guards in place (which can be very annoying...) that forces to give that information later.

dionisiydk:
It can work even more different:
Imagine that the debugger will just show another tab "+ notUnderstoodSelector" (like in the browser "+ Inst. side method" tab) when DNU is signalled. Only on "accept changes" it will install the method. The tab will allow to categorize method in place (like in the browser).

MarcusDenker:
Yes, that would be a good next step... on the other hand, I will not have time to implement it. To me it looks like a good next step after merging this PR

dionisiydk (approved):

bencoman:
Thanks @MarcusDenker.  I think this is particularly important for newcomers - to reduce the number of decisions that they feel they don't know the answer for, increases their sense of mastery.

Issue #6571, opened 2020-06-11
Browser selects default tab when selection is changed instead of current one


dionisiydk:
**Describe the bug**
Browser does not keep the selected tab when selection in lists is changed.
**To Reproduce**
- Browse Announcer class.
- Select Comment tab
- Select Announcement class

Browser will show definition tab for Announcement class
**Expected behavior**
Browser should show a Comment tab as previously selected.

**Version information:**
Pharo 9.

**Expected development cost**
It is related to new notebook widget which does not support some of expected API in same way like it was with old widget.
It should not be difficult to fix. It just requires a debugging session.

Ducasse:
Denis do you have an idea where this bug can come from?

dionisiydk:
@Ducasse if fixed it in PR #6716. See commits f459b8e and 1c2fdee :
New notebook widget incorrectly passed selection indexes in notification methods (the argument with old index was always a current one).
Can I ask why a new widget was implemented instead of improving current one? We also loose multiple tab selection with this change (no comment+method or two methods at same time now)

PR #6716, opened 2020-06-27
Fix updating tabs in Calypso broken by new notebook widget

dionisiydk:
Try to edit setUp method in setUp tab. After accepting changes the browser will show two setUp tabs. That is the bug caused by new notebook (new widget) integration.

In the history I found that #basicUpdateTools was altered to workaround initial tab updating issue: selecting new methods in same class did not update the tab content. This workaround was wrong. At least it should be implemented in explicit place responsible for tab selections (#restoreSelectedTools). But little debugging shows that it was just an issue with new tab implementation (or superclasses which it's based on):
- tab removal and addition incorrectly update the currently selected tab

Therefore here I provide correct fix and revert Calypso method

### DIFF
--- a/src/Morphic-Widgets-PolyTabs/TabSelectorMorph.class.st
+++ b/src/Morphic-Widgets-PolyTabs/TabSelectorMorph.class.st
@@ -32,6 +32,16 @@ TabSelectorMorph >> addTab: aStringOrMorph [
 ]
 
 { #category : #adding }
+TabSelectorMorph >> addTab: aStringOrMorph beforeIndex: index [
+
+	self tabs add: (self newLabelMorph: aStringOrMorph) beforeIndex: index.
+	selectedIndex >= index ifTrue: [ 
+		"there is now new tab before selected tab"
+		self selectedIndex: selectedIndex + 1 ]. 
+	self updateTabs
+]
+
+{ #category : #adding }
 TabSelectorMorph >> addTab: aStringOrMorph selected: selectedStringOrMorph [
 	"Add a new tab with the given text or morph and alternate for when selected."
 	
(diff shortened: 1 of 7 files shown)


src/Spec2-Adapters-Morphic/SpNotebookTabSelectorMorph.class.st
```diff
@@ -48,11 +48,6 @@ SpNotebookTabSelectorMorph >> removeTabIndex: anInteger [
 	"Remove the tab at the given index and
 	adjust any selected (next or first if was last)."
 
-	self tabs removeAt: anInteger.
-
-	self tabs
-		ifNotEmpty: [ self selectedIndex: (self selectedIndex - 1) \\ self tabs size + 1 ]
-		ifEmpty: [ self selectedIndex: 0 ].
-
+	super removeTabIndex: anInteger.
```
dionisiydk: super implementation performs initial selection reset (self selectedIndex: 0).
It is required to trigger updates of selection when tab is removed. Otherwise new selected tab with same index as removed one will not be updated.

src/Morphic-Widgets-PolyTabs/TabSelectorMorph.class.st
```diff
@@ -31,6 +31,16 @@ TabSelectorMorph >> addTab: aStringOrMorph [
 	self updateTabs
 ]
 
+{ #category : #adding }
+TabSelectorMorph >> addTab: aStringOrMorph beforeIndex: index [
+
+	self tabs add: (self newLabelMorph: aStringOrMorph) beforeIndex: index.
+	selectedIndex >= index ifTrue: [ 
+		"there is now new tab before selected tab"
+		self selectedIndex: selectedIndex + 1 ].
```
dionisiydk, line 40: This was a missing part. If we do not shift selection index then widget will think that selection is not changed and will not update inserted tab in case when it will be seleted in subsequent calls.

dionisiydk:
f459b8e commit fixes #6571

Ducasse:
Thanks you denis!

Ducasse:
Having tests is cool. Because this is the only way to control complexity.

dionisiydk:
failed test was really related: #6728.

PR #6718, opened 2020-06-28
Use-globals-for-OCEnvironmentScope

MarcusDenker:
OCEnvironmentScope just wraps the environment (Smalltalk globals in the default case). This is not needed as we can just implement the API there.

- add the needed API to Dictionary
  - yes, this should be *only* on a subclass that is dedicated to be used for environments, this is future work
 - yes, we shoudl not need #hasTempVector there ==> future work
- Change code to use the enviroment as Scope
- move OCEnvironmentScope to deprecated

The extension methods are  kept as Slot extension method together with the other variable related method that needs to be cleaned with a refactoring to have a dedicated dictionary for environment (which might just be SystemDictionary)

(I can't help it but my brain rally loved doing things in a step-by-step way... I tried to have my own fork to create that *huge* change that then is perfect... It just does not work for me)

### DIFF
--- a/src/Deprecated90/OCEnvironmentScope.class.st
+++ b/src/Deprecated90/OCEnvironmentScope.class.st
@@ -7,7 +7,7 @@ Class {
 	#instVars : [
 		'environment'
 	],
-	#category : #'OpalCompiler-Core-Semantics'
+	#category : #'Deprecated90-OpalCompiler-Core'
 }
 
 { #category : #'instance creation' }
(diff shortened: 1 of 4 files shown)


dionisiydk:
> add the needed API to Dictionary
> * yes, this should be _only_ on a subclass that is dedicated to be used for environments, this is future work

Are you talking about SystemDictionary? Why you did not put methods there?

dionisiydk:
I am just curious what it will break? Probably only tests?

MarcusDenker:
yes, it was just tests... should be a simple next step. the #hasTempVector has just one caller, so that will be fixable, too

dionisiydk (approved):
forgot to press

PR #6729, opened 2020-06-28
[Testing] Process suspend resume terminate issue

dionisiydk:
PR enables skipped tests to show "the process isTerminate" issue in particular scenarious.
The explicit test is added for the issue to avoid any relation to the class under test:
- testSuspendResumeTerminateIssue discovers the issue with Process>>isTerminated method when the process is suspended in the middle and then resumed. 
- testSuspendResumeTerminateIssueWithoutSuspend proves that #suspend in the middle is required trigger for the issue

Idea to merge original PR #6280 with TestExecutionEnvironment and keep discovered issue in separate PR for the record and investigation.

### DIFF
--- a/src/SUnit-Tests/ProcessMonitorTestServiceTest.class.st
+++ b/src/SUnit-Tests/ProcessMonitorTestServiceTest.class.st
@@ -65,8 +65,6 @@ ProcessMonitorTestServiceTest >> testAllowRunningBackgroundProcessesToFinishButF
 { #category : #tests }
 ProcessMonitorTestServiceTest >> testAlwaysPassBackgroundHalt [
 	| process haltWasPassed halt |
-	self flag: 'it fails runKernelTests on bootstrap image due to Process suspend issue'.
-	self skip.
 	
 	haltWasPassed := false.
 	halt := Halt new.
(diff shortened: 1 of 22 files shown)


src/SUnit-Tests/ProcessMonitorTestServiceTest.class.st
```diff
@@ -0,0 +1,556 @@
...
+	self assert: suspendedError class equals: UnhandledError. 
+	self assert: suspendedError exception identicalTo: warning.
+]
+
+{ #category : #tests }
+ProcessMonitorTestServiceTest >> testSuspendResumeTerminateIssue [
+	"The test discovers the issue with Process>>isTerminated method
+	when process was suspended in the middle and then resumed.
+	During the bootstrap process on CI it fails RunKernelTests stage.
+	But it is working on full image"
```
dionisiydk, line 525: This is separate test for the issue. It is important to keep it in this class. It seems the case appears when tests are running in concrete order.

src/SUnit-Tests/ProcessMonitorTestServiceTest.class.st
```diff
@@ -0,0 +1,556 @@
...
+	process := [ started := true. Processor activeProcess suspend. ended := true ] newProcess.
+	process priority: Processor activePriority + 1.
+	processes add: process.	
+	process resume.
+	self assert: started description: 'process should be started'.
+	self assert: process isSuspended description: 'process should be suspended'.
+
+	process resume.
+	self assert: ended description: 'process should be ended'.
+	self assert: process isTerminated description: 'process should be terminated'
```
dionisiydk, line 538: And this is failed assertion. As you see the last instruction in the block is executed (ended is true) but process is not terminated. 
I did more deep debugging and found that in that case process pc is zero (and isTerminated relies on greater pc number).
According to Eliot the zero pc is a sign of VM bug in the mapping code between jitted code and bytecode. But it is strange that it appears only with #suspend call at the middle of process (see next test which proves it)

dionisiydk:
I created better test without original PR changes

Issue #6674, opened 2020-06-24
SyntaxErrorDebugger does not work properly


dionisiydk:
**Describe the bug**
I found that syntax error debugger does not really work for a long time (I checked from Pharo 4).
The debugger allows to fix the problem source code in the editor. But accepting changes produces incorrect result:
- it does not close the editor (like it was in Pharo 8 and early)
- it compiles the method but with original bad sources. (you can see it in the browser and check #decompile to prove that fixed text was used to compile it) .

**To Reproduce**
In playground call: 
```Smalltalk
Announcer compile: 'badMethod
^'
```
In opened editor fix the text to make it correct smalltalk syntax (add any literal at the end)
Now accept changes (cmd+s).
Browse Announcer and select new method. You will see bad sources and get another syntax error (from highlighting). 

**Expected behavior**
- Accepting changes should close editor and proceed compilation.
- compiled method should point to actual sources which it was created with.

**Expected development cost**
It should not be difficult if you are familiar with SyntaxErrorNotification and compiler.

MarcusDenker:
One idea: why do we not allow these to be loaded?


If we add to OpalCompiler:

isInteractive
	^self isInteractiveFor: compilationContext requestor

and then change the method:		
		
parseMethod
	
	^ (self compilationContext optionParseErrors or: [ self isInteractive not ])
		ifTrue: [self parserClass parseFaultyMethod: source contents]
		ifFalse: [self parserClass parseMethod: source contents]

==> methods like that can just be loaded. The syntax error is compiled to be raised at runtime (and there it is not working, too).
But 99% of all cases where we load syntactically incorrect code we will not execute it, so it would be a huge step even without fixing the the tool (which we should do, too)

dionisiydk:
I definitely agree about it. Users will fix incorrect syntax in the browser like any other issue.
+100 for removing all this complex machinery for syntax correction in fly. It did not work for a long time and nobody complain.

dionisiydk:
Also it will help to resolve circular dependency between AST and Compiler (which is why I really found this problem)

MarcusDenker:
and at runtime we can just raise a RuntimeSyntaxError, which would be a subclass of Halt.

MarcusDenker:
I merged the PR... there are of course lots of cleanup steps and lots of "next steps", but this issue is fixed.

(the SyntaxErrorDebugger itself is not yet removed, but it is only used for expressions (doit) where it is used even less and it actually works...)

PR #6730, opened 2020-06-29
Allow-Compiling-Syntax-Errors-NonInteractively

MarcusDenker:
This is the first step for #6674
- add #isInteractive to easily check if a compiler instance is interactive
- use faulty parsing when non-interactive

This way we can load and compile all methods with syntax errors, the syntax error is raised at runtine

Next steps:
- add Rule to warn if a method conatins syntax errors
- raise a Halt Subclass at runtime so we can reture the syntaxerror debugger

### DIFF
--- a/src/Kernel/RuntimeSyntaxError.class.st
+++ b/src/Kernel/RuntimeSyntaxError.class.st
@@ -1,0 +1,10 @@
+"
+When compiling syntactically incorrect code, we compile raising this exception.
+
+This way the debugger opens and the programmer can easily fix the faulty method
+"
+Class {
+	#name : #RuntimeSyntaxError,
+	#superclass : #Halt,
+	#category : #'Kernel-Exceptions'
+}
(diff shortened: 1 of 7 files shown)


src/Kernel/RuntimeSyntaxError.class.st
```diff
@@ -0,0 +1,10 @@
+"
+When compiling syntactically incorrect code, we compile raising this exception.
+
+This way the debugger opens and the programmer can easily fix the faulty method
+"
+Class {
+	#name : #RuntimeSyntaxError,
+	#superclass : #Halt,
```
dionisiydk: I think it should be an Error subclass. Why you decide to make it a Halt?

MarcusDenker: Hmm... right. My brain thinks Halt opens the debugger because I want to, while error is an error. And I wanted to open debugger :-)
But then, this is actually a error, and error opens the debugger, too.

So yes, should be Error subclass

dionisiydk (approved):

PR #6735, opened 2020-06-29
Dictionary-no-hasTempVector

MarcusDenker:
We should not need to implement hasTempVector on the level of the global scope.

This refactors pushOuterVectors: to stop checking at the level of the MethodScope.

### DIFF
--- a/src/OpalCompiler-Core/IRTranslator.class.st
+++ b/src/OpalCompiler-Core/IRTranslator.class.st
@@ -92,7 +92,7 @@ IRTranslator >> pushOuterVectors: scope [
 	| scopesWithVector sc |
 	scopesWithVector := OrderedCollection new.
 	sc := scope.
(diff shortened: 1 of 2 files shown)


dionisiydk (approved):

src/OpalCompiler-Core/IRTranslator.class.st
```diff
@@ -92,7 +92,7 @@ IRTranslator >> pushOuterVectors: scope [
 	| scopesWithVector sc |
 	scopesWithVector := OrderedCollection new.
 	sc := scope.
-	[ sc outerScope isNil ] whileFalse: [ 
+	[ sc outerScope isInstanceScope ] whileFalse: [
```
dionisiydk, line 95: Maybe better would be to ask a scope to pushTempVector. Then recursively it will travers all scopes up to OCInstanceScope

PR #6737, opened 2020-06-29
Rewrite basic editor and tests

jordanmontt:
This is related to #6625 
The next step is to add an ExampleBasedEditor (you can build a rule based on Pharo's code) and to integrate this tool with MatchTool.

[Stef's observations](https://github.com/pharo-project/pharo/pull/6625#issuecomment-645313804) have been attended and the following changes have been added:

- *Protocol name* has been renamed to *package name*.
- Class creation has been extracted to a different method.
- The UIManager has been separated from rule creator.
- The default template rule is not hard-coded anymore. It is loaded from a saved rule.
- Several tests have been added.

### DIFF
--- a/src/RewriteRule-Tests/BasicEditorTest.class.st
+++ b/src/RewriteRule-Tests/BasicEditorTest.class.st
@@ -1,0 +1,92 @@
+Class {
+	#name : #BasicEditorTest,
+	#superclass : #TestCase,
+	#instVars : [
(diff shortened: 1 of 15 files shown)


jordanmontt:
Same error as before  #6599. The error is caused for moving RewriteRuleChangesBrowser to RewriteRuleEditor package. I will do it again.

PR #6741, opened 2020-06-30
Added Rewrite Basic Editor.

jordanmontt:
(Copy-pasted from: #6737 ) The next step is to add an ExampleBasedEditor (you can build a rule based on Pharo's code) and to integrate this tool with MatchTool.

Stef's observations have been attended and the following changes have been added:

- Protocol name has been renamed to package name.
- Class creation has been extracted to a different method.
- The UIManager has been separated from rule creator.
- The default template rule is not hard-coded anymore. It is loaded from a saved rule.
- Several tests have been added.

### DIFF
--- a/src/BaselineOfRewriteTool/BaselineOfRewriteTool.class.st
+++ b/src/BaselineOfRewriteTool/BaselineOfRewriteTool.class.st
@@ -1,0 +1,21 @@
+"
+The baseline of RewriteTool. This is necessat for install the tool from GitHub.
+"
+Class {
+	#name : #BaselineOfRewriteTool,
+	#superclass : #BaselineOf,
+	#category : #BaselineOfRewriteTool
+}
+
+{ #category : #baselines }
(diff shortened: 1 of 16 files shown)


Ducasse:
- Pay attention setUp method should do a super setUp.

Ducasse:
- you can replace Smalltalk globals by self class environment

Ducasse:
- you can replace Smalltalk globals by self class environment
- (RBCustomTransformationRule subclasses collect: [ :each | each ]) => RBCustomTransformationRule subclasses

Ducasse:
Can you tell me how you load your packages?
I think that we should have a baselineOfRewriter loading your packages.

jordanmontt:
> Can you tell me how you load your packages?
> I think that we should have a baselineOfRewriter loading your packages.

I have this same code in [my repo](https://github.com/jordanmontt/RewriteTool-Spec2) and I have the baseline there. I will add it here as well.

PR #6742, opened 2020-06-30
Cleanup-OCAbstractLocalVariableTOMERGE

MarcusDenker:
- remove OCAbstractLocalVariableTOMERGE
- remove old TemporaryVariable (will come back later, never used so we can do this without deprecation)


MarcusDenker:
hmm.. some bug that removing a class does not remove the extension method from the extending package? to be explored...

dionisiydk (approved):

MarcusDenker:
I will redo this is two steps

PR #6743, opened 2020-06-30
ReMethodHasSyntaxErrorRule

MarcusDenker:
add ReMethodHasSyntaxErrorRule, it checks if a method has a syntax error (e.g. because the preference to allow it was enabled)

### DIFF
--- a/src/GeneralRules/ReMethodHasSyntaxErrorRule.class.st
+++ b/src/GeneralRules/ReMethodHasSyntaxErrorRule.class.st
@@ -1,0 +1,40 @@
+"
+Methods can be compiled with syntax errors, fix the code and recompile the method
+"
+Class {
+	#name : #ReMethodHasSyntaxErrorRule,
+	#superclass : #ReAbstractRule,
+	#category : #'GeneralRules-Migrated'
+}
+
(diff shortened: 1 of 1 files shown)


src/GeneralRules/ReMethodHasSyntaxErrorRule.class.st
```diff
@@ -0,0 +1,40 @@
...
+{ #category : #accessing }
+ReMethodHasSyntaxErrorRule class >> uniqueIdentifierName [
+	"This number should be unique and should change only when the rule completely change semantics"
+	
+	^'MethodSyntaxError'
+]
+
+{ #category : #running }
+ReMethodHasSyntaxErrorRule >> basicCheck: aMethod [
+	^aMethod ast isFaulty
```
dionisiydk, line 24: To make it fast we can introduce "aMethod hasSyntaxError" which will check some explicit property to avoid ast analysis (it can be also "self sends: #signalRuntimeSyntaxError" check).

Also we can mark a class that it includes such problems to indicate bad packages and classes in the browser.

dionisiydk (approved):

PR #6744, opened 2020-06-30
This changes doIt expressions to be compiled with runtime syntax errr…

MarcusDenker:
…os. As for methods, this only affects non-interactive use.

A problem is that you can not actually change the DoIt method.


dionisiydk (approved):

MarcusDenker:
The only downside is that you an not edit the doit with a syntax error to then get the changed result. But then I never ever (ever!) needed that...

MarcusDenker:
The additional commit now fixes changing a doit with a syntax error in the debugger. It now seems to work exactly like the SyntaxErrorDebgger: eval on accept, then proceed just returns that value.

MarcusDenker:
I will redo this with, but with a preference so that we can have it off for now...

PR #6748, opened 2020-07-01
Expand OSWindow with additional bindings for the generic renderer

ronsaldo:
This PR introduces additional bindings into OSWindow by adding the following features to the OSWindowGenericRenderer API:
- Specification of clipping rectangles
- Text rendering through the Pharo fonts by creating a texture atlas.
- Subpixel antialiased rendering via a two pass text rendering operation.
- Support for converting static forms into cached textures.
- Support for rendering into textures through the SDL2 render targets.


Ducasse:
Hi ronie 
two points: 
- I did not see any tests. Do you have some tests around?
- You have some as yet unclassified

S

ronsaldo:
I just noticed that I had forgotten to push one of my commits. As for the tests, most of OSWindow does not have tests.
I can add additional tests for the font texture atlas creation, the additional font methods. But for the rest of OSWindow, I need some guidelines on writing those tests. I can add test on whether things are not crashing, but testing on whether the rendering is actually working, that is much trickier to test. It also might require some guards for the case of headless CI slaves.

PR #6750, opened 2020-07-01
Speedup-RelationSlot-write

MarcusDenker:
RelationSlot was compiling a reflective write. But we can do better;

- refactor write: to:  so we do most of the stuff in another method
- implement emitStore: to call that method and use bytecodes to acces ivars

This PR was done to speed up Moose which uses RelationSlots a lot.

### DIFF
--- a/src/Slot-Examples/RelationSlot.class.st
+++ b/src/Slot-Examples/RelationSlot.class.st
@@ -81,9 +81,10 @@ RelationSlot >> emitStore: aMethodBuilder [
 		addTemp: tempName;
 		storeTemp: tempName;
 		popTop;
-		pushReceiver;
+ 		pushLiteral: self;
 		pushInstVar: index;
 		pushTemp: tempName;
+		pushReceiver;
 		send: #updateOld:new:in:;
 		storeInstVar: index
 ]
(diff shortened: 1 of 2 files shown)


src/Slot-Examples/RelationSlot.class.st
```diff
@@ -71,17 +71,22 @@ RelationSlot >> checkValue: aValue [
...
+	"
 	| tempName |
 	tempName := '0slotTempForStackManipulation'.
 	aMethodBuilder
 		addTemp: tempName;
 		storeTemp: tempName;
 		popTop;
-		pushReceiver;
-		pushLiteralVariable: (AdditionalBinding key: #slot value: self);
+ 		pushLiteral: self;
```
dionisiydk, line 84: This is quite interesting change. What makes #pushReceiver faster than pushLiteral:?
Or is there other reason for this change?

src/Slot-Examples/RelationSlot.class.st
```diff
@@ -71,17 +71,22 @@ RelationSlot >> checkValue: aValue [
...
+		pushInstVar: index;
 		pushTemp: tempName;
-		send: #writeSlot:value:
+		pushReceiver;
```
dionisiydk, line 87: Hm.. During my first comment I did not notice this extra #pushReceiver.
Could you explain the logic please?

dionisiydk: Ok, I see. Self here is not a receiver. Sorry for the noise.

PR #6765, opened 2020-07-03
willCreateBlock-support-fullBlock

MarcusDenker:
willCreateBlock does not detect full blocks, this leads to to problems in the debugger that temps are read from the wrong context

### DIFF
--- a/src/Kernel-BytecodeEncoders/EncoderForSistaV1.class.st
+++ b/src/Kernel-BytecodeEncoders/EncoderForSistaV1.class.st
@@ -486,6 +486,15 @@ EncoderForSistaV1 class >> isCreateFullBlock: compiledBlock code: code at: pc [
 ]
 
 { #category : #'instruction stream support' }
+EncoderForSistaV1 class >> isCreateFullBlockAt: pc in: code [
+	^ self
+		  extensionsAt: pc
+		  in: code
+		  into: [ :extA :extB :nExtBytes | 
+		  (code at: pc + nExtBytes) = 16rF9 ]
+]
+
+{ #category : #'instruction stream support' }
 EncoderForSistaV1 class >> isExtension: bytecode [
 	"Answer if the bytecode is an extension bytecode, i.e. one that extends
 	 the range of the following bytecode."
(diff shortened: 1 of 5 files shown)


src/Kernel-BytecodeEncoders/EncoderForSistaV1.class.st
```diff
@@ -485,6 +485,15 @@ EncoderForSistaV1 class >> isCreateFullBlock: compiledBlock code: code at: pc [
 				compiledBlock = (code literalAt: (code at: pc + nExtBytes + 1) + (extA bitShift: 8) + 1)] ]
 ]
 
+{ #category : #'instruction stream support' }
+EncoderForSistaV1 class >> isCreateFullBlockAt: pc in: code [
+	^ self
+		  extensionsAt: pc
+		  in: code
+		  into: [ :extA :extB :nExtBytes | 
+		  (code at: pc + nExtBytes) = 16rF9 ]
```
dionisiydk, line 494: Why we can't put this logic into isCreateBlockAt:in: ? It would be good to add a comment about it

MarcusDenker: I am not yet at all sure where to put what.. I think it is good to have, on the level of the encoder, two different methods to check (ad we have two different ways to create blocks).

MarcusDenker:
StepThroughTest>>testStepThrough fails after this.

Which is interesting... and needs some analysis.

MarcusDenker:
unix-64 / Tests-unix-64 / testStepOverAssignment – Unix64.Debugger.Tests.Base.AssignmentAndLiteralDebuggerTest

unix-64 / Tests-unix-64 / testNewDebugSession – Unix64.Debugger.Tests.Base.AssignmentAndLiteralDebuggerTest

unix-64 / Tests-unix-64 / testStepIntoAssignment – Unix64.Debugger.Tests.Base.AssignmentAndLiteralDebuggerTest

pavel-krivanek (approved):

Ducasse:
Marcus I integrate it and we will see.

PR #6781, opened 2020-07-06
UndefinedSlot

MarcusDenker:
- FIx classbuilder:  move #installSlotsAndVariables to before recompiling methods
- add UndefinedSlot

The idea of UndefinedSlot is that for the cast a class is loaded where the slot definition does not exist, we create this special UndefinedSlot.
-  It has the same name
- has the definition of the slot (the AST) 
- prints itself as the definition (the AST). This might be slightly different then the prefered printout for the slot, but contains all infos to create the correct slot later
- reads return nil
- writes are ignored

On read and write, the UndefindedSlot will check if the class of the original definition now has been loaded and if yes, will rebuild the class.

### DIFF
--- a/src/Kernel/UndefinedSlot.class.st
+++ b/src/Kernel/UndefinedSlot.class.st
@@ -56,7 +56,7 @@ UndefinedSlot >> read: anObject [
 
 { #category : #accessing }
 UndefinedSlot >> slotClassName [
-	^ast arguments first binding name 
+	^ast arguments first variable name 
 ]
 
 { #category : #'meta-object-protocol' }
--- a/src/Kernel/UndefinedSlot.class.st
+++ b/src/Kernel/UndefinedSlot.class.st
@@ -1,0 +1,67 @@
+"
+When loading a class where a Slot is used that is not present in the System, we need to model
+this slot somehow.
+
+The UndefinedSlot ingores reads and write (it returns nil like undeclared variables). But on
+access, it checks if the Slot has been loaded and if yes, it rebuilds the class definition.
+"
+Class {
+	#name : #UndefinedSlot,
+	#superclass : #Slot,
+	#instVars : [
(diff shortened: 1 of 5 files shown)


src/Kernel/UndefinedSlot.class.st
```diff
@@ -0,0 +1,67 @@
...
+{ #category : #'meta-object-protocol' }
+UndefinedSlot >> read: anObject [
+	"Undeclared slots read nil always, but check if they can repair the class"
+	 self checkClassRebuild.
+	^ nil
+]
+
+{ #category : #accessing }
+UndefinedSlot >> slotClassName [
+	^ast arguments first binding name
```
dionisiydk: Do we have #variable now?
```suggestion
	^ast arguments first variable name 
```
Then we should deprecate #binding.

src/Kernel/UndefinedSlot.class.st
```diff
@@ -0,0 +1,67 @@
...
+{ #category : #accessing }
+UndefinedSlot >> slotClassName [
+	^ast arguments first binding name 
+]
+
+{ #category : #'meta-object-protocol' }
+UndefinedSlot >> write: aValue to: anObject [
+	"Undeclared slots ignore writes, but check if they can repair the class"
+	 self checkClassRebuild.
+	^ aValue
```
dionisiydk, line 66: The question: what is the better behaviour in that case.
I believe we should raise an error in all such cases. Fail as soon as possible with meaningful message. Same is related to UndeclaredVariable.

dionisiydk (approved):

src/Kernel/UndefinedSlot.class.st
MarcusDenker: I wanted to do the same as UndeclaredVariable

PR #6782, opened 2020-07-06
6778-ShiftClassBuilder-does-not-recompile-SharedPool-users-when-it-changes

tesonep:
When creating a SharedPool the associations should become from the old to the new associations.
As they don't share the same hierarchy.

### DIFF
--- a/src/Shift-ClassInstaller/ShiftClassInstaller.class.st
+++ b/src/Shift-ClassInstaller/ShiftClassInstaller.class.st
@@ -231,7 +231,7 @@ ShiftClassInstaller >> migrateClassTo: newClass [
 		(newClass classPool includesKey: oldAssociation key)
 			ifTrue: [ 
 				newAssociation := newClass classPool associationAt: oldAssociation key.
-				newAssociation value: oldAssociation value.
+				newClass classVarNamed: oldAssociation key put: oldAssociation value.
 				newClassVariableBindings add: newAssociation.
 				oldClassVariableBindings add: oldAssociation. ]].
 	
--- a/src/Shift-ClassInstaller/ShiftClassInstaller.class.st
+++ b/src/Shift-ClassInstaller/ShiftClassInstaller.class.st
@@ -130,8 +130,7 @@ ShiftClassInstaller >> installInEnvironment: newClass [
 
 	"I only install in the environment if there is not oldClass installed."
 	(self installingEnvironment hasClassNamed: builder name) ifFalse:[
-		self installingEnvironment installClass: newClass withName: builder name.
-	].
+		self installingEnvironment installClass: newClass withName: builder name ].
 
 	self
 		updateBindings: (self installingEnvironment bindingAt: builder name)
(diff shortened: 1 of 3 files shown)


src/Shift-ClassInstaller/ShiftClassInstaller.class.st
```diff
@@ -224,15 +224,25 @@ ShiftClassInstaller >> migrateClassTo: newClass [
 	slotsToMigrate := newClass class allSlots reject:[:e | builder builderEnhancer hasToSkipSlot: e ].
 	slotsToMigrate do: [ :newSlot | oldClass class slotNamed: newSlot name ifFound: [ :oldSlot | newSlot write: (oldSlot read: oldClass) to: newClass ] ].
 	
-	oldClass classVarNames do: [ :aClassVar |
-		(newClass hasClassVarNamed: aClassVar) ifTrue:[ newClass classVarNamed: aClassVar put: (oldClass classVarNamed: aClassVar) ].
-	].
+	oldClassVariableBindings := OrderedCollection new.
+	newClassVariableBindings := OrderedCollection new.
+	
+	oldClass classPool associations do: [ :oldAssociation | | newAssociation |
```
dionisiydk, line 230: We should use classVariable API:
```suggestion
	oldClass classVariables do: [ :oldAssociation | | newAssociation |
```

src/Shift-ClassInstaller/ShiftClassInstaller.class.st
```diff
@@ -224,15 +224,25 @@ ShiftClassInstaller >> migrateClassTo: newClass [
...
+		(newClass classPool includesKey: oldAssociation key)
```
dionisiydk, line 231:
```suggestion
		(newClass hasClassVarNamed: oldAssociation key)
```

src/Shift-ClassInstaller/ShiftClassInstaller.class.st
```diff
@@ -224,15 +224,25 @@ ShiftClassInstaller >> migrateClassTo: newClass [
...
+			ifTrue: [ 
+				newAssociation := newClass classPool associationAt: oldAssociation key.
```
dionisiydk, line 233:
```suggestion
				newAssociation := newClass classVariableNamed: oldAssociation key.
```

src/Shift-ClassInstaller/ShiftClassInstaller.class.st
```diff
@@ -224,15 +224,25 @@ ShiftClassInstaller >> migrateClassTo: newClass [
...
+				newClass classVarNamed: oldAssociation key put: oldAssociation value.
```
dionisiydk, line 234:
```suggestion
				newAssociation write: oldAssociation read
```
And can you rename these temps to have newVariable and oldVariable?

dionisiydk (changes requested):

dionisiydk:
For the future work:

I think the same migration is required for some types of Slot. There are cases when they are pushed as literals (RelationSlot for example).

Another general issue is what to do with variable class change. Imagine we are changing existing ClassVariable to be a WeakClassVariable. It will require recompilation of methods. And simple become trick will not be enough

dionisiydk (approved):

PR #6785, opened 2020-07-06
Introduce-LocalVariable

MarcusDenker:
Instead of the old structure with OCTemVariable with a subclass OCArgumentVariable, this now tries to model the level of the name analysis the same as what we have on the AST:

-> We have an abstract class LocalVariable with subclasses ArgumentVariable and TemporaryVariable
-> #isTemp now answers false for ArgumentVariable

compieler tests are green... this will allow us to clean up the #isArg, isTemp and #isArgOrTemp on the ast level easier, I hope...

Needs a careful review

### DIFF
--- a/src/Debugging-Core/Context.extension.st
+++ b/src/Debugging-Core/Context.extension.st
@@ -30,7 +30,7 @@ Context >> lookupSymbol: aSymbol [
 Context >> lookupTempVar: aSymbol [
 	| var |
 	var := self lookupVar: aSymbol.
-	var isTemp ifFalse: [ ^self error: var name, ' is not a temp but, ', var class name].
+	var isLocalVariable ifFalse: [ ^self error: var name, ' is not a temp but, ', var class name].
 	^var
 ]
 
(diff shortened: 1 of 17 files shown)


MarcusDenker:
@hogoww another run against the wall :-) It will fall... 

What is nice is that now the semantic level is in line what you started to do on the AST level, which I think should simplify further cleanups there

dionisiydk (approved):
I do not see any class deprecation. Do you think they are redundant for these variables?

MarcusDenker:
I wanted to add subclasses for OCTempVariable and OCArgumentVariable.

Another thing that needs to be fixed is isLocal vs. isLocalVariable. #isLocal on slot is used to check if the Slot is from a Trait or not.

dionisiydk:
> I wanted to add subclasses for OCTempVariable and OCArgumentVariable.
> 
> Another thing that needs to be fixed is isLocal vs. isLocalVariable. #isLocal on slot is used to check if the Slot is from a Trait or not.

Yes. It should be simple rename. There are not much senders. For the name I would use #isDefinedByOwningClass.

PR #6789, opened 2020-07-07
Reflectivity-LiteralVariable-Cleanup

MarcusDenker:
A small cleanup related to references to LiteralVariable, mostly touching MetaLinks (Reflectivity)

- in #entities we can now use Variable
- simplify #visitVariableValue:
- #testLinkOnClassVarForObject and testLinkObjectToClassVarName are now green! (they where on self skip)

Calypso related cleanup: 
- #usesUndeclares -> usesUndeclareds (private method, even though a canditate for an API method)
- simplify

### DIFF
--- a/src/Reflectivity/RFASTTranslator.class.st
+++ b/src/Reflectivity/RFASTTranslator.class.st
@@ -307,8 +307,7 @@ RFASTTranslator >> visitVariableNode: aVariableNode [
 
 { #category : #reflectivity }
 RFASTTranslator >> visitVariableValue: aVariable [
-	self flag: #pharoTodo. "needs to be extendend to other kinds of variables and cleaned"
-	((aVariable isKindOf: LiteralVariable) or: [ (aVariable isKindOf: Slot) ]) ifTrue: [ self emitPreamble: aVariable. self emitMetaLinkBefore: aVariable. ].
+	(aVariable isKindOf: Variable) ifTrue: [ self emitPreamble: aVariable. self emitMetaLinkBefore: aVariable. ].
 	aVariable emitValue: methodBuilder.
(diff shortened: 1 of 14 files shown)


src/Reflectivity/RFASTTranslator.class.st
```diff
@@ -307,8 +307,7 @@ RFASTTranslator >> visitVariableNode: aVariableNode [
 
 { #category : #reflectivity }
 RFASTTranslator >> visitVariableValue: aVariable [
-	self flag: #pharoTodo. "needs to be extendend to other kinds of variables and cleaned"
-	((aVariable isKindOf: LiteralVariable) or: [ (aVariable isKindOf: Slot) ]) ifTrue: [ self emitPreamble: aVariable. self emitMetaLinkBefore: aVariable. ].
+	(aVariable isKindOf: Variable) ifTrue: [ self emitPreamble: aVariable. self emitMetaLinkBefore: aVariable. ].
 	aVariable emitValue: methodBuilder.
-	((aVariable isKindOf: LiteralVariable) or: [ (aVariable isKindOf: Slot) ]) ifTrue: [self emitMetaLinkAfterNoEnsure: aVariable].
+	(aVariable isKindOf: Variable) ifTrue: [self emitMetaLinkAfterNoEnsure: aVariable].
```
dionisiydk, line 312: why we still need isKindOf check?

dionisiydk (approved):

src/Reflectivity/RFASTTranslator.class.st
MarcusDenker: it should not be needed, I will do another cleanup for this

PR #6793, opened 2020-07-07
Removing old completion Morphs

tesonep:
Removing the old completion and keybinding.


MarcusDenker:
close and reopen to trigger ci run

PR #6794, opened 2020-07-08
Reflectivity-Cleanups-2

MarcusDenker:
Some cleanups, most related to Reflectivity but two others, too:

- LiteralVariable>>#isCascade is not needed
- remove isKindOf: from visitVariableValue:
- Ring: testMethod exectuted code that was not part of assert:
- ReGlobalVariablesUsageRule: us #isClassOrTrait, else direct references to Traits are false positives

### DIFF
--- a/src/GeneralRules/ReGlobalVariablesUsageRule.class.st
+++ b/src/GeneralRules/ReGlobalVariablesUsageRule.class.st
@@ -25,7 +25,7 @@ ReGlobalVariablesUsageRule >> check: aMethod forCritiquesDo: aCriticBlock [
 					(self isKnownGlobal: node name) not and: [
 						(Smalltalk globals
 							at: node name
-							ifAbsent: [ Object ]) isClass not ] ] ] ]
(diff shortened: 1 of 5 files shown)


astares (approved):

src/GeneralRules/ReGlobalVariablesUsageRule.class.st
```diff
@@ -16,17 +16,12 @@ ReGlobalVariablesUsageRule class >> checksMethod [
 
 { #category : #running }
 ReGlobalVariablesUsageRule >> check: aMethod forCritiquesDo: aCriticBlock [
-	"Beware that ClassVariables respond true to #isGlobal. 
-	 The trick with 'ifAbsent:[Object] is basically a way not select ClassVariables (or Pools)"
 	aMethod ast allChildren
 		select: [ :node |
 			node isVariable and: [ 
-				node isGlobal and: [
+				node variable isGlobalVariable and: [
```
dionisiydk, line 22: we should have something like isClassBinding to distinguish globals like Smalltalk from classes or traits

PR #6801, opened 2020-07-08
5193-Running-Executable-Comments-Checker-on-200-000-Classes-images-should-work-faster

### LINKED ISSUE #5193: Running Executable Comments Checker on 200 000+ Classes images should work faster
and 1.8M methods. It is also taking several times the memory of the initial image.

MarcusDenker:
This speeds up both selecting packages and then running the tests by pre-checking of the sourceCode of the method contains a string >>>

This avoids to call #comments on the method which is  very slow due to creating the AST.

fixes #5193 at least to some extent: It is now quite usable for the image, which means it should be *much* faster for a larger image.

### DIFF
--- a/src/DrTests-CommentsToTests/DTCommentTestConfiguration.class.st
+++ b/src/DrTests-CommentsToTests/DTCommentTestConfiguration.class.st
@@ -9,17 +9,14 @@ Class {
 
 { #category : #converting }
 DTCommentTestConfiguration >> asTestSuite [
-	| suite |
+	| suite methods |
 	suite := TestSuite new.
 	suite := TestSuite named: 'Test Generated From Comments'.
-	self items
-		reject: [ :each | each isAbstract ]
-		thenDo: [ :c | 
-			c methods
-				do: [ :m | 
-					(m methodClass compiledMethodAt: m selector) comments
-						do: [ :com | 
-							(com includesSubstring: '>>>')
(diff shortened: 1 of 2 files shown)


src/DrTests-CommentsToTests/DTCommentTestConfiguration.class.st
```diff
@@ -9,17 +9,14 @@ Class {
...
-		reject: [ :each | each isAbstract ]
-		thenDo: [ :c | 
-			c methods
-				do: [ :m | 
-					(m methodClass compiledMethodAt: m selector) comments
-						do: [ :com | 
-							(com includesSubstring: '>>>')
-								ifTrue: [ suite addTest: (CommentTestCase comment: com class: c selector: m selector) ] ] ] ].
+	methods := (self items reject: [ :each | each isAbstract ]) flatCollect: [ :each | each methods ].
+	methods := methods select: [:each | each sourceCode includesSubstring: '>>>' ].
```
dionisiydk, line 16: That was quite ugly code. We should refactor it further to have good API:
- aMethod hasExecutableComment
-  aMethod collectExecutableComments.

dionisiydk (approved):

src/DrTests-CommentsToTests/DTCommentTestConfiguration.class.st
MarcusDenker: yes, the idea was to do just the tiniest local change... there is an api to get the example objects that should be used here

PR #6804, opened 2020-07-09
Compiler-DeadCode-Clean

MarcusDenker:
Tiny cleanup in the extension methods of the compiler:

- allTempNames in  UndefinedObject not needed
- asOneArgSelector not needed
- literalIndexOf:ifAbsent: â> OCLiteralList  (it has just one user who uses it on a literal list)


Ducasse:
Why you do not want to keep asOneArgSelector?

MarcusDenker:
It's a very specialised method, it does

#fasdas:qwerqwr: asOneArgSelector   ->#fasdasqwerqwr:

For just turning #asdasf into a one arg selector, we have #asMutator 


String/Symbol has already a huge API, as this method was an extension method that is clearly not API but just a private method that is not used by the package, I would like to remove it.

dionisiydk (approved):

Ducasse:
Thanks for the explanation. Indeed it is too specialised for nothing.

PR #6805, opened 2020-07-09
Variables-isGlobal-Cleanup

MarcusDenker:
On the level of both the AST and the Variables, #isGlobal returns true for all Literal Variables. That is not good, e.g.:

(SmalltalkImage classVariableNamed: #CompilerClass) isGlobal

- This PR checks and fixes all senders that send #isGlobal to a Variable 
- Then it changes #isGlobal to be in sync with isGlobalVariable 

The AST level is not touched! There #isGlobal still returns true for class var nodes (future work)
Another future work is to decide if we merge isGlobalVariable and isGlobal on the level of the Variables and just have one of the two.

### DIFF
--- a/src/Kernel/GlobalVariable.class.st
+++ b/src/Kernel/GlobalVariable.class.st
@@ -27,6 +27,12 @@ GlobalVariable >> emitValue: methodBuilder [
 ]
 
 { #category : #testing }
+GlobalVariable >> isGlobal [
+
+	^ true
+]
+
+{ #category : #testing }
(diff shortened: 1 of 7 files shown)


Ducasse:
Could you add a comment to isGlobal because this is not clear what is the difference between isGlobal and isGlobalVariable
We should ***really*** add method level comments.

MarcusDenker:
this is the next step: I want to get rid of #isGlobal and just have #isGlobalVariable.

PR #6814, opened 2020-07-10
AST-deprecate-methodComments

MarcusDenker:
#methodComments is a strange method... it returns all comments at the level of the method, but including comments on arguments.

There are no clients.  It is not clear why this is needed as an API method.


dionisiydk (approved):

Ducasse:
And it would be much better to have in a visitor if we need it.

PR #6815, opened 2020-07-10
6008-Two-methods-to-test-if-a-node-is-an-argument

### LINKED ISSUE #6008: Two methods to test if a node is an argument
related #6001
@MarcusDenker 
There is isArg on RBVariable node (which uses the binding property)
and there is isArgument on RbProgramNode and RBArgument node.

That may solve the problem of #6001, but this is *very* error prone in my opinion.
Whatdo you think?

MarcusDenker:
Now that the semantics of arguments vs temps are the same on the AST and the Variables, we can get rid of #isArg 
(prior cleanups already removed almost all users on the AST level)

- add #isArgumentVariable on the Variable Hierarchy
- fix all users of isArg to use that, deprecate the method

- fix users of #isArg on the AST level to use #isArgument
- deprecate the method

- fix Ring to use isArgumentVariable, too

fixes #6008


MarcusDenker:
cleaning #isArgOrTemp is the step to be done after this.

dionisiydk (approved):

PR #6816, opened 2020-07-10
Speedup-class-removal

MarcusDenker:
This speeds up class deletion by not putting deleted classes into the Obsoletes

As we did not recompile, existing references still referenced the obsolete class anyway. Just the syntax coloring was hiding that.

Thus, it is better to make that explicit by adding a Rule that shows when a method references obsolete classes.

small additional cleanup: ReGlobalVariablesUsageRule uses now #isGlobalClassNameBinding instead of having a reference to the Smalltalk globals.


jecisc:
Awesome!! That you Marcus!

Is it possible to backport to Phaor 8 or not?

MarcusDenker:
Yes, should be possible to back port (maybe without the rules as they already rely on the better "first class variables" model...)

jecisc:
It's mostly the removal speed up that interest me because the speed regression added 7min of build on a Jenkins job I run often and it's a little bothersome to have to wait 7 more minutes when you need the build. :)

src/Kernel/Class.class.st
```diff
@@ -885,7 +885,6 @@ Class >> removeFromSystem: logged [
 		self superclass addObsoleteSubclass: self].
 
 	myCategory := self category.
-	self isReferenced ifTrue: [Undeclared declare: self name asSymbol from: Smalltalk globals].
```
dionisiydk, line 888: The problem is what will happen when after removal we will load (create) class back? 
I did not check but I assume that methods pointing to old obsolete class will keep using it instead of new loaded class.

MarcusDenker: There is no difference, as we did not recompile the methods (which would make it even slower).

MarcusDenker: At least in my test it was still referencing the Obsolete class after re-loading. Just the syntax highlighter was happy and showing it as an existing class

dionisiydk: But this is a bug. Maybe I misunderstood.
Let's see an example: there is ClassA and there is a #methodWithClassA :
```Smalltalk
MyClassB class>>methodWithClassA 
  ^ClassA
```
"MyClassB methodWithClassA == ClassA" is true.
Now remove ClassA and create it again.
What will be the result of "MyClassB methodWithClassA == ClassA"? 
If it is not true it is a bug.

MarcusDenker: Hmm... I tested it again and now it works.

Another idea: can we just add the class always to the Undeclared dict? Without the #isReferenced check?

This would accumulate entries, but we have that problem already. (e.g. imagine loading a method with an Undeclared and then delete it, the entry stays forever).

MarcusDenker: I committed a version that does this

dionisiydk (approved):

dionisiydk:
This is good change. But I think PR to manage Undeclared with weak structure is still relevant. It is not critical I think but it would be nice to have.
Now any unload operation will keep garbage there.

PR #6820, opened 2020-07-11
6790-BaselineOfMD-V2

### LINKED ISSUE #6790: BaselineOfMicrodown is not loaded
Microdown packages are in the repo the baselineOfMicrodown but not loaded

Ducasse:
fixes: #6790

### DIFF
--- a/src/BaselineOfIDE/BaselineOfIDE.class.st
+++ b/src/BaselineOfIDE/BaselineOfIDE.class.st
@@ -194,11 +194,6 @@ BaselineOfIDE >> baseline: spec [
 				repository: repository;
 				loads: #('Code' 'CodeTests') ].
 		
-		"No idea why we need to define it here because it is expressed as dependency inside
-		microdown baseline!"	
-		spec baseline: 'Pillar' with: [ spec 
-				loads: #('rich text exporter');
-				repository: repository ].
 		spec baseline: 'Microdown' with: [ 
 			spec 
 				repository: repository;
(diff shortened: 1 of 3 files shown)


Ducasse:
I do not really understand because I made the Clap part for P8 conditional and all the builds are loading in P8 and P9 in travis. I'm puzzled.

Ducasse:
Thanks Pablo, do you know if I should also  commit the new version (in sync normally with the PR) of the packages in the repo?

tesonep:
Are there changes in the repository of Microdown? If they are maybe it is a good idea to sync them.
I do not understand completely the question. If my change work we need to push it to the microdown repo.

Ducasse:
No there is no changes. I tagged the version I pushed so that I can follow changed.
First I would like to get MD in.

Ducasse:
The conflict is resolved and I should check what is the problem

Ducasse:
I do not get why I lost a class in the process. So the tests are failing because the test context is not set. I going super slowly to understand how I can push in the PR. I thought it would be simple
but I have problem to get your changes and commit in this PR. May be because I'm tired.

Ducasse:
I found a way to pull the PR. 
I'm looking for the PRDocumentSample.
It is in Pillar-Tests-Model and should be moved to core. 
Now if I load from microdown repo I will lose the changes. So I loaded from pharo
I will do it tomorrow. 
Now I have to check because travis works. But tomorrow since I'm dead

Ducasse:
I will merge this one and fix the Pillar packages with the missing classes afetr because I cannot spent 3 hours on this.

PR #6827, opened 2020-07-11
Nice-testing-Methods-for-Variable

MarcusDenker:
The names of the testing methods for Variables (and VariableNodes) are not good. #isGlobal works, but it breakes down with #isInstance... 

As one step, this PR just makes sure that all the "nice" variable check methos exist on the Variable hierarchy

- add missing (e.g. isTemporaryVariable)
- the short versions forward to the new and have a comment that they will be deprecated
- add comments to the version on Variable for those that are not obvious (eg isLiteralVariable, the term is not nice but that is what they are called on the Bytecode level)

This PR does *not*
- change any senders
- modify the AST level
- isSelf/isSuper are skiped for now (they should be isSelfVariable, too)

### DIFF
--- a/src/Kernel/Variable.class.st
+++ b/src/Kernel/Variable.class.st
@@ -114,8 +114,9 @@ Variable >> isClassVariable [
 ]
 
 { #category : #testing }
-Variable >> isGlobal [ 
-	^false
+Variable >> isGlobal [
+	"this will be deprecated"
+	^self isGlobalVariable
 ]
 
 { #category : #testing }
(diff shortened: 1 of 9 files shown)


src/Kernel/Variable.class.st
```diff
@@ -175,6 +194,11 @@ Variable >> isTemp [
 	^ false
 ]
 
+{ #category : #testing }
+Variable >> isTemporaryVariable [
```
dionisiydk, line 198: I think a shorter version will be better: #isTempVariable. 
I would also rename TemporaryVariable to TempVariable. 
Otherwise we should use longer names for subclasses to be consistent (OCCopyingTempVariable. OCVectorTempVariable )

dionisiydk (approved):

src/Kernel/Variable.class.st
MarcusDenker: yes, possible... there are already #hasTemporaryVariableNamed: (and similar), but they where not used a lot

What is important is to not shorten "Variable" to "Var" as that is the difference between old and new reflective methods (e.g. instVarNamed: vs instanceVariableNamed:)

PR #6833, opened 2020-07-13
Speedup-pharoDocCommentNodes

MarcusDenker:
- add property API to RBComment (which shows that it might be better a subclass of RBProgramNode?)
- implement #hasDocComment: quick check for >>> and check if property is alrready defined.
- #pharoDocCommentNodes checks hasDocComment, caches result in property
- DocCommentIconStyler uses hasDocComment

Benchmark: 
[(Behavior>>#methodDict) ast pharoDocCommentNodes] bench.

### DIFF
--- a/src/AST-Core/RBComment.class.st
+++ b/src/AST-Core/RBComment.class.st
@@ -29,7 +29,8 @@ Class {
 	#instVars : [
 		'contents',
 		'start',
-		'parent'
+		'parent',
+		'properties'
 	],
 	#category : #'AST-Core-Nodes'
 }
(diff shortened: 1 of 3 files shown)


MarcusDenker:
Simplistic benchmarks:

"large comment, no docCommment"
[(Behavior>>#methodDict) ast pharoDocCommentNodes] bench.
 "'4.635 per second'" 
"'555076.554 per second'"
"one small docComment"
[(RBMethodNode>>#pharoDocCommentNodes) ast pharoDocCommentNodes] bench.
 "'3900.660 per second'" 
"'710242.055 per second'"

It really speeds up syntax highlighting.. browsing through methods quickly (or just down the stack in the debugger) feels much more fluid.

pavel-krivanek (approved):

Ducasse:
Marcus we should make RBComment subclass of RBProgramNode in the subsequent PR.

PR #6846, opened 2020-07-14
LocalVariable-declaringNode

MarcusDenker:
add a query method to LocalVariable #declaringNode. It returns the AST node that declared the variables (that is, the argument in the method signature, the block arg  or the RBVariable that is in the | | temp definition in the AST.

- remove #definingScope as it is the same as scope (and just used in tests)
- add Tests

### DIFF
--- a/src/OpalCompiler-Core/ArgumentVariable.class.st
+++ b/src/OpalCompiler-Core/ArgumentVariable.class.st
@@ -13,6 +13,11 @@ ArgumentVariable class >> semanticNodeClass [
 	^RBArgumentNode 
 ]
 
+{ #category : #queries }
+ArgumentVariable >> declaringNode [
+	^ scope node arguments detect: [ :each | each name = name ]
+]
+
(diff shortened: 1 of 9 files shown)


src/OpalCompiler-Core/ArgumentVariable.class.st
```diff
@@ -13,6 +13,11 @@ ArgumentVariable class >> semanticNodeClass [
 	^RBArgumentNode 
 ]
 
+{ #category : #queries }
+ArgumentVariable >> declaringNode [
```
dionisiydk, line 17: I think we are using "defining" word for this kind of meaning. There is definingClass in methods and variables. Would it better to follow it here too?

MarcusDenker: ah, right... we have isDefinition, too, on the AST level (which should use this code). Added to the backlog to rename it.

PR #6868, opened 2020-07-16
speedup-ReVariableReferencedOnceRule

MarcusDenker:
simpify ReVariableReferencedOnceRule by using #usingMethods. Faster even though we iterate to find all methods, not just the first.


dionisiydk (approved):
good stuff :)

PR #6871, opened 2020-07-16
CompiledCode-useLiteralAt

MarcusDenker:
antother fix..

### DIFF
--- a/src/Kernel/CompiledCode.class.st
+++ b/src/Kernel/CompiledCode.class.st
@@ -487,7 +487,7 @@ CompiledCode >> literalsAt: anIndex [
 { #category : #literals }
 CompiledCode >> literalsDo: aBlock [
 	"Evaluate aBlock for each of the literals referenced by the receiver."
-	1 to: self numLiterals do:
+	1 to: self numLiterals - self literalsToSkip do:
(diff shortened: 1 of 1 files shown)


MarcusDenker:
This fixes e.g.

(SendsDeprecatedMethodToGlobalRuleTest>>#testBasicCheck ) hasLiteral: (SendsDeprecatedMethodToGlobalRule binding).

src/Kernel/CompiledCode.class.st
```diff
@@ -235,11 +235,11 @@ CompiledCode >> frameSize [
 { #category : #literals }
 CompiledCode >> hasLiteral: literal [
 	"Answer whether the receiver references the argument, literal."
-	2 to: self numLiterals - 1 do: "exclude superclass + selector/properties"
+	1 to: self numLiterals - self literalsToSkip do: "exclude superclass + selector/properties"
```
dionisiydk, line 238: Now it starts with 1. Do you know why it was 2 before?

MarcusDenker: yes, is was 2 to - 1, now it is 1 to -2, but #literalAt: does "+1".

Ducasse:
This looks good to me. This is nice to see this code improving.

PR #6873, opened 2020-07-17
Variables-SmallCleanups

MarcusDenker:
- fix some tiny mistakes of the last PRs
- rename declaringNode to definingNode
- implement isDefinition in terms of definingNode
- add #instanceVariables to Behavior returning Slots
- remove #isSlot introduces in the Variable hierarhchy (we have isInstanceVariable)
- remove #isObservable: Slot users should not add extension method if it an be avoided


MarcusDenker:
failing test not related

PR #6874, opened 2020-07-17
ReRuleManager-cleanups-related-to-enabled

MarcusDenker:
- remove classAdded: on the instance side: dead code
- implement enabled and enabled: on RBLintRule
- clean up all the code that just was there because of these methods not implemenented


Ducasse:
Looks good.

PR #6880, opened 2020-07-17
reduce-usage-RBVariableNode-subclasses

MarcusDenker:
reduce references to RBVariableNode subclasses

The AST should only know variables. Givign meaning to it is done by semantic analysis.


Ducasse:
Looks good.

PR #6881, opened 2020-07-17
whichMethodsReferToCleanUp

MarcusDenker:
The story of literals on CompiledMethod is such a mess...

Therse are the demensions:

1) literals use the last two entries as poor-mans variables. Do we want to include them or not?
2) Do we want to dive into Arrays and Pragmas?
3) Do we want to dive into Blocks?
4) Do we want to check for #+ and similar who would be there if we would not just not put them?

Combined, this is nearly impossible... 

What this PR does: we can not use #hasLiteral: , as that should be the version that (nearly) shows what is real. It should *not* look into embedded blocks.
So we can no use it here. The negative point is that we now dive into Arrays and Pragmas, which makes this method slower then needed if you use is with
global variable bindings.

Why do we do this crap? Why not have a higher level model?


dionisiydk (approved):

MarcusDenker:
There are some strange tests failures.. I will close this and redo it later. Needs some analysis.

PR #6883, opened 2020-07-17
6719-Dictionary-creation-API-should-accept-associations-too

Ducasse:
Fixes: 6719: add newFromAssociation, add test, fixed some protocol names.

### DIFF
--- a/src/Collections-Unordered/Dictionary.class.st
+++ b/src/Collections-Unordered/Dictionary.class.st
@@ -88,10 +88,10 @@ Dictionary class >> newFrom: aDict [
 { #category : #'instance creation' }
 Dictionary class >> newFromAssociations: anAssocCollection [
 	"Return a dictionary with the associations defined in anAssocCollection. If the same key appears twice in anAssocCollection, the last one enumerated win."
-	
-	^ anAssocCollection 
-		inject: self new 
-		into: [ :dict :assoc | dict add: assoc. dict ]
+	| newDict |
+	newDict := self new: anAssocCollection size.
+ 	anAssocCollection do: [:each | newDict add: each ].
+	^ newDict
 ]
 
 { #category : #'instance creation' }
--- a/src/Collections-Unordered/Dictionary.class.st
+++ b/src/Collections-Unordered/Dictionary.class.st
@@ -86,6 +86,15 @@ Dictionary class >> newFrom: aDict [
 ]
 
 { #category : #'instance creation' }
+Dictionary class >> newFromAssociations: anAssocCollection [
+	"Return a dictionary with the associations defined in anAssocCollection. If the same key appears twice in anAssocCollection, the last one enumerated win."
+	
+	^ anAssocCollection 
+		inject: self new 
+		into: [ :dict :assoc | dict add: assoc. dict ]
+]
+
+{ #category : #'instance creation' }
 Dictionary class >> newFromKeys: keys andValues: values [
 	"Create a dictionary from the keys and values arguments which should have the same length."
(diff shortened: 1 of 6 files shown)


Ducasse:
This is super strange that it crashes because this is just the addition of a new method.

src/Collections-Unordered/Dictionary.class.st
```diff
@@ -85,6 +85,15 @@ Dictionary class >> newFrom: aDict [
 "
 ]
 
+{ #category : #'instance creation' }
+Dictionary class >> newFromAssociations: anAssocCollection [
+	"Return a dictionary with the associations defined in anAssocCollection. If the same key appears twice in anAssocCollection, the last one enumerated win."
+	
+	^ anAssocCollection 
+		inject: self new 
+		into: [ :dict :assoc | dict add: assoc. dict ]
```
dionisiydk: Maybe it's just me. But I found it quite complicated. Why not use simple version with do?
```Smalltalk
result := Dictionary new: anAssocCollection size.
 anAssocCollection do: [:each | result add: each ].
^result
```

Ducasse: Tx denis.
I followed the implementation of kasper. Now I should bench them to see which one is faster and pick the fastest one.

Ducasse:
I should fix 

testDictionaryPublicProtocolCompatibility – MacOSX64.Collections.Tests.Sequenceable.OrderedDictionaryTest and friends.

Ducasse:
Tests are green.

Ducasse:
Now there is something wrong with Dictionary >> newFrom: because this is the only newFrom: that expects another instance of the same class.

dionisiydk (approved):

Ducasse:
Denis my impression is that Dictionary class >> newFrom: is wrong and that it should accept a collection of binding. 
Because nearly all the other implementors are done like that.
What do you think?

S

dionisiydk:
> Denis my impression is that Dictionary class >> newFrom: is wrong and that it should accept a collection of binding.
> Because nearly all the other implementors are done like that.
> What do you think?

I think I agree. But to be honest I never used these methods in practice and I did not really remember them. 
From the other side we can be more clever and support both cases with double dispatch. For example:
```Smalltalk
Dictionary>>addAll: aCollection
   aCollection importItemsIntoDictionary: self

Dictionary class>>newFrom: aCollection 
	| newDictionary |
	newDictionary := self new: aCollection size.
        aCollection importItemsIntoDictionary: newDictionary.
        ^newDictionary

Collection>>importItemsIntoDictionary: aDictionary
   "receiver is expected to include associations to be able to contribute into aDictionary"
   self do: [:each | aDictionary add: each]

Dictionary>>importItemsIntoDictionary: aDictionary
   self associationsDo: [:each | aDictionary add: each]

```

Ducasse:
But what the double dispatch will bring to us? that if we have different hierarchy of dictionary we only have one definition in the collection?

Ducasse:
I close it.

dionisiydk:
> But what the double dispatch will bring to us? that if we have different hierarchy of dictionary we only have one definition in the collection?

We will be able to use both kind of objects in these messages (an array of associations and a dictionary):
```Smalltalk
aDictionary addAll: {#key -> #value. #key2 -> #value}.
aDictionary2 := Dictionary withAll: {#ke3 -> #value. #key4 -> #value}.
aDictionary addAll: aDictionary2.
aDictionary3 := Dictionary withAll: aDictionary2.
```

PR #6887, opened 2020-07-18
Variables-CopiedArgs

MarcusDenker:
introduce OCCopyingArgumentVariable. Arguments can end up to be copied, too. We need to model that.

### DIFF
--- a/src/OpalCompiler-Core/OCAbstractMethodScope.class.st
+++ b/src/OpalCompiler-Core/OCAbstractMethodScope.class.st
@@ -17,7 +17,7 @@ Class {
 
 { #category : #'temp vars - copying' }
 OCAbstractMethodScope >> addCopyingTemp: aTempVar [
-	^copiedVars at: aTempVar name put: (OCCopyingTempVariable new
+	^copiedVars at: aTempVar name put: (aTempVar copiedVarClass new
 			originalVar: aTempVar originalVar;
 			name: aTempVar name;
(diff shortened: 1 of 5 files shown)


src/OpalCompiler-Core/OCAbstractMethodScope.class.st
```diff
@@ -17,7 +17,7 @@ Class {
 
 { #category : #'temp vars - copying' }
 OCAbstractMethodScope >> addCopyingTemp: aTempVar [
-	^copiedVars at: aTempVar name put: (OCCopyingTempVariable new
+	^copiedVars at: aTempVar name put: (aTempVar copiedVarClass new
 			originalVar: aTempVar originalVar;
 			name: aTempVar name;
 			escaping: aTempVar escaping;
```
dionisiydk, line 23: it really needs to be refactored to "aTempVar createCopiedVariable" (or similar)

dionisiydk (approved):
It will bring us two similar classes for temp and arg copies.
Can we have just one OCCopyingLocalVariable? 
Then methods like #isArgument will delegate to originalVar (originalVar isArgument)

src/OpalCompiler-Core/OCAbstractMethodScope.class.st
MarcusDenker: noted for the backlog.

PR #6888, opened 2020-07-18
simplify-ReVariableAssignedLiteralRule

MarcusDenker:
ReVariableAssignedLiteralRule by getting the assignmentNodes via the slots and check directly.


MarcusDenker:
Not sure if it really speeds up: it was faster in the benchmark as it ended a lot in the RBParseTreeSearcher case.

But I still vote for including it now, as it is *much* simpler

Ducasse:
Ok I could not really understand the logic.

PR #6894, opened 2020-07-18
6890-Variablevariable-is-deprecated

### LINKED ISSUE #6890: Variable>>#variable is deprecated
In the pharo build log:

The method Variable>>#variable called from ArgumentVariableTest>>#testDeclaringNode has been deprecated. Use receiver directly instead.

MarcusDenker:
- fix the depcrecated call. fixes #6890
- add CompiledMethod>>#accessesField: and test, use it in InstanceVariable>>isAccessedIn:
- add #definingNode which where missed on the other PR (and yes, this is really a wrapper.. to be done later)
- remove empty OCVariableSpecialisationTest
- add #definedVariable to get all class and invars of a class
- remove deprecated #instanceVariables on ClassDescription

### DIFF
--- a/src/OpalCompiler-Core/TemporaryVariable.class.st
+++ b/src/OpalCompiler-Core/TemporaryVariable.class.st
@@ -20,7 +20,15 @@ TemporaryVariable >> copiedVarClass [
 
 { #category : #queries }
 TemporaryVariable >> definingNode [
-	^ scope node temporaries detect: [ :each | each name = name ]
+	^ scope node temporaries 
+		detect: [ :each | each name = name ]
+		ifNone: [ 
+			" ugly workaround to support temps defined by primitives"
+			| pragma |
+			pragma := scope node methodNode pragmas detect: [ :each | each isPrimitiveError ].
+			pragma ifNil: [ ^nil ].
+			^ RBVariableNode named: (pragma argumentAt: #error:) value asString
+			]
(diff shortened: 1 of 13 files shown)


MarcusDenker:
(I merge this as it fixes two really annoying bugs in addition)

src/OpalCompiler-Core/TemporaryVariable.class.st
```diff
@@ -20,7 +20,15 @@ TemporaryVariable >> copiedVarClass [
...
-	^ scope node temporaries detect: [ :each | each name = name ]
+	^ scope node temporaries 
+		detect: [ :each | each name = name ]
+		ifNone: [ 
+			" ugly workaround to support temps defined by primitives"
+			| pragma |
+			pragma := scope node methodNode pragmas detect: [ :each | each isPrimitiveError ].
+			pragma ifNil: [ ^nil ].
+			^ RBVariableNode named: (pragma argumentAt: #error:) value asString
+			]
```
dionisiydk, line 31: This "error pragma" case really deserves a test and a comment with method example.
I do not know what it is about really.

MarcusDenker: yes, I will do that... it needs some much better implementation, this is just a workaround to not have the rules fail on all those methods

PR #6901, opened 2020-07-19
Improving debugger infrastructure step 1: a Trait to unify the debugger API used by the system to open debuggers

StevenCostiou:
This PR is an intermediate step of a general improvement of the debugger infrastructure.

This work is about decoupling the debugger opening from the UI (for now, the logic of opening debuggers and handling debugger errors is in `UIMorphicManager`), unifying debuggers API to open them (for now, there at least 3 different ways of requesting the opening of a debugger, and many different users of those 3 ways), and finally cleaning the debugger infrastructure to start further work and improvements on the debugging model.

Instead of a super huge PR, we chose to introduce some independant parts of this work first in order to get early feedback.
Therefore, the introduced features are not to be used by the system yet.

In this PR, you will find:
- Some package renaming and reorganization for consistency
- A Trait `TDebugger`, which all debuggers must (soon) use to be recognized by the system as a debugger.

**`TDebugger` class comment:**

Use me if you want to be recognized as a debugger.
I specify the interface that all debuggers must implement to be used as debuggers.
Conversely, classes recognized as debuggers are all able to start debugging from a given debug session.

The API that all debuggers will expose by using me is:

`availableAutomatically:` a boolean telling if the debugger is (de)activated.
`rank:` an integer giving a priority to the debugger (against other debuggers priorities). A high priority debugger will be preferably opened by the system rather than a low priority debugger.

`handlesContext: aContext`: returns a boolean, tells if the debugger can handle the given context (e.g., a SUnit debugger will be able to handle test cases contexts). By default, any context is handleable.

`debugSession: aDebugSession`: Explicit requirement that must be implemented by all debuggers.
This API method is the entry point to open a debugger on a DebugSession instance.
It is the preferred debugger opening entry point used by the system to open a debugger.

### DIFF
--- a/src/BaselineOfGeneralTests/BaselineOfGeneralTests.class.st
+++ b/src/BaselineOfGeneralTests/BaselineOfGeneralTests.class.st
@@ -25,6 +25,7 @@ BaselineOfGeneralTests >> baseline: spec [
 			package: 'Collections-Arithmetic-Tests';
 			package: 'ConfigurationCommandLineHandler-Tests';
 			package: 'MetacelloCommandLineHandler-Tests';
+			package: 'Debugger-Filters-Tests';
 			package: 'Debugger-Model-Tests';
 			package: 'EmergencyDebugger-Tests';
 			package: 'EmbeddedFreeType-Tests';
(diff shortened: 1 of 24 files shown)


MarcusDenker:
Failing test: 

ReleaseTest>>testExplicitRequirementMethodsShouldBeImplementedInTheirUsers

MarcusDenker (approved):
ok

PR #6906, opened 2020-07-19
Variables-deprecate-isUndeclared

MarcusDenker:
- do the deprecation of #isUndeclared to #isUndeclaredVariable
- fix all senders
- move all deprated methods to Deprecated90

### DIFF
--- a/src/AST-Core/RBVariableNode.class.st
+++ b/src/AST-Core/RBVariableNode.class.st
@@ -80,7 +80,7 @@ RBVariableNode >> hasIncompleteIdentifier [
 	"check if there is any variable declared in my scope that starts with my name"
 
 	"declared vars are never incomplete"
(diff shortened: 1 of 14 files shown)


src/AST-Core/RBVariableNode.class.st
```diff
@@ -80,7 +80,7 @@ RBVariableNode >> hasIncompleteIdentifier [
 	"check if there is any variable declared in my scope that starts with my name"
 
 	"declared vars are never incomplete"
-	self binding isUndeclared ifFalse: [ ^false ].
+	self binding isUndeclaredVariable ifFalse: [ ^false ].
```
dionisiydk, line 83: still #binding here

dionisiydk (approved):

PR #6917, opened 2020-07-20
Simplify initializeCommandLineHandlerAndErrorHandling for default UI …

### LINKED ISSUE #6915: Simplify initializeCommandLineHandlerAndErrorHandling for default UI manager
- when setting the default UI manager we should use the #default: setter method that is available instead of putting an activated instance into the class var of UIManager (the method #default already does what we want)

astares:
…manager

Fix  #6915


MarcusDenker (approved):
ok

PR #6925, opened 2020-07-20
Variable-accessesRef-cleanup

MarcusDenker:
- add CompiledCode>>#accessesRef: 
- use it in LiteralVariable>>#isAccessedIn:
- simplify ClyVariable>>isAccessedIn:

### DIFF
--- a/src/Kernel/CompiledCode.class.st
+++ b/src/Kernel/CompiledCode.class.st
@@ -79,6 +79,13 @@ CompiledCode >> accessesField: varIndex [
 ]
 
 { #category : #scanning }
+CompiledCode >> accessesRef: literalVariable [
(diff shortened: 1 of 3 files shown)


src/Kernel/CompiledCode.class.st
```diff
@@ -78,6 +78,13 @@ CompiledCode >> accessesField: varIndex [
 	^ (self readsField: varIndex) or: [ self writesField: varIndex ]
 ]
 
+{ #category : #scanning }
+CompiledCode >> accessesRef: literalVariable [
```
dionisiydk, line 82: I think it needs a better name #accessesLiteralVariable: . And same for readsRef: and writesRef:

dionisiydk (approved):

PR #6926, opened 2020-07-20
Variables-isBlockVar-via-Variable

MarcusDenker:
#isBlockVar can be implemented much simplei via the variable.

in additon, use isDefinition in methodOrBlockArgStyleFor: which shows the intention of the code better.

### DIFF
--- a/src/AST-Core/RBArgumentNode.class.st
+++ b/src/AST-Core/RBArgumentNode.class.st
@@ -18,10 +18,3 @@ RBArgumentNode >> acceptVisitor: aProgramNodeVisitor [
 RBArgumentNode >> adaptToSemanticNode [
 	" I can't provide more semantics "
 ]
-
-{ #category : #testing }
-RBArgumentNode >> isBlockVar [
-	"true if a variable node is defined by a block node"
-
-	^ (self whoDefines: self name) isBlock
-]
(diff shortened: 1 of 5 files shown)


astares:
@MarcusDenker 

we have #isClassVariable but #isBlockVar ... why not use isBlockVariable?
Maybe this is for another issue - but we should use proper names.

MarcusDenker:
yes, I often do first just the implementation change and then later clean naming independently. 

But isBlockVariable is not good: you would think that there would be a BlockVariable class.

so the real name should be something like "isDefinedByBlock" or similar.

Ducasse:
I do not get why RBProgramNode does not need to define isBlockVar, a better name should be IsBlockVariable to be coherent with isSelfVariable.

Ducasse:
Removing isBlockVar from ProgramNode means that we will have check that we are on a VariableNode first. 
Or to be sure that are. And again I would let it so that people can avoid to use isKindOf:

MarcusDenker:
I do not get why RBProgramNode does not need to define isBlockVar --> because you should call it only on Variables. Like isDefinition.

MarcusDenker:
IsBlockVariable and isSelfVariable are different: a Variable is either a SelfVariable or not. this "isBlock" is like isDefinition: it can be true for nodes that are isArgumentVariable and for nodes that are isTempVariable.

MarcusDenker:
(the only user of isBlockVar is Shout... it only calls it on Variables. To be in sync with #isDefinition, it is good that it is *not* defined on RBProgramNode. It was a mistake. it should never have  been defined there)

Ducasse:
Ok may be I did not get it. 
The method comments should be better.

PR #6930, opened 2020-07-21
Variable-isSelf-isSuper-deprecation

MarcusDenker:
- On the level of the Variable hierarchy, deprecate isSelf, isSuper and isSelfOrSuper --> *Variable
- reduce senders of isSelfOrSuper (I think we should remove it)
-

### DIFF
--- a/src/Deprecated90/OCSelfVariable.class.st
+++ b/src/Deprecated90/OCSelfVariable.class.st
@@ -26,16 +26,6 @@ OCSelfVariable >> initialize [
 	name := 'self'
 ]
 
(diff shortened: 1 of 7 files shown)


Ducasse:
The changes are good. Now I think that this is important to keep isSelfVariable and other so that people building tools do not have to reintroduce it. Because sometimes we need it.

MarcusDenker:
isSelfVariable: of course. the question is if we need isSelfOrSuperVariable, for the 3 cases they can just do the "or" themselves

PR #6931, opened 2020-07-21
executeRenameRefactoringBy-move-to-Variable

MarcusDenker:
move the executeRenameRefactoringBy: extensions methods from the RBVariableNode subclasses to the Variable hierarchy

### DIFF
--- a/src/NautilusRefactoring/GlobalVariable.extension.st
+++ b/src/NautilusRefactoring/GlobalVariable.extension.st
@@ -1,6 +1,6 @@
 Extension { #name : #GlobalVariable }
 
 { #category : #'*NautilusRefactoring' }
-GlobalVariable >> executeRenameRefactoringBy: aNautilusRefactoring inClass: aClass [
-	aNautilusRefactoring renameClass: value 
+GlobalVariable >> executeRenameRefactoringBy: aNautilusRefactoring [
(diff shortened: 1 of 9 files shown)


Ducasse:
We will have to think how to rename and where to place this nautilus refactoring.

src/NautilusRefactoring/GlobalVariable.extension.st
```diff
@@ -1,6 +1,6 @@
 Extension { #name : #GlobalVariable }
 
 { #category : #'*NautilusRefactoring' }
-GlobalVariable >> executeRenameRefactoringBy: aNautilusRefactoring inClass: aClass [
-	aNautilusRefactoring renameClass: value 
+GlobalVariable >> executeRenameRefactoringBy: aNautilusRefactoring [
+	aNautilusRefactoring renameClass: self definingClass
```
dionisiydk, line 5: It looks like a bug. There is no definingClass for globals

guillep: Probably because there are no tests about this.

MarcusDenker: yes, but if we require 100% perfect tests before doing anything, we should stop now

PR #6933, opened 2020-07-21
Variables Refactoring: isSelfVariable-Part2

MarcusDenker:
- Implement isSelfVariable, isSuperVariable and isThisContextVariable on RBProgramNode
- do *NOT* change isSelf, isSuper  yet
- deprecate #isThisContext
- fix many users of isSelf, isSuper
- replace all users of isSelfOrSuper (there were two, one should use #isReservedVariable...)

More to follow, the other cases require more then just a trivial replacement (as they asume #isSelf to return a good result on the non-name analysed AST)


Ducasse:
Looks good to me. Once the refactoring will be finished I will check all the code of our little interpreter and we will have to check with carolina.

PR #6935, opened 2020-07-21
Variables-isReferenced-cleanup

MarcusDenker:
- Implement isAccessedIn: , isReadIn: and #isWrittenIn: on the whole Variable hierachy

- behavior #hasMethodsAccessingSlot: can now be generalised to hasMethodAccessingVariable:

- isReferenced on the LiteralVariable Hierachy can now use #hasMethodAccessingVariable:

- simplify #isReferenced on Behavior, implement it on MetaClass

### DIFF
--- a/src/Kernel/Variable.class.st
+++ b/src/Kernel/Variable.class.st
@@ -100,6 +100,11 @@ Variable >> hash [
 ]
 
 { #category : #testing }
+Variable >> isAccessedIn: aMethod [
+	^self usingMethods includes: aMethod
+]
+
+{ #category : #testing }
 Variable >> isArgumentVariable [
 	^false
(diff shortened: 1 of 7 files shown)


src/Kernel/Variable.class.st
```diff
@@ -192,6 +203,12 @@ Variable >> isWritable [
 	^ true
 ]
 
+{ #category : #testing }
+Variable >> isWrittenIn: aCompiledCode [
+	^aCompiledCode ast variableWriteNodes
+		anySatisfy: [ :node | node binding == self ]
```
dionisiydk, line 209: still #binding here

dionisiydk (approved):

PR #6941, opened 2020-07-21
6821-BlockClosureTest-should-not-compare-string-but-use-the-dumpVisitor

### LINKED ISSUE #6821: BlockClosureTest should not compare string but use the dumpVisitor
The tests should not be based on the default prettyprinter!

Check also newsource senders.

MarcusDenker:
we can just compare the ASTs in this case. fixes #6821


Ducasse:
Marcus is printString not using the pretty printer because each time we change it or select another one the tests may break.

MarcusDenker:
yes, printString uses the pretty printer. The test used to check the exact output, now instead it just checks that the AST after pretty printing is the same as before, which should always be the case. This way we still test printing, but without relying on exact output of the pretty printer.

PR #6948, opened 2020-07-22
RBComment-cleanup

MarcusDenker:
Now that RBComment is a subclass of RBProgramnode, we can remove the duplicated methods


guillep:
This PR is not tested. The only place properties are tested are for (the abstract) RBProgramNode.

image

I hope this does not break anything.

MarcusDenker:
yes, it's an API... it does not need to be re-tested for each subclass. 
(I did just copy the methods over as RBComment was not a subclass of RBProgramNode)

PR #6949, opened 2020-07-22
ReUtilityMethodsRule-disable

MarcusDenker:
ReUtilityMethodsRule is not very useful. I can not remember even *one* instance that I acted in it. But it is quite slow. I propose to disable it.


Ducasse:
Marcus instead of going one by one on rule and deciding based on your experience, we should look at the profile because rules that do not help you, will help novices and this is important to keep them.

MarcusDenker:
There is some support for beginner rules, but it is not used.. I will close this and check if we can use it

PR #6966, opened 2020-07-23
6965-speedup-senders-of-hasSelectorspecialSelectorIndex-check-too-much

### LINKED ISSUE #6965: speedup senders of: hasSelector:specialSelectorIndex: check too much
I think hasSelector:specialSelectorIndex: iterates too much in the current version.

- refersToLiteral: already dives into blocks
- so the loop could be just around the special var check:

'hasSelector: selector specialSelectorIndex: specialOrNil

	"refersToLiteral: traverses all blocks, but only for no-special literals"
	(self refersToLiteral: selector) ifTrue: [ ^ true ].
	specialOrNil ifNil: [ ^ false ].
	"if the selector is special, scan all blocks and myself"
	^ self withAllBlocks anySatisfy: [ :aCompiledCode | aCompiledCode scanFor:
				  self encoderClass firstSpecialSelectorByte + specialOrNil ]'

I will write some tests to check all the cases

MarcusDenker:
- add test #testHasSelector and #testHasSelectorSpecialSelectorIndex
- recategorize CompiledCodeTest
- implement hasSelector:specialSelectorIndex:  to only iterate the blocks for special selectors

fixes #6965

### DIFF
--- a/src/Kernel/CompiledCode.class.st
+++ b/src/Kernel/CompiledCode.class.st
@@ -295,15 +295,15 @@ CompiledCode >> hasSelector: selector specialSelectorIndex: specialOrNil [
 	 If you do, you may call this method directly to avoid recomputing 
 	 the special selector index all the time.
 	
-	 I traverse the method and all the compiled blocks in the literals 	 
-	"
+	 I traverse the method and all the compiled blocks in the literals"
 
-	^ self withAllBlocks
-		anySatisfy: [ :aCompiledCode | 
-			(aCompiledCode refersToLiteral: selector)
-				or: [ specialOrNil isNotNil
-						and: [ aCompiledCode
-								scanFor: self encoderClass firstSpecialSelectorByte + specialOrNil ] ] ]
+	
+	(self refersToLiteral: selector) ifTrue: [ ^ true ].
+	"refersToLiteral: traverses all blocks, but only for no-special literals"
+	specialOrNil ifNil: [ ^ false ].
+	"if the selector is special, scan all blocks and myself"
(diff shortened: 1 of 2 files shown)


MarcusDenker:
[ #thisSelectorDoesNotExist senders ] bench 

before: "'6.377 per second'"
after: "'13.764 per second'"

src/Kernel/CompiledCode.class.st
```diff
@@ -295,15 +295,15 @@ CompiledCode >> hasSelector: selector specialSelectorIndex: specialOrNil [
...
-						and: [ aCompiledCode
-								scanFor: self encoderClass firstSpecialSelectorByte + specialOrNil ] ] ]
+	 I traverse the method and all the compiled blocks in the literals"
+
+	
+	(self refersToLiteral: selector) ifTrue: [ ^ true ].
+	"refersToLiteral: traverses all blocks, but only for non-special literals"
+	specialOrNil ifNil: [ ^ false ].
+	"if the selector is special, scan all blocks and myself"
+	^ self withAllBlocks anySatisfy: [ :aCompiledCode | aCompiledCode scanFor:
```
dionisiydk, line 307: for more optimizations we should add #withAllBlocksDo: to avoid extra objects allocation.

dionisiydk (approved):

src/Kernel/CompiledCode.class.st
MarcusDenker: yes, another thing is that we should use the same pattern for all these methods, right now there are 3 ways that we implement a method like this... (besides this: have local* method doing it locally, then do a recursive call, third way is to  implement the method on all literals and just call it when iterating over the literals, leading to recursion on embedded blocks)

PR #6967, opened 2020-07-23
ReUtilityMethodsRule-speedup-and-test

MarcusDenker:
- speedup ReUtilityMethodsRule by implementing it using better APIs
- add visitor methods to the ones that do not raise it
- add test


MarcusDenker:
Strange test failure.. I will redo it with just the test

MarcusDenker:
test is now ok. failure not related

PR #6969, opened 2020-07-23
accessesRef-test-andspeedup

MarcusDenker:
- add a test for accessesRef:
- speedup the method by scanning literals, add comment

### DIFF
--- a/src/Kernel/CompiledCode.class.st
+++ b/src/Kernel/CompiledCode.class.st
@@ -81,8 +81,10 @@ CompiledCode >> accessesField: varIndex [
 { #category : #scanning }
 CompiledCode >> accessesRef: literalVariable [
 	"Answer whether this method accesses the LiteralVariable"
+	"we do not need to call readsRef: or #writesRef:, as if the variable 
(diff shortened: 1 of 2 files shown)


src/Kernel/CompiledCode.class.st
```diff
@@ -81,8 +81,10 @@ CompiledCode >> accessesField: varIndex [
 { #category : #scanning }
 CompiledCode >> accessesRef: literalVariable [
```
dionisiydk, line 82: Now we can remove this method actually. LiteralVariable can call #hasLiteral: directly.
But maybe it's good to have for consistency. I would at least rename it to #accessesLiteralVariable: (I do not get what Ref means here)

dionisiydk (approved):

src/Kernel/CompiledCode.class.st
MarcusDenker: yes, it was about getting a nice API on CompiledMethod: read/write and access for each kind of var (low level arg), then read/write/access for Var in general. renaming *Ref to *LiteralVariable is in the backlog.

PR #6978, opened 2020-07-25
ReInconsistentMethodClassificationRule-add-repair

MarcusDenker:
the ReInconsistentMethodClassificationRule tells us that the category is wrong, but there is no repair button.

- This Rule addes the repair button using the RBTransformation subclass RBMethodProtocolTransformation/
- add a compatibility method to RBTransformation to make them easily usable in the Rules


hogoww (approved):
Looks good to me :)

PR #6982, opened 2020-07-25
Use #do: in testCollectCreatesANewMatrix

### LINKED ISSUE #6981: Use #do: in testCollectCreatesANewMatrix
better use #do: instead of collect: as we just assert something and do not use the result of #collect:

astares:
Fix #6981


MarcusDenker (approved):
ok

PR #6988, opened 2020-07-27
ThisContextVariable-fast-usingMethods

MarcusDenker:
speed up #usingMethods of ThisContextVar the same as we have for super: scan bytecode instead of AST (which is slow if we have to create all the ASTs)

- add #readsThisContext to CompiledCode with Test
- implement usingMethods

This is now much faster:
ThisContextVariable instance usingMethods size

### DIFF
--- a/src/Kernel-BytecodeEncoders/EncoderForSistaV1.class.st
+++ b/src/Kernel-BytecodeEncoders/EncoderForSistaV1.class.st
@@ -795,6 +795,15 @@ EncoderForSistaV1 class >> pushNilByte [
 	^ 79
 ]
 
+{ #category : #'instruction stream support' }
+EncoderForSistaV1 class >> readsThisContextFor: compiledMethod [
+	"Answer whether compiledMethod reads thisContext"
+	| scanner |
+	scanner := InstructionStream on: compiledMethod.
+	^ scanner scanFor: 
+		[:instr |  instr = 82 ]
+]
+
 { #category : #'bytecode decoding' }
 EncoderForSistaV1 class >> selectorToSendOrItselfFor: anInstructionStream in: method at: pc [
 	"If anInstructionStream is at a send bytecode then answer the send's selector,
--- a/src/Kernel-BytecodeEncoders/EncoderForSistaV1.class.st
+++ b/src/Kernel-BytecodeEncoders/EncoderForSistaV1.class.st
@@ -797,7 +797,7 @@ EncoderForSistaV1 class >> pushNilByte [
 
 { #category : #'compiled method support' }
 EncoderForSistaV1 class >> readsSelfFor: compiledMethod [
-	"Answer whether compiledMethod reads thisContext"
+	"Answer whether compiledMethod reads self"
 	| scanner |
 	scanner := InstructionStream on: compiledMethod.
 	^ scanner scanFor: 
--- a/src/Kernel-BytecodeEncoders/EncoderForSistaV1.class.st
+++ b/src/Kernel-BytecodeEncoders/EncoderForSistaV1.class.st
@@ -795,6 +795,15 @@ EncoderForSistaV1 class >> pushNilByte [
 	^ 79
 ]
 
+{ #category : #'compiled method support' }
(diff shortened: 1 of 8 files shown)


src/OpalCompiler-Core/SelfVariable.class.st
```diff
@@ -35,3 +35,14 @@ SelfVariable >> isSelfVariable [
...
 ]
+
+{ #category : #queries }
+SelfVariable >> usingMethods [
+	"as super sends are doing a pushSelf, too, we need to still check the AST level sometimes"
+
+	^ SystemNavigation new allMethods select: [ :method | 
+		  method readsSelf and: [ 
+			  method sendsToSuper not or: [ 
+				  method ast variableNodes anySatisfy: [ :varNode |
```
dionisiydk, line 46: I think we do not need ast traversal here. It's enough to just detect a self send.
We can also add #sendsToSelf in addition to #sendsToSuper.

src/Kernel/CompiledCode.class.st
```diff
@@ -440,6 +440,13 @@ CompiledCode >> isRealPrimitive [
 	^self isPrimitive and: [ self isQuick not ]
 ]
 
+{ #category : #testing }
+CompiledCode >> isReturnSelf [
```
dionisiydk, line 444: #returnsSelf sounds better

src/Kernel-BytecodeEncoders/EncoderForV3PlusClosures.class.st
```diff
@@ -335,6 +335,24 @@ EncoderForV3PlusClosures class >> pushNilByte [
 	^ 115
 ]
 
+{ #category : #'compiled method support' }
+EncoderForV3PlusClosures class >> readsSelfFor: compiledMethod [
+	"Answer whether compiledMethod reads thisContext"
```
dionisiydk: a comment is about thisContext instead of self.

src/Kernel-BytecodeEncoders/EncoderForSistaV1.class.st
```diff
@@ -795,6 +795,24 @@ EncoderForSistaV1 class >> pushNilByte [
 	^ 79
 ]
 
+{ #category : #'compiled method support' }
+EncoderForSistaV1 class >> readsSelfFor: compiledMethod [
```
dionisiydk, line 799: There is #in: end in similar methods. And it sounds better: #readsSelfIn: and readsThisContextIn:

dionisiydk (approved):

src/OpalCompiler-Core/SelfVariable.class.st
dionisiydk: Notice I introduced singletons for these variables. So we have very simple queries like:
```Smalltalk 
SuperVariable instance usingMethods.
ThisContextVariable instance usingMethods
```

MarcusDenker: the problem is that "super something" actually is "pushSelf doSuperSend", so scanning for self access gives us all the super, too.

MarcusDenker: self can be referenced standalone in addition (e.g. self something ifTrue: [^self].)

src/Kernel-BytecodeEncoders/EncoderForSistaV1.class.st
MarcusDenker: I made it follow the existing isSuperSendFor: method...

MarcusDenker: The best here would be to have just the block on the encoder and do the scanning on the method in CompiedCode. Most other scanning methods do that. I have added that to the backlog

src/Kernel/CompiledCode.class.st
MarcusDenker: This is a prior existing method.

dionisiydk:
yes, you are right. Thanks for answers

PR #6989, opened 2020-07-27
Tests-hasLiteral-hasLiteralSuchThat

MarcusDenker:
add test for #hasLiteral and #hasLiterealSuchThat:

-> shows that the thwo methods implement a different semantics.


MarcusDenker:
The PR does not change anything. The refactoring is just so that the implementation of #hasLiteralSuchThat: follows that os #refersToLiteral: wrt to scanning the Pragmas (to not call #penultimateLiteral on the Blocks).

dionisiydk (approved):

PR #6990, opened 2020-07-27
Reduce-References-RBVariableNode-subclasses2

MarcusDenker:
another pass to reduce the users of RBVariableNode references. Instead, use RBVariableNode. The name analysis takes care of the rest 

There are some left that will be done in following PRs:

RBSelfNode
		- NewTools: testShouldConsiderHaltNode
		- identifierNamed:at: 
		
RBSuperNode
		- identifierNamed:at: 
		- rewriteASTToSimulateExecutionInADifferentContext:
		- testDoesApplyForSuperMessageSendNodes
		
RBThisContextNode
		identifierNamed:at: 
		testThisContextNodeDump -> needs visitor
		testBuildThisContextHeuristic -> needs visitor

### DIFF
--- a/src/AST-Core/RBSequenceNode.class.st
+++ b/src/AST-Core/RBSequenceNode.class.st
@@ -176,7 +176,7 @@ RBSequenceNode >> addSelfReturn [
 	| node |
 	self lastIsReturn
 		ifTrue: [ ^ self statements last ].
-	node := RBReturnNode value: (RBSelfNode new).
+	node := RBReturnNode value: (RBVariableNode named: 'self').
 	^ self addNode: node
 ]
 
--- a/src/AST-Core/RBSequenceNode.class.st
+++ b/src/AST-Core/RBSequenceNode.class.st
@@ -176,7 +176,7 @@ RBSequenceNode >> addSelfReturn [
 	| node |
 	self lastIsReturn
 		ifTrue: [ ^ self statements last ].
-	node := RBReturnNode value: (RBVariableNode named: 'self').
+	node := RBReturnNode value: RBVariableNode selfNode.
 	^ self addNode: node
 ]
 
--- a/src/OpalCompiler-Core/RBMethodNode.extension.st
+++ b/src/OpalCompiler-Core/RBMethodNode.extension.st
@@ -38,7 +38,7 @@ RBMethodNode >> ensureCachedArgumentNames [
 RBMethodNode class >> errorMethodNode: selector errorMessage: messageText [	
 	| message |
 	message := RBMessageNode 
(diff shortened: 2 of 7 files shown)


src/AST-Core/RBSequenceNode.class.st
```diff
@@ -176,7 +176,7 @@ RBSequenceNode >> addSelfReturn [
 	| node |
 	self lastIsReturn
 		ifTrue: [ ^ self statements last ].
-	node := RBReturnNode value: (RBSelfNode new).
+	node := RBReturnNode value: (RBVariableNode named: 'self').
```
dionisiydk: We should avoid such strings ('self', 'super'). Pablo introduced something like "RBVariableNode selfNode" in his PR. We need to push it here.

src/OpalCompiler-Core/RBMethodNode.extension.st
```diff
@@ -38,12 +38,12 @@ RBMethodNode >> ensureCachedArgumentNames [
 RBMethodNode class >> errorMethodNode: selector errorMessage: messageText [	
 	| message |
 	message := RBMessageNode 
-		receiver: RBSelfNode new
+		receiver: (RBVariableNode named: self)
```
dionisiydk: and here is a bug due to 'self' duplications.

dionisiydk (changes requested):

src/AST-Core/RBSequenceNode.class.st
MarcusDenker: My idea was to make sure that people understand that Variables have no meaning before name analysis. We need to be consistent: We can not do it for temps (humanly impossible to know what kind of temp and which offset). Sp why do it for Self?

src/OpalCompiler-Core/RBMethodNode.extension.st
MarcusDenker: fixed

src/AST-Core/RBSequenceNode.class.st
MarcusDenker: Hmmm "RBVariableNode selfNode" makes sense... but only for self, super and thisContext.

dionisiydk: Just that there is a duplication of code now. 
But this particular method has already semantics meaning: we return a self. So it needs to build a valid structure.

MarcusDenker: I will update the OR to use #selfNode, superNode and thisContextNode

dionisiydk (approved):

PR #6991, opened 2020-07-27
refersToLiteral-DoNotLookIntoProperties

MarcusDenker:
There is some *really* strange thing implemented for Method Properties: #refersToLiteral: checks both the key and, if the value is an array or a symbol, even that.

This makes no sense: of course we want to look into Pragmas. But properties? This means that if we put some property (e.g. #sendHalt returning true), it will find all methods as method refering to the property key.

We should see properties more as "dynamic object specific ivars", just additional state that can be assiged to a compiledMethod.  #refersToLiteral: should *not* bother!

This, as  side effect, speeds up senders of a little (and a lot if you use method properties a lot)


MarcusDenker:
To check the current behavior: check senders of #isFFIMethod. You get *lots* of methods that do not reference #isFFIMethod any form. 

"But that is nice, this way we can query for properties". Nope: if someone put false as the value if the property, it still answers to to #referstoLiteral:. It's just a bad idea.

MarcusDenker:
would it be ok to merge this?

Issue #6877, opened 2020-07-17
"run tests" on a package or a tag


hogoww:
**Describe the request**
The ctrl+T/right click + "run tests" on a package allows to run tests of all classes in the package.
It would be great to have it for tags as well.

**Expected behavior**
Be able to "run tests" on a package or a tag.

**Expected development cost**
It doesn't look that complicated to do, but with very little knowledge of calypso, it took me quite a while to just start to grasp the necessary concepts.
If nobody else does it, I'll pick this up later.

hogoww:
@dionisiydk I'll try do it for tomorrow's sprint, unless you want to take it =)

dionisiydk:
Hi @hogoww . I do not have time for this. 
It is not very trivial as you may think. You will need to implement something like ClyRunTestsFromClassGroupCommand.
But the trick to allow calypso to recognize tests in any kind of selected items is to annotate them with ClyTestResultProperty. See ClySUnitEnvironmentPlugin>>#decorateBrowserItem:ofPackage:. You will need to implement #decorateBrowserItem:ofClassGroup:.

hogoww:
I do not know if it's trivial as this is very dependent on how this is coded (which I don't know) and that I have very little experience with GUI.

But thanks for the answer and the hints !

hogoww:
@dionisiydk is there some sort of cache for query classes or something?
I tried to put my code on a new image twice, and it doesn't work *right away*
it starts working after a *while* (I wasn't able to narrow it down).
The general problem is that the item selected is wrongly a package instead of a tag.

It works after a while with those changes:
https://github.com/hogoww/pharo/commit/b10a7b81acab98d2ee6ec3027fc0143eb7522c6c

dionisiydk:
> @dionisiydk is there some sort of cache for query classes or something?

Yes. There is a cache of queries. In context menu there is a command Extra/Update to forcibly update it.

hogoww:
My apologies, I don't know where the context menu is ^^"

dionisiydk:
On any list (class list, method list, ...)

пн, 27 июл. 2020 г., 14:19 hogoww <notifications@github.com>:

hogoww:
Thank you :) 
It seems unrelated.
I'm back on : "The general problem is that the item selected is wrongly a package instead of a tag" (until it's correctly typed)
I have no idea where that can come from currently.
(and reaching that conclusion is super weird to me..)

dionisiydk:
[...]

PR #6992, opened 2020-07-27
"Run test" from tags

hogoww:
As described in #6877(thanks to @dionisiydk !)
Enables the use of run tests from a tag.
I do not know how to test it though.

(if someone wants to reproduce it locally, use `ClassAnnotation resetCache.`)


dionisiydk (approved):

Ducasse:
To me running tests should be the responsibility of a TestRunner (not UI) object that can be used by all the users.

Ducasse:
So Integrate this change but really the way tests are executed should be revisited.

hogoww:
I kind of agree.
To me, UI can have shortcuts to run tests (such as this), but should delegate to the test runner.
Another change that I would like to do (in that sense) is: when getting a result from Calypso's tests, clicking on it open a test runner to browse the test results.

But anyway, that was just a "quick" shortcut for calypso in the meantime.

PR #6995, opened 2020-07-27
Fixes #6993 against P9: Incorrect error message on failed parsing of a number with high radix

### LINKED ISSUE #6993: Incorrect error message on failed parsing number with high radix
An error message specifying an invalid character as the upper bound for literal numbers with a high radix is shown by the parser.

To reproduce: perform an inspect on `123r` (ie high radix and no digits after the radix indicator). The following error message is shown: `Reading a number failed: a digit between 0 and ± expected ->`. But only `$A` - `$Z` are valid digits (even for a very high radix).

The message should therefore read: `Reading a number failed: a digit between 0 and Z expected ->`

Error is present in at least Pharo 8.0 and 9.0 and is OS independent.

I will send a PR for this issue.

ErikOnBike:
This is a fix for the incorrect error message on failed parsing number with high radix.
This is PR against P8 branch.

### DIFF
--- a/src/NumberParser/NumberParser.class.st
+++ b/src/NumberParser/NumberParser.class.st
@@ -377,11 +377,12 @@ NumberParser >> nextScaledDecimal [
 NumberParser >> nextUnsignedIntegerBase: aRadix [ 
 	"Form an unsigned integer with incoming digits from sourceStream.
 	Fail if no digit found.
+	Since only characters $A - $Z are supported, limit the character range shown on error.
 	Count the number of digits and the lastNonZero digit and store int in instVar "
 	
 	| value |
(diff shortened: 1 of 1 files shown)


MarcusDenker:
strangely the CI only ran against P8... I will close  and re-open to force a new run

MarcusDenker:
PR was checked against P9

StephanEggermont:
Please rename

ErikOnBike:
A hint what you expect would be helpful.

StephanEggermont:
The list of PRs is not readable when the title only refers to an issue number. A description of the issue and perhaps the chosen solution helps with understanding the stream of changes

PR #6998, opened 2020-07-28
Adding support for Commander2 commands in Calypso context menu

tesonep:
Extending commander SycOpenSourceCodeMenu to handle commander2 mixed commands


src/Commander-Activators-ContextMenu/CmdOpenContextMenuCommand.class.st
```diff
@@ -95,7 +95,7 @@ CmdOpenContextMenuCommand >> registerContextMenuItemsFor: aCommandItem withBuild
 	RecursionStopper during: [ 		
 		submenu := self activationStrategy buildMenuInContext: context.		 
 		submenu isActive ifTrue: [ 					
-			super registerContextMenuItemsFor: aCommandItem withBuilder: aBuilder.
+			super doRegisterContextMenuItemsFor: aCommandItem withBuilder: aBuilder.
```
guillep: is super needed now that the method got renamed?
Or looking at the refactoring this could still be the same version as before: `super registerContextMenuItemsFor: aCommandItem withBuilder: aBuilder.` right?

PR #6999, opened 2020-07-28
Ring2-Compatibility-isVariable

MarcusDenker:
make sure that the is*Variable related methods are up do date in Ring2


MarcusDenker:
This just adds some is* methods to Ring2 that Moose needs. (We should look into how Moose uses Ring2 and add tests to Pharo that cover this use...)

PR #7005, opened 2020-07-29
PDF and SVG cairo exporters: do not require garbage collection to flush the file

### LINKED ISSUE #7004: PDF and SVG cairo exporters: do not require garbage collection to flush the file
The cairo library requires that the following destroy functions are called to flush the internal buffer to disk.
```C
  cairo_surface_destroy(surface);
  cairo_destroy(cr);
```
Which is translated to our bindings as:
```smalltalk
	self destroySurfaceHandle: handle value
	self destroyContextHandle: contextHandle value. 
```
performed by `AthensCairoSurface class>>#finalizeResourceData:` which is executed when the object is finalized.
So, in the current implementation, we suggest to let the garbage collector finalize the instance, which will flush to disk (file in disk may have 0 bytes before).

For example, the tests execute this code in `AthensCairoExportSurfaceTest>>tearDown`:
```Smalltalk
flushExportFile
	"This triggers call to destroy functions (the garbage collector will do it via finalize)"

	surface := nil.
	Smalltalk garbageCollect
```

**Describe the request**
It would be nice to rewrite this method to do not use the Morphic dependency at all and do not use explicit GC because that is a huge performance killer.
(Pavel)

**Expected behavior**
The same behavior but triggered explicitly via some method in `AthensCairoExportSurface`, without requiring to do a gc.
However, if such method was not executed explicitly, the gc will come at some point and it has to do it.

**Expected development cost**
I'm working on it... <1 hour?

**Additional context**
Reported originally here: https://github.com/ObjectProfile/Roassal3/issues/246

tinchodias:
Fixes #7004


Ducasse:
Thanks a lot for the nice comments and all the work.

PR #7006, opened 2020-07-29
Variables-definedVariables-tests

MarcusDenker:
- add Test for #definedVariables and make sure it works on the class side

In addition, do some trivial cleanups:

- tempNamed: does not need the temp
- #manifestOf: should nake sure to use the optimized ifNotNil:
- isAccessedIn: can be a bit clearer

### DIFF
--- a/src/Kernel/Behavior.class.st
+++ b/src/Kernel/Behavior.class.st
@@ -631,7 +631,7 @@ Behavior >> deepCopy [
 	^ self shallowCopy
 ]
 
-{ #category : #'as yet unclassified' }
+{ #category : #'accessing instances and variables' }
 Behavior >> definedVariables [
 	"return all the Variables defined"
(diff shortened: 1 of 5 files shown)


src/Kernel/Behavior.class.st
```diff
@@ -631,6 +631,12 @@ Behavior >> deepCopy [
 	^ self shallowCopy
 ]
 
+{ #category : #'accessing instances and variables' }
+Behavior >> definedVariables [
+	"return all the Variables defined"
+	^ self slots
```
dionisiydk, line 637: It should use #localSlots. Class does not define variables of its traits.
For all variables we need another method #ownedVariables (with name following slot property #owningClass).

MarcusDenker: Thanks! added to backlog

PR #7008, opened 2020-07-29
ReservedVariable-nameIsReserved

MarcusDenker:
We need an easy way to check if a name is the name of a Rerverved var in a way that it does not hardcode the string. This wil (after we use this everywhere) allow us to define new reserved vars (e.g. thisProcess to use the new bytecode) by just adding a subclass to ReserverdVariable.


MarcusDenker:
sadly needs to be redone. PR is not loadable, thus I can not fix it.

PR #7009, opened 2020-07-29
6751-Syntactically-wrong-code-can-be-compiled-without-Error

### LINKED ISSUE #6751: Syntactically wrong code can be compiled without Error
**Describe the bug**
Compilation of syntactically wrong code should raise an error to at least warn the developer that something bad will happens.

**To Reproduce**
Execute (got from real use case):
```
Object subclass: #Test
	instanceVariableNames: ''
	classVariableNames: ''
	package: 'AAA'.
Test compile: 'initialize
	((Lifeware includesKey: #DummyPoolVar) 
		and: [(Lifeware at: #DummyPoolVar) notNil]) 
			ifFalse: 
				[| nameSpace |
				nameSpace := #{NameSpace} value new.
				Lifeware at: #DummyPoolVar put: nameSpace.
				nameSpace
					environment: Lifeware;
					name: #DummyPoolVar.
				nameSpace at: #Var put: 1].
	(self importDefinitions 
		detect: [:e | e path = #(#Lifeware #DummyPoolVar)]
		ifNone: []) isNil 
		ifTrue: 
			[self setImportDefinitions: (Array 
						with: (#{GeneralNameSpaceImport} value path: #(#Lifeware #DummyPoolVar)))]'
```

**Expected behavior**
image
As per the Pharo 9.0.0
Build information: Pharo-9.0.0+build.348.sha.9675774d715aa7943f91d1501b073384441bce14 (64 Bit)

**Screenshots**
image

**Version information:**
 - OS: Win 10 - V2004
 - Version: 440

**Expected development cost**
5 mins to revert a wrong commit?
20 mins to add a test on this

MarcusDenker:
add a new compiler option #optionParseErrorsNonInteractiveOnly. Only if that one is enabled (or the more general optionParseErrors), we allow syntax errros in non-interactive mode

This includes two simple fixes that are relevant if you enable the option:

- implement #errorMessage on RuntimeSyntaxError to be compatible to SyntaxErrorNotification
- fix the one user that traps SyntaxErrorNotification to react to RuntimeSyntaxError

fixes #6751

### DIFF
--- a/src/Kernel/RuntimeSyntaxError.class.st
+++ b/src/Kernel/RuntimeSyntaxError.class.st
@@ -18,11 +18,16 @@ RuntimeSyntaxError class >> signal: aNode [
 ]
 
 { #category : #accessing }
+RuntimeSyntaxError >> errorMessage [
+	^errorNode errorMessage
+]
+
+{ #category : #accessing }
 RuntimeSyntaxError >> errorNode: aNode [
 	errorNode := aNode
 ]
 
 { #category : #accessing }
 RuntimeSyntaxError >> messageText [
-	^errorNode errorMessage
+	^self errorMessage
 ]
--- a/src/OpalCompiler-Core/CompilationContext.class.st
+++ b/src/OpalCompiler-Core/CompilationContext.class.st
@@ -269,6 +269,16 @@ CompilationContext class >> optionParseErrors: aBoolean [
 ]
 
 { #category : #'options - settings API' }
+CompilationContext class >> optionParseErrorsNonInteractiveOnly [
+	^ self readDefaultOption: #optionParseErrorsNonInteractiveOnly
+]
+
+{ #category : #'options - settings API' }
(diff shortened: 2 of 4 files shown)


VincentBlondeau:
Thanks Marcus for the fix! Could you tell what will be the default behavior for this?

MarcusDenker:
The default is to raise the error. I still believe that allowing compiler errors for non-interactive code is a good idea, but it needs more tool support.

MarcusDenker:
And of course the SyntaxErrorDebugger it now raises is broken... the whole thing started with "why fix it, we don' need that".

MarcusDenker:
So if you now want to load hundreds of methods with errors, you can decide: raise the SyntaxErrorDebugger for each (good luck), or enable the option and load all the methods, then query for all broken methods (#isFaulty on the AST ) and fix them from the normal browser.

I have on my todo to add a faster #isFaulty on the CompiledMethod that does not need to reify the AST if the method is clearly not faulty.

Ducasse:
Marcus we should document this.

PR #7010, opened 2020-07-29
Improving Taskit Execution time

tesonep:
Removing long waits for GC


sbragagnolo:
Great Pablo. Those waits were not necessary finally. I would have to have some powerfull garbage collect strategy for being able to remove all thos gc loops. Thanks!

PR #7016, opened 2020-07-30
ReservedVariable-nameIsReserved-and-Test

MarcusDenker:
This adds a nameIsReserved: check to ReservedVariable plus a test.

The idea is to use that in the compiler and e.g. when we check variable name validity. If we do that, we can add new Reserved Variables by just adding a subclass.

(this is not as simple to do to use it in Slot class as it has implications for dependencies, ReservedVariable is for now in the compiler package,
therefore this PR does not do that change yet. The Compiler change will be done im combination with removing all hard-coded references to the variable names)

### DIFF
--- a/src/Slot-Tests/ReservedVariableTest.class.st
+++ b/src/Slot-Tests/ReservedVariableTest.class.st
@@ -1,0 +1,18 @@
+Class {
+	#name : #ReservedVariableTest,
+	#superclass : #TestCase,
+	#category : #'Slot-Tests-VariablesAndSlots'
+}
+
+{ #category : #tests }
+ReservedVariableTest >> testNameIsReserved [
+ 	"we have these reserved vars: self, super, thisContext"
+ 	self assert: (ReservedVariable nameIsReserved: 'self').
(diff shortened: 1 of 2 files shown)


src/Slot-Tests/ReservedVariableTest.class.st
```diff
@@ -0,0 +1,18 @@
...
+}
+
+{ #category : #tests }
+ReservedVariableTest >> testNameIsReserved [
+ 	"we have these reserved vars: self, super, thisContext"
+ 	self assert: (ReservedVariable nameIsReserved: 'self').
+ 	self assert: (ReservedVariable nameIsReserved: 'super').
+ 	self assert: (ReservedVariable nameIsReserved: 'thisContext').
+
+  	"true and false and nil are compiled as lierals, not reserved vars"
```
StephanEggermont: literals

StephanEggermont (changes requested):

src/Slot-Tests/ReservedVariableTest.class.st
MarcusDenker: fixed!

PR #7017, opened 2020-07-30
CleanBlocks-Experiment

MarcusDenker:
!!!! Highly Experimental!!!

This is disabled by default.

If you want to experiment, you can enable the #optionCleanBlockClosure compiler option.

What does it do? For blocks that do not access anything outside (#isClean is true), it statically creates a FullBlockClosure and pushes it with #pushLiteral.

This avoids having to create the FullBlockClosure from the CompiledBlock at runtime, which is much faster.

This is not yet complete, see the class CleanBlockClosure for more comments and TODOs.

### DIFF
--- a/src/Kernel/CleanBlockClosure.class.st
+++ b/src/Kernel/CleanBlockClosure.class.st
@@ -1,0 +1,61 @@
+"
+!!!! Highly Experimental!!!
+
+This is disabled by default.
+
+If you want to experiment, you can enable the #optionCleanBlockClosure compiler option.
+
+- This should not be a subclass of FullBlock (as it does not need the receiver ivar, for example)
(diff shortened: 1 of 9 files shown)


MarcusDenker:
Maybe I need clarify: it is experimental, but:

- If the option is turned off, this is turned off
- the changes done in other parts of the system (isEmbeddedBlock instead of isCompiledBlock) are preparation for when this is turned on, but they do make sense for the current system. (e.g. this allowed to remove one isKindOf: in the compiler)

This means: we should merge this. It then allows people to recompile the system to do e.g. benchmarks. I used a recompiled system a bit and it is usable, the only problem is that debugging of cleanBlocks is not yet working.

MarcusDenker:
I will merge this so that we can easily test it (the code is only active if you turn on the preference for the compiler)

PR #7018, opened 2020-07-30
ReNotOptimizedIfRule

MarcusDenker:
First step of a Rule that warns if you use ifTrue: ifFalse: ifTrue:ifFalse: ifFalse:ifTrue:  in a way that it can not be optimized.

- need a test
- need to run it on the whole image to see if we get any false positives

### DIFF
--- a/src/GeneralRules-Tests/ReSmalllintTest.class.st
+++ b/src/GeneralRules-Tests/ReSmalllintTest.class.st
@@ -389,6 +389,11 @@ ReSmalllintTest >> testNoClassComment [
 ]
 
 { #category : #tests }
+ReSmalllintTest >> testNotOptimizedIf [
+	self ruleFor: self currentSelector plusSelectors: {#ifTrueBlocks}
+]
+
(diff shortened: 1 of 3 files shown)


guillep:
We were just discussing with Pablo:
- since it should hold that x == x value, particularly for blocks
- And that the message send to ifTrue:ifFalse: will send `value` to the argument
- then the compiler could transform automatically `condition ifTrue: […] ifFalse: x` into `condition ifTrue: […] ifFalse: [ x value ]`

And we get the best of both worlds, right?
We write it nicely, and it avoids always the slow path.

We will work on that in the sprint :)

MarcusDenker:
That sounds great!

MarcusDenker:
I am 100% in agreement with Guille that this should be done automatically. 

I propose we add rules just until this is done. (ifTrue: is just one, there are more like and:, to:do: ...).

As soon as the system is then good enough to do these automatically, we remove the rule.

astares:
Great!

PR #7019, opened 2020-07-30
Changing defaults to retain blanks

guillep:
The formatter should by default:
 - retain blank lines between statements (we all the time group related statements and we are losing that information)
 - add an empty line between the method signature and the body


MarcusDenker:
failing test on all 3 platforms: RBDumpVisitorTest>>testDumpOnSelfClassMethods

MarcusDenker (approved):

PR #7020, opened 2020-07-30
Adding scopeLevel to requestor scope: it returns 0.

StevenCostiou:
Fixes a bug that we see in the nwe debugger.
There is no issue related, we've seen that problem with marcus.


MarcusDenker (approved):

MarcusDenker:
a test would be nice.. I add it to the backlog but merge this for now.

PR #7021, opened 2020-07-31
Playground-Color-defined-Variables

MarcusDenker:
When you create a binding in the workspace (a:=1 for exampe), that var should be recognized by the Pretty Printer .

This PR for now just uses the same #symbol as tempvar, but that can be improved later so we can define a special color.


src/Shout/SHRBTextStyler.class.st
```diff
@@ -1028,6 +1028,8 @@ SHRBTextStyler >> resolveStyleFor: aVariableNode [
 	"here we should add support for #classVar"
 	aVariableNode isClassVariable ifTrue: [ ^#globalVar].
 	aVariableNode isInstanceVariable ifTrue: [ ^#instVar]. 
+	"for now we use temp variable here, we could now color them with their own color"
+	aVariableNode isWorkspaceVariable ifTrue: [ ^#tempVar].
```
dionisiydk, line 1032: Now we can refactor it into "aVariableNode styleSymbol" and use it directly in senders of #resolveStyleFor:

PR #7022, opened 2020-07-31
fix #6937

### LINKED ISSUE #6937: Opening a Browser from a message list that contains class side methods
**Describe the bug**
Subscript out of bounds exception when browsing code

**To Reproduce**
1. implementors message list tat contains class methods (e.g. Go to the Playground, type "uniqueInstances", implementors of)
2. Click on one of the class side methods
3. Try to Browse (e.g. via the menu)

hogoww:
#6937


guillep:
The fact that tests passed means nothing. Because tests were green while the bug is there. So this PR should add a test too

hogoww:
This is not working, and I'm having a lot of trouble reproducing this in tests, so I'll just close it for now

PR #7026, opened 2020-07-31
6952 debugger line location is uncorrect when using a breakonce

### LINKED ISSUE #6952: Debugger line location is uncorrect when using a Breakonce
**Describe the bug**
I put a break there:
image

And the debugger opens there:
image

Clicking on restart doesn't change the position of the "Where is" selection (screenshot after clicking on restart):
image

And clicking on over just changes the stack top in the var list

**To Reproduce**
1. Execute:
```
Object compile:  'systemIconName
	self class isMeta ifTrue: [ ^ self class printString].
	^ #classIcon '.
```
2. Set a Break once breakpoint between "^" and "self". See up screenshot
3. Click on another class to refresh the browser icon list
4. See debugger

**Expected behavior**
The good frame should be selected in the method source code pane at the beginning, on restart and on step in.

**Version information:**
 - OS: win 10
 - Version: Pharo9.0.0
build: 541, commit: a3a6597

StevenCostiou:
Fixes #6952.
This fix looks simple but that is not because it has implications:

- Breakpoints once do not remove themselves after hitting, which fixes the highlight bug in the debugger
- Breakpoint now include (de)activation logic
- Breakpoints are now a property of the nodes on which they are installed
- Breakpoint icon styler has been fixed accordingly

Not that commit 5f4cbf3 is reverted by d408f74 (don't bother to review).

@VincentBlondeau please review.
@dionisiydk could you have a look?

### DIFF
--- a/src/NewTools-Debugger-Breakpoints-Tools/StHaltAndBreakpointController.class.st
+++ b/src/NewTools-Debugger-Breakpoints-Tools/StHaltAndBreakpointController.class.st
@@ -45,7 +45,8 @@ StHaltAndBreakpointController >> areHaltsAndBreakpoinsEnabledFor: aProgramNode [
 
 { #category : #control }
 StHaltAndBreakpointController >> disableBreakpointNode: aProgramNode [
-	aProgramNode allBreakpointLinks do: [ :link | link condition: [false]; disable ].
+	self flag: 'Fix for PR 6952: should be removed in newtools merge 4.2.2'.
+	aProgramNode breakpoints do: [ :bp | bp disable ].
 	self skipNode: aProgramNode
 ]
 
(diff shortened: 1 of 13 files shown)


StevenCostiou:
Additional comment: even if reviewed and test pass, do not merge yet.
I want to add an improvement on the gutter icons for breakpoints to differentiate them with options (once, variable breakpoints, etc.).

MarcusDenker (approved):
looks good

StevenCostiou:
Now it is a bit annoying :/ 
Failing tests are from NewTools, which should be fix in the merge and not through this PR.
I will fix them here and the fix will also be part of the merge.

StevenCostiou:
After thinking more: please review and merge.
I will do another PR with icon styler improvements.

MarcusDenker (approved):

Issue #7015, opened 2020-07-29
Add a comment to explain why it is better to have  [aBlock value] instead of aBlock


Ducasse:
```
keyAtIndex: anIndex ifAbsent: aBlock
	^ (anIndex > 0 and: [ anIndex <= self size ])
		ifTrue: [ orderedKeys at: anIndex ]
		ifFalse: [ aBlock value ]
```

=> 

```
keyAtIndex: anIndex ifAbsent: aBlock
	^ (anIndex > 0 and: [ anIndex <= self size ])
		ifTrue: [ orderedKeys at: anIndex ]
		ifFalse: aBlock
```

MarcusDenker:
No, the ifTrue:iFalse: compiled to jumps.. if the block is removed, it does not.

Ducasse:
If we never document this nobody will ever be able to grow knowledge. 
We should add a comment to the method. This is the minimum. 
I hate all this crap in Pharo that never lets me learn. 


That is the problem: only if the compiler statically sees blocks, it can compile #ifTrue: and friends to jumps.

So if you remove the block here, it will instead compile a real send and this compile a block for the #ifTrue: case, and a send #ifTrue:ifFalse:
making this very very slow.

This means this is a valid transformation for all the cases where it is *not* an optimised send. There it would speed things up (e.g. imagine do:),
as it would avoid a block creation.

For the optimised case, it slows things down.

The nice thing is that all this is statically decidable. e.g. I have already on my TODO do do a Rule that warns you when you use ifTrue: and friends
without blocks.

Bytecode:


'41 <40> pushTemp: 0
42 <50> pushConstant: 0
43 <63> send: >
44 <C4> jumpFalse: 50
45 <40> pushTemp: 0
46 <4C> self
47 <72> send: size
48 <64> send: <=
49 <B0> jumpTo: 51
50 <4E> pushConstant: false
51 <C3> jumpFalse: 56
52 <01> pushRcvr: 1
53 <40> pushTemp: 0
54 <70> send: at:
55 <B1> jumpTo: 58
56 <41> pushTemp: 1
57 <79> send: value
58 <5C> returnTop’

vs.

'49 <40> pushTemp: 0
50 <50> pushConstant: 0
51 <63> send: >
52 <C4> jumpFalse: 58
53 <40> pushTemp: 0
54 <4C> self
55 <72> send: size
56 <64> send: <=
57 <B0> jumpTo: 59
58 <4E> pushConstant: false
[...]

PR #7037, opened 2020-08-03
Compiler should rewrite the inlined methods that does not have a block.

tesonep:
- Add opal plugin to improve inlining
- Add tests for them
- Create a baseline to load them during the bootstrap process. 

It should make #7015 obsolete


Ducasse:
Pablo you got some Morphic in there too.

Ducasse:
Pablo what do we do of this?

PR #7051, opened 2020-08-06
Adding the tools to stylize the text with IconStyler in real time

lesueur-philippe:
Adding in calypso the underlining and overlining of error nodes.
There are two parts 
- the copy of IconStyler to "reset its sub-classes", with new sub-classes
- the modification of ClyMethodEditor to allow the styling in method tabs


Ducasse:
One solution to remove the duplication between iconStyler and iconStylerRT is probably to make sure that icon styler can get a list of the subclasses it will query instead of doing 

```
styleText: textModel withAst: ast [
 	self allSubclassesDo: [ :styleClass | 
 			styleClass shouldStyle
 				ifTrue: [ ast acceptVisitor: (styleClass style: textModel) ] ]
```

dionisiydk:
Hi @lesueur-philippe . 
Could you explain why do you consider the old styler as not real time? When I edit code the highlighting is definitely changed dynamically as I typed. What the difference?

lesueur-philippe:
Hi,
The real time styling of the text is made by SHRBTextStyler and only
changes the color and the typo of the texte. IconStylerRealTime uses the
same model as IconStyler to overline or underline the text and create icons
with messages on the left bar.
But IconStyler was doing it on compilation where IconStylerRealTime does it
when the text is changed.

So, it is a complementary styler to further the feedback on errors.

Le mar. 18 août 2020 à 18:43, Denis Kudriashov <notifications@github.com> a
écrit :

src/Reflectivity-Tools/IconStylerRealTime.class.st
```diff
@@ -0,0 +1,239 @@
...
+	^true
+]
+
+{ #category : #style }
+IconStylerRealTime class >> style: textModel [
+	^self new textModel: textModel.
+]
+
+{ #category : #api }
+IconStylerRealTime class >> styleText: textModel withAst: ast [
```
dionisiydk, line 24: I believe we can avoid separate styler hierarchy. We can have a method on IconStyler like:
```Smalltalk
IconStyler>>styleTextForRealTimeHighlighting: textModel withAst: ast
	self allSubclassesDo: [ :styleClass | 
			styleClass shouldStyleTextInRealTime ifTrue: [...]]
```
Then we just need to decide which classes should return true from #shouldStyleTextInRealTime.

dionisiydk (changes requested):
And you did not comment new classes.

src/Reflectivity-Tools/SemanticVariableIconStyler.class.st
```diff
@@ -0,0 +1,49 @@
+Class {
+	#name : #SemanticVariableIconStyler,
+	#superclass : #IconStylerRealTime,
+	#category : #'Reflectivity-Tools-RealTimeStyling'
+}
+
+{ #category : #styling }
+SemanticVariableIconStyler >> addIconStyle: aNode from: start to: stop color: aColor [
```
dionisiydk, line 8: It seems this method is duplicated in all styler classes. Maybe I do not see the difference. But it is also a sign to refactor common parts.

Ducasse:
Yes I agree denis. For now we will integrate it and I already created an issue https://github.com/pharo-project/pharo/issues/7126

PR #7052, opened 2020-08-06
Adding the nesting of the comments and strings with smart characters …

lesueur-philippe:
Adding the nesting of the comments and strings with smart charachters and copySelection.
Adding the unnesting of the comments and strings with smart characters and copySelection.
Paste and copySelection are dependent of RBProgramNode's 'bestNodeFor:' which does not yet include comment nodes.

bestNodeFor: will be taken care of in an other PR
The absence of modification only renders the changes useless for comments
but does not break the previous implementation of paste and copySelection.

Inside paste, there is commented code which hints at a potential nesting inside symbols.

### DIFF
--- a/src/Rubric/RubTextEditor.class.st
+++ b/src/Rubric/RubTextEditor.class.st
@@ -2390,7 +2390,7 @@ RubTextEditor >> shiftEnclose: aKeyboardEvent [
 ]
 
 { #category : #nesting }
-RubTextEditor >> shouldNestCar: aChar [
+RubTextEditor >> shouldNestCharacter: aChar [
 	^'"''' includes: aChar.
 ]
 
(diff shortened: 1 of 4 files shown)


src/Rubric/RubTextEditor.class.st
```diff
@@ -2343,6 +2389,11 @@ RubTextEditor >> shiftEnclose: aKeyboardEvent [
 	^true
 ]
 
+{ #category : #nesting }
+RubTextEditor >> shouldNestCar: aChar [
```
Ducasse: Can you rename this method shouldNestCharacter:?
Thanks

Ducasse:
Philippe 

I typed:

```
'l''iode'
```

I select it and I press '
I get 

```
'''l''''iode'''
```

Is it what we want?

Indeed in previous Pharo I got 

```
''l''iode''
```
which is broken

PR #7058, opened 2020-08-07
Safer completion class access

pavel-krivanek:
Editing of package comments can cause DNUs. Use safer access to the completion class.
(Problem reported and patch provided by the Richard Uttner)

### DIFF
--- a/src/HeuristicCompletion-Model/CoClassBasedFetcher.class.st
+++ b/src/HeuristicCompletion-Model/CoClassBasedFetcher.class.st
@@ -54,7 +54,7 @@ CoClassBasedFetcher >> forHierarchy [
 { #category : #'hierarchy-fetching' }
 CoClassBasedFetcher >> onSuperclass [
 	
-	completionClass superclass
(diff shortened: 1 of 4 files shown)


Ducasse:
Hi Pavel,
Guille is asking the bug you have because he would like to make it robust from a user perspective but not hiding the problem. Because may be there is a use not using correctly completion

Ducasse:
The editor should not configure the package comment to use a completion that requires a class because there is no class.

Ducasse:
I add you so that we do not forget to discuss it.

Ducasse:
May be my change to comment mode solved the problem.

guillep:
Yes, I'd prefer to avoid having nil checks everywhere if a correct configuration can issue the right behavior...
@pavel-krivanek do you have some reproducible cases so we can test with @Ducasse changes?

pavel-krivanek:
hmm, we are using slightly older version of P9 for development and it seems that this problem is no more valid. We can close this PR then. Thank you

Ducasse:
Thanks pavel let us know in case it reappears.

PR #7082, opened 2020-08-11
[WIP] Fixes #7057: Commenting out a piece of code containing comments should maintain the inside comments

### LINKED ISSUE #7057: Commenting out a piece of code containing comments should maintain the inside comments
Right now, if one selects the following piece of code and presses " to comment it, then the comment inside is no longer a comment and the piece of code does not compile.

```Smaltlalk
code.
"comment"
code.
```
-> select all and press " ->
```Smalltalk
"code.
"comment"
code."
```
-> this code does not compile because the inside comment is interpreted as code
**Expected Behaviour**
```Smalltalk
"code.
""comment""
code."
```

ackerleytng:
Still a work in progress! Kicking off this pull request first

### DIFF
--- a/src/NECompletion-Morphic/CompletionDefaultKeyboardEvent.class.st
+++ b/src/NECompletion-Morphic/CompletionDefaultKeyboardEvent.class.st
@@ -1,0 +1,5 @@
+Class {
+	#name : #CompletionDefaultKeyboardEvent,
+	#superclass : #CompletionKeyboardEvent,
+	#category : #'NECompletion-Morphic'
+}
--- a/src/NECompletion-Morphic/CompletionDoubleQuoteKeyboardEvent.class.st
+++ b/src/NECompletion-Morphic/CompletionDoubleQuoteKeyboardEvent.class.st
@@ -1,0 +1,50 @@
+"
+I represent a CompletionKeyboardEvent specialized for when the event's keyCharacter is $""
+"
+Class {
+	#name : #CompletionDoubleQuoteKeyboardEvent,
+	#superclass : #CompletionKeyboardEvent,
+	#category : #'NECompletion-Morphic'
+}
+
+{ #category : #'actions - enclosing' }
+CompletionDoubleQuoteKeyboardEvent >> replacementAfterEncloseActionOnSelection: selectedString [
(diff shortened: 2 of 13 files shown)


ackerleytng:
Here's my proposed implementation:

I chose to replace every double quote in the selection with two double quotes, for consistency:

+ `code "comment" code` becomes `"code ""comment"" code"`
+ `invalid " code` becomes `"invalid "" code"`
+ `code "" code` becomes `"code """" code"`
+ `invalid """ code` becomes `"invalid """""" code"`

If we had ignored paired double quotes, i.e. allow `code "" code` to become `"code "" code"`,
then we would not have been able to restore these two cases from the enclosed versions

+ `case " one` becomes `"case "" one"`
+ `case "" two` becomes `"case "" two"`

since the right hand side is identical.

Ducasse:
Pay attention that we will integrate before #7052

ackerleytng:
I was also looking into refactoring to try and use classes to represent keyboard events instead of using many if elses throughout the CompletionEngine code.

Another option would be to add methods to the KeyboardEvent class. What do people think of these two options?

ackerleytng:
@Ducasse, please feel free to merge #7052 and close both #7082 and #7057.

I saw some more issues surrounding TextEditor/Rubric at https://github.com/orgs/pharo-project/projects/20 and I would like to look into refactoring Rubric instead.

<blockquote><div> GitHub</div><div><strong><a href="https://github.com">Build software better, together</a></strong></div><div>GitHub is where people build software. More than 50 million people use GitHub to discover, fork, and contribute to over 100 million projects.</div></blockquote>

Ducasse:
You are welcome :)

Ducasse:
Right now the team is in slow mode because many of them are in vacation.

Ducasse:
What I would love to have is that when evaluating an expression the expression should be in "comment" and not directly modifying the text!

Ducasse:
I would like to know the status in latest version now that the parser and comments/strings escaping behavior got integrated.

guillep:
Hi @ackerleytng I think this behaviour is already in Pharo9 by default, and it covers also copy/past within comments. Could you check? I'll close the issue, tell us if this needs to be reopened.

PR #7088, opened 2020-08-13
Merge NewTools-0.4.2.4

### LINKED ISSUE #83: The operations of the Debugger should have shortcuts.

### LINKED ISSUE #90: Cancelling the Return popup DNU
Pharo-9.0.0+build.580.sha.9514baad3830e43a539d94c24719b0317623995e (64 Bit)

- Click on the Return button => a pop up is open
- cancel => BOOM

### LINKED ISSUE #92: The debugger code pane status' color is noisy
The color does not mean anything.
Users consider it noise.

### LINKED ISSUE #93: Exceptions/Errors in #printOn: make the debugger crazy
When there is an error in a #printOn: or one of those methods used by the debugger/inspector, the debugger/inspector fails to correctly open. Sometimes I get the emergency debugger, sometimes another debugger.
Clicking on a different frame in the stack makes the debuggers multiply.

 => Then I cannot fix the bug in the debugger. I have to close all debuggers and open fix it in a browser

### LINKED ISSUE #7038: [Pharo 9] Programming via debugger: adding methods to classes from the hierarchy
_An enhancement is a slight modification of an existing behavior. If you want a new feature use the 
feature request template._

**Describe the request**

Trying out Pharo9 now (the last version I used was 7 I believe). When I send non-existent messages and debugger opens, there's a "Create" button to create a new method in the receiver. Previously (in 7) it allowed to select the class where to create a new method, based on class hierarchy. Now (in 9) it by default creates a new method in the class the instance of which has received the message... This is not what I always would like to do -- is there a way to bring the old behavior back?

**Expected behavior**

Would love to see a way how to introduce methods to classes I select from the inheritance hierarchy -- especially useful when I want to patch my system's standard classes (like String or Number) with extension methods specific to my project.

**Expected development cost**

No idea

**Screenshots**

**Version information:**
 - OS: Ubuntu 18.04
 - Version:
```
Pharo9.0.0
build: 572, commit: 308d3ba
```

StevenCostiou:
https://github.com/pharo-spec/NewTools/releases/tag/v0.4.2.4

Contains debugger fixes and improvements:
- Fixes https://github.com/pharo-spec/NewTools/issues/83
- Fixes https://github.com/pharo-spec/NewTools/issues/90
- Fixes https://github.com/pharo-spec/NewTools/issues/92
- Fixes https://github.com/pharo-spec/NewTools/issues/93
- Fixes #7038
- The debugger now requires to define all variables before compiling from its code pane (as in Calypso editors)
- Added back test methods that were not pushed into NewTools

### DIFF
--- a/src/NewTools-Debugger-Breakpoints-Tools/StRemoveBreakpointCommand.class.st
+++ b/src/NewTools-Debugger-Breakpoints-Tools/StRemoveBreakpointCommand.class.st
@@ -1,0 +1,20 @@
+"
+Removes a breakpoint from a selected breakpoint item in a StObjectBreakpointInspection. 
+"
+Class {
+	#name : #StRemoveBreakpointCommand,
+	#superclass : #SpToolCommand,
+	#category : #'NewTools-Debugger-Breakpoints-Tools-Model'
+}
+
+{ #category : #default }
+StRemoveBreakpointCommand class >> defaultName [
+	^'Remove'
+]
+
+{ #category : #executing }
+StRemoveBreakpointCommand >> execute [
+
+	(self transform value: self context) removeInPresenterContext:
+		self context
+]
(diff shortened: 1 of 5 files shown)


StevenCostiou:
Failing tests seems unrelated.
Please review.

PR #7093, opened 2020-08-13
Issue 7092: Better HaloMorph Handles

darth-cheney:
Updating HaloMorph handles (the buttons around the halo) to look cleaner.
  
This entails:
- A slightly larger size for the handles (30px, was 20px);
- Completely circular handles (was rounded corners);
- Solid color background (was gradient, which was causing issues);
- No border (looks more cartoonish with the border)
  
The halos will now look like this:
  
HaloMorphNew


Ducasse:
Oh thanks!!!! This is great!

PR #7097, opened 2020-08-16
Fix equals method: check against anObject

ErikOnBike:
The #= method did inadvertently check the instance variable #mergedParts against itself instead of anObject's value for that instance variable.


Ducasse:
Indeed! Thx

PR #7105, opened 2020-08-16
deprecate-Context-lookupSymbol

MarcusDenker:
Context>>#lookupSymbol: is not well named... the name implies that it would do somehting similar as lookupVar:, but it actually reads the variable content.
- add #readVariableNamed: 
- fix the users (just in testss) of #lookupSymbol: to use #readVariableNamed:
- simplify #tempNamed: put:


Ducasse:
Oh yes this is much better. I would have never guessed that lookupSymbol: is not looking up a method

PR #7114, opened 2020-08-17
CleanBlock-pcInOuter

MarcusDenker:
this implements #pcInOuter for gfhe case that #optionCleanBlockClosure is enabled.

This makes #sourceNode work for clean blocks:

CompilationContext optionCleanBlockClosure: true.
[ 1 + 2 ] sourceNode


src/OpalCompiler-Core/RBVariableNode.extension.st
```diff
@@ -12,7 +12,7 @@ RBVariableNode >> binding: aSemVar [
 
 { #category : #'*opalcompiler-core' }
 RBVariableNode >> isClean [
```
dionisiydk, line 14: On the nodes level #isClean sounds quite bad. It would be nice to find better name

PR #7119, opened 2020-08-17
Deprecate-namedTempAt-namedTempAtput

MarcusDenker:
With closures, the idea that a temp variable has just an offset in the context is not true anymore. The story is *much* more complex!
To simplify it, we have added meta-objects that know how to read and write temps: we have made temps into first class variables.

The old "DebuggerMethodMap" API was providing some "virtual" offset based accessign scheme which now makes no sense at all anymore.

This PR thus deprecates namedTempAt:/namedTempAt:put:, the methods are not used anymore in the system and just by being there are highly confusing.

### DIFF
--- a/src/Debugging-Core/Context.extension.st
+++ b/src/Debugging-Core/Context.extension.st
@@ -44,7 +44,10 @@ Context >> methodReturnConstant: value [
 Context >> namedTempAt: index [
 	"Answer the value of the temp at index in the receiver's sequence of tempNames."
 	"NOTE: this list of temp names is completely virtual! It omits temp vectors but 
-	adds every variable that could be acccessed from a source perspective but might not be"
+	adds every variable that could be acccessed from a source perspective but might not be.
+	Do *not* use this!"
+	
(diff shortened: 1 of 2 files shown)


src/Debugging-Core/Context.extension.st
```diff
@@ -44,7 +44,10 @@ Context >> methodReturnConstant: value [
 Context >> namedTempAt: index [
 	"Answer the value of the temp at index in the receiver's sequence of tempNames."
 	"NOTE: this list of temp names is completely virtual! It omits temp vectors but 
-	adds every variable that could be acccessed from a source perspective but might not be"
+	adds every variable that could be acccessed from a source perspective but might not be.
+	Do *not* use this!"
+	
+	self deprecated: 'Please access temps by name using #tempNamed:'.
```
dionisiydk, line 50: These methods also need to be move to Deprecated90

dionisiydk (approved):

src/Debugging-Core/Context.extension.st
MarcusDenker: ok, in backlog

PR #7120, opened 2020-08-17
Missing offsets when mapping ast<->bytecode and when the first node is a fullblock

### LINKED ISSUE #7087: Debugger : step into doesn't work in this case.
**Describe the bug**
When doing step into inside visitNode:, the debugger redirect to a dictionary error

**To Reproduce**
1. Load commit 4a0b9b6
2. Put an halt inside RBParser>>parseIncompleteExpression
3. Open Playground and type: )
4. Step over until doSemanticAnalysis
5. Step into
6. Repeat step 4 and 5
7. Step into analyse:
8. Step into until visitMethodNode:
9. Step into visitNode:
10. Step into until visitSequenceNode:
11. Step through the block
12. Step into visitNode:
13. The next step into should cause the error

**Expected behavior**
KeyNotFound: key 57 not found in Dictionary

StevenCostiou:
Fixes #7087 needs review by @MarcusDenker 

When mapping ast to bytecodes, if the first node of the method is a block, sometimes the bytecodeIndex of the block IR is 2, sometimes it is 1.
When that bytecodeIndex is 2, it shifts the first mapped node by 1, and the initialPC of the method is not mapped.
Example: 
```Smalltalk
method: arg
  [arg] on: Error do: [nil]
```
Imagine method initialPC is `40`.
The first mapped node is `[arg]`, and is computed as :  `bytecodeIndex + startpc - 1` 
Its value is 2 + 40 - 1 = 41.
As it is the first mapped node, the mapping starts at `41`and the initiaPC `40` is missing from the mapping.
When debugging and entering a method with the debugger, if the node for the initialPC is requested then there is a key not found error because it is missing.

This PR fixes the mapping by checking that all indexes  between the initialPC of a method and the first mapped offset are all mapped to the first mapped node.
In the case above, it checks that all indexes from 40 to the first mapped offset (41->`[arg]`) are mapped, if not it maps 40 to 41 to the node mapped at 41 (`[arg]`).

I am not sure it is the *right* way to do this. Maybe it should be done during the mapping rather than afterwards by fixing missing indexes.

### DIFF
--- a/src/OpalCompiler-Core/OCBytecodeToASTCache.class.st
+++ b/src/OpalCompiler-Core/OCBytecodeToASTCache.class.st
@@ -35,20 +35,26 @@ OCBytecodeToASTCache >> bcToASTMap [
 
 { #category : #private }
 OCBytecodeToASTCache >> fillMissingBCOffsetsWithLastBCOffsetNodes [
-	"It happens that different bytecode offsets map to the same AST node.
+	 "It happens that different bytecode offsets map to the same AST node.
 	These cases are detected when there is no node mapped between, for example, bcOffset 46 and bcOffset 50.
 	In that case, we take every possible bytecode index between 46 and 50 (i.e., 47, 48, 49),
-	and we map them to the same node as the last mapped bytecode offset, here 46."
+	and we map them to the same node as the last mapped bytecode offset, here 46.
+	
+	When a fullblocks is the first node in a method body, it is possible (when?) the offsets do not start at the first offset of the method (why?). 
+	In that case, we also fix the missing offsets by mapping to them the first mapped node."
 
-	| sortedBCOffsets |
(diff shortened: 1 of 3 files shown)


StevenCostiou:
Failing tests are unrelated.

MarcusDenker:
At 40 we push the copied var on the stack, which will be consumed by the closure creation bytecode:


"'49 <40> pushTemp: 0
50 <F9 00 01> fullClosure:a CompiledBlock: [ aVisitor ] NumCopied: 1
53 <11> pushLit: Error
54 <40> pushTemp: 0
55 <F9 00 01> fullClosure:a CompiledBlock: [ aVisitor ] NumCopied: 1
58 <A2> send: on:do:
59 <D8> pop
60 <58> returnSelf'"

I think we can merge this as a workaround now but then should make sure that the mapping is correctly done later

MarcusDenker (approved):

PR #7131, opened 2020-08-19
remove-ReUnderscoreAssignmentRule

MarcusDenker:
remove ReUnderscoreAssignmentRule as we do not even allow to compile code with that anymore


Ducasse:
Yes!

PR #7162, opened 2020-08-21
Rename RubWorkspaceBar into RubWorkspaceBarRuler

### LINKED ISSUE #7111: Rename RubWorkspaceBar into RubWorkspaceBarRuler
This is a ruler.

sic-f:
Rename RubWorkspaceBar to RubWorkspaceBarRuler

Fixes #7111


astares:
We should property deprecated by making a copy and deprecating the old one. Otherwise we directly break peoples code

guillep:
Was this class for public consumption or only a private one for Rubric?

I understand the importance of deprecations but at least this should be
done with criteria and not blindly. And this decision should be properly
documented in comments and the issue.

This said, I'm ok for deprecating it if needed. But I want to make explicit
the need for deprecating (or not) :)

PR #7170, opened 2020-08-24
Size reporter

NicolasAnquetil:
Added a SizeReporter that computes various metrics over the system

### DIFF
--- a/src/Tool-SystemReporter/AbstractReporter.class.st
+++ b/src/Tool-SystemReporter/AbstractReporter.class.st
@@ -1,3 +1,14 @@
+"
+Abstract class of reporters:
+Classes displaying a window with information about the system. 
+Intended to be easily copy/pasted for external use (email, ...)
+
+Use by calling #open
+
+To extend a Reporter:
+- add a method  'reportXYZ: aStream'
+- insert a line 'add: #XYZ method: #reportXYZ' to #initialize
+"
 Class {
 	#name : #AbstractReporter,
 	#superclass : #Model,
--- a/src/Tool-SystemReporter/AbstractReporter.class.st
+++ b/src/Tool-SystemReporter/AbstractReporter.class.st
@@ -1,0 +1,190 @@
+Class {
+	#name : #AbstractReporter,
+	#superclass : #Model,
+	#instVars : [
+		'categories',
+		'categoriesSelected',
+		'categoryList',
+		'sourceTextModel'
+	],
+	#category : #'Tool-SystemReporter'
+}
+
+{ #category : #'instance creation' }
+AbstractReporter class >> open [
+	^ self new open
+]
+
+{ #category : #operations }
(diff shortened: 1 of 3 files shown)


src/Tool-SystemReporter/AbstractReporter.class.st
```diff
@@ -0,0 +1,201 @@
...
+Intended to be easily copy/pasted for external use (email, ...)
+
+Use by calling #open
+
+To extend a Reporter:
+- add a method  'reportXYZ: aStream'
+- insert a line 'add: #XYZ method: #reportXYZ' to #initialize
+"
+Class {
+	#name : #AbstractReporter,
```
dionisiydk, line 13: It would be nice to avoid such common names. Can you rename it to AbstractSystemReporter at least?

src/Tool-SystemReporter/SizeReporter.class.st
```diff
@@ -0,0 +1,84 @@
+"
+A subclass of AbstractReporter that computes various metrics over the image
+"
+Class {
+	#name : #SizeReporter,
```
dionisiydk, line 5: Also here. What size means here? 
It reports source code statistics. And better would be to name it accordingly like SourceCodeSizeReporter

src/Tool-SystemReporter/SizeReporter.class.st
```diff
@@ -0,0 +1,84 @@
...
+	self header: 'Number of LinesOfCode' on: aStream.
+	aStream 
+		nextPutAll: 'a lot !' ; cr
+
+]
+
+{ #category : #reporting }
+SizeReporter >> reportNbClasses: aStream [
+	| imageClasses |
+	imageClasses := Smalltalk globals allClasses.
```
dionisiydk, line 59: you should introduce #environment variable in superclass and use it here:
```Smalltalk
AbstractReporter>>initialize
   ...
   environment := self class environment

SizeReporter>>reportNbClasses: aStream
    classes := environment allClasses
```

dionisiydk: Also please use full names. "Nb" here is not good.

src/Tool-SystemReporter/SizeReporter.class.st
NicolasAnquetil: agreed

src/Tool-SystemReporter/AbstractReporter.class.st
NicolasAnquetil: I agree that AbstractReporter is not very informative, but AbstractSystemReporter is not much better.
Any other idea ?

src/Tool-SystemReporter/SizeReporter.class.st
NicolasAnquetil: agreed

PR #7172, opened 2020-08-24
Simplify-AddionalMethodState-LiteralQueries

MarcusDenker:
AdditionalMethodState should not care at all about the literal queries #hasLiteralSuchThat: and #refersToLiteral:

We just look into Pragmas, and that we can do from the CompiledMethod level by querying the pragmas directly.

This allows us to remove all the (very confusing) code related to literals from AdditionalMethodState

No deprecation needed as this is implementatoon details, not external API.


Ducasse:
When this is nicer to read and does the same then this is great.

PR #7174, opened 2020-08-24
Variable-writeInContext-should-return-Value

MarcusDenker:
#write:inContext: is a reflective version of a variable assignment. The value of "var := aValue" is aValue, thus the return of the reflective operation should do the same.

The clients (e.g. tempNamed:put:) show that this is indeed what we should do.

### DIFF
--- a/src/Debugging-Core/Context.extension.st
+++ b/src/Debugging-Core/Context.extension.st
@@ -380,8 +380,7 @@ Context >> tempNamed: aName [
 Context >> tempNamed: aName put: anObject [
 	"Assign the value of the temp with name in aContext"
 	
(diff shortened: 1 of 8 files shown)


Ducasse:
I read your comment but I would like to understand why the write need to return the value.

MarcusDenker:
if you evaluate "temp := 5", the value of the whole assignment is 5. Now if we repace the assignment with a "thisContext tempNamed: #temp put: 5" the result should be 5, too. This was done by #tempNamed: explicitly returning the value, showing that we better do it on the level of the variable #write:inContext: (as you could replace the assignment with "(self lookupTempVar: #temp) write: anObject inContext: self")

PR #7176, opened 2020-08-25
CleanBlock-accessVariables-ForDebugger

MarcusDenker:
The debugger needs to be able to access temps that are actually defined in the outerContext (and migh not even exist in the context itself).

For CleanBlocks, we can easily detect the case in Context>>#outerContext (outerContext is nil only in this case).

- implement #outerContext for the clean case
- testExampleAccessOuterFromCleanBlock tests it (example forces clean block comiplation using compiler options pragma)

With this the old Debugger works with Clean Blocks!

### DIFF
--- a/src/Kernel/Context.class.st
+++ b/src/Kernel/Context.class.st
@@ -1233,9 +1233,9 @@ Context >> objectSize: anObject [
 { #category : #accessing }
 Context >> outerContext [
 	"Answer the context within which the receiver is nested."
-
+	
 	^closureOrNil ifNotNil:
-		[closureOrNil outerContext]
+		[closureOrNil outerContext ifNil: ["if the outer is nil, this is a CleanBlock" self sender]]
 ]
 
 { #category : #accessing }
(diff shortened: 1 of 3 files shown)


MarcusDenker:
This is actually quite fun that we can access temps via the context in Clean Blocks. e.g. we could change Dictionary>>#at:

at: key 
	"Answer the value associated with the key."
	<compilerOptions: #(+ optionCleanBlockClosure)>

	^ self at: key ifAbsent: [thisContext receiver errorKeyNotFound: (thisContext readVariableNamed: #key)]

and it is much faster:

[Smalltalk globals at: #Object] bench 

before: "'13428372.326 per second'"
after: "'18923645.013 per second'"

MarcusDenker:
The downside is that it requires the AST. But we can refactor at:ifAbsent:to pass the key in the block via #cull and just
use the thisContext receiver instead of self to have a clean block.

guillep:
I personally prefer readable code to fast code at this point in the system.
We should get optimizations in some other way. Please don't add the
thisContext there

Le mar. 25 août 2020 à 10:03, Marcus Denker <notifications@github.com> a
écrit :

MarcusDenker:
yes, I did not want to do that change... I just found it an interesting observation. A real solution needs to do that (kind of) transparently. E.g. we could have some kind of compiler hint and then compile the block like that, even though the code stays readable. But all that is for after clean blocks are used for real.. so nothing for now.

MarcusDenker:
(the PR does not do this change, it just makes temp reading work for clean blocks)

dionisiydk (approved):

PR #7180, opened 2020-08-25
Merge request: Oups new debugger infrastructure

### LINKED ISSUE #7085: [Debugger]  changing the debugger in the setting does not change the debugger.
How can we get back the old debugger?

StevenCostiou:
This PR contains **Oups** and changes to merge it into Pharo 9.

**Oups** is a **Debugger Infrastructure**, that controls in a unified way how debuggers are found in the system and how they are opened - including error handling for errors occuring while opening debuggers, which previously ended up in the mighty emergency evaluator.

The goal of **Oups** is to facilitate the understanding, the evolution and the addition of system debuggers and their mechanics. To do so it regroups all debugger opening code under a unified mechanism that consists in creating debugger opening requests and submitting them to the system. The system then decides how to handle those requests depending on the UI mode.

# Infrastructure overview
Screenshot 2020-08-25 at 12 59 13

# Changes
1. Adds **Oups** (French for *Oops*, like in *Oops my debugger failed again*), a new debugger infrastructure:
- single debugger-opening entry point named `OupsDebugRequest`
- the best available debugger is selected through a debugger selection strategy
- if a debugger fails to open, it automatically selects another one to not disrupt the user
- a new mode *handle debugger errors* that open others debuggers to debug exceptions raised from inside a debugger
- to be automatically recognized as a debugger by the system, tools must now use the `TDebugger`trait
2. Cleans debugger settings:
- debuggers are now (de)activable in the settings
- each debugger has its own priority
3. Cleans the debugger opening call chain that were in `MorphicUIManager` and in `Process` 
4. Removes `Smalltalk tools debugger/debuggers`:
- remove registration from all debuggers
- remove accesses from all the system and replace them by requests to **Oups**
- cleans `Smalltalk tools`
5. System and tool fixes for this merge:
- debugging settings
- NewTools
- GT debugger
- EmergencyDebugger
- Baselines and dependencies

# Notes
- Includes changes to NewTools that are not yet pushed nor merged from NewTools. These changes will be pushed on NewTools once this PR is merged to avoid problems while loading NewTools in current Pharo image (before the merge).
- A technical paper was written and submitted to [IWST](https://esug.github.io/2020-Conference/cfpIWST2020.html), and contains details about the **Oups** infrastructure and the motivation behind it. We'll make it available ASAP to help reviewing that PR.
- This will require more work after merging. For example, the API to request debuggers could be improved and also centralized. Debugging/Debugger settings should be revisited and reclassified. Etc. However, this PR is big enough, and these changes can be studied after when a stable version of **Oups** is merged.

# Authors
@StevenCostiou 
@dupriezt 
@cdlm 

Please preferably tag all of us if any problem or issue related to **Oups**.

# Fixes
Fixes #7085


StevenCostiou:
There is a big dependency problem, as now `Exception` call methods from `Oups` which introduces a wrong dependency between the two.

StevenCostiou:
Tests passed, the remaining failing test is unrelated.
Now what can be improved is the API to create debug requests, and to make all users of debug request call that API instead of instantiating a request each time.

src/Kernel/Process.class.st
```diff
@@ -198,32 +198,17 @@ Process >> debug [
...
-Process >> debug: context title: title full: bool [
-	
-	^ UIManager default
-		debugProcess: self 
-		context: context 
-		label: title 
-		fullView: bool
-]
-
 { #category : #debugging }
```
Ducasse, line 201: Could this method packaged with Oups? because it continues to promote the dependency to UIManager?

Ducasse:
Nice class comments for example DebugRequest is cool :).
I like the idea that we refer to a request object instead of a singleton. 
I will wait that pablo have the time to look at the code too. 
But it looks good.

src/Kernel/Process.class.st
StevenCostiou: The problem is (at least) two-fold:
- Before **Oups**, everything went through the UIManager, but instead of a simple API the full logic was implemented there and the API was crappy. We need a centralized API somewhere, we need to devise where if not in the UIManager to put this API so that it is accessible to most of the code without introducing other dependencies.
- Now, the problem is that **Oups** depends on UIManager to know if it can open graphical debuggers. When we designed it, we knew we did not like it. However, we decided to keep that problem for the next steps. The goal would be to remove that dependency and be as independent as possible so that the debugger infrastructure could be loaded the soonest as possible in the system (ideally, in the bootstrap?). For now, I've identified that it could be loaded just after Kernel, if not for that UIManager dependency.

So the answer is yes, we need to remove that dependency. I advise we keep that problem for after :)

Ducasse:
Ok thanks for the explanation.

PR #7189, opened 2020-08-26
Cleanup-reflective-pragma

MarcusDenker:
some methods (mostly related to perfom:) had a lefover #reflective: pragma. This seems to be a leftover from some experiment, das there is no code using this.


dionisiydk (approved):

hogoww (approved):
There's other changes (simple cleanup) that might be related to another pr though.

PR #7192, opened 2020-08-27
CompiledMethod-isFaulty

MarcusDenker:
- add a quick check for #isFaulty for CompiledMethod that does not need to reify the AST.
- use it in ReMethodHasSyntaxErrorRule


hogoww (approved):
Tests failure look unrelated to me.

Issue #7165, opened 2020-08-23
IMPORTANT: selector extraction DOES NOT work anymore!


Ducasse:
in P8 

selecting `between: 2 and:` in the following and press Cmd-M  find between:and: 
```
1 between: 2
and: 3
```

This is not the case in P9.

Ducasse:
A systematic analysis

## in P8
in Playground
-selecting all expression works
-selecting between: 2 and: works
-placing caret inside between: or and: works

In calypso method pane replacing fully the method body
-selecting all expression works
-selecting between: 2 and: works
-placing caret inside between: or and: works

In calypso class pane replacing fully the class body
-selecting all expression works
-selecting between: 2 and: works
-placing caret inside between: or and: works

In calypso class comment pane replacing fully the comment
-selecting all expression works
-selecting between: 2 and: works
-placing caret inside between: or and: works

In calypso class comment pane replacing with a free text  comment starting
-selecting all expression works
-selecting between: 2 and: works
-placing caret inside between: or and: works

Ducasse:
## in P9
in Playground
-selecting all expression DOES NOT work
-selecting `between: 2 and:` DOES NOT work
-placing caret inside `between:` or `and:` DOES NOT work

In calypso method pane replacing fully the method body
-selecting all expression DOES NOT work
-selecting `between: 2 and:` DOES NOT work
-placing caret inside `between:` or `and:`  DOES NOT work

In calypso class pane replacing fully the class body
-selecting all expression works
-selecting `between: 2 and:` DOES NOT work
-placing caret inside `between:` or `and:` DOES NOT work different version

In calypso class comment pane replacing fully the comment
-selecting all expression works
-selecting `between: 2 and:` DOES NOT work
-placing caret inside `between:` or `and:` (because parseNode: is not implemented on comment)

In calypso class comment pane replacing with a free text  comment starting
-selecting all expression works
-selecting `between: 2 and:` DOES NOT work
-placing caret inside `between:` or `and:` raise an error (because parseNode: is not implemented on comment)

Ducasse:
I started to extract the logic of selector extraction from Rubric so that it can be reused in commands for example. 
The first attempt is available at: 
https://github.com/Ducasse/mybecher/tree/master/myBecher-Xtractor

[...]

PR #7204, opened 2020-08-28
Quick fix of #7165.

lesueur-philippe:
To be tested.


Ducasse:
Thanks a lot because this bug was breaking a lot of interaction.

PR #7220, opened 2020-08-31
7208-sharePools-is-defined-in-Class-but-called-from-superclass

MarcusDenker:
This is a first step to clean it up

- move includesSharedPoolNamed: from behavior to ClassDescription
- move all the definition methods to Class (they have their own version in Metaclass already)
  These where the methods that send e.g.  sharedPoolsString in ClassDescription even though pools are only defined in Class
- add #sharedPool to ClassDescription for sharedPoolsString
- recategorize sharedPoolsString

We should check what the API for pools should be on Metaclass... I think we should move all the API to Class, but I am not sure.

This PR is a good first step for both directions and it should not introduce any backward compatibility problems.

### DIFF
--- a/src/Kernel/Behavior.class.st
+++ b/src/Kernel/Behavior.class.st
@@ -892,13 +892,6 @@ Behavior >> includesSelector: aSymbol [
 	^ self methodDict includesKey: aSymbol
 ]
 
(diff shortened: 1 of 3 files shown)


Ducasse:
This may have an impact of the fluid class API but it will be easy to fix.

Issue #7059, opened 2020-08-07
obsolete method reference to WorldMorph>>initialize


j-brant:
There are two WorldMorph>>initialize methods in the image. It appears that one was run which hooked up two blocks to be run when tools were (un)registered. After the code was run, the method was recompiled, but the blocks remain hooked up in the registry. You can find the reference path to the uninstalled initialize method using: 
```
ReferenceFinder findPathTo: 
    (CompiledMethod allInstances 
        select: [:each | 
            each methodClass = WorldMorph 
                and: [each selector = #initialize 
                and: [WorldMorph>>#initialize ~~ each]]]) first
```

This returns a path: 
```
1 - Smalltalk
2 - a SystemDictionary(lots of globals)...etc...
3 - an Array(#HEPragma->HEPragma #LGitP...etc...
4 - #SmalltalkImage->SmalltalkImage
5 - SmalltalkImage
6 - a Dictionary(#CompilerClass->OpalCo...etc...
7 - an Array(#CompilerClass->OpalCompil...etc...
8 - #Tools->a PharoCommonTools
9 - a PharoCommonTools
10 - an Announcer
11 - a SubscriptionRegistry
12 - an IdentitySet(a WeakAnnouncementSu...etc...
13 - an Array(nil a WeakAnnouncementSubs...etc...
14 - a WeakAnnouncementSubscription(init...etc...
15 - initialize worldState := WorldSta...etc...
16 - WorldMorph>>initialize
17 - WorldMorph>>#initialize
```

This happens in both 8 & 9 images. The method's source is different in 9, but the effect is the same.

Pharo 9.0.0
Build information: Pharo-9.0.0+build.578.sha.a20ed8a531064fac7e141b13565f92eaa1bb584e (64 Bit)

Pharo 8.0.0
Build information: Pharo-8.0.0+build.1124.sha.0932da82f08175e906b0e2a8052120c823374e9f (64 Bit)

There are likely several other methods with the same issue as there are around 80 compiled methods in my 9 image that are not installed in their classes: 
```
CompiledMethod allInstances 
	select: [ :each | 
		[ each methodClass >> each selector ~~ each ] 
			on: Error
			do: [ :ex | ex return: true ] ]
```

Ducasse:
Thanks John. This may be related to the use of fullBlockclosures!

MarcusDenker:
The WorldMorph>>initialize goes away after resetting the ToolRegistry announcer:

[...]

PR #7221, opened 2020-08-31
TextAction-do-not-share-block

MarcusDenker:
Issue  #7059 shows that TextAction class #intialize can not be GCed after recompiling the methods, as shown by this code:

CompiledMethod allInstances 
	select: [ :each | 
		[ each methodClass >> each selector ~~ each ] 
			on: Error
			do: [ :ex | ex return: true ] ]
		
The reason is that the [] is referenced by all instances, which in turn references the method. 

Solution: use nil, as "nil value" and [] value are both nil.

### DIFF
--- a/src/Text-Core/TextAction.class.st
+++ b/src/Text-Core/TextAction.class.st
@@ -48,7 +48,8 @@ TextAction class >> actOnClickBlock: aBlock [
 
 { #category : #'class initialization' }
 TextAction class >> initialize [
-	ActOnClickBlock := [ ]
+	"nil value and [] value are both nil, but sharing a block would hold on to its outer method"
+	ActOnClickBlock := nil
 ]
 
 { #category : #evaluating }


Ducasse:
Marcus how the code snippet show it?

S.

MarcusDenker:
It should show TextAction class>#initialize as one of the non-installed methods.

Ducasse:
Ah ok but why it is not installed. I do not get how this is related to the non GCed behavior.

MarcusDenker:
The block is referenced in all the sub-instances of TextAction (as the default value). Then the block references the method it is from. Recompiling the system now recompiled the method in TextAction, but the existing instances still point to the block which point to the old method. This is why the old method then can not be GCed.

dionisiydk:
clean blocks will avoid this issue completely

dionisiydk:
Will clean blocks avoid this issue completely?

MarcusDenker:
if the block is a clean block, then this block will not have the issue.

Thus this case: yes, with clean blocks that would work. 

This means that in general, clean blocks will lead to these problems happening less, but it will be a bit more confusing as changing the code (adding a self call, for example), will suddenly introduce it...

PR #7223, opened 2020-09-01
Pragma: cache Pragmas in the Pragma class

MarcusDenker:
This is a first idea how to cache pragmas
-> Pragma class has a Dictionary with one entry per selector 
-> It references the Pragmas weakly

CompiledMethods add their pragmas to the cache when they are added in a class.

Speed to get all pragmas:

[Pragma allInstalled] bench "'799.320 per second'".
[PragmaCollector allSystemPragmas] bench  "'54.923 per second'"

But it will now grow with the number of pragmas, not number of methods.

Pragma allNamed: aSymbol can replace the use of the PragmaCollector in many cases (to be explored)

This PR just adds the Cache and tests, it is not yet used. The next step is to test it a bit, compare it with #allSystemPragmas and
the implement allSystemPragmas to use the cache.

### DIFF
--- a/src/Kernel-Tests/PragmaTest.class.st
+++ b/src/Kernel-Tests/PragmaTest.class.st
@@ -29,6 +29,11 @@ PragmaTest >> setUp [
 
 ]
 
+{ #category : #'tests - cache' }
+PragmaTest >> testAllNamed [
+	self assert: ((Pragma allNamed: #primitive:) first selector) equals: #primitive:
+]
+
 { #category : #tests }
 PragmaTest >> testArgumentAt [
 	| pragma |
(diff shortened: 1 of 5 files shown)


MarcusDenker:
I checked and the bootstrapped image PragmaCollector allSystemPragmas and Pragma allInstalled  are the same:


PragmaCollector allSystemPragmas asSet = Pragma allInstalled asSet  ==> true

MarcusDenker:
failing test not related

MarcusDenker:
Full description: https://blog.marcusdenker.de/speeding-up-pragma-access

<blockquote><div><strong><a href="https://blog.marcusdenker.de/speeding-up-pragma-access">Speeding up Pragma Access - Marcus Denker Blog</a></strong></div><div>Accessing all Pragmas (e.g. to find all methods that have a pragma) is quite slow. PragmaCollector was a first try to improve: you pay the price to</div></blockquote>

PR #7224, opened 2020-09-01
ClassOrganization: store sourcePointer, not RemoteString instances

MarcusDenker:
ClassOrganization stores RemoteString instances. But that is not needed: we can store instead the sourcepointer (as we do with CompiledMethod).

Why is this better? it removes one instance of RemoteString per class comment from the image, saving memory espeically for images with many well documented classes.

### DIFF
--- a/src/Kernel/ClassDescription.class.st
+++ b/src/Kernel/ClassDescription.class.st
@@ -211,25 +211,14 @@ ClassDescription >> classComment: aString [
 ClassDescription >> classComment: aString stamp: aStamp [
 	"Store the comment, aString or Text or RemoteString, associated with the class we are organizing.  Empty string gets stored only if had a non-empty one before."
 
-	| pointer header oldCommentRemoteString oldComment oldStamp preamble |
+	| pointer header oldComment oldStamp preamble |
 	oldComment := self organization classComment.
 	oldStamp := self organization commentStamp.
 
 	aString string = oldComment string ifTrue: [ ^ self ].
 
-	aString isRemoteString ifTrue: [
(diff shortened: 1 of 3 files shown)


Ducasse:
What was exactly the remoteString and why was it ever needed?

src/Kernel/ClassDescription.class.st
```diff
@@ -211,25 +211,14 @@ ClassDescription >> classComment: aString [
...
-			newComment: aString string 
-			oldStamp: oldStamp 
-			newStamp: aStamp.
-		^ self organization classComment: aString].
-
-	oldCommentRemoteString := self organization commentRemoteString.
-	pointer := oldCommentRemoteString 
-			ifNil: [0] 
-			ifNotNil: [oldCommentRemoteString sourcePointer].
+	pointer := self organization commentSourcePointer ifNil: [0].
```
dionisiydk, line 220: Could you check that editing comment in one browser updates the another browser which shows  it at same time?
(It looks suspicious that announcing was removed)

MarcusDenker:
RemoteString is used by the Source/Changes condenser. 

On the level of the image model: no, we do not use it as we can just create on demand.

The fact that class setting on the level of ClassDescription was checking for isRemoteString was really strange...

src/Kernel/ClassDescription.class.st
MarcusDenker: it just removed the case that someone calls comment:stamp on a class with a RemoteString as a parameter... I checked and this does not happen anymore (and was very odd in general)

the announcement is raised at the end of the method.

PR #7225, opened 2020-09-02
CatalogBrowser-use-Pragma-not-PragmaCollector

MarcusDenker:
CatalogBrowser can use the API that we have to get pragmas from one class. This ommits iterating the whole system to find pragmas


Ducasse:
Indeed! Scanning everything when we can just check a couple of methods.

PR #7231, opened 2020-09-02
CopyingLocalVariable-CopiedLocalVariable

MarcusDenker:
- change CopyingLocalVariable to CopiedLocalVariable ("this variable is copied"), in line with #createCopiedVariableFor:
- simplify OCUndeclaredVariableWarning>>#defaultAction a little

### DIFF
--- a/src/OpalCompiler-Core/CopiedLocalVariable.class.st
+++ b/src/OpalCompiler-Core/CopiedLocalVariable.class.st
@@ -2,7 +2,7 @@
 A copying variable is an arg or temp var that is copied into a block that later reads this variable
 "
 Class {
-	#name : #CopyingLocalVariable,
+	#name : #CopiedLocalVariable,
 	#superclass : #LocalVariable,
 	#instVars : [
 		'originalVar'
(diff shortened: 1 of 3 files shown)


src/OpalCompiler-Core/CopiedLocalVariable.class.st
```diff
@@ -2,7 +2,7 @@
 A copying variable is an arg or temp var that is copied into a block that later reads this variable
```
dionisiydk, line 2: The comment still uses an old word copying

src/OpalCompiler-Core/LocalVariable.class.st
```diff
@@ -42,7 +42,7 @@ LocalVariable >> astNodes [
 
 { #category : #converting }
 LocalVariable >> createCopiedVariableFor: aScope [
```
dionisiydk, line 44: Also senders of this method and senders of senders still use copying word.

dionisiydk (commented):
Also continuing rename idea here we can refactor: 
```Smalltalk
doCopying
	| copying |
	copying := scope tempVars select: [ :each | each isEscapingRead].	
	copying do: [ :each | scope addCopyingTemp: each]
```
The analyzer here does the job of scope. It should be simple:
```Smalltalk
doCopying
	scope copyEscapingLocals
```
And I think it may lead to other simplifications in senders.

src/OpalCompiler-Core/CopiedLocalVariable.class.st
MarcusDenker: added to backlog

Issue #7214, opened 2020-08-30
escaping strings and symbols in copy paste


Ducasse:
If I type

``` 
l'idiot
```
and select it and type '
I get 

``` 
'l''idiot'
```
and this is good

Now if I have '' and paste it inside strings are not escaped.

guillep:
This is the same problem as #7219. The paste, based on the AST, is not correctly done because the AST is wrongly built.

Ducasse:
cool :)

PR #7232, opened 2020-09-02
Fix #7219

### LINKED ISSUE #7219: Copy paste raises error
```
asFluidDefinitionString

	^ String streamContents: 
		[:strm |
		strm nextPutAll: self name.
		
		self hasTraitComposition ifTrue: [
			strm
				crtab;
				nextPutAll: 'uses: ';
				print: self traitComposition ].
		self slots ifNotEmpty: [  
			strm
				crtab;
				nextPutAll: 'slots: ';
				nextPutAll: self slotDefinitionString]]
```

when I copy paste `nextPutAll: self slotDefinitionString` I get a dnu

guillep:
The parsing of ASTs for copy past and others is not producing a valid AST, because they are using wrongly the parser's API.

=> Produce a valid AST (for faulty methods)

### DIFF
--- a/src/Rubric/RubSmalltalkEditor.class.st
+++ b/src/Rubric/RubSmalltalkEditor.class.st
@@ -257,10 +257,7 @@ RubSmalltalkEditor >> bestNodeInTextAreaWithoutSelection [
 									stop >= self textArea text size ifFalse: [ end :=  self textArea string copyFrom: stop + 1 to: self textArea text size ].
 									beginning,end ]
 					 ifTrue: [ self textArea string ].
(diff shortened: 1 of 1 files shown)


guillep:
Do not integrate yet, this needs to be handled correctly for faulty expressions too.

guillep:
I updated the PR to fix too #7214. The final solution is that if in scripting mode, it should use a method parser, otherwise an expression parser.

Ducasse:
I wonder why we do not use the mode instance because each mode encapsulates the associated parsing selector. 
I merged it but may be we can discuss this later.

PR #7234, opened 2020-09-02
Fix #7217

### LINKED ISSUE #7217: SHRBTextStyler should be more robust
For example it is used in the context of Microdown and 

```
Trait named: TBehavior
   uses: 
```
raises an error. 

In particular, this method 
```
visitParseErrorNode: anErrorNode
	self 
		addStyle: #invalid
		forNode: anErrorNode
```
with anErrorNode equals to 
 RBParseErrorNode( RBEOFToken)

leds to an error.

guillep:
RBEOFToken should have an empty value. Otherwise it has a wrong stop and source interval.


Ducasse:
Cool to see this fixed. One day I will start to write tests for the styler.

PR #7236, opened 2020-09-02
CleanBlockClosure-Decompiler

MarcusDenker:
This PR adds support to decompile Clean Blocks.

### DIFF
--- a/src/Flashback-Decompiler/FBDDecompiler.class.st
+++ b/src/Flashback-Decompiler/FBDDecompiler.class.st
@@ -486,6 +486,22 @@ FBDDecompiler >> pushActiveContext [
 	simulatedStack addLast: builder codeThisContext 
 ]
 
+{ #category : #'closure support' }
+FBDDecompiler >> pushCleanClosure: aCompiledBlock [
+
+	| savedSimStack savedSequence |
+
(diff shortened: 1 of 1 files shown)


src/Flashback-Decompiler/FBDDecompiler.class.st
```diff
@@ -486,6 +486,22 @@ FBDDecompiler >> pushActiveContext [
...
+	| savedSimStack savedSequence |
+
+	savedSimStack := simulatedStack.
+	savedSequence := currentSequence.
+
+	simulatedStack addLast: (self
+		extractFullBlock: aCompiledBlock numCopied: 0).
+
+	simulatedStack := savedSimStack.
+	currentSequence := savedSequence.
```
dionisiydk, line 501: No tests for this?

MarcusDenker: There were tests failing for clean blocks. This method is a variation of the existing method for normal block (just removed all code related to copied vars). 

Tests are green with the change.

We need in general a good regression tester for the decompiler

MarcusDenker: (that is, decompile and recompile the whole image and then run all tests)

PR #7246, opened 2020-09-05
7240-PRRichTextComposer-does-not-need-to-use-PragmaCollector

### LINKED ISSUE #7240: PRRichTextComposer does not need to use PragmaCollector
PRRichTextComposer class>>#languageStylerFor: 

this can be rewritten to use Pragma allNamed:

Can I just submit to Pharo9 or is this some other repo?

MarcusDenker:
use Pragma>>#allNamed: instead of PragmaCollector in #languageStylerFor:, fixes #7240

### DIFF
--- a/src/Pillar-ExporterRichText/PRRichTextComposer.class.st
+++ b/src/Pillar-ExporterRichText/PRRichTextComposer.class.st
@@ -171,8 +171,7 @@ PRRichTextComposer class >> initialize [
 PRRichTextComposer class >> languageStylerFor: aLanguage [
 	| stylers |
 	stylers := Dictionary
-		newFromAssociations: ((PragmaCollector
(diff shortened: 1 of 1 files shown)


Ducasse:
Yes and this logic should really be revisited I did not look at it but to me it is far too complex for nothing

Ducasse:
I will manage it the microdown repo. 
https://github.com/pillar-markup/MicroDown/issues/92

MarcusDenker:
I close this but keep the issue open until this is in Pharo9. 
(but it shows again the negative side of "external packages": this case would be fixed. Now it is open until.. when?)

MarcusDenker:
Checked with Stef: easiest is to commit it now, will later be merged upstream

Issue #7244, opened 2020-09-04
Athens needs more unit tests


ronsaldo:
**Describe the problem**
Athens has few unit tests. Athens classes that do not require actual rendering must be tested.

**Classes involved**
All of the classes in the Athens-* packages. The worst offenders are in Athens-Core.

**Proposal**
Missing test cases must be written.

**Version information:**
 - Version [e.g. Pharo 9.0]
 - Commit [e.g. 22]

**Expected development cost**
One afternoon for the transform classes.
The rendering methods may require devising a strategy for testing by comparing with reference images. This may be very expensive to test.

PR #7253, opened 2020-09-07
Add unit tests for AthensTransform and family.

ronsaldo:
This is a partial fix for https://github.com/pharo-project/pharo/issues/7244.


Ducasse:
Thanks ronie this is cool.

PR #7255, opened 2020-09-07
merge-spec2-0.7.3

estebanlm:
https://github.com/pharo-spec/Spec/releases/tag/v0.7.3

- fix a leak on dialogs, which were not being removed from the windows list in application
- enhance demo of SpNotebookPresenter
- made string columns not sortable by default
- fix IconStyler usage
- fix `SpCodePresenter>>#overridingContextMenu`
- enhance comments and organisation (WIP)


Ducasse:
Should we integrate it? What is the semantics of WIP? We keep it until you say this is ok?

estebanlm:
cool, tests are green. We can proceed (then I can submit a newtools merge)

PR #7269, opened 2020-09-08
ResetBC-2-AST-Mapping-on-Recompile

MarcusDenker:
With the BC->AST mapping cache, we need to make sure to reset it as soon as we regenerate the bytecode (as it could be regenerated with a modiefied AST or e.g. metalinks added)

This fixes the mapping for example when doing a "Debug test" from the browser menu

### DIFF
--- a/src/OpalCompiler-Core/RBMethodNode.extension.st
+++ b/src/OpalCompiler-Core/RBMethodNode.extension.st
@@ -58,6 +58,8 @@ RBMethodNode >> generate [
 
 { #category : #'*OpalCompiler-Core' }
 RBMethodNode >> generate: trailer [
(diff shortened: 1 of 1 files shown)


pavel-krivanek (approved):

src/OpalCompiler-Core/RBMethodNode.extension.st
```diff
@@ -58,6 +58,8 @@ RBMethodNode >> generate [
 
 { #category : #'*OpalCompiler-Core' }
 RBMethodNode >> generate: trailer [
+	"if the bytecode is regenerated, we need to reset the cache"
+	bcToASTCache := nil.
```
dionisiydk, line 62: Good change. But the test would be very nice to have.

PR #7273, opened 2020-09-08
ClyDeprecatedClassExample should be in Deprecated90

### LINKED ISSUE #7272: ClyDeprecatedClassExample should be in Deprecated90
Class is deprecated and should move to Deprecated90 package

astares:
Fix #7272


dionisiydk:
It's not a class to remove with deprecation.  
It is an example of how deprecated classes look in the browser. 
I guess no references confused you. And the reference is actually missing. I was sure there is a method referencing this class to demonstrate a syntax highlighting. But it's not there

PR #7274, opened 2020-09-09
flattenOn-No-If

MarcusDenker:
flattenOn: checks for class and then does on if... to me it looks cleaner to add to those two classes a version of #flattenOn: that does what needs to be done and simplify the #flattenOn: on Collection.

This is tested by the existing test  #testFlattened


svenvc:
Look good

Issue #7251, opened 2020-09-07
Introduce a #isAssociation test method 


MarcusDenker:
**Describe the problem**

Interestingly, Association has no method to check in the style #isAssociation. Instead, there is isVariableBinding which everyon uses.

This is not good as #isVariableBinding implies that there is more going on then just checking if the Object is an Association.

What to do: introduce #isAssociation and start to use it. #isVariableBinding need to be carefully checked, one user that is save to change is #AdditionalMethodState. 

It is interesting how much easier to understand e.g. AdditionalMethodState>>#pragmas is after not using #isVariableBinding.

PR #7278, opened 2020-09-09
7251-Introduce-a-isAssociation-test-method-

MarcusDenker:
- add #isAssociation check method
- use it in AdditionalMethodState

Now it is much clearer that there is nothing going on regarding Variables....

other users have to be checked and #isVariableBinding can be removed at some point from Association (as all Variables are now subclass of Variable)


MarcusDenker:
fixes https://github.com/pharo-project/pharo/issues/7251

PR #7279, opened 2020-09-09
Variables-better-name-for-isBlockVar

MarcusDenker:
#isBlockVar is a bad name. It implies that it checks if a node is some kind of Variable, but instead it checks if a Variable is defined by a block.

- use #isDefinedByBlock
- deprecated isBlockVar
- add test
- fix the two senders

### DIFF
--- a/src/AST-Core/RBVariableNode.class.st
+++ b/src/AST-Core/RBVariableNode.class.st
@@ -125,8 +125,8 @@ RBVariableNode >> isClassVariable [
 
 { #category : #testing }
 RBVariableNode >> isDefinedByBlock [
-	"true if a variable node is defined by a block node"
-	^variable scope isBlockScope
+	"true if a variable node is defined by a block"
+	^variable isDefinedByBlock
 ]
 
 { #category : #testing }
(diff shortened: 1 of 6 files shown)


src/AST-Core/RBVariableNode.class.st
```diff
@@ -117,18 +117,18 @@ RBVariableNode >> isArgumentVariable [
...
 { #category : #testing }
 RBVariableNode >> isClassVariable [
 	^variable isClassVariable
 
 ]
 
+{ #category : #testing }
+RBVariableNode >> isDefinedByBlock [
+	"true if a variable node is defined by a block node"
+	^variable scope isBlockScope
```
dionisiydk: we should delegate to variable here:
```Smalltalk
^variable isDefinedByBlock
```

dionisiydk (approved):

src/AST-Core/RBVariableNode.class.st
```diff
@@ -117,18 +117,18 @@ RBVariableNode >> isArgumentVariable [
...
-
 { #category : #testing }
 RBVariableNode >> isClassVariable [
 	^variable isClassVariable
 
 ]
 
+{ #category : #testing }
+RBVariableNode >> isDefinedByBlock [
+	"true if a variable node is defined by a block"
```
dionisiydk, line 128: Some comments look really useless when it's just a split of words in the class name and methods by space

dionisiydk: Actually it starts to be useless with rename

PR #7280, opened 2020-09-09
Variables-variable-not-binding

MarcusDenker:
Historically, sending #binding to a Class gives you the association in the systemDictionary.

We have now changed this so that those objects are the Variables representing the concept of accessing the global variable.

That is, 

OrderedCollection variable

now returns the *Variable* that references OrderedCollection.

To be in sync everywhere, we should use the selector #variable to get Variables. We did that already on the AST node level, this PR does the same change on the level of Behaviors.

-> the implemetation is now #variable
-> binding is retained as a compatibility method.

A later PR then can change all users of binding to use #variable instead

### DIFF
--- a/src/Kernel/Class.class.st
+++ b/src/Kernel/Class.class.st
@@ -238,14 +238,6 @@ Class >> basicDeclareClassVariable: aClassVariable [
 ]
 
 { #category : #compiling }
-Class >> binding [
-   "Answer a binding for the receiver, sharing if possible"
-   | binding |
-	binding := self environment associationAt: self name ifAbsent: [LiteralVariable key: nil value: self].
-   ^binding value == self ifTrue: [binding] ifFalse: [LiteralVariable key: nil value: self]
-]
-
-{ #category : #compiling }
 Class >> bindingOf: varName [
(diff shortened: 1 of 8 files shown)


src/Kernel/Class.class.st
```diff
@@ -1338,6 +1330,14 @@ Class >> usesPoolVarNamed: aString [
 	^self allSharedPools anySatisfy: [:each | each usesClassVarNamed: aString]
 ]
 
+{ #category : #compiling }
+Class >> variable [
```
dionisiydk, line 1334: I think #variable is too simplistic name for this purpose. It's quite confusing due to existing instanceVariables, classVariables and other variants of actual variables defined by class.
Previous #binding was also not clear name but at least it was not related to actual variables of the class.

We can have something like:
```Smalltalk
aClass environmentVariable.
```
With the meaning of a variable containing the class in its environment.

Or maybe better:
```Smalltalk
aClass referenceVariable
```
With the meaning of a variable to reference the class. I think I would prefer this variant

Ducasse:
Marcus I agree with denis on this point. environmentVariable is a good suggestion to me.

MarcusDenker:
Ok, I will revert it for now.

PR #7282, opened 2020-09-09
More-PragmaCollector-Improvements

MarcusDenker:
PragmaCollector has just one way to create itself: by filtering all Pragmas. 

This is not a good idea, as this means that it always has to iterate over all Pragmas. For large systems, this is slow however you do it
(iterate all methods or iterate all pragmas of the cache and check if the methods are still installed)

For one, PragmaCollector is often not needed. But existing clients that rely on announcements of added / removed pragmas can benefit from a simple change: When creating PragmaCollector, we can
pre-set which Pragma selectors we are interested in.

This PR implements creating PragmaCollector with #selectors:  and changes both PragmaMenuBuilder and KMKeymapBuilder to use this new scheme.

### DIFF
--- a/src/Deprecated90/PragmaCollector.extension.st
+++ b/src/Deprecated90/PragmaCollector.extension.st
@@ -1,0 +1,7 @@
+Extension { #name : #PragmaCollector }
+
+{ #category : #'*Deprecated90' }
+PragmaCollector class >> allSystemPragmas [
+	self deprecated: 'use Pragma allInstalled instead'.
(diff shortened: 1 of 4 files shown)


dionisiydk (approved):

Ducasse:
Marcus could you improve the class comments because you mentioned in the method comment that we should use selector: but this is really unclear why and what selector is used for.

MarcusDenker:
I improved the comments

PR #7287, opened 2020-09-10
Remove the implicit coercion of float indices into integers in #at: and family

ronsaldo:
We found this that looks like a really bad smell. This kind of type coercions if needed should always be done by the user.


Ducasse:
Indeed this is hiding information...

Ducasse:
Pablo can you have a look?

MarcusDenker (approved):
looks good

PR #7292, opened 2020-09-11
BlockMenuBuilder-not-needed

MarcusDenker:
BlockMenuBuilder was just there because PragmaMenuBuilder was too slow.

Now we can remove it. Not deprecation, as: added on Pharo9, has comment "Please, do not use me".


Ducasse:
One less!

PR #7293, opened 2020-09-11
RBVariable-do-not-use-Subclasses

MarcusDenker:
- Forward all checks #isSelf and friends to the Variable, never check the name
- Visitor forwards to the Variable, too.

TODO:
- #selfNode and superNode need to not use RBVariableNode subclasses

### DIFF
--- a/src/OpalCompiler-Core/TemporaryVariable.class.st
+++ b/src/OpalCompiler-Core/TemporaryVariable.class.st
@@ -15,7 +15,7 @@ TemporaryVariable class >> semanticNodeClass [
 
 { #category : #visiting }
 TemporaryVariable >> acceptVisitor: aProgramNodeVisitor node: aNode [
-	^ aProgramNodeVisitor visitTemporaryNode: self
+	^ aProgramNodeVisitor visitTemporaryNode: aNode
 ]
 
 { #category : #queries }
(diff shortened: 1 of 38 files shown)


src/OpalCompiler-Core/TemporaryVariable.class.st
```diff
@@ -13,6 +13,11 @@ TemporaryVariable class >> semanticNodeClass [
 	^RBTemporaryNode 
 ]
 
+{ #category : #visiting }
+TemporaryVariable >> acceptVisitor: aProgramNodeVisitor node: aNode [
+	^ aProgramNodeVisitor visitTemporaryNode: aNode
```
dionisiydk, line 18: We should also cleanup methods with temp in name to have localVariable meaning. Here it would be #visitLocalVariableNode:.

dionisiydk (approved):
Finally no become :)

dionisiydk (changes requested):
Many related tests failed

src/OpalCompiler-Core/TemporaryVariable.class.st
MarcusDenker: There is some real problem in the AST... #visitArgumentNode: is called for *declarations*, while visitTemporaryNode: is how you would expect it. There is #visitTemporaryDeclarationNode: for the temp declaration... which falls back to call #visitArgumentNode, too.

This requires quire some non-backward compatible refactoring to get right....

dionisiydk (approved):

Ducasse:
Yes removing a misplace become is always a little joy!

PR #7294, opened 2020-09-12
7191-aCollection-max-aBlock-could-be-defined-on-Set-too-

### LINKED ISSUE #7191: aCollection max: aBlock could be defined on Set too.
Right now `aCollection max: aBlock` identifies the collection element for a block computation, but it does not work for Set.

sic-f:
Implement Set >>max:

Fixes  #7191


sic-f:
Not sure what specifically is the problem in the failing build

MarcusDenker:
failing test is not related

PR #7296, opened 2020-09-13
PragmaMenuBuilder-no-PragmaCollector

MarcusDenker:
PragmaMenuBuilder was only caching a PragmaCollector because of the slowness of accessing all Pragmas.

After merging https://github.com/pharo-project/pharo/pull/7291 ,  there will be no long living instances of PragmaMenuBuilder.
Thus we can remove the collector and directly get the Pragmas.


Ducasse:
Oh this is cool!!!!

PR #7297, opened 2020-09-14
KMPragmaKeymapBuilder-remove-PragmaCollector

MarcusDenker:
- KMPragmaKeymapBuilder manages adding/removing methods with pragmas itself. We do therefore not need PragmaCollector but can just use the Pragma API directly.
- remove unused ivar
- class side: cleanup code a little


MarcusDenker:
I will merge this, it is now open for 14 days...

Issue #7306, opened 2020-09-15
Menubar is messed up in latest Pharo9 build


MarcusDenker:
**Describe the bug**
The menu bar has entries related to debugging

MenubarMorph reset.

fixes it.

Which is odd, because this is called as part of the cleanup scripts which should be executed as the last thing in the build

PR #7308, opened 2020-09-15
7306-Menubar-is-messed-up-in-latest-Pharo9-build

MarcusDenker:
checking if the reset is the cause for issue #7306


MarcusDenker:
I checked the build artefact: Menu looks good again.

PR #7320, opened 2020-09-16
OpalCompiler-parse-should-doSemanticAnalysis

MarcusDenker:
if we call #doSemanticAnalysis in #parse of OpalCompiler, we can remove the calls from the users.

### DIFF
--- a/src/OpalCompiler-Core/OpalCompiler.class.st
+++ b/src/OpalCompiler-Core/OpalCompiler.class.st
@@ -324,7 +324,6 @@ OpalCompiler >> compilationContextClass: aClass [
 OpalCompiler >> compile [
 	| cm |
 	[ 	[	self parse.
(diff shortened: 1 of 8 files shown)


src/OpalCompiler-Core/OpalCompiler.class.st
```diff
@@ -325,7 +325,6 @@ OpalCompiler >> compile [
 	| cm |
 	[ 	[	self parse.
 			self transformDoit.
```
dionisiydk: Now #doSemanticAnalysis will be performed two times: in #parse and in #transformDoit. Or I may miss something

MarcusDenker: I have now change the implementation to have a different explicit code path when parsing/compiling doIts, this solves that problem

PR #7321, opened 2020-09-16
Fix for #hasMethodReturn in FullBlockClosure

dionisiydk:
It fixes #hasMethodReturn in FullBlockClosure when the return is performed in one of embedded blocks.
The implementation introduces withAllBlocksDo: and allBlocksDo: methods to have similar API to subclasses query (withAllSubclassesDo:)


dionisiydk:
Tests pass on linux. So I think other failures are unrelated

PR #7322, opened 2020-09-16
Clean nested block api

dionisiydk:
Optimization to use withAllBlocksDo: instead of withAllBlocks to avoid collecting all blocks in an intermediate set.
It's done on top of #7321 to separate a bug fix and cleanup (which can be discussible)

### DIFF
--- a/src/Kernel/CompiledCode.class.st
+++ b/src/Kernel/CompiledCode.class.st
@@ -931,6 +931,14 @@ CompiledCode >> withAllBlocks [
 ]
 
 { #category : #accessing }
+CompiledCode >> withAllBlocksDo: aBlock [ 
+	
+	aBlock value: self.
+	self allBlocksDo: aBlock
+	
+]
+
+{ #category : #accessing }
 CompiledCode >> withAllNestedLiteralsDo: aBlockClosure [ 
 	"This method traverses all the nested literals.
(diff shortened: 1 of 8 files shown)


src/Kernel/CompiledCode.class.st
```diff
@@ -928,8 +932,8 @@ CompiledCode >> withAllNestedLiteralsDo: aBlockClosure [
 	As a Block or Method can have literals in the nested blocks. 
 	This is used to query all the selectors used in a method for example"
 	
-	self withAllBlocks do: [ :aCompiledCode | 
-		aCompiledCode allLiterals do: aBlockClosure ]
+	self withAllBlocksDo: [ :aCompiledCode | 
+		aCompiledCode literals do: aBlockClosure ]
```
dionisiydk, line 936: We definitely do not need a class of a method to traverse here. But #literals do not look into pragmas (the last literal in general). 
I checked that users of this method are not affected by this change (they do not care about pragmas or whatever last literal can return)

dionisiydk:
crashes on some platforms are unrelated

src/Kernel/CompiledCode.class.st
MarcusDenker: yes, correct!

MarcusDenker (approved):

PR #7323, opened 2020-09-16
CompiledBlock-reallyNoTrailer

MarcusDenker:
We are creating CompiledBlock instances like methods: with an empty trailer. But that wastes 4 bytes per CompiledBlock, as the trailer is never set (only Methods store the source pointer there).

This PR creates CompiledBlock without trailing bytes

### DIFF
--- a/src/Kernel/CompiledBlock.class.st
+++ b/src/Kernel/CompiledBlock.class.st
@@ -15,6 +15,12 @@ Class {
 	#category : #'Kernel-Methods'
 }
 
+{ #category : #accessing }
+CompiledBlock >> endPC [
+	"Answer the index of the last bytecode."
+	^ self size
+]
+
 { #category : #scanning }
 CompiledBlock >> hasMethodReturn [
 	"Answer whether the receiver has a method-return ('^') in its code."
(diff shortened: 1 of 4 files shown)


Ducasse:
marcus so endPC of a compiled method is not really the last place where the PC can point but the space used by the compiled method since endPC point at the end of the method trailer.
Is my analysis correct?
It would be good to describe it in the class comment.

MarcusDenker:
Yes! Now imagine a real Solution: CompiledMethod should be a Variable class storing the literals. It should have some instance vars, too: source (for the trailer or a String), byte codes (a ByteArray), flags (an integer to store e.g. which compiler options where enabled, this would store properties if needed), selector, class. Add a subclass for method with pragmas.

We then can share the ByteArray over all methods and blocks where the bytecode is the same, paying back the memory cost.
(of all the methods in the system, there are just 35% of byte arrays that are different).

MarcusDenker:
Ah, and this would then remove a lot of code... CompiledMethodTrailer would go away, as we can store the sourcePointer as does ClassOrganizer now, all the complexity of "is the 2nd last literal a selector or a property array?", all the calculations to access literals...

MarcusDenker:
And on the VM level, the whole concept of a strange object that is variable for both pointers and non-pointers would go away. This should simplify the GC, or not?

Ducasse:
I do not know but this is not something that we should try to do incrementally. 
For this we should have a full version to check if it is really worth. 
My point was just that we document what we have because endPC is misleading for me. 
it could be fullSize.

PR #7324, opened 2020-09-17
Deprecate-RBVariable-Subclasses

MarcusDenker:
The PR moves the RBVariable Subclasses to the Depcated package.

all class comments of the classes got a short explanation of what we did (why they are depcated now)


MarcusDenker:
Very strange... the move to deprecated got lost.  I will redo

PR #7328, opened 2020-09-17
Added a closeAllDebugger method to trait TDebugger class, that users …

### LINKED ISSUE #7313: Close all Debuggers menu option is not working.
If you have many open debuggers, and you do right button on the taskbar and then in close all debuggers it generates DNUs.

dupriezt:
…have to implement.

This method is called when using the "close all debuggers" button in the taskbar's rightclick menu.
Added 'do nothing' implementations of this method to EDEmergencyDebugger and TDebuggerTest so that they comply with this new requirement. (These implementations do nothing on purpose, see method comments)

Fixes #7313

### DIFF
--- a/src/Debugger-Model/TDebugger.trait.st
+++ b/src/Debugger-Model/TDebugger.trait.st
@@ -34,6 +34,11 @@ TDebugger classSide >> availableAutomatically: anObject [
 	availableAutomatically := anObject
 ]
 
+{ #category : #accessing }
+TDebugger classSide >> closeAllDebuggers [
+	^ self explicitRequirement
+]
+
 { #category : #'instance creation' }
 TDebugger classSide >> debugSession: aDebugSession [
 	self explicitRequirement
--- a/src/Debugger-Model-Tests/TDebuggerTest.class.st
+++ b/src/Debugger-Model-Tests/TDebuggerTest.class.st
@@ -6,6 +6,13 @@ Class {
 	#category : #'Debugger-Model-Tests-Core'
 }
 
+{ #category : #accessing }
+TDebuggerTest class >> closeAllDebuggers [
+	"Do nothing. This class is a test class, not a real debugger, so it cannot stay open like real debugger. Therefore, it has nothing to do to 'close all of its debuggers'"
+
+	^ self
+]
+
 { #category : #'instance creation' }
 TDebuggerTest class >> debugSession: aDebugSession [
 	^true
--- a/src/EmergencyDebugger/EDEmergencyDebugger.class.st
+++ b/src/EmergencyDebugger/EDEmergencyDebugger.class.st
@@ -75,6 +75,13 @@ EDEmergencyDebugger class >> availableAutomatically [
 	^ true
 ]
 
+{ #category : #'utilities api' }
+EDEmergencyDebugger class >> closeAllDebuggers [
+	"Do nothing, because this message does not make sense for the Emergency Debugger. If the emergencyDebugger  is open, it has the focus so the user cannot right-click on the taskbar and click on the 'close all debuggers' button anyway"
+
+	^ self
+]
+
 { #category : #'instance creation' }
 EDEmergencyDebugger class >> debug: aDebugSession [
 	self flag: 'Not good, must add and test the error too!'.


dupriezt:
I reflected the change this PR makes to EDEmergencyDebugger in the corresponding class in the NewTools repo https://github.com/pharo-spec/NewTools

<blockquote><div> GitHub</div><div><strong><a href="https://github.com/pharo-spec/NewTools">pharo-spec/NewTools</a></strong></div><div>Contribute to pharo-spec/NewTools development by creating an account on GitHub.</div></blockquote>

StevenCostiou:
> I reflected the change this PR makes to EDEmergencyDebugger in the corresponding class in the NewTools repo https://github.com/pharo-spec/NewTools 

Sorry for this, ED is now in Pharo and the version in NewTools is deprecated. I should remove it from NewTools so that the reference version is the one in Pharo and there is no extra work requried.

<blockquote><div> GitHub</div><div><strong><a href="https://github.com/pharo-spec/NewTools">pharo-spec/NewTools</a></strong></div><div>Contribute to pharo-spec/NewTools development by creating an account on GitHub.</div></blockquote>

src/Debugger-Model/TDebugger.trait.st
```diff
@@ -34,6 +34,11 @@ TDebugger classSide >> availableAutomatically: anObject [
 	availableAutomatically := anObject
 ]
 
+{ #category : #accessing }
+TDebugger classSide >> closeAllDebuggers [
+	^ self explicitRequirement
+]
```
StevenCostiou, line 40: I agree with the changes.

However, why not having the default behavior (do nothing) in TDebugger instead of an explicit requirement, then debugger classes redefine it to implement their own logic?

I am asking genuinely, I do not know about what would be the best solution.
I would say, generally, provide a default behavior and avoid redefining the same default behavior everywhere, but I understand that the idea is to force all debuggers to reimplement that method. 

In the other hand, there are nothing constraining debuggers to close all of their instances, it is just a practicality offered to developers.

dupriezt: The idea is indeed to force all debuggers to reimplement that method.
Having it as an explicit requirement means that an explicit error will pop up the next time someone (like Pablo here) clicks "close all debuggers" and it doesn't work. So the user will be told what the error is (missing implementation of #closeAllDebuggers) instead of the "close all debuggers" button silently not working.

StevenCostiou: I know what it does, but what I mean is that it looks like a code smell.

Ensure that all debuggers to which we ask "close all debuggers" can answer => **yes**
Duplicating the same empty method in all debuggers that do not need it redefined => **smell**
Having a default empty method in the Trait like I suggested => **smell**

How can we do it better?
Maybe "close all debuggers" is the wrong naming, and if we find the naming corresponding to the exact behavior we are looking for, it will be easier to find a solution.

Maybe this method do not belong to the Trait, and should be in the debugger system itself?
It seems odd that, if you ask "close all dbg" to GT, it closes only the GT debuggers **windows**... maybe that should be a "system-level" method that closes/terminates all windows/presenters/models/instances of all TDebugger users?

StevenCostiou:
```
Maybe this method do not belong to the Trait, and should be in the debugger system itself?
It seems odd that, if you ask "close all dbg" to GT, it closes only the GT debuggers windows... maybe that should be a "system-level" method that closes/terminates all windows/presenters/models/instances of all TDebugger users?
```

Ok, my bad, the "close all debuggers" method is already a system-level method.
So from there, we need a way to avoid the smells above.

Ducasse: Hello guys what is the next step?
Should I integrate this PR?
Tx

dupriezt: > Should I integrate this PR?

Steven says it's a smell, so I guess not.

StevenCostiou: It's an open discussion. We could say that we integrate it as is so that it fixes the problem, and take care of the problem in a second time.

MarcusDenker: Yes, doing things step by step is good

MarcusDenker (approved):

PR #7335, opened 2020-09-18
All exceptions now respond to smartDescription.

### LINKED ISSUE #7334: Exception does not implement smartDescription while some of its subclasses do
Sometimes, `smartDescription` is used to display information about exceptions, but that method is only implemented in `MessageNotUnderstood`. This breaks polymorphism.

Just add generic `smartDescription` method returning the default description in`Exception`.

StevenCostiou:
Fixes #7334

### DIFF
--- a/src/Kernel/Exception.class.st
+++ b/src/Kernel/Exception.class.st
@@ -372,6 +372,11 @@ Exception >> signalerContext [
 ]
 
 { #category : #accessing }
+Exception >> smartDescription [
+	^self description
+]
+
+{ #category : #accessing }
 Exception >> tag [
 	"Return an exception's tag value."
 


dionisiydk:
#smartDescription is really bad name. Considering that the only user is  Process>>debug I would rename it to #titleForDebugger. That would have really clear purpose.

From the other side I do not understand why #description is not enough and why we can't just use this existing method (with better implementation where it's needed).

dionisiydk (approved):
Good as it's a fix but please consider my proposal

StevenCostiou:
> From the other side I do not understand why #description is not enough and why we can't just use this existing method (with better implementation where it's needed).

I agree with that. Maybe also provide a `defaultDescription` that just returns a `printString` and is called by default by `description`.

StevenCostiou:
I opened an issue to improve the API https://github.com/pharo-project/pharo/issues/7347
Could we merge this one?
It's slowing me down in the debugger development.

PR #7341, opened 2020-09-19
LiteralSharing-Optional

MarcusDenker:
Lots of methods and blocks refer to literals that are equal. 

This is not yet turned on by default as literals can be changed... which leads to bugs that are now local, but with this change get to have a potential global effect. (making exactly those literals read only that are shared here is a future step).

ImageCleaner new statistics "141930" 

ImageCleaner shareLiterals

ImageCleaner new statistics "118979"

### DIFF
--- a/src/Collections-Native/ByteArray.class.st
+++ b/src/Collections-Native/ByteArray.class.st
@@ -212,6 +212,12 @@ ByteArray >> isLiteral [
 	^ self class == ByteArray
 ]
 
+{ #category : #testing }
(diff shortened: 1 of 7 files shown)


MarcusDenker:
As this PR is not actually activating read-only literals nor calling the sharing code, this PR is safe to integrate

MarcusDenker:
(but I can not integrate it without having at least one review)

dionisiydk (approved):

Issue #7344, opened 2020-09-21
completion scope failure with unfinished brackets and temp


pavel-krivanek:
**Describe the bug**
When you write a method that contains a temporary variable and use unfinished brackets during typing (it is more easily discoverable with _smart characters_ setting turned off), you get a nil scope of the code completion

**To Reproduce**
To any class, write this method code: 
```smalltalk
test 

	| temp |
	[ [  ]
```

It will produce an error: `MessageNotUnderstood: receiver of "lookupVar:" is nil`

MarcusDenker:
This seems to be fixed by this PR: https://github.com/pharo-project/pharo/pull/7348

MarcusDenker:
merged

PR #7348, opened 2020-09-21
Fix-Completion-isSelfVariable

MarcusDenker:
This fixes a bug where #isSelfVariable is send to a RBVariableNode with a nil variable. Reason: Semantic analysis was broken and *that* was hidden behing an on: Error do: handler...

- removes the error handler
- implements parseNode of the CoASTResultSetBuilder to use the compiler facade (remove a ref to RBParser and simplify)
- remove the "self visitNode: anEnglobingError value.â


pavel-krivanek (approved):
fixes #7344

PR #7354, opened 2020-09-22
Fix race conditions and bugs in OSWindow

ronsaldo:
This PR fixes race conditions in OSWindow that are typically triggered when closing an OSWindow. Additional related bugs were  found in the process:
- The per world cycle delay is refactored into a single delay per frame for all windows in the Morphic render loop. And the default default is changed from 50 fps (which ends being running in 30 fps) which makes the UI much more responsive.
- The keyDown event generated by the OSWindowMorphicEventHandler is sent into an event queue that nobody is consuming. This PR removes the event queue, and delivers this event to Morphic.

### DIFF
--- a/src/Morphic-Core/WorldState.class.st
+++ b/src/Morphic-Core/WorldState.class.st
@@ -115,6 +115,32 @@ WorldState class >> discoveredMenuTitle [
 	^ DesktopMenuTitle ifNil: [DesktopMenuTitle := 'World']
 ]
 
+{ #category : #'drawing cycle' }
+WorldState class >> doDrawCycleWith: aBlock [
+	aBlock ensure: [
+		self doInterCycleWait
+	]
+]
+
+{ #category : #'drawing cycle' }
+WorldState class >> doInterCycleWait [
+	| waitTime newCycleTime |
+	newCycleTime := Time millisecondClockValue.
+	waitTime := self serverMode
+		ifTrue: [ 50 ]
+		ifFalse: [
+			| expectedEndTime lapse |
+			lapse := self interCycleLapse.
+			expectedEndTime := self lastDrawCycleTime + self interCycleLapse.
+			"This minimum is required to prevent freezing the system during startup.
+			The times two in the lapse is to allow some slack due to precision issues."
+			expectedEndTime - newCycleTime min: (lapse*2).
+		].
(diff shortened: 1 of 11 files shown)


src/Morphic-Core/WorldState.class.st
```diff
@@ -644,31 +671,6 @@ WorldState >> initialize [
 
 ]
 
-{ #category : #'update cycle' }
-WorldState >> interCyclePause: milliSecs [
-	"delay enough that the previous cycle plus the amount of delay will equal milliSecs.  If the cycle is already expensive, then no delay occurs.  However, if the system is idly waiting for interaction from the user, the method will delay for a proportionally long time and cause the overall CPU usage of Pharo to be low.
-	If self serverMode returns true then, always do a complete delay of 50ms, independant of my argument. This prevents the freezing problem described in Mantis #6581"
-
-	| wait wait2 |
```
estebanlm, line 652: wait2 is a horrible name :)
Can't be changed with something more descriptive ?

estebanlm (changes requested):
just change that variable name :)

src/Morphic-Core/WorldState.class.st
ronsaldo: That method was completely changed and moved into the class side. Can you review the new version of that method please?:

```smalltalk
{ #category : #'drawing cycle' }
WorldState class >> doInterCycleWait [
	| waitTime newCycleTime |
	newCycleTime := Time millisecondClockValue.
	waitTime := self serverMode
		ifTrue: [ 50 ]
		ifFalse: [
			| expectedEndTime lapse |
			lapse := self interCycleLapse.
			expectedEndTime := self lastDrawCycleTime + self interCycleLapse.
			"This minimum is required to prevent freezing the system during startup.
			The times two in the lapse is to allow some slack due to precision issues."
			expectedEndTime - newCycleTime min: (lapse*2).
		].
	waitTime > 0 ifTrue: [ (Delay forMilliseconds: waitTime) wait ].
	LastCycleTime := newCycleTime.
]
```

estebanlm: duh! that happens when you review too early in the morning ;)

estebanlm (approved):

PR #7355, opened 2020-09-23
Fix some more test related to read-Only literals

MarcusDenker:
enable #optionReadOnlyLiterals by default.  This compiles all literals as read only that return true to isSharedLiteral

### DIFF
--- a/src/Collections-Tests/ArrayTest.class.st
+++ b/src/Collections-Tests/ArrayTest.class.st
@@ -753,9 +753,9 @@ ArrayTest >> testPrinting [
 ArrayTest >> testShuffleBy [
 
 	| base random shuffleByResult |
-	base := #(1 2 3 4 5).
+	base := #(1 2 3 4 5) copy.
 	random := (Random seed: 42).
-	shuffleByResult := #(2 5 4 3 1).
+	shuffleByResult := #(2 5 4 3 1) copy.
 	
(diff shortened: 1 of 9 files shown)


MarcusDenker:
Nice! It's green! But I have to now make make 3 parts out of this: fix Spec, a PR for all the rest, and a PR to finally enable

MarcusDenker:
The PR now just contains some fixes that are save to merge, it does not yet enable it. Spec changes have been moved to its own issue: https://github.com/pharo-project/pharo/issues/7359

src/Collections-Tests/ArrayTest.class.st
```diff
@@ -753,9 +753,9 @@ ArrayTest >> testPrinting [
 ArrayTest >> testShuffleBy [
 
 	| base random shuffleByResult |
-	base := #(1 2 3 4 5).
+	base := #(1 2 3 4 5) copy.
```
Ducasse, line 756: why do we need the two copy?
If shuffled by change base I understand the copy but I do not get the need for the result.

MarcusDenker: if we hand over a literal array, it will be modified. that means, after running the test, the literal is not in sync with the source. (bugs like that are impossible to find...)

If we hand over the copy, only the copy will be modified.

Alternative are: use Array with:with:... or {}.

PR #7358, opened 2020-09-24
Compiler-Remove-ClassScope

MarcusDenker:
The compiler compiles always for an environment (normally the global one). 

When you compile for another environment, there is one big question: Do we ignore the environment set by the classes inside of the environment?

The current implementation does that: the Scope hierarchy is a parallel hierarchy. That is we have methodScope-instanceScope-classScope->environment. All lookup is carefully implemented to use that, not the environment of the class.

Problems:
- This is a lot of machinery to support something strange.
- #hasBindingThatBeginsWith: is out of sync with #bindingsOf: (as it does not know the outer environment that the compiler used)
- The environment that we set in the compiler is not at all reflected in the static entity used.  That is, there is no way to implement #scope for instance variables and get rid of InstanceScope, too.
- #scope for Class Variables is already the class, not the (not persisted) compiler internal scope. There is no way to lookup vars from that scope and come to the same result as the compiler did, as we just do not know the scope set in the compiler.

If we add two contraints:

1. Envionment of classes in some instance of SystemDictionary has to to be that instance
2. Lookup starts for the class that we compile for even for globals

Then we can just remove the concept that the compiler has an additonal scope. The scope of a class it the class, it's outerScope is the environment of the class.

This then enables us to add #scope to any Variable, and looking up the name of the variable in that scope will return the variable.

We have right now no use cases that use the current complex scheme... it was used for tests, but that can be replaced by #bindings:. 

But I wonder if there are use cases.


Ducasse:
Broken tests are unrelated.

src/OpalCompiler-Core/OCInstanceScope.class.st
```diff
@@ -10,6 +10,11 @@ Class {
 	#category : #'OpalCompiler-Core-Semantics'
 }
 
+{ #category : #'instance creation' }
+OCInstanceScope class >> for: aClass [
+	^ self new slots: aClass allSlots; outerScope: aClass; yourself
```
dionisiydk, line 15: I you going to get rid of OCInstanceScope as well?

MarcusDenker: Yes! That is the next step: we do not need to add all slots to a scope, we can just look the up in #lookupVar: of the class.

Then the class is the scope of a Slot (this is, we can implement #scope there, too.. nice for uniformity: every Variable has a scope, lookupVar: withe the name of the var will return that variable).

MarcusDenker: this then shows that self, super and thisContext are variables of the environment scope... there is one instance of SelfVariable per SystemDictionary instance.

dionisiydk: Indeed, It looks very nice

PR #7360, opened 2020-09-24
More-CleanBlock-Fixes

MarcusDenker:
More changes to make an image compiled with clean blocks more usable.

- force some clean blocks in tests
- some other test fixes
- fix #printOn:

KernelClassFilter is not used, we should remove the whole package

With this, the  new debugger can debug clean blocks. What is not yet working is the new emergency debugger


dionisiydk (approved):

Ducasse:
This test is breaking

osx-64 / Tests-osx-64 / testTallyInstructions – MacOSX64.Kernel.Tests.Methods.BlockClosureTest

MarcusDenker:
the failing test should be fixed with the second commit. I do not see it failing on the CI anymore

Ducasse:
Ok then I push!

PR #7368, opened 2020-09-25
7338-Second

Ducasse:
Fix from martin


MarcusDenker:
failing test not related: test2BarPlots, it is failing for another CI run, too

Issue #2949, opened 2019-03-21
Context>>copyStack crashes the VM


VincentBlondeau:
Run on latest build of Pharo8.

https://ci.inria.fr/pharo-ci-jenkins2/blue/organizations/jenkins/Test%20pending%20pull%20request%20and%20branch%20Pipeline/detail/PR-2940/3/pipeline/55

Stack trace:
```
Running tests in 582 Packages


Segmentation fault Thu Mar 21 10:42:23 2019



VM: 201901051900 https://github.com/OpenSmalltalk/opensmalltalk-vm.git

Date: Sat Jan 5 20:00:11 2019 CommitHash: 7a3c6b6

Plugins: 201901051900 https://github.com/OpenSmalltalk/opensmalltalk-vm.git


C stack backtrace & registers:

	eax 0x062091e0 ebx 0x000000cc ecx 0xbffe466c edx 0x00000000

	edi 0x06932240 esi 0x06932240 ebp 0xbffe0908 esp 0xbffe0900

	eip 0x05ca55c3

0   ???                                 0x05ca55c3 0x0 + 97146307

1   Pharo                               0x00095cf3 reportStackState + 770

2   Pharo                               0x000960b1 sigsegv + 213

3   libsystem_platform.dylib            0x9b54ddeb _sigtramp + 43

4   ???                                 0xffffffff 0x0 + 4294967295

5   Pharo                               0x00018189 interpret + 757

6   Pharo                               0x0009718c -[sqSqueakMainApplication runSqueak] + 439

7   Foundation                          0x9487af70 __NSFirePerformWithOrder + 419

8   CoreFoundation                      0x9b5d52be __CFRUNLOOP_IS_CALLING_OUT_TO_AN_OBSERVER_CALLBACK_FUNCTION__ + 30

9   CoreFoundation                      0x9b5d520f __CFRunLoopDoObservers + 399

10  CoreFoundation                      0x9b5c602e __CFRunLoopRun + 974

11  CoreFoundation                      0x9b5c59ea CFRunLoopRunSpecific + 394

12  CoreFoundation                      0x9b5c584b CFRunLoopRunInMode + 123

13  HIToolbox                           0x9720eb5d RunCurrentEventLoopInMode + 259

14  HIToolbox                           0x9720e777 ReceiveNextEventCommon + 163

15  HIToolbox                           0x9720e6bd _BlockUntilNextEventMatchingListInModeWithFilter + 92

[...]

PR #7372, opened 2020-09-25
copyTo-workaround-against-crashes

MarcusDenker:
add the woraround as described in #2949


akgrant43 (approved):
CI test failures are pre-existing, Gt has been using the same workaround successfully for a while now.

dionisiydk (approved):

PR #7373, opened 2020-09-25
7257-Copy-pasting-pragma-body-with-argument-inside-another-pragma-raises-error

### LINKED ISSUE #7257: Copy pasting pragma body with argument inside another pragma raises error
I will try to reproduce it.

### LINKED ISSUE #7389: Error when tring to copy text from code pane: MNU receiver of #getCommentsFor: isNil
**Describe the bug**
MNU receiver of #getCommentsFor: isNil

**To Reproduce**
1. Open the browser
2. Create a class `A`
3. add a method to this class:
```smalltalk
test

	| a b |

	a := (Object
		name: 'a') a: #().
	b := (Foo
		name: 'b').
```
4. try to copy (CTRL+C) the text `) a: #().`

**Expected behavior**
The text should be copied without any try to parse it.

**Version information:**
Pharo-9.0.0+build.699.sha.2ac6d670db10b42b7404cf629042b611f4c6e8ec (64 Bit)

MarcusDenker:
- change #stop in RBMethodNode to only fall back on decompiled code if the source is nil
- fix a test

fixes #7257

### DIFF
--- a/src/AST-Core/RBMethodNode.class.st
+++ b/src/AST-Core/RBMethodNode.class.st
@@ -708,7 +708,7 @@ RBMethodNode >> statements: aCollection [
 
 { #category : #accessing }
 RBMethodNode >> stop [
-	^ self newSource size
+	^(self sourceCode ifNotNil: [self sourceCode] ifNil: [ self formattedCode ]) size
 ]
 
 { #category : #accessing }
--- a/src/AST-Core/RBMethodNode.class.st
+++ b/src/AST-Core/RBMethodNode.class.st
@@ -708,7 +708,7 @@ RBMethodNode >> statements: aCollection [
 
 { #category : #accessing }
 RBMethodNode >> stop [
-	^(self sourceCode ifNotNil: [self sourceCode] ifNil: [ self formattedCode ]) size
+	^(self sourceCode ifNotNil: [:src | src] ifNil: [ self formattedCode ]) size
 ]
 
 { #category : #accessing }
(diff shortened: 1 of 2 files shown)


src/AST-Core/RBMethodNode.class.st
```diff
@@ -708,7 +708,7 @@ RBMethodNode >> statements: aCollection [
 
 { #category : #accessing }
 RBMethodNode >> stop [
-	^ self newSource size
+	^(self sourceCode ifNotNil: [self sourceCode] ifNil: [ self formattedCode ]) size
```
Ducasse: marcus may be can rewrite it as 

(self sourceCode ifNotNil: [ :s | s ] ifNil: [ self formattedCode ]) size

MarcusDenker: I committed an update with that

src/AST-Core/RBMethodNode.class.st
```diff
@@ -708,7 +708,7 @@ RBMethodNode >> statements: aCollection [
 
 { #category : #accessing }
 RBMethodNode >> stop [
-	^ self newSource size
+	^(self sourceCode ifNotNil: [:src | src] ifNil: [ self formattedCode ]) size
```
demarey, line 711: Why not simply `^ (self sourceCode ifNil: [ self formattedCode ]) size`?

MarcusDenker: yes. But I will do that in another PR. I want this in as it destroys my productivity.

MarcusDenker: We should in general not hold up fixes due to trivial things like this. Better do another PR

PR #7376, opened 2020-09-25
7374-DNU-isDynamicArray-when-editing-code

### LINKED ISSUE #7374: DNU #isDynamicArray when editing code
**Describe the bug**

**To Reproduce**
1. add a method with (self sourceCode ifNotNil: [self sourceCode] ifNil: [ self formattedCode ]) size
2. remove the last )

The real problem behind is that the AST that the compiler creates has a message send without parent pointer.
But I think we should do two steps:

1. workaround, as it is really annoying
2. second issue for the real problem

The workaround is to just check for nil in #isInlinedLoop

MarcusDenker:
Workaround. Can be removed when the AST is not created with nil parent anymore. fixes #7374


MarcusDenker:
I merge this because it is really bad for productivity.. it's just an intermediate workaround, there is an issue for the real problem

PR #7380, opened 2020-09-25
EpMonitor: Remove old code for subscribing to test announcer.

### LINKED ISSUE #7379: EpMonitor: Remove old code for subscribing to test announcer
**Describe the problem**
By 2015, Epicea logged test runs and results. For example, when Test Runner run a suite, the log registered the successes and failures. At some point, that became a package extension, which is still in SmalltalkHub as "Epicea-SUnit". That package requires this code in EpMonitor that I propose to delete in this issue. Why? 1. it hasn't been used for years but it'd be easy to recover after delete; 2. AFAIR, the feedback I got was that logging test runs implied too much noise in the log of changes.

**Classes involved**
EpMonitor

tinchodias:
By 2015, Epicea logged test runs and results. For example, when Test Runner run a suite, the log registered the successes and failures. At some point, that became a package extension, which is still in SmalltalkHub as "Epicea-SUnit". That package requires this code in EpMonitor that I propose to delete in this issue. Why? 1. it hasn't been used for years but it'd be easy to recover after delete; 2. AFAIR, the feedback I got was that logging test runs implied too much noise in the log of changes.

Fixes #7379.

### DIFF
--- a/src/Epicea/EpMonitor.class.st
+++ b/src/Epicea/EpMonitor.class.st
@@ -9,7 +9,6 @@ Class {
 	#instVars : [
 		'systemAnnouncer',
 		'log',
-		'testAnnouncer',
 		'entryReferenceByEvent',
 		'jobAnnouncer',
 		'announcer',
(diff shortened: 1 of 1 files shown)


tinchodias:
To load this code, ensure the code change logging is not enabled with `EpMonitor current disable`.

MarcusDenker (approved):

Issue #6318, opened 2020-05-07
DoubleWordArray is not usable...


stacksmith:
DoubleWordArray is not capable of storing DoubleWords - only 32-bit values work.    It appears to allocate and read, but to store, primitive 61 expects a SmallInteger.

Looking at Squeak (which has no such issues), DoubleWordArray class is defined ass `ArrayedCollection variableDoubleWordSubclass: #DoubleWordArray...` while Pharo defines it as `ArrayedCollection variableWordSubclass: #DoubleWordArray`.  

Is there a plan to fix this?    This was reported a while back: [related issue 5956](https://github.com/pharo-project/pharo/issues/5956).   There is even a proposed patch there.  It seems fairly essential to store 64-bit arrayed data on 64-bit machines.

Example:
`{1<<32} as: DoubleWordArray `

svenvc:
Could you provide changes/fixes with tests ?

stacksmith:
@svenvc   I am slightly out of my comfort zone, but the file (by  leves@caesar.elte.hu ) you linked from issue 5956 contained among other things 3 definitions (see below) that **apparently** fix the problem.   

After filing in I had no issues storing larger ints.  I have reservations, because it looks like the fix introduces `variableDoubleWordSubclass:..` and redefines `DoubleWordArray`, but looking at DoubleWordArray with a browser still shows `variableWordSubclass`.  It acts as if it uses the new layout correctly, and seems stable.

**I will therefore respectfully ask for someone who really understands the VM and the object layout and creation mechanism to look it over and integrate into Pharo.**  It seems like a trivial fix.

Here is the snippet I am using:
```
!Class methodsFor: 'subclass creation - variableWord' stamp: 'ul 3/28/2020 15:44'!
variableDoubleWordSubclass: className instanceVariableNames: instVarNameList classVariableNames: classVarNames package: cat

	^ self
		variableDoubleWordSubclass: className
		instanceVariableNames: instVarNameList
		classVariableNames: classVarNames
		poolDictionaries: ''
		package: cat! !

!Class methodsFor: 'subclass creation - variableWord' stamp: 'ul 3/28/2020 15:43'!
variableDoubleWordSubclass: t instanceVariableNames: f classVariableNames: d poolDictionaries: s package: cat

	^ self classInstaller
		make: [ :builder | 
			builder
				superclass: self;
				name: t;
				layoutClass: DoubleWordLayout;
				slots: f asSlotCollection;
[...]

Issue #5956, opened 2020-03-21
Wrong format of DoubleWordArray


guillep:
Report from discord

> I investigated a bit more, and found that it is not a VM bug but the
> lack of support for variable double word classes in Pharo images.
> So, primitive 61 will fail not because of a VM bug but because the
> receiver, a DoubleWordArray is not a variable double word subclass but a
> variable word subclass in Pharo. So, it has 32-bit fields instead of
> 64-bit fields. Shame...

**To Reproduce**

Evaluate
```smalltalk
{ 1 << 32 } as: DoubleWordArray
```

**Expected behavior**
A double word array with a single element of value (1<<32)

**Version information (please complete the following information, you can get it from Help->about):**

I think this applies to all 64 bit images we have up to today.

guillep:
We should check class layouts correspond in the `DoubleWordArray` class

svenvc:
Here is a changeset by leves@caesar.elte.hu that is relevant, though it also contains other changes (I believe we renamed digitAt and friends).

So, this needs to be checked.

He writes:

> I've attached a changeset. If you load it before the Cryptography packages, CryptographyHashing will work in Pharo 8 and in the current Pharo 9 alpha. Some tests will fail, because random strings are generated with the Random class, which is different in Squeak and Pharo but still,
everything will work.

[Pharo-DoubleWordArray-patch.cs.txt](https://github.com/pharo-project/pharo/files/4397695/Pharo-DoubleWordArray-patch.cs.txt)

The last 3 changes are the relevant ones, I think.

MarcusDenker:
I merged the image level changes, but it is not completely solved yet: I could not change the layout of the classes as they could not be loaded from git with the changed layout. I reopen this issue for this next step.

(I need to check what exactly fails)

MarcusDenker:
(The changes in the image are only those related to create variableDoubleWordSubclass and variableDoubleByteSubclass)

MarcusDenker:
This has been fixed!

PR #7382, opened 2020-09-26
kindOfSubclass-ask-Layout

MarcusDenker:
#kindOfSubclass has a huge if... and misses cases. e.g. there is no variableDoubleWordSubclass: and no variableDoubleByteSubclass (and the case for compiledmethod is missing, too).

This PR implements kindOfSubclass on the AbstractLayoyt hierarchy instead. This is a first step for solving #6318 and #5956


Ducasse:
Yes and I think that in the future we should be able to get rid of it.

PR #7383, opened 2020-09-26
ReservedVariable-Environment

MarcusDenker:
The classes modeling self, super and thisContext reference the global environment directly (via SystemNavigation new).

This PR instead adds the environment as state. This will allow us to later have different instances for different environments. For now the envrionment remains hardcodes, but just in #initialize instead of all #usingMethods implementations.

(The idea is that all subclasses of Variable should implement #scope. They should know the scope that will return them when looking up by name using #lookupVar:. The scope for self, super and thisContext will be the global environment)

The PR in addition adds a missing visitor methid to the completion builder.

### DIFF
--- a/src/HeuristicCompletion-Model/CoASTResultSetBuilder.class.st
+++ b/src/HeuristicCompletion-Model/CoASTResultSetBuilder.class.st
@@ -111,6 +111,11 @@ CoASTResultSetBuilder >> visitGlobalNode: aRBVariableNode [
 ]
 
 { #category : #visiting }
+CoASTResultSetBuilder >> visitInstanceVariableNode: aRBVariableNode [ 
+	^ self visitNode: aRBVariableNode
+]
+
+{ #category : #visiting }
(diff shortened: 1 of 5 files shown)


src/HeuristicCompletion-Model/CoASTResultSetBuilder.class.st
```diff
@@ -110,6 +110,11 @@ CoASTResultSetBuilder >> visitGlobalNode: aRBVariableNode [
 	^ self visitVariableNode: aRBVariableNode
 ]
 
+{ #category : #visiting }
```
Ducasse, line 113: Is this method part of this PR?

MarcusDenker: I ran into a DNU and decided to just fix it as part of this PR.

 (the code completion re-implements the visitor without using the Trait or inheriting, thus the backward compatibility method put in place is not helping here...)

PR #7385, opened 2020-09-27
betterDefinitionOnSubcomponents

Ducasse:
Fixing missing quotes


Ducasse:
This is strange because on my machine all the tests are green. 
I should have lost a change.

PR #7386, opened 2020-09-27
NewDefinitionUsingStreamPR

Ducasse:
fixing definition without slots.... there is duplication and I start to get annoyed.


Ducasse:
Broken tests are unrelated (roassal)

PR #7388, opened 2020-09-28
7387-typeOfClass-should-delegate-to-the-Layout-Hierarchy

### LINKED ISSUE #7387: #typeOfClass should delegate to the Layout Hierarchy
#typeOfClass is a huge "IF" that actually misses cases: DoubleWordArray and DoubleByteArray  even if the layout is correct will answer wrongly.

Solution: forward to the layout, implement #typeOfClass  there. Then it will for DoubleByteArray automatically return the default case, the name of the layout class (#DoubleByteLayout), which should make it then possible to save classes like that with Git.

We have to keep the special symbols for now to stay compatible with Monticello, but with some care we could default to the name of the layout class later for all cases.

MarcusDenker:
#typeOfClass is a huge "IF" that actually misses cases: DoubleWordArray and DoubleByteArray even if the layout is correct will answer wrongly.

Solution: forward to the layout, implement #typeOfClass there. Then it will for DoubleByteArray automatically return the default case, the name of the layout class (#DoubleByteLayout), which should make it then possible to save classes like that with Git.

we can remove #isCustomLayout now, too. Custom layout now use the name of the layout class by default.

### DIFF
--- a/src/Kernel/AbstractLayout.class.st
+++ b/src/Kernel/AbstractLayout.class.st
@@ -109,11 +109,6 @@ AbstractLayout >> isBits [
 ]
 
 { #category : #testing }
-AbstractLayout >> isCustomLayout [
-	^false
-]
-
-{ #category : #testing }
 AbstractLayout >> isFixedLayout [
 	^false
 ]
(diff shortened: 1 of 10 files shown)


MarcusDenker:
fixes https://github.com/pharo-project/pharo/issues/7387

Ducasse:
Yes more polymorphism else cases!!!

Issue #7359, opened 2020-09-24
Spec: some tests modify literal arrays


MarcusDenker:
**Describe the bug**
Recompile the image with #optionReadOnlyLiterals enabled, these methods modify literal arrays:

SpAbstractListPresenterTest>>#setUp
SpAbstractListPresenterTest>>#testSetSortingBlockBeforeItems
SpAbstractListPresenterTest>>#testSortingBlock
SpComponentListPresenterTest>>#testSetSortingBlockBeforeItems
SpComponentListPresenterTest>>#testSortingBlock
SpDropListPresenterTest>>#testSortingBlock

A trivial way of solving the failing test is to add a "copy" after the literal array.

MarcusDenker:
How to reproduce: add this pragma to the method:

	<compilerOptions: #(+ optionReadOnlyLiterals )>

demarey:
Now on hold until PR is integrated into Spec, then merged into Pharo

PR #7391, opened 2020-09-29
7359-Spec-some-tests-modify-literal-arrays

MarcusDenker:
Workaround for #7359: the real fix will be integrated with the next spec integration, adding copy here will unblock the work on read-only literals.


MarcusDenker:
failing tests not related

PR #7403, opened 2020-10-01
Name Analysis: Lookup Slots In Class

MarcusDenker:
Instead of adding all Slots to the instance scope, we can do the lookup in the Class itself (#lookupVar:).

- Improve hasBindingThatBeginsWith: on class to take slot names into account
- add #scope to Slot: the scope of a slot is the defining class

### DIFF
--- a/src/Kernel/Slot.class.st
+++ b/src/Kernel/Slot.class.st
@@ -301,6 +301,11 @@ Slot >> readInContext: aContext [
 ]
 
 { #category : #accessing }
+Slot >> scope [
+	^ definingClass
+]
+
+{ #category : #accessing }
 Slot >> scope: aScope [
 	"ignored, subclasses can override to analyze the scope they are to be installed in"
 ]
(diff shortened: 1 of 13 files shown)


src/Kernel/Slot.class.st
```diff
@@ -300,6 +300,11 @@ Slot >> readInContext: aContext [
 	^self read: aContext receiver
 ]
 
+{ #category : #accessing }
+Slot >> scope [
+	^ definingClass
```
dionisiydk, line 305: This may be not correct. But I did not check users of #scope.  
Is #scope about a visibility of a variable? 
If true then I think it should be #owningClass because the visibility of the variable inherited from a trait is a target class. 
From the other side I am not sure (never used it) does method of a class with stateful trait have an access to a trait variables?

dionisiydk (approved):

src/Kernel/Slot.class.st
MarcusDenker: Yes, it is #owningClass... the class where the slot is in that has the offsets correct to read the slot (the slot in the Trait knows nothing about the class...)

I will put it on the backlog, I want to write tests for alll the cases (this is not used so we can merge nevertheless)

PR #7404, opened 2020-10-01
5956-Wrong-format-of-DoubleWordArray

### LINKED ISSUE #5956: Wrong format of DoubleWordArray

MarcusDenker:
- add test testDoubleArraysHaveCorrectLayout
- fix MC #kindOfSubclass to ask the layout for a string
- add minimal old style class definition methods for DoubleWord and DoubleByte subclasses
- change the layout of DoubleByteArray/DoubleWordArray, the fixes make sure that Iceberg does indeed save this fix

fixes #5956

### DIFF
--- a/src/Collections-Native/DoubleByteArray.class.st
+++ b/src/Collections-Native/DoubleByteArray.class.st
@@ -6,7 +6,7 @@ For further comments read comments of `WordArray`.
 Class {
 	#name : #DoubleByteArray,
 	#superclass : #ArrayedCollection,
-	#type : #words,
+	#type : #DoubleByteLayout,
 	#category : #'Collections-Native-Base'
 }
 
--- a/src/Collections-Native/DoubleByteArray.class.st
+++ b/src/Collections-Native/DoubleByteArray.class.st
@@ -6,7 +6,7 @@ For further comments read comments of `WordArray`.
 Class {
 	#name : #DoubleByteArray,
 	#superclass : #ArrayedCollection,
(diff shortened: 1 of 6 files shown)


MarcusDenker:
I fear the problem is now that the bootstrap is done with an old Pharo, so we can not load the definition of a class with a the layout type #DoubleWordLayout. 

I improved the error printout in #kindOfSubclass, but the bootstrap does not see that change:

'Errors in script loaded from /builds/workspace/uest_and_branch_Pipeline_PR-7404/64/bootstrap/scripts/bootstrap.st'
Error: Unrecognized class type
MCClassDefinition(Object)>>error:
MCClassDefinition>>kindOfSubclass

MarcusDenker:
I changed the PR to just be the code changes, but not change the layout (as we can not load this in the bootrap/reload)

I suggest we merge this as an intermediate step

PR #7408, opened 2020-10-01
Debugger-should-show-Workspace-bindings

MarcusDenker:
Imagine you define a binding in a workspace:

temp := 1.

and then you "doit" this code:

self halt

you expect the variable to be shown in the debugger as a temp. 

The PR is a fix to do exactly that.
- we need to make sure to add the doit compiled method to the AST cache
- #allTemps of the Requesterscope needs to return a list of all Variables defined by the Requestor

And it works! it seems there are sometimes odd bindings for "self" generated, we need to check why but that is another PR

### DIFF
--- a/src/AST-Core/RBMethodNode.class.st
+++ b/src/AST-Core/RBMethodNode.class.st
@@ -376,6 +376,11 @@ RBMethodNode >> initialize [
 ]
 
 { #category : #testing }
+RBMethodNode >> isDoIt [
+	^ self selector isDoIt
+]
+
+{ #category : #testing }
 RBMethodNode >> isFaulty [
 	(self arguments anySatisfy: [:each | each isFaulty]) ifTrue:[ ^true].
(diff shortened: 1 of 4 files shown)


StevenCostiou (approved):
:)

dionisiydk (approved):
We should have tests for this

MarcusDenker:
Hmm... more complex then I thought... the problem is that the AST insists on printing using the original source, but the transformed method has the AST not in sync. If we make sure they are, then interactive use (e.g. eval in the workspace) has the wrong offset infos to print a message in the requestor.

This shows again that we need a better Doit Model... 

I think an intermediate solution is to fall back to pretty printing when doing a printOn: on an AST node, but just in case it is a DoIt.

MarcusDenker:
This now fixes the problem by relying on the decompiled code in DoITs when asking #sourceCode.

This is more a workaround then a fix, but it should be an ok solution for now.

PR #943, opened 2018-02-22
21400 Use category "utilities" instead of "utils" in NECClassVarTypeG…

astares:
…uesserTest RBParseTreeRewriterTest RPackageTestCase SystemNaviga

NECClassVarTypeGuesserTest
RBParseTreeRewriterTest
RPackageTestCase
SystemNavigationTest
TextAlignment (class/instance)
TextAlignment
TraitTest

Only recategorization, no change in behavior

PR #7415, opened 2020-10-04
Fix_for_PluggableSliderMorph

majella67:
Fix for Pharo spec project SpSliderPresenter #value returns values in the range 0..1 and not #min..#max #943


Ducasse:
Thanks a lot!!!

PR #7418, opened 2020-10-05
SourceFileBufferedReadWriteStreamTest-readOnlyLiteral

MarcusDenker:
SourceFileBufferedReadWriteStreamTest should not use literal strings directly when writing (as it changes the literals...)

### DIFF
--- a/src/Tests/SourceFileBufferedReadWriteStreamTest.class.st
+++ b/src/Tests/SourceFileBufferedReadWriteStreamTest.class.st
@@ -120,7 +120,7 @@ SourceFileBufferedReadWriteStreamTest >> testEnsureWrittenPositionFlushesComplet
 SourceFileBufferedReadWriteStreamTest >> testNextPutAllStartingAt [
 	| stringStream |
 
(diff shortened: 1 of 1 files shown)


Ducasse:
While reading the changes, ie. copying the argument. I wondering if we could not provide a method that will send the copy to the argument and one that does not know. Because we put the logic on the client and I would prefer to have it internalised.

MarcusDenker:
I do not think that there are many use cases for that. We do it here only because we create a WriteStream on a literal.. which you just should not do.

Ducasse:
Oki
So why not changing WriteStream with: '' into WriteStream with: String new.
This way we do not promote a special usage.

MarcusDenker:
empty arrays like objects are an interesting case in general: they can only grow by being copied, in which case the literal is not changed.

The Stream re-uses the underlying string as long as it does not need to grow, it creates a copy as soon as it grows.

(old Smalltalk -80 used become far more for these things (not sure if for arrays, but for Dictionary it did...), as it was a very cheap operation with the object table)

PR #7419, opened 2020-10-05
Playground-Self-Bindings

MarcusDenker:
The Playground got binding to #self added in some cases (e.g. "Doit" 'self halt' in the playground and then check the playground bindings).

The reason for that is that CoVariableValueMessageHeuristic was directly calling #bindingOf:, which adds the binding if it is not there.

The solution is to first check with  #hasBindingOf:


MarcusDenker:
maybe we should explore if we can make the creation of Requestor bindings explicit, having bindingOf: and lookupVar: creating them on the fly is odd... but I would do that as  another step as it would be a change with much more potential impact (and I do not even know it is possible)

PR #7425, opened 2020-10-05
7414-division-of-number-by-array-fails-with-doesNotUnderstand-isZero

### LINKED ISSUE #7414: division of number by array fails with doesNotUnderstand isZero
**Describe the bug**
pharo supports array arithmetic, eg:
5 + #(3 2) answers #(8 7)
This works for subtraction and multiplication also, but not division

**To Reproduce**
In a playground, evaluate 10 / #(20) 

**Expected behavior**
It should answer #(1 / 2).

**Version information:**
Pharo 9. 

**Expected development cost**
minimal. Implementing  

```
isZero
   ^false.
```
should do it. The Collection class implements + - * etc for array arithmetic, so that might be a good place.

MarcusDenker:
Fixes 7414

- implement #isZero
- add some tests

### DIFF
--- a/src/Collections-Arithmetic/Collection.extension.st
+++ b/src/Collections-Arithmetic/Collection.extension.st
@@ -109,6 +109,11 @@ Collection >> floor [
 ]
 
 { #category : #'*Collections-arithmetic-collectors' }
+Collection >> isZero [
+   ^ false
+]
+
+{ #category : #'*Collections-arithmetic-collectors' }
 Collection >> ln [
 	^self collect: [:each | each ln]
 ]
--- a/src/Collections-Arithmetic-Tests/CollectionArithmeticTest.class.st
+++ b/src/Collections-Arithmetic-Tests/CollectionArithmeticTest.class.st
@@ -65,6 +65,14 @@ CollectionArithmeticTest >> testAverageWithEmptySet [
 ]
 
 { #category : #tests }
+CollectionArithmeticTest >> testDevision [
+	| collection |
+	collection := #(2 4).
+	self assert: collection / 2 equals: #(1 2).
+	self assert: 10 / #(20) equals: {1/2}
+]
+
+{ #category : #tests }
 CollectionArithmeticTest >> testRunningAverage [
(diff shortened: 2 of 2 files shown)


Ducasse:
may be we could also check that the argument is a number before asking for isZero because it looks strange to have isZero in Collection.

src/Collections-Arithmetic/Collection.extension.st
```diff
@@ -108,6 +108,11 @@ Collection >> floor [
 	^ self collect: [:a | a floor]
 ]
 
+{ #category : #'*Collections-arithmetic-collectors' }
+Collection >> isZero [
```
guillep, line 112: We are reviewing this issue with Stef and we don't like to pollute collections with this kind of interface.

We were playing a bit around and it would seem that replacing `SmallInteger >> #/`with:

```smalltalk
/ aNumber 
	"Primitive. This primitive (for /) divides the receiver by the argument
	and returns the result if the division is exact. Fail if the result is not a
	whole integer. Fail if the argument is 0 or is not a SmallInteger. Optional.
	No Lookup. See Object documentation whatIsAPrimitive."

	<primitive: 10>
	aNumber = 0 ifTrue: [^(ZeroDivide dividend: self) signal].
	^(aNumber isMemberOf: SmallInteger)
		ifTrue: [(Fraction numerator: self denominator: aNumber) reduced]
		ifFalse: [super / aNumber]
```

This checks zero directly with `= 0`, meaning this will only work with numbers, but in case the denominator is not a number then it will fall back to the superclass method that eventually makes the double dispatch with `adaptToNumber:andSend:`. This will probably make the slower case of a zero division a bit slower, but it's already slow since the exception will force the reification of the stack...

On the other side, this change removes some pressure on the API, we do not need to implement `isZero` in all possible denominators, just `adaptToNumber:andSend:`.

Ducasse:
We played with it with guille and I think that we have a nicer solution.

guillep:
See https://github.com/pharo-project/pharo/pull/7436

PR #7434, opened 2020-10-06
7412-preparing-the-integration-of-fluid-class-part-2

Ducasse:
Adding a new message to encode the kind of the layout.

### DIFF
--- a/src/Kernel/AbstractLayout.class.st
+++ b/src/Kernel/AbstractLayout.class.st
@@ -8,8 +8,14 @@ Class {
 	#instVars : [
 		'host'
 	],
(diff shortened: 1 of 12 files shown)


MarcusDenker:
the commit somehow removes the #category  on the level of the git format:

#'Kernel-Layout' is turned into #Kernel

Ducasse:
yes I do not know why

Ducasse:
I think that this is the move to package refactoring that is changing this and I do not know what to do.

Ducasse:
I do not understand because I commit I removed explicitly from the selection the class definition changes :(. 
If I would have blindly committed I could understand but this is not the case :(

PR #7435, opened 2020-10-06
Variable-scope-addTests

MarcusDenker:
- add tests for #scope for Slot, Temps and Globals
- fix #scope on GlobalVariable and Slot

### DIFF
--- a/src/Kernel/GlobalVariable.class.st
+++ b/src/Kernel/GlobalVariable.class.st
@@ -33,5 +33,6 @@ GlobalVariable >> isGlobalVariable [
 
 { #category : #accessing }
 GlobalVariable >> scope [
-	^ self definingClass environment
+	"The scope of Globals is Smalltalk globals"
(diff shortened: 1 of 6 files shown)


src/Kernel/GlobalVariable.class.st
```diff
@@ -33,6 +33,5 @@ GlobalVariable >> isGlobalVariable [
 
 { #category : #accessing }
 GlobalVariable >> scope [
-	"The scope of Globals is Smalltalk globals"
-	^ Smalltalk globals
+	^ self definingClass environment
```
dionisiydk: I wonder that there is definingClass for globals. What is that?

MarcusDenker: Ups, true, that makes not sense. I will put it back for now to a hard-coded global environment. (I am reluctant to add an environment pointer to GlobalVariable as it would cost some memory...)

PR #7436, opened 2020-10-06
Cleaning integer division

guillep:
- zero division with non-numbers is checked in superclass (and it's already a slow case... No need for optimization)
 - removing special fraction case, it's managed by superclass

Improvement over #7425


MarcusDenker (approved):
ok

PR #7438, opened 2020-10-07
7437-prepare-fluid-class-integration-part-3

Ducasse:
Fixes: 7437 moving some extensions to classDefinition nodes.


Ducasse:
Much better now. 
I will slowly arrive to the point that I have less and less to integrate

PR #7448, opened 2020-10-08
Deprecate-debuggerMap

MarcusDenker:
Finally deprecate DebuggerMethodMapOpal


Ducasse:
You have some as yet unclassified.

MarcusDenker:
I will redo it.. there is some problem moving a class to Deprecated when it has already deprecated methods

PR #7450, opened 2020-10-08
7443-Compiling-a-method-with-a-non-declared-temps-raises-a-DNU

### LINKED ISSUE #7443: Compiling a method with a non declared temps raises a DNU
```
| foo|
bar := 1.
```

raises  rubScrolledTextMorph>> bindings DNU

```
allTemps
	"the requestor has the temps defined in the Requestor as WorkspaceVariables"
	^requestor bindings associations 
```

MarcusDenker:
implement #bindings to fix #7443


dionisiydk (approved):

MarcusDenker:
I think we should merge this even though the code that directly needed it has been reverted.

the idea is that having all requestors implementing #bindings is good, and it is nice for then trying to do another pass on https://github.com/pharo-project/pharo/pull/7408

PR #7459, opened 2020-10-09
2020-10-09-Bottom

### LINKED ISSUE #7458: Fluid Class integration part 5
Now integrating the class builder and the message syntax

Ducasse:
fixes: #7458 adding the fluidbuilders and tests

### DIFF
--- a/src/Shift-ClassBuilder-Tests/FluidClassBuilderTest.class.st
+++ b/src/Shift-ClassBuilder-Tests/FluidClassBuilderTest.class.st
@@ -1,0 +1,243 @@
+Class {
+	#name : #FluidClassBuilderTest,
+	#superclass : #TestCase,
+	#instVars : [
+		'builder'
+	],
+	#category : #'Shift-ClassBuilder-Tests'
+}
+
+{ #category : #running }
+FluidClassBuilderTest >> builder [ 	
+
+	^  builder
+]
+
+{ #category : #'tests - mandatory' }
+FluidClassBuilderTest >> removeFromSystem [
+
+	self class environment at: #MyClass 
+		ifPresent: [:p | p removeFromSystem: false ]
+		ifAbsent: [  ] .
+	
+]
+
+{ #category : #running }
+FluidClassBuilderTest >> setUp [
+	super setUp.
+	builder := FluidClassBuilder new 
+		superclassToBuild: Object;
+		classNameToBuild: #Point33;
+		package: 'FakedPackage'
+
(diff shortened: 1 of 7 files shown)


Ducasse:
Simply adding the fluidclassBuilder and UndefinedObject>> < to shiftClassBuilder package does not seem to work 
I get

```

You will need to log out and log back in for the limits to take effect.
For more information please see
https://github.com/OpenSmalltalk/opensmalltalk-vm/releases/tag/r3732#linux
[31mMessageNotUnderstood: Context>> #sourceNodeExecuted
[0mContext(Object)>>doesNotUnderstand: #sourceNodeExecuted
UndefinedObject>>doesNotUnderstand: #classSide
[31mSubscriptOutOfBounds: 0
[0mArray(Object)>>errorSubscriptBounds:
Array(Object)>>at:
Array(SequenceableCollection)>>last
BlockClosure>>endPC
BlockClosure>>abstractBytecodeMessagesDo:
BlockClosure>>isClean
Context>>printOn:
Context>>printDetails:
Context>>errorReportOn:
```
so I will try to add it in an extra package.

src/Shift-ClassBuilder-Tests/FluidClassBuilderTest.class.st
```diff
@@ -0,0 +1,243 @@
...
+	builder := Object < #Point.
+	self assert: builder superclassToBuild equals: Object.
+	self assert: builder classNameToBuild equals: #Point.
+	
+]
+
+{ #category : #'tests - mandatory' }
+FluidClassBuilderTest >> testCreateBuilderWithNil [
+
+	builder := nil < #NewProtoObject.
```
dionisiydk, line 132: #< is really bad message for nil object. It will eat errors in many comparison cases when variables are nil: 
```Smalltalk
aNumber < 10
aString < 'anotherString'
```

Ducasse:
What do you propose?
I proposed <| (it looks like uml).
and it can be changed super simply.

src/Shift-ClassBuilder-Tests/FluidClassBuilderTest.class.st
Ducasse: BTW I think that expecting that aString is nil in a comparison like <'anotherstring' is definively a bug so I do not understand why getting an error from the classbuilder is worse than now.
I will move this discussion to https://github.com/pharo-project/pharo/issues/7471

PR #7464, opened 2020-10-10
7458-Fluid-Class-integration-part-5

### LINKED ISSUE #7458: Fluid Class integration part 5

Ducasse:
Fixes: #7458 now by creating a separate package


Ducasse:
I really do not get why UndefinedObject >> < is not working while it is working in my image and it is part of the build.

Issue #7470, opened 2020-10-11
Fluid Class: A class definition can be expanded


Ducasse:
Now I do not know how to make a menu working in calypso. 
Here is the email that Denis sent me. I made it works but after I could not find why it was not working. 

I created a context and a command. 

Hi Stef.

You will need to introduce new contexts for Calypso to attach commands into these specific tools:

- ClyClassDefinitionContext as a subclass of ClySourceCodeContext.
You will need to return its instance from #createTextContext method of the class definition editor. See as example ClyMethodCodeEditorToolMorph>>#createTextContext. 

- ClyClassCommentContext as a subclass of ClyTextEditorContext (or if you have any ast for comment syntax you can try to reuse ClySourceCodeContext as superclass).
And return it from ClyClassCommentEditorToolMorph>>createTextContext.

With new context you will be able to annotation commands to enable them in these tools. For example:

YourCommand class >> sourceCodeMenuActivation
   <classAnnotation>
   ^SycSourceCodeMenuActivation byRootGroupItemOrder: 1.2 for: ClyClassDefinitionContext

It will add a new command to Source context submenu (which also opens by cmd+t).

PR #7476, opened 2020-10-11
7470 fluid class a class definition can be expanded

dionisiydk:
The problem was that new context was not used for selected AST nodes of class definition. 
Here it replaces ClySourceCodeContext in #createTextContext method. 
In addition I cleaned overriding methods of new context class.  ClySourceCodeContext implements them based on AST and it was working fine for class AST. No extra logic needed there (at least not for previous changes)


dionisiydk:
It fixes #7470

Ducasse:
Thanks denis. I will look to learn.

Issue #7256, opened 2020-09-07
HiSpecExampleTest(HiExamplesTest)>>testAllExamples is Timing out on the CI quite often


MarcusDenker:
**Describe the bug**

The CI fails often due to HiSpecExampleTest(HiExamplesTest)>>testAllExamples running into a timeout

MarcusDenker:
Even with a 3 minute timeout it fails sometimes. I propose to not run this on the CI

PR #7480, opened 2020-10-11
Better way to design HiExamplesTest

dionisiydk:
This is an example of how we can use a flexible SUnit design for cases like HiExamplesTest. 
SUnit allows to override a logic how to build a test suite for specific test class. Instead of running a collection of "custom validation rules" inside a single test we can represent them as a separate test instances for each rule.   

So here #testSelectors is overridden to return selectors of target example class. And #performTest is overridden to execute corresponding example methods instead of a methods of test class.

Check a suite of those tests:
```Smalltalk
HiSpecExampleTest buildSuite tests
```
Screen Shot 2020-10-11 at 19 08 19

With this approach the time to run HiExamplesTest should not exceed any limit because each example is running as a separate test which is quite fast.

### DIFF
--- a/src/Hiedra-Tests/HiExamplesTest.class.st
+++ b/src/Hiedra-Tests/HiExamplesTest.class.st
@@ -1,8 +1,16 @@
 Class {
 	#name : #HiExamplesTest,
 	#superclass : #TestCase,
+	#instVars : [
+		'example'
+	],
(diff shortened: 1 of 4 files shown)


dionisiydk:
It is related to #7256

Ducasse:
This is great. I will add it to the testing booklet.

PR #7484, opened 2020-10-12
7481-testSharedSlotNodeArePolymorphicToRBVariableNodes-is-failing-in-default-image

### LINKED ISSUE #7481: testSharedSlotNodeArePolymorphicToRBVariableNodes is failing in default image
And this is not related to fluid class definition (it happens when it is not enabled).

Ducasse:
Fixes: #7481 using the correct class parser is definitively a must.


Ducasse:
I have no idea how to find the tearDown method that is not well categorised in particular because I did not touch anything in this PR. 
That rule is superbad because it has the information but does not let us know it. 
I HATE so much such practice.

Ducasse:
this is HiExamplesTest so not related and I will fix it with the CRAPPY rules...

PR #7485, opened 2020-10-12
7473-Fluid-Class-clean-definitionWithSlotsdefinitionWithoutSlot

Ducasse:
Check all the tests of ClassDefinitionParser.
So far this is not related to the issue but 7481 but this is going in the same direction.


Ducasse:
Broken tests are unrelated and I will fix tomorrow the categorisation while correcting that badly designed rules. "oh yes there is badly categorised method" ..... bohhhhhhh crap.

PR #7491, opened 2020-10-13
7490-definition-on-classDescription-can-be-simplified

### LINKED ISSUE #7490: definition on classDescription can be simplified
As a first step.... can we just fix classDescription

Ducasse:
Fixes: #7490 just simplifying the definition of classDescription and subclasses to start with.


Ducasse:
The broken tests are not related.

PR #7497, opened 2020-10-13
[WIP] Fix Layout DoubleWordArray

MarcusDenker:
Change the layout of DoubleWordArray and DoubleByteArray to see why it fails to build with this fix


MarcusDenker:
I need to look at this in detail: the change needs a bootstrap code change, it seems. For later

PR #7502, opened 2020-10-14
Add backwards compatibility Pharo9 to Pharo8

guillep:
During development several changes made Pharo8 and Pharo9 incompatible.
We should reintroduce some deprecations to give one version of compatibility allowing migration.

 - OSWindowHandle should have been deprecated
 - Key should have been deprecated
 - send:toSuperOf: should be able to generate code for traits just as before sista compilation was the main one


MarcusDenker (approved):
Ok! The super is indeed not working for traits, but the place this is checked is wrong: the compiler backend is not the place to check things like that, we need to forbid it on the level of Traits.

PR #7508, opened 2020-10-14
Introduce symbolic platform independent name to the morphic keyDown and keyUp events

ronsaldo:
There is no platform independent tracking of the named keys in the morphic keyboard events for key down and key up events. The lack of this introduces limits the flexibility of what keys can actually be used in different morphs, and it also introduces bugs in the way that shortcuts are processed (they use the keyStroke event which should only be used for text input, and they are sensitive to caps lock which is wrong).


ronsaldo:
Do not integrate yet. I was just pointed to the existent KeyboardKey class, so I am modifying this PR and that class.

ronsaldo:
Please review the new version. This one uses the existent KeyboardKey class.

src/OSWindow-Core/OSKeySymbols.class.st
```diff
@@ -15,55 +15,29 @@ Class {
 		'OSK_7',
```
Ducasse, line 17: Ronie why only some are kept?

ronsaldo: I moved the generation and setting of these constants into a literal in the data method. I removed the key symbols that do not have proper mapping into a KeyboardKey and where I could not find a corresponding X11 key code to introduce a mapping for them. This is to ensure that all of the named symbols have a corresponding mapping in KeyboardKey. As for the removed symbols in the end, there are either: rare to be present, or rarely handled by an end-user application (e.g.: the Power button for shutting down the machine).

Ducasse: Could you add this kind of comment in the class?
Thanks.

Ducasse:
Thanks this is important to document decision. Because we can all read the code but understanding the reasons is different.

Issue #7357, opened 2020-09-23
DNU #text when doing a printIt in the Playground


MarcusDenker:
**Describe the bug**
The requestor used when executing code in the workspace misses a #text method

**To Reproduce**
Try to "print it" in a playground:

   [:each | [ :each ] ]

MarcusDenker:
@estebanlm

estebanlm:
and what has to be that text?

MarcusDenker:
it seems to be the code. In general, it looks like all this could be refactored and be much simpler...

MarcusDenker:
SpCodeInteractionModel>>#text needs to be implemented as:

text
	^self owner text


and it works

MarcusDenker:
another method I need:


SpCodeInteractionModel>>#bindings


bindings
	"no additional bindings provided by the Requestor"
	^#() asDictionary

MarcusDenker:
This has been fixed!

PR #7511, opened 2020-10-15
Doit-RequestorScope

MarcusDenker:
This is another try to not lose the RequestorScope for doits and to have #allTemps including the RequestorBindings.

- when you have a binding the the playground, the debugger will list it as one of the local variables
- With another change (that will be another PR), these variables can be accessed in the code, as they should be


MarcusDenker:
But we an only merge it after we add 

SpCodeInteractionModel>>#bindings

This PR thus depends on https://github.com/pharo-project/pharo/issues/7357 being resolved first

MarcusDenker:
Removed the dependency on the other PR ==> this can be merged if green

PR #7518, opened 2020-10-16
Cleanup-headerFlagForEncoder

MarcusDenker:
Small cleanup of code not needed


src/Kernel/CompiledMethod.class.st
```diff
@@ -124,13 +124,7 @@ CompiledMethod class >> handleFailingNewMethod: numberOfBytes header: headerWord
 CompiledMethod class >> headerFlagForEncoder: anEncoderClass [
 	anEncoderClass == PrimaryBytecodeSetEncoderClass ifTrue: [ ^ 0 ].
 	anEncoderClass == SecondaryBytecodeSetEncoderClass ifTrue: [ ^ SmallInteger minVal ].
```
dionisiydk, line 126: It should be simple "anEncoderClass headerFlagForMethods"

MarcusDenker: I was thinking of that, but wanted to not do it as part of the cleanup as it does change the semantics: the number now is not defined by the bytecode set, but by it being registered primary or secondary.

PR #7521, opened 2020-10-18
[Do not merge] Make debugIt operation to work transparently with temps

dionisiydk:
Make debugIt operation to work transparently with temps. No ThisContext trick anymore:

Screen Shot 2020-10-18 at 22 37 39

The part I did not solve is code highlighting. On screenshot #temp variable is red as undeclared. But the evaluation works correctly and I believe this approach is better than current form with ThisContext. The highlighting can be fixed later.

The implementation is based on special kind of variable (ForeignVariable) which provides a unified access to variables from any context. 
#asDoItForContext: method fills the compilation context with extra bindings using imported temps of original context as foreign variables.
Therefore OCExtraBindingScope defines allTemps as all defined bindings.
The doIt compilation is modified in the way that #asDoIt and #asDoItForContext return semantically analyzed ast ready for use.

Some changes were required for GT debugger which did not use variable objects but var names and indexes. Now it's based on first class variables


dionisiydk:
The experiment is interesting but there are too many problems which I am not ready to solve.
For debugIt cases the changes are fine. But currently the code for doIts is common for other cases like mustBeBoolean deoptimization and conditional breakpoints. They really need thisContext as parameter because they cache the expression method. In my approach the original context is stored inside expression variables and therefore they become invalid as soon as this context terminates.

Another found issue is related to copied temps:
```Smalltalk
blockContext astScope allTemps
```
It does not return actual copied temp but original method temp.

dionisiydk:
I did better fix with requestor scope

PR #7524, opened 2020-10-19
Variables-isTempVariable

MarcusDenker:
This is the final renaming of the is* methods of Variables. We still had isTemp, to unify it with all the other checks, we rename is to isTempVariable

All senders are changed, the isTemp methods are deprecated where it makes sense.

### DIFF
--- a/src/OpalCompiler-Core/TemporaryVariable.class.st
+++ b/src/OpalCompiler-Core/TemporaryVariable.class.st
@@ -26,13 +26,7 @@ TemporaryVariable >> definingNode [
 ]
 
 { #category : #testing }
-TemporaryVariable >> isTemp [
-
-	^ true
-]
-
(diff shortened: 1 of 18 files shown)


src/OpalCompiler-Core/TemporaryVariable.class.st
```diff
@@ -26,13 +26,7 @@ TemporaryVariable >> definingNode [
...
 
 { #category : #testing }
-TemporaryVariable >> isTemp [
-
-	^ true
-]
-
-{ #category : #testing }
-TemporaryVariable >> isTemporaryVariable [
+TemporaryVariable >> isTempVariable [
```
dionisiydk, line 29: What about class rename to TempVariable?

dionisiydk (approved):
Very good.

PR #7527, opened 2020-10-19
Introduce ForeignVariable to model a unified access to variables

dionisiydk:
Introduce special ForeignVariable to model a unified access to variables from any context. 
For example it allows to transparently use temps from the local context in doIt expressions:
```Smalltalk
| temp |
temp := 100.
foreignVariable := ForeignVariable named: #temp fromContext: thisContext.
 Smalltalk compiler
	bindings: {foreignVariable};
	evaluate: '1+temp'.
```
Using this variable we will be able to improve debugIt behavior over expression with temps.


MarcusDenker (approved):
Nice!

PR #7531, opened 2020-10-20
7421-Variable-breakpoints-do-not-work-on-class-variables

### LINKED ISSUE #7421: Variable breakpoints do not work on class variables
**Describe the bug**
Setting a variable breakpoint on a class variable causes a `MessageNotUnderstood` exception

**To Reproduce**
1. Create a class with a class variable
```Smalltalk
Object subclass: #MyClass
	instanceVariableNames: ''
	classVariableNames: 'myClassVariable'
	package: 'MyPack'
```
2. Create a method using the class variable, for example a basic setter.
```Smalltalk
myClassVariable: aValue
	myClassVariable := aValue
```
3. Select `myClassVariable` in the code of that method, right-click, and choose "Break on accesses to myClassVariable"
image

4. You will get the error: `MessageNotUnderstood: Metaclass>>#newBreakpointForVariable:`
image

**Version information:**
Ubuntu 20.04

**Additional context**
The variations of variable breakpoints ("Break on reads to myClassVariable" and "Break on writes to myClassVariables") do not work either and cause a similar exception, albeit on slightly different methods (`#watchVariableReads:inClass:`, `#watchVariableWrites:inClass:`) respectively)

MarcusDenker:
fixes #7421 

- implement breakpoints on Class variables
- add tests

### DIFF
--- a/src/Reflectivity-VariableBreakpoint/Class.extension.st
+++ b/src/Reflectivity-VariableBreakpoint/Class.extension.st
@@ -2,7 +2,11 @@ Extension { #name : #Class }
 
 { #category : #'*Reflectivity-VariableBreakpoint' }
 Class >> link: aMetaLink toVariableNamed: vName option: accessStrategy [
-	^ self link: aMetaLink toSlotNamed: vName option: accessStrategy
+	^aMetaLink
+		installOnVariableNamed: vName
+		for: self
+		option: accessStrategy
+		instanceSpecific: self intanceSpecificMetaLinksAvailable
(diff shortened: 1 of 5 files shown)


MarcusDenker:
I think we can simplify things now a lot with the first class variables (not for this PR but for later): instead of using names of variables we can very early get the Variable itself and then pass that around. This should allow to unify a lot of code between the different kinds of variables that is now there multiple times (for temps, for slots, for class vars...)

PR #7533, opened 2020-10-20
Improve the method isValidGlobalName

### LINKED ISSUE #7532: isValidGlobalName should be updated and improve
**Describe the bug**
`MyClass` is apparently now used in class template, and therefore, it cannot be used as a global variable name.

**To Reproduce**
1. `'MyClass' isValidGlobalName` should returns `false`

bergel:
Fix #7532

### DIFF
--- a/src/Collections-Strings/String.class.st
+++ b/src/Collections-Strings/String.class.st
@@ -1874,17 +1874,16 @@ String >> isString [
 
 { #category : #testing }
 String >> isValidGlobalName [
-
+	"Answer whether the string receiver can be used as a global name (e.g., name of a class, name of a trait)"
 	self ifEmpty: [ ^ false ].
 	
 	"reserverd default names"
-	self = 'NameOfSubclass' ifTrue: [ ^ false ].
-	self = 'TNameOfTrait' ifTrue: [ ^ false ].
+	(#('NameOfSubclass' 'MyClass' 'TNameOfTrait') includes: self) ifTrue: [ ^ false ].
 	
 	^ (self first isLetter 
-				and: [self first isUppercase]) 
-				and: [ self allSatisfy: [:character | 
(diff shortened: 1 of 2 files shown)


MarcusDenker:
failing tests, e.g:

FluidClassBuilderTest>>testCreateClassWithFullExpandedDefinitionKeepsTheMinimum

src/Collections-Strings/String.class.st
```diff
@@ -1874,17 +1874,16 @@ String >> isString [
...
 { #category : #testing }
 String >> isValidGlobalName [
-
+	"Answer whether the string receiver can be used as a global name (e.g., name of a class, name of a trait)"
 	self ifEmpty: [ ^ false ].
 	
 	"reserverd default names"
-	self = 'NameOfSubclass' ifTrue: [ ^ false ].
-	self = 'TNameOfTrait' ifTrue: [ ^ false ].
+	(#('NameOfSubclass' 'MyClass' 'TNameOfTrait') includes: self) ifTrue: [ ^ false ].
```
dionisiydk, line 1881: I did not know about this logic. But it's very surprizing. Why we can't have classes with those names? What makes them so special?

dionisiydk: It would be really nice to explain it in the method comment

astares: @bergel also it should be "reserved" - not "reserverd"

guillep:
@bergel there are broken tests and pending comments in the PR, and a couple of months of inactivity. I'll close it for now, please do not hesitate to reopen it.

astares:
@guillep one can not reopen as outside contributor due to permissions.

guillep:
I can reopen it myself if required, no problem :)

PR #7537, opened 2020-10-21
Context-printOn-no-isClean

MarcusDenker:
#isClean is quite slow as it traps DoesNotUnderstand for each bytecode. 

This PR adds a check that we do not use it as long as we are not compiling with clean blocks enabled.

I will add a faster scanner for clean blocks to the backlog (I really do not like this DNU handler based bytecode scanning...)


MarcusDenker:
Nope. I will do a special isCleanChecker subclass of InstructionClient instead

PR #7538, opened 2020-10-21
Move SystemAnnouncement extensions from Calypso

dionisiydk:
Calypso introduced nice methods for SystemAnnouncement like affectsMethod: , affectsClass:. They can be reused by other tools to analyze system changes. And therefore they should be packaged together with Announcement classes.


MarcusDenker (approved):
I think it is good to move them: this way we can use it to simplify all the methods that listen to announcements to see if a method was changed that they are interested in.

PR #7543, opened 2020-10-22
Fix crashes and memory leaks in OSSDL2AthensRenderer

ronsaldo:
This PR fixes crashes and memory leaks in OSSDL2AthensRenderer by removing the AthensCairoSDLSurface class and surface a plain AthensCairoSurface. This prevents a memory leak on intermediate SDL2 surface that was created for Cairo to render on it since that surface was never being released. This also prevents crashes due to an extra 1 pixel hack that is in the main AthensCairoSurface, but it was not being replicated in the AthensCairoSDLSurface subclass and matching SDL_Surface for pixel storage.


Ducasse:
Thanks ronie. This is cool to have you here working on this.

PR #7544, opened 2020-10-22
Do not disable freetype text hinting in Cairo

ronsaldo:
We are currently disabling all of the freetype face hinting options when rendering text with Cairo. This makes small text look horrible.


guillep:
Ah this is cool! Could we have a screenshot?

ronsaldo:
Technically for some reason the BitBlt version looks a bit bolder. I am wondering on whether this is due to freetype, or due to the way that we are compositing onto screen (is not normal alpha blending). 

BitBlt:
bitblt

Athens No Hinting (old version):
athens-nohinting

Athens Hinting (new version):
athens-hinting

For a greater impact, I recommend loading the images on separate Gimp Layers and comment it. What I notice is that the text without the hinting disabled looks slightly sharper, but it is still thin.

PR #7551, opened 2020-10-23
7535-Variable-Breakpoints-Wrong-Icon-in-gutter

### LINKED ISSUE #7535: Variable Breakpoints: Wrong Icon in gutter
**Describe the bug**
When adding a breakpoint for a variable, the icon is wrong

**To Reproduce**
1. select some ivar use in a method
2. menu, Debugging, "break on read to var"
4. See Metalink icon at the right, not the icon for breakpoints

**Expected behavior**
We should get only a Breakpoint icon

We did fix this for "normal" breakpoint. There the reason was that in #install, we were adding the metalink
before adding the breakpoint to the #breakpoints annotation on the AST node.

If we check the situation here, this is not possible as the link only knows after it has places where it will be placed

First thing to check: does #hasBreakpoint return true after adding the breakpoint to the variable?

MarcusDenker:
This improves how styling of MetaLinks vs Breakpoints is done.

- Now the Matalink has an option so we can configure if we want it to render the default gutter styler or not
- Breakpoint, Watch and ExecutionCounter now turn that off

This way we do not have to hard-code any knowledge in MetaLinkIconStyler>>#shouldStyleNode:

There is still an update problem for Variable Breakpoints (the method is not re-rendered after adding one). 
That will be fixed in anther step.

### DIFF
--- a/src/Reflectivity/MetaLink.class.st
+++ b/src/Reflectivity/MetaLink.class.st
@@ -58,6 +58,7 @@ MetaLink class >> defaultOptions [
 	+ optionWeakAfter 						  "do not use #ensure: for #after "
 	- optionAnnounce                    "do we announce adding / removing links? Slow"
 	- argsAsArray                       "pass all arguments in one array. By default off as it adds an array creation"
+	+ optionStyler							  "should the metalink be visualized in the Tools?"
 	)
 ]
 
--- a/src/Reflectivity-Tools/MetaLinkIconStyler.class.st
+++ b/src/Reflectivity-Tools/MetaLinkIconStyler.class.st
@@ -29,10 +29,6 @@ MetaLinkIconStyler >> iconLabel: aNode [
 
 { #category : #testing }
 MetaLinkIconStyler >> shouldStyleNode: aNode [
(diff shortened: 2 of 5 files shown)


src/Reflectivity/MetaLink.class.st
```diff
@@ -58,6 +58,7 @@ MetaLink class >> defaultOptions [
 	+ optionWeakAfter 						  "do not use #ensure: for #after "
 	- optionAnnounce                    "do we announce adding / removing links? Slow"
 	- argsAsArray                       "pass all arguments in one array. By default off as it adds an array creation"
+	+ optionStyler							  "should the metalink be visualized in the Tools?"
```
dionisiydk, line 61: It's quite confusing that all visible metalinks (like breakpoint) retract the optionStyler. 
What does it really mean to be visualized in Tools and what the difference from how breakpoints are shown?

dionisiydk (approved):

src/Reflectivity/MetaLink.class.st
MarcusDenker: yes, I am not even sure if metalinks should by default show the styler... to be improved, but the merged PR is a good first step

PR #7552, opened 2020-10-23
DNU in RubCurrentLineBarDecorator for #compose

### LINKED ISSUE #7550: DNU in RubCurrentLineBarDecorator for #compose
- latest Pharo 9 image on Windows (but also on Ubuntu)

```Evaluate 
|provider fontFolder file client done|
provider := FreeTypeFontProvider current.
fontFolder := provider imageRelativeFontFolder.
fontFolder ensureCreateDirectory.
file := fontFolder / 'download.zip'.
file deleteIfAbsent: [].
client := ZnClient new
			 	url: 'https://fonts.google.com/download?family=Inter';
				downloadTo: file.
fontFolder openInOSFileBrowser.
ZipArchive new
	readFrom: file;
	 extractAllTo: fontFolder.
	
file deleteIfAbsent: [].

done := Set new.
provider platformImageRelativeDirectories
		do: [ :each | provider updateFromDirectory: each done: done ]
		displayingProgress: 'Loading image relative font files'. 
		
font := FreeTypeFont fromFile: fontFolder / 'static' / 'Inter-Light-slnt=0.ttf' pointSize: 10.

StandardFonts listFont: font. 
StandardFonts menuFont: font. 
StandardFonts codeFont: font.
```
to install a new Font.

Then open the system browser, gives a DNU:

image

astares:
Fix #7550

### DIFF
--- a/src/FreeType/FreeTypeFont.class.st
+++ b/src/FreeType/FreeTypeFont.class.st
@@ -296,6 +296,11 @@ FreeTypeFont >> displayUnderlineOn: aDisplayContext from: baselineStartPoint to:
 	self displayLineGlyphOn: aDisplayContext from: s to: e
 ]
 
+{ #category : #emphasis }
+FreeTypeFont >> emphasized: code [
+	"empty stub, see https://github.com/pharo-project/pharo/issues/7550"
+]
+
 { #category : #accessing }
 FreeTypeFont >> face [
 	"Validate, and answer, the receiver's face"


Ducasse:
Hi torsten

Can you check because your change looks related to freetype and not to rub?

astares:
@Ducasse Yes - the title is misleading. Because this is what I first thought and got (and thats why the bug is opened with this title). 


But if you check what is described in the bug the DNU is a problem of a missing implementation in FreeType.

One only gets this problem when using a FreeType font (see script in the original bug which downloads and uses a font from Google font collection). This way it should be reproducible.

Ducasse:
Ok thanks for checking. 
What I know is that Rubric decorators use DNU to work (this way they are not forced to duplicate the API and forward to the inner decorator - I do not really like it) but this is why I was also waiting.

PR #7565, opened 2020-10-23
Cleanup FreeTypeCacheLinkedList

### LINKED ISSUE #7564: Cleanup FreeTypeCacheLinkedList
- use ifNil:ifNotNil: instead of  == nil checks with ifTrue:ifFalse:
- use ifNotNil: instead of == nil ifFalse:
- remove unnecessary dots

astares:
fix #7564

### DIFF
--- a/src/FreeType/FreeTypeCacheLinkedList.class.st
+++ b/src/FreeType/FreeTypeCacheLinkedList.class.st
@@ -12,28 +12,28 @@ FreeTypeCacheLinkedList >> add: link after: otherLink [
 	"Add otherLink  after link in the list. Answer aLink."
 
 	| savedLink |
-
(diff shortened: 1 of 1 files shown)


Ducasse:
Torsten we should pay attention with the == nil -> ifNil because == is not preemptable while ifNil is (this is bad but this is like that) so doing what we think is a refactoring without impact may change the program execution in presence of concurrency.
So I think that you changes are ok but we should pay attention to this.

MarcusDenker:
ifNil: is fine, it is compiled to the same bytecode as "== nil ifTrue:". The problem is #isNil, as it introduces a message send where if it replaces "== nil".

I have in my TODO to add a note for that to the transformation rules.

MarcusDenker (approved):

PR #7579, opened 2020-10-23
Cleanup FluidClassBuilder-Tests package

### LINKED ISSUE #7578: Cleanup FluidClassBuilder-Tests package
- add missing comment to TViewModelMock
- tag TViewModelMock into "Mocks", while putting the test classes into tag "Base"
- put #builder into accessing
- put the two removeFromSystem methods into "private - cleanup"
- rename #removeFromSystem into #removeTestArtifactsFromSystem (as a better name as the test is not removed when calling with self) to have a more intention revealing selector

astares:
Fix #7578


Ducasse:
Thanks! I should do another pass on the fluid class (and remove this builder but should sit with Pablo).

PR #7594, opened 2020-10-27
7591-We-have-isImmediateClass-add-isImmediateObject

### LINKED ISSUE #7591: We have isImmediateClass, add isImmediateObject
We have isImmediateClass, so every time we check if an *object* is immediate, we do "someObject class isImmediateClass".

During the bootstrap, there is some situation where the class is not yet correct and that test returns false.

I would like to add isImmediateObject. By default it just returns "self  class isImmediateClass", but for SmallInteger and SmallFloat it would have an implementation on those classes returning true.

MarcusDenker:
- add #isImmediateObject
- add test
- use in #isReadOnlyObject and #setIsReadOnlyObject:

this will alow us to (in a second step) simplify #isReadOnlyLiteral, as with this we can just call #beReadOnlyLiteral on all literals

fixes #7591

### DIFF
--- a/src/Kernel-Tests/ProtoObjectTest.class.st
+++ b/src/Kernel-Tests/ProtoObjectTest.class.st
@@ -171,6 +171,15 @@ ProtoObjectTest >> testIfNotNilIfNil [
 ]
 
 { #category : #'tests - testing' }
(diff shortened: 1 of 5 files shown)


dionisiydk (approved):
It seems a Character is also immediate object

PR #7597, opened 2020-10-27
Variables-use-pushLiteral

MarcusDenker:
for both MetaLinks and Slots we added code in the past to not use pushLiteral: directly.

The idea was that this would be needed for read-only litereals.

But the realiszaton of read-only literals is now that only literals are compiled read-only, we can use the literal frame 
to e.g. refernce Variables or MetaLinks without any problem.


dionisiydk (approved):

dionisiydk:
Tests look unrelated

PR #7605, opened 2020-10-28
FontFamilyAbstract should be renamed to AbstractFontFamily

### LINKED ISSUE #7604: FontFamilyAbstract should be renamed to AbstractFontFamily
- rename FontFamilyAbstract into AbstractFontFamily  
- implement #isAbstract to return true
- provide the old class name as deprecated class in Deprecated90 package
- tag class properly
- adopt subclasses like FreeTypeFontFamily in package FreeType to subclass AbstractFontFamily instead of FontFamilyAbstract

- start to add some tests

more to come in other cleanups

astares:
Fix #7604

### DIFF
--- a/src/Deprecated90/FontFamilyAbstract.class.st
+++ b/src/Deprecated90/FontFamilyAbstract.class.st
@@ -1,0 +1,14 @@
+"
+This class is deprecated - use class `AbstractFontFamily`
+"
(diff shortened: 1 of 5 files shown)


Ducasse:
Yes a part of the system that we never looked at. It really needs love.

PR #7607, opened 2020-10-28
isReadOnlyLiteral-cleanup

MarcusDenker:
- all Literals are now compiled as readOnly
- we can now call beReadOnlyObject on Immediate objects (it's a nop in this case)

This allows us to simplify the whole story and remove #isReadOnlyLiteral, this was now useful to be able to do this step-by-step,
but it is not neded anymore: Literals in the literal frame are now read-only.


MarcusDenker:
Due to the bootstrap, we need to do it in two steps

PR #7621, opened 2020-10-30
Deprecated withSqueakLineEndings. Fixed #7446

### LINKED ISSUE #7446: we should deprecated withSqueakLineEndings into withInternalLineEndings
withSqueakLineEndings should not be used anymore and it is not used but it is not using the automatic rewriting.

olekscode:


tesonep:
Why not moving it to the deprecated package as an extension method?

Ducasse:
Pablo we will do it later.

PR #7633, opened 2020-10-30
7585-Fix-dependency-to-get-a-green-bar-again

### LINKED ISSUE #7585: Fix dependency to get a green bar again
I should check the test of the kernel test (classPrinterDefinition because it may introduce bad dependency).

tesonep:
Fix #7585 

- Reorganizing the dependencies of the ClassParser tests.
- Adding some example class to not use Text, as it introduce an undesired dependency


Ducasse:
Pablo this is strange because testPoolUsers – Unix64.Kernel.Tests.Classes.SharedPoolTest is broken

Ducasse:
Argh... I continued to fix and added other tests :(.
What is the best way to handle the conflict?

Ducasse:
I removed the conflicting zone and I will check them in another PR

PR #7635, opened 2020-10-30
Fixed #7627. Changing canonicalClassName of True and False to Boolean

### LINKED ISSUE #7627: When creating a method with the debugger receiving a Boolean argument, aBoolean should be the name of the argument
**Description**

When having a Do-not-understood message is possible to create a new method in the debugger. If the method receives a boolean as an argument the name of the argument is aTrue or aFalse, it would be better if it is aBoolean. 

**Expected development cost**
Easy

**Version information:**
 - Version: Pharo 9
- OS: all operating systems

carolahp:


tinchodias:
I loaded it and fixes #7627 perfectly. +2 green tests

PR #7641, opened 2020-10-30
6731 use announcements instead of ad hoc observer in breakpoints

### LINKED ISSUE #6731: Use announcements instead of ad-hoc observer in Breakpoints
The implemented observer in Breakpoint was meant to be trivial, but now it reached a point where any client could be notified of breakpoints events.
The breakpoint notifications should be transformed to announcements and the mechanics modified accordingly.

dupriezt:
Fixes #6731

- Typo correction (registerInsterestToSystemAnnouncement -> registerInterestToSystemAnnouncement)
- Removed unused BreakpointActivationNotification class
- Breakpoint now uses SystemAnnouncer instead of an ad-hoc announcement system.
- BreakpointNotification and its subclasses are deprecated, to be replaced with BreakpointAnnouncement and its subclasses (which are copies)
- Reflected methods of the MetalinkChanged system announcement onto the BreakpointAnnoucement (and subclasses) system announcement.
- Tests of breakpoint notifications updated to use SystemAnnouncer instead of the old ad-hoc observer pattern of the Breakpoint class.


src/Reflectivity/BreakpointHit.class.st
```diff
@@ -0,0 +1,18 @@
+Class {
+	#name : #BreakpointHit,
+	#superclass : #BreakpointAnnouncement,
+	#instVars : [
+		'valueOrNil'
```
dionisiydk, line 5: Can it be a meaningful name? What is stored inside?

dionisiydk (approved):
It's a good change. 
My comment is for existing code which is refactored here but the question is still relevant

MarcusDenker:
tests fail with 

Breakpoint class(Object)>>doesNotUnderstand: #observers

dupriezt:
All tests of BreakpointTest were failing because setUp and tearDown tried to mute the observer pattern of the Breakpoint class during the tests (to avoid sending breakpoint notifications to tools, for breakpoint events happening because of the tests).
I replaced this mute code with code that mutes the SystemAnnouncer (that now handles breakpoint notifications thanks to earlier commits in this PR).
However, muting the SystemAnnouncer means the Breakpoint class is no longer receiving notifications like MethodChanged during the breakpoint tests. Some tests (like testModifyMethodWithBreakpoint) test that the Breakpoint class removes the breakpoints of methods that get modified, so these tests fail. These tests simultaneously need SystemAnnouncer to not send out breakpoint notifications to the rest of the system, while still sending notifications like MethodChanged to the Breakpoint class. To address this, I changed the setUp so that the Breakpoint class registers to the placeholder SystemAnnouncer used during the breakpoint tests.

These two changes fix all the tests but these two (full code below):
1. BreakpointTest>>#testTargetInstanceGarbageCollected
2. SystemDependenciesTest>>#testExternalSUnitKernelDependencies

I don't see the link between this PR and test 2.
As for test 1, it fails because the targetInstance of a breakpoint does not get garbage collected as it should. I don't see how this PR could have affected garbage collection in that way...

````Smalltalk
testTargetInstanceGarbageCollected
	|bp object|	
	bp := Breakpoint new.
	object := Object new.
	bp scopeTo: object.
	object := nil.
	Smalltalk garbageCollect.
	
	self assert: bp targetInstance isNil.
	self deny: bp isObjectCentric
```

```Smalltalk
testExternalSUnitKernelDependencies

	| dependencies | 
	
	self longTestCase.
	
	dependencies := self externalDependendiesOf: (
		BaselineOfPharoBootstrap compilerPackageNames,
		BaselineOfTraits corePackages,
		BaselineOfPharoBootstrap kernelPackageNames,
		BaselineOfPharoBootstrap multilingualPackageNames,
		BaselineOfPharoBootstrap kernelAdditionalPackagesNames,
		BaselineOfPharoBootstrap sUnitPackageNames).
	
	self assertCollection: dependencies hasSameElements: self knownSUnitKernelDependencies.
```

src/Reflectivity/BreakpointHit.class.st
dupriezt: I'm guessing this variable is meant for the subclass VariableBreakpoint, to store the value the variable has at the moment its breakpoint triggers.

MarcusDenker:
I think testTargetInstanceGarbageCollected fails as the class definition of Breakpoint loses the WeakSlot, it is saved as a standard class definition even though it is using a slot.

(you an see that if you look at the diff here on GitHub)

dupriezt:
> 
> 
> I think testTargetInstanceGarbageCollected fails as the class definition of Breakpoint loses the WeakSlot, it is saved as a standard class definition even though it is using a slot.
> 
> (you an see that if you look at the diff here on GitHub)

Ah, indeed. I changed it back (by checking the "Fluid" check box in calypso and editing the class definition) and now testTargetInstanceGarbageCollected passes again.
However, Iceberg doesn't see this change to the class definition, so it does not let me commit anything ("There is nothing to commit")

MarcusDenker:
can you try to turn on the "fluid class definition" in the browser?

dupriezt:
I did:
1. 
image
2. Checked "Fluid" checkbox
image
3. Added "WeakSlot"
image
4. Unchecked the "Fluid" checkbox

But iceberg doesn't see any changes to commit.
image

dupriezt:
I could modify the text file Breakpoint.class.st by hand to add WeakSlot and use command line git to push (without iceberg), unless there's a way to do it from the image.

MarcusDenker:
modify by hand might be the best, we need to check what the bug is indendently

dupriezt:
> modify by hand might be the best

Done

MarcusDenker (approved):

StevenCostiou:
Hi guys, this is nice work. @MarcusDenker @dionisiydk @dupriezt I'm taking the opportunity here to remark things that could be fixed easily without opening a new issue. I would like to know what you think about it:

- We should simply remove the obsolete classes, they were never supposed to be used outside the register/unregister observer API in Breakpoint class, it would have been an improper use. This is now dead code.

- I would like to keep the register/unregister methods: they are an API to subscribe to breakpoints announcement rather than calling directly SystemAnnouncer. Note that it seems bad practice to merge methods with a "do not use" comment. 

- Could we use a local observer instead of the system announcer? What would be the pros/cons? Is it better practice to always use the system announcer?

- Should we not use the announcer **weak**?

dionisiydk:
In addition to these questions:
There are objects similar to Breakpoint: ExecutionCounter and Watcher. IMHO they all implement very similar management code which means there is a bunch of duplicated code: 
- how to register/unregister instances
- how to organize registry
- how to announce registration

Currently ExecutionCounter and Watcher work using MetaLink announcement logic (optionAnnounce: true). Why Breakpoint is different?
Can we refactor code to unify them (with common superclass for example) ?

It's also very clear to me that advanced breakpoint features (condition and instance specificity) are applicable to counter and watcher. But current design does not allow it.

StevenCostiou:
Breakpoints do use optionAnnounce, but that only notifies that a metalink has changed (if I correctly recall).
We had the need to announce more precise information about the breakpoints (un)installation and hit.

I do not know how that can be generalized to watch and counter, but there could be a common abstract logic (the registering, etc.) while the concrete implementation should decide what to announce and when.

True that counters and others should benefit from conditions and instance scopes. I opened an issue or a task in a github project somewhere about that but I never had the time.

MarcusDenker:
Right now Metalinks announce changes double if optionAnnounce is set to true.

MarcusDenker:
We another problem with MetalinkChanged: speed. Norbert was trying to use MetaLink to do code coverage analysis, and with how things are, it is completely unusable as soon as you set many links: every link setting triggers UI update, which is very very slow to the point that is not usable.

Now that we have announcements for breakpoints, we can use those to update the UI.

As a first step I want to make the announcements of MetaLinks optional again by default. Then we can see in second step what the best direction is.

noha:
Marcus is right. I used it on ~500 methods and it took nearly 3 minutes to complete. The use case of UI updates is just a single one which  operates on a single node. And I don't understand what the name MetalinkChanged means if it is fired when you add a link to/remove from a RBNode. And what are the announcements good for if I add it to 500 nodes and get 500 announcements. Seems a bit misaligned to me.

noha:
Furthermore I do not know the reasons why the SystemAnnouncer is used for this. What is the rationale behind that? To update all browsers?

StevenCostiou:
> Furthermore I do not know the reasons why the SystemAnnouncer is used for this. What is the rationale behind that? To update all browsers?

I think I understand why the system announcer was chosen, in the other hand breakpoints announcements are not meant to be notified to the system. At the beggining it was for meta-debugging tools, and it was only on-demand by registering to the Breakpoint class' API.

The best would be to have a local announcer (that we use *weak*) in the Breakpoint class. Maybe a generalization could later be defined and implemented to have that available for all metalink tools.

dionisiydk:
> Marcus is right. I used it on ~500 methods and it took nearly 3 minutes to complete. The use case of UI updates is just a single one which operates on a single node. And I don't understand what the name MetalinkChanged means if it is fired when you add a link to/remove from a RBNode. And what are the announcements good for if I add it to 500 nodes and get 500 announcements. Seems a bit misaligned to me.

To clarify: I did not suggest to always announce any metalink changes. The optionAnnounce was very good addition. 
My point is that ExecutionCounter and Watcher needs same functionality as Breakpoint or otherwise. Do you agree with that?

With changes in this PR breakpoint events are published by SystemAnnouncer which returns them back to how MetalinkChanged works - also through SystemAnnouncer. 
And it makes me think about general solution by introducing MetalinkAdded, MetalinkRemoved and MetalinkHit events and using optionAnnounce for breakpoints. So that all such breakpoint like objects will behave equally.  

Notice that we have a conditional subscriptions. So users can be notified specifically for breakpoints: 
```Smalltalk
SystemAnnouncer when: (MetalinkAdded where: [:ann | ann object isKindOf: Breakpoint ]) do: #logCr. 
``` 
It can be wrapped in more friendly code for our use case.

What users of breakpoint events do we have now? The browser needs them to show corresponding icons over methods and AST nodes in editor.. What else?

StevenCostiou:
> To clarify: I did not suggest to always announce any metalink changes. The optionAnnounce was very good addition.
> My point is that ExecutionCounter and Watcher needs same functionality as Breakpoint or otherwise. Do you agree with that?

Yes. But generalizing too much might lead to constraints that  are unnecessary. For instance, it has no importance that I am notified that a metalink was added when in reality the information I want is "a breakpoint was added on that variable". And receiving both is a waste of resource if I'm explicitly interested in only one of them.

> With changes in this PR breakpoint events are published by SystemAnnouncer which returns them back to how MetalinkChanged works - also through SystemAnnouncer.
> And it makes me think about general solution by introducing MetalinkAdded, MetalinkRemoved and MetalinkHit events and using optionAnnounce for breakpoints. So that all such breakpoint like objects will behave equally.

I believe this would be good, as in `MetaLinkChanged` we currently have a lot of stuff to check which case we are in (add, remove, etc.).

> Notice that we have a conditional subscriptions. So users can be notified specifically for breakpoints:
> 
> ```smalltalk
> SystemAnnouncer when: (MetalinkAdded where: [:ann | ann object isKindOf: Breakpoint ]) do: #logCr. 
> ```
> 
> It can be wrapped in more friendly code for our use case.

That wrapping is done in the breakpoint announcements. Whatever additional information they hold is encapsulated in those announcements. Without them, we lose a lot of precision.
Same argument as above, we need a *breakpoint added* announcement and not a metalink one for which we have to check the encapsulated object.

However, they should not be announced by the system but by something more specific. I don't believe that the system really cares that we hit or added a breakpoint... what are the cases? 

> What users of breakpoint events do we have now? The browser needs them to show corresponding icons over methods and AST nodes in editor.. What else?

In the system, I think none currently. We use those in other debugging projects or for monitoring breakpoints in debugging studies. 
It was more appropriate to add that directly in the image, so that it is an open facility for everyone to use.

dionisiydk:
> > Furthermore I do not know the reasons why the SystemAnnouncer is used for this. What is the rationale behind that? To update all browsers?
> 
> I think I understand why the system announcer was chosen, in the other hand breakpoints announcements are not meant to be notified to the system. At the beggining it was for meta-debugging tools, and it was only on-demand by registering to the Breakpoint class' API.
> 
> The best would be to have a local announcer (that we use _weak_) in the Breakpoint class. Maybe a generalization could later be defined and implemented to have that available for all metalink tools.

MetalinkChanged is actually announced when breakpoint is installed (and by some reason twice, a bug?). It triggers a browser update.
Now there is an extra BreakpointAdded event. 
So I see what you mean by local announcer. But I wonder why you are not happy with system one? 

SystemAnnouncer is supposed to publish changes about any kind of system modification. Should we consider a breakpoint here? I think so because It affects a system behavior (it halts execution somewhere).
A possible use case is logging breakpoints in Epicea. The crashing debug session can be restored after restart.

dionisiydk:
> MetalinkChanged is actually announced when breakpoint is installed (and by some reason twice, a bug?). It triggers a browser update.

For the record:
- MetalinkChanged is per method event (but double event is probably a big)
- BreakpointAdded is for full breakpoint setup. Variable breakpoint for example will trigger multiple MetalinkChanged but the only BreakpointAdded

dionisiydk:
> Yes. But generalizing too much might lead to constraints that are unnecessary. For instance, it has no importance that I am notified that a metalink was added when in reality the information I want is "a breakpoint was added on that variable". And receiving both is a waste of resource if I'm explicitly interested in only one of them.

Yes, agree. It's what I commented after.

dionisiydk:
> However, they should not be announced by the system but by something more specific. I don't believe that the system really cares that we hit or added a breakpoint... what are the cases?

Tools can rely on these events instead of MetalinkChanged which may improve performance when breakpoint affect a lot of places. 
But the current behaviour does not need it. So it can be safely replaced by local announcer.

noha:
My opinion is that a change towards explicit metalink action like added, removed is good because control can be more fine grained. But I also think that we overuse the SystemAnnouncer. If we add MetalinkChanged to the system announcer to announce the addition of a metalink in order to update a UI it gets a bit awkward and brittle. I think in the future those need be separated by topics. Tools can have multiple subscriptions. And as we can see here is that in sequence executed announcements can slow down the whole system

MarcusDenker:
What I did:

- remove the obsolete classes
- correctly raise BreakPoint announcements when installing / removing BreakPoint and VariableBreakPoint
- change the links to not raise announcements by default (for now). Do that per link install, not per method.
- make #optionStyler false by default, too.

See https://github.com/pharo-project/pharo/pull/7672 and https://github.com/pharo-project/pharo/pull/7676

I think we need to carefully distinguish levels: MetaLinks are so low level that we can not afford to update the UI.

With the two PRs, all the updating is done with Breakpoint announcements (just for Watcher it still uses MetaLink announce
to update, to be fixed). Setting metalink is now fast again.

TODO:

- we should have a hierarchy for MetaLinkAnnouncer if we keep it (changed, added, removed) and hook it up to when the link is changed

- Steven: Could we use a local observer instead of the system announcer? What would be the pros/cons? Is it better practice to always use the system announcer? ==> if we keep the MetaLink announcer, yes, that is what I would do

- We need a Special Announcement for Watcher to update the UI

- do we need BreakPointHit announcement?

And these from Steven:

- I would like to keep the register/unregister methods: they are an API to subscribe to breakpoints announcement rather than calling directly SystemAnnouncer. Note that it seems bad practice to merge methods with a "do not use" comment.

- Should we not use the announcer weak?

PR #7644, opened 2020-10-31
Rename #simulated into #isSimulatedStyle

### LINKED ISSUE #7643: Rename #simulated into #isSimulatedStyle
- for #simulated and #simulated: it is better to use #isSimulatedStyle and #beSimulatedStyle
- rename the iVar simulated into isSimulatedStyle also
- introduce proper deprecations with transform rules for #7636 simulated
- initialize with false, then we can avoid to set it with false
- point people to beSimulatedStyle in #simulated: deprecation

astares:
Fix #7643


Ducasse:
This is good that you start to add tests because this part is really in need for love. 
I started to migrate the fontChooser to Spec2 but I stopped in the middle.

PR #7656, opened 2020-11-02
7653-Removal-of-SystemReporter-breaks-smalltalkCI

### LINKED ISSUE #7653: Removal of SystemReporter breaks smalltalkCI
**Describe the bug**
SystemReporter was removed from the system (https://github.com/pharo-project/pharo/pull/7637) but it is used by SmalltalkCI that is now unusable.
```bash
NewUndeclaredWarning: SmalltalkCIPharo>>imageInfo (SystemReporter is Undeclared)

NewUndeclaredWarning: SmalltalkCIPharo>>imageInfo (SystemReporter is Undeclared)

VariableNotDeclared: receiver of "new" is nil
```

**To Reproduce**
Run a build with Pharo 9 and Smalltalk CI

Ducasse:
Fixes: #7653 Just create a subclass to support backward compatibility for SmalltalkCI

### DIFF
--- a/src/NewTools-SystemReporter/SystemReporter.class.st
+++ b/src/NewTools-SystemReporter/SystemReporter.class.st
@@ -1,0 +1,9 @@
+"
+I'm just a placeholder to support migration to never version. 
+In particular SmalltalkCI should not depend on user interface class. 
+"
+Class {
+	#name : #SystemReporter,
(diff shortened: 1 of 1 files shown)


Ducasse:
This is will be changed the PR of Christophe that extracts a low level object from StStystemReporter. 
Right now I integrate it and we will see.

PR #7657, opened 2020-11-02
7584-Fluid-Complex-slot-should-force-display-of-fluid-class-definition

Ducasse:
Make sure that we select the correct parser given a class and a given syntax.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -172,13 +172,14 @@ def bootstrapImage(){
     }
     
       } finally {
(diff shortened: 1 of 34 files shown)


Ducasse:
Yes the bootstrap is broken probably because a class with complex slot was in the past sneeking in unnoticed.

Ducasse:
I will have to "sit with Pablo or Guille to understand"

Ducasse:
This is nice I read all your PRs and I could not have guessed. And this is good to improve the Jenkins to give us a log.

PR #7661, opened 2020-11-02
7347 improve exception description api

### LINKED ISSUE #7347: Improve Exception description API
Following https://github.com/pharo-project/pharo/pull/7335 integration.
`Exception` implements a `description` method and a `smartDescription` method that returns more detailed information.
Instead a better naming should be chosen:
- use `description`to provide the *smart description*
- use `defaultDescrption`when there is no *smart description* implmented

maxwills:
Closes issue: #7347 
Simplified API: removed smartDescription.
Exception >> # description is the "default description" and handles the creation of the description string. It is overridden in subclasses for specialized descriptions.
Subclasses' (MessageNotUnderstood and VariableNotDeclared) # description method returns a specialized description string when their 'message' field is not Nil. Otherwise, they use the base class default description.

Additional methods and tests included related to changes.

#7347


Ducasse:
For me this is ok but better have more eyes.

maxwills:
I can't link this Pull Request to make it appear as if it Closes the Issue. if anyone knows what's missing, please let me know.

MarcusDenker:
very strange, I could not add the issue from here as a linked issue, but I could add the PR from the issue, very strange

src/Kernel/VariableNotDeclared.class.st
```diff
@@ -16,11 +16,18 @@ VariableNotDeclared >> classSymbol [
...
-	message ifNil: [^self description].
+	message ifNil: [^super description].
 
-	^self classSymbol printString
+	^self messageBasedDescription
+]
+
+{ #category : #accessing }
+VariableNotDeclared >> messageBasedDescription [
+	"Returns a textual representation of the exception based on the message field"
```
StevenCostiou: The method is not returning anything here while the comment states it does

src/Kernel-Tests/ExceptionTest.class.st
```diff
@@ -52,6 +52,13 @@ ExceptionTest >> runWithNoHandlers: aBlock [
 	^result
 ]
 
+{ #category : #test }
+ExceptionTest >> testDescription [
+
+	self flag: #toImplement.
+	self assert: false
```
StevenCostiou: This test is unimplemented, this hsould be fixed

src/Kernel-Tests/VariableNotDeclaredTest.class.st
```diff
@@ -0,0 +1,22 @@
...
+   "Tests that the description method produces a message-based text description of the exception, if the message field is not nil, or it uses the default description text instead if message nil"
+
+	| aVariableNotDeclared defaultDescription descriptionWithMessage |
+	aVariableNotDeclared := MyVariableNotDeclaredDummy new.
+	"The default description when message is nil"
+	defaultDescription := aVariableNotDeclared description.
+	aVariableNotDeclared message: 'aNonNilValue'.
+		"The message-based description. It is a constant in our dummy class overriden accesor."
+	descriptionWithMessage := aVariableNotDeclared description.
+	self assert: defaultDescription ~~ descriptionWithMessage
```
StevenCostiou: Maybe use a `deny:equals`rather than `~~`?

src/Kernel-Tests/MyMessageNotUndestoodDummy.class.st
```diff
@@ -0,0 +1,17 @@
...
+	#name : #MyMessageNotUndestoodDummy,
+	#superclass : #MessageNotUnderstood,
+	#category : #'Kernel-Tests-Exceptions'
+}
+
+{ #category : #accessing }
+MyMessageNotUndestoodDummy >> messageBasedDescription [ 
+	"Overrides the method for testing"	
+	^ 'messageBasedDescription'
+]
```
StevenCostiou: I would prefer that we directly instantiate and fill a real `MessageNotUnderstood` to test it rather than creating a dummy class. It's fine to do so but here I think we can easily create a MNU object and work with it to test its description behavior.

src/Kernel-Tests/MyVariableNotDeclaredDummy.class.st
```diff
@@ -0,0 +1,16 @@
...
+	#name : #MyVariableNotDeclaredDummy,
+	#superclass : #VariableNotDeclared,
+	#category : #'Kernel-Tests-Exceptions'
+}
+
+{ #category : #accessing }
+MyVariableNotDeclaredDummy >> messageBasedDescription [ 
+	"Overrides the method for testing"	
+	^ 'messageBasedDescription'
+]
```
StevenCostiou: Same remark as for the MNU above

StevenCostiou (changes requested):
Instead of introducing a new selector `messageBasedDescription`, could we rather introduce a `defaultDescription` at the highest level (in `Exception`?) and having by default `description` in `Exception` calling that default description message, while subclasses of `Exception` can redefine `description`? (therefore replacing `messageBasedDescription` from that PR).

Ducasse:
Hello can we move on this issue because it looks quite simple from the outside so I would like to close it by integrating the changes so can one of you fix it.

src/Kernel/VariableNotDeclared.class.st
maxwills: The method was removed based on your remarks

src/Kernel-Tests/ExceptionTest.class.st
maxwills: There was no test for that method. When I tried to jump to the related test that code was automatically added. After your remark, I implemented a test, that, in my opinion, doesn't add value, but "its now implemented". Should I just remove it instead?

Please review if fixed.

src/Kernel-Tests/MyMessageNotUndestoodDummy.class.st
maxwills: I tried to instantiate. However, the reason for the dummy class is that there are some required fields, for the production of the description, that I am not capable of manually creating them. Overriding that method was a quick fix.
For this class "MessageNotUnderstood" I used a real MessageNotUnderstood exception for the test in the latest push.

src/Kernel-Tests/MyVariableNotDeclaredDummy.class.st
maxwills: The reason of the dummy was the same as the previous. Unfortunately, for this case, I was not able to generate a working "VariableNotDeclared" exception. Please take a look if the solution is acceptable.

StevenCostiou (approved):
Thanks Max

PR #7669, opened 2020-11-04
Simplify-MetaLinkInstaller

MarcusDenker:
This PR simplifies MetaLinkInstaller by taking advantage of the new unified Variable hierarchy.

Instead of using names, we get the Variable object early and hand that over to the installer.

This alllows us to unify the code there for all Variables to be just #installPermaLink:onVariable:

PermaLink does not need to keep the method, as that information is already accessible from the Temporay Variable object 
(it was not used anyway)

### DIFF
--- a/src/Reflectivity/MetaLink.class.st
+++ b/src/Reflectivity/MetaLink.class.st
@@ -263,26 +263,33 @@ MetaLink >> installOn: aNode [
 
 { #category : #'installing - link installer' }
 MetaLink >> installOnClassVarNamed: aClassVarName for: aClassOrObject option: option instanceSpecific: instanceSpecific [
-	| permalink |
+	| permalink variable |
+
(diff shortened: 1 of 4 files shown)


src/Reflectivity/MetaLink.class.st
```diff
@@ -263,26 +263,33 @@ MetaLink >> installOn: aNode [
 
 { #category : #'installing - link installer' }
 MetaLink >> installOnClassVarNamed: aClassVarName for: aClassOrObject option: option instanceSpecific: instanceSpecific [
```
dionisiydk, line 265: I think this API methods can also be simplified. It can be a user responsibility to choose a variable instance and pass it to install method.

dionisiydk (approved):

src/Reflectivity/MetaLink.class.st
MarcusDenker: yes, we should see how to simplify things even further.

PR #7670, opened 2020-11-04
identityIncludesFontProvider: should be test extension

### LINKED ISSUE #7668: identityIncludesFontProvider: should be test extension
the method #identityIncludesFontProvider: is only used by a test and not necessary for the API, so lets 
package it as an extension of class LogicalFontManager into package "Fonts-Infrastructure-Tests"

astares:
Fix #7668


MarcusDenker (approved):

Ducasse:
Now I do not like that we have test doing extension. If we need a testing method in the class it is better to add it.

PR #7675, opened 2020-11-05
Speedup-containsHalt

MarcusDenker:
Speed up #containsHalt by tagging methods with halts in Semantic Analysis.

As halts are not very common (and should never be commited), it is ok to pay for the overhead in terms of memory
(we pay only for methods containing a halt).

Slow down of compile time in minimal (just a fast check while already visiting the AST).

Next steps:
- this now can repalce #hasHalts use in Newtools
- maybe we could treat the #haltOrBreakpointForTesting  at this level, not sure.

### DIFF
--- a/src/Kernel/CompiledCode.class.st
+++ b/src/Kernel/CompiledCode.class.st
@@ -183,12 +183,6 @@ CompiledCode >> comment [
 	^ self ast firstPrecodeComment
 ]
 
-{ #category : #testing }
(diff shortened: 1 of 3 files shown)


MarcusDenker:
Can not be merged as it needs a move of isHaltNode out of Newtools which is a cross cutting change that I might have the energy to do in 3 month or so.

PR #7676, opened 2020-11-05
MetaLink-DoNotAnnounceByDefault

MarcusDenker:
This PR does two things:

- make sure we do not announce MetaLink changes double
- all announcing is now governed by the #optionAnnounce which is false by default
- make #optionStyler false by default, too.


noha (approved):

MarcusDenker:
AARGH... this, too, needs changes in *BOTH* Pharo and NewTools at the same time.

MarcusDenker:
Ah, good, it was another problem: tests are ok now

PR #7689, opened 2020-11-07
7687-Cannot-remove-a-Trait-written-with-the-fluid-syntax

### LINKED ISSUE #7687: Cannot remove a Trait written with the fluid syntax
Because RB is trying to parse... and we should update the definition patterns defined in definitionPatterns methods.

Ducasse:
Fixes: #7687 but I do not understand why we can remove a class (while we could remove a trait).


Ducasse:
Broken test is unrelated: ZnClient on windows

Issue #7699, opened 2020-11-09
The Debugger stack view is missing package column


pavel-krivanek:
**Describe the bug**
The new debugger stack view is missing package column that was added this year in this commit to the GTDebugger:
https://github.com/pharo-project/pharo/commit/d854d60ee6c3500863d96f177ac81dbf41124d2e

estebanlm:
sorry, but... why do we want package column in debugger?

pavel-krivanek:
Because then you may see where the stack may not be important for you... See the typical usage of moldable debugger when you use debugger plugins to group/hide announcements processing logic. If you see the package in stack, you clearly see where the flow returned to your code you are interesting in.

PR #7700, opened 2020-11-09
Return the package column to the debugger

pavel-krivanek:
issue #7699


StevenCostiou:
I have sync newtools with that change, otherwise it would have been overwritten in the next merge (unfortunately the way we work now, NewTools repo is the ref).

PR #7701, opened 2020-11-09
Signalling a VariableNotDeclared whenever we attempt to read or to write an undeclared variable

### LINKED ISSUE #7083: After removing an instance variable from a class, code using this instance variable still *works*
In any class, create this method:
```Smalltalk
someMethod
  i := i + 1.
  ^i squared
```
Then save, and as `i` does not exist, create an instance variable.
Now go into the class definition, remove `i` and execute in the playground:

`YourClass new someMethod`

It *works*! Or at least it does *not not work*.
If you `debugIt` and step into, it lets you step over the assignment and the return of the undeclared variable without problem (values are clearly wrong if you look at `stackTop`).

Happens in both `GTDebugger` and `StDebugger`, from both `GTPlayground` and `StPlayground`.

StevenCostiou:
Fixes https://github.com/pharo-project/pharo/issues/7083

Instead of transparently writing and reading undeclared variables, we signal a variable not declared exception. The idea is to give a chance to the debugger to catch it and propose the user to define it.

If we do not do that, the code continues its execution and it was (and can still be) the cause of bugs super hard to understand.

Now, this is a suggestion and works remains:
- how to test that code?
- what's the impact?
- if merged, the debugger will not be able to know what to define (a class? a variable?) and maybe it requires a simple heuristic to do that (or simply ask the user)

### DIFF
--- a/src/Kernel/UndeclaredVariable.class.st
+++ b/src/Kernel/UndeclaredVariable.class.st
@@ -37,11 +37,22 @@ UndeclaredVariable >> definingClass [
 { #category : #'code generation' }
 UndeclaredVariable >> emitStore: methodBuilder [
 
+	methodBuilder pushLiteralVariable: VariableNotDeclared binding.
+	methodBuilder pushLiteralVariable: self.
+	methodBuilder send: #signal:.
+	methodBuilder popTop.	
+	
 	methodBuilder storeIntoLiteralVariable: self
 ]
 
 { #category : #'code generation' }
 UndeclaredVariable >> emitValue: methodBuilder [
+	
+	methodBuilder pushLiteralVariable: VariableNotDeclared binding.
+	methodBuilder pushLiteralVariable: self.
+	methodBuilder send: #signal:.
+	methodBuilder popTop.
+	
 	methodBuilder pushLiteralVariable: self
 ]
 


StevenCostiou:
It seems it actually crashes the CI because it raises a VariableNotDeclared.

dionisiydk:
it's great initiative @StevenCostiou .
The main issue is how undeclared variable becomes declared one. You need to recompile all user methods otherwise all of them will continue raise an error (the error logic becomes "hardcoded" when compiled).

Testing is really easy to do. I would point you to ForeighVariableTest as an example but by some reason it is not in the image (I will push a PR with tests). So here is my approach:
```Smalltalk
testReadUndeclared
	| temp var ast tempReadDoIt |
	temp := 100.
	var := UndeclaredVariable named: #temp.
	ast := [ temp + 2 ] sourceNode body asDoit doSemanticAnalysis.
	ast variableNodes first variable: var.
	tempReadDoIt := tempAST generateWithSource.
	
	self should: [tempReadDoIt valueWithReceiver: self arguments: #()] raise: VariableNotDeclared
```

MarcusDenker:
Another problem is that VariableNotDeclared is a subclass of MessageNotUnderstood (due to the hack that is used to do it now).

We should for now use our own exception.. it needs to fall back to returning "nil" when there is no UI.

dionisiydk:
> Another problem is that VariableNotDeclared is a subclass of MessageNotUnderstood (due to the hack that is used to do it now).
> 
> We should for now use our own exception.. 

I don't think it's a big issue. It was done like that just to allow "Define missing class" dialog in debugger similar to "Define missing method". As soon as we will enable this error by default it should be really straightforward to clean this feature from such trick. And then hierarchy can be fixed.

> it needs to fall back to returning "nil" when there is no UI.

I think it just needs a setting similar to "Deprecation warning" and "Shadow variables". It does not really depend on UI. On CI it's super useful to see such problems while now they are almost hidden (no errors usually).

Ducasse:
Ok guys let us know what we do.

StevenCostiou:
Hi Thanks Denis and Marcus, I want to continue this but I cannot until next monday.
@dionisiydk thanks for the unit test example, I will use it and push it with the next changes if you are ok

dionisiydk:
> Hi Thanks Denis and Marcus, I want to continue this but I cannot until next monday.
> @dionisiydk thanks for the unit test example, I will use it and push it with the next changes if you are ok

Sure. DoItVariableTests are now in the image.

Ducasse:
Hi guys

what should we do with this one?

S.

guillep:
I'm not sure. I see these are not extension methods. I'm afraid too much compiler code is creeping into the kernel.

guillep:
From @StevenCostiou 

> I am currently in holidays and will not return to work until the 4th of January, and unfortunately I do not know yet what will be my priorities, if some PR are waiting for my approval better just close them and I will reopen them later in the year if that seems appropriate

PR #7704, opened 2020-11-09
Unify-OCInstanceScope-Behavior

MarcusDenker:
The compiler was modeling a dedicated scope just for the self/super/thisContext vars

We can remove this OCInstanceScope and simplify things by
- moving the handling of self/super/thisContext to the global scope. This makes a lot of sense as there is
one "self".
- merge OCInstanceScope with the class scope (which is the class itself


src/System-Support/SystemDictionary.class.st
```diff
@@ -275,6 +278,13 @@ SystemDictionary >> hasClassOrTraitNamed: aString [
 	^ false
 ]
 
+{ #category : #'dictionary access' }
+SystemDictionary >> lookupVar: name [
+	"Return a var with this name.  Return nil if none found"
+	name isString ifFalse: [ ^nil ].
```
dionisiydk, line 284: I know it was there before but I would really remove this check. It should be an explicit error if type protection is really required.

src/System-Support/SystemDictionary.class.st
```diff
@@ -405,6 +415,11 @@ SystemDictionary >> renameClassNamed: oldName as: newName [
 	oldClass rename: newName
 ]
 
+{ #category : #removing }
+SystemDictionary >> reservedVariables [
+	^vars ifNil: [ vars := ReservedVariable lookupDictionary]
```
dionisiydk, line 420: what is the reason to not name it #reservedVariable ?

dionisiydk (approved):

MarcusDenker:
Build fails due to ReservedVariable not yet loaded (it is reading an Undeclared, that is nil):


UndefinedObject>>doesNotUnderstand: #lookupDictionary
SystemDictionary>>reservedVariables
SystemDictionary>>lookupVar:
HEExtendedInstaller class(Class)>>bindingOf:
HEGlobalVariable>>asLiteralIn:

src/System-Support/SystemDictionary.class.st
MarcusDenker: No reason, should be done

src/System-Support/SystemDictionary.class.st
MarcusDenker: yes, I added it already in the todo, but I wanted to do it independently of this change

src/System-Support/SystemDictionary.class.st
MarcusDenker: (I was thinking of using the term "PseudoVariable" instead of Reserved, but that is independed of this change)

src/OpalCompiler-Core/OCASTSemanticAnalyzer.class.st
```diff
@@ -204,7 +204,7 @@ OCASTSemanticAnalyzer >> visitInlinedBlockNode: aBlockNode [
 { #category : #visiting }
 OCASTSemanticAnalyzer >> visitMethodNode: aMethodNode [
 
-	scope := compilationContext scope newMethodScope. 
+	scope := OCMethodScope new outerScope: compilationContext scope.
```
dionisiydk, line 207: I think it should be extracted into CompilationContext>>#newMethodScope. And then maybe it can be merged with #scope method.

dionisiydk: And it's really just two users of "compilationContext scope".  So it can be transformed into single #newMethodScope method

src/System-Support/SystemDictionary.class.st
dionisiydk: Actually I also consider PseudoVariable name after my first changes on Variable hierarchy. But I can't recall what stopped me from using it. 
Now pseudo sounds better than reserved in the way what we model with it.

dionisiydk: But for the name of container variable I would choose the name #reservedVariables because it sounds more general: pseudo vars are reserved by default but we may have others non pseudo but reserved vars.

Having it written I am starting recall my thoughts about these names. What does pseudo actually mean? I think many classes in Variable hierarchy deserves this status preudo. Imagine very special slots. Pseudo feels like characteristics of very special implementation. From the other side it does not matter how variable is implemented to be reserved.

Maybe all of it is just a demagogy :)

src/OpalCompiler-Core/OCASTSemanticAnalyzer.class.st
MarcusDenker: Yes, I will merge this and do another PR with the small fixes

PR #7710, opened 2020-11-09
DoItVariable instead of ForeignVariable and tests 

dionisiydk:
ForeignVariable is renamed to DoItVariable as it better reflects an intention.
Also tests are added which were lost in my original commits


Ducasse:
Just a question the DoitVariable are only used in Doit? or for something else.

dionisiydk:
> Just a question the DoitVariable are only used in Doit? or for something else.

Yes. It's for doIt execution. 
It looks more general underhood but it's difficult to find a proper name. And having in mind that there is an only user I think DoItVariable is much clear to to understand an intention.

MarcusDenker (approved):
naming is always a problem.. for me either it ok. It is true that this now only used for DoIts.

PR #7714, opened 2020-11-10
Added <debuggerCompleteToSender> pragma to user-level object error-handling protocol methods

StevenCostiou:
Those methods, such as `shouldNotImplement` should not be shown in the debugger. They bring little value and worse, add mental noise.

The expected impact is that those methods will not show up in the system debugger anymore (customised debugger may choose to show them).

### DIFF
--- a/src/Kernel/Object.class.st
+++ b/src/Kernel/Object.class.st
@@ -815,15 +815,15 @@ Object >> enclosedElement [
 { #category : #'error handling' }
 Object >> error [
 	"Throw a generic Error exception."
-
-	^self error: 'Error!'.
+	<debuggerCompleteToSender>
+	^ self error: 'Error!'
 ]
 
 { #category : #'error handling' }
-Object >> error: aString [ 
+Object >> error: aString [
 	"Throw a generic Error exception."
-
-	^Error new signal: aString
+	<debuggerCompleteToSender>
+	^ Error new signal: aString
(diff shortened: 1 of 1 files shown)


Ducasse:
Steven did you see the remark of Martin McClure on discord about this?

Ducasse:
As a general principle I do not like to hide logic because then you cannot understand. 
For me seeing a subclassResponsibility is important. 
So if we hide it what do I get?

guillep:
I don't think this PR is ok, please see my comments in https://github.com/pharo-spec/NewTools/issues/124

StevenCostiou:
Let me show you: 

**Before**

Screenshot 2020-11-10 at 14 13 47

**After**

Screenshot 2020-11-10 at 14 14 01

Tell me which one is better, and we can either reject or merge the PR :)

guillep:
Ok, I did not understand right the impact of the change. Seeing the screenshots I understand better. I think it is ok, we can test it!

guillep:
@tesonep you opened the original issue, can you check?

StevenCostiou:
> Ok, I did not understand right the impact of the change. Seeing the screenshots I understand better. 

My fault ;) I should have put the screenshots first, they are more explicit :)

macta:
I'm glad you guys are fixing things like this - they do help.

PR #7715, opened 2020-11-10
Compute and use the screen scale factor in OSWorldRenderer

ronsaldo:
This PR adds the computation code for the screenScaleFactor into OSWorldRenderer. The effect of these changes is in replicating the previous behavior before the VM were not marked as DPI aware with the old renderer by drawing into a smaller resolution display and then scale it up when composing into the window.

This also enables texture filtering which is required to make scaling look good, and it fixes the methods for updating a rectangle in the OSSDL2AthensRenderer.


guillep (changes requested):
Hi Ronie, I believe the changes are good. My only point: can the magic numbers some little comment to say what they are and why?

There is a `0.5` in `convertWindowMouseEventPosition:` and a `0.1` that magically changed to `0.05` in `fetchDPI`.

ronsaldo:
@guillep Are these changes good enough?

guillep (approved):
Yes! Thanks!!

PR #7719, opened 2020-11-11
 Introduce package Collections-Sequenceable-Tests

### LINKED ISSUE #7718: Introduce package Collections-Sequenceable-Tests
- to further modularize and break Collections-Tests into pices we should have a package Collections-Sequenceable-Tests
  and move accordingly the classes testing package Collections-Sequenceable there

astares:
Fix #7718


gcotelli (approved):

MarcusDenker:
I sometimes wonder if we should not merge some of the smaller package into one... is is really good to have Collections-Sequenceable in it's own package? In the past we split up packages just because monticello was faaaar too slow.

MarcusDenker:
@gcotelli thanks for the review!

PR #7727, opened 2020-11-12
Add ExternalAddress  class>>#allocate:bytesDuring:

gcotelli:
Add `ExternalAddress  class>>#allocate:bytesDuring:` to perform an allocation and freeing the memory after a block evaluation so people don't forget to free the memory. Replace senders doing this pattern with this method.

There are some changes due to the formatting, but the only real change are the new senders of `allocate:bytesDuring:`


gcotelli:
It looks like the only failing test is unrelated and it was failing before

gcotelli:
I've just solved the merge conflicts

PR #7747, opened 2020-11-17
Disable the dpi based computation screen scale factor computation in OS X

ronsaldo:
The screen scale factor that is computed through the screen DPI provided by SDL2 is producing wrong results in OS X, which ends making the UI very large and blurry.


MarcusDenker (approved):
I tested the download artefact: looks good

PR #7749, opened 2020-11-17
Variables defined with #primitive:error: are shown as undefined, fixes 7748

### LINKED ISSUE #7748: Variables defined with #primitive:error: are shown as undefined
**Describe the bug**
Browse e.g the #clone method. The IDE companies that "ec" is undefined. But it is set to the error value by the compiler

MarcusDenker:
- The error variable has to be added as a variable with write status
- delegate code generation to the variable when storing the TOS.
- factor out primitiveErrorVariableName

fixes #7748

### DIFF
--- a/src/OpalCompiler-Core/OCASTSemanticAnalyzer.class.st
+++ b/src/OpalCompiler-Core/OCASTSemanticAnalyzer.class.st
@@ -214,15 +214,19 @@ OCASTSemanticAnalyzer >> visitMethodNode: aMethodNode [
 
 { #category : #visiting }
 OCASTSemanticAnalyzer >> visitPragmaNode: aPragmaNode [
+
+	| var |
 	super visitPragmaNode: aPragmaNode.
+	aPragmaNode selector = #compilerOptions: ifTrue: [ 
+		aPragmaNode asPragma sendTo:
+			aPragmaNode methodNode compilationContext ].
 	
-	"we parse compiler options here so the rest of the compilation can react to them"
-	(aPragmaNode selector = #compilerOptions:)
-		ifTrue: [ aPragmaNode asPragma sendTo: aPragmaNode methodNode compilationContext ].
-	
-	"primitives with a primitive error define a variable that can be referenced in the body"
-	aPragmaNode isPrimitiveError ifTrue: [  
-		self declareVariableNode: (RBVariableNode named: (aPragmaNode argumentAt: #error:) value asString) ]
+	"if the pragma is a primitive that defines an error variable, we need to declare a temp
+	for it"	
+	aPragmaNode isPrimitiveError ifFalse: [ ^ self ].
+	var := RBVariableNode named: aPragmaNode primitiveErrorVariableName.
+	self declareVariableNode: var.
+	var variable markWrite
 ]
 
 { #category : #visiting }
--- a/src/OpalCompiler-Core/OCASTSemanticAnalyzer.class.st
+++ b/src/OpalCompiler-Core/OCASTSemanticAnalyzer.class.st
@@ -215,7 +215,7 @@ OCASTSemanticAnalyzer >> visitMethodNode: aMethodNode [
 { #category : #visiting }
 OCASTSemanticAnalyzer >> visitPragmaNode: aPragmaNode [
 
-	| var |
+	| varNode |
 	super visitPragmaNode: aPragmaNode.
 	aPragmaNode selector = #compilerOptions: ifTrue: [ 
 		aPragmaNode asPragma sendTo:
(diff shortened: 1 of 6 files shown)


src/OpalCompiler-Core/OCASTSemanticAnalyzer.class.st
```diff
@@ -214,15 +214,18 @@ OCASTSemanticAnalyzer >> visitMethodNode: aMethodNode [
 
 { #category : #visiting }
 OCASTSemanticAnalyzer >> visitPragmaNode: aPragmaNode [
+
+	| varNode |
 	super visitPragmaNode: aPragmaNode.
+	aPragmaNode selector = #compilerOptions: ifTrue: [ 
+		aPragmaNode asPragma sendTo:
+			aPragmaNode methodNode compilationContext ].
```
dionisiydk, line 222: Maybe it makes sense to push this knowledge about such special pragmas into superclass to have #visitCompilerOptionsPragma: and #visitPrimitiveErrorPragma:

src/OpalCompiler-Core/OCASTSemanticAnalyzer.class.st
```diff
@@ -214,15 +214,18 @@ OCASTSemanticAnalyzer >> visitMethodNode: aMethodNode [
...
-	"we parse compiler options here so the rest of the compilation can react to them"
-	(aPragmaNode selector = #compilerOptions:)
-		ifTrue: [ aPragmaNode asPragma sendTo: aPragmaNode methodNode compilationContext ].
-	
-	"primitives with a primitive error define a variable that can be referenced in the body"
-	aPragmaNode isPrimitiveError ifTrue: [  
-		self declareVariableNode: (RBVariableNode named: (aPragmaNode argumentAt: #error:) value asString) ]
+	"if the pragma is a primitive that defines an error variable, we need to declare a temp
+	for it"	
+	aPragmaNode isPrimitiveError ifFalse: [ ^ self ].
```
dionisiydk, line 226: Then we will avoid such ifTrue checks in all visitors

src/OpalCompiler-Core/OCASTTranslator.class.st
```diff
@@ -654,10 +654,15 @@ OCASTTranslator >> visitParseErrorNode: anErrorNode [
...
-	methodBuilder addPragma: aPragmaNode asPragma.
-	aPragmaNode isPrimitiveError ifTrue: [
-		methodBuilder storeTemp: (aPragmaNode argumentAt: #error:) value].
 
+	| var |
+	methodBuilder addPragma: aPragmaNode asPragma.
+	
+	"if the pragma is a primitive that defines an error variable, we need to store error value 
+	which is on the stack"
+	aPragmaNode isPrimitiveError ifFalse: [ ^ self ].
```
dionisiydk, line 663: And here

src/OpalCompiler-Core/OCASTTranslator.class.st
```diff
@@ -654,10 +654,15 @@ OCASTTranslator >> visitParseErrorNode: anErrorNode [
...
+	var := aPragmaNode methodNode scope lookupVar: aPragmaNode primitiveErrorVariableName.
+	var emitStore: methodBuilder.
```
dionisiydk, line 665: Maybe it can be done more elegantly considering that the semantics analyzer will generate a variable node and therefore it should be visited together with other variables.
We can have #visitPrimitiveErrorVariable: on the base visitor class and translator will override it with simple #emitStore: call.

src/OpalCompiler-Core/TemporaryVariable.class.st
```diff
@@ -16,13 +16,7 @@ TemporaryVariable >> acceptVisitor: aProgramNodeVisitor node: aNode [
...
 	^ scope node temporaries 
 		detect: [ :each | each name = name ]
-		ifNone: [ 
-			" ugly workaround to support temps defined by primitives"
-			| pragma |
-			pragma := scope node methodNode pragmas detect: [ :each | each isPrimitiveError ].
-			pragma ifNil: [ ^nil ].
-			^ RBVariableNode named: (pragma argumentAt: #error:) value asString
-			]
+		ifNone: [ nil ]
```
dionisiydk, line 19: It would be nice to avoid nil returns. Let it fail if no node exists. And superclass method can be removed. 
There is only RBVariableNode>>isDefinition sender and the logic can be modified by #isDefinedByNode: message (returning false from Variable as default).

src/OpalCompiler-Core/PrimitiveErrorVariable.class.st
```diff
@@ -0,0 +1,35 @@
...
+]
+
+{ #category : #visiting }
+PrimitiveErrorVariable >> acceptVisitor: aProgramNodeVisitor node: aNode [
+	^ aProgramNodeVisitor visitTemporaryNode: aNode
+]
+
+{ #category : #queries }
+PrimitiveErrorVariable >> definingNode [
+	^node
```
dionisiydk, line 29: I think it would be better to call this variable #definingNode. 
I know it was an existing logic to use a variable node as result. But I wonder why it done like that?
I would expect a pragma node to be a definition. Or even relax to method node as default for other temps (which will simplify the implementation).

dionisiydk (approved):

MarcusDenker:
I will merge this, we can later improve

PR #7754, opened 2020-11-17
ReLiteralArrayContainsCommaRule should ignore all UFFI selectors

### LINKED ISSUE #7753: ReLiteralArrayContainsCommaRule should ignore all UFFI selectors
Scenario:
  - open the browser on a method including FFI calls (for example WBWindowsWebBrowser class>>#shellExecute:file:parameters:directory:show:)
  - the system highligts the call in yellow and critisizes that Literal array contains a #,"

image

But in lint rule ReLiteralArrayContainsCommaRule there UFFI selector #ffiCall: is left out, but not
other selectors like #ffiCall:module

astares:
Fix #7753


astares:
CI failing unrelated

PR #7757, opened 2020-11-17
7755 re excessive arguments rule should exclude ffi callouts 

### LINKED ISSUE #7755: ReExcessiveArgumentsRule should exclude FFI callouts
The lint rule ReExcessiveArgumentsRule critisizes methods with an excessive number of arguments. Currently the limit is by default 5 arguments.

Unfortunately it also laments on UFFI calls, who often have several arguments and quickly hit that limit:

image

Proposal:
The **ReExcessiveArgumentsRule** should exclude FFI callouts as it is done in other rules too (like ReLiteralArrayContainsCommaRule)

astares:
Fix #7755

### DIFF
--- a/src/GeneralRules/ReExcessiveArgumentsRule.class.st
+++ b/src/GeneralRules/ReExcessiveArgumentsRule.class.st
@@ -30,6 +30,12 @@ ReExcessiveArgumentsRule >> argumentsCount [
 
 { #category : #running }
 ReExcessiveArgumentsRule >> basicCheck: aMethod [
+
+	"If method includes UFFI calls then it is OK to have high number of arguments"
+	(aMethod ast allChildren anySatisfy: [:node | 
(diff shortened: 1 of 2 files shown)


src/GeneralRules/ReExcessiveArgumentsRule.class.st
```diff
@@ -30,6 +30,12 @@ ReExcessiveArgumentsRule >> argumentsCount [
 
 { #category : #running }
 ReExcessiveArgumentsRule >> basicCheck: aMethod [
+
+	"If method includes UFFI calls then it is OK to have high number of arguments"
+	(aMethod ast allChildren anySatisfy: [:node |
```
dionisiydk, line 35: There is #isFFIMethod

astares: @dionisiydk Oh fine - lets handle this in https://github.com/pharo-project/pharo/pull/7789

PR #7765, opened 2020-11-18
Move-isHaltNode-to-AST 

MarcusDenker:
#isHaltNode is an extension method if NewTools-Debugger. This PR moves it to the AST package.

NOTE: after this is merged, we need to commit the change to the NewTools repository, too


Ducasse:
Do we open another PR for the NOTE: after this is merged, we need to commit the change to the NewTools repository, too

MarcusDenker:
I committed it:

https://github.com/pharo-spec/NewTools/commit/d7cb4085549d8b0da7e7ff3066ea13ccdcccc2f2

PR #7774, opened 2020-11-19
7770 cleaning old definition printer

Ducasse:


Ducasse:
I do not get why it would be breaking since it was internal method logic refactoring.
let us try again.

Ducasse:
I would like to test this now that #7816 has been integrated.

Ducasse:
The tests are broken which is already an improvement.

Ducasse:
To see with the merge of marcus fix if the test passes

MarcusDenker:
Yes, looks better. Still some other tests failing, but far less

Ducasse:
This is strange because I do not see the difference.

##one 
T2OldPharoClassDefinitionPrinterTest>>#testTrait3AndTagTests.
Got 'Trait named: #Trait3  uses: Trait2  package: ''Tests-Traits-MOP''' instead of 'Trait named: #Trait3   uses: Trait2   package: ''Tests-Traits-MOP'''.
Stacktrace

TestFailure
Got 'Trait named: #Trait3
	uses: Trait2
	package: ''Tests-Traits-MOP''' 
instead of 'Trait named: #Trait3
	 uses: Trait2
	 package: ''Tests-Traits-MOP'''.

## two 
testAddTraitPattern

## three
T2OldPharoClassDefinitionPrinterTest>>#testTBehavior

MarcusDenker:
My fixe was for the other failing ones (the ones that made DNU for named: package, they where from Epicea)

MarcusDenker:
(i think we did the close/reopen twice)

Ducasse:
A new issue has been pushed: https://github.com/pharo-project/pharo/pull/7845

PR #7775, opened 2020-11-19
Merge-MethodNameEditor-RBMethodNameEditor 

MarcusDenker:
MethodNameEditor was not used, merge it into it's subclass RBMethodNameEditor

No deprecation needed, as this is not public API.


Ducasse:
We will have to redo it in Spec2 and to remove the dependency to the ui from this package.

PR #7780, opened 2020-11-19
Cleanup-Unused-Ivars

MarcusDenker:
There are over 120 classes with instance variables that are not referenced.

This PR removes some and adds a (for now skipped) test in  NoUnusedVariablesLeftTest which we can turn on as soon as we have checked all the problematic classes.

### DIFF
--- a/src/Beacon-Core-Tests/MemoryLoggerTest.class.st
+++ b/src/Beacon-Core-Tests/MemoryLoggerTest.class.st
@@ -1,9 +1,6 @@
 Class {
 	#name : #MemoryLoggerTest,
 	#superclass : #TestCase,
-	#classInstVars : [
-		'previousRecordings'
-	],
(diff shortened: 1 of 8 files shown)


MarcusDenker:
I wrote a snipped about the implementation of the test: https://blog.marcusdenker.de/finding-unused-variables-an-example-for-first-class-variables

<blockquote><div><strong><a href="https://blog.marcusdenker.de/finding-unused-variables-an-example-for-first-class-variables">Finding Unused Variables: An example for First Class Variables - Marcus Denker Blog</a></strong></div><div>As you might be aware, we added to Pharo the concept of “Firsts class Variables” that is, there is a meta object for every variable in the system.</div></blockquote>

astares (approved):
Nice one! Yes - this really deserves cleaning

PR #7803, opened 2020-11-21
Add additional linux platform checking

### LINKED ISSUE #7802: Add additional linux platform checking
UnixPlatform already has support for #isLinux but we could enhance to check for a few prominent 
Linux derivates / distributions like Ubuntu, Fedora, ...

astares:
Fix #7802


astares:
CI failures unrelated

astares:
Any reviewer?

guillep:
I'm not particularly against it... But it has a problem: why stop with these distros? :)
Somebody will always complain their distro is not in the list.

Maybe a solution more like linux's `uname` that returns the name of the current distro is more flexible from that point of view.

thoughts?

noha:
No, uname is not more flexible. There is is something called Linux standard base (https://en.wikipedia.org/wiki/Linux_Standard_Base) that defines a common system for major linux distros. The canonical way of getting information about an LSB system is 

`lsb_release -a`

But this is a binary. So relying on /etc/os-release is the proper way to do. But I'm with Guille here that making individual linux distros a mention in the image is not such a good idea. Neither is doing a substring search over the file ;)
If we define our linux target to be a LSB linux target which is feasible then I would make it detect linux and the Linux distro object offers access to the dictionary in the os-release file. In my system it looks like 

cat /etc/os-release
NAME="Ubuntu"
VERSION="20.04.1 LTS (Focal Fossa)"
ID=ubuntu
ID_LIKE=debian
PRETTY_NAME="Ubuntu 20.04.1 LTS"
VERSION_ID="20.04"
HOME_URL="https://www.ubuntu.com/"
SUPPORT_URL="https://help.ubuntu.com/"
BUG_REPORT_URL="https://bugs.launchpad.net/ubuntu/"
PRIVACY_POLICY_URL="https://www.ubuntu.com/legal/terms-and-policies/privacy-policy"
VERSION_CODENAME=focal
UBUNTU_CODENAME=focal

which should be easy to parse and put into a dictionary

<blockquote><div><strong><a href="https://en.wikipedia.org/wiki/Linux_Standard_Base">Linux Standard Base - Wikipedia</a></strong></div></blockquote>
<blockquote><div>Ubuntu</div><div><strong><a href="https://ubuntu.com/">Enterprise Open Source and Linux | Ubuntu</a></strong></div><div>Ubuntu is the modern, open source operating system on Linux for the enterprise server, desktop, cloud, and IoT.</div></blockquote>
<blockquote><div><strong><a href="https://help.ubuntu.com/">Official Ubuntu Documentation</a></strong></div></blockquote>
<blockquote><div>Launchpad</div><div><strong><a href="https://bugs.launchpad.net/ubuntu/+bugs">Bugs : Ubuntu</a></strong></div><div>Ubuntu also includes a wide variety of software through its network of software repositories. Once your system is installed you can simply call up a list of all the existing tools out there and choose any of them for immediate installation over the internet.
</div></blockquote>
<blockquote><div>Ubuntu</div><div><strong><a href="https://ubuntu.com/legal/data-privacy">Data privacy | Ubuntu</a></strong></div><div>Ubuntu is an open source software operating system that runs from the desktop, to the cloud, to all your internet connected things.</div></blockquote>

astares:
@guillep 
uname can only be applied to general Unix. It gives you "Linux" on Linux.  The "/etc/os-release" solution and the LDB approach is the way forward in Linux world

@noha 
OK - I reworked and generalized the PR:
 - removed the specific #isXXXX checking 
 - parse the /etc/os-release to have access to all properties 
 - added better utility methods to check for the distribution and Linux derivate following the Linux specs

guillep:
I think @noha kind of understood what I explained poorly :)
What I meant is that I'd prefer a method `distributionName` rather than 10 `is[Distro]` methods

astares:
@guillep Yes - and thats what and why I wrote: the isXXX methods now [got removed](https://github.com/pharo-project/pharo/pull/7803/commits/97b6c9f9be5868f671e1021edf768e60e95d514a) (and I included the other changes suggested too)

So please review again how the PR looks now so we can move on. Thanks in advance!

So for instance you can now  check using:

```Smalltalk
Smalltalk os isLinuxDistributionWithID: 'Ubuntu'
Smalltalk os isLinuxDerivateOf: 'debian'
```
Comparision is case insensitive (thats why I gave the example here differently)

One can also access (on Linux) all parameters using
```Smalltalk
Smalltalk os linuxIdentificationParameters inspect
```

References:    
- [https://www.freedesktop.org/software/systemd/man/os-release.html](https://www.freedesktop.org/software/systemd/man/os-release.html)
- [https://gist.github.com/natefoo/814c5bf936922dad97ff](https://gist.github.com/natefoo/814c5bf936922dad97ff)

src/System-Platforms/UnixPlatform.class.st
```diff
@@ -106,6 +103,21 @@ UnixPlatform >> lineEnding [
...
+	"see https://www.freedesktop.org/software/systemd/man/os-release.html"
+
+	| props extractor |
+	props := Dictionary new.
+	self isLinux ifFalse: [ ^props ].
+	extractor := [ '/etc/os-release' asFileReference contents linesDo: [ :each | 
+		props at: (each copyUpTo: $=) asSymbol put: (each copyAfter: $=) ]].
+
+	"Do nothing when it does not exist"
+	[ extractor value ] on: FileDoesNotExistException do: [ self error: 'Only available on modern distributions' ].
```
noha: Can we replace "modern distributions" with "distributions supporting the linux standard base" ?

noha (approved):

src/System-Platforms/UnixPlatform.class.st
astares: @noha  Yes - we can ;)

I changed the comment

MarcusDenker (approved):

PR #7815, opened 2020-11-22
Fix side-effect in MCOrganizationDefinition

LinqLover:
The usage of `#sort` instead of `#sorted` manipulates the passed category array, which not only restricts the argument type to Array but also is very unexpected and confusing behavior. Furthermore, this led to recent test failures in smalltalkCI [1] because the passed array could not be a read-only object.

With this PR, the parameter is #sorted into a copy instead so both bugs should be resolved.

Please review! This is my first contribution to Pharo, so I hope I did not miss any formalities.
Please squash this PR when merging.

[1] https://travis-ci.org/github/hpi-swa/smalltalkCI/jobs/742264299#L457

### DIFF
--- a/src/Monticello/MCOrganizationDefinition.class.st
+++ b/src/Monticello/MCOrganizationDefinition.class.st
@@ -12,7 +12,7 @@ Class {
 
 { #category : #'instance creation' }
 MCOrganizationDefinition class >> categories: aCollection [
-	^ self new aCollection: aCollection
+	^ self new categories: aCollection
 ]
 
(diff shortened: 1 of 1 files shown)


LinqLover:
I don't see the CI is failing. Looking at the Jenkins overview page, it seems as if it was never red. Could someone confirm that my small patch is not responsible for any regressions? :-)

MarcusDenker:
Failing tests are not related to the change.

MarcusDenker (approved):
looks good.

PR #7833, opened 2020-11-23
`#ensureCreateDirectory` Shouldn't Resolve

seandenigris:
`#ensureCreateFile` already has the correct behavior, which is to return self. Otherwise, one may ask a FileLocator to ensure and get back a FileReference.


dionisiydk (approved):

MarcusDenker (approved):

seandenigris:
Thanks everyone :)

PR #7836, opened 2020-11-23
Clean the OSWindowGenericRenderer texture APIs for properly implementing a text renderer

ronsaldo:
These changes remove the broken text rendering attempt through the SDL2 renderer API. Instead we are adding the missing APIs for implementing text rendering at the scene graph renderer level by cleaning the texture APIs.


guillep:
> These changes remove the broken text rendering attempt through the SDL2 renderer API. Instead we are adding the missing APIs for implementing text rendering at the scene graph renderer level by cleaning the texture APIs.

Hi @ronsaldo could you add some comments to the code?
Could you detail here why the SDL2 text rendering API is broken? Is it SDL2 text rendering API that we should not use or it was our bindings that were wrong? Or you're just moving that low level API to the scene graph renderer?

guillep:
image

@MarcusDenker, that's kinda obvious, there are no tests for all this :)

MarcusDenker:
I am just adding the tag always when the CI would be green if we would not have the failing test on Win and the crash on mac... it has no other meaning then the CI green would have.

ronsaldo:
> Could you detail here why the SDL2 text rendering API is broken? Is it SDL2 text rendering API that we should not use or it was our bindings that were wrong?

There is no SDL2 text rendering API, it was just a convenience API that I added some months ago and it never worked (completely unusable due to bad performance).

> Or you're just moving that low level API to the scene graph renderer?

I am reimplementing the text rendering at the scene graph renderer via texture drawing. I adapted the same mechanism for drawing with the gpu to use the texture drawing API via multiple passes.

guillep:
> > Could you detail here why the SDL2 text rendering API is broken? Is it SDL2 text rendering API that we should not use or it was our bindings that were wrong?
> 
> There is no SDL2 text rendering API, it was just a convenience API that I added some months ago and it never worked (completely unusable due to bad performance).
> 
> > Or you're just moving that low level API to the scene graph renderer?
> 
> I am reimplementing the text rendering at the scene graph renderer via texture drawing. I adapted the same mechanism for drawing with the gpu to use the texture drawing API via multiple passes.

Ok!

PR #7839, opened 2020-11-24
Simplify-ReUnaryAccessingMethodWithoutReturnRule

MarcusDenker:
- use #category instead of #categoryOfElement: on the class organizer
- use noneSatisfy:

### DIFF
--- a/src/GeneralRules/ReUnaryAccessingMethodWithoutReturnRule.class.st
+++ b/src/GeneralRules/ReUnaryAccessingMethodWithoutReturnRule.class.st
@@ -20,11 +20,10 @@ ReUnaryAccessingMethodWithoutReturnRule class >> uniqueIdentifierName [
 
 { #category : #running }
 ReUnaryAccessingMethodWithoutReturnRule >> basicCheck: aMethod [
-	(aMethod numArgs > 0 or: [ aMethod isAbstract ]) ifTrue: [ ^ false ].
-	((aMethod methodClass organization categoryOfElement: aMethod selector) asString beginsWith: 'accessing') ifFalse: [ ^ false ].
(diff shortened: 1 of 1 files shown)


src/GeneralRules/ReUnaryAccessingMethodWithoutReturnRule.class.st
```diff
@@ -20,11 +20,10 @@ ReUnaryAccessingMethodWithoutReturnRule class >> uniqueIdentifierName [
...
 { #category : #running }
 ReUnaryAccessingMethodWithoutReturnRule >> basicCheck: aMethod [
-	(aMethod numArgs > 0 or: [ aMethod isAbstract ]) ifTrue: [ ^ false ].
-	((aMethod methodClass organization categoryOfElement: aMethod selector) asString beginsWith: 'accessing') ifFalse: [ ^ false ].
-	aMethod ast nodesDo: [ :each | each isReturn ifTrue: [ ^ false ] ].
 
-	^ true
+	(aMethod numArgs > 0 or: [ aMethod isAbstract ]) ifTrue: [ ^ false ].
+	(aMethod category beginsWith: 'accessing') ifFalse: [ ^ false ].
+	^ aMethod ast allChildren noneSatisfy: [ :each | each isReturn ]
```
gcotelli, line 26: `allChildren` is producing a new collection on each call so I don't know if this change has some performance impact. On the other hand `nodesDo:` is also producing intermediate collections for iterating the children , so maybe this change has no measurable impact on performance.

MarcusDenker: yes, it adds just one more creation of an intermediate collection...we do that *a lot* everywhere... every node does it when the visitor asks for #children...

PR #7842, opened 2020-11-24
Calypso: Don't Show "Should Implement" Protocol for Abstract Classes

seandenigris:
A  false positive scenario was reported on Discord today because there is no way to tell Calypso not to warn about "should be implemented" for abstract classes.

Here is the behavior with the fix in place:
2020-11-24T084035343244-0500


seandenigris:
Hold off on this. May not be an issue

seandenigris:
It appears that the actual problem was that Calypso shows the "should not implement" protocol even if empty because the target class isAbstract. It appears to work as expected in P9.

PR #7847, opened 2020-11-24
Ban rule "class not referenced" for WebBrowser subclasses

### LINKED ISSUE #7846: Ban rule "class not referenced" for WebBrowser subclasses
- the critique "Class not referenced" is OK for subclasses of WebBrowser - so we should ban the rule

astares:
Fix #7846


MarcusDenker:
maybe better is to implement #isUsed like in TestCase and similar classes

astares:
@MarcusDenker Did'nt know about #isUsed. I changed the PR to use it, no need for the banned rule and manifest now.

Ready for a second review

MarcusDenker (approved):

PR #7854, opened 2020-11-25
CompiledMethod-hasProperty-use-includesProperty 

MarcusDenker:
speedup #hasProperty: a little by using includesProperty: instead of #propertyAt:ifAbsent:


MarcusDenker:
Actually more speedup then I expected...

[SystemNavigation new allHalts] bench  

- now: 30/sec (low level scanning for selectors, but missing all full blocks)
- using #containsHalt and #7779 merged: 23/second (that is, checking property that is set be the compiler)
- with this improvement: 38 per second

benchmark is bad in any way (e.g. we even find a different amount of method), but nevertheless interesting

PR #7855, opened 2020-11-25
Merge Spec2 0.8 and NewTools 0.5

### LINKED ISSUE #123: [Debugger] When creating a method the focus should be in the code pane
When a method is created in the debugger, after selecting the package.
The keyboard focus should be in the code pane and not in the stack frames list.

### LINKED ISSUE #128: use Breakpoint announcement in StObjectBreakpointInspection>>#initializePresenters
StObjectBreakpointInspection>>#initializePresenters is listening to MetaLinkChanged, but it is only interested in Breakpoints.

We can now change it to listen to breakpoint announcements

estebanlm:
together because there are some packages that moved from one to the other :)

NewTools 0.5
===
**New Inspector is now default in Pharo!**

- Inspector Introduces concept of "inspection concept" which allows to configure how the presentation will be show (and if it will be shown). For now it does not do much, but potential is huge. 
- Inspector uses `SpMillerPaginatorPresenter`.
- Inspector uses popover to allow command expansion in the toolbar (otherwise is too big).
- Inspector panes are closeable now.
- Inspector, Playground now understand <meta+g> command
- Playground pages now can "open in new page".
- Activation in single click (which triggers column expansion) is now default for inspections (it was double click, but it felt weird).
- Added several inspections (extensions to objects) to match what we had before.
- Most commands now delegate its behavior to their contexts to allow better handling of them.
- Introduce "icon providers" to handle how icons are provided in applications.
- Refactor the way icons are used in `StPharoApplication` to allow switching of icon sets and the gradual transition from current naming to a better, semantic way of name icons.
- Fixes https://github.com/pharo-spec/NewTools/issues/128
- Fixes https://github.com/pharo-project/pharo/pull/7700
- Fixes https://github.com/pharo-spec/NewTools/issues/123
- Several bugfixes.

Spec 0.8
===
- `Spec2-Tools` was moved to `NewTools-Methods`, `NewTools-ChangeSorter` and `NewTools-KeymapBrowser`.
- `SpCodePresenter` commands now are delegating functionality to the presenter (this allow better handling of them).
- output transmissions now can be composed, to allow a single activation to trigger to different child presenters.
- `SpWindowPresenter` now understands withDecorations/withoutDecorations
- `SpWindowPresenter>>#centered` and `SpWindowPresenter>>#centerRelativeTo:` now work correctly
- Modal dialogs now close on ESC correctly, even when focus is in a text field/area.
- `SpBoxLayout` homogeneous property now works correctly.
- several style modifications.
- several bugfixes


Ducasse:
esteban none of the builds built.

MarcusDenker:
tests look good...

PR #7858, opened 2020-11-25
7857-Cleaning-ends-of-class-names-in-Calypso

### LINKED ISSUE #7857: Cleaning ends of class names in Calypso.
SycRefactoringPreview is a presenter so we should name it: SycRefactoringPreviewPresenter for example.
This is annoying to have to guess and be forced to look at the superclass to know.

Ducasse:
fixes: #7857 just some renaming to make things clearer.


Ducasse:
Zinc test.

PR #7876, opened 2020-11-27
optimize hasProperty: on Morph a little

MarcusDenker:
- do not allocate block in MorphExtension>>#hasProperty, just use nil
- Morph>>#hasProperty: does not need to push the ivar twice

- remove GLMBrick>>#hasProperty:, which the same as the implementation in the superclass


MarcusDenker:
I looked a the other methods of the property API: there are a lot more that can be simplified. And Brick we should not touch (it seems to need the super methods to fullfill a trait).

(these methods are called a lot... )


I will close this and make  combined PR with all the suggested improvements.

PR #7878, opened 2020-11-27
simplify SubscriptionRegistry>>#subscriptionsHandling:

### LINKED ISSUE #7764: simplify SubscriptionRegistry>>#subscriptionsHandling:
**Describe the problem**
SubscriptionRegistry>>#subscriptionsHandling: uses Array>>streamContents: as an optimisation

But this should never be needed as the size of registrations is never large enough were this would make sende.

** Solution **

just use select: and convert to an Array. This could be a chance to add select:as: to unify as we have already collect:as:

dupriezt:
Fixes #7764 
- Added Collection>>#select:as:
- SubscriptionRegistry>>#subscriptionsHandling: now uses #select:as: to build the array instead of #streamContents:

Note: for Collection>>#select:as: I initially tried to use #addAll: to fill the collection to return, but Array does not support #add:. So I used the same method that is used in Collection>>#collect:as: -> #fill:from: .Even though it looks a bit strange to provide #fill:from: with a block that does nothing, it makes #select:as: work when it's asked to produce an Array

### DIFF
--- a/src/Collections-Abstract/Collection.class.st
+++ b/src/Collections-Abstract/Collection.class.st
@@ -1184,6 +1184,16 @@ Collection >> select: aBlock [
 ]
 
 { #category : #enumerating }
+Collection >> select: aBlock as: aClass [
+	"Evaluate aBlock with each of the receiver's elements as argument.
+	Collect the elements for which aBlock evaluates to true in an instance of aClass. Answer the resulting collection"
+	
+	| selectedElements |
+	selectedElements := (self select: aBlock).
+	^ (aClass new: selectedElements size) fillFrom: selectedElements with: [ :each | each ] ; yourself
+]
+
+{ #category : #enumerating }
 Collection >> select: selectBlock thenCollect: collectBlock [
 	"Utility method to improve readability."
 
--- a/src/Announcements-Core/SubscriptionRegistry.class.st
+++ b/src/Announcements-Core/SubscriptionRegistry.class.st
@@ -133,10 +133,9 @@ SubscriptionRegistry >> subscriptionsForClass: subscriberClass [
 
 { #category : #accessing }
 SubscriptionRegistry >> subscriptionsHandling: anAnnouncement [
-	^ Array streamContents: [ :s|
-			subscriptions do: [:each| 
-				(each handlesAnnouncement: anAnnouncement)
-					ifTrue: [ s nextPut: each ]]]
+	"Returns an Array containing the subscriptions that handle anAnnouncement"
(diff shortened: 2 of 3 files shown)


src/Collections-Abstract/Collection.class.st
```diff
@@ -1183,6 +1183,16 @@ Collection >> select: aBlock [
 	^newCollection
 ]
 
+{ #category : #enumerating }
+Collection >> select: aBlock as: aClass [
+	"Evaluate aBlock with each of the receiver's elements as argument.
+	Collect the elements for which aBlock evaluates to true in an instance of aClass. Answer the resulting collection"
+	
+	| selectedElements |
+	selectedElements := (self select: aBlock).
```
gcotelli, line 1192: Why not use `Object>>as:` ?
So this can be 
`^(self select: aBlock) as: aClass` and will avoid to fill the collection if it is of the same type.

gcotelli (changes requested):

src/Collections-Abstract/Collection.class.st
```diff
@@ -1183,6 +1183,16 @@ Collection >> select: aBlock [
 	^newCollection
 ]
 
+{ #category : #enumerating }
+Collection >> select: aBlock as: aClass [
```
dionisiydk, line 1187: This method deserves a test

src/Collections-Abstract/Collection.class.st
dupriezt: I tried this replacement, but the image freezes when I save the new #select:as: method.
Maybe because it's used by the announcer system, which tries to announce that this method was changed, but this method fails?

src/Collections-Abstract/Collection.class.st
dupriezt: I added a simple one.
`CollectionTest>>#testSelectAs`
```Smalltalk
testSelectAs
	| collection result |
	collection := OrderedCollection new add: 1; add: 3; add: 2; yourself.
	result  := collection select: [:i | i even not] as: Array.
	self assert: result class equals: Array.
	self assert: result equals: #(1 3).
```

guillep:
We are discussing, we agree with the comments. The implementation of `select:as:` should simply move the definition from `subscriptionsForClass:`.

Also the test only covering `OrderedCollection` is not representative of this refactor: the subscriptions is an `IdentitySet`, which cannot be transformed to an array using `as:`.

I'll close it for now. Please reopen if those points are going to be reworked

PR #7881, opened 2020-11-27
[fix] fixes selector extraction when asking for it at the first chara…

hogoww:
…cter of a method
@guillep That's probably not the best fix, but it outlines the problem =)
Let me know if you want some improvements

### DIFF
--- a/src/Tools-CodeNavigation/CNSelectorExtractor.class.st
+++ b/src/Tools-CodeNavigation/CNSelectorExtractor.class.st
@@ -53,12 +53,9 @@ CNSelectorExtractor >> extractSelectorFromAST: ast atPosition: aPosition [
 	   self foo:|#bar => Here a user may want foo: or bar.
 	For now we decided to favor foo: to motivate people to indent code correctly"
 	| offset position bestNodeAtPosition |
-	"special case when asking for a selector at the first character of a method"
-	offset := aPosition = 1 ifTrue: [ 0 ]
-		ifFalse: [
-		(ast sourceCode at: aPosition - 1) isSeparator
-			ifTrue: [ 0 ]
-			ifFalse: [ -1 ]].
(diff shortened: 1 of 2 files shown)


src/Tools-CodeNavigation/CNSelectorExtractor.class.st
```diff
@@ -53,9 +53,12 @@ CNSelectorExtractor >> extractSelectorFromAST: ast atPosition: aPosition [
 	   self foo:|#bar => Here a user may want foo: or bar.
 	For now we decided to favor foo: to motivate people to indent code correctly"
 	| offset position bestNodeAtPosition |
-	offset := (ast sourceCode at: aPosition - 1) isSeparator
```
guillep, line 56: Why not `(aPosition - 1 max: 0)`?

guillep (changes requested):

src/Tools-CodeNavigation/CNSelectorExtractor.class.st
hogoww: max: 0 would fail, because it would ask for at: 0 in a 1 based context.
I though max: 1 would be good, but actually it finds a letter, which is not a separator, set offset to -1, and fails later in the code.

hogoww: (And I did this to fix the problem in the simplest way possible, didn't even try to do something pretty)

guillep: Well, there is no need to rush :).

I'd prefer a solution with `or:` which would read better and make it only two branches explicitly.

`(aPosition = 1 or: [(ast sourceCode at: aPosition - 1) isSeparator]) ...`

hogoww: No need to rush a merge indeed! 
I agree, I like your way better, and it's commited :)

Ducasse:
All tests are green.

PR #7884, opened 2020-11-27
remove-unused-charScannerPrimitive

### LINKED ISSUE #6841: primitive unusable since squeak 2.8
primScanCharactersFrom: startIndex to: stopIndex in: sourceString rightX: rightX stopConditions: stops kern: kernDelta
	"Primitive. This is the inner loop of text display--but see 
	scanCharactersFrom: to:rightX: which would get the string, 
	stopConditions and displaying from the instance. March through source 
	String from startIndex to stopIndex. If any character is flagged with a 
	non-nil entry in stops, then return the corresponding value. Determine 
	width of each character from xTable, indexed by map. 
	If dextX would exceed rightX, then return stops at: 258. 
	Advance destX by the width of the character. If stopIndex has been
	reached, then return stops at: 257. Optional. 
	See Object documentation whatIsAPrimitive.
	Historical note: this primitive has been unusable since about Squeak 2.8 when the shape of the CharracterScanner class changed. It is left here as a reminder that the actual primitive still needs supporting in the VM to keep old images such as Scratch1.4 alive - tpr"
	<primitive: 103>
	^self basicScanByteCharactersFrom: startIndex to: stopIndex in: sourceString rightX: rightX

tesonep:
Removing unused primitive.
Fixes #6841

### DIFF
--- a/src/Text-Scanning/CharacterScanner.class.st
+++ b/src/Text-Scanning/CharacterScanner.class.st
@@ -309,28 +309,12 @@ CharacterScanner >> plainTab [
 ]
 
 { #category : #scanning }
-CharacterScanner >> primScanCharactersFrom: startIndex to: stopIndex in: sourceString rightX: rightX stopConditions: stops kern: kernDelta [
-	"Primitive. This is the inner loop of text display--but see 
-	scanCharactersFrom: to:rightX: which would get the string, 
-	stopConditions and displaying from the instance. March through source 
-	String from startIndex to stopIndex. If any character is flagged with a 
-	non-nil entry in stops, then return the corresponding value. Determine 
-	width of each character from xTable, indexed by map. 
-	If dextX would exceed rightX, then return stops at: 258. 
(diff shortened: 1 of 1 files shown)


MarcusDenker (approved):

Ducasse:
Pablo we have three failing tests.

windows-64 / Tests-windows-64 / testDoIt – Windows64.Fuel.Tests.Core.Base.FLContextSerializationTest
6s
windows-64 / Tests-windows-64 / testFixObsoleteSharedPools – Windows64.ReleaseTests.Tests.ObsoleteTest
10s
windows-64 / Tests-windows-64 / testObsoleteClasses – Windows64.ReleaseTests.Release.ReleaseTest

MarcusDenker:
The win slave was in very odd state that friday.. but it is unstable very often... the last run it crashed.

(I do not think that anything of this is related to this change)

MarcusDenker:
So, finally it was able to run all tests without crashing... failing tests are not related.

Ready to merge

PR #7888, opened 2020-11-27
Fixed #6391. Filed out methods are now saved into files that have colons in their names

### LINKED ISSUE #6391: fileouting a setter save the same file than a getter in calypso
The two methods are stored in a file with the same name so .... this is quite annoying. 
Because when you need to file out a method, it means that you need it.

olekscode:


Ducasse:
oleksandr : is not accepted on windows.
So could you replace it with DoubleDot?

olekscode:
I'm on it!

olekscode:
Done

However, now `Collection>>do:` will be filed out as `Collection-do...st`
Not very pretty but it should work

olekscode:
I was thinking of underscores,
so for example, we would have `Collection-do_.st`
But theoretically underscores can be used in a method name, so in a crazy scenario when we have `Class >> name_` and `Class >> name:`, they would both file out as `Class-name_.ston`

Ducasse:
What we want is that we get a file name with a unique name. 
So replace : by A and it will work.

olekscode:
@Ducasse I have replaced `:` with `..`
I think it should work fine and always be unique (because `.` can not be used in method names)

gcotelli:
We can use the same logic used for naming the files in the filetree format (It is using dots for sure but I don't remember how many)

PR #7893, opened 2020-11-27
Compile-Empty-Blocks-as-CleanBlocks

MarcusDenker:
Clean block compilation is not yet enabled due to more testing needed to be done for the debugger. But that does not mean that we can not use them already for one specific case: [] the empty block.

There are over 800 uses in the image. This, too, will allow us to find problems early with CleanBlocks if we start to use them step by step.

### DIFF
--- a/src/OpalCompiler-Core/OCASTTranslator.class.st
+++ b/src/OpalCompiler-Core/OCASTTranslator.class.st
@@ -538,11 +538,19 @@ OCASTTranslator >> visitCascadeNode: aCascadeNode [
 
 { #category : #'visitor-double dispatching' }
 OCASTTranslator >> visitFullBlockNode: aBlockNode [
+
 	| compiledBlock |
 	compiledBlock := self compilationContext astTranslatorClass new translateFullBlock: aBlockNode.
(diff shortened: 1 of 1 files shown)


MarcusDenker:
[[]] bench

before: "34981387.245 per second" 
after: "185133399.960 per second" 

~ 5.3 times faster

MarcusDenker:
nice: WeakAnnouncerTest inherits tests from AnnouncerTest. They use empty blocks. They are not GCed as the blocks are referenced. But with clean blocks, the Weak announcers now do work as the block has not refs, GC happens, test fails.

testHandleSubscriberClass
	announcer when: AnnouncementMockA do: [ ].
	self assert: (announcer handleSubscriberClass: self class).
	self deny: (announcer handleSubscriberClass: AnnouncementMockA).

guillep:
Why not just applying the change locally in an image and use it for a couple of days? I feel all these things are random. We need a more systematic approach for something as deep in the system.

MarcusDenker:
Yes, I will do that...

guillep:
Thanks!

I'll make some time this week to update the bootstrap to set compiler options explicitly (encoder, full blocks). Like that we can avoid a recompilation step and we will see problems in this kind of things earlier too :)

PR #7894, opened 2020-11-27
Move old definition printing to a the legacy printer

guillep:
Make the system use the legacy printer
Remove redundant/unused methods

### DIFF
--- a/src/EpiceaBrowsers-Tests/EpApplyPreviewerTest.class.st
+++ b/src/EpiceaBrowsers-Tests/EpApplyPreviewerTest.class.st
@@ -164,16 +164,19 @@ EpApplyPreviewerTest >> testCategoryRenameWithPreviousRollback [
 
 { #category : #tests }
 EpApplyPreviewerTest >> testClassAdditionWithClassRemoved [
+
 	| aClass classDefinition |
 	aClass := classFactory newClass.
-	classDefinition := aClass definition.
+	classDefinition := aClass definitionString.
 	self setHeadAsInputEntry.
 
 	aClass removeFromSystem.
 
 	self assertOutputsAnEventWith: [ :output | 
 		self assert: output class equals: EpClassAddition.
-		self assert: output behaviorAffected definitionSource equals: classDefinition ]
+		self
+			assert: output behaviorAffected definitionSource
+			equals: classDefinition ]
 ]
 
 { #category : #tests }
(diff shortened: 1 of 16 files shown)


Ducasse:
testTraitAdditionWithTraitRemoved – Unix64.EpiceaBrowsers.Tests.Integration.EpApplyPreviewerTest
<1s
testClassAdditionWithSuperclassChanged – Unix64.EpiceaBrowsers.Tests.Integration.EpApplyPreviewerTest
<1s
testClassAdditionWithInstanceVariablesChanged – Unix64.EpiceaBrowsers.Tests.Integration.EpApplyPreviewerTest
<1s
testClassAdditionWithClassRemoved – Unix64.EpiceaBrowsers.Tests.Integration.EpApplyPreviewerTest
<1s
testExampleClassSide – Unix64.Slot.Tests.Examples.ExampleSlotWithStateTest
<1s
testUndeclareSlotFixWhenSlotIsLoaded – Unix64.Slot.Tests.ClassBuilder.SlotErrorsTest


I would like to integrate them so that I can fix them over the week-end.

guillep:
I'll fix the tests, please do not integrate until then.

guillep:
The Epicea tests fail the first time and not the second one.
Checking the transcript it seems this is the fault of a deprecation rewriting:

````
Automatic deprecation code rewrite: The method ClassDescription>>#definition called from EpApplyPreviewerTest>>#testClassAdditionWithSuperclassChanged has been deprecated. definition
```

guillep:
I've fixed all tests but one: `testUndeclareSlotFixWhenSlotIsLoaded`.

This test is abusing of the class definition to recompile a class dynamically using the class defintion string:

```
checkClassRebuild
	"break the recursion while rebuilding"
	classIsRebuild ifTrue: [ ^ self].
	(self definingClass environment hasClassNamed: self slotClassName) ifFalse: [ ^ self ].
	classIsRebuild := true.

	"we rebuild the class, this triggers instance migration"
>>>	self definingClass compiler evaluate: self definingClass definitionString.

	"recompile all methods that access me to generat code for the loaded definition"
	self usingMethods do: [:each | each recompile].
```

however, with the new definition, the line `self definingClass compiler evaluate: self definingClass definitionString.` evaluates to a fluid class builder, and does not re-create the class. Thus the test will never work.

I tried to put that expression in `(...) build` but I did not get it to work. Ideas?

Ducasse:
Ok I will check

Ducasse:
This is strange because when I load your PR. I will have the EpApplyPreviewerTest that fails.
And yes the second times there are green. But I do not get it because you said that you fixed them.

Ducasse:
```
self definingClass compiler evaluate: self definingClass definitionString.
```
this returns builder and not a class. We should then send it the `build` message. 

so it should probably be

```
ShiftClassInstaller new 
makeWithBuilder: (self definingClass compiler evaluate: self definingClass definitionString) shiftClassBuilder 
```
Yes this is not nice.

Ducasse:
Once the fix I did is integrated we will be able to just to 
```
(Object << #MYpoint.....) install
```

Ducasse:
Guille I will integrated and we fix the borken test in another issue

Issue #7901, opened 2020-11-29
Infinite recursion on send of not-understood message to nil


martinmcclure:
**Describe the bug**
Any send of a not-understood message to nil will cause infinite recursion, unless the send is to an undefined variable.
This appears to be introduced by PR #7834, commit 7438a3658, on 26 November.
In UndefinedObject>>doesNotUnderstand:, a VariableNotDeclared is created with a nil variableNode. That nil variableNode is eventually sent a message, starting another evaluation of the method.

**To Reproduce**
Evaluate "nil zork"
Get infinite recursion

**Expected behavior**
Debugger on "zork not understood"

**Version information:**
 - OS: Linux

**Expected development cost**
PR to follow :-)

PR #7902, opened 2020-11-29
Fix infinite recursion upon send of not-understood message to nil

martinmcclure:
Issue #7901


dionisiydk (approved):
We will get rid of all this logic with proper implementation of #7701

PR #7907, opened 2020-11-29
Cleanup tests and mock package of DrTest

### LINKED ISSUE #7906: Cleanup tests and mock package of DrTest
- we should rename DrTests-TestCoverage-MocksForTests into DrTests-TestCoverage-Tests-Mocks as it provided mocks for the DrTest plugin tests
- Also add a class comment to MockForCoverage 
- categorize some uncategorized methods
- tag classes properly
- rename ivar "dTConf" into "pluginConfiguration" to better depict what it is
- remove unnecessary dots at end of method

astares:
Fix #7906


astares:
OK, here CI is right - this one needs more work.

astares:
Will do another PR, we close this one

PR #7917, opened 2020-11-30
optimize-hasDropShadow

MarcusDenker:
#hasDropShadow is called in the main morphic rendering loop. make sure to return the correct result as fast as possible

- pre-check extensions. Most morphs do not have them
- do not use [false], but false.
- small cleanup for #visible


MarcusDenker:
morph := Morph new.
[ morph hasDropShadow  ] bench  

before: 15783917/sec
after: 109261562/sec

speedup of factor 7 for a morph with no extension

pavel-krivanek (approved):

PR #7921, opened 2020-11-30
7914-Fluid-builder-should-encapsulate-call-to-install-the-class

### LINKED ISSUE #7914: Fluid builder should encapsulate call to install the class
Right now the fluid class builder does nto know how to install a class and the users should 
explicitely invoke the ShiftClassInstaller

```
ShiftClassInstaller new makeWithBuilder: ( Object << #X ) shiftClassBuilder 
```

and it should be 

```
( Object << #X ) install
```

Ducasse:
fixes: #7914 introduce install to encapsulate class installation logic.


Ducasse:
The green were green but I wanted to push the use of install too.

PR #7931, opened 2020-12-01
7927-halt-left-in-OmBlockrefreshIfNeededStartingAtsince 

### LINKED ISSUE #7927: #halt left in OmBlock>>#refreshIfNeededStartingAt:since:
**Describe the problem**
There is a #halt left in OmBlock>>#refreshIfNeededStartingAt:since:

MarcusDenker:
remove #halt

fixes #7927


MarcusDenker:
@tinchodias

tinchodias:
Looks good to remove that ugly code. I hope it is not needed!

MarcusDenker:
The idea is that when all halt sends are gone we add a test that will fail and forces people to *never* commit code like this.

PR #7933, opened 2020-12-01
Cleanup tests and mock package of DrTest (Second attempt)

### LINKED ISSUE #7906: Cleanup tests and mock package of DrTest

astares:
Second try to fix #7906

now after the renaming the package with the new name should be included


astares:
CI Exception unrelated

astares:
Any reviewers? Looks complicated but it is not:
  - renamed the package
  - added tags to the classes
  - renamed an ivar "dTconf" into "pluginConfiguration"
  - categorized method1ForMock  and method2ForMock in category "sample methods"

guillep (approved):
If the build passes, it's ok.

I wonder why the mocks cannot be in the same package than the tests...

astares:
@guillep Thanks for the review. I tried to have the mocks inside the test package - but then DrTest gives different results for the coverage which would lead to failing tests. For sure we could adopt the tests (to respect the coverage including the tests) - but then this would mean to adopt them anytime the package changes (for instance when new tests get added later or other).
 
So it is OK as it is now I guess: a specific mock package that is coverage tested.

guillep:
Ah, that makes sense, could be nice to have that in a comment somewhere.
Otherwise I'll ask the same in 6 months :P

astares:
Define "somewhere" - what we need is a clean path to "comments for packages" :)

guillep:
> Define "somewhere" - what we need is a clean path to "comments for packages" :)

Well, right now the package manifest hosts package comments. Could be a solution

PR #7959, opened 2020-12-03
Do not add DoIts to the cache for now. fixes #7955

### LINKED ISSUE #7955: ASTCache can cause huge leaks
**Describe the bug**
ASTCache can keep instances that may survive memory cleanups

**To Reproduce**
1. in Playground, evaluate `RGEnvironment new halt.`, this will open a debugger and cause leaking `StPharoApplication`. Close the debugger.
2. in the same Playground, evaluate: `RGEnvironment someInstance.`
3. reset StPharoApplication: `StPharoApplication classVarNamed: #Current put: nil.` (in the same Playground). This should remove the momory leak with exception of the Playground that is still keeping reference to its old application.
4. Close this Playground
5. Open a new Playground and do GC: `Smalltalk garbageCollect`
6. Evaluate `ReferenceFinder findPathTo: RGEnvironment someInstance.` 

This will show that the old instances are still kept because AST cache record for DoIt `ReferenceFinder findPathTo: RGEnvironment someInstance` keeps compilation context that points to the old playground (that points to the old application that holds the old debugger that holds the old RGEnvironment instance)

Pharo_rlO4wEyMJi

All code:
```smalltalk
"in the first Playground:"
RGEnvironment new halt.
RGEnvironment someInstance.
StPharoApplication classVarNamed: #Current put: nil.

"in the second Playground after closing of the first one"
Smalltalk garbageCollect.
ReferenceFinder findPathTo: RGEnvironment someInstance.
```

MarcusDenker:
Do not add DoIts to the cache for now. fixes #7955

### DIFF
--- a/src/AST-Core/ASTCache.class.st
+++ b/src/AST-Core/ASTCache.class.st
@@ -75,9 +75,12 @@ ASTCache class >> shutDown [
 
 { #category : #accessing }
 ASTCache >> at: aCompiledMethod [
-	^ self 
-		at: aCompiledMethod
-		ifAbsentPut: [ self class cacheMissStrategy getASTFor: aCompiledMethod ]
+	"for doit methods, the ast is stored in the method propery"
+	^ aCompiledMethod propertyAt: #ast ifAbsent: [ 
+		  self
+			  at: aCompiledMethod
+			  ifAbsentPut: [ 
+			  self class cacheMissStrategy getASTFor: aCompiledMethod ] ]
 ]
 
 { #category : #initialization }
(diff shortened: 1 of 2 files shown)


guillep:
This breaks the debugger

image

MarcusDenker:
What makes is harder to test is that now the latest Newtools merge broke it in general... (the reading from the debugger of vars of the requestor)

The tests, we need these changes merged: https://github.com/pharo-spec/NewTools/pull/122

guillep:
I don't follow entirely what is happening in new tools, but I'd say that the memory leak can wait a couple of days. It's better to have a working debugger.

The memory leak has been there for months, and we know that a workaround could be to reset the AST cache + the SpPharoApplication.

Looking at the new commit, I think that yes it's better to put the AST in the method rather than a global cache, that could fix a lot of problems. Also I'd put all those variable thingys in the method and not in the AST. But I don't know what are the consequences, what else could break. Depending on the global AST cache is too brittle...

guillep:
Looking for the AST in the method, then in the global cache seems like a good trade off. @MarcusDenker is this ready?

MarcusDenker:
Yes! I checked it again, all good.

MarcusDenker:
(the debugger to access vars from the requestor needs the latest newtools changes, but that is independent).

PR #7972, opened 2020-12-04
Cleanup-EncoderForV3PlusClosures

MarcusDenker:
the EncoderForV3PlusClosures hierarchy still had the non-closure version EncoderForV3 as a dedicated superclass.

We do not (and never will) support non-closure bytecode sets..

This PR cleans up the Hierachy and removes EncoderForV3 and the test methods like #supportsClosures


Ducasse:
Yes yes yes

PR #7976, opened 2020-12-04
7953 huge debugger st pharo application memory leak

### LINKED ISSUE #7953: Huge debugger/StPharoApplication memory leak
The closed debuggers are never really destroyed. They are kept in the StPharoApplication.

**To Reproduce**
1. eval `RGEnvironment new halt`
2. do garbage collection and check the class instances
```smalltalk
10 timesRepeat: [Smalltalk garbageCollect ].
RGEnvironment someInstance.

ReferenceFinder findPathTo: RGEnvironment someInstance.
```

Pharo_KOeiQYgtXz

**Version information:**
 - OS: Widnows
 - Version: Pharo 9.0.0, 
 - Build information: Pharo-9.0.0+build.922.sha.0f93f71a2e0ee67f6f76a5765315a2620084897c (64 Bit)

StevenCostiou:
Fixes #7953
I believe this is due to the fact that:
- debug sessions and other debugger model objects holds a reference to the exception that provoked a debugger opening
- the exceptions references the interrupted process and context that prevent garbage collection

This may be a side effect of https://github.com/pharo-spec/NewTools/issues/136 where the debugging processes are never released after closing a debugger (under investigation).

With this fix, now the reproduction example of the issue gives the following result:

Screenshot 2020-12-04 at 14 09 40

Trying the same on instances of `StDebugger` returns nil.

### DIFF
--- a/src/Debugger-Model-Tests/DebuggerModelTest.class.st
+++ b/src/Debugger-Model-Tests/DebuggerModelTest.class.st
@@ -24,6 +24,15 @@ DebuggerModelTest >> setUp [
 ]
 
 { #category : #tests }
+DebuggerModelTest >> testClearDebugSession [
+
+	session clear.
+	self assert: session exception equals: nil.
+	self assert: session interruptedContext equals: nil.
+	self assert: session interruptedProcess equals: nil
+]
+
+{ #category : #tests }
 DebuggerModelTest >> testCorrectlyCreateDebugSession [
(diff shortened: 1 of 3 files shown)


Ducasse:
Thanks Steven! Let us try. 
I would really like to know how we could have tools to find such bug.
Please archive this bug somewhere.

PR #7978, opened 2020-12-04
Fix rename refactoring issue - 7947

### LINKED ISSUE #7947: Refactoring - Rename Method Bug
**Bug of Rename Method in GUI**
The renameMethod UI does not have a parameter selected by default, which causes a SubscriptOutOfBounds error and pressing the Up or Down button.

**To Reproduce**
1. Select method to rename (preferably one with arguments)
2. Click on Up or Dn button
3. See error

**Expected behavior**
If we have an argument selected by default or we disable the buttons until one is selected, it should not throw an error

**Screenshots**
imagen

@Ducasse

Lin777:
Fixes #7947 

@Ducasse


guillep (changes requested):
We should write a test for this

Lin777:
Okay, I'll check how to add tests for views

guillep:
You can check the tests in spec2!

Ducasse:
@evelyn do you plan to complement this PR or you do another one?

src/SystemCommands-RefactoringSupport-Tests/SycMethodNameEditorTest.class.st
```diff
@@ -77,6 +77,23 @@ SycMethodNameEditorTest >> testButtonDownClickShouldSelectionIndexBeDecremented
...
+	arrayAfterClick := methodNameEditor argumentsList items.
+	elt2 := arrayAfterClick at: 2.
+	elt1 := arrayAfterClick at: 1.
+	self
+		assert: arrayBeforeClick
+		equals:
+			(arrayAfterClick
+				at: 1 put: elt2;
+				at: 2 put: elt1;
+				yourself)
```
guillep: This test looks strange... why doing at:put: on the array after click? it's not clear what this test is actually testing.

src/SystemCommands-RefactoringSupport-Tests/SycMethodNameEditorTest.class.st
```diff
@@ -77,6 +77,23 @@ SycMethodNameEditorTest >> testButtonDownClickShouldSelectionIndexBeDecremented
 	self assert: methodNameEditor argumentsList selection selectedIndex equals: 2
 ]
 
+{ #category : #'tests - up/down buttons' }
+SycMethodNameEditorTest >> testButtonDownClickWhitoutSelectOption [
```
guillep: typo: whitout => without

Lin777:
@Ducasse I was working on other, but I already add tests for this PR

src/SystemCommands-RefactoringSupport-Tests/SycMethodNameEditorTest.class.st
```diff
@@ -177,9 +201,9 @@ SycMethodNameEditorTest >> testEditParameterShouldHaveGhostText [
 ]
 
 { #category : #'tests - initial state' }
-SycMethodNameEditorTest >> testInitialArgumentListHasNoSelection [
+SycMethodNameEditorTest >> testInitialArgumentListAutoSelection [
 	
-	self assert: methodNameEditor argumentsList selection isEmpty
+	self assert: methodNameEditor argumentsList selection selectedIndexes equals: #(1)
```
guillep, line 192: Why not testing `isNotEmpty` ?
Unless you really care about the first one.

What happens in the list is empty?

guillep (changes requested):
I don't understand some of the tests. Could you explain?

src/SystemCommands-RefactoringSupport-Tests/SycMethodNameEditorTest.class.st
Lin777: This is because now the first argument is selected (if it exists) by default, before it gave an error that when pressing Up or Down button because it did not refer to any selected element

src/SystemCommands-RefactoringSupport-Tests/SycMethodNameEditorTest.class.st
Lin777: my mistake, I correct it

src/SystemCommands-RefactoringSupport-Tests/SycMethodNameEditorTest.class.st
Lin777: It was to show that it changed positions, but it was strange, now I did it in a simpler way and correct a previous similar test

PR #7979, opened 2020-12-05
RBMethodNode #pragmas is the only instance variable for AST children that has an accessor checking for nil

MarcusDenker:
The reason I think is to not waste an OrdereCollection for each method that has no Pragmas.

This PR changed it to be initialized with an empty array (which later will be shared when we enable literal sharing) 
 
It changes to use an array for storage, this is no problem as adding pragmas is not happening at afte ran AST is created (they are created bu the compiler once).


gcotelli (approved):
LGTM!

PR #7981, opened 2020-12-07
add Trait>>#isUsed.

### LINKED ISSUE #7971: Class Not Referenced rule: take Traits into account (isUsed for Traits)
When a Trait is not referenced in the code as a global but has traitUsers, the Code critique does not know it.

We should check to take traitUsers into account, something like:

isUsed
	^super isUsed or: [ self traitUsers notEmpty]

MarcusDenker:
There is already a test

fixes #7971

### DIFF
--- a/src/TraitsV2/Trait.class.st
+++ b/src/TraitsV2/Trait.class.st
@@ -238,6 +238,12 @@ Trait >> isTraitExclusion [
 	^ false
 ]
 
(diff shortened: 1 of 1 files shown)


astares:
@MarcusDenker I guess #traitUsers could be used to find references to traits. Because at the moment Calypso only is able to show references to a class - but not references to a Trait yet.

PR #7982, opened 2020-12-07
7980-fix-checkClassRebuild

### LINKED ISSUE #7980: fix checkClassRebuild
with the old definition changes this method should be revisited. 
It should use a builder instead of the compiler. 
Now what we can do is the following: 

(Object << #MYpoint.....) install

read https://github.com/pharo-project/pharo/pull/7894

MarcusDenker:
- fix checkClassRebuild
- commit a deprecation transform in asEpiceaRingDefinition that comes from running all Slot tests

fixes #7980


MarcusDenker:
I merge it as it fixes a failing test on the main CI build

PR #7988, opened 2020-12-07
fixes #7924: Enhance CodeImporter with better comment

### LINKED ISSUE #7924: Enhance CodeImporter with better comment
**Describe the request**

We should enhance `CodeImporter` class comment with something like this:

> CodeImporter reads chunk format. The chunk format is a code format where elements are delimited by exclamation marks. An element is preceded by some kind of annotation that says what kind of element it is / how it should be read.
For example, the annotation CodeImportTestCaseTestClass #methodsFor: 'some protocol' #stamp: 'GuillermoPolito 5/2/2012 13:35 says that the thing that follows is a method for the class CodeImportTestCaseTestClass, its protocol, and some other metadata like the last modification stamp.
> A chunk can then contain a method, a class comment, a "package organization" which is like a package definition, or an arbitrary "doIt" expression. The definition of classes are integrated within arbitrary expressions.

Bajger:
CodeImporter class has improved comment with better explanation of code chunks, using microdown format.

### DIFF
--- a/src/CodeImport/CodeImporter.class.st
+++ b/src/CodeImport/CodeImporter.class.st
@@ -1,39 +1,43 @@
 "
+## Description
 I'm an object in charge of import source files.  I know a format object that knows how to parse the files, and I import the parsed results into the image.
 
 I handle doIts normally, but some special cases like class organization, class comment or methods are handled via a double dispatch (See my method extensions for that).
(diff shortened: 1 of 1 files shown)


guillep:
This is great!!! Thanks!

PR #7992, opened 2020-12-08
Fixing mem leak after debugger proceeding

StevenCostiou:
This is a NewTools mirror fix, fixing memory leaks issues with the debugger:
- https://github.com/pharo-spec/NewTools/issues/136
- https://github.com/pharo-spec/NewTools/issues/125 (for real)

Basically, when a debugger is opened to debug a UI process, another ui process is spawned so that we can debug the first one and we need to terminate this process after debuggging. This is done by terminating the debug session. 

However:
- Proceeding the execution programmatically closes the debugger, first by resuming the debug session then by terminating it, which happens in a function called by the spec window closing mechanism through a callback. This somehow prevents the spec mechanism to go through the end of its closing procedure, and leaks instances of `SpWindow` (https://github.com/pharo-spec/NewTools/issues/125). I still do not understand the details of that problem.
- If we do not terminate the session in the callback (for instance when using the close button), the ui process created for debugging is never terminated and haunts the image (https://github.com/pharo-spec/NewTools/issues/136).

The problem is solved by:
- Separating the closing from the proceed button and the close button, and not terminating the session in the first case,
- ensuring that when we programmatically close the debugger with the proceed button, the debug session is terminated.

I am not super happy by this fix, and it will be further investigated to produce a nice solution --- but for now it is needed.

### DIFF
--- a/src/NewTools-Debugger/StDebugger.class.st
+++ b/src/NewTools-Debugger/StDebugger.class.st
@@ -378,19 +378,16 @@ StDebugger >> canExecuteRestartCommand [
 
 { #category : #actions }
 StDebugger >> clear [
-	
-	extensionToolsNotebook pages
-		do: [ :page | page activePresenter windowIsClosing ].
+
+	extensionToolsNotebook pages do: [ :page | 
+		page activePresenter windowIsClosing ].
 	extensionTools := nil.
+	self unsubscribeFromSystemAnnouncer.
 	self removeActionsForSession: self session.
 
-	"If I have been proceeded the window has been closed programmatically and I do not need to do this".
-	programmaticallyClosed 
-		ifTrue: [ ^self ].
-
-	self debuggerActionModel clearDebugSession.
-	self unsubscribeFromSystemAnnouncer
-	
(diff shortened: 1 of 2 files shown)


Ducasse:
Hi steven 
Thanks. Now I have a question. Can we write tests to identify that such problems will not show up again. I have the impression that we are a bit running after bugs in this moment. 
And I would really like to see what we can put in place to control more our process.

S.

StevenCostiou:
Quick answers ;)

> Thanks. Now I have a question. Can we write tests to identify that such problems will not show up again. I have the impression that we are a bit running after bugs in this moment.

Writing tests is mandatory, but it require more work to set them because we cannot terminate processes from the debugger as it would terminate the active process... i.e., the test runner. So it's an important TODO that I cannot do right now. Also, I think this is a super hard bug that needs to be reproducible for experiments.

Problem with those recent bugs is that we didn't see them coming, else there would already be tests ;)


> And I would really like to see what we can put in place to control more our process.

Yes, it is becoming more and more complicated for the debugger, as many changes in spec, newtools or in the core have an impact that either make existing code that was working but that become bugs, or highlight hidden bugs.

PR #7997, opened 2020-12-09
TestSkip should be TestSkipped

### LINKED ISSUE #7996: TestSkip should be TestSkipped
the name is misleading
- its not a test that it should start with "Test..."
- It is an exception that is signaled 

```Smalltalk
SkippingTest signal 
```

would be better

astares:
Fix #7996


astares:
CI problem unrelated

Ducasse:
Ok for me. Tx

PR #7998, opened 2020-12-09
7993-Completion-broken-while-editing

### LINKED ISSUE #7993: Completion broken while editing ...
```
named: aName uses: aTraitCompositionOrCollection instanceVariableNames: string package: aPackageName env: anEnvironment
	| trait |
	trait := self classInstaller
		make: [ :builder | 
			self
				configureBuilder: builder
				withName: aName
				traitComposition: aTraitCompositionOrCollection
				instanceVa^^^Cursor^^^ someSlots
				packageName: aPackageName ].
	self assert: [ trait class class = MetaclassForTraits ].
	self assert: [ trait class superclass = Trait ].
	^ trait
```

MarcusDenker:
- Make sure that CoASTResultSetBuilder implements the new visitor methods for Global/Temp/Arg variables.
- #visitGlobalNode: and #visitTemporaryNode: can be removed

fixes #7993

### DIFF
--- a/src/HeuristicCompletion-Model/CoASTResultSetBuilder.class.st
+++ b/src/HeuristicCompletion-Model/CoASTResultSetBuilder.class.st
@@ -124,7 +124,7 @@ CoASTResultSetBuilder >> visitGlobalVariableNode: aRBVariableNode [
 
 { #category : #visiting }
 CoASTResultSetBuilder >> visitInstanceVariableNode: aRBVariableNode [ 
-	^ self visitNode: aRBVariableNode
+	^ self visitVariableNode: aRBVariableNode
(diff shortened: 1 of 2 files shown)


guillep (changes requested):
A test could be easily added by extending `CoASTResultSetBuilderTest`. There are lots of examples there.

Ducasse:
I just wonder if this bug arrived because of a recent change?

MarcusDenker:
yes, I merged yesterday an improvement for the visitor (to call #visitGlobalVariableNode: and so on).

Ducasse:
Ah ok I prefer this.

PR #8002, opened 2020-12-09
Class>>#usesTrait: and Class>>#usesTraitTransitively:

### LINKED ISSUE #8000: Add Class>>#useTrait: and Class>>#useTraitTransitively:
I suggest adding #useTrait: and #useTraitTransitively: to Class, so that there is an easy way of asking a class whether it directly uses a trait (#useTrait:), or whether it or one of its superclasses does (#useTraitTransitively).

dupriezt:
Fixes: #8000

### DIFF
--- a/src/Kernel/Class.class.st
+++ b/src/Kernel/Class.class.st
@@ -1217,18 +1217,6 @@ Class >> usesPoolVarNamed: aString [
 	^self allSharedPools anySatisfy: [:each | each usesClassVarNamed: aString]
 ]
 
-{ #category : #'as yet unclassified' }
-Class >> usesTrait: aTrait [
-	"Returns whether self is among the users of aTrait. Note that this will return false if aTrait is used by a superclass of self and not self itself. If you want to get true in this case, use #usesTraitTransitively:"
-	^ aTrait users includes: self
-]
-
-{ #category : #'as yet unclassified' }
-Class >> usesTraitTransitively: aTrait [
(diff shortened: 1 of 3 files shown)


tesonep (changes requested):
These methods should be classified and also they should be in the T2Traits package as extensions. The kernel should not have trait related code; as traits are loadable after. The initial kernel does not support traits.

dupriezt:
> These methods should be classified and also they should be in the T2Traits package as extensions. The kernel should not have trait related code; as traits are loadable after. The initial kernel does not support traits.

Done

Ducasse:
I read the nice comment of usesTraitTransitively: and I was shocked because I was thinking that the the method was looking in the subtraits (the composite). So this made me think about the method name. 
May be we should name it usesTraitWithSuperclasses: ?
Now I do not like that superclasses: is last word. 
It would be nice if we could get a pattern like with withAll:.... 
usesTraitSuperclassly:

or we could have

aClass locallyUseTrait: T1

and 

aClass useTrait: T1 (which is transitively:).

Ducasse:
BTW I would love to have some tests.

dupriezt:
I renamed the methods.
#usesTrait: for whether a class or its superclasses uses the trait
#usesTraitLocally: for whether a class directly has the usage of the trait in its class declaration

Both methods start with "use", so that they are side-by-side in the method list of Class.

I also added tests for them.

Ducasse:
Oh this is super cool. 
Tx thomas.

PR #8024, opened 2020-12-09
Class>>#asRingDefinition should not send #definition but #definitionString

### LINKED ISSUE #8023: Class>>#asRingDefinition should not send #definition but #definitionString
Catched another one after running tests: also Class>>#asRingDefinition should not send #definition but #definitionString

astares:
Fix  #8023


olekscode:
Related issue: https://github.com/pharo-project/pharo/issues/8026

PR #8031, opened 2020-12-09
ReSendsUnknownMessageToGlobalRule laments wrongly on class variables

### LINKED ISSUE #8030: ReSendsUnknownMessageToGlobalRule laments wrongly on class variables
The rule **ReSendsUnknownMessageToGlobalRule** wrongly laments on class variables as "globals". 

Scenario 1:
=========
Look at RenrakuBaseTestCase>>#tearDownscreamerRule for example:

image

where ScreamerRule is a class variable and not a global variable. So the critics shows a problem - but there is none. The code if fully working.

Scenario 2:
=========
Another example:  RenrakuBaseTestCase>>#setupScreamerRule - if you open it in the browser the critics will lament on message sends to the class variable ScreamerRule - although this is fully working as a method

image

astares:
Fix #8030


Ducasse:
May be we should add a test to cover it.

PR #8033, opened 2020-12-10
Add a lint rule for classes to have no "Uncategorized" classes in the package

### LINKED ISSUE #8032: Add a lint rule for classes to have no "Uncategorized" classes in the package
If one or more class tags exist in the package but a class is not using it or not tagged with an own class tag then the system browser shows this class as "Uncategorized".

image

This can for instance easily happen when you have an uncategorized/untagged class is in the package and a manifest class gets added (which is usually categorized in "Manifest").

We should provide a lint rule that checks this. 

To heal the situation the user should put the class into one of the existing tag categories or create a custom one (like "Base", if there is no other name)

astares:
Fix #8032


astares:
image

astares:
CI Problem is unrelated

astares:
Any reviewers?

astares:
Any reviewers?

guillep (approved):

PR #8040, opened 2020-12-10
CoTypeInferencer should inherit from RBProgramNodeVisitor

MarcusDenker:
This way it inherits backward compatible methods for e.g. #visitGlobalVariableNode:


Ducasse:
Tests are green on linux and mac...

PR #8043, opened 2020-12-10
8036-Spotter-uses-definition-for-class-instead-of-definitionString

### LINKED ISSUE #8036: Spotter uses definition for class instead of definitionString
But this is painful to find.

Ducasse:
fixes: #8036 tryingt o fix spotter using definition instead of definitionString


Ducasse:
May be there is a problem around here

GTSpotterExtensionsOnClassTest>>#testBasic has been deprecated. definition

PR #8045, opened 2020-12-10
Fix class variable entanglement issue

### LINKED ISSUE #8044: Copying class using class browser 'entangles' class variables
**Describe the bug**
When copying a class using the context menu 'Copy' command in the class browser, the class variables in the new class are 'entangled' with the old class. In other words, when the value of the class variable in the old class is changed, the value of the class variable in the new (copied) class also changes and vice versa.

**To Reproduce**
1. Open a class browser on the DateAndTime class.
2. Right-click on the class and select 'Copy'. Fill in a name for the copy, e.g. 'DateAndTimeCopy'.
3. Open a new playground window:
  a. Type `DateAndTime localTimeZone.` and print it. If you're in time zone CET like me, this will print `a LocalTimeZone(LT1:00)`.
  b. Type `DateAndTimeCopy localTimeZone.` and print it. This will print the same value.
  c. Type `DateAndTime localTimeZone: (TimeZone offset: -8 hours).` and do it.
  d. Performs step 3b again. It should print `a LocalTimeZone(LT1:00)`, but instead it prints `a TimeZone(PST)`.

**Expected behavior**
Above, in step 3d, it should have printed `a LocalTimeZone(LT1:00)` since we didn't change the value of the local time zone class variable in the `DateAndTimeCopy`.

**Screenshots**
Class browser context menu 'Copy' command
(Screenshot 1 - Class browser context menu 'Copy' command)

Screenshot 2020-12-10 at 12 23 54
(Screenshot 2 - Example results showing bug)

**Version information:**
 - OS: macOS Mojave 10.14.6 (18G6042)
 - Affected versions: Pharo9.0.0 (build: 950, commit: 9a9de50) and Pharo8.0.0 (build: 1124, commit: 0932da8)

**Expected development cost**
Zero. I will provide a fix.

**Additional context**
Pharo version 7.0 does not have this issue.

jvalteren:
Add failing test to demonstrate issue #8044.

### DIFF
--- a/src/Kernel-Tests-Extended/ClassTest.class.st
+++ b/src/Kernel-Tests-Extended/ClassTest.class.st
@@ -154,6 +154,72 @@ ClassTest >> testClassSide [
 ]
 
 { #category : #tests }
+ClassTest >> testClassVariableEntanglement [ 
+	| firstClassName firstClass secondClassName secondClass thirdClassName thirdClass |
+
+	firstClassName := 'MyFirstClass'.
+	firstClass := Object subclass: firstClassName
+		instanceVariableNames: ''
+		classVariableNames: 'MyVar'
+		package: self categoryNameForTemporaryClasses.
+	[ 
+
+		firstClass class compile: 'myVar
+	^ MyVar'.
+		firstClass class compile: 'myVar: anObject
+	MyVar := anObject'.
+
+		secondClassName := 'MySecondClass'.
+		secondClass := Object subclass: secondClassName
+			instanceVariableNames: ''
+			classVariableNames: 'MyVar'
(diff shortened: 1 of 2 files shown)


Ducasse:
@tesonep I added you as a reviewer because the shift class builder is involved.

MarcusDenker:
I checked: already in Pharo8 we were re-using the binding, that means the bug exists there, too.

It is true that this is not good: even sharing the *value* of the var would be bad, thus the solution of the PR I think is good.

By using the string, we create new variables that point to nil, thus the copied class is completely independent.

MarcusDenker:
(closed and re-opend to run tests again, was still in a bad state form last week)

PR #8046, opened 2020-12-10
7923-Broken-traits-definitions

Ducasse:
adding missing bits.


Ducasse:
I do not understand why I can get a conflict. 
I did not touch this code.

MarcusDenker:
My theory: There was a rewrite that was done in your image automatically, but at the same time we merged already a PR with the same rewrite.

Ducasse:
:)

pavel-krivanek:
Is the failure caused by the content of this PR? I do not see any reason why it should.

Ducasse:
I relaunch it.

PR #8051, opened 2020-12-10
Fixing undo refactory

### LINKED ISSUE #5010: Undoing a refactoring is not working anymore
and this is important to be able to do it.

Lin777:
Fixes #5010 

Fixed Undo Refactoring, it now works with all refactorings that are done by command, except with Remove Method this because the command of this refactoring uses the SycSilentlyRemoveMethodStrategy class to remove the method, and not a refactoring class.

@Ducasse


Ducasse:
Good can you open an issue about this behavior of the removeMethod?

Ducasse:
In the future we will have to be smarter than just have 20 refactorings undone. Can you open a bug entry for that.
I would like to know if we cover the following scenario. 
- I rename a method and it touches 22 method
- I use a refactoring that use another one. 
- when we undo do we undo the composite including all the contained operations or not. 

And we will clean the code around the singleton because we do not care that new raises an error.

Lin777:
@Ducasse I'm not sure if I understood what you said correctly, this is that I understood:

- change the name of a method, which changes its 20 references, when doing undo it must change the name of the method to the previous one including the 20 references that it changed.

- use a refactoring that uses another internally, and doing undo should cover undoing changes from both the first refactoring and the internal one.

Is it correct? or did you mean something else?

Issue #8011, opened 2020-12-09
Unable to bootstrap on macOS 11.0.1


jgfoster:
**Describe the bug**
Bootstrap instructions don't work

**To Reproduce**
1. Fork/Clone this repository to your macOS 11.0.1 machine
2. Follow the README.md instructions to see this error:

```
jfoster@vienna pharo % git status
On branch Pharo9.0
Your branch is up to date with 'origin/Pharo9.0'.

nothing to commit, working tree clean
jfoster@vienna pharo % git log -1
commit 26abf5dc31dd3d9f57ee8927fa0d8753e60af006 (HEAD -> Pharo9.0, upstream/Pharo9.0, origin/Pharo9.0)
Merge: 0b3b95ac86 da7c02fcfe
Author: Guille Polito <guillermopolito@gmail.com>
Date:   Wed Dec 9 21:38:50 2020 +0100

    Merge pull request #8008 from astares/8004-tagsForClasses
    
    tagsForClasses
jfoster@vienna pharo % BUILD_NUMBER=42 BOOTSTRAP_ARCH=64 bash ./bootstrap/scripts/bootstrap.sh
+ set -e

...

+ ./pharo Pharo.image --no-default-preferences /Users/jfoster/code/pharo/bootstrap/scripts/prepare_image.st --save --quit
./pharo: line 11: /Users/jfoster/code/pharo/pharo-vm/Pharo.app/Contents/MacOS/Pharo: Bad CPU type in executable
jfoster@vienna pharo % 
```

I assume that it is trying to run a 32-bit executable. `make` exhibits a similar error.

astares:
If I'm not mistaken bootstrap in general is based on 32 bit and then converts the image to 64 bit, no?
But you have to ask @guillep for details

jgfoster:
Unfortunately, macOS no longer runs 32-bit applications, so the process is currently broken.

MarcusDenker:
https://github.com/pharo-project/pharo/pull/8064 has been merged

PR #8064, opened 2020-12-10
Bootstrapper in 64 and 32 bits

carolahp:
Implementing the optional argument BOOTSTRAPPER_ARCH to define the architecture of the Pharo system that runs the bootstrap process (32 or 64 bits).
When not specified, its default value is 64 bits.

Use as follows: 
```BUILD_NUMBER=42 BOOTSTRAP_ARCH=32 BOOTSTRAPPER_ARCH=32 bash ./bootstrap/scripts/bootstrap.sh```


jgfoster:
This works for me on macOS 10.15.7 (Catalina) and 11.0.1 (Big Sur). This fixes #8011.

Ducasse:
Thanks for the feedback.

PR #8067, opened 2020-12-11
RBGenericNodeVisitor-tests 

MarcusDenker:
- This PR adds tests for RBGenericNodeVisitor
- adds #visit:anySatisfy: and #visit:noneSatify: 
- move RBCommentNodeVisitorTest close to the other visitor tests


Ducasse:
I would extract RBGenericVisitor new into a method. Else this is nice.

MarcusDenker:
I do not see the "RBGenericVisitor new", it already has a class side API visit:anySatisfy: ... and so on.

PR #8070, opened 2020-12-11
Make use of current working directory primitive on windows

guillep:
to avoid FFI if possible

### DIFF
--- a/src/System-Platforms/Win32WideString.class.st
+++ b/src/System-Platforms/Win32WideString.class.st
@@ -61,6 +61,7 @@ Win32WideString >> asString [
 
 	codepage := 65001.
 	out := ByteArray new: (self size * 4) + 1.
+	out pinInMemory.
 
 	r := OSPlatform current
 		wideCharacterToMultiByteCodepage: codepage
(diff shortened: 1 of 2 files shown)


astares:
Why avoid FFI? To my knowledge it was moved from primitive to FFI in the past?

guillep:
> Why avoid FFI?

The current working directory is required at startup and bootstrap times, in which we don't have FFI enabled or available yet (respectively). Also as you may notice the FFI implementation of this method was not using uffi, but rather using `ExternalFunction` by hand (again, because uffi is not available at bootstrap time), which is not only verbose but it is also tying it to one FFI backend...

> To my knowledge it was moved from primitive to FFI in the past?

AFAIR, this was implemented through FFI **only for windows** because the primitive was not correctly working on Windows (and that was added later, so historical reasons). Unix and OSX are and were always using a primitive. This just makes windows to work like the others :).

I think that obtaining the working directory is something that is pretty standard and basic functionality, and having it available as a primitive is valuable.

guillep:
Also, please notice that this primitive is already there in the VM from some time!

astares:
@guillep Thanks for the details - just wanted to know because if I'm not mistaken we once had it like it is now.

PR #8114, opened 2020-12-15
Improve Compiler plugin implementation

### LINKED ISSUE #8113: Improve Compiler plugin implementation
Strange hierarchy:
- at the moment we have OCCompilerASTPlugin saying it is "I am an abtract superclass for compiler plugins." in the class comment
- but the "dynamic" (block closure based) compiler plugins like OCCompilerDynamicASTPlugin are not implemented as a subclass - (they inherit from Object directly)
- also 'OCCompilerASTPlugin isAbstract ' returns false at the moment
- additionally:  not all the methods in OCCompilerASTPlugin are required for a plugin - only #transformation: and #priority methods define the API called from the compiler side

=> We should cleanup and unify into the following hierarchy
  - OCCompilerPlugin  (abstract)
  -- OCDynamicCompilerPlugin
  -- OCStaticCompilerPlugin
   --- FFICompilerPlugin
   --- ASTPluginMeaningOfLife

Other code problems we should directly fix:
- also OCCompilerASTPPlugin **has a class side subclassResponsibility for #priority** which is fulfilled by both subclasses (ASTPluginMeaningOfLife and FFICompilerPlugin) - but it still returns a **hard coded instance side #priority** of 100 
- #transform is categorized/classified as API - but it is not public API (its called privately)
- fix typo in ASTPluginMeaningOfLife>>#transform
- the plugin classes should be in a class category 'Plugins' similar to how their tests were tagged

astares:
fix #8113


astares:
image

MarcusDenker:
One thing to care about: there might be users that subclass the current abstract class. 

Another idea was that it had AST in it's name as there might me other plugins added later that are doing things at different stages of the compiler. There will be for sure not just one CompilerPlugin.

astares:
OK - so what do you propose:

- OCCompilerPlugin (abstract)
-- OCASTCompilerPlugin (abstract)
--- OCDynamicASTCompilerPlugin
--- OCStaticASTCompilerPlugin
---- FFICompilerPlugin
---- ASTPluginMeaningOfLife
-- [OCOtherStageCompilerPlugin] (abstract) then later
--- ...

guillep:
I had a brief look and it looks ok to me. I like the class comments.

Is this safe for integration?

astares:
This is for P10 - but we develop for P11 now.

I created a new up to date PR against "Pharo11" branches (see https://github.com/pharo-project/pharo/pull/11761)

Issue #8077, opened 2020-12-13
Dropping a code file produce a DNU: ByteString did not understand #utf8Decoded


Ducasse:
```
requestDropReference: i

	| fileName |

	fileNames ifNil: [ 
		"The file names are not stored in the event, they need to be obtained by a VM primitive"
		^ FileReference requestDropReference: i ].
	
	"The event already knows the files (case of events transformed from OSWindow events)"
	fileName := self fileNames at: i.
	^ (FilePathEncoder decode: fileName) asFileReference
```

MarcusDenker:
The problem is FilePathEncoder class>>#decode:, it calls #utf8Decoded on a string but it is only implemented in ByteArray

From Discord discussion: "if you edit the method with the following:  aString asByteArray utf8Decoded it works"

PR #8116, opened 2020-12-15
Correctly decode file names from SDL2, at SDL2

guillep:
- DropFilePlugin is not used nor built in Pharo9 anymore (remove the associated code)
- Correctly decode file names from SDL2, at SDL2
  => users receive already decoded strings, no need to decode them

### DIFF
--- a/src/FileSystem-Core/FileReference.class.st
+++ b/src/FileSystem-Core/FileReference.class.st
@@ -49,20 +49,6 @@ FileReference class >> newTempFilePrefix: prefix suffix: suffix [
 	^ tmpDir / fileName
 ]
 
-{ #category : #'instance creation' }
-FileReference class >> primDropRequestFileName: dropIndex [
-	"Primitive. Return the file name for some file that was just dropped onto the application.
-	Fail if dropIndex is out of range or the primitive is not supported."
-	<primitive: 'primitiveDropRequestFileName' module:'DropPlugin'>
-	^ self primitiveFail 
-]
-
-{ #category : #'drag-and-drop' }
-FileReference class >> requestDropReference: anInteger [ 
-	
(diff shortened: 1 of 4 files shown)


guillep:
Fixes #8077

Ducasse:
Oh Excellent. 
Guille you got two methods from a class that does not look related to this issue.

guillep:
That is Tonel reordering the files, there is nothing to do about it I think...

Ducasse:
ah ok I was thinking that you modified it. I do not get why it is included. 

{ #category : #'*Morphic-Base' }		 { #category : #'*Morphic-Base' }
 Form >> floodFillTolerance [		 Form class >> floodFillTolerance [
 	^ self class floodFillTolerance		 	^ FloodFillTolerance ifNil: [FloodFillTolerance := 0.0]
 ]		 ]

 
  { #category : #'*Morphic-Base' }		 { #category : #'*Morphic-Base' }
 Form class >> floodFillTolerance [		 Form >> floodFillTolerance [
 	^ FloodFillTolerance ifNil: [FloodFillTolerance := 0.0]		 	^ self class floodFillTolerance
 ]		 ]

But I do not get why

guillep:
The code did not change: they were written in different order. These are two methods with the same selector, but one is in the class side and the other is instance side. Somehow, in previous versions this was written out as "instance side, then class side", but in my commit tonel decided to do the other way around.

Maybe it's because the previous version of this file was old, and tonel was changed in between?

Because when there is a new tonel version we don't migrate all the source base (we shouldn't need to, it should be backwards compatible).

Ducasse:
ah ok I did see not it. but now I understand. This is several times that I see this pattern.

PR #8121, opened 2020-12-15
ClassParser-reduce-NewUndeclaredWarning

MarcusDenker:
When running the tests of the ClassParser, we get  alot of NewUndeclaredWarning messages in the transcript.

This is a workaround to do what the exception does, without logging.  (the code looks not that great, but it is a nice example of a client that will allow us to improve the Exception handling in the future)


MarcusDenker:
works. There are more of these log entries from other uses, I will check them step by step. This one can be merged

Ducasse:
Thanks I did not that the classParser was a bad guy.

PR #8130, opened 2020-12-16
8129-Debugger-should-log-the-exception-always

### LINKED ISSUE #8129: Debugger should log the exception always
**Describe the bug**
The debugger should log the exception in the PharoDebug.log as the first thing and it should be done always. Does not matter the UI manager used or if there is or not debuggers. Today if there is an exception during the initialization of the session it is not logged. 

Today 
```
requestDebuggerOpeningFor: anException

	<debuggerCompleteToSender>
	self class environment
		at: #OupsDebugRequest
		ifPresent: [ :requestClass | 
			(requestClass newForException: anException) submit.
			^ self ].
	Smalltalk
		logError: anException printString
		inContext: anException signalerContext

```

We need to log the exception always. It is not true the debuggers handle it. 
Ways to reproduce: 
- Generate an error during the startup process
- Generate an error during Morphic UI rendering.

tesonep:
This is not a pretty way to solve this. But in the meantime, we need to fix it.

The error log should be generated always. Sometimes is the only clue we have to see if the error occurs.
See the issue for details in which cases it does not work.
Fix #8129

### DIFF
--- a/src/UIManager/UIManager.class.st
+++ b/src/UIManager/UIManager.class.st
@@ -639,20 +639,23 @@ UIManager >> request: aStringOrText initialAnswer: defaultAnswer title: aTitle e
 UIManager >> requestDebuggerOpeningFor: anException [
 
 	<debuggerCompleteToSender>
+	Smalltalk logError: anException printString inContext: anException signalerContext.	
+	
 	self class environment
 		at: #OupsDebugRequest
 		ifPresent: [ :requestClass | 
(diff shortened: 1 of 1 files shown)


StevenCostiou:
My 2 cents since you merged it before review: it does not seem worse than before to me, so it's ok.
Are there tests on that part of the code (I do not remember)? I feel that testing the log is always generated is now necessary, as you highlighted that it is a problem when that log is not generated.

PR #8138, opened 2020-12-17
Fix Abstract instance variables issue

### LINKED ISSUE #8100: RB - Abstract instance variable refactoring doesn't work correctly
Abstract instance variable, not works well overrides some changes, and finally are bad results.

For example we can get some results like this:

```
method
    ^ a + self b
```

where a and b are instance variables

Lin777:
Fixes #8100 

Before fix we got:
```
method
      ^ a +self b
```

after fix:

```
method 
     ^ self a + self b
```
this was due because the secuence of refactorings was executed in diferents models, in concequence only changes of last variable were executed and not of all. I changed to reuse the model in the sequence of refactorings (of this refactory) and now works well.
@Ducasse


Ducasse:
This is cool.
Do you have a test to cover this?

Lin777:
this refactoring was only a commad, it use a set of RBAbstractInstanceVariableRefactorings, due this it wasn't have tests.

should be better if I create a compose refactoring to this refactoring?

PR #8139, opened 2020-12-17
8127-FFICalloutAPITestmethod2-sends-non-implemented-call

Bajger:
Removed FFICalloutAPITest>>method2, since it has no senders in this package (UnifiedFFI-Tests). Also this method sent #call:, which is not implemented on this test.


guillep:
Thanks!

PR #8141, opened 2020-12-18
Debugger fuel out stack command

StevenCostiou:
Actually fixes a new tools issue: https://github.com/pharo-spec/NewTools/issues/148

Now, what I do not really like is that it introduces a dependency from fuel to 2 new tools debugger packages.
Maybe the package with fuel commands (and actions for GT) for debuggers should be externalized from Fuel, but then where to put it?

If the PR passes the CI, I will leave that to think about to the community...

Screenshot 2020-12-18 at 09 10 28


StevenCostiou:
Ok, not building because of dependencies...

MarcusDenker:
I downloaded the build artefact: I can fuel out a stack and dropping that on the image opens a debugger.

StevenCostiou:
Super :) Now if someone could review and merge, I need this PR to be accepted before doing the mirror changes in NewTools.

MarcusDenker (approved):

Ducasse:
You can package it in a separate package this way the new debugger will not depend on fuel.

StevenCostiou:
> You can package it in a separate package this way the new debugger will not depend on fuel.

Yes this is what I did last week ;)

PR #8152, opened 2020-12-18
8120-RPackageOrganizer-and-RPackage-should-report-about-if-a-package-is-a-test-package

### LINKED ISSUE #8120: RPackageOrganizer and RPackage should report about if a package is a test package.
To avoid string manipulation everywhere we should introduce 
`RPackageOrganizer >> testPackages` and `RPackage >> isTestPackage`

Bajger:
added method for determining test packages, based on package name
added method for selecting testing packages from package organizer
added unit test methods for package organizer and package 
refactored image cleaner method for obtaining test package names
added test method for image cleaner

### DIFF
--- a/src/RPackage-Core/RPackage.class.st
+++ b/src/RPackage-Core/RPackage.class.st
@@ -957,6 +957,19 @@ RPackage >> isForeignClassExtension: categoryName [
 	^ categoryName first = $* and: [(self isYourClassExtension: categoryName) not]
 ]
 
+{ #category : #testing }
+RPackage >> isTestPackage [
+	"1. Test package ends with suffix -Tests. Suffix is case sensitive.
+	 2. Or test package contains '-Tests-' in middle.
+	Some examples: "
+	"RPackage named: 'MockPackage-Tests' isTestPackage >>> true"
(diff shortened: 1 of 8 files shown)


Ducasse:
Super cool!.
Thanks again. 
We were busy today fixing nasty bugs. So this is great because I would have done it that nicely!!!

src/RPackage-Core/RPackage.class.st
```diff
@@ -957,6 +957,19 @@ RPackage >> isForeignClassExtension: categoryName [
...
 
+{ #category : #testing }
+RPackage >> isTestPackage [
+	"1. Test package ends with suffix -Tests. Suffix is case sensitive.
+	 2. Or test package contains '-Tests-' in middle.
+	Some examples: "
+	"RPackage named: 'MockPackage-Tests' isTestPackage >>> true"
+	"RPackage named: 'MockPackage-tests' isTestPackage >>> false"
+	"RPackage named: 'MockPackage' isTestPackage >>> false"
+	"RPackage named: 'MockPackage-Tests-Package' isTestPackage >>> true"
```
Bajger: I just realized I miss parantheses here. I'll add another commit to fix this.

PR #8159, opened 2020-12-21
8158-ContextquickStep-is-unused-and-strange 

### LINKED ISSUE #8158: Context>>#quickStep us unused and strange
**Describe the problem**
Context>>#quickStep is not used. It is not clear what it does (and why it is needed).

A real problem is that the state of being in this "QuickStep mode" is modelled as a class variable. This means that two
instances of Context (e.g. two open debuggers using quicksStep) would interfere with each other.

I propose to remove the code

MarcusDenker:
Context>>#quickStep is not used. It is not clear what it does (and why it is needed).

A real problem is that the state of being in this "QuickStep mode" is modelled as a class variable. This means that two
instances of Context (e.g. two open debuggers using quicksStep) would interfere with each other.

this PR removes the code

fixes #8158

### DIFF
--- a/src/Debugging-Core/Context.extension.st
+++ b/src/Debugging-Core/Context.extension.st
@@ -80,40 +80,6 @@ Context >> pcRangeContextIsActive: contextIsActive [
 ]
 
 { #category : #'*Debugging-Core' }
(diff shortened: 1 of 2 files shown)


guillep (approved):

MarcusDenker:
failing tests not related

Issue #8198, opened 2020-12-28
Reorganize World Menu


gcorriga:
Reorganize the WorldMenu to declutter the Tools menu.

Do this by introducing two new menus - Develop and VCS.
The Develop menu contains all tools for developing within the image - System Browser, Playground, etc.
The VCS menu contains all tools for version control and collaborative work - Iceberg, Change tools etc.

The Tools menu can then be used for additional tools that can be loaded in the image - e.g. Roassal3.

Screenshots:
![image](https://user-images.githubusercontent.com/7032660/103236691-703fe380-493d-11eb-898e-d3f4a418c6af.png)
![image](https://user-images.githubusercontent.com/7032660/103236711-8057c300-493d-11eb-9fbc-57fd0c8a0a70.png)
![image](https://user-images.githubusercontent.com/7032660/103242742-48f21200-494f-11eb-874b-04c639a7ed27.png)
![image](https://user-images.githubusercontent.com/7032660/103242759-57402e00-494f-11eb-83a7-4b4111617469.png)
![image](https://user-images.githubusercontent.com/7032660/103242774-6626e080-494f-11eb-917f-601cb689afcf.png)
![image](https://user-images.githubusercontent.com/7032660/103242808-7fc82800-494f-11eb-9f03-614377593b42.png)

macta:
From Discord -

TM:  Pretty much every application avoids 2 word menus - and those that did seemed to get rid of them... VCS is used by IntelliJ - but actually its not just version control stuff in that menu why not "Organise" - for things you use to organise code, and for Tools - its name has had its day as everything is a tool - I vote for "Community" in part because I like the community feel that Pharo pushes - and why Develop? I would push for "Browse" this would give you:

Pharo, Browse, Organise, Debug, System, Community, Windows, Help - which seems to give a nice complimentary feel. 

(I would also move File Browser into Organise - and remove all "Browser" suffixes), also Process Browser might go into Debug
Reading the menu words again,  I might also think of your journey and change to: 

Pharo, Browse, Debug, Organise, Community, System, Windows, Help

GC: the idea is
'Develop' has all the tools to manage code changes in the image
'VCS' has all the tools for collaboration - version control etc.
I don't like 'Community' too much since there aren't any community-related items there. 'Contrib' would be more precise (but I don't particularly like it). We could go for Applications
anyway I think we all agree more or less with splitting the current Tools menu, and we are now debating the details
[...]

PR #8201, opened 2020-12-29
8198 reorganize world menu

gcorriga:
Reorganize the world menu as per [(https://github.com/pharo-project/pharo/issues/8198)] .

Change is structured as follows:
- change order score for the main menus from 0,1,2,etc to 0,10,20...99
- structure the menu in groups. Individual menu items use these groups as parents as opposed to the main menus.

Changes to individual menu items are as per screenshots in the linked issue.


Ducasse:
Under VCS we should have iceberg as first entry and I would put (Monticello Browser) and (catalog Browser) 
to indicate that they are deprecated.

Ducasse:
DependencyAnalyser could go on tools or develop.
Iceberg could also have an etnry on develop because we use it all the time.

gcorriga:
Iceberg is meant to be the first entry - not sure why that's not happening when loading the code. I'll take a look.

Not sure if wrapping menu items in parenthesis does convey the right message - I would suggest adding '(Legacy)' to the labels:
image

gcorriga:
The reason the Iceberg menu item isn't included is that the standard pharo repository entry as set up in the image does not include the Iceberg packages.

How should I include this change? Do I need to send a separate PR for Iceberg?

gcorriga:
> DependencyAnalyser could go on tools or develop.
> Iceberg could also have an etnry on develop because we use it all the time.

The idea behind the current set up is to split the menu items between Develop (tools to browse and change the behavour of the image) and VCS (tools to manage code changes, packages, and collaboration). This is regardless of how often the tools are used.

Ducasse:
Excellent idea. 
This is important that people know the status of things. 
Could you put (preview) after drTest ?

gcorriga:
(Preview) added.

Can you please confirm how we should proceed with the 'Iceberg' menu entry?

astares:
@Ducasse Can you please ellaborate why DrTest still a preview? Would be nice to know more details. I guess it is a useful tool. 
It was a preview in Pharo 8, is now part of the image since over a year and got cleaned now also in Pharo 9. Why declaring it as "Preview" for Pharo 9 still?

Ducasse:
Because it means that they are some parts that we want to improve but did not get the time and julien left for a Java shop :(

Bajger:
Hi @gcorriga ! Can we rename VCS menu item to just "Version Control" or "Versioning"? I know, such abbreviation is quite obvious to SW developers, but still it is abbreviation. 
Also, what remains in Tools? Is it just Roassal? Can it be "Community" or rather "Visualization"? I just think stating "Tools" is a bit ambiguous compared to other development or debug tools. 
Thanks for your insight!

gcorriga:
@Bajger happy to change from VCS to something else, but I'm not keen on using either multiple words or -ing suffixes for a top-level menu name. So I would suggest some other alternatives:

Source
Versions
Changes

As for tools, it needs to be kept at least for pharo 9 for backwards compatibility. But in general the idea is to keep it as the default entry point for any non-standard tools that someone might want to have loaded in the image. E.g. if you load Teapot, the two menu items would be found there.

Ducasse:
Can you use Version Control?

I would like to integrate this so that we can get feedback else it will rot.

Bajger:
Ok, got it! Maybe we could ask for some feedback on discord (kind of survey with just buttons 1, 2, 3 for each alternative)...Just curious what other think...

Ducasse:
Yes this is a good idea but I'm more than super busy in this moment so if you want to take the lead on this be our guest :)

Bajger:
I replied with results on original issue. I guess there were quite few other ideas in place, I was trying to put it together.

gcorriga:
I've updated the PR with the latest changes given the feedback received.

Ducasse:
Nice!

Ducasse:
Broken tests are unrelated.

Ducasse:
I revert because this is not good and it is broken see.... 
https://github.com/pharo-project/pharo/issues/8412

PR #8206, opened 2020-12-30
8169 - Compiled code 'literals' do not behave the same that compiled code  'hasLiteral'

### LINKED ISSUE #8169: CompiledCode >> #literals  do not behave the same that CompiledCode >> #hasLiteral:
**Describe the bug**
After the new change of blocks considered as literals, the behavior of the methods CompiledCode >> #literals and CompiledCode >> #hasLiteral: were supposed to be the same and giving all the literals of the methods (even if they are included in the block which is in the literals arrays.

As a consequence, the large image support is broken:
https://github.com/pharo-project/aleph/issues/3

**To Reproduce**
Execute:

```
Object compile: 'test	
	[Object new ] value. Array new ' classified: 'test'.
method := Object >> #test.
{ 
method literals includes: (Smalltalk  bindingOf: #Object).
method hasLiteral: (Smalltalk  bindingOf: #Object).
method literals includes: (Smalltalk  bindingOf: #Array).
method hasLiteral: (Smalltalk  bindingOf: #Array).
 }
```

**Expected behavior**
The return should be 4 times true. the first one is false.

**Version information:**
 - OS: Win10
 - Version: Pharo9.0.0
build: 997, commit: 5bf82e9

**Expected development cost**
time to think what should be the behavior and doing the PR + tests.

VincentBlondeau:
Add method 'literalsEvenTheOnesInTheInnerBlocks' that returns the literals also included in the inner blocks of the method.

Temporary fixes #8169.

### DIFF
--- a/src/Kernel-Tests/CompiledCodeTest.class.st
+++ b/src/Kernel-Tests/CompiledCodeTest.class.st
@@ -170,6 +170,36 @@ CompiledCodeTest >> testHasSelectorSpecialSelectorIndex [
 ]
 
 { #category : #'tests-literals' }
+CompiledCodeTest >> testLiteralsDoNotConsiderTheInnerBlockLiterals [
+
+	"Bug or feature , I do not know..."
+
+	| method block |
+	method := self class compiler compile: 'method 
+		<pragma: #pragma>
+		test := 1+2.
+		Class.
+		self doIt: [ 
+		test := 1 - 2.
+		test := #(arrayInBlock).
+		Object.
+		self name ].
+		^#(#array) '.
+	block := (self class compiler evaluate: '[ 
+		test := 1 - 2.
+		test := #(arrayInBlock).
+		Object.
+		self name ]') compiledBlock.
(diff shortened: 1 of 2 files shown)


VincentBlondeau:
Failures are not related

MarcusDenker:
This is a dark part of the system.. whatever we do, it will be wrong form some "perspective". I will have a look

VincentBlondeau:
Maybe a bright perspective and a dark one could cohesist?
Let me know if you need sth.

MarcusDenker:
I think this is a good next step. It is tempting to just think that #literals can be the "high level" method that returns the literals of the blocks, too. But that is problematic, as we then have no way of just getting the literals of the object that represents the low level compiledMethod or compiledBlock.

This compiledMethod *does not* have these literals, so #literals returning them would be not good.

So for me adding a method is a good intermediate solution. 

It is true #hasLiteral: is currently out of sync.

MarcusDenker (approved):

PR #8217, opened 2021-01-02
Fixed #7967. Renamed the method Integer >> take:, wrote a better comment and better variable names

### LINKED ISSUE #7967: Integer >> take: needs a better method name, better comment, and better variable names
I think that `Integer >> take:` should be given a more meaningful name (e.g. `numberOfCombinationsTaken:` or `binomialCoefficientsWith:`).

Also, the implementation of `Integer >> take:` is a bit unclear:
- comment is hard to read
- names `kk`, `num`, `denom`, `factor` could be better

```Smalltalk
Integer >> take: kk
	"Return the number of combinations of (self) elements taken kk at a time.  
	For 6 take 3, this is 6*5*4 / (1*2*3).  Zero outside of Pascal's triangle.  Use a trick to go faster."
	" 6 take: 3  "

	| num denom |
	kk < 0 ifTrue: [^ 0].
	kk > self ifTrue: [^ 0].
	num := 1.
	self to: (kk max: self-kk) + 1 by: -1 do: [:factor | num := num * factor].
	denom := 1.
	1 to: (kk min: self-kk) do: [:factor | denom := denom * factor].
	^ num // denom
```

olekscode:
…ent and better variable names

### DIFF
--- a/src/Math-Operations-Extensions/Integer.extension.st
+++ b/src/Math-Operations-Extensions/Integer.extension.st
@@ -430,6 +430,31 @@ Integer >> nthRootTruncated: aPositiveInteger [
 ]
 
 { #category : #'*Math-Operations-Extensions' }
+Integer >> numberOfCombinationsTaken: k [
+	"Return the number of combinations of (self) elements taken k at a time. 
+	It is calculated as C(n,k) = n! / (k! (n-k)!) 
+	For 6 numberOfCombinationsTaken: 3, this is 6*5*4 / (1*2*3)"
+	
+	"6 numberOfCombinationsTaken: 3 >>> 20"
+
+	| numerator denominator |
+	k < 0 ifTrue: [^ 0].
+	k > self ifTrue: [^ 0].
+	
+	numerator := 1.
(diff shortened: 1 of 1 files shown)


Ducasse:
Thanks Oleksandr I will merge it. 
Now 
- would be nice to have a test.
- I was wondering about the numberOfCombinationsTaken: k [
 	"Return the number of combinations of (self) elements taken k at a time. :

I was wondering if the name should not be 
numberOfCombinationsFor: k [
 	"Return the number of combinations of (self) elements taken k at a time. 

or 

numberOfCombinationsTaking: k [
 	"Return the number of combinations of (self) elements taken k at a time. 

Because when I read it, it feels like there is something missing taken: what.

olekscode:
Yes, perhaps "Taking" would be better

olekscode:
In fact, this function is a bit tricky to name.
In mathematics it's a two-argument function C(n, k)
And sometimes it is semantically hard to get from C(n, k) to (n C: k) or (k C: n)

PR #8221, opened 2021-01-03
Refactor ChangeSetClassChangesTest to use ClassFactoryForTestCase.

### LINKED ISSUE #8097: ChangeSetClassChangesTest should use ClassFactoryForTest
and avoid to have code like that!

```
testChangeClassCategoryAddsNewChangeRecord
	"Changing the class category for a class should result in a change
	record being updated in the current change set."

	"At the start of this test, JunkClass should not exist, and there should be
	no change records pertaining to it in the change set."

	self deny: (self class environment hasClassNamed: 'JunkClass').
	self assert: (ChangeSet current changeRecorderFor: (testingEnvironment classNamed: #JunkClass)) thisName equals: 'nil'.	"Remove bogus change records created as side effect of preceding assert"
	ChangeSet current removeClassChanges: 'nil'.	"Define a class and save its definition"
	Object
		subclass: #JunkClass
		instanceVariableNames: 'zzz'
		classVariableNames: ''
		poolDictionaries: ''
		package: 'DeleteMe-1'.	"Forget about JunkClass in the change set"
	ChangeSet current removeClassChanges: 'JunkClass'.	"Redefine the class, changing only the class category"
	Object
		subclass: #JunkClass
		instanceVariableNames: 'zzz'
		classVariableNames: ''
		poolDictionaries: ''
		package: 'DeleteMe-2'.	"A change record should now exist in the change set"
	self
		assert:
			(self
				isDefinition: (ChangeSet current changeRecorderFor: (testingEnvironment classNamed: #JunkClass)) priorDefinition
				equivalentTo:
					'Object subclass: #JunkClass
	instanceVariableNames: ''zzz ''
	classVariableNames: ''''
	poolDictionaries: ''''
	category: ''DeleteMe-2''')

```

gcorriga:
Enhance ClassFactoryForTestCase with logic to redefine an already defined class.

Fixes #8097.

### DIFF
--- a/src/SUnit-Core/ClassFactoryForTestCase.class.st
+++ b/src/SUnit-Core/ClassFactoryForTestCase.class.st
@@ -183,6 +183,15 @@ ClassFactoryForTestCase >> newClassUsing: aTraitComposition [
 ]
 
 { #category : #creating }
+ClassFactoryForTestCase >> newClassWithInstanceVariableNames: aString [
+
+	^ self 
+		newSubclassOf: self defaultSuperclass
+		instanceVariableNames: aString
+		classVariableNames: ''
+]
+
+{ #category : #creating }
 ClassFactoryForTestCase >> newSubclassOf: aClass instanceVariableNames: ivNamesString classVariableNames: classVarsString [
 
 	^ self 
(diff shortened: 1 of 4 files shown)


Ducasse:
Thanks I will have a look. This is basically impossible to follow the change on github :)

Ducasse:
Hi giovanni

Are you sure that the changeset is not poluted by the test execution. 
the new teardown
```
"running"
tearDown
	factory cleanUp.
	ChangeSet newChanges:	changeSet.
	super tearDown
```
Does not look like removing the old change set. Or may be I got wrong this newChanges: 
I saw that you kept the current changeset so this is probably working. 
Now changeSet is one of these old classes.

gcorriga:
I've reworked the logic so that #tearDown removes only the change set installed in the #setUp method.

PR #8192, opened 2020-12-27
8189-testNoUnusedInstanceVariablesLeft-is-skipped

Ducasse:
fixes: 8189 - well a part of it.

PR #8228, opened 2021-01-04
Remove unused ivar selectedItem in deprecated TreePresenter

### LINKED ISSUE #8226: Remove unused ivar selectedItem in deprecated TreePresenter
a) we could either ignore deprecated classes when we test for unused ivars in image
b) or we could clean them too

I guess b) is the better option as we only have ivar selectedItem in deprecated TreePresenter to fix.

astares:
Fix #8226

### DIFF
--- a/src/Spec-Core/TreePresenter.class.st
+++ b/src/Spec-Core/TreePresenter.class.st
@@ -9,7 +9,6 @@ Class {
 		'menuBlockHolder',
 		'rootsHolder',
 		'selectionHolder',
-		'selectedItem',
 		'autoMultiSelection',
 		'columnInset',
 		'doubleClick',


astares:
CI problems (PotentialOutDatedDependencyWarning) not caused by this PR. 

Any reviewer/approvers?

Ducasse:
I do not know what is the context serialisation test I imagine that it is not related.

Ducasse:
Hi torsten did you see this little helper I did https://github.com/pharo-project/pharo/pull/8192
Because I wanted to get the class and the iv. 
Now I was surprised that my PR broke the build.

astares:
@Ducasse Yes - I saw the helper. But not all the cases could be fixed by just removing the unused ivars...

PR #8240, opened 2021-01-04
8168-CompiledMethodreferencedClasses-should-not-take-class-variables-into-account 

### LINKED ISSUE #8168: CompiledMethod>>#referencedClasses should not take class variables into account
`CompiledMethod>>#referencedClasses ` also takes into account the values of class variables. Ideally it should not as the results can change depending on the value of class variables. 

This creates a class with a class variable

```
Object subclass: #AATest
	instanceVariableNames: ''
	classVariableNames: 'VarA'
	package: 'AST-Core-Tokens'.
	
#AATest asClass class compile: 'initializeVar
	VarA := Object'.
	
#AATest asClass class compile: 'varA
	^ VarA'.
```

The method `#varA` has no referenced classes
```
(AATest class >> #varA) referencedClasses .  "an IdentitySet()"
```

Now we initialize the variable to point to a class.
```
AATest initializeVar.
```

If we call now `referencedClasses` we get a dependency to `Object` just because it is stored inside the class variable.
```
(AATest class >> #varA) referencedClasses. "an IdentitySet(Object)" .
```

Happens in Pharo 8 / Pharo 9

MarcusDenker:
fixes #8168 by only taking globals into account.

### DIFF
--- a/src/Kernel-Tests-Extended/ClassTest.class.st
+++ b/src/Kernel-Tests-Extended/ClassTest.class.st
@@ -421,7 +421,11 @@ ClassTest >> testReferencedClasses [
 	}
 		do: [ :assoc | 
 			self assert: assoc key referencedClasses notEmpty.
-			self assert: (assoc key referencedClasses asSet includesAll: assoc value asSet)]
+			self assert: (assoc key referencedClasses asSet includesAll: assoc value asSet)].
+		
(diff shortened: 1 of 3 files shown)


chisandrei:
The change looks good. Indeed #referencedClasses takes now only global classes into account.

PR #8252, opened 2021-01-04
8242 has package matching extension name uses symbol concatenation

MarcusDenker:
This removes the string concatenation in both methods using the idea to check directly


MarcusDenker:
very odd.. the tests do not fail in the image after doing the change, but in an image bootrapped with the change. Need to check why

MarcusDenker:
I will redo this for Pharo10

Issue #8204, opened 2020-12-30
MCPackageLoader>>#basicLoadDefinitions loads classes twice


chisandrei:
The method `MCPackageLoader>>#basicLoadDefinitions` loads classes twice. This is mentioned in its comment. However, the fact that classes is loaded twice leads to a significant slowdown especially in code that uses traits. 

<img width="910" alt="Screenshot 2020-12-30 at 15 40 50" src="https://user-images.githubusercontent.com/3534711/103358565-27f80100-4ab6-11eb-99bc-c9b6458c1116.png">

The screenshot above shows a profiler on loading GT.  `loadClassDefinition:` which is the first pass to load classes and traits takes 7.3% of the time. Then `tryToLoad:` which should load methods also loads again classes.  This time loading the same class definitions as before takes 30% of the time!!! 

<img width="1026" alt="Screenshot 2020-12-30 at 15 50 40" src="https://user-images.githubusercontent.com/3534711/103358812-ce440680-4ab6-11eb-87ae-bfb03f6a463a.png">

If we expand we see that most time is spend in the `TraitBuilderEnhancher`. For example when migrating class instances `Behaviour class >>#allInstances` is now called a lot, taking a significant amount of time. 

I'm not sure if there are any side-effects but loading also class definitions while loading method definitions should not be necessary. One maybe naive solution is to have `tryToLoad: ` as below. It does not attempt to load again class definitions.

```
tryToLoad: aDefinition
	aDefinition isClassDefinition ifTrue: [ ^ false ].
		
	[aDefinition addMethodAdditionTo: methodAdditions] on: Error do: [errorDefinitions add: aDefinition].'
```

I tried the above solution on loading the following code in a Pharo 9 image:
```
Time millisecondsToRun: [ 
	Metacello new
        repository: 'github://j-brant/smacc';
        baseline: 'SmaCC';
        load.

	Metacello new
        repository: 'github://PolyMathOrg/PolyMath';
        baseline: 'PolyMath';
        load.
	
	Metacello new
 		baseline:'Seaside3';
 		repository: 'github://SeasideSt/Seaside:master/repository';
 		load.
		
	Metacello new
    	baseline: 'PetitParser2';
    	repository: 'github://kursjan/petitparser2';
    	load.
	
	Metacello new
  		baseline: 'Moose';
  		repository: 'github://moosetechnology/Moose:development/src';
  		load.
 ] 
```

[...]

PR #8282, opened 2021-01-05
8204-MCPackageLoaderbasicLoadDefinitions-loads-classes-twice 

MarcusDenker:
As suggested in #8204, do not load the class definition twice


MarcusDenker:
I will redo this for Pharo10

PR #8286, opened 2021-01-05
fixing-eval-in-interactive-session

tesonep:
Doing a clean up in the detection of the --interactive image parameter.


astares:
## Bug description: copied from Discord discussion

astares:
I try on Linux (here Linux Mint)
```
mkdir test1
cd test1
wget -O- https://get.pharo.org/64/90+vm | bash
./pharo-ui Pharo.image eval "1 inspect"
```
I get a regular Pharo 9 with VM which is evaluating someting (here a simple inspect). But it is is hard to key in text in a Playground as backspace and other things are not working with this VM. Same when I use https://get.pharo.org/64/90+vmLatest. Same problem as was discussed on <http://forum.world.st/Can-anyone-get-recent-Pharo9-builds-to-work-on-Linux-tt5125641.html>

@estebanlm#7458 suggested last week to fetch the latest headless VM from Zero conf, so I did:
```
mkdir test2
cd test2
wget -O- https://get.pharo.org/64/90+vmHeadlessLatest | bash
./pharo-ui Pharo.image eval "1 inspect"
```
which opens the UI and really solves the issue in the playground. But this combination seems to totally **ignore the command line handler eval** expression of the command line ... no inspector is open. Usually I use this to instruct the image to do something (like loading a project or other).

<blockquote></blockquote>
<blockquote><div><strong><a href="http://forum.world.st/Can-anyone-get-recent-Pharo9-builds-to-work-on-Linux-tt5125641.html">Pharo Smalltalk Developers - Can anyone get recent Pharo9 builds to work on Linux? | Threaded View</a></strong></div></blockquote>

astares (approved):
Works for me

VincentBlondeau:
Linked to https://github.com/pharo-project/opensmalltalk-vm/issues/123

VincentBlondeau:
The VM passes the --interactive parameter in with an add at last strategy (see: vm_parameter_vector_insert_from (in parameterVector.c )).

The command line handler should be able to catch this as last argument.

tesonep:
Fix https://github.com/pharo-project/opensmalltalk-vm/issues/123

tesonep:
Now it can handle in any position, and it is ignored when processing the command line handlers.

PR #8297, opened 2021-01-05

### LINKED ISSUE #2586: Deleting a class with references gives extra "Do you want to delete anyway" msg

gcorriga:

1) if the class has no references, subclasses, or users, perform a refactoring to remove the class
2) if the class has references or users, ask the user what to do - browse the references/users, or do nothing.
3) if the class has subclasses, ask the user what to do - delete keeping subclasses, browse the subclasses, or do nothing.

Fixes #2586.


src/Calypso-SystemTools-Core/ClyBrowserMorph.extension.st
```diff
@@ -105,12 +87,61 @@ ClyBrowserMorph >> confirmUnusedVariablesInDefiningClass: variables [
...
+	 | hasReferences hasSubclasses hasUsers |
+	 hasReferences := self hasReferencesToClasses: classes.
+	 hasSubclasses := ((classes flatCollect: [ :each | each subclasses ]) 
+		                   copyWithoutAll: classes) notEmpty.
+	 hasUsers := ((classes flatCollect: [ :each | each users ]) 
+		              copyWithoutAll: classes) notEmpty.
+
+	 hasReferences | hasSubclasses | hasUsers ifFalse: [ 
+		 ^ {SycSilentlyRemoveClassStrategy new} ].
+
```
dionisiydk, line 120: Cool that you follow the method removal approach. 
Now with your nice changes and tests I think this logic for references analysis can be moved to SycRemoveClassStrategy (and same refactoring for method removal logic).
(some thoughts for possible improvements)

src/Calypso-SystemTools-Core/ClyBrowserMorph.extension.st
```diff
@@ -105,12 +87,61 @@ ClyBrowserMorph >> confirmUnusedVariablesInDefiningClass: variables [
...
+{ #category : #'*Calypso-SystemTools-Core' }
+ClyBrowserMorph >> hasReferencesToClasses: classes [
+	
+	| query references |
+	
+	query := (ClyClassReferencesQuery toAny: classes from: self systemScope).
+	references := query execute.	
+	^(references notEmpty and: [
+		((references items collect: [:each | each origin])
+			difference: (classes flatCollect: [ :each | { each . each classSide } ])) notEmpty])
```
dionisiydk, line 99: I also thought to extract this logic into ClyClassReferencesQuery itself (or special subclass). 
- query ignoreSelfReferences
- ClyExternalClassReferencesQuery

dionisiydk (approved):
Good stuff but I did not play with live image

MarcusDenker:
somehow the image can not be build:

  ClyBrowserMorphTest
  ClyBrowserMorphTest>>#testHasReferencesToClass
  ClyBrowserMorphTest>>#testRequestRemoveClassStrategyForParentClass
  ClyBrowserMorphTest>>#testRequestRemoveClassStrategyForReferencedClass
  ClyBrowserMorphTest>>#testRequestRemoveClassStrategyForUnreferencedClass
  ClyBrowserMorphTest>>#testRequestRemoveClassStrategyForUsedTrait

gcorriga:
It looks like I got the dependency order mixed up. I'll look into fixing that.

gcorriga:
The tests above pass because the test class causing issue seems to be missing.

Cleanest thing to do is to close the PRs and submit again.

PR #8305, opened 2021-01-07

### LINKED ISSUE #2586: Deleting a class with references gives extra "Do you want to delete anyway" msg

gcorriga:

if the class has no references, subclasses, or users, perform a refactoring to remove the class
if the class has references or users, ask the user what to do - browse the references/users, or do nothing.
if the class has subclasses, ask the user what to do - delete keeping subclasses, browse the subclasses, or do nothing.

Fixes #2586.
New version of #8297.


gcorriga:
The issues seem related to defining new test packages.

1) how can I replicate the problem
2) what's the standard solution in these cases?

guillep:
Hi @gcorriga sorry for coming late here.
There seem to be merge conflicts now.

Also, would you mind explaining what you mean by the following? I do not see any conversation, do you mean the tests failures?

> The issues seem related to defining new test packages.
> 
> 1. how can I replicate the problem
> 2. what's the standard solution in these cases?

gcorriga:
@guillep the comment was referring to some spurious errors I was seeing when running the tests. If I remember correctly the errors were due to the fact that I was reinstating some packages which were previously included in the iceberg package definition.
This was causing some old tests that had not been deleted to be pulled from the repo when merging the code.

gcorriga:
While trying to fix the merge issues, it looks like I've turned this merge request into a huge one with 900+ commits.

I'll close this PR and submit a clean one.

PR #8307, opened 2021-01-07
SycRemoveMethodStrategy refactoring

gcorriga:
Refactoring the logic to select the SycRemoveMethodStrategy when removing methods using a ClyBrowserMorph.


src/Calypso-SystemQueries/ClyClassExternalReferencesQuery.class.st
```diff
@@ -0,0 +1,14 @@
...
+}
+
+{ #category : #execution }
+ClyClassExternalReferencesQuery >> buildResult: aQueryResult [
+	| toExclude |
+	
+	super buildResult: aQueryResult.
+	toExclude := self variableQuery resultItems flatCollect: [ :each | { each . each classSide } ].
+	aQueryResult items removeAllSuchThat: [ :each | toExclude includes: each origin ]
+]
```
dionisiydk, line 14: Another method #selectMethod: should also use new filter logic. Add test to see that it does not work correctly with this PR.

But full implementation can be simplified by overriding the internal filter method:
```Smalltalk
 doesMethod: aMethod useVar: aVariable 
        "aVariable is a GlobalVariable instance in this query"
        aMethod origin ~= aVariable value and: [super doesMethod: aMethod useVar: aVariable]
```
This way all parts of query API will follow new logic.

Now it looks like this filter method deserves a better name. Something like #selectsMethod:referencingVar:

src/Calypso-SystemQueries/ClyMessageExternalSendersQuery.class.st
```diff
@@ -0,0 +1,15 @@
...
+	#superclass : #ClyMessageSendersQuery,
+	#category : #'Calypso-SystemQueries-Queries'
+}
+
+{ #category : #execution }
+ClyMessageExternalSendersQuery >> buildResult: aQueryResult [
+
+	super buildResult: aQueryResult.
+	aQueryResult items removeAllSuchThat: [ :each | selectors includes: each selector ]
+]
```
dionisiydk, line 15: Same fix is needed here

src/Calypso-SystemTools-Core/ClyBrowserMorph.extension.st
```diff
@@ -111,26 +81,44 @@ ClyBrowserMorph >> isMethodSelected: aMethod [
...
-	plural := '' asPluralBasedOn: senderCount.
+	strategies := SycRemoveMethodStrategy createForBrowser: self methods: methods.
+	strategies size = 1 ifTrue: [ ^strategies first ].
 	
-	caption := selectors size > 1 
-		ifTrue: [ 'There are {1} sender{2}' format: {senderCount. plural}]
-		ifFalse: ['The message #{1} has {2} sender{3}' format: {selectors anyOne. senderCount. plural}].	
+	title := selectors size > 1 
+		ifTrue: [ 'Methods have senders']
+		ifFalse: ['Method #{1} has sender{s' format: {selectors anyOne}].
```
dionisiydk:
```suggestion
		ifFalse: ['Method #{1} has senders' format: {selectors anyOne}].	
```

src/Calypso-SystemTools-Core/SycRemoveMethodStrategy.extension.st
```diff
@@ -13,6 +13,18 @@ SycRemoveMethodStrategy class >> createAllForBrowser: aBrowser [
 	^result
 ]
 
+{ #category : #'*Calypso-SystemTools-Core' }
+SycRemoveMethodStrategy class >> createForBrowser: aBrowser methods: aCollection [
```
dionisiydk: I would name this method a bit different:
```Smalltalk
SycRemoveMethodStrategy createAllForMethods: methods browser: browser
```
It would follow the name of method #createAllForBrowser:.

src/Calypso-SystemTools-QueryBrowser-Tests/MockClyBrowser.class.st
```diff
@@ -0,0 +1,34 @@
+Class {
+	#name : #MockClyBrowser,
```
dionisiydk: All Calypso classes are prefixed with Cly. Why mock should be different? 
I would use name ClyBrowserMock in that case

src/SystemCommands-ClassCommands/SycRemoveClassStrategy.class.st
```diff
@@ -0,0 +1,88 @@
...
+	#category : #'SystemCommands-ClassCommands'
+}
+
+{ #category : #testing }
+SycRemoveClassStrategy class >> canExecuteWithReferences: hasReferences subclasses: hasSubclasses users: hasUsers [
+	self subclassResponsibility 
+]
+
+{ #category : #'instance creation' }
+SycRemoveClassStrategy class >> createForBrowser: aBrowser classes: aCollection [
```
dionisiydk: And here I also think that more readable form would be:
```Smalltalk
SycRemoveClassStrategy createAllForClasses: classes browser: browser
```

src/SystemCommands-ClassCommands/SycRemoveClassStrategy.class.st
```diff
@@ -0,0 +1,88 @@
...
+		             copyWithoutAll: aCollection) notEmpty.
+
+	hasReferences | hasSubclasses | hasUsers ifFalse: [ 
+		^ { SycSilentlyRemoveClassStrategy new } ].
+
+	^ self
+		  createForBrowser: aBrowser
+		  references: hasReferences
+		  subclasses: hasSubclasses
+		  users: hasUsers
```
dionisiydk, line 40: And then this method can be #createAllForReferences:subclasses:users:browser:. 
(I don't really like so many arguments but I don't have any suggestion to fix that)

dionisiydk (approved):
All my notes can be addressed later if you prefer.

dionisiydk:
And another note. With this changes we lost information about senders count which I believe is very useful.

Could you add it to strategy descriptions? (as part of list item strings). 
For example "Remove, then browse 10 senders". 
The number of subclasses and trait users would be also very nice to see in class removal dialog.

MarcusDenker:
There are tests failing that are related

gcorriga:
right, let me have a look at the failures first and we can address @dionisiydk's points later.

gcorriga:
The failures are caused by my reinstating of the SystemCommands-MethodCommands-Tests package.

This has the side effect of reinstating some test classes which were not properly deleted from the repo. These tests fail since they require deprecated code.

SycRenameArgOrTempCommandTest
SycInlineTempCommandTest
SycConvertTempToinstVarCommandTest

I'll update the PR to delete these classes.

src/Calypso-SystemQueries/ClyClassExternalReferencesQuery.class.st
gcorriga: Not sure I got your suggestion here about #selectMethod: .

I'll look at refactoring the new query classes to use #doesMethod:useVar:, we can look at renaming it after that.

src/Calypso-SystemTools-QueryBrowser-Tests/MockClyBrowser.class.st
gcorriga: Ideally we should replace the class with a more general MockObject class - without necessarily going as far as including Mocketry or similar.

I'll rename it to ClyMockBrowser for now.

Ducasse:
Hi giovanni I want to integrate it but I see that there is a conflict.

gcorriga:
Hi @Ducasse, this should have now been brought in line with the current Pharo9.0 branch.

FYI - I didn't really need to fix any merge conflicts.

Ducasse:
Tx.
I think that sometimes git gets confused.

Ducasse:
I think that we waited too long :( I would close this issue for now.

PR #8315, opened 2021-01-08
8267 some pillar tests should be in a real test package

kasperosterbye:
fixes 8267

Moved tests to separate package, and added that package to baseline of GeneralTests


kasperosterbye:
It seems like the class comment in PRCodeSemanticActionTest.class.st is confusing confusing the system somehow. I am not sure it is a bug of the commit system, or if I did something wrong here.

kasperosterbye:
I somehow manages to confuse me and the system on this one - closing it now.

PR #8321, opened 2021-01-08
8214-Global-shortcuts-only-works-when-a-Pharo-window-is-focused

### LINKED ISSUE #8214: Global shortcuts only works when a Pharo window is focused
**Describe the bug**
Global shortcuts like CTRL + O + I, SHIFT + ENTER or CTRL + O + T only works when a Pharo window is focused or outside of the initial extent of the native window.
You need to give the focus to an open window to be able to perform it. Giving the focus to the World or the Windows Pharo native window is not enough

In addition, if the pointer is in the initial extent of the native window it is working. Meaning that if you give the focus to the world and go to the initial extent, it is working. In the green zone there:
image

**To Reproduce**
1. Go to pharo
2. Enlarge the native window
3. Click on the world (maybe twice to close the world menu)
4. Try the shortcut CTRL + o +i in the new extent zone -> does not work
5. Move the pointer to the initial extent
6. Try the shortcut CTRL + o +i in the new extent zone -> works

**Expected behavior**
works everywhere

**Version information:**
Pharo9.0.0
build: 1000, commit: f92e709
Windows 10

**Expected development cost**
Not too much

**Additional context**
Reproducible in vanilla Pharo image with the headless VM

tesonep:
- When the key events are created they should be created with a mouse position that is scaled to the scaled size of the world. If the world changes the scalling. When the position  is not correctly scalled the out-of-bounds detection is broken when checking against scalled bounds.

- The SDL Form renderer should use the size of the scalled form. If it is not done, the invalidation of the window is not correctly propagated.

Fix #8214


guillep:
@VincentBlondeau could you check with this fix?

VincentBlondeau:
Working well for me :) Issue fixed! Thanks!

Ducasse:
Hi Pablo we got a little conflict. I tried to see it but GH failed to display the code :( so I let you do it.

guillep:
Fun conflict :/. We saw with Pablo that that file was being recognized as binary by github, so it's probably related. We should do the merge locally.

guillep:
I'm fixing the merge

PR #8328, opened 2021-01-09
Unused ivars in SliderPresenter

### LINKED ISSUE #8327: Unused ivars in SliderPresenter
SliderPresenter has unused ivars (its deprecated but we need to fix it)

astares:
Fix #8327


Ducasse:
borken test is unrelated

PR #8331, opened 2021-01-10
8216-FluidClass-API-for-traits-

### LINKED ISSUE #8216: FluidClass API for traits
```
Object << #MyClass2
	uses: TMockWithComplexSlot;
	slots: {};
	sharedVariables: {};
	package: 'Athens-Examples'
```

Ducasse:
Fix: #8216 and add tests

### DIFF
--- a/src/FluidClassBuilder-Tests/FluidClassBuilderTest.class.st
+++ b/src/FluidClassBuilder-Tests/FluidClassBuilderTest.class.st
@@ -96,7 +96,7 @@ FluidClassBuilderTest >> testBuilderTraitsClass [
 	
 	builder uses: TViewModelMock class.
 	self assert: builder traitsToBuild equals: TViewModelMock class.
-	self shouldnt: [ builder fillShiftClassBuilder ] raise: MessageNotUnderstood 
+	builder fillShiftClassBuilder 
 ]
 
 { #category : #'tests - gathering' }
--- a/src/FluidClassBuilder-Tests/FluidClassBuilderTest.class.st
+++ b/src/FluidClassBuilder-Tests/FluidClassBuilderTest.class.st
@@ -102,8 +102,8 @@ FluidClassBuilderTest >> testBuilderTraitsClass [
 { #category : #'tests - gathering' }
 FluidClassBuilderTest >> testBuilderTraitsWithComposition [
 	
-	builder uses: TViewModelMock + TMockWithComplexSlot.
-	self assert: builder traitsToBuild equals: TViewModelMock + TMockWithComplexSlot.
+	builder uses: TViewModelMock + TViewModelMock2.
+	self assert: builder traitsToBuild equals: TViewModelMock + TViewModelMock2.
 	self shouldnt: [ builder fillShiftClassBuilder ] raise: MessageNotUnderstood 
 ]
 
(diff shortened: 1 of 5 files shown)


Ducasse:
@estebanlm Try it and let me know. 
I will now improve the slots default and layout of complex slots on the class definition

src/FluidClassBuilder-Tests/FluidClassBuilderTest.class.st
```diff
@@ -87,7 +87,24 @@ FluidClassBuilderTest >> testBuilderSharedVariables [
 FluidClassBuilderTest >> testBuilderTraits [
 	
 	builder uses: TViewModelMock.
-	self assert: builder traitsToBuild equals: TViewModelMock 
+	self assert: builder traitsToBuild equals: TViewModelMock.
+	self shouldnt: [ builder fillShiftClassBuilder ] raise: MessageNotUnderstood
```
dionisiydk: Why not allow test to simply fail? 
I am sure that good books describe such assertions as an anti pattern. 
 
With this style you will not get a normal debugger in case of DNU. And error log will miss a lot of contextual information (depends how shouldnot:raise: is implemented). Therefore it will complicate resolving the bug

Ducasse:
Got a slip in the PR src/Rubric/RubTextEditor.class.st  but after 25 min I could not find how to republish this PR without it :(

@Denis why not. For now I just want to make sure that I have an assertion and since I could not find a way to have an assertion I use shouldnt:raise: which to me is valid.

src/TraitsV2/MetaclassForTraits.class.st
```diff
@@ -53,6 +53,16 @@ MetaclassForTraits >> isClassTrait [
 	^ true
 ]
 
+{ #category : #testing }
+MetaclassForTraits >> isEmpty [ 
+	"Since we can have potential an empty array in the fluid definition (to be revisited) 
+	when the trait composition is aTrait it received the message isEmpty and would break
+	traitComposition 
+		ifNotNil: [ :t | t isEmpty ifFalse: [shiftClassBuilder traitComposition: traitComposition ]]."
+	^ false
```
dionisiydk, line 62: #isEmpty here looks really strange. It would be better to find different fix to the problem

src/TraitsV2/Trait.class.st
```diff
@@ -268,6 +268,17 @@ Trait >> isClassTrait [
...
 ]
 
+{ #category : #'as yet unclassified' }
+Trait >> isEmpty [ 
+	"Since we can have potential an empty array in the fluid definition (to be revisited) 
+	when the trait composition is aTrait it received the message isEmpty and would break
+	traitComposition 
+		ifNotNil: [ :t | t isEmpty ifFalse: [shiftClassBuilder traitComposition: traitComposition ]]."
+
+	^ false
```
dionisiydk, line 278: And here.

dionisiydk:
> @denis why not. For now I just want to make sure that I have an assertion and since I could not find a way to have an assertion I use shouldnt:raise: which to me is valid.

To me it is equivalent to using "self assert: true" to bypass the rule. it's not a real assertion and it complicates debugging if test fails.

src/TraitsV2/Trait.class.st
Ducasse: Like what? runtime checks? 
Right now I do not have the answer. First I want to know what is the argument of uses: because it can be a trait or a trait composition and I would prefer to have traits: with only one collection, now the problem is that T1 + T2 is a trait so it means that traits: argument would be collection with only one argument so it sucks.  
So this is a problem because there is no intersection between traits: and uses: arguments.
I do not like uses: because it looks strange. I would prefer trait: because at the end if we keep T1 + T2 then I get a single trait (composed) in any way I turn it around.
So may be the solution is to have trait: and remove uses: and in that case I will be able to remove isEmpty. 
But we should decide what do we want for the fluid syntax for trait.

PR #8333, opened 2021-01-10
8332-Better-and-nicer-display-of-slots

Ducasse:
Fixes: 8332 Nicer display of slots and complex slots


Ducasse:
Broken tests are unrelated.

Ducasse:
Broken tests are zinc tests. So unrelated.

PR #8335, opened 2021-01-11
Make temps visible in doIt scripts (debugIt command).

dionisiydk:
PR makes temps visible directly in all doIt scripts. 
Previously all temp users were converted to reflective calls like "ThisContext readTempVariable: #tempName". Now you will see normal variable names debugger (try debugIt command).

Screen Shot 2021-01-11 at 00 10 12

Implementation details:

1. It simplifies #asDoItContext: by removal the temp rewrite stage. 
With recently introduced DoItVariable  locals are now transparently visible during compilation of doIt method using requestor context.
In future it will be possible to remove ThisContext argument from debugger DoIt method . But now I keep it for compatibility (change will require NewTools update).
2. It introduces asReflectiveDoItContext: to keep previous logic with rewritten temps in all other places: mustBeBoolean machinary and conditional breakpoints.
These places are special because doIt's there are cached: they are executed over different contexts. In future we can improve them to use special kind of variables and get transparent variables in all places
3. It refactors semantics analyzer to not rely on #isLocalVariable message and use a dispatch to the variable instead. 
It is requried to allow DoItVariable represent external temps during compilation. Otherwise compiler is trying to call some locals specific code during doIt compilation and it fails (DoItVariable is not a local for newly compiled doIt methods).
4. By default Variable>>#asDoItVariable returns self. Only locals are converted to DoItVariable 
The idea is that only variables which can't be used directly in doIts needs to be converted. And it is only local variables.

### DIFF
--- a/src/AST-Core/RBValueNode.class.st
+++ b/src/AST-Core/RBValueNode.class.st
@@ -43,7 +43,7 @@ RBValueNode >> evaluate [
 { #category : #evaluating }
 RBValueNode >> evaluateForContext: aContext [
 	"evaluate the AST taking variables from the context"
-	^(self asDoitForContext: aContext) 
+	^(self asReflectiveDoItForContext: aContext) 
 		generateWithSource valueWithReceiver: aContext receiver arguments: {aContext}.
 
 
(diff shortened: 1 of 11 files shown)


dionisiydk:
tests are related. Needs more work

dionisiydk:
Close to avoid noise. I will reopen it when ready

PR #8336, opened 2021-01-11
fixes #8267

### LINKED ISSUE #8267: Some pillar tests should be in a real test package
classes from Pillar-ExporterRichText-Tests should be in a real test package so one is able to
load/unload to have a "production code only" image without any tests

kasperosterbye:
This is really trivial, so trivial that I am not sure if something else was asked for.
1) Promoted the Test-tag to a package
2) Included the newly created test package in the baseline GeneralTests


kasperosterbye:
removed dependency, and close this to try again.

PR #8337, opened 2021-01-11
Remap OS keys to use Home and not Begin

### LINKED ISSUE #8334: Home key (return to the start of the line) does not work
**Describe the bug**
Pressing the home key ("début" on french keyboard) does not bring the cursor to the start of the line as it should.

**Version information:**
 - OS: Windows 10. Azerty french keyboard

guillep:
Fix #8334 

@dupriezt could you check this fix?
To test, you can change the line I changed in the PR and then do `OSKeySymbols initialize`.


dupriezt:
> 
> 
> Fix #8334
> 
> @dupriezt could you check this fix?
> To test, you can change the line I changed in the PR and then do `OSKeySymbols initialize`.

It works. Thanks!

guillep:
Thanks for testing fast :)

PR #8338, opened 2021-01-11
Fixes8267 some pillar tests should be in a real test packag

kasperosterbye:
The tests are now in a package of their own.


MarcusDenker:
When I download the build artefact, the tests seems to be not there anymore (e.g PRCodeSemanticActionTest)

The problem is that the ne new package is not loaded by the bootrap, I think.

kasperosterbye:
I believe that to be true (it not being loaded). I find it in general very hard to figure out the overall architecture of the system, and in particular the boot/build. I think I just close this one, and find some other bug to fiddle.

PR #8339, opened 2021-01-11
improve-win32-platform-specific-implementations

tesonep:
- Implementing fromCString in a correct way, as it is using the encoder as UTF8
- Adding support for easing the command line testing.
- Improving the implementation of Win32WideString and its conversion to/from Pharo Strings.

### DIFF
--- a/src/RPackage-Core/RPackage.class.st
+++ b/src/RPackage-Core/RPackage.class.st
@@ -1522,17 +1522,6 @@ RPackage >> tagsForClasses [
 	^self classTags reject: [:each | each isRoot] thenCollect: [:each | each name]
 ]
 
-{ #category : #testing }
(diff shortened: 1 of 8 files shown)


src/RPackage-Core/RPackage.class.st
```diff
@@ -1522,6 +1522,17 @@ RPackage >> tagsForClasses [
 	^self classTags reject: [:each | each isRoot] thenCollect: [:each | each name]
 ]
 
+{ #category : #testing }
+RPackage >> testSuite [
```
guillep: I think this should be repackaged closer to the tests or the user, right?

tesonep: Yes, I will change

PR #8347, opened 2021-01-12
8174 st peel to first command looks wrong

### LINKED ISSUE #8174: StPeelToFirstCommand looks wrong
So far it is the same as copyStackToClipboard

execute
	self debugger copyStackToClipboard

StevenCostiou:
Fixes https://github.com/pharo-project/pharo/issues/8174


Ducasse:
Steven what is peel action? I never used it.

StevenCostiou:
When you manage to open a debugger on a running recursion (with cmd + .), peel to first like pops the stack to the first recursive call so you can see where it started. 

Now that I explain that, I think the test I wrote is not totally accurate, I should check.

Ducasse:
Thanks I will try to see because I'm not clear with myself. :)

PR #8351, opened 2021-01-12
Smalltalk-restartMethods

MarcusDenker:
This PR does two things:
- we add #restartMethods to SmalltalkImage
- implement a fall back in Behavior
- implement it in MorphicCoreUIManager and MorphicUIManager which spawn a new morphic process]

For now we call this in ImageCleaner>>#cleanupMethods

The goal of this PR is to make sure that the morphic main loop is restarted so that the method MorphicRenderLoop>>#doOneCycleWhile: on stack is in sync with the code.

Later we need to add #cleanupMethods to more classes to make sure that all methods are restarted.


MarcusDenker:
I checked the build artefact: with this restart, the method on the stack has now a valid sourcepoiner

PR #8357, opened 2021-01-13
fixes #8352. Basically it was already fixed

### LINKED ISSUE #8352: Inspecting a Text no longer shows the formatted text
**Describe the bug**
It used to be (before "newtools" I  believe) that `'An all bold text' asText allBold inspect` would bring up an inspector which would show the rendered text as first presentation.

**To Reproduce**
doing this in a playground: `'An all bold text' asText allBold inspect` 

**Screenshots**
Old behaviour:
image

New behaviour:
image

**Version information:**
 - The error manifests itself in Pharo 9, build 1030, and not in build 6 (sorry, it was just a random old image I had)

**Expected development cost**
I expect some method has to be implemented in class Text, with the right pragma. I just do not know which pragma and which method.

kasperosterbye:
Just had to copy two methods from String to Text.

The SpTextPresenter was used for Inspecting strings, and works for text too.

### DIFF
--- a/src/NewTools-Inspector-Extensions/Text.extension.st
+++ b/src/NewTools-Inspector-Extensions/Text.extension.st
@@ -1,0 +1,25 @@
+Extension { #name : #Text }
+
+{ #category : #'*NewTools-Inspector-Extensions' }
+Text >> inspectionText [
+	<inspectorPresentationOrder: -10 title: 'Text'>
+	
+	^ SpTextPresenter new
+		text: (self truncateWithElipsisTo: 10000);
+		yourself
+]
+
+{ #category : #'*NewTools-Inspector-Extensions' }
+Text >> truncateWithElipsisTo: maxLength [
+	"Return myself or a copy suitably shortened but with elipsis added"
+	"Copied from String"
+
+	^ self size <= maxLength
+		ifTrue:
+			[self]
+		ifFalse:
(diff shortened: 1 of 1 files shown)


kasperosterbye:
Unclear to me if this is me doing something wrong, or if there is a pipeline problem.

MarcusDenker:
It is too complex to contribute. This is one example: "but is has to be done in another repository". Even I can not give you a list for which projects this is true. We really need to improve that.

MarcusDenker:
So the reason why I did not yet merge is that Spec and newtools are "external" packages and one should commit to another repository.

kasperosterbye:
Ahh - I see.
There is also something else at play here, namely that once things are integrated into Pharo (not in their original repo anymore), documentation is often lost. In the original repo there might have been readme files, but those are not moved into Pharo (and it is not clear where they should go).
I'll leave this one open.

MarcusDenker:
I will do a commit to the newtools repo later today...

MarcusDenker:
I propose to not add truncateWithElipsisTo: as an extension for newtools, but instead add it to Text (as we have it on String already).

kasperosterbye:
Yes, that sounds right. I was wondering if one should inline it in the method itself, as it is unlikely to be used for any other purpose

PR #8359, opened 2021-01-13
Using rub text decorators to show the next instruction to execute

StevenCostiou:
This is experimental. This has also been pushed to NewTools, but if it triggers instability, then revert the PR.

Screenshot 2021-01-13 at 09 56 27

### DIFF
--- a/src/NewTools-Debugger-Tests/StDebuggerTest.class.st
+++ b/src/NewTools-Debugger-Tests/StDebuggerTest.class.st
@@ -161,14 +161,25 @@ StDebuggerTest >> testCodeChangeAfterContextChange [
 { #category : #'tests - code pane' }
 StDebuggerTest >> testCodeSelectionAfterStepping [
 
-	| dbg |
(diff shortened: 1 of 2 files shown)


MarcusDenker:
StDebuggerTest>>testCodeSelectionAfterStepping is failing on the CI

dionisiydk:
This is cool. 
Did you look how to support tooltips in addition? We can show variable values this way.

StevenCostiou:
> This is cool.
> Did you look how to support tooltips in addition? We can show variable values this way.

Actually yes and I had a basic prototype last year. And I put pressure on Esteban to have the necessary API to capture mouse pointer in the debugger presenter, but I did not do anything since. 

Now that we can have presenters in tooltips, I think we could easily have a raw inspection when hovering a variable (maybe when pressing cmd or ctrl key I do not know).

StevenCostiou:
Do not merge yet, there is a blind spot in the test I need to fix it.

MarcusDenker (approved):

Issue #8341, opened 2021-01-11
Some methods have Invalid source pointers


pavel-krivanek:
**Bug description**
During bootstrap, we condense changes file. It produces big sources file and almost empty changes file. This process does not take care of source pointers of compiled methods that are present in the stacks of active processes, so their source pointer gets invalid.

If then the changes file reaches the size when the source pointer is valid again, it thinks that the real source can be found there it  tries to parse that location. But this location is, of course, random, so it finds only invalid junk and a try to get an AST raises a syntax error. We condense sources file with the Morphic world already installed, so the UI process stack source links are corrupted. When there is a general error and the debugger is supposed to be displayed, it raises the syntax error as described so a very random syntax error without real sense is displayed. The syntax error window restarts the UI process, so this happens the only very first time. The next errors display normal debugger.

Restarting of the UI process does not solve everything because there are other affected processes as well. 

**To Reproduce**
1. get fresh image
2. increase size of the changes file:
```smalltalk
s := String ofSize: 1000000.
100 timesRepeat: [ 
    ChangesLog default logChange: s
     ].
```
3. save image
4. start image again and evaluate something what raises error (like `1 asdf`)

Check:
```smalltalk
CompiledMethod allInstances
    select: [ :each | (SourceFiles fileIndexFromSourcePointer: each sourcePointer) > 1 ]
```

Thank's Marcus for the investigation

pavel-krivanek:
Related issue: #4967

guillep:
We can reproduce this problem in an image that crashed by printing  `thisContext stack`. We have investigated this and we found this issue is triggered since build 982, although we don't understand how it is related.

This issue is actually two issues: 
1- the source pointers are wrong in some methods
2- somehow we try to get the AST of those invalid methods when just printing the stack

Point nbr 2 is caused because printing a block is trying to get the AST and parse the source code.

![image](https://user-images.githubusercontent.com/708322/104340339-deb5a180-54f8-11eb-8f3f-1570dc54f1d8.png)

guillep:
Also, getting the source code of a block is protected by a `onErrorDo:` but this only catches errors, and not notifications. And since syntax errors raise `SyntaxErrorNotification`...

MarcusDenker:
I think I managed to reset the Morphic process with this PR: https://github.com/pharo-project/pharo/pull/8351

(the method with the wrong source pointer is not there anymore)

guillep:
[...]

PR #8360, opened 2021-01-13
PRRichTextComposer-resetStylers 

MarcusDenker:
PRRichTextComposer needs to reset the stylers as it holds on to methods that might be recompiled or have the sources condensed.

Another step for #8341


Ducasse:
Thanks marcus!

PR #8361, opened 2021-01-13
MenubarMorph-restartMethods 

MarcusDenker:
MenubarMorph holds onto lots of methods via the menu pragmas. This PR tries to fix it by calling #reset.

We need to check if that works for the CI build.

(This is another cleaning step for #8341)


MarcusDenker:
Sadly it does not work
wrong

PR #8367, opened 2021-01-13
Correctly throw a syntax error notification in unfinished statements

guillep:
When there is an unfinished statement, it correctly generates an error node in a faulty parser, but in the case of a strict parser it is not generating the corresponding exception.

 - add tests
 -  fix the issue by throwing the notification


tinchodias (approved):
Perfect, I loaded the PR and couldn't save the "unfinished statement" method, as expected.

In particular, I tried with the following code, which is similar to the code that made us discover the bug in the Roassal3 repo (reported [here](https://github.com/pharo-vcs/tonel/issues/93)):

```
method1

self assert: [  ] description: 'dasdasdas'.
self assert: [  ] description: 'dasdasdas' ].
self assert: [  ] description: 'dasdasdas'.
```

guillep:
Nice! If you check my tests I also test leading parenthesis and braces.

PR #8369, opened 2021-01-14
MenubarMorph-Improve

MarcusDenker:
a small improvement for MenubarMorph:
- menuBarItems are only used in #open, we do not need to hold on to them

I tried to keep the API  the same (that is, keep the setter for menuBarItems) in case this class is used for non-world menu uses.

(This is another cleaning step for #8341)


MarcusDenker:
does not work.

PR #8374, opened 2021-01-14
Minimal failing tests for peculiar ASTs where metalinks break

maenu:
I was trying to collect coverage of one of my packages by instrumenting every node with metalinks. Compared to the last time I touched metalinks, the time overhead has been greatly reduced. Instrumenting hundreds of methods and running more than one hundred examples is now possible within seconds, no longer minutes or hours *tip of the hat*. But I noticed certain ASTs for which metalinks, that should not change the result of the execution, do actually change the result of the execution (either by errors, or in some cases actually changing values on the stack). Searching my soup of nodes by types and binary search on hashes, allowed me to find the culprits. I tried reducing them to minimal code and added them as tests, which are currently failing. The tests instrument the problematic nodes only, then compares the result of the execution with and without instrumentation, as well as the side-effect of the metalink. I have no concrete idea on how to make the tests pass, Marcus probably has, just a few hints:
 - pragma nodes: no clue how metalinks on them make sense, but the raised error seems to be too late in the pipeline, I think it should fail way earlier, with a more actionable message
 - literal array nodes: not sure if the size or the nesting or the combination causes the problem
 - dynamic array nodes: seem to cause problems when used as arguments
 - ==, =, > message nodes: those seem to be weird edge cases, reducing my convoluted code to a minimal example was quite hard, maybe it has to do with special messages like size, or optimizations?

### DIFF
--- a/src/Reflectivity-Tests/ReflectiveMethodTest.class.st
+++ b/src/Reflectivity-Tests/ReflectiveMethodTest.class.st
@@ -152,6 +152,35 @@ ReflectiveMethodTest >> testSetLinkOnClassVariableAndUninstall [
 
 ]
 
+{ #category : #tests }
+ReflectiveMethodTest >> testSetLinkOnDynamicArrayArgument [
+
+	| link a b |
+	link := MetaLink new
+		        metaObject: #node;
+		        selector: #tagExecuted;
+		        yourself.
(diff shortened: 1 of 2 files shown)


MarcusDenker:
Thanks! I will have a look

MarcusDenker:
I now added the tests to my backlog, I will fix them one by one with new PRs

PR #8380, opened 2021-01-18
RB - Add message to kwon how many occurrences were replaced

### LINKED ISSUE #7948: Refactoring - Find and replace Bug
**Find and replace UI bug**

When the Find and replace refactoring is used and it does not find occurrences it does not give you any message

@Ducasse

Lin777:
Fixes #7948


src/Refactoring-Core/RBFindAndReplaceRefactoring.class.st
```diff
@@ -106,12 +108,18 @@ RBFindAndReplaceRefactoring >> findOccurrencesIn: rbMethod [
...
+	occurrences := occurrences + 1.
 	flag := true.]
                       onFailure: [  ] .
 						flag ifTrue: [ self findOccurrencesIn: (rbMethod modelClass methodFor: rbMethod selector).
 							^ self]]].
 ]
 
+{ #category : #initialization }
+RBFindAndReplaceRefactoring >> initialize [ 
+	occurrences := 0.
```
tesonep, line 121: Hi @Lin777 please add a call to 
```
super initialize
```
Just in case somebody initialize something in the superclass (Now or in the future)

src/SystemCommands-SourceCodeCommands/SycExtractMethodCommand.class.st
```diff
@@ -70,6 +71,15 @@ SycExtractMethodCommand >> setUpOptionToOverrideExistingMethodDuring: aRefactori
 		ref refactoringWarning: 'Method ', selector printString, ' will override method in ', class name]. 
 ]
 
+{ #category : #execution }
+SycExtractMethodCommand >> setUpOptionToSearchInHierarchyDuring: refactoring [
+
+	refactoring setOption: #searchInAllHierarchy toUse:  [ 
+		UIManager default
```
tesonep, line 78: For this PR is ok, but maybe it is a good idea to extract all calls to UIManager to another object. 
In this way we can handle it in a nicer way. 
I would love to have a better interaction with the user, maybe even explaining the consequences of the decision. Something like the repair window in Iceberg, that shows us what is going to do with one or other option.

src/Refactoring-Core/RBExtractMethodAndOccurrences.class.st
```diff
@@ -69,7 +69,7 @@ RBExtractMethodAndOccurrences >> findOccurrencesOf: aSelector [
 			model: self model 
 			find: aSelector
 			of: class
-			inAllHierarchy: true.
+			inAllHierarchy: (self options at: #searchInAllHierarchy) value.
```
tesonep, line 72: I know that this is like this before, but maybe we want to use a nicer object for the options.
Something that wraps a dictionary can be an initial start, but for example that we can send the message #searchInAllHierarchy and not just like a symbol. Using symbols seems nice but it creates kind of strange coupling between both parts of the system. Also, I have a question, why the #value message?

tesonep (changes requested):

src/Refactoring-Core/RBFindAndReplaceRefactoring.class.st
Lin777: Hi @tesonep, now understand why always call super initialize, even when that method is empty. Thanks

src/Refactoring-Core/RBExtractMethodAndOccurrences.class.st
Lin777: umm... ok, yeah we can improve that, for the moment NautilusRefactoring initialize many options, even overridden some options changed inside commands. I'll think how improve this. I call #value method to execute the block (of that option) and get a boolean from user.

src/SystemCommands-SourceCodeCommands/SycExtractMethodCommand.class.st
Lin777: I'll check how Iceberg do that to have some ideas of how improve user interactions in Refactorings

src/Refactoring-Core/RBExtractMethodAndOccurrences.class.st
Ducasse: Yes this is a bit the mess.

Ducasse:
I will tag the issue with more work needed so that we know the status. Let us know when you are done.

src/SystemCommands-SourceCodeCommands/SycExtractMethodCommand.class.st
Ducasse: We will have to understand how to express the flow. 
May be here we should have have a call to the driver and a driver configuration 
because this is what the option is doing but this interaction could be move to the driver.

Ducasse:
Pablo may be we can integrate this PR and we do another one to fix the problems we spotted

tesonep:
@Lin777 Can I integrate it and you can work in the following PR, as Stef suggested?

Lin777:
@tesonep of course, I'll work in those changes in following PRs

PR #8381, opened 2021-01-18
RB - Fix strings and symbols comparing when extract method

### LINKED ISSUE #4972: Extract method is not working ok with string/symbol literals
If you have a method including a symbol literal and already exists in the same class another method returning an string equivalent to it, the extract method will use the existing method which is incorrect because it can broke existing code.

For example:
 ```smalltalk
TestClass>>method1

  self doSomething: #symbol

TestClass>>symbolString

  ^ 'symbol'
```
if you select `#symbol` and use the extract method option the resulting code will be:
 ```smalltalk
TestClass>>method1

  self doSomething: self symbolString
```

Lin777:
Fixes #4972


src/AST-Core/RBLiteralValueNode.class.st
```diff
@@ -40,8 +40,9 @@ RBLiteralValueNode class >> value: aValue start: aPosition stop: anotherPosition
 { #category : #comparing }
 RBLiteralValueNode >> = anObject [ 
 	^ super = anObject 
-		and: [ self value = anObject value 
-		and: [ self value species = anObject value species ] ]
+		and: [ self value = anObject value
```
dionisiydk, line 43: I think #literalEqual: should be used here:
```Smalltalk
and: [ self value literalEqual: anObject value ]
```
And other conditions can be removed this way.

Also it would be nice to have a test describing this change.

PR #8382, opened 2021-01-18
RB - Permute parameters when add an argument

### LINKED ISSUE #5852: Add argument refactoring breaks argument order
image

Asks for a new selector, for example:

```
toto:
```
=>
```
newArgument:toto:
```

Then, a second dialog asks for the value of the new argument. For example, I set `nil`.
However, the new argument is used at the end, breaking all senders.

```
foo toto: 42.
```
=>
```
foo newArgument: 42 toto: nil
```

Maybe we need a smarter dialog to edit this refactoring. A single dialog that allows editing the new selector, and the list of arguments, plus sort them.

Lin777:
Fixes #5852


src/Refactoring-Core/RBAddParameterRefactoring.class.st
```diff
@@ -29,6 +29,15 @@ RBAddParameterRefactoring class >> addParameterToMethod: aSelector in: aClass ne
 		initializer: init
 ]
 
+{ #category : #'instance creation' }
+RBAddParameterRefactoring class >> addParameterToMethod: aSelector in: aClass newSelector: newSelector permutation: aColl initializer: init [ 
+	^self new addParameterToMethod: aSelector
```
Ducasse, line 26: I do not understand why this method does not call the next one and also why it does not need a model?

src/Refactoring-Core/RBAddParameterRefactoring.class.st
```diff
@@ -40,6 +49,18 @@ RBAddParameterRefactoring class >> model: aRBSmalltalk addParameterToMethod: aSe
 		yourself
 ]
 
+{ #category : #'instance creation' }
+RBAddParameterRefactoring class >> model: aRBSmalltalk addParameterToMethod: aSelector in: aClass newSelector: newSelector permutation: aColl initializer: init [ 
+	^(self new)
```
Ducasse: Parentheses not needed.

src/Refactoring-Core/RBAddParameterRefactoring.class.st
```diff
@@ -40,6 +49,18 @@ RBAddParameterRefactoring class >> model: aRBSmalltalk addParameterToMethod: aSe
...
+		model: aRBSmalltalk;
+		addParameterToMethod: aSelector
+			in: aClass
```
Ducasse: Tabs are strange

src/Refactoring-Tests-Core/RBAddParameterTest.class.st
```diff
@@ -46,6 +47,7 @@ RBAddParameterTest >> testAddParameterThatReferencesGlobalAndLiteral [
 		               addParameterToMethod: oldSelector
 		               in: RBRefactoryTestDataApp
 		               newSelector: newSelector
+						   permutation: #(1 0)
```
Ducasse: strange tabbing

Ducasse:
Evelyn I wanted to integrate but some of the tests are broken.

unix-64 / Tests-unix-64 / testSycAddMessageArgumentCommandIsProperlyStoreOn – Unix64.SystemCommands.RefactoringSupport.Tests.SycRefactoringStoreOnTest

Lin777:
@Ducasse I will correct it and let you know when everything works fine

src/Refactoring-Core/RBAddParameterRefactoring.class.st
Lin777: All refactorings have two ways to initialize an instance, one with a model, an another with a default model, the defaul model is defined just like that:

```
model

	^ model
		ifNil: [ model := ( RBClassModelFactory rbNamespace onEnvironment: self defaultEnvironment )
				name: self printString;
				yourself
			]
		ifNotNil: [ model ]
```

Lin777:
@Ducasse I already fixed the broken test, and the other comments

Ducasse:
Super I will check when the build is finished

PR #8385, opened 2021-01-18
RB - Driver proposed to split UI, Command and Refactorings

Lin777:
The refactoring commands are acoplated with the UI, so a class is needed to intermediate that (driver), with the aim that command only interacts with driver class


Ducasse:
We should create a new package Refactoring-UI and may be copy the Syc..PreviewRefactoring.
RB should not depend on calypso and on Spec2.

PR #8398, opened 2021-01-20
RB - Remove duplicated code of SycMessageDescription and SycMessageOriginHack classes

### LINKED ISSUE #8397: RB - Remove argument from source context don't recognize the selected argument in some cases
For instance:

I have a method #with:and:
I want remove a method argument from source code context of #someSenderMethod method
```
SomeClass >> someSenderMethod
^ self with: nil and: nil
```
I select the method, this show me a preview to select the argument that I want remove

imagen

I select the second argument
And when it procedes remove the first argument, not the second

Lin777:
- Now only uses RBMethodName not SycMessageDescription
- The class sended to refactoring is directly obtained of context, not using SycMessageOriginHack like before

- With this change I also found an issue in command of Remove argument , but this issue was fixed in this PR

Fixes #8397

@Ducasse


src/SystemCommands-MessageCommands/SycRemoveMessageArgumentCommand.class.st
```diff
@@ -64,16 +64,17 @@ SycRemoveMessageArgumentCommand >> defaultMenuItemName [
 
 { #category : #execution }
 SycRemoveMessageArgumentCommand >> prepareFullExecutionInContext: aToolContext [
+	| parseTree |
 	super prepareFullExecutionInContext: aToolContext.
-	
-	argumentName := originalMessage argumentNames size = 1 
-		ifTrue: [ originalMessage argumentNames first]
+	parseTree := class parseTreeFor: originalMessage selector.
```
dionisiydk, line 69: PR breaks message related refactorings called from source code menu.
Here for example "class parseTreeFor: originalMessage selector" will fail because #class will be from the selected method but the command is not about method but about the message.
You can change any message in the code editor: rename or change signature. This message does not need to be implemented by the selected class.
For example open browser on: 
```
AnnouncerTest>>testAnnounceClass
	self assert: (announcer announce: AnnouncementMockA) class equals: AnnouncementMockA
```
Put a cursor into #announce: and open Source Code menu (cmd+t). There will be a command "Remove argument". It will fail with this PR.

The Hack object in my implementation was to adopt **method** refactorings to **message** refactoring where you do not really select a method but a message and where there is no an implementor class. The proper change would be to really implement message based refactoring and make method refactoring use it. And maybe method refactorings can be removed at the end.

dionisiydk (changes requested):

src/SystemCommands-MessageCommands/SycRemoveMessageArgumentCommand.class.st
Lin777: I understand, I'll think how do that, thanks for the advice. 

By the moment I'll close this PR.

PR #8401, opened 2021-01-21
Mouse over event is sent multiple times

tesonep:
Mouse over should only be sent once. Now it is sent again in every loop.


guillep (approved):
Seems legit

PR #8426, opened 2021-01-26
8379-Debugger-Contextat-infinite-loop-when-pressing-Through-in-a-test-case-condition

### LINKED ISSUE #2410: Debugger step through mustBeBoolean case hangs
(from fogbugs)

Debugger hangs on step through block which internally analysed for bool result.
Try step through:
```Smalltalk
#(2) select: [:i| i]
```
Problem somehow related to mustBeBoolean error. It works normally with different internal errors:
```Smalltalk
#(2) select: [:i| i / 0]
```
Debugger fallback into #/ method at ZeroDivide statement. But with mustBeBoolean trick it not works.

### LINKED ISSUE #7394: Code simulator issue with Step through
**Describe the bug**
When debugging code with intensive error handling with Step through, the code simulator may end in infinite recursion because of a wrong context index access.

**To Reproduce**
File-in attached PdmDebuggerTest and follow instructions in PdmDebuggerTest>>#letsFail

[PdmDebuggerTest.zip](https://github.com/pharo-project/pharo/files/5298457/PdmDebuggerTest.zip)

The failure looks like this:
```
 Context>>at:
Context>>at:
Context>>at:
Context>>pop
Context>>send:super:numArgs:
Context(InstructionStream)>>interpretNext2ByteSistaV1Instruction:for:extA:extB:startPC:
Context(InstructionStream)>>interpretNextSistaV1InstructionFor:
EncoderForSistaV1 class>>interpretNextInstructionFor:in:
Context(InstructionStream)>>interpretNextInstructionFor:
Context>>step
Context>>stepToHome:
[ctxt stepToHome: aContext] in Process>>stepToHome:
Process>>evaluate:onBehalfOf:
Process>>stepToHome:
DebugSession>>stepThrough:
StDebuggerActionModel>>stepThrough:
StDebugger>>stepThrough
```

It probobly does one pop more than expected which leads to access to the Context at index 0 (SmallInteger). This raises an exception which, during processing, wants to probably access the Context again and the image ends in infinite recursion.

### LINKED ISSUE #8379: [Debugger] "Context>>at:" infinite loop when pressing Through in a test case condition
-- Maybe duplicate of https://github.com/pharo-project/pharo/issues/8146 --
**Describe the bug**
Debugger crash see below

**To Reproduce**
1. Execute:
`ReNoPrintStringInPrintOnRuleTest  compile: 'testHalt self halt'.`
2. Execute
`ReNoPrintStringInPrintOnRuleTest buildSuite run`
3. In the debugger scroll in the stack to the frame right after the "Undefined Object  >> Doit" which is "TestSuite >> run".
4. Click on 'Through'
5. Enjoy the firework in PharoDebug.log  [PharoDebug-1.log](https://github.com/pharo-project/pharo/files/5826564/PharoDebug-1.log)
6. Press Alt + . to see all the debug windows opening

**Expected behavior**
No 🎆 🎆 🎆 🎆 

**Screenshots**
m5S9EEY0QM

**Version information:**
Pharo9.0.0
build: 1053, commit: 563ebe6

**Expected development cost**

**Additional context**

tesonep:
When doing a jump in inside a method (in this case on:do: and ensure:) we need to guarantee that the return value of the method is pushed in the original context.

Fix #8379
Fix #2410 
Fix #7394

### DIFF
--- a/src/Debugging-Core/Context.extension.st
+++ b/src/Debugging-Core/Context.extension.st
@@ -245,38 +245,43 @@ Context class >> simulatePrimitiveNumber: num with: simulator [
 Context >> stepToHome: aContext [
 	"Resume self until the home of top context is aContext.  Top context may be a block context."
 
-	| home ctxt here error context |
+	| home ctxt here error wrapperContext |
 	here := thisContext.
-	ctxt := self step.
-	ctxt = self ifFalse: [ "Insert ensure and exception handler contexts under aSender"
-		error := nil.
-		context := aContext insertSender: (Context contextOn: UnhandledError do: [ :ex | 
+	error := nil.
+	
+	wrapperContext := aContext insertSender: (Context contextOn: Exception do: [ :ex | 
 				            error
 					            ifNil: [ 
-						            error := ex exception.
-						            ex resumeUnchecked: here jump ]
-					            ifNotNil: [ ex pass ] ]) ].
+						            			error := ex.
+												ex resumeUnchecked: here jump ]
+					            ifNotNil: [ 
+										ex pass ] ]).
 	home := aContext home.
-	home == ctxt home ifTrue: [ 
-		^ { 
-			  ctxt.
-			  nil } ].
 
-	[ 
-	ctxt := ctxt step.
-	error ifNotNil: [ "remove above inserted ensure and handler contexts"
-		context ifNotNil: [ aContext terminateTo: context sender ].
+	"A first Step to do at least one step"
+	ctxt := self step.
+
+	home == ctxt home ifTrue: [ 
 		^ { 
 			  ctxt.
 			  error } ].
 
-	home == ctxt home ] whileFalse: [ 
-		home isDead ifTrue: [ 
+	[  ctxt := ctxt step.
+		error ifNotNil: [ 
+		"remove above inserted ensure and handler contexts"
+		aContext terminateTo: wrapperContext sender.
 			^ { 
 				  ctxt.
-				  nil } ] ].
+				  error } ].
+
+		home == ctxt home ] whileFalse: [ 
+			home isDead ifTrue: [ 
+				^ { 
+					  ctxt.
+					  nil } ] ].
 	"remove above inserted ensure and handler contexts"
-	context ifNotNil: [ aContext terminateTo: context sender ].
+	aContext terminateTo: wrapperContext sender.
+	
 	^ { 
 		  ctxt.
 		  nil }
--- a/src/Debugging-Core/Context.extension.st
+++ b/src/Debugging-Core/Context.extension.st
@@ -253,6 +253,7 @@ Context >> stepToHome: aContext [
 				            error
 					            ifNil: [ 
 						            			error := ex.
+												ex inspect.
 												ex resumeUnchecked: here jump ]
 					            ifNotNil: [ 
 										ex pass ] ]).
@@ -262,6 +263,8 @@ Context >> stepToHome: aContext [
 	ctxt := self step.
 
 	home == ctxt home ifTrue: [ 
+		"remove above inserted ensure and handler contexts"
+		aContext terminateTo: wrapperContext sender.
 		^ { 
 			  ctxt.
 			  error } ].
--- a/src/Debugging-Core/Context.extension.st
+++ b/src/Debugging-Core/Context.extension.st
@@ -253,7 +253,6 @@ Context >> stepToHome: aContext [
 				            error
 					            ifNil: [ 
 						            			error := ex.
-												ex inspect.
 												ex resumeUnchecked: here jump ]
 					            ifNotNil: [ 
 										ex pass ] ]).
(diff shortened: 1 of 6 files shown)


StevenCostiou:
I understand the test but not the changes :'( 
Could you explain I want to understand this is important, could you explain in the conversation here why it is needed to simulate a return value?

Ducasse:
Pablo was explaining to me so I will try to repeat.
The problem is that when there is an exception that happens during the step through 
the exception was raised as UnhandledException and the step trhough logic could not catch it and opening another debugger. I know it is vague and probably wrong. Because in some cases it was simply freezing the image. Now this is why we will write test for instructionStream and after context and it will take us some time but at the end we will know :)

src/Debugging-Core/Context.extension.st
```diff
@@ -245,38 +245,46 @@ Context class >> simulatePrimitiveNumber: num with: simulator [
...
+	error := nil.
+	
+	wrapperContext := aContext insertSender: (Context contextOn: Exception do: [ :ex | 
 				            error
 					            ifNil: [ 
-						            error := ex exception.
-						            ex resumeUnchecked: here jump ]
-					            ifNotNil: [ ex pass ] ]) ].
+						            			error := ex.
+												ex inspect.
```
guillep: I think this inspect is a debugging leftover ;)

src/Debugging-Core/Context.extension.st
```diff
@@ -245,38 +245,46 @@ Context class >> simulatePrimitiveNumber: num with: simulator [
 Context >> stepToHome: aContext [
```
guillep, line 245: This is strange, why is this method an extension of the debugger?

src/Debugging-Core/Context.extension.st
```diff
@@ -245,38 +245,46 @@ Context class >> simulatePrimitiveNumber: num with: simulator [
...
-		context := aContext insertSender: (Context contextOn: UnhandledError do: [ :ex | 
+	error := nil.
+	
+	wrapperContext := aContext insertSender: (Context contextOn: Exception do: [ :ex | 
 				            error
 					            ifNil: [ 
-						            error := ex exception.
-						            ex resumeUnchecked: here jump ]
-					            ifNotNil: [ ex pass ] ]) ].
+						            			error := ex.
```
guillep, line 255: Before this was `ex exception` any idea of why the change?

src/Debugging-Core/Context.extension.st
tesonep: I remove it

src/Debugging-Core/Context.extension.st
tesonep: Because before it was an UnhandledError

StevenCostiou: Maybe because ex is an unhandled exception that is an exception refering to the exception that is not handled (and it has that accessor). 
Probably there they have different signaler contexts and the important one is the unhandled exception's context? (but why?)

src/Debugging-Core/Context.extension.st
tesonep: Some time ago we move it out of Context to make a smaller bootstraped image.
This code is only used by the debugger, it is not used by the exception handling mechanism.

src/Debugging-Core/Context.extension.st
StevenCostiou: ah sorry I commented and saw your answer after refreshing the page.
So now ex is not the unhandled error anymore?

tesonep:
When a context is suspended in a message send, it already moved the PC to the next bytecode to execute, but the return value of the send message is still not pushed in the stack. 
Using #jump in a Context just restarts the context in the next PC. 
If jump is used in a context that has been interrupted in a message send, a return value should be pushed to the stack. As in Pharo we have a return value after a message sent. 

If you check the implementation of jump, this is done for the context of the sender of the message #jump, as the jump will never return. So to leave a healthy context it push a nil. However, it cannot be done for the receiver of the message jump, as it will depends when the context has been suspended. The push only needs to be done if the context is suspended in a message send. If it is suspended in a back-jump it is not needed to push and will produce a crash to push one thing extra.

So, it depends of the users of jump that require this handling.

This is happening in 3 different situations: 

- #ContextOn:Do:, this is adding a wrapping context that is implementing a  on:do:, so we have a message send, and we need to push a return value (even if discarded, as the bytecode do a pop)
- #ContextEnsure: is exactly the same reason, as it is adding a #ensure: and it is doing the #jump inside the block, a push is needed.
- The third case is a consequence of having Context >> stepHome: working. Context >> stepHome: returns a pair of objects, the first one is the context where it stopped stepping, and the second is the exception if there was one. When Process >> stepHome: uses the method in Context gets the return and do the following: if there was no exception takes the stopping context and put it as the suspended context of the receiver, if there is an error the context that is put as suspended context is the signaler context of the exception. This looks nice, as the process is now suspended in the signaling of the exception. However, this context is in an invalid state, it cannot resume, because the signal message send where it is blocked has never returned, so a value has to be push as the return value of the #signal message.

Ducasse:
Pablo thanks for the explanation I will have to digest it. 
Do you think that we could add this text as a comment of the jump method?

tesonep:
In the comment we have something like this:

```
	Self MUST BE a top context (ie. a suspended context or an abandoned context that was jumped out of).  A top context already has its return value on its stack (see Interpreter>>primitiveSuspend and other suspending primitives).
```

In this case we were not respecting this requirement, that was the problem. We were jumping to a method that was not the top (because we were modifying it to add a send in the middle). This is the key of the problem. I will add some comment if this is not preserved.

tesonep:
I have improved the comments in jump and the three users that required to push an extra value in the stack.

PR #8430, opened 2021-01-27
Removing suspicious update code in the debugger inspector, and fixing variable update problem with recursive call

### LINKED ISSUE #7286: Debugger's botton panel doesn't update parameters when steping into a recursive call
**Describe the bug**
The bottom panel of the debugger does not update parameters when steping into a recursive call

**To Reproduce**
1. Call 'debug it' on a recursive function
2. Step into the recursive call
3. Temp variables will still show the parameters for the first call

**Expected behavior**
Temp variables should reflect the parameters of the call.

**Screenshots**
image

**Version information:**
 - OS: Windows 10
 - Version: 1904
 - Pharo-9.0.0+build.652.sha.713b9dbf032dfb9c09406e86323a21624bfa149f (64 Bit)

**Additional context**
The bytecode debugger's extension shows the correct temp variables as shown in the print.

StevenCostiou:
Mirror change pushed in NewTools.
Fixes #7286
Suspicious code has been removed, so it remains to be seen how the debugger bottom inspector behaves in the next days after that PR is merged.

### DIFF
--- a/src/NewTools-Debugger-Tests/StDebuggerInspectorTest.class.st
+++ b/src/NewTools-Debugger-Tests/StDebuggerInspectorTest.class.st
@@ -52,6 +52,21 @@ StDebuggerInspectorTest >> testSetDefaultLayoutSpec [
 ]
 
 { #category : #tests }
+StDebuggerInspectorTest >> testShouldUpdateContext [
+
+	| debugger |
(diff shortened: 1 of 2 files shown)


StevenCostiou:
Unrelated test failure

dupriezt (approved):
Looks good, and works.

PR #8433, opened 2021-01-27
remove some unused class variables 1

MarcusDenker:
remove some unused class variables


kasperosterbye:
I went through the removed class variables, and indeed they look like they are not needed. 
I did not verify these were the only unneeded class variables.

I vote for accepting this PR

PR #8436, opened 2021-01-28
added a comparisson to KeyboardKey

estebanlm:
because otherway `Character down alt` = `Character down alt` gives false (since they are not the same object).

### DIFF
--- a/src/System-Platforms-Tests/KeyboardKeyTest.class.st
+++ b/src/System-Platforms-Tests/KeyboardKeyTest.class.st
@@ -8,6 +8,16 @@ Class {
 }
 
 { #category : #tests }
+KeyboardKeyTest >> testEqual [ 
+
+	self 
+		assert: (KeyboardKey fromCharacter: Character arrowUp) 
+		equals: (KeyboardKey new 
+			withValue: 65362;
+			yourself)
+]
+
+{ #category : #tests }
 KeyboardKeyTest >> testNamed [
 	"These assertions have been written using data available in KeyboardKey class>>#initializeKeyTable"
 	self assert: (KeyboardKey named: 'SHIFT_L') equals: (KeyboardKey value: 65505).
--- a/src/System-Platforms/KeyboardKey.class.st
+++ b/src/System-Platforms/KeyboardKey.class.st
@@ -1088,6 +1088,19 @@ KeyboardKey class >> windowsVirtualKeyTable [
 	^WindowsVirtualKeyTable
 ]
 
+{ #category : #comparing }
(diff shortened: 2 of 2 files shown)


Ducasse:
Esteban please add tests!

Ducasse:
We should not integrate a change like this without tests. Sorry but we should eat our own dog food and I'm sure you agree.

estebanlm:
well, I can argue that I was fixing a problem that emerged because of something that *was* integrated without tests :)
But anyway, test it there.

estebanlm:
Still, this problem was shown after a fuel materialization that included a KeyboardKey. 
Maybe there is a problem in fuel, not in `KeyboardKey` (since KeyboardKey is more like a flyweight?)

guillep:
Keyboard Key is a flyweight and there are tests for equality, so the problem may be elsewhere

MarcusDenker:
should this be merged?

estebanlm:
nah (even if I do not think it would hurt)

estebanlm:
in fact, I will insist with this, since we have some entries like this: 
```Smalltalk
KeyboardKey class>>valueForUnixPlatform: aKeyValue

	^self unixVirtualKeyTable at: aKeyValue ifAbsent: [ (self basicNew withValue: aKeyValue) ] 
```
and then, comparison redefinition has sense (even if fuel needs to be adapted to use the flyweight).

guillep:
> in fact, I will insist with this, since we have some entries like this:
> 
> ```smalltalk
> KeyboardKey class>>valueForUnixPlatform: aKeyValue
> 
> 	^self unixVirtualKeyTable at: aKeyValue ifAbsent: [ (self basicNew withValue: aKeyValue) ] 
> ```
> 
> and then, comparison redefinition has sense (even if fuel needs to be adapted to use the flyweight).

Then in this case yes, it could make sense to add a couple of equality tests for cases of keys created outside the flyweight

estebanlm:
bump? 
@MarcusDenker, @guillep ?

guillep:
We need the equality test.

estebanlm:
it is there: 
```Smalltalk
KeyboardKeyTest >> testEqual [ 

	self 
		assert: (KeyboardKey fromCharacter: Character arrowUp) 
		equals: (KeyboardKey new 
			withValue: 65362;
			yourself)
]
```

estebanlm:
re-bump ?
@MarcusDenker, @guillep ?

MarcusDenker (approved):

guillep:
Sorry, did not see it!

PR #8438, opened 2021-01-28
Move-ArgumentNamesTest-to-Debugger-Model-Tests-Core

MarcusDenker:
ArgumentNamesTest was in Tools-Test-Debugger, move it to where the to where debugger model tests are: Debugger-Model-Tests-Core


dupriezt (changes requested):
Looks good.
I suggest adding a comment to Object>>#canonicalArgumentName to explain what it does:
*"Imagine you write a method using this object (self) as argument, I return how you would likely name that argument.
Used by DynamicMessageImplementor in the debugger to create a template method when an execution calls a method that does not exist."*

(#canonicalArgumentName is the method that is being tested by ArgumentNamesTest)

dupriezt (approved):

PR #8440, opened 2021-01-29
fixes #8416: SpStringTableColumn beEditable and acceptAction has not effect

### LINKED ISSUE #8416: SpStringTableColumn beEditable and acceptAction has not effect
The action specified by `SpStringTableColumn>>onAcceptEdition:` is not activated even if `beEditable` is called.

The problem is seen in that there is no senders for `SpStringTableColumn>>acceptEdition`.

The situation can/should be corrected in method 'SpMorphicTableCellBuilder>>visitStringColumn:'. 

It is a bit unclear (due to lacking implementation) what parameters the `acceptAction` block should accept. My guess is that culling the block with `item` and the newly typed text would be appropriate.

kasperosterbye:
There is one aspect of this I am less certain about, namely the change in `SpAbstractTextPresenter >> text:`. It fixes what I consider a bug, as the formatting (bold etc) is lost in the previous implementation. This keeps it.

### DIFF
--- a/src/Spec2-Core/SpAbstractTextPresenter.class.st
+++ b/src/Spec2-Core/SpAbstractTextPresenter.class.st
@@ -311,8 +311,18 @@ SpAbstractTextPresenter >> text [
 { #category : #api }
 SpAbstractTextPresenter >> text: aString [
 	"Set the text of the text presenter"
+	"Attempts to keep text attributes (formatting) of the field"
 
-	text := aString
+	| attributes newValue |
+	aString isText ifTrue: [ 
+		text := aString.
+		^ self ].
+	newValue := aString.
+	self text isText ifTrue: [ 
+		attributes := self text attributesAt: 1 forStyle: TextStyle default.
+		newValue := newValue asText.
+		attributes do: [ :attr | newValue addAttribute: attr ] ].
+	text := newValue
 ]
 
 { #category : #simulating }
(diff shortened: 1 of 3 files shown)


kasperosterbye:
As usual, I am not sure if the problem is mine or somewhere else

guillep:
@estebanlm maybe you would like to review this one?

src/Spec2-Core/SpAbstractTextPresenter.class.st
```diff
@@ -311,8 +311,18 @@ SpAbstractTextPresenter >> text [
 { #category : #api }
 SpAbstractTextPresenter >> text: aString [
 	"Set the text of the text presenter"
-
-	text := aString
+	"Attempts to keep text attributes (formatting) of the field"
```
estebanlm, line 314: this change is actually incorrect.
Spec is not Text aware nor it will be. An attempt to mix it here will just add a functionality that we do not support, and we will not support that way. 
I understand people may want to use a Text in Spec with Morphic (but Spec is agnostic of its backend). 
Instead, I suggest to use create a special presenter that is attached to the Text backend.

estebanlm:
I merged the edition changes in spec (and they will be available in the image later this week, after a new merge)

src/Spec2-Core/SpAbstractTextPresenter.class.st
kasperosterbye: Thanks Esteban. I understand where Spec is going. It is a reasonable choice to have some widgets which are String only, and some which do character based formatting. I was just not aware of which special presenter can use Text without morphic.

estebanlm: just to be clear, there is *not* such presenter right now, you need to create it :)

PR #8441, opened 2021-01-29
fixes #8434 SpStringTableColumn sorting is cumbersome to get right

### LINKED ISSUE #8434: SpStringTableColumn sorting is cumbersome to get right
I tried to make the `sortFunction:` in class `SpStringTableColumn` work, but get inconsistent errors. In this example

```smalltalk
classNameSort := [ :c1 :c2 | c1 name < c2 name ].
methodCountSort := [ :c1 :c2 | c1 methodDictionary size < c2 methodDictionary size ].
SpTablePresenter   new
	addColumn: ((SpStringTableColumn title: 'Name' evaluated: #name) 
			sortFunction: classNameSort);
	addColumn: ((SpStringTableColumn title: 'Methods' evaluated: [:c| c methodDictionary size]) 
		sortFunction: methodCountSort) ;
	items: Smalltalk globals allClasses;
	openWithSpec
```

I get an error (when pressing the sort icon in the column header) in method:

```smalltalk
FTDescendingSortingState>>executeSortingOn: aDataSource
	aDataSource sortWith: self context ascendingSortFunction reversed` 
```

Saying that a FullBlockClosure does not understand `reversed`. Alas, that method expects a sortFunction.

Trying to make the two blocks into sortFunctions as in:

```smalltalk
classNameSort := [ :c1 :c2 | c1 name < c2 name ] asSortFunction.
methodCountSort := [ :c1 :c2 | c1 methodDictionary size < c2 methodDictionary size ] asSortFunction.
....

```

This makes the method 

```CollatorBlockFunction>>value: anObject1 value: anObject2
	| result |
	result := self collate: anObject1 with: anObject2.
	^result <= 0
```

fail with `True does not understand <= 0`, which is because the block returns a boolean rather than threeway compare (-1,0,1).

Next attempt is to make the blocks use three way compares:

```smalltalk
classNameSort := [ :c1 :c2 | c1 name threeWayCompareTo:  c2 name ]  .
methodCountSort := [ :c1 :c2 | c1 methodDictionary size threeWayCompareTo:  c2 methodDictionary size ]  .
...
```

This makes the code break in `Array>>mergeFirst:middle:last:into:by:` because block returns an integer when a boolean was expected.

### The right argument to `sortFunction:`

Alas, there is one kind of argument which actually seems to work:

```smalltalk
classNameSort := [ :c1 :c2 | c1 name threeWayCompareTo:  c2 name ]  asSortFunction .
methodCountSort := [ :c1 :c2 | c1 methodDictionary size threeWayCompareTo:  c2 methodDictionary size ] asSortFunction .
```

### My suggestion is to do the following two changes:

1. Change the comment `SpStringTableColumn>>sortFunction: aSortFunction` to indicate that its argument must be a threway sortFunction.
2. Add a sumplementary `SpStringTableColumn>>compareFunction: aBlock` which should accept a boolean valued compare block.

kasperosterbye:
Added a possibility for using a boolean compare for sorting, added an example, and changed the comment of sortFunction:

### DIFF
--- a/src/Spec2-Core/SpStringTableColumn.class.st
+++ b/src/Spec2-Core/SpStringTableColumn.class.st
@@ -67,6 +67,15 @@ SpStringTableColumn >> beSortable [
 ]
 
 { #category : #api }
+SpStringTableColumn >> compareFunction: aBlock [
+	"sets the sorting of this column to use the two argument aBlock for comparing (aBlock should return a boolean) "
+	"it is often simpler to use a boolean compare than the threeway compare required by sortFunction: "
+
+	super sortFunction: [ :i1 :i2| (aBlock value: i1 value: i2) ifTrue:[-1] ifFalse:[1]] asSortFunction .
+	self isSortable: aBlock notNil
+]
+
+{ #category : #api }
 SpStringTableColumn >> displayBackgroundColor [
 
 	^ backgroundColorAction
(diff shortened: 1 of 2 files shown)


guillep:
@estebanlm maybe you would like to review this one?

estebanlm:
added directly to pharo-spec

PR #8446, opened 2021-01-29
Comment Interval>>#rangeIncludes: and more complete test

### LINKED ISSUE #8444: Interval >> rangeIncludes: should be private and have a better comment
**Describe the bug**

`Interval >> rangeIncludes:` checks only if the number is withing the upper and lower bounds of the interval.
However, in the presence of a step, this is not enough to check inclusion.
And there are many users in the image of this method, that we should review.

For example, the following expressions are true:
```
(1 to: 10 by: 2) rangeIncludes: 4.
(10 to: 1 by: -2) rangeIncludes: 3.
```

In some places rangeIncludes is used seemingly as an optimization when you know that the interval will have step = 1 or -1.
That should be **at least** properly documented in the method and the users.

Also, current tests are not testing the cases above.

cdlm:
...to show that values in between steps are still included in the interval's range.

fixes #8444.

### DIFF
--- a/src/Collections-Sequenceable-Tests/IntervalTest.class.st
+++ b/src/Collections-Sequenceable-Tests/IntervalTest.class.st
@@ -635,24 +635,16 @@ IntervalTest >> testPermutationsDo [
 
 { #category : #tests }
 IntervalTest >> testRangeIncludes [
-	self
-		assert: ((1 to: 10)
-				rangeIncludes: 3).
-	self
-		assert: ((1 to: 10 by: 2)
-				rangeIncludes: 3).
-	self
-		deny: ((1 to: 10)
-				rangeIncludes: 0).
-	self
-		deny: ((1 to: 10)
-				rangeIncludes: 11).
-	self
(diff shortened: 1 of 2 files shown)


guillep:
> fixes #8444.

But not entirely... I'd say users should be reviewed and it should be documented why they use and it is enough with this method (and not the more precise `includes:`)

cdlm:
I had a quick look and all uses in the base image concern text spans (step other than ±1 meaningless) of some sort: text editing, message keywords in RSAST.

Using `rangeIncludes:` when the interval represents a span avoids the loop from `Collection>>#includes:`… but surely `Interval` could do a bounds check when step is ±1 and fall back to the loop otherwise, and that would make `rangeIncludes:` mostly redundant.

cdlm:
I'll change the comment in the method to mention it's a faster but special use check; but it feels overkill to add a comment to each and every sender…

The real solution would be extracting a `Span` superclass (implicit unit step or no step-related behavior at all, just bounds) out of `Interval` (explicit step)

PR #8447, opened 2021-01-29
7889-GZipReadStream--fileIntoNewChangeSet-fullFileName

tesonep:
Removing unused GZip support in the drag and drop


MarcusDenker (approved):
ok

PR #8452, opened 2021-01-29
7987-Class-definition-should-not-be-evaluated-if-it-contains-a-syntax-error

### LINKED ISSUE #7987: Class definition should not be evaluated if it contains a syntax error
**Describe the bug**
The accepting of the class definition with broken syntax tries to evaluate the definition which causes a confusing DNU

**To Reproduce**
1. Go to to some class in Calypso, e.g. `RBPragmaNode`
2. remove the closing quotes from the end of instance variables definition so the definition now looks like: 
```
RBProgramNode subclass: #RBPragmaNode
	instanceVariableNames: 'selector keywordsPositions arguments left right
	classVariableNames: ''
	package: 'AST-Core-Nodes'
```
3. The syntax highlighting will complain
4. Accept it
5. DNU: ```PrimitiveFailed: primitive #withArgs:executeMethod: in UndefinedObject failed```

MarcusDenker:
fixes #7987 by defining a fail block in the compiler eval call

### DIFF
--- a/src/Calypso-SystemQueries/ClySystemEnvironment.class.st
+++ b/src/Calypso-SystemQueries/ClySystemEnvironment.class.st
@@ -156,6 +156,7 @@ Is this really what you want to do?') asText
 	[class := classCompiler
 		source: aString;
 		requestor: aController;
+		failBlock: [ ^ nil ];
 		logged: true;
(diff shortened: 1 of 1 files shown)


Ducasse:
Oh cool.

Ducasse:
I did not know it would be that simple.

PR #8453, opened 2021-01-29
8450-ChangeSorter-should-open-and-work-with-new-Spec2

tesonep:
Fixing the ExternalCodeBrowser and the ChangeSorter to use the new version of Spec2.


Ducasse:
Thanks pablo for fixing such boring issues. 
Not even fun.

PR #8455, opened 2021-01-29
8454-Removing-more-users-of-oldDefinition

### LINKED ISSUE #8454: Removing more users of oldDefinition
#oldDefinition is deprecated and should be replaced with #definitionString

tesonep:
Removing more uses of #oldDefinition, replacing it with #definitionString


MarcusDenker (approved):
ok

PR #8456, opened 2021-01-29
2681 debugger step into does not go into accessors or methods with primitives 

### LINKED ISSUE #2681: Debugger StepInto does not go into accessors or methods with primitives
When stepping into an accessor or a method with a primitive, the debugger just step over with no explanation. If not stepping into a primitive or a simple accessor is understandable, it should at least warn the user somehow that a value was returned or a primitive was executed.

StevenCostiou:
Fixes #2681

### DIFF
--- a/src/Kernel-Tests/CompiledCodeTest.class.st
+++ b/src/Kernel-Tests/CompiledCodeTest.class.st
@@ -332,3 +332,8 @@ CompiledCodeTest >> testRefersToLiteralsReturnsTrueWhenLiteralIsInPragmaSelector
 	self assert: (self compiledMethod1
 				refersToLiteral: #pragma1:foo:)
 ]
+
+{ #category : #'tests - debugging support' }
+CompiledCodeTest >> testStepIntoQuickMethods [
+	self deny: self compiledMethod1 stepIntoQuickMethods 
+]
--- a/src/Kernel-Tests/ContextTest.class.st
+++ b/src/Kernel-Tests/ContextTest.class.st
@@ -215,14 +215,49 @@ ContextTest >> testStepIntoQuickMethod [
 	self assert: quickMethod isQuick.
 	
 	"The quick method contains 3 bytecodes:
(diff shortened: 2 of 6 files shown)


Ducasse:
Nice :)
Guille explained to me the primitive and realPrimitive concepts. 
And also the kind of duplication in this class. 
This is really interesting because the logic is quite close to the one in the VM and many be some parts could be better implemented but first tests. 
:)

StevenCostiou:
I updated the tests and the pc calculation to get the return self working.
But:
- The solution is not perfect, I do not understand what Pablo suggests with the primitive. For now it should stay like this if it works, while next week we'll work on tests with @Ducasse to understand best and to ease refactoring later.
- Pablo says there might be a huge performance problem with exceptions mechanics that also use the execution simulation, but we do not know yet until we execute benchmarks. So for now, I will push later another commit with a setting that will disable that stepping by default until we get more knowledge of the impact of that PR. Should not be merged until the Setting is there.
-  Sindarin is breaking hell with these modifications, it does not even build, I will have to fix that before we can merge.

StevenCostiou:
Forgot to comment my last changes:
- The idea here is to make it dynamically configureable through a per-method flag, by default the stepping is deactivated -> the current PR code is about this mechanics,
- these commits form the behavior I would like to achieve, but how it is implemented can be discussed,
-   after that behavior is correctly implemented, or when the current implementation is fixed, I will add a control of that in the debugger -> the debugger, when stepping into, will set this flag and ensure that after stepping the flag is set back to false.
That way, this kind of stepping will only be available when using the debugger model and not in any other case.

Ducasse:
The code looks nice :)
I like that we do not rely on metalink because to me this is a nogo.
And indeed the instructionstream is the way to invest.
S

MarcusDenker (approved):

StevenCostiou:
Thank you for the review :)
After it is merged, I will do another PR for the debugger to use that and enable the stepping into accessors.

Issue #7616, opened 2020-10-29
implementors and senders browsing broken in the debugger


pavel-krivanek:
**Describe the bug**
The debugger strangely uses different broken way how to detect selector for browsing of implementors and senders using keyboard shortcut (for multipart keyword messages)

**To Reproduce**
1. evaluate `self halt` to open a debugger
2. got to `SpCodeSelectionCommand>>#evaluateSelectionAndDo:`
3. add the bottom od the method, there is a call:
```smalltalk
^ self 
		evaluate: selection 
		andDo: aBlock
```
4. place the cursor in the middle of `evaluate:` part and press `Ctrl+M` (or platform equivalent)
5. it will browse implementors of `selection`

If you do the same in the same method in Calypso, it works correctly and browses implementors of `evaluate:andDo:`

Ducasse:
Yes we know. Guille was working on fixing it.

MarcusDenker:
We now merged https://github.com/pharo-project/pharo/pull/7660

But I think this now needs an additional change so this will be used in Spec2

Ducasse:
the hot fix 
```
SpCodePresenter >> selectedSelector
	^ (CNSelectorExtractor new extractSelectorFromSelection: self selectedTextOrLine)
```

pavel-krivanek:
thanks!

pavel-krivanek:
The hot-fix does not seem to work, at least in the case described above, it still shows only implementors of `evaluate:`

maxwills:
I am working on this. Will resume after lunch.

StevenCostiou:
Fixed by https://github.com/pharo-project/pharo/pull/8462

PR #8462, opened 2021-01-31
Fixes issue #7616 implementors and senders browsing broken in the debugger

maxwills:
Fixes issue #7616 - implementors and senders browsing broken in the debugger.
This correct the problem reported for the SpCodePresenter. Now it properly browses or suggests the appropriate selectors when seeking implementors.

### DIFF
--- a/src/Tools-CodeNavigation-Tests/CNSelectorExtractorTest.class.st
+++ b/src/Tools-CodeNavigation-Tests/CNSelectorExtractorTest.class.st
@@ -17,7 +17,14 @@ CNSelectorExtractorTest >> initialize [
 ]
 
 { #category : #test }
-CNSelectorExtractorTest >> testExtractSelectorFromSourceAtPosition_BorderCases_ReturnsNil [
+CNSelectorExtractorTest >> testExtractSelectorFromSourceAtPosition [
+
+	self flag: #toImplement.
+	self assert: false
+]
+
+{ #category : #test }
+CNSelectorExtractorTest >> testExtractSelectorFromSourceAtPositionBorderCasesReturnsNil [
 
    |source position|
 	"Border cases first"
@@ -45,7 +52,7 @@ CNSelectorExtractorTest >> testExtractSelectorFromSourceAtPosition_BorderCases_R
 ]
 
 { #category : #test }
-CNSelectorExtractorTest >> testExtractSelectorFromSourceAtPosition_GeneralCases_ExtractsSelector [
+CNSelectorExtractorTest >> testExtractSelectorFromSourceAtPositionGeneralCasesExtractsSelector [
 
 	"Note that the tested method (#extractSelectorFromSource:atPosition:) uses internally #extractSelectorFromAST:atPosition:"
 	"Also uses RBParser >> #parseFaultyExpression: "
--- a/src/Tools-CodeNavigation-Tests/CNSelectorExtractorTest.class.st
+++ b/src/Tools-CodeNavigation-Tests/CNSelectorExtractorTest.class.st
@@ -17,13 +17,6 @@ CNSelectorExtractorTest >> initialize [
 ]
 
 { #category : #test }
-CNSelectorExtractorTest >> testExtractSelectorFromSourceAtPosition [
-
-	self flag: #toImplement.
-	self assert: false
-]
-
-{ #category : #test }
 CNSelectorExtractorTest >> testExtractSelectorFromSourceAtPositionBorderCasesReturnsNil [
 
    |source position|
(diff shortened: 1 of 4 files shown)


Ducasse:
Thanks max!!!

Ducasse:
I will integrate it. Now could you add some tests because the new methods can be feed with many case that can break.

maxwills:
> I will integrate it. Now could you add some tests because the new methods can be feed with many case that can break.

Sure, I'll add tests now.

maxwills:
Added the tests and it's ready to be reviewed.

Ducasse:
This is cool to have more tests. Today we discuss and we will do another pass on the extractor during the next sprint. To extract receiver.

src/Tools-CodeNavigation-Tests/CNSelectorExtractorTest.class.st
```diff
@@ -0,0 +1,97 @@
...
+	
+	"a string source, invalid Position"
+	source := 'astring'.
+	position := 10000.
+	self assert: (selectorExtractor extractSelectorFromSource: source atPosition: position) isNil.
+
+]
+
+{ #category : #test }
+CNSelectorExtractorTest >> testExtractSelectorFromSourceAtPosition_GeneralCases_ExtractsSelector [
```
StevenCostiou: I find the underscores in method naming suspicious. 
Usually that convention is not used, and we prefer camel case naming convention for methods.
Could the name of the test method be shorntened to avoid the use of underscores?

src/Tools-CodeNavigation-Tests/CNSelectorExtractorTest.class.st
```diff
@@ -0,0 +1,97 @@
...
+{ #category : #initialization }
+CNSelectorExtractorTest >> initialize [
+	"comment stating purpose of instance-side method"
+	"scope: class-variables  &  instance-variables"	
+			
+	selectorExtractor := CNSelectorExtractor new.
+]
+
+{ #category : #test }
+CNSelectorExtractorTest >> testExtractSelectorFromSourceAtPosition_BorderCases_ReturnsNil [
```
StevenCostiou: Same remark about underscores.

StevenCostiou (changes requested):

StevenCostiou:
@Ducasse @MarcusDenker could you review the `CNSelectorExtractor >> extractSelectorFromNode: aNode` method?
I just wonder if there are easier ways to manipulate `RBNodes` and recover a selector from it and maybe you know a way?

StevenCostiou:
> @Ducasse @MarcusDenker could you review the `CNSelectorExtractor >> extractSelectorFromNode: aNode` method?
> I just wonder if there are easier ways to manipulate `RBNodes` and recover a selector from it and maybe you know a way?

Or, merge it after my naming remarks are addressed and improve all of that during next sprint ;)

src/Tools-CodeNavigation-Tests/CNSelectorExtractorTest.class.st
Ducasse: +1

Ducasse:
I checked and the code looks of for me. 
Now indeed I do not like _ and there is no method with underscore in Pharo. Except for hidden FFI

Ducasse:
Ping Max!

StevenCostiou:
@maxwills could you fix the method names when you can?
It is the only thing blocking for a merging and it is a quick and trivial change ;)

maxwills:
I'll make the corrections this afternoon.

src/Tools-CodeNavigation-Tests/CNSelectorExtractorTest.class.st
maxwills: I removed the underscores because it seems like it is not used anywhere else, as Stéphane mentioned.
However, in my opinion, this highlights an issue (A **Naming Convention Issue** in general). Is there an official convention? There are always a reasoning behind convention rules, and we could avoid errors or suggesting naming based on personal preferences (like I did).

The "Tests framework" (Don't know the name) needs a particular name for tests methods to make the "connection" to the actual object method implementation.
I don't like the underscore also, I just used as a reflex. The idea(in previous work) was to have Unit tests with names that clearly describe the purpose by using the tested "methodName_inputOrCase_expectedOutcome" format. I am not saying that this is good or bad for Pharo, I am just saying that the current state of tests (and namings) is not very good in my opinion.

Having "shortened" names will surely break the current "connection" of the method being tested and its test method. Also, I am not a fan of suggestions from personal preferences.

maxwills:
Done, please review

maxwills:
@Ducasse and @StevenCostiou. By the way, yesterday, I noticed the messages and pings from previous Friday. Then, I realized that I had a bad notifications configuration, so I didn't received emails. I have corrected that, and from now on, I should reply faster. Sorry about potential frustrations, will not happen again. :(

@StevenCostiou, pinging for review.

StevenCostiou (approved):

StevenCostiou:
@maxwills do not forget to specify which issue it will close, you can do this in the PR first comment:
Fixes #7616 
It automatically closes the issue when your PR is merged :)

PR #8468, opened 2021-02-01
8467-Fuel-is-using-an-old-API-from-MultiByteFileStream

### LINKED ISSUE #4346: Fuel should not depend on Deprecated file streams
The tests of Fuel are depending on deprecated file streams. 
One reference can just be removed since there are no senders, but the other is used in tests.

Maybe @theseion knows what to do about it.

### LINKED ISSUE #8467: Fuel is using an old API from MultiByteFileStream
**Describe the bug**
Fuel is using #nextBytesPutAll: 
This method is not present in new Streams as it is assuming the type of the receiver (it specs the receiver to be a binary stream). We need to replace it with #nextPutAll:

tesonep:
The message nextBytesPutAll: is not implemented in streams.
It should be replaced by using nextPutAll: with a binary stream.

### DIFF
--- a/src/Fuel-Core/FLBufferedWriteStream.class.st
+++ b/src/Fuel-Core/FLBufferedWriteStream.class.st
@@ -138,7 +138,7 @@ FLBufferedWriteStream >> nextBytesPutAll: collection [
 		ifFalse: [
 			self flushBuffer.
 			collection size > (self buffer size / 2)
-				ifTrue: [ stream nextBytesPutAll: collection ]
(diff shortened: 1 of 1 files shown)


tesonep:
This is a fix of an issue detected in a test FLContextSerializationTest >> #testDoIt. That is why this change does not have new tests. This error appears because we have now a class with a longer comment (yes... we turn on the light and the boiler starts)

MarcusDenker (approved):

PR #8472, opened 2021-02-01
RB - Expose variable refactoring commands in source context

### LINKED ISSUE #8345: Expose refactoring push up instance variable from class definition
Right now, when we do a right click on an instance variable in a class definition we do not see all refactorings that apply to it.

image

**Describe the request**
In particular, we don't have protect, push up nor push down.
It would be nice if both lists of refactorings are created in the same way to avoid having duplicated declarations.

**Expected behavior**

Have the same list we have when we do a right click on the variables in Calypso's variable list.

image

**Version information:**
 - Version: Observed in build Pharo-9.0.0+build.880.sha.ead4e5a5597d939a558ffcde702f838a8d867274 (64 Bit)

Lin777:
Fixes #8345

Now we have exposed all variable refactoring commands in source context.

imagen
imagen


Ducasse:
Good now one of these days we will have to use commander 2 commands.

PR #8476, opened 2021-02-02
cleanup-ExternalObject-class-initialize 

MarcusDenker:
ExternalObject class>>#initialize was sending a non-existing selector #addToStartUpList:after:

We cleaned that up everywhere else already, so we can do it here, too.


MarcusDenker:
This change leads to a crash when building the full image from the bootstrap image.

MarcusDenker:
Second run was successful

PR #8489, opened 2021-02-03
Add simple mock objects implementation to SUnit

gcorriga:
This PR adds a simple implementation of mock objects to SUnit, suitable for inclusion in the Pharo 9 base image.

While simpler and less sophisticated than other libraries, this implementation is still quite powerful.

This implementation takes a different approach compared to BabyMock/BabyMock2 and Mocketry:
- there are no methods such as `#should`, `#can`, or `#be:`
- the mocks are more strict in their behaviour - users need to send all and only the messages defined, in the same order as they were defined.
- the mocks need to be manually verified using the `TestCase>>verify:` extension.

These limitations are on purpose, mainly for two reasons:
- to discourage the use of these objects unless they are really needed.
- to keep the implementation simple so that it can be integrated in Pharo's SUnit library, as opposed to being its own framework.

### DIFF
--- a/src/BaselineOfSUnit/BaselineOfSUnit.class.st
+++ b/src/BaselineOfSUnit/BaselineOfSUnit.class.st
@@ -39,6 +39,7 @@ BaselineOfSUnit >> baseline: spec [
 				package: 'SUnit-Tests';
 				package: 'SUnit-Visitor' with: [ spec requires: #('SUnit-Core') ];
 				package: 'SUnit-Visitor-Tests' with: [ spec requires: #('SUnit-Visitor') ];
+				package: 'SUnit-MockObjects' with: [ spec requires: #('SUnit-Core') ];
 				package: 'SUnit-UI';
 				package: 'SUnit-Support-UITesting-Tests';
 				package: 'SUnit-Support-UITesting';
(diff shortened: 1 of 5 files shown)


astares:
Do we really need that as as part of the base image?

There are other mock frameworks:
- https://github.com/astares/Pharo-Teachable
- https://github.com/dionisiydk/Mocketry
- ...
and it would be hard to draw a line what should be included and what not.

@gcorriga If you need it in your projects why not just create a dependency in the baseline?

<blockquote><div> GitHub</div><div><strong><a href="https://github.com/astares/Pharo-Teachable">astares/Pharo-Teachable</a></strong></div><div>A teachable object for Pharo. Contribute to astares/Pharo-Teachable development by creating an account on GitHub.</div></blockquote>
<blockquote><div> GitHub</div><div><strong><a href="https://github.com/dionisiydk/Mocketry">dionisiydk/Mocketry</a></strong></div><div>Mock objects framework. Contribute to dionisiydk/Mocketry development by creating an account on GitHub.</div></blockquote>

gcorriga:
@astares yes I think it would be useful to have a simple mock object implementation in the base image. I have no preference whether to use this one or yours, but I would prefer to have something simpler than Mocketry.

As a matter of fact, this is something I’ve extracted from some project of mine and it’s already in its baseline ;)

PR #8497, opened 2021-02-04
fixes #4856 - print on methods should avoid calling blindly print string on their elements

### LINKED ISSUE #4856: printOn: methods should avoid calling blindly printString on their elements...
since it just creates intermediary streams for nothing.

This is ugly! and slow and consuming memory for nothing. 

printOn: aStream

	aStream nextPut: $(;
		nextPutAll: g0Size printString; space;
		nextPutAll: g1Size printString; space;
		nextPutAll: g0Leading printString; space;
		nextPutAll: g1Leading printString; space;
		nextPutAll: charSize printString; space;
		nextPutAll: streamPosition printString.
	aStream nextPut: $).

kasperosterbye:
fixes #4856. There is a rule (`ReNoPrintStringInPrintOnRule`) which marks printString inside printOn: as bad style, so there is hope for the future. Almost all cases could be fixed by changing `aStream nextPutAll: obj printString` into `aStream print: obj`.

### DIFF
--- a/src/Clap-Core/ClapValidationReport.class.st
+++ b/src/Clap-Core/ClapValidationReport.class.st
@@ -45,6 +45,6 @@ ClapValidationReport >> isSuccess [
 ClapValidationReport >> printOn: aStream [
 	problems do: [ :each |
 		aStream
-			nextPutAll: each printString;
+			print: each;
 			cr ]
 ]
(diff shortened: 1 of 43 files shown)


MarcusDenker:
There are failing test in TonelWriterTest that seem to be related

kasperosterbye:
I think it is in order now - the fail seems unrelated to this PR.

MarcusDenker (approved):
looks good

PR #8507, opened 2021-02-04
Adding deprecations for Opal Compiler that were recommended by DepMiner and accepted by Pablo

olekscode:
Those methods were present in the Opal Compiler of Pharo 8 but deleted in Pharo 9.
DepMiner tool (https://github.com/olekscode/DepMiner) recommended to return them with deprecations.

Those recommendations were rejected by @Ducasse, @MarcusDenker, and @guillep.
However, they were accepted by @tesonep.

As far as I remember, @guillep did not want to accept those methods because they are "private" and therefore should not be deprecated.

### DIFF
--- a/src/Deprecated90/OCAbstractVariable.class.st
+++ b/src/Deprecated90/OCAbstractVariable.class.st
@@ -23,6 +23,24 @@ OCAbstractVariable >> asString [
 	^ self name
 ]
 
+{ #category : #generated }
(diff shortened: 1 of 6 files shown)


MarcusDenker (approved):
OCAbstractVariable is itself deprecated.

It does not hurt to add these so it is in sync with the class that replaced it.

Issue #8443, opened 2021-01-29
Crash on step over


Alisu:
**Describe the bug**
The image froze on some step over.

**To Reproduce**
1. Open an image
2. Open a playground
3. Type [ 'pouet'. [ ^42 ] value ] ensure: [ ^ 'popouet' ] in the playground
4. Now step over, step over, step into, step over, step over, step over and image is frozen.

**Expected behavior**
Image does not froze.

**Version information:**
 - OS: OSX mojave
On Pharo 9 and Pharo 8

Additionnal:

It looks like an infinite debugger opening loop.

Alisu:
It looks like the same.
Another way to reproduce:
-Open an image
-Open a playground
-Type [ 'pouet'. [ ^42 ] value ] ensure: [ ^ 'popouet' ] in the playground
-Step over, over, into, into, into, into, over

StevenCostiou:
Thanks for the report Théo!
I will have a look next week.

Ducasse:
I hope that this is not another one of the instruction stream bugs.

StevenCostiou:
The stack (thousands of them) from the log, it does not go deeper.

```
Context>>cannotReturn:
	Receiver: Process>>doTerminationFromYourself
	Arguments and temporary variables: 
		result: 	a Process in ni
	Receiver's instance variables: 
		sender: 	nil
		pc: 	nil
		stackp: 	1
		method: 	Process>>#doTerminationFromYourself
		closureOrNil: 	nil
		receiver: 	a Process in nil


Process>>doTerminationFromYourself
	Receiver: a Process in nil
	Arguments and temporary variables: 

	Receiver's instance variables: 
		nextLink: 	nil
		suspendedContext: 	nil
		priority: 	40
		myList: 	nil
		name: 	nil
		env: 	nil
		effectiveProcess: 	nil
		terminating: 	true
		level: 	nil
```
It is obvious it cannot return, but it should not loop like that.
Started investigating, @Alisu any idea what it could be?

StevenCostiou:
So what I did:
- I disabled all debuggers
- I cut the Oups entry point to log to the Transcript instead of doing anything like trying to open debuggers
This example still hangs with the same stack. So it's at the level of the interpreter and not the debugger infrastructure.
More: the debugger infrastructure can do nothing about it, it is never sollicitated in that execution (if I did not miss anything).

StevenCostiou:
[...]

PR #8509, opened 2021-02-04
Fix simulation of return from unwind block like in following example:…

dionisiydk:
Fix simulation of return from unwind block like in following example: 
```Smalltalk
[ #some ] ensure: [ ^#unwindReturn ]
```

### DIFF
--- a/src/Kernel/Context.class.st
+++ b/src/Kernel/Context.class.st
@@ -1713,20 +1713,25 @@ Context >> return: value [
 ]
 
 { #category : #controlling }
-Context >> return: value from: aSender [ 
+Context >> return: value from: aSender [
 	"For simulation.  Roll back self to aSender and return value from it.  Execute any unwind blocks on the way.  ASSUMES aSender is a sender of self"
-
 	| newTop context |
 	aSender isDead ifTrue: [
 		^ self send: #cannotReturn: to: self with: {value} super: false ].
 	newTop := aSender sender.
 	context := self findNextUnwindContextUpTo: newTop.
+	(context notNil and: [ context unwindBlock = closureOrNil ]) ifTrue: [
+		"Here the return happens inside the unwind block like in following example: 
+				[ #test ] ensure: [ ^ #returnFromUnwindBlock ]
+		In such cases #findNextUnwindContextUpTo: returns the corresponding #ensure context
+		because it is always a sender of unwind block.
+		Here we detect this case and lookup the next unwind context after the current ensure section (current unwind)"
+		context := context findNextUnwindContextUpTo: newTop ].
 	context ifNotNil: [
-		^ self send: #aboutToReturn:through: to: self with: {value. context} super: false] .
+		^ self send: #aboutToReturn:through: to: self with: {value. context} super: false ].
 	self releaseTo: newTop.
 	newTop ifNotNil: [ newTop push: value ].
 	^ newTop
-
 ]
 
 { #category : #controlling }
--- a/src/Debugger-Model-Tests/StepOverTest.class.st
+++ b/src/Debugger-Model-Tests/StepOverTest.class.st
@@ -5,6 +5,11 @@ Class {
 }
 
 { #category : #helper }
+StepOverTest >> methodWithUnwindReturn [
+	[ #body ] ensure: [ ^#unwindReturn ]
+]
+
+{ #category : #helper }
 StepOverTest >> step1 [
 	self step2.
 ]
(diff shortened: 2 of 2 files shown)


dionisiydk:
Fixes #8443

dionisiydk:
While it fixes original example it is more complex problem.
I have another similar test which also hangs the image.

But if tests are fine PR can be integrated. It's a good step anyway

src/Kernel/Context.class.st
```diff
@@ -1713,20 +1713,25 @@ Context >> return: value [
...
 		^ self send: #cannotReturn: to: self with: {value} super: false ].
 	newTop := aSender sender.
 	context := self findNextUnwindContextUpTo: newTop.
+	(context notNil and: [ context unwindBlock = closureOrNil ]) ifTrue: [
+		"Here the return happens inside the unwind block like in following example: 
+				[ #test ] ensure: [ ^ #returnFromUnwindBlock ]
+		In such cases #findNextUnwindContextUpTo: returns the corresponding #ensure context
+		because it is always a sender of unwind block.
+		Here we detect this case and lookup the next unwind context after the current ensure section (current unwind)"
+		context := context findNextUnwindContextUpTo: newTop ].
```
StevenCostiou, line 1729: If I understand correctly, before your fix the `aboutToReturn:through:` is sent with the `ensure:` context as context.
Then the fix makes sure that the returned context is the next context to unwind, which is the return in the ensure block.

Did you identify what causes the hanging in the first case?

StevenCostiou (approved):

StevenCostiou:
> I have another similar test which also hangs the image.
Could you share that other test?

> But if tests are fine PR can be integrated. It's a good step anyway
I approved because, as far as I understand, it looks good and all tests are passing.
Perhaps @guillep or @tesonep should have a look?

src/Kernel/Context.class.st
dionisiydk: The real bug still exists: when simulation is going into #aboutToReturn:through: the image hangs. 
My fix just removes the one specific case where simulation should not go this way. But the real problem is still here. And it's easy to reproduce:
```Smalltalk
[
   [ 100 + 200 ] ensure: [^600]
] ensure: [ #done ]
```
Step inside [^600] block. Then step over it will hang the image.

dionisiydk: So the situation is much worser. We can't step over return inside ensure:
```Smalltalk
[ ^100 ] ensure: [ 10+30]
```
Step into block [^100]. Then step over it will hang the image.
I checked also Pharo 7 - same issue

dionisiydk: So why the image hangs in all those cases? This is very interesting story :) 
The short answer: bugs in the debugger lead to the state of target process where it never ends.

The last call of any process is the termination which ends with #suspend.
```Smalltalk
newProcess
	^Process
		forContext: 
			[self value.
			Processor terminateActive] asContext
		priority: Processor activePriority
```
The block here is actually never returned and it must not return. If you will remove the #terminateActive call then processes will end with BlockCannotReturn error (see cannotReturn:) because VM will not find the sender to continue the process. 

And it is what really happens when we got hanging in the debugger: 
- #terminateActive call does not suspend the current process and simply returns.
- VM calls #cannotReturn: which triggers the debugger:
  - new debugger is requested
  - debugger tries to #suspend activeProcess but it does nothing (see below) and process continues. 
  - as process continues the method #cannotReturn: returns.
- cannotReturn: has no sender and therefore VM calls cannotReturn: again - Infinite recursions began.
  
Now why #terminateActive and #suspend can be not working.

The problem is how debugger simulates the #activeProcess calls during stepping operations:. 
```Smalltalk
process := Process activeProcess
```
The StepOver this expression is evaluated by the debugger process (UIProcess) but still the result will be the process under the debugger (not a UIProcess). 
To achieve this the debugger wraps all stepping operations using #effectiveProcess trick:
```Smalltalk
step
	^Processor activeProcess
		evaluate: [suspendedContext := suspendedContext step]
		onBehalfOf: self
```
The active process here is UIProcess (which handles StepOver click). It sets the variable #effectiveProcess to the process under the debugger. During the block execution the Processor will always use #effectiveProcess as an active one instead of the real active process which executes the debugger command. 
For example you can step over "Processor terminateActive" and it will not terminate the UIProcess while the command is really executed by UIProcess.

Now it should be clear how the normal process completion becomes broken: if #effectiveProcess variable is not reset by the debugger the #terminateActive call will never kill the current process.
And same issue happens with #suspend call. When debugger is requested it uses "Processor activeProcess" to get a process for debugging and to suspend it but it will never return the actual current process. Therefore the #suspend from the debugger will not stop the current process (it will suspend some other process) and the current process will continue execution which will cause the buggy recursion.

Now why effectiveProcess can be not reset? Method #evaluate:onBehalfOf: contains the ensure logic to reset the variable back. 
The block here is supposed to do the context/stack manipulations. And some of them are really not trivial. The processing of unwinds (ensure parts) are their responsibility. So it's obvious that there can be some bugs or missing parts. Especially that we don't have good tests for simulation logic.

For this concrete issue with StepOver the problem is in the way how debugger intercepts the return from the callee. The current trap is not enough for the returns inside unwind contexts. And therefore the simulated code is not stopped on the current context and completely finishes the execution up to last #terminateActive in the #newProcess. So #terminateActive is executed with effective process and it hangs the image.

The trap needs to be improved and maybe it's not really easy to do. But what is more important is to find a general fix for described problem. 
I hope you understand my explanation. I will push some obvious changes soon: #newProcess should ensure the termination of actual active process despite on any assigned effectiveProcess.

Ducasse:
Hi denis
Just that you know steven is dead sick at home. 
S.

StevenCostiou:
Hi, sorry I am now on holidays even if I keep an eye from time to time.
After I come back on March 8th I will review if this is still open.

isCzech:
Hi, please check PR #10383; I think the PR fixes the issue. Thanks.

guillep:

guillep:

PR #8521, opened 2021-02-05
Instruction stream tests first pass

StevenCostiou:
With @Ducasse we pair-programmed to understand the instruction stream and context classes.
Along the way, we write tests and fix little things (comments, dead code, etc.).

This is a first pass, there will be many others.

All comments are welcome.

### DIFF
--- a/src/Debugging-Core-Tests/DebuggingInstructionStreamTest.class.st
+++ b/src/Debugging-Core-Tests/DebuggingInstructionStreamTest.class.st
@@ -61,19 +61,15 @@ DebuggingInstructionStreamTest >> testFollowingByteAccessByteCodeConstituentByte
 { #category : #'tests - structure' }
 DebuggingInstructionStreamTest >> testFollowingByteWithSuper [
 
-	| compiledMethod stream client stream2 client2 |
+	| compiledMethod stream client |
 	compiledMethod := self class >> #methodWithSuper.
 	stream := InstructionStream on: compiledMethod.
 	client := InstructionClient new.
 	stream interpretNextInstructionFor: client.
-	
-	self assert: stream followingByte identicalTo: (compiledMethod at: 43)
-	
-	
-	"stream2 := InstructionStream on: compiledMethod.
-	client2 := InstructionClient new.
-	stream2 interpretNextInstructionFor: client2.
-	stream2 interpretNextInstructionFor: client2."
+	self flag: 'improve this test to avoid magic number'.
+	self
+		assert: stream followingByte
+		identicalTo: (compiledMethod at: 43)
 ]
 
 { #category : #'tests - interpretation' }
--- a/src/Debugging-Core-Tests/DebuggingInstructionStreamTest.class.st
+++ b/src/Debugging-Core-Tests/DebuggingInstructionStreamTest.class.st
@@ -1,0 +1,145 @@
+Class {
+	#name : #DebuggingInstructionStreamTest,
+	#superclass : #TestCase,
+	#category : #'Debugging-Core-Tests'
+}
+
+{ #category : #helpers }
+DebuggingInstructionStreamTest >> method1 [
+
+	| a |
+	a := 1.
+	^ 1 + 2
+]
+
+{ #category : #helpers }
+DebuggingInstructionStreamTest >> methodWithBlock [
+
+	[| a |
+	a := 1.
+	^ 1 + 2] value
+]
+
+{ #category : #helpers }
+DebuggingInstructionStreamTest >> methodWithSuper [
+
+	super printString.
+	^ 42
+]
+
+{ #category : #'tests - structure' }
+DebuggingInstructionStreamTest >> testFirstByte [
+	|compiledMethod stream|
+	compiledMethod := self class >> #method1.
+	stream := InstructionStream on: compiledMethod.
+	self assert: stream firstByte equals: (compiledMethod at: compiledMethod initialPC)
+]
+
+{ #category : #'tests - structure' }
+DebuggingInstructionStreamTest >> testFollowingByteAccessByteCodeConstituentByteByByte [
+
+	| compiledMethod stream client |
+	compiledMethod := self class >> #methodWithSuper.
+	stream := InstructionStream on: compiledMethod.
+	client := InstructionClient new.
+	
+	"Answer the next byte.
+	For example for this bytecode 
+	<EB 00> superSend 
+	<D8> pop
+	 If the current byte is EB, the following byte is 00.
+	 It is not the following bytecode, that is D8"
+ 	stream pc: stream pc + 1.	
+	self assert: stream followingByte identicalTo: (compiledMethod at: compiledMethod initialPC + 2).
+	self deny: stream followingByte identicalTo: stream followingBytecode.
+	stream pc: stream pc + 1.
+	self assert: stream followingByte identicalTo: (compiledMethod at: compiledMethod initialPC + 3).
+	stream pc: stream pc + 1.
+	self assert: stream followingByte identicalTo: stream followingBytecode
+]
+
+{ #category : #'tests - structure' }
+DebuggingInstructionStreamTest >> testFollowingByteWithSuper [
+
+	| compiledMethod stream client stream2 client2 |
+	compiledMethod := self class >> #methodWithSuper.
+	stream := InstructionStream on: compiledMethod.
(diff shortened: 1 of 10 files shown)


Ducasse:
Thanks Steven. We should continue because this is a super fun way to learn. 
I will ask marcus or pablo to review what we did and also get feedback on the renaming we propose.

src/Debugging-Core-Tests/DebuggingInstructionStreamTest.class.st
```diff
@@ -0,0 +1,141 @@
...
+	stream := InstructionStream on: compiledMethod.
+	client := InstructionClient new.
+	
+	"Answer the next byte.
+	For example for this bytecode 
+	<EB 00> superSend 
+	<D8> pop
+	 If the current byte is EB, the following byte is 00.
+	 It is not the following bytecode, that is D8"
+ 	stream pc: stream pc + 1.
```
guillep, line 52: I don't understand why the +1 here. It is assuming 00? But by reading the comment I would have thought your stream is at EB.

src/Debugging-Core-Tests/DebuggingInstructionStreamTest.class.st
```diff
@@ -0,0 +1,141 @@
...
+]
+
+{ #category : #'tests - interpretation' }
+DebuggingInstructionStreamTest >> testInterpretNextInstruction [
+	| compiledMethod stream client currentPC |
+	compiledMethod := self class >> #method1.
+	stream := InstructionStream on: compiledMethod.
+	client := InstructionClient new.
+	currentPC := stream pc.
+	stream interpretNextInstructionFor: client.
```
guillep, line 82: This is to me a too generic name for the test. It really depends on what bytecode you are...

src/Debugging-Core-Tests/DebuggingInstructionStreamTest.class.st
```diff
@@ -0,0 +1,141 @@
...
+
+{ #category : #'tests - structure' }
+DebuggingInstructionStreamTest >> testNextByte [
+
+	| compiledMethod stream client |
+	compiledMethod := self class >> #methodWithSuper.
+	stream := InstructionStream on: compiledMethod.
+	client := InstructionClient new.
+	stream interpretNextInstructionFor: client.
+	self assert: stream nextByte
```
guillep, line 94: What is the difference between nextByte and followingByte?

src/Debugging-Core-Tests/DebuggingInstructionStreamTest.class.st
```diff
@@ -0,0 +1,141 @@
...
+
+{ #category : #'tests - structure' }
+DebuggingInstructionStreamTest >> testNextPc [
+
+	| compiledMethod stream bytecode |
+	compiledMethod := self class >> #method1.
+	stream := InstructionStream on: compiledMethod.
+	bytecode := stream firstByte.
+	self
+		assert: (stream nextPc: bytecode)
```
guillep, line 119: this method looks fishy. What happens if a same bytecode appears many times in the stream? Does it only take the first one? The name of the test is not really very intention revealing of what it is testing

src/Debugging-Core-Tests/DebuggingInstructionStreamTest.class.st
```diff
@@ -0,0 +1,141 @@
...
+	| compiledMethod stream client stream2 client2 |
+	compiledMethod := self class >> #methodWithSuper.
+	stream := InstructionStream on: compiledMethod.
+	client := InstructionClient new.
+	stream interpretNextInstructionFor: client.
+	
+	stream2 := InstructionStream on: compiledMethod.
+	client2 := InstructionClient new.
+	stream2 interpretNextInstructionFor: client2.
+	stream2 interpretNextInstructionFor: client2.
```
guillep, line 136: This line is repeated on purpose?

src/Debugging-Core/InstructionStream.extension.st
```diff
@@ -24,12 +24,17 @@ InstructionStream >> debuggerMap [
 InstructionStream >> firstByte [
 	"Answer the first byte of the current bytecode."
 
-	^self method at: pc
+	^self method at: self method initialPC
```
guillep, line 27: This new implementation does not go well with the comment. The comment is supposed to return the first byte of the **current bytecode**. The new implementation returns the first byte of the method.

src/Debugging-Core/InstructionStream.extension.st
```diff
@@ -253,37 +258,20 @@ InstructionStream >> interpretV3PlusClosureJumpIfCond [
 
 { #category : #'*Debugging-Core' }
 InstructionStream >> nextByte [
-	"Answer the next bytecode."
+	"Answer the next bytecode"
+	self flag: 'Should this method be renamed nextBytecode? Is it possible thta it returns only a byte?'.
 
 	^self method at: pc
```
guillep, line 264: What is the difference with followingByte? I was thinking this one advanced the stream, but it does not

src/Kernel-Tests/CompiledCodeTest.class.st
```diff
@@ -30,6 +30,18 @@ CompiledCodeTest >> testAccessesRef [
...
+{ #category : #'tests - accessing' }
+CompiledCodeTest >> testBytecodeSizeForPc [
+
+	| method |
+	method := self class >> #method1.
+
+	self
+		assert: (method bytecodeSizeForPc: method initialPC)
+		equals:
+		(method encoderClass bytecodeSize: (method at: method initialPC))
```
guillep, line 42: Again, this test really depends on the input, and different inputs do different things. Maybe they need different tests...

guillep (commented):
I have some comments, there is one that looks like a bug, so be careful...

src/Debugging-Core-Tests/DebuggingInstructionStreamTest.class.st
Ducasse: Yes. We eat two.

src/Debugging-Core/InstructionStream.extension.st
Ducasse: Indeed this is what we faced. nextByte is the nextByteCode while followingByte is the following byte of the byte code

Ducasse:
Thanks we will check your comment. I wish to have the dumbledoor stuff to remember :).

src/Debugging-Core-Tests/DebuggingInstructionStreamTest.class.st
Ducasse: If I remember one is a bytecode and the other a byte so we found that super confusing.

src/Debugging-Core-Tests/DebuggingInstructionStreamTest.class.st
Ducasse: Here we wanted to show is that if we have a bytecode with two bytes, the followingByte (not bytecode) is the one following the one pointed by the current PC
So we point at EB and followingByte should be 00. 
May be the test is wrong this is what we want to test.

StevenCostiou:
Hi, thanks for the reviews!
Sorry I am now on holidays even if I keep an eye from time to time.
After I come back on March 8th I will come back to that.

src/Debugging-Core-Tests/DebuggingInstructionStreamTest.class.st
Ducasse: agreed

src/Debugging-Core/InstructionStream.extension.st
Ducasse: agreed!

src/Kernel-Tests/CompiledCodeTest.class.st
Ducasse: Probably we were started to see.

src/Debugging-Core/InstructionStream.extension.st
Ducasse: to me the current definition is super bad

```
followingByte
	"Answer the next bytecode."

	^self method at: pc + 1
```

Ducasse:
Stevn I will close it and I rescued it because I still would to continue.

guillep:
This was already redone and integrated in Pharo10

PR #8522, opened 2021-02-05
Fix issue 8491 #valueAt: fails to yield...

### LINKED ISSUE #8491: BlockClosure>>#valueAt: fails to yield to higher priority process
This example fails to answer true:
```
x := 0.
[ [ x := 1 ] forkAt: 41 ] valueAt: 42.
x == 1.  "false - wrong"
```
The problem is the Processor yield in #valueAt won't check for higher priority processes; yield's primitive 167 only checks within active priority.

I suggest either: (A) fix #valueAt: the same way as #valueUnpreemptively (inline yield without the primitive)
or better: (B) fix Process>>#priority: to take care of the situation in the spirit of:
```
Process >> priority: anInteger
	"Set the receiver's priority to anInteger.
	... and make sure the active process won't continue running below a higher priority runnable process."

	(anInteger between: Processor lowestPriority and: Processor highestPriority)
		ifTrue: [
			priority := anInteger.
			(self isActiveProcess and: [anInteger < Processor nextReadyProcess priority])
				ifTrue: [ [self resume] fork. self suspend ] ]  "force active process give up"
		ifFalse: [ self error: 'Invalid priority: ' , anInteger printString ]
```
Methods #valueAt: (and #valueUnpreemptively) could then be cleaned up:
```
BlockClosure >> valueAt: blockPriority 
	"Evaluate the receiver (block), with another priority as the actual one 
	and restore it afterwards. The caller should be careful with using 
	higher priorities."

	| activeProcess outsidePriority |
	activeProcess := Processor activeProcess.
	outsidePriority := activeProcess priority.
	activeProcess priority: blockPriority.
	^self ensure: [activeProcess priority: outsidePriority]
```
Build information: Pharo-9.0.0+build.1093.sha.5a650fec081de6a41678530844647baad6723797 (64 Bit)

Regards, 
Jaromir

isCzech:
Fixes issue 8491 #valueAt: fails to yield... and adds a test

### DIFF
--- a/src/Kernel-Tests/BlockClosureTest.class.st
+++ b/src/Kernel-Tests/BlockClosureTest.class.st
@@ -434,6 +434,16 @@ BlockClosureTest >> testSuppressInformUsingStringMatchOptions [
 ]
 
 { #category : #'tests - evaluating' }
+BlockClosureTest >> testValueAt [
+	| val oldPriority |
+	val := 0.
+	oldPriority := Processor activePriority.
+	[ [ val := 1 ] forkAt: oldPriority + 1 ] valueAt: oldPriority + 2.
+
+	self assert: val equals: 1
+]
+
+{ #category : #'tests - evaluating' }
 BlockClosureTest >> testValueWithArguments [
 
 	self shouldnt: [ aBlockContext valueWithArguments: #() ] raise: ArgumentsCountMismatch.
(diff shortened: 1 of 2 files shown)


Ducasse:
Hello this change can have strong impact. So we should pay really attention. I would not integrate it in P9 but consider it for P10. We have to enter beta phase soon.

guillep:
Thanks for the PR!

`BlockClosure >> valueAt: ` is being deprecated in Pharo10. See https://github.com/pharo-project/pharo/pull/10004/files

Issue #8524, opened 2021-02-06
Process>>#priority: allows lower priority process run while higher priority one scheduled


isCzech:
In the following example a higher priority process keeps waiting in the ProcessorScheduler queue until a lower priority process gets suspended!
```
Transcript flush; cr.
Processor activeProcess priority: 42.
[ Transcript nextPut: $A ] forkAt: 41.
Processor activeProcess priority: 40.
Transcript nextPut: $B; flush.
500 factorial.
Transcript nextPut: $C; flush.
500 factorial.
Transcript nextPut: $D; flush.
10 milliSeconds wait.   "suspends the active process"
Transcript nextPut: $E; flush.
```
Answers:

BCDAE
BCDAE
BCADE
BCDAE
BCDAE
BACDE
BCDAE
(A never appears before B; interrupts may change the order of C, D and A which is expected)

Reason: the `#priority:` method won't make sure the active process doesn't jump over a scheduled process when lowering it's priority - as a result the active process will continue running until it suspends and won't let the higher priority process a chance to run.

Current implementation:
```
priority: anInteger
	"Set the receiver's priority to anInteger."

	(anInteger between: Processor lowestPriority and: Processor highestPriority)
		ifTrue: [ priority := anInteger ]
		ifFalse: [ self error: 'Invalid priority: ' , anInteger printString ]
```
Proposed solution:
```
priority: anInteger
	"Set the receiver's priority to anInteger."

	(anInteger between: Processor lowestPriority and: Processor highestPriority)
		ifTrue: [
			priority := anInteger.
			(self isActiveProcess and: [anInteger < Processor nextReadyProcess priority])
				ifTrue: [ [self resume] fork. self suspend ] ]
		ifFalse: [ self error: 'Invalid priority: ' , anInteger printString ]
```
The problem with this solution is the #nextReadyProcess may get interrupted in the middle which may make it's answer usless. So I guess either this whole method should run as a primitive or in the interim, always force the active process suspend momentarily to let ProcessorScheduler do the job:
```
priority: anInteger
	"Set the receiver's priority to anInteger."

	(anInteger between: Processor lowestPriority and: Processor highestPriority)
		ifTrue: [
			priority := anInteger.
			self isActiveProcess ifTrue: [ [self resume] fork. self suspend ] ]
		ifFalse: [ self error: 'Invalid priority: ' , anInteger printString ]
```

[...]

PR #8526, opened 2021-02-06
Fix Process>>#priority: incorrectly allowing lower priority process run while …

isCzech:
…higher priority one is waiting

fixes issue #8524 , covers the root cause of #8491 )


Ducasse:
Hello this change can have strong impact. So we should pay really attention. I would not integrate it in P9 but consider it for P10. We have to enter beta phase soon.

guillep:
Superseeded by #10004

PR #8531, opened 2021-02-08
RB - Improve messages push up method refactoring

### LINKED ISSUE #8478: Enhance Push Up method refactoring message
When pushing up a method that references an instance variable, and that instance variable is not in the superclass, there is just a message not very clear and with a "Yes/No" action.

**Expected behavior**
We should have better feedback to make the decision.
Also, maybe the refactoring could propose to push up the instance variable too.

Lin777:
Fixes #8478

Now we have better feedback to make the decision.
Also, we have the option to push up the instance variable too (if it is referenced in push up method).

### DIFF
--- a/src/Refactoring-Core/RBPullUpMethodRefactoring.class.st
+++ b/src/Refactoring-Core/RBPullUpMethodRefactoring.class.st
@@ -100,7 +100,9 @@ RBPullUpMethodRefactoring >> checkInstVarsFor: aSelector [
 								with: each
 								with: class)) 
 						ifTrue: [ self pushUpVariable: each ]
-						ifFalse: [ self refactoringError: 'Do you want push up anyway?' ] ]]
+						ifFalse: [ self refactoringError: 'You are about to push your method without the instance variable it uses. 
+						It will bring the system is an inconsistent state. But this may be what you want.
+						So do you want to push up anyway?' ] ]]
 ]
 
 { #category : #preconditions }
--- a/src/Refactoring-Core/RBPullUpMethodRefactoring.class.st
+++ b/src/Refactoring-Core/RBPullUpMethodRefactoring.class.st
@@ -96,10 +96,11 @@ RBPullUpMethodRefactoring >> checkInstVarsFor: aSelector [
 	class instanceVariableNames do: 
 			[:each | 
 			((class whichSelectorsReferToInstanceVariable: each) includes: aSelector) ifTrue: 
-					[self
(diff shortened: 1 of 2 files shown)


guillep:
Hi @Lin777 could you put a screenshot?

Lin777:
Hi @guillep, 

For example, I have these classes
```
Object subclass: #MyClassA
	instanceVariableNames: ''
	classVariableNames: ''
	package: 'example'

MyClassA subclass: #MyClassB
	instanceVariableNames: 'b a'
	classVariableNames: ''
	package: 'example'

MyClassB >> exampleMethod
	^ a + 6
```
- Select #exampleMethod and press push up refactoring option, this will open this warning, to ask if you want also PULL UP the instance variable referenced in that method. 
imagen

If you press YES, the method and instance variable will be push up to superclass, else if you press NO, other warning will be open

imagen

if you press YES, the method will be push up but instance variable reference will be broken, else if you press NO, the refactoring is aborted

guillep:
I'm not fond of the cascade of dialog windows, but ok, this improves the current state

Ducasse:
Thanks evelyn. This looks good and a nice step.
For the do you want push up anyway?  
it would be good to write

You can push you method without the instance variable it uses. It will bring the system is an inconsistent state. 
But may be this is what you want. Do you want to push up anyway?

src/Refactoring-Core/RBPullUpMethodRefactoring.class.st
```diff
@@ -96,10 +96,11 @@ RBPullUpMethodRefactoring >> checkInstVarsFor: aSelector [
...
 			[:each | 
 			((class whichSelectorsReferToInstanceVariable: each) includes: aSelector) ifTrue: 
-					[self
-						refactoringError: ('<1p> refers to <2s> which is defined in <3p>' expandMacrosWith: aSelector
+					[ (self confirm: ('<1p> refers to #<2s> which is defined in <3p>. Do you want push up variable #<2s> also?' expandMacrosWith: aSelector
 								with: each
-								with: class)]]
+								with: class)) 
+						ifTrue: [ self pushUpVariable: each ]
+						ifFalse: [ self refactoringError: 'Do you want push up anyway?' ] ]]
```
Ducasse: You could use the following text: 

'You are about to push your method without the instance variable it uses. It will bring the system is an inconsistent state. But this may be what you want. 
So do you want to push up anyway?'

Ducasse:
Evelyn I rephrase the message.

Ducasse:
Hi Evelyn did you see my comment above?

PR #8535, opened 2021-02-08
8353-startStepping-is-calling-an-undefined-method 

### LINKED ISSUE #8353: startStepping is calling an undefined method
```
startStepping: aMorph
	"Add the given morph to the step list. Do nothing if it is already being stepped."
	^self startStepping: aMorph at: Time millisecondClockValue selector: #stepAt: arguments: nil stepTime: nil
```

So it may means that the stepping mechanism of Morph is not working anymore.

MarcusDenker:
move two methods down in the hierachy, fixes #8353


kasperosterbye:
Looks good to me - as discussed in issue #8353, two methods was moved down the hierarchy.

PR #8539, opened 2021-02-09
8503-Slot-remove-in-Stateful-Traits

### LINKED ISSUE #8503: Slot remove in Stateful Traits
**Describe the bug**
instance variable removal ( -- ), see TaRemoveSlot needs a test an fix

MarcusDenker:
- add tests for deletion of slots from a Trait: #testChangingDeleteSlot and #testChangingDeleteSlot2Slots
- TaRemoveSlot: change TaRemoveSlot to use a collection of vars to delete from the Trait (as it can define multiple slots)
- add CDTraitSlotRemoveNode to make the classparser work

fixes #8503


SimonXHolland:
Thanks, Marcus! Will have a play.

PR #8543, opened 2021-02-09
8042 deprecate and remove pointdist because there is pointdistance to with exactly the same implementation

### LINKED ISSUE #8042: Deprecate and remove Point>>dist: because there is Point>>distanceTo: with exactly the same implementation
Both methods are implemented in Pharo 9:

- `Point >> dist:` (7 senders)
- `Point >> distanceTo:` (49 senders)

```Smalltalk
Point >> dist: aPoint 
	"Answer the distance between aPoint and the receiver."
	| dx dy |
	dx := aPoint x - x.
	dy := aPoint y - y.
	^ (dx * dx + (dy * dy)) sqrt
```
```Smalltalk
Point >> distanceTo: aPoint 
	"Answer the distance between aPoint and the receiver."
	| dx dy |
	dx := aPoint x - x.
	dy := aPoint y - y.
	^ (dx * dx + (dy * dy)) sqrt
```

`Point >> dist:` should be deprecated:

```Smalltalk
Point >> dist: aPoint 
	"Answer the distance between aPoint and the receiver."
	
	self
		deprecated: 'Use #distanceTo: instead'
		transformWith: '`@rec dist: `@arg' -> '`@rec distanceTo: `@arg'.
		
	^ self distanceTo: aPoint
```

Additionally, all 7 senders of `Point >> dist:` must be rewritten to use `distanceTo:`

dupriezt:
Fixes #8042 

 - Deprecated Point>>#dist: (in favour of Point>>#distanceTo:)
- Removed test PointTest>>#testDist (it was the same as PointTest>>#testDistanceTo)
- Senders of Point>>#dist: rewritten to use Point>>#distanceTo:

### DIFF
--- a/src/Athens-Morphic/PolygonMorph.extension.st
+++ b/src/Athens-Morphic/PolygonMorph.extension.st
@@ -118,7 +118,7 @@ for the current vertice C.
 			next := vert atWrap: i + 1.
 			tangent := (next - prior) / 2.
 			lenpc := current distanceTo: prior.
-			lencn := next dist: current.
+			lencn := next distanceTo: current.
 			lenpc = 0 ifTrue:[
 				ctrl1 := prior]
 			ifFalse:[	ctrl1 := current - (tangent / (1 + (lencn / lenpc)))].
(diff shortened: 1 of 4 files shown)


Ducasse:
Thanks!!

PR #8545, opened 2021-02-10
Fixes #4011 Character nextObject crashes Pharo too

### LINKED ISSUE #4011: $x nextObject crashes Pharo too
A VM related problem was discussed for Squeak:

http://lists.squeakfoundation.org/pipermail/vm-dev/2019-July/031175.html

but it crashes Pharo too.

kasperosterbye:
A very simple PR, which just prevent the image from crashing in two very unlikely and specific cases.
A test was added as well, to make sure nothing bad is happening. Before this PR, this test would crash the image.

### DIFF
--- a/src/Kernel-Tests/ProtoObjectTest.class.st
+++ b/src/Kernel-Tests/ProtoObjectTest.class.st
@@ -192,6 +192,20 @@ ProtoObjectTest >> testIsNil [
 	self deny: ProtoObject new isNil
 ]
 
+{ #category : #tests }
+ProtoObjectTest >> testNextObject [
+	"In issue 4011 (https://github.com/pharo-project/pharo/issues/4011), "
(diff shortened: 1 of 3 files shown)


Ducasse:
Thanks I asked Pablo to look at it so that we can discuss it. 
I reviewed it but I would like to know more and probably all immediate objects should not support nextObject and nextObject is a scary method that should only be used in specific case.

kasperosterbye:
I tried to find the documentation for which immediate objects exist in Pharo. The only place I found something was in [Pharo by example chapter 58](https://ci.inria.fr/pharo-contribution/job/UpdatedPharoByExample/lastSuccessfulBuild/artifact/book-result/PharoObjectModel/PharoObjectModel.html), where it seemed to be those three (char, int, float).

PR #8557, opened 2021-02-11
Form class>>#services changed to allow easy extension of the options …

### LINKED ISSUE #8556: Options for when an image file is drag'n'dropped onto a pharo image are not easily extendable
When a user drags and drop an image file (jpeg, png...) onto a pharo image, two choices are offered.
image

These choices are defined by the `Form class>>#services` method.
```smalltalk
services
	^ Array 
		with: self serviceOpenImageInWindow
		with: self serviceImageAsBackground 
```

Issue: This method hard-codes the two possible services.
It would be nice if pharo applications could dynamically add other options for what to do when an image file is received.
For example, `Form class>>#services` could instead look for all methods with a given pragma (#formService?), so that other applications would just have to add extension methods with that pragma to the `Form` class to add other options.

dupriezt:
…when an image file is drag'n'dropped onto the image

Fixes #8556

### DIFF
--- a/src/Graphics-Files/Form.extension.st
+++ b/src/Graphics-Files/Form.extension.st
@@ -113,7 +113,7 @@ Form >> resourceTag [
 { #category : #'*Graphics-Files-FileRegistry' }
 Form class >> serviceImageAsBackground [
 	"Answer a service for setting the desktop background from a given graphical file's contents"
-
+	<formService>
 	^ SimpleServiceEntry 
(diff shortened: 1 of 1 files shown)


dupriezt:
The test failures seem unrelated

Ducasse:
Yes nice addition most of the services would benefit from this. And the pragma lookup is self contained so this is good.

PR #8559, opened 2021-02-11
TaRenameSlot-Use-TaRenamedSlotWrapper

MarcusDenker:
TaRenameSlot was copying the slot and then changing the name.

I think it is better to make the composition explicit by adding TaRenamedSlotWrapper. This special slot wraps the original slot and changes the name, but it forwards everything else to the original slot.


Ducasse:
Marcus don't you want to have a couple of tests?

MarcusDenker:
There are tests: I added test for aliasing when I fixed it, they are in T2TraitWithAliasTest

MarcusDenker:
This PR just changes how it is implemented: instead of copying the slot to rename it, it wraps it instead

PR #8565, opened 2021-02-12
8564-Running-tests-from-the-command-line-does-not-execute-parametrized-cases

### LINKED ISSUE #8564: Running tests from the command line does not execute parametrized cases
Running tests from the command line does not execute parametrized cases.
The behavior should be the same as that in the TestRunner.

tesonep:
Fixes #8564


tesonep:
We have a problem: the limitation of the JUnit XML format parsed by Jenkins. It does not support having more information about the parametrized test.

guillep:
> We have a problem: the limitation of the JUnit XML format parsed by Jenkins. It does not support having more information about the parametrized test.

Interesting, I'll look for a spec of the JUnit XML format. Do you have a link? Would be nice to have it linked in the issue for future reference.

guillep:
There is one in form of an xsd here:

https://github.com/windyroad/JUnit-Schema

<blockquote><div> GitHub</div><div><strong><a href="https://github.com/windyroad/JUnit-Schema">windyroad/JUnit-Schema</a></strong></div><div>XSD for Apache Ant's JUnit output. Contribute to windyroad/JUnit-Schema development by creating an account on GitHub.</div></blockquote>

tesonep:
I had checked the same XSD.
Maybe we can try to see what happens if the method name is not a Java valid symbol.

guillep:
I don't think Jenkins will check that... We can try that...

Issue #2978, opened 2019-03-22
Process>>#isTerminated is Wrong


guillep:
A process that is on the bottom context, but whose context has already started but not terminated says that the process is terminated.

The following code (easily translated to a test) shows the problem.
We start a process from a manually made bottom context.
That context will stop in a semaphore.
While that context is suspended there, the process says it is terminated, and this prevents us from resuming it (because resuming asks if the process is terminated).

However, the process is clearly not terminated! We have some code after the semaphore that should be executed.

```smalltalk
masterSemaphore := Semaphore new.

context := [ 
	| semaphore |
	semaphore := Semaphore new.
	masterSemaphore	signal.
	semaphore wait.
	
	"Some more code that should be executed, so this process should not be terminated yet."
	self error: 'Throw exception'
	
] asContext.

p := Process forContext: context priority: 40.
p resume.

masterSemaphore wait.
self deny: p isTerminated.
```

StephanEggermont:
Broken since at least Pharo 2.0? Works in Squeak 5.2

guillep:
Another way to reproduce it. `newProcess` will create a process with a bottom context doing `[self value. Processor activeProcess]`. So in the following

```smalltalk
p := [ 1+2 ] newProcess.
p step.	
p isTerminated
```

The stepping operation will just push `self` to the stack, we are before the message send of `value`, but still the process says it is terminated.

aranega:
@guillep with @Larcheveque we checked a little bit the issue, and we found a strange comparison in the `#isTerminated` method. There is many comparison checking if the suspended context is dead, and the final one checks if the suspended context's `startpc` is smaller than the it's `pc`. This looks weird, because it should be almost always the case, unless the process as not started.

```smalltalk
isTerminated
	self isActiveProcess ifTrue: [ ^ false ].
	
	^ suspendedContext isNil or: [
		"If the suspendedContext is the bottomContext it is the block in Process>>newProcess.
		If so, and the pc is greater than the startpc, the block has already sent and returned
		from value and there is nothing more to do."
		suspendedContext isBottomContext and: [
			suspendedContext isDead 
			"The pc of the suspendedContext is set to nil in #terminate explicitly.
				Leaving this line in for safety."
			or: [suspendedContext pc > suspendedContext startpc ] ] ]   "HERE"
``` 

[...]

PR #8567, opened 2021-02-13
[Very critical fix] Processor>>#terminateRealActive

dionisiydk:
PR is a general fix against the image hanging due to the buggy code simulation in the debugger (during stepping). It should prevent the most critical kind of infinite recursion where it is impossible to recover with cmd+.

PR introduces two methods:
- Processor>>#terminateRealActive
- Processor>>realActiveProcess

The first method here is to terminate the active process that is **really** currently running despite on any effective process installed. 	
It is very critical to ignore the simulation trick of the debugger (the effective process) during normal completion of the process. This method ensures that every process is able to complete. 

The second method here is to suspend the **real** active process during the debugger request triggered from the unhandled error. 
When unhandled error is encountered the execution must be stopped. To stop the execution system needs to find an active process to suspend it. Therefore it is crucial to find the **real** running process. Otherwise the suspend will not stop the execution (it will suspend a wrong process). In most cases the return from the error signal will end up with infinite recursion triggering new debuggers at every step. 

Related changes here:
- Simulation over Process>>#suspend primitive just returns the receiver. We never need to perform real suspend during stepping because we either step over suspended process (a debugger precondition) or we are stepping the wrong process and it would hang the simulator (debugger UI process).
- Context>>#isDead takes into account isEndOfProcessTermination. Now we will consider the context as dead when it is the last completed context in the process
- Three tests to perform full simulation of process execution from the first instruction up to the final "self terminate" instruction of the process. They should be important invariants for any simulation changes:
  - using StepInto
  - using StepOver
  - using StepThrough


src/Kernel-Tests-Extended/ProcessTest.class.st
```diff
@@ -389,6 +389,39 @@ ProcessTest >> testNewProcessWith [
 	self assert: { receivedArgument1. receivedArgument2 } equals: passedArguments.
 ]
 
+{ #category : #tests }
+ProcessTest >> testNormalProcessCompletionWithLeftEffectiveProcess [
```
dionisiydk, line 490: This is a test which reproduces the image hanging. With PR test should pass.

src/Kernel/BlockClosure.class.st
```diff
@@ -510,6 +512,40 @@ BlockClosure >> on: exception fork: handlerAction [
 			  nil ]
 ]
 
+{ #category : #exceptions }
+BlockClosure >> on: exception fork: handlerAction return: answerAction [
```
dionisiydk, line 510: This method was packaged in UnifiedFFI. I think it should be closer to no #return: version.

src/Kernel/ProcessorScheduler.class.st
```diff
@@ -313,6 +313,31 @@ ProcessorScheduler >> terminateActive [
 	activeProcess effectiveProcess terminate
 ]
 
+{ #category : #'process state change' }
+ProcessorScheduler >> terminateRealActive [
```
dionisiydk, line 331: Read this comment for more details about the problem

dionisiydk:
SindarinDebugger and its tests include wrong assumption about process being terminated:
```Smalltalk
SindarinDebugger>>isExecutionFinished
	"Returns whether the debugged execution is finished"
	^ process isTerminating
```
Sindarin uses #terminating flag of the process as the process termination indicator. But it is wrong. 
There is explicit method #isTerminated which should be used there. 
@StevenCostiou Do you have a particular reason to use #terminating flag here?

StevenCostiou:
I do not know, this part was done by @dupriezt 
Changing to `#isTerminated` makes Sindarin tests failing, so I think it comes from there: either the sindarin debugger does not configure its debugging contexts properly, or the test miss something in their configuration.
Maybe @dupriezt can give more context?

dionisiydk:
> I do not know, this part was done by @dupriezt
> Changing to `#isTerminated` makes Sindarin tests failing, so I think it comes from there: either the sindarin debugger does not configure its debugging contexts properly, or the test miss something in their configuration.

Tests are failing because they put particular expectation on the termination context where debugger is supposed to detect the end of execution. Tests perform the stepping up to the few steps inside #terminateActive until the terminating is set. But the process is considered terminated when it is returned from the #value, not when it is performing #terminateActive. You can check it in workspace: 
```Smalltalk
p := Process
  forContext: 
     [self value.
     isTerminated := Processor activeProcess isTerminated.
     Processor terminateActive] 
  priority: Processor activePriority
p resume.
```
So the debugger should follow the #isTerminated logic. And tests needs to be fixed accordingly

dionisiydk:
I see the problem with isTerminated. The first stepInto operation in new process does not step into #value of target block. The context pc in that case is not the initialPC and the process think it corresponds to the return state. Somewhere in these places we need improvements

Ducasse:
Thanks you all. 
This is super important that we get this right.

Ducasse:
Hi denis 
the guys will come back to this issue and give you feedback. 
S.

dionisiydk:
> Hi denis
> the guys will come back to this issue and give you feedback.
> S.

Hi @Ducasse . It is very interesting issue. Maybe I will find time to write a blog post about it.

src/Debugger-Oups/OupsDebugRequest.class.st
```diff
@@ -55,8 +55,15 @@ OupsDebugRequest class >> newForContext: aContext [
...
+	"By default we assume that the debugger is requested from the process itself as the result of anException. In that case we must use #realActiveProcess instead of #activeProcess.
+	When debugger opens it must stop the current process. Otherwise the error signal will return to the sender and it will continue execution. Obviously it may cause an unexpected behavior and in worst case it may break the system completely (with some infinite recursion triggering debuggers at every step).
+	Therefore in this scenario we use #realActiveProcess to reliably get a really running process. 
+	For more details see #realActiveProcess and #terminateRealActive comments"
 	<debuggerCompleteToSender>
-	^ self basicNew exception: anException; yourself
+	^ self basicNew 
+			exception: anException;
+			process: Processor realActiveProcess;
+			yourself
```
dionisiydk, line 66: I will leave @StevenCostiou to refactor it later if needed. 
My suggestion is to make the process retrieval an explicit for the users and remove the lazy initialization.

src/Debugger-Oups/OupsDebugRequest.class.st
```diff
@@ -55,8 +55,15 @@ OupsDebugRequest class >> newForContext: aContext [
 
 { #category : #'instance creation' }
 OupsDebugRequest class >> newForException: anException [
```
dionisiydk, line 57: Please sync this changes with actual repository if it exists. I did not find a separate project for Debugger-Oops.

src/Sindarin-Tests/SindarinDebuggerTest.class.st
```diff
@@ -362,15 +362,16 @@ SindarinDebuggerTest >> testIsExecutionFinished [
...
 	scdbg
 		stepOver;
 		step;
 		step;
 		stepOver;
 		step;
-		stepOver;
-		step.
-	"Reached `self isActiveProcess` in Process>>#terminate"
+		stepOver.
```
dionisiydk: Due to the changes in termination method the number of steps here were fixed

src/Sindarin-Tests/SindarinDebuggerTest.class.st
```diff
@@ -362,15 +362,16 @@ SindarinDebuggerTest >> testIsExecutionFinished [
...
+	"Reached `Processor terminateRealActive` method.
+	It is a special termination method which is considered as the end for simulation and should not be stepped"
```
dionisiydk: Please sync this changes with actual repository if it exists. I did not find a separate project for Sindarin

dionisiydk:
Tests are unrelated.
And PR is ready for merge

tesonep (approved):
It looks sound. I will check to rerun the tests because I don't like the error in the FFI marshalling. But if everything goes fine (= this error disappear  I will integrate)

src/Kernel-Tests-Extended/ProcessTest.class.st
```diff
@@ -7,6 +7,36 @@ Class {
 	#category : #'Kernel-Tests-Extended-Processes'
 }
 
+{ #category : #running }
+ProcessTest >> setUp [ 
+	super setUp.
+	
+	self executionProcessMonitor disable
```
dionisiydk, line 14: Process monitoring of TestExecutionEnvironment should be disabled because tests here are about methods involving into monitoring logic.

src/Kernel/Context.class.st
```diff
@@ -629,6 +629,14 @@ Context >> doPrimitive: primitiveIndex method: meth receiver: aReceiver args: ar
 			ifFalse: [ ^self send: (arguments at: 2) to: (arguments at: 1) with: (arguments at: 3) lookupIn: (arguments at: 4) ]
 		].
 
+	primitiveIndex = 88 ifTrue: ["Process>>suspend"
```
dionisiydk, line 635: This is another important change to make debugging safe for stepping over #suspend method (there are cases when suspend is executed over real active process which is an actual UI process of the debugger).

src/Sindarin-Tests/SindarinDebuggerTest.class.st
```diff
@@ -362,16 +362,10 @@ SindarinDebuggerTest >> testIsExecutionFinished [
...
-		stepOver;
-		step;
-		stepOver;
-		step.
-	"Reached `self isActiveProcess` in Process>>#terminate"
-	self assert: scdbg isExecutionFinished
+	
+	[ scdbg isExecutionFinished ] whileFalse: [ scdbg stepOver ].
+
+	self assert: scdbg currentProcess isTerminated
```
dionisiydk, line 394: This change makes test independent from the actual process stack during the self termination.

@StevenCostiou please sync these changes back to actual repository if it exists

dionisiydk:
> It looks sound. I will check to rerun the tests because I don't like the error in the FFI marshalling. But if everything goes fine (= this error disappear I will integrate)

Hi @tesonep . Please review my final commit. It is the end of story here.

dionisiydk:
> > It looks sound. I will check to rerun the tests because I don't like the error in the FFI marshalling. But if everything goes fine (= this error disappear I will integrate)
> 
> Hi @tesonep . Please review my final commit. It is the end of story here.

One related test is failing. I am checking

isCzech:
> I see the problem with isTerminated. [...]

Guys, I've updated #2978 with a suggested fix to #isTerminated. Wondering if it might impact this as well.... Currently a lot of bottom context processes were wrongly reported #isTerminated as terminated.

isCzech:
> > I do not know, this part was done by @dupriezt
> > Changing to `#isTerminated` makes Sindarin tests failing, so I think it comes from there: either the sindarin debugger does not configure its debugging contexts properly, or the test miss something in their configuration.
> 
> Tests are failing because they put particular expectation on the termination context where debugger is supposed to detect the end of execution. Tests perform the stepping up to the few steps inside #terminateActive until the terminating is set. But the process is considered terminated when it is returned from the #value, not when it is performing #terminateActive. You can check it in workspace:
> 
> ```smalltalk
> p := Process
>   forContext: 
>      [self value.
>      isTerminated := Processor activeProcess isTerminated.
>      Processor terminateActive] 
>   priority: Processor activePriority
> p resume.
> ```
> 
> So the debugger should follow the #isTerminated logic. And tests needs to be fixed accordingly

Inspired by your example I opened an issue #8656 about creating processes like this one - when you provide a closure instead of a context to #forContext:priority: it may cause a series of consequences for the image and potentially crash it. I'm not sure if your example just contains a mistake or was meant to be run as it is... but an issue is there regardless. Again, I was wondering it this issue might be related to your one.

src/Kernel/Context.class.st
```diff
@@ -973,9 +981,23 @@ Context >> isContext [
 
 { #category : #testing }
 Context >> isDead [
-	"Has self finished"
+	"Context is dead when
+	- it is self finished (returned to sender) 
+	- it is at the last instruction of process termination (it is unwinded and process is suspended)"
 
-	^ pc isNil
+	^ pc isNil or: [  self isEndOfProcessTermination ]
```
dionisiydk, line 977: And another important change. 
I checked users of #isDead. The critical senders were already accompanied with #isEndOfProcessTermination as part of this PR to fix existing problems. For the rest of users it also makes sense to take into account this condition. And probably it even fixes some bugs.

dionisiydk:
And finally green.
I completed changes from my side. It's ready for a final review and merge

tesonep:
Cool, I will check it

isCzech:
Hi, how will the new #isTerminated deal with a process that is at the last instruction of the bottom context? Like e.g. 
```p := Process forContext: [Processor activeProcess suspend] asContext  priority: 40. p resume```
Its context isn't dead but it's effectively done working, thus could be considered terminated; resuming it would raise an exception... Or am I missing something? Thanks, Jaromir

**Comment by dionisiydk** [comment]
> Hi, how will the new #isTerminated deal with a process that is at the last instruction of the bottom context? Like e.g.
> `p := Process forContext: [Processor activeProcess suspend] asContext priority: 40. p resume`
> Its context isn't dead but it's effectively done working, thus could be considered terminated; resuming it would raise an exception... Or am I missing something? Thanks, Jaromir

Hi @isCzech .

Such process will not be considered as terminated. And after resume it will really continue execution:
- it will return to nil sender
- vm will send #cannotReturn:
- #cannotReturn: will raise an error

The problems which I investigated and fixed here requires to have very concrete "end of process" to fix simulation/debugging of the execution up to the last instruction. Therefore PR introduces little restriction on the form of the process.
As everything it can be improved but it can be done later. We would need to avoid hardcoded #doTerminationFromYourself as a criteria for the "end process method". 

Do you have a **real** use case for #forContext: method which will be broken with my change?

**Comment by isCzech** [comment]
Hi, @dionisiydk 

> > Hi, how will the new #isTerminated deal with a process that is at the last instruction of the bottom context? Like e.g.
> > `p := Process forContext: [Processor activeProcess suspend] asContext priority: 40. p resume`
> > Its context isn't dead but it's effectively done working, thus could be considered terminated; resuming it would raise an exception... Or am I missing something? Thanks, Jaromir
> 
> Hi @isCzech .
> 
> Such process will not be considered as terminated.

Thanks, that's an interesting point; after all, terminated processes are the ones you declare terminated, it's not a given, it's a matter of a definition :) In my mind processes at their last instruction were terminated by default even without an explicit termination but maybe it just adds an unnecessary complexity...

> 
> * it will return to nil sender
> * vm will send #cannotReturn:
> * #cannotReturn: will raise an error
> 
> The problems which I investigated and fixed here requires to have very concrete "end of process" to fix simulation/debugging of the execution up to the last instruction. Therefore PR introduces little restriction on the form of the process.
> As everything it can be improved but it can be done later. 

A cosmetic change I'd suggest:
```
doTerminationFromYourself 
	"Stop this process forever from the process itself. 
	Unwind to execute pending ensure:/ifCurtailed: blocks before terminating.
	It assumes that self is the active process. "		
	thisContext unwindForTermination.
>>>	^self suspend    "add a return to remove the second last bytecode instruction"
```
which would simplify your final condition in `Context >> isEndOfProcessTermination` to:
	`^pc = self endPC`
	instead of `^pc = (self endPC - 1)`


> We would need to avoid hardcoded #doTerminationFromYourself as a criteria for the "end process method".
> 

I've had an idea how to avoid hardcoding #doTerminationFromYourself to your termination criteria: you can add a "trap" method, let's call it Process>>#terminated, and forward **all** terminated processes to it. Your #doTerminationFromYourself would send `self terminated` rather than `self suspend`:
```
doTerminationFromYourself 
	"Stop this process forever from the process itself. 
	Unwind to execute pending ensure:/ifCurtailed: blocks before terminating.
	It assumes that self is the active process. "		
	thisContext unwindForTermination.
>>>	^self terminated

terminated
	thisContext terminateTo: nil.   "set thisContext sender to nil"
	self suspend.
	^thisContext restart   "restart and suspend again in case of sending resume to a terminated process"
```
Although this approach doesn't avoid hardcoding a method selector as a criterion, it at least avoids hardcoding a reference to a bytecode position ;) 

With a bit of effort even #doTerminationFromAnotherProcess could forward terminated processes to the #terminated trap, unifying the two termination paths:
```
doTerminationFromAnotherProcess 
	...
	"replace the bottom line with:"
	(suspendedContext notNil and: [ suspendedContext isBottomContext ]) ifTrue: [
>>>		suspendedContext setSender: nil receiver: self method: (Process>>#terminated) arguments: {}]

isTerminated
	self isActiveProcess ifTrue: [ ^ false ].
	
	^ suspendedContext isNil or: [
		suspendedContext isDead or: [
			suspendedContext methodClass == Process and: [
				suspendedContext selector == #terminated ]]]
```
At the first glance all tests are green... 

> Do you have a **real** use case for #forContext: method which will be broken with my change?

No, except maybe if someone wanted to avoid #terminate because it automatically involves unwinds, but other than that, not really. Your slight change in termination criteria just raised my curiosity :)
best,
Jaromir

dionisiydk:
Hi @isCzech .

Thanks for your review.
#terminated trap is definitely interesting idea. I would use some fancy name for this like tombstone. 
But I don't want to push it here. It is already too many changes in current logic. So feel free to push it in another PR.

isCzech:
> Hi @isCzech .
> 
> Thanks for your review.
> #terminated trap is definitely interesting idea. I would use some fancy name for this like tombstone.
> But I don't want to push it here. It is already too many changes in current logic. So feel free to push it in another PR.

> We would need to avoid hardcoded #doTerminationFromYourself as a criteria for the "end process method".

Hi @dionisiydk ,

Thanks, I sent a PR #8845 as you suggested... Resisted the graveyard inspiration and changed the name of the trap to "#endProcess" ;)

isCzech:
Hi @dionisiydk ,
I've opened an issue #8992 (Unwind mechanism during termination is broken - skips multiple unwind blocks and corrupts the image) which addresses a problem that may be related to your issue. I wonder what impact the suggested fix of #8992 could have on this. It only touches #doTerminationFromAnotherProcess and leaves #isTerminated as (broken as) it is. It may surely be considered for merging (if at all) after your PR but still I thought you should know about the termination/unwind issues described in #8992 and PR #8993.
Thanks,
Jaromir

estebanlm:
same as with  #8993, we want this, but is very dangerous so I am deferring.

dionisiydk:
@tesonep could you please change the source branch to realActiveProcessMerged10 ? I do not have permissions for this.

dionisiydk:
I don't understand why CI merges the PR differently than github shows the changes.
For example #testIsExecutionFinished method is original one in the PR image from CI but here you can see that it is changed. 
The latest CI run was using a version of my code merged into Pharo10 branch. The result is still the same: wrong merge.
@tesonep do you have any idea what can be wrong? The problem appears after you switched PR to Pharo10 branch.

BTW I noticed that CI images are still with Pharo9.0 prefix in the names.

dionisiydk:
broken tests are still unrelated

dionisiydk:
So I found what is the issue. 
The confused merge was only for Sindarin packages. 
In previous Pharo versions they were copies from the official repo (Pharo-Spec/ScriptableDebugger). So that they we not downloaded for every image build. 
In Pharo 10 (maybe the end of Pharo 9) it was changed. Now every build downloads dependencies.

The confusing part is that you can't see this problem. The packages were not removed from this repo. And therefore it's impossible to notice that any code changes will be lost. The diff normally shows only my modifications. 
If packages would be properly cleaned it will be trivially to see what is going on: many packages will be marked as added or removed.

@estebanlm could you please review what other code is still left as a dead in Pharo repo. So that in future it will not kill the other brains.

dionisiydk:
I reported a separate issue #9822

dionisiydk:
The prerequisite for the integration is https://github.com/pharo-spec/ScriptableDebugger/pull/14
It requires new branch for ScriptableDebugger. It is now only master. Then integration this new version into the image. And only after that this PR can be validated again

tesonep:
We are seeing this PR with @StevenCostiou in the following weeks.
Do you have a breaking example? 
Is it just hanging when stepping or an error in the debugger?

dionisiydk:
> We are seeing this PR with @StevenCostiou in the following weeks.
> Do you have a breaking example?
> Is it just hanging when stepping or an error in the debugger?

The original issue was here: https://github.com/pharo-project/pharo/issues/8443. Just retry the example from the description.
This PR does not fix it. But the PR addresses a general issue why image freezes completely in such "broken debugger" scenarios (#8443 is a simulation bug). So with this PR the #8443 example is recoverable by cmd+. . 

Generally without this PR it is super hard to do any improvements/fixes to the debugger because any error there breaks the image completely.

dionisiydk:
Another example: try to step over a continuation resume. 

Generally just try to think for what code the debugger hooks to interrupt execution would not work (hooks in runUntilErrorOrReturnFrom: and some related methods). In all such cases the debugger leaves the process in inconsistent state where it will **never** stop execution.

Ducasse:
Thanks denis. Now every two weeks we sit together and review PRs.
We will have a look. We were also thinking that in Java a process cannot terminate another one (if I remember correctly).

dionisiydk:
> Thanks denis. Now every two weeks we sit together and review PRs.
> We will have a look. We were also thinking that in Java a process cannot terminate another one (if I remember correctly).

In Java it is really amazing shit I would say. You can't implement a behaviour which will control the life time of arbitrary processes. There is just no way to reliably kill the process with java API (at least it was the case in my Windows Java 7 time)

Ducasse:
A naive and honest question: what are the scenario where you need to have this?

dionisiydk:
> A naive and honest question: what are the scenario where you need to have this?

For example to control the time of process execution to protect the system from some hanging tasks which it serves.

isCzech:
Hi, I noticed Crash on step over #8443 issue has been mentioned here; I've just sent a PR #10383 to fix the root cause of the issue. Please check the explanation of the behavior.
Thanks,
Jaromir

guillep:

guillep:

PR #8571, opened 2021-02-15
Adding missing return to break the recursion.

tesonep:
Cutting the recursion to not try to get again if the string is smaller than the buffer. It is done in the recursive step.


guillep (approved):
👍

PR #8584, opened 2021-02-18
Use hashMultiply for SmallInteger>>hash

martinmcclure:
A few days ago, Nikhil Marathe made a blog post [1] about experiences with Pharo, asserting that "Dictionary access seems to be really slow in Pharo." Later discussion on Discord found that in Nikhil's application using an IdentityDictionary made the application perform ~100X faster (2009msec vs 188664 msec).

This did not sound right to me, so I took a look. It turns out that Nikhil's application uses integer keys. SmallInteger>>hash just returns self, whereas SmallInteger>>identityHash is returns self hashMultiply. This seemed likely to explain the huge performance difference. 

I asked Nikhil to try adding hashMultiply to SmallInteger>>hash, and sure enough, with that change IdentityDictionary and Dictionary perform similarly in Nikhil's application.

So here's the PR to make that one-word change.

[1] https://nikhilism.com/post/2021/experiencing-smalltalk/

### DIFF
--- a/src/Collections-Unordered-Tests/DictionaryTest.class.st
+++ b/src/Collections-Unordered-Tests/DictionaryTest.class.st
@@ -647,18 +647,21 @@ DictionaryTest >> testNewFromPairs [
 
 { #category : #tests }
 DictionaryTest >> testNilHashCollision [
-	"Ensures that fixCollisionsFrom: does the right thing in the presence of a nil key"
+
+	"Ensures that fixCollisionsFrom: does the right thing in the presence of a nil key.
+	To do that, we must have a key with the same hash as nil."
+
 	| dict key |
-	self supportsNilKey
-		ifFalse: [ ^ self ].
+	self supportsNilKey ifFalse: [ ^ self ].
 	dict := self collectionClass new.
-	key := nil hash. "any key with same hash as nil"
-	dict at: key hash put: 1.
+	key := ObjectWithSettableHash new.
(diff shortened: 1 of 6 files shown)


gcotelli (commented):
LGTM

noha (approved):

MarcusDenker:
Sadly there are failing tests, e.g. IdentityDictionaryTest

martinmcclure:
I'll take a look.

StephanEggermont:
Are there existing Dictionaries that need to be rehashed?

martinmcclure:
> Are there existing Dictionaries that need to be rehashed?

Yes, just making this change in an existing image freezes the image (for me, anyway). Going through the bootstrap process avoids this problem.

At least one test is written to rely on the hash of a SmallInteger being itself. I'm working on fixing that test, and will check for others.

martinmcclure:
Test fixes pushed. They pass for me, we'll see what Jenkins thinks.

martinmcclure:
OK, tests are happier now. :-)
After this change to SmallInteger>>hash, #testIsHealthy would fail in about one of five builds, depending on what identity hash was given to nil. The chances of a problematic nil identity hash are much lower, and even if that fails the test should not fail because in the larger table the hashes of 3 and 0 no longer hash to the same slot.

MarcusDenker (approved):

PR #8591, opened 2021-02-19
Unused-ClassVar-CleanUp-Deprecated

MarcusDenker:
remove unused class vars from CP1253TextConverter and LanguageEnvironment


Ducasse:
The broken tests are not related.

PR #8597, opened 2021-02-22
RB - Rename method only in specific packages

### LINKED ISSUE #6366: Pharo rename refactoring should update references only in local methods
**Describe the bug**
While using the rename classes refactoring, methods comming from traits referencing the classes get recompiled in the class. 

**To Reproduce**
- create a Trait X
- create a class Y using X
- create a class Z
- Add a method to X referencing Z
- Rename Z to anything

**Expected behavior**
I expect the reference in Y to be renamed but the method is recompiled in Y also. 

I suppose the refactoring is updating the references in all methods of the class instead of doing it only in local methods.

Present in Pharo 8.

### LINKED ISSUE #8513: Renaming should let me decide....
Imagine that you have two classes with method foo

When you do rename foo -> bar
The system says there are two implementors yes/no
if you say no

then you are stuck.
It would be interesting to have scoped rename that ask you the scope (package that should be impacted
by your refactoring). 

I can do it work in with scope package. 
But it would be good to have it as a menu item.

Lin777:
Add option to rename methods only in specific packages.

Fixes #6366
Fixes #8513

### DIFF
--- a/src/Calypso-SystemTools-Core/ClySystemBrowserContext.class.st
+++ b/src/Calypso-SystemTools-Core/ClySystemBrowserContext.class.st
@@ -172,6 +172,14 @@ ClySystemBrowserContext >> refactoringScopes [
 ]
 
 { #category : #'user requests' }
+ClySystemBrowserContext >> requestMultiplePackages: queryTitle [
+
+	^ tool searchDialog 
+		requestMultipleObjects: queryTitle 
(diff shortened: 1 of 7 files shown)


MarcusDenker:
RBRemoveDirectAccessToVariableTransformationTest>>testRefactoring is failing

PR #8600, opened 2021-02-22
Cleanup-unused-classvars-FT2Handle 

MarcusDenker:
remove a class var that is not referenced in FT2Handle


kasperosterbye:
I have not been able to find any usage of those class variables either, so I think it is safe.

Unrelated to the issue, most cases of classVariables are bad design.

PR #8609, opened 2021-02-22
RB - Improve convert temporary to instance variable refactoring

### LINKED ISSUE #3543: Refactoring idea: Make instance variable
I have a test class where all classes use a `log` temporary variable. I would like to make it an instance variable.

Maybe we could have a "Make instance variable" refactoring?

Lin777:
Fixes #3543

With the new features, this refactoring now works like this:

Script refactoring:
```
(RBTemporaryToInstanceVariableRefactoring 
    class: MyClassA
    selector: #someMethod
    variable: 'log') execute
```

Before refactoring:
``` 
Object subclass: #MyClassA
	instanceVariableNames: ''
	classVariableNames: ''
	package: 'example'

MyClassA >> someMethod 
    |log aNumber|
    log := self newLog.
    log isNil.
    aNumber := 5.

MyClassA >> anotherMethod
    #(4 5 6 7) do: [:e | | log |
        log := e ]

MyClassA subclass: #MyClassB
	instanceVariableNames: 'log'
	classVariableNames: ''
	package: 'example'
```

After refactoring:
``` 
Object subclass: #MyClassA
	instanceVariableNames: 'log'
	classVariableNames: ''
	package: 'example'

MyClassA >> someMethod 
    | aNumber |
    log := self newLog.
    log isNil.
    aNumber := 5.

MyClassA >> anotherMethod
    #(4 5 6 7) do: [:e | 
        log := e ]

MyClassA subclass: #MyClassB
	instanceVariableNames: ''
	classVariableNames: ''
	package: 'example'
```

### DIFF
--- a/src/Refactoring-Core/RBTemporaryToInstanceVariableRefactoring.class.st
+++ b/src/Refactoring-Core/RBTemporaryToInstanceVariableRefactoring.class.st
@@ -67,6 +67,12 @@ RBTemporaryToInstanceVariableRefactoring >> class: aClass selector: aSelector va
 RBTemporaryToInstanceVariableRefactoring >> preconditions [
 	^(RBCondition definesSelector: selector in: class)
 		& (RBCondition definesInstanceVariable: temporaryVariableName asString in: class) not
+		& (RBCondition withBlock: [ 
+			(class allSubclasses anySatisfy: 
+				[ :cls | cls definesInstanceVariable: temporaryVariableName asString ])
+			ifTrue: [ self refactoringWarning: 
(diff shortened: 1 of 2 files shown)


Ducasse:
Hi Evelyn 

This is nice and I will integrate it. 
Now I was thinking that if we subclass already defines an instance with the same name we could ask for a new name.
This way we can then later check if we want to merge or not the methods to use the same name. 

BTW I have the impression that we would like a Refactoring: ReplaceUsersOfInstanceVarXByY
Do you see what I mean. 
This way we could have two instance variables + users and be able to merge them. 
Right now we have to do it by hand. We should open a new issue with it.

PR #8610, opened 2021-02-23
validate callback failure on exit

estebanlm:
... and try to answer default value if callback failed.

this improves TFFI stability but it does not prevents all crashes (because a default answer can be a wrong answer a lot of times)

### DIFF
--- a/src/ThreadedFFI/FFICallback.extension.st
+++ b/src/ThreadedFFI/FFICallback.extension.st
@@ -1,6 +1,24 @@
 Extension { #name : #FFICallback }
 
 { #category : #'*ThreadedFFI' }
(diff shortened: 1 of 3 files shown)


tesonep:
@estebanlm are some changes missing? Like the users of the methods? 😄

estebanlm:
not really since the users are for the moment in the Gtk backend (I know, I should push them up :P)

Ducasse:
@estebanlm Just a novice question, why do you initialise by default success and not failed

estebanlm:
because by default is success. A callback fails just if an error that pops up a debugger happens, which is not very frequent (hopefully).

guillep (approved):
Ok!

PR #8616, opened 2021-02-23
Rule for setUp and tearDown should be applied for TestResource too

### LINKED ISSUE #8615: Rule for setUp and tearDown should be applied for TestResource too
there are two rules 
- ReShouldSendSuperSetUpAsFirstMessage
- ReShouldSendSuperTearDownAsLastMessage

Both should also work for **TestResource** subclasses, not only for **TestCase** subclasses as 
they also provide a #setUp and #tearDown where lint should check for similar call order

astares:
Fix  #8615


MarcusDenker:
Failing test:

Error
If a class has a method in its trait composition, the method should not be implemented in the class. Violations:  - ReShouldSendSuperTearDownAsLastMessage - ReShouldSendSuperSetUpAsFirstMessage
Stacktrace
TestFailure
If a class has a method in its trait composition, the method should not be implemented in the class.
Violations: 
- ReShouldSendSuperTearDownAsLastMessage
- ReShouldSendSuperSetUpAsFirstMessage
ReleaseTest>>assertValidLintRule:withExceptions:
ReleaseTest>>testLocalMethodsOfTheClassShouldNotBeRepeatedInItsTraits

astares:
Fixed

MarcusDenker (approved):

PR #8619, opened 2021-02-23
RB - Refactor methods to use Factory class instead of RB*Change classes directly

### LINKED ISSUE #3176: RB*Change should not shortcut the factory.
RB*Change have a factory but only the change manager use it. 
Each change directly refer to other changes. 
The change hierarchy should access to the refactoryfactory and it should be used
else it defeats the purpose.

Lin777:
Fixes #3176


Ducasse:
Yes it was clearly a violation of the architecture. Either using or not the factory but all following the same.

PR #8633, opened 2021-02-25
improve browse history when a method belongs to multiple packages

estebanlm:
improve the browse of versions by allowing selection of repository if method to query is present in more than one.


Ducasse:
The only breaking test is not related.

PR #8651, opened 2021-02-26
8595 enhance debugger to show the dnu call in the stack

### LINKED ISSUE #8595: Enhance Debugger to show the DNU call in the stack
**Describe the request**
I concede that it is nice that the DNU method is not needed all the time and it can loose beginner in the debugging process. However, I think it should be still nice to have it in the stack. First, one may want to debug the DNU method (a custom one for instance). Or second, access to the instance on which the DNU has been called:
image
In the above example, Object new cannot be inspected. I do no want to create a new "toto" method but browse the api of the "Object new" to find the existing equivalent.

**Expected behavior**
Either do not trim the stack and point to the current method (not the DNU one), or show the current stack top.

**Expected development cost**
1h for Steven

**Version information:**
Pharo9.0.0
build: 1145, commit: 2990df5

StevenCostiou:
Fixes #8595

When the debugger opens on an exception, now provides helper contextual objects to see:
- the exception
- relevant objects provided by the exception (*e.g.*, the receiver of a MNU in case of a DNU)

Screenshot 2021-02-26 at 16 31 52

### DIFF
--- a/src/NewTools-Debugger-Extensions/Exception.extension.st
+++ b/src/NewTools-Debugger-Extensions/Exception.extension.st
@@ -7,13 +7,7 @@ Exception >> signalContext [
 ]
 
 { #category : #'*NewTools-Debugger-Extensions' }
-Exception >> stDebuggerInspectorNodes [
+Exception >> stDebuggerInspectorNodesFor: aStDebuggerContext [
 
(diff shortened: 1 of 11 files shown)


Ducasse:
This looks cool.

PR #8658, opened 2021-02-28
Fix ZnBufferedReadStream>>#setToEnd for ZipArchive usage

### LINKED ISSUE #8657: ZipArchiveError 'can''t find EOCD position' while using Tabular XLSL importer
Looks like the same as https://github.com/VincentBlondeau/Tabular/issues/2

Trying to import an XLSL file with tabular. The file is somewhat big (around 1MB), and opens fine as a zip file. Looking in an hex editor, the End Of Central Directory is there.

 - OS: Windows10
 - Version: Pharo 8 and also latest 9.

Issue happens in `ZipArchive class>>#findEndOfCentralDirectoryFrom:
`
to get the size of the file, it does:
```
	stream setToEnd.
	fileLength := stream position.
```

`ZnBufferedReadStream>>#setToEnd` does not set the buffered position when called. The size reported is smaller than it is actually and the EOCD is beyond that position.

terpon:
Fixes #8657

### DIFF
--- a/src/Zinc-Character-Encoding-Core/ZnBufferedReadStream.class.st
+++ b/src/Zinc-Character-Encoding-Core/ZnBufferedReadStream.class.st
@@ -314,7 +314,7 @@ ZnBufferedReadStream >> readInto: collection startingAt: offset count: requested
 { #category : #accessing }
 ZnBufferedReadStream >> setToEnd [
 	
-	stream setToEnd
+	self position: stream size
 ]
 
 { #category : #private }


Ducasse:
Thanks!!!

guillep:
Thanks for the contribution. I'm checking how to add a test for this now.

guillep:
I opened a separate issue to create a test. If somebody wants to give a try on that, it would be great.

https://github.com/pharo-project/pharo/issues/10010

PR #8675, opened 2021-03-01
8544 new debugger impossible to step into debugger complete to sender methods

### LINKED ISSUE #8544: [New Debugger] Impossible to step into <debuggerCompleteToSender> methods
**Describe the bug**
With new debugger it is impossible to debug the implementation of low level methods marked with <debuggerCompleteToSender>.
For example debug "self error: 'test'" expression and step into #error: method. The debugger will not show the callee context and it will stay like nothing happens.

**Expected behavior**
StepInto should always allow to dive into low level parts.

Originally #debuggerCompleteToSender pragma was introduced to skip the too low level parts of stack when debugger just shows (like to not show Halt machinery when "self halt" is triggered).
But now it seems that the new debugger applies such filters all the time for every step/update operation. That's wrong.

StevenCostiou:
Fixes #8544

Dynamic filtering of the stack:
- stepping **into** kernel code is now possible
- stepping **over** filters stack if we're not into kernel code (if we are, filtering is off)
- global disabling of stack filtering overrides the dynamic filtering
- button in the top-right side of the debugger stack to toggle stack filtering on/off

About the button:
- there is no mention if the stack is currently filtered or not
- it is not explicit what the button do, we have to look at the fly-by-help to see it
Those last two points are not perfect, but to me this can work as is for now as it should satisfies expert developers while begginners will not care. The other option was a checkbox, but it will not work as is in the spec presenter.
I suggest to merge for now and to open another usability issue for that particular point. This could make a nice introduction issue for new contributors.

Short demo:

https://user-images.githubusercontent.com/26929529/109515435-5b034480-7aa7-11eb-9397-85bc166588bd.mov

### DIFF
--- a/src/NewTools-Debugger-Commands/StDebuggerToggleFilterStackCommand.class.st
+++ b/src/NewTools-Debugger-Commands/StDebuggerToggleFilterStackCommand.class.st
@@ -1,0 +1,24 @@
+"
+I toggle the stack filtering in the debugger on and off, and I trigger a debugger update.
+"
+Class {
+	#name : #StDebuggerToggleFilterStackCommand,
+	#superclass : #CmCommand,
+	#category : #'NewTools-Debugger-Commands'
+}
+
+{ #category : #default }
+StDebuggerToggleFilterStackCommand class >> defaultDescription [
+	^'Toggle stack filtering on and off'
+]
+
+{ #category : #default }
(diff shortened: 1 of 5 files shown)


dionisiydk:
I did not try in live. But it definitely looks good. Thanks @StevenCostiou

PR #8677, opened 2021-03-01
7756 broken class definitions combining traits and slots

Ducasse:


Ducasse:
This is wrong, it includes another commit so I close it and I will redo it.

PR #8681, opened 2021-03-03
8680-Calypso-is-showing-bad-the-shortcuts

tesonep:
Making Calypso to use the KM printer and improving the KMShortcutPrinter implementation to handle combinations and single shortcuts


Ducasse:
To me this looks nice.

PR #8682, opened 2021-03-03
7756 broken class definitions combining traits and slots

Ducasse:


Ducasse:
Still includes the other commits

PR #8690, opened 2021-03-04
8689-DropListBox-is-rendering-the-text-on-top-of-the-button

tesonep:
RubAbstractTextArea should delegate to the superclass when setting the extent and not in wrapping mode.


Ducasse:
Broken is unrelated

PR #8692, opened 2021-03-04
Activate smart characters even when in completion position

### LINKED ISSUE #8687: Smart characters does not work properly in some cases with spaces
**Describe the bug**

Sometimes smart characters are not inserted, and instead selection is replaced by a single character.

**To Reproduce**

1. write down `a  b  c` (notice there are two spaces between a and b, and between b and c.)
2. select `  b` (with the two spaces, starting just after the a)
3. type a smart character (e.g., opening parenthesis or quote)
4. => the selection is replaced by the typed character

**Expected behavior**

 => the selection is surrounded by the typed character

Also, notice that if selection is between the spaces, it works!
1. write down `a  b  c` (notice there are two spaces between a and b, and between b and c.)
2. select ` b` (with only one space, starting just after the first space)
3. type a smart character (e.g., opening parenthesis or quote)
4. => the selection is surrounded by the typed character

guillep:
Fixes #8687 
With Tests

### DIFF
--- a/src/NECompletion-Morphic/CompletionEngine.class.st
+++ b/src/NECompletion-Morphic/CompletionEngine.class.st
@@ -187,8 +187,8 @@ CompletionEngine >> handleKeystrokeBefore: aKeyboardEvent editor: anEditor [
 	keyCharacter := aKeyboardEvent keyCharacter.
 	controlKeyPressed := aKeyboardEvent controlKeyPressed.
 	
-	self editor atCompletionPosition ifFalse: [ 
-		^ (self smartInputWithEvent: aKeyboardEvent ) notNil ].	
+	(self smartInputWithEvent: aKeyboardEvent )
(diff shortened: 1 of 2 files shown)


astares:
@guillep Could we become a little bit more descriptive in issues so other can reproduce too what issues are about ... what is situation, how to solve it and what should be the final result ...

guillep:
Well, I was pretty informative in #8692, no? What is missing there?

guillep:
Sorry, I actually mistook the issue number :)

guillep:
Please, check #8687 and tell me what you would like (I could add the Pharo build, yes, but it happens since a while :) ...)

astares:
@guillep OK - the corrected issue number includes a good description. Thanks!

Ducasse:
Broken tests are unrelated

PR #8696, opened 2021-03-05
8641-valueWithExit-should-answer-the-result-of-value

### LINKED ISSUE #8641: #valueWithExit should answer the result of #value:
Current implementation of #valueWithExit returns either nil (if the receiver exited using the exit block) or the receiver:

`[:exit | 1 ] valueWithExit`  answers: "[ :exit | 1 ]" instead of just 1.

It should answer the result of #value: (or nil in case of a non-local return).

Fix:
```
BlockClosure>>#valueWithExit
-  self value: [ ^nil ]
+  ^self value: [ ^nil ]
```

Pharo 9.0.0
Build information: Pharo-9.0.0+build.1162.sha.3b661311d4f2c35cfb60d371924f9754770a1414 (64 Bit)

MarcusDenker:
- add test
- fix as described in the issue

fixes #8641

### DIFF
--- a/src/Kernel-Tests/BlockClosureTest.class.st
+++ b/src/Kernel-Tests/BlockClosureTest.class.st
@@ -474,6 +474,11 @@ BlockClosureTest >> testValueWithArgumentsWithOrderedCollection [
 ]
 
 { #category : #'tests - evaluating' }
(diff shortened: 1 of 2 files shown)


MarcusDenker:
failing test not related

PR #8706, opened 2021-03-07
Fix #8656 Process class >> #forContext:priority: potentially crashing an image

### LINKED ISSUE #8656: Process class >> #forContext:priority: potentially crashing an image
`#forContext:priority:` may cause an image crash when supplying e.g. a block instead of a context:

(save your image first!)

create a new process:

   `p := Process forContext: ["whatever"] priority: Processor activePriority`

and then try to start a Process Browser or just do-it:

   `p isTerminated`

The thing is both Process Browser and #isTerminated send `suspendedContext isBottomContext` and that fails, starting a debugger and in some cases even windows start popping up (probably because processes with invalid "contexts" are still persisting in the image and something tries to deal with them... gc maybe?).

I guess the easiest and safest fix is to send #asContext to the aContext argument in #forContext:priority:
```
Process class >> forContext: aContext priority: anInteger 
	"Answer an instance of me that has suspended aContext at priority 
	anInteger."

	| newProcess |
	newProcess := self new.
	newProcess suspendedContext: aContext asContext.       "here's the fix"
	newProcess priority: anInteger.
	Processor activeProcess installEnvIntoForked: newProcess.
	^newProcess
```
If not then at least a condition has to be added to make sure the aContext argument really is a context.

Regards,
Jaromir

isCzech:


isCzech:
#8656-Process-class-#forContext-priority-potentially-crashing-image

guillep (approved):
Tests are green.

I like the change, it's not dangerous, and can catch many problems early on.

PR #8709, opened 2021-03-08
RB - Migrate some class commands to commander2

Lin777:
Migrate commands `SycAddSubclassCommand, SycInsertSubclassCommand and SycInsertSuperclassCommand` to commander2


MarcusDenker:
bootstrap fails with

```
  SycAddSubclassCommand
  SycAddSubclassCommand>>#fullBrowserTabActivation
```

PR #8712, opened 2021-03-08
Fix #8654 - Process>>#terminate should answer self consistently

### LINKED ISSUE #8654: Process>>#terminate causes a walk-back when terminating a terminated process
Process>>#terminate causes a walk-back when terminating a terminated process because it returns nil when signaling ProcessAlreadyTerminating. The method should always return self and repeated termination should be harmless, I guess.

```
Process>>#terminate
	"Stop the process that the receiver represents forever.  Unwind to execute pending ensure:/ifCurtailed: blocks before terminating."
	terminating := self isTerminating
-		ifTrue: [ ^ ProcessAlreadyTerminating signal ]         "replace this line by:"
+		ifTrue: [ ProcessAlreadyTerminating signal. ^self ]
		ifFalse: [ true ].

	self isActiveProcess 
		ifTrue: [ self doTerminationFromYourself ]
		ifFalse: [ self doTerminationFromAnotherProcess]
```

isCzech:
Fixes #8654

### DIFF
--- a/src/Kernel-Tests-Extended/ProcessTest.class.st
+++ b/src/Kernel-Tests-Extended/ProcessTest.class.st
@@ -458,6 +458,24 @@ ProcessTest >> testTerminateActive [
 	self deny: lastStatementEvaluated.
 ]
 
+{ #category : #'tests - creation' }
+ProcessTest >> testTerminateAnswersSelf [
+
+	| process |
+
+	"verify #terminate returns a process that is terminated"
+	process := [] newProcess terminate.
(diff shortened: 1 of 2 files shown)


dionisiydk (changes requested):
Good point. But please add a test for this.

guillep:
I'm with @dionisiydk here, would you mind adding some tests showing that multiple termination does indeed work properly?

isCzech:
will do asap, sorry

PR #8733, opened 2021-03-09
8732 class abstract tool is not really abstract

astares:


MarcusDenker:
MetacelloNotification: Loaded -> NewTools-MethodBrowsers-tonel.1 --- tonel:///builds/workspace/uest_and_branch_Pipeline_PR-8733/64/src --- cache
MetacelloNotification: Loading -> NewTools-ChangeSorter-tonel.1 --- tonel:///builds/workspace/uest_and_branch_Pipeline_PR-8733/64/src --- cache
  AbstractTool

astares:
@MarcusDenker But AbstractTool is provided (now as a subclass of the renamed class) as you see on the changed files

My guess is that CI only catches the first commit, not the full PR (with the second commit) and thats why it was failing

Can you restart it?

Ducasse:
Let us try to rebuild it. 
This is strange.

guillep (approved):
Good!

guillep:
I'm sorry, I'll revert this. I was in trance, I did not see there were dependency issues.

Of course, the deprecated package is loaded at the very end, to make sure there are no users of it in the build.

Looking at the logs, I'd say there are users of it in NewTools-ChangeSorter ?
@astares would you mind redoing a new PR? (I'd do it otherwise to make up from my mistake, but given that there is still work to do in this one...)

PR #8739, opened 2021-03-09
TaskIt: TKTRunner and TKTAbstractExecutor should be abstract

### LINKED ISSUE #8738: TaskIt: TKTRunner and TKTAbstractExecutor should be abstract
TKTRunner and TKTAbstractExecutor are not referenced. They are providing abstract
behavior and should implement #isAbstract to return true

astares:
Fix #8738

### DIFF
--- a/src/TaskIt/TKTAbstractExecutor.class.st
+++ b/src/TaskIt/TKTAbstractExecutor.class.st
@@ -15,6 +15,12 @@ Class {
 	#category : #'TaskIt-Kernel'
 }
 
+{ #category : #testing }
+TKTAbstractExecutor class >> isAbstract [
+
+	^self == TKTAbstractExecutor 
+]
+
 { #category : #accessing }
 TKTAbstractExecutor >> exceptionHandler [
 	^ exceptionHandler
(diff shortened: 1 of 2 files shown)


noha:
As long as we just copy projects into pharo we should take care either doing pull request on the original source or doing it in both. Otherwise it is very likely these changes get reverted on next copy. Repository is at https://github.com/pharo-contributions/taskit


<blockquote><div> GitHub</div><div><strong><a href="https://github.com/pharo-contributions/taskit">pharo-contributions/taskit</a></strong></div><div>TaskIt is a library that ease Process usage in Pharo. It provides abstractions to execute and synchronize concurrent tasks, and several pre-built mechanisms that are useful for many application dev...</div></blockquote>

MarcusDenker:
I do not think that this "I by chance know that the package has a repo but you can not know when fixing things in Pharo" will ever scale. It already now does not work.

noha:
Nobody is satisfied with the situation. But as long as we rely on metacello there is no escape. So we just need to minimize potential for regressions

astares:
@noha Yes - an unfortunate situation. Still one can not see what is managed internal and what external.

Ideally an external project aligns back with changes in Pharo to go more freely for the contributions. For instance Sven is now pulling back changes for Zinc and Zodiac.

Issue #8759, opened 2021-03-10
Deprecation of ClassDescription >> definition broke polymorphism with CompiledMethod >> definition


ClotildeToullec:
**Describe the problem**
```CompiledMethod >> definition``` is meant to be polymorphic with ```ClassDescription >> definition``` (see the method comment).
Since ```ClassDescription >> definition``` has been deprecated, there is no polymorphism anymore.

**Classes involved**
CompiledMethod

**Proposal**
Deprecate ```CompiledMethod >> definition``` and replace it by ```definitionString```.

**Version information:**
 - Pharo 9.0

**Expected development cost**
I'll make a PR

**Additional context**
This polymorphism is needed in Moose

PR #8760, opened 2021-03-10
In CompiledMethod, replace #definition by #definitionString

ClotildeToullec:
in order to keep polymorphism with ClassDescription >> definitionString


ClotildeToullec:
Related to #8759

MarcusDenker (approved):
Yes!

PR #8770, opened 2021-03-10
8659 expression leading to error is not highlighted anymore 

### LINKED ISSUE #8659: Expression leading to error is not highlighted anymore.
I do not know if this is a bug in the debugger, in the code/text mapping or the SpCodePresenter. 
Now when we have an error and the debugger displays the stack, then the expression is not highlighted
anymore.

StevenCostiou:
Fixes #8659

### DIFF
--- a/src/NewTools-Debugger/StDebugger.class.st
+++ b/src/NewTools-Debugger/StDebugger.class.st
@@ -1232,8 +1232,7 @@ StDebugger >> updatePredicatesForContext: aContext [
 StDebugger >> updatePresenter [
 
 	self updateStackFromSession: self session.
-	self buildContextMenus.
-	self updateCodeFromContext
+	self buildContextMenus
 ]
 
 { #category : #'updating actions' }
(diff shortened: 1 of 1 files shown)


Ducasse:
Steven if I remember correctly esteban told me that this is a bug he introduced while fixing another one. 
So I do not know if I should integrate this one. I will ask @estebanlm.

StevenCostiou:
> Steven if I remember correctly esteban told me that this is a bug he introduced while fixing another one.
> So I do not know if I should integrate this one. I will ask @estebanlm.

This was due to a refactoring we discussed with Esteban. 

The problem seems that the ui elements (toolbar, code) have to be updated after the window presenter has opened (why?).
This fixes that problem while being compatible with the refactoring.

I also requested a review from Esteban because I do not know if he will like it.

StevenCostiou:
Nevertheless, I would prefer that we accept that fix because it is really a problem that when the debugger opens you don't see the code highlight nor the toolbar buttons are up to date.

Esteban if you think it deserves more work, let us open an issue in the NewTools repository and take our time to understand and fix it there. Meanwhile users will not be annoyed by the problem.

Ducasse:
Ok with me.

PR #8772, opened 2021-03-10
Update UnhandledError to reflect the ability for the underlying Exception to resume.

kurtkilpela:
The changes in this pull request were suggested by @martinmcclure on Pharo's discord server. The changes enable unhandled exception processing techniques used in https://github.com/GemTalk/RemoteServiceReplication to function GemStone and Pharo.

### DIFF
--- a/src/Kernel/UnhandledError.class.st
+++ b/src/Kernel/UnhandledError.class.st
@@ -55,5 +55,5 @@ UnhandledError >> exception: anError [
 { #category : #testing }
 UnhandledError >> isResumable [
 	
-	^ false
+	^ true
 ]
--- a/src/Kernel/UnhandledError.class.st
+++ b/src/Kernel/UnhandledError.class.st
@@ -54,6 +54,10 @@ UnhandledError >> exception: anError [
 
 { #category : #testing }
 UnhandledError >> isResumable [
-	
-	^ true
+
+	"UnhandledError can be signaled for more than non-resumable Errors.
+	If UnhandledError reflects the underlying Exception, we are able to
+	resume the underlying Exception."
+
+	^ self exception isResumable
 ]
--- a/src/Kernel-Tests/ExceptionTest.class.st
+++ b/src/Kernel-Tests/ExceptionTest.class.st
@@ -53,6 +53,20 @@ ExceptionTest >> runWithNoHandlers: aBlock [
 ]
 
 { #category : #test }
+ExceptionTest >> testDefaultAction [
+	"Ensure that the resuming an UnhandledError when the underlying
+	Exception is resumable, results in the resumption value serving
+	as the result of the original #signal."
+
+	| expected result |
+	expected := #marker.
+	[ result := MyResumableTestError signal ]
+		on: UnhandledError
+		do: [ :ex | ex resume: expected ].
+	self assert: result equals: expected
+]
+
+{ #category : #test }
 ExceptionTest >> testDefaultDescription [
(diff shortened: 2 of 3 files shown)


Ducasse:
Thanks!

MarcusDenker:
just two timeouts of network on MacOS

MarcusDenker (approved):

dionisiydk:
You should add tests

src/Kernel/UnhandledError.class.st
```diff
@@ -55,5 +55,5 @@ UnhandledError >> exception: anError [
 { #category : #testing }
 UnhandledError >> isResumable [
 	
-	^ false
+	^ true
```
dionisiydk: It deserves a good comment why it must be resumable.
There is #resumeUnchecked: method to resume any exception. Would not it be better to use it explicitly for your goal?

Now it feels very odd to have false for Error>>isResumable and true for UnhandledError>>isResumable.

kurtkilpela: Using #resumeUnchecked: w/ the changes to #defaultAction and #raiseUnhandledError would be sufficient. It could also be worked around using #resignalAs:. 

I'll revert this method and add a test for the return value of #defaultAction.

kurtkilpela: After a bit of discussion w/ @martinmcclure, we decided it may be appropriate to have UnhandledError reusability reflect that of the underlying Exception. The changes in #defaultAction and #raiseUnhandledError allow the resumption value of UnhandledError to serve as the resumption value of the original #signal.

In the event that the underlying Exception is not resumable, an IllegalResumeAttempt is signal.

This would allow us to avoid the use of #resumeUnchecked: while making UnhandledError more generally useful.

dionisiydk (approved):
Looks great, thanks

dionisiydk (changes requested):
tests are related

kurtkilpela:
The tests I wrote in this PR appear to pass in pharo image created via Pharo Launcher. They don't appear to pass on the CI server. Is there something different about the CI server vs development image that would cause this behavior?

kurtkilpela:
Actually, the test can and does fail locally if I am running in the TestRunner. If I run the test directly by using the circle to the left of the test selector, the test passes.

kurtkilpela:
The difference is in TestResult>>#runCase: vs TestResult>>#runCaseForDebug:. Getting the tests to pass in both scenarios will require a bit more work. I mark this as a draft until I'm able to return to it next week.

dionisiydk:
> The difference is in TestResult>>#runCase: vs TestResult>>#runCaseForDebug:. Getting the tests to pass in both scenarios will require a bit more work. I mark this as a draft until I'm able to return to it next week.

Yes, the problem is that the test suite catches an error , reports it in a test result and continues to run the next test. In that case UnhandledError never happens because all errors are handled.
In case of single test in debug mode the exception is passed out after reporting. Therefore the execution continues up to  UnhandledError signal. 

You can find the solution to that problem in other tests. For example look at #testUnhandledErrorWhenNoHandlers. The trick is to run the code under test in a separate process where you can be sure that no outer handlers exist. See the method #runWithNoHandlers:

kurtkilpela:
Thank you @dionisiydk, your suggestion worked as expected.

dionisiydk (approved):
tests are unrelated

MarcusDenker (approved):

PR #8773, opened 2021-03-11
Added the possibility to create a new method in the class definition tab

sonibla:
In the Browser, in the class definition it's now possible to replace the class definition with a new or existing method.
One small issue: before accepting changes, the syntax coloring always thinks it's a class definition but it is acceptable.


Ducasse:
Oh yes!!!!
I will check because I always hated that we lost this possibility!
Thanks a lot.

Ducasse:
I reviewed the code and the implementation and I will integrate it. 
Now we will have to make sure that it works also in Fluid class definition mode. 
Right now it does not. I do not have the time to see why.

PR #8774, opened 2021-03-11
8694-Instance-of-DoItVariable-did-not-understand-markRead 

### LINKED ISSUE #8694: Instance of DoItVariable did not understand #markRead
I got this in the transcript.

Instance of DoItVariable did not understand #markRead
Instance of DoItVariable did not understand #markRead
Instance of DoItVariable did not understand #markRead
Instance of DoItVariable did not understand #markRead
Instance of DoItVariable did not understand #markRead
Instance of DoItVariable did not understand #markEscapingWrite
Instance of DoItVariable did not understand #markEscapingWrite
Instance of DoItVariable did not understand #markEscapingWrite
Instance of DoItVariable did not understand #markRead
Instance of DoItVariable did not understand #markRead

MarcusDenker:
DoItVariable needs to forward the read/write usage messages as it can wrap a temporary variable.

fixes #8694

### DIFF
--- a/src/Kernel/DoItVariable.class.st
+++ b/src/Kernel/DoItVariable.class.st
@@ -179,6 +179,31 @@ DoItVariable >> key [
 	^self name
 ]
 
+{ #category : #'read/write usage' }
+DoItVariable >> markEscapingRead [
+	actualVariable markEscapingRead
+]
+
+{ #category : #'read/write usage' }
+DoItVariable >> markEscapingWrite [
+	actualVariable markEscapingWrite
+]
+
+{ #category : #'read/write usage' }
+DoItVariable >> markRead [
+	actualVariable markRead
+]
+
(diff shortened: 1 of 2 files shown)


src/Kernel/DoItVariable.class.st
```diff
@@ -179,6 +179,31 @@ DoItVariable >> key [
 	^self name
 ]
 
+{ #category : #'read/write usage' }
+DoItVariable >> markEscapingRead [
+	actualVariable markEscapingRead
+]
```
dionisiydk, line 185: I don't think it is the right thing to do.
When DoItVariable represents an external temp the actual temp is already analyzed for escape operations. Therefore a compilation of doIts should not change it.
I think the correct approach is to do nothing in these methods. Leave them empty.

Issue #8778, opened 2021-03-11
`ChangeSet current currentMessageList` can not be opened with `Smalltalk tools messageList`


rolandbernard:
**Describe the bug**
Trying to open the message browser (`Smalltalk tools messageList`, `ClyOldMessageBrowserAdapter`) with the changed message list returned from the change set (`ChangeSet>>currentMessageList`), will throw `#realClass was sent to nil` **if** I previously changed a comment.

**To Reproduce**
1. Go to any Class (e.g. AST-Core > Formatter)
2. Change the comment for the class.
3. Executing the following (e.g. in the playground):
```smalltalk
Smalltalk tools messageList
	browse: ChangeSet current changedMessageList
	title: 'Current Change Set'
```

**Expected behavior**
The message browser should open with a list of the changed methods.

**Screenshots**

**Version information:**
 - OS: Windows 10
 - Version: tested in Pharo 9.0.0 and 8.0.0

**Expected development cost**
I don't think it requires a lot of work.

**Additional context**
The exception is thrown in `ClyOldMessageBrowserAdapter>>open` because it executes `each parent realClass`, but parent is nil. It seems to me that this could be replaced with `each realParent` (which handles the case where the parent is nil), but I don't know enough about Pharo to understand if this has the same semantics. (It definitely has the same behavior if the parent is not nil)

The stack:
```
UndefinedObject(Object)>>doesNotUnderstand: #realClass
UndefinedObject>>doesNotUnderstand: #realClass
[ :each | ClyClassComment of: each parent realClass ] in ClyOldMessageBrowserAdapter>>open in Block: [ :each | ClyClassComment of: each parent realClas...etc...
[ :stream | 
1 to: self size do: [ :index | 
	(selectBlock value: (each := self at: index))
		ifTrue: [ stream nextPut: (collectBlock value: each) ] ] ] in Array(SequenceableCollection)>>select:thenCollect: in Block: [ :stream | ...
Array class(SequenceableCollection class)>>new:streamContents:
Array(SequenceableCollection)>>select:thenCollect:
ClyOldMessageBrowserAdapter>>open
ClyOldMessageBrowserAdapter class>>browse:title:
UndefinedObject>>DoIt
OpalCompiler>>evaluate
RubSmalltalkEditor>>evaluate:andDo:
RubSmalltalkEditor>>highlightEvaluateAndDo:
[ textMorph textArea editor highlightEvaluateAndDo: ann action.
[...]

PR #8779, opened 2021-03-11
Changed ClyOldMessageBrowserAdapter>>open to using realParent to handle cases where parent is nil

rolandbernard:
My proposed change to solve Issue #8778.
I am not very familiar with Pharo, so I am not 100% sure this is the correct solution.
If parent is not nil the result will be the same, given the implementation of realParent.


MarcusDenker:
I did not yet really look at it, but my "gut feeling" is that it would be nice to not rely on the Ring concept of "parent". Even if we get a Ring object representing a comment, we should be able to ask for the class it is from, using an API that is not Ring special (as is the whole "parent" concept)

Ducasse:
I will try to allocate some time once I have finished with my current task.

MarcusDenker:
I had a look: as a fix it is good. The code checks if it is a ring comment, thus calling #realParent is ok. 

I propose to merge it, we can  then later continue cleaning ring APIs.

PR #8780, opened 2021-03-12
8185-When-fluid-class-RGClassDefinitionTest-has-one-test-failing

### LINKED ISSUE #8185: When fluid class RGClassDefinitionTest has one test failing
testAsClassDefinitionSourceDefinition

MarcusDenker:
easy to fix: we have to use #new so it uses the fluid printer when in fluid mode

fixes #8185


Ducasse:
funnily I did it on purpose that it selected the old format. 
Thanks

PR #8781, opened 2021-03-12
RB - Introduce new refactoring "Copy package"

### LINKED ISSUE #8568: Copy (updating scoped references)
I would love to copy a full package and get all the classes referencing each other 
renamed too, referencing to the copied classes. 

We could also make that if we select a group of classes and copy them all the get the same behavior. 

So we could have 
   surface copy: just copy the classes and the methods
   scope copy: surface copy + references updated

Lin777:
Fixes #8568

This new refactoring consists in copy a complete package with all defined classes inside it, if the copy classes has references between them, these references are renamed also. For example:

The following package has these classes, 

imagen

and these classes has these references between them:
```
MyClassA >> foo
	^ MyClassB

MyClassB >> example4 
	#MyClassA1
```

After copy #Example package we obtain this result:

imagen

and references were updated

```
MyClassACopy >> foo
	^ MyClassBCopy

MyClassBCopy >> example4 
	#MyClassA1Copy
```

It can also be noted that the hierarchy was correctly updated with the copied classes

### DIFF
--- a/src/Calypso-SystemTools-FullBrowser/SycCopyPackageCommand.extension.st
+++ b/src/Calypso-SystemTools-FullBrowser/SycCopyPackageCommand.extension.st
@@ -1,0 +1,15 @@
+Extension { #name : #SycCopyPackageCommand }
+
+{ #category : #'*Calypso-SystemTools-FullBrowser' }
+SycCopyPackageCommand class >> fullBrowserMenuActivation [
+	<classAnnotation>
+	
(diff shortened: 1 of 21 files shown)


Ducasse:
Hello Evelyn 

This is test is not working 
 testRefactoring – MacOSX64.Refactoring2.Transformations.Tests.RBRemoveDirectAccessToVariableTransformationTest

PR #8784, opened 2021-03-14
8669 aFloat asScaledDecimal loses precision

pablo-s:
In class Float, the method asScaledDecimal was overriden to set a custom scale (14) for floats.

The reasons for this number are:
1) IEEE 754 double precision numbers use 53 bits to encode mantissa. This means that the maximum amount of decimal digits that can be represented is log10(2^53) = 15,xx.
2) At least one of the decimal digits has to be placed to the left of the decimal point.

Therefore, if we round down the max amount of digits and we remove the digit that will not be to the right of the decimal point, we obtain a scale of 14.

### DIFF
--- a/src/Kernel-Tests/FloatTest.class.st
+++ b/src/Kernel-Tests/FloatTest.class.st
@@ -13,6 +13,23 @@ FloatTest >> classToBeTested [
 	^ Float
 ]
 
+{ #category : #test }
+FloatTest >> testAsScaledDecimal [
+
+	self
+		assert: 0.00000000000006 asScaledDecimal asString
+		equals: '0.00000000000006s14'.
+	self
+		assert: 9.99999999999999 asScaledDecimal asString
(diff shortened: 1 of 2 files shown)


pablo-s:
Updated the description of this PR with more details.

Ducasse:
Thanks for your contribution.

guillep:
Hi @pablo-s this is great! It could be nice to add some tests and commentd in the PR. For example, documenting why you chose 15, and why it is enough, would be nice. Otherwise this is just a magic number and its rationalite is lost...

pablo-s:
Thanks to all of you for your comments!

@guillep I have updated the scale value and I also included the reasoning. You can find it in: the first comment of this PR, and also in the method Float >> asScaledDecimal, where the "magic number" is used. I also included some tests.

Please, let me know if anything else is needed.

guillep (approved):
Good!

PR #8826, opened 2021-03-18
use new evaluation method (fixes playground)

estebanlm:
fixes a problem introduced with the merge of spec 0.8.5


estebanlm:
@MarcusDenker this is the fix for the problems I introduced

MarcusDenker (approved):

PR #8842, opened 2021-03-19
6970-win32ModuleName-not-implemented

### LINKED ISSUE #6970: win32ModuleName not implemented
latest Pharo 9

image

MarcusDenker:
I think this is a leftover from the transition. We can just make turn it into a subclass-responsability now

fixes #6970


tesonep (changes requested):
We are keeping that as backward compatibility, in case anyone loads a FFILibrary done with the old API.

MarcusDenker:
Ok, I will change the PR to add a comment instead

StephanEggermont (approved):

PR #8845, opened 2021-03-21
Fix #8844 Avoid hardcoded #doTerminationFromYourself as a criteria

### LINKED ISSUE #8844: Avoid hardcoded #doTerminationFromYourself as a criteria for the isTerminated method
As as complementary proposal to PR #8567 I'd like to suggest a way to avoid hardcoding #doTerminationFromYourself and a particular position of the pc pointer to the termination criteria in the upcoming PR. It involves adding a "trap" method, let's call it Process>>#endProcess, and forwarding every processes being terminated to it. #endProcess would suspend the receiver and use a new <isTerminated> pragma to indicate the termination status of the receiver.

The idea has been also presented in #8567 (https://github.com/pharo-project/pharo/pull/8567#issuecomment-802321883).

In case of accidental resuming a terminated process the #endProcess method would just suspend it again eliminating the 'cannot return' errors that would otherwise be raised.

isCzech:
Fixes #8844

### DIFF
--- a/src/Kernel/Process.class.st
+++ b/src/Kernel/Process.class.st
@@ -273,7 +273,8 @@ Process >> endProcess [
 	<isTerminated>
 	thisContext terminateTo: nil.   "set thisContext sender to nil"
 	self suspend.
-	^thisContext restart   "restart and suspend again in case of sending resume to a terminated process"
+	"the following line commented out because current debugger step over tests fail with this feature"
+	"^thisContext restart"   "restart and suspend again in case of sending resume to a terminated process"
 
 ]
 
--- a/src/Kernel/Process.class.st
+++ b/src/Kernel/Process.class.st
@@ -240,12 +240,10 @@ Process >> doTerminationFromAnotherProcess [
 		[ UnwindError signalIn: stackCopy ] forkNamed: 'Unwind error during termination'.
 		ctxt terminateTo: ctxt sender.
 		ctxt := self popTo: suspendedContext bottomContext ].
-	"The suspendedContext of process that has never run will have 
-	a pc equal to the startpc. #isTerminate would thus answer false
-	even though the process is dead. To prevent that we set the pc
-	of the suspendedContext to nil (#isDead ---> true)."
+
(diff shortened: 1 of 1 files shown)


MarcusDenker:
Somehow it looks as if the tests time out on all platforms

isCzech:
> Somehow it looks as if the tests time out on all platforms

Thanks, it looks like some debugger tests don't like the feature preventing 'cannot return' errors (i.e. simply suspending accidentally resumed terminated processes, without raising an error). So I just commented the feature out (leaving the other code intact) to make it available for future consideration. Sorry for missing this :)

dionisiydk:
Hi @isCzech .

Sorry that I was not clear. But I expected that you will push your changes on top of mine.

Now it will cause a conflict with my PR (#8567 ). And considering the size of my changes I would be very happy to avoid remerging. It will require another pass on testing and rethinking of what I get. 
Not sure that I will have the time for this soon. But the PR fixes very critical problems. it is better to integrate it first and then do other improvements. 

So please wait the integration of #8567

isCzech:
> Hi @isCzech .
> 
> Sorry that I was not clear. But I expected that you will push your changes on top of mine.
> 
> So please wait the integration of #8567

Hi @dionisiydk ,
sure! I hope the proposed changes will still be compatible :)

Ducasse:
Hi denis

I heard today Pablo telling that he had a look and like it but wants guille to have a look too so that they can make sure that it is ok. 

S.

isCzech:
Hi All,
I've opened an issue #8992 (Unwind mechanism during termination is broken - skips multiple unwind blocks and corrupts the image) which addresses a related but more important problem. Adjusting #isTerminated suggested here is of a way lower priority.

estebanlm:
same as with  #8993, we want this, but is very dangerous so I am deferring.

Ducasse:
Yes this is good to relabel it.

guillep:

estebanlm:
Ok for me :)

guillep:

Ducasse:
Agreed because after they become baby monsters that grow up and that we cannot handle.

PR #8862, opened 2021-03-23
Add the trace operation to Array2D class

rakki-18:
The operation to find the trace (sum of the diagonal elements of the matrix) of a matrix has been added in this PR.

### DIFF
--- a/src/Collections-Sequenceable/Array2D.class.st
+++ b/src/Collections-Sequenceable/Array2D.class.st
@@ -179,6 +179,17 @@ Array2D >> = aMatrix [
 					and: [ aMatrix privateContents = contents ] ] ]
 ]
 
+{ #category : #'row/column operations' }
+Array2D >> Trace [
+	"returns the trace, the sum of diagonal elements of the matrix"
+	
+	| result |
+	result := 0.
+	[numberOfRows = numberOfColumns] assert.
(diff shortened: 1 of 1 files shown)


StephanEggermont:
Does this implementation have better performance than the more idiomatic one using #inject:into: ?

StephanEggermont (changes requested):
Method name should start with a lower case letter

rakki-18:
@StephanEggermont I have changed the method name now to `tr`. Should I squash the two commits and rebase my PR?

rakki-18:
> Does this implementation have better performance than the more idiomatic one using #inject:into: ?

I am not sure yet.

Ducasse:
Thanks for your contribution. 
Now pay attention that a better implementation of Array2D is available in http://github.com/pharo-containers/Container-Array2d

Then on your code: I think that it is better not to consider Array2D as a Matrix and it is better to enhcance PMMatrix from the Polymath package. We should remove from Array2D all idea of math support.

rakki-18:
@Ducasse Thank you. I will look more into enhancing PMMatrix then. I will also look into the better [implementation](https://github.com/pharo-containers/Containers-Array2D) of Array2D.

Ducasse:
Great!

PR #8867, opened 2021-03-24
[Calypso] Refactor and bugfix on ClySystemEnvironment

sonibla:
Hello there

I refactored and improved the class compilation in ClySystemEnvironment.
Now it is more readable, tested, and some #todo removed.

Issue fixed: when trying to create a method only with a name from the class definition, a "Subscript out of bonds" error was created.

Possible improvements: more robust parsing of class/trait definition (see isTraitDefinition, isClassDefinition methods for example), for example support comments

Improvement of #8773 , but the error already existed before the PR

### DIFF
--- a/src/Calypso-SystemQueries-Tests/ClyExtendedMethodGroupProviderTest.class.st
+++ b/src/Calypso-SystemQueries-Tests/ClyExtendedMethodGroupProviderTest.class.st
@@ -11,5 +11,5 @@ ClyExtendedMethodGroupProviderTest >> classSampleWhichHasGroup [
 
 { #category : #running }
 ClyExtendedMethodGroupProviderTest >> groupProviderClass [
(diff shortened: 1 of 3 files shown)


Ducasse:
Thanks! It was on my list but the logic of this code always made me sick.

dionisiydk:
Hi.
It's improvement of course but all this logic should be replaced by class parser

Ducasse:
We have one CDClassDefinitionParser. It should be made more robust.

Ducasse:
Hi @sonibla 
could you help on the new bug https://github.com/pharo-project/pharo/issues/8878
Because I think that we introduced a regression. We cannot define testcase anymore in fluid class mode and 
we get an error when we want to redefine an existing class?
I do not have cycles to look now.

sonibla:
Ok i'll look

PR #8868, opened 2021-03-24
Add VirtualMachine>>#freeSize

svenvc:
With VirtualMachine>>#memorySize you can find out how much memory the heap/image is consuming.

Right now, there is no way to find out how much free memory is available, although this calculation does exist in VirtualMachine>>statisticsReport.

Make this explicit by adding VirtualMachine>>#freeSize


MarcusDenker (approved):

svenvc:
Thank you

Issue #8878, opened 2021-03-25
Fluid class and more got broken by recent calypso method edition or related


Ducasse:
How to reproduce
- select fluid in the class browser
- define a class  TestCase123
- then change it name TestCase567
```
TestCase << #TestCase123
	slots: { };
	sharedVariables: {  };
	tag: 'Kernel';
	package: 'SUnit-Core'
```

And it crashes with 

```
TestCase  Variable or expression expected -><< #TestCase567
	slots: { };
	sharedVariables: {  };
	tag: 'Kernel';
	package: 'SUnit-Core'
```

In addition:

- Define the class 

```
Object << #MyMyMy
	slots: {};
	sharedVariables: {};
	package: 'Kernel'
```

- Redefine it again, you get prompted and answer yes and boum you get an error

```
compileANewClassFrom: newClassDefinitionString notifying: aController startingFrom: oldClass

	"The receiver's textual content is a request to define a new class. The
	source code is defString. If any errors occur in compilation, notify
	aController."

	| newClassName |
	newClassName := self extractNameFrom: newClassDefinitionString.
	((self
		  isOverridingExistingClassWhenDefiningClassNamed: newClassName
		  startingFrom: oldClass) and: [ 
		 (self confirmToOverrideExistingClassNamed: newClassName) not 
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^


]) 
		ifTrue: [ ^ nil ].

	(self isFluidDefinition: newClassDefinitionString)
		ifTrue: [ 
			ClassDefinitionPrinter showFluidClassDefinition: true.
			^ self
				  defineNewFluidClassFrom: newClassDefinitionString
				  notifying: aController
				  startingFrom: oldClass ]
		ifFalse: [ 
			(self isTraitDefinition: newClassDefinitionString)
				ifTrue: [ 
					^ self
						  defineTrait: newClassDefinitionString
						  notifying: aController
						  startingFrom: oldClass ]
				ifFalse: [ 
					^ self
						  defineNewClassFrom: newClassDefinitionString
						  notifying: aController
						  startingFrom: oldClass ] ]
```

Ducasse:
May be related to recent changes made by @sonibla

Ducasse:
may be this one or related https://github.com/pharo-project/pharo/pull/8773

Ducasse:
[...]

PR #8883, opened 2021-03-26
Fix missing return in ClySystemEnvironment

sonibla:
Hello

This PR fixes an issue mentioned in https://github.com/pharo-project/pharo/issues/8878 with the missing return in ClySystemEnvironment>>confirmToOverrideExistingClassNamed:


sonibla:
Duplicated with https://github.com/pharo-project/pharo/pull/8884
https://github.com/pharo-project/pharo/pull/8884 adds a test case, so I close my PR

PR #8884, opened 2021-03-26
[fix] fixes #8878: Fluid class and more got broken by recent calypso...

### LINKED ISSUE #8878: Fluid class and more got broken by recent calypso method edition or related

hogoww:
Was just a missing return.

### DIFF
--- a/src/BaselineOfCalypso/BaselineOfCalypso.class.st
+++ b/src/BaselineOfCalypso/BaselineOfCalypso.class.st
@@ -146,6 +146,8 @@ BaselineOfCalypso >> baseline: spec [
 
 			package: #'Calypso-SystemTools-Core' with: [
 				spec requires: #(#'Calypso-SystemQueries' #'Calypso-Browser' 'SystemCommands' ). ];
+			package: #'Calypso-SystemTools-Core-Tests' with: [
+				spec requires: #(#'Calypso-SystemTools-Core' ). ];
 			 
 			package: #'Calypso-SystemTools-FullBrowser' with: [
 				spec requires: #(#'Calypso-SystemTools-Core' ). ];
--- a/src/Calypso-SystemQueries-Tests/ClySystemEnvironmentTest.class.st
+++ b/src/Calypso-SystemQueries-Tests/ClySystemEnvironmentTest.class.st
@@ -215,38 +215,3 @@ ClySystemEnvironmentTest >> testIsTraitDefinition [
 	uses: {}
 	package: ''Calypso-SystemQueries-Domain''')
 ]
-
-{ #category : #test }
-ClySystemEnvironmentTest >> testRedefineAClassShouldRedefineIt [
-	| controlerDummy newClassDefinitionString newClass oldUIManager |
-	controlerDummy := RubScrolledTextMorph new.
-	dummySystemEnvironment globals: Smalltalk globals.
-	(self class environment classNamed: #MyClass) ifNotNil: [ self error: 'Need to rename test class to not override existing class' ].
-
-	newClassDefinitionString := 'Object << #MyClass
-    package: ''Calypso-SystemQueries-Tests'''.
-
-
-	oldUIManager := UIManager default.
-	[
-		UIManager default: DummyUIManager new.
-		[ 
-		
-			dummySystemEnvironment compileANewClassFrom: newClassDefinitionString notifying: controlerDummy startingFrom: nil.
-			newClassDefinitionString := 'Object << #MyClass
-		    slots: {#s};
-		    package: ''Calypso-SystemQueries-Tests'''.
-
-			dummySystemEnvironment compileANewClassFrom: newClassDefinitionString notifying: controlerDummy startingFrom: nil.
-
-		
-		] 
-			on: ProvideAnswerNotification 
-			do:[ :e | e resume: true ]
-	]
-	ensure: [ UIManager default: oldUIManager ].
-
-	newClass := self class environment classNamed: #MyClass.
-	[ self assert: newClass allSlots size equals: 1 ] 
-		ensure: [ (self class environment classNamed: #MyClass) removeFromSystem ]
-]
--- a/src/Calypso-SystemQueries-Tests/ClySystemEnvironmentTest.class.st
+++ b/src/Calypso-SystemQueries-Tests/ClySystemEnvironmentTest.class.st
@@ -215,3 +215,38 @@ ClySystemEnvironmentTest >> testIsTraitDefinition [
 	uses: {}
 	package: ''Calypso-SystemQueries-Domain''')
 ]
+
+{ #category : #test }
+ClySystemEnvironmentTest >> testRedefineAClassShouldRedefineIt [
+	| controlerDummy newClassDefinitionString newClass oldUIManager |
+	controlerDummy := RubScrolledTextMorph new.
+	dummySystemEnvironment globals: Smalltalk globals.
+	(self class environment classNamed: #MyClass) ifNotNil: [ self error: 'Need to rename test class to not override existing class' ].
+
+	newClassDefinitionString := 'Object << #MyClass
+    package: ''Calypso-SystemQueries-Tests'''.
+
+
+	oldUIManager := UIManager default.
+	[
+		UIManager default: DummyUIManager new.
+		[ 
+		
+			dummySystemEnvironment compileANewClassFrom: newClassDefinitionString notifying: controlerDummy startingFrom: nil.
+			newClassDefinitionString := 'Object << #MyClass
+		    slots: {#s};
+		    package: ''Calypso-SystemQueries-Tests'''.
+
+			dummySystemEnvironment compileANewClassFrom: newClassDefinitionString notifying: controlerDummy startingFrom: nil.
+
+		
+		] 
+			on: ProvideAnswerNotification 
+			do:[ :e | e resume: true ]
+	]
+	ensure: [ UIManager default: oldUIManager ].
+
+	newClass := self class environment classNamed: #MyClass.
+	[ self assert: newClass allSlots size equals: 1 ] 
+		ensure: [ (self class environment classNamed: #MyClass) removeFromSystem ]
+]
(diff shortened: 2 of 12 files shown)


sonibla:
Duplicated with https://github.com/pharo-project/pharo/pull/8883
Let's keep this one though, because of the test case you added @hogoww  :)

Ducasse:
This does not fix the problem.
This fixes only the one of the two bugs.

Ducasse:
Do you try the first scenario?

sonibla:
For the first scenario, i think it's caused by https://github.com/pharo-project/pharo/pull/7921 @Ducasse 
Or maybe it's because we use a method definition notebook, see @Alisu comment on  https://github.com/pharo-project/pharo/issues/8878

hogoww:
Alright, so, after working on it for a while:
We are using ClyClassDefinitionEditorToolMorph to edit classes when having selected a class. 
Unlike when we have selected a package, in which case we use a ClyClassCreationToolMorph.

In the first case we can edit BOTH methods & classes.
In the second case, we can edit only classes.

The second case works as expected (as far as I can see)
The first case fails when we don't return a class using Fluid.
This happens for example when there is a syntax error.
Therefore it treats it as a method, which is the fallback case.
(see ClyClassDefinitionEditorToolMorph >> applyChanges).

I've thought of several ways to fix this, but I need tests because there's too many cases.
In this last commit there is the *basis* for the tests to be written.
I'm not experienced in writing calypso tests, and I had a lot of trouble.
I added (possibly useless) mocks for that.
I haven't done the Trait ones yet.
They should ease providing a fix that doesn't remove behavior somewhere else.

While trying a random fix, I've also noted reordering of the tabs.
These tests are not for that case.
They are there to test that we can 
    (1) Use successfully every class creation cases
    (2) Get the right syntax error (or other if I discover other things while fixing it

sonibla:
Actually there is something else (the first issue @Ducasse mentioned in https://github.com/pharo-project/pharo/issues/8878)
When you create a new class in fluid using the class definition panel, the class is created but an error message appears on the editor
This is because applyChangesAsClassDefinition returns false even if the class was created, to fix that I thought of fixing the class compiler, that returns aFluidClassBuilder instead of the new class (see discussion in https://github.com/pharo-project/pharo/issues/8878)

Getting the right syntax error is not clear : in the class definition editor panel you can create a class or a method, so if entered string can't be a class or a method, then what error should we show ? (class syntax error or method syntax error)

hogoww:
> This is because applyChangesAsClassDefinition returns false even if the class was created, to fix that I thought of fixing the class compiler, that returns aFluidClassBuilder instead of the new class (see discussion in #8878)

I witnessed that too, that's actually why I've been doing the tests, to fix everything and not risk failing another one =) 

> Getting the right syntax error is not clear : in the class definition editor panel you can create a class or a method, so if entered string can't be a class or a method, then what error should we show ? (class syntax error or method syntax error)

As far as I can see, there is no possible conflict (in simple cases at least).
Trait/Objects using the shift class builder send a message to a global. This cannot be mistaken for a method.
FluidParser also uses a receiver.
The first part of a method is forced to be a selector.

I think the only case that might be a problem is trying to define a method with an uppercase selector, but I'll see about that one later

hogoww:
Btw, I am not sure what a syntax error look like for the shift class builder, everything I tried opens to open the debugger. (might be normal, I do not know either).
I'm open to ideas :)

Ducasse:
Thank you all for this great effort. 
I love the tests :).
I should have written some long time ago.
I loved that I have to rethink my process. 
I'm checking the PR now. 
Thank again. This is cool to have you around!

Ducasse:
I loaded your PR in my image and I do not get it because I still have the following behavior

```
Morph << #AthensDemoMorph
	slots: { #surface . #current . #frame . #pharoLogo . #esugBallon };
	tag: 'Morphic';
	package: 'Athens-Examples'
```

If I add 2 at the end of the class name I get 

```
Morph  Variable or expression expected -><< #AthensDemoMorph2
	slots: { #surface . #current . #frame . #pharoLogo . #esugBallon };
	tag: 'Morphic';
	package: 'Athens-Examples'
```

Ducasse:
Do you get the same in your image?

Alisu:
The fix is still not in here (unless I missed it). The idea was to have tests first to be sure we do not break something else along the way.
For example:
```
defineNewFluidClassFrom: newClassDefinitionString notifying: aController startingFrom: oldClass

	"Precondition: newClassDefinitionString is a fluid class"

	| newClass builder |
	[ 
	builder := (self classCompilerFor: oldClass)
		            source: '(' , newClassDefinitionString , ') install';
		            requestor: aController;
		            logged: true;
		            "for now a super ugly patch"evaluate.
newClass := builder ifNotNil: [builder build] ]
		on: OCUndeclaredVariableWarning
		do: [ :ex | "we are only interested in class definitions"
			ex compilationContext noPattern ifFalse: [ ex pass ].
			"Undeclared Vars should not lead to the standard dialog to define them but instead should not accept"
			self inform: 'Undeclared Variable in Class Definition'.
			^ nil ].

	^ newClass isBehavior
		  ifTrue: [ newClass ]
		  ifFalse: [ nil ]
```
should fix the redefinition problem but the following code:
```
Morph << #AthensDemoMorph
	slots: { #surface . #current . #frame . #pharoLogo  #esugBallon };
	tag: 'Morphic';
	package: 'Athens-Examples'
```
will produce
```
Morph Variable or expression expected -><< #AthensDemoMorph
	slots: { #surface . #current . #frame . #pharoLogo  #esugBallon };
	tag: 'Morphic';
	package: 'Athens-Examples'
```
instead of 
```
Morph << #AthensDemoMorph
	slots: { #surface . #current . #frame . #pharoLogo End of statement expected -> #esugBallon };
	tag: 'Morphic';
	package: 'Athens-Examples'
```
hence producing a new bug.

hogoww:
As theo said :)
You can also see that the tests are still incomplete (didn't do traits), and failing (because it's not fixed yet!)

Problem if it is not a class, it should be treated as a method.
However, if there is a syntax error (for example at least...) it is treated as a method instead of showing that syntax error.

I am going to (try) to fix it now :)
However, I am still missing the syntax error case for the shift class builder cases (traits/regular classes) !
(error case = resulting in syntax error. 
Otherwise it opens a debugger, so behavior is fine !)

hogoww:
Ok, so a few final notes.
I tagged the missing tests (syntax error for regular Objects/Traits) as I do not know if it's possible.
I'm of course willing to remove them if asked to, I just hoped someone might have an idea first! 

I changed the test for the redefinition that should actually work to use another solution.
The initial one was dangerous as it causes a crash in case of the bug.
See previous code at the end of this message.
I'm still not a fan of that solution, but at least, it looks safe. Ish. Well, there's guards to limit the possible damages. I think. :p

Lastly, I like how the tests are looking, but I am not a fan of my namings.
Please feel free to request changes for them !

This PR is done on my part unless there is reviews to be taken into account !



"test"
testRedefineAClassShouldRedefineIt
	|  newClass oldUIManager |
	self checkClassDoesntExists.

	self compileString: 
'Object << #' , className , '
	package: ''Calypso-SystemQueries-Tests'''.
	
	oldUIManager := UIManager default.
	[
		UIManager default: NonInteractiveUIManager new.
		[ 
			self compileString: 
'Object << #' , className , '
	slots: {#s};
	package: ''Calypso-SystemQueries-Tests'''.
		] 
			on: ProvideAnswerNotification 
			do:[ :e | e resume: true ]
	]
	ensure: [ UIManager default: oldUIManager ].

	newClass := self class environment classNamed: #MyClass.
	[ self assert: newClass allSlots size equals: 1 ] 
		ensure: [ self deleteClassIfExists ]

Ducasse:
Thanks pierre I checked on the latest image and all the behavior is fixed.
Thanks again. 
I will let the build built and I will integrate it. 
S.

sonibla:
Thanks for all your efforts @hogoww , this PR introduces a lot of improvements !
I noticed a small issue : applyChangesAsClassDefinition doesn't returns true at the end in case class compilation succedded (this is the official missing return fixing PR xD)

hogoww:
@sonibla Apologies, I don't understand what's wrong ^^"
In my image, the return true at the end is indeed present for that method.

(And yes, so many missing returns :p)

Ducasse:
Borken tests are unrelated

sonibla:
I didn't have the last version, my bad
Everything's perfect !

Issue #8854, opened 2021-03-22
Transcript noise from RubShoutStylerDecorator 


pavel-krivanek:
**Bug description**
When there is a syntax error. the Shout Styler messes the Transcript with messages
`SubscriptOutOfBounds: 0`
on every keystroke, The logging is caused by this method:
`RubShoutStylerDecorator>>#style:`

**To Reproduce**
1. open Transcript
2. open Calypso
3. start to write a completely empty method but without a method selector, start the method text with a number, for example.
4. write something

![pppEVTxijM](https://user-images.githubusercontent.com/572494/111998711-94febe00-8b1c-11eb-81e2-faf2fb9c9927.gif)

olekscode:
I can not reproduce this issue.
I think it was already fixed.

Steps to reproduce:
1. Open Transcript
2. Open System Browser (Calypso)
3. Choose any method, delete everything, start typing
(nothing happens in Transcript)

Alternatively
3. Create a new method, delete everything, start typing
(nothing happens in Transcript)

MarcusDenker:
it needs to start with a number

olekscode:
Ah, indeed!

MarcusDenker:
fixed by merging https://github.com/pharo-project/pharo/pull/9292

PR #8891, opened 2021-03-26
Fixing #8854. Removed Transcript noise from RubShoutStylerDecorator

olekscode:


Ducasse:
I'm not sure that your fix is good because the other was returning the text in case of error and now it returns the receiver or nil

MarcusDenker:
yes, I would like to keep the logging and fix the real cause... 

So one odd thing: the start position of the node (coming from the scanner token originally) seems to be "0" as kind of a "null" value, in the sense of "I have no clue". But then it is used as an offset, which fails as we are 1-based

MarcusDenker:
I close this... we can not remove the transcript output but have to fix the cause of the problem

Issue #8383, opened 2021-01-18
Prevent the infinite opening of debuggers


StevenCostiou:
In certain cases, it happens that the system tries to open debuggers in loop, for instance:
- A loop with a failure in your code, in case it runs in a separate process from the UI's
- A debugger that opens successfully but whose contents triggers the opening of another debugger, and so on

In both cases, it causes the image to hang. The idea is to first prevent that hanging by giving back the hand to the user with information about the problem. A second step would be to give users a means to debug that problem (*i.e.*, a safe debugger opening).

StevenCostiou:
This concerns issues #8146 #8379 #8324 and maybe #2410 #7394 #5878.
These issues are all (I think) about another problem, however the investigation of those problems are blocked by infinite debugger openings.

Issue #8324, opened 2021-01-09
running some tests with errors freezes pharo


Ducasse:
Pharo 9.0.0
Build information: Pharo-9.0.0+build.1029.sha.0eaa422d0d9681484c04883c4c0e1db3c09e5f81 (64 Bit)
![Screenshot 2021-01-09 at 20 20 36](https://user-images.githubusercontent.com/432549/104106804-3addb300-52b8-11eb-8ab6-c355edc8354d.png)

I suspect an endless loops may be due to Oups?

Ducasse:
I'm not sure.

Ducasse:
I spent some time to see if I can reproduce it and I can. 
I load 

```
Metacello new
  baseline: 'ContainersHashTable';
  repository: 'github://pharo-containers/Containers-HashTable/src';
  load.
```

I add TCopy to the CTHashSetTest class

```
TestCase << #CTHashSetTest
	uses: TCloneTest + TAddForUniquenessTest + TCopyTest;
	slots: { #set . #empty . #full . #collectionWith3Elements };
	tag: 'SUnit';
	package: 'Containers-HashTable-Tests'.
```

Then I run the test
I click on the SUnit Runner on the first red test 
This opens a debugger and if I click on the stack -> image frozen.

Ducasse:
![Screenshot 2021-01-09 at 20 40 00](https://user-images.githubusercontent.com/432549/104107211-e12ab800-52ba-11eb-82d7-0a3a9ef8247c.png)

Ducasse:
may be this is related to the bug we found on the fullblock closure.

StevenCostiou:
@Ducasse @MarcusDenker going to settings>debugger>handle debugger errors 
Now, it opens a GT debugger on a Spec debugger error.

I do not know if this is Oups.

What happens is that there is an deprecation warning in FullBlockClosure, that provokes a loop of warning exceptions that should not happen.
I am investigating to confirm, this is just what I see after 1 minutes of investigation.

StevenCostiou:
Ok, I managed to make it work: it seems this is because Warnings (and Deprecations) actually raise standard errors instead of warning errors in Oups. Warnings and Deprecations are normally handled differently as they are not real errors.

I am investigating before making a PR with a fix.

StevenCostiou:
Ok after investigating: it is not fixed yet, it requires more work.

StevenCostiou:
**Current analysis and theory**
[...]

PR #8893, opened 2021-03-26
Rolling back to the old morphic behavior (but in OUPS) when opening debuggers

StevenCostiou:
This fixes in particular https://github.com/pharo-project/pharo/issues/8324 by avoiding locking the image because of infinite debugger openings.

This rollbacks from smart debugger opening (when a debugger fails, tries to open another one) to just open the default system debugger and log error if any.
There actually was a bug in that rollbacked version that is now fixed.

The smart version problems should be investigated (issue https://github.com/pharo-project/pharo/issues/8383).


Ducasse:
Thanks Steven.

Ducasse:
Waiting for the builds

PR #8896, opened 2021-03-26
add isDoubleQuote method

### LINKED ISSUE #5279: could we introduce isDoubleQuote in Character
because this is useful to have.

rakki-18:
Fixes #5279 
Added a new method that checks if the given Character is a double quote or not.

### DIFF
--- a/src/Kernel/Character.class.st
+++ b/src/Kernel/Character.class.st
@@ -798,6 +798,13 @@ Character >> isUppercase [
 ]
 
 { #category : #testing }
+Character >> isDoubleQuote [
+	"Returns whether the receiver is a Double Quote"
+
+	^ self = $"
+]
+
+{ #category : #testing }
 Character >> isVowel [
 	"Answer whether the receiver is one of the vowels, AEIOU, in upper or 
 	lower case."


src/Kernel/Character.class.st
```diff
@@ -797,6 +797,14 @@ Character >> isUppercase [
 
 ]
 
+{ #category : #testing }
+Character >> isDoubleQuote [
+	"Returns whether the receiver is a Double Quote"
+
+	| in |
+	^ (in := self asInteger) == 34
```
guillep: There is no need of a temporary variable here.
That would simplify the code like

```smalltalk
^ (self asInteger) == 34
```

Also remember message precedence: unary methods are always executed before binary methods, and those are before keyword methods. This means that in this case, parenthesis are not required. That would give us

```smalltalk
^ self asInteger == 34
```

Finally, to compare Characters, you don't need to extract their internal code point value. Instead, Characters can be expressed as literal objects. For example, the character `a` is  `$a`, the character `4` is  `$4` (this is kinda the same as single quotes in C. This means you could obtain a literal double quote character by doing `$"`, and with that we could just compare characters!

```smalltalk
^ self == $"
```

Finally, remember in most object oriented languages there are two equality operations: object equality (`=` in Pharo, `.equals()` in Java) and object identity (`==` in both Pharo and Java). Object equality checks if two objects have the same value. Object identity checks if they are **the same object**. In general, we should not use object identity unless really required.
Then I'd simplify further the code to:

```smalltalk
^ self = $"
```

guillep (changes requested):

rakki-18:
@guillep Thank you for the review. I have updated the PR.

PR #8901, opened 2021-03-27

### LINKED ISSUE #2586: Deleting a class with references gives extra "Do you want to delete anyway" msg

gcorriga:

if the class has no references, subclasses, or users, perform a refactoring to remove the class
if the class has references or users, ask the user what to do - browse the references/users, or do nothing.
if the class has subclasses, ask the user what to do - delete keeping subclasses, browse the subclasses, or do nothing.

Fixes #2586.
New version of #8297, #8305.


Ducasse:
I wanted to merge it in my image and I got an improper store (after I review the code looks nice to me).

Ducasse:
```
initializeWithName: nameString
superclassName: superclassString
traitComposition: traitCompositionString
classTraitComposition: classTraitCompositionString
category: categoryString 
instVarNames: ivarArray
classVarNames: cvarArray
poolDictionaryNames: poolArray
classInstVarNames: civarArray
type: typeSymbol
comment: commentString
commentStamp: stampStringOrNil
	name := nameString asSymbol.
	superclassName := superclassString ifNil: ['nil'] ifNotNil: [superclassString asSymbol].
	traitComposition := traitCompositionString.
	classTraitComposition := classTraitCompositionString.
	category := categoryString.
	name = #CompiledMethod ifTrue: [type := #compiledMethod] ifFalse: [type := typeSymbol].
	comment := commentString withInternalLineEndings.
	commentStamp := stampStringOrNil ifNil: [self defaultCommentStamp].
	variables := OrderedCollection  new.
	self addVariables: ivarArray ofType: MCInstanceVariableDefinition.
```

Ducasse:
ivarArray is holding 'encoder' instead of an array!

Ducasse:
I do not understand why it is breaking on my image so I restarted to the build.

Ducasse:
So I'm in a funny situation since I cannot load the code to test the behavior. 
I will integrate it and we will see.

MarcusDenker:
You can download the CI build image instead. This can be faster than loading

PR #8902, opened 2021-03-28
Added RewriteRuleEditor to world menu

StephanEggermont:
But is still missing an icon

### DIFF
--- a/src/NautilusRefactoring/RewriteRuleEditor.class.st
+++ b/src/NautilusRefactoring/RewriteRuleEditor.class.st
@@ -32,6 +32,16 @@ RewriteRuleEditor class >> defaultSpec [
 		yourself	
 ]
 
+{ #category : #'world menu' }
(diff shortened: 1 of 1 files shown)


MarcusDenker:
Even though I think there is an improved version of a rewrite editor somewhere, I vote for adding this for now so we can use what is there

astares:
@MarcusDenker you mean that one: https://github.com/jordanmontt/RewriteToolsSet ?

<blockquote><div> GitHub</div><div><strong><a href="https://github.com/jordanmontt/RewriteToolsSet">jordanmontt/RewriteToolsSet</a></strong></div><div>A set of tools that allows you to do complex searches and code transformations using the refactor engine in Pharo!  - jordanmontt/RewriteToolsSet</div></blockquote>

Ducasse:
Yes that one. 
We will have to remove the menu but you can include if you want. 
After we could also do a PR to include the work of Sebastian in NewTools but we will do it. 
This is not that we have nothing to do.

PR #8903, opened 2021-03-29
RB - Replace method's senders by another

### LINKED ISSUE #2493: Improving refactoring: how to merge on method usage into another
For example imagine that I have 
- method inBibSet: which is called several time and is equivalent to another method bibSet:. 
I would like to the able to say use bibSet: instead of inBibSet:

S.

### LINKED ISSUE #5851: Replace senders refactoring can only replace senders with same amount of arguments
image

It is not clear from the message dialog that we have this restrictrion:

image

So when we add a keyword (and thus, a new argument), it just fails

image

It would be nice to have a smarter dialog. Either adding arguments is forbidden (and clear), or it should be allowed and too to edit the arguments (even sort them!)

Lin777:
Fixes  #5851
Fixes  #2493

You can see the new behavior of refactoring in the following video: https://www.youtube.com/watch?v=zPzUHFWWpV4


Ducasse:
Super nice!!!

PR #8905, opened 2021-03-29
[Redo #8894] Introduce bestNodeForAtPosition: by moving selector extractor heuristic.

### LINKED ISSUE #8887: Rename variable takes wrong selection
```smalltalk
does: e maches: aString
       body of the method
```

- put cursor just after "e"
- Cmd + r
- => proposes rename does:matches: instead of e

guillep:
Introduce bestNodeForAtPosition: by moving selector extractor heuristic.

Fixes #8887

Use `self currentEditedAST` instead of `self editingMethod astForStylingInCalypso`
astForStylingInCalypso yields the AST of the original method, but does not take into account editions

### DIFF
--- a/src/AST-Core-Tests/RBProgramNodeTest.class.st
+++ b/src/AST-Core-Tests/RBProgramNodeTest.class.st
@@ -441,6 +441,84 @@ RBProgramNodeTest >> testBestNodeWithMethodSelectorGivesCommentNode [
 	self assert: ast isCommentNode.
 ]
 
+{ #category : #'testing-adding' }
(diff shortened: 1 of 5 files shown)


Ducasse:
Yes I was looking for currentEditedAST and started to do one before reverting.

MarcusDenker:
Error
Found methods with haltan OrderedCollection(ClyMethodCodeEditorToolMorph>>#toto)
Stacktrace
TestFailure
Found methods with haltan OrderedCollection(ClyMethodCodeEditorToolMorph>>#toto)
ReleaseTest>>testMethodsContainNoHalt

Issue #8525, opened 2021-02-06
Errors : Insufficient buffer size to read current working directory, at the launch of a new image, with 'not supported yet. and ever' StrikeFont error.


labordep:
I create a new Pharo 9 image with the launcher (Pharo 9.0 - 64bit (development version, latest)). 
When I start the freshly created image with the play button I have two errors 'Insufficient buffer size to read current working directory' (See my gif to check the problem).
I'm on Windows 10 Pro x64 version 19042.746.

When I close these errors and I navigate in the browser, I have rapidly a new StrikeFont>>asFreetypeFont error called 'not supported yet. and ever' (See my gif to check the problem).

**Screenshots**
![image](https://user-images.githubusercontent.com/49183340/107118509-d47d8d80-6881-11eb-9cf4-cc556a5620a0.png)
![image](https://user-images.githubusercontent.com/49183340/107118526-e7905d80-6881-11eb-8bc6-45f975450f07.png)
![image](https://user-images.githubusercontent.com/49183340/107118543-f840d380-6881-11eb-84f8-6350737cb0ed.png)

** Video (gif) **
![PharoError](https://user-images.githubusercontent.com/49183340/107118608-3d650580-6882-11eb-9481-67169ab68b8a.gif)

**Version information:**
 - OS: Windows 10 Pro x64 version 19042.746.
 - Version: Pharo-9.0.0+build.1110.sha.cea57d5a282796e03048cfdeaa851c99b8ee6c52 (64 Bit)

MideAkin:
Try reducing the length of your image name.

labordep:
Hello,

Thanks for the reply.

Same problem with an image rename « P.image ».

The patch is not very long : C:\Users\xxxxx\Documents\Pharo\images\P\P.image


Pierre
 

De : MideAkin <notifications@github.com> 


Try reducing the length of your image name.

Ducasse:
Thanks pierre we will come back to you with Pablo.

labordep:
Hi all, a little update about this problem : 

- VMs updated today
![image](https://user-images.githubusercontent.com/49183340/111029457-f330f000-83fc-11eb-9635-7a53875eb174.png)

- Pharo Launcher 2.2 already up to date

- Pharo 9 - 664 bits image created today
![image](https://user-images.githubusercontent.com/49183340/111029486-22476180-83fd-11eb-90bf-8580fa36e16f.png)

[...]

PR #8911, opened 2021-03-29
8525-Errors--Insufficient-buffer-size-to-read-current-working-directory-at-the-launch-of-a-new-image-with-not-supported-yet-and-ever-StrikeFont-error

tesonep:
I cannot reproduce exactly #8525, but I have found issues with the handling of the current working directory and vm directory when executing with a path with utf8 characters.

This is affecting not only the resolution of the working directory, but the resolution of FFI calls as the libraries are next to the VM.``


tesonep:
@labordep I have a possible solution with this

Ducasse:
This looks good to me but I'm a noob in windows

PR #8912, opened 2021-03-29
Add SDL2 class >> version

tinchodias:
The FFI call was already present but it is not direct to obtain
the result from the workspace. Now it only printing 
`SDL2 version`.

Added in class-side in coincidence with other similar methods
in that class.

Also added a printOn: to ease reading.


Ducasse:
thanks martin!

PR #8924, opened 2021-03-30
RB - Improve Extract Method Refactoring

### LINKED ISSUE #8871: Extract method in the superclass is not fully working
Take pillar 8eec3c949d2b1a61ce1f5b3efa547dfed45f7288 and take an use of Smalltalk os lineEnding.

When we do extract method, we are asked if we want to search in the complete hierarchy. 
This should lead to 
 - check and extract that expression in the top most class (the top most class using `Smalltalk os lineEnding.`. This way we get lineEnding in the class higher in the hierarchy
 - then we should check the full hierarchy below this top most class and extract the method to call the extracted method.

Lin777:
Fixes #8871

Now you can specify the class where you want the method to be extracted .

You can see the new behavior in the following video: https://www.youtube.com/watch?v=fCCp9he8Zq0


Ducasse:
The videos is super cool :)!!!

PR #8926, opened 2021-03-31
fixes #8918 the SpLabelledContainer examples are over engineered

### LINKED ISSUE #8918: The SpLabelledContainer examples are over engineered
I might be out on a limb here, but when I look for an example of a `SpLabelledContainer` - in my case I wanted a `SpLabelledTextInput`, the examples on the class side are written to provide the most reuse across the subclasses of `SpLabelledContainer` rather than providing the easiest to understand way to use a widget.

A good example is this:
```smalltalk
SpTextInputFieldPresenter>>example
	<sampleInstance>

	^ self new
		extent: 300 @ 25;
		placeholder: 'This is a ghost';
		text: 'This is a text';
		openWithSpec
```

The example in `SpLabelledTextInput` is like this:
```smalltalk
SpLabelledTextInput>>example
	<sampleInstance>
	^ super example
```
which calls:
```smalltalk
SpLabelledContainer>>example
	<sampleInstance>
	| example |
	example := self abstractExample.
	self setUpExample: example.
	example openWithSpec.
	example withWindowDo: [ :window | window title: self name asString , ' example' ].
	^ example
```
which further forces me to understand:
```smalltalk
SpLabelledContainer>>abstractExample
	| example |
	example := self new.
	example label: 'I am a label'.
	^ example
```
... and ultimately one end up with the right thing on the screen. But not very helpful.

Ultimately I figure out that the right way to do it is:
```smalltalk
SpLabelledTextInput>>kaspersExample
	<example>
	^ (SpLabelledTextInput new
	label: 'simple text field'; 
	text: 'text to be edited';
	whenTextChanged:  [ :newText| Transcript nextPutAll: newText;cr;endEntry  ] )
openWithSpec
``

kasperosterbye:
Added simpler examples to the four concrete subclasses.

### DIFF
--- a/src/Spec2-Core/SpLabelledDropList.class.st
+++ b/src/Spec2-Core/SpLabelledDropList.class.st
@@ -64,20 +64,20 @@ SpLabelledDropList >> selectedItem [
 
 { #category : #'api-shortcuts' }
 SpLabelledDropList >> setIndex: anIndex [
-	^self dropList setIndex: anIndex
+	^self dropList selectIndex: anIndex
 ]
 
 { #category : #'events-shortcuts' }
 SpLabelledDropList >> whenSelectedItemChanged: aBlock [
-	self dropList whenSelectedItemChanged: aBlock
+	self dropList whenSelectedItemChangedDo: aBlock
 ]
 
 { #category : #'events-shortcuts' }
 SpLabelledDropList >> whenSelectionChanged: aBlock [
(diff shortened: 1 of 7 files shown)


Ducasse:
Hi kasper

We discussed with esteban and I do not understand anything ;(
I opened https://github.com/pharo-spec/Spec/issues/1023

S.

Ducasse:
I will close the PR for now. And I like the example BTW

PR #8927, opened 2021-03-31
ReleaseTest: testNoShadowedVariablesInMethods

MarcusDenker:
- add #testNoShadowedVariablesInMethods
- move all tests about Variables to its own category

This test makes sure that there are no methods with shadowing variables other than the 4 examples we use for testing


Ducasse:
I do not get why this test is not working.
osx-64 / Tests-osx-64 / testAs – MacOSX64.Collections.Strings.Tests.Base.ByteSymbolTest

MarcusDenker:
running the tests again, it does not fail anymore

PR #8928, opened 2021-03-31
Speedup-reflective-read-IndexedSlot

MarcusDenker:
Read and Write of indexed instance variables via the variable (IndexedSlot or a subclass) is using the mirror method on thisContext.

(thus we can read ivars from objects without sending a message to them, Variables thus can be understood to be special kinds of Mirror objects)

```Smalltalk
read: anObject
	^ thisContext object: anObject instVarAt: index
```

But this means that we have to reference thisContext... which the VM does not like (it is slow).

The actual primitives do not care about the receiver, we can just copy them and implement them in IndexSlot:

```Smalltalk
read: anObject
	^ self object: anObject instVarAt: index
```

A simple naive benchmark:

```
slot := Point slotNamed: #x.
[slot read: 4@4 ] bench  

before: "'12863984.406 per second'"
after: "'28472842.495 per second'"
```

Which means this is indeed faster, with not changing the complexity too much.

### DIFF
--- a/src/Kernel/IndexedSlot.class.st
+++ b/src/Kernel/IndexedSlot.class.st
@@ -49,11 +49,38 @@ IndexedSlot >> isVirtual [
 ]
 
 { #category : #'meta-object-protocol' }
+IndexedSlot >> object: anObject instVarAt: anIndex [
+	"Primitive. Answer a fixed variable in an object. The numbering of the 
+	 variables corresponds to the named instance variables. Fail if the index 
+	 is not an Integer or is not the index of a fixed variable. Essential for the
+	 debugger. See  Object documentation whatIsAPrimitive."
+
+	<primitive: 73>
+	"instead of copying the fail code here, just call the method that has it"
+	^thisContext object: anObject instVarAt: anIndex
+]
+
+{ #category : #'meta-object-protocol' }
+IndexedSlot >> object: anObject instVarAt: anIndex put: aValue [ 
+	"Primitive. Store a value into a fixed variable in the argument anObject.
+	 The numbering of the variables corresponds to the named instance
+	 variables.  Fail if the index is not an Integer or is not the index of a
+	 fixed variable.  Answer the value stored as the result. Using this
+	 message violates the  principle that each object has sovereign control
+	 over the storing of values into its instance variables. Essential for the
+	 debugger. See Object documentation whatIsAPrimitive."
+
+	<primitive: 74>
+	"instead of copying the fail code here, just call the method that has it"
+	^thisContext object: anObject instVarAt: anIndex put: aValue 
+]
+
+{ #category : #'meta-object-protocol' }
 IndexedSlot >> read: anObject [
-	^ thisContext object: anObject instVarAt: index
+	^ self object: anObject instVarAt: index
 ]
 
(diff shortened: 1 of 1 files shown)


pavel-krivanek (approved):

src/Kernel/IndexedSlot.class.st
```diff
@@ -48,12 +48,39 @@ IndexedSlot >> isVirtual [
...
+	 The numbering of the variables corresponds to the named instance
+	 variables.  Fail if the index is not an Integer or is not the index of a
+	 fixed variable.  Answer the value stored as the result. Using this
+	 message violates the  principle that each object has sovereign control
+	 over the storing of values into its instance variables. Essential for the
+	 debugger. See Object documentation whatIsAPrimitive."
+
+	<primitive: 74>
+	"instead of copying the fail code here, just call the method that has it"
+	^thisContext object: anObject instVarAt: anIndex put: aValue
```
dionisiydk, line 75: Years ago I extracted all mirrors to MirrorPrimitives class. I think it can be used here. 
I wonder if there will be any performance difference comparing to such internal primitive call

MarcusDenker: Yes, I was thinking: why not have a class Mirror that has the method on the class side, but that one can instantiate for an object, too (and provide the API on that instance, too).

This way we can use the class side methods here and not pay the cost to create a mirror instance.

MarcusDenker: Having the mirror methods in a dedicated class is for sure the way to go: I want to add some methods so we can get the slots of an instance without sending messages. (I am experimenting with a "Reflective Object Graph Visitor" that uses the power of being able to annotate instance variables to denote sub-graphs...)

src/Kernel/IndexedSlot.class.st
```diff
@@ -48,12 +48,39 @@ IndexedSlot >> isVirtual [
...
+
+	<primitive: 74>
+	"instead of copying the fail code here, just call the method that has it"
+	^thisContext object: anObject instVarAt: anIndex put: aValue 
+]
+
 { #category : #'meta-object-protocol' }
 IndexedSlot >> read: anObject [
-	^ thisContext object: anObject instVarAt: index
+	^ self object: anObject instVarAt: index
```
dionisiydk, line 80: I mean there is already a method for this:
```Smalltalk
MirrorPrimitives fixedFieldOf: anObject at: anIndex
```

src/Kernel/IndexedSlot.class.st
```diff
@@ -48,12 +48,39 @@ IndexedSlot >> isVirtual [
...
 ]
 
 { #category : #'meta-object-protocol' }
 IndexedSlot >> write: aValue to: anObject [
-	^ thisContext object: anObject instVarAt: index put: aValue
+	^ self object: anObject instVarAt: index put: aValue
```
dionisiydk, line 85: And this:
```Smalltalk
MirrorPrimitives fixedFieldOf: anObject at: anIndex put: aValue
```
It avoids spreading those primitives around the image.

MarcusDenker: Ah, I will check! This should be used here, yes!

PR #8940, opened 2021-04-02
ReClassVariablesShadowingRule

MarcusDenker:
Add a rule that shows when a variable defined by a class is shadowing the outer variable (a global)

### DIFF
--- a/src/GeneralRules/ReClassVariablesShadowingRule.class.st
+++ b/src/GeneralRules/ReClassVariablesShadowingRule.class.st
@@ -1,0 +1,47 @@
+"
+This rule checks if a instance or class variable shadows a global
+"
+Class {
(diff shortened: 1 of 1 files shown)


MarcusDenker:
# Code Critique

What we need next is a Code Critique rule. This for once warns developers early, but in addition it will
make it much easier to fix the 6 problem cases that the release test exposed.

Checking for current senders of #definedVariables leads us to ReVariableAssignedLiteralRule, we can use this as a template.
The main method doing the check looks like this:

```Smalltalk
check: aClass forCritiquesDo: aCriticBlock

	aClass definedVariables do: [ :variable | 
		variable isShadowing  ifTrue: [ 
			aCriticBlock cull: (self critiqueFor: aClass about: variable name) ] ]
```

Issue #8942, opened 2021-04-02
Cleanup some class name in Drtests


Ducasse:
**Describe the problem**
A clear and concise description of what the problem is.
Global state? Bad names? Inconsistent APIs? Missing documentation? Circular dependencies?

**Classes involved**
What are the classes and methods involved in the desired cleanup?

**Proposal**
Do you require a refactoring? The addition of comments?

**Version information:**
 - Version [e.g. Pharo 9.0]
 - Commit [e.g. 22]

**Expected development cost**
As you know, nothing happens magically, so do you have an idea of the development effort required? 
In addition, are you willing to help because after all Pharo is yours too?

**Additional context**

MarcusDenker:
It helps to describe the cleanup in a way that even someone else can do it.

Else, if the idea is to do it now,  instead of opening an issue, one can just directly do a Pull Request. 

Issues are only needed for things that are supposed to be done late *and* possibly by someone else.

Ducasse:
I do not like just to do a PR because sometimes they fail and I need to log stuff. 
Now I do not want to spend more time in describing what can be done during that time.

PR #8947, opened 2021-04-02
fixes: #8942 cleaning class names using the cool tools of nour.

Ducasse:


Ducasse:
there is class that squeezed in

PR #8949, opened 2021-04-02
5388-Improving-RBNode-API: add #hasArguments

### LINKED ISSUE #5388: Improving RBNode API
I waiting for P9. I have some methods to improve the API of the RBNode and subclasses with more tests. 
hasTemporaries,
hasArguments...
It is nicer that forcing people to violate encapsulation.

MarcusDenker:
#hasTemporaries was already added to RBMethodNode

This PR adds a test and #hasArguments to RBMethodNode

fixes #5388

### DIFF
--- a/src/AST-Core-Tests/RBMethodNodeTest.class.st
+++ b/src/AST-Core-Tests/RBMethodNodeTest.class.st
@@ -136,6 +136,14 @@ RBMethodNodeTest >> testCommentsParent [
 		equals: #testCommentsParent.
 ]
 
+{ #category : #tests }
+RBMethodNodeTest >> testHasArguments [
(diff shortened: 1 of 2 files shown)


Ducasse:
may be we should name it hasParameters?

MarcusDenker:
The parameters vs arguments would be a *huge*, *huge* change. Many, many methods, instance variable names,  even class names. It would have an impact everywhere.

We need to plan that well *and* we need to discuss it. This is not something we can do halfway with just one method.

I think it is better to first keep it unified for now and use the term that we use everywhere.

Ducasse:
Ok but this would really improve the API.

PR #8950, opened 2021-04-03
RB - Method's Refactoring using selected scope

### LINKED ISSUE #8638: Remove method (cmd-x) does not respect scoped mode
- I worked on a project 
- Set myself in a scope view here package
- I removed cmd-x printOn: on my class
- I got The message ....

It would be good to enhance and propose the following: 
  - in the scope view you have that many senders
  + globally you have that many senders.

Lin777:
Fixes  #8638

Now the refactorings of the methods are executed in the selected scope, in case the environment is 'current image', it gives you the option to change for another environment, either package, the hierarchy or just the class.

This change is very useful when one has a method with a very common name and we want to rename it, edit it or delete it, since the changes will only be executed in the selected Scope 

_Pending video_


Ducasse:
I was thinking that we should probably do something when the model is not set.

PR #8959, opened 2021-04-05
RB - Clean use of "Smalltalk globals"

### LINKED ISSUE #6510: RBExpandReferencedPoolRefactoring refers directly to Smalltalk globals
```
poolVariableNamesIn: poolName 
	^(Smalltalk globals at: poolName ifAbsent: [Dictionary new]) classPool keys 
		collect: [:name | name asString]
```

Lin777:
Fixes #6510


Ducasse:
Evelyn in fact PoolDictionaries per se do not exist anymore in Pharo. 
Context: in the past a pool dictionary was a dictionary stored in Smalltalk and a class could "import" such poolDictionary and the compiler would use the value when a pooldictionary key was used. 

Nowadays a pool dictionary is represented by a subclass of SharedPool and the previous key and values of the pool dictionary 
are just class variables of the SharedPool subclass. Have a look at from example TextConstant. 

So we should probably integrated this PR. Now from the internals and API of RB one of these days we will have to remove poolDictionary. 
This is why instead of using class definition strings, we should use a classBuilder. Indeed I think that the tests should be able to manipulate classes that are not installed in the system but we should check. 
A possible strategy may be to in a first step: use the classBuilder and invoke the classPrinter and compare the strings.

PR #8960, opened 2021-04-05
RB - Correct selection scope

### LINKED ISSUE #8957: Navigating overridden methods is broken
In calypso there are the little arrows that allow to look for overridden methods.

image

In Pharo 9.0.0
Build information: Pharo-9.0.0+build.1278.sha.325e1924aab0781936bb59e421acee3753a33098 (64 Bit)

Clicking in the arrow does not lead us to the overridden methods, but does something strange. Clicking the arrow up asks something about selecting another scope (?) (don't mind the button colors, I'm playing with that).
image

I don't understand what this is, I never selected a scope before, so this kind of gets in my way.

Lin777:
Fixes #8957

At the moment the scope selection when we are in the global environment fails, so as not to cause problems for the developers I'm removing the part of the selection until I have a version that works well, in this type of case.

You can see the new behavior in the following video: https://www.youtube.com/watch?v=4OMcQ-wvLB4


src/Epicea/EpRBPropagateRefactoring.class.st
```diff
@@ -40,6 +45,11 @@ EpRBPropagateRefactoring >> setOption: aString toUse: aBlockClosure [
 	targetRefactoring setOption: aString toUse: aBlockClosure 
 ]
 
+{ #category : #accessing }
+EpRBPropagateRefactoring >> transform [
```
Ducasse: Hi evelyn 
Are you sure that we need this method?

src/Epicea/EpRBPropagateRefactoring.class.st
```diff
@@ -30,6 +30,11 @@ EpRBPropagateRefactoring >> initializeWith: aRefactoring [
 	targetRefactoring := aRefactoring 
 ]
 
+{ #category : #accessing }
+EpRBPropagateRefactoring >> preconditions [
```
Ducasse: Hi evelyn 
Are you sure that we need this method?

src/Refactoring-Core/RBAbstractClass.class.st
```diff
@@ -37,11 +37,21 @@ Class {
...
+	^ self subclassResponsibility
+]
+
 { #category : #'class initialization' }
 RBAbstractClass class >> initialize [
 	LookupSuperclass := Object new
 ]
 
+{ #category : #accessing }
+RBAbstractClass class >> named: arg1 [
```
Ducasse: Could you renamed arg1 to aString? if this is a string?

src/Refactoring-Core/RBAbstractClass.class.st
```diff
@@ -37,11 +37,21 @@ Class {
 	#category : #'Refactoring-Core-Model'
 }
 
+{ #category : #accessing }
+RBAbstractClass class >> existingNamed: arg1 [
```
Ducasse: Could you renamed arg1 to aString? if this is a string?

src/Epicea/EpRBPropagateRefactoring.class.st
Lin777: I already corrected it, it was an error that I put while testing code critics for Ale Cossio's project

Ducasse:
Unrelated broken tests. Something went wrong on the server.

PR #8969, opened 2021-04-06
7655-Epicea-fileout-should-produce-Pharo-code-and-not-ombu-definition

### LINKED ISSUE #7655: Epicea fileout should produce Pharo code and not ombu definition
When we want to file out code we do not want to get ombu. 
I have no idea how I can reload ombu definitions BTW.

MarcusDenker:
This PR changes the menu entry so that it is clear what is done.
fixes #7655


Ducasse:
Marcus looks like some tests are related to your changes.

Unix64.Files.Tests.Core.FileRegistryTest

but I saw that the test is broken in other issues so probably not linked.

tinchodias (commented):
thanks

PR #8970, opened 2021-04-06
7827-oldDefinition-should-use-the-associated-class-definition-printer

### LINKED ISSUE #7827: oldDefinition should use the associated class definition printer
All the senders of oldDefinition should be revisited and the definition in class description should delegated to the appropriate class Definition printer.

MarcusDenker:
I checked them: RGBehavior was not using the ClassDefinitionPrinter.
This PR  implements a tests an makes it work with "ClassDefinitionPrinter legacy"

fixes #7827


Ducasse:
Nice! Indeed the printers are working nicely.

PR #8979, opened 2021-04-07
LocalTimeZone class variable is shadowing the class of the same name.

pavel-krivanek:
change name to LocalTimeZoneCache


MarcusDenker:
Two other tests are reading the variable reflectively using the old name:

OCBytecodeDecompilerTest>>testLiteralVariableClassVariable
IRBuilderTest>>testLiteralVariableClassVariable

The need to be updated, too

MarcusDenker (approved):

Issue #8984, opened 2021-04-08
ASTCache does not have limit, it grows to have all methods even removed ones


tesonep:
The introduction of the test ReleaseTest >> testNoShadowedVariablesInMethods made evident a problem with the ASTCache. This is keeping all elements that are asked. It is not working as a real cache as it does not have a retention time nor a limit. 

This is producing that all the tests starting to fail with timeout, even making the Windows slaves to run out of time. As having thousands and thousands of AST forces the GC to traverse a Heap much bigger than expected in the test configuration. 

This effect can be clearly seen running twice, in an image before and after the new test:
```
[ 
	ReleaseTest buildSuite run.
	TKTCommonQueueWorkerPoolTest buildSuite run ] timeToRun.
```

In an older image, the time is ~ 15 seconds. 
In a newer image, the time is ~ 53 seconds.

In the newer image the AST cache has 141425 elements, in the old 2761.
The memory used by Pharo is 1.04 GB in the newer one, and 261 MB in the old one.

The test is not the issue, as it is asking the AST of all methods, the problem is the cache, we need to limit it.

tesonep:
I have found it by checking the statistics of test executions. I have checked that the execution time of taskit tests have been duplicated. This is a cool example why the work of @lesueur-philippe, about collecting information from the test execution, could be very interesting and we need to automate some alarms from this data.

MarcusDenker:
as a workaround (and yes, just workaround) we can add a ASTCache reset to that test

guillep:
Would be nice to do a real analysis for the cache. When we are doing heavy code analysis, having the cache in general does not pay off, because we will have to parse everything anyways, unless you force the parsing of the whole set of methods at least twice. And even! in that case, the user probably knows what he/she's doing and the caching could be done more efficiently at the client side.

Probably the cache makes more sense for code navigation, where we are for sure looking at a small enough working set of methods. And in that case maybe a cache bound to, say max 10-20 asts, is enough (and has ~90% hits) => and it would not even require a reset ever!

jecisc:
Currently it is a subclass of WeakIdentityKeyDictionary. Maybe we could make it a LRUCache instead?

privat:
Code could be improved (and a proper data structure used) but the global issue is fixed by #13379 I think

PR #8986, opened 2021-04-08
reset the ASTCache in #testNoShadowedVariablesInMethods

MarcusDenker:
This is a workaround for #8984, that is, it should fix the problems if running the tests on the CI but we need to rethink the mechanism behind the ASTCache


MarcusDenker:
The failing tests in ClassTest #testUsesPoolVarNamed and testHasPoolVarNamed are not caused by this change, but by
https://github.com/pharo-project/pharo/pull/8981

I will open an issue

PR #8993, opened 2021-04-08
#8992 Unwind mechanism during termination is broken, skips multiple unwind blocks and corrupts the image

### LINKED ISSUE #8992: Unwind mechanism during termination is broken - skips multiple unwind blocks and corrupts the image
Try the following example (three nested #ensure:):
```
| p |
p := [
	[
		[ ] ensure: [
			[ ] ensure: [
				Processor activeProcess suspend. 
				Transcript show: 'x1']. 
			Transcript show: 'x2']
	] ensure: [
		Transcript show: 'x3']
] fork.
Processor yield.
p terminate
```
===> x1

The unwind procedure is broken in that it prints just x1 and skips not only x2 but x3 as well ! You'd like to see them all.

I'll try to explain:

To terminate a suspended process we use two different mechanisms:
(1) we try to complete the inner-most unwind block halfway through its execution, and then
(2) we use simulation (!) to unwind the remaining unwind blocks that haven't started their execution yet

This approach introduces three bugs (or at least an undesirable outcome):

(i) Current implementation only attempts to complete the inner-most unwind block halfway through unwind (this part prints 'x1' in the example above). In my opinion it's desirable to attempt to complete the outer-most unwind block under currently evaluation (that would print 'x2').

(ii) What's even more disturbing is the termination unwind skips an outer unwind block that haven't even been started yet when it immediately precedes an inner unwind block (this is where 'x3' was skipped).

(iii) The worst part is #popTo is unable to deal with non-local returns during unwinds and the execution ends up in the wrong stack. This is causing the Unwind errors and the semantics of non-local returns is completely nonsensical as a result. It may be argued non-local returns shouldn't be used in unwind blocks in the first place (because they are 'dangerous') but that doesn't mean we shouldn't fix the procedure :)

If you execute (but first save your image):
```
[ self error ] valueUninterruptably

"and later:"

Semaphore new waitTimeoutMSecs: 50
```
it'll completely ruin your image - why? The termination via #popTo will leave the image in a corrupted state with an effective process still set rendering the image almost unusable. The worst part is the user won't notice anything and a simple code will run as usual - until an error or a crash happens. One of a few ways to see something's off is e.g.
`Processor activeProcess = UIManager default uiProcess` answers false (!)
or inspecting `Processor` to see the effective process is set during 'normal' work (no debugging involved).

My proposal:

Unless there's a serious reason I overlooked miserably, using simulation to complete unwinds is a mistake. It looks like it's not only complicated but also wrong. The easiest and most consistent solution is to simply extend the existing mechanism for completing halfway-through unwind blocks and let it deal with *all* unwind blocks - and that's it. The resulting  unwind (and non-local return) semantics is clean and intuitive, even consistent with e.g. VisualWorks. Check out the code in the PR.

I have to mention there are two tests failing with the new approach:

#testTerminationShouldProceedAllEnsureBlocksIfSomeWasFailed

The test uses the Unwind error as a means to achieve the projected result. Fixing the Unwind errors makes this test indeed fail :) The test's name suggests it tries to show the unwind should continue after an error during unwind but at the moment this logic more or less requires the presence of the unwind error. Other situations involving errors covered in (i) or (ii) don't follow the logic anyway - which is ok and can be adjusted to cover all reasonable cases and not just the unwind errors situations. 

#testIsTerminatingForcedTermination

As I understand it, this test does `process terminate` and should prove it unwinds correctly but it in some very convoluted way generates an exception and ends up terminating the process via #doTerminationFromYourself which in my opinion partly beats the purpose of the test to show the `terminator` is indeed NOT identical to `process`. I suspect it uses the Unwind error again and so it fails when the unwind error bug is fixed. The tested process is simple so no exception should be involved during termination unwind.

If I have misunderstood the purpose of these two tests could the author please correct me? Thanks!
best,
Jaromir

PS: so far tests are green but changing the approach from simulated termination (i.e. as if 'direct') to true 'indirect' termination may affect how some test are designed.

PPS: the role of #isTerminating is unclear to me - what did the author try to achieve exactly? I'd appreciate if someone could enlighten me :) Thanks!

PPPS: I'm aware #isTerminated is horribly wrong and I suggested a quick fix in Issue #2978 here: https://github.com/pharo-project/pharo/issues/2978#issuecomment-787130744. As Denis noted however, he's pushing another fix as part of a big PR #8567, so I'm sending a PR only for #doTerminationFromAnotherProcess.

Build information: Pharo-9.0.0+build.1295.sha.fbc079ba405c0b7289fdc7dc8dbc7f1f4fac3daa (64 Bit)
(Win 10 Pro 20H2)

isCzech:
Fixes #8992

### DIFF
--- a/src/Kernel/Process.class.st
+++ b/src/Kernel/Process.class.st
@@ -240,7 +240,7 @@ Process >> doTerminationFromAnotherProcess [
 		(suspendedContext findContextSuchThat: [:ctx | 
 			ctx closure == outerMost unwindBlock]) ifNotNil: [:inner | 
 				"Let's finish the unfinished unwind context only (i.e. up to inner) and return here"
-				suspendedContext runUntilErrorOrReturnFrom: inner.
+				outerMost := (suspendedContext runUntilErrorOrReturnFrom: inner) first.
 				"Update the receiver's suspendedContext (the previous step reset its sender to nil);
 				return, if the execution stack reached its bottom (e.g. in case of non-local returns)."
 				(suspendedContext := outerMost sender) ifNil: [^self]]]. 
--- a/src/Kernel/Process.class.st
+++ b/src/Kernel/Process.class.st
@@ -217,7 +217,7 @@ Process >> doTerminationFromAnotherProcess [
 	Unwind to execute pending ensure:/ifCurtailed: blocks before terminating.
 	It assumes that self is not the active process
 	and the termination is requested from another process"
-	| ctxt oldList |
+	| ctxt oldList outerMost unwindBlock |
 	"Always suspend the process first so it doesn't accidentally get woken up"
 	oldList := self suspend.
 	suspendedContext ifNil: [^self].
@@ -226,26 +226,44 @@ Process >> doTerminationFromAnotherProcess [
 	in Semaphore>>critical:. So if waiting object is interesting on this situation we will ask it to handle it. In case of Semaphore>>critical, Semaphore will pop the suspendedContext so that we leave the ensure: block inside	Semaphore>>critical: without signaling the semaphore.
 	This methods allow to not be restricted only on Semaphore case."
 	suspendedContext := oldList handleProcessTerminationOfWaitingContext: suspendedContext.
-	"If we are terminating a process halfways through an unwind, try to complete that unwind block first."
-	(suspendedContext findNextUnwindContextUpTo: nil) ifNotNil: [:outer |
-		(suspendedContext findContextSuchThat: [ :c | c closure == outer unwindBlock ]) ifNotNil: [ :inner |
-			"This is an unwind block currently under evaluation"
-			suspendedContext runUntilErrorOrReturnFrom: inner ] ].		
-	ctxt := self popTo: suspendedContext bottomContext.
-	[ ctxt == suspendedContext bottomContext ] whileFalse: [
-		"There was a problem during termination. Make the user aware of the problem
-		but ensure that the current process will be properly terminated."
-		| stackCopy |
-		stackCopy := ctxt copyStack.
-		[ UnwindError signalIn: stackCopy ] forkNamed: 'Unwind error during termination'.
-		ctxt terminateTo: ctxt sender.
-		ctxt := self popTo: suspendedContext bottomContext ].
-	"The suspendedContext of process that has never run will have 
-	a pc equal to the startpc. #isTerminate would thus answer false
-	even though the process is dead. To prevent that we set the pc
-	of the suspendedContext to nil (#isDead ---> true)."
-	(suspendedContext notNil and: [ suspendedContext isBottomContext ]) ifTrue: [
-		suspendedContext pc: nil ]
+
+	"If terminating a process halfways through an unwind, try to complete that unwind block first;
+	if there are multiple such nested unwind blocks, try to complete the outer-most one; the inner
+	blocks will be completed in the process."
+	ctxt := suspendedContext.
+	[(ctxt := ctxt findNextUnwindContextUpTo: nil) isNil] whileFalse: 
+		"Contexts under evaluation have already set their complete (tempAt: 2) to true."
+		[(ctxt tempAt:2) ifNotNil: [outerMost := ctxt]].
+	outerMost ifNotNil: [
+		"This is the outer-most unwind context currently under evaluation;
+		let's find an inner context executing outerMost's argument block (tempAt: 1)"
+		(suspendedContext findContextSuchThat: [:ctx | 
+			ctx closure == (outerMost tempAt: 1)]) ifNotNil: [:inner | 
+				"Let's finish the unfinished unwind context only (i.e. up to inner) and return here"
+				suspendedContext runUntilErrorOrReturnFrom: inner.
+				"Update the receiver's suspendedContext (the previous step reset its sender to nil)"
+				suspendedContext := outerMost]]. 
+
+	"Now all unwind blocks caught halfway through have been completed; 
+	let's execute the ones still pending. Note: #findNextUnwindContextUpTo: starts
+	searching from the receiver's sender but the receiver itself may be an unwind context."
+	ctxt := suspendedContext.
+	ctxt isUnwindContext ifFalse: [ctxt := ctxt findNextUnwindContextUpTo: nil].
+	[ctxt isNil] whileFalse: [
+		(ctxt tempAt: 2) ifNil: [
+			ctxt tempAt: 2 put: true.
+			unwindBlock := ctxt tempAt: 1.
+			"Create a context for the unwind block and execute it on the unwind block's stack. 
+			Note: using #value instead of #runUntilErrorOrReturnFrom: would lead to executing 
+			the unwind on the wrong stack preventing the correct execution of non-local returns."
+			suspendedContext := unwindBlock asContextWithSender: ctxt.
+			suspendedContext runUntilErrorOrReturnFrom: suspendedContext].
+		ctxt := ctxt findNextUnwindContextUpTo: nil].
+
+	"Set the context to its endPC and its sender to nil for the benefit of isTerminated."
+	ctxt := suspendedContext.
+	ctxt terminateTo: nil.
+	ctxt pc: ctxt endPC
 ]
 
 { #category : #private }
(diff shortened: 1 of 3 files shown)


isCzech:
As explained in #8992 the test failures are a result of fixing the termination bugs (unwind errors during termination).

dionisiydk (changes requested):
I will look into the logic later.
But your should really add tests to describe found issues.

src/Kernel/Process.class.st
```diff
@@ -226,26 +226,42 @@ Process >> doTerminationFromAnotherProcess [
...
-	(suspendedContext notNil and: [ suspendedContext isBottomContext ]) ifTrue: [
-		suspendedContext pc: nil ]
+
+	"If terminating a process halfways through an unwind, try to complete that unwind block first;
+	if there are multiple such nested unwind blocks, try to complete the outer-most one; the inner
+	blocks will be completed in the process."
+	ctxt := suspendedContext.
+	[(ctxt := ctxt findNextUnwindContextUpTo: nil) isNil] whileFalse: 
+		"Contexts under evaluation have already set their complete (tempAt: 2) to true."
+		[(ctxt tempAt:2) ifNotNil: [outerMost := ctxt]].
```
dionisiydk:
```suggestion
		[ctxt unwindComplete ifNotNil: [outerMost := ctxt]].
```

src/Kernel/Process.class.st
```diff
@@ -226,26 +226,42 @@ Process >> doTerminationFromAnotherProcess [
...
+	outerMost ifNotNil: [
+		"This is the outer-most unwind context currently under evaluation;
+		let's find an inner context executing outerMost's argument block (tempAt: 1)"
+		(suspendedContext findContextSuchThat: [:ctx | 
+			ctx closure == (outerMost tempAt: 1)]) ifNotNil: [:inner |
```
dionisiydk:
```suggestion
			ctx closure == outerMost unwindBlock]) ifNotNil: [:inner | 
```

dionisiydk:
```suggestion
			ctx closure == outerMost unwindBlock]) ifNotNil: [:inner | 
```

src/Kernel/Process.class.st
```diff
@@ -226,26 +226,42 @@ Process >> doTerminationFromAnotherProcess [
...
+
+	"Now all unwind blocks caught halfway through have been completed; 
+	let's execute the ones still pending. Note: #findNextUnwindContextUpTo: starts
+	searching from the receiver's sender but the receiver itself may be an unwind context."
+	ctxt := suspendedContext.
+	ctxt isUnwindContext ifFalse: [ctxt := ctxt findNextUnwindContextUpTo: nil].
+	[ctxt isNil] whileFalse: [
+		(ctxt tempAt: 2) ifNil: [
+			ctxt tempAt: 2 put: true.
+			unwindBlock := ctxt tempAt: 1.
```
dionisiydk:
```suggestion
			unwindBlock := ctxt unwindBlock.
```

src/Kernel/Process.class.st
```diff
@@ -226,26 +226,42 @@ Process >> doTerminationFromAnotherProcess [
...
+				suspendedContext := outerMost]]. 
+
+	"Now all unwind blocks caught halfway through have been completed; 
+	let's execute the ones still pending. Note: #findNextUnwindContextUpTo: starts
+	searching from the receiver's sender but the receiver itself may be an unwind context."
+	ctxt := suspendedContext.
+	ctxt isUnwindContext ifFalse: [ctxt := ctxt findNextUnwindContextUpTo: nil].
+	[ctxt isNil] whileFalse: [
+		(ctxt tempAt: 2) ifNil: [
+			ctxt tempAt: 2 put: true.
```
dionisiydk:
```suggestion
			ctxt unwindComplete: true.
```

src/Kernel/Process.class.st
```diff
@@ -226,26 +226,42 @@ Process >> doTerminationFromAnotherProcess [
...
+				"Update the receiver's suspendedContext (the previous step reset its sender to nil)"
+				suspendedContext := outerMost]]. 
+
+	"Now all unwind blocks caught halfway through have been completed; 
+	let's execute the ones still pending. Note: #findNextUnwindContextUpTo: starts
+	searching from the receiver's sender but the receiver itself may be an unwind context."
+	ctxt := suspendedContext.
+	ctxt isUnwindContext ifFalse: [ctxt := ctxt findNextUnwindContextUpTo: nil].
+	[ctxt isNil] whileFalse: [
+		(ctxt tempAt: 2) ifNil: [
```
dionisiydk:
```suggestion
		ctxt unwindComplete ifNil: [
```

dionisiydk (changes requested):
only brief look: there are methods on contexts for "unwind variables"

isCzech:
Hi @dionisiydk ,
sorry for not responding earlier, I was not aware of your commenting this. Thanks for suggesting replacing tempAt with accessors - way more readable!

Ducasse:
We are somehow reorganising the engineering team as well as our research team (this happen every 12 years),  under pressure with contracts and release, and flooded by interns (we did an effort because of COVID students do not find internships). I checked the tests and this is super nice that we get all these tests. @estebanlm this is issue may be related to the issue of the unwind on GTK.

estebanlm:
hi, all of this is very important and we definitively want this code... but the nature of this change is so risky that I will defer the inclusion of this code to first iterations of P10 (so we have time to debug if something happens).

isCzech:
Hi @estebanlm , @Ducasse ,

Thank you for your interest! This code is a part of a more complete package overhauling the unwind process during termination. Cuis already integrated the complete code; you can check it out (update the image to latest changes) - including many more tests and examples. Squeak has integrated a portion of it so far. The difference between Pharo and Cuis/Squeak is Pharo refactored #terminate into smaller methods a while ago (and unfortunately including the bugs) but I believe the whole overhauled code is still applicable with reasonable effort.

There's a danger though a future code may try to "learn to live" with the current termination bugs and deficiencies, balance them out on "higher levels" (e.g. debugger) or even "use" them: I've already found two tests that kind of do that and propose to mark them 'expected failure'. They use the "unwind error" to achieve certain outcome but this "unwind error" shouldn't be there at all in the first place :) From this angle I'd say the sooner the better.

The relevant discussions are here:
http://forum.world.st/Solving-multiple-termination-bugs-summary-amp-proposal-td5128285.html#a5130200
https://lists.cuis.st/mailman/archives/cuis-dev/2021-June/003187.html
https://lists.cuis.st/mailman/archives/cuis-dev/2021-May/003171.html

Thanks again, please let me know how I can help.

Ducasse:
Thanks for the info. Our problem is that we do not change the ground in post beta pre release phase. 
The system is like that since at least 10 years. And we have customers. If you have more tests please let include them in the PR.

As a general principle we do not copy code from other systems because we do not want that people think that we are stealing their code. We do not look at their solution for political reasons.

isCzech:
> @Ducasse wrote:
> As a general principle we do not copy code from other systems because we do not want that people think that we are stealing their code. We do not look at their solution for political reasons.

This is my solution and I'm offering it to you. As well as I offered it to Cuis and Squeak. And as a sole author I'm offering to adapt the solution to your particular situation in Pharo 10 - did you mean this is against your policy? I hope not :)

Ducasse:
No this is not against our policy. This is why this is important that you do it. 
Now I just wanted to tell that we (the core Pharo team) do not look at other code because we do not want that people think that we are stealing any clever ideas. In the past some silly people told us that Pharo destroyed Smalltalk, while many others think the inverse :).

isCzech:
>  No this is not against our policy. This is why this is important that you do it.

Thanks, I'm relieved :)

> Now I just wanted to tell that we (the core Pharo team) do not look at other code because we do not want that people think that we are stealing any clever ideas. In the past some silly people told us that Pharo destroyed Smalltalk, while many others think the inverse :).

I can only say without your sharing so many resources I'd probably not be doing any Smalltalk at all :D Thanks again!

Ducasse:
:)
Thanks. May be we will have to revisit the concurrent programming book. But I would like to release Pharo by Example 90. 

Pharo 90 was super difficult with Covid impact and we enhanced so many things that it went far too lengthly. We are brainstorming on how to change our process to get faster. But large changes are just large. We hope to open the Pharo 10 alpha branch by the end of July. And we want to learn how to handle subtle PR more efficiently. 
So I hope Pharo 10 will be in another age :)

guillep:

guillep:

PR #8999, opened 2021-04-08
RB - Add a default refactoring scope

Lin777:
You can see the new behavior in this video: https://www.youtube.com/watch?v=NkxGP0P9klA

### DIFF
--- a/src/SystemCommands-RefactoringSupport/TRefactoringCommandSupport.trait.st
+++ b/src/SystemCommands-RefactoringSupport/TRefactoringCommandSupport.trait.st
@@ -17,5 +17,11 @@ TRefactoringCommandSupport >> setUpModelFromContext: aContext [
 	| browser scope |
 	browser := aContext browser.
 	scope := browser defaultNavigationScope.
+	scope description = 'current image' ifTrue: [ 
+		| scopes choosenScope |
+		scopes := browser allNavigationScopes.
+		choosenScope := scopes select: [ :e | e isKindOf: RefactoringSettings scopeRefactoring ].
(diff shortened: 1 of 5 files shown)


MarcusDenker:
MetacelloNotification: Loaded -> Commander-Spec2-Compatibility-tonel.1 --- tonel:///builds/workspace/uest_and_branch_Pipeline_PR-8999/64/src --- cache
MetacelloNotification: Loading -> SystemCommands-RefactoringSupport-tonel.1 --- tonel:///builds/workspace/uest_and_branch_Pipeline_PR-8999/64/src --- cache
  SycClassCommand
  SycShowChangeRefactoringScopeCommand

src/SystemCommands-RefactoringSupport/TRefactoringCommandSupport.trait.st
```diff
@@ -17,5 +17,11 @@ TRefactoringCommandSupport >> setUpModelFromContext: aContext [
 	| browser scope |
 	browser := aContext browser.
 	scope := browser defaultNavigationScope.
+	scope description = 'current image' ifTrue: [
```
dionisiydk:
```suggestion
	scope isCurrentImage ifTrue: [ 
```
Even if this method not exist it is how you should approach such cases. Code should not rely on string comparisons.

dionisiydk (changes requested):

src/SystemCommands-RefactoringSupport/TRefactoringCommandSupport.trait.st
Lin777: I understand, thank you

PR #9003, opened 2021-04-09
Rename FTMultipleSelection to FTMultipleSelectionStrategy.

NourJihene:
FTMultipleSelection is a strategy which inherits from FTSelectionStrategy


Ducasse:
I like this change too. 
@estebanlm what do you think?
To me it will be more regular.

estebanlm (approved):

PR #9004, opened 2021-04-09
Rename FTSimpleSelection to FTSimpleSelectionStrategy.

NourJihene:
FTSimpleSelection is a strategy which inherits from FTSelectionStrategy.


Ducasse:
@estebanlm I like this change.
What do you think?

estebanlm (approved):

PR #9007, opened 2021-04-09
Rename ExampleRadioButtonModel to RadioButtonModelExample

NourJihene:
Example suffix at the end not the beginning.


Ducasse:
I would love to integrate this one. Now next time consider to group the renaming of a class and its subclasses because else we can have dull conflicts like this one.

PR #9008, opened 2021-04-09
Rename TabSelected to TabSelectedState.

NourJihene:
TabSelected inherits from TabState


Ducasse:
We should start with this one because it is a leave. Else all the other renaming in this hierarchy will fail.

PR #9010, opened 2021-04-09
Rename TabEnabled to TabEnabledState

NourJihene:
TabEnabled inherits from TabState.


Ducasse:
Great. I thought that tabEnabled was an announcement!!!!!

Ducasse:
I thought that I integrated them in the right order but something went wrong. :(

PR #9019, opened 2021-04-09
Rename SimpleBorder to SimpleBorderStyle

NourJihene:
SimpleBorder is a BorderStyle


Ducasse:
Indeed either we have 
Border 
    SimpleBorder

or we have 

BorderStyle
    SimpleBorderStyle

But not a mix of the two.
Now we have 

```
BorderedMorph new borderStyle: (BorderStyle inset width: 2); openInWorld.
```

So I was thinking that I would prefer to have border (and it implies less renaming). 
But this is nice to pass a borderStyle as argument to the message borderStyle:

Ducasse:
We should integrate this one once the subclasses are integrated else all the subclasses PR will break.

PR #9025, opened 2021-04-09
Rename TabPanelBorder to TabPanelBorderStyle.

NourJihene:
TabPanelBorder is a BorderStyle.


Ducasse:
Nour I have the impression that it would be easier to wait for a new build with all the integrated changes and to do the rename of a hierarchy in one shot.

PR #9031, opened 2021-04-09
Rename IconicListItem to IconicListItemMorph.

NourJihene:
IconicListItem is a Morph which inherits  from BorderedMorph.


Ducasse:
I like the regularity.

PR #9039, opened 2021-04-12
Improve "Set up score command"

### LINKED ISSUE #5192: Popup when scoping packages should have the already selected packages
56NQ3eq18Z

Lin777:
Fixes #5192


Ducasse:
I cannot really test but the code looks ok. 
Sorry not to check it before.

PR #9040, opened 2021-04-12
RB - Fix move method with empty protocols

### LINKED ISSUE #8929: Move to package of defining class enh
When we move a method to a class it would be to
- propose the same protocol for the method
- if this is not possible to ask the system for a protocol
when there is a protocol, it would be good to preselect the protocol because this can be super boring.

Lin777:
Fixes #8929

# 1. Move methods to another package
## 1.1. Method that is extension
- Move to another package, there isn't problem because it has the extension protocol by default and the name of the package where it is extended.
- Move to the package where its proprietary class is defined, here there is a problem because in the previous version it was moved with the "as yet unclassified" protocol, **now you have the option to change the protocol so that the moved method is not declassified .**
## 1.2. Method that is not extension,
- Move to another package, the protocol is changed to "extension" and the name of the extension package
# 2. Move methods to another class
## 2.1. The method has protocol
When moving to another class the same protocol is used
## 2.2. The method doesn't have protocol
It was moved to the other class declassified, **now you have the option to change the name of the protocol when moving the method .**

### DIFF
--- a/src/Calypso-SystemTools-Core/ClySystemBrowserContext.class.st
+++ b/src/Calypso-SystemTools-Core/ClySystemBrowserContext.class.st
@@ -217,34 +217,6 @@ ClySystemBrowserContext >> requestSingleClass: queryTitle from: classCollection 
 ]
 
 { #category : #'user requests' }
-ClySystemBrowserContext >> requestSingleMethodTag: queryTitle [
-	^self requestSingleMethodTag: queryTitle suggesting: ''
-]
-
-{ #category : #'user requests' }
-ClySystemBrowserContext >> requestSingleMethodTag: queryTitle suggesting: suggestedTag [
-	| knownTags ui selectedTag |
-	knownTags := (SystemNavigation default allExistingProtocolsFor: true)
-		reject: [ :each | each beginsWith: '*' ].
-	knownTags := knownTags asSortedCollection: [ :a :b | a asLowercase < b asLowercase ].	
-	ui := ListDialogWindow new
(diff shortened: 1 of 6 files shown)


Ducasse:
Evelyn my point was that when we moved an extension to its home class, we should ask the system for the protocol and propose this protocol as the protocol of the method in in its home class. So the point is really what is the default protocol that we offer.

Ducasse:
I'm not sure that we should use this logic. 
I remember that there was a part of the system that could tell us what is the protocol for a given selector. 
I'm looking for it. 

```
knownTags := knownTags asSortedCollection: [ :a :b | a asLowercase < b asLowercase ].	
 	ui := ListDialogWindow new
 		getList: [ :r | knownTags select: [ :e | r search: e ] ];
 		displayBlock: [ :e | e ];
 		initialAnswer: suggestedTag;
 		acceptNewEntry: true;
 		title: queryTitle;
 		yourself.
```

Ducasse:
I found it this is MethodClassifier. 
We should always ask an instance of this class to find the correct protocol for us.

Ducasse:
So all the logic of the handling of protocols should go via this class.
If this is not good then we should improve but the tools should not all do something on their own.

Lin777:
Ok, so I will put as a suggestion in the protocol selection preview the value suggested by MethodClassifier, so the user will take the suggestion or change the protocol.

So I should change the behavior of copying the protocol from the source to the destination (when we move a method from one class to another), as long as the system suggests us?

PR #9043, opened 2021-04-12
RB - Add preview rename class refactoring

### LINKED ISSUE #9032: Rename class refactoring does not show preview
I renamed a class and I could not see the preview of the changes

Lin777:
Fixes  #9032

Now the refactoring has a preview before apply the changes, however the changes are not shown change by change but all in one (see image)

imagen


Ducasse:
Just a question why this is not the same tool that before with the tick? showing a list with all the changes.

Lin777:
As far as I review it is due to how the changes are handled, each type of change has a class that handles its changes, in this case the class that handles the changes is: RBRenameClassChange

PR #9045, opened 2021-04-12
8974-Adding-a-breakpoint-using-the-context-menu-does-not-work-in-the-latest-build

### LINKED ISSUE #8974: Adding a breakpoint using the context menu does not work in the latest build
**Bug description**
When trying to add a Breakpoint from the context menu in the method source code editor, the latest build will not add a breakpoint. No breakpoint is visible on the side of the editor nor in the 'Browse breakpoints'-window. The code will also not break when executing, so I think there is a problem adding the breakpoint.

**To Reproduce**
1. Open the latest Pharo image
2. Open any method
3. Try to add a breakpoint using the context menu (Right click->'Debugging'->'Add breakpoint')
4. See that no breakpoint was added

**Expected behavior**
A breakpoint should be created and visible on the left of the code and in the list of all breakpoints.

**Version information:**
 - OS: Windows 10 64bit
 - Pharo Version: Pharo9.0.0 build: 1289, commit: 2fed4c7, date: 7 April 2021

**Expected development cost**
I do not know. I am currently not sure where the problem is.

**Additional context**
After doing some git bisection, it seems the issue starts with commit 71a5c21719c27beae2ab6a83c744e6c5c1fdc29e.

MarcusDenker:
This PR changes ClyMethodCodeEditorToolMorph/ClyMethodEditorToolMorph to store the ast.

- #buildTextMorph, when the view is created, gets the ast from the unchanged method
- every keystroke (see #textChanged:) parses a new AST from the text buffer
- #findSourceNodeAt: and selectedSourceNode now do not create new ASTs

This makes breakpoints work again for *non edited* methods. This should get us back very close to the old behavior
and makes breakpoints usable for not edited methods.

I think we can even improve it to not clear the breakpoints on accept (if no edit was done between setting the Break and accepting):

- We need to make sure that if we accept a method with a set breakpoint that we compile the method not from the text buffer but from the AST (and make sure the reflective method is correctly setup)
- Not re-initialize the ast in the #applyChanges method

### DIFF
--- a/src/Calypso-SystemTools-Core/ClyMethodCodeEditorToolMorph.class.st
+++ b/src/Calypso-SystemTools-Core/ClyMethodCodeEditorToolMorph.class.st
@@ -41,6 +41,8 @@ ClyMethodCodeEditorToolMorph >> applyChanges [
 	currentMethod := methodClass >> selector.
 	self tagAndPackageEditingMethod: currentMethod.
 	self switchToMethod: currentMethod.
+	"update the AST to the just compiled method. This clears breakpoints. To be improved"
+	ast := self initializeAST.
 	^true
 ]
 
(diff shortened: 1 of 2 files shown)


MarcusDenker:
failing tests not related.

This should be tested a bit before merging

pavel-krivanek (approved):

MarcusDenker:
We tested it a bit more, let's merge

PR #9055, opened 2021-04-13
Allow the user to delete a class even if the class has references.

### LINKED ISSUE #9048: We should be able to remove a class even if it has a reference!!!
Screenshot 2021-04-12 at 22 06 39

gcorriga:
Allow a user to delete a class even if the class has references.

Fixes #9048


Ducasse:
thanks!!!

MarcusDenker (approved):

PR #9061, opened 2021-04-14
7305-HiExampleCommit-classiceberg-v1-5-6-contains-huuuuge-String

### LINKED ISSUE #7305: HiExampleCommit class>>#iceberg_v1_5_6 contains huuuuge String
**Describe the bug**
HiExampleCommit class>>#iceberg_v1_5_6 contains huuuuge String

The FileOut of the method is 678Kb... just imagine: that is quite some part of the image size.

MarcusDenker:
This PR removes the two examples and the ~1MB strings... this is not good to have this in the image.

Even just browing the methods is not working.

This is data, not code, and should be loaded from the network.

fixes #7305


tinchodias (approved):
Fine, from time to time I sync changes in this repo to tinchodias/Hiedra and when I do it, I can refactor it to do not loose the 2 examples as you suggest. I can put the large contest in a file in that repo, and a `ZnEasy get: '...'` obtains it.

PR #9062, opened 2021-04-15
DrTest-Simplify-Docomment-detection

MarcusDenker:
DrTests does all the doc comment detecting, parsing and running all manually, not relying on the PharoDocComment package at all.

This is the first step to change that: we use PharoDocComment abstraction to find the methods. NOTE: the actually execution is still does not use PharoDocComment, this will be another step

- remove not needed example method
- improve comment and categorization in CommentTestCase
- add a #hasDocComment test for method and class
-  use this to simplify DrTest


Ducasse:
great!

PR #9067, opened 2021-04-15
SHTextStyler-Cleanup-Processhandling

MarcusDenker:
SHTextStyler still had all the machinery to style in a spawn process. We do not use this anymore.

- remove unused class GLMHighlighterTextStylerDecorator
- remove ivars background process and monitor in SHTextStyler and all the code that used it

much simpler... next step: hand in the AST from the editor to avoid parsing twice for every keystroke.


guillep (commented):
I personally would prefer not to. What happens with long methods? How much does it take to syntax highlight them? What woulg be nice is to enhance this, not to just remove it... Sometimes complexity is necessary to cover some cases (for example, large images). But we have the choice on how to approach the complexity.

MarcusDenker:
But we did not use it for years...

MarcusDenker:
Another aspect is: this was done because parsing is slow, but parsing has to be done by the Editor anyways, so we just have to visit and add styles.

guillep:
> But we did not use it for years...

That is because you removed the usage. So to me that's not a serious argument.

To me the question is: when we have long methods, is this required?
If it is required, we should reintroduced (potentially improved). If not required, we can remove it.
But I don't see such analysis. To me what I see is that the system degrades on the aspects that are not in your critical path...

MarcusDenker:
I don't think I removed it... at least I do not remember. I think it was part of a bug fix as the mechanism with the thread was not working.

 I just saw all this code that is not used, but we can let it.

PR #9081, opened 2021-04-16
Continue improving DrTests Doc comments plugin

MarcusDenker:
- CommentTestCase move all the code to the instance side
-  pass doc comment to the  CommentTestCase, simplify the state  (e.g selector)

- Simplify #asTestSuite and do not skip Abstract Classes
- improve CommentsToTestsTest


Ducasse:
Great!

PR #9084, opened 2021-04-16
9069-Typing-in-Transcript-gives-a-DNU-doItContext-CompletionEngine

### LINKED ISSUE #9069: Typing in Transcript gives a DNU #doItContext (CompletionEngine)
Pharo 9
- open the Transcript
- type something
- Pharo gives error: "Instance of ThreadSafeTranscript did not understand #doItContext"

image

MarcusDenker:
- fixes #9069 by implementing #doItContext returning nil
- categorize a method


estebanlm:
not convincing of this adding methods strategy, but it does not seems to be another way ;)

PR #9085, opened 2021-04-16
RB - Update deprecate method refactoring and add command

### LINKED ISSUE #8534: RB - Check deprecation refactoring
Check RBDeprecateMethodRefactoring, if all transformations and logic work correctly

Lin777:
Fixes #8534

Current behavior

Script:
```
	(RBDeprecateMethodRefactoring 
		deprecateMethod: #called:on: 
		in: RBRefactoryTestDataApp 
		using: #callFoo) execute
```

Before refactoring:
```
RBRefactoryTestDataApp >> called: anObject on: aBlock 
	Transcript
		show: anObject printString;
		cr.
	aBlock value
```

After refactoring:
```
RBRefactoryTestDataApp >> called: anObject on: aBlock 
	self
		deprecated: 'Use #callFoo instead'
		on: '16 April 2021'
		in: 'Pharo-9.0.0+build.1327.sha.a1d951343f221372d949a21fc1e86d5fc2d2be81 (64 Bit)'.
	Transcript
		show: anObject printString;
		cr.
	aBlock value
```

## Command access

imagen


src/Refactoring-Core/RBDeprecateMethodRefactoring.class.st
```diff
@@ -26,57 +49,48 @@ Class {
...
 ]
 
-{ #category : #transforming }
-RBDeprecateMethodRefactoring >> deprecateMethod: aSelector in: aClass using: newSel [
-	oldSelector := aSelector.
-	newSelector := newSel.
-	class := self classObjectFor: aClass
+{ #category : #'intance creation' }
+RBDeprecateMethodRefactoring class >> model: aRBSmalltalk deprecateMethod: aSelector in: aClass using: newSelector [
+	^(self new)
```
Ducasse: no need for parenthesis

Ducasse:
I will integrate it. After we should think about how we could use deprecatetransformwith: because it is much better.

Ducasse:
For the version it is better to use Pharo90 or Pharo 80 instead of the full hash.

MarcusDenker (approved):

PR #9088, opened 2021-04-18
4443-We-should-have-a-code-critic-rule-checking-for-dot-at-the-end-of-method-signature

### LINKED ISSUE #4443: We should have a code critic rule checking for dot at the end of method signature
Because it is not good style and makes Pharo code non-compatible with some other dialect for no good reason.

MarcusDenker:
This PR adds a new rule ReMethodSignaturePeriodRule

fixes #4443: We should have a code critic rule checking for dot at the end of method signature


astares (approved):

astares:
@MarcusDenker  Can we have a "repair action"?

MarcusDenker:
I think it should be possible (by just recompiling without the dot). Would be a nice next step.

PR #9095, opened 2021-04-19
RB - Fix move method changing protocols

### LINKED ISSUE #8929: Move to package of defining class enh

Lin777:
Fixes #8929

Now when we move a method to a class or other package, you can update the protocol, and this open a preview to select a protocol with a suggestion of MethodClassifier.

You can see the new behavior in this video: https://youtu.be/r22E1BUZxrE


Ducasse:
Hello evelyn 
I have one question: why do you need to have protocol as an instance variable?
This is ok for me but I wanted to understand. 
S

PR #9097, opened 2021-04-19
ReleaseTest-Dot-end-of-Signature

MarcusDenker:
- add release teset #testMethodSignaturePeriod
- fix one problematic case

The other 11 cases are all in Roassal and Iceberg, we need to fix them there.


astares (changes requested):
Would you mind renaming into #testNoPeriodAtEndOfMethodSignature which is more intention revealing

astares (commented):
Looks good. Thanks!

PR #9098, opened 2021-04-19
Fix for #9096 string >>> as time is very permissive

### LINKED ISSUE #9096: String >>> asTime is very permissive
`String >>> asTime` seems to accept any string that beings with a digit. For example the following strings don't throw an error but instead produce erroneous times:

```smalltalk 
'1250 Fake St, New York, USA 12345' asTime.     "2:00 am"
'13000 AM' asTime.                              "4:00 pm"
'25:69:1239' asTime.                            "2:29:39 am"
'1,3,6' asTime.                                 "1:00 am"
```

**Expected behavior**
I would expect all of these examples to produce an error.

**Version information:**
 - OS: MacOS 11.2.1
 - Pharo 9.0.0
Build information: Pharo-9.0.0+build.1335.sha.05eecc36981bdaaae2a32ad8e06c4217eff1dc3f (64 Bit)

**Expected development cost**
I'm willing to develop this and create a pull request

ctSkennerton:
Hi,

This change adds in additional checks to `Time class >>> hour:minute:second:nanoSecond:` and `Time class >>> readFrom:` to make sure that the given string is a valid time. I also extended `TimeTest >>> testReadFromWithError` to include a number of strings that (I think) not should be a time or are otherwise invalid times.

However, one of the existing tests `TimeTest >>> testSeconds` fails with these new changes. In this test, the string `20:33:14.321-05:00` is used and the final part `-05:00` causes the test to error. I believe that this is supposed to be a timezone designation however I don't see any mention of timezones being recognized by the `Time` object (neither the class comment or the `readFrom` comment mention timezone parsing). If this format is required to be valid, I'll change my code so that it does not error for this instance.

### DIFF
--- a/src/Fuel-Tests-Core/FLBasicSerializationTest.class.st
+++ b/src/Fuel-Tests-Core/FLBasicSerializationTest.class.st
@@ -264,20 +264,22 @@ FLBasicSerializationTest >> testDate [
 
 { #category : #tests }
 FLBasicSerializationTest >> testDateAndTime [
-	
+
 	| initialTime initialDate |
 	initialTime := Time fromSeconds: 76020.
 	initialDate := Date fromSeconds: 3492288000.
-	self assertSerializationEqualityOf: (DateAndTime date: initialDate time: initialTime).
-	
-	initialTime := (Time hour: 24 minute: 60 second: 60).
+	self assertSerializationEqualityOf:
+		(DateAndTime date: initialDate time: initialTime).
+
+	initialTime := Time hour: 0 minute: 0 second: 0.
 	initialDate := Date year: 3050 month: 12 day: 31.
-	self assertSerializationEqualityOf: (DateAndTime date: initialDate time: initialTime).
-	
-	initialTime := (Time hour: 24 minute: 60 second: 60).
+	self assertSerializationEqualityOf:
+		(DateAndTime date: initialDate time: initialTime).
+
+	initialTime := Time hour: 0 minute: 0 second: 0.
 	initialDate := Date year: 1600 month: 12 day: 31.
-	self assertSerializationEqualityOf: (DateAndTime date: initialDate time: initialTime).
-	
+	self assertSerializationEqualityOf:
+		(DateAndTime date: initialDate time: initialTime)
 ]
 
 { #category : #tests }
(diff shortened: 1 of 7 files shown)


MarcusDenker:
there seem to be problems leading to a OmFileStoreReadingError: 

```
OmFileStoreReadingError: Reading entry from File @ /builds/workspace/uest_and_branch_Pipeline_PR-9098/64/bootstrap-cache/pharo-local/ombu-sessions/Pharo9.0-PR-64bit-3d7ebab.vsscbmymbdjex09tzy2u1w7c.ombu@131498: Trailing characters after time
[ :error |
				(OmFileStoreReadingError 
					readingError: error
					on: self fileReference
					position: readStream position) signal ] in [ :readStream | [
		^ aBlockClosure value: readStream ]
			on: Error
			do: [ :error |
				(OmFileStoreReadingError 
					readingError: error
					on: self fileReference
					position: readStream position) signal ] ] in OmBlockFileStore(OmFileStore)>>readEntriesWith: in Block: [ :error |...
```

svenvc:
The code changes look good to me and make total sense.

The failure could be related to a wrong time being read that now signals an error. Maybe a time stamp on a method somewhere ?

ctSkennerton:
It's my fault, I didn't understand how `Time` worked with `DateAndTime` class and didn't run all of the tests so I never saw it. In `DateAndTime class >>> readFrom:defaultOffset:` a complete time stamp like `2013-03-04T23:47:52.876+01:00` is parsed piece-wise in different classes. The first part,  `2013-03-04` is handled by `Date class >>> readFrom:`, the stream is then passed to `Time class >>> readFrom:` which parses the `23:47:52.876` part, and the final timezone part, `+01:00` is handled directly in `DateAndTime`. What this means is that trailing characters (that could contain the timezone) must be accepted in `Time class >>> readFrom:`; my change made any string after time an error. 

I believe the most straight forward course of action would be to peek for the timezone in `Time class >> readFrom:` and accept that; throwing an error for other strings.

MarcusDenker:
There are failing tests (I copy those form the Linux slave, but they fail on all platforms):

```
unix-64 / Tests-unix-64 / testDateAndTime – Unix64.Fuel.Tests.Core.Base.FLBasicSerializationTest
<1s
unix-64 / Tests-unix-64 / testTime – Unix64.Fuel.Tests.Core.Base.FLBasicSerializationTest
<1s
unix-64 / Tests-unix-64 / testDateAndTime – Unix64.Fuel.Tests.Core.Base.FLGZippedBasicSerializationTest
<1s
unix-64 / Tests-unix-64 / testTime – Unix64.Fuel.Tests.Core.Base.FLGZippedBasicSerializationTest
<1s
unix-64 / Tests-unix-64 / testDateAndTime – Unix64.Fuel.Tests.Core.Base.FLInMemoryBasicSerializationTest
<1s
unix-64 / Tests-unix-64 / testTime – Unix64.Fuel.Tests.Core.Base.FLInMemoryBasicSerializationTest
<1s
unix-64 / Tests-unix-64 / testChronologyConstants – Unix64.Kernel.Tests.Fluid.FluidClassDefinitionPrinterTest
<1s
unix-64 / Tests-unix-64 / testChronologyConstants – Unix64.Kernel.Tests.Fluid.LegacyClassDefinitionPrinterTest
<1s
unix-64 / Tests-unix-64 / testChronologyConstants – Unix64.Kernel.Tests.Fluid.OldClassDefinitionPrinterTest
```

MarcusDenker:
FluidClassDefinitionPrinterTest>>testChronologyConstants seems to still have a problem

ctSkennerton:
This test passes on my computer so I'm not sure what to do. In Jenkins the order of the `sharedVariables` array in the output is different than what I get on my computer. I think I might be a little out of my depth, any guidance on how I could resolve this?

MarcusDenker:
What I do in this situation is to download the build artifact (the image that the CI builds). This is the second file in this list:

https://ci.inria.fr/pharo-ci-jenkins2/blue/organizations/jenkins/Test%20pending%20pull%20request%20and%20branch%20Pipeline/detail/PR-9098/3/artifacts

This way I can see if the image that the CI was building if I run the tests there if it is working or not.


<blockquote><div><strong><a href="https://ci.inria.fr/pharo-ci-jenkins2/blue/organizations/jenkins/Test%20pending%20pull%20request%20and%20branch%20Pipeline/detail/PR-9098/3/artifacts">Jenkins</a></strong></div></blockquote>

ctSkennerton:
thanks for the tip. `FluidClassDefinitionPrinterTest>>testChronologyConstants` is fixed but there are failures with four tests with a "Test took too much time error" I don't think any of my code should have affected these tests. Can you clarify though whether this is something I should fix?

Ducasse:
I will integrate this changes. Now we are more relax.
Thanks a lot for your contribution.

PR #9100, opened 2021-04-19
6319-SmallInteger-as31BitInteger-wrong-on-64-bit-systems 

### LINKED ISSUE #6319: SmallInteger as31BitInteger wrong on 64-bit systems
as31BitInteger is implemented as `^self`.  On 64-bit systems, the result may be greater than 31 bits as the name implies.

`16r123456789ABCDEF as31BitSmallInt highBit`

MarcusDenker:
as the method is not used (other than in a test), I propose to remove it

fixes #6319

### DIFF
--- a/src/Kernel/Integer.class.st
+++ b/src/Kernel/Integer.class.st
@@ -289,11 +289,6 @@ Integer >> anyMask: mask [
 	^0 ~= (self bitAnd: mask)
 ]
 
-{ #category : #accessing }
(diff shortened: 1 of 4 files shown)


MarcusDenker:
(not sure if removing is the right thing to do... as I have no idea who would need this...)

guillep:
I think removing it is ok. There is no 64 bit equivalent. No real users, not even in the vm.
And to me this has limited usage...

PR #9113, opened 2021-04-20
Refactor semantics analyzer to not rely on #isLocalVariable message

dionisiydk:
1. It refactor the semantics analyzer to not rely on #isLocalVariable message and use a dispatch to the variable instead: #analyzeRead:by: and analyzeWrite:by:
The changes are based on recently introduced visit\~ConcreteTypeOf\~Variable: methods in AST visitors.  
Related to this idea several renames are introduced: #analyseEscapingRead: and #analyseEscapingWrite:  are replaced by #analyzeLocalVariableRead: and #analyzeLocalVariableWrite: accordingly.
The driver for this refactoring is DoItVariable issue (recently fixed #8774) when it represents the external temp during a compilation. Compiler is trying to call some locals specific code during doIt compilation and it was failing because DoItVariable is not a local for newly compiled doIt methods. The Marcus fix introduced missing methods. But this PR allows to avoid it completely. 

2. It optimizes how variables are represented in doIt expressions. The idea is that only variables which can't be used directly in doIts needs to be converted. And it is only local variables. 
Therefore Variable>>#asDoItVariable will returns self as default behaviour. And local variables will be converted to DoItVariable instances.

### DIFF
--- a/src/Kernel/DoItVariable.class.st
+++ b/src/Kernel/DoItVariable.class.st
@@ -68,11 +68,6 @@ DoItVariable >> actualVariable: aVariable [
 	name := actualVariable name
 ]
 
-{ #category : #converting }
-DoItVariable >> asDoItVariableFrom: aContext [
-	^self
-]
-
 { #category : #accessing }
 DoItVariable >> doItContext [
 	^ doItContext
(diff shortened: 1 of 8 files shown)


MarcusDenker:
nice, I will have a look

MarcusDenker (approved):

PR #9130, opened 2021-04-22
fixing-socket-availableData

tesonep:
The primSocketReceiveDataAvailable: should not be called directly, as it should handle the nil in the socketHandle.

This nil is introduced when the image is saved and the socket state is set to nil. 
If we don't handle it properly, the primitive fails with a PrimitveFailure, and it should fail with a ConnectionClosed error.


Ducasse:
Pablo can you add a comment in the  primSocketReceiveDataAvailable: to mention it?
Thanks

PR #9132, opened 2021-04-23
9077-SpValueHolder-did-not-untderstand-asTest-in-new-playground-About-menu

MarcusDenker:
- fix emitStore: in both SpObservableSlot and ObservableSlot
- add test: #testObservableSlotAssignReturnsValue

fixes 9077


MarcusDenker:
This PR needs to be split in two: one for the main image, one for Spec

MarcusDenker:
I removed the spec changes from this PR, I will do that as a PR to the Spec repo

This means that this PR does not fix the issue, though.

MarcusDenker:
Spec is already fixed

PR #9137, opened 2021-04-24
Fix #9136 - inconsistent handling of exception return

### LINKED ISSUE #9136: Inconsistent explicit and implicit exception return
Exception `return` should be handled the same whether it's called explicitly or not: The last line should send return: value to the exception instead of self - the exception return calls handlerContext return and although at the moment it's true the self is the handlerContext, in theory someone can e.g. change `return` definition and then the two returns won't be the same... Plus I think this is more consistent and readable. Thanks,
```
evaluateSignal: exception
	"The following primitive is just a marker used to find the evaluation context. 
	See MethodContext>>#isHandlerOrSignalingContext. "

	<primitive: 199>
	<debuggerCompleteToSender>
	| value |
	exception privHandlerContext: self contextTag.
	value := self exceptionHandlerBlock cull: exception.	
	"default return if not otherwise directed in handle block"
-	self return: value
+	exception return: value
```

isCzech:
Fixes #9136

### DIFF
--- a/src/Kernel/Context.class.st
+++ b/src/Kernel/Context.class.st
@@ -753,8 +753,8 @@ Context >> evaluateSignal: exception [
 	| value |
 	exception privHandlerContext: self contextTag.
 	value := self exceptionHandlerBlock cull: exception.	
-	"return from self if not otherwise directed in handle block"
-	self return: value
+	"default return if not otherwise directed in handle block"
+	exception return: value
 ]
 
 { #category : #'special context access' }


dionisiydk (approved):

isCzech:
Hi,
Apologies, I changed my mind: the two kinds of return as in
```
[] on: Error do: [:ex | ex return]
```
and
```
[] on: Error do: []
```
really are two distinct operations (and probably should remain as such). They are equivalent in Pharo thanks to their current implementation. I'm closing the PR.

dionisiydk:
Hi @isCzech 

I am more convinced with your example that your proposal is correct.
Imagine that #return: is redefined for some exception class. Then the version with "ex return" will works differently than with simple block result. So they are not equivalent with the current implementation

isCzech:
Thanks @dionisiydk for your feedback, I've done some more testing and realized there will be a problem with Exception>> #outer - I'll come up with a test showing it later - so the "fix" probably wouldn't be as simple as I suggested. Besides, as Stephane writes in Deep into Pharo (2013, p. 278), ANSI is not clear regarding the potential difference - or equivalence - between the two. What happened is I'd experimented with a modification of #return and realized the implicit return (or abandon as it's referred to in Deep Into Pharo) will become different from #return :)

isCzech:
Hi @dionisiydk , you were right, my bad - I got confused and the suggested modification works well with #outer. I'm reopening the PR for your consideration :) Sorry for the confusion. Thanks

dionisiydk (approved):

Ducasse:
I suggest that we keep this one for Pharo 10 because we really want to make sure that we do not have changes that can shake to o much the system. Pharo 90 should have been out in April. :(

guillep:

guillep:

PR #9141, opened 2021-04-26
RB - Fix namespace with aliased methods, and inline refactoring errors

### LINKED ISSUE #6218: Inline senders refactoring does not correctly manage errors
**Describe the bug**
When doing an inline senders refactoring of a method that has many senders, the "current image" scope will not show many of the changes (like the removal or the other impacted methods).

**To Reproduce**
1. Go to Object>>printOn:
2. On the method editor, right click on the method signature
3. Select `Inline senders`
4. The preview does not have all the possible senders of `printOn:` (and, I did not test it, but I think it will rename all of them without warning)!

image

**Expected behavior**
The preview should show the entire list of changes

**Version information:**
 - Version: Pharo 9

**Expected development cost**
I think it will not be easy, the Refactorings are too coupled with the UI...

**Additional context**
The problem seems to come from the following method that eats all errors.
In this case, the refactoring seems to throw an error "This method is called by N users, do you want to see them?".

```
removeMethod
	self onError: 
			[self performComponentRefactoring: (RBRemoveMethodRefactoring 
						model: self model
						removeMethods: (Array with: selector)
						from: class)]
		do: []
```

### LINKED ISSUE #9041: RB - RBNamespace doesn't support alias
RBNamespace model doesn't support alias (methods and variables) when we are using Traits

## To Reproduce

    Go to Object>>printOn:
    On the method editor, right click on the method signature
    Select Inline senders
    Show an error, to get the value of method #c3 this is due #c3 is alias for #c2

### LINKED ISSUE #9042: RB - Inline method refactoring doesn't handle well, handling calls to super or overrides
Current version
- does not differentiate between a sender that is called as super or as self
- does not consider those senders that override the method

Lin777:
Fixes  #9041
Fixes #6218
Fixes #9042

You can watch the new behavior in: https://youtu.be/H9S3nmEHL9s

imagen

This refactoring only changes the methods called with SELF and SUPER as receiver, this because if the subclasses override the method, the calls to this method, even from other classes, remain intact.

### DIFF
--- a/src/Refactoring-Core/RBAbstractClass.class.st
+++ b/src/Refactoring-Core/RBAbstractClass.class.st
@@ -429,6 +429,18 @@ RBAbstractClass >> isTrait [
 ]
 
 { #category : #'method accessing' }
+RBAbstractClass >> localSelectors [
+	| selectors |
+	selectors := Set new.
+	selectors addAll: self newMethods keys.
+	self isDefined 
+		ifTrue: 
+			[selectors addAll: self realClass localSelectors.
+			removedMethods ifNotNil: [removedMethods do: [:each | selectors remove: each ifAbsent: []]]].
+	^selectors
+]
+
+{ #category : #'method accessing' }
 RBAbstractClass >> methodFor: aSelector [
 	^ self newMethods
 		at: aSelector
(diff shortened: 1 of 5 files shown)


Ducasse:
Thanks evelyn 
Do you have tests for this?

Lin777:
Hi Stef,

Yes, I added tests for Inline all senders refactoring

Ducasse:
OK I integrated it but what is the class RBInlineMethodRefactoring1 why do we have this class?

Lin777:
Sorry Stef, I uploaded a version prior to the final. I fixed this error in this PR:  https://github.com/pharo-project/pharo/pull/9239

PR #9146, opened 2021-04-26
RB - Add preview of refactorings, improve and add tests to Realize Class Refactoring

Lin777:
- Add previews
- Improve preconditions of Realize Class Refactoring
- Add tests to Realize Class Refactoring

## Realize Class Refactoring

Video: https://youtu.be/UAoiiEK1Xy4

Complete the set of defined methods of this class, by generating a "self shouldBeImplemented" method for all abstract methods defined in its superclass hierarchy. Where an abstract method is a method sending "self subclassResponsibilty.
Shows a warning if this class has abstract methods on its own.


MarcusDenker:
TestFailure
the following methods have unused temporary variables and should be cleaned: an OrderedCollection(SycInlineTempCommand>>#asRefactorings)

Issue #9150, opened 2021-04-26
Right click is broken on variables in the Calypso browser


VincentBlondeau:
**Bug description**
Actions from the contextual menu in the browser is broken:
![image](https://user-images.githubusercontent.com/4706329/116115413-a9ede480-a6ba-11eb-8ddd-2faeb47d56bb.png)

Change from Marcus in 341451ff4d649135c6964f481e7c16eb75ab62f0 was overriden in d4504b29062f04c6741d4f8f2401b61198398156

```
ClyClassDefinitionContext(Object)>>doesNotUnderstand: #lastSelectedMethod
SycExtractSetUpRefactoring>>isApplicable
[ :e | e isApplicable ] in SycCmCommand class>>addItemsToSubmenu:forContext: in Block: [ :e | e isApplicable ]
[ :stream |
		1 to: self size do: [:index |
			each := collectBlock value: (self at: index).
			(selectBlock value: each)
				ifTrue: [ stream nextPut: each ]]] in Array(SequenceableCollection)>>collect:thenSelect: in Block: [ :stream |...
Array class(SequenceableCollection class)>>new:streamContents:
Array(SequenceableCollection)>>collect:thenSelect:
SycSourceCodeCmCommand class(SycCmCommand class)>>addItemsToSubmenu:forContext:
[ 
		submenu := self activationStrategy buildMenuInContext: context.
		submenu isActive ifTrue: [ 
			self
				doRegisterContextMenuItemsFor: aCommandItem
				withBuilder: aBuilder.
			self cmCommandClass addItemsToSubmenu: submenu forContext: context.

			"Registration builder uses #name to bind children to parent"
			submenu rootGroup name: aCommandItem name.
			submenu registerContextMenuItemsWithBuilder: aBuilder ] ] in SycOpenSourceCodeMenuCommand(SycOpenContextMenuCommand)>>registerContextMenuItemsFor:withBuilder: in Block: [ ...
FullBlockClosure(BlockClosure)>>ensure:
RecursionStopper>>stopMethod:during:
RecursionStopper class>>during:
SycOpenSourceCodeMenuCommand(SycOpenContextMenuCommand)>>registerContextMenuItemsFor:withBuilder:
CmdCommandActivator>>registerContextMenuItemsFor:withBuilder:
CmdCommandMenuItem>>registerContextMenuItemsWithBuilder:
[ :each | 
[...]

PR #9151, opened 2021-04-26
Fix 9150-Right-click-is-broken-on-variables-in-the-Calypso-browser 

VincentBlondeau:


Ducasse:
Thanks vincent. 
You fix is strange :)
I would not have expected it like that at all. 
To me this looks strange that something related to extract method is called on classvariables.

Ducasse:
But I can reproduce it also on instance variables :(.

Ducasse:
I confirm that your fix is correct. It fixes also instance var.

VincentBlondeau:
The fix is just a revert of the old version like described in #9150

Ducasse:
@Evelyn may be this will impact the extract Setup.

PR #9176, opened 2021-04-26
Mark abstract classes in Microdown package

### LINKED ISSUE #9175: Mark abstract classes in Microdown package
- a few classes are considered abstract (according to class comment) but return false for #isAbstract    
- so make this more explicit

astares:
Fix #9175

### DIFF
--- a/src/Microdown/MicAbstractAnnotatedBlock.class.st
+++ b/src/Microdown/MicAbstractAnnotatedBlock.class.st
@@ -10,6 +10,12 @@ Class {
 }
 
 { #category : #testing }
+MicAbstractAnnotatedBlock class >> isAbstract [
+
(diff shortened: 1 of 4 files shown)


MarcusDenker (approved):

Ducasse:
I do not like this pattern. 
We do not need it.

Ducasse:
Can we wait a bit before integrating changes?
I do not want to have all our classes cripple with such stupid code.

Ducasse:
I will revert because I will not integrate this code in Microdown and Pillar because we real discussion. 
Sorry to look harsh but soon someone will come with a rule and we will do super dull work. 
I prefer to spend my time fix the parser of class definition or bugs in microdow.

astares:
???

PR #9185, opened 2021-04-27
8897 restart quick method smashes stack

### LINKED ISSUE #8888: Stepping Over a method returning self hangs the image
**Bug description**
When doing a step over on a message send that is returning self (it is a quick primitive it hangs). Maybe it is the same for other quick primitives.

**To Reproduce**
1. Debug any test.
2. Restart in the method #performTest.
3. Click Over many times
4. See the explotion.

**Expected behavior**
Do the step over.

**Version information:**
 - OS: ANY
 - Pharo Version: 9 Latest

### LINKED ISSUE #8897: Restart Quick method smashes stack
**To Reproduce**
1. Write in a playground `#symbol asSymbol`
2. Debug it using  `Cmd shift d`
3. Step into `asSymbol`
4. Restart

Now the stack is empty

StevenCostiou:
Fixes #8897
Fixes #8888


StevenCostiou:
Unrelated test failure.
Can someone review?

src/Debugging-Core/Context.extension.st
```diff
@@ -304,16 +309,23 @@ Context >> stepToHome: aContext [
...
+
 	"Simulate the execution of bytecodes until either sending a message or 
 	returning a value to the receiver (that is, until switching contexts)."
 
 	| context |
-	[ self isDead or: [ self willSend or: [  self willReturn or: [ self willStore or: [self willCreateBlock ] ] ] ] ]
-		whileFalse: [
+	self method isQuick ifTrue: [ 
+		self method primitive = self returnSelfPrimitive ifFalse: [ 
+			pc := pc + self quickMethodPrimitiveBytecodeSize.
```
guillep: I wonder if this works for other quick primitives too, like return false, true, accessors... Also I don't remember if very used numbers (0, 1, ...) are compiled as quick too or not.

StevenCostiou: No it doesn't, it is more complicated than anticipated.
Requires more work.

StevenCostiou: There are two cases if I understand it right:
- quick prims that push something on the stack then return it (like an instvar or a number), these methods executes two bytecodes: the push and the return
- quick prims with optimized push+return, that only execute one bytecode: `return:true` or `returnSelf`

For the first case, when we restart a quick method context we need to put the pc at the first bytecode so we have to skip the primitive: `pc := pc + quickPrimitiveSize`.

For the second case, when we restart a quick method context, we just need to go at the end of the method: `pc := pc + self method endPC`

StevenCostiou: I tried to make it simple, and fixed directly related problems in the way.
It can be reviewed already.

MarcusDenker:
there is one test failing that looks related:

```
Error
Denial failed
Stacktrace
TestFailure
Denial failed
ContextTest(TestAsserter)>>assert:description:resumable:
ContextTest(TestAsserter)>>deny:description:resumable:
ContextTest(TestAsserter)>>deny:description:
ContextTest(TestAsserter)>>deny:
ContextTest>>testStepIntoQuickMethodInCompiledBlock
```

StevenCostiou:
It looks better, but there is a new failing test I do not know if it has shown in other PR or if this is a side effect (I do not see the relation and the test passes in my image).

MarcusDenker:
I don't think it is related to this PR

Ducasse:
I want to learn more on the VM and context to be able to carefully review such kind of changes.

Ducasse:
So what do we do?

StevenCostiou:
Yes :)
It fixes two annoying bugs, but it touches kernel stuff so... :)

MarcusDenker:
I checked the bugs in both  #8897 and  #8888 with an image that has this PR loaded: works

MarcusDenker:
- I think the #stepIntoQuickMethod: mechanism should be better documented (there should be a comment describing why it is needed). 
- I wonder if there can be any concurrency issues... imagine two debuggers working on two different stacks, but setting the quickstep annotation on the same compiledMethod. Could that lead to  interference?

The #stepIntoQuickMethod: mechanism is not added with this PR, though, thus this is more something that we should look at later, it seems to me.

MarcusDenker (approved):
other than the note about #stepIntoQuickMethod:, it looks good to me

PR #9200, opened 2021-04-28
Add possibility to wrap text in the text comparator

sonibla:
Hello there

I implemented the possibility to wrap texts in SpDiffPresenter (default behavior is unchanged, i.e horizontal scrollbars)

wrapped-example

Example from screenshot:
```
SpDiffPresenter exampleWithOptions presenter beWrapped: true
```


Ducasse:
thanks. I will let esteban handle it.

Ducasse:
@estebanlm

estebanlm:
yeah, I am aware. No time to review yet :)

PR #9202, opened 2021-04-28
Allow-Shadowing-RequestorVars

MarcusDenker:
This is a fix for https://github.com/pharo-spec/NewTools/issues/221

- introduce the concept of #allowShadowing for Variables. These variables can be shadowed and will not raise and error
- OCASTSemanticAnalyzer>>#declareVariableNode:as: now checks for allowShadowing

Add a test testDoItRequestorShadow that tests that we can indeed declare a temp if we evaluate aganst a requestor.

### DIFF
--- a/src/Kernel/Variable.class.st
+++ b/src/Kernel/Variable.class.st
@@ -72,6 +72,11 @@ Variable >> acceptVisitor: aProgramNodeVisitor node: aNode [
 	^ aProgramNodeVisitor visitVariableNode: aNode
 ]
 
+{ #category : #testing }
+Variable >> allowShadowing [
+	^false
+]
+
 { #category : #converting }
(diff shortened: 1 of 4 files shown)


src/Kernel/Variable.class.st
```diff
@@ -72,6 +72,11 @@ Variable >> acceptVisitor: aProgramNodeVisitor node: aNode [
 	^ aProgramNodeVisitor visitVariableNode: aNode
 ]
 
+{ #category : #testing }
+Variable >> allowShadowing [
```
dionisiydk: it should be allow**s**~ with **s** at the end.

MarcusDenker: fixed, thanks!

dionisiydk (approved):

Issue #9208, opened 2021-04-29
RBSequenceNode periods, statements, temporaries point to the same instance


enwiro:
**Bug description**
We try to import from sourcecode to FAST pharo. But the RBSequenceNode raised a DNU on isReturn.

**To Reproduce**
1. Import FASTPharo 
2. ```Smalltalk FASTSmalltalkImporterVisitor new runWithSource: (Object >> #at:) sourceCode ``` execute it

**Expected behavior**
The error come from RBSequenceNode periods, statements, temporaries point to the same instance.

**Version information:**
 - OS: Linux
 - Pharo Version 9

**Expected development cost**
PR in progress

enwiro:
Working on it

MarcusDenker:
merged

PR #9209, opened 2021-04-29
9208 rb sequence node periods statements temporaries point to the same instance

enwiro:
RBSequenceNode periods, statements, temporaries point to the differents instances of OrderedCollection.


enwiro:
Fix #9208

MarcusDenker (approved):

Issue #9128, opened 2021-04-22
Fix broken doc comments


MarcusDenker:
**Bug description**
Now that DrTest uses the DocComments infrastructure, we can use it to fix all broken executable comments.
(we should then add a ReleaseTest to never accept broken examples)

There are now 24 Errors and 28 Failures (most of those related to tests that check immutable literals)
837 are green.

![brokenDoccoments](https://user-images.githubusercontent.com/413057/115724093-ec44b800-a380-11eb-95b8-96b1ceebbab2.jpg)

MarcusDenker:
another step done as part of https://github.com/pharo-project/pharo/pull/9254

PR #9218, opened 2021-04-30
Broken-doc-comments-fix-errors 

MarcusDenker:
this PR fixes all doc comments that have errors but one which is in Roassal (I have opend an issue there for that).

One step for https://github.com/pharo-project/pharo/issues/9128
(we need to fix all failing doc comments after this)

### DIFF
--- a/src/ClassParser/Collection.extension.st
+++ b/src/ClassParser/Collection.extension.st
@@ -13,16 +13,14 @@ Collection >> selectEvery: aBlock [
 
 { #category : #'*ClassParser' }
 Collection >> selectEveryFirst [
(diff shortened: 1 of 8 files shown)


Ducasse:
I have two questions
- should we use class name? self does not work?
- I do not get why {1 . 2. 3} is better than #(1 2 3).
-

MarcusDenker:
- #(1 2 3) is read-only, we can not modify it (all literals are read-only, you get an error when executing this in the playground, too)
- It is not working as the doc coments are not compiled in the right scope. This should be fixed, but I wanted to first get all tests green so we can add a release test... after we can improve.

Ducasse:
Ok thanks for the explanation.

PR #9219, opened 2021-04-30
7313 close all debuggers menu option is not working 

### LINKED ISSUE #7313: Close all Debuggers menu option is not working.

StevenCostiou:
Fixes #7313


MarcusDenker (approved):
looks good

PR #9220, opened 2021-04-30
fix 9213 DNU when browsing implementors

### LINKED ISSUE #9213: DNU when browsing implementors with the cursor at the end of the string
**Bug description**
browsing implementors on a random string, with the cursor at the end, in the playground raises a DNU.

**To Reproduce**
- Open a playground
- Type something random
- Put the cursor at the end of this string
-  browse implementors 
- DNU

**Expected behavior**
Browse implementors, as when putting the cursor in the string or highlighting

**Screenshots**
Not that relevant

**Version information:**
Ubuntu 16
Latest Pharo (9)

**Expected development cost**
Not sure, I'll start working on this.

### LINKED ISSUE #9267: Selector resolving and selection in code editor not working
**Bug description**
Before, if you tried to find senders or implementors without explicit selection, Pharo tried to find the last one in the expression (select current line). Now it does not do that.

**To Reproduce**
1. open Playground
2. write `yourself` and stay with the cursor after the `f`character
3. press `Ctrl+N` to browse senders
4. Error: `#asSymbol was sent to nil`

It seems to try to interpret it as a Playground variable.

There is a very strange case when you put the cursor **inside** the given selector. Then it tries to find senders of `noMethod`

hogoww:
fixes #9213.

This breaks (at least) the test SPCodePresenterTest >>  #testSelectedSelector.
I left it as is for now and I'm awaiting for some feedback / opinion on this.
The breaking assert in the test is not relevant IMO in this test.
I'll see if the CI finds more breaking tests and their relevance.

The problem is two fold.
First, when reproducing the issue, the string is parse (rightfully) as a Variable (an unresolved one), which is put in a doIt (#noMethod).
The selector extraction is looking for a selector and only encounters a variable.
Therefore, the first selector it's actually encountering is the doIt selector #noMethod.
There is a test showing it in the selector extraction &  the fix only in the doIt case.
I am unsure whether this can happen some other way.

The second problem is that when putting the caret at the end of a selector, it's too big for the string
'isTest' size = 6
Position of cursor when at the end = 7.
Therefore it was never extracted as a selector.
Fix without test. 
Unsure where this should go if anywhere.

Thanks for reading, awaiting for CI & feedbacks :)

### DIFF
--- a/src/Spec2-Code/SpCodePresenter.class.st
+++ b/src/Spec2-Code/SpCodePresenter.class.st
@@ -197,7 +197,8 @@ SpCodePresenter >> browseSelectedSelectorIfGlobal: globalBlock ifNotGlobal: nonG
 	| variableOrClassName env |
 
 	variableOrClassName := self selectedSelector.
-
+	variableOrClassName ifNil: [ ^ self inform: 'No string could be parsed to answer your query.' ].
+	
 	env := self environment.
 	
 	self flag: #bug.
--- a/src/Tools-CodeNavigation/CNSelectorExtractor.class.st
+++ b/src/Tools-CodeNavigation/CNSelectorExtractor.class.st
@@ -82,7 +82,7 @@ CNSelectorExtractor >> extractSelectorFromNode: aNode [
 			node := node parent messages first ] ].
 	
 	"In the case of doIt methods"
-	node selector = #noMethod & aNode isVariable
+	(node selector = #noMethod and: [ aNode isVariable ])
 		ifTrue: [  ^ aNode name ].
 
 
--- a/src/Tools-CodeNavigation/CNSelectorExtractor.class.st
+++ b/src/Tools-CodeNavigation/CNSelectorExtractor.class.st
@@ -82,7 +82,8 @@ CNSelectorExtractor >> extractSelectorFromNode: aNode [
 			node := node parent messages first ] ].
 	
 	"In the case of doIt methods"
-	node selector = #noMethod ifTrue: [  ^ aNode name ].
+	node selector = #noMethod & aNode isVariable
+		ifTrue: [  ^ aNode name ].
 
 
 	^ node selector
--- a/src/Tools-CodeNavigation/CNSelectorExtractor.class.st
+++ b/src/Tools-CodeNavigation/CNSelectorExtractor.class.st
@@ -80,6 +80,10 @@ CNSelectorExtractor >> extractSelectorFromNode: aNode [
 		(node ~= originalNode and: [ 
 			 node parent notNil and: [ node parent isCascade ] ]) ifTrue: [ 
 			node := node parent messages first ] ].
+	
+	"In the case of doIt methods"
+	node selector = #noMethod ifTrue: [  ^ aNode name ].
+
 
 	^ node selector
 ]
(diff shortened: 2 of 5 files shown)


hogoww:
As far as my manual testing go, this is consistent with expected behavior, I think

MarcusDenker:
SpCodePresenterTest>>testSelectedSelector fails:
```Error
Got #anObjectdo instead of #noMethod.
Stacktrace
TestFailure
Got #anObjectdo instead of #noMethod.
SpCodePresenterTest(TestAsserter)>>assert:description:resumable:
SpCodePresenterTest(TestAsserter)>>assert:description:
SpCodePresenterTest(TestAsserter)>>assert:equals:
SpCodePresenterTest>>testSelectedSelector
SpCodePresenterTest(TestCase)>>performTest
```

hogoww:
> SpCodePresenterTest>>testSelectedSelector fails:
> 
> ```
> Got #anObjectdo instead of #noMethod.
> Stacktrace
> TestFailure
> Got #anObjectdo instead of #noMethod.
> SpCodePresenterTest(TestAsserter)>>assert:description:resumable:
> SpCodePresenterTest(TestAsserter)>>assert:description:
> SpCodePresenterTest(TestAsserter)>>assert:equals:
> SpCodePresenterTest>>testSelectedSelector
> SpCodePresenterTest(TestCase)>>performTest
> ```

See my comment about that in previous post =)

hogoww:
I extracted the logic of the spec test into the Extractor tests.
I left the nil case which is irrelevant for extractor tests.
I added more tests, and am now treating more cases.
Should be ready for integration, if the CI says it's okay :)

I'm of course available for changes if you spot anything !

estebanlm:
so, what´s the next step of this? 
for me the change is good, we can adapt `testSelectedSelector` in consequence.

src/Spec2-Code/SpCodePresenter.class.st
```diff
@@ -197,7 +197,8 @@ SpCodePresenter >> browseSelectedSelectorIfGlobal: globalBlock ifNotGlobal: nonG
 	| variableOrClassName env |
 
 	variableOrClassName := self selectedSelector.
-
+	variableOrClassName ifNil: [ ^ self inform: 'No string could be parsed to answer your query.' ].
```
estebanlm, line 200: in a second thought I do not like this. 
what to show if there are no results (or showing it at all) is a responsibility of the users of this.

estebanlm (changes requested):
In second thoughts, I prefer is you do not call directly the UI from that method. 
If needed, add a new keyword to the message ifNone: or something, so the users can choose what to do (probably the same as now, but in the right place :P)

src/Tools-CodeNavigation/CNSelectorExtractor.class.st
```diff
@@ -80,6 +80,11 @@ CNSelectorExtractor >> extractSelectorFromNode: aNode [
 		(node ~= originalNode and: [ 
 			 node parent notNil and: [ node parent isCascade ] ]) ifTrue: [ 
 			node := node parent messages first ] ].
+	
+	"In the case of doIt methods"
+	node selector = #noMethod & aNode isVariable
```
estebanlm: not 
```Smalltalk
node selector = #noMethod and: [ aNode isVariable ]
```
?

src/Spec2-Code/SpCodePresenter.class.st
hogoww: I think that's a good point.
However, given how the messages are done here, I'm very unsure about putting an "extra case" in the sender.
doBrowseSenders/Implementors just sends a message with two blocks and does not expect a result.

That would look like:
doBrowseSenders

    res := self
        browseSelectedSelectorIfGlobal: [ :global | 
            self systemNavigation browseAllSendersOf: global ]
        ifNotGlobal:  [ :selector |
            self systemNavigation browseAllReferencesTo: selector ].
       res ifNotNil: [  self inform: blablabla ]

I feel that's not better.
Moreover it creates a bit of code duplication for the two senders of this message.
Also, this is kind of a UI method already, isn't it?

estebanlm (commented):
yes I am ok with that :) 
(I need to learn how this thing of changes work, I do not even know where this comment will arise :P)

src/Tools-CodeNavigation/CNSelectorExtractor.class.st
hogoww: Looks like it matters very little in this case.
I can change it if you'd like =)

estebanlm:
yes, please... is about good practices more than immediate consequences :)

hogoww:
Done !

hogoww:
Failing tests look unrelated

estebanlm:
what are we waiting to integrate this?

hogoww:
One "requested change was not tagged as resolved although it was.
Apart from that, I don't know.
I don't intend to change anything else so far :)

MarcusDenker:
I downloaded and tested the build artefact: fixes the problem

MarcusDenker (approved):

PR #9221, opened 2021-04-30
add upperbound on seed in random

### LINKED ISSUE #6231: Random enters invalid state when large seed is given
To reproduce:
`(Random seed: 16rFFFFFFFFFFFFFFFF) next; inspect`

Now its seed is negative and it outputs funky results.

rakki-18:
Aims to solve #6231

When a large seed is given to Random, it starts giving negative values.
Every seed greater than 1e12 seems to be producing negative values.

This PR adds an upper bound( m = 2^31 - 1) to the value that a seed can take.

### DIFF
--- a/src/Random-Core/Random.class.st
+++ b/src/Random-Core/Random.class.st
@@ -184,11 +184,13 @@ Random >> seed [
 
 { #category : #initialization }
 Random >> seed: aNumber [
-	"Refer #privateNextSeed and [1], seed should be positive"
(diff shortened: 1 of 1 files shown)


MarcusDenker (commented):
I had a look at privateNextSeed and to me this looks like the right thing to be done, yes.

MarcusDenker:
I would prefer to get another review before merging

MarcusDenker (approved):

rakki-18:
@MarcusDenker Any update on this PR?

PR #9223, opened 2021-04-30
8892 doing a step over an assert in a test breaks the debugger

### LINKED ISSUE #8892: Doing a Step Over an assert in a test breaks the debugger
- Have a test with a failing assert
- Restart the method
- Do Over until the assert fails and open a new window
- Close the newly open window.
- Restart the original test
- Click on over or where is and nothing happens.

StevenCostiou:
Fixes #8892
Along the way, improvements: less magic, more logic.

### DIFF
--- a/src/Kernel/Context.class.st
+++ b/src/Kernel/Context.class.st
@@ -1766,7 +1766,7 @@ Context >> runUntilErrorOrReturnFrom: aSender [
 	"Insert ensure and exception handler contexts under aSender"
 	error := nil.
 	context := aSender insertSender: (Context
-		contextOn: Error, Halt do: [:ex |
+		contextOn: Error, Halt, TestFailure do: [:ex |
 			error ifNil: [
 				error := ex.
 				topContext := thisContext.
--- a/src/Kernel/Context.class.st
+++ b/src/Kernel/Context.class.st
@@ -778,6 +778,17 @@ Context >> exceptionHandlerBlock [
 
 ]
 
+{ #category : #'private-exceptions' }
+Context >> exceptionsToCaptureWhenStepping [
+
+	| exceptionsToCapture exceptionsSet |
+	exceptionsToCapture := Exception allSubclasses select: [ :c | 
+		                       c captureIfSignalledWhenStepping ].
+	exceptionsSet := ExceptionSet new.
+	exceptionsToCapture do: [ :e | exceptionsSet add: e ].
+	^ exceptionsSet
+]
+
 { #category : #'system simulation' }
 Context >> failPrimitiveWith: maybePrimFailToken [
 	"The receiver is a freshly-created context on a primitive method.  Skip the callPrimitive:
@@ -1766,7 +1777,7 @@ Context >> runUntilErrorOrReturnFrom: aSender [
 	"Insert ensure and exception handler contexts under aSender"
 	error := nil.
 	context := aSender insertSender: (Context
-		contextOn: Error, Halt, TestFailure do: [:ex |
+		contextOn: self exceptionsToCaptureWhenStepping do: [:ex |
 			error ifNil: [
 				error := ex.
 				topContext := thisContext.
--- a/src/NewTools-Debugger-Tests/ContextTest.extension.st
+++ b/src/NewTools-Debugger-Tests/ContextTest.extension.st
@@ -1,0 +1,7 @@
+Extension { #name : #ContextTest }
+
+{ #category : #'*NewTools-Debugger-Tests' }
+ContextTest >> testFailedAssertionContextInSenderChain [
+
+	
+]
--- a/src/NewTools-Debugger/Context.extension.st
+++ b/src/NewTools-Debugger/Context.extension.st
@@ -1,0 +1,20 @@
+Extension { #name : #Context }
+
+{ #category : #'*NewTools-Debugger' }
+Context >> findFailedAssertionContextInSenderChain [
+
+	^ (self findContextSuchThat: [ :ctx | 
+		   (ctx hasTemporaryVariableNamed: #actual) and: [ 
+			   ctx hasTemporaryVariableNamed: #expected ] ]) ifNil: [ 
+		  StDebuggeAssertionFailureContextNotFound new signal ]
+]
+
+{ #category : #'*NewTools-Debugger' }
+Context >> findFailedAssertionResultsInSenderChain [
+	|assertionContext assertionResult|
+	assertionContext := self findFailedAssertionContextInSenderChain.
+	assertionResult := StDebuggerAssertionFailureResult new.
+	assertionResult actual: (assertionContext readVariableNamed: #actual).
+	assertionResult expected: (assertionContext readVariableNamed: #expected).
+	^assertionResult
+]
--- a/src/NewTools-Debugger/StDebuggeAssertionFailureContextNotFound.class.st
+++ b/src/NewTools-Debugger/StDebuggeAssertionFailureContextNotFound.class.st
@@ -1,0 +1,8 @@
+"
+I am raised when a context cannot find in its sender chain an assertion failure context, that is, a context that compared two value, an actual and an expected.
+"
+Class {
+	#name : #StDebuggeAssertionFailureContextNotFound,
+	#superclass : #Exception,
+	#category : #'NewTools-Debugger-Model'
+}
--- a/src/Kernel-Tests/ExceptionTest.class.st
+++ b/src/Kernel-Tests/ExceptionTest.class.st
@@ -52,6 +52,16 @@ ExceptionTest >> runWithNoHandlers: aBlock [
 	^result
 ]
 
+{ #category : #'tests - handling' }
+ExceptionTest >> testCaptureIfSignalledWhenStepping [
+	self deny: Exception captureIfSignalledWhenStepping.
+	self assert: Error captureIfSignalledWhenStepping.
+	self deny: (Error allSubclasses anySatisfy: [ :c| c  captureIfSignalledWhenStepping]).
+	self assert: (UnhandledException withAllSubclasses select:[:c| c captureIfSignalledWhenStepping]) size equals: 1.
+	self assert: (UnhandledException withAllSubclasses select:[:c| c captureIfSignalledWhenStepping]) first identicalTo: Halt.
+	
+]
+
 { #category : #test }
 ExceptionTest >> testDefaultAction [
 	"Ensure that the resuming an UnhandledError when the underlying
--- a/src/Kernel/Error.class.st
+++ b/src/Kernel/Error.class.st
@@ -12,6 +12,11 @@ Class {
 	#category : #'Kernel-Exceptions'
 }
 
+{ #category : #testing }
+Error class >> captureIfSignalledWhenStepping [
+	^self == Error
+]
+
 { #category : #private }
 Error >> isResumable [
 	"Determine whether an exception is resumable."
--- a/src/Kernel/Exception.class.st
+++ b/src/Kernel/Exception.class.st
@@ -50,6 +50,11 @@ Exception class >> - anotherException [
 		yourself
 ]
 
+{ #category : #testing }
+Exception class >> captureIfSignalledWhenStepping [
+	^false
+]
+
 { #category : #exceptionselector }
 Exception class >> handles: exception [
 	"Determine whether an exception handler will accept a signaled exception."
--- a/src/Kernel/Halt.class.st
+++ b/src/Kernel/Halt.class.st
@@ -75,6 +75,11 @@ Class {
 	#category : #'Kernel-Exceptions'
 }
 
+{ #category : #testing }
+Halt class >> captureIfSignalledWhenStepping [
+	^self == Halt
+]
+
 { #category : #halting }
 Halt class >> fromCount: anInteger [
 	"Always halt after a count has been reached"
--- a/src/NewTools-Debugger-Tests/StDebuggerAssertionContextTest.class.st
+++ b/src/NewTools-Debugger-Tests/StDebuggerAssertionContextTest.class.st
@@ -1,0 +1,44 @@
+Class {
+	#name : #StDebuggerAssertionContextTest,
+	#superclass : #TestCase,
+	#instVars : [
+		'session'
+	],
+	#category : #'NewTools-Debugger-Tests-Model'
+}
+
+{ #category : #running }
+StDebuggerAssertionContextTest >> tearDown [
+	session ifNotNil:[session clear].
+	super tearDown 
+]
+
+{ #category : #tests }
+StDebuggerAssertionContextTest >> testFindFailedAssertionContextInSenderChain [
+	| debugger |
+	debugger := StTestDebuggerProvider new debuggerWithFailingAssertionContext.
+	session := debugger session.
+	self assert: session interruptedContext findFailedAssertionContextInSenderChain notNil.
+	session clear.
+	debugger := StTestDebuggerProvider new debuggerWithDNUContext. 
+	session := debugger session.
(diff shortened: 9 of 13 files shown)


StevenCostiou:
Hmm lots of failures...
Looking at why.

StevenCostiou:
Ah, I added a dependency to `TestFailure` in the `Context` class...

StevenCostiou:
All tests are passing now.
Only problem is this error, that I don't know if it is related to my changes or not:
`OmFileStoreReadingError: Reading entry from File @ /builds/workspace/uest_and_branch_Pipeline_PR-9223/Tests-unix-64/pharo-local/ombu-sessions/Pharo9.0-PR-64bit-3b1ecfc.bteyne3iltp9pv405xojhlh7j.ombu@10878978: Illegal leading byte for utf-8 encoding`

Additionally, my changes require careful review :)
I need that review before pushing mirror changes into NewTools.

MarcusDenker (approved):
I tested the scenario described in the bug report, this now works nicer (no extra debugger opening).

The refactorings make things much clearer (no "aContext at: 2") any more, but reads variables by name instead, for example

src/NewTools-Debugger-Tests/ContextTest.extension.st
```diff
@@ -0,0 +1,7 @@
+Extension { #name : #ContextTest }
+
+{ #category : #'*NewTools-Debugger-Tests' }
+ContextTest >> testFailedAssertionContextInSenderChain [
```
dionisiydk, line 4: It looks like a committed extension test. Besides it is empty.

src/NewTools-Debugger/StDebuggeAssertionFailureContextNotFound.class.st
```diff
@@ -0,0 +1,8 @@
+"
+I am raised when a context cannot find in its sender chain an assertion failure context, that is, a context that compared two value, an actual and an expected.
+"
+Class {
+	#name : #StDebuggeAssertionFailureContextNotFound,
+	#superclass : #Exception,
```
dionisiydk, line 6: Why it is not a subclass of Error?
It should be a very good reason to subclass the Exception. And it deserves a very good comment

src/NewTools-Debugger/Context.extension.st
```diff
@@ -0,0 +1,20 @@
+Extension { #name : #Context }
+
+{ #category : #'*NewTools-Debugger' }
+Context >> findFailedAssertionContextInSenderChain [
+
+	^ (self findContextSuchThat: [ :ctx | 
+		   (ctx hasTemporaryVariableNamed: #actual) and: [ 
+			   ctx hasTemporaryVariableNamed: #expected ] ]) ifNil: [
```
dionisiydk, line 8: It is very fragile logic. The names #actual and #expected is kind of SUnit pattern to name variables in the tests. 
For example your logic will not fail for test like:
```Smalltalk
actual := 10 + 20.
expected := 2.
self assert: actual / 0 equals: expected
```
I did not check what debugger will show in that case. But the logic looks wrong.

src/NewTools-Debugger/Context.extension.st
```diff
@@ -0,0 +1,20 @@
...
+]
+
+{ #category : #'*NewTools-Debugger' }
+Context >> findFailedAssertionResultsInSenderChain [
+	|assertionContext assertionResult|
+	assertionContext := self findFailedAssertionContextInSenderChain.
+	assertionResult := StDebuggerAssertionFailureResult new.
+	assertionResult actual: (assertionContext readVariableNamed: #actual).
+	assertionResult expected: (assertionContext readVariableNamed: #expected).
+	^assertionResult
```
dionisiydk, line 19: I would probably implement it a bit differently and not relying on variable names:
```Smalltalk
Context>>isAssertionMethod
    ^"better logic here based on selectors (maybe using pragmas to mark assertions)"

Context>>asAssertionResult
  "Assertions are implemented with convention to have first argument as an actual value and the second argument as expected result"
   self isAssertionMethod ifFalse: [ StDebuggeAssertionFailureContextNotFound signal ].
   ^ StDebuggerAssertionFailureResult actual: self arguments first expected: self arguments second
```
And then the find method will be more simple and readable:
```Smalltalk
Context>> findFailedAssertionResultsInSenderChain
     assertionContext := self findContextSuchThat: [ :ctx | ctx isAssertionMethod ].
     assertionContext ifNil: [ ^  StDebuggeAssertionFailureContextNotFound signal ].
     ^assertionContext asAssertionResult
```

dionisiydk: And I would package those methods into SUnit as it can be quite useful for other projects. 
At the end it can be classes:
- AssertionMethodNotFoundError 
- AssertionMethodResult

and methods:

- #isAssertionMethod 
- #asAssertionMethodResult

dionisiydk: And it would be cool to make it flexible to introduce other kind of result objects for specific assertions.
Particular assertion method can be marked with pragma with specified class of result like: 
```Smalltalk
TestCase>>assert: actual equals: expected
    <assertionResult: #AssertionCompareResult>
   ...
```
And for the collection testing: 
```Smalltalk
TestCase>>assertCollection: actualCollection includesAny: subCollection
   <assertionResult: #CollectionSubsetAssertionResult>
   ...
```
Then the failing inspector StFailingAssertionInspector should not be hardcoded but the debugger will show all inspections of concrete result object (in some preferred order).

In general we will be able to reuse "assertions plugin" of the debugger in other frameworks. Imagine Mock objects assertions, "should expressions".

src/NewTools-Debugger/Context.extension.st
dionisiydk: Another example of why relying on var names is a bad idea:
```Smalltalk
assert: actualNumber closeTo: expectedNumber
```
There will be no assertion tab in the debugger for such assertion failure

src/Kernel/Context.class.st
```diff
@@ -778,6 +778,17 @@ Context >> exceptionHandlerBlock [
 
 ]
 
+{ #category : #'private-exceptions' }
+Context >> exceptionsToCaptureWhenStepping [
+
+	| exceptionsToCapture exceptionsSet |
+	exceptionsToCapture := Exception allSubclasses select: [ :c |
```
dionisiydk, line 785: It can decrease the performance of StepOver button in the debugger (or other related). 
And more important it affects the process termination which performs the unwinds using this simulation machinery. See the method Process>>#doTerminationFromAnotherProcess.

I think you can just implement your own ExceptionSet class:
```Smalltalk
SimulationRunTrapExceptionSet>>handles: anException
   ^anException captureIfSignalledWhenStepping

Context>>runUntilErrorOrReturnFrom: aSender
   ...
   aSender insertSender: (Context
		contextOn: SimulationRunTrapExceptionSet new do: [:ex |
   ...
```
And it all deserves better names

dionisiydk: And this way you will avoid the implementation like "self == Error". Simply return true or false as it really relevant for all subclasses

dionisiydk: Possible alternative to #captureIfSignalledWhenStepping is #shouldInterruptSimulation

src/NewTools-Debugger-Tests/ContextTest.extension.st
StevenCostiou: Oops mistake.. sorry...

src/NewTools-Debugger/StDebuggeAssertionFailureContextNotFound.class.st
StevenCostiou: I don't remember, this PR is more than a week old, maybe no reason.
I will check.

src/NewTools-Debugger/Context.extension.st
StevenCostiou: Two things:
- the debugger only look for those when it has insurance it is a test failure, so somewhere down the stack there is a test failure (just for information)
- it is better than before (even in GT) where those values were fetched from magic indexes in the test failure context
I am open to suggestions about how to do better though :)

src/Kernel/Context.class.st
StevenCostiou: > It can decrease the performance of StepOver button in the debugger (or other related).
> And more important it affects the process termination which performs the unwinds using this simulation machinery. See the method Process>>#doTerminationFromAnotherProcess.
> 
> I think you can just implement your own ExceptionSet class:
> 
> ```smalltalk
> SimulationRunTrapExceptionSet>>handles: anException
>    ^anException captureIfSignalledWhenStepping
> 
> Context>>runUntilErrorOrReturnFrom: aSender
>    ...
>    aSender insertSender: (Context
> 		contextOn: SimulationRunTrapExceptionSet new do: [:ex |
>    ...
> ```

But does it decreases performance in a significant way?
How does it affect the termination, do you mean in terms of performance?
Thing is, there is no other way of solving this than adding the test failure exception to that list, and it has to be built dynamically as you don't know beforehand which exception are going to interrupt the simulation.
For instance, I can't just add TestFailure, it is not in the Kernel, I need a dynamic way of finding it. I do not know what would be the fastest way, I did not understand your suggestion with `SimulationRunTrapExceptionSet`.

> And this way you will avoid the implementation like "self == Error". Simply return true or false as it really relevant for all subclasses

I will check, but I think it does not work. If you return true for a super class it enables interuption of simulation for all subclasses which is not supposed to happen in certain cases.


> Possible alternative to #captureIfSignalledWhenStepping is #shouldInterruptSimulation

Thanks! I'll take it :)

src/NewTools-Debugger/Context.extension.st
StevenCostiou: > It is very fragile logic. The names #actual and #expected is kind of SUnit pattern to name variables in the tests.
> For example your logic will not fail for test like:
> 
> ```smalltalk
> actual := 10 + 20.
> expected := 2.
> self assert: actual / 0 equals: expected
> ```
> 
> I did not check what debugger will show in that case. But the logic looks wrong.

Actually this produces a divide by zero exception, so as the debugger is not in a test failure context it is not going to try displaying the diff view.

src/NewTools-Debugger/Context.extension.st
StevenCostiou: > I would probably implement it a bit differently and not relying on variable names:
> 
> ```smalltalk
> Context>>isAssertionMethod
>     ^"better logic here based on selectors (maybe using pragmas to mark assertions)"
> 
> Context>>asAssertionResult
>   "Assertions are implemented with convention to have first argument as an actual value and the second argument as expected result"
>    self isAssertionMethod ifFalse: [ StDebuggeAssertionFailureContextNotFound signal ].
>    ^ StDebuggerAssertionFailureResult actual: self arguments first expected: self arguments second
> ```
> 
> And then the find method will be more simple and readable:
> 
> ```smalltalk
> Context>> findFailedAssertionResultsInSenderChain
>      assertionContext := self findContextSuchThat: [ :ctx | ctx isAssertionMethod ].
>      assertionContext ifNil: [ ^  StDebuggeAssertionFailureContextNotFound signal ].
>      ^assertionContext asAssertionResult
> ```

Thanks for the suggestion, it would be better by marking the assertion methods with pragmas and then accessing the arguments.
 I will open an issue.

src/Kernel/Context.class.st
dionisiydk: > How does it affect the termination, do you mean in terms of performance?

Yes. Of course it is all tiny benchmarks. But if you would measure the performance and memory of process termination it will be significant I believe. It requires collection of all subclasses of Exception - 342 classes in base image.

> I did not understand your suggestion with `SimulationRunTrapExceptionSet`

It's simple:

`[] on: Error, Halt do: [:er | ]` is equivalent to `[] on: (ExceptionSet new add:  Error; add: Halt) do: [:er | ]`

The #, message just produces an instance of ExceptionSet. You can pass there any object which understand #handles: message. Therefore you can implement your own analogue of ExceptionSet and avoid the concatenation of exception classes.

StevenCostiou: > 
> `[] on: Error, Halt do: [:er | ]` is equivalent to `[] on: (ExceptionSet new add: Error; add: Halt) do: [:er | ]`
> 
> The #, message just produces an instance of ExceptionSet. You can pass there any object which understand #handles: message. Therefore you can implement your own analogue of ExceptionSet and avoid the concatenation of exception classes.

Ah, that I know, but it cannot work like that because I need to add `TestFailure` directly to the exception set as it is not in the kernel. I need to find it dynamically. 

Maybe I can build an "extendable exception set" with extensions methods, that I dynamically recover through pragmas. Looking for pragmas in a single class should be much faster than looking in the exception class hierarchy like I did but I don't know for sure. I'm ok to fix the design (it would be better with your proposal I think), but I'm not up to compare performance of both solutions. I would need help on that side.

dionisiydk: Hi @StevenCostiou . It seems you did not get my example:
```Smalltalk
SimulationRunTrapExceptionSet>>handles: anException
   ^anException captureIfSignalledWhenStepping

Context>>runUntilErrorOrReturnFrom: aSender
   ...
   aSender insertSender: (Context
		contextOn: SimulationRunTrapExceptionSet new do: [:ex |
   ...
```
There is no reference to TestFailure there. But it does exactly same job as your solution with classes concatenation.

StevenCostiou: Ah. Sorry I still don't understand.
How would SimulationRunTrapExceptionSet work?
It has to:
- be in the kernel
- contain all the exception classes that interrupt the simulator
To satisfy this last point, it has to find those classes... which would have to include TestFailure.

dionisiydk: > Ah. Sorry I still don't understand.
> How would SimulationRunTrapExceptionSet work?
> It has to:
> 
> * be in the kernel

It's fine to be in Kernel as it represents a set of exceptions to be trapped during the simulation. Ideally all these simulation methods (runUntilErrorOrReturn and friends) should in own package but it is where we are.

> * contain all the exception classes that interrupt the simulator
>   To satisfy this last point, it has to find those classes... which would have to include TestFailure.

See the #handles: method in my example. It just asks the given exception if the simulation should be interrupted (no exception classes are hardcoded there). With my name it looks better:
```Smalltalk
SimulationRunTrapExceptionSet>>handles: anException
   ^anException shouldInterruptSimulation
```
Compare it with:
```Smalltalk
ExceptionSet>>handles: anException
	"Determine whether an exception handler will accept a signaled exception."
	^ exceptions anySatisfy: [ :exception | exception handles: anException ]
```
Does it makes sense now? I don't know how to explain it better.

StevenCostiou: Ah, yes I understand. I missed that above in your previous message, I understand now. It greatly improves the solution.
Thank you for explaining again :)

Well I realize now... you have put the entire solution (for the simulation interruption) into this discussion... would you consider implementing it and submitting a pull request? :)

dionisiydk: Hi @StevenCostiou .

I did not know when I will have time for this. Feel free to push it forward

dionisiydk: > ```smalltalk
> SimulationRunTrapExceptionSet>>handles: anException
>    ^anException captureIfSignalledWhenStepping
> 
> Context>>runUntilErrorOrReturnFrom: aSender
>    ...
>    aSender insertSender: (Context
> 		contextOn: SimulationRunTrapExceptionSet new do: [:ex |
> ```

Actually it is much simpler. I wonder why none of us did not see it: just handle the base class Exception and that's all:
```Smalltalk
aSender insertSender: (Context
 	contextOn: Exception new do: [:ex | 
                     ex shouldInterruptSimulation ifFalse: [ ex pass ].

```

dionisiydk: And I did PR #9422

PR #9232, opened 2021-05-02
fixes: 9231 new* into newTopToBottom and newLeftToRight

Ducasse:


Ducasse:
Ok since nobody cares about this issue I will just push it.

PR #9233, opened 2021-05-03
RB - Fix rename inst variable shortcut command

### LINKED ISSUE #9215: Ctrl + R to rename a variable in a System Browser does not work
**Bug description**
Ctrl + R shortcut  to rename a variable in a System Browser does not work

**To Reproduce**
Open a Pharo 9 64 image.
Open a System Browser on a method that uses an instance variable for example on AthensCairoPathBuilder >> context: 
Highlight the instance variable and click on Ctrl + R  and it raises an error. Instance of ClyMethodSourceCodeContext did not understand #requestSimpleVariable:from:
If we right click on the variable and rename it it works.

**Expected behavior**
Open the window dialog to rename the variable

**Screenshots**
image

Lin777:
Fixes  #9215


Ducasse:
@Evelyn do you have a test that covers this change?

src/SystemCommands-VariableCommands/SycMergeVariableCommand.class.st
```diff
@@ -12,6 +12,20 @@ SycMergeVariableCommand class >> defaultMenuIconName [
 	^ #merge
 ]
 
+{ #category : #activation }
+SycMergeVariableCommand class >> fullBrowserShortcutActivation [
+
```
Ducasse, line 17: is it normal to get such empty methods?

Lin777:
> @evelyn do you have a test that covers this change?

I'm not sure if there is a way to validate commands, as far as I saw, for the refactoring commands there are no tests to validate their correct activation. I will check if it exists for other commands, and if it exists I will add tests for this change

src/SystemCommands-VariableCommands/SycMergeVariableCommand.class.st
Lin777: I handled it like this because it inherited from a class that had its own activation, however if it is incorrect I can modify the hierarchy of the classes

PR #9234, opened 2021-05-03
RB - Improve rename method presenter

### LINKED ISSUE #5855: Cannot extract method with arguments
Extracting a method, I specify a keyword selector.
1. it is not clear if we should add the argument in the selector or in the list
2. it is not clear if the argument list should contain 
   1. argument names for the extracted method,
   2. or expressions for the replaced expression 

Trying to add the argument name for the extracted method (case 2.i) fails telling it is an invalid selector but it is not clear why
image

Trying to add an expression (case 2.ii) fails but it is not clear why
image

### LINKED ISSUE #9139: When I rename a method I cannot rename its parameter
I thought that now we could but I could not.
And when I remove the last parameter I cannot add one anymore. 
Screenshot 2021-04-25 at 22 15 17
Screenshot 2021-04-25 at 22 14 40

### LINKED ISSUE #9194: Add Argument Refactoring generates a MNU in RBMessageNode
1. Have a method like 

```
aMethod

	^ 42
```

2.  Have another method 

```
anotherMethod 

	^ self aMethod
```

3. Select "aMethod" and do Cmd+T and select "Add Argument"

image

Boom:

image

Lin777:
Fixes #9194
Fixes #9139
Fixes #5855

Video: https://youtu.be/mYeFbpMQhs8

Now you can rename arguments when you are renaming or extractirng methods

Rename method presenter
imagen

When you are adding parameters you can add more than one parameter, and also you can rename parameters

Add argument presenter
imagen


Ducasse:
@Lin777 this is cool. 
Now do you invoke the RenameArgumentRefactoring because the input from the user may be bogus.

Ducasse:
I read the code and I saw it. Quite a lot of work!

Lin777:
> @Lin777 this is cool.
> Now do you invoke the RenameArgumentRefactoring because the input from the user may be bogus.

Yes, in this change I focus on the view, but I will improve the renaming using RenameArgumentRefactoring

PR #9235, opened 2021-05-03
RB - Fix extract set up command

Lin777:
Changes hierarchy of Extract setUp command because I had a problem with a SourceCodeDefinition (It's a temporal fix)


Ducasse:
Ok let us know.

PR #9237, opened 2021-05-03
Fix-DocComments-Self-Reference 

MarcusDenker:
Doc comments where not correctly evaluating "self" references. A doc comment of the style

```
	"self >>> PharoDocCommentTest"
```
	
in class PharoDocCommentTest should evaluate the "self" to be the class

- add test #testExpressionReferencingSelf
- Cleanup/Simplify tests
- fix #evaluate in PharoDocCommentExpression to evaluate against the class as the receiver


Ducasse:
I like the tests :) with thisContext :)

PR #9243, opened 2021-05-03
Test no period in method signature downto 2

MarcusDenker:
With Iceberg merges, we have just 2 methods with periods in the first line of a method left.

To not add new cases, this PR makes sure we test for 2. The two cases are in Roassal, an issue tracker entry has been added tot he issue tracker and the next merge for sure will fix this.


MarcusDenker:
ups, did this with the wrong image

PR #9246, opened 2021-05-04
Fix some more doc comments

MarcusDenker:
- Collection>>#unzip: fix formatting
- SequenceableCollection>>#atWrap:put: , we can not compare an array with an OrderedCollection
- Context>>#setNamedPrimitiveInformationFrom:toMethod: To me this is too low level for a doc comment
- Point>>#sideOf: was just wrong


Ducasse:
Marcus is the formatting breaking them? Because we could then use the Parser to avoid to get such brittleness.

MarcusDenker:
Yes, somehow the current implementation expects everything in one line. You can see that already from how it is syntax highlighted. I will have a look...

Ducasse:
This is probably a hack of the first version.

Issue #9250, opened 2021-05-05
Context Menu (Right mouse menu) in Method-Sourcecode does not work


hoemac:
**Bug description**
ContextMenu in SourceCode of arbitrary Method in SystemBrowser does not work. 

**To Reproduce**
1. Downloaded latest Pharo 9 image  May 5th 2021 
2. Open a System Browser
3. Browse arbitrary method and put cursor anywhere in the source code 
4. Click right mouse button 

**Expected behavior**
Context menu dows not work,  appears with debug information after running into an exception (see screenshot)  

**Screenshots**
![grafik](https://user-images.githubusercontent.com/38470493/117168931-e6ab9100-adc8-11eb-87cf-cde6c23543aa.png)

![grafik](https://user-images.githubusercontent.com/38470493/117169560-7bae8a00-adc9-11eb-959b-8b1e57ac4686.png)

**Version information:**
 - OS: Windows
 - Version: 10
  - Pharo Version [ Pharo 9.0.0
Build information: Pharo-9.0.0+build.1390.sha.6e3dceda6ba1cfea9bfd09147470dfd1c23164e7 (64 Bit)]

**Additional context**
Showstopper for Pharo 9 (on Windows 10) for me if if no copy/paste etc is possible.

rolandbernard:
I also encountered this issue. It is caused by the changes made with #9103.
The AND-operation in `SycDeprecateClassCommand class>>canBeExecutedInContext:` should probably be lazy (i.e. `and: [...]`).

MarcusDenker:
This is fixed by merging the PR (I checked the latest version)

PR #9251, opened 2021-05-05
Fixes the issue #9250: The context menu now opens correctly

rolandbernard:
This fixes issue #9250.
The only change here is replacing the `&` with a `and: [...]` to avoid calling `lastSelectedClass` if no class is selected.


Ducasse:
Thanks!!! Indeed!

MarcusDenker (approved):

PR #9254, opened 2021-05-06
DocComments-fixes-tests 

MarcusDenker:
- add testMultipleDocComments: test that we can add multiple comments with doc comments
- add testMultipleDocCommentsInOneComment: test that one comment can contain multiple >>> expressions
- fix more existing doc comments


MarcusDenker:
as this just adds two tests that are green and else just changes comments, I will merge it. (even if the doc comment fixes I did should be wrong, they are still just comments...)

PR #9260, opened 2021-05-06
9145-Cursor-is-too-small-when-using-display-scaling

### LINKED ISSUE #9145: Cursor is too small when using display scaling
**Bug description**
The cursor is not scaling according to the parameters in the options.

mstsc_a7gVOeVHBG

**To Reproduce**
image

**Expected behavior**
bigger cursor

**Version information:**
 - OS: Win 10
Pharo9.0.0
build: 1354, commit: 1846e54, date: 23 April 2021

**Expected development cost**
Not too much

**Additional context**
Linked with https://github.com/pharo-project/pharo/issues/2883 ?

tesonep:
- Adding support for system cursors
- Adding support for overriding cursors of Pharo with System Cursors
- Adding support for changing the mapping per platform
- Fixing the scaling of cursors 
- Fixing the conversion to SDL cursors to handle different sizes (required for scalling)
- Using the scaling of the World in the setting

Fix #9145

### DIFF
--- a/src/Deprecated90/OSSDL2BackendWindow.extension.st
+++ b/src/Deprecated90/OSSDL2BackendWindow.extension.st
@@ -1,0 +1,9 @@
+Extension { #name : #OSSDL2BackendWindow }
+
+{ #category : #'*Deprecated90' }
+OSSDL2BackendWindow >> convertCursor: cursor withMask: mask [
+
+	self deprecated: 'Use #convertCursor:withMask:andScale:'.
+
+	^ self convertCursor: cursor withMask: mask andScale: 1
+]
--- a/src/Graphics-Display Objects/Cursor.class.st
+++ b/src/Graphics-Display Objects/Cursor.class.st
@@ -1079,6 +1079,17 @@ Cursor >> printOn: aStream [
 	self storeOn: aStream base: 2
 ]
 
+{ #category : #primitives }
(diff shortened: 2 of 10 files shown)


VincentBlondeau:
Working well with a display factor of 1.
What happens with a display factor superior or inferior:
image

C6zCzHR4Uw

tesonep:
@VincentBlondeau are you testing with the image product of the CI? 
https://ci.inria.fr/pharo-ci-jenkins2/job/Test%20pending%20pull%20request%20and%20branch%20Pipeline/job/PR-9260/1/artifact/bootstrap-cache/Pharo9.0-PR-64bit-74e5bc9.zip

Or loading the PR directly. 

Loading the PR branch in a Pharo image does not update correctly the changes... I think we have found an issue in iceberg as the methods are being used at the same time (it should not happen, of course).

I have tested with the produced image in the CI, and looks that is working. Can you try with that one?

Thanks a lot.

<blockquote></blockquote>

VincentBlondeau:
Tried on the PR image. It works fine if the scale factor is an integer (1, 2, 3) but not otherwise (0.5, 1.5, 2.5,...): it is then behaving like the bug described above

tesonep:
I have fixed the rounding because cursors width should be multiple of 8

PR #9264, opened 2021-05-07
add knownPaths for openbsd

qbit:
OpenBSD ports installs things into `/usr/local`, also X11 libraries live in `/usr/X11R6/lib`, not 100% sure if X11 stuff is needed.


estebanlm (approved):

estebanlm:
not sure about X11R6 either, but let's see

MarcusDenker (commented):
maybe we can first try without adding the X11 path?

qbit:
Removed X11R6 for now!

Issue #9252, opened 2021-05-05
Adding a breakpoint throws an exception in the latest build


rolandbernard:
**Bug description**
Attempting to add a breakpoint in the current image from the method editor in the system browser will throw an exception (subclassResponsibility).

**To Reproduce**
1. Open a method in the system browser. 
2. Try to add a breakpoint (either using the context menu or the left side).

**Expected behavior**
A breakpoint should be set without throwing an exception.

**Screenshots**
![Capture11](https://user-images.githubusercontent.com/18496639/117194333-44010b80-ade4-11eb-8e89-2e7a5c53eaad.PNG)

**Version information:**
 - OS: Windows 10
 - Pharo Version 9.0.0, build: 1390, commit: 6e3dced, date: 4 May 2021

**Expected development cost**
I don't know. I can take a closer look at this later this week, if no one else fixes it.

**Additional context**
It works in fa306ee1ac254116ab697044213430d30c484947 and breaks in 4a9cc4ad9c4f219a1a6fe175c54e070fec160ec6. So it seems this issue starts with the merge of #9146.

PR #9266, opened 2021-05-10
RB - Fix show previews of refactorings from Source Code

Lin777:
This change is necessary so that all refactorings have a preview of the changes, in this case the refactorings from source code

Related to: #9261 and #9252


MarcusDenker:
There is now a conflict due to merging #9252

PR #9276, opened 2021-05-12
improving-drtests-plugin-support

tesonep:
Spliting the responsability of DrTests presenter in two, so the plugins can configure and give a different presenter


Ducasse:
Pablo noUnusedVariable test seems to fail.

tesonep:
Great, the test has found some variables I left by mistake... I like when something protects the world from myself 😄

astares:
@Ducasse / @tesonep  Did you see that the PR contains #classSide in Tonel not #class. Is this correct?

PR #9277, opened 2021-05-13
Rewrote and improved String >> skipAnySubstring:startingAt:

marcor:
- Moved part of the code to new separate method String >> hasSubstring:At:
- Added tests 
- Temporarily renamed original method to String >> oldSkipAnySubstring:startingAt:

This should fix issue #6191. 

I tried my hand at this after studying @dupriezt previous pull request #6230 and the discussion. Most of the new tests come from there.

I used detect:ifFound:ifNone: and allSatisfy: for a tighter control of the loops. One difference compared to the orginal method is that every time a delimiter is found, the iteration of the delimiters array restarts from the first element instead of continuing to the end.

I mentioned in the documentation that the order of the delimiters array can have an effect on the result. I don't think it could be otherwise.

I noticed an increase in performance for more complex/longer input, and worse but still comparable performance in other cases, but I haven't really profiled this for the time being.

### DIFF
--- a/src/Collections-Abstract-Tests/TAsStringCommaAndDelimiterTest.trait.st
+++ b/src/Collections-Abstract-Tests/TAsStringCommaAndDelimiterTest.trait.st
@@ -95,18 +95,17 @@ TAsStringCommaAndDelimiterTest >> testAsStringOnDelimiterLastEmpty [
 { #category : #'tests - as string comma delimiter sequenceable' }
 TAsStringCommaAndDelimiterTest >> testAsStringOnDelimiterLastMore [
 
-	| delim multiItemStream result last allElementsAsString tmp |
+	| delim multiItemStream result last allElementsAsString |
 	
 	delim := ', '.
 	last := ' and '.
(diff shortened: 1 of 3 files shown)


marcor:
Actually though, this piece of code:

`'text' findBetweenSubstrings: #('te' 'x')`

still returns the same wrong result. It seems the problem is not entirely in skipAnySubstring like I assumed.

Ducasse:
Thanks this is nice!!!

marcor:
> I don't think it could be otherwise.

On second thought the order of delimiters seems to have an effect on the result only in situations like these:

`#('abc', 'abcdef')`

If a short delimiter matches the start of a longer one, and precedes it in the array, the longer one will never, ever have a chance to be found. In the old method it could probably happen, based on chance. @dupriezt method solved this I believe, at the cost of more string comparisons.

It could make sense to sort the array according to the length of the delimiters before starting the search. Or advise the sender to take care of it. As of now the only sender is String >> findBetweenSubstrings:.

PR #9281, opened 2021-05-16
fixing light/dark themes disabled colors

estebanlm:
(this will work better with graying also icons, which will be present on spec controls)


Ducasse:
broken tests are unrelated.

PR #9292, opened 2021-05-19
fix SHRBTextStyler messing Transcript.

pavel-krivanek:
When you type a method with opened parenthesis,  SHRBTextStyler writes SubscriptOutOfBounds error on every keystroke.

### DIFF
--- a/src/Shout/SHRBTextStyler.class.st
+++ b/src/Shout/SHRBTextStyler.class.st
@@ -849,9 +849,9 @@ SHRBTextStyler >> addAttributes: attributes forNode: anRBNode [
 { #category : #formatting }
 SHRBTextStyler >> addAttributes: attributes from: start to: stop [
 	charAttr
-		from: start
-		to: stop
-		put: attributes
+		from: (start min: charAttr size) 
+		to: (stop min: charAttr size)
+		put: attributes 
 ]
 
 { #category : #converting }


MarcusDenker (approved):
Yes, I think this is a good first fix.

I want to check why some of the offsets start with 0 instead of 1 (it seems it is used as kind of a "nil" value). But that change would be quite intrusive and thus for Pharo10.

I think we should therefore accept this fix now.

kasperosterbye:
Sorry, but it seems like it is still not functioning in build: 1425, commit: a821063, date: 21 May 2021, which has the error correction suggested (https://github.com/pharo-project/pharo/pull/9292/files). 

Starting a method with a number still produce an error logged to Transcript.

The problem in the workaround for `SHRBTextStyler>>addAttributesfrom:to:`. The line:

```smalltalk
from: (start min: charAttr size)
```
should have been:
```smalltalk
from: (start max: 1) 
```


<blockquote><div> GitHub</div><div><strong><a href="https://github.com/pharo-project/pharo/pull/9292">fix SHRBTextStyler messing Transcript. by pavel-krivanek · Pull Request #9292 · pharo-project/pharo</a></strong></div><div>When you type a method with opened parenthesis,  SHRBTextStyler writes SubscriptOutOfBounds error on every keystroke.</div></blockquote>

PR #9320, opened 2021-05-22
Refactor Form >> fromUser

### LINKED ISSUE #5918: Refactor Form>>fromUser
`Form >> fromUser` freezes event management and makes the user select a rectangle from the screen. Then it builds a form from the selected rectangle by clipping the receiver form.

- This functionality should not be part of the form, it should be extracted
- It should get a world/display by argument instead of accessing global state (and thus use `Rectangle fromUser: gridPoint` with a world)

AleCossioCh:
Fix #5918

- Refactor all classes that used Form >> fromUser
- Add fromUser:inWorld:

### DIFF
--- a/src/Graphics-Canvas/BitmapFillStyle.class.st
+++ b/src/Graphics-Canvas/BitmapFillStyle.class.st
@@ -34,7 +34,7 @@ BitmapFillStyle class >> fromForm: aForm [
 { #category : #'instance creation' }
 BitmapFillStyle class >> fromUser [
 	| fill |
-	fill := self form: Form fromUser.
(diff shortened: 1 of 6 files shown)


estebanlm:
hi, 
thanks for your contribution, but I'd like to know what's the purpose if this refactor?
Because we are in frozen mode, which means no other than bugfixes should be allowed, and if this can wait until  we open P10, I would prefer to hold the integration and redirect there as soon as we open the new development branch.

AleCossioCh:
Hi @estebanlm,
I tried to get a world by argument instead of accessing global state. The issue is #5918
But, yes indeed! I think it can wait.

Ducasse:
We can just tag it as Pharo 10

PR #9323, opened 2021-05-24
7974 file class  open and truncate for write file named is a clunky name

### LINKED ISSUE #7974: File class >> openAndTruncateForWriteFileNamed: is a clunky name
I'm sure that we can do better.

AleCossioCh:
Fix #7974
- Add deprecation >>openAndTruncateForWriteFileNamed:

### DIFF
--- a/src/Files/File.class.st
+++ b/src/Files/File.class.st
@@ -581,6 +581,13 @@ File class >> open: fileName writable: writableFlag [
 ]
 
 { #category : #'file creation' }
+File class >> openAndTruncateForWriteFileNamed: aName [
+	self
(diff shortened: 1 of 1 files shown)


Ducasse:
Alejandra I should have told you that you can use deprecated: transformWith: check the actual users because like that the system can rewrite users of the functionality automatically.

Ducasse:
Yes!
Thanks. 
Now you should pay attention because we are in beta so we do not accept changes that changes too much. 
This one should be ok.

estebanlm:
uhm... I am not so sure Iike the new name more than the old one :P

Ducasse:
After discussion we will deprecated this method.

PR #9325, opened 2021-05-26
ReleaseTest-DocComments

MarcusDenker:
This PR adds DocCommentsReleaseTest, a test class that creates a test for each doc comment (see #buildSuite on the class side).

- for now just test instance side methods (DrTest does that, we need to fix it there first and fix the broken cases)

This adds 892 tests

### DIFF
--- a/src/PharoDocComment-Tests/DocCommentsReleaseTest.class.st
+++ b/src/PharoDocComment-Tests/DocCommentsReleaseTest.class.st
@@ -21,5 +21,7 @@ DocCommentsReleaseTest class >> buildSuite [
 	methods do: [ :method | 
 		method pharoDocCommentNodes do: [ :docComment | 
 			suite addTest: (CommentTestCase for: docComment) ] ].
+	"make sure to clean up the AST Cache to allow the GC to clean up"
+	ASTCache reset.
 	^ suite
 ]
--- a/src/PharoDocComment-Tests/DocCommentsReleaseTest.class.st
+++ b/src/PharoDocComment-Tests/DocCommentsReleaseTest.class.st
@@ -6,16 +6,16 @@ See #buildSuite on the class side.
 Class {
 	#name : #DocCommentsReleaseTest,
 	#superclass : #TestCase,
-	#category : #'PharoDocComment-Tests'
+	#category : #'PharoDocComment-Tests-Base'
 }
 
 { #category : #'building suites' }
 DocCommentsReleaseTest class >> buildSuite [
-		| suite methods |
+	| suite methods |
 	suite := TestSuite named: 'Test Generated From Comments'.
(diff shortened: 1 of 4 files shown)


src/PharoDocComment-Tests/DocCommentsReleaseTest.class.st
```diff
@@ -0,0 +1,25 @@
...
+	#category : #'PharoDocComment-Tests'
+}
+
+{ #category : #'building suites' }
+DocCommentsReleaseTest class >> buildSuite [
+		| suite methods |
+	suite := TestSuite named: 'Test Generated From Comments'.
+	"we do not run examples in tests (as they are used just for testing and might be broken by
+	design"
+	methods := Smalltalk globals allMethods reject: [ :method | method methodClass isTestCase ].
```
Ducasse: can you remove the Smalltalk globals ref?

src/PharoDocComment-Tests/DocCommentsReleaseTest.class.st
```diff
@@ -0,0 +1,25 @@
...
+"
+Class {
+	#name : #DocCommentsReleaseTest,
+	#superclass : #TestCase,
+	#category : #'PharoDocComment-Tests'
+}
+
+{ #category : #'building suites' }
+DocCommentsReleaseTest class >> buildSuite [
+		| suite methods |
```
Ducasse: tab too much

src/PharoDocComment-Tests/DocCommentsReleaseTest.class.st
MarcusDenker: yes. But it shows that we are missing a mechanism: the only thing I can do is to use the environment of the test class. But that is not what I mean: I want to write "System, give me all methods, regardless of the environment that this class is in". 

Of course "Smalltalk globals" is not good: for one it is a global reference. And it hardcodes "give me all methods from the global environment", which is not the same as "give me all methods, not only those from classes in the global environment"

Right now there is only one environment, so "self class environment" and "Smalltalk globals" happen to point to the same environment.

I will use self class environment for now.

Ducasse:
Super nice.
testNoUnusedTemporaryVariablesLeft – MacOSX64.ReleaseTests.CleanCode.NoUnusedVariablesLeftTest
testNoShadowedVariablesInMethods – MacOSX64.ReleaseTests.Release.ReleaseTest

MarcusDenker:
failing tests are timeouts on the mac machine, it sometimes happens. I think it is running slow due to virtualisation.

MarcusDenker:
Ah, I think it makes sense to add a reset for the ASTCache, this way the cached doc comments node instances get cleaned up.
(I have an idea how to speed up parsing, then we will not need to cache them anymore, but that is for Pharo10)

Issue #9324, opened 2021-05-25
Remove `accesing` protocol


akevalion:
**Problem**
`accesing` is not an english word, but you can see it some times from protocols list.

**Classes involved**
a Set [10 items] (DAPackageAnalyzerWindow IceBackendWriterAdapter RGEnvironmentBackend QuotedPrintableMimeConverterTest RBErrorToken IceTipWorkingCopyDiffModel LGitTag AbstractSpecLayoutAction SDL_TextInputEvent IceRemote)

**Proposal**
Rename the protocol `accesing` to `accessing`
```Smalltalk
violating:= Set new.
SystemNavigation default
	allClasses do: [ :class | 
		(((class protocols ifNil: #()), 
		(class class protocols ifNil: #())) includes: 'accesing') 
			ifTrue: [ violating add: class ].
		 ].
```

**Version information:**
 - Version [e.g. Pharo 9.0]

**Expected development cost**
This should be an easy issue

![image](https://user-images.githubusercontent.com/10532890/119516511-d0984d00-bd44-11eb-9d42-b8e7edf1db63.png)

MarcusDenker:
These need to be commited to the iceberg repository:

IceBackendWriterAdapter
IceTipWorkingCopyDiffModel
LGitTag
IceRemote


The rest I think can be done in the main pharo repo (AbstractSpecLayoutAction is from spec1)

MarcusDenker:
All merged!

PR #9336, opened 2021-05-28
9324-Remove-accesing-protocol

MarcusDenker:
fix all image cases (outside of iceberg/libgit)

one step for #9324


gcotelli (approved):
LGTM

PR #9337, opened 2021-05-28
[Enhancement] Smoother drag and drop in calypso

hogoww:
I know we're in a freeze, but this might still be relevant for p10.

When drag and dropping something in a package, the current behavior is to open the package and display the tags for that package.
However, it never closes it.
Drag n dropping a class in a package can be quite tedious to do.
If we do not hover directly on the right package, tags opened for other packages.
The opened tags are preventing access to the package that we want.

With this small fix, it collapses whichever package was opened, allowing for significantly smoother drag and drop operations.

However, I do not know how to test this without passing the rest of the day doing it..
Suggestions welcome :)

Pierre

### DIFF
--- a/src/Calypso-Browser/ClyQueryViewMorph.class.st
+++ b/src/Calypso-Browser/ClyQueryViewMorph.class.st
@@ -115,7 +115,7 @@ Class {
 		'changesWasInitiatedByUser',
 		'navigationSelector',
 		'dataSourceClass',
-		'expandedItem'
+		'expandedItemByDrag'
 	],
 	#category : #'Calypso-Browser-Table'
 }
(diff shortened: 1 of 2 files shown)


src/Calypso-Browser/ClyQueryViewMorph.class.st
```diff
@@ -114,7 +114,8 @@ Class {
 		'highlighting',
 		'changesWasInitiatedByUser',
 		'navigationSelector',
-		'dataSourceClass'
+		'dataSourceClass',
+		'expandedItem'
```
dionisiydk: It would be better to have more concrete name like expandedItemByDrag. 

But generally this state needs to be managed together with other expansion state like tree expansion in class view for example. 
ClyDataSource is responsible for this. And it handles the cases where expanded items are updated (when they are removed for example). Maybe it does not matter for drag scenario but I can see how the view can get inconsistent state and possibly some errors can happen.

dionisiydk: From the other side I would expect different fix for the problem with "too fast drag". And I think there is existing issue with description:
Normally UI (non Pharo UI) requires to hang the drag item for a while over the drop target to force the expansion. Now it is expanded immediately. And that's what we need to change.

hogoww: I completely agree with everything you said, and that was my initial objective.
Buuuuuuut, I wasn't able to do it :/
This fix is not meant to be a definitive one, just to have an easier time dealing with drag and drops.

I'll do the rename though thanks ! =)

hogoww:
To be honest, after looking at this again, I think it's also pretty nice to do this way.
Even though it is indeed unconventional. 
(But I'm not a fan of the code i provided to do so).

guillep:
I've checked this locally, and while it gives the impression of a smoother drag, it does not allow to drag classes into tags anymore.

Ideally, the package should expand after some delay only.

PR #9358, opened 2021-05-28
testNoUnusedInstanceVariablesLeft-set-to-zero

MarcusDenker:
We can now check for isEmpty in testNoUnusedInstanceVariablesLeft, MockWithComplexSlot is for now in valid exceptions.


astares:
😎

PR #9366, opened 2021-05-28
9365 cpu watcher should use correct startup  shutdown category

### LINKED ISSUE #9365: CPUWatcher should use correct startup / shutdown category
"startup-shutdown" should be "startup - shutdown" so we do not have it twice in the list

astares:
Fix #9365


Ducasse:
Strange the build failed on all platforms.

astares:
@Ducasse 
Strange CI - as it is only category stuff.

But its green on unix-64

Ducasse:
Ok much better now. Probably an hiccup.

PR #9377, opened 2021-05-31
9107-Syntax-highlighter-does-not-refresh-when-a-method-is-saved

### LINKED ISSUE #9107: Syntax highlighter does not refresh when a method is saved
**Bug description**
When a method with undefined variables is saved, the browser asks to define them.
When we select our choices, the method is edited and finally compiled => ok!

But, the syntax highlighter still highlights wrongly the just defined variables.
It seems it is not refreshed.

Happened when defining instance variables.

Latest Pharo 9

MarcusDenker:
This PR fixes #9107. We need to do some changes as the browser now keeps the AST in some cases where it used to re-parse.

- move the ReparseAfterSourceEditing handler from #compiledMethod to #parse
- make sure to raise ReparseAfterSourceEditing not only for the case we add a temp but for global, class, and Ivars, too
- remove the now unused (private) method  #lookUpVariable

Doing this change it got clear that we should re-evaluate if it would not be better to remove the whole #ReparseAfterSourceEditing concept.

But that will have to wait for Pharo10 as it would be a large change.

### DIFF
--- a/src/OpalCompiler-Core/OCUndeclaredVariableWarning.class.st
+++ b/src/OpalCompiler-Core/OCUndeclaredVariableWarning.class.st
@@ -12,21 +12,21 @@ OCUndeclaredVariableWarning >> declareClassVar [
 
 	self methodClass instanceSide
 		addClassVarNamed: node name asSymbol.
-	^ self lookUpVariable
+	(ReparseAfterSourceEditing new newSource: self requestor text) signal
 ]
 
 { #category : #correcting }
 OCUndeclaredVariableWarning >> declareGlobal [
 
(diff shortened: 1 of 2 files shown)


dionisiydk:
Sadly no test :(
And without test it's difficult to understand the actual problem scenario (out of the highlighting case)

Ducasse:
Agreed!

MarcusDenker:
I will try to come up with tests.. no that easy... and I want to remove the whole ReparseAfterSourceEditing because it is so complex (hard to understand and test, but that has to wait for Pharo10)

PR #9380, opened 2021-05-31
Fix several method category issues

### LINKED ISSUE #9379: Fix several method category issues
- MockSourceEditor has uncategorized methods
- RGBehaviorStrategyUser has uncategorized methods
- RGTraitTransformation has uncategorized methods
- we should follow "xxxx - yyyyy" not "xxxx-yyyyy" category name pattern
- category for tests in TestCase subclass is not "testing" but "tests" and not "testing-yyyyy" but "tests - yyyyy" 
- utility methods should be in "utilities" method category (fix lint lamenting)
- small formatting issues
- remove some unnecessary dots at end of method
- ...

astares:
Fix #9379

### DIFF
--- a/src/AST-Core-Tests/NumberParserTest.class.st
+++ b/src/AST-Core-Tests/NumberParserTest.class.st
@@ -22,7 +22,7 @@ NumberParserTest >> classToBeTested [
 	^ NumberParser
 ]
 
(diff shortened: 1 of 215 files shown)


Ducasse:
Thanks. I prefer these long PR over many small because the small ones cost more energy to handle them.

PR #9385, opened 2021-06-02
Method category cleanup - pass 2

### LINKED ISSUE #9384: Method category cleanup - pass 2
- fixed needed by categorizing uncategorized methods
- use "xxx - yyyy" pattern in some method categories (where it makes sense) instead of "xxx-yyy"
- remove a few dots at end of method body

astares:
fix #9384


Ducasse:
This is strange since all the Fuel tests are breaking. I imagine that this is unrelated.

astares:
Changes are simple. Can you restart CI on this? 🤔

MarcusDenker:
If I lool at the diff, it seems that the PR removed some extension methods, this could be the reason. I will restart

MarcusDenker:
The Fuel tests are failing in another PR check, too, so it seems to be not caused by this PR

PR #9387, opened 2021-06-02
Avoid calling the link and changing the control, which invalidates th…

### LINKED ISSUE #9386: When VariableBreakpoints hit in the debugger, we can't live edit the method
- install a varibale breakpoint on the variable of a class
- run a method of that class touching that variable
- when the debugger opens, try to edit the method and save
- 
There will be an exception because the method on the stack is a `ReflectiveMethod` and not a `CompiledMethod`.
In many cases the execution is messed up just after.

StevenCostiou:
…e link and replaces the compiled method by its reflective method.

Fixes #9386

### DIFF
--- a/src/Reflectivity-Tests/VariableBreakpointTest.class.st
+++ b/src/Reflectivity-Tests/VariableBreakpointTest.class.st
@@ -377,6 +377,15 @@ VariableBreakpointTest >> testInstall [
 ]
 
 { #category : #'tests - installation' }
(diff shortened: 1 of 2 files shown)


StevenCostiou:
2k tests failing?

Ducasse:
Probably a problem with fuel loading. @estebanlm

MarcusDenker (approved):

MarcusDenker:
test now are good with fuel fixed, I merge

PR #9393, opened 2021-06-03
3211-Creating-a-missing-class-in-debugger-still-shows-it-missing-in-browsercritique

### LINKED ISSUE #3211: Creating a missing class in debugger still shows it missing in browser/critique
If you have a test method (but could be any method) referencing a missing class (either load it from a filein, or save it using the new leave undeclared menu option) and the create the class via the debugger by running the test, the class is correctly created however Calypso (possibly the code critic) still shows the class as undeclared.  Opening a new browser doesn't work - however running the test again, its happy the class is there. Re-saving the method does correct it (so something is out of sync).

An easy way to prove this is to run through the [pharo-exercism](https://exercism.io/my/tracks/pharo-smalltalk) first exercise (hello-world or two-fer).

image

### LINKED ISSUE #3246: Compiling a class does not reset the cache of compiled method using it

MarcusDenker:
This fixes #3211 by making sure to recompile all methods that use the class. We can just use #usingMethods, as the class builder has already made sure that the class binding (the Global Variable) is the correct one.

### DIFF
--- a/src/OpalCompiler-Core/OCUndeclaredVariableWarning.class.st
+++ b/src/OpalCompiler-Core/OCUndeclaredVariableWarning.class.st
@@ -90,7 +90,7 @@ OCUndeclaredVariableWarning >> defaultAction [
 OCUndeclaredVariableWarning >> defineClass: className [ 
 	"Prompts the user to define a new class."
 	
-	| classSymbol systemCategory classDefinition |
+	| classSymbol systemCategory classDefinition classBinding |
 	classSymbol := className asSymbol.
 	systemCategory := self methodClass category
 		ifNil: [ 'Unknown' ].
@@ -108,8 +108,10 @@ OCUndeclaredVariableWarning >> defineClass: className [
 		source: classDefinition;
 		logged: true;
 		evaluate.
-	^ (node owningScope lookupVar: className)
-		ifNil: [self error: 'should be not happen']
+	classBinding := node owningScope lookupVar: className.
+	"make sure to recompile all methods referencing this class"
+	classBinding usingMethods do: [:method | method recompile].
+	^classBinding 
 ]
 
 { #category : #correcting }


src/OpalCompiler-Core/OCUndeclaredVariableWarning.class.st
```diff
@@ -108,8 +108,10 @@ OCUndeclaredVariableWarning >> defineClass: className [
 		source: classDefinition;
 		logged: true;
 		evaluate.
-	^ (node owningScope lookupVar: className)
-		ifNil: [self error: 'should be not happen']
+	classBinding := node owningScope lookupVar: className.
+	"make sure to recompile all methods referencing this class"
+	classBinding usingMethods do: [:method | method recompile].
```
dionisiydk, line 113: It is unclear why it fixes any problem. The class definition evaluation (like in the browser) is supposed to recompile all required methods. Why it does not work in that case?

It looks like that the defining class not from the warning should lead to the same issue. Is it?

dionisiydk:
And sadly no test :) . Is it difficult to reproduce?

src/OpalCompiler-Core/OCUndeclaredVariableWarning.class.st
MarcusDenker: If we define a class, the AST of the methods that reference that class still has the wrongly analysed variable, thinking
it is Undeclared (if the AST exists).

(the compiled methods are ok, the Undeclared mechanism made sure to update the bindings without the need to recompile).

For now, I just added a recompile step to make sure the methods and the AST (if there are some) are in sync… I wanted to avoid adding lots of code now in Pharo9, forcing a recompile should be an ok intermediate fix.

What we need to do for real is to instead fix the AST name analysis for this variable.

I will do that better version, but as part of Pharo10, I think, as part of the project to simplify handling of Undeclared Variables  (not doing it in the debugger or the Exception).

(I should have added more info to the issue tracker entry…)

PR #9396, opened 2021-06-03
fix emergency debugger

estebanlm:
add a more or less working version of ED, just modifying the essential: 
it nows open a separated window with its own event handler to process events. 
the rest should work as expected...


Ducasse:
This is super cool to have an example of how we can have different handler!

estebanlm:
tests pass... is maybe not the best solution possible, but is one that seems to be working (at least, it cannot be worst than before, without any emergency debugger at all).

PR #9403, opened 2021-06-05
Fixes #9402: DumpVisitor new contents

### LINKED ISSUE #9402: DumpVisitor new contents
DumpVisitor should give an API to get the result. 
Doing stream contents in client is not that nice.

Ducasse:
- Add contents to RBDumpVisitor
- Update dump
- Extract Smalltalk compiler from test methods
- Add visit: to RBRProgramNodeVisitor


MarcusDenker (commented):
(This shows again that we need some nice way to express "I want the default compiler of the system" vs. "the compiler of this class")

PR #9408, opened 2021-06-06
Fixes #8668: Scaled Decimal multiplication loses precision

### LINKED ISSUE #8668: Scaled Decimal multiplication loses precision
**Describe the bug**
1.003s3 * 1.006s4.

 => 

1.0090s4

instead of something like:

1.009018

**Expected behavior**

This happens because scaled decimal multiplication is just taking the largest scale between the two scales. One possibility to keep precision is to multiply the scales instead (and taking care of the cases where scale = 0)

carlotxra:
For multiplication of 2 scaled decimals we answer a scaled decimal with a scale equal to the sum of the operand scales


guillep (approved):
Looks ok to me

PR #9412, opened 2021-06-07
RB - Fix refactoring presenters

### LINKED ISSUE #5894: Refactoring dialogs should not be modal
Or not as they are right now.
Right now, we have a succession of modal dialogs.
Each modal dialog is put in a "FullScreenMorph", thus putting it on top, and putting a dark gray background behind it (which BTW does not fully work because we have access to the main top menu and the taskbar, and that somehow workarounds the modality).

image

In this situation, the method/class where I started the refactoring from gets hidden / partially covered.
And the dialog does not help me remembering what I was doing.
This gets even worse when there is a sequence of dialogs (gimme a new method name, gimme a new argument name).

It seems this behaviour comes from using Spec2 `openModalWithSpec`

image

If refactoring dialogs are made modal, they should at least contain all the information required to make the decision:
 - the code before
 - the code after
 - the impacted methods/classes

Otherwise, I'd propose we stick to the simple non-modality.

### LINKED ISSUE #9060: Enhance Refactoring Preview Widget
The refactoring preview is pretty naive for the moment

image

**Describe the request**
We should be able to:
 - do multiple selection and check/uncheck what is selected
 - we should also have a (de)select all button maybe?
 - right click give a context menu with some navigation options
 - modal is completely wrong, if the changes are complex, I cannot go and verify my change in the browser
    - this means we should be able to navigate to the modified item from the preview

### LINKED ISSUE #9263: preview default button should be apply
so that we can do 

- command r 
- edit
- enter (for rename)
- enter (for apply)

Lin777:
Fixes #9060 
Fixes #5894
Fixes #9263

Use ENTER -> accept button
Use ESC -> cancel button
Use backspace -> select/unselect items in a table

### DIFF
--- a/src/SystemCommands-RefactoringSupport/CmdCommand.extension.st
+++ b/src/SystemCommands-RefactoringSupport/CmdCommand.extension.st
@@ -10,7 +10,7 @@ CmdCommand >> confirmRefactoringInContext: aToolContext by: aCommandActivator [
 	submitDialog openDialogWithSpec
 		okAction: [ submitDialog accept.
 			aCommandActivator applyCommandResult ];
-		cancelAction: [ submitDialog cancel ]]
+		cancelAction: [ :p | p beCancel ]]
 	ifFalse: [ submitDialog alert: 'There is no change produced by this refactoring' ].
 	CmdCommandAborted signal
 ]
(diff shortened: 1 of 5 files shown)


Ducasse:
yes yes yes!

PR #9418, opened 2021-06-08
Modify UnixResolver>>xdgParseUserDirLine: to handle the home directory

### LINKED ISSUE #9410: UnixResolver>>xdgParseUserDirLine: returns root if XDG_DOCUMENTS_DIR is set to $HOME/
**Bug description**

If the environmental variable `XDG_DOCUMENTS_DIR` in `~/.config/user-dirs.dirs` (on Linux systems) is set to

```
XDG_DOCUMENTS_DIR="$HOME/"
```

`UnixResolver>>xdgParseUserDirLine:` will end up calling `self home / ''`, which will return the root directory. Note that this problem only occur if there is a trailing slash after `$HOME` specifically. There is no problems if you only have `$HOME` or a slash after another level. For example,

```
XDG_DOCUMENTS_DIR="$HOME/Documents/"
```

works without any problem.

Found in https://github.com/feenkcom/gtoolkit/issues/1927. @akgrant43 found the exact cause of the problem.

**To Reproduce**

1. Set the environmental variable `XDG_DOCUMENTS_DIR` to `$HOME/` in `~/.config/user-dirs.dirs` on a Linux system.
2. Logout and log in again. This needs to be done for the change to take effect.
3. Run `FileLocator documents.`.

**Expected behavior**

`UnixResolver>>xdgParseUserDirLine:` should return the home directory if `XDG_DOCUMENTS_DIR` is set to `$HOME/`. 

**Screenshots**

No screenshots.

**Version information:**
 - OS: Fedora Linux
 - Version: 33
 - Pharo Version Glamorous Toolkit v0.8.695 (but the problem is in Pharo. See @akgrant43's comment in the linked issue).

**Expected development cost**

I don't know how big change is needed, so I abstain from making an estimate. And if I had enough knowledge to make an estimate, I would probably have fixed myself and created a pull request.

**Additional context**

Not what I know of.

akgrant43:
Currently a line of the form:

XDG_DOCUMENTS_DIR="$HOME/"

in `~/.config/user-dirs.dirs` will answer the root directory, when the user's home directory is obviously intended.

Fixes: pharo-project/pharo#9410

### DIFF
--- a/src/FileSystem-Core/UnixResolver.class.st
+++ b/src/FileSystem-Core/UnixResolver.class.st
@@ -66,7 +66,7 @@ UnixResolver >> xdgParseUserDirLine: aStream [
 	(#($$ $/) includes: firstChar) ifFalse: [ ^ nil ].
 	path := firstChar = $$
 				ifTrue: [ (aStream next: 5) = 'HOME/' ifFalse: [ ^ nil ].
-					       self home / (aStream upTo: $") ]
+					       self home / ((aStream upTo: $") ifEmpty: [ '.' ]) ]
 				ifFalse: [ self resolveString: '/', (aStream upTo: $") ].
 	^ path
 ]
--- a/src/FileSystem-Tests-Core/UnixResolverTest.class.st
+++ b/src/FileSystem-Tests-Core/UnixResolverTest.class.st
@@ -1,0 +1,37 @@
+"
+SUnit tests for PlatformResolver
+"
+Class {
+	#name : #UnixResolverTest,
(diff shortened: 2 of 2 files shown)


akgrant43:
The CI failure looks to be unrelated.

Ducasse:
@tesonep hi pablo can you have a look?

MarcusDenker (approved):
I am not a specialist for that part of the system, but the tests and fix look good to me

akgrant43:
Thanks!

PR #9417, opened 2021-06-07
Method category cleanup - pass 3

astares:
Fix  #9416

PR #9422, opened 2021-06-08
Simplify debugger trap to stop the simulation

dionisiydk:
Simplify a debugger trap to stop a code simulation for specific exception classes. Now we simply handle all Exceptions and ask the instance if it #shouldInterruptExecutionSimulation.

There should be a better general implementation of #shouldInterruptExecutionSimulation based on UnhandledError. But it is for the next time


StevenCostiou:
Nice...
I did not see it either.
However, I see now that my code was wrong as my test to check if Error should interrupt the simulation was discarding all its subclasses, which introduces a bug (before it was `on: Error do: ...`.

I saw that you removed the tests but I would like tests for those things.
It tells which exceptions are intented to interrupt the execution, and if one of them stops doing that the tests will tell us.
It affects the debugger and the general understanding.

@estebanlm we need that fix.

StevenCostiou (approved):

dionisiydk:
> I saw that you removed the tests but I would like tests for those things.
> It tells which exceptions are intented to interrupt the execution, and if one of them stops doing that the tests will tell us.

It was very primitive tests. I don't think they matter especially with current logic (simple true/false constant return). We need real tests simulating StepOver code with interesting exception cases. 
There are already such tests for Halt. I would add them for other cases but I am kind of avoiding making changes to the code to prevent conflict my another PR #8567. I will add new tests but later. As I mentioned #shouldInterruptCodeExecution can be implemented in general way and all these will be there.

StevenCostiou:
I agree the tests are primitive, but tests do not need to only test complex stuff.
In particular, they are important for:
- documenting simply without to look to a complex test,
- make the CI fail if, by mistake, we change the boolean of one of the exception class (or we remove the method by mistake for example)
That is valid even though you cover that in testing the logic of the simulator.

MarcusDenker:
Oh, not merging https://github.com/pharo-project/pharo/pull/9417 created a conflict (but should be easy to fix)

MarcusDenker (approved):

PR #9425, opened 2021-06-09
Method category cleanup - pass 4

### LINKED ISSUE #9424: Method category cleanup - pass 4
- it should be "tests" to categorize test methods, not "test"
- it should be "examples" to categorize example methods, not "example"
and other fixes / alignments (similar to https://github.com/pharo-spec/Spec/issues/1094 in Spec/Spec2 but for the base image)

astares:
Fix #9424


MarcusDenker:
there is now a conflict

PR #9426, opened 2021-06-10
9308 code critics   improving critic description api change api hints architectural

### LINKED ISSUE #9308: CodeCritics - Improving Critic Description (API Change, API Hints, Architectural)
For example, kernel is displayed twice, but it is not known how it differs or it is just a duplicate critic  

imagen

AleCossioCh:
Fix: #9308 
May need refactoring

### DIFF
--- a/src/Renraku/ReSmalllintChecker.class.st
+++ b/src/Renraku/ReSmalllintChecker.class.st
@@ -16,8 +16,9 @@ Class {
 
 { #category : #adding }
 ReSmalllintChecker >> addCritique: aCritique [
-
-	^ critiques add: aCritique
+  (((critiques  collect: [ :x | x sourceAnchor sourceEntity name ]) includes: aCritique sourceAnchor sourceEntity name))
+  ifFalse:[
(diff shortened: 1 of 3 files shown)


AleCossioCh:
Ok, I'm going to check those failed test and fix it!

AleCossioCh:
I fixed testCriticsOf and testToDoOf and now all tests pass.
I checked the rest of the failed tests and they all run perfectly, although the server returns weird messages from files I didn't modify. Suggestions would be great

MarcusDenker:
I wonder if this is the right way to fix it... to me it looks like the rule might be implemented wrongly? It seems to make no real sense in the context of the tool? It seems that it checks rules for packages, but the critique browser does not know how to show that.

to me a better fix (for now) would be to not add any rule to the critique browser that has answered true for #checksPackage

AleCossioCh:
Thanks @MarcusDenker for your suggestion! Could you please review the implementation I did?  
(I also deleted the changes made it before)

PR #9428, opened 2021-06-10
remove old catalog browser

estebanlm:
because it is more confusing than helping :)


Ducasse:
Esteban

Found undeclared Variables:  BaselineOfCatalog in: SystemDependenciesTest>>#testExternalIDEDependencies

estebanlm:
yeah, i know... it was harder than expected (as always)... full of hidden dependencies :(
anyway, we'll see tomorrow :)

estebanlm:
ok, now tests are passing :)

PR #9434, opened 2021-06-11
9430-Playground-Define-Global-lead-to-DNU 

### LINKED ISSUE #9430: Playground: Define Global lead to DNU
COUNT := 0 in playground -> Boum :(

MarcusDenker:
this fixes #9430

I have two tests 

- StPlaygroundPagePresenterTest>>#testDoEvaluateDeclareGlobal
- StObjectContextPresenterTest>>#testDoEvaluateDeclareGlobal

That I will commit to the Newtools repo later after this is merged.

I have added to the backlog:
- remove the possibility to select "add class var" when in the playground
- check #declareInstVar: and #declareClassVar for the case when this is not the playground

All this will be *much* simpler in Pharo10 when we remove this code from the exception.

### DIFF
--- a/src/OpalCompiler-Core/OCUndeclaredVariableWarning.class.st
+++ b/src/OpalCompiler-Core/OCUndeclaredVariableWarning.class.st
@@ -17,9 +17,9 @@ OCUndeclaredVariableWarning >> declareClassVar [
 
 { #category : #correcting }
 OCUndeclaredVariableWarning >> declareGlobal [
-
-	Smalltalk at: node name asSymbol put: nil.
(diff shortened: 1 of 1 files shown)


Ducasse:
super 
would be nice to use self environment instead of this Smalltalk global variable

MarcusDenker:
I am trying to do the absolut minimal changes now that we are in freeze.

MarcusDenker:
This includes to not change that. It should not be self environment, but the environment the compiler compiles for. but that is a change that should not be done now.

Ducasse:
Ok my point is that I'm removing Smalltalk and they pop up like mushrooms.

Issue #9440, opened 2021-06-14
RB - Copy class refactoring needs support of traits and slots


Lin777:
For the moment copy class refactoring only support:

- copy instance variables
- copy class variables
- copy methods

AlexisCnockaert:
indirectly I fixed this problem when fixing `InsertNewClassRefactoring`, now it takes Traits and Slots into account (just tested it!)

PR #9441, opened 2021-06-14
RB - Add variables when we copy a class

### LINKED ISSUE #9335: Copy all package is totally bogus
We should either fix it or remove it.,

Lin777:
Fixes #9335

- Fix copy of variables when we create a copy of a class, however it still doesn't work well when we a class uses traits or has slots.
- For the moment this refactoring's command was deactivated
- Create issue to support traits and slots when we copy a class [(#9440)](https://github.com/pharo-project/pharo/issues/9440)

### DIFF
--- a/src/Refactoring-Core/RBCopyClassRefactoring.class.st
+++ b/src/Refactoring-Core/RBCopyClassRefactoring.class.st
@@ -98,6 +98,21 @@ RBCopyClassRefactoring >> copyMethodsOf: rbClass1 in: rbClass2 [
 	]
 ]
 
+{ #category : #transforming }
+RBCopyClassRefactoring >> copyVariables [
+	aClass instanceVariableNames do: [ :varName |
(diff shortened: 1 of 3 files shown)


Ducasse:
Yes this is a good idea to desactivate it.

Ducasse:
We should have a 
- copy to self contained as a name 
- and keep the old copy package

Lin777:
with this:

> copy to self contained as a name

you mean we make the copies of the classes (inside the package) by their name or their definition?

Lin777:
As far as I review there was no command to copy a package in previous versions.

Ducasse:
To me 
- "copy to self contained" => cop + refer to the copied classes
and I thought that there was an old copy that just copied the classes

Lin777:
I checked the older versions, and I couldn't find any command to copy packages. 

imagen

I will check better, but at the moment I did not find it

PR #9442, opened 2021-06-14
RB - Add push down methods transformation command

### LINKED ISSUE #9407: Push down to only one subclass is a nice operation
Push down a method is pushing the methods to all subclasses and often this is not something we need. 
Of course Push down to Single Subclass is not a refactoring in the same that it can break the program 
but this is a super handy transformation when we need it.

Lin777:
Fixes #9407

- Update push down refactoring to only push down in some classes
- add command of push down methods in specific classes

### DIFF
--- a/src/SystemCommands-MethodCommands/SycPushDownMethodInSomeClassesCommand.class.st
+++ b/src/SystemCommands-MethodCommands/SycPushDownMethodInSomeClassesCommand.class.st
@@ -1,0 +1,68 @@
+Class {
+	#name : #SycPushDownMethodInSomeClassesCommand,
+	#superclass : #SysRefactoringMethodCommand,
+	#instVars : [
+		'classes'
+	],
+	#category : #'SystemCommands-MethodCommands'
+}
+
+{ #category : #activation }
+SycPushDownMethodInSomeClassesCommand class >> methodContextMenuActivation [
+	<classAnnotation>
+	
+	^ SycMethodMenuActivation byRootGroupItemOrder: 21 for: ClyMethod asCalypsoItemContext
+]
+
+{ #category : #converting }
+SycPushDownMethodInSomeClassesCommand >> asRefactorings [
+	"Return push down method refactoring using method"
+	| refactoring |
+	refactoring := RBPushDownMethodRefactoring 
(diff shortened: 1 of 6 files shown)


src/SystemCommands-MethodCommands/SycPushDownMethodInSomeClassesCommand.class.st
```diff
@@ -0,0 +1,68 @@
...
+			withItems: (class methods sort: [ :a :b | a asString < b asString ]) asOrderedCollection
+			selecting: selectedMethods. 
+	dialog cancelled ifTrue: [ CmdCommandAborted signal ].
+	methods := dialog presenter selectedItems.
+	dialog := RBVariablesSelectionPresenter 
+		   label: 'Push down in classes: '
+			withItems: (class subclasses sort: [ :a :b | a name < b name ]) asOrderedCollection
+			selecting: selectedClasses. 
+	dialog cancelled ifTrue: [ CmdCommandAborted signal ].
+	classes := (dialog presenter selectedItems) collect: #name.
```
Ducasse: Can you use collect: [:each | each name]

src/SystemCommands-MethodCommands/SycPushDownMethodInSomeClassesCommand.class.st
```diff
@@ -0,0 +1,68 @@
...
+SycPushDownMethodInSomeClassesCommand >> prepareFullExecutionInContext: aToolContext [
+	super prepareFullExecutionInContext: aToolContext.
+	self selectMethodsAndClasses.
+]
+
+{ #category : #execution }
+SycPushDownMethodInSomeClassesCommand >> selectMethodsAndClasses [
+	|selectedMethods selectedClasses dialog class|
+	class := methods first origin.
+	selectedMethods := methods copy asOrderedCollection.
```
Ducasse: I was wondering why do we need to copy the method

Lin777: My mistake, I already remove the use of unnecessary temporary variables

PR #9455, opened 2021-06-16
Adding missing check for pool contained in superclasses

arturozambrano:
Fix on the writers query of pool vars. 
Now it consider pools being defined in super classes.


Ducasse:
Oh this is cool!

Ducasse:
We should move this behavior to class itself and that ClyClassVariable invoked it because 
this is unclear that we will keep ClyClassVariable in the future and this way it will be easier to write test for it.

MarcusDenker:
There is ClassVariable, too. I wanted to see at some point (not for Pharo9) if we can not replace ClyClassVariable with ClassVariable.

I suggest we do this change now and then improve in Pharo10

Ducasse:
Ok then open an issue to remember.

PR #9487, opened 2021-06-17
Improve failBlock usage in NumberParser

### LINKED ISSUE #9486: NumberParser failBlock is used inconsistently
**Bug description**
`NumberParser` includes an instance variable called `failBlock`. If you look at the references this variable is used inconsistently across the methods. In some methods is evaluated by sending the message `value` as if it was a block without arguments, and in other methods is evaluated by sending `cull:cull:`. So currently, if `failBlock` is configured with a two-argument block and the input given to the parser makes it to use one of the methods sending value it will fail.

**To Reproduce**
Open a Playground and evaluate:

```
|p|
p := NumberParser new.
p on: '-'.
p failBlock: [ :a :b | ].
p nextInteger
```

**Expected behavior**
If two-argument blocks are to be expected (as is suggested by `expected:` method that provides an error description and the index of the stream), then `fail` method needs to be updated so two-argument blocks will not fail.

**Version information:**
 - Pharo Version 9

**Expected development cost**
Easy to fix. I will propose a PR.

gcotelli:
Fixes #9486
Add some test cases

### DIFF
--- a/src/BaselineOfGeneralTests/BaselineOfGeneralTests.class.st
+++ b/src/BaselineOfGeneralTests/BaselineOfGeneralTests.class.st
@@ -13,6 +13,7 @@ BaselineOfGeneralTests >> baseline: spec [
 		spec 
 			package: 'Debugging-Utils-Tests';
 			package: 'Tests';
+			package: 'NumberParser-Tests';
 			package: 'AST-Core-Tests';			
 			package: 'RPackage-Tests';
 			package: 'Monticello-Tests';	"required by MonticelloMocks"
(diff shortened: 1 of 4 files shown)


gcotelli:
The build failures look unrelated (only on Windows): 
```
+ ./pharo bootstrap.image initializePackages --packages=packagesKernel.txt --protocols=protocolsKernel.txt --save
+ ./pharo bootstrap.image loadHermes TraitsV2.hermes --save
./pharo: line 13:    98 Segmentation fault      "$DIR"/"pharo-vm/Pharo.exe" "$image_name" "$@"
```

PR #9492, opened 2021-06-21
RB - Fix extract method refactoring

### LINKED ISSUE #9405: Extract method miss methods
I have the following code

```
visitMonospace: aFormat

	aFormat children do: [ :each | each accept: self ].
	self fixBlockWithoutMicText: aFormat
```

```
visitMonospace: aFormat

	aFormat children do: [ :each | each accept: self ].
	self fixBlockWithoutMicText: aFormat
```

```
visitStrike: aFormat

	aFormat children do: [ :each | each accept: self ].
	self fixBlockWithoutMicText: aFormat
```
and 

```
visitParagraph: aParagraph

	^ aParagraph children do: [:each | each accept: self ]
```

When I ask to extract `aFormat children do: [ :each | each accept: self ]` into visitChildrenOf: 
I'm asked to the check the hierarchy when I say ok, the method is extracted none of the other methods is touched.

Then after if I select aParagraph children do: [:each | each accept: self ]
then the system tells me if I want to use the extracted method. 

S.

Lin777:
Fixes #9405
Fix issue to find occurrences in differents scenarios when we are extracting a method
imagen


Ducasse:
I tried to understand what was the point with the return but I could not.

Lin777:
Sorry for responding late, the eliminate returns part is because there are some cases where the occurrence does not need a return, however the "similar method" has a return, as in the following example:
``` Smalltalk
visitParagraph: aParagraph
    ^ aParagraph children do: [: each | each accept: self]

visitStrike: aFormat
    aFormat children do: [: each | each accept: self].
    self fixBlockWithoutMicText: aFormat
```
because #visitParagraph: has a return when we extract `aFormat children do: [: each | each accept: self]` does not use #visitParagraph: as a pre-existing method, with the removal of returns (in cases where the extraction does not require returns , like this) it looks for occurrences in the methods eliminating the returns of the similar methods to the extraction

PR #9494, opened 2021-06-21
9493-Compiler-error-when-using-whileTrue

### LINKED ISSUE #9493: Compiler error when using whileTrue
**Bug description**

When having the following

```smalltalk
[ [  ​] whileTrue.
   ​^
   | cursor here after the second brackets
```

type a colon (`:`)

And there is an exception

image

Looks like a semantic analysis bug?

MarcusDenker:
If we  have a #whileTrue with a block that has an argument, answer false for #isInlineWhile.

There was a call to #notify:. which makes no sense.
Executing this code will lead to a runtime error, which is what we expect. 

 If we want to give a hint to the programmer, we could add a rule, but it should not be hard-coded in the #isInlineWhile check.

fixes #9493

### DIFF
--- a/src/OpalCompiler-Core/RBMessageNode.extension.st
+++ b/src/OpalCompiler-Core/RBMessageNode.extension.st
@@ -127,7 +127,7 @@ RBMessageNode >> isInlineWhile [
 	self isCascaded ifTrue: [^ false].
 	(#(whileFalse: whileTrue: whileFalse whileTrue) includes: self selector) ifFalse: [^ false].
 	self receiver isBlock ifFalse: [^ false].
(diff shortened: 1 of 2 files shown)


Ducasse:
I have a question: why raising an error is not good? because like that the error could be caught at compile time?

MarcusDenker:
Because it raises an error, e.g. while it is doing semantic analysis after typing a key for the syntax highlighter.

All things that we want to tell the programmer should be done as part of a critique rule, not as part of a "stop the world and wait for input" mechanism. E.g. we want to (have to) remove all the interaction that is now in the "undeclared variable" exception, too.

PR #9511, opened 2021-06-21
AthensBalloonPath>>#paintStrokeUsing:on: sends unimplemented method

### LINKED ISSUE #9510: AthensBalloonPath>>#paintStrokeUsing:on: sends unimplemented method
image

astares:
Fix #9510


MarcusDenker (approved):
as calling the method leads to DNU, removing it is the right thing to do I think

PR #9521, opened 2021-06-24
Ban Lint Time class>>#dateAndTimeFromSeconds: and use #fromSeconds:

### LINKED ISSUE #9520: Ban Lint Time class>>#dateAndTimeFromSeconds: and use #fromSeconds:
Lint issue

image

astares:
Fix #9520


astares:
This simple PR seems to reveal a bug (which I do not know if it is from loading the PR or will be there also when final image is built.


 - The PR is banning a rule on an extension method - so it is adding a manifest 
 - but in the resulting package **some classes are seen as "Uncategorized" after loading the PR**:

image

I guess this is because of Package **Kernel** with **Chronoly** tag and additional **Kernel-Chronolog** package with same name.

PR #9526, opened 2021-06-25
9525-Moving-method-from-one-class-to-another-should-automatically-use-the-original-protocol

### LINKED ISSUE #9516: Upon moving methods to a different class avoid popup requesting protocol
In Pharo 9 when moving a method from a class to another one, the system requests developer to provide the target protocol.
This is repeated for every method if more than one are moved.
This happens regardless whether we drop on the class or on specific protocol.

The default behavior should be:
- use the protocol on which we drop moved methods without further interaction
- request target protocol ONLY ONCE when dropping on a class

### LINKED ISSUE #9525: Moving method from one class to another should automatically use the original protocol
**Bug description**
When you drag&drop a method from one class to another. It asks for the protocol name for every single method you move. I think, by default, it should just use the original protocol name.

MarcusDenker:
This PR removes the #fillProtocols send in SycMoveMethodsToClassCommand

The idea is that we want to just move the method, keeping the category.

fixes #9525
fixes #9516

### DIFF
--- a/src/SystemCommands-MethodCommands/SycMoveMethodsToClassCommand.class.st
+++ b/src/SystemCommands-MethodCommands/SycMoveMethodsToClassCommand.class.st
@@ -41,16 +41,10 @@ SycMoveMethodsToClassCommand >> defaultMenuItemName [
 
 { #category : #execution }
 SycMoveMethodsToClassCommand >> execute [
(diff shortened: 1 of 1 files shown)


Ducasse:
Thanks for this one....

PR #9537, opened 2021-06-25
Add release test - no "utlils" method categorization, use "utilities" instead

### LINKED ISSUE #9536: Add release test - no "utlils" method categorization, use "utilities" instead
we should add a release test to not introduce new additional cases

astares:
Fix #9536


astares:
Note that this is also already supported additionally by an existing lint rule:

image

PR #9539, opened 2021-06-29
9517 modifying method with variable breakpoints in the debugger sliently remove the breakpoint but leaves metalinks behind in other methods

### LINKED ISSUE #9517: Modifying method with variable breakpoints in the debugger sliently remove the breakpoint but leaves metalinks behind in other methods
- Put a variable breakpoint on the variable of a class,
- execute code that touches that variable to open a debugger (it breaks),
- modify the method and save.

Consequences:
- the variable breakpoint is removed (you can't see it anymore in the tools, and the breakpoint object is not in the memory anymore),
- all other methods referencing the variable still have a metalink on them, and they continue to break while they should not if the breakpoint is uninstalled,
- the modified method should continue breaking if it still uses the target variable, but it does not.

Instead, the system should:
- maintain the breakpoint until there are no more use of the target variable,
- no leave active metalinks related to previous breakpoint after they are uninstalled

StevenCostiou:
Fixes #9517 

The bug prevented that we modify methods in the debugger that popped because of a variable breakpoint.
It required a fix in the Reflectivity core, and one in the Reflectivity tools.

The code around seems overly complicated, so a general redesign will be later required for some parts of Reflectivity (basically the object-centric and the permalink part).

### DIFF
--- a/src/Reflectivity-Tests/LinkInstallerTest.class.st
+++ b/src/Reflectivity-Tests/LinkInstallerTest.class.st
@@ -923,7 +923,7 @@ LinkInstallerTest >> testUninstallLinkOnSlotOrVar [
 		
 	"After installing, registry must be correct"	
 	permalink := (linkRegistry permaLinksFor:link) asArray first.
-	self assert: (linkRegistry registeredClassesAt: ReflectivityExamples2) asArray first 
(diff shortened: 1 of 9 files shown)


StevenCostiou:
The failing test seems unelated.

MarcusDenker (approved):
ok

PR #9540, opened 2021-06-29
Modify Array storeString

sonibla:
Hello

This PR introduces extremely few changes so I don't know if it's appropriate

Store strings of Arrays leave a space after the last element, I removed it by changing Array>>#storeOn:
`#(1 2) storeString` is now `'#(1 2)'` instead of `'#(1 2 )'`

This space causes a few test failures on code that is supposed to run on Gemstone/VisualWorks


svenvc:
Looks good to me

Issue #9291, opened 2021-05-19
Undo undoes the second to last change if the last change is new line


pavel-krivanek:
**Bug description**
The **undo** command undoes the second to last change if the last chnage in the editor was pressing Enter to do a new line.

![BA0Xi5JWRy](https://user-images.githubusercontent.com/572494/118778743-b107ad00-b88a-11eb-8110-0ec04938024b.gif)

MarcusDenker:
This might be fixed with merging https://github.com/pharo-project/pharo/pull/9544

MarcusDenker:
I tested it and it seems to work now

PR #9544, opened 2021-07-03
Fixed the Undo and Redo in the RubTextEditor

### LINKED ISSUE #3903: Undo works incorectly in the debugger
And the code is completely unusable and you have to write it again...

rolandbernard:
This should fix at least some issues with the Undo and Redo operations described in #3903. The main problem was that new lines and tab characters are not added to the history. By moving the `openTypeIn` (and also the `addDeleteSelectionUndoRecord`) from being used only for normal keys to being used for all changes. This will then include the new line or tab in the history when `closeTypeIn` is called, the same way it works for normal characters.

Also connected to: #9291

### DIFF
--- a/src/Rubric/RubTextEditor.class.st
+++ b/src/Rubric/RubTextEditor.class.st
@@ -260,6 +260,7 @@ RubTextEditor >> backWord: aKeyboardEvent [
 { #category : #'typing/selecting keys' }
 RubTextEditor >> backspace: aKeyboardEvent [
 	| result startIndex |
+	self closeTypeIn.
 	result := aKeyboardEvent shiftPressed
 		ifTrue: [ self backWord: aKeyboardEvent keyCharacter ]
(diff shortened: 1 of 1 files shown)


MarcusDenker:
I downloaded the build artefact and tested the examples from the issue discussion: 

- Playground works now nicely, even with newline or space
- The debugger works much better, there is still the problem that when you press undo one more than edits where done that the text editor is emptied. But the undos before that now work.
-

rolandbernard:
>   The debugger works much better, there is still the problem that when you press undo one more than edits where done that the text editor is emptied. But the undos before that now work.

OK, I did not test that. This is actually also the current behaviour. (I never noticed)

MarcusDenker:
Yes, I think this is an independent problem, no need to fix it as part of this PR

MarcusDenker (approved):

PR #9556, opened 2021-07-06
9548-Menus-shortcut-rendering-broken

tesonep:
Creating a new font when using shortcut reminder, as it was reusing the same font.


guillep (approved):
Tested locally and it works

PR #9562, opened 2021-07-08
9560-Does-not-understand-while-using-a-non-defined-variable-starting-with-uppercase 

MarcusDenker:
Improve the workaround to turn off variable suggestions for class comments.

(the idea is to do the minimal workaround, a major cleanup is planned for Pharo10)


MarcusDenker:
something went wrong wit he PR, will redo

PR #9563, opened 2021-07-08
Improve the workaround to not show the  undeclared fix menu when in c…

### LINKED ISSUE #9560: Does not understand while using a non-defined variable starting with uppercase
**Bug description**
Open a playground
Type `Foo`
Do-it or Print-it

image

Seen in build 1519

MarcusDenker:
…omment mode.

This PR is a workaround: as many classes can be used a requestor, adding a test method might lead to more problems, therefore the idea
is to detect that case when it is happening in comment mode.

Pharo10 will clean this up and not do the variable suggestion as part of the exception.

fixes #9560

### DIFF
--- a/src/OpalCompiler-Core/OCUndeclaredVariableWarning.class.st
+++ b/src/OpalCompiler-Core/OCUndeclaredVariableWarning.class.st
@@ -149,10 +149,14 @@ OCUndeclaredVariableWarning >> node: aVariableNode [
 
 { #category : #correcting }
 OCUndeclaredVariableWarning >> openMenuIn: aBlock [
-	| alternatives labels actions lines caption choice name interval |
+	| alternatives labels actions lines caption choice name interval requestor |
 	
-	"Turn off suggestions when in RubSmalltalkCommentMode" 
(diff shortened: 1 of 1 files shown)


guillep:
I tested the workaround locally and I did not have an error when typing a global in the playground. I did not know, though, how to test the case that forced the introduction of the original workaround. Was is that typing in the comments pane was broken?

MarcusDenker:
Yes, executing a doit in the comments pane with an Undeclared opened the menu, and if you then selected "define temp" it did not work as the requestor is the text editor directly instead of a model of the browser. (and it does not implement the needed API).

This is all not great (and hard to test!)... the "fix" I did was meant to be the simplest thing we can do now, as anything that fixes it in a cleaner way would be hard to get stable.

I hope that the idea for Pharo10 to not do this inside of the exception will make it easier to test (and less brittle).

guillep:
Ok, because I tested typing in the comment pane just in case and i did not had any errors.
I think this is good enough to integrate.

MarcusDenker:
Thanks!

PR #9572, opened 2021-07-09
9571-Using-Clipboard-Tools-does-not-work-aka-Virtual-Paste

### LINKED ISSUE #9571: Using Clipboard Tools does not work (a.k.a Virtual Paste)
Using a tool to handle the clipboard does not work with P9.
These tools are implemented by sending a virtual paste event. 
This is usually implemented by sending a virtual Ctrl/Cmd + V event.

These events are handled by SDL, but they are handled by sending three events.

We need to correctly detect this sequence and handling the event.

tesonep:
Fix #9571

### DIFF
--- a/src/OSWindow-SDL2/OSSDLPasteEventFilter.class.st
+++ b/src/OSWindow-SDL2/OSSDLPasteEventFilter.class.st
@@ -1,0 +1,155 @@
+Class {
+	#name : #OSSDLPasteEventFilter,
+	#superclass : #Object,
+	#instVars : [
+		'next',
+		'lastEvents',
+		'currentState'
+	],
+	#pools : [
+		'SDL2Constants'
+	],
+	#category : #'OSWindow-SDL2-Base'
+}
+
+{ #category : #'last events' }
+OSSDLPasteEventFilter >> addEvent: aEvent [
+
+	self lastEvents add: aEvent
+]
+
+{ #category : #'events-processing' }
+OSSDLPasteEventFilter >> changeEventsToIncludeModifier [
+
+	| modifierToUse defaultModifier |
+
+	"We modify the KeyUp/KeyDown events to include the correct modifier.
+	This is because tools to handle copy/paste use virtual events, and SDL is dropping the modifier.
+	
+	The correct modifier is calculated depending of the current platform modifier"
+
+	modifierToUse := 0.
+	defaultModifier := OSPlatform current defaultModifier.
+	
+	defaultModifier = KMModifier command ifTrue: [ modifierToUse := KMOD_GUI ].
+	defaultModifier = KMModifier alt ifTrue: [ 	modifierToUse := KMOD_LALT ].
+	defaultModifier = KMModifier ctrl ifTrue: [ modifierToUse := KMOD_LCTRL ].
+
+	self lastEvents do: [ :anEvent | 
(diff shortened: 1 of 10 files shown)


src/OSWindow-SDL2/OSSDLPasteEventFilter.class.st
```diff
@@ -0,0 +1,155 @@
...
+		ifTrue: [ 
+			self addEvent: aSDLEvent.
+			self currentState: 2.
+			 ^ self ].
+
+	(self currentState = 1 and: [aSDLEvent isClipboardUpdateEvent]) 
+		ifTrue: [ 
+			self addEvent: aSDLEvent.
+			self currentState: 2.
+			 ^ self ].
```
guillep: Aren't these two cases the same?

src/OSWindow-SDL2/OSSDLPasteEventFilter.class.st
```diff
@@ -0,0 +1,155 @@
...
+OSSDLPasteEventFilter >> dispatchEvent: aSDLEvent [
+	
+	"This filter will handle the case when an external Clipboard handling tool is used.
+	When an external tool modifies the clipboard and send a virtual paste command, it sends three events:
+	
+	1) SDL_CLIPBOARDUPDATEEVENT
+	2) SDL_KEYDOWN with $v as the key
+	3) SDL KEYUP with $v as the keysym 
+	
+	If we find this three events, we need to modify the last two events to include defaul system modifier.
```
guillep: this three events => these three events

src/OSWindow-SDL2/OSSDLPasteEventFilter.class.st
```diff
@@ -0,0 +1,155 @@
...
+{ #category : #accessing }
+OSSDLPasteEventFilter >> currentState: aValue [
+
+	currentState := aValue
+]
+
+{ #category : #'events-processing' }
+OSSDLPasteEventFilter >> dispatchEvent: aSDLEvent [
+	
+	"This filter will handle the case when an external Clipboard handling tool is used.
```
guillep: I don't know if it is worth to note that this seems to be a OSX SDL2 bug.

src/OSWindow-SDL2/OSSDLPasteEventFilter.class.st
tesonep: Yes, but I write it like this to make it more clear.

src/OSWindow-SDL2/OSSDLPasteEventFilter.class.st
tesonep: Yes, I will add that

src/OSWindow-SDL2/OSSDLPasteEventFilter.class.st
tesonep: Jajja, it is duplicated... I will fix it
 :D

MarcusDenker:
```
Error
the following classes have unused instance variables and should be cleaned: an OrderedCollection(OSSDL2Driver)
Stacktrace
TestFailure
the following classes have unused instance variables and should be cleaned: an OrderedCollection(OSSDL2Driver)
NoUnusedVariablesLeftTest(TestAsserter)>>assert:description:resumable:
NoUnusedVariablesLeftTest(TestAsserter)>>assert:description:
NoUnusedVariablesLeftTest(TestCase)>>performTest
```

PR #9575, opened 2021-07-10
9495-Color-selection-dialog-broken

tesonep:
We cannot reuse the engine, it is not possible to ensure the state of it in the programming model we have.


tesonep:
For Pharo 10, we should reimplement it using Athens directly.

PR #9580, opened 2021-07-12
Fix global declaration in playground

estebanlm:
declaring a global in playground was raising DNU because of some hardcoded code in `OCUndeclaredVariableWarning`. 
I didn't fix for real the issue, but I added more hardcode to prevent an undesirable error just before release :)

NOTE: this code needs to be refactored for real, just no time now :(

### DIFF
--- a/src/OpalCompiler-Core/OCUndeclaredVariableWarning.class.st
+++ b/src/OpalCompiler-Core/OCUndeclaredVariableWarning.class.st
@@ -151,8 +151,13 @@ OCUndeclaredVariableWarning >> node: aVariableNode [
 OCUndeclaredVariableWarning >> openMenuIn: aBlock [
 	| alternatives labels actions lines caption choice name interval |
 	
-	"Turn off suggestions when in RubSmalltalkCommentMode" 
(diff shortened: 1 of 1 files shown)


src/GeneralRules/ReConsistencyCheckRule.class.st
```diff
@@ -32,5 +32,5 @@ ReConsistencyCheckRule >> initialize [
 
 { #category : #accessing }
 ReConsistencyCheckRule >> name [
-	^ 'Uses "size = 0" instead of "isEmpty"'
+	^ 'Checks for empty collection using #size instead of #isEmpty" or #isNotEmpty'
```
estebanlm: I have no idea why this is here :(

MarcusDenker: the PR was against Pharo10 wich does not yet have that change, I rebased it

MarcusDenker:
This should be fixed already with https://github.com/pharo-project/pharo/pull/9563 (I hope?)

estebanlm:
ah, no idea. 
but I can just close if that is the case :)

MarcusDenker:
I will double check

MarcusDenker:
I checked and it is working in the latest version

PR #9581, opened 2021-07-12
Enable display empty slots by default

estebanlm:
I am changing this setting because "first impression counts", and this is what users expect. 
I know this is not the ideal solution, but changing the setting will solve the problem now, while we can fix the more general problem that prevents the expand keybinding to work.


estebanlm:
test passing

estebanlm:
I know I shouldn't be the one merging this, but I need this in for the release :P

PR #9588, opened 2021-07-14
[Compiler] Reify OCSemanticScope's representing the meaning of the code: a method or a doIt expressions

dionisiydk:
It refactors the compiler by reifying OCSemanticScope's representing the meaning of the code: a method or a doIt expressions:
    - OCMethodSemanticScope
    - OCReceiverDoItSemanticScope
    - OCContextualDoItSemanticScope
The configuration parameters of Compiler and CompilerContext #class, #context, #receiver, #noPattern are removed (but accessor methods are kept for compatibility).
The code paths over them are replaced by delegation to a semantic scope.Thus the parsing and compilation methods are slightly simplified without ifNil/ifTrue checks spreaded in many places in the compiler.
In future the existing users of configuration API (class: , receiver: ) needs to be cleaned to be based directly on semantic scopes instead of separate parameters. It will allow many interesting feature.
    
As a side effect of these changes (and it's driver actually) the OCContextualDoItSemanticScope brings a transparent access for temps in doIts expressions in the debugger.
Now the debugIt in the debugger shows real temp vars instead of reflective calls (ThisContext readVariableNamed:):

debuggerWithTemps

The reflective doIt with ThisContext is also kept with new name: #asReflectiveDoItForContext: . It is used by mustBeBoolean machinery. In future it can be simplified with special kind of DoItVariable.


dionisiydk:
Cool! Tests are green from first try.
Only windows build is crashed - unrelated to PR.

dionisiydk:
Broken tests are still unrelated

dionisiydk:
A long time since I saw the green PR here :)

src/AST-Core/RBMethodNode.class.st
```diff
@@ -683,6 +682,14 @@ RBMethodNode >> selectorParts [
 	^ self keywords collect: [:each | each asSymbol]
 ]
 
+{ #category : #'accessing compiled method' }
+RBMethodNode >> semanticScope: aSemanticScope [
```
guillep, line 684: I would prefer to stop crippling compiler stuff in the AST.

src/OpalCompiler-Core/CompilationContext.class.st
```diff
@@ -498,6 +499,16 @@ CompilationContext >> interactive: anObject [
 	interactive := anObject
 ]
 
+{ #category : #accessing }
+CompilationContext >> logDoItEvaluation: aSourceString [ 
+	self logged ifFalse: [ ^self ].
+	
+	Smalltalk globals 
+			at: #SystemAnnouncer
```
guillep, line 466: This hidden dependency looks strange. I know it was there before...

src/OpalCompiler-Core/IRTranslator.class.st
```diff
@@ -92,7 +92,7 @@ IRTranslator >> pushOuterVectors: scope [
 	| scopesWithVector sc |
 	scopesWithVector := OrderedCollection new.
 	sc := scope.
-	[ sc outerScope isBehavior ] whileFalse: [ 
+	[ sc outerScope isNil or: [ sc outerScope isBehavior ] ] whileFalse: [
```
guillep, line 91: Why is there a new isNil check now there?

src/OpalCompiler-Core/RBMethodNode.extension.st
```diff
@@ -73,6 +73,15 @@ RBMethodNode >> generateIR [
 	^ self ir: ir
 ]
 
+{ #category : #'*OpalCompiler-Core' }
+RBMethodNode >> generateMethod [
```
guillep, line 77: I think this change goes also in a bad direction.
I know this is an extension method, but this should be in the compiler, and the AST should not know the `compilationContext`

guillep (changes requested):

src/OpalCompiler-Core/CompilationContext.class.st
dionisiydk: I agree but it was like that before. I just moved it into CompilationContext from OpalCompiler >> logDoIt

src/OpalCompiler-Core/IRTranslator.class.st
dionisiydk: Previously the most outer scope was a class of method. Now the top is semantics scope. 
I was needed the minimal change here to be compatible with both cases to allow this change to be made: 
- during the development I does the testing/debugging with new compiler class (a copy of original one). Then I pushed the changes to the main class.

So the logic here needs to be improved once it will be integrated. And it can be based on a scope API instead of hardcoding the type of object here (something like #isTopScope method or similar).

src/OpalCompiler-Core/RBMethodNode.extension.st
dionisiydk: It's again not a new method. See #generate and #generateWithSource. I just renamed it to be more descriptive.
But generally I agree with you.

src/OpalCompiler-Core/RBMethodNode.extension.st
```diff
@@ -73,6 +73,15 @@ RBMethodNode >> generateIR [
...
+	| methodTrailer |
+	methodTrailer := self compilationContext optionEmbeddSources
+			ifTrue: [ CompiledMethodTrailer new sourceCode: source ]
+			ifFalse: [self compilationContext compiledMethodTrailer ].
+	^self generate: methodTrailer
+]
+
 { #category : #'*OpalCompiler-Core' }
 RBMethodNode >> generateWithSource [
```
dionisiydk, line 86: This can be removed later with deprecation

src/AST-Core/RBMethodNode.class.st
dionisiydk: I don't think I introduced some new dependency. It's a refactoring of what we have now regarding the method vs doIt compilation and semantics analysis. And being related to the semantics analysis it is packaged under compiler package as all other code for semantics analysis.

I agree that the current dependencies are not good. And this PR is my attempt to provide a foundation to fix this situation. 
The current code paths contain many kind of isDoit checks and they are quite difficult to follow. RBSemanticScope is introduced to remove such checks and simplify the logic. But it is still very initial version (while it brings a nice feature for temps access from doIts).

dionisiydk:
somehow I removed all my comments :(

src/AST-Core/RBMethodNode.class.st
dionisiydk: I don't think I introduced some new dependency. It's a refactoring of what we have now regarding the method vs doIt compilation and semantics analysis. And being related to the semantics analysis it is packaged under compiler package as all other code for semantics analysis.

I agree that the current dependencies are not good. And this PR is my attempt to provide a foundation to fix this situation.
The current code paths contain many kind of isDoit checks and they are quite difficult to follow. RBSemanticScope is introduced to remove such checks and simplify the logic. But it is still very initial version (while it brings a nice feature for temps access from doIts).

src/OpalCompiler-Core/CompilationContext.class.st
dionisiydk: I agree but it was like that before. I just moved it into CompilationContext from OpalCompiler >> logDoIt

src/OpalCompiler-Core/IRTranslator.class.st
dionisiydk: Previously the most outer scope was a class of method. Now the top is semantics scope.
I was needed the minimal change here to be compatible with both cases to allow this change to be made:

during the development I does the testing/debugging with new compiler class (a copy of original one). Then I pushed the changes to the main class.
So the logic here needs to be improved once it will be integrated. And it can be based on a scope API instead of hardcoding the type of object here (something like #isTopScope method or similar).

src/OpalCompiler-Core/RBMethodNode.extension.st
dionisiydk: It's again not a new method. See #generate and #generateWithSource. I just renamed it to be more descriptive.
But generally I agree with you.

Ducasse:
So what do we do?
Integrate to support next pass?

guillep:
If @MarcusDenker is ok with the changes, I'd only like if we can do a list of follow ups :)

dionisiydk:
> If @MarcusDenker is ok with the changes, I'd only like if we can do a list of follow ups :)

I don't have a full list. There are too many things to improve there.
My next target is to simplify tools - existing users of compiler. Instead of various variables like doItReceiver, doItContext, isScripting, classOrMetaclass they should keep an instance of concrete semantic scope in the single variable. It will simplify a lot of the code and It will fix some issues related to doItVariables. 

Other ideas I want to try:
- remove "requestor" logic from compiler. It is about handling the compilation errors and I clearly see how simpler it will be with normal error signal/handling. Now #requestor is mixed everywhere in various layers of tooling/compiler.
- simplify the "fix syntax and continue" logic - also related to requestor 
- simplify semantics analysis. I think many bad dependencies in "compilation parties" can be fixed here. Idea is to always parse/compile in faulty mode but raise syntax/semantics errors on the final step of the full process. Now these errors are signaled just in place where they are detected. And it forces to have many information required for them in the lower layers where they are not really needed.

PR #9589, opened 2021-07-14
deprecate glamour and gt tools

estebanlm:
explained in subject :)

(last action of Pharo 9 ;) )


astares:
Maybe it makes sense to also move them to Deprecated90 package already. Makes it easier to remove them 
then for P10.

estebanlm:
nah, too many classes.
we will remove them part by part... having them in their packages easies our work, in fact (because we still need to clean dependencies).

Esteban

PR #9627, opened 2021-07-20
Better IPv6 String Support (Issue 9626)

### LINKED ISSUE #9626: NetNameResolver doesn't deal with IPv6 Strings/ByteArrays
**Bug description**
We have `NetNameResolver class >> #addressFromString:`, which expects IPv4 address strings like we are used to seeing, ie "192.168.0.3" etc.
  
Additionally, we have `NetNameResolver class >> stringFromAddress:` which takes a ByteArray of size 4 also expecting an IPv4 value.
  
This pull request introduces new methods that enable the IPv6 equivalents of the above methods. It also updates those original methods to check to see if a given address string/bytes is IPv6 and, if so, to use the new methods instead.

**To Reproduce**
1. Execute the following in a workspace: `NetNameResolver addressFromString: ''::2001:db8''
2. You will see an error -- but this is a valid IPv6 address (for example)

**Expected behavior**
We expect IPv6 strings, like '::2001:db8', to be compatible with the String->ByteArray and ByteArray->String methods that are already available in the class side of NetNameResolver. These existing methods have generic names, making it seem like they apply to all valid IP addresses, when in fact they are IPv4 only.
  
For example, in the above case we would want `NetNameResolver addressFromString: '::2001:db8'` to give us the following ByteArray: `#[ 0 0 0 0 0 0 0 0 0 0 0 0 32 1 13 184 ]`.
  
We also expect the reverse to work, that `NetNameResolver stringFromAddress: #[ 0 0 0 0 0 0 0 0 0 0 0 0 32 1 13 184 ]` gives us 
`::2001:db8`.
  
Note that the IPv6 string format is more complicated than the IPv4 version, in particular when dealing with "compressing" the longest contiguous range of 0 values into the substring `::`.
**Screenshots**

**Version information:**
 - OS: Pop_os 20.10 x64
 - Version: [e.g. 1.6.1]
 - Pharo Version 10.0.0 (Pharo-10.0.0+build.5.sha.1df15ac7e2e02c8b262b1ad29521855eafad0423 (64 Bit))

**Expected development cost**
I already have a fix that I will submit as a PR shortly.

darth-cheney:
## What ##
This PR implements better IPv6 String/ByteArray support in `NetNameResolver`, as discussed in Issue #9626.
  
## Tests ##
There are tests defined in the `NetworkIPv6StringTest` class. All green.

### DIFF
--- a/src/Network-Kernel/NetNameResolver.class.st
+++ b/src/Network-Kernel/NetNameResolver.class.st
@@ -57,28 +57,125 @@ NetNameResolver class >> addressForName: hostName timeout: secs [
 ]
 
 { #category : #'address string utils' }
-NetNameResolver class >> addressFromString: addressString [ 
+NetNameResolver class >> addressFromIPv6String: aString [
+	"Return a ByteArray corresponding to an IPv6 address in text format"
+	| stream headStream headBytes tailStream tailBytes hextets |
+	headStream := WriteStream on: ByteArray new.
+	tailStream := WriteStream on: ByteArray new.
+	
+	"Obtain a collection of the string hextets from the
+	original IPv6 string. Any instance of a double colon (::)
+	will be present in the list as an empty string."
+	hextets := self splitIPv6HextetsOn: aString.
+	stream := hextets readStream.
+	(stream upTo: '') do: [ :hextet |
+		headStream nextPutAll: (ByteArray readHexFrom: hextet) ].
+	(stream upToEnd) do: [ :hextet |
+		tailStream nextPutAll: (ByteArray readHexFrom: hextet) ].
+	headBytes := headStream contents.
+	tailBytes := tailStream contents.
+	
+	"If the total number of bytes is less than 16, that means
+	we have a compressed range of 0-value hextets specified
+	using the double colon (::). We add these at the end of
+	the headBytes collection"
+	(16 - (headBytes size + tailBytes size)) timesRepeat: [
+		headBytes := headBytes,#[0] ].
+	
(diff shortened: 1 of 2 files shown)


Ducasse:
Thanks for your contribution.

svenvc:
Although I know little about IPv6 address representation, this looks good overall.

Am I right to assume this only deals with this aspect, the conversion, not yet with actually using IPv6 addresses ?

The test methods should not be unclassified.

darth-cheney:
Hi @svenvc,
  
>  Am I right to assume this only deals with this aspect, the conversion, not yet with actually using IPv6 addresses ?
  
That is the correct assumption. Does using IPv6 addresses not work? It seems like this capability is present based on other methods already existing in `NetNameResolver`. 
  
I found [this post from the mailing list](http://forum.world.st/IPv6-support-VM-td4934215.html) and repeated the steps. I got the same result -- there is a primitive error when attempting to connect using an IPv6 address.
  
Perhaps I should post about this to the squeak-vm list since it seems to be the case across the board.  
  
I leave it to the Pharo maintainers to decide whether or not this current PR should be merged without full IPv6 support.

Ducasse:
Thanks for the PR. We will have a look and we want to improve this aspect.

tesonep:
I really like the implementation of the conversion, even if we still does not have IPv6 support I will integrate the PR because it is a good step ahead. Thanks a lot @darth-cheney

Ducasse:
Thanks Pablo! And Darth-Cheney!

PR #9645, opened 2021-07-22
9612-CleanBlockChecker-looks-to-be-used-before-it-is-loaded

### LINKED ISSUE #9612: CleanBlockChecker looks to be used before it is loaded
Hi marcus

Do you have an idea because it is blocking the context cleaning PR?
https://github.com/pharo-project/pharo/pull/9573

```
isClean
	"Answer if the receiver does not close-over any variables other than globals, and does
	 not ^-return (does not close over the home context).  Clean blocks are amenable to
	 being created at compile-time."
	self numCopiedValues > 0 ifTrue:
		[^false].

	^ CleanBlockChecker check: self method
```

MarcusDenker:
this PR fixes #9612 by moving InstructionClient and CleanBlockChecker to Kernel-Methods

### DIFF
--- a/src/Kernel/CleanBlockChecker.class.st
+++ b/src/Kernel/CleanBlockChecker.class.st
@@ -9,7 +9,7 @@ Class {
 	#instVars : [
 		'isClean'
 	],
-	#category : #'Debugging-Core'
+	#category : #'Kernel-Methods'
 }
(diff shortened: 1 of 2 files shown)


Ducasse:
Oh this is nice!

PR #9653, opened 2021-07-22
Cleanup-ClyMethodCreationToolMorph

MarcusDenker:
ClyMethodCreationToolMorph is not used --> as it is internal, we can remove it without deprecation


MarcusDenker:
Ah, no, it is somehow created by magic... but the class has no references.

PR #9655, opened 2021-07-22
ReleaseTest-testNoEquivalentSuperclassMethods

MarcusDenker:
This adds a first version of a new release test: #testNoEquivalentSuperclassMethods

The idea is that we do not want to waste space by having any method in a class where the superclass has already the same method.

We have already a rule that checks it: ReEquivalentSuperclassMethodsRule

This PR adds a ReleaseTest.

- for now (like ReEquivalentSuperclassMethodsRule) we compare the AST, this should be improved as e.g. variable names are now enough to not be the same
- the test adds for now just a backstop: we have 617 cases in the image, it makes sure that it does not get worse

### DIFF
--- a/src/ReleaseTests/ReleaseTest.class.st
+++ b/src/ReleaseTests/ReleaseTest.class.st
@@ -259,6 +259,31 @@ ReleaseTest >> testNoEmptyPackages [
 	self assertEmpty: violating
 ]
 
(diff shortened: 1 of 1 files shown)


MarcusDenker:
as this is just a test I merge it (else it will degenerate even more and this PR will have to get updated again)

PR #9672, opened 2021-07-23
9620-changes-in-pharo-10-in-welcome-window-points-to-version-8-instead-of-10

### LINKED ISSUE #9620: changes in pharo 10 in welcome window points to version 8 instead of 10
We need to do the change https://github.com/pharo-project/pharo/pull/9606  but pointing to the Pharo10 changelog

MarcusDenker:
This PR just updates the end of the method #changLog so we know that the 10 changelog link is already there.

I do not think that we should now do anything else.

fixes #9620


astares (changes requested):
Pharo10ChangeLogs.md is the one for Pharo 1.0

You need to point to Pharo100ChangeLogs.md

astares (approved):

PR #9683, opened 2021-07-23
RB - Add scope manager

Lin777:
Add Scope environments manager


Ducasse:
I did not understand why all the commands are touched but may be this is a calypso stuff. 
I will integrate it so that we can try and fix it.

Lin777:
Hello @Ducasse 

In fact, I changed those methods to extensions so as not to have the message "is Undeclared" when loading them.

PR #9699, opened 2021-07-25
use newtools 0.6.9

estebanlm:
https://github.com/pharo-spec/NewTools/releases/tag/v0.6.9

remove dependencies


astares:
Do we also update for Pharo 10?

estebanlm:
nope
Pharo 10 will use Pharo10 branch (a development branch) :)

PR #9704, opened 2021-07-26
Use Pharo10 branches for Spec and Newtools

estebanlm:
use Pharo10 branch of Spec and NewTools (to keep development branches aligned).
in fact, I made a mistake when I pushed "dev-1.0" :(


estebanlm:
I do not think the failures have something to do with this :)

PR #9707, opened 2021-07-27
Tests to make sure that OpalCompiler uses compilerClass under circums…

dvmason:
…tances where methods are automatically recompiled.

This commit is just the tests, the next commit may propose a fix.


Ducasse:
This is super cool! Thanks a lot.

PR #9724, opened 2021-08-02
RB - Migrate Deprecate method refactoring

Lin777:
Migrate deprecate method refactoring from RB1 to RB2


Ducasse:
This test is also broken

```
unix-64 / Tests-unix-64 / testAPI – Unix64.Refactoring2.Transformations.Tests.RBTransformationTest
```

PR #9736, opened 2021-08-05
Added #combinations: method

jordanmontt:
Added the `#combinations: aNumber` method that return the combination taking only `aNumber`. This is useful when you want to obtain only the combination for a `k` number as shown in the formula: 

equation


astares:
Why not provide a unit test too?

jordanmontt:
Yes you are right. I will put the test in the same class (`ArrayTest`) as the test `testCombinations`

Issue #9744, opened 2021-08-06
Remove Glamour


Ducasse:
Due time arrived: Glamour out.

noha:
🥇 

- __

Ducasse:
:)

PR #9745, opened 2021-08-06
Fixes: #9744. Remove Glamour! Finally.

Ducasse:


Ducasse:
I will merge and we fix the broken tests.

PR #9767, opened 2021-08-11
Removed SpChooserPresenter from Pharo

jordanmontt:
Because now the `SpChooserPresenter` is in spec repo, not in Pharo. See: https://github.com/pharo-spec/Spec/pull/1133


Ducasse:
Just a question SpChooserPresenter is not used by RB in Pharo?

jordanmontt:
@Ducasse it's used only by `RBEnvironmentVisitor `. But I have already talked with the bolivian team (Semantics, Juan Pablo...) and they are going to fix it by using the new Chooser in Spec rather than this one. They are going to do it right away as soon as we accept this PR.

Ducasse:
Ok but it means that I should integrate the package of Spec. 
Normally as soon as the package of spec is integrate we will just have a dirty package in RB. 
Did you change the name of the component?

jordanmontt:
No, I haven't changed the name of the component.

Issue #9703, opened 2021-07-26
Longterm Cleanup: No duplicated Method in the hierarchy


MarcusDenker:
**Describe the problem**
We should check all methods that are just the same as a method that exists in the hierarchy already.

- There is a rule that detect it
- A release test is there: testNoEquivalentSuperclassMethods, it, for now, tests for the number of cases we have in the system

**What needs to be done**
We need to remove the methods. But: this sounds trivial, but in practice, there will be more complex cases (e.g. some FFI related methods seem to be auto-generated, some #subclassResponsibility overrides the need to be replaced by tagging the class as Abstract, some cases might be there for tests...).

If you do a PR, please mention this issue tracker entry so we keep track on what has been already done.

**How do we find them?**

Inspect this code:

```
methods := SystemNavigation new allMethods reject: [:method | 
	method isFromTrait or: [method methodClass inheritsFrom: FFIExternalStructure ]].
methods := methods reject: [:method | method methodClass == ExternalAddress].

methods := methods select: [:method |
method methodClass superclass
 		ifNotNil: [ :superclass | (superclass lookupSelector: method selector)
 			ifNotNil: [ :overridenMethod | method ast = overridenMethod ast ]
 			ifNil: [ false ] ]
 		ifNil: [ false ]].
methods
```

(you might see that we test here for AST equality... yes, this means that we for now do not find cases where variable names
are different. But as we have already >600 cases, this is a good first step).

The idea is to fix this step by step. This can be nice for new contributors: just fixing some cases (e.g all found in one package) makes a very nice contribution for a first-timer.

MarcusDenker:
This PR added a release test that serves as a back-stop: https://github.com/pharo-project/pharo/pull/9655
(It has to be updated while we clean up and, when finished, not just raise and error but tell what failed, see the commented code at the end)

MarcusDenker:
all easy cases done...

PR #9789, opened 2021-08-20
Reduce-duplicated-Methods-2

MarcusDenker:
This PR reduces the methods that are eqally defined in the superclass (see issue #9703)

- Update the test to check for 474 instead of 620


Ducasse:
This is strange because we fixed many tests with Pablo and I see them still broken. 
I will integrate this fix. But we should check why the dependencies are still failing.

MarcusDenker:
the PR was checked before the test fixed where merged

PR #9793, opened 2021-08-23
ReadOnlyLiterals-Remove-Preference

MarcusDenker:
In Pharo9 we changed the compiler to compile all literals (that is, where #isLiteral returns true) as read-only.

For testing, we added a compiler preference. Now that this is the default, we can remove the preference as it never makes sense to turn it off

(and even if someone would need to, they can just post-process the literals and send #beWritableObject themselves)


MarcusDenker:
failing tests not related (I opened issue tracker entries for all failing tests)

dionisiydk (approved):

PR #9794, opened 2021-08-23
Remove-Package-Deprecated90

MarcusDenker:
This PR removes Deprecated90
(we need to remove the code from the git repo later, too)


MarcusDenker:
ByteString>> #convertToEncoding: is using class TextConverter which is Deprecated and thus not there anymore

MarcusDenker:
This needs multiple steps, I have added an issue for the TextConverter

PR #9799, opened 2021-08-23
Cleaning up `ValueLink` objects

massimo-nocentini:
This PR primarily fixes the creation of `Link` objects as happen in the message `Object>>#~~>`. 

The current implementation sends `#basicNew` then sets some properties and, finally, returns the newly created link without a send of `#initialize`. The proposed change dispatches over the argument of the binary message in order to maintain the class of the links the sender is pushing on. Moreover, also the following points changes:
- A refinement of the test case about `#add:` for `LinkedList` objects when a sender wants to add `nil` to it.
- A new test class for `ValueLink` objects and two test cases:
  - `#testPushingOnValueLink` that records the subject of this PR as explained above;
  - `#testNilAsLink` that shows how `nil` is related to links creation.

### DIFF
--- a/src/Collections-Support-Tests/ValueLinkForTesting.class.st
+++ b/src/Collections-Support-Tests/ValueLinkForTesting.class.st
@@ -1,5 +1,10 @@
 "
 I am a variant of `ValueLink` to show the behavior of `Object>>#~~>`.
+
+In particular, I will be used to test expressions of the form `a ~~> b` 
+that create a link (with `a` as value) of the same class of the link `b`, 
+so I don't introduce any new methods but provide myself to check that newly 
+created links keep the class of the chain we are pushing on.
 "
 Class {
 	#name : #ValueLinkForTesting,
--- a/src/Collections-Support-Tests/ValueLinkForTesting.class.st
+++ b/src/Collections-Support-Tests/ValueLinkForTesting.class.st
@@ -1,0 +1,8 @@
+"
+I am a variant of `ValueLink` to show the behavior of `Object>>#~~>`.
+"
+Class {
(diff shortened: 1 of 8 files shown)


pavel-krivanek (approved):

src/Collections-Support-Tests/ValueLinkForTesting.class.st
```diff
@@ -0,0 +1,8 @@
+"
+I am a variant of `ValueLink` to show the behavior of `Object>>#~~>`.
+"
```
dionisiydk, line 8: Why do you need a new subclass without any new methods?

massimo-nocentini: The new subclass has been introduced to show that the expression `a ~~> b` creates a link (with `a` as value) of the same class of the link `b`, so no additional methods are necessary to test it. Of course, in real usage those subclasses would hold new methods.

MarcusDenker:
failing tests:

- ReleaseTest>>testNoShadowedVariablesInMethods
- CommentTestCase>>testIt

src/Collections-Support-Tests/ValueLinkForTesting.class.st
Ducasse: Could you add this to the class comment? Because a subclass without no new method IS strange

massimo-nocentini: Added in 464c813

MarcusDenker:
24 tests have failed, but most of them due to  #9818. Review is on hold till that is fixed

PR #9815, opened 2021-08-24
RecurseReadOnly-LiteralArray

MarcusDenker:
Since Pharo9. literals are immutable.
But we did not recurse into the Literal Arrays.

This PR:
- add #beReadOnlyLiteral and test
- use it in the compiler

### DIFF
--- a/src/Collections-Sequenceable-Tests/ArrayTest.class.st
+++ b/src/Collections-Sequenceable-Tests/ArrayTest.class.st
@@ -603,6 +603,13 @@ ArrayTest >> testAtWrap2 [
 ]
 
 { #category : #tests }
+ArrayTest >> testBeReadOnlyLiteral [
+	self assert: (Array new) beReadOnlyLiteral isReadOnlyObject.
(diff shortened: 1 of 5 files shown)


MarcusDenker:
Some tests time out, they are not related. Green on Linux.

src/Collections-Sequenceable-Tests/ArrayTest.class.st
```diff
@@ -602,6 +602,13 @@ ArrayTest >> testAtWrap2 [
 	self assert: (tabTest atWrap: -2) equals: 5
 ]
 
+{ #category : #tests }
+ArrayTest >> testBeReadOnlyLiteral [
+	self assert: (Array new) beReadOnlyLiteral isReadOnlyObject.
```
Ducasse, line 607: No need parenthese over Array new, String new....

MarcusDenker: I will fix it with the next iteration (which is to improve the sharing code to take the nested arrays into account)

PR #9816, opened 2021-08-24
DeadCode-Cleanup-08-24

MarcusDenker:
- remove Deprecated Compiler classes/methods to make it easier to show the compiler code
- remove more duplicated methods (for #9703)


Ducasse:
"Cut the fat" is the hidden name for Pharo 10 and I love it.

PR #9818, opened 2021-08-24
Loading documentation reader.

Ducasse:


MarcusDenker:
There where lots of failing tests... e.g. due to halt in the code.

This now means that all the work to get the CI green was for nothing

MarcusDenker:
These are:

ProperMethodCategorizationTest>>testNoUtilsMethods
ProperMethodCategorizationTest>>testSetUpMethodInSUnitTestsNeedsToBeInRunningProtocol
ReleaseTest>>testMethodsContainNoHalt
ProperMethodCategorizationTest>>testNoUncategorizedMethods
ReleaseTest>>testUndeclared
ReleaseTest>>testNoEquivalentSuperclassMethods


I will add an issue tracker entry for each of them.

MarcusDenker:
But this is getting boring

Ducasse:
Do not worry we will fix them.
The problem is that the critic browser sucks!

astares:
@Ducasse yes it sucks. But we need a new one anyway due to Spec2, no?

Best would be to have immediate feedback directly in the system browser (instead of a separate tool) as in other modern IDE's like Eclipse. There is a yellow/red exclamation mark on packages and classes icons if the rules are violated:

image

and a **Problems** tab (not a pane as in calypso) showing errors and warnings:

image

PR #9819, opened 2021-08-24
Unify #layoutForType: with the #kind methods.

MarcusDenker:
- make sure to use the right symbols in #kind, implement a fall back on ObjectLayout that returns the name of the layout
- instead of hard-Coding the symbols in #layoutForType:, iterate over the subclasses instead.

This is all just internal to Monticello, maybe we should rename #kind to #mcType (as it is in Ring2) later to make this clearer.


MarcusDenker:
Now 28 tests fail due merging https://github.com/pharo-project/pharo/pull/9818

MarcusDenker:
I sadly have to stop doing any work until this is fixed. I will now not go through 28 failing tests to check if they fail due to my change or #9818

MarcusDenker:
Failing tests are related

PR #9820, opened 2021-08-24
Ring2: simplify #mcType and unify with #kind

MarcusDenker:
#mcType should have a fallback on  RGLayout returning the layoutName.

This makes it extensible and allows to remove the hardcoded cases for #DoubleByteLayout and  #DoubleWordLayout
(and follows the implementation of #kind on the layout classes, see #9819)

### DIFF
--- a/src/Ring-Monticello/RGDoubleByteLayout.extension.st
+++ b/src/Ring-Monticello/RGDoubleByteLayout.extension.st
@@ -1,7 +1,0 @@
-Extension { #name : #RGDoubleByteLayout }
-
-{ #category : #'*Ring-Monticello' }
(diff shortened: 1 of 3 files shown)


MarcusDenker:
Lots of failing test, most likely due to #9818, but someone needs to check them one by one to be sure

Ducasse:
To test new version of HelpBrowser

Ducasse:
Most of the tests are fixed in the latest version but they are not loaded and we do not get why.

Ducasse:
Since a build takes 1 hour it limits the possibility to try and kills energy for stupid things that in addition are already tested.

PR #9821, opened 2021-08-24
Adding doc

Ducasse:


MarcusDenker:
Lots of failing test, most likely due to #9818, but someone needs to check them one by one to be sure

PR #9823, opened 2021-08-25
Cleanup-Deprecated-Metacello

MarcusDenker:
remove deprecated methods from Metacello (and one from DrTests)


MarcusDenker:
tests have failed, but most of them due to  #9818. Review is on hold till that is fixed

PR #9824, opened 2021-08-25
Deprecate-DebuggerMethodMapOpal

MarcusDenker:
After having all the methods in DebuggerMethodMapOpal deprecated in Pharo9, we can now continue the cleanup

- remove all these deprecated methods
- move the class DebuggerMethodMapOpal to Deprecated10
- deprecate and move all the methods referencing it


MarcusDenker:
tests have failed, but most of them due to #9818. Review is on hold till that is fixed

PR #9826, opened 2021-08-25
Exposing the ABI of a function to the image.

tesonep:
The image can pass the ABI type so LibFFI uses the correct one


MarcusDenker:
tests will have failed, but most of them due to #9818. Review is on hold till that is fixed

MarcusDenker:
26 tests are failing

PR #9827, opened 2021-08-25
Adding initial iteration of ExternalProjectLoader

tesonep:
Adding a simple loader to expose ExternalProjects to be loaded.
We have a nice UI in NewTools, and the idea is that we can start to script them and move them away from the image repository.


MarcusDenker:
Just some timeouts of the Win slave

PR #9834, opened 2021-08-26
Expand Dictionary documentation

ctSkennerton:
- Expand Dictionary class comment
    - Convert to markdown format
    - Add in examples for main messages
- Expand comments for dictionary selectors.
- Add in comments for some selectors


Ducasse:
Excellent idea.
Could you change in the comment `Class >>> foo` as `Class >> #foo`

ctSkennerton:
thanks for your comment, I updated it

PR #9836, opened 2021-08-26
LiteralSharing-NestedArrays

MarcusDenker:
This PR improves literal sharing (note: not yet active)

- take into account the content of nested arrays when building the table
- unify the nested arrays to use the one version from the table
- formatting test fix

### DIFF
--- a/src/Collections-Sequenceable/Array.class.st
+++ b/src/Collections-Sequenceable/Array.class.st
@@ -314,6 +314,16 @@ Array >> printOn: aStream [
 	super printOn: aStream
 ]
 
+{ #category : #enumerating }
+Array >> recursiveDo: aBlock [
+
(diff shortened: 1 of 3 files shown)


src/Collections-Sequenceable/Array.class.st
```diff
@@ -314,6 +314,16 @@ Array >> printOn: aStream [
 	super printOn: aStream
 ]
 
+{ #category : #enumerating }
+Array >> recursiveDo: aBlock [
```
dionisiydk, line 318: Why not move this method to the base class Collection ?

dionisiydk (approved):

src/Collections-Sequenceable/Array.class.st
MarcusDenker: yes, good idea as a next step

PR #9838, opened 2021-08-26
Removing old Event handling and VM World Renderer

tesonep:
WIth the new SDL based UI the old event backend is not needed anymore


Ducasse:
Super cool!
Yes another piece of unused code is out! Excellent

PR #9845, opened 2021-08-27
9798 deprecated class text converter is still used

axlgithub:


src/Compression/CompiledMethodTrailer.extension.st
```diff
@@ -31,13 +31,12 @@ CompiledMethodTrailer >> encodeEmbeddedSourceZip [
...
+	| stream length encodedLength |
 
 	[data isString] assert.
-	utf8str := data convertToEncoding: 'utf8'.
 	
-	stream := ((ZipWriteStream on: (ByteArray new: utf8str size))
-		nextPutAll: utf8str asByteArray;
+	stream := ((ZipWriteStream on: (ByteArray new: data size))
+		nextPutAll: data asByteArray;
 		close;
```
guillep, line 40: There is no encoding in the new version. Shouldn't the asByteArray be replaced by utf8Encoded?

src/Deprecated90/String.extension.st
```diff
@@ -2,6 +2,11 @@ Extension { #name : #String }
 
 { #category : #'*Deprecated90' }
 String >> convertFromEncoding: encodingName [
+	self 
+        deprecated: 'Use #utf8Decoded or #decodeWith: encoding on a byte array. Using a string is not valid. You should have a byte array as input'
+        on: '2021-08-27'
+        in: #Pharo10.
+
 	^self convertFromWithConverter: (TextConverter newForEncoding: encodingName)
```
guillep, line 10: We should replace these implementations to be able to remove the TextConverter class

src/Deprecated90/String.extension.st
```diff
@@ -22,6 +27,11 @@ String >> convertFromWithConverter: converter [
 
 { #category : #'*Deprecated90' }
 String >> convertToEncoding: encodingName [
+	self 
+        deprecated: 'Use #utf8Encoded or #encodeWith: .'
+        on: '2021-08-27'
+        in: #Pharo10.
+
 	^self convertToWithConverter: (TextConverter newForEncoding: encodingName).
 ]
```
guillep, line 36: Same as above

src/Monticello/MCVersionInfoWriter.class.st
```diff
@@ -25,7 +25,7 @@ MCVersionInfoWriter >> writeVersionInfo: aVersionInfo [
 	#(name message id date time author) do: [:sel | 
 		stream 
 			nextPutAll: sel; space;
-			print: (((aVersionInfo perform: sel) ifNil: ['']) asString convertToEncoding: 'latin-1' ); space ].
+			print: (((aVersionInfo perform: sel) ifNil: ['']) asString encodeWith: #latin1 ); space ].
```
guillep, line 28: This used to return a string and is now a ByteArray, is that ok?

guillep (changes requested):

src/Deprecated90/String.extension.st
guillep: This should be the most important point in the PR :)

guillep:
Superseeded by https://github.com/pharo-project/pharo/pull/10012

PR #9849, opened 2021-08-30
9848-Allow-reflective-execution-if-CompiledMethod-without-sending-to-the-receiver 

MarcusDenker:
implements #valueWithReceiver:arguments: as described in the issue tracker entry:

-  add receiver:withArguments:executeMethod: that uses primitive 188
- use it in #valueWithReceiver:arguments:

this is tested in CompiledMethodTest>>#testValueWithReceiverArguments (via #valueWithReceiver:arguments:)

### DIFF
--- a/src/Kernel/CompiledMethod.class.st
+++ b/src/Kernel/CompiledMethod.class.st
@@ -960,6 +960,12 @@ CompiledMethod >> readsField: varIndex [
 	^ super readsField: varIndex 
 ]
 
+{ #category : #evaluating }
+CompiledMethod >> receiver: aReceiver withArguments: anArray executeMethod: anObject [
+	<primitive: 188>
+	self primitiveFailed
+]
+
 { #category : #literals }
 CompiledMethod >> refersToLiteral: aLiteral [
 	"Answer true if any literal in this method is literal, even if embedded in array structure."
@@ -1161,7 +1167,7 @@ CompiledMethod >> usesUndeclareds [
 
 { #category : #evaluating }
 CompiledMethod >> valueWithReceiver: aReceiver arguments: anArray [ 
-	^ aReceiver withArgs: anArray executeMethod: self
+	^self receiver: aReceiver withArguments: anArray executeMethod: self
 ]
 
 { #category : #copying }


src/Kernel/CompiledMethod.class.st
```diff
@@ -960,6 +960,12 @@ CompiledMethod >> readsField: varIndex [
 	^ super readsField: varIndex 
 ]
 
+{ #category : #evaluating }
+CompiledMethod >> receiver: aReceiver withArguments: anArray executeMethod: anObject [
+	<primitive: 188>
```
dionisiydk, line 965: This method looks quite strange if defined in this class. You can't reuse it anywhere without actual method instance.

src/Kernel/CompiledMethod.class.st
```diff
@@ -1161,7 +1167,7 @@ CompiledMethod >> usesUndeclareds [
 
 { #category : #evaluating }
 CompiledMethod >> valueWithReceiver: aReceiver arguments: anArray [ 
-	^ aReceiver withArgs: anArray executeMethod: self
+	^self receiver: aReceiver withArguments: anArray executeMethod: self
```
dionisiydk, line 1170: What about using the MirrorPrimitive class here?
```suggestion
	^MirrorPrimitives withReceiver: aReceiver andArguments: anArray execute: self
```

dionisiydk (changes requested):

src/Kernel/CompiledMethod.class.st
MarcusDenker: the problem is that this kernel functionality, but MirrorPrimitives is not in the Kernel.

i wanted to do a pass over MirrorPrimitives and use them everywhere, but that should be done as a project in itself.

src/Kernel/CompiledMethod.class.st
MarcusDenker: yes, it is a bit of a hack, like the mirror primitive methods on Context.

dionisiydk (approved):

MarcusDenker:
I propose to merge this and to open a new issue about using MirrorPrimitives

dionisiydk:
Sure Marcus. That's why I approved

PR #9858, opened 2021-08-31
Improve ProperClassCategorizationTest>>#testProperClassTagCasing

### LINKED ISSUE #9856: Improve ProperClassCategorizationTest>>#testProperClassTagCasing
Better use  

```
Smalltalk packages
```
as it is shorter and returns the packages already sorted (and also use better var name in block iteration)

astares:
Fix #9856

### DIFF
--- a/src/ReleaseTests/ProperClassCategorizationTest.class.st
+++ b/src/ReleaseTests/ProperClassCategorizationTest.class.st
@@ -11,7 +11,7 @@ Class {
 ProperClassCategorizationTest >> testProperClassTagCasing [
 	|violations|
 	violations := OrderedCollection new.
-	Smalltalk packages
+	RPackageOrganizer default packages
         do: [:package | package classTags do: [:classTag | classTag name first isLowercase ifTrue: [ violations add: (package -> classTag) ] ] ].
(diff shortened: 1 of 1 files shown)


MarcusDenker:
I think the reason that there is no Smalltalk reference is the idea to reduce references to globals... but then, it is true that "RPackageOrganizer default packages" is not really beautiful, either.

I sometimes think we need an explicit and easy way to send messages to "the image" that does not involve Globals, nor a Singleton pattern (which in the end not that much better...)

astares:
OK, how to proceed?

Introduce a 

```Smalltalk
RPackageOrganizer defaultPackages
```
and use that? But it would be a reference to a global too, too a global class name.

Ducasse:
Yes for example. 
We should not put it in Smalltalk :).

MarcusDenker:
I really think we need to do an analysis of the Smalltalk references... today it is just clear "do not use it", but the alternatives are in some cases very ugly or even impossible, and in no case documented. (e.g. we need a way to send a message to "the system" without going via a class, e.g. to aks for the default compiler)

MarcusDenker (approved):

PR #9863, opened 2021-08-31
Cleanup-Duplicate-methods-30-08 

MarcusDenker:
This PR removes some methods that are just copies of the same method in the superclass

another step for https://github.com/pharo-project/pharo/issues/9703

### DIFF
--- a/src/Fuel-Platform-Pharo-Core/FLPharoPlatform.class.st
+++ b/src/Fuel-Platform-Pharo-Core/FLPharoPlatform.class.st
@@ -39,13 +39,13 @@ FLPharoPlatform >> absolutePathOf: aString [
 	^ aString asFileReference fullName
 ]
 
-{ #category : #'accessing - compiler' }
+{ #category : #'accessing-compiler' }
 FLPharoPlatform >> addAndClassifySelector: aSymbol inClass: aClass withMethod: aCompiledMethod inProtocol: aString [
 
 	aClass addAndClassifySelector: aSymbol withMethod: aCompiledMethod inProtocol: aString
 ]
 
-{ #category : #'accessing - system' }
+{ #category : #'accessing-system' }
 FLPharoPlatform >> imageName [
 	^ Smalltalk image shortImageName
 ]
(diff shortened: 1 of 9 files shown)


src/Fuel-Platform-Pharo-Core/FLPharoPlatform.class.st
```diff
@@ -39,13 +39,13 @@ FLPharoPlatform >> absolutePathOf: aString [
 	^ aString asFileReference fullName
 ]
 
-{ #category : #'accessing - compiler' }
+{ #category : #'accessing-compiler' }
```
astares, line 42: Please use "accessing - compiler" and not "accessing-compiler". Otherwise we have shadowed categories.

astares (changes requested):
- PR unfortunately changes categories wrongly (and creates new pairs of categories in category dialog)
- also I thought that Fuel is managed externally in theseions repository and loaded. No?

src/Fuel-Platform-Pharo-Core/FLPharoPlatform.class.st
MarcusDenker: But I wonder how this happens... there seems to be different code in the image than in the repository. I just commited, no changes done by me on the level of the categories

MarcusDenker:
The problem is that even I do not know which packages are managed externally. There is no way to discover that.

src/Fuel-Platform-Pharo-Core/FLPharoPlatform.class.st
astares: Bad - I had similar trouble yesterday evening with the pull request for Microdown - Iceberg showed me more than I have actually changed (like Baseline or other). 

What I did: I downloaded freshly the pharo repo from github and also used a fresh vm +image from zeroconf

astares:
Yes - this was never sorted out / discussed properly with a decision (or documentation)

MarcusDenker:
I close this... easier to redo than to fix the PR

PR #9895, opened 2021-09-03
Release test for #new categorization

### LINKED ISSUE #9894: Release test for #new categorization
The #new method on class side should be categorized in 'instance creation'.

Actually this is already OK except 4 cases that will be fixed with PR 9893 and PR 9891.

So we should add a release test to check this. It then will actually be green when 
- https://github.com/pharo-project/pharo/pull/9893
- https://github.com/pharo-project/pharo/pull/9891
are merged.

astares:
Fix #9894

Side note: this test can only become green when PR 9893 and PR 9891 have been merged first.

### DIFF
--- a/src/ReleaseTests/ProperMethodCategorizationTest.class.st
+++ b/src/ReleaseTests/ProperMethodCategorizationTest.class.st
@@ -105,6 +105,14 @@ ProperMethodCategorizationTest >> testHashMethodNeedsToBeInComparingProtocol [
 
 ]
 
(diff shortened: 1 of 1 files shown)


astares:
PR 9893 and PR 9891 have been merged. So this could be reviewed/merged too now.

Ducasse:
We discussed and we prefer not having this rule.

Ducasse:
We think that it is too strict.

PR #9897, opened 2021-09-03
Release test for #basicNew categorization

### LINKED ISSUE #9896: Release test for #basicNew categorization
The #basicNew method on class side should be categorized in 'instance creation' which is already
true for the standard image.

We should add a release test to ensure that we keep this quality level.

astares:
Fix #9896

### DIFF
--- a/src/ReleaseTests/ProperMethodCategorizationTest.class.st
+++ b/src/ReleaseTests/ProperMethodCategorizationTest.class.st
@@ -49,6 +49,14 @@ ProperMethodCategorizationTest >> explanationFrom: aCol [
 					separatedBy: [ str nextPutAll: ' ,'] ]).
 ]
 
+{ #category : #'tests - instance creation' }
+ProperMethodCategorizationTest >> testBasicNewMethodNeedsToBeInInstanceCreationProtocol [
+	"The #basicNew methods should be in method protocol 'instance creation'"
+	
+	self assureAll: #basicNew areCategorizedIn: #'instance creation' whenSubclassOf: Object class
+
+]
+
 { #category : #'tests - object' }
 ProperMethodCategorizationTest >> testCloneMethodNeedsToBeInCopyingProtocol [
 	"The #clone methods should be in method protocol 'copying'"


MarcusDenker:
I like to test this: For the core system, I want to not accept code that degrades (there should be no reason to ever commit to the Kernel code that has this categorized differently).

But we have code in the system where things are not that strict... and we had feedback where a committer got very annoyed as when something (not core) is in active development, fixing things like this might not be a priority.

Would it make sense to restrict these tests to those parts of the system that are more "Core"?

astares:
Does not make sense to distinguish into Core because then we open the discussion on the boundaries (Kernel or Kernel + primary projects vs. external loaded projects).

Do we want a clean **release image** with clear rules or not?

To me it makes sense as cleaning has shown already a much more clean, consistent and structured release. If that is not wanted I can stop.

MarcusDenker:
I find tests like that very (very!) important. e.g. just today the existing tests catched two problems

See https://github.com/pharo-project/pharo/issues/9996 and https://github.com/pharo-project/pharo/issues/9995

If we would not have release tests, this code would stay in the image forever as nobody would know that code was merged that added a #halt and an unused ivar.

astares:
>where a committer got very annoyed

So we stop because a single individual got annoyed? Do we really want to boil it down to this reason?

Maybe we have a different understanding but my reality is like this:

We still only have a few commiters and a common interest should be to 
 - increase the number of commiters / contributors (and not only on the size of users) to bring things to more shoulders
 - increase the number of features / code (which will grow over time anyway)
 - increase the quality of the code (and yes doing this equally and more strict fosters this)
 - increase the trust into Pharo
 - decrease the complexity
 
To be able to do this we need rules - for community, processed and for code. Otherwise it is 
 - hard to follow what is happening
 - hard to understand code and processes
 - hard to contribute
 - hard to find things easily on expected locations (like an expected category "instance creation" when it comes to
   methods like #basicNew who exactly do what the category describes)

otherwise Pharo and community and users will not grow or just get stuck on a certain level because only the 
"wizards" know how to do things or "witchery" is needed all the time of the few who are knowledable why things
are done different each time.

Pharo was forked in 2008 and unfortunately we made several years no progress on the quality side - even our own code 
was still bad, uncategorized, not properly done or documented, even same things have been done differently each time. 
Silly test code was creeping in where even a single #setUp method was all the time in a different category or not calling super 
setUp and so on and so on. 

I started this whole quality and cleanup initiative to change exactly that and many hours have been invested from many sides
even in very trivial stuff but that sums up into what we have now: better, more aligned packages, more easy to understand
and navigate because they follow certain patterns and rules ensure that we keep the status quo.

And we are by no means finished. Missing definitions enforced by rules still lead to trouble. We have still tough open topics like it is even for you hard to distinguish which packages are managed externally and which ones internally.

A release like Pharo 9 is basically a snapshot of development state. Every day is a "relase day" - and thats why we test with **continuous integration** (CI) to keep (and ideally) grow in the quality level for each single release build.

And yes while it is alpha things might easily break when we have bigger changes - but that does not mean that we should not return quickly in case we get into a bad state. There is a reason why we monitor and do tests or check via CI as we want to constantly improve and not fall back!

I understand that one might require more time to bring code into better shape but then git easily allows for that: one can have a branch, resynch with current state easily and manage things outside of image until they are stable enough for integration.

When it would be required to already have things integrated or showed to others then one can integrate already in smaller pieces (but with same level of quality and following the rules) to already to see if things collide or better integrate with other packages also in the move.

But being "annoyed" is a bad excuse - there are tools like Automatic categorizer and other that quickly help you
even in experimental stage.


Let me tell you a true story: 
I once met a team which really had some trouble: they had not used versioning systems yet (just zipped code state from time to time) and they asked me as an external in helping them with consultation and better processes.

At that time there was one very knowledgable guy in the team who developed mostly on his own, he was the "local hero". When he wanted to work from home he zipped the code and worked. When he returned he extracted his code state (overwriting destroying anything others had changed meanwhile) because his code was always the "best". He never had the time to explain what he did and he was so busy to not even follow a few coding standard. Yes - they had rules - and while all followed them easily and properly you can guess who constantly violated them. In the end this just helped him keeping this position and not really giving the others any chance to contribute something useful (although they have written decent and good code).

From his point of view anything was satisfying because regarding the set he found many excuses (there was "no time" this week to follow what the team agreed to and so on). As he also wanted to adopt or remove the rules like he required for any new idea. In the end anything looked different, the team did not know which thing to follow and which not. He (for the moment) stayed  the most knowledgable just because he was knowing his own bad decisions and history. 

But the overall code quality was lousy, more and more bugs lead to trouble and the rest of the team was extremely unsatisfied. Team did good work - but it was useless because the chain was only as strong as its weakest link. And still no one wanted to really veto against his bad behavior. Introducing a versioning system in that team was helpful - no one lost code anymore, quality improved. Hardest part was to convince that the rules make sense and avoided conflicts in the team due to "bad and adopted excuses".

Rules and coding them as release tests have the following benefit:
 - everyone can look at them 
 - they get applied and constantly monitor the quality measures
 - we can ensure that things stay proper and consistent (without wasting too much time in reviewing manually 
    if new contributions increase or decrease quality level)
 - they can give new contributors the trust that their contribution is also properly done and fits to the rest of the image 

But returning to the issue I stay with my points and the PR raised:
 - basicNew is an "instance creation" method and should therefore stay in that category (because this is where one would 
   expect it)
 - it does not make sense to ask for that consistency only for specific Kernel packages - the scope of the release tests 
    is the "image" (=release) so thats why it is proposed that way 

If someone is "annoyed" then he could/should 
 - either work in a branch and wait with integration
  - or use available tools like the automatic method categorizer making this a no-brainer

PR #9900, opened 2021-09-05
Expand String class comment.

ctSkennerton:
- convert to Markdown
- expand examples
- add in some links to collaborating classes.


Ducasse:
Excellent!
I'm thinking how we could reuse the executable comments so that we are sure that the comments will be always up to date.

PR #9911, opened 2021-09-06
modify Spec2-Common-Widgets dependencies

estebanlm:
since it should be is Spec2-CommonWidgets now :)


estebanlm:
I don't know why is the timeout on mac, but otherwise it is working :)

MarcusDenker (approved):

MarcusDenker:
The mac timeout I think is not related

PR #9920, opened 2021-09-07
Pharo10 - 9038-windows edges grabbing for resizing not optimal

tesonep:
Porting to P10


MarcusDenker:
one test times out, not related, I opend an issue to make this test time out less in the future: https://github.com/pharo-project/pharo/issues/9922

PR #9923, opened 2021-09-08
9909-Deprecate-typeOfClass-and-kind-and-add-mcType-instead

### LINKED ISSUE #9909: Deprecate #typeOfClass and #kind and add #mcType instead
**Describe the problem**
The methods #typeOfClass and #kind have lead to confusion in the past. 

They look like generally useful methods to get a symbol describing a layout. But in reality, they are just useful for Monticello backward compatibility (e.g. some layouts have special symbols like #normal, while others just use the name of the layout).

They are now used only by monticello (after merging #9908). Thus we can deprecate them and have a very explicity #mcType method for internal monticello use.

MarcusDenker:
- deprecate #typeOfClass
- deprecate #kind
- add #mcType and #mcTypeSymbol (can not be mcType as all classes need to return their correct #mcType, including the Layout Classes)

fixes #9909

### DIFF
--- a/src/Deprecated10/Behavior.extension.st
+++ b/src/Deprecated10/Behavior.extension.st
@@ -1,0 +1,12 @@
+Extension { #name : #Behavior }
+
+{ #category : #'*Deprecated10' }
+Behavior >> typeOfClass [
+
+	self
+		deprecated: 'Should use mcType instead'
+		on: '08 September 2021'
+		in: #Pharo10
+		transformWith: '`@rec typeOfClass' -> '`@rec mcType'.
+	^ self mcType
+]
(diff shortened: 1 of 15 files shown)


MarcusDenker:
The idea why I think this is good: typeOfClass and kind are far too general as selectors. This is Monticello internal and should maybe even later be part of the monticello package as extensions.

(and all this confused so much that even we made mistakes in the past... and used this in other code...)

Ducasse:
Could you add a comment in the methods saying that nobody should use this beside monticello?

MarcusDenker:
yes, that is a good idea. The same kind of comment is to be done for #kindOfSubclass, which makes only sense for ST80 style class definitions (e.g. no dedicated string for CompiledMethod).

I think the best is to merge this PR, then I do another one with the comments.

Issue #9921, opened 2021-09-08
unify #isScripting and #isForScripting


MarcusDenker:
**Describe the problem**
Some part of the system uses isScripting, others #isForScripting for the check if a mode/editor/tool is working on methods vs. expressions.

** What need to be done ***

We should unify and have only one selector for this

MarcusDenker:
TODO:

- [x] add #isScripting in system everywhere where we just has #isForScripting https://github.com/pharo-project/pharo/pull/11412
- [x] add #isScripting in Spec everywhere where we just has #isForScripting https://github.com/pharo-spec/Spec/pull/1288
- [x] Change senders in Spec https://github.com/pharo-spec/Spec/pull/1289
- [x] Change senders in Image https://github.com/pharo-project/pharo/pull/11416
- [x] Replace #isWorkspace with #isScripting https://github.com/pharo-project/pharo/pull/11416

PR #9926, opened 2021-09-08
9925-Defining-variables-in-the-debugger-DNU-isForScripting 

### LINKED ISSUE #9925: Defining variables in the debugger: DNU #isForScripting
**Bug description**
There is a confusion between isForScripting and isScripting (see https://github.com/pharo-project/pharo/issues/9921).

This lead to a bug that when you try to define a variable, the wrong method is sent (this happens in some cases in the debugger)

I propose as a workaround to add #isForScripting for this case, we can then backport this to Pharo9, too.
(the real solution is to unify the two, but that will be a bigger change)

MarcusDenker:
Workaround, fixes #9925

### DIFF
--- a/src/Rubric/RubAbstractSmalltalkMode.class.st
+++ b/src/Rubric/RubAbstractSmalltalkMode.class.st
@@ -19,6 +19,12 @@ RubAbstractSmalltalkMode >> isCompletionEnabled [
 ]
 
 { #category : #testing }
+RubAbstractSmalltalkMode >> isForScripting [
+	"needs to be unified is #isScripting"
+	self isScripting
+]
+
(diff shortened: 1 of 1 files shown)


MarcusDenker:
For the real problem, I added https://github.com/pharo-project/pharo/issues/9921 (we need to unify isForScripting and isScripting, but that is a multi-step thing due to newtools vs image). And I added an issue to explore backporting to Pharo9: https://github.com/pharo-project/pharo/issues/9927 (of the workaround only, of course)

guillep:
Yes, I think this is good. Tests are on the way, but since this problem is only reproduced from the debugger, those tests come in new tools.

PR #9937, opened 2021-09-08
enhance code changes browser a little bit

estebanlm:
do not use morphs and display modification time first (it gives a better idea of what is happening)


estebanlm:
tests are ok :)
the idea is that the code changes now will look like this: 

image

(more changes to come, but this is the first small step)

MarcusDenker (approved):

Issue #8242, opened 2021-01-04
#hasPackageMatchingExtensionName: uses symbol concatenation


chisandrei:
Related to issue https://github.com/pharo-project/pharo/issues/8202, there is another symbol concatenation in `RPackageOrganizer>>#hasPackageMatchingExtensionName:`

https://github.com/pharo-project/pharo/blob/c3cf32fbfd7c80b85b59c8df103ce36e8af1665e/src/RPackage-Core/RPackageOrganizer.class.st#L456

```
hasPackageMatchingExtensionName: anExtensionName

	(self hasPackageExactlyMatchingExtensionName: anExtensionName)
		ifTrue: [ ^true ].
		
	packages keysDo: [ :aSymbol | 
		(anExtensionName beginsWithEmpty: aSymbol, '-' caseSensitive: false)
			ifTrue: [ ^ true]].
	^ false
```

Not sure if this causes performance problems but it creates not needed symbols so we can also remove it.  However it's the same check as in `RPackageOrganizer >> packageMatchingExtensionName:` so the same solution could be used.

MarcusDenker:
from @syrel:

We could also avoid concatenation all together, which would be more than 50% faster.

Since we always concatenate `aSymbol` with `'-'` we do not have to care about empty string (`withEmpty:` part of `beginsWithEmpty:`). Next since `-` is constant we can directly compare the value at a destination index with `-`

```smalltalk
aSymbol := #Package.
anExtensionName := 'Package-Tag-Examples'.

aSymbolIndex := anExtensionName findString: aSymbol startingAt: 1 caseSensitive: false.
aMinusIndex := aSymbolIndex + aSymbol size.
hasPrefix := aSymbolIndex > 0
	and: [ (aMinusIndex <= anExtensionName size)
		and: [ (anExtensionName at: aMinusIndex) = $- ] ]
```

MarcusDenker:
The Symbol concatenation has been removed, I think it is best to close this issue.

I tried the other suggestion, but the bootrap/full image build show that this is not that easy.

I think the better strategy is to get rid of prefix matching alltogether in the context of the Rpackage cleanups

PR #9941, opened 2021-09-09
NoSymbolCreation-hasPackageMatchingExtensionName 

MarcusDenker:
as described in #8242, this PR avoids a symbol creation in RPackageOrganizer>>#hasPackageMatchingExtensionName:

The issue tracker entry has another idea to speed this up even more which we should try afterward (the issue tracker entry thus stays open even after merging this)


MarcusDenker:
I added an issue tracker entry about symbols in general: https://github.com/pharo-project/pharo/issues/9962

PR #9959, opened 2021-09-14
9924-CDBehaviorDefinitionNode-uses-subclassDefiningSymbol-not-unique

### LINKED ISSUE #9924: CDBehaviorDefinitionNode uses #subclassDefiningSymbol: not unique
**Bug description**
CDBehaviorDefinitionNode was using a mix of mc symbols and #subclassDefiningSymbol to model the layout. We have unified this to use just the subclassDefiningSymbol (because this is what the code assumed, mc symbols were only used in tests and when parsing).

The problem with subclassDefiningSymbol is now that it is just there for backward compatibility and encodes what the old class definition uses. E.g. there is no dedicated symbol for CompiledMethodLayout. User-defined layouts are not supported (they just answer #subclass:). 

We need to use something that is unique per layout. This could be the name of the layout as a symbol. Maybe better: the layout class itself. This would allow us to simply call #subclassDefiningSymbol for the cases we need that string.

MarcusDenker:
- store the layoutClass in CDBehaviorDefinitionNode, not a symbol
- fix all tests to check for that
- fix CDClassDefinitionParser to do the right thing
- simplify #buildFromAST: which can now just take the layout from the CDBehaviorDefinitionNode

fixes #9924

### DIFF
--- a/src/ClassParser/CDClassDefinitionParser.class.st
+++ b/src/ClassParser/CDClassDefinitionParser.class.st
@@ -20,9 +20,19 @@ CDClassDefinitionParser class >> fromASTNode: aNode [
 
 { #category : #parsing }
 CDClassDefinitionParser >> handleClassName: aNode withType: aSymbol [
+	| layout |
 	
 	self handleClassName: aNode.
-	classDefinition classKind: aSymbol
+	layout := FixedLayout.
+	aSymbol = #variableWordSubclass: ifTrue: [ layout := WordLayout ].
+	aSymbol = #ephemeronSubclass: ifTrue: [ layout := EphemeronLayout ].
+	aSymbol = #weakSubclass: ifTrue: [ layout := WeakLayout ].
+	aSymbol = #variableByteSubclass: ifTrue: [ layout := ByteLayout  ].
+	aSymbol = #variableSubclass: ifTrue: [ layout := VariableLayout ].
+	aSymbol = #immediateSubclass: ifTrue: [ layout := ImmediateLayout ].
+	aSymbol = #subclass: ifTrue: [ layout := FixedLayout ].
+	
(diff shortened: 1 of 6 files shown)


src/ClassParser/CDClassDefinitionParser.class.st
```diff
@@ -20,9 +20,19 @@ CDClassDefinitionParser class >> fromASTNode: aNode [
...
 	self handleClassName: aNode.
-	classDefinition classKind: aSymbol
+	layout := FixedLayout.
+	aSymbol = #variableWordSubclass: ifTrue: [ layout := WordLayout ].
+	aSymbol = #ephemeronSubclass: ifTrue: [ layout := EphemeronLayout ].
+	aSymbol = #weakSubclass: ifTrue: [ layout := WeakLayout ].
+	aSymbol = #variableByteSubclass: ifTrue: [ layout := ByteLayout  ].
+	aSymbol = #variableSubclass: ifTrue: [ layout := VariableLayout ].
+	aSymbol = #immediateSubclass: ifTrue: [ layout := ImmediateLayout ].
+	aSymbol = #subclass: ifTrue: [ layout := FixedLayout ].
```
dionisiydk, line 33: It can be simplified with some lookup method based on  #subclassDefiningSymbol.

dionisiydk (approved):

src/ClassParser/CDClassDefinitionParser.class.st
Ducasse: Yes :)
I will integrate this one and we continue.

MarcusDenker: I added https://github.com/pharo-project/pharo/issues/9961

PR #9960, opened 2021-09-14
String #beginsWith:caseSensitive and #endsWith:caseSensitive: 

dionisiydk:
1. Introduce #beginsWith:caseSensitive and #endsWith:caseSensitive: String methods using existing primitive approach for substring search (see current beginsWith:) 
The new #begins method should replace quite strangely named #beginsWithEmpty:caseSensitive (in another PR). The old method actually does not work properly for wide strings (covered by tests for new methods).

2. Remove duplicated #beginWith:  in String subclasses. Simply check the bytes  type of string in the base class to decide if super implementation is required (the type check was already there but only for the argument). 
The logic of those  methods is not that simple actually due to the optimization parts. And therefore I believe it is a better  to not spread it over classes.

3. Optimize endsWith: using same logic as #beginsWith:

### DIFF
--- a/src/Collections-Strings-Tests/StringTest.class.st
+++ b/src/Collections-Strings-Tests/StringTest.class.st
@@ -554,6 +554,30 @@ StringTest >> testAt [
 ]
 
 { #category : #tests }
+StringTest >> testBeginsWithCaseSensitive [
+	| w1 w2 |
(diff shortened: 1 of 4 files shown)


MarcusDenker (approved):
We reviewed it an fixed one comment: ok to merge

PR #9967, opened 2021-09-15
Cleanup-Deprecated-09-15

MarcusDenker:
another removal of deprecated methods that have no senders.


Ducasse:
Marcus I would keep dist because it may be used externally.

astares (approved):

PR #9968, opened 2021-09-15
Compiler Segmentation Fault

codereport:
I came across a compiler seg fault:
```
stack page bytes 8192 available headroom 5576 minimum unused headroom 5944

	(Segmentation fault)
```
If I recall, the problem has something to do with `mustBeBooleanDeOptimize` because we are return an integer from the `=` method. 

Hopefully the problem is easy to fix!


MarcusDenker:
Thanks! We will have a look.

MarcusDenker:
If I run this in Pharo10 running on an intel Mac, I get a DNU for #ifFalse:, as expected.

This is a the #= method was changed and does not return a boolean, but an integer:

```
= aSimpleClass
    ^ (var = aSimpleClass var) asBit
```

Which version of Phro are you using?

MarcusDenker:
( I close this for now, if you can reproduce it in Pharo10, please re-open)

dvmason:
To see the bug, you seem to have to have a loop. So simple linear code won't exhibit the problem

MarcusDenker:
But this code in a loop would just open *a lot* of debuggers?

guillep:
Hi Guys,

Thanks for the report and the PR with the test.
However, I loaded the PR but cannot reproduce the issue.
I'd gladly work on the issue if we have a reproducible case.

In the meantime, I'll close it. Feel free to reopen if you have more info.

codereport:
I am using:
```
Pharo 10.0.0
Build information: Pharo-10.0.0+build.254.sha.0a86a3d9b5984e174267b756f437baf5f7c7c6ab (64 Bit)
```
But any Pharo10 should work.

Steps to reproduce bug:
1. Add the following method to SmallInteger:
```st
ifFalse: aFullBlockClosure 
   ^ self = 1 ifFalse: aFullBlockClosure 
```
2. Go to `testBugFailure` in `ClassParser-Tests` / `ASTClassBuilderTest`
3. Run `testBugFailure`
4. Pharo should crash

I would have pushed the code from #1 but I can't figure out how to.

codereport:
@guillep @MarcusDenker For some reason, there is no "Re-open PR button". I think you need to.

MarcusDenker:
But this is not the same code... maybe opening an issue instead of a PR would be good? The code now has nothing to do with the PR

guillep:
Thanks! That test case is indeed beneficial. I confirm I could reproduce the bug.
I'll reopen a new issue instead of a PR with all the info!

Thanks again!

PR #9975, opened 2021-09-16
Cleanup-Deprecated-Spec-Examples-Tests

MarcusDenker:
This PR removes the Test and Examples packages from Spec1. removing those now makes the next step easier to find out where spec1 is still used.


Ducasse:
Yes yes yes

Issue #9931, opened 2021-09-08
FFITypesTest>>testSignedLongLong sometimes failing on the CI


MarcusDenker:
**Bug description**
On the CI (only seen on the mac for now) we have very seldomly:

```
Error
Got 2147483647 instead of -2147483649.
Stacktrace
TestFailure
Got 2147483647 instead of -2147483649.
FFITypesTest(TestAsserter)>>assert:description:resumable:
FFITypesTest(TestAsserter)>>assert:description:
FFITypesTest(TestAsserter)>>assert:equals:
[ :int |
		|ref|
		ref := ByteArray new: FFIInt64 externalTypeSize.
		ref signedLongLongAt: 1 put: int.
		self assert: (ref signedLongLongAt: 1) equals: int ] in FFITypesTest>>testSignedLongLong
Array(SequenceableCollection)>>do:
FFITypesTest>>testSignedLongLong
```

MarcusDenker:
can fail on win, too

MarcusDenker:
Does not happen anymore...

MarcusDenker:
happening again

guillep:
This is very probably a JIT vs Interpreter difference. The test works ok as soon as you execute it again. So probably there is a border case that is not well implemented in the interpreter.

guillep:
We worked on this this afternoon with @Ducasse 

we have a fix (or several fixes because there are several bugs) in the VM here: https://github.com/pharo-project/pharo-vm/pull/1066

demarey:
Since the fix is merged in the VM repository, I will close the issue.

PR #9978, opened 2021-09-17
Cleanup-Duplicated-Methods-09-17

MarcusDenker:
remove some more methods that are duplicated in the hierarchy.
(another step for #9703)


MarcusDenker:
tests look good, randomly failing test is FFITypesTest>>testSignedLongLong (see #9931)

PR #9984, opened 2021-09-17
9565 pharo 9  seaside  process dying on cloud instance

StevenCostiou:
Fixes #9565 (see that issue for a detailed analysis) and #9919.

Basically we changed P9 to kill the image when an error occured in command line.
In P8, it handled the error without quitting, we fixed it so that it worked again.

PR #9985, opened 2021-09-17
9565 pharo 9  seaside   process dying on cloud instance

### LINKED ISSUE #9565: Pharo 9 + Seaside - Process dying on cloud instance
A few weeks back it was discussed on Discord and Pharo ML what would be minimal costs to run a Pharo web application online.
So I wanted to check on Azure cloud from Microsoft with a minimal Linux VM running what would be the average costs in running something with Pharo using the cloud.

To perform this test I used Pharo and Seaside on a cloud virtual machine (using Microsoft Azure). The Seaside image is exposed on a public IP address. I've let it run overnight and every second day the Pharo process/virtual machine was gone. It died inbetween which was reproducible. Unfortunately Pharo left no traces why it died - there was no debug log or other.

Maybe it is related to the resource restricted tiny Linux VM with not much RAM. 

Peter Svenson tried the same with similar results.

StevenCostiou:
Fixes #9565 => see the P9 PR https://github.com/pharo-project/pharo/pull/9984

### DIFF
--- a/src/Settings-Graphics/GraphicFontSettings.class.st
+++ b/src/Settings-Graphics/GraphicFontSettings.class.st
@@ -10,7 +10,7 @@ Class {
 { #category : #fonts }
 GraphicFontSettings class >> displayScaleFactorForStyleNamed: aSymbol [
 
-	^ (#(#(#superTiny 0.6) #(#tiny 0.8) #(#small 1) #(#medium 1.3) #(#large 1.6) #(veryLarge 2) #(huge 2.7)) 
+	^ (#(#(#superTiny 0.6) #(#tiny 0.8) #(#small 1) #(#medium 1.4) #(#large 1.6) #(veryLarge 2) #(huge 2.7))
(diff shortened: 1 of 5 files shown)


astares:
@StevenCostiou Be aware that this PR against Pharo 10 is having more changes (5 changes files) than the actual fix (which is only in a single method). This gives the conflicts.

You have to be on the level of upated Pharo10 branch to provide a PR for Pharo 10.

StevenCostiou:
> @StevenCostiou Be aware that this PR against Pharo 10 is having more changes than the actual fix. This gives the conflicts.
> 
> You have to be on the level of upated Pharo10 branch to provide a PR for Pharo 10.

Ah, yes, I will fix that.

svenvc:
This PR changes more files than just CommandLineUIManager >> handleDebugRequest: 

I am not sure this is right.

StevenCostiou:
It's wrong I have to start again from a P10 image, I cannot manage to do it from the same as my P9 image.
But I cannot do it now.

At least the fix is known.

PR #9989, opened 2021-09-18
Cleanup-Duplicated-Methods-09-18

MarcusDenker:
remove more methods that are not needed as the same method exists in a superclass

another step for #9703 

(365 to go)


MarcusDenker:
There are two tests failing, but they are not related. I have added new issue tracker entries for those: https://github.com/pharo-project/pharo/issues/9996 https://github.com/pharo-project/pharo/issues/9995

Issue #9992, opened 2021-09-19
Missing #pharo10.x and #pharo10.0.x in Metacello ...


Ducasse:
```
metacelloPlatformAttributes
    "Returns the tags for the conditional platform loading in Metacello. Pay attention the order is important: from most  to least general." 
    "For release integrators, we should not have #'pharo1.3x' **and** #'pharo1.4.x'"
   ^ #(#squeakCommon #pharo #'pharo9.x' #'pharo9.0.x' #'pharo10.x'  #'pharo10.0.x')
```

PR #9994, opened 2021-09-19
fix: #9992 add pharo10 for metacello

Ducasse:


Ducasse:
Got broken because MD lost a method :( so added it back and rekicking it to see

Ducasse:
This is strange because in master and dev of microdown I published newFromAssociations: so I do not get why it is not loading.

Ducasse:
ah the build is not redone from the PR closing because the changes in not visible... of course.

Issue #9996, opened 2021-09-20
Test failing on CI: NoUnusedVariablesLeftTest>>testNoUnusedInstanceVariablesLeft


MarcusDenker:
**Bug description**
A PR got merged that leads to a test failure on the CI (for the build and for all CI runs we do from now on): #testNoUnusedInstanceVariablesLeft

```
Error
the following classes have unused instance variables and should be cleaned: an OrderedCollection(MicrodownInlineParserTest)
Stacktrace
TestFailure
the following classes have unused instance variables and should be cleaned: an OrderedCollection(MicrodownInlineParserTest)
NoUnusedVariablesLeftTest(TestAsserter)>>assert:description:resumable:
NoUnusedVariablesLeftTest(TestAsserter)>>assert:description:
NoUnusedVariablesLeftTest>>testNoUnusedInstanceVariablesLeft
```

MarcusDenker:
has been fixed upstream

Issue #9995, opened 2021-09-20
Test failing on CI: ReleaseTest>>testMethodsContainNoHalt


MarcusDenker:
**Bug description**
A PR got merged that leads to a test failure on the CI (for the build and for all CI runs we do from now on): #testMethodsContainNoHalt

```
Error
Found methods with haltan OrderedCollection(MicInputfileBlock>>#path)
Stacktrace
TestFailure
Found methods with haltan OrderedCollection(MicInputfileBlock>>#path)
ReleaseTest(TestAsserter)>>assert:description:resumable:
ReleaseTest(TestAsserter)>>assert:description:
ReleaseTest>>testMethodsContainNoHalt
ReleaseTest(TestCase)>>performTest
```

MarcusDenker:
has been fixed upstream

PR #9997, opened 2021-09-20
Cleanup: remove unreferenced deprecated class in Spec-Deprecated

MarcusDenker:
This PR removes all deprecated classes from the package spec-deprecated that are not referenced.

The package now just contains one extension method, removing that is the next step


MarcusDenker:
There are tree tests failing, but they are not related
- see  #9996 #9995 for the failing tests related to halt and an usused ivar
- We are seeing StSpotterModelTest>>testAnnounceQueryEndedIsSentOnce failing on the mac slave sometimes, I will open an issue for that

PR #9998, opened 2021-09-20
9970-ReConsistencyCheckRule-is-a-very-bad-name 

### LINKED ISSUE #9970: ReConsistencyCheckRule is a very bad name
**Describe the problem**
ReConsistencyCheckRule is the rule that 'Checks for empty collection using #size instead of #isEmpty" or #isNotEmpty'

The name is really bad. We should have a more descriptive name

MarcusDenker:
Change the name to ReUseIsEmptyNotSizeRule, we keep the uniqueIdentifierName as before so all bans e.g. are not affected

fixes #9970


MarcusDenker:
There are two tests failing, but they are not related (see  #9996 #9995 for the failing tests)

PR #9999, opened 2021-09-20
Protocol for all class side methods `#reset` changed to `initialization`

### LINKED ISSUE #2607: #reset methods should be in 'initialization' protocol and a test should ensure we do not have regression
	
Original case: https://pharo.fogbugz.com/f/cases/20683
Status: Work Needed
Project: Usability
Date: 13 November 2017 1:33:33 pm
	
Description:
	
First we should correct and integrate the subcases. 

We should add a test/rule to ensure #reset methods are in 'initialization' by convention. 

See : http://forum.world.st/Convention-for-quot-reset-quot-methods-td5019493.html

gypsydave5:
starts to fix #2607, at least for this repo.

Should finally add the test when all of the other repos are fixed.

### DIFF
--- a/src/AST-Core/ASTCache.class.st
+++ b/src/AST-Core/ASTCache.class.st
@@ -61,7 +61,7 @@ ASTCache class >> default: anASTCache [
 	default := anASTCache
 ]
 
-{ #category : #initialization }
+{ #category : #'class initialization' }
 ASTCache class >> reset [
 	<script>
 	self default reset.
--- a/src/Beacon-Core/MemoryLogger.class.st
+++ b/src/Beacon-Core/MemoryLogger.class.st
@@ -36,7 +36,7 @@ Class {
 	#category : #'Beacon-Core-Loggers'
 }
 
-{ #category : #registering }
+{ #category : #'class initialization' }
 MemoryLogger class >> reset [
 	self instance reset
 ]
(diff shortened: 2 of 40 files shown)


MarcusDenker:
Failing tests not related, open issue for the failing tests:

- https://github.com/pharo-project/pharo/pull/10000
- https://github.com/pharo-project/pharo/issues/9996
- https://github.com/pharo-project/pharo/issues/9995

MarcusDenker:
Ah, the question came up if it should not be "class initialization" instead (see issue tracker)

gypsydave5:
> Ah, the question came up if it should not be "class initialization" instead (see issue tracker)

Could do with some guidance on this point!

astares:
@gypsydave5 
- on the instance side the category is called "initialization" and typically hosts instance methods like #initialize
- on the class side the category is often called "class initialization" and typically hosts class methods like #initialize (for class initialization) or #reset (to reset state of class variables or other)

So if you adopt your PR to put the method into a category named "class initialization" instead of "initialization" because you moved it to the class side now then I guess your PR will be happily accepted.

gypsydave5:
> adopt your PR to put the method into a category named "class initialization" instead of "initialization"

easy enough!

gypsydave5:
OK, this should now be the same but with `class initialization` protocol instead of `initialization`.

(now I have to do the same with all the other PRs 🤣)

gypsydave5:
@astares @MarcusDenker - is this now in a state where it can be merged? The failed check looks Win specific, a regression that's been fixed? I'd rerun the check but I'm not sure how (or if I can).

MarcusDenker (approved):

Issue #6311, opened 2020-05-05
Enhance Completion to be more fluid in large images


VincentBlondeau:
**Describe the request**
When writing a piece of code containing a message that is not existing yet, each addition of a letter to the non existing message will take a second. It seems that each completion request triggers the creation of a new "generator" in the CoFetcher. It doesn't seems to reuse the previously generated one and should since it is the same word we are writing.

**Expected behavior**
A more fluid one

**Expected development cost**
Ask @guillep and @tesonep ;)

**Screenshots**
![LF9SxRQdVt](https://user-images.githubusercontent.com/4706329/81094471-d639af80-8f03-11ea-8ca7-38db02275072.gif)

**Version information:**
 - OS: W10
 - Version: Pharo-9.0.0+build.288.sha.bc25507fbe677d4df5f4abddb533931599c76bba (64 Bit)

**Additional context**
Only visible in large images

VincentBlondeau:
Profiling report:
```
 - 24973 tallies, 24998 msec.

**Tree**
--------------------------------
Process: (60) Input Event Fetcher Process: InputEventFetcher>>waitForInput
--------------------------------
7.4% {1859ms} InputEventFetcher>>installEventLoop
  7.4% {1859ms} InputEventFetcher>>eventLoop
    7.4% {1859ms} InputEventFetcher>>waitForInput
--------------------------------
Process: (40) Morphic UI Process: nil
--------------------------------
88.9% {22225ms} MorphicRenderLoop>>doOneCycle
  88.9% {22225ms} WorldMorph class>>doOneCycle
    88.9% {22225ms} WorldMorph>>doOneCycle
      88.9% {22225ms} WorldState>>doOneCycleFor:
        61.6% {15390ms} WorldState>>doOneCycleNowFor:
          |59.1% {14777ms} WorldMorph>>runStepMethods
          |  |59.1% {14777ms} WorldState>>runStepMethodsIn:
          |  |  59.1% {14775ms} CoCompletionEngine(CompletionEngine)>>resetCompletionDelay
          |  |    59.1% {14775ms} CoCompletionEngine(CompletionEngine)>>openMenu
          |  |      59.0% {14751ms} NECMenuMorph class>>engine:position:
          |  |        59.0% {14747ms} NECMenuMorph>>engine:position:
          |  |          59.0% {14741ms} NECMenuMorph>>narrowCompletion
          |  |            59.0% {14740ms} NECMenuMorph>>selected:
[...]

PR #10002, opened 2021-09-21
6311-Enhance-Completion-to-be-more-fluid-in-large-images

tesonep:
Improving the code completion by adding a decorator that can handle smoothly the narrowing and unnarrowing of the query.

When an unnarrowing has been performed the decorator just returns the previously stored results without needing to perform the same query again.


tesonep:
Fix #6311 in P10

guillep (approved):
Good

Issue #10001, opened 2021-09-21
CI: StSpotterModelTest has sometimes failing test on the mac CI


MarcusDenker:
**Bug description**

- StSpotterModelTest>>testAnnounceQueryEndedIsSentOnce failied in #9997 and #9999
- StSpotterModelTest>>testSpotterModelShouldWaitToPerformActualSearch failed in  #9999

We should check if these test are somehow timing related

MarcusDenker:
The reason is that the mac slave is very slow sometimes and the test hard-codes timing here:

```
waitUntilEnded: model

	[ (StSpotterModel delayBeforeQuery + 50 milliSeconds) wait ] 
	doWhileTrue: [ 
		model scheduler workers anySatisfy: [ :each | each isBusy ] ].
```

MarcusDenker:
After discussion with esteban, we decided to turn it off on the pharo CI for now

https://github.com/pharo-spec/NewTools/pull/315

PR #10004, opened 2021-09-21
Fix #8524: Fix Process>>#priority: incorrectly allowing lower priority process run while …

### LINKED ISSUE #8524: Process>>#priority: allows lower priority process run while higher priority one scheduled

tesonep:
When changing the priority of a process the process should be reschedule.


MarcusDenker:
Failing tests not related, open issue for the failing tests:

- https://github.com/pharo-project/pharo/issues/9996
- https://github.com/pharo-project/pharo/issues/9995

src/Kernel/ProcessorScheduler.class.st
```diff
@@ -176,6 +176,23 @@ ProcessorScheduler >> highestPriority: newHighestPriority [
...
+	The primitive does not try to resume a process in a higher priority.
+	By forcing the creation of a new process and suspending this one, 
+	we force the rescheduling of processes. 
+	
+	This message should not be used directly. 
+	The only application so far is when changing the priority of a process.
+	"
+
+	modifiedProcess isActiveProcess ifFalse: [ ^ self ].
+	[ modifiedProcess resume ] fork.
```
dionisiydk, line 192: I wonder if there can be a corner case where forked process will be executed before the following #suspend. The rescheduling can be triggered  just after the fork by other processes/events in the image. 
I think #fork needs to be done with higher priority than the modifiedProcess to avoid it.

src/Kernel/Process.class.st
```diff
@@ -448,10 +448,12 @@ Process >> priority [
...
 { #category : #accessing }
 Process >> priority: anInteger [
-	"Set the receiver's priority to anInteger."
+	"Set the receiver's priority to anInteger.
+	When changing the process priority we need to yield to make it reschedule the processor"
 
 	(anInteger between: Processor lowestPriority and: Processor highestPriority)
-		ifTrue: [ priority := anInteger ]
+		ifTrue: [ priority := anInteger. 
+			self isActiveProcess ifTrue: [ [self resume] fork. self suspend ] ]
```
dionisiydk: It seems #interpriorityYield: has no users in PR.
Should not it is be used here? Or is it inlined here but the method is not removed?

src/Kernel/ProcessorScheduler.class.st
guillep: Probably, but I think we should build a test for such a thing.
So far, there is not really much code that changes the priority of the running process dynamically, so the question that comes from there is "is it worth it?"

src/Kernel/Process.class.st
guillep: yes, that's strange. We will check. Thanks!

PR #7634, opened 2020-10-30
Use "Smalltalk vm wordSize" to determine the version of the VM: 32 or 64

VincentBlondeau:
Change isWin32 sends to isWindows to includes the isWin64.
Tested on Win64

Fixes:
#6709
#2246
#5968

Issue #5968, opened 2020-03-23
Adjust CallbackForWin64X64 class >> isForCurrentPlatform


bencoman:
Per #2246, it seems `SmalltalkImage current getSystemAttribute: 1001` only returns 'Win32', even on 64-bit images, because that is the historical name or the API regardless of word size.   So the existing code...
```
CallbackForWin64X64 class >> isForCurrentPlatform
	| processor |
	(#('Win32' 'Win64') includes: (SmalltalkImage current getSystemAttribute: 1001)) ifFalse:
		[^false].
	processor := (SmalltalkImage current getSystemAttribute: 1003) asLowercase.
	^processor first = $x and: [processor = 'x64' or: [('x86#64*' match: processor)]]
```
...works because it matches 'Win32', but then the redundant 'Win64' confounds proper understanding.  Propose to remove the 'Win64' and add a similar comment as proposed in #2246 for **OSPlatform class >> currentPlatformName**.

MarcusDenker:
fix was merged

PR #10008, opened 2021-09-21
Fix #7634: Use "Smalltalk vm wordSize" to determine the version of the VM: 32 or 64

guillep:
Redo https://github.com/pharo-project/pharo/pull/7634/files

Correctly ask if the vm is 32 or 64 bits and if we are in windows platform.

Fixes:
#6709
#2246
#5968


JanBliznicenko:
This change probably caused issue #10026

PR #10011, opened 2021-09-21
Deprecating documentPath.

### LINKED ISSUE #10007: VirtualMachine>>#documentPath is wrong
**Bug description**

**To Reproduce**
1. open a playground

**Expected behavior**

tesonep:
Fix #10007


guillep:
Actually, `documentPath` should just be deprecated. It is not equivalent to `imagePath`. I'm closing this issue on request of the author.

PR #10012, opened 2021-09-21
Redo #9845 -- Fix 9798: Deprecated class TextConverter is still used

### LINKED ISSUE #9798: Deprecated class TextConverter is still used
**Bug description**
TextConverter is deprecated, but it is used bu the method 
- String>>#convertFromEncoding:
- String>>#convertToEncoding:

These are still used in Monticello and by the Method Trailer encoding

guillep:
Replace convertFromEncoding:/convertToEncoding: with decodeWith:/encodeWith:
Replace users and implementation so deprecated encoder classes can be removed

Fix #9845
Fix #9798


svenvc:
LGTM

PR #10016, opened 2021-09-21
Deprecate #beginsWithEmpty:caseSensitive. Users should use beginsWith…

dionisiydk:
It deprecates String method #beginsWithEmpty:caseSensitive. It was oddly named and it was inconsistent to a simple #beginsWith: analog in case of empty prefix.
Now users should use #beginsWith:caseSensitive and take care about empty prefix on their own side when needed.
The image users are adopted accordingly

### DIFF
--- a/src/Collections-Strings-Tests/StringTest.class.st
+++ b/src/Collections-Strings-Tests/StringTest.class.st
@@ -577,20 +577,6 @@ StringTest >> testBeginsWithCaseSensitive [
 	self assert: (w1 beginsWith: w2 asUppercase caseSensitive: false).
 ]
 
-{ #category : #tests }
-StringTest >> testBeginsWithEmptyCaseSensitive [
-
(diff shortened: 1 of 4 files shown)


src/Collections-Strings-Tests/StringTest.class.st
```diff
@@ -577,20 +577,6 @@ StringTest >> testBeginsWithCaseSensitive [
 	self assert: (w1 beginsWith: w2 asUppercase caseSensitive: false).
 ]
 
-{ #category : #tests }
-StringTest >> testBeginsWithEmptyCaseSensitive [
-
-	self deny: ('éèàôüößäóñíá' beginsWithEmpty: 'ß' caseSensitive: true).
```
dionisiydk, line 583: I found many other StringTests with strange encodings. Not sure what to do with them.  But such cleaning required a separate issue

MarcusDenker (approved):

PR #10000, opened 2021-09-21
9946-CI-Flaky-test-ProcessSpecificTesttestDynamicVariable

MarcusDenker:
a bit longer timeout, hopefully fixes #9946

PR #10019, opened 2021-09-22
Revert "9946-CI-Flaky-test-ProcessSpecificTesttestDynamicVariable"

MarcusDenker:
Reverts pharo-project/pharo#10000


MarcusDenker:
looking at the CI tests, this actually seems to be ok for now with 10 seconds.

PR #10028, opened 2021-09-23
 Fix #10026: Cannot use Iceberg - Pharo 10 Windows 64bits FFI

### LINKED ISSUE #10026: Cannot use Iceberg - Pharo 10 Windows 64bits FFI
**Bug description**
On clean image launch, two debugger windows pop.
First one is "Error: Can't find the requested origin". Looks like some system settings error.
Second one is "#substrings: was sent to nil". It comes from libgit2_init FFI call.

**To Reproduce**
  1. Open Iceberg
  2. Try to repair pharo repository
  3. Point to existing repo on hard drive
  4. Debugger pops up.

On the second try, just opening Iceberg triggers the error.

**Call stack**

UndefinedObject>>doesNotUnderstand: #substrings:
FFIWindowsLibraryFinder(FFILibraryFinder)>>ldLibraryPaths
FFIWindowsLibraryFinder(FFILibraryFinder)>>paths
[...]
LGitRepository(Object)>>ffiCall:options:
LGitRepository>>repository_open:bare_path:

**Additional informatio**
Last stack level is ```FFILibraryFinder>>ldLibraryPaths
	^ (OSEnvironment current at: 'LD_LIBRARY_PATH' ifAbsent: [ '' ]) substrings: ':'```

Inspecting `OSEnvironment current`  Meta is is a PlatformIndependentEnvironment (right-click on browse open Browser on Undefined Object though).
at:ifAbsent: returns nil. But it probably should be a Win32Environment?

**Version information:**
 - OS: Window10 64bits
 - Pharo 10 64bits

guillep:
Fix #10026 

Adapt platform recognition for Windows 64 bits.

### DIFF
--- a/src/System-OSEnvironments/Win32Environment.class.st
+++ b/src/System-OSEnvironments/Win32Environment.class.st
@@ -20,7 +20,7 @@ Class {
 
 { #category : #testing }
 Win32Environment class >> isDefaultFor: aPlatform [ 
-	^ aPlatform isWin32
+	^ aPlatform isWindows
 ]
 
 { #category : #accessing }


MarcusDenker (approved):

astares:
@guillep / @MarcusDenker / @tesonep 

The original problem in #10026 was that we had a DNU when starting the image on Windows (nil was not able to understand #substring:)

If one looks into the debugger when the problem appears because of  FFILibraryFinder>>#ldLibraryPaths

image

asks for LD_LIBRARY_PATH and if absent gives a block just to make sure to always have a string. So if LD_LIBRARY_PATH is not there to have at least an empty string that can then also understand #substring: 

The problem appeared because a PlatformIndependentEnvironment was used and  PlatformIndependentEnvironment>>#at:ifAbsent: was just returning "nil" (and is ignoring any argument including the block

image

So while the PR fixes the problem now by giving the right environment (Windows) I guess this problem can still happen again.

Wouldnt it make sense that additionally PlatformIndependentEnvironment>>#at:ifAbsent: would be implemented as

``` Smalltalk
at: aKey ifAbsent: aBlock

	^ aBlock value
```
instead of hard returning nil
``` Smalltalk
at: aKey ifAbsent: aBlock

	^ nil
```

MarcusDenker:
at:ifAbsent: not evaluating the block looks strange, yes

astares:
So I opened https://github.com/pharo-project/pharo/issues/10041 so we can care

Issue #10021, opened 2021-09-22
CI: Test failing related to text encoding


MarcusDenker:
**Bug description**
These tests are failing now on the CI:

```
CompiledMethodTrailerTest>>testEmbeddingCompressedSourceCode
GoferOperationTest>>testPush
MCDirectoryRepositoryTest(MCRepositoryTest)>>testAddAndLoad
MCSerializationTest>>testMcdSerialization
MCSerializationTest>>testMczSerialization
MCMczInstallerTest>>testInstallFromFile
MCMczInstallerTest>>testInstallFromStream
```

This might be caused by merging https://github.com/pharo-project/pharo/pull/10012

MarcusDenker:
fixed, I opnend https://github.com/pharo-project/pharo/issues/10050 for the one remaining

PR #10038, opened 2021-09-24
10021 - Fixing the export of Monticello.

tesonep:
The problem is that is using strings and saving them to the zip file.
Keeping that behavior, because if not we are not able to generate valid MCZ files.
Also, this code has a ticket to go away.


tesonep:
Fix #10021

PR #10040, opened 2021-09-24
Keep isQuick checks to the single place it matters

### LINKED ISSUE #10036: Defining quick methods in the debugger raise DNU #previousPc
**Bug description**

If you send a non implemented message in the debugger, the debugger offers the "Create" button.
Clicking on it creates a method from a template.
If we replace that template by a quick method definition, accepting it throws an exception.

In Pharo 10 Pharo-10.0.0+build.152.sha.66e37c8bbd3c9e84ab12d25f3c754296ab7a3edd (64 Bit)

guillep:
Fix #10036 
`isQuick` checks should only be made when the method PC is restarted (and that happens when a method is compiled/recompiled in the context of the debugger)

### DIFF
--- a/src/Debugger-Model/DebugSession.class.st
+++ b/src/Debugger-Model/DebugSession.class.st
@@ -88,14 +88,6 @@ DebugSession >> detectUIProcess [
 
 ]
 
-{ #category : #context }
-DebugSession >> downInContext: aContext [
-	"move down the context stack to the previous (enclosing) context"
-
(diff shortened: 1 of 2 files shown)


guillep:
Tests on https://github.com/pharo-spec/NewTools/pull/294

guillep:
Hi guys, I've merged the latest fixes in Pharo10 into this PR, so the fixes upstream are taken into account here too

PR #10043, opened 2021-09-24
Smooth drag and drop2

hogoww:
Following #9337 , second attempt at a smoother drag and drop.
This is closer to other tools's drag and drop by adding a delay before expanding a package list for example.
It is configurable by Code browsing / Calypso /DragAndDrop delay.
Default delay is currently at 1500 ms

Current limitation is that the target of the drag and drop gets the event *on move*.
So you need to move after the delay is passed for it to open.

Let me know what you think.

Pierre


Ducasse:
to check the tests

src/Calypso-Browser/ClyDataSource.class.st
```diff
@@ -519,14 +539,13 @@ ClyDataSource >> updateItems: dataSourceItems [
...
 	| dropTargetItem result |
 	rowIndex = 0 | (rowIndex > self numberOfRows) ifTrue: [ ^false ].
 	self dragTransferType == type ifFalse: [ ^false ].
 	dropTargetItem := self elementAt: rowIndex.		
-	result := queryView wantsDropPassenger: aPassanger at: dropTargetItem.
-	
-	result & dropTargetItem hasChildren ifTrue: [ dropTargetItem expand ].
+	result := queryView wantsDropPassenger: aPassenger at: dropTargetItem.
 	
+	(result and: [ dropTargetItem hasChildren and: [ self isUserFocusingOnItem: dropTargetItem ]])  ifTrue: [ dropTargetItem expand ].
```
dionisiydk: Cool that you succeed with this feature. 

But I believe the DataSource is not the right place to put this logic and related new state. 
I suggest to move it into ClyQueryViewMorph. All this machinery can be done inside the call to #queryView:
```Smalltalk
queryView wantsDropPassenger: aPassenger at: dropTargetItem.
```
At the end it is a property of tree view to do the expansion on the mouse hover. And the tree view here is a queryView component.
So even the existing last line should be moved there:
```Smalltalk
result & dropTargetItem hasChildren ifTrue: [ dropTargetItem expand ].
```

src/Calypso-Browser/ClyDataSourceItem.class.st
```diff
@@ -49,9 +49,23 @@ Class {
 		'childrenDataSource',
 		'browserItem'
 	],
+	#classInstVars : [
+		'DragAndDropDelay'
```
dionisiydk: And same here. I would move it to ClyQueryViewMorph.

src/Calypso-Browser/ClyDataSource.class.st
```diff
@@ -66,7 +66,10 @@ Class {
 		'parentItem',
 		'lastFilteredDataSource',
 		'queryView',
-		'dirty'
+		'dirty',
+		'something',
```
dionisiydk: I guess #something is some garbage left in the PR

src/Calypso-Browser/ClyDataSource.class.st
```diff
@@ -66,7 +66,10 @@ Class {
...
+		'lastSelectedItem',
```
dionisiydk: I would rename #lastSelectedItem to be more explicit: #targetDropItem

src/Calypso-Browser/ClyDataSource.class.st
```diff
@@ -66,7 +66,10 @@ Class {
...
+		'timeInitialEvent'
```
dionisiydk: And #hoverOverDropItemStartTime instead of #timeInitialEvent

dionisiydk (changes requested):

hogoww:
Thanks a lot for the review @dionisiydk !
I haven't had time to fix it, but I will when I come back from holidays :)

guillep:
@hogoww could you take a look at this?

hogoww:
I was leaving it for this friday's sprint. 
Already checked it out, just need a few fixes

hogoww:
I did the requested changes.
However I am investigating why it does not work properly for protocols/classes expansion

hogoww:
Should be fine now.
Added some tests cases.
However i'm still not a fan of the naming.
Also, the package in which I put the test class is not... right. 

Suggestion welcome =]

Pierre

dionisiydk (approved):
Good job :)

dionisiydk:
> Also, the package in which I put the test class is not... right.

UI tests can go to Calypso-SystemTools-FullBrowser-Tests

src/Calypso-Browser/ClyQueryViewMorph.class.st
```diff
@@ -433,12 +435,33 @@ ClyQueryViewMorph >> hasRealQuery [
 	^self query ~~ ClyUnknownQuery instance
 ]
 
+{ #category : #'drag and drop support' }
+ClyQueryViewMorph >> hasUserBeenFocusingOnItem: aSelectedItem [
+	targetDropItem ifNil: [ ^ false ].
+	hoverOverDropItemStartTime ifNil: [ ^ false].
+	aSelectedItem browserItem = targetDropItem browserItem ifFalse: [ ^ false ].
+	 ^ (hoverOverDropItemStartTime copy addSeconds: aSelectedItem class dragAndDropDelay / 1000) < Time now
```
dionisiydk: I would moved the dragAndDropDelay into ClyQueryViewMorh. And make it an instance of duration (2 seconds). In that case this code will be very simple and readable:
```Smalltalk
hoverOverDropItemStartTime + DragAndDropDelay < Time now
```

src/Calypso-Browser/ClyQueryViewMorph.class.st
```diff
@@ -677,6 +700,23 @@ ClyQueryViewMorph >> setUpDataSource: aDataSource [
 	self ensureSelectedItemIfNeeded
 ]
 
+{ #category : #'drag and drop support' }
+ClyQueryViewMorph >> shouldExpand: aSelectedItem [
```
dionisiydk: maybe name like #isReadyToExpandForDrop:

hogoww: That's better.
My problem is that I'd like to show that this is indeed a tester, but one that modifies the state

src/Calypso-Browser/ClyQueryViewMorph.class.st
hogoww: Good point, thanks !

hogoww:
Perfect, I'll fix that tomorrow, thanks =)

hogoww:
Note for myself, there's also 3 tests failing, because the class variable is not well defined.

src/Calypso-Browser/ClyQueryViewMorph.class.st
hogoww: Not so great for editing the setting though, it displays a duration.
So you have to know that it should be an integer... 
Although I like the idea, this does not really fit for user interaction, i think.
Still, I tried to put it anyway in #isReadyToExpand so the code is more readable.

Ducasse:
So can we fix this and integrate it soon?

hogoww:
I've been working on this for two more hours this morning.
Lots of small stuff, that makes this a pain

hogoww:
Still not ready, there's a crutial missing test case.

hogoww:
Turns out playing with time is not that easy.
The underlying logic works, but there's still a problem which makes it inactive.
I burned my batterie to fix all of that, and therefore cannot continue. (too much diff, trying to create a new image kinda fails...

I'll finish that on monday..... Hopefully.

src/Calypso-Browser/ClyQueryViewMorph.class.st
dionisiydk: For settings you just need extra two accessors: 
```Smalltalk
ClyQueryViewMorph class>>dragAndDropDelaySeconds: aNumber
    DragAndDropDelay := aNumber seconds

ClyQueryViewMorph class>>dragAndDropDelaySeconds
    ^DragAndDropDelay asSeconds
```

src/Calypso-SystemTools-FullBrowser/ClyFullBrowserMorph.class.st
```diff
@@ -129,6 +129,20 @@ ClyFullBrowserMorph class >> registerToolsOn: registry [
 	registry register: self as: #browser
 ]
 
+{ #category : #settings }
+ClyFullBrowserMorph class >> settingsDragAndDropDelayOn: aBuilder [
+	<systemsettings>
+	(aBuilder setting: #dragAndDropDelay)
```
dionisiydk: And here it would be "aBuilder setting: #dragAndDropDelaySeconds"
I would also define this method on ClyQueryViewMorph directly because it's a configuration of this widget and not the full browser.

src/Calypso-Browser/ClyQueryViewMorph.class.st
hogoww: I feel seconds are not a good granularity.
Microseconds are too low.
Milliseconds seems ok, from a user perspective.

However there is a whole in the Duration/Time stuff and some methods are not available,
Microseconds, which is how I calculate the time is not, particularly.
Makes for harder calculation.

Anyway, that could be improved slightly, but I'm tired of this PR, so that;ll be good enough

hogoww:
Thanks for all your comments @dionisiydk !
They were quite helpful =]

hogoww:
Failling tests cases are unrelated, as far as I can see

MarcusDenker:
- [CI] Failing randomly: StSpotterModelTest>>testSpotterModelShouldWaitToPerformActualSearch #10520

src/Calypso-Browser/ClyQueryViewMorph.class.st
```diff
@@ -114,17 +114,47 @@ Class {
...
+	],
+	#classVars : [
+		'DragAndDropDelay'
 	],
 	#category : #'Calypso-Browser-Table'
 }
 
+{ #category : #accessing }
+ClyQueryViewMorph class >> dragAndDropDelay [
+	"Modified by settings CodeBrowsing/Calypso: ClyFullBrowserMorph >> #settingsDragAndDropDelayOn:"
```
dionisiydk, line 129:
```suggestion
	"Modified by settings CodeBrowsing/Calypso: ClyQueryViewMorph >> #settingsDragAndDropDelayOn:"
```

dionisiydk (approved):

src/Calypso-Browser/ClyQueryViewMorph.class.st
dionisiydk: Good stuff anyway. 
I am just curious what was missing in the duration/time algebra and how we can improve it to make it suitable for such cases

src/Calypso-Browser/ClyQueryViewMorph.class.st
hogoww: right, thanks x')

guillep (approved):
Thanks for all the combined effort!

Issue #9777, opened 2021-08-15
SystemNavigation default allUnimplementedCalls size -> DNU


Ducasse:
executing the following expression `SystemNavigation default allUnimplementedCalls size`
produces an error.

Ducasse:
when doing meth messages we get a GlobalVariable????

Ducasse:
```
(WidgetExample class>>#exampleBasicControls) messages

>>>
(#isDefault: #newListFor:list:selected:changeSelected:help: #minHeight: #vResizing: #useDefaultOKButton #newEditableDropListFor:list:getSelected:setSelected:addToList:ghostText:getEnabled:help: #newRow: #value: #red #'->' #newPluggableDialogWindow: #contentMorph: #openModal #newCheckboxFor:getSelected:setSelected:label:help: #wheel: #newMorphDropListFor:list:getSelected:setSelected:help: #extent: #newMorphListFor:list:getSelected:setSelected:help: #newLabel: #newTreeFor:list:selected:changeSelected: #@ #newButtonFor:getState:action:arguments:getEnabled:label:help: #NewValueHolder->NewValueHolder #newVerticalSeparator #builder #newTitle:for: #with: #newDropListFor:list:getSelected:setSelected:help: #newTextEditorFor:getText:setText: #newColumn: #newTextEntryFor:getText:setText:help: #currentWorld #minWidth: #setSelectedIndex: #newButtonFor:action:label:help: #newLabelGroup: #collect: #newSliderFor:getValue:setValue:help: #items: #new #disable #model: #color: #newRadioButtonFor:getSelected:setSelected:label:help:)
```

Ducasse:
In P9 and P10

Ducasse:
```
(WidgetExample class>>#exampleBasicControls) localMessages
```

Ducasse:
```
addSelectorTo: set 
	"If this instruction is a send, add its selector to set."

	| selectorOrSelf |
	(selectorOrSelf := self selectorToSendOrSelf) == self ifFalse:
		[set add: selectorOrSelf]
```

is adding global variable.

Ducasse:
Bingo, `(WidgetExample class>>#exampleBasicControls) localMessages` breaks
 
```
addSelectorTo: set 
	"If this instruction is a send, add its selector to set."

	| selectorOrSelf |
	(selectorOrSelf := self selectorToSendOrSelf) == self ifFalse:
		[
		selectorOrSelf class = GlobalVariable 
[...]

PR #10044, opened 2021-09-24
Fix InstructionStream>>nextPc:

Alisu:
In the case there was an extended bytecode nextPC: was not skipping it and return a pc in the middle of the extension.
Partially correct #9777.


src/Kernel-BytecodeEncoders/EncoderForSistaV1.class.st
```diff
@@ -418,6 +418,20 @@ EncoderForSistaV1 class >> instVarWriteScanBlockFor: varIndexCode using: scanner
...
 
+{ #category : #accessing }
+EncoderForSistaV1 class >> instructionSizeAt: pc of: aMethod [
+
+	| tempPc bytecode accumalatedSize |
+	tempPc := pc.
+	bytecode := aMethod at: tempPc.
+	accumalatedSize := 0.
+	[ self isExtension: bytecode ] whileTrue: [ 
+		accumalatedSize := accumalatedSize + (self bytecodeSize: bytecode).
```
Ducasse: accumulated

PR #10045, opened 2021-09-24
9971 cleanup ring deprecate the meta class the non meta class base class 

### LINKED ISSUE #9971: Cleanup Ring: deprecate #theMetaClass #theNonMetaClass #baseClass
**Describe the problem**
We deprecated #theMetaClass #theNonMetaClass #baseClass in the kernel, but Ring still uses these methods

**Proposal**
To keep Ring in sync with the kernel model, we need to deprecate and fix the senders.

olekscode:
Fixed #9971. Deprecated #theMetaClass and #theNonMetaClass in Ring in favor of #classSide and #instanceSide. #baseClass also needs to be deprecated and rewritten with #instanceSide but I did not understand if it is safe or not

### DIFF
--- a/src/Ring-Core/RGBehaviorStrategy.class.st
+++ b/src/Ring-Core/RGBehaviorStrategy.class.st
@@ -412,14 +412,24 @@ RGBehaviorStrategy >> storeName [
 
 { #category : #'accessing - deprecated parallel hierarchy' }
 RGBehaviorStrategy >> theMetaClass [
-
-	^ self owner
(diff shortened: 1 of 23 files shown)


olekscode:
I have opened a new issue for this sub-problem: #10046

MarcusDenker:
Failing tests are not related:
- https://github.com/pharo-project/pharo/issues/9931

MarcusDenker (approved):

PR #10048, opened 2021-09-26
Fixes #10042 enhance help system to be microdown based

### LINKED ISSUE #10042: Enhance Help System to be Microdown based
There are several ways in which this can be done. My proposal is to:

- Export the existing help into Microdown files in the doc (or docs folder)
- The status of the [NewTools-DocumentationReader](https://github.com/pharo-spec/NewTools-DocumentationReader) is that it is actually able to present Microdown files in a manner rather similar to the existing help system, so once the files exist, then can be presented.
- In the existing help system, the help topics are ordered. A simple index.txt file should allow to give both order and to load the documents from github directly. The current NewTools-DocumentationReader reads files, so some extension is needed
- Maintenance of help is done in the Microdown files in the doc folder, which gives versioning and pull-requests

I have made an exporter which seems to work. 

It required a bit of refactoring to the existing help system, and I want to hear if the above would be a way to go.
If so, I will make a PR which adds the exporter to `HelpSystem-Core` so we can start using it.

If there are other plans for the help system, please just comment and close this issue.

kasperosterbye:
Refactor of CustomHelp, added HelpExporter, and included generated help files in the doc directory.


Ducasse:
It looks good to me. 
Let us wait for the tests to run.

kasperosterbye:
As usual, interpreting the output from jenkins is beyond me. Something went wrong, but if is my code is not clear.

Ducasse:
It does not look connected to your changes :)

PR #10051, opened 2021-09-27
9947-SycSourceCodeRefactoringCommand-is-shown-in-menu 

### LINKED ISSUE #9947: SycSourceCodeRefactoringCommand is shown in menu
image

MarcusDenker:
fixes #9947 by adding the abstract check

In another pass, we could now check all overrides of canBeExecutedInContext: and remove the isAbtract check there and do a super call instead (which they often already do)

### DIFF
--- a/src/Commander-Core/CmdCommand.class.st
+++ b/src/Commander-Core/CmdCommand.class.st
@@ -99,7 +99,8 @@ Class {
 
 { #category : #testing }
 CmdCommand class >> canBeExecutedInContext: aToolContext [
-	^true
+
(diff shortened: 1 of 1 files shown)


astares:
That does not fix the issue - after applying

```Smalltalk
canBeExecutedInContext: aToolContext
	^ self isAbstract not
```
in CmdCommand the command is still there

image

astares (changes requested):
Does not fix the issue

MarcusDenker:
very odd... I  will check

PR #10060, opened 2021-09-28
Small cleanups - class side initialize and other in Monticello

### LINKED ISSUE #10017: Small cleanups - class side initialize and other in Monticello
Similar to #9882 we want to clean Monticello (classes starting with MC)

astares:
Fix #10017

### DIFF
--- a/src/Metacello-Core/MetacelloPlatform.class.st
+++ b/src/Metacello-Core/MetacelloPlatform.class.st
@@ -18,17 +18,17 @@ MetacelloPlatform class >> current [
 	^ Current
 ]
 
-{ #category : #'initialize-release' }
+{ #category : #'class initialization' }
 MetacelloPlatform class >> initialize [
   "noop ... use #select"
 ]
 
-{ #category : #'initialize-release' }
+{ #category : #'class initialization' }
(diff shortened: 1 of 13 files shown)


astares:
Chicken and egg problem - remaining tests need to fixed in Tonel - but this is managed externally

Ducasse:
This breaks tonel tests (they are probably too brittle).

astares:
@Ducasse no, they are not too brittle - the problem is that I would have to 
 - have the changes of this PR
 - adopt the tonel tests at the same time 
 but unfortunately Tonel is an external package.
 
So it can not be done **with a single PR** and therefore the PR is impossible to get green.

Ducasse:
I think that we should revisit tonel tests not to be dependent on MC package protocol.
(at least this is what I understand). 
Because once this is done we will not have this dependency anymore.

astares:
Tonel uses the mock classes defined in MC packages, which I guess is fine as they belong to standard image. 
The issue just arises as we have to adopt on two places at the same time.

Or maybe there is a way:
 - to change Tonel and provide a new release first
 - then make a PR with the changes on MC mock classes side + the new version of Tonel in one go

PR #10074, opened 2021-09-29
10073-Cleanup-class-parser-support-for-slots-in-old-style-class-def-not-needed 

### LINKED ISSUE #10073: Cleanup class parser: support for slots: in old style class def not needed
**Describe the problem**
We do not need to support slots: in old style class defs as we never print them like that anymore. Code loading is still supported as the class creation methods themselves are retained.

There is already a comment in CDClassDefinitionParser>>#parseSelectorPart:withArgument:

```
	ifTrue: [ ^ self handleSlotsNodesFromArrayNode: aNode ].
	self flag: #todo. 
	"there is no such definition anymore in a non fluid definition
	So we should clean also the test CDSlotParserArgumenting. They should use 
	the fluid parser.
```

MarcusDenker:
This PR fixes #10073:

- remove the code in CDClassDefinitionParser>>#parseSelectorPart:withArgument:
- remove the tests that test this functionality. For Fluid Class Definitions, these cases are tested with other tests


Ducasse:
I do not remember why I said that slots: was not needed anymore. I'm confused :) and tired.

MarcusDenker:
The only reason to keep it would be backward compatibility, but the class parser is not used in that case: we still keep the class creation methods with the slots: variant, which are called in that case (e.g. for file in). Monticello does not need it, we use another way to store/load slot definitions.

Ducasse:
Ah it was in old style class format, Ok I got it. 
Indeed we do not need them. I thought it was in fluid and I was puzzled.

PR #10086, opened 2021-10-01
10078-To-check-CompiledMethodlocalReadsSelf-does-isFFIMethod-check

### LINKED ISSUE #10078: To check: CompiledMethod>>#localReadsSelf does isFFIMethod check
**Bug description**
The method CompiledMethod>>#localReadsSelf:

```
localReadsSelf
	"methods that are compiled with FFI have a self in the source but not in the bytecode"
	^self isFFIMethod or: [ super localReadsSelf ]
```

The idea is that, when you search if there is a "self" accessed in the code, you want to see the ffi self sends (as they are self sends conceptionally).

But on the level of the method, it does *not* access self.

**To discuss**
#localReadsSelf should be the state of the bytecode, it should not try to be clever. Everything else we should do on another level.

e.g. a better place could be to do this in SelfVariable>>#usingMethods.

MarcusDenker:
This PR fixes #10078:

- Move the FFI check from #localReadSelf to #usingMethods of SelfVariable
- add a test and comment to  testReadsSelf. I do not execute an FFI method here to not add a dependency / make this test platform specific
- add tests for #usingMethods for self:
	- check that we can find the test method itself (with the self assert)
	- check that if we use super, we do not find it
	- add test to check for a #isFFIMethod that does not read self, but we find it still by #usingMethods of SelfVariable

### DIFF
--- a/src/Kernel/SelfVariable.class.st
+++ b/src/Kernel/SelfVariable.class.st
@@ -36,11 +36,13 @@ SelfVariable >> readInContext: aContext [
 
 { #category : #queries }
 SelfVariable >> usingMethods [
-	"as super sends are doing a pushSelf, too, we need to still check the AST level sometimes"
+	"Super sends are doing a pushSelf, too, we need to still check the AST level sometimes.
+	FFI methods after a first call do not have a self send in the bytecode, but one in the code"
 
 	^ environment allMethods select: [ :method | 
+		method isFFIMethod or: [  
 		  method readsSelf and: [ 
 			  method sendsToSuper not or: [ 
 				  method ast variableNodes anySatisfy: [ :varNode | 
-					  varNode variable == self ] ] ] ]
+					  varNode variable == self ] ] ] ] ]
 ]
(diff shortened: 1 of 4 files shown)


MarcusDenker:
Failing tests not related:
- https://github.com/pharo-project/pharo/issues/10090

src/Kernel/SelfVariable.class.st
```diff
@@ -36,11 +36,13 @@ SelfVariable >> readInContext: aContext [
...
 SelfVariable >> usingMethods [
-	"as super sends are doing a pushSelf, too, we need to still check the AST level sometimes"
+	"Super sends are doing a pushSelf, too, we need to still check the AST level sometimes.
+	FFI methods after a first call do not have a self send in the bytecode, but one in the code"
 
 	^ environment allMethods select: [ :method | 
+		method isFFIMethod or: [  
 		  method readsSelf and: [ 
 			  method sendsToSuper not or: [ 
 				  method ast variableNodes anySatisfy: [ :varNode |
```
dionisiydk, line 46: Some useful method can be extracted here: 
```Smalltalk
method ast includesVariable: self
```

PR #10088, opened 2021-10-01
10062-Cleanup-Ring-RGElementDefinitiontheNonMetaParentName

### LINKED ISSUE #10062: Cleanup Ring: RGElementDefinition>>#theNonMetaParentName should be deprecated
**Describe the problem**
 RGElementDefinition>>#theNonMetaParentName is in a category "deprecated" but not formally deprecated. It has one sender

**Proposal**
Deprecate #theNonMetaParentName and fix the one user

MarcusDenker:
This PR fixes RGMethodDefinition>>#basicAsMCMethodDefinition to not use theNonMetaParentName. As this was the only user and it was already in a deprecated category, the method was removed.

I addition, from running all Ring tests, we commit some transformations #theMetaClass to #classSide  (form a prior change that, too, brings the Ring API closer to the main model API again)

### DIFF
--- a/src/EmergencyDebugger-Tests/EDDebuggingAPITest.class.st
+++ b/src/EmergencyDebugger-Tests/EDDebuggingAPITest.class.st
@@ -51,7 +51,7 @@ EDDebuggingAPITest >> testChangeRecordsForMethod [
 	| changes |
 	self prepareMethodVersionTest.
 	changes := debugAPI
-		changeRecordsForMethod: (EDMockObjectForTests >> #mWithVersion) asHistoricalRingDefinition.
+		changeRecordsForMethod: (EDMockObjectForTests >> #mWithVersion).
 	self assert: changes size equals: 2.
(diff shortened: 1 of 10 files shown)


MarcusDenker:
two failing test are related to the change

- MCStReaderTest>>testCommentWithoutStyle
- MCStReaderTest>>testMethodWithStyle

MarcusDenker:
Fixed failing tests.. they expose a real bad bug in #methodClass on Ring1 method objects... #methodClass was escaping the model, but it should return the Ring parent!

MarcusDenker:
Now EDDebuggingAPITest>>testChangeRecordsForMethod is failing

MarcusDenker:
This is not that easy...  While looking at #instanceSideParentName, I realized that it is common to create Ring1 methods that are not connected to a Ring class .  I will for now close it and to the deprecated renaming change first, and unify theNonMetaParentName with instanceSideParentName (which was just a rename). #instanceSideParentName after then needs some thought.

PR #10092, opened 2021-10-01
10067-CDClassDefinitionParserparseSelectorPartwithArgument-misses-subclass-definition-symbols

### LINKED ISSUE #10067: CDClassDefinitionParser>>#parseSelectorPart:withArgument: misses subclass definition symbols
**Bug description**
CDClassDefinitionParser>>#parseSelectorPart:withArgument:  hard - codes the subclass defintion symbols. This way some are missed (variableDoubleWordSubclass and variableDoubleByteSubclass).

** What needs to be done**

- add missing tests for variableDoubleByteSubclass: and variableDoubleWordSubclass:
- use #subclassDefiningSymbol in the Layout Hierarchy to get a complete list of all symbols instead of hard-coding

MarcusDenker:
This PR fixes #10067

The problem was that CDAbstractClassDefinitionParser did not know all subclass definition symbols. variableDoubleByteSubclass: and variableDoubleWordSubclass: were missing.

- we add tests for the two cases (which fail)
- we add #allSubclassDefiningSymbols that gets all the defined symbols from the layout classes

Now we can just use #allSubclassDefiningSymbols instead of hard coding the symbols and the tests are now green!

### DIFF
--- a/src/ClassParser-Tests/CDDoubleByteClassParserTest.class.st
+++ b/src/ClassParser-Tests/CDDoubleByteClassParserTest.class.st
@@ -1,0 +1,27 @@
+Class {
+	#name : #CDDoubleByteClassParserTest,
+	#superclass : #CDClassDefinitionParserTest,
+	#category : #'ClassParser-Tests'
+}
+
+{ #category : #helpers }
+CDDoubleByteClassParserTest >> classDefinitionString [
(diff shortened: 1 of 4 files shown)


Ducasse:
Cool that you improve it.

PR #10107, opened 2021-10-05
Remove the stupid rule about the instance creation protocol.

Ducasse:


astares:
Why do you declar it as a stupid rule? We should keep it as it serves a purpose.

We had some people using "instance-creation" on a few places instead of the regular "instance creation" and we fixed these cases.  
The tests make sure they do not get accidentially reintroduced - give an array of wrong examples and how the category should be named.

image

Ducasse:
Because the rule is stupid. 
We should NOT change instance creation into instance-creation. Period.
And we are fed up of rules and a system that does not help us. 
See my email in the Pharo about Cops vs. Help.

We do not want rules without any help to fix them.
We are fedup to have more to do like rename utils instead of utilities both are CORRECT
and having both is ok. 

I'm working on hector a tool to FIX for us such stupid conventions
and others like unused instance variables. Why a program cannot remove unused instance variables
why am I to do it?

MarcusDenker:
But it checks the opposite: that nobody adds instance-creation (because the dash is not a good idea)

Ducasse:
ok but still do we have to redo a PR just for that?

Ducasse:
I do not want to code for Pharo anymore. Too boring.

astares:
@Ducasse You have seemed to judged too early - as Marcus said (and I tried to explain): it checks exactly the opposite. 

So it follows our usual pattern of cleaning and making sure with tests that the "bad" things do not get reintroduced and we can keep our quality level (or increase it constantly). So the test definitely makes sense and we should not remove it.

There is no PR or change needed.


----------------------------------------------------------------------------------------------------
**Side note:**  
I do not know what Hector is about. But personally I think we do not need yet another new tool. We have already rules displayed in browser. So the user can directly see critiques already. We just need more alignment between the rules and the final release tests and people should learn not to ignore the code critiques.

As an info:
What seems to be little known is that it is possible to also provide an "autofix" for a code critique/check - so if a rule is displayed it can be automatically fixed with a single click (if provided and implemented for the critique) 

So if a Renraku rule returns true for #providesChange then the Calypse command class "ClyFixCritiqueCommand" sends #change 
to the rule to let it implement an "autofix" for the rule. So there is some support already and I remember that it was used in the past. 

The class comment in the root class ReAbstractCritique says:

```
A critique has the #providesChange method which returns a boolean value specifying whether the critique can provide a change which will resolve the issue. The #change method can be used to obtain an object of RBRefactoryChange kind.
```

Still this mechanism seems to be incomplete a little bit (the announcement "ReCritiqueAutofixed" is there but not sent while "ReCritiqueBanned" is).

So before we burn our cycles on the next fancy tool we should reuse and focus more on using the already existing (maybe unfinished) infrastructure. Except we want to throw our Renraku infrastructure and replace it with the next idea.

astares:
> I do not want to code for Pharo anymore. Too boring.

Yes - getting more disciplined, more rule based, more consistent in what Pharo and its code looks like and 
more serious in all efforts and plans can be tough and hard. And annoying - I know. Sure it can also lead to a more  
exhausting or even boring day. 

But we have to raise the quality level (and test that we keep it) to set the bar stepwise higher to play in the league of  serious and professional tools. Remember that we dont want to go back to ugly and a mess in the code but funny toys. This was the past.

Don't let a mood hit the mind or influence decisions. Take a break - look what has been achieved already - there are already many fruits from all the work. And there is more to invent for the future.

I hope (and I'm pretty sure) your fun returns...

astares:
BTW: I found out where the Autofix of the Renraku "critiques" infrastructure could be applied:

Lets make an example:

 - create a class "ClassA"
   - create a class "ClassB" as subclass of "Class A" with a method #someMethod to return $b
   - create a class "ClassC" as subclass of "Class A" with a method #someMethod to return $c
   
 So both subclasses define an equal method - but the superclass not.
 
Now Renraku/Calypso not only show the lint message but also offer a button to fix it "automagically":
  
image
 
So one can fix the issue with a single click on the button labeled "Automatically resolve the issue".  

We could do similar things for fixing other issues in the image (like wrong categories) by using the already provided infrastructure and mechanisms of Rules and Renraku. If more rules would provide an "Autofix" it would be also easier for newbees.

Also the user does not need to click each and every lint issue resolver. If one open the code critique browser there is
an "Update" button which is sending the #update to the rule that provide an automatic repair.

I guess the button should have better be called "Autofix" instead of just "Update". It sends the #update message to every rule who provides an automatic fix.

image

Check #ReMissingSubclassResponsibilityRule (which is using ReMissingMethodCritique from Renraku) for such an example implementation.

Ducasse:
You do not understand anything in fact. This is sad.

Ducasse:
I will release a version of the new documentation reader and stop. 
I will make sure that none of my future projects has anything related to Pharo core. 
I will not touch Pharo. And bet that I will succeed...
Pharo is full of crap but for sure let us annoy people with categorisations that we can fix massively
but this is probably more exciting to do them one by one. 
Let us not improve the code formatter it is better to do it by hand.

astares:
You wrote that this piece of code (quote) "_change instance creation into instance-creation._" (/quote) which 
it is definitely not doing. Its not changing categories, its a test method: it just verifies that we stay cleanly 
with "instance creation" and not accept bad namings like "instance-creation". 

You now lament although the misunderstanding was on your side. You called things already "stupid" in the topic without
giving a clear description of what you see as being wrong, having a closer look at the code or reading properly the 
first responses (who explain it already). 

And finally you give the impression to try to impose your opinion with a "threatening" gesture to quit if others do not share the same thoughts?! Sorry - but such a feedback did not help in any way on the issue discussed here!

For the mentioned reasons I still propose that we close that PR without integrating it. Additionally I will try to refactor these 
tests to use better names so it is more obvious and we can become more intention revealing. We could for instance make it 
less confusing by writing:

```Smalltalk
testProperInstanceCreationProtocolIsUsed

  self assertProtocolName: #'instance creation' notAcceptingProtocolNamesLike: #(#'instance-creation') 
```


Regarding your other comments:
Do you really think it is nice to constantly work on thousands of small BORING and often obvious cleanups while 
others can use their time to work on more exciting tasks or tools? It's also NOT FUN to clean up the room again and again 
when others can make the same room dirty again and again at the same time. 

We want to improve and not move backwards: so we have release tests to ensure that we keep that status quo - also for 
categories:
- people do not get lost anymore because methods are uncategorized or categorized differently each time with
  stupid names or typos
- it helps people find they way more easily (and also saves time) when one can expect for instance a #setUp/#tearDown 
  methods in protocol "running" in every test class, ...


I also see no real contradiction in our views/goals: the points you raised and mentioned in the post to the mailinglist are 
TOTALLY VALID. I agree with you by 100% that the system should not only have rules but also support the users 
to easily spot violations and support to quickly bringing things back into good shape.

So where it get's annoying is especially:
 - when we introduce now more release tests (who are running late in the process)
 - but we do not show the same issues as lint problems in the browser (and also do not give people the chance and tool to easily fixe them when they appear)

This is a point we need to address and work on and I share your thoughts that we need better tool support on this
topic directly in Pharo. So we need a way to easier fix violations of given rules and as you see from my last comments: 
in Renraku there is already infrastructure available and it is now up to us all to start using it even more for the better.
 
I try to allocate some time to look into this "autofixing" issues more deeply so we all get less annoyed...

astares:
So lets start: here is a way to let people quickly "autofix" the issue with unused ivars:  https://github.com/pharo-project/pharo/issues/10125

astares:
And the next one to improve and make more transparent what the test is doing: https://github.com/pharo-project/pharo/pull/10132

astares:
@Ducasse thanks for closing.

I took a few minutes to provide another proposal to improve the situation:

https://github.com/pharo-project/pharo/issues/10139 with PR in https://github.com/pharo-project/pharo/pull/10140

astares:
Related: #10170 (a refactoring to bring the rules and tests into alignment)

astares:
And more to cleanup and improve the situation: see #10226

PR #10111, opened 2021-10-05
10110-isReferenced-is-not-working-for-PoolVariables

### LINKED ISSUE #10110: #isReferenced is not working for PoolVariables
**Bug description**
isReferenced is just searching the Pool defining class, not the user 

**Expected behavior**
It should check if this is a Pool var and if yes, fall back to the slow super class implemention

MarcusDenker:
fixes #10110

- add a test for IsReferenced for the case of Pool variables
- add a comment to both usingMethods and isReferenced explaining why we treat pools specially

We could speed up the search for pools to look in the pool users in addition to the definer if we see that it is too slow.

### DIFF
--- a/src/Kernel/ClassVariable.class.st
+++ b/src/Kernel/ClassVariable.class.st
@@ -75,8 +75,9 @@ ClassVariable >> isPoolVariable [
 
 { #category : #testing }
 ClassVariable >> isReferenced [
+	"For pools, fall back to the slow full search of all methods"
+	self definingClass isPool ifTrue: [ ^ super isReferenced ].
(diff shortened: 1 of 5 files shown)


MarcusDenker:
For the speedup I opened https://github.com/pharo-project/pharo/issues/10113, we need to do it as the #testNoUnusedClassVariablesLeft is now too slow. But it more important to merge this first so we do not block the PR for the "senders of" improvement.

dionisiydk (approved):

guillep:
Thanks Marcus, I'll produce a backport of this, otherwise Spec builds will never work :)

PR #10115, opened 2021-10-06
Deprecation-Rewrites-Ring 

MarcusDenker:
These are some automatic rewrites triggered by running all the Ring tests


MarcusDenker:
as these are just automatic rewrites, I take the liberty to merge the PR even though I did it

PR #10126, opened 2021-10-06
Provide autofix rule to have no unreferenced ivars left

### LINKED ISSUE #10125: Provide autofix rule to have no unreferenced ivars left
Provide an ReNoUnusedInstanceVariableRule that can autofix

astares:
fix #10125


MarcusDenker:
In a second step, we then can remove ReIvarNeitherReadNorWrittenRule which does the same but without repair (and the new name is better)

MarcusDenker (approved):

astares:
@MarcusDenker maybe we keep ReIvarNeitherReadNorWrittenRule and provide a different autofix: generate the accessors.

Then the user can decide for the variable:
 - if it is still required then autofixing to accessors make sense
 - if it is not required (anymore) then autofixing to removal makes sense

PR #10128, opened 2021-10-07
Backport #10111: isReferenced-is-not-working-for-PoolVariables

guillep:
Backport #10111

This is required for the correct behavior of latest spec/newtools (and not breaking its tests)

- extend isReferenced for class pool variables
- adding comments to usingMethods
- tests


MarcusDenker (approved):

MarcusDenker:
testNoUnusedClassVariablesLeft might now time out more often (on Pharo10 is seems to be just under the timeout treshold, just due having less class variables...), but I think we should merge and I fix that next.

PR #10130, opened 2021-10-07
enabling-debug-window-in-windows-P10

tesonep:
Adding hook to configure additional options in Windows.
This might fail if the VM is not prepared, so it is handling that scenario.
Part of pharo-project/opensmalltalk-vm#125

### DIFF
--- a/src/OSWindow-SDL2/SDLWindowsPlatform.class.st
+++ b/src/OSWindow-SDL2/SDLWindowsPlatform.class.st
@@ -7,7 +7,7 @@ Class {
 	#category : #'OSWindow-SDL2-Bindings'
 }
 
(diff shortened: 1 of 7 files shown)


src/OSWindow-SDL2/SDLWindowsPlatform.class.st
```diff
@@ -7,6 +7,24 @@ Class {
 	#category : #'OSWindow-SDL2-Bindings'
 }
 
+{ #category : #'as yet unclassified' }
```
Ducasse: Pablo you have an unclassified

tesonep: Thanks Stef, I have fixed it.

PR #10137, opened 2021-10-08
[P9] 10085-Pharo-does-not-work-on-network-drive-on-Windows

tesonep:
Adding support for UNC Network paths.
They are required to correctly support window shares (SMB).

### DIFF
--- a/src/FileSystem-Path/Path.class.st
+++ b/src/FileSystem-Path/Path.class.st
@@ -117,7 +117,10 @@ Path class >> from: aString delimiter: aDelimiterCharacter [
 	
 	pathClass :=  ((self isAbsolutePath: aString delimiter: aDelimiterCharacter) or: 
 							[self isAbsoluteWindowsPath: aString]) 
-		ifTrue: [ AbsolutePath ]
+		ifTrue: [ 
+			((aString size >= 2) and: [ aString second = aDelimiterCharacter ])
+				ifTrue: [ NetworkPath ]
+				ifFalse: [AbsolutePath] ]
 		ifFalse:[ RelativePath ].
 	
 	^ pathClass withAll: (aDelimiterCharacter split: aString)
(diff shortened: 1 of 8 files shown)


src/FileSystem-Path/Path.class.st
```diff
@@ -117,7 +117,10 @@ Path class >> from: aString delimiter: aDelimiterCharacter [
 	
 	pathClass :=  ((self isAbsolutePath: aString delimiter: aDelimiterCharacter) or: 
 							[self isAbsoluteWindowsPath: aString]) 
-		ifTrue: [ AbsolutePath ]
+		ifTrue: [ 
+			((aString size >= 2) and: [ aString second = aDelimiterCharacter ])
+				ifTrue: [ NetworkPath ]
+				ifFalse: [AbsolutePath] ]
```
guillep, line 123: Maybe for a next iteration, but, shouldn't this be dispatched back to the store? Because the NetworkPath is a windows concern, right?

src/FileSystem-Path/Path.class.st
```diff
@@ -137,12 +140,15 @@ Path class >> isAbsoluteUnixPath: aString [
...
-	^ ((aString size > 2) and: [ aString first isLetter ])
-		ifTrue: [ (aString second = $:) and: [aString third = $\] ] "e.g. C:\..."
-		ifFalse: [ false ]
+	(aString first = $\ and: [ 
+		 aString size = 1 or: [ aString second ~= $\ ] ]) ifTrue: [ ^ true ]. "e.g. \file"
 
+	^ (aString size > 2 and: [ aString first isLetter ])
+		  ifTrue: [ "e.g. C:\..." 
+		  aString second = $: and: [ aString third = $\ ] ]
+		  ifFalse: [ false ]
```
guillep, line 151: Same, shouldn't this be dispatched to the store?

guillep (approved):

src/FileSystem-Path/Path.class.st
tesonep: Yes, that can be nicer. I will do that change in P10.

PR #10138, opened 2021-10-08
10085-Pharo-does-not-work-on-network-drive-on-Windows-P10

### LINKED ISSUE #10085: Pharo does not work on network drive on Windows
**Bug description**
When Pharo VM is run from network drive, it is not able to find any file, including .changes, .sources, .dll etc.
This is an issue on company/university systems with user homes on network drives.
Same files work properly when copied to C drive or flash drive (on the same PC).

Pharo3
Pharo
Pharo2

**Version information:**
 - OS: Windows 10
 - Pharo 9 and 10 (stable and latest VM)

tesonep:
Adding support for UNC Network paths.
They are required to correctly support window shares (SMB).

Fix #10085

### DIFF
--- a/src/FileSystem-Path/NetworkPath.class.st
+++ b/src/FileSystem-Path/NetworkPath.class.st
@@ -1,0 +1,68 @@
+"
+I am a path to express UNC paths e.g.,	
+	#( '\\MachineName.x.y.z\directory\file' '\\MachineName.x.y.z\directory' '\\MachineName.x.y.z' ) .
+
+This is useful to represent files in SMB shares as used by Windows or Samba.
+
+A NetworkPath acts as an absolute path but with a different translation to string.
+"
+Class {
+	#name : #NetworkPath,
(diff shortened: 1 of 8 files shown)


src/FileSystem-Path/NetworkPath.class.st
```diff
@@ -0,0 +1,68 @@
+"
+I am a path to express UNC paths e.g.,	
+	#( '\\MachineName.x.y.z\directory\file' '\\MachineName.x.y.z\directory' '\\MachineName.x.y.z' ) .
+
+This is useful to represent files in SMB shares as used by Windows or Samba.
+
+A NetworkPath acts as an absolute path but with a different translation to string.
+"
+Class {
+	#name : #NetworkPath,
```
dionisiydk: Maybe UNCNetworkPath would be a better name. The current one sounds too generic

dionisiydk (approved):
that's nice

PR #10140, opened 2021-10-08
Provide Autofix for ReProperMethodProtocolNameRule

### LINKED ISSUE #10139: Provide Autofix for ReProperMethodProtocolNameRule
The system provides some rules for proper categorization and to avoid bad category names. This is shown in browser
as lint rule and provided in class ReProperMethodProtocolNameRule and subclasses.

image

Additionally as part of the release tests we check that at least for the image that rules is followed so we have a
clean and consistent naming and also do not reintroduce cases with bad names for categorization (this is ensured using ProtocolConventionsTest).

While we have these mechanisms in place it can be annoying: as the release tests runs late in the integration process one needs to fix these cases already when coding a change / PR.

Unfortunately as one can see on the first screenshot these violations are shown as "infomations" for severit - not as real "warnings". Also one has to fix them by manual recategorization - there is no support from the system side yet to quickly fix them.

See also 
 - discussion in https://github.com/pharo-project/pharo/pull/10107
 - post of Stef to ML: https://lists.pharo.org/empathy/thread/BUJNB3ZRHWGOFZ3RSDMDCP5EJHGJANFU

astares:
Fix #10139 
- remove "information" severity (so that the default "warning" is used)
- improve the lint message by shown current and suggested protocol name
- provide an autofix so code can easily repaired to follow lint rul
- move all #protocolIdioms to the class side (so we can later better reuse them in the release tests)

### DIFF
--- a/src/Renraku/ReProperMethodProtocolNameForAccessingRule.class.st
+++ b/src/Renraku/ReProperMethodProtocolNameForAccessingRule.class.st
@@ -8,7 +8,7 @@ Class {
 }
 
 { #category : #idioms }
-ReProperMethodProtocolNameForAccessingRule >> protocolIdiom [ 
+ReProperMethodProtocolNameForAccessingRule class >> protocolIdiom [ 
 
 	^self use: 'accessing' insteadOf: #('accessor' 'accessors' 'acessing' 'acccessing' 'accesing' 'acesing')
 ]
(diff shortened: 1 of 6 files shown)


astares:
CI problems unrelated. Maybe it is possible to restart CI?

MarcusDenker (approved):

PR #10164, opened 2021-10-13
Migrating RBRenamePackage into Refactoring2

miguelcamp:
Using a composite transformation since RBCopyPackage will be composite and this will preserve the same hierarchy


demarey:
Class comments are missing.
Could you add them? Else, it looks ok for me.

miguelcamp:
I have added the missing class comments. Sorry about the delayed response

Ducasse:
Thanks!

demarey:
Thanks Miguel

PR #10182, opened 2021-10-17
Rule to check class side #reset methods are in 'class initialization' protocol

gypsydave5:
Relating to #2607, this is a start on writing a Renraku rule. I'd like some feedback and advice!


gypsydave5:
I've just added an automated fix for this rule

MarcusDenker:
One failing test: testAndMakeSureSuperTearDownIsCalledAsLastMessageInTearDownMethodsOfTestCases (missing "super tearDown"

gypsydave5:
Bah - I remember thinking about that. Thanks Marcus!

gypsydave5:
@MarcusDenker - just a qq - what tests should I be running locally so that I don't do this again?

MarcusDenker:
We need to add an easy way to run a subset of tests, e.g. by adding a button to the git commit UI.

MarcusDenker (approved):
ok

For the instance side, #reset can have many meanings, but class side it is good to have it in a defined category.

guillep (approved):
Good

PR #10183, opened 2021-10-18
Use-equivalentTo-ReEquivalentSuperclassMethodsRule 

MarcusDenker:
This PR changes ReEquivalentSuperclassMethodsRule and testNoEquivalentSuperclassMethods to use #equivalentTo:

 #equivalentTo: in the end compares the AST, too, but it first does some quick checks to avoid that in most
 cases (numArgs and numLiterals).


demarey:
It is just a quick fix for the rule. #equivalentTo: has more problems that should be addressed: see issue #10184

PR #10186, opened 2021-10-18
Enable-SharedLiterals

### LINKED ISSUE #10187: testMutateVariableObject modifies literals
**Bug description**
testMutateVariableObject uses literals to test.. which means that it changes the literals in the method.

Worse, as soon as you share literals, it modifies #(1 2 3) which is used 162 times in the system.

**What to do**
Just copy the literal...
```guineaPigs := {#[1 2 3] copy . #(1 2 3) copy}.```

MarcusDenker:
This PR enables literal sharing. As part of #cleanUpForRelease:
 -  we create a table with all the literals (even recursing into literal arrays). 
 -  then we make sure that we use this one literal in all methods.

For now, the last step does not recurse into arrays (TBD).


guillep (approved):
Looks ok!

MarcusDenker:
For the next step I added https://github.com/pharo-project/pharo/issues/10190

Issue #8170, opened 2020-12-24
Pharo image freeze on Windows after some hours Idle


VincentBlondeau:
**Describe the bug**
After some time keeping open a pharo image with the headless image, the image becomes non-responsive but is still open.

The error is:
Error: Bad BitBlt arg (Fraction?); proceed to convert.
[dump.txt](https://github.com/pharo-project/pharo/files/5739298/dump.txt)

**To Reproduce**
1. Open an image
2. Wait some time
3. See that it is frozen

**Expected behavior**
No freeze

**Version information:**
VM: Win32 - x86_64 - Win64 - CoInterpreterWithQueueFFI VMMaker-tonel.1 uuid: be795315-f06d-0d00-aab5-813a0e395221 Oct 13 2020
StackToRegisterMappingCogit VMMaker-tonel.1 uuid: be795315-f06d-0d00-aab5-813a0e395221 Oct 13 2020
e829a1da - Commit: e829a1da - Date: 2020-10-13 15:43:09 +0200

Image: Pharo9.0.0 [Build information: Pharo-9.0.0+build.997.sha.5bf82e9abc5aea283f18e74bbe27ee9464483ea9 (64 Bit)]

**Expected development cost**
No clue

**Additional context**
It may be due also to a windows session locking

VincentBlondeau:
I ran the WinDbg utility on the frozen image for each running thread. Here is the the result:
[WinDbg-SDLCrash.txt](https://github.com/pharo-project/pharo/files/5763965/WinDbg-SDLCrash.txt)

I try to let the image still in a freeze mode, let me know if other commands should be run.

tesonep:
Thanks for the debug info @VincentBlondeau 
I will check why we are not getting SDL events.

VincentBlondeau:
I can reproduce the issue when the Windows user session got locked and when I login in, the image freeze.
Let me know if I can help you with more debugging info.

NB: this issue will be critical to be fixed for Lifeware.

Ducasse:
We will fix it :)

fxgallego:
I can't reproduce the same with Pharo 9 but something similar happens when you unplug one display if you have more than one connected. P9 window is still visible but hang and doesn't redraw anything. In my case it doesn't depends on which display the Pharo windows is, it alway hangs when you unplug one. Interesting that the Pharo Launcher doesn't shows the same behavior.
Having the image open for days, locking, unlocking windows session, even suspend works ok, but unplug display did the trick.

[...]

PR #10188, opened 2021-10-18
[P10] 8170-Pharo-image-freeze-on-Windows-after-some-hours-Idle-P10

tesonep:
Adding a new event that records changes in the display or resolutions.
This is needed to correctly handle Remote Desktop connection and Disconnection.

Adding in the drawing of the renderer because we need to ensure that the error does not destroy everything.
The problem is if we have the exception in a mutex, and the mutex is not liberated.

Spliting the initialization of constants.
The Pharo 9 bootstrap is still building the image first in V3 bytecode and the number of literals is limited.
These methods can be inlined when V3 bytecode is left behind in the bootstrap.


tesonep:
Fix #8170 for P10

demarey:
@VincentBlondeau Would it be possible for you to test the fix and tell us if everything is now ok?

PR #10195, opened 2021-10-19
10190-Literal-Sharing-unify-nested-arrays

### LINKED ISSUE #10190: Literal Sharing: unify nested arrays
If you look at ImageCleaner #shareLiterals, it recurses into nested literal arrays to build the table.

But then it only unifies the literals of methods using the table. 

We should in addition go over the arrays recursively and unify the contents using the table.

MarcusDenker:
This impoves literal sharing: unify all the nested arrays using the table that we created.

fixes #10190

### DIFF
--- a/src/Kernel-Tests/ObjectTest.class.st
+++ b/src/Kernel-Tests/ObjectTest.class.st
@@ -42,7 +42,7 @@ ObjectTest >> testBeRecursivelyReadOnlyObject [
 	self should: [ assoc key: 2 ] raise: ModificationForbidden.
 	self should: [ assoc value key: 2 ] raise: ModificationForbidden.
 	
-	array := #(1 #(1 2)).
+	array := Array with: 1 with: (Array with:1 with: 2).
 	array beRecursivelyReadOnlyObject.
 	self should: [ array at: 1 put: 2 ] raise: ModificationForbidden.
 	self should: [ array second  at: 1 put: 2 ] raise: ModificationForbidden.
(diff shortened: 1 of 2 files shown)


svenvc:
Any idea on how much memory is being recycled this way ?

In any case, it is impressive that this is possible. Little steps multiplying.

MarcusDenker:
It seems the image ~ 2 MB smaller with just doing literal sharing as it we merged today. I did not test this additonal improvement yet.

MarcusDenker:
(but image size is not a good measure.. it seems to be a bit different each time and of course loading/deleting code change sit all the time) but gives just some impression)

MarcusDenker:
Hmm.... Tests show that there are problems

MarcusDenker:
Looking at the tests, it seems that #(1 2) is changed to #(2 2) somewhere. This array has to nested in another array (else we would the the problem already with the current version).

svenvc:
but wouldn't the readonly property guard against that modification ?

MarcusDenker:
For the version alrady integrated I found a test that tested read-only itself that was making a literal read-write and changed it.

But for this case, my theory is that it might be some FFI related test.

I need to check all nested literal arrays that contain #(1 2), maybe it will be obvious where it happens.

MarcusDenker:
I was testBeRecursivelyWritableObject

MarcusDenker:
This is now working, I merge it

PR #10199, opened 2021-10-20
Changing uses: and trait: to use traits: in the class definition.

tesonep:
Changing all uses of trat: and uses: to traits:


demarey:
@tesonep @Ducasse I'm surprised to see this change since the related issue #10212 is still under discussion.
It looks strange to use `traits` for a trait composition.

Ducasse:
for the user we a class can have a list of traits. 
It does not have to know that this is in fact a single trait composition. 
So to be consistent and simple we decided with pablo to use traits: {} and we got the same than slots: {}

PR #10209, opened 2021-10-20
Remove class of the RBSelectorEnvirionment if there is no more selectors associated to the class

VincentBlondeau:
For a RBSelectorEnvironment, #classes should not contains the class anymore


VincentBlondeau:
1 Failure that is not related:
osx-64 / Tests-osx-64 / testAnnounceQueryEndedIsSentOnce – MacOSX64.NewTools.Spotter.Tests.StSpotterModelTest

VincentBlondeau:
I see that the diff is just unreadable... Is there a parameter to set so the commit is github compliant?

PR #10218, opened 2021-10-21
9922-RenderBugsTesttestForward-checks-for-timing-leads-to-timeouts-on-slow-CI 

MarcusDenker:
I propose to remove the tests... unit tests are good, but tests that just check for a specific fixed bug without being structured as unit-tests are not that of a nice idea..

fixes 9922


MarcusDenker:
Just the spotter test timing out on mac, else green

MarcusDenker:
This PR fixes a timeout, normally I merge those without a review if they take >1 week. But this I opted to be a bit radical (too much?) and removed the whole class. 

 It would be nice to have a review from someone else, too.

guillep:
I don't think these tests are good. The only one that looks good is the stickiness one...

MarcusDenker:
We can just move that over to MorphTest, it only uses the teardown to do "m topRendererOrSelf delete", which it already does at the end.

```
testForTiltedStickyness

	| m |
	m := Morph new openCenteredInWorld.
	self assert: m topRendererOrSelf isSticky not.
	m beSticky.
	self assert: m topRendererOrSelf isSticky.
	m addFlexShell.
	m topRendererOrSelf rotationDegrees: 45.0.
	self assert: m topRendererOrSelf isSticky.
	m beUnsticky.
	self assert: m topRendererOrSelf isSticky not.
	m topRendererOrSelf delete.
```

So we can merge this PR and then commit the testForTiltedStickyness as a test for MorphTest

MarcusDenker:
I added a PR for the test that makes sense as #10351

Issue #10211, opened 2021-10-20
addSearch: -> addRule:???


Ducasse:
- `addSearch: anAssociation` does not exist.
- It looks like it could simply be implemented as: 

```
addSearch: anAssociation
 self addRule: (RBSearchRule searchFor: anAssociation key thenDo: anAssociation value)
```

We should check all the senders of addSearch: 
```
longMethods
	| detector matcher |
	detector := self new.
	detector name: 'Long methods'.
	matcher := self parseTreeSearcher.
	matcher
		addSearch: '`.Stmt' -> 
					[:aNode :answer | 
					(aNode children inject: answer
						into: [:sum :each | matcher executeTree: each initialAnswer: sum]) + 1].
	detector methodBlock: 
			[:context :result | 
			(matcher executeTree: context parseTree initialAnswer: 0)
				>= self longMethodSize
					ifTrue: [result addClass: context selectedClass selector: context selector]].
	^detector
```

Ducasse:
`addSearches:` should be added too as a simple loop calling addSearch:
Adding some tests would be good too.

PR #10220, opened 2021-10-21
fix: #10211 addSearch: -> addRule: adding two little tests to start

alesshosry:
Started to add some tests and method definition for missing methods.

### DIFF
--- a/src/AST-Core-Tests/RBParseTreeSearcherTest.class.st
+++ b/src/AST-Core-Tests/RBParseTreeSearcherTest.class.st
@@ -116,3 +116,21 @@ RBParseTreeSearcherTest >> testSearchingComposedSelector2 [
 					initialAnswer: 0)
 		equals: 1
 ]
+
+{ #category : #'tests - add search' }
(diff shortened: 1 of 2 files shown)


guillep:
Hi @AlessHo,

For documentation purposes, could you give some detail in the PR on:
 - what tests did you add?
 - why are they important?
 - Are there follow-ups expected?

Ducasse:
We should go over RBDummyLint Test class side methods and check how we can turn them into executed tests.

Ducasse:
- there are two tests in the PR. :)

alesshosry:
Hi @guillep ,

1. 2 new tests are added under RBParseTreeSearcherTest subclass: **testSimpleAddSearch** and **testSimpleAddSearches**.
2. The purpose of adding these tests is to test **addSearch** and **addSearches** methods that were missing and added recently under RBParseTreeSearcher subclass. 
3. For the moment no more changes will be implemented, I will be checking the RBParseTreeSearcher and how it works ... then I will add some comments etc ... before implementing the same on RBDummyLint.

PR #10222, opened 2021-10-21
9973-Different-behavior-in-Playground-for-deprecated-method-transformed-vs-nontransformed-deprecations

### LINKED ISSUE #9973: Different behavior in Playground for deprecated method (transformed vs. nontransformed deprecations)
1. Create a simple class Foo

```Smalltalk
Object subclass: #Foo
	instanceVariableNames: '' 
	classVariableNames: ''
	package: 'MyKillerApp'
```

2. Now implement on class side a **class method with a deprecation** like:

```Smalltalk
foo

	self deprecated: 'this is deprecated'
``` 	
When you interactively evaluate this in a Playground you get a **deprecation notification** and can step into the debugger.
image

Now when you implement the method using a **deprecation with a transformation** then

```Smalltalk
foo

       self deprecated:  'Please use #bar instead' transformWith:  '`@receiver foo' 
						-> '`@receiver bar'
```
then it will just run **without any depreaction notice**

image

For sure the transformation catches the signal to perform the task of changing the sender method.

But in interactive playing in Playground it might be wished to also get a deprecation notification for the 
methods with a transformation.

MarcusDenker:
fixes #9973 as described in the issue tracker entry

### DIFF
--- a/src/Kernel/Deprecation.class.st
+++ b/src/Kernel/Deprecation.class.st
@@ -232,7 +232,7 @@ Deprecation >> transform [
 	self shouldTransform ifFalse: [ ^ self signal].
 	self rewriterClass ifNil:[ ^ self signal ].
 	aMethod := self contextOfSender compiledCode method.
-	aMethod isDoIt ifTrue:[^ self]. "no need to transform doits"
+	aMethod isDoIt ifTrue:[ ^ self signal ]. "no need to transform doits"
 	node := self contextOfSender sourceNodeExecuted.
 	RecursionStopper during: [
 		rewriteRule := self rewriterClass new 


astares (approved):
Yes works fine.

When calling

```
Foo foo
```
interactively it throws the deprecation.

But when having a Bar>>something like

```Smalltalk
something
     Foo foo
```
and calling it in Playground using

```Smalltalk
Bar something
```
it correctly gets rewritten into:

```Smalltalk
something
     Foo bar
```
so I guess this is expected and acceptable.

astares (approved):
@MarcusDenker yes - looks good

PR #10224, opened 2021-10-21
Remove empty protocols in ZnStaticFileServerDelegate

### LINKED ISSUE #10223: Remove empty protocols in ZnStaticFileServerDelegate
image

unneeded:
- "accessing - deprecated parallel hierarchy"
- "accessing - deprecated parallel hierarchy " (with trailing space)

both are without any method

astares:
Fix #10223

@svenvc

### DIFF
--- a/src/Kernel/ClassDescription.class.st
+++ b/src/Kernel/ClassDescription.class.st
@@ -175,7 +175,7 @@ ClassDescription >> authors [
 	^(self methods, self classSide methods) collect: [ :each | each author ] as: Bag
 ]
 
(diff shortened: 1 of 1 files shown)


svenvc:
I never wrote, nor never saw such silly protocol names.

IOW, I did not do this.

MarcusDenker:
Interestingly, the PR just touches ClassDescription. Maybe the space confuses the system...

astares:
Yes - and the class File in Tonel has no method category like that. Strange. 

Lets see what happens when the PR is integrated, then at least there should be only one category.

PR #10227, opened 2021-10-21
Provide rule to have no leading or trailing space in method category name

### LINKED ISSUE #10226: Provide rule (and autofix) to have no leading or trailing space in method category name
We have methods where the category name has accidentially has a leading or trailing space
in the method category name.

So we should provide a lint rule that shows these cases and also allows to autofix them:

```
Object allSubclasses select: [:each | 
	each organization allCategories 
		anySatisfy: [:cat | (cat endsWith: ' ')
			or: [ cat endsWith: ' ' ]]		
]
```

one is for instance fixed in https://github.com/pharo-project/pharo/pull/10224/files of PR #10224)

astares:
Fix #10226

### DIFF
--- a/src/GeneralRules/ReNoSpaceAroundMethodCategoryNameRule.class.st
+++ b/src/GeneralRules/ReNoSpaceAroundMethodCategoryNameRule.class.st
@@ -1,0 +1,56 @@
+"
+Provide rule to have no leading or trailing space in method category name
+"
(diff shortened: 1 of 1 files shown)


astares:
The two failing tests are totally unrelated (and only fail on Mac)

MarcusDenker (approved):

PR #10233, opened 2021-10-21
Release test to ensure we have no protocol names with leading or trailing spaces

### LINKED ISSUE #10232: Release test to ensure we have no protocol names with leading or trailing spaces
When #10224 and #10230 (who fix protocol names with leading or trailing spaces) we should add a release
test to make sure we keep the current status and do not introduce such cases again.

astares:
Fix #10232


astares:
Can only become green when https://github.com/pharo-project/pharo/pull/10230 has been merged first

kasperosterbye:
I like this one. I assume there was a problem from time to time, and they are hard to eyeball. There are no instances in the P10 image. I vote for merge.

MarcusDenker (approved):

PR #10235, opened 2021-10-21
Autofixes on class side raise DNU

### LINKED ISSUE #10234: Autofixes on class side raise DNU
When one has a lint rule with an autofix (for example "Inconsistent method classification" )
 - then autofix work on instance side
 - but not **on class side on gets an error**

as soon as one clicks on Autofix

image

image

astares:
Fix #10234


MarcusDenker:
I tested it with repairing a wrong category (once class side, once instance side). Works!

PR #10236, opened 2021-10-21
Cleanup-Deprecated9-More

MarcusDenker:
This PR removes more usused code from the Deprecated90 package

#leadingChar and #unicodeToMacRoman have still senders, I will open another issue for those


kasperosterbye:
It seems the errors in CI is unrelated to this. The changes are fine. I say merge.

In general there are a lot of deprecated classes. For one like me it can be very hard to figure out why something was deprecated and what to use instead. I like that the deprecated classes stick around for a while. But better if it was policy that in order for something to be deprecated you added a comment (package/class/method depending on what it is).

MarcusDenker:
We need class deprecation to at least warn the user (that is, reading the global should lead to transcript output). 
Even better would be automatic transforms for that as we have for methods... this is wanted for too long, I had not yet time to work on that.

kasperosterbye:
While the rewrite is super cool, I have sometimes experienced a certain amount of alienation, hence my wish for explanations.

kasperosterbye:
Btw, is there a "button" I could press stating "I vote for merge"?

MarcusDenker:
There is a "review" button where one can approve

kasperosterbye:
In the files pane - I have found it... Thanks.

PR #10243, opened 2021-10-21
10231-Cleanup-check-references-to-EncoderForSistaV1

### LINKED ISSUE #10231: [Cleanup] check references to EncoderForSistaV1
**Describe the problem**
Now that we have only EncoderForSistaV1, we can simplify all the places where we check for it (or set it explicitly).

Most of these have been already fixed, but if you check references to EncoderForSistaV1, it shows there are still some more
cases we have to clean

MarcusDenker:
clean up more code related to bytecode sets

fixes #10231


Ducasse:
Unix64.Kernel.Tests.Methods.BlockClosureTest

MarcusDenker:
yes, took the wrong value. Fixed

PR #10244, opened 2021-10-21
"Tools-Test" package should be "Tools-Tests"

astares:
Fix #10242

PR #10248, opened 2021-10-21
Release test for proper packages

### LINKED ISSUE #10247: Release test for proper packages
We should have a release test to check for proper and clean packages. 

1. When #10242 is fixed by integrating #10244 we could start by implementing a test that should check that we do not have packages ending with "Test" (as test packages should end with "Tests").

2. we could rename ProperManifestCategorizationTest (as it has only a single test testProperManifestCategorization
  which is also related to packages)

3. we could also put #testProperClassTagCasing into this test class as this is a check on the level of proper packages
    and remove #ProperClassCategorizationTest

astares:
Fix #10247


astares:
Can only become green when  #10244 is integrated

kasperosterbye (approved):
Two different methods are used to get the packages: `Smalltalk packages` and ` RPackageOrganizer default packages`. I suggest to pick one way to do it.

astares:
All green - but one platform is missing

image

kasperosterbye:
Notice made a comment - I recommend it being merged anyways.

guillep (approved):
Ok!

PR #10270, opened 2021-10-23
[Pharo9] Reactivating debugger opening strategies by default.

StevenCostiou:
When a debugger fails because of an error:
- tries to open the next available debugger (depending on the settings priorities)
- P9: opens GT, if GT fails it opens the emergency debugger
- P10: opens the emergency debugger (ED) immediately

If ED fails, this means the error is too critical to be handled (at least, in the current state of the system)


guillep:
Sorry Steven, we cannot integrate this in Pharo 9.

Also, Pharo10 is entering freezing mode, so we can discuss it, but it needs to be stable.

StevenCostiou:
No problem, the code is there for when it can be integrated.
Until then the emergency debugger will not be available though.

PR #10277, opened 2021-10-27
speedup-specialSelectorIndexOrNil 

MarcusDenker:
This is a small speedup for #specialSelectorIndexOrNil: to stop iterating when it finds a hit.

For a trivial test, this made senders of for #+ a little bit faster.

There are a lot more potentials for optimizing (e.g. check BytecodeEncoder>>specialSelectors instead of iterating slowly using specialSelectorAt:), but that is for later.

The method is tested by CompiledCodeTest>>#testHasSelectorSpecialSelectorIndex

### DIFF
--- a/src/System-Support/SmalltalkImage.class.st
+++ b/src/System-Support/SmalltalkImage.class.st
@@ -1811,12 +1811,11 @@ SmalltalkImage >> specialSelectorAt: anInteger [
 
 { #category : #'special objects' }
 SmalltalkImage >> specialSelectorIndexOrNil: selector [
-	| specialIndex |
 	1 to: Smalltalk specialSelectorSize do:
 		[:index | 
 		(Smalltalk specialSelectorAt: index) == selector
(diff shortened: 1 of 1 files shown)


Ducasse:
What is faster? How much?

Ducasse:
I never understood why this behavior is in a namespace.

MarcusDenker:
Just a micro-optimisation..

[#+ senders] bench

before: "'4.362 per second'"
after: "'4.436 per second'" 

(+ ist the first in the list, so for this it avoids the loop nearly completely)

For the normal case this just removes some assignment, that is not measurable.

This is called per class, so in systems with lots of classes this will have a larger impact.

What we need to do for real: call it just once per environment (use the same trick we use at the class level there).

Or invest the memory and add the special selectors to the literal frame the same as we are adding #ifTrue: and friends
(would cost ~800Kb in the current image, but the vm uses number of literals as a complexity heuristic for the JIT I think, so we have to asses the impact there, too)

PR #10278, opened 2021-10-27
10273-RBSmalllintContextprimitiveComputeLiterals-can-just-use-EncoderspecialSelectors

### LINKED ISSUE #10273: RBSmalllintContext>>#primitiveComputeLiterals can just use Encoder>>#specialSelectors
**Describe the problem**
RBSmalllintContext>>#primitiveComputeLiterals creates exactly the array that Encoder>>#specialSelectors returns

**Classes involved**
RBSmalllintContext, Encoder

**Proposal**
As a first step, we can just use Encoder>>#specialSelectors and replace primitiveComputeLiterals

MarcusDenker:
use "BytecodeEncoder specialSelectors" as a simpler way to get all special selectors.

(later we need to unify  #specialSelectors on BytecodeEncoder and SmalltalkImage)

fixes #10273


Ducasse:
I do not see why Smalltalk would need to know it. 
ByteCodeEncoder looks better to me. And in addition RB refactorings could be initialized with it.
S

PR #10283, opened 2021-10-27
Reduce usage of NativeBoost calls for UnifiedFFI-Legacy

### LINKED ISSUE #10282: Reduce usage of NativeBoost calls for UnifiedFFI-Legacy
There is a package providing some #nbSomething selectors like #nbFloat32AtOffset:, #nbFloat64AtOffset: and many other.
They are legacy from NativeBoost times - but some classes still call these methods.

We should inline them so they are not used in standard image anymore.

astares:
Fix #10282

### DIFF
--- a/src/Athens-Cairo/AthensCairoStrokePaint.class.st
+++ b/src/Athens-Cairo/AthensCairoStrokePaint.class.st
@@ -141,7 +141,7 @@ AthensCairoStrokePaint >> setDashesOn: aCanvas [
 	
 	dashLenghts withIndexDo: [:len :i |
 		
-		buf nbFloat64AtOffset: (i-1)*8 put: len asFloat
+		buf doubleAt: (i-1)*8 + 1 put: len asFloat
 		].
 
 	aCanvas
--- a/src/Athens-Cairo/AthensCairoSurface.class.st
+++ b/src/Athens-Cairo/AthensCairoSurface.class.st
@@ -209,7 +209,7 @@ AthensCairoSurface class >> findSurface: surfaceId [
 	
 	
 	(self primFindSurface: surfaceId handle: buf) ifTrue: [
-		^ buf nbUlongAt: 1
+		^ buf unsignedLongAt: 1
 		].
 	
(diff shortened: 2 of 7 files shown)


MarcusDenker (approved):
I think this is good, but I would like to have a review from someone more knowledgable with FFI before merging

astares:
@MarcusDenker It just inlines the methods - so it substitutes to use the new calls and not the old NB ones from UnifiedFFI-Legacy. From that POV it should be safe to integrate or can be reviewed even without much knowledge on the inner workings of UFFI.

So the exchange pattern is simple and I will explain on a concrete example:

1. if you look at method #setDashesOn:  it uses the old #nbFloat64AtOffset: method from the legacy package "UnifiedFFI-Legacy":
```Smalltalk
setDashesOn: aCanvas 
	| buf |
	buf := ByteArray new: 8*dashLenghts size.	
	dashLenghts withIndexDo: [:len :i |
		buf nbFloat64AtOffset: (i-1)*8 put: len asFloat ].

	aCanvas
		primSetDashesLengths: buf
		count: dashLenghts size
		offset: dashOffset
```

2. if you look at ByteArray>>#nbFloat64AtOffset: method you will see it already calls the now valid #doubleAt: method from "FFI-Kernel":

```Smalltalk
nbFloat64AtOffset: zeroBasedOffset
	"Read 64-bit float at ZERO-based index."
	^ self doubleAt: zeroBasedOffset + 1
```

3. So the replacement/inlining pattern is to now use the #doubleAt: (with the correct index increased by one) in our original method:

```Smalltalk
setDashesOn: aCanvas 
	| buf |
	buf := ByteArray new: 8*dashLenghts size.	
	dashLenghts withIndexDo: [:len :i |
		buf doubleAt: (i-1)*8 + 1 put: len asFloat ].

	aCanvas
		primSetDashesLengths: buf
		count: dashLenghts size
		offset: dashOffset
```

So we should end up in not using the old Native boost methods (#nbWhatever selectors) from the legacy package anymore

PR #10285, opened 2021-10-28
Added offset-based copy method to ExternalAddress to parallel *offset…

lello-at-lifeware:
Added a copy method to ExternalAddress based on 0-based offsets


Ducasse:
Thanks for the contribution.

PR #10288, opened 2021-10-29
Revert "10273-RBSmalllintContextprimitiveComputeLiterals-can-just-use-EncoderspecialSelectors"

### LINKED ISSUE #10286: CoAvoidRepeatedFetcher>>hasAlreadyReturned: fails when editing any method
**Bug description**
The method: 
```smalltalk
CoAvoidRepeatedFetcher >>hasAlreadyReturned: anEntry

	^ elements includes: anEntry contents asSymbol
```
fails when trying to edit any method. It fails with `#asSymbol was sendt to nil`.

**To Reproduce**
1. Go to any class, but reproducible from `ZnClient>>addPathSegment:`
2. Try to type `request foo` in the empty line just after the method comment

**Expected behavior**
No error should happen

**Version information:**
 - OS: [macOS version 11.6]
 - Pharo Version [Pharo10.0.0 build: 233, commit: 6f901cd, date: 28 October 2021]

MarcusDenker:
Reverts pharo-project/pharo#10278

#specialSelectors contains a nil, it is not just a list of the special selectors.

### DIFF
--- a/src/HeuristicCompletion-Model/SystemNavigation.extension.st
+++ b/src/HeuristicCompletion-Model/SystemNavigation.extension.st
@@ -19,7 +19,9 @@ SystemNavigation >> allSentMessagesInClass: aClass do: aBlock [
 			(literal isMemberOf: Array) ifTrue: ["might be performed"
 				literal do: [:x | (x isSymbol) ifTrue: [ x == method selector ifFalse: [ aBlock value: x ]]]]]].
 	"The following may be sent without being in any literal frame"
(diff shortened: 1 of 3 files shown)


MarcusDenker:
I willl open another issue tracker entry with a summary about #specialSelectors, as the situation is more complex than a small cleanup.

MarcusDenker:
(the nil of course has to be there as there used to be #class, which we do not treat special anymore.. the sad thing is that I should have known that...)

PR #10293, opened 2021-10-30
Added IdentitySet>>identityIncludes: for improved performance

### LINKED ISSUE #10292: Override #identityIncludes: on IdentitySet
` IdentitySet` currently doesn't override `#identityIncludes:`. The abstract implementation uses a loop to find the argument, which does not make much sense for an `IdentitySet`. The performance of `IdentitySet..identityIncludes:` is currently O(n) when id should be O(1) (thanks to @maenu for discovering this).

theseion:
Fixes #10292


MarcusDenker (approved):
looks good, tests are green

PR #10309, opened 2021-11-02
Microoptimization-MethodDictionary-scanFor

MarcusDenker:
All #scanFor: methods use == nil instead of #isNil to avoid one message send.

This PR unifies the situation and does this for MethodDictionary and the OCLiteralSet, too.

"implementors of" got 2% faster for me with this change.


MarcusDenker:
This PR unifies all #scanFor: methods to use #== instead of isNil  (just two where not doing that)

massimo-nocentini (approved):

Issue #9715, opened 2021-07-31
Pharo 9 looks bad on Windows HiDPI Devices (Microsoft Surface)


astares:
**DEVICE:** Microsoft Surface Laptop with Windows 10

The Fonts in Pharo 9 look washed out - while the fonts in Pharo 8 look very sharp in comparision. Maybe that is related to

![image](https://user-images.githubusercontent.com/5980033/127747215-98347d2d-18ad-46eb-82da-658976b4089a.png)

astares:
I guess this is due to some changes in Pharo 9 for HiDPI Support.

What works for me better is when I follow the instructions from @KubaO as given in 
https://github.com/pharo-project/pharo/issues/2833

and changing OSSDL2BackendWindow>>#fetchDPI method by adding 

```Smalltalk
fetchDPI
	| displayIndex ddpi hdpi vdpi newDiagonalDPI newScreenScaleFactor |
	"In OS X disable the computation of a DPI based scale factor. These values
	are completely wrong and we are getting scale factors that are very large
	even in non-retina display. This workaround does not affect the support for
	retina display because that is handled by fetching the size of the drawing surface."
	Smalltalk os isMacOSX ifTrue: [ 
		diagonalDPI := verticalDPI := horizontalDPI := self screenScaleFactorBaseDPI.
		screenScaleFactor := 1.0.
		^ false
	].

	displayIndex := sdl2Window getDisplayIndex.
	displayIndex < 0 ifTrue: [ ^ false ].
	
	ddpi := ByteArray new: 4.
	hdpi := ByteArray new: 4.
	vdpi := ByteArray new: 4.
	(SDL2 getDisplay: displayIndex ddpi: ddpi hdpi: hdpi vdpi: vdpi) = 0 ifFalse: [ ^ false ].
	
 	ddpi floatAt: 1 put: 96.0.
	ddpi floatAt: 1 put: 96.0.
	ddpi floatAt: 1 put: 96.0.
	
	newDiagonalDPI := ddpi floatAt: 1.
	"Allow automatic scale of factor increments of 5%. We need to
	round due to precision issues which may result in an almost 1 scale factor, but not exactly one.
	This has to be at least a divisor of 25% which are normal scale factor increments that are selectable in Windows.
	"
	newScreenScaleFactor := newDiagonalDPI / self screenScaleFactorBaseDPI roundTo: 0.05.
[...]

Issue #9038, opened 2021-04-11
Windows edges grabbing for resizing not optimal


fedemennite:
**Bug description**
There are a number of issues that make windows edges mouse grabbing for resizing problematic, especially on high resolution displays on Windows.
The real fix would probably be to have Pharo to use the OS window manager. Alternatively one or more of the following suggestions could be evaluated/implemented
- windows edges grab areas probably need to be slightly larger
- it can quite easily happen that you end up grabbing the window for placement instead of grabbing it for resizing. Isn't the upper bar area of the window enough for that?
- the resize cursor should also show up while over the edges of windows not currently having the focus
Although not critical, this issue seems to have detrimental impact on productivity

**To Reproduce**
Move the mouse over the edges of an open window, with or without focus                                                                                                                                                                   

**Expected behavior**
Windows should be easier to resize without requiring sniping skills ;)

**Version information:**
 - OS: Windows
 - Version: 10
 - Pharo Version 9

**Expected development cost**
Medium to High

tesonep:
@fedemennite @VincentBlondeau Can you tell me if you see an improvement with it?
Thanks.

NSUSpray:
> * the resize cursor should also show up while over the edges of windows not currently having the focus

Causes an unpleasant user experience. And still relevant (Pharo 10).

Ducasse:
Sorry but P10 is only get important updates. So if this is fixed in P11, then the issue is fixed.

NSUSpray:
I also apologize, but now I checked in Pharo 11 and this issue is still there.
![Pharo-windows-edges](https://user-images.githubusercontent.com/15806293/190892955-af31859e-2908-43be-bf30-d54532ed7ee1.gif)

Ducasse:
thanks.

MarcusDenker:
I added two issues:

- [ ]  Forward port 9038 to Pharo 11: Windows edges grabbing for resizing not optimal https://github.com/pharo-project/pharo/issues/10347
- [ ] Forward port 9038 to Pharo 10: Windows edges grabbing for resizing not optimal  https://github.com/pharo-project/pharo/issues/11696

pavel-krivanek:
fun fact, it sometimes changes the cursor for windows without focus

![Pharo_GMut86MUTT](https://user-images.githubusercontent.com/572494/225394677-e5d40de4-2536-4e81-8f92-28ea8399c6be.gif)

pavel-krivanek:
[...]

PR #10325, opened 2021-11-04
10319-OSSDL2BackendWindowfetchDPI-seems-broken

### LINKED ISSUE #10319: OSSDL2BackendWindow>>#fetchDPI seems broken
**Bug description**
`OSSDL2BackendWindow>>#fetchDPI` can't fetch correct DPI settings, newScreenScaleFactor temp is always 1.0d0
I tested on Windows 10 machines, both virtual and physical

**To Reproduce**
1. In Windows display settings (right click on desktop>Display settings), set the scale factor (DPI) to anything but 100%
2. Open a Pharo image (notice the scale of Pharo is wrong compared to other apps)
3. Inspect `OSSDL2BackendWindow allInstances first screenScaleFactor`.
4. You get `1` instead of your actual scale factor

**Expected behavior**
`OSSDL2BackendWindow>>#fetchDPI` should get the correct DPI value, so that Pharo scales correctly

**Screenshots**
On the screenshot, notice Pharo's scale factor set to 1, with this setting we except Pharo to scale like other apps, but it can't because `OSSDL2BackendWindow>>#screenScaleFactor` returns `1` instead of `1.75`
pharo-DPI

**Version information:**
 - OS: Windows 10
Pharo 9.0.0
Build information: Pharo-9.0.0+build.1564.sha.f5f541cb35c8609abfec418e35d5506715c7987d (64 Bit)

tesonep:
Fix #10319 
It is related to #9038 
@sonibla can you check if it improves #9038?

### DIFF
--- a/src/OSWindow-SDL2/OSSDL2BackendWindow.class.st
+++ b/src/OSWindow-SDL2/OSSDL2BackendWindow.class.st
@@ -211,10 +211,6 @@ OSSDL2BackendWindow >> fetchDPI [
 	vdpi := ByteArray new: 4.
 	(SDL2 getDisplay: displayIndex ddpi: ddpi hdpi: hdpi vdpi: vdpi) = 0 ifFalse: [ ^ false ].
 		
- 	ddpi floatAt: 1 put: 96.0.
-	ddpi floatAt: 1 put: 96.0.
-	ddpi floatAt: 1 put: 96.0.
-		
 	newDiagonalDPI := ddpi floatAt: 1.
 	"Allow automatic scale of factor increments of 5%. We need to
 	round due to precision issues which may result in an almost 1 scale factor, but not exactly one.


astares:
@tesonep Be aware that this reverts https://github.com/pharo-project/pharo/pull/9716/files which was done to fix https://github.com/pharo-project/pharo/issues/9715 which might then reappear, no?

<blockquote><div> GitHub</div><div><strong><a href="https://github.com/pharo-project/pharo/pull/9716">Pharo 9 looks bad on Windows HiDPI Devices (Microsoft Surface) by astares · Pull Request #9716 · pharo-project/pharo</a></strong></div><div>Fix #9715</div></blockquote>

tesonep:
Yes, the previous PR was disabling the scaling basically. It is true that the scaled version is blured, but it is better than have an smaller world. I will put a setting to disable automatic scaling.

guillep:
Needs to be forward ported to Pharo 10

astares:
@tesonep: it was already merged by @guillep.  Will you open another issue so we can have it as a setting?

Issue #10253, opened 2021-10-22
Improve Code Completion in Large Images


tesonep:
- [ ] Keyboard input freezes and queues events until the lookup ends.
- [ ] Frozen looking for non-existing symbols as receivers
- [ ] Widowing and Narrowing of queries.

Continuation of Issue #6311

MarcusDenker:
CoPackageImplementedMessagesFetcher>>#entriesDo: could be slow for a package with lots of classes.

guillep:
> CoPackageImplementedMessagesFetcher>>#entriesDo: could be slow for a package with lots of classes.

Yes, but in practice it is not, even for large images in lifeware, so no need to optimise there. The bottleneck is somewhere else: when no matches are found and all collections should be iterated for nothing.

PR #10326, opened 2021-11-04
Completion index

guillep:
Fix for https://github.com/pharo-project/pharo/issues/10253.
It includes extensions to the code completion system to use indexed accesses to classes.
 - extend system navigation with class begins with accesses
 - refactor fetchers to make the filter a fetcher property => this is necessary so the filter can be applied directly on the access and not afterwards

Yet to be tested with large images.


guillep:
This is a WIP, we need to still measure performance before integrating.

PR #10327, opened 2021-11-09
Added a MakeAbstract command to execute a transformation to make a class abstract …

miguelcamp:
Executes a transformation to make a class abstract by adding a method isAbstract. Should the way abstract classes are defined changes, this command would also have to change

### DIFF
--- a/src/SystemCommands-ClassCommands/SycCMakeAbstractCommand.class.st
+++ b/src/SystemCommands-ClassCommands/SycCMakeAbstractCommand.class.st
@@ -4,9 +4,6 @@ I am a command to make a class abstract by adding a method that returns whether 
 Class {
 	#name : #SycCMakeAbstractCommand,
 	#superclass : #SycClassCmCommand,
-	#instVars : [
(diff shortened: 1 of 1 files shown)


MarcusDenker:
There is an instance variable defined but not used:

```
Error
the following classes have unused instance variables and should be cleaned: an OrderedCollection(SycCMakeAbstractCommand)
Stacktrace
TestFailure
the following classes have unused instance variables and should be cleaned: an OrderedCollection(SycCMakeAbstractCommand)
NoUnusedVariablesLeftTest(TestAsserter)>>assert:description:resumable:
NoUnusedVariablesLeftTest(TestAsserter)>>assert:description:
NoUnusedVariablesLeftTest(TestCase)>>performTest
```

MarcusDenker:
Yes, the tagging using a method on the class side is just what we can do now... With Fluid Class Definitions we can move this later to the class definition itself. (and then we will remove all those methods and update the refactoring).

Thus: ok to merge

PR #10342, opened 2021-11-16
Correct casing for STON-Core extensions

### LINKED ISSUE #10341: Cleanup: STON-Core extensions should use correct casing
we should use "*STON-Core" and not "*ston-core" for the extensions:

```
badExtensions := CompiledMethod allInstances select: [:each | each protocol notNil and: [ each protocol beginsWith: '*ston-core']]
```

astares:
Fix #10341


guillep (approved):
Good!

Issue #10290, opened 2021-10-29
[CI] ZnHTTPSTest>>testGmailEncrypted started failing on the CI


MarcusDenker:
**Bug description**
#testGmailEncrypted is failing on all 3 platforms with ZnTooManyRedirects

```
Stacktrace
ZnTooManyRedirects
ZnClient>>executeWithRedirectsRemaining:
ZnClient>>executeWithRedirectsRemaining:
ZnClient>>executeWithRedirectsRemaining:
ZnClient>>executeWithRedirectsRemaining:
[ self executeWithRedirectsRemaining: self maxNumberOfRedirects ] in ZnClient>>executeWithRetriesRemaining:
FullBlockClosure(BlockClosure)>>on:do:
ZnClient>>executeWithRetriesRemaining:
[ self executeWithRetriesRemaining: self numberOfRetries ] in [ 
		[ self executeWithRetriesRemaining: self numberOfRetries ]
			on: Error
			do: self ifFailBlock ] in ZnClient>>executeWithTimeout
FullBlockClosure(BlockClosure)>>on:do:
[ 
		[ self executeWithRetriesRemaining: self numberOfRetries ]
			on: Error
			do: self ifFailBlock ] in ZnClient>>executeWithTimeout
[ ^ block value ] in ZnClient>>withTimeoutDo:
[ activeProcess
			psValueAt: index
			put: anObject.
		aBlock value ] in ZnConnectionTimeout(DynamicVariable)>>value:during:
FullBlockClosure(BlockClosure)>>ensure:
ZnConnectionTimeout(DynamicVariable)>>value:during:
ZnConnectionTimeout class(DynamicVariable class)>>value:during:
ZnClient>>withTimeoutDo:
ZnClient>>executeWithTimeout
[ result := self executeWithTimeout ] in ZnClient>>execute
[ ^ block value ] in ZnClient>>withProgressDo:
[ activeProcess
			psValueAt: index
			put: anObject.
		aBlock value ] in ZnSignalProgress(DynamicVariable)>>value:during:
FullBlockClosure(BlockClosure)>>ensure:
ZnSignalProgress(DynamicVariable)>>value:during:
ZnSignalProgress class(DynamicVariable class)>>value:during:
ZnClient>>withProgressDo:
ZnClient>>execute
ZnClient>>get
ZnClient>>get:
ZnHTTPSTest>>testGmailEncrypted
ZnHTTPSTest(TestCase)>>performTest
```

[...]

PR #10355, opened 2021-11-16
10353-Backport-Pharo9-CI-ZnHTTPSTesttestGmailEncrypted-started-failing-on-the-CI- 

### LINKED ISSUE #10353: Backport Pharo9: [CI] ZnHTTPSTest>>testGmailEncrypted started failing on the CI
**Bug description**
We need to backport the fix of #10290 as #testGmailEncrypted is failing on the CI for Pharo9

MarcusDenker:
backport #10290

### DIFF
--- a/src/Zinc-HTTP/ZnClient.class.st
+++ b/src/Zinc-HTTP/ZnClient.class.st
@@ -915,9 +915,9 @@ ZnClient >> maxNumberOfRedirects [
 	"Return the maximum number of HTTP redirect that will be followed.
 	Note that when the #followsRedirect option is false, 
 	no redirects will ever be followed regardless of the value of this option.
-	Defaults to 3. A ZnTooManyRedirects will be signalled when the count reaches zero."
+	Defaults to 5. A ZnTooManyRedirects will be signalled when the count reaches zero."
 
-	^ self optionAt: #maxNumberOfRedirects ifAbsent: [ 3 ]
+	^ self optionAt: #maxNumberOfRedirects ifAbsent: [ 5 ]
 ]
 
 { #category : #options }


src/Zinc-HTTP/ZnClient.class.st
```diff
@@ -915,9 +915,9 @@ ZnClient >> maxNumberOfRedirects [
 	"Return the maximum number of HTTP redirect that will be followed.
 	Note that when the #followsRedirect option is false, 
 	no redirects will ever be followed regardless of the value of this option.
-	Defaults to 3. A ZnTooManyRedirects will be signalled when the count reaches zero."
+	Defaults to 5. A ZnTooManyRedirects will be signalled when the count reaches zero."
 
-	^ self optionAt: #maxNumberOfRedirects ifAbsent: [ 3 ]
+	^ self optionAt: #maxNumberOfRedirects ifAbsent: [ 5 ]
```
tesonep, line 920: We should not change the global redirect to fix a test. The test should be modified, this has a really bad impact.
Has this change been impacted in Pharo10?

tesonep (changes requested):

src/Zinc-HTTP/ZnClient.class.st
MarcusDenker: This is the fix that was done in the Zinc upstream. It is already in Pharo10

MarcusDenker: see the discussion in https://github.com/pharo-project/pharo/issues/10290

svenvc: > We should not change the global redirect to fix a test. The test should be modified, this has a really bad impact.
> Has this change been impacted in Pharo10?

3 was a bit too low, in certain valid cases, like in the test, more than 3 redirects do happen.
The main objective is to prevent infinite loops, as these do happen occasionally.

tesonep: Ok, perfect

PR #10361, opened 2021-11-16
Fix Zinc extensions should use correct casing (package name)

### LINKED ISSUE #10360: Cleanup zinc extensions should use correct casing (package name)
In Pharo standard image we have some extensions (protocol name on extension method with a leading *) left that use not the name of the package correctly (wrong casing). This was fixed for Metacello-Core, STON, OpalCompiler and Renraku packages already (see #10343, #10337, #10356 and others)

This ticket here is for **Zinc** specifically. Having this as a separate PR will hopefully also allow @svenvc to reintragrate this change back into the original Zinc framework.

Details:
This extension issues seem to be a leftover from times (old pharo versions or even back from Squeak times) where the extension have been manually entered with a * following the name. Back then the casing and package name was not checked - so this is now a leftover in source code.

Later when Calypso was used as system browser there was a better way: a package selection dialog  was introduced to  to select the exact package when defining an extension.

We can pick out the cases from Zinc using:
```Smalltalk
badExtensions := CompiledMethod allInstances select: [:each | each protocol notNil and: [ each protocol beginsWith: '*zinc-']]
```

if we want to check which extensions do not use an uppercase package name we can use the following script to check
the whole image:

```Smalltalk
| condition |
condition := [:protocol | (protocol beginsWith: '*') and: [ protocol second isLowercase  ]].
casesToFix := CompiledMethod allInstances select: [:each | each protocol notNil and: [ condition value: each protocol ]].
badExtensionProtocols := (wrong collect: [:each | each protocol ]) asSet asSortedCollection
```

astares:
Fix #10360 

Something that possibly @svenvc can also merge back into https://github.com/svenvc/zinc

### DIFF
--- a/src/Zinc-HTTP/ZnUrl.extension.st
+++ b/src/Zinc-HTTP/ZnUrl.extension.st
@@ -1,6 +1,6 @@
 Extension { #name : #ZnUrl }
 
-{ #category : #'*zinc-http' }
+{ #category : #'*Zinc-HTTP' }
 ZnUrl >> retrieveContents [
 	"Download and return the resource that I refer to.
 	This will typically return a String or a ByteArray.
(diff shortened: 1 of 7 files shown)


MarcusDenker (approved):

MarcusDenker:
- [CI] Test run on the mac sometimes is terminated #10462

PR #10376, opened 2021-11-17
10375-Selecting-Packages-in-Calypso-produces-MNU

### LINKED ISSUE #10375: Selecting Packages in Calypso produces MNU
When selecting a package, there is an MNU in RPackage.

tesonep:
Fixing broken method in Renraku


tesonep:
Fix #10375

PR #10383, opened 2021-11-18
Fix #8443 Crash on step over

### LINKED ISSUE #8443: Crash on step over

isCzech:
When debugging things like this (and similar, described in the Issue #8443):
```
[^2] ensure: [Transcript cr; show: 'done']
```
if we step into the protected block [^2] and then step over ^2, we crash the image or get an infinite recursion etc.
This behavior happens because #return:from: supplies firstUnwindContext to #resume:through: (via #aboutToReturn:through:) but by the time this argument can be used in #resume:through: the #runUntilErrorOrReturnFrom: method inserts another unwind context before this supplied firstUnwindContext! As a result firstUnwindContext is no longer the first unwind context and the computation derails because it misses the new unwind context inserted during simulation.

The proposed fix communicates the fact the computation is in simulation by sending nil instead of the firstUnwindContext (from #return:from: to #resume:through: via #aboutToReturn:through:) causing a fresh search for the actual first unwind context is invoked at the right time - before executing #resume:through:

This creates a close link between a simulation environment and the execution environment which is not ideal so I've included a note explicitly pointing at the relationship. In theory though even VM could send nil as firstUnwindContext value and the code would still work so I guess the little hack is justifiable.

Best regards,
Jaromir

### DIFF
--- a/src/Kernel/Context.class.st
+++ b/src/Kernel/Context.class.st
@@ -1642,7 +1642,8 @@ Context >> resume: value through: firstUnwindContext [
 	| context unwindBlock |
 	self isDead 
 		ifTrue: [ self cannotReturn: value to: self ].
-	context := firstUnwindContext.
+	context := firstUnwindContext ifNil: [thisContext findNextUnwindContextUpTo: self].
+	"note: nil value of firstUnwindContext is sent by #return:from: to invoke fresh search for first unwind context"
 	[ context isNil ] whileFalse: [	
 		context unwindComplete ifNil:[
 			context unwindComplete: true.
@@ -1697,13 +1698,13 @@ Context >> return: value [
 Context >> return: value from: aSender [ 
 	"For simulation.  Roll back self to aSender and return value from it.  Execute any unwind blocks on the way.  ASSUMES aSender is a sender of self"
 
-	| newTop context |
+	| newTop |
 	aSender isDead ifTrue: [
 		^ self send: #cannotReturn: to: self with: {value} super: false ].
 	newTop := aSender sender.
-	context := self findNextUnwindContextUpTo: newTop.
-	context ifNotNil: [
-		^ self send: #aboutToReturn:through: to: self with: {value. context} super: false] .
+	(self findNextUnwindContextUpTo: newTop) ifNotNil: [
+		^ self send: #aboutToReturn:through: to: self with: {value. nil} super: false]. 
+		"note: nil is sent to #resume:through via #aboutToReturn:through: to invoke fresh search for next unwind context there"
 	self releaseTo: newTop.
 	newTop ifNotNil: [ newTop push: value ].
 	^ newTop


Ducasse:
Thanks for the report and fix.
Now pay attention that we use crash as the vm quits unexpectedly. Is it what you mean?

Ducasse:
Do you have a test or more that would cover this behavior (the "crash") and/or the fix?

isCzech:
> Thanks for the report and fix. Now pay attention that we use crash as the vm quits unexpectedly. Is it what you mean?

Freeze, sorry I meant freeze :) The VM didn't "crash" but froze as reported in the issue. The root cause is described in the PR and thanks for reminding me - I'll think about how to test it if possible (or if my experience allows). Or at least provide a few examples... I used just this one:
```[^2] ensure: [Transcript cr; show: 'done']```
The same flawed behavior shows when you
1. step over ^2, or
2. step into ^2 and then step over return:through:, or
3. step further into return:through: and then step over resume:through:

In each of the cases the cause is identical - the insertion of the extra unwind context by runUntilErrorOrReturnFrom: but not taking it into account when executing resume:through: during simulation.

The difference between runtime and simulation of the example is the VM sends the first unwind context it found via aboutToReturn:through: and it is immediately propagated to resume:through: without any interference...

**Comment by isCzech** [comment]
Hi,
The fix covers the examples reported by Denis in #8509 - they won't freeze any more. And maybe some Debugger issues in #8567 may benefit as well:
```
[
   [ 100 + 200 ] ensure: [^600]
] ensure: [ #done ]
"Step inside [^600] block. Then step over it will hang the image."
```
and 
```
[ ^100 ] ensure: [ 10+30]
"Step into block [^100]. Then step over it will hang the image."
```
I can't think of a proper "test" for this situation. 
To better understand the issue I suggest this experiment: Replace #aboutToReturn:through: with this version (removing firstUnwindContext argument from the picture)
```
aboutToReturn: result through: firstUnwindContext 
	"Called from VM when an unwindBlock is found between self and its home.
	 Return to home's sender, executing unwind blocks on the way."

	self methodReturnContext return: result
```
... and voilà - the problem's gone :)
(To prevent any misunderstanding - modifying #aboutToReturn:through: is NOT a fix - it's just an experiment aimed to shed more light on the issue. The fix is presented in this PR :) )

**Comment by Ducasse** [comment]
Thanks for the explanation

**Comment by dionisiydk** [comment]
Nice fix. I am wondering why it also fixes the case with return from ensure block. My fix for this was in another method as you can see in  #8509 . I will look deeply.

> I can't think of a proper "test" for this situation.

Did you try to add similar test like in my PR?

The original issue leaded me to to the general problem with any debugger bugs where image freezes in infinite recursion. So I stopped doing any fixes until it will be sorted with #8567  . Sadly it's almost a year when I pushed it and it still waiting review. And now it requires complex merging with separate projects (managed out of pharo repo) - again related PR is waiting several months.

**Comment by dionisiydk** [comment]
Hi @isCzech .

It looks good. But you should add test from my PRD here.

**Comment by isCzech** [comment]
Hi, I’m AFK, will do. Thanks.

**Comment by isCzech** [comment]
Wow, you've got an impressive testing suite under Debugger tests :o Yes, we need a similar test like yours for a method like:
```
methodWithBodyReturn [
	[ ^#bodyReturn ] ensure: [ #unwind ]
]
```
Do stepThrough until inside [ ^#bodyReturn ] and then stepOver and verify it's ok.

Frankly, I don't feel like promising I'd do it promptly - it'll take me hours to familiarize myself with your testing environment; I've never built tests like this :) So yes, it'll be a great joy to learn but at the moment I'm under pressure to finish something :|

Denis, you also mentioned [Very critical fix] Processor>>#terminateRealActive #8567. Yes, I remember that one, I commented regarding my fixes #8993 in #terminate and #isTerminated. I tried to understand your fix in #8567 but gave up eventually as it was too complex for me and I couldn't keep all the details in my head. One of the problems, in my opinion, is that #terminate is polluted with bugs with unpredictable consequences god knows where, so I concluded for myself these #terminate bugs should be cleared out of the way first, because they are at the deepest level in Kernel. Similar may be true for this stepOver bug - it lies deep, very well hidden and makes complex changes like yours even harder.

Ducasse:
We are also under pressure :(

dionisiydk:
> Similar may be true for this stepOver bug - it lies deep, very well hidden and makes complex changes like yours even harder.

The reason why I stopped doing concrete fixes for debugger is because it is a nightmare. Even while checking this PR with two small changes I froze the image several times. So any mistake during a debugging session of a debugger bug leads the the image crash usually. It's always like in the old school C programs - start from scratch every time.
Now there is a general solution to this issue. So changes like yours and other fixes should be much simpler to do like for any other bugs. But it is still not integrated.

guillep:
Superseeded by #10429

PR #10390, opened 2021-11-23
Fix so that Rename Method refactoring does not start with a disabled text input

### LINKED ISSUE #10332: Renaming a Method starts with a disabled editbox.
Doing a rename of a method with Crl+R / Cmd+R the name of the method in the window is disabled, it cannot be edited.
@miguelcamp

miguelcamp:
Fix for issue #10332

### DIFF
--- a/src/SystemCommands-RefactoringSupport/SycMethodNameEditorPresenter.class.st
+++ b/src/SystemCommands-RefactoringSupport/SycMethodNameEditorPresenter.class.st
@@ -109,7 +109,6 @@ SycMethodNameEditorPresenter class >> openOn: aMethod [
 	"I take a RBMethodName as parameter and open the refactoring UI in a modal to rename it."
 	|temp|
 	temp := self on: aMethod.
-	self selectorInput beNotEditable.
 	^ temp openBlockedDialogWithSpec
 ]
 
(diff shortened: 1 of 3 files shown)


src/SystemCommands-RefactoringSupport/SycMethodNameEditorPresenter.class.st
```diff
@@ -346,7 +358,7 @@ SycMethodNameEditorPresenter >> initializeWidgets [
 	addButton label: '+'.
 	selectorInput text: methodName selector.
 	previewResult label: methodName methodName.
-	selectorInput beNotEditable.
+	"selectorInput beNotEditable."
```
kasperosterbye, line 361: Perhaps this line should be deleted (not leaving code in comments)

src/SystemCommands-RefactoringSupport/SycMethodNameEditorPresenter.class.st
```diff
@@ -136,6 +135,19 @@ SycMethodNameEditorPresenter class >> openOn: aMethod withInvalidArgs: aSet canR
 	^ temp openBlockedDialogWithSpec
 ]
 
+{ #category : #specs }
+SycMethodNameEditorPresenter class >> openOn: aMethod withInvalidArgs: aSet canRenameArgs: aBoolean1 canRemoveArgs: aBoolean2 canAddArgs: aBoolean3 canEditName: aBoolean4 [
```
kasperosterbye, line 139: Perhaps some better argument names?

kasperosterbye (approved):
I think it is good to go. I added a few detailed comments which should be simple to correct, but works fine without.

Is there a way for you to add a test to make sure the error do not sneak back in? That would be really nice.

Ducasse:
Miguel this is **really** annoying that we cannot use rename anymore in Pharo.
When you push a change please make sure that it works! 
And if you break something (this is ok) makes sure that you react fats to comments so that we can integrate them fast.

Ducasse:
I will integrate these changes and I hope that they work because to me I cannot use Pharo now.

PR #10434, opened 2021-11-24
Fix a rare case where the primitiveRegisterSurface fails without setting true or false.

### LINKED ISSUE #10433: Call to surface primitive must return a boolean but return self when failed
```
"protocol: #private"

primRegisterSurface: aCairoSurfaceHandle dispatch: sqSurfaceDispatchPtr surfaceId: idHolder
"
	Register a new surface with the given handle and
	the set of surface functions. The new ID is returned
	in surfaceID. Returns true if successful, false 
	otherwise."

	<primitive: 'primitiveRegisterSurface' module: 'SurfacePlugin'>
```

The primitive can fail without returning true or false. The fallback code should return false.
**To Reproduce**
No idea.

**Expected behavior**
Not crashing.

**Expected development cost**
I send the PR in few minutes.

Alisu:
Fix #10433.
Add a ^false in fallback code.

### DIFF
--- a/src/Athens-Cairo/AthensCairoSurface.class.st
+++ b/src/Athens-Cairo/AthensCairoSurface.class.st
@@ -308,6 +308,7 @@ AthensCairoSurface class >> primRegisterSurface: aCairoSurfaceHandle dispatch: s
 	otherwise."
 
 	<primitive: 'primitiveRegisterSurface' module: 'SurfacePlugin'>
+	^false
 ]
 
 { #category : #private }


MarcusDenker:
- randomly failing test (one that is already fixed)

MarcusDenker (approved):

MarcusDenker:
after re-run now random failure:

- [CI] Failing randomly: SemaphoreTest>>testMutualExclusion #10469

PR #10436, opened 2021-11-24
Reflectivity-AddPragmaSupport 

MarcusDenker:
This PR adds support (and a test) for metalinks on Pragmas. (Test was from #8374)

- add testSetLinkOnPragma
- adds support in the compiler (see #visitPragmaNode:)
- adds support to ask for reifications on RBPragmaNode

### DIFF
--- a/src/Reflectivity-Tests/ReflectiveMethodTest.class.st
+++ b/src/Reflectivity-Tests/ReflectiveMethodTest.class.st
@@ -211,6 +211,29 @@ ReflectiveMethodTest >> testSetLinkOnInstanceVariableAndUninstall [
 ]
 
 { #category : #'tests - links' }
+ReflectiveMethodTest >> testSetLinkOnPragma [ 
+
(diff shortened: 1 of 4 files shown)


guillep:
I don't think this should go for Pharo 10.
This goes in the *experimental range of things* to me, and we should stop adding features as we talked.

MarcusDenker:
It is not used outside of the test.. and it could be seen as a bug

MarcusDenker:
But I have no problem to postpone that one to Pharo11

guillep:
> It is not used outside of the test.. and it could be seen as a bug

That is not the point, the point is that as soon as it is inside the system, people will use it. And bugs will arrive. And that will put more pressure on us :)

I propose that this is kept in a separate project/branch or even a repository. It can be downloaded optionally, and later in the future we can evaluate it for integration when we know it is stable. Reality is that you will probably do many iterations on it until you're happy with it too.

Issue #10437, opened 2021-11-24
Wrong completion in keyword selectors


guillep:
**Bug description**
When completing keyword selectors, under some conditions the selector is duplicated.

**To Reproduce**
Imagine I’m trying to autocomplete #asStringOn:delimiter:.

Case 1) if we do not type the colon, it works ok

1 asStringOn
=complete=>
1 asStringOn:  delimiter: .

Case 2) if we type the colon, it will complete duplicating the first keyword altogether

1 asStringOn:
=complete=>
1 asStringOnasStringOn:  delimiter: 

Case 3) if we start typing the second keyword, it will complete duplicating the first keyword separated by a colon

1 asStringOn:de
=complete=>
1 asStringOn:asStringOn:  delimiter: 

We can start by writing some tests based on this :)

Similar problems happens with more than one keyword.

Ducasse:
I will check once the changes is integrated if my breaking scenario is fixed else I will report it. 
Basically

```
transform:| + completion
=> 
transformtransform:....
```

guillep:
Hi @Ducasse , were you able to test this?

Ducasse:
Yes and it worked :) I will continue to test but P10 looks to be in the right direction. 
implementors and senders are getting much better and completion too.

PR #10438, opened 2021-11-24
Fix completion of keyword selectors

guillep:
- Extend rubric iterate words backward and forward by selector when in smalltalk mode.
- Added completion tests for #10437

As a nice side effect, navigating code with ctrl+arrow now takes into account colons in keyword selectors too!

### DIFF
--- a/src/Rubric/RubSmalltalkEditor.class.st
+++ b/src/Rubric/RubSmalltalkEditor.class.st
@@ -818,6 +818,12 @@ RubSmalltalkEditor >> isSmalltalkEditor [
 	^ true 
 ]
 
+{ #category : #private }
+RubSmalltalkEditor >> isWordCharacter: aCharacter [
+	"In smalltalk code, colons make part of selectors too and are part of a word"
+	^ aCharacter isAlphaNumeric or: [aCharacter = $:]
+]
+
 { #category : #'menu messages' }
 RubSmalltalkEditor >> jumpToNextKeywordOfIt [
 	"Jump to the next keyword after the cursor - this is for legacy support"
--- a/src/Rubric/RubSmalltalkEditor.class.st
+++ b/src/Rubric/RubSmalltalkEditor.class.st
@@ -819,9 +819,18 @@ RubSmalltalkEditor >> isSmalltalkEditor [
 ]
 
 { #category : #private }
-RubSmalltalkEditor >> isWordCharacter: aCharacter [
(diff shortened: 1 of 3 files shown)


dionisiydk (approved):
That's nice. 
It would be cool to also implement the navigation between words inside selector. It was a feature in some first versions of Pharo or Squeak.

src/Rubric/RubSmalltalkEditor.class.st
```diff
@@ -818,6 +818,12 @@ RubSmalltalkEditor >> isSmalltalkEditor [
 	^ true 
 ]
 
+{ #category : #private }
+RubSmalltalkEditor >> isWordCharacter: aCharacter [
+	"In smalltalk code, colons make part of selectors too and are part of a word"
+	^ aCharacter isAlphaNumeric or: [aCharacter = $:]
```
kasperosterbye: I am able to get into trouble on this if I try to edit an assignment where there is no space between the variable and the `:=`. 
image
when I start typing `dumb` I get a completion for `dumb`, accepting the completion removes the `:`

guillep: Very good catch! Thanks Kasper, I'll make a test on that.

src/Rubric/RubSmalltalkEditor.class.st
```diff
@@ -819,9 +819,18 @@ RubSmalltalkEditor >> isSmalltalkEditor [
...
+	"By default, group alphanumeric and non-alphanumeric separately"
+	| character |
+	character := aString at: index.
 	"In smalltalk code, colons make part of selectors too and are part of a word"
-	^ aCharacter isAlphaNumeric or: [aCharacter = $:]
+	^ character isAlphaNumeric or: [
+		"Check that it is a keyword selector.
+		 - followed by a colon
+		 - but not by an assignment "
+		character = $: and: [
```
kasperosterbye, line 831: Looks good now. I do not believe there are any other loop-holes.

kasperosterbye:
I do not know if this is the same problem, or a separate issue. But let me try to explain.

If I start writing a new method which I want to be named say `asCamel`. When I finish writing the method name and press return to get to the next line, the autocompleter expands my method to be `asCamelCase`. Then the only way to compete the name `asCamel` is to press escape to stop the proposal, and then return. 

Perhaps I am just unfit for automatic popup :(

PR #10560, opened 2021-12-02
in fact, Keymapping-Pragmas package needs to be in UI group

estebanlm:
since is use there :)


estebanlm:
again... errors not related (but I am wonder why so many spec-related random errors. I think tomorrow I will work on that)

MarcusDenker (approved):

Issue #10519, opened 2021-11-29
[CI] Failing randomly: StPlaygroundPageTest>>testContents


MarcusDenker:
**Bug description**
StPlaygroundPageTest>>testContents sometimes fails randomly:

```
Error
Got '' instead of 'test 2'.
Stacktrace
TestFailure
Got '' instead of 'test 2'.
StPlaygroundPageTest(TestAsserter)>>assert:description:resumable:
StPlaygroundPageTest(TestAsserter)>>assert:description:
StPlaygroundPageTest(TestAsserter)>>assert:equals:

StPlaygroundPageTest(TestCase)>>performTest
```

seen on win  #10504

maxwills:
I created a PR for this issue here: https://github.com/pharo-spec/NewTools/pull/318
I have a question though. When contributing, should I create the pull request targeting Pharo9.0 or Pharo10 branch?

MarcusDenker:
PR was merged upstream

PR #10566, opened 2021-12-03
Fix 10565 - Extract freetype to its own baseline

estebanlm:


estebanlm:
error not related

src/BaselineOfFreeType/BaselineOfFreeType.class.st
```diff
@@ -0,0 +1,30 @@
...
+}
+
+{ #category : #baselines }
+BaselineOfFreeType >> baseline: spec [
+	<baseline>
+	
+	spec for: #common do: [ 
+		spec package: 'FreeType'.
+		spec package: 'FreeType-Morphic'.
+		spec package: 'FreeType-Tests'.
```
astares, line 14: Dependencies among packages should be declared, no?

Like for example:
```Smalltalk
spec package: 'FreeType-Tests' with: [ spec requires: #('FreeType' 'FreeType-Morphic') ]
```

estebanlm:
not really needed, since dependency is explicit in groups

astares:
@estebanlm 

It's also possible to load (single) packages - or reference them when loading from other baselines, not only the groups. So dependencies should be defined clearly on package level, additonally for the groups  IMHO

MarcusDenker:
random failure:

- [CI] Failing randomly: StPlaygroundPageTest>>testContents #10519

src/BaselineOfFreeType/BaselineOfFreeType.class.st
MarcusDenker: I will merge the current state for now...

MarcusDenker (approved):

PR #10611, opened 2021-12-05
Fixes 10610 - CairoFontExtents is missing a field 

### LINKED ISSUE #10610: CairoFontExtents is missing a field
**Bug description**
Subject says :) 
In addition, 
- that means if you execute `AthensCairoCanvas>>#text:extents:` you will crash. 
- the field missing is `width`.
This is not used in the system, but it may be used by athens users.

**To Reproduce**
In a cairo canvas, execute: 
```Smalltalk
	textExtents := CairoFontExtents new.
	aCanvas text: text extents: textExtents.
```
and enjoy the crash :)

**Expected behavior**
It should not crash

estebanlm:
Fixes https://github.com/pharo-project/pharo/issues/10610


estebanlm:
test error not related

Issue #10625, opened 2021-12-08
Test failing on CI: RubEditingAreaTest>>testEmbeddedAnchorInTextAreDrawnAtRightPosition


MarcusDenker:
**Bug description**
RubEditingAreaTest>>testEmbeddedAnchorInTextAreDrawnAtRightPosition started failing on the CI for all platforms.

maybe related to merging https://github.com/pharo-project/pharo/pull/10558 ?

kasperosterbye:
They all (all RubEditingAreaTest ) run on P10 build: 340, commit: fa6ca09, date: 8 December 2021.
They were green yesterday.
But it seem plausible that the merge causes the problems as the method was not present before the merge. There was a problem in that it opened some morphs on the (non-existing) screen, but I just checked the actual code in the repository, which is the same as in what I downloaded just now.

Is FormCanvas defined in headless mode?

MarcusDenker:
it for sure is something related to headless vs not headless... it is green if I run it in the same image that the CI has problem with

kasperosterbye:
OK, let me try to make a PR to see if I can find out where the problem is. I have some ideas to try out.

PR #10627, opened 2021-12-08
Is FormCanvas available?

kasperosterbye:
This PR is for trying to figure out what is the error in #10625.
Do not merge just now even if succeeding.

### DIFF
--- a/src/Rubric-Tests/RubEditingAreaTest.class.st
+++ b/src/Rubric-Tests/RubEditingAreaTest.class.st
@@ -87,6 +87,7 @@ RubEditingAreaTest >> testEmbeddedAnchorInTextAreDrawn [
 { #category : #tests }
 RubEditingAreaTest >> testEmbeddedAnchorInTextAreDrawnAtRightPosition [
 	| circle longTextWithCircleAtEnd canvas|
+	(Smalltalk globals includesKey: #FormCanvas) ifFalse: [^self].
 	circle := CircleMorphCounter new extent: 50 @ 50;color: Color blue.
 	longTextWithCircleAtEnd := ((String loremIpsum: 10000), 
 		(String value: 1)) asText 


MarcusDenker:
I am sure that the class exists. The image is the same... it just fails the test because it is running headless or something like that.

kasperosterbye:
Now - When I press details above I get this image: 
image

When pressing the yellow osx-64 I get something like this:
image

I can download the log, but I can not find out what actually caused the osx-64 to fail - where do you and Stef look that I overlook?

tesonep:
The class exists in the image, the image used in the tests is the same that is produced. 
You can get the image with the changes from the artefacts of the CI execution. 
For example, for this PR the artifacts are here:
https://ci.inria.fr/pharo-ci-jenkins2/blue/organizations/jenkins/Test%20pending%20pull%20request%20and%20branch%20Pipeline/detail/PR-10627/1/artifacts

And the image you wish is in the .zip file **bootstrap-cache/Pharo10.0-PR-64bit-9fc141b.zip**
In this case, as the PR has been integrated, the test will fail in the latest image.
To execute the test and see the error you can do in a terminal:

```bash
./pharo Pharo10.0-PR-64bit-634eaa6.image test --junit-xml-output "Rubric.*"
````
This will generate the same .xml files that the CI (it is the command executed by the CI).

<blockquote><div><strong><a href="https://ci.inria.fr/pharo-ci-jenkins2/blue/organizations/jenkins/Test%20pending%20pull%20request%20and%20branch%20Pipeline/detail/PR-10627/1/artifacts">Jenkins</a></strong></div></blockquote>

kasperosterbye:
Thanks for pointing to the artifacts tab @tesonep. When downloading the Pharo10.0-PR-64bit-9fc141b image and starting it on my machine all rubrics tests are green.

PR #10633, opened 2021-12-09
Fix metacello lock

estebanlm:
# Using Metacello lock

I have to work with different "in development" versions of several projects that depend one in the other. For example, I develop Spec2 and at the same time Spec-Gtk that depends on the first. Or I work on NewTools that also depends on Spec2.
I tend to have the three projects (well, some more too ;) ) in the same image, so I can work on any one that requires attention.

My "working environment" includes usually this projects: 

- `Spec2`, branch `dev-1.0`
- `Spec-Gtk`, branch `master`
- `NewTools`, branch `dev-1.0`
- `Test Project 1` (for example these days is the `PharoLauncher`, branch `Pharo9`)
- ...
- `Test Project N`

In Metacello, project dependencies can be done pointing to a branch or a tag. By default, it takes "master" branch (which should be changed in the near future, as soon as we have some time to spend there), but is recommended that each project declares the correct version through a tag.

And here is the problem: 

- `Spec2` depends on `SpecCore` branch `Pharo10` 
- `Spec-Gtk` depends on `Spec2` branch `Pharo10`
- `NewTools` depends on `Spec2` branch `Pharo10`
- `PharoLauncher` depends on `Spec2` branch `Pharo9` !
... and like that.

As you can see, I have a major dependency mismatch. But I need the versions I first declared since I am developing all of them.
To fix this, "regular" behavior would be to load in order `SpecCore -> Spec2 -> Spec-Gtk -> NewTools -> PharoLauncher`.

*But then Metacello tries to do **the right thing** and will load the correct dependency version of each project!*
And that will let my image with a lot of wrong versions! (and potentially not working, since Spec2 is used also for the tools in the image).

What to do?

## Metacello lock to the rescue

Thinking on how to solve this problem I remembered Metacello comes since the beginning with a functionality to deal with this situation: the capability to `lock` a project so the version loaded will remain untouched. 
To lock a project should be as easy as executing: 

```Smalltalk
Metacello new baseline: 'MyProject'; lock.
```

The problem is that in a "live environment" as Pharo every functionality that is not used or tested, eventually stops working. This is our own version of Lamark law (https://en.wikipedia.org/wiki/Lamarckism#Lamarck's_evolutionary_framework). 
And since nobody used this in years, and tests were not there... **the lock of projects was not working**. 

So, It was needing a fix, urgently since I am not the only one dealing with "large chains of dependencies".
After some time digging to understand the design and trying solutions, I decided to simplify the mechanism first developed that was not working (a mechanism that previewed a confirmation hook when trying the load), and I re-implemented it as easy as possible:  **Now, when you lock a project, its load will be skipped when found as a dependence**.

In this new implementation, lock works not *by project* but *by baseline*: Since a project can have different baselines, you need to lock each baseline of the project to lock the whole project. 

That way, I can load all projects with the versions I want: 

```Smalltalk
Metacello new baseline: 'Spec2'; repository: 'github://pharo-spec/Spec:dev-1.0'; load.
Metacello new baseline: 'SpecCore'; lock.
Metacello new baseline: 'Spec2'; lock.

Metacello new baseline: 'Spec-Gtk'; repository: 'github://pharo-spec/SpecGtk:master'; load.
Metacello new baseline: 'SpecGtk'; lock.

Metacello new baseline: 'NewTools'; repository: 'github://pharo-spec/NewTools:dev-1.0'; load.
Metacello new baseline: 'NewTools'; lock.

Metacello new baseline: 'PharoLauncher'; repository: 'github://pharo-project/pharo-launcher'; load.
```

I can then unlock my projects by executing: 

```Smalltalk
Metacello new baseline: 'SpecCore'; unlock.
Metacello new baseline: 'Spec2'; unlock.
Metacello new baseline: 'SpecGtk'; unlock.
Metacello new baseline: 'NewTools'; unlock.
```


estebanlm:
@guillep I'd like to convert this PR explanation into a blog post, do you think it is worthy?

Ducasse:
yes yes yes!

Ducasse:
I wanted to read it but with the diffs it was painful :)

guillep:
> @guillep I'd like to convert this PR explanation into a blog post, do you think it is worthy?

yes :)

src/Metacello-Core/MetacelloProjectSpecForLoad.class.st
```diff
@@ -85,10 +85,18 @@ MetacelloProjectSpecForLoad >> performCurrentVersionTestAgainst: vrsn operator:
...
+
+displayString := 'Project: ' , spec name.
+	
   spec versionString ~~ nil
     ifTrue: [ displayString := displayString , ' ' , spec versionString ].
-  MetacelloNotification signal: displayString.
+  	MetacelloNotification signal: displayString.
+
+	spec isLocked ifTrue: [ 
+		MetacelloNotification signal: 'Project is locked, skipping.'.
```
pavel-krivanek, line 96: we should use custom notification class or at least an exception tag here

pavel-krivanek (changes requested):

src/Metacello-Core/MetacelloProjectSpecForLoad.class.st
estebanlm: I use what Metacello uses to do logging, the intention is not other than the information to arrive to the transcript :P

pavel-krivanek (approved):

PR #10635, opened 2021-12-10
Cleanup: STONWriter categorization

### LINKED ISSUE #10634: Cleanup: STONWriter categorization
- in STONWriter many methods (even accessors) are badly categorized in 'initialize-release'
- the #initialize method should go into category 'initialization' (lint rule)

image

astares:
Fix #10634

### DIFF
--- a/src/STON-Core/STONWriter.class.st
+++ b/src/STON-Core/STONWriter.class.st
@@ -95,12 +95,12 @@ STONWriter class >> on: writeStream [
 		yourself
 ]
 
-{ #category : #'initialize-release' }
+{ #category : #accessing }
 STONWriter >> asciiOnly: boolean [
 	asciiOnly := boolean
 ]
 
-{ #category : #'initialize-release' }
+{ #category : #initialization }
 STONWriter >> close [
 	writeStream ifNotNil: [
 		writeStream close.
@@ -206,7 +206,7 @@ STONWriter >> error: aString [
 	^ STONWriterError signal: aString
 ]
 
-{ #category : #'initialize-release' }
+{ #category : #private }
 STONWriter >> escape: char with: anObject [
 	"Instruct me to escape char with object, either a replacement string or #pass"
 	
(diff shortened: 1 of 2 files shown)


svenvc:
I am sorry but you are changing semantics by moving things around to different categories.

Some of the methods that you call accessors are meant to be used during initialisation to set up the instance, and are not meant to be use arbitrarily afterwards.

You are moving some methods to private which were not meant to be there.

astares:
@svenvc 

1. the instance side #on: method is called from class side #on: method only - there is also no need to have it public
    as one does not change the stream while writing  
    => that's why it has been moved to 'private - initialization'

2. the #escape:with: should be in "private" as this is only called locally and one should not give an own custom
    escaping as the STON format is clearly defined 

    If it would stay public and one would give a different escaping (during initialization or whenever different from
    what the class itself configures in the instance) it would undermine the STON definition.

3. #initialize should go to "initialization" for sure as we try to have a common style among the classes for that and
    it also violates the lint rule (as seen in screenshot), this would actually satisfy the lint 

    
Having the other ones in
 A) in "initialization" which seems to be your POV as you see them used during initial setup of the instance before writing
 B) or distributed between "accessing" (due to their accessor nature) and "public" (for #optimizeForLargeStructures)

could be discussed. 

I understand you point that they are used initially and therefore should not just be "accessors" and it as valid as the other opinion (which is my POV) that a user would expect accessors in accessing and check that category anyway if he need different setup than default and wants to configure before writing.

So I can switch from B) to A) if you want me to adopt the PR. 

But the ones mentioned in 1 - 3 should stay (if there are no other detailed arguments against them) for the mentioned reasons above to fix the issue and have better categorization.

Just tell me how how we could/should proceed.

astares:
@svenvc: I provided an alternative PR in #10661 following what was written above. Could you have a look which one would be acceptable. Thanks in advance!

astares:
OK - we close that one and use #10661

PR #10636, opened 2021-12-10
10214 improving comments and adding tests to tree searcher

### LINKED ISSUE #10214: Improving comments and adding tests to TreeSearcher
TreeSearcher needs love...
We will add comment and tests

alesshosry:
Adding more tests + Enhancing class RBParseTreeSearcher

### DIFF
--- a/src/AST-Core-Tests/RBParseTreeSearcherTest.class.st
+++ b/src/AST-Core-Tests/RBParseTreeSearcherTest.class.st
@@ -55,10 +55,8 @@ RBParseTreeSearcherTest >> testDynamicArrayWithStatementListPattern [
 { #category : #'tests - to de refined' }
 RBParseTreeSearcherTest >> testSearchListComplex [
 
-	"		
-		- Here we looking for multiple lists in one match pattern at the same time. 
-		- Lists can be any receiver or args
-	"
+	"Here we are looking for multiple lists in one match pattern at the same time. 
+	Lists can be any receiver or arguments"
 
 	| dict |
 	searcher
--- a/src/AST-Core-Tests/RBParseTreeSearcherTest.class.st
+++ b/src/AST-Core-Tests/RBParseTreeSearcherTest.class.st
@@ -46,24 +46,370 @@ RBParseTreeSearcherTest >> testDynamicArrayWithStatementListPattern [
 
 	| dict |
 	searcher matches: '{`.@stmts.}' do: [ :aNode :answer |  dict:= searcher context ].
-	dict := searcher executeTree: (self parseExpression:'{ (1@2) . Color red . 3}').
+	dict := searcher executeTree: (self parseExpression:'{ (1@2) . Color red . 3 }').	
+	"self halt."
 	self assert: (dict at: (RBPatternVariableNode named: '`.@stmts')) size equals: 3
 
 ]
 
 { #category : #'tests - to de refined' }
+RBParseTreeSearcherTest >> testSearchListComplex [
+
+	"		
+		- Here we looking for multiple lists in one match pattern at the same time. 
+		- Lists can be any receiver or args
+	"
+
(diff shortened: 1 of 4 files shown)


Ducasse:
Thanks a lot Aless for your contribution. We will have a look. 
Having more tests is super nice. Thanks again.

src/AST-Core-Tests/RBParseTreeSearcherTest.class.st
```diff
@@ -46,24 +46,487 @@ RBParseTreeSearcherTest >> testDynamicArrayWithStatementListPattern [
 
 	| dict |
 	searcher matches: '{`.@stmts.}' do: [ :aNode :answer |  dict:= searcher context ].
-	dict := searcher executeTree: (self parseExpression:'{ (1@2) . Color red . 3}').
+	dict := searcher executeTree: (self parseExpression:'{ (1@2) . Color red . 3 }').	
+	"self halt."
```
gcotelli, line 50: Remove this debugging code, even if it's commented.

src/AST-Core-Tests/RBParseTreeSearcherTest.class.st
```diff
@@ -46,24 +46,487 @@ RBParseTreeSearcherTest >> testDynamicArrayWithStatementListPattern [
...
 
 ]
 
+{ #category : #'tests - to de refined' }
+RBParseTreeSearcherTest >> testSearchListComplex [
+
+	"		
+		- Here we looking for multiple lists in one match pattern at the same time. 
+		- Lists can be any receiver or args
+	"
```
gcotelli:
```suggestion
	"Here we are looking for multiple lists in one match pattern at the same time. 
	Lists can be any receiver or arguments"
```

src/AST-Core-Tests/RBParseTreeSearcherTest.class.st
```diff
@@ -46,24 +46,487 @@ RBParseTreeSearcherTest >> testDynamicArrayWithStatementListPattern [
...
+		matches: '`@receiver assert: `@arg equals: true'
+		do: [ :aNode :answer | dict := searcher context ].
+
+	"so in below example, there is no match with the suggested pattern. Which is why, dict is nil "
+
+	searcher executeTree:
+		(self parseExpression: 'self assert: token isLiteralToken.').
+
+	self assert: dict isNil
+	"self deny: dict isNotNil"
```
gcotelli:
```suggestion
	"Here we are looking for multiple lists in one match pattern at the same time. 
	Lists can be any receiver or argument"

	searcher
		matches: '`@receiver assert: `@arg equals: true'
		do: [ :aNode :answer | self fail ].

	searcher executeTree:
		(self parseExpression: 'self assert: token isLiteralToken.').
```

gcotelli (changes requested):
- Use proper names instead of `dict`
- Method comments are formatted in an strange way
- The cases when the searcher is not matching can be changed to do `self fail` in the do block instead of checking for nil (I'm proposing one of that changes in this review)
- There's a lot of almost repeated assertions, like
```
self
		assert:
		(dict at: (RBPatternVariableNode named: '`@arg')) formattedCode
		equals: 'offset + count'.
```
extract it to something receiving the context (dict) and the variable name, something like:
`assertCodeForVariableNodeNamed: name in: context equals: expectedCode`

MarcusDenker:
This PR just needs a pass over the comments of the reviewer

Ducasse:
AlessHo can you have a look?

MarcusDenker:
@AlessHo

AlessHo:
Hello, Okay

alesshosry:
Hello,

I just saw the messages.
Ok, noted for the comments I will adjust them this week.
Thank you for your reviews.

PR #10645, opened 2021-12-12
Fixes #10644 Process>>terminate - fix failing test etc.

### LINKED ISSUE #10644: Process>>terminate - fix failing test, correctly terminate non-local returns in ensure blocks and nested terminations etc.
I'd like to extend my work on #terminate (see #8993 recently merged) to fix failing testIsTerminatingForcedTermination (currently as expected failure) and further deficiencies like correctly terminate non-local returns in ensure blocks, correctly terminate nested errors in ensure blocks and nested terminations, and allow to safely debug BlockCannotReturn scenarios.

E.g. the following examples will be evaluated correctly and safely:
```
[[[] ensure: [^2]] ensure: [^42]] fork
"any debugging possible unless pressed Proceed"
```
```
[
	[
		[ ] ensure: [
			[self error: 'outer error'] ensure: [
				self error: 'inner error'.
				'x1' trace]. 
			'x2' trace]
	] ensure: [
		'x3' trace].
	'x4' trace
] fork.
Processor yield.
"should print x1 x2 x3"
```
```
[
	[ ] ensure: [
		[self error: 'unwind test'] ensure: [
			^'x1' trace]. 
		'x2' trace]
] ensure: [
	'x3' trace].
'x4' trace
"should print x1 x3"
```
The test testIsTerminatingForcedTermination can be removed from expectedFailure because it passes with the proposed PR.

The test testTerminationShouldProceedAllEnsureBlocksIfSomeWasFailed currently marked as expected failure can be removed because it refers to the UnwindError exception no longer used by the current version of #terminate.

A PR and tests will be submitted.
Regards,
Jaromir Matas

isCzech:
fix failing testIsTerminatingForcedTermination (currently as expected failure) and further deficiencies like correctly terminate non-local returns in ensure blocks, correctly terminate nested errors in ensure blocks and nested terminations, and allow to safely debug BlockCannotReturn scenarios.
E.g. the following examples will be evaluated correctly and safely:
[[[] ensure: [^2]] ensure: [^42]] fork
"any debugging possible unless pressed Proceed"
More scenarios in tests pushed separately soon.

This work extends and simplifies changes to #terminate merged  previously in #8993.


dionisiydk:
Did you forget to remove #testIsTerminatingForcedTermination from expected failures?

BTW there is a pragma <expectedFailure> to mark given test directly.

src/Kernel/Context.class.st
```diff
@@ -1772,6 +1772,31 @@ Context >> runUntilErrorOrReturnFrom: aSender [
...
 ]
 
+{ #category : #private }
+Context >> runUnwindTo: aContext onBehalfOf: aProcess [
+	"Run self on behalf of aProcess on aProcess's (i.e. self's) stack until aContext returns. Avoid a context that cannot return. Note: self must be a stack top context. Note: to avoid infinite recursion of MNU error: e.g. a wrapper around the message sentTo: receiver in #doesNotUnderstand: must be implemented.
+	This method is meant to be used exclusively by Process>>#terminate."
+
+	Processor activeProcess
+		evaluate: [ 
+			| here unwindBottom newTop |
```
dionisiydk: I would split this method on two parts: 
- Context>>runUnwindsTo: with this code inside #evaluate (or similar).
- And Process>>#runUnwindsBetween:to: which would wrap "aContext runUnwindsTo:" with #evaluate:onBehalfOf:. See other senders of #evaluate:onBehalfOf: . They are all on the Process side.

src/Kernel/Process.class.st
```diff
@@ -213,63 +213,46 @@ Process >> debugWithTitle: title [
...
+		suspendedContext := nil.  "Disable terminating process while running its stack in active process below.
+		If terminating a process halfways through an unwind, try to complete that unwind block first; if there are multiple such nested unwind blocks, try to complete the outer-most one; nested unwind blocks will be completed in the process; see testTerminationDuringUnwind, testNestedUnwind.
+		Note: Halfway-through blocks have already set the complete variable (unwindComplete) in their defining #ensure:/#ifCurtailed contexts from nil to true; we'll search for the bottom-most one.
+		Note: #findNextUnwindContextUpTo: starts searching from the receiver's sender but the receiver 	itself may be an unwind context (see testTerminateEnsureAsTopContext)."
+		ctxt := top.
+		ctxt isUnwindContext ifFalse: [ctxt := ctxt findNextUnwindContextUpTo: nil].
+		[ctxt isNil] whileFalse: [
+			ctxt unwindComplete ifNotNil: [
+				outerMost := ctxt].
+			ctxt := ctxt findNextUnwindContextUpTo: nil].
```
dionisiydk: Let's introduce Context>>findMostOuterUnwindContext

src/Kernel/Process.class.st
```diff
@@ -213,63 +213,46 @@ Process >> debugWithTitle: title [
...
+		Note: newTop sender points to the former outerMost sender, i.e. the next unexplored context."
+		ctxt := newTop ifNil: [top] ifNotNil: [newTop sender].
+		ctxt isUnwindContext ifFalse: [ctxt := ctxt findNextUnwindContextUpTo: nil].
+		[ctxt isNil] whileFalse: [
+			ctxt unwindComplete ifNil: [
+				ctxt unwindComplete: true.
+				newTop := ctxt unwindBlock asContextWithSender: ctxt.
+				newTop runUnwindTo: newTop onBehalfOf: self].
+			ctxt := ctxt findNextUnwindContextUpTo: nil].
+		suspendedContext := top setSender: nil receiver: self method: (Process>>#endProcess) arguments: {}
```
dionisiydk: And I would extract it into Context>>beDead method

isCzech:
> Did you forget to remove #testIsTerminatingForcedTermination from expected failures?
> 
Yes, I did :)

src/Kernel/Context.class.st
isCzech: yes, I was considering this and checking typical senders of #evaluate:onBehalfOf: but then I decided against; no "big" reason though: Context it just seemed to me a better place for a method manipulating context similarly as #runUntilErrorOrReturnFrom: . #runUnwindTo:onBehalfOf: was conceived by pruning #runUntilErrorOrReturnFrom: as much as possible :) But you're right, it could as well be under Process...

src/Kernel/Process.class.st
isCzech: I haven't made my mind about this yet. There might be some more changes here in the future and splitting the method into smaller ones makes it, for me at least, less readable. I need to see the whole code in this case :) (I don't do this generally but in this particular case with a tricky code like #teminate I prefer less fragmented code). When the code "settles" I'm definitely not against fragmenting. I just know how many times I reshuffled this code which is why I'm presenting it as an unfragmented method.

src/Kernel/Process.class.st
```diff
@@ -213,63 +213,46 @@ Process >> debugWithTitle: title [
...
+		terminating := true.
+		"Always suspend the process first so it doesn't accidentally get woken up"
+		oldList := self suspend.
+		suspendedContext ifNil: [^self].
+		"Figure out if we are terminating the process while waiting inside special object like in Semaphore>>critical:. So if waiting object is interesting on this situation we will ask it to handle it. In case of Semaphore>>critical, Semaphore will pop the suspendedContext so that we leave the ensure: block inside Semaphore>>critical: without signaling the semaphore.
+		This methods allow to not be restricted only on Semaphore case."
+		suspendedContext := oldList handleProcessTerminationOfWaitingContext: suspendedContext.
+
+		top := suspendedContext.
+		suspendedContext := nil.  "Disable terminating process while running its stack in active process below.
```
dionisiydk: I am not sure about the idea of nil'ed suspendedContext. When debugger shows ZeroDivide in my example I can check the process in the playground: 
- it is terminated: p isTerminated "==> true"
- p suspendedContext "==> is nil"

This state is lying because the process is not yet terminated: it is in the middle of termination.  Why this nil is required? We have #terminating flag which protects from multiple termination.

isCzech: Yes, #terminating could be enough... I don't think nil is "required" - I didn't use it for anything; it just seemed safer just nil it. suspendedContext of a terminating process has no longer any meaning during termination (except the last step) because it's not used during termination. But you're right it implicates #isTerminated value... I'll think about this. However, if there's no practical use for suspendedContext or isTerminated showing false during termination - I'd say it's more a philosophical question :)

src/Kernel/Context.class.st
isCzech: one of the reasons I put #runUnwind:onBehalfOf: under Context is I can emphasize the fact it's evaluated "on behalf of" in the name of the method; and other could be the process being terminated is the patient being operated on, so making it a receiver (the one object "responsible" for the action in the OO philosophy) didn't seem right :)

src/Kernel/Process.class.st
dionisiydk: > Yes, #terminating could be enough... I don't think nil is "required" - I didn't use it for anything; it just seemed safer just nil it. 

I prefer to keep the system simple. Any line of code raises the question "why it is here?". It's better to remove redundant code. Otherwise you should describe the need for this code with a test.

>> suspendedContext of a terminating process has no longer any meaning during termination (except the last step) because it's not used during termination. 

But it is used during termination. All unwinds are started from the suspendedContext. I think the unwind logic can better reflect the unwind steps: it can assign #newTop to #suspendedContext on each iteration. It will make the "outside termination" to be consistent with the "inside termination" (from the process itself).

>> However, if there's no practical use for suspendedContext or isTerminated showing false during termination - I'd say it's more a philosophical question :)

Having the system in the consistent state is very important. Especially when we implement such a very corner cases. It's always hard to debug and when in one place the process is shown as dead and in another place it is not dead that is a nightmare to figure things out.

dionisiydk (approved):
Good stuff

src/Kernel/Process.class.st
yes, you have a point indeed and I hear your arguments! My objective now is to make sure it works flawlessly (and I have tested it quite thoroughly) before any further polishing. But I'm working on your comments, thanks again!

isCzech: Hi, thanks for these great points!
> > Yes, #terminating could be enough... I don't think nil is "required" - I didn't use it for anything; it just seemed safer just nil it.
> 
> I prefer to keep the system simple. Any line of code raises the question "why it is here?". It's better to remove redundant code. Otherwise you should describe the need for this code with a test.
> 
Oh, I know why - if the suspendedContext is not nil the process being terminated can be resumed (e.g. by a higher priority process or thanks to preemption) and that would interfere with its termination in progress. Setting suspendedContext to nil prevents such scenario (see Process >> resume). In case this makes a good sense, I will update the comment; even I forgot the reasoning :D 

> > > suspendedContext of a terminating process has no longer any meaning during termination (except the last step) because it's not used during termination.
> 
> But it is used during termination. All unwinds are started from the suspendedContext. I think the unwind logic can better reflect the unwind steps: it can assign #newTop to #suspendedContext on each iteration. It will make the "outside termination" to be consistent with the "inside termination" (from the process itself).
> 
yes, right; I meant it has no reasonable meaning beyond the point the real unwinds start (i.e. after top := suspendedContext). The point is even if I set suspendedContext to newTop each time, its consistency is lost when the active process jumps onto the terminating process's stack and starts executing it's code. I remember when I first drafted this new approach I used suspendedContext the way you suggest but it kept failing for some reason so I gave up and left it as nil and used a variable instead (I'm not saying it's wrong; I just didn't manage to make it work that way).

> > > However, if there's no practical use for suspendedContext or isTerminated showing false during termination - I'd say it's more a philosophical question :)
> 
> Having the system in the consistent state is very important. Especially when we implement such a very corner cases. It's always hard to debug and when in one place the process is shown as dead and in another place it is not dead that is a nightmare to figure things out.
>
yes, I couldn't agree more; consistency is absolutely key; however in this case I don't know how to make isTerminated showing false during termination when I need suspendedContext = nil without further complicating isTerminated; actually maybe we could make isTerminated false if isTerminating = true -?? That should fix the consistency issue, what do you think? I'm not sure if all tests would agree with this change right away but it could be dealt with independently of this PR. Other option could be to change Process >> resume to take isTerminating into account and then we could remove nilling suspendedContext and leave isTerminated as is -?? (still independently of this PR)

dionisiydk: > Oh, I know why - if the suspendedContext is not nil the process being terminated can be resumed (e.g. by a higher priority process or thanks to preemption) and that would interfere with its termination in progress. Setting suspendedContext to nil prevents such scenario (see Process >> resume). In case this makes a good sense, I will update the comment; even I forgot the reasoning :D

Right, I see that #resume method contains nil check for suspendedContext.
I think for the real solution we should explicitly model the process status - possibly replace #isTerminating flag with #status variable (choose a better name) and reify the ProcessStatus hierarchy. It will allow to decouple the life cycle state of a process from its execution state like suspendedContext, pc, endProcess method.

Anyway it's for future work. PR is already a great improvement.

isCzech: Hi @dionisiydk , you wrote
> I think the unwind logic can better reflect the unwind steps: it can assign #newTop to #suspendedContext on each iteration.

FYI, you're right, I tried it and it works ok and makes isTerminated answer correctly; the only issue is you're not protected against accidental resumption of the terminating process.

Ducasse:
Hi jaromir
Thanks a lot! Now pay attention because you have a lot of changes in your commit. 
May be you should merge first P11 in your branch to realign it. 

S

isCzech:
Hi Stephane,
My apologies for the mess… I don’t know what happened, only wanted to push one commit (512bc5d) but when my repo synchronized it pushed 118 commits, never happened before but it’s surely my mistake :(
Thanks; have a nice day,
Jaromir

--

Jaromír Matas

***@***.***

Ducasse:
Do not worry :)
Probably your version was far and got confused.

MarcusDenker:
replaced by https://github.com/pharo-project/pharo/pull/11308, I close this

PR #10646, opened 2021-12-12
Make Athens not coupled with FT2

estebanlm:
because we need to be able to load Athens without FreeType fonts (most of the times, when using Athens, you want to use the Athens primitives for that).


estebanlm:
oops. This one needs the FT2 approved first, I will comeback later!

MarcusDenker (approved):

PR #10652, opened 2021-12-13
10651-Resuming-a-exception-that-has-been-signalled-twice-does-not-return-correctly

### LINKED ISSUE #10651: Resuming a exception that has been signalled twice does not return correctly
Resuming an exception that has been signalled twice does not return correctly the value of the resume.

```Smalltalk
testResignalExceptionThatHasBeenSignaledTwice

	| shouldBe17 |
	shouldBe17 := nil.

	[ shouldBe17 := self methodSignallingZeroDivide ]
		on: ZeroDivide
		do: [ :e | e resume: 17 ].

	self assert: shouldBe17 notNil
```

```Smalltalk
methodSignallingZeroDivide

	^ [ 1 / 0 ]
		  on: ZeroDivide
		  do: [ :e | e signal ]
```

In Pharo 9 and previously to the change in PR #10429 the return value is correctly resumed.

tesonep:
There has been a regression as the same exception is not returning to the correct signalling context when resuming.

- Returning to the previous change introduced in #10429
- Fix #10651
- Adding Tests

### DIFF
--- a/src/Kernel-Tests/ExceptionTest.class.st
+++ b/src/Kernel-Tests/ExceptionTest.class.st
@@ -30,6 +30,14 @@ ExceptionTest >> assertSuccess: anExceptionTester [
 	self should: [ ( anExceptionTester suiteLog first) endsWith:  'succeeded'].
 ]
 
+{ #category : #utilities }
+ExceptionTest >> methodSignallingZeroDivide [
+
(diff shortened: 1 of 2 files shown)


MarcusDenker:
just randomly failing tests:
- [CI] Failing randomly: StPlaygroundPageTest>>testContents #10519

PR #10653, opened 2021-12-13
Register bootstrapped projects

estebanlm:
Boostrapped external projects need to be registered into metacello registry (otherwise metacello ignores their existence and that can have consequences).


estebanlm:
oops... this needs also another PR before. I need to stop branching from branchs :P

pavel-krivanek (approved):

PR #10657, opened 2021-12-13
Feature proposition : Anonymous visitor

hogoww:
We were talking about the possibility of creating an anonymous visitor with @MarcusDenker a while back.
We kinda forgot about it for a while.
As I am needing it now, I fixed it for the new AST and I'm sharing it in case this interest enough people to merge.
(The code is fully generated by a custom tool, so I haven't added the generation with the PR)

Pierre.

### DIFF
--- a/src/AST-Core/RBAnonymousVisitor.class.st
+++ b/src/AST-Core/RBAnonymousVisitor.class.st
@@ -1,0 +1,364 @@
+"
+For the Class part:  I represent a visitor for the RB AST which allows to dynamically set the behavior for each kind of nodes.
+
+You can register a block for any kind of node using the message #registerBlock:for:
+The kind of node is taken from the visit method name: visitMethodNode => #methodNode
+I works as any other visitor !
+ 
+Example:
+Counting the number of argument nodes
+| visitor count |
+count := 0.
+visitor := RBAnonymousVisitor new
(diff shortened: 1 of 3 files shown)


MarcusDenker:
Interesting! I will have a look and play with it.

Question: with Pharo10 in Freeze, would it make sense to have this in it's own github repo for now?

hogoww:
I had it in a private repo so far, so maybe.
It's a very small project, which can depend on RB.
Could make sense to have a its own projects for different kinds of visitors.

I'd love to add something I tend to "reify" the anonymous visitor in a class.
Basically : anonymousVisitor reify as 'MyCoolVisitor'.
Which would create the MyCoolVisitor class, and add the methods from the register blocks.

Anyway, lots of other stuff are doable.

MarcusDenker:
I am not against adding it to Pharo eventually... but we should be a bit conservative... (I used to just be in favor of adding early, but got a bit more cautious over time...)

 e.g. it is always good to have a user for something before integrating it.

guillep:
I particularly do not like the use of symbols to identify cases.
This is at least brittle: a simple typo leads to a missbehavior.
Having different setters/getters would make the code more clear and easy to manage (and fail early with a DNU in the case of an error!)

The PR has no tests, and to me not a clear motivation of why we would want to have this...

Also, I'd not put this in Pharo 10 as we are already in freeze

PR #10662, opened 2021-12-15
fix a couple of regressions

estebanlm:
for some reasons, this changes were rolled back in other PRs (probably my fault ;) )

### DIFF
--- a/src/Keymapping-Pragmas/PragmaMenuAndShortcutRegistration.extension.st
+++ b/src/Keymapping-Pragmas/PragmaMenuAndShortcutRegistration.extension.st
@@ -1,12 +1,12 @@
 Extension { #name : #PragmaMenuAndShortcutRegistration }
 
-{ #category : #'*Keymapping-Core' }
+{ #category : #'*Keymapping-Pragmas' }
 PragmaMenuAndShortcutRegistration >> attachShortcutCategory: aSymbol to: aClass [ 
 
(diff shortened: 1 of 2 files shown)


MarcusDenker:
interesting: windows crash with

```
fred is number 1
charlie is number 2
elmer is number 32021-12-15T11:54:21.468+01:00
WARNING:  Had to GC to make room for more external objects.
If this happens often, it would be a good idea to either:
- Raise the maxExternalSemaphores size.
- Write your code to explicitly release them rather than wait for finalization.

Returning an object but pool is full./pharo: line 11: 24135 Segmentation fault      "$DIR"/"pharo-vm/PharoConsole.exe" --headless "$@"
```


But that seems to be not related to the change

MarcusDenker (approved):

MarcusDenker:
The tfPointerAddress on ByteArray...  was this added somehow?

MarcusDenker:
tfPointerAddress on Bitmap I mean...

estebanlm:
yes, I moved it to `ThreadedFFI-Graphics-Extensions`, but for some reason it was not removed from `ThreadedFFI` causing a duplicate. The final image has  `tfPointerAddress` correct (since ThreadedFFI-Graphics-Extensions overrides ThreadedFFI),

estebanlm:
yeah, the windows error is not related.

PR #10665, opened 2021-12-15
Remove ZnUrlCopy

### LINKED ISSUE #10664: Remove ZnUrlCopy
The latest image contains ZnUrlCopy. I read the class comment.

This is horrible and incomplete as well.

More than a year ago https://github.com/svenvc/zinc/commit/7a42881f14144e0d9dcc2b3aa98a075f189eae69 added the preservation of URL query parameter order.

Of course, merging all of Zinc is hard because the many cosmetic changes that took place. I want to do it, but I can't do it alone.

The problem at hand can be fixed easily by changing the superclass of ZnMultiValueDictionary to OrderedDictionary. That's it.

Then ZnUrlCopy can be removed and the 2 users can revert back to ZnUrl.

I'll try to do a PR.

astares:
Fix #10664

### DIFF
--- a/src/Deprecated10/ZnUrlCopy.class.st
+++ b/src/Deprecated10/ZnUrlCopy.class.st
@@ -1,0 +1,913 @@
+"
+PLEASE DO NOT USE - see https://github.com/pharo-project/pharo/issues/10664
+
+I'm a copy of ZnUrl because ZnrUrl query does not use an dictionary garantying ordered keys.
+May be we could change ZnUrl to be able to change the dictionary used there...
+I do not have the time to address this point.
+
+"
+Class {
+	#name : #ZnUrlCopy,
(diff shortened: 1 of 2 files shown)


astares:
- class was deprecated and moved to Deprecated10
- class comment is pointing to #10664
- the two users have been switched to use ZnUrl

svenvc:
Thanks.

I don't see the change in users.

I would not move it to deprecated, just delete it: it was a temporary measure, no one uses it.

astares:
The other changes are in "Microdow"n - which is managed externally. Unfortunately not like other external packages (it does not work with clear releases). 

image

So I guess we need some help here.

Ducasse:
It will be removed with the new version of MD we are preparing.

Issue #10676, opened 2021-12-28
Improve performance of Month class>>indexOfMonth:


chisandrei:
When parsing the `DateAndTime` objects the method `Month class>>indexOfMonth:` is called. This ends up doing:

```
Month class>>#indexOfMonth: aMonthName 
	1 to: 12 do: [ :i |  (aMonthName, '*' match: (MonthNames at: i)) ifTrue: [^i] ].
 	self error: aMonthName , ' is not a recognized month name'.
```

`aMonthName` however is a symbol as defined in `ChronologyConstants`. This means `aMonthName, '*'` creates a `Symbol` which is slow.

A different way to implement this method would be as below. This does not use a symbol conversion nor calls match:

```
Month class>>#indexOfMonth: aMonthName 
	1 to: 12 do: [ :i | ((MonthNames at: i) beginsWith: aMonthName) ifTrue: [^i] ].
 	self error: aMonthName , ' is not a recognized month name'.
```

More details at https://github.com/feenkcom/gtoolkit/issues/2310

guillep:
Thanks for catching that!

I have two comments:
 - first, using `beginsWith:` changes a bit the semantics: `match:` is case insensitive, while the `beginsWith:` is case sensitive. See for example: `'asd' match: 'ASD' "true"`
 - the other, I think your proposal can be further enhanced by just using an `indexOf:`. In my machine doing something like `1 to: 12 do: [ :i | ((MonthNames indexOf: aMonthName) ~= 0) ifTrue: [^i] ].` is in average ~4x faster than your solution, but still losing the case insensitivity

Now, as it is this is a an API breaking change, in the core image there are not many users, but It's difficult to say if there are external users, so to take with care.

chisandrei:
Indeed, `match:` is cases insensitive. Made another change where I switched to `beginsWith:caseSensitive:`.  

But also this still introduces an API change as something like `Month indexOfMonth: '*ebruar'` will no longer work. Not sure if there are any users of that, or if that behaviour is wanted.

The smallest change that would avoid creating symbols would be `1 to: 12 do: [ :i |  (aMonthName asString, '*' match: (MonthNames at: i)) ifTrue: [^i] ].`; so just calling `asString` to avoid the symbol creation. But also using `beginsWith:caseSensitive:` seems reasonable.

chisandrei:
[...]

PR #10677, opened 2021-12-28
Improve #indexOfMonth: to not create symbols and compare without using #match

chisandrei:
Possible fix for https://github.com/pharo-project/pharo/issues/10676.  Update`Month class >> indexOfMonth:` to not call `match:` and avoid creating symbols.

### DIFF
--- a/src/Kernel-Chronology-Extras/Month.class.st
+++ b/src/Kernel-Chronology-Extras/Month.class.st
@@ -46,7 +46,7 @@ Month class >> february [
 { #category : #accessing }
 Month class >> indexOfMonth: aMonthName [ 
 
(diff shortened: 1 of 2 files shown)


src/Kernel-Chronology-Extras/Month.class.st
```diff
@@ -45,8 +45,8 @@ Month class >> february [
 
 { #category : #accessing }
 Month class >> indexOfMonth: aMonthName [
```
Bajger, line 47: Can you provide example calls here (e.g.:  `Month indexOfMonth: 'Jan' >>> 1` or similar )?

Bajger (changes requested):
Please add few examples of `indexOfMonth:` method, so it could be explored using finder.

MarcusDenker (approved):
The problem from the issue discussion of being case insensitive is solved

I think this is good to merge

PR #10682, opened 2022-01-03
Fixes #10673 by adding missing method method to display a MessageTally identified

### LINKED ISSUE #10673: DNU in MessageTally>>#displayStringOn:
**Bug description**
`MessageTally>>#displayStringOn:` contains send of unimplemented method `displayIdentifierOn:`.

**To Reproduce**
1. evaluate `(MessageTally spyOn: [ self notNil ]) inspect`

carolahp:
Test solution executing
```
mt := MessageTally  
	spyOn: [ self notNil ]
	reportOtherProcesses: false 
	cutoff: 1 
	openResultWindow: true 
	closeAfter: false. "if true, the instance will be closed immediately and its identifier will not be displayed"
mt inspect.
mt close
```

### DIFF
--- a/src/Tool-Profilers/MessageTally.class.st
+++ b/src/Tool-Profilers/MessageTally.class.st
@@ -544,11 +544,10 @@ MessageTally >> into: leafDict fromSender: senderTally [
 { #category : #testing }
 MessageTally >> isClosed [
 
-	^ Timer isNil 
-		and: [ class isNil 
+	^ class isNil 
 		and: [ method isNil 
 		and: [ tally isNil 
-		and: [ receivers isNil ] ] ] ]
+		and: [ receivers isNil ] ] ] 
 ]
 
 { #category : #comparing }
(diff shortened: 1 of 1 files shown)


Ducasse:
In P9 we got

```
MessageTally >> displayIdentifierOn: aStream
	class displayStringOn: aStream.
	self method methodClass ~~ class 
		ifTrue: [ aStream nextPut: $(; print: self method methodClass; nextPut: $) ].
	aStream 
		nextPutAll: '>>';
		store: self method selector
```

Ducasse:
@carolahp may be this is better to use the version of P9 and not to have the test to the Time class variable

MarcusDenker:
review asks question: more work needed

carolahp:
> @carolahp may be this is better to use the version of P9 and not to have the test to the Time class variable

@Ducasse the version of P9 fails when executing  ```(MessageTally spyOn: [ self notNil ]) inspect```
because the instance variables method and class are set to ```nil``` at the end of ```spyOn:```

Maybe I could check only these IVs and not the class var ```Time``` in my method ```isClosed```?

Ducasse:
Ok :) Yes I found strange to check the status of a class variable in a printOn because the class variable looks like a global

MarcusDenker:
Just randomly failing test:

MarcusDenker (approved):

PR #10683, opened 2022-01-04
add-SystemDictionary-loopUpVarDeclare

MarcusDenker:
This PR adds the lookupVar:declare: method, this is needed for https://github.com/pharo-spec/Spec/pull/1220


MarcusDenker:
as this is just an used method,  I merge it

PR #10684, opened 2022-01-04
Fixes pharo-spec/Spec#1219

carolahp:
Fixing the refactoring preview visualization by remembering the scrollbar position and setting it again after UI rebuild. 

@Ducasse you have more examples of lists scrolling up after ticking an item?

This solution breaks encapsulation of FTTableMorph by writing a private variable from outside, but it is simpler than fixing the calculations used to get the first shown item from the scroll bar position

### DIFF
--- a/src/Morphic-Widgets-FastTable/FTTableMorph.class.st
+++ b/src/Morphic-Widgets-FastTable/FTTableMorph.class.st
@@ -1111,6 +1111,11 @@ FTTableMorph >> showIndex [
 	^ showIndex
 ]
 
+{ #category : #private }
+FTTableMorph >> showIndex: anInteger [
+	showIndex := anInteger
+]
+
 { #category : #menu }
 FTTableMorph >> showMenuForIndex: aTuple [
 	| menu rowIndex columnIndex |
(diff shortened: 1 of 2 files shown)


Ducasse:
The problem is that if we have to fix all the users then there is something wrong. 
@estebanlm Esteban do you have an idea?

estebanlm:
I am sorry, this solution is bad. 
1) It breaks layers by going directly to the morphic widget.
2) the presenter uses very-incorrectly a rebuild mechanism (probably inherited from old spec1), but this is wrong and it will not work correctly on spec2 (and very probably will make the component leak).

Spec2 already has the api `#updateItemsKeepingSelection:` which should be the desired behavior instead rebuilding the component.

In anycase, let's discuss the issue because I do not even understand what is a "ticking" window :P 

(closing this)

estebanlm:
It would be also possible to add an api `#updateItemsKeepingDisplayedViewport:` if needed, but let's not start hacking around when we can fix/improve Spec :)

Ducasse:
@estebanlm I agree with you this is why I put you in the loop. 
Now the PR of Caro was to fix https://github.com/pharo-spec/Spec/issues/1219
which is super annoying when you have long lists. 
Do you think that #updateItemsKeepingSelection: scroll to show the item that got unselected?
Because in fact the user want to see the following one, right now you have to scroll down and find the position after every unselection of a tick. Is the #updateItemsKeepingDisplayedViewport: what you have in mind for this
problem? Can you give some indication to Caro?
Can we remove the rebuild: method if is obsolete?

estebanlm:
well, that's why I suggested to implement an api `updateItemsKeepingDisplayedViewport:` to keep the displayed area even if no selected item. But I should check what that presenter does to see if we can replace it for something more solid (and remove the obsolete stuff)

PR #10686, opened 2022-01-05
Fixes #10263 Setting minimum size for browser window

### LINKED ISSUE #10263: Calypso browsers should have Minimal Size
When it is resized smaller than the size, Calypso hides the widgets.

carolahp:


carolahp:
minimum size is used for browser and senders/implementors window

Issue #10522, opened 2021-11-29
[CI] Failing randomly: Spec related tests with DNU SmallFloat64 >> #nextCell:


MarcusDenker:
**Bug description**
Sometimes Spec related tests fail with a DNU  SmallFloat64 >> #nextCell:

Happend on linux in  #10508

```
Error
Message not understood: SmallFloat64 >> #nextCell:
Stacktrace
MessageNotUnderstood
Message not understood: SmallFloat64 >> #nextCell:
SmallFloat64(Object)>>doesNotUnderstand: #nextCell:
SpMorphicBoxLayoutCalculatorLeftToRight>>calculateCellFor:
[ :each | self calculateCellFor: each ] in SpMorphicBoxLayoutCalculatorLeftToRight>>calculateFor:in:
Array(SequenceableCollection)>>do:
Morph>>submorphsDo:
SpMorphicBoxLayoutCalculatorLeftToRight>>calculateFor:in:
SpMorphicBoxChildLayout(SpMorphicBaseBoxLayout)>>layoutLeftToRight:in:
SpMorphicBoxChildLayout(TableLayout)>>layout:in:
Morph>>doLayoutIn:
[ self doLayoutIn: self layoutBounds ] in Morph>>computeFullBounds
FullBlockClosure(BlockClosure)>>on:do:
Morph>>computeFullBounds
Morph>>fullBounds
[:m | | subBox | 
		(m visible) ifTrue: [
			subBox := m fullBounds.
			box 
				ifNil:[box := subBox copy]
				ifNotNil:[box := box quickMerge: subBox]]] in Morph>>submorphBounds
Array(SequenceableCollection)>>do:
Morph>>submorphBounds
Morph>>privateFullBounds
Morph>>doLayoutIn:
[ self doLayoutIn: self layoutBounds ] in Morph>>computeFullBounds
FullBlockClosure(BlockClosure)>>on:do:
Morph>>computeFullBounds
Morph>>fullBounds
Morph>>minExtent
[:m| | sizeX sizeY |
		m disableTableLayout ifFalse:[
			n := n + 1.
			size := m minExtent. sizeX := size x. sizeY := size y.
			sizeX < minX
				ifTrue:[sizeX := minX]
				ifFalse:[sizeX > maxX ifTrue:[sizeX := maxX]].
			sizeY < minY
				ifTrue:[sizeY := minY]
				ifFalse:[sizeY > maxY ifTrue:[sizeY := maxY]].
[...]

PR #10697, opened 2022-01-08
align commit of external projects with its merged version

estebanlm:
this pull request changes the way external projects are declared when included in the image.
before, this was a hard coded commit making the number to be out of sync in development phase (when external projects are pointing to a branch and not a tag) making the development process more cumbersome than needed. 

With this change, we search and add the commit as part of the bootstrap, making the build be always pointing to the exact merged commitish.


Ducasse:
Once this is in place I can try with microdown and BC I'm working on a release.

MarcusDenker:
just a random test failiure:

PR #10707, opened 2022-01-11
10678-Step-Over-fails-for-leaving-simple-accessors 

### LINKED ISSUE #10678: Step Over fails for leaving simple accessors
**Bug description**
When you step into a simple accessor method and the leave it using "step over", bad things happen.

**To Reproduce**
1. Load this simple class with one instance variable, accessor and a method with halt that calls the accessor
   [TstClass.zip](https://github.com/pharo-project/pharo/files/7793737/TstClass.zip)
2. call `TstClass new msg`
3. in the debugger, do two times "Into"
4. you are now in the `accessor` method, do "Over`
5. `PrimitiveFailed: primitive #objectAt: in CompiledMethod failed`

image

**Version information:**
Pharo 10.0.0
Build information: Pharo-10.0.0+build.354.sha.37934d7c40480f096548689ab93d4cad88f0344b (64 Bit)

MarcusDenker:
This is a workaround to fix #10678

For a real fix we need to check if we really need to do the check and call to updatePCForQuickPrimitiveRestart for every step... this should only be done when starting (or restarting) to debug a method.

I will open a new issue tracker for that.

### DIFF
--- a/src/Debugger-Model/DebugSession.class.st
+++ b/src/Debugger-Model/DebugSession.class.st
@@ -564,7 +564,7 @@ DebugSession >> stepToFirstInterestingBytecodeIn: aProcess [
 	make sure that we step correctly over the first primitive bytecode"
 	| suspendedContext |
 	suspendedContext := aProcess suspendedContext.
-	suspendedContext method isQuick
+	(suspendedContext method isQuick and: [ suspendedContext pc == suspendedContext method initialPC ])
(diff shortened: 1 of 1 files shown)


pavel-krivanek (commented):
I tested it for several days and it works as expected

MarcusDenker:
randomly failing SpDemoTest>>testSmokeTestForDemoPages, just on the mac

PR #10728, opened 2022-01-14
10730 Update comments kernel wrt responsibilities/behavior

ErikOnBike:
Some comments are not fully accurate or reveal too much implementation detail. This update hopes to make the comments more clear, especially for newcomers.
See https://github.com/pharo-project/pharo/issues/10730


Ducasse:
Thanks Erik.

Issue #10265, opened 2021-10-22
Improve Breakpoints


tesonep:
- [x] Working in methods
- [x] Working in blocks
- [x] Adding / Removing does not alter the timestamp
- [x] Removing returns to the previous method without affecting the code

MarcusDenker:
For the third point it seems to already work like that (timestamp is not changed).  I added a test: https://github.com/pharo-project/pharo/pull/10324

carolahp:
@MarcusDenker I leave here the first UI improvements we just spoke about

- [x] **Improve Debugging submenu**

    When inserting a breakpoint, using the current debugging submenu, it is hard to know where the breakpoint will be inserted. 
    For example:

    <img src="https://user-images.githubusercontent.com/4822303/149376915-ca34ffd3-e189-4d10-953c-4d18902b73e3.png" width=50% height=50%>


    To solve this, the submenu should display the code section where the breakpoint will be inserted, similarly to:

    <img src="https://user-images.githubusercontent.com/4822303/149371780-03f3268e-475e-463f-bc46-9bf720b7cf3e.png" width=50% height=50%>


- [x] **Prevent accidental breakpoint insertion**

    A single click on the bar of the code editor inserts a breakpoint, and no confirmation is asked (this can freeze the system in core methods).
    
    A single click should open the debugging menu (as I just discovered is currently the case when right-clicking).
    And breakpoints should be inserted using double click (@MarcusDenker what do you think?)

MarcusDenker:
all TODOs are done, so we can close this issue

PR #10736, opened 2022-01-15
Use double click on the left bar to install a breakpoint in the browser

dionisiydk:
It switches the breakpoint installation to the double click instead of single click according to proposal from #10265 . 
And the highlighting of ast node is added for breakpoint menu activation.

The breakpoint menu is still activated with right click on the left side bar. The single click  does not work properly because it conflicts with double click event. To see the problem you can play with this code:
```Smalltalk
CmdOpenTextLeftBarMenuCommand>>contextClickActivation
	<classAnnotation>
	^CmdTextLeftBarClickActivation for: CmdToolContext "it was byYellowButtonFor: "
```


Ducasse:
Thanks Denis!

Ducasse:
Broken tests are unrelated.

PR #10738, opened 2022-01-15
TFFI not compatible to UFFI - missing #tfPointerAddress on Integer

### LINKED ISSUE #10737: TFFI not compatible to UFFI - missing #tfPointerAddress on Integer
ThreadedFFI should be compatible to UFFI but there is a method missing:

image

as it was discussed on Discord it can be easily added:

```Smalltalk
tfPointerAddress 

    ^ExternalAddress new fromInteger: self
```

astares:
Fix #10737 by adding missing #tfPointerAddress

### DIFF
--- a/src/ThreadedFFI/Integer.extension.st
+++ b/src/ThreadedFFI/Integer.extension.st
@@ -1,0 +1,7 @@
+Extension { #name : #Integer }
+
+{ #category : #'*ThreadedFFI' }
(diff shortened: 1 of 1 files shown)


astares:
**CI Problem is unrelated** (but something that might require a closer look. So I opened https://github.com/pharo-project/pharo/issues/10756)

Bajger (approved):

PR #10749, opened 2022-01-15
Do not use deprecated method in ClyNotebookPageMorph>>#initialize

### LINKED ISSUE #10748: Do not use deprecated method in ClyNotebookPageMorph>>#initialize
- #buildWithSpec is now deprecated and we should use #build
- this is now enabled in latest Pharo 10 - so once Calypse is opened this deprecation transforms this method and makes the "Calypso-Browser" package dirty

image

astares:
Fix #10748


Ducasse:
Unrelated broken tests (It looks like).

PR #10758, opened 2022-01-16
Fixes #10740 Expand on non fluid does not raise a DNU anymore

### LINKED ISSUE #10740: Missing expandedClassDefinitionString for OldPharoClassDefinitionPrinter
This is strange but we lost it.
- Select a class in the class pane
- Select the class name in the definition
- Bring the menu and choose expand.
- It breaks.

Ducasse:


Ducasse:
The failure is super strange. I do not get it even if it looks related.

PR #10775, opened 2022-01-20
10695-CmdBrowse-raise-a-DNU

### LINKED ISSUE #10695: Cmd+Browse raise a DNU
And now I have a reproducible case in latest Pharo10.

- Type a class name of a class that does not exist eg. Poin
- Cmd+Browse
- in the pop up list pick Point
- Boum

```
visitVariableNode: aVariableNode
	| var |	
	var := self resolveVariableNode: aVariableNode.
	var analyzeRead: aVariableNode by: self
```
analyzeRead: was sent to nil

MarcusDenker:
This fixes the #analyzeRead:by: DNU that we see when using "Browse it…"  (this happned for "Print it", too, though).

This fixes the base problem, but it is only working in the Code Browser for now.

For the Playground, it will work after we integrate a missing method (see https://github.com/pharo-spec/NewTools/pull/334)

fixes #10695

### DIFF
--- a/src/OpalCompiler-Core/OCUndeclaredVariableWarning.class.st
+++ b/src/OpalCompiler-Core/OCUndeclaredVariableWarning.class.st
@@ -196,8 +196,7 @@ OCUndeclaredVariableWarning >> openMenuIn: aBlock [
 			labels add: each.
 			actions
 				add: [ 
-					self substituteVariable: each atInterval: interval.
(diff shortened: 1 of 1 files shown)


MarcusDenker:
Random failing test:

testWorkerPoolDoesNotExceedPoolSizeWhenSchedulingTasksInParallel

PR #10784, opened 2022-01-21
Cleanup-canBeExecutedInContext

MarcusDenker:
after doing the isAbstract check in the superclass (see #10770), we can omit it from all the methods that override it if we make sure to do a super call.

In addition:
- fix some super calls that were not returning anything 
- make sure the methods are all categorized the same 
- remove some Pharo7 specific code in ClyIcebergShowMethodVersionCommand


Ducasse:
Reviewing code with super send is difficult. 
I notice that there is a dependency from calypso to iceberg.

MarcusDenker:
the dependcy was already there, this PR is not there to fix that, I just removed the issue described in the "self flag:", the
code that was there for Pharo7.

Yes, we should experiment to review from the image, text is not a good medium...

PR #10793, opened 2022-01-21
Rename defaultSpec to defaultLayout

demarey:


src/Refactoring-Core/RBReplacePreview.class.st
```diff
@@ -100,7 +100,7 @@ RBReplacePreview >> initializeWidgets [
 
 { #category : #'instance creation' }
 RBReplacePreview >> open [
-	^ self "openDialogWithSpec"openModalWithSpec
+	^ self "openDialogWithSpec"openModal
```
astares, line 103: @demarey 
should the comment not also have "openDialog" instead of "openDialogWithSpec" ?
As it is merged I guess this needs to be fixed with a new issue.

demarey: yes but more probably, the comment should be removed

PR #10795, opened 2022-01-22
Remove Pharo 9 from Metacello attributes

jecisc:
Fixes #10794

PR #10797, opened 2022-01-22
Revert "Remove Pharo 9 from Metacello attributes"

Ducasse:
Reverts pharo-project/pharo#10795


Ducasse:
Let us check why the build failed.

jecisc:
Probably a dependency of Pharo whose dependencies are set for Pharo 9 but not 10

jecisc:
It seems to be TaskIt

Issue #8631, opened 2021-02-25
#allInstVarNames should be #allInstanceVariableNames


astares:
We should rename allInstVarNames into #allInstanceVariableNames and deprecated #allInstVarNames (if possible)

MarcusDenker:
we use the abbreviated forms for compatibility with the "old" ivars, too. (e.g. to have instVarAt:), while the long form is for "Variables as Objects".

It used to be that allInstVarNames maps to the offsets of ivars, while slots do not (there can be slots that have no index). 

That is, the offset in #allInstVarNames should be a valid argument for #instVarAt:, while #allInstanceVariableNames should be in sync with #allInstanceVariables which is #allSlots.

MarcusDenker:
The todo here is:
- add allInstanceVariableNames to ClassDescription
- make sure all users of #allInstVarNames that are not reasoning about offsets use it
- change #allInstVarNames to only return slots with offsets to be in sync with #instVarAt:
- Check all users of #instVarAt: --> all clients should be rewritten in terms of the first-class variable API. 

(this is not for Pharo10)

MarcusDenker:
I was looking at this again and I think for now it is better to keep it as it is: short names -> old API, compatible with ST80.

With Variables being objects, we need the ...names methods less, and I would not add things like allSlotNames. Just use #allSlots and #slots (and get the name from there if needed).

PR #10811, opened 2022-01-27
Fix #8630: fix allInstVarNames for Traits with State

### LINKED ISSUE #8630: Traits with State: #allInstVarNames not in sync with #allSlots
**Describe the bug**
This should be true:

```
TDebuggerTest class allInstVarNames size = TDebuggerTest class allSlots size
```

This leads to one of the problems shown by issue https://github.com/pharo-project/pharo/issues/7952

MarcusDenker:
- add #testAllInstVarNames
- Implement allInstVarNames in terms of #allSlots in ClassDescription

We should make sure later that #allInstVarNames is in sync with #instVarAt: for the ST80 offset API (as soon as we use non-offset-based Slots for real).

For that, we have another issue tracker entry: https://github.com/pharo-project/pharo/issues/8631

This PR cleans some unused Pragma from Spec in addition (not really needed as that code will be removed anyway)

### DIFF
--- a/src/Kernel-Tests/BehaviorTest.class.st
+++ b/src/Kernel-Tests/BehaviorTest.class.st
@@ -25,6 +25,20 @@ BehaviorTest >> sampleMessageWithFirstArgument: firstArgument "This is a comment
 ]
 
 { #category : #tests }
+BehaviorTest >> testAllInstVarNames [
+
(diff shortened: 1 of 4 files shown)


MarcusDenker:

PR #10832, opened 2022-01-31
Add back beautifulcomments

estebanlm:
this PR adds back BeautifulComments, but as external projects.

EDIT: I sneaked also some refactors to BaselineOfIDE because it was starting to look very ugly :P


Ducasse:
Should I need to merge to integration branch

estebanlm:
yep

Ducasse:
Ok I will.

Ducasse:
Do you have an idea how I should do it because not every packages of Microdown should be loaded in Pharo. 
So may be the baseline is loading the correct subset. 
For now I will push everything in master and integration.

estebanlm:
I am loading just the packages included in "RichText" group of Microdown.
BeautifulComments is just a single package.

Ducasse:
should I change the baseline of BeautifulComments to use something else than dev of Microdown?

estebanlm:
nope. I load Microdown before, with the correct version.

Ducasse:
shit when I load BC after MD I get 233 failures :(

estebanlm:
oops
I am pushing the first version of this.

Ducasse:
So I do not know what to do and I do not have the time nor energy. 
Could we not put BC in Pharo? Our tools are not good to manage projects.

Ducasse:
Yes this is good to only load this package. Probably that BC is doing the same.

estebanlm:
now this is ready :)

MarcusDenker:
aargh... now merging 

Integrate Roassal3 v0.9.9e #10846

before leads to a conflict...

estebanlm:
do not merge yet :)

estebanlm:
now yes

PR #10838, opened 2022-02-01
Increasing minimum size for Calypso browser

carolahp:
New minimum size prevents widgets from hiding.

Note: In Calypso Browser, the middle toolbar (the one between the navigation views and the source code editor) constraints the minimum size of its container (which also contains the widgets). 

To allow the browser to have a smaller size, the middle toolbar must be split into smaller toolbars, which are able to resize together with their container

### DIFF
--- a/src/Calypso-Browser/ClyBrowserMorph.class.st
+++ b/src/Calypso-Browser/ClyBrowserMorph.class.st
@@ -123,9 +123,7 @@ Class {
 		'navigationStarted',
 		'plugins',
 		'navigationEnvironment',
-		'systemScope',
-		'toolbars',
-		'toolbarPanels'
+		'systemScope'
 	],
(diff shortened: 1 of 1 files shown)


src/Calypso-Browser/ClyBrowserMorph.class.st
```diff
@@ -123,7 +123,9 @@ Class {
 		'navigationStarted',
 		'plugins',
 		'navigationEnvironment',
-		'systemScope'
+		'systemScope',
+		'toolbars',
```
demarey: It looks like you have unused i.v.

carolahp: Oh you are right, I commited that by accident!

demarey (approved):

PR #10841, opened 2022-02-01
Inserting variable breakpoint when the selected node is variable and the user double clicks left panel

carolahp:
Insert a VariableBreakpoint that breaks when the variable is accessed.
The problem of inserting breakpoints in pseudo-variables is not solved yet.


MarcusDenker:
The question I have: why would setting a breakpoint on one variable set a breakpoint in all accesses?
 
I would *never* want that! if I set a breakpoint on a variable, I want to break *on that one* access.

MarcusDenker:
The "break on all read/write" should just be there as a menu, and maybe even only on the variable definition.

carolahp:
> The "break on all read/write" should just be there as a menu, and maybe even only on the variable definition.

You are right, I just fixed the PR to:
- Allow the insertion of normal breakpoints to variables
- Remove the insertion of variable breakpoint on all accesses when the user double clicks the left panel

As a result, when a variable node is selected and the user double clicks the left panel, a normal breakpoint is inserted

dionisiydk (approved):

PR #10842, opened 2022-02-01
Fix: preserve pinning between ffi calls.

### LINKED ISSUE #10215: Image build by the CI sometimes is too large (128MB)
**Bug description**
The build on the CI should result in an image slighly smaller than 60MB. But often, it is 128 MB.

**To Reproduce**
Just have a look at http://files.pharo.org/image/100/ 

Compresses the zip is often 38MB (which is the 128MB image), sometimes correct 22MB (that is the 60M image)

** More infos **

I did a space tally analysis (full file attached). This does not show any problems, the image thinks it has just 65MB of allocated instances.
[STspace.txt](https://github.com/pharo-project/pharo/files/7387206/STspace.txt)

guillep:
Some ffi calls require that arguments are pinned to work correctly.
Pin arguments before calls, save the previous state and restore pinning after call.

The main objective of this PR is to avoid having large snapshots with big holes due to pinned objects.

### DIFF
--- a/src/ThreadedFFI-Graphics-Extensions/Bitmap.extension.st
+++ b/src/ThreadedFFI-Graphics-Extensions/Bitmap.extension.st
@@ -3,6 +3,5 @@ Extension { #name : #Bitmap }
 { #category : #'*ThreadedFFI-Graphics-Extensions' }
 Bitmap >> tfPointerAddress [
 
-	self pinInMemory.
 	^ PointerUtils oopForObject: self
 ]
--- a/src/ThreadedFFI-UFFI-Tests/TFUFFIDerivedTypeMarshallingTest.class.st
+++ b/src/ThreadedFFI-UFFI-Tests/TFUFFIDerivedTypeMarshallingTest.class.st
@@ -86,19 +86,6 @@ TFUFFIDerivedTypeMarshallingTest >> testMarshallingByteArrayAsStringArgument [
 ]
 
 { #category : #tests }
-TFUFFIDerivedTypeMarshallingTest >> testMarshallingByteArrayWillPinTheObject [
-	| aByteArray |
-	
-	aByteArray := ByteArray new: 10.
-	
-	self fillByteArray: aByteArray size: aByteArray size.
-	
-	self assert: aByteArray equals: #[1 2 3 4 5 6 7 8 9 10].
-	
-	self assert: aByteArray isPinnedInMemory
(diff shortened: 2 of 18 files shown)


estebanlm:
I a little bit concerned how this change will affect performance. Any idea about it?

guillep:
It will most probably not. Previously it was also doing the pinning. This one only introduces a call to setPinnedInMemory: at the end.

dionisiydk:
Hi.

Does it mean that the pinning is now hidden from the users and it's now done automatically? 

How it would survive the cases where the library keeps the passed objects (pointers) internally? (to performs some code asynchronously for example).  

And imagine the following scenario with concurrent FFI calls:
- argA is unpinned in memory
- process1 wants to call func(argA)
- process1 pins argA
- process1 returns from func(argA)
- process1 is about to unpin the argA but process2 yields it just before the unpin call.
- process2 wants to call func(argA) with same argument as process1.
- process2 pins argA before func call but process 1 yields it (as some image activity triggering processor scheduling).
- now process1 continues the return from func and unpins argA
- process2 continues and passes argA to func in unpinned state (by process1) - BUG
- and it can be even worse with unblocked FFI calls. 

Previously without "unpin recovery" such problems did not exist.

dionisiydk:
> The main objective of this PR is to avoid having large snapshots with big holes due to pinned objects.

Do you mean the image snapshot? Probably the snapshot primitive can do some extra work to handle pinned objects.

guillep:
> > The main objective of this PR is to avoid having large snapshots with big holes due to pinned objects.
> 
> Do you mean the image snapshot? Probably the snapshot primitive can do some extra work to handle pinned objects.

Nope it cannot.

The Snapshot primitive has two main modes: quit and no quit.
If quitting, we can do all the post-treatment we want.
However, if not quitting, we should keep all pinned objects in the same place in memory because an external library (e.g., SLD, Freetype) could have a pointer to that object. So it's important we don't unpin that object, and we don't move it.

If we did not have "save and not quit" we wouldn't have this problem.

The safe place to unpin objects is the startup of the image, but that kind of breaks the purpose, because snapshots will still be large :)

I'll answer to your other message separately.

guillep:
> Previously without "unpin recovery" such problems did not exist.

Well, pinning/unpinning and ffi calls are all primitives, thus they are not suspension points.
So it should never happen.

PR #10844, opened 2022-02-01
Highlighting the selected node when displaying contextual menu

### LINKED ISSUE #10843: When the contextual menu is shown Calypso browser, the selected node should be highlighted in the source code
The selected node should be highlighted when a contextual menu is displayed

carolahp:
Fixes [issue](https://github.com/pharo-project/pharo/issues/10843)

### DIFF
--- a/src/Calypso-Browser/ClyTextEditor.class.st
+++ b/src/Calypso-Browser/ClyTextEditor.class.st
@@ -18,6 +18,8 @@ ClyTextEditor class >> contextMenuOn: aBuilder [
 		ifFalse: [ ^self ].
 		
 	context := aBuilder model browserTool createCommandContext.
+	context selectedTextInterval ifEmpty: [ 
+		context showSourceNode].
 	CmdContextMenuActivation registerContextMenuOn: aBuilder inContext: context
 ]
 


dionisiydk:
It breaks simple copy/paste commands from context menu.
Try for example paste the text from the clipboard into the cursor position.

MarcusDenker:
It feels a bit odd, too: a lot of the highlight is under the menu.

And it highlights other occurrences, too, which is confusing.

carolahp:
> It breaks simple copy/paste commands from context menu. Try for example paste the text from the clipboard into the cursor position.

Yes, my implementation is selecting the text which is wrong, a good solution should only highlight the text instead of selecting it

carolahp:
> It feels a bit odd, too: a lot of the highlight is under the menu.

This is true, I was thinking about complementing this solution automatic highlighting of the currently selected node when the caret moves and when the user clicks on the text
 
> And it highlights other occurrences, too, which is confusing.

Yes, a good solution should highlight only the node referred by the cursor position or caret position (not other occurrences)

estebanlm:
I do not like this, it breaks how context menu works in any editor I have in mind. 
Can I know what is the rationale of it? The issue does not explain why it needs to be like that, and a priory I am against such change. 
Now, if this is really, really needed (and I still need to know the reason, "it was asked by X" is not a reason, I need to know why this is asked), then: 
a) evaluate not using selection buy a simple highlight
b) I'll consider adding a setting (even if not exposed), because I am almost sure 99% of users will hate it, at least at the beginning ;)

carolahp:
The idea behind this issue is to provide users with better feedback before applying changes through the debugging tools in the contextual menu. 
A few weeks ago I proposed the solution of adding, in the debugging menu, the name of the node where the debugging operation will be applied (eg adding a breakpoint), but I feel this is not a good solution and we should aim to find one based on highlighting. 

a) I agree that the solution should be based on highlighting and not on selection (I PR the selection solution not realizing the problems it would bring)
b) This is a radical change, so yes, I think making it optional is a good idea

c) This change is even more radical if we highlight the selected node as the caret moves through the code (eg, in eclipse they highlight the line where the caret is positioned and this is useful because they have line based breakpoints... in our case, we install breakpoints in nodes, therefore our highlighting would be different). This is just an idea I propose and it also requires further discussion

Ducasse:
so I suggest to close this issue. Correct?

Issue #3156, opened 2019-04-10
Use faster string primitives


Ducasse:
VMMaker.oscog-sk.2367
Author: sk
Time: 19 April 2018, 12:02:35.661622 pm
UUID: 0c2401e3-1450-4f73-8e81-958f50171595
Ancestors: VMMaker.oscog- nice.2366

** new primitive to compare strings (slang + JIT)
answers negative smi, 0 or positive smi (instead of 1, 2 or 3 in the MiscPlugin)

* Slang (primitiveCompareWith)
order is optionnal. 
comparison loop performed in rawCompare: string1 length: strLength1 with: string2 length: strLength2 accessBlock: accessBlock

* JIT (genPrimitiveStringCompareWith)
the JIT primitive does not take order as parameter (assumed asciiOrder)
quick jump if one of the strings is empty

Ducasse:
we should not use anymore 

compare: string1 with: string2 collated: order
	"Return 1, 2 or 3, if string1 is <, =, or > string2, with the collating order of characters given by the order array."

	| len1 len2 c1 c2 |
	<primitive: 'primitiveCompareString' module: 'MiscPrimitivePlugin'>
	
	<var: #string1 declareC: 'unsigned char *string1'>
	<var: #string2 declareC: 'unsigned char *string2'>
	<var: #order declareC: 'unsigned char *order'>

	len1 := string1 size.
	len2 := string2 size.
	1 to: (len1 min: len2) do:
		[:i |
		c1 := order at: (string1 basicAt: i) + 1.
		c2 := order at: (string2 basicAt: i) + 1.
		c1 = c2 ifFalse: 
			[c1 < c2 ifTrue: [^ 1] ifFalse: [^ 3]]].
	len1 = len2 ifTrue: [^ 2].
	len1 < len2 ifTrue: [^ 1] ifFalse: [^ 3].

Ducasse:
Hi Stephane,
If you google 'vm-dev VMMaker.oscog-sk.2367'
and click one of the first hits https://marc.info/?l=squeak-vm-dev&m=152413936110744&w=2

you'll see the changes from Sophie:
The new InterpreterPrimitives>>primitiveCompareWith should be linked to primitive 158.
(and only for ByteString comparison).

jecisc:
Is there any news on this change?
It's a great performance improvement. I tried to integrate it some years ago but I remember someone else in the team wanted to do it but it seems that the code in Pharo 9 is still using the old primitive.

[...]

PR #10861, opened 2022-02-06
Add faster ByteString comparison primitive.

jecisc:
/!\ This just adds the methods but it does not use them in the actual default implementation of String comparison. This is just a first step.

See: https://github.com/pharo-project/pharo/issues/3156


jecisc:
I'll open a better PR later

PR #10863, opened 2022-02-07
5879-Printing-an-exception-should-not-rely-on-the-AST

### LINKED ISSUE #5879: Printing an exception should not rely on the AST
These methods are added later in Pharo, they should not be used to just print an exception

MarcusDenker:
Allow block printing if the compiler is not available.

(the error handler already kind of solved it in practice (as it would catch the case of a missing compiler, too), but this improves it: printing a block without the compiler should not be an error).

fixes #5879


MarcusDenker:
just a flaky test:

interestingly failing the first time on all platforms the same time.

Bajger (approved):

PR #10866, opened 2022-02-08
testNoUncategorizedMethods-lowerbound

MarcusDenker:
This PR changes the number we check from 30 to 3.

(it is impossible to ever get this done if we do not do that, as people are constantly adding new cases...)


astares:
Yes - that was the idea behind these checks. We have to ensure the quality level stays or improves over time - but that we do not fall back into bad quality.

PR #10871, opened 2022-02-08
9646-unload-package-Deprecated90-in-Pharo10

MarcusDenker:
- move all methods that are in Deprecated90 package that still have users to Deprecated10

fixes  #9646 (using the workaround of just positioning the real work...)

PR #10873, opened 2022-02-08
cleanup: rewrite users of unicodeToMacRoman to use Zn encoders

### LINKED ISSUE #10872: Deprecate Character>>#unicodeToMacRoman after fixing the usage
**Describe the problem**
 Character>>#unicodeToMacRoman was put in the Deprecated90 package, but not deprecated. It still has users (it is now in Deprecate10).

TODO
- fix users of #unicodeToMacRoman 
- Deprecate it for real with "self deprecated:" so clients can be fixed

guillep:
Fix https://github.com/pharo-project/pharo/issues/10872
Better version of part of https://github.com/pharo-project/pharo/pull/10871


MarcusDenker:

MarcusDenker (approved):

PR #10876, opened 2022-02-10
Deprecate KMCategoryTarget

### LINKED ISSUE #8485: Has KMCategoryTarget been superseded by KMCategoryBinding?
KMCategoryTarget looks like a previous version of KMCategoryBinding which is no longer used (no references to the class). 
I believe KMCategoryTarget was copied from KMCategoryTarget.

MarcusDenker:
fixes #8485


kasperosterbye (approved):
Perhaps put in the class comment or somewhere that one should use `KMCategoryBinding` instead. But OK as is!

Issue #8179, opened 2020-12-26
When FluidClass on Epicea Tests break


Ducasse:
So I will fix them :)

Ducasse:
So far I failed to fix them :(
I published in my repo what I found

Ducasse:
Notes with Pablo in the train

For the fluid class builder on class side

We should return a builder prefilled with information from the instance side. 

For this we should 

Object << #Point

=> 

Object class << #Point class
	slots: {#uniqueInstance}
	traits: {TFlying + }
	package: ....
The steps to get there are:

- grab the instance (Point)
- create a class side builder to prefilled itself from the instance side
- then fill up the slots

MarcusDenker:
After https://github.com/pharo-project/pharo/pull/10882 is merged, we have just two failing tests:

EpRevertTest>>#testClassRemoval
EpRevertTest>>#testTraitRemoval

The mechanism to create a revertion of a removal (class and trait), which would be an adding of that class, is not working.

MarcusDenker:
I tracked down the cause: EpMonitor>>#behaviorRemoved: has workarounds that do not work for Fluid

- for Obsolete classes, the code after "Fix class name (remove AnObsolete prefix)
- for the category, see the code after "Fix category"

PR #10882, opened 2022-02-10
Epicea-Fluid-fixes

MarcusDenker:
This PR is a step for  #8179

- add an empty method #install to Behavior
- make sure to use it in Epicea when creating classes

There seems to be another problem with Reverting deletion, that should be checked after


MarcusDenker:
the class side #intall is not a good idea, as there are already classes implementing it... but checking for ">>" looks really like a hack.

MarcusDenker:
- just FFITypesTest>>testSignedLongLong failing

Ducasse:
introducing fluidInstall is a nice idea.

PR #10885, opened 2022-02-11
just adding two little tests for fraction accessing.

Ducasse:


MarcusDenker:
CI seems to be in an odd state: lots of network related tests time out.

As this PR just adds test (which are green), this ok to merge

MarcusDenker (approved):

Ducasse:
I think that we should have a look at the microdown tests (May be they were not added but many of them are using network).

PR #10891, opened 2022-02-14
10889-equal-CompiledBlocks-in-the-literal-frame-stored-only-once

### LINKED ISSUE #10889: equal CompiledBlocks in the literal frame stored only once
**Bug description**
We store only those objects in the literal frame that are not already #literalEqual: to another object there.

Now this leads to an interesting over-optimization in the case of CompiledBlocks: if there are two blocks that are #=, we store
only one CompiledBlock.

At first, this looks like a nice memory optimization. From a pure execution point of view, this is even true. But it leads to problems for mapping execution to the AST and thus e.g. text highlights in the debugger.

**Proposal** 
implement in CompiledCode

```
literalEqual: other
	^false
```

With a nice comment describing why this is a good idea.

MarcusDenker:
- add literalEqual: for CompiledBlock
- add test testLiteralEqual

fixes #10889

### DIFF
--- a/src/Kernel-Tests/BlockClosureTest.class.st
+++ b/src/Kernel-Tests/BlockClosureTest.class.st
@@ -168,6 +168,26 @@ BlockClosureTest >> testIsClean [
 ]
 
 { #category : #tests }
+BlockClosureTest >> testLiteralEqual [
+	"Check that if we have two clean blocks with the same code, they are not #literalEqual:"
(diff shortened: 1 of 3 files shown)


pavel-krivanek (approved):
it is fine but I would use `literalEquals:`

MarcusDenker:
Yes, I will make a note of literalEquals: instead of literalEqual: for later

PR #10893, opened 2022-02-15
10860-Debugger-hightlights-the-wrong-line-in-test-for-asserts-inside-two-identical-blocks

### LINKED ISSUE #10860: Debugger hightlights the wrong line in test for asserts inside two identical blocks
**Bug description**
Debgger highlights the first line that has the same code as the one where it is actually halted. This is happens only when assertion is tested inside a block

**To Reproduce**
Run the following tests
```st
1) testWrongHeightlightWithCollections

2)	| array |
3)	array := { 0. 0. 0 }.
4)	array do: [ :each | self assert: each equals: 0 ].
5)	array at: 2 put: 1.
6)	array do: [ :each | self assert: each equals: 0 ]
```

``` st
testWrongHeightlightSingleValue

	| value |
	value := 0.
	10 timesRepeat: [self assert: value equals: 0].
	value := 1.
	10 timesRepeat: [self assert: value equals: 0].
```

```

**Expected behavior**
In both tests, the debugger should highlight  line 6, but instead it highlights line 4.
Very weird. It took me a lot of time to figure it out.

**Version information:**
 - OS: MacOs 
 - Version: 12.1
 - Pharo Version 9

**Expected development cost**
As you know nothing happens magically, so do you have an idea of the development effort required? 
In addition, are you willing to help because after all Pharo is yours too?

**Additional context**
Add any other context about the problem here.

MarcusDenker:
- fix #isCreateFullBlock:code:at: to search for blocks in the literals using #==

- add test #testPcInOuter for CompiledBlock
- move tests to CompiledBlockTest

fixes #10860

### DIFF
--- a/src/Kernel-BytecodeEncoders/EncoderForSistaV1.class.st
+++ b/src/Kernel-BytecodeEncoders/EncoderForSistaV1.class.st
@@ -476,7 +476,7 @@ EncoderForSistaV1 class >> isCreateFullBlock: compiledBlock code: code at: pc [
 	^self extensionsAt: pc in: code into:
 		[:extA :extB :nExtBytes| 
 			(code at: pc + nExtBytes) = 16rF9 and: [ 
-				compiledBlock = (code literalAt: (code at: pc + nExtBytes + 1) + (extA bitShift: 8) + 1)] ]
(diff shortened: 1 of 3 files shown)


MarcusDenker:
I think we should merge this... it is now waiting for a review for 7 days

PR #10901, opened 2022-02-15
Update zinc to the latest version

### LINKED ISSUE #10328: Enhance ZnUrl to be able to parse according to rfc3986 & rfc8089
In my understanding of things, the current `ZnURL>>parseFrom: string defaultScheme: defaultScheme` do not appear to parse according the the [2005 rfc3986 standard for URI](https://datatracker.ietf.org/doc/html/rfc3986#appendix-A).

In particular, it is not part of the scheme if it uses // or not. Mostly this does not matter, but it matters for the file uri.

The detailed syntax of the file uri is in the [2017 rfc8089](https://datatracker.ietf.org/doc/html/rfc8089#section-2), and in particular it specifies that the host part can be left out (that is `file:path` is allowed - path must start with a `/`). Leaving out the host part (the part starting with `//` and up til but not including the next `/` means the local file system. 

Thus: `file:/seg1/seg2/seg3`, `file://localhost/seg1/seg2/seg3`, and `file:///seg1/seg2/seg3` are all the same. The first with no host specified means localhost. The second explicitly mentions localhost. The third leaves the host empty (nothing between `//`and the following`/`) which then defaults to localhost.

What `ZnUrl` should be able to do is to accept all three, or in particular include the single `/`format. 

On an aside, for Pharo it might be useful to support a hostname called `workingdirectory` (or `wdir` for those who do not like typing).
@svenvc

### LINKED ISSUE #10760: Failing on CI: ZnHTTPSTest>>testSNIPage: SSL Exception: connect failed [code:-5]
**Bug description**
ZnHTTPSTest>>testSNIPage started failing on the CI on all three platforms with:

SSL Exception: connect failed [code:-5]

*

demarey:
Update zinc to the latest version as of 2022-02-15


MarcusDenker:
This should contain that fix, to be checked: https://github.com/pharo-project/pharo/issues/10328

svenvc:
This is a very big commit ;-)

I went over the changes and they look OK to me (but I might have missed something). Let's see what happens.

In any case, thanks a lot @demarey

demarey:
It is still a work in progress as there are failing tests and validation.
@svenvc I did some PR on Zinc repository that will be far easier to  review.
This PR was issued this way:
1. Load Zinc from its repository into a fresh Pharo image
2. Select changes that can already be introduced

This way, we only keep a few amount of changes, letting time to fix these changes (e.g. test skip for flaky tests).

MarcusDenker:
With the CI in a better shape it looks much better:

- for all platform: SystemDependenciesTest>>testExternalLocalMonticelloDependencies
- just on Windows: ZnServerTest>> testTooManyConcurrentConnections

PR #10943, opened 2022-02-23
temporary revert FFI pinning as this breaks CI

### LINKED ISSUE #10940: Problems sometimes in startup:  #setPinnedInMemory: ... failed
**Bug description**

Since we merged https://github.com/pharo-project/pharo/pull/10842 I get sometimes images that are in a bad state after saving.

I have not yet found a way to reproduce, it works fine in most cases. The object where it fails is not always the same.
This time:

```
PrimitiveFailed: primitive #setPinnedInMemory: in Bitmap failed
```

But it happened with some Freetype related object, too

demarey:


tesonep:
Yes, I think it is the best. There is code using FFI that is not correctly handling the pinning when passing an object outside the image.

Issue #10623, opened 2021-12-08
iceberg throws a DNU #definitions when checking out a different version of Spec 


estebanlm:
**Bug description**

**To Reproduce**
1. Add a location to pharo-spec
2. repair the repository by selecting "discard image changes and load repository version".

For the moment I workarounded the issue by implementing: 
```Smalltalk
MCPackageLoader >> definitions [
	
	^ MCDefinitionIndex definitions: #()
]
``` 
and seems to be working but I cannot be sure this is the intended behavior.

**Expected behavior**
It should not raise a DNU

**Stack trace**

MCPackageLoader(Object)>>doesNotUnderstand: #definitions
IceClassDefinition>>removeFromPatcher:
IceMCSnapshotPatchVisitor>>visitRemoval:
IceRemoval>>accept:
[ 
		anIceNode value accept: self ] in IceMCSnapshotPatchVisitor(IceTreeVisitor)>>visitTreeNode: in Block: [ ...
FullBlockClosure(BlockClosure)>>ensure:
IceMCSnapshotPatchVisitor(IceTreeVisitor)>>withNode:do:
IceMCSnapshotPatchVisitor(IceTreeVisitor)>>visitTreeNode:
IceNode>>accept:
[ :each | each accept: self ] in IceMCSnapshotPatchVisitor(IceTreeVisitor)>>visitChildrenOf: in Block: [ :each | each accept: self ]
Array(SequenceableCollection)>>do:
IceNode(IceAbstractNode)>>childrenDo:
IceMCSnapshotPatchVisitor(IceTreeVisitor)>>visitChildrenOf:
IceMCSnapshotPatchVisitor>>visitNoModification:
IceNoModification>>accept:
[ 
		anIceNode value accept: self ] in IceMCSnapshotPatchVisitor(IceTreeVisitor)>>visitTreeNode: in Block: [ ...
FullBlockClosure(BlockClosure)>>ensure:
IceMCSnapshotPatchVisitor(IceTreeVisitor)>>withNode:do:
IceMCSnapshotPatchVisitor(IceTreeVisitor)>>visitTreeNode:
IceNode>>accept:
IceWorkingCopyPatcherVisitor>>visitPackage:
IcePackageDefinition>>accept:
IceWorkingCopyPatcherVisitor>>visitNoModification:
IceNoModification>>accept:
[ 
[...]

PR #10950, opened 2022-02-25
Make the patcher more robust

guillep:
Make the patcher more robust by not modifying definitions that are not present in a patcher.
Part of https://github.com/pharo-project/pharo/issues/10623.

Integrate before it's iceberg counterpart.


guillep:
Only one flaky test failing from ProfStef

imagen

PR #10951, opened 2022-02-25
FIX issue #10933: Scaled decimals comparison result depends on the number creation method

### LINKED ISSUE #10933: Scaled decimals comparison result depends on the number creation method
**Bug description**
Scaled decimals comparison is highly depending on the way how the number instance was created. 

```smalltalk
17.23s2 = 17.23s2 >>> true.
(17.234 asScaledDecimal: 2) = 17.23s2 >>> false.
```

The comparison fails because the internal representation is very different

Pharo_1oaKgyTi1V

...which allows change of the precision:

`17.234s3->17.234s3`

...but the comparison should be probably smarter

Alisu:
AS title says 'fix issue #10933'.
Added tests for equality and inequality fro scaled decimals.
Fix by comparing integer part then rounded part of the fraction.
Added rounded part calculation.

### DIFF
--- a/src/Kernel-Tests/ScaledDecimalTest.class.st
+++ b/src/Kernel-Tests/ScaledDecimalTest.class.st
@@ -298,6 +298,46 @@ ScaledDecimalTest >> testScaleExtension [
 ]
 
 { #category : #tests }
+ScaledDecimalTest >> testScaledDecimalsEquality [
+
+	
+	self assert: 1.234s2 = 1.23s2. "1.23 = 1.23"
+	
+	self deny: 1.023s3 = 1.23s2. "1.023 != 1.23 "
+	
+	self assert: 1.230s3 = 1.23s2.  " 1.230 = 1.23"
+]
+
+{ #category : #tests }
+ScaledDecimalTest >> testScaledDecimalsInequality [
+
+	
+	self deny: 1.234s2 < 1.23s2. 
+	self assert: 1.023s3 < 1.23s2. 
+	self deny: 1.230s3 < 1.23s2.  
+	self deny: 1.234s2 < 1.111s2.
+	self assert: 1.111s2 < 1.234s2.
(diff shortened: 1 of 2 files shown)


MarcusDenker:
I think we should put this in pharo11 right when we start

pavel-krivanek (approved):

nicolas-cellier-aka-nice:
-1 on this change. This is loosing the property that = is transitive.

Even if they print the same, two decimals with different fraction will behave differently.
They are not equal.

Take:

    f1 := 1/100.
    f2 := (1/4+ 1) / 100.
    d1 := f1 asScaledDecimal: 2.
    d2 := f2 asScaledDecimal: 2.

If you decide that `d1 = d2`, then we expect `d1 = f1 ==> (d2 = f1)` by transitivity.

You also loose the fact that equal numbers have equal hash which I carefully ensured in Squeak in the past.

    (d2 = f2) ==> (d2 hash = f2 hash).

Now try this:

    d3 := 1000.00s2.
    (d1 = d2) ==> ((d1*d3) = (d2*d3)).

This change will increase the level of astonishment IMO.

The only way you can achieve this goal is to round the decimal to the specified scale at creation.

MarcusDenker:
Thanks! I did a revert PR and re-opened the issue tracker entry.

nicolas-cellier-aka-nice:
That said, ScaledDecimal are not ideal, but they deserve more thought.
If we round at creation we get a sort of FixedPoint with decimal scale, but unlimited range (no overflow - just a limitation of addressable memory space). Maybe it's good enough for monetary apps.
Or maybe we want decimal scale with high precision like the BigDecimal found in some languages...
For example, in Java, the scale is increased in case of multiplication so that the operation remain exact.
https://docs.oracle.com/javase/7/docs/api/java/math/BigDecimal.html

PR #10953, opened 2022-02-25
Fix issue 10949 by checking correctly windows platform (cannot update available font list on win64)

rvillemeur:


rvillemeur:
The windows platform was not correctly checked. It was only checking win32 platform, omitting win64. Code now check the more generic windows platform.

guillep:
Very cool! Thanks!

PR #10966, opened 2022-03-02
Allow to alternate colors of a table

demarey:
This functionality is very useful to better distinguish lines on a big table


MarcusDenker (approved):
ok

PR #10967, opened 2022-03-02
Renamed protocol name from `accesing` to `accessing`

jordanmontt:
Fixed typo in the protocol name of the method `ZnStaticFileDecoratorDelegate >> path:`.

Renamed from "accesing" to "accessing"


astares:
Yes - but Zinc is managed externally - where this needs to be fixed by @svenvc and (re-reintegrated)

Thats why I wrote a rule like ReProperMethodProtocolNameForAddingRemovingRule and used that in ProtocolConventionsTest>>#testProperAccessingProtocolIsUsed which at the moment looks like

image

which reflects the current case.

MarcusDenker:
This PR needs to be done on the Zinc repository

svenvc:
Picked up upstream: https://github.com/svenvc/zinc/commit/8042884dea032b9250f18506fcdce3f029cc6e74

PR #10972, opened 2022-03-07
ensure no literal is kept pinned in memory when building an image

estebanlm:
this should prevent the images of extreme huge size :)


estebanlm:
the image builds and test pass, but image is still 124m :( 
I will need to continue digging on this :((

MarcusDenker (approved):
Would it make sende to implement #pinInMemory for all the immediate classes, too? (for symmetry)

PR #10978, opened 2022-03-09
10935-senders-of-in-playground-shows-senders-of-noMethod

### LINKED ISSUE #10935: "senders of" in playground shows senders of #noMethod
**Bug description**
If you have just a selector (e.g. to:do:) in the playground and you do "senders of" , it browses #noMethod  

The method #extractSelectorFromNode: has a check and does the right thing if the extractor sees this as a variable,
but not for the case of a error node.

It seems to work with this additional check:

```
aNode isParseError ifTrue: [ ^ aNode value asSymbol ifEmpty: [ nil ] ].
``

MarcusDenker:
add another check for the DoIt case,  besides variables, add some simple fallback for error nodes.

This makes the common case work that you have a selector in the playground like "to:do:" and do "senders of" for it.

A test would be nice, the existing case for the variable is not tested, either

fixes #10935


MarcusDenker:

PR #10989, opened 2022-03-12
Expand the class comment for Bag

ctSkennerton:
- give some examples of constructing, enumerating, and accessing for Bag objects
- clarify that the `do:` message will return all occurrences of a stored object

### DIFF
--- a/src/Collections-Unordered/Bag.class.st
+++ b/src/Collections-Unordered/Bag.class.st
@@ -2,6 +2,39 @@
 I represent an unordered collection of possibly duplicate elements.
 	
 I store these elements in a dictionary, tallying up occurrences of equal objects. Because I store an occurrence only once, my clients should beware that objects they store will not necessarily be retrieved such that == is true. If the client cares, a subclass of me should be created.
+
+## Creating 
+Bags can be created from other collections or built up progressively using the `Bag>>#add:` message. The example below shows four different ways to create a `Bag` with 3 occurrences of 1.
+```
+b := #(1 1 1) asBag.
+""or""
+b:= Bag newFrom: #(1 1 1).
+""or""
+b := Bag new add: 1; add: 1; add: 1.
+""or"" 
+b add: 1 withOccurrences: 3.
+
+b valuesAndCounts.  ""Dictionary(1->3)""
+```
+
+## Enumerating
+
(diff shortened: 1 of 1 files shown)


Ducasse:
Normally we should not expect that do: should return a value so I would not mention it.

src/Collections-Unordered/Bag.class.st
```diff
@@ -104,7 +137,11 @@ Bag >> cumulativeCounts [
 
 { #category : #enumerating }
 Bag >> do: aBlock [ 
-	"Evaluate aBlock with each of the receiver's elements as the argument."
+	"Evaluate aBlock with each of the receiver's elements as the argument.
+	
+	This will print all of the occurrences of the value. Use doWithOccurrences:
```
Ducasse: Hello I do not get the this will print...
Because do: does not print anything.

src/Collections-Unordered/Bag.class.st
```diff
@@ -2,6 +2,39 @@
...
+b valuesAndCounts.  ""Dictionary(1->3)""
+```
+
+## Enumerating
+
+It is possible to iterate over each individual occurrance of an object using the `Bag>>#do:` message, or pairs of objects and their occurances using `Bag>>#doWithOccurrences:` or the synonymous messages `Bag>>#associationsDo:` and `Bag>>#keysAndValuesDo:`.
+
+```
+b := #(1 1 1 1) asBag.
+b do: [:obj | obj].  ""will be 1. 1. 1. 1.""
```
Ducasse: I would use do: with a side effect

| nbElts |
nbElts := 0.
b := #(1 1 1) asBag.
b do: [:each | nbElts := nbElts + 1].

src/Collections-Unordered/Bag.class.st
ctSkennerton: Yes. This was poor wording on my part. I replaced with "enumerate"

src/Collections-Unordered/Bag.class.st
ctSkennerton: Yes, this wasn't clear. I've updated the example to use a counter as you suggest.

Ducasse:
super!

PR #10998, opened 2022-03-15
workaround to prevent loss of UI when switching to currently fullscre…

johnhwoods:
…en mode on Linux as this is currently unsupported.

### DIFF
--- a/src/Graphics-Display Objects/DisplayScreen.class.st
+++ b/src/Graphics-Display Objects/DisplayScreen.class.st
@@ -179,10 +179,15 @@ DisplayScreen >> fullscreen [
 
 { #category : #'screen managing' }
 DisplayScreen >> fullscreen: aBoolean [
+    "Set fullscreen status as per aBoolean"
+    
+    "Guard clause to prevent setting fullscreen on Unix platforms"
+    (aBoolean and: [Smalltalk os isUnix])
+        ifTrue:[^ self inform: 'Fullscreen not supported under Unix O/S'].
 
-	SystemAnnouncer uniqueInstance announce: (FullscreenAnnouncement new 
-		displayScreen: self;
-		fullscreen: (LastScreenModeSelected := aBoolean))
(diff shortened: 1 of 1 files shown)


tesonep:
Thanks a lot for the PR, it is nice to have new contributors.

Hi, the feature is working there is a problem with SDL (the base library that we are using for graphics) and the linux distribution. We have to ship a version with the VM but this version might work different depending on the installed system. The best way to tackle this is to use the pharo version installed using the system level dependencies (https://pharo.org/download more instructions in https://software.opensuse.org/download.html?project=devel:languages:pharo:latest&package=pharo9-ui) 

Another issue with your solution is that it is disabling a feature, we try to fix them instead of disabling features.


<blockquote><div><strong><a href="https://pharo.org/download">Pharo - Download</a></strong></div></blockquote>

johnhwoods:
With respect, that  isn't the best way to fix it at all.  The best way to fix it is not to to ship product where a click on a main menu item causes it to crash unpredictably.  That's hardly going to encourage people to believe that Pharo is solid platform.  My code change is invisible to those not using Linux.  The workaround is urgent.  We can always raise a bug ("linux fullscreen workaround must be fixed properly") once it is in the code.  But it is fracking ridiculous to ship 'stable' versions that are so unstable on major platforms.  Pharo 8 did fullscreen properly on all platforms. Until Pharo 9 can do it properly this is Sev 1.

johnhwoods:
This is what the crash looks like on my laptop.  It looks very similar on the 3 other Linux systems to which I have access and it obviously affects others, such as the user who raised 10965 in the first place.  This isn't just 'not working' this is tragic levels of fail.  https://youtu.be/AsHMNTuAoa4

<blockquote><div><strong><a href="https://youtu.be/AsHMNTuAoa4">https://www.youtube.com/watch?v=AsHMNTuAoa4&feature=youtu.be</a></strong></div></blockquote>

astares:
I tried today with latest Pharo 11 VM and image on Ubuntu 22.04.1 LTS and it worked.

PR #11004, opened 2022-03-15
Change UserInterrupt detection to use character codes rather than sca…

johnhwoods:
…ncodes

Direct reference to key scancodes is discouraged because it bypasses O/S translation, nixing regional & language settings and also risks defeating accessibility settings possibly other nice things like kb macros.


johnhwoods:
This seems to have failed for reasons that seem unrelated to the code changes made... ?

MarcusDenker:
yes, the failures are not related: 
- some network timeout

tesonep:
@johnhwoods Thanks a lot for the PR. I think it is right.

@Alisu can you check this? I think it was introduced the change because an issue with AZERTY layouts. Can you check if it is still working?

Alisu:
For AZERTY layouts, there is no problem with the coma ',' but it does not work with the dot '.' because  to make a dot you have to use shift (in QWERTY it is shift+',' to make a '.')

Ducasse:
then it should be with shift. we should not use scan code.

PR #11005, opened 2022-03-16
ClassDefinitionPrinter-Comments

MarcusDenker:
- add Class Comments to LegacyClassDefinitionPrinter OldPharoClassDefinitionPrinter

- unify categories a bit
- add #isAbstract
- add subclass responsibility methods
-  a isNil refactoring

### DIFF
--- a/src/Kernel/ClassDefinitionPrinter.class.st
+++ b/src/Kernel/ClassDefinitionPrinter.class.st
@@ -101,7 +101,8 @@ ClassDefinitionPrinter class >> for: aClass [
 
 { #category : #testing }
 ClassDefinitionPrinter class >> isAbstract [
-	^true
+		
+	^ self == ClassDefinitionPrinter
 ]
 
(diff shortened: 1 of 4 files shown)


src/Kernel/ClassDefinitionPrinter.class.st
```diff
@@ -99,6 +99,11 @@ ClassDefinitionPrinter class >> for: aClass [
 		yourself
 ]
 
+{ #category : #testing }
+ClassDefinitionPrinter class >> isAbstract [
```
tesonep, line 103: Does this kind of definition inherited by the classes and should be overriden?

tesonep (changes requested):

src/Kernel/ClassDefinitionPrinter.class.st
MarcusDenker: Indeed, fixed

PR #11014, opened 2022-03-18
use fixed spec version (for release)

estebanlm:
use fixed spec version for release.

https://github.com/pharo-spec/Spec/releases/tag/v1.1.0

# Changes overview

- deprecation of class side defaultSpec, now all layouts can be defined in instance side (but defaultLayout can still be defined in class side, in case of need)
- deprecation of all WithSpec suffixes. Now to open a presenter window you just use #open/#openModal, etc. (this was a lot more work as you may think).
- added #asWindow, #asModalWindow to allow user to modify the window presenter (adding extent, ot title, for example) without needingto override `initializeWindow:` and before the window is opened.
- all layouts are dynamic and can be modified on runtime (API depending the layout). 
- enhancement SpApplication with notion of iconProvider
- added some standard dialogs (inform, request, confirm) to SpApplication
- added a progress bar standard dialog/processor (SpJobListPresenter) 
- enhanced Common-Widgets: SpFilteringListPresenter, etc.
- many improvements to SpCodePresenter, including syntax highlight, better code operations, text search, etc. 
- implemented drag&drop (not all widget, but required).
- enhanced styleSheets usage
- box, grid and scrollable layouts implement SpTAlignable (to better place its contents).
- enhanced API for building grid layouts.
- fix any remaining memory leak (some may remain, as always... but we removed all we found ;)

... and many, many tweaks and bugfixes!

### DIFF
--- a/src/BaselineOfPharo/BaselineOfPharo.class.st
+++ b/src/BaselineOfPharo/BaselineOfPharo.class.st
@@ -18,7 +18,7 @@ BaselineOfPharo class >> declareExternalProjects [
 
 	^ { 
 	PharoExternalProject newName: 'Iceberg' owner: 'pharo-vcs' project: 'iceberg' version: 'dev-2.0' sourceDir: nil.
-	PharoExternalProject newName: 'Spec2' owner: 'pharo-spec' project:'Spec' version: 'Pharo10'.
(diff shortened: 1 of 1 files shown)


Ducasse:
Sounds cool. 
I will start to check the tutorials (when time allows it).

PR #11016, opened 2022-03-18
Backport Fix #11001: Stateful Traits: Adding One Corrupts Existing Instances

### LINKED ISSUE #11001: Stateful Traits: Adding One Corrupts Existing Instances
**Bug description**
Adding a stateful trait leads to corrupt existing instances

**To Reproduce**
1. Create a class and a stateful trait:
```smalltalk
Object subclass: #MyClass
	instanceVariableNames: 'var'
	classVariableNames: ''
	package: ''.
Trait named: #TMyTrait
	instanceVariableNames: 'tvar'
	package: ''.
TMyTrait compile: 'tvar
	^ tvar'
```
1. Inspect an instance: `MyClass new`
1. Add the trait to the class:
```smalltalk
Object subclass: #MyClass
	uses: TMyTrait
	instanceVariableNames: 'var'
	classVariableNames: ''
	package: ''.
```
1. From the inspector open on the instance variable, try to inspect the state of the trait: `self tvar`

**Expected behavior**
You would expect the result to be nil, but the result is a float. I have managed to crash the VM in certain cases, but don't have a reproducible case for that.

**Screenshots**
Screen Shot 2022-03-15 at 10 54 05 AM

**Version information:**
 - OS: macOS 12.2.1 on M1
 - Pharo 9

**Expected development cost**
I don't know; wish I could be more helpful here!

seandenigris:
Correctly calculating when a class should migrate the instances.

### DIFF
--- a/src/Shift-Changes/ShSlotChangeDetector.class.st
+++ b/src/Shift-Changes/ShSlotChangeDetector.class.st
@@ -10,7 +10,7 @@ Class {
 { #category : #initialization }
 ShSlotChangeDetector >> initialize [
 	super initialize.
-	builderAccessor := [ :e | e layoutDefinition allSlots asArray ].
+	builderAccessor := [ :e | e allSlots asArray ].
 	classAccessor := [ :e | e allSlots asArray].
-	comparer := [ :a :b |  self compareSlotCollection: a with: b ]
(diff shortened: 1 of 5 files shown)


seandenigris:
Thanks! 👍

PR #11020, opened 2022-03-21
fix row background color loss when hovering a table with alternated color configured

demarey:


demarey:
failing tests are timeouts. not related to the PR

PR #11026, opened 2022-03-22
11023-CmdCommand-should-implement-isUsed

### LINKED ISSUE #11023: CmdCommand should implement #isUsed
If you browse a subclass of CmdCommand, the system says that the class is not used.

But this hierarchy discovers subclasses reflectively --> they are all used

MarcusDenker:
add #isUsed, fixes #11023


MarcusDenker:
the reflective discovery seems to no all commands, but just some

PR #11029, opened 2022-03-22
Align a few categories in Zinc with the rest of system

### LINKED ISSUE #11027: Align a few categories in Zinc with the rest of system
Protocols follow mostly the "accessing - xxxx" pattern where xxxx is a specific topic for accessing methods. Having that aligned would not only make it more consistent, this notation allows later also for using a tree (instead of a list) as a possible view for method categories to "dive in".

At the moment only Zinc is not following that pattern in the standard image, we have:
 - "accessing request"
 - "accessing-delegated"
 - "accessing-path"
 - "accessing-query"

image

astares:
Fix #11027

### DIFF
--- a/src/Zinc-HTTP/ZnClient.class.st
+++ b/src/Zinc-HTTP/ZnClient.class.st
@@ -1440,7 +1440,7 @@ ZnClient >> setBasicAuthenticationUsername: username password: password [
 
 ]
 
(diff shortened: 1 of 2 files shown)


svenvc:
Merged upstream

MarcusDenker (approved):

Issue #11018, opened 2022-03-19
SmallInteger>>isPrime is inefficient


johnhwoods:
**Bug description**
Smallinteger>>isPrime is inefficient.  The use of the well-known 6k±1 optimization is recomended

**To Reproduce**
Seconds to confirm the primality of the biggest SmallInteger prime using current method: 
`(Time microsecondsToRun: [ 1152921504606846883 isPrime ]) / 1.0e6.`

Seconds ditto using 6k±1 algorithm: I expect this to be approx half an order of magnitude faster.
```Smalltalk
(Time microsecondsToRun: [ | p s i  |
	p := 1152921504606846883.
	s := p sqrtFloor.
	i := 5.
	[ i < s ] whileTrue: [ 
		p \\ i = 0 ifTrue:[^false].
		i := i + 2.
		p \\ i = 0 ifTrue:[^false].
		i := i + 4.]]) / 1.0e6. 
```

**Expected development cost**
0, I will provide the new `isPrime` method and test it for both performance and accuracy against the 1000 largest SmallIntegers and a million random SmallIntegers.

johnhwoods:
It's not really a 'bug' but I don't know how to apply a different label :-D

johnhwoods:
Related: `Integer>>primeFactors` uses the same algorithm and can similarly be improved by using 6k±1

Ducasse:
Improvements are always cool

johnhwoods:
Done the code but just doing the tests before submitting PR.

Ducasse:
Yes we love tests.... our life insurance :)

johnhwoods:
With the 1000 biggest `SmallInteger`s, all `isPrime` results matched the old ones, total time saving ~45%
of 1 million random `SmallInteger` tests
100% match between result of `isPrime` 
87.9%  `isPrime` ran in 0µs to run in both old and new versions
0.035% `isPrime` ran in a few µs (310 in 1µs, 38 in 2µs, 3 in 3µs and 1 in 38µs)
6.0% ran faster in the new version at a 38.3% time saving.
6.0% ran slower, 258.1% time cost, but this is an artefact caused by the fact it only happens when the old test returns  0µs and the new one 1µs.  To confirm, the same numbers were run through the tests back to back and the new method was fractionally faster, but but both performed on average at about 0.5µs

So I'll commit, push and submit the PR.

Ducasse:
Fixed by PR #11018

[...]

PR #11030, opened 2022-03-22
11007 isPrime does not work for large numbers

### LINKED ISSUE #11007: isPrime does not work for large numbers
**Bug description**
`(105850338e131072+1) isPrime` is causing an error `Error: Cannot truncate this number` after a while.

### LINKED ISSUE #11046: Integer>>isPrime isBroken
Some idiot (yes it was me) has broken isPrime, it fails for specific squares  Obvious bug in hindsight.  `121 isPrime = true` obviously brain still fluffed from Covid.  PR incoming.

johnhwoods:
`LargeInteger>>isPrime` overrides `Integer>>isPrime` with `^self isProbablyPrime` - a 25 iteration Miller/Rabin test.

However, because the fractional indices in the algorithm were expressed as floats, this stopped `isPrime` being useful above `Float fmax` (approx 1.8e308).  However, there's no performance reason for such a ceiling (and, indeed, many of the LargeIntegers we want to test for probable primality are cryptographic numbers in excess of 2^1024 and therefore larger than this).   The new method will return true, in a minute or so, on average 2020 PC, for the 23rd Mersenne Prime, 2^11213-1, a 3376 digit number.


src/Math-Operations-Extensions/Integer.extension.st
```diff
@@ -146,29 +146,18 @@ Integer >> isPrime [
 Integer >> isProbablyPrime [
 	"See isProbablyPrimeWithK:andQ:randomIndex: for the algoritm description."
```
astares: can you change typo: algorithm

johnhwoods: Hah, missed that.  In my defence that typo is in the original method :-)

MarcusDenker:
failig test not related:

Ducasse:
Does it fix #11046

Ducasse:
I'm confused what is the difference with PR about isPrime that got integrated

johnhwoods:
Hi, there were three separate issues which I idiotically turned into four.

1) isPrime didn't work for large numbers (where it redirects to isProbablyPrime which used a method which contained floats instead of fractions for its random indices in a Miller/Rabin method.  I changed the floats to fracs and it worked)  This was #11007, opened by Pavel.

2) I noticed when doing it that the SmallInteger isPrime had a 2x inefficiency so I raised #11018, and fixed this.

3) I also noticed primeFactors could be improved, so I raised #11045.  Whilst doing this I realised that

4) Unfortunately when I did (2) I introduced a stupid bug.  So I raised #11046.

Have I earned a ban yet?  :-D

MarcusDenker (approved):

PR #11047, opened 2022-03-23
Fix for specific failure of isPrime for certain squares introduced by…

johnhwoods:
… bug in my earlier code.  Also changed test code to cover these specific cases.


Ducasse:
It means that our tests were not good enough too :).

johnhwoods:
That's a very generous interpretation :-)

PR #11072, opened 2022-03-29
fix DNU in VTermTestRunner

demarey:
When running tests with VTermTestRunner, i.e. in headless mode and no Junit output, the test runner throws a DNU while trying to serialize the stack with Fuel.


Ducasse:
Tx!!!

MarcusDenker (approved):

PR #11079, opened 2022-03-30
Fix iceberg version to 2.0.6 for release

estebanlm:
so we can fix the changes


MarcusDenker:
failign test is network related timeout

MarcusDenker (approved):

PR #11086, opened 2022-03-31
Fix for issue 11070: Interval>>= and Interval>>hash do not match

### LINKED ISSUE #11070: Interval>>= and Interval>>hash do not match
**Bug description**
When `=` is implemented, `hash` should also be implemented such that equal objects have equal hashes.

**To Reproduce**
```Smalltalk
a := 2 to: 10 by: 3.
b := 2 to: 8 by: 3.
self assert: a = b.
self assert: a hash = b hash
```

**Expected behavior**
These two assertions should both succeed, or both fail.
As of now, the former succeeds and the latter fails

**Version information:**
 - OS: N/A
 - Pharo Versions 9, 10

**Expected development cost**
$0.  There are two solutions, both easy to implement.

1) Precondition: Add to interval test so that the current bug is detected
2) Either
2a) reimplement `Interval>>=` to use the value of `stop` rather than `self last`
or 
2b) reimplement `Interval>>hash` to use the value of `self last` rather than `stop`

Proposal is 2b) in line with bugraisers view that `stop` is not a legitimate state variable of `Interval` :-D

johnhwoods:
Intervals that are equal (`#=`) to each other --- e.g. `(2 to: 10 by: 3)` and `(2 to: 9 by: 3)` should have equal `hash`es.  Intervals that share a start, stop and size --- e.g. `(2 to: 10 by: 3)` and `(2 to: 10 by: 4)` --- and are not equal (`#~=`) to each other should not (excepting collisions) have equal `hash`es: indeed the different `step` size means they differ in all but the first element.

### DIFF
--- a/src/Collections-Sequenceable-Tests/IntervalTest.class.st
+++ b/src/Collections-Sequenceable-Tests/IntervalTest.class.st
@@ -446,6 +446,16 @@ IntervalTest >> testEquals4 [
 ]
 
 { #category : #tests }
+IntervalTest >> testEquals5 [
+	"Interval>>= and Interval>>hash do not match #11070"
+	self assert: (2 to: 10 by: 3) equals: (2 to: 9 by: 3). 
+	self assert: (2 to: 10 by: 3) hash equals: (2 to: 9 by: 3) hash. 
+
+	self deny: (2 to: 10 by: 3) equals: (2 to: 10 by: 4). 
+	self deny: (2 to: 10 by: 3) hash equals: (2 to: 10 by: 4) hash	
+]
+
+{ #category : #tests }
 IntervalTest >> testExtent [
 	self assert: (1 to: 10) extent equals: 9.
 	self assert: (1 to: 10 by: 2) extent equals: 9.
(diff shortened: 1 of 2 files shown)


svenvc:
Equal objects should have equal hashes, true.
The reverse is not a requirement (this is called a collision).
Even returning the same hash value for each instance would be OK, be it a very bad one.

Interval is a pretty bad class, it combines too many roles which are in conflict. But like Stef said it is dangerous to make changes since critical code depends on it.

johnhwoods:
> Equal objects should have equal hashes, true. The reverse is not a requirement (this is called a collision).

And a good hash function is 1) fast to compute and 2) *minimises collisions for unequal objects*

> But like Stef said it is dangerous to make changes since critical code depends on it

Any code that depends on `Interval>>hash` being wrong is already buggy 
--- *especially* if it depends on it being wrong in the way this fix fixes.

astares:
Proposed to fix #11070


@johnhwoods You have to mention "Fix #issuenumber" to associate PR with raised issue

MarcusDenker (approved):

PR #11087, opened 2022-04-01
Fix #11065: Ordered dictionaries should not ignore order when compared.

### LINKED ISSUE #11065: OrderedDictionary: #= ignores order
**Bug description**
equality in OrderedDictionary ignores the order:

```
                (OrderedDictionary with: #a -> 1 with: #b -> 2) = (OrderedDictionary with: #b -> 2 with: #a -> 1)
```

answers true

We should think about if that is really good or if this was an oversight.

johnhwoods:
If you want to ignore ordering when comparing OrderedDictionaries, just send them the message asDictionary first, just like you might send asSet to OrderedCollections to compare them in an order indifferent manner.

Note that this is a bit of a 'mare of a class.  Far too many methods and its purpose seems fundamentally unclear.  However these changes are simple and just fix `=`, the relevant test and a comment.

### DIFF
--- a/src/Collections-Abstract/Collection.class.st
+++ b/src/Collections-Abstract/Collection.class.st
@@ -292,6 +292,12 @@ Collection >> asDictionary [
 ]
 
 { #category : #converting }
+Collection >> asIdentityDictionary [
+
+	^ self as: IdentityDictionary
+]
+
+{ #category : #converting }
 Collection >> asIdentitySet [
 	^(IdentitySet new: self size) addAll: self; yourself
 ]
--- a/src/Collections-Abstract/Collection.class.st
+++ b/src/Collections-Abstract/Collection.class.st
@@ -292,12 +292,6 @@ Collection >> asDictionary [
 ]
 
 { #category : #converting }
-Collection >> asIdentityDictionary [
-
-	^ self as: IdentityDictionary
-]
-
-{ #category : #converting }
 Collection >> asIdentitySet [
 	^(IdentitySet new: self size) addAll: self; yourself
 ]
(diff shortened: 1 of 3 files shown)


svenvc:
LGTM

MarcusDenker:
Failing test:

OrderedIdentityDictionaryTest(OrderedDictionaryTest)>>testEquals

johnhwoods:
Ahah, of course, `testEquals` contains a method to ensure you can still compare contents regardless of order:
`od1 asDictionary = od2 asDictionary` --- but this strategy doesn't work for the Identity flavor because - obvs - `#asDictionary` is not guaranteed to work from `IdentityDictionary`  

What's the best solution?  Creating the apparently missing `asIdentityDictionary` method, and using that instead in an overriding testEquals for the Identity flavor?

Ducasse:
I would not add asIdentityDictionary in collection because there plenty of collections for which it does not make much sense. We can change the tests or use the `as: IdentityDictionary` within the test.

johnhwoods:
> there plenty of collections for which it does not make much sense. 

I hear what you're saying but there is already an existing `Collection>>asDictionary` which *also* doesn't make much sense for plenty of collections (specifically any who have an element that dNU `#key`).  Furthermore, there's an `asIdentitySet` in addition to `asSet` so ... either we've already got too many `asX` methods in `Collection` or we can have one more.  I find it hard to accept that we have exactly the right number :-D

Ducasse:
My gut feeling tells me that we have too many of them :) and you right that we do not have the right number.

johnhwoods:
Ok, I will not add another - I will instead modify the test :-) Push incoming imminently

MarcusDenker (approved):

MarcusDenker:
why did you close it?

johnhwoods:
Erm - because i'm an idiot, oops.  What do I need to do now?

MarcusDenker:
I think if you un-delte the branch 11065-OrderedDictionary--ignores-order in your repo it can be re-opened

johnhwoods:
Ah, I seem to have "lost" that repo.  I still have the changes, but I suppose if I create a same-name branched it won't actually be the same branch and will need a new PR?  Sorry.  Bit of a noob.

MarcusDenker:
yes, then I you need to do a new PR

PR #11104, opened 2022-04-07
10780-Spotter-crashes-in-Pharo-9

### LINKED ISSUE #10780: Spotter crashes in Pharo 9
**Bug description**
I'm trying to migrate my projects to Pharo 9 but when I'm using Spotter I get crashes on this method:
Collection>>#spotterItemsFor:

It is calling another method that is not implemented in Pharo 9 but is in Pharo 10. 

It would be nice to backport this in Pharo 9 to avoid crashing the Spotter. 

**To Reproduce**

I open the spotter then click on the right arrow of the World section

**Version information:**
 - OS: Windows
 - Version: 11
 - Pharo Version 9

**Expected development cost**

Should be pretty straightforward since it works in P10

MarcusDenker:
add the missing method to Collection (#collectionSizeThreshold).

I add it to the image, not the Newtools package as a workaround, this is correctly fixed in Pharo10 already.

fixes #10780

### DIFF
--- a/src/Collections-Abstract/Collection.class.st
+++ b/src/Collections-Abstract/Collection.class.st
@@ -433,6 +433,13 @@ Collection >> collect: collectBlock thenSelect: selectBlock [
 	^ (self collect: collectBlock) select: selectBlock
 ]
 
+{ #category : #hotfix }
+Collection >> collectionSizeThreshold [ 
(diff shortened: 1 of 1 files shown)


MarcusDenker:
failing tests not related

Issue #11098, opened 2022-04-05
Failing on CI (mac only): SpJobListPresenterTest>>testJobIsFinishedWhenWaitingMoreThanWorkBlockDuration


MarcusDenker:
**Bug description**
We have SpJobListPresenterTest>>testJobIsFinishedWhenWaitingMoreThanWorkBlockDuration failing consistently, but only 
on the mac.

```
Error
Got 92 instead of 100.
Stacktrace
TestFailure
Got 92 instead of 100.
SpJobListPresenterTest(TestAsserter)>>assert:description:resumable:
SpJobListPresenterTest(TestAsserter)>>assert:description:
SpJobListPresenterTest(TestAsserter)>>assert:equals:
SpJobListPresenterTest>>testJobIsFinishedWhenWaitingMoreThanWorkBlockDuration
SpJobListPresenterTest(TestCase)>>performTest
```

MarcusDenker:
There is a PR: https://github.com/pharo-spec/Spec/pull/1268

MarcusDenker:
PR was merged in Spec

MarcusDenker:
But sadly it is still failing in the current PR runs with Pharo11

PR #11119, opened 2022-04-12
11107-Small-speedup-for-Dictionaryat 

### LINKED ISSUE #11107: Small speedup for Dictionary>>#at:
While looking at clean blocks, I saw that Dictionary>>#at: on every call creates a block for the failure case. This is not clean as it references self and the argument.

```
at: key 
	"Answer the value associated with the key."

	^ self at: key ifAbsent: [self errorKeyNotFound: key]
```

The rest of the code path is carefully written to never create any block.

We could just inline #at: ifAbsent:

```
at: key
	"Answer the value associated with the key. 
	To avoid block creation, we do not re-use #at:ifAbsent:"

	^ (array at: (self findElementOrNil: key))
		  ifNil: [ self errorKeyNotFound: key ]
		  ifNotNil: [ :assoc | assoc value ]
```

and the same for SmallDictionary:

```

	at: key
		"Answer the value associated with the key.
		To avoid block creation, we do not re-use #at:ifAbsent:"

		| index |
		index := self findIndexForKey: key.
		index = 0 ifTrue: [ ^ self errorKeyNotFound: key ].
		^ values at: index
```

Naive Micro-benchmark shows that it is ~20% faster. 

```

[ Smalltalk globals at: #Object ] bench 
before: "'9091170.897 per second'" 
after: "'11411410.436 per second'" 
```

Does it make sense to do this? Of course, the real solution is that the system should inline automatically.

MarcusDenker:
manual inlining in Dictionary>>#at: to avoid having to create the block for every call. Speeds up micro benchmark ~20%

closes #11107

### DIFF
--- a/src/Collections-Unordered/Dictionary.class.st
+++ b/src/Collections-Unordered/Dictionary.class.st
@@ -339,10 +339,13 @@ Dictionary >> associationsSelect: aBlock [
 ]
 
 { #category : #accessing }
-Dictionary >> at: key [ 
-	"Answer the value associated with the key."
+Dictionary >> at: key [
+	"Answer the value associated with the key. 
+	To avoid block creation, we do not re-use #at:ifAbsent:"
 
-	^ self at: key ifAbsent: [self errorKeyNotFound: key]
+	^ (array at: (self findElementOrNil: key))
+		  ifNil: [ self errorKeyNotFound: key ]
+		  ifNotNil: [ :assoc | assoc value ]
 ]
 
 { #category : #'nested dictionaries' }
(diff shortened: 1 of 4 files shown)


MarcusDenker:
failing test not related:

- testJobIsFinishedWhenWaitingMoreThanWorkBlockDuration #11098

MarcusDenker:
I was thinking about this... I had an idea of a general solution. Not yet sure if it will work, but I close this for now while I check my idea (we can easily re-open the PR if the idea turns out to have some flaw...).

PR #11145, opened 2022-04-22
10815 cleanup prefix change from rb to ast for abstract syntax tree related classes

### LINKED ISSUE #10815: Cleanup: Prefix change from "RB" to "AST" for Abstract syntax tree related classes
From the historic point of view Smalltalk started I guess Smalltalk refactorings had their roots in the Refactoring Browser (RB) and can be traced back to work at UIUC (University of Illinois, Urbana-Champaign) by Don Roberts, John Brant, and Ralph Johnson

 - https://wiki.c2.com/?RefactoringBrowser
 - https://refactory.com/refactoring-browser
 - https://www.researchgate.net/publication/220346807_A_Refactoring_Tool_for_Smalltalk

Following this history the classes like RBNode had an "**RB**" prefix for **RefactoringBrowser**.

But logically now in newer Pharo we have put them into packages for "AST" - Abstract Syntax Tree

image

If you think about it for a while it would be more logically to have them prefixed with "AST" instead of "RB". Yes - we know it is heritic and means effort and deprecation/transition/migration - but for Pharo we want to constantly improve, get clean and do for the better.

Additionally this would also help people who are comming from other (more mainstream) languages to find it in alignment with the AST concepts and namings there.

astares:
Fix #10815 10815


MarcusDenker:
```
+ /builds/workspace/est_and_branch_Pipeline_PR-11145/64/bootstrap-downloads/vmtarget/pharo --headless Pharo11.0-PR-monticello_bootstrap-64bit-8d0f06a.image --no-default-preferences st /builds/workspace/est_and_branch_Pipeline_PR-11145/64/bootstrap/scripts/02-monticello-bootstrap/02-bootstrapMonticello.st --save --quit
SystemNotification: Creating Snapshot of: AST-Core
Instance of Association did not understand #needsFullDefinition
Association(Object)>>doesNotUnderstand: #needsFullDefinition
classVariablesNeedFullDefinition
  self error: 'No decompiler available' in RBProgramNode class(Class)>>classVariablesNeedFullDefinition in Block: classVariablesNeedFullDefinition...
anySatisfy: t1
  self error: 'No decompiler available' in Array(Collection)>>anySatisfy: in Block: anySatisfy: t1...
```

I think we need better support for deprecating classes and automatic renaming if we want to do something like this.

The impact is *huge* (all clients everywhere need to be updated)... without a working deprecation mechanism, this will be hard to do.

MarcusDenker:
I close this for now

astares:
@MarcusDenker The approach was to rename and then have a compatible deprecated subclass for the transition. But it already failed when loading the PR back - looks like RB stuff is involved in more things than wanted/required.

Will give it another try later to try to find the cause

PR #11148, opened 2022-04-22
10963-Memory-AllProtocol-uses-a-lot-of-memory

### LINKED ISSUE #10963: [Memory] AllProtocol uses a lot of memory
**Describe the problem**
Every class and metaclass has a AllProtocol. As a subclass of Protocol it has a name ivar (even though the name is fixed) and a methodSelctors IdentitySet, which is never used.

But maybe even AllProtocol is not needed at all, the only users seem to be old code like the Monticello browsers.

Memory:

```
AllProtocol new sizeInMemory. "32"  
IdentitySet new sizeInMemory. "24"
IdentitySet new array sizeInMemory. "48"

AllProtocol allInstances size * (32+24+48) /1024 / 1024.0 "1.9226455688476563"
```

We should first refactor to not need the identitySet and the name, then check if we can remove it altogether.

MarcusDenker:
This PR Refactors the Protocol hierarchy. By not inheriting from Protocol, AllProtocol does not need to allocate space for the two ivars that it never uses (name and methodSelectors).

(there are numberOfClasses*2 instances of AllProtocol)

fixes #10963

### DIFF
--- a/src/Kernel/AbstractProtocol.class.st
+++ b/src/Kernel/AbstractProtocol.class.st
@@ -78,7 +78,7 @@ AbstractProtocol >> name [
 ]
 
 { #category : #accessing }
-AbstractProtocol >> name: arg1 [ 
+AbstractProtocol >> name: aString [ 
 	^ self subclassResponsibility
 ]
 
--- a/src/Kernel/AbstractProtocol.class.st
+++ b/src/Kernel/AbstractProtocol.class.st
@@ -1,0 +1,94 @@
+"
+Abstract superclass for Protocol, AllProtocol should not inherit from Protocol to avoid wasting memory
+"
+Class {
+	#name : #AbstractProtocol,
+	#superclass : #Object,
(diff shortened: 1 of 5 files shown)


MarcusDenker:
failing test not related:

- Failing on CI (mac only): SpJobListPresenterTest>>testJobIsFinishedWhenWaitingMoreThanWorkBlockDuration #11098

src/Kernel/AbstractProtocol.class.st
```diff
@@ -0,0 +1,94 @@
...
+	^ self subclassResponsibility
+]
+
+{ #category : #accessing }
+AbstractProtocol >> name [
+	^ self subclassResponsibility
+]
+
+{ #category : #accessing }
+AbstractProtocol >> name: arg1 [
```
demarey: arg1 should have a better name

demarey (approved):
looks good for me.
There is only one method argument to rename

PR #11156, opened 2022-04-25
11063-ExternalChangesBrowserbuildMenu-should-use-selectedItemsSorted 

### LINKED ISSUE #11063: ExternalChangesBrowser>>buildMenu: should use selectedItemsSorted
**Bug description**
File in all selected menu item uses selectedItems which causes problems filing in as dependencies are not honoured.
It should use selectedItemsSorted.

**To Reproduce**
1. Place a breakpoint in buildMenu: and inspect selectedItems
2. Observe that the order of the items is not the same as on the UI

**Expected behavior**
Code should be filed in in order to honour dependencies.

**Version information:**
 - OS: [MacOS]
 - Version: [12.0.1]
 - Pharo Version [9]

**Expected development cost**
Nil - Change the code to use selectedItemsSorted.

MarcusDenker:
Fix ExternalChangesBrowser as described in the issue tracker entry: use #selectedItemsSorted

fixes 11063

### DIFF
--- a/src/Tool-ExternalBrowser/ExternalChangesBrowser.class.st
+++ b/src/Tool-ExternalBrowser/ExternalChangesBrowser.class.st
@@ -142,7 +142,7 @@ ExternalChangesBrowser >> buildMenu [
 	aMenu 	addItem: [ :item | 
 			item
 				name: 'File in all selected';
-				action: [ self fileIn: changes selectedItems ] ].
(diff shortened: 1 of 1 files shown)


MarcusDenker:
Just
- Failing on CI (mac only): SpJobListPresenterTest>>testJobIsFinishedWhenWaitingMoreThanWorkBlockDuration #11098

StewMacLean:
The change looks good - thanks!

PR #11171, opened 2022-04-28
10957-Memory-sharedPools-OrderedCollections-waste-memory

### LINKED ISSUE #10957: [Memory] sharedPools OrderedCollections waste memory
**Describe the problem**
Class>>#sharedPools instantiates an OrderedCollection. But almost no class actually imports a shared pool:

```
(Smalltalk allClasses select: [ :each | each sharedPools isEmpty  ]) size  "9530"
```

This wastes quite some memory:

```
OrderedCollection new sizeInMemory "32"
(OrderedCollection new instVarNamed: #array) sizeInMemory "88"

(9530 * (32+88)) / 1024 / 1024.0 "1.0906219482421875"

```

==> we could save some memory if we would lazy initialize the sharedPools ivar

MarcusDenker:
This PR moves the creation of the empty OrderedCollection from #sharedPools to #addSharedPool:.

This avoids allocating an OrderedCollection when e.g. creating AST (which happens on comping a method or looking at it in the browser).

This PR will not make the boostrapped image smaller, but it saves memory later (e.g. for newly loaded classes)

fixes #10957

### DIFF
--- a/src/Kernel/Class.class.st
+++ b/src/Kernel/Class.class.st
@@ -992,7 +992,7 @@ Class >> sharedPoolOfVarNamed: aString [
 { #category : #'pool variables' }
 Class >> sharedPools [
 	"Answer an orderedCollection of the shared pools declared in the receiver.
-	Note: To save memory, the variable is inialized by #addSharedPool:"
+	Note: To save memory, the variable is initialized by #addSharedPool:"
 
 	^ sharedPools ifNil: [ super sharedPools ]
 ]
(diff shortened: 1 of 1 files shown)


astares (approved):

MarcusDenker:
(we need to wait for the CI to check the PR, it is shown green but when the CI is down this happens sadly. (it shows "1 successful check", while there should be 2)

Ducasse:
Changes is good to me.

MarcusDenker:
Uhh... the CI id dead... I have no idea if this really works (I ran the kernel tests, but not all the tests)

PR #11173, opened 2022-04-29
Fixes #11165 begins with and ends with should return true on empty string

### LINKED ISSUE #11165: beginsWith: and endsWith: should return true on empty string
A [long discussion in mailing list](https://lists.pharo.org/empathy/thread/6IMEMATJWKJR2VV3LBUKD55PJPTA2PGZ) on correct behaviour of `beginsWith:` and `endsWith:` should be if passed the empty string.

- as it stands, `'something' beginsWith: ''` returns false
- suggested behaviour: `'something' beginsWith: ''` returns true.

In particular, the argument for change is this (due to Richard O'Keefe)
```
We have many ways to define a partial order
on strings.
x <= y iff y beginsWith: x
x <= y iff y endsWith: x
x <= y iff y includesSubCollection: x
x <= y iff y includesSubSequence: x
These things are supposed to obey laws:
if a beginsWith: b , c
then a beginsWith: b
if a endsWith: b , c
then a endsWith: c
if a includesSubCollection: b , c
then a includesSubCollection: b
and a includesSubCollection: c
if a includesSubSequence: b , c
then a includesSubSequence: b
and a includesSubSequence: c.

We also expect the usual rules of equality
to hold.  So
(1) a beginsWith: a
(2) a = '' , a
(3) THEREFORE a beginsWith: ''

(1) a endsWith: a
(2) a = a , ''
(3) THEREFORE a endsWith: ''

(1) a includesSubCollection: a
(2) a = '' , a
(3) THEREFORE a includesSubCollect: ''

Reasoning about strings (as values) gets
enormously more complicated if the operations
do not follow simple sensible rules, and
having '' be the least string under these
orderings and having '' be a prefix and a
suffix of any string is essential if the
rules are going to be simple and coherent.

'' is to strings (and more generally
empty sequences are to sequences) pretty
much what 0 is to integers.   Denying that
'abc' beginsWith: '' is *structurally*
just like denying that 0 <= 123.
```

From a formal (formal languages) point of view, the empty string is the identity element for concatenation.

kasperosterbye:
Letting the CI check.

### DIFF
--- a/src/Collections-Abstract-Tests/TBeginsEndsWith.trait.st
+++ b/src/Collections-Abstract-Tests/TBeginsEndsWith.trait.st
@@ -36,7 +36,7 @@ TBeginsEndsWith >> testsBeginsWith [
 { #category : #'tests - begins ends with' }
 TBeginsEndsWith >> testsBeginsWithEmpty [
 	
-	self deny: (self nonEmpty beginsWith:(self empty)).
+	self assert: (self nonEmpty beginsWith:(self empty)).
 	self deny: (self empty beginsWith:(self nonEmpty))
 ]
 
@@ -51,6 +51,6 @@ TBeginsEndsWith >> testsEndsWith [
 { #category : #'tests - begins ends with' }
 TBeginsEndsWith >> testsEndsWithEmpty [
 	
-	self deny: (self nonEmpty endsWith: self empty).
+	self assert: (self nonEmpty endsWith: self empty).
 	self deny: (self empty endsWith: self nonEmpty)
 ]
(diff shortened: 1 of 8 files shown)


svenvc:
LGTM

MarcusDenker:
mac CI is right now not reachable as the inria building lost internet connection (I went back home). This is why the CI check does not terminate

kasperosterbye:
Great weekend suspense ... :-)

kasperosterbye:
In particular commit https://github.com/pharo-project/pharo/pull/11173/commits/3f4d3ff3a385bf59b304d4613a2b8088883cc2ec is interesting, as those are places in the code which needed the new definition

kasperosterbye:
I assume the PR will restart automatically when servers are back in order?

MarcusDenker:
I forced a restart

MarcusDenker:
Tests look good!

MarcusDenker (approved):

PR #11176, opened 2022-04-29
6720-Dictionary--newFrom-logic-can-be-improved- 

### LINKED ISSUE #6720: Dictionary >> newFrom: logic can be improved.
```
Dictionary >> newFrom: aDict 
	"Answer an instance of me containing the same associations as aDict.
	 Error if any key appears twice."
	| newDictionary |
	newDictionary := self new: aDict size.
	aDict associationsDo:
		[:x |
		(newDictionary includesKey: x key)
			ifTrue: [self error: 'Duplicate key: ', x key printString]
			ifFalse: [newDictionary add: x]].
	^ newDictionary
```

I do not get why we test because how the source dictionary can contain two keys with the same name. 
Since newFrom: expects a dictionary then it is not possible. 
Am'I missing something obvious?

In addition we could change this method to take advantage of #6719

MarcusDenker:
Change #newFrom: to do the same as #newFromAssociations: Instead of error, "If the same key appears twice, the last one enumerated will win"

fixes #6720

### DIFF
--- a/src/Collections-Unordered/Dictionary.class.st
+++ b/src/Collections-Unordered/Dictionary.class.st
@@ -160,17 +160,13 @@ Class {
 { #category : #'instance creation' }
 Dictionary class >> newFrom: aDictionaryOrCollectionOfAssociations [
 	"Answer an instance of me containing the same associations as the argument.
-	 Error if any key appears twice."
+	If the same key appears twice, the last one enumerated will win"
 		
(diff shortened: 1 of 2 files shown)


MarcusDenker:
TraitCompositionTest has one failing test that relies on an Error for this case to be raised.

svenvc:
Makes sense I guess

PR #11185, opened 2022-05-01
Resolves #11184: printStringLimitedTo:Using: will fail to actually respect limit

### LINKED ISSUE #11184: Object >> printStringLimitedTo:Using: will fail to actually respect the desired limit when including '...etc...'. Further, that is included even when not needed.
Rather than respecting the limit, the '...etc...' part will exceed it. Further, will tack on '...etc...' even when exactly at the limit.

Easy way to see the issue:
`12 printStringLimitedTo: 2 >>> '12...etc...'`

This should just return '12'. Also, in general, the '...etc...' part, when it shows up, extends the string past the desired limit.

As far as fixing it... Already wrote the fix, just figuring out all the stuff re github and will hopefully submit it shortly.

Psy-Kosh:
Changes Object >> printStringLimitedTo:Using so that it actually properly respects the limit, and that it doesn't add '...etc...' when exactly at the limit and thus not needed.

### DIFF
--- a/src/Kernel-Tests/ObjectTest.class.st
+++ b/src/Kernel-Tests/ObjectTest.class.st
@@ -207,8 +207,8 @@ ObjectTest >> testDisplayString [
 ObjectTest >> testDisplayStringLimitedString [
 
 	| actual |
-	actual := Object new displayStringLimitedTo: 4.
-	self assert: actual equals: 'an O...etc...'
+	actual := Object new displayStringLimitedTo: 8.
+	self assert: actual equals: 'an O[..]'
 ]
 
 { #category : #tests }
@@ -243,8 +243,8 @@ ObjectTest >> testInstVarNamedPut [
 ObjectTest >> testPrintLimitedString [
 
 	| actual |
-	actual := Object new printStringLimitedTo: 4.
-	self assert: actual equals: 'an O...etc...'
+	actual := Object new printStringLimitedTo: 8.
+	self assert: actual equals: 'an O[..]'
 ]
 
 { #category : #'tests - printing' }
--- a/src/Kernel/Object.class.st
+++ b/src/Kernel/Object.class.st
@@ -1649,11 +1649,12 @@ Object >> printStringLimitedTo: limit [
 { #category : #printing }
 Object >> printStringLimitedTo: limit using: printBlock [
 	"Answer a String whose characters are a description of the receiver
-	produced by given printBlock. It ensures the result will be not bigger than given limit"
+	produced by given printBlock. It ensures the result will be not bigger than given limit.
+	Given that the '...etc...' is itself nine characters long, limit should be at least 10."
 	| limitedString |
-	limitedString := String streamContents: printBlock limitedTo: limit.
-	limitedString size < limit ifTrue: [^ limitedString].
-	^ limitedString , '...etc...'
+	limitedString := String streamContents: printBlock limitedTo: limit + 1.
+	limitedString size <= limit ifTrue: [^ limitedString].
+	^ (limitedString truncateTo: limit - 9), '...etc...'
 ]
 
 { #category : #streaming }
--- a/src/Kernel/Object.class.st
+++ b/src/Kernel/Object.class.st
@@ -1650,11 +1650,11 @@ Object >> printStringLimitedTo: limit [
 Object >> printStringLimitedTo: limit using: printBlock [
 	"Answer a String whose characters are a description of the receiver
 	produced by given printBlock. It ensures the result will be not bigger than given limit.
-	Given that the '...etc...' is itself nine characters long, limit should be at least 10."
+	Given that the '[..]' is itself four characters long, limit should be at least 5."
 	| limitedString |
 	limitedString := String streamContents: printBlock limitedTo: limit + 1.
 	limitedString size <= limit ifTrue: [^ limitedString].
-	^ (limitedString truncateTo: limit - 9), '...etc...'
+	^ (limitedString truncateTo: limit - 4), '[..]'
 ]
 
 { #category : #streaming }


MarcusDenker:
We have right now some CI problems (that is, problems with the internet connection). This way sadly the CI run will be delayed

Psy-Kosh:
>

> We have right now some CI problems (that is, problems with the internet connection). This way sadly the CI run will be delayed

Darn, that sucks. I hope that gets resolved quickly.

Oh, hey, I noticed you change the title of this request. Just so I know for the future, if am doing a pull request that resolves one issue in the tracker, the title for the pull request should exactly correspond to the title of the issue, not just reference the issue number? Thanks.

MarcusDenker:
For the title, I like to see which issue it is... the number does not say much.
To me, it is not needed to match exactly. I just want to be able to find "that PR that fixes printStringLimitedTo" quickly...

Psy-Kosh:
> For the title, I like to see which issue it is... the number does not say much.
> To me, it is not needed to match exactly. I just want to be able to find "that PR that fixes printStringLimitedTo" quickly...

Alright, and thanks.

Psy-Kosh:
So, looks like the commit has some test failures. Is there any easy way for me to find out what the failures were? That is, if it was my code, if it was some other code that was also rolled into that build, old code that was failing previously, or is the fail part of the CI issues themselves? Thanks.

Ah, looks like am able to actually reach the CI server now. Well, sorta. Let's see if I can get it to actually tell me what's failing..

Aaaand it's down.

MarcusDenker:
Yes, there is (again) a problem with the CI.

MarcusDenker:
Finally a CI run: There are two failing tests caused by the PR:

- ObjectTest>>#testPrintLimitedString
- ObjectTest>>#testDisplayStringLimitedString

Psy-Kosh:
Yay for an actual run. And wow that I failed to find the tests in the pharo image in the first place. I even looked in Kernel-Tests. I think I looked for something like "printing" without looking in "Object >> ObjectTests".

Anyways, what's the proper procedure in this case? The test failed because my code changed the behavior of the method, but changing the behavior of the method was the whole point. I mean, now that I can see that there actually were tests there, it'd make sense to change the tests, but I feel that's not something I myself ought be doing without someone else reviewing that. (is there a general "right way" to handle "the whole point of this is I'm changing the behavior of this, so the tests should be changed too" or should I go ahead and change the tests to what I think they should be? For instance,

`12345678910987654321 printStringLimitedTo: 15` should produce `'123456...etc...'` since the total length of the output is 15, but that's a change to the old behavior.

(Anyways, thank you also for helping with my beginner level flailing attempts to actually contribute)

MarcusDenker:
In a case like this we change the tests to reflect the new behavior.

Psy-Kosh:
> In a case like this we change the tests to reflect the new behavior.

Thanks. I mean, obviously that was needed, but wasn't sure if I should do so "silently" without comment, since updating tests when a test fails seems like bad habit to get into without having eyes on the specific reason. Anyways, took care of that, also did one other tweak, replaced the` '...etc...'` with cuis style `'[..]'` because (once a friend pointed that one out to me), sure seemed a whole lot better. (also, yay, finally got iceberg set up so I can push with it instead of dropping to command line to do the push)

And again, thanks for helping me out with all this.

MarcusDenker:
Failing not related, ready for review

Psy-Kosh:
Thanks! Oh, one other question: How do I link the pull request to the issue I opened for it? I mean, the request subject uses the Resolves keyword and references the issue number, so not sure what more I need to do to link the issue and the pull request.

hogoww:
> Thanks! Oh, one other question: How do I link the pull request to the issue I opened for it? I mean, the request subject uses the Resolves keyword and references the issue number, so not sure what more I need to do to link the issue and the pull request.

You just have to mention the number of the issue somewhere in the PR , prefixed by # :)

hogoww (approved):

Psy-Kosh:
> You just have to mention the number of the issue somewhere in the PR , prefixed by # :)

Thanks, but that's what's confusing me since it's mentioned right at the top in the subject of the PR. I should mention here that it resolves #11184? Or it needed to be mentioned in a commit message, or..? And thanks again.

hogoww:
> > You just have to mention the number of the issue somewhere in the PR , prefixed by # :)
> 
> Thanks, but that's what's confusing me since it's mentioned right at the top in the subject of the PR. I should mention here that it resolves #11184? Or it needed to be mentioned in a commit message, or..? And thanks again.

If you click on the hyperlink created by your last message, you will see that the issue is now "has been mentionned in pull request ..."
Guess it doesn't work in titles, or something.
When this PR is merged, the message in the issue will change, and maybe even auto close the issue (I am not sure of that last bit)

MarcusDenker:
The "fixes #11184" just works in the git commit body itself, I think. (which is annoying).

I try to add it manually if  I see that it is missing (which I guess might need special rights...the git issue tracker it a bit hard to use sometimes...)

Psy-Kosh:
> The "fixes #11184" just works in the git commit body itself, I think. (which is annoying).

Wait, the commits, not the pull request? That indeed seems annoying. 

> I try to add it manually if I see that it is missing (which I guess might need special rights...the git issue tracker it a bit hard to use sometimes...)

Thanks, yeah, I'd looked to see if I could, but no option that I could find for me to do so. Sure seems like it needs special permissions to do a manual link. Wait... you manually linking it now has it saying that _I_ linked it. That is, in the issue, it says I linked it a couple days ago, but prior to you linking it, it only said that I mentioned it? (am I being dumb about github, or...?)

Also, just to make sure I understand right, is there anything further I need to do on this pull request at this point, or at this point it's simply waiting until it gets merged? and again, thanks.

MarcusDenker:
> Thanks, yeah, I'd looked to see if I could, but no option that I could find for me to do so. Sure seems like it needs special permissions to do a manual link. Wait... you manually linking it now has it saying that _I_ linked it. That is, in the issue, it says I linked it a couple days ago, but prior to you linking it, it only said that I mentioned it? (am I being dumb about github, or...?)
> 

I linked it. 

The problem is that the issue tracker features like tagging and that is only for those with commit rights, which is really a bad idea... maybe we have to explore "protected branches" more and indeed give more people commit rights (to all but the protected branch).

> Also, just to make sure I understand right, is there anything further I need to do on this pull request at this point, or at this point it's simply waiting until it gets merged? and again, thanks.

Yes, it now needs to be merged. For that we need to decide if it is a good thing to do. For that the best is if people would review it, but sadly not many people do that...

Psy-Kosh:
> Yes, it now needs to be merged. For that we need to decide if it is a good thing to do. For that the best is if people would review it, but sadly not many people do that...

Thanks. Typically how many reviews are looked for for Pharo stuff?

MarcusDenker:
We try for a minimum is 1, and even that is hard to get often.

This PR has one review, so this already is on a good way.

MarcusDenker (approved):

PR #11187, opened 2022-05-04
11186-Clean-Blocks-fix-to-make-debugger-work

### LINKED ISSUE #11186: Clean Blocks: fix to make debugger work
**Bug description**
The debugger does not open in a clean block (e.g. "[1 halt] value" with clean blocks enabled).

**To Reproduce**
1. enable clean blocks
2. recompile the image (OpalCompiler recompileAll)
3.  execute "[1 halt] value" 

**Expected behavior**
Debugger should open

**First Fix**

Change Context>>#home to do one more ifNil: check:

```
home

	"Answer the context in which the receiver was defined, i.e. the context from which an ^-return should return from."

	closureOrNil ifNil: [ ^ self ].
	closureOrNil outerContext ifNil: [ ^ self ].
	^ closureOrNil outerContext home
```

After this, we have just three failing tests:

- DebugSessionContexts2Test>>#testInterruptedContext
- DebugSessionContexts2Test>>#testContextsAfterStepInto
- StepOverTest>>#testStepOverReturnInUnwindBlock

I propose to add this ifNil check to make the debugger work, this way we can test it to see what is wrong much easier.

MarcusDenker:
- nil check in Context>>#home (this allows us to open the debugger from a clean block, we need to check next if it works correctly)

- fix two tests that need  a fix for clean blocks

fixes #11186

### DIFF
--- a/src/Debugger-Model-Tests/DebugSessionContexts2Test.class.st
+++ b/src/Debugger-Model-Tests/DebugSessionContexts2Test.class.st
@@ -36,6 +36,8 @@ DebugSessionContexts2Test >> testContextsAfterStepInto [
 
 	session stepInto.
 	expectedMethod := (self class >> #setUp) literalAt: 2.
+	"for cleanblocks, we have to get the compiledBlock"
+	expectedMethod isBlock ifTrue: [ expectedMethod := expectedMethod compiledBlock  ].
 
 	self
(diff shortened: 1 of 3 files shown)


MarcusDenker:
I would like to merge this, it is now 26 days waiting for a review.

PR #11188, opened 2022-05-04
Rename BlockClosure startpc to compiledBlock

### LINKED ISSUE #11177: BlockClosure ivar startpc should be renamed
**Describe the problem**
BlockClosure has an instance Variable "startpc". This used to be the startpc. But now, it has one read accessor "startpcOrOuterCode" that is just used by Fuel. The accessor is is called compiledBlock/compiledBlock: 

**Proposal**
- Deprecate startpcOrOuterCode and fix Fuel
- rename ivar to "compiledBlock"

The two steps are independent. But note: the rename can not be done in a running system, but we have to modify the source
and commit to git outside of Pharo.

MarcusDenker:
fixes #11177

### DIFF
--- a/src/Kernel/BlockClosure.class.st
+++ b/src/Kernel/BlockClosure.class.st
@@ -39,7 +39,7 @@ Class {
 	#type : #variable,
 	#instVars : [
 		'outerContext',
-		'startpc',
+		'compiledBlock',
 		'numArgs'
 	],
 	#category : #'Kernel-Methods'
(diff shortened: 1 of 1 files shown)


MarcusDenker:
This is a minimal change to just rename the variable, we have to check e.g. the class comment later

MarcusDenker:
Test failing: SpJobListPresenterTest>>testJobIsFinishedWhenWaitingMoreThanWorkBlockDuration
(we really need to disable that test.. it uses delays and that just does not work on the CI)

pavel-krivanek (approved):

PR #11197, opened 2022-05-05
Making completion support $_ in class and method names

### LINKED ISSUE #10986: Underscore in class name breaks text completion
**To Reproduce**
1. Create a class named `My_Class`
```
Object subclass: #My_Class
	instanceVariableNames: ''
	classVariableNames: ''
	package: ''
```
2. In a playground, type `My_`
3. First issue: no text completion entries are showing up
phIssue_1

4. Type fastly `Class` so that playground content is `My_Class`
5. Second issue: text completion is showing up for classes starting with `Class` instead of `My_Class`
phIssue_2

6. Go back to `My`, then type slowly `_Clas` so that playground content is `My_Clas`
7. When `My_Class` text completion shows up, press enter
8. Third issue: now the playground contents is `My_My_Class` instead of `My_Class`
phIssue_3

**Version information:**
 - OS: Win10 and Linux Debian 11
Pharo 10.0.0
Build information: Pharo-10.0.0+build.473.sha.a073a65d1153db338c3be8764da632e2ce71d901 (64 Bit)

carolahp:
Fix #10986 by preventing completion to break when classes and methods contain underscore

### DIFF
--- a/src/NECompletion/Character.extension.st
+++ b/src/NECompletion/Character.extension.st
@@ -3,5 +3,5 @@ Extension { #name : #Character }
 { #category : #'*NECompletion' }
 Character >> isCompletionCharacter [
 	"I defined the logic that completion can only happen with alphanumeric : characters."
-	^ self isAlphaNumeric or: [ self = $: ]
+	^ self isAlphaNumeric or: [ { $: . $_ } includes: self  ]
 ]
--- a/src/NECompletion/Character.extension.st
+++ b/src/NECompletion/Character.extension.st
@@ -3,5 +3,5 @@ Extension { #name : #Character }
 { #category : #'*NECompletion' }
 Character >> isCompletionCharacter [
 	"I defined the logic that completion can only happen with alphanumeric : characters."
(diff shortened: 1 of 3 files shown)


src/NECompletion/Character.extension.st
```diff
@@ -3,5 +3,5 @@ Extension { #name : #Character }
 { #category : #'*NECompletion' }
 Character >> isCompletionCharacter [
 	"I defined the logic that completion can only happen with alphanumeric : characters."
-	^ self isAlphaNumeric or: [ self = $: ]
+	^ self isAlphaNumeric or: [ { $: . $_ } includes: self  ]
```
gcotelli:
```suggestion
	^ self isAlphaNumeric or: [ #(:  _) includes: self  ]
```

MarcusDenker (approved):

hogoww:
@gcotelli Just so you know, you introduced a bug on the code of Carolina which was correct.
I added a test case so we can catch it in the future in the new PR #11487

Issue #11195, opened 2022-05-05
[Meta] Steps needed for CleanBlocks


MarcusDenker:
This issue collects all TODOs to be done to be able to enable CleanBlocks by default

- [x] Clean Blocks: self is nil on Debugger #13591 
- [x] fix self in debugger inspector https://github.com/pharo-spec/NewTools/pull/542
- [ ] Test failing with clean blocks enabled: Debugger #13755 
- [x] Test failing clean blocks enabled: Misc #13780 
- [ ] Test failing with clean blocks enabled: Fuel #13755

DONE:
- [x] With CleanBlocks enabled: tests fail when trying to Fuel out stack #13657
- [x] Announcements... Failing tests: AnnouncerTest, WeakAnnouncerTest https://github.com/pharo-project/pharo/issues/11220
- [x] [CleanBlocks] HDTestReport>>#serializeError:of: fails if we can not find the #signalerContext #11908
- [x] Implement Context>>#activeHome without using #outerContext #11965
- [x] [Clean Blocks] Job>>#owner assumes valid outerContext #11968
- [x] Implement BlockClosure>>#homeMethod (without using #home) #11966
- [x] Implement #findMethodContextSuchThat: #11967
- [x] ConstantBlockClosure: raise error if called with wrong number of args #11901
- [x] Debugger fixes: https://github.com/pharo-project/pharo/issues/11186
- [x] :BlockClosureTest>>testTallyInstructions, MessageTallyTest>>testTallySends #11209
- [x] ReflectivityControlTest>>testConditionDisableEnableNoArguments https://github.com/pharo-project/pharo/issues/11218
- [x] Use Fuel 5 https://github.com/pharo-project/pharo/issues/10318
- [x] Fuel needs to be able to serialize CleanBlocks https://github.com/pharo-project/pharo/pull/11894

MarcusDenker:
as there are only two issues with failing tests, we can close this meta-issue

seandenigris:
Wow, this is really exciting, but we have been working on this so long that I have to go back and dig up the experiments I did to remind me why I'm excited about it ha ha

PR #11209, opened 2022-05-09
Fix-messagetally-tests-Cleanblocks

MarcusDenker:
Two tests fixed, one step for #11195


MarcusDenker:
as this just changes tests, I merge it

PR #11225, opened 2022-05-12
CleanBlock-Hierachy-Cleanup 

### LINKED ISSUE #11217: Check BlockClosure comments for mention of startpc
**Describe the problem**
In https://github.com/pharo-project/pharo/pull/11188 we renamed the ivar "startpc" to "compiledBlock". 

As that change was done by modifying the code with an editor (as it is not a change you can do from within the running system), the change was as limited as possible. It especially did not update the class comment.

**What has to be done**

- Check and improve the class comment of BlockClosure. 
- check referenced to compiledBlock and senders of #startpc to make sure everything is ok.

MarcusDenker:
- make CleanBlockClosure a subclass of BlockClosure
- BlockClosure is now tagged abstract
- override #receiver  in CleanBlockClosure
- push down #sender
- Improve comments ( fixes #11217 )

This saves one pointer per clean block, with ~10K clean blocks in the current image (if turned on), this is nice.

### DIFF
--- a/src/Kernel/BlockClosure.class.st
+++ b/src/Kernel/BlockClosure.class.st
@@ -15,14 +15,13 @@ Implementation:
 
 Instance variables:
 	outerContext <Context|nil> context that defined me
-	startpc: <SmallInteger> (pc = program counter) offset of my first bytecode instruction in the compiledMethod bytecode  
+	compiledBlock <CompiledBlock>  
 	numArgs: <SmallInteger> my number of arguments
 
-I am created at runtime through a special bytecode:
-closureNumCopied: x numArgs: y bytes z1 to z2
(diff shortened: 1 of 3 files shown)


MarcusDenker:
(before merging, we need to check if this works with turning on CleanBlocks)

MarcusDenker:
Nope... needs VM support. For now, we have to keep the nil receiver...

```
 0x16d0c0038 M [] in INVALID RECEIVER>exampleAccessOuterFromCleanBlock 0x10310dab8 is in generated runtime
 0x16d0c0070 M MethodMapExamples>exampleAccessOuterFromCleanBlock 0x10a42e1f8: a(n) MethodMapExamples
```

MarcusDenker:
This looks now good: by adding the var in both subclasses, the layout is the same.
(we can push the var up later, but that can  not be done from within Pharo)

PR #11235, opened 2022-05-16
Reset SelectorTable on #cleanUp

MarcusDenker:
This PR makes sure to nil out the SelectorTable on #cleanUp and #rehash.

This means that the download image will be a bit smaller (WeakSet of size 140K entries).

### DIFF
--- a/src/Collections-Strings/Symbol.class.st
+++ b/src/Collections-Strings/Symbol.class.st
@@ -149,6 +149,8 @@ Symbol class >> internCharacter: aCharacter [
 Symbol class >> internSelector: aStringOrSymbol [
 	| selector |
 	aStringOrSymbol isDoIt ifTrue: [ ^self ].
+	(aStringOrSymbol == #UndefinedMethod) ifTrue: [ ^self ].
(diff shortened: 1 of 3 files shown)


MarcusDenker:
Hmm... not sure if it works:

```
./pharo  pharo.image eval '(Symbol classVarNamed: #SelectorTable) size'
```

should be nil, but it is not. So *something* either initializes the table or the cleanup does not work.

MarcusDenker:
Next step: add Transcript logging to see what is happening during the build

MarcusDenker:
So it seems the release clean is the problem:

+ /builds/workspace/est_and_branch_Pipeline_PR-11235/64/bootstrap-downloads/vmtarget/pharo --headless Pharo11.0-PR-64bit-e22ad15.image --no-default-preferences clean --release
#resetSelectorTable
#rebuildSelectorTable

pavel-krivanek (approved):

MarcusDenker:
So, it works!

- saved ~1MB for the saved "cleaned" image
- avoids interning selectors on compile, just do it on install (which means it will not be done for DoIts and FFI methods).
- simplify #internSelector: a little (no temp needed)

Issue #11226, opened 2022-05-12
Commits with ... (introduced by GH) in summary cannot be loaded back in Pharo


Ducasse:
When the first line of a commit is too long, GH cuts it and place a ... instead. 
This breaks the loading of the code 

I do not know if this changes is correct.
Changed 
```smalltalk
print: (((aVersionInfo perform: sel) ifNil: ['']) asString encodeWith: #'latin-1' ) asString; space ].
```
into
```smalltalk
print: (((aVersionInfo perform: sel) ifNil: ['']) asString encodeWith: #utf8 ) asString; space ].
```

```smalltalk
writeVersionInfo: aVersionInfo
	(self isWritten: aVersionInfo)
		ifTrue: [^ stream nextPutAll: '(id '; print: aVersionInfo id asString; nextPut: $) ].
	stream nextPut: $(.
	#(name message id date time author) do: [:sel | 
		stream 
			nextPutAll: sel; space;
			print: (((aVersionInfo perform: sel) ifNil: ['']) asString encodeWith: #utf8 ) asString; space ].
	stream nextPutAll: 'ancestors ('.
	aVersionInfo ancestors do: [:ea | self writeVersionInfo: ea].
	stream nextPutAll: ') stepChildren ('.
	aVersionInfo stepChildren do: [:ea | self writeVersionInfo: ea].
	stream nextPutAll: '))'.
	
	self wrote: aVersionInfo
```

MarcusDenker:
There is  an issue tracker entry

[Iceberg] commit message is not encoded as UTF8 #10781

I think this is related.

MarcusDenker:
I just saw that with another repository. With the fix I could load it

MarcusDenker:
PRs merged for Pharo10 and Pharo11

PR #11239, opened 2022-05-17
Pharo11-UFT8-Monticello

MarcusDenker:
This commit changes the encoding as described in the Issue  #11226

The MC reader and installer already assume utf8 and fall back to latin1 on error... we could do the same if needed
(see MCMczReader>>#contentStreamForMember: )


Inao0 (approved):
👍

dionisiydk (approved):

Issue #10999, opened 2022-03-15
Window damage not repaired until window is active under Linux


johnhwoods:
**Bug description**
Unlike Pharo 8 and earlier, Pharo 9 and later windows do not repair drag, overlap or collapse damage until the window is clicked.  This behavior is Linux only 

**To Reproduce**
1. On a Linux system
2. Open a Pharo 9 or later image
3. Drag another application window over the top of the image window
4. Image window contents are destroyed and not refreshed until window is clicked.

**Expected behavior**
Application windows are expected to redraw themselves

**Screenshots**
![image](https://user-images.githubusercontent.com/4273151/158376539-9df64d23-2992-45c2-8f19-17762222e3f0.png)

**Version information:**
 - OS: Ubuntu 21 / Windows 10
 - Pharo Version 8,9,10

Was there a major change in window handling or something after Pharo 8?  Possibly related to #10965?

PR #11240, opened 2022-05-18
Redraw world on SDL_WINDOWEVENT_EXPOSED

BlazingMagpie:
Related to #10999
Since Pharo 9 on Linux, OS window wouldn't redraw areas that were covered by anything outside it until other draw event inside it would be called. For example, dragging one window over Pharo window would leave traces of that window; they would stay until one would click on Pharo window to open a context menu or mouse over any morph with different on-hover appearance.
SDL sends SDL_WINDOWEVENT_EXPOSED event in such cases. I have seen no apparent way to find dirty rectangle that needs redrawing and simply redrawing entire window seems to be common approach elsewhere, so the solution is to simply redraw entire window when we receive the event. Everything but event handler were already made.
This patch has been tested by few Linux users and it doesn't seem to cause other issues so far.

### DIFF
--- a/src/OSWindow-Core/OSWindowMorphicEventHandler.class.st
+++ b/src/OSWindow-Core/OSWindowMorphicEventHandler.class.st
@@ -317,6 +317,12 @@ OSWindowMorphicEventHandler >> visitWindowDropEvent: anEvent [
 ]
 
 { #category : #visiting }
+OSWindowMorphicEventHandler >> visitWindowExposeEvent: anEvent [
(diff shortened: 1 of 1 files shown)


gcotelli (approved):
I've tried in Pharo 10 running on Manjaro Linux and now the windows redraw correctly when switching between apps. I didn't notice any odd behavior. :love_you_gesture:

guillep:
Thanks all!!

PR #11241, opened 2022-05-18
Cleanup: OneCharacterSymbols: Symbol allSymbols contains one-character symbols twice 

### LINKED ISSUE #11242: Symbol allSymbols contains one-character symbols twice
**Bug description**

This is fun:

```
Symbol allSymbols asBag occurrencesOf: 'a'
```

MarcusDenker:
This PR removes the initialize and use of OneCharacterSymbols.

- OneCharacterSymbols is not used by #intern:, just allSymbols
- Why add a symbol for all characters anyway? We will intern them as needed.
- The current scheme we could them twice... as they are interned via #intern: and inside the OneCharacterSymbols.

The class variable is not yet removed.

This removed the unused method #selectorsContaining:, too

### DIFF
--- a/src/Collections-Strings/String.class.st
+++ b/src/Collections-Strings/String.class.st
@@ -740,8 +740,7 @@ String >> asString [
 
 { #category : #converting }
 String >> asSymbol [
-	"Answer the unique Symbol whose characters are the characters of the 
-	string."
+	"Answer the unique Symbol whose characters are the characters of the string."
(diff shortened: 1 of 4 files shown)


MarcusDenker:
Only testNoUnusedClassVariablesLeft failing, can not be fixed in this PR (as this can not be done from the image)

pavel-krivanek (approved):
Looks fine

PR #11244, opened 2022-05-20
11243-Symbol-classrebuildSelectorTable-hould-not-use-allInstancesDo

### LINKED ISSUE #11243: Symbol class>>#rebuildSelectorTable hould not use allInstancesDo:
**Describe the problem**
We are using allInstancesDo: in #rebuildSelectorTable:

```
rebuildSelectorTable
	| selectorTable |
	selectorTable := WeakSet new.
	CompiledMethod
		allInstancesDo: [ :method | 
			| selector |
			selector := method selector.
			selector ifNotNil: [ selectorTable add: selector ] ].
	^selectorTable.
```

We should try to avoid #allInstancesDo: as it can be slow when the image gets very large.

MarcusDenker:
do not use allInstancesDo: and simplify, fixes #11243

### DIFF
--- a/src/Collections-Strings/Symbol.class.st
+++ b/src/Collections-Strings/Symbol.class.st
@@ -210,14 +210,10 @@ Symbol class >> readFrom: strm [
 
 { #category : #cleanup }
 Symbol class >> rebuildSelectorTable [
-	| selectorTable |
-	selectorTable := WeakSet new.
-	CompiledMethod
-		allInstancesDo: [ :method | 
-			| selector |
-			selector := method selector.
-			selector ifNotNil: [ selectorTable add: selector ] ].
-	^selectorTable.
+
+	^ SystemNavigation new allMethods
+		  collect: [ :method | method selector ]
+		  as: WeakSet
 ]
 
 { #category : #private }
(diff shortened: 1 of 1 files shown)


MarcusDenker:
Changed to use Weak IdentitySet, see  #11248

src/Collections-Strings/Symbol.class.st
```diff
@@ -210,14 +210,10 @@ Symbol class >> readFrom: strm [
...
-	| selectorTable |
-	selectorTable := WeakSet new.
-	CompiledMethod
-		allInstancesDo: [ :method | 
-			| selector |
-			selector := method selector.
-			selector ifNotNil: [ selectorTable add: selector ] ].
-	^selectorTable.
+
+	^ SystemNavigation new allMethods
```
dionisiydk, line 214: #allMethods is quite expensive itself. I wonder is it really faster than #allInstancesDo: in that case?

MarcusDenker: The idea was to avoid allInstancesDo: as it gets slower with the number of objects (not good for large images with lots of data, but not much code).

We are just interested in those symbols that are used as a method name of *installed* methods, not all methods (even though there should be no difference when it is called, I think)

PR #11254, opened 2022-05-20
Extracting the logic of screenshot and selection

### LINKED ISSUE #5918: Refactor Form>>fromUser

### LINKED ISSUE #5919: Refactor `Rectangle>>fromUser`
`Rectangle >> fromUser` freezes the display to let the user select a rectangle from the screen (drawing the selection in between). Then it returns a rectangle instance with the size selected.

This functionality does not belong to the Rectangle class, It generates a rectangle, but that does not mean it should be in the rectangle class. It should be extracted.

bloomabir:
Extracting the logic of screenshot and selection of areas to an object of class Screenshot.

Fix #5919
Fix #5918


src/Graphics-Canvas/BitmapFillStyle.class.st
```diff
@@ -34,7 +34,7 @@ BitmapFillStyle class >> fromForm: aForm [
 { #category : #'instance creation' }
 BitmapFillStyle class >> fromUser [
 	| fill |
-	fill := self form: Form fromUser.
+	fill := self form: Screenshot new formScreenshotFromUserSelection.
```
astares, line 37: #formScreenshotFromUserSelection is not a good selector
- "screenshot" in the selector is redundant as the receiver is already Screenshot (instance)
- having the type of the result in the selectors name is also not good as other types could be returned as long as they work polymorphic

I would do it with a more convinient method with a more simple intention revealing selector:

```Smalltalk
Screenshot fromUserSelection
```
which returns the form

src/Morphic-Core/Screenshot.class.st
```diff
@@ -0,0 +1,147 @@
...
+	originRect := self 
+		waitUserSelectionOriginRectangle: originRect
+		transformBlock: [ :f :pt | (pt grid: gridPoint) extent: 0 @ 0 ]
+		inWorld: aWorld.
+					
+	aWorld activeHand showTemporaryCursor: Cursor bottomRight.
+
+	^ 	self 
+			waitUserSelectionOriginRectangle: originRect
+			transformBlock: [ :f :pt | f origin corner: (pt grid: gridPoint) ]
```
pavel-krivanek, line 65: Why `f` when it is a rectengle? Consider more descriptive names like `aPoint`

The same few lines below.

pavel-krivanek (commented):
It would be nice to move it to `Morphic-Base` but I understand it may not be that easy considering the original code. `Morphic-Core` is supposed to contain only things required to display a world with simple morph and allow the most basic user interaction with it.

MarcusDenker:
One test is failing on all plattforms:

SystemDependenciesTest>>testExternalMorphicCoreDependencies


```

Error
Given Collections do not match!  additions : #(#'Graphics-Files')  missing: #()
Stacktrace
TestFailure
Given Collections do not match!
	additions : #(#'Graphics-Files')
	missing: #()
SystemDependenciesTest>>testExternalMorphicCoreDependencies
SystemDependenciesTest(TestCase)>>performTest
```

guillep (approved):
I think this is good enough.

We can rename the selectors later @astares.
@bloomabir is not working in the team anymore.

Issue #11246, opened 2022-05-20
Fluid and Trait tests create class that are not clean up


Alisu:
**Describe the problem**
Fluid tests create a Point3 class for test that is not clean up.
Same for Trait tests that create Trait 1 2 3 4 5 .

**Proposal**
Create the classes on setup and remove them on tearDown.

**Additional context**
Half a sprint.

MarcusDenker:
Point3 has been fixed, the Trait 1-4 are there already in a fresh image, Tait5 does not exist.

So this seems to be fixed

jecisc:
I've fixed a lot of things in Fluid cleaning recently so it's highly possible that all of those are now fixed

PR #11259, opened 2022-05-20
One Step for: Tests create class that are not clean up #11246

Alisu:
This is a startup to issue #11246.
In this PR we generate class for ClassTest and we do the same for TraitTest with the introduction of a Trait to add helper where they are needed. (Not ideal solution but anyway...)
Purely remove Point3 class.


MarcusDenker:
Lot of tests failing with:

```
Message not understood: ClassTest >> #generateExampleClass1

```

guillep:
59 tests are failing, and this is too old.
I propose this PR is reopened if somebody takes care of it.

PR #11262, opened 2022-05-22
9828-Inspector-columns-cannot-be-resized 

### LINKED ISSUE #9828: Inspector columns cannot be resized
**Bug description**
Try inspecting any object with instance variables. You get 2 columns named "Variable" and "Value". For long instvar names, you might want to drag the column border to the right - which is possible, but the final position is always reset to the initial default after releasing the drag.

3HQIL8VcLd

lopezca:
`FTColumnResizerMorph>>updateFromEvent` computes the updated column widths from the resizer event data; at the end of the method, it updates the wrong variables and nothing happens. This fix changes the receiver at the end as we need to update the `column model width`, not the `column width`.


jgfoster (approved):
Looks like a good catch!

MarcusDenker (approved):

estebanlm:
this will not work, it needs to be reverted
reason: there is no model in non-spec tables.

estebanlm:
(I already have a fix for this, I was trying to commit when I saw this change)

PR #11291, opened 2022-05-24
Simpler (and slightly faster) Symbol>>#intern: 

MarcusDenker:
This PR simplifies Symbol class >> #intern:

Dispatch the creation of the symbol to Symbol (do nothing), BytString (simple case) and WideString (needs to check #isOctedString in addition).

This means 
- we do one message send instead of sending #isSymbol and #isOctetString,
- we do not check #isOctetString for ByteString (the most common case)

We should benchmark this... it might be not faster

### DIFF
--- a/src/Collections-Strings/ByteString.class.st
+++ b/src/Collections-Strings/ByteString.class.st
@@ -184,6 +184,12 @@ ByteString >> byteSize [
 	^self size
 ]
 
+{ #category : #converting }
+ByteString >> createSymbol [
+
+	^(ByteSymbol basicNew: self size) string: self
+]
+
 { #category : #comparing }
 ByteString >> findSubstring: key in: body startingAt: start matchTable: matchTable [
 	
--- a/src/Collections-Strings/Symbol.class.st
+++ b/src/Collections-Strings/Symbol.class.st
@@ -118,19 +118,10 @@ Symbol class >> initialize [
 ]
 
 { #category : #'instance creation' }
-Symbol class >> intern: aStringOrSymbol [ 
+Symbol class >> intern: aStringOrSymbol [
 
-	^(self lookup: aStringOrSymbol) ifNil:[
-		| aClass aSymbol |
-		aStringOrSymbol isSymbol ifTrue:[
-			aSymbol := aStringOrSymbol.
-		] ifFalse:[
-			aClass := aStringOrSymbol isOctetString ifTrue:[ByteSymbol] ifFalse:[WideSymbol].
-			aSymbol := aClass basicNew: aStringOrSymbol size.
-			aSymbol string: aStringOrSymbol.
-		].
-		NewSymbols add: aSymbol.
-		aSymbol].
+	^ (self lookup: aStringOrSymbol) ifNil: [ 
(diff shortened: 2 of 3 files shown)


MarcusDenker:
- Needs some form of benchmark 
- do we really need to do the #isOctetString check? If we do not do it, the only thing that change is that a symbol of the kind 'hello' asWideString asSymbol will be created as a WideSymbol, and thus 'hello' asSymbol will use a WideSymbol instance, too. But there will never be two symbols for the string.

MarcusDenker:
Small naive benchmark. If we remove the adding of the resulting symbol to the table, we can use #bench to check the speed of the changed code. I made for both the current and the new intern code a version "intern2:" with just the "add" to NewSymbol commented out.

```
[Symbol intern2: 'akldjfalskdfhalksdfjaslkhdfj'] bench  


"'1809292.824 per second'"
"'1582279.888 per second'"

1809292.824/1582279.888 "1.1434720479743594"

```

Thus the refactored code is ~14 percent faster

MarcusDenker:
For the question if we should do the #isOctetString for WideSymbol creation, we should do another Issue after (as it is not that simple to decide if  that is a good thing or not)

svenvc:
I looked at the Symbol interning code before, I found it confusing and a bit of a mess, it looks to be more complex than it needs to be, probably again due to historical reasons.

For this it is very good that you try to improve it.

Regarding the isOctetString I think the reason might be to make sure that the following is true:
```
'ABC' asSymbol == 'ABC' asWideString asSymbol.
```
Since Symbols are tested with #==, two different subclass instances, ByteSymbol and WideSymbol can never be identical, while the following is true due to the way #= is implemented
```
'ABC' = 'ABC' asWideString.
```
The question is how often will such a strange case occur (i.e. a Latin1 string that becomes a WideSymbol) and is it necessary to pay the (heavy) price for this at symbol interning time ?

MarcusDenker:
But:

```
'ABC'  = 'ABC' asWideString. "true"
```

Thus, the search for an existing symbol (using #like on the WeakSet) will find the symbol regardless if it looks for a wide string or not.

if I change the (current) #intern: to use 

```
	aClass := aStringOrSymbol isWideString ifTrue:[WideSymbol] ifFalse:[ByteSymbol].
```
 I get:

```
'ABC' asSymbol == 'ABC' asWideString asSymbol "true"
 'ABC' asWideString asSymbol == 'ABC' asSymbol  "true"
```

But, now asking for the symbol *after* having asked for the equal wide symbol returns the wide symbol
(eval in one doit due to to GC):

```
'ABCCC' asWideString asSymbol class. 
'ABCCC'  asSymbol class "WideSymbol"
```

But to me this is not a problem, as the only thing that matters is identiy.

svenvc:
I think we can try the change.
I would add a specific test case and maybe a comment.
BTW, would it not be nicer to change the #isWideString test to ask the String class for its Symbol class counterpart ?

MarcusDenker:
I propose to first merge this PR (without the OctetString change), then I open another issue and copy the discussion over.

svenvc:
Beautiful !

PR #11302, opened 2022-05-30
ReleaseTest-Symbol-unused-classvar

MarcusDenker:
- remove Symbol from testNoUnusedClassVariablesLeft
- testNoUncategorizedMethods can check for 2 (it could be 1 if not a new one woul have been added in the meantime)


MarcusDenker:
I merged it as these are just two tests that are green

PR #11133, opened 2022-04-15
Updated Fuel (and users of Fuel) to 5.0.4

theseion:
Also opened PR's for [NewTools](https://github.com/pharo-spec/NewTools/pull/357) and [Roassal3](https://github.com/ObjectProfile/Roassal3/pull/439).

PR #11317, opened 2022-06-01
11316-ClassFactoryForTestCasedeleteClasses-should-call-deleteClass

### LINKED ISSUE #11316: ClassFactoryForTestCase>>#deleteClasses should call #deleteClass:
**Bug description**
The PR https://github.com/pharo-project/pharo/pull/11133 exposes a bug in ClassFactoryForTestCase

ClassFactoryForTestCase>>#deleteClasses should call #deleteClass:

#deleteClasses now calls #delete:, but that does not remove the class from createdClasses

MarcusDenker:
The PR #11133 exposes a bug in ClassFactoryForTestCase

ClassFactoryForTestCase>>#deleteClasses should call #deleteClass:

#deleteClasses now calls #delete:, but that does not remove the class from createdClasses

fixes #11316


theseion:
Good catch!

Issue #11319, opened 2022-06-01
#compactSymbolTable leads to hash collisions


MarcusDenker:
Some simplistic benchmarks... 

Just after image startup of the downloaded image:

```
['asdasdf' asSymbol ] bench. "'1178179.564 per second'" 
```

Now do "Symbol compactSymbolTable":

```
['asdasdf' asSymbol ] bench. "'652087.348 per second'"
```

But if instead, we do a rehash: "Symbol rehash", which adds all instances to a new set (which leads
to growing naturally to a much larger size than 25%):

```
['asdasdf' asSymbol ] bench.   "'2083532.094 per second'"
```

svenvc:
Good observation.

But this benchmark creates the symbol once and then finds it millions of times as already existing, so you are basically testing hashtable/dictionary lookup, which is apparently better when there is some free space.

Still, it is important.

svenvc:
I would indeed try this

MarcusDenker:
I merged #11320, that is, #compactSymbolTable is not used anymore

(we later should deprecate it, with calling #rehash instead)

PR #11320, opened 2022-06-02
Experiment to not use #compactSymbolTable (see #11319)

MarcusDenker:
- in #cleanup call #rehash 
- in #cleanUpAfterMorphicInitialization call Symbol rehash

(see #11319 for more infos)


svenvc:
LGTM

MarcusDenker:
Build on the CI does not show any big improvements of load time, but it not slower either.

For the downloaded artifact, the #asSymbol lookup is indeed faster

PR #11323, opened 2022-06-03
Simplify-subscriptionsHandling 

### LINKED ISSUE #7764: simplify SubscriptionRegistry>>#subscriptionsHandling:

MarcusDenker:
Fixes #7764 (but not tries to implement select:as:)

### DIFF
--- a/src/Announcements-Core/SubscriptionRegistry.class.st
+++ b/src/Announcements-Core/SubscriptionRegistry.class.st
@@ -123,10 +123,7 @@ SubscriptionRegistry >> subscriptionsForClass: subscriberClass [
 
 { #category : #accessing }
 SubscriptionRegistry >> subscriptionsHandling: anAnnouncement [
-	^ Array streamContents: [ :s|
-			subscriptions do: [:each| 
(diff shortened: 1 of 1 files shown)


MarcusDenker:
Failing test is the expected 

- https://github.com/pharo-project/pharo/issues/11324

Ducasse:
I could not understand why the proposed change is better
the current implementation avoid the conversion into an array
by creating it upfront. So to me it is better.

MarcusDenker:
On the Mailinglist it was suggested that the number of subscribers is never large enough for the overhead of #streamContents:

But then, I did not benchmark it.

I should not have open the issue in the first place: the person suggesting the change should have opened the issue *and* pushed for the fix.

PR #11325, opened 2022-06-03
Cleanup Symbol class categories

MarcusDenker:
- move all methods related to the symbol table to category 'symbol table'
- add #hasInterned: 
- deprecate lookup:, instead use #findInterned:
- deprecate findInternedSelector: and remove test

The deprecated methods are not yet in a dedicated package, will be done later after we add it


svenvc:
With all the cleanups and simplifications you have already done, maybe we could one day make a proper SymbolTable class, which would be better, design wise.

MarcusDenker:
Yes, that is an idea for the future

Issue #10958, opened 2022-02-28
[Memory] commentSourcePointer is not needed for MetaClasses


MarcusDenker:
**Bug description**
the source pointer to comments is stored in ClassOrganization. Now both the class and the metaclass have a ClassOrganization instance, but the source pointer for the comment is only needed once.

**Why is this a problem?**
It is not a severe problem (else this would have already been fixed a loooong time ago). But it wastes a little bit of memory.
(just some ~75Kb in the current Pharo10).

**What needs to be done**
The reason why this is stored in the ClassOrganization is just that it was too hard to change Class itself without the bootstrap way back when it was added (ST80 did not save class comments in the .sources/changes).

That means: we just need to move the commentSourcePointer to Class, then MetaClass will not waste space with it anymore,

Ducasse:
Yes!

astares:
Cool! 

Yes bootstrap allows for things that would not have been possible back at the time in Squeak. Pharo rocks!

MarcusDenker:
merged https://github.com/pharo-project/pharo/pull/11449

MarcusDenker:
merged!

PR #11326, opened 2022-06-03
Add commentSourcePointer to ClassDescription

MarcusDenker:
commentSourcePointer is now defined in ClassOrganization, which wastes space.

To be able to move it, we first have to add the ivar using a manual PR.

Fist step for https://github.com/pharo-project/pharo/issues/10958

### DIFF
--- a/src/Kernel/ClassDescription.class.st
+++ b/src/Kernel/ClassDescription.class.st
@@ -12,7 +12,8 @@ Class {
 	#name : #ClassDescription,
 	#superclass : #Behavior,
 	#instVars : [
-		'organization'
+		'organization',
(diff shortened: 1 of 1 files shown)


MarcusDenker:
adding the unused var makes two tests fail:

- NoUnusedVariablesLeftTest>>testNoUnusedInstanceVariablesLeft
- BehaviorTest>>testInstSize

But as the PR is done by manually editing and not loadable, the only way is to merge and fix in a second step the tests

MarcusDenker:
I am merging this and then 
- do a second PR to fix the tests
- a PR to not store the class comment in the class organiser after that

Issue #11342, opened 2022-06-12
RBClass


Ducasse:
- isSharedPool
- removePoolDictionary:
are not invoked or deserve tests

tx CB2

Ducasse:
I'm working on writing tests

PR #11347, opened 2022-06-13
Fixes: #11345 not use  theNonMetaclass and theMetaclass anymore in RB

Ducasse:
convert theNonMetaclass and theMetaclass to their counterpart.

PR #11349, opened 2022-06-13
Fixes: #11342 added some tests to cover to methods that were not called.

Ducasse:
Added some patterns in the definition pattern list.


MarcusDenker:
same as https://github.com/pharo-project/pharo/pull/11347, this PR has odd failures on Windows 

"Tests failing in OSEnvironmentTest with "Illegal leading byte for utf-8 encoding"

Ducasse:
I do not see how this can be related to my changes.

PR #11350, opened 2022-06-13
Method Trailer: avoid perform: in encode for most used cases

MarcusDenker:
Method trailer decoding hard-codes the three most used trailer kinds to avoid a #perfom:

This PR does the same, but for encoding

This avoids #asSymbol and perform: for each compile (and every install in addition  if the trailer has to use variable length encoding).


MarcusDenker:
I put a needs more work as the green CI is wrong: it just did not run the CI at all

MarcusDenker:
failing tests not related

pavel-krivanek (approved):

PR #11358, opened 2022-06-13
Cleanup asSymbol ReservedVariable

MarcusDenker:
reduce the need to call #asSymbol on strings by using symbols when creating self/super/thisContext RBVariablesNodes.

(the code looks not that nice, there seem to be more possibilities for improvements later. This PR just reduces the need to create symbols from the strings)


MarcusDenker:
failing tests are not related

guillep (approved):

Issue #11359, opened 2022-06-14
Tests failing on CI: RSDSMTest>>testLabelsTopPosition on Linux


MarcusDenker:
**Bug description**
After merging , https://github.com/pharo-project/pharo/pull/11328, RSDSMTest>>testLabelsTopPosition is failing often on the Linux CI

```
Error
Assertion failed
Stacktrace
TestFailure
Assertion failed
RSDSMTest(TestAsserter)>>assert:description:resumable:
RSDSMTest(TestAsserter)>>assert:description:
RSDSMTest(Object)>>assert:
[ :box : label | 
		| x x1 x2 |
		x := label encompassingRectangle bottomRight x.
		x1 := box encompassingRectangle topLeft x.
		x2 := box encompassingRectangle topRight x.
		self assert: (x between: x1 and: x2)
		 ] in RSDSMTest>>testLabelsTopPosition
Array(SequenceableCollection)>>with:do:
RSDSMTest>>testLabelsTopPosition
RSDSMTest(TestCase)>>performTest
```

MarcusDenker:
@akevalion

akevalion:
Hi I will check it

akevalion:
I have change the test, will be present in next roassal3 release

MarcusDenker:
@akevalion it would be good to not wait too long: this test fails for *every* run, which means we have to manually check every PR run just to check if indeed just this one tests failed or more.

akevalion:
I will fix it now

akevalion:
@MarcusDenker I have created a pull request, I am waiting the CI results

MarcusDenker:
Fixed by merge

PR #11364, opened 2022-06-15
9705-Cleanup-There-should-be-no-uncategorized-methods 

### LINKED ISSUE #9705: Cleanup: There should be no uncategorized methods
**Describe the problem**
All methods should be categorized. 

There is already #testNoUncategorizedMethods which checks for 57 problematic classes.

**How do I find the problematic ones?**
inspect:

```
Smalltalk globals allBehaviors select: [ :class | class  protocols includes: #'as yet unclassified'. ]
```

**What needs to be done**
All the cases where it makes sense (e.g. for tests) have to be added as false positives to the test, the rest needs to be fixed.

MarcusDenker:
I have added the last case (StDebuggerCommand) as a valid excpetion for now so that we can enable the test in a way that it prints a good error message.

(I will add to the issue tracker entry about StDebuggerCommand to remove it from the test if it gets fixed)

Fixes #9705

### DIFF
--- a/src/ReleaseTests/ProperMethodCategorizationTest.class.st
+++ b/src/ReleaseTests/ProperMethodCategorizationTest.class.st
@@ -141,12 +141,11 @@ ProperMethodCategorizationTest >> testNoUncategorizedMethods [
 		             class protocols includes: Protocol unclassified ].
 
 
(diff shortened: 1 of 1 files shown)


MarcusDenker:
Failing tests not related

- Tests failing on CI: RSDSMTest>>testLabelsTopPosition on Linux #11359

astares (approved):

PR #11365, opened 2022-06-15
10041-PlatformIndependentEnvironmentatifAbsent-should-evaluate-absent-block

### LINKED ISSUE #10041: PlatformIndependentEnvironment>>#at:ifAbsent: should evaluate absent block
looks like 
```Smalltalk
at: aKey ifAbsent: aBlock

	^ nil
```
but should look like 
```Smalltalk
at: aKey ifAbsent: aBlock
        "As there is nowhere to look up the key just evaluate the given block"
	^ aBlock value
```

see discussion in  #10028

MarcusDenker:
This PR just implements the method as described in the issue tracker entry

fixes #10041

### DIFF
--- a/src/System-OSEnvironments/PlatformIndependentEnvironment.class.st
+++ b/src/System-OSEnvironments/PlatformIndependentEnvironment.class.st
@@ -16,8 +16,8 @@ PlatformIndependentEnvironment class >> isAvailable [
 
 { #category : #accessing }
 PlatformIndependentEnvironment >> at: aKey ifAbsent: aBlock [
-
-	^ nil
+   "As there is nowhere to look up the key just evaluate the given block"
+	^ aBlock value
 ]
 
 { #category : #accessing }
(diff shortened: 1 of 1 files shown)


Ducasse:
To me PlatformIndependentEnvironment is a confusing name. 
Is it a language environment or a shell like?
Then PlatformIndependent does not convey the fact that it looks like it is a Null one. 
I do not really understand because if this is an OS environment what is the at:ifAbsent: method doing (I did not see any comment so I have no idea what it is supposed to do). Is it used to get variable such $HOME value?

So please at the minimum add a clear comment

MarcusDenker:
I just fixed the issue tracker entry. To me everything else is better to add another issue, as it is another problem.

MarcusDenker:
I improved the comment a bit more

Ducasse:
Yes let us open another issue.

Issue #11356, opened 2022-06-13
Cleanup: Wrongly named Icons


MarcusDenker:
**Describe the problem**
If you put a traceCr into the #ifPresent: case at the end of ThemeIcons>>#iconNamed:ifNone:, you see a lot of cases where
we use the old style name for icons and Forms.

We should change all the places where we run into the slow backward compatibility case.

astares:
Good catch!

MarcusDenker:
More to do:

- Roassal uses strings: https://github.com/ObjectProfile/Roassal3/issues/456
-

akevalion:
Can we close this issue?

PR #11371, opened 2022-06-15
Icons-Renames-3 

MarcusDenker:
More Icon Renames, a step for https://github.com/pharo-project/pharo/issues/11356

### DIFF
--- a/src/Calypso-SystemPlugins-Critic-Queries/ClyCritiqueSeverityProperty.class.st
+++ b/src/Calypso-SystemPlugins-Critic-Queries/ClyCritiqueSeverityProperty.class.st
@@ -51,7 +51,7 @@ ClyCritiqueSeverityProperty >> color: anObject [
 
 { #category : #accessing }
 ClyCritiqueSeverityProperty >> iconName [
-	^ 'small', name capitalized, 'Icon'
+	^ 'small', name capitalized
 ]
 
 { #category : #testing }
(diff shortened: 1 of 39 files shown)


astares (approved):

Ducasse:
In fact I do not understand why we remove the Icon from the names

astares:
Because it is redundant and also clear from the selector that this the **name** of an icon. We manage a collection of icons (and their name depict what they visually represent).

If you would manage / return a collection of "status" you would also write the **name**:

#active
#stopped

and not

#activeStatus
#stoppedStatus

If you would manage / return a collection of "colors" you would also write the **name**:

#green
#red

and not

#redColor
#greenColor

At least that is how I would see it and I guess this is the intention here.

Ducasse:
Ok I understand. Now it means that we will break all existing users of the old icon names and to me the added value is infinitisimal. So probably that we can get the icon version for a while. Now to me this kind of changes are not worth: The ratio changes / breakage is not in favor.

astares:
@Ducasse 

My understanding of the PR's of @MarcusDenker is that we just fix the places in the image where the wrong name had been
used and where access is slower. By including his changes we become consistent and the access is faster.

As **we do not change** ThemeIcons>>iconNamed:ifNone: itself (which can also deal with the old "xxxIcon" names - we still do not break old usage in external packages. So no animal will be harmed and people can migrate their packages whenever they like without having a problem.

As you can see on:

image

using the right name is better and faster. But using the old name is still possible and we are still backward compatible.

Ducasse:
Ah ok so this is better.

PR #11372, opened 2022-06-15
TestRunner

aboubacardiawara:
The package contains:
- TestRunner2 allows executing tests without UI.
- TestRunnerTreePresenter is a basic UI to run tests. It is based on TestRunner2.

### DIFF
--- a/src/TestRunner-Core/RunExecutionMode.class.st
+++ b/src/TestRunner-Core/RunExecutionMode.class.st
@@ -1,0 +1,34 @@
+"
+I ececute a test suite then return its TestResult.
+"
+Class {
+	#name : #RunExecutionMode,
+	#superclass : #AbstractExecutionMode,
+	#category : #'TestRunner-Core-Configuration'
+}
+
+{ #category : #execution }
+RunExecutionMode class >> executeCase: aTestCase [
+
+	| aResult |
+	aResult := TestResult new.
+	[ 
+	[ 
+	aTestCase announce: TestCaseStarted withResult: aResult.
+	CurrentExecutionEnvironment runTestCase: aTestCase.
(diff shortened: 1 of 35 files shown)


tesonep:
Thanks @aboubacardiawara I will do a review so we can integrate it.

src/TestRunner-Core/TestRunner2.class.st
```diff
@@ -0,0 +1,330 @@
...
+- runPackage:
+- runPackages:
+- switchToDebugMode
+- switchToRunMode
+- accumulateResult
+- doNotAccumulateResult
+- runFailures
+"
+Class {
+	#name : #TestRunner2,
```
astares, line 16: Why not call it AdvancedTestRunner then (following what the comment says)?

aboubacardiawara: Yess i think it is a good idea to change the name.

src/TestRunner-Core/RunExecutionMode.class.st
```diff
@@ -0,0 +1,34 @@
...
+	#category : #'TestRunner-Core-Configuration'
+}
+
+{ #category : #execution }
+RunExecutionMode class >> executeCase: aTestCase [
+
+	| aResult |
+	aResult := TestResult new.
+	[ 
+	[
```
dionisiydk, line 16: It seems it is a full duplication of TestResult>>runCase: method. It can become: 
 ```Smalltalk
[aResult runCase: aTestCase] ensure: [ ... ]
```

dionisiydk: Actually more simple - full method is just "^aTestCase run"

Issue #8661, opened 2021-03-01
Chinese characters cannot input in Pharo


liangbing64:
**Describe the bug**
Latest Image 1168  and VM.
There is no Chinese character input prompter box.
 
Pharo 80  Chinese character input works fine.
Use 80X64 VM to open Pharo 90.
Chinese character input prompter box can open , but input and diplay incorrectly
 Maybe this is VM error.

**To Reproduce**
1. Shift Chinese input state
2. input strings in text pane
3. no Chinese input prompter

**Expected behavior**
Chinese input prompter appear

**Screenshots**
![image](https://user-images.githubusercontent.com/8159448/109457045-08b92800-7a95-11eb-8eb4-c787d89771a9.png)

**Version information:**
 - OS: [Win10]
 - Image: 1168]
 - VM: 90X64 last March 1

guillep:
Hi @liangbing64 ,

the windowing system has been moved from the VM to a normal pharo library in Pharo9. This is probably a side effect of that move. In the past weeks I've started playing with IME support too, there is an ongoing discussion on discord, I invite you to take a look and comment in here: https://discord.com/channels/223421264751099906/582493455574499348/821013293215907904

<blockquote><img src="https://discord.com/assets/ee7c382d9257652a88c8f7b7f22a994d.png" width="48" align="right"><div><img src="/assets/07dca80a102d4149e9736d4b162cff6f.ico" height="14"> Discord</div><div><strong><a href="https://discord.com/channels/223421264751099906/582493455574499348/821013293215907904">Discord - A New Way to Chat with Friends & Communities</a></strong></div><div>Discord is the easiest way to communicate over voice, video, and text.  Chat, hang out, and stay close with your friends and communities.</div></blockquote>

liangbing64:
I use OpenSmalltalk Pharo VM（stable）to run Pharo 9 and 10，the result is  Chinese input prompter can display and work , but backspace key doesnot work.

shixiongfei:
Is there any new progress on this issue?
I have the same problem with Pharo90 and Pharo100.

Ducasse:
@guillep

guillep:
IME support is still work in progress in: https://github.com/tomooda/PharoIM
I know @tomooda has done some progress on that front.
You can help checking the code there, testing it, or improving it.

[...]

PR #11379, opened 2022-06-16
fix 8661: enables IM support in SDL2

tomooda:
This PR provides support for IM-related support for SDL2 library, and solves https://github.com/pharo-project/pharo/issues/8661 .

Some natural languages needs a support agent so-called IM to input letters/symbols including emojis.
IMs are most often provided by OSs and Desktop environments, including macOS, Linux desktops and Windows platforms.
IM accepts keyboard inputs and offers candidate letters/symbols/strings so that the user can choose what should be poured into a text widget.

SDL2 provides a generic API to support IMs.
An SDL2 application should respond to two events: TextEditing and TextInput.
TextEditing is triggered when the user is choosing from candidate strings, and TextInput is triggered when the choice is done.

This PR is to define the events at different layers of frameworks (SDL and morphic) and implements handlers for them.

### DIFF
--- a/src/Morphic-Core/AbstractWorldRenderer.class.st
+++ b/src/Morphic-Core/AbstractWorldRenderer.class.st
@@ -299,6 +299,12 @@ AbstractWorldRenderer >> logRenderingTimeMeasurements: measurements [
 	self class logRenderingTimeMeasurements: measurements
 ]
 
+{ #category : #events }
+AbstractWorldRenderer >> requestTextEditingAt: aRectangle [ 
+	
+	self subclassResponsibility
+]
+
 { #category : #operations }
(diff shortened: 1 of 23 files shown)


MarcusDenker:
Strangely the tests do not finish running and the CI runs are aborted

MarcusDenker:
We need to to find out why the tests can not be run on the CI. I marked it "needs more work" for now.

tomooda:
I found an obsolete instance of MyTestClass is making some troubles. Also found some other problems. I'll dig them out, and then update this PR.

MarcusDenker:
Tests look good now! the two failing tests seem to be not related:

- known failing test on linux for Roassal
- timeout testWorkerPoolDoesNotExceedPoolSizeWhenSchedulingTasksInParallel

PR #11383, opened 2022-06-17
Simplify-BytecodeGenerator-AdditionalLiterals 

MarcusDenker:
The old bytecode set could only have 255 literals. We therefore had to take care to store the additional literals only if we had space.

This is now not beeded anymore and we can simplify


MarcusDenker:
failing tests not related:

- Tests failing on CI: RSDSMTest>>testLabelsTopPosition on Linux #11359

pavel-krivanek (approved):

PR #11395, opened 2022-06-21
11394-Cursor-classresizeForEdge-should-hard-code-common-cases 

### LINKED ISSUE #11394: Cursor class>>#resizeForEdge: should hard-code common cases
**Describe the problem**
Cursor class>>#resizeForEdge: calls #asSymbol during interactive use (moving the mouse over window edges).

There are not many cases, we can easily hard-code the common ones and fall back on perform: for the less common ones.

MarcusDenker:
Hard code the common cases

fixes #11494

### DIFF
--- a/src/Graphics-Display Objects/Cursor.class.st
+++ b/src/Graphics-Display Objects/Cursor.class.st
@@ -880,13 +880,14 @@ Cursor class >> resizeBottomRight [
 
 { #category : #'instance creation' }
 Cursor class >> resizeForEdge: aSymbol [
-	"most common cases hard coded to avoid perform: and asSymbol"
-	aSymbol == #right ifTrue: [ ^self resizeRight ].
-	aSymbol == #left ifTrue: [ ^self resizeLeft ].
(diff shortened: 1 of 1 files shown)


Ducasse:
marcus just a question. Is is faster? Is perform: bad?

MarcusDenker:
It is ~55 times faster..  

But speed is not really the reason for this one, it is called just ones while moving the mouse over a window border.

It is more that I wanted to reduce the rate of Symbol interning during interactive use... I think all of them together
do matter a little (put a traceCr in #asSymbol of String in Pharo10 and compare to Pharo11), but it is more to make it
easier to analyse Symbol interning in the future. (removing the noise).

```
[Cursor resizeForEdge:  #topLeft] bench  

old: "'1675463.322 per second'"
new: "'92913294.541 per second'"

```

(edited as more symbols does not mean slower)

Ducasse:
Cool! Could you add such information in the comments? Because we should educate people.

MarcusDenker:
it has already

"most common cases hard coded to avoid perform: and asSymbol"

PR #9765, opened 2021-08-11
Pr new user interrupt

Alisu:
Adding a new command 'cmd+,' user interrupt that kill all processes.
It respawns an UI process and answer a collection of the killed processes.
The SDL changes are used for adding the shortcut.
Resolve #9762

PR #11397, opened 2022-06-21
revert #9765

### LINKED ISSUE #11396: We should revert https://github.com/pharo-project/pharo/issues/9762
https://github.com/pharo-project/pharo/issues/9762

It was but using kill the system for real so I do not see the point.

Describe the request
UserInterrupt sometimes does not kill the right process, thus interruption never happen.
You need to relaunch an image.

Expected behavior
Having a second user interrupt command that kill all processes.
You don't need to relaunch the image.

Expected development cost
See PR.

guillep:
Fix https://github.com/pharo-project/pharo/issues/11396


Ducasse:
Great!

MarcusDenker (approved):

PR #11402, opened 2022-06-23
Compiler-Cleanup-OldDoItTransformations 

MarcusDenker:
Due to using DoItVariable, we do not need to transform temp read in #DoItIn: methods anymore to use the transformation of variable read and write  (#readVariableNamed: and write)

This PR just deletes the unused code


dionisiydk (approved):
Nice. 
But I remember when I tried to remove the context argument It broke some of the external dependencies. Possibly NewTools expects this argument for DoIt evaluations.

If it's true it would require to adopt tools first to use #withPossibleArguments: version of evaluation method. But maybe it's already adopted.

PR #11405, opened 2022-06-24
Removed RBBasicLintRule >> filteredResult method

### LINKED ISSUE #11340: filteredResult in RBBasicLintRule is not used
I'm not sure that we should remove it but nobody calls this method. 
Tx CB2

jordanmontt:
Fix #11340


MarcusDenker:
This PR exposes (like two others) the bug https://github.com/pharo-project/pharo/issues/11407
(which is completely unrelated to what is done in the PR)

MarcusDenker:
redid the PR to avoid the encoding problem

PR #11408, opened 2022-06-24
CloseAllWindows

RemiDufloer:
We add a tool on taskBar menu to close all windows. When you click on the item an openModal window is created to check if you really want to do so

Capture d’écran 2022-06-24 à 16 56 59

Capture d’écran 2022-06-24 à 16 57 15

Capture d’écran 2022-06-24 à 16 57 29

Author : @RomainDeg and me

### DIFF
--- a/src/Morphic-Widgets-Taskbar/SystemWindow.extension.st
+++ b/src/Morphic-Widgets-Taskbar/SystemWindow.extension.st
@@ -186,7 +186,7 @@ SystemWindow >> taskbarButtonMenu: aMenu [
 		icon: self theme windowCloseForm
 		subMenu: submenu.
 	submenu
-		addToggle: 'close all windows' translated
+		addToggle: 'all windows' translated
 		target: self
 		selector: #taskbarCloseAllWindows
 		getStateSelector: nil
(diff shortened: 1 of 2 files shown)


RemiDufloer:
We added modal windows to confirm closing actions

jgfoster:
* It appears that we already have "Close all->unchanged windows". What is the added benefit of the new command? Is that it closes windows that have unsaved edits without warning you about the edits? That seems more dangerous than helpful!
* It seems that it would be better to have "Close all" not repeated in the submenus.
* I don't think we have close confirmations (unless you will lose unsaved edits) and I don't think the benefit is worth the extra confirmation.

RemiDufloer:
- This new operation create the possibility for the users to close all window and for the window unchanged, a modal window is created, to ask the user if he wants to save or not the changes.
- why ?
- if you don't want extra confirmation maybe just create an option on settings for the beginners ?

jgfoster:
@RemiDufloer, I don't feel very strongly about this, I was just reading through items with "please review!" and offering some feedback. Primarily, I don't see that there is much added value for the new menu option; but I do like consistency and it seems inconsistent to have two submenus start with "close all" and four not start with "close all". 

Screen Shot 2022-07-13 at 11 22 37 AM

src/Morphic-Widgets-Taskbar/SystemWindow.extension.st
```diff
@@ -185,6 +185,12 @@ SystemWindow >> taskbarButtonMenu: aMenu [
 		add: 'Close all'
 		icon: self theme windowCloseForm
 		subMenu: submenu.
+	submenu
+		addToggle: 'close all windows' translated
```
guillep:
```suggestion
		addToggle: 'all windows' translated
```

guillep:
Hi all, I've fixed a bit the message so it avoids the redundant "Close".
I believe that a "close all windows" action is good, we can further iterate it later.

PR #11412, opened 2022-06-27
Cleanup-add-isScripting 

MarcusDenker:
First step for https://github.com/pharo-project/pharo/issues/9921

- add #isScripting where we just had #isForScripting


MarcusDenker:
checked with Esteban, I merge

PR #11417, opened 2022-06-28
Replacing method command refactorings with transformations from Ref2

miguelcamp:
Replaced refactorings with transformations from refactoring 2 and one command that used neither


MarcusDenker (approved):

MarcusDenker:
@miguelcamp refactorings "remove method" and "push up" are broken in Pharo11, it might be due to merging this

MarcusDenker:
- https://github.com/pharo-project/pharo/issues/11435
- https://github.com/pharo-project/pharo/issues/11432

miguelcamp:
Ah then it is probably better to revert these changes while I figure out how to fix it. Sorry about that.

PR #11437, opened 2022-07-04
Allow creating Pharo methods using the Bytecode IR

nrainhart:
This PR extends the OpalCompiler to allow defining methods using the bytecode-like IR.

### DIFF
--- a/src/AST-Core-Tests/RBMethodNodeTest.class.st
+++ b/src/AST-Core-Tests/RBMethodNodeTest.class.st
@@ -11,6 +11,30 @@ Class {
 RBMethodNodeTest >> methodWithArg: someArgName and: someAnotherArgName [
 ]
 
+{ #category : #helpers }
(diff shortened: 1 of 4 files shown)


dionisiydk:
> This PR extends the OpalCompiler to allow defining methods using the bytecode-like IR.

Could you describe the purpose of this feature? When we are going to use it instead of regular code?

nrainhart:
The goal is to be able to create methods that don't have a syntactic representation in Pharo code, but could be constructed using bytecode.

That's useful for bytecode prototyping (which is our current motivation), and may also be used for low-level optimizations. E.g. as far as I understand we're not currently using the `bytecodePrimPointX`/`bytecodePrimPointY` bytecodes. Implementing `Point` methods as `opalBytecodeMethod`s would allow use these bytecodes without modifying the compilation process.

PR #11441, opened 2022-07-07
fix-tests-commentSourcePointer

MarcusDenker:
fix two failing tests after add class comment pointer to ass description.

After we use it, we will remove the validExcpetion from testNoUnusedInstanceVariablesLeft


MarcusDenker:
( I merge this as it is just fixing tests that should be green)

PR #11442, opened 2022-07-07
fix ShiftClassInstaller that fails to migrate some objects

tomooda:
The following expression fails because the ShiftClassInstaller>>copyObject:to: measures the indexed variable size by `size` which should be `basicSize`.
```
ShiftClassInstaller new 
    oldClass: MethodDictionary;  
    copyObject: (MethodDictionary new: 3) to: MethodDictionary
```


MarcusDenker:
I will close and reopen to run with the latest (exactly this run has some failing tests that are fixed)

MarcusDenker:
failing test not related

PR #11446, opened 2022-07-08
Pharo Image support for Pharo VM's arithmetic SIMD bytecodes

nrainhart:
This PR adds changes required to integrate https://github.com/pharo-project/pharo-vm/pull/444 into the Pharo VM:
- Renames `FloatArray` to `Float32Array`
- Adds `Float64Array` class
- Adds both FloatArray classes to the Special Objects Array, so that they can be used by the Pharo VM

### DIFF
--- a/src/Collections-Abstract-Tests/CollectionRootTest.class.st
+++ b/src/Collections-Abstract-Tests/CollectionRootTest.class.st
@@ -125,6 +125,29 @@ CollectionRootTest >> testDetectIfFoundWhenSomethingIsFoundIgnoringTheFoundObjec
 	self assert: wasFound
 ]
 
+{ #category : #'tests - iterating' }
+CollectionRootTest >> testDo2 [
+	"dc: Bad test, it assumes that a new instance of #speciesClass allows addition with #add:. This is not the case of Interval for which species is Array."
+	"res := self speciesClass new.  
+	self collection do: [:each | res add: each class].
+	self assert: res = self result. "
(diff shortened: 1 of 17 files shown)


src/Collections-Abstract-Tests/CollectionRootTest.class.st
```diff
@@ -125,6 +125,29 @@ CollectionRootTest >> testDetectIfFoundWhenSomethingIsFoundIgnoringTheFoundObjec
 	self assert: wasFound
 ]
 
+{ #category : #'tests - iterating' }
+CollectionRootTest >> testDo2 [
```
nrainhart: This test popped up while doing the automatic rename from `FloatArray` to `Float32Array`. Should we remove it?

src/Collections-Sequenceable-Tests/LinkedListTest.class.st
```diff
@@ -811,6 +811,29 @@ LinkedListTest >> testCreateAs [
 		self assert: nextElementOfLinkedList equals: nextElementOfAnotherCollection]
 ]
 
+{ #category : #'tests - iterating' }
+LinkedListTest >> testDo2 [
```
nrainhart: Ditto

src/Collections-Abstract-Tests/CollectionRootTest.class.st
nrainhart: Yes. Tests are indeed failing because of that. I'll delete them

PR #11447, opened 2022-07-08
Revert "9921-unify-isScripting-and-isForScripting "

MarcusDenker:
Reverts pharo-project/pharo#11416

trying to check if this broke class definition compiling


MarcusDenker:
Nope, not the problem

PR #11449, opened 2022-07-09
Class Comments: use ivar in ClassDesciption

MarcusDenker:
This PR:

- changes all methods related to class comments in ClassDescription to use the source pointer of ClassDescription
- Change ClassOrganzier to forward to the class and remove the ivar there

after this we should change all code that calls the methods on the organizer and deprecate them

### DIFF
--- a/src/Kernel/ClassDescription.class.st
+++ b/src/Kernel/ClassDescription.class.st
@@ -1,28 +1,12 @@
 "
-Please comment me using the following template inspired by Class Responsibility Collaborator (CRC) design:
-
-For the Class part:  State a one line summary. For example, ""I represent a paragraph of text"".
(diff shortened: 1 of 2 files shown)


dionisiydk (approved):

MarcusDenker:
The MAC CI is down, but the change is not platform specific

Bajger (approved):
I just wonder that no changes in unit tests needed. Nice!

MarcusDenker:
we need to change one unit test back: the one that tests for un-used ivars has ClassDiscription as a false positive, I will fix that later.

MarcusDenker:
strangely another PR shows that this introduces problems, I revert it for now

PR #11453, opened 2022-07-11
README.md and Bootstrap cleanup

jgfoster:
* Update README.md links to 11.0
* Use defaults for BUILD_NUMBER and BOOTSTRAP_ARCH if not provided (no need to require them)
* Provide Dockerfile example
* Remove instructions for adding 32-bit libraries to Docker container
* Remove previous VM so we don't get an error if the bootstrap is rerun
* Use the latest VM for the bootstrap


astares:
@jgfoster 
Why do you switch to 11 (leaving out 10). Usually the last stable (Pharo 10 at the moment) should be used, as Pharo 11 is alpha and a moving target still. No?

jgfoster:
@astares, that's a good question and to be honest I didn't think about it very much; I just assumed that it should be the latest. But now, having thought about it, I think I can make a case for '11'. 

* The "Download Pharo" link talks about a "stable version" and this leads visitors to the Launcher download which will give a choice of versions. So people who are just looking for Pharo will use the Launcher. 
* Given that the changes are made to a `README.md` in a branch named `Pharo11`, I think it makes sense for the "latest build" to point to the latest build of Pharo11. This is especially true when discussing the Bootstrap process—any bootstrap from this branch will create a Pharo11 image.

But, my primary interest is in the bootstrap process and if you think it would be better for the Pharo11 branch to point to the latest build of Pharo10, I won't quibble; mostly I appreciate getting some feedback!

astares:
Maybe @guillep can give some feedback here

MarcusDenker:
Failing test is not related:

- SpJobListPresenterTest>>testJobIsFinishedWhenWaitingMoreThanWorkBlockDuration

MarcusDenker:
Lots of people are on holidays right now, so reviewing is a bit slower than in normal times

bootstrap/scripts/2-download.sh
```diff
@@ -23,12 +23,13 @@ function download_to {
...
-    mkdir ${BOOTSTRAP_DOWNLOADS}/vmtarget
-    cd ${BOOTSTRAP_DOWNLOADS}/vmtarget
-    ${BOOTSTRAP_REPOSITORY}/bootstrap/scripts/getPharoVM.sh 90 vm $BOOTSTRAP_ARCH
-    cd -
-    echo "Target VM: $(${VM} --version)"
+	rm -rf "${BOOTSTRAP_DOWNLOADS}/vmtarget"
+	# Downloads a SPUR vm for the configured architecture
+	mkdir ${BOOTSTRAP_DOWNLOADS}/vmtarget
+	cd ${BOOTSTRAP_DOWNLOADS}/vmtarget
+	${BOOTSTRAP_REPOSITORY}/bootstrap/scripts/getPharoVM.sh 110 vm $BOOTSTRAP_ARCH
```
guillep: I think the entire PR is good.
The only line that makes me a bit worried is this one indeed.
I'd like to build (unstable) version X with a stable version <X.

So I'd feel better if we put Pharo 10 in here. What do you think?

jgfoster: If a new primitive was added to the VM and a new method using that primitive was added to the image, then they should be used together. Trying to debug an image with a new primitive without VM support would be difficult. It seems to me that they should all stay together.

astares: Yes - but at the moment we do not have this as a case, right? So maybe it makes sense to go with the last *stable* (<X) until we have to solve such a situation and have a need to go the *unstable* (X) route. No?

jgfoster: I guess I don't feel that strongly, but it seems that we also do not have the case where the 11 VM doesn't work. So maybe go with the latest VM "until we have to solve such a situation." Also, it seems that if we need the latest VM and it doesn't come in automatically, it will not be very obvious where to go look to solve the problem. So, which is more likely, the latest VM is broken or the latest VM is needed but missing? Also, my impression is that the VM is on version 9 and it is being used by several versions of Pharo. If the same thing is being used for Pharo 9, 10, and 11, what advantage is there in using the Pharo10 VM if it is the same as 9 and 11? If they are going to use the same thing (which I don't know), then the numbers should match. But, again, my interest is in other parts of the script; I don't care that much here and can change if it will help get the rest accepted.

astares: Discussion is mainly about the risk through dependencies: if we have the stable VM for the bootstrap we would go with less risk. When using stable with < 10 for the bootstrap the VM is not the moving target and the bootstrap process should work on that end. 
That means contributions can be done for image and bootstrap Smalltalk code more savely (usually we have more contributions to image side than to VM code). 

Only when situation would be needed to change all three at the same time: VM + image + boostrap for a new feature introduction (primitive, ....) the unstable version (here X in our case now) would make sense. At least that is how I see it.

jgfoster: Changed to VM 100 as requested.

jgfoster:
@guillep , @astares , Any further comments or requests?

Ducasse:
Guille is on vacation for two weeks

PR #11455, opened 2022-07-11
Add an "exit condition" to the UI loop to allow a former UI process t…

isCzech:
…o finish its last cycle and terminate automatically when replaced by a new UI. This change could allow UI processes blocked on custom semaphores to continue waiting for a signal and finish their previous job before terminating automatically. Useful for keeping the UI responsive during evaluation of examples like this:

s := Semaphore new.
[1/0. s signal] fork.
s wait

This is a first approxiamation of the solution and may require some polishing :)

### DIFF
--- a/src/Kernel/Process.class.st
+++ b/src/Kernel/Process.class.st
@@ -366,7 +366,7 @@ Process >> isBlocked [
 	^false
 ]
 
-{ #category : 'testing'. }
+{ #category : 'testing' }
 Process >> isSuspended [
 	"A #suspended process will not run until sent #resume. New processes
 	start in this state (though the #fork family sends #resume internally),
--- a/src/Kernel/Process.class.st
+++ b/src/Kernel/Process.class.st
@@ -359,10 +359,11 @@ Process >> isReady [
 
 { #category : 'testing' }
 Process >> isBlocked [
-	"Answer true if blocked on a semaphore."
-	self isActiveProcess ifTrue: [ ^false ].
-	self isTerminated ifTrue: [ ^false ].
-	^myList class == Semaphore
+	"Answer true if blocked on a condition variable.
+	A process is blocked if it is waiting on some list
+	(e.g. a Semaphore), other than the runnable process lists."
+	myList ifNotNil: [ :list | ^list class ~~ ProcessList ].
+	^false
 ]
 
 { #category : 'testing'. }
--- a/src/Kernel/Process.class.st
+++ b/src/Kernel/Process.class.st
@@ -358,6 +358,14 @@ Process >> isReady [
 ]
 
 { #category : 'testing' }
+Process >> isBlocked [
+	"Answer true if blocked on a semaphore."
+	self isActiveProcess ifTrue: [ ^false ].
+	self isTerminated ifTrue: [ ^false ].
+	^myList class == Semaphore
+]
+
+{ #category : 'testing'. }
 Process >> isSuspended [
 	"A #suspended process will not run until sent #resume. New processes
 	start in this state (though the #fork family sends #resume internally),
--- a/src/Polymorph-Widgets/MorphicUIManager.class.st
+++ b/src/Polymorph-Widgets/MorphicUIManager.class.st
@@ -600,7 +600,7 @@ MorphicUIManager >> showWaitCursorWhile: aBlock [
 MorphicUIManager >> spawnNewProcess [
 
 	UIProcess := [
-		MorphicRenderLoop new doOneCycleWhile: [ self uiProcess isActiveProcess ]
+		MorphicRenderLoop new doOneCycleWhile: [ self uiProcess isActive ]
 	] newProcess priority: Processor userSchedulingPriority.
 	UIProcess name: 'Morphic UI Process'.
 	UIProcess resume
--- a/src/Polymorph-Widgets/MorphicUIManager.class.st
+++ b/src/Polymorph-Widgets/MorphicUIManager.class.st
@@ -599,11 +599,13 @@ MorphicUIManager >> showWaitCursorWhile: aBlock [
 { #category : 'ui process' }
 MorphicUIManager >> spawnNewProcess [
 
-	UIProcess := [
-		MorphicRenderLoop new doOneCycleWhile: [ self uiProcess isActive ]
+	| newUIProcess |
+	newUIProcess := [
+		MorphicRenderLoop new doOneCycleWhile: [ newUIProcess isActive ]
 	] newProcess priority: Processor userSchedulingPriority.
-	UIProcess name: 'Morphic UI Process'.
-	UIProcess resume
+	newUIProcess name: 'Morphic UI Process'.
+	UIProcess := newUIProcess.
+	newUIProcess resume
 ]
 
 { #category : 'ui process' }
--- a/src/Polymorph-Widgets/MorphicUIManager.class.st
+++ b/src/Polymorph-Widgets/MorphicUIManager.class.st
@@ -600,7 +600,7 @@ MorphicUIManager >> showWaitCursorWhile: aBlock [
 MorphicUIManager >> spawnNewProcess [
 
 	UIProcess := [
-		MorphicRenderLoop new doOneCycleWhile: [ UIManager default uiProcess isActiveProcess ]
+		MorphicRenderLoop new doOneCycleWhile: [ self uiProcess isActiveProcess ]
 	] newProcess priority: Processor userSchedulingPriority.
 	UIProcess name: 'Morphic UI Process'.
 	UIProcess resume
--- a/src/Polymorph-Widgets/MorphicUIManager.class.st
+++ b/src/Polymorph-Widgets/MorphicUIManager.class.st
@@ -599,11 +599,11 @@ MorphicUIManager >> showWaitCursorWhile: aBlock [
 { #category : 'ui process' }
 MorphicUIManager >> spawnNewProcess [
 
-	UIProcess := [ MorphicRenderLoop new doOneCycleWhile: [ true ] ] newProcess priority:
-		             Processor userSchedulingPriority.
-	UIProcess
-		name: 'Morphic UI Process';
-		resume
+	UIProcess := [
+		MorphicRenderLoop new doOneCycleWhile: [ UIManager default uiProcess isActiveProcess ]
+	] newProcess priority: Processor userSchedulingPriority.
+	UIProcess name: 'Morphic UI Process'.
+	UIProcess resume
 ]
 
 { #category : 'ui process' }
--- a/src/Debugger-Oups/OupsDebuggerSystem.class.st
+++ b/src/Debugger-Oups/OupsDebuggerSystem.class.st
@@ -137,11 +137,9 @@ OupsDebuggerSystem >> spawnNewUIProcess [
 { #category : 'helpers' }
 OupsDebuggerSystem >> spawnNewUIProcessIfNecessary: aDebugRequest [
 
-	"If aDebugRequest is about debugging the UI process, we must create a new UI process to take its place. Because the debugged process will be suspended at some point, and suspending the UI process means freezing the UI of the image; similarly if the UI is currently blocked, create a new UI and let the blocked UI finish its last cycle only."
+	"If aDebugRequest is about debugging the UI process, we must create a new UI process to take its place. Because the debugged process will be suspended at some point, and suspending the UI process means freezing the UI of the image; for simplicity, in case of any debug request create a new UI and let the currently active UI finish its last cycle only."
 
-	(aDebugRequest debugSession isAboutUIProcess or: [
-		self defaultUIManager uiProcess isBlocked]) ifTrue: [ 
-			self spawnNewUIProcess ]
+	self spawnNewUIProcess
 ]
 
 { #category : 'helpers' }
--- a/src/Debugger-Oups/OupsDebuggerSystem.class.st
+++ b/src/Debugger-Oups/OupsDebuggerSystem.class.st
@@ -137,10 +137,11 @@ OupsDebuggerSystem >> spawnNewUIProcess [
 { #category : 'helpers' }
 OupsDebuggerSystem >> spawnNewUIProcessIfNecessary: aDebugRequest [
 
-	"If aDebugRequest is about debugging the UI process, we must create a new UI process to take its place. Because the debugged process will be suspended at some point, and suspending the UI process means freezing the UI of the image"
+	"If aDebugRequest is about debugging the UI process, we must create a new UI process to take its place. Because the debugged process will be suspended at some point, and suspending the UI process means freezing the UI of the image; similarly if the UI is currently blocked, create a new UI and let the blocked UI finish its last cycle only."
 
-	aDebugRequest debugSession isAboutUIProcess ifTrue: [
-		self spawnNewUIProcess ]
+	(aDebugRequest debugSession isAboutUIProcess or: [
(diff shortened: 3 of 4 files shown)


src/Kernel/Process.class.st
```diff
@@ -358,6 +358,14 @@ Process >> isActiveProcess [
 	^ self == Processor activeProcess
 ]
 
+{ #category : #testing }
+Process >> isBlocked [
+	"Answer true if blocked on a semaphore."
+	self isActiveProcess ifTrue: [ ^false ].
+	self isTerminated ifTrue: [ ^false ].
+	^myList class == Semaphore
```
dionisiydk: Funny I did not think that we are speaking about Semaphore case specifically. There can be any other blocking scenarious. And generally it is not only about blocking. Just think that we are running some long computation at the end (which may periodically sleep or not):
```Smalltalk
[1/0] fork.
[ 10 factorial ] repeat
```

isCzech: of course... baby steps :D I've just covered the most pressing case for the beginning ;)

MarcusDenker:
failing test not related:

- SpJobListPresenterTest>>testJobIsFinishedWhenWaitingMoreThanWorkBlockDuration

src/Kernel/Process.class.st

I've been thinking about your comments and has done some testing in Cuis to make sure I'm not presenting an idea that is only applicable in Pharo; it works :)

> I did not think that we are speaking about Semaphore case specifically. 

I'd like to clarify what I mean by "blocked": I'd like to consider it one of the classic 5 process states: active, runnable, suspended, terminated and waiting (= blocked); I chose #isBlocked to be compatible with Squeak/Cuis terminology. Because Pharo and Cuis implement Mutex using a Semaphore I think the only case when a process is waiting is when it is waiting on a semaphore. Please correct me if I missed some "waiting" scenario.

> There can be any other blocking scenarios. 

What is it you mean by blocking; what other blocking scenarios did you have in mind?

> And generally it is not only about blocking. Just think that we are running some long computation at the end (which may periodically sleep or not):
```
[1/0] fork.
[ 10 factorial ] repeat
```
Yes, this is a valid point; it'd be great to address this one as well; however, it seems to me it's a slightly different sort of a problem and I'd suggest to deal with it separately from the simpler "waiting" scenario. The point is in the example above the UI is not "idly" waiting for something but it's actively "busy" doing something. I expect a slightly different (or more general) approach might be needed, and hence I'd like to leave it aside for the moment.

In your example, I'd expect the UI just go on working; why should it stop and look around; any other process than UI wouldn't... or maybe the UI should spawn a new UI process in such a case, I don't have an opinion yet :) But what about in the following example, should the UI deal with the 1/0 error?
```
[1/0] fork.
[ 10000 factorial. Processor yield ] repeat
```
I was still inclined to say no but look at this example:
```
[1/0] fork.
[ 10000 factorial. 5 seconds asDelay wait ] repeat
```
Why keep the UI idle for 5 seconds and not let it deal with the 1/0 error in the meantime? (this PR addresses this scenario).

Or even this example:
```
[[1/0] fork.
[ 10000 factorial. Processor yield ] repeat] fork
```
Of course you'd expect the UI take care of the 1/0 error; and I believe the UI should ideally behave as any other process running the same example because the computation’s result shouldn’t depend on whether it runs in the UI or not because it's an implementation detail the user shouldn't need to know or care about.

So to summarize: I suggest to implement the "easy" part, i.e. spawn a new UI if the old UI is waiting on a semaphore, and deal with the "busy" UI scenario later. 

And of course, there are these other issues with "sub-cycles":
```
[1/0] fork.
5 seconds asDelay wait.
self confirm: 'test?'
```
and who knows what else... Why not find a clean solution for just the "waiting" UIs first :)

dionisiydk: > I was still inclined to say no but look at this example:
> 
> ```
> [1/0] fork.
> [ 10000 factorial. 5 seconds asDelay wait ] repeat
> ```
> 
> Why keep the UI idle for 5 seconds and not let it deal with the 1/0 error in the meantime? (this PR addresses this scenario).

Well, increase the priority of the forked process and PR will not help:
 ```Smalltalk
[1/0] forkAt: Processor activePriority + 1.
[ 10000 factorial. 5 seconds asDelay wait ] repeat
```

To be honest I more inclined now that this fix is not a good idea at all. My examples just demonstrate inconsistencies which will beat the user in a real life scenarios. And in the real life I have no idea what happens inside the code I am playing in the playground:
```Smalltalk
SomeNewCoolLib new execute
```
If the code does some multithreaded computations/networking it will wait at random time for the remote data and at random time it will do the computation when data received. Now if there is a bug then in some cases I can luckily get the debugger (when the background error is thrown exactly during the waiting period in the UI) and at the another time it will be just a locked UI (when the background error is thrown during a computation time in the UI). 
It means that the same code started in the workspace may occasionally show a different behaviour. I would prefer to always have a same one. Otherwise it is a breaking brain moment. 

So I would not distinguish the cases when UI is locked on semaphore or locked on computation. It does not matter IMHO - UI is just busy.

Another thought. All these problems we discussing here is about the user experience with workspace (playground). Maybe it should just never run code in the UI Process. Then the problem will just gone.

isCzech: That's a lot of brain food, thank you!

> Another thought. All these problems we discussing here is about the user experience with workspace (playground). Maybe it should just never run code in the UI Process. Then the problem will just gone.

Yes, it occurred to me when I learned the Playground examples are run "in the UI" but it is still beyond my skill level to try to analyze and implement the alternative; there must have been a reason Squeak and its forks (and VW too I guess) chose this strategy; it'd help to know why :)

dionisiydk: > I'd like to clarify what I mean by "blocked": I'd like to consider it one of the classic 5 process states: active, runnable, suspended, terminated and waiting (= blocked); I chose #isBlocked to be compatible with Squeak/Cuis terminology. Because Pharo and Cuis implement Mutex using a Semaphore I think the only case when a process is waiting is when it is waiting on a semaphore. Please correct me if I missed some "waiting" scenario.

I guess you are right about the current state of the image. There is implementation of another synchronization primitive but it is not integrated. See PrimitiveLock from https://github.com/pharo-ide/ReadWriteLock/tree/master/ReadWriteLock . 
Anyway it's a good idea to not hardcode the classes and rely on messages.

isCzech: > There is implementation of another synchronization primitive but it is not integrated. See PrimitiveLock from https://github.com/pharo-ide/ReadWriteLock/tree/master/ReadWriteLock .

Thanks for pointing this out; in that case it'd be better to implement #isBlocked using
```
isBlocked
	"Answer true if blocked on a condition variable.
	A process is blocked if it is waiting on some list
	(e.g. a Semaphore), other than the runnable process lists."

	myList ifNotNil: [ :list | ^list class ~~ ProcessList ].
	^false
```

or like in Squeak (by Eliot):
```
isBlocked
	"Answer true if blocked on a condition variable.
	 A process is blocked if it is waiting on some list
	 (e.g. a Semaphore), other than the runnable process lists."

	| myPriority |
	"Grab my priority now. Even though evaluation is strictly right-to-left,
	accessing Processor could involve a send."
	myPriority := priority.
	^myList
		ifNil: [false]
		ifNotNil: [:list | list ~~ (Processor waitingProcessesAt: myPriority)]
```

to be somewhat future-proof :)

isCzech:
Hi Denis,

> Well, increase the priority of the forked process and PR will not help:

[1/0] forkAt: Processor activePriority + 1.
[ 10000 factorial. 5 seconds asDelay wait ] repeat

Perfect example :)

I understand now that when the [1/0] process requests a new debugger the UI is waiting on a ProcessList, i.e. not on a Semaphore, and yet we'd like the UI to open a new debugger. It's clear then, as you suggested, that distinguishing where exactly the UI is waiting is not helpful.

I'd like to suggest the simplest possible solution here: *Always* create a new UI process for a new debug request. This way you'll make sure the current UI will be able to finish whatever it is doing (a Playground example or any deferred action) and automatically terminate when it reaches the end of its current cycle. And at the same time the new UI will be able to open the requested debugger immediately and start acting as a proper new UI.

An alternative could be trying to distinguish when the UI is idly waiting (on a ProcessList between cycles or an intercycle delay semaphore) or busy running it's intracycle errands (e.g. by placing a boolean flag inside the cycle's code) but I can't see any added value here.

So unless I missed some fatal downside I prefer the former, simple solution.

> It means that the same code started in the workspace may occasionally show a different behavior. I would prefer to always have a same one. Otherwise it is a breaking brain moment.

Can't agree more :) It seems to me though "any" code fragment should now produce identical outcome regardless whether it's run in the UI (Playground) or otherwise (forked) ("any" code meaning, of course, except e.g. a code referring to the UI process itself :) etc. ). Or at least we're getting closer to the "ideal".

Surprisingly, even the following example works, i.e. opens a debugger instead of getting stuck, but in this case it is also thanks to some unrelated system interrupts at play preempting the old UI and allowing the new UI run..

[1/0] fork.
[ 10 factorial ] repeat

Please check the latest commit. I'd very much appreciate your view.

isCzech:
failing test in CI: CoNarrowHistoryFetcherTest>>testNarrowingReturnsSameElementsThatCallingDirectly
... no idea how could be related
The same test OK in the image

dionisiydk (approved):

dionisiydk:
@isCzech I like how simple it is now. Thanks for this work.
I would just add tests for isBlocked method. 

CoNarrowHistoryFetcherTest is not related. I have another PR with same errors.

guillep:
Hi, I was checking this, long discussion.
Can somebody do an executive summary? @isCzech @dionisiydk do you think this is *good enough* to integrate?

dionisiydk:
Hi @guillep .
Summary:
PR is a general solution to the following issue:
```Smalltalk
s := Semaphore new.
[1/0. s signal] fork.
s wait
```
Execute this code from the playground and the image will be locked. The debugger from the background process will not be shown.  Another example without semaphore:
```Smalltalk
s := Semaphore new.
[1/0. s signal] fork.
[] repeat.
```
The explanation is quite simple: to open the debugger the UI process should be "free" to process the "debugger request". If the UI process is busy on something then any new UI action (like a debugger request) will wait until the UI process will be free.

The proposed solution is quite elegant: 
- debugger request should always start new UI process.
- old UI process is left to be completed by itself (using "stop condition" for its UI loop) 

Does it make sense?

dionisiydk:
@isCzech could you resolve conflicts?

isCzech:
Hi @dionisiydk I wonder why has the conflict popped up but I have no experience in this regard. IMO nothing changed in the original Pharo code and the suggestion in this PR is still valid. May I ask  you to doublecheck? I trust you have lot of experience :)

dionisiydk:
> Hi @dionisiydk I wonder why has the conflict popped up but I have no experience in this regard. IMO nothing changed in the original Pharo code and the suggestion in this PR is still valid. May I ask you to doublecheck? I trust you have lot of experience :)

It seems you did everything.
Tests look unrelated. I guess any PR is failing now

demarey:
I restarted the build on CI and there are failing tests.
`SystemDependenciesTest>>#testExternalMorphicCoreDependencies` looks now broken.
Also, while, at first sight, it does not look related, we have around 10 other tests failing. See https://ci.inria.fr/pharo-ci-jenkins2/blue/organizations/jenkins/Test%20pending%20pull%20request%20and%20branch%20Pipeline/detail/PR-11455/8/tests/ for more details.
Does the change pass all image tests locally on a fresh P11 image?
If so, maybe these changes can be rebased on top of last P11 (for which CI is green: https://ci.inria.fr/pharo-ci-jenkins2/job/Test%20pending%20pull%20request%20and%20branch%20Pipeline/job/Pharo11/)

<blockquote><div><strong><a href="https://ci.inria.fr/pharo-ci-jenkins2/blue/organizations/jenkins/Test%20pending%20pull%20request%20and%20branch%20Pipeline/detail/PR-11455/8/tests/">Jenkins</a></strong></div></blockquote>
<blockquote><div><strong><a href="https://ci.inria.fr/pharo-ci-jenkins2/job/Test%20pending%20pull%20request%20and%20branch%20Pipeline/job/Pharo11/">Pharo11 [Pharo] [Jenkins]</a></strong></div></blockquote>

jecisc:
I updated the base to be Pharo 12

src/Polymorph-Widgets/MorphicUIManager.class.st
```diff
@@ -648,11 +648,11 @@ MorphicUIManager >> showWaitCursorWhile: aBlock [
 { #category : #'ui process' }
 MorphicUIManager >> spawnNewProcess [
```
dionisiydk, line 600: For Pharo12 method neeeds to be copied to MorphicCoreUIManager (where it is duplicated)

daniels220:
I am pretty familiar with Dolphin Smalltalk's UI code, and it indeed uses a very similar core approach of exiting the message loop when that particular process is no longer the One True UI Process, allowing the new one to take over smoothly. A neat bit of convergent evolution!

However I have one concern: Message processing in Dolphin is much more fine-grained, as it can rely on Windows to do most of the heavy lifting WRT determining need-to-paint—so there's no internal concept of a render cycle, it just processes `WM_PAINT` messages as they are posted by the OS. As such, even with multiple UI processes in play, there is a reasonable expectation that messages will be handled in order, and a given need for painting will be satisfied only once. In Pharo, I would be concerned about the tail-end of a previous UI process clashing with the new one—`WorldState>>doOneCycleFor:` does a _lot_ of work before returning to the check in `doOneCycleWhile:` (and yeah, there are several calls in between these two, but if there's only one World it looks like a straight shot—if there's more than one, the situation only gets worse). The equivalent in Dolphin handles **one** mouse/keyboard event OR paints **one** Win32 "window"/Morph-analogue.

This actually _would_ already cause problems with the existing case where a new UI process is spawned, that of debugging the UI process, except that there is an awful hack of a workaround already in place. In `DebugSession>>terminate`:

```
self interruptedProcess
	ifNotNil: [
		"Assume the user closed the debugger. Simply kill the interrupted process."
		self interruptedProcess terminate.
		self clear.
		Smalltalk installLowSpaceWatcher.  "restart low space handler" ]
	ifNil: [
		"Assume the interrupted process was resumed."
		"Kill the active process if the error was in the UI as there should be only one UI process."
		self isAboutUIProcess
			ifTrue: [Processor terminateActive] ]
```

It looks like `interruptedProcess` is nil'ed when hitting "Resume"? But going by the comments, this is the (only!) reason that debug-then-resume'ing the UI process doesn't permanently leave two UI processes running and fighting over the event queue, painting, etc. But...okay, I thought this would mean that you couldn't resume a DoIt, because it would kill the process right away. For some reason it doesn't, and I don't have time to chase down why. What _does_ happen is that the debugger window remains open until the DoIt finishes, which is just weird—shouldn't it close just as or even before the process resumes execution? But in that case it clearly _would_ end up killing the process. So there's some definite improvement possible here once UI processes can cooperate better.

I'm not sure how to address this cleanly without a need for massive refactoring, or indeed at all—I know too little about how Pharo interacts with the host OS to be confident. I have some vague ideas about how to tease things out if anyone is interested, but it'd be easiest to chat about it on Discord or something rather than try to write it up all at once here.

(Another difference that's worth noting and may be relevant—in Dolphin the UI process blocks entirely waiting for new events from the OS, relying on the idle process to take over and quiesce the VM. Pharo seems instead to explicitly poll for events every ~16ms and execute a (potential) render cycle each time. Clearly there is dirty-checking as idle CPU usage is minimal, but the image still wakes and executes at least some code each cycle.)

tesonep:
Stale PR

PR #11463, opened 2022-07-13
DoIts without method header and reformatting

dionisiydk:
DoIts without method header and reformatting:
simpleDoIt

- RBParser is refactored to simplify method and doIts parsing/compilation
 - common parts are extracted to parseMethodBodyInto:
    - pragmas are supported in doIts (see #testPragmaInExpression):
      - it was some strange constraint

      - compilation flags can be now enabled in workspace
  - RBReturnNode is improved with better var name for position: #start instead of #return 
  - RBDoItMethodNode is introduced to represent method nodes in doIt ASTs
 - SHRBTextStyler>>isWorkspace uses isScripting which respects semantic scope to correctly highlight DoIts without headers

PR does not fix highlighting in the debugger and inspector for no-header DoIts. These fixes are in their repos:
- https://github.com/pharo-spec/NewTools/pull/400
- https://github.com/pharo-spec/Spec/pull/1296

### DIFF
--- a/src/AST-Core/ASTCacheMissStrategy.class.st
+++ b/src/AST-Core/ASTCacheMissStrategy.class.st
@@ -16,5 +16,5 @@ Class {
 
 { #category : #accessing }
 ASTCacheMissStrategy >> getASTFor: aCompiledMethod [
-	^ aCompiledMethod parseTree doSemanticAnalysisIn: aCompiledMethod methodClass
+	^ aCompiledMethod parseTree
 ]
--- a/src/DrTests-CommentsToTests-Tests/CommentsToTestsTest.class.st
+++ b/src/DrTests-CommentsToTests-Tests/CommentsToTestsTest.class.st
@@ -28,7 +28,7 @@ CommentsToTestsTest >> testCommentWithSyntaxError [
 	docComment := thisContext method ast pharoDocCommentNodes first.
 	commentTestCase := CommentTestCase for: docComment.
 	
-	self should: [commentTestCase testIt] raise: SyntaxErrorNotification
+	self should: [commentTestCase testIt] raise: RuntimeSyntaxError
 ]
 
 { #category : #tests }
--- a/src/OpalCompiler-Core/OpalCompiler.class.st
+++ b/src/OpalCompiler-Core/OpalCompiler.class.st
@@ -259,17 +259,7 @@ OpalCompiler >> compile: textOrString [
 { #category : #private }
 OpalCompiler >> compileDoItFromAST [ 
 	| method |
-	ast := self semanticScope transformDoItExpressionToMethodAST: ast.
-	ast compilationContext: self compilationContext.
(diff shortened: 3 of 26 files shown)


MarcusDenker:
Only failing test is 

- SpOptionListPresenterTest>>testDoAcceptIsNotExecutedWhenCancelingDialog

Which is not related to the change

MarcusDenker:
Nice speedup for #evaluate:

```
[ Smalltalk compiler evaluate: '1+2' ] bench

before:  "'4431.741 per second'"
after:  "'28202.360 per second'"
```

src/AST-Core/ASTCacheMissStrategy.class.st
```diff
@@ -16,5 +16,5 @@ Class {
 
 { #category : #accessing }
 ASTCacheMissStrategy >> getASTFor: aCompiledMethod [
-	^ aCompiledMethod parseTree doSemanticAnalysisIn: aCompiledMethod methodClass
+	^ aCompiledMethod parseTree
```
dionisiydk, line 19: With little fix in #parseTree method (see below) the extra semantic analysis is not needed any more.

src/DrTests-CommentsToTests-Tests/CommentsToTestsTest.class.st
```diff
@@ -28,7 +28,7 @@ CommentsToTestsTest >> testCommentWithSyntaxError [
 	docComment := thisContext method ast pharoDocCommentNodes first.
 	commentTestCase := CommentTestCase for: docComment.
 	
-	self should: [commentTestCase testIt] raise: SyntaxErrorNotification
+	self should: [commentTestCase testIt] raise: RuntimeSyntaxError
```
dionisiydk, line 31: This difference is from the previous reformatting logic of ast methods. There were two stages:
1. initial parsing of source code in faulty mode
2. reformatting and parsing again with normal mode which causes syntax notification.

Now the second step is removed and therefore only runtime syntax error can be thrown.

src/OpalCompiler-Core/OpalCompiler.class.st
```diff
@@ -259,17 +259,7 @@ OpalCompiler >> compile: textOrString [
 { #category : #private }
 OpalCompiler >> compileDoItFromAST [
```
dionisiydk, line 260: This is the main change to remove the header and reformatting

MarcusDenker:
- ClassHierarchyPrinterTest>>testOnlyRBASTNodes needs to be update for the DoItMethodNode

src/DrTests-CommentsToTests-Tests/CommentsToTestsTest.class.st
MarcusDenker: I always wanted to change the doc comments to store the AST, not the source, but that is for another PR of course

PR #11474, opened 2022-07-25
fix the top and bottom of the rectangle of text input region so that …

tomooda:
…the candidate window can avoid the text line at the caret.


Ducasse:
tx!!!!

Issue #11425, opened 2022-06-30
#skipOnPharoCITestingEnvironment seems to not work on WIN and MAC


MarcusDenker:
**Bug description**
testJobIsFinishedWhenWaitingMoreThanWorkBlockDuration is failing on win even thought it does 

"	self skipOnPharoCITestingEnvironment."

This is related to environment variables:

```
onPharoCITestingEnvironment
	
	^ Smalltalk os environment at: 'PHARO_CI_TESTING_ENVIRONMENT' ifPresent: [ true ] ifAbsent: [ false ]

```

MarcusDenker:
on MAC, too:

SpJobListPresenterTest>>#testJobIsFinishedWhenWaitingMoreThanWorkBlockDuration

was failing even though it is skipped

MarcusDenker:
What is odd: if skipOnPharoCITestingEnvironment would not work, we should see much more tests failing.

MarcusDenker:
I checked the CI: the variable was not defined globally, I added it back

MarcusDenker:
Still not working correctly

MarcusDenker:
After adding it to the global config had no effect, I added the environment variable to all the CI machine runner setups

MarcusDenker:
The test testJobIsFinishedWhenWaitingMoreThanWorkBlockDuration does not have #skipOnPharoCITestingEnvironment...

so it seems the mechanism works, but some methods miss the tag.

Issue #11368, opened 2022-06-15
Failing on CI (random): CoNarrowHistoryFetcherTest>>testNarrowingReturnsSameElementsThatCallingDirectly


MarcusDenker:
**Bug description**
CoNarrowHistoryFetcherTest>>testNarrowingReturnsSameElementsThatCallingDirectly fails on the CI sometimes.

We though that the collection used (which was weak) was the problem (see https://github.com/pharo-project/pharo/pull/11355) but it now happened again.

MarcusDenker:
```
Error
Given Collections do not match. Got   left := #(#asLocalStringYMDHM #asLinkPrepend: #asLiteralNode #asLayoutFrame #asLocal #asLGitExternalString #asLink #asLowercase #asLargerPowerOfTwo #asLocalClassScope).  instead of   right :=#(#asLocalStringYMDHM #asLinkPrepend: #asLocalStringYMDHM #asLinkPrepend: #asLiteralNode #asLayoutFrame #asLocal #asLGitExternalString #asLink #asLowercase). Got 2 additional element(s) in the left collection:  #(#asLargerPowerOfTwo #asLocalClassScope)
Stacktrace
TestFailure
Given Collections do not match. Got 
	left := #(#asLocalStringYMDHM #asLinkPrepend: #asLiteralNode #asLayoutFrame #asLocal #asLGitExternalString #asLink #asLowercase #asLargerPowerOfTwo #asLocalClassScope).
 instead of 	 right :=#(#asLocalStringYMDHM #asLinkPrepend: #asLocalStringYMDHM #asLinkPrepend: #asLiteralNode #asLayoutFrame #asLocal #asLGitExternalString #asLink #asLowercase).
Got 2 additional element(s) in the left collection: 	#(#asLargerPowerOfTwo #asLocalClassScope)
CoNarrowHistoryFetcherTest(TestAsserter)>>assert:description:resumable:
CoNarrowHistoryFetcherTest(TestAsserter)>>assert:description:
CoNarrowHistoryFetcherTest(TestAsserter)>>assertCollection:equals:
CoNarrowHistoryFetcherTest>>testNarrowingReturnsSameElementsThatCallingDirectly
CoNarrowHistoryFetcherTest(TestCase)>>performTest
```

MarcusDenker:
Failing tests not related:

- [CI] Failing test: ProperMethodCategorizationTest>>testNoUncategorizedMethods #11669
[...]

PR #11482, opened 2022-07-27
[Compiler] Stateless OpalCompiler

dionisiydk:
1. PR makes OpalCompiler stateless (with temp state deprecation):

With the recent changes to the compiler it is now clear that we can easily remove #source and #ast variables and reuse the compiler instance for parsing/compilation of multiple sources. This way we can get rid of CompilationContext moving all its state to the compiler.  So that we will have logical reference from the method/AST to the producer compiler instance (currently it is the reference to the CompilationContext). Then many tools can be simplified by reusing a same compiler instance for own needs (highlighting for example). 

 Here PR performs main changes to the users of #compiler: 

   - Variables #source and #ast are removed (source is temporally deprecated for integration). 
   - API is changed to be #compile:/parse:/#evaluate: with source code as a parameter:

```Smalltalk
OpalCompiler new noPattern: true; compile: '1+2'.
```
Versus old style:
```Smalltalk
OpalCompiler new source: '1+2'; noPattern: true; compile.
```
The new style API was always there but it is now implemented without instance variables #source and #ast. 
Most of the changes in PR is such a rewrite from old style to new one.

   - #source:, #compile, #evaluate methods are deprecated (will be removed after NewTools and Spec fixes).
   - All users of compiler are adopted accordingly

The deprecation of #source variable is required to get safe independent adoption of NewTools and Spec. After that the variables will be removed.

The inline of CompilerContext will be done in separate PR.

2. Environment parameter is moved from CompilationContext into OCSemanticScope as #targetEnvironment which uses the environment of #targetClass by default. 

   - it makes the #environment configuration really working. Previously it did nothing. See the changes in #testCompileWithEnvironment. 

3. Introduce compilation option #optionEmbeddAST to remove #compileDoItFromAST method. Now #doIt compilation simply enables #optionEmbeddAST option (just as #optionEmbeddSources option) and a scope indirection is not needed anymore. See #generateMethod.


src/AST-Core/CompiledMethod.extension.st
```diff
@@ -25,10 +25,9 @@ CompiledMethod >> parseTree [
 	"returns an AST for this method, do not cache it. (see #ast for the cached alternative)"
 	| ast |
 	ast := self methodClass compiler 
-		source: self sourceCode;
 		failBlock: [^ self decompile ];
 		class: self methodClass;
-		parse.
+		parse: self sourceCode.
```
dionisiydk, line 30: Such move of #source: argument is the most of the changes here to remove the state from the compiler

src/OpalCompiler-Core/Behavior.extension.st
```diff
@@ -70,6 +68,14 @@ Behavior >> evaluate: aString [
 	^self compiler evaluate: aString
 ]
 
+{ #category : #'*OpalCompiler-Core' }
+Behavior >> lookupLocalVar: aName [
+	^ self classLayout 
+		resolveSlot: aName asSymbol 
+		ifFound: [:var | var] 
+		ifNone: [ self innerBindingOf: aName ]
```
dionisiydk, line 76: The local lookup is introduced to find any variable defined in the class without the full environment lookup. It is required to allow the override of an #environment of a class during a compilation. 

We should really review and unify all current lookup methods: #lookupVar:#lookupLocalVar:/#bindingOf:/#innerBindingOf:.

src/OpalCompiler-Core/CompilationContext.class.st
```diff
@@ -120,6 +119,16 @@ CompilationContext class >> optionCleanBlockClosure: aBoolean [
 	^ self writeDefaultOption: #optionCleanBlockClosure value: aBoolean
 ]
 
+{ #category : #'options - settings API' }
+CompilationContext class >> optionEmbeddAST [
+	^ self readDefaultOption: #optionEmbeddAST
+]
```
dionisiydk, line 135: Here is new option #optionEmbeddAST

src/OpalCompiler-Core/CompilationContext.class.st
```diff
@@ -488,7 +498,7 @@ CompilationContext >> noPattern: anObject [
 	anObject ifTrue: [
 		semanticScope := semanticScope asDoItScope.
 		"when we compile an expression, embedd sources, as the resulting method will never be installed"
-		self parseOptions: #(+ #optionEmbeddSources).
+		self parseOptions: #(+ #optionEmbeddSources + #optionEmbeddAST).
```
dionisiydk, line 522: And #optionEmbeddAST is now default option for doIts

src/OpalCompiler-Core/RBMethodNode.extension.st
```diff
@@ -85,11 +85,14 @@ RBMethodNode >> generateIR [
...
 RBMethodNode >> generateMethod [
-	| methodTrailer |
+	| methodTrailer newMethod |
 	methodTrailer := self compilationContext optionEmbeddSources
 			ifTrue: [ CompiledMethodTrailer new sourceCode: source ]
 			ifFalse: [self compilationContext compiledMethodTrailer ].
-	^self generate: methodTrailer
+	newMethod := self generate: methodTrailer.
+	self compilationContext optionEmbeddAST ifTrue: [ 
+		 newMethod propertyAt: #ast put: self ].
```
dionisiydk, line 100: And this is where optionEmbeddAST is used

src/OpalCompiler-Tests/OCEnvironmentScopeTest.class.st
```diff
@@ -10,22 +10,26 @@ Class {
 { #category : #'tests - environment' }
 OCEnvironmentScopeTest >> testCompileWithEnvironment [
 	"testing compiling with environments other then Smalltalk globals. Note: the class in the environment has to point to the environment. And we have to compile for a class in the environment"
-
-
 	| environment method return | 
 	environment := SystemDictionary new.
 	environment at: #MyClass put: Point copy.
-	(environment at: #MyClass) environment: environment.
+	environment at: #MyClass2 put: Association.
```
dionisiydk, line 16: This test demonstrates the environment override. In the current image it does not work.

MarcusDenker:
failing tests not related

guillep:
What's the next step here?
I see many merge conflicts.

dionisiydk:
> What's the next step here? I see many merge conflicts.

I am waiting @MarcusDenker review for this idea.

Conflicts are resolved

dionisiydk:
conflicts were badly fixed . I will revisit it. But it does not prevent the review

MarcusDenker:
The PR https://github.com/pharo-project/pharo/pull/13152 now adds some more state.. (see discussion)

 I need to think about this more (stateless compiler, CompilationContext is for sure now holding on to too much state...)

privat:
I added the cleaning of the state in CompilationContext and OpalCompiler to #12883, I will think about that.

Ducasse:
So what is the status? I have the impression that the issue is not going to be integratable.

tesonep:
Stale PR

PR #11483, opened 2022-07-28
Replace CommandTestRunner by a NewCommandLineRunner

aboubacardiawara:
Replacing :
    - CommandLineTestRunner 
    - VTermTestRunner
By NewCommandLineRunner.
This One covers their behaviors by using reporters (XmlReporter, StdReporter, XTermReporter)


aboubacardiawara:
The Bootstrap-64 step failed due to this error:
```
NewUndeclaredWarning: TestCommandLineHandler>>defaultReporter (StdReporter is Undeclared)
Instance of UndeclaredVariable did not understand #asExportedLiteral
```
The class already exists, and i tried many change and each times it tells another Object is undeclared.

Do you have any idea of the cause ?

jecisc:
@tesonep This PR is more than 3 years old. Should we keep it?

PR #11487, opened 2022-07-29
[Quick review] fix #11265: Underscore in class name breaks text completion

### LINKED ISSUE #11265: Underscore in class name breaks text completion (2)
(https://github.com/pharo-project/pharo/issues/10986 is fixed, this is another issue)

**Bug description**
Text completion doesn't show up after "_"

**To Reproduce**
Create a class named My_Class

Before "_", no issue

Once you type "_" no entries are showing up

Example with another character: (expected behavior)

**Version information:**
Windows 10 
Pharo 10.0.0
Build information: Pharo-10.0.0+build.519.sha.180dea8d1995a5199e3c7f6ffcae4b9880ffad77 (64 Bit)

hogoww:
fix #11265 + a small test.
We tried to add a test to check that there were results, however, it seems untested currently, and we didn't add required accesses.
We preferred to keep the fix small, and test that the behavior was as expected.

### DIFF
--- a/src/NECompletion-Tests/CompletionEngineTest.class.st
+++ b/src/NECompletion-Tests/CompletionEngineTest.class.st
@@ -48,7 +48,10 @@ CompletionEngineTest >> keyboardEventFor: char useCommand: command [
 		charCode: char asciiValue
 		hand: nil 
 		stamp: Time now.
-	event key: (KeyboardKey fromCharacter: char).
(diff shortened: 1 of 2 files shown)


hogoww:
Failing test case look unrelated

MarcusDenker:
failing tests not related

Issue #11263, opened 2022-05-22
bytecode and symbolicBytecodes: would be good to be consistent


Ducasse:
```
(Rectangle >> #width) bytecode
>>> #[1 126 0 126 97 92]
```

```
(Rectangle >> #width) symbolicBytecodes  
	25 <01> pushRcvr: 1 
	26 <7E> send: x 
	27 <00> pushRcvr: 0 
	28 <7E> send: x 
	29 <61> send: - 
	30 <5C> returnTop
```
It would be nice to have bytecodes and symbolicBytecodes. 

In addition it would be nice to have a more verbose verbose 

```
(Rectangle >> #width) symbolicBytecodes  
	25 <01> pushRcvr: 1  (corner)
	26 <7E> send: x 
	27 <00> pushRcvr: 0 (origin)
	28 <7E> send: x 
	29 <61> send: - 
	30 <5C> returnTop
```

hogoww:
About consistency, is it only the missing 's' at bytecode (=> PR is here #11488)?
Or is it also about decimal vs hexa representation?
If so, would you want to show both in hexa, or both in decimal?

Looking at the printing of IV now.

MarcusDenker:
main topic has been done. For printing the temp names: the problem is that mapping offsets to names for temps requires the compiler and AST and all the infrastructure (temps stored in temp vectors vs not, copies temps getting different offsets in different closures...)

This would complexify the whole implementation a lot and would mean that the package that SymbolicBytecode is in now requires the compiler (or even more to make it pluggable).

Consider that every use in practive is done via the inspector:

![bytecode](https://user-images.githubusercontent.com/413057/206734657-1b93f4ff-c883-4b3b-b9c9-daf4f7c6f753.jpg)

Where when you select the temp read bytecode, it will hightlight exactly the variable in the source.

PR #11491, opened 2022-07-29
Symbolic bytecode improvement

hogoww:
As suggested #11263 , improvement of the printing of symbolic bytecodes with variables names for temps, remote temps, & instances variables

### DIFF
--- a/src/Kernel/CompiledMethod.class.st
+++ b/src/Kernel/CompiledMethod.class.st
@@ -1116,12 +1116,6 @@ CompiledMethod >> tags [
 	^{protocol}
 ]
 
-{ #category : #'source code management' }
-CompiledMethod >> tempNames [
-	"on the level of the compiled method, tempNames includes argument names"
-	^self ast argumentNames, self ast temporaryNames
-]
-
 { #category : #accessing }
 CompiledMethod >> trailer [
 	"Answer the receiver's trailer"
--- a/src/Kernel/CompiledMethod.class.st
+++ b/src/Kernel/CompiledMethod.class.st
@@ -1116,6 +1116,12 @@ CompiledMethod >> tags [
 	^{protocol}
 ]
 
+{ #category : #'source code management' }
+CompiledMethod >> tempNames [
+	"on the level of the compiled method, tempNames includes argument names"
+	^self ast argumentNames, self ast temporaryNames
+]
+
 { #category : #accessing }
 CompiledMethod >> trailer [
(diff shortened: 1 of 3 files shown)


src/Kernel/CompiledMethod.class.st
```diff
@@ -1116,12 +1116,6 @@ CompiledMethod >> tags [
 	^{protocol}
 ]
 
-{ #category : #'source code management' }
```
Ducasse: Can you explain why tempNames are removed?

hogoww: It was not removed, it was pulled up in CompiledCode.
This is because compiled blocks may also use it  for certain purpose such as tooling !
Such as in this case

hogoww: However this is worth a fix in the comment

Ducasse:
Can you check because none of the builds is building and with the same error not related to git

hogoww:
Absolutely no idea.
I'll try to redo it sometime next week.

Ducasse:
NewUndeclaredWarning: CompiledCodeTest>>method (test is Undeclared)
NewUndeclaredWarning: CompiledCodeTest>>method (test is Undeclared)
[ERROR] 2022-08-12 22:57:29.000 growOldSpaceByAtLeast (Replacement/generated/64/vm/src/gcc3x-cointerp.c:48922):Required space is bigger than the headroom. Could not allocate


this error is strange.

guillep:
There is a consistent error in all builds:

```
Syntax Error on line 3: 'End of statement expected'

===================================================

1: UndefinedMethod arg1

2: 

3: 	^ arg1 + arg2

```

So this seems related...

hogoww:
> There is a consistent error in all builds:
> 
> ```
> Syntax Error on line 3: 'End of statement expected'
> 
> ===================================================
> 
> 1: UndefinedMethod arg1
> 
> 2: 
> 
> 3: 	^ arg1 + arg2
> ```
> 
> So this seems related...

Clearly, but I haven't touched any code using a **+**
That\'s why I want to redo it, I'm hopping it is  it might be a side effect that randomly occurred.

guillep:
Looking at the code, I wonder if this has something to do with decompilation.

It looks like this is a method with two args that are not properly decompiled. But the symbolic bytecode should have nothing to do with that... Unless the symbolic bytecode is used during decompilation...

PR #11492, opened 2022-07-29
Solving Printing infinite loop and crashing image in Printing of Array with self Array in it

DanielCamSan:
The problem was that when we are trying to print an Array the image crash and it was that it entering in an infinite loop so now is refering to the printing as the same way than OrderedCollection

### DIFF
--- a/src/Collections-Sequenceable/Array.class.st
+++ b/src/Collections-Sequenceable/Array.class.st
@@ -306,9 +306,16 @@ Array >> printAsSelfEvaluatingFormOn: aStream [
 
 { #category : #printing }
 Array >> printOn: aStream [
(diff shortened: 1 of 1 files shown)


svenvc:
But this only checks a very simple case (one level deep). You can create self-referring structures that are much deeper before they cycle. No ?

jordanmontt:
I agree with @svenvc. This is not a correct solution.
We need to be able to detect loops no matter how deep they are.

Actually, we need to fix the `printOn:` method in Collection to handle loops and that should fix Array `printOn:` too.

MarcusDenker:
fixed by merging https://github.com/pharo-project/pharo/pull/11916

PR #11493, opened 2022-07-29
[improvement] add microsecond to API

hogoww:
Useful for people using microsecondsToRun:
The naming microSecond or microsecond is inconsistent to, I took one randomly.

This seems to be an oversight that is just missing from the API.

### DIFF
--- a/src/Kernel-Tests/DurationTest.class.st
+++ b/src/Kernel-Tests/DurationTest.class.st
@@ -280,6 +280,17 @@ DurationTest >> testLessThan [
 ]
 
 { #category : #tests }
+DurationTest >> testMicroSeconds [
+	#(#(5 0 5000) #(1000005 1 5000) #(-5 0 -5000) #(-1000005 -1 -5000) #(1234000567 1234 567000) #(-1234000567 -1234 -567000))
+		do: [ :each | 
+			| duration | 
+			duration := Duration microSeconds: each first.
+			self assert: duration asSeconds equals: each second.
+			self assert: duration nanoSeconds equals: each third ]
+	"argument (milliseconds)        seconds nanoseconds"
+]
+
+{ #category : #tests }
 DurationTest >> testMilliSeconds [
 	#(#(5 0 5000000) #(1005 1 5000000) #(-5 0 -5000000) #(-1005 -1 -5000000) #(1234567 1234 567000000) #(-1234567 -1234 -567000000))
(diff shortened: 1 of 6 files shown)


hogoww:
All the failing test cases are passing on my local machine.
Will try restart the CI..

hogoww:
Tests cases are still indeed passing locally but not in the CI.
Any comment welcome, as I am currently out of ideas to fix this.

MarcusDenker:
The test FluidClassDefinitionPrinterTest>>#testChronologyConstants fails because the order of the variables somehow changed.. but only when printing the class definition for the fluid case, the other printouts (non-fluid) sort the class vars alphabetically and have no problem.


we could just change the test to be:

```
testChronologyConstants

	self
		assert: (self forClass: ChronologyConstants)
		equals: 'SharedPool << #ChronologyConstants
	sharedVariables: { #SqueakEpoch . #SecondsInDay . #DayNames . #MonthNames . #SecondsInHour . #MicrosecondsInDay . #NanosInSecond . #SecondsInMinute . #NanosInMillisecond . #DaysInMonth . #NanosInMicroSecond . #HoursInDay . #MinutesInHour };
	tag: ''Chronology'';
	package: ''Kernel'''
```

hogoww:
> The test FluidClassDefinitionPrinterTest>>#testChronologyConstants fails because the order of the variables somehow changed.. but only when printing the class definition for the fluid case, the other printouts (non-fluid) sort the class vars alphabetically and have no problem.
> 
> we could just change the test to be:
> 
> ```
> testChronologyConstants
> 
> 	self
> 		assert: (self forClass: ChronologyConstants)
> 		equals: 'SharedPool << #ChronologyConstants
> 	sharedVariables: { #SqueakEpoch . #SecondsInDay . #DayNames . #MonthNames . #SecondsInHour . #MicrosecondsInDay . #NanosInSecond . #SecondsInMinute . #NanosInMillisecond . #DaysInMonth . #NanosInMicroSecond . #HoursInDay . #MinutesInHour };
> 	tag: ''Chronology'';
> 	package: ''Kernel'''
> ```

Already done, and they pass locally (see previous commit)

MarcusDenker:
but if I download the build artefact, the test fails

hogoww:
Good point.
I always forget we can do that.
My theory is that the formater may have changed the order.
I'll update when I can

PR #11498, opened 2022-08-02
DrTests: Rerunning test

aboubacardiawara:
Rerunning tests in DrTests (TestRunner plugin) didn't work.
It is fixed in this commit

.


Ducasse:
We decided with Aboubacar to close this PR so that he can produce one or two separated ones.

PR #11504, opened 2022-08-05
CoverageRunner deleted

aboubacardiawara:
There is already a coverage runner in DrTests (see DTTestCoveragePlugin) and the one removed can't do anything newer.
It is more interesting to improve DTTestCoveragePlugin.


Ducasse:
Yes!

PR #11508, opened 2022-08-07
[cleanup] remove empty test classes

hogoww:
Multiple test classes are currently empty in a new image.
This PR removes them.
There was also an empty test superclass.

Let me know if something may be an issue.


hogoww:
Failing test cases are related, I'll take a look and document them

hogoww:
One test case failed in the CI, unrelated.
I'll let better people than me judge whether this clean up makes sense.
I don't see much incentive of keeping empty test classes, but I'm sure I could find reasons.

PR #11509, opened 2022-08-08
Cleanup on The package TestRunner

aboubacardiawara:
The cleanup concerns the following point:
- category
- code formating
- classes reference


MarcusDenker:
failing tests not related

estebanlm:
I like it, but I would do two steps : 
1. integrate
2. remove old TestRunner
3. rename NewTestRunner, NewCommandLineRunner classes to just TestRunner and CommandLineRunner.

estebanlm (approved):

estebanlm:
also: CommandLineRunner needs to be renamed to  CommandLineTestRunner
and we need an UI

and now that I see, names are messed up since TestRunner is the current name of the UI (and your NewTestRunner is an executor).

estebanlm (changes requested):
we need to align names and API a little bit :)

Ducasse:
This is cool that you review this.

aboubacardiawara:
Thanks for your review. I will apply your suggested changes.

Ducasse:
I'm trying to find the code but I do not find it in the image in the latest P11.

aboubacardiawara:
> I'm trying to find the code but I do not find it in the image in the latest P11.

It can be found after cloning the Pharo project again in a loaded image. Before that I couldn't find it either

Ducasse:
Thanks!

PR #11516, opened 2022-08-12
SIMD bytecode support

nrainhart:
Adds missing support for extending the bytecode interpreter to handle SIMD bytecodes.


nrainhart:
The CI failure does not appear to be related to the changes in this PR

PR #11535, opened 2022-08-17
11533-Ring-RGMethodDefinitionisDoIt-missing

### LINKED ISSUE #11533: Ring: RGMethodDefinition>>#isDoIt missing
**Bug description**
Ring Methods can not be checked for #isDoIt

This leads to a DNU in the VersionBrowser when channging the view (e.g. change it to "Source").

As Ring never models DoIts, we can just add a method returning false.

MarcusDenker:
add #isDoIt so that RB methods can be used where compiledmethods are used.

fixes #11533


MarcusDenker:
test not related

PR #11542, opened 2022-08-18
Fix #11540: Selecting a string and pressing Implementors breaks in Calypso New Class pane

### LINKED ISSUE #11540: Selecting a string and pressing Implementors breaks in Calypso New Class pane
Capture 2022-08-18 at 09 27 27

guillep:
Use #selectedClassOrMetaClass instead of asking for the #editingMethod. Each tool will extract the correct class on demand


guillep:
Failure not related

PR #11545, opened 2022-08-19
Updating version of Iceberg to 2.0.7, for libgit 1.4.4 -- Backport to Pharo 10

tesonep:


tesonep:
Updating to use libgit 1.4.4 to solve problems with HTTPS

PR #11553, opened 2022-08-19
Stream should be declared as abstract class

### LINKED ISSUE #11552: Stream should be declared as abstract class
Class comment says it is abstract. but #isAbstract returns false

astares:
Fix #11552


astares:
Simple change but CI fails with libgit:

image

astares:
@MarcusDenker can you please restart CI for this one

MarcusDenker (approved):

PR #11554, opened 2022-08-19
Review if #scanToken: in RBScanner is still needed

### LINKED ISSUE #11543: Review if #scanToken: in RBScanner is still needed
The comment in RBScanner(class)>>on:  method recommends to use #scanTokens:

"Reader may consider scanTokens: as a better way to use the class."

image

but #scanToken: is only used in two tests:

image

I think we can deprecate and remove it and also remove the unnecessary part in the comment on RBScanner(class)>>on: then.

Side note: there is also a #scanToken: in MCScanner class as well, which I guess can get removed too

astares:
Fix #11543

- delete the two tests who use scanToken:
- remove unnecessary part in RBScanner(class)>>#on comment
- introduce a Deprecated11 package (also in Baseline)
- deprecate the two #scanToken: methods
- correctly categorize baseline: in "baselines" as in superclass to satisfy/fix lint rule


astares:
Unrelated CI trouble for Mac,  others are green

PR #11555, opened 2022-08-19
Optimize setting of initial position in PositionableStream>>on:

### LINKED ISSUE #11536: Optimize setting of initial position in PositionableStream>>on:
When using the class side #on: method on a positionable stream:

```Smalltalk
ReadStream on: '3+4'
```
the instance side method PositionableStream>>on: is called which sets the position to zero:

```Smalltalk
on: aCollection

	collection := aCollection.
	readLimit := aCollection size.
	position := 0.
	self reset
```

and additionally calls #reset which also sets the position to zero again for a second time in ReadStream>>reset

```Smalltalk
reset
	"Set the receiver's position to the beginning of the sequence of objects."

	position := 0
```

We could optimize this by only setting the position in the reset method and remove setting it already in PositionableStream>>on:

So I propose that PositionableStream>>on: just needs to look like:

```Smalltalk
on: aCollection

	collection := aCollection.
	readLimit := aCollection size.	
	self reset
```

astares:
Fix #11536

### DIFF
--- a/src/Collections-Streams/PositionableStream.class.st
+++ b/src/Collections-Streams/PositionableStream.class.st
@@ -441,7 +441,6 @@ PositionableStream >> on: aCollection [
 
 	collection := aCollection.
 	readLimit := aCollection size.
-	position := 0.
 	self reset
 ]
 


Ducasse:
Ok the reset method is doing the same just after. 
Now I understand.

MarcusDenker:
I will do a revert to check if this is the reason for the broken build: the CI failed for this with the same error message as we now have for the build

PR #11564, opened 2022-08-21
Fix case insensitive auto-completion setting

### LINKED ISSUE #11445: Fix that unticking the Case Sensitive checkbox does not provide case insensitive completions?
I have this unticked for case insensitive completions:

image

But in the Playground, the completion popup is still case sensitive?

image

Here I would expect that `maxFloat` would also be suggested, but it is not? Am I misunderstanding how this setting behaves? I have never seen it change anything so far.

image

**Version information:**
 - OS: macOS 
 - Version: 10.14.6
 - Pharo Version 11.0

**Expected development cost**
I am a Pharo novice by any means. I tried to fix this myself by patching in `asLowercase` methods in the `RB...Node` classes, but that does not do anything and is highly likely the incorrect place to change this.

```smalltalk
select: [ :each | each beginsWith: ((self completionToken: offset) asLowercase) ]
```

Maybe someone can point me in the right direction where to approach this? 

I found this pull request: https://github.com/pharo-project/pharo/pull/8895 but my code in Pharo 11.0 looks a bit different to that, so I do not know how to approach this.

fmqa:
Attempts to fix #11445 

Replaces hardcoded calls to `CoCaseSensitiveBeginsWithFilter>>#filterString` with the factory method call where relevant, also adds an overload to `#allClassNamesStartingWith` allowing case-insensitive retrieval of class names by prefix

### DIFF
--- a/src/HeuristicCompletion-Model/CoCompletionContext.class.st
+++ b/src/HeuristicCompletion-Model/CoCompletionContext.class.st
@@ -155,7 +155,7 @@ CoCompletionContext >> isScripting [
 { #category : #narrowing }
 CoCompletionContext >> narrowWith: aString [ 
 
-	self completion replaceFilterWith: (CoCaseSensitiveBeginsWithFilter filterString: aString)
+	self completion replaceFilterWith: (CoBeginsWithFilter caseSensitive: NECPreferences caseSensitive filterString: aString)
 ]
 
 { #category : #accessing }
--- a/src/HeuristicCompletion-Model/CoGlobalVariableFetcher.class.st
+++ b/src/HeuristicCompletion-Model/CoGlobalVariableFetcher.class.st
@@ -27,6 +27,7 @@ CoGlobalVariableFetcher >> entriesDo: aBlock [
 	self systemNavigation
 		allClassNamesStartingWith: filter completionString 
 		do: [ :e | aBlock value: (NECGlobalEntry contents: e node: astNode) ]
+		caseSensitive: (NECPreferences caseSensitive)
 ]
 
 { #category : #accessing }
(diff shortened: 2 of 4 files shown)


Ducasse:
I think that this is the way to go. I did not get time today to look at it.

hogoww:
Code looks good to me, could someone restart the CI to check the test cases please?
This CI failure looks unrelated to me

Ducasse:
We have many related tests broken. 
This looks strange 
 CoMockSystemNavigation >> #allClassNamesStartingWith:do:caseSensitive:
I restart the build to make sure that this is related.

fmqa:
I think the changeset contains a mistake. `self allClassNamesStartingWith: aString do: aBlock caseSensitive: false` should be `self allClassNamesStartingWith: aString do: aBlock caseSensitive: true`. I missed this when performing the delegated call. I will fix this, the tests should pass afterwards.

fmqa:
> IceGitHubError: API rate limit exceeded for 128.93.193.1. (But here's the good news: Authenticated requests get a higher rate limit. Check out the documentation for more details.)

CI fails due to this :(

Ducasse:
Yes this is a problem. Pablo explained me why the authentification is difficult to do but we will have to see. 
On our gitlab server it would be simpler but we want to push our visibility on github. So we will have to discuss this.

fmqa:
MNUs in test happen because the Mock class `CoMockSystemNavigation` needs to support the newly added message.
I will take care of this sometime today.

fmqa:
> SubclassResponsibility: MCPrivateHostableRepository class had the subclass responsibility to implement #basicDescription

More CI trouble here :/

fmqa:
Can we restart the CI here once more? There is some unrelated bootstrapping problem with Iceberg, I wonder if it's reproducible.

fmqa:
Same CI error also happens on [another PR](https://ci.inria.fr/pharo-ci-jenkins2/blue/organizations/jenkins/Test%20pending%20pull%20request%20and%20branch%20Pipeline/detail/PR-11655/1/pipeline)

fmqa:
CI issue tracked in #11657

MarcusDenker:
Failing tests not related:

Ducasse:
Looks ok to me.

PR #11583, opened 2022-08-22
Attempt to fix scrolling behavior in RubAbstractTextArea & friends

### LINKED ISSUE #10666: Iceberg's commit window has non-functional horizontal scroll bars
**Bug description**
When looking at a diff using iceberg's commit window, the horizontal scrollbar has incorrect bounds and does not allow for scrolling, therefore it's impossible to see the changes in long lines.

**To Reproduce**
1. Go to Iceberg
2. Click on: a repository where you'd like to do a commit.
3. Click on commit.
4. Select a change with a long line
5. Try to scroll horizontally to see the changes.

**Expected behavior**
Window should allow looking at all the code in the diff sections.

**Screenshots**
image

**Version information:**
 - OS: Windows
 - Version: 10
 - Pharo Version:
    - Pharo-9.0.0+build.1573.sha.0e09d756fc99383cbb498565200d9e5e9841ce11 (64 Bit)

**Expected development cost**
🤷‍♂️

fmqa:
Attempts to fix #10666

Adds a flag that, if enabled, automatically resizes the text area horizontally so that its width equals the approximate point size of the longest text line in the editing pane. The flag is enabled by default for `RubScrolledTextMorph`'s textArea, thus alleviating issues with scrolling in Iceberg & Epicea's changeset viewers.

### DIFF
--- a/src/Rubric/RubAbstractTextArea.class.st
+++ b/src/Rubric/RubAbstractTextArea.class.st
@@ -994,7 +994,7 @@ RubAbstractTextArea >> handleEdit: editBlock [
 	editBlock value.
 	"Fit to width on edit"
 	(self grow isNotNil & self grow) ifTrue: [
-		self width: (1 max: self maxLineWidth)
+		self width: (1 max: self maxLineWidth + self margins left + self margins right)
 	].
 	self selectionChanged	"Note new selection"
 ]
@@ -1328,7 +1328,7 @@ RubAbstractTextArea >> maxLineWidth [
 	w := 0.
 	self string linesDo: [ :line | w := w max: (self font widthOfString: line) ].
 	"check this later, not really sure about this calculation"
-	^ w * 6 // 5 + self margins left + self margins right
+	^ w * 6 // 5
 ]
 
 { #category : #accessing }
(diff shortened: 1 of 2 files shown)


MarcusDenker:
Failing test not related

lopezca:
Thank you @fmqa this fix is really helpful.

Ducasse:
Thanks for the test and feedback

src/Rubric/RubAbstractTextArea.class.st
```diff
@@ -991,6 +992,10 @@ RubAbstractTextArea >> handleEdit: editBlock [
 	"Ensure that changed areas get suitably redrawn"
 
 	editBlock value.
+	"Fit to width on edit"
+	(self grow isNotNil & self grow) ifTrue: [
```
Ducasse, line 996: I will integrate it but I would have preferred to have grow initialized to false and to avoid the isNil check.

Ducasse:
I will integrate it now and it would be good to produce another one where grow is initialized to false and like that we avoid having a nil or true or false around and we can remove the isNotNil check.

fmqa:
@Ducasse #11645 avoids the isNotNil message as per your recommendation

Ducasse:
Thanks :)!

PR #11585, opened 2022-08-24
Make explicit hostname and port work with bitbucket

guillep:
Make explicit hostname and port available to bitbucket in addition to gitlab.

- Refactor gitlab support to abstract it and make it work for bitbucket
- Add some tests
- [TODO] I moved some methods from iceberg extensions to these MC repos, we will need to remove the iceberg extensions for this to work

### DIFF
--- a/src/Metacello-Bitbucket/MCBitbucketRepository.class.st
+++ b/src/Metacello-Bitbucket/MCBitbucketRepository.class.st
@@ -1,6 +1,6 @@
 Class {
 	#name : #MCBitbucketRepository,
-	#superclass : #MCGitBasedNetworkRepository,
+	#superclass : #MCPrivateHostableRepository,
 	#category : #'Metacello-Bitbucket'
 }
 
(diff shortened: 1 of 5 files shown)


Ducasse:
so should we integrate it or do we have to wait?

guillep:
Let's wait, gimme a couple of hours, we are testing with a real server :)

labordep:
Thanks @guillep, this is working in my internal bitbucket repository and can I see the content of my project.
But I have another problem and I don't know if it's link to this evolution.

When I'm doing that :
```
Metacello new
     baseline: 'Cade'; 
     repository: 'bitbucket://my.super.server:foo/SMOCK/cade:master';
     load.

"Cade is my project, and SMOCK my repository of code"
```
My project is cloning localy in a loop, first loop is creating each repository folder in my PC but after there is another cloning steps and I have an error message : 

image

If I ignore this error and I continue the loading, the repository try to re-clone more and more... 

And my source code and packages was never loaded.

labordep:
The problem seems to be that
 - metacello tries first to clone using ssh and it fails
 - then it clones in https and configures the remote using https
 - and then after cloning, when loading the code it tries to get the repository by its ssh url, but it is not there
 - so it reclones it again and fails because the directory is already there

guillep:
I think this issue fixes the Metacello problem. The remaining problem lies in Iceberg. I'll open a separate issue for it.

PR #11596, opened 2022-08-25
[Pharo9] Use iceberg 2.0.5.1

guillep:
Iceberg 2.0.5.1 based on 2.0.5 which is pharo 9 compatible


guillep:
Ok, the failing tests are related to the following:

imagen

Very probably the tear down is not working properly?

PR #11600, opened 2022-08-25
Fix behavior of sum: on collections of doubles

VincentBlondeau:
When summing the values, sum: does operations that have a different order that just iterating on the original collection. This makes the order the same.
With this collection: #(33.33d0 33.33d0 33.34d0).
The current implementation is doing:
33.33d0 + 33.33d0 + 33.33d0 + 33.34d0 - 33.33d0 = 99.99999999d0
The fixed implementation is doing:
33.33d0 - 33.33d0 + 33.33d0 + 33.33d0 + 33.34d0 = 100d0


svenvc:
LGTM, but the diff is weird, it removed and adds identical methods

MarcusDenker:
failing test not related:

PR #11606, opened 2022-08-26
Revert "Optimize setting of initial position in PositionableStream>>on:"

### LINKED ISSUE #11605: [Image Build Pharo11] Bootrap broken since Build
**Bug description**
The CI can not build the image or run CI checks. The reason is that some PR that we merged broke the boostrap

It fails since Build 171: https://ci.inria.fr/pharo-ci-jenkins2/job/Test%20pending%20pull%20request%20and%20branch%20Pipeline/job/Pharo11/171/

Changes merged:

- Optimize setting of initial position in PositionableStream>>on:  #11555
- Removed empty abstractCompletionEngine class  #11562
- Improve #basicSize documentation and add tests   #11590
- Cleanup ReferenceFinder>>#searchIndicesIn: #11592

Maybe we merged one of those without waiting for a successful CI run?

MarcusDenker:
Reverts pharo-project/pharo#11555

Check if this is the reason for the broken build

### DIFF
--- a/src/Collections-Streams/PositionableStream.class.st
+++ b/src/Collections-Streams/PositionableStream.class.st
@@ -447,6 +447,7 @@ PositionableStream >> on: aCollection [
 
 	collection := aCollection.
 	readLimit := aCollection size.
+	position := 0.
 	self reset
 ]
 


MarcusDenker:
Success!

Ducasse:
This is really strange because the code is doing

```
   position := 0.
   self reset.

reset
   position := 0.

```
and the change just removed the first `position := 0`

astares:
@MarcusDenker @Ducasse 

Yes - really strange, as the code change looks easy and clear.  But streams are used often.
The image also froze when I committed the change from the image - so I changed it on file level to commit the PR.

If I read correctly today after the merge the image build froze with an adoptToNumberAndSend: or something like that - this is sometimes the case when the emergency debugger comes up. Maybe a strange side effect. Would be interesting to know more deeply what caused it during the bootstrap process.

Ducasse:
what was the stack or log when you did it at the image level?
Editing directly in GH is not a good idea.

astares:
There was no stack and log, just a freeze. Just do the change yourself in pharo and commit to a branch locally (which might work).
Then again load this branch with the change  from a fresh image.

I did not Edit on GH.  I changed on file Level locally and then pushed branch to GH to create PR.

Ducasse:
Yes this is the same. Local edit on the file system because system breaks are dangerous.

Ducasse:
There is probably an endless loop because the compiler/changes should be using a stream while compiling. Now this is super strange that your off changes got that impact. 
We will see. I should have see that the built was broken. I was not careful enough.

PR #11607, opened 2022-08-26
Categorize test classes. No modification

Ducasse:
Categorize test classes. No modification


Ducasse:
There is one related failure but I will fix it in another iteration

PR #11614, opened 2022-08-29
adding entries in the debug world menu to enable / disable all breakpoints in the image

### LINKED ISSUE #11599: [Feature Request][Debugger] Disable/Enable all debug breakpoints
Hi,

This would very cool to disable in one click all breakpoint but without lost them.
From a user point of view, a solution can be a toggle button in the main Pharo Debug menu (and also inside the debugger) to switch breakpoints on/off : 
image

This function exists in some IDE but most of the time user can be confused because he don't remember states of the activation (or not) : The debug menu can be display the state of breakpoints activation with a color or an icon when they are not active for the garanted that users doesn't believe that there are active. 
image

This features is very usefull when you debug, particularly graphical loops and i/o events.

Thanks.

Sorry I cannot set that this topic is related to "debugger" tag :(

adri09070:
Fix #11599 

I added entries in the debug world menu so that we can disable and tenable all breakpoints in the image.

This is something that could be done programatically but a shortcut is not too much for that as this is something that not everybody knows we can do and, sometimes, we regularly want to remove/put again breakpoints. And people that don't know do that manually

### DIFF
--- a/src/Reflectivity/Breakpoint.class.st
+++ b/src/Reflectivity/Breakpoint.class.st
@@ -88,13 +88,25 @@ Breakpoint class >> cleanUp [
 
 { #category : #'world menu' }
 Breakpoint class >> debugWorldMenuOn: aBuilder [
+
 	<worldMenu>
-	
 	(aBuilder item: #'Remove all Breakpoints')
 		parent: #Breakpoints;
 		order: 2;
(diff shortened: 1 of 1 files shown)


adri09070:
CI is red but when you look for détails, the CI is green...

MarcusDenker:
Failure not related:

PR #11615, opened 2022-08-29
Improving the speed of the compiler when compiling methods with a lot of literals

tesonep:
I have tested that a method with 32k literals is compiled 50 times faster. 
For small & common methods there is not impact.
I have replaced the implementation using an array with linear scan to use a specialized Dictionary that uses #literalEquals:.
It has not impact in small methods because before it was using an array and a Set to guarantee uniqueness, so the impact is similar.

I would backport it to P10.

### DIFF
--- a/src/Collections-Strings/Symbol.class.st
+++ b/src/Collections-Strings/Symbol.class.st
@@ -499,7 +499,7 @@ Symbol >> isUnary [
 { #category : #comparing }
 Symbol >> literalEqual: other [
 
-	^ self class == other class and: [self == other]
+	^ self == other
 ]
 
 { #category : #'system primitives' }
--- a/src/Collections-Strings/Symbol.class.st
+++ b/src/Collections-Strings/Symbol.class.st
@@ -496,6 +496,12 @@ Symbol >> isUnary [
 	^ self precedence = 1
 ]
 
+{ #category : #comparing }
+Symbol >> literalEqual: other [
+
+	^ self class == other class and: [self == other]
+]
+
 { #category : #'system primitives' }
 Symbol >> numArgs: n [
 	"Answer a string that can be used as a selector with n arguments.
--- a/src/OpalCompiler-Core/OCLiteralList.class.st
+++ b/src/OpalCompiler-Core/OCLiteralList.class.st
@@ -8,7 +8,6 @@ Class {
 	#superclass : #Object,
 	#instVars : [
 		'literalsDict',
-		'lastIndex',
 		'first'
 	],
 	#category : #'OpalCompiler-Core-Extras'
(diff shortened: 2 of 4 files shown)


src/Collections-Strings/Symbol.class.st
```diff
@@ -496,6 +496,12 @@ Symbol >> isUnary [
 	^ self precedence = 1
 ]
 
+{ #category : #comparing }
+Symbol >> literalEqual: other [
+
+	^ self class == other class and: [self == other]
```
guillep: why not just `self == other`? It is not possible that  `self == other ` but  `self class ~~ other class`

guillep (commented):
Super nice! :D

src/OpalCompiler-Core/OCLiteralList.class.st
```diff
@@ -1,46 +1,63 @@
...
+I am optimised to guarantee that literals all unique and have an unique index.
 "
 Class {
 	#name : #OCLiteralList,
-	#superclass : #OrderedCollection,
+	#superclass : #Object,
 	#instVars : [
-		'equalitySet'
+		'literalsDict',
+		'lastIndex',
```
dionisiydk: It seems #lastIndex is actually a size of #literalsDict. So this variable can be replaced

src/OpalCompiler-Core/OCLiteralList.class.st
```diff
@@ -1,46 +1,63 @@
 "
-Holds a unique ordered collection of literals
+Holds a unique ordered collection of literals.
+I uses an special dictionary inside that compares literals using #literalEquals message.
+I am optimised to guarantee that literals all unique and have an unique index.
 "
 Class {
 	#name : #OCLiteralList,
```
dionisiydk, line 7: I think OCLiteralList and OCLiteralsDictionary can be merged. So it can be even simpler.

dionisiydk (approved):

src/OpalCompiler-Core/OCLiteralList.class.st
tesonep: I didn't like to do that, because I want not to extend the Dictionary for implementing OCLiteralList. Before OCLiteralList was a subclass of OrderedCollection, so it has all the methods of OrderedCollection. However, from the intended use not all the API is valid. I am extending the Dictionary just to change the comparison (I could have used PluggableDictionary, but I got less performance with the use of blocks), but I try to keep it hidden from the clients of OCLiteralList.  To me, it is more important to have a clear API for the users. 

For example, the #add: method in Collections returns the added object, but I need to return the index of the newly added literal, if you see the old code it was performing two searches in the collection and this knowledge and complexity was pushed to the client of OCLiteralList, just to keep the compatibility with Collections. Also, breaking the compatibility with Collection is bad, so I decided to separate it from the Collection hierarchy.

src/Collections-Strings/Symbol.class.st
tesonep: Good catch.

MarcusDenker (approved):
Nice!

src/OpalCompiler-Core/OCLiteralList.class.st
tesonep: Thanks, good catch here.

astares:
Really nice! Would be interesting to see the speed effect this has on package/code loading (like loading Seaside)

MarcusDenker:
Failing test not related

PR #11620, opened 2022-08-31
Metacello should initialize also subclasses of classes defining initialize

sonibla:
Hello there

Metacello initializes loaded classes if they define initialize on class side, but not their subclasses
I think it's a bug (correct me if I'm wrong), so I tweaked the postload to also call initialize on subclasses

### DIFF
--- a/src/Monticello/MCMethodDefinition.class.st
+++ b/src/Monticello/MCMethodDefinition.class.st
@@ -284,15 +284,18 @@ MCMethodDefinition >> overridenMethodOrNil [
 
 { #category : #installing }
 MCMethodDefinition >> postloadOver: aDefinition [
+
 	super postloadOver: aDefinition.
-	self class initializersEnabled ifTrue: [
-		(self isInitializer
-			and: [ self actualClass isTrait not 
-					and: [ aDefinition isNil or: [ self source ~= aDefinition source ]]]) ifTrue: [
-				self actualClass instanceSide initialize ] ].
+	self class initializersEnabled ifTrue: [ 
+		(self isInitializer and: [ 
+			 self actualClass isTrait not and: [ 
+				 aDefinition isNil or: [ self source ~= aDefinition source ] ] ]) 
+			ifTrue: [ 
+				self actualClass instanceSide withAllSubclassesDo: [ :e | 
+					e initialize ] ] ].
 	"Postloading of FFI fields. This code will be called when loading FFI structures that are not by default in the image. This is NOT dead code."
-	self isExternalStructureFieldDefinition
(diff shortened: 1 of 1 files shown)


svenvc:
But won't the subclass be initialized when it is loaded itself ?

What is your use case ?

sonibla:
My use case:

AbstractClass defining initialize and a class instance variable
Subclass inheriting both

After loading them with metacello, only the abstract class is initialized

sonibla:
When metacello reaches postload, all instances of MCMethodDefinition are sent #postloadOver:, this method is responsible for initializing the class.
There is an instance of MCMethodDefinition for AbstractClass class>>#initialize, but there is no for the subclass (that makes sense because initialize is only defined in AbstractClass), but I also want the subclass to be initialized

sonibla:
And it seems my change is more radioactive than I thought, the buid failed when loading packages

astares:
> but I also want the subclass to be initialized

If you want/need the subclass to be initialized **then you need to provide/implement an explizit #initialize method**.

Just inheriting is not a criteria and not enough and it is good that this mechanism is working like this!

If a code loader would call #initialize on any subclass that just inherits this method could easily bring you in trouble.

As an example: if you would have a database framework providing an abstract superclass that is 
holding all (open) database connections in a class variable

```Smalltalk
Object subclass: #AbstractDatabaseManager
	instanceVariableNames: '' 
	classVariableNames: 'OpenConnections'
	package: 'Athens-Cairo-Tests'
```	

it might have initialized the connection pool/collection like:

```Smalltalk
initialize
    OpenConnections := OrderedCollection new
```	

Now anytime you would load a concrete subclass in custom code package it would (all the time a subclass
appear) reset the open connections. Which would be unwanted and to make things worse: if you would load
10 subclasses it would reinitialize it 10 times - which is also not good.

So changing this would introduce bad side effects: if you load an own custom package with a custom
subclass this would do a call on the superclass initialize again. As these class #initialize methods 
in the superclass often initialize class variables to a clean state (referencing important objects) it would badly
reinitialiize anytime a subclass gets loaded and might do dome unwanted freeing of objects that still need to
be referenced or other.

So calling only the ones with an explicit #initialize implementation  is the behavior ever since (not only in 
Metacello, but also loading from Changesets *.cs files or other fileouts *.st files) - in Squeak, Pharo, Cuis but 
also other Smalltalks:

When loading the classes who provide a class side #initialize method as an implementation/real method
will receive a call to it. This also makes it easier to understand, browse and find.	

So the current design is by intention and not a mistake in Metacello. If you need a call explicitly provide 
a class side #initialize method.

PR #11532, opened 2022-08-16
Creates Graphical UI for AndreasProfiler and TimeProfiler

carolahp:
Creating a graphical user interface to profile code using both AndreasProfiler and TimeProfiler.
Fixes [issue #10260 Integrate Profiler improvements](https://github.com/pharo-project/pharo/issues/10260)

PR #11621, opened 2022-08-31
Test profiler UI

carolahp:
Fixing problems in PR #11532

### DIFF
--- a/src/BaselineOfIDE/BaselineOfIDE.class.st
+++ b/src/BaselineOfIDE/BaselineOfIDE.class.st
@@ -94,7 +94,7 @@ BaselineOfIDE >> additionalInitialization [
 	self makeCalypsoDefaultBrowser.
 ]
 
-{ #category : #baseline }
+{ #category : #baselines }
 BaselineOfIDE >> baseline: spec [
 	<baseline>
 	| repository |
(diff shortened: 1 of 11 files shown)


MarcusDenker:
Again, this is submitted against Pharo10, but normally new features are added to Pharo11 and backported if very important.

MarcusDenker:
(I added the needs more work tag so we do not accidentially merge into Pharo10 without knowing that this is really what we want)

astares:
@MarcusDenker : yes and fully agree with you on this, new features should go to Pharo 11 and not Pharo 10. 

The problem: a part of that code for the UI has already been merged as #11532 into Pharo10 as I already mentioned a few days back on:

https://github.com/pharo-project/pharo/pull/11532

but this merge into Pharo 10 (!) had several problems:
 - it was merged without a review / approval
 - it was merged into Pharo10 although this is a stable and already released version
 - it was providing changed code but also new packages that have not been added to the Baseline and therefore was in the source but not in the final image 

I'm sure @Ducasse merged with good will - but I guess the ESUG day had been too stressful to find the time to have a proper look on what had been merged. And @carolahp now just tried to move forward and fix it. Looks like the idea is to complete it inside of Pharo 10 with even more PRs - but IMHO this is not a good idea.

I would clearly not include this into Pharo 10 and roll back even the incomplete #11532 for Pharo10 from last week. Many people have invested lots of work and time to provide and release a stable Pharo 10 and now with merging PRs we endanger it by allowing further new development in it??? 

I think would be better to roll back the merge of #11532.

We should not surprise Pharo users who want to rely on a stable finished release with new features and new code additions
and put their applications and builds with it into a risk.

Rule to follow is very simple:
New developments should only go to the new development branch/new not yet released iteration version (which is Pharo 11 nowadays) and only if there is a critical fix or a safe performance improvement it could be backported to Pharo 10 if there is a wide interest for a backport fix. 

If Pharo users would not be able to rely on proper release management it would be a tough ride for them and endanger the use of Pharo to be considered for serious work...

MarcusDenker:
This is the Pharo10 PR. I think we can close it, as we reverted the PR

PR #11644, opened 2022-09-04
FTTableMorph: Allow expansion/collapse of nodes using arrow keys

### LINKED ISSUE #11643: Enhance package/class trees in system browser to allow expanding/collapsing groups using the keyboard
The package tree & class hierarchy in the system browser is only partially navigable using the keyboard.
Moving the selected row up/down using the arrow keys is possible. However, it is not possible to expand/collapse parent nodes using the keyboard.

**Describe the request**
Improve keyboard navigation by allowing expansion/collapse of rows using arrow keys. This mirrors the behavior seen in other common GUI environments.

**Expected behavior**
GIVEN I am running the Pharo IDE
AND I start the System Browser (i.e. Calypso)
AND the package navigation pane on the left is focused
AND I navigate using the up/down arrows to a *collapsed* node with children
AND I press the *right arrow* key
THEN the node is expanded and its immediate children are shown

GIVEN I am running the Pharo IDE
AND I start the System Browser (i.e. Calypso)
AND the package navigation pane on the left is focused
AND I navigate using the up/down arrows to an *expanded* node with children
AND I press the *left arrow* key
THEN the node is collapsed and all of its children are hidden

**Expected development cost**
Minor. Can be done by adding some additional announcement types to `FTTableMorph`

**Version information:**
 - OS: Ubuntu x64
 - Version: 22.04.1
 - Pharo 10

**Additional context**
Related to #11618, where we attempt to complete the missing pieces of Pharo's keyboard navigation support

fmqa:
Fixes  #11643

Implementation strategy:

* Adds an additional announcement class `FTSelectionRejected` that is used to signal attempted out-of-bounds selections in a FTTreeMorph
* FTTreeMorph announces `FTSelectionRejected` whenever an out-of-bounds selection is attempted
* ClyQueryViewMorph registers a handler for this announcement
* The handler checks the direction vector, if it is to the right, expand the node. Otherwise collapse it.

### DIFF
--- a/src/Calypso-Browser/ClyNoChildren.class.st
+++ b/src/Calypso-Browser/ClyNoChildren.class.st
@@ -1,0 +1,32 @@
+"
+I am ClyNoChildren, an exception indicating that an operation on a parent in a parent-child structure failed due to the absence of child elements. 
+"
+Class {
+	#name : #ClyNoChildren,
+	#superclass : #Error,
+	#instVars : [
+		'parent'
+	],
+	#category : #'Calypso-Browser-Exceptions'
+}
+
+{ #category : #signalling }
+ClyNoChildren class >> signalFor: aParentObject [
+	"Create and signal a ClyNotFound exception for aParentObject in the default receiver."
+			
+	^ self new
+		parent: aParentObject;
+		signal
+]
+
+{ #category : #accessing }
+ClyNoChildren >> parent [
+
+	^ parent
+]
+
+{ #category : #accessing }
+ClyNoChildren >> parent: anObject [
+
+	parent := anObject
+]
--- a/src/Calypso-Browser/ClyCollapsedDataSource.class.st
+++ b/src/Calypso-Browser/ClyCollapsedDataSource.class.st
@@ -229,7 +229,7 @@ ClyCollapsedDataSource >> retrieveChildrenOf: aDataSourceItem [
 	| childQuery childDataSource |
 	childQuery := queryView 
 		queryToExpand: aDataSourceItem 
-		ifAbsent: [self error: 'cant retrieve children from ', aDataSourceItem printString].
+		ifAbsent: [ClyNoChildren signalFor: aDataSourceItem].
 	
 	childDataSource := ClyCollapsedDataSource on: childQuery.
 	childDataSource openOn: queryView.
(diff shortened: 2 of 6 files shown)


Ducasse:
@estebanlm can you have a look? Tx

NSUSpray:
It works in the panes of packages, classes and protocols, with any possible switch positions on the middle separator (AllPackages/ScopedView etc.). But when I press to the Right in the *methods* pane, an error is displayed (see attached image):
> Error: cant retrieve children from a ClyDataSourceItem(open)

Windows 10 x64

изображение

fmqa:
There is some problem with `ClyDataSourceItem>>#hasChildren` - it does not give the correct answer in all cases. Especially in the package pane it seems to answer `true` on leaf nodes.

For now, maybe we should address this enhancement by raising a more specific exception type in the `ClyCollapsedDataSource` (which might be a better idea than using `Object>>#error:` anyway).

I will open a separate issue ticket for `ClyDataSourceItem>>#hasChildren` once I understand the problem with it a bit better.

MarcusDenker:
errors not related

dionisiydk:
> There is some problem with `ClyDataSourceItem>>#hasChildren` - it does not give the correct answer in all cases. Especially in the package pane it seems to answer `true` on leaf nodes.

You should use #supportsExpansion instead to detect if the item can be expanded at all:
```Smalltalk
item supportsExpansion ifTrue: [
   item isExpanded ifFalse: [ item expand ] ]
```

src/Calypso-Browser/ClyNoChildren.class.st
```diff
@@ -0,0 +1,32 @@
+"
+I am ClyNoChildren, an exception indicating that an operation on a parent in a parent-child structure failed due to the absence of child elements. 
+"
+Class {
+	#name : #ClyNoChildren,
```
dionisiydk: Please add a suffix Error to the name: ClyNoChildrenError

dionisiydk (approved):
I removed my comment. You are right about the problems with expansion logic

estebanlm (approved):

src/Calypso-Browser/ClyNoChildren.class.st
fmqa: Done

PR #11647, opened 2022-09-04
FTTableMorph: Pass some keyboard events to the filter field if it is focused

### LINKED ISSUE #11646: System browser's package filter does not respond to keyboard events despite having keyboard focus
**Bug description**
The system browser has a textual input field at the bottom that allows for filtering by package name.
However, this field does not actually respond to the common editing keys (left arrow, right arrow, home, end) 

**To Reproduce**

1. Open the system browser using the world menu or CTRL+O+B
2. Click on the filter text input at the bottom of the package browser
3. Enter some text e.g. 'abc'
4. Press the left arrow or home key

**Expected behavior**
The filter text input's cursor position responds to the keystroke, changing the cursor position 

**Screenshots**
Filter

**Version information:**
 - OS: Ubuntu x64
 - Version: 22.04.1
 - Pharo 10

**Expected development cost**
Minor. In `FTTableMorph`, the focus of needs to be checked and the event needs to be passed to the filter control. 

**Additional context**
This relates to `FTTableMorph`, more specifically the keybinding configuration in `FTTableMorph >> initializeKeyBindings`

fmqa:
Fixes #11646

This triggers actions on `FTTableMorph` or on its `RubTextFieldMorph` child depending on whether the filter input has keyboard focus or not.

### DIFF
--- a/src/Calypso-Browser/FTTableMorph.extension.st
+++ b/src/Calypso-Browser/FTTableMorph.extension.st
@@ -22,6 +22,13 @@ FTTableMorph >> filterField [
 ]
 
 { #category : #'*Calypso-Browser' }
+FTTableMorph >> filterFieldHasKeyboardFocus [
+	| field |
+	field := self filterField.
+	^ field isNotNil and: [ field hasKeyboardFocus ]
+]
+
+{ #category : #'*Calypso-Browser' }
 FTTableMorph >> filterString [
 	(function isKindOf: FTFilterFunction) ifFalse: [ ^'' ]. 
 	^function pattern ifNil: [ '' ]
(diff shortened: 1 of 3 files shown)


fmqa:
I'm not sure if there's a better way to do event forwarding.

I tried to forward the event as-is using `self filterField handleEvent: event` but that is ignored due to the event's `wasHandled` flag being set to true.

As an alternative, I tried `event wasHandled: false. self filterField handleEvent: event` but that locks up the event loop.

Therefore, this PR uses explicit forwarding. But I would not mind changing it if there's a better/shorter way to achieve the desired result.

PR #11651, opened 2022-09-05
Adding an API in OCBytecodeToASTCache, RBMethodNode and RBBlockNode to map AST nodes to a pc range.

adri09070:
`OCBytecodeToASTCache` only provides a way to easily map bytecode offsets to AST nodes but it doesn't allow to do it the other way around.

I implemented an API in this class to do so:

- `OCBytecodeToASTCache>>#pcsForNode:`that returns a collection of pcs of bytecodes that are mapped to the argument node.
-  `OCBytecodeToASTCache>>#firstBcForNode:` that returns the first bytecode offset in the pc range mapped to the argument node, and nil if the argument node isn't mapped to any bytecode offset.
- `OCBytecodeToASTCache>>#lastBcForNode:` that returns the last bytecode offset in the pc range mapped to the argument node, and nil if the argument node isn't mapped to any bytecode offset.

This API is also in `RBBlockNode`and `RBMethodNode`to have an easier AST -> pc mapping, just like it already is for the pc -> AST mapping

### DIFF
--- a/src/OpalCompiler-Core/OCBytecodeToASTCache.class.st
+++ b/src/OpalCompiler-Core/OCBytecodeToASTCache.class.st
@@ -62,6 +62,12 @@ OCBytecodeToASTCache >> firstBcOffset [
 	^ firstBcOffset
 ]
 
+{ #category : #astAndAstMapping }
+OCBytecodeToASTCache >> firstBcOffsetForNode: aNode [
+
+	^ (self pcsForNode: aNode) at: 1 ifAbsent: nil
+]
+
 { #category : #initialization }
 OCBytecodeToASTCache >> generateForNode: aNode [
 	| methodOrBlockIr currentBcOffset |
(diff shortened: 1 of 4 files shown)


dionisiydk:
Side comment. 
Why we really need OCBytecodeToASTCache? The use case to find a source node for given pc does not look expensive. It seems to me that it will be simpler to have just a lookup API on the level of IR without the need of the full bytecode map:
```Smalltalk
RBMethodNode>>sourceNodeForPC: anInteger
	^ self ir nodeForPC: anInteger
```

dionisiydk (approved):

MarcusDenker:
The problem with just using the IR is that then we always have to keep the IR. With the mapping, we do not need to keep the IR around.

MarcusDenker (approved):

dionisiydk:
> The problem with just using the IR is that then we always have to keep the IR. With the mapping, we do not need to keep the IR around.

But it is what we have now:
```Smalltalk
method := MethodMapExamples>>#helperMethod13.
ast := method ast.
ast sourceNodeForPC: 10.
ast propertyAt: #ir "==> an IRMethod"
```
And one more observation:
```Smalltalk
((ast propertyAt: #ir) instVarNamed: #compiledMethod) == method "==> false"
```
So when we built IR with a bytecode map we perform second compilation and keep new method in a variable.

MarcusDenker:
It is like that now, yes, but is that good? 

the second compilation is not good, this should be the same method.

dionisiydk:
Sounds like the place to improve :).

PR #11655, opened 2022-09-06
Fix "undo" operation erroneously removing the dirty flag when undo operation performed after saving

### LINKED ISSUE #11649: Undo shortcut (ctrl+z) does not mark method as dirty in System browser
**Bug description**
Restoring method using ctrl+z to last compiled state does not mark the method as dirty, the code browser thinks the method definition has no change.

**To Reproduce**
1. Go to a method definition using the System browser.
2. Add something to the method.
3. Compile the method with ctrl+s.
4. Undo changes (using ctrl+z) on the method until its initial state (step 1).
5. The method can't be recompiled with ctrl+s.

**Expected behavior**
After undo to previous state of the method, it should be possible to recompile it with ctrl+s.

**Version information:**
 - OS: Mac, Linux
 - Pharo Version: 9 and 10

fmqa:
Fixes #11649

This makes sure that the editor is always marked "changed" after an undo operation.

### DIFF
--- a/src/Calypso-Browser/ClyTextEditor.class.st
+++ b/src/Calypso-Browser/ClyTextEditor.class.st
@@ -51,3 +51,12 @@ ClyTextEditor >> printIt [
 				do: [ '<error in printString: try ''Inspect it'' to debug>' ].
 			self afterSelectionInsertAndSelect: printString ]
 ]
+
+{ #category : #operations }
+ClyTextEditor >> undo [
+
+	super undo.
+
+	self editingState undoManager hasPrevious ifFalse: [ 
+		self browserTool changesCancelled]
+]
--- a/src/Calypso-Browser/ClyTextEditor.class.st
+++ b/src/Calypso-Browser/ClyTextEditor.class.st
@@ -51,12 +51,3 @@ ClyTextEditor >> printIt [
 				do: [ '<error in printString: try ''Inspect it'' to debug>' ].
 			self afterSelectionInsertAndSelect: printString ]
 ]
-
-{ #category : #operations }
-ClyTextEditor >> undo [
-
-	super undo.
-
-	self editingState undoManager hasPrevious ifFalse: [ 
-		self browserTool changesCancelled]
-]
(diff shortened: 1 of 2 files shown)


fmqa:
I will test this on my development image for a some time to make sure nothing breaks by accident before marking it ready

src/Calypso-Browser/ClyTextEditor.class.st
```diff
@@ -51,12 +51,3 @@ ClyTextEditor >> printIt [
 				do: [ '<error in printString: try ''Inspect it'' to debug>' ].
 			self afterSelectionInsertAndSelect: printString ]
 ]
-
-{ #category : #operations }
-ClyTextEditor >> undo [
```
dionisiydk: This override was added to reset the dirty state when all changes are cancelled by multiple cmd+z.
You can see that without this method the dirty indication does not disappear when the first edit is cancelled by cmd+z

fmqa: Yes I understood that. The problem here is that it interacts poorly with the save state, causing issues like #11649

If one accepts/saves the changes after performing an edit, then undos the edit via ctrl+z, it clears the dirty indication even though it shouldn't - thus preventing the user from saving after reverting a change.

fmqa: Just checking whether the undo stack is empty is not enough to determine whether the content is unchanged.

We interpret "dirty" as "differs from the saved state".

The negation "not dirty" is  "does not differ from the saved state". Or alternatively "not dirty" ≡ "same as saved state".

And, "undo stack is empty" as a condition is not equivalent to "same as saved state".

Therein lies the issue with the original code. I understand the intent of it but it seems to be based on a faulty assumption

fmqa:
CI fails due to #11657
_Edit: Seems that this is no longer the case as of 8381e90cf9511d6bc5f0a04c8b589e06ec809d2b_

src/Calypso-Browser/ClyTextEditor.class.st
dionisiydk: I see.
Extra guard for #changesCancelled will prevent the issue:
```Smalltalk
ClyTextEditorToolMorph>>changesCancelled
	self pendingText = self editingText ifFalse: [ ^self ].
         ...
```

fmqa: Agreed. That makes sense.

fmqa: I would also suggest adding an additional comment for future reference

MarcusDenker:
Failing tests not related:

MarcusDenker (approved):

NSUSpray:
This solution works for me.

PR #11659, opened 2022-09-07
RubTextEditor: Bind the MENU key to open the context menu

### LINKED ISSUE #11658: Enhance text editor with menu key binding
When a text editor is in focus, pressing the menu (≣) key should open the context menu, similar to what a right click in the editing area would do.

This mirrors the behavior seen in native OS widget toolkits.

**Describe the request**
An additional key binding for the menu key (≣) is added, which - when pressed - would have the effect of opening the context menu

**Expected behavior**
GIVEN a text area is focused
AND I press the menu ≣ button on the keyboard
THEN a context menu appears

**Expected development cost**
Minor. An additional keybinding is to be added.

**Version information:**
 - OS: Ubuntu x64
 - Version: 22.04.1
 - Pharo 10

**Additional context**
Part of #11618

fmqa:
Fixes #11658 

* Adds an additional key binding to RubTextEditor, binding the menu key (≣) to the context menu
* `KeyboardKey class >> menu` is also added to facilitate further bindings elsewhere in Pharo later

### DIFF
--- a/src/System-Platforms/KeyboardKey.class.st
+++ b/src/System-Platforms/KeyboardKey.class.st
@@ -1097,6 +1097,12 @@ KeyboardKey class >> metaRight [
 	^ self named: #META_R
 ]
 
+{ #category : #'accessing - keys - special' }
+KeyboardKey class >> menu [
+
+	^ self named: 'MENU'
+]
+
 { #category : #'accessing - keys - symbols' }
 KeyboardKey class >> minus [
 	^ self named: #MINUS
--- a/src/System-Platforms/KeyboardKey.class.st
+++ b/src/System-Platforms/KeyboardKey.class.st
@@ -1100,7 +1100,7 @@ KeyboardKey class >> metaRight [
 { #category : #'accessing - keys - special' }
 KeyboardKey class >> menu [
 
-	^ self named: 'MENU'
+	^ self named: #MENU
 ]
 
 { #category : #'accessing - keys - symbols' }
--- a/src/OSWindow-Core/OSKeySymbols.class.st
+++ b/src/OSWindow-Core/OSKeySymbols.class.st
@@ -278,7 +278,7 @@ OSKeySymbols class >> data [
 		(16r40000061 OSK_KP_9 kp_9)
 		(16r40000062 OSK_KP_0 kp_0)
 		(16r40000063 OSK_KP_PERIOD KP_Decimal)
-		(16r40000065 OSK_MENU menu) "win32 menu key / pharo issue#11659"
+		(16r40000065 OSK_APPLICATION menu) "win32 menu key / pharo issue#11659"
 		"(16r40000066 OSK_POWER power)"
 		(16r40000067 OSK_KP_EQUALS kp_equal)
 		(16r40000068 OSK_F13 f13)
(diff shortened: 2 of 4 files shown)


src/System-Platforms/KeyboardKey.class.st
```diff
@@ -1097,6 +1097,12 @@ KeyboardKey class >> metaRight [
 	^ self named: #META_R
 ]
 
+{ #category : #'accessing - keys - special' }
+KeyboardKey class >> menu [
+
+	^ self named: 'MENU'
```
NSUSpray: Most likely, 'MENU' should be replaced with #MENU. At least in other similar methods, a symbol is used, not a string.
```suggestion
	^ self named: #MENU
```

NSUSpray:
I have no effect in Windows 10 x64. I made the changes manually. Whether it is necessary or not, but the `KeyboardKey class >> #initialize` I called just in case.

fmqa:
@NSUSpray Interesting. Did you try it on a new edit window? Changing this on a running image would only affect new RubTextEditor instances. Existing instances won't be affected.

NSUSpray:
Sure. When I open a new System Browser or a new Playground, I don't see any effect. However, I noticed that the classes of these text editors is not RubTextEditor, but RubScrolledTextMorph and SpRubScrolledTextMorph.

fmqa:
You are right. I tested on a Windows 10 VM just now and it has no effect. This only seems to work on Linux.

guillep:
Hi all, this is a super great work, I love having more shortcutty interactions :)

The cause of that could be multiple:
 - SDL events do not send us key presses to the menu key (unlikely?)
 - we don't have a correct mapping with Window's menu key in `KeyboardKey` => I'd look for the cause around here!

fmqa:
The missing mapping might be a problem. `KeyboardKey.class.st` contains this:

```st
{ #category : #windows }
KeyboardKey class >> initializeWindowsVirtualKeyTable [
	self flag: #pharoTodo. "Some values are missing, need to gather them from users who are able to."
	WindowsVirtualKeyTable := Dictionary new.
	...    
```

NSUSpray:
I found that in `KeyboardKey class >> # initializeKeyTable` the key corresponding to the virtual code 0xA5 (VK_RMENU) is designated not as Menu, but as Alt_R (kVK_RightOption). Could this be the issue?

fmqa:
@NSUSpray Would you mind testing again with the latest commit. The extra symbol mapping in revision c3f06a0eb4b71a94a7e6ef97cda0204c066bae35 works for me on a Windows VM

NSUSpray:
Yes, now it works!

I found that the corresponding Windows constant is correctly named VK_*APP*S (VK_MENU is the Alt key and VK_RMENU is the right Alt key). Maybe the original name of the constant as OSK_*APP*LICATION is reasonable and should not be changed?

NSUSpray:
I would also like to note that the menu opens at the position of the *mouse* cursor (pointer), while according to the canons it should open at the position of the *text* cursor (caret).

MarcusDenker:
Failing tests not related:

Ducasse:
So is this block only for not categorised methods?

fmqa:
I don't see the uncategorized methods. All added methods in the PR have a category.

PR #11664, opened 2022-09-09
SDL_HINT_VIDEO_X11_NET_WM_BYPASS_COMPOSITOR=0 on Linux

LucFabresse:
Hi,

This is not as bad as @martinmcclure on my system: Manjaro (Gnome 42.4 X11).
Anyway, I did a pass on the fix of @ezeaguerre (thanks!) and here the code.
It would be better if you could give it a try.
IMO, it does not hurt to integrate it because it allows one to add specific SDL hints for Linux.


ezeaguerre:
Awesome! A unix platform makes far more sense than the null one I think.

MarcusDenker:
Failing tests not related:

demarey (approved):

PR #11668, opened 2022-09-12
clean String reading from FFI to return bytes to allow decoding with any decoder.

### LINKED ISSUE #11665: ExternalAddress>>#fromCString should return raw bytes
**Bug description**
Currently `ExternalAddress>>#fromCString` returns a byte array utf8 decoded.
It works in case the content is utf8 encoded but it fails when the content has another encoding (e.g. Windows).
`#fromCString` should return raw bytes to let the user choose the appropriate decoder to read the String.
`ExternalData>>#readString`uses `ExternalAddress>>#fromCString` and also always returns an utf8 string.
It should also return raw bytes

We should see the impact on users of these methods: maybe define a new API to get raw bytes and add a deprecation rule

demarey:
fixes #11665

### DIFF
--- a/src/FFI-Kernel/ExternalAddress.class.st
+++ b/src/FFI-Kernel/ExternalAddress.class.st
@@ -245,5 +245,5 @@ ExternalAddress >> utf8StringFromCString [
 	WARNING: the referenced data MUST ends with a NULL character (byte 0).
 	"
 
-	^ self bytesFromCString utf8Decoded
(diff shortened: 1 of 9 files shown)


Ducasse:
This looks good to me.

Ducasse:
Broken tests are unrelated.

MarcusDenker:
Failing tests not related:

PR #11672, opened 2022-09-12
Enabling formating of uncompiled method

### LINKED ISSUE #11640: Can't format a new method before compiling it
**Bug description**
Error when formatting a method that isn't yet compiled

**To Reproduce**
1. Browse any class in Calypso browser
2. Go on the tab "+ Inst. side method"
3. Write anything (but don't compile)
4. Press ctrl+shift+F to format
5. Instance of ClyTextEditorContext did not understand #methodEditor

**Version information:**
 - OS: Windows
 - Pharo-11.0.0+build.182.sha.160dedc341254e1e1ec8a0dc0a2776c615834b15 (64 Bit)

carolahp:
Fix #11640

### DIFF
--- a/src/Calypso-Browser/ClyTextEditorContext.class.st
+++ b/src/Calypso-Browser/ClyTextEditorContext.class.st
@@ -13,6 +13,11 @@ ClyTextEditorContext >> browser [
 ]
 
 { #category : #public }
+ClyTextEditorContext >> editor [
+	^ tool
+]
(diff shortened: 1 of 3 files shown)


Ducasse:
I hope that moving from methodEditor to editor will not make us lose some functionality. We will see.

Ducasse:
Broken tests are not related

Issue #11670, opened 2022-09-12
[CI] Failing test: ReleaseTest>>testUndeclared


MarcusDenker:
**Bug description**
ReleaseTest>>testUndeclared is failing

```
Found undeclared Variables:  SpMethodToolbarPresenter in: AbstractMessageCentricBrowser>>#initializePresenters
```

MarcusDenker:
The method looks like this:

```
initializePresenters

	messageList := self instantiate: MessageList.
	toolbarPresenter := self instantiate: SpMethodToolbarPresenter
```

The class SpMethodToolbarPresenter does not exist in the image

MarcusDenker:
Same problem in VersionBrowser>>#initializePresenters

MarcusDenker:
might be already fixed, we wait for the CI to do a build. (see #11677)

MarcusDenker:
fixed my merging https://github.com/pharo-spec/NewTools/pull/417

Issue #11669, opened 2022-09-12
[CI] Failing test: ProperMethodCategorizationTest>>testNoUncategorizedMethods


MarcusDenker:
**Bug description**
ProperMethodCategorizationTest>>testNoUncategorizedMethods is failing since some days

```
the following classes have uncategorized methods: an OrderedCollection(LGitMessageDecoder class)
```

MarcusDenker:
I did a PR https://github.com/pharo-vcs/libgit2-pharo-bindings/pull/67

MarcusDenker:
Even though the PR was merged, the test is still failing

MarcusDenker:
The problem is fixed in libgit (PR is merged).

But we need to do a new release and change Iceberg (I think) to load that.

MarcusDenker:
this is fixed

PR #11683, opened 2022-09-14
Add more executable comments for Collection

privat:
Fell free to comment

### DIFF
--- a/src/Collections-Abstract/Collection.class.st
+++ b/src/Collections-Abstract/Collection.class.st
@@ -860,6 +860,11 @@ Collection >> doWithIndex: elementAndIndexBlock [
 { #category : #private }
 Collection >> emptyCheck [
 
+	"Signal CollectionIsEmpty if the collection is empty"
+
+	"#(10 20) emptyCheck >>> #(10 20)"
+	"[#() emptyCheck] on: CollectionIsEmpty do: [ 'oops' ] >>> 'oops'"
+
 	self isEmpty ifTrue: [self errorEmptyCollection]
 ]
 
@@ -1029,6 +1034,10 @@ Collection >> hash [
 Collection >> identityIncludes: anObject [ 
 	"Answer whether anObject is one of the receiver's elements."
 
+	"({#hello. #world} identityIncludes: #hello) >>> true"
+	"({'hel','lo'. 'world'} identityIncludes: 'hello') >>> false"
+	"({'hel','lo'. 'world'} includes: 'hello') >>> true"
+
 	self do: [:each | anObject == each ifTrue: [^true]].
 	^false
 ]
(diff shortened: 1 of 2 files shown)


src/Collections-Abstract/Collection.class.st
```diff
@@ -228,11 +242,13 @@ Collection >> anySatisfy: aBlock [
 
 { #category : #converting }
 Collection >> asArray [
-	"Answer an Array whose elements are the elements of the receiver.
+	"Answer an new Array whose elements are the elements of the receiver.
```
Ducasse: I'm not sure about the new part because if you send #(1 2 3) asArray it probably returns self so just the receiver and not a copy.

src/Collections-Abstract/Collection.class.st
```diff
@@ -288,11 +310,26 @@ Collection >> asCommaStringAnd [
...
+	
+	"|x| x := {'one' -> 1. 'two' ->2} asDictionary. x asDictionary == x >>> true"
+
 	^ self as: Dictionary
 ]
 
 { #category : #converting }
 Collection >> asIdentitySet [
+
+	"Return a new IdemtitySet based on self."
```
Ducasse: Idemtity -> Identity

src/Collections-Abstract/Collection.class.st
privat: You are right. The method builds a new array, but is redefined in Array to return self.
So I will assume for now that all the `asX` methods try to return self if applicable.

MarcusDenker:
Failing tests not related:

hogoww:
I haven't read everything (you did a LOT of them !)
All I read look very good.
One question though, why use bags in the test cases of the collections ? 
Also, you can use #() to create arrays with statics elements in them rather than {} which creates it at run time.

privat:
>  One question though, why use bags in the test cases of the collections ?

I used bags because `asBag` is shorter to write than `asOrderedCollection` and also this shows examples on non sequenceable collections.

>  Also, you can use #() to create arrays with statics elements in them rather than {} which creates it at run time.

From an example point of view, I'm not sure what is the *recommended* usage of Pharoers.

My external point of view is that `#()` is more cumbersome than '{}' since you cannot mutate it or edit the source code to add points or fractions. Basically I see `#()` as a low level optimization, but optimization is not required when writing docs (and premature optimization is the root of all evil bla bla bla)

The only nice thing of `#()` (beside speed and memory efficiency) is that you only need space as separator.
Ok `{1.2.3.4.5.6}` is equally source-code efficient, but some people might find it hard to parse, so it is not something you want in an example.

Multiple times I tried to settle on `#{}` but the parser does not approve my way of making a compromise. :)

hogoww:
All of that is very true !
But executable comments are more than documentation.
There are also unit test cases.

AFAIK They were supposed to be executed in the CI/Test runners, I do not know if they do it currently.

My point is that performance may matter, although probably not **Not that much**

Ducasse:
What we should see is that it would be good to limit dependency between classes.
To me Collection, Array and OrderedCollection are really in the core while we could imagine that Bag can be packaged separately. For example what I envision is a minimal core and that we have well packaged collections that people can load when they want. This is the http://github.com/pharo-containers/ project. 
I have Stack, ...

In the past we did the mistake to "improve Date and Time" while now I would have prefered to clone Date and Time and improves it and reduce the core of Date and Time in Core.
I would like to have a way smaller core and well packaged libraries.

privat:
@Ducasse I thought that Bag was popular, but there is only 49 `Bag` and 10 `asBag`, whereas there is 418 `Set` and 278 `asSet` in my pharo11 image. So I will reroll the PR and restrict myself to more core classes : Collection, Array and OrderedCollection.

I assume it is OK to add `Set`, `Interval` and `String` to the happy few?

Ducasse:
Yes!

privat:
I did a big reroll, and moved things I like less in a distinct PR #11699.
I can split the PR further if it is too much to review at once.

privat:
@Ducasse if you merged #11699 , you should merge this one also

PR #11687, opened 2022-09-15
Symbol a little more as block

privat:
Implements Symbol>>value:value: so symbol can be used in more methods of collection (and other classes)

You can now write the following

```
(#(1 9 0 6 6) sorted: #>=) >>> #(9 6 6 1 0)
(#(2r101 2r011 2r100) inject: 0 into: #bitXor:) >>> 2r010 
```

Maybe we can simulate the whole API of blocks on symbols, but I'm not sure: 1. that this is a good idea nor 2. what is best way to do it without too much boilerplate code, duplication and technical dept. Maybe with traits?

### DIFF
--- a/src/Collections-Strings/Symbol.class.st
+++ b/src/Collections-Strings/Symbol.class.st
@@ -408,6 +408,11 @@ Symbol >> createSymbol [
 
 { #category : #evaluating }
 Symbol >> cull: anObject [ 
+
+	"A Symbol can be mostly used as a BlockClosure. See value: for details"
+	
+	"#asUppercase cull: 'hello' >>> 'HELLO'"
+
 	^anObject perform: self.
 ]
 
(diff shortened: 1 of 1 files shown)


svenvc:
I kind of like this style but many people do not.

This is something that comes up from time to time, it has been discussed before.

jordanmontt:
Indeed this is a hot debate :)
What I know is that using `Symbol` as blocks is not recommended because the `perform:` message makes it significantly slower

MarcusDenker:
The fun thing is that it is actually faster. If you use a block, the VM has to create the block from the CompiledBlock at runtime. The Symbol is there and #value 

```
[#(1) do: #yourself] bench 

[#(1) do: [:each | each yourself]] bench 

```
This gets better if you enable "clean block closures", but even there the Symbol case is a little bit faster when I try it.

jordanmontt:
Interesting! So why we discourage people of using Symbol as blocks ?

privat:
Encouraging using symbol as block is very fragile. So it is better to not expand the feature that much.

PR #11688, opened 2022-09-15
RFC: Teach Dictionary to simulate being a Block

privat:
I'm mostly playing with Pharo here. :)

So the deal is to teach Dictionary the meaning ~of life~ `value:` so its instances can be used on *enumerating* methods of Collections (and others).

A dictionary used to transform elements of a collection:

```
|d|
d := {-1->'negative'. 0->'zero'. 1->'positive'} asDictionary.

((-10 to: 10 by: 5) sign collect: d)
  >>> { 'negative'. 'negative'. 'zero'. 'positive'. 'positive' }"
```

A dictionary used to filter elements of a collection:

```
|d|
d := {$A->false. $T->true. $G->true. $C->false} asDictionary.

('AGTCTGCACTC' count: d)
  >>> 5"
```

### DIFF
--- a/src/Collections-Unordered/Dictionary.class.st
+++ b/src/Collections-Unordered/Dictionary.class.st
@@ -495,6 +495,14 @@ Dictionary >> collect: aBlock [
 	^newCollection
 ]
 
+{ #category : #evaluating }
+Dictionary >> cull: anOject [
+
+	"DO NOT USE. Method only here to simutate BlockClosure. See value: for details."
+	
+	^self at: anOject 
+]
+
 { #category : #adding }
 Dictionary >> declare: key from: aDictionary [ 
 	"Add key to the receiver. If key already exists, do nothing. If aDictionary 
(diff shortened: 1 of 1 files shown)


Ducasse:
I do not like the idea that a dictionary could be used as a blockclosure.
I do not like the general use of cull: because it is reflective. 
Having a dynamic language does not mean that we can treat eggs as apples.

privat:
@Ducasse I agree with your point of view. The proposal is brittle and eventually will fail in the hand of the user with weird dnu on random symbol objects.

A cleaner and robust approach could be to deal with real block objects something like

    Symbol>>asBlock
      return [:k| self at: k]

But I do not think there is any gain to write

    (-10 to: 10 by: 5) sign collect: d asBlock

instead of

    (-10 to: 10 by: 5) sign collect: [:k|d at:k]

It is basically the same length, and the explicit block makes the code more understandable.
So I close the PR

Ducasse:
I agree :)
And I like that we use the support of syntax to delimit types. 
Because this is not because Pharo is dynamically typed that we have to blur all frontier. 
I would love to have a nice type inferencer that helps me.

PR #11689, opened 2022-09-16
DrTest see extension classes

privat:
With the executable comment plugin of DrTest, selecting a package now gives access to classes extended in that package.

For classes extended in a package, only extended methods are tested (not the whole class).
For classes defined in a package, all methods are tested, including those extended in other package. This is the old behavior and  is compatible with the system browser view (and better be safe than sorry).

The behavior is compatible when multiple packages are selected and a same class is defined/extended more than once.

In the list view, extended classes are displayed with the right color (gray by default) to help the user to visualize that they selected only a subset of the methods (the subset related to the selected packages).


MarcusDenker:
failing test not related:

Failing tests not related:

Ducasse:
Cool we should integrate the work on aboubacar too.

PR #11693, opened 2022-09-16
Get the variable names of objects

privat:
This comes from a question on the Discord. I thought it was a nice exercise for me :)

The context, if I'm correct, is that you are playing on the playground and call complex methods that call complex methods, etc.
At some point, you want to know the name of temporary variable that hold some objects.

The proposed method Context>>variableNamesOf: just do that

```
|a b|
a := 'hello'.
b := 'world'.
(thisContext variableNamesOf: a) >>> #(#a).
b := a.
(thisContext variableNamesOf: a) >>> #(#a #b).
```

### DIFF
--- a/src/Kernel-Tests/ContextTest.class.st
+++ b/src/Kernel-Tests/ContextTest.class.st
@@ -48,7 +48,7 @@ ContextTest class >> returnNonActiveContextOfBlock [
 	^ Block value
 ]
 
-{ #category : #'as yet unclassified' }
+{ #category : #utils }
 ContextTest >> DoIt [
 
 	"Mock DoIt method that does not do things"
@@ -231,7 +231,7 @@ ContextTest >> testReturn [
 ]
 
 { #category : #tests }
-ContextTest >> testSearchPlayground [
+ContextTest >> testSearchDoIt [
 
 	|doit|
 	
@@ -247,11 +247,11 @@ ContextTest >> testSearchPlayground [
 		method: aCompiledMethod
 		arguments: #().
 		
-	self assert: (aMethodContext searchPlayground) equals: doit
+	self assert: (aMethodContext searchDoIt) equals: doit
 ]
 
 { #category : #tests }
-ContextTest >> testSearchPlaygroundNil [
+ContextTest >> testSearchDoItNil [
 
 	aMethodContext := Context
 		sender: thisContext
@@ -259,7 +259,7 @@ ContextTest >> testSearchPlaygroundNil [
 		method: aCompiledMethod
 		arguments: #().
 		
-	self assert: (aMethodContext searchPlayground) equals: nil
+	self assert: (aMethodContext searchDoIt) equals: nil
 ]
 
 { #category : #tests }
--- a/src/Kernel-Tests/ContextTest.class.st
+++ b/src/Kernel-Tests/ContextTest.class.st
@@ -405,6 +405,37 @@ ContextTest >> testSteppingReturnSelfMethod [
 	self assert: newContext identicalTo: aMethodContext
 ]
 
+{ #category : #'as yet unclassified' }
+ContextTest >> testvariableNamesOf [
+
+	| a b c |
+	a := 'hello'.
+	b := 'hel' , 'lo'.
+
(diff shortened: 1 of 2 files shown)


src/Kernel/Context.class.st
```diff
@@ -2251,6 +2262,16 @@ Context >> unwindTo: aContext [
...
 ]
 
+{ #category : #'debugger access' }
+Context >> variableNamesOf: anObject [
+
+	"Retrieve the names of the arguments and temporary variables, if any, that refer to the given object."
+
+	^ (1 to: self size)
+		  select: [ :i | (self at: i) == anObject ]
+		  thenCollect: [ :i | self home method tempNames at: i ]
```
MarcusDenker: I think this might not work for temp variables that are allocated in a temp Vector (espcaping vars, when they are written to from a closure).

In smalltalk 80, there is a direct mapping temp var --> offset in the context, but with the closure implementation, they might be in an array that then is referenced from the context.

privat: You are so right, and thanks to your explanations I reimplemented this with Variables (♥️ reification).
That was a lot easier!

Please review again.

Ducasse:
Pay attention you have too much fun :)

jordanmontt:
Hello, I did not understand the do it context is needed. Is it only for the tests?
`Context >> searchDoIt`

src/Kernel/Context.class.st
```diff
@@ -2251,6 +2262,16 @@ Context >> unwindTo: aContext [
...
 ]
 
+{ #category : #'debugger access' }
+Context >> variableNamesOf: anObject [
+
+	"Retrieve the names of the arguments and temporary variables, if any, that refer to the given object."
+
+	^self astScope allTemps
+		select: [ :v | (v readInContext: self) == anObject ]
+		thenCollect: [ :v | v name ]
```
dionisiydk, line 2272: Why you want just names? We should stop think in terms of primitive objects. Just return full objects representing temp variables.
I would also rename the method to #tempVariablesContaining: . Then you can also define allVariablesContaining: which will lookup the receiver in addition.

src/Kernel/Context.class.st
```diff
@@ -1778,6 +1778,17 @@ Context >> runUntilErrorOrReturnFrom: aSender [
 	].
 ]
 
+{ #category : #'debugger access' }
+Context >> searchDoIt [
```
dionisiydk, line 1782: There are other search methods in context but with the word #lookup. I would follow this name here: #lookupDoItContext

src/Kernel-Tests/ContextTest.class.st
```diff
@@ -367,6 +405,60 @@ ContextTest >> testSteppingReturnSelfMethod [
...
+		c := nil.
+		e := a.
+		self assert: (thisContext variableNamesOf: a) sorted equals: #( #a #d #e #t #x #z ).
+		].
+	
+	a := nil
+]
+
+{ #category : #tests }
+ContextTest >> testvariableNamesOfWithArgs [
```
dionisiydk, line 457: many tests are missing uppercase letter after the test word

MarcusDenker:
Failing tests not related:

- [CI] two failing test in SpCommandTest #11702

jordanmontt:
If you don't declare the temps variables you get an empty array

Capture d’écran 2022-09-21 à 23 31 41

I don't know if that is expected or not. But one normally doesn't declare the names of the temp variable in the Playground

MarcusDenker:
Yes, these Variables in the playground are not temps, but WorkspaceVariables, what the compiler does is to add a special "RequestorScope" to the scope hierarchy that has the defintions.

This means that if #variableNamesOf: would ask the scope (and make sure the parent scope is asked, too), we could could find those variables (reading works with #readInContext:, too).

guillep:
Hi all, what do we do with this?

privat:
I will reroll when I have the time.

jecisc:
This PR is here since a long time and the team has a lot of things to do. For now we are closing this PR because it seems that other things have a higher priority.

If someone has a big need for it we can reopen later and do another pass.

PR #11695, opened 2022-09-18
deprecate taskbarIcon and systemIcon

estebanlm:
do not use taskbarIcon and systemIcon. Use taskbarIconName and systemIconName instead.

### DIFF
--- a/src/Calypso-SystemTools-Core/ClyClassIconTableDecorator.class.st
+++ b/src/Calypso-SystemTools-Core/ClyClassIconTableDecorator.class.st
@@ -16,9 +16,10 @@ ClyClassIconTableDecorator class >> decorateMainTableCell: anItemCellMorph of: a
 	| icon |
 	super decorateMainTableCell: anItemCellMorph of: aDataSourceItem.
 	    
-   icon := Object environment at: aDataSourceItem name asSymbol
-            ifPresent: [ :aClass | aClass systemIcon ]
-            ifAbsent: [ Object systemIcon ].
+   icon := self iconNamed: (Object environment 
+		at: aDataSourceItem name asSymbol
+		ifPresent: [ :aClass | aClass systemIconName ]
+		ifAbsent: [ Object systemIconName ]).
 
     anItemCellMorph definitionMorph: icon asMorph
 ]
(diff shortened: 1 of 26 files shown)


Ducasse:
It would be nice to be able to get rid of this kind of code `Smalltalk tools fileList`

estebanlm:
yes, eventually all those Smalltalk singletons needs to be move to the running application (StPharoApplication, in this case).
this will easy things, but still a lot of things to think and re-design :/

In any case, my next public enemy N. 1 is Object>>#iconNamed:, which is also a huge problem ;)

Ducasse:
Before redesigning we could have a tools registry in StApplication and we can move all the `Smalltalk tools` to `self toolNamed:`.
I can do it if you want or we can take 1 hours per week and do some pair programming and do it. It can be fun.

estebanlm:
no idea why that test is failing, it does not seems related with the change..

Ducasse:
I agree that killing Object>>iconNamed: is good.

MarcusDenker:
one failing test is a known failing test:

but these are new and do not fail currently in Pharo11:

- SpMorphicBoxLayoutTest>>testClassIsDisplayedWithCorrectAlignment
- SpDemoTest>>testSmokeTestForDemoPages

MarcusDenker:
for testSmokeTestForDemoPages this is a generic on: Error handler, we need to check what the real error is:

```
	on: Error
			do: [ :e | 
				self fail: 'Broken demo page: ' , demoPage asString , '

```

MarcusDenker:
It seems both are due to the same problem:

- the rewrite in the deprecated method Object>>#systemIcon is missing a @:

```
	transformWith: '`@receiver systemIcon' -> '`@receiver iconNamed: `@receiver systemIconName'.
```

MarcusDenker:
Just checked: tests are green after that, I will try to edit the PR directly

estebanlm:
oops :)

MarcusDenker:
Test look good now, but of course now the MacosX executor is down...

MarcusDenker (approved):

estebanlm:
yes, that's exactly what I'm doing now :)
but moving things takes time too, so... slowly :(

Issue #11697, opened 2022-09-19
[CI] Mac executor down due to network issues


MarcusDenker:
**Bug description**
Due to networking problems, the mac executor is down

MarcusDenker:
Works again!

PR #11698, opened 2022-09-19
Prevent BlockClosure>>benchFor: to loop infinitely

privat:
`[ 100 factorial ] benchFor: 5` hangs since `5 wait` silently fails (5 is not a duration) thus `run` never became `false`.

The PR adds the catch of exceptions, ensure run is set to false, and reraises the exception on the main thread (so the user can see and debug it).

Questions for reviewers: It is OK to just `exeption signal` to reraise the exception in the correct thread?


Ducasse:
Often a pattern used is to force conversion of the parameter

```
benchFor: aParam
    param := aParam asSeconds
```

But here it is not that nice since we can pass seconds and millisecond even if benching using milliseconds does not make that much sense.

S

MarcusDenker:
Failing tests not related:

- [CI] Mac executor down due to network issues #11697

PR #11699, opened 2022-09-20
Add executable comments on Collection for do: and stream related

privat:
This is a partial reroll of #11683 that focuses on example with do blocks and streams

I do find these examples complex and big. But I lack imagination for better alternatives.

The issue for stream is that you need to setup a stream then validate it. But since executable examples need to be on a single line, this makes the lines hard to read.

There is an equivalent issue with methods that takes a `do` block whose value is not returned in the expression.
Therefore you need to setup something before the block, altering in the `do` block, then validating that something after me method.
I tried to setup a stream (and putting things inside), a collection (and adding elements), a single variable (and assigning it). I'm not sure wich one I prefer. :(

### DIFF
--- a/src/Collections-Abstract/Collection.class.st
+++ b/src/Collections-Abstract/Collection.class.st
@@ -1075,8 +1075,8 @@ Collection >> printOn: aStream delimiter: delimString last: lastDelimString [
 		aStream print: elem]
 	separatedBy: [
 		n = sz
-			ifTrue: [aStream print: lastDelimString]
-			ifFalse: [aStream print: delimString]]
+			ifTrue: [aStream nextPutAll: lastDelimString]
+			ifFalse: [aStream nextPutAll: delimString]]
 ]
 
 { #category : #enumerating }
--- a/src/Collections-Abstract/Collection.class.st
+++ b/src/Collections-Abstract/Collection.class.st
@@ -340,22 +340,29 @@ Collection >> asSortedCollection: aSortBlock [
 
 { #category : #printing }
 Collection >> asStringOn: aStream delimiter: delimString [
+
 	"Print elements on a stream separated
 	with a delimiter String like: 'a, b, c'
 	Uses #asString instead of #print:."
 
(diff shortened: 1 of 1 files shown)


src/Collections-Abstract/Collection.class.st
```diff
@@ -340,23 +340,30 @@ Collection >> asSortedCollection: aSortBlock [
...
 { #category : #printing }
 Collection >> asStringOn: aStream delimiter: delimString [
+
 	"Print elements on a stream separated
 	with a delimiter String like: 'a, b, c'
 	Uses #asString instead of #print:."
 
-	self do: [:elem | aStream nextPutAll: elem asString]
-		separatedBy: [aStream nextPutAll: delimString]
+	"String streamContents: [:s| 'abcd' asStringOn: s delimiter: '->'] >>> 'a->b->c->d'"
```
Ducasse, line 348: May be this is too early but you should pay attention that >>> is not sugar but a real binary message so we should put parentheses around String ...] so hat the >>> is sent to the resulting result 

I think that we should do that all over the comments. 
Normally we should have an extension of DrTests that is check comments.

privat: I tried to add the parenthesis, but the current implementation of executable comments seems to split the line on the first >>> then parse and evaluate the left side and the right side independently. Precedence and anything is therefore just ignored. So, missing parenthesis as not signaled when I run the tests.

Fun example: `(1 >>> 1)` is considered a syntactic error because unmatched `(` in the first part

Ducasse:
We should discuss in front of Pharo. Because when I execute
```
"(String streamContents: [:s| 'abcd' asStringOn: s delimiter: '->']) >>> 'a->b->c->d'"

'a->b->c->d'->'a->b->c->d'

```

I get the correct result. 
but when I execute the same without the () I get an empty string because the message >>> is sent to the block and not the result of the parenthesed expression as in the first case. 
So I think that I'm right and that we should always make sure that the left side of >>> is executed before >>> is sent.

PR #11701, opened 2022-09-21
11613-Review-RBLiteralNode-classvalue 

### LINKED ISSUE #11613: Review RBLiteralNode class>>#value:
- RBLiteralNode class>>#value: checks for the class using equality

shouldnt it check for #isArray or #isByteArray to find out if it should use RBLiteralArrayNode or RBLiteralValueNode?

I know that the #isByteArray is not yet in the image, but we can easily add it.

MarcusDenker:
add a comment explaining why we do not use #isByteArray here, fixes #11613

(I do not like the method, maybe the check should be done by the user...

### DIFF
--- a/src/AST-Core/RBLiteralNode.class.st
+++ b/src/AST-Core/RBLiteralNode.class.st
@@ -22,7 +22,9 @@ RBLiteralNode class >> isAbstract [
 ]
 
 { #category : #'instance creation' }
-RBLiteralNode class >> value: aValue [ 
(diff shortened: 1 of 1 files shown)


MarcusDenker:
Failing tests not related:

- [CI] two failing test in SpCommandTest #11702

dionisiydk (approved):

Ducasse:
It took me while to understand. May be the comment should say value argument can be x or y.

PR #11716, opened 2022-09-22
Faster Array>>asX and Array>>as:

privat:
Important low-level change where conversion from Array are made faster. (up to x3 (at least) for `asOrderedCollection`)

This is useful as `{1. 2. 3} asFoo` (or `{1. 2. 3} as: Foo`) is the current de facto syntax to have initialized collections of various types with user defined content.

Note that subclasses of Collection that redefine `newFrom:` should also redefine `newFromArray:` or bad things could happen.
This is because the default (inherited) implementations of `newFromArray:` uses `new:` and `add:` (or `at:put:` for ArrayedCollection).

### DIFF
--- a/src/Collections-Abstract/Collection.class.st
+++ b/src/Collections-Abstract/Collection.class.st
@@ -28,7 +28,7 @@ Collection class >> newFromArray: anArray [
 	"Fast initialization with the items of a given array.
 	This initializes elements faster that the generic withAll: or newFrom: methods.
 
-	The main selling point is that dyamic arrays like {1. 2. 3} are really fast in Pharo.
+	The main selling point is that dynamic arrays, like {1. 2. 3}, are really fast in Pharo.
 	So other collections can be easily and efficiently initialized with `{1. 2. 3} asFoo` syntax.
 
 	Examples:
--- a/src/Collections-Abstract/Collection.class.st
+++ b/src/Collections-Abstract/Collection.class.st
@@ -333,7 +333,7 @@ Collection >> asIdentitySet [
 { #category : #converting }
 Collection >> asNewArray [
 
-	"Like asArray: but return a copy is self is already an Array.
+	"Like asArray: but return a copy if self is already an Array.
 	This ensures that the result is always a new Array"
 
 	"'foo' asNewArray >>> 'foo' asArray"
--- a/src/Collections-Abstract/Collection.class.st
+++ b/src/Collections-Abstract/Collection.class.st
@@ -30,6 +30,9 @@ Collection class >> newFromArray: anArray [
 
 	The main selling point is that dynamic arrays, like {1. 2. 3}, are really fast in Pharo.
 	So other collections can be easily and efficiently initialized with `{1. 2. 3} asFoo` syntax.
+	
+	Important: Subclasses of Collection that redefine withAll: or newFrom: should also redefine this method
+	either by having a proper implementation (specific to Arrays) or by calling the redefined versions of withAll:/newFrom:.
 
 	Examples:
 	{ 1. 2. 3 } asSet >>> Set new add: 1; add:2; add:3; yourself
--- a/src/Collections-Abstract/Collection.class.st
+++ b/src/Collections-Abstract/Collection.class.st
@@ -23,6 +23,37 @@ Collection class >> isAbstract [
 ]
 
 { #category : #'instance creation' }
+Collection class >> newFromArray: anArray [
+
+	"Fast initialization with the items of a given array.
+	This initializes elements faster that the generic withAll: or newFrom: methods.
+
+	The main selling point is that dyamic arrays like {1. 2. 3} are really fast in Pharo.
+	So other collections can be easily and efficiently initialized with `{1. 2. 3} asFoo` syntax.
+
+	Examples:
+	{ 1. 2. 3 } asSet >>> Set new add: 1; add:2; add:3; yourself
+	{ 1. 2. 3 } asOrderedCollection >>> OrderedCollection new add: 1; add:2; add:3; yourself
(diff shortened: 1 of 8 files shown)


src/Collections-Abstract/Collection.class.st
```diff
@@ -22,6 +22,37 @@ Collection class >> isAbstract [
 	^self name = #Collection
 ]
 
+{ #category : #'instance creation' }
+Collection class >> newFromArray: anArray [
+
+	"Fast initialization with the items of a given array.
+	This initializes elements faster that the generic withAll: or newFrom: methods.
+
+	The main selling point is that dyamic arrays like {1. 2. 3} are really fast in Pharo.
```
astares: typo: dynamic

jordanmontt:
> Note that subclasses of Collection that redefine newFrom: should also redefine newFromArray: or bad things could happen.
This is because the default (inherited) implementations of newFromArray: uses new: and add: (or at:put: for ArrayedCollection).

Maybe you can add this in a method comment like that the developers can know :)

hogoww:
Presentation talking about this topic (thanks Jean!)
https://rmod-files.lille.inria.fr/Team/PharoPresentations/2022-DynamicInitializationOfCollection-Privat.pdf

hogoww (changes requested):
nevermind, I misread :D

src/Collections-Abstract/Collection.class.st
```diff
@@ -296,6 +330,18 @@ Collection >> asIdentitySet [
 	^(IdentitySet new: self size) addAll: self; yourself
 ]
 
+{ #category : #converting }
+Collection >> asNewArray [
+
+	"Like asArray: but return a copy is self is already an Array.
```
Ducasse: Like asArray: but return a copy if self is already an Array.

src/Kernel/Object.class.st
```diff
@@ -140,6 +140,17 @@ Object class >> newFrom: aSimilarObject [
 		ifFalse: [self basicNew]) copySameFrom: aSimilarObject
 ]
 
+{ #category : #'instance creation' }
+Object class >> newFromArray: anArray [
```
Ducasse, line 144: Do we want this on Object?
I would not.

src/Collections-Sequenceable/Array.class.st
```diff
@@ -136,13 +136,48 @@ Array class >> new: sizeRequested [
...
 { #category : #converting }
 Array >> asArray [
 	"Answer with the receiver itself."
 
 	^ self
 ]
 
+{ #category : #converting }
+Array >> asDictionary [
+
```
Ducasse: I would not add asDictionary because it is really strange to me and can only work if we have binding in the array.

guillep: I'm not against this change per se.
@Ducasse , asDictionary already exists. And works only for collections with associations.
So, this change is semantically equivalent.

Of course, the rationale behind this is speed, but if the change is questioned for Dictionaries, then all the rest of the changes are too. The question is: Do we want `{ a->b. c->d } asDictionary` to be preferred to  `Dictionary new at: a put: b; at: c put: d; yourself`?.

Regarding performance, however, this only cuts one message send, right? Instead of

```
anArray asDictionary
anArray as: Dictionary
Dictionary newFromArray: anArray
...
```

It would be:
```
anArray asDictionary
Dictionary newFromArray: anArray
...
```

right?
Is there much to gain in here?

src/Kernel/Object.class.st
privat: As the comment of the method says, this is only for backward compatibility.

Some existing clients can (directly or not) rely on the fact that `{1. 2. 3} as: SomeClass` calls back `SomeClass>>newFrom:`. With the change proposed in the PR the likely DNU selector `newFromArray:` will be called instead. So to have a semantically equivalent change, I just globally forward `newFromArray:` to `newFrom:` unless a better `newFromArray:` is redefined.

Side note: I'm not sure about the *backward compatibility policy* of Pharo (is there an implicit or an explicit one?) I can adapt the PR to what you prefer: I can drop `Object class>>newFromArray:` (and zut for guys that are on the way). I can also harden the current proposal for compatibility: move specific behavior Collection>>newFromArray into selected subclasses (so weird subclasses of Collection that redefine newFrom: or withAll: will not break). I can even add type guards to that weird subclasses of OrderedCollection or Array do not feel the change (fragile base class bla bla bla).

Other note: the rationale and laws of Object-level methods like `Object>>as:` or `Object class>>newFrom:` are not completely evident for me. While this is not the point of the PR to discuss these, feel free to say if I broke some beahviors or rules.

src/Collections-Sequenceable/Array.class.st
privat: Since there was an existing `asDictionary` in Array, I just adapted it to use the fast path from `Collection class>>newFromArray:` and be equivalent with `as: Dictionnary` (as `Array>>as:` uses the fast path by default).

Note: see the doc of `Collection class>>newFromArray:` for where is the gain

privat: I forgot the rule of  ``do not comment before your first coffee''.

It seems that there was in fact no previous definition of `asDictionary` therefore I'm not sure why I added it.
I will remove it since there is almost no gain as noted @guillep

MarcusDenker: There is #asDictionary in Pharo (on Collection). It has 90 senders.

MarcusDenker: I propose to make *this* PR solely about speeding up what we have and do change as part of another step.

privat: I removed asDictionary

MarcusDenker:
Failing tests not related:

PR #11717, opened 2022-09-23
POC BlockClosure>>valueAtCompileTime

privat:
This is a proof of concept.

@MarcusDenker talked about a need to have heavy or complex objects (initialized collections for instances) evaluated at compile time instead of at runtime.

This PR offers a generalized way to evaluate any expression at compile time, with obviously the support of the Opal compiler.
Instead of a syntax extension, the magic is hidden behind a special method `BlockClosure>>valueAtCompileTime`.

Why blocks?
Blocks are a very powerful and standard Pharo element that encloses a semi-autonomous expression whose evaluation could be skipped, repeated, deferred, etc.

Since the time of the evaluation of a block is flexible, the idea is to also permit the evaluation of a block **before** the evaluation of the method that contains it (ahead of time) with the special message (the obnoxious selector is a *mise en guarde*).

While classic time travel schemes cause known time continuum issues, the trick here is that the compiler (opal) statically detects such a message send and then evaluate the block itself and store the value as a literal of the Compiled Method (that is not a real literal, but a simple a shared object stored in the method).

### DIFF
--- a/src/AST-Core/RBMessageNode.class.st
+++ b/src/AST-Core/RBMessageNode.class.st
@@ -212,6 +212,12 @@ RBMessageNode >> isUnary [
 	^arguments isEmpty
 ]
 
+{ #category : #testing }
+RBMessageNode >> isValueAtCompileTime [
+
+	^ self selector = #valueAtCompileTime and: [ receiver isBlock ]
+]
+
 { #category : #accessing }
 RBMessageNode >> keywords [
 	^ self selector keywords.
--- a/src/AST-Core/RBMessageNode.class.st
+++ b/src/AST-Core/RBMessageNode.class.st
@@ -215,7 +215,7 @@ RBMessageNode >> isUnary [
 { #category : #testing }
 RBMessageNode >> isValueAtCompileTime [
 
-	^ self selector = #valueAtCompileTime and: [ receiver isBlock ]
+	^ self selector = #valueAtCompileTime and: [ receiver isBlock and: [ receiver hasArguments not ]]
 ]
 
 { #category : #accessing }
--- a/src/Kernel-Tests/ValueAtCompileTimeTest.class.st
+++ b/src/Kernel-Tests/ValueAtCompileTimeTest.class.st
@@ -40,16 +40,12 @@ ValueAtCompileTimeTest >> testValueAtCompileTimeNoLiteralBlock [
 
 { #category : #tests }
 ValueAtCompileTimeTest >> testValueAtCompileTimeNoParameters [
-
-	|v|
 	
 	"Parameters are not accepted"
 	
-	self flag: 'FIXME: add compiler errors or warnings'.
-	
-	v := [ :x | {x} ] valueAtCompileTime.
-	
-	self assert: v equals: { nil }
+	self flag: 'FIXME: add compiler errors or warnings instead of raising runtime error'.
+		
+	self should: [ [ :x | {x} ] valueAtCompileTime ] raise: Error
 ]
 
 { #category : #tests }
--- a/src/Kernel-Tests/ValueAtCompileTimeTest.class.st
+++ b/src/Kernel-Tests/ValueAtCompileTimeTest.class.st
@@ -1,0 +1,101 @@
+Class {
+	#name : #ValueAtCompileTimeTest,
+	#superclass : #TestCase,
+	#category : #'Kernel-Tests-Methods'
+}
+
+{ #category : #tests }
+ValueAtCompileTimeTest >> testValueAtCompileTime [
+
+	|v|
+	
+	v := [ 1+2 ] valueAtCompileTime.
+	
+	self assert: v equals: 3
+]
+
+{ #category : #tests }
+ValueAtCompileTimeTest >> testValueAtCompileTimeLocalTempsAreOK [
+
+	|v|
+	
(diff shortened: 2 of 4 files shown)


privat:
The point of the present PR was to gather some feedback (or even to stimulate discussion). Anyway, here are some issues I thought during the last week.

The main one is that it is a compiler hack, thus making the language as a whole more complex for the user and for the other tools.
Currently, in Pharo, magic behavior (and syntactic sugar) is very limited and does not alter the behavior of programs (it is mainly optimizations).

Another issue is related to the wording of the semantic. Compile time might not be the correct message to send (pun) to the user. Whereas the distinction of runtime and compiletime is mandatory in some other programming languages, Pharo is not restricted in that way.
Will the user really understand when such block are evaluated?

The 'when' issue can also be generalized if there are multiple compiletime blocks. What is the order of the evaluation? That question matters when blocks do side effects like assignments of global variables or mutations of global objects. Nested compiletime blocks can also be funny if we choose to evaluate the deepest first (or not).

Related to the previous issue is the binding (context) available at this compiletime. In the proposal self is bound to nil as there is no instance, but it is not clear what other variables are bound to.
Especially, the user may want to refer to shared variables of the class. Worse, a fancy user may want to refer to objects initialized in other compiletime blocks of the same method.

The result of the evaluation is also stored in a magical way without the control of the user. Although, they can store it at runtime in a temporary variable (outside the block) so it can feel like a poor man's constant static variable.

The last issue I see lies on the initial needs. What are the real-world use cases this proposal is trying to solve?
And how the existing language facilities do not solve them.

For instance, having shared variables (class variables) computed by the class initialize or through lazy initialization in a getter seem to be the classic way to address the main need, with a better time specification and a better storage specification.

Ping @Ducasse and @MarcusDenker

Ducasse:
I will check!

MarcusDenker:
On the CI when loading the Kernel tests (which contain the tests for valueAtCompileTime), we get an error 
"Instance of FullBlockClosure did not understand #pushLiteral:"

```
MetacelloNotification: Loading -> Kernel-Tests-tonel.1 --- tonel:///builds/workspace/est_and_branch_Pipeline_PR-11717/64/src --- cache
	Instance of FullBlockClosure did not understand #pushLiteral:
	FullBlockClosure(Object)>>doesNotUnderstand: #pushLiteral:
	OCASTTranslatorForValue(OCASTTranslator)>>emitValueAtCompileTime:
	OCASTTranslatorForValue(OCASTTranslator)>>visitMessageNode:
	RBMessageNode>>acceptVisitor:
	OCASTTranslatorForValue(RBProgramNodeVisitor)>>visitNode:
	OCASTTranslatorForValue(OCASTTranslator)>>visitNode:
	OCASTTranslatorForEffect(OCASTTranslator)>>visitAssignmentNode:
	OCASTTranslatorForEffect>>visitAssignmentNode:
	RBAssignmentNode>>acceptVisitor:
	OCASTTranslatorForEffect(RBProgramNodeVisitor)>>visitNode:
	OCASTTranslatorForEffect(OCASTTranslator)>>visitNode:
	[:each | self visitNode: each] in OCASTTranslatorForEffect>>visitSequenceNode: in Block: [:each | self visitNode: each]
	OrderedCollection>>do:
	OCASTTranslatorForEffect>>visitSequenceNode:
	RBSequenceNode>>acceptVisitor:
	OCASTTranslatorForEffect(RBProgramNodeVisitor)>>visitNode:
	OCASTTranslatorForEffect(OCASTTranslator)>>visitNode:
	OCASTTranslator>>visitMethodNode:
	RBMethodNode>>acceptVisitor:
	OCASTTranslator(RBProgramNodeVisitor)>>visitNode:
	OCASTTranslator>>visitNode:
	RBMethodNode>>generateIR
	RBMethodNode>>generate:
	RBMethodNode>>generateMethod
	OpalCompiler>>compileMethodFromAST
	OCMethodSemanticScope>>compileMethodFromASTBy:
```

Maybe we execute a Block that has arguments or temps without setting up the stack correctly?

MarcusDenker:
we could try to add a 

```
numArgs: compiledBlock numArgs;
```
to the creation of the block that we execute

privat:
Thanks @MarcusDenker using a better numArgs did something. But the tests did fail to run on osx. this is weird.

MarcusDenker:
I restarted the CI...

Ducasse:
I like your questions and this is why I'm not convinced that I would like to have compiled time blocks
because I do not really understand their semantics and use.

dionisiydk:
There are few missing bits in the implementation:
- what to do with failed blocks? Imagine "[2/0] valueAtCompileTime"
- what to do with non clean blocks? Imagine "[self someMessage] valueAtCompileTime". They should be forbidden.

In both cases you have to implement a proper feedback to the user as a syntax error. Imagine that you will load a bad code (with such broken block) to the image. It should follow all current hooks with faulty parsing/compilation.

Also I think that such feature should be modeled on the level of semantics analysis and not just ast-translator. And possibly it is required to solve the problems I described. 
For example special selectors ifTrue/ifFalse/etc are reified their meaning during the semantics analysis. Here you introduces another kind of special selectors which means another semantical element in the language.

Ducasse:
I would not integrate this in P11. I suggest to close this issue for now.

PR #11719, opened 2022-09-25
Make title bar layout dependent on new settings option and OS

### LINKED ISSUE #11718: Windows/Linux layout for the title bar elements
The label area elements in the Windows and Linux desktop environments look out of place: in Pharo windows, the close and resize window buttons are on the left, and the window menu button is on the right, while in system windows they are in the opposite order. This state of affairs creates an unpleasant user experience for beginners and occasional Pharo users.

I suggest adding a settings option that allows you to choose how the title bar elements will be laid out: macOS style (close button on the left), Windows/Linux style (close button on the right), and Platform Dependent.

**Expected behavior**

In the *Appearance > Morphic > Windows* section of the Settings Browser, there is a “Layout of label area elements” setting, which is a drop-down list with three options:

- Platform Dependent (selected by default)
- macOS
- Windows/Linux.

When choosing the Windows/Linux option, or when choosing the Platform Dependent option under Windows or Linux, in all Pharo windows, the window menu button is placed on the left, and the window close and resize buttons are placed on the right.

**_I have already done and am going to push changes_**

Screenshot

**Version information**
 - OS: Windows, Linux, macOS
 - Version: any
 - Pharo Version: 11

NSUSpray:
Fixes #11718

- Adds a new instance variable `labelAreaLayout` to the ThemeSettings class and adds access methods for it
- Makes the window label area configuring dependent on the value of this variable
- Adds a “Label area layout” option to the “Windows” section of the Settings Browser
- Bonus: adds a method `ThemeSettings >> #applyToDisplayedWindows` for applying settings to the all displayed SystemWindow instances

### DIFF
--- a/src/Polymorph-Widgets/UITheme.class.st
+++ b/src/Polymorph-Widgets/UITheme.class.st
@@ -1288,14 +1288,24 @@ UITheme >> configureWindowDropShadowFor: aWindow [
 UITheme >> configureWindowLabelAreaFor: aWindow [
 	"Configure the label area for the given window."
 
-	|padding|
+	| macOSLayout addCloseBox addCollapseExpandBoxes addMenuBox padding paddingStep |
+	macOSLayout := self labelAreaLayout = #macOS
+		or: [ self labelAreaLayout = #platformDependent and: Smalltalk os isMacOS ].
+	addCloseBox := [ aWindow hasCloseBox ifTrue: [aWindow addCloseBox ] ].
+	addCollapseExpandBoxes := [
+		aWindow hasCollapseBox ifTrue: [aWindow addCollapseBox ].
+		aWindow hasExpandBox ifTrue: [aWindow addExpandBox ] ].
+	addMenuBox := [ aWindow hasMenuBox ifTrue: [aWindow addMenuControl ] ].
 	padding := 0.
 	aWindow labelArea
 		addMorphBack: (Morph new extent: aWindow class borderWidth @ 0).
-	aWindow hasCloseBox ifTrue: [aWindow addCloseBox. padding := padding + 1].
-	aWindow hasCollapseBox ifTrue: [aWindow addCollapseBox. padding := padding + 1].
-	aWindow hasExpandBox ifTrue: [aWindow addExpandBox. padding := padding + 1].
-	aWindow hasMenuBox ifTrue: [padding := padding - 1].
+	macOSLayout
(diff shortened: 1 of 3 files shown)


NSUSpray:
I’m beginner in Pharo and I have some doubts about my decision.

The first is about the need for a `translated` message: the manual recommends it, but I don’t see it being used in the existing `PolymorphSystemSettings` class code.

The second is about the size of the `configureWindowLabelAreaFor:` method: it was initially rather big, but it became even a little larger. I couldn’t think of a good split into multiple methods and just refrained from unnecessary duplication.

The third is about the execution time of the `applyToDisplayedWindows` method: it should rarely be called, but it still takes a long time to execute. I wonder if it can be optimized and if there is another way to apply the new settings to existing windows.

Ducasse:
Thanks. translated is a left over. At some point we moved the translated message send into the system instead of forcing all clients to have to send it (and forget it). So one day we should check the current status.

For the rest I have to check the code. Now we can do it incrementally. Make it works, make it right, make is fast.

src/Polymorph-Widgets/UITheme.class.st
```diff
@@ -1288,14 +1288,24 @@ UITheme >> configureWindowDropShadowFor: aWindow [
 UITheme >> configureWindowLabelAreaFor: aWindow [
```
Ducasse, line 1288: To avoid all the checks we could have a little hierarchy OSWindowLayout and subclasses. But we can do that in a second step.

src/Polymorph-Widgets/UITheme.class.st
```diff
@@ -2369,6 +2381,11 @@ UITheme >> initializeForms [
 		at: #windowMenuOver put: self newWindowMenuOverForm
 ]
 
+{ #category : #accessing }
+UITheme >> labelAreaLayout [
```
Ducasse, line 2385: As a general comment (not about your code). I do not like that UITheme gets a setting.
The setting should put a configuration object to the UITheme which once configured should not refer to the setting object. This violates basic layering principle.

Ducasse: I would prefer to have a classVariable in UITheme that holds the layout and that the settings push the correct layout object or symbol for now in that UITheme. This way we can remove Setting and UITheme should continue to work. 
This is clear not the case in the current UITheme design and it sucks!

MarcusDenker:
Failing tests not related:

PR #11721, opened 2022-09-25
Add Point>>#isInRectangle

Ducasse:
- Add tests
- Add isInRectangle:
- Add executable comments to containsPoint:
- Add executable comment to isInRectangle:
- Recategorise is...Triangle...

### DIFF
--- a/src/Kernel/Point.class.st
+++ b/src/Kernel/Point.class.st
@@ -397,7 +397,20 @@ Point >> isFloatPoint [
 	^ x isFloat and: [ y isFloat ]
 ]
 
-{ #category : #geometry }
+{ #category : #testing }
+Point >> isInRectangle: aRectangle [ 
+	"Answer true whether the receiver is inside the argument (following Rectangle>>#containsPoint: semantics"
+	
+	"(50@50 isInRectangle: (0@0 corner: 100@100)) >>> true"
+	"(0@0 isInRectangle: (0@0 corner: 100@100)) >>> true"
+	"(0@10 isInRectangle: (0@0 corner: 100@100)) >>> true"
+	"(100@100 isInRectangle: (0@0 corner: 100@100))>>> false"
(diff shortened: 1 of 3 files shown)


src/Kernel/Point.class.st
```diff
@@ -397,7 +397,20 @@ Point >> isFloatPoint [
 	^ x isFloat and: [ y isFloat ]
 ]
 
-{ #category : #geometry }
+{ #category : #testing }
+Point >> isInRectangle: aRectangle [
```
astares: I would call it 

 #is**Inside**Rectangle:
 
 to be in alignment with
 
  #is**Inside**Circle:

src/Kernel/Point.class.st
```diff
@@ -397,7 +397,20 @@ Point >> isFloatPoint [
...
+	"(0@0 isInRectangle: (0@0 corner: 100@100)) >>> true"
+	"(0@10 isInRectangle: (0@0 corner: 100@100)) >>> true"
+	"(100@100 isInRectangle: (0@0 corner: 100@100))>>> false"
+	"(100@90 isInRectangle: (0@0 corner: 100@100))>>> false"
+	
+	^ aRectangle containsPoint: self
+]
+
+{ #category : #testing }
 Point >> isInsideCircle: a with: b with: c [
```
astares, line 401: a, b, c as argument names is not very descriptive. Could we have a comment what they depict if one does not have the mentioned paper at hand?

Ducasse: I have no idea. This is not me that wrote it. 
I just moved it to the testing protocol.

src/Kernel/Point.class.st
Ducasse: Good suggestion. I will fix this.

astares: Thanks!

MarcusDenker:
Failing tests not related:

PR #11722, opened 2022-09-26
11720-Improve-ReUtilityMethodsRule 

### LINKED ISSUE #11720: Improve ReUtilityMethodsRule
List methods that have one or more arguments and do no refer to self or an instance variable. These methods might be better defined in some other class or as class methods.

report error for 
```
Point >> isInRectangle: aRectangle

     ^ aRectangle containtsPoint: self
```
while clearly the method do use self!!!

MarcusDenker:
Do not check for self sends, but if the method reads self

fixes #11720


MarcusDenker:
Failing tests not related:

PR #11726, opened 2022-09-27
11725-Startup-Slow-in-OSX-previous-to-Monterrey

### LINKED ISSUE #11725: Startup Slow in OSX previous to Monterrey
OSX is slow to resolve the local address (it performs a DNS request for the hostname).
That takes around 5 seconds.
Pharo should not relly on the local address in the startup.
Ombu and GlobalIdentifier uses it.

https://apple.stackexchange.com/questions/175320/why-is-my-hostname-resolution-taking-so-long

tesonep:
- Delaying the generation of the UUID for the computer when it is not loaded
- Using the OmTimeStamp strategy instead of the UUID (it is not random)

### DIFF
--- a/src/EpiceaBrowsers/EpSettings.class.st
+++ b/src/EpiceaBrowsers/EpSettings.class.st
@@ -86,7 +86,7 @@ EpSettings class >> storeNameStrategySettingsOn: aBuilder [
 		description: 'Choose how will be the Epicea logs named.';
 		parent: #epicea;
 		target: self;
-		default: OmRandomSuffixStrategy;
+		default: OmTimeStampSuffixStrategy;
 		domainValues: OmSessionStoreNamingStrategy availableStrategies.
 ]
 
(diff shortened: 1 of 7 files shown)


MarcusDenker (approved):

demarey (approved):
I would like to see a comment explaining why we now avoid to use UUIDs:
UUID generation somehow leads to a FQDN resolution on OS X. This call implies a process block of around 5 seconds on many OS X systems.

tesonep:
~~I have found an issue, please do not integrate yet. I will fix it~~
I fixed it, I need to ensure the computer and secret UUID were generated before saving in an empty machine.

MarcusDenker:
Failing tests:
NoUnusedVariablesLeftTest>>testNoUnusedTemporaryVariablesLeft

the following methods have unused temporary variables and should be cleaned: an OrderedCollection(GlobalIdentifierTest>>#testBackwardCompatibility4)

PR #11737, opened 2022-09-30
Class definition popup uses the default class syntax (fluid with the new default)

### LINKED ISSUE #11622: Class definition pop-up uses old class syntax
When one executes code from a Playground with a class that does not exists, a pop-up appears.
If one selects the option **Define new class**
Capture d’écran 2022-08-31 à 21 00 04

The old class definition syntax is used.
Capture d’écran 2022-08-31 à 21 00 13

privat:
Fix #11622


MarcusDenker:
failing test not related

MarcusDenker (approved):

PR #11761, opened 2022-10-05
Improve compiler plugin implementation for P11

### LINKED ISSUE #8113: Improve Compiler plugin implementation

astares:
Fix #8113


astares:
Same code (which is now for Pharo 11) had been reviewed by @guillep in PR https://github.com/pharo-project/pharo/pull/8114 which was back then a PR against Pharo 10.

guillep (approved):

PR #11765, opened 2022-10-07
Move RBReplacePreview to NautilusRefactoring since it inherits from p…

Ducasse:
…resenter and core refactoring should not depend on UI!


Ducasse:
Broken tests are not related

PR #11767, opened 2022-10-07
Remove obsolete code in RBConcition - left over of the PhD of D. Robe…

Ducasse:
…rts.
After discussion with J Brant because I could not understand how it was used. I removed this unused declaration of type of condition.


Ducasse:
Tests are unrelated.

PR #11772, opened 2022-10-10
simplify Closure>>#arguments by using #collect:

MarcusDenker:
simplify #arguments by using #collect:

### DIFF
--- a/src/Kernel/Context.class.st
+++ b/src/Kernel/Context.class.st
@@ -292,11 +292,7 @@ Context >> activeOuterContext [
 Context >> arguments [
 	"returns the arguments of a message invocation"
 	
-	| arguments numargs |
-	numargs :=  self compiledCode numArgs.
-	arguments := Array new: numargs.
-	1 to: numargs do: [ :i | arguments at: i put: (self tempAt: i) ].
-	^ arguments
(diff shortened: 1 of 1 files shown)


src/Kernel/Context.class.st
```diff
@@ -292,11 +292,7 @@ Context >> activeOuterContext [
 Context >> arguments [
 	"returns the arguments of a message invocation"
 	
-	| arguments numargs |
-	numargs :=  self compiledCode numArgs.
-	arguments := Array new: numargs.
-	1 to: numargs do: [ :i | arguments at: i put: (self tempAt: i) ].
-	^ arguments
+	^(1 to: self numArgs) collect: [:i | self tempAt: i]
```
privat, line 295: I believe that `to:do:` is inlined? Do we feel the loss of this optimization here?
But maybe, since we are dealing with Context, and possibly metaprograming or debuging (who else cares about a collection of the arguments of a method?), the loss is probably negligible.

MarcusDenker: Yes, it is less optimized (it creates the intermediate interval, uses a real closure). But it is not used anywhere right now  (the debugger does not call it to show the stack)

privat (approved):

PR #11774, opened 2022-10-10
Multiline Executable Comments

privat:
This set of commits extends and simplifies executable comments.

First, it extends executable comments to the whole comment block, allowing multi-line comments.
This is the original intended behavior, since single-line expression is a strong limitation on readability, whereas readability is a required quality of examples.
Moreover, a double click on the inner part of the double quote selects the whole comment block, making DoIt inspections easy.

Obviously, this breaks existing executable comments with multiple expressions or which include text.
But surprisingly, there were not a lot of them (~15) and were easy to fix. Just add a couple of `"` to separate expressions and text.

Second, it handles the executable expressions as a whole, making `>>>` a standard binary operator.
Previously, the `>>>` was explicitly and manually separated. Making both left, and right parts parsed, highlighted and evaluated independently.
This previous behavior has three issues:
* it made `>>>` a syntactic construct with a lower precedence, lower than keyword message send. Whereas `>>>` in a full expression has a standard binary infix operator precedence. Thus making *DoIt* and automatic testing behaving differently.
* It prevented the left and right parts to share context, especially temporary variables needed by the example.
* It was bloated code, because handling both parts separately was superfluous, and because the parser already knows how to separate the operands of a binary infix operator.

The change of precedence shows that ~80 tests were in fact broken, but accepted by DrTest and CI testing nevertheless because of the weird precedence of `>>>` that was used.

### DIFF
--- a/src/Calypso-SystemQueries/ClyLocalClassScope.class.st
+++ b/src/Calypso-SystemQueries/ClyLocalClassScope.class.st
@@ -43,11 +43,11 @@ ClyLocalClassScope class >> metaLevelForHierarchyOf: aClass [
 	"Subclasses should decide what meta level of given class
 	should be used to retrieve/build hierarchy.
 	For example superclass hierarchy of ProtoObject class 
-	can stop at it according to instance side hierarchy 
-		ProtoObject superclass >>> nil
-	Or it can follow full superclass chain which will ends at Object and ProtoObject
-		ProtoObject class superclass >>> Class
-	This method adds such decision to concrete kind of local scope
+	can stop at it according to instance side hierarchy" 
+		"ProtoObject superclass >>> nil"
+	"Or it can follow full superclass chain which will ends at Object and ProtoObject"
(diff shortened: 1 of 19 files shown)


privat:
Tests failed because DrTests did not show me metaclasses. Added a commit to fix that.

That raises an interesting question: how do the current CI collects and runs the executable comments?

Edit: the answer is the method `DocCommentsReleaseTest class>>#buildSuite` that builds a pseudo test suite made of all the executable comments.

MarcusDenker:
DocCommentsReleaseTest class>>#buildSuite creates a "virtual" test suite

privat:
Can I assume the remaining failure is unrelated to the PR?

MarcusDenker:
yes, very odd failure just on win:

- OmSessionStoreTest(OmStoreTest)>>testEntryForAbsentEntry

Ducasse:
Cool today it was budget day...

MarcusDenker (approved):

PR #11776, opened 2022-10-10
11775 feature request adding breakpoint groups and break on count breakpoints

Max-Zurbriggen:
Created BreakpointGroup class with functionality
Added break on count as breakpoint functionality (only backend and API, no UI yet)

Did some code cleanup, including an adjustment to the "once" method which now only returns the state (replacing the "oneShot" method) instead of setting the variable, instead use "once: true" now to align it with the rest of the API. (This will require a change in a test case in https://github.com/pharo-spec/ScriptableDebugger)


src/Reflectivity/Breakpoint.class.st
```diff
@@ -288,14 +316,36 @@ Breakpoint >> enable [
...
+{ #category : #accessing }
+Breakpoint >> group [
+	"returns the group of the breakpoint"
+	^ group.
+]
+
+{ #category : #api }
+Breakpoint >> group: aBreakpointGroup [
+	"puts the breakpoint in a new group and sets the group variable to that group"
+	group removeBreakpoint: self.
```
MarcusDenker, line 330: as the group uses a Set, I think we do not need to remove before adding (as the set will just ignore if we add twice)

MarcusDenker (approved):
- it is really good to change #oneShot to use the term "once", the oneShot term was leftover when we thought that we could remove the breakpoint instead of disabling it for once breakpoints, which was not a good idea.

- the removal from the group is not needed as it uses sets (so we can just add the same multiple times with no problem). Not a huge problem, does not hold up merge

src/Reflectivity/Breakpoint.class.st
Max-Zurbriggen: this is to remove the breakpoint from its old group, so it is necessary since otherwise multiple groups (each with their own set) would contain that breakpoint. At the moment a breakpoint is only intended to be in one group at once (except for the "All Breakpoints" group).

MarcusDenker: ah, ok

PR #11778, opened 2022-10-11
Update comment to reflect refactoring actions

balsa-sarenac:
`RBRemoveAllSendersRefactoring` has different comment than the command calling it. Updated comments not to mislead someone.

### DIFF
--- a/src/SystemCommands-MessageCommands/SycRemoveAllSendersCommand.class.st
+++ b/src/SystemCommands-MessageCommands/SycRemoveAllSendersCommand.class.st
@@ -1,5 +1,5 @@
 "
-I am a command to launch the Remove All Senders Refactoring: the call to this method in all senders is removed (as long as possible)
+I am a command to launch the Remove All Senders Refactoring: removes all possible senders from a method.
 "
 Class {
 	#name : #SycRemoveAllSendersCommand,


Ducasse:
I think that the difference between pushup and pullup is that push up is from the subclass to the superclass 
while pull up is from the superclass point of view.
Can you check this aspect?

balsa-sarenac:
Thanks for the clarification, I was not aware of the terminology. I will modify the PR so it only integrates comment fix.
Also, similar thing is with parameters and arguments (AddParameter refactoring, but AddArgument command) is there something similar in that scenario as well?

Ducasse:
Normally we say that a parameter is the name of the method and that that the argument is the actual value. 
So from a refactoring point of view we could have only parameter and in addParametre the system should prompt the user for a default value = argument.

src/SystemCommands-MessageCommands/SycRemoveAllSendersCommand.class.st
```diff
@@ -1,5 +1,5 @@
 "
-I am a command to launch the Remove All Senders Refactoring: the call to this method in all senders is removed (as long as possible)
+I am a command to launch the Remove All Senders Refactoring: removes all possible senders from a method.
```
demarey, line 2: The comment is still not clear for me, especially `removes all possible senders from a method.`
Maybe something like: `remove all message sends to this method` ?

balsa-sarenac: This comment was inconsistent with the refactoring it uses (the `RBRemoveAllSendersRefactoring`).
I have changed the comment so that it is the same as `RBRemoveAllSendersRefactoring`.
What it does is it removes all (that it possibly can) messages that are sent from the source code of this message.

Below is an example taken from `RBRemoveAllSendersRefactoring` class comment. 

Example Script
----------------
```smalltalk
| refactoring options |
refactoring := RBRemoveSenderRefactoring 
   remove: (90 to: 105) "node position to be removed "
   inMethod: #caller1
   forClass: RBRefactoryTestDataApp.
options := refactoring options copy.
options at: #inlineExpression put: [:ref :string | false].
refactoring options: options.
refactoring execute.
```
Before refactoring:
```smalltalk
RBRefactoryTestDataApp >> caller1
 | anObject |
 anObject := 5.
 self called: anObject + 1
  on1: 
   [:each | 
   each printString.
   ^anObject]
```
After refactoring (notice that the call to printstring was removed):
```smalltalk
RBRefactoryTestDataApp >> caller1
 | anObject |
 anObject := 5.
 self called: anObject + 1
  on1: 
   [:each | 
   ^anObject]
```

So both comments should be updated to reflect the usage.

PR #11782, opened 2022-10-11
Reflectivity: more tests for multiple links

MarcusDenker:
while tracking down a bug in client code we wanted to be sure that this (combining before and instead) is working on the level of Reflectivity. It does!

Adding the tests as examples and to make sure it will not break


MarcusDenker:
as these are just green tests, I merge

PR #11783, opened 2022-10-11
Categorize refactorings and transformations

balsa-sarenac:
Some refactorings are not true refactorings but instead code generations or simply transformations. Added tag 'Transformation' to those refactorings.
Also, added a `kind` class method that tells whether an refactoring is a refactoring or transformation.


privat:
For my knowledge, what criterion do you use to distinguish both kinds?
Is there a precise definition of *true* refactoring? (and *true* transforming?)

Ducasse:
No behavioral changes!

PR #11784, opened 2022-10-11
Syntactic check a single >>> on executable comments

privat:
This PR is a follow-up of  #11774

Since executable comments are multiline, it could be easy to mistakenly put multiple `>>>` in a single comment block. Especially since the previous behavior was to process each line independently. e.g.

```
"1+2 >>> 3.
3+4 >>> 7"
```

The behavior proposed here is to reject such code and explicitly enforce the following rule:

> to be syntactically valid as an executable comment, there should be exactly one single triple > message send, and it should be the last statement.

In the existing code base (~2000 executable comments), there were only two errors; and some couples of `"` fixed both.

Note: there is no commit related to the UI yet, they may come in a future PR.

### DIFF
--- a/src/Kernel/AsciiCharset.class.st
+++ b/src/Kernel/AsciiCharset.class.st
@@ -181,9 +181,9 @@ AsciiCharset class >> maxValue [
 
 { #category : #casing }
 AsciiCharset class >> toLowercase: aCharacter [
-	"(AsciiCharset toLowercase: $A) >>> $a.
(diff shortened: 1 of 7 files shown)


MarcusDenker:
I added a label [PharoDocComments] and tagged some open issues with that

Ducasse:
Yes I like this 

```
to be syntactically valid as an executable comment, there should be exactly one single triple > message send, and it should be the last statement.
```

MarcusDenker (approved):

PR #11785, opened 2022-10-11
Pharo10: CI: when getting pharo version/prefix on PR, do not use --first-parent

privat:
Backport of #11780 on pharo10.

Let's see the if CI is happy


MarcusDenker:
merged, as the Pharo11 PR looks good

PR #11788, opened 2022-10-12
ClassParser: CDVariableNode  should use AST of the variable definition

### LINKED ISSUE #6410: Slot definitions with message cascades in them break class definition parser
And then classes that contain them cannot be edited.

**To Reproduce**
Create slot subclass with `#printOn:` defined as follows:
 ```
printOn: aStream
	aStream 
		nextPutAll: self class name;
		nextPutAll: ' new yourself; yourself'.
```
Add this slot to a class; save; to edit class definition. 

**Version information:**
Pharo 8 and 9

**Additional context**
Now this example is silly, but I have slots with many parameters, that are also optional, and in definition only parameters with non-default values are shown.
And now I cannot edit definitions of classes that use them.

MarcusDenker:
This PR  changes CDVariableNode to not hard-code one initialization message but instead store the AST of the Slot.

The AST has all the information we need, see #testComplexSlots

small cleanup
- remove not needed isGlobalClassNameBinding which is not needed -


MarcusDenker:
Note: this PR does not change anything from the "big picture": the variable definition is parsed, not evaluated. We just read the data from the AST instead of creating a special (very limited) datastructure.

Ducasse:
Thanks for the information.

Ducasse:
I will read when I have time.

PR #11789, opened 2022-10-13
CopyClassRefactoring copies class comment as well

### LINKED ISSUE #11520: Copy class refactoring should copy the class comment
When copying a package the classes lose their comments.

balsa-sarenac:
CopyClassRefactoring is a composite refactoring that uses InsertClassRefactoring.
Extend InsertClassRefactoring with the option to specify class comment for the inserted class.

Fixes: #11520 

Note: this will also copy the demo template. Problem with this is that pharo won't warn you about your class not having a comment. I tried to fix this one as well, but couldn't figure it out completely. If anyone has any suggestions, let me know.

### DIFF
--- a/src/Refactoring-Core/RBCopyClassRefactoring.class.st
+++ b/src/Refactoring-Core/RBCopyClassRefactoring.class.st
@@ -68,7 +68,8 @@ RBCopyClassRefactoring >> copyClass [
 		addClass: className
 		superclass: aClass superclass
 		subclasses: #()
-		category: self category).
+		category: self category
+		comment: aClass comment).
 ]
 
 { #category : #copying }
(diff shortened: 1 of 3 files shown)


Ducasse:
What is the demo template?

balsa-sarenac:
The starting template that gets automatically generated on class creation. Sorry for the bad word choice.

Ducasse:
For the starting class template we will have to look on the side of the ClassParser because we could say that by default we use the Fluid syntax and we can check how to generate it without hardcoding a macro but this is not related to refactorings.

MarcusDenker:
For the class comment: You can check with ClassDescription>>#hasComment if the class really has a comment. It returns false if asking the comment would return the template.

PR #11791, opened 2022-10-13
add missing utf8StringFromCString methods

estebanlm:
as is said... we deprecated readStringUTF8 in ExternalData but not its counterpart in ByteArray

### DIFF
--- a/src/UnifiedFFI/ByteArray.extension.st
+++ b/src/UnifiedFFI/ByteArray.extension.st
@@ -141,6 +141,11 @@ ByteArray >> readString [
 { #category : #'*UnifiedFFI' }
 ByteArray >> readStringUTF8 [
 
+	self 
+		deprecated: 'Please use #utf8StringFromCString instead' 
+		transformWith: '`@receiver readStringUTF8' 
+						-> '`@receiver utf8StringFromCString'.
+
 	^ (ExternalData fromHandle: self type: ExternalType string)
 		  utf8StringFromCString
 ]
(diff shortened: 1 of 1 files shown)


estebanlm:
to be honest, @demarey, I am kind of annoyed by this change (I know, I should have complained before :P).
As I see it, this is just a name change, I do not see the gain on it (not really very different utf8StringFromCString from readStringUTF8). 
The real problem is that it breaks backward compatibility for, IMO, no real gain. 
See this: https://github.com/pharo-vcs/libgit2-pharo-bindings/releases/tag/v3.0.0
and like this... how many other libs?
We need to think careful this changes, sometimes just deprecate is not enough.

<blockquote><div> GitHub</div><div><strong><a href="/pharo-vcs/libgit2-pharo-bindings/releases/tag/v3.0.0">Release v3.0.0 · pharo-vcs/libgit2-pharo-bindings</a></strong></div><div>This version uses utf8StringFromCString method family, which is just found in Pharo 11.
Hence, everybody wanting to backport from here needs to add such method (a rename of readStringUTF8) to their...</div></blockquote>

Rydier:
The new name makes no sense to me.
Strings in Pharo are just strings, not utf8strings.
What we're doing (presumably), is reading a string from the contents, interpreted as utf8 bytes...
What was wrong with the old name?

demarey:
@estebanlm @ this change was part of a bigger change: https://github.com/pharo-project/pharo/pull/11668
The important part was to introduce `#bytesFromCString`.  Then `#utf8StringFromCString` is a utility method to get the bytes decoded with UTF8 decoder.
`#readString` had to be deprecated as it was *always* returning an UTF8 decoded String while `#bytesFromCString` now returns a byte array.
`#readStringUTF8` may have been kept. maybe the name is clearer than `#utf8StringFromCString` ?

Ducasse:
To me the new name is clearer.

guillep:
To me a clearer name is `decodeStringFromUtf8CString` if you want. `utf8StringFromCString` may be interpreted as "it returns a utf8 encoded string", which is wrong.

estebanlm:
my opinion was that it was a change with so little gain I don't think it worth breaking backward compatibility.
if we want better names, Guille's proposal seems better than the newer (and the newer is in fact worst than readStringUTF8, since that old one at least was indicating there was an operation happening, the read).

Ducasse:
So may be we should change it because I thought it was exactly what guille mentioned as wrong :)

PR #11795, opened 2022-10-13
Fix #11794: Enhance CI to use the right VM version

### LINKED ISSUE #11794: Enhance CI to use the right VM version
PR https://github.com/pharo-project/pharo/pull/11678 introduced the ability to run the latest VM in odd PR builds.
However, we are still hardcoding to use the VM for Pharo 9 (either stable or latest pharo9).
Right now that is not a problem because it's backwards compatible.

Right now
```bash
TEST_VM_VERSION="90"
```
We should change that to use the VM of the correct Pharo version.

guillep:
Make PHARO_VM_VERSION match the Pharo version
Make test scripts use PHARO_VM_VERSION

### DIFF
--- a/bootstrap/scripts/envversion.sh
+++ b/bootstrap/scripts/envversion.sh
@@ -97,7 +97,7 @@ function set_version_variables() {
 		set_version_pull_request_variables
 	fi
 	
-	PHARO_VM_VERSION="90"
+	PHARO_VM_VERSION=${PHARO_SHORT_VERSION}
 	
 	# Prefix the commit hash with a g, to be compatible with git-describe commit hashes
 	# https://git-scm.com/docs/git-describe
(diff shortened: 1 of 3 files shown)


demarey:
CI is failing:
```bash
++ git describe --long --tags --first-parent

++ cut -d- -f 1

++ cut -c 2-

++ cut -d. -f 1-2

fatal: not a git repository (or any of the parent directories): .git
```

guillep:
Yes, it seems that the script `set_version_variables` requires access to the git repository to extract the version number.
I'll see how I fix this.

guillep:
imagen

Now it seems good! Failing tests seem unrelated...

guillep:
Checked, I re-run this PR build a could of times. Latest VM was used in odd builds, stable VM in even builds, good to go!

Issue #11799, opened 2022-10-16
Offers nice way to compute file size


Ducasse:
Magritte has a nice little helper that produces file size.

Ducasse:
I adapt it and rename it.

```
IntegerTest >>testFileSizeString
	self assert: 1000 asFileSizeString equals: '1000 B'.
	self assert: 1024 asFileSizeString equals: '1 KB'.

	self assert: (1000 * 1000) asFileSizeString equals: '976 KB'.
	self assert: (1024 * 1024) asFileSizeString equals: '1 MB'.

	self assert: (1000 * 1000 * 1000) asFileSizeString equals: '953 MB'.
	self assert: (1024 * 1024 * 1024) asFileSizeString equals: '1 GB'.

	self assert: (1000 * 1000 * 1000 * 1000) asFileSizeString equals: '931 GB'.
	self assert: (1024 * 1024 * 1024 * 1024) asFileSizeString equals: '1 TB'.
```

```
Integer >> asFileSizeString
	"Return the receiver as a string with the adequate file size identified, e.g. '50 KB'"
	
	"(1000 * 1000 * 1000) asFileSizeString >>> '953 MB'"
	"(1024 * 1024 * 1024) asFileSizeString >>> '1 TB'"
	
	#( 'B' 'KB' 'MB' 'GB' 'TB' 'PB' 'EB' 'ZB' 'YB' )
		inject: self
		into: [ :value :each |
			value < 1024
				ifFalse: [ value // 1024 ]
				ifTrue: [ ^ value asString , ' ' , each ] ]
```

svenvc:
Note that there already is Integer>>#humanReadableSIByteSize

Ducasse:
Ok I did not know!

`"Print a string with an SI binary unit represation of myself."`
-> `Return a string with an SI ...` 


Ok we should add a decent comment without obvious typo, and an executable comment and a test.

Ducasse:
I will update my PR

Ducasse:
strange my PR extension is not shown.

MarcusDenker:
https://github.com/pharo-project/pharo/pull/11800 was merged

PR #11800, opened 2022-10-16
#11799 a nice little help method for converting integers into corresp…

Ducasse:
…onding file size.

### DIFF
--- a/src/FileSystem-Core/Integer.extension.st
+++ b/src/FileSystem-Core/Integer.extension.st
@@ -1,28 +1,27 @@
 Extension { #name : #Integer }
 
 { #category : #'*FileSystem-Core' }
-Integer >> humanReadableSIByteSize [
-	"Return the receiver as a string with SI binary (International System of Units) file size, e.g. '50 KB'"
+Integer >> humanReadableByteSizeString [
+	"Return the receiver as a string with the adequate file size identified, e.g. '50 KB'"
 	
-	"(1000 * 1000 * 1000) humanReadableSIByteSize >>> '953 MB'"
-	"(1024 * 1024 * 1024) humanReadableSIByteSize >>> '1 TB'"
+	"(1000 * 1000 * 1000) humanReadableByteSizeString >>> '953 MB'"
+	"(1024 * 1024 * 1024) humanReadableByteSizeString >>> '1 GB'"
 	
-	^ String streamContents: [ :s|
-		self humanReadableSIByteSizeOn: s ]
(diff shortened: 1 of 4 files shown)


svenvc:
Note that there already is Integer>>#humanReadableSIByteSize

Ducasse:
Now this is not the same

```
"(1000 * 1000 * 1000) humanReadableSIByteSize >>> '1.00 GB'"
"(1024 * 1024 * 1024) humanReadableSIByteSize '>>> 1.07 GB'"

"(1000 * 1000 * 1000) asFileSizeString >>> '953 MB'"
"(1024 * 1024 * 1024) asFileSizeString >>> '1 GB'"
```

So which one is correct.

marcor:
First one is correct, second should read KiB, MiB, etc.

svenvc:
Yeah, this is annoying and confusing. In standard SI terminology, the K, M, G are multiples of 1000, while in computer science they are sometimes multiples of 1024. And indeed, the KiB notation exists to make this clear, but not many really use this in practice, relying more on context.

Ducasse:
So I do not really know what to do. Personally on my harddisc I want 1024 and not 1000 so we could have 

```
(1000 * 1000 * 1000) asFileSizeString
=> 
(1000 * 1000 * 1000) humanReadableByteSize
```
and I would not add the i because nobody will ever use that.

Ducasse:
Now what I do not like about humanReadableByteSize and humanReadableSIByteSize is that it does not return a number so this is not a size but a string. This is why my original method was asFileSizeString.
I can rename humanReadableByteSize into humanReadableByteSizeString
and 
humanReadableSIByteSize into humanReadableSISizeString

Ducasse:
Because I do not understand the byte part in humanReadableSIByteSizeString while I can understand it in humanReadableByteSizeString

marcor:
SI is synonymous of "metric", I don't think it would be clear what "size" refers to. For comparison Cuis uses these method names:  printStringAsBytes and printStringAsBytesDecimal. Apparently in Pharo "humanReadable" is only used in 2 classes, Integer and Duration.

Ducasse:
I should say that I'm neither fond of human readable nor print. I would prefer as...
Now the Cuis names are not that good either.

marcor:
I'll throw asByteSizeString in the mix. :D

Ducasse:
Ok so it would be asByteSizeString and asSIByteSizeString or asByteSizeString and asSISizeString.

Ducasse:
Finally I let the methods as there are named. We can decide later.

Issue #11803, opened 2022-10-18
Return value of inspect should be the receiver and not a SpWindowPresenter 


SabineMa:
a long time ago....nearly 10 years now, there was a discussion in nabble:

http://forum.world.st/inspect-returns-iOf-SystemWindow-and-not-receiver-td4692399.html#a4692587

I was very surprised about a change in pharo - > inspect did not return the receiver itself but  an instance of SystemWindow.

We are moving from Pharo 8 to Pharo 10 this days. And what do I see today?

inspect did not return the receiver itself but  an instance of SpWindowPresenter.
Who is interested in the SpWindowPresenter when doing an inspect?

I will not repeat my arguments here, they are all discussed in the dabble thread.
Most people did accept my opinion and it was changed in 2013.

I will change this in my code by creating my own inspect method but I think the community should discuss this again. I think this is a bad solution. (I start to repeat my arguments...other Smalltalk dialects, debugging etc...)

I am amused about the code carousel/Back into Future.

I did not change my opinion. 

Code as it is:
```
inspect
	"Create and schedule an Inspector in which the user can examine the receiver's variables."
	 ^ Smalltalk tools inspector inspect: self.
```

Code as it should be:

```
inspect
	"Create and schedule an Inspector in which the user can examine the receiver's variables."
	  Smalltalk tools inspector inspect: self.
     ^ self
```

Ducasse:
Thanks sabina.
I agree with you :).

Ducasse:
Send a PR.

jordanmontt:
+1 (first time seeing this). I liked Marcus answer on the forum:

> So maybe we can
        -> add #inspector that returns the tool
        -> make inspect be like it was before (with documentation).
        Marcus

so,

```st
inspect
    "Create and schedule an Inspector in which the user can examine the receiver's variables."
    Smalltalk tools inspector inspect: self.
    ^ self
```

and

```st
inspector
    "Create and schedule an Inspector in which the user can examine the receiver's variables."
    ^ Smalltalk tools inspector inspect: self
```

SabineMa:
I just got a new pharo 11 image to write the test as Stephane suggested. Then I could not reproduce the problem. 
3 inspect returns 3. Everything is fine, I can not explain what I did yesterday.
It seems that I was confused yesterday. Or I did an inspect on an inspect. 
(3 inspect inspect returns the window presenter).
So, this PR should NOT be integrated. SORRY for the noise (I'm embarrassed)

[...]

PR #11804, opened 2022-10-18
change inspect to return receiver / add inspector

SabineMa:
https://github.com/pharo-project/pharo/issues/11803

### DIFF
--- a/src/Tool-Base/Object.extension.st
+++ b/src/Tool-Base/Object.extension.st
@@ -18,9 +18,19 @@ Object >> browseHierarchy [
 
 { #category : #'*Tool-Base' }
 Object >> inspect [
-	"Create and schedule an Inspector in which the user can examine the receiver's variables."
+	"Create and schedule an Inspector in which the user can examine the receiver's variables.
+	Return the receiver."
 
-	^ Smalltalk tools inspector inspect: self
+	Smalltalk tools inspector inspect: self.
+	^ self
+]
+
+{ #category : #'*Tool-Base' }
+Object >> inspector [
+	"Create and schedule an Inspector in which the user can examine the receiver's variables.
+	Return the inspector window."
+
+	^ Smalltalk tools inspector inspect: self.
 ]
 
 { #category : #'*Tool-Base' }


jordanmontt (approved):

Ducasse:
Would be good to add a test to cover that we do not get this regression.

SabineMa:
I can add

testInspector
self assert: 1 inspector class equals: SpWindowPresenter

and

testInspect
self assert: 1 inspect equals: 1


Is this ok?
I am not sure to which class.
what is the condition that the regression will not occur?

src/Tool-Base/Object.extension.st
```diff
@@ -18,9 +18,19 @@ Object >> browseHierarchy [
...
+	^ self
+]
+
+{ #category : #'*Tool-Base' }
+Object >> inspector [
+	"Create and schedule an Inspector in which the user can examine the receiver's variables.
+	Return the inspector window."
 
-	^ Smalltalk tools inspector inspect: self
+	^ Smalltalk tools inspector inspect: self.
```
astares, line 33: Please no unnecessary dots at the end.

astares (changes requested):

jordanmontt:
Also maybe you could add a comment in the methods as a form of documentation. For example, in the `inspect`method saying that if you can the window object you should use `inspector`

Ducasse:
Hi sabina 
Yes this sounds good to me and they should be packaged with the same tools-tests or something like that.

SabineMa:
https://github.com/pharo-project/pharo/issues/11803#issuecomment-1283836624
Please do not integrate

SabineMa:
Everything is ok as it is. I was confused.

jordanmontt (commented):
As discussed in the issue, I also think we should add:

- deprecate basicInspectand renamed it to basicInspector. We still return the spec window.
- refactor inspectOnce to always return self.

SabineMa:
I already wrote the code but I have auth problems for doing the pull request. 
I have to manage other stuff with you product and will come back here asap.

jordanmontt:
Hello @SabineMa , did you have the time to commit your changes?
If not, tomorrow there is a Pharo sprint, this can be an issue for that :)

SabineMa:
closed because done here: https://github.com/pharo-project/pharo/pull/11886

PR #11824, opened 2022-10-24
Improve bootstrap.sh

### LINKED ISSUE #11822: Improve bootstrap.sh
The **bootstrap.sh** includes the following line:

image

that calls **envvars.sh** but the same script also calls **1-clean.sh** right after and **1-clean.sh** is again calling **envvars.sh**:

image

So we can optimize and only need to call it a single time. We should remove the first call in  **bootstrap.sh**

astares:
Fix #11822


MarcusDenker:
Strange: the CI did not run for this PR

PR #11825, opened 2022-10-24
Transition MoveMethodToClassSide refactoring to use driver object

balsa-sarenac:
This PR introduces base `RBDriver` class and `RBMoveMethodsToClassSideDriver`.
In order for driver to check preconditions a small refactoring was needed of the `MoveMethodToClassSideRefactoring`. Preconditions are split on `applicabilityConditions` and `breakingChangeConditionss` and calling preconditions for move method refactorings will just invoke these two methods:
```smalltalk
RBMoveMethodToClassSideRefactoring >> preconditions
     ^ self applicabilityConditions & self breakingChangeConditionss
```
Also, in `RBRefactoring` `performChanges` was extracted from `execute` method so that driver can use the refactoring API for applying changes to the image without breaking any of the refactorings.

*Note* this also fixes a bug where move to class side command was calling `MoveToClass` refactoring and would just move method to class side. Now, a method is moved to class side and a new instance side method is created that calls newly created class side method and thus doesn't break the application.


Ducasse:
Cool I want to see this. 
I will have a look as soon as I can. Probably tomorrow morning or later today.

Ducasse:
I do not know why but I cannot load the PR into my image because of an iceberg bug. I will ask for help tomorrow. 
I think that indeed precondition raising pop up should be move to the driver. After preconditions on applicability should be kept on the refactoring so that we can use the refactoring without the driver.

Ducasse:
I changed the title to remove the back tick and see if I can load the PR. 
I'm trying to load the branch from your repo

Ducasse:
Im wondering why checkBreakingChanges and checkApplicabilities are defined the RBDriver class.

Ducasse:
I would rename breakingChanges -> breakingChangeConditions because the method return conditinos and not changes
and may be applicability -> applicabilityCondition (just to be in sync with breakingChangeConditions).

Now I got a bit better why you defined checkBreakingChanges on the driver class.

balsa-sarenac:
The reason why you cannot load the image could be the reason the CI is failing.

jordanmontt:
Hello @balsa-sarenac @Ducasse

I just tried and I was able to checkout to this PR. You need to add the balsa fork as a remote repo and then checkout to the branch `transition-move-method-to-class-side-refactoring-to-use-driver-object`

I also run all the tests that have refactor in their name and everything is green.

@PalumboN Just tried also and it works for him too.

Ducasse:
Thanks. In fact I removed the back ticks from the title. 
Because I could load the branch directly from the repo. 
Now when loading in PR, iceberg should create the correct repo.

PR #11826, opened 2022-10-24
made rename method refactor "Spec2 compliant"

estebanlm:
it was using old api and in consequence it was broken with latest changes


Ducasse:
cool

PR #11830, opened 2022-10-25
Package 'Coverage' as a higher level API for code coverage

privat:
I propose an independent package that offers a simple API to compute code coverage and to deal with low-level stuff.
While it uses metalink internally, the user is not directly exposed to it.

Currently, compiled methods and sequences of statements (bodies of methods and block) can be tracked.
This is a compromise between efficiency and precision that is good enough precision to track covered paths.

Here is a quick usage:

```st
| collector coverage |
collector := CoverageCollector new. "Instantiate"
collector methods: Point methods, Rectangle methods.  "Configure with the methods to watch."
coverage := collector runOn: [ (1@1 corner: 2@2) center ].  "Setup, execute and teardown."
(coverage methods includes: Point>>#x) >>> true.  "Inspect the results"
(coverage nodes size > 10) >>> true. "Covered paths are also available"
```

Existing usage of metalinks in SUnit and DrTest are also updated to use this new package instead.

### DIFF
--- a/src/BaselineOfCoverage/BaselineOfCoverage.class.st
+++ b/src/BaselineOfCoverage/BaselineOfCoverage.class.st
@@ -1,0 +1,15 @@
+Class {
+	#name : #BaselineOfCoverage,
+	#superclass : #BaselineOf,
+	#category : #BaselineOfCoverage
+}
+
+{ #category : #baselines }
+BaselineOfCoverage >> baseline: spec [
+	<baseline>
+	spec
+		for: #common
+		do: [ 
+			spec
+				package: 'Coverage' ]
+]
--- a/src/BaselineOfCoverage/package.st
+++ b/src/BaselineOfCoverage/package.st
@@ -1,0 +1,1 @@
+Package { #name : #BaselineOfCoverage }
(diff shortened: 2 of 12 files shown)


privat:
@MarcusDenker can you review?

privat:
I tried to fix the package/baseline issues, but I do not know what I am doing (i should read https://github.com/pharo-open-documentation/pharo-wiki/blob/master/General/Baselines.md )

<blockquote><div> GitHub</div><div><strong><a href="https://github.com/pharo-open-documentation/pharo-wiki">pharo-wiki/Baselines.md at master · pharo-open-documentation/pharo-wiki</a></strong></div><div>Wiki related to the Pharo programming language and environment. - pharo-wiki/Baselines.md at master · pharo-open-documentation/pharo-wiki</div></blockquote>

astares:
Can you name the packages "Coverage-Core" and "Coverage-Core-Tests" ?
Would be better as we can/will have additional "Coverage-UI", "Coverage-Tools" or other later.

Ducasse:
Cool :)
Eventually I would like to have a branch level coverage.

MarcusDenker:
It loads now! I am not sure if the way we load it is "correct", it seems we do not use the BaseLine at all but instead add the package the baseline of SUnit and DrTest.

MarcusDenker (approved):
Looks good!

privat:
Pushed some improvements (and a bugfix)

PR #11831, opened 2022-10-25
fix OSWIndowDriver shutdown while saving

estebanlm:
the event loop needs to be stopped while saving the image, otherwise reopening the image after save (and not quit) will keep the event loop running, even when in non-interactive mode.

### DIFF
--- a/src/OSWindow-SDL2/OSSDL2Driver.class.st
+++ b/src/OSWindow-SDL2/OSSDL2Driver.class.st
@@ -231,8 +231,7 @@ OSSDL2Driver >> initialize [
 "		initVideo;
 		initJoystick;
 		initGameController;
-"		initEverything.
-	self setupEventLoop.
+"		initEverything
 ]
 
 { #category : #'window creation and deletion' }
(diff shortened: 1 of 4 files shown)


src/OSWindow-SDL2/OSSDL2Driver.class.st
```diff
@@ -361,33 +360,54 @@ OSSDL2Driver >> setGLAttributes: glAttributes [
...
+			forkAt: Processor lowIOPriority ] 
 		ifFalse: [ 
-			[ self eventLoopProcessWithoutPlugin ] 
-			forkAt: Processor lowIOPriority ].
-	].
+			self class hasPlugin 
+				ifTrue: [ 
+					[ self eventLoopProcessWithPlugin ] 
+					forkAt: Processor lowIOPriority ]
+				ifFalse: [
```
tinchodias, line 378: I wonder what's the history of this plugin. In P10 and 11, at least, I only got false from `OSSDL2Driver hasPlugin`.

Issue #11727, opened 2022-09-28
Double click and drag does not select whole words


privat:
**Bug description**

You can double-click on a word to select it.
But if you drag, the word is unselected and dragging around selects characters instead of whole words.

This behavior is inconsistent with most UI elements of common applications that deals with text (text editors, web browsers, terminal emulators, pdf viewers, etc.)

**To Reproduce**
1. Open a playground
2. Write `hello world`
3. Double click between the two `l` of `hello` and keep the mouse button down
4. See that the whole world `hello` is selected (this is good)
5. Move the cursor towards the `r` of `world` while maintaining the mouse button down
6. See that `hel` is unselected, only `lo` remains selected (this is bad)
7. See also that only `wor` is selected and that `ld` is unselected (this is also bad)

extra: try the same scenario in non pharo applications where there is selectable text.

**Expected behavior**

step 6: `hello` should remain selected
step 7: when entering `world` the whole word should be selected.

**Remark**

Using shift+control and selecting with arrows, have a whole-word behavior consistent with other application.

**Version information:**
 - OS: Debian testing 
 - Pharo Version: 11

StephanEggermont:
I don’t understand step 3. Double click ends with a mouse up. What is the exact behavior you see?

akevalion:
I got the the same result, also I compared with vscode, atom, html textarea editor.

privat:
@StephanEggermont 3.1. press left mouse button, 3.2. quickly release it, 3.3 quickly press it again. 3.4 keep it pressed (do not release it)

akevalion:
Another thing,

- Open a playground
- Write hello world
- Double click between the two l of hello and keep the mouse button down
- See that the whole world hello is selected (this is good)
- Move the cursor towards the r of world while maintaining the mouse button down
- you also can press `ESC` to cancel the cut event.

akevalion:
After review rubric code, Mouse double click only works after the second mouse up.
Then we have to modify the mouse double click event process to detect mouse double click with drag and drop events
Please check this class
`MouseClickState`

akevalion:
I have fixed the MouseClickState, then now it accepts doubleclick drag state, but the text editor still does not recognise the drag selection word

Ducasse:
Thanks a lot. 
This usability glitches are important!!!!

jordanmontt:
Hi, so as this issue was solved in #11838 , we should close it and also #11744 
@akevalion

PR #11838, opened 2022-10-28
11727 double click and drag does not select whole words

akevalion:
This pull request solves this issue: 
- https://github.com/pharo-project/pharo/issues/11727


akevalion:
Pair programming @privat

Ducasse:
May be this is impacting the ecompletion tests. We will fix it after.

guillep:
This PR is also creating another issue: now clicking on a tab will activate a double click (and maximise) when a second click happens outside the tab.

This happens at least in calypso when the first click happens on the tab, and the second tab happens in the text area.

PR #11846, opened 2022-11-02
Closes #11817. The class creation text suggestion in Calypso now uses the current cl…

### LINKED ISSUE #11817: When the system prompts the user for a class creation the package is empty
Imagine you refer in your package Pack1 in a method to class NotDefinedYet.
Calypso will ask you to know if this is a class, classVariable.
When you say that this is a class, the template does not use Pack1 but an empty string and 
this can lead to a mess.

maxwills:
...class package instead of an empy String. Closes #11817
Now the new class suggestion text template suggests as package the current class package instead as an empty String.


Ducasse:
Thanks! Sound good.

PR #11849, opened 2022-11-03
Fixes the problem related to backwards selection. Closes #11848

### LINKED ISSUE #11848: Selecting code in backwards sense produces wrong behavior in text editors.
**Bug description**
Selecting code in backwards sense produces wrong behavior of the text editors.

**To Reproduce**
1. Open a Playground
2. Write: Hello World
3. Put cursor right after the 'd', then click and drag until the word World is fully selected.
4. Press the delete key (or try to perform any action with the selected text, like copying to the clipboard).
5. See error: Nothing is deleted (Or copied).

**Expected behavior**
The selected text should be deleted (Or whatever action was performed)

**Version information:**
  - Pharo 11

**Additional context**
I identified the pr that introduced the bug and the related code. I'll propose a fix in a few minutes, which seems to work, but I'm not sure of the side effects.
Also, maybe other people are not experiencing this issue, can someone else confirm?

maxwills:
Closes #11848
Please take a look @akevalion.
It behaved well in my manual testings, although I am not sure about other side effects.

### DIFF
--- a/src/Rubric-Tests/RubEditingStateTest.class.st
+++ b/src/Rubric-Tests/RubEditingStateTest.class.st
@@ -1,0 +1,66 @@
+"
+A RubEditingStateTest is a test class for testing the behavior of RubEditingState
+"
+Class {
+	#name : #RubEditingStateTest,
+	#superclass : #TestCase,
+	#category : #'Rubric-Tests-Editing-Core'
+}
+
+{ #category : #tests }
+RubEditingStateTest >> testSelectionStart [
+
+	| m |
+	"Create a RubTextFieldArea which calls the tested method"
+	m := RubTextFieldArea new
+		     width: 300;
+		     maxLength: 30;
+		     yourself.
+	self assert: m string isEmpty.
+	m privateReplaceFrom: 1 to: 0 with: 'one two three'.
+
+	m selectFrom: 1 to: 3.
+	self assert: m selection asString equals: 'one'.
+	self assert: m selectionStart stringIndex equals: 1.
+
+	m selectFrom: 4 to: 0.
+	self assert: m selection asString equals: 'one'.
+	self assert: m selectionStart stringIndex equals: 1.
+
+	m selectFrom: -10 to: 1000.
+	self assert: m selection asString equals: 'one two three'.
+	self assert: m selectionStart stringIndex equals: 1.
+
+	m selectFrom: 100 to: -1000.
+	self assert: m selection asString equals: 'one two three'.
+	self assert: m selectionStart stringIndex equals: 1
+]
+
+{ #category : #tests }
+RubEditingStateTest >> testSelectionStop [
+
+	| m |
+	"Create a RubTextFieldArea which calls the tested method"
+	m := RubTextFieldArea new
+		     width: 300;
+		     maxLength: 30;
+		     yourself.
+	self assert: m string isEmpty.
+	m privateReplaceFrom: 1 to: 0 with: 'one two three'.
+
+	m selectFrom: 0 to: 3.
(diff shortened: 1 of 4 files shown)


Ducasse:
Oh great because I cannot remove a line with delete if I select it from the end of the line to the beginning.

akevalion:
It looks good for me. Please check this pr:@privat. We have modified the method `selectionStop` because when you select a word, the caret position is at the end of the selection and not in mouse cursor position, maybe a new possible issue

privat:
The change in `selectionStop`was in fact a bad idea, as it seems that it is used (as its name implies) to know the stop index of the current selection.

But I do not really understand the rest of the changes.

maxwills:
There essentially 3 things changed in the code. 
1. The selectionStop method (which is the main change that fixes the issue), but introduced a random error, which I'll try to explain in the next point.
2. The previous version of this line https://github.com/pharo-project/pharo/pull/11849/files#diff-698bdf9f6c9a52f01d072e30e9207555b35513082babd125e22f2dcd8cf8832fR2129
sometimes had the value of mark equal to 0 in some use cases that I could not reproduce reliably, which raises an exception. I just introduced the bounds checking, which prevents the issue and has no practical side effects.
3. In my manual testings I noticed excessive flickering of the selection visuals when doing the "double-click and drag" to select words. That is why I changed parts of the code to use a new "invisible deselect" that uses the existing selectInvisiblyFrom:to: ("Select the designated characters, inclusive.  Make no visual changes."). 
To be honest, I see an improvement in the flickering, probably, but might be just me, and I have no formal or automated way to test that.

Point 3 might be optionally removed from the commit.

<blockquote><div> GitHub</div><div><strong><a href="https://github.com/pharo-project/pharo/pull/11849">Fixes the problem related to backwards selection. #11848 by maxwills · Pull Request #11849 · pharo-project/pharo</a></strong></div><div>Closes #11848
Please take a look @akevalion.
It behaved well in my manual testings, although I am not sure about other side effects.</div></blockquote>

guillep:
Hi guys this is a really good catch! I just was using the latest P11 and it was very complicated with these issues...
Could I ask to add a couple of tests?

privat (approved):

Ducasse:
For the tests I have the impression that looking at the tests we did for the completion and caret position can be a good help.

maxwills:
The class in question (RubEditingState) had no unit test. I added a couple tests for the modified method. However, covering the complete list of methods with tests might take some work. Can we create a separated issue for that?

Ducasse:
This is cool. If you can open an issue it would be great and pointing to your first tests to encourage people.

guillep:
> The class in question (RubEditingState) had no unit test. I added a couple tests for the modified method. However, covering the complete list of methods with tests might take some work. Can we create a separated issue for that?

I think that adding a couple of tests for the modification is more than enough! Thanks a lot!
If we all add 2/3 tests each time we change something that was untested, eventually we will have no more fear to change it, and we will have hundreds of tests :)

Ducasse:
I do not get why the build does not work. I restarted it but I got the same problem and I think that this not linked with the changes.

Ducasse:
load : FluidClassBuilder-Tests-tonel.1
	postload : baseline [BaselineOfPharo] >> postload:package:
+ /builds/workspace/est_and_branch_Pipeline_PR-11849/64/bootstrap-downloads/vmtarget/pharo --headless Pharo11.0-PR-64bit-3eb2d38.image --no-default-preferences eval --save 'Smalltalk vm parameterAt: 43 put: 32'

Segmentation fault Fri Nov  4 17:36:15 2022

MarcusDenker:
Failing test not related

PR #11860, opened 2022-11-07
#instVarNamed: does not need to send #asSymbol to the name

MarcusDenker:
ClassDescription>>#slotNamed:ifFound:ifNone:  converts the String to a symbol.

This was needed as AbstractLayoutScope>>#resolveSlot:ifFound:ifNone: was checking for identity.  

If we use #= there, we do not need to convert the string to a symbol.


MarcusDenker:
Failign test not related (timeout of a test on win)

Issue #11868, opened 2022-11-09
Cleaning up some refactorings that are just transformations.


Ducasse:
See the commit...

PR #11869, opened 2022-11-09
Fix: #11868 generate hash and printOn: are NOT refactoring so indicat…

Ducasse:
…e this to the user and the developer!!!!


Ducasse:
Broken tests are not related.

PR #11879, opened 2022-11-10
Clean blocks simpler checks

MarcusDenker:
simplify the checks on CleanBlockClosure about accessing self, super and ivars: by definition, a clean block does not access those


MarcusDenker:
failing test not related (timeout networking)

PR #11882, opened 2022-11-10
AST: add a recursive #hasNonLocalReturn and #isConstant

MarcusDenker:
This PR 

- adds a #hasNonLocalReturn method.It return true of there is a non-local return in the AST subtree This is mostly interesting to be called on a Block level: "does this block need a context to return or not?"
- add a #isConstant check for RBBlockNode. Blocks that just return a constant value can be optimized not only to be pre-compiled (as they are a special kind of clean), but we can even speed up execution (that is, implement a kind of block with a fast #value that returns the constant without the need to evaluate another method)


MarcusDenker:
error not related

PR #11885, opened 2022-11-13
11803 return value of inspect should be the receiver and not a sp window presenter 

SabineMa:

PR #11886, opened 2022-11-13
11803 return value of inspect should be the receiver and not a sp window presenter

hogoww:
redo of #11885 at the request of @SabineMa :)
(@SabineMa, please check that I did that properly please !)


SabineMa:
@hogoww I checked. It is exactly what I wanted. Thank you for your help.

Ducasse:
tx!

hogoww:
Unix only failure.
Looks unrelated to me.

PR #11887, opened 2022-11-13
New refactoring changes preview presenter

balsa-sarenac:
A new implementation of `RBRefactoringPreviewPresenter`. This is a cleaned and smaller version of `SycRefactoringPreviewPresenter`. You can test it with DeprecateMethod refactoring (only refactoring that uses it for the moment). 

While developing I've noticed that similar logic is implemented in multiple places:
* ChangesBrowser
* RBReplaceBrowser
* RewriteRuleChangesBrowser
* SycRefactoringPreviewPresenter
* RBRefactoringPreviewPresenter (this PR)
Maybe there is a way to reduce duplication here and reuse some of the logic in multiple places?

### DIFF
--- a/src/Refactoring-Core/RBDeprecateMethodDriver.class.st
+++ b/src/Refactoring-Core/RBDeprecateMethodDriver.class.st
@@ -16,10 +16,18 @@ Class {
 	#instVars : [
 		'methodToDeprecate',
 		'methodToReplaceDeprecatedMethod',
-		'refactoring'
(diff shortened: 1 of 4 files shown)


Ducasse:
Cool 
We will have to pay attention to dependencies and probably package the driver and presenters in a separate Refactorings-UI package in the future.

Ducasse:
Indeed we should reduce the duplication.

PR #11890, opened 2022-11-14
11881-RBBlockNodeisClean-takes-every-optimized-block-to-be-not-clean 

### LINKED ISSUE #11881: RBBlockNode>>#isClean takes every optimized block to be not clean
RBBlockNode>>#isClean has a check:

```
	self isInlined ifTrue: [ ^ false ].
```

But consider a clean block with an ifTrue: inside:

```
[ true ifTrue: [ 1 ] ] 
```

There is no reason why this block can not be compiled as a clean block. Removing the check allows us to compile some more bocks as clean blocks if the option is enabled:

```
CleanBlockClosure allInstances size 

"before: 12440" 
"after: 12729"
```

MarcusDenker:
- fix #isClean on the AST to treat optimized blocks correctly
- use (for now) the AST for BlockClosure>>#isClean
- move OCCleanBlockTest to be #isClean on the AST tests

fixes #11881


src/AST-Core-Tests/RBBlockNodeTest.class.st
```diff
@@ -25,6 +25,34 @@ RBBlockNodeTest >> testHasNonLocalReturn [
...
+	self assert: [ 1 + 2 ] sourceNode isClean.
+	self assert: [ :a | a + 2 ] sourceNode isClean.
+	self assert: [ :a :b | a + b + 3 ] sourceNode isClean.
+	self assert: [ | a | a := 1. a + 3 ] sourceNode isClean.
+	self assert: [ true ifTrue: [  ] ]  sourceNode isClean.
+	"optimized blocks"
+	self assert: [ 
+		| local2 | true ifTrue: [local2 := 1 ] ] isClean. "optimized blocks are clean"
+	self deny: [ true ifTrue: [escpWrite := 1 ] ] isClean. "But not if we have a real escape"
+]
```
demarey, line 54: shouldn't these tests be shared with BlockClosureTests?

MarcusDenker: There is 

[ ] isClean
and
[ ] sourceNode isClean

For now this is the same, but later #isClean on BlockClosure hierachy will not delegate to the AST.

Maybe having the same code with just  "sourceNode" removed and a comment in both tests to keep them in sync would be an idea

MarcusDenker: (the #sourceNode send can. not be in the Kernel tests as they are carefully structured to not need the AST/Compiler)

PR #11891, opened 2022-11-14
Fixes #11802 - Adds full consistency checks to class Time

### LINKED ISSUE #11802: Inconsistent Error message or inconsistent validation in Time
**Bug description**

`Time hour: 24 minute: 0 second: 0 `

raises debugger with the error message "An hour must be 24 or less". But it is exactly 24. 
So, the error message should be "An hour must be less than 24"
Same for Minutes.

Why then, is seconds allowed to have 60 seconds?

Bildschirmfoto 2022-10-18 um 12 51 39

Error message does not fit to the logic.

I dont like that limitation. In Pharo 8, there was no restriction like that and it was possible to make things like:

`(Time hour: 15 minute: 120 second: 10) printString -> "'5:00:10 pm'"`

which is correct 

**Version information:**
 - OS: iOS Mac M1
 - Pharo Version [10]

kasperosterbye:
Adds consistency checks for class Time.
I think this is the right solution, but awaits the integration test

### DIFF
--- a/src/Kernel-Chronology-Extras/Time.extension.st
+++ b/src/Kernel-Chronology-Extras/Time.extension.st
@@ -151,9 +151,9 @@ Time class >> hour: hour minute: minute second: second  nano: nano [
 Time class >> hour: hour minute: minute second: second  nanoSecond: nanoCount [
 	"Answer a Time - only second precision for now"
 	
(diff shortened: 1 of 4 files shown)


svenvc:
LGTM

PR #11892, opened 2022-11-14
Turn off ReLongMethodsRule for test methods

MarcusDenker:
ReLongMethodsRule checks for long methods as in normal code, it is better to break them up.

But for tests, this is not a problem to have long test methods.

This pr makes sure to not warn for long methods in tests


Ducasse:
Nice

PR #11894, opened 2022-11-14
ConstantBlockClosure: active when compiling with clean blocks enabled

MarcusDenker:
Experiment to see what will fail

- add a special kind of clean block: constant clean blocks. 
- change the compiler to instantiate those for all #isConstant bocks

### DIFF
--- a/src/Debugging-Core/CompiledBlock.extension.st
+++ b/src/Debugging-Core/CompiledBlock.extension.st
@@ -18,7 +18,7 @@ CompiledBlock >> pcInOuter [
 			literalOffset ifNotNil: [
 				| literal |
 				literal := (self outerCode literalAt: literalOffset + 1).
-				((literal isKindOf: CleanBlockClosure) and: [literal compiledBlock = self  ])ifTrue: [ ^ instructionStream pc ].  ].
+				((literal isKindOf: CleanBlockClosure) and: [literal compiledBlock == self  ])ifTrue: [ ^ instructionStream pc ].  ].
 			instructionStream pc: (instructionStream nextPc: (outer at: instructionStream pc))].	
 		
 	self error: 'block not installed in outer code'.
--- a/src/Debugging-Core/CompiledBlock.extension.st
+++ b/src/Debugging-Core/CompiledBlock.extension.st
@@ -18,7 +18,7 @@ CompiledBlock >> pcInOuter [
 			literalOffset ifNotNil: [
 				| literal |
 				literal := (self outerCode literalAt: literalOffset + 1).
-				(literal class == CleanBlockClosure and: [literal compiledBlock = self  ])ifTrue: [ ^ instructionStream pc ].  ].
+				((literal isKindOf: CleanBlockClosure) and: [literal compiledBlock = self  ])ifTrue: [ ^ instructionStream pc ].  ].
 			instructionStream pc: (instructionStream nextPc: (outer at: instructionStream pc))].	
 		
 	self error: 'block not installed in outer code'.
--- a/src/Fuel-Tests-Core/FLBlockClosureSerializationTest.class.st
+++ b/src/Fuel-Tests-Core/FLBlockClosureSerializationTest.class.st
@@ -106,6 +106,8 @@ FLBlockClosureSerializationTest >> testBlockClosureRemovedClean [
 	"Raise no error when materializing a clean closure whose method was removed."
 
 	| aClass aClosure |
+	"to be removed when we compile constant closure by default"
+	[  ] class == ConstantBlockClosure ifFalse: [ self skip ].
 	aClass := self classFactory silentlyNewClass.
 	self classFactory
 		silentlyCompile:  'methodWithClosure  ^ [ 42 ]'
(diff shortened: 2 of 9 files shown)


MarcusDenker:
(lots of tests will fail for sure)

MarcusDenker:
So we can 
- recompile the image with this
- kernel tests and compiler tests are ok
- bootrap works
- tests can be run but many fail

Problems:
- #pcInOuter does always finds the first block if there are multiple blocks that are equal
- the fake block has problems (maybe due to that) wrt to the pc->AST mapping

MarcusDenker:
Failing tests:

```
Announcer:

testAccessingSubscribers – MacOSX64.Announcements.Core.Tests.Tests.AnnouncerTest
testHandleSubscriberClass – MacOSX64.Announcements.Core.Tests.Tests.AnnouncerTest
testSubscribersForClass – MacOSX64.Announcements.Core.Tests.Tests.AnnouncerTest
testAccessingSubscribers – MacOSX64.Announcements.Core.Tests.Tests.WeakAnnouncerTest
testHandleSubscriberClass – MacOSX64.Announcements.Core.Tests.Tests.WeakAnnouncerTest
testSubscribersForClass – MacOSX64.Announcements.Core.Tests.Tests.WeakAnnouncerTest



Debugger:

testStepIntoDeadContextShouldRaiseException – MacOSX64.Debugger.Model.Tests.Core.StepIntoTest
testStepIntoUntilTermination – MacOSX64.Debugger.Model.Tests.Core.StepIntoTest

Process:
testIsNotTerminatedWhenItIsInsideLastTerminationMethod – MacOSX64.Kernel.Tests.Extended.Processes.ProcessTest
```

All of these fail for CleanBlocks, too.

The only difference to cleanblocks is this: we do not check numArgs for value to raise exception:

```
testBind – MacOSX64.Collections.Sequenceable.Tests.Ordered.OrderedCollectionTest
testCull – MacOSX64.Kernel.Tests.Methods.BlockClosureTest
testCullCull – MacOSX64.Kernel.Tests.Methods.BlockClosureTest
testCullCullCull – MacOSX64.Kernel.Tests.Methods.BlockClosureTest
testCullCullCullCull – MacOSX64.Kernel.Tests.Methods.BlockClosureTest
```

Not easy to solve outside of implementing special primitives, as checking would be too slow (but to be checked).

MarcusDenker:
The latest version turns constant block on when you enable clean blocks (they are special kinds of clean blocks after all)

Here is a bench:


Morph has a lot of method reading from properties using this pattern:

```
minHeight
	"answer the receiver's minHeight"
	^ self
		valueOfProperty: #minHeight
		ifAbsent: [2]
```

the [2] means we create a closure for every execution and then we evaluate it.

Thus with ConstantBlock implemented in the compiler, we get both the positive
effect of pre-compilation *and* the faster #value, the bytecode looks like this:

```
49 <4C> self
50 <20> pushConstant: #minHeight
51 <21> pushConstant: [2]
52 <A2> send: valueOfProperty:ifAbsent:
53 <5C> returnTop'
```

```	
morph := Morph new.
[morph  minHeight] bench		
```
```	
"constant block: '106336381.647 per second'" 
" Full Block:	  '36323454.327 per second'"

106336381.647 / 36323454.327 

	=> “2.9274853842289414"
```
Thus nearly 3 times faster.

MarcusDenker:
The fuel tests have to take into account during the transition: is clean block enabled or not?

dionisiydk:
> ```
> "constant block: '106336381.647 per second'" 
> " Full Block:	  '36323454.327 per second'"
> 
> 106336381.647 / 36323454.327 
> 
> 	=> “2.9274853842289414"
> ```
> 
> 
>     
>   
> 
> Thus nearly 3 times faster.

This is cool.
I also compared a regular CleanBlock and a Constant one. In my test it gives 30% improvements.

dionisiydk:
But with extra logic to fix bad number of args the performance is equivalent to raw clean blocks. 
I use this code:
```Smalltalk
value: anObject
	numArgs ~= 1 ifTrue: 	[self numArgsError: 1].
	^literal
```
And bench: 
```Smalltalk
<compilerOptions: #(+ optionCleanBlockClosure)>
b := [ :a | 1 ].
[ b value: 10 ] bench 
```
Please try same bench from your side. I can be wrong. But it seems clean blocks are good enough to not introduce extra complexity. Anyway it is a cool experiment :)

MarcusDenker:
If the eval speed is the same as Clean Blocks, we do not really gain anything... then we should not do it at all.
(as these blocks are all clean)

Ducasse:
I do not get why we need this check.
numArgs ~= 1 ifTrue: 	[self numArgsError: 1].

Why this is not a compile time error?

S

MarcusDenker:
We can not, as the block is compiled independely of the method that later is send to it. (it is just an object that stands in for a block, later someone calls #value or #value:.)

In BlockClosure, it is done as part of the fail code:

```
value
	"Activate the receiver, creating a closure activation (MethodContext)
	 whose closure is the receiver and whose caller is the sender of this
	 message. Supply the copied values to the activation as its copied
	 temps. Primitive. Essential."
	<primitive: 207>
	numArgs ~= 0 ifTrue:
		[self numArgsError: 0].
	^self primitiveFailed
```


I think the best is to do 2-3 subclasses for the case of 1, 2 (and maybe 3) args and not care about the others, see https://github.com/pharo-project/pharo/issues/11901

MarcusDenker:
(see https://github.com/pharo-project/pharo/issues/11901 for more)

MarcusDenker:
failing tests not related

- windows-64 / Tests-windows-64 / testDoIt – Windows64.Fuel.Tests.Core.Base.FLContextSerializationTest
- windows-64 / Tests-windows-64 / testEntryForAbsentEntry – Windows64.Ombu.Tests.OmSessionStoreTest

PR #11898, opened 2022-11-15
Implement #asSlot to return a correct Slot instances for complex slots

MarcusDenker:
This PR implements #asSlot to work with Slots that are not InstanceVariableSlots. It falls back to create a UndefinedSlot if a problem happens


MarcusDenker:
error not related

PR #11900, opened 2022-11-15
11883-CleanBlockChecker-does-not-recurse-into-the-blocks

### LINKED ISSUE #11883: CleanBlockChecker does not recurse into the blocks
**Bug description**
CleanBlockChecker is used by BlockClosure>>#isClean to have a way to ask a block if it could be clean.

This will be removed as soon as we compile clean blocks by default (then CleanBlockClosure returns true and the others false, no scan needed).

Until then, we should make sure that both are on sync (or change to go via the AST for all cases).

With CleanBlocks turned OFF, we get now:

```
[ 1 do: [ ^2 ] ]  isClean  "true"
[ 1 do: [ ^2 ] ]  sourceNode isClean "false"
```

MarcusDenker:
CleanBlockChecker was added so we do not need the AST for a BlockClosure to decide it a Block could be clean.

But with clean blocks soon in a state that we can use them, we do not need it anymore: the closure itself that compiler creates knows it it is clean or not.

fixes #11883 

(until clean blocks are enabled by default, it delegates the check to the AST)


MarcusDenker:
(in case we do not enable clean blocks for Pharo11, I will add this back)

MarcusDenker:
Test timeout (win only):
- FLContextSerializationTest>>testDoIt

PR #11905, opened 2022-11-16
IRBuilder: add support for outerContext flag for Closure creation 

MarcusDenker:
The block creation bytecode supports to create blocks without an outer context.

This PR adds support to IRBuilder to create blocks like that. It is not active.

To enable it in the compiler, add the argument to the end of the block creation in OCASTTranslator>>#visitBlockNode:  "outerContextNeeded: aBlockNode hasNonLocalReturn"

Note: for sure this will need changes to work, I can recompile the image but Job>>#owner makes problems


MarcusDenker:
I merge this as this is not active (just adds the API to play with it for the future)

PR #11907, opened 2022-11-16
CompiledBlocks: compile quick returns

MarcusDenker:
This PR makes sure to compile Quick returns if a block has just a non-local return, just as we do for methods.


MarcusDenker:
A naive benchmark, we have to bench creation + execution together:	

```
tt
	^[ ^ 1 ] value 
```

no Quick return: "'26069375.575 per second'"
with Quick return: "'30247736.053'"

30247736.053/26069375.575 "1.16"

PR #11916, opened 2022-11-17
Fix print on recursive array

### LINKED ISSUE #11479: Recursive Array cannot be printed
**Bug description**
An Array with itself as element cannot be printed, keep in loop.

**To Reproduce**
Playground example:
```st
l := { 1 . 2 . 3 } asArray.
l at: 1 put: l.
```

**Expected behaviour**
To have the same behaviour as OrderedCollection (recursive print until a threshold)

**Version information:**
 - OS: [MacOS]
 - Pharo Version [Pharo 10.0.0]

PalumboN:
Fix https://github.com/pharo-project/pharo/issues/11479

with @jordanmontt @sbragagnolo and @privat 
in the Pharo Sprint on 28/10

----------

Maybe we can add another issue to show a "good" string. 
Now the printed string is not very useful, it doesn't show the elements after recursion:
```
l1 := {1 . 2 . 3 . 4 . 5}.
l1 at: 3 put: l1.
l1 "an Array(1 2 an Array(1 2 an Array(1 2 an Array(1 2 an Array( ..."
```

It's the same for any collection, not only arrays.

### DIFF
--- a/src/Kernel/Object.class.st
+++ b/src/Kernel/Object.class.st
@@ -1353,6 +1353,14 @@ Object >> isSelfEvaluating [
 	^ self isLiteral
 ]
 
+{ #category : #'self evaluating' }
+Object >> isSelfEvaluating: alreadyVisited [
+
+	(alreadyVisited identityIncludes: self) ifTrue: [ "Recursive structure cannot be a literal" 
+		^ false ].
+	^ self class == Array or: [ self isLiteral ]
+]
+
 { #category : #testing }
 Object >> isStream [
 	"Return true if the receiver responds to the stream protocol"
@@ -1903,6 +1911,14 @@ Object >> shouldBeImplemented [
 Object >> shouldBePrintedAsLiteral [
 
 	^self isLiteral
+]
+
+{ #category : #'literal testing' }
+Object >> shouldBePrintedAsLiteral: alreadyVisited [
+
+	(alreadyVisited identityIncludes: self) ifTrue: [ "Recursive structure cannot be a literal" 
+		^ false ].
+	^ self class == Array or: [ self isLiteral ]
 ]
 
 { #category : #'error handling' }
--- a/src/Kernel/Object.class.st
+++ b/src/Kernel/Object.class.st
@@ -1262,6 +1262,12 @@ Object >> isLiteral [
 	^ false
 ]
 
+{ #category : #testing }
+Object >> isLiteral: alreadyVisited [
+
+	^ self isLiteral
+]
+
 { #category : #'class membership' }
 Object >> isMemberOf: aClass [
 	"Answer whether the receiver is an instance of the class, aClass."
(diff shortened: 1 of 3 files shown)


jordanmontt:
Nice!

src/Kernel/Object.class.st
```diff
@@ -1353,6 +1353,14 @@ Object >> isSelfEvaluating [
 	^ self isLiteral
 ]
 
+{ #category : #'self evaluating' }
+Object >> isSelfEvaluating: alreadyVisited [
```
jordanmontt, line 1368: If I understand correctly, this receiver of `isSelfEvaluating:` will be always a subclass of `Collection`. So, it would not make more sense to put this method in Collection instead of Object?

Ducasse: Yes it looks like :)

MarcusDenker:
failing tests not related

src/Kernel/Object.class.st
```diff
@@ -1353,6 +1353,14 @@ Object >> isSelfEvaluating [
...
+
+	(alreadyVisited identityIncludes: self) ifTrue: [ "Recursive structure cannot be a literal" 
+		^ false ].
+	^ self class == Array or: [ self isSelfEvaluating ]
```
privat: The type check on Array here seems to make the logic quite complex.
I understand that it is to avoid an infinite reduction on `self isSelfEvaluating` (and that is the core issue of the PR) but it is weird to ask a method in a superclass to have responsibility for the shenanigans of some of its subclasses.

jordanmontt:
Also, when this PR will be merged, this will close #11492 too as they are refering to the same issue.
Closes #11492

privat:
Loosely related, but the specificities of the Object methods  `isLiteral` and `shouldBePrintedAsLiteral` are not obvious. Maybe you could add a comment? Also `isSelfEvaluating` need some comment.

privat:
By the way, how does `isLiteral` behave on a recursive array? Maybe also fix it if needed.

src/Kernel/Object.class.st
PalumboN: Thanks!

src/Kernel/Object.class.st
PalumboN: No, the check is ok because the "self evaluated" and "literal" printing only works for literals:
```st
#(1 2 3 4 5)
{(1@2). 1}
```

If you evaluate those expressions, Pharo will build Array objects, no subclasses of it.

PalumboN: Anyway, we remove the check because is in the entry point method `Array >> isSelfEvaluating`

PalumboN:
Thanks for the comments! We already fix (or answer) them with @inao0 😄

MarcusDenker (approved):

PR #11918, opened 2022-11-17
Complete symbols

### LINKED ISSUE #11748: Code completion should complete symbols
**Describe the request**

Symbols are used a lot in Pharo, internally (as names and selectors) but also explicitly as literal #foo for a variety of usages.
Currenlty the completion popup is not triggered when writing a token that starts with # and some letters.

Having completion on symbols would help to avoid spelling mistakes.

**Expected behavior**

A popup completion window should appear on `#` plus two letters and propose existing symbols

privat:
This enables code completion on literal symbols.
Currently, it uses the whole system information; a future improvement could be to prioritize symbols existing in the package, for instance.

Note: seriously, this was more complex than I expected, as I needed to create 3 new classes and to add a bunch of methods and instance variables to some other classes (without counting the small refactorization). So maybe I did something wrong.

Note2: I'm not sure to understand the relation between the packages `NECompletion` and `HeuristicCompletion` ; especially since there is existing code that seems to do what I wanted `RBLiteralValueNode>>#completionEntries:`. Is this dead code or something?

Ping @guillep for review since he worked on the completion engine.

Fix #11748


guillep:
> Note: seriously, this was more complex than I expected, as I needed to create 3 new classes and to add a bunch of methods and instance variables to some other classes (without counting the small refactorization). So maybe I did something wrong.

As I see it, you added 3 classes, one is an extracted class. The other two:
 - added capabilities to collect literals
 - added a fetcher to fetch symbols (and not just selectors)

so it's not so crazy to me

> Note2: I'm not sure to understand the relation between the packages NECompletion and HeuristicCompletion ; especially since there is existing code that seems to do what I wanted RBLiteralValueNode>>#completionEntries:. Is this dead code or something?

Yes, dead code.

guillep (approved):
Change is good, it's a pity it does not have tests.

It could be made later

PR #11919, opened 2022-11-18
11901-ConstantBlockClosure-raise-error-if-called-with-wrong-number-of-args

MarcusDenker:
fix error handling

### DIFF
--- a/src/Kernel/ConstantBlockClosure.class.st
+++ b/src/Kernel/ConstantBlockClosure.class.st
@@ -11,6 +11,8 @@ We create a dummy compiledBlock which and use it to find the pcInOuter / do pc-a
 
 the RBBlockNode has two methods: #isConstant returns true, #constantValue returns the literal for constant blocks.
 
+This class implements 0-arg blocks, with correct error handling for args >0. Subclasses for 1-3 args are implemented.
+
 
 	
 "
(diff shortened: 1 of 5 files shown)


src/Kernel/ConstantBlockClosure.class.st
```diff
@@ -21,6 +23,21 @@ Class {
 	#category : #'Kernel-Methods'
 }
 
+{ #category : #'instance creation' }
+ConstantBlockClosure class >> numArgs: numArgs literal: aLiteral [
+	| classToUse |
+	classToUse := numArgs caseOf:
```
dionisiydk, line 29: It seems #caseOf message is going to be deprecated. Its protocol is "casing-To be deprecated".

src/Kernel/ConstantBlockClosure.class.st
```diff
@@ -21,6 +23,21 @@ Class {
...
+	{
+		[0]->[ self ].
```
dionisiydk, line 31: I think it would be more clear to have explicit ConstantBlockClosure0Arg class and keep ConstantBlockClosure as abstract

dionisiydk:
An it would be nice to have more tests for each kind of block

src/Kernel/ConstantBlockClosure.class.st
MarcusDenker: there are *many* users of caseOf. I tried to implement it without, but this was the nicest solution.

src/Kernel/ConstantBlockClosure.class.st
MarcusDenker: I add it to the TODO, I was thinking the same. Maybe adding the 4 arg version is good, too (to have all the ones coverd that we can call without using #valueWithArguments:)

PR #8791, opened 2021-03-15
Add opal plugin to improve inlining

guillep:
Redo #7037, the old PR became a bit mad after we updated it

PR #11923, opened 2022-11-19
Use-optimizable-ifTrue 

MarcusDenker:
fix some cases where ifTrue: was used with a variable instead of a block, which means that it can not compile an optimized ifTrue:

### DIFF
--- a/src/FileSystem-Core/AbstractFileReference.class.st
+++ b/src/FileSystem-Core/AbstractFileReference.class.st
@@ -157,14 +157,14 @@ AbstractFileReference >> binaryReadStreamDo: aBlock [
 AbstractFileReference >> binaryReadStreamDo: doBlock ifAbsent: absentBlock [
 	^ self isFile 
 		ifTrue: [ self binaryReadStreamDo: doBlock ]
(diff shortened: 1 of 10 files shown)


src/FileSystem-Core/AbstractFileReference.class.st
```diff
@@ -157,14 +157,14 @@ AbstractFileReference >> binaryReadStreamDo: aBlock [
 AbstractFileReference >> binaryReadStreamDo: doBlock ifAbsent: absentBlock [
 	^ self isFile 
 		ifTrue: [ self binaryReadStreamDo: doBlock ]
-		ifFalse: absentBlock
+		ifFalse: [ absentBlock value ]
```
dionisiydk, line 160: Should not it be part of compiler optimization logic? 
It's strange that we are forced to write such code.

MarcusDenker: Yes, in the long term this is the goal. There was a first experiment (see https://github.com/pharo-project/pharo/pull/8791).

As soon as we solve this better, we of course will remove the rules.

PR #11927, opened 2022-11-20
11912-Compiler-add-option-for-constant-blocks

### LINKED ISSUE #11912: Compiler: add option for constant blocks
In https://github.com/pharo-project/pharo/pull/11894 we added a specialized kind of clean block, the ConstantBlockClosure.

This is now active together with optionCleanBlockClosure, but as getting clean blocks to a state to turn it on by default will require lots of small fixes, we should devide and conquer and first try to just fix everything for constant blocks only.

-> add  compiler option #optionConstantBlockClosure

MarcusDenker:
add an option to turn on constant block that is independend of the clean block option. fixes #11912


dionisiydk:
we should have a test for new option

MarcusDenker:
ah, true, I will add one

PR #11929, opened 2022-11-21
Fixes #11925, ExternalBrowser does not show class definition from ST files Bug

### LINKED ISSUE #11925: ExternalBrowser does not show class definition from ST files
**Bug description**

**To Reproduce**
Warning: this cannot be reproduced until issue #11762 is fixed by PR #11924

Execute the following in a workspace:
```smalltalk
ExternalBrowser browseStream: (ReadStream on: 'Object subclass: #KaOTest
	instanceVariableNames: ''xxx''
	classVariableNames: ''''
	package: ''KaoTest''!

!KaOTest methodsFor: ''playing'' stamp: ''KasperOsterbye 11/18/2022 08:36''!
foo
	^ ''bla''! !
')
```

**Expected behavior**
Selecting the class should show its class definition in the code presenter
**Screenshots**
image

**Version information:**
This has been the case all way back to Pharo 6 (as loaded from the Launcher)

**Expected development cost**
The problem is in RingChunkImporter, where `visitDoItChunk:` does not recognise class definitions.
In any case, ExternalBrowser should most likely use RGChunkImporter now

kasperosterbye:
Fixes issue: #11925.
This patch makes sure `visitDoItChunk:` delegates to `classDefinition:with:` for a wider range of class definitions. It is `classDefinition:with` which does the actual work.

### DIFF
--- a/src/Ring-OldChunkImporter/RingChunkImporter.class.st
+++ b/src/Ring-OldChunkImporter/RingChunkImporter.class.st
@@ -215,7 +215,7 @@ RingChunkImporter >> visitDoItChunk: aChunk [
 
 	| contents |
 	contents := aChunk contents.
-	('*ubclass:*instanceVariableNames:*classVariableNames:*poolDictionaries:*category:*'
+	('*ubclass:*instanceVariableNames:*'
 		match: contents) ifTrue:[^self classDefinition: contents with: aChunk].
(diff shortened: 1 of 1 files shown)


MarcusDenker:
Failure *not* related: 

- FLContextSerializationTest>>testDoIt

MarcusDenker (approved):

Issue #11934, opened 2022-11-21
[CI] Flaky test:  FLContextSerializationTest>>testDoIt  times out


MarcusDenker:
**Bug description**
We see   FLContextSerializationTest>>testDoIt  timing out on the CI often, mostly on win, sometimes on Linux

MarcusDenker:
Main issue is https://github.com/pharo-project/pharo/issues/12006

In this issue we should do some workaround to have the Ci green

PR #11942, opened 2022-11-22
EpHasImpactVisitor>>visitProtocolRemoval: handle protocols that only contain trait methods.

### LINKED ISSUE #11941: EpHasImpactVisitor>>visitProtocolRemoval: incorrectly marks event as having an impact
**Bug description**
EpHasImpactVisitor>>visitProtocolRemoval: incorrectly answer true for protocols that only have trait inherited methods. In this case removing the protocol doesn't have any impact since no methods are removed and the protocol will be kept anyway due to the trait method(s).

**To Reproduce**
RBSmalllintTestLocalMethodsSameThanTraitObject's `private` protocol contains only trait methods, so the protocol removal will have no affect (this can be tested by removing it in Calypso, the protocol will still exist and all methods will be present).

The following script should pass:

```
self assert: ((RBSmalllintTestLocalMethodsSameThanTraitObject methodsInProtocol: #private) allSatisfy: #isFromTrait).
event := EpProtocolRemoval
	behavior: RBSmalllintTestLocalMethodsSameThanTraitObject asRingDefinition 
	protocol: #private.
visitor := EpHasImpactVisitor inEnvironment: self class environment.
self assert: (event accept: visitor) not.
```

however the final assertion will fail.

**Expected behavior**
The above script passes.

**Version information:**
 - OS: Ubuntu
 - Version: 22.04
 - Pharo Version: 10 & 11

**Expected development cost**
PR on the way.

akgrant43:
See: https://github.com/pharo-project/pharo/issues/11941

### DIFF
--- a/src/EpiceaBrowsers-Tests/EpHasImpactFilterTest.class.st
+++ b/src/EpiceaBrowsers-Tests/EpHasImpactFilterTest.class.st
@@ -118,3 +118,25 @@ EpHasImpactFilterTest >> testMethodWithoutImpact [
 	anEvent := OmEntry content: (EpMethodAddition method: aClass >> #a).
 	self deny: (impactFilter accepts: anEvent)
 ]
+
+{ #category : #tests }
+EpHasImpactFilterTest >> testProtocolRemovalOfTraitProtocol [
+	| anEntry aClass |
+
+	aClass := classFactory
+		newSubclassOf: classFactory defaultSuperclass 
+		uses: TAssertable
+		instanceVariableNames: '' 
+		classVariableNames: '' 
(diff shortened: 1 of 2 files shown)


Ducasse:
Tx I love the tests :)

MarcusDenker:
Failing tests are not related:

- [CI] Flaky test: FLContextSerializationTest>>testDoIt times out #11934
- and a FileDoesNotExistException for ombu on win that sometimes happens

Linux and Mac are green

MarcusDenker (approved):

akgrant43:
Thanks, Marcus & Stef!

PR #11950, opened 2022-11-22
Fix #11949 less help class

### LINKED ISSUE #11949: start removing help classes
Now with the new document browser.... we can start removing help classes

Ducasse:


Ducasse:
We should fix all the system dependencies tests once all Help Browser will be removed.

Ducasse:
Broken tests are not related.

Issue #11797, opened 2022-10-14
Dependency Tab is updated *every* time you select a new method


MarcusDenker:
**Bug description**
There is a Dependency tab now in the browser.

The problem is that it is updated (showing a progress bar) *every* time you change which method you look at.

This slows down things *a lot* and is very ugly

maxwills:
I've been experiencing this issue to. It's very annoying. In some cases, taking more than one second for the update to complete.

It is a little tricky to reproduce. This is how I manage to do it:

1. Browse a class with numerous dependencies:
   - Do this (DelayNullScheduler browse). 
   - It should open a System Browser with the class. Notice that the Dependency tab is not present, so changing methods is instantaneous.
2.   With the class DelayNullScheduler selected in in System Browser, now click directly in the 'Kernel' package. The class should still be selected, but now the Dependencies tab is visible.
3. Select any method from the class and experience the delay.

![Screen Shot 2022-11-02 at 11 13 55](https://user-images.githubusercontent.com/1684221/199504017-8067b561-fb5e-4efa-ba3b-c86b9525e285.png)

Ducasse:
Indeed and this is worse than that the all the note panes are all refreshed even when they are not visible.

demarey:
With @Ducasse, we investigated this issue.
In Calpyso, tools (what is inside tabs) are created lazily. This is good.
The bad news is that any change on the selection triggers a rebuild of all tools.
It is triggered by `ClyBrowserMorph>>#changeStateBy:` (see https://github.com/pharo-project/pharo/blob/Pharo11/src/Calypso-Browser/ClyBrowserMorph.class.st#L228) that will call #rebuildAllTools.
We then get to #basicUpdateAllTools (see https://github.com/pharo-project/pharo/blob/Pharo11/src/Calypso-Browser/ClyNotebookManager.class.st#L95-L106). 
You can reproduce by putting a halt in some tool (ex: halt in `DABrowserToolMorph>>#daPackageFor:`.

We should revisit this to discard tools no more relevant to the context, inform tools of the change and let them decide if they need an update (e.g. new model) or not, and add new tools relevant to the context.

[...]

PR #11952, opened 2022-11-23
do not rebuild all Calypso tools at each selection change

demarey:
tools are now compared with the context with a method #isValidInContext: defined on the ClyToolMorph. This method has to be implemented on each tool to avoid rebuild.


demarey:
fixes #11797

MarcusDenker:
I dowloaded the build artefacts: works!

MarcusDenker (approved):

PR #11982, opened 2022-11-25
11844-Renaming-a-package-loses-all-the-package-extensions

tesonep:
#Fix 11844.
Could not reproduce the issue, but we have written tests for validating it is working.


Ducasse:
Thanks!

Ducasse:
Pablo may be I broke my package using a refactoring?
I do not remember what I did. I will check.

PR #11999, opened 2022-11-27
Fix spechelp problem

estebanlm:


Ducasse:
Tx it fixes it.

Ducasse:
I should have looked for SpecHelp

PR #12000, opened 2022-11-27
11966-Implement-BlockClosurehomeMethod-without-using-home 

### LINKED ISSUE #11966: Implement BlockClosure>>#homeMethod (without using #home)
**Bug description**
the #home of a BlockClosure is often used just to find the home *method*. 

```
[[]home method] value 
-->  "UndefinedObject>>#DoIt"
```
But this is *static* information, the block knows it's outerCode, and we can follow it till we find the method statically where
the block is defined.

** what to do **
Implement #homeMethod without using #home.

MarcusDenker:
add #homeMethod and test

### DIFF
--- a/src/Kernel/BlockClosure.class.st
+++ b/src/Kernel/BlockClosure.class.st
@@ -297,6 +297,14 @@ BlockClosure >> home [
 	^ outerContext ifNotNil: [ outerContext home ]
 ]
 
+{ #category : #accessing }
+BlockClosure >> homeMethod [
+	"return the home method. Use the static outerCode if the closure has not outerContext"
+	^ self home
+		ifNotNil: [:homeContext | homeContext method ]
+		ifNil: [self compiledBlock outerCode method ]
+]
+
 { #category : #exceptions }
 BlockClosure >> ifCurtailed: aBlock [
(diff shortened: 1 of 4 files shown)


src/Kernel/BlockClosure.class.st
```diff
@@ -297,6 +297,14 @@ BlockClosure >> home [
 	^ outerContext ifNotNil: [ outerContext home ]
 ]
 
+{ #category : #accessing }
+BlockClosure >> homeMethod [
+	"return the home method. If no #home is available due to no outerContext, use the compiledBlock"
+	^ (self home
+		   ifNotNil: [ :homeContext | homeContext ]
+		   ifNil: [ self compiledBlock ]) method
```
dionisiydk, line 305: What is the reason to not always use "self compiledBlock method"?

MarcusDenker: it is  slower, but I wonder if we care 

```
block := [[]]value. 

[block homeMethod ] bench "'53052503.698 per second'" 
[block homeMethod2 ] bench "'28656341.263 per second'"  

53052503.698/28656341.263 "1.8513355634307516"  
```

PR #12004, opened 2022-11-28
11928-SystemDictionaryallMethods-rerturns-Traits-methods

### LINKED ISSUE #11928: SystemDictionary>>#allMethods rerturns Traits methods
**Bug description**
SystemDictionary>>#allMethods returns all the methods that are implementation details of Traits (TraitedClass>>#doRebuildMethodDictionary and all the methods there) and all methods that the traits adds to a class.

```
Smalltalk globals allMethods count: [ :each | each selector =  #doRebuildMethodDictionary 
] "324"
```
Then check "implementors of" #doRebuildMethodDictionary -> there is just one.

**Why?**

The reason is that these methods are returned when sending #methods.

To try:

```
Smalltalk globals allMethods size "134983"
```

Now change #allMethods to send #localMethods to the behavior:

```
Smalltalk globals allMethods size "121367"
```

What is a solution?

- we could use #localMethods there
- but in general this is not nice: why do I have to decide #methods vs. #localMethods? Is there at any point a reason *not* to use #localMothods?

MarcusDenker:
Instead of fixing #allMethods, this adds #methods, this way we can get all the methods with traits installed copies with #allMethods. 

- add #methods to SystemNavigation and Smalltalk globals: #allMethods returns with trait installes, methods without
- inline Behavior>>#localMethods
- fix all senders of allMethods to use #methods

fixes #11928

### DIFF
--- a/src/Collections-Strings/Symbol.class.st
+++ b/src/Collections-Strings/Symbol.class.st
@@ -199,7 +199,7 @@ Symbol class >> readFrom: strm [
 { #category : #'selector table' }
 Symbol class >> rebuildSelectorTable [
 
-	^ SystemNavigation new allMethods
+	^ SystemNavigation new methods
 		  collect: [ :method | method selector ]
 		  as: WeakIdentitySet
 ]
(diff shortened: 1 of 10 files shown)


MarcusDenker:
bootstrap fails:

```
Error: localMethods not found in Abort class
```

MarcusDenker:
So this is not possible: We can never call #localMethods by default, as traits are supposed to be optional and thus #localMethods might not exist.

PR #12007, opened 2022-11-29
Context-use-homeMethod 

MarcusDenker:
- replace "home method" with the new #homeMethod which works for blocks without a home
- fix a comment in Hermes


MarcusDenker:
failing test not related:

- ProperMethodCategorizationTest>>testNoUncategorizedMethods fails on CI #12009

PR #12018, opened 2022-11-29
This PR introduced a basic use of the new string primitive imlpemente…

### LINKED ISSUE #3156: Use faster string primitives

jecisc:
…d by Sophie.

Fix [#https://github.com/pharo-project/pharo/issues/3156](https://github.com/pharo-project/pharo/issues/3156)

There are two more steps to improve further the perfs:
- Remove passing the ASCII order when possible because the new primitive works faster if we do no pass this optional parameter
- Update the users of #String>>#compare:with:collated: to compare with positive/0/negative instead of 0/1/2 and avoid to send "sign +2" 

Benches: 

[ 'This is a test' = 'This is a test' ] bench.

"After: '18266371.200 per second'"
"Before: '10208252.099 per second'"

[ 'This is a test' = 'This isn''t a test' ] bench. "After: '42902022.000 per second'"
"Before: '42747076.139 per second'"

[ 'This' = 'This' ] bench.
"After: '18408173.296 per second'"
"Before: '10456056.377 per second'"

[ 'This' = 'That' ] bench.
"After: '20356481.311 per second'"
"Before: '10439757.545 per second'"

### DIFF
--- a/src/Collections-Strings/ByteString.class.st
+++ b/src/Collections-Strings/ByteString.class.st
@@ -184,6 +184,37 @@ ByteString >> byteSize [
 	^self size
 ]
 
+{ #category : #comparing }
+ByteString >> compareWith: anotherString [
+
+	"Return a negative Smi, 0 or a positive Smi if self is anotherString, with the collating order of characters given optionnally by the order array."
+
+	<primitive: 158>
+	^ self compareWith: anotherString collated: AsciiOrder
+]
+
+{ #category : #comparing }
+ByteString >> compareWith: anotherString collated: order [
+
+	"Return a negative Smi, 0 or a positive Smi if self is anotherString, with the collating order of characters given optionnally by the order array."
+
+	<primitive: 158>
+	| len1 len2 c1 c2 |
+	len1 := self size.
+	len2 := anotherString size.
+	1 to: (len1 min: len2) do: [ :i | 
+		c1 := order at: (self basicAt: i) + 1.
+		c2 := order at: (anotherString basicAt: i) + 1.
+		c1 = c2 ifFalse: [ 
+			^ c1 < c2
+				  ifTrue: [ -1 ]
+				  ifFalse: [ 1 ] ] ].
+	len1 = len2 ifTrue: [ ^ 0 ].
+	^ len1 < len2
+		  ifTrue: [ -1 ]
+		  ifFalse: [ 1 ]
+]
+
 { #category : #converting }
 ByteString >> createSymbol [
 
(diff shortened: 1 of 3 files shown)


Ducasse:
good initiative!!!

I think that some broken tests are related 

```
 testRecategorizeClassRegisterTheClassMethodsInTheNewPackage – MacOSX64.Monticello.Tests.RPackage.RPackageClassesSynchronisationTest

 testRecategorizeClassWithMetaClassMethodsRegisterAllClassMethodsInTheNewPackage – MacOSX64.Monticello.Tests.RPackage.RPackageClassesSynchronisationTest
 testReorganizeClassByRenamingExtensionCategoryToAnotherExtensionCategoryAddMethodsToTheNewPackage – MacOSX64.Monticello.Tests.RPackage.RPackageClassesSynchronisationTest

 testRepackagedAnnouncementWhenModifyMethodByMovingFromDifferentExtensionCategories – MacOSX64.Monticello.Tests.RPackage.RPackageMCSynchronisationTest
 testModifyMethodByMovingFromExtensionCategoryToExtensionCategoryMoveItFromExtendingPackageToSecondExtendingPackage – MacOSX64.Monticello.Tests.RPackage.RPackageMethodsSynchronisationTest

```

jecisc:
I’ll take a look when I’ll have the time thanks

jecisc:
It does not seems that easy to find the place in the code that breaks the tests. I'll look more into it another time

jecisc:
I reduced the problem to this code:

```st
'ZZZZZ' sameAs: 'zzzzz'
```

The new primitive does not give the same result than the old one

hogoww:
Yes, that's why i ended up not fixing it 2 years ago, the primitives do not have the same behaviors.
Cool that you have found such a smal' reproduceable case!

guillep:
@jecisc maybe we can introduce the new code without the usage, like that we can debug it later? That would mean reverting the change to `String >> compare: aString caseSensitive: aBool`.

What do you think?

jecisc:
> @jecisc maybe we can introduce the new code without the usage, like that we can debug it later? That would mean reverting the change to `String >> compare: aString caseSensitive: aBool`.
> 
> What do you think?

Done!

PR #12022, opened 2022-11-30
12006-Performance-Regression-in-Fuel-Serialization-of-Blocks--CompiledMethods 

### LINKED ISSUE #11934: [CI] Flaky test:  FLContextSerializationTest>>testDoIt  times out

### LINKED ISSUE #12006: Performance Regression in Fuel Serialization of Blocks & CompiledMethods
The CI is breaking with the test FLContextSerializationTest.testDoIt
This test is failing because Fuel has a performance regression.

In Pharo 11:

```
[FLContextSerializationTest run: #testDoIt] timeToRun. "0:00:00:05.642"
```

In Pharo 10:
```
[FLContextSerializationTest run: #testDoIt] timeToRun. "0:00:00:00.016"
```

I have profiled and I see that the issue is in the implementation of: 

FLFullGeneralMapper>>#shouldSerializeCompiledMethodFully:

MarcusDenker:
Fixes #12006 by compiling the closure without outer context

### DIFF
--- a/src/Fuel-Core/BlockClosure.extension.st
+++ b/src/Fuel-Core/BlockClosure.extension.st
@@ -11,9 +11,9 @@ BlockClosure >> cleanCopy [
 
 { #category : #'*Fuel-Core' }
 BlockClosure >> cleanOuterContext [
-	"Clean my outerContext preserving just the receiver and method, which are the only needed by a clean block closure."
-
-	outerContext := outerContext cleanCopy
+	"Clean my outerContext preserving just the receiver and method, which are the only needed by a clean block closure. outerContext can already be nil if the block was compiled without context"
+	
+	outerContext ifNotNil: [ outerContext := outerContext cleanCopy ]
 ]
 
 { #category : #'*Fuel-Core' }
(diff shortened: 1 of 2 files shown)


tesonep:
Hi Marcus, this fix our problem with the test, but it is not hiding the performance regression when serializing blocks with outerContext? That is not the case of the doIt, but for other blocks?

MarcusDenker:
The problem is that there is no solution: *if* we have a full block with outer context, we have to serialize it, which can lead to problems that are hard to control.

What Fuel tries to do is to analyze and serialize clean blocks a bit better (but it confuses "clean" and "needs no outerContext").

So one idea can be to improve that, e.g. to nil out the outerContext for all blocks with no non-local return when storing.

Even better: the plan is to enable Constant, Clean and Blocks without outer by default in Pharo. Them we can implement serialization to follow that: serialize the static blocks for constant/clean, there is no outer if there is no return...

But if you try to serialize a block *with* outerContext, you might nevertheless serialize too much... that is just a property of the model if closures... they are very hard to serialize without running into trouble.

But there are not many closures with a non-local return, so it should work for many many cases quite nicely?

Not sure if my understanding is completely correct, though.

theseion (approved):

theseion:
In version 5 of Fuel I added facilities to control the level of the serialised graph and the absolute number of objects. You could use that as a workaround for serialising exceptions:

```smalltalk
FLSerializer new
    limitDepthTo: 50;
    ...
```

That limit will of course apply to all objects, so every object with a distance above 50 from the root will be replaced with `nil`, which may limit the amount of useful information you get.

PR #12023, opened 2022-11-30
use Context>>#homeMethod

MarcusDenker:
- use #homeMethod in Context #methodNode and #methodSelector
- simplify printOn: #selector already does the nil check and fallback to default
- use homeMethod in explicitRequirement and BytecodeEncoder

### DIFF
--- a/src/Kernel/Object.class.st
+++ b/src/Kernel/Object.class.st
@@ -894,9 +894,9 @@ Object >> explicitRequirement [
 	| originalMethod originalArguments errorBlock originalReceiver callingContext originalSelector |
 	errorBlock := [ ^ self error: 'Explicitly required method' ].
 	callingContext := thisContext sender.
-	originalMethod := callingContext compiledCode method.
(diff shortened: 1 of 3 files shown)


src/Kernel/Object.class.st
```diff
@@ -894,9 +894,9 @@ Object >> explicitRequirement [
 	| originalMethod originalArguments errorBlock originalReceiver callingContext originalSelector |
 	errorBlock := [ ^ self error: 'Explicitly required method' ].
 	callingContext := thisContext sender.
-	originalMethod := callingContext compiledCode method.
+	originalMethod := callingContext homeMethod.
 	originalMethod isFromTrait
-		ifFalse: errorBlock.
+		ifFalse: [ errorBlock value].
```
privat, line 899: For curiosity, why this change?

MarcusDenker: just to use the optized ifFalse... I guess not important here as speed does not matter

Issue #11220, opened 2022-05-11
[CleanBlocks] Announcements defined via # when:do:


MarcusDenker:
**Bug description**
when:do:  with a CleanBock as a parameter does not work.

**To Reproduce**
1. Set the cleanblock option in the compiler Setting
2. Do "OpalCompiler recompileAll"
3. run #testHandleSubscriberClass

**More...**

It calls #receiver on the block, in AnnouncementSubscription>>#valuable:

```
valuable: aValuable
	"Used when subscriber should be extracted from valuable object"
	self action:  aValuable.
	self subscriber: aValuable receiver.
	
```

receiver is now exactly what a clean block does *not* know. 

The problem now is that yes, the static case is trivial to solve: We can just disable clean
blocks for now (for a transition period) in all methods that send when:do: with a static block. 

But that does not help if the when:do: parameter is not the block statically, but a variable or result of some expression. The cases should be happening not that often, though,

We know the method statically where the block comes from (via the outerCode pointer), so we can make a good
Transcript output for that case to warn users.

MarcusDenker:
The static case can be solved, when the when:do: has the block directly, we can just walk up the stack to find the sender:

```
valuable: aValuable
	"Used when subscriber should be extracted from valuable object"
	| receiver |
	self action:  aValuable.
	receiver := aValuable receiver.
	"Clean blocks do not know the receiver"
	(receiver isNil and: [aValuable isBlock and: [ aValuable isClean ]]) ifTrue: [
		"A clean block was passed. In most cases, the block is defined statically, in this case we can discover
		the receiver. This fails if the block is defined in another method, we print a Warning"
		| contextOfSubsriptionSend |
		contextOfSubsriptionSend := thisContext sender sender.
		(contextOfSubsriptionSend method == aValuable compiledBlock method) 
			ifTrue: [ receiver := contextOfSubsriptionSend receiver ] 
			ifFalse: [ self traceCr: 'WARNING: can not find the receiver, use #when:do:for: in' ,contextOfSubsriptionSend method displayString ] ].
	self subscriber: receiver

```

But this does not work for method like e.g. SpExecutableLayout>>#whenChildrenAddedDo:

```
whenChildrenAddedDo: aBlock

	self announcer
		when: SpChildrenAdded
		do: aBlock

```

In practice, the blocks are never clean, as they at least do self sends or reference a temp...

What we could do:
- refactor in the system these cases to use when:do:for:  
- have the code above with the Transcript output for external code 
[...]

PR #12027, opened 2022-12-01
ConstantBlocks-Fix-Announcement-Tests 

MarcusDenker:
do not use constant blocks to test announcements


Ducasse:
Can you document in the tests why we have  [nil yourself] and not []?

MarcusDenker:
yes, I need to explain it better, especially as that is just an intermediate step to be able to turn on Constant Blocks. I am doing a pass

MarcusDenker:
I now changed it to:

"we do not use constant blocks here, instead use a clean block.
 	
 	Using Constant blocks with announcements makes no senese, as they do nothing.
 	
    For Clean Blocks, we have to the Announcments API as they  require to access the receiver of the home context, for clean blocks this has to be fixed, see https://github.com/pharo-project/pharo/issues/11220"


https://github.com/pharo-project/pharo/issues/11220 stays open, of course (this is just a workaround, the real fix comes with fixing that one for Clean Blocks)

PR #12030, opened 2022-12-01
Replacing baseColor method call from class to instance side

jordanmontt:
Instead of calling the method baseColor from the class side of UITheme, calling it from the instance.
This is a necessary step to after integrate the new color palette.


Ducasse (commented):
Why not UITheme current baseColor?

jordanmontt:
I don't know, if was an automatic refactor proposed by Calypso

Ducasse:
:) it is better to use the theme singleton from UI theme than from Smalltalk because we want to remove all the Smalltalk ui

jordanmontt:
Ok done. So we should remove the Calypso rule also. The one that says to use Smalltalk ui

Ducasse:
Indeed because this is really a bad rule

PR #12045, opened 2022-12-03
Fix 12044 introduces hexString and related tests

### LINKED ISSUE #12044: Introduce hexString as alias to hex
```
testHexString
	self assert: 0 hexString equals: '16r0'.
	self assert: 12 hexString equals: '16rC'.
	self assert: 1234 hexString equals: '16r4D2'.
	self assert: 11234 hexString equals: '-16r4D2'
```

Ducasse:


guillep:
But integers already have a method `hex` that does this!

Ducasse:
Yes I know. 
Now hexString makes clear that we get a string and not a number (of course we cannot have a number hex) 
and like that this is also coherent with the other hexString. 
I can understand that this is handy to type hex instead of hexString but it looks ugly. Because why Color asHexString is not the hex.

MarcusDenker:
Failing test:

IntegerTest>>testHexString

MarcusDenker:
Failing test not related:

- [CI] Fail on WIN only: MicFileResourceReferenceTest>>testFileReferenceExtensionMethod #12054
- [CI] ReleaseTest>>testUndeclared is failing: "MicFileIncluder in: MicRootBlock>>#includeFor:" #12052

PR #12047, opened 2022-12-03
12046 has non standard palette is always returning false

Ducasse:


jordanmontt (approved):

jordanmontt:
When Hernan will make his PR, we can delete the entire class of ColorForm because apparently is not used.

Ducasse:
In fact we should check because pablo told me that in Cairo he has to always transform. 
So may be Cairo is ARBG and form is RBGA (in addition I guess that what I removed in a kind of indexed palette).

PR #12055, opened 2022-12-06
11968-Clean-Blocks-Jobowner-assumes-valid-outerContext

### LINKED ISSUE #11968: [Clean Blocks] Job>>#owner assumes valid outerContext
**Bug description**
Job>>#owner needs an outerContext for the block is executes:

```
owner

	^ block outerContext receiver
```

We should think about how to avoid that. The problem is that we create the job with BlockClosure>>#asJob

```
asJob

	^ Job block: self
```

MarcusDenker:
Change Job to not rely on the homeContext receiver for the owner, but instead use the object that created the Job with #asJob.

In addion, add  #asJobWithOwner: to set it explicitly

### DIFF
--- a/src/Jobs/BlockClosure.extension.st
+++ b/src/Jobs/BlockClosure.extension.st
@@ -3,7 +3,7 @@ Extension { #name : #BlockClosure }
 { #category : #'*Jobs' }
 BlockClosure >> asJob [
 	"the owner is the object where the job was started, not where the block was created"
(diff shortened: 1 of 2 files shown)


MarcusDenker:
Not meant to be merged, more and idea how this could be solved

guillep (approved):
Ok.

Still, using thisContext (or closure home contexts) for something as trivial as a progress bar model seems like an overkill to me

PR #12063, opened 2022-12-07
11965-Implement-ContextactiveHome-without-using-home

### LINKED ISSUE #11965: Implement Context>>#activeHome without using #home
Blocks compiled without outer context (Clean Blocks or Blocks that have no non-local return) do not know, by definition, their home Context. But they know statically the home *method*.

This means that we can implement Context>>#activeHome for these blocks, it is defined as:

```
activeHome
	"If executing closure, search senders for the activation of the original
	 (outermost) method that (indirectly) created my closure (the closureHome).
	 If the closureHome is not found on the sender chain answer nil."

	| methodReturnContext |
	self isBlockContext ifFalse: [^self].
	self sender ifNil: [^nil].
	methodReturnContext := self methodReturnContext.
	^self sender findContextSuchThat: [:ctxt | ctxt = methodReturnContext]
```

Instead of using #methodReturnContext (which is an alias of #home), we can walk up the stack till we find a context which method is the homeMethod of the block.

MarcusDenker:
- implement #activeHome to not need outerContext
- add test
- refactor existing test

fixes #11965

#activeHome returns the #home if it is currently on the stack, the debugger uses that to check if a “save and proceed”
is possible when editing code in a Block. It is implemented to search up the sender chain until it finds the #home:

```Smalltalk
activeHome
	| methodReturnContext |
	self isBlockContext ifFalse: [^self].
	self sender ifNil: [^nil].
	methodReturnContext := self methodReturnContext.
	^self sender findContextSuchThat: [:ctxt | ctxt = methodReturnContext]
```

But it can be rewritten to do the same, but checking for the #homeMethod:

```Smalltalk
activeHome
	| homeMethod |
	self isBlockContext ifFalse: [^self].
	homeMethod := self homeMethod.
	^self findMethodContextSuchThat: [:ctxt | ctxt method == homeMethod]
```

With #homeMethod being implemented to delegate to the bock:

```Smalltalk
Context>>homeMethod
	"Answer the method in which the receiver was defined, i.e. the context from which an ^-return ] should return from. Note: implemented to not need #home"

	^ closureOrNil ifNil: [ self method ] ifNotNil: [ :closure | closure homeMethod ]
```

Where, if no #outerContext is available, it asks the CompiledBlock:

```Smalltalk
BlockClosure>>homeMethod
	"return the home method. If no #home is available due to no outerContext, use the compiledBlock"
	^ (self home
		   ifNotNil: [ :homeContext | homeContext ]
		   ifNil: [ self compiledBlock ]) method
```

Which uses the static  #outerCode chain (CompiledBlocks encode have a back-pointer to the enclosing block or method),
with #method following #outerCode until it reaches a CompiledMethod:

```Smalltalk
CompiledBlock>>method
	"answer the compiled method that I am installed in, or nil if none.”
	^self outerCode method
```

### DIFF
--- a/src/Kernel-Tests/ContextTest.class.st
+++ b/src/Kernel-Tests/ContextTest.class.st
@@ -140,13 +140,20 @@ ContextTest >> testActivateReturnValue [
 ContextTest >> testActiveHome [
 	| block |
 	"Active BlockContext Home And ActiveHome"
-	block := [ thisContext home == thisContext activeHome].
+	block := [ self. "to make it non-clean" thisContext home == thisContext activeHome].
+	self assert: block class equals: FullBlockClosure.
 	self assert: block value.
 	
 	"for Method Context"
 	self
 		assert: aMethodContext activeHome identicalTo: aMethodContext;
-		assert: aMethodContext activeHome identicalTo: aMethodContext home
+		assert: aMethodContext activeHome identicalTo: aMethodContext home.
+	
+	"non-active block context"
+	nonActiveBlockContext := self class returnNonActiveContextOfBlock.
+	self assert: (nonActiveBlockContext activeHome) isNil.
+		
+	
 ]
 
 { #category : #private }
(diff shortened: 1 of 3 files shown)


MarcusDenker:
Bootstrap error:

```

Instance of CleanBlockClosure did not understand #asExportedLiteral
CleanBlockClosure(Object)>>doesNotUnderstand: #asExportedLiteral
[ :e | e asExportedLiteral ] in HEMethod>>fillLiteralsAndBytecode: in Block: [ :e | e asExportedLiteral ]
Array(SequenceableCollection)>>collect:
HEMethod>>fillLiteralsAndBytecode:
HERing2ToHermesBuilder>>visitMethod:
```

I will move the test to the compiler package for now and open an issue for Hermes to support CleanBlocks

MarcusDenker:
Failing teset not related:

- [CI] test failing on CI: SystemDependenciesTest>>testExternalBasicToolsDependencies #12066

PR #12072, opened 2022-12-09
Factorize the list of VM parameters

privat:
Prevent obsolete comments or code by having it at one place.

Bonus:

* Spur in not optional anymore.
* Fix read-only/write information.
* Spell and tab fixes.


privat:
I thought that `SystemReporter` (from the package System-Reporter in this repository) was responsible for the list when you go Menu -> System -> System Reporter -> VM parameters.

In fact the responsible lives in newtools. https://github.com/pharo-spec/NewTools/pull/448
(And now I'm not sure what is the point of SystemReporter)

PR #12074, opened 2022-12-10
Create a release test so that instance side and class side methods are properly categorized

### LINKED ISSUE #11974: Create a release test so that instance side and class side initialize methods are properly categorized
- instance side #initialize method should be in "initialization"
- class side #initialize method should be in "class initialization"

and we already have lint rules for that

We have to fix a few cases in Spec2
 - in Spec2 https://github.com/pharo-spec/Spec/issues/1331
 - in NewTools https://github.com/pharo-spec/NewTools/issues/437
 - in Roassal https://github.com/ObjectProfile/Roassal3/issues/488
 - in Zinc https://github.com/svenvc/zinc/issues/104
 
but we should also make sure that no new cases get introduced into the standard image

So we should write a release test

astares:
Fix #11974


astares:
CI fails on Windows but this is a CI problem.

astares:
One test failing on OSX, unrelated

PR #12087, opened 2022-12-10
Add missing keyDown: message in FTSearchFunction

### LINKED ISSUE #11724: Browser method filter is broken
**Bug description**
In the system browser, when a class defines a lot of methods, you can go quickly to a method by starting to type the method name first letters as shown below.
Capture d’écran 2022-09-27 à 16 06 10

**To Reproduce**
1. Open the system browser
2. Select `AI-Algorithms-Graph` package
3. Select `AIBFS` class
4. Select the first method
5. Type `s`=> `start` method is selected
6.  Play a bit with method selection and filtering. After a while, the filter is not available anymore

fmqa:
Fixes #11724

From what I see, this was also missing in Pharo 10 as well, but I'm not quite sure why it works there.

### DIFF
--- a/src/Morphic-Widgets-FastTable/FTSearchFunction.class.st
+++ b/src/Morphic-Widgets-FastTable/FTSearchFunction.class.st
@@ -23,6 +23,11 @@ FTSearchFunction >> ghostText [
 ]
 
 { #category : #'event handling' }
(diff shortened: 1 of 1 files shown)


Ducasse:
Let us try and we will see :)

PR #12095, opened 2022-12-10
Fix class editor not refreshing when switching between class/instance sides

### LINKED ISSUE #12088: Fluid class definition is not updated when we click on class side.
Capture 2022-12-10 at 17 11 52
Capture 2022-12-10 at 17 11 44

fmqa:
Fixes #12088 

This checks whether the metalevel also matches within the validity computation. If the metalevel does not match, the tool validity is signalled as false.


Ducasse:
Indeed it works so I integrate it!!!!
Thanks for a lot for this cool contribution.

Issue #11191, opened 2022-05-05
PharoShortcuts>>#outdentShortcut not used


pavel-krivanek:
**Bug description**
Some shortcuts defined in PharoShortcuts class are not really used like `outdentShortcut` and others. It adds confusion where the shortcut is really defined (it is done in `RubTextEditor class>>#buildShortcutsOn:`)

PR #12096, opened 2022-12-11
RubTextEditor/TextEditor: Make ctrl+backspace delete the previous word

### LINKED ISSUE #2914: Under Windows the key to remove a word is not SHIFT+Backspace but CTRL + Backspace
	
Original case: https://pharo.fogbugz.com/f/cases/13882
Status: Work Needed
Project: Keymapping
Original Author: Vincent Blondeau
Date: 21 August 2014 4:30:15 pm
	
Description:
	
When you want to remove a word from an editor, the current Pharo keybinding is Shift+BackSpace and Shift+Suppr. But the default Windows binding is Ctrl+BackSpace and CtrlShift+Suppr.

fmqa:
Fixes #2914 

* Ctrl+Backspace now deletes the previous word
* Ctrl+Delete now deletes the next word (note that this never worked, even with the previous key-binding shift+delete due to an off-by-one error)
* Updates PharoShortcuts accordingly (I am aware that they are currently unused - this is noted by #11191, but it is important to keep these updated as we may want to use them at some point)

### DIFF
--- a/src/Tool-Base/PharoShortcuts.class.st
+++ b/src/Tool-Base/PharoShortcuts.class.st
@@ -174,7 +174,7 @@ PharoShortcuts >> delShortcut [
 { #category : #'keymaps - editor edit' }
 PharoShortcuts >> delWordShortcut [
 
-	^ OSPlatform current isMacOS
+	^ (OSPlatform current isMacOS)
 		ifTrue: [ Character $d ctrl asShortcut ]
 		ifFalse: [ Character delete ctrl asShortcut ]
 ]
--- a/src/Tool-Base/PharoShortcuts.class.st
+++ b/src/Tool-Base/PharoShortcuts.class.st
@@ -175,7 +175,7 @@ PharoShortcuts >> delShortcut [
 PharoShortcuts >> delWordShortcut [
 
 	^ (OSPlatform current isMacOS)
-		ifTrue: [ Character $d ctrl asShortcut ]
+		ifTrue: [ $d ctrl asShortcut ]
 		ifFalse: [ Character delete ctrl asShortcut ]
 ]
 
--- a/src/Tool-Base/PharoShortcuts.class.st
+++ b/src/Tool-Base/PharoShortcuts.class.st
@@ -249,7 +249,8 @@ PharoShortcuts >> forwardLineShortcut [
 
 { #category : #'keymaps - editor navigation' }
 PharoShortcuts >> forwardWordShortcut [
-	^ Character arrowRight ctrl
+	"Equivalent to: ctrl+right on Windows/Linux, option/alt+right on OSX"
+	^ Character arrowRight meta
 ]
 
 { #category : #'keymaps - editor style' }
--- a/src/Tool-Base/PharoShortcuts.class.st
+++ b/src/Tool-Base/PharoShortcuts.class.st
@@ -60,6 +60,7 @@ PharoShortcuts >> backwardWordDeleteShortcut [
 
 { #category : #'keymaps - editor navigation' }
 PharoShortcuts >> backwardWordShortcut [
+	"Equivalent to: ctrl+left on Windows/Linux, option/alt+left on OSX"
 	^ Character arrowLeft meta
 ]
 
--- a/src/Tool-Base/PharoShortcuts.class.st
+++ b/src/Tool-Base/PharoShortcuts.class.st
@@ -174,7 +174,9 @@ PharoShortcuts >> delShortcut [
 { #category : #'keymaps - editor edit' }
 PharoShortcuts >> delWordShortcut [
 
-	^ Character delete ctrl asShortcut
+	^ OSPlatform current isMacOS
+		ifTrue: [ Character $d ctrl asShortcut ]
+		ifFalse: [ Character delete ctrl asShortcut ]
 ]
 
 { #category : #'keymaps - code' }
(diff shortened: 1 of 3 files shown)


Ducasse:
Thanks we will have a look. This is great to see some help on this side.

NSUSpray:
I am not familiar with macOS. Shouldn't the Ctrl option be Windows-only? If Shift is not macOS specific, then why was Shift used?

fmqa:
Shift is probably a mistake. Never seen a platform that uses it. OSX uses
"Option" (Alt) instead:
https://www.macworld.com/article/234442/how-to-delete-text-in-all-sorts-of-ways-on-your-mac.html

We need to reach a consistent state first. Right now, word navigation is
bound to ctrl+left / ctrl+right where word deletion is shift+backspace /
shift+delete which makes no sense regardless of platform.

Ideally we want to put the platform-specific decision logic into
PharoShortcuts but Rubric has to be refactored to use it first

Spray ***@***.***> schrieb am Mo., 12. Dez. 2022, 03:12:

src/Tool-Base/PharoShortcuts.class.st
```diff
@@ -249,7 +249,7 @@ PharoShortcuts >> forwardLineShortcut [
 
 { #category : #'keymaps - editor navigation' }
 PharoShortcuts >> forwardWordShortcut [
-	^ Character arrowRight meta
+	^ Character arrowRight ctrl
```
VincentBlondeau: I'm not sure that this change will work on Mac.

src/Tool-Base/PharoShortcuts.class.st
```diff
@@ -462,7 +462,7 @@ PharoShortcuts >> selectBackwardLineShortcut [
 
 { #category : #'keymaps - editor selection' }
 PharoShortcuts >> selectBackwardWordShortcut [
-	^ Character arrowLeft meta shift
+	^ Character arrowLeft shift ctrl
```
VincentBlondeau: I'm not sure that this change will work on Mac.

src/Tool-Base/PharoShortcuts.class.st
```diff
@@ -479,7 +479,7 @@ PharoShortcuts >> selectForwardLineShortcut [
 
 { #category : #'keymaps - editor selection' }
 PharoShortcuts >> selectForwardWordShortcut [
-	^ Character arrowRight shift meta
+	^ Character arrowRight shift ctrl
```
VincentBlondeau: I'm not sure that this change will work on Mac.

src/Tool-Base/PharoShortcuts.class.st
```diff
@@ -174,7 +174,7 @@ PharoShortcuts >> delShortcut [
 { #category : #'keymaps - editor edit' }
 PharoShortcuts >> delWordShortcut [
 
-	^ Character delete shift asShortcut
+	^ Character delete ctrl asShortcut
```
VincentBlondeau: Good for Windows. Not sure about Mac support.

src/Tool-Base/PharoShortcuts.class.st
```diff
@@ -55,7 +55,7 @@ PharoShortcuts >> backwardLineShortcut [
 { #category : #'keymaps - editor edit' }
 PharoShortcuts >> backwardWordDeleteShortcut [
 
-	^ Character backspace shift
+	^ Character backspace ctrl
```
VincentBlondeau: Good for Windows. Not sure about Mac support.

src/Tool-Base/PharoShortcuts.class.st
fmqa: > Control-D: Delete the character to the right of the insertion point. Or use Fn-Delete.
> Fn-Delete: Forward delete on keyboards that don't have a Forward Delete   key. Or use Control-D.

Not sure how to map this

fmqa: We can make it conditional for now.

src/Tool-Base/PharoShortcuts.class.st
fmqa: Changed + Added comment

src/Tool-Base/PharoShortcuts.class.st
fmqa: Changed + Added comment

src/Tool-Base/PharoShortcuts.class.st
fmqa: Changed + Added comment

src/Tool-Base/PharoShortcuts.class.st
fmqa: Changed + Added comment

src/Tool-Base/PharoShortcuts.class.st
```diff
@@ -249,6 +251,7 @@ PharoShortcuts >> forwardLineShortcut [
 
 { #category : #'keymaps - editor navigation' }
 PharoShortcuts >> forwardWordShortcut [
+	"Equivalent to: ctrl+right on Windows/Linux, option/alt+right on OSX"
```
NSUSpray, line 255: It seems you forgot to add the same comment for the _**backward**WordShortcut_ method.

NSUSpray:
It works fine for me.

src/Tool-Base/PharoShortcuts.class.st
fmqa: Thanks for catching this. On my phone now (holidays...) but this can be fixed quickly.

fmqa:
I think this is about ready at this point. Would be great if Dr. @VincentBlondeau or an OSX user signs off on it.

Ducasse:
Broken tests are unrelated.

PR #12117, opened 2022-12-16
Introduce release test to check that "selectors send but not implemented" cases do not grow

### LINKED ISSUE #12116: Introduce release test to check that  "selectors send but not implemented" cases do not grow
We have many unimplemented calls in the image. Fixing them is a long term cleanup. 
See https://github.com/pharo-project/pharo/issues/10834 for more details.

But we should already introduce a release test that checks for the number of cases does not 
go up and so (with #10834) can go down over time to reach zero. This way we can avoid that 
new cases get introduced and we can improve for better quality.

astares:
Fix #12116


astares:
CI Failing is unrelated

astares:
Any reviews?

gcotelli (approved):

MarcusDenker (approved):

PR #12130, opened 2022-12-19
fixes #11997: Improving Array2d API 

### LINKED ISSUE #11997: Improving Array2d API
```
Array2DTest >> testRowsColumnsPut
	| m |
	m := Array2D rows: 3 columns: 2.
	m rowsColumnsPut: [ :row :column | column * 10 + row ].
	self assert: (m at: 1 at: 1) equals: 11.
	self assert: (m at: 1 at: 2) equals: 21.
	self assert: (m at: 2 at: 1) equals: 12.
	self assert: (m at: 2 at: 2) equals: 22.
	self assert: (m at: 3 at: 1) equals: 13.
	self assert: (m at: 3 at: 2) equals: 23.
```

n1toxyz:
Implements the proposed fix by
- implementing #rowsColumnsDo:, #rowsColumnsPut: in Array2D
- refactoring #rows:columns:tabulate: in Array2D class
- implementing #testRowsColumnsDo, #testRowsColumnsPut in Array2DTest

### DIFF
--- a/src/Collections-Sequenceable-Tests/Array2DTest.class.st
+++ b/src/Collections-Sequenceable-Tests/Array2DTest.class.st
@@ -251,6 +251,38 @@ Array2DTest >> testRowVector [
 	self assert: m2 numberOfRows equals: 1.
 ]
 
+{ #category : #tests }
+Array2DTest >> testRowsColumnsDo [
+
+	| m col |
+	m := Array2D rows: 3 columns: 2.
+	m rowsColumnsPut: [ :row :column | column * 10 + row ].
+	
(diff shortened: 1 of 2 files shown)


Ducasse:
Hi 

Array2D should be removed from the core and we should favor http://github.com/pharo-containers/
Can you do a PR on this version?
We should stop thinking that Pharo should contain all the datastructures of the world but think more modularly. 
So we should remove array2d from Pharo 11.
S.

Ducasse:
I opened an issue in https://github.com/pharo-containers/Containers-Array2D/issues/14

PR #12135, opened 2022-12-20
Calypso: self in <script:> is the selected class

### LINKED ISSUE #12107: Calypso: Self on a <script:> pragma is the origin trait instead of the user class
**Bug description**

When a trait method defines a <script:> and you browse a trait user in Calypso and execute this script using the icon button, self refers to the trait instead of the user.

I think nobody would expect that a script defined on a trait is really executed on the trait and not on the user class. But I may be missing a scenario.

**To Reproduce**

1. Open Transcript.
2. Evaluate to read a traced "TraitWithScript": 
```
t := Trait named: #TraitWithScript.
t class compile: 'traceYourself
	<script: ''self traceYourself'' >
	self traceCr'.
t traceYourself.
```
3. Evaluate to read a traced "OneUserOfScripts", as I'd expect:
```
u := Object subclass: #OneUserOfScripts uses: t.
u traceYourself.
```
4. Evaluate `u class browse` and click on the method icon to run script, to read a traced "TraitWithScript" as I wouldn't expect.

**Expected behavior**
I expected that "self" in a <script: 'self' > pragma is bound to the method class instead of the method origin.

**Version information:**
 Pharo 11, probably others too.

**Expected development cost**
The main question is the impact on existing pragmas on traits that assume that self is the trait instead of the user. I didn't query it on the image, but anyway external projects are affected, of course. **What's your opinion?**

The solution would be simple: change `ClyClassScript>>#implementorClass` to answer  `implementorMethod methodClass` instead of `implementorMethod origin`

tinchodias:
To fix #12107

Also renamed an argument anObject->aCompiledMethod, that's unrelated but so handy to do it...

### DIFF
--- a/src/Calypso-SystemPlugins-ClassScripts-Queries/ClyClassScript.class.st
+++ b/src/Calypso-SystemPlugins-ClassScripts-Queries/ClyClassScript.class.st
@@ -94,7 +94,7 @@ ClyClassScript >> iconName [
 
 { #category : #accessing }
 ClyClassScript >> implementorClass [
-	^ implementorMethod origin
+	^ implementorMethod methodClass
 ]
 
 { #category : #accessing }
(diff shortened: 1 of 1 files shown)


astares (approved):

MarcusDenker:
Not related: 

MarcusDenker (approved):

Issue #12131, opened 2022-12-19
Introduce a non-misleading not equals selector: #!=


demarey:
In Smalltalk 80, `~=` was chosen as selector to answer whether the receiver and argument do not represent the same object.
 `~=` is misleading since `~` is used in mathematics to describe proportionality or approximation (see https://en.wikipedia.org/wiki/Approximation).
I propose to introduce `!=`, in addition to `~=`, and use it everywhere in Pharo standard image. This selector is already used in well known languages for this purpose and is more meaningful, especially for people learning Smalltalk.
For coherence, I propose to use `!==` instead of `~~`.

demarey:
See related Phep: https://github.com/pharo-project/pheps/pull/12

astares:
If we want to align with mathematics then "3+4" would need to be done different in Smalltalk too.

We would have to do it without objects and do it as in other languages/mathemtics were both side equally treated..
But as we are aware 
 3 + 4
 4 +3 
 is different due to the different receiver of the message #+
 
 For the topic of supporting != it would then also mean to support <> as well as for instance in SQL you
 can have that also for NOT EQUAL
 
 
![image](https://user-images.githubusercontent.com/5980033/212373134-82a5b181-aaf4-4097-ba85-2da7194b2ff8.png)

PR #12139, opened 2022-12-20
12131 introduce a non misleading not equals selector 

demarey:


Ducasse:
Hi christophe 
this is really a change that introduces incompatibility. So at least we should keep ~= 
now my question is how do we express identity difference because it is ~~

MarcusDenker:
this will not be that easy:

some package are meant to run on quite old Pharo versions (all the Zinc packages, Fuel)

So this can not be done for those... 

and yes, we have to keep the compatibility methods, but that was always the idea I think.

What has to be checked is special selectors:

BytecodeEncoder class>>#specialSelectors

```
	^ #(#+ #- #< #> #'<=' #'>=' #= #'~=' #* #/ #'\\' #@ #bitShift: #'//' #bitAnd: #bitOr: #at: #at:put: #size #next #nextPut: #atEnd #'==' nil "class" #'~~' #value #value: #do: #new #new: #x #y)
```

and EncoderForSistaV1>>#genSendSpecial:numArgs: shows that there is even a bytecode for ~=, which means that this would have impact on performance, too

MarcusDenker:
To me this is a typical project of a complexity that would be good for a "Pharo Enhancement Proposal" (see https://github.com/pharo-project/pheps) as it has lots of non-trivial implications




<blockquote><div> GitHub</div><div><strong><a href="https://github.com/pharo-project/pheps">GitHub - pharo-project/pheps: Pharo Enhancement Proposals</a></strong></div><div>Pharo Enhancement Proposals. Contribute to pharo-project/pheps development by creating an account on GitHub.</div></blockquote>

demarey:
> Hi christophe this is really a change that introduces incompatibility. So at least we should keep ~= now my question is how do we express identity difference because it is ~~

Yes, this is the idea. There is more information on the issue: https://github.com/pharo-project/pharo/issues/12131.

> To me this is a typical project of a complexity that would be good for a "Pharo Enhancement Proposal" (see https://github.com/pharo-project/pheps) as it has lots of non-trivial implications

I just started one: https://github.com/pharo-project/pheps/pull/12

<blockquote><div> GitHub</div><div><strong><a href="https://github.com/pharo-project/pheps">GitHub - pharo-project/pheps: Pharo Enhancement Proposals</a></strong></div><div>Pharo Enhancement Proposals. Contribute to pharo-project/pheps development by creating an account on GitHub.</div></blockquote>

MarcusDenker:
I close this for now (it has conflicts)

astares:
I would not do != just to align with mainstream, because then you would also have to support <> for NOT EQUAL. 

So the question is where do stop then:

image

PR #12142, opened 2022-12-20
9932-RegistryThreshold-in-Socket-seems-to-be-unused

### LINKED ISSUE #9932: RegistryThreshold in Socket seems to be unused
**Describe the problem**
RegistryThreshold in Socket seems to be unused... it has accessors, but these have no senders.

Maybe it is accessed from the VM? We should add a note or remove it.

MarcusDenker:
remove unused registryThreshold class var and accessors. fixes #9932


astares (approved):

astares:
We could have deprecated it - but as it is unused it makes sense to remove directly

Issue #12081, opened 2022-12-10
Another duplicates of itemchooser


Ducasse:
DialogItemsChooserUI should not be used anymore. SpChooserPresenter should be used instead. 
![Capture 2022-12-10 at 14 44 12](https://user-images.githubusercontent.com/432549/206858308-ba92da4f-7298-447d-9379-52ec30273e03.png)

PR #12146, opened 2022-12-22
12081 another duplicates of itemchooser

n1toxyz:
Fixes issue #12081 as proposed.


n1toxyz:
I noticed that I don't need a separate Dialog class in Spec as #openDialog already opens a presenter in a dialog window.
Thus we can replace two classes with just a few lines.

Ducasse:
Thanks this is cool. Ideally we should rewrite the Finder to use Spec2.0. But this is good to have one less chooser in the image!

Ducasse:
I will relaunch the build because it is broken on all platforms and this is not related to the change.

Ducasse:
This is cool to see less code :). Tx!

PR #12167, opened 2022-12-30
New rule to remove unnecessary separators and dots

### LINKED ISSUE #12166: [Feature Request] New rule to remove unnecessary separators and dots
Many methods contain tabs or spaces at the end of each code line. Also some final dot that is not necessary.

We should have a lint rule detecting such methods together with an autofix option

astares:
Fix #12166

### DIFF
--- a/src/Renraku/ReCompactSourceCodeCritic.class.st
+++ b/src/Renraku/ReCompactSourceCodeCritic.class.st
@@ -1,0 +1,30 @@
+"
+A code critique to take care of unnecessary seperators at end of source code lines and unnecessary dot at end of method
+"
+Class {
+	#name : #ReCompactSourceCodeCritic,
+	#superclass : #ReAbstractCritique,
+	#category : #'Renraku-Critiques'
+}
+
+{ #category : #'instance creation' }
+ReCompactSourceCodeCritic class >> for: anEntity selector: aSelector by: aRule [
+	^ (self for: anEntity by: aRule)
+		tinyHint: aSelector;
+		yourself
+]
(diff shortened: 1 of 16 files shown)


astares:
image

astares:
CI issue on Windows is unrelated

src/Renraku/ReCompactSourceCodeCritic.class.st
```diff
@@ -0,0 +1,30 @@
+"
+A code critique to take care of unnecessary seperators at end of source code lines and unnecessary dot at end of method
```
vonbecmann: separators is misspelled

astares: fixed

vonbecmann (approved):

vonbecmann (approved):

Ducasse:
I like that it fixes the problem and do it without reformatting the code :) super cool.

demarey:
remove unnecessary separators is really nice!
On the other hand, I do not like removing trailing dot. All expressions can end with a dot. Removing the last dot in a method forces the writer to remember to add a missing dot before adding new expressions to a method. It is easier to always put the dot.

astares:
@demarey 

1. If you do not like it, then ban the rule for your package. Simple.

2. But finishing with a dot every method just for the case later there might eventually (who really knows?) be another expression I would call "premature dotting". 

See a detailed discussion on this in the thread:   https://twitter.com/TorstenAstares/status/1609666189382254593

PR #12178, opened 2022-12-31
Formatter adds unnecessary empty line before (first) method comment

### LINKED ISSUE #12175: Formatter adds unnecessary empty line before (first) method comment
Write a simple unformatted method including a method comment like:
```Smalltalk
foo
"This is a simple method" ^3+4
```
and you **format it** then you will get

```Smalltalk
foo

    "This is a simple method"

     ^ 3 + 4
```

with an additional empty newline between the selector and the comment. This is annoying.

There would be no such line if Pharo default formatter settings would follow the usual Smalltalk systems/Smalltalk books and the Pharo with Style guide (see https://github.com/SquareBracketAssociates/Booklet-PharoWithStyle/blob/master/Chapters/PWS-Guidelines-1.pdf) where the method comment is usually right in the second line after the selector of the method.

image

So it should better format into 
```Smalltalk
foo
    "This is a simple method"

     ^ 3 + 4
```
without the additional line between the selector and the first comment.

astares:
- introduce an additional setting #numberOfNewLinesAfterMethodSignatureWithMethodComment beside existing #numberOfNewLinesAfterMethodSignature

Fix #12175

### DIFF
--- a/src/EnlumineurFormatter/EFContext.class.st
+++ b/src/EnlumineurFormatter/EFContext.class.st
@@ -42,7 +42,8 @@ Class {
 		'numberOfNewLinesAfterMethodComment',
 		'newLinesAfterMethodSignature',
 		'periodAtEndOfMethod',
-		'periodAtEndOfBlock'
+		'periodAtEndOfBlock',
+		'newLinesAfterMethodSignatureWithMethodComment'
 	],
 	#category : #'EnlumineurFormatter-Core'
 }
@@ -285,6 +286,17 @@ EFContext >> numberOfNewLinesAfterMethodSignature [
 { #category : #accessing }
 EFContext >> numberOfNewLinesAfterMethodSignature: aBoolean [
 	newLinesAfterMethodSignature := aBoolean
+]
+
+{ #category : #accessing }
+EFContext >> numberOfNewLinesAfterMethodSignatureWithMethodComment [
+
+	^ newLinesAfterMethodSignatureWithMethodComment
+]
+
+{ #category : #accessing }
+EFContext >> numberOfNewLinesAfterMethodSignatureWithMethodComment: aBoolean [
+	newLinesAfterMethodSignatureWithMethodComment := aBoolean
 ]
 
 { #category : #accessing }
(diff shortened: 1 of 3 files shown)


astares:
We can now distinguish in the settings:

image

- between methods with a comment (where we want to only have 1 newline to directly follow in the next line after method signature) 
- and methods without a comment (where we want to only have 2 newlines to have an empty line after the method signature before the method body starts)

astares:
Tests are green in image:

image

Lets see what the CI tells us.

Ducasse:
Tx. I'm always annoyed by the extra line before the first comment. I should have fixed it long time ago.

jordanmontt (approved):

src/EnlumineurFormatter/EFContext.class.st
```diff
@@ -287,6 +288,17 @@ EFContext >> numberOfNewLinesAfterMethodSignature: aBoolean [
...
 
+{ #category : #accessing }
+EFContext >> numberOfNewLinesAfterMethodSignatureWithMethodComment [
+
+	^ newLinesAfterMethodSignatureWithMethodComment
+]
+
+{ #category : #accessing }
+EFContext >> numberOfNewLinesAfterMethodSignatureWithMethodComment: aBoolean [
+	newLinesAfterMethodSignatureWithMethodComment := aBoolean
```
gcotelli:
```suggestion
EFContext >> numberOfNewLinesAfterMethodSignatureWithMethodComment: anInteger [
	newLinesAfterMethodSignatureWithMethodComment := anInteger
```

gcotelli (changes requested):

src/EnlumineurFormatter/EFContext.class.st
astares: @gcotelli Good catch!
But then this is already wrong in the setting that was there before (newLinesAfterMethodSignature) which I just used as template.

I will change it

astares: @gcotelli fixed in https://github.com/pharo-project/pharo/pull/12178/commits/00831674cee35ea901bb14f75fd17f03c83a2fda

gcotelli (approved):

astares:
@gcotelli / @MarcusDenker Thanks!

PR #12182, opened 2023-01-01
Recursion found

### LINKED ISSUE #12180: Recursion found
- open critiques browser on a package
- add a clean code critique and search for cases
- click on "Transform"

Pharo hangs with a recursion

image

astares:
Fix #12180


jordanmontt:
I think we should eventually remove this tool. There is the new version that we should add to NewTools.

astares:
@jordanmontt yes - but firt we should just delete the strange recursion to make it work again and then work on other topics.

jordanmontt (approved):

astares:
CI only fails on Windows which is unrelated.

hogoww (approved):

PR #12185, opened 2023-01-01
Cleanup: Condense AST- packages

### LINKED ISSUE #12184: Cleanup: Condense AST- packages
- using the cleanup to remove unnecessary tabs, spaces and final dots in methods

astares:
Fix #12184


astares:
Reloads into a fresh image and tests are green. 

image

Ready for a review and integration

gcotelli (approved):

hogoww (approved):
Basic cleanup, looks good to me!

MarcusDenker:
RBExtractMethodToComponentParametrizedTest>>testMoveWithoutSelfReference is failing, we should check that

astares:
The range in this test was dependent on the size of source code in RBReadBeforeWrittenTester>>#copyDictionary:
which is a method from the "AST-Core-Visitors" package.

Fixed the test now. Lets see what CI will tell us

hogoww:
Looks good.
May I ask why that -1? 
I'm curious.
Is that because you updated a method used by a test which needed to be "badly" formatted ?

Edit: Too slow :D Thanks !

astares:
@hogoww Reason is simple: the test method #testMoveWithoutSelfReference in RBExtractMethodToComponentParametrizedTest is kind of simulating an extract method refactoring.  

It is (ab)using RBReadBeforeWrittenTester>>#copyDictionary: method for that from the package "AST-Core-Visitors" that I canged.

You see that the "extract range" is being setup right in the beginning of the test method:

image
 
So before the cleaning (for instance in a Pharo 10 image) you have the source code length for this method with a size of **286** as it is written in the test 
 
 ````Smalltalk
 (RBReadBeforeWrittenTester>>#copyDictionary:) sourceCode size "286"
 ```` 
 
So in Pharo 10 it was extracting this part from the source code of the method:

image

 
 As my PR now cleaned there was an unnecessary space after the selector that got removed. See the diff:
 
 
image
 
So the new source code size for is 285 as a maximum now in Pharo 11.

Thats why the test failed because it tried to simulate a selected range exceeding the max size of characters in the source code string - so 286 would exceed the bounds:
 
image

So fixing by adopting the test with -1 and adopting from 286 to 285 was the obvious solution to solve it:

image

hogoww:
@astares I had understood from the answer previous to my question, but thank you for the details !

Issue #12127, opened 2022-12-19
[CI]  With latest VM: Segmentation fault when running all tests


MarcusDenker:
**Bug description**
Every second CI rum (that is, when it uses the -lastest vm) leads to a Segmentation fault on all three platforms. See e.g.

https://ci.inria.fr/pharo-ci-jenkins2/blue/organizations/jenkins/Test%20pending%20pull%20request%20and%20branch%20Pipeline/detail/PR-12069/3/pipeline

MarcusDenker:
It works now on Linux, but Win and Mac still show the error, see e.g.

https://ci.inria.fr/pharo-ci-jenkins2/blue/organizations/jenkins/Test%20pending%20pull%20request%20and%20branch%20Pipeline/detail/PR-12216/1/pipeline

<blockquote><div><strong><a href="https://ci.inria.fr/pharo-ci-jenkins2/blue/organizations/jenkins/Test%20pending%20pull%20request%20and%20branch%20Pipeline/detail/PR-12216/1/pipeline">Jenkins</a></strong></div></blockquote>

MarcusDenker:
Now even Linux fails again: 

https://ci.inria.fr/pharo-ci-jenkins2/blue/organizations/jenkins/Test%20pending%20pull%20request%20and%20branch%20Pipeline/detail/PR-12220/3/pipeline/46

<blockquote><div><strong><a href="https://ci.inria.fr/pharo-ci-jenkins2/blue/organizations/jenkins/Test%20pending%20pull%20request%20and%20branch%20Pipeline/detail/PR-12220/3/pipeline/46">Jenkins</a></strong></div></blockquote>

guillep:
Hi Marcus,

The crash should be consistent in all platforms. I wonder why it worked in linux earlier :).
I don't think there is a need to follow up all the traces of the CI.
We have investigated the problem, it is in the latest unstable VM, due to a change integrated in december (https://github.com/pharo-project/pharo-vm/pull/518). Yesterday we merged a fix (https://github.com/pharo-project/pharo-vm/pull/524) but the build failed due to a network issue. We are relaunching the build right now, it should be solved during the day.

MarcusDenker:
The cases with Linux working did download the -latest vm, I see in the logs:

Downloading the latest pharoVM:
	http://files.pharo.org/get-files/110/pharo-vm-Linux-x86_64-latest.zip

MarcusDenker:
[...]

PR #12197, opened 2023-01-02
checkIfMustRefresh in OmBlockFileStore>>blockForLocalNameAsInteger:

### LINKED ISSUE #12196: OmBlockFileStore>>entryFor:ifPresent:ifAbsent: quietly returns partial results
**Bug description**
When reading a large ombu file `OmBlockFileStore>>entryFor:ifPresent:ifAbsent:` may only return partial results.

The bug is due to `#checkIfMustRefresh` not being called when searching for an entry.

Note that this bug will not appear if another operation has previously caused the entire file metadata to be loaded.

**To Reproduce**
The following test will trigger the problem:

```
testEntryForIfPresentIfAbsent
	"Check that #entryFor:ifPresent:ifAbsent: loads the file past the first block"
	| ombuContent ombuFile mr1 entry mr2 |

	ombuContent := (self class class >> #multiBlockOmbu) comments second.
	ombuFile := self directoryForTests / 'bigStore.ombu'.
	self deny: ombuFile exists.
	ombuFile writeStreamDo: [ :stream |
		stream << ombuContent ].
	bigStore := OmBlockFileStore
		named: 'bigStore'
		inDirectory: ombuFile parent.
	targetReference := OmReference 
		globalName: 'bigStore'
		localName: '199'.
	mr1 := bigStore mustRefresh.
	self assert: mr1
		description: 'mustRefresh should be true prior to search'.
	entry := bigStore
		entryFor: targetReference
		ifPresent: [ :found | found ]
		ifAbsent: [ nil ].
	mr2 := bigStore mustRefresh.
	self assert: entry isNotNil
		description: 'entry not found'.
	self assert: mr2 not
		description: 'mustRefresh should be false after to search'.
```

Note that this won't run as it is missing supporting methods (but a PR is on the way).

**Expected behavior**
The entire ombu file metadata is loaded in to `OmBlockFileStore` and the entry found.

**Screenshots**
N/A.

**Version information:**
 - OS: Ubuntu
 - Version: 22.04
 - Pharo Version: 10

**Expected development cost**
PR on the way.

**Additional context**
N/A.

akgrant43:
to ensure the internal state is up to date prior to retrieving the requested block.

Fixes: https://github.com/pharo-project/pharo/issues/12196

### DIFF
--- a/src/Ombu-Tests/OmBlockFileStoreTest.class.st
+++ b/src/Ombu-Tests/OmBlockFileStoreTest.class.st
@@ -1,8 +1,6607 @@
 Class {
 	#name : #OmBlockFileStoreTest,
 	#superclass : #OmFileStoreTest,
+	#instVars : [
+		'targetReference',
+		'bigStore'
+	],
 	#category : #'Ombu-Tests'
 }
+
+{ #category : #accessing }
+OmBlockFileStoreTest class >> multiBlockOmbu [
+	"The second comment in this method is used to generate test data for #testEntryForIfPresentIfAbsent"
+"OmEntry {
+	#tags : {
+		#author : 'AlistairGrant',
+		#time : DateAndTime [ '2023-01-02T14:26:40.404694+01:00' ],
+		#prior : OmReference [ 'Pharo10-SNAPSHOT-64bit-14f5413.8z1nmd2yzd4bxueyxpz9rzxql', '5' ],
+		#self : OmReference [ '1' ]
+	},
+	#content : EpMethodAddition {
+		#method : RGMethodDefinition {
+			#annotations : IdentityDictionary {
+				#className : #OmBlockFileStore,
+				#isMetaSide : false
+			},
+			#name : #internalStateIssue,
+			#protocol : #'test data',
+			#sourceCode : 'internalStateIssue\r\r\t^ 1',
+			#stamp : 'AlistairGrant 1/2/2023 14:26',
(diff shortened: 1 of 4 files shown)


MarcusDenker:
Failing test:

ProperlyImplementedSUnitClassesTest>>testAndMakeSureSuperTearDownIsCalledAsLastMessageInTearDownMethodsOfTestCases

akgrant43:
Hi Marcus,

Thanks for the feedback.

I put a comment in the method about why the `super tearDown` was placed at the top of the method - placing the `tearDown` at the bottom results in:

```
        "OmFileStoreTest>>tearDown"
	store fileReference ensureDelete.
	otherStore fileReference ensureDelete.
       "OmStoreTest>>tearDown"
	store flush.
	otherStore flush.
```

which, while harmless, doesn't make any sense.

However, given that it is harmless, I've just removed the change.

akgrant43:
The CI failure appears to be unrelated to this PR (same VM crash as in several other recent PRs).

MarcusDenker:
Yes, we have the problem that every second CI run uses the newest VM, but that VM has a bug.

see https://github.com/pharo-project/pharo/issues/12127

MarcusDenker:
Failing test not related:

MarcusDenker (approved):
The only thing I do not like that much is the huge comment... but then, we have test data like that already in the system and by being a comment it is on disk only

PR #12202, opened 2023-01-02
ReCompactSourceCodeCritic should change origin method - not methodClass

### LINKED ISSUE #12201: ReCompactSourceCodeCritic should change origin method - not methodClass
**Bug description**
In #12190 @gcotelli found a bug that #ffiLibraryName was added although ReCompactSourceCodeCritic should have no code changes other than removing unnecessary tabs, spaces and final dots.

I now found out that it inlined #ffiLibraryName from a trait inheritance (from TCairoLibrary) as there is a bug left in ReCompactSourceCodeCritic as it wrongly uses the compiled methods #methodClass instead of the #origin. This causes trouble when applying the auto-fix to methods inherited from traits. 

For the case of a compiled method that is answering #isFromTrait with true (which means it is inherited from a trait) then the auto repair will compile the change in the class and not the trait where the method was originally defined.

**To Reproduce**
1. Create a package "**ZZZ**"
2. Create a trait **TBar**
3. Create a method **#bar** in TBar with an unnecessary dot, tab or space - for example:

```Smalltalk
bar

	^ 3+4. 
```

4. Create a class **Foo** that uses the trait **TBar**

```Smalltalk
Object subclass: #Foo
	uses: TBar
	instanceVariableNames: ''
	classVariableNames: ''
	package: 'ZZZ'
```

5. Now in Calypso select the method #bar while still having the class Foo selected.
6. Click on "automatic resolve the issue" for the lint rule that laments on "Unnecessary separators or final dots found in method body"

NOW THE SYSTEM WRONGLY SUGGESTS TO FIX THE ISSUE IN **Foo>>#bar** ALTHOUGH THE PROBLEM IS IN **TBar>>#bar** WHERE #bar WAS ORIGINALLY DEFINED

**Expected behavior**
The expected behavior is that the issue is resolved in the trait for trait inherited methods

That is easy to fix by changing ReCompactSourceCodeCritic>>#change 
from

```Smalltalk
change
	| cm |
	cm := self sourceAnchor sourceEntity.

	^ (RBRefactoryChangeFactory instance)
		addMethodSource: (ReMethodSourceCleaner clean: cm sourceCode) in: cm methodClass
```
to 

```Smalltalk
change
	| cm |
	cm := self sourceAnchor sourceEntity.

	^ (RBRefactoryChangeFactory instance)
		addMethodSource: (ReMethodSourceCleaner clean: cm sourceCode) in: cm origin
```
so the method will be fixed on the origin of definition always. 

Either 
- in the class defining the method or 
- for train inherited method in the trait that defined the trait method

astares:
Fix #12201

### DIFF
--- a/src/Renraku/ReCompactSourceCodeCritic.class.st
+++ b/src/Renraku/ReCompactSourceCodeCritic.class.st
@@ -20,7 +20,7 @@ ReCompactSourceCodeCritic >> change [
 	cm := self sourceAnchor sourceEntity.
 
 	^ (RBRefactoryChangeFactory instance)
-		addMethodSource: (ReMethodSourceCleaner clean: cm sourceCode) in: cm methodClass
+		addMethodSource: (ReMethodSourceCleaner clean: cm sourceCode) in: cm origin
 ]
 
 { #category : #testing }


astares:
CI failing is unrelated

Please review and integrate so I can move on with the cleaning of more packages

MarcusDenker:

MarcusDenker (approved):

PR #12205, opened 2023-01-03
12128-Poor-performance-with-large-WriteStreams

### LINKED ISSUE #12128: Poor performance with large WriteStreams
`WriteStream>>pastEndPut:` limits the growth of the `WriteStream` to 1,000,000 slots. This causes bad performance when writing large streams >2 million items if you are using `nextPut:`. If you use `nextPutAll:` to write the same object, the performance is better as it appears to grow the underlying collection by 1/4 current size. For example, compare these two examples: 
```
Time millisecondsToRun: [
	| stream |
	stream := WriteStream on: String new.
	100000000 timesRepeat: [stream nextPut: $0]]
```
and using `nextPutAll:`:
```
Time millisecondsToRun: [
	| stream |
	stream := WriteStream on: String new.
	100000000 timesRepeat: [stream nextPutAll: '0']]
```
The first one using `nextPut:` takes 6.1 seconds on my machine, but the second one takes 2.6 seconds.

**Request**
Please remove `min: 1000000` from the `WriteStream>>pastEndPut:` method (and update the comment).

**Expected behavior**
With the change above, the running time for the first example above is 1 second on my machine. Furthermore, in large images that take longer for the garbage collector to run, it will be much greater. Here's a graph of the time differences between the current code (in red) and the proposed change in blue. Each data point is the time in microseconds to insert 1 million characters:
image

**Version information:**
 - Pharo Version: All current versions have the same code.

MarcusDenker:
fixes #12128 as described in the issue tracker entry

### DIFF
--- a/src/Collections-Streams/WriteStream.class.st
+++ b/src/Collections-Streams/WriteStream.class.st
@@ -231,10 +231,9 @@ WriteStream >> on: aCollection from: firstIndex to: lastIndex [
 
 { #category : #private }
 WriteStream >> pastEndPut: anObject [
-	"Grow the collection by doubling the size, but keeping the growth between 20 and 1000000.
-	Then put <anObject> at the current write position."
+	"Grow the collection, then put <anObject> at the current write position."
 
-	collection := collection grownBy: ((collection size max: 20) min: 1000000).
+	collection := collection grownBy: (collection size max: 20).
 	writeLimit := collection size.
 	collection at: (position := position + 1) put: anObject.
 	^ anObject


tesonep:
Hi Marcus, I am not sure if duplicating the size of the underlying collection is good in all the scenarios. Maybe we need to add it as a growth rate factor, because it only affects strange scenarios like this one, adding a single element millions of times, maybe the WriteStream should have a way of setting how to behave.

j-brant:
Of course, it may not be "good in all the scenarios". If only one character more is added, then we should have grown by 1 character. However, we can say that growing by a fixed amount is O(n^2), while growing by some factor that is based on the current size is O(n). Of course, the constants in the running time matter. For example, if you grow by a factor of `(size/1000000) ceiling` your running time isn't going to be good. However, the current implementation causes problems with using `contents` to read a file (e.g., `'filename' asFileReference contents`). This uses `nextPut:`. Furthermore, if you do this in a large image (e.g., 1GB or more), the running time gets much worse as GC time starts to dominate the overall time. If you are dealing with large streams like these, most likely your image is going to be large.

As an example, consider the bytes written to create a 40MB stream. Growing by 1MB will have to copy 39 times to get to 40MB: 1MB + 2MB + 3MB + … + 39MB=(39*40/2)MB=780MB copied. If we double in size, we copy 1MB + 2MB + 4MB +… + 32MB = 63MB copied. So the overall number of bytes written for the 1MB growth will be 820MB vs 103MB for doubling (I'm ignoring the first 1MB as my proposed changes have no affect on the first 1MB characters). In addition to the number of additional writes, you also have to pay for the additional GC time to collect the 39 previous arrays vs. the 6 for doubling. The additional GC times makes the running times much worse in large images.

guillep (approved):
We discussed here in the team during the sprint. The change is good!!
There is a potential follow up to configure streams with a growth factor also.

PR #12207, opened 2023-01-03
fix TFProcessLocalWorker with callbacks

estebanlm:
we need some tweaks for using TFProcessLocalWorker with callbacks

### DIFF
--- a/src/UnifiedFFI/LibC.class.st
+++ b/src/UnifiedFFI/LibC.class.st
@@ -69,6 +69,13 @@ LibC >> fgetc: stream [
 	 ^self ffiCall: #(int* fgetc(void* stream))
 ]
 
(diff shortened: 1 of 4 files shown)


src/UnifiedFFI/LibC.class.st
```diff
@@ -69,6 +69,13 @@ LibC >> fgetc: stream [
 	 ^self ffiCall: #(int* fgetc(void* stream))
 ]
 
+{ #category : #'api - accessing' }
+LibC >> getpid [
+	"Get character from stream."
```
gcotelli: Is this comment right? I looks unrelated to the FFI call

MarcusDenker (approved):

PR #12222, opened 2023-01-04
fix Set class comment

pavel-krivanek:
- fix Set comment (nil management)
- prevent no hash method warning (the = message in Set is just optimized for speed)


pavel-krivanek:
The error is:
> IceGitHubError: API rate limit exceeded for 128.93.193.1. (But here's the good news: Authenticated requests get a higher rate limit. Check out the documentation for more details.)

hogoww (approved):
Documentations changes look good !
I'm never using the manifest though, so I don't know.
But Coming from Pavel, I'm never worried ;)

PR #12256, opened 2023-01-05
Refactorings cleanup

balsa-sarenac:
* added `Unused` tag to refactorings that are not exposed
* refactored `InlineMethod` for better readability


Ducasse:
Broken tests are not related.

PR #12258, opened 2023-01-06
convert numpad keys with NumLock turned off

### LINKED ISSUE #12257: PgUp, PgDown, Home and End are ignored on NumPad
**Bug description**

report from Aik-Siong Koh:

> 
> I have an OMEN 17" notebook.
> The keyboard has NumPad with Num Lock.
> Superimposed on the numbers are Home, End, PgUp, etc.
> When Num Lock is on, 1,2,3,4... works everywhere in Windows and apps.
> When Num Lock is off, Home, End, PgUp, etc. work in Windows and every app except in Pharo (9, 10 or 11).
>     In Pharo, pressing the Home key does nothing at all.
> 
> FYI. Above the Numpad are two keys: (HOME / PG UP) and (END / PG DN)
> They work in Pharo with the FN key to get HOME and END.

 I have here a notebook with Numpad where I can reproduce it. Pharo is getting information from SDL2 and there nothing really seems to tell us that the incomming event is PgDown or similar key.

If I log incomming events in OSSDL2BackendWindow>>#visitKeyDownEvent:, I'm getting these values:

```
Num 3
a SDL_KeyDownEventan Array(835757 a SDL_Keysym#(4096 91 1073741915) 768 1 1)
Num PgDown
a SDL_KeyDownEventan Array(860624 a SDL_Keysym#(0 91 1073741915) 768 1 1)
PgDown
a SDL_KeyDownEventan Array(886723 a SDL_Keysym#(0 78 1073741902) 768 1 1)
```

(ignore the first number, the values in SDL_Keysym are mod, scancode, sym)

So the information about the key event only differs in the Num Lock modifier (4096). It is way too low-level. 

It would be nice if someone can try to reproduce it on Linux. I think it may be not Windows specific issue.

I see now the only solution - to the the ugly manual conversion

pavel-krivanek:
- convert numpad keys with NumLock turned off
- set NumLock modifier

fixes #12257

### DIFF
--- a/src/OSWindow-SDL2/OSSDL2BackendWindow.class.st
+++ b/src/OSWindow-SDL2/OSSDL2BackendWindow.class.st
@@ -89,7 +89,7 @@ OSSDL2BackendWindow >> clipboardText: aText [
 
 { #category : #'event handling' }
 OSSDL2BackendWindow >> convertButtonState: mouseState modState: modState modifiers: modifiers [
-	| shift ctrl alt ralt gui |
+	| shift ctrl alt ralt gui numLock|
 	modifiers buttons button1: (mouseState bitAnd: SDL_BUTTON_LMASK) ~= 0.
 	modifiers buttons button2: (mouseState bitAnd: SDL_BUTTON_MMASK) ~= 0.
 	modifiers buttons button3: (mouseState bitAnd: SDL_BUTTON_RMASK) ~= 0.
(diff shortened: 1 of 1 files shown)


pavel-krivanek:
> Why not adding those mappings in the table so they are mapped by `self mapSpecialCharacter: keysym sym` ?

This method does not take into account NumLock status and just converts the character. But we may consider creating some table for it instead of just a row of ifs. Originally, I expected to need to convert only four keys, so the table would be a slower overkill, but now it may make sense. But maybe it may make sense to create an even more general table that would handle the conversion of any key, including various modifiers.

MarcusDenker:
failing test not related:

-  testReentrantCalloutsDuringCallback(#runner->[ TFWorker named: 'fortest' ]) – Windows64.ThreadedFFI.Tests.TFCallbacksTest

Issue #12264, opened 2023-01-07
remove Numeric symbol from source code.


vonbecmann:
**Describe the problem**
#Numeric symbol is the first statement in
Number>>#asScaledDecimal
Float>>#printPaddedWith:to:

**Proposal**
remove it

vonbecmann:
it was merged.

PR #12265, opened 2023-01-07
remove #Numeric from source code.

vonbecmann:
fix:#12264


vonbecmann:
how i can see the test failures?

astares (approved):

PR #12294, opened 2023-01-09
extract return from ifTrue:ifFalse: expression and remove final dot

### LINKED ISSUE #12293: extract return from WeakMessageSend asMinimalRepresentation
extract return from ifTrue:ifFalse: in WeakMessageSend>>#asMinimalRepresentation

Pharo image: Pharo11.0.0#110425
Virtual machine used: CoInterpreter * VMMaker-tonel.1 uuid: aa6234f4-799d-0d00-90d1-c90e092bce84 Jun 10 2022
StackToRegisterMappingCogit * VMMaker-tonel.1 uuid: aa6234f4-799d-0d00-90d1-c90e092bce84 Jun 10 2022
v9.0.15 - Commit: b487900 - Date: 2022-06-02 09:48:57 +0200

Platform Name: unix

Actual Result:
```
asMinimalRepresentation

	self isReceiverOrAnyArgumentGarbage
		ifTrue: [^nil]
		ifFalse:[^self].
```
Expected Result:
```
asMinimalRepresentation

	^ self isReceiverOrAnyArgumentGarbage
		ifTrue: [ nil ]
		ifFalse:[ self ]
```

vonbecmann:
fix: #12293

### DIFF
--- a/src/Kernel/WeakMessageSend.class.st
+++ b/src/Kernel/WeakMessageSend.class.st
@@ -74,9 +74,9 @@ WeakMessageSend >> asMessageSend [
 { #category : #converting }
 WeakMessageSend >> asMinimalRepresentation [
 
-	self isReceiverOrAnyArgumentGarbage
-		ifTrue: [^nil]
-		ifFalse:[^self].
+	^ self isReceiverOrAnyArgumentGarbage
+		  ifTrue: [ nil ]
+		  ifFalse: [ self ]
 ]
 
 { #category : #private }


astares:
If not already we should write a lint rule/critique together with an autofix possibility. I guess we have more cases like this and we could then migrate not only the image but also clean up external packages (Seaside, GLORP, ...) with it as well.

vonbecmann:
> If not already we should write a lint rule/critique together with an autofix possibility. I guess we have more cases like this and we could then migrate not only the image but also clean up external packages (Seaside, GLORP, ...) with it as well.

im searching for this expression
```
`@receiver ifTrue:[^ `@exp] ifFalse:[^ `@exp2]
```
using the expression finder in github://jordanmontt/RewriteToolsSet

MarcusDenker (approved):

PR #12295, opened 2023-01-09
ReSentNotImplementedRule: take Pragma into account

MarcusDenker:
SystemNaviation>>#allUnimplementedCalls did have the feature that one can tag selectors that should be ignored. This was used in just one method

- ReSentNotImplementedRule now takes this pragma into account
- change the bad terminology "allUnimplementedCalls" to be in sync with the rule: #allSentNotImplementedSelectors


MarcusDenker:
Crash on Win, not related

src/System-Support/SystemNavigation.class.st
```diff
@@ -236,19 +236,28 @@ SystemNavigation >> allSentMessages [
...
-SystemNavigation >> allUnimplementedCalls [
+SystemNavigation >> allSentNotImplementedSelectors [
 
 	"Answer all methods where a message is sent but the selector is not implemented anywhere in the system."
 
-	^ self allMethods select: [ :meth | 
+	^ self allMethods select: [ :meth |
 		  | ignored |
-		  ignored := (meth pragmaAt: #ignoreUnimplementedCalls:)
+		  ignored := (meth pragmaAt: #ignoreNotImplementedSelectors:)
```
astares: As we use this checking the pragma part now a second time could we not delegate this question to the method itself and extract it into an own "accessing" method like #allIgnoredNotImplementedSelectors?

```Smalltalk
allIgnoredNotImplementedSelectors
      ^ (self pragmaAt: #ignoreNotImplementedSelectors:)
                        ifNotNil: [ :pragma | pragma argumentAt: 1 ]
			ifNil: [ #(  ) ].
```

astares: this would then make the code more easy to read as you just reject "allIgnoredNotImplementedSelectors"

MarcusDenker: DONE

astares (approved):

astares:
@MarcusDenker 
The only problem remaining is that we have two places now to "ban the lint rule"
1. the method itself (using the pragma)
 
 ```Smalltalk
 foo 	
	<ignoreNotImplementedSelectors: #(zork)>
	self zork 
```
 
2.  the package manifest (when you click the red "X" to ban the rule

image


The PR just covers the first

MarcusDenker:
the whole package manifest implementation is so bad that I just ignore it, I have to admit.

- it stores RingMethod definition store strings, with a useless date
- is not updated when code is refactored
- it is *never* cleaned up. Full of store strings of methods that have been removed years ago

The idea to store data as code like that is just wrong. It's the old "when it is about code, Smalltalkers immediately forget everything they know about objects  and try to map *everything* to classes and methods.

MarcusDenker:
But the "self" rule is another rule... it never makes sense to be ignored as it means that the method either should be moved,  or there is #subclassResponsability missing.

astares:
@MarcusDenker Yes - the package manifest banning is lets say "not optimal". For instance if we would rename classes like "RBxxxx" into "AST" (as it is more about the abstract syntax tree than about the historical refactoring browser) it would not rename in the manifest definitions.

Renreaku needs a review as well. As many other. But step by step.... lets clean and rewrite for the better.

guillep:
A thought, the mechanism to locally disable a rule with a pragma is nice. It's a pity that it's not generalised for all rules :)

PR #12305, opened 2023-01-10
Fix hasSelector:specialSelectorIndex: on FFI methods that are compiled

### LINKED ISSUE #12304: FFI method does not respond correctly to hasSelector:specialSelectorIndex:
**Bug description**
FFI method does not respond correctly to hasSelector:specialSelectorIndex:
This leads to incorrect results of senders of.

**To Reproduce**
We have a test that reproduces it. PR incoming!

tinchodias:
FFI methods must delegate to the original non-compiled method to respond.

Fix #12304


tinchodias:
Pair-programmed with @tesonep

MarcusDenker (approved):

PR #12333, opened 2023-01-14
Variables: Fix Prrinting and Fluid Class definition that uses ClassVariable 

MarcusDenker:
When using Fluid Class definitons, we can define Classs Variables (aka Shared Variables) that are implemented by subclasses of ClassVariable.

(and an example, look at WeakClassVariable)

This PR makes sure that
- the printString is the definition, that is,   #MyVar => ClassVariable
- The class definitons shows #MyVar => ClassVariable, but the short form if the variable is just a ClassVariable

This will make it possible to use self-defined Class Variable kinds and we can start to experiment, e.g. we can 
- have a class var with the value in the definition, see InitializedClassVariable and LazyClassVariable
- we can experiment with more intelligent class vars that work as a Cache and manage the memory they allocate in a more clever way

### DIFF
--- a/src/Kernel/ClassVariable.class.st
+++ b/src/Kernel/ClassVariable.class.st
@@ -35,11 +35,17 @@ ClassVariable >> definingClass: anObject [
 ]
 
 { #category : #printing }
-ClassVariable >> definitionString [
+ClassVariable >> definitionOn: aStream [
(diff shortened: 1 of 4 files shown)


Ducasse:
Sounds good. I still on my todo to do a pass on the printers* because some logic is duplicated but this is simple.

PR #12370, opened 2023-01-15
Refactoring/categorizing unexposed refactorings

balsa-sarenac:
Investigated all refactorings and refactorings2 (transformations) and added `Unused` tag to the ones that are not exposed or not composed.


Ducasse:
Cool! and thanks

Ducasse:
This is strange that none of builds went through.

PR #12373, opened 2023-01-17
Add #normalizeCamelCase and test

hernanmd:
Add method to split a camel cased String into its "normalized" string version, separated by spaces.

```smalltalk
'PinkFloyd' normalizeCamelCase >>> 'Pink Floyd'
```

### DIFF
--- a/src/Collections-Strings-Tests/StringTest.class.st
+++ b/src/Collections-Strings-Tests/StringTest.class.st
@@ -1655,6 +1655,17 @@ StringTest >> testMatch [
 ]
 
 { #category : #tests }
+StringTest >> testNormalizeCamelCase [
+
+	self assert: '' normalizeCamelCase equals: String empty.
+	self assert: 'CamelCase' normalizeCamelCase equals: 'Camel Case'.
+	self assert: 'CamelCase123' normalizeCamelCase equals:'Camel Case123'.
+	self assert: 'ABC' normalizeCamelCase equals: 'ABC'.
+	self assert: 'TresTristesTigres' normalizeCamelCase equals: 'Tres Tristes Tigres'.
+	self assert: 'TresTristesTigresABC' normalizeCamelCase equals: 'Tres Tristes Tigres ABC'
+]
+
+{ #category : #tests }
 StringTest >> testNumArgs [
 	"This is about http://code.google.com/p/pharo/issues/detail?id=237"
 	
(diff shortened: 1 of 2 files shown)


src/Collections-Strings-Tests/StringTest.class.st
```diff
@@ -1654,6 +1654,17 @@ StringTest >> testMatch [
...
 
+{ #category : #tests }
+StringTest >> testNormalizeCamelCase [
+
+	self assert: '' normalizeCamelCase equals: String empty.
+	self assert: 'CamelCase' normalizeCamelCase equals: 'Camel Case'.
+	self assert: 'CamelCase123' normalizeCamelCase equals:'Camel Case123'.
+	self assert: 'ABC' normalizeCamelCase equals: 'ABC'.
+	self assert: 'TresTristesTigres' normalizeCamelCase equals: 'Tres Tristes Tigres'.
+	self assert: 'TresTristesTigresABC' normalizeCamelCase equals: 'Tres Tristes Tigres ABC'
```
astares: please add at least one test with two wording parts as input **separated by space** to cover that case

self assert: 'CamelCase HelloWorld' normalizeCamelCase equals: ...

src/Collections-Strings/String.class.st
```diff
@@ -2178,6 +2178,19 @@ String >> match: text [
 	^ self startingAt: 1 match: text startingAt: 1
 ]
 
+{ #category : #converting }
+String >> normalizeCamelCase [
+	"
+	'TheRollingStones' normalizeCamelCase >>> 'The Rolling Stones'
+	"
+	^ self class streamContents: [ :out |
+		self do: [ :e |
```
astares: why do you call the block parameter "out" and not "s" (like Stream) or better "stream" 
why do you call the block parameter "e" and not "c" (like Character) or better "char" or even better "character"

astares (changes requested):

src/Collections-Strings/String.class.st
hernanmd: Ok, good suggestions, I just integrated them and fixed the case to not write two consecutive blank spaces where there is a single space in the input string.

src/Collections-Strings-Tests/StringTest.class.st
hernanmd: Added

astares (approved):

PR #12387, opened 2023-01-18
Remove ReCompactSourceCodeRule

jordanmontt:
This new rule appears in Calypso all the time. Basically all the methods in the system break it.

Also, each time that you refactor a method, of if you use the formatter, it will add spaces and you will have the rule showing up in Calypso. The formatter breaks the rule also. We don't have plans of changing the formatter in a near future. This rule will complain even if you have an space in the middle of the method. 

I propose to simply delete this rule from Pharo.


astares:
@jordanmontt 
>This new rule appears in Calypso all the time. Basically all the methods in the system break it.

Thats why I already clean them!
Are you aware of all these Cleaning Issues and PRs already:  https://github.com/pharo-project/pharo/issues?q=%22Cleanup%3A+Condense%22

>The formatter breaks the rule also.
OK, but then why dont you delete the formatter in the PR too?

Please do not ignore what others do and just delete the rule, just because you may not like it or be annoyed by it. 

Our hopefully still common goal for and with Pharo since version 1.0 is to improve to get a nice, pretty, clean and 
consistent Pharo language and system and do it better than in the old days of Smalltalk.


I worked on many cleanups:
 -  better categorization, 
  - on having proper #setUp and #tearDown messages, 
  - on having standardized name for Test package, 
  - better dependency and many, many other. 
  
I invested hours in such often boring work since a few years and there was always someone lamenting in the first place. 
Sometimes people say: "why care" - and the answer is "because we want to improve even with small steps".

Compare an old Pharo system with a more recent like P11 and you might spot the difference.
Pharo is more clean and is more consistent through this cleaning and we have introduced release tests to keep the 
quality level and consistency. 

Even when more quality assurance means to also annoy someone as it reminds to be and stay disciplined.

About ReCompactSourceCodeRule introduction I can say:

 1. the rule is there to notify (and also remove clutter and unnecessary stuff) from the source code of a method
 
 2. it helps to clean and condense methods (which is for instance also useful for tiny images)
 
 3. as you see it reveals that even the formatter has some flaws by introducing unnecessary/unwanted separators 
     and spaces into the code once used. Nothing is perfect but it's good that this we have this finding and 
     we then should fix that too. 

What you do in this PR here is not a solution or step forward it is "you just kill the bearer of the bad news"  

And yes - it is not perfect yet (PERFECTION IS NOT A REALITY) but I'm already on it to improve it.

PRIMARY NEXT ACTIONS
- for the standard image I already provide step by step PR's for the P11 package now so this is just a temporay situation (but it takes some time and it would help if people would review instead of lament)

- for you own packages it is a simple click on the red cross and you can ban the lint rule on the package level

- ideally we can change also that the cleaning is applied when you accept/save a method

hernanmd:
Maybe the best is to add setting to enable/disable this rule

astares:
@hernanmd 

Check in Settings Browser under "Code Browsing" -> "Quality Assitant" -> "Toggle rules to run"

image

**There already is the possibility in the settings to disable a rule completely if you do not like or need it.** 
You can do this with ANY lint rule right out of the box.

image

So one can 

 - ban the rule on method level
 - ban the rule on class level
 - ban the rule on package level
 - enable/disable the rule on global level (through settings)

but please do not delete it as it is useful and needed for further cleanup.

astares:
image

jordanmontt:
Hello @astares

> Are you aware of all these Cleaning Issues and PRs already: https://github.com/pharo-project/pharo/issues?q=%22Cleanup%3A+Condense%22

I don't follow that much the Pharo repository so I was not aware of all your PRs and issues.

> Please do not ignore what others do and just delete the rule, just because you may not like it or be annoyed by it.

What do you mean with this? I'm not ignoring other people's work. I didn't even know that was you who created this rule.

> I invested hours in such often boring work since a few years and there was always someone lamenting in the first place.
Sometimes people say: "why care" - and the answer is "because we want to improve even with small steps".

My intention with this PR was not to underestimate your work nor your efforts. I personally appreciate that you put time to improve Pharo with things that other people find boring to do or don't have the time.

> About ReCompactSourceCodeRule introduction I can say: (...)

I understand this. For me, the problem for me is that the formatter itself breaks this rule and that we don't have a clear definition of if it is bad or not to have spaces after each statement. Also, I don't see why it is bad to have a dot in the last statement of a method.

This is my personal opinion and it is not an attack to anyone, but I still think that this rule is not necessary and it is too intrusive.

<blockquote><div> GitHub</div><div><strong><a href="https://github.com/pharo-project/pharo">Issues · pharo-project/pharo</a></strong></div><div>Pharo is a dynamic reflective pure object-oriented language supporting live programming inspired by Smalltalk. - Issues · pharo-project/pharo</div></blockquote>

jordanmontt:
And thank you for showing how to disable rules from the system browser. I will use it to disable several rules, for example the one about the "do not use do: use collect: or select: instead"

astares:
>the problem for me is that the formatter itself breaks this rule

Yes - and as I wrote the solution is then not to remove the lint rule but fix the formatter then even when this can not be done immediately

>Also, I don't see why it is bad to have a dot in the last statement of a method.

If you think you need the you can just use them. 
But there must be a "default" (to have consistency in the image) which is not to have them 
- as they are unnecessary and 
- as they are a kind of "premature dotting" for message that might eventually get added later 

image

Ducasse:
Torsten still there is a problem when we use the code formatter because the code formatter produces extra spaces. 
And this means that users will have problems too. 
Sebastian is not an idiot and he is working on improving Pharo too.

astares:
I recommend **not to merge this PR** as it will remove the lint rule and its autofix completely.
Then
 - this would not improve the situation and move us into the wrong direction
 - it would also mean the cleaning activities we already do for the packages would stop and can not be continued
I also see that people use the lint and autofix already and clean their methods. Step by step.
 
If the formatter is wrongly adding spaces it need to be fixed.

For the user experience:
- I've already shown that the rule could be easily switched off (see screenshots above)
- additionally I work on a better user experience for the critique and cleaning of the final dot (so people who still want them can go with final dots despite the image default) - see #12431 which soon will have also a new PR

@jordanmontt 
But please do close this PR without merging and do not remove the possibility to cleanup

Issue #12188, opened 2023-01-02
Should String>>asNumber be an extension


astares:
Currently String>>asNumber is an extension from "AST-Core-Parser" package.

Do we need to provide it as an extension?

Maybe we should check if it could be a regular method in the package where String is introduced
('Collections-Strings-Base')

Number is defined in Kernel

guillep:
Already solved

PR #12388, opened 2023-01-18
Move #asNumber from AST-Core to Kernel (where it is also already tested)

hernanmd:
String>>asNumber is a core method and shouldn't be in AST-Core as reported by https://github.com/pharo-project/pharo/issues/12188


astares (approved):

MarcusDenker (approved):

MarcusDenker:

PR #12395, opened 2023-01-19
Add Generator class>>onDo:

privat:
Generators in Pharo are nice, and very powerful.
But transforming a callback-based `do:`-like method into a generator is not as straightforward as it could be.

This PR proposes the class sside `onDo:` method to ease this transformation.

E.g. you have a callback based do-like method that values a block on each element.

```st
Integer primesUpTo: 10000000 do: [ :each| each traceCr ]
```

Now, processing the elements as a stream of elements is as easy as:

```st
|generator|
generator := Generator onDo: [ :block | Integer primesUpTo: 10000000 do: block ].
(generator next: 5) asArray >>> #(2 3 5 7 11)
```

For comparison, the classic `on:` approach is:

```st
|generator|
generator := Generator on: [ :gen | Integer primesUpTo: 10000000 do: [ :prime| gen yield: prime]].
(generator next: 5) asArray >>> #(2 3 5 7 11)
```

### DIFF
--- a/src/Collections-Streams-Tests/GeneratorTest.class.st
+++ b/src/Collections-Streams-Tests/GeneratorTest.class.st
@@ -131,6 +131,17 @@ GeneratorTest >> testNext [
 ]
 
 { #category : #tests }
+GeneratorTest >> testOnDo [
+
+	| generator |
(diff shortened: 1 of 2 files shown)


Ducasse:
Yes!

PR #12419, opened 2023-01-23
Introducing a Color Palette to cache the allocation of Color objects by tht UITheme

jordanmontt:
## Description

We were looking at the object allocations using method proxies and we saw that Pharo allocates between 300 and 700 Color objects per second. We also saw that in 20 seconds, out of 12000 created Color objects, only 42 were unique, leaving 11958 unnecessarily created objects.

For this, we propose to add the concept of a Color Palette to the UITheme to cache the creation of Color objects. 

## Implementation

UIThemeColorPalette is a class that uses a Dictionary to cache the colors once they were created. It is used by UITheme like this:

```st
UITheme >> #backgroundColor
	^ colorPalette backgroundColor
```

```st
UIThemeColorPalette >> #backgroundColor
	^ paletteDictionary at: #backgroundColor ifAbsentPut: [ colorConfigurator backgroundColor ]
```

The classes ColorConfigurator, there is one subclass for each of the themes dark and light, are the ones that know what color to return. It is used by the UIThemeColorPalette

```st
DarkThemeColorConfigurator >> #backgroundColor

    ^ Color r: 0.130 g: 0.130 b: 0.130
```

Also moved the creation of ThemeSettings from the class side of UITheme to the instance side.

_Note:_
As the responsibility of creating the colors is dispersed in different classes and these classes are heavily used in the system; we only did a patch that fixes only a part of the problem. A complete solution is needed as a redesign of the ui theme to address the encountered problems with the excessive allocation of objects. The Flyweight design pattern also can be also used.


Ducasse:
cool!

astares:
@jordanmontt / @Ducasse 

Nice!

For the concept of a "theme" object and customization I guess over time the number of extension methods on 
the theme will increase. 

I wonder if it would make sense to group them into an own class (like a kind of "subtheme").

 ```Smalltalk
 UISubtheme subclass: #GrowlLightTheme
	instanceVariableNames: ''
	classVariableNames: ''
	package: 'Polymorph-Widgets-Themes'
```
because then you do not use the method on the standard theme but ask the (standard) theme for the "growl" theme first.

image

So we divide the theme into subthemes and make them less complex and easier to follow/navigate.
What do you think?

**Comment by jordanmontt** [comment]
Hello @astares ,

Thank you for your interest in this PR. I have updated it with a Description and a clearer title. If you are interested, we found this problem with the color allocations using this memory allocation profiler that we are starting to build https://github.com/jordanmontt/illimani-memory-profiler. The development is in the prototype stage. If you have any suggestions please feel free to let me know.

As for your suggestion, I like your idea and it seems good. But I need to discuss it with my supervisors if I can/should do it as the purpose of this PR is part of the investigation that we are doing about the memory allocations :)

<blockquote><div> GitHub</div><div><strong><a href="https://github.com/jordanmontt/illimani-memory-profiler">GitHub - jordanmontt/illimani-memory-profiler</a></strong></div><div>Contribute to jordanmontt/illimani-memory-profiler development by creating an account on GitHub.</div></blockquote>

**Comment by jordanmontt** [comment]
The PR is now ready. Please review :)

**Comment by MarcusDenker** [comment]
"the following classes have uncategorized methods: an OrderedCollection(UIThemeColorConfigurator UIThemePalette)"

**Comment by jordanmontt** [comment]
Thanks Marcus, I made a commit fixing this

**Comment by jordanmontt** [comment]
Warning: this will break any custom themes that people may have. I am thinking of https://github.com/Ducasse/Cozy/ for example. Has some custom blue themes that will probably not work properly once this PR will be merged


<blockquote><div> GitHub</div><div><strong><a href="https://github.com/Ducasse/Cozy">GitHub - Ducasse/Cozy: A little package to create a more cozy working environment</a></strong></div><div>A little package to create a more cozy working environment - GitHub - Ducasse/Cozy: A little package to create a more cozy working environment</div></blockquote>

**Comment by astares** [comment]
Just as an info: I remember @estebanlm has some themes for Pharo too

File: src/Polymorph-Widgets/UIThemePalette.class.st [hunk 1]
```diff
@@ -0,0 +1,659 @@
...
+		'paletteDictionary'
+	],
+	#category : #'Polymorph-Widgets-Themes'
+}
+
+{ #category : #colors }
+UIThemePalette >> backgroundColor [
+
+	^ paletteDictionary at: #backgroundColor ifAbsentPut: [ colorConfigurator backgroundColor ]
+]
```
PalumboN, line 18: These methods make me think if we don't want (or it's already implemented maybe) some mechanism to cache the result of a method.

I'm thinking something like
```st
UIThemePalette >> backgroundColor [
	<cached>

	^ colorConfigurator backgroundColor
]
```
And the magic happens (the first time the method is executed, the sentence is evaluated and the returned object is stored, then it returns the stored result for the next invocations).

PalumboN: And how is it invalidated? 🤔

PalumboN (approved):
Coollll 🎉 

We can open an issue with the remaining work that you say in the description, maybe someone in the community is interested ;)

src/Polymorph-Widgets/UIThemePalette.class.st
guillep: there is already `[] memoised`. But, how is it invalidated ? :)

src/Polymorph-Widgets/PharoDarkTheme.class.st
```diff
@@ -14,265 +14,133 @@ Class {
...
+
+	(aButton valueOfProperty: #noBorder ifAbsent: [ false ]) ifTrue: [ 
+		^ SimpleBorderStyle new
+			  width: 0;
+			  baseColor: Color transparent ].
 	^ SimpleBorderStyle new
-		width: 1;
-		baseColor: self buttonColor lighter
+		  width: 1;
+		  baseColor: self buttonColor lighter
```
guillep: There is still a color allocation here!

src/Polymorph-Widgets/PharoDarkTheme.class.st
```diff
@@ -14,265 +14,133 @@ Class {
...
-{ #category : #'accessing - colors' }
-PharoDarkTheme >> enabledColor [ 
-	^ self textColor
-]
-
-{ #category : #'basic-colors' }
-PharoDarkTheme >> errorPopperBackgroundColor [
-	^ Color red muchDarker
+		  width: 1;
+		  baseColor: self buttonColor lighter
```
guillep: Color allocation here!

src/Polymorph-Widgets/PharoDarkTheme.class.st
```diff
@@ -285,271 +153,157 @@ PharoDarkTheme >> initialize [
...
-	^ self textColor
-]
+	"Answer the color to use for a label when disabled."
 
-{ #category : #'accessing - colors' }
-PharoDarkTheme >> menuBorderColor [
-	^ self borderColor
+	^ aLabel owner ifNil: [ aLabel paneColor muchDarker ] ifNotNil: [ 
+		  aLabel owner color isTransparent ifTrue: [ aLabel owner paneColor muchLighter ]
+			  ifFalse: [ aLabel owner color muchLighter ] ]
```
guillep: Color allocationS here!

src/Polymorph-Widgets/ThemeSettings.class.st
```diff
@@ -347,9 +347,9 @@ ThemeSettings >> maxBalloonHelpLineLength: anInteger [
 ThemeSettings >> menuBorderColor [
 	"Answer the value of menuColor"
 
-	^ self flatMenu  
-		ifFalse: [#raised]
-		ifTrue: [self derivedMenuColor twiceDarker]
+	^ self flatMenu 
+		ifFalse: [ #raised ]
+		ifTrue: [ self derivedMenuColor twiceDarker ]
```
guillep, line 352: Color allocation here!

guillep (changes requested):
I agree that the iterations can be added on top.

I'd like however to see class comments explaining the new architecture.
There are so many methods in these classes just doing color xx color yy, that's it's difficult to follow:
 - when are objects instantiated
 -  how instantiates them
 - are they cached?
 - where? and when?

Also, it would be nice that this change comes with a small migration guide of how to migrate old themes to the new schema.

src/Polymorph-Widgets/UIThemePalette.class.st
jordanmontt: When you change the theme, for example from light to dark, you create a new palette object with a new color configurator. That is how is invalidated. Nevertheless, there is no an api to invalidate a color without changing the theme, this is no different from current's implementation

jordanmontt: I like the <cached> pragme that Nahuel proposed. But for this is a separated project. Maybe a Phep?

src/Polymorph-Widgets/PharoDarkTheme.class.st
jordanmontt: For this one, I don't know how to cache it. Do you think is a good idea to have 3 methods. For example: `disabledLabelWhenOwnerIsNil` `disabledLabel` `disabledLabelWhenTransparent` ?

jordanmontt:
I will add better class comments explaining the architecture and a migration guide for creating new themes in a new commit

Ducasse:
I think that all these lighter and others darker are really a bad practice.
So this is good to remove them.

MarcusDenker:
there are now conflicts

Ducasse:
Of course when we have full system changes then it impacts pending changing.

jordanmontt:
I will close this PR and I will open a new one.

Issue #12412, opened 2023-01-23
"Browse this version in Calypso" Error after clicking an unloaded package


hernanmd:
**Bug description**

A NotFound exception is raised from the DAMessageSendAnalyzer after clicking a package in the Calypso browser on an unloaded package. This was found when trying to reproduce https://github.com/pharo-project/pharo/issues/9684

**To Reproduce**

![calypso_external_browse](https://user-images.githubusercontent.com/4825959/214012710-9e359498-b0fa-4003-a911-f905acdbdb4c.gif)

**Version information:**
 - OS: macOS
 - Pharo Version 11

**Expected development cost**
We need to check if Calypso should calculate dependencies for unloaded packages.

MarcusDenker:
PR merged

Issue #9684, opened 2021-07-23
When browsing an external package via iceberg, system loops.


Ducasse:
This loops forever when invoked from the external class browser 

```
RBBehavior >> inheritsFrom: aClass 
    "Answer whether the argument, aClass, is on the receiver's superclass 
    chain."

    | aSuperclass |
    aSuperclass := self superclass.
    [aSuperclass == nil]
        whileFalse: 
            [aSuperclass == aClass ifTrue: [^true].
            aSuperclass := aSuperclass superclass].
    ^false
```

Ducasse:
How to reproduce it.
- open iceberg
- select a package not loaded in the system
- bring Browse in Calypso
- click on the package
- click on any class ...
I do not know what is unresolved but it is not nil so we get in an endless loop.

Ducasse:
```
shouldBeActivatedInContext: aBrowserContext
	^ RSUMLCalypsoSettings useCalypsoUML 
		and: [ aBrowserContext isClassSelected 
			and: [ aBrowserContext lastSelectedClass inheritsFrom: BaselineOf ] ].	
```

Ducasse:
```
aSuperclass := self superclass.
```
returns an unresolved and then boum....

PR #12423, opened 2023-01-24
Fix for "Browse this version in Calypso" Error after clicking an unloaded package

hernanmd:
A NotFound exception is raised from the DAMessageSendAnalyzer after clicking a package in the Calypso browser on an unloaded package. This was found when trying to reproduce #9684.

See https://github.com/pharo-project/pharo/issues/12412 for video to reproduce. This commit preserves #rPackage behavior by answering a virtual RPackage so Calypso does not break trying to analyze dependencies of an unloaded package.


MarcusDenker:
this is this, but on linux:

MarcusDenker (approved):

PR #12425, opened 2023-01-24
Fix Calypso browsing unloaded package artifacts from Iceberg

hernanmd:
Add a check in RGBehavior>>inheritsFrom: so unresolved superclasses gets out of an infinite recursion. See https://github.com/pharo-project/pharo/issues/9684 for details.
Implements #definitionString hooks to display the class definition String in the Calypso browser for non-loaded packages, using MCSnapshot information.
Add missing #needsSlotClassDefinition and #allowsShadowing in RGBehavior, which were causing red morph of death while browsing some methods.
Add #buildMicrodownUsing:withComment in RGBehavior, needed for Calypso when rendering the Comment tab.


Ducasse:
THANKS!!!

MarcusDenker:
known random failing tests on win:
-  [CI] System variable environment on windows does not use adequate encodings #11407

src/Ring-Core/RGBehavior.class.st
```diff
@@ -222,6 +222,16 @@ RGBehavior >> asYetUnclassifiedProtocolName [
 	^ Protocol unclassified
 ]
 
+{ #category : #accessing }
+RGBehavior >> buildMicroDownUsing: aMicMicrodownTextualBuilder withComment: aString [
```
guillep: I wonder if this is ok here or this should be an extension method.

src/Ring-Core/RGBehavior.class.st
```diff
@@ -486,17 +503,16 @@ RGBehavior >> includesSelector: aString [
...
 	| aSuperclass |
 	aSuperclass := self superclass.
-	[aSuperclass == nil]
-		whileFalse: 
-			[aSuperclass == aClass ifTrue: [^true].
-			aSuperclass := aSuperclass superclass].
-	^false
+	[ aSuperclass isNil ] whileFalse: [
+		aSuperclass == aClass ifTrue: [ ^ true ].
+		aSuperclass hasResolvedName ifFalse: [ ^ false ].
```
guillep, line 504: Yes, good catch. Could we have a test for this?

guillep: I mean, the case of iterating the hierarchy when a superclass is not resolved. That's an important ring case.

src/Ring-Core/RGPackage.class.st
```diff
@@ -43,6 +43,16 @@ RGPackage >> addoptToParentStub [
 	^ self parent addPackage: self.
 ]
 
+{ #category : #accessing }
+RGPackage >> buildMicroDownUsing: aMicMicrodownTextualBuilder withComment: aString [
```
guillep: Same as before. Should this be an extension method?

src/Tool-DependencyAnalyser-Tests/DAMessageSendAnalyzerTest.class.st
```diff
@@ -13,6 +13,16 @@ DAMessageSendAnalyzerTest >> setUp [
...
 ]
 
+{ #category : #tests }
+DAMessageSendAnalyzerTest >> testRPackage [
+
+	self assert: (analyzer rPackage isKindOf: RPackage).
+	analyzer := DAMessageSendAnalyzer on: 'Unloaded-Dummy-Package'.
+	self
+		assert: (analyzer rPackage isKindOf: RPackage)
+		description: 'It test that rPackage still answer a RPackage even when the analyzer was instantiated with an unexisting or unloaded package'
```
guillep: I'm not fond of tests doing isKindOf, but ok.

This kind of tests prevent future extensibility. Anyways, if required the test can be removed in the future.

guillep: A stronger test should be to send a message to the `analyzer rPackage` asserting expected classes or methods in the package

```smalltalk
self assert: analyzer rPackage definedClasses equals: ...
```

That asserts that the returned object works correctly, catches the bug too, and leaves the door open to extensions :)

src/Tool-DependencyAnalyser/DAMessageSendAnalyzer.class.st
```diff
@@ -130,7 +130,11 @@ DAMessageSendAnalyzer >> possibleDeadCode [
 
 { #category : #accessing }
 DAMessageSendAnalyzer >> rPackage [
-	^ packageAnalysis rPackageSet packages detect: [ :each | each packageName = self packageName ]
+	"Answer a <RPackage> matching the receiver's package name. If we are browsing an unloaded package, then answer a virtual RPackage so the message send analyzer could preserve its behavior with loaded packages in #manuallyResolvedDependencies"
```
guillep: good catch!

This is the kind of thing that a type inferencer should be able to catch super easily...

guillep (changes requested):

src/Ring-Core/RGBehavior.class.st
hernanmd: Ok, added as extension methods in the Microdown repository.

src/Ring-Core/RGPackage.class.st
hernanmd: Done

src/Tool-DependencyAnalyser-Tests/DAMessageSendAnalyzerTest.class.st
guillep: I don't understand, the test is gone, is that on purpose?

src/Tool-DependencyAnalyser/DAMessageSendAnalyzer.class.st
guillep: This fix is not in the PR anymre?

src/Tool-DependencyAnalyser-Tests/DAMessageSendAnalyzerTest.class.st
hernanmd: I re-added the missing methods. 
Added as testRPackageIsAnsweredWithLoadedPackage

src/Tool-DependencyAnalyser/DAMessageSendAnalyzer.class.st
hernanmd: I don't know what happened, but I retrieved the changes and re-commited again

MarcusDenker (approved):

PR #12429, opened 2023-01-26
add a "parse plugin" mechanism to OpalCompiler

estebanlm:
this is o be used to transform the ast in the parse phase (otherwise plugins are used in compile phase).
we will use this mechanism to add custom highlighter.


src/OpalCompiler-Core/CompilationContext.class.st
```diff
@@ -466,7 +481,8 @@ CompilationContext >> getClass [
 CompilationContext >> initialize [
 	options := Set new.
 	astTransformPlugins := OrderedCollection withAll: self class defaultTransformationPlugins.
-	semanticScope := OCMethodSemanticScope targetingClass: nil class
+	semanticScope := OCMethodSemanticScope targetingClass: nil class.
+	astParseTransformPlugins := OrderedCollection new: 2
```
privat: introduce `defaultParseTransformationPlugins` to be consistent with the other plugins?

privat (approved):
Did you consider renaming the other plugin? E.g. from `astTransformPlugins` to `astCompileTransformPlugins` etc.

src/OpalCompiler-Core/CompilationContext.class.st
guillep: Is it really needed to allocate it with 2?

src/AST-Core/RBMessageNode.class.st
```diff
@@ -474,6 +474,14 @@ RBMessageNode >> sentMessages [
 		yourself
 ]
 
+{ #category : #accessing }
+RBMessageNode >> start: newStart [
+	| oldStart |
+
+	oldStart := self start.
+	keywordsPositions := keywordsPositions collect: [ :each | each - oldStart + newStart ]
```
guillep: This could use a test

src/OpalCompiler-Core/OpalCompiler.class.st
```diff
@@ -199,12 +205,22 @@ OpalCompiler >> bindings: aDictionary [
...
+
+	plugins := compilationContext astParseTransformPlugins ifEmpty: [ ^ self ].
+	plugins sort: [ :a :b | a priority > b priority ]. "priority 0 is sorted last"
+	plugins do: [ :each | ast := each transform: ast ]
+]
+
 { #category : #private }
 OpalCompiler >> callPlugins [
-	| plugins  |
+	| plugins |
```
guillep, line 219: Now that we have some other plugins maybe we need to rename this? And keep of course the following for backwards compatibility

```smalltalk
callPlugins
   self callCompileTimePlugins
```

src/Shout/Behavior.extension.st
```diff
@@ -0,0 +1,10 @@
+Extension { #name : #Behavior }
+
+{ #category : #'*Shout' }
+Behavior >> syntaxHighlighter [
```
guillep: I don't think this should be in the PR, right?

Unless the PR is about pluggable highlighting, not about parse plugin mechanism.

src/Shout/SHRBTextStyler.class.st
```diff
@@ -14,7 +14,8 @@ Class {
 		'font',
 		'isForWorkspace',
 		'pixelHeight',
-		'workspace'
+		'workspace',
+		'plugins'
```
guillep: Same as above for all changes in this file.
I don't think this should be in the PR, right?

Unless the PR is about pluggable highlighting, not about parse plugin mechanism.

guillep (changes requested):

src/OpalCompiler-Core/CompilationContext.class.st
estebanlm: ok

estebanlm: it is a size optimization: most of the time there will not be parse plugin so allocate 10 slots for it seemed a lot, but anyway, that's already addressed.

src/Shout/Behavior.extension.st
estebanlm: is about both. One is the proof/first use of the other.

estebanlm:
> Did you consider renaming the other plugin? E.g. from `astTransformPlugins` to `astCompileTransformPlugins` etc.

not really, this is out of my scope :)
I will let @MarcusDenker to lead that :P

src/Shout/Behavior.extension.st
guillep: Then changing the PR title was enough.

I don't understand why the commit title makes me happy, personally, it does not make me happy.

estebanlm:
so, can this be taken care, please?
@MarcusDenker ?

src/OpalCompiler-Core/OpalCompiler.class.st
MarcusDenker: yes, we can improve naming im a next step.

MarcusDenker (approved):

PR #12443, opened 2023-01-27
Moving stepping down to Morph

### LINKED ISSUE #12398: Review RubSmalltalkEditor>>#tallySelection
I wonder why RubSmalltalkEditor>>#tallySelection is calling #lineSelectAndEmptyCheck: twice in the same method

```Smalltalk
tallySelection
	"Treat the current selection as an expression; evaluate and tally it."

	self lineSelectAndEmptyCheck: [ ^ self ].
	self lineSelectAndEmptyCheck: [ ^ self ].
	self tally: self selectionAsStream
```

guillep:
Fixes #12398

- Moving stepping down to Morph
- SystemWindow forces all models (so nil, collections) to step, so deal with that locally for now

### DIFF
--- a/src/Kernel/Object.class.st
+++ b/src/Kernel/Object.class.st
@@ -1978,30 +1978,6 @@ Object >> species [
 	^self class
 ]
 
-{ #category : #stepping }
-Object >> stepAt: millisecondClockValue in: aWindow [
-
-	^ self stepIn: aWindow
-]
-
-{ #category : #stepping }
-Object >> stepIn: aWindow [
-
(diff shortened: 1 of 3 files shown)


Ducasse:
argh I integrated fixes of cyril that are conflicting. Sorry about it.

MarcusDenker:
Cyril's PR seems to fix the same issue: https://github.com/pharo-project/pharo/issues/12398

jecisc:
A fix was integrated and it is working

Issue #12441, opened 2023-01-27
[CI] testExternalMorphicCoreDependencies is failing


MarcusDenker:
**Bug description**
testExternalMorphicCoreDependencies is failing

```
Graphic-Files is missing in knownMorphicCoreDependencies
```

No idea why it started failing

PR #12446, opened 2023-01-27
Fix #12287 adding group model to parent window

### LINKED ISSUE #12287: Protection of unsaved content does not work if the window being closed has multiple tabs
**Bug description**
Protection of unsaved content does not work if the window being closed has multiple tabs.

**To Reproduce**
1. browse implementors of `testBasics` (or any other harmless method), it has three implementors
2. modify the source code in the first showed implementor in any way but do not save it
3. select the method name `testBasics` and press `Ctrl+M` or alternative to show the implementors of the method, a new tab with the implementors is shown
4. close the whole window
5. the window is closed without warning about the unsaved content

**Where protection works**
1. browse implementors of `testBasics` (or any other harmless method), it has three implementors
2. modify the source code in the first showed implementor in any way but do not save it
3. go to the second implementor
4. close the whole window
5. `Changes have not been saved` question dialog is opened

ValentinBourcier:
Fixing issue: https://github.com/pharo-project/pharo/issues/12287

### DIFF
--- a/src/Calypso-Browser/ClyBrowserMorph.class.st
+++ b/src/Calypso-Browser/ClyBrowserMorph.class.st
@@ -284,7 +284,8 @@ ClyBrowserMorph >> createWindowGroupFrom: currentWindow [
 	group addWindow: currentWindow.
 	(group openInWindowLabeled: currentWindow label)
  		extent: ext;
-		position: pos.
+		position: pos;
+		model: group.
 	^group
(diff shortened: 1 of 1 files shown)


MarcusDenker:
- CI] testExternalMorphicCoreDependencies is failing #12441

guillep (approved):
Good!

Issue #12431, opened 2023-01-26
Better user experience for final dot critique and cleaning 


astares:
By default we would like to remove unnecessary separators from method body to istay clean.

See 
    https://twitter.com/TorstenAstares/status/1609666189382254593
    https://twitter.com/TorstenAstares/status/1609952076804628482
and #12387

This includes unnecessary and unwanted spaces, tabs but also final dots at the end of the method.

Despite the cleaning and the default in the Pharo standard image some users might nonetheless
want to continue with them for their own packages.  They also might not want to be criticized for 
their the final dots.

So we should increase the user experience by providing a setting where they can disable this
default in the settings browser like:

![image](https://user-images.githubusercontent.com/5980033/214922724-260fce85-3139-4f5d-a580-2d50970084c6.png)

MarcusDenker:
one idea could be to split the rule into two: whitespace and dot.

For whitespace, I want to experiment then with an automation: accepting code should do the whitespace remove automatically.

jordanmontt:
I like the idea of splitting the rule into two rules: white spaces and dots :)

astares:
No need to split but just configure - as this is possible with the same code. Already working on it. Please stay patient.

And yes - cleaning on accept would also be nice. But step by step.

MarcusDenker:
What confuses me is that I alway get a message about a dot, but there is no dot. Yes, it says "or", but my brain can not hanle that.

For me a split would really improve the situation.

astares:
It's not finished yet ...

astares:
PR is in #12455

PR #12448, opened 2023-01-27
Fix conflict between formatter and rules: spaces before new lines

PalumboN:
After discussion in https://github.com/pharo-project/pharo/pull/12387 between @jordanmontt and @astares we change the formatter to not enter in collision with the rule.

Beyond the rules that any of us want in his code, I think that is important that the tools inside the environment won't conflict between them (ie. the formatter should not break other systems like linter, compiler, highlighter, etc ). 
Should be nice to maintain this equilibrium, if you change something in the system keep in mind that should be in harmony with the rest of the components.

> PR made in a Pharo Sprint with @hernanmd

### DIFF
--- a/src/EnlumineurFormatter-Tests/EFBlockExpressionTest.class.st
+++ b/src/EnlumineurFormatter-Tests/EFBlockExpressionTest.class.st
@@ -298,8 +298,7 @@ EFBlockExpressionTest >> numberOfSpaceChangedAfterIndentCharacterConfiguration [
 		newLineAfterCascade:true;
 		newLineBeforeFirstCascade: true;
 		numberOfSpacesInIndent: 1;
-		indentStyle: #space;
-		numberOfSpacesInIndent: 2;
+		indentStyle: #tabulation;
 		indentExtraSpaces:0
 ]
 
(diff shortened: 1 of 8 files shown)


Ducasse:
This is cool. 
As a default for EF I would like to have no extra line between the selector and the first comment - it is on my todo since ages.

astares:
Nice as the formatter is the right place to take action.

Will try to check it but I'm busy with
https://github.com/pharo-project/pharo/issues/12431
which should also help to improve the situation. Willl work on that first.

guillep:
@astares I'd like if you can make a review when you have time :)

astares:
Will do!

src/EnlumineurFormatter-Tests/EFBlockExpressionTest.class.st
```diff
@@ -124,6 +124,18 @@ EFBlockExpressionTest >> dontFormatCommentWithStatementConfiguration [
 		indentExtraSpaces:0
 ]
 
+{ #category : #configurations }
+EFBlockExpressionTest >> example: col [
```
jordanmontt: What is this? It looks like some leftover

jordanmontt (approved):

jordanmontt:
Clever solution, each time that there is a new line the spaces at the end of the statement are removed

astares:
looks good from cases I tried to format - but I havent checked very deeply (not yet had the time to check code)

PalumboN:
Thanks for the review!

@Ducasse I think that you can configure that with this setting:
image

Ducasse:
Yes I did it but I was stupid not to put it on by default :)

astares (approved):

guillep:
Thanks!

PR #12455, opened 2023-01-28
12431 Better user experience for final dot critique and cleaning

### LINKED ISSUE #12431: Better user experience for final dot critique and cleaning

astares:
Fix #12431

### DIFF
--- a/src/Renraku-Tests/ReMethodSourceCleanerTest.class.st
+++ b/src/Renraku-Tests/ReMethodSourceCleanerTest.class.st
@@ -7,8 +7,117 @@ Class {
 	#category : #'Renraku-Tests-Utilities'
 }
 
-{ #category : #tests }
+{ #category : #'private - test data' }
+ReMethodSourceCleanerTest >> aleadyCleanedMethodStringKeepingDot [
+	^'dirtyMethodWithUnnecessarySpacesAndFinalDot
+	"Do not format this method - it by intention should include unnecessary spaces and a final dot"
+
+	|temp|
+
+	temp := 3 + 4.'
+]
+
+{ #category : #'private - test data' }
+ReMethodSourceCleanerTest >> alreadyCleanedMethodString [
+	^'dirtyMethodWithUnnecessarySpacesAndFinalDot
+	"Do not format this method - it by intention should include unnecessary spaces and a final dot"
+
+	|temp|
+
+	temp := 3 + 4'
+]
+
+{ #category : #'private - test data' }
+ReMethodSourceCleanerTest >> dirtyMethodWithUnnecessarySpacesAndFinalDot [
+	"Do not format this method - it by intention should include unnecessary spaces and a final dot"
+
+	|temp|
+
+	temp := 3 + 4
+]
+
+{ #category : #'private - test data' }
+ReMethodSourceCleanerTest >> properMethodWithFinalDot [
+	"This is a method without any unnecessary separators in the method body but a final dot"
+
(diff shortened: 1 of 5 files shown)


src/Renraku-Tests/ReMethodSourceCleanerTest.class.st
```diff
@@ -7,8 +7,117 @@ Class {
...
+
+{ #category : #'private - test data' }
+ReMethodSourceCleanerTest >> alreadyCleanedMethodString [
+	^'dirtyMethodWithUnnecessarySpacesAndFinalDot
+	"Do not format this method - it by intention should include unnecessary spaces and a final dot"
+
+	|temp|
+
+	temp := 3 + 4'
+]
```
jordanmontt, line 32: It looks like you have this method duplicated

src/Renraku-Tests/ReMethodSourceCleanerTest.class.st
```diff
@@ -7,8 +7,117 @@ Class {
...
+	|temp|
+
+	temp := 3 + 4
+]
+
+{ #category : #'private - test data' }
+ReMethodSourceCleanerTest >> properMethodWithFinalDot [
+	"This is a method without any unnecessary separators in the method body but a final dot"
+
+	^ 3 + 4.
```
jordanmontt, line 47: Maybe add a warning in the comment to prevent people to format this method

jordanmontt (approved):

src/Renraku-Tests/ReMethodSourceCleanerTest.class.st
astares: No - as you see and as the selector should reveal: one returns the cleand method with a dot and one without. It's also not an own method body - it returns a string we compare agains the #dirtyMethodWithUnnecessarySpacesAndFinalDot in the tests

src/Renraku-Tests/ReMethodSourceCleanerTest.class.st
astares: Made it more clear in https://github.com/astares/pharo/commit/4904ac8610c2f07e369be13a25efb174b07400ba

src/Renraku-Tests/ReMethodSourceCleanerTest.class.st
astares: But I added comment in https://github.com/astares/pharo/commit/4904ac8610c2f07e369be13a25efb174b07400ba
too

jordanmontt: Didn't see the dot at the end :)

astares: That's why ideally there should be none :)

astares:
CI failing tests unrelated

guillep (approved):

guillep:
Thanks!

PR #12468, opened 2023-01-30
Fix DrTest UI when plugin selection change

privat:
A recent optimization (66dfc5bd4e967f09492d4cf1ce413aac11f287f5) disabled the update of plugin-related presentation (eg the package list) when a plugin is selected in the dropbox.

This made "executable comments" untestable with DrTest :(


Ducasse:
Broken tests are unrelated.

PR #12475, opened 2023-01-30
Disable setting formatIncompleteIdentifiers by default

privat:
Syntax highlighting is very slow when coloring garbage source code (e.g. a bad coy-paste in an editor window).

```st
[ StPlayground openContents: (String loremIpsum: 10000) ] timeToRun "14 seconds"
```

The culprit is `SHRBTextStyler>>#formatIncompleteSelector:` that calls `Symbol class>>#selectorThatStartsCaseSensitive:skipping:` that is insanely inefficient.

This method is used to distinguish bad selector that are prefix of existing selectors from just bad selectors.
Personally, I'm not sure to understand why that matters in the first place, especially since the default theme just color both of them red, so the default user could not see any difference anyway.

Luckily, this formatting is protected by a setting (introduced by 46360f7b7f5f031701a68d1eda2474a6983007d1). This PR just made the default to false. (My first idea was to nuke the whole thing).

Now, the default user can highlight garbage 4275% more efficiently!

```st
[ StPlayground openContents: (String loremIpsum: 10000) ] timeToRun "0.32 seconds"
```

### DIFF
--- a/src/Shout/SHRBTextStyler.class.st
+++ b/src/Shout/SHRBTextStyler.class.st
@@ -340,7 +340,7 @@ SHRBTextStyler class >> defaultStyleTable [
 { #category : #styles }
 SHRBTextStyler class >> formatIncompleteIdentifiers [
 
-	^ formatIncompleteIdentifiers ifNil: [ formatIncompleteIdentifiers := true ]
+	^ formatIncompleteIdentifiers ifNil: [ formatIncompleteIdentifiers := false ]
 ]
 
 { #category : #styles }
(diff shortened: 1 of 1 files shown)


guillep (approved):
Yes

MarcusDenker:
besides speed, this has the problem that it shows those selectors not colored red when browsing code... the highligher does not distinguish between "someone is typing" and "just show this full method". The "incomplete" color only makes sense while typing...

privat:
Distinguishing incomplete selectors while editing could make sense, but only for the selector currently edited, not the whole source code.

I don't think a Text object has the knowledge of a cursor or something, therefore the cursor information should be given to the highlighter, or responsibility to deal with incomplete selectors moved to something else.

If the feature is that much unloved, maybe just remove it completely instead of playing with default values. This will remove virtually dead code and remove a setting. It could be brought back later with a better design.

jordanmontt (approved):

PR #12479, opened 2023-01-30
NumberParser accept and ignore underscore

privat:
The proposal is to extends number syntax to accept and ignore the underscore character `_` (ASCII 95).

Many languages (including Python https://peps.python.org/pep-0515/ , Java https://docs.oracle.com/javase/7/docs/technotes/guides/language/underscores-literals.html or Ruby) accept some forms of numeric literal that ignore `_`.

The idea is to permit long literals that are still readable, eg. `1_000_000_000` is easier for a human than `100000000` especially since in the previous literal a zero is missing (I'm a tricky deceitful fellow).

Usually, there is not a lot of constraint on the placement of `_` to accommodate various locales and usages or radix.

However, duplicate or training `_` are usually rejected as syntax error.
The PR just ignore any `_` without constrains. We might add them, or add them as rules for case that feel more being a syntax warning than a syntax error.

### DIFF
--- a/src/NumberParser-Tests/NumberParserTest.class.st
+++ b/src/NumberParser-Tests/NumberParserTest.class.st
@@ -314,6 +314,20 @@ NumberParserTest >> testFloatmin [
 ]
 
 { #category : #'tests - Integer' }
+NumberParserTest >> testIntegerParseUnderscore [
+
+	self assert: (NumberParser parse: '1234') equals: 1234.
+	self assert: (NumberParser parse: '1_234') equals: 1234.
+	self assert: (NumberParser parse: '__1__2__3__4__') equals: 1234.
+	self should: [NumberParser parse: '__'] raise: Error.
+
+	self assert: (NumberParser parse: '16r1234_5678') equals: 16r12345678.
+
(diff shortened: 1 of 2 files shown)


svenvc:
I am not so sure it is a good idea to do this by default.
In NeoCSV I have my own number parser that supports this, as an option:
```
NeoNumberParser>>#digitGroupSeparator: separatorCharacter
	"Set the digit group separator to separatorCharacter.
	The are skipped while parsing digit characters.
	The default is nil (nothing being skipped)"
	
	digitGroupSeparator := separatorCharacter 
```
If you don't make it an option, you can't do strict checking anymore.

guillep:
I'd like to have this in Pharo. I think that this can have a lot of value when working with numbers.
And I think it would be nice to to have also a number printer that adds separators...

I think this should become a phep.

estebanlm:
phep is here: https://github.com/pharo-project/pheps/pull/18

please let's discuss there :)

guillep:
Closed in favor of https://github.com/pharo-project/pharo/pull/17760

PR #12494, opened 2023-01-31
fix: DeprecateClassCommand DNU

### LINKED ISSUE #9551: Fix and enable SycDeprecateClassCommand
I want to deprecate a class and first I'm forced to give an existing class.
Second if I give a class name I get DNU.

S

balsa-sarenac:
Fixes DNU and exposes `RBDeprecateClassRefactoring`.

Closes: #9551


MarcusDenker:
Yes, we should turn in on again.

I would like to simplify it even more: now we have to select a class that is the replacement, but not every deprecated class is just a rename.

To me a DeprecateClassCommand should just add the method to the class side, and do nothing else.

Maybe we should have a class rename refactoring in addition that does the rename of users, add a new class, refactor users...

MarcusDenker:
But this PR could be a good first step

MarcusDenker (approved):

MarcusDenker:
Failing test not related:

- [CI] failing test: RBReplaceSubtreeTransformationTest>>testTransformOneOfManyStatements #12508

PR #12499, opened 2023-01-31
Stop `->` insertions in editor contents

privat:
When compiling (ctrl+s), compilation errors alters the content with `-> Error messages` that are often unwelcome.

This PR proposes a new default behavior: just a growl notification to inform that the compilation/installation was not performed. The rest of the Rubric editor (highlighting and icons) should already give all the information on compilation errors, no need to also hijack the source code.

Previous `->` behavior can still be configured in the settings.


guillep:
I think the idea of removing this is great.

I'll take some minutes in the afternoon to review the PR in detail :)

privat:
Initially, I just wanted to remove the feature entirely, but it seems that the code was almost ready to accept alternative behavior (overengineering or leftover code?) so I just used this existing facility. In fact, most of the PR code is about the system setting.

astares:
"just a growl notification to inform" - this includes some danger as people might overlook and think the code is good while
the code pane is still "dirty".

Changing the source code is not nice (and maybe also annoying) but my understanding is that current mechanism starts to show where the error in parsing happens. 

So for instance if a method would be wrong like this

image
 
it will inset the cursor on the right place:
image
so you can correct it by typing (or deleting that part).

Maybe we check how regular code editors (VisualStudio, Eclipse and others) give a feedback and align Pharo on that.

privat:
> this includes some danger as people might overlook and think the code is good while the code pane is still "dirty".

Maybe, but it's already the case with the insertion mechanism (especially when `->` are added in invisible parts of the text). So, not really a new problem here.

Nevertheless:

* I added a flash on the editor window to give feedback that something is happening.
* I also move the cursor in front of the error to show where the error happens. It's the same behavior with the insertion mechanism, but without the insertion.

Note: for some reason, new selection do not scroll the editor text to ensure it is visible. It's already the case with the insertion mechanism, so not a new bug, but it feels weird.
I tried to find a way to make the selection visible, but failed. If someone has a hint, I will take it.

> Maybe we check how regular code editors (VisualStudio, Eclipse and others) give a feedback and align Pharo on that.

Yes we can. But it could be done in future PR. The present PR is about stopping messing with the contents. Future PR will require more refactorization and UI design.

MarcusDenker:
If you look at the Playground, there we get the floating error message instead of destruction of the code.

I wonder if we could not do the same by default in all browsers. The problem of course is that this is implemented in the workspace by hooking in the exception of the parser that comes after accept.

For the browser, this is not enough as we need to show it not only on accept, but when we look at the code.

privat:
You are right about the playground, but it seems to be a toolkit issue, and it's far for my knowledge level. :(

Rubric (the code editor widget) and Calypso (the code browser) that embed it seem to be pure Morphic. Whereas the Playground seems a Spec2 that embed a Rubric adapter (`StPlaygroundInteractionModel` points to a `SpCodePresenter` that points to a `SpMorphicCodeAdapter` that points to a `SpRubScolledTextMorph` that points to instances of classes from the Rubric package).

The popup widget seems fully provided by Spec2 and used by the workspace, not a Rubric thing.

Looking at the call chain, in case of compilation error in the playground, `SpCodeInteractionModel>>#notify:at:in:` is called and instantiates a `SpCodePopoverErrorPresenter` (from Spec2-Code) that inherits `SpPopoverContentPresenter` (from Spec2-Core).

I did not find an equivalent text popup in the Morphic world, I don't know how hackish it could be to embed a Spec2 thing in pure Morphic code.

I can accept pointers to help me to understand all those things. 

Nevertheless, I'm currently not sure what are the next steps here. :/

privat:
@MarcusDenker:

> For the browser, this is not enough as we need to show it not only on accept, but when we look at the code.

I'm not sure to understand what you mean.

MarcusDenker:
I think the playground implements it by just having an on:do: handler for the compiler exceptions, which works as it is shows only on explicit compile (doit, print-it).

The code browser (when just looking at a method) compiles for highlighting, but that uses the mode where we parse in "faulty" mode, thus no exception is raised, the whole mechanism needs to work differently.

privat:
To extends what I wrote above, the playground just hooks itself as the requestor of the compiler, so that `OpalCompiler>>#compile` directly calls `SpCodeInteractionModel>>#notify:at:in:` in case of a syntax error that deal with the UI directly (for instance, it removes the `->` in the string before creating the popup).

It's the same path for the Calypso/Rubric case, except that the requestor is not the same, therefore`RubScrolledTextMorph>>#notify:at:in:` is called instead. It’s this code path that is modified in the present PR so that the default behavior of (plain) Rubric is to stop altering the text contents.

It could be nice to unify the UI (some tools use the spec thing and some other use the rubric thing) but it could be done in other PR (multiple PR).
Obviously, if Rubric is rewritten in Spec2 or just linked with Spec2 popup, this is an improvement over the current PR.

Note for above: there is a lot of code-area-text-input-widget in a lot of tools that expect code input from the human (senders/callers list, debugger, profiler, inspector), the full list of these tools could help the future UI design choices.
There are other tools that just display code in a read-only way (like the finder). But humans can still select and doit. The reporting behavior in case of syntax error is variable, but basically uses the Spec2 popup or the pure Rubric way (`->` insertion before the PR or system popup + flash with the PR). The only tool I tried that use something else is the old profiler that causes a DNU on syntax error (need more investigation, or not if the tool will be deprecated).

Furthermore, as @MarcusDenker wrote, it also could be nice to clean up the interactions* between the compiler and its various clients, but it seems to me that it is out of the scope of the current PR.

Obviously, if `notify:at:in:` is replaced by something more elegant, the current PR is to be adapted.

*there are some weird design choices and the rationale (or history) is often missing (or difficult to find). Addressing all of them will be a lot of work, and a lot of discussion.

guillep:
Do we have a decision on this? :) I like the idea, I'm not sure about the repercussions

privat:
I dunno. Maybe I should rewrite as something more conservative. Possibly only targetting the edition in calypso because it already offers a better error feedback than plain rubric, and it will limit the scope or unwanted repercussions.

privat:
New proposal at  #13073

PR #12501, opened 2023-02-01
Profiler: simplify world menu

MarcusDenker:
This PR removes the old Time Profiler from the world menu.

With just one entry, we can change the name of the Profiler now to be just "Profiler"


astares (approved):

astares:
How heretic - some books now need to be rewritten and adjusted with new screenshots 😉

guillep (approved):

PR #12504, opened 2023-02-01
Introduce new string comparison primitive 

jecisc:
But do not use it because there is a bug in the VM side still


guillep (approved):
Good!

PR #12512, opened 2023-02-02
Extract set up now classify setUp in #running

### LINKED ISSUE #11729: RBExtractSetUpMethodRefactoring should put #setUp into "running" protocol
There is a RBExtractSetUpMethodRefactoring that allows to extract an assignment in a test method within a SUnit test class into a #setUp method.

Unfortunately the #setUp is then in the same method protocol as the testing method (which gives a lint error) as the #setUp and #tearDown methods are usually in "running" protocol.

So we should extend this refactoring so it puts the final #setUp method into "running" protocol already automatically so the flow is not interrupted.

Example:

Create an SUnit test class

```Smalltalk
TestCase subclass: #FooTest
	instanceVariableNames: '' 
	classVariableNames: ''
	package: 'Foo'
```

now create a method protocol "tests" and write a testing method like this:
 

```
testCollection

	| coll |
	coll := OrderedCollection new.
	self assert: coll isEmpty
```

then select the variable definition and assignment and choose "Extract setup" from the context menu

image

the refactoring will convert the temporary variable into an instance variable and create a #setUp method automatically:

image

but once you "APPLY" the refactoring you will notice that the #setUp method is generated in the same method protocol as our testing method (in "tests"). 

But this gives a lint error as #setUp should for best practices and consistency reason be in "running" method protocol:

image

For sure one can use the "auto repair" of the lint issue to recategorize but it would be better if the **RBExtractSetUpMethodRefactoring** would put the resulting #setUp method already into the "running" protocol

jecisc:
+ factorization of code

Fixes #11729

### DIFF
--- a/src/Refactoring-Core/RBAbstractClass.class.st
+++ b/src/Refactoring-Core/RBAbstractClass.class.st
@@ -208,13 +208,13 @@ RBAbstractClass >> compile: aString [
 ]
 
 { #category : #'method accessing' }
-RBAbstractClass >> compile: aString classified: aSymbolCollection [ 
+RBAbstractClass >> compile: aString classified: aSymbolCollection [
 	| change method |
-	change := model 
+	change := model
 				compile: aString
 				in: self
 				classified: aSymbolCollection.
-	method := modelFactory rbMethod 
+	method := modelFactory rbMethod
 				for: self
 				source: aString
 				selector: change selector.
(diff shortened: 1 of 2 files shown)


MarcusDenker (approved):

astares:
@jecisc / @MarcusDenker 

Nice - thanks!

PR #12519, opened 2023-02-03
Cache inital hash of strings

### LINKED ISSUE #11580: Consider caching or hard coding String initialHash
**Describe the request**
`String >> #hash` uses `ByteString hash` as an initial hash. This should be cached or hard coded.

**Expected behavior**
`String >> #hash` sends `ByteString hash` to compute an initial hash. This sometimes shows up during profiling. `ByteString hash` has be a constant value for an image otherwise all collections would have to be rehashed. `Behavior` implements `#basicIdentityHash` and uses primitive 175 to return the classes index in the class table.  This returns `13312` in my image.

It looks like the idea was once to use `self class hash` or similar so that `ByteString` and `WideString` would hash differently but this was abandoned due to the need for rehashing.

Instead of sending `ByteString hash` every time we hash a string we can either cache this value or use hard code value.

**Expected development cost**
> As you know nothing happens magically, so do you have an idea of the development effort required? 

Depends a bit on the approach chosen. Simply hard coding a value in `String >> #hash` should not be too much effort. Caching depends a bit on where and how to cache, the easiest would probably be a class variable.

> In addition, are you willing to help because after all Pharo is yours too?

Sure. If there is consensus what should be done how.

**Version information:**
 - OS: any
 - Version: any
 - Pharo Version: at least 10 and 11

jecisc:
Fixes #11580

Warning: Since String is a core class, I had to do the changes by hand in the files directly and not in Pharo. I hope I did not do any syntax error :)


jecisc:
Adding a class variable to String seems to not be that easy. The bootstrap failed. 

Does anyone have an idea on how to update the string class to work with the bootstrap?

jecisc:
Trying to reopen now that we can update Espell once again

jecisc:
Closing since Guille will propose a simpler implementation by hardcoding the value :)

Issue #12221, opened 2023-01-04
[CI] Random failing: TKTNewProcessTaskRunnerTest>>#testNewProcessTaskRunnerRunsInNewProcessEveryTime


MarcusDenker:
**Bug description**
TKTNewProcessTaskRunnerTest>>#testNewProcessTaskRunnerRunsInNewProcessEveryTime fails sometimes randomly

```
Error
Got 9 instead of 10.
Stacktrace
TestFailure
Got 9 instead of 10.
TKTNewProcessTaskRunnerTest(TestAsserter)>>assert:description:resumable:
TKTNewProcessTaskRunnerTest(TestAsserter)>>assert:description:
TKTNewProcessTaskRunnerTest(TestAsserter)>>assert:equals:
TKTNewProcessTaskRunnerTest>>testNewProcessTaskRunnerRunsInNewProcessEveryTime
TKTNewProcessTaskRunnerTest(TestCase)>>performTest

```

MarcusDenker:
merged a skip

PR #12520, opened 2023-02-03
Rename PlatformIndependentEnvironment into NullOSEnvironment

### LINKED ISSUE #11375: PlatformIndependentEnvironment does not convey its role
PlatformIndependentEnvironment is a confusing name.
Is it a language environment or a shell like?
Then PlatformIndependent does not convey the fact that it looks like it is a Null one.
I do not really understand because if this is an OS environment what is the at:ifAbsent: method doing.
 Is it used to get variable such $HOME value?

S

jecisc:
Fixes #11375


MarcusDenker:

astares (approved):

PR #12526, opened 2023-02-03
Move deprecated classes to deprecated package and remove their users.

jecisc:
Partly fixed 10840 but some deprecated classes are still in Spec


jecisc:
There are some weird changes. I'll redo a PR

PR #12527, opened 2023-02-03
Move deprecated classes to deprecated package and remove their users.

jecisc:
Partly fixed 10840 but some deprecated classes are still in Spec and Roassal


astares:
I'm not sure but I guess the Deprecated11 package need to appear in Baseline too for being recognized and loaded during bootstrap.

Maybe just look up where "Deprecated10" in Pharo 10, "Deprecated9" package in Pharo 9 was mentioned in source code by scanning through the system using Finder.

jecisc:
It’s already loaded, I just need to replace the usage of deprecated classes in spec.
I did it in my image but I forgot to commit in spec

MarcusDenker:
I am not sure if the whole idea of the Deprecated package is still really good. 

Another thing is that deprated classes: the users do not know. There is no output in the logs, there is no warning. 
As such, we have to be much more careful with removing them.

jecisc:
This PR requires https://github.com/pharo-spec/Spec/pull/1351 before integration

PR #12528, opened 2023-02-03
Refactoring UI cleanup

balsa-sarenac:
Small changes to `refactoring-ui`. This is the first step towards refactoring of refactoring presenters by changing their prefix to `St` and moving them to `Refactoring-UI`


Ducasse:
We should pay attention because the new package should be declared in the baseline somewhere.

Then we should check the following error

```
Error

Found undeclared Variables:  RBDeprecateMethodDriver in: SycDeprecateMessageCommand>>#execute RBMoveMethodsToClassSideDriver in: SycMoveMethodsToClassSideCommand>>#execute
Stacktrace

TestFailure
Found undeclared Variables: 
RBDeprecateMethodDriver in: SycDeprecateMessageCommand>>#execute
RBMoveMethodsToClassSideDriver in: SycMoveMethodsToClassSideCommand>>#execute
ReleaseTest>>testUndeclared
ReleaseTest(TestCase)>>performTest

```

balsa-sarenac:
I've added it to `BaselineOfBasicTools` since that is where the `Refactoring-Core` and `Refactoring-Changes` were. Do you think there is a better place to put it in?

src/Refactoring-UI/StRefactoringPreviewPresenter.class.st
```diff
@@ -165,35 +165,35 @@ RBRefactoringPreviewPresenter >> initializePresenters [
 ]
```
jordanmontt, line 158: It is not necessary to call initialize the Layout by yourself. Now you should use `defaultLayout` method in which you need to return the desire layout, so the same layout that is built in `initializeLayout`

src/Refactoring-UI/StRefactoringPreviewPresenter.class.st
```diff
@@ -137,7 +137,7 @@ RBRefactoringPreviewPresenter >> initializeLayout [
...
```
jordanmontt, line 131: also instead of sending the message `padding:` to each presenter you can set up the spacing of the layout with `aLayout spacing: ` like that you don't need to set the padding each time

src/Refactoring-UI/StRefactoringPreviewPresenter.class.st
```diff
@@ -90,7 +90,7 @@ RBRefactoringPreviewPresenter >> columns [
...
 
 { #category : #initialization }
-RBRefactoringPreviewPresenter >> connectPresenters [
+StRefactoringPreviewPresenter >> connectPresenters [
 
 	super connectPresenters.
```
jordanmontt: why calling super connectPresenters?

jordanmontt (changes requested):

src/Refactoring-UI/StRefactoringPreviewPresenter.class.st
balsa-sarenac: Hey, thanks for the review! This is my first Spec window, so I was looking in multiple places to figure out how to create it and naturally I've copied some stuff that I shouldn't have. It makes much more sens to use `defaultLayout`, I've changed it.

src/Refactoring-UI/StRefactoringPreviewPresenter.class.st
balsa-sarenac: That's really handy, thanks for the tip!

src/Refactoring-UI/StRefactoringPreviewPresenter.class.st
balsa-sarenac: I've removed it now, it makes no sense to have it here. Good catch!

src/Refactoring-UI/StRefactoringPreviewPresenter.class.st
jordanmontt: No problem, it is my pleasure! :)
The ìnitializeLayout`was a solution that we used to do but then the `defaultLayout` method was added! The chang is really recent.

src/Refactoring-UI/StRefactoringPreviewPresenter.class.st
```diff
@@ -114,30 +113,25 @@ RBRefactoringPreviewPresenter >> initializeLayout [
...
-			 add: (SpBoxLayout newTopToBottom
-					  add: scopesLayout
-					  withConstraints:
-						  [ :constraints | constraints height: 30 ] yourself;
-					  add: table;
-					  add: diffPresenter;
-					  yourself);
-			 add: actionsLayout expand: false;
-			 yourself)
+	^ self layout: (SpBoxLayout newTopToBottom
```
jordanmontt: There is only one thing left to do, and that is to return the layout without assignment in.
So, you only need to remove the `self layout: (...)` something like this:

```
defaultLayout
    ^ (SpBoxLayout newTopToBottom
        add: (SpBoxLayout newTopToBottom
        ...
```

jordanmontt (changes requested):

src/Refactoring-UI/StRefactoringPreviewPresenter.class.st
balsa-sarenac: That makes sense, I've changed it as well. Thanks!

src/Refactoring-UI/StRefactoringPreviewPresenter.class.st
balsa-sarenac: Oh, that's great, and the new way of defining layout is much better.

jordanmontt (approved):

PR #12529, opened 2023-02-04
12518-ImageCleaner-shareLiterals-shouldnt-consider-Strings-and-Symbols-the-same 

### LINKED ISSUE #12518: ImageCleaner shareLiterals shouldn't consider Strings and Symbols the same
**Bug description**
Symbols can be replaced to Strings by ImageCleaner shareLiterals
If in a method you have for example 'something' and #something , then run ImageCleaner shareLiterals, we get a String instead of the #something symbol at runtime

**Additional context**
Related to https://github.com/pharo-project/pharo/issues/12420

MarcusDenker:
use #identityIndexOf: instead of indexOf:, fixes #12518


MarcusDenker:

PR #12531, opened 2023-02-04
Cleanup unused refactorings

balsa-sarenac:
Removed some duplicated ones, renamed ones that are transformations and other minor cleanups.


Ducasse:
Thanks balsa it was a supercool pair programming session

PR #10830, opened 2022-01-31
Remove Author

guillep:
Lots of cleanups
 - remove the Author class (and with it the useless popup)
 - remove dependency from many users

Issue #12575, opened 2023-02-06
OmFileStoreReadingError Illegal leading byte after DeprecationPerformedNotification


MarcusDenker:
**Bug description**

we see this now often on the CI:

DeprecationPerformedNotification: Automatic deprecation code rewrite: The method Float class>>#epsilon called from RSTransformable>>#updateCurrentShape: has been deprecated. Use machineEpsilon instead

OmFileStoreReadingError: Reading entry from File @ /Users/lse/Workspace/workspace/est_and_branch_Pipeline_PR-12541/Tests-osx-64/pharo-local/ombu-sessions/Pharo11.0-PR-64bit-3f8475b.218301751A20.ombu@11403266: Illegal leading byte for utf-8 encoding

svenvc:
It could be that this is something that we have seen before, maybe it was not fully fixed.

I think this is related to a position in a binary stream that is being moved (probably backward) without regards for utf8 character boundaries, like going back 100 bytes. Next a higher level character decoding stream is used to read. Most of the time this works (ASCII), but sometimes you are unlucky and end up in the middle of a character. That is a likely cause for the above error.

Here is an example:
```
| bytes |
bytes := 'Francais' utf8Encoded.
(ZnCharacterReadStream on: (bytes readStream setToEnd; skip: -4)) next.
```
This gives you $c as you would expect going 4 positions backwards.
However:
```
| bytes |
bytes := 'Français' utf8Encoded.
(ZnCharacterReadStream on: (bytes readStream setToEnd; skip: -4)) next.
```
gives you the error as $ç is encoded using 2 bytes and the stream is positioned between them.

The method ZnUTF8Encoder>>#backOnStream: can do this correctly, but it is not further integrated into streams. It does not have to be used on each step, doing it only as the last step seems OK:
```
| bytes |
bytes := 'Français' utf8Encoded readStream.
bytes readStream setToEnd; skip: -4.
ZnCharacterEncoder utf8 backOnStream: bytes.
(ZnCharacterReadStream on: bytes) next.
```
Which gives $ç

jecisc:
I checked the code of Ombu and it does not seems to skip back. But it does set the position in the stream explicitly in some places. 

(The problem seems to be in one of the senders of #readEntriesWith:)

svenvc:
OK, as far as I can see, there are 3 senders that manipulate the position:

- OmFileStore>>#nextEntryFromPosition:
[...]

PR #12584, opened 2023-02-07
Remove dependency of ChangeSet on Author

jecisc:
There is a plan to remove the need of Author from the image. This is a little step toward that goal by removing its use in ChangeSet.  Change extracted from:â¯https://github.com/pharo-project/pharo/pull/10830


MarcusDenker:
- [CI] OmFileStoreReadingError Illegal leading byte after DeprecationPerformedNotification #12575

PR #12586, opened 2023-02-07
Remove code that is not loaded from the git repository

jecisc:


jecisc:
I removed the packages present in the repository but not loaded in the image

jecisc:
This is weird because the bootstrap fails because it seems to rely of Zinc-FileSystem. But Zinc-FileSystem is not loaded in Pharo anymore. 

I don’t know the bootstrap process really well so I’ll need to investigate why it works like that

astares:
Would we keep the old Deprecated packages at least so people can load them to migrate their code?

PR #12587, opened 2023-02-08
Update vm parameter 45 description

### LINKED ISSUE #5715: Update vm parameter 45 description
Currently, the vm parameter 44 and 45 descriptions are:

```
#44	size of eden, in bytes
#45	desired size of eden, in bytes (stored in image file header)
```

This was true before Spur, but it is not the case anymore. 

Before new space and eden were the same thing. 

Now the new space is composed of the eden (5/7 of the new space) and two survivor spaces (1/7 of the new space each). 

Parameter 44 is the size of the eden in bytes while parameter 45 should be the desired size of the **new space== in bytes. 

For more info:
https://github.com/OpenSmalltalk/opensmalltalk-vm/issues/476

jecisc:
Fixes #5715

### DIFF
--- a/src/System-Support/SystemNavigation.class.st
+++ b/src/System-Support/SystemNavigation.class.st
@@ -236,14 +236,6 @@ SystemNavigation >> allSentMessages [
 ]
 
 { #category : #query }
-SystemNavigation >> allSymbolsStartingWith: aString do: aBlock [
-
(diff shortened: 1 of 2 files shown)


jecisc:
The failing test is a Network error

jecisc:
This one should be easy to review for someone knowing the VM

PR #12591, opened 2023-02-08
Do not resolve absolute paths to open a file

### LINKED ISSUE #590: PharoLauncher cannot determine versions of Pharo 11 images
The current private VM for spur images is too old to be able to execute the primitive to get the current directory of Pharo 11 images. 
Because of that, the command to determine the version of an image is failing. 

A possible solution would be to change the behavior of this command to just return in the studout the version of the image via an eval command line. This would minimize the code to execute in the Pharo image to run. And then let the launcher create the file.

jecisc:
Currently to open a file we always resolve the path relatively to the… working directory

This commit changes that to resolve only in case the path is not absolute. In case of absolute path we can just use the path as it is. This could help use resolve https://github.com/pharo-project/pharo-launcher/issues/590

### DIFF
--- a/src/FileSystem-Core/FileSystem.class.st
+++ b/src/FileSystem-Core/FileSystem.class.st
@@ -668,10 +668,10 @@ FileSystem >> open: aResolvable writable: aBoolean [
 	that path using the specified access mode."
 
 	| path |
-	path := self resolve: aResolvable.
-	^ store handleClass
-		open: (FileReference fileSystem: self path: path)
-		writable: aBoolean
+	path := aResolvable isAbsolute
+		        ifTrue: [ aResolvable ]
+		        ifFalse: [ self resolve: aResolvable ].
+	^ store handleClass open: (FileReference fileSystem: self path: path) writable: aBoolean
 ]
 
 { #category : #private }


MarcusDenker:
failing test not related

MarcusDenker (approved):

src/FileSystem-Core/FileSystem.class.st
```diff
@@ -668,10 +668,10 @@ FileSystem >> open: aResolvable writable: aBoolean [
...
 
 	| path |
-	path := self resolve: aResolvable.
-	^ store handleClass
-		open: (FileReference fileSystem: self path: path)
-		writable: aBoolean
+	path := aResolvable isAbsolute
+		        ifTrue: [ aResolvable ]
+		        ifFalse: [ self resolve: aResolvable ].
+	^ store handleClass open: (FileReference fileSystem: self path: path) writable: aBoolean
```
dionisiydk, line 674: It would be nice to have a test covering this logic.
Is it a bug fix or just an optimization? If former then the test is really needed.

tesonep (changes requested):
I agree that we need a test. 
I don't like the isAbsolute question, we can delegate to the absolute path to do nothing.
There is already the extension point, we can implement the AbsolutePath >> #asResolvedBy: as just returning self.
Also, we can remove the awful if in Path >> #resolvePath:

jecisc:
I tried to do that but then my image bugs in some places because here we resolve a path with a path, but I think that in other places we do it with other resolvable and we don’t expect a path as result.

PR #12604, opened 2023-02-09
Compiler: enable optionInlineTimesRepeat and optionInlineRepeat by default

### LINKED ISSUE #11169: Compiler: It supports compiling timesRepeat: as optimised code: Check if we can enable
The compiler has support to compile  timesRepeat: and #repeat to jumps (like to:do:), but they are disabled by default.

See #optionInlineTimesRepeat and #optionInlineRepeat

We should think about enabling them by default.

MarcusDenker:
enable the following by default:

	(+ optionInlineTimesRepeat 		'Inline timesRepeat: if specific conditions are met (See isInlineTimesRepeat)')
	(+ optionInlineRepeat 			'Inline repeat if specific conditions are met (See isInlineRepeat)')
				
fixes #11169

### DIFF
--- a/src/Kernel-Tests/CompiledBlockTest.class.st
+++ b/src/Kernel-Tests/CompiledBlockTest.class.st
@@ -13,9 +13,9 @@ CompiledBlockTest >> testLiteralEqual [
 
 	| value |
 	value := 0.
-	10 timesRepeat: [self assert: value equals: 0].
+	(1 to: 10) do: [self assert: value equals: 0].
 	value := 1.
-	10 timesRepeat: [self assert: value equals: 0].'.
+	(1 to: 10) do: [self assert: value equals: 0].'.
 
 	compiledBlocks := methodToTest literals select: [ :each | each isCompiledBlock ].
 	self assert: compiledBlocks size equals: 2.
@@ -30,9 +30,9 @@ CompiledBlockTest >> testPcInOuter [
 
 	| value |
 	value := 0.
-	10 timesRepeat: [self assert: value equals: 0].
+	(1 to: 10) do: [self assert: value equals: 0].
 	value := 1.
-	10 timesRepeat: [self assert: value equals: 0].'.
+	(1 to: 10) do: [self assert: value equals: 0].'.
 
 	compiledBlocks := methodToTest literals select: [ :each | each isCompiledBlock ].
 	self assert: compiledBlocks size equals: 2.
--- a/src/Kernel-Tests/CompiledBlockTest.class.st
+++ b/src/Kernel-Tests/CompiledBlockTest.class.st
@@ -13,6 +13,7 @@ CompiledBlockTest >> testLiteralEqual [
 
 	| value |
 	value := 0.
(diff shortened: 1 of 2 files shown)


MarcusDenker:
test to fix:

- CompiledBlockTest>>testLiteralEqual
- CompiledBlockTest>>testPcInOuter

src/Kernel-Tests/CompiledBlockTest.class.st
```diff
@@ -13,9 +13,9 @@ CompiledBlockTest >> testLiteralEqual [
 
 	| value |
 	value := 0.
-	10 timesRepeat: [self assert: value equals: 0].
+	(1 to: 10) do: [self assert: value equals: 0].
 	value := 1.
-	10 timesRepeat: [self assert: value equals: 0].'.
+	(1 to: 10) do: [self assert: value equals: 0].'.
```
demarey, line 19: Why is `timesRepeat:` replaced by an `interval do:`(that could also have been a `to:do:`)?

MarcusDenker: the test needs a block. the PR changes timesRepeat: to not compile the argument as a block.

As the code is not executed, but is executable in principle I wanted a minimal change...

MarcusDenker: to:do: is compiled optimzid, too, so it would have the same problem as timesRepeat:

MarcusDenker:
- [CI] OmFileStoreReadingError Illegal leading byte after DeprecationPerformedNotification #12575

src/Kernel-Tests/CompiledBlockTest.class.st
demarey: ok, thanks

demarey (approved):

src/Kernel-Tests/CompiledBlockTest.class.st
```diff
@@ -13,9 +13,9 @@ CompiledBlockTest >> testLiteralEqual [
 
 	| value |
 	value := 0.
-	10 timesRepeat: [self assert: value equals: 0].
+	(1 to: 10) do: [self assert: value equals: 0].
```
jecisc, line 17:
```suggestion
	"Do not use #to:do: because it is optimized and will prevent us of testing the behavior we want."
	(1 to: 10) do: [self assert: value equals: 0].
```

src/Kernel-Tests/CompiledBlockTest.class.st
```diff
@@ -30,9 +30,9 @@ CompiledBlockTest >> testPcInOuter [
...
```
jecisc, line 35:
```suggestion
	"Do not use #to:do: because it is optimized and will prevent us of testing the behavior we want."
	(1 to: 10) do: [self assert: value equals: 0].
```

jecisc (commented):
I propose to add comments to avoid future mistakes

PalumboN (approved):

PR #12611, opened 2023-02-10
Update matchesTypes to fix DNU in matchesTypes

### LINKED ISSUE #12610: ExternalDropHandler, "ZnMimeType DNU #beginsWith"
I noticed this bug when trying out Drag-and-Drop from the OS. The underlying ZnMimeType has a different signature than needed, so that the method `matchesTypes;` throws the Error mentioned.
Direct fix which seems to work is the following:

Original:
```smalltalk
matchesTypes: types
        (self type isNil or: [types isNil])
            ifTrue: [^false].
        ^types anySatisfy: [:mimeType | mimeType beginsWith: self type]
```
Fix:
```smalltalk
matchesTypes: types
	(self type isNil or: [types isNil])
		ifTrue: [^false].
	^types anySatisfy: [:mimeType | mimeType main beginsWith: self type]
```

I submitted a Pull request with that change

MatFluor:
Fixes: #12610 

ZnMimeType has accessors which are not used in the `matchesTypes:` method. This fixes it, but it is untested if it has repercussions.

### DIFF
--- a/src/Morphic-Core/ExternalDropHandler.class.st
+++ b/src/Morphic-Core/ExternalDropHandler.class.st
@@ -135,7 +135,7 @@ ExternalDropHandler >> matchesExtension: aExtension [
 ExternalDropHandler >> matchesTypes: types [
 	(self type isNil or: [types isNil])
 		ifTrue: [^false].
-	^types anySatisfy: [:mimeType | mimeType beginsWith: self type]
+	^types anySatisfy: [:mimeType | mimeType main beginsWith: self type]
 ]
 
 { #category : #accessing }


jecisc:
Can you tell me what kind of files you are dropping and on which OS? 
Because I cannot reproduce the bug for now

MatFluor:
Here's a way to reproduce it:

```smalltalk
csvHandler := ExternalDropHandler new type: 'text' extension: 'csv' action: [ UIManager default inform: 'CSV file added' ].
ExternalDropHandler registerHandler: csvHandler.
```
And then grab a CSV-File from the OS and drop it into the World.

Tested with the following combinations:
P10 and P11 (latest) on Win 11
P10 and P11 (latest) on Ubuntu-derivate (ZorinOS)

guillep (approved):
I tested it, ok!

guillep:
Thanks!

PR #12613, opened 2023-02-10
Remove help packages that are not loaded in the image

jecisc:
Subpart of https://github.com/pharo-project/pharo/pull/12586 that should not impact the bootstrap


guillep (approved):
Yes! Thanks!

PR #12615, opened 2023-02-10
Fix float comparision with precision and start to test Math-Operation…

### LINKED ISSUE #11153: Strange behavior of Float >> closeTo:precision:
I find this strange:
```st
3 closeTo: 3.001 precision: 0.001. "true"
0 closeTo: 0.001 precision: 0.001. "false"
```
Why does this happen?
Because there is a guard clause for 0 in the implementation
```st
Float >> closeTo: num precision: aPrecision
 	"Tell whether the receiver and arguments are close from each other given a precision"
	
	num isNumber ifFalse: [ ^ [self = num] onErrorDo: [false]].
	self = 0.0 ifTrue: [^ num abs < aPrecision].
	num = 0 ifTrue: [^self abs < aPrecision].
	^self = num asFloat
		or: [(self - num) abs / (self abs max: num abs) < aPrecision]
```
But then I am confused, why do we divide by `(self abs max: num abs)` ?? 🧐

jecisc:
…s-Extensions

Fixes #11153


jecisc:
Fixes this:

image

jecisc:
Failures seems related

jecisc:
CI failure not related. Win slave seems out

guillep (approved):
Good!

PR #12764, opened 2023-02-13
Unnecessary characters rule should be an information and not a warning

### LINKED ISSUE #12614: Unnecessary characters rule checks spaces inside a String
The rule unnecessary characters found in method body checks for characters inside a String. For example see photo below:

Capture d’écran 2023-02-09 à 21 59 02

It detects one space inside a String. For me this should not be the case since there is no code formatting inside the content of a String

astares:
Fix #12614


jordanmontt (changes requested):
Agree for this PR. But for me it does not fix the linked issue :)

astares:
See my comment in #12614

jordanmontt (approved):

PR #12768, opened 2023-02-13
11368-Failing-on-CI-random-CoNarrowHistoryFetcherTesttestNarrowingReturnsSameElementsThatCallingDirectly

### LINKED ISSUE #11368: Failing on CI (random): CoNarrowHistoryFetcherTest>>testNarrowingReturnsSameElementsThatCallingDirectly

MarcusDenker:
I propose to run the tests on a Set, this way it works like the test was originally written.

This is of course a workaround to get the tests on the CI green (we could alternatively skip it on the CI)

That it (sometimes) does not work correctly for Array is very strange, we should open an issue for that then.

fixes #11368


jecisc (approved):

jecisc:
CI failure unrelated

PR #12771, opened 2023-02-13
 Introducing a Color Palette to cache the allocation of Color objects by tht UITheme

jordanmontt:
Continuation of #12419 
I also removed a lot of method from `UITheme` that did not have any senders. Also, all the reviews and comments of the last PR (#12419) were addressed.


jordanmontt:
Ok, so there are still merge conflicts :(

jecisc:
If it is troublesome to manage a big change like that, maybe it’d be nice to cut it in multiple PRs by cherry picking the changes?
Like one to replace things like `self theme testColorError` by `TestResult defaultColorBackGroundForErrorTest`. One to remove the methods with no senders. One to introduce the dictionary. One to use the dictionary.

jordanmontt:
Thanks @jecisc I will try to solve the merge conflicts. If I cannot I will split this into multiple PR's

jordanmontt:
Merge conflicts solved :)

MarcusDenker:
This failure is related: 

RSMapBuilderTest>>testHighlightOnNode

jordanmontt:
Yes this is a failure of Roassal. I discussed with @akevalion

It is an easy fix. We need to replace the line `color := self theme testColorPassing` by `color := TestResult defaultColorBackGroundForPassingTest` in the method `RSThemedMapBuilder>>#highlightBoxes`

But first we need to integrate this PR to do this change because the method.

jecisc:
Would it be possible to do a first PR with what is needed for the roassal change, but keeping a green CI, then updating roassal, then merging this PR? 
integrating a broken test would mess with the CI for multiple days

akevalion:
hello there, I will create a pull fixing this issue

jordanmontt:
> Would it be possible to do a first PR with what is needed for the roassal change, but keeping a green CI, then updating roassal, then merging this PR? integrating a broken test would mess with the CI for multiple days

Yes but it will create 2 more PRs. First one PR changing `self theme testColorPassing` by `TestResult defaultColorBackGroundForPassingTest` for all classes except for Roassal. Wait to be integrate it and then do another PR to Roassal to update the same. And then the rest of this PR... 

If it is necessary I will do it, but if possible I would like to avoid it. This are changes  are difficult to load in a new image as they change the UITheme that is used everywhere.  We spent one whole afternoon with Guille only to load this changes in a new image. That image now is broken so I will need to do that again.

Ducasse:
I agree with seb we can have an idiot test broken for making sure that an important change gets through.

Ducasse:
Seb just a question before I integrate it. Did you deprecate some "public" methods (but now this is difficult who is using the UITheme).

jordanmontt:
The PR for updating Roassal is coming this afternoon. Milton already fixed it in the Roassal repository now we only need to update the Roassal version in Pharo.

> Seb just a question before I integrate it. Did you deprecate some "public" methods (but now this is difficult who is using the UITheme).

I didn't deprecate any public method but the API s the same. The difference is that now `UITheme` returns a color with `colorPalette aColor` instead of `Color new`. So the users of UITheme should not break because there are the same methods that return colors.

But I deleted like 40 methods that didn't had any users inside the image.

akevalion:
pr created!

Ducasse:
Capture 2023-02-21 at 17 51 52


I will revert the change

jecisc:
How did you hit that?

Since the beginning of the afternoon I’m using an image with this PR and everything works fine after I migrated my themes to the new mechanism

Ducasse:
I took the last one (oops deleted my image) and I could not open a tool.

jordanmontt:
Do you remember which tool? Like that we can see. @Ducasse 

One question, were you using a custom theme? Because if yes, maybe that is causing the problem. In the afternoon I helped Cyril to migrate his custom themes. We made an error and we got the EM, but then me migrated them without any problem

PR #12773, opened 2023-02-13
Remove SerialPort that relies on a vm plugin that is not in the vm anymore

### LINKED ISSUE #12767: Remove SerialPort?
It was brought in https://github.com/pharo-project/pharo/pull/12692 that SerialPort vm plugin does not exist anymore. If that is really the case, maybe the package should be removed from the system

jecisc:
Fixes #12767


astares (approved):

jecisc:
This was approved by Pablo: 

image

PR #12774, opened 2023-02-13
Revert back to `Apple Computer, Inc.`

10allday:


MarcusDenker:
Can you tell us why? But then, the account looks strange, all activity is private and the linked twitter looks odd, too

guillep:
Yes, I propose to close this until we have more information.

PR #12776, opened 2023-02-14
Factorize code in Zinc

jecisc:


astares:
@svenvc

jordanmontt (approved):

src/Zinc-Tests/ZnClientTest.class.st
```diff
@@ -1,6 +1,6 @@
 Class {
 	#name : #ZnClientTest,
-	#superclass : #TestCase,
+	#superclass : #ZnAbstractServerTest,
```
noha, line 3: So why is a client test a specification of a server test?

noha:
We really need to check with @svenvc . It does not make sense to change zinc without the sync with upstream.

noha:
Reviewing this PR I think introducing a class hierarchy for that is not the best option. The methods #withServerDo: and #randomPort are independent and therefor are candidates to be class methods. So deduplicating code by moving them as class side methods under ZnNetworkingUtils seems better.  They can still be extension methods from the test package but they generally useful. But let's see what @svenvc thinks

jecisc:
I was not sure about the class hierarchy indeed. I’d be open to change the way to factorize and I’d love to have the opinion of sven.

I am checking this because we currently have random failures in Pharo CI because sometimes tests are running in multiple images and some ports are used by other images. I’d like to see if we could improve the situation, maybe with a retry on a different port if one port is used. But before that, I’d like to remove the duplication to not implement the retry mechanism in multiple places

svenvc:
The core problem is that it used to be, and probably still is, very difficult to reliably check, across all platforms and operating systems, if starting a server on some port was successful or not.

Using random ports within a certain range makes it possible to, most of the time, run multiple tests in parallel, be it in the same image or on the same machine across images. Obviously, this is not perfect.

I know that, at least on some platforms, it is possible to to let the OS allocate a free port. The withServerDo: construct could support that.

In any case, changes like this need to be done in the Zinc repositories first, so that they can be tested independent of the main Pharo image and across Pharo versions (currently from Pharo 7 upwards).

jecisc:
I’ll close this PR and I opened an issue about the ci failures

Issue #12763, opened 2023-02-13
[CI] Windows testrunner has problems


MarcusDenker:
**Bug description**
The windows test runner on the CI seems too have some consistent problem (maybe disk space?)

MarcusDenker:
We get

```
Failed to extract bootstrap64.tar.gz
```

e.g https://ci.inria.fr/pharo-ci-jenkins2/blue/organizations/jenkins/Test%20pending%20pull%20request%20and%20branch%20Pipeline/detail/PR-12760/1/pipeline/48

<blockquote><div><strong><a href="https://ci.inria.fr/pharo-ci-jenkins2/blue/organizations/jenkins/Test%20pending%20pull%20request%20and%20branch%20Pipeline/detail/PR-12760/1/pipeline/48">Jenkins</a></strong></div></blockquote>

MarcusDenker:
I restarted the two virtual runners

MarcusDenker:
Did not help. Now we get

```
ci@WIN10-CI C:\builds\workspace\Test pending pull request and branch Pipeline\PR-12768\Tests-windows-64>bash -c 'bootstrap/scripts/runTests.sh 64 Tests-windows-64' 
+ CACHE=bootstrap-cache
+++ dirname bootstrap/scripts/runTests.sh
++ cd bootstrap/scripts
++ pwd -P
+ SCRIPTS='/cygdrive/c/builds/workspace/Test pending pull request and branch Pipeline/PR-12768/Tests-windows-64/bootstrap/scripts'
+ . /cygdrive/c/builds/workspace/Test pending pull request and branch Pipeline/PR-12768/Tests-windows-64/bootstrap/scripts/envvars.sh
bootstrap/scripts/runTests.sh: line 15: /cygdrive/c/builds/workspace/Test: No such file or directory
```

MarcusDenker:
It is now working again

MarcusDenker:
"Failed to extract bootstrap64.tar.gz" is back

MarcusDenker:
New problem:

```
Cannot contact pharo-ci-jenkins2-win10: java.lang.InterruptedException
Sending interrupt signal to process
After 20s process did not stop
```

MarcusDenker:
fixed by 

https://github.com/pharo-project/pharo/pull/12887

PR #12777, opened 2023-02-14
add some Duration tests

pavel-krivanek:
add testAsMicroseconds, multiplication etc


pavel-krivanek:
some windows tests failed, not related

MarcusDenker:
it is good to add  a link to the bug report of why this is failing, I just copy the title, this way we have this:

- [CI] Windows testrunner has problems #12763

The reason is that this way we know which problem still happens and which is the one that makes most sense to fix

jordanmontt (approved):

PR #12780, opened 2023-02-15
categorization tests should print culprints if any

demarey:
Get informative messages when categorization tests fail

### DIFF
--- a/src/ReleaseTests/ProperMethodCategorizationTest.class.st
+++ b/src/ReleaseTests/ProperMethodCategorizationTest.class.st
@@ -142,7 +142,7 @@ ProperMethodCategorizationTest >> testInstanceSideInitializeMethodNeedsToBeInIni
 
 	self 
 		assert: violating size <= 4 
-		description: 'Instance side #initialize methods should be in "class initialization" method category:' , violating asString
+		description: 'Instance side #initialize methods should be in "initialization" method category:' , violating asString
 ]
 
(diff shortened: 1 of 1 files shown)


src/ReleaseTests/ProperMethodCategorizationTest.class.st
```diff
@@ -131,11 +133,16 @@ ProperMethodCategorizationTest >> testInstanceSideInitializeMethodNeedsToBeInIni
...
+		cls methods do: [:method | 
+			(method selector = #initialize and: [ method isClassSide not and: [( method package name beginsWith: 'SmalltalkCI') not ] ])
+				ifTrue: [ instanceSideInitializeMethods add: method ]]].
 
 	violating := instanceSideInitializeMethods select: [:m | m protocol ~= #initialization ].
 
-	self assert: violating size <= 4 description: 'Instance side #initialize methods should be in "initialization" method category'
+	self 
+		assert: violating size <= 4 
+		description: 'Instance side #initialize methods should be in "class initialization" method category:' , violating asString
```
jordanmontt: I think there is a typo. The category shouldn't be "initialization" instead of "class initialization" since we are checking instance side initialize methods and not class side?

demarey: yes, there is a typo. thank you!

jecisc (approved):

PR #12782, opened 2023-02-15
FIx #wantsSteps deprecated calls

### LINKED ISSUE #12595: DNU #wantSteps on method remove
**Bug description**
When removing a method from the method browser --> DNU as #wantSteps is send to a non-morph

1 implementors of #acceptVisitorNeverFinishes: 
2 remove that method with cmd-x
3 see DNU

wantSteps should check if the model is a morph or a model?

jecisc:
The deprecated calls seems to happen when the model of a window is a collection and not an actual model

Fixes #12595

### DIFF
--- a/src/Morphic-Widgets-Windows/SystemWindow.class.st
+++ b/src/Morphic-Widgets-Windows/SystemWindow.class.st
@@ -1956,7 +1956,13 @@ SystemWindow >> wantsHalo [
 SystemWindow >> wantsSteps [
 	"Return true if the model wants its view to be stepped.  For an open system window, we give the model to offer an opinion"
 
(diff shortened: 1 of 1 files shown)


jecisc:
[CI] Windows testrunner has problems [CI] Windows testrunner has problems #12763

MarcusDenker (approved):

MarcusDenker:
We have in https://github.com/pharo-graphics/Bloc/issues/207 a situation where the receiver is a SyntaxErrorDebuger

I think we should really remove the deprecated method from object and check if the obects understands #wantSteps

jecisc:
https://github.com/pharo-project/pharo/pull/12793

PR #12778, opened 2023-02-15
Improve faulty parsing

privat:
Some improvements (and tests) related to the parsing of faulty code.

Basically, these are test cases on faulty code and the required fixes to the tests become green.
This PR (and future ones) focuses on having usable and consistent AST when parsing broken code.

Note: since I hacked some low level image stuff, I hope I did not break too many tests :)

PR #12785, opened 2023-02-16
Tests that CI handle properly Syntax Errors during tests

privat:
The best way to test the CI is to do a PR that runs the CI I assume :)

Cf https://github.com/pharo-project/pharo/pull/12778#issuecomment-1431549288


privat:
It could be related to #12595

privat:
CI broken as suspected.

jecisc:
> It could be related to #12595

There is a PR for this issue: https://github.com/pharo-project/pharo/pull/12782

But I don’t see how this issue could product this effect on the CI?

jecisc:
Is this still needed?

privat:
Is the CI still broken? Let's see...

privat:
CI still broken :(

MarcusDenker:
What this PR does to test:


```
 testParseExpressionDontFreeze
 	"Force a parse error (from a simple tool class).
 	The image (and CI) should not freeze :)"
 	self parserClass parseExpression: '#'
```	

We would expect that this leads to a failing test. But instead, the image quits.	

Handling Syntax errors is a bit, uhm, arcane... what happens is that the exception in the end
calls #syntaxErrorNotificationDefaultAction: on the UIManager.

And the non-interactive UI manager (and CommandLineUIManager) explicitly exit:

```
syntaxErrorNotificationDefaultAction: aSyntaxErrorNotification

	"display and log the syntax error"
	super syntaxErrorNotificationDefaultAction: aSyntaxErrorNotification.

	"in noninteractive mode simply quit"
	^ self exitFailure
```


While DummyUIManager>>#syntaxErrorNotificationDefaultAction: does something more resonable:

```
syntaxErrorNotificationDefaultAction: anException
	self traceCr: '*** SYNTAX ERROR ***'.
	self traceCr: anException.
	self
		traceCr: (thisContext stack first: (20 min: thisContext stack size)).
	anException return
```


we should maybe not do the  "self exitFailure" for NonInteractive and CommandLine

(and we should get rid of this #syntaxErrorNotificationDefaultAction: alltogether...)

privat:
A radical solution to the issue #12910 :)

privat:
Thanks to #12910, the CI now behave nicely.

MarcusDenker (approved):

PR #12787, opened 2023-02-16
Add document browser

demarey:


Ducasse:
Christophe what is the problem?

demarey:
DocumentBrowser has an implicit dependency to Iceberg - GitHub plugin.
I need to check how to solve this

astares (approved):

PR #12796, opened 2023-02-17
Condense condensation of sources

jecisc:
It takes so much time to review the condense PR that we might want to merge them all at once in the CI is good with it. 
This PR contains most the other PRs of @astares


jecisc:
The CI broke and it is hard to find from where it comes from. I’ll try to du a PR with just a subset of the changes.

PR #12800, opened 2023-02-17
Add #closeTo: on sequenceable collection

jecisc:
This is originally part of PolyMath in Core package but we are trying to remove this package and the closest package this methods could go to is Math-Operations-Extensions. Thus we are proposing this to Pharo.

### DIFF
--- a/src/Math-Operations-Extensions/SequenceableCollection.extension.st
+++ b/src/Math-Operations-Extensions/SequenceableCollection.extension.st
@@ -1,0 +1,21 @@
+Extension { #name : #SequenceableCollection }
+
+{ #category : #'*Math-Operations-Extensions' }
+SequenceableCollection >> closeTo: aSequenceableCollection [
+	"Return true if all my elements are close to the elements of same index of the parameter"
+
+	"#(1.9283901234902349 2.302949083493849) closeTo: #(1.9283901234902348 2.302949083493899) >>> true"
+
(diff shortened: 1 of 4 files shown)


src/Math-Operations-Extensions/SequenceableCollection.extension.st
```diff
@@ -0,0 +1,21 @@
+Extension { #name : #SequenceableCollection }
+
+{ #category : #'*Math-Operations-Extensions' }
+SequenceableCollection >> closeTo: aSequenceableCollection [
+	"Return true if all my elements are close to the elements of same index of the parameter"
+
+	"#(1.9283901234902349 2.302949083493849) closeTo: #(1.9283901234902348 2.302949083493899) >>> true"
+
+	self with: aSequenceableCollection do: [ :a :b | (a closeTo: b) ifFalse: [ ^ false ] ].
```
demarey: code duplication. 
Could use 
```smalltalk
self closeTo: aSequenceableCollection precision: Float closeToPrecision
``` 
where `#closeToPrecision` will return `0.0001` (extracted from `Float>>#closeTo:`)

jecisc: Oh good catch :) I’ll update the PR

MarcusDenker (approved):

jordanmontt (approved):

PR #12804, opened 2023-02-18
Fail doc comment release test if a comment test has a syntax error.

jecisc:
Currently the CIâ¯will juste break if that happens and the feedback is really really poor


jecisc:
Test failure are related. Apparently we already have executable comments with failures in the system

jecisc:
I think I did not understood well enough the executable comments structure. There is a bug in there but my PR is not the right way to fix it. I’ll open an issue about that.

PR #12810, opened 2023-02-19
extract return. fix: #12809

vonbecmann:
extract return from ifTrue: ifFalse: expression. i didn't replace == nil with isNil, not sure about it.

### DIFF
--- a/src/Kernel/Class.class.st
+++ b/src/Kernel/Class.class.st
@@ -798,11 +798,11 @@ Class >> possibleVariablesFor: misspelled continuedFrom: oldResults [
 	results := misspelled correctAgainstDictionary: self classPool continuedFrom: oldResults.
 	self sharedPools do: [:pool |
 		results := misspelled correctAgainstDictionary: pool classPool continuedFrom: results ].
-	self superclass == nil
+	^ self superclass == nil
 		ifTrue:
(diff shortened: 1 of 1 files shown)


src/Kernel/Class.class.st
```diff
@@ -798,11 +798,11 @@ Class >> possibleVariablesFor: misspelled continuedFrom: oldResults [
 	results := misspelled correctAgainstDictionary: self classPool continuedFrom: oldResults.
 	self sharedPools do: [:pool |
 		results := misspelled correctAgainstDictionary: pool classPool continuedFrom: results ].
-	self superclass == nil
+	^ self superclass == nil
```
vonbecmann, line 801: i didn't replace == nil with isNil because im not sure it is a good idea. idk the consequences.

vonbecmann:
> This issue has either a default title or empty body. We would appreciate it if you could provide more information. Note: I am not a very intelligent bot, I can only react to new comments. Please add a comment for me if you update the body or title.

body was updated

src/Kernel/Class.class.st
jecisc: I think it can safely be replaced by #ifNil:ifNotNil: here

jecisc (approved):

PR #12816, opened 2023-02-20
Improve faulty parsing of cascades

privat:
Improve the current parsing of faulty cascades and enrich RBInvalidCascadeErrorNode, change the logic and add documentation and tests (lots of tests).

The PR ensures that tokens in a faulty cascade are not lost when consumed during the error recovery.

Note that currently (and previously, the PR does not introduce that, it just generalizes it) a faulty cascade node is a legal `RBCascadeNode` instance but contains errors nodes in its `messages` instance variable.
The class `RBInvalidCascadeErrorNode` is still only used for the first item of a cascade node if there is an error on it.
Also, the class name is bad (and redundant), maybe `RBMessageError` could a better name. But the PR does not change that. I could if some comments ask for it :p
Anyway, client of faulty AST should still be careful when manipulating them.

A first logic change is that when an unexpected token is found after a `;`, then the whole cascade is aborted and the unexpected token is not consumed (the previous behavior was to drop the token and continue the cascade). It means that `a foo; 5 bar` will be parsed as two statements: an unfinished cascade `a foo ;` then a non-faulty message send `5 bar`.
It is more consistent with how other unexpected tokens are handled. (e.g. `a foo 5 bar` is parsed as two statements: a unfinished statement `a foo` and a non a faulty message send `5  bar`). This behavior also makes sense in the context of method edition with live feedback and highlighting where a new statement is added (thus initially unfinished) in the middle of existing code.

A second logic change is that if the first element is not a message send, then it is considered the receiver of the following messages. e.g. `a;foo;bar` will have 3 elements, a first faulty one as `a` alone is not a full message send, then two correct message sends `a foo` and `a bar`. This change reduces the number of faulty things in the AST.

### DIFF
--- a/src/AST-Core-Tests/RBErrorNodeParserTest.class.st
+++ b/src/AST-Core-Tests/RBErrorNodeParserTest.class.st
@@ -141,6 +141,202 @@ RBErrorNodeParserTest >> testFaultyCascade [
 ]
 
 { #category : #tests }
+RBErrorNodeParserTest >> testFaultyCascadeCascadeMessageExpected [
+
+	| node |
+	node := self parserClass parseFaultyExpression: '1 foo;'.
+
+	self assert: node isFaulty.
+	self assert: node isCascade.
+
+	self deny: node messages first isFaulty.
+	self assert: node messages first selector equals: #foo.
+
+	self assert: node messages second isParseError.
+	self assert: node messages second sourceInterval equals: (7 to: 6).
+	self
+		assert: node messages second errorMessage
+		equals: 'Cascade message expected'.
+	self assert: node messages second formattedCode equals: ''.
+
+	self
+		assert: node formattedCode
+		equals: '1<r><t>foo;<r><t>' expandMacros
+]
+
+{ #category : #tests }
+RBErrorNodeParserTest >> testFaultyCascadeCascadeMessageExpected2 [
+
+	| node |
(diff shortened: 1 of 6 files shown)


privat:
About `RBCascadeNode` that contains `RBParseErrorNode` objects inside its `messages` (instead of only `RBMessageNode` objects). I think this is fragile because clients of the AST will likely not expect that elements of `messages` can not be messages. However, I might worry to much as buggy existing clients (and future ones) can be fixed :)

An alternative approach could be to get a specific `RBEnglobingErrorNode` subclass for faulty cascades (in a similar way that is used to handle faulty complex nodes like `RBBlockErrorNode`). Such an approach will shield clients of the AST of unaware usage of broken cascades, but still allow them to handle them specifically if they want (or just handle them in a general way as indirect instances of `RBEnglobingErrorNode`).

Another approach could be to wrap broken elements  of a cascade (i.e. broken messages) in a new class `RBMessageErrorNode` that can mimic some operations of legitimate `RBMessageNode` instances (thus being compatible with most clients to try to fool them). However, this feels hackish and will likely be a mess since the there are a lot of methods (and extension methods) in `RBMessageNode` (and some that might be tricky to implement in a degraded way).

Anyway, future improvement of the handling of broken cascades will be done in future PR. This one can be merged as is if people are ok.

MarcusDenker (approved):

PR #12817, opened 2023-02-20
Improve faulty parsing (and more!) of byte array literals

privat:
Parsing: keep the content of faulty elements in byte array literals by interpreting the faulty element as classic literal array elements (instead of just skipping them).

And more: inject a RuntimeSyntaxError signal durint the compilation of faulty byte arrays literals, and make `RBLiteralByteArrayNode>>#value` more robust by returning a classic array if faulty (instead of a weird deep exception during the execution of the method).

Note: I'm not sure if this is the correct approach here. A safer alternative could be to return a `RBLiteralByteArrayErrorNode` instead of a legal `RBLiteralByteArrayNode` with broken elements. The current approach (unmodified by the PR, since it just improve the parsing, the error information and add some workaround around faulty RBLiteralByteArrayNode) is still fragile. Maybe the other way could also be simpler (less workarounds). Discussion?


guillep:
Should this PR be closed after the integration of https://github.com/pharo-project/pharo/pull/12818 ?

PR #12818, opened 2023-02-20
Improve faulty parsing of byte array literal (alternative version)

privat:
Parsing keeps the content of faulty elements in byte array literals by interpreting the faulty element as classic literal array elements (instead of just skipping them), and put them in an instance of `RBLiteralByteArrayErrorNode` instead of in a faulty `RBLiteralByteArrayNode` instance as previously.
The old way was fragile and broke clients (including compilation and execution), since a `value` of a `RBLiteralByteArrayNode` can only contain bytes.

Note: This is an alternative implementation of #12817 — there can be only one[.](https://www.youtube.com/watch?v=sqcLjcSloXs) that is somewhat simpler (you can compare the diffs) and more robust.
The only drawback is that `RBEnglobingErrorNode` (the superclass of the error node used) need love:
* weird spaces in formatting.
* `contents` should have a `s` (I think)
* mandatory error is problematic when the error is in the content (so no specific error message on the whole compound)
But I consider these to be out of the scope of the PR and should be solved independently.


privat:
Maybe a discussion point (that also applies to #12817 ): the choice currently (and previously) done in case of error is to consume elements until a closing bracket `]`. An alternative policy could be to consider having an unfinished literal (someone forgot the `]`, or is still writing and did not enter it yet) then continue the parsing as if a `]` was present (the node is still an error node, obviously).

MarcusDenker:
I restarted the CI run and now this looks odd: It shows an error and I can not restart it anymore. Very very odd

guillep (approved):
Looks good!

I wonder, why a lot of tests are removed, is that because they seem superfluous or it is a mistake?

MarcusDenker:
I think I did the conflict resolution wrong, the tests should have been added, not removed

privat:
I rebased the commits to solve the merge

src/AST-Core-Tests/RBErrorNodeParserTest.class.st
```diff
@@ -363,6 +363,29 @@ RBErrorNodeParserTest >> testFaultyCascadeMessageExpected4 [
 	self
 		assert: node formattedCode
 		equals: '1<r><t>;<r><t>bar' expandMacros
+RBErrorNodeParserTest >> testFaultyLiterabByteArray [
```
guillep: There is a syntax error here...

guillep: There is a missing closing bracket between lines 365 and 366?

guillep (changes requested):
The ci fails because of the syntax error I think

privat:
oops. I shouldn't have solved the conflicts by hand in a text editor before the first coffee.

privat:
tests are ok (but the windows box is dead)

MarcusDenker (approved):

MarcusDenker:
I like this solution better than the alternative, as here we know that we have a wrong ByteArray

guillep (approved):

PR #12819, opened 2023-02-21
InitializeSlot causes strange behavior that leads to error

### LINKED ISSUE #12811: InitializeSlot causes strange behavior that leads to error
Create a class with the following definition:

```
Object << #TestClass
	slots: { #slotA => InitializedSlot default: OrderedCollection new };
	sharedVariables: {};
	package: ''
```

click out and back to the class definition, it changes to:

```
Object << #TestClass
	slots: { #slotA => InitializedSlot default: an OrderedCollection };
	sharedVariables: {};
	package: ''
```

`an OrderedCollection` shows red indicating a syntax error.  If you try to edit the class definition the following error appears:

image

 - OS: Windows 10
 - Pharo Version 10

astares:
Fix #12811


jordanmontt:
It would be nice if you can add some tests

MarcusDenker:
One think to note: the default value is written to the instance after "new". For a collection, you would want a *new* collection for each instance. 

The problem with "OrderedCollection new" is that it will be evaluated just once (when you create / update the class).

Inao0 (approved):
I checked that one with @tesonep.

MarcusDenker:
Did you chech the CI tests?

unix-64 / Tests-unix-64 / testPrintingUsingPrintStringForBlocks – Unix64.VariablesLibrary.Tests.Tests.InitializedSlotTest
unix-64 / Tests-unix-64 / testPrintingUsingStoreStringWithoutBlocks – Unix64.VariablesLibrary.Tests.Tests.InitializedSlotTest

are now failing for all CI runs.

to me it looks like you merged this even though the CI tun failed completely for the last commit (no tests were run, so we can not even say if these tests fail due to this PR)

guillep:
Let's revert it, no need to make a fuzz when it can be solved easily :)
The CI is very charged on a sprint day.

PR #12823, opened 2023-02-21
Filter on substrings in DrTest

### LINKED ISSUE #12788: DrTest: Filter with includes substring instead of begins with
It happens a lot to me that I want to run the tests of a project that is part of a meta-project. For example AI-HierarchicalClustering-Test. To select this package in DrTest I type "Hier" in the package list filter. 

Currently there is not result because my package starts with "AI-", but in the previous test runner, it was filtering by also returning the elements including the pattern given. 

I wonder if this could not be a better behivor also for DrTest. 

Anyone has an opinion on that?

jecisc:
Currently it was filtereng elements only if they begun by the pattern. Now, they filter if the item includes the pattern

Fixes #12788


MarcusDenker:
Tests failing in DTFilterableListPresenterTest

privat (approved):

PR #12824, opened 2023-02-21
Sync zinc

jecisc:
Brings: 
- [Add ZnCharacterEncoder>>#ensureAtBeginOfCodePointOnStream: to be used](https://github.com/svenvc/zinc/commit/952f5f4a689206736d38cbe9a00a6661628859a0)
- [Protocol name change testing->tests](https://github.com/svenvc/zinc/commit/87f1e9b28861c4d95d9c3da7e4ff26c405d03bb6)
- [Rename isFixed -> isFixedLength, isVariable -> isVariableLength](https://github.com/svenvc/zinc/commit/439b9a18112e250011fcbafa82783de814edbaab)
- [Add a more comprehensive ZnCharacterStreamTest>>#testReadStreamManipuâ¦](https://github.com/svenvc/zinc/commit/8f1a2554de51fc8c7f9f848b392526d9f066f2bd)

We hope that this will help to fix issue 12575


svenvc:
LGTM

Thanks !

MarcusDenker:
Two failing tests:

OmBlockFileStoreTest>>testMultibyteUnicodeCharacterCrossingTwoBlocks
	
```Reading entry from File @ /tmp/ombu-tests/testingStoree4io677ch48goj24wbq67vs0v.ombu@393216: Message not understood: ZnBufferedReadStream >> #back```

OmFileStoreTest>>testLoadFromCorruptFile4

```Message not understood: ZnBufferedWriteStream >> #peek```

**Comment by svenvc** [comment]
OK, I will look into them tonight.

**Comment by svenvc** [comment]
This is way harder than I thought.

ZnBufferedReadStream>>#back was indeed missing, but adding it only results in the problem that the new position logic tries to move too much backwards (past the beginning). The question is why ?

Looking at the test, I don't understand the intent, nor how this Om File/Block Store works or is designed.

The second error is more conceptual: the write stream inherits the new position logic that was meant for a read stream, hence peek and back are not understood. This is normal. 

The question is, should position on an encoded write stream also make sure you do not end up inside one character ? Probably.

I think I could rewrite the second test to circumvent the issue (the goal is just to generate a bogus file).

It is also important to note what I did with the positioning of course assumes that the input, the underlying bytes, are correctly encoded, else all bets are off. Is assuming that too much of an ask ?

**Comment by jecisc** [comment]
Maybe @tinchodias can provide some clues about the way ombu works? :)

**Comment by svenvc** [comment]
Here is a version of the second test that works (by circumventing the issue):
```
testLoadFromCorruptFile4
	"Content is corrupted"

	| loadedStore stream |
	store newEntry: (OmEntry content: 41).
	store newEntry: (OmEntry content: 42).
	store flush.

	"Override some characters"
	stream := store fileReference binaryWriteStreamDo: [ :writeStream |
		writeStream position: 50; nextPutAll: '**' utf8Encoded ].

	loadedStore :=
		(OmFileStore
			named: store globalName
			inDirectory: self directoryForTests)
			yourself.

	self
		should: [ loadedStore refresh; entries ]
		raise: OmFileStoreReadingError
```
And as far as I understand, it is clearer.

**Comment by jecisc** [comment]
I added this updated test in the PR

**Comment by tinchodias** [comment]
already fixed?

**Comment by jecisc** [comment]
@tinchodias Sven had two problems. 

He fixed the second one, but not the first one:

>ZnBufferedReadStream>>#back was indeed missing, but adding it only results in the problem that the new position logic tries to move too much backwards (past the beginning). The question is why ?
>
> Looking at the test, I don't understand the intent, nor how this Om File/Block Store works or is designed.

**Comment by jecisc** [comment]
The failing test in question is `OmBlockFileStoreTest>>testMultibyteUnicodeCharacterCrossingTwoBlocks`

**Comment by svenvc** [comment]
> already fixed?

The test #testMultibyteUnicodeCharacterCrossingTwoBlocks is acting up.

What is the purpose of OmBlockFileStore vs OmFileStore ?
Are the blocks (which seem to be in the same file) a way to speed up accessing ? 
Are positions being cached ?

The test seems to be about splitting a multi-byte character over a block boundary.

The new position: logic will now move back until it is at a character start. However, if we are at the start (of a block I am guessing), that will fail, since we cannot go back further.

Does this make sense ?

**Comment by svenvc** [comment]
@jecisc the following commit in Zinc is part of the changes regarding #position: (but not directly relevant to the 2 issues)

**Comment by svenvc** [comment]
With all the latest Zinc code loaded, the following shows the current problem:

```
bytes := (String new: 64 withAll: (Character value: 266)) utf8Encoded.

(ZnCharacterReadStream on: (ZnBufferedReadStream on: bytes allButFirst readStream)) in: [ :stream |
  stream position: 0; next ].
```

The character in the example is encoded using 2 bytes in UTF8, chopping off the first byte of the byte sequence makes the sequence invalid. With the new position: logic the stream tries to move backwards from the first position to a valid character start, which does not exist, hence it hits the 'cannot move back from beginning' error.

Basically, I think this is correct behaviour.

**Comment by svenvc** [comment]
It feels to me as if the OmBlockFileStore logic should be updated to prevent it from splitting an encoded character, if that makes sense or is even possible.

**Comment by tinchodias** [comment]
Hello, I didn't understand well what changed to make the test fail. But I'd be happy to explain a bit of the Ombu and Epicea if you can help me to get it.

An .ombu file is a sequence of ston serializations. The naive way to inspect the contents:
```smalltalk
ombuFile := ('./pharo-local/ombu-sessions' asFileReference filesMatching: '*.ombu') anyOne.

changeEntries := OrderedCollection new.

OmCurrentGlobalName
	value: 'Test'
	during: [
		ombuFile readStreamDo: [ :readStream |
			| aReader |
			aReader := STON reader
				on: readStream;
				allowComplexMapKeys: true;
				yourself.
			
			[ readStream atEnd ] whileFalse: [
				changeEntries add: aReader next ] ].				
		].

changeEntries inspect.
```

But this became much more complex when there were issues with large files that took too long to be parsed in the browser. I implemented the `OmBlockFileStore`, that reads the file by blocks of a fixed number of bytes. 

The test that is failing is associated with this discussion:
http://forum.world.st/Non-deterministic-ombu-errors-td4970592.html#a4975137

I'll continue tomorrow. Cheers

<blockquote><div><strong><a href="http://forum.world.st/Non-deterministic-ombu-errors-td4970592.html#a4975137">Pharo Smalltalk Developers - Non-deterministic ombu errors</a></strong></div><div>Non-deterministic ombu errors. Hi, I'm having some non-deterministic errors in the Travis builds, the error is always something like: *OmFileStoreReadingError: Reading entry from File...</div></blockquote>

jecisc:
@tinchodias le problem seems to be that UTF-8 does not have a fixed number of bytes for a character. And it happens that the blocks can chain in the middle of a character and this breaks IIUC

jecisc:
@tinchodias Do you have an idea to improve the situation? 
It’s a little hard to know how to fix this issue without having lot of knowledge about ombu

tinchodias:
true... let me check today/tomorrow

jecisc:
Thank you Martin! :) 
If we could fix this issue it will make the CI much more robust

tinchodias:
Hi @jecisc I debugged the test but need to continue tomorrow.

tinchodias:
I'm trying to reimplement the OmBlockFileStore: work on the binary stream without utf-8 encoding, just scan occurrences of `'Ombu {' utf8Encoded` plus an extra byte to differentiate it from the literal, which could appear inside a method's source code.

jecisc:
Thank you Martin! :)

tinchodias:
I went back yesterday to this issue... I was a bit blocked on how to fix it "elegantly".

I ended writing an anolog of Zinc's position: but that skips forward instead of backwards, if the encoding requires it.
For the moment this is my best approach... what's your opinion @svenvc ?

Code:
```
ZnEncodedReadStream>>
positionAtNextBeginCodePointStartingAt: aPosition
	"Set the byte position in the underlying/wrapped binary stream to aPosition.
	This is not a character based position! Positions are zero based.
	I will move forward if aPosition is not at the beginning of a code point."

	super position: aPosition.
	self encoder skipToValidCharacterOnStream: stream.
	peeked := nil
```
and:
```
ZnUTF8Encoder>>
skipToValidCharacterOnStream: stream
	"Ensure that the current position of stream is a the beginning of an encoded code point,
	if not move further forward. This is necessary when a position in the binary stream is set,
	not knowing if that position is on a proper encoded character boundary."

	[ (stream peek bitAnd: 2r11000000) == 2r10000000 ] whileTrue: [ stream next ]
```

but it would require more on the hierarchy of encoders and streams.

tinchodias:
I pushed this commit with my draft fix: https://github.com/tinchodias/pharo/commit/0ff5bcc2effeab5e0aae248cc9d0461420dd2456

What do you think?

svenvc:
Interesting, indeed, both going backwards as well as going forwards are possible when positioning in between encoded character boundaries.

I will have a look and this about it. It would make the interface pretty complicated though, to offer both options.

svenvc:
Thanks @tinchodias for the excellent proposal, I took it over with some slight renaming:

https://github.com/svenvc/zinc/commit/be478d9bdeabc8369e9a8f532c3577fe1cc5d32a

The difference can be illustrated as follows:
```
"the fourth byte position (zero based) is inside the encoded code point for è (at the second byte)"

(ZnCharacterReadStream on: 'élève' utf8Encoded readStream)
	position: 4; "implicit backward move"
	next. "$è"
	
(ZnCharacterReadStream on: 'élève' utf8Encoded readStream)
	positionForward: 4;
	next. "$v"
```

tinchodias:
Great @svenvc it looks very good, I like the names. @jecisc -> How should I collaborate with the adaptation to Ombu? if I push the change to my repo as a branch of this PR branch, then you can pull it and push it to your repo. No?

jecisc:
I’m planning to integrate Sven changes to this pr Monday (I don’t have my computer before Sunday evening :( )
If you branch from there and link me the branch I can check it up Monday :)

tinchodias:
Yes of course no problem!

jecisc:
@tinchodias I just did another sync of Zinc in this PR

jecisc:
@tinchodias in the end do you have a new version of the ombu blocs? Can I integrate it to this PR or does it still need more work?

jecisc:
I checked Martin repo and I tried to propose a new version with Sven's new methods. Let's see if that fixes the CI

jecisc:
Yes it passed!

Have you ever dreamed of never seen "OmFileStoreReadingError: Reading entry from File blabla: Illegal leading byte for utf-8 encoding" ever again in your CI? If you have you will love this PR because if everything goes well, you will never need that again! (At least in P11 and later)

Fixes #12575

MarcusDenker (approved):

tinchodias:
Thanks @jecisc sorry for delay. Great that you could adapt the code to the new selectors! cheers.

svenvc:
Thank you all for the enormous effort, great work.

PR #12825, opened 2023-02-21
Revert " Introducing a Color Palette to cache the allocation of Color objects by tht UITheme"

Ducasse:
Reverts pharo-project/pharo#12771


Ducasse:
So I will revert the revert since Cyril and Seb told me that it is working.

PR #12830, opened 2023-02-22
Executable comments should manage string escaping

jecisc:
Let's imagine that we add an executable comments the should check that the result is `This is "exagerated" at least`. Since it is in a comment, we need to escape some caracters and it becomes `This is ""exagerated"" at least`.  Currently the compared value did not remove the escaped double quotes.


svenvc:
I understand the fix, but where does the doubled double quote come from ?

I mean, it should not be there in the first place, right ?

The doubled double quote is a way to have a double quote inside a comment, but it is only an external representation, a syntax. When you ask the second comment from parsed source code, it should be gone.

svenvc:
```
(STON class>>#toJsonString:) comments second.
```
If in the source code, the second comment contains an embedded double quote, the parser should resolve it and just return the comments' content, without the doubled double quote, just like it removes the beginning and ending double quotes.

Do you agree ?

jecisc:
Now that you bring it, it is indeed strange that the double quote is here in the first place.

It comes from the ast

```st
(STON class>>#toJsonString:) ast comments second
```

I’ll try to see if I can change my fix to remove the escape before in the process of executing the executable comment which seems more logical

svenvc:
Maybe @MarcusDenker has an idea ?

MarcusDenker:
I will check..

privat:
I fell that the escaping of the comment contents is not done at the right place.

The earliest place is `RBScanner>>#scanComment` and we can see that escaping is managed here, but that both `"` are nextPut in the contents of the buffer. I think that's a mistake and only one should be put.

I'm not sure how many clients of the AST expect that comments are kept escaped but I think that most need what the programmer means (not the verbatim source text) — maybe only few programmers know that `"` can be escaped in comments.

svenvc:
I tried commenting out the double nextPut:

```
scanComment
	| start stop |
	start := stream position.
	buffer reset.
	self step.
	self atEnd
		ifTrue: [ ^ self scannerError: 'Unmatched " in comment.' ].
	[ currentCharacter = $" and: [ "buffer nextPut: currentCharacter." self step ~= $" ] ]
		whileFalse: [ characterType = #eof
				ifTrue: [ ^ self scannerError: 'Unmatched " in comment.' ].
			buffer nextPut: currentCharacter.
			self step ].
	stop := self atEnd
		ifTrue: [ stream position ]
		ifFalse: [ stream position - 1 ].
	comments add: (RBCommentToken value: (buffer contents copyFrom: 1 to: buffer contents size -1) start: start stop: stop)
```

That seem to work, but now the syntax highlighting of the executable comment is broken and flagged by a rule.

guillep:
@jecisc what should we do with this?

guillep:
We discussed with @jecisc . We agree with @svenvc and @privat in that the comment should be already unscaped.
There is probably a PR that will arrive later in the afternoon with that.
We will keep this PR open and rework it a bit for the test.

jecisc:
This is not so simple. Since the RBComment node has a source that is shorter than what is actually present in the code, it messes with the syntax highlight.

For each escaped character, the syntax highlight stop 1 character earlier.

This is a big change to do. Too much for me right now. I propose to merge my original fix except if someone with more knowledge on the parser and AST is willing to do this change

svenvc:
Maybe, but the problem is similar to other literal constants: single quotes embedded in strings or let's say a byte array. In both cases the external source code representation will be different than the internal object. In the end, it seems only the [start,stop] interval in the source code is relevant.

I could understand that you want to move forward, but this should then remain an open issue.

jecisc:
I added another test. The goal is that if one day the parser is updated to remove the escaping characters, it will fail and we will now we have to remove the fix I introduced

guillep:
Hi all, I created a new issue from this discussion. I propose we integrate this one, the new tests look good. Somebody that would like to work in the parser could do the further improvements :).

svenvc:
OK

jecisc:
Test failure not related

privat:
Since I'm waist-deep in the parser and scanner, I might have a look. For the moment, I propose that people do not include comments in comments (and stay safe on the shore) :p

privat:
Ok, I looked at it and I understand the coloring issue mentioned by @svenvc. Note that the same coloring issue should also be present with the current PR.

To get the expression, the content of the comment is used. In case of escaped double-quotes in the comment, an unescaping must be done (either by something in the executable comment package, as in the PR. or by RBScanner from the AST package as I prefer), then the string is parsed and compiled and executed...

For highlighting, `SHRBTextStyler` is simply extended with `styleDocExpression:in:` that independently highlight the commented expression (the escaped version obviously, as the AST is used for highlighting and nested comments should be treated correctly) to get an autonomous colorful Text object only for the comment.
Then the hack: the method simply takes the style information of each character in the small text and copy it to the corresponding character in the original source code.
As you may guess, if there is an escaped double quote, then the corresponding character is off by one, and if there is two escaping (as like Sith they often go by pairs), the whole color is off by two, etc. As you can see in the screenshot

Capture d’écran du 2023-02-24 13-49-43

So I will PR something to fix that, this seems doable

PR #12833, opened 2023-02-23
Refactoring `TDebugger class>>#handlesContext:` to `TDebugger>>#handlesDebugSession:`

adri09070:
In the Oups debugger system (`OupsDebuggerSelector>>#nextDebuggerForSession:`), to choose the next debugger that should open, it relies on the method `TDebugger class>>#handlesContext:` that returns true by default.

The context itself is often not enough to decide which debugger should open because it provides far less information than the debug session. Notably, it provides: the exception that opened the debugger, the process under debug and it also tells if the error was in the UI process.

This allows users to better decide when their specific debugger should open.

This also allows to centralise in `TDebugger class>>#handlesDebugSession:` the code that checks if the exception is handleable by a debugger: 

```Smalltalk
aDebugSession exception isHandleableBy: self
```

This condition is necessary and can't be removed because, by default, a debugger can handle ANY exception except errors that occurred during this debugger's opening


adri09070:
@StevenCostiou

MarcusDenker:
failing tests look related:

testHandlesContext – MacOSX64.Debugger.Model.Tests.Core.TDebuggerTest
 testNextDebuggerForDebuggerFailure – MacOSX64.Debugger.Oups.Tests.Tests . strategies.OupsDebuggerSelectorTest
estOpenDebuggerError – MacOSX64.Debugger.Oups.Tests.Tests . strategies.OupsDebuggerSelectorTest

adri09070:
I fixed one of them. The other two fail because I moved the condition `aDebugSession exception isHandleableBy: aDebugger` from `OupsDebuggerSystem` to `TDebugger class`. 

As OupsDummyDebugger does not use this trait, the tests fail because Oups chooses the same debugger twice without this condition.

So, I can:

- force `OupsDummyDebugger>>#handlesDebugSession:` to call `TDebugger class>>#handlesDebugSession:`. This is ugly and I don't like that because it highlights the fact that a debugger could handle its own errors if it overrides `#handlesDebugSession:` without calling `super handlesDebugSession: aDebugSession`
- moving this condition to where it was before, in `OupsDebuggerSelector>>#nextDebuggerForSession:`. Debuggers could still do additional checks of the exception by overriding `TDebugger class>>#handlesDebugSession:`. But I feel like it duplicates the exception check.

This condition exists only because of the exception `OupsDebuggerFailure` and is annoying
I don't really know which is better

adri09070:
I kept the second solution to ensure that the exception handling condition is checked for every debugger.
The first one can't work because it is not possible to call super as the debugger is a direct user of the `TDebugger` trait

adri09070:
Why is the CI red ? Everything looks fine now

MarcusDenker:
The win test runner again has problems

guillep (approved):
Good! thanks!

PR #12835, opened 2023-02-23
NumberParser: kill `requestor`

privat:
Remove requestor in NumberParser since it is an unused and very limited API.


jecisc (approved):
Seems good.
Worst case this can be managed with a catch

guillep (approved):
Yes!

PR #12838, opened 2023-02-24
Deprecate all nativeboost methods

### LINKED ISSUE #12781: ByteArray >> nbInt64AtOffset == ByteArray >> nbInt32AtOffSet
**Bug description**
The code of nbInt32AtOffSet is exactly equivalent to the one in nbInt64AtOffset.
This is also the case for at:put: version.
A ByteArray answering this answers 32 bits for both those selector.

I do not know who uses this , but this looks like a bug to me.
At least, I couldn't use it.

**Expected behavior**
The 64 version should take into account 64 bits, not 32.

**Version information:**
 - Pharo 11

**Expected development cost**
Trivial.
Either a removal because this is unused, or a fix that takes into account 64 bytes.

If I missed something, I hope someone will be able to explain it to me =)

### LINKED ISSUE #12837: Deprecate all nativeboost methods
**Describe the problem**
Those methods are the one starting with 'nb' in Byte array (ex: nbFloat32AtOffset:)

Inao0:
Fix #12837 and Fix #12781


guillep (approved):
Great! Thanks!

PR #12839, opened 2023-02-24
Remove zinc file system from bootstrap

jecisc:
Try to remove Zinc-FileSystem by removing the reference from the bootstrap. 

I need to see if the CI will work now

### DIFF
--- a/src/Zinc-FileSystem/package.st
+++ b/src/Zinc-FileSystem/package.st
@@ -1,1 +1,0 @@
-Package { #name : #'Zinc-FileSystem' }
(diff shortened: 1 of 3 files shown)


src/Zinc-FileSystem/package.st
```diff
@@ -1 +0,0 @@
-Package { #name : #'Zinc-FileSystem' }
```
guillep, line 1: Is this removal ok?

guillep:
Still failing: 

Error: No versionInfo for package: Zinc-FileSystem in File @ /builds/workspace/est_and_branch_Pipeline_PR-12839/64/bootstrap-cache/pharo-local/package-cache

src/Zinc-FileSystem/package.st
jecisc: Yes, for now we keep an empty package because the bootstrap fails without it.
I’m trying to remove the references to it in the bootstrap to not need it anymore. 
but from your second comment, this still needs work

jecisc:
Failing test not related :)

guillep (approved):
Good!! Thanks!

PR #12843, opened 2023-02-24
InitializeSlot causes strange behavior that leads to error

guillep:
Re-made PR of https://github.com/pharo-project/pharo/pull/12819

InitializeSlot causes strange behavior that leads to error

Tests are failing, they should be fixed before integration.

### DIFF
--- a/src/VariablesLibrary-Tests/InitializedSlotTest.class.st
+++ b/src/VariablesLibrary-Tests/InitializedSlotTest.class.st
@@ -5,6 +5,33 @@ Class {
 }
 
 { #category : #tests }
+InitializedSlotTest >> testBlockWithNonSharedDefault [
+	"If we use a block the resulting default value should not be shared among instances"
+
+	| slot objectA objectB defaultA defaultB |
+	slot := #slot1 => InitializedSlot default: [ OrderedCollection new ].
+	aClass := self make: [ :builder | builder slots: { slot } ].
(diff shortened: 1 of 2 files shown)


MarcusDenker:
failing tests:

InitializedSlotTest>>testPrintingUsingPrintStringForBlocks
InitializedSlotTest>>testPrintingUsingStoreStringWithoutBlocks

MarcusDenker:
The solution is not good: any use of an object that is created at accept time would lead to very strange effects, as this would share the object between instances.

a better (for now) solution is to restrict the the default value to be a literal (read only) or a clean block (evaluated each time the class is instantiated).

When creating slots via evaluation, we can just raise an error. The class parser then can show a nice user level error message when parsing the class definition. 

```
testPrintOn

 	| slot |
 	slot := #slot1 => InitializedSlot default: [ OrderedCollection new ].
	self
		assert: slot definitionString
		equals: '#slot1 => InitializedSlot default: [ OrderedCollection new ]'.

	slot := #slot1 => InitializedSlot default: #literal.
	self
		assert: slot definitionString
		equals: '#slot1 => InitializedSlot default: #literal'.

	"On eval, we raise an error, the ClassParser can show a nicer error in the class definition"
	self should: [#slot1 => InitializedSlot default: 5@2] raise: Error.
	self should: [#slot1 => InitializedSlot default: OrderedCollection new] raise: Error
```

I will close this and commit a version with this solution.

PR #12852, opened 2023-02-24
remove method Object asOrderedCollection

### LINKED ISSUE #11146: Object>>#asOrderedCollection should be deprecated
**Bug description**
In my code, I was getting in an instance variable (`shapes`) this content: `#(nil)` wondering where it comes from. In the end, I realized that it was caused by this code: 

```smalltalk
newShapes := shapes copy asOrderedCollection.
```` 

The problem was, of course, that the instance variable was not properly initialized and then `nil asOrderedCollection` returns an OrderedCollection with nil. 

I propose that the `Object>>#asOrderedCollection` should be deprecated and replaced with something with more explicit name. I'm not sure what the name should be (`withinOrderedCollection` or something like that). 

Notice that there is no `Object>>#asArray` nor `Object>>#asSet`

### PR DESCRIPTION
**Pull request by jordanmontt**
Remove the method `Object>>#asOrderedCollection` and updated `Symbol>>with:`

Fixes #11146

### DIFF
--- a/src/Kernel/Object.class.st
+++ b/src/Kernel/Object.class.st
@@ -360,13 +360,6 @@ Object >> asLinkPrepend: anObject [
 ]
 
 { #category : #converting }
-Object >> asOrderedCollection [
-	"Answer an OrderedCollection with the receiver as its only element."
-
-	^ OrderedCollection with: self
-]
-
-{ #category : #converting }
 Object >> asString [
 	"Answer a string that represents the receiver."
 
--- a/src/Kernel/Object.class.st
+++ b/src/Kernel/Object.class.st
@@ -360,6 +360,16 @@ Object >> asLinkPrepend: anObject [
 ]
 
 { #category : #converting }
+Object >> asOrderedCollection [
+
+	self
+		deprecated:
+		'The usage of this method is not recommended. We want to have a smaller Object api. We will remove this method in the next Pharo version.'
+		transformWith: '`@receiver asOrderedCollection' -> 'OrderedCollection with: `@receiver'.
+	^ OrderedCollection with: self
+]
+
+{ #category : #converting }
 Object >> asString [
 	"Answer a string that represents the receiver."
 
--- a/src/Kernel/Object.class.st
+++ b/src/Kernel/Object.class.st
@@ -360,16 +360,6 @@ Object >> asLinkPrepend: anObject [
 ]
 
 { #category : #converting }
-Object >> asOrderedCollection [
-
-	self
-		deprecated:
-		'The usage of this method is not recommended. We want to have a smaller Object api. We will remove this method in the next Pharo version.'
-		transformWith: '`@receiver asOrderedCollection' -> 'OrderedCollection with: `@receiver'.
-	^ OrderedCollection with: self
-]
-
-{ #category : #converting }
 Object >> asString [
 	"Answer a string that represents the receiver."
 
--- a/src/Collections-Strings/Symbol.class.st
+++ b/src/Collections-Strings/Symbol.class.st
@@ -320,7 +320,8 @@ Symbol class >> thatStartsCaseSensitive: leadingCharacters skipping: skipSym [
 
 { #category : #'instance creation' }
 Symbol class >> with: aCharacter [
-	^self newFrom: aCharacter asOrderedCollection
+
+	^ self newFrom: (OrderedCollection with: aCharacter)
 ]
 
 { #category : #'instance creation' }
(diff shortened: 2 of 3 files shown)

### CODE REVIEW

**Review by privat — APPROVED** [review]

**Comment by astares** [comment]
You might break some external packages relying on that - maybe we should better deprecate it first  (with a transformation) in Pharo 11 and then remove in Pharo 12. At least this would help people migrate.

**Comment by jordanmontt** [comment]
The 2 failing tests are related. It is in the graph algos library. I fixed that in this PR:  https://github.com/pharo-project/pharo/pull/12851

We need to merge that PR first

**Comment by jordanmontt** [comment]
Now all the tests are passing!

**Comment by astares** [comment]
@jordanmontt We break peoples code for all those external packages who use it at the moment. Why delete, why not a deprecation with a transformation first?

**Comment by jordanmontt** [comment]
It is just an issue that we did with @hernanmd for the Pharo Sprint. Since Pharo 11 is the development version, we think is better to just delete the method as it is said in one comment of the issue.

**Comment by jecisc** [comment]
Personally I agree that methods that have high chances to be used outside Pharo are better to be deprecated for 1 version of Pharo because it helps a lot with the migration. And here there was already a project that got broken with the graph algo, proving that this method is used.
The deprecation does not cost much, and the automatic transformation helps a lot to not break projects

**Comment by astares** [comment]
@jordanmontt Yes - Pharo 11 is alpha and in development. But at one point in time it will be released. 

There are people who use Pharo in their daily work, a few have even built a small business with like @SabineMa. Often they require time to move own code from one version to the next. But we would like to support them in migration, no?

If we just remove the method(s) we accept that external code who is using the method directly breaks. But this might make it
impossible for the users to load the code or run it in the new version of Pharo to fix all issues. They would need to fix in the old version of Pharo - but sometimes they already want to switch to more recent classes or methods they only find in the new image.
 
So with a proper deprecation and (if possible also a transformation) we can simply declare the method as deprecated and support them for one single iteration. So we keep them in Pharo 11 but deprecated and throw them finally out of the image in the Pharo 12 iteration.

People when porting from P10 to P11 can then still use their code code but they will get a warning that they need to change the callers. And they can fix it directly in P11 image while adopting.

With a transformation it is even more easy as when the code gets called the system automagically converts the caller. 

So we should not break by intention and offer proper deprecation wherever possible.

Also see http://forum.world.st/Deprecation-guide-for-methods-classes-and-packages-td5085081.html


<blockquote><div><strong><a href="http://forum.world.st/Deprecation-guide-for-methods-classes-and-packages-td5085081.html">Pharo Smalltalk Users - Deprecation guide for methods, classes and packages</a></strong></div><div>Deprecation guide for methods, classes and packages. Hi, From Bug #20900/ PR 1818 it looked like it is not really clear on how to "deprecate" correctly. Therefore I summarized and wrote a short...</div></blockquote>

**Comment by privat** [comment]
The method exists since at least 6 years ago (Pharo7?) https://github.com/pharo-project/pharo/blame/a23630c03222dd574461c5e5ac8ae3ecb5aa3932/src/Kernel/Object.class.st#L363 (blame stopped at source format change, I did not search further). A depreciation is more than welcome.

<blockquote><div> GitHub</div><div><strong><a href="https://github.com/pharo-project/pharo">pharo/src/Kernel/Object.class.st at a23630c03222dd574461c5e5ac8ae3ecb5aa3932 · pharo-project/pharo</a></strong></div><div>Pharo is a dynamic reflective pure object-oriented language supporting live programming inspired by Smalltalk. - pharo/src/Kernel/Object.class.st at a23630c03222dd574461c5e5ac8ae3ecb5aa3932 · pharo...</div></blockquote>

**Comment by jordanmontt** [comment]
Created related issue: https://github.com/pharo-project/pharo/issues/12881

**Comment by jecisc** [comment]
Thank you!
If you add this method as an extension of Depreacted11, it will be removed automatically when we will remove Depreacted11 from the baseline also

File: src/Kernel/Object.class.st [hunk 1]
```diff
@@ -359,6 +359,16 @@ Object >> asLinkPrepend: anObject [
 	^ self asLink asLinkPrepend: anObject
 ]
 
+{ #category : #converting }
```
astares, line 362: May we can already **package it into "Deprecated 11" package as an extension** as we later remove this package (and so we do not forget to remove this method)

So put it into *Deprecated11 category please

jordanmontt: It is already into Deprecated11 package. See the last commit

jecisc (approved):

src/Kernel/Object.class.st
astares: Cool - thanks!

PR #12854, opened 2023-02-24
Improve faulty parsing: rename content as contents (a s was missing)

privat:
The English language is weird (syntax doesn't even fit on a postcard...)

Note the impact of such a small letter change was larger than I expected (and having a styler that semi randomly signals MNU while editing is not fun)


MarcusDenker:
There is now a merge conflict

privat:
I did a merge to solve a conflict likely introduced by Iceberg when it moved code around :(

MarcusDenker:
tests in RBCodeSnippetTest are failing (I think it uses the #content method)

MarcusDenker (approved):

Issue #12807, opened 2023-02-18
Executable comment syntax error leads to CI crash


jecisc:
I created a PR that added executable comments. But I did a syntax error in one of those and the CI build ended up crashing with really poor feedback. 

I have the impression that this is caused because there is a problem when launching the RuntimeSyntaxError. 

To see the problem, you can debug PharoDocCommentTest>>#testExpressionSyntaxError and inspect `node expression evaluate` to see the error it raises. 

The error have a bug when trying to print itself: `#errorMessage was sent to nil`

```
UndefinedObject(Object)>>doesNotUnderstand: #errorMessage
UndefinedObject>>doesNotUnderstand: #errorMessage
RuntimeSyntaxError>>errorMessage
RuntimeSyntaxError>>messageText
[ :stream | | mt |
		stream << self class name.
		(mt := self messageText) isEmptyOrNil
			ifFalse: [ stream << ': ' << mt ] ] in RuntimeSyntaxError(Exception)>>defaultDescription in Block: [ :stream | | mt |...
String class(SequenceableCollection class)>>new:streamContents:
String class(SequenceableCollection class)>>streamContents:
RuntimeSyntaxError(Exception)>>defaultDescription
RuntimeSyntaxError(Exception)>>description
RuntimeSyntaxError(Exception)>>printOn:
[:s | self printOn: s] in RuntimeSyntaxError(Object)>>printStringLimitedTo: in Block: [:s | self printOn: s]
String class(SequenceableCollection class)>>streamContents:limitedTo:
RuntimeSyntaxError(Object)>>printStringLimitedTo:using:
RuntimeSyntaxError(Object)>>printStringLimitedTo:
RuntimeSyntaxError(Object)>>printString
MorphicUIManager(UIManager)>>requestDebuggerOpeningFor:
RuntimeSyntaxError(Exception)>>debug
MorphicUIManager>>unhandledErrorDefaultAction:
RuntimeSyntaxError(Exception)>>unhandledErrorAction
UnhandledError>>defaultAction
UndefinedObject>>handleSignal:
Context>>handleSignal:
Context>>handleSignal:
UnhandledError(Exception)>>signal
UnhandledError class>>signalForException:
RuntimeSyntaxError(Exception)>>raiseUnhandledError
RuntimeSyntaxError(Exception)>>defaultAction
UndefinedObject>>handleSignal:
RuntimeSyntaxError(Exception)>>pass
[ :ex |
			ActiveWorld := priorWorld.
[...]

PR #12857, opened 2023-02-24
Improve scanning: escape comments

### LINKED ISSUE #12844: Comments should be unscaped in the AST
I fell that the escaping of the comment contents is not done at the right place.

The earliest place is `RBScanner>>#scanComment` and we can see that escaping is managed here, but that both `"` are nextPut in the contents of the buffer. I think that's a mistake and only one should be put.

I'm not sure how many clients of the AST expect that comments are kept escaped but I think that most need what the programmer means (not the verbatim source text) — maybe only few programmers know that `"` can be escaped in comments.

_Originally posted by @privat in https://github.com/pharo-project/pharo/issues/12830#issuecomment-1440643339_

privat:
Few people know (I even did not know that last week) but in pharo you can escape double-quotes in comments by doubling the double-quote. (there are only a few uses of this syntactic feature in the pharo base code)

```st
"A comment with a "" inside"
```

For some reason (?), the pharo scanner did not unescape the double double-quote and let clients deal with possibly unexpected content with superfluous `"` (because when people comment `"a""b"` what they really mean is `a"b`).
One of these client was the executable comments (see #12807) where nested comment break the whole thing, so the PR tries to fix the root cause.

* Teach the scanner to unescape the comment contents when creating the comment token.
* Fix the highlighting in executable comments with nested comments.
* Teach EFFormatter to put back double `"` when formatting comments.

fix #12844
related to #12807 (half fix, the CI issue is still problematic)
replace #12830

### DIFF
--- a/src/AST-Core-Tests/RBErrorNodeParserTest.class.st
+++ b/src/AST-Core-Tests/RBErrorNodeParserTest.class.st
@@ -170,32 +170,6 @@ RBErrorNodeParserTest >> testFaultyCascade [
 ]
 
 { #category : #tests }
-RBErrorNodeParserTest >> testFaultyLiterabByteArray [
-
-	| source node |
-	source := '#[ 1 hello 3]'.
-	node := self parserClass parseFaultyExpression: source.
-
-	self assert: node isFaulty.
-	self assert: node isLiteralByteArrayError.
-
-	self deny: node content first isFaulty.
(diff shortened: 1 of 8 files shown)


svenvc:
LGTM (not actually tested)

privat:
Tests all green (except windows that is kaput)

guillep (approved):
Thanks!

PR #12863, opened 2023-02-26
Fix #10929 Graphic UI selection glitch with selected text on large screens

### LINKED ISSUE #10929: Graphic UI selection glitch with selected text on large screens
Hi,
I have a large screen (5120x1440) and I display the Pharo Window on all the available surface.
This is great but I have a funny bug :D when I select a text (from a source-code, or anything) the selection rectangle render is not visible.

See some screenshots after.

My Pharo desktop : 
image

And the selection bug in differents cases :
DirtySelection

On this gif we can see that there is "an invisible line", at the left off it the render is visible and at the right the render is cutted. May be this is the same problem for the vertical axe.

This is a problem when my selection is completely at the left of this line, I don't see the selection :S 
InvisibleSelection

This invisible line appear to be at 4095 pixels of the beginning of my screen (with a measure app), is the render clipped outside 4096 pixels ?

My config : 
Pharo 10.0.0
Build information: Pharo-10.0.0+build.399.sha.c4426366e83c26b8a002ef0680abe3a89a0b1a99 (64 Bit)
On Windows 10 Pro 64bits.

labordep:
My reflexion in the concerned thread here : #10929

### DIFF
--- a/src/FormCanvas-Core/BalloonEngine.class.st
+++ b/src/FormCanvas-Core/BalloonEngine.class.st
@@ -426,7 +426,11 @@ BalloonEngine >> flush [
 BalloonEngine >> initialize [
 	| w |
 	super initialize.
-	w := Display width > 2048 ifTrue: [ 4096 ] ifFalse: [ 2048 ].
+	
+	"Get the display width to initialize width of the bitmap, if the width is zero or negative, use a standard value as 2048"
+	w := Display width.
+	w <= 0 ifTrue:[w := 2048].
+	
 	externals := OrderedCollection new: 100.
 	span := Bitmap new: w.
 	self
@@ -750,4 +754,10 @@ BalloonEngine >> resetIfNeeded [
 	self primSetColorTransform: colorTransform.
 	self primSetDepth: self primGetDepth + 1.
 	postFlushNeeded := false
+]
+
+{ #category : #accessing }
+BalloonEngine >> span [
+
+	^ span
 ]
--- a/src/FormCanvas-Core/BalloonEngine.class.st
+++ b/src/FormCanvas-Core/BalloonEngine.class.st
@@ -431,22 +431,29 @@ BalloonEngine >> flush [
 
 { #category : #initialization }
 BalloonEngine >> initialize [
-	| w |
+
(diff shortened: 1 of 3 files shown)


MarcusDenker:
- [CI] Windows testrunner has problems #12763

src/FormCanvas-Core/BalloonEngine.class.st
```diff
@@ -426,7 +426,11 @@ BalloonEngine >> flush [
 BalloonEngine >> initialize [
 	| w |
 	super initialize.
-	w := Display width > 2048 ifTrue: [ 4096 ] ifFalse: [ 2048 ].
+	
+	"Get the display width to initialize width of the bitmap, if the width is zero or negative, use a standard value as 2048"
+	w := Display width.
+	w <= 0 ifTrue:[w := 2048].
```
demarey: 2048 is magic number.
It should be nice to extract it in a method like `defaultBitmapWidth`.
And also, renaming `w`to `width` would be nice too.

labordep: thanks @demarey, I will do that :)

src/FormCanvas-Core/BalloonEngine.class.st
```diff
@@ -50,6 +50,13 @@ BalloonEngine class >> debug: aBoolean [
 	Debug := aBoolean
 ]
 
+{ #category : #accessing }
+BalloonEngine class >> defaultBitmapWidth [
+	"Return default bitmap width"
```
demarey: The comment is a paraphrase of the selector. I would remove it.

demarey (approved):
Thank you for the changes and the contribution.
I suggest a last change but is not mandatory for the inclusion of this PR.

src/FormCanvas-Core/BalloonEngine.class.st
jecisc: I agree with Christophe

labordep: Ok ;-)
I will remove that.

jecisc (approved):

PR #12866, opened 2023-02-27
Deprecating documentPath.

### LINKED ISSUE #10007: VirtualMachine>>#documentPath is wrong

guillep:
Fix #10007


jecisc:
There is a failure in the CI:
```
Command line handler failed

Invalid arguments: "/builds/workspace/est_and_branch_Pipeline_PR-12866/64/bootstrap-cache/Pharo11.0-PR-compiler-64bit-355bec7.image loadHermes Hermes-Extensions.hermes --save"

script returned exit code 1
```

I’ve never seen this error so it might be related

MarcusDenker:
```
Command line handler failed
Invalid arguments: "/builds/workspace/est_and_branch_Pipeline_PR-12866/64/bootstrap-cache/Pharo11.0-PR-compiler-64bit-bea2264.image loadHermes Hermes-Extensions.hermes --save"
script returned exit code 
```

hernanmd:
The problem here is that the command line handler initialize its `arguments` i.v. as follows:

```smalltalk
initialize

	| documentPath  |

	arguments := OrderedCollection withAll: Smalltalk arguments.
	documentPath := Smalltalk vm documentPath.
	documentPath isEmptyOrNil
		ifFalse: [ arguments addFirst: documentPath ]
```

Then when in CLI we pass a command such as "test", that is what "documentPath" answer (it is just `getSystemAttribute: 2`) for instance: 

```bash
./pharo-ui Pharo.image test --junit-xml-output --stage-name=Tests-osx-64 '.*'
```

So it seems to be a naming problem which we could resolve by renaming the selector `documentPath` to something meaningful such as `commandName` or even more generic `vmFirstParameter`.

MarcusDenker:
superceeded by https://github.com/pharo-project/pharo/pull/15577

PR #12869, opened 2023-02-27
12858-CopiedLocalVariable-answers-false-to-isTemporaryVariable-even-when-it-is 

### LINKED ISSUE #12858: CopiedLocalVariable answers false to #isTemporaryVariable even when it is
I have a project manipulating the AST and porting it to more recent versions of Pharo it is broken. The reason is that some temporary variables are now CopiedLocalVariable in the AST and it answers false to #isTemporaryVariable. 

I think that when they represent a temporary variable, they should answer true.

MarcusDenker:
- change superclass of CopiedLocalVariable to TemporaryVariable 
- add a test by enhancing testOptimizedBlockReadInBlock

(did not rename it to CopiedTempVariable, we could think about that)

fixes #12858

### DIFF
--- a/src/OpalCompiler-Core/CopiedLocalVariable.class.st
+++ b/src/OpalCompiler-Core/CopiedLocalVariable.class.st
@@ -3,7 +3,7 @@ A copied variable is an arg or temp var that is copied into a block that later r
 "
 Class {
 	#name : #CopiedLocalVariable,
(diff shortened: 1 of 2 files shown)


jecisc (approved):

jecisc:
CI is good except windows

PR #12868, opened 2023-02-27
Improve faulty parsing on variables

privat:
Some work on the parsing of variable (temps and block arguments).

Most work is aimed at removing FIXME in `RBCodeSnippet class >> #badExpressions`

Main changes are:
* Bad temps and blocks args are preserved in the faulty ast and in formatted code
* `BRParser>>#nextBar` to simplify the code (and make it less random) when dealing with `|` as separator
* unexpected block parameters declaration `:foo` are treated like a bad variable.

See individual commits for details.

PR #12874, opened 2023-02-27
Rewrite some error messages to follow the style `Foo expected`

privat:
The format `Foo expected` (with an uppercase, no article and no verb) is used on almost all error messages. This PR changes some noncompliant ones.

Also, on `| expected`, `varable` or `parameter` are added, because they are also expected, and because it helps to understand the context.

Note: this PR might conflict with #12868, so merge the other before.
You might consider reviewing this one in the meantime.


jecisc (approved):
Lgtm

MarcusDenker (approved):

PR #12879, opened 2023-02-28
Improve faulty parsing of methods

privat:
This PR is a little larger than the previous one but it's mainly small commits.

* Extends RBCodeSnipped with full methods (including broken pattern, arguments and pragmas).
* Fix various clients of faulty pragma nodes (not that numerous in fact).

Note: I do not change the compiler behavior when dealing with faulty AST. Especially, no RuntimeSyntaxicError is signaled is the method contains buggy pragmas or parameters. But maybe it's not the job of the compiler to do such policy decision about what faulty code is acceptable to execute or not?


privat:
Merge with P11 to solve conflict with #12868 (I initially thought that both PR were not conflictual)

MarcusDenker (approved):

PR #12882, opened 2023-02-28
Use MethodClassifier protocol for new methods instead of the selected protocol

demarey:
It will avoid critics and broken release tests.
If there is no protocol found by MethodClassifier, then the selected protocol is used as fallback


demarey:
done with @jecisc

demarey:
tests are green. Windows tests step reached a timeout

MarcusDenker (approved):

PR #12887, opened 2023-03-01
Skipping test in windows as it is hanging

tesonep:
The test is hanging in windows, we don't know why... but should be skipped 😮


jecisc:
Maybe @svenvc has an idea?

MarcusDenker:
Windows is green!

This one fails on linux only, which is odd but unrelated:
CompletionContextTest>>testReceiverGlobalVarNew

PR #12888, opened 2023-03-01
Add tests on executable comments

jecisc:
My PR https://github.com/pharo-project/pharo/pull/12830 was not integrated because a better fix was proposed. This on is only harvesting the tests that I added to ensure no regression happens in the future

### DIFF
--- a/src/PharoDocComment-Tests/PharoDocCommentTest.class.st
+++ b/src/PharoDocComment-Tests/PharoDocCommentTest.class.st
@@ -7,6 +7,18 @@ Class {
 	#category : #'PharoDocComment-Tests-Base'
 }
 
+{ #category : #helpers }
+PharoDocCommentTest class >> stringWithDoubleDoubleQuotes [
+
+	^ 'This is ""exagerated"" at least'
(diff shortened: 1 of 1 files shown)


privat:
the integrated PR #12857 did contain tests on comments in executable comments, one even has nested executable comment, because Yo Dawg, I herd you like executable comments, so I put an executable comment in your executable comment so you can execute comments while you execute comments.

https://github.com/pharo-project/pharo/pull/12857/files#diff-6463bd3741296f2d964a968076387c493c1a11b115a07fe99bf72ba57952e13c

<blockquote><div> GitHub</div><div><strong><a href="https://github.com/pharo-project/pharo/pull/12857">Improve scanning: escape comments by privat · Pull Request #12857 · pharo-project/pharo</a></strong></div><div>Few people know (I even did not know that last week) but in pharo you can escape double-quotes in comments by doubling the double-quote. (there are only a few uses of this syntactic feature in the ...</div></blockquote>

jecisc:
Oh good! Thank you :)

PR #12890, opened 2023-03-01
Microdown-RichTextComposer is now managed outside

### LINKED ISSUE #12889: Pharo11 clean image has dirty packages
If you get a clean image of Pharo 11 and recalculate the dirty packages, we can see that the image is dirty because Microdown-RichTextComposer is not totally loaded. 

It’s probable that it is loaded too early in the loading process.

jecisc:
We can remove it from Pharo. It is already loaded from the microdown repository.
Fixes #12889


jecisc:
Ci seems to have a random failure

MarcusDenker (approved):

Issue #12502, opened 2023-02-01
[CI] test runs sometimes aborted due to SubscriptOutOfBounds error


MarcusDenker:
**Bug description**
Sometimes (not often) the test run is aborted due to a call to 

 SubscriptOutOfBounds  class>>signalFor:lowerBound:upperBound:in:

This seem to happen during drawing, leading to the emergency debugger opening which logs and quits.

```


a DebugSession
Original error: test session.
Debugger error: a UndefinedObject

[ Set new ] in OupsDummyDebugger class>>dummySession in Block: [ Set new ]

a DebugSession
Original error: test session.
Debugger error: Error

[ Set new ] in OupsDummyDebugger class>>dummySession in Block: [ Set new ]

a DebugSession
Original error: test session.
Debugger error: Error

[ Set new ] in OupsDummyDebugger class>>dummySession in Block: [ Set new ]
SubscriptOutOfBounds: 1
[:err |
		"Handle a drawing error"
		| errCtx errMorph |
		errCtx := thisContext.
		[
			errCtx := errCtx sender.
			"Search the sender chain to find the morph causing the problem"
			[errCtx notNil and:[(errCtx receiver isMorph) not]] 
				whileTrue:[errCtx := errCtx sender].
			"If we're at the root of the context chain then we have a fatal drawing problem"
			errCtx ifNil:[^self handleFatalDrawingError: err description].
			errMorph := errCtx receiver.
			"If the morph causing the problem has already the #drawError flag set,
			then search for the next morph above in the caller chain."
			errMorph hasProperty: #errorOnDraw
		] whileTrue.
		errMorph setProperty: #errorOnDraw toValue: true.
		"Install the old error handler, so we can re-raise the error"
		err signal.
	] in WorldState>>displayWorldSafely: in Block: [:err |...
FullBlockClosure(BlockClosure)>>cull:
[:ex | errorHandlerBlock cull: ex] in FullBlockClosure(BlockClosure)>>onErrorDo: in Block: [:ex | errorHandlerBlock cull: ex]
FullBlockClosure(BlockClosure)>>cull:
Context>>evaluateSignal:
Context>>handleSignal:
SubscriptOutOfBounds(Exception)>>pass
[ :ex | 
			"This should do it unless you don't screw up the bounds"
			fullBounds := bounds.
			ex pass ] in SpWindow(Morph)>>computeFullBounds in Block: [ :ex | ...
FullBlockClosure(BlockClosure)>>cull:
[...]

PR #12892, opened 2023-03-02
Improve faulty scanning on comments

privat:
Comment was the only (remaining) scanned token to possibly signal a syntax error notification (and possibly launch a recovery editor in the UI) instead of producing an error token. This PR change that.

The issue was that comments are stripped from the scanner flow, but AFTER the next token is scanned (so that we can attach the comments on the token returned by `next`).
~~The proposal is if one of the comment is a bad comment (the only current case is an unterminated comment), we have to rewind the stream position (and the scanned state) just before the faulty token so that the next `next` will scan and return this faulty token.~~
The proposal is if one of the comment is a bad comment (the only current case is an unterminated comment), we just remember it, so it will be the next token (returned by `next`).

Note that EFFormatter have some issues with comments, so tests are let as it but will catch on future improvements.

### DIFF
--- a/src/AST-Core-Tests/RBCodeSnippet.class.st
+++ b/src/AST-Core-Tests/RBCodeSnippet.class.st
@@ -49,6 +49,18 @@ RBCodeSnippet class >> badExpressions [
 		self new source: '#'.
 		self new source: '$'.
 		self new source: ':'.
+		self new source: ''; isFaulty: false. "emptyness is ok"
+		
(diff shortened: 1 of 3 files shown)


privat:
I did not like the strategy used, so I couldn't sleep, so I changed it... now to bed!

MarcusDenker:

MarcusDenker (approved):

Issue #12883, opened 2023-02-28
[Meta-issue] improve faulty compiling


privat:
Parsing and compiling is an everyday job and is performed when you install a package, switch branch, compile a method or add a character in a method editor (because of highlighting).

AST-Code (and its client) and OpalCompiler have/had a limited approach when dealing with faulty code.

The historical approach was basically to fire the UI in case of code error, so the human can fix it (in a not-so interactive way).
This approach is somewhat limited if tools have to work during code edition (or on a random piece of code) to provide better user feedback.

Pharo included basic opt-in faulty parsing (and compiler related options) so AST clients can try to deal with faulty code (highlighting for instance). This Issue track improvement and discussions.

## Bad API

The main issue of the current API are:

* [ ] a lack of isolation with UI: magic requestor instance variable, compiler drives the UI presentation on syntax error or semantic error (with interactive menu)
  → mostly eliminated. Some requestor thing remains, but I wait for open season in spec/newtools repositories
* [x] a bloated code recovery: notification ping-pong (SyntaxErrorNotification, OCUndeclaredVariableWarning, ReparseAfterSourceEditing).
  → Now fixed
* [x] complex effects on the configuration: requestor and its interactiveness, optionParseErrors, optionParseErrorsNoInteractive. optionSkipSemanticWarnings
  → Now fixed
* [x] lack of robustness: fragile `errorblock` and `failBlock`, unclear exception signaler, unclear notification signaled and possibly resumed
  → Now fixed

## Current debugger GUI on faulty code

Some notes on error recovery. i.e. what happens when a faulty node is executed.

### Evaluate expression (DoIt method)

```st
Smalltalk compiler permitFaulty: true; evaluate: '1+¿'.
```

* [x] launch a debugger with `RuntimeSyntaxError: Unknown character`
* [x] sane stack trace (nil>>DoIt, OC>>evaluateDoIt, OC>>evaluate, OC>>evaluate:)
* [x] present original code
* [x] visible cursor and error located in the code
* [x] recovery. no way to recover from here, cannot modify a doit method, but it is a limitation of the debugger, not ours, so I add a checkmark

### Execute non installed method

```st
nil executeMethod: (Smalltalk compiler permitFaulty: true; compile: 'foo ^ 1+¿')
```

* [x] launch a debugger with `RuntimeSyntaxError: Unknown character`
* [x] sane stack trace (nil>>foo, nil>>executeMethod:)
* [x] present original code.
[...]

PR #12894, opened 2023-03-03
12807-Executable-comment-syntax-error-leads-to-CI-crash 

### LINKED ISSUE #12807: Executable comment syntax error leads to CI crash

MarcusDenker:
This is a fix for better error messages for Doc comments.

- add a way to test for not well formed doc comments independ if there is a syntax error (#isWellFormed)
- in #evaluate we check that and raise and error, the we recompile and let the compiler give us good error messages for the expression that we compile

An improvement that could be made as a next step if to add a specific exception for the doc commen error and describe how the expression is not well formed.

fixes #12807

### DIFF
--- a/src/PharoDocComment-Tests/PharoDocCommentTest.class.st
+++ b/src/PharoDocComment-Tests/PharoDocCommentTest.class.st
@@ -36,9 +36,9 @@ PharoDocCommentTest >> testExpressionNoAssociation [
 	"(3 + 4 >>> 7) value"
 	| node |
 	node := thisContext method ast pharoDocCommentNodes first.
-	self assert: node expression isFaulty.
-	self should: [ node expression evaluate ] raise: RuntimeSyntaxError.
-	self should: [ (CommentTestCase for: node) testIt ] raise: RuntimeSyntaxError
+	self deny: node expression isWellFormed.
+	self should: [ node expression evaluate ] raise: Error.
+	self should: [ (CommentTestCase for: node) testIt ] raise: Error
 ]
 
 { #category : #tests }
@@ -82,9 +82,12 @@ PharoDocCommentTest >> testExpressionSyntaxError [
 	"3 + 4) >>> 7"
 	| node |
 	node := thisContext method ast pharoDocCommentNodes first.
+	"We get a syntax error"
 	self assert: node expression isFaulty.
-	self should: [ node expression evaluate ] raise: RuntimeSyntaxError.
-	self should: [ (CommentTestCase for: node) testIt ] raise: RuntimeSyntaxError
+	"but the structure of the >>> looks ok"
+	self assert: node expression isWellFormed.
+	self should: [ node expression evaluate ] raise: SyntaxErrorNotification.
+	self should: [ (CommentTestCase for: node) testIt ] raise: SyntaxErrorNotification
 ]
 
 { #category : #tests }
@@ -96,8 +99,8 @@ PharoDocCommentTest >> testExpressionZero [
 	| node |
 	node := thisContext method ast pharoDocCommentNodes first.
 	self assert: node expression isFaulty.
-	self should: [ node expression evaluate ] raise: RuntimeSyntaxError.
-	self should: [ (CommentTestCase for: node) testIt ] raise: RuntimeSyntaxError
+	self should: [ node expression evaluate ] raise: Error.
+	self should: [ (CommentTestCase for: node) testIt ] raise: Error
 ]
 
 { #category : #tests }
@@ -118,8 +121,8 @@ PharoDocCommentTest >> testMultipleDocCommentsInOneComment [
 	| node |
 	node := thisContext method ast pharoDocCommentNodes first.
 	self assert: node expression isFaulty.
-	self should: [ node expression evaluate ] raise: RuntimeSyntaxError.
-	self should: [ (CommentTestCase for: node) testIt ] raise: RuntimeSyntaxError
+	self should: [ node expression evaluate ] raise: Error.
+	self should: [ (CommentTestCase for: node) testIt ] raise: Error
 ]
 
 { #category : #tests }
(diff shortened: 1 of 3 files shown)


privat:
I'm not fan of having a new vocabulary `isWellFormed`. I would prefer to keep `isFaulty` on executable comment meaning that the syntactic form is not good.

src/PharoDocComment-Tests/PharoDocCommentTest.class.st
```diff
@@ -96,8 +99,8 @@ PharoDocCommentTest >> testExpressionZero [
 	| node |
 	node := thisContext method ast pharoDocCommentNodes first.
 	self assert: node expression isFaulty.
-	self should: [ node expression evaluate ] raise: RuntimeSyntaxError.
-	self should: [ (CommentTestCase for: node) testIt ] raise: RuntimeSyntaxError
+	self should: [ node expression evaluate ] raise: Error.
```
privat, line 102: Error is too generic

privat:
By the way, I'm working on semantic errors (cf #12883) and I evaluate the possibility to extends `isFaulty` to include the presence of semantic errors.

src/PharoDocComment-Tests/PharoDocCommentTest.class.st
MarcusDenker: I will add a dedicated Exception in a next PR

MarcusDenker:
> I'm not fan of having a new vocabulary `isWellFormed`. I would prefer to keep `isFaulty` on executable comment meaning that the syntactic form is not good.

The whole problem comes from using "isFaulty" for both and raising an error when the doc comment is not well formed  or a syntactically wrong expression.

If we want better error messages, we have to distinguish the two. There is no way to have better error messages by checking for #isFaulty only, as the whole expression this is true but we have no idea where.

One fesure of #isFaulty is that an AST like that can be executed and raises the error at runtime (if it executes a code path that has it). While here the whole doc comment just says "no, I can not be executed because either the doc comment is not well formed or there is a syntax error somewhere". 

For both we can give better infos statically, but we have to distingush the two cases.

MarcusDenker:
I Propose to merge this as a workaround now and then use this as one of the places in the system that can be improved by extending the isFaulty to cover semantic errors.

privat:
You can merge, I added a todo in #12883 to remember about executable comments

PR #12895, opened 2023-03-03
InitializedSlot-check-default 

MarcusDenker:
fixes #12819

When using InitializedSlot, it is tempting to use "OrderedCollection new" as the default. But that is not a good idea, as this instance would be created at slot definition time.

A solution is to restrict the the default value to be a literal (read only) or a clean block (evaluated each time the class is instantiated).

When creating slots via evaluation, we can just raise an error. The class parser then can show a nice user level error message when parsing the class definition. 

```
testPrintOn

 	| slot |
 	slot := #slot1 => InitializedSlot default: [ OrderedCollection new ].
	self
		assert: slot definitionString
		equals: '#slot1 => InitializedSlot default: [ OrderedCollection new ]'.

	slot := #slot1 => InitializedSlot default: #literal.
	self
		assert: slot definitionString
		equals: '#slot1 => InitializedSlot default: #literal'.

	"On eval, we raise an error, the ClassParser can show a nicer error in the class definition"
	self should: [#slot1 => InitializedSlot default: 5@2] raise: Error.
	self should: [#slot1 => InitializedSlot default: OrderedCollection new] raise: Error
```

### DIFF
--- a/src/VariablesLibrary-Tests/InitializedSlotTest.class.st
+++ b/src/VariablesLibrary-Tests/InitializedSlotTest.class.st
@@ -5,6 +5,33 @@ Class {
 }
 
 { #category : #tests }
+InitializedSlotTest >> testBlockWithNonSharedDefault [
+ 	"If we use a block the resulting default value should not be shared among instances"
+
+ 	| slot objectA objectB defaultA defaultB |
+ 	slot := #slot1 => InitializedSlot default: [ OrderedCollection new ].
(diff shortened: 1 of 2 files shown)


MarcusDenker:
I propose to merge this as a first step (so people do not run into the same idea of using "OderedCollection new" as a default value).

I added to my backlog
- improve error messages (maybe have a dedicated exception)
- check the ClassParser and explore how to handle this case there (for user feedback)

PR #12897, opened 2023-03-04
Add new attributes to SDL_MouseWheelEvent

### LINKED ISSUE #12896: SDL_MouseWheelEvent has 2 new attributes since SDL 2.0.18
Current Pharo 11 VM comes with a newer version (*) which counts with 2 new attributes: preciseX and preciseY. They are ignored by current FFI bindings.

type | name | description 
-- | -- | --
float | preciseX | the amount scrolled horizontally, positive to the right and negative to the left, with float precision (added in 2.0.18)
float | preciseY | the amount scrolled vertically, positive away from the user and negative toward the user, with float precision (added in 2.0.18)

(Source: https://wiki.libsdl.org/SDL2/SDL_MouseWheelEvent)

(*): vm has SDL 2.24.1, BTW they changed version notation in the middle.

**Describe the request**
I will do a PR to add these 2 attributes, which is inoffensive: the "low-level" SDL2 event is converted to a OSWindow event in `OSSDL2BackendWindow>>#visitMouseWheelEvent:`, and I'd let it untouched for the moment.

**Expected behavior**
Preliminar test showed that these values can be useful for smooth scrolling.

tinchodias:
Proposed to fix #12896.
In addition, also included binding to 'direction' attribute, added in SDL 2.0.4.

Post load action: `SDL_MouseWheelEvent rebuildFieldAccessors`

To test it:
1. Create a `printOn:` in `SDL_MouseWheelEvent` (see below)
2. In `OSSDL2Driver>>#processEvent:`, add the expression `mappedEvent traceCr.` right after `mappedEvent := sdlEvent mapped.`.
3. Open Transcript and scroll with the wheel or typical two-fingers in track pad.

Example of printOn: to see the new attributes:
```
printOn: stream 

	super printOn: stream.
	
	stream print: (self preciseX @ self preciseY).
	
	stream print: self direction.
```

### DIFF
--- a/src/OSWindow-SDL2/SDL_MouseWheelDirection.class.st
+++ b/src/OSWindow-SDL2/SDL_MouseWheelDirection.class.st
@@ -1,0 +1,54 @@
+"
+I represent the scroll direction types for the Scroll event.
+
+I'm a binding to the following enum in SDL2 library:
+
+```
+typedef enum
+{
+    SDL_MOUSEWHEEL_NORMAL,    /**< The scroll direction is normal */
+    SDL_MOUSEWHEEL_FLIPPED    /**< The scroll direction is flipped / natural */
+} SDL_MouseWheelDirection;
+```
+
+
+"
+Class {
+	#name : #'SDL_MouseWheelDirection',
+	#superclass : #FFIEnumeration,
+	#classVars : [
+		'SDL_MOUSEWHEEL_FLIPPED',
+		'SDL_MOUSEWHEEL_NORMAL'
+	],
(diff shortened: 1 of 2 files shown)


MarcusDenker:
faling test related (simple fix needed)

- ProperMethodCategorizationTest>>testClassSideInitializeMethodNeedsToBeInClassInitializationProtocol


failing test not related:

tinchodias:
Thanks Marcus

MarcusDenker:

jecisc (approved):

PR #12900, opened 2023-03-05
no more RBScanner errorBlock

privat:
PR #12892 removed the last usage of errorBlock in RBScanner.

This PR does the **almost** final cleanup of now unused code.
Because existing images still refer to some undead method of RBScanner (`on:errorBlock:` and `errorBlock:`) they are kept as empty shells.
Note: for some reason, moving these two methods as extension methods in Deprecated11 also break Iceberg. Possible the "move" is not atomic so the removal of the method in AST-Core is done before it is added in Deprecated11 letting the low level code experiences MNU errors.

### DIFF
--- a/src/AST-Core/RBParser.class.st
+++ b/src/AST-Core/RBParser.class.st
@@ -258,7 +258,7 @@ RBParser >> initialize [
 { #category : #accessing }
 RBParser >> initializeParserWith: aString [
 	source := aString.
-	self scanner: (self scannerClass on: (ReadStream on: aString) errorBlock: self errorBlock)
+	self scanner: (self scannerClass on: (ReadStream on: aString))
 ]
 
 { #category : #'private - classes' }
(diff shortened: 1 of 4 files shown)


guillep (approved):
Good!

MarcusDenker:
Failing test that maybe realted:

BCBeautifulCommentsSettingsTest>>testRenderingOfMyOwnComment


CI problem not related:

privat:
Hum... I cannot check out the branch anymore :/ all I got is a weird nonsensical emergency alert popup.

Capture d’écran du 2023-03-06 09-01-23

Maybe moving around low level code broke some Iceberg expectations.

privat:
Moved back old code as No-op to ease the loading :(

My assumption is that old compiled methods in fresh image are still using the killed API, but the RBScanner methods are moved/removed before such client method are updated (and recompiled).
Future images should not contain call to the deprecated method of RBScanner, so the removal can be complete at that time.

Now I can reproduce the BCBeautifulCommentsSettingsTest issue, I will investigate.

MarcusDenker:
Yes, I think so... code loading is not atomic sadly

privat:
I think it was a syntactic error in the depreciation clause (so the method just failed).

MarcusDenker (approved):

PR #12904, opened 2023-03-06
new class SycInspectNodeCommand to simply inspect an AST node

privat:
When hacking on AST elements, it's nice to be able to inspect a given AST node of some code shown in an editor.

Capture d’écran du 2023-03-06 08-36-41


MarcusDenker:
ZnHTTPSTest>>testGmailEncrypted timeout, mac only, not related

MarcusDenker (approved):

PR #12905, opened 2023-03-06
Backport fix to bugged undo-redo commands after using completion

carolahp:
Changing the Completion Engine behavior from replacing the full user typed word with its completed version, to insert only the missing part of the word.


MarcusDenker:
failing tests not related

PR #12907, opened 2023-03-06
Ensure #setUp is classified in running

jecisc:
We currently have a release test that #setUp should be classified in #running but the current MethodClassifier put it in initialization. This update the behavior by adding a list of methods we want to see in specific protocols + add a test

### DIFF
--- a/src/Tool-Base/MethodClassifier.class.st
+++ b/src/Tool-Base/MethodClassifier.class.st
@@ -13,10 +13,20 @@ Class {
 	#classInstVars : [
 		'prefixMapping',
 		'keywordSuffixMapping',
-		'pragmaMapping'
+		'pragmaMapping',
+		'exactMapping'
 	],
 	#category : #'Tool-Base-Utilities'
 }
+
+{ #category : #private }
+MethodClassifier class >> buildExactMapping [
+
+	exactMapping := Dictionary new
+		                at: 'setUp' put: 'running';
+		                at: 'tearDown' put: 'running';
+		                yourself
+]
 
 { #category : #private }
 MethodClassifier class >> buildKeywordSuffixMapping [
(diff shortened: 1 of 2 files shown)


MarcusDenker (approved):

src/Tool-Base/MethodClassifier.class.st
```diff
@@ -49,53 +59,54 @@ MethodClassifier class >> buildPragmaMapping [
...
+		                 at: 'requires' put: 'testing';
+		                 at: 'reset' put: 'initialization';
+		                 at: 'set' put: 'initialization';
+		                 at: 'should' put: 'asserting';
+		                 at: 'shouldnt' put: 'asserting';
+		                 at: 'signal' put: 'signalling';
+		                 at: 'sort' put: 'sorting';
+		                 at: 'test' put: 'tests';
+		                 at: 'total' put: 'accessing';
+		                 at: 'version' put: 'versions';
```
astares, line 106: @jecisc why not "versioning" instead of "versions" if we follow the Someth**ing** pattern

jecisc: This was already here, I just changed the formatting. It's possible there are more things to improve. I just wanted to align the release tests and the method classifier

astares:
@jecisc And I do not understand: "setUp is in running and MethodClassifier put it in running too" what the issue is?

Is it about symbol (#running) vs. string ("running") or what?

jecisc:
We have a release test that says that "setUp" method should be in "running" protocol. 

But the method categorizer of Pharo was putting "setUp" in the "initialization" protocol by default because the prefix map says that things starting by "set" are going in the initialization protocol

jecisc:
Oh, I see that I wrote too fast in my original message. It should be: 

We currently have a release test that #setUp should be classified in #running but the current MethodClassifier put it in initialization

Sorry for that

astares:
@jecisc Ah - OK. Thanks for the clarification!

PR #12908, opened 2023-03-06
Add BaselineOf:>>#pharoVersionsFrom:

jecisc:
Add an helper when we want to sepcify in a configuration that we want a package to be loaded in Pharo X and later versions.

### DIFF
--- a/src/Metacello-Base/BaselineOf.class.st
+++ b/src/Metacello-Base/BaselineOf.class.st
@@ -106,6 +106,12 @@ BaselineOf >> baseline: spec [
 	"
 ]
 
+{ #category : #baselines }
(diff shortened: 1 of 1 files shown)


jecisc:
To use like this: https://github.com/jecisc/pharo-scripts/blob/master/settings/BaselineOfPharoSettings/BaselineOfPharoSettings.class.st

<blockquote><div> GitHub</div><div><strong><a href="https://github.com/jecisc/pharo-scripts">pharo-scripts/BaselineOfPharoSettings.class.st at master · jecisc/pharo-scripts</a></strong></div><div>My personal settings for my Pharo images. Contribute to jecisc/pharo-scripts development by creating an account on GitHub.</div></blockquote>

MarcusDenker (approved):

PR #12909, opened 2023-03-06
Improve faulty parsing. Add semantic snippets

privat:
This PR does 3 related things

* Improve and fortify the `parse`, `doSemanticAnalysis` and `compile` methods of RBCodeSnippet.
  * and improve the related tests.
  * add an `onError:` version to have a "non-faulty" mode that throws errors.
  * and add the related tests on `onError` version.
* Teach OCASTTranslator to emit runtime error when trying to assign read-only variable (instead of crashing or corrupting the image).
* Add new snippets related to semantic errors and warnings
  * Some are OK as is, others need fixmes and skips

These change feels the minimal set to have a bunch of new test that pass CI reasonably (with fixme and skips).

For the sake of simplicity (and reviewability). The following is let to future PR:

* Handle semantic error in the AST
* Test semantic warnings (currently, they are just ignored by RBCodeSnippetTest)
* Do something meaningful with backend error (they should be caught at semantic-time!).

### DIFF
--- a/src/AST-Core-Tests/RBCodeSnippet.class.st
+++ b/src/AST-Core-Tests/RBCodeSnippet.class.st
@@ -39,7 +39,10 @@ Class {
 { #category : #accessing }
 RBCodeSnippet class >> allSnippets [
 
-	^ self badExpressions, self badMethods
+	^ {
+		  self badExpressions.
+		  self badMethods.
+		  self badSemantic } flattened
 ]
 
 { #category : #accessing }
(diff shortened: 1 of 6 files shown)


Ducasse:
You look to have a lot of fun and this is super nice to see this :)

MarcusDenker:
failing test not related: TKTCommonQueueWorkerPoolTest>>testWorkerPoolDoesNotExceedPoolSizeWhenSchedulingTasksInParallel

MarcusDenker (approved):

PR #12912, opened 2023-03-06
Improve faulty parsing - simplify RBParser error API

privat:
Error handling in the whole compilation infrastructure of Pharo is sometime really weird, possibly for a lot of reasons (going from stacked historical compatibility hacks; to insane design choices that could be used as counter example and case study in a graduated software engineering course).

This PR tries to clean up (a little) the current status. Now that `RBScanner` is a little better (cf #12900) it is the turn of `RBParser`.

It is important to remember that `RBParser` is a simple *utility* component in the whole compilation infrastructure. Its main client is `OpalCompiler` that is a far more complex beast (unfortunately). Therefore:

* Providing the possibility of interactive recovery on syntax error is overengineering (and seems not really used by any client). The simpler used solution is to inform the client of errors and let it run a new parser.
* `RBParser` is not the place to have complex UI/UX/event-programming design. `OpalCompiler` and its clients, like the Rubric code editor, are (maybe) a better place to do fancy things.
* Syntax errors are a normal thing to expect, and most of the time, manipulating faulty AST is what one really want.

The main thing of the PR is to kill `errorBlock` and stop using `ReparseAfterSourceEditing`.

The proposed API is now:

* low level: a class toggle `beFaulty` (and instance variable `isFaulty`) to activate fault-tolerant parsing that produce AST that contains `RBParseErrorNode`, and otherwise signals `SyntaxErrorNotification`.
* helpers methods `parseFaulty*` are preserved (and still do not signal SyntaxError).
* helpers `Xxxx:OnError:` are also preserved, the behavior change a little so that the given block is used to intercept the syntax error signal in a compatible way. The only change is that there is no recovery (no client never cared about that anyway). The value of the block will be the value of the whole `Xxx:OnError:` call.

Note that bugs in the PR might break the usability of images by limiting the possibility to compile new code. So good reviewing is needed :)

### DIFF
--- a/src/AST-Core-Tests/RBFormatterTest.class.st
+++ b/src/AST-Core-Tests/RBFormatterTest.class.st
@@ -19,8 +19,7 @@ RBFormatterTest >> formatClass: aClass selector: aSymbol [
 	source := aClass sourceCodeAt: aSymbol.
 	tree1 := self parserClass parseMethod: source.
 	tree2 := self parserClass
-		parseMethod: (self formatterClass new format: tree1)
-		onError: [ :err :pos | self assert: false ].
+		parseFaultyMethod: (self formatterClass new format: tree1).
 	self assert: tree1 equals: tree2
 ]
 
--- a/src/AST-Core-Tests/RBParserTest.class.st
+++ b/src/AST-Core-Tests/RBParserTest.class.st
@@ -1441,13 +1441,11 @@ RBParserTest >> testParserErrorsWithErrorBlock [
 
 	#(#('self foo. + 3' 1) #('self 0' 1) #('self asdf;;asfd' 1))
 		do: [:each | |ast errorCount|
-			   [ ast := self parseError: each first onError:
-					[:msg :pos :parser | parser parseErrorNode: msg]]
-						on: ReparseAfterSourceEditing do: [].
+			ast := self parserClass parseFaultyExpression: each first.
 
-				errorCount := 0.
-				RBParseErrorNodeVisitor visit: ast do: [ :n | errorCount := errorCount + 1 ].
-				self assert: errorCount equals: each last ]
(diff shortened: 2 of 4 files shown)


MarcusDenker:
Bootrap failure:

```
    [+] Preparing bootstrapInstance of RGInstanceVariableSlot did not understand #allowsShadowing
RGInstanceVariableSlot(Object)>>doesNotUnderstand: #allowsShadowing
OCASTSemanticAnalyzer>>declareVariableNode:as:

```

privat:
The error is weird.

MarcusDenker:
Re-running is now triggers

- [CI] full image load is broken #12926

MarcusDenker:
Very strange. We get

```
RGInstanceVariableSlot(Object)>>doesNotUnderstand: #allowsShadowing
OCASTSemanticAnalyzer>>declareVariableNode:as:
OCASTSemanticAnalyzer>>declareTemporaryNode:
```

RGInstanceVariableSlot  is Ring2 RGInstanceVariableSlot instance, which models slots in Ring2 and should be API compatible to Slots).

But what is strange as RGInstanceVariableSlot actually does implement #allowsShadowing. 

Maybe the code that the boostrap uses is not up to date compared to the version used by Pharo11?

(I always wanted to push for including everything that the bootstap uses into the main image to make it more visible and easier to keep up to date... but that is another thing for another time)

privat:
Yes, this is really strange. Either the change introduced a subtle bug, or just uncover a potential one (or both).

Anyway, after thinking more about that, I'm not entirely sure about the PR. Maybe keeping an error block but with a failBlock semantic, ie the returned value is not used to recover compilation, but become the value of the high-level public parsing method called (the one that give the whole AST)

privat:
Better attempt proposed #13042 , so close this one

PR #12915, opened 2023-03-07
Improve styler

### LINKED ISSUE #12914: Shout attributes wrong attributes to some nodes
I noticed that shout is attributing wrong styles to some rb nodes. 

We cannot see it with the default theme because the different elements share the same color in those. The tests are also using the colors of the default theme, but if you create a custom theme, the problems are showing.

I found the places the errors are made.

I'll do some PRs to add a new tests infrastructure based on the attributes assigned and not the color. I'll also fixe the errors I found. Lastly I'll make shout delegate the resolution of variables styles to the AST instead of managing it in shout

### LINKED ISSUE #12925: Shout: Lot of styles are not assigned by shout anymore
Lot of styles in our style tables are not assigned anymore by shout. 

I'll remove them in #12914 

I'm just keeping this issue to track the things to do

jecisc:
This PR contains multiple changes:
- Introduction of an infrastructure to better test the styler that do not rely on the colors of the current theme
- Add tests on multiple elements to style (I'll add more soon)
- Fix bug where argument was detected as a temporary variable
- Fix bug where block pattern temp var was detected as a temporary variable
- Fix bug where shared pools vars were detected as class variables
- Fix bug where class var is considered a global for style
- Remove multiple style that are not used by Shout such as: excessCode, unfinishedComment, unfinishedString, binary, unary, $, -, stringSymbol, literalArray, methodArg, workspaceVar, incompleteBinary, incompleteUnary, undefinedBinary, undefinedUnary, arrayStart, arrayEnd, arrayStart1, arrayEnd1, externalCallType, externalCallTypePointerIndicator, primitiveOrExternalCallStart, primitiveOrExternalCallEnd, primitive, pragmaUnary, pragmaBinary, externalFunctionCallingConvention, module

Fixes #12925 
Fixes #12914


MarcusDenker:
- SpCodeAdapterTest fails, it seems to nedd #unary 
- NoUnusedVariablesLeftTest>>testNoUnusedClassVariablesLeft

MarcusDenker:
Full image load failure, but that might not be due to this PR?

```
MetacelloNotification: Loading -> Chest-tonel.1 --- github://pharo-spec/Chest:v0.4.2 --- cache
Error: Package Chest depends on the following classes:
  StPlaygroundInteractionModel
  StDebugger
  StDebuggerExtensionVisitor
  StDebuggerExtensionInspectorNodeBuilderVisitor
  ChestCodeScriptingInteractionModel
  ChestCodeScriptingInteractionModel>>#debugger:
  ChestCodeScriptingInteractionModel>>#addBinding:
  ChestCodeScriptingInteractionModel>>#addSelfAndThisContextBindings
  ChestCodeScriptingInteractionModel>>#bindingOf:
  ChestCodeScriptingInteractionModel>>#bindings
  ChestCodeScriptingInteractionModel>>#debugger:
  ChestCodeScriptingInteractionModel>>#hasBindingOf:
  ChestCodeScriptingInteractionModel>>#removeBinding:
  StDebugger>>#bindings
  StDebugger>>#interactionModel
  StDebuggerExtensionInspectorNodeBuilderVisitor>>#visitChest:
  StDebuggerExtensionVisitor>>#visitChest:
```

privat:
Nice!

> The goal is to test more things, remove all unused styles, maybe rename #keyword into #selector and delegate more of the style application to the AST

You had my curiosity, but now you have my attention. What are your plan about the AST?

jecisc:
I just finished to:
- Remove all unused style
- Fix bug where wrong style was assigned (we did not see it because the different style share the same color **in the current theme**)
- Add a test for **every** style that is assigned by Shout currently so that we do not have regressions
I updated my first comment in order to list all the changes I did

jecisc:
I still need to fix the failing tests

jecisc:
> Nice!
> 
> > The goal is to test more things, remove all unused styles, maybe rename #keyword into #selector and delegate more of the style application to the AST
> 
> You had my curiosity, but now you have my attention. What are your plan about the AST?

If you check SHRBTextStyler>>#resolveStyleFor: you can see a lot of ifs. I want to kill those by delegating the style name to RBVariableNode>>variable

privat:
Unrelated note: I do not know if you plan to care but currently, IconStyler (the thing that add icons in the vertical icon column in calypso text area) also have some non-disjoint responsibility with the styler since it overlays code text with color and fancy decorations. The major difference seems that it does its job at with RubTextSegmentMorph (so at the UI level, not the Text level) but 1. Color choice in one might conflict with color choice in the other. 2. IconStyler seems to hardcode colors instead of relying on theme or something. Maybe IconStyler could use the same style information as the styler for a more consistent and unified UX?

jecisc:
> Unrelated note: I do not know if you plan to care but currently, IconStyler (the thing that add icons in the vertical icon column in calypso text area) also have some non-disjoint responsibility with the styler since it overlays code text with color and fancy decorations. The major difference seems that it does its job at with RubTextSegmentMorph (so at the UI level, not the Text level) but 1. Color choice in one might conflict with color choice in the other. 2. IconStyler seems to hardcode colors instead of relying on theme or something. Maybe IconStyler could use the same style information as the styler for a more consistent and unified UX?

I like your idea. I don't know if I'll have the time because originally I was only looking at the styler to customize my syntax highlighting and I got lost in the bug fixing and cleaning ahah And people already reported my some bugs that needs to be fixed.
It's at least worth to open an issue because it's bad to have hardcoded values and that the icon styler does not use the shout style table

privat:
Also, It might interest you, but I added more than 250 RBCodeSnippet to systematically test various buggy pieces of code ­— and legitimate non-buggy pieces of code, but those are boring :)

I added a simple test to run the styler on all those pieces of code. Basically it just checks that no internal exception like MNU or other thing are fired when dealing with faulty AST. https://github.com/pharo-project/pharo/blob/Pharo11/src/Shout-Tests/RBCodeSnippetTest.extension.st and https://github.com/pharo-project/pharo/blob/Pharo11/src/Shout-Tests/RBCodeSnippet.extension.st

Do you have any idea to test more than "execution went fine" and check that the result is consistent? Annotating 250 code snippets with an expected styler result seems to be a menial task. Maybe there is something that is easier to check automatically.

<blockquote><div> GitHub</div><div><strong><a href="https://github.com/pharo-project/pharo">pharo/RBCodeSnippetTest.extension.st at Pharo11 · pharo-project/pharo</a></strong></div><div>Pharo is a dynamic reflective pure object-oriented language supporting live programming inspired by Smalltalk. - pharo/RBCodeSnippetTest.extension.st at Pharo11 · pharo-project/pharo</div></blockquote>
<blockquote><div> GitHub</div><div><strong><a href="https://github.com/pharo-project/pharo">pharo/RBCodeSnippet.extension.st at Pharo11 · pharo-project/pharo</a></strong></div><div>Pharo is a dynamic reflective pure object-oriented language supporting live programming inspired by Smalltalk. - pharo/RBCodeSnippet.extension.st at Pharo11 · pharo-project/pharo</div></blockquote>

jecisc:
In this PR I added some utilites to help test the style of a method.

Before, shout was testing the color of the styled nodes. 

In this PR, I added a Mock styler that will not change the style (the colors, emphasis, ...), but it tags the characters with the style that was found for them (instVar, classVar, pragma, ...).

If you want to test really the style that is applied to some elements of your snippets you can check SHRBStyleAttributionTest that I added to this PR

jecisc:
testGmailEncrypted

src/OSWindow-SDL2/SDL_TextEditingEvent.class.st
```diff
@@ -138,24 +138,22 @@ SDL_TextEditingEvent >> start: anObject [
...
 { #category : #accessing }
 SDL_TextEditingEvent >> strlen [
+
 	| len |
 	len := 0.
-	12 to: 12 + 32 - 1 do:
-		[ :i |
-		(self getHandle nbUInt8AtOffset: i) = 0
-			ifTrue: [ ^ len ].
+	12 to: 12 + 32 - 1 do: [ :i |
```
demarey, line 144: Argh. All these magic numbers are really bad!
(I know it is not related to this PR)

src/Shout-Tests/SHRBMockSharedPoolStyler.class.st
```diff
@@ -0,0 +1,11 @@
+"
+I am a shared pool used for styler tests.
+"
+Class {
+	#name : #SHRBMockSharedPoolStyler,
```
demarey, line 5: I do not see where it is used

src/Shout-Tests/SHRBTextStylerTest.class.st
```diff
@@ -31,96 +31,13 @@ SHRBTextStylerTest >> style: aText [
...
-]
-
 { #category : #running }
 SHRBTextStylerTest >> tearDown [
 
 	SHRBTextStyler formatIncompleteIdentifiers: oldSetting.
 	super tearDown
 ]
 
-{ #category : #tests }
```
demarey, line 56: Where are the new tests equivalent to those removed?

src/OSWindow-SDL2/SDL_TextEditingEvent.class.st
jecisc: I avoided to commit them multiple time but I forgot once :( should be pretty safe tho

src/Shout-Tests/SHRBMockSharedPoolStyler.class.st
jecisc: It is used as a shared pool in the test class to be able to test shared pool style

src/Shout-Tests/SHRBTextStylerTest.class.st
jecisc: Yes, but not relying on the current theme! Some of those were in reality failing but we didn’t see it since it was relying on the theme

src/Shout-Tests/SHRBMockSharedPoolStyler.class.st
jecisc: 0FC59509-323C-4181-A181-FB79A0D451D6

MarcusDenker (approved):

src/Shout-Tests/SHRBTextStylerTest.class.st
demarey: ah, sorry, I did not see the `load diff` hiding a lot of changes.

demarey (approved):

PR #12919, opened 2023-03-07
Improve faulty compilation - semantic error

privat:
This PR extends AST API with basic methods for semantic error

* `addError:` to let compiler add semantic error information on Node
* `isError` to let AST user knows if the node is erroneous
* `isFaulty` is extended to encompass both syntax error (`RBParseErrorNode`) and other kind of errors (`isError`)

The internal implementation is to use a node property `#error`.

Currently, (I'm not aware of a plan to add more semantic error, so this is not the better term) there seems to be only 2 semantic errors: unwritable variable (for some reason read-only and reserved are distinguished) and too much argument (for block and methods), the latter was moved from backend to the semantic analysis (because the whole point of semantic analysis is to get these errors before compilation).
The status of undeclared variable as error is unclear and seems to depend on UI related stuff (and possibly the humidity level).

The code is deliberately written as simple as possible, first attempts were too much complex for a limited gain.

A possible future improvement could be to unify and reify errors and warnings (strings are poor).
Maybe look at how critiques does the deal.

UI related concerns (like nice icons in the icon bar) will be added in a future PR.

Another question is how much people like (or rely) on the current semantic error/warnings exception hierarchy (`OCSemanticError`, `OCStoreIntoReadOnlyVariableError`, `OCStoreIntoReservedVariableError`, `OCSemanticWarning`, `OCShadowVariableWarning`, `OCUndeclaredVariableWarning`)

### DIFF
--- a/src/AST-Core-Tests/RBCodeSnippet.class.st
+++ b/src/AST-Core-Tests/RBCodeSnippet.class.st
@@ -442,8 +442,8 @@ RBCodeSnippet class >> badSemantic [
 		"Backend errors"
 		"FIXME: syntax error are thrown by the compiler even in *faulty* mode.
 		 FIXME: semantic analysis does not catch the error (backend issue)"
-		self new source: 'foo ^ [ :a1 :a2 :a3 :a4 :a5 :a6 :a7 :a8 :a9 :a10 :a11 :a12 :a13 :a14 :a15 :a16 | a1 ]'; isMethod: true; isFaulty: true; skip: #compile; skip: #testDoSemanticAnalysisOnError. "Too many arguments"
-		self new source: 'a1: a1 a2: a2 a3: a3 a4: a4 a5: a5 a6: a6 a7: a7 a8: a8 a9: a9 a10: a10 a11: a11 a12: a12 a13: a13 a14: a14 a15: a15 a16: a16 ^ a1'; isMethod: true; isFaulty: true; skip: #compile; skip: #testDoSemanticAnalysisOnError. "Too many arguments"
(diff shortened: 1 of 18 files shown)


MarcusDenker:
Broken build, not related to this change:

- [CI] full image load is broken #12926

MarcusDenker (approved):

demarey (commented):
I think #isFaulty and #isError should go in #testing protocol and not #errors.
It is the standard in the Pharo image.

demarey:
So, are comments useful if merged without answer ???

privat:
@demarey where is the standard/style guide about protocols/categories on Pharo? I did not find one.

Personally, I prefer when methods are grouped by concerns, but if there are some rules I can follow them.

demarey:
@privat You have some rules encoded in the Method classifier. Foe example, you can see methods starting with `is` are classified in the `testing protocol`. https://github.com/pharo-project/pharo/blob/Pharo11/src/Tool-Base/MethodClassifier.class.st#L84
At the end, it is up to you to choose the protocols.

<blockquote><div> GitHub</div><div><strong><a href="https://github.com/pharo-project/pharo">pharo/MethodClassifier.class.st at Pharo11 · pharo-project/pharo</a></strong></div><div>Pharo is a dynamic reflective pure object-oriented language supporting live programming inspired by Smalltalk. - pharo/MethodClassifier.class.st at Pharo11 · pharo-project/pharo</div></blockquote>

privat:
@demarey thank you for the pointer. Although a list of heuristics on substrings to determinate some default protocol is not what I envisioned for "rule" or a "standard" :)
They are fine as they seem able to guess something that works (and far more usable than "as yes unclassified"), but I do not think programmers should follow them blindly.

Issue #12917, opened 2023-03-07
[CI] testGmailEncrypted timeout


MarcusDenker:
**Bug description**
testGmailEncrypted times out often

MarcusDenker:
There is ZnHttpsTest and ZnHttpsTests, one has the timeout, one not

Solution: unify the two

svenvc:
The plural one, ZnHttpsTests is from the Zinc repository:

https://github.com/svenvc/zinc/tree/master/repository/Zinc-Zodiac-Tests.package/ZnHTTPSTests.class

From my point of view, ZnHttpsTests can be removed.

<blockquote><img src="https://opengraph.githubassets.com/970f7e5dcea0bcdfffea549ec7336428d96a4f9c77d72a969d05ab61f228adf5/svenvc/zinc" width="48" align="right"><div><img src="https://github.githubassets.com/favicons/favicon.svg" height="14"> GitHub</div><div><strong><a href="https://github.com/svenvc/zinc">zinc/repository/Zinc-Zodiac-Tests.package/ZnHTTPSTests.class at master · svenvc/zinc</a></strong></div><div>Zinc HTTP Components is an open-source Smalltalk framework to deal with the HTTP networking protocol. - zinc/repository/Zinc-Zodiac-Tests.package/ZnHTTPSTests.class at master · svenvc/zinc</div></blockquote>

MarcusDenker:
In Pharo11, there is just ZnHTTPSTest (no s) and it has no timeout.


So maybe the sync did not work correctly?

svenvc:
Ah OK, hmm, I saw this is my work pharo 11 image from some time ago, where we were doing the integration.

BTW, they are in different packages also:

- ZnHttpsTests is in Zinc-Zodiac-Tests
- ZnHttosTest is in Zinc-Zodiac

That last package does not exist in my Zinc repository.

jecisc:
The problem is that in Pharo we have the package Zinc-Zodiac and in Zinc we have Zinc-Zodiac-Core and Zinc-Zodiac-Tests

So Iceberg was not able to find the diffs during the last sync since the packages are different. 

I'll try to check was is the equivalent of what was integrated before and sync the name changes

MarcusDenker:
Still happening

MarcusDenker:
ah, no, PR was merged

PR #12930, opened 2023-03-07
Speed up ClassVar>>#isReferenced

### LINKED ISSUE #12928: #isReferenced is slow because of shared pools
The releases tests checking if variables are referenced are really slow (30sec for 1 test) and I see that 2/3 of the time is spend in finding the pool users of pool variables. 

Maybe there are things to improve on the implementation of shared pool so that they do not slow down that much such methods?

jecisc:
Fixed #12928


jecisc:
This will make the release tests way faster :)

MarcusDenker:
- [CI] testGmailEncrypted timeout #12917

MarcusDenker (approved):

PR #12932, opened 2023-03-07
Synch zinc zodiac

jecisc:
In Zinc the package Zinc-Zodiac is called Zinc-Zodiac-Tests

This PR aligns this

### DIFF
--- a/src/BaselineOfZodiac/BaselineOfZodiac.class.st
+++ b/src/BaselineOfZodiac/BaselineOfZodiac.class.st
@@ -9,12 +9,12 @@ BaselineOfZodiac >> baseline: spec [
 	<baseline>
 	spec for: #'common' do: [
 		spec 
-			package: 'Zinc-Zodiac';
+			package: 'Zinc-Zodiac-Tests';
 			package: 'Zodiac-Extra';
 			package: 'Zodiac-Tests'.
 
 		spec 
-			group: 'Core' with: #('Zinc-Zodiac' 'Zodiac-Extra');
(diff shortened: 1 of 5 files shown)


svenvc:
Yes, that is it. Thank you

MarcusDenker:
Failing tests:

- ZnHTTPSTests>>testGmailEncrypted fails with to many redirects, this was #10290
- NoUnusedVariablesLeftTest>>testNoUnusedTemporaryVariablesLeft for ZnHTTPSTests>>#testGetPharoVersion
- ReleaseTest>>testTestCasesEndsWithTestOrTestCase
- ProperlyImplementedSUnitClassesTest>>testTestClassesShouldBeSingular

jecisc:
A few comments:
- The test still fails because of too many redirection: I propose to increase the limit in the test
- The class was renamed in Pharo to not have the "s" because there is a rule that test case needs to ed with "TestCase" or "Test" in Pharo to be coherent
- The test about pharo version was commented in zinc. I propose to just remove it

I'll update this PR and do one in Zinc too

jecisc:
The little sister of this PR: https://github.com/svenvc/zinc/pull/112

demarey (commented):
Are you sure we do not need the tests with Gitlab and pharo version?
We could add them in another class not packaged in Zodiac if needed.
I would ask to Pharo devs before removing them.

jecisc:
This fix a failing test of the ci. Maybe we can merge it and create a new PR for those two tests?

jecisc:
https://github.com/pharo-project/pharo/issues/12939

PR #12933, opened 2023-03-07
add RBCodeSnippetTest>>#testCodeImporter

privat:
Extends RBCodeSnippetTest to test CodeImporter

CodeImporter is a client of OpalCompiler and might be confronted to faulty code.
This PR does not change the behavior of the importer (just add test). Therefore, compilation errors are expected to be produced on faulty code.
In classical use, such compilation errors are caught by the buggy SyntaxErrorDebugger. But the present test only check (and capture) the SyntaxErrorNotification signals, and does not check the recoverability. See #12910 for related rants.

The test is a little complex because there are basically two modes, one for expressions and one for methods, and we need a placeholder class and to retrieve the method in the latter.

Note: I do not know how much the CodeImporter is used in real life.


privat:
Errors seems related to #12935 . Merge this other one and rerun the tests

privat:
PR #12995 was merged and now the test are OK (except random windows weirdness)

MarcusDenker (approved):

PR #12935, opened 2023-03-07
Improve faulty compilation: sanify undeclared variables

privat:
Previously, running

```st
{
OpalCompiler new evaluate: 'a := 1. a'.
OpalCompiler new evaluate: 'a'.
OpalCompiler new options: #(+optionSkipSemanticWarnings); evaluate: 'a := 2. a'.
OpalCompiler new options: #(+optionSkipSemanticWarnings); evaluate: 'a'.
}
```

gave `#{1 1 nil nil}` (and no exception or nothing).

That is so weird.

* unrelated `a` variables are bound together in the first 2 expressions.
* related `a` variables are not bound together in the third expression.
* `optionSkipSemanticWarnings` should just skip warnings, not change the whole semantic of the language.

The PR tries to address this mess:

* Undefined variables are grouped together in a same semantic analysis. Basically playing as implicit temps
* OCUndeclaredVariableWarning stop creating globally shared undefined variable by default, and just use the implicit temp.

Now the same code give `1 nil 2 nil` \o/

### DIFF
--- a/src/AST-Core-Tests/RBCodeSnippet.class.st
+++ b/src/AST-Core-Tests/RBCodeSnippet.class.st
@@ -402,8 +402,13 @@ RBCodeSnippet class >> badSemantic [
 	| list |
 	list := {
 		"Undefined variable"
-		"For some reason, this is just a warning in non-faulty mode, not an error."
-		self new source: 'a := 10. ^ a'; isFaulty: false; value: nil. "Weird, I expected to be 10"
+		"In non-faulty mode, undeclared variables are shared at the method level like implicit temporaries.
+		 However, specific context/requestors can change the deal."
+		self new source: 'a := 10. ^ a'; isFaulty: false; value: 10.
+		self new source: '[ a := 10 ]. ^ a'; isFaulty: false; value: nil.
+		self new source: '[ a := 10 ] value. ^ a'; isFaulty: false; value: 10.
+		self new source: 'b := [ a ]. a := 10. ^ b value + 1'; isFaulty: false; value: 11.
+		self new source: 'b := [ ^ a ]. a := 10. ^ b value + 1'; isFaulty: false; value: 10.
 		self new source: '^ a'; isFaulty: false; value: nil.
 
 		"Uninitialized variable"
--- a/src/Calypso-SystemPlugins-Undeclared-Queries-Tests/ClyClassWithUndeclares.class.st
+++ b/src/Calypso-SystemPlugins-Undeclared-Queries-Tests/ClyClassWithUndeclares.class.st
@@ -14,10 +14,10 @@ ClyClassWithUndeclares class >> withFloatReturn [
 
 { #category : #'with undeclares' }
 ClyClassWithUndeclares >> method1WithUndeclares [
-	^undeclaredStubInstVar1 printString
+	^UndeclaredStubInstVar1 printString
 ]
 
 { #category : #'with undeclares' }
 ClyClassWithUndeclares >> method2WithUndeclares [
-	undeclaredStubInstVar2 := 100
+	UndeclaredStubInstVar2 := 100
 ]
--- a/src/Kernel-Tests/CompiledCodeTest.class.st
+++ b/src/Kernel-Tests/CompiledCodeTest.class.st
@@ -212,22 +212,22 @@ CompiledCodeTest >> testLiteralsDoNotConsiderTheInnerBlockLiterals [
 	| method block |
 	method := self class compiler compile: 'method
 		<pragma: #pragma>
-		test := 1+2.
+		Test := 1+2.
 		Class.
 		self doIt: [
-		test := 1 - 2.
-		test := #(arrayInBlock).
+		Test := 1 - 2.
+		Test := #(arrayInBlock).
 		Object.
 		self name ].
 		^#(#array) '.
 	block := (self class compiler evaluate: '[
-		test := 1 - 2.
-		test := #(arrayInBlock).
+		Test := 1 - 2.
+		Test := #(arrayInBlock).
 		Object.
 		self name ]') compiledBlock.
 	self assertCollection: method literals equals: {
 			2.
-			(UndeclaredVariable key: #test value: nil).
+			(UndeclaredVariable key: #Test value: nil).
 			block.
 			#doIt:.
 			#( #array ).
--- a/src/Kernel/UndeclaredVariable.class.st
+++ b/src/Kernel/UndeclaredVariable.class.st
@@ -12,6 +12,21 @@ Class {
 }
 
 { #category : #'instance creation' }
+UndeclaredVariable class >> getWithName: aString [
+	"Undeclared variable from faulty code use the following rule:
+
+	If the name starts with a uppercase, the undeclared variable is registered, and if one already exists it is reused.
+	It means that the undeclared variable is shared (implicit global) and that future global variable (like a class) with the same name will replace it.
+
+	If the name starts with a lowercase, a new unregistered undeclared variable is returned.
+	"
+
+	^ aString first isUppercase
+		  ifTrue: [ self registeredWithName: aString ]
+		  ifFalse: [ self named: aString ]
+]
+
+{ #category : #'instance creation' }
 UndeclaredVariable class >> registeredWithName: aString [
 	| varName |
 	varName := aString asSymbol.
--- a/src/OpalCompiler-Core/OCASTSemanticAnalyzer.class.st
+++ b/src/OpalCompiler-Core/OCASTSemanticAnalyzer.class.st
@@ -137,6 +137,7 @@ OCASTSemanticAnalyzer >> undeclaredVariable: variableNode [
 	^ OCUndeclaredVariableWarning new
 		  node: variableNode;
 		  compilationContext: compilationContext;
+		  undefinedVariable: var;
 		  signal
 ]
 
--- a/src/OpalCompiler-Core/OCASTSemanticAnalyzer.class.st
+++ b/src/OpalCompiler-Core/OCASTSemanticAnalyzer.class.st
@@ -8,7 +8,8 @@ Class {
 	#instVars : [
 		'scope',
 		'compilationContext',
-		'blockCounter'
+		'blockCounter',
+		'undeclaredVariables'
 	],
 	#category : #'OpalCompiler-Core-Semantics'
 }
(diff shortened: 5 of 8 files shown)


privat:
I may need a bootstrap specialist to understand why undefined variables need to be global :( @guillep?

privat:
OK. I might have it....

Undeclared variables are not just a gimmick of faulty code. It's basic Pharo behavior.
When removing a class (or any global), or an instance variable (or any shared variable). Existing methods with those variables are updated to use undeclared variables (as placeholder) instead.
If an equivalent variable with the same name are added back, they can then take it rightful place in existing code.
I can of like it, it's between pure dynamically executed code (where you resolve binding at runtime without any compilation phase) and pure statically compiled code where one everything is compiled it's immutable.

Moreover, when installing new code from a package (with many classes and methods), global variables might refer to classes not yet loaded (because static dependency graph is hard?) so registered undefined variables are used, until a class with the good name is loaded and can replace the undefined variable.

And package loading is obviously used during the bootstrap. This what we get in the falling CI log:

```
#new was sent to nil
UndefinedObject(Object)>>doesNotUnderstand: #new
UndefinedObject>>doesNotUnderstand: #new
MCWorkingCopy>>initialize
```

Whereas `MCWorkingCopy>>initialize` only `new` is `ancestry := MCWorkingAncestry new`.
And a previous warning in the log warned us that `NewUndeclaredWarning: MCWorkingCopy>>initialize (MCWorkingAncestry is Undeclared)`.
One can imagine that when `MCWorkingAncestry` was eventually loaded, the undefined reference in `MCWorkingCopy>>initialize` was not updated because the PR made it an unregistered plain undefined variable (instead of a registered global one).

Unrelated node: it seems that in previous Pharo versions (or even is was a Smalltalk thing?), bindings were simple plain associations, updating a binding was easy: just change the value. Now binding are reified into rich objects, so updating them need a *become* operation.

privat:
Since I do not want to close the PR yet since the behavior is still inconsistent, I propose a new rule:
"*undefined variables are registered (global) when they start by an uppercase*", therefore.

```st
{
OpalCompiler new evaluate: 'a := 1. a'.
OpalCompiler new evaluate: 'a'.
OpalCompiler new options: #(+optionSkipSemanticWarnings); evaluate: 'a := 2. a'.
OpalCompiler new options: #(+optionSkipSemanticWarnings); evaluate: 'a'.
}
```

gives `{1. nil. 2. nil}, but

```st
{
OpalCompiler new evaluate: 'A := 1. A'.
OpalCompiler new evaluate: 'A'.
OpalCompiler new options: #(+optionSkipSemanticWarnings); evaluate: 'A := 2. A'.
OpalCompiler new options: #(+optionSkipSemanticWarnings); evaluate: 'A'.
}
```

gives `{1. 1. 2. 2}`

Let's see how the CI likes it.

privat:
Tests are better but I'm still not sure about the whole thing. Thoughts?

MarcusDenker:
there is now a conflict

privat:
Yeah, merge conflict, but I would like to get your input and discuss the issue further before trying to solve the conflict (even if it's an easy solve)

MarcusDenker:
Yes, sorry for being a bit slow: I have blocked time this afternoon to check an comment on all the remaining open PRs

MarcusDenker:
Yes, your understanding of "Undeclared" is correct. Here are some notes that I wrote some time ago (I should have sent it to you before...)

This explains the mechanism that we have now and then has some raw ideas of an alternative implementation.

# Undeclared Variables

Undeclared Variables that are used (read or written), but not defined. As such, we have no information other than the name.

They are handled differently in interactive and non-interactive mode: In interactive mode, we raise an exception that shows to the developer a menu that allows the developer to define the variable.

- The resulting infrastructure on the compiler side is a bit complex, see the exception "ReparseAfterSourceEditing".
- The user interaction can be annoying as it raises dialogs for each var. 

For the case of compiling an Undeclared:

The compiler will add an association-like object (UndeclaredVariable) to the Undeclared dictionary (a Dictionary pointed to by the global variable "Undeclared"). It then compiles standard global access: it uses the bytecodes pushLiteraVariable/storeIntoLiteralVariable to modify the second ivar of the Variable Object (binding) which is referenced from the literal frame of the method.

When defining new Globals, it will check if an Undeclared of that name already exists, copy over the value and remove it entry from the Undeclared Dictionary. This fixes all accessing methods without needing a recompile.

For undeclared local methods, it will just recompile all methods of the class as soon as you add a new ivar, the value stored in the global is not retrieved (it is just one value, for an ivar it makes not much sense to recover that. Thus a new ivar starts with nil even though code with an Undeclared of that name has been written to before).

# Good Properties

- We can load uses before definitions in non-interactive mode
- We are asked to define a Variable of compile when in interactive mode
- When loading the definition, the uses are fixed automatically with no intervention needed
	- For Globals, this works even with no need for the compiler!
	- For Globals, this means that it fixes even hot methods that are on the stack
	
- We do support limited execution capabilities, which allows code to not just load but *be executed* with undeclared variables
	But Caveat! It only is correct for Globals! For ivars it just has one global representing what should be per-object state.


# Problem 1: Undeclared can be executed at runtime with no User Feedback

If you have an Undeclared, it reads as nil and we can write to it as a global. 

This is for one good, as being able to execute such code in non-interactive mode is nice.

But on the other hand, it is bad:
- If an Undeclared is sneaking in, the Programmer might find it very late. Executing code with tests will not find it!
- It is very very confusing to the developers. We are getting bug reports for this regularly!
  - Non-interactive compile happens more often than you think, e.g. on removing an ivar from a class
  - This is especially bad as the resulting execution uses a Global for a variable that is instance state and hides the fact that execution is completely incorrect, see the next point.

# Problem 2: All Undeclared Variable are modeled as Globals

An undeclared instance var or temp leads to one shared global of that name.

We could to better: Variables names that are not capitalized we know are either temps or instance state.
Temps are only accessed in one method (and written to first).

We could see "Undeclared write" as examples that teach the Variables which role it has. Even the first step:
compiling what should be a temp as an ivar is much better than what we have now, as it would be correct for ivars.
Maybe we can come up with a heuristic if a variable should be a temp or not.


# Small bad properties:
- To fix the instance state, we need the compile and thus rely on the compiler anyway.
- Due to combining "save" and "compile", we forcing the programmer to define a variable every time a method is saved

	
# Would it be a problem to rely on recompilation to fix undeclared variable access?
We need the compiler to be present *and* methods that are on the stack would not be updated.

But we do not have to rely on the compiler, we could have a fall back: Recompilation would only be needed to remove the Undeclared read to make it fast.

By late-binding Undeclared access, we could in case that a recompile is not possible, make the late-bound version do the right thing. 
- This would just be slower
- but it would work. This would e.g. mean that methods on the stack would be able to work after a var is defined

If we compile explicit read and write to the Undeclared, we could even fix that and replace it (without the need of a compiler) with the now valid variable after it has been defined. This would then just leave the difference of "pushLiteralVariable" vs. "pushLiteral, send, execute quickReturn" when e.g loading code without a compiler.

	
# What do we want?

- In non-interactive mode: code with Undeclards is loadable 
- The resulting code can be executed in non-interactive mode
- We want an exception in interactive mode on execute (which shows a menu: define this var!)

# How do we get this?

- We compile a message to an instance of an Undeclared Variable
-- for globals (vars with capital letter), we just compile a #read and #write
-- for ivars and temps, we compile a send that hands over the current context

- read and write check if we are in IDE mode or not
	yes: Show menu
	no: simulate a ivar/global/temp, by storing the state on the global, as state per var and instance for ivars and even as context state if we know that this is a temp.

- read and write check if a var has been defined in the meantime on read and write and need to be able to forward to the real var in this case, without recompile.

- We need to support fixing the var access if a compiler is available
	- either at read/write or at definition of vars. The current model does the latter, and I think we should do the same.
		- read/write checks if a definition has been added and *forwards*
		- any new definition of variables check if they can fix read/writes of Undeclareds by re-compiling

- We have different kinds of Unknown Variable classes and select the right one according to a heuristic. 
	- at the beginning, we just support ivar and global, heuristic is if the var name is capitalized
	- I think we can support temps, too (just accessed in one method, it is assigned first). But this then needs the possibility to mutate a temp to an instance var as more code is loaded. (in a way, we would teach the var its role slowly).
	
- With this in place, we would not need to ask to define variables of "save". We could just compile an Undeclared and the developer fixes it with the gutter "repair" icon or as part of running the tests. Fixing undeclared vars would thus work the same as fixing a DNU.
As a side effect, we can radically simplify the exception handling in the compiler and remove the "ReparseAfterSourceEditing" mechanism.

- We can simplify the #doesNotUnderstand: on nil, as it now tries to find out if it happens on a nil that comes from a Undefined Global to propose creating classes at execution time.

privat:
Thant you for the long text. I'm glad I reverse-engineered the code and the specification mostly correctly :)

I though little more, I'm fairly certain that I do not like (in no particular order)

* that `optionSkipSemanticWarnings` changes the semantic of the language
* that capitalization changes the semantic of the language. It is OK for heuristics used to generate suggestion or critiques in a GUI, but not really for a compiler to guess the behavior of programs. (a possible change can be to put explicitly in the language specification (the postcard?) that identifiers of global variables are separated from the ones or other variables (but this is not enough if we want to also enforce the rules of read-only globals vs read-write globals (or shared)).
* that the result of a compilation depends on the nature of the caller (or which subclass of Morph it is an instance of)
* that compiled code behave in a quite unpredictable way (or fragile or vulnerable) source of bugs and programmer frustration.

Eg the following example that circumvent the static checking that prevent writing on read-only global:

```st
"clean up if you want to replay"
Smalltalk globals removeKey: #UnknownB ifAbsent: [].
Smalltalk globals removeKey: #AClass ifAbsent: [].
Undeclared removeKey: #UnknownB ifAbsent: [].

{
'Object subclass: #AClass'.
'AClass compile: ''foo UnknownB := 42'''.
'Object subclass: #UnknownB'.
'UnknownB'.
'AClass new foo'.
'UnknownB'.
} collect: [ :s | Smalltalk compiler evaluate: s ]
```

However, I think I understand the main needs and hopes:

* A notion of undeclared variable should exist; because of package loading or variable deletion (class removal, ivar removal, etc.)
* Repairing code is not an easy task (e.g. simply recompiling does not change old versions in the stack), so generate compiledMethods that use undeclared variables with some late binding (slow) or (easily) reparable bytecode. But maybe also for compiledMethods that use declared variables (far more numerous) as some of them might become undeclared (or worse, undeclared then redeclared as another kind of variable)?
* Having error at runtime (with possible fix, like we do for DNU for instance) is far better than loosely defined and silent semantic.

MarcusDenker:
Good summary!

And yes, I agree with the list of points

privat:
I do not like the approach. So close the PR. Made a distinct open issue for the moment:  #13006

PR #12945, opened 2023-03-08
Shout: Rename some styles

### LINKED ISSUE #12920: Shout: Rename keyword into selector
In shout we had #keyword #binary #unary (and equivalents with #undefinedKeyword and #incompleteKeyword) but currently only keyword is set by the styler. 

I propose to rename that style to #selector 

I'll propose that soon. I'm just creating an issue not to forget

### LINKED ISSUE #12924: Shout: Merge left/right parenthesis and block start/block end
Currently, in shout, we have #leftParenthesis and #rightParenthesis. I don't see why we would need a different style for both. I propose to rename those to #parenthesis directly. 

Same with #blockStart and #blockEnd. Let's rename that #block. 

I'll propose that later.

jecisc:
Some styles have bad names and this PR fixes that.

- #keyword is used to style any selector => Renamed into #selector
- #incompleteKeyword is used to style any selector => Renamed into #incompleteSelector
- #undefinedKeyword is used to style any selector => Renamed into #undefinedSelector
- #pragmaKeyword is used to style any selector => Renamed into #pragma

Fixes #12920

This also merges left/rightParenthesis into parenthesis, blockStart/blockEnd into block, left/rightBrace into brace.
Because... Come on, who wants their parenthesis different colors when they are matching?! (This does not change the fact that you can have different colors for imbricated parenthesis)

Fixes #12924


jecisc:
Failure seems unrelated

MarcusDenker (approved):

PR #12946, opened 2023-03-08
SDL2 on MacOS required to postpone focus

### LINKED ISSUE #10981: MacOS Monterey: On startup, image is unresponsive
**Bug description**

The image doesn't react to left clicks and window title bar is grayed, as if the OS window didn't grab focus. This is fixed when you switch to other MacOS window and come back to Pharo.

According to chats, this only happens in MacOS Monterey.

**To Reproduce**

- Download Pharo with: `curl https://get.pharo.org/64/100+vmLatest | bash`
- From the terminal, do `./pharo-ui Pharo.image`
- Keep moving the cursor with the track-pad immediately after pressed enter on the terminal until the image opens. 
- The window opens and single-left-click doesn't produce any effect on the World. Another symptom is that window title bar is greyed. Strangely, double-left-click works normally.
- That buggy behavior stays until you switch to another OS window (e.g. cmd+tab), and come back to Pharo. This time, the image gets the mouse focus and works normally.

**Version information:**
 - OS: MacOS Monterey (12.1)
 - Pharo Version: 10

tinchodias:
- Fix #10981
- Revert previous unsuccessful fix: https://github.com/pharo-project/pharo/pull/11082

### DIFF
--- a/src/OSWindow-Core/OSWorldRenderer.class.st
+++ b/src/OSWindow-Core/OSWorldRenderer.class.st
@@ -129,14 +129,14 @@ OSWorldRenderer >> doActivate [
 	driver afterSetWindowTitle: Smalltalk image imageFile fullName onWindow: osWindow.
 	driver startUp: true.
 
-	"SDL2 on MacOS required to postpone focus.
-	See: https://github.com/pharo-project/pharo/issues/10981"
-	[ osWindow focus ] fork.
-
 	world worldState doFullRepaint.
 	world displayWorld.
 
-	OSWindowClipboard new beDefault
+	OSWindowClipboard new beDefault.
+
+	"SDL2 on MacOS presented a bug if this message send is done earlier on this method.
+	See: https://github.com/pharo-project/pharo/issues/10981"
+	osWindow focus
 ]
 
 { #category : #operations }
--- a/src/OSWindow-Core/OSWorldRenderer.class.st
+++ b/src/OSWindow-Core/OSWorldRenderer.class.st
@@ -129,7 +129,9 @@ OSWorldRenderer >> doActivate [
 	driver afterSetWindowTitle: Smalltalk image imageFile fullName onWindow: osWindow.
 	driver startUp: true.
 
-	osWindow focus.
+	"SDL2 on MacOS required to postpone focus.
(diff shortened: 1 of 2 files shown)


guillep:
Can you give more details? the focus needs to be postponed until when?
The fork looks like a short term solution also...

guillep:
BTW just to give a positive note: I think that it's super mega great and important that you look into this.
My comment is just that we would need to have a better comment in case this bites us again in the future :P

jecisc:
I confirm what Guille says, it's awesome that you look at this problem because it's really a pain in the a** :)

MarcusDenker:
Yes! This is nice!

Failing tests not related:

-  [CI] System variable environment on windows does not use adequate encodings #11407

tinchodias:
thanks

@guillep The PR is based on all the observations I did on https://github.com/pharo-project/pharo/issues/10981 plus guessing. 

Sadly I didn't arrive to a comprehensive explanation of the bug's cause. But I think this kind of issues could happen in SDL because it is an abstraction over several platforms, each one with their own trickiness, their evolution along the time, etc. But I was so annoyed by this issue that I tried to guess a fix, at least. 

We could:
- write a small C program to check the bug reproduces there, and then report it to SDL2 project;
- or debug via lldb or something;
- or read the SDL2 for Macos. IIUC, this is the function that this "osWindow focus" calls: https://github.com/libsdl-org/SDL/blob/release-2.24.1/src/video/cocoa/SDL_cocoawindow.m#L1971

<blockquote><div> GitHub</div><div><strong><a href="https://github.com/libsdl-org/SDL">SDL/SDL_cocoawindow.m at release-2.24.1 · libsdl-org/SDL</a></strong></div><div>Simple Directmedia Layer. Contribute to libsdl-org/SDL development by creating an account on GitHub.</div></blockquote>

tinchodias:
Hi!
1. Was this change a fix in your computers? it works on my Macbook 2018, maybe not others.
2. Or should I search more the cause of the bug before merging?
In any case, this change shouldn't break anything, it mostly moves the SDL2 FFI call that requests window focus a few lines below, in the same method.

guillep:
lets try it

PR #12949, opened 2023-03-08
Remove dead code in SHTextStyler

### LINKED ISSUE #10080: Syntax Highlighting: unused SHTextStyler>>#styleInBackgroundProcess:
There is still some (complex) code that used to be used to run text styling in a background process. 

See  SHTextStyler>>#styleInBackgroundProcess:

**Describe the request**
We need to decide what to do: just keeping code around so we can look at it in the future.. for that we have Git.

- the code is not used for years
- It was turned off as is lead to lots of race-conditions
- We solved the case of Huge methods (which are not nicely solved with this either) by not styling huge literals

jecisc:
Fixes #10080

### DIFF
--- a/src/Shout/SHTextStyler.class.st
+++ b/src/Shout/SHTextStyler.class.st
@@ -11,8 +11,6 @@ My public interface is...
 
 	style: aText - modifies the TextAttributes of aText, but does not change the string, then sends #stylerStyled: to the view.
 
-	styleInBackgroundProcess: aText - performs style: in a background process, then sends #stylerStylednBackground: to the view.
-
 	styledTextFor: aText - answers a formatted and styled copy of aText
 
 	unstyledTextFrom: aText - answers a copy of aText with all TextAttributes removed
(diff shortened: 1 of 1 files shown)


Ducasse:
This is cool to clean. The changes are good. Now I was wondering if using the background process was not a good idea for syntax highlighting.

jecisc:
From what Marcus told me it’s not really possible because the text editor is not thread safe

MarcusDenker:
Another thing is that Shout used it's own parser, now we use the normal AST which we have to create anyway for the tools to work (e.g. menu, code completion, and so on).

In practice sadly they all re-parse, but instead of investing time in a complex workaround like this we should make sure that Shout and all other tools re-use the AST from the code editor instead of re-parsing.

MarcusDenker (approved):

MarcusDenker:
And if applying styles to the editor would be too slow, it should be solved on the level of the editor for all users who want to style text, nut just like this for only the code highlighter

privat:
You should update the class comments that still talk about removerd stuff

jecisc:
I'm not finished yet with the changes in shout. If it's alright with you i'd like to update the comment in a later PR

PR #12950, opened 2023-03-08
Remove font logic from Shout since it is all dead code

jecisc:
Iâ¯deprecated some stuff that might be used by Spec


MarcusDenker:

jecisc:
I changed the base to be Pharo 12 and this will need to be updated to create a Deprecated12 instead of using Deprecated11

src/Deprecated11/SHRBTextStyler.extension.st
```diff
@@ -0,0 +1,23 @@
+Extension { #name : #SHRBTextStyler }
+
+{ #category : #'*Deprecated11' }
```
privat: Deprecated12?

src/Deprecated11/SHRBTextStyler.extension.st
```diff
@@ -0,0 +1,23 @@
...
+SHRBTextStyler >> font: aFont [
+
+	self deprecated: 'This variable is never used and was removed.'
```
privat: if never used, maybe just kill it?

jecisc: The variable was never used internally, but I don’t know if the class has a user that set the font. #font: is too common for me to check all senders and guess it the receiver is the styler or not ahah

src/Deprecated11/SHRBTextStyler.extension.st
jecisc: Yeah I originally did this Pr on P11 and I need to update it. I’ll do it today

MarcusDenker:
ReleaseTest>>testUndeclared

'''Found undeclared Variables:  textAttributesByPixelHeight in: MicSmalltalkTextStyler class>>#initialize'''

MarcusDenker (approved):

Issue #12126, opened 2022-12-19
[CI] Fail on WIN random: OmSessionStoreTest(OmStoreTest)>>testEntryForAbsentEntry


MarcusDenker:
**Bug description**
OmSessionStoreTest(OmStoreTest)>>testEntryForAbsentEntry fails on the CI on Windows only randomly

```

FileDoesNotExistException
'C:\Users\ci\AppData\Local\Temp\ombu-tests\test-sessions-other\Pharo.38D0549E180.ombu'
```

jecisc:
I don't have time to find the origin, but I can reproduce it locally on one of my machine

jecisc:
I added a logs in OmFileStore>>#flushEntryBuffer, OmFileStore>>#flush and OmFileStore>>#readEntriesWith:

When the test pass I get:

```
Flush
Create D:\Temp\ombu-tests\test-sessions-main\Pharo 12.0 - 64bit (next development version, latest)-01.2F97BA4A5200.ombu
Flush
Flush
Flush
```

When it fails I get 


```
Reading D:\Temp\ombu-tests\test-sessions-other\Pharo 12.0 - 64bit (next development version, latest)-01.2F98127211C0.ombu
```

==========

Going a little further.

When the test pass, we never go in #checkIfMustRefreshBlock: but when the test fails, we go in #checkIfMustRefreshBlock:

===========

A more precise divergeance point is in OmFileStore>>#entryFor:ifPresent:ifAbsent:.

When the test passes, the reference global name and the store global name are the same. But when the test fails, they are different.

============

Trying with more logs:

Passing:

```
Setup
Reference initilized with global name: nil
Store initialized with global name: Pharo 12.0 - 64bit (next development version, latest)-01.31FD2FC61E80
Reference initilized with global name: nil
Store initialized with global name: Pharo 12.0 - 64bit (next development version, latest)-01.31FD2FE4A300
Test
Reference initilized with global name: Pharo 12.0 - 64bit (next development version, latest)-01.31FD2FC61E80
Reference global name: Pharo 12.0 - 64bit (next development version, latest)-01.31FD2FC61E80 store global name: Pharo 12.0 - 64bit (next development version, latest)-01.31FD2FE4A300
Flush D:\Temp\ombu-tests\test-sessions-main\Pharo 12.0 - 64bit (next development version, latest)-01.31FD2FC61E80.ombu
[...]

PR #12953, opened 2023-03-08
Shout: Manage syntax highlight of dynamic array

### LINKED ISSUE #12923: Shout: braces in return do not honor their style
If a dynamic array is in a return, the style of the return is applied and not the style defined for #leftBrace and #rightBrace.

I'll see to fix that soon

### LINKED ISSUE #12941: Shout: Braces styling does not work like parenthesis and block styling
For parenthesis and blocks we can give different styles depending on the imbrication. It is not possible for dynamic arrays.

I propose to manage it as we manage the others.

I'll do a PR when my other PRs on styling will be integrated

jecisc:
The braces have a style declared in Shout but this is never used. 

This PR introduce the highlighting of dynamic array and makes it like parenthesis or block: have some imbrications

Fixes #12941 
Fixes #12923


MarcusDenker:

MarcusDenker (approved):

estebanlm:
this PR seems to have broke something.
1) I have an "unary"  style tag missing (maybe others). 
2) For some reason it breaks styles like my solarized dark: 

Screenshot from 2023-03-09 23-10-33

I am not saying this is a mistake, but I would like to understand the rational in the removals and what is intended to be replaced for.

jecisc:
This was done in a previous PR.

Some styles were removed because shout was never assigning them. Since the Theme system was already changed in P11, we are also cleaning shout and doing the breaking changes now. I was planning to do a PR to your style after the work on shout is done

estebanlm:
but the shout does assign them (even if many times they are not used).
and there is the SpCodeAdapterTest to prove it :)
(in fact I wonder how is possible you had a green test with it)

jecisc:
I’ll try to check today in an old image how it is possible we ended up with “unary” or “binary” in the style because I’ve never seen it assigned anywhere in the styler.

estebanlm:
unary, binary and keyword.
all them can be used (even if they may not be used in the styles present in the image)

PR #12954, opened 2023-03-08
Setting image version in the header of the image file

tesonep:
- Adding accessors to the image version that is stored in the image header.
- Saving the image version that is in SystemVersion current to the image header.
- Adding a Release test to validate if the image is correctly generated
- Adding a call to Smalltalk vm saveImageVersionInImageHeader in the bootstrap process

This PR will be green when the PR of the VM is integrated.
https://github.com/pharo-project/pharo-vm/pull/549


tesonep:
The value is stored as unsigned 16-bit integer in the offset 70 of the header. 
It was an unused field, that in older images should be 0.

MarcusDenker:
not related:

But this is failing and caused by the PR:

- ReleaseTest>>testVersionInImageHeaderIsCorrect

jecisc (approved):
Nice!

Maybe we could also update VirtualMachine>>#parameterLabels

demarey:
Very nice!
I would maybe use the first 3 bits to encode the version of the image numbering. It would start with 1 and could go up to 7 (meaning 6 changes in the numbering of the images, that is already a lot). This would allow change of the numbering in the future.
Another option to change the numbering would be to change the image header version number, and then use the whole field to store the version number.
WDYT?

guillep:
> Very nice! I would maybe use the first 3 bits to encode the version of the image numbering. It would start with 1 and could go up to 7 (meaning 6 changes in the numbering of the images, that is already a lot). This would allow change of the numbering in the future. Another option to change the numbering would be to change the image header version number, and then use the whole field to store the version number. WDYT?

Oh, versionning the versionning. When will it stop? :D

tesonep:
> Nice!
> 
> Maybe we could also update VirtualMachine>>#parameterLabels

I have added it

jecisc:
Awesome! Thank you :3

demarey (approved):

PR #12956, opened 2023-03-08
12955-TFStringType-is-leaking-when-allocates-the-string-for-calling-a-function

### LINKED ISSUE #12955: TFStringType is leaking when allocates the string for calling a function
This demonstrates the memory leak associated with TFStringTypes not being deallocated correctly:
1) Open the OS Process Monitor/Task Manager and note the current memory usage of Pharo
2) Execute:
```smalltalk
| string |
string := String new: 1000 withAll: $=.
10000000 timesRepeat: [LibC resultOfCommand: 'echo "', string, '"'].
```
Observe that the memory of the Pharo OS process steady increases. This simulates the behaviour of my app when it is running. After running continuously for some time memory reaches GBs of size, which then starts to cause problems!

tesonep:
- Improving the free of allocated strings to avoid memory leak.
- Fix 12955


Ducasse:
Thanks pablo!

PR #12958, opened 2023-03-08
add RBCodeSnippetTest>>testCompileWithRequestor

privat:
This adds a new test set for OpalCompiler when a (non-interactive) requestor is given (currently with more than 256 faulty or weird snippets to systematically test various corner cases).

Internal API of the OpalCompiler<->requestor relationship is quite hairy and most mocked method were found by reverse engineering (static and dynamic approaches).
Some code paths might have not been discovered yet :)

Future PR will try to also mock and test interactive requestors, but this is harder since real windows are open and human user interactions are expected without any clear control from the compiler nor the requestor (e.g. the common "unknown variable foo, please correct or cancel" menu).
Moreover, the semantic of the Pharo language change in interactive mode, so it might be another kind of testing challenge.

### DIFF
--- a/src/OpalCompiler-Tests/OCMockRequestor.class.st
+++ b/src/OpalCompiler-Tests/OCMockRequestor.class.st
@@ -1,0 +1,84 @@
+"
+I'm a test utility class that tries to mock a requestor API and just log callbacks
+"
+Class {
+	#name : #OCMockRequestor,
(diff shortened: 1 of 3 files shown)


MarcusDenker:
failure not related:

MarcusDenker (approved):

PR #12959, opened 2023-03-08
Object remove bindingOf: and hasBindingOf:

privat:
Because I just do not like them.
And, whatever they might be useful to should be put in a more specific class.
And, they do not even have any documentation.
And, I had a bad day so I'm grumpy.

More seriously, I'm not really sure if they are only related to some specific code path by requestors of OpalCompiler or if they are used for other concerns, so let's have the CI show us what will break.

### DIFF
--- a/src/CodeImport/FileCompilerRequestor.class.st
+++ b/src/CodeImport/FileCompilerRequestor.class.st
@@ -13,6 +13,12 @@ Class {
 	#category : #'CodeImport-Utilities'
 }
 
+{ #category : #binding }
+FileCompilerRequestor >> bindingOf: aString [
+
+	^ nil
+]
+
 { #category : #accessing }
 FileCompilerRequestor >> contents [
 	^ contents
@@ -36,6 +42,12 @@ FileCompilerRequestor >> fileReference [
 { #category : #accessing }
 FileCompilerRequestor >> fileReference: anObject [
 	fileReference := anObject
+]
+
+{ #category : #binding }
+FileCompilerRequestor >> hasBindingOf: aString [
+
+	^ false
 ]
 
 { #category : #initialization }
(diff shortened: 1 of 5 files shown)


MarcusDenker:
yes... bindingOf:, hasBindingOf: are what used to be there, there was not #lookupVar: outside of the compiler scopes.

Which was not nice... having one way to query for Variables is nicer, but we kind of got stuck in finishing this...


Historically, there where just dictionaries implemented as collections of associations, the term "binding" stands for the association gotten from the dictionary, so that changing the value of that object changes the dictionary:

```
MyGlobal binding
```
returned the association, it can be written to using #value:   

```
MyGlobal binding value: 1.
```

would be the "reflective" version of "MyGlobal := 1". This is where then the idea came that we could just use (or ab-use?) this to turn these binding associations into subclasses of Variable, as the dictionary just needs the key/value API, it does not care that it uses Assocations. And with that be able to have "real" GlobalVariable objects while staying compatible and not introducing many additional objects.

I, too, think that getting rid of #bindingOf: is the right direction.

(I wonder if we got too cautious with cleaning up all this mess...)

MarcusDenker:
Looking at the CI: there are quite some tests failing, so it seems that some places use it still.\

e.g
CoCompletionEngineTest>>testHomeGoesToStartOfLine
OCCompilerTest>>testNotInteractiveSiblingBlocksInstanceVariableShadowing

privat:
Some failed tests are out-of-tree (Spec thing), a clean removing need a coordinated effort :/

My images also failed with really weird issues. E.g. compiled method that were added to classes in a broken way: executable, but not shown by Iceberg on commits, presented on the 4th panel in Calypso but no code shown when selected, and weird exception windows popped semi-randomly.
Tried with fresh image and branching to the PR code. So possibly there is also something.

Another possible 'temporary' approach could be to act on the caller level. For instance, protect MNU with a `respondsTo:`. It's still hackish but move the hack into a controlled space instead of hoarding Object.

privat:
Close it for now. I will think about it

PR #12962, opened 2023-03-09
correcting computation of sets of protocols to remove and add 

### LINKED ISSUE #12961: Method categorization stopped to work
**Bug description**
In latest Pharo 11, the method categorization using writing a protocol name stopped to work. Drag&Drop still works.

**To Reproduce**
1. Go to some method, e.g. `AIBFS>>#end` which is in `accessing` protocol on the instance side
2. click on protocol selection button in the right bottom corner 

image

3. Write a new protocol name in the text field, e.g. `protocol`

image

4. press OK
5. the button in the right-bottom corner shows new protocol name (`protocol`) but no new protocol is created in the class protocols list
6. if you go to another method and return back to it, it shows the old protocol (`accessing`)

**Version information:**
Pharo 11.0.0
Build information: Pharo-11.0.0+build.614.sha.4b6cd19c8978da8465124eb3e42f11badb84c787 (64 Bit)

pavel-krivanek:
revert to version before build 586

fixes #12961

@demarey, please, review in context of the original issue, https://github.com/pharo-project/pharo/pull/12882

### DIFF
--- a/src/Calypso-SystemTools-Core/ClyMethodEditorToolMorph.class.st
+++ b/src/Calypso-SystemTools-Core/ClyMethodEditorToolMorph.class.st
@@ -231,8 +231,8 @@ ClyMethodEditorToolMorph >> tagEditingMethod: aMethod [
 	| existingTags removedTags newTags |
 	self applyChangesBy: [
 		existingTags := aMethod tags reject: [:each | each beginsWith: '*'].
-		newTags := existingTags reject: [ :each | methodTags includes: each ].
-		removedTags := methodTags reject: [ :each | existingTags includes: each ].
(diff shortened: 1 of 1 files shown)


demarey:
Thanks @pavel.
Indeed, there is a problem with the change. I thought there was an error on tag computation but no. I will dig a bit to find the problem with auto-categorization

jecisc (approved):

jecisc:
I checked it and with the PR of Pavel it seems to work fine. 
I cannot reproduce the bug we encountered with Christophe in the original PR. We probably have a broken image after all our changes.
This can be integrated

PR #12967, opened 2023-03-09
New fix code importer

Ducasse:
Now you can drop a file out that uses fluidclass without breaking anything else.
All the three menu items that you can use from dropping the file are working. 
I added some tests. 

What this PR is not doing is fixing the Ring code importer since it is a lot more complex but this is not used in the filein.


MarcusDenker:
there are failing tests, eg in

- SystemDependenciesTest
- MCStReaderTest
- ChunkImportTestCase

Ducasse:
Ok so we should tag it for Pharo 12 because I do not want to fix all these tests now

Ducasse:
If people want to load code in P11 they will have to use the normal class definition.

Issue #10856, opened 2022-02-04
Cleanup Behavior>>#protocols should return a Protocol, not symbol


MarcusDenker:
**Describe the problem**
Ring adds an extension method Behavior>>#protocols which has some problems:

- it sends a message not implemented in Behavior, it should be moved to ClassDestriptiom
- it should not be in the Ring package (it is used a lot everywhere)
- it is not good that it returns a category symbol... on the organizer is implemented to return a *Protocol* object instead.

**Classes involved**
Behavior

```
protocols
	
	^ self organization categories copy
```

ClassOrganization

```
protocols 
	^ self protocolOrganizer protocols.

categories
	^ self protocolOrganizer protocolsNames
```

**Proposal**
make sure the API is unified.

jecisc:
Not for P11 I think

PR #12968, opened 2023-03-09
Rename "category" variables to "protocol"

jecisc:
We have 3 way to call protocols in the image and this is crazy to understand the code.

This is a tiny step toward unification: Renaming some variables "category" to use "protocol"

### DIFF
--- a/src/CodeExport-Traits/TraitedClass.extension.st
+++ b/src/CodeExport-Traits/TraitedClass.extension.st
@@ -1,15 +1,14 @@
 Extension { #name : #TraitedClass }
 
 { #category : #'*CodeExport-Traits' }
-TraitedClass >> fileOutLocalMethodsInCategory: aSymbol on: aFileStream [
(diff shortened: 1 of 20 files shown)


MarcusDenker:
but take care: category is a symbol, while protocol is an instance of protocol (see class Protocol)

This got very confusing by someone adding #protocol to CompiledMethod and Behavior>>#protocols both to return categories... (aka symbols).

So this isn a mess and needs a bit more thinking.

MarcusDenker:
There is https://github.com/pharo-project/pharo/issues/10856

MarcusDenker:
- Does Protocol as an object even make sense?
- There is allProtocol, but we never use it
- if we keep Protocol, we should not have "#protocol" returning just a symbol, because that would be confusing
- what do we get from just renaming the term category to protocol if it stays a symbol?
- What do we even want? e.g. it feels that people would like to be able to tag methods as both 'private' and 'acessing', which neither protocols nor categories can do

jecisc:
I'll close this until we discuss for Pharo 12

PR #12971, opened 2023-03-10
better fix for #12858: check isTempVariable

MarcusDenker:
better fix for #12858: check isTempVariable:
- CopiedLocalVariable is a decorator, it should not be a subclass of TemporaryVariable because it can wrap an argument var, too
- implement isTempVariable to delegate to the original var
- add a method for Shout


MarcusDenker:

jecisc (approved):

PR #12973, opened 2023-03-10
11407-CI-System-variable-environment-on-windows-does-not-use-adequate-encodings

tesonep:
- Correctly getting the Environment entries, as they can be Win32WideStrings.

- Getting the strings with GetEnvironmentStringsW so it is a Win32WideString


tesonep:
Fix #11407

jecisc:
Windows failed and this is the OS impacted by this PR. I'll try to relaunch it

jecisc (approved):
CI pass and test this so it should be good to integrate

PR #12975, opened 2023-03-10
Polymath needs the largest exponant argument. Pharo implements the constant but didn't provide an accessor

jecisc:
Polymath needs the largest exposant argument. Pharo implements the constant but didn't provide an accessor

### DIFF
--- a/src/Math-Operations-Extensions/Float.extension.st
+++ b/src/Math-Operations-Extensions/Float.extension.st
@@ -177,6 +177,13 @@ Float class >> fmax [
 ]
 
 { #category : #'*Math-Operations-Extensions' }
+Float class >> fmaxLn [
+	"Answer the logarithm of the maximum finite floating point value representable."
+
+	^ MaxValLn
+]
+
+{ #category : #'*Math-Operations-Extensions' }
 Float class >> fmin [
 	"Answer minimum positive representable value."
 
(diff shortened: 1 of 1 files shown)


src/Math-Operations-Extensions/Float.extension.st
```diff
@@ -176,6 +176,13 @@ Float class >> fmax [
 	^MaxVal
 ]
 
+{ #category : #'*Math-Operations-Extensions' }
+Float class >> fmaxLn [
```
astares: "fmaxLn" ? What an ugly name.

Java has Float.MAX_VALUE

Why not call the selectors #maxValue, #maxValueLength or similar then and maybe deprecate the old "fxxx" names?

jecisc: Here this is the logarithm on the max value. And the max value is called fmax (I don't like the name personally) so I followed this way of naming things

astares: #maxValueLogarithm then? 

And changing 
 - from "fmax" to #maxValue deprecating the #fmin
 - from "fmin" to #minValue deprecating the #fmin

jecisc: I can name it #maxLog 

For the renaming it'll be in Pharo 12 because we are starting to only do bugfix in this version

astares (approved):

PR #12978, opened 2023-03-13
Improve OCUndeclaredVariableWarning>>declareTempAndPaste

privat:
Low-hanging fruit. A small cleanup of declareTempAndPaste that reduces the hackish level a little

* reclassify and remove hack
* remove unused math (and a parameter)


Ducasse:
Thanks for cleaning this prehistorical code (and it was not the original compiler but a "new one" but with left over of other hacks).

MarcusDenker:
Yes, and the whole idea (to fix the code during exception handling) is just plain bad.

estebanlm:

privat:
@estebanlm yes, it is intended pf P12. Unfortunately, I'm not able to set the pharo12 tag myself.

MarcusDenker (approved):

MarcusDenker (approved):

PR #12979, opened 2023-03-13
make ReparseAfterSourceEditing not resumable

privat:
Low-hanging fruit. A small improvement of ReparseAfterSourceEditing that reduces the hackish level a little.

ReparseAfterSourceEditing is a BAD Notification, as senders expect it to be always caught and never resumed.
Change code to catch error is one of these two expectations are not met.

Maybe change the superclass? But `Error` is not a better choice, since the semantic is not the right one.
What about a new subclass of Exception called `HackishLongJump` for hackish long jumps? :)

### DIFF
--- a/src/OpalCompiler-Core/ReparseAfterSourceEditing.class.st
+++ b/src/OpalCompiler-Core/ReparseAfterSourceEditing.class.st
@@ -18,6 +18,19 @@ ReparseAfterSourceEditing class >> signalWithNewSource: aSourceCode [
 		signal
 ]
 
+{ #category : #handling }
+ReparseAfterSourceEditing >> defaultAction [
+	"Senders do not expect the exception to be resumed"
+
+	self unhandledErrorAction
+]
+
+{ #category : #testing }
+ReparseAfterSourceEditing >> isResumable [
+
+	^false
+]
+
 { #category : #accessing }
 ReparseAfterSourceEditing >> newSource [
 


privat:
First, the CI crashed badly, I'm not sure why. CI should be more robust.

Second, failed tests are related to two out-of-tree tests: `StObjectContextPresenterTest>>#testDoEvaluateDeclareGlobal` and `StPlaygroundPagePresenterTest>>#testDoEvaluateDeclareGlobal`, that lives in another project https://github.com/pharo-spec/NewTools/

Both these tests feel wrong because they hijack internal compiler stuff (and partial, since only global are tested and not any other reparation). Moreover, it seems that on the OC side, the only `ReparseAfterSourceEditing signal` that "could" return (with two different approaches) are the ones in `OCUndeclaredVariableWarning>>#declareGlobal` and `OCUndeclaredVariableWarning>>#substituteVariable:atInterval:` (but in practice never return except in weird tests).

Since this PR is only related to a small cleanup and robustness, maybe it does not worth trying too much :(



<blockquote><div> GitHub</div><div><strong><a href="https://github.com/pharo-spec/NewTools">GitHub - pharo-spec/NewTools: All development tools for Pharo, developed with Spec</a></strong></div><div>All development tools for Pharo, developed with Spec - GitHub - pharo-spec/NewTools: All development tools for Pharo, developed with Spec</div></blockquote>

PR #12981, opened 2023-03-13
Improve ClyClassDefinitionEditorToolMorph>>applyChanges

privat:
Low-hanging fruit. A small cleanup of ClyClassDefinitionEditorToolMorph that reduces the hackish level a little

* Do not reparse the class definition as a method definition in case of failure
* Do not assume a syntax error is related to a source code change


estebanlm:

privat:
CI results are weird, relaunch.

MarcusDenker:
I think the code was there to allow to just remove all text in a class def pane, write a method and accept.

This is indeed strange... I guess this is there "because this always worked", but is it used? To me it makes no sense and the cleanup is nice

MarcusDenker (approved):

PR #12984, opened 2023-03-13
Improve the default comparision value 

jecisc:
Do not use a random precision but use one that proved itself

This value is the default used by Polymath and comes from Numerical Recipes, 3rd edition, Section 5.7, which is pages 229-230

**This is not for Pharo 11 but Pharo 12!**


jecisc:
Bootstrap is timing out with this PR :/ I'll need to investigate why

Issue #12991, opened 2023-03-13
Do we need to parse a method to be able to copy it?


jecisc:
I have a bug currently because the parser has a problem and is invoked by RubSmalltalkEditor>>#copySelection in order to copy the current selection. 

But I wonder if we need to parse the content of a RubSmalltalkEditor in order to be able to copy its content? This seems fragile for a feature that should be easier to implement than to rely on parsing but I might miss something

privat:
I just looked at that because I was curious (while I have many things to do, but eh, we all have different ways to procrastinate). The whole parsing thing is related to the feature I just discovered of doing "smart" escape-aware copy.

Open a playground, and write

```st
'hell''o', 'wor''ld' traceCr
```

Select `hel''lo` without the enclosing quotes. And copy.
Then paste: you get `hell'o` with a single quote. The escaped `'` is unescaped, so smart!
Now select only a double `''` (in either hello or world), copy and paste gives a single one, it works for substrings of literal strings, so smart!
Now select from the `h` to hello up to the `d` of world, then copy-paste... and you get no unescaping but the verbatim text, so smart!

It also works the same for comment because `"hello""world"` is not two comments but a single comment with an escaped double quote.

> Do we need to parse a method to be able to copy it?

For a dumb copy, no need to parse... but for a smart one that is aware of escaping?..
No, don't need to parse either, a scanner seems sufficient.

privat:
Note that if the editor tried to keep a reference to the current ast, there is no need to reparse. But running redundant parsers (style for instance) seems some kind of cultural thing (it's ok I think, it makes jobs...)

jecisc:
> Note that if the editor tried to keep a reference to the current ast, there is no need to reparse. But running redundant parsers (style for instance) seems some kind of cultural thing (it's ok I think, it makes jobs...)

It might not be cultural, but just that people are trying to add new cool features and are not aware of how things could be cleaner because they don't know the system fully ahah
But we can iterate after to improve that

privat:
It's more a joke that something serious, I even added a :+1: to #12963

PR #12992, opened 2023-03-13
Fix infinite loop in faulty literal array parsing

### LINKED ISSUE #12886: Infinite loop when there is insecable spaces in literal array
In the latest Pharo 11 image I have a playgroud with some code and comments. If I try to copy (cmd + c) some text, my image locks on what seems to be an infinite loop. 

I found that the problem comes from the fact that I had an array with insecable spaces in it. Apparently it produces a faulty AST but this seems to end up in a loop?

I know @privat is working on that so it might interest you.

The code that produced the problem:

```st

(housing column: #median_income) groupByBins: { 0 . 1.5 . 3 . 4.5 . 6 . Float infinity } labeled: #(1 2 3 4 5).
```

Try to copy paste that in a playground, then do ctrl + a and ctrl + c. Your image will freeze because of the insecable space between the 4 and the 5 in `#(1 2 3 4 5)`

jecisc:
This is important for P11 because it can freeze the image on the parsing.  The problem was that in the case of faulty literal array we never steped to look at the next character and we looped on the parsing of the faulty character of the faulty array. I added a regression test also

Fixes #12886


src/AST-Core/RBParser.class.st
```diff
@@ -1234,19 +1232,26 @@ RBParser >> parseVariableNode [
 { #category : #'error handling' }
 RBParser >> parserError: aString [
 	"Let the errorBlock try to recover from the error. Answer a ParseNode (possibly an RBParseErrorNode) or signal there is new source"
-	| errorNode errorMessage errorPosition newSource |
+
+	| errorNode errorMessage errorPosition newSource token |
+	token := currentToken.
+	self step.
```
privat: Do not always consume tokens here. parserError means that there is a problem here, not that the current token has to be discarded. We usually want to keep it and instead consider that the current syntactic construction is unfinished.

src/Rubric/RubSmalltalkEditor.class.st
```diff
@@ -192,15 +192,14 @@ RubSmalltalkEditor >> bestNodeInTextArea [
...
-
 	start := self textArea startIndex.
 	stop := self textArea stopIndex.
 
-	[node := RBParser parseMethod: self textArea string onError: [
-		RBParser parseFaultyExpression: self textArea string]
-	] on: Error do: [^nil].
+	[ node := RBParser parseMethod: self textArea string onError: [ :aString | RBParser parseFaultyExpression: aString ] ]
+		on: Error
+		do: [ ^ nil ].
```
privat: this is a bad use of the parser API (that is also bad, and unintuitive).

The correct code might be to replace the whole thing by a simpler

```st
RBParser parseFaultyExpression: self textArea string.
```

without `onError:` nor `on:do:`

privat (changes requested):

src/Rubric/RubSmalltalkEditor.class.st
jecisc: I'll do that thanks

src/AST-Core/RBParser.class.st
jecisc: Do you know in what case I should step or not step?

jecisc: I tried to step in #parseLiteralArrayObject instead but then it does not produce a faulty node anymore. I don't know if this is fine?

privat: never, it's not the responsibility of `parserError` to do this kind of decision, but it is the job of its callers that have a better knowledge of the syntactic situation.

jecisc:
@privat I did a new commit taking into account your review.
The problem is that the generated node is not considered faulty while I think it should.

Do you think this PR can be integrated as is it to fix the infinite loop at least? Also, do you also think the node should be faulty? If yes, do you have an idea on how to manage that? Is it the AST produced that is wrong or should #isFaulty be improved to consider the generated AST of my test to be faulty?

Once thing I find weird also is that the generated AST contains a VariableNode with an UnknownVariable binding for the character 4. That does not seems right to have a variable node in a literal array but maybe we cannot guess better?

privat:
For P11, I think you should drop everything except the fix (and simplification) in RubSmalltalkEditor related to the bad use of the parser API.

The test you added is misleading because the API is bad and should not be used that way, and I'm not sure something meaningful should be expected.
We might just rework the API for P12.

privat:
I can still look the parser code if something could be done for P11

jecisc:
I updated the PR to not change anything about the AST but just improve the way RubSmalltalkEditor gets the AST

src/Rubric/RubSmalltalkEditor.class.st
```diff
@@ -191,16 +191,11 @@ RubSmalltalkEditor >> bestNodeInString: source at: selectionInterval editingMode
...
+	| start stop |
 	start := self textArea startIndex.
 	stop := self textArea stopIndex.
 
-	[node := RBParser parseMethod: self textArea string onError: [
-		RBParser parseFaultyExpression: self textArea string]
-	] on: Error do: [^nil].
-
-	^node bestNodeFor: (start to: stop)
+	^ (RBParser parseFaultyExpression: self textArea string) bestNodeFor: (start to: stop)
```
MarcusDenker: if you look at RubSmalltalkEditor>>#bestNodeInTextAreaWithoutSelection, it uses:

```
	node := self isScripting
		ifTrue: [ RBParser parseFaultyExpression: intervalTrimmedText ]
		ifFalse: [ RBParser parseFaultyMethod: intervalTrimmedText ].
```

privat (approved):

MarcusDenker:
So I wonder if just parsing with #parseFaultyExpression: is good enough: the editor can have either methods or not (in the playground for example). 

This mode is explicit (#isScripting). 

Maybe we should check for #isScripting and use parseFaultyMethod/Expression depending on it

privat:
I think Marcus is right, except that I believe that here the method only care for tokens, and not for trees (c.f. #12991), therefore a faulty expression AST is as good for the task that a non-faulty method (for the same source text).
I approve the new commit

MarcusDenker:
ah, ok. 

I merge it (CI looks good, one timeout that seems not related)

MarcusDenker (approved):

MarcusDenker:
Ah, I was wrong and yes, we have to use #parseFaultyExpression: for both cases.

I will do a new PR

PR #12994, opened 2023-03-14
[CI only] Disable writting in an undeclared variable, to see what happens

privat:
Do not merge. Not for Pharo11, not even for Pharo12 (yet).

This is just to check how far the CI can go if we disable writing to undeclared variables.
I do not expect to go that far, but CI, please, surprise me!


MarcusDenker:
it fails now for global variables.

```
SourceFiles := SourceFileArray new.
```

I guess because with Undeclared, we can assign and later it is fixed of the Variable is added to Smalltalk globals.

privat:
Yeah, I saw that, so I pushed a fix to initialize SourceFiles without relying on undefined variable. (`Smalltalk globals at: #SourceFiles put: SourceFileArray`)
BTW, `at:put:`  is used a few lines above in the same file for the transcript.

There is less than a dozen of globals in the full image that are not classes `Smalltalk environment reject: #isBehavior.`
It's not a few symbols that will be making the rules :)

privat:
> I guess because with Undeclared, we can assign and later it is fixed of the Variable is added to Smalltalk globals.

Just to be sure, can you confirm that, the only way to have a global variable (that are not Behavior or not provided by the bootstrap) is to manually put something in the system dictionary `Smalltalk globals at: #Foo put: bar`

privat:
Close the experiment

PR #12998, opened 2023-03-14
re-Enable Format Incomplete Variables

MarcusDenker:
We turned off incomplete formatting as it is too slow for selectors.

But the preference was used for incomplete Variables, too, where the feature feels quite nice.

This PR enables (and removes the preference) for the variable case.

This is a minimal change for Pharo11.

It does not clean up more, which can be done later (we need to check the subclass in microdown)


jecisc (approved):

jecisc:
I like that it kills some of the tests relying on the applied color and not the symbol associate with the style :)

Those tests are not reliable!

PR #13002, opened 2023-03-14
Simplify Categorizer

jecisc:
In the part the Categorizer was used to manage methods protocols but it is not the case anymore.  This commit contains multiple simplifications of the Categorizer:
- Remove dead code related to protocol
- Remove the concept of Default and Null categories since it was only for protocols 
- Remove some other dead code that was just a copy of the ClassOrginizer but is not used by the Categorizer 

This makes the code muuuuuuch clearer. We can do even better using a dictionary instead of 3 arrays but this is for a next PR ;)


jecisc:
This is broken because Pharo 12 build is not yet working. 
I did a PR to fix the encountered problem we have here

MarcusDenker (approved):

PR #13004, opened 2023-03-14
Fix Pharo 12 loading 

jecisc:
Pharo 12 build is broken because of problems in baselines.

I fixed the TaksIt Baseline to load in all future versions of Pharo.
I also fixed BaselineOfFuelPlatform to work in all future versions of Pharo


MarcusDenker:
The problem is that we load fuel somehow from the fuel repo, while the source is in the pharo repo, too.

Very odd

```

MetacelloNotification: Fetched -> BaselineOfFuelPlatform-cypress.1 --- github://theseion/FuelPlatform:v5/repository --- github://theseion/FuelPlatform:v5/repository
MetacelloNotification: Loading -> BaselineOfFuelPlatform-cypress.1 --- github://theseion/FuelPlatform:v5/repository --- github://theseion/FuelPlatform:v5/repository
MetacelloNotification: Loaded -> BaselineOfFuelPlatform-cypress.1 --- github://theseion/FuelPlatform:v5/repository --- github://theseion/FuelPlatform:v5/repository
Error: Could not determine image version. Using fallback

```

jecisc:
This requires: https://github.com/theseion/FuelPlatform/pull/5

tesonep:
Maybe we should duplicate it in Pharo repository. The same for Pharo 11

jecisc:
We already duplicated apparently but it's still the version in the external project that is loaded. When I'll have some time I'll check to use the version that is in Pharo instead of the remote one.

For P11 are you sure since we are in feature freeze?

tesonep:
If it is loading the master of an external repository or a branch we can
control, we need to change it. We need a reproducible build

jecisc:
I found the problem. BaselineOfFuel>>#fuelPlatform: is explicitly referencing the external version of FuelPlatform. 

I'll do a PR to fix that in P11 then I'll see if I find what to replace for Pharo 12 (IIRC we need a different version of Fuel for each Pharo version?)

jecisc:
https://github.com/pharo-project/pharo/pull/13010

jecisc:
I merge P11 and tried to setup FuelPlatform for P12.

jecisc:
Since it synch with the P11 branch to get the Fuel baseline changes it's hard to review but the two commits that are not in P11 are:
- fd0d81450d970e0ff998dd6e8bfebb17040e6220
- ce9657c1bd9f05f63173cb255e3fda18372c18d2

Issue #13006, opened 2023-03-14
[Meta-issue] undeclared variables


privat:
Undeclared variable is the possibility to have a variable "temporarily" bound to nothing at (method-)compile time, then be automatically repaired when the real variable appears (with or without the need of a compiler according to what kind of variable appears). The thing is complex and low-level, have a lot of drawbacks and surprising semantic and some fragile policy to limit the user ability to shoot themself in the foot.

See discussion in #12935 for more explanation and impacts

## Disclaimer

While UX is important, tooling and presentation are out of scope of the issue.
The focuses are on the specification of the language (including good practices), and compiler code (frontend and backend) to enforce and implement.

## Simple summary

Current behavior (rough explanation because I'm not even sure to not miss things) is that undeclared symbols might be bound to one of those

* Compilation error. Only in interactive code editors. There is still the possibility to keep the variable undeclared for those that starts with an uppercase (heuristic) when the "undeclared" menu appears
* Registered undeclared variables that act like globals (system-wide read-write shared variable) that lead to poor bugs, side effects and programmer incomprehension. They live in a global dictionary called `Undeclared`.
* ~~Unregistered undeclared variables (totally unshared) where each variable is bound to a single usage (i.e. a single read or a single store). Basically, they that act like black hole sinks where you can put everything or infinite source of nil where you can only read nil. This leads to more programmer incomprehension, but at last, have limited side effects :)
  It seems that there exists only for situations for front-end only compilation (highlighting for instance) where you do not want to fill `Undeclared` with garbage, but the API that activate/deactivate this is bad. see #12883 for general rants~~ fixed in #13007
* Workspace variables (only in the workspace, and maybe some DoIt related evaluation). Displayed as red (undeclared) until a first execution, but it's a distinct issue.

The choice of one or the other of these behaviors is mostly circumstantial and depend on the tool, its configuration and the system configuration... (but weather humidity seems ruled out)

## Possible solutions

Not mutually exclusives, and include solution that do not work (or very badly) for reference and discussion.

### Stop registering undeclared variables

  * **will not do** tried in initial version #12935
  * pro: clean and simple
  * con: break package loading and bootstrap.
  * con: break variable removal and reparation.

### Policy based on the capitalization of the name

  * **will not do**
  * Explanation: only capitalized undeclared variables are globally shared (working version of #12935)
  * pro: works
  * con: new random rule
  * con: do not solve the problem (just hide it better)

### Do not share registered declared variables

  * **will not do**
[...]

PR #13007, opened 2023-03-14
Improve faulty compilation - unify handling of undeclared variables

privat:
Note: This is another attempt on  #12935.

Previously, running

```st
{
OpalCompiler new evaluate: 'a := 1. a'.
OpalCompiler new evaluate: 'a'.
OpalCompiler new options: #(+optionSkipSemanticWarnings); evaluate: 'a := 2. a'.
OpalCompiler new options: #(+optionSkipSemanticWarnings); evaluate: 'a'.
}
```

gave #{1 1 nil nil} (and no exception or nothing).

That is not nice.

* bad: undefined variable have a weird semantic (see #13006 , this out of scope of the PR).
* worse: a random unrelated option of the compiler change the semantic.

I have no clue but if I try to guess "why?" such different behavior exists, I suppose it because optionSkipSemanticWarnings might be *circumstantially* used on frontend-only (parse for highlighting for instance) instead of on full compilation.
With optionSkipSemanticWarnings, undeclared variables are not registered, so do not fill the global `Undeclared` with garbage. (but maybe that's not the reason and I'm thinking too much)

Nevertheless, this is bad.
In the PR I propose to

1. unify the behavior (adding optionSkipSemanticWarnings does not change the compiled code)
2. undeclared variable are only registered if a compiled method is produced (by the backend), and again independently of random options.

Now

```st
{
OpalCompiler new evaluate: 'a := 1. a'.
OpalCompiler new evaluate: 'a'.
OpalCompiler new options: #(+optionSkipSemanticWarnings); evaluate: 'a := 2. a'.
OpalCompiler new options: #(+optionSkipSemanticWarnings); evaluate: 'a'.
}
```

give #{1 1 2 2 }, it's still discutable but at least it is consistent (and memory efficient).

the PR will also help future and more aggressive PR by reducing the initial behavioral space.

### DIFF
--- a/src/Kernel-Tests-Extended/CompiledMethodTest.class.st
+++ b/src/Kernel-Tests-Extended/CompiledMethodTest.class.st
@@ -794,6 +794,150 @@ CompiledMethodTest >> testSendsSelector [
 	self deny: ((CompiledCode >> #sendsSelector:) sendsSelector: #doBreakfastForMe)
 ]
 
+{ #category : #'tests - literals' }
+CompiledMethodTest >> testUndeclaredReparationWithClass [
+
+	| method c c2 |
+
+	Smalltalk globals removeKey: #TestUndeclaredVariable ifAbsent: [].
+	Undeclared removeKey: #TestUndeclaredVariable ifAbsent: []. "Because obsolete may remain"
+
+	method := self class compiler compile: 'x ^TestUndeclaredVariable'.
+	self assert: method usesUndeclareds.
+	self assert: (nil executeMethod: method) equals: nil.
+
+	"Adding a class DOES repair"
+	c := Object subclass: #TestUndeclaredVariable.
+	self deny: method usesUndeclareds.
+	self assert: (nil executeMethod: method) equals: c.
+
+	"Removal of class DOES undeclare users"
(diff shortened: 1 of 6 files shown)


privat:
CI logs say `#name:superclassName:traitComposition:classTraitComposition:category:instVarNames:classVarNames:poolDictionaryNames:classInstVarNames:type:comment:commentStamp: was sent to nil`

yeah, I understand that `nil` was unprepared to receive such a message.

The question is more, how this PR can achieve that?

privat:
I get it, the method called is `MCClassDefinition class>>#name:superclassName:traitComposition:classTraitComposition:category:instVarNames:classVarNames:poolDictionaryNames:classInstVarNames:type:comment:commentStamp:` (what a nice selector!) except that MCClassDefinition is still nil, therefore something is wrong. :(

privat:
I hope I found the bug. I had to dig in the bootstrap and semi-blindly isolate faulty situations with handwritten `.st` files and inferring how damn those pesky `!` interleaving method definitions with expressions in `.st` files works.

privat:
Tests are ok minus a unrelated windows test

MarcusDenker (approved):

PR #13008, opened 2023-03-14
Escape '|' and '.' characters to prevent regex interpretation

### LINKED ISSUE #13001: 'Find & Replace' broken for some characters
**Bug description**

The Find & Replace is broken for some characters like `|` or `.`.
It could be related to a bad interpretation of regex, since these characters are meaningful in regex.

Note that escaping the characters `\|` or `\.` or protecting in character classes `[|]` or `[.]` does achieve nothing/
Also note that checking or unchecking the "Regular expression" do not seem to have any effect.

**To Reproduce**

Search for `|` or `.` or any string containing them in a text that have them (or not).

Capture d’écran du 2023-03-14 08-34-21

pavel-krivanek:
fixes #13001

### DIFF
--- a/src/Text-Edition/FindReplaceService.class.st
+++ b/src/Text-Edition/FindReplaceService.class.st
@@ -38,7 +38,7 @@ FindReplaceService >> caseSensitive: aBoolean [
 { #category : #accessing }
 FindReplaceService >> convertedFindString [
 	| specials |
-	specials := '^$:\+*[]()?'.
+	specials := '|^$:\+*[]()?.'.
 	^String
 		streamContents: [:s | self findString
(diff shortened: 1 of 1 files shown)


privat:
Good catch.
You might want to add `{` and `}` to the special characters, and I'm not sure why `:` in the list

pavel-krivanek:
updated

```smalltalk
':' asRegex matches: ':'
```
fails with RegexSyntaxError:  no terminating ":"` while 
```smalltalk
'\:' asRegex matches: ':'
``` 
not so it has some reason ;-)

MarcusDenker (approved):

PR #13010, opened 2023-03-15
Use internal FuelPlatform

jecisc:
This commit does:
- Use the internal pharo FuelPlatform instead of the external one
- Simplify the baseline to keep only what is managed by Pharo current version
- Sync with Fuel (There was only a return extraction)
- Replace deprecated calls in UFFI (it happened automatically and I'm too lazy to do a PR just for that)

In FuelPlatform we still explicitly reference Pharo11 tag because IIRC we need to update FuelPlatform for each version of Pharo so it should break in Pharo 12.  This does not fix Pharo 12 but at least it will make Pharo 11 more reproduceable.

### DIFF
--- a/src/BaselineOfFuel/BaselineOfFuel.class.st
+++ b/src/BaselineOfFuel/BaselineOfFuel.class.st
@@ -19,86 +19,20 @@ BaselineOfFuel >> addPostloadHacks [
 
 { #category : #baselines }
 BaselineOfFuel >> baseline: spec [
-  <baseline>
-  spec
-	for: #common
-	do: [
+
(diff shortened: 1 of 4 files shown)


jecisc:
When this is integrated, I'll get my Pharo 12 branch, merge it and see if I can add the FuelPlatform for Pharo 12 myself (I don't know the internal of Fuel but if it's just doing an equivalent of what is in the image it should be fine)

jecisc:
Tests passed. If it's integrated, I'll take a look to update Fuel in Pharo 12 also after merging Pharo11 in Pharo12

PR #13012, opened 2023-03-15
Fix style of pragmas

jecisc:
The opening of a pragma did not had the style of the syntax highlighter of the pragma. This fixes it.


src/Shout/SHRBTextStyler.class.st
```diff
@@ -1077,7 +1077,9 @@ SHRBTextStyler >> visitParseErrorNode: anErrorNode [
 { #category : #visiting }
 SHRBTextStyler >> visitPragmaNode: aPragmaNode [
 
-	aPragmaNode selectorParts with: aPragmaNode keywordsPositions do: [ :keyword :position | self addStyle: #pragma from: position to: position + keyword size ].
+	aPragmaNode selectorParts
+		with: aPragmaNode keywordsPositions
+		do: [ :keyword :position | self addStyle: #pragma from: position - 1 to: position + keyword size ].
```
privat: The -1 seems off. Is it because the given position is wrong (it should not be)? or is it because you want to style `<` (but this will fail to highlight `<    pragmaaaaaa     >`)? or something else and I need mode coffee?

The following methods on aPragmaNode might help:

* `left` `right` should correspond to the position oh opening and closing `<` `>`
* `keywordsIntervals` seems to exist to avoid doing math to compute the full span of each keyword

jecisc: Oh thanks! 

I'll add tests with those cases :)

jecisc:
@privat does that seems better for you? :)

src/Shout/SHRBTextStyler.class.st
```diff
@@ -1076,8 +1076,15 @@ SHRBTextStyler >> visitParseErrorNode: anErrorNode [
 
 { #category : #visiting }
 SHRBTextStyler >> visitPragmaNode: aPragmaNode [
+	"Lets imagin this pragma: `<  test: #toto test2: 2 >`
 
-	aPragmaNode selectorParts with: aPragmaNode keywordsPositions do: [ :keyword :position | self addStyle: #pragma from: position to: position + keyword size ].
+	We want to highlight the `<`, the `test:`, the `test2:` and the `>` but we do not want to highlight the arguments."
+
+	self addStyle: #pragma from: aPragmaNode left to: aPragmaNode left.
+	aPragmaNode selectorParts
```
privat: Maybe something like this could simplify more (untested)

`aPragmaNode keywordsIntervals do: [ : i | self addStyle: #pragma from: i start to: i stop ]`

jecisc: Test is green with it, I added that. Thanks

privat (approved):

MarcusDenker (approved):

PR #13013, opened 2023-03-15
"Close" instead of "Cancel" in the button label of the Find & Replace…

### LINKED ISSUE #12985: Find and Replace "cancel" button should be close
In the Find and Replace dialog we have a "cancel" button. But it does not cancel the actions we did already. It's just here to close the window. 

IMO it should be renamed "Close"

pavel-krivanek:
… dialog.

fixes #12985

### DIFF
--- a/src/Rubric-SpecFindReplaceDialog/SpRubFindReplaceDialog.class.st
+++ b/src/Rubric-SpecFindReplaceDialog/SpRubFindReplaceDialog.class.st
@@ -15,7 +15,7 @@ Class {
 		'entireCheckBox',
 		'replaceButton',
 		'replaceAllButton',
-		'cancelButton',
+		'closeButton',
 		'replaceTextInput',
 		'replaceLabel'
 	],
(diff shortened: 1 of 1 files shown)


privat (approved):

src/Rubric-SpecFindReplaceDialog/SpRubFindReplaceDialog.class.st
```diff
@@ -105,7 +105,7 @@ SpRubFindReplaceDialog >> initializeDialogWindow: aDialogWindowPresenter [
 			self replace ].
 	replaceAllButton := aDialogWindowPresenter addButton: 'Replace All' do: [ :presenter |
 			self replaceAll ].
-	cancelButton := aDialogWindowPresenter addButton: 'Cancel' do: [ :presenter |
+	cancelButton := aDialogWindowPresenter addButton: 'Close' do: [ :presenter |
```
gcotelli: Maybe rename also the button from `cancelButton` to `closeButton`

pavel-krivanek: yeah, thanks! Done

MarcusDenker (approved):

PR #13016, opened 2023-03-15
move bootstrap/src into main src

privat:
I'm not sure why those files are isolated. They need more love.


jecisc:
Hi Jean, 

I already tried to do that but failed. 

The bootstrap scripts need to be updated to point to the new location but then the bootstrap fails because the original image starting the bootstrap already registered the baseline location to be at a different place and it create a conflict. 

When I was going to have the time I planned to update the original image but a first step can be to empty its metacello registry in the bootstrap script to make it work. 

I can check to finish what I started soon if it can help you

jecisc:
And yes, my goal was to give them more love too ahah (But only 24h in a day)

pavel-krivanek:
you should consider moving to Mercury :)

privat:
I thought that symlinks will do the trick :(
The failed message is weird anyway `Error: Name not found: coreTests`

jecisc:
This is because P12 branch is not working yet. 
Well... Now it is working but just because Pablo merged my PR fixing the CI of P12 a few seconds ago!

If you merge Pharo 12 (after pulling) in your branches for P12, they should work now

privat:
There will be a bottleneck for P12 testing, this PR is clearly not urgent.
But the beginning of the cycle is the right moment to hack and break the bootstrap anyway :)

jecisc:
It was also on my todo list so we can do that together :)

privat:
It worked.
I did not really expect that.

now what?

jecisc:
If you change the base to be P12 and it does not break the CI it can be merged and it would be a great step to improve the bootstrap!

privat:
tests are still ok, windows is still drunk or something

jecisc:
I've never seen this failure. I hope it's not a new random failure :(

What you did is awesome because now we'll be able to easily update the bootstrap code and remove the deprecations and all :)

privat:
> I've never seen this failure. I hope it's not a new random failure :(

I don't really appreciate windows that much, maybe it's reciprocal.

> What you did is awesome because now we'll be able to easily update the bootstrap code and remove the deprecations and all :)

Thanks ! :blush:
Some other bootstrap experts should look at it.

Ducasse:
I'm slowly starting to look at the bootstrap and I identified some deadcode or logic
- make files
- Old classes

```
prepareEnvironmentForHermes
	ringEnvironment := self originRepository asRing2EnvironmentWith: self allPackagesForHermes.
	ringEnvironment fixProtoObjectClassSuperclass.
	ringEnvironment addGlobalsNamed: #(Smalltalk Transcript FileStream MacRomanTextConverter ChangeSet SourceFiles Processor Display Sensor UTF8TextConverter SystemOrganization Undeclared TextConstants).
	ringEnvironment clean

FileStream -> empty
Sensor -> does not exit
UTF8TextConverter -> does not exit anymore
```

jecisc:
Don’t hesitate to open issues :)
Once the bootstrap will be more accessible by moving the sources, then we’ll be able to fix those issues more easily :)

jecisc (approved):

Ducasse:
I think that where there to clearly identify the bootstrap which is basically not the code that we browse using Pharo but only a part? No?

PR #13023, opened 2023-03-16
Use dictionary in SystemOrganizer

### LINKED ISSUE #12602: Enhance  Categorizer/SystemOrganizer  by switching from three Arrays to an OrderedDictionary
**Describe the request**
I have a package with ~580 classes in 78 categories.  In Pharo 10, for a class in that package, opening a browser takes 1.5 seconds,  changing from flat to hierarchical view takes 3.6 seconds, according to TimeProfiler.  

Doing the same browsing operations in e.g. `System-Support-Image` package is instantaneous.  

The TimeProfiler shows the majority of the time is spent in the `Categorizer>>listAtCategoryNamed:` method, which does an `indexOf:ifAbsent:` lookup.  

From the looks of the Categorizer class it seems like there is a lot of accounting code for keeping track of where each category's class lists begin and end which could just be a simple array as the value in the dictionary.   Key = category name, value = class list

To see for yourself run: 
```
RPackageOrganizer default createPackageNamed: 'Foo'.
1 to:600 do:[:ind | |classNameSuffix categoryNameSuffix |
	 classNameSuffix :=ind asString. 
	categoryNameSuffix :=(ind//8) asString.
	Object subclass: ('Foo' , classNameSuffix) asSymbol
	instanceVariableNames: ''
	classVariableNames: ''
	package: 'Foo-Data-',categoryNameSuffix  ].
```

Then try to browse one of those classes and also switch from flat to hierarchical view. 

**Expected behavior**
no degradation in speed of browsing or class display as packages increase in size...... I guess.  

**Expected development cost**
Less than half a day?  I don't know how to swap in a different version of the SystemOrganizer or what relies on it if anything now. 

**Version information:**
 - OS: macos m1 arm
 - Pharo Version 10

jecisc:
The system organizer was introduced before the dictionaries. 

Its management is unreadable because it emulate the behavior of a dictionary with 3 arrays.
This PR removes that and is using a dictionary instead. The code is much more readable (really check it out you'll seen) and is way faster. 

The hierarchy view with 600 classes in calypso went from 3.5sec to show to 350ms.

Fixes #12602

I did another change. In the past, when you removed an element forma category that had only this element, the category was kept. And then the image cleanup was asking to remove those from time to time.
Now, when we remove the last element of a category, we remove it directly and I removed the image cleanup since it will now never happen anymore to have empty categories in the SystemOrganizer.


jecisc:
Error seems related but I have no idea of the origin of the problem. 
I got the artifiact of Jenkins and RPackage is able to find _UnpackagedPackage in this image, so I guess that the error is produced by another test that changes the state of the image, but... Which one?

jecisc:
The refactoring tests are taking 30sec less to run now :)

jecisc:
The refactoring tests are taking 30sec less to run now :)

MarcusDenker (approved):

PR #13026, opened 2023-03-16
Improve #12992: Copying strings in editor

### LINKED ISSUE #13027: Syntax Error while copying code
**Bug description**
When copying from a code editor, if the code is faulty, the system shows a Syntax Error dialog instead of putting the selected text in the clipboard.
is it a bug or a feature? either way, it's annoying.

**To Reproduce**
1. Any method in the system browser.
2. In the  method code area, write any text that don't compile, such as: '1 this should not be accepted'.
3. Select the text, or part of the text and try to copy via hotkeys or context menu.
4. See error: A Syntax Error dialog appears.

**Expected behavior**
The selected code is copied and not judged.

**Version information:**
 - Pharo 11

**Expected development cost**
I'm pushing a PR now for this. However, if that behavior was expected or linked to other smart feature, it will probably be broken, so please check.

MarcusDenker:
we need to use parseFaulty for the method, too

### DIFF
--- a/src/Rubric/RubSmalltalkEditor.class.st
+++ b/src/Rubric/RubSmalltalkEditor.class.st
@@ -194,7 +194,7 @@ RubSmalltalkEditor >> bestNodeInTextArea [
 	"We build the AST (which can be faulty when we are scripting) then ask the best node for the interval."
 	^ (self isScripting
 		   ifTrue: [ RBParser parseFaultyExpression: self textArea string ]
-		   ifFalse: [ RBParser parseMethod: self textArea string ]) bestNodeFor: (self textArea startIndex to: self textArea stopIndex)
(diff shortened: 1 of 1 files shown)


maxwills:
Whoops, I did a PR related to this. I guess the change overlooks some behavior.

> we need to use parseFaulty for the method, too

Can you elaborate a little more please? I am not sure why it is necessary any parsing, when I just want to copy what is selected 
(Feel free to just close the deny the PR if there is something missing)

MarcusDenker:
There is a discussion in https://github.com/pharo-project/pharo/issues/12991 

\

privat:
I installed P12 and was bitten by it. Thanks, @MarcusDenker

privat (approved):

PR #13028, opened 2023-03-16
fixes and closes #13027 bug when copying faulty code

### LINKED ISSUE #13027: Syntax Error while copying code

maxwills:
fixes and closes #13027 bug when copying faulty code

### DIFF
--- a/src/Rubric/RubSmalltalkEditor.class.st
+++ b/src/Rubric/RubSmalltalkEditor.class.st
@@ -1080,6 +1080,19 @@ RubSmalltalkEditor >> selectedSymbol [
 	^ nil
 ]
 
+{ #category : #'menu messages' }
+RubSmalltalkEditor >> selectedText [
+	"Answer the text contained in selection interval"
+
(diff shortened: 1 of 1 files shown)


MarcusDenker:
This was fixed by https://github.com/pharo-project/pharo/pull/13026

PR #13029, opened 2023-03-16
Faulty compiler: new CodeError class

privat:
This PR adds a new error super class for syntactic and semantic errors: `CodeError`.

The changes are rather brutal since:

* SyntaxErrorNotification become an Error (instead of a notification). The name stays unchanged for now.
* OCSemanticError is stripped of its responsibility to report error on the UI (yes, it was this bad).

Note: `OCSemanticWarning`, especially `OCUndeclaredVariableWarning` are let untouched (yet!)

Let's see if CI thinks this to too much change to handle...


privat:
tests are ok, except windows. But I added more commits for some related cleanup

MarcusDenker (approved):

PR #13036, opened 2023-03-17
Faulty code - test Monticello

privat:
This teaches Monticello to test method loading with numerous faulty code.

I also added a warning and a fix in Monticello if for some reason there is a mismatch between the declared selector and the real one in the method definition.


privat:
```
Error: Package Refactoring-Core-Support depends on the following classes:
  RBCodeSnippetTest
  RBCodeSnippetTest>>#testMonticello
```

no it don't, silly. :/

privat:
Package dependency in Pharo is garbage. I will just stop using extension methods. :rage:

MarcusDenker (approved):

PR #13037, opened 2023-03-17
Faulty code: CodeError are resumable

privat:
This PR feels like nothing, but is an achievement based on numerous previous PR on cleaning, improvements and bugfixes.

It shows that error management in compilation starts to become consistent.

It also means that when a debugger windows pop up in syntax or semantic error, user can "resume" and a faulty method might be produced.

e.g. try the following in the playground

```st
Smalltalk compiler evaluate: '1+'
```

You will get a syntax error. Then click on "▶️ proceed", and the code will be executed anyway (you proceeded, it's your fault) and thus you will a runtime error (whatever as hard you try, a thing is missing, and no decent value can be produced).

Note: I expect that some things might break, although faulty AST and faulty CompiledMethods are tested a lot, but I likely missed some code paths or usage (especially in GUI)

### DIFF
--- a/src/AST-Core/CodeError.class.st
+++ b/src/AST-Core/CodeError.class.st
@@ -34,6 +34,13 @@ CodeError >> errorMessage [
 	^ messageText
 ]
 
+{ #category : #'private - testing' }
+CodeError >> isResumable [
+
+	"When resumed, faulty AST are produced"
+	^true
+]
+
 { #category : #accessing }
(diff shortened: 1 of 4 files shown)


privat:
Note: @MarcusDenker this might help you to import bad or old code :)

MarcusDenker:
Nice!

MarcusDenker (approved):

PR #13040, opened 2023-03-17
Unify package unregistration

### LINKED ISSUE #13039: RPackageOrganizer>>#basicUnregisterPackage: and #basicUnregisterPackageNamed: have different behaviors
When looking at those methods we can see some duplication and that they do not do exactly the same thing with the package to unregister.

I have the impression that basicUnregisterPackage: should just be:

```st
basicUnregisterPackage: aPackage 
    ^ self basicUnregisterPackageNamed: aPackage name
```

jecisc:
We had basicUnregisterPackageNamed: and basicUnregisterPackage: to unregister packages but basicUnregisterPackageNamed: was missing two things:
- Logging multiple remove
- Unregistring MC working copy

I moved all the behavior in basicUnregisterPackageNamed: and removed the other method after fixing the two senders. I'm not sure of the consequences but I'll let the CI show me if this is ok or not.

Fixes #13039

### DIFF
--- a/src/RPackage-Core/RPackageOrganizer.class.st
+++ b/src/RPackage-Core/RPackageOrganizer.class.st
@@ -282,31 +282,20 @@ RPackageOrganizer >> basicRegisterPackage: aPackage [
 ]
 
 { #category : #'private - registration' }
-RPackageOrganizer >> basicUnregisterPackage: aPackage [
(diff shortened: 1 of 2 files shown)


privat (approved):
I know nothing about that, but I saw nothing weird.

MarcusDenker (approved):

PR #13042, opened 2023-03-18
Faulty parsing: change logic of errorBlock

privat:
This is a rewrite of #12912. I didn't like how it was done (but you might want to re-read it to compare).

Here is the new proposal, less radical but much cleaner than the current state anyway (so good enough for now).

* `RBParser` gets a `isFaulty` flag used for all "faulty" parsing (no more weird `RBParser class>>errorNodeBlock`).
  For contxt, faulty mode produces `RBParseErrorNode`, then continue the parsing in degraded mode to produce faulty AST. In 'non-faulty' mode, exceptions are signaled (that just stops the parsing).
* `errorBlock` is kept (for the moment?) because it is used a lot, but is only for error notification (and most of the time non-local return). Reparation was broken by design and never performed by client at this level (not even by a test!!!). Therefore,  reparation is no more possible (and parseErrorNode is now private).
* SyntaxErrorNotification is therefore signaled if neither isFaulty or errorBlock wanted to do anything
* If resumed, `RBParseErrorNode` is produced (as in isFaulty mode). This was in the previous PR.

Next PR will be likely about cleaning up more parser error things (messages and positions).


privat:
note: OpalCompiler uses the name `failBlock` and `RBParser` uses `errorBlock`. Which one do you prefer?

MarcusDenker:
I somehow prefer failBlock

MarcusDenker (approved):
getting rid of the errorNodeBlock mechanism is very good. 

For the errorBlock, I somehow prefer failBlock, but that is I guess because the compiler API uses that term.

And yes, it would be nice to get rid of it both on the level of the  compiler and the parser

PR #13044, opened 2023-03-18
Faulty parser - better error position and messages

privat:
In the parser, the management of faulty parsing with `RBParseErrorNode` and especially with `RBEnglobingErrorNode` was hacked in parallel of the existing Exception and errorBlock error handling, including redundancy (and inconsistency) on error messages and error positions.

This PR tries to simplify the mess a little, especially:

* new `errorPostion` in `RBParseErrorNode` (and `RBEnglobingErrorNode`) to pinpoint the exact location, you should put the cursor or popup message (or insert `->` messages in source code if it's your kink).
* new `RBEnglobingErrorNode class>>#from:contents:` to transform a `RBParseErrorNode` into a `RBEnglobingErrorNode` (or a subclass) reusing the information and adapting the span of the error (that could also be adapted further if needed).


privat:
unrelated windows failure

MarcusDenker (approved):

PR #13045, opened 2023-03-18
Compiler: remove getSourceFromRequestorSelection

privat:
For some reason, the compiler did callback the UI editor to get the selection string to evaluate, misunderstanding that it should already have it. If editors give garbage to the compiler, it's not the responsibility of the compiler to repair. (I am not able rightly to apprehend the kind of confusion of ideas that could provoke such a design).

Moreover, kill selectionAsStream in editors as a string is OK for the compiler, no need of a stream.


privat:
unrelated windows error

MarcusDenker (approved):

MarcusDenker:
Yes... I wonder why this was there. The tools should give the compiler the right code from the start, and I think they always do.

PR #13046, opened 2023-03-18
Some cleanings of ClassDescription

jecisc:
- Rename some "Category" temporary variables into "protocolName"
- Make sure compilation methods use the same variable names on their different versions (for example always use #code and not sometimes #code and sometimes #text)
- Improve some variables names (like sel => selector)
- Remove a method that was just calling another method with a close name
- Remove conditional in ClassDescription>>#compile:classified:notifying: because the change stamps will not be used later if acceptsLoggingOfCompilation returns false anyway. No need to make the code more complicated

### DIFF
--- a/src/Kernel/ClassDescription.class.st
+++ b/src/Kernel/ClassDescription.class.st
@@ -27,12 +27,10 @@ ClassDescription >> acceptsLoggingOfCompilation [
 ]
 
 { #category : #'accessing - method dictionary' }
-ClassDescription >> addAndClassifySelector: selector withMethod: compiledMethod inProtocol: category [
+ClassDescription >> addAndClassifySelector: selector withMethod: compiledMethod inProtocol: protocol [
 	| priorMethodOrNil priorOriginOrNil oldProtocol newProtocol |
 
-	priorMethodOrNil := self
-		compiledMethodAt: selector
-		ifAbsent: [ nil ].
+	priorMethodOrNil := self compiledMethodAt: selector ifAbsent: [ nil ].
 	priorMethodOrNil ifNotNil: [ priorOriginOrNil := priorMethodOrNil origin ].
 
 	self addSelectorSilently: selector withMethod: compiledMethod.
@@ -41,9 +39,9 @@ ClassDescription >> addAndClassifySelector: selector withMethod: compiledMethod 
 	SystemAnnouncer uniqueInstance suspendAllWhile: [
 		self organization
 			classify: selector
-			under: (category = Protocol unclassified
+			under: (protocol = Protocol unclassified
 				ifTrue: [ oldProtocol ]
-				ifFalse: [ category ]) ].
+				ifFalse: [ protocol ]) ].
 	newProtocol := self organization categoryOfElement: selector.
 
 	self isAnonymous ifTrue: [ ^ self ].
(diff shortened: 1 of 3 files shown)


src/Monticello/MethodAddition.class.st
```diff
@@ -76,7 +76,7 @@ MethodAddition >> notifyObservers [
 			priorCategoryOrNil = category ifFalse: [
        			SystemAnnouncer uniqueInstance methodRecategorized: compiledMethod oldProtocol: priorCategoryOrNil ] ].
 	"The following code doesn't seem to do anything."
```
privat, line 78: Maybe remove it if it doesn't do anything?

src/Kernel/ClassDescription.class.st
```diff
@@ -373,186 +371,149 @@ ClassDescription >> commentStamp: changeStamp [
...
 
-	^ self compileSilently: code classified: category notifying: nil
+	^ self compileSilently: code classified: protocolName notifying: nil
 ]
 
 { #category : #compiling }
-ClassDescription >> compileSilently: code classified: category notifying: requestor [
-	"Compile the code and classify the resulting method in the given category, leaving no trail in the system log, nor in any change set, nor in the 'recent submissions' list. This should only be used when you know for sure that the compilation will succeed."
+ClassDescription >> compileSilently: code classified: protocolName notifying: requestor [
+	"Compile the code and classify the resulting method in the given protocol, leaving no trail in the system log, nor in any change set, nor in the 'recent submissions' list. This should only be used when you know for sure that the compilation will succeed."
```
privat, line 440: The original comment makes no sense. The previsible success of a compilation is not related to leaving a trace.

src/Kernel/ClassDescription.class.st
```diff
@@ -373,186 +371,149 @@ ClassDescription >> commentStamp: changeStamp [
...
+	self copyCategory: protocolName from: class classified: protocolName
 ]
 
 { #category : #copying }
-ClassDescription >> copyCategory: cat from: aClass classified: newCat [
-	"Specify that one of the categories of messages for the receiver is the
-	third argument, newCat. Copy each message found in the category cat in
-	class aClass into this new category."
+ClassDescription >> copyCategory: protocolName from: aClass classified: newProtocolName [
+	"Specify that one of the protocol of messages for the receiver is the
```
privat: 'protocol of messages ' sounds weird.

privat: I don't understand the whole sentence in fact.

src/RPackage-Core/ClassDescription.extension.st
```diff
@@ -2,9 +2,9 @@ Extension { #name : #ClassDescription }
 
 { #category : #'*RPackage-Core' }
 ClassDescription >> compileSilently: code [
-	"Compile the code and classify the resulting method in the given category, leaving no trail in the system log, nor in any change set, nor in the 'recent submissions' list. This should only be used when you know for sure that the compilation will succeed."
```
privat, line 5: These should not be extension methods. It's just variations on existing selector familes

privat (changes requested):
You should use `source` or `sourceCode` (instead of `code ` or `text`) for better consistency with other parts of Pharo (compiler, CompiledMethod, search in source UI, the source file sibling of the image file, etc.)

src/Monticello/MethodAddition.class.st
jecisc: I prefer not to touch that in that PR because I'm not sure it does nothing ahah

src/Kernel/ClassDescription.class.st
jecisc: I changed the comment

src/RPackage-Core/ClassDescription.extension.st
jecisc: Done

jecisc:
CI error not related

privat (approved):

MarcusDenker (approved):

PR #13047, opened 2023-03-18
Compiler: stop playing notification pingpong

privat:
Big step, but I tried to make it as small and simple as possible for testability and reviewability (especially commit by commit).

Up to now, automatic (mandatory!) code reparation on compilation error (the "undefined variable foo, please do something or cancel" menu) is managed through a complex exceptions mess and badly designed responsibility assignment.
Unfortunately, getting rid of it is less easy as it seems (and it seems hard!)

This PR does the main part of the cleanup:

* UI related stuff still in the compiler is grouped and moved to a simple single dedicated method `parseForRequestor` (still bad, but much less bad)
* `OCUndeclaredVariableWarning` stops opening the menu by itself as default action. All that remains is the console warning for CI.
* `parseForRequestor` catches `OCUndeclaredVariableWarning` instead of `ReparseAfterSourceEditing`, open the menu, and retry the parsing all by itself (previously it was done by `OCUndeclaredVariableWarning`)
* `ReparseAfterSourceEditing` is removed
* I repeat, `ReparseAfterSourceEditing` is removed
* Tests are added (there was none, except a very bad and limited one)

What is not done and let for future PR

* move the menu UI and reparation logic outside `OCUndeclaredVariableWarning`
* simplify exception hierarchy `OCSemanticWarning`
* kill `parseForRequestor` for good


Ducasse:
Great. You see this is a legacy from the "interactive" view of the world. 
I do not blame them. They were inventing everything so stepping back and thinking in terms of layers and minimizing the dependency and strange logical flow in a context where you have 64kb.... 
Now this is our responsibility to clean and I love your effort.

privat:
Tests failed because of a single test in NewTools. Not nice.
The test tries to intercept some exceptions but because the ping-pong is removed, then the menu is open and the test fails.

I do not want to add a buggy workaround here in compiler (like identifying the specific requestor or test) so I should fix the test in the other repository.

Ducasse:
Sure! We should target all green tests else it will block us for no reason.

privat:
Test does not seem really fixable now, so I propose to simply skip it https://github.com/pharo-spec/NewTools/pull/492.
We will re-enable it (or reimplement it) once the compiler cleaning is done and its API stabilized.

MarcusDenker (approved):

MarcusDenker:
Very nice... such a huge step to be able to remove ReparseAfterSourceEditing!

Issue #12285, opened 2023-01-08
GitHub Rate limiting for CI (yet again)


astares:
Looks like GitHub Rate limiting for CI problem again:
see https://github.com/pharo-project/pharo/branches

We already had this in the past: 
- see #11222
- see #10926

astares:
[https://ci.inria.fr/pharo-ci-jenkins2/job/Test%20pending%20pull%20request%20and%20branch%20Pipeline/job/Pharo11/423/display/redirect](https://ci.inria.fr/pharo-ci-jenkins2/job/Test%20pending%20pull%20request%20and%20branch%20Pipeline/job/Pharo11/423/display/redirect)

![image](https://user-images.githubusercontent.com/5980033/211222025-69c498fb-ca68-48d8-82e2-2a5725ffb52e.png)

So this is CI machine of INRIA

![image](https://user-images.githubusercontent.com/5980033/211222270-530888b4-bfbf-4c74-ae94-439e008da6ac.png)

astares:
See more details in
[https://docs.github.com/en/rest/overview/resources-in-the-rest-api?apiVersion=2022-11-28#rate-limiting](https://docs.github.com/en/rest/overview/resources-in-the-rest-api?apiVersion=2022-11-28#rate-limiting)

astares:
One can verify the rate limit using
```
curl -i https://api.github.com/users/pharo-project
```
which gives: 

![image](https://user-images.githubusercontent.com/5980033/211222524-3baf308b-1c2f-4363-884c-3730ac9f1dca.png)

astares:
So according to [documentation](https://docs.github.com/en/rest/overview/resources-in-the-rest-api?apiVersion=2022-11-28#rate-limiting) this is due to doing anonymous requests (=unauthenticated)
```
"For unauthenticated requests, the rate limit allows for up to **60** requests per hour."
```
which would fit what "pharo-project" has at the moment (see curl example with the limit of 60 requests per hour)

If we would use a **dedicated authenticated user** then then we could go up to **5000** request per hour:
```
User-to-server requests are limited to **5,000** requests per hour and per authenticated user.
```

@MarcusDenker This ticket was closed (as you marked this new case as duplicate from an already closed old one) 

But the problem persists when too many requests hit CI. Is there a way we can work with an authenticated user in the CI?

privat:
[...]

PR #13048, opened 2023-03-19
Compiler get isFaulty

privat:
Parsing and compiling in a "faulty" way is not really a preference toggle, it's mainly the responsibility of client to state what they want and what kind of results they are ready to deal with.

So this PR deprecate the options `#optionParseErrorsNonInteractiveOnly` `#optionParseErrors` `#optionSkipSemanticWarnings` and propose a single compilation flag.
This simplifies the code of clients and the internal logic of the compiler.

Also, I did not like the `options:` way because the syntax is cumbersome, long names are easy to misspell, and spelling errors are silently ignored.

I use a flag named `isFaulty`, that I also exposed in the parser #13042. But now I have a second thought. *faulty* means *containing a fault or defect; imperfect or defective*, that make sense for an AST or a method, but not really for the parser or the compiler. Here are some alternative names. I will change the name to the one you prefer.

* `acceptFaulty`
* `tolerateFaulty`
* `supportFaulty`
* `permitFaulty`
* something else?

### DIFF
--- a/src/AST-Core/CompiledMethod.extension.st
+++ b/src/AST-Core/CompiledMethod.extension.st
@@ -27,7 +27,7 @@ CompiledMethod >> parseTree [
 	ast := self methodClass compiler
 		source: self sourceCode;
 		failBlock: [^ self decompile ];
-		options: #( + #optionParseErrors #optionSkipSemanticWarnings ) ;
+		isFaulty: true;
 		class: self methodClass;
 		parse.
 	ast compilationContext compiledMethod: self.
--- a/src/Calypso-SystemTools-Core/ClyMethodEditorToolMorph.class.st
+++ b/src/Calypso-SystemTools-Core/ClyMethodEditorToolMorph.class.st
@@ -91,7 +91,7 @@ ClyMethodEditorToolMorph >> currentEditedAST [
 
 	^ self methodClass compiler
 		  source: self pendingText;
-		  options: #( + optionParseErrors + optionSkipSemanticWarnings );
+		 isFaulty: true;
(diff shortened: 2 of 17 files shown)


jecisc:
My personal favorite is permitFaulty.
My least favorite is supportFaulty because I have the impression it’s a request « do you support faulty » instead of an option.

privat:
I'm used to break the bootstrap at preparation or build. But it's the first time I manage to produce full images but that break this early in its execution.

privat:
Independently of the bootstrap issues, I'm having some second thoughts

MarcusDenker:
At some point I thought about renaming the flag to be about "SyntaxError" (back when it was just about that). 
"faulty" is a bit odd... 

Now you generalized it to cover semantic, too (so nice!), which in turn makes me less sceptical about the term "faulty" (as "error" would not describe it anymore well), so keeping it makes sense.

For the compiler, level, for me "acceptFaulty" or "permitFaulty" would work.

privat:
Rhaaaa #12285

MarcusDenker:
Failing test that is related:
CompiledMethodTest>>testIsFaulty

privat:
At least, the image is ok.
I'm still thinking about it, it might not be the right PR to do now. I do not have a precise roadmap, I'll go with what seems right according to my inspiration and this one feels wrong :)

privat:
What I did not like in the PR is that the parser API is not nice. So I propose #13080 first, then I can come back and implement 'permitFaulty' on the compiler part only

privat:
new version at #13084

PR #13049, opened 2023-03-19
Patch PNGReadWriter to allow and ignore excess chunks after IEND

svenvc:
Make PNGReadWriter>>#processNextChunk return #IEND instead of self and use that in PNGReadWriter>>#nextImage to stop parsing

### DIFF
--- a/src/Graphics-Files/PNGReadWriter.class.st
+++ b/src/Graphics-Files/PNGReadWriter.class.st
@@ -751,7 +751,7 @@ PNGReadWriter >> nextImage [
 	transparentPixelValue := nil.
 	unknownChunks := Set new.
 	stream skip: 8.
-	[ stream atEnd ] whileFalse: [ self processNextChunk ].
+	[ stream atEnd or: [ self processNextChunk = #IEND ] ] whileFalse.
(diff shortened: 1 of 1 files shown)


jecisc (approved):
Maybe a new test can be nice but this PR seems good anyway 👍

svenvc:
> Maybe a new test can be nice but this PR seems good anyway 👍

Normally, I would agree, but this is about parsing a 'bogus' PNG file, a test needs that resource. Either you include it, but that takes more space than we would like, or you refer to a URL, but that is very brittle over time, as various Zinc tests have shown and couples the graphics to the network code.

privat:
Having png assets to tests a png package seams reasonable to me. Another possibility could be to base64 encode images in tests and generate the png on the fly, but this could take up more space in fact.

privat:
there is methods and classes larger than `iVBORw0KGgoAAAANSUhEUgAAAAEAAAABCAYAAAAfFcSJAAAADUlEQVQIW2P4v5ThPwAG7wKklwQ/bwAAAABJRU5ErkJggg==` :)

svenvc:
Danke

PR #13050, opened 2023-03-19
Make unreachable statements a warning instead of an error

privat:
Forbidding to compile method with "grammatically" unreachable code is excessive.

Syntax and semantic errors should not prevent to compile or execute source code with bad style, but only source code that are non-meaningful.
Unreachable code is now signaled by a simple warning (yellow background and :warning: icon).

Moreover, adding an early return in a method when hacking around is a standard practice and programmers did have to use random ugly tricks like opaque predicates (quite transparent, but opaque enough for a parser). e.g. `true ifTrue: [^ blabla ].` or comment the rest of the method to achieve their goals.
Such tricks are far worse than warnings because they are not flagged by compilers, editor UI or critiques.


jecisc (approved):
Seems good to me

MarcusDenker (approved):
Very nice!

PR #13055, opened 2023-03-19
Faulty code: strip OCUndeclaredVariableWarning of its reparation duties

privat:
This is the continuation of #13047

This moves the `openMenu` method and various reparation method of `OCUndeclaredVariableWarning` to a new class `OCCodeReparator` (I did not think that much about the name, but it fits).

The "job" of `OCUndeclaredVariableWarning` is now just to be a notification and the reparation is someone else responsibility. The benefits are:

* We can more easily hack and improve (or kill) the OCSemanticWarning hierarchy of classes.
* The reparator can be improved independently of notification (or even the compiler), be extended to be usable on other kind of reparation, or in other kind of scenarios. For instance, right-click on a source code token with an error or warning, or left-click on an error or warning icon in the icon bar of calypso.

The PR is about moving the responsibility. The original code and its limitations are let as is with as little change as possible. The only improvements of `OCCodeReparator` that were not originally in `OCUndeclaredVariableWarning` are:

* low coupling: only need a node and a requestor (and the requestor should be removable with more work). The reparator does not know the compiler, the compiler context or the notification, for instance.
* lots of tests. I also renamed class `OCCompilerExceptionsTest` from #13047 as `OCCodeReparatorTest` to fit the name convention of test classes.

The only thing I do not like is the new `openMenu` API that returns an "extended" boolean (true, false, nil) because I need to communicate 3 possible outcomes. A better design (for that, but also overall for `OCCodeReparator`) should be done, but I think this is good enough for the current PR.


MarcusDenker:
lots of failing tests (but the image is build and survives running all tests)

privat:
I Got 99 Failing Tests But Bootstrap Ain't One. All are in CDClassSomething, so I might have missed a client of OCUndeclaredVariableWarning

MarcusDenker (approved):

PR #13057, opened 2023-03-20
RBNotice to reify error and warning messages on AST nodes

privat:
On the TODO list of the faulty compiler, there is the idea of offering a better API to manage errors and warnings on nodes.
Strings are nice, but reifying concepts is nicer.

So I propose a simple `RBNotice` class with two subclasses `RBErrorNotice` and `RBWarningNotice`. I needed a word that encompass both errors and warnings (and that was not already taken by other classes or concepts). "Notice" seems fine.

A node has a collection of notices (stored at `#notices` in the property dictionary) and each notice knows its node.
Currently, beside the node, a notice has a `messageText` (not sure of the name, I reused the one of Exception, but I do not like it that much) and a couple of `isError` and `isWarning` predicates.

`RBProgramNode` is extended with some methods to manage and query notices, replacing some existing methods. I put them in the `notice` protocol.

Some clients that picked information directly in `properties` were updated to use the new API.

The only tricky part is for RBParseErrorNode (and subclasses) as they already have all the error information.
I initially thought about making RBParseErrorNode compatible with RBErrorNotice, so it can be its own notice.
But this seems so ugly, so I use a special class `RBErrorNodeNotice` that acts as an adapter. But I'm still not sure.

Overall, this is a simple (but extendable) model, but I'm ready to accept suggestions, critiques, and attempts at bikesheeding :)


MarcusDenker:
I like the design. Having RBErrorNodeNotice as an adapter is actually nice, it does not cost a lot and feels better than making  RBParseErrorNode API compatible to be the Notice.

MarcusDenker (approved):

jecisc (approved):

PR #13078, opened 2023-03-20
Move machineEpsilon to Float.

jecisc:
This is a first step to improve the default comparision precision. The CI crashes so I'm extracting a subset of the PR to see if it passes

### DIFF
--- a/src/Kernel/Float.class.st
+++ b/src/Kernel/Float.class.st
@@ -214,6 +214,18 @@ Float class >> initialize [
 ]
 
 { #category : #constants }
+Float class >> machineEpsilon [
+	"Answer the machine epsilon or macheps, defined by wikipedia asCalypsoItemContext
+	https://en.wikipedia.org/wiki/Machine_epsilon
+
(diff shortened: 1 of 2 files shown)


src/Kernel/Float.class.st
```diff
@@ -213,6 +213,18 @@ Float class >> initialize [
 	NegativeZero := 1.0 / Infinity negated
 ]
 
+{ #category : #constants }
+Float class >> machineEpsilon [
+	"Answer the machine epsilon or macheps, defined by wikipedia asCalypsoItemContext
```
privat: asCalypsoItemContext

privat (approved):

src/Kernel/Float.class.st
MarcusDenker: darn code completion, I wonder if it is not better to turn it off in comments

jecisc: Good catch ahah

MarcusDenker (approved):

PR #13080, opened 2023-03-20
Fauty code: RBParser parse only faulty code

privat:
RBParser is able (and tested) to produce AST with error nodes and precise error and warning information (notices)

The handling of error with error blocks or syntax error notifications is inferior because they are activated too early, without so much context: there is no AST yet, so such errors are only identified with a message string and a single position.
Moreover, they add complexity to the Parser code.

This PR tries to do what  #12912 could not: remove errorBlock from RBParser.
What changed since then is a lot a cleanup and the RBNotice API, so I hope this will work now.

Features

* no more beFaulty, errorBlock or even SyntaxErrorNotification: the parser only goal is to generate AST, not to care about your feeling or water down bad news. You ask an AST, you get an AST, no questions asked, not many flavors to choose.
* new method `RBProgramNode>>checkFaulty:` that simulates the old parser handling of error.
   Therefore, clients that rely on beFaulty, errorBlock and/or SyntaxErrorNotification: can just use this method on the resulting AST and should get the same behavior.
  Note that the receiver is an AST node, not a parser, this is neat.
* Method on the class side of RBParser `parse(Faulty)?(Expression|Method):(onError:)?` are still here, and use `checkFaulty:` to achieve the same results but in a smoother way.

### DIFF
--- a/src/AST-Core-Tests/RBCodeSnippetTest.class.st
+++ b/src/AST-Core-Tests/RBCodeSnippetTest.class.st
@@ -146,7 +146,7 @@ RBCodeSnippetTest >> testMonticello [
 	self assert: MCMockClassE methods size equals: 0.
 
 	snippet isFaulty ifTrue: [
-			self should: [ definition load ] raise: CodeError.
+			self should: [ 	[ definition load ] on: Warning do: [ :e | e resume ]. "Ignore Selector missmatches" ] raise: CodeError.
 			self assert: MCMockClassE methods size equals: 0.
 			^ self ].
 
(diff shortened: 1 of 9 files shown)


privat:
OMG! Bootstrap passed! :tada:

jecisc:
Congrats :P

jecisc:
The ci does not seems to like it

privat:
It's an acquired taste. I just have to insist enough.

jecisc:
I hope you got their consent :)

privat:
CI green except windows that like nothing anyway

jecisc:
This was a timeout on the windows slave :( I hope it’s not a new random failure

MarcusDenker (approved):

PR #13082, opened 2023-03-20
ClassOrganization>>#classify:under:suppressIfDefault: is always called with the same parameter

jecisc:
Little code simplification


jecisc:
Tests passed except what seems to be random windows failures

MarcusDenker (approved):

PR #13083, opened 2023-03-20
Start to inline ProtocolOrganizer behavior in ClassOrganization

jecisc:
Start to inline ProtocolOrganizer behavior in ClassOrganization


jecisc:
Something went wrong, I'll redo it later

PR #13084, opened 2023-03-20
Compiler gets permitFaulty

privat:
This is a rewrite of #13048 and includes code of #13080 that is not merged yet (so bigger commit list and diff)

Parsing and compiling in a "faulty" way is not really a preference toggle, it's mainly the responsibility of client to state what they want and what kind of results they are ready to deal with.

So this PR stop relying on the options #optionParseErrorsNonInteractiveOnly #optionParseErrors #optionSkipSemanticWarnings and propose a single compilation flag `permitFaulty`.
This simplifies the code of clients and the internal logic of the compiler.

Also, I did not like the options: way because the syntax is cumbersome, long names are easy to misspell, and spelling errors are silently ignored.
Note that the 3 options are now unused. I did not remove or deprecate them because doing so break the image at startup. Maybe something related to the system settings, but I'm not sure. (see https://ci.inria.fr/pharo-ci-jenkins2/blue/organizations/jenkins/Test%20pending%20pull%20request%20and%20branch%20Pipeline/detail/PR-13048/1/pipeline if you can help me with that)

The flag `permitFaulty` is stored in the compilerContext but accessible from the Compiler (I'm not a fan or having two classes is not the concern of the PR).

The main difference with #13048 is that there is a new default policy:

* for parsing and semantic analysis, permitFaulty is true by default.
* for compiling and evaluating the flag is false by default.

This simple policy is consistent with the new parser default policy (cf #13080) but also consistent with the usage, since I removed a lot of `options:` in clients that wanted only an AST in silence. A looked at every sender of parse, and I did not find a single client, except some tests, that required a non-faulty parsing and semantic analysis.

### DIFF
--- a/src/AST-Core/CompiledMethod.extension.st
+++ b/src/AST-Core/CompiledMethod.extension.st
@@ -27,7 +27,6 @@ CompiledMethod >> parseTree [
 	ast := self methodClass compiler
 		source: self sourceCode;
 		failBlock: [^ self decompile ];
-		options: #( + #optionParseErrors #optionSkipSemanticWarnings ) ;
 		class: self methodClass;
 		parse.
 	ast compilationContext compiledMethod: self.
--- a/src/AST-Core/RBErrorNodeNotice.class.st
+++ b/src/AST-Core/RBErrorNodeNotice.class.st
@@ -7,6 +7,12 @@ Class {
 	#category : #'AST-Core-Notice'
 }
 
+{ #category : #'error handling' }
+RBErrorNodeNotice >> errorPosition [
+
+	^ node errorPosition
+]
+
 { #category : #accessing }
 RBErrorNodeNotice >> messageText [
 
--- a/src/AST-Core/RBNotice.class.st
+++ b/src/AST-Core/RBNotice.class.st
@@ -11,6 +11,12 @@ Class {
 	#category : #'AST-Core-Notice'
 }
 
+{ #category : #'error handling' }
+RBNotice >> errorPosition [
+
+	^ node start
+]
+
 { #category : #testing }
 RBNotice >> isError [
 
--- a/src/AST-Core/RBProgramNode.class.st
+++ b/src/AST-Core/RBProgramNode.class.st
@@ -308,12 +308,12 @@ RBProgramNode >> checkFaulty: aBlock [
 	"Search for the first error, to signal"
 	errorNotice := self allErrorNotices first.
 
-	aBlock ifNotNil: [ aBlock cull: errorNotice messageText cull: errorNotice node errorPosition cull: self ].
+	aBlock ifNotNil: [ aBlock cull: errorNotice messageText cull: errorNotice errorPosition cull: self ].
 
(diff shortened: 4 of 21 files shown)


privat:
All tests passed except `CompiledMethodTest>>#testIsFaulty`.

The thing is that this test goes by:

```st
testIsFaulty
	|  cm |
	cm := OpalCompiler new
				source: 'method 3+';
				permitFaulty: true;
				compile.
	self assert: cm isFaulty.
	self deny: (OCASTTranslator>>#visitParseErrorNode:) isFaulty
```

but the failed assert is `CompiledMethod>>#isFaulty` that goes by:

```st
isFaulty
 	"check if this method was compiled from syntactically wrong code"
	| ast |
	"fast pre-check: all methods with syntax errors reference this symbol"
  	(self refersToLiteral: #signalSyntaxError:) ifFalse: [ ^false].
	"we have to parse the ast here as #ast does not know that this needs optionParseErrors"
	ast := self methodClass compiler
				source: self sourceCode;
				parse.
	^ ast isFaulty
```

and the issue is in `CompiledMethod>>#sourceCode` that goes by:

```st
sourceCode
	"Retrieve or reconstruct the source code for this method."

	| trailer |
	trailer := self trailer.
	trailer sourceCode ifNotNil: [:code | ^ code ].
	trailer hasSourcePointer ifFalse: [ ^ self codeForNoSource ].

	^ self getSourceFromFile ifEmpty: [ self codeForNoSource ]
```

And `CompiledMethodTrailer>>#sourceCode` that goes by

```st
sourceCode
	"Answer the source code of compiled method as it was decoded.
	Note: it does not attempts to read from source files using sourcePointer,
	nor reconstruct the source code using temp names"

	kind == #EmbeddedSource ifTrue: [ ^data ].
	kind == #EmbeddedSourceWide ifTrue: [ ^data ].
	kind == #EmbeddedSourceQCompress ifTrue: [ ^data ].
	kind == #EmbeddedSourceZip ifTrue: [ ^data ].
	^ nil
```

Except that, for some reason, the kind is *now* `#NoTrailer` and previously it was `#EmbeddedSource`.

Therefore, it now has to do `CompiledMethod>>#codeForNoSource` that goes by:

```st
codeForNoSource
	"if everything fails, decompile the bytecode"
	"If there is no compiler, we cannot decompile it"
	Smalltalk hasCompiler ifFalse: [ ^ nil ].

	 ^(self compiler decompileMethod: self) formattedCode
```

but `(self compiler decompileMethod: self) formattedCode` return the following:

```st
method
	3 + (RuntimeSyntaxError signalSyntaxError:
		 '<an unprintable nonliteral value>')
```

That is completely legal and valid Pharo code, thus is considered a non-faulty method.

So:

* `CompiledMethod>>#isFaulty` seems not that reliable.
* I was responsible because in my precipitation to clean options, I removed an `optionEmbeddSources` in the failing test... oops...

privat:
All green, but review (and merge) #13080 first

MarcusDenker (approved):

PR #13085, opened 2023-03-21
Improve SystemOrganizer namings

jecisc:
Instead of using the vocabulary around "categroy", make it around "packageName" which is way more understandable.

### DIFF
--- a/src/Monticello/MCOrganizationDefinition.class.st
+++ b/src/Monticello/MCOrganizationDefinition.class.st
@@ -147,5 +147,5 @@ MCOrganizationDefinition >> summary [
 { #category : #unloading }
 MCOrganizationDefinition >> unload [
 
-	categories do: [ :pacakgeName | (SystemOrganization listAtCategoryNamed: pacakgeName) ifEmpty: [ SystemOrganization removePackageNamed: pacakgeName ] ]
+	categories do: [ :pacakgeName | (SystemOrganization classNamesInPackageNamed: pacakgeName) ifEmpty: [ SystemOrganization removePackageNamed: pacakgeName ] ]
 ]
--- a/src/Monticello/MCOrganizationDefinition.class.st
+++ b/src/Monticello/MCOrganizationDefinition.class.st
@@ -147,5 +147,5 @@ MCOrganizationDefinition >> summary [
 { #category : #unloading }
 MCOrganizationDefinition >> unload [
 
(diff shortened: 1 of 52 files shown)


MarcusDenker:
```
Instance of SystemOrganizer did not understand #addCategory:

```

jecisc:
I reverted a part of my last commit. I cannot find the last senders of this method easily. I'll do it later with the CI logs and my other cleanings

src/Monticello/MCOrganizationDefinition.class.st
```diff
@@ -146,8 +146,6 @@ MCOrganizationDefinition >> summary [
 
 { #category : #unloading }
 MCOrganizationDefinition >> unload [
-	categories
-		do: [ :category | 
-			(SystemOrganization isEmptyCategoryNamed: category)
-				ifTrue: [ SystemOrganization removeCategory: category ] ]
+
+	categories do: [ :pacakgeName | (SystemOrganization classNamesInPackageNamed: pacakgeName) ifEmpty: [ SystemOrganization removePackageNamed: pacakgeName ] ]
```
privat: pancakes?  noice!

privat (commented):
With 23 commits, 50 files impacted and a big +584-573 diff, this PR is hard to review :(

src/Monticello/MCOrganizationDefinition.class.st
jecisc: I'll fix that :)

jecisc:
I renamed only the methods in one class :(
When I look for stuff using the category naming convention, I have almost 6000 results. If I do PRs too small I'm afraid to still be doing PRs on this stuff in 5 years

But I'll try to do smaller still for the next ones!

privat:
doing reformatting or mass renaming is independent commits can help.
Not one commit per file, but a single commit of mass renaming, then (or before) a single commit of reformatting, then additional commits if needed
the full diff will still be insane, but a commit-by-commit review will be a lot easier to spot weirdness or typo

jecisc:
Marcus noticed that sometimes the categories are not packages but package tags.. I'll close it for now. We need more analysis

MarcusDenker:
Categories are strings that have a package name as a prefix.

So it depends (I think) on which package exists already how to interpret a category like

Hello-Core-Something

It's a bit of a mess and needs some analysis of how to proceed.

PR #13087, opened 2023-03-21
Faulty code: keep AST and source information in faulty CompiledMethod

privat:
https://github.com/pharo-project/pharo/pull/13084#issuecomment-1477133046 showed that `CompiledMethod>>#isFaulty` (and more generally decompilation) can be unreliable if source information is not preserved for faulty methods (and doit methods).

Therefore, this PR forces the compiler to preserve this information.
While the question of source&ast persistence and efficient storage is an important one, this PR is only focussed on the reliability of faulty compiled methods.


privat:
Windows...

MarcusDenker (approved):

PR #13092, opened 2023-03-21
12996-Remove-ReLiteralArrayContainsSuspiciousTrueFalseOrNilRule 

### LINKED ISSUE #12996: Remove ReLiteralArrayContainsSuspiciousTrueFalseOrNilRule
Now Iceberg propose us to check the rules on commit and this is great!
But we have some rules that do not make sens or that have a looooot of false positive. 

Pharo 12 might be a good opportunity to clean that. If we give better critics, then people might have more motivation to fix the critics remaining. 

In that direction I propose to remove ReLiteralArrayContainsSuspiciousTrueFalseOrNilRule because this rule makes no sense anymore because we might want to have a symbol in the literal arrays!

MarcusDenker:
fixes #12996


jecisc (approved):
YES!

privat (approved):

jecisc:

With Jean and me doing a lot of PR this one comes more and more often :(

PR #13093, opened 2023-03-21
Make documentation more readable on github

jecisc:
Add `st` tags to doc to make it more readable on github


privat:
`st` tags are nice, but are newline character intended? (not that I am against them, I like my `\n`)

jecisc:
I think the documentation was generated from Pharo with cr line ending. But nobody uses cr anymore so I guess my code editor replaced the line endings?

I'm not sure it is bad because as I said, nobody except Pharo uses cr anymore

privat:
Yep, 0A is the superior line ending :p

jecisc:
The CI has its usual problems but this does not change the code of Pharo

MarcusDenker:
we should really change the default line ending in Pharo to be the unix one...

jecisc:
I’m 100% for it! But I don’t know where to start ahah

privat:
> I’m 100% for it! But I don’t know where to start ahah

Some `tr` and `sed` could do the trick :p

Ducasse:
Cyril the first step would be use newline instead of cr. 
Like we could change the newline without having to change everything. 
I do not like to be forced about oine ending when I do not need.

PR #13094, opened 2023-03-21
12948-Broken-ast-method-on-compiled-block

### LINKED ISSUE #12948: Broken ast method on compiled block
**Bug description**

Specific combinations of block and global produce problematic CompiledBlock objects where the `ast` method does return a wrong node thus breaking others methods like `ir` and inspection (and possibly more).

It can be shown at least on latest Pharo11

**To Reproduce**

Run in a playground:

```st
| method block |
method := OpalCompiler new compile: 'foo Object. [ self ]. ^ #bar'.
block := method literals detect: #isCompiledBlock.
block ir
```

for some reason, `block ast` gives the return node (RBReturnNode `^#bar`) instead of the block node (RBBlockNode).

@MarcusDenker you might like this one :)

MarcusDenker:
remove the optimization to not emit code for

```
aVariable.
```

We now push and pop the variable, this makes sure that decompiled code will not loose the variable access.

Optimizations should be done on anothe level.

fixes #12948

### DIFF
--- a/src/Kernel-Tests-Extended/BlockClosureTest.extension.st
+++ b/src/Kernel-Tests-Extended/BlockClosureTest.extension.st
@@ -53,8 +53,8 @@ BlockClosureTest >> testTallyInstructions [
 	| expected |
 	expected := Smalltalk compiler compilationContext
 		            optionCleanBlockClosure
-		            ifTrue: [ 26 ]
-		            ifFalse: [ 27 ].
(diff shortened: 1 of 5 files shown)


MarcusDenker:
I propose to fix it in Pharo11 (this PR is for the 11 branch)

privat:
What about a test? for instance, one based on the example of #12948

Issue #10773, opened 2022-01-20
Meta issue for all problems around "Browse it"


MarcusDenker:
**Bug description**
Browse it needs some work 

We found that "Browse it" is supposed to mean "execute this, then browse the class of the result". Discussions have shown that this feature is used a lot and thus essential.

Accidentally, if you "browse it" of a global variable, it browses the class (it executes the code, browse the class of the result).
This has lead to people thinking of "browse it" as "browse the definition if it", because it does so for global references to classes.

Thus quite quickly, people think that when selecting a selector, "Browse it" should do what "implementors of" does and browse all the methods with that selector.

The problem is: the two features can not exist at the same time. "Browse it" can either be "evaluate this and an browse class" *or* it can be a non-evaluating "browse the definition of this". It can *not* be both, as it is not clear when it would evaluate and when not.

There are two more problems (not directly related): 

1. When not selecting anything, it automatically selects and that selection seems to be wrong in some cases.
2. When the selection is done manually and the selected text is not syntactically correct, the compiler complaints. But that is kind of expected, or not? "doIt" and "inspect it" do the same.

- [ ] Browse Class in code pane only works if entire class selected (senders/implementors correctly looks at AST) #9615
- [ ] evaluate: aString andDo: aBlock should better handle buggy expressions. #5013

This issue shows that a "browse Definitons of this" feature is missing.

 "implementors if" works like that to some extent, but it is too low level as it shows *methods* with a name, not class definitions or variable definitions.

already closed:

- [x] Browsing a Symbol produces an error / browse UndefinedObject #8889
- [x]  Browse does not bring the class #10733
- [x] Cmd+Browse raise a DNU #10695	
- [x] "Browse it" evaluates code. Can lead to bad side effects #8515
- [x] analyzeRead:by: sent to nil #9644
- [x] Another DNU when browsing class #9991

MarcusDenker:
The code uses is different between NewTools and the browser (this is what #8889 mentions).

Both have problems, but the first step could be to make sure that they use the same code.

Issue #8515, opened 2021-02-05
"Browse it" evaluates code. Can lead to bad side effects


MarcusDenker:
**Describe the bug**
add this line in the playground:

```
Smalltalk snapshot: true andQuit: true
````

The put the cursor anywhere on the line. CMD-B --> image quits

The reason is that "Browse it" evaluates code. This is not a good idea, as this can have all kinds of evil side effects.

**Comment by kasperosterbye** [comment]
How is this different from doit, printit and inspectit? - I mean - why is it more evil that BrowseIt has side effects than InspectIt has?

**Comment by MarcusDenker** [comment]
I think I expect "browse it" to be purely navigation. Like "implementors of" and "senders of". When I execute code I think about possible side effect, when I navigate I do not.

**Comment by kasperosterbye** [comment]
Do you want `Object newAnonymousSubclass browse` to be without side effects?. I would have no idea of how to come up with a solution, but I am looking forward to seeing what you have of ideas.

**Comment by MarcusDenker** [comment]
no, just the interaction: if I put the courser on the "Smalltalk" and do "alt-b", I do not want the image to quit.

```
Smalltalk snapshot: true andQuit: true
```

I expect it to  realise that Smalltalk is a var pointing to an object and thus browse that class.

if the cursor is on #snapshot, I expect it to browse the implementors of #snapshot: andQuit:

MarcusDenker:
To me "browse it" was meant as an intelligent version that merges both "implementors of" and "Class references to" that in addition knows about variable content.

MarcusDenker:
I am not against a "Doit And browse", but that needs to be explicit that this is a "doit" interaction, not a "code navigation" interaction.

kasperosterbye:
I was just too dense to grasp what you were after. If you put the cursor anywhere in something which is a name of a class, alt-b should just read the variable, not the whole line. 
That is a good idea!
I normally double-click the variable before alt-b.

fxgallego:
Anything in in smalltalk can have side effects.

The term Browse it, means evaluate and browse. It's the same as Do it, Debug it, Inspect it, Print it, it is part of the **same family**. In Dolphin it works the same way and it is something completely normal that you evaluate a line or a message send and browse the result, typically in debug sessions. I believe that it's not only hard to implement this in a safer way but also non desirable. I would consider it a bug not working as Inspect it and the others.

Ducasse:
+100%
For me I expect Browse it to be like Execute and browse. 
[...]

PR #13095, opened 2023-03-21
Rubric/Calypso improve menu entries (especially browse)

privat:
I just discovered why *Browse class* did behave so badly: it is a "do it" action that evaluates then browse the class, and not a navigation that statically resolves class names.

I'm very fine with this new acquired knowledge, but I blame the rubric menu to be so bad:

* "browse" in is the same submenu "code search..." with entries like "senders" and "source code with strings", instead of with other related doit actions (like "inspect it" for instance).
* The label is "Browse full class" that do not hint any kind of evaluation or doit action.

Because hacking compilers requires high quality hacking tools, this PR does the following:

* Move "browse it" with other "doit" action menu (I use the browser icon).
* Precise "Do &" in front of "doit" actions. This helps discoverability of tools.
* Add some menu entry for actions that have keybinding but no menu entries: diff & widen.
* Small fixes because they were easy to do.

### DIFF
--- a/src/Rubric/RubEditingMode.class.st
+++ b/src/Rubric/RubEditingMode.class.st
@@ -23,9 +23,6 @@ RubEditingMode class >> extendedSearchMenuOn: aBuilder [
 		label: 'Code search...' translated;
 		withSeparatorAfter;
 		with: [
-					(aBuilder item: #'Browse full class' translated)
-						keyText: 'b';
-						selector: #browseFullClass.
 					(aBuilder item: #'Senders of it' translated)
 						keyText: 'n';
 						selector: #sendersOfIt.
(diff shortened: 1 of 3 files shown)


MarcusDenker:
I do not like "browse it" to evaluate, to me there should be a real difference between *Navigation* and *Execution*

There are some issues open regarding this:

- "Browse it" evaluates code. Can lead to bad side effects #8515
- Meta issue for all problems around "Browse it" #10773

privat:
I did not know about those issues, they seem political :)
Frankly, I do not care that much if "browse it" evaluates or not, what I do care is that the current menu is misleading. And that's what the PR is trying to fix.

guillep:
Yes, I think that renaming the menu to "Do it and browse" is a good trade-off :)

MarcusDenker (approved):

privat:
Tests are fine, except windows that... segfaulted this time. Oh, it's a new one.

```
Returning an object but pool is full./pharo: line 11:  3363 Segmentation fault      "$DIR"/"pharo-vm/PharoConsole.exe" --headless "$@"
```

MarcusDenker:
yes, it is a good solution

PR #13096, opened 2023-03-21
rubric: doit actions on empty selection selects the current expression

privat:
doit actions (and related ones like inspect or debug) are a nice feature.

However, a thing that I do not like is how it is not aware of the AST and force us to select things.
Especially because if no selection, it defaults on the whole line.
The issue is that the line is often a poor approximation of code meaningful elements (like expressions or statements).

This PR changes the default selection to be the current expression instead of the line.
It changes only rubric (and calypso) and does not touch newtools (debugger or workspace).

Examples. Write the following in a new method panel in Calypso:

```st
foo

1 + 
  100000 timesRepeat: [ (Integer primesUpTo: 100) last factorial ].
```

* Put your cursor on 100, ctrl-I and you get an inspector on 100
* Put your cursor on 100, ctrl-B and you get a browser on the SmallInteger class
* Put your cursor on Integer, ctrl-B and you get a browser on the Integer class
* Put your cursor on factorial, ctrl-shift-D and you can debug the factorial code (and prove yourself that 96192759682482119853328425949563698712343813919172976158104477319333745612481875498805879175589072651261284189679678167647067832320000000000000000000000 make sense)
* Put your cursor just after `[` or `]`, ctrl-I and you can inspect a block
* Put your cursor on `timesRepeat:`, right click, *Profile it*, and you get a profile report
* Put your cursor at the end of the line (after the dot), ctrl-P and `100001` is printed

You can try exactly the same in workspace (or a rubric without this PR) and suffer how bad working with line boundaries by default is.

Note: if you like lines a lot, you can still double-click at their beginning or end to select them, I suppose.

### DIFF
--- a/src/Rubric-Tests/RubSmalltalkEditorTest.class.st
+++ b/src/Rubric-Tests/RubSmalltalkEditorTest.class.st
@@ -99,6 +99,15 @@ RubSmalltalkEditorTest >> assertNodeError [
 ]
 
 { #category : #assertions }
+RubSmalltalkEditorTest >> assertNodeReturn [
+	| node |
+
+	node := self executeBestNodeFor: '^'.
+
+	self assert: node isReturn
+]
+
+{ #category : #assertions }
 RubSmalltalkEditorTest >> assertNodeSelector: aSelector [
 	| node |
 
@@ -464,7 +473,7 @@ RubSmalltalkEditorTest >> testBestNodeWithValidPostionOnMethodPeriodEnd [
 
 	self
 		positionAfter: '9.';
-		assertNodeValue: 9
+		assertNodeReturn
 ]
 
 { #category : #'tests - bestNodeIn' }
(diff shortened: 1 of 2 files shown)


privat:
There is not a lot of reaction. I thought about it a little more.

While defaulting on the line in the absence of an explicit selection is a very poor approximation for code, it is also a very simple and straightforward one.
In the middle of some intense coding session, with a lot of emotional pressure, you do not want to think about what current expression the cursor is on, and do some mental parsing heuristic, with the constant fear of fumbling it.
The default line selection also have the advantage to work on plain text (without AST)

jecisc:
I have seen this PR but I do not know what to think about it. 

On one side, I already got bitten by the fact that I though only the node under my cursor would be evaluated.
On the other side, I tend to evaluate the line while selecting nothing.

I am not sure what is the best here.

MarcusDenker:
Mayne we could test it a bit.

I think when you have the cursor on e.g. a Variable, it kind of makes sense that the action would default to that variable.

```
Smalltalk snapshot: true andQuit: true 
```

when putting the cursor on "Smalltalk" and do a "browse it", I am kind of not expecting what now happens... for "implementors of" works better, as it uses the AST already (it used to take the line and then try to find a selector from that, which was very confusing, too)

jecisc:
Do we integrate to test it? I'm in favor of at least a test

PR #13098, opened 2023-03-21
Faulty code: position fixes on some ast nodes (and various cleanup)

privat:
Added extensive smoke tests on AST nodes, then corrections and cleanup.

Nothing so fancy, but some source code positions were not set or were off-by-one. Thus, causing issues with inconsistent intervals and source code extraction.


privat:
Unrelated windows failure

MarcusDenker (approved):

PR #13099, opened 2023-03-21
Refactoring push down method command to use driver

balsa-sarenac:
Remove UI interaction from command to driver class.

### DIFF
--- a/src/Refactoring-Core/RBPushDownMethodRefactoring.class.st
+++ b/src/Refactoring-Core/RBPushDownMethodRefactoring.class.st
@@ -59,9 +59,8 @@ RBPushDownMethodRefactoring >> classes: aCollection [
 { #category : #actions }
 RBPushDownMethodRefactoring >> generateChanges [
 
-	self preconditions check.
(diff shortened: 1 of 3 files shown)


jecisc:
osx-64 / Tests-osx-64 / testExternalBasicToolsDependencies – MacOSX64.System.DependenciesTests.SystemDependenciesTest4s
Error
Given Collections do not match!  additions : #(#'Commander-Core')  missing: #()
Stacktrace
TestFailure
Given Collections do not match!
	additions : #(#'Commander-Core')
	missing: #()
SystemDependenciesTest>>testExternalBasicToolsDependencies
SystemDependenciesTest(TestCase)>>performTest

We are trying to kill the users of Commander to use Commander2 instead. So I think we do not want this dependency

balsa-sarenac:
@jecisc Thanks for the review! I've removed the usage of `Commander`.

jecisc:
Thank you! Having to remove bad dependencies is really hard sometimes so it's better not to introduce them at all :)

PR #13106, opened 2023-03-22
Add some timestamps and logs to the bootstrap

jecisc:
To help find bottlenecks or speed regression

### DIFF
--- a/bootstrap/scripts/1-clean.sh
+++ b/bootstrap/scripts/1-clean.sh
@@ -5,6 +5,8 @@
 set -x
 set -e
 
+echo $(date -u) "Bootstrap: Begin cleanup of artifacts of previous runs"
+
 SCRIPTS="$(cd "$(dirname "${BASH_SOURCE[0]}")" ; pwd -P)"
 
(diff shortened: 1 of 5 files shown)


privat:
Nice.

It’s a shame that Jenkins does not already timestamp its log.
An alternative could be to pipe any jenkins script inovcation in `ts` https://manpages.debian.org/bullseye/moreutils/ts.1.en.html

<blockquote><div><strong><a href="https://manpages.debian.org/bullseye/moreutils/ts.1.en.html">ts(1) — moreutils — Debian bullseye — Debian Manpages</a></strong></div></blockquote>

jecisc:
Yes, I also wanted to add those for when we play with the bootstrap locally :)

privat (approved):

MarcusDenker:
the slow part seems to be just 3-prepare.sh, so we see there just one timestamp:

slow

jecisc:
I've seen that, I plan to iterate. This is just a first tiny step for better bootstrap :)

PR #13112, opened 2023-03-22
Improving debug session to be able customize debug contexts model

StevenCostiou:
This is required to test the debugger model and the debugger UI without the `DebugContext` class multiple calls to confirmation popups.

### DIFF
--- a/src/Debugger-Model/DebugSession.class.st
+++ b/src/Debugger-Model/DebugSession.class.st
@@ -32,8 +32,29 @@ Class {
 	#classVars : [
 		'LogDebuggerStackToFile'
 	],
+	#classInstVars : [
+		'contextModelClass'
+	],
 	#category : #'Debugger-Model-Core'
 }
+
+{ #category : #accessing }
+DebugSession class >> contextModelClass [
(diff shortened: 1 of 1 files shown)


jecisc (approved):
Good. 

Do you need it for a bugfix on P11? If not, maybe the target should be P12?

StevenCostiou:
Yes I need it on P11 because it is required for a P11 PR on NewTools

jecisc:
Ok! If the CI doesn't find anything suspicious we can merge

StevenCostiou:
Cyril do you know what is the origin of the github rate limit?
Can it be related to our changes?

jecisc:
In the pharo-repository it comes from the fact that we have way more PR than in the past and our CI hits this limit. 
We need to find a way to workaround this problem :( 
It's just hard to find the time to solve this

MarcusDenker:
no,we see that very often recently. This is I think not just Pharo, the whole Inria CI is behind one IP and  counts to the same limit.

MarcusDenker (approved):

PR #13115, opened 2023-03-22
Code Snippets: extends tests to check notices

privat:
Another big diff :(

Each code snippet is provided the information about the expected notices (all errors and warnings).
This is the second commit (the big one). Even if the notices are in a condensed form, it's still big.
Note: if you are brave, you can annotate, in the diff, the errors and warning messages you don't expect (or don't like).

Some tests are extended to check that the notices are the right ones (third commit)

This helps to show that there are still some strange error messages and warnings in some configurations.
It's the point of mass testing, you can check many kinds of syntactic variations.

Therefore, I fixed one: statement that starts with a plain closer `)}]` reported "unexpected token" instead of "missing opener". (last commit)
It also shows that CodeImport do not like dry closing `]`, but is fine otherwise with `]` in found later in the code. It is to de investigated.
But that also shows the benefit of running all snippets on many clients.

I stop here, or the PR will be too huge :)

### DIFF
--- a/src/AST-Core-Tests/RBCodeSnippet.class.st
+++ b/src/AST-Core-Tests/RBCodeSnippet.class.st
@@ -150,11 +150,24 @@ RBCodeSnippet class >> badExpressions [
 				            #( 3 6 7 ''']'' expected' )
 				            #( 1 6 7 '''}'' expected' ) )).
 		        (self new
-			         source: ') ] }';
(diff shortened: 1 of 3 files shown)


jecisc (approved):
I read the code and that seems good

MarcusDenker (approved):

Issue #13114, opened 2023-03-22
Inspecting large Text with a lot of style is slow


privat:
Running in Pharo11 (recent version)

```st
{ 
[ (String loremIpsum: 100000) asText inspect ] timeToRun. "monochrome"
[ (SHRBTextStyler new privateStyle: (String loremIpsum: 100000) asText) inspect ] timeToRun. "mainly red text"
[ RBCodeSnippet styleAllWithError] timeToRun. "many colors, bgcolors and styles"
}
```

gives 1s, 2s and 6s. Whereas the text size to display in the last case is more than half sorter than the others (43063).
Profiler shows that 94% is spent in `SpWindowPresenter>>#openWithLayout:`

PR #13116, opened 2023-03-22
Speed up FreeTypeFont>>widthOf:

jecisc:
This method is heavily called in rubric. This will make ti faster to display text in rubric.

This bring #13114 from 6sec to 4sec. Not perfect but still better


jecisc:
In an image I used a while, I have 95 instances of free type font. This cache should not impact much the memory

privat (approved):
Nice.

As always with cache, it's hard to measure the cost/benefit in the long term.
But 6s to 4s is good short term benefits anyway :)

Just by curiosity, I looked and there is already a LRUCache class in (core?) Pharo, but I don't know what it's worth.

privat:
Maybe it could be good for Pharo11?

MarcusDenker:
at some point we have to stop changing Pharo11... I will merge in Pharo12, at least we have to test it

PR #13119, opened 2023-03-23
Run array cleanings

jecisc:
While checking of the time it takes to render text I was checking the class RunArray. I'm proposing here some cleanings I did while checking the code.

- Remove references to RunArray in RunArray
- use #isNil, #isEmpty, #last... instead of lower level code
- Remove unused methods
- Other small cleanings to make the code more readable

### DIFF
--- a/src/Text-Core/RunArray.class.st
+++ b/src/Text-Core/RunArray.class.st
@@ -422,7 +422,7 @@ RunArray >> do: aBlock [
 	runs doWithIndex: [ :run :index |
 		| v |
 		v := values at: index.
-		[ aBlock value: v ] timesRepeat: run ]
+		run timesRepeat:[ aBlock value: v ] ]
 ]
 
 { #category : #private }
--- a/src/Text-Core/RunArray.class.st
+++ b/src/Text-Core/RunArray.class.st
@@ -153,17 +153,18 @@ RunArray >> , aRunArray [
 	"Answer a new RunArray that is a concatenation of the receiver and
 	aRunArray."
 
-	| new newRuns |
+	| newRuns |
 	aRunArray class = self class ifFalse: [
+		| new |
 		new := self copy.
 		"attempt to be sociable"
-		aRunArray do: [ :each | new addLast: each ].
+		new addAll: aRunArray.
 		^ new ].
-	runs size = 0 ifTrue: [ ^ aRunArray copy ].
-	aRunArray runs size = 0 ifTrue: [ ^ self copy ].
-	(values at: values size) ~= (aRunArray values at: 1) ifTrue: [ ^ self class runs: runs , aRunArray runs values: values , aRunArray values ].
+	runs ifEmpty: [ ^ aRunArray copy ].
+	aRunArray runs ifEmpty: [ ^ self copy ].
+	values last ~= aRunArray values first ifTrue: [ ^ self class runs: runs , aRunArray runs values: values , aRunArray values ].
 	newRuns := runs copyReplaceFrom: runs size to: runs size with: aRunArray runs.
-	newRuns at: runs size put: (runs at: runs size) + (aRunArray runs at: 1).
+	newRuns at: runs size put: runs last + aRunArray runs first.
 	^ self class runs: newRuns values: (values copyReplaceFrom: values size to: values size with: aRunArray values)
 ]
 
@@ -172,8 +173,8 @@ RunArray >> = otherArray [
 	"Test if all my elements are equal to those of otherArray"
 
 	self == otherArray ifTrue: [ ^ true ].
-	self species == otherArray species ifFalse: [ ^ false ].
 	otherArray class = self class ifFalse: [ ^ self hasEqualElements: otherArray ].
+	self species == otherArray species ifFalse: [ ^ false ].
 
 	"Faster test between two RunArrays"
 	^ (runs hasEqualElements: otherArray runs) and: [ values hasEqualElements: otherArray values ]
@@ -184,7 +185,7 @@ RunArray >> add: value [
 	"Add value as the last element of the receiver."
 
 	lastIndex := nil. "flush access cache"
-	(runs size = 0 or: [ values last ~= value ])
+	(runs isEmpty or: [ values last ~= value ])
 		ifTrue: [
 			runs := runs copyWith: 1.
 			values := values copyWith: value ]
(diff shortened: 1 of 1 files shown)


MarcusDenker:
tests failing in MicResourceSettingsTest

src/Text-Core/RunArray.class.st
```diff
@@ -153,90 +153,57 @@ RunArray >> , aRunArray [
...
-	(runs size=0 or: [values last ~= newObject])
-		ifTrue:
-			[runs := runs copyWith: anInteger.
-			values := values copyWith: newObject]
-		ifFalse:
-			[runs at: runs size put: runs last + anInteger]
-]
-
-{ #category : #adding }
-RunArray >> addFirst: value [
```
privat, line 218: I assume that addFirst and addLast is to be compatible with OrderedCollection. I would keep them

src/Text-Core/RunArray.class.st
```diff
@@ -452,12 +419,10 @@ RunArray >> detectMin: aBlock [
 RunArray >> do: aBlock [
 	"This is refined for speed"
 
-	1 to: runs size do: [:i |
-		| r v |
-		v := values at: i.
-		r := runs at: i.
-		[( r := r - 1) >= 0]
-			whileTrue: [aBlock value: v]]
+	runs doWithIndex: [ :run :index |
```
privat, line 422: the initial implementation is faster since `to:do:` and `whileTrue:` are inlined.
Usually I do not care that much, but users of RunArray might need some kind of perf.

src/Text-Core/RunArray.class.st
```diff
@@ -522,7 +488,8 @@ RunArray >> includes: anObject [
 
 { #category : #'self evaluating' }
 RunArray >> isSelfEvaluating [
-	^ self class == RunArray and: [values isSelfEvaluating]
+
+	^ values isSelfEvaluating
```
privat, line 492: I think the test is a guard to prevent instance of subclasses to be wrongfully declared as self evaluating. I think I saw those in other classes.
Better be safe than sorry here, bugs caused here will be hard to track.

src/Text-Core/RunArray.class.st
```diff
@@ -737,18 +680,17 @@ RunArray >> setRuns: newRuns setValues: newValues [
 
 { #category : #accessing }
 RunArray >> size [
-	| size |
-	size := 0.
-	1 to: runs size do: [:i | size := size + (runs at: i)].
-	^size
+
+	^ runs sumNumbers
```
privat, line 684: here also, the initial implementation is faster because of inlining, and `size` might be called a lot.
Maybe the impact is minimal but without perf measures, I would prefer to keep the original code. better be safe.

src/Text-Core/RunArray.class.st
```diff
@@ -153,90 +153,57 @@ RunArray >> , aRunArray [
...
+	| newRuns |
+	aRunArray class = self class ifFalse: [
+		| new |
+		new := self copy.
+		"attempt to be sociable"
+		new addAll: aRunArray.
+		^ new ].
+	runs ifEmpty: [ ^ aRunArray copy ].
+	aRunArray runs ifEmpty: [ ^ self copy ].
+	values last ~= aRunArray values first ifTrue: [ ^ self class runs: runs , aRunArray runs values: values , aRunArray values ].
```
privat, line 165: Formatting is sometime inconstant.

privat (commented):
I'm questioning some possible loss of perf and loss of API.

src/Text-Core/RunArray.class.st
jecisc: RunArray is an internal collection of Text that should not be used somewhere else. 
Adding elements to this should only be managed by the Text class or the RunArray itself. Those I think that we can remove the dead code that just make us lose time to read it

src/Text-Core/RunArray.class.st
jecisc: I think that #timesRepeat: is now inlined too. @MarcusDenker do you know?

src/Text-Core/RunArray.class.st
jecisc: As I said in a previous comment, this is an internal class of Text. I don't think it's good to subclass this kind of classes. You don't want to rely on the internal implementation of other stuff

src/Text-Core/RunArray.class.st
jecisc: I'll add to my todo to check that. I'll add a Work Needed label in the meantime

src/Text-Core/RunArray.class.st
jecisc: I'm using the Pharo formater, I do not format by hand :(

src/Text-Core/RunArray.class.st
privat: RunArray could be used independently of Text. There is no reason to limit usage, nor anything in the comment of the class that refer to Test

jecisc: If that is the case maybe it should not be packaged in Text-Core but in a Collection package?

privat: Packages seems to me weird and unreliable, I think that because there is no namespace and because dependency resolution is mostly faith-based, people do not care that much.

Where do you store a type-agnostic data structure but with a specific use case (long sequences of equal elements) and a single main client in the core classes? There is no good answer here :( so status quo might win.

jecisc: Personally, I do not use classes that seems to be part of something unrelated to me. 

Having a Collection implemented in Text-Core make it a Text collection to me that should not be used outside the Text management. 

To have something clear here, I think that we should either packages it in a general package like Collections-Sequenceable or assume it's specific to text and remove code that is unspecific to tests. 

After having this discussion, I would opt to move it to Collections-Sequenceable.

privat: Seriously, I do not know. Maybe it's out of scope of the current PR.
But it's possible that some people are using RunArray (it even has a general non-prefixed public name, it's an array that runs!) so I would put the method back in to be safe. They are not that big

jecisc: I agrée that this out of the scope of this PR. I’m just wondering how things should be ideally :)

src/Text-Core/RunArray.class.st
jordanmontt: If it is planned to move this class to Collections pacakage, I agree with Jean to keep the guard

jecisc:
I'll close this PR and do other ones (multiple for simplicity of review) with the feedback I got here.

Ducasse:
RunArray is not internal it is an optimized sequence.

PR #13124, opened 2023-03-23
Add a way to know the real processor architecture

demarey:
is useful to determine whether Pharo is running in translated / emulation mode (32 bit on a 64 bit system)


jecisc (approved):
Seems good to me!

demarey:
I will repackage methods to avoid new dependencies

jecisc (approved):

jecisc:
The failure seems random

PR #13127, opened 2023-03-23
CodeSnippet: can update itself thaks to metaprogramming and code transformation

privat:
Maintaining and updating code snippets is tedious.

* If a test fails on a snippet, finding the original definition in the source code that defines snippets is not fun.
* Updating expectations in the source code is also risk-prone.
* Manually filling expectation for new tests (on all existing snippets) or new snippets (for all existing test) is bad use of a programmer time.

A good programmer is lazy and conscientious, thus automatize tedious or error-prone tasks.
It's what this PR does. Provide way to automatize all the use cases presented above than to meta-programming and the package Refactoring2-Transformations.

As a proof of usability, all the updates of #13115 were done while developing the features proposed by the present PR.

### DIFF
--- a/src/AST-Core-Tests/RBCodeSnippet.class.st
+++ b/src/AST-Core-Tests/RBCodeSnippet.class.st
@@ -1737,6 +1737,29 @@ RBCodeSnippet >> raise: anObject [
 	raise := anObject
 ]
 
+{ #category : #updating }
+RBCodeSnippet >> searchDefinition [
+	"Get the AST node defining self in the class side method.
+	The source code is used as a key to distinguishes other snippets.
+	Assume the AST of the class side method respect the format."
+
+	| ast candidate |
+	candidate := OrderedCollection new.
+	ast := (self class class >> default group) ast.
+	ast nodesDo: [ :node |
+		| parent |
+		(node isLiteralNode and: [
+			 parent := node parent.
+			 node value = self source and: [
+				 parent isMessage and: [ parent selector = #source: ] ] ])
+			ifTrue: [
+				parent isCascaded ifTrue: [ parent := parent parent ].
+				candidate add: parent ] ].
+	^ candidate size = 1
+		  ifTrue: [ candidate first ]
+		  ifFalse: [ nil ]
+]
+
 { #category : #accessing }
 RBCodeSnippet >> skip: aSymbol [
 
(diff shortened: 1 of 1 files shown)


privat:
@balsa-sarenac you might like to review this PR since I use refactoring/transformation tools

src/AST-Core-Tests/RBCodeSnippet.class.st
```diff
@@ -1479,14 +1479,33 @@ RBCodeSnippet class >> styleAllWithError [
 	bigtext inspect
 ]
 
+{ #category : #updating }
+RBCodeSnippet class >> updateAllSnippets [
+	"This script will update all definitions of all snippets.
+	Beware of code loss and double check the generated definitions."
+
+	"I did not manage to update all in a single transformation (with a preview!)
+	So do them one by one with a progress bar :("
```
balsa-sarenac, line 1488: I think you can accumulate changes from all refactorings and preview them with `StRefactoringPreviewPresenter`. Only problem is that presenter requires environment and  scopes. In this case it would make sense to have presenter open with default environment and without scopes.
I could try to refactor it and then we can refactor this to preview changes

src/AST-Core-Tests/RBCodeSnippet.class.st
```diff
@@ -1718,6 +1857,38 @@ RBCodeSnippet >> textWithNode: aNode message: aString at: aPosition [
 	^ text
 ]
 
+{ #category : #updating }
+RBCodeSnippet >> updateDefinition [
+	"Do the update of the definition of self in the "
```
balsa-sarenac: This comment seems unfinished.

src/AST-Core-Tests/RBCodeSnippet.class.st
```diff
@@ -1718,6 +1857,38 @@ RBCodeSnippet >> textWithNode: aNode message: aString at: aPosition [
...
+	self definitionRefactoring ifNotNil: [ :transformation |
+		transformation
+			transform;
+			execute ]
+]
+
+{ #category : #updating }
+RBCodeSnippet >> updateExpectations [
+	"Update the value of expectations according what tools produce.
+	Do not blindly trust it, as a tool cannot jusge itself."
```
balsa-sarenac:
```suggestion
	"Update the value of expectations according what tools produce.
	Do not blindly trust it, as a tool cannot judge itself."
```

balsa-sarenac (commented):
Left some inline comments. The code looks good, it's nice and easy to follow.

src/AST-Core-Tests/RBCodeSnippet.class.st
privat: Thanks! I will improve this part in another PR

MarcusDenker (approved):

PR #13131, opened 2023-03-23
Refactorings clean up

balsa-sarenac:
* removing UI logic from refactorings
* add (T) - transformation prefix to realize class
* small cleanups

### DIFF
--- a/src/Refactoring2-Transformations/RBInsertClassTransformation.class.st
+++ b/src/Refactoring2-Transformations/RBInsertClassTransformation.class.st
@@ -74,19 +74,23 @@ RBInsertClassTransformation >> addClass: aName superclass: superclassName subcla
 RBInsertClassTransformation >> preconditions [
 
 	| cond |
-	cond := ((RBCondition isMetaclass: (self model classObjectFor: superclass))
-				errorMacro: 'Superclass must not be a metaclass') not.
-	cond := subclasses inject: cond into: [:sub :each |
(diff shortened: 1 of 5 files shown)


src/Refactoring2-Transformations/RBInsertClassTransformation.class.st
```diff
@@ -72,21 +72,27 @@ RBInsertClassTransformation >> addClass: aName superclass: superclassName subcla
...
+				         isImmediateSubclass: (self model classObjectFor: each)
+				         of: (self model classObjectFor: superclass)) ].
 	^ cond & (RBCondition isValidClassName: className)
-		& (RBCondition isGlobal: className in: self model) not
-			& (RBCondition isSymbol: category)
-			& ((RBCondition withBlock: [category isEmpty not])
-					errorMacro: 'Invalid category name')
+	  & (RBCondition isGlobal: className in: self model) not
+	  & (RBCondition isSymbol: category)
+	  & ((RBCondition withBlock: [ category isNotEmpty ]) errorMacro:
```
MarcusDenker, line 94: Not for this PR (as this was already like that), but & is not that good as it evaluates all the cases and it is not optimized, while #and: is optimized and does not evaluate the block if the sender is false.

MarcusDenker: Ah, this is the & on RBCondition... so my comment is wrong

MarcusDenker (approved):

PR #13132, opened 2023-03-23
Faulty code: improve shadowing policy and implementation

privat:
Currently, shadowing a variable has a very weird semantic.

1 - in interactive mode, all shadowing are forbidden
2 - in non-interactive mode (background method recompilation or file loading), all shadowing are permitted. You can even have a temporary variable named `self` if you want.

I think this need to be improved for several reasons:

* having different language rules if you are in a visual editor or not is bad, and confuse users.
* shadowing is most of the time a bad style, but not error-worthy: the rule of the nearest scope is usually well understood, and its semantic is clear and stable. Warnings and critiques are made for bad style.
* allowing `self` and other reserved name to be shadowed in non-interactive mode is too lenient, and might confuse users more.
* shadowing error can appear without the active involvement of the method. A method with a temporary variable named `foo` has a clear and precise behavior, then someone adds an instance variable `foo` in some distant superclass, causing the method to become illegal (although its behavior is unchanged). And unlike undeclared variables, the user is not even notified of such cases. (undeclared is another issue, because it changes the behavior).

Moreover, independently of the semantic of the language, the current implementation is very complex with a lesser kind of notification ping-pong.
On any shadowing, the semantic analyzer signals an OCShadowVariableWarning notification.
And load it with a lot of contextual information that breaches encapsulation.
If uncaught, the default action of this notification it to check if the initial UI/requestor is interactive or not.
If not interactive, the notification fills the transcript with most of the time useless info.
In interactive mode, the notification will signal a new exception, a semantic error this time, that will fail the compilation by being caught, and display an error message `->`.

This PR solves both issues at once: it proposes a better semantic, and throws away the previous implementation.

The proposed semantic is the following:
* shadowing is accepted but 1. causes a warning (yellow background + ⚠️ icon) 2. except for the 3 reserved variables that causes an error.
* it's the same rules for interactive and non-interactive modes.

Note: I update many tests, and remove some that relied on the OCShadowVariableWarning notification (that is dead). I think that it did not worth the time to convert them since RBCodeSnippet already covers more cases.


jecisc (approved):
I like the unification!

jecisc:
We can also update the default themes to show instance variables and temporary variables different color. Then it becomes even easier to distinguish what is the scope of the shadowed variable! 
I have that in my personal theme

MarcusDenker (approved):

MarcusDenker:
Yes, I think this is good. 

For resereved variables might be interesting to allow shadowing for two reasons:

-  experiments, but there it is no problem to use a compiler with a modified semantic analyzer for an experiment like that.

-  in case we add more reserved vars (e.g there was the idea at some point to introduce thisProcess), if we do that, we might want to allow shadowing in case someone uses the variable. But that should be done when/if we do that.


So ok for me.

PR #13133, opened 2023-03-23
Faulty code: improve error on unexpected block parameter

privat:
Previous error messages on unexpected `:foo` was weird, this PR tries to fix that


jecisc (approved):
Seems good to me

MarcusDenker (approved):

PR #13137, opened 2023-03-24
Speed up symbol creation

jecisc:
Please do not review and do not merge.

My image crash and cannot be open again.
I cannot load my code in a new image because it breaks it.
My local bootstrap produce a non usable image, I have no idea why :(

I just want the CI to give me a working image please! I hope this one will be good TwT


jecisc:
I'm at a point where the bootstrap does not work with my changes. For that I need to update espell I guess but I have a really hard time to do that because of multiple reasons:
- The project has not been build since 2 years and last time it was in a developement version of Pharo 10
- The vm dependency does not exists anymore (maybe there is a redirection but Iceberg does not seems to like it) so I tried to update it to pharo-vm:pharo-9
- Pharo 10 is not able to load the vm dependency, Iceberg always end up crashing (it works in P11)
- The Pharo 10 release had changes in the compiler that cause espell to not load normally
- It's hard to update things without knowing the projects, and even in the image used for the current bootstrap some tests are not passing, so I cannot really rely on those tests

I would really need a hand here to update espell without losing too much time :)

jecisc:
I'll not have the time to finish this for now. Let's close it

PR #13140, opened 2023-03-24
Move runarray to collections support

jecisc:
Move RunArray to Collections-Support that seems to be a nice package for this class.

I took care of cutting the dependencies with Text-Core.

I also did a little cleaning in the tests. Mostly by adding some more initialization to the setup and by removing a duplicated test. I did that in a separated commit


jecisc:
Github API rate...

src/Text-Core/RunArray.extension.st
```diff
@@ -0,0 +1,42 @@
+Extension { #name : #RunArray }
+
+{ #category : #'*Text-Core' }
+RunArray class >> scanFrom: strm [
```
privat, line 4: is this one really used? Since `writeScanOn` do not exist anymore?

jecisc: It has one non test sender that is PositionableStream>>#nextChunkText

src/Collections-Support-Tests/RunArrayTest.class.st
```diff
@@ -22,46 +23,37 @@ RunArrayTest >> collectionWith5Elements [
...
+	array add: $A.
+	array add: $A.
+	array add: $B.
+	self assert: (array at: 3) equals: $B
 ]
 
 { #category : #'tests - accessing' }
 RunArrayTest >> testAt [
 
-	| array |
```
privat: What is the rationale of making `array` an ivar? I understand that it factorizes code, but at the same time it makes tests more complex because the occasional reader have to understand where that `array` comes from.
It also prevents to copy-paste the test code into a workplace (or doit parts of it), or simply use it as a complete functionnal example.

src/Text-Core/RunArray.extension.st
privat: Maybe update the reference to `writeScanOn` in the comment?

src/Collections-Support-Tests/RunArrayTest.class.st
jecisc: It was only because almost all tests started like that but I can see your point. I can revert this commit

src/Text-Core/RunArray.extension.st
jecisc: I can also move it to the only package using it instead of Text-Core

jecisc: In fact the sender is dead. We can kill all the scanning methods :)
But I'll do that in another PR if it's alright

privat: +1

You're the boss, I'm only a commenter :)

src/Collections-Support-Tests/RunArrayTest.class.st
jecisc: I reverted

jecisc:
Github API limit again and again :(

privat (approved):

jecisc:
This PR was reviewed and is green. Maybe we can merge? :) (I need this one to do another cleaning ahah)

PR #13143, opened 2023-03-24
Speed up some fuel tests to not hit default time limit

### LINKED ISSUE #13142: [CI] FLCreateClassSerializationTest takes too much time
The CI often randomly fails on windows because of FLCreateClassSerializationTest having tests taking too much time.

We should check if the tests have a performance problem or not. If it has, fix it. If it does not have any, increase the time limit.

jecisc:
The goal is to have less random failures on CI (and we save a few seconds of test :) On my machine this saves 6seconds. It's not much but it's better than nothing).

#assertCollection:hasSameElements: can be really slow for big collections because it does a #difference:. I propose instead to check the siwe of the collections and check that the two collections as set are equivalent.

Fixes #13142


privat:
`TestAsserter>>#assertCollection:hasSameElements:` should not spend time computing difference and intersection if the sets are the same in the first place.
If you improve perf of the original assertion, everybody wins!

jecisc:
True. 
New PR: https://github.com/pharo-project/pharo/pull/13144

PR #13144, opened 2023-03-24
Speed up #assertCollection:hasSameElements: in case they are equals.

### LINKED ISSUE #13142: [CI] FLCreateClassSerializationTest takes too much time

jecisc:
Better fix for https://github.com/pharo-project/pharo/pull/13143

Fixes #13142

### DIFF
--- a/src/SUnit-Core-Traits/TAssertable.trait.st
+++ b/src/SUnit-Core-Traits/TAssertable.trait.st
@@ -61,7 +61,9 @@ TAssertable >> assertCollection: actual equals: expected [
 
 { #category : #asserting }
 TAssertable >> assertCollection: actual hasSameElements: expected [
-	"Assert that a collection contains the same elements as the given collection. Order is not checked, only the presence/absence of elements."
+	"Assert that a collection contains the same elements as the given collection. Order is not checked, only the presence/absence of elements.
+
+	For performance reasons we check first that they are not equals because difference computation takes long on collections that are not really small."
 
 	[ self assert: actual equals: expected ]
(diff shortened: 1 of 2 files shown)


privat (changes requested):
Why not simple if for early return and less indentation levels?
And why not two level of optimization:

```st
actual = expected ifTrue: [^self].

actualSet := actual asSet. "Or asBag is occurences is important, i don't know)"
expectedSet := expected asSet.
actualSet = expectedSet ifTrue: [ ^self].

"compute missing and additional here,
and you already have the set (or bag)
so it can also be faster
..."
```

jecisc:
I was not feeling really good about calling an assert method that does not do assertions but it's dumb. I'll just do plain return yeah. 
I also like the idea of the set/bag. I'll check the behavior of the current assertion and do the same.

privat:
How much faster? We want numbers!

privat (approved):

jecisc:
Some tests went from 3.5seconds to 40ms :)
Is that fast enough? ;)

jecisc:
The CI did not like it

jecisc:
The tests are passing in my image

jecisc:
Problem comes from Pharo itself

jecisc:
[CI] Fail on WIN random: OmSessionStoreTest(OmStoreTest)>>testEntryForAbsentEntry #12126

PR #13145, opened 2023-03-24
More explicit error message for slot usage in bootstrap

jecisc:
The Pharo bootstrap does not manages slots in the bootstrap currently. 
This gives a better error message explaining that we cannot have a slot and giving the location to the user. 

This requires: https://github.com/pharo-project/pharo/pull/13016
Because it makes it so much easier to update the bootstrap this way ;)


jecisc:
I got bitten by that while trying to improve the collections of Pharo : (
The error message was not helping

MarcusDenker (approved):

MarcusDenker:

MarcusDenker:

PR #13146, opened 2023-03-24
Unify protocol of baselines

jecisc:


jecisc:
I also added a release test. 

With the unification, the system should automatically classify the #baseline: method of BaselineOfXX in the right protocol

jecisc:
Github API limit rate..

MarcusDenker (approved):

jecisc:
Oh boy, this will be long with the number of external changes to do

jecisc:
I'm too lazy to fix all external projects for something this trivial so I'll just close.

PR #13147, opened 2023-03-24
Synch with p11

jecisc:
The CI of Pharo 12 is currently broken because NewTools needs something added only in P11. 

This sync should fix the problem


jecisc:
For some context:

The debugger had some changes. 
This PR was made to P11: https://github.com/pharo-project/pharo/pull/13112

And a new PR of new tools depends on this change. https://github.com/pharo-spec/NewTools/pull/487
Since this was integrated, Pharo 12 builds are broken. But a sync with P11 should fix the problem.

MarcusDenker (approved):

PR #13148, opened 2023-03-24
Faulty code: introduce invalid variables

privat:
AST can have syntax errors. And declarations of block parameters and temporary variables can be broken.
However, we do not want the local variables declared inside to be:

* simply ignored as their future possible usage might be good, and we do not want a bunch of noisy and distracting 'undeclared' (on any error or warning) on them. The main issue to fix is the bad AST, not the variable.
* simply considered good variables as we do not know if they are good variable, and we do not want either a bunch of noisy and distracting 'unused', or 'masked' (on any error or warning) on them either.

The solution is to have a kind of variable that just ignore bad usage on it, and do not interfere with other variable (and play the role of the top element of a lattice).
This is what this PR does.

As shows in snippets. this reduce the number of warnings in case of syntax errors.


jecisc:
There is a merge conflict now

MarcusDenker:
Failing tests:
- RBCodeSnippetTest>>testSimpleFormattedCode
- RBCodeSnippetTest>>testFormattedCode

MarcusDenker (approved):

Issue #12880, opened 2023-02-28
compile: .... methods should return CompiledMethod object


demarey:
Strangely, all compile: .... methods return a selector.
They should return the `CompiledMethod` object.

privat:
Do you have an example?

```st
(Smalltalk compiler compile: 'foo ^1') class >>> CompiledMethod 
```

demarey:
```smalltalk
ClassDescription>>#compile:classified:notifying:
String compile: 'foo' classified: 'foo' notifying: nil >>> "#foo"
```

privat:
My guess it's that a legacy thing and since that there are a lot of clients that expect that such methods return selectors there is no real gain to change that. Maybe @MarcusDenker knows the history better.

MarcusDenker:
yes, this is legacy.. I wonder if it is not better to add new APIs and keep the old ones compatible

privat:
I opened a meta-issue #12883 to rant about the compilation API for instance :)

privat:
My position on this specific issue: `compile:*` is legacy, maybe we could deprecate it? The way forward is the `compiler` message. But the API is still under heavy changes. It could be a Pahro12 goal

```st
String compile: 'foo' classified: 'foo' notifying: nil >>> "#foo"
```

should become something like

```st
String compiler protocol: 'foo'; install: 'foo' >>> aCompiledMethod
```

PR #13152, opened 2023-03-26
OpalCompiler gain install

privat:
The API `compile*` of ClassDescription has some defaults

* cannot specify some parameters like optimization options or faulty mode;
* returns a symbol instead of a compiled method (cf #12880);
* relies on the requestor (and I want to reduce its bad influence).

The proper solution is to move the responsibility to finish the job to the compiler, since it already has almost all information. In the long run, I want to provide a better API (I'm writing something), but I need this little step first, and I also need to run CI on it to see what it breaks.

### DIFF
--- a/src/Kernel/ClassDescription.class.st
+++ b/src/Kernel/ClassDescription.class.st
@@ -393,18 +393,10 @@ ClassDescription >> compile: sourceCode classified: protocolName withStamp: chan
 		source: sourceCode;
 		requestor: requestor;
 		failBlock:  [ ^ nil ];
-		compile.
-
-	logSource ifTrue: [
-		self
-			logMethodSource: (requestor ifNotNil: [ :r | r text ] ifNil: [ sourceCode ]) "the requestor text might have been changed by the compiler and may be different thant text argument"
-			forMethod: method
-			inCategory: protocolName
-			withStamp: changeStamp ].
-
-	self addAndClassifySelector: method selector withMethod: method inProtocol: protocolName.
-
-	self instanceSide noteCompilationOf: method meta: self isClassSide.
+		protocolName: protocolName;
+		changeStamp: changeStamp;
+		logged: logSource;
+		install.
 
 	^ method selector
 ]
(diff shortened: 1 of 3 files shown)


MarcusDenker:

jecisc:
I like to see this responsibility going from ClassDescription to the compiler. I often though that ClassDescription was doing too much in the compilation

jecisc (approved):

MarcusDenker (approved):

MarcusDenker:
- By adding state to the compiler, it goes another direction of PR #11482 (which removes the source). I am not sure about the stateless compiler, but I think even if we decide to go in that direction, the state could be somewhere else (there is in general a huge question about state in the form of the CompilerContext... it feels like that has too much data (requestor!?) for sure. (I will copy this discussion to the other PR).

Another related idea: it is quite hard to edit the AST of a method and then compile and install it.

- the "Refactoring Browser" was, in the true ST80 tradition, build "on top" of an unchanged ST80 meta model. (this seems to be kind of a religion for some people: ST80 was handed down from the Gods, it can not be improved! Mortals have to build on top!)
- As such, it can only install methods by pretty printing the AST and then use the odd API on ClassDescription...

With having the AST available, people are tempted to just get the AST of a method and change it, using a Refactoring or just the API of the AST itself.

They might now naively think that the AST would be "causally connected" and that changing it changes the running system.
(just like changing a class, e.g. adding a method, changes the system directly)

 After that is not the case, they expect for sure some "easy" way to install the changed AST, to then discover there is none...

jecisc:
I personally use this:

{ #category : #'*Chanel' }
RBMethodNode >> install [
	('Updating: ' , self method printString) record.
	^ self methodClass compile: self formattedCode
]

It’s easy but based on the pretty print like you said

privat:
AST are nice, but at some point you need plain source code to put in your code repository. Also, AST are insanely hard to represent to users, so a formatted view is a good compromise.
Anyway, it is something that deserves some deeper reflection... and possibly multiple future PR :)

@jecisc I do not know what `Chanel` is, and putting the install method on an AST node make me uneasy.

MarcusDenker (approved):

PR #13159, opened 2023-03-27
Clean copying methods of ClassDescription

jecisc:
Some of them were dead code. Some of them were really complicated for nothing. In the end, the only use of those methods is to copy the methods of a class in another class.  Before, what was done was to get all the procols of the origin class, get all the selectors in them, and try to copy the selectors doing various check since we didn't had a compiled method instance. 

Here I propose a much simpler version were we just iterate over the methods directly. This is simpler, requires less code and reduces the API of ClassDescription easily.

### DIFF
--- a/src/Deprecated12/ClassDescription.extension.st
+++ b/src/Deprecated12/ClassDescription.extension.st
@@ -1,0 +1,74 @@
+Extension { #name : #ClassDescription }
+
+{ #category : #'*Deprecated12' }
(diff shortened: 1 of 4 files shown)


jecisc:
The little bonus is that it removes 3 methods with "Category" instead of "Protocol" in the name.

jecisc:

privat (approved):

MarcusDenker (approved):

PR #13160, opened 2023-03-27
Simplify and deprecate ClassDescription>>allProtocolsUpTo:

jecisc:
allProtocolsUpTo: is not used in the image so I propose to deprecate it.  In case an external tool is using it, I updated the method to be simpler so it can probably just be inlined.


jecisc:

privat (approved):

MarcusDenker:
there is now a merge conflict

jecisc:
Fixed

MarcusDenker:

PR #13162, opened 2023-03-27
CodeImport - stop requestor madness

### LINKED ISSUE #11400: Compiler Cleanup: CodeImport should not use fake Requestors to hook into compilation errors
**Describe the problem**
The only way to hook into compilation errors with the old compiler API was to implement a fake requestor and implement #notify:at:in:.

CodeImport uses that and implements FileCompilerRequestor and MethodChunkCompilerRequestor. They just re-raise the exception that called the notify in the first place:

```
notify: message at: location in: code
	| newSource |
	self flag: #hack.
	"Should use the new OPalWarnings directly instead of recreating a SyntaxErrorNotification"
	newSource := SyntaxErrorNotification 
		inClass: STCommandLineHandler
		withCode: code
		doitFlag: false
		errorMessage: message
		location: location.
		
	"If the syntax error notification is resumed, then the source was corrected and we have to announce that parsing can restart."					
	ReparseAfterSourceEditing signalWithNewSource: newSource
```

This looks at the first glance trivial. But it actually is not that easy: when the SyntaxErrorNotification opens the Syntax error debugging tool and you edit the code, the resulting *method* does not know the source. The reason is that methods do not store storecode, but get a source pointed only when installed in a class, as the .sources/.changes format needs a class name...

MethodChunk>>#importFor:logSource: now would compile a correct method with the edited code, but then install the source code with the syntax error as code of that method.

A solution thus needs to rewrite MethodChunk>>#importFor:logSource: to generate the method with embedded sources and make sure to store that source, not the original.

privat:
This remove made-up requestors in CodeImport

Tests seems ok (but tests might be lacking).
I also tried to drag-n-drop various `.st` files with various level of brokenness and things seems working.
In case of syntax error, you get the code error exception. You can close the window (and installation is aborted) or you can proceed, that installs faulty methods. Then you can fix the method on your own time.

@MarcusDenker Is there some use cases I'm missing?


privat:
Ouch, I'm not sure what I broke, but I broke it hard.

privat:
Retry CI is case of a transient issue

MarcusDenker (approved):

MarcusDenker:
I think this (loading .st files) is the use-case that it is supposed to support. So with tests green this is good to merge

MarcusDenker:
- ReleaseTest>>testUndeclared: Found undeclared Variables:  requestor in: CodeImporter>>#evaluate

jecisc:
I never liked this request or thing. It’s great to see it been removed.

IMO this was complicating the compilation while it was not it’s responsibility

privat:
Requestors are bad, the plan is to remove them #12883 
There is more in deep details and thought in the blog posts

MarcusDenker (approved):

PR #13165, opened 2023-03-27
Faulty code: make OCASTSemanticAnalyzer (almost) faulty only

privat:
~~This PR need both #13163 and #13164.
For now, it is just here to test the CI and gather possible feedback (and joyful reactions of people).~~
needed PR are merged, and the code is rebased, It's now easy to review (and to rejoice!)

`OCASTSemanticAnalyzer` signals errors in non-faulty mode, and only annotates the AST in faulty mode.
This PR change that: now the analyzer does not care about the faulty mode (almost*), and it just annotates the AST. The signaling of errors is no more its responsibility.

It's the same change that was done to the parser a few PR ago (#13080).
You want to do semantic analysis on a bad AST full of semantic error? Then you get an AST marked up in red, and a silent stare from the semantic analyzer.
If you like exceptions, you can then run `checkFaulty` on the AST (it was introduced for faulty parsing).
It is what the compiler will do in non-faulty-mode, so you get your exceptions.

The benefits are that it simplifies the code of the semantic analyzer, and make its behavior consistent with the one of the parser. It also avoids AST to be only partially analyzed.

If you follow the PR series, you know that error information are now reified as `RBNotice` instances, and `CodeError` is a simple exception used as messenger. Therefore, `checkFaulty` only signals `CodeError` and there is no more need of a complex hierarchy of semantic errors (that only a few tests cared about, by the way).
So the PR also remove `OCSemanticError` and its subclasses.

(*) About the "almost" of the PR title. Only two places remains where the semantic analyzer gets a different behavior in faulty mode:
* For undeclared variables. It is a complex issue to address. cf #13006 . We need to do more cleanup, more improvements and grind more xp.
* For workspace variables, because workspace variables are created at semantic-time in compilation mode to avoid mass creation of random workspace variables each time you highlight the code. I do not like the current design of workspace variables, but's is another story and will be likely solved later — in DLC or something :).

### DIFF
--- a/src/CodeImportCommandLineHandlers/STCommandLineHandler.class.st
+++ b/src/CodeImportCommandLineHandlers/STCommandLineHandler.class.st
@@ -137,7 +137,7 @@ STCommandLineHandler >> handleError: error reference: aReference [
 { #category : #installing }
 STCommandLineHandler >> handleErrorsDuring: aBlock reference: aReference [
 	aBlock
-		on: Error, OCSemanticWarning, OCSemanticError
+		on: Error, OCSemanticWarning
 		do: [ :e | self handleError: e reference: aReference ]
 ]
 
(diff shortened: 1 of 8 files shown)


dionisiydk:
I just read description , no time yet to see the code :( . But it's definitely a great improvement. I was thinking about it for a long time.

privat:
I rebased the PR, it was the simplest way to remove leftovers of the required PR.
Let's run the CI to see if I did not fumble the rebase :)

PR #13168, opened 2023-03-28
ClassComments-Via-Class 

MarcusDenker:
This PR changes all cliens to as the class directly for comments, not the #organizer

This prepares to move the classCommentSourcePointer to ClassDescription

This turns for now off the automatic class template, this should be done in the UI, not in the model

### DIFF
--- a/src/Kernel/ClassDescription.class.st
+++ b/src/Kernel/ClassDescription.class.st
@@ -197,12 +197,12 @@ ClassDescription >> classComment: aString stamp: aStamp [
 	"Store the comment, aString or Text or RemoteString, associated with the class we are organizing.  Empty string gets stored only if had a non-empty one before."
 
 	| pointer header oldComment oldStamp preamble |
-	oldComment := self organization classComment.
-	oldStamp := self organization commentStamp.
+	oldComment := self comment.
+	oldStamp := self commentStamp.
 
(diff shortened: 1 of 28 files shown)


src/Kernel/ClassDescription.class.st
```diff
@@ -334,8 +333,7 @@ ClassDescription >> classesThatImplementAllOf: selectorSet [
 
 { #category : #'accessing - comment' }
 ClassDescription >> comment [
-    "Answer the receiver's comment. (If missing, supply a template) "
-    ^self instanceSide organization classComment ifEmpty: [ self classCommentBlank ]
```
privat, line 338: yes!

privat (approved):

jecisc (approved):
I like to see all those #organization vanish :)

Do you plan to move the empty class template in the UI soon? (just wondering)

MarcusDenker:
For the UI? yes, I will check where the browser calls #comment and do the check there

MarcusDenker:
Even though I am not anymore that convinced about the template...

privat:
I like the removal of the template, in a model POV I think that "no comment" is important information to preserve.
The template should be used when developers want to comment the class: i.e. not on default presentation, but when entering in edition mode.

PR #13169, opened 2023-03-28
ProtocolOrganizer cleanings

jecisc:
This does two things:
- Put code import specific methods in CodeImport package
- Inline some behavior of ProtocolOrganizer in ClassOrganization

This is madness:

```st
aClass organization protocolOrganizer hasProtocolNamed: aString	
```

I'm trying to make Demeter a little happier even if there are still work to do about that.


privat (approved):

MarcusDenker:
And this will make it easier to merge ProtocolOrganizer and ClassOrganizer

jecisc:
It's the end goal indeed :) 
Since changing the kernel is backfiring a lot on me recently, I'm trying to do tiny steps to not bring the wrath of the gods

jecisc:

Again :(

jecisc:
Failures seems unrelated

PR #13172, opened 2023-03-28
[POC] Faulty Code: make undeclared variable a semantic error

privat:
Do not merge, this is just a concept PR to see how far we are currently in the CI.
Code is ugly, do lot look at it.

So, a little explanation.
Undeclared variable (#13006) is a warning in non-interactive mode, but an error in interactive mode.
The first step of what I tried is to make it always an error.

Obviously, all the non-interactive clients of the compiler that deal (for good or bad reasons) with undeclared variables are failing on this new kind of error.
So the second step is to pass each failing clients in faulty mode (`permitFaulty`). Code will be executed and/or installed even if they contain undeclared variable. (note #13152 is needed for that).

In order to be able to bootstrap, 3 clients need to be updated:
* CodeImport for initial .st files, why are there undeclared variable here? I don't know.
* Monticello for import of package, there are undeclared variables because it's the way Monticello works, I'm not sure if we can change that easily.
* TraitsV2 because traits are used I assume, but it's nor clear why there are undeclared variables in traits

The bootstrap is working, but now many tests are failing. The 3rd step is to check them all. 111 is not that much.

I'm still not sure what to do with undeclared variables, but I hope this experiments will help up to decide.

### DIFF
--- a/src/AST-Core-Tests/RBCodeSnippet.class.st
+++ b/src/AST-Core-Tests/RBCodeSnippet.class.st
@@ -328,12 +328,14 @@ RBCodeSnippet class >> badExpressions [
 		        (self new
 			         source: '| | a';
 			         formattedCode: 'a';
-			         isFaulty: false;
+			         isParseFaulty: false;
+			         value: nil;
 			         notices: #( #( 5 5 5 'Undeclared variable' ) )). "This one is legal, it is an empty list of temporaries, the | are dismissed"
 		        (self new
 			         source: '|| a';
 			         formattedCode: 'a';
-			         isFaulty: false;
+			         isParseFaulty: false;
+			         value: nil;
 			         notices: #( #( 4 4 4 'Undeclared variable' ) )). "Same, but are messing with the Scanner"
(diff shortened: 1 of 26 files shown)


privat:
That was a lot easier than expected. The only broken tests that should remain are 3 in NewTools (that princess is in another castle) and all the ones in OCCodeReparatorTest that I introduced to tests reparation of the undefined variable menu.
Because for the experiment, I just do not signal the Undeclared notification, so there is no menu.

My temporary conclusion is that making UndeclaredVariable a real error seems quite possible in a very near future.
It does not solve its bad runtime behavior, but at least having an official error might help.

@MarcusDenker what do you think?

MarcusDenker:
Merge conflict (due to merging ttps://github.com/pharo-project/pharo/pull/13152 I guess?)

MarcusDenker:
This looks promising! This means that as soon as we start Newtools for Pharo12, we could really do that step

privat:
Experiment is conclusive

PR #13173, opened 2023-03-28
Update behaviour of #protocolNamed:

jecisc:
We have an incoherence between ClassOrganization and ProtocolOrganizer about #protocolNamed:

ClassOrganization is returning nil if the protocol is missing.
ProtocolOrganizer is returning a nameless protocol.

I was trying to think about which I prefer, but I don't like either!

Here is what I propose:
- Add ClassOrganization>>#protocolNamed:ifAbsent: to deal with the possibility of a missing protocol
- Make #protocolNamed: throw a NotFound error in case the protocol does not exist.

This PR also fix other Demeter's law violations and reduces the exposition of protocolOrganizer in the system. (Slowly but surely we will kill it!)

### DIFF
--- a/src/EpiceaBrowsers/EpHasImpactVisitor.class.st
+++ b/src/EpiceaBrowsers/EpHasImpactVisitor.class.st
@@ -164,15 +164,10 @@ EpHasImpactVisitor >> visitProtocolRemoval: aProtocolRemoved [
 	- the protocl exists, and
 	- there are methods in the protocol that are not trait methods"
 
-	self
-		behaviorNamed: aProtocolRemoved behaviorAffectedName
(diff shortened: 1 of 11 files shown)


privat (approved):

MarcusDenker:
failing tests are related

jecisc:
I fix two of the tests. 

For #testOnlyWeakSubscriptions it passes in my image.. I have no idea of the error origin.. I hope this was a side effect of the other test failures? Lets see what the CI says

jecisc:
Also the log of the test was really bad so I did this PR: https://github.com/pharo-project/pharo/pull/13178

MarcusDenker (approved):

PR #13175, opened 2023-03-28
[Refactoring] push down method in some classes driver

balsa-sarenac:
* fixed a bug in `RBPushDownMethodDriver` where methods were selected after refactoring was created thus making no difference in the outcome
* introduced driver for `SycPushDownMethodsInSomeClassesCommand`

### DIFF
--- a/src/Refactoring-UI/RBPushDownMethodDriver.class.st
+++ b/src/Refactoring-UI/RBPushDownMethodDriver.class.st
@@ -32,10 +32,10 @@ RBPushDownMethodDriver class >> model: aRBNamespace scopes: refactoringScopes pu
 RBPushDownMethodDriver >> execute [
 
 	| changes refactoring |
-	refactoring := self refactoring.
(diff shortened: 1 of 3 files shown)


guillep:
hi balsa, could you add some tests about the fixed bug?

balsa-sarenac:
Hello @guillep, I can certainly add some tests.
The bug is related to order of invoking UI presenter and other methods. In order to test it, I would need to mock the UI presenter call and refactor the Driver to enable testing. I've found `MockObject` in `SUnit-MockObjects` but it appears to be unused. Can I use it for testing this? Do you have some other suggestion?

guillep:
Maybe you don't need to mock anything. Your `RBPushDownMethodDriver` has a reference to the dialog, and maybe you can cancel it programmatically, no?

For example, in a test you can:
 - initialize the driver
 - make it open the dialog
 - get the dialog and cancel it
 - make it continue and see it has the intended behaviour
?

PR #13176, opened 2023-03-29
remove  commentSourcePointer from ClassOrganization

### LINKED ISSUE #10958: [Memory] commentSourcePointer is not needed for MetaClasses

MarcusDenker:
this removes  commentSourcePointer from ClassOrganization, forwards for now the methods without deprecation

### DIFF
--- a/src/Fuel-Tests-Core/FLTCreateClassOrTraitSerializationTest.trait.st
+++ b/src/Fuel-Tests-Core/FLTCreateClassOrTraitSerializationTest.trait.st
@@ -145,8 +145,10 @@ FLTCreateClassOrTraitSerializationTest >> testCreateWithComment [
 	
 	materializedClassOrTrait := self resultOfSerializeRemoveAndMaterialize: aClassOrTrait.
 
-	self assert: 'test comment' = materializedClassOrTrait comment asString.
-	self assert: 'test stamp' = materializedClassOrTrait commentStamp.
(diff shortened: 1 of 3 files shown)


MarcusDenker:
FLCreateClassSerializationTest>>testCreateWithComment
FLCreateTraitSerializationTest>>testCreateWithComment

MarcusDenker:
So fuel just serializes the sourcePointer because it serializes the organization completely. This makes these tests green, but it is actually dangerours: if you load the class in anther image, this pointer will point just to a random place in the .changes

MarcusDenker:
I suggest to

- change the tests for now to test for comments to be nilled, as this is much better than a wrong sourcePointer
- open an issue for Fuel to change serialization of classes to explicitly store the comment

jecisc (approved):
:)

PR #13178, opened 2023-03-29
Improve announcements printing

jecisc:
With this PR the printing of announcement subscriptions will print the subscribing object and the announcement class. This is useful when an error related to announcements happens in the CI

### DIFF
--- a/src/Announcements-Core/AnnouncementSubscription.class.st
+++ b/src/Announcements-Core/AnnouncementSubscription.class.st
@@ -86,6 +86,19 @@ AnnouncementSubscription >> makeWeak [
 			announcementClass: announcementClass)
 ]
 
+{ #category : #printing }
+AnnouncementSubscription >> printOn: aStream [
+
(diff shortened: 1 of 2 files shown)


privat (commented):
the fact that there is no relation between AnnouncementSubscription and WeakAnnouncementSubscription (both specialize Object) is worrisome

jecisc:
@privat i just started to give them a common superclass but it takes time because I need to bootstrap x)
If I do the change in my image directly, the image dies =X

it would be nice if pushing up an instance var didn’t reset the content of the var..

privat:
Atomic ivar push up is hard :)
At least you can push methods up more easily

MarcusDenker (approved):

jecisc:
> the fact that there is no relation between AnnouncementSubscription and WeakAnnouncementSubscription (both specialize Object) is worrisome

https://github.com/pharo-project/pharo/pull/13185

PR #13179, opened 2023-03-29
Removing unused instance variable `topContext` in DebugContext

### LINKED ISSUE #13111: The DebugContext class seems to have unused variables and accessors

adri09070:
Fixes #13111 

The instance varaible `topContext` of the class `DebugContext`:

- is not used inside the class `DebugContext`
- has no getter (so is not used by any other class)
- has a setter that is used once in `DebugSession>>#createModelForContext:` but, well, it is useless to call a setter if the set variable isn't used anywhere

That's why I removed this instance variable

### DIFF
--- a/src/Debugger-Model/DebugContext.class.st
+++ b/src/Debugger-Model/DebugContext.class.st
@@ -33,8 +33,7 @@ Class {
 	#name : #DebugContext,
 	#superclass : #Object,
 	#instVars : [
-		'context',
-		'topContext'
+		'context'
 	],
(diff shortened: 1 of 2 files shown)


adri09070:
The PR is for Pharo11 but it should be merged for Pharo12 too

MarcusDenker:
we merge all Pharo11 changes once per week to Pharo12 automatically (mostly mondays)

MarcusDenker (approved):

PR #13184, opened 2023-03-29
Compiler source code is String

privat:
Ancient code artifacts and scriptures suggests that at some point in time, the compiler could have required the source code to be a stream.
It is not possible since some time (the parser takes a string before I even start to hack it) but some old clients still uselessly wrap the source code on a stream reader. Moreover, OpalCompiler host its source ivar as a stream, and internally only use its `contents`.

So this PR is mainly a cleanup

* OpalCompiler: use a String for its `source`
* many OpalCompiler clients: give String instead of a Stream.
* few clients: give String instead of a Text.

Commit 9865d55b9a5341d52837daf20d24413c15481565 will be removed in the final version of the PR.
I added it to force CT to catch possible missing clients.


privat:
I'm quite happy, only newtools and spec fails, so I missed apparently nothing on the first try!

MarcusDenker (approved):

PR #13187, opened 2023-03-29
Inline protocol organizer

### LINKED ISSUE #10959: [Memory] Every class and meta class has both a ProtocolOrganizer and a ClassOrganization
**Describe the problem**
For every class, we have 2 ProtocolOrganizer and 2 ClassOrganization instances. This wastes memory 

**Why is that a problem?**
For one, it wastes memory. But in addition, it adds not-needed complexity. 

**Proposal**
There are two steps:
- merge ProtocolOrganizer and ClassOrganization classes. There should be only ProtocolOrganizer
- after this is done, we should review and ask if we really need the -Organizer? In the end, protocols are just a collection of protocols. Why do we even need an organizer for that?

jecisc:
Sorry for the big PR and long commit list but this change is not easy to make since everything in the system relies on this ><

This PR inline ProtocolOrganizer into ClassOrganization.

Most of the behavior should be the same. I know that there is one case where the behavior is different because #classify:inProtocolNamed: was not announcing the creation of new protocols in case the protocol did not exist yet. Which was an error IMO.

I think this part of the system needs more tests but this was too much to do just for 1 PR.
I have the impression that we can do cleaner about the announcements and some other stuff but once again that can be done later.

Please CI, don't be too harsh on me with this one


jecisc:
The logs do not really help... :(

Ducasse:
argh there is a conflict.

MarcusDenker (approved):

PR #13189, opened 2023-03-29
UndeclaredVariable: Refuse to read or write undeclared variable at runtime

privat:
This is a suggestion of @MarcusDenker , so all praise should go to him.

The thing with undeclared variables, is that they are used as a placeholder until a real global variable is declared; either as a named class, or as whatever you put in `Smalltalk globals` (yes, it means that its `at:put:` is magic).
Method code reparation is also magic and does not need a compiler: bytecode stay the same, just a `become:` on the undeclared variable and voilà!

The idea is to use late binding (message send) to implement store and read instead of store/read bytecode.
This is slower, **but** it allows having an error at runtime if the variable is still undeclared, and a correct behavior if the variable did become something else (it is the point of message sending btw).

Note that if the variable is already known at compile time to be a global (or a shared variable), then the efficient bytecode is generated. Moreover, if the method is recompiled, then the efficient bytecode will also be used.

The code of the PR is just a test to see how far the CI can go.

### DIFF
--- a/bootstrap/scripts/01-initialization/01-init.st
+++ b/bootstrap/scripts/01-initialization/01-init.st
@@ -2,5 +2,3 @@ SmalltalkImage classPool at: #SpecialSelectors put: (Smalltalk specialObjectsArr
 
 Class subclasses: (Array with: ProtoObject class).
 ProtoObject class classLayout slotScope parentScope: Class classLayout slotScope.
-
-RBProgramNode classPool at: #FormatterClass put: BISimpleFormatter.
--- a/bootstrap/scripts/02-monticello-bootstrap/03-bootstrapMonticelloRemote.st
+++ b/bootstrap/scripts/02-monticello-bootstrap/03-bootstrapMonticelloRemote.st
@@ -1,3 +1,4 @@
+| mcPackages |
 mcPackages := #(
  'Network-Kernel'
  'Network-MIME'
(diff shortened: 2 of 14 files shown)


privat:
OMG! It worked. Failling tests are expected and all seems related to compilation tests. I will look at them tomorrow or Friday (tomorrow is a little busy).

privat:
This was very easy in fact, there was some real bugs protected by the surprising semantic of undeclared variable. I will PR something to solve them independently.

What remains is
* The discovery or a bug in the decompiler where a temp declaration is missed and a undeclared global is therefore wrongly used instead. That gives superficially the same behavior so is was not detected before. I will open an independent issue. 
* The debugger (in newtools 😡) where apparently the define new class action does not work because I stopped sending messages to undefined variables, and apparently there is a fragile stack of hacks to divert MNU errors into undeclared variable access. I did not look that much into it but from far away it looks like very bad code.
I'm not even sure that this code path is used in real life that much since undeclared are mostly forbidden in interactive mode.

What could also be interesting is to  roughly measure time overhead.

privat:
Anyway, the results of the experiment are very good.
@MarcusDenker I propose to reroll the PR but only to prevent write access, with the intention to merge it this time.
What do you think?

privat:
Superseded by #13201

PR #13199, opened 2023-03-30
Remove RBInstanceVariableNode crufts

privat:
#13189 shows that some leftover from a presumably removed class RBInstanceVariableNode remains, and was not detected because used in an evaluated string — eval is evil :)

Some method `visitInstanceVariableNode:` did still exist with a parameter named `aRBInstanceVariableNode` but I think they are still used via `Slot>>#acceptVisitor:node:`. I'm not sure, so I do not touch


jecisc (approved):
Before we had multiple AST variable nodes. Now we only have one that knows its binding. This was indeed a leftover :)

privat:
CI ok, windows error unrelated

MarcusDenker (approved):

PR #13201, opened 2023-03-30
Faulty code: forbid write to undeclared variables

privat:
This is a cleaned up version of #13189 that should de production ready.
This PR depends on #13195 (3 first commits)

The main change is that only write to undeclared variables at runtime is forbidden.
Read still transparently return nil (or at least the value of the undeclared).

Because static writes to undeclared variables are very rare, and dynamic ones almost nonexistent and always bugs, this should be unnoticeable in production.
But this will bring a little more sanity in the runtime semantic of the language; prevent most bad behaviors (and sneaky and tricky bugs) related to undeclared variables; and reduce user incomprehension during dev and learning.

Another change is names:
* The runtime error class is named `UndeclaredVariableWrite`
* The runtime message is named `runtimeUndeclaredWrite:`. A specific selector is used, so it is easier to track methods that use the late-binding implementation (I will look at them in the bootstrapped image if the gods of CI and rate limits favor us)

Note: MOP-based writes to the undeclared variable are still allowed because `Dictionary>>#add:` is used when registering an undeclared to the global `Undeclared`. But the `add:` method executes `element value: anAssociation value`, and `element` is here an undeclared variable, so signaling an error on `value:` fails at compile-time.
Since touching `Dictionary` seems too much for this PR, I did not change anything here.

### DIFF
--- a/bootstrap/scripts/01-initialization/01-init.st
+++ b/bootstrap/scripts/01-initialization/01-init.st
@@ -2,5 +2,3 @@ SmalltalkImage classPool at: #SpecialSelectors put: (Smalltalk specialObjectsArr
 
 Class subclasses: (Array with: ProtoObject class).
 ProtoObject class classLayout slotScope parentScope: Class classLayout slotScope.
-
-RBProgramNode classPool at: #FormatterClass put: BISimpleFormatter.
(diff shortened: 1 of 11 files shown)


jecisc (approved):
I checked the code and it seems good. I like the change also.

MarcusDenker (approved):
Nice!

PR #13206, opened 2023-03-30
Rename SystemOrganizer "elements" into "Behavior"

jecisc:
In the system we have ClassOrganization and SystemOrganizer that use "element" as a way to describe their content, but why use this generic work when we know what they manipulate?

ClassOrganization is manipulating protocols and SystemOrganizer is manipulating behaviors!

Here I'm doing the renaming in SystemOrganizer. 
- #categoryOfElement: becomes #categoryOfBehavior (and in the futur I also want to kill this "category" vocabulary)
- #removeElement: becomes #removeBehavior

In addition, now both those methods support getting a Behavior and not a Behavior name as parameter. 
This is because the current SystemOrganizer is a glue between the class meta model and packages. It holds only symbols. But we want to remove the need of this glue and manipulate objects directly! So I'm allowing to give objects here and not just dead symbols.

I also got a bunch of deprecation rewrite that I shipped. I have the impression that sometimes those rewrites are not taken into account in Iceberg and I had to force the computation of dirty packages.

And one last thing (yes, really the last now). I removed a method that only had 1 sender and this sender was only calling this one method (#basicRemoveElement:)

### DIFF
--- a/src/Deprecated12/SystemOrganizer.extension.st
+++ b/src/Deprecated12/SystemOrganizer.extension.st
@@ -6,3 +6,10 @@ SystemOrganizer >> categoryOfElement: behaviorName [
 	self deprecated: 'Use #categoryOfBehavior: instead' transformWith: '`@rcv categoryOfElement: `@arg' -> '`@rcv categoryOfBehavior: `@arg'.
 	^ self categoryOfBehavior: behaviorName
 ]
+
+{ #category : #'*Deprecated12' }
+SystemOrganizer >> removeElement: behaviorName [
+
+	self deprecated: 'Use #removeBehavior: instead' transformWith: '`@rcv removeElement: `@arg' -> '`@rcv removeBehavior: `@arg'.
+	^ self removeBehavior: behaviorName
+]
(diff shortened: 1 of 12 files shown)


jecisc:
Also, for now, I kept #classesInCategory:, but it's on my list of methods to rename because it does not return only classes but also traits. Thus it should be #behaviorsInCategory: IMO.

We have a lot of bugs due to things like local/non local methods, classes/traits/behavior, etc. I think we should be really strict on the namings to avoid those at the maximum and avoid using "classes" instead of "behaviors".

See:

```st
(Smalltalk organization classesInCategory: 'VariablesLibrary-Observable') allSatisfy: #isClass "false"
```

BUT

```st
(Smalltalk organization classesInCategory: 'VariablesLibrary-Observable') allSatisfy: #isBehavior "true"
```

Ducasse:
is protocolNameOfElement: introduced by deprecating its previous name?

jecisc:
Yes, it is still not the best name but it is better than the previous one. 
it is on my todo to iterate more on it

guillep:
thanks!

PR #13210, opened 2023-03-31
Faulty code - compiler beef up failblock

privat:
OpalCompiler has a failBlock parameter. However, it is used only 1. with requestor, 2. to exit the compilation after error notification or canceled reparation.

This PR extends the `failBlock` to be used without requestor with a (I hope) simple, straightforward semantic.
And add a lot of tests to check various configurations.

The full error API is the following:
I do not know how to draw flowcharts, so I use some pseudo Pharo code.

```st
permitFaulty
   ifTrue: [ ^ generate possible faulty AST or compiled method.
      "never raise exception, call failblock or notify requestor when permitFaulty" ].
requestor
   ifNotNil: [ ^ self quirksMode where you simulate the classic behavior (with UI things).
      "but with more clean code. failblock is mandatory (as with the classical behavior)" ].
ast isFaulty "is there some syntactic or semantic errors?"
   ifFalse: [ ^ generate non-faulty AST or compiled method "everything went fine" ].
failBlock
   ifNotNil: [ ^ failBlock value: aRBNoticeError.
      "call the failblock with the first error, the value of the failblock (if not local returned
      or something) is used as the result of the full compiler call" ].
CodeError new signal. "Raise an exception"
^ generate possible faulty AST or compiled method. "If exception is resumed, continue in faulty mode"
```

I'm writing a blogpost to explain the full rationale, GitHub PR cover letter is not the best place for details :)

### DIFF
--- a/src/AST-Core-Tests/RBCodeSnippet.class.st
+++ b/src/AST-Core-Tests/RBCodeSnippet.class.st
@@ -1791,11 +1791,18 @@ RBCodeSnippet >> group: aString [
 RBCodeSnippet >> hasAllNotices: someNotices [
 
 	(self notices ifNil: [0] ifNotNil: [self notices size]) = someNotices size ifFalse: [ ^ false ].
-	^ someNotices allSatisfy: [ :each |
-		  self notices anySatisfy: [ :each2 |
-			  each node start = each2 first & (each node stop = each2 second)
-			  & (each position = each2 third)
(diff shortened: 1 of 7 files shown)


privat:
Bootstrap is bad. An PR does too much things. So need work

privat:
Too many concerns, I will split the PR

PR #13214, opened 2023-03-31
Fast table should not manage keydown as keystroke

### LINKED ISSUE #12434: FastTable inserts two characters when opening the filter
**Bug description**

When pressing a key in a fast table, with filter enabled (e.g., in calypso method lists), the first character is inserted twice in the filter pop up.

**To Reproduce**
1. Go to Calypso, navigate to a class with methods, click on the method list
2. Type a character
3. You will see two times the character in the filter

**Expected behavior**
The character should appear only once.

**Screenshots**
imagen

**Version information:**

Pharo 11.0.0
Build information: Pharo-11.0.0+build.474.sha.cd8926c428df0badb660702c102c1a320862f5fb (64 Bit)

**Additional context**
I think this is quite recent. The pharo launcher could help in looking for the version where this got introduced, that could lead to the PR and related changes.

guillep:
Do not manage a keydown as a keystroke. Otherwise a key event may be managed twice.

Fix https://github.com/pharo-project/pharo/issues/12434

### DIFF
--- a/src/Morphic-Widgets-FastTable/FTSearchFunction.class.st
+++ b/src/Morphic-Widgets-FastTable/FTSearchFunction.class.st
@@ -23,11 +23,6 @@ FTSearchFunction >> ghostText [
 ]
 
 { #category : #'event handling' }
-FTSearchFunction >> keyDown: anEvent [
-	^ self keyStroke: anEvent
-]
-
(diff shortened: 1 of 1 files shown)


Ducasse:
Broken test is unrelated

PR #13216, opened 2023-03-31
Refactor: insert new class refactoring

balsa-sarenac:
`RBInserClassTransformation` and `RBInsertClassRefactoring` have the same duplicate logic. This PR addresses this issue by:
* migrating usages of `RBInserClassTransformation` to refactoring
* tagging transformation as unused
* renaming the refactoring so it's more clear that it will produce a new class

### DIFF
--- a/src/Refactoring2-Transformations/RBInsertNewClassTransformation.class.st
+++ b/src/Refactoring2-Transformations/RBInsertNewClassTransformation.class.st
@@ -1,35 +1,16 @@
-"
-Adds a class in the image, optionally inside an hierarchy (with superclass or subclasses).
-
-Usage:
-| transformation |
-transformation := (RBInsertClassTransformation
-	addClass: #InsertedClass
-	superclass: #RBTransformationTest
-	subclasses: (Array with: RBInsertClassTransformationTest)
-	category: #'Refactoring2-Transformations-Tests')
-	transform.
-(ChangesBrowser changes: transformation model changes changes) open
-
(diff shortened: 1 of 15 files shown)


balsa-sarenac:
The build failed, but I think it's unrelated issue.

jecisc:
Indeed, the issue is unrelated

src/Refactoring2-Transformations/RBInsertNewClassTransformation.class.st
```diff
@@ -1,35 +1,16 @@
...
-	category: #'Refactoring2-Transformations-Tests')
-	transform.
-(ChangesBrowser changes: transformation model changes changes) open
-
-Preconditions:
-- superclass must be an existing class
-- each subclass also must be an existing class
-- name of the class to be added must be non-existent in the system
-- name of the category must be a valid one
-"
```
jecisc: There is a bug currently in Pharo 12 that makes some comments vanish. And it happened here.

(A fix for this bug was just integrated)

jecisc (changes requested):

src/Refactoring2-Transformations/RBInsertNewClassTransformation.class.st
balsa-sarenac: Nice catch @jecisc! I've just returned the removed comment. Glad that was fixed, it was a nasty bug.

PR #13217, opened 2023-03-31
Handle Cmd/Ctrl + right arrow during completion.

### LINKED ISSUE #13120: Code completion "open detail"
**Bug description**
It is not possible to open the "detail" window from the code completion menu.
image

As a side-note, the message mentions "cmd" which is not a key on Windows, but this is minor if this is only text, and as long as e.g. Alt or Ctrl can be used instead.

**To Reproduce**
1. Type in some code, press TAB to autocomplete a class name.
2. Try to open the "detail" window as suggested by the text on the completion menu by pressing cmg + ->
3. Observe that nothing happens

**Expected behavior**
A details window gets opened.

**Screenshots**
Here you have a screenshot of the what is I believe the key message here, but this message does not seem to have any senders.
image

**Version information:**
 - OS: Windows
 - Version: 11
 - Pharo Version 10 and 11

**Expected development cost**
For somebody knowing where/when to send this message it's probably not more than 1h of work including some manual testing.

**Additional context**
None

guillep:
Fix #13120

Do not close if Ctrl/Cmd is pressed
Open detail window if cmd/ctrl right is pressed

### DIFF
--- a/src/HeuristicCompletion-Tests/CoCompletionEngineTest.class.st
+++ b/src/HeuristicCompletion-Tests/CoCompletionEngineTest.class.st
@@ -41,24 +41,6 @@ CoCompletionEngineTest >> testAdvanceWordDoesNotCloseCompletionContext [
 ]
 
 { #category : #'tests - interaction' }
-CoCompletionEngineTest >> testAdvanceWordGoesToNextWord [
-
-	| text |
-	text := 'self mEthOdThatDoesNotExist toto'.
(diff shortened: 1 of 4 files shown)


MarcusDenker:
- CoCompletionEngineTest>>testAdvanceWordGoesToNextWord,  but on mac only

guillep:
Windows timed out, but otherwise everything is green!

PR #13218, opened 2023-03-31
Fix selection of multiword

guillep:
A string of N characters have N+1 caret positions. `#nextWord:` should take that into account.

Fix forwardDelete: also for mac machines

### DIFF
--- a/src/NECompletion-Tests/CompletionEngineTest.class.st
+++ b/src/NECompletion-Tests/CompletionEngineTest.class.st
@@ -283,6 +283,25 @@ CompletionEngineTest >> testReplaceTokenAfterMovingCaretToMiddleOfWordWithFollow
 	self assert: editor text equals: 'self toto something that follows'
 ]
 
+{ #category : #tests }
+CompletionEngineTest >> testReplaceTokenWithCaretBeforeEndOfTextWithSpecialCharacterReplacesEntireWord [
+
+	"If the caret is at the end of a word, replace the entire word"
+
+	| text |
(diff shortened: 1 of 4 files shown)


jecisc:
We have failing tests:

osx-64 / Tests-osx-64 / testPreviousWord – MacOSX64.Rubric.Tests.Editing.Core.RubTextEditorTest<1s
osx-64 / Tests-osx-64 / testSelectWordMarkPoint – MacOSX64.Rubric.Tests.Editing.Core.RubTextEditorTest<1s

They are present in the 3 OS

jecisc:
Oh, a fix was pushed right when I commented ahah

src/NECompletion-Tests/CompletionEngineTest.class.st
```diff
@@ -229,18 +229,18 @@ CompletionEngineTest >> testReplaceKeywordTokenWithCaretInTheEndOfWordWithFollow
...
 
 	editor textArea openInWorld.
 	controller openMenu.
 
-	controller context replaceTokenInEditorWith: 'mEthOdThatDoesNotExist:'.
+	controller context replaceTokenInEditorWith: 'toto'.
 
-	self assert: editor text asString equals: 'self mEthOdThatDoesNotExist: something that follows'
+	self assert: editor text asString equals: 'toto'
 ]
```
PalumboN: Take care with this test, we changed the semantics.
I think that we should leave the test as it was and add a new one with the new example.

PalumboN (commented):
Good tests! I left one comment about one of them 😇

PR #13226, opened 2023-03-31
Finder: Ignoring combinations selector for collections

### LINKED ISSUE #13153: Certain Finder searches hang entire application/image
**Bug description**
If you use Finder to search for `'aojfwepoqprjiowjoiajwojfoecmiaowejc' . $w` (any sufficiently long string does it), Finder will cause the entire image to hang. Clicking the red X doesn't work. On Mac I have to force quit the entire application.

**To Reproduce**
1. Open Finder.
2. Select "Examples" to search.
3. Search for something like `'aojfwepoqprjiowjoiajwojfoecmiaowejc' . $w`.
4. Notice that the entire image hangs and clicking the red X does nothing.

**Expected behavior**
This shouldn't hang the entire image. Worst case scenario it should just return an error.

**Screenshots**
Screenshot 2023-03-26 at 6 47 46 PM

**Version information:**
 - OS: mac OS
 - Version: Ventura 13.2.1
 - Pharo Version: 12.0 64 bit

**Expected development cost**
I'm still a Pharo noob, so I have no idea. I'm willing to help do any additional testing that is needed!

jordanmontt:
Pair programming with @ValentinBourcier

Fix issue https://github.com/pharo-project/pharo/issues/13153 ignoring the method combinations. This needs to be discussed. Maybe another better way of doing this is to, as @privat  said in the issue, to put a timeout for each method and to ignore the methods that timed out.

We hear opinions :)

### DIFF
--- a/src/Tool-Finder/Collection.extension.st
+++ b/src/Tool-Finder/Collection.extension.st
@@ -5,3 +5,9 @@ Collection class >> approvedSelectorsForMethodFinder [
 
 	 ^ self selectors
 ]
+
+{ #category : #'*Tool-Finder' }
+Collection class >> forbiddenSelectorsForMethodFinder [
+
+	^ #( #combinations )
+]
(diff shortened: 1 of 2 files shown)


jecisc (approved):
I prefer the timeout but this is a quick way to improve things so let's go :)

jecisc:
Maybe we could also have a tests performing the search on a big string and a big collection to ensure it does not takes more than a amount of time?

MarcusDenker (approved):

PR #13228, opened 2023-03-31
Socket error handling can get in a loop

### LINKED ISSUE #13227: Socket error handling can get in a loop
Most primitives on Socket signal a SocketError when they fail.

This SocketError is constructed using the method socketErrorMessage where the idea is to translate numeric error codes into text.

This method calls socketError which calls primSocketError:

However, when this last method fails, we start all over again.

I propose the following implementation to shortcut this loop:
````
primSocketError: socketID
	"Return an integer encoding the most recent error on this socket. Zero means no error."

	<primitive: 'primitiveSocketError' module: 'SocketPlugin'>
	^ SocketError signal: 'Cannot access socket error code'
````
I am pretty sure I ran into this while hacking but it is hard to reproduce.

svenvc:
Fix a potential loop in primSocketError:


noha (approved):

svenvc:
Thx!

PR #13229, opened 2023-03-31
FFI: Add +optAutoReleaseReturnedValue option

tinchodias:
Add +optAutoReleaseReturnedValue option.

This simplifies a common case where the C function returns an object what must be freed/released/destroyed by the caller.

Example of bindings to [a Cairo function](https://www.cairographics.org/manual/cairo-Paths.html#cairo-copy-path) as it's currently done in [Alexandrie](https://github.com/pharo-graphics/Alexandrie):

```
AeCairoContext>>
currentPath
	"Return a copy of the current path."

	^ self unownedCurrentPath
		  autoRelease;
		  yourself
```
and:
```
AeCairoContext>>
unownedCurrentPath
	"Return a copy of the current path.
	Warning: Sender is responsible of freeing (or sending autoRelease) the returned object."

	^ self ffiCall: #( AeCairoPath cairo_copy_path #( self ) )
```

With this new option, both methods can be rewritten as a single method:
```
AeCairoContext>>
currentPath
	"Return a copy of the current path."

	^ self
		  ffiCall: #( AeCairoPath cairo_copy_path #( self ) )
		  options: #( + optAutoReleaseReturnedValue )
```

### DIFF
--- a/src/UnifiedFFI/ExternalObject.extension.st
+++ b/src/UnifiedFFI/ExternalObject.extension.st
@@ -2,8 +2,9 @@ Extension { #name : #ExternalObject }
 
 { #category : #'*UnifiedFFI' }
 ExternalObject >> autoRelease [
-	self isExternalAddress ifFalse: [ ^ self ].
-	^ self getHandle autoRelease
+	self getHandle isExternalAddress ifFalse: [ ^ self ].
+	self getHandle autoRelease.
+	^ self
 ]
 
 { #category : #'*UnifiedFFI' }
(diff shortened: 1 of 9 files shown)


src/UnifiedFFI/ExternalObject.extension.st
```diff
@@ -2,8 +2,9 @@ Extension { #name : #ExternalObject }
 
 { #category : #'*UnifiedFFI' }
 ExternalObject >> autoRelease [
-	self isExternalAddress ifFalse: [ ^ self ].
-	^ self getHandle autoRelease
+	self getHandle isExternalAddress ifFalse: [ ^ self ].
+	self getHandle autoRelease.
+	^ self
```
tinchodias, line 7: this fix is the only change that can affect users that don't explicitly set the new option

tinchodias:
Programmed with @tesonep !

MarcusDenker:

jecisc:
Maybe this should be rebased for Pharo 12?

tinchodias:
Hi, I don't know current status of Pharo 11, but I guess so.

MarcusDenker:
I changed it to pharo12

PR #13230, opened 2023-03-31
ProtocolOrganizer: Inline in ClassOrganization most behavior not dealing with instance variables directly

jecisc:
This PR inline more of the protocol organizer behavior into ClassOrganization. It mostly inline the behavior that does not deal with instance variables of ProtocolOrganizer.

List of changes:
- ClassOrganization now understand #addProtocol: 
- ClassOrganization>>#addProtocolNamed: now return the protocol
- ClassOrganization know how to return the #allProtocol 
- Add #classify:inProtocolNamed: in ClassOrganization. THIS INCLUDES A CHANGE IN BEHAVIOR! If the protocol does not exist yet, we announce its creation. In the past it was created silently and I think this is a bug.

### DIFF
--- a/src/CodeImport/ClassOrganization.extension.st
+++ b/src/CodeImport/ClassOrganization.extension.st
@@ -24,5 +24,5 @@ ClassOrganization >> internalChangeFromString: protocolSpecs [
 		| name methods |
 		name := spec first asSymbol.
 		methods := spec allButFirst asSet.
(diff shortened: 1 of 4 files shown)


jecisc:
Since there is a change in behavior, maybe this is the part that makes the CI crash? Let's see

jecisc:
Ok, the problem comes from this subset :/
Still no usable log

jecisc:
I'll close it a try a new PR

PR #13239, opened 2023-03-31
Protocols: Allow to compile a method with an instance of Protocol

jecisc:
Here is a new step for going from symbol manipulation to usage of protocols. 

With this pr we get:
- We can now give an instance of Protocol to the compiler to compile a method instead of a symbol (but the symbol works too!)
- I renamed a bunch of methods to make that explicit
- I renamed the impacted variables
- In order to know if a protocol is unclassified, we now have #isUnclassified instead of comparing to `Protocol unclassified` (Btw, this method returns a string and not a protocol, so I'll change that later probably

This will help us when we will update CompileMethod>>protocol to return a protocol. We will have no problem with the compilation and to know if this protocol is unclassified

### DIFF
--- a/src/Kernel/ClassDescription.class.st
+++ b/src/Kernel/ClassDescription.class.st
@@ -362,33 +362,33 @@ ClassDescription >> commentStamp: changeStamp [
 ]
 
 { #category : #compiling }
-ClassDescription >> compile: sourceCode classified: protocolName [
-	"Compile the argument, code, as source code in the context of the receiver and install the result in the receiver's method dictionary under the protocol, protocolName.
+ClassDescription >> compile: sourceCode classified: protocol [
+	"Compile the argument, code, as source code in the context of the receiver and install the result in the receiver's method dictionary under the protocol as parameter.
 	nil is to be notified if an error occurs.
 	The argument code is either a string or an object that converts to a string or a PositionableStream on an object that converts to a string."
 
-	^ self compile: sourceCode classified: protocolName notifying: nil
+	^ self compile: sourceCode classified: protocol notifying: nil
 ]
 
 { #category : #compiling }
-ClassDescription >> compile: sourceCode classified: protcolName notifying: requestor [
+ClassDescription >> compile: sourceCode classified: protcol notifying: requestor [
 
-	^ self compile: sourceCode classified: protcolName withStamp: Author changeStamp notifying: requestor
+	^ self compile: sourceCode classified: protcol withStamp: Author changeStamp notifying: requestor
 ]
 
 { #category : #compiling }
-ClassDescription >> compile: sourceCode classified: protocolName withStamp: changeStamp notifying: requestor [
+ClassDescription >> compile: sourceCode classified: protocol withStamp: changeStamp notifying: requestor [
 
 	^ self
 		  compile: sourceCode
-		  classified: protocolName
+		  classified: protocol
 		  withStamp: changeStamp
 		  notifying: requestor
 		  logSource: self acceptsLoggingOfCompilation
 ]
 
 { #category : #compiling }
-ClassDescription >> compile: sourceCode classified: protocolName withStamp: changeStamp notifying: requestor logSource: logSource [
+ClassDescription >> compile: sourceCode classified: protocol withStamp: changeStamp notifying: requestor logSource: logSource [
 
 	| method |
 
@@ -396,7 +396,7 @@ ClassDescription >> compile: sourceCode classified: protocolName withStamp: chan
 		source: sourceCode;
 		requestor: requestor;
 		failBlock:  [ ^ nil ];
-		protocolName: protocolName;
+		protocolName: protocol;
 		changeStamp: changeStamp;
 		logged: logSource;
 		install.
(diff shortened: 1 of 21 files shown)


src/OpalCompiler-Core/OpalCompiler.class.st
```diff
@@ -1,30 +1,3 @@
-"
```
privat: why did you remove the class comment?

src/OpalCompiler-Core/OpalCompiler.class.st
```diff
@@ -558,7 +531,7 @@ OpalCompiler >> productionEnvironment: anObject [
 
 { #category : #accessing }
 OpalCompiler >> protocolName: aString [
```
privat: update the selectors (and the parameter name).
No need to deprecate them, they are a few days old

src/OpalCompiler-Tests/OCTargetCompiler.class.st
```diff
@@ -1,9 +1,3 @@
-"
```
privat: You do not like class comments?

src/System-Sources/ClassDescription.extension.st
```diff
@@ -1,5 +1,16 @@
 Extension { #name : #ClassDescription }
 
+{ #category : #'*System-Sources' }
+ClassDescription >> logMethodSource: aText forMethod: aCompiledMethod inProtocol: protocol withStamp: changeStamp [
```
privat, line 4: do not add new methods with insane selector. Merge with the ...inProtocolNamed:.. one instead

src/Kernel/ClassDescription.class.st
```diff
@@ -362,41 +362,41 @@ ClassDescription >> commentStamp: changeStamp [
...
+ClassDescription >> compile: sourceCode classified: protocol withStamp: changeStamp notifying: requestor logSource: logSource [
 
 	| method |
 
 	method := self compiler
 		source: sourceCode;
 		requestor: requestor;
 		failBlock:  [ ^ nil ];
-		protocolName: protocolName;
+		protocolName: protocol;
```
privat: you can rename the getter/setters

src/Kernel/Protocol.class.st
```diff
@@ -38,6 +38,13 @@ Protocol >> addMethodSelector: aSymbol [
 	^ methodSelectors add: aSymbol
 ]
 
+{ #category : #converting }
+Protocol >> asSymbol [
+	"This method is here to allow compiler To compile with a Protocol"
```
privat, line 43: the compiler does not call `asSymbol`.
the compiler does no care. it just call the installation methods of the class with whatever was given

src/Kernel/String.extension.st
```diff
@@ -0,0 +1,8 @@
+Extension { #name : #String }
+
+{ #category : #'*Kernel' }
+String >> isUnclassified [
+	"Return true if I am the name of Uncliassified protocol"
```
privat: 'Uncliassified'

privat: I'm not fan of this one.
Moving away from symbols and strings is good, and reifing concepts is great, but adding back random methods to string feels not nice.

src/OpalCompiler-Core/OpalCompiler.class.st
jecisc: I did not. It's probably because I loaded the change of source pointers in my image :(

I'll add it back by hand

src/OpalCompiler-Core/OpalCompiler.class.st
jecisc: Done!

src/Kernel/String.extension.st
jecisc: I don't like it either but I'm not sure we can do without it until we transition.. Or do you have another idea?

src/Kernel/Protocol.class.st
jecisc: Well.. It did not work before I added this method and it works now. It's probably on the methods delegated to the ClassDescription

src/System-Sources/ClassDescription.extension.st
jecisc: I did originally but my image crashed and apparently I had a problem while loading back the code

src/Kernel/Protocol.class.st
privat: find it and kill it! :crossed_swords:

src/Kernel/String.extension.st
privat: A possibility: for clients that manipulate protocols either as string or Protocol (I assume only external clients, because internal one should only deal with Protocol instance?), you can transform the variable early with some `asProtocol` or something (no-op on Protocol but something on String)

jecisc: You prefer to have #asProtocol on String?

privat: yes, because `asXYZ` has a clear traditional meaning of retuning the instance of XYZ, the one of the given name. eg `asPackage` or even `asClass`.

Whereas `isUnclassified` means nothing obvious on String instances. is `'foo'` classified? what is a classified string? is there digits or Chinese characters inside?
I naturally expect that predicates (testing protocol) on strings answer relatively to intrinsic property of the string content. e.g. `isAllDigits`

Moreover, the selector `isUnclassified` gives no indication it is related to protocol in any way.

privat: Note: in fact, I don't know if Protocol objects are global (one Protocol instance by name on the whole system, this feels bad) or local to a class (each class has its own collection of protocols, this feels better). For the latter, `asProtocol` does not work and you should ask the targeted class to do the conversion string->protocol

src/Kernel/ClassDescription.class.st
```diff
@@ -362,41 +362,41 @@ ClassDescription >> commentStamp: changeStamp [
...
 	nil is to be notified if an error occurs.
 	The argument code is either a string or an object that converts to a string or a PositionableStream on an object that converts to a string."
 
-	^ self compile: sourceCode classified: protocolName notifying: nil
+	^ self compile: sourceCode classified: protocol notifying: nil
 ]
 
 { #category : #compiling }
-ClassDescription >> compile: sourceCode classified: protcolName notifying: requestor [
+ClassDescription >> compile: sourceCode classified: protcol notifying: requestor [
```
Ducasse, line 374: missing o

jecisc:
I will close this one and rework that later

PR #13244, opened 2023-04-01
[POC] make OCUndeclaredVariableWarning a CodeError (new attempt)

privat:
New attempt at #13172
Things have improved a lot on the undeclared side, so let's check again


privat:
Less hackish and more promising than #13172, but still some work to do.

However, CodeImport, Monticello and Traits seems to strongly rely on undeclared variables then late reparation to work.
Therefore, we need to think more.

PR #13247, opened 2023-04-01
Protocols: Allow to compile a method with a real protocol

jecisc:
Attempt number 2!

This PR allows to compile a method under a certain protocol... But giving the instance of the protocol instead of a string!

This is a necessary step if we want to be able to use the Protocol class in the image instead of manipulating strings. For example, it will make it easier to make sure CompiledMethod>>protocol return a protocol because this takes care of every places this is done to recompile a method. 

Compared to yesterday I made some adaptations after the review I got.
- I no longer try to improve the #unclassified management (We'll see what to do there later)
- I no longer add any weird extensions on String. I now added #isProtocol 
- I no longer add #asSymbol magic to Protocol. This was probably not really explicit indeed


jecisc:
I have the impression I still have a bug here :( Some methods are badly classified in my image after loading this change
I'll check that this evening or tomorrow.

MarcusDenker:
- ProperMethodCategorizationTest>>testNoUncategorizedMethods

jecisc:
Tests passed again after I added a new test

src/CodeExport/ClassDescription.extension.st
```diff
@@ -120,7 +120,9 @@ ClassDescription >> printProtocolChunk: protocolName on: aFileStream withStamp:
 			 strm
 				 nextPutAll: self name;
 				 nextPutAll: ' methodsFor: ';
-				 print: protocolName asString.
+				 print: (protocol isProtocol
+						  ifTrue: [ protocol name ]
+						  ifFalse: [ protocol asString ]).
```
privat: I would have used the other way around

```st
print: (protocol isString
   ifTrue: [ protocol ]
   ifFalse: [ protocol name ]).
```

Why, because random objects used as protocol (because of a bug or whatever) will "behave" instead of failing, and you might have protocols with very strange names in your system.

And if for some reason all `isProtocol` can be inverted to `isString`, then you can even remove `isProtocol` if you with.

src/Kernel/ClassDescription.class.st
```diff
@@ -28,16 +28,19 @@ ClassDescription >> acceptsLoggingOfCompilation [
...
 	priorMethodOrNil := self compiledMethodAt: selector ifAbsent: [ nil ].
 	priorMethodOrNil ifNotNil: [ priorOriginOrNil := priorMethodOrNil origin ].
 
 	self addSelectorSilently: selector withMethod: compiledMethod.
 
 	oldProtocol := self organization protocolNameOfElement: selector.
 	SystemAnnouncer uniqueInstance suspendAllWhile: [
-		self organization classify: selector under: (protocol = Protocol unclassified
+		"The next part is really ugly with the double if.. BUT! I will fix that in a new PR soon... I want to remove the usage of `Protocol unclassified` as a way to say 'Do not upatde the protocol' because this is counter intuitive."
+		self organization classify: selector under: ((protocol isProtocol
```
privat: same thing here. `isString` instead of `isProlocol` (does matter less if you kill this part)

src/Kernel/ClassDescription.class.st
```diff
@@ -489,6 +492,16 @@ ClassDescription >> definitionStringFor: aConfiguredPrinter [
 	^ self subclassResponsibility
 ]
 
+{ #category : #'queries - protocols' }
+ClassDescription >> ensureProtocol: aProtocol [
```
privat, line 495: nice!

src/Kernel/ClassOrganization.class.st
```diff
@@ -98,16 +98,23 @@ ClassOrganization >> classify: aSymbol inProtocolNamed: aProtocolName [
...
+ClassOrganization >> classify: selector under: protocol [
 
-	| oldProtocolName forceNotify oldProtocols |
+	| oldProtocolName forceNotify oldProtocols protocolName |
+	"The next section deserve more cleanings.
+	
+	Some code was added to make it possible to classify giving a real protocol.
+	Here to keep a small change, I just ask the name to the protocol and use that for compatibility.
+	Later, I plan to update this code once more to directly use the actual object I'm givin instead of doing this little trick."
+	protocolName := protocol isProtocol ifTrue: [ protocol name ] ifFalse: [ protocol ].
```
privat: same here, `isString`

src/Monticello/MethodAddition.class.st
```diff
@@ -41,7 +41,9 @@ MethodAddition >> compile: aString classified: aString1 withStamp: aString2 logS
 
 { #category : #operations }
 MethodAddition >> createCompiledMethod [
-
+	"CyrilFerlicot: Why do we need to explicitly write to the source? Why do we call the compiler?
+	
+	Can't we just call #compile:classfied: here?"
```
privat, line 46: My understanding is that when installing a method in a class, its source is stored in the sources file and/or? the change file (epicea thing?), therefore the full original source is needed. But it's just what I deduced from various clues in the code and I might be wrong (or right, but still very far from the truth)

src/Monticello/MethodAddition.class.st
```diff
@@ -87,9 +89,11 @@ MethodAddition >> priorCategoryOrNil: anObject [
 
 { #category : #operations }
 MethodAddition >> writeSourceToLog [
-	logSource ifTrue: [
-		myClass logMethodSource: text forMethod: compiledMethod
-			inProtocolNamed: category withStamp: changeStamp
```
privat, line 92: the original formatting is very funny

src/OpalCompiler-Core/OpalCompiler.class.st
```diff
@@ -437,21 +437,19 @@ OpalCompiler >> install [
 	method := self compile.
 	method ifNil: [ ^ method ].
 
-	protocolName ifNil: [ protocolName := Protocol unclassified ].
+	protocol ifNil: [ protocol := Protocol unclassified ].
+	protocol := class ensureProtocol: protocol.
```
privat, line 440: maybe `ensureProtocol: nil` could return an unclassified protocol? Or maybe not, it's a little weird. I don't know

src/CodeExport/ClassDescription.extension.st
jecisc: I applied your advice and removed #isProtocol! :)

src/Monticello/MethodAddition.class.st
jecisc: I'm always weird out by that formatting

src/OpalCompiler-Core/OpalCompiler.class.st
jecisc: I just did that, thanks for the idea

src/Fuel-Platform-Pharo-Core/FLPharoPlatform.class.st
```diff
@@ -40,9 +40,9 @@ FLPharoPlatform >> absolutePathOf: aString [
 ]
 
 { #category : #'accessing-compiler' }
-FLPharoPlatform >> addAndClassifySelector: aSymbol inClass: aClass withMethod: aCompiledMethod inProtocol: aString [
+FLPharoPlatform >> addAndClassifySelector: aSymbol inClass: aClass withMethod: aCompiledMethod inProtocol: aProtocolName [
```
privat: `aProtocol` instead of `aProtocolName`? (same for bellow)

privat (approved):

src/Monticello/MethodAddition.class.st
MarcusDenker: yes, it needs the source. MethodAddition is odd:, it is used by MCPackageLoader>>#basicLoadDefinitions to seperate method compilation from method installation and then notification of change to the observers.

This shows that Monticello just hacked stuff on top instead of improving the meta model of the language...

MarcusDenker (approved):

Issue #13015, opened 2023-03-15
TextConstants should not contain fonts only constant


Ducasse:
Right now 
TextConstants has a at: #TextSharedInformation
and this contains fonts that are sharing.

Ducasse:
Regarding Issue #13015, I have a few questions that I would like to discuss with you. Firstly, I am trying to understand what you mean by "fonts" in this context. As far as I understand, constants here is CtrlE, CtrlF etc. However, I am not sure about the inclusion of fonts in TextConstants. Could you please clarify this for me?

Assuming that fonts do indeed refer to the styles of text such as bold and italic, the question arises as to where they should be placed. Should we create a separate class for fonts, or should we place them in a different location within the code? Additionally, it would be helpful to know if we need to remove TextSharedInformation from classVariableNames and other mentions in the codebase.

Ducasse:
My main point is that in a constant we should have constants :)
Now we should check because there is probably a  font manager or something like that where we could move the font part.
Having a separate SharedPool to hold the fonts looks more natural to me. 
I did not have the time right now to look deeper into it but looking around seems good to me.

shreeshjha:
I would like to state that I have not completely removed the TextSharedInformation element, but rather preserved it in the TextConstant Class with commented-out code. Additionally, the same can be said for the accessibility of this element. To improve the situation, I have added a comment suggesting the use of a separate SharedPool to store the fonts, which would be more appropriate. At present, I am reviewing the codebase to identify a more efficient approach to resolving this issue. I would like to add that I am currently addressing an issue related to the build, which is currently failing. I am actively working to resolve this matter and will provide updates as soon as possible.

Ducasse:
Cool!

PR #13249, opened 2023-04-02
Refactor TextConstants to separate font and non-font constants (Issue : #13015)

shreeshjha:
This pull request addresses issue #13015 by refactoring the TextConstants class to separate font and constants. Currently, TextConstants contains both constant and fonts which can cause confusion in the code. By separating them, the code becomes more organized and easier to maintain.

### DIFF
--- a/src/Text-Core/TextConstants.class.st
+++ b/src/Text-Core/TextConstants.class.st
@@ -87,11 +87,10 @@ Class {
 		'RightMarginTab',
 		'Space',
 		'Tab',
-		'TextSharedInformation'
 	],
 	#category : #'Text-Core-Base'
 }
-
+"TextSharedInformation"
 { #category : #accessing }
 TextConstants class >> BS [
 	^BS
@@ -502,10 +501,10 @@ TextConstants class >> Tab [
 	^Tab
 ]
 
-{ #category : #accessing }
+"{ #category : #accessing }
 TextConstants class >> TextSharedInformation [
 	^TextSharedInformation
-]
+]"
 
 { #category : #'backward compatibility' }
 TextConstants class >> at: aSymbol [
@@ -516,7 +515,7 @@ TextConstants class >> at: aSymbol [
 { #category : #'class initialization' }
 TextConstants class >> initialize [
 
-	TextSharedInformation ifNil: [ TextSharedInformation := Dictionary new].
+	"TextSharedInformation ifNil: [ TextSharedInformation := Dictionary new]."
 	Ctrll := Character value: 153.
 	Ctrld := Character value: 132.
 	ESC := Character value: 160.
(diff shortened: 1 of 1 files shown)


MarcusDenker:
Full image load fails with:

```
MetacelloNotification: Loading -> Text-Core-tonel.1 --- tonel:///builds/workspace/est_and_branch_Pipeline_PR-13249/64/src --- cache
STONReaderError: At character 1105: 'invalid input'
```

Ducasse:
May be this is due to a bootstrap problem (did not check).

jecisc (changes requested):
A little piece of advice: Do not comment code. Remove it. 

If we have a problem here in the future, we can use a git blame to find the old implementation. 
Keeping commented code will just make us lose time reading it and wonder "why is this here? Should I keep it working if we remove the comments?"

And most of the time, you lose way more time asking yourself those questions than just using a git blame if one day there is a problem here.

shreeshjha:
> A little piece of advice: Do not comment code. Remove it.
> 
> If we have a problem here in the future, we can use a git blame to find the old implementation. Keeping commented code will just make us lose time reading it and wonder "why is this here? Should I keep it working if we remove the comments?"
> 
> And most of the time, you lose way more time asking yourself those questions than just using a git blame if one day there is a problem here.

Thanks @jecisc I completely agree with you that keeping commented code can cause confusion and waste time. I will make sure to remove it in the next update once I figure out how to eliminate the issue that it consisted, and I appreciate your advice to use git blame (I wasn't aware of this feature, and it's good to know that it can be useful in finding the old implementation) to find the old implementation if necessary.

MarcusDenker:
so the problem is caused directly by the commenting out: the syntax of Tonnel (the one stored on git) does not allow commenting out like that.

```

 "{ #category : #accessing }
 TextConstants class >> TextSharedInformation [
 	^TextSharedInformation
 ]"

 { #category : #'backward compatibility' }
 "TextConstants class >> at: aSymbol [
 	^ self classPool at: aSymbol ifAbsent: [self TextSharedInformation at: aSymbol]
 ]"
```

shreeshjha:
> so the problem is caused directly by the commenting out: the syntax of Tonnel (the one stored on git) does not allow commenting out like that.
> 
> ```
> 
>  "{ #category : #accessing }
>  TextConstants class >> TextSharedInformation [
>  	^TextSharedInformation
>  ]"
> 
>  { #category : #'backward compatibility' }
>  "TextConstants class >> at: aSymbol [
>  	^ self classPool at: aSymbol ifAbsent: [self TextSharedInformation at: aSymbol]
>  ]"
> ```

Thanks @MarcusDenker bringing the problem to my attention. I understand that the problem was caused by the commenting out, and that the syntax of Tonnel does not allow commenting out like that (will work on this error on next commit and Thanks again for the observation).

guillep:
Hi @shreeshjha ! Thanks for the PR. Please, do not just edit the code in github, that can lead to a lot of problems :)

I'll close this PR until this is fixed, then we can reopen it or open a new PR.

shreeshjha:
Hello @guillep Thanks for taking the time to review my PR. I'll look into this issue soon and I'll make sure to follow the correct process in the future :)

PR #13251, opened 2023-04-03
12885-Objectassert-fails-with-non-booleans-contradicting-its-own-comment 

### LINKED ISSUE #12885: Object>>assert: fails with non booleans, contradicting its own comment
**Describe the problem**
```smalltalk
Object » assert: aBlock
	"Throw an assertion error if aBlock does not evaluates to true.
	We check for true explicitly to make the assertion fail for non booleans"

	<debuggerCompleteToSender>
	self assert: aBlock description: 'Assertion failed'
```

According to the method comment, it should be OK to do `self assert: #meh` — the assertion should fail just like with `self assert: false`. However the comment is lying because `assert:description` assumes the block returns a boolean and signals `NonBooleanReceiver`.

**Classes involved**
- `Object » assert:`
- `Object » assert:description:`

**Proposal**
Either update the comment, or change `Object » assert:description:` to use `aBlock value == true ifFalse: […`

**Version information:** Pharo-11.0.0+build.549.sha.06617dfe20afc4af3abb8960015414652714ad2e

**Expected development cost**
It will take longer to commit and merge the change than to do it in the image.

MarcusDenker:
make #assert: allow to be more stable for non-booleans
fixes #12885

### DIFF
--- a/src/Kernel-Tests/ObjectTest.class.st
+++ b/src/Kernel-Tests/ObjectTest.class.st
@@ -27,6 +27,15 @@ ObjectTest >> testAs [
 	self assert: coll1 identicalTo: coll2
 ]
 
+{ #category : #tests }
+ObjectTest >> testAssert [
+
+	self shouldnt: [Object new assert: true] raise: AssertionFailure.
+	self shouldnt: [
+		"make sure that we non-boolean arguments just fails"
+		self should: [Object new assert: 1] raise: AssertionFailure ] raise: [NonBooleanReceiver]. 
+]
+
 { #category : #'tests - accessing' }
 ObjectTest >> testBasicSize [
 	"Numbers are not indexable"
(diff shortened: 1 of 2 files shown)


jecisc (commented):
Maybe we should add a test about that because it's the kind of things that can easily be removed in the future because `== true ifFalse:` is often simplified into `ifFalse:`?

privat:
Nitpick: while the boundary between error and fail in unit tests is sometime slim (and in practice do not matter that much as both are failures), a non-boolean given to an assert seems inconsistent with the rest of the language. I would have updated the comment instead.

demarey (approved):

jecisc (approved):
I don't know if it should be an error or an assertion failure, but the non boolean error is often not really nice in the CI to find the origin of the problem and this is already a step forward so I think we can integrate and maybe improve further later by making it an error if we find that more consistent

PR #13256, opened 2023-04-03
Remove ClassOrganization>>allProtocolNames

jecisc:
#allProtocolNames return the name of the protocols + #'--all--', but this #'--all--' is never used. 

This removes #allProtocolNames and adds #protocolNames instead that does not deal with the instance of AllProtocol. 

The end goal is to remove AllProtocol from that system because this is not really used and makes the MM too complicated for what it does.

### DIFF
--- a/src/Deprecated12/ClassOrganization.extension.st
+++ b/src/Deprecated12/ClassOrganization.extension.st
@@ -10,8 +10,8 @@ ClassOrganization >> addCategory: aString [
 { #category : #'*Deprecated12' }
 ClassOrganization >> allCategories [
 
-	self deprecated: 'Use #protocolsNames instead.' transformWith: '`@rcv allCategories' -> '`@rcv protocolsNames'.
-	^ self protocolsNames
+	self deprecated: 'Use #protocolNames instead.' transformWith: '`@rcv allCategories' -> '`@rcv protocolNames'.
+	^ self protocolNames
 ]
 
 { #category : #'*Deprecated12' }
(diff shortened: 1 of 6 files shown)


src/Deprecated12/ClassOrganization.extension.st
```diff
@@ -10,8 +10,8 @@ ClassOrganization >> addCategory: aString [
 { #category : #'*Deprecated12' }
 ClassOrganization >> allCategories [
 
-	self deprecated: 'Use #allProtocolsNames instead.' transformWith: '`@rcv allCategories' -> '`@rcv allProtocolsNames'.
-	^ self allProtocolsNames
+	self deprecated: 'Use #protocolsNames instead.' transformWith: '`@rcv allCategories' -> '`@rcv protocolsNames'.
```
MarcusDenker: protocolsNames should be protocolNames, I think

MarcusDenker (approved):

src/Deprecated12/ClassOrganization.extension.st
jecisc: Indeed! I'm fixing that :)

privat (approved):
I dislike how sometimes you have a random `all` at the beginning of method selectors. Like `allThings` that returns a collection of things. I always have a mental image of https://knowyourmeme.com/memes/all-the-things.

Especially for classes, as I expect implicitly that `things` return the things defined in the class and `allThings` returns things defined in class and in its super-classes. (`allThingsUpTo: nil`)

privat (approved):

jecisc:
Yes same!
I would have expected the protocols with all superclasses for this method

PR #13257, opened 2023-04-03
Rename #methodsInProtocolNamed: into #methodSelectorsInProtocol:

jecisc:
When I renamed the categories of ClassOrganization I did a mistake in the name of this method. This PR fixes this mistake.

The name is pretty long but I plan to iterate over this part of the system still and the names will improve (most #protocolNamed: will just become #protocol: and accept real instances of protocol)

I also simplified RPackage>>importProtocol:forClass: and GoferCleanup>>cleanupProtocols:

### DIFF
--- a/src/CodeExport/ClassDescription.extension.st
+++ b/src/CodeExport/ClassDescription.extension.st
@@ -7,7 +7,7 @@ ClassDescription >> fileOutChangedMessages: aSet on: aFileStream [
 
 	| org |
 	(org := self organization) protocolNames do: [ :protocolName |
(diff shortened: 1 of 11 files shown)


privat (commented):
I might suggest that you already use the final name (without `Name`). This will reduce the diff in future PR.

You can still do handle only string (for the moment), or try to accept both string and protocol if `ClassOrganization>>#methodSelectorsInProtocolNamed:` is simple to adapt

jecisc:
I'll do that

privat (approved):

MarcusDenker (approved):

PR #13259, opened 2023-04-03
Remove references to SystemOrganizer

jecisc:
SystemOrganizer will probably be removed and the new organizer will be RPackageOrganizer. 

In order to plan this change I propose to remove all the references to SystemOrganizer.

This goes with: https://github.com/pharo-spec/NewTools-DocumentBrowser

### DIFF
--- a/src/Fuel-Platform-Core/FLPlatform.class.st
+++ b/src/Fuel-Platform-Core/FLPlatform.class.st
@@ -95,25 +95,9 @@ FLPlatform class >> extensionProtocolName [
 FLPlatform class >> fixMetacello [
 ]
 
-{ #category : #accessing }
-FLPlatform class >> hacksCategoryName [
-	^ 'FuelPlatformHacks'
-]
-
 { #category : #testing }
 FLPlatform class >> isResponsibleForCurrentPlatform [
 	^ self subclassResponsibility
-]
-
-{ #category : #'private-convenience' }
-FLPlatform class >> removeModifications [
-
-	((Smalltalk globals at: #( Smalltalk globals at: #PackageInfo )) named: self extensionProtocolName) extensionMethods do: [ :methodReference |
-		methodReference actualClass removeSelector: methodReference selector ].
-	((Smalltalk globals at: #( Smalltalk globals at: #PackageInfo )) named: self hacksCategoryName) in: [ :hacks |
-		hacks classes do: [ :classOrTrait | classOrTrait removeFromSystem ].
-		(Smalltalk globals at: #PackageOrganizer) default unregisterPackage: hacks ].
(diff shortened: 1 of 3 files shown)


astares:
IDEALLY: Deprecated ... then removed ;)

jecisc:
@astares I already did a PR to add the API of SystemOrganizer to RPackageOrganizer. 

Most of the API will probably be updated to be nicer in the future and we will be able to deprecate it in RPackageOrganizer.

jecisc:
It will be hard to deprecate the SystemOrganizer in itself because we want to change directly the way things are done. 

Instead of saving a dictionary of symbols and throwing announcements everywhere, we want to make RPackage manipulate classes directly.

src/Fuel-Platform-Core/FLPlatform.class.st
```diff
@@ -107,13 +107,13 @@ FLPlatform class >> isResponsibleForCurrentPlatform [
 
 { #category : #'private-convenience' }
 FLPlatform class >> removeModifications [
-	((Smalltalk globals at: #(Smalltalk globals at: #PackageInfo)) named: self extensionProtocolName) extensionMethods do: [ :methodReference |
+
+	((Smalltalk globals at: #( Smalltalk globals at: #PackageInfo )) named: self extensionProtocolName) extensionMethods do: [ :methodReference |
```
privat: Is this code really used and functional?
`#( Smalltalk globals at: #PackageInfo )` is the literal array `#(#Smalltalk #globals #at: #PackageInfo)`, not a message send.
And it used as a key of a dictionary that usually contains symbols?

This is a lot for a monday :/

jecisc: This method is dead code so I'll just remove it :)

privat (approved):

src/Fuel-Platform-Core/FLPlatform.class.st
```diff
@@ -95,27 +95,11 @@ FLPlatform class >> extensionProtocolName [
...
-	^ 'FuelPlatformHacks'
-]
-
 { #category : #testing }
 FLPlatform class >> isResponsibleForCurrentPlatform [
 	^ self subclassResponsibility
 ]
 
-{ #category : #'private-convenience' }
-FLPlatform class >> removeModifications [
```
MarcusDenker, line 109: Fuel has all this code because it is supposed to be working on all old version of Pharo and all current and many (all?) versions of Squeak. This means of course that this "feature" at some point will cost a lot.

 Already this change is not how to do it, there are subclasses for FLPlatForm... 

And no, I do not think this makes any sense for something so dependend on implementation details to try to be compatible like that.

MarcusDenker (approved):

PR #13263, opened 2023-04-03
Snippets: split badExpressions into 3 methods

privat:
badExpressions was very large, so split it in into 3, one for tokens and literals, and one for simple expression (message sends, assignments and returns).

Diffs are big, but it is mostly code move.

Also, added nice script for each one to have a simple and fast overview.


privat:
tests are ok, windows is not

jecisc (approved):

MarcusDenker (approved):

PR #13264, opened 2023-04-03
Faulty code: improve error block node

privat:
RBErrorBlockNode are produce on block with a missing closing `]`.
Unfortunately, its contents did not include the temps possibly declared inside, and did have a bad stop value.

### DIFF
--- a/src/AST-Core-Tests/RBCodeSnippet.class.st
+++ b/src/AST-Core-Tests/RBCodeSnippet.class.st
@@ -137,7 +137,7 @@ RBCodeSnippet class >> badExpressions [
 			         notices: #( #( 1 2 4 ''')'' expected' ) )).
 		        (self new
 			         source: '[ ';
-			         notices: #( #( 1 1 3 ''']'' expected' ) )).
+			         notices: #( #( 1 2 3 ''']'' expected' ) )).
 		        (self new
 			         source: '#[ ';
 			         notices: #( #( 1 2 4 ''']'' expected' ) )).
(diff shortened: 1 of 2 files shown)


privat:
windows CI felt into a hole I presume

src/AST-Core-Tests/RBCodeSnippet.class.st
```diff
@@ -579,10 +579,10 @@ RBCodeSnippet class >> badExpressions [
 			         source: '[ | ';
 			         formattedCode: '[ | | ';
 			         notices: #( #( 3 3 5 '''|'' or variable expected' )
-				            #( 1 3 5 ''']'' expected' ) )).
+				            #( 1 4 5 ''']'' expected' ) )).
 		        (self new
 			         source: '[ | 1';
-			         formattedCode: '[ | | 1';
+			         formattedCode: '[ | | . 1';
```
jecisc, line 554: There is a dot here because you consider that the end of a unclosed block is after its temporaries declaration?

jecisc (approved):

src/AST-Core-Tests/RBCodeSnippet.class.st
privat: Bad temps are stored into a synthetic first statement of the statement sequence to prevent ast visitors to mistake them for legit temp declarations. It's a ok design I think.

The formatter was not developed to format fauly ast (sometimes even non faulty ast :p ) so it just prints both bar (because temp declarations have two bars) and a dot (because statement, even synthetic , have a dot).

I have a wip branch that tries to fix theses formatting things.
But it's more cosmetic than useful, because formatting broken ast make little sense. The idea to check it in snippets is more to verify that parts are not lost, not the punctuation.

MarcusDenker:
there is now a conflicet

src/AST-Core-Tests/RBCodeSnippet.class.st
privat: And why there wasn't a dot before the PR. Because the contents of faulty blocks were the content of the statement sequence, not the statement sequence itself. Therefore, losing possible legitimate temps, and losing the information that the contents are statements. So the content were separated with simple space instead of dots.

MarcusDenker (approved):

PR #13266, opened 2023-04-04
deprecate classesThatImplementAllOf: method. fix: #12832

vonbecmann:
deprecate ClassDescription>>#classesThatImplementAllOf: method. has no senders and a lot of critics


MarcusDenker:
Windows CI times out after 35 min

MarcusDenker (approved):
yes, it is good to deprecate that.

PR #13267, opened 2023-04-04
Remove AllProtocol

jecisc:
AllProtocol was introduced to be used in the tools to display all the methods. But now Calypso implements a more fancy mechanisms to have also "deprecated" or "overridden" categories and is not using AllProtocol. Currently this just waste memory and make the code more complexe so I propose to remove it. 

This simplifies the API of ClassOrganization and will make it easier to inline in ClassDescription in the future. 

Also, this remove a behavior that I did not like because this piece of code

```st
MyClass compile: 'toto' classified: '--all--' 
```

Was acting as if we would have written:

```st
MyClass compile: 'toto' classified: 'as yet unclassified' 
```

And I also do not like how 'as yet unclassified' is managed. Take a look at this madness we had before:

```st
MyClass compile: 'foo ^ 1' classified: #toto.
MyClass compile: 'foo ^ 1' classified: #'as yet unclassified'.
(MyClass>>#foo) protocol. "#toto"

MyClass compile: 'foo ^ 1' classified: #'-- all --'.
(MyClass>>#foo) protocol. "#toto"
```

Crazy right?

Now the behavior will be:

```st
MyClass compile: 'foo ^ 1' classified: #toto.
MyClass compile: 'foo ^ 1' classified: #'as yet unclassified'.
(MyClass>>#foo) protocol. "#toto"

MyClass compile: 'foo ^ 1' classified: #'-- all --'.
(MyClass>>#foo) protocol. "#'-- all --"
```

This is not yet as clean as I want it but it's already less magical!


src/Kernel/Protocol.class.st
```diff
@@ -1,17 +1,32 @@
 "
-A Protocol is a simple value holder representing a protocol.
-It's composed of a name and a set of method selectors
+A Protocol is a cetagorization of methods in Pharo.
```
privat: cetagorization

src/Kernel/Protocol.class.st
```diff
@@ -1,17 +1,32 @@
...
 	#instVars : [
 		'name',
 		'methodSelectors'
 	],
 	#category : #'Kernel-Protocols'
 }
 
+{ #category : #'instance creation' }
+Protocol class >> ambiguous [
+	^ #ambiguous
```
privat, line 22: ???

src/Kernel/Protocol.class.st
```diff
@@ -28,6 +43,17 @@ Protocol class >> name: nm methodSelectors: methods [
 		yourself
 ]
 
+{ #category : #accessing }
+Protocol class >> nullProtocolName [
+
+	^ #'no messages'
```
privat, line 51: This one make no sense, it's a presentation thing. Is it even used?

src/Kernel/Protocol.class.st
```diff
@@ -38,6 +64,17 @@ Protocol >> addMethodSelector: aSymbol [
 	^ methodSelectors add: aSymbol
 ]
 
+{ #category : #private }
+Protocol >> canBeRemoved [
+	^ self isEmpty
```
privat: canBeRemoved feels superfluous

src/Kernel/Protocol.class.st
```diff
@@ -47,6 +84,16 @@ Protocol >> initialize [
...
 ]
 
+{ #category : #testing }
+Protocol >> isEmpty [
+	^ self methodSelectors isEmpty
+]
+
+{ #category : #testing }
+Protocol >> isExtensionProtocol [
+	^ self name first = $*
```
privat, line 95: maybe add comment or something on added content you plan to remove or change in a foreseeable future

privat (approved):

src/Kernel/Protocol.class.st
jecisc: Thanks I'll fix that

src/Kernel/Protocol.class.st
jecisc: It is on my TODO list to check what uses that and probably kill that because I don't like it. 

But it's for another PR :)

src/Kernel/Protocol.class.st
jecisc: It is on my TODO list to check what uses that and probably kill that because I don't like it. 

But it's for another PR :)

src/Kernel/Protocol.class.st
jecisc: I'll remove it. I hesitated to remove it already and seen you think the same is enough for me to remove it :)

src/Kernel/Protocol.class.st
jecisc: All things added here are mostly push down from the abstract superclass that I removed. 
But I'll add some comment yes

jecisc:
The CI failure is related but the log produced does not help me to find the problem. 

I'll split this PR into smaller ones to find the culprit.

PR #13268, opened 2023-04-04
do not compute message sends dependencies  in browser dependencies tool

### LINKED ISSUE #13204: DABrowserToolMorph maked Calypso slow
Calypso can be pretty slow when we have a lot of classes/code.

I found that one thing making it slow is DABrowserToolMorph.

In order to setup the dependencies it looks for the implementors of some methods and this is pretty slow.

For example currently it take 0.6sec to compute those dependencies when I expand or collapse the kernel package on my machine. 0.6 sec is really noticeable and we can have bigger packages than kernel.

demarey:
do not compute message sends dependencies  in browser dependencies tool because it is slow and is done quite often within Calypso


jecisc (approved):

jecisc:
Fixes #13204

MarcusDenker:
- [CI] Win testrunner times out after 35min #13282

MarcusDenker (approved):

PR #13270, opened 2023-04-04
Announcement support clean blocks

### LINKED ISSUE #11220: [CleanBlocks] Announcements defined via # when:do:

PalumboN:
Fix https://github.com/pharo-project/pharo/issues/11220

The issue with the current Announcement API is that assumes as the subscriber _the object where the action block is created_. 

https://github.com/pharo-project/pharo/blob/d9f862003a88bfc9a17ff12ab7a11abf1fe154a9/src/Announcements-Core/AnnouncementSubscription.class.st#L99-L104

This has two main issues:
1. Could exist cases where the action block is not created in the "subscriber object" (some utils method chain).
2. It does **not work with blocks without context** (as clean or constant blocks) 😢 

So, I changed the Announcement API to send the subscriber object explicitly. Changing all
```
when: event do: action
```
by
```
when: event do: action for: subscriber
```

### Related PRs

- https://github.com/pharo-spec/Spec/pull/1374
- https://github.com/pharo-spec/NewTools/pull/502
- https://github.com/ObjectProfile/Roassal3/pull/521


src/Announcements-Core/Announcer.class.st
```diff
@@ -88,22 +88,16 @@ Announcer >> weak [
 	^ WeakSubscriptionBuilder on: self
 ]
 
-{ #category : #'registration api' }
-Announcer >> when: anAnnouncementClass do: aValuable [
```
privat, line 92: maybe deprecate it instead of removing it?

With a degraded behavior? Something like

```st
aValuable isClean ifTrue: [ self error: 'something bla bla' ].
^ self when: anAnnouncementClass do: aValuable for: aValuable receiver
```

src/Announcements-Core-Tests/AnnouncerSubscriberMockA.class.st
```diff
@@ -22,10 +22,8 @@ AnnouncerSubscriberMockA >> announcer: anAnnouncer [
 
 { #category : #events }
 AnnouncerSubscriberMockA >> registerEvents [
-	"we do not use constant blocks here, instead use a clean block.
+	"We do not use constant blocks here, instead use a clean block.
+	Using Constant blocks with announcements makes no senese, as they do nothing."
```
privat: senese (and some other, grep them)

privat (commented):
I know nothing about announcements in Pharo.

Neither <https://github.com/pharo-open-documentation/pharo-wiki/blob/master/PharoProjects/Announcer.md> nor <https://github.com/pharo-project/pharo/blob/Pharo11/doc/Announcements/Announcements.md> (why two pages?) present `when:do:` or `when:do:for:`. They only have `when:send:to:`.

So my question is why the `for:` is needed? The explanation should be present in the documentation of the method.
I imagine (but it is only a free guess, I did not look at the code) that this could be related to the registration lifespan and GC. The announcement being freed if the subscriber is reclaimed.

jecisc:
Announcements are something public so I definitely vote to deprecate the old API!
I know that I use them in some of my projects and I'm sure other people do to :)

PalumboN:
@jecisc @privat I deprecated the related methods with rule deprecations to find the correct subscriber 👍 

Now I'm checking/editing the wiki :)

Thanks for the reviews!

src/Deprecated11/Announcer.extension.st
```diff
@@ -0,0 +1,24 @@
+Extension { #name : #Announcer }
+
+{ #category : #'*Deprecated11' }
```
PalumboN, line 3: Maybe this should go to a new Deprecated12 package 🤔

jecisc: Yes because Deprecated11 was removed :)

PalumboN:
The merge with Pharo12 was breaking the image, I moved the changes to other PR -> https://github.com/pharo-project/pharo/pull/13590

Closing this

PR #13271, opened 2023-04-04
Remove references to SystemOrganization

jecisc:
SystemOrganization is a global variable holding the default SystemOrganizer. But this value can also be accessed via `Smalltalk organizer`. 

In order to avoid the multiple global variables, I am removing in the PR all references to SystemOrganization and in a future pull request I will kill this global variable and instead just save its instance un a normal instance variable of SystemDectionary, or better, SmalltalkImage if it is not to hard (because why the hell would it be the SystemDictionary that holds the PackageOrganizer?!)

Also, some classes of RPackage already have a method to access the SystemOrganizer of their organization. So I use this instead of using the global

### DIFF
--- a/src/CodeExport/RPackage.extension.st
+++ b/src/CodeExport/RPackage.extension.st
@@ -2,22 +2,21 @@ Extension { #name : #RPackage }
 
 { #category : #'*CodeExport' }
 RPackage >> fileOut [
-	| internalStream |
 
+	| internalStream |
 	internalStream := (String new: 1000) writeStream.
 
-	self classTags do: [ :each |
-		SystemOrganization
-			fileOutCategory: each categoryName
-			on: internalStream ].
+	self classTags do: [ :each | self systemOrganization fileOutCategory: each categoryName on: internalStream ].
 
 	classExtensionSelectors keysAndValuesDo: [ :className :selectors |
-		selectors do: [ :selector | | extendedClass |
+		selectors do: [ :selector |
+			| extendedClass |
 			extendedClass := self class environment classNamed: className.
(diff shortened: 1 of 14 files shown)


privat (approved):

PalumboN (approved):

MarcusDenker:
Boostrap fails with

```

MetacelloNotification: Fetched -> BaselineOfPharoBootstrapProcess-tonel.1 --- tonel:///builds/workspace/est_and_branch_Pipeline_PR-13271/64/bootstrap/src --- tonel:///builds/workspace/est_and_branch_Pipeline_PR-13271/64/bootstrap/src
MetacelloNotification: Loading -> BaselineOfPharoBootstrapProcess-tonel.1 --- tonel:///builds/workspace/est_and_branch_Pipeline_PR-13271/64/bootstrap/src --- tonel:///builds/workspace/est_and_branch_Pipeline_PR-13271/64/bootstrap/src
MetacelloNotification: Loaded -> BaselineOfPharoBootstrapProcess-tonel.1 --- tonel:///builds/workspace/est_and_branch_Pipeline_PR-13271/64/bootstrap/src --- tonel:///builds/workspace/est_and_branch_Pipeline_PR-13271/64/bootstrap/src
MetacelloNotification: Loading baseline of BaselineOfPharoBootstrapProcess...
IceGenericError: error reading from the zlib stream
IceLibgitErrorVisitor>>visitGenericError:
IceLibgitErrorVisitor>>visitERROR:
```

jecisc:
I will try to restart the ci because I would be really surprised that this is breaking this part of the bootstrap because this is code from the espell image

MarcusDenker (approved):

PR #13275, opened 2023-04-04
Remove ClassOrganization>>allProtocol

jecisc:
This is a subset of #13267 to find what part of the PR crashes the system. 
It removes the instance variable and both methods manipulating AllProtocol and replace them by #protocols instead. 

Let's see if this subset of changes breaks the system or not


jecisc:
Build is passing so it is better than the previous PR but some test failures are related. I’ll check to fix them soon

privat (approved):

PR #13277, opened 2023-04-04
Faulty code: move code reparator to RBNotice

privat:
This PR is a simple cleanup, but I do not like big PR.

The original idea was to simplify `OpalCompiler >> checkNotice:` by removing conditions and type checks.
A simple solution was to move the reparation concern to the RBNotice.

Currently, OCCodeReparator is still the fragile and limited class that was extracted from the original OCUndeclaredVariableWarning, but there is no reason that clients have to deal with too much detail about that.
So let's imagine that any RBNotice might propose some possible ways to repair itself. It simplifies the client code that do not have to deal about specific errors and warnings.

PR #13278, opened 2023-04-04
hotfix: add missing return in ClassOrganization>>#ensureProtocol:

privat:
Fix unrelated CI issues, for instance seen in #13277


jecisc (approved):
Indeed, this might be the origin of some of my problems too :)
Thanks!

jecisc:
I’ll add on my todo to add a test on this scenario also

PR #13279, opened 2023-04-04
Bootstrap: allow to authetificate github requests during the bootstrap

jecisc:
On thing we might test to fix the Github limit rate problem on the CI we can try to authenticate the requests our github request. 

This PR adds this possibility. If GITHUB_TOKEN and GITHUB_USER environment variables are present, they will be used to build the image. And it will remove this credential at the end. Next step is to add credentials on Jenkins

### DIFF
--- a/src/BaselineOfIDE/BaselineOfIDE.class.st
+++ b/src/BaselineOfIDE/BaselineOfIDE.class.st
@@ -265,13 +265,27 @@ BaselineOfIDE >> loadIceberg [
 		load.
 	(Smalltalk classNamed: #Iceberg) enableMetacelloIntegration: true.
 
+	Smalltalk os environment at: #GITHUB_TOKEN ifPresent: [ :token |
+		| credentials |
+		credentials := (Smalltalk classNamed: #IceTokenCredentials) new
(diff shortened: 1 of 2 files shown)


privat (approved):
Nice. And this seems easier than I imagined.

Note: maybe we should check that credentials does not leak in the produced image: GC is unreliable, and even if strings are collected, raw bytes with secrets might be not overwritten and still present in the file.

MarcusDenker (approved):

jecisc:
Next step on this one would be to generate a personal github token and set it in the Jenkins build environment variables. Probably via the Credential Plugin to not have it visible anywhere. 

Once this is done we can restart this PR to see if there is any problem with that.

jecisc:
> Nice. And this seems easier than I imagined.
> 
> Note: maybe we should check that credentials does not leak in the produced image: GC is unreliable, and even if strings are collected, raw bytes with secrets might be not overwritten and still present in the file.

It should be possible to create a read only token on github and since Pharo is open source then we do not have to care that someone could find it while looking at the raw memory state

PR #13280, opened 2023-04-04
Faulty code: add permitUndeclared

privat:
~~This PR is based on #13277 (6 first commits).~~ (was merged)

Currently, the static status of undeclared variables is not great. Basically, they are statically accepted unless some heuristics in the compiler decided that you are in an interactive text editor or something, and you get a GUI menu instead.
The lack of a precise specification of the Pharo language cause confusion for users (or at least for myself).

Unfortunately, recent experiments (#13172 then #13244) showed that some important parts of Pharo rely on variables statically undeclared (with the hope that they will be repaired later), including CodeImport, Monticello and Traits.
Therefore, currently, forbidding globally the static use of undeclared variable seems very hard.

So, I would like to try another (easier) approach: do not let the compiler decide with some uncontrollable heuristic but make it an explicit parameter: `permitUndeclared`.

The PR defines the infrastructure for it, and sets up a default value that is (I hope) sufficiently compatible with the current policy. Now the heuristic is at a single place, at least :)

Future PR will then be able to update clients to use a better explicit value and drop the heuristic to use a sane static default value. I would like that `permitUndeclared` be false by default unless specific tool require it to be true, but we'll see.

Note: I initially wanted `permitFaulty` to play the role of `permitUndeclared`, but this was complex to have without breaking too many things (mostly tests), and possibly introducing unexpected side effects and complex bug.
So maybe in the long run `permitUndeclared` will be merged with `permitFaulty`, but I do not know yet.

### DIFF
--- a/src/AST-Core/RBNotice.class.st
+++ b/src/AST-Core/RBNotice.class.st
@@ -45,6 +45,12 @@ RBNotice >> isError [
 ]
 
 { #category : #testing }
+RBNotice >> isUndeclaredNotice [
+
+	^ false
+]
+
+{ #category : #testing }
 RBNotice >> isWarning [
 
 	^ false
--- a/src/OpalCompiler-Core/CompilationContext.class.st
+++ b/src/OpalCompiler-Core/CompilationContext.class.st
@@ -35,7 +35,8 @@ Class {
 		'bindings',
 		'compiledMethodClass',
 		'semanticScope',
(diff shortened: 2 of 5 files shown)


privat:
Tests passes, except windows (as usual) and unrelated issue #13281

MarcusDenker (approved):

Issue #13246, opened 2023-04-01
Class comments can get lost when commiting code


jecisc:
I recently noticed that in different commits some of my class comments are vanishing. 

For now I don't know how this happens.

privat:
I noticed the same behavior. I do not recall what I did with the classes that lost their comments.

MarcusDenker:
This could related to moving the commentSourcePointer to ClassDescription, I will have a look

privat:
I found a way to reproduce.

* Choose a class you like with a nice comment
* Go to the class definition panel
* Add an ivar, and accept
* Now the comment is dead in the class and in all its subclasses

privat:
I now have a bunch of changes that are a pain to commit because a lot of class comments are lost :(
Maybe I will commit them as is and try to recover the lost comments with command-line git-fu

MarcusDenker:
I am working on fixing this

PR #13286, opened 2023-04-05
Faulty code: don't permit undeclared by default

privat:
This PR defaults `permitUndeclared` to false, meaning that unless instructed otherwise with the setter `permitUndeclared:`, `OpalCompiler` will consider the (static) use of undeclared variables to be errors.
Moreover, `OCUndeclaredVariableNotice` is officially promoted to `RBErrorNotice`. Until now, it was in some limbo state, neither an error nor a warning. This change means that an AST node with such a notice will answer true to `isFaulty`.

In order for this change to be successful, the following clients are adapted:
* Traits: `TaAbstractComposition` sets permitUndeclared
* Monticello: `MethodAddition` sets permitUndeclared
* All clients of `ClassDescription>>compile:*`, that includes CodeImport and some tests in newtools, permitUndeclared is set according to a heuristic (is a requestor present) that is comparible with the current usage.
  I think that's the right move since I consider `ClassDescription>>compile:*` a legacy API, so maintaining its behavior make sense.

Tests are also added. `RBCodeSnippet` gains a new ivar `isFaultyMinusUndeclared` (the name is not nice, but it is not used that much) that correspond to being faulty, while `permitUndeclared` is set to true.
It is used for the Monticello test on snippets and for a new test explicitly made to check `permitUndeclared:`

There are many small changes with specific goals, so there are a lot of commits (oneliner sometime).
The global diff is less big than I expected.

### DIFF
--- a/src/AST-Core-Tests/RBCodeSnippet.class.st
+++ b/src/AST-Core-Tests/RBCodeSnippet.class.st
@@ -299,12 +299,14 @@ RBCodeSnippet class >> badExpressions [
 			         source: '| | a';
 			         formattedCode: 'a';
 			         isParseFaulty: false;
+			         isFaultyMinusUndeclared: false;
 			         value: nil;
 			         notices: #( #( 5 5 5 'Undeclared variable' ) )). "This one is legal, it is an empty list of temporaries, the | are dismissed"
 		        (self new
 			         source: '|| a';
 			         formattedCode: 'a';
 			         isParseFaulty: false;
+			         isFaultyMinusUndeclared: false;
 			         value: nil;
 			         notices: #( #( 4 4 4 'Undeclared variable' ) )). "Same, but are messing with the Scanner"
 		        (self new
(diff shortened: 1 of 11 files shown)


jecisc:
Hi,
Can you elaborate on ClassDescription>>compile:* been legacy? Do you have plans to use something else in the future? 

I have the impression that this is the entry point of most method compilation happening in the image no?

privat:
> Can you elaborate on ClassDescription>>compile:* been legacy? 

* Receiver (the class) is dubious
* Name `compile` is bad, because it also installs
* Returns a symbol. cf #12880
* Not configurable with optimization options or faulty-modes
* Most of them ask you for a requestor, but requestors are bad
* Hard to extend: adding a new parameter double the number of selectors

For me, `ClassDescription>>compiler` is the way forward.

privat:
Bootstrap passed :tada:, but test failed :sob:  because I fumbled a rebase trying to restore some lost comments (cf #13246). So let's try again

privat:
Tests are ok, Windows is still running, but who cares?

MarcusDenker (approved):

PR #13287, opened 2023-04-05
Fix some deprecated sends

MarcusDenker:
Fix some cases where deprecation refactorings are happening while running the tests


privat (approved):

jecisc (approved):
Thanks! 
I think I did those changes alongside the deprecation but I have the impression that in some cases Iceberg is not notified of some changes

jecisc:

PR #13292, opened 2023-04-05
enable Fluid Class Definition by default

MarcusDenker:
This PR makes the fluid definitions default. But it still can be turned off

### DIFF
--- a/src/Kernel/ClassDefinitionPrinter.class.st
+++ b/src/Kernel/ClassDefinitionPrinter.class.st
@@ -161,7 +161,7 @@ Object << #Faked
 { #category : #configure }
 ClassDefinitionPrinter class >> showFluidClassDefinition [
 
-	^ ShowFluidClassDefinition ifNil: [ ShowFluidClassDefinition := false ]
+	^ ShowFluidClassDefinition ifNil: [ ShowFluidClassDefinition := true ]
 ]
 
 { #category : #configure }


privat (approved):

jecisc:
One test failed on linux but not osx.

testChangeClassCategoryAddsNewChangeRecord – Unix64.System.Changes.Tests.Base.ChangeSetClassChangesTest

*But* it does print a fluid class definition. So it might be related.

```
Got 'Object << #ClassForTestToBeDeleted247  slots: {};  package: ''CategoryForTestToBeDeleted-TestPackage1''' instead of 'Object << #ClassForTestToBeDeleted247  slots: {};  package: ''TestPackage2'''.
Stacktrace
TestFailure
Got 'Object << #ClassForTestToBeDeleted247
	slots: {};
	package: ''CategoryForTestToBeDeleted-TestPackage1''' instead of 'Object << #ClassForTestToBeDeleted247
	slots: {};
	package: ''TestPackage2'''.
ChangeSetClassChangesTest(TestAsserter)>>assert:description:resumable:
ChangeSetClassChangesTest(TestAsserter)>>assert:description:
ChangeSetClassChangesTest(TestAsserter)>>assert:equals:
ChangeSetClassChangesTest>>assertDefinition:equivalentTo:
ChangeSetClassChangesTest>>testChangeClassCategoryAddsNewChangeRecord
ChangeSetClassChangesTest(TestCase)>>performTest
```

Ducasse:
I took latest P12 + changed to fluid as marcus did and the test passes

Ducasse:
I do not understand why it is passing on Mac and not on Linux!

Ducasse:
I do not know how to reproduce it. May be this is problem with cr/lf?

Ducasse:
I do not get why we get something so different for packages
TestPackage2 vs. CategoryForTestToBeDeleted-TestPackage1

MarcusDenker:
on a second run, the tests are green

MarcusDenker:
But fails randomly.. as the CI is already too unstable I revert

PR #13297, opened 2023-04-05
Faulty code: make undeclared write resumable with the right value

privat:
Kernel runtime error `UndeclaredVariableWrite` is now resumable, and keep the right value (pun on *correct* and *rvalue*). The value is stored in the exception, so easy to access if someone cares.

This is a different (and better) implementation than the one in (not yet merged) PR #13238.
And also contains a test.


jecisc (approved):
I like this approach

MarcusDenker (approved):

MarcusDenker:
ups, too early merge

MarcusDenker:
but the test is green of the CI on the mac and linux, so it looks good

PR #13298, opened 2023-04-05
Do not disable system announcements during rpackage tests

jecisc:
RPackage tests were disable the system announcer but this causes multiple problems such as:
- Automatic transformation was happening but no tool were aware of it (Calypso did not update the source, Iceberg did not catch the change...)
- Coding in debugger was dangerous because no system announcements were happening 
- The system was vulnerable during the execution of those tests

So I propose to just keep the system announcer while testing RPackage

Also I shipped some deprecation rewrites now that Iceberg notice them :)

### DIFF
--- a/src/Monticello-Tests/RPackageClassesSynchronisationTest.class.st
+++ b/src/Monticello-Tests/RPackageClassesSynchronisationTest.class.st
@@ -494,36 +494,37 @@ RPackageClassesSynchronisationTest >> testReorganizeClassByRemovingExtensionCate
 ]
 
 { #category : #'tests - operations on protocols' }
-RPackageClassesSynchronisationTest >> testReorganizeClassByRenamingClassicCategoryToExtensionCategoryMoveAllMethodsFromParentPackageToExtendingPackage [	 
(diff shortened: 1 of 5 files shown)


privat (approved):
This is interesting. We want announcements to not necessarily be global (especially for tests), but in practice it's hard to achieve.

jecisc:
Yes!

The problem here is that the way to not do system announcement impact everything.

Ideally, you want an announcer specific for the code generated by your new package organizer and keep the global one for the code of the tests itself.

Maybe a solution would be to make RPackageOrganizer hold an announcer and use this one. Default RPackageOrganizer would use the system announcer and the test would have their own instance that is not the global one?

jecisc:
Let's note that for now this is hard to do because it's SystemOrganizer that launches the announcements. But when RPackageOrganizer will replace SystemOrganizer, it should be easy not to use global announcer everywhere :)

jecisc:
testOnlyWeakSubscriptions fails and seems to be related.

jecisc:
RPackage tests where using strong subscription to announcements. I added a fix to make all subscriptions weak

MarcusDenker (approved):

PR #13299, opened 2023-04-05
Use date and time as Stdio windows file name distinguisher

jecisc:
Stdio on windows will sometimes create a new file and since we can have multiple ones, we were generating a random number in the file name. Since the is the only dependency to Random-Core and that I would like to extract it from the KernelGroup of the bootstrap, I propose to use a timestamp instead.

I also added a test to be sure the dependency does not get added again. One last change is the update of the manually resolved dependencies of Files package

### DIFF
--- a/src/System-DependenciesTests/SystemDependenciesTest.class.st
+++ b/src/System-DependenciesTests/SystemDependenciesTest.class.st
@@ -593,6 +593,16 @@ SystemDependenciesTest >> testExternalUIDependencies [
 	self assertCollection: dependencies hasSameElements: self knownUIDependencies
 ]
 
+{ #category : #tests }
+SystemDependenciesTest >> testFilesShouldNotDependOnRandom [
+	"In the past Files was depending on Random-Core only to print a random number in a file name.
+	Now, there is no dependency between Files and Random-Core and if possible we would like to keep it this way in order to extract Random-Core from the KernelGroup of the Bootstrap (and maybe of the bootstrap in itself generally)"
+
+	| dependencies |
+	dependencies := self externalDependendiesOf: #( 'Files' ).
+	self deny: (dependencies includes: 'Random-Core')
+]
+
 { #category : #accessing }
 SystemDependenciesTest >> tonelCorePackageNames [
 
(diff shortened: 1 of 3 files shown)


src/System-DependenciesTests/SystemDependenciesTest.class.st
```diff
@@ -593,6 +593,16 @@ SystemDependenciesTest >> testExternalUIDependencies [
...
 ]
 
+{ #category : #tests }
+SystemDependenciesTest >> testFilesShouldNotDependOnRandom [
+	"In the past Files was depending on Random-Core only to print a random number in a file name.
+	Now, there is no dependency between Files and Random-Core and if possible we would like to keep it this way in order to extract Random-Core from the KernelGroup of the Bootstrap (and maybe of the bootstrap in itself generally)"
+
+	| dependencies |
+	dependencies := self externalDependendiesOf: #( 'Files' ).
+	self deny: (dependencies includes: 'Random-Core')
```
privat, line 603: This feels weird. If you care about random things not being pulled. Why not just list the dependencies you want instead of blacklisting some that you do not want?

privat (approved):

src/System-DependenciesTests/SystemDependenciesTest.class.st
jecisc: That is something that I would like to do later but it requires to check the current dependencies of Files and think about which ones make sense or not and for now I just wanted to focus the extraction of Random-Core from the bootstrap core

jecisc: Some things seems simple, but then I end up spending hours in simple things and not finishing my original fix x_x

Ducasse:
cyril can we use the nextFileNamed utils in filereference? Because like that it numbers the file automatically

jecisc:
I am not sure we have a file reference of the previous stdio here so I went with the time stamp.

But I don’t think it will cause too much trouble to the user because this is a hack anyway happening only on Windows when PharoConsole.exe is not used. And people will even be able to find easily the most recent one since it uses a file compatible timestamps

PR #13300, opened 2023-04-05
Faulty parser: better faulty literal (byte) arrays

privat:
`RBArrayErrorNode` (and `RBByteArrayErrorNode`) were created to handle (byte) arrays with missing closer (or opener).
Therefore, on `#[ 0 foo 2 ]` where the error is not on the brackets, but on the contents, an additional and wrong error message was produced ("*] expected*").

Producing a valid `RBLiteralArrayNode` with bad contents did not work because clients (many known ones and possible unknown ones) assumed that `RBLiteralArrayNode` always have a legal literal value.
Various alternative solution were tried, but not with more success.

The proposed solution is to still have a `RBArrayErrorNode` (or `RBByteArrayErrorNode`) but without an error notice attached.
While possibly unintuitive, it allows an error node to not have specific error information by itself, but to contain nodes that have.

### DIFF
--- a/src/AST-Core-Tests/RBCodeSnippet.class.st
+++ b/src/AST-Core-Tests/RBCodeSnippet.class.st
@@ -162,32 +162,25 @@ RBCodeSnippet class >> badExpressions [
 			         source: '#( 0 1r2 4 )';
 			         formattedCode: '#( 0 1 r2 4 )';
 			         notices:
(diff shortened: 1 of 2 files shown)


privat:
Windows is out, others are in

MarcusDenker (approved):

Issue #7233, opened 2020-09-02
Parse expression should never return a sequence node


guillep:
**Describe the bug**

`RBParser >> parseExpression:` could return a sequence node.

**To Reproduce**

```smalltalk
RBParser parseExpression: '
  1+2.
  2+3.'
```

**Expected behavior**
An error

**Version information:**
Pharo 9.0.0
Build information: Pharo-9.0.0+build.634.sha.658326a3eb4f26b45dc0efe988aa945d5470876b (64 Bit)

MarcusDenker:
I think this is because for a "Doit" or "PrintIt" you actually can select multiple lines (more than an expression).

This is parsed with parseExpression:, 

I guess this is why the ST80 Compiler used  "parse without method pattern" as a terminology... (see #noPattern:)

privat:
Not sure if it's a bug. Bad terminology at most.

"Expression" seems to be user rather broadly in Pharo as any piece of code you might execute or expect a value. This could be a contamination from languages where sequences of expressions are also expressions.
What was the context where you wanted a real expression and not a sequence (if you remember)?

To add to the confusion, the parser also offers a `parseDoIt` that injects a synthetic return node at the last statement.
`parseDoIt` seems to be the one used for `DoIt` or `PrintIt` and for `OpalCompiler#>>evaluate:`

The terminology "noPattern" is so much bad that it might underflow.

Calypso, Rubric and Spec2-Code seems to use the terminology `isScripting` (only in a boolean form?). That is not that bad.
But I'm not sure if it is worth the hassle to rename for the sake of renaming, but it could be an opportunity in case there is an unrelated API change.

hernanmd:
I get a ` "RBSequenceNode(1 + 2.
2 + 3)"`

So this could be closed (or renamed)?

PR #13301, opened 2023-04-05
Vocabulary: propose "script" in compiler

privat:
`OpalCompiler>>noPattern` is historic and bad:

 * it conflicts with *pattern code* the thing with ``` ` ``` and `@` to match AST (in rewriting tools for instance)
 * it has a negation, and you should not have a negation in the name of boolean flag
 * I'm not sure that all people understand what the method pattern is
 
I tried to look at various parts of the Pharo tools to see what is the used vocabulary.

* `RBParser` use `expression`, like in `parseExpression` That is ok but misleading since sequences of statements (`|x| x:=self. x bar`) or return statements (`^ 42`) are parsed the same but technically aren't expressions. See  #7233
* `ExpressionEvaluated` is a `SystemAnnouncement` used when someone runs a DoIt (or other action). (I do not understand why you need to announce them, but it's not the focus here)
* `CodeImport` use `DoIt`, like in `DoItChunk`. That is better than `noPattern` but maybe too specific?
* Various GUI editor seems to use `isScripting`, or `beForScripting` I kind of like it (see title)
* `Monticello` use `script`, like in `MCScriptDefinition` and related
* but `Metacello` seems to use `DoIt`, as in `preLoadDoIt`
* `SHRBTextStyler` have `isForWorkspace`, that can be used to highlight source code outside the workspace :) and seems the only one to use "workspace" this way.
* Surprisingly, `playground` seems not used
* `script:` is also a pragma, that takes a string to evaluate

Since I'm cleaning up the compiler API, I would like to have your input.
And if nobody cares, I will go ahead with "script".
The commit is a simple preview of what it might look like.

### DIFF
--- a/src/Debugger-Model-Tests/StepIntoTest.class.st
+++ b/src/Debugger-Model-Tests/StepIntoTest.class.st
@@ -42,7 +42,7 @@ StepIntoTest >> testStepIntoDoIt [
 
 	| method node |
 	method := nil compiler
-		          noPattern: true;
+		          isScripting: true;
 		          compile: '| a b | a := 40. b := a + 2. ^ b'.
 	self settingUpSessionAndProcessAndContextForBlock: [
 		nil executeMethod: method ].
--- a/src/Debugger-Model/DebugContext.class.st
+++ b/src/Debugger-Model/DebugContext.class.st
@@ -156,7 +156,7 @@ DebugContext >> recompileCurrentMethodTo: aText notifying: aNotifyer [
 	self context method isInstalled ifFalse: [
 		^ self selectedClass compiler
 			protocol: self selectedMessageCategoryName;
-			noPattern: self selectedMessageName isDoIt;
+			isScripting: self selectedMessageName isDoIt;
 			requestor: aNotifyer;
 			compile: aText.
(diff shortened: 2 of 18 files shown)


jecisc:
Workspace is the ancestor of the playground.
It would not surprise me that #beForWorkspace was to highlight the code of this old playground. 

I don’t like #noPattern either and prefer what you proposed. I like the “scripting” vocabulary!

jecisc (approved):

Ducasse:
Yes script or snippet is good. 

It would be good to effectively clean all the API because this is confusing in many places.
Thanks for all the jobs you are doing.

MarcusDenker (approved):
Yes, it is good to unify the terms. Both "notPattern" and "Workspace" are not good. "script" is nice (and we use it already in some places)

MarcusDenker:
There is now a conflict

privat:
I did a rebase with improvements:

* adapt clients
* add deprecation
* add Context>>compiler and use it

privat:
Tests are green, please review

MarcusDenker (approved):

PR #13302, opened 2023-04-05
Some compiler cleanup (remove unused methods and classes)

privat:
Remove unused methods and classes

### DIFF
--- a/src/OpalCompiler-Core/Behavior.extension.st
+++ b/src/OpalCompiler-Core/Behavior.extension.st
@@ -131,9 +131,3 @@ Behavior >> recompileBasic: selector from: oldClass [
 	selector == newMethod selector ifFalse: [self error: 'selector changed!'].
 	^ newMethod
 ]
-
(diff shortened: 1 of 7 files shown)


privat:
All tests green (and windows gray)

src/OpalCompiler-Core/Behavior.extension.st
```diff
@@ -67,6 +67,7 @@ Behavior >> compilerClass [
 
 { #category : #'*OpalCompiler-Core' }
 Behavior >> evaluate: aString [
+	self deprecated: 'You should not even use it as `self` does not impact the result'.
```
jecisc:
```suggestion
	self deprecated: 'You should not even use it as `self` does not impact the result' transformWith: '`@rcv evaluate: `@arg' -> '`@rcv compiler evaluate: `@arg'.
```

We could maybe add a transformation

jecisc (approved):

MarcusDenker (approved):

PR #13303, opened 2023-04-05
Compiler simplify OCSemanticScope

privat:
The class hierarchy `OCSemanticScope` is used to abstract the contextual semantic of a source code: method definition, basic script, or advanced script in a context (for debugger).

For some reason, the current code is far more complex than required. This PR tries to simplify it.

### DIFF
--- a/src/OpalCompiler-Core/CompilationContext.class.st
+++ b/src/OpalCompiler-Core/CompilationContext.class.st
@@ -542,11 +542,9 @@ CompilationContext >> noPattern [
 
 { #category : #accessing }
 CompilationContext >> noPattern: anObject [
-	anObject ifTrue: [
-		semanticScope := semanticScope asDoItScope.
-		"when we compile an expression, embedd sources, as the resulting method will never be installed"
-		self parseOptions: #(+ #optionEmbeddSources).
-	 ]
+
+	(anObject and: [ self noPattern not ]) ifTrue: [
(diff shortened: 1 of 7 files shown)


privat:
* all tests: green
* windows: U

src/OpalCompiler-Core/CompilationContext.class.st
```diff
@@ -542,11 +542,9 @@ CompilationContext >> noPattern [
 
 { #category : #accessing }
 CompilationContext >> noPattern: anObject [
-	anObject ifTrue: [
-		semanticScope := semanticScope asDoItScope.
-		"when we compile an expression, embedd sources, as the resulting method will never be installed"
-		self parseOptions: #(+ #optionEmbeddSources).
```
jecisc, line 548: Is this parse option removal intended?

jecisc (approved):
This seems easier to read indeed

src/OpalCompiler-Core/CompilationContext.class.st
privat: Yep, it is wrongly placed and duplicates one that is correctly placed: https://github.com/pharo-project/pharo/blob/Pharo12/src/OpalCompiler-Core/OpalCompiler.class.st#L335

MarcusDenker (approved):

MarcusDenker:
there is no a conflict

PR #13306, opened 2023-04-06
Remove SystemOrganisation global

jecisc:
We have a SystemOrganization global variable (for now... It's the next one to kill on my list) but a type in the Bootstrap creates a SystemOrganisation (with an s instead of the z) global.

This removes this unwanted global.

### DIFF
--- a/src/Pharo30Bootstrap/PBImageBuilder50.class.st
+++ b/src/Pharo30Bootstrap/PBImageBuilder50.class.st
@@ -289,7 +289,7 @@ PBImageBuilder50 >> createInitialObjects [
 	self bootstrapInterpreter evaluateCode:
 		'Smalltalk globals at: #Transcript put: nil.'.
 	self bootstrapInterpreter evaluateCode:
-		'Smalltalk globals at: #SystemOrganisation put: nil.'.
+		'Smalltalk globals at: #SystemOrganization put: nil.'.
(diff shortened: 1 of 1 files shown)


src/Pharo30Bootstrap/PBImageBuilder50.class.st
```diff
@@ -289,7 +289,7 @@ PBImageBuilder50 >> createInitialObjects [
 	self bootstrapInterpreter evaluateCode:
 		'Smalltalk globals at: #Transcript put: nil.'.
 	self bootstrapInterpreter evaluateCode:
-		'Smalltalk globals at: #SystemOrganisation put: nil.'.
+		'Smalltalk globals at: #SystemOrganization put: nil.'.
```
privat: I think you should just remove it since it worked without the correct name

Ducasse:
Build is broken

src/Pharo30Bootstrap/PBImageBuilder50.class.st
jecisc: Using the right global seems to break the bootstrap anyway so let's remove it yeah

MarcusDenker:
We might add a special DeprecatedGlobal later to warn clients that use it, but for early Pharo12 this is ok to remove it

MarcusDenker (approved):

PR #13308, opened 2023-04-06
Skip windows broken tests

### LINKED ISSUE #13282: [CI] Win testrunner times out after 35min
**Bug description**
The test runner for Windows times out (35min) while running the tests.

Seen on both Pharo11 and Pharo12

jecisc:
The windows CI is ending up in timeout since the latest used VM because two tests and ending un OutOfMemory and the VM currently opens a native window blocking the execution of Pharo.

This skips those two tests until this issue is fixed (And it should get fixed before P11 release in my opinion because that means that first it is impossible to allocate 1Go on windows but also if that happens the execution get blocked. This is a show stopper for me

Fixes #13282

### DIFF
--- a/src/Kernel-Tests/AllocationTest.class.st
+++ b/src/Kernel-Tests/AllocationTest.class.st
@@ -12,10 +12,14 @@ AllocationTest >> testOneGBAllocation [
 	"Documentating a weird bug in the allocator"
 
 	| sz array failed |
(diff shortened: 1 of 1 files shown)


MarcusDenker (approved):

astares:
@jecisc is there already an issue for "pharo-vm"?

jecisc:
No idea for Pharo-vm but I opened one here with the P11 tag and Pablo knows

PR #13305, opened 2023-04-06
Multiple cleaning of Protocol

jecisc:
- Remove Protocol class>>nullProtocolName because it is useless
- Rename #ambiguous to #traitProtocolConflictName and move it to TraitV2 package
- #canBeRemoved has been removed (yes, you, you can be removed ;) ) and #isEmpty is used instead. Those are equivalent now that AllProtocol magic has been removed
- Class comment has been improved
- I added some comments on methods has #isExtensionProtocol
- #canBeRenamed and #isVirtualProtocol were dead code
- Protocol class>>#defaultName has been removed in favor of #unclassified
- Some small other code cleanings

PR #13311, opened 2023-04-06
Remove usage of Protocol>>#canBeRemoved

jecisc:
This is a subset of #13305 to find which part of the PR crashes the bootstrap.

#canBeRemoved was a method present to deal with virtual protocols. Since virtual protocols do not exist anymore, we can just remove this method and use #isEmpty instead because we can remove a protocol only if it is empty. The code is still readable and explicit enough. For example:

```st
oldProtocols
		select: #isEmpty
		thenDo: [ :aProtocol | self removeProtocol: aProtocol ].
```

I think it is easy to understand that we want to remove the empty protocols only

### DIFF
--- a/src/Refactoring2-Transformations/RBRemoveProtocolTransformation.class.st
+++ b/src/Refactoring2-Transformations/RBRemoveProtocolTransformation.class.st
@@ -54,9 +54,6 @@ RBRemoveProtocolTransformation >> preconditions [
 			includes: protocol ]
 			errorString: 'Protocol named ', protocol, ' does not exist' )
 		& ( RBCondition
-			withBlock: [ (self definingClass realClass organization protocolNamed: protocol ) canBeRemoved ]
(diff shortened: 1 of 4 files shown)


src/Refactoring2-Transformations/RBRemoveProtocolTransformation.class.st
```diff
@@ -53,9 +53,6 @@ RBRemoveProtocolTransformation >> preconditions [
 			protocolsFor: self definingClass realClass )
 			includes: protocol ]
 			errorString: 'Protocol named ', protocol, ' does not exist' )
-		& ( RBCondition
-			withBlock: [ (self definingClass realClass organization protocolNamed: protocol ) canBeRemoved ]
-			errorString: 'Protocol named ', protocol, ' is a virtual protocol and it cannot be removed' )
```
jecisc, line 58: I removed this condition because virtual protocols do not exist anymore

MarcusDenker (approved):

Issue #13310, opened 2023-04-06
Deprecate OneShotArmed / doOnlyOnce: mechanism in ProtoObject


astares:
There is an old mechanism using a global called #OneShotArmed to run a block only once. It's ugly, has to be done manually (one shot and rearm) and has no senders in the image.

We should remove it without any replacement.
I propose to
 - deprecate it for Pharo 11
 - and remove it in Pharo 12 iteration

Implementation is in 
 - ProtoObject>>#doOnlyOnce:
 - ProtoObject>>#rearmOneShot
 - DoOnlyOnceIconStyler
which are all unused

jecisc:
P11 is about to be released. 
I would prefer to deprecate in P12 and remove in P13

astares:
@jecisc  The deprecation does not change any behavior and it is a good time to have the deprecation now . So ideally we can get it in.

Doing it with the next would mean to wait another full iteration and in worst case we forget to remove in P13 (as it is in ProtoObject and the #deprecation methods are in Object you will not see it striked through). 

My thinking is:
 - new features should go into new P12
 - bugfixing AND deprecations can still be applied to P11

jecisc:
I don't like to integrate deprecation so close to the release because in some cases it can lead to problems. 

Just yesterday we found that a deprecation was causing problems in Pharo 12 because RPackage was disabling system announcement for its tests, but the deprecation mecanism was sending an announcement for the tools to act on its behavior. Since the announcement were disabled, the tools never got the announcement and the tools ended up in a bastard state and cause some really weird behavior like Iceberg not noticing some code changes to commit.

The release should happen really really soon, I prefer that we focus on bug fixes.
We lived with this method in the image since years now I guess. We can live with it for another Pharo iteration.

astares:
As written, there are no references/senders. Very safe. So no side effect to expect.
I personally would put it in P11 so people who might use it can already see from the deprecation that it will be removed and react. It would give the chance to get rid of it earlier and as long as release is prepared deprecations should be allowed. 

But I'm fine with the P12/P13 solution as well.

PR #13312, opened 2023-04-06
Deprecate OneShotArmed / doOnlyOnce: mechanism in ProtoObject (P11)

astares:
Deprecation for issue #13310


astares:
We will deprecated in P12 and delete in P13. See https://github.com/pharo-project/pharo/issues/13310
So this PR is invalid

PR #13313, opened 2023-04-06
Revert "enable Fluid Class Definition by default"

MarcusDenker:
Reverts pharo-project/pharo#13292


jecisc (approved):
This cause the CI to be unstable and I cannot reproduce the problem locally. 

Reverting is good in that case I think

PR #13314, opened 2023-04-06
Remove the need to give a SystemOrganizer to SystemDictionary

jecisc:
SystemDictionary needs a SystemOrganizer to add or remove classes. Currently, users creating a SystemDictionary are creating this by hand. This PR simplifies it by doing that during the initialization.

This also remove a bunch of other SystemOrganizer references a little everywhere in the system and the bootstrap which will help to replace the SystemOrganizer by a RPackageOrganizer.


jecisc:
This seems to cause an infinit loop somewhere

jecisc:
Ohh this highlight a bug that was already present.

SystemDictionary references SystemOrganizer and vice versa. 
There is a tests checking that we can use #storeOn: on a SystemDictionary but it does the test on a partially initialized SystemDictionary that does not have a SystemOrganizer yet.

MarcusDenker (approved):

Issue #13318, opened 2023-04-06
RBVariableNode with comment goes 1 character too far 


jecisc:
I have a piece of code with a variable in parenthesis and a comment. When that happens, the RBVariableNode produced in the AST has a stop going one character too far as you can se here:

<img width="1440" alt="image" src="https://user-images.githubusercontent.com/9519971/230359513-43ce29c9-4013-4a96-afd0-f1d707205b92.png">

I expect the close parenthesis not to be included in the variable node since the opening parenthesis is not.

Maybe this can interest @privat

privat:
Interesting...

privat:
The location on the node is off. I will look at that. I think there might be related issues for other kind of nodes.

Issue #12942, opened 2023-03-08
AST visitor does not visit comments


jecisc:
It might be good to have a #visitCommentNode: in the AST visitor

PR #13319, opened 2023-04-06
Fix bug in syntaxt highlighting of comments

### LINKED ISSUE #13316: Bug in syntax highlight
In some cases the syntax highlight can be off. For example:

image

I do not know yet of this is a syntax highlighter but or a parser bug producing a bad AST

jecisc:
If we put a comment next to a variable in parenthesis, then the comment is part of the VariableNode in the AST.  Shout was not hanlding such comments until now.

This fixes the problem by visiting the comments of the variable nodes of the AST and adds a regression test.

Fixes #13316

Let's note that there is still a bug there because the closing parenthesis is considered part of the variable node. But this is not a but of Shout but of the parser (cf. https://github.com/pharo-project/pharo/issues/13318)

### DIFF
--- a/src/Shout/SHRBTextStyler.class.st
+++ b/src/Shout/SHRBTextStyler.class.st
@@ -1101,10 +1101,10 @@ SHRBTextStyler >> visitSequenceNode: aSequenceNode [
 
 { #category : #visiting }
 SHRBTextStyler >> visitVariableNode: aVariableNode [
-	self
-		addStyle: (self resolveStyleFor: aVariableNode)
-		attributes: (self resolveVariableAttributesFor: aVariableNode)
-		forNode: aVariableNode
+
+	self addStyle: (self resolveStyleFor: aVariableNode) attributes: (self resolveVariableAttributesFor: aVariableNode) forNode: aVariableNode.
+
+	self visitMethodComments: aVariableNode comments
 ]
 
 { #category : #accessing }
--- a/src/Shout/SHRBTextStyler.class.st
+++ b/src/Shout/SHRBTextStyler.class.st
@@ -966,6 +966,12 @@ SHRBTextStyler >> visitCascadeNode: aCascadeNode [
 ]
 
 { #category : #visiting }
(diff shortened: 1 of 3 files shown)


src/Shout/SHRBTextStyler.class.st
```diff
@@ -1101,10 +1101,10 @@ SHRBTextStyler >> visitSequenceNode: aSequenceNode [
...
 { #category : #visiting }
 SHRBTextStyler >> visitVariableNode: aVariableNode [
-	self
-		addStyle: (self resolveStyleFor: aVariableNode)
-		attributes: (self resolveVariableAttributesFor: aVariableNode)
-		forNode: aVariableNode
+
+	self addStyle: (self resolveStyleFor: aVariableNode) attributes: (self resolveVariableAttributesFor: aVariableNode) forNode: aVariableNode.
+
+	self visitMethodComments: aVariableNode comments
```
privat: Visiting comments is not automatic? Because comments can be attached to any nodes.

jecisc: From what I remember, the AST visitor does not visit comments yet.

Comments are not yet well integrated in the AST. There have been improvements during the past few years but there are still work to do.
Visiting comments is something that we need to do. But as always, improvements require time. Cf. https://github.com/pharo-project/pharo/issues/12942

jecisc:
I added the Need More Work tag even if this already improve things and should not break much test because I want to add a test with Jean's example: https://github.com/pharo-project/pharo/issues/13316#issuecomment-1498919836

src/Shout/SHRBTextStyler.class.st
privat: Hum. Since adding explicit comment visit on each node is insane and having a better visitor takes work and time. A global workaround can be to perform a second quick pass only to highlight comments (that override the previous style).

jecisc:
I added some more tests and fixed a bug related to comments in RBLiteralValueNode too

MarcusDenker (approved):

PR #13321, opened 2023-04-06
Playground not able to evaluate simple #timesRepeat:

### LINKED ISSUE #13309: Playground not able to evaluate simple #timesRepeat:
**Bug description**
executing of `5 timesRepeat: [ Smalltalk garbageCollect  ]` in the Playground is causing `Instance of WorkspaceVariable did not understand #index:`

Note: everywhere else (Transcript, Calypso) this evaluation works.

**Version information:**
- Pharo 11.0.0
- Build information: Pharo-11.0.0+build.668.sha.660303e6c0a6b7d9709b670d89051d035d80c0b0 (64 Bit)

confirmed on Pharo 12 too

astares:
Fix #13309

### DIFF
--- a/src/OpalCompiler-Core/OCASTTranslator.class.st
+++ b/src/OpalCompiler-Core/OCASTTranslator.class.st
@@ -442,7 +442,7 @@ OCASTTranslator >> nextUniqueInlineID [
 	parentTranslator ifNotNil: [ ^ parentTranslator nextUniqueInlineID ].
 
 	nextUniqueInlineID := nextUniqueInlineID + 1.
-	^ nextUniqueInlineID asString asSymbol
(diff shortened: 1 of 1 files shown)


MarcusDenker:
"IceGitHubError: API rate limit exceeded for 128.93.193.1. (But here's the good news: Authenticated requests get a higher rate limit. Check out the documentation for more details.)"

Because we fixed that only for Pharo12 builds :-)

jecisc:
Since we change the github plugin credentials, I prefer not to backport that to Pharo 11 so close to the release because I want to be sure it will not cause any problem on the credentials.

jecisc:
Test failure seems unrelated

PR #13323, opened 2023-04-06
Try to move workspace variables further

privat:
( ... until they fell off the table?)

Workspace variables are a very problematic feature with a lot of defaults (UI, spec, implementation).

The compiler have to deal with heuristics and uncooperative requestors (GUI windows) to make them work, including:

* Having (and keeping) references on GUI windows when managing scope
* Heuristic trying to guess if we are compiling or not (just highlighting)
* Fear of asking if a variable exists, because asking so might create it

So this PR tries to clean up the current design (*implementation* design, not *GUI* design) a little.

I also tried to no impact the current UI and spec of (workspace variables are still as bad/good as before). Because it is out of scope of the PR, but also because changing what workspace variables are implies hacking GUI tools that lives in another repositories (and I'm not a GUI guy and PR synchronization between repository are a pain).

So, what is on the menu:

* Reactivate the "undeclared" menu on scripts/playground because it is useful in the playground for classes, for instance.
  It sill buggy, but Pharo11 (and some previous versions I assume) has the same bugs (or worse bugs)
* Do not add requestor in scopes *except* for workspace-type requestors, that still need it (grandfather clause) until I (or someone else) hack Spec/NewTools
* Handle workspace variables with the same idea that was used for undeclared  (see #VariableNotDeclared): get a local placeholder variable at parse/semantic-time and only register it (here to the requestor/GUI) at compile-time.
  This has a nice side effect: undeclared variable are now "blue" in the workspace (I can change that if you do not like it).
* remove most of the insane methods `lookupVar:declare:` and `lookupVarForDeclaration:` that are now unneeded (I suppose)

### DIFF
--- a/src/OpalCompiler-Core/CompilationContext.class.st
+++ b/src/OpalCompiler-Core/CompilationContext.class.st
@@ -699,7 +699,7 @@ CompilationContext >> requestor: anObject [
 	requestor := anObject
 ]
 
-{ #category : #'as yet unclassified' }
+{ #category : #utilities }
 CompilationContext >> requestorCanDeclare: aRequestor [
 	"Heuristic to know if the possible (legagy) requestor can declare new bindings.
 	Because there is no sane API on requestors, we just typecheck until all requestors are dead and buried"
--- a/src/OpalCompiler-Core/CompilationContext.class.st
+++ b/src/OpalCompiler-Core/CompilationContext.class.st
@@ -701,6 +701,22 @@ CompilationContext >> requestor: anObject [
 	requestor := anObject
 ]
 
+{ #category : #'as yet unclassified' }
+CompilationContext >> requestorCanDeclare: aRequestor [
+	"Heuristic to know if the possible (legagy) requestor can declare new bindings.
+	Because there is no sane API on requestors, we just typecheck until all requestors are dead and buried"
+
+	aRequestor ifNil: [ ^ false ].
+	(aRequestor isKindOf: SpCodeScriptingInteractionModel) ifTrue: [
+		^ true ].
+	(aRequestor respondsTo: #interactionModel) ifTrue: [
+		| req2 |
+		req2 := aRequestor interactionModel.
+		aRequestor = req2 ifTrue: [ ^ false ].
+		^ self requestorCanDeclare: req2 ].
+	^ false
+]
+
 { #category : #accessing }
 CompilationContext >> requestorScopeClass [
 	^ requestorScopeClass ifNil: [ requestorScopeClass := OCRequestorScope ]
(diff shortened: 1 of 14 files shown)


src/OpalCompiler-Core/CompilationContext.class.st
```diff
@@ -699,6 +699,24 @@ CompilationContext >> requestor: anObject [
 	requestor := anObject
 ]
 
+{ #category : #'as yet unclassified' }
```
jecisc: The ci will probably not like the unclassified method

jecisc (changes requested):
I like it if we simplify the management of this variable. 

I did a fast review of the code, seems good to me but it would be nice to have a new review of someone with more knowledge of Opal.

jecisc:
There are a bunch of failures:

unix-64 / Tests-unix-64 / testEvaluateOnCompileErrorOnError – Unix64.Spec2.Code.Tests.SpCodePresenterTest9s
unix-64 / Tests-unix-64 / testExternalCompilerDependencies – Unix64.System.DependenciesTests.SystemDependenciesTest<1s
unix-64 / Tests-unix-64 / testExternalDisplayDependencies – Unix64.System.DependenciesTests.SystemDependenciesTest<1s
unix-64 / Tests-unix-64 / testExternalFileSystemDependencies – Unix64.System.DependenciesTests.SystemDependenciesTest<1s
unix-64 / Tests-unix-64 / testExternalLocalMonticelloDependencies – Unix64.System.DependenciesTests.SystemDependenciesTest<1s
unix-64 / Tests-unix-64 / testExternalMetacelloDependencies – Unix64.System.DependenciesTests.SystemDependenciesTest<1s
unix-64 / Tests-unix-64 / testExternalMonticelloDependencies – Unix64.System.DependenciesTests.SystemDependenciesTest<1s
unix-64 / Tests-unix-64 / testExternalMorphicCoreDependencies – Unix64.System.DependenciesTests.SystemDependenciesTest<1s
unix-64 / Tests-unix-64 / testExternalMorphicDependencies – Unix64.System.DependenciesTests.SystemDependenciesTest<1s
unix-64 / Tests-unix-64 / testExternalSUnitDependencies – Unix64.System.DependenciesTests.SystemDependenciesTest<1s
unix-64 / Tests-unix-64 / testExternalSUnitKernelDependencies – Unix64.System.DependenciesTests.SystemDependenciesTest<1s
unix-64 / Tests-unix-64 / testExternalUFFIDependencies – Unix64.System.DependenciesTests.SystemDependenciesTest

src/OpalCompiler-Core/CompilationContext.class.st
privat: indeed

privat:
There is a lot of bad hack to protect from Spec/Newtools. I do not like that.
Maybe just wait that a Pharo12 branch open on these repositories and fix the API there.

privat:
Doing a reroll: #13393

PR #13325, opened 2023-04-06
Remove deprecated modelwants steps in in p12

### LINKED ISSUE #13324: Remove deprecated Model>>#wantsStepsIn: in P12
The method Model>>#wantsStepsIn: was deprecated in Pharo 11 (but not moved to Deprecated11 package)
we can/should remove it now in Pharo 12 (it has no senders)

astares:
Fix #13324


jecisc (approved):
If the test are fine, this is good for integration

PR #13327, opened 2023-04-06
Cleanup small cleanups in baseline of base libraries package

### LINKED ISSUE #13326: Cleanup: small cleanups in BaselineOfBaseLibraries package
- class comment
- tagging of classes
- fix lint that class is not referenced

astares:
Fix #13326


src/BaselineOfBaseLibraries/ManifestBaselineOfBaseLibraries.class.st
```diff
@@ -0,0 +1,426 @@
...
+
+
+
+
+
+
+
+
+
+)
```
jecisc: I think there are a few line returns too much here :)

jecisc:
When an issue is not big and you fix it directly, do not hesitate to do the PR without opening an issue. The message in the PR can replace the need of an issue and it reduces the number of mails of the people subscribed to the project.

astares:
Hi @jecisc 
I use this function in Iceberg - so I would need an issue number first to create a branch (and that branch number/name matches the issue):

image

Is there any other function one could use?

jecisc:
I see,
I find it faster to just create a new local branch simply but I can understand your workflow :)

src/BaselineOfBaseLibraries/ManifestBaselineOfBaseLibraries.class.st
astares: Resolved

jecisc (approved):

astares:
For sure there are many roads to Rome.
But I always though this is the official workflow - because it is built into Iceberg...no?

jecisc:
At some point, the contributions to Pharo after we moved to github needed to follow a strict workflow because we had a lot of constraints and the tools were still new. But since, Iceberg and Pharo were improved and the contribution process is simpler and it allows other workflows as any other git projects. (Sometime I just do a change by editing a file in github interface and asking github.com to do the PR for me! Useful when you just want to fix a typo without having to do anything locally)

Now we can accept pr with branches of any name. So we can manually create our own branch without having to use the github plugin. We can still use it, but it's not required. Just a possibility.

PR #13329, opened 2023-04-06
Cleanup: Better ratio of commented classes

### LINKED ISSUE #13328: Cleanup: Better ratio of commented classes
Still many classes in the system had been left uncommented (without a class/traits comment).

```Smalltalk
stereotypes := OrderedCollection new.
stereotypes
	addAll: Smalltalk allTraits;
	addAll: Smalltalk allClasses.
	
stereotypes removeAll: TestCase allSubclasses.
stereotypes removeAll: BaselineOf allSubclasses.
violating := stereotypes select: [:each | each comment isEmpty ]
```

Using that script we could add a release test that checks that the size is not growing and over time fix more and more classes by commenting them

astares:
Fix #13328

### DIFF
--- a/src/ReleaseTests/ReleaseTest.class.st
+++ b/src/ReleaseTests/ReleaseTest.class.st
@@ -411,6 +411,21 @@ ReleaseTest >> testRPackageOrganizer [
 ]
 
 { #category : #tests }
+ReleaseTest >> testRatioOfCommentedClasses [
+
+	| stereotypes violating |
+	stereotypes := OrderedCollection new.
+	stereotypes
+		addAll: Smalltalk allTraits;
+		addAll: Smalltalk allClasses.
+
+	stereotypes removeAll: TestCase allSubclasses.
+	stereotypes removeAll: BaselineOf allSubclasses.
+	violating := stereotypes select: [ :each | each comment isEmpty ].
+	self assert: violating size <= 914
+]
+
+{ #category : #tests }
 ReleaseTest >> testShouldFileLocatorBeBeforeSystemSettingsPersistenceInStartupList [
 	"This test documents issue https://pharo.manuscript.com/f/cases/17721"
 


jecisc (approved):
I like this idea :)

jecisc:
Test failure unrelated

src/ReleaseTests/ReleaseTest.class.st
```diff
@@ -410,6 +410,21 @@ ReleaseTest >> testRPackageOrganizer [
...
+	| stereotypes violating |
+	stereotypes := OrderedCollection new.
+	stereotypes
+		addAll: Smalltalk allTraits;
+		addAll: Smalltalk allClasses.
+
+	stereotypes removeAll: TestCase allSubclasses.
+	stereotypes removeAll: BaselineOf allSubclasses.
+	violating := stereotypes select: [ :each | each comment isEmpty ].
+	self assert: violating size <= 914
```
privat, line 425: Maybe split the method into two, one that computes the violating classes and one that asserts its size

privat (approved):
good idea.

src/ReleaseTests/ReleaseTest.class.st
astares: Why? As long as there is no other method requiring this it is not needed and we can all the time add if needed.

jecisc:
To avoid this scenario:
- You comment a class of the system, uncommented classes goes down to 913 and test still pass
- You do a commit with a new uncommented class and the number goes back to 914 and the test still pass but we have a new uncommented class 

We can either:
- Check that the uncommented classes are all part of the list of actual violations and then new uncommented classes would fail the test
- Or check that the uncommented classes are exactly a list, and then we would also have to remove the classes we comment form the list and we ensure we do not remove the comment later. This is better but might be a hassle.

Personally I like solution 1 but I would keep the code to generate the tests in a method comment so that we can regenerate the test when we comment a bunch of classes and we think about it.

astares:
Lets keep it simple - we had/have similar release tests who just checked for size of violations. We then applied cleaning and step by step brough the counters down adopting the numbers.

The thing is that we keep improving and the test just assures that the quality level does not decrease....

PR #13331, opened 2023-04-06
Use Protocol>>#unclassified instead of #defaultName

jecisc:
We have two getter to have the unclassified protocol name. #unclassified and #defaultName. This make the readability of the code harder and is useless since it is pretty obvious everywhere else in the code that by default we have an unclassified method.

I removed the need of #defaultName to use #unclassified everywhere.


jecisc:
Tests failure are unrelated

MarcusDenker (approved):

PR #13334, opened 2023-04-06
Simplify protocol removals

jecisc:
This commit simplifies the removal of protocols. Here is the list of changes:

- #removeProtocol: was removing only empty protocols while we have #removeProtocolIfEmpty:. Thus I deprecated it because I prefer to have a selector explicitly saying it will work only if the protocol is empty.
- #removeProtocolIfEmpty: now accept a real protocol as parameter. This will help with the reified Protocol integration 
- #removeProtocolNamed: became obsolete with the previous improvement and is now deprecated
- #removeEmptyProtocols has been simplified to just iterate over the protocols and call #removeProtocolIfEmpty:. Also there was a bug where protocol removals were announced twice and this does not happen anymore.

The API of ClassOrganization lost 2 new methods and this will help with its future inlining :)


src/Kernel/ClassOrganization.class.st
```diff
@@ -91,7 +91,7 @@ ClassOrganization >> classify: selector under: protocol [
 	oldProtocols := self protocolsOfSelector: selector.
 
 	self classify: selector inProtocolNamed: protocolName.
-	(oldProtocols select: #canBeRemoved) do: [ :e | self removeProtocol: e ].
+	(oldProtocols select: #canBeRemoved) do: [ :e | self removeProtocolIfEmpty: e ].
```
privat: `#canBeRemoved` → `#isEmpty` ? and if `removeProtocolIfEmpty` is worth its name, maybe the select is superfluous.

src/Kernel/ClassOrganization.class.st
```diff
@@ -305,7 +305,7 @@ ClassOrganization >> protocolsOfSelector: aSelector [
 	^ (self protocols select: [:each | each includesSelector: aSelector ]) asArray
 ]
 
-{ #category : #'backward compatibility' }
+{ #category : #removing }
```
privat, line 310: why protocol change?

privat (approved):

src/Kernel/ClassOrganization.class.st
jecisc: Done thank you for the catch

src/Kernel/ClassOrganization.class.st
jecisc: I moved it because it is used by a lot of things so I don't think it is just for backward compatibility

MarcusDenker (approved):

PR #13339, opened 2023-04-06
Remove general help package from the repo

jecisc:
Now that we do not load it anymore since https://github.com/pharo-project/pharo/commit/12e9797a33d20e74a820f06ce9a88fc53076f614 , we can remove the General help package


Ducasse:
Yes may be we should do that for P11 too.

jecisc:
Since those packages are not loaded it's not a problem for p11.

This PR was just to clean a little the repo so that we don't have to think about those old packages anymore

PR #13341, opened 2023-04-06
Striked through symbols who are names of a deprecated global

### LINKED ISSUE #13322: Deprecated classes only striked through on class side
- Pharo 12 or earlier
- implement a class side method #isDeprecated returning true for a class 

```Smalltalk
isDeprecated 

	^true
```

Now on the class side the class is striked through in Calypso to show that it should not be used:

image

but on instance side it is not striked through:

image

It would make sense to strike it through also on the instance side in class definition (Fluid or not)

jecisc:
And also tag Deprecated12 as deprecated

Fixes #13322


jecisc:
For this one I think review is important. Not really for the code, but for the idea in itself. 

Here is the result:

image

src/Shout/SHRBTextStyler.class.st
```diff
@@ -991,22 +991,24 @@ at once, but for ordinary literal arrays, style every node"
...
-	self addStyle: (self literalStyleSymbol: value) attributes: { link } forNode: aLiteralValueNode
+	value isSymbol
+		ifTrue: [
+			self class environment
+				at: value
+				ifPresent: [ :aGlobal |
+					attributes add: (TextClassLink className: value).
+					(aGlobal isClass and: [ aGlobal isDeprecated ]) ifTrue: [ attributes add: TextEmphasis struckOut ] ]
+				ifAbsent: [ attributes add: (TextMethodLink selector: value) ] ]
+		ifFalse: [ TextClassLink class: value class ].
```
privat, line 1016: code is complex. Is there a way to move some parts in a new method?

privat (approved):

src/Shout/SHRBTextStyler.class.st
jecisc: One of my other PR will simplify a little the conditions :)

MarcusDenker (approved):

MarcusDenker:
I think it is good to enable it.

If I remember correctly, we disabled it because due to some bad timing, we had lots if classes used that were deprecated in a release.

PR #13346, opened 2023-04-07
Remove ClassOrganization>>hasOrganizedClass

jecisc:
The announcement system of a class organization happens only if it has a class associated. But we should *always* have an associated class.  Since the long terme goal is to clean ClassOraganization and move the remaining behavior to ClassDescription, I propose to remove all those checks directly because once it will be in ClassDescription it will be absolutly sure that we have the organized class.

I also inlined 2 methods that became simpler and are called only in one place. This is to reduce ClassOrganization API size since we want to inline it in the future without polluting too much ClassDescription API.

Let's see if this will break something

### DIFF
--- a/src/Deprecated12/ClassOrganization.extension.st
+++ b/src/Deprecated12/ClassOrganization.extension.st
@@ -45,28 +45,12 @@ ClassOrganization >> listAtCategoryNamed: aName [
 ]
 
 { #category : #'*Deprecated12' }
-ClassOrganization >> notifyOfAddedCategory: protocolName [
-
-	self deprecated: 'Use #notifyOfAddedProtocol: instead.' transformWith: '`@rcv notifyOfAddedCategory: `@arg' -> '`@rcv notifyOfAddedProtocol: `@arg'.
-	self notifyOfAddedProtocol: protocolName
-]
-
-{ #category : #'*Deprecated12' }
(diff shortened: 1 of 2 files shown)


src/Deprecated12/ClassOrganization.extension.st
```diff
@@ -44,13 +44,6 @@ ClassOrganization >> listAtCategoryNamed: aName [
 	^ self methodSelectorsInProtocol: aName
 ]
 
-{ #category : #'*Deprecated12' }
-ClassOrganization >> notifyOfAddedCategory: protocolName [
```
privat, line 48: You removed a `Deprecated12` method? Are we already at Pharo13 and I slept too much?

privat (approved):

src/Deprecated12/ClassOrganization.extension.st
jecisc: I think this methods are really internal to the usage of ClassOrganization, and since we plan to kill it during Pharo 12, I don't think it's worth keeping the deprecation. 

But I can also keep the deprecated method without the transformation and explain why it was removed if people prefer that. 

But if someone is calling this method in their code, they are playing a really dangerous game now that I know better this part of the system...

privat: I do not think there is value to keep internal API method, especially when the whole MM is changed.

MarcusDenker (approved):

PR #13347, opened 2023-04-07
Rename ReleaseTests into Release-Tests

jecisc:
This change was already done in https://github.com/pharo-project/pharo/commit/1bd5aa57559e671b187f69e019ce300d381ab88b but it was not done in a correct way. The old package was not removed and the new package was not loaded in the image.

This:
- Updates Release-Tests to the latest version
- Load it in the baseline
- Remove old one


jecisc:
Lot of tests are failing... I wonder if it is caused by the fact that release tests are now executed later and maybe other tests are dirtying the system?

jecisc:
In order to check if what I said is true, I created: https://github.com/pharo-project/pharo/pull/13349

jecisc:
I'll close this one until #13349 passes

PR #13348, opened 2023-04-07
Add support to deprecate globals

jecisc:
This PR adds support to deprecated global variables.

Here are the implementation details:
- GlobalVariable now has a #isDeprecated flag 
- GlobalVariable now has a #isDeprecated method returning true if the global is deprecated or if its value is a deprecated class
- Shout now style as deprecated the deprecated globals

I also added tests to check the styler does its job on deprecated classes and on deprecated globals.

I tried to make this change the simplest and less invasive possible to not complicate too much the system.

Next step would be to warn the user if we access to a deprecated global but the same still needs to be done for deprecated classes so I did not include that in this PR. (I'm already happy that I was able to do the deprecated global without too much troubles :D )


privat (approved):
very nice!

jecisc:
Sadly it breaks the CI: 

```
PrimitiveFailed: primitive #primitiveChangeClassTo: in Association failed

Association(ProtoObject)>>primitiveFailed:

Association(ProtoObject)>>primitiveFailed

Association(Object)>>primitiveChangeClassTo:

SystemDictionary>>declare:from:

SystemDictionary>>at:put:

ShSmalltalkGlobalsEnvironment(ShGlobalsEnvironment)>>installClass:withName:

ShiftClassInstaller>>installInEnvironment:

[] in ShiftClassInstaller>>make in Block: []

ShiftClassInstaller>>make
```

Apparently something is doing assomptions on the structure of GlobalVariables. Maybe I'll have to create a DeprecatedGlobalVariable subclass instead. Idk, I'll have to check.

privat:
`primitiveChangeClassTo:` says that the format must match, so if you add an ivar to Global, then I assume that an Association instance cannot be upgraded to a Global instance

jecisc:
I'll have to discuss with Marcus to know what is the best way to solve this.

MarcusDenker:
One problem: the bootrap still creates Associations (there is an issue tracker entry for that). 

one idea: this ivar will be nearly always be nil, so maybe using a property for it instead could be an idea, then we just use memory when the class is deprecated

MarcusDenker:
Idea for later: with Globals knowing that they are deprecated, we could compile accesses to them to do something different, e.g. Transcript output. Or we could implement re-writing ag runtime, if we add the new name e.g. as a pragma to the #isDeprecated method.

jecisc:
Thank you for the idea Marcus. When I’ll have my computer and time I’ll update this PR to use a property as you said.

Do you know in which part of the bootstrap the associations are generated? If it’s in espell we will need to wait for it to be fixed to update it. If it’s somewhere else, this should not be that hard to update.

jecisc:
Now I am getting a really weird error. I got the same with another PR but I never found the root of the problem

MarcusDenker:
This error is very strange. I can not see how the changes of this PR lead to the class builder having to do anything different

MarcusDenker:
Maybe the class builder creates a new class for just changing the comment?

jecisc:
D74861EF-421A-4A5A-9012-8F35CEE94457

before the class builder error there is this error that I also have on a totally unrelated PR

MarcusDenker (approved):

PR #13349, opened 2023-04-07
Execute Release-Tests last on CI

jecisc:
We have a lot of tests in Pharo and it might happen that some of them do not respect the principle of letting the system clean with their tearDown. To help to spot those impacts, I propose to run the releases tests last. Like this, if some bad generated code/or modifications of system entities happens, we might catch it with the verifications of release tests.

I expect this commit to break the tests in the CI but it'll help me find the culprits and fix them.


privat:
Nice.

Since tests force the coverage (of) some code, thus performs code transformation because of deprecation, is there a release test that check that the code is pristine/unmodified?

jecisc:
There is none yet, but I would love to have that :)

jecisc:
But we will need some work to get it because in Pharo 12 it comes with dirty packages, so this test would fail currently

privat (approved):

jecisc:
More than 10 tests failures because the tests are dirtying the image D:

privat:
`primitiveChangeClassTo:` says that the format must match, so if you add an ivar to Global, then I assume that an Association instance cannot be upgraded to a Global instance

jecisc:
Here is a first fix: https://github.com/pharo-spec/NewTools/pull/504

jecisc:
I have no idea why the CI has troubles with this PR. I'm redoing it https://github.com/pharo-project/pharo/pull/17783

PR #13350, opened 2023-04-07
AST: fix node positions, and test them systematically

### LINKED ISSUE #13318: RBVariableNode with comment goes 1 character too far

privat:
fix #13318

In the AST, value nodes (i.e. expressions) can be parenthesized.
All parentheses positions are saved in the AST (for styling, for instance).
There are therefore two starts: the position of the first parenthesis (called `start`), and the position of the interesting part of the expression, called `startWithoutParentheses`.
Same thing with the stop position.

`RBVariableNode` and `RBParseErrorNode` did not always return the correct value for start or stop, leading to weird styling, and possible unexpected behavior on editor for right-click menu.

Fixes are the 2 first commits.

The other commits are about testing.
`RBCodeSnippet` is extended with a way to identify an AST node at a given position is the source.
The method `nodeForOffset:` is used for that.
It is the simplest and more reliable one as it only considers `start to: stop` and do not have hacks or heuristics.
Other methods for the *same* jobs exists, `bestNodeFor:` and `bestNodeForPosition:` that are filled with hacks and heuristics. Another one also exists in rubric `RubSmalltalkEditor>>#bestNodeInString:at:editingMode:shouldFavourExpressions:onError:` and I assume gives also different results.

The diff is big because all snippets are updated. I checked each one and discovered some oddities that could deserve a fix (or not) in future PR:

* comments are not considered, because they are not real `children` nor visited with `nodesDo`.
* the `.` of statements do not belong to the statement.
* but the `;` of cascade belong to the message that follows it.
* `:` of block parameters do not belong to the variable.

Some additional thought: maybe the same approach could be used to test styling, using a string where each character correspond to a given style/color.

### DIFF
--- a/src/AST-Core-Tests/RBCodeSnippet.class.st
+++ b/src/AST-Core-Tests/RBCodeSnippet.class.st
@@ -1728,15 +1728,15 @@ RBCodeSnippet class >> styleWithError: snippets [
 ]
 
 { #category : #updating }
-RBCodeSnippet class >> updateAllSnippets [
+RBCodeSnippet class >> updateSnippets: snippets [
 	"This script will update all definitions of all snippets.
 	Beware of code loss and double check the generated definitions."
 
 	"I did not manage to update all in a single transformation (with a preview!)
 	So do them one by one with a progress bar :("
 
-	<script>
-	self allSnippets
+	<script: 'self updateSnippets: self allSnippets'>
+	snippets
 		do: [ :snippet |
 			snippet definitionRefactoring ifNotNil: [ :refactoring |
 				refactoring
--- a/src/AST-Core-Tests/RBCodeSnippet.class.st
+++ b/src/AST-Core-Tests/RBCodeSnippet.class.st
@@ -52,6 +52,7 @@ RBCodeSnippet class >> allSnippets [
 		  self badTokens.
 		  self badMethods.
 		  self badSemantic.
+		  self badPositions.
 		  self badVariableAndScopes } flattened
 ]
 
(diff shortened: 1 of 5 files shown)


jecisc (approved):
Reading the code it seems good

privat:
tests are ok. One `TestTookTooMuchTime` on unix, but I assume it's unrelated

Ducasse:
Did I get it correctly that the message stop does not exist anymore?

privat:
> Did I get it correctly that the message stop does not exist anymore?

I'm not sure that I understand what you mean by "*message stop*"

jecisc:
I think this is the method #stop (#start)

privat:
#start and #stop exist for every AST node, it is used in most tools that present source code to the user, including code editor, styler, or the debugger to highlight the current node to execute.

Ducasse:
Yes I know now I saw that the method stop was removed so this is why I asked.

privat:
> Yes I know now I saw that the method stop was removed so this is why I asked.

Methods fly away, but selectors remain. :)
`RBVariableNode >> start`, `RBParseErrorNode >> start` and `RBParseErrorNode >> stop` were removed because the inherited ones from `RBVAlueNode` are the right ones.

Ducasse:
Oki :)

MarcusDenker (approved):

Issue #13351, opened 2023-04-07
Release tests should check that the code of the image is unchanged


jecisc:
I created a PR to launch the releases tests last. When this will be integrated we might want to add a release test to check if the code of the image is still clean because multiple things could update it such as:
- Tests manipulating some classes or other entities
- Automatic transformations of deperecations
- Packages with classes/methods with the same name as another package
- Conflicting extension methods
- ...

PR #13352, opened 2023-04-07
Compiler: remove OCASTTranslator subclasses

### LINKED ISSUE #13330: Compiler: Remove OCASTTranslator subclasses for effect/value
OCASTTranslator uses two sub-instances, depending if it visits for effect or for value. 
This in the end just is the difference of adding a pop or not.

We should check if this instead could be be implemented using an instance variable that models this state.

- Much simpler, less magic
- it simplifies subclassing --> the existing subclasses for Reflectivity can be simplified

privat:
Fix #13330

See issue for explanation of the problem.

The idea of my solution is to always visit in *value* mode, that means that values are always pushed on the stack with the *normal* visitor API.
But offers two helper method `visitEffectNode:` and a special `visitEffectInlinedBlockNode:` that just do the visit then pop the value.
This way, (internal) client can use either `self visitNode:` if they want a value (instead of the insane `valueTranslator visitNode:`); or `self visitEffectNode:` if they just want the effect (instead of the insane `effectTranslator visitNode:`).
These two internal ivar are thus unused and their class removed.

What about perf? Pushing a value then poping it is wasted CPU time!
First, we do not care, IR or bytecode is not necessarily the correct place to do optimizations. It's the VM job. And more the bytecode is naive, more the performance increase numbers sold by VM people are impressive!!!
Second, we care a little about space, so I added a simple optimization where a `pop` can cancel a previous `push` (both are annihilated) when it's safe to do so. But it's very simple and tries to be equivalent to what the duettist visitors valueTranslator and effectTranslator were doing.

Note: there are a lot of commits because it's open-heart surgery. Hacking the compiler live need steady hands. So you can see the incremental changes if you wish. I almost not used the overlay compiler because I had weird issues with it


privat:
Bootstrap is passing and most tests :tada:
But I did break metalink and related (coverage). And I'm not sure what I did wrong.

Some tests failed because the bytecode is not exactly the same as they expected (tests can be picky sometime)

privat:
I repaired most of metalink, and I updated some tests to reflect some bytecode change.

privat:
Found the issue in Reflectivity that caused random crashes, it was related to a fragile design in Reflectivity, and the fact that the VM do not care if a CompiledMethod has a return or not, so will execute garbage bytes.

privat:
All tests are green!

Some special remarks:

* Reflectivity/Metalinks are very fragile. The code is complex, but having now a single visitor might help a lot to improve things. I added some `skip` (you can find them in some commits) that I removed later once fixed. But in my precipitation, I removed some `skip` that I did not add. So I assume I did manage to fix things as a side effect :p
* Relying on precise bytecode is **bad**. Tests might be OK, because if you change the bytecode then you can change the test, but tools like the debugger/decompiler/context/etc. should be agnostic to the bytecode.
* The debugger having to rely on the disassembler to map pc to source code is **very bad**. There are too many fragile assumptions when doing so. And the disassembler is not that reliable.

Anyway, that was a fun 2 days!

src/Debugger-Model-Tests/StepOverTest.class.st
```diff
@@ -60,7 +60,8 @@ StepOverTest >> stepB3 [
 
 { #category : #tests }
 StepOverTest >> stepC1 [
-	self stepC2
+	self stepC2.
+	^ self
```
jecisc: I guess this is needed by tests? I though that if there is no return, this was automatically added to the bytecode (or to something else, I don't really know this part of the system **yet**)

src/Reflectivity-Tests/VariableBreakpointMockClass.class.st
```diff
@@ -36,8 +36,7 @@ VariableBreakpointMockClass >> methodWithTempsAndArg: arg [
 ]
 
 { #category : #accessing }
-VariableBreakpointMockClass >> v1 [
-	^ v1
+VariableBreakpointMockClass >> v1 [ ^v1
```
jecisc, line 39: Is this formatting normal?

jecisc (approved):
That seems good to me.
As always it's good to have a second opinion since I do not know this part of the system. But the code seems much more readable with those changes.

I check for the skip and they already pass in an image without this change. Maybe we should do a little script to run tests with skips and report if some of them pass and run it from time to time. But that has nothing to do with this PR :)

src/Debugger-Model-Tests/StepOverTest.class.st
privat: The test that used the example assumed circumstantial things that were not preserved by my changes.
 So I updated the example, it was easier that updating the test.

src/Reflectivity-Tests/VariableBreakpointMockClass.class.st
privat: Some test is rewriting the method (without formatting?). So I committed the result because I was tired of unselecting it at each commit. Cf #13351

src/Debugger-Model-Tests/StepOverTest.class.st
jecisc: If this return is needed I would add a comment I think.
I know that if I found this code I would probably remove the return if there is no comment >:)

Ducasse:
This is super nice to see all the work around the compiler :). I will let marcus review the code before pressing on the button
but this is really a great feeling.

privat:
> This is super nice to see all the work around the compiler :)

Thank you, I did a bunch of PR, but this one was very hard. I'm quite proud :)

MarcusDenker (approved):

PR #13355, opened 2023-04-08
extract return from Number>>#raisedTo: and add some tests

### LINKED ISSUE #13354: extract return from Number raisedTo
extract return from Number>>#raisedTo:

Pharo image: Pharo12.0.0#120137

Platform Name: unix

vonbecmann:
fix: #13354
hope the tests work in every architecture(32 or 64)

### DIFF
--- a/src/Kernel-Tests/NumberTest.class.st
+++ b/src/Kernel-Tests/NumberTest.class.st
@@ -126,6 +126,18 @@ NumberTest >> testRaisedTo [
 ]
 
 { #category : #tests }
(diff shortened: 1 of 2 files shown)


MarcusDenker (approved):

jecisc:
Thank you!

Don’t worry too much, extracting returns like this is really safe usually when there is a message with two branches that are always executed and both have a return in the last statement. I cannot think of any case where it could break something ;)

Worst case scenario, the tests should catch the problem.

PR #13356, opened 2023-04-08
Make protocols an array in ClassOrganization

jecisc:
ClassOrganization is storing protocols into an Identity set but almost all operations applied on them  transform the protocols into an array. I think we will save memory and maybe even time by just keeping an array of protocols. Adding/Removing might be a tiny bit slower but we will avoid a lot of casting.

I also renamed some temporaries to make the code more readable.

I hope this will not break anything because I think some parts lack tests in ClassOrganization. I plan to add them, but I want to finish my cleanings before because some of the behaviors with missing tests I currently found are in methods on my killing list or my list of methods to improve.

### DIFF
--- a/src/CodeExport/ClassOrganization.extension.st
+++ b/src/CodeExport/ClassOrganization.extension.st
@@ -32,14 +32,14 @@ ClassOrganization >> putCommentOnFile: aFileStream forClass: aClass [
 ClassOrganization >> stringForFileOut [
 
 	^ String streamContents: [ :aStream |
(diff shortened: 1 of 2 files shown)


jecisc:
Error seems unrelated

privat (approved):
Note that transforming an internal collection as an array when giving them to the client has the advantage of protecting the object consistency if the client tempers with the array.

MarcusDenker (approved):

PR #13358, opened 2023-04-08
Always announce creation of new protocols

jecisc:
In the current Pharo, we do not always announce the creation of a new protocol.

In the recent past we had two classes to classify a protocol:
- ClassOrganization with #classify:under: 
- ProtocolOrganizer with #classify:inProtocolNamed: 

The problem is that ProtocolOrganizer was not announcing any change happening and it was the responsibility of ClassOrganization. This means that any user of #classify:inProtocolNamed: does not announce the creation of a new protocol if the protocol in parameter does not exist yet.

While merging ClassOrganization and ProtocolOrganizer I tried to merge both methods and unify the behavior but this caused the bootstrap to crash.

Today I am trying once again to announce all protocol creations. The origin of the previous crash is an infinite loop that was caused by the management of TraitedMetaclass. It was going like this:
- We want to load a class with traits and class side methods in the trait
- The class is added to the system
- The organization of the TraitedMetaclass is initialized
- Creating the organization adds all methods to #'as yet unclassified'
- Adding the methods to this protocol first create the protocol
- Creating the protocol announce the creation if we do not silently add the protocol
- RPackage listen to this event and ask the protocols of the class to do some checks
- To get the protocols we need the organization of the TraitedMetaclass
- The organization did not finish its initialization, the variable is still need, and here we go again in the defensive strategy.  Go back in step 3 and enjoy your ride of "We have an infinit loop"

With the removal of the silent add of this PR, I directly save the instance of ClassOrganization before initializing the protocols so that RPackage is happy and can ask the organization what are the current protocols. 

Let's see if that is enough to fix this bug of silent protocol add.
If this is not enough, I will have to keep the silent add, but I will use it **only** for the initialization of the class organization and nothing more. (We currently have 4 places in the system that use this silent protocol addition).

Little bonus: 1 less method in ClassOrganization (and no other method that got more complexe). This mean 1 tiny step toward a ClassOrganization we can inline in ClassDescription :)

This PR is really small but it was not easy to find the problem.

### DIFF
--- a/src/Kernel/ClassDescription.class.st
+++ b/src/Kernel/ClassDescription.class.st
@@ -846,11 +846,11 @@ ClassDescription >> organization [
 	"Answer the instance of ClassOrganizer that represents the organization
 	of the messages of the receiver."
 
-	self basicOrganization
-		ifNil: [ self basicOrganization: (self isTrait
-						ifTrue: [ ClassOrganization new ]
-						ifFalse: [ ClassOrganization forClass: self ]) ].
-	^self basicOrganization setSubject: self.	"Making sure that subject is set correctly. It should not be necessary."
+	self basicOrganization ifNil: [
+		| classOrganization |
+		self basicOrganization: (classOrganization := ClassOrganization new).
+		self isTrait ifFalse: [ classOrganization initializeClass: self ] ].
+	^ self basicOrganization setSubject: self "Making sure that subject is set correctly. It should not be necessary."
 ]
 
 { #category : #organization }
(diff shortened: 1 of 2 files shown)


jecisc:
If this pass, we can merge both ways to classify a method and that will allow us to simplify multiple things in the protocol management of ClassOrganization :)

jecisc:
This will clash with  #13356 but if we have to chose which one to integrate first I would like this one because it is required for multiple other cleanings I want to do on protocols :)

privat (approved):

MarcusDenker (approved):

Ducasse:
Ok marcus was faster than me :)

PR #13360, opened 2023-04-09
Clean ClassRenameFixTest

jecisc:
ClassRenameFixTest is way too complicated for what it does. This aim to simplify it and make it cleaner.

Currently running this test was leaving the system in a dirty state leaving an empty package. (cf. https://github.com/pharo-project/pharo/pull/13349)

Now, instead of saving all the changes and trying to revert it, the code generated is all in a specific package that we remove in the tear down. I removed a double initialization.
And finally I inlined #renameClassUsing: because this method was doing almost everything and the test nothing


jecisc:
[CI] Ring tests cause segmentation fault on GC #13364

MarcusDenker (approved):

PR #13362, opened 2023-04-09
use and: expression instead of ifTrue:ifFalse: expression

### LINKED ISSUE #13361: use and expression instead of ifTrue ifFalse expression
search in the Kernel packages the following expression 
```
`@receiver ifTrue:[ `@exp ] ifFalse:[ false ]
```
and replace it with an and: expression.

Pharo image: Pharo12.0.0#120139

Platform Name: unix

vonbecmann:
fix: #13361

### DIFF
--- a/src/Kernel/CompiledMethod.class.st
+++ b/src/Kernel/CompiledMethod.class.st
@@ -503,8 +503,7 @@ CompiledMethod >> hasProperty: aKey [
 
 	| extendedMethodState |
 	^ (extendedMethodState := self penultimateLiteral) isMethodProperties
-		  ifTrue: [  extendedMethodState includesProperty: aKey ]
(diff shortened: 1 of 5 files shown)


jecisc (approved):
This is nice! 

I have a project to automatically do this kind of cleanings and I never thought about this one. I think I’ll add it because I like this and it makes the code clearer.

MarcusDenker (approved):

vonbecmann:
> This is nice!
> 
> I have a project to automatically do this kind of cleanings and I never thought about this one. I think I’ll add it because I like this and it makes the code clearer.

there also this expression
```
`@receiver ifTrue:[ true ] ifFalse:[ `@exp ]
```
but the change is different. maybe a guard clause. don't know

PR #13367, opened 2023-04-09
Refactoring `RBRefactoringWarning` and `RBRefactoringError`

balsa-sarenac:
`RBRefactoringError` was resumable and was used instead of `RBRefactoringWarning`, and `RBRefactoringFailure` was then used instead of `RBRefactoringError`. This PR addresses these issues.

This change was bigger than initially anticipated. Main reason is because Refactorings and Tranformations handle refacotring errors differently.
- refactorings used `RBRefactoringError` for resumable warnings, and `RBRefactoringFailure` for the non-resumable ones
- transformations only used `RBRefactoringError`

Because of this difference tests started failing.

8 tests were impacted by this change, where 5 of them were easily fixed. The other 3 were a bit more problematic because the behaviour of parametrized test was different for refactoring and transformation.
Since all transformations that were impacted are unused. I've decided to remove 2 them for parametrized tests all together (https://github.com/pharo-project/pharo/pull/13367/commits/2a7856822db190f360707fe715ae17e666191101), since fixing the tests would require duplicating testing code. And the third one was single parametrized, and I've just fixed it to make it work (https://github.com/pharo-project/pharo/pull/13367/commits/b5d6695ffbbb07100baea66de8e538e7dfe7956e).

I think that this is a nice step forwards for refactorings, and a nice move towards merging of refactorings and transformations.


Ducasse:
I see that around 8 tests are impacted.

Ducasse:
Let me know when I should remove the draft tag.

balsa-sarenac:
@Ducasse it's ready for a review. Pay attention to updated description of the PR.
Also, I've asked Marcus and he gave me permissions to update my PRs with tags and also manage "Refactoring Engine board" and close old/outdated refactoring issues, so that's cool

PR #13368, opened 2023-04-09
ASTCache add statistics, then use LRUCache

privat:
Solving resource issues need precise numbers, so in order to understand  #8984 I added a cache statistic to the ASTCache, then used a LRUCache instead.

Note that ASTCache is a subclass of WeakIdentityKeyDictionary that associates a compiled method to an AST. Most compiled methods rests in the system for long time, but actively developed part of the system should have a higher turn over. However, using a weak dictionary here seems a bad idea. The weak part permits unused method to be garbage collected (the cache do not hold then strongly), but long living method (most of them!) will stay in the cache forever.

Some stats with WeakIdentityKeyDictionary

* after a full `Smalltalk recompile` then a `ASTCache reset`: `an ASTCache#9 0%`
* after DrTest all tests (1248) of `OpalCompiler-Tests`: `an ASTCache#67 74%`
* (reset astcache)  all critiques on Kernel (that has the most methods, 4153): `an ASTCache#4317 99%` make sense because the same AST is used again and again
* (reset astcache)  a single critique 'Dead Block' on Kernel (that has the most methods, 4153): `an ASTCache#4317 0%` make sense because the same AST is not reused

Some stats with LRUCache

* after a full `Smalltalk recompile` then a `ASTCache reset`: `an ASTCache#3 3/16 [ 1 ] 99%`
* after DrTest all tests (1248) of `OpalCompiler-Tests`: `an ASTCache#16 16/16 [ 1 ] 99%`
* (reset astcache)  all critiques on Kernel (that has the most methods, 4153): `an ASTCache#16 16/16 [ 1 ] 99%`
* (reset astcache)  a single critique 'Dead Block' on Kernel: `an ASTCache#16 16/16 [ 1 ] 98%`

### DIFF
--- a/src/AST-Core/ASTCache.class.st
+++ b/src/AST-Core/ASTCache.class.st
@@ -91,6 +91,17 @@ ASTCache >> at: aCompiledMethod [
 				  self class cacheMissStrategy getASTFor: aCompiledMethod ] ]
 ]
 
+{ #category : #printing }
+ASTCache >> printElementsOn: stream [
+
+	stream
+		nextPut: $#;
+		print: self size;
+		space;
+		print: (self statistics hitRatio * 100.0) rounded;
+		nextPut: $%
+]
+
 { #category : #initialization }
 ASTCache >> reset [
 	self removeAll.
--- a/src/AST-Core/ASTCache.class.st
+++ b/src/AST-Core/ASTCache.class.st
@@ -99,7 +99,7 @@ ASTCache >> at: aCollection ifAbsentPut: aRBMethodNode [
 { #category : #accessing }
 ASTCache >> at: aCollection put: aRBMethodNode [
 
-	self cache at: aCollection put: [ aRBMethodNode ]
+	self cache at: aCollection put: aRBMethodNode
 ]
 
 { #category : #accessing }
--- a/src/AST-Core/ASTCache.class.st
+++ b/src/AST-Core/ASTCache.class.st
@@ -108,7 +108,7 @@ ASTCache >> cache [
 	^ cache ifNil: [ "I don't want to depend on System-Caching. So, no system caching, no cache :P"
 		  Smalltalk globals
 			  at: #LRUCache
-			  ifPresent: [ :class | cache := class new ] ]
+			  ifPresent: [ :class | cache := class new beThreadSafe; keyIndex: IdentityDictionary new ] ]
 ]
 
 { #category : #printing }
--- a/src/AST-Core/ASTCache.class.st
+++ b/src/AST-Core/ASTCache.class.st
@@ -88,6 +88,12 @@ ASTCache >> at: aCompiledMethod [
 ]
 
 { #category : #accessing }
+ASTCache >> at: aCollection ifAbsentPut: aRBMethodNode [
+
+	self cache at: aCollection ifAbsentPut: aRBMethodNode
+]
+
+{ #category : #accessing }
 ASTCache >> cache [
 
 	^ cache ifNil: [ cache := LRUCache new ]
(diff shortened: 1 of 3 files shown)


jecisc:
Once we use a LRUCache I think we can remove at least one reset in the image. IIRC there was one in a test to clear the memory because it was going over all the methods ast of the image

privat:
> Once we use a LRUCache I think we can remove at least one reset in the image

Since the LRUCache is not weak, it holds strongly the compiled methods (and their AST). So maybe we can keep the resets, they are not expensive anyway.

Also, I let the default size of 16, but maybe it's too low. Eg. If editors use the astcache on each tab, then it might be not enough. Stats are available so it can be easy to see in the long run anyway :)

privat:
Full image build

* the PR (89763c9): 8m 39s
* the base (Pharo12 97c3483): 8m 46s
* I assume the improvement is noise. Or maybe in the bootstrap use case, a 16-slots LRU is more efficient than an infinite-slots WeakIdentityKeyDictionary

privat:
Tests failed. There is a lot of `a DoubleLinkedList is empty`, some scary `
[ERROR] sqDeallocateMemorySegmentAtOfSize (src/memoryUnix.c:204):sqDeallocateMemorySegment... munmap: Invalid argument` 

hmmm..

jecisc:
I got `[ERROR] sqDeallocateMemorySegmentAtOfSize (src/memoryUnix.c:204):sqDeallocateMemorySegment... munmap: Invalid argument` on a PR recently. 

The problem was that I got an infinit loop and it ended up crashing Pharo after a while with this error

privat:
I can reproduce at home with the image from the artifacts of the job if I run tests in console, but I cannot reproduce manually when unit testing on some packages with drtest.

I tried to force tread safety (there is a flag for it in LRUCache). But it did not solve the problem.

I do not have a lot of ideas :/

It could be:
* [ ] A bug in my PR, but even if the code is small, my eyes are tired.
* [ ] A bug in LRUCache or DoubleLinkedList. Are these classes used a lot and reliable?
* [ ] Some concurrency issue. I didn't see that the old ASTcache was thread safe, but maybe.
  → Trying in progress with `beThreadSafe`
* [x] WeakIdentityKeyDictionary (ald ASTCache) use object identity. Whereas LRUCache use a standard Dictionary, so AST of equal CompiledMethod might clashes.
  → Trying in progress with a `IdentityDictionary` internally
* [x] Some clients of ASTCache that have some (very) broken expectations of ASTCache not being a cache. Like storing information in AST nodes or using nodes as key for some dictionaries. Thus, behaving very badly when the AST are ejected from the cache and a new AST (with new node and new identities) is given back later.
  → #13371 shows it's the case
* [x] a combination of some previous ones

jecisc:
Maybe it comes from the order of the ran tests to produce the crash?
In that case you might want to try to run the tests with the command line but with the `-interactive` flag to see if that also happens?

privat:
Yes, I did my run in interactive move.
I do not know how jenkins run them.

`RPackageExtensionMethodsSynchronisationTest` are soooo slooow on my machine... why? [update: I know why, because the image is already corrupted]

privat:
I think it was the `IdentityDictionary` thing.

Since the dictionary used internally by LRUCache is used to store internal nodes of the linked list, collision on CompiledMethods caused confusion on the linked node, and corruption of the linked list. But #13371 shows some additional fragility that should be solved first.

privat:
Because of the expectation on ASTCache about persistence, see #13371, a LRUCache just won't work as is. So close the PR :(

PR #13369, opened 2023-04-10
Rename Protocol class>>#ambiguous into #traitConflictName

jecisc:
'ambiguous' is an ambiguous name in itself because it is hard to understand how it is useful with just the name. I propose to rename it into Protocol class>>traitConflictName.

Also, since it is an implementation detail of traits, I propose to move it to the TraitsV2 package.

I finally did some renamings toward the goal of removing the "category" vocabulary.

### DIFF
--- a/src/Kernel/Protocol.class.st
+++ b/src/Kernel/Protocol.class.st
@@ -13,11 +13,6 @@ Class {
 }
 
 { #category : #'instance creation' }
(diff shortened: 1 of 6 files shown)


privat (approved):

MarcusDenker (approved):

Ducasse:
I like the changes!

PR #13371, opened 2023-04-10
CI-Test with ASTCache disabled (do not merge)

privat:
While cache invalidation is one of the two hard things in computer science, knowing that a cache is a cache is an easy one.

This PR is not intended to be merged, but is here to understand possible issues with #13368 (and also #8984)

### DIFF
--- a/src/AST-Core/ASTCache.class.st
+++ b/src/AST-Core/ASTCache.class.st
@@ -95,7 +95,7 @@ ASTCache >> at: aCompiledMethod ifAbsentPut: aBlock [
 	(aCompiledMethod propertyAt: #reflectiveMethod) ifNotNil: [ :rf | ^ rf ast ].
 
 	"Look in the (almost infinite) cache"
-	self weakDictionary at: aCompiledMethod ifPresent: [ :ast | ^ ast ].
+	"self weakDictionary at: aCompiledMethod ifPresent: [ :ast | ^ ast ]."
 
 	"We tried everything we could. So compute and store it"
 	^ self at: aCompiledMethod put: aBlock value
--- a/src/OpalCompiler-Core/OCAbstractMethodScope.class.st
+++ b/src/OpalCompiler-Core/OCAbstractMethodScope.class.st
@@ -228,13 +228,13 @@ OCAbstractMethodScope >> removeTemp: tempVar [
 OCAbstractMethodScope >> setCopyingTempToAllScopesUpToDefTemp: aVar to: aValue from: aContext [
 	"we need to update all the copies if we change the value of a copied temp"
 
-	self = aVar scope
-		ifTrue: [ ^ aVar writeFromLocalContext: aContext put: aValue ].
-	(self lookupVar: aVar name) writeFromLocalContext: aContext put: aValue.
-	^self outerScope
-		setCopyingTempToAllScopesUpToDefTemp: aVar
-		to: aValue
-		from: (self nextOuterScopeContextOf: aContext)
+	(self lookupVar: aVar name)
+		writeFromLocalContext: aContext
+		put: aValue.
+	^ super
+		  setCopyingTempToAllScopesUpToDefTemp: aVar
+		  to: aValue
+		  from: aContext
 ]
 
 { #category : #'temp vars' }
--- a/src/OpalCompiler-Core/OCAbstractMethodScope.class.st
+++ b/src/OpalCompiler-Core/OCAbstractMethodScope.class.st
@@ -140,7 +140,8 @@ OCAbstractMethodScope >> localTemps [
 OCAbstractMethodScope >> lookupDefiningContextForVariable: var startingFrom: aContext [
 	"Is this the definition context for var? If it not, we look in the outer context using the corresponding outer scope. If found, we return the context"
 
-	^self = var scope
+	"Because AST are not reliable, we do the correspondance on var names"
+	^ (self localTempNames includes: var name)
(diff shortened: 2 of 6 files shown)


privat:
CI timed out :(
At least the cache is useful (or a test went into an infinite loop? Is there a time limit per test?)

privat:
Try again with a single-slot LRU cache (ie just remember the last one)

privat:
CI went further a lot. According to tests, it appears that Reflexcitivty/Metalinks assume the AST cache is reliable.
Since metalinks are added to AST nodes, it makes sense that losing AST will cause trouble. AST are stored in ReflectiveMethod, so an AST seems correclty kept around.

But maybe the issue is in `ReflectiveMethod>>#installCompiledMethod` that put the AST to the ASTCache. But doing so should be considered "best effort"/"a hint" only.

Since a CompiledMethod knows its possible ReflectiveMethod (that holds strongly the AST), it could be possible to bypass the cache

privat:
It did not work that much :/

## A second problem, link installation

There is an issue on the API.

If you do something like:

```st
(AClass>>#aSelector) ast link: aLink
```

* the method gives you an AST (ast1).
* you add a link on it: `RBProgramNode>>#link:`.
* that will call `createTwin` and instantiate a new `ReflectiveMethod`.
* that will call `ReflectiveMethod>>#compiledMethod:`.
* that does `ast := compiledMethod ast`, so the AST of the `ReflectiveMethod` might be ast1, or **another** AST if there is a cache miss. In this case your link will be added to a dandling AST. And there might be an inconsistency on the `linkCount` of the `ReflectiveMethod` causing more problems

Using a 1slot LRU, will probably work in this example, but it's circumstantial.

A possible API solution could be to not lose the AST of the node and reuse it for the ReflectiveMethod, so pass around the AST until the ReflectiveMethod instantiation.
Another solution could be to prevent adding links to a random AST nodes found in a box on the street.
For instance, we could check that the AST node belong to a method associated to a reflective method. So the client has to prepare the reflective method first.
Possibly we can simply propose `CompiledMethod>>#reflectiveAST` that ensures that a ReflectiveMethod exists and that the returned AST is strongly held.

## A third problem, link results

The previous problem does not explain the failing tests, as a 1slot-LRU is enough to (circumstantially) work on most test cases, because those tests test on a single compiled method and use a fresh enough AST.

The reason of the failure is that some tests uninstall the link, thus removing the twin, thus removing the strong hold on the AST **before** checking the AST nodes for some result tag.
But if you ask for an AST to a method that used to have metalinks, there is a non-zero chance that you get a new, fresh AST that won't contain what the test expects.

With a 1slot-LRU, again, most of the tests should success. Except that in the current PR that disabled the cache, the `ASTCache default at: compiledMethod put: ast` of `installCompiledMethod` is ignored so, a lot of test fails.
But, #13368 that has a (now) working 16-slots LRU cache pass most tests. But, I insist, circumstantially, as real workloads with a non-infinite cache will likely fail semi-randomly.

Not that the `Coverage` package (that I wrote) suffers from these things because it can install a lot of metalinks on a lot of methods, and it wrongly collects the results after uninstallation. I did not understand at the time (I was young and naive) that `CompiledMethod>>#ast` used some cache and that the returned AST should maybe not be considered persistent :(

A solution at the API level could be to not remove the ReflectiveMethod on the last uninstalled link, so the AST could be held strongly.
But it will require the user to trigger the removal of the ReflectiveMethod in some way.

Maybe the garbage collector and weak references could help here.
If all links and AST nodes are garbage collected, then the ReflectiveMethod can be uninstalled.

BUT. If we involve the GC, maybe both the whole ASTCache and metalink issues can be solved easily: instead of a `WeakIdentityKeyDictionary` we should use a `WeakIdentityKeyAndWeakValueDictionary` where we return the same AST as long as someone has a strong reference on a part of it (because of `#parent` AST are basically strongly connected) so:
* AST cache won't cost that much memory (because the AST is sill hell otherwise).
* Pseudo-persistence of AST will be preserved, and metalinks users can still feel young and naive if they wish.

I don't like having an API that requires some specific magic features of :stars: memory :rainbow: management :magic_wand:. But cache invalidation is notoriously hard, so a little magic could help :p

privat:
I fully removed the cache again, but keep the strong ref on ReflectiveMethod.
Tests timed out again. So I blacklisted Reflextivity-Tests (340 x3 tests skipped).

In the last CI run, ASTCache always give you a new AST (and won't remember it) unless you are a script (doit) or have a reflective method.
There are more failures: 159 x3, but it is far less than I expected. 
What seems to fail are:

* Things based on reflectivity & coverage. Obviously, see the previous wall of text
* Debugger, possibly because the Context can give the AST of the currently execute node, but the debugger might expect some persistence, I don't know.
* things related to temporary variables (mapping?). I do not know what it is. Maybe some tests try to extract TemporaryVariable instances from an AST, but since the AST is not persistent, then TemporaryVariable are not persistent either (unlike other global/shared variables)
* things related to breakpoints. I think it's based on reflectivity in fact.

Small conclusion: the broken clients 1. are not that numerous, 2. are highly specialized. Bootstrap, GUI and every day Pharo tasks seems unaffected when the ASTCache is on strike.
So possibly there is a way to avoid relying on the ASTCache at all (except for perf reasons, so making astcache a real cache), and without too much memory management magic. We can start to look at each one of these clients and understand why they care that much about persistence.

privat:
Close for the moment. Maybe I will revisit the issue

PR #13372, opened 2023-04-10
extract return from Rectangle>>#quickMerge: and use self class 

### LINKED ISSUE #13370: extract return from Rectangle quickMerge
extract return from Rectangle>>#quickMerge: and use self class instead of Rectangle.

Pharo image: Pharo12.0.0#120141

Platform Name: unix

vonbecmann:
fix: #13370

### DIFF
--- a/src/Kernel/Rectangle.class.st
+++ b/src/Kernel/Rectangle.class.st
@@ -723,7 +723,7 @@ Rectangle >> printOn: aStream [
 
 { #category : #'rectangle functions' }
 Rectangle >> quickMerge: aRectangle [
-	"Answer the receiver if it encloses the given rectangle or the merge of the two rectangles if it doesn't. THis method is an optimization to reduce extra rectangle creations."
+	"Answer the receiver if it encloses the given rectangle or the merge of the two rectangles if it doesn't. This method is an optimization to reduce extra rectangle creations."
 	| useRcvr rOrigin rCorner minX maxX minY maxY |
 
 	aRectangle ifNil: [ ^self ].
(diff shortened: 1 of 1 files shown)


jecisc (approved):

src/Kernel/Rectangle.class.st
```diff
@@ -751,10 +751,10 @@ Rectangle >> quickMerge: aRectangle [
 			[ useRcvr := false.
 			rCorner y ]
 		ifFalse: [ corner y ].
-	useRcvr
-		ifTrue: [ ^ self ]
+	^ useRcvr
+		ifTrue: [ self ]
 		ifFalse:
-			[ ^ Rectangle
+			[ self class
```
vonbecmann: i'm not sure about this change. CharacterBlock inherits from Rectangle.

jecisc: I have seen no sender of this method that seems to be on CharacterBlock.

IMO this is a good change because if we merge CharacterBlocks we expect a new character block. Same with other subclasses other people could do

vonbecmann: reverting change.

vonbecmann:
using reference to Rectangle. because it's not the same

Ducasse:
Why did you revert? Because may be this is ok to have self class.
I just mentioned that this is not equivalent - not that the change was a problem per se.

vonbecmann:
> Why did you revert? Because may be this is ok to have self class. I just mentioned that this is not equivalent - not that the change was a problem per se.

i change my mind, i wasn't sure about it. there's no problem.

PR #13373, opened 2023-04-10
ClassOrganization classification cleanings

jecisc:
This PR does some cleanings in ClassOrganization.

## Cleaning of #classify:inProtocolNamed:

I updated #classify:inProtocolNamed: in multiple ways:
- Now the method can take a protocol name OR a real protocol as parameter
- If the method is been reclassified and the old protocol becomes empty, we remove it. In the past this was only done if this method was invoked through #classify:under:
- In the end, this method was doing exactly the same as #classify:under: but without the reclassification announcement. Since I had to rename the method to remove the "inProtocolNamed:" I decided to rename it #silentlyClassify:under: in mirror of #classify:under:

## Cleaning of #classify:under: 

Last change is in #classify:under: with the goal of making the code more readable than it was before. And I think I did a good job on that.

## Futur step

There is one thing I will need to check after this change, it's the senders of #silentlyClassify:under:. We should make sure that we do not forget to announce some classifications.
I found 4 senders of this method:
- #classify:under: which is natural
- A method invoked by the class builder to copy the protocols of the old class into the new one. This one should naturally happen silently but instead of classifying everything in the new class, maybe it should just copy the collection of protocols? This seems cleaner and better optimized
- The two last senders comes from traits. Since this is not an easy part of the system I will need to understand the mechanic and check why they need a silent classification. Maybe it is necessary, or maybe it is just a bad use of the API since #classify:under: and #classify:inProtocolNamed: were not really revealing of their difference...


jecisc:
If I want to make the CI happy I will have to fix the copy issue first apprently

jecisc:
But to make the improvement on the copy I first need https://github.com/pharo-project/pharo/pull/13356 to be integrated so I will pause the improvements on protocols until this is done probably.

src/Kernel/ClassOrganization.class.st
```diff
@@ -240,12 +217,9 @@ ClassOrganization >> protocolNameOfElement: aSelector [
 { #category : #accessing }
 ClassOrganization >> protocolNameOfElement: aSelector ifAbsent: aBlock [
 
-	^ (self protocolsOfSelector: aSelector)
-		  ifEmpty: [
-			  (organizedClass includesSelector: aSelector)
-				  ifTrue: [ Protocol unclassified ]
-				  ifFalse: [ aBlock value ] ]
```
jecisc: Here I removed this hack because I checked in a running image and I have no selector that is in the class but not in a protocol and I think we should not allow that state to happen

jecisc:
We have some broken tests. Some categories seems to be missing from what release tests are saying. 

This points that we are missing tests on recategorization. When I'll have time I'll check to add some and look for the problem in my PR

jecisc:
Every test I do in my image works fine, I cannot reproduce the errors of the CI :(

jecisc:
Maybe the problem comes from #copyFrom: since I added 30 assertions on #classify:under: and they all pass. 

I'll need to add some tests on that part too.

jecisc:
I added tests on the copy and the test was already green. It's hard to find the origin of this problem so I might just split this one in smaller PRs.

jecisc:
It seems that my recent fix succeeded to fix this PR :)

MarcusDenker (approved):

Issue #13364, opened 2023-04-09
[CI] Ring tests cause segmentation fault on GC


jecisc:
I saw multiple times segmentation fault in the CI while running the tests. This seems to be caused by Ring tests. 

For example I have a case with testBehavior.

Since this happens during a GC I have no idea if this comes from Ring or from the VM.

```
Segmentation fault Sun Apr  9 03:43:01 2023

/builds/workspace/est_and_branch_Pipeline_PR-13360/Tests-unix-64/pharo-vm/lib/pharo
PharoVM version:5.0-Pharo 10.0.3-7d7f8d3 built on Apr  7 2023 23:09:25 Compiler: 5.4.0 20160609 5.4.0 20160609 [Production Spur 64-bit VM]
Built from: CoInterpreter VMMaker-tonel.1 uuid: c3153026-27b5-0d00-9fe0-21900fd32d39 Apr  7 2023
 With:StackToRegisterMappingCogit VMMaker-tonel.1 uuid: c3153026-27b5-0d00-9fe0-21900fd32d39 Apr  7 2023
 Revision: 7d7f8d3 - Commit: 7d7f8d3 - Date: 2023-04-07 22:44:03 +0200

C stack backtrace & registers:
	rax 0x19399c38 rbx 0x7f8f41fbb758 rcx 0xffffffffff91fccb rdx 0x15c98290
	rdi 0x7f8b3e3209a0 rsi 0x7f8b3e3209a0 rbp 0xffffffffffffff rsp 0x7ffcc21e07d0
	r8  0x1df43780 r9  0x1df43778 r10 0xffffffffffffff r11 0x7f8b3e3209a0
	r12 0x00000001 r13 0x00000000 r14 0x000a8c01 r15 0x1925d9e0
	rip 0x7f8f41c9b360

*/builds/workspace/pharo-vm_pharo-10/build-stockReplacement/build/vm/libPharoVMCore.so(+0x65360)[0x7f8f41c9b360]
/builds/workspace/pharo-vm_pharo-10/build-stockReplacement/build/vm/libPharoVMCore.so(+0xb1eec)[0x7f8f41ce7eec]
/builds/workspace/pharo-vm_pharo-10/build-stockReplacement/build/vm/libPharoVMCore.so(doReport+0xb5)[0x7f8f41ce81d5]
[...]

PR #13376, opened 2023-04-10
Clean MCMethodDefinitionTest

jecisc:
This issue is one of the many steps needed to fix #13349.

MCMethodDefinitionTest is not really nice and leaves the system in a dirty states.
First it generates a package that it does not remove.
Second it recompiles its own code 
Third it updates the Monticello mock package present in the image

This is an cleaning so that every code compilation happening is done in generated packages that get cleaned so that we do not dirty the code commited to Pharo.

### DIFF
--- a/src/Monticello-Tests/MCMethodDefinitionTest.class.st
+++ b/src/Monticello-Tests/MCMethodDefinitionTest.class.st
@@ -9,8 +9,10 @@ Class {
 	#category : #'Monticello-Tests-Base'
 }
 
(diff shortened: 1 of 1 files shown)


jecisc:
[CI] Ring tests cause segmentation fault on GC #13364 
 
 Still with #testBehavior 

 [CI] test runs sometimes aborted due to SubscriptOutOfBounds error #12502

MarcusDenker (approved):

PR #13378, opened 2023-04-10
refactor ASTCache

privat:
#13368 and #13371 are a mess because they tried to do too many things at once, like refactoring the class and changing the behavior.

So the PR only does the refactoring. Therefore, if successful, the behavior change will be easier to do.

* Keep WeakIdentityKeyDictionary but use it as a ivar
* Cleanup methods
* Assume ReflectiveMethods strongly hold the AST

### DIFF
--- a/src/AST-Core/ASTCache.class.st
+++ b/src/AST-Core/ASTCache.class.st
@@ -17,7 +17,10 @@ Code that wants to modify the AST without making sure the compiledMethod is in s
 "
 Class {
 	#name : #ASTCache,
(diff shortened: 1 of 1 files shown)


jecisc (approved):

MarcusDenker (approved):

MarcusDenker:
If we now set the #ast property on ReflectiveMethods and the compiled version of them, then we can just remove the whole #getASTFor: mechanism and the check for the #reflectiveMethod in the ASTCache

PR #13379, opened 2023-04-10
ASTCache: use a double weak dictionary

privat:
Based on #13378 for the refactor. Look at the last commit only.

Instead of a `WeakIdentityKeyDictionary`, that keep forever uninteresting AST.
Use a `WeakIdentityKeyWeakValueDictionary`! But because this class does not exist, and because I do not want to implement is (and because ephemerons are coming), I just simulate it with a `WeakIdentityKeyDictionary` of single-cell `WeakArrays`, and a special loop to remove garbage collected value.

For the rationale, look at https://github.com/pharo-project/pharo/pull/13371#issuecomment-1501517196

### DIFF
--- a/src/AST-Core/ASTCache.class.st
+++ b/src/AST-Core/ASTCache.class.st
@@ -105,9 +105,12 @@ ASTCache >> at: aCompiledMethod ifAbsentPut: aBlock [
 ASTCache >> at: aCompiledMethod put: aRBMethodNode [
 
 	"Cleanup weak AST. Note `associations` return a copy, so the iteration is safe"
+	| weakRef |
 	self weakDictionary associations do: [ :each |
 		(each value at: 1) ifNil: [ self weakDictionary removeKey: each key ] ].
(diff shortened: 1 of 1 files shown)


privat:
Test passed, but the code is ugly.

Some stats (see #13368 for other numbers)

* after DrTest all tests (1248) of OpalCompiler-Tests: `an ASTCache#2 73%` (only 2 items instead of 67)
* after all critiques on Kernel: `an ASTCache#127 99%` (only 127 items instead of 4317)
* after one critique on Kernel: `an ASTCache#199 0%`
* after all tests (342) of Reflectivity-Tests: `an ASTCache#619 97%`

So the numbers are lower but are not that reliable: elements are collected on future GC, and it seems that `WeakIdentityKeyDictionary` have inconsistent size value: `{self associations size. self size} >>> #(129 619)`. so possible the 619 of above is false.

jecisc (approved):

PR #13383, opened 2023-04-10
Remove #protocolsOfSelector: and implement #protocolOfSelector:

jecisc:
This remove #protocolsOfSelector: because a selector has only 1 protocol, so why should we manipulate a collection when we can only have 0 or 1 element?

Instead I created a method #protocolOf: that returns the protocol of a method selector or nil. I adapted the users accordingly to make the code cleaner.

In future PR this will also help me to clean more code.
I plan to deprecate #protocolNameOfSelector:[ifAbsent:] to use this new method. Some senders can use a real protocol, and others can just send #name to the return of #protocolOf:

### DIFF
--- a/src/Kernel/ClassOrganization.class.st
+++ b/src/Kernel/ClassOrganization.class.st
@@ -75,7 +75,7 @@ ClassOrganization >> classify: aSymbol inProtocolNamed: aProtocolName [
 { #category : #'backward compatibility' }
 ClassOrganization >> classify: selector under: aProtocol [
 
-	| oldProtocolName forceNotify oldProtocols protocolName |
+	| oldProtocolName forceNotify oldProtocol protocolName |
 	"The next section deserve more cleanings.
 	
 	Some code was added to make it possible to classify giving a real protocol.
@@ -87,10 +87,10 @@ ClassOrganization >> classify: selector under: aProtocol [
 	forceNotify := (self includesSelector: selector) not.
 	oldProtocolName := self protocolNameOfElement: selector.
 	(forceNotify or: [ oldProtocolName ~= protocolName or: [ protocolName ~= Protocol unclassified ] ]) ifFalse: [ ^ self ].
-	oldProtocols := self protocolsOfSelector: selector.
+	oldProtocol := self protocolOf: selector.
 
 	self classify: selector inProtocolNamed: protocolName.
-	oldProtocols do: [ :protocol | self removeProtocolIfEmpty: protocol ].
+	oldProtocol ifNotNil: [ self removeProtocolIfEmpty: oldProtocol ].
 	oldProtocolName ifNotNil: [ self notifyOfChangedSelector: selector from: oldProtocolName to: protocolName ]
 ]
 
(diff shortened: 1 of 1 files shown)


src/Kernel/ClassOrganization.class.st
```diff
@@ -268,16 +265,18 @@ ClassOrganization >> protocolNames [
 	^ self protocols collect: [ :protocol | protocol name ]
 ]
 
-{ #category : #accessing }
-ClassOrganization >> protocols [
+{ #category : #protocol }
+ClassOrganization >> protocolOf: aSelector [
```
privat: With the simple name I'm afraid users (myself included) will mistakenly use a CompiledMethod as argument, because I don't care about the selector that much, but I care about the method.

privat (commented):
I not sure, if a method is in no protocol. Do we get nil or do we get asyetuncladdified?

src/Kernel/ClassOrganization.class.st
jecisc: A method of a class is always in a protocol even if it is an instance of Protocol named #as yet unclassified.

The case where it returns nil is if we ask the protocol of a method that is not in the class. Instead we should probably raise an error but that changes the current behavior and I prefer do this change later when the protocol management will be cleaner in the image

jecisc:
A method of a class is always in a protocol even if it is an instance of Protocol named #as yet unclassified.

The case where it returns nil is if we ask the protocol of a method that is not in the class. Instead we should probably raise an error but that changes the current behavior and I prefer do this change later when the protocol management will be cleaner in the image

src/Kernel/ClassOrganization.class.st
jecisc: I will rename it into #protocolOfSelector: tomorrow.

jecisc:
I have no idea how we end up in the error of the CI.. I'll check that later

src/Kernel/ClassOrganization.class.st
privat: Another possibility could be to behave correctly if a method is given.

jecisc: Well, if we have a compiled method, I plan to update #protocol to return an instance of protocol instead of a protocol name in the near future. 

I'm not sure it's worth to have a class knowing how to return the protocol of a method when a method will be able to directly return it soon

jecisc: I renamed into #protocolOfSelector: because I don't think the class should have the responsibility of returning the protocol of a method since it should be the responsibility of the method.

privat: That makes a lot of sense

jecisc:
Error seems unrelated.

The revert I had to do show that somewhere in the system a class has methods that are not categorized in the class organization!
This is bad and should be fixed, but in a later PR.

MarcusDenker (approved):

MarcusDenker:
For the whole Protocol stuff... the questions are always on my mind:

- is Protocol really the right term? Protocol implies "this class implements a Protocol"
-- but we do not use this at all like that ("accessing" is not something like that at all)
-- "Protocol" in the sense of "it implements the XXX protocol" would much more like a Trait, but for traits we want to tag methods as "accessing", too

- why just allow one? why not allow to put a method into "accessing" and "public API" at the same time?

PR #13393, opened 2023-04-11
Try to move workspace variables further (lite)

privat:
This is a reroll of #13323 ~~without~~ with less sugar or GUI concern

Workspace variables are a very problematic feature with a lot of defaults (UI, spec, implementation).

The compiler have to deal with heuristics and uncooperative requestors (GUI windows) to make them work, including:

* Having (and keeping) references on GUI windows when managing scope
* Heuristic trying to guess if we are compiling or not (just highlighting)
* Fear of asking if a variable exists, because asking so might create it

So this PR tries to clean up the current design (*implementation* design, not *GUI* design) a little.

I also tried to no impact the current UI and spec of (workspace variables are still as bad/good as before). Because it is out of scope of the PR, but also because changing what workspace variables are implies hacking GUI tools that lives in another repositories (and I'm not a GUI guy and PR synchronization between repository are a pain).

So, what is on the (reduced) menu:

* Handle workspace variables with the same idea that was used for undeclared  (see #VariableNotDeclared): get a local placeholder variable at parse/semantic-time and only register it (here to the requestor/GUI) at compile-time.
  This has a nice side effect: undeclared variable are now "blue" in the workspace (I can change that if you do not like it).
* remove most of the insane methods `lookupVar:declare:` and `lookupVarForDeclaration:` that are now unneeded (I suppose)

What was removed:

* Reactivate the "undeclared" menu → #13392
* Do not add requestor in scopes *except* for workspace-type requestors, that still need → wait for open season in spec/newtools

### DIFF
--- a/src/Debugging-Core/Context.extension.st
+++ b/src/Debugging-Core/Context.extension.st
@@ -64,7 +64,7 @@ Context >> lookupVar: aSymbol [
 
 { #category : #'*Debugging-Core' }
 Context >> lookupVar: aSymbol declare: aBoolean [
-	^ self astScope lookupVar: aSymbol declare: aBoolean
+	^ self astScope lookupVar: aSymbol
 ]
 
 { #category : #'*Debugging-Core' }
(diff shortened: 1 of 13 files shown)


privat:
Ok, I remember now why I needed to hack spec/newtools. It's not *that* independent :(
Pushed an attempt to make it minimal.

MarcusDenker (approved):

PR #13398, opened 2023-04-11
Compiler: simplify DoIt

privat:
The management of DoIt (executable scripts) is a little hairy and complex.

* Invisible synthetic return statements are injected in `parseDoIt` and `asDoIt`, but not in `parseExpression`.
* `asDoit` return a `RBMethodNode` instead of a `RBoItMethodNode`
* `asDoit` results have an inconsistent source
* some AST methods are unneeded extension methods

This PR tries to make the behavior less complex:

* original source is not altered (even after `asDoit`)
* no returns are injected in the AST. The default return behavior is implemented at compile time by the transformator, as it already decides for classic methods with/without a final return.


jecisc (approved):
Nice!

Only thing I'm not sure about is OCASTTranslator >> visitMethodNode: since I don't know well the compiler, but the code seems good :)

MarcusDenker (approved):
Nice!

PR #13402, opened 2023-04-12
Completion can extends keyword messages

privat:
Completion is sometime very frustrating (but the code is worse).
One thing that did bug me enough to hack the code is the absence of completion on keyword messages after the first argument.

Now it tries to complete an existing keyword message with additional keyword parts.

Capture d’écran du 2023-04-11 20-27-57

Note: also replace `isKeyword` on some AST nodes with something more general and more efficient


jecisc (approved):
Nice!

MarcusDenker (approved):

PR #13403, opened 2023-04-12
Make Smalltalk environnement return a RPackageOrganizer

jecisc:
This is a step toward the goal of integrating RPackage in the default metamodel and coupling it more to remove the magic of SystemOrganizer. 

I already made RPackageOrganizer polymorphic with SystemOrganizer. Now `Smalltalk environnement` return the default RPackageOrganizer! RPackageOrganizer still wraps SystemOrganizer, but it is on death row.


jecisc:
Once this is integrated, I just need to update something in Microdown and we will be down to only 1 reference of SystemOrganizer in the image.

jecisc:
Espell is not happy with that change :(

jecisc:
This time the bootstrap worked but 20 tests by OS are failing. That's progress!

jecisc:
If i download the artifact and run the failing tests they are all green :(

jecisc:
I have some code that is just missing in the new image I don't know why. For example ClassFactoryForTestCaseTest>>#tearDown is just not present in the artifact

jecisc:
I got the tests to pass! this one will be a huuuuge help to finally clean RPackage!

src/RPackage-Tests/RPackageOrganizerSystemTest.class.st
```diff
@@ -0,0 +1,31 @@
...
+	"When no environment is specified, point to the default one"
+
+	self assert: RPackageOrganizer new environment identicalTo: Smalltalk globals
+]
+
+{ #category : #tests }
+RPackageOrganizerSystemTest >> testThatExistingPackageNamesDoesNotContainIllegalCharacters [
+	| illegalCharacters |
+
+	illegalCharacters := #( $\ $/ $: $* $? $" $< $> $| ).
```
privat, line 27: so `;` or `¿` are fair-game? Or starting with ` -` is OK?
I would prefer a whitelist (alphanum and `-`?) and that a test predicate is introduced outside the tests so that tools, rules or whatever can rely on the same policy.
Since there is already `String>>#isValidGlobalName` and `String>>#isValidSelector`, having `String>>#isValidPackageName` makes sense

src/System-Support-Tests/SystemDictionaryTest.class.st
```diff
@@ -102,7 +102,7 @@ SystemDictionaryTest >> testSetOrganizationSetBackPointer [
 
 	| sd so |
 	sd := SystemDictionary new.
-	so := SystemOrganizer new.
+	so := RPackageOrganizer new.
```
privat: so `ro`, no?

src/RPackage-Tests/RPackageOrganizerSystemTest.class.st
jecisc: I'm not a big fan of this but here I'm just migrating an existing test :)
We can iterate on this in future PRs

src/System-Support-Tests/SystemDictionaryTest.class.st
jecisc: I gave better names (systemDictionary and packageOrganizer because I **hate** abbreviations in code)

src/RPackage-Tests/RPackageOrganizerSystemTest.class.st
privat: Oh, right. It was in the deleted file. I'm fine with future PR then

jecisc:
Windows is the only problem on the CI currently

privat (approved):

MarcusDenker (approved):

PR #13405, opened 2023-04-12
Remove ClassFactoryForTestCaseTest>>#testDuplicateClassWithNewName

jecisc:
It has no sender and the test associated creates obsolete classes in the system.


jecisc:
There is one failure that seems linked to the AST cache refactoring and not my PR

privat (commented):
`duplicateClassWithNewName:` is no more tested?

jecisc:
It’s true it was tested transitively. I’ll just update the existing test instead of removing it.

jecisc:
> `duplicateClassWithNewName:` is no more tested?

I checked and it is used by some other tests in FluidClassBuilder tests and ClassTest. So it should be fine to remove this test.

PR #13406, opened 2023-04-12
Cut dependency from Collection-Streams to UIManager

jecisc:
Collection-Streams had 2 methods dealing with the UI which is a bad dependency IMO. I could have added thos methods as an extention of the UI packages but they have no sender in the image. I propose to deprecate them. If a user wants to deal with a display bar he can do it himself.


privat (approved):

MarcusDenker (approved):

Ducasse:
Thanks :)

PR #13407, opened 2023-04-12
Fix dependies of RPackage-Core

jecisc:
Mulitple problems:
- UIManager was declare as a dependency but it was Transcript the real dependency
- Jobs, Announcements-Core and Monticello were missing as dependencies
- #packageDefiningOrExtendingMethod: and #packageDefiningOrExtendingSelector:inClass: were bringing an unwanted dependency on Traits while the code is dead and is calling unimplemented methods ==> Remove them. No need for deprecation, those methods can only lead to errors

### DIFF
--- a/src/RPackage-Core/ManifestRPackageCore.class.st
+++ b/src/RPackage-Core/ManifestRPackageCore.class.st
@@ -9,5 +9,7 @@ Class {
 
 { #category : #'meta-data - dependency analyser' }
 ManifestRPackageCore class >> manuallyResolvedDependencies [
-	^ #(#UIManager)
+
+	<ignoreForCoverage>
+	^ #(#Jobs #'Transcript-Core' #'Announcements-Core' #Monticello)
 ]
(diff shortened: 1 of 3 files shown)


privat (approved):

MarcusDenker (approved):

Ducasse:
When I see all these dependencies on Core package I see that we do not pay attention to such aspects.

MarcusDenker:
- SystemDependenciesTest is failing

jecisc:
> When I see all these dependencies on Core package I see that we do not pay attention to such aspects.

It’s the reason I’m trying to cut some dependencies from time to time :)

jecisc:
> * SystemDependenciesTest is failing

It’s probably because some dependencies were present but missed by the tests. I’ll update them.

jecisc:
I updated the system dependencies tests. 

Since RPackage-Core had missing dependencies in his list of manually resolved dependencies, the tests were not correct and I fix them. 

It's sad because it shows that we have more unwanted dependencies than we thought we had. But at least we now know that we need to kill 3 dependencies in the system.

MarcusDenker (approved):

PR #13415, opened 2023-04-12
Change Protocol  to use an array for methodSelectors directly

MarcusDenker:
- 40K less IdentitySets alocated
- all the arrays are much smaller

I changed the API slghtly to be more sane and return "self", not the added element or the collection changed.

### DIFF
--- a/src/Kernel/Protocol.class.st
+++ b/src/Kernel/Protocol.class.st
@@ -40,7 +40,9 @@ Protocol >> addAllMethodsFrom: aProtocol [
 
 { #category : #'adding-removing' }
 Protocol >> addMethodSelector: aSymbol [
-	^ methodSelectors add: aSymbol
+	(methodSelectors includes: aSymbol) ifTrue: [ ^ self ].
+	
+	methodSelectors := methodSelectors copyWith: aSymbol
 ]
 
 { #category : #testing }
@@ -59,7 +61,7 @@ Protocol >> initialize [
 
 	super initialize.
 
-	methodSelectors := IdentitySet new.
+	methodSelectors := Array empty.
 	name := self class unclassified
 ]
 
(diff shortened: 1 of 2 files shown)


MarcusDenker:
very odd crash on full image load

jecisc:
A PR of Esteban also had a segmentation fault while loading the packages. Maybe there is a bug in the VM?

jecisc:
See: https://ci.inria.fr/pharo-ci-jenkins2/blue/organizations/jenkins/Test%20pending%20pull%20request%20and%20branch%20Pipeline/detail/PR-13412/2/pipeline/25

<blockquote><div><strong><a href="https://ci.inria.fr/pharo-ci-jenkins2/blue/organizations/jenkins/Test%20pending%20pull%20request%20and%20branch%20Pipeline/detail/PR-13412/2/pipeline/25">Jenkins</a></strong></div></blockquote>

src/Kernel/Protocol.class.st
```diff
@@ -59,7 +61,7 @@ Protocol >> initialize [
 
 	super initialize.
 
-	methodSelectors := IdentitySet new.
+	methodSelectors := Array empty.
```
jecisc:
```suggestion
	self removeAllMethodSelectors.
```

Maybe we could rename it #resetMethodSelectors instead later

jecisc (approved):

MarcusDenker:
I got the same crash on #13415

privat (approved):

privat:
I got the same CI issue. I opened #13419

jecisc:
testHaltOnCount – MacOSX64.Debugging.Utils.Tests.HaltTest

It's the second time I see this random failure this week :(

PR #13418, opened 2023-04-12
Calypso: check if node is nil before calling isCommentNode and cie

### LINKED ISSUE #13414: Browser window resizing causes Red Square of Death
**Bug description**
For some methods, resizing of the browser window is causing Red Square of Death

**To Reproduce**
1. open Spotter, type `methodAdded`
2. go to `ChangeSet>>#methodAdded:`
3. increase the browser size
4. you get someting like

image

**Version information:**
 Pharo 11.0.0
Build information: Pharo-11.0.0+build.672.sha.87a0f07f1706902298dd944d7f17a4f9611a0a21 (64 Bit)

privat:
fix #13414 (maybe)

I could not reproduce the bug on my machine :( but there are only 4 senders of `isCommentNode`, two of them did check is the receiver is nil, and the other two did not.
Now all four of them check nil.
Disclaimer: I have no idea what these methods does.

@pavel-krivanek can test and report if this fix your issue?


privat:
new CI crash #13419? Pharo11 isn't rock solid?

pavel-krivanek (approved):
seems to fix the issue

jecisc (approved):
I think that using guard clauses might be more readable but this fixes the bug at least :)

privat:
> I think that using guard clauses might be more readable but this fixes the bug at least :)

The existing code use stairs of `and:`, so I did just maintain the existing flow and style.

tesonep:
I have merged P11 to get the fix of @guillep about the primitive

guillep (approved):
Good!

PR #13421, opened 2023-04-12
AST: starting visiting comments

privat:
Tries to make comments, first-class citizens of the AST. Or at least second-class citizens... Still better than the cargo hold.

This includes:

* new method `childrenAndComments` to have both (direct) children and (direct) comments
* `nodesDo:` (and related) visit comments
* `nodeForOffset:` can return a comment node
* add `visitCommentNode:` to the `RBProgramNodeVisitor`.

The first three should be safe since clients using them cannot expect specific nodes and just select what they need.
The last one should also be safe because comments are not automatically visited (yet!), but it might allow custom clients to use the hook for custom visit.

The API of `RBProgramNodeVisitor` need love, but there are a lot of clients, so it's not simple to improve it.
Maybe a simpler approach could be having a second visitor, then switch clients one by one.

### DIFF
--- a/src/AST-Core-Tests/RBCodeSnippet.class.st
+++ b/src/AST-Core-Tests/RBCodeSnippet.class.st
@@ -2541,8 +2541,7 @@ RBCodeSnippet >> updateExpectations [
 	ast := self parse.
 	self formattedCode: ast formattedCode withSeparatorsCompacted.
 	self isParseFaulty: ast isFaulty.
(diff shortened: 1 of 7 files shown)


jecisc:
YES! \0/

This is on my todo list since 4 years now (but not enough time…). Thank you :)

MarcusDenker (approved):

Issue #13419, opened 2023-04-12
CI VM crash. MetacelloNotification BaselineOfDisplay postload:package


privat:
CI crashes the VM at the full image step. eg https://ci.inria.fr/pharo-ci-jenkins2/blue/organizations/jenkins/Test%20pending%20pull%20request%20and%20branch%20Pipeline/detail/PR-13418/1/pipeline

```
MetacelloNotification: Evaluating -> baseline [BaselineOfDisplay] >> postload:package:
Segmentation fault Wed Apr 12 16:36:41 2023
/builds/workspace/est_and_branch_Pipeline_PR-13418/64/bootstrap-downloads/vmtarget/pharo-vm/lib/pharo
PharoVM version:5.0-Pharo 10.0.3-59e68d0 built on Apr 12 2023 11:45:43 Compiler: 5.4.0 20160609 5.4.0 20160609 [Production Spur 64-bit VM]
Built from: CoInterpreter VMMaker-tonel.1 uuid: 64857c2d-82b5-0d00-9fdd-ac5b0fd32d39 Apr 12 2023
 With:StackToRegisterMappingCogit VMMaker-tonel.1 uuid: 64857c2d-82b5-0d00-9fdd-ac5b0fd32d39 Apr 12 2023
 Revision: 59e68d0 - Commit: 59e68d0 - Date: 2023-04-12 11:07:13 +0200
[...]
All Smalltalk process stacks (active first):
Process      0x10000c88808 priority 40
       0x280007d28 I DisplayScreen>setExtent:depth: 0x340d1ea00: a(n) DisplayScreen
```

guillep:
Fixed in #13427

PR #13422, opened 2023-04-12
faulty code: test runtimeSyntaxError

privat:
New Kernel tests that ensure that signaled `RuntimeSyntaxError` exceptions are correct and that their contexts are sane with usable source code information.

Because of all the good work about faulty compilation, there is no special case and faulty code is correctly caught and the source code information is useful. So debugger should behave correctly.
These tests check 3 scenarios: doit (scripts), installed method, and uninstalled method. And they will help to prevent future regressions.

Note: I understand what `debuggerCompleteToSender` does, but I have trouble understanding the English of the symbol string. What "*complete*" means here?

Update: I did another pass and added commits

I removed the AST node from the Exception `RuntimeSyntaxError`. That make the model saner. And the node was superfluous. Because the context (and debuggers and everything) can (now) retrieve the original source code and AST node exactly like other exceptions.
So now, the exception only have a `messageText` like most other exceptions, and is computed early at compile time.
Doing so, I realized the that `RuntimeSyntaxError` what the general runtime exception also for other kind of runtime semantic error. It could be improved because there is "*syntax*" in `RuntimeSyntaxError` (it's hidden in the middle if you don't see it).

### DIFF
--- a/src/Kernel-Tests-WithCompiler/RuntimeSyntaxErrorTest.class.st
+++ b/src/Kernel-Tests-WithCompiler/RuntimeSyntaxErrorTest.class.st
@@ -21,7 +21,7 @@ RuntimeSyntaxErrorTest >> testFaultyInstalledMethod [
 			self
 				assert: error description
 				equals: 'RuntimeSyntaxError: Unknown character'.
-			ctx := error signalContext skipDebuggerComplete.
+			ctx := error signalContext filterDebuggerStack.
 			self assert: ctx sourceNodeExecuted sourceCode equals: '¿' ].
 
 	method removeFromSystem
@@ -42,7 +42,7 @@ RuntimeSyntaxErrorTest >> testFaultyScript [
 			self
 				assert: error description
 				equals: 'RuntimeSyntaxError: Unknown character'.
-			ctx := error signalContext skipDebuggerComplete.
+			ctx := error signalContext filterDebuggerStack.
 			self assert: ctx sourceNodeExecuted sourceCode equals: '¿' ]
 ]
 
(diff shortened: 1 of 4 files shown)


privat:
Note: To confirm, I did also run a GUI session with faulty code, debugger, and reparation attempts. All checks are checked! :tada:

See Sect. "Current debugger GUI on faulty code"in #12883

Capture d’écran du 2023-04-12 20-46-47

jecisc:
CI VM crash. MetacelloNotification BaselineOfDisplay postload:package #13419 
 
 :/

guillep:
Found the cause. That method calls

```
supportsDisplayDepth: pixelDepth
	"Return true if this pixel depth is supported on the current host platform.
	Primitive. Optional."
	<primitive: 91>
	^#(1 2 4 8 16 32) includes: pixelDepth
```

But primitive 91 (which was unused) was redefined in the latest VM:

```
	"Perm Space primitives"
		(90 primitiveMoveToPermSpace)
		(91 primitiveMoveToPermSpaceInBulk)
		(92 primitiveIsInPermSpace)
```

guillep:
One solution is to remove the <primitive: 91> from the offending method in P>=11

jecisc (approved):
I also don't like the name debuggerCompleteToSender.

The next step you talked about seems good to me from what you are saying.

MarcusDenker:
yes, "debuggerCompleteToSender" is not good. And there was the idea that one negative aspect of pragmas is that they have no place to put documentation... there were Pragma implementations in other Smalltalks that required to create a subclass of Pragma for each pragma symbol used... it could be interesting to think how to add something like that (but maybe optional)

MarcusDenker (approved):

PR #13426, opened 2023-04-13
Make protocol of super method a priority in MethodClassifier

### LINKED ISSUE #13413: Automatic method categorization does not repsect superclass category
**Bug description**
When creating a method, the automatic assigning of a protocol ignores the superclass classification.

**To Reproduce**
1. go to class `AIBFS` (first package in the image, second class)
2. create empty method `addNodeFor: aModel`
3. this method has superclass implementation in the protocol `building - graph` but despite that, it is automatically added to the protocol named `adding`. 

The behavior is actually even more crazy for other cases: 
- when you first select a different protocol (like `accessing`) and create the method there, it is anyway automatically added to the protocol `adding`
- if you first create the protocol of the desired name (`building - graph`) and create the new method inside it, it gets anyway recategorized as `adding`. Moreover, the protocol `building - graph` is removed!

It does that for other method names where it guesses the protocol name like `isRunningTasks`.

**Version information:**
 Pharo 11.0.0
Build information: Pharo-11.0.0+build.672.sha.87a0f07f1706902298dd944d7f17a4f9611a0a21 (64 Bit)

jecisc:
MethodClassifier suggest to classify a method in the same protocol than its overriden method if one exist.  But this suggestion has the lowest priority of all. Iâ¯suggest to give it the highest priority, especially since we have ReInconsistentMethodClassificationRule that yells at us if it is not the case.

I also adde a test to avoid an accidental reversal of the behavior.

Fixes #13413

### DIFF
--- a/src/Tool-Base/MethodClassifier.class.st
+++ b/src/Tool-Base/MethodClassifier.class.st
@@ -198,8 +198,20 @@ MethodClassifier >> classifyAll: aCollectionOfMethods [
 { #category : #accessing }
 MethodClassifier >> listToFindProtocols [
 
-	^ #( #protocolByKnownMapping: #protocolAccessorFor: #protocolByKnownKeywordSuffix: #protocolByKnownPragma: #protocolByKnownPrefix: #protocolByOtherImplementors:
-	     #protocolInSuperclassProtocol: )
+	^ #( #protocolInSuperclassProtocol: #protocolByKnownMapping: #protocolAccessorFor: #protocolByKnownKeywordSuffix: #protocolByKnownPragma:
+	     #protocolByKnownPrefix: #protocolByOtherImplementors: )
+]
+
+{ #category : #accessing }
+MethodClassifier >> protocol [
+
+	^ protocol
+]
+
+{ #category : #accessing }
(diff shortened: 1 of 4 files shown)


src/Tool-Base/MethodClassifier.class.st
```diff
@@ -198,8 +198,20 @@ MethodClassifier >> classifyAll: aCollectionOfMethods [
...
+]
+
+{ #category : #accessing }
+MethodClassifier >> protocol [
+
+	^ protocol
+]
+
+{ #category : #accessing }
+MethodClassifier >> protocol: anObject [
```
demarey: why is there a setter?

jecisc: I just generated the accessors but I can remove this one since protocol is just an implementation detail

demarey (approved):

privat (approved):

MarcusDenker:
the PR is against Pharo12, but maybe this is something to fix in Pharo11, too?

jecisc:
https://github.com/pharo-project/pharo/pull/13432

> the PR is against Pharo12, but maybe this is something to fix in Pharo11, too?

PR #13428, opened 2023-04-13
Pharo 11 weekly synch

jecisc:
This one is really interesting for the old primitive fix


jecisc:
Hum... Seen the list of changes I thing I messed something up with my synch because it revert some changes we want to keep

privat:
You should not include the commits with the stable versions of dependencies

PR #13429, opened 2023-04-13
Weekly Pharo11 sync

jecisc:
Weekly Pharo 11 sync. 

In this merge I rejected the change of versions in the baselines so that P12 keep loading the integration branches


jecisc:
Hum.. Something about finalization seems to have gotten lost somewhere

PR #13432, opened 2023-04-13
Backport #13426: Make protocol of super method a priority in MethodClassifier

jecisc:
Backport the fact that the protocol of the super method has the priority in MethodClassifier


jecisc:
I did not miss the Github API limit rate.. :)

privat (approved):

PR #13435, opened 2023-04-13
Remove dead code in Bootstrap

### LINKED ISSUE #13434: Bootstrap: Method #createInitialObjects is present 2 times in the bootstrap
I am trying to update some things in the bootstrap and to locate where some initializations are executed I did a textual search in the pharo repo. 

I got surprised because I found 2 locations that seems to be duplicated code.
- PBImageBuilder50 >> createInitialObjects
- PBBootstrap30 >> createInitialObjects

It would be great to be able to remove this duplication because it wake it harder to update the bootstrap if we don't have a good knowledge of it.

jecisc:
Originally the Bootstrap was planned to be an external project that should have been able to bootstrap any version of Pharo.

But since the introduction of the bootstrap, it lives with Pharo. 

Here is a first cleaning step that removes what I think is dead code inherited from the time where more than one version of Pharo could have been build.

Fixes #13434

Next step: Rename the classes to remove the 30 and 50 that are still in the names.


guillep:
Have you seen this very green CI? 
imagen
 :D

privat (approved):

astares:
@guillep still too much white - any chance to get it even greener?

PR #13436, opened 2023-04-13
Reduce dependencies of SessionManager in the core of Pharo

jecisc:
Currently in the KernelGroup of the bootstrap we have 4 packages that are only needed by the SessionManager of Pharo.  I want to check if we can extract the SessionManager from the core of the bootstrap and make it load via Hermes instead of Espell.

Here is a first step toward that goal. Now all registrations of KernelGroup classes are done in SessionManager class>>#initializeKernelRegistrations to avoid dependencies to SessionManager. 

If this pass, then the only dependency we will still have is SmalltalkImage. But that is for another PR since it is not so simple to cut this dependency.  I also removed a method that seems to be dead code.: SessionManager>>initializedSessionManager


jecisc:
This timed out.. Now I need to find why D=

jecisc:
Close in favor of:  #13920

PR #13442, opened 2023-04-14
Add logs to locate bug in protocols

jecisc:
Ideally, all methods of a class should always be in a protocol of the class, even if it is the unclassified protocol  Sadly, some previous bootstrap crashes have shown that reality is harsh and this is not the case. This is why ClassOrganization>>#protocolNameOfElement:ifAbsent: comes with a hack that prevents the users to just use #protocolOfSelector: directly.

Let's add some logs to see if we can find the culprit and remove this ugly hack.


jecisc:
No wonder my other PR crashed. This happens a looooooot 😱

Now I will need to hunt for the place that creates methods without setting the protocol in ClassOrganization :(

jecisc:
Hum... This happens when we ask the protocol of a method that is in a superclass but not directly in the class we are manipulating I have the impression. 

I am not sure we want to do that. 

Instead of implementing it on ClassOrganization it should be on ClassDescription and if the method is in a superclass, it should ask it the protocol instead of returning Protocol unclassified IMO.

jecisc:
I located some sources of the problem so I'll close this one for now

PR #13443, opened 2023-04-14
Keep pushing protocols usage

jecisc:
The goal is to keep pushing the fact that we can use Protocols directly as parameter of ClassOrganizer methods and reduce the API size in order to move this in ClassDescription later.

#addProtocol: and #addProtocolNamed: are now merged and #hasProtocolNamed: become #hasProtocol:


jecisc:
Failures seems unrelated

privat (approved):

MarcusDenker (approved):

Issue #13447, opened 2023-04-14
DateAndTime is way less precise on Windows than on Unix


jecisc:
On windows if you do `DateAndTime now = DateAndTime now` you get true because you need to wait at least 1 millisecond to get two different DateAndTime instances from #now. On unix you get false because they have a precision at the nanosecond.

This is problematic because first the API let you think you can manipulate this at the level of the nano second. And second because DateAndTime now is used to generate unique IDs in the system and that means that if we create ID too fast on windows, we end up with same IDs.

svenvc:
That is why, for example, UUIDGenerator also uses a counter and a random element.

You should also check if the file exists.

Like Stef suggested, we already have FileReference class>>#newTempFilePrefix:suffix:

labordep:
Hi, we have this problem in our UI library (Toplo), our tests on Windows (Pharo 11 and Pharo 12) are not passing.
To test, DoIt on this in a Windows playground and a Mac playground : 
`DateAndTime now < DateAndTime now`
The result in Windows is **false**, in Mac it is **true**.

For a unit test we need to write `aTimestamp <= anotherTimestamp` instead of `aTimestamp < anotherTimestamp`. This is dirty :(

Another precision, in Windows I have already 6 zeros digits in my `DateAndTime now` nanos value and 3 zeros digits on Mac. So the result of `DateAndTime <` is not the same because of :
![image](https://github.com/pharo-project/pharo/assets/49183340/adf5f55e-a980-4b0f-91df-bb535c5f3b69)

guillep:
The code that computes that is here: https://github.com/pharo-project/pharo-vm/blob/de760673b5f108d32576b47f4bbdfd3b230520b0/extracted/vm/src/common/heartbeat.c#L113
There are different implementations between windows and *nix based systems.

In *nix systems we are using gettimeofday, maybe we should move to clock_gettime as proposed by this post https://stackoverflow.com/questions/69099750/what-is-the-precision-of-the-gettimeofday-function?

The equivalent of clock_gettime in windows is no trivial, but we could port such a solution :)
https://stackoverflow.com/questions/5404277/porting-clock-gettime-to-windows

PR #13446, opened 2023-04-14
Fix typo introduced in merge

jecisc:
During a recent merge I introduced a typo breaking the code highlighting. This fixes it

PR #13448, opened 2023-04-14
Fix random failure of Ombu in windows.

jecisc:
The problems comes from the fact that we use the current time to set the name of Ombu stores.  But here we create 2 stores in less than 1 millisecond and DateAndTime on windows is sadly not able to give us a precision under 1 millisecond for now... See https://github.com/pharo-project/pharo/issues/13447.

So here is a quick hack to wait 1 millisecond before producing a new name on windows.

This was really not easy to locate :(

### DIFF
--- a/src/Ombu-Tests/OmSessionStoreTest.class.st
+++ b/src/Ombu-Tests/OmSessionStoreTest.class.st
@@ -27,6 +27,16 @@ OmSessionStoreTest >> tearDown [
 ]
 
 { #category : #tests }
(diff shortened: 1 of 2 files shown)


jecisc:
CI is bad because of https://github.com/pharo-project/pharo/pull/13446

Once it is integrated I'll restart it

tinchodias:
Ohh it was that! good catch.

MarcusDenker (approved):

privat (approved):

PR #13449, opened 2023-04-14
Introduce annotations

privat:
Playing with the compiler, AST transformations, IR manipulations and language extensions is fun, but sometimes you wish having some sane way to isolate your experimentation attempts from the compiler code base.

Currently, OpalCompiler can be extended with plugins (two set of them, one for the frontend and one for the backend).
It is used by the FFI for instance (I don't know the detail that much).
Another approach is to subclass some core classes of the compiler and use them instead of the vanilla ones. This is what reflectivity/metalinks does, but has a great cost: a strong binding to the compiler architecture (with a lot of fragile base class issues). Moreover, multiple users of such a radical approach cannot be easily combined.
On the parser front, there is no plugin available. Subclassing is always an option (it is what the pattern parser does) but has drawbacks. Alternative approaches like extensible grammars are not available in Pharo, but are always a nightmare anyway.

This PR is a proposal to be able to experiment on independent and integrated extensions of OpalCompiler that include the parser and the IR generation. It is based on the existing plugin infrastructure.

On the parser front, there is a single extension point: a new *annotation* syntax based on a pseudo-message syntax (like pragma) with a selector and possible arguments.
The use of selectors (instead of a soup of random characters) force each extension to be named. They can be searched (like pragmas). They are also more understandable by random users, which might guess what such an extension does by its name.
To distinguish from normal selectors, there are two things: they have no receiver (like pragma), and they start with `@`. This symbol was chosen because it's the one used by some other languages (python and java for instance) for their own annotations.
Besides that, they can be used anywhere a message send is expected (with the same parsing precedence), and with the same kind of arguments (including other annotations).

```st
@foo. "a unary annotation `@foo`"
@foo: x bar: y. "a keyword annotation `@foo:bar:`"
^ x := y foo: (@bar: z + 1). "a keyword annotation `@bar:` in an expression that takes an expression"
```

About language theory. There is no conflict nor ambiguity. `@` in annotations was previously parsed as a syntax error (binary operator with a missing receiver).
For the AST representation, there is a simple (and nice) trick: annotations are genuine message send nodes, but with a new AST node `RBAnnotationMarkNode` (that corresponds to a lone `@`) used as a receiver.

On the processing front.
By default, annotations are simply rejected by the compiler (*unknown annotation*).
In order to manipulate them, plugins with the responsibility to handle them must be attached to the compiler.
They can do classic AST search and transformation to handle the annotations they know about.
Node replacements are an effective approach since they make the annotation disappear (so the compiler won't signal an error).

AST transformations are pleasant, but what if we want to play also with the IR?
`RBAnnotationMarkNode` are usable by the AST→IR transformer to emit IR code thanks to `emitValue:`.
This idea is similar to that is done for variables with their `emitValue:` and `emitStore:` methods.
Therefore, plugins can replace instances of `RBAnnotationMarkNode` with instances of their own subclasses and provide meaningful `emitValue:` implementations.

An alternative lightweight approach is to use `emitValueBlock:` to assign a custom IR generation block to annotation instances in the AST. In this case, no need to have a subclass, and plugins can simply use the existing instances.

Fun fact: new `RBAnnotationMarkNode` instances can also be injected in AST by some plugin where they were not present in order to use their `emitValue:` facility.

And it's basically all. Most of the needed code is already available. The PR is in fact very small, tests and examples takes the major part of it.
By the way, you might want to look at the examples (and, by the law of lexicographic order, they are at the end of the diff).
There are two examples that introduces a toy `@meaning:` annotation to show basic usage (one with AST transformation, the other one with IR emission).
There are more realistic examples. One proposes `@binding:` to get the binding of a variable (so you get a `Variable` instance at runtime, instead of the value of the variable). The last twos propose `@constexpr:` to evaluate an expression at compile-time (one example with AST transformation, the other with IR emission).

Limitation and future work:

* The AST injector point (annotation node) is limited to values (aka expressions). This is not a big limitation because most things are expressions, and for most things that are not expressions, you can enclose them in a block.
* The plugin API can be rough with nested classes and blocks. But "production" code should be able to use block-less approaches with classes and correctly assigned responsibilities.
* The annotation API is very simple (by design) but it might not be its final form.

### DIFF
--- a/src/AST-Core-Tests/RBCodeSnippet.class.st
+++ b/src/AST-Core-Tests/RBCodeSnippet.class.st
@@ -47,6 +47,7 @@ Class {
 RBCodeSnippet class >> allSnippets [
 
 	^ {
+		  self badAnnotations.
 		  self badSimpleExpressions.
 		  self badExpressions.
 		  self badTokens.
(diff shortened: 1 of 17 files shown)


privat:
Tests passed (modulo #13446)

MarcusDenker (approved):

MarcusDenker:
Let's experiment with that. I really like the idea
- backward compatible from the grammer POV
- the possibility to do changes in the IR level is very nice

noha:
This looks almost like a change that should have a PhEP. Or say it in other words, I wonder which changes should have a PhEP and which don't need that

tinchodias:
Hi, I didn't get use scenarios of these annotations... do you have some example or demo?

PR #13454, opened 2023-04-16
Cleaning of #classify:inProtocolNamed:

jecisc:
I updated #classify:inProtocolNamed: in multiple ways:

- Now the method can take a protocol name OR a real protocol as parameter
- If the method is been reclassified and the old protocol becomes empty, we remove it. In the past this was only done if this method was invoked through #classify:under:
- In the end, this method was doing exactly the same as #classify:under: but without the reclassification announcement. Since I had to rename the method to remove the "inProtocolNamed:" I decided to rename it #silentlyClassify:under: in mirror of #classify:under:
- Now this method has tests


MarcusDenker (approved):

MarcusDenker:
There is now a merge conflict

jecisc:
Fixed!

jecisc:
Test failure is not releated. It should be linked to latest Iceberg changes.

MarcusDenker (approved):

PR #13455, opened 2023-04-16
Introduce ClassDescription>>protocolOfSelector:

jecisc:
In the current system we have ClassOrganization>>protocolNameOfElement:ifAbsent:. that includes a hack when a method is present in the class but is not in its protocols (and some logs I added on the CI suggest that this is used a lot to get the protocol of a method in a superclass). In that case it returns Protocol unclassified.

I wan to remove totally the use of the method anyway to just manipulate protocols and not protocol names. If someone need the name they can just call #name of the protocol.

In order to do two birds one stone, I suggest the addtion of #protocolOfSelector on class description directly. This method would return the protocol of the method even if it is in a super class. I might update this PR to start replacing the users of #protocolNameOfElement:ifAbsent: by this one

### DIFF
--- a/src/Kernel-Tests/ClassDescriptionTest.class.st
+++ b/src/Kernel-Tests/ClassDescriptionTest.class.st
@@ -109,6 +109,25 @@ ClassDescriptionTest >> testOrganization [
 	self assert: (aClassOrganizer isKindOf: ClassOrganization)
 ]
 
+{ #category : #tests }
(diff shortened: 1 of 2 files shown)


jecisc:
For this one I would like to discuss with Marcus because I might break the coherence of the system.

jecisc:
After checking how most other methods of ClassDescription I think I might break the consistency with this method so I will close this issue for now and attack the problem with another angle.

PR #13456, opened 2023-04-16
Adding some comments to Protocol

jecisc:
Still trying to find the problematic change of #13305


jecisc:
We have our winner for the modification that crashes the CI and it is just a change of comment

jordanmontt (approved):

jordanmontt:
CI failure: `Warning: Package Kernel-Processes depends on the following classes:`

No tests were executed, seems like a random failure?

jecisc:
The bootstrap crashed at the loading step and this is not random. This happens all the time with this change. But I have no idea what happens because, as you can see, the change is not anything special...

jecisc:
With the recent cleaning the error is still present but it changed

jecisc:
It seems that does the same hack for Context and CompiledBlock does not work totally. Some tests are failing and seems related.

MarcusDenker:
Just two test failing. Interestingly, they fails because the timestamp of some methods in Point is the string 'nil'


RGReadOnlyImageBackendTest>>testMethod
here it is Point >> #x

SourceFileArrayTest>>testTimeStamp
Point>>#dotProduct:


The stamp looks lile this (for many methods in Point):

 "'Point methodsFor: ''point functions'' stamp: ''nil'''"

MarcusDenker (approved):

MarcusDenker:
Yes, we need to check why it is 'nil', but being 'nil' is in practice no problem (same as empty).

I merge and we can move forward, as this is blocking lots of improvements

PR #13457, opened 2023-04-16
Use #protocolOfSelector: in CompiledMethod>>#category

jecisc:
I thought the method was storing its protocol but in fact it is asking the class for its protocol name using #protocolNameOfElement:

I want to remove this method because in case the method is not present in the class, it returns Protocol unclassified. For me, if the method is not here, you return nil or an error. I'm going with the nil to simplify the transition currently. 

Since the change impacts a lot of places in the system, I will try to replace it little by little to see who was relying on the hack. First place is CompiledMethod>>#category. Let's see if the CI agrees


jecisc:
In the futur I will also make this method return a protocol later

jecisc:
The current problem is that #classify:under: does not manage the case where the protocol is nil.

My plan is to change that because currently it uses Protocol unclassified to say "do not categorize" and it's in my plans to change that so that we can use nil instead.

But since https://github.com/pharo-project/pharo/pull/13373 changes a lot the way to classify, I'll need this PR integrated first. So I'll pause this PR until the other one is fixed and integrated.

jecisc:
This one is passing now with recent improvements :)

MarcusDenker (approved):

PR #13458, opened 2023-04-17
Remove ClassOrganization>>#silentlyRenameProtocolNamed:toBe: in favor of always announcing the rename

jecisc:
When we rename a protocol we announce the change in the image so that some tools like Epicea can manage this change.  Currently when renaming a RPackage, the extension renaming is not announced because it uses a silent rename.

I'm trying to avoid silent actions with protocol so let's see if anything breaks if we announce the change.

Since Epicea is one of the main consumer of the meta model change announcements, I added some missing tests to check that my changes to not impact Epicea's behavior.

### DIFF
--- a/src/Deprecated12/ClassOrganization.extension.st
+++ b/src/Deprecated12/ClassOrganization.extension.st
@@ -105,7 +105,7 @@ ClassOrganization >> renameCategory: oldName toBe: newName [
 ClassOrganization >> silentlyRenameCategory: oldName toBe: newName [
 
 	self
-		deprecated: 'Use #renameProtocolNamed:toBe: instead.'
+		deprecated: 'Use #renameProtocolNamed:toBe: instead. Since Pharo12 it is not possible to not announce a protocol rename anymore because we should always announce those changes.'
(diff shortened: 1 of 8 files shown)


src/Deprecated12/ClassOrganization.extension.st
```diff
@@ -105,7 +105,7 @@ ClassOrganization >> renameCategory: oldName toBe: newName [
 ClassOrganization >> silentlyRenameCategory: oldName toBe: newName [
 
 	self
-		deprecated: 'Use #silentlyRenameProtocolNamed:toBe: instead.'
-		transformWith: '`@rcv silentlyRenameCategory: `@arg1 toBe: `@arg2' -> '`@rcv silentlyRenameProtocolNamed: `@arg1 toBe: `@arg2'.
-	self silentlyRenameProtocolNamed: oldName toBe: newName
+		deprecated: 'Use #renameProtocolNamed:toBe: instead.'
```
jordanmontt: Maybe a little comment to this method to explain why a user should not use `silentlyRenameProtocolNamed:`?

jordanmontt (approved):

src/Deprecated12/ClassOrganization.extension.st
jecisc: Done!

MarcusDenker (approved):

Issue #13070, opened 2023-03-20
[CI] Linux: OSEnvironmentTest fail with "Illegal continuation byte for utf-8 encoding"


MarcusDenker:
Linux: OSEnvironmentTest fail with "Illegal continuation byte for utf-8 encoding"

For some user submitting PRs, the CI fails on Linux:

```
unix-64 / Tests-unix-64 / testAsDictionary – Unix64.System.OSEnvironments.Tests.OSEnvironmentTest
unix-64 / Tests-unix-64 / testAssociations – Unix64.System.OSEnvironments.Tests.OSEnvironmentTest
unix-64 / Tests-unix-64 / testKeys – Unix64.System.OSEnvironments.Tests.OSEnvironmentTest
unix-64 / Tests-unix-64 / testValues – Unix64.System.OSEnvironments.Tests.OSEnvironmentTest
```

for example:

```
Error
Illegal continuation byte for utf-8 encoding
Stacktrace
ZnInvalidUTF8
Illegal continuation byte for utf-8 encoding
ZnUTF8Encoder>>error:
ZnUTF8Encoder>>errorIllegalContinuationByte
ZnUTF8Encoder>>nextCodePointFromStream:
[ :stream |
		[ byteStream atEnd ] whileFalse: [ | codePoint |
			codePoint := self nextCodePointFromStream: byteStream.
			(codePoint > 255 and: [ stream originalContents isWideString not ])
				ifTrue: [ | wideString position |
					position := stream position.
					wideString := WideString from: stream originalContents.
					stream on: wideString; setFrom: position + 1 to: position ].
			stream nextPut: (Character value: codePoint) ] ] in ZnUTF8Encoder(ZnUTFEncoder)>>decodeBytes:
String class(SequenceableCollection class)>>new:streamContents:
String class(SequenceableCollection class)>>streamContents:
ZnUTF8Encoder(ZnUTFEncoder)>>decodeBytes:
ByteArray>>decodeWith:
ByteArray>>utf8Decoded
FFIExternalString>>basicHandle:at:
FFIExternalString(FFIExternalType)>>handle:at:
FFIExternalArray>>at:
UnixEnvironment>>environAt:
UnixEnvironment>>keysAndValuesDo:
UnixEnvironment(OSEnvironment)>>valuesDo:
[ :stream|
		self valuesDo: [ :value | stream nextPut: value ]] in UnixEnvironment(OSEnvironment)>>values
[...]

PR #13459, opened 2023-04-17
Refactoring BlockClosureTest >> testCannotReturn

jordanmontt:
Refactoring the test `BlockClosureTest >> testCannotReturn`. Previously, it has a never reached assertion `self assert: false` to fail in case that the `block value` returns correctly. The intention of the test is to ensure that `block value` produces an error. Now I'm testing the same behavior but with `should: raise: withExceptionDo:` which seems more adequate.


jecisc (approved):
Way better!

jecisc:
[CI] Linux: OSEnvironmentTest fail with "Illegal continuation byte for utf-8 encoding" #13070

PR #13466, opened 2023-04-18
Improve #renameProtocolNamed:toBe:

jecisc:
In the continuity of the current changes I'm making, here is a cleaning of the protocol renaming.

Changes:
- Now we have way more tests
- You can provide Protocol instead of protocol names
- You can provide nils and it will cancel the renaming (this will be useful to remove the hacks around `Protocol unclassified`. Maybe it should raise an error instead but this is out of the scope of this PR)
- It will also make it easier later to update all the announcements to hold an instance of Protocol instead of the protocol names


jecisc:
I forgot to add that it also fixes a bug where a protocol creation was not announced in the system. 

If the new protocol did not exist, it was created silently. Now we announce it.

MarcusDenker:
SystemDependenciesTest is failing

jecisc:
I don't find the origin pf the issue. Locally it works. 

I night try to split this one into multiple PRs

jecisc:
So... I had to keep the hack that do not announce all protocol creation and removal in the end because RPackage bug when listening to those announcements.

Once it is fix, we can simplify this further, but it's not easy to find a good fix for RPackage with how complex it is right now.

jecisc:
Tests passed now. 

It has a hack I want to remove later but it will help us progress at least. This is ready to be reviewed.

privat (commented):
I'm not a native speaker, but `rename:into:` sounds weird. What about `rename:as:` (or `rename:to:`)?

jecisc:
I asked some friends from the US and they said they'd use "as:". I'll update the PR accordingly

jecisc:
Done!

MarcusDenker (approved):

jecisc:
Failures seems unrelated

PR #13470, opened 2023-04-18
Factorize UndeclaredVariableErrors

privat:
Create a new superclass UndeclaredVariableError for UndeclaredVariableRead and UndeclaredVariableWrite.
Also add `variableNode` that will be used by StDebugger to create a new class on an undeclared variable


jecisc (approved):
Good!

MarcusDenker (approved):

PR #13472, opened 2023-04-18
Compiler: explicit the need of requestor scope

privat:
Require https://github.com/pharo-spec/Spec/pull/1377

This is the remaining part of  #13323 that was missing in #13393

Requestors that need special bindings or the ability to declare new bindings by themselves must announce it.
Because there is no requestor API, introspection is used. Previously hard-coded class names are removed.

Note: I renamed `canDeclareVariableName:` to `canAddBindingOf:` to be more consistent with existing `*binding*` selectors.

### DIFF
--- a/src/OpalCompiler-Core/CompilationContext.class.st
+++ b/src/OpalCompiler-Core/CompilationContext.class.st
@@ -501,6 +501,17 @@ CompilationContext >> interactive [
 		  ifFalse: [ true ]
 ]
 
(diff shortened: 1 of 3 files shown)


privat:
Tests did not fail, although https://github.com/pharo-spec/Spec/pull/1377 is not merged.
Some tests might be lacking in the spec world :/

MarcusDenker (approved):

PR #13473, opened 2023-04-19
Completion menu: some love on the detail window

privat:
* Details in completion menu showed a bunch of `'CompiledMethod'` instead of the correct class names
* Forced reset after unhandled keyboard event makes control commands unusable. So disable it.
* Open the detail window by default. What is the point of hiding it and making it very hard to discover?
* Provide correct detail information for completion when continuing a keyword selector with many arguments.


privat:
In can split in smaller PR if reviewers prefer

jecisc (approved):
I tried it and it works fine.

jecisc:
`Provide correct detail information for completion when continuing a keyword selector with many arguments.` ==> This one I wanted it for years <3

PR #13479, opened 2023-04-19
Compiler: remove #optionEmbeddSources 

MarcusDenker:
this PR removes the #optionEmbeddSources

The whole CompiledMethodTrailer concept is far too complex... this PR is a small step to allow to simplify alter. The compiler now sets the #source property for all  methods ot creates, thus we do not need the option to embedd in the Trailer on the level of the compiler This PR

- removes the  #optionEmbeddSources 
- reduces the users of #generateWithSource

The next step will be to remove generateWithSource and simplify CompiledMethodTrailer

### DIFF
--- a/src/OpalCompiler-Core/CompilationContext.class.st
+++ b/src/OpalCompiler-Core/CompilationContext.class.st
@@ -156,16 +156,6 @@ CompilationContext class >> optionConstantBlockClosure: aBoolean [
 ]
 
 { #category : #'options - settings API' }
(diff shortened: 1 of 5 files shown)


jecisc (approved):
I am wondering, what happens if someone compiles with this option?

MarcusDenker:
you can just add any kind of option, they are just ignored.

The compiler will always add the #source property, so the user would even get a method with source...

privat (approved):

PR #13483, opened 2023-04-19
use {} instead of Array with:with:

MarcusDenker:
Another trivial PR using {} instead of with:with: of Array

### DIFF
--- a/src/Collections-Strings/String.class.st
+++ b/src/Collections-Strings/String.class.st
@@ -2519,7 +2519,7 @@ String >> stemAndNumericSuffix [
 		[suffix :=  stem last digitValue * position + suffix.
 		position := position * 10.
 		stem := stem copyFrom: 1 to: stem size - 1].
-	^ Array with: stem with: suffix
+	^ { stem . suffix }
 ]
 
 { #category : #printing }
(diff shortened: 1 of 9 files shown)


privat (approved):
4 more PR to go? The longest of the family seems `ArrayedCollection class>>#with:with:with:with:with:with:` :)

jordanmontt (approved):

astares:
YES, there is a benefit:
 - the { } notation is shorter to write and also quicker to read
 - you often do not need to put it into extra braces (....) as you already have the { ... }
 
BUT:

As the with:with: ... messages are limited **they really reveal that you have a design problem** 
in trying to give more and more parameters until you reach limits
like #with:with:with:with:with:with: and then **it lets you stop to start thinking** if you really need so many arguments.
   
So the #with:with: notations forces one to think about a **better design then to provide a real object 
to be passed** when more parameters need to be given.

While { } notation on th other side does not have such a limit and is not revealing the design problem. 
So basically you can pass 20, 40, 200 and more elements which would be less optimal in 
reading and can foster a non-object oriented design with "passing large arrays with different elements around".

PR #13486, opened 2023-04-19
Faulty code: improve highlight and icons

privat:
Summary:

* correctly highlight opener/closer on error node with missing opener/closer or problematic content
* icons don't crash on bad byte literal
* ensure error position is marked (missing token, unexpected end of file)
* ensure 0-with underlining are visible (missing token, unexpected end of file)

```st
foo
{
#[ 1 256 2 ].
+ 10.
max:10.
10 max:.
10 10.
10).

```

Calypso screenshot (include icons and text segmentd):

Capture d’écran du 2023-04-19 13-44-01

Playground screenshot (no icon or text segments):

Capture d’écran du 2023-04-19 13-45-50


jecisc:
It might be nice to add a test in SHRBStyleAttributionTest

Except that the code seems good

MarcusDenker (approved):

PR #13488, opened 2023-04-19
Use Compiler not #copyWithTrailerBytes

MarcusDenker:
- Context>>#initializeTryNamedPrimitiveTemplateMethod can just use the compiler, no #copyWithSource: needed
- testPrintOnBlockDefinedInMethodWithoutSourceCode can use the compiler
- remove testInstalledMethodsWithIncorrectTrailerKind,  a step to remove all other trailer kinds
- add CompiledMethod>>valueWithReceiver: and test

### DIFF
--- a/src/Kernel/CompiledMethod.class.st
+++ b/src/Kernel/CompiledMethod.class.st
@@ -1155,6 +1155,11 @@ CompiledMethod >> usesUndeclareds [
 ]
 
 { #category : #evaluating }
+CompiledMethod >> valueWithReceiver: aReceiver [
+	^self receiver: aReceiver withArguments: #() executeMethod: self
+]
+
+{ #category : #evaluating }
 CompiledMethod >> valueWithReceiver: aReceiver arguments: anArray [
 	^self receiver: aReceiver withArguments: anArray executeMethod: self
 ]
--- a/src/Kernel/CompiledMethod.class.st
+++ b/src/Kernel/CompiledMethod.class.st
@@ -938,7 +938,13 @@ CompiledMethod >> readsField: varIndex [
 ]
 
 { #category : #evaluating }
-CompiledMethod >> receiver: aReceiver withArguments: anArray executeMethod: anObject [
+CompiledMethod >> receiver: aReceiver withArguments: anArray executeMethod: aMethod [
+	"execute aMethod.
+	This method takes aMethod as an argument so we can call primitive 188.
+	Clients should use #valueWithReceiver:arguments:
+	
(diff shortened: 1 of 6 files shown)


src/Kernel/CompiledMethod.class.st
```diff
@@ -1154,6 +1154,11 @@ CompiledMethod >> usesUndeclareds [
 	^ false
 ]
 
+{ #category : #evaluating }
+CompiledMethod >> valueWithReceiver: aReceiver [
+	^self receiver: aReceiver withArguments: #() executeMethod: self
```
privat, line 1165: Aren't too much synonym for the same thing?
In fact, I did not even realize that `CompiledMethod>>#receiver:withArguments:executeMethod:` existed.
(and why the compiled method is both a receiver and an argument here?)

Up to now, I used `ProtoObject>>#executeMethod:`. That one uses `ProtoObject>>#withArgs:executeMethod:` that is the same primitive number 188. Weird.

src/Kernel/Context.class.st
```diff
@@ -156,17 +156,16 @@ Context class >> initializePrimitiveSimulators [
...
 	<primitive:'' module:'' error: errorCode>
 	^ Context primitiveFailTokenFor: errorCode'.
-	method := Smalltalk compiler
+	TryNamedPrimitiveTemplateMethod := Smalltalk compiler
 				class: UndefinedObject;
 				source: source;
-				compile.
-	TryNamedPrimitiveTemplateMethod := method copyWithSource: source
+				compile
 ]
```
privat, line 166: Could be even simpler, I think. `UndefinedObject` is the default

```st
Smalltalk compiler compile: source
```

src/Kernel/CompiledMethod.class.st
MarcusDenker: CompiledMethod>>#receiver:withArguments:executeMethod:  takes a bit odd parameters so it can use Primitive 188. We have it on CompiledMethod so we can execute the method without sending a message to an object (which we would need to instantiate, and we might not want to send a message e.g. for proxies, so having that API is nice).

I wanted to add #valueWithReceiver:  for all the (many!) users that execute methods that have no args. A bit like we have #value on block.

src/Kernel/Context.class.st
MarcusDenker: fixed

src/Kernel/CompiledMethod.class.st
MarcusDenker: I added a comment in #receiver:withArguments:executeMethod:

privat (approved):

PR #13491, opened 2023-04-20
Improve style for parenthesis, comments and syntax error

privat:
This is a continuation of #13490. So merge it first if you want a better global diff.
In the meantime, you might look at individual commits.

* Always style parenthesis. It's done globally at `visitNode` instead of at some individual selected node.
* Style parenthesis in the right order when nested, eg `( ( a ) + ( ( b ) ) )` (it's a detail)
* Always style comments. In fact, it was already the case and individual call at `visitCommentNodes:` where just superfluous. Thus, the method is removed.
* Ensure that syntax error positions are always marked invalid. Because the position can be outside the node and overflows onto other node, it should be done last.
  Ideally, if there is a code error in the source, there should always be some red somewhere (it's not completely true yet)

### DIFF
--- a/src/Shout/SHRBTextStyler.class.st
+++ b/src/Shout/SHRBTextStyler.class.st
@@ -1050,15 +1050,16 @@ SHRBTextStyler >> visitMethodNode: aMethodNode [
 	self visitMethodComments: aMethodNode.
 	aMethodNode arguments do: [ :argument | self addStyle: #patternArg forNode: argument ].
 	link := TextMethodLink selector: aMethodNode selector.
-	aMethodNode selectorParts
-		with: aMethodNode keywordsPositions
-		do:
-			[ :keyword :position |
-			self
-				addStyle: #patternSelector
-				attribute: link
-				from: position
-				to: position + keyword size - 1 ].
+	aMethodNode isDoIt ifFalse: [
+		aMethodNode selectorParts
+			with: aMethodNode keywordsPositions
+			do:
+				[ :keyword :position |
+				self
+					addStyle: #patternSelector
+					attribute: link
+					from: position
+					to: position + keyword size - 1 ] ].
 	aMethodNode pragmas do: [ :each | self visitNode: each ].
 	self visitNode: aMethodNode body
 ]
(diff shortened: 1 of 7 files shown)


src/Shout/SHRBTextStyler.class.st
```diff
@@ -836,6 +836,10 @@ SHRBTextStyler >> style: aText ast: ast [
 	parentheseLevel := 0.
 	braceLevel := 0.
 	self visitNode: ast.
+	"Second pass to ensure that all syntax errors are visible"
+	ast allErrorNotices do: [ :notice | 
+		notice isSyntaxError ifTrue: [
+			self addStyle: #invalid from: notice position to: notice position ] ].
```
jecisc, line 842: With this, do we need to style the other invalid positions or could this deal with all invalid things at once?
For example can we remove visitParseErrorNode: or not?

jecisc (approved):

src/Shout/SHRBTextStyler.class.st
privat: This one adds red to a single character that might be outside the span of the node. Especially for 0-width errors where all positions are necessary outside.
It is also used by a few bad number tokens that have a span and a error position outside.
The visitParseErrorNode take care of styling the full node, not only the error position.

I didn't change that much the way Pharo report errors. I changed most of the API and the implementation but I did not touch a lot the error messages or their positions.
Some might feel weird because they are mechanical errors "I did not expect this character now in the stream" where the error position is where you are in the stream and the error is reported with `blabla->` pointing the unwanted character and stopping the parsing.

The parser does not work this way anymore and always continue in recovery mode to give you an AST, so the unwanted character might be consumed in the error node, but it might instead be used for something else meaningful in the AST. Leaving the error character outside the error node.
The question is therefore how to report such error visually, especially if the error node is empty?
I simply choose to force the highlighting of the character at the error position, because whatever the character might be reinserted to, it is still a culprit for the parsing law.

Note: This is consistent with how the styler worked previously with ad hoc heuristic (for instance it considered that the error position is always the first character that follows its span). I just make it correct (use real information available in the AST, not heuristics) and systematic.

PR #13494, opened 2023-04-20
Ensure renaming of protocol announce the changes

jecisc:
When renaming a protocol, if the new protocol does not exist already, the renaming is done by changing the name of the Protocol.  This is efficient in term of perf but it has the drawback that the new protocol creation and the old protocol removal are not annouced in that case. 

I propose to unify the cases where the new protocol exists or not to always announce everything.

Future step: 
Once this is done, we can remove the ProtocolRename announcement because nobody uses it


jecisc:
So this is the change that crashes the other PR. Now I need to find why...

jecisc:
It seems the system is left in a bad state after executing EpApplyTest>>#testCategoryRenameWithExtension with this change.

jecisc:
With my recent cleanings in RPackage and Monticello this now works!

MarcusDenker (approved):

PR #13497, opened 2023-04-20
Deprecate SystemOrganization

jecisc:
We have 3 ways to access the SystemOrganization currently:
- A global variable `SystemOrganization`
- A direct access to `SystemOrganizer default`
- Asking to SmalltalkImage like `Smalltalk organization`

Some even do `Smalltalk globals organization`!

Let's stop with all those ways to access it and deprecate the global variable as a first step


jecisc:
Apparently Espell does not like that :(

jecisc:
This PR shows that Pharo's globals are hardcoded in 7 places in the system/bootstrap

It might be nice to find a way to reduce this number.

PR #13498, opened 2023-04-20
OpalCompiler: permit clients to use receiver: after context: as (a noop)

privat:
Some clients (e.g. Spec/Newtools) use both `receiver:` and `context:` when setting up a compiler.
This PR implements the previous behavior where the context is not lost if a redundant receiver is set.

This fixes an issue where DoIt evaluation of temps in the debugger is currently broken


jecisc (approved):
I tried it and it works

PR #13501, opened 2023-04-20
SHRBTextStyler class>>#blueStyleTable distinguish some variables

privat:
Update the default blue style of SHRBTextStyler to distinguish some category of variables.
I did not use too many colors here to avoid adding too much noise.

* local variables (including temps, args and block args) remain blue
* receiver/class based variables (ivars and shareds) become cyan (it is the color of self)
* global variables become purple (that color is not used in the default blue theme)

Capture d’écran du 2023-04-20 16-09-26


jecisc (approved):
At some point I should also do as in my own theme for all kind of variables: 
image

MarcusDenker (approved):

PR #13507, opened 2023-04-21
CompiledMethod simplification: remove the last user of #copyWithSource:

MarcusDenker:
The idea is that we want to radically simplify the CompiledMethodTrailer. One step is to not support storing source in the trailer, but to instead put it as a methodProperty. This PR changes the "onFail:" when storing sources to log and store the sourse as  a property.

### DIFF
--- a/src/System-Sources/CompiledMethod.extension.st
+++ b/src/System-Sources/CompiledMethod.extension.st
@@ -25,5 +25,6 @@ CompiledMethod >> putSource: source withPreamble: preambleBlock [
 			"Update with new source pointer"
 			self setSourcePointer: newSourcePointer ]
 		onFail: [
-			self becomeForward: (self copyWithSource: source) ]
+			self traceCr: 'Warning, could not write sources'.
+			self propertyAt: #source put: source ]
 ]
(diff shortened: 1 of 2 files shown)


src/System-Sources/CompiledMethod.extension.st
```diff
@@ -25,5 +25,6 @@ CompiledMethod >> putSource: source withPreamble: preambleBlock [
 			"Update with new source pointer"
 			self setSourcePointer: newSourcePointer ]
 		onFail: [
-			self becomeForward: (self copyWithSource: source) ]
+			self traceCr: 'Warning, could not write sources'.
```
jecisc: I'm not sure #traceCr: is part of the kernel so if this gets called during the beginning of the bootstrap it might crash. Maybe there should be a #respondsTo: if that can happen?

MarcusDenker: But how can we log? logging is quite essential... else we never know what happens.

MarcusDenker: I removed it

jecisc: We don't have the transcript but we have the stdio. 

Maybe we could find a way to use Transcript if present to log, and if it's not present log on Stdout?

jecisc: Hum... We do have the NonInteractiveTranscript. Maybe we could move the #traceCr: & co methods to this package so that we can log in the beginning of the bootstrap?

jecisc (approved):

privat (approved):

PR #13510, opened 2023-04-21
remove all unsused encodings from CompiledMethodTrailer.

MarcusDenker:
This PR removes all unused encoding from CompiledMethodTrailer.

In the end, CompiledMethodTrailer  is amazing how much complexity it adds...


jecisc (approved):
Seen the complexity of the removed code, this will be great to reduce the system complexity!

Issue #13508, opened 2023-04-21
[Meta issue] detaching AST and CompilationContext?


privat:
AST, and especially the root node `RBMethodNode` have a lot of responsibilities. As a grammatical representation of the source code, obviously, but also to manage its own compilation.

**Update:** checked boxes are resolved

Currently, some compilation concerns are not clearly assigned between `RBMethodNode` and `OpalCompiler`
Methods of `RBMethodNode` with the main compilation (or related) responsibility (simple wrappers excluded) are:

* [ ] bcToASTCache
* [x] compiledMethod #13514
* [x] doSemanticAnalysis #13517
* [ ] generate:
* [ ] generateIR
* [ ] generateMethod
* [x] method #13514
* [ ] scope

This means that the ASTs have to store a lot of data through the god-data-class CompilationContext (that includes potentially GUI windows). Beside storing, this data is exposed and used by the following methods that rely on the existence of a compilation context in a method node:

* [ ] `CDClassNameNode>>#existingBindingIfAbsent:` to get the environment of the scope of the node (why?)
* [x] `CompiledMethod>>#parseTree` to store the compiled method in the context (who use that?) #13514
* [x] `OCASTSemanticAnalyzer>>#visitPragmaNode:` to set compilerOptions: given with pragma  #13511
* [x] `OCASTTranslator>>#translateFullBlock:` to initialize the compilationContext of the methodBuilder ivar (maybe the ivar should be in the translator)  #13511
* [x] `OCASTTranslator>>#visitMethodNode:` same  #13511
* [x] `RFASTTranslator>>#visitMethodNode:` same  #13511
* [ ] all `RBMessageNode>>#isInline...` to check the `optionInline...` flag of the compilationContext and decide if the expression should be inlined
* [x] `RBProgramNode>>#asDoItForContext:` only one client #13520
* [x] `RBMethodNode>>#method` because the compiled method might be stored in the compilation context #13514
* [ ] `RBMethodNode>>#methodClass` this one is surprising, a method node does not know its own class but need a full context and scope
* [ ] `RBMethodNode>>#methodClass:` this one updates or creates a compilationContext. Some clients might not expect this side effect
* [ ] `RBMethodNode>>#semanticScope:` this one also update or create a compilationContext
* [x] `RBValueNode>>#evaluateForContext:` only used in a single test #13520
* [x] `RBValueNode>>#evaluateForReceiver:` #13520
[...]

PR #13514, opened 2023-04-21
AST: cleanup extension methods from OpalCompiler

privat:
As identified in #13508, AST node of methods (`RBMethodNode`) provides some compilation-related services.
Especially, there are various methods named `*method*` or `*generate*` that seem redundant and could be simplified (and documented!)

Main changes are:

* remove `generateWithSource` and its 3 clients because there is always a source
* deprecate `generate` in favor of `generateMethod` that is more explicit.
  Note: the big diff of `OCBytecodeDecompilerExamplesTest` is related to the mass automatic deprecation of this method
* store the compiledMethod cache as a property of the node, and not in the compilationContext (one less ivar in the data-god-class, and one less dependency on the compilation context!)
* update `method` (I'm not that fan of the name, but I did not change it) to not retrieve the installed method as a fallback. That seems just wrong and bad and wrong again. But I'm afraid that will break clients. Let's hear what the CI says

### DIFF
--- a/src/AST-Core/CompiledMethod.extension.st
+++ b/src/AST-Core/CompiledMethod.extension.st
@@ -28,6 +28,6 @@ CompiledMethod >> parseTree [
 		source: self sourceCode;
 		class: self methodClass;
 		parse.
-	ast compilationContext compiledMethod: self.
+	ast compiledMethod: self.
 	^ast
 ]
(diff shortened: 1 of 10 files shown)


jecisc (approved):

privat:
9 failures, but in fact 3 because 3 os, but in fact 1 because 2 are caused by an insufficient teardown after the first one failed. not that bad

MarcusDenker (approved):

PR #13519, opened 2023-04-23
Clean info of MethodChangeRecord

jecisc:
MethodChangeRecord saves an #info variable that contains first the last source pointer and the protocol of a removed method. The protocol is never used so I propose to remove it and transform this ugly array (why? why? why an array?!)  into a variable storing the last source pointer.

I have the impression that the changes source code can be improved further but that will be for another time (or maybe it will be removed before)


MarcusDenker:
I think MethodChangeRecord used to be the model that the changesorter used. And then this was hacked in, I guess.

Now we use Ring for modeling historical code, so this is not needed anymore

PR #13521, opened 2023-04-23
Compiler get more scope responsibilities

privat:
Another part of #13508 and a follow up of  #13517

This moves some responsibility from CompilationContext to OpalCompiler, thus slimming the former, reducing coupling and improving cohesion.

Are moved totally 

* `requestor` \o/
* `bindings` additional bindings, a nice feature but not used that much it seems
* `interactive` a predicate related to requestor
* `class:` used to configure the semantic scope on a class

Are moved partially

* `getClass` only one client remain
* `semanticScope` the hunt continue

Are not moved yet:

* plugins
* `environement` and `productionEnvironment` it's not clear what they are used for. They seem very fragile


jecisc:
This one has conflict now

src/OpalCompiler-Core/OpalCompiler.class.st
```diff
@@ -221,13 +224,16 @@ OpalCompiler >> ast: anObject [
...
 ]
 
 { #category : #accessing }
-OpalCompiler >> bindings: aDictionary [
+OpalCompiler >> bindings: aCollectionOfAssociations [
 	"allows to define additional binding, note: Globals are not shadowed"
-	self compilationContext bindings: aDictionary
+	"bindings are LiteralVariables, not simple Associations.
+	This method receives any collection with Associations or Variables inside and converts those to AdditionalBindings"
+	bindings := (aCollectionOfAssociations asDictionary associations
```
jecisc, line 235: The `asDictionary associations` seems redondant if we already have a collection of associations. Is it to cast LiteralVariables into Associations? Then why do the #isAssociation check?
I wonder if there is a real need for this or if this is just a lost on perfs.

Of course this is not a proble of the current PR since it was already done like this.

src/OpalCompiler-Core/CompilationContext.class.st
```diff
@@ -16,7 +16,6 @@ Class {
 	#name : #CompilationContext,
 	#superclass : #Object,
 	#instVars : [
-		'requestor',
```
jecisc, line 19: Nice! :)

jecisc (approved):

src/OpalCompiler-Core/OpalCompiler.class.st
privat: I just moved the method. But you're right, that weird. Maybe @MarcusDenker knows if there is a real need or if it is some leftover code

MarcusDenker: yes, looks like it can be simplified, we can check later

MarcusDenker (approved):

MarcusDenker:
There is now a conflict

MarcusDenker:
It seems I merged this too fast: the last PR shows problems.

I added a revert: https://github.com/pharo-project/pharo/pull/13573

PR #13522, opened 2023-04-23
Cleanup-CompiledMethod-Source-Embedding

MarcusDenker:
This PR removes the last user of embedding source in the trailer.

With that done, it removes the code and test in CompiledMethodTrailer, too


jecisc (approved):
What a nice step to simplify the system!

Issue #13377, opened 2023-04-10
[CI] testReentrantCalloutsDuringCallback(#runner->[ TFWorker named: 'fortest' ]) randomly fail


jecisc:
This is not the first time I see this kind of failure. It happens once in a while.

Failing test:
testReentrantCalloutsDuringCallback(#runner->[ TFWorker named: 'fortest' ]) 

Last failure I've seen is on windows but I don't know if this is OS specific.

Trace:

```
Error
Got 448 instead of 7.
Stacktrace
TestFailure
Got 448 instead of 7.
TFCallbacksTest(TestAsserter)>>assert:description:resumable:
TFCallbacksTest(TestAsserter)>>assert:description:
TFCallbacksTest(TestAsserter)>>assert:equals:
TFCallbacksTest>>testReentrantCalloutsDuringCallback
TFCallbacksTest(TestCase)>>performTest
```

jecisc:
New failure:

```
Error
Got 1 instead of 7.
Stacktrace
TestFailure
Got 1 instead of 7.
TFCallbacksTest(TestAsserter)>>assert:description:resumable:
TFCallbacksTest(TestAsserter)>>assert:description:
TFCallbacksTest(TestAsserter)>>assert:equals:
TFCallbacksTest>>testReentrantCalloutsDuringCallback
TFCallbacksTest(TestCase)>>performTest
```

jecisc:
I didn't see this one in a while

PR #13525, opened 2023-04-23
RBMethodMode: compiledMethod sould be a simple accessor

privat:
RBMethodNode had two confusing methods:

* `method` that returns the associated compiled method if one exists (nil otherwise)
* `compiledMethod` that does the same but forces a compilation (with a possible limited context) if none exists.

While #13514 cleaned up the implementation, it did not touch the name or the semantic.
After analysis of the clients, it appears that almost none use `method` (the name is bad anyway) and most expect that `compiledMethod` is a simple accessor that does not necessarily trigger a compilation (I might have the same expectation, in fact).
Only two groups of tests seems to use `compiledMethod` to have a compilation (and are updated).

So, the PR propose to have:

* `compiledMethod` does not compile and may return nil
* `method` is deprecated in favor of `compiledMethod`

### DIFF
--- a/src/OpalCompiler-Core/RBMethodNode.extension.st
+++ b/src/OpalCompiler-Core/RBMethodNode.extension.st
@@ -8,7 +8,7 @@ RBMethodNode >> bcToASTCache [
 { #category : #'*OpalCompiler-Core' }
 RBMethodNode >> compiledMethod [
 	"Retrieve the associated CompiledMethod (cached version).
-	If no CompiledMethod was was generated, nil is returned.
+	If no CompiledMethod was generated, nil is returned.
 
 	When the AST is recompiled (see `generateMethod`), the cache is updated.
 	However, the cache is not reset if the AST is modified."
--- a/src/OpalCompiler-Core/RBMethodNode.extension.st
+++ b/src/OpalCompiler-Core/RBMethodNode.extension.st
@@ -7,13 +7,13 @@ RBMethodNode >> bcToASTCache [
 
 { #category : #'*OpalCompiler-Core' }
 RBMethodNode >> compiledMethod [
-	"Retrieve or generate a CompiledMethod (cached version).
-	The cache is not reset even is the AST is modified but can be overrided with `compiledMethod:`.
-	See limitations in `generateMethod`.
-	
-	To get a CompiledMethod without compiling it, see `method`"
+	"Retrieve the associated CompiledMethod (cached version).
+	If no CompiledMethod was was generated, nil is returned.
 
-	^ self propertyAt: #compiledMethod ifAbsentPut: [ self generateMethod ]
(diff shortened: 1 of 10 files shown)


privat:
I spend some hours on the failed test `ReflectivityReificationTest>>#testReifyMethod`. It made so little sense.
What failed is: `self assert: tag equals: ReflectivityExamples >> #exampleIvarRead.`

Where `tag` should be the instrumented method got though metalink reflectivity. It's the point of the test.
The thing is that `tag` was correct, it's `ReflectivityExamples >> #exampleIvarRead` that was weird: it contains the original compiled method, as if the reflective method was uninstalled.
Could the PR do subtle changes, causing some uninstallation of the link and the reflective method?
Could this be related to weird race conditions in the AST Cache?
Maybe I awaken some dormant bug lurking in the reflectivity package?

I tried to inspect the code, add breakpoints, and check the difference with another image without the commits.
The more I dig, the less things made sense. The primitive wrapper was correctly installed, the real instrumented compiled method has ok, but during the execution of the instrumented code the instrumented code is uninstalled, and if executed through a debugger, the test become green.
As a last measure, I added a breakpoint at `Behavior>>#methodDict` to see what was responsible for the uninstallation... it was `Deprecation>>#transform` that intercepted the depreciation of `RBMethodNode>>method` and installed a new compiled method, overwriting the instrumented version in a quite brutal manner.
Because, you see, the deprecated `method` call is not present in the original source, it is installed trough meta-programming and synthetic AST nodes. (but transformation don't care, I assume).
Anyway, if someone recompile blindly the source, you get the original method without the instrumentation!
And a failed test because a new version of `ReflectivityExamples >> #exampleIvarRead` was not expected.

src/OpalCompiler-Core/RBMethodNode.extension.st
```diff
@@ -7,13 +7,13 @@ RBMethodNode >> bcToASTCache [
 
 { #category : #'*OpalCompiler-Core' }
 RBMethodNode >> compiledMethod [
-	"Retrieve or generate a CompiledMethod (cached version).
-	The cache is not reset even is the AST is modified but can be overrided with `compiledMethod:`.
-	See limitations in `generateMethod`.
-	
-	To get a CompiledMethod without compiling it, see `method`"
+	"Retrieve the associated CompiledMethod (cached version).
+	If no CompiledMethod was was generated, nil is returned.
```
jecisc:
```suggestion
	If no CompiledMethod was generated, nil is returned.
```

jecisc (approved):

jecisc:
[CI] testReentrantCalloutsDuringCallback(#runner->[ TFWorker named: 'fortest' ]) randomly fail #13377

MarcusDenker (approved):

PR #13527, opened 2023-04-24
CompiledMethodTrailer: use 4 bytes as sourcePointer

MarcusDenker:
try to just encode the sourcePointer with #asLiteralArrrayOfSize: 4

This PR is created by editing the source directly and tries to do the minimum. If it succeeds, we can remove more code from CompiledMethodTrailer

### DIFF
--- a/src/Kernel/CompiledMethodTrailer.class.st
+++ b/src/Kernel/CompiledMethodTrailer.class.st
@@ -167,7 +167,9 @@ CompiledMethodTrailer class >> trailerKinds [
 CompiledMethodTrailer >> clear [
 	kind := #NoTrailer.
 	size := 4.
-	data := encodedData := method := nil
+	data := 0.
+	encodedData :=  #[0 0 0 0].
+	method := nil
 ]
 
 { #category : #'creating a method' }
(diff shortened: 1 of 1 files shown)


jecisc (approved):
Nice!

MarcusDenker:
With 4bytes we can encode offsets into the .changes and .sources file each for 2GB files (we need one bit for making the distinction).

I will ask on the Mailinglist if this is enough (but I think if this is not enough, this can only be that someone abuses the source storage to store data). But in case, we could just use 5 byte (or even full 64bit with 8). What matters is that this is *fixed*.

What is odd, too, is the encoding use for the offset right now. Naively I would take the offset in the file and shift it by 1, so that we can use the last bit for sources/changes.

But what it does:

```
sourcePointerFromFileIndex: index andPosition: position
	"Return a sourcePointer encoding the given file index and position"

	| hi lo |
	(index = 1 or: [index = 2])
		ifFalse: [self error: 'invalid source file index'].
	position < 0 ifTrue: [self error: 'invalid source code pointer'].
	hi := position // 16r1000000 *2 + index.
	lo := position \\ 16r1000000.
	^ hi * 16r1000000 + lo
```

Why?

MarcusDenker:
So the large image that is created here: https://ci.inria.fr/pharo-ci-jenkins2/job/latest-large-image/

has a .changes of ~4GB


<blockquote><div><strong><a href="https://ci.inria.fr/pharo-ci-jenkins2/job/latest-large-image/">latest-large-image [Jenkins]</a></strong></div></blockquote>

MarcusDenker:
I will merge this for now and then make it larger in a next step

PR #13533, opened 2023-04-24
Improve ClassDescription>>#addAndClassifySelector:withMethod:inProtocol:

jecisc:
ClassDescription>>#addAndClassifySelector:withMethod:inProtocol: was manipulation protocol names. This PR is pushing the use of real protocols.

This also removes 2 users of #protocolNameOfElement: that contains a hack heavily used in the system. Let's see if this method is part of the culprits or if it can use the cleaner version. 

Next steps:
- Use real protocol in the announcements
- Remove the `self class environment at: #RPackage ifPresent: [ :rPackageClass |` since we want RPackage to be part of the default MM


jecisc:
This one apparently needs the hack

jecisc:
This one pass now! :)

MarcusDenker (approved):

PR #13534, opened 2023-04-24
Refactor block ensure during termination to allow override

Mathilde411:


src/Kernel/BlockClosure.class.st
```diff
@@ -845,6 +845,14 @@ BlockClosure >> valueUnpreemptively [
 	^result
 ]
 
+{ #category : #scheduling }
+BlockClosure >> valueUnwindInContext: ctxt [ 
+	| ensureContext |
+	ensureContext := self asContextWithSender: ctxt.
+	ensureContext runUntilErrorOrReturnFrom: ensureContext
+
+]
```
jecisc: For this kind of methods that can be obscure to people that do not know the internals of Block management/context/.. it might be good to add a method comment. Or maybe move the one of #doTerminationFromAnotherProcess there?

jecisc (changes requested):

src/Kernel/BlockClosure.class.st
guillep: Yes! Agree, thanks @jecisc !

guillep:
Tests were green the previous run! Now with the comment move they should be ok too!

Thanks @BastouP411 and @jecisc for the review !

jecisc:
I did not notice but this was merged in Pharo 11. 

This kind of changes should probably be integrated in Pharo 12since the release is right around the corner

PR #13535, opened 2023-04-24
Rename some "category" in MCMethodDefinition

jecisc:
To favor "protocol" vocabulary


jecisc:
This is just renaming... I don't get why this does not work

jecisc:
This one is not so easy and I have other PR to take care of before so I'll close it for now

PR #13537, opened 2023-04-24
Do not hardcode stylers

jecisc:
In the past I had some personnal IconStyler and I noticed now that they stopped working. So reason is that in the past all stylers were applied. Now I found a hardcoded list of stylers in IconStyler class>>#withStaticStylers.  This is really bad because it means that nobody can extend the styles. 

I now collect the stylers to apply in a way that allow to add your own. I think we can do better but this already improves the situation

### DIFF
--- a/src/Reflectivity-Tools/IconStyler.class.st
+++ b/src/Reflectivity-Tools/IconStyler.class.st
@@ -12,6 +12,12 @@ Class {
 	#category : #'Reflectivity-Tools-Breakpoints'
 }
 
+{ #category : #testing }
+IconStyler class >> isStaticStyler [
+
+	^ false
+]
+
 { #category : #style }
 IconStyler class >> withStaticStylers [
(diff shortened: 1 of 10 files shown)


MarcusDenker (approved):

src/Reflectivity-Tools/IconStyler.class.st
```diff
@@ -22,8 +28,8 @@ IconStyler class >> withStaticStylers [
 	"
 
 	^ self new
-		stylerClasses:  {BreakpointIconStyler. CounterIconStyler. DisabledBreakpointIconStyler. FlagIconStyler. HaltIconStyler. MetaLinkIconStyler. SemanticWarningIconStyler. WatchIconStyler. DocCommentIconStyler};
-		yourself
+		  stylerClasses: (self allSubclasses select: #isStaticStyler);
```
Ducasse, line 31: please use a block instead of a symbol

PR #13539, opened 2023-04-24
Classify method before installing them

jecisc:
The management of protocol is currently hard to clean further because some hacks are present.

One of the hack is in #protocolNameOfElement:. This method check if an element is in the protocols of the class and if it is not, it will check if the method is included in the organized class. If it is, then it will return the unclassified protocol. 

One of the case where we rely on this hack is when we first compile a method. The method gets added to the method dictionary and then we classify it, meaning that it is not yet into any protocol. I would like to see what happens if we classify it before we add it to the method dictionary. Meaning that for the first compilation, the classification should have the behavior expected without the hack of #protocolNameOfElement:

Let's see if Pharo can handle that

### DIFF
--- a/src/Kernel-Tests/ClassOrganizationTest.class.st
+++ b/src/Kernel-Tests/ClassOrganizationTest.class.st
@@ -85,6 +85,31 @@ ClassOrganizationTest >> testClassifyUnder [
 ]
 
 { #category : #tests }
+ClassOrganizationTest >> testClassifyUnderWithNil [
+	"Set the base for the test"
+	| unclassified|
+	unclassified := Protocol unclassified.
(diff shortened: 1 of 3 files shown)


jecisc:
In order to make this pass I had to add some nil management to #classify:under:. This was the plan already to be coherent with the rest of the API! The code is currently ugly but a big cleaning is coming soon. 

I also added a test. This is what interest me the most on this to be sure my future cleaning will not break the system.

jecisc:
Tests passed!

It does not bring much except the tests and the help of relying less on the hack of #protocolNameOfElement:. The complex double imbricated if will be removed soon.

MarcusDenker (approved):

PR #13540, opened 2023-04-24
Scope: generalize target class

privat:
For the semantic analysis, scopes are nested objects, so demanding a variable binding is either answer or delegated to an outer scope.
Except for `targetClass` and `targetEnvironment` that were only available at a single scope level (the weirdly named SemanticScope).
So this PR push up `targetClass` and `targetEnvironement` (renamed as environment) so all scope can access it, and update some clients.

Especially, it updates `RBMetodNode>>methodClass` to simply rely on its scope instead of asking compilationContext.
The side effect is that `RBMetodNode scope` must be setup earlier in the semantic analysis, as some plugin examples expect it.
This feel like a good move, but I'm willing to consider alternatives if someone have a better idea.

### DIFF
--- a/src/AST-Core/RBMethodNode.class.st
+++ b/src/AST-Core/RBMethodNode.class.st
@@ -470,7 +470,7 @@ RBMethodNode >> method [
 { #category : #'accessing - compiled method' }
 RBMethodNode >> methodClass [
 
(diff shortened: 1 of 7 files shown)


jecisc (approved):
This seems good to me

MarcusDenker (approved):

MarcusDenker:
yes, I think it is ok

PR #13544, opened 2023-04-24
Reflectivity: make tests more reliable (and fix a bug)

### LINKED ISSUE #13516: Metalinks: link uninstallation do seems to leak cruft
Testing of `Reflectivity-Tests` are unreliable when there is some failures or errors, but also in the case of success.
Especially, `link uninstall` seems to left things around. 

To reproduce, run tests of `LinkInstallerTest` repeatedly.
See also #13514

privat:
should fix #13516

Tests teardown in metalinks were limited and some leaked metalinks cause failure on retry.
Moreover, unexpected leaks could make the test green but fail subsequent tests.

The PR forces each test to clean after itself. Failure to do so means an error in the test or elsewhere and should be investigated.

* add missing link uninstallations on tests that missed some
* new superclass ReflectivityTestCase that offer a teardown that detect metalink leaks, fails the corresponding test, and try to clean the leak.
* fix a bug preventing removal of *instead* metalinks on global variables. When using instead, the global is no more read, thus the method with the *instead* instrumentation won't be listed on `usingMethods`, thus the instrumentation won't be removed. The solution is to inject the global as a forced literal to be present on `usingMethods`.

### DIFF
--- a/src/Reflectivity-Tests/LinkInstallerTest.class.st
+++ b/src/Reflectivity-Tests/LinkInstallerTest.class.st
@@ -31,9 +31,6 @@ LinkInstallerTest >> setUp [
 LinkInstallerTest >> tearDown [
 	ReflectivityExamples2 new removeModifiedMethodWithInstVarAccess.
 	ReflectivityExamples new removeTemporaryMethods.
-	Smalltalk garbageCollectMost.
-	MetaLink uninstallAll.
-	Smalltalk garbageCollectMost.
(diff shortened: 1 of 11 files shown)


jecisc (approved):
"single method" tear down are generally in an ensure, but since the global tear down will look for all leaks then it should be good to clean after the tests.

Thanks

MarcusDenker (approved):

PR #13546, opened 2023-04-25
CompiledMethodTrailer: more cleanup, do not use it for #endPC

MarcusDenker:
- remove the kind ivar from CompiledMethodTrailer
- change CompiledMethod endPC to not instantiate a  CompiledMethodTrailer
- simplify setSourcePointer:
- less sends to #trailer

### DIFF
--- a/src/Kernel/CompiledMethod.class.st
+++ b/src/Kernel/CompiledMethod.class.st
@@ -310,6 +310,12 @@ CompiledMethod class >> toReturnSelfTrailerBytes: trailer [
 	^self newBytes: 3 trailerBytes: trailer nArgs: 0 nTemps: 0 nStack: 0 nLits: 2 primitive: 256
 ]
 
+{ #category : #constants }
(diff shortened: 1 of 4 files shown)


src/Kernel/CompiledMethod.class.st
```diff
@@ -398,7 +404,7 @@ CompiledMethod >> displayStringOn: aStream [
 { #category : #accessing }
 CompiledMethod >> endPC [
 	"Answer the index of the last bytecode, as the trailer size varies we have to delegate"
```
jecisc, line 406: Maybe the comment should be updated since it’s not really délégation anymore? But this is a detail

jecisc (approved):

src/Kernel/CompiledMethod.class.st
MarcusDenker: yes, I will fix it next

PR #13547, opened 2023-04-25
Simplify #classify:under:

jecisc:
Since #13373 has bad effects on the system and I cannot find the root, here is a first cleaning to find the origin of the problem


jecisc:
So this part already produces the problem.. Let's dig more.

jecisc:
With this change, all classes with unclassified methods seems to be from traits. So I guess that Traits have a mecanic were it adds methods to a class without classifying them?

jecisc:
This one is now integrating into https://github.com/pharo-project/pharo/pull/13373

We can close

PR #13548, opened 2023-04-25
Autocompletion must replace from start of the word if case-insensitiv…

Mr-Dispatch:
…e completions are turned on, otherwise wrongly-cased letters are not overwritten.


jecisc:
The test failure is a known random failure.

privat (approved):

MarcusDenker (approved):

PR #13549, opened 2023-04-25
Ensure methods are in a protocol

jecisc:
All methods should be in a protocol when installed in a class.  Currently, this is not the case. Some methods are added silently and never added to the class protocols. Then we have hacks that check if a method is classified or not in the protocol management.

Here I ensure that if a method is not classified while been added to a class, we put it in the unclassified protocol.  If the method is installed by Opal or if the method was already classified in the past, then this will do nothing.

I hope that this move will help me to deal with the hacks in protocols preventing me from simplifying further the protocol management.

### DIFF
--- a/src/Kernel/ClassDescription.class.st
+++ b/src/Kernel/ClassDescription.class.st
@@ -86,8 +86,13 @@ ClassDescription >> addSelector: selector withMethod: compiledMethod [
 
 { #category : #'accessing - method dictionary' }
 ClassDescription >> addSelectorSilently: selector withMethod: compiledMethod [
+
 	super addSelectorSilently: selector withMethod: compiledMethod.
-	self instanceSide noteAddedSelector: selector meta: self isMeta
+	self instanceSide noteAddedSelector: selector meta: self isMeta.
+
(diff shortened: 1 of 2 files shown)


jecisc:
It works but the CI time increased a lot :/ 

I'm not sure it's a good move to integrate the change in that state

MarcusDenker:
The whole point of #addSelectorSilently:withMethod: is to *not* raise any announcements. That's why it is called "silent".

jecisc:
Yes indeed. I need to see what is the best plan here.
I think I'll pause that and do other cleanup to see what is the best way to do this.
Either fix the sender to categorize the method. Or classify silently. Or maybe both. Fix some senders that should announce classification and classify silently for the few cases that are left. 

It's still hard to find really the places that add methods without classifying them so I need to investigate that before pushing this PR.

MarcusDenker:
With "silent" means no Announcing, why whould you need to turn off the announcer for calling  ClassOrganization>>#silentlyClassify:under: ? 

Following the "silently" idea, it should do whatever it does without announcing anything.

PR #13550, opened 2023-04-25
CompiledMethodTrailer-Cleanup-even-more 

MarcusDenker:
... more cleanups


MarcusDenker:
it turns out that the compiler does not even need to know about it... the compiler can always create a method with an empty trailer, and all users that want to preserve the source pointer can easily do that by setting it.

This PR is the first step towards that

jecisc (approved):
Nice cleaning!

PR #13551, opened 2023-04-25
Factorize some code in recompilation management

jecisc:
Factorize some code in trait method recompilation.


jecisc:
Hum... Seen that we have a test in failure I guess that mean that the order of calls is important. I'll need to check that in more details.

jecisc:
There is one thing I find a little weird and inconstant in recompilation:
- In a non traited class, recompilation is silent
- In a traited class, recompilation of trait method is silent
- In a traited class, recompilation of a method that is not from a trait is not silent

Why recompiling a mthod from a non traited class is silent but from a traited class it is not silent?

MarcusDenker:
yes, this seems to be wrong that it emits the Announcement, the refactoring now makes it clearer that something is wrong

(#addselector:withRecompiledMethod: now calls addselector:withMethod: )

MarcusDenker (approved):

MarcusDenker:
recompilation is much slower in Pharo12:

Pharo12:
[OpalCompiler recompileAll ] timeToRun "0:00:00:54.25"
 
Pharo11:
 [OpalCompiler recompileAll ] timeToRun "0:00:00:41.821"

PR #13552, opened 2023-04-25
Rename #category into #protocolName

jecisc:
And clean up a little

This is a subpart of #13535, but since it fails with a strange error while the only changes were on names I'm splitting the PR.


jecisc:
I guess the problem comes from the loader that probably uses the name of instance variables to create the MCMethodDefinition while reading a mcz or something. I'll have to see if I can update that to stay compatible with previous version while unifying the names in Pharo. 

I'll do that in the other PR.

PR #13557, opened 2023-04-26
AST: add a scanner state

privat:
#13553 identified issues on the parsing of array literals.
#13554 tried to improve the situation with combined keyword and identifiers in literal arrays, but inconsistencies remains.

The main issue is the presence of `RBParser>>#patchLiteralArrayToken` that takes tokens returned by the scanner and tries to recombine them, but does a bad job because it's very complex to do correctly, especially if you don't want to lose information.

The solution proposed by this PR is to solve the problem at the origin: in the scanner, freeing the parser of retrofitting some lexer job.
The classical approach to solve this problem is to use *lexer states* (not to be confused with automaton states) that correspond to scanning policy. Basically, the state controls what the next token is.
I added a `state` ivar to the scanner and use it to have alternative paths in `scanToken`. This way identifiers and binops are treated as identifiers and binops in the default state but are treated as symbols in the *literalArray* state.
The state is currently controlled by the parser in `RBParser>>#parseLiteralArray`.
Lexer states are notorious tricky (because you have to think about lookaheads), but they do the job when used correctly, and they simplify the code a lot.

I also needed to do some factorization in the scanner to prevent code duplication. Especially because true, false and nil are named literal values that need special care.

So, what are the free benefits? i.e. things that now fixed out of the box

* Retrieve some lost comments
* Fix some highlighting issue
* Fix some formatting issue
* Correctly parse `#(foo:==)` as `#(#foo: #==)` whereas the bad `patchLiteralArrayToken` produced `#(#foo: #= #=)`

What remains to do?

Special characters in literal arrays are still parsed as symbols. As a remainder, they are `^[]{}:;.` (note `()#` are also special but are meaningful in arrays)
I'm not sure if this is a good idea, so I let them the current `patchLiteralArrayToken` (it is basically only what remain here) and I did not push up that to the scanner state. In a future PR I will finish the job (kill or move).
So what do you prefer for `#(a . b c`... ?

* `#(#a #'.' #b #c`... current behavior. Accept most garbage and try to make them symbols.
  Note: I don't like this one for a lot of reasons. It also confuses newbies `{1. 2}` (ok)  vs. `{1 2}` (an error) vs. `#(1 2)` (ok) vs. `#(1. 2)` (ok but not what you might expect)
* `#(#a <syntax error unexpected '.' literal wanted> #b #c`... first alternative. Report an error and continue the parsing of array elements
* `#(#a <missing closing parenthesis>. b c`... second alternative. Report an error and assume the array is now finished. This one make sense when programming, as an unclosed literal arrays will be detected earlier and not consume the rest of the source code

### DIFF
--- a/src/AST-Core/RBScanner.class.st
+++ b/src/AST-Core/RBScanner.class.st
@@ -32,7 +32,8 @@ Class {
 		'characterType',
 		'classificationTable',
 		'comments',
-		'nextToken'
+		'nextToken',
+		'state'
 	],
 	#classVars : [
 		'CascadePatternCharacter',
@@ -619,6 +620,18 @@ RBScanner >> scanUnknownCharacter [
 	errorToken := self scanError: 'Unknown character'.
 	errorToken location: tokenStart. "Unlike other scanner errors, error location is here before the bad token"
 	^ errorToken
+]
+
+{ #category : #accessing }
+RBScanner >> state [
+
+	^ state
+]
+
+{ #category : #accessing }
+RBScanner >> state: anObject [
+
+	state := anObject
 ]
 
 { #category : #private }
--- a/src/AST-Core/RBScanner.class.st
+++ b/src/AST-Core/RBScanner.class.st
@@ -599,11 +599,17 @@ RBScanner >> scanToken [
 	case statement. Didn't use Dictionary because lookup is pretty slow."
 
 	characterType = #eof ifTrue: [ ^ RBEOFToken start: tokenStart + 1 ]. "The EOF token should occur after the end of input"
-	characterType = #alphabetic ifTrue: [^self scanIdentifierOrKeyword].
+	characterType = #alphabetic ifTrue: [
+		^ state = #literalArray
+			ifTrue: [self scanSymbolOrNamedLiteral]
+			ifFalse: [self scanIdentifierOrKeyword]].
 	(characterType = #digit
 		or: [currentCharacter = $- and: [(self classify: stream peek) = #digit]])
 			ifTrue: [^self scanNumber].
-	characterType = #binary ifTrue: [^self scanBinarySelector].
+	characterType = #binary ifTrue: [
+		^ state = #literalArray
+			ifTrue: [ self scanLiteralBinary]
+			ifFalse: [ self scanBinarySelector]].
 	characterType = #special ifTrue: [^self scanSpecialCharacter].
 	currentCharacter = $' ifTrue: [^self scanLiteralString].
 	currentCharacter = $# ifTrue: [^self scanLiteral].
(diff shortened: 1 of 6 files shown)


src/AST-Core/RBScanner.class.st
```diff
@@ -514,6 +496,29 @@ RBScanner >> scanName [
...
+			[^RBLiteralToken
+				value: false
+				start: tokenStart
+				stop: self previousStepPosition].
+	name = 'nil'
+		ifTrue:
+			[^RBLiteralToken
+				value: nil
+				start: tokenStart
+				stop: self previousStepPosition].
```
jecisc: Since we have at least 4 times this code with just a different value maybe we could add a method #literalTokenForValue:?

But this is a detail here

jecisc:
I like the option 2 and 3 because each time I had a "." in a literal array it was a mistake. I have a little preference for option 2.

It will probably break some code but since we can load it and patch it, it should be fine

jecisc (changes requested):
This is shadowing the temp in RBScanner>>#isSelectorNonBinary

src/AST-Core/RBScanner.class.st
privat: Done. The new method is also used elsewhere

privat:
Update: I removed the token patching altogether, and directly generate the AST node when a Special token is involved, this fixes the remaining issues on styling and comments.
I also added a warning, so the parsing is still compatible but you get a yellow background on rubric/calypso (not spec/newtools unfortunately)

privat:
Bootstrap segfault now :(

Ducasse:
HUGE THANKS. I really really love to see this part of Pharo improving.

privat:
I might have found the bootstrap issue.

The segfault occurred during the loading of the Array class (somewhere inside `ShiftClassInstaller`) because some things in the model were corrupted (nevertheless, the VM should be more robust).
It's likely that the corruption was caused during the parsing of chunks performed by code importer or Monticello. `String>>parseLiterals` (that ultimately calls `RBParser>>parseLiterals:`) is used to split random strings into tokens to recognize the kind of chunks and extract data (seriously, this is very fragile and hackish).
Unfortunately, `parseLiterals:` needed love (the argument was not used for instance), had no test and poor error management. So the removal of `patchLiteralArrayToken` just silently broke it.

New commits of the PR update `parseLiterals:` and add tests. I hope this solves the bootstrap issue.

MarcusDenker (approved):

MarcusDenker:
Regarding the question: yes, 2 or 3 would be good. a . in a literal array is always a mistake. Not sure about 2 or 3... 2 feels better, but maybe we should focus on code writing, which means 3.

PR #13564, opened 2023-04-27
Cleanup: do not reference CompiledMethodTrailer in ReflectiveMethod

MarcusDenker:
- do not use generate: in ReflectiveMethod
- start deprecation, for now just deprecated #trailer (and keep it in the package for now)
- fix comment #endPC

### DIFF
--- a/src/Kernel/CompiledBlock.class.st
+++ b/src/Kernel/CompiledBlock.class.st
@@ -22,7 +22,7 @@ CompiledBlock >> displayStringOn: stream [
 
 { #category : #accessing }
 CompiledBlock >> endPC [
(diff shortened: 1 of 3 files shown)


MarcusDenker:
this PR removes the last two references to CompiledMethodTrailer outside of the compiler. And the compiler itself just passes the argument through to the bytecode generator, which does not use it anymore.

This means we can, after merging this, cleanup the compiler and remove all the trailer arguments, which should make the API much simpler

jecisc (approved):

MarcusDenker:
strangely enough, this seems to lead to problems. I will revert

PR #13568, opened 2023-04-27
Revert "Cleanup: do not reference CompiledMethodTrailer in ReflectiveMethod"

MarcusDenker:
Reverts pharo-project/pharo#13564


MarcusDenker:
But it does not work.

So I checked: all PRs merged where green. But now the build is broken.

PR #13569, opened 2023-04-27
Add compiler option to not optimise special sends

guillep:
This helps in having method proxies that capture everything.
Otherwise, optimised sends are doing static type predictions, avoiding message sends (thus lookup and method proxies).

Done with @PalumboN , @tesonep , @jordanmontt , @BastouP411

### DIFF
--- a/src/OpalCompiler-Tests/OCCalledMethodProxy.class.st
+++ b/src/OpalCompiler-Tests/OCCalledMethodProxy.class.st
@@ -1,0 +1,73 @@
+Class {
+	#name : #OCCalledMethodProxy,
+	#superclass : #Object,
+	#instVars : [
+		'originalMethod',
+		'called'
+	],
+	#category : #'OpalCompiler-Tests-AST'
+}
+
+{ #category : #'instance creation' }
+OCCalledMethodProxy class >> on: aMethod [
+
+	^ self new
+		  originalMethod: aMethod;
+		  yourself
+]
+
+{ #category : #'accessing - pragmas & properties' }
+OCCalledMethodProxy >> cachePragmas [
+	
+	^ originalMethod cachePragmas
+]
+
+{ #category : #accessing }
+OCCalledMethodProxy >> called [
+
+	^ called
+]
+
+{ #category : #initialization }
+OCCalledMethodProxy >> initialize [
+
+	super initialize.
+	called := false
+]
+
+{ #category : #accessing }
+OCCalledMethodProxy >> methodClass: aClass [ 
+	
+	originalMethod methodClass: aClass
+]
+
+{ #category : #accessing }
+OCCalledMethodProxy >> originalMethod [
+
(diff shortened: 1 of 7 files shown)


src/OpalCompiler-Tests/OCCalledMethodProxy.class.st
```diff
@@ -0,0 +1,76 @@
...
+	originalMethod := anObject
+]
+
+{ #category : #evaluation }
+OCCalledMethodProxy >> run: aSelector with: arguments in: aReceiver [
+
+	| notFirstTime |
+	notFirstTime := called.
+	called := true.
+	notFirstTime ifFalse: [ 	1 haltOnce ].
```
PalumboN: halt

PalumboN (approved):
Nice!

src/OpalCompiler-Tests/OCCalledMethodProxy.class.st
jordanmontt: There is a forgotten halt in here

src/OpalCompiler-Tests/OCSpecialSelectorTest.class.st
```diff
@@ -0,0 +1,122 @@
...
+	| proxy |
+	optimisationsActive := true.
+	self compileExample: #exampleSimpleBlockiVar.
+	instance iVar: #valueToTest.
+	[
+		instance class addSelectorSilently: #exampleSimpleBlockiVar withMethod: method.
+		proxy := self proxy: BlockClosure >> #value.
+
+		self deny: proxy wasCalled.
+		self assert: instance exampleSimpleBlockiVar equals: #valueToTest.
```
jordanmontt, line 97: Maybe it is worth to add a comment in here to say why we send a message instead of calling `(self executeMethod: method onReceiver: instance)`. In the dojo we saw that is because of the JIT compiler. Else will be difficult to understand the why

jordanmontt (changes requested):

src/OpalCompiler-Tests/OCCalledMethodProxy.class.st
guillep: doh!

guillep: doh!

src/OpalCompiler-Tests/OCSpecialSelectorTest.class.st
guillep: Yes!

jordanmontt (approved):

MarcusDenker:
There are tests failing in OCBytecodeGeneratorTest

MarcusDenker:
Windows CI has problems (diskpace?) everything else is green

MarcusDenker (approved):

PR #13572, opened 2023-04-27
Fix case were method is in a class but not in a protocol

jecisc:
When a class has a slot, it generates an initialization method. But during the first creation of a class, this method got removed from its protocol when the class copied the organization of the old class.

This PR puts the organization copy before the initialization of slots to ensure the slot stay classified. It also adds a test.

This problem was detected by: #13533


jecisc:
Test failures seems to be only reflectivity ones. But let's ask again the CI just in case

MarcusDenker (approved):

PR #13573, opened 2023-04-27
Revert "Compiler get more scope responsibilities"

MarcusDenker:
Reverts pharo-project/pharo#13521


MarcusDenker:
This was merged but it breaks the tests of Reflectivity I think.

It seems that the ReflectiveMethod>>#compileAST does not know the right class, leading to bad effects (the methods are not installed correctly)

jecisc:
The problem seems to come from the last commit: https://github.com/pharo-project/pharo/pull/13521/commits/63a51b6ee8e0fd581f2a1712f2f772f6844d9331

Before this merge the build was green

Issue #13543, opened 2023-04-24
Registered surface not being deallocated


StewMacLean:
This is related to issue AthensCairoSurface not being deallocated #13053

Within AthensCairoSurface>>finializeResourceData:

a primitive is called via:

self unregisterSurfaceWithId: id.

As reported in #13053, the surface is not being found for the provided id. 

Using XCode Instruments to monitor allocated memory it shows that the primitive is failing to deallocate the surface's memory as the following screenshots demonstrate:

![03 createManualSurface](https://user-images.githubusercontent.com/97324099/234130863-f9eaa656-615d-4678-a383-8d1df88dd304.png)

![05 primitiveCreateManualSurface leak](https://user-images.githubusercontent.com/97324099/234130912-c54614ac-53cb-46ba-912f-c83f47967ff4.png)

**To Reproduce**
1. Using XCode > Open Developer Tools > Instruments > Leaks 
attached to the Pharo process
3. Execute 10000 timesRepeat: [AthensCairoSurface extent: 512 @ 512].
4. Observe the memory leak as reported in Instruments

**Expected behavior**
The memory for the surface should be deallocated by the primitive.

**Version information:**

OS: macOS
Version: 12.0.1
Pharo Version 10

**Expected development cost**
This is an issue with the VM. If the associated problem with finding the surface is fixed, this may fix this issue too?

MarcusDenker:
PR was merged

tinchodias:
Excelent!!! thanks

PR #13574, opened 2023-04-27
13526-Attempt-to-release-a-shared-cairo-font-face

### LINKED ISSUE #13526: Attempt to release a shared cairo_font_face
**Bug description**
We are getting recently very often this inform box during image saving: `Attempt to release a shared cairo_font_face. Please report this on pharo-dev mailing list`.  Sometimes even a dozen of them.

It comes from: 'CairoFontFaceExecutor>>#finalize' and this PR describes the background of it: https://github.com/pharo-project/pharo/pull/565

> A cairo_font_face could still be referenced from external heap and further used.
> Thus we must NOT release the FT_Face until the external reference count drops to 1.
> 
> Thus the CairoFontFace is not a good place for retaining a strong reference to the FT2Face, because the reference will vanish as soon as the Smalltalk object is garbage collected, while some external cairo_font_face still point on it.
> It's better to have the strong reference in the executor, and avoid garbageCollecting the executor while there are external references to the resource data...
> 
> The solution is to use a specific CairoFontFaceExecutor.
> If the resource is shared, then we have no other choice than regularly polling the ref count in a new Process.
> This process must stop whenever the session change.

It is probably related to changes in ephemerons. So first question is, have you seen it as well? @nicolas-cellier-aka-nice

tesonep:
Fix #13526 and #13543

- The surfaces should be released even if they are not registered in the SurfacePlugin. The Plugin handles this scenario. Also, the findSurface primitive is only working for 32bits pointers.
- The CairoFontFace can be released when not used (it is using an internal reference counting mechanism, so it is safe to release them).
- The CairoFontFace should keep a strong reference to the FTFace, before it was kept in the executor (ugly... )
- The executor is not needed anymore and we can use the ExternalAddress as handle to finalize.


tesonep:
Now it is safe to keep the reference in teh CairoFontFace as now using Epehemerons the objects is not dereferenced until the object is finalized.

https://github.com/pharo-project/pharo/pull/565

guillep (approved):
Good!

jecisc:
CI hit github API rate limit. This was fixed only in P12

Ducasse:
Thanks Pablo this is good to see all these improvements.

Ducasse:
Should we backport it to P11?

MarcusDenker:
the PR is for Pharo11, we should merge it here and then sync all the Pharo11 changes next week to Pharo12

PR #13575, opened 2023-04-27
Compiler: Cleanup trailer for the frontend

MarcusDenker:
Cleanup the Compiler, it does not need to care about CompiledMethodTrailer.

### DIFF
--- a/src/OpalCompiler-Core/IRMethod.class.st
+++ b/src/OpalCompiler-Core/IRMethod.class.st
@@ -125,15 +125,18 @@ IRMethod >> compiledBlock: scope [
 { #category : #translating }
 IRMethod >> compiledMethod [
 
-	^ self compiledMethodWith: CompiledMethodTrailer empty
+	^compiledMethod
+		ifNil: [self generate ]
+		ifNotNil: [compiledMethod]
 ]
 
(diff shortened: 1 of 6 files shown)


src/OpalCompiler-Core/IRMethod.class.st
```diff
@@ -125,15 +125,18 @@ IRMethod >> compiledBlock: scope [
 { #category : #translating }
 IRMethod >> compiledMethod [
 
-	^ self compiledMethodWith: CompiledMethodTrailer empty
+	^compiledMethod
+		ifNil: [self generate ]
+		ifNotNil: [compiledMethod]
```
jecisc, line 130:
```suggestion
	^compiledMethod ifNil: [self generate ]
```

The ifNotNil branch is redundant here

jecisc (approved):
I did an esthetic suggestion, else it seems good!

It's incredible how much complexity this trailer was adding everywhere

src/OpalCompiler-Core/IRMethod.class.st
MarcusDenker: I made a note to do this with a later PR

Issue #13512, opened 2023-04-21
[Meta] Simplify CompiledMethod trailer: just support fixed sourcePointers


MarcusDenker:
**Bug description**

CompiledMethodTrailer is quite complex for what it does. In the end, the compiledMethod stores a sourcePointer in the 4 last bytes. We need 1 bit to destinguish sources from .changes, so the 32bits should be good for an offset in 2GB .changes and .sources

But the encoding used is quite strange, for one, we use the last byte to define which encoding we use. This is needed as with 3 bytes, we can not encode the sourcePointer already for our ~40MB .sources, so we use a #VarLengthSourcePointer for those:

```
Smalltalk globals methods count: [ :method | method trailer kind == #VarLengthSourcePointer ].
 "19841"
```

Variable length sourcePointers are not that nice in general, as we have to create another compiledMethod (and do a become) if we want to set it.

TODO

Remove the need for embedding sources in the trailer
- [x] CompiledMethod simplification: remove the last user of #copyWithSource: #13507
- [x] remove all unsused encodings from CompiledMethodTrailer. #13510
- [x] check users of generateWithSource: removed by https://github.com/pharo-project/pharo/pull/13514
- [x] fix Fuel to not use source embedding (see FLCompiledMethodCluster class>>#setTrailerWithSourceCode) #13522
- [x] CompiledMethodTrailer: remove #EmbeddedSource and #EmbeddedSourceWide encoding #13522
- [x] CompiledMethodTrailer: just use #asByteArrayOfSize: 4 to encode sourcePointer #13527
- [x] Cleanup CompiledMethodTrailer extendedSourcePointer #13545
- [x] simplify CompiledMethod>>#setSourcePointer: #13546
- [x] simplify CompiledMethod>>#endPC, we do not need to instantiate CompiledMethodTrailer anymore #13546
- [x] Cleanup: CompiledMethodTrailer should not create methods #13555
- [x] Cleanup: Fuel does not need to use #copyWithTrailerBytes: #13560
- [x] do not use CompiledMethodTrailer in ReflectiveMethod>>#compileAST #13564
- [x] Compiler: Cleanup for trailer removal in IRByteCodeGenerator #13565
- [x] remove CompiledMethodTrailer usage in the compiler (not used by the bytecode generator anymore) #13575
- [x] Deprecate class CompiledMethodTrailer  #13579
- [x] introduce CompiledMethod>>#clearSourcePointer and #sourcePointer: #13578
[...]

PR #13580, opened 2023-04-28
Set #trailerSize to 5

### LINKED ISSUE #13616: CompiledMethod trailer: use more then 4 bytes
After the CompiledMethodTrailer is now gone (see #13512), we use a fixed 4 byte trailer.

This means, .sources and .changed can be 2GB each. 

We need to check to set the trailer to a larger number 
- to test that everything works so we can change it
- use 5 bytes now, as the large image that the CI buillds (for Pharo9) generates a 4GB .changes

Later I want to experiment to use 8 bytes and store a Soil object ID.

The PR https://github.com/pharo-project/pharo/pull/13580 shows that there are strange problems: the image builds, it even is usable (but not "senders of"). And the tests fail for Traits. So *somewhere* we seem to not take the trailer size into account corrrectly.

MarcusDenker:
This means that we can grow sources to all practical sizes (this is 512GB per file, I think)

### DIFF
--- a/src/Kernel/CompiledMethod.class.st
+++ b/src/Kernel/CompiledMethod.class.st
@@ -202,8 +202,8 @@ CompiledMethod class >> subclassResponsibilityMarker [
 
 { #category : #constants }
 CompiledMethod class >> trailerSize [
-	"we use the last 4 byte to store the source pointer, this allows for 2GB .sources/.changes"
-	^ 4
+	"we use the last 5 byte to store the source pointer"
+	^ 5
 ]
 
 { #category : #queries }


jecisc (approved):

MarcusDenker:
Hard crash when running traits tests...

astares:
For the records:

The last PR now changes the trailer size to 5.


	https://github.com/pharo-project/pharo/pull/13580


This way .changes/sources size can be 512GB. 

The most important things is, though, that this validates that we can change the space easily, e.g
use 64bit and store some ObjectID...

The fixed trailer does have an impact on speed for bytecode interpretation (via the Context).

e.g. when looking for senders of special selectors, it has to fall back to use InstructionStream scanFor:,
as the selectors are not in the literals.

[#+ senders] bench

Pharo11 : "'3.891 per second'"
Pharo12 : "'4.913 per second'"

This is some 25% faster, due to not having to instantiate a CompiledMethodTrailer for #endPC.

	Marcus

On 28 Apr 2023, at 11:07, Marcus Denker <marcus.denker@inria.fr> wrote:

Hi,

We did (im Pharo12) a simplification of how method trailers are handled. Below is the class comment of the now Deprecated and empty
class CompiledMethodTrailer. 

The issue [https://github.com/pharo-project/pharo/issues/13512](https://deref-gmx.net/mail/client/lxTIyvDBkqs/dereferrer/?redirectUrl=https%3A%2F%2Fgithub.com%2Fpharo-project%2Fpharo%2Fissues%2F13512) has a list of all the intermediate steps done.

==========

CompiledMethodTrailer was removed in Pharo12

It was responsible for encoding and decoding various kinds of compiled method trailer data, but that had many downsides:

•	variable length trailers forced often called methods like CompiledMethod>>#endPC to instantiate an instance
•	setting a sourcePointer that required a different length trailer (e.g  embedding sources) needed to create a new compiled method and do a become
•	to avoid that, the compiler supported to use an existing trailer and delegated instance creation of CompiledMethod to it.
•	using 1 byte for the trailer kind lead to the need use Variable length encoding already for our ~40MB source file for some methods.


All in all, the CompiledMethodTrailer added a lot of complexity for nothing.

What do we do now?

•	use use a fixed-length trailer (see #trailerSize)
•	writing and reading that is not that complex (see #setSourcePointer, #sourcePointer:)

Q: I used to hand a trailer to the compiler, what should I do now?

The compiler now creates always CompiledMethods with a fixed, but empty trailer. Just use #sourcePointer to set it after.

Q: I used to clear the trailer using #copyWithTrailerBytes:, what should I do now?

There is now #cearSourcePointer. Just use #copy before if you want a copy.

Q: I want to have a non-installed method, but with source code.
The compiler now sets a #source property (and setting a source pointer via #sourcePointer: removes it).
If you use the compiler, you need to do nothing special anymore! All methods come with source via the property by default.

PR #13581, opened 2023-04-28
13468-Calypso-New-Trait-command-does-not-create-the-Trait

### LINKED ISSUE #13468: Calypso: New Trait command does not create the Trait
In Calypso:
- Right click on the class pane
- Click on New Trait
- Give the name you want
- Click on "OK"

I expected to get my new trait but nothing is happening

tesonep:
We can remove the isKindOf: message by using teh fluidInstall message that is implemented in Behavior and in the FluidBuilders.

Fix #13468


jecisc:
testExternalIDEDependencies now fails because there is a dependency added on FluidClassBuilder.

Maybe we should just update the test because it makes sense that the IDE depends on the class builder.

Ducasse:
Yes we should update the test.

MarcusDenker:
alternative fix was merged: https://github.com/pharo-project/pharo/pull/14088

PR #13582, opened 2023-04-28
Do not rely on #protocolNameOfElement: in Traits

jecisc:
#protocolNameOfElement: is a method containing a hack that I aim to remove. I updated some trait API and tests to use #protofolOfSelector: instead and keep pushing this hack removal.

Some bonus:
- categorize methods before adding them to the method dictionary to be sure we do not rely on the hack of #protocolNameOfElement: (this will help me with future improvements and should not change the behavior)
- Remove a useless #removeEmptyProtocols because nor #silentlyClassify:under: is taking care of removing the empty protocols
- Rename some category into protocol


jecisc:
Tests passed except windows slave

PR #13583, opened 2023-04-28
Simpler sourcePointer encoding

MarcusDenker:
Naively, I thought that we just shift the position and use the last bit to encode .changes vs .sources.

Let's try that

### DIFF
--- a/src/System-Sources/SourceFileArray.class.st
+++ b/src/System-Sources/SourceFileArray.class.st
@@ -156,10 +156,9 @@ SourceFileArray >> fileAt: index ifAbsent: aBlock [
 { #category : #'private - sourcepointer conversion' }
 SourceFileArray >> fileIndexFromSourcePointer: anInteger [
 	"Return the index of the source file which contains the source chunk addressed by anInteger"
+	"1 is sources file, 2 is changes file"
 
-	^ (anInteger allMask: 1)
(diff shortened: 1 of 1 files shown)


MarcusDenker:
Nice, just the tests for the old (very odd) encoding are failing. The new encoding is so trivial that we do not even need to test it like that.

I will do a another PR to skip the tests for now

src/System-Sources/SourceFileArray.class.st
```diff
@@ -157,23 +157,16 @@ SourceFileArray >> fileAt: index ifAbsent: aBlock [
 SourceFileArray >> fileIndexFromSourcePointer: anInteger [
 	"Return the index of the source file which contains the source chunk addressed by anInteger"
 
-	^ (anInteger bitAnd: 16r1000000) ~= 0
-		ifTrue: [ 1	"sources file" ]
-		ifFalse: [ anInteger >= 16r1000000
-				ifTrue: [ 2	"changes file" ]
-				ifFalse: [ 0	"compatibility with StandardSourceFileArray" ] ]
+	^ (anInteger allMask: 1)
```
MarcusDenker: this could be just 

^ (anInteger bitAnd: 1) + 1

MarcusDenker:
all green but linux, that is 

jecisc (approved):

PR #13588, opened 2023-04-28
Delegate #at:at: related methods to dictionary instance variable

### LINKED ISSUE #13587: OrderedDictionary DNU #errorKeyNotFound: - Implementation of #at:at:
**Bug description**
When sending #at:at: to an instance of OrderedDictionary, it signals a MessageNotUnderstood error, telling me it does not understand #errorKeyNotFound:

**To reproduce**
Open a playground and evaluate the following:
```Smalltalk
OrderedDictionary new at: #key1 at: #key2
```

**Version information**
All known versions, i.e. Pharo versions 7, 8, 9, 10, 11 etc.

**Context**
OrderedDictionary subclasses Collection (not Dictionary). It holds a dictionary instance variable, which it delegates to, for example in the case of #at:, #at:ifAbsent: etc. However, the implementation of #at:at: does not delegate to its dictionary instance variable. Instead, it has the same method implementation that Dictionary has, which sends #errorKeyNotFound: to self in case one of the keys is not found. However, the method #errorKeyNotFound: is not implemented on OrderedDictionary.

**Potential solutions**
The simplest thing that could work is to implement the method #errorKeyNotFound: on OrderedDictionary, but then I think we'd miss an opportunity for improvement.

However, I think OrderedDictionary has incorrectly implemented the methods #at:at:, #at:at:ifAbsent:, #at:at:ifAbsentPut: and #at:at:put:. They should all delegate to the dictionary instance variable instead. The use of #species in #at:at:ifAbsentPut: and #at:at:put: ensures the correct class of the value for the first key.

A downside of refactoring the #at:at: methods is that the error message text is 'key XYZ not found in Dictionary', instead of referencing the class name OrderedDictionary. I think this is an acceptable implementation consequence.

jvalteren:
Fixes #13587. 

Includes simple test case that triggered bug in previous implementation.

### DIFF
--- a/src/Collections-Sequenceable/OrderedDictionary.class.st
+++ b/src/Collections-Sequenceable/OrderedDictionary.class.st
@@ -171,44 +171,22 @@ OrderedDictionary >> at: aKey [
 
 { #category : #'nested dictionaries' }
 OrderedDictionary >> at: firstKey at: secondKey [
-	"Return the object stored in the second dictionary at secondKey. The second dictionary is accessed via the key firstKey."
-
-	"
-	(Dictionary new
-		at: #top at: #below1 put: 1;
-		at: #top at: #below1 put: 2;
-		at: #top at: #below1)
-	>>>
-	2"
-
-	^ self at: firstKey at: secondKey ifAbsent: [ self errorKeyNotFound: secondKey ]
+	^ dictionary at: firstKey at: secondKey
 ]
 
 { #category : #'nested dictionaries' }
 OrderedDictionary >> at: firstKey at: secondKey ifAbsent: aZeroArgBlock [
-		"Return the object stored in the second dictionary at secondKey. The second dictionary is accessed via the key firstKey. Execute aZeroArgBlock in case one of the key is wrong."
-		
-	| subDictionary |
-	subDictionary := self at: firstKey ifAbsent: [ ^ aZeroArgBlock value ].
-	^ subDictionary at: secondKey ifAbsent: aZeroArgBlock
+	^ dictionary at: firstKey at: secondKey ifAbsent: aZeroArgBlock
 ]
 
 { #category : #'nested dictionaries' }
 OrderedDictionary >> at: firstKey at: secondKey ifAbsentPut: aZeroArgBlock [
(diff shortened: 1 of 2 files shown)


svenvc:
LGTM

jvalteren:
Thanks @svenvc! See https://github.com/svenvc/NeoJSON/issues/18 as well ;-)

src/Collections-Sequenceable/OrderedDictionary.class.st
```diff
@@ -171,44 +171,22 @@ OrderedDictionary >> at: aKey [
 
 { #category : #'nested dictionaries' }
 OrderedDictionary >> at: firstKey at: secondKey [
-	"Return the object stored in the second dictionary at secondKey. The second dictionary is accessed via the key firstKey."
```
Ducasse, line 174: Could you keep the comment and the executable comment.

Ducasse:
Can you rescue the method comments for each of the methods you changed?
We want a strongly documented system and comments are important.

src/Collections-Sequenceable/OrderedDictionary.class.st
jvalteren: Hi @Ducasse, thanks for your feedback (sorry for my late reply). None of the methods in OrderedDictionary that delegate to the dictionary instance variable have these comments, so I chose to remove them for consistency.
As I see it, they are implementation comments and, since the new implementation delegates to an instance of  dictionary, I feel that the comments should not be retained.

Please take a look at methods like #at:ifAbsent:, #includes:, #keyAtValue: etc. None of them have these comments, while their implementation in Dictionary does.

What do you think?

jvalteren:
> Can you rescue the method comments for each of the methods you changed? We want a strongly documented system and comments are important.

I am in favor of a strongly documented system too! But IMO the context is a bit different here. Please see my reply to your review comment above.

Ducasse:
Well I prefer to have more comments than less. I will add it back.

Ducasse:
I redid the PR plus fixed more methods to be consistent, added more tests and rescued the method comments. 
So I will close this PR

PR #13589, opened 2023-04-28
Fix/filter

guillep:
Second iteration on fix #13214.

Nowadays, the fast table filter works the first time.
And as soon as the filter text box loses focus it does not work anymore.
This is because the system is disabling text events, and they are re-enabled when a text editor regains focus.


guillep:
Everything green except windows that fails

PR #13590, opened 2023-04-28
Clean blocks for announcement

### LINKED ISSUE #11220: [CleanBlocks] Announcements defined via # when:do:

PalumboN:
_Come from PR https://github.com/pharo-project/pharo/pull/13270_
Fix https://github.com/pharo-project/pharo/issues/11220

The current Announcement API assumes as the subscriber _the object where the action block is created_. 

https://github.com/pharo-project/pharo/blob/d9f862003a88bfc9a17ff12ab7a11abf1fe154a9/src/Announcements-Core/AnnouncementSubscription.class.st#L99-L104

This has two main issues:
1. Could exist cases where the action block is not created in the "subscriber object" (some utils method chain).
2. It does **not work with blocks without context** (like clean or constant blocks) 😢 

So, I changed the Announcement API to send the subscriber object explicitly. Changing all
```
when: event do: action
```
by
```
when: event do: action for: subscriber
```
using a smart deprecation rewrite 👀

### DIFF
--- a/doc/Announcements/Announcements.md
+++ b/doc/Announcements/Announcements.md
@@ -32,17 +32,17 @@ Announcement subclass: #GameLostAnnouncement
 	   
 ### Step 2 - Publishers and subscribers
 
-If an object wants to announce an event it needs someone to make the announcement to. This is typically an instance of class `Announcer` which acts as the mediator between the object that has to announce something (publisher) and one or many (anonymous) subscribers who are interested in the event.
+If an object wants to announce an event it needs someone to make the announcement to. This is typically an instance of class `Announcer` which acts as the mediator between the object that has to announce something (publisher) and one or many (subscriber) subscribers who are interested in the event.
 
 ```st
 | announcer |
 announcer := Announcer new.
-announcer announce: MyInterestingAnnouncement new
 ```
 
-Using `announce:` we can make an announcement - but since nobody is interested yet nothing will happen.
+#### Example 1 - Sending a message to the subscriber
 
-Lets add some consumers/subscribers. Subscribers just register on the Announcer instance to note that they are interested on a particular event/announcement: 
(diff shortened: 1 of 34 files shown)


jecisc (approved):

Ducasse:
We should deprecate when:do: if it is not available anymore.

PalumboN:
Yes! Now it is deprecated -> https://github.com/pharo-project/pharo/pull/13590/files#diff-f41aaf7ebd75a6b03d3bd3f2c4089d9a48bfc7f4363932b87c882c8d0c9bf40b

<blockquote><div> GitHub</div><div><strong><a href="https://github.com/pharo-project/pharo/pull/13590">Clean blocks for announcement by PalumboN · Pull Request #13590 · pharo-project/pharo</a></strong></div><div>Come from PR #13270
Fix #11220
The current Announcement API assumes as the subscriber the object where the action block is created.

  
    
      pharo/src/Announcements-Core/AnnouncementSubscript...</div></blockquote>

MarcusDenker (approved):

Issue #10184, opened 2021-10-18
Improve CompiledMethod>>#equivalentTo:


MarcusDenker:
CompiledMethod>>#equivalentTo: uses the AST to compare the methods. The problem with that is that methods with just different variable names are not seen as equivalent

*** What to do ***

We should implement #equivalentTo: like #= by comparing bytecode and literals and pragmas, but ignoring properties and class binding.

PR #13595, opened 2023-04-28
Add test on CompiledCode>>#equivalentTo:

jecisc:
See title


MarcusDenker:
this method has problems... see #10184

MarcusDenker (approved):
the things tested here are ok, so good to merge (the issue with the variable names is a different thing and stays open)

PR #13604, opened 2023-05-01
Simplify classify:under:

jecisc:
This is one of the last method of ClassOrganization to improve to manage nils well and deal with real protocols.  I already tried to improve it but it causes troubles in the release tests. Here is a simple simplifcation of the method as a first step to improve it.


jecisc:
Hum... This breaks some tests which is weird because I though I didn't change the logic.

MarcusDenker (approved):

PR #13606, opened 2023-05-01
Compiler: get requestor

privat:
#13521 was merged too fast (and reverted by  #13573) and cause regressions, so this is a first part that takes care of the requestor


jecisc:
Like in the last PR the Reflectivity tests are broken so I'll put a "Need more work" label for now

privat:
Re-reroll into a smaller subset

jecisc (approved):

privat:
Windows CI is broken, others are green

MarcusDenker (approved):

PR #13615, opened 2023-05-02
Make protocol classification a setting

### LINKED ISSUE #13597: automatic protocol categoriser does weird stuff
**Bug description**
it autocategorise methods even if I am creating it inside a category)

**To Reproduce**
Calipso, select a protocol (e.g. private)
type a method that would be categorised if not in protocol, e.g. `isATest`.
See how it puts the method in `adding` instead let it in `private`.

**Expected behavior**
If I add a method in an existing category, I want it to stay there

Pharo 11

jecisc:
As esteban asked, now the automatic classification is a setting and the old behavior is brought back by default.

Fixes  #13597

### DIFF
--- a/src/Tool-Base/MethodClassifier.class.st
+++ b/src/Tool-Base/MethodClassifier.class.st
@@ -10,6 +10,9 @@ Class {
 	#instVars : [
 		'protocol'
 	],
+	#classVars : [
+		'ShouldAlwaysClassify'
+	],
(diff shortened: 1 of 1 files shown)


jecisc:
@estebanlm This is the change of behavior you asked me, I targeted P11

estebanlm (approved):

estebanlm:
thanks!

PR #13618, opened 2023-05-02
First step at class name deprecation

### LINKED ISSUE #13365: Feature: deprecated class names
Deprecated classes have a `isDeprecated` class-side method (that defaults to the package, so moving a class to a deprecated package is enough), but there is no facility to deprecate just a class name. i.e no feature change, instances stay the same, etc.

A trick can be to have a deprecated named class as a subclass of a new named class, but there are still 2 classes that might cause issues:

* new will instantiate the bad one (so class methods must be added to finish the trick).
* `isKindOf: OldName` fails on the new named class (but the message is questionable).
* `on: OldName do:` won't catch new named exceptions.
* both class are not really linked together (it's an information that tools can use)

In the case of pure class renaming, I propose to just have additional entries in the global dictionary.
Such deprecated names can be identified when `assoc key ~= assoc value name` and styled accordingly.
Rules and automatic code rewritings (à la `#deprecated:transformWith:when:`) can also be added.

To install those names, the `initialize` class-side method could be used.
Maybe an additional property of the *fluid* class syntax can also be proposed. eg `deprecatedAliases`.

```st
Object << #MyClass
	slots: {};
	sharedVariables: {};
	deprecatedAliases: { #MyOldClass };
	package: ''
```

I'm not sure if there are other implications

jecisc:
Here is a first step at class name deprecation. 
We can now use the method ClassDescription>>deprecatedAliases: to set some deprecated aliases to a class. It will add a deprecated global variable to the SystemDictionary that can be referenced but that will be strikedthrought. 

Currently you can set this method in a class initialization.

I see a few next steps:
- Make Tonel export the information
- Add this to FluidClassBuilder so that we do not need to define it in the class initialization
- Warn users on accesses (This is for all deprecated global variables)
- Add a way to automatically rename the class references 

Fixes #13365


jecisc:
I'm getting the weird error :(

```
Warning: Package Kernel-Processes depends on the following classes:
  nil
  Abort
  AbstractDelayTicker
  AbstractDelayTicker>>#isAbstract
  AbstractDelayTicker>>#millisecondsUntilTick:
  AbstractDelayTicker>>#nowTick
  AbstractDelayTicker>>#restoreResumptionTimes:
  AbstractDelayTicker>>#saveResumptionTimes:
  AbstractDelayTicker>>#tickAfterMilliseconds:
  AbstractDelayTicker>>#waitForUserSignalled:orExpired:
  AbstractLayout
  AbstractLayout>>#isAbstract
  AbstractLayout>>#=
  AbstractLayout>>#allSlots
  AbstractLayout>>#allSlotsDo:
  AbstractLayout>>#allVisibleSlots
  AbstractLayout>>#checkIntegrity
  AbstractLayout>>#checkSanity
  AbstractLayout>>#definesSlot:
  AbstractLayout>>#extend:
  AbstractLayout>>#extendByte
  AbstractLayout>>#extendDoubleByte
  AbstractLayout>>#extendDoubleWord
  AbstractLayout>>#extendImmediate
  AbstractLayout>>#extendVariable:
...
```

privat:
Nice.

Is it not clear when the new method should be called (nor by who).
And because the feature will have possible complex interaction with class-related meta-operations (installing, removing, renaming, etc.), maybe start to add some tests.

jecisc:
> Nice.
> 
> Is it not clear when the new method should be called (nor by who). And because the feature will have possible complex interaction with class-related meta-operations (installing, removing, renaming, etc.), maybe start to add some tests.

I added some tests and I handled two cases:
- Aliasing to an existing global now raises a warning (except in the global is the actual class, that means that we probably are just reaplying the aliasing)
- Removing a class should remove the aliases from the globals

jecisc:
Also thank you for reminding me not to be lazy on tests :)

src/Kernel/Class.class.st
```diff
@@ -870,21 +870,23 @@ Class >> removeFromSystem: logged [
...
+	self superclass ifNotNil: [ "If we have no superclass there's nothing to be remembered" self superclass addObsoleteSubclass: self ].
 
 	"we add the class to Undeclared so that if references still exist, they will  be automatically fixed
 	if this class is loaded again. We do not check if references exist as it is too slow"
 	Undeclared declare: self name asSymbol from: Smalltalk globals.
 	myCategory := self category.
 	self environment forgetClass: self logged: logged.
+
+	"In case the class has deprecated aliases we need to remove them from the system dictionary."
+	self deprecatedAliases do: [ :alias | self environment removeKey: alias ifAbsent: [  ] ].
```
privat: I'm not sure how strong the name ownership is (or should be). Maybe remove the alias from the environment only if it still refer to the class to be removed?

privat (approved):

src/Kernel/Class.class.st
jecisc: Oh I see what you mean. I'll add new assertions about that and manage this case.

jecisc: I'm checking that but I see that once we have a global variable installed, if we create a class of the same name, it is not been registered to the system dictionary, 

This seems to big to treat in this specific PR IMO.

jecisc: I'll still do a new commit that manages the case tho

jecisc:
Tests finally passed!

PR #13619, opened 2023-05-02
Remove usage of #protocolNameOfElement: in Monticello

jecisc:
This method contains a hack that should be removed. This PR has for goal to see if the hack is exploited in Monticello or not. 

If this PR is green then we will be able to reduce the number of clients of this method that will be removed in the future

### DIFF
--- a/src/Monticello/MethodAddition.class.st
+++ b/src/Monticello/MethodAddition.class.st
@@ -44,11 +44,14 @@ MethodAddition >> createCompiledMethod [
 	"CyrilFerlicot: Why do we need to explicitly write to the source? Why do we call the compiler?
 	
 	Can't we just call #compile:classfied: here?"
-	compiledMethod := myClass compiler permitUndeclared: true; compile: text asString.
+
+	compiledMethod := myClass compiler
+		                  permitUndeclared: true;
+		                  compile: text asString.
 	selector := compiledMethod selector.
 	self writeSourceToLog.
 	priorMethodOrNil := myClass compiledMethodAt: selector ifAbsent: [ nil ].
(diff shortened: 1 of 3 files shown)


src/Monticello/MethodAddition.class.st
```diff
@@ -44,11 +44,14 @@ MethodAddition >> createCompiledMethod [
...
-	compiledMethod := myClass compiler permitUndeclared: true; compile: text asString.
+
+	compiledMethod := myClass compiler
+		                  permitUndeclared: true;
+		                  compile: text asString.
 	selector := compiledMethod selector.
 	self writeSourceToLog.
 	priorMethodOrNil := myClass compiledMethodAt: selector ifAbsent: [ nil ].
-	priorCategoryOrNil := myClass organization protocolNameOfElement: selector
+	priorCategoryOrNil := (myClass organization protocolOfSelector: selector) ifNotNil: [ :protocol | protocol name ]
```
privat, line 54: When I read this diff-line, I'm second guessing if having a helper method `protocolNameOf` that takes care of nil is not that bad.

jecisc: Well, ideally I want to iterate further to use real protocol in most part of the system and not manipulate the protocol name. I'm just doing it step by step and for now I want to kill the users of the hacked #protocolNameOfElement:  

I wanted to check once I would use real protocols in most places if we still had a lot of places that needed to deal with protocol names instead of protocols to see if it makes sense to introduce a #protocolNameOfSelector: that does not have the hack that #protocolNameOfElement: has

jecisc:
The hack is removed now. I think we can close this one. I'll check if it is better to manipulate a real protocol here later.

PR #13621, opened 2023-05-02
Do not reclassify already classified method

jecisc:
Currently the ClassOrganization has a guard to not reclassify a method if it is already in a protocol and the old and new protocol are the unclassified one.  I don't know why we would do that only in the case a protocol is unclassified. IMO as soon as the old and new protocols are the same we should get in the guard clause.

Doing tiny PR because this part of the system is currently really fragile.

### DIFF
--- a/src/Kernel-Tests/ClassOrganizationTest.class.st
+++ b/src/Kernel-Tests/ClassOrganizationTest.class.st
@@ -85,6 +85,39 @@ ClassOrganizationTest >> testClassifyUnder [
 ]
 
 { #category : #tests }
+ClassOrganizationTest >> testClassifyUnderUnclassified [
+	"Ensure unclassified is acting as any other protocol because that was not the case in the past."
+
+	"Set the base for the test"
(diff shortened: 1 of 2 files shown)


privat (commented):
This changes the behavior if one asks to classify a method to unclassified? Maybe a test is missing.

jecisc:
For classification, the unclassified protocol should behave like any other protocol IMO. 

If you want to classify an unclassified method, it's the same as classifying a method that was already in another protocol. 
If you want to unclassify a classified method, it's the same as changing the protocol of a method

And ideally I want to remove most of the hacks that make ClassOrganization have special treatments for unclassified. Ideally it should be a protocol like any other. It's just that it's the default value when a new method is created without a protocol. 

And we have tests for protocol classification, but I can add new ones to ensure the behavior is the same.

privat:
I agree that unclassified should be a protocol that behave like the others, but this was not the case before, so adding tests may help the new consistent behavior to be more robust

privat (approved):

PR #13624, opened 2023-05-02
Simplify code export

jecisc:
Some cleanup and removing some calls to a method that will be removed


jecisc:
The test failure is known. Tests passed.

MarcusDenker (approved):

PR #13625, opened 2023-05-02
Code Completion: multiple fixes on replacements annoying bugs

### LINKED ISSUE #9329: CC - Cursor backwards and then autocomplete, there is a bug
when you move the cursor backwards and then autocomplete, there is a bug

privat:
Code replacement was broken if the old string what shorter than the new one and was not a prefix.
I assume it was some old hard-coded expectation on completion, but new completion engines are better.

e.g.

1. write `a foo`
2. completion menu open
3. go back 1 character
4. chose `fontNumber`
5. old behavior: you get `a footNumber` (a possible valid selector, but not the same concern)
5. new fixed behavior: you get `a fontNumber`

The second fix is related to "Put a space after completion" setting (that is true by default) that added a space even it there was already a space. Forcing the programmer to remove one

e.g.
1. write `a fo bar`
2. go back at the end of `fo` and add a `n`
3. completion menu open
4. chose `fontNumber`
5. old behavior: you get `a fontNumber  bar` (with two spaces and the cursor in the middle of them)
5. new behavior: you get `a fontNumber bar` (with a single space and the cursor after it)

Update: fixed a third bug. The completion was broken when completing at the 2nd position. i.e. between the first and the second characters of the word, i.e. with a single character prefix.
Both completion and the replacement were broken because of two additional (but broken) `- 1` when calling `previousWord:`. With a single character prefix, the prefix started at the previous word (including possible spaces and newlines), preventing the completion heuristic to find something meaningful and making replacements hazardous.

e.g.
1. write `a f`
2. press tab
3. previous behavior: you get no completion suggestions
3. new fixed behavior: you get plenty of completion suggestions

e.g.
1. write `a fx`
2. completion menu open
3. go back 1 character
4. chose `fontNumber`
5. previous behavior: you get `a fxNumber` (yes `nt` vanished)
5. previous behavior without the update: you get `fontNumber` (yes `a ` vanished)
6. new fixed behavior after the update: you get `a fontNumber`

Fix #9329 I assume

### DIFF
--- a/src/NECompletion-Morphic/CompletionEngine.class.st
+++ b/src/NECompletion-Morphic/CompletionEngine.class.st
@@ -326,6 +326,10 @@ CompletionEngine >> replaceTokenInEditorWith: aString [
 	"Try to correctly replace the current token in the word.
 	The code editor should be able to do it by himself"
 	wordEnd := (self editor nextWord: self editor caret - 1).
+	"Do not insert an aditional space if there is already one"
+	(aString last = $  and: [
+		wordEnd <= self editor text size and: [
+			(self editor text at: wordEnd) = $ ]]) ifTrue: [ wordEnd := wordEnd + 1 ].
 	"If the returned index is the size of the text that means that the caret is at the end of the text and there is no more word after, so add 1 to the index to be out of range to select the entierely word because of the selectInvisiblyFrom:to: remove 1 just after to be at the end of then final word"
 	wordEnd > self editor text size ifTrue:[ wordEnd := wordEnd + 1 ].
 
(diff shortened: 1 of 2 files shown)


jecisc:
Thank you for that! A few completion issues are bugging me since a while and those two are part of it!

privat:
Please open issues for the other ones that still bother you

privat:
Update: fixed a third bug (see updated description)

svenvc:
Thanks for taking care, several aspect of completion in Pharo 11 were quite annoying.

jecisc:
Maybe this could go in Pharo11 if it does not rely on recent improvements of the system?

Ducasse:
Yes thanks for me I got annoyed often but I jumped over everything (now I will record it to be able to report). 
See my comment in the other issues with keywords based completion.

Ducasse:
Oh and there are tests!!!! So cool. I remember when we wrote the first ecompletion tests with guille because we wanted to understand clearly the issue. This is really nice to see other tests!!!!

Issue #13315, opened 2023-04-06
Fluid: Check skipped tests #ChangeSetClassChangesTest


MarcusDenker:
**Bug description**
If we enable Fluid by default, we have odd tests failing, but only when running all tests:

```
Error
Got 'Object << #ClassForTestToBeDeleted246  slots: {};  package: ''CategoryForTestToBeDeleted-TestPackage1''' instead of 'Object << #ClassForTestToBeDeleted246  slots: {};  package: ''TestPackage2'''.
Stacktrace
TestFailure
Got 'Object << #ClassForTestToBeDeleted246
	slots: {};
	package: ''CategoryForTestToBeDeleted-TestPackage1''' instead of 'Object << #ClassForTestToBeDeleted246
	slots: {};
	package: ''TestPackage2'''.
ChangeSetClassChangesTest(TestAsserter)>>assert:description:resumable:
ChangeSetClassChangesTest(TestAsserter)>>assert:description:
ChangeSetClassChangesTest(TestAsserter)>>assert:equals:
ChangeSetClassChangesTest>>assertDefinition:equivalentTo:
ChangeSetClassChangesTest>>testChangeClassCategory
ChangeSetClassChangesTest(TestCase)>>performTest



Error
Got 'Object << #ClassForTestToBeDeleted247  slots: {};  package: ''CategoryForTestToBeDeleted-TestPackage1''' instead of 'Object << #ClassForTestToBeDeleted247  slots: {};  package: ''TestPackage2'''.
Stacktrace
TestFailure
Got 'Object << #ClassForTestToBeDeleted247
	slots: {};
	package: ''CategoryForTestToBeDeleted-TestPackage1''' instead of 'Object << #ClassForTestToBeDeleted247
	slots: {};
	package: ''TestPackage2'''.
ChangeSetClassChangesTest(TestAsserter)>>assert:description:resumable:
ChangeSetClassChangesTest(TestAsserter)>>assert:description:
ChangeSetClassChangesTest(TestAsserter)>>assert:equals:
ChangeSetClassChangesTest>>assertDefinition:equivalentTo:
ChangeSetClassChangesTest>>testChangeClassCategoryAddsNewChangeRecord
ChangeSetClassChangesTest(TestCase)>>performTest
```

PR #13628, opened 2023-05-03
set again fluid to be enabled by default

MarcusDenker:
...  to check which tests fail

### DIFF
--- a/src/Kernel/ClassDefinitionPrinter.class.st
+++ b/src/Kernel/ClassDefinitionPrinter.class.st
@@ -161,7 +161,7 @@ Object << #Faked
 { #category : #configure }
 ClassDefinitionPrinter class >> showFluidClassDefinition [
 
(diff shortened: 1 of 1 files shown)


jecisc:
The failing tests are the one in ChangeSetClassChangesTest and IIRC they are random failures

Ducasse:
This strange that the package is not correct.

Ducasse:
```
'Object << #ClassForTestToBeDeleted249  slots: {};  package: ''CategoryForTestToBeDeleted-TestPackage1''' instead of 'Object << #ClassForTestToBeDeleted249  slots: {};  package: ''TestPackage2'''.
```

MarcusDenker:
Yes, this has to do with old class defintion printing the category, while fluid prints the package name. 

Maybe AbstractEnvironmentTestCase does something wrong, There is https://github.com/pharo-project/pharo/issues/13315

PR #13635, opened 2023-05-03
Ensure #addSelector:withMethod: classifies methods

jecisc:
This method seems to be the last place were methods are added without been added to a protocol. 

Since it does not classify the method we just add it to Protocol unclassified to be sure it is at least in one protocol.


jecisc:
Test failure does not seems related

PR #13637, opened 2023-05-03
Complete super message send

### LINKED ISSUE #13054: [Feature Request] Code completion on super
There is basically a single selector (and set of arguments) that make sense on a super call.
This could be an easy task.

privat:
On super send, there is one preferred selector, the current selector (see ReSendsDifferentSuperRule).
So this PR teach completion to propose a full message send, including the arguments.

Fix #13054

See it in action: I redefine `printOn:` in some class then write `super p` and a popup appear with the first choice

Capture d’écran du 2023-05-03 15-15-15

Then I just press enter, et voilà!

Capture d’écran du 2023-05-03 15-24-29


Ducasse:
oh I love it :)

jecisc (approved):
Yes! Maybe worth a test to be sure we do not lose this enhancement in the future. But it will be great for overriding methods!

PR #13638, opened 2023-05-03
Completion complete empty tokens

### LINKED ISSUE #9327: CC - Add placeholders in the text editor (FP)
Add placeholders in the text editor to facilitate the edition of keyword messages

privat:
This is the second part of the journey. In one big PR. The first part was #13636

A big PR because there are some changes (and fixes) in various parts of the system, but that don't have much sense alone.
If reviewers think that is too much, I can move cleaning and refactoring in a distinct *prior* PR.

The first 3 commits are about `RubSmalltalkEditor>>#atCompletionPosition` that extends the completion point to anywhere in the code except in the indentation space (between the beginning of the line and the first non separator character). So `tab` will try to complete almost anywhere — the important word here is *try* :)

The following commits are about corresponding the caret position to some AST node. Some methods already exists (like `RBProgramNode>>#nodeForOffset:`) but focuses on a precise position. But code completion is related to what is before the cursor. So I introduce and use `RBProgramNode >> nodeBeforeOffset:`.

The magic commit is *CoASTHeuristicsResultSetBuilder>>#visitValueNode:  add the possibility to complete a new message send after any value*.
What it does is detecting the situation of an empty token completion after a value node, then tries to complete a message send from scratch with this node as a receiver. There is some magic involved, but because of the complexity of CoCompletionEngine I did not manage to find a cleaner way.

The rest of the commits are small bugfixes that I added after playing with it for a while.

OK. But is all this useful?

yes in some cases...

For `foo between: bar |`, where `|` is the caret, if you press *tab*, the completion menu opens and proposes `and:` at the first choice.

Capture d’écran du 2023-05-03 17-42-44

`Point |` proposes `x:y:` at the first choice.

Capture d’écran du 2023-05-03 17-45-37

`printOn: aStream   super |` proposes the full message send `printOn: aStream` at the first choice (require #13637 )

Capture d’écran du 2023-05-03 18-09-02

and less in other cases because existing heuristics are sometime very poor, thus asking for a completion on an empty prefix can give a lot of unfiltered garbage.
So heuristics could be improved anyway (or just rewrite from scratch because I was forced to hack it and I don't like the code at all).

The GUI need also some love (and tweaks), the completion menu closes itself less often (because current closing rely on invalid completion positions, but now most completion positions are valid) and I noticed some minor graphical glitches (tool still functional). I fixed all runtime errors I noticed when trying to complete weird things at weird places, so it should just work for you (or at least not crash)

About tests. ~~There are no tests on the completion part~~ because I don't understand the testing framework of the completion engine. I will try to add at least some smoke tests to ensure that completion menu open and don't crash on various kinds of empty tokens, maybe by extending the ones I added in #13636.
**Update:** I added an extensive smoke test with code snippets, this forces to try to get completion results on corner case source codes. It allowed me to spot and fix another bug.

Should fix #9327

### DIFF
--- a/src/AST-Core/RBEnglobingErrorNode.class.st
+++ b/src/AST-Core/RBEnglobingErrorNode.class.st
@@ -105,6 +105,15 @@ RBEnglobingErrorNode >> isEnglobingError [
 	^true
 ]
 
+{ #category : #replacing }
+RBEnglobingErrorNode >> replaceNode: aNode withNode: anotherNode [
+
+	self contents: (contents collect: [ :each |
+			 each == aNode
+				 ifTrue: [ anotherNode ]
+				 ifFalse: [ each ] ])
+]
+
 { #category : #accessing }
 RBEnglobingErrorNode >> stop: anInterger [
 	stop := anInterger
--- a/src/AST-Core/RBMessageNode.class.st
+++ b/src/AST-Core/RBMessageNode.class.st
@@ -365,7 +365,7 @@ RBMessageNode >> replaceNode: aNode withNode: anotherNode [
 		ifTrue:
 			[self receiver: anotherNode.
 			(parent notNil and: [parent isCascade])
-				ifTrue: [parent messages do: [:each | each receiver: anotherNode]]].
+				ifTrue: [parent messages do: [:each | each isMessage ifTrue: [ each receiver: anotherNode]]]].
 	self arguments: (arguments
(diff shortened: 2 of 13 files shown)


privat:
CI failure caused by a change in the GUI behavior (the completion auto closes itself less often, see description for details).
Mark the test as *expectefFailure* until we know what behavior we want.

MarcusDenker (approved):

PR #13640, opened 2023-05-04
CommentSourcePointer-move-methods-down 

MarcusDenker:
to prepare moving the commentSourcePointer ivar down to class (as it is nil for MetaClass), this PR

- adds forwarding methods to MetaClass
- move methods down to Class

### DIFF
--- a/src/Kernel/Class.class.st
+++ b/src/Kernel/Class.class.st
@@ -461,7 +461,7 @@ Class >> comment [
 Class >> comment: aStringOrText [
 	"Set the receiver's comment to be the argument, aStringOrText."
 
-	self instanceSide classComment: aStringOrText
+	self classComment: aStringOrText
 ]
 
 { #category : #'accessing - comment' }
--- a/src/Kernel/Class.class.st
+++ b/src/Kernel/Class.class.st
@@ -313,6 +313,50 @@ Class >> classBuilder [
 		^ self classInstaller new builder
 ]
 
+{ #category : #'accessing - comment' }
+Class >> classComment: aString [
+	"Store the comment, aString or Text or RemoteString, associated with the class we are orgainzing.  Empty string gets stored only if had a non-empty one before."
+	^ self comment: aString stamp: '<historical>'
+]
+
+{ #category : #'accessing - comment' }
+Class >> classComment: aString stamp: aStamp [
+	"Store the comment, aString or Text or RemoteString, associated with the class we are organizing.  Empty string gets stored only if had a non-empty one before."
+
+	| pointer header oldComment oldStamp preamble |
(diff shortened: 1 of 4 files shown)


src/Kernel/Class.class.st
```diff
@@ -313,6 +313,50 @@ Class >> classBuilder [
...
+Class >> classComment: aString stamp: aStamp [
+	"Store the comment, aString or Text or RemoteString, associated with the class we are organizing.  Empty string gets stored only if had a non-empty one before."
+
+	| pointer header oldComment oldStamp preamble |
+	oldComment := self comment.
+	oldStamp := self commentStamp.
+
+	aString string = oldComment string ifTrue: [ ^ self ].
+
+	pointer := self instanceSide commentSourcePointer ifNil: [0].
```
jecisc:
```suggestion
	pointer := self commentSourcePointer ifNil: [0].
```

IIUC the goal, here it should always be equivalent to self?

src/Kernel/Class.class.st
```diff
@@ -313,6 +313,50 @@ Class >> classBuilder [
...
+				nextPutAll: ' commentStamp: '.
+			aStamp storeOn: strm.
+			strm nextPutAll: ' prior: '; nextPutAll: pointer printString ].
+		file nextChunkPut: header; cr ].
+
+	SourceFiles
+		writeSource: aString
+		preamble: preamble
+		onSuccess: [ :newSourcePointer |
+			self instanceSide commentSourcePointer: newSourcePointer]
```
jecisc:
```suggestion
			self commentSourcePointer: newSourcePointer]
```

Same

src/Kernel/Class.class.st
```diff
@@ -406,6 +450,45 @@ Class >> classVariablesNeedFullDefinition [
...
+    ^ self commentSourcePointer
+		  ifNil: [ '' ]
+		  ifNotNil: [ :ptr | SourceFiles sourceCodeAt: ptr ]
+]
+
+{ #category : #'accessing - comment' }
+Class >> comment: aStringOrText [
+	"Set the receiver's comment to be the argument, aStringOrText."
+
+	self instanceSide classComment: aStringOrText
```
jecisc:
```suggestion
	self classComment: aStringOrText
```

jecisc (approved):

jecisc:
The global changes is good. I suggested some code cleanings as a bonus

src/Kernel/Class.class.st
MarcusDenker: yes, I did not look at the methods yet as they should be deprecated in a next step, but of course we can do this change now already

PR #13642, opened 2023-05-04
Small cleanup on protocol cleaning

jecisc:
- cleanUpProtocolsForClass: is always called with the organized class as parameter ==> Remove the parameter
- The name is not really explicit. What cleanup? I propose removeUnexistingSelectorsFromProtocols
- The method had no test which is now fixed 

Bonus: ClassOrganizationTest should use the actual organization of the class and not create a duplicated one to keep coherence in the tests.

### DIFF
--- a/src/Kernel/ClassOrganization.class.st
+++ b/src/Kernel/ClassOrganization.class.st
@@ -55,13 +55,6 @@ ClassOrganization >> classify: selector under: aProtocol [
 	oldProtocol ifNotNil: [ self notifyOfChangedSelector: selector from: oldProtocol name to: newProtocol name ]
 ]
 
-{ #category : #cleanup }
-ClassOrganization >> cleanUpProtocolsForClass: aClass [
-	"remove all entries that have no methods"
-
-	self allMethodSelectors do: [ :each | (aClass includesSelector: each) ifFalse: [ self removeElement: each ] ]
-]
-
 { #category : #copying }
 ClassOrganization >> copyFrom: otherOrganization [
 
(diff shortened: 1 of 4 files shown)


src/Kernel/ClassOrganization.class.st
```diff
@@ -232,6 +225,15 @@ ClassOrganization >> removeProtocolIfEmpty: aProtocol [
 	self notifyOfChangedProtocolNamesFrom: oldProtocolNames to: self protocolNames
 ]
 
+{ #category : #cleanup }
+ClassOrganization >> removeUnexistingSelectorsFromProtocols [
+	"For each protocol, remove the selectors that are not present in the class."
```
MarcusDenker: "Unexisting" in english sounds strange... but it exists: "Adjective. unexisting (not comparable) Not in existence."

MarcusDenker (approved):

MarcusDenker:
There are now conflicts

jecisc:
I'll ask some native english speaker friends if they can think about a better name

MarcusDenker:
the alternative is "nonexistent" https://www.merriam-webster.com/dictionary/unexisting

But I am fine as long as it has a dictionary entry

<blockquote><div><strong><a href="https://www.merriam-webster.com/dictionary/unexisting">Definition of UNEXISTING</a></strong></div><div>nonexistent… See the full definition</div></blockquote>

jecisc:
Native English speaker says that nonexistent is a better word indeed. I'll resolve the conflict and rename

jecisc:
I renamed to use nonexistent

jecisc:
I forgot to rename the tests and I don't have the time now. I'll do that later

MarcusDenker (approved):

PR #13643, opened 2023-05-04
move commentSourcePointer from ClassDescription to Class

MarcusDenker:
move source pointer down


jecisc (approved):

MarcusDenker:
BehaviorTest>>testInstSize is failing (as it should with one ivar less for Metaclass)

I propose to merge this manually created PR and fix the test next

PR #13647, opened 2023-05-04
Completion autoclose menu

privat:
Follow up of  #13638
This reproduces the previous behavior of the completion menu that closes itself when "exiting" a word (with left and right arrows, for instance).
And it also adds a test because this behavior was not tested.

Previous behavior used `atCompletionPosition` to close the menu, but that does not work anymore because we almost always at a completion position now.
To reproduces the behavior, we need to check that the current completion context is still valid for the current caret. To do that, I just store the start of the completion token in the completion context as an identifier. If the start changes, it means that the current context is not appliable, so the menu is closed.

The difference with the previous behavior is the start of words does not close the menu.
It's not perfect, but it's better.

### DIFF
--- a/src/HeuristicCompletion-Model/CoCompletionContext.class.st
+++ b/src/HeuristicCompletion-Model/CoCompletionContext.class.st
@@ -10,6 +10,7 @@ Class {
 		'engine',
 		'source',
 		'position',
+		'completionTokenStart',
 		'completionBuilder',
 		'completionClass',
 		'completion'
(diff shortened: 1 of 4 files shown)


jecisc:
This broke 3 tests:

osx-64 / Tests-osx-64 / testCompletionOpenOnUnderscore – MacOSX64.NECompletion.Tests.CompletionEngineTest7s
osx-64 / Tests-osx-64 / testReplaceTokenWithCaretInTheMiddleOfWordWithFollowingWordsReplacesEntireWord – MacOSX64.NECompletion.Tests.CompletionEngineTest6s
osx-64 / Tests-osx-64 / testReplaceTokenWithCaretInTheMiddleOfWordWithFollowingWordsReplacesEntireWord2 – MacOSX64.NECompletion.Tests.CompletionEngineTest

privat:
tests fixed, unrelated (but windows related) failures

MarcusDenker (approved):

PR #13649, opened 2023-05-04
Start protocol announcements testing + add feature to Announcer

jecisc:
The primary goal of this PR is to create the basic infrastructure to test the announcements related to protocols. 

Indeed, I added tests on protocols to ensure the behavior is correct but none of them checks the announcements. Here are a first bunch of tests about the addition of protocols. 

We have two special cases:
- testRenameProtocolAsAnnounceNewProtocol is skipped because the current behavior is buggy but not so easy to fix currently
- testRenameProtocolAsAnnounceClassReorganizedOnce was failing when I implemented it because the ClassReorganized event was fired two times by the method with the same values. This is a waste and we should fire the event only once.

To fix the double event firing I had to add a new feature on Announcer. 

The feature can be used with the method #prevent:during:. This method takes an announcement class or an annoucement set as parameter and will silence all announcements of this kind during the execution of a block.

Like this I can call methods that announce ClassReorganized without actually announcing it in my specific case. But other events such as ProtocolAdded can still be announced and are working just as expected.

### DIFF
--- a/src/Announcements-Core-Tests/AnnouncerTest.class.st
+++ b/src/Announcements-Core-Tests/AnnouncerTest.class.st
@@ -143,6 +143,19 @@ AnnouncerTest >> testPreparationAnnouncementDeliveryWhenNoSubscriptions [
 ]
 
 { #category : #'tests - subscribing' }
+AnnouncerTest >> testPreventDuring [
+
+	| announcement instance |
+	announcer when: AnnouncementMockA do: [ self fail ] for: self.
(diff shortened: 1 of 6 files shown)


jecisc:
The build time of this PR is really fast compared to other ones I did recently :O 41m 13s
I wonder if we lost all the time because of the double announcement.

MarcusDenker:
sadly the reason is that windows is green, but did not run any tests

jecisc:
That was too good to be true ahah

MarcusDenker (approved):

PR #13650, opened 2023-05-04
Do not classify silently methods

jecisc:
We have two way of classifying methods in the system. One that announce the classification and one that does not.  I found during the past few months places that were using the silent classification when they should have used the normal one, causing lost of informations in Epicea and other tools. 

We currently have only one last place with this silent classification: traits.  I wonder if this is really needed. Here is a test to see if the image builds, if the tests passes and if the bootstrap timae changes with this change.


jecisc:
The image built, the built time was not slower than before and all tests are green. I think is good

MarcusDenker:
A question is if we want to get notifications for the traits methods... to me, everything that is implementation level of traits should not be visible, nobody is classifying a method when a trait installs it's method in the class that uses it.

We already have the problem that "OpalCompiler recompileAll" should be completely silent, yet Traits do raise announcements that they should not. With this, we would get even more of those or not?

jecisc:
I agree with you that we do not want traits to "dirty" the code.

I would still like to remove the silent classification from ClassOrganization tho. The reason is that IMO it is not the right place to do that. If we want a "silent" classification then we have to duplicate the whole API. The current silent classification is not so silent because it currently announce the new protocol creation if it does not exist. What it does not announce is the classification only.

To ensure we have a silent recompilation/less operations that makes tools such as Iceberg dirty when there is no change in code at all is one of the next step in my cleanings. It's the reason I'm starting to add tests on the announcements made. 
To deal with the silent classification in Traits I would like to disable the announcements during the classification instead of duplicating the protocol API in ClassOrganization.

MarcusDenker:
Ok, that makes sense!

MarcusDenker (approved):

PR #13658, opened 2023-05-05
Introduce ClassDescription>>#protocolNames and start using it

jecisc:
This PR is the first step toward #10856.

It does multiple things:
- Push down Behavior>>#protocols to ClassDescription because Behavior is not responsible for protocol management
- It introduces ClassDescription>>#protocolNames as a first step to inline ClassOrganization in it
- It migrates most of the users of #protocols to use #protocolNames. The last user I found is in an external project D=
- Deprecate #protocolNames even if this deprecation will probably not survive until the release, it's to see if I forgot a user somewhere and have logs in the CI. 
- Introduce ClassDescriptionProtocolsTest that will test everything related to protocols in ClassDescription. Most of the code that will end up here will be the code of ClassOrganizationTest, but it allows me to keep track of the migration and be sure everything is well tested

Next steps on this issue:
- Replace the usage of #protocols in Microdown once this PR is integrated
- Move #protocols from Ring to ClassDescription and make it return the real protocols and not the protocol names
- Update all the code doing `x organization protocols` to `x protocols`

### DIFF
--- a/src/Collections-Sequenceable-Tests/OrderedDictionaryTest.class.st
+++ b/src/Collections-Sequenceable-Tests/OrderedDictionaryTest.class.st
@@ -735,19 +735,15 @@ OrderedDictionaryTest >> testDictionary [
 
 { #category : #tests }
 OrderedDictionaryTest >> testDictionaryPublicProtocolCompatibility [
-	| dictionary |
 
+	| dictionary |
 	dictionary := self emptyDictionary.
-	{dictionary -> Dictionary.
-	dictionary class -> Dictionary class}
-		do: [:assoc |
-			assoc value protocols
-				reject: [:protocol |
-					#('private' 'print' 'undeclared' 'copy' 'compar' '*')
-						anySatisfy: [:each | protocol asString beginsWith: each]]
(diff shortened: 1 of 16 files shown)


Ducasse:
just a little question. 
How protocolnames deals with the protocols of method extensions?

S

jecisc:
#protocolNames return all the protocols currently present in the class, including the extension protocols. If you have extension methods, it will return things such as `*Kernel`. 

One of my short term goal is to have #protocols return real protocols and not names so that we can easily do things such as `aClass protocols reject: [ :protocol | protocol isExtensionProtocol ]`. 

Long terme goal is to not manage extensions via protocols so that we do not pollute them but that's not for the next few weeks

jecisc:
The CI is not happy. I'll need to check what is the problem because it is pretty obscure.

jecisc:
So, I think my problem originates from the fact that #protocols is called by the code of the automatic rewrite and we end up in an infinit loop when it tries to rewrite.

MarcusDenker (approved):

PR #13661, opened 2023-05-05
Added `thisProcess` pseudovariable using existing bytecode.

Mathilde411:
Added the `thisProcess` pseudovariable to access the current process. Added code coloration along with it.

### DIFF
--- a/src/Kernel/ThisProcessVariable.class.st
+++ b/src/Kernel/ThisProcessVariable.class.st
@@ -12,6 +12,12 @@ ThisProcessVariable class >> variableName [
 	^#thisProcess
 ]
 
+{ #category : #'code generation' }
+ThisProcessVariable >> emitValue: methodBuilder [
+
+	methodBuilder pushThisProcess 
+]
+
 { #category : #testing }
 ThisProcessVariable >> isThisProcessVariable [
 	^true
--- a/src/Kernel/ThisProcessVariable.class.st
+++ b/src/Kernel/ThisProcessVariable.class.st
@@ -1,0 +1,24 @@
+"
+I model thisProcess keyword
+"
+Class {
+	#name : #ThisProcessVariable,
+	#superclass : #ReservedVariable,
+	#category : #'Kernel-Variables'
+}
+
+{ #category : #accessing }
+ThisProcessVariable class >> variableName [
+	^#thisProcess
+]
+
+{ #category : #testing }
+ThisProcessVariable >> isThisProcessVariable [
+	^true
+]
+
+{ #category : #debugging }
+ThisProcessVariable >> readInContext: aContext [
+	"We expect this method to be called either from the same process or by debugger process that will read the correct process. Otherwise we cannot guarantee to read the correct process"
+	^ Processor activeProcess
+]
--- a/src/AST-Core-Tests/RBParserTest.class.st
+++ b/src/AST-Core-Tests/RBParserTest.class.st
@@ -468,6 +468,12 @@ RBParserTest >> testCreationThisContexNodeWithExpression [
 ]
 
 { #category : #'tests - CreationProtocol' }
+RBParserTest >> testCreationThisProcessNodeWithExpression [
+	self compare: (self parserClass parseExpression: 'thisProcess')
+		  to: RBVariableNode thisProcessNode 
+]
+
+{ #category : #'tests - CreationProtocol' }
 RBParserTest >> testCreationVariableNodeWithExpression [
 	self compare: (self parserClass parseExpression: 'variable')
 		  to: (RBVariableNode named: #variable start: 1)
--- a/src/AST-Core/RBVariableNode.class.st
+++ b/src/AST-Core/RBVariableNode.class.st
@@ -28,6 +28,8 @@ RBVariableNode class >> identifierNamed: anIdentifierName at: aPosition [
 		ifTrue: [ ^ self selfNode start: aPosition ].
 	anIdentifierName = #thisContext
 		ifTrue: [ ^ self thisContextNode start: aPosition ].
+	anIdentifierName = #thisProcess
+		ifTrue: [ ^ self thisProcessNode start: aPosition ].
 	anIdentifierName = #super
 		ifTrue: [ ^ self superNode start: aPosition ].
 	^ self named: anIdentifierName start: aPosition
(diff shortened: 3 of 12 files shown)


src/Kernel/ThisProcessVariable.class.st
```diff
@@ -0,0 +1,24 @@
...
+
+{ #category : #testing }
+ThisProcessVariable >> isThisProcessVariable [
+	^true
+]
+
+{ #category : #debugging }
+ThisProcessVariable >> readInContext: aContext [
+	"We expect this method to be called either from the same process or by debugger process that will read the correct process. Otherwise we cannot guarantee to read the correct process"
+	^ Processor activeProcess
```
privat, line 29: This one is problematic. I do not like the idea to silently return possible a bad value. I fear that this will bite people and cause inconsistencies.

I prefer to have an active runtime error saying that `thisProcess` in unreliable, or a passive error (return nil) instead of having a bad value.

I'm not sure how exactly contexts and processes are associated. If a context can be dead and associated to no context, returning nil might make sense.
If a context can be associated to a single process at most, searching the correct process can be a solution.
I prefer a slow correct answer to a fast wrong one.

privat (commented):
From a language point of view, introducing new global identifiers might cause breakages.
What about possible shadowing (there is none in the main pharo repository, but it might be not that representative)?
Maybe explicitly allowing shadowing for this pseudo-variable could be a backward compatibility gesture.

privat:
Is introducing new pseudo-variables the right approach? Pseudo-variables are hack and do not behave like variables (see `ReservedVariable>>#usingMethods`, or `readInContext:` for instance). Were alternative approaches considered?

src/Kernel/ThisProcessVariable.class.st
Mathilde411: When exploring the senders, we saw that it is not used outside of the debugger, so we don't expect this to be a problem. But returning nil could also be a solution to be sure we have a consistent output.

Mathilde411:
> Is introducing new pseudo-variables the right approach? Pseudo-variables are hack and do not behave like variables (see `ReservedVariable>>#usingMethods`, or `readInContext:` for instance). Were alternative approaches considered?

This specific implementation was chosen as it uses an extension of the same bytecode that's being used by `thisContext`. That's why we chose to use a pseudovariable. We didn't really consider alternative ways of implementation.

guillep:
In fact, the addition of `thisProcess` has been in "doing" state for long(years). All the compiler had support for this except for the frontend. This PR just adds the frontend.

The benefit of this is the ability to access the current process without sending messages and thus is very handy in particular for instrumentation. Right now, accessing the current process through `Processor activeProcess` is very complex and can complexify instrumentation (cf to its implementation) because it has a lot of code to support debugging.

Regarding the reified variables, we just followed the existing pattern. If you ask me, I am not entirely convinced about this `ThisProcessVariable` singleton, but that is the way right now to plug into the system as far as we understood.

MarcusDenker:
- Yes, the Singleton is not good, I wanted to instead have one instance per "Smalltalk globals".
- I vote for adding thisProcess even though it adds some backward compatibility challenge
- But I want to explore some idea that I had, but as such, this is research at first: the whole idea of "reifications" like thisContext should I think be generalized. I will write some notes about the idea that I had there.

MarcusDenker:
- Yes, the Singleton is not good, I wanted to instead have one instance per "Smalltalk globals".
- I vote for adding thisProcess even though it adds some backward compatibility challenge
- But I want to explore some idea that I had, but as such, this is research at first: the whole idea of "reifications" like thisContext should I think be generalized. I will write some notes about the idea that I had there.

privat:
I think I understand the need, you want some way to access a low-level VM information. Because it's low level, the classic approach of primitive methods is not suitable, as you do not want to depend on message sending.

So you have some fancy byte code that push some cool value on the stack, but no way to express this fancy byte code in the Pharo language.
The core issue is not new, and I kind of understand that `thisContext` was introduced as a make-believe variable to solve the same kind of problem eons ago.
But having make-believe variables are very hackish and forces a lot of MOP and tooling code to go along with the lie.

So what are the choices
* keep a pseudo-variable. But maybe improve them, so they fit better in the metamodel?
* a new syntactic construct? Does the need worth the hassle?
* a make-believe message send? Like some kind of [intrinsic function](https://en.wikipedia.org/wiki/Intrinsic_function) ? I kind of like this one, and it is generalizable to other use cases. This need more thought, but that's something that worth investigating.
* some embedded free-style ir or bytecode (à la C/C++ asm declaration). While awful for almost all software engineering quality metrics, they are a nice way to play with low-level stuff and experiment.
* something else?

MarcusDenker (approved):

guillep:
> * keep a pseudo-variable. But maybe improve them, so they fit better in the metamodel?

I think this is the simplest solution, implementation-wise, and in terms of friction. Variables are traditionally bound at compile time, so they do not have late-bound semantics that could be overridden.
 
> * a new syntactic construct? Does the need worth the hassle?

I don't think it's worth it. I do not expect people to write code with `thisProcess`, so there will be very few cases. But core parts of the system require an easy way to access the current process.

> * a make-believe message send? Like some kind of [intrinsic function](https://en.wikipedia.org/wiki/Intrinsic_function) ? I kind of like this one, and it is generalizable to other use cases. This need more thought, but that's something that worth investigating.

I don't like this one. Message sends have late-bound semantics. I don't like to add new "special cases" like #to:do that cannot be overridden.

I like the idea of an intrinsic, but not without a way to statically bound a send. I would like to explore the idea of using the annotations to do statically-bound sends. Then those statically-bound sends could be defined as intrinsecs, macros, and other special support like this one. However, I don't see how annotations could annotate a send (I want to annotate A send, not an expression tree).

I was thinking at something like the following, where `message:` is statically bound, but `bar` is late bound.

```
foo  @Array>>message: param bar
```

> 
> * some embedded free-style ir or bytecode (à la C/C++ asm declaration). While awful for almost all software engineering quality metrics, they are a nice way to play with low-level stuff and experiment.

Well, the whole point of this is to be able to access it easily. Maybe it would be nice to have a macro system built statically bound messages where we receive the IR builder as argument?

> 
> * something else?

guillep:
Thanks @BastouP411 !

PR #13664, opened 2023-05-05
Fix bug (issue #13663): enabling the setting ‘Show lines numbers for all browsers’ only showed line numbers for the first method selected in a class

### LINKED ISSUE #13617: Calypso: Saving a method removes the line numbers
In Calypso there is a setting to always show line numbers and I enabled it. 

The problem is that now my browsers keep switching between showing and not showing line numbers. When I save a method, the line numbers are vanishing and then at some point they pop again. 

This perturb me a lot.

### LINKED ISSUE #13663: Enabling the setting ‘Show lines numbers for all browsers’ only shows line numbers for the first method selected in a class
**Bug description**
Enabling the setting ‘Show lines numbers for all browsers’ only shows line numbers for the first method selected in a class

**To Reproduce**
* In the Settings, enable ‘Show lines numbers for all browsers’
* In a System Browser, select a class and select a method (note that line numbers are shown)
* Select another method (in the same class)

Line numbers are unexpectedly not shown. When selecting another class and method, the line numbers reappear, then disappear again when another method is selected in the same class.

**Version information:**
Pharo-12.0.0+build.293.sha.f544baec532b913167d1ab83a345514efe6004a2

Rinzwind:
Scenario from the issue (#13663):

> * In the Settings, enable ‘Show lines numbers for all browsers’
> * In a System Browser, select a class and select a method (note that line numbers are shown)
> * Select another method (in the same class)
>
> Line numbers are unexpectedly not shown. When selecting another class and method, the line numbers reappear, then disappear again when another method is selected in the same class.

### DIFF
--- a/src/Calypso-Browser/ClyTextLineNumbersSwitchMorph.class.st
+++ b/src/Calypso-Browser/ClyTextLineNumbersSwitchMorph.class.st
@@ -58,7 +58,7 @@ ClyTextLineNumbersSwitchMorph >> attachToTextMorph [
 	self updateLabel.
 	self addMorph: label.
 	self class showLineNumbers
-		ifTrue: [ self toggle ]
+		ifTrue: [ textMorph withLineNumbers ]
 ]
 
 { #category : #operations }


Rinzwind:
@request-info: body has been updated.

jecisc:
I wonder if this is related to: https://github.com/pharo-project/pharo/issues/13617

jecisc (approved):
I tested it and it works well! It also fixes https://github.com/pharo-project/pharo/issues/13617 !

If other agrees I'm in favor of backporting this in P11 because this bug is really annoying if we enable the setting

PR #13668, opened 2023-05-06
Clean some code using better condition branches

jecisc:
Just trying to make some code more readable.


privat (approved):
Can the simplified patterns be a rule with an autofix?

jecisc:
> Can the simplified patterns be a rule with an autofix?

It is not *so* simple since someone could implement #compliedMehod:ifAbsent: on its object without implementing #compiledMethod:ifPresent:
It's the same problem with #ifEmpty:/#ifNil:...

PR #13673, opened 2023-05-09
Pharo 11 sync

jecisc:
I hope this will cause no problem because there were multiple conflicts


src/BaselineOfPharo/BaselineOfPharo.class.st
```diff
@@ -7,24 +7,43 @@ Class {
...
 
+{ #category : #'external projects' }
+BaselineOfPharo class >> beautifulComments [
+	<externalProject>
+
+	^ PharoExternalProject 
+		newName: 'BeautifulComments' 
+		owner: 'pillar-markup' 
+		project: 'BeautifulComments' 
+		version: 'Pharo12'
```
demarey, line 18: should be 'integration'

src/BaselineOfPharo/BaselineOfPharo.class.st
```diff
@@ -47,36 +66,109 @@ BaselineOfPharo class >> externalProjects [
...
 
+{ #category : #'external projects' }
+BaselineOfPharo class >> microdown [
+	<externalProject>
+	
+	^ PharoExternalProject 
+		newName: 'Microdown' 
+		owner: 'pillar-markup' 
+		project: 'Microdown' 
+		version: 'Pharo12'
```
demarey, line 95: should be 'integration'

src/BaselineOfPharo/BaselineOfPharo.class.st
```diff
@@ -7,24 +7,43 @@ Class {
...
+
+{ #category : #'external projects' }
+BaselineOfPharo class >> documentBrowser [
+	<externalProject>
+	
+	^ PharoExternalProject 
+		newName: 'DocumentBrowser' 
+		owner: 'pharo-spec' 
+		project: 'NewTools-DocumentBrowser' 
+		version: 'Pharo12'
```
demarey, line 46: should be 'integration'

jecisc:
Thank you for the review @demarey but this PR is now useless because another sync was integrated. 

But the versions used are not integration in the PR that was integrated so we need to update it. 

Currently it's hard to update Pharo 12 because the CI has some troubles and we don't know if it comes from a recent change or from the CI in itself (because the build time is much slower than before and some files cannot be written but we cannot find a change that could cause that)

PR #13678, opened 2023-05-09
Remove some Monticello dependencies from RPackage

jecisc:
RPackage should not depend on Monticello. Instead Monticello should listen to the system announcements and do what it needs to do.  It is really hard to understand how all those announcements are managed. I found some cases where Montciello already listen to some announcements to do what it needs to do so let's see if we can just remove some of the couplings and check if we have failing tests.


Ducasse:
thanks I hope that one day we will be able to remove MC :)

jecisc:
Done in another PR.

PR #13679, opened 2023-05-09
Completion: improve completion on super

privat:
When completing on super, the current message is proposed only if it is a redefinition.

This PR also add tests (in fact, most person-hours was spent on the test part)


jecisc (approved):
Nice!

MarcusDenker (approved):

PR #13682, opened 2023-05-09
Cleanup: Metacello-Platform 

MarcusDenker:
- remove empty and unused MetacelloPharo30Platform
- move some extension methods to the classes for classes in MetaCello and Monticello


jecisc (changes requested):
There is still a reference in bootstrap/scripts/03-metacello-bootstrap/01-loadMetacello.st

PR #13683, opened 2023-05-09
Completion: replacement put caret at the right place

privat:
When replacing super send including arguments, the caret position was wrong.

This PR fix, rationalize, document and test the caret placement after a completion

### DIFF
--- a/src/HeuristicCompletion-Tests/CoCompletionEngineTest.class.st
+++ b/src/HeuristicCompletion-Tests/CoCompletionEngineTest.class.st
@@ -154,11 +154,14 @@ CoCompletionEngineTest >> testExitingWordClosesCompletionContext [
 	controller openMenu.
 
 	self assert: controller hasCompletionContext.
+	self assert: self editorTextWithCaret equals: 'self mEthOdThatDoesNotExis|t toto'.
 	self assert: controller completionToken equals: 'mEthOdThatDoesNotExis'.
 	editor keyDown: (self keyboardEventFor: Character arrowRight).
+	self assert: self editorTextWithCaret equals: 'self mEthOdThatDoesNotExist| toto'.
 	self assert: controller completionToken equals: 'mEthOdThatDoesNotExist'.
 	self assert: controller hasCompletionContext.
 	editor keyDown: (self keyboardEventFor: Character arrowRight).
+	self assert: self editorTextWithCaret equals: 'self mEthOdThatDoesNotExist |toto'.
 	self assert: controller completionToken equals: ''.
 	self deny: controller hasCompletionContext.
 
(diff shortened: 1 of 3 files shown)


jecisc:
I have one problem with the current completion but I did not check if/how other IDE were managing that. 

Imagine I have:
`self meth| toto`

and the end result I want is 

`self method otherMethod toto`

here is what I would do:

1) `self meth| toto`
2) Use auto complete and end up with: `self method |toto`
3) type "otherMe" and end up with `self method otherMe|toto`
4) accept the change
5) end up with `self method otherMethod |` and the toto vanished. 

So either I have to type again toto or do a space and left arrow before typing the second method. 

I wonder if there could be a way to prevent that. I guess I should check other IDE to see their behaviors.

privat:
I have the same issue, but I don't know what I really want, so I did not change that. So this PR just make the current behavior consistent and tested.

A quick review of other IDE (and shell completion) behavior could help a lot. I don't know what they do, but I feel that I complain a lot less about their UI choices (I mainly use completion in vim and bash, so I'm not that representative anyway)

privat:
bash with tab:
bash complete up to the shortest common prefix by default, the completion list is shown on a second `tab`

* `foo aFi|` → `foo aFile |` add a space and put the caret after
* `foo aDirec|` → `foo aDirectory/|` no space (it's contextual!)
* `foo aFi|aFile` →`foo aFile|aFile` do not overwrite and do not insert a space
* `foo aFi| aFile` →`foo aFile| aFile` do not insert a second space and put the caret before the existing one

vim with plain text-based ^n/^p completion with default config:
active completion (^n/^p) complete automatically the first choice.
if multiple choice exists the completion menu is opened and can be used to cycle around them (^n/^p or down/up) and enter confirm (undoing the free initial completion).

* `Ph|` → `Pharo|` no space
* `Ph|Pharo` → `Pharo|Pharo` no overwrite, no space
* `Ph| Pharo` → `Pharo| Pharo` no overwrite, no new space

privat:
X- Other (current) alternative with space then left arrow

* `self meth| toto`
* tab → `self method |toto`
* otherMe → `self method otherMe|toto`
* space left → `self method otherMe| toto`
* tab → `self method otherMethod| toto`

Y- you can also disable the space addition in the setting (*Put a space after completion*).

* `self meth| toto`
* tab → `self method| toto`
* space otherMe → `self method otherMe| toto`
* tab → `self method otherMethod| toto`

X is usually too late, and I also just rewrite the overwrote message.
Y is ok except that you just keep adding space (not a big overhead but might be less annoying)

Other possible new behavior could be:

A- put the caret before the space if there is already one, but after the space if there is none.
This is a smart variation of Y.

* `self meth| toto`
* tab → `self method| toto`
* space otherMe → `self method otherMe| toto`
* tab → `self method otherMethod| toto`

B- or disable overwriting

* `self meth| toto`
* tab → `self method |toto`
* otherMe → `self method otherMe|toto`
* tab → `self method otherMethod| toto`

A- seems nice. I'm afraid that B is too drastic and that overwriting has its uses, especially because going left and right on the completion menu does not close it.

privat:
I implemented the A policy

guillep:
I think that in other IDEs it’s less of a problem because other languages chain messages and operations with other than spaces.

aFoo.doThis().what.doThat()

while that is for us:

aFoo doThis what doThat

=> but the conflict happens because pressing “space” triggers completion.
Right?

Ducasse:
I like that we can finally get a nice solution because it is a huge productivity improvement. 
Thank you all for this kind of love. Because such "details" are important :)

PR #13687, opened 2023-05-10
Basic type analysis for better completion

### LINKED ISSUE #12976: Autocompletion should prioritise the methods implemented on the current receiver
When we send #new to a class, we could prioritize the methods of the class in the code suggestion. See example in the next message

### LINKED ISSUE #13681: [Feature Request] Code completion: react on `self class`
When we are doing `self class`, the receiver will be the class side of the object we are currently into. Maybe we could use that in order to refine the completion?

I had a case just today where I typed `self class |` and the propositions were really off what I wanted while the list could have been better by proposing in priority the methods of the class side. 

I don't know how the code completion works so I have no idea of the development cost.

privat:
This PR propose a simple general type visitor.

Because it is simple (a single class) and agnostic enough of specific usages, I put it in the `AST` package.
This is a work in progress, but I would like to PR the first steps. The Visitor is currently fully usable, provide tests, good results for simple cases and a simple API (that may evolve).

As A POC, I added a new heuristic for the completion of message send.
Capture d’écran du 2023-05-09 20-49-42

The feature should be quite powerful and could replace some existing ad hoc heuristics.

fix #13681
fix #12976

### DIFF
--- a/src/AST-Core-Tests/ASTTypingVisitorTest.class.st
+++ b/src/AST-Core-Tests/ASTTypingVisitorTest.class.st
@@ -1,0 +1,113 @@
+"
+An ASTTypingVisitorTest is a test class for testing the behavior of ASTTypingVisitor
+"
+Class {
+	#name : #ASTTypingVisitorTest,
+	#superclass : #TestCase,
+	#category : #'AST-Core-Tests-Type'
+}
+
+{ #category : #asserting }
+ASTTypingVisitorTest >> assertTypedTree: aNode [
+
+	aNode nodesDo: [ :node |
+		(self expectedNodeType: node) ifNotNil: [ :type |
+			self
+				assert: (node propertyAt: #type ifAbsent: [ nil ])
+				equals: type ] ]
+]
+
+{ #category : #examples }
+ASTTypingVisitorTest >> exampleLoop [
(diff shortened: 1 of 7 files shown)


jecisc (approved):
Nice!

If you dealt with #class there is an easy one that is close: https://github.com/pharo-project/pharo/issues/12976

privat:
It's already managed by the PR. It's the whole point of a global semantic approach, it avoids using a bunch of ad hoc heuristics.
Capture d’écran du 2023-05-10 07-28-48

privat:
Some numbers, if you care.
All methods of AST-Core (`'AST-Core' asPackage methods collect: #parseTree`)

```
Number of AST: 1570
Number of value nodes: 20883
Individual typing (one new visitor per method)
  duration:0:00:00:00.025
  typed nodes: 7196 (34%)
Agretated typing (a single visitor reused. Allow better knowledge of ivar and simple inter-procedural analysis)
  duration:0:00:00:00.026
  typed nodes: 7777 (37%)
+fixed point typing (reiterate the analysis until a fix point is reached)
  duration:0:00:00:00.177
  steps:6
  typed nodes: 9688 (46%)
```

I tried to look at the untyped nodes. Most are argument variable nodes as they are never typed (yet).

MarcusDenker (approved):

Issue #13675, opened 2023-05-09
Compiler: improve ReservedVariables


MarcusDenker:
**Describe the problem**
The ReservedVariables (ala PseudoVariables) are singletons and are created via methods likes #thisContextNode.

Why is this a problem? For one, they are too special, all other variables are just looked up in the scope by name, which is perfectly possible for ReservedVariables, too, as SystemDictionary knows them. 

The idea of setting the variable early and the use of a Singleton is from before they where looked up via the SystemDictionary.
Now we can simplify:

- SystemDictionary should instantiate and not use singleton (we should have one instance of "self" per environment)
- check all uses  #thisContextNode, #selfNode, #superNode. Just create an AST node with the name and let the name analysis later do the lookup and put the instance it finds via SystemDictionary
- Think about Terminology. Traditionally they are called "PseudoVariables", maybe we should change the terminology everywhere

privat:
What it `self` or `thisContext` (or `super`, but this one is weirder)?
As source code elements or AST nodes, the answer is easy, they are occurrences of words in some method definitions.
But as variable, with a scope, the answer is less obvious.

I was surprised when I discovered that reserved variable are singletons, but it kind of make sense as, for instance, the name `self` is visible everywhere and correspond everywhere to the current receiver.

What will be the benefits of having a `self` per environment? How can two `self` differs?

MarcusDenker:
They used to be hard coded in #lookupVar:, thisContext in the methodScope and self/super in an "instance scope" that was used to lookup ivars/slots instead of  #lookupVar: in Behavior and using that as a Scope. 

The lookup we have now feels much better (to me). At first I was not 100% sure about them being part of the global scope, I guess the alternative would be the method scope, maybe because this would allow us to not have singletons.

I think the reason for not having singletons would be that one then could easier have environments that use a re-defined SelfVariable class. Conceptionally it feels odd that a SystemDictionary would be forced to use one shared instance. e.g. if the environment defines SelfVariable, it should have an instance of *that* class as the SelfVariable, not forced one of another SystemDictionary instance.

privat:
From a pure low-level OO design, what are the benefits of having multiple `self`. How the state or the behavior of one instance will be different of the state or behavior of another instance? What will be the implementation of `=`?

`self` can be seen as a pseudo 0th parameter of a method definition, so a method scope might have sense. (so one `self` per method?)
[...]

PR #13688, opened 2023-05-10
ReservedVariable: No need for Singleton

MarcusDenker:
With pseudoVariables being looked up just like any other variable during semantic analysis, we can make them more normal and remove the Singleton pattern. It was needed back when they where defined in the method scope: There was no way to get the shared instance. Now we can just use #lookupVar: on any scope.

Some tests use non-analysed AST modes, we can just create instances using #new for those.

one step of  #13675

### DIFF
--- a/src/AST-Core/RBVariableNode.class.st
+++ b/src/AST-Core/RBVariableNode.class.st
@@ -49,22 +49,22 @@ RBVariableNode class >> named: aName start: aPosition [
 
 { #category : #'instance creation' }
 RBVariableNode class >> selfNode [
-	^ (self named: #self) variable: SelfVariable instance
+	^ self named: #self
 ]
 
 { #category : #'instance creation' }
 RBVariableNode class >> superNode [
-	^ (self named: #super) variable: SuperVariable instance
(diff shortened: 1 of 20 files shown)


MarcusDenker:
Sadly the refactoring engine works on the non-name analyzed AST, yet expects to destinguish self and super from normal variables. 

I re-parses *a lot*, and some of the trees then are transfomed without making a copy (as it assumes they are the result of a re-parse). This means we can not just use #ast  (and in addition to name analysis, benefit from the cache).

The refactoring engine looks like huge mess... it should never re-parse but conceptionally work on ASTs that are already there.

As a workaround, we should change all the (far to many) #parseTree methods to do name analysis, this is what #parseTree now does in CompiledMethod, too

MarcusDenker:
This worked! The change done is minimal: just add #doSemanticAnalysis everywhere where RB creates a parseTree.
Many many more improvements should be done (e.g. the model of methods of RB should model the AST an not always reparse just to get all self sends...) , but that is for another time.

privat (approved):

PR #13689, opened 2023-05-10
Improve some error logging

jecisc:
This updates two things:
- Subscript out of bounds prints the signaler in the error message when the signaler is known
- #basicAt: has better logging when we are out of bounds

This will help debug error in #13456 and multiple other PRs


svenvc:
LGTM

MarcusDenker (approved):

jecisc:
The failure seems unrelated and the build is not done because of windows

PR #13690, opened 2023-05-10
Remove SystemOrganizer

jecisc:
This commit moves all the logic of the SystemOrganizer into RPackageOrganizer and removes the class.

The logic of "categories" is still here and needs to be removed in the future, but that will be in the future steps.


jecisc:
Test failure is unrelated (random failure on halt test)

MarcusDenker (approved):

PR #13692, opened 2023-05-10
Improve removal of protocols and add tests

jecisc:
This make ClassDescription>>removeProtocol: work with a real instance of Protocol and add multiple tests on protocol removal.

This will also help to make CompiledMethod>>protocol return a real protocol because I found some code doing `class removeProtocol: method protocol`.


MarcusDenker:
Failing test:

EpCodeChangeIntegrationTest>>testProtocolRemoval

jecisc:
The test is passing in my local image.. :/

jecisc:
I found the problem that is not really related to this PR so I will do another PR with a test and a fix

MarcusDenker (approved):

PR #13693, opened 2023-05-10
Add highlight text color to theme for debugger stack coloring

jecisc:
I plan to add a feature to the debugger to highlight stack elements that are in the same class or package than the selected context in the debugger. In order to do that I need a highlight text color so I propose to add one in the theme so that I can use it in NewTools

I also improved some protocols (we can still do way better)

image

image


jecisc:
I opened the NewTools pendant of this PR https://github.com/pharo-spec/NewTools/pull/521

It can be checked once this PR is reviewed and merged

privat (approved):

Issue #13686, opened 2023-05-09
Another improvement for completion


Ducasse:
imagine I have 

`1 between: 12| and: 14`

when I complete would like to get `1 between: 12 and: 14 |` and not 

`1 between: 12 and: | and: 14`

Ducasse:
I played with Pharo 12 completion and it is neat. 

May be it was already addressed but I got `1 between:  |and:` and it would be better to have `1 between: | and:`

privat:
About `1 between: 12| and: 14` this is interesting. I'm trying to guess when such a situation occurs.
Some cleaning and refactoring might be needed before, as nothing in the engines take care of what follows the cursor.

About the second on the caret position, I thought I have fixed it (and even added tests for it). But I can reproduce. I will look at it.

Ducasse:
Thanks Jean I thought also that you fixed it.

PR #13696, opened 2023-05-10
Completion mini fixes

privat:
Two small fixes:

* #13687 added #type property on some AST node. NECCompletion knows to process them, but was not robust enough. Fix that.
* Fix the bad caret position when completing at the end of the text. Fix one half of #13686. And add more tests


privat:
unrelated CI issue (a morph SubscriptOutOfBounds)

PR #13699, opened 2023-05-10
Completion browse with Ctrl-B

### LINKED ISSUE #13584: meta+left shortcut used for 2 things, breaking CompletionEngine's "close details"
**Bug description**
`Meta key + left arrow` shortcut (`cmd + <-` or `ctrl + <-`) is used for two actions that could be activated under the same condition, leading to only one actually working.

1) Move cursor next to previous word
2) Hide details of `CompletionEngine`/`NECMenuMorph`

As a result, the cursor move happens, but there is no way to just hide details without moving the cursor.

Also, the "cmd" words in these completion shortuct labels are hardcoded instead of using "ctrl" on non-mac systems: https://github.com/pharo-project/pharo/issues/6169

**To Reproduce**
1. Open Playground
2. Type `Tr` and wait for completion popup to open
3. Notice the "cmd<- close detail" gray text at the bottom of the popup
4. Press `cmd + <-` or `ctrl + <-`
5. Cursor moved, instead of just closing the details

**Expected behavior**
There should be separate keyboard shorcut for opening and closing the detail and for moving cursor.

**Version information:**
 - Probably all OS, tested on Windows, not sure about mac
 - Probably all Pharo versions, tested on 7-12

privat:
fix #13584 

Completion menu use Ctrl-B to browse instead of hijacking Ctrl-right (in fact unlike documentation says, Ctrl-left was not hijacked).

I needed to hack on browseIt actions (and related) on rubric/calypso, so I also did a small cleanup, by merging some methods and removing unused ones.

Help test still shows a hard-coded "Cmd-B" text info. I did not find a way to make the text platform independent reliably (and I don't even have mac)

### DIFF
--- a/src/Calypso-Browser/ClyTextEditor.class.st
+++ b/src/Calypso-Browser/ClyTextEditor.class.st
@@ -57,21 +57,6 @@ ClyTextEditor >> notify: aString at: anInteger in: aStream [
 ]
 
 { #category : #operations }
-ClyTextEditor >> printIt [
-	"Treat the current text selection as an expression; evaluate it. Insert the
-	description of the result of evaluation after the selection and then make
-	this description the new text selection."
-
-	| printString |
-	self
-		evaluateSelectionAndDo: [ :result |
-			printString := [ result asScriptResultStringInCalypso ]
-				on: Error
-				do: [ '<error in printString: try ''Inspect it'' to debug>' ].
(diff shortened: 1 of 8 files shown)


privat:
tests ok, unrelated SubscriptOutOfBounds

MarcusDenker (approved):

PR #13700, opened 2023-05-11
Completion: better update

privat:
The completion menu was updated on every key down (even unrelated, like a lone control key down) but the menuMorph was not refreshed (because refreshing cause the selection to be reset).
This caused graphical glitches when the menu content did not fit the initial completion menu area.

This PR store the current completion token into the completion context, so it becomes easy to see if the completion need an update and the menu need a refresh.

This should remove the last GUI issue with the completion menu (no issue number but I wanted to fix it anyway)
Feel free to open new ones :)

### DIFF
--- a/src/HeuristicCompletion-Model/CoCompletionContext.class.st
+++ b/src/HeuristicCompletion-Model/CoCompletionContext.class.st
@@ -25,7 +25,6 @@ CoCompletionContext class >> engine: aCompletionEngine class: aClass source: aSt
 	^ self new
 		completionClass: aClass;
 		engine: aCompletionEngine;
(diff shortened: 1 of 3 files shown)


jecisc:
Nice! I got this glitch today but I didn’t know how to reproduce it.

MarcusDenker:
tests failing in CoASTResultSetBuilderTest

privat:
yeah. HeuristicCompletion is complex (and fragile), and the tests set up and use objects a little differently :(
What is strange is that windows did not fail the same tests.

jecisc:
Windows is broken but tells us tests are fine.. (but they are not run)

image

MarcusDenker (approved):

PR #13702, opened 2023-05-11
Removal of nonexisting protocol should not announce anything

jecisc:
This fixes a bug where sometime we were announcing the removal of a protocol that did not even exist in the class. I added some tests around that.

### DIFF
--- a/src/Kernel-Tests/ProtocolAnnouncementsTest.class.st
+++ b/src/Kernel-Tests/ProtocolAnnouncementsTest.class.st
@@ -51,6 +51,30 @@ ProtocolAnnouncementsTest >> testClassifyUnderAnnounceNewProtocol [
 ]
 
 { #category : #tests }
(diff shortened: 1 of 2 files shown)


privat (approved):

privat:
Most pharo removal methods signal an error when you remove an absent item (so you have to use `ifAbsent:`).
Maybe it could be consistent to do the same with protocols?

jecisc:
I agree. It's something I would like to have also but there are so many things to clean that I postponed this.
My priorities on protocols currently were:
- Deal with real protocols instead of strings
- Tests and clean announcements
- Remove the "category"/"heading"/'tag" denomination
- Inline ClassOrganization

I was keeping the error management for later if I still had energy on this subject ahah
I already changed the "silence" to "error" for #protocolNamed: but I should also do that in other places indeed

Issue #13136, opened 2023-03-24
ReClassVariablesShadowingRule does not update after changing class defintion


jecisc:
If we add a class variable with the same name as a global/class then we cannot access to this class in the code since the variable is "shadowed". Maybe we should have the same warning as in the shadow instance variables declaration?

privat:
Interesting. What should also be done, at least, is a dedicated critique that can be displayed under the text area in class edition, and during attempt to commit.

jecisc:
I agree! 

To me shadowing of globals should work as closely as possible to shadowing of instance variables ideally. (I like consistency :) )

MarcusDenker:
There is already ReClassVariablesShadowingRule, the name is a bit wrong, it warns in the class defintion pane if you add a shadowing ivar or class variable

privat:
The rule exists, but it does not appear in the "critiques" box in calypso, not highlight variable in the class definition code (fluid or not).
![Capture d’écran du 2023-03-28 19-54-02](https://user-images.githubusercontent.com/135828/228392130-90195aa5-f287-4842-bdad-cd89053a66a9.png)

jordanmontt:
We were checking with @ValentinBourcier this issue. We found out that class `SycRenameArgOrTempCommand` is the one that calls this refactor. After, one of the superclass `CmdCommand`has a method `confirmRefactoringInContext: by:` calls the refactoring preview presenter (see image). And then, that presenter in the method `accept` applies the refactor. So, this is very hard coded.

We think that the best is to move this logic to the driver @balsa-sarenac 
The refactors should be able to call the driver with the option "no confirmation" and the driver is the one that should be responsible of applying the refactor with or without presenter. We also think that the logic of applying the refactors should be moved from `SycRenameArgOrTempCommand>>accept` to the driver because now the presenter it also has business logic.

<img width="498" alt="Capture d’écran 2023-03-31 à 11 38 21" src="https://user-images.githubusercontent.com/33934979/229084799-79b84a0e-1181-4054-954a-45a19e634542.png">

PR #13704, opened 2023-05-11
Allow-Slots-Shadow-PseudoVariables 

MarcusDenker:
This is maybe a bit controversial:

PseudoVariables are just Variables, there is nothing that should, on the level of the class builder, raise a hard error if you want to shadow "self" with a slot named "self".

The tools will warn you that the variable is shadowing an existing variable of that name already, we could add a special rule (or improve the description) for this case in addition.

A side effect is that this means that thisProcess as a instance variable would now still be possible and thus introducing thisProcess is less of a backward compatibility problem.

### DIFF
--- a/src/Kernel/Slot.class.st
+++ b/src/Kernel/Slot.class.st
@@ -43,17 +43,11 @@ Slot class >> checkValidName: aSymbol [
 	aSymbol startsWithDigit ifTrue: [
 		^ InvalidSlotName signalFor: aSymbol ].
 	
-	(self isPseudovariableName: aSymbol) ifTrue: [
+	(#(#true #false #nil) includes: aSymbol) ifTrue: [
 		^ InvalidSlotName signalFor: aSymbol ].
-
 
 	(aSymbol allSatisfy: [ :aCharacter | aCharacter isAlphaNumeric or: [ aCharacter = $_ ] ]) ifFalse: [
 		^ InvalidSlotName signalFor: aSymbol ]
-]
-
-{ #category : #testing }
-Slot class >> isPseudovariableName: aSymbol [
-	^ #('self' 'true' 'false' 'nil' 'thisContext' 'thisProcess' 'super') includes: aSymbol
 ]
 
 { #category : #testing }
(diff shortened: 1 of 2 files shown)


src/Kernel/Slot.class.st
```diff
@@ -43,19 +43,13 @@ Slot class >> checkValidName: aSymbol [
...
+	(#(#true #false #nil) includes: aSymbol) ifTrue: [
 		^ InvalidSlotName signalFor: aSymbol ].
 
-
 	(aSymbol allSatisfy: [ :aCharacter | aCharacter isAlphaNumeric or: [ aCharacter = $_ ] ]) ifFalse: [
 		^ InvalidSlotName signalFor: aSymbol ]
 ]
 
-{ #category : #testing }
-Slot class >> isPseudovariableName: aSymbol [
```
privat, line 55: Keeping a list of named literals into an autonomous method is ok, I think.
Maybe a method of Symbol?

privat (approved):
#13136 is still not fixed. I'm all OK to transform hard errors into warnings, but don't forget the warning part!

privat:
The real controversy will be for named literal, allowing an ivar named `true` could be a lot of fun.

MarcusDenker:
> The real controversy will be for named literal, allowing an ivar named `true` could be a lot of fun.

I think on the lowest level, we should allow as much as possible. And then on the tool level warn and discourage very odd uses and guide users to not by mistake do strange things

src/Kernel/Slot.class.st
MarcusDenker: I have just removed the check. This will allow "nil true false", but the will never be used, thus this is save to do

privat:
> I think on the lowest level, we should allow as much as possible. And then on the tool level warn and discourage very odd uses and guide users to not by mistake do strange things.

I agree, but for the specific case of named literal, there is an additional scanner/parser issue as nil true and false are not recognized as variables but as literal.
So currently. An ivar named `#true` can only access trough low-level VM stuff, high-level MOP stuff, synthetic AST, or custom scanner/parser.
But it's also an issue for any variables with a weird name, e.g. `#'1e ☭'`.

The named literal thing seems only important in literal arrays `#(true) = {true}` and `#(true) ~= #(#true)` whereas other identifiers are symbols `#(self) = #(#self)`

PR #13705, opened 2023-05-11
CompletionEngine do not automatically open empty completion menus

### LINKED ISSUE #13571: Hide autocompletion when the list is empty
It happens often that the autocompletion is empty, for example while typing a new method name that is pretty specific. In that case we have an empty completion window and often it prevents me from seeing some code that interest me.

It would be great to hide the window if we have 0 result.

privat:
Completion menu is not displayed after completion delay when it is empty.
For context, the delay starts when the last keyboard input is a plain letter.

An empty menu is still presented on explicit `tab`. It's the right behavior, as it gives an explicit GUI feedback.

Fix #13571


jecisc (approved):
Thanks

MarcusDenker (approved):

MarcusDenker (approved):

PR #13715, opened 2023-05-12
finally remove AbstractTool

MarcusDenker:
merge after https://github.com/pharo-spec/NewTools/pull/523


MarcusDenker:
The code moved was just moved, more improvement for sure are possible in future PRs

src/Manifest-Core/TheManifestBuilder.class.st
```diff
@@ -524,6 +524,36 @@ TheManifestBuilder >> removeAllToDo: fp of: ruleId version: versionId [
...
+		ifTrue: [
+			classesToRemove
+				do: [ :classToRemove |
+					classToRemove subclasses notEmpty
+						ifTrue: [
+							(self confirm: (self removedClassHasSubclassesMessageFor: classToRemove name))
+								ifTrue: [ classToRemove removeFromSystem ] ]
+						ifFalse: [ classToRemove removeFromSystem ] ] ].
+	^ result
+]
```
jecisc, line 546: This method is called only by a dead method so we can remove it because it's really weird for the manifest to implement some classes removal. 

Can be done in a next PR

jecisc (approved):

PR #13716, opened 2023-05-12
Dead code cleanup in Metacello pharo platform 

MarcusDenker:
cleanup some methods that send selectors that do not exist, but are never called because overriden in subclass


Ducasse:
May be some people use the downloadZip* methods. 
Now I'm ok to remove the methods

MarcusDenker:
all of them are overridden in the subclass with versions that use Zinc.

PR #13717, opened 2023-05-13
BlockPath

MarcusDenker:
this PR adds #literalPath for CompiledBlocks. This allows a serializer that operates under the assumption that code is the same between the system that serializes and that materializes to store bocks by storing the class name, selector and path of the block, similar to how it would store classes and methods (by class name and class name  + #selector).

This for now supports  normal blocks and constant blocks, but can be easily extenden for general clean blocks.

### DIFF
--- a/src/Kernel/CompiledCode.class.st
+++ b/src/Kernel/CompiledCode.class.st
@@ -151,6 +151,13 @@ CompiledCode >> author [
 	^''
 ]
 
+{ #category : #literals }
+CompiledCode >> blockViaLiteralPath: anArray [
+	"return the compiledBlock for a path (see CompiledBlock>>literalPath)"
+	anArray ifEmpty: [ ^self ].
+	^ (self literals at: anArray first) blockViaLiteralPath: anArray copyWithoutFirst
+]
+
 { #category : #accessing }
 CompiledCode >> bytecode [
 	self
--- a/src/Kernel/ConstantBlockClosure.class.st
+++ b/src/Kernel/ConstantBlockClosure.class.st
@@ -38,9 +38,22 @@ ConstantBlockClosure class >> numArgs: numArgs literal: aLiteral [
 		literal: aLiteral
 ]
 
+{ #category : #literals }
+ConstantBlockClosure >> blockViaLiteralPath: anArray [ 
+	"return the compiledBlock for a path (see CompiledBlock>>literalPath)"
+	^ self
+]
+
 { #category : #accessing }
 ConstantBlockClosure >> literal: anObject [
 	 literal := anObject
+]
+
+{ #category : #accessing }
+ConstantBlockClosure >> literalPath [
+	"return an array with the offsets in the literals from the homeMethod to here"
+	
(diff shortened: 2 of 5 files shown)


src/Kernel/ConstantBlockClosure.class.st
```diff
@@ -38,11 +38,24 @@ ConstantBlockClosure class >> numArgs: numArgs literal: aLiteral [
 		literal: aLiteral
 ]
 
+{ #category : #literals }
+ConstantBlockClosure >> blockViaLiteralPath: anArray [ 
+	"return the compiledBlock for a path (see CompiledBlock>>literalPath)"
+	^ self
```
privat, line 44: I'm afraid that always retuning self with cause hard to track bugs if a wrong path is given.
I like exact result and early failures.

`anArray isEmpty ifTrue: [ ^self ] ifFalse: [ raise some index error or something ]`

src/Kernel/ConstantBlockClosure.class.st
```diff
@@ -38,11 +38,24 @@ ConstantBlockClosure class >> numArgs: numArgs literal: aLiteral [
 		literal: aLiteral
 ]
 
+{ #category : #literals }
+ConstantBlockClosure >> blockViaLiteralPath: anArray [
```
privat, line 42: why `ConstantBlockClosure` and not `CleanBlockClosure`?

src/Kernel/ConstantBlockClosure.class.st
```diff
@@ -38,11 +38,24 @@ ConstantBlockClosure class >> numArgs: numArgs literal: aLiteral [
...
 { #category : #accessing }
 ConstantBlockClosure >> literal: anObject [
 	 literal := anObject
 ]
 
+{ #category : #accessing }
+ConstantBlockClosure >> literalPath [
+	"return an array with the offsets in the literals from the homeMethod to here"
+	
+	 ^self outerCode literalPath, {self outerCode indexOfLiteral: self}
```
privat, line 56: looking at `CompiledBlock >> literalPath`, I'm wondering if there is a way to move the path construction responsibility to the outercode.

`CompiledBlock >> literalPathFor:` that takes a literal, so this method could simply be `^self outerCode literalPathFor: self`

src/Kernel/CompiledCode.class.st
```diff
@@ -151,6 +151,13 @@ CompiledCode >> author [
 	^''
 ]
 
+{ #category : #literals }
+CompiledCode >> blockViaLiteralPath: anArray [
```
privat, line 155: Could the method be generalized to any literal?

src/Kernel/ConstantBlockClosure.class.st
MarcusDenker: for CleanBlockClosure is is more complex, as it can contain other clean blocks

src/Kernel/ConstantBlockClosure.class.st
MarcusDenker: ok

MarcusDenker:
I will redo this (taking the comments into account)

PR #13722, opened 2023-05-13
Adjust icon and key text spacing in ToggleMenuItemMorph

Rinzwind:
See the messages in commits 61d43abbfeaf9099 and b02c232ac7082544. Before&after screenshots of the method context menu in the System Browser are shown below. The first change is that the text for ‘Do it’, ‘Do & print it’ and ‘Accept’ is slightly shifted to the right so that these align with the text of the other menu items, which is most noticeable for the first two compared to the other ‘Do …’ menu items. The second change is that on the items ‘Source code’, ‘Breakpoints’ and ‘Debugging’, the keyboard shortcuts are shifted slightly to the left so that they are more clearly separated from the submenu markers.

<p align="center">
Before After
</p>

### DIFF
--- a/src/Morphic-Base/MenuMorph.class.st
+++ b/src/Morphic-Base/MenuMorph.class.st
@@ -1040,6 +1040,15 @@ MenuMorph >> leftArrowStroked: evt [
 ]
 
 { #category : #accessing }
+MenuMorph >> maxItemsIconWidth [
+
+	^ self menuItems inject: 0 into: [ :maxWidth :menuItemMorph |
+		menuItemMorph hasIcon
+			ifTrue: [ maxWidth max: menuItemMorph iconForm width ]
+			ifFalse: [ maxWidth ] ]
+]
+
+{ #category : #accessing }
 MenuMorph >> menuItems [
 	^submorphs select: [ :m | m isMenuItemMorph ]
 ]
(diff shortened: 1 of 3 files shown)


Rinzwind:
I considered similarly applying a method `#maxItemsSubMenuMarkerWidth` in `ToggleMenuItemShortcut>>#boundsForKeyText:font:`, but the resulting visual distinction between menus with and without submenus seemed a bit odd.

The spacing between the key text and submenu marker should maybe be increased a bit further.

The vertical alignment within each item morph is not changed in this pull request, but is something that could use improvement as well:

<p align="center">
Alignment
</p>

Ducasse:
Thanks Kris. This is good to have people fixing this. 
BTW we are working also on new widgets and the menus are gorgeous there.

Ducasse:
This is true that I do not understand why doit is displayed as command D (while we do not press shift) and why debugging is shift command h. 

BTW I tried to see the difference between the two screenshots you showed but I could not spot the difference.

Rinzwind:
The differences are subtle, and were more clear to me given that I had been staring at it for a while, I have edited my first message now to explicitly point them out.

guillep:
Oh this is nice! Thanks!!

Issue #12420, opened 2023-01-23
Equality between Symbols and Strings having same characters sequences "wrongfully" returns true


fedemennite:
**Bug description**
Equality between Symbols and Strings having same characters sequences "wrongfully" returns true.
Many Smalltalk dialects such as Gemstone would return false.

**To Reproduce**
These are not failing:
self assert: ('string' = #string)
self assert: (#string = 'string')

**Expected behavior**
These should not fail instead
self assert: ('string' = #string) not
self assert: (#string = 'string') not

**Screenshots**
None

**Version information:**
 - OS: Any
 - Version: Any
 - Pharo Version 10

**Expected development cost**
Fixing the equality implementation for symbols is quite easy.
However there is code in the Pharo image that breaks under this assumption.

**Additional context**
Changing this behavior would increase code portability across dialects, making easier to have consistent and portable code behavior.
The proposed change can possibly have impact on a number of existing packages.
Therefore a first level of fixing this could be to make this behavior configurable, so that core classes (and possibly third party packages) can be tested against the new more restrictive behavior while old code can continue to work depending on the user needs.
In any case such a change should likely be targeted for new major version of Pharo.

As a side note: I personally have no doubts on the new proposed behavior, however this change could also spawn a philosophical discussion on what should be the best behavior.

MarcusDenker:
This would be a quite big change... we rely on equality between strings and symbol in many, *many* places... 

A lot a lot of work. Is there any other good reason for this other than compatibility?

fedemennite:
It seems to me also an easy way to miss an explicit way (explicit code) to rely on the speed (identity = comparison) and lower memory consumption implied by the symbols uniqueness. Basically one the symbols core purpose.

fedemennite:
FYI, this issue implicitly caused issue https://github.com/pharo-project/pharo/issues/12518
Of course there we can have a relatively quick fix without having to solve this issue now

MarcusDenker:
I think for this the right path is to write a "Pharo Enhancement Proposals" (see https://github.com/pharo-project/pheps)

This would
- explain all the positive (why this is needed and better)
- possible negative aspect
- Analysis of the Pharo core of where it relies on equality and now simple it is fix those
- Impact on users of Pharo. What is a migration path for them?

[...]

PR #13727, opened 2023-05-15
Implements optional strict Symbol and String comparison

carolahp:
[**This is a draft PR, do not review yet please**]

Seeks to solve the issue https://github.com/pharo-project/pharo/issues/12420

Currently in Pharo, the result of comparing a symbol and a string with the same contents is true. 
For example:
```#hello = 'hello'``` returns `true`.

This PR seeks to implement "strict" comparison of strings and symbols, meaning that the previous example should return false

For this:
- first (to prevent system crashes) we must fix all comparisons between strings and symbols from Pharo's core, replacing them by symbol/symbol comparisons
- secondly, we must modify the comparison methods to make them 'strict' depending on a configuration setting

image


MarcusDenker:
If you fix a merge conflict, why not remove the "merge conflicts" tag, too?

PR #13729, opened 2023-05-15
Remove dead code in session manager

jecisc:
This method is never called in the image or the bootstrap scripts.


MarcusDenker (approved):

jecisc:
The error seems to be a random failure even if I've never seen it before.

PR #13730, opened 2023-05-15
Method announcements should use real protocols

jecisc:
We have different method announcements that deal with protocols but were using protocol names up to now.

I updated those to be able to use real protocols instead of protocol names. This will allow us to simplify some code in the future. 

Updated classes:
- MethodAdded
- MethodRemoved
- MethodUpdated
- MethodRecategorized

Future steps:

I found two methods at least that can now be simplified with this change:
- RPackageOrganizer>>#systemMethodRecategorizedActionFrom:
- MCPackageManager class>>#managersForClass:category:do:

Those two methods are in parts of the system that are too complicated for what they are doing and the use of a real protocol could simplify them a little bit. If this is integrated, I'll simplify them


jecisc:
Apparently this makes the tests timeout.

MarcusDenker (approved):

PR #13731, opened 2023-05-15
Simplify Fuel management

jecisc:
Fuel is a project that work in multiple smalltalk versions and there is a compatibility system with FuelPlatform to keep things compatible. 

This seems overkill because since years now the Pharo version of Fuel is part of Pharo and is never loaded from an external repository. This means that the Pharo version is managed in Pharo and has no reason to be loaded anywhere else. 
Since the compatibility management adds 4 packages, lot of code, complexifies the code and is not really useful in our case, maybe, we should just keep the Pharo version of things in Pharo. 

This compatibility thing also add an extra step for us at each release because we need to create a new platform for fuel at each release.

This PR removes those 4 packages and simplifies fuel management. Removing a release step at the same time.


MarcusDenker:
I think it is no secret that I am not a fan of the "it has to work on all smalltalks (and all versions of those), even if I have to implement a whole ST80 to get there!" approach to things...

Fuel should see Pharo as something that it can improve to make implementing a serializer simpler, instead of using no new features due to "compatibility". E.g. the serializer should forward to the layout instead of checking with #IsBytes... or it should impove the API for binary streams for everyone, instead of adding it's own methods.

For Fuel things are even harder: How to support new features (DoubleByteArray, Ephemerons, remove of classDescription, trailer simplifications, clean blocks....) without this leading to an *explosion* in complexity if you have to parametrize all that?

But: Fuel has a maintainer, and was always synced upstream. 
For now I add issue tracker entries for all the changes I have to do to Fuel for the cleanups. 

https://github.com/theseion/Fuel/issues

But I honestly do not have any idea how to do these in a code base that is supposed to be backward compatible to all old Pharo versions and even Squeak?



<blockquote><div> GitHub</div><div><strong><a href="https://github.com/theseion/Fuel">Issues · theseion/Fuel</a></strong></div><div>Fuel, the Smalltalk object serializer. Contribute to theseion/Fuel development by creating an account on GitHub.</div></blockquote>

jecisc:
I heard that in general, something exported in a version of a smalltalk or even a version of Pharo is not materializeable in other smalltalks/versions of Pharo. 

If this is the case, I think that keeping the project in one place and managing all versions of Pharo/Other smalltalks will just add more and more and more complexity for really little gain. 

I didn't see much new features of Fuel recently and I have the impression that we spend more time on compatibility than on other things.

MarcusDenker:
Fuel (for now) assumes that the code is the same in both images (the one used to serialize, then the one that materializes). 

There is some support to serialize CompiledMethod and CompiledBlocks (e.g. when serializing a long existing block whose home method has already been modified). There is some support to store classes, there is some support for code changes when ivars are removed/added/re-ordered.  

In practice I think this can work to some extend to move instances from one image to another with changed code, But there is no real code history support and it will be quite brittle, I think.

theseion:
I agree with all you've said. I wrote the compatibility mainly for my own sanity because it made it simpler for me to maintain it for both Pharo and Squeak for multiple versions at the same time. Up until now it was always more or less straight forward to make Fuel work with changes in Pharo and Squeak but these big changes that you're doing right now are definitely something else. I'm not going to invest time into trying to keep this compatible with Squeak and older Pharo versions. Fuel can deal with code changes between images but not always with big changes such as between major versions, so it's not a huge deal to make a break.

I think the way forward for Pharo will be to have distinct versions for each version of Pharo. Not sure yet about how I'll deal with version numbers but that doesn't have to concern you much. I'll probably also get rid of version specific platform classes, maybe of FuelPlatform in general at some point but it would be too much work to just drop FuelPlatform just yet.

theseion:
And now I that @jecisc already did all the work 😆. Thanks!

theseion:
I've integrated your changes and will open a new PR for upgrading Fuel. You can close this PR.

jecisc:
Thank you max!
I was gonna contact you to discuss about this try PR but I got busy :)

theseion:
In https://github.com/pharo-project/pharo/pull/14132

PR #13738, opened 2023-05-16
Cleanup-MetacelloPharoPlatform 

MarcusDenker:
The plattform code of Pharo is split in two packages, which I think makes no sense (we will never load this code in an older Pharo, but backport fixes there)

- Move everything from Metacello-Platform to Metacello-PharoCommonPlatform
- move all methods up to MetacelloPharoCommonPlatform
- remove all unsent messages

The empty class and package stay there as we have to first edit the bootstrap script


jecisc (approved):
Yes! :)

PR #13736, opened 2023-05-16
remove Object bindingOf: and hasBindingOf:

MarcusDenker:
- implement them in the places where normal use needs them -- there might be more

PR #13739, opened 2023-05-16
13723-Binding-for-variable-in-Playground-is-nil-after-assigning-a-number-to-it 

MarcusDenker:
Check for identity, not equality


privat (approved):
Cood catch. I saw the issue but did not found what I did wrong.

jecisc (approved):

jecisc:
Failure seems unrelated

MarcusDenker:
failing tests are due to a spec change got merged, but it needs https://github.com/pharo-project/pharo/pull/13736 to be green (why the spec test run was ok for that I do not undstand)

PR #13745, opened 2023-05-16
Move-isObservableSlot

MarcusDenker:
#isObservableSlot was part of a deprecated package, move it to the correct package

### DIFF
--- a/src/VariablesLibrary/Slot.extension.st
+++ b/src/VariablesLibrary/Slot.extension.st
@@ -1,0 +1,7 @@
+Extension { #name : #Slot }
+
+{ #category : #'*VariablesLibrary' }
+Slot >> isObservableSlot [
+	
+	^ false
+]


MarcusDenker:
We need to merge this here before merging the removal there: https://github.com/pharo-spec/Spec/pull/1393

jecisc (approved):

MarcusDenker:
This sadly does not work: https://github.com/pharo-spec/Spec/pull/1393 fails with this method missing, even though it is now in another package.

jecisc:
This is weird because I see the method that was overridden in the latest image: 
image


Maybe the build did not take the latest image yet?

jecisc:
I have a similar case with: https://github.com/pharo-spec/ScriptableDebugger/pull/65

The method is present in the core Pharo. And this PR removing it from another project is failing

PR #13751, opened 2023-05-17
turn on clean blocks, second try

MarcusDenker:
- try to us signalContext in HRTestReport


MarcusDenker:
So, the fix seems to work! But we have lots of failing tests in the debugger and Fuel.

I made an issue for those.

I wil close this and later commit the fix for the HDTestRunner

PR #13756, opened 2023-05-17
Merge SHTextStyler and SHRBTextStyler

jecisc:
SHTextStyler is never used and only has one subclass. I porpose to merge both of them for now. 

With this change, we can inherit from RBProgramNodeVisitor instead of using TRBProgramNodeVisitor. 

With this, we have no user of TRBProgramNodeVisitor. I think we should deprecate and remove it in the future because it cause us to have to maintain 2 full API of the AST visitor increasing the complexity of the AST maintenance. Especially since the visitor is still been improved (for example to manage better the comments). What do you think about this next step?


src/Shout/SHRBTextStyler.class.st
```diff
@@ -3,8 +3,7 @@ I'm the one who visits node to coloring the code
 "
 Class {
 	#name : #SHRBTextStyler,
-	#superclass : #SHTextStyler,
-	#traits : 'TRBProgramNodeVisitor',
+	#superclass : #RBProgramNodeVisitor,
 	#classTraits : 'TRBProgramNodeVisitor classTrait',
```
MarcusDenker: I guess the trait is not needed anymore?

MarcusDenker: I wonder why the trait was removed but the class side not?

MarcusDenker:
SycRemoveClassStrategyTest>>testCreateStrategiesForUsedTrait is failing, it uses hat trait to test something that is now failing after it is not used

jecisc:
I updated the tests to use a trait that is directly from Calypso (in a packages that is a dependency of the Calypso package been tested) and that has way more users.

MarcusDenker (approved):

src/Shout/SHRBTextStyler.class.st
jecisc: For the class side, this is a bug I had often when working on Moose. I reported it but I never had the time to check the origin and fix it :(

And no, the trait is not needed anymore and I wonder what to do with it. 
- Keep it for external users that might want a visitor that does not inherit from the actual visitor
- Deprecate it and remove it in P13
- Remove it (I prefer option 2)

I would vote to deprecate it and remove it in P13 because the actual visitor cannot use the Trait IIRC because it is loaded before the traits are in the image. And it causes a big duplication that can be hard to maintain since we are improving the AST a lot recently.

PR #13758, opened 2023-05-17
Fix #13478: Method search now only applies to selector

### LINKED ISSUE #13478: Calypso: Method search does not filter properly with class name keyword
**Bug description**
Searching for methods with a word contained in the class name doesn't filter correctly.

**To Reproduce**
1. Open browser on a class, for example `Collection`
2. Use cmd+f to open a "Choose method" window (ClyBrowserSearchDialogWindow)
3. Enter a filter to search for a method whose name contains the class name, for example `collection` (or even `collect`)
4. See that all methods are still present, even if their name do not match the filter

**Expected behavior**
Only methods matching the filter should appear, regardless of the name of the class they belong to.

**Screenshots**
image

**Version information:**
All Pharo versions.

**Note**
If this is indeed considered as a bug by the team, I will consider making a PR.

Gabriel-Darbord:
The issue was that the query was using `ClyItemNameFilter` which sends `name` to a `CompiledMethod`, which returns something like `MyClass>>mySelector`.
This fix proposes to add `ClyItemSelectorFilter` specialized for methods (and possibly other items?) that only ask for the `selector`.
To determine which to use, the `isMethodQuery` testing method is added to the `ClyQuery` hierarchy, and `ClyQueryViewMorph>>#enableFilterUsing:` is modified to use it to make the choice.

### DIFF
--- a/src/Calypso-Browser/ClyQueryViewMorph.class.st
+++ b/src/Calypso-Browser/ClyQueryViewMorph.class.st
@@ -343,7 +343,11 @@ ClyQueryViewMorph >> enableFilter: anItemStringFilterClass using: aStringPattern
 { #category : #controlling }
 ClyQueryViewMorph >> enableFilterUsing: aStringPattern [
 
-	self enableFilter: ClyItemNameFilter using: aStringPattern
+	self
+		enableFilter: (selection rootDataSource query isMethodQuery
+				 ifTrue: [ ClyItemSelectorFilter ]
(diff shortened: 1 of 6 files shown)


jecisc:
I talked with Gabriel IRL but for the sake off logging information:

I have the impression it would be better to improve the Name filter to filter on the selector for compiled method instead of adding a new filter.

jecisc:
I'm closing because a better fix has been added in another PR

PR #13767, opened 2023-05-18
Remove unused UITheme forms and add support for scale 2.0 icons

Rinzwind:
This removes unused ‘forms’ from UITheme.


MarcusDenker (approved):

Rinzwind:
I have added three more commits that further clean up the ‘forms’ in UITheme, and two that extend ThemeIcons to support using the [scale 2.0 icons from the default icon pack](https://github.com/pharo-project/pharo-icon-packs/tree/16fe361b60871496735fff56b4c1dc93d36924b9/png-scale2.0). This improves the icons when using display scale factor 2:

```smalltalk
WorldMorph displayScaleFactor: 2.
SourceCodeFonts setSourceCodeFonts: SourceCodeFonts sizeSmall * WorldMorph displayScaleFactor.
ToggleWithSymbolMenuItemShortcut
	classVarNamed: 'SymbolFont'
		put: (LogicalFont familyName: 'Lucida Grande' pointSize: 10 * WorldMorph displayScaleFactor);
	classVarNamed: 'SymbolTable' put: nil.
ThemeIcons current: ((ThemeIcons named: 'svgPack') in: [ :themeIcons |
	(themeIcons respondsTo: #loadIconsFromUrlUsingScale:) ifTrue: [
		themeIcons loadIconsFromUrlUsingScale: WorldMorph displayScaleFactor ] ];
	yourself).
```

Before&after showing the top of a System Browser:

<p align="center">

</p>

<p align="center">

</p>

The world scale factor can be set to the reciprocal of the display scale factor for ‘Retina display’ scaling:

```smalltalk
World scaleFactor: WorldMorph displayScaleFactor reciprocal
```

The World menu:

<p align="center">

</p>

One question that still needs to be resolved: how should the new instance variable ‘scale’ be initialized for the current ThemeIcons instance when this branch gets checked out in an existing image? Do I just use an accessor with lazy initialization?

Ducasse:
Thanks this is super cool!
BTW do you know that the future is https://github.com/plantec/Toplo, if you want to contribute they are working on theme and skins.

<blockquote><div> GitHub</div><div><strong><a href="https://github.com/plantec/Toplo">GitHub - plantec/Toplo: A widget framework on top of Bloc</a></strong></div><div>A widget framework on top of Bloc. Contribute to plantec/Toplo development by creating an account on GitHub.</div></blockquote>

Rinzwind:
I didn’t know about that, I will take a look!

A problem with the ‘Retina display’ scaling snippets I gave above is that text becomes too big in some areas because of the additional font scaling done in SpFontStyle. I have opened a pull request to, as a temporary solution, add a class variable to turn that off (https://github.com/pharo-spec/Spec/pull/1397):

```smalltalk
SpFontStyle applyDisplayScaleFactor: false
```

Comparison of inspectors before and after setting it to false:

<p align="center">
 
</p>

src/Polymorph-Widgets/ThemeIcons.class.st
```diff
@@ -300,20 +301,20 @@ ThemeIcons >> iconNamed: aSymbol [
...
 ]
 
 { #category : #accessing }
 ThemeIcons >> iconNamed: aSymbol ifNone: aBlock [
 
-	self icons at: aSymbol asSymbol ifPresent: [ :icon | ^ icon ].
+	self icons at: aSymbol asSymbol ifPresent: [ :icon | ^ icon scaledToSize: icon extent * (self currentWorld displayScaleFactor / scale) ].
 	"Trying the old way"
 
 	((aSymbol endsWith: 'Icon') or: [ (aSymbol endsWith: 'Form') ]) ifTrue: [
```
MarcusDenker, line 314: Not for this PR: but we really need to find a way to get rid of the "old way", to force people to change the names. I did a session to reduce #asSymbol sends during normal use, this happend *a  lot* that this code path was taken. I checked again, and, oh wonder, *new code* was written that end here agan, for every time we open a menu. 

Why even change to a "new way" if there is no way to slowly refactor away from the "old way"?

(not for this PR)

MarcusDenker (approved):
For me it looks good, but the problem is that I have not really any knowledge about this part of the system.

Therefore we need another review in addition to mine

src/Polymorph-Widgets/ThemeIcons.class.st
Ducasse: We should add more rules :)

Rinzwind:
Thanks for merging it! I think my question on the initialization of ‘scale’ in ThemeIcons may have been missed? Using a build 374 image, pulling in the latest commit will break because ‘scale’ is not initialized. Here’s a fix for that: commit 19db8c4267262740.

Ducasse:
argh!
Hi kris you see this is important to make simple autonomous issues because else it requires a lot of brain power from us. 
Can you do a new PR with the changes. Because I do not know how to easily integrate the commit.

Ducasse:
Else I will redo the change and submit a PR.

Ducasse:
I' m redoing your changes and producing a PR now.

Rinzwind:
Sorry, I just wasn’t quite sure how to handle it. I considered using slot initialization but that didn’t do what I expected: see issue #13839. Using `#scale => InitializedSlot default: 1` would still leave the variable nil on the existing current ThemeIcons. It might also have been somewhat unclear as the variable doesn’t need to be initialized for a new ThemeIcons: it gets set in the `#loadIconsFromUrlUsingScale:` method. In GemStone/S, I would have implemented an override of `#migrateFrom:instVarMap:`, see the [section on ‘Customized Instance Variable Mappings’](https://downloads.gemtalksystems.com/docs/GemStone64/3.6.x/GS64-ProgGuide-3.6/10-ClassHistory.htm#pgfId-828058) in chapter 10 ‘Class versions and Instance Migration’ in the GemStone/S v3.6 Programming Guide which says: “When you migrate an object, `migrateFrom:instVarMap:` is sent to the new instance with the argument of the old instance.” I couldn’t find an equivalent for that in `#migrateInstancesTo:` and `#copyObject:to:` in ShiftClassInstaller.

carolahp:
Hi @Rinzwind thanks for this contribution. I am sorry I got to check this only today. This is exactly what we were missing and the reason why I created the 3 versions of icons. 

@Ducasse and @MarcusDenker I can review the new [PR](https://github.com/pharo-project/pharo/pull/13837/files)

Ducasse:
Hi caro

Please send a new PR if I break something. 

S

PR #13768, opened 2023-05-19
Use more protocol object in RPackage

jecisc:
The main change of this PR is to use more real protocols instead of protocol name. But this also includes some cleanings such as dead code removal or variable renaming.


MarcusDenker:
failing tests look related

jecisc:
I fixed a test. A second one still fails, I'll take a look later

Ducasse:
Lot of tests are broken

jecisc:
It's hard to track all implications of those changes so I'll probably extract one portion in another PR to wake it easier. 

RPackage is complex and it's not easy to unify things

jecisc:
3 tests are still failing but it's way less than before :)

jecisc:
Tests finally passed on this one :)

MarcusDenker (approved):

PR #13777, opened 2023-05-21
Add #getTemporaryDirectory to UnixPlatform and WinPlatform

gcorriga:
Add #getTemporaryDirectory to UnixPlatform and WinPlatform, with related tests.

Do not add yet an abstract #getTemporaryDirectory to OSPlatform as macOS might require additional work to support NSTemporaryDirectory.


src/System-Platforms-Tests/UnixPlatformTest.class.st
```diff
@@ -0,0 +1,24 @@
...
+	#category : #'System-Platforms-Tests-Unix'
+}
+
+{ #category : #tests }
+UnixPlatformTest >> testGetTemporaryDirectoryFromTMPDIR [
+
+	| value |
+	OSPlatform current isUnix ifFalse: [ ^ self skip ].
+
+	value := OSEnvironment current at: 'TMPDIR' ifAbsent: [  ].
```
tesonep: Why not return nil in the ifAbsent block, it is more clear.

tesonep:
```suggestion
	value := OSEnvironment current at: 'TMPDIR' ifAbsent: [ nil ].
```

src/System-Platforms-Tests/WinPlatformTest.class.st
```diff
@@ -0,0 +1,21 @@
...
+	#category : #'System-Platforms-Tests-Windows'
+}
+
+{ #category : #tests }
+WinPlatformTest >> testGetTemporaryDirectoryFromTMP [
+
+	| value |
+	OSPlatform current isWindows ifFalse: [ ^ self skip ].
+
+	value := OSEnvironment current at: 'TMP' ifAbsent: [  ].
```
tesonep: Why not return nil in the ifAbsent block, it is more clear.

```suggestion
	value := OSEnvironment current at: 'TMP' ifAbsent: [ nil ].
```

tesonep (changes requested):

jecisc:
I am wondering what is the use of those new methods? We already have `FileLocator temp`, what is the difference?

gcorriga:
@jecisc I was not aware of
```smalltalk
FileLocator temp
```

Looking at the underlying behaviour it doesn’t appear to be correct - e.g. for Windows the TMP env variable should have precedence over TEMP.
I will rework my PR to move the correct behaviour in the right place.

jecisc:
Thank you! :)

jecisc:
The tests passed and now it improves the current implementation so it's way better!

The code seems good but I'm not an expert on this part of the system so another review might be nice

Rinzwind:
The [output of the Windows test running step in build 3 of pull request #16337](https://ci.inria.fr/pharo-ci-jenkins2/job/Test%20pending%20pull%20request%20and%20branch%20Pipeline/job/PR-16337/3/execution/node/81/log/?consoleFull) shows the implementation of `#temp` on WindowsResolver introduced in this pull request in commit 78e0652dae75e007 is causing an error:

```
<testsuite name="Epicea-Tests" tests="55" timestamp="12:39:40.92 pm" seed="1053542408"  failures="0" errors="110" time="24.537">                                                            
	<testcase classname="Windows64.Epicea.Tests.EpAnnouncementsTest" name="testMonitorAnnouncesUpdateWhenDisabled" time="0.631">
		<error type="SymbolNotFoundError" message="Could not find symbol named: #GetTempPath2W searching in module: 'Kernel32'">
SymbolNotFoundError
Could not find symbol named: #GetTempPath2W searching in module: 'Kernel32'
TFFIBackend>>primLoadSymbol:module:
[…]
Win64Platform(WinPlatform)>>getTempPath:buffer:
Win64Platform(WinPlatform)>>getTempPath
WindowsResolver>>temp
[…]
EpTestLogBuilder>>newDirectory
[…]
EpAnnouncementsTest(EpMonitorIntegrationTest)>>setUp
		</error>
```

This might be due to the Jenkins agent not meeting the [requirements for ‘GetTempPath2W’](https://learn.microsoft.com/en-us/windows/win32/api/fileapi/nf-fileapi-gettemppath2w#requirements) (‘Windows 10 Build 20348’ / ‘Windows Server Build 20348’).

@gcorriga: Could [‘GetTempPathW’](https://learn.microsoft.com/en-us/windows/win32/api/fileapi/nf-fileapi-gettemppathw) be used instead?

gcorriga:
@Rinzwind I believe it should be possible to use GetTempPathW. No issues from me if you want to go ahead with this change.

Please note, I am on paternity break until April 2nd so I won't be able to help much.

Rinzwind:
OK, I have opened a pull request to change it: pull request #16342.

PR #13778, opened 2023-05-21
Fix #13478: Filter on `calypsoName` instead of `name`

### LINKED ISSUE #13478: Calypso: Method search does not filter properly with class name keyword

Gabriel-Darbord:
As discussed with @jecisc in #13758 regarding #13478.

Add the new `calypsoName` method to RPackage, Class, CompiledMethod and ClyBrowserItem (are there others?).
For a CompiledMethod, this method answers its selector, fixing the issue.
For the others, it has the same behavior as calling the `name` method.

### DIFF
--- a/src/Calypso-NavigationModel/Class.extension.st
+++ b/src/Calypso-NavigationModel/Class.extension.st
@@ -1,0 +1,7 @@
+Extension { #name : #Class }
+
+{ #category : #'*Calypso-NavigationModel' }
+Class >> calypsoName [
+
+	^ self name
+]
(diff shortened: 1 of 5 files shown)


Gabriel-Darbord:
The failed CI looks like the fabled random test failure...

jecisc:
This failing test appears a lot recently

jecisc (approved):
I prefer that to the other PR! :)

jecisc:
Maybe adding a tests can be good :)

PR #13782, opened 2023-05-22
organization, not organisation

MarcusDenker:
Just fix a case we #organisation is used instead of  the #organization

### DIFF
--- a/src/Gofer-Core/GoferCleanup.class.st
+++ b/src/Gofer-Core/GoferCleanup.class.st
@@ -27,7 +27,7 @@ GoferCleanup >> cleanupCategories: aWorkingCopy [
 GoferCleanup >> cleanupProtocols: aWorkingCopy [
 
 	aWorkingCopy packageSet extendedClasses do: [ :class |
-		(aWorkingCopy packageSet extensionCategoriesForClass: class) do: [ :protocolName | class organization removeProtocolIfEmpty: protocolName ] ].
+		(aWorkingCopy packageSet extensionCategoriesForClass: class) do: [ :protocolName | class removeProtocolIfEmpty: protocolName ] ].
 
 	aWorkingCopy packageSet classesAndMetaClasses do: [ :class |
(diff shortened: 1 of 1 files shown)


src/Gofer-Core/GoferCleanup.class.st
```diff
@@ -27,7 +27,7 @@ GoferCleanup >> cleanupCategories: aWorkingCopy [
 GoferCleanup >> cleanupProtocols: aWorkingCopy [
 
 	aWorkingCopy packageSet extendedClasses do: [ :class |
-		(aWorkingCopy packageSet extensionCategoriesForClass: class) do: [ :protocolName | class organisation removeProtocolIfEmpty: protocolName ] ].
+		(aWorkingCopy packageSet extensionCategoriesForClass: class) do: [ :protocolName | class organization removeProtocolIfEmpty: protocolName ] ].
```
jecisc:
```suggestion
		(aWorkingCopy packageSet extensionCategoriesForClass: class) do: [ :protocolName | class removeProtocolIfEmpty: protocolName ] ].
```

src/Gofer-Core/GoferCleanup.class.st
```diff
@@ -27,7 +27,7 @@ GoferCleanup >> cleanupCategories: aWorkingCopy [
...
 
 	aWorkingCopy packageSet classesAndMetaClasses do: [ :class |
 		(aWorkingCopy packageSet coreCategoriesForClass: class) do: [ :category | class organization removeProtocolIfEmpty: category ] ]
```
jecisc:
```suggestion
		(aWorkingCopy packageSet coreCategoriesForClass: class) do: [ :category | class removeProtocolIfEmpty: category ] ]
```

jecisc (commented):
We can do even better because I added #removeProtocolIfEmpty: to ClassDescription already even if it just delegates to the ClassOrganization for now since it manipulates directly an instance variable

PR #13783, opened 2023-05-22
Inline ClassOrganization>>extensionProtocols

jecisc:
Still inlining ClassOrganization into ClassDescription


MarcusDenker (approved):

jecisc:
The error is the coverage test randomly failing

MarcusDenker (approved):

MarcusDenker:
The build now fails with

```
 Instance of ClassOrganization did not understand #extensionProtocols
```

PR #13784, opened 2023-05-22
merge Pharo11 changes into Pharo12

estebanlm:
subject :)


jecisc:
The error is weird.

It happens because the Pharo 12 fuel platform is not loaded before the post load is executed. Because of that it selects another platform that has broken code for Pharo 12 (which is why I think we should simplify Fuel management: https://github.com/pharo-project/pharo/pull/13731).

But I don't understand why the package was not loaded because this PR does not touch the BaselineOfFuelPlatform

jecisc:
The problem is that the version is considered to be Pharo 11 in this build:

image

estebanlm:
uhm... yes. because there is no tag.

jecisc:
This seems to still be built for Pharo 11

estebanlm:
yeah, PR do not take into account tags made in the clone, apparently. So we have a problem.

jecisc:
It's weird that all the other synched of P11 in P12 I did before were able to build and not this one

estebanlm:
someone removed FLPharo11Platform, I will reintroduce it just to make this pass

jecisc:
An easier solution would be to update BaselineOfFuelPlatform>>#baseline: like this:

```st
baseline: spec

	<baseline>
	spec for: #common do: [
		spec postLoadDoIt: #addPostloadHacks.

		spec
			package: 'Fuel-Platform-Core';
			package: 'Fuel-Platform-Pharo-Core' with: [ spec requires: 'Fuel-Platform-Core' ];
			package: 'Fuel-Platform-Pharo-12' with: [ spec requires: 'Fuel-Platform-Pharo-Core' ].

		spec group: 'FuelPlatformCurrentGroup' with: #( 'Fuel-Platform-Pharo-12' ) ]
```

estebanlm:
easier than what?
no, no... I will reintroduce the platform (temporary), modifying baselines is the door of hell.

jecisc:
Fuel-Platform-Pharo-11 package was also removed from Pharo 12 so you will need to introduce it in the baseline anyway.

Here the problem is that since the Fuel-Platform-Pharo-12 package is loaded like this:

```st
spec for: #'pharo12.x' do: [
		spec package: 'Fuel-Platform-Pharo-12' with: [ spec requires: 'Fuel-Platform-Pharo-Core' ].
		spec group: 'FuelPlatformCurrentGroup' with: #( 'Fuel-Platform-Pharo-12' ) ]
```

it is not loaded if the version number of Pharo is 11 and not 12. Because #'pharo12.x.x' will not be in the attributes. But we can just move that to the #common group since we know that we are in Pharo 12 and if we keep the same management of Fuel, we should have Fuel-Platform-Pharo-13 in Pharo 13 instead of Fuel-Platform-Pharo-12. 

(I don't like this management and I did a PR to simplify it but that requires some discussion)

jecisc:
Oh, but I see that since the Major version of Pharo will be 11 and not 12, it will not select this platform anyway... So I get it that it's easier to add back the Pharo 11 platform. Sorry for the noise

estebanlm:
> Oh, but I see that since the Major version of Pharo will be 11 and not 12, it will not select this platform anyway... So I get it that it's easier to add back the Pharo 11 platform. Sorry for the noise

is how the bootstrap works. It recognizes the pharo version through the tags assigned (it looks the biggest/closest) tag named vN.N.N*. Since my merge start point is previous the v12.0.0, it takes that version (v11.0.0) and hence it assigns that number to the PR. This will be fixed in the future, but this case is particular. 

Anyway, this would have not happened if the branches would not have diverged so much, and for so much time.

PR #13785, opened 2023-05-22
Reduce users of #organization

jecisc:
Reduce the number of senders of #organization by finishing the inline of #protocolNames


jecisc:
This CI immediately failed.. Weird..

jecisc:
I also deprecated #allMethodSelectors because it is now equivalent to ClassDescription>>#selectors since I made sure that the methods are now always categorized.

jecisc:
The CI seems totally broken on this one but I'm not sure it's linked to my changes.
I don't really know on this one.

PR #13789, opened 2023-05-23
Migrate variables of VariableLayout only if old instance was also variable

jecisc:
When we recompile a class with a variable layout, we migrate the variables from the old objects to the new instances. But we should not do that if the old object was not a variable layout already because then it will try to access fields that are not accessible via #basicAt:

### DIFF
--- a/src/Shift-ClassInstaller/ShiftClassInstaller.class.st
+++ b/src/Shift-ClassInstaller/ShiftClassInstaller.class.st
@@ -74,26 +74,28 @@ ShiftClassInstaller >> copyClassSlotsFromExistingClass [
 
 { #category : #copying }
 ShiftClassInstaller >> copyObject: oldObject to: newClass [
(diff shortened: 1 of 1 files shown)


MarcusDenker (approved):

MarcusDenker:
We should in a later step forward this work to the class Layout. In Soil (https://github.com/ApptiveGrid/Soil) this lead to a much easier to understand code for  serializing/deserializing (and specially it one does not forget a Layout, e.g. this code here can not work with CompiledMethodLayout, and nobody knows...)

<blockquote><div> GitHub</div><div><strong><a href="https://github.com/ApptiveGrid/Soil">GitHub - ApptiveGrid/Soil: A playground for a pharo OO database</a></strong></div><div>A playground for a pharo OO database. Contribute to ApptiveGrid/Soil development by creating an account on GitHub.</div></blockquote>

PR #13792, opened 2023-05-23
Inline ClassOrganization>>#protocolNamed:[ifAbsent:] 

jecisc:
Move the methods to ClassDescription and update senders. 

I also removed a method that was dead code (#removeAllMethodsInProtocol:). It was also calling methods that do not exists.

Requires https://github.com/pharo-project/pharo/pull/13785 before merging.


jecisc:
Failing tests are not related

MarcusDenker (approved):

PR #13795, opened 2023-05-24
13791-PseudoVariables-should-be-read-via-DoItVariable 

### LINKED ISSUE #13791: PseudoVariables should be read via DoItVariable
When executing a DoIt in the Debugger, self, super, thisContext accesses are compiled to the push bytecodes.

This means they push the state of the execution of the DoIt method, not the values of the context that the debugger looks at.

If we add to ReservedVariable the same method as we have for temps:

```
asDoItVariableFrom: aContext
	^ DoItVariable fromContext: aContext variable: sel
```

the lookUp done when compiling the doIt will find the shadowing DoItVariable and force code generation for the reflective path, ending up calling #readInContext:    

For thisContext, is means that it will return the context that we look at in the debugger, for the others it means we can improve what is returned later easily by improving  #readInContext: , e.g. when reading self in a block, we could improve by return the home context receiver, not the current context receiver, for example

MarcusDenker:
- add #asDoItVariableFrom: to ReservedVariable
- add tests for ThisContextVariableTest 
- remove empty ReservedVariableTest

fixes #13791

### DIFF
--- a/src/Kernel/ReservedVariable.class.st
+++ b/src/Kernel/ReservedVariable.class.st
@@ -37,6 +37,11 @@ ReservedVariable class >> variableName [
 ]
 
 { #category : #converting }
+ReservedVariable >> asDoItVariableFrom: aContext [
+	^ DoItVariable fromContext: aContext variable: self
+]
+
+{ #category : #converting }
 ReservedVariable >> asString [
 
 	^ self name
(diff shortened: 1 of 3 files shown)


MarcusDenker:
I think the problem is not due to this PR, but not sure, I added https://github.com/pharo-project/pharo/issues/13796

MarcusDenker:
One test has to be fixed, sadly it is faster to re-create the PR than to update it...

MarcusDenker:
New PR:

https://github.com/pharo-project/pharo/pull/13812

PR #13798, opened 2023-05-24
Disable Epicea debug mode in tests

### LINKED ISSUE #13796: oldProtocol often nil in ClassDescription>>#notifyRepackage:method:oldProtocol:newProtocol:
**Bug description**
On the CI we see a lot

```
 EpMonitor could not log a change. You can enable debugging with "EpMonitor current debugMode: true".
```

If I do that, I see a lot of teste failing when e.g. running the Newtools debugger tests

For these cases, oldProtocol nil in ClassDescription>>#notifyRepackage:method:oldProtocol:newProtocol: and we get a DNU for #name.

If EpMonitor debugMode is off, the tests are green

jecisc:
Epicea has a debug mode to catch errors. This is good for the users to get less noise when there is a bug (and Pharo been a big project, we have bugs for sure). But this is not good for the development of Pharo because we cannot know when an error happens.

The best solution for me would be to enable the debug mode only for released versions. But we don't have thi information currently in the image and this would require to update the bootstrap to add the info in SystemVersion. Here I'm going for the second best thing I could think of: let the errors pass during the tests so that we can at least catch what the tests are unveiling. 

I expect this PR to fail because Marcus found some log producing errors. But at least we'll be able to detect them better and fix them before integrating this.

When this will be mergeable, this will mean that no test will fail even with debug mode off. Fixes #13796


MarcusDenker (approved):

MarcusDenker:
We should turn on debug mode in general during development (in addition to have it always in during tests, which is done by this PR)

PR #13799, opened 2023-05-24
Debug CI

jecisc:
We have some PR failing with a strange bug. It seems that at some points, we have blocks that do not have the right layout.  In order to understand what is going on when things are working fine, here I just add some logs.


src/Shift-ClassInstaller/ShiftClassInstaller.class.st
```diff
@@ -76,14 +76,29 @@ ShiftClassInstaller >> copyClassSlotsFromExistingClass [
 ShiftClassInstaller >> copyObject: oldObject to: newClass [
 
 	| newObject |
+	newClass name = #CompliedBlock ifTrue: [
```
Gabriel-Darbord: Did you mean **Compiled**Block?

PR #13823, opened 2023-05-26
Change the `bench` method to return a `BenchmarkResult` instance

### LINKED ISSUE #12019: print ou of bench are not easily comparable
instead of 
```
"After: 
'18266371.200 per second'"
"Before: 
'10208252.099 per second'"
```
I would like to have 
```
"After: 
'18 266 371.200 per second'"
"Before: 
'10 208 252.099 per second'"
```

Gabriel-Darbord:
Plus change `BenchmarkResult>>#printOn:` to look better.
@MarcusDenker's idea! :)


svenvc:
LGTM

PR #13826, opened 2023-05-26
Move CompiledMethod>> basicNew: and basicNew to CompiledCode

MarcusDenker:
CompiledMethod has overrides for #basicNew and #basicNew:

But CompiledBlock does not override them. This is bad, as both can not be created with basicNew

This PR moves the method to CompiledCode


jecisc (approved):
Yes! :)

PR #13828, opened 2023-05-26
improved TempsReadBeforeWrittenRule

Enzo-Demeulenaere:
When temporaries are read before written, those temporaries will be displayed in the rule warning


Ducasse:
Nice addition. It is always frustrating to have a warning not helping us to fix it while it knows the  exact information

PR #13829, opened 2023-05-26
13797-Monticello-can-not-create-the-class-CompileCode-subclasses 

### LINKED ISSUE #13797: Monticello can not create the class CompileCode subclasses
**Bug description**

CompiledMethod and CompiledCode use both the CompiledMethodLayout. When this layout was added (maybe already in Apple ST80?), the language was not fixed to be able to create classes like that. Instead it uses #variableSubclass... and then hard-codes the name if the class (slighlty improved in Pharo to check for the layout, but that has it's problems, as it just supports updating classes with that layout, not creating new ones. But with Fluid enabled,  it works correctly)

Now Monticello hard codes just CompiledMethod by name, check

MCClassDefinition>>#initializeWithName:superclassName:traitComposition:classTraitComposition:category:instVarNames:classVarNames:poolDictionaryNames:classInstVarNames:type:comment:commentStamp:

```
name = #CompiledMethod
		ifTrue: [ type := #compiledMethod ]
		ifFalse: [ type := typeSymbol ].
```

We should at least add support for all classes here.

MarcusDenker:
- add all the names of the classes that have CompiledMethodLayout to the hard coded way so we are able to create them

fixes #13797

### DIFF
--- a/src/Monticello/MCClassDefinition.class.st
+++ b/src/Monticello/MCClassDefinition.class.st
@@ -274,9 +274,9 @@ MCClassDefinition >> initializeWithName: nameString superclassName: superclassSt
 	traitComposition := traitCompositionString.
 	classTraitComposition := classTraitCompositionString.
 	category := categoryString.
-	name = #CompiledMethod
-		ifTrue: [ type := #compiledMethod ]
-		ifFalse: [ type := typeSymbol ].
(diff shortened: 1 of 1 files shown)


jecisc (approved):

jecisc:
Failure seems unrelated

PR #13833, opened 2023-05-26
Color: little cleaning and optimization

Enzo-Demeulenaere:
little cleaning of Color

### DIFF
--- a/src/Colors/Color.class.st
+++ b/src/Colors/Color.class.st
@@ -113,7 +113,7 @@ Color class >> colorFrom: parm [
 	"(Color colorFrom: #(0.0 0.0 1.0)) >>> (Color blue)"
 
 	| aColor firstParm |
-	(parm isKindOf: self class)
+	(parm isKindOf: self)
 		ifTrue: [ ^ parm ].
 	parm isSymbol
 		ifTrue: [ ^ self perform: parm ].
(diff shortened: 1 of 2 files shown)


src/Colors/Color.class.st
```diff
@@ -113,13 +113,13 @@ Color class >> colorFrom: parm [
 	"(Color colorFrom: #(0.0 0.0 1.0)) >>> (Color blue)"
 
 	| aColor firstParm |
-	(parm isKindOf: Color)
```
guillep, line 116: All this smells like we need a double dispatch

src/Colors/Color.class.st
```diff
@@ -622,7 +621,7 @@ Color class >> initializeIndexedColors [
 				[ grayVal := v / 32.0.
 				a
 					at: index
-					put: (Color
+					put: (self class
```
jecisc:
```suggestion
					put: (self
```

We already are class side so this should be #self

src/Colors/Color.class.st
```diff
@@ -113,13 +113,13 @@ Color class >> colorFrom: parm [
 	"(Color colorFrom: #(0.0 0.0 1.0)) >>> (Color blue)"
 
 	| aColor firstParm |
-	(parm isKindOf: Color)
+	(parm isKindOf: self class)
```
jecisc:
```suggestion
	(parm isKindOf: self)
```

We are class side so this should be self

jecisc (changes requested):

jecisc:

jecisc (approved):

Issue #13806, opened 2023-05-24
Run coverage button is enabled with unselected packages ( Dr Tests )


CarolinaMelendez:
**Bug description**
In the test coverage plugin, when no package is selected and the run coverage button is pressed, it shows an exception "#collect: was sent to nil". Furthermore, it is not possible to go to Tests Runner, it shows the same exception.

**To Reproduce**
1. Go to DrTest
2. Select "Test coverage" plugin
3. Click on 'Run coverage' button
5. See exception 
6. Then select "Tests Runner" plugin
7. See exception

**Expected behavior**
The run coverage button shouldn't be enabled or a message "You have to choose one package" should be displayed.

**Version information:**
 - Pharo Version: 12.0

Ducasse:
Thanks carolina!
Good catch. Are you planning to work on it?

CarolinaMelendez:
Yes, I would like to solve it. 
As a practice too

MarcusDenker:
PR was merged

PR #13836, opened 2023-05-26
Run coverage button is enabled with unselected packages ( Dr Tests )

CarolinaMelendez:
#13806
This PR contains a modification to prevent the exception of "collect was sent to nil"


Ducasse:
Thanks for the contribution. It would be great to have a test covering this bug.

Ducasse:
failing test is unrelated

PR #13840, opened 2023-05-28
13769-Image-hangs-when-copying-a-text-from-Transcript

tesonep:
Configuring the NumberParser to use a max exponent (absolute value).
To avoid entering a end-less loop when parsing by error an string that looks like a number (e.g., a commitish 123e456)


Ducasse:
Thanks!

PR #13846, opened 2023-05-29
Fix bug in Epicea

jecisc:
This bug is hidden because of Epicea error handling. In order to disable this error handler we need to fix it

### DIFF
--- a/src/Epicea/EpMonitor.class.st
+++ b/src/Epicea/EpMonitor.class.st
@@ -233,7 +233,7 @@ EpMonitor >> behaviorRemoved: aClassRemovedAnnouncement [
 EpMonitor >> behaviorRemovedImpliesMethodRemoved: aMethodInAnObsoleteBehavior defaultPackageName: aSymbol [
 
 	| packageName |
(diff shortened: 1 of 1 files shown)


src/Epicea/EpMonitor.class.st
```diff
@@ -233,7 +233,8 @@ EpMonitor >> behaviorRemoved: aClassRemovedAnnouncement [
 EpMonitor >> behaviorRemovedImpliesMethodRemoved: aMethodInAnObsoleteBehavior defaultPackageName: aSymbol [
 
 	| packageName |
-	packageName := (Smalltalk organization packageForProtocol: aMethodInAnObsoleteBehavior protocol inClass: aMethodInAnObsoleteBehavior methodClass) name.
+	packageName := ((Smalltalk organization packageForProtocol: aMethodInAnObsoleteBehavior protocol inClass: aMethodInAnObsoleteBehavior methodClass) ifNil: [
```
Ducasse, line 236: should I mention that I hate such Smalltalk everywhere?

MarcusDenker (approved):

PR #13848, opened 2023-05-29
Fix System Browser scaling

Rinzwind:
This pull request fixes some methods in ‘Calypso-Browser’ that used width and height constants without taking the display scale factor into account.

Screenshot of a System Browser after these changes:

<p align="center">

</p>

The above screenshot was made after doing the snippets for ‘Retina display’ scaling from the comments in pull request #13767 (the message sent to SpFontStyle requires merging [Spec pull request 1397](https://github.com/pharo-spec/Spec/pull/1397)):

```smalltalk
WorldMorph displayScaleFactor: 2.
SourceCodeFonts setSourceCodeFonts: SourceCodeFonts sizeSmall * WorldMorph displayScaleFactor.
ToggleWithSymbolMenuItemShortcut
	classVarNamed: 'SymbolFont'
		put: (LogicalFont familyName: 'Lucida Grande' pointSize: 10 * WorldMorph displayScaleFactor);
	classVarNamed: 'SymbolTable' put: nil.
ThemeIcons current: ((ThemeIcons named: 'svgPack') in: [ :themeIcons |
	(themeIcons respondsTo: #loadIconsFromUrlUsingScale:) ifTrue: [
		themeIcons loadIconsFromUrlUsingScale: WorldMorph displayScaleFactor ] ];
	yourself).
World scaleFactor: WorldMorph displayScaleFactor reciprocal.
SpFontStyle applyDisplayScaleFactor: false.
```

PR #13850, opened 2023-05-29
Fix System Browser tables scaling

Rinzwind:
This pull request fixes methods related to the tables at the top of a System Browser that used width and height constants without taking the display scale factor into account.

Screenshot after these changes (compare with the one in pull request #13848):

<p align="center">

</p>


guillep:
Oh this looks great!!

I'll invoke @tesonep and @carolahp that know more about this display scale factor thingy :)

PR #13852, opened 2023-05-30
Skip testRatioOfCommentedClasses for now

MarcusDenker:
Skip testRatioOfCommentedClasses for now, it fails the CI on git for spec and newtools and it is not clear why.

it adds lots of noise to the CI checks, it is better to skip it for now I think.


jecisc (approved):
My guess is that since we load the project two times with 2 different versions, we might have some hanging classes with no comments still in the image or something like this

PR #13855, opened 2023-05-30
13591-Clean-Blocks-self-is-nil-on-Debugger 

### LINKED ISSUE #13591: Clean Blocks: self is nil on Debugger
If I stop a debugger inside a Clean Block and evaluate for `self` it is nil :(

image

MarcusDenker:
- #readInContext: reads the receiver of the home context
- change Context>>#home to fall back on #activeHome
- add two tests

fixes #13591


jecisc (approved):
The code seems good. I don't really know this part of the system so if someone else can review it might be good. But no obvious problem here :)

PR #13857, opened 2023-05-30
Fix bug in Epicea management

jecisc:
This fixes a bug found by disabling the error catching of Epicea.

I also renamed some variables to reduce the number of warnings in the system


MarcusDenker (approved):

MarcusDenker:
StDebuggerTest>>testUpdateExtensionSubscription is a random failure

PR #13858, opened 2023-05-30
Cut dependency of Collection-Strings over SessionManager

jecisc:
I would like to extract the session manager from the code been loaded by Espell in the Bootstrap. Since my PR cutting most of the dependencies of the kernel group is failing the CI, here is a smaller version taking care of the Collection-Strings package.


jecisc:
The CI seems happy but I cannot find the Symbol class in the SessionManager registration on the produced image which means that there is a problem here. 

This is probably the origin of why my other PR fails.

jecisc:
CI passed and the session manager in the generated image is good!

Ducasse:
the failing test  is not related.

MarcusDenker (approved):

PR #13866, opened 2023-05-31
Remove Epicea catch all

jecisc:
Epicea has a catch all mecanism but this catched also bugs of the system. With Marcus we would prefer the bugs to pass so that we can find and fix them instead of silenting them


MarcusDenker (approved):

MarcusDenker:
Yes, I think we should remove the Error handler.
- if Epicea is enabled, you want it to work, all the time
- if you do not need it, you can disable it
- for deployment, you want error handling with logging and that should be higher up
- this was hiding errors... we had tests not related to Epicea that should have failed but did not due to the error handler

PR #13868, opened 2023-06-01
Finish to remove #organization from protocol tests

jecisc:
As a step to remove ClassOrganization, this PR makes sur all protocols tests are dealing with classes and not organizations


jecisc:
Tests passed except windows that is doing nothing

MarcusDenker (approved):

PR #13872, opened 2023-06-01
Use new class deprecation when possible

### LINKED ISSUE #10840: Cleanup Deprecated classes
**Describe the problem**
There are quite a number of classes deprecated outside of the Deprecated packages.

```
(Smalltalk allClasses select: #isDeprecated) reject: [ :class | class package isDeprecated ]
```
54

As using them does not even print something in the log, we can not just remove them, as users never got notified.
(there might very well be users in the system even)

This should be improved in Pharo11.

jecisc:
This PR uses the new class deprecation system introduced in Pharo 12.

For now I've put all of them in the class initialization, but maybe it would be better to group all of those declarations in one place so that it is easier to remove later until we have a better deprecation mecanism?


MarcusDenker (approved):

MarcusDenker (approved):

Ducasse:
what means deprecateAliases:?
I have the impression that this method is not intention reveal.  Is it something like deprecateClassForwardingMessageTo:?

jecisc:
#deprecatedAliases: is to define some aliases for the class that are deprecated. 

In the long term, this should not be visible to the user. We want to update Tonel and the Fluid class builder to be able to define those in the class definition.

PR #13874, opened 2023-06-01
Just a tiny change to use the #compiler on Context in more cases

MarcusDenker:
... just some cleanups while reading code

### DIFF
--- a/src/OpalCompiler-Tests/OCClosureCompilerTest.class.st
+++ b/src/OpalCompiler-Tests/OCClosureCompilerTest.class.st
@@ -327,9 +327,7 @@ OCClosureCompilerTest >> doTestDebuggerTempAccessWith: one with: two [
 
 { #category : #running }
 OCClosureCompilerTest >> evaluate: aString in: aContext to: anObject [
-	^self class compiler source: aString;
-		context: aContext;
-		evaluate
+	^ aContext compiler evaluate: aString
 ]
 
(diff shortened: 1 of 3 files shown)


src/OpalCompiler-Tests/OCClosureCompilerTest.class.st
```diff
@@ -327,9 +327,7 @@ OCClosureCompilerTest >> doTestDebuggerTempAccessWith: one with: two [
 
 { #category : #running }
 OCClosureCompilerTest >> evaluate: aString in: aContext to: anObject [
-	^self class compiler source: aString;
-		context: aContext;
-		evaluate
+	^ aContext compiler evaluate: aString
```
jecisc, line 330: For a next step: The third argument can be removed and maybe this whole method can be inlined.

jecisc (approved):

src/OpalCompiler-Tests/OCClosureCompilerTest.class.st
MarcusDenker: yes, lots of next steps needed. For tests in addition I have to think again about the possibility to have two compilers... if the tests use the method from context or "Smalltalk compiler", this is hard to run the tests not using the default compiler (so maybe in the tests we have to use OpalCompiler directly, so we can replace the reference)... this is why I did not continue in the tests for now

PR #13877, opened 2023-06-01
ThisProcess: try to implement reflective read better

MarcusDenker:
This PR tries to implement reading of thisProcess... it searches all Processes for the context. if it does not find it, it retunes the current process (using thisProcess).

if you now print or inspect "thisProcess" in the debugger, it now returns the process that you expect. #testReadInContext shows that you can read reflectively from the current thisContext, too.

But I am not sure about the fallback, yet it seems to work

### DIFF
--- a/src/Kernel/ProcessorScheduler.class.st
+++ b/src/Kernel/ProcessorScheduler.class.st
@@ -142,6 +142,14 @@ ProcessorScheduler >> backgroundProcess [
 	^ BackgroundProcess
 ]
 
+{ #category : #querying }
+ProcessorScheduler >> findProcessForContext: aContext [
+	 
+	^ Process allInstances 
+		detect: [ : process | process hasContext: aContext ]
+		ifNone: [  thisProcess ]
+]
+
 { #category : #'priority names' }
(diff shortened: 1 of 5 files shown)


jecisc (approved):

src/Kernel/ProcessorScheduler.class.st
```diff
@@ -142,6 +142,14 @@ ProcessorScheduler >> backgroundProcess [
 	^ BackgroundProcess
 ]
 
+{ #category : #querying }
+ProcessorScheduler >> findProcessForContext: aContext [
+	 
+	^ Process allInstances 
+		detect: [ : process | process hasContext: aContext ]
+		ifNone: [  thisProcess ]
```
privat, line 150: retuning `thisContext` as a default seems a bad idea. Retuinig nil or raising an error (or calling a block) seems a safer approach.

But in which condition a context can exist without an owing process? Independent instantiation of a Context object (`Context basicNew findProcess`)? Escaped context reference on dead process?

Maybe add some tests on these cases

PR #13881, opened 2023-06-02
Introduce CompiledMethod>>protocolName and start to use it

jecisc:
In order to make CompiledMethod>>protocol return a real protocol, it would be nice to have CompiledMethod>>protocolName to ease with the transition. 

This PR adds this method and start to replace the users. There are still multiple users that needs to be updated but that protocol managements is still inconsistent and I'm afraid that doing too many changes in the PR will make it harder to find potential troubles.


MarcusDenker:
Failing test: EpCodeChangeIntegrationTest>>testMethodModificationOfProtocol

jecisc:
Test failure is from Spec and is now fixed

MarcusDenker (approved):

PR #13882, opened 2023-06-02
Some simple cleanups (lookupVar: category, #home use ifNotNil...)

MarcusDenker:
- move some #lookupVar: to the compiler package from debugging (a bit better place), some to "lookup" where the compiler is not used (e.g. Behavior).
- trivial clean nextReadyProcess
- trivial cleanup to Context>>#home from the mail


MarcusDenker:
Just already fixed test failing: https://github.com/pharo-spec/Spec/pull/1406

jecisc (approved):

PR #13886, opened 2023-06-05
Fix logging of protocol in Opal

jecisc:
When we compile a method we save its sources in the source file. This log includes the protocol. 

Currently, it logs the protocol given to Opal and logs before we classify it. But this is not always the final protocol. In some cases, the classification does not use what is given as parameter.

I propose to classify the method before logging it and to ask the protocol directly to the method instead of using the one given as parameter.


jecisc:
What is breaking the tests is not the change of logged protocol but the fact that with this PR we add the method to the class before printing it in the source file. But I don't know why yet.

jecisc:
9 tests are fixed. There is still 2 failures but they'll be harder to find because my tests pass locally

jecisc:
The failure is really weird. From the image produced by the CI, MCWorkingCopy announcements are not registered. But the registration is done in the bootstrap and the code I changed does not have any relation with the code registering those announcements...

jecisc:
It is now green!

MarcusDenker (approved):

PR #13897, opened 2023-06-05
remove-isReservedVariable-check-from-ReCollectionMessagesToExternalObjectRule

MarcusDenker:
This PR removes the only use of #isReservedVariable outside of the compiler

Another step towards making these variables less special


jecisc:
The error is a new one but I don't think it is related.

jecisc (approved):

PR #13900, opened 2023-06-05
Split Finder package between model and UI

jecisc:
Since the Finder will need to be rewritten, here is a splitting of the Tool-Finder package to extract the UI in Tools-Finder-UI


Ducasse:
unrelated failed test

PR #13918, opened 2023-06-06
Enable-Fluid-Skip-Tests

MarcusDenker:
- enable fluid by default
- skip two tests in ChangeSetClassChangesTest which are related to the Category vs. Package confusion in the test setup code  I changed #13315 to track this


jecisc (approved):

Ducasse:
I removed many subclass:.... so I will continue to clean such users.

PR #13919, opened 2023-06-06
Reduce references to Smalltalk

jecisc:
To access the RPackageOrganizer of the system we had a lot of ways in the past. I removed multiple of them to replace them by `Smalltalk organization`. But referencing globals is not that good. It's hard to find the best solution, but there is already the concept of #environment in the system and the environment knows the default package organizer. So we can just go through this environment to get it, thus reducing the references to a global. 

Under the hood it will still kinda access a global, but when the environment will be managed well, the package organizer will automatically work well too with this change.


jecisc:
@Ducasse I think you will like this one

MarcusDenker (approved):

Ducasse:
Sooooo nice :)

Ducasse:
This is so cool to use correct abstractions instead of globals

PR #13921, opened 2023-06-06
Remove some references to `RPackage class>>organizer`

jecisc:
We have a tons of way to access the default RPackageOrganizer of the image and I removed at least 3 ways already.  In this process of unification I found yet another way to access it, via `RPackage class>> organizer`. 

This is a first step to unify the access to the default RPackageOrganizer for senders of #organizer. 

This is only a part of it because multiple senders are in external projects and there are more in Pharo but I didn't wanted to do a PR that would be a beast.


jecisc:
This is not so simple. I'll close this one for now and come back to it later with a better way

PR #13922, opened 2023-06-06
Compiler-Reserved-Vars-are-just-readOnly

MarcusDenker:
This PR simplifies what happens when you try to assign to a pseudvariable.

Right now, we get a special error message 'Assigment to reserved variable'. We need to rename the term anyway (they should not be "reserved", and in documentation we use PseudoVariable).

While starting to look at that, I wonderd: does it really add anything to have the special error notice? There is already 'Assignment to read-only variable', and these variables are just one case of those.

This PR thus simplifies to remove the special check, #analyzeWrite:by: now just checks for #isWritable


MarcusDenker:
after thinking about it more... this is for later. It feels like the compiler should only care about read-only, but having nice explanations for errors is good. Maybe this can be moved into it's own visitor later... so for now lets keep it

PR #13923, opened 2023-06-06
Add benchmark comparison

Gabriel-Darbord:
A BenchmarkComparison class has been added in the same spirit as BenchmarkResult.
An instance prints the frequency of both results and the factor between them:
```st
[ (Date current year / 100) asInteger * 100 ] benchCompareTo: [
  (Date current year // 100) * 100 ]. "694997.401/s * 1.203 = 836151.770/s"
```

Plus some updates to the Kernel-Chronology-Extras package.

### DIFF
--- a/src/Kernel-Chronology-Extras/BenchmarkComparison.class.st
+++ b/src/Kernel-Chronology-Extras/BenchmarkComparison.class.st
@@ -1,0 +1,60 @@
+Class {
+	#name : #BenchmarkComparison,
+	#superclass : #Object,
+	#instVars : [
+		'referenceResult',
+		'comparisonResult'
+	],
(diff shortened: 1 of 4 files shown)


Ducasse:
Hello gabriel this is nice. 
BTW do you know SMark because this is a nice little framework (that could be improved) to run benchmarks

Gabriel-Darbord:
I didn't know about SMark, sounds interesting.
With @BastouP411 we started developing this comparison based on a discussion with Marcus and #12019.
It's convenient to have a quick comparison tool without having to install a dependency.

Ducasse:
Sure. This is a good idea. 
We also want to extend the parser to support better printing of large numbers.

MarcusDenker (approved):

PR #13933, opened 2023-06-07
Cleanup: methods duplicated in the hierarchy / tag selectors

MarcusDenker:
- remove some methods duplicated in the hierarchy 
- tag some tests that use selectors that they compile
- remove #traitOrClassOfSelector (dead code)

### DIFF
--- a/src/TraitsV2-Tests/TraitPureBehaviorTest.class.st
+++ b/src/TraitsV2-Tests/TraitPureBehaviorTest.class.st
@@ -190,7 +190,7 @@ TraitPureBehaviorTest >> testOwnMethodsTakePrecedenceOverTraitsMethods [
 	self assert: (trait methodDict at: #m12) sourceCode equals: 'm12 ^12'
 ]
 
-{ #category : #'as yet unclassified' }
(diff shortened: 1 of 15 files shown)


src/TraitsV2-Tests/TraitPureBehaviorTest.class.st
```diff
@@ -190,8 +190,9 @@ TraitPureBehaviorTest >> testOwnMethodsTakePrecedenceOverTraitsMethods [
 	self assert: (trait methodDict at: #m12) sourceCode equals: 'm12 ^12'
 ]
 
-{ #category : #'tests - applying trait composition' }
+{ #category : #'as yet unclassified' }
```
jecisc: I guess this is not intended

jecisc (changes requested):
One protocol seems to have been removed

jecisc (approved):

src/TraitsV2-Tests/TraitPureBehaviorTest.class.st
MarcusDenker: fixed

PR #13940, opened 2023-06-07
Clleanup: VariableNotDeclaredTest and some #ignoreNotImplementedSelectors:

MarcusDenker:
VariableNotDeclared has been deprecated, this removes the test VariableNotDeclaredTest and the subclass MyVariableNotDeclaredDummyClassForTest

in addition some trivial cleanups:

- add ignoreNotImplementedSelectors: pragma to more tests
- remove compatibility method named:trait: which was sending a non existing selector
- remove shortcut that is sending non-existing #cancelWithoutConfirmation in TextEditor (class is mor or less dead code anyway)


jecisc (approved):

src/TraitsV2/Trait.class.st
```diff
@@ -73,11 +73,6 @@ Trait class >> named: aSymbol package: aString [
 		env: self environment
 ]
 
-{ #category : #'old pharo 8 syntax' }
-Trait class >> named: aSymbol trait: aCompositionOrArray [
-	^ self named: aSymbol trait: aCompositionOrArray package: 'Unclassified'
-]
```
jecisc, line 79: Maybe this one should be deprecated instead of removed

MarcusDenker: the idea is that methods that un-conditionally send non existing selectors are dead code. By definintion, this can not have users

PR #13943, opened 2023-06-08
Use thisProcess in some tests

MarcusDenker:
This PR changes some methods to use thisProcess instead of "Processor activeProcess"

- Just in tests for now
- after this, we have 80 usages of activeProcess remaining


MarcusDenker:
conflicts... and I have no idea how to remove that change, as it is a subclass of a deprecated class. easier to close the PR and redo later

jecisc:
An alternative is to revert the change but commit only the reversion of the conflicting file. It’s usually what I am doing

PR #13947, opened 2023-06-08
Remove ReUtilityMethodsRule

### LINKED ISSUE #12997: Remove ReUtilityMethodsRule
Now Iceberg propose us to check the rules on commit and this is great!
But we have some rules that do not make sens or that have a looooot of false positive. 

Pharo 12 might be a good opportunity to clean that. If we give better critics, then people might have more motivation to fix the critics remaining. 

In that direction I propose to remove ReUtilityMethodsRule because because it is not a bad thing to have utility methods in Pharo! Without them our methods would be much longer and less readable.

jecisc:
Utilisty methods are not to be overused, but they are necessary to keep our code clean. For example, why is it bad to have an utility method to factorize code?

Since this rule give a false positive most of the time I propose to remove it.

Fixes #12997


MarcusDenker:
I vote for removed. But I tried already twice over the years and I was always told that the rule it good and should stay.

I *never* *ever* acted on this rule. That is, I never changed code (that is, inlined a method into all the senders) due to the rule telling me that a method is a utility method.

MarcusDenker:
Another idea: keep the rule for the CritiqueBrowser for now, but disable it in the code browser

jecisc:
For some rules I think they can have a lot of false positive but still be pertinent. In that case I would like to be able to not have them all the time enabled, but we can keep the rule.

For this rule in particular, I never acted on it because it has never been something else than a false positive, in the way that the code would have been bad if I inlined the code (less readable, duplicated, ...)

MarcusDenker (approved):

guillep:
Thanks @jecisc and reviewers!

PR #13951, opened 2023-06-08
CompiledMethod>>protocol should return a protocol

jecisc:
CompiledMethod>>protocol should return a protocol instead of a protocol name. If a user needs a protocol name, he can use #protocolName instead.

The goal of this PR is to make the system work with this change and then we can iterate to simplify some code when it will be easier to use a real protocol instead of a protocol name.

Depends on: https://github.com/pharo-project/pharo/pull/13967


jecisc:
There is only one failure left but this one is in Roassal. This is fixed upstream but the fix is not in Pharo yet because roassal works only with releases and not with an integration branch.

MarcusDenker (approved):

PR #13952, opened 2023-06-08
Add #enumerationsDo: and #enumerationsOfSize:

hernanmd:
Add #enumerationsDo: and #enumerationsOfSize: methods to SequenceableCollection. These methods generate all the possible combinations of N numbers including repetitions.
Add test

Example:
```smalltalk
 

(0 to: 1) enumerationsDo: [ :each | Transcript show: each printString ; cr ]

#(0 0)
#(0 1)
#(1 0)
#(1 1)
```

### DIFF
--- a/src/Collections-Abstract-Tests/TEnumeratingTest.trait.st
+++ b/src/Collections-Abstract-Tests/TEnumeratingTest.trait.st
@@ -84,3 +84,27 @@ TEnumeratingTest >> testFlattened [
 	"we do not try to flatten CompiledMethods"
 	self assert: { Object>>#halt } flattened equals:  { Object>>#halt }
 ]
+
+{ #category : #'tests - enumerating' }
+TEnumeratingTest >> testProductWithRepetitions [
+
+   | collection result |
+   collection := #(1 2 3).
+   result := collection productsOfSize: 2.
+   self assert: result size equals: 9.
+   self assert: (result includesAll: #(#(1 1) #(1 2) #(1 3) #(2 1) #(2 2) #(2 3) #(3 1) #(3 2) #(3 3))).
+]
+
+{ #category : #'tests - enumerating' }
+TEnumeratingTest >> testProductWithRepetitionsIteration [
+
+   | collection index elements |
+   collection := #(1 2 3).
+	index := 1.
+	elements := #(#(1 1 1) #(1 1 2) #(1 1 3) #(1 2 1) #(1 2 2) #(1 2 3) #(1 3 1) #(1 3 2) #(1 3 3) #(2 1 1) #(2 1 2) #(2 1 3) #(2 2 1) #(2 2 2) #(2 2 3) #(2 3 1) #(2 3 2) #(2 3 3) #(3 1 1) #(3 1 2) #(3 1 3) #(3 2 1) #(3 2 2) #(3 2 3) #(3 3 1) #(3 3 2) #(3 3 3)).
+   collection productsDo: [ : product |
+	   self 
+			assert: product
+			equals: (elements at: index).
+		index := index + 1 ]
+]
--- a/src/Collections-Abstract-Tests/TEnumeratingTest.trait.st
+++ b/src/Collections-Abstract-Tests/TEnumeratingTest.trait.st
@@ -28,6 +28,44 @@ TEnumeratingTest >> simpleCollection [
 ]
 
 { #category : #'tests - enumerating' }
+TEnumeratingTest >> testCartesianProductDo [
+
+   | collection index elements |
+
+   collection := #(1 2).
+	index := 1.
+	elements := #(#(1 1) #(1 2) #(2 1) #(2 2)).
+   collection permutationsWithRepetitionsDo: [ : product |
+	   self 
+			assertCollection: product hasSameElements: (elements at: index).
+			index := index + 1 ]
+]
+
+{ #category : #'tests - enumerating' }
+TEnumeratingTest >> testCartesianProductDo3Elements [
+
+   | collection index elements |
+   collection := #(1 2 3).
+	index := 1.
+	elements := #(#(1 1 1) #(1 1 2) #(1 1 3) #(1 2 1) #(1 2 2) #(1 2 3) #(1 3 1) #(1 3 2) #(1 3 3) #(2 1 1) #(2 1 2) #(2 1 3) #(2 2 1) #(2 2 2) #(2 2 3) #(2 3 1) #(2 3 2) #(2 3 3) #(3 1 1) #(3 1 2) #(3 1 3) #(3 2 1) #(3 2 2) #(3 2 3) #(3 3 1) #(3 3 2) #(3 3 3)).
+   collection permutationsWithRepetitionsDo: [ : product |
+	   self 
+			assert: product
+			equals: (elements at: index).
+		index := index + 1 ]
+]
+
+{ #category : #'tests - enumerating' }
+TEnumeratingTest >> testCartesianProductOfSize [
+
+   | collection result |
+   collection := #(1 2 3).
+   result := collection permutationsWithRepetitionsOfSize: 2.
+   self assert: result size equals: 9.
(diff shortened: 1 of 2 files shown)


privat:
name `enumeration` feels very bad. What you are doing seems to be an n-ary Cartesian power (E x E x ... x E)

hernanmd:
> name `enumeration` feels very bad. What you are doing seems to be an n-ary Cartesian power (E x E x ... x E)

A n-ary cartesian power of a set is a collection of all ordered n-tuples of elements from the set, where repetition of elements is allowed. So although this is correct, it feels like a pedantic way of naming a method. For example: a method named cartesianProductWithRepetitions or alike would be cryptic for those who do not remember about cartesian products, and does not follow similar names to methods with similar functionality like #combinationsDo: and #permutationsDo:

guillep:
Please, let's choice another tone for the discussion.

I personally agree with @privat : `#enumerationsXXX` does not tell me what it does at all.
I also agree with @hernanmd in that `cartesianProductWithRepetitions` is probably a bit too verbose.

Now, I want methods to tell me what they do, have tests that show it and have nice comments explaining what I should expect from them. To me an "enumeration" is just an ordered list.

Maybe we can choose a better name that satisfies both?

guillep:
Also, the PR should not only add a new method and a test, I believe it should say why we need this method in the standard library, what's the usage, why this is not an extension in a separate package, and so on.

hernanmd:
> Also, the PR should not only add a new method and a test, I believe it should say why we need this method in the standard library, what's the usage, why this is not an extension in a separate package, and so on.

It is useful  to count the number of possible combinations of n elements from a set. For example, if we have a set of size k, then the number of possible k-tuples with repetition is k^n, since each element can be chosen n times. So it gives the total number of possible outcomes for n independent events, for example, bytecode combinations of a method.

It can also be used to generate test cases. For example, if we have a method that takes a Collection as argument, we can generate test cases by selecting n-tuples of input values from the Cartesian product of the input domains. This ensures that we test all possible combinations of input values, which is important for detecting bugs and ensuring correctness.

This has other usage scenarios: In cryptography for example the security of some encryption schemes relies on the difficulty of finding the number of possible combinations of a large set of elements. Also in probability theory, where the number of possible combinations of a set of events is used to calculate the probability of certain outcomes, think about drawing a flush in a standard deck of N cards which can be calculated by dividing the number of possible flush combinations by the total number of possible combinations of K types of cards.

Also most programming languages provide built-in functions or modules for generating the n-ary Cartesian power of a set. It is included by default in [Python's itertools (product)](https://docs.python.org/3/library/itertools.html#itertools.product), in C++ [std::cartesian_product()](https://en.cppreference.com/w/cpp/ranges/cartesian_product_view), in Rust [itertools as iproduct()](https://docs.rs/itertools/latest/itertools/macro.iproduct.html)

I agree the name could be changed to something more familiar, although I didn't find any better name so far.

jecisc:
osx-64 / Tests-osx-64 / testEnumerationsOfSizeWithRepetitions – MacOSX64.Collections.Sequenceable.Tests.Base.ArrayTest<1s

```
Error
Message not understood: Array >> #enumerateWithSize:subSize:in:do:
Stacktrace
MessageNotUnderstood
Message not understood: Array >> #enumerateWithSize:subSize:in:do:
Array(Object)>>doesNotUnderstand: #enumerateWithSize:subSize:in:do:
Array(SequenceableCollection)>>enumerationsOfSize:do:
Array(SequenceableCollection)>>enumerationsOfSize:
ArrayTest>>testEnumerationsOfSizeWithRepetitions
ArrayTest(TestCase)>>performTest
```

guillep:
I like better the new name! It seems a good trade-off between conciseness and intention reveal.

I see there are a couple of merge conflicts after the merge of the Pharo12 branch, It's strange

Ducasse:
@hernanmd can you handle it so that we integrate it?
Tx

Ducasse:
Why did you close it? was it not a good addition?

hernanmd:
Mmm that was strange. I just synchronized changes through the Github web interface and add the commits manually, never actually clicked close!

I add one more test also here in this PR

Ducasse:
Cool that they are not lost.

jordanmontt:
Hello, I come late to the issue.

Normally the Cartesian product is for different sets and **does not** allow repetitions and it returns a list of ordered pairs (If you do the Cartesian product of 3 sets you will get a list of ordered pairs of 3 elements). See (https://en.wikipedia.org/wiki/Cartesian_product or the linked libraries for other languages https://docs.python.org/3/library/itertools.html#itertools.product).

For example look in Python `product('ABCD', 'xy') --> Ax Ay Bx By Cx Cy Dx Dy`. 

As @privat pointed out, this is a n-ary Cartesian product **but** with the same set n-times.

In the proposed example the result is the cartesian product of the set `#(0 1)` with itself.

```st
(0 to: 1) enumerationsDo: [ :each | Transcript show: each printString ; cr ]
#(0 0)
#(0 1)
#(1 0)
#(1 1)
```

If you try this example but with a set of 3 elements `#(0 1)` as a result you will set a collection of 27 elements or 3-elements ordered pairs. So, the same as doing the triple Cartesian product of `#(0 1)` with itself. This feels wrong IMO.

So, for me we need to rename the method and to propose a better API that allows to do the actual Cartesian product of different sets. Like in Python e.g.
- `product('ABCD', 'xy') --> Ax Ay Bx By Cx Cy Dx Dy`.
- `product('123', '456', '789') --> 147 148 149 157 158 159 167 168 169 247 248 249 257 258 259 267 268 269 347 348 349 357 358 359 367 368 369`

If the user wants repetitions he/she can use the Cartesian product with the same set. Like in `product('01', '01') --> 00 01 10 11`.

If the name "Cartesian product" seems too "fancy" or whatever adding a small comment solves the problem :) The name `products` does not tell me anything and it's actually confusing as discussed previously, it is a product yes but of the same set n-times...

Anyways, most of the people that want that they probably have heard of cartesian product (in the linked C example the function is called cartesian product)

<blockquote><div><strong><a href="https://en.wikipedia.org/wiki/Cartesian_product">Cartesian product - Wikipedia</a></strong></div></blockquote>
<blockquote><div>Python documentation</div><div><strong><a href="https://docs.python.org/3/library/itertools.html">itertools — Functions creating iterators for efficient looping</a></strong></div><div>This module implements a number of iterator building blocks inspired by constructs from APL, Haskell, and SML. Each has been recast in a form suitable for Python. The module standardizes a core set...</div></blockquote>

hernanmd:
Why the re-open @jecisc ?

guillep:
Hi @jordanmontt @hernanmd how can we get to an agreement and move forward? :)
I think the addition is nice, let's flush the little details and take a decision ;)

jecisc:
> Why the re-open @jecisc ?

I just closed and reopen to restart the CI

jordanmontt:
Hi @guillep,

If you ask me, we need to work on two points for integrating this PR:

- I will convert the method to `cartesianProduct:` to have more flexibility. Like this we can have the Cartesian product of different sets too., not only for one that is repeated n-time. For example, `'ABCD' cartesianProduct: 'xy'`.
- The implementation is good. I see that is based on the implementation of the `combinations` method (`combinations:atATimeDo:`). The problem is that is you don't send the message copy, you will have wrong results. This is a problem also in the combinations method. Look this two following screenshots:

Without sending the message copy we get the same number (wrong result).
Capture d’écran 2023-08-03 à 12 32 03

Copy fixes the problem.
Capture d’écran 2023-08-03 à 12 32 29

jordanmontt:
Thanks @hernanmd for applying the suggestions!
At the end, we decided to have the name permutations with repetitions. Because what Hernan actually wanted is that and not the Cartesian product. For me the Pr is ready to be merged

jordanmontt (approved):

jecisc:
Let's wait for the CI result and then we can merge :)

PR #13953, opened 2023-06-08
Make initial extent of SettingBrowser, ThreadSafeTranscript, FinderUI, EpUnifiedBrowserPresenter and EpLogBrowserPresenter take the display scale factor into account

Rinzwind:
This pull request makes the initial extent of SettingBrowser, ThreadSafeTranscript, FinderUI, EpUnifiedBrowserPresenter and EpLogBrowserPresenter take the display scale factor into account.


guillep:
Nice!!! As always, maybe @carolahp can review this scale factor issue? :)

PR #13959, opened 2023-06-09
13957--RBRefactoringonErrordo--should-accept-both-0-and-1-arg-block

### LINKED ISSUE #13957: RBRefactoring>>#onError:do:  should accept both 0 and 1 arg block
Capture 2023-06-09 at 07 33 07

MarcusDenker:
RBRefactoring>>#onError:do:  is called with both 0 and 1 arg error bl…ocks. a 1-arg block feels natural, as on:do: takes an optional parameter, too.

This PR therefore changes #onError:do:  to work with both 1 and 0-arg error blocks

(the whole idea of #onError:do: is strange.. do we really want to hide why a refactoring can not be done?)

fixes #13957


Ducasse:
I did it too.

PR #13961, opened 2023-06-09
13942-TraitCompositionTesttestAliasCompositions--passes-but-for-the-wrong-reason

### LINKED ISSUE #13942: TraitCompositionTest>>#testAliasCompositions  passes, but for the wrong reason
**Bug description**
TraitCompositionTest>>#testAliasCompositions  has some checks for raising Error. 

```
self
		should: [
			self t2
				setTraitCompositionFrom:
					self t1
						@
							{(#alias -> #m11).
							(#alias2 -> #alias)} ]
		raise: Error
```

it raises an error because #setTraitCompositionFrom: is a DNU

MarcusDenker:
The system does allow one of the cases, it will just use the last alias

fixes #13942


jecisc (approved):
Nice catch!

PR #13964, opened 2023-06-09
compilationContext is lazily initialise so the acessor should be used

Inao0:
In OpalCompiler, the compilation context is initialized lazily.
```
compilationContext
	^ compilationContext ifNil: [ compilationContext := self compilationContextClass default ]
```

Therefore all access to the compilationContext should use the accessor. 
In particular, `OpalCompiler >> addPlugin:` was not using the accessor, which raised an exception about a message not understood being sent to nil.


jecisc (approved):

guillep:
Thanks @Inao0 :)

PR #13965, opened 2023-06-09
Modified the UUID class to use a primitive and added parsing options

Mathilde411:
The UUID class has been modified to use the primitive from pharo-project/pharo-vm#610.

Parsing methods were also added mainly for version 1 RFC 4122 UUIDs.

See: https://datatracker.ietf.org/doc/html/rfc4122


src/Network-UUID/UUID.class.st
```diff
@@ -142,36 +135,119 @@ UUID >> asString36 [
...
+	
+	"Should be deprecated, but is still used by iceberg."
+	"self
+		deprecated:
+		'The UUIDs are no longer a subclass of ByteArray, and should not be directly modified.'."
+	^ uuidData at: aNumber
+]
+
+{ #category : #accessing }
+UUID >> at: aNumber put: aByte [
```
Mathilde411, line 179: This Method should be deprecated, but is currently used by Iceberg. Will change it after Iceberg is updated.

jecisc:
The bootstrap is failing with:

```
MetacelloNotification: Loading -> Iceberg-CompatibleUserName.1686126007 --- https://github.com/pharo-vcs/Iceberg.git[dev-2.0] --- cache
Error: You are about to load new versions of the following packages
that have unsaved changes in the image:
  Iceberg

If you continue, you will lose these changes:
MCVersionLoader>>warnAboutLosingChangesTo:ifCancel:ifMerge:
MCVersionLoader>>checkForModificationsIfCancel:ifMerge:
MCVersionLoader>>basicLoadWithNameLike:
```

This error is a new one to me on the CI

src/Network-Tests/UUIDGeneratorTest.class.st
```diff
@@ -18,67 +18,75 @@ UUIDGeneratorTest >> setUp [
...
+	self assert: UUIDGenerator default equals: UUIDGenerator default.
 ]
 
 { #category : #testing }
 UUIDGeneratorTest >> testSpeed [
+	|array|
+	array := ByteArray new: 16.
 	self
-		should: [ 100000 timesRepeat: [ generator next ] ]
+		should: [ 100000 timesRepeat: [ UUIDGenerator default placeFields: array ] ]
```
guillep: I don't like this test.

First, it's breaking the public API. Why testing `placeFields:` and not directly `UUID new`.
Then, testing speed like this is just a smoke test. Why 100000 and not more (or less?) :D.
I'd remove this test altogether.

src/Fuel-Tests-Core/FLBasicSerializationTest.class.st
```diff
@@ -796,7 +796,7 @@ FLBasicSerializationTest >> testTrue [
 FLBasicSerializationTest >> testUUID [
 
 	self assertSerializationEqualityOf: (UUID fromString: 'a3b64357-377a-5b41-b575-1c653084a121').
-	self assertSerializationEqualityOf: UUID nilUUID.
+	self assertSerializationEqualityOf: NilUUID new.
```
guillep: What is the rationale behind NilUUID ?
Why a different class?

If having a nil uuid is really important, why couldn't we have `UUID nilUUID`? That way the implementation is kinda hidden (one class, two classes, the user should not care, it's easier to refactor)

src/Network-Tests/UUIDGeneratorTest.class.st
```diff
@@ -18,67 +18,75 @@ UUIDGeneratorTest >> setUp [
...
+
 	| generator1 generator2 uuid1 uuid2 time1 time2 |
 	generator1 := UUIDGenerator new.
 	generator2 := UUIDGenerator new.
-	uuid1 := generator1 next.
-	uuid2 := generator2 next.
+	uuid1 := NilUUID new.
+	uuid2 := NilUUID new.
+	generator1 placeFields: uuid1 uuidData.
+	generator2 placeFields: uuid2 uuidData.
```
guillep: This test breaks the public interface. We should kill it probably or rewrite it :)

Let's discuss

src/Network-UUID/ByteArray.extension.st
```diff
@@ -0,0 +1,8 @@
+Extension { #name : #ByteArray }
+
+{ #category : #'*Network-UUID' }
+ByteArray >> generateUUIDInPlace [
+	<primitive: 'primitiveMakeUUID' module: 'UUIDPlugin'>
+	self size = 16 ifFalse: [ self error: 'The byte array size has to be 16 bytes.' ].
+	UUIDGenerator default placeFields: self
```
guillep, line 7: Nice!!

src/Network-UUID/NilUUID.class.st
```diff
@@ -0,0 +1,13 @@
+"
+UUID with all fields set to 0.
```
guillep: My point: We don't need a separate class for this :) Just a different instance is enough!

src/Network-UUID/UUID.class.st
```diff
@@ -1,6 +1,5 @@
 "
 I am UUID.
-I am a ByteArray.
```
guillep, line 3: Cuack :D

src/Network-UUID/UUID.class.st
```diff
@@ -187,13 +273,69 @@ UUID >> printString [
...
+	"Returns the variant of the UUID. See https://tools.ietf.org/html/rfc4122"
+
+	^ (uuidData at: 9) >> 5
+]
+
+{ #category : #comparing }
+UUID >> version [
+	"Returns the version of the UUID. See https://tools.ietf.org/html/rfc4122"
+
+	^ (uuidData at: 7) >> 4
```
guillep, line 356: It's nice to have a real UUID implementation!

guillep (changes requested):
I left some comments.

I think we should make Iceberg not need to look at the internals, and we should discuss the API :)

src/Network-UUID/UUID.class.st
Mathilde411: 🦆

src/Network-UUID/NilUUID.class.st
Mathilde411: NilUUID has been removed :)

src/Network-Tests/UUIDGeneratorTest.class.st
Mathilde411: NilUUID has been removed :)

src/Network-Tests/UUIDGeneratorTest.class.st
```diff
@@ -18,67 +18,70 @@ UUIDGeneratorTest >> setUp [
...
 	"different generators, different node id"
-	self deny: uuid1 reverse readStream uint32 equals: uuid2 reverse readStream uint32.
+	self
+		deny: uuid1 uuidData reverse readStream uint32
+		equals: uuid2 uuidData reverse readStream uint32.
 	"same generators, similar clock, within 100 milliseconds (stored time is in microseconds)"
 	time1 := self timeFromUUID: uuid1.
 	time2 := self timeFromUUID: uuid2.
-	self assert: (time2 - time1) abs < (100 * 1000)
+	"self assert: (time2 - time1) abs < (100 * 1000)"
```
guillep, line 65: Should we remove the commented code?

guillep (approved):
I think this is ok, modulo one line of code that is commented.

Question: are you planning to change Iceberg to not access the UUID directly?

Mathilde411:
> I think this is ok, modulo one line of code that is commented.
> 
> Question: are you planning to change Iceberg to not access the UUID directly?

I was waiting for the changes to be merged since it is not in the main repo, but yes.

src/Network-Tests/UUIDGeneratorTest.class.st
guillep, line 65:
```suggestion
```

src/Network-UUID/UUID.class.st
```diff
@@ -187,13 +288,69 @@ UUID >> printString [
...
+
+{ #category : #accessing }
+UUID >> timestamp [
+	"Returns the timestamp of the version 1 UUID as a local DateAndTime. See https://tools.ietf.org/html/rfc4122"
+
+	| timestamp |
+	self isTimeBased ifFalse: [
+		self error: 'Only time-based UUIDs have a timestamp.' ].
+
+	timestamp := (uuidData at: 7) & 16r0f.
```
guillep, line 320:
```suggestion
	timestamp := (uuidData at: 7) & 16r0f.
	"The timestamp bytes in the UUID bytearray are scrambled. Iterate them in the good order"
```

src/Network-Tests/UUIDGeneratorTest.class.st
```diff
@@ -18,67 +18,70 @@ UUIDGeneratorTest >> setUp [
...
+	"UUID are not necessarly of version 1->3, time can be random"
```
guillep:
```suggestion
```

guillep:
CI is green! changes are good! Thanks @BastouP411

PR #13967, opened 2023-06-09
Update the way we deal with protocols from other classes

jecisc:
When I started to use real protocols instead of protocol names, I updated the protocol API to raise an error if we used a protocol from another class in a class. 

Now that I have more insight on how protocols are managed, I think that the system would be way simpler if the API could accept those and adapt to create/use a protocol of the same name instead. This is the goal of this PR.

Some example of times where it is useful:
- In traits when they categorize the methods because the protocol can come from the user class for example.
- In refactoring when you do a push up/down

Most changes are tests added


jecisc:
Test failures are not related to this PR.

Ducasse:
You will have to explain what is an external v/s internal :)

PR #13968, opened 2023-06-10
Move some session manager registrations

jecisc:
Subpart of #13920 to help find the source of the bootstrap error

### DIFF
--- a/src/Kernel/SmallInteger.class.st
+++ b/src/Kernel/SmallInteger.class.st
@@ -26,9 +26,7 @@ SmallInteger class >> basicNew [
 
 { #category : #'class initialization' }
 SmallInteger class >> initialize [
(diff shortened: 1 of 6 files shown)


jecisc:
Test failures are not from this PR

Ducasse:
Did you discuss with G/P because the session manager was done to reduce the dependencies.
And each class had to register to the session manager. 
Now regrouping the registration can make sense but also some tradeoffs.

jecisc:
The goal is to only group the registration of the kernel. I would like to reduce the dependencies of the kernel to reduce what espell is loading and making the bootstrap more reliable and faster. With the session manager extracted from the Kernel group, we can extract at least 5 other packages used only by the session manager such as UIManager

PR #13970, opened 2023-06-10
Rename packages announcements

jecisc:
Here is a step toward the integration of RPackage in the Metamodel.

Announcements related to packages in the system are not consistent with the rest of the system currently:
- Their name does not match the other announcement names in the system such as ClassAdded or ProtocolAdded
- Their name starts with an R as all classes in RPackage. Since we are integrating them in the MM, we should drop the R to be consistent with the rest of Pharo MM
- We have a package announcement hierarchy and a category announcement hierarchy

This PR fixes the first two problems:
- RPackageRegistered -> PackageAdded
- RPackageRenamed -> PackageRenamed
- RPackageUnregistered -> PackageRemoved

**Future work:**
For the third problem we will have tp work some more. Currently the CategoryAdded announcement is fired each time we create a package or a package tag. PackageAdded is fired when a package is created. In the future I would like to have PackageAdded and PackageTag added and I want to remove CategoryAdded because I want to remove the system of Category that mixes Packages and PackageTags.

### DIFF
--- a/src/Calypso-SystemQueries/PackageRenamed.extension.st
+++ b/src/Calypso-SystemQueries/PackageRenamed.extension.st
@@ -1,0 +1,6 @@
+Extension { #name : #PackageRenamed }
+
+{ #category : #'*Calypso-SystemQueries' }
+PackageRenamed >> canAffectResultOfMethodQuery: aMethodQuery [
+	^true
+]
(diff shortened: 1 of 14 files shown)


Ducasse:
Nice step!

Ducasse:
Broken tests are unrelated.

PR #13973, opened 2023-06-10
13972 create a separate data for tests for rb1 and rb2

Ducasse:
13972 create a separate data for tests for rb1 and rb2
also fix the as yet unclassified. 
RB* tests are all green


Ducasse:
Nice all tests pass :)

PR #13977, opened 2023-06-11
Emh/merge t set up into class

Ducasse:
Inline and remove spurious trait in refactoring tests.


Ducasse:
All tests are passing. So merge!

PR #13979, opened 2023-06-11
Emh/failure in selector

Ducasse:
simply and automatically ensuring an invariant on method name.
all tests are green on my machine.


Ducasse:
All tests are green. So I integrate because this is blocking me.

PR #13980, opened 2023-06-11
Removing RBDummyCLass and subclasses - all tests green

Ducasse:
A new stream of enhancements...
- first Removing RBDummyCLass and subclasses - all tests green


Ducasse:
Still cleaning more.... I will push it in this stream so wait before integrating it.

Ducasse:
I did not succeed to manage the class data as real class but at least I clean the strings. 
It will be easier for the following iteration.

Ducasse:
Finally good!

PR #13987, opened 2023-06-11
Rework UITheme tree forms for better display scaling support

Rinzwind:
This pull request reworks the tree forms in UITheme for better display scaling support: the scaling of the predrawn forms in `#newTreeExpandedForm` and `#newTreeUnexpandedForm` is replaced by a method `#newTreeForm:size:color:` that redraws the forms at the scaled size.

Screenshot of the top left part of a System Browser before and after these changes while using display scale factor 2 (see pull request #13848):

<p align="center">
 
</p>


Ducasse:
Nice to replace the form stored once for all by a computation and to cache it too :)

PR #13989, opened 2023-06-12
tag more tests with ignoreNotImplementedSelectors: pragma

MarcusDenker:
... slowly...

### DIFF
--- a/src/Deprecated12/SHRBTextStyler.extension.st
+++ b/src/Deprecated12/SHRBTextStyler.extension.st
@@ -19,12 +19,3 @@ SHRBTextStyler class >> initialTextAttributesForPixelHeight: aNumber [
 	self deprecated: 'Use #initialTextAttributes instead.' transformWith: '`@rcv initialTextAttributesForPixelHeight: `@arg1' -> '`@rcv initialTextAttributes'.
 	^ self initialTextAttributes
 ]
-
-{ #category : #'*Deprecated12' }
-SHRBTextStyler class >> initialTextAttributesForPixelHeight: aNumber styleTable: aStyleTable [
-
-	self
-		deprecated: 'Use #initialTextAttributesForStyleTable: instead.'
-		transformWith: '`@rcv initialTextAttributesForPixelHeight: `@arg1 styleTable: `arg2' -> '`@rcv initialTextAttributesForStyleTable: `@arg2'.
-	^ self initialTextAttributesForStyleTable: aStyleTable
-]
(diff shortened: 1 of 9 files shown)


src/Deprecated12/SHRBTextStyler.extension.st
```diff
@@ -19,12 +19,3 @@ SHRBTextStyler class >> initialTextAttributesForPixelHeight: aNumber [
...
 ]
-
-{ #category : #'*Deprecated12' }
-SHRBTextStyler class >> initialTextAttributesForPixelHeight: aNumber styleTable: aStyleTable [
-
-	self
-		deprecated: 'Use #initialTextAttributesForStyleTable: instead.'
-		transformWith: '`@rcv initialTextAttributesForPixelHeight: `@arg1 styleTable: `arg2' -> '`@rcv initialTextAttributesForStyleTable: `@arg2'.
-	^ self initialTextAttributesForStyleTable: aStyleTable
-]
```
jecisc, line 30: Is this removal intentional?

src/TraitsV2-Tests/ClassTraitTest.class.st
```diff
@@ -50,6 +50,7 @@ ClassTraitTest >> testConflictsAliasesAndExclusions [
 { #category : #tests }
 ClassTraitTest >> testInitialization [
 	| classTrait |
+		<ignoreNotImplementedSelectors: #(m2ClassSide:)>
```
jecisc:
```suggestion
	<ignoreNotImplementedSelectors: #(m2ClassSide:)>
```

src/Deprecated12/SHRBTextStyler.extension.st
MarcusDenker: no, and I guess I have to close this PR again as I have no clue how to fix it, as the image I used I did a spec pr before and spec thinks that it owns this method

MarcusDenker: I opened:

SHRBTextStyler class>>#initialTextAttributesForPixelHeight: is in Deprecated12, but spec thinks it is in spec #13990

MarcusDenker: So how do I fix this PR? I do not want to redo that trivial copy and past yet again...

MarcusDenker: In my image, this method is in Deprecated12, I can even edit it, but when I commit it is ingored

jecisc: https://stackoverflow.com/questions/215718/how-do-i-reset-or-revert-a-file-to-a-specific-revision

This seems to be what you need

MarcusDenker: that's too complex for me (and will nor work, as there is no verson in git as iceberg just does not see this change)

PR #13992, opened 2023-06-12
removing spurious class compilation and logging in Epicea in RBRefact…

Ducasse:
…oringChangeTest (more to come)

So there is place that remove the class I just created!!! Grgrgr meta stuff.


Ducasse:
I do not know how to control release test

Ducasse:
I will fix it in another PR because do not have the time to understand how to do it properly with git.

PR #13998, opened 2023-06-12
Use real protocols in MCPackageManager

jecisc:
Move from protocol names to real protocols in MCPackageManager.

I also removed CompiledMethod>>mcWorkingCopy that seems to be dead code.


jecisc:
I'm doing bigger cleaning in this part of the system so I'll wait that it is done before redoing this change.

PR #13999, opened 2023-06-12
Handle single-sized collection in stdev method.

### LINKED ISSUE #13993: Make stdev answer 0 for one-sized collections
stdev is a method implemented in Collection of the package Collections-Arithmetic.

Right now, the result of sending stdev to one-sized collection is a ZeroDivide exception.

The standard deviation mathematically involves taking the square root of the variance, and the variance formula requires at least two values to compute meaningful results.

But in practice, calculating the stdev for a single value will always yield a result of zero since there is no variability or deviation from the single value (because there are no other values to compare it to). So technically we can calculate the standard deviation for a single value, although it may not provide meaningful information in that context.

Answering 0 is consistent with other programming languages implementations such as Python:

```python
import numpy as np
a = np.array([[4]])
print(np.std(a))

"0.0" 
```

R on the other side returns NA (it does not raise an Exception though)

```r
v <- c(1)
s<-sd(v)
print(s)
```

**Expected behavior**
Return a zero integer when sending #stdev to one sized collections.

**Expected development cost**
One-line addition to #stdev method.

**Screenshots**
Screenshot 2023-06-12 at 12 35 30

**Version information:**
 - Pharo Version: 12

hernanmd:
This PR solves the issue https://github.com/pharo-project/pharo/issues/13993

Single-sized collection will raise a ZeroDivide exception when stdev is received. However, we can determine the standard deviation for a single value, althouth it might not be useful, it is better zero to avoid unnecessary raising of exceptions.

Add standard deviation tests for collection of 3 numbers, empty and single-element case.

### DIFF
--- a/src/Collections-Arithmetic-Tests/CollectionArithmeticTest.class.st
+++ b/src/Collections-Arithmetic-Tests/CollectionArithmeticTest.class.st
@@ -127,3 +127,25 @@ CollectionArithmeticTest >> testRunningMin [
 
 	self assert: result equals: {1 . 1 . 2 . 2}
 ]
+
+{ #category : #tests }
+CollectionArithmeticTest >> testStdev [
+
+	| collection |
+	collection := #(1 2 3).
+	self assert: collection stdev equals: 1
+]
+
+{ #category : #tests }
+CollectionArithmeticTest >> testStdevIfEmpty [
+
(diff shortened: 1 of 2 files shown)


jecisc:
Test failures seems unrelated

jecisc (approved):

MarcusDenker (approved):

PR #14006, opened 2023-06-12
Clean redundant preconditions by defining a common supercase.

Ducasse:


Ducasse:
Tests not related to this PR. I will fix the one with the case of the instance variable of the metaclass.

Issue #13990, opened 2023-06-12
SHRBTextStyler class>>#initialTextAttributesForPixelHeight: is in Deprecated12, but spec thinks it is in spec


MarcusDenker:
**Bug description**
The system is confused about SHRBTextStyler class>>#initialTextAttributesForPixelHeight: 

- this is in Deprecated12
- but spec has it, too

So if you load spec, you can end up in a situation where the method then is removed from pharo if you commit a pharo PR

MarcusDenker:
fixed by merging the PRs

PR #14011, opened 2023-06-13
move the extension methods of Spec to the class SHRBTextStyler

MarcusDenker:
one step of #13990


MarcusDenker:
The spec PR is here:

https://github.com/pharo-spec/Spec/pull/1413

We need to first merge the pharo change

PR #14012, opened 2023-06-13
Clean manifest builder test and cut dependency to Monticello

jecisc:
I started this change by cutting the dependency of ManifestBuilder to Monticello. Monticello was used as a way to get a system package but we can directly ask the package without this useless dependency.

I also did other cleanings such as removing duplicated cleanings or introducing a better setup for BuilderManifestTest and SmalllintManifestCheckerTest.  I also removed some things such as double point at the end of some statements.


jecisc:
Failures are not related

Ducasse:
difficult to read :)

PR #14013, opened 2023-06-13
first fix some release test and may be a test is readded by the tests

Ducasse:
first fix some of the release tests and may be a test is readded by the tests


jecisc (approved):
Those variables are indeed not used and are good to be removed

PR #14015, opened 2023-06-13
Fix broken release tests + improve a test

jecisc:
I fixed the two failing release tests. One by adding an exception and one by removing unused variables (tests referencing the class are green in my image).

I also updated testNoUnusedInstanceVariablesLeft to not hardcode the violations but let the users declare them. 

FInally, I fixed some other violations that were in the valid exceptions but did not make sense in reality.

Future steps:
- Use the new exception pragma in testNoUnusedClassVariablesLeft and testNoUnusedTemporaryVariablesLeft
- Take into account this pragma in the rules

### DIFF
--- a/src/Kernel-Tests/MockWithComplexSlot.class.st
+++ b/src/Kernel-Tests/MockWithComplexSlot.class.st
@@ -15,6 +15,7 @@ Class {
 { #category : #initialization }
 MockWithComplexSlot >> initialize [
 
+	<ignoreUnusedVariables: #( #aComplexSlot )>
 	self class initializeSlots: self.
 	super initialize
 ]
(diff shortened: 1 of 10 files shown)


Ducasse:
Thanks I did a PR to fix the problem in RB tests.

Ducasse:
We have one test broken and related to the this changes.

```
osx-64 / Tests-osx-64 / testAbstractInstanceVariable(#rbClass->RBAbstractInstanceVariableRefactoring #constructor->#variable:class:) – MacOSX64.Refactoring2.Transformations.Tests.Parametrized.RBAbstractInstanceVariableParametrizedTest
```

jecisc:
Tests are green now (Windows is waiting for its timeout)

PR #14021, opened 2023-06-13
Change colors of the ‘grab’ and ‘rotate’ halo morph handles for better contrast with their icons

Rinzwind:
This pull request changes the colors of the ‘grab’ and ‘rotate’ halo morph handles for better contrast with their icons.

Before&after:

<p align="center">
 
</p>

### DIFF
--- a/src/Morphic-Base/HaloMorph.class.st
+++ b/src/Morphic-Base/HaloMorph.class.st
@@ -43,7 +43,7 @@ HaloMorph class >> classicHaloSpecifications [
 		---------				------		-----------		-------------------------------		---------------"
 	(addMenuHandle:		left			top				(red)							none)
 	(addDismissHandle:		leftCenter	top				(red		muchLighter)			#haloDismissIcon)
-	(addGrabHandle:			center		top				(black)							none)
+	(addGrabHandle:			center		top				(gray)							none)
 	(addDragHandle:			rightCenter	top				(brown)							none)
 	(addDupHandle:			right		top				(green)							none)
 	(addDebugHandle:		right		topCenter		(blue	veryMuchLighter)		none)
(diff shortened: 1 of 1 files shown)


Rinzwind:
Instead of changing the colors, the icons could be changed from black to white, as was actually the case for the ‘grab’, ‘rotate’ and ‘drag’ handles in [Squeak 2.8](https://squeak.js.org/run/#url=https://freudenbergs.de/vanessa/squeakjs&files=[Squeak2.8.image,Squeak2.8.changes,SqueakV2.sources]&swapButtons=true):

<p align="center">

</p>

jecisc (approved):

jecisc:
This is a better color indeed. For the text color I agree that it makes better contraste. Maybe the color could be choose depending on the luminance of the background color?

Rinzwind:
There’s a send of `#color:` in `#createHandleAt:color:iconName:` in HaloMorph for that still:

https://github.com/pharo-project/pharo/blob/3ac2aa82a21c64acbcb8cad01526418fd7ba1163/src/Morphic-Base/HaloMorph.class.st#L725-L726

It doesn’t have the intended effect anymore because the `#depth` of the Forms is no longer 1. A possible fix for that is to replace the above two lines by:

```smalltalk
image form: (aColor luminance > 0.5
	ifTrue: [ form ]
	ifFalse: [ form collectColors: [ :formColor |
		Color h: formColor hue s: formColor saturation
			l: 1.0 - formColor lightness alpha: formColor alpha ] ]).
```

Screenshots of that with the old colors, and with the new colors introduced by this pull request:

<p align="center">
 
</p>

For reference, in Pharo 9 the Forms still had depth 1, screenshot:

<p align="center">

</p>

PR #14025, opened 2023-06-14
Reduce Monticello coupling in RPackage

jecisc:
My goal is to kill all dependencies of RPackage to Monticello. Here is a new step.

Instead of making RPackage listen to the MCWorkingCopyCreated announcement, I make Monticello directly ensure we have the package in the system.

I also provided a small cleaning by deprecating #packageExactlyMatchingExtensionName: in favor of #packageNamedIgnoreCase:ifAbsent:. The reasons are that first I don't find the name really revealing. For me "extension" is about extension methods, but here we get a package name as parameter and this has nothing to do with extension methods. And also there is no need to have multiple methods doing the same things with a slight implementation change. This improves the consistance of RPackage.

### DIFF
--- a/src/Deprecated12/RPackageOrganizer.extension.st
+++ b/src/Deprecated12/RPackageOrganizer.extension.st
@@ -1,0 +1,11 @@
+Extension { #name : #RPackageOrganizer }
+
+{ #category : #'*Deprecated12' }
+RPackageOrganizer >> packageExactlyMatchingExtensionName: anExtensionName [
+	"only look for a package for which the name match 'anExtensionName', making no difference about case. Return nil if no package is found"
+
+	self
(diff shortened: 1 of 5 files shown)


jecisc:
Some tests were failing because with the previous implementation registering a MCWrokingCopy was creating an instance of RPackage in all instances of RPackageOrganizer. 

With this change, the creation of a package happens only in one RPackageOrganizer. By default this is the default one of the system, and I added a hood to be able to specify the package organization we want. This should fix the tests that are using their own instance of RPackageOrganizer.

jecisc:
One less Monticello reference in RPackage :)

MarcusDenker (approved):

PR #14026, opened 2023-06-14
Remove RPackageOrganizer package name cache

jecisc:
In RPackageOrganizer we have a cache on the package names because we are returning a sorted list of the packages names but this is not good. The model should not be responsible for sorting names. This is the responsibility of the UI!

I checked the senders, and ideally, none of them should need the package names sorted. I just found one place where it could have an impacte because it will change the order in which the tests are executed. But if this cause too much trouble I'll sort the package there until we can fix those tests order problems.

This simplifies a little more RPackage management


jecisc:
This has unexpected test failures. This is hard to understand from where they are coming from. I'll try to keep the same order for tests running

MarcusDenker (approved):

jecisc:
Tests seems to have passed if we exclude the windows ones :)

PR #14027, opened 2023-06-14
SHRBTextStyler-simplify-stylehandling 

MarcusDenker:
the pragma was read both in #setStyleTableNamed: of the preferences and #styleTableNamed: of  SHRBTextStyler.

- just querry in #styleTableNamed:, this will make changes later simpler
- fix a bug introduced before (class missing)
- move return out of blocks in some #styleNameIn: methods (to unify style with the others)


jecisc (approved):
Nice one!

PR #14028, opened 2023-06-14
Fluid: show empty slots on the class side, too

MarcusDenker:
Fluid shows empty slots on the instance side so it is easy to add more. But on the class side, it does not.

This PR just applies the same #displayEmptySlots check


MarcusDenker:
This fixes now the class side: if was always printing the ; at the end of the slots

I opted for a local fix for now, in a later change we should move the printing of the ; to the main defintion print method

PR #14029, opened 2023-06-14
Remove some unecessary packages

jecisc:
This PR removes some unnecessary packages from the Pharo repository. 

- Calypso-SystemTools-Core-Tests: I tried to load those tests but the code is broken and almost all tests are broken/failing. There is not much interest in keeping them, especially since we plan to work on Calypso in the future
- System-Reporter now is replaced by NewTools-SystemReporter
- Tools-CriticBrowser now has an equivalent in NewTools
- Microdown-Calypso contains code that is now in other packages
- Commander-SpecSupport is based on Spec1 and can be removed


MarcusDenker (approved):

Ducasse:
thanks cyril

PR #14033, opened 2023-06-14
Make use of CompiledMethod>>protocol

jecisc:
Now that we have CompiledMethod>>protocol returning a real protocol we can use it in places that needs a method protocol instead of showing the implementation details of using ClassDescription>>#protocolOfSelector:


MarcusDenker (approved):

MarcusDenker:
```
MessageNotUnderstood(Exception)>>signal
ObjectsAsMethodsExample(Object)>>doesNotUnderstand: #protocol
```

I think you just have to implement that in ObjectsAsMethodsExample

jecisc:
I'll take a look thanks!

jecisc:
I'll close this one for now.

PR #14036, opened 2023-06-14
Move WorkingCopy creation to Monticello

jecisc:
New step at decoupling Monticello and RPackage. 

This change moves the creation of the working copies to Monticello by listening to RPackage announcements


jecisc:
Failing tests are showing some missing announcements from RPackage. I tried to add the missing ones but the code is really hard to follow because of the ping pong between Monticello and RPackage and also the lack of coherence in RPackage. I found at least 5 methods to register packages and each of them have their particularities. Some firing announcements, some not, some doing the same thing but with different implementations. 

I think I'll try to clean some of this stuff before fixing this PR.

jecisc:
With the recent improvements this one now passes!

This one includes https://github.com/pharo-project/pharo/pull/14065 but the changes are still small enough so that we can review them.

MarcusDenker (approved):

PR #14048, opened 2023-06-15
Enh/refact pass on preconditions

Ducasse:


MarcusDenker:
CI is not running test, I will close and reopen

Ducasse:
I started to do one change because the API sucks but this is generating too many broken tests and after three days trying to stabilize it is better to integrate this and I will see how I can attack the problem in a different ways. But the code is not good because it mixes the name of the class and the class itself!

PR #14049, opened 2023-06-16
Deprecated CompiledMethod>>tagWith:

jecisc:
#tagWith: is missleading because it let the user think that there can be multiple tags on a method, but that is not the case. There can be only one protocol.
This change deprecate it in favor of the already existing #protocol: method.

I also reduced the number of #asSymbol sent since the protocol management can deal with strings.


jecisc:
I have the impression the fail is not linked to this PR: missing workspace /builds/workspace/est_and_branch_Pipeline_PR-14049/64 on pharo-ci-jenkins2-unix-3

MarcusDenker (approved):

PR #14053, opened 2023-06-16
7218-CDClassDefinitionParser-raises-an-error-when-copying-ProtoObject-definition

### LINKED ISSUE #7218: CDClassDefinitionParser raises an error when copying ProtoObject definition
```
ProtoObject subclass: #ProtoObject
	instanceVariableNames: ''
	classVariableNames: ''
	package: 'Kernel-Objects'.
ProtoObject superclass: nil
```

but the new fluid syntax parser handles it well.

MarcusDenker:
This fixes #7218

When defining nil subclasses with non-fluid, at some point the ST80 idea of having class creation methods on nil was replaced by doing a subclass of ProtoObject and then changing the superclass. This was done as due to the explosion of creation methods, the old way did not scale.

With fluid, we are back with nil subclass. Why not do this for non-fluid, too? We just need to support the old method and the one printed now in case fluid is disabled

### DIFF
--- a/src/Calypso-SystemQueries/ClySystemEnvironment.class.st
+++ b/src/Calypso-SystemQueries/ClySystemEnvironment.class.st
@@ -265,6 +265,8 @@ ClySystemEnvironment >> initialize [
 ClySystemEnvironment >> isClassDefinition: newClassDefinitionString [
 	| defTokens |
 	defTokens := newClassDefinitionString findTokens: Character separators.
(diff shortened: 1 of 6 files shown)


MarcusDenker:
Nope, dumb idea: this works for the nil subclasses in the image. But there might be external ones e.g with traits. 

So the only solution is to extend the parser to parse the whole complex case

PR #14062, opened 2023-06-19
Move Monticello diffs to MonticelloGUI

jecisc:
The package Tool-Diff implements morphs to do diffs. But it also contains all the code of the diffs of Monticello. IMO this is not the place of this code so I'm proposing to move the Diff code specific to Monticello in MonticelloGUI.


jecisc:
Here also I have the impression that the dependency warnings are just wrong and do not represent what is actually in the code...

jecisc:
I did an alternative that the CI seems to prefer

PR #14063, opened 2023-06-19
Move Monticello diff merge to MonticelloGUI

jecisc:
Subpart of https://github.com/pharo-project/pharo/pull/14062

Since Monticello seems to have some weird errors let's try a smaller PR.

This PR moves the diff morphs specific to Monticello from Tool-Diff to MonticelloGUI package


jecisc:
Yet another strange error...

PR #14066, opened 2023-06-19
ClassBuilder users: Do not set  'Unclassified' category/package explicitly

### LINKED ISSUE #14057: ClassBuilder users: do not set 'Unclassified' Package/Category
**Describe the problem**

If you look at ShiftClassBuilder>>#initialize, the classbuilder uses the string  ‘Unclassified’ for category as a default value.

In addition, many users set this explicitly, e.g. Class>>#newSubclass (search for the string 'Unclassified'. to find more)

# TODO

Change all clients to just not set the category. This will make is easier later to cleanup package management in the class builder

MarcusDenker:
fixes #14057


jecisc (approved):
Nice!

PR #14067, opened 2023-06-19
Move Monticello diffs to its own package

jecisc:
Alternative to #14062

The package Tool-Diff implements morphs to do diffs. But it also contains all the code of the diffs of Monticello. IMO this is not the place of this code so I'm proposing to move the Diff code specific to Monticello in Monticello-GUI-Diff


jecisc:
I made sure that Morphic does not depend on Monticello stuff like it was the case for now.

MarcusDenker (approved):

PR #14071, opened 2023-06-20
rename ReservedVariable to PseudoVariable 

MarcusDenker:
- rename ReservedVariable to PseudoVariable to be in sync with how wecall them

- test method is #isPseudoVariable
- do not check specially for storing: they are read-only, not reserved

This PR does not change the ivar in SystemDictionary (for later), it keep the hard error for shadowing for now

No deprecation needed as this is internal, not public API


jecisc (approved):
I think it might be nice to add a deprecated alias since this can be the variable of a RBVariableNode that the user can use and extend to manipulate the AST.

MarcusDenker:
I will add an alias with the next iteration

PR #14074, opened 2023-06-20
ClassBuilder: reduce the need of #asSlotCollection and  #asTraitComposition

MarcusDenker:
improve the usage/api of ClassBuilder:
- reduce the need of #asSlotCollection by adding #slotsFromString: following the existing #sharedVariablesFromString:
- reduce the sends of #asTraitComposition,  this is done by the #traitComposition: method of the class builder


jecisc (approved):
Nice :)

PR #14076, opened 2023-06-20
Fix package file out

jecisc:
Leftover API not removed (Maybe from a merge?)


MarcusDenker (approved):

guillep (commented):
Why not a test?

jecisc:
I wanted to fix it fast and I added to my todo list to add a test about that soon

PR #14082, opened 2023-06-21
PseudoVariable: add deprecations / cleanups

### LINKED ISSUE #13675: Compiler: improve ReservedVariables

MarcusDenker:
- add deprecated versions of #isReservedVariab
- fix comments / error messages
- rename ivar

### DIFF
--- a/src/AST-Core/RBProgramNode.class.st
+++ b/src/AST-Core/RBProgramNode.class.st
@@ -796,6 +796,15 @@ RBProgramNode >> isPseudoVariable [
 ]
 
 { #category : #testing }
+RBProgramNode >> isReservedVariable [
+	self
+		deprecated: 'Please use #isPseudoVariable instead'
+		transformWith: '`@receiver isReservedVariable' -> '`@receiver isPseudoVariable'.
+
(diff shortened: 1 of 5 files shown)


jecisc (approved):
Nice!

Thank you for the deprecations. I have a project based on the AST and I know that it's often hard to update it when the AST changes without deprecations :)

PR #14086, opened 2023-06-21
RPackage: Do not synchronize Categories and Packages throught announcements

jecisc:
Currently we have a duplication of info between RPackage and what was the SystemOrganizer. 

SystemOrganizer fires Category announcements and RPackage listen to it to synchronize itself. I propose here to let the code of the SystemOrganizer update itself the RPackage model instead of using the announcements. Announcements that I plan to remove with the merging of SystemOrganizer and RPackageOrganizer


jecisc:
It seems that Espell does not like it if we do like this..

jecisc:
I'll close this one for now and redo it with smaller PRs

PR #14087, opened 2023-06-21
Epicea: Clean package addition and fix bugs

jecisc:
The management of packages and package tags in Epicea is really bad because it is built on the concept of "categories" that is a mix of packages and package tags. This introduced unnecessary complications and bugs in Epicea. 

Here is the first step to remove the usage of this old package management style.

Big changes:
- The initialization of EpCategoryChange has been simplified
- EpCategoryRegistration has been renamed EpPackageAddition and the methods and temporaries related were changed as well
- EpPackageAddition>>categoryName cannot be used anymore since a Package event should not deal with any category name
- #visitPackageChange: has been added to the CodeChangeVisitor to distinguish Packages and tags and simplify the code (before the code was dealing with both, now, we'll know when we have a package and when we have a tag). Eg pBrowseVisitor>>visitPackageChange:
- Package addition visitors now deal only with Packages and no category anymore
- Epicea browser now print "package" next to a package change and not "package tag" as it used to print (This was REALLY missleading to me)

Low hanging fruits:
- Add <script> pragmas to help while developping Epicea
- Use #packageOrganizer to access the default RPackageOrganizer instead of other methods

Next steps:
- Clean the other Category events
- EpPackageAddition should not inherit from EpCategoryChange
- EpApplyVisitor>>visitPackageAddition: does nothing and this is probably a bug.
- EpInverseVisitor>>visitPackageAddition: return a PackageAddition which is really weird. I think we miss a PackageRemoval event
- Update package addition tests


jecisc:
The only thing that should impact the users here is that package addition will be displayed with "Package" instead of "PackageTag" in the browser. 

The other important things is that I found that package addition are not applied by epicea and the revertion of this action is not good. So those changes cannot be replayed or reverted.

Ducasse:
Yes it was always a problem so good that you clean it.

PR #14088, opened 2023-06-21
Fix trait definition in Calypso when fluid is enabled

### LINKED ISSUE #13345: Cannot create trait with fluid class builder
In the latest Pharo 12 I am not able to create a new trait if I use the fluid class builder. It fails silently without anything happening.

This is weird that we do not have any test on that :O

MarcusDenker:
- ClySystemEnvironment>>#defineTrait:notifying:startingFrom: needs to use #fluidInstall
- The check is done already in the caller (for both classes and traits), removed (just like for classes already done)
- ClySystemEnvironment>>#defineNewClassFrom:notifying:startingFrom: should call #fluidInstall
- simplify ClySystemEnvironment>>#compileANewClassFrom:notifying:startingFrom:, we do not need to handle fluid specially

Minimal changes, more cleanups needed later (there is an issue for #compileANewClassFrom:notifying:startingFrom: already)

fixes #13345

### DIFF
--- a/src/Calypso-SystemPlugins-Traits-Queries/ClySystemEnvironment.extension.st
+++ b/src/Calypso-SystemPlugins-Traits-Queries/ClySystemEnvironment.extension.st
@@ -3,24 +3,14 @@ Extension { #name : #ClySystemEnvironment }
 { #category : #'*Calypso-SystemPlugins-Traits-Queries' }
 ClySystemEnvironment >> defineTrait: defString notifying: aController startingFrom: aClass [
 
-	| defTokens keywdIx oldTrait newTraitName trait |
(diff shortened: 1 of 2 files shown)


jecisc (approved):

Ducasse:
Hi marcus 

I would like to know but not online why we need fluidInstall and how we should improve so that we can avoid the isKindOf and use an install message

MarcusDenker:
The fluid definition evaluates to a builder, the old one creates the class with the eval and returns that.

In this case here I think an idea can be to use the ClassParser (it can parse both kinds  to a tree representation) and then use the #buildFromAST: feature of the Class Installer, which uses this AST like tree to create and install the class. This would avoid eval completely.

PR #14094, opened 2023-06-22
Epicea: Rename CategoryAddition and CategoryRemoval

jecisc:
Rename EpCategoryAddition and EpCategoryRemoval into EpPackageTagAddition and EpPackageTagRemoval.

It also updates the announcements to listen to from CategoryAdded and CategoryRemoved to PackageTagAdded and PackageTagRemoved (since we want to remove the categories soon).

The API of the visit methods are still using the category API instead of the package and package tag API for now. There is still EpCategoryRenamed that uses the old names and announcements but I found a bug in RPackage that prevents to do the change right now. I'll need to fix it before finishing this change.

Next steps:
- Split the EpCategoryChange hierarchy between EpPackageChange and EpPackageTagChange hierarchies
- Update the API used by the visit methods to not use categories API
- Rename EpCategoryRenamed after fixing the RPackage bug

Includes https://github.com/pharo-project/pharo/pull/14092


jecisc:
There is a problem with this change because we are losing some package removal events. 

The previous way a package removal was done is not good. I'll have to implement a new EpCodeChange for that to fix it in a clean way.

jecisc:
The changes needed are too big. I'll close this one and try to do a smaller step. Everything is interconnected and it is hard to fix this.

PR #14095, opened 2023-06-22
ClassInstaller: improve #update:to: to do the #fillFor:

### LINKED ISSUE #14084: ShiftClassInstaller class>>#update:to: should do the #fillFor:
if you look at the users of the #update:to: API of the class intaller, they all use #fillFor. Which makes sense: updating means that everything but the things you want to change are the same.

Proposal:
- update:to:  should do the #fillFor: by default
- renove it from the users

MarcusDenker:
The change is backward compatible as calling #fillFor: twice is not a problem. 

using #update:to:  without the fillFor: made no sense, as the class then would loose everything (ivars, layout, superclass)

fixes  #14084


jecisc (approved):
I like that change

PR #14096, opened 2023-06-22
Compiler: allow shadowing of PseudoVariables

MarcusDenker:
This PR unified how variable shadowing is handled: the compiler allows it now for *all* variables, on the level of the tool we warn the user.

Why is this good?

- Uniform, no variable is special due to it's name. There is no use of #isPseudoVariable anywhere anymore!

- if people used thisProcess as a tamp or arg name, it will now work and not raise a hard error. Good for backward compatibility

- removes one  case of  a hard compiler error. Instead accept the code and show the user the error so it can be fixed.


guillep (changes requested):
The tests for shadowing are missing.

This should at least test that the right shadowing variable is written, and not the shadowed one.

PR #14105, opened 2023-06-23
Epicea: Reduces calls to #addCategory:

jecisc:
#addCategory: is a method that should be removed since we want to remove the concept of category.

Here is a PR introducing a method on RPackageOrganizer called #ensureTagNamed:inPackageNamed: to create a tag in a package and some replacements of #addCategory: using this method.

There are still some senders because replacing them breaks the tests. I think there is one place where an announcement is not made creating a slight difference between the SystemOrganizer and the PackageOrganizer. I'll try to fix that.

Depends on https://github.com/pharo-project/pharo/pull/14102


guillep (approved):
LGTM

jecisc:
I removed the "tests passed" tag because I just updated the API I used since it was not managing the events right and I want to see in Epicea tests are still green

jecisc:
Tests are still passing :)

PR #14109, opened 2023-06-23
8782-Pharo-8-Build-1145-File-Browser-error

### LINKED ISSUE #8782: Pharo 8 Build 1145 File Browser error
If you have an emacs autosave file in a given directory, you have a link of the following form
```
john@u20:~$ ls -l .#st_style.txt 
lrwxrwxrwx 1 john john 24 Mar 12 23:44 .#st_style.txt -> john@u20.4844:1615582216
```
Not quite sure exactly what kind of symlink that is, or what other apps create similar, it's not something I've seen anywhere else.  
This causes the debugger to open on a `FileDoesNotExistException` if you open the File Browser in that location.

**To Reproduce**
1. > emacs newfile.txt &
2. Type something then save.
3. Type something more, but don't save
4. . Return to command line
5. . > ls -la
6. Ensure you have a symlink as above
7. Go into a pharo image and visit that directory with the File Browser.
Screenshot from 2021-03-13 00-01-03

**Expected behavior**
I expected the File Browser to either ignore the file or display it as ls -l does

**Screenshots**
Attached

**Version information:**
 - Ubuntu 20.04.1 in a VM on Windows 10
 - Fresh 64bit Pharo 8.0 build 1145

**Expected development cost**
No idea.  Can't we just throw FileNotFound away here as surely only files which exist are going to be fed into the method to retrieve their attributes?  Probably a more elegant solution exists.

** Workaround
Save all emacs buffers before File Browsing?

### LINKED ISSUE #11694: FileList signals a FileDoesNotExistException when opened on a directory containing a ‘dangling’ symbolic link
**Bug description**
`FileList` signals a `FileDoesNotExistException` when opened on a directory containing a ‘dangling’ symbolic link

**To Reproduce**
In the following example, the inspector is opened and shows the contents of the directory, but opening the `FileList` signals a `FileDoesNotExistException`:

```smalltalk
command := 'mkdir /tmp/testdir && ln -s /tmp/testdir/file /tmp/testdir/link'.
(LibC system: command) = 0 ifFalse: [ Error signal ].
fileReference := '/tmp/testdir' asFileReference.
StInspector openOn: fileReference.
FileList openOn: fileReference
```

**Version information:**
> Pharo 11.0.0
> Build information: Pharo-11.0.0+build.213.sha.ecf04188c6ad2a3c2480ca3fe3b9a2988baa1b7a (64 Bit)

MarcusDenker:
This PR allows to use the FileList with a directory that has symlinks which point to non-exisiting files.

The Symlink model of FileSystem seems to be not good, it always operates on the target of the symlink. This PR therefore hardens the FileList to not try to read size/date and permissions of symlinks (it would show the ones of the target, which is not waht you want)

Fixes #8782

### DIFF
--- a/src/Tool-FileList/FileListGridNode.class.st
+++ b/src/Tool-FileList/FileListGridNode.class.st
@@ -14,15 +14,18 @@ FileListGridNode >> fileName [
 
 { #category : #'user interface' }
 FileListGridNode >> filePermissions [
+	item isSymlink ifTrue: [ ^ 'symlink' ].
 	^ self theme newTextIn: self text: item permissions asString
 ]
 
 { #category : #'user interface' }
 FileListGridNode >> fileSize [
+	item isSymlink ifTrue: [ ^ 'symlink' ].
 	^ self theme newTextIn: self text: item humanReadableSize asString
 ]
 
 { #category : #'user interface' }
 FileListGridNode >> modificationDate [
+	item isSymlink ifTrue: [ ^ 'symlink' ].
 	^ self theme newTextIn: self text: item modificationTime asString
 ]


jecisc (approved):

guillep (changes requested):
This needs to have tests.

Also, why returning a string 'symlink' everywhere?

MarcusDenker:
The problem is that FileSystem does not support symlinks for real. As soon as you call e.g. permissions, is tries to get the permissions from the target.

This is more a workaround so people do not get the Error when browsing, a real solution would an extension of FileSystem to support Symlinks, but that is a larger project.

Testing on the level of FileList... hmm... not sure how to do that. We would need to add support for creating of links or use some linux callout to ln... and then somehow open the fileList and somehow check that no errror happens when browsing a directory.

Maybe we just let the error for now and close the PR, I do not think I will work on either of these

PR #14110, opened 2023-06-23
Packages should know their organizer

jecisc:
The RPackage instances do not directy know their organizer, instead they are pointing a class variable that can be updated with a weird method. This leads to really weird behavior. For example if you ask a new organizer instance to register a package, it end up registering it in another PackageOrganizer instance.  The goal of the PR is to allow to save the organizer in an instance variable. 

Since we have a lot of ways to register a package, this will not set the variable in all casses but this is a start. (I want to do small steps since it impacts **a lot** of things)

### DIFF
--- a/src/RPackage-Core/RPackage.class.st
+++ b/src/RPackage-Core/RPackage.class.st
@@ -76,7 +76,8 @@ Class {
 		'metaclassExtensionSelectors',
 		'classes',
 		'name',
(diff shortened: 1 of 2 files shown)


jecisc:
Failures seems unrelated

Ducasse:
What is a package organizer?

PR #14112, opened 2023-06-23
Speedup FLGZippedBasicSerializationTest>>#testWideString

jecisc:
From around 9sec to less than a second


jecisc:
It takes 5.4sec in general on the CI runner of OSX and it should take less than a sec with this change

jecisc:
It now takes 6ms :)

MarcusDenker (approved):

PR #14113, opened 2023-06-23
Deprecate RPackageOrganizer>>#ensureExistAndRegisterPackageNamed:

jecisc:
RPackageOrganizer>>#ensureExistAndRegisterPackageNamed: is bad because it does not register the new category in the SystemOrganizer (and even if I want to kill the SystemOrganizer, we still need it until everything using it is migrated) and it does not launch any PackageAdded event, and this is necessary if we want to remove the CategoryAdded event.

I propose to deprecate it and to make the users use #registerPackageNamed: instead (this method also does the check that the package exists).

Requires #14110


jecisc:
The image seems to work like this but Espell does not seems to like it :(

jecisc:
The conflict to manage here is too big. I'll create a new PR later.

PR #14119, opened 2023-06-25
RPackage: Removing category should remove the tags

jecisc:
Currently we have the SystemOrganizer and the RpackageOrganizer that have duplicated info and are synchronized. But this synchronization have some bugs and one of the is that #removeCategory: does not remove RPackageTag when the category is not a package. This PR fixes that to be sure we remove the RPackageTag also.

This is necessary to be able to migrate Epicea from category management to Package/Tags management (and kill slowly categories)

### DIFF
--- a/src/Calypso-SystemTools-Core/ClyMethodCreationToolMorph.class.st
+++ b/src/Calypso-SystemTools-Core/ClyMethodCreationToolMorph.class.st
@@ -131,7 +131,7 @@ ClyMethodCreationToolMorph >> setUpModelFromContext [
 	context isMethodGroupSelected ifFalse: [ ^ self ].
 	selectedGroup := context lastSelectedMethodGroup.
 	(selectedGroup isKindOf: ClyExternalPackageMethodGroup) ifTrue: [ ^ extendingPackage := selectedGroup package ].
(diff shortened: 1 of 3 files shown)


src/Calypso-SystemTools-Core/ClyMethodCreationToolMorph.class.st
```diff
@@ -131,7 +131,7 @@ ClyMethodCreationToolMorph >> setUpModelFromContext [
 	context isMethodGroupSelected ifFalse: [ ^ self ].
 	selectedGroup := context lastSelectedMethodGroup.
 	(selectedGroup isKindOf: ClyExternalPackageMethodGroup) ifTrue: [ ^ extendingPackage := selectedGroup package ].
-	(selectedGroup isKindOf: ClyMethodsInProtocolGroup) ifTrue: [ ^ methodProtocol := selectedGroup tag ]
+	(selectedGroup isKindOf: ClyMethodsInProtocolGroup) ifTrue: [ ^ methodProtocol := selectedGroup protocol ]
```
jecisc, line 134: This is automatic transformation

MarcusDenker (approved):

PR #14121, opened 2023-06-26
Remove some more categories in Epicea

jecisc:
Now that I fixed a RPackage bug we can remove some more category usage in Epicea. There are still some category usage due to other bugs in RPackage but I'll fix that later.


MarcusDenker:
failing test not related

MarcusDenker (approved):

PR #14124, opened 2023-06-26
Allow repeated file-in (Named ChangeSet)

### LINKED ISSUE #9966: ChangeSet class>>#basicNewChangeSet: does not allow to load a .st or .cs file twice
**Bug description**
When you load a .cs or .st file, a change set if created. Now if you try to load it again (e.g. because you manually edited it), you get 'Sorry that name is already used'.

This behaviour is implemented in ChangeSet class>>#basicNewChangeSet:

Historically there used to be another option to load without creating a dedicated change set, but this option was removed (maybe because it was not clear why it was needed?)

**What we need to do **
The method basicNewChangeSet: should just synthesise a new name if the name already exists. e.g. append a number and increase it (like we do when saving the image).

privat:
For some reason, file-in (drag and drop a file the select "install",) goes trough the creation of a named changeset and fails if the same file is filed-in again.

Two issues, here:

* error is reported as an GUI info popup, so non actionable and non debuggable (and the error message is very bad and nonsensical to the user action).
* why the error exits at the first place?

The whole thing is complex and probably need love, in the meantime the issue is a showstopper. So the current PR is a quick workaround : append a UUID to avoid name collision.

### DIFF
--- a/src/System-Changes-FileServices/ChangeSet.extension.st
+++ b/src/System-Changes-FileServices/ChangeSet.extension.st
@@ -5,7 +5,8 @@ ChangeSet class >> fileIntoNewChangeSet: fullName [
 	"File in all of the contents of the currently selected file, if any, into a new change set."
 	fullName ifNil: [^ self ].
 	fullName asFileReference readStreamDo: [ :readStream |
-		self newChangesFromStream: readStream named: fullName asFileReference basename ]
+		"Use UUID to avoid name collision. I'm not sure way a name is needed nor why is sould be unique BTW"
+		self newChangesFromStream: readStream named: fullName asFileReference basename withoutPeriodSuffix , '_' , UUID new asString ]
 ]
 
 { #category : #'*System-Changes-FileServices' }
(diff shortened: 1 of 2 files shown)


privat:
Oh, the issue was already reported a few years ago #9966

MarcusDenker:
failing tests not related

MarcusDenker (approved):

MarcusDenker:
I wonder if we could not just re-use the existing changeset if it already exists.

privat:
I dot really know what a changeset is, what the changeset is used for when "file in", why it has a name, why the name should be unique, nor what will be the possible ill effects we reuse a changeset or reuse a name. So I just played it safe.

But if you have a better fix, just PR it or update the current PR. UUID is quite an ugly workaround that I do not really like.

Ducasse:
a changeset was a was to exhcngae code: a list of definition and there were tools to merge and build them. We should remove this. And the when we fileout something it should be in tonel format but this is another story

PR #14125, opened 2023-06-26
Calypso: add test for trait and class creation

MarcusDenker:
We did not have tests for ClySystemEnvironment>>#compileANewClassFrom:notifying:startingFrom:

This adds a first test and simplifies the code a bit more


Ducasse:
Cool!

PR #14143, opened 2023-06-28
Remove hack

Ducasse:
- Remove the ugly in: (SycMessageOriginHack of: originalMessage) and related
- Remove some spurious precondition checks (less popup) in rename method.


Ducasse:
There is something strange  I cannot get the build to restart :(

MarcusDenker:
The

MarcusDenker:
So other PRs have no problems, this runs the tests and prints:

Running tests of SystemCommands-RefactoringSupport-Tests


There is no other info in the log, strangely

PR #14144, opened 2023-06-28
Remove instance var driver

Ducasse:
Introduce a driver for RemoveInstanceVariableRefactoring 
Changed  RBRemoveInstanceVariableRefactoring into Into RemoveInstanceVariableRefactoring2
to support the driver logic.
Remove RemoveInstanceVariableRefactoring  
RBRemoveIV2 Will renamed later


Ducasse:
I do not understand why my PR is not loading. I can load it in a fresh image without problem and I did not touch the related methods. I will check but I really do not like coding around this browser

Ducasse:
I need real help because 
MetacelloNotification: Loading -> Calypso-SystemTools-FullBrowser-tonel.1 --- tonel:///builds/workspace/est_and_branch_Pipeline_PR-14144/64/src --- cache
Error: Package *Calypso-SystemTools-FullBrowser depends on the following classes:
  SycRemoveVariableCommand
  SycRemoveVariableCommand>>#fullBrowserMenuActivation
  SycRemoveVariableCommand>>#fullBrowserShortcutActivation


is wrong in my image there is not SycRemoveVariableCommand anymore.
I recompile the methods to see. Nothing changed
and now I have a bug I never got...UNKNOWN package when I tried to checkedout my PR to see.

Capture 2023-06-28 at 21 27 47

Ducasse:
I recompiled the methods and yes the system is in a super bad state.

Capture 2023-06-28 at 21 33 00

Ducasse:
As we can see recompiling the method in the SAME class did change the class. 
Houston we have a bug.

Ducasse:
Now it passes 
Borken tests are unrelated.

PR #14155, opened 2023-06-30
[P11] Traits has to install method when is generated code

tesonep:
When a method is not in the trait, we install it always.
It is useful to generate methods used by Talents.


jecisc:
There was no problem after merging this in P12. It should be safe for p11 but I'll restart the CI because it failed

Issue #14140, opened 2023-06-28
#foo keywords should return #()


Ducasse:
Three actions should be done
- add tests for keywords
```
#foo:bar: keywords 
> #(#foo: #bar:)
#name keywords
> #(#name )
#+ keywords
> #(#+)
```
- improve the method comments to tell that keywords return the receiver when this is a unary or binary message
- introduce a method call strictKeywords

```
#foo:bar: strictKeywords 
> #(#foo: #bar:)
#name strictKeywords
> #()
#+ strictKeywords
> #()
```

MarcusDenker:
pr merged

PR #14157, opened 2023-06-30
Fixed `String>>#keywords`

Mathilde411:
Fixed problem from pharo-project/pharo#14140

Added comment and tests for method `String>>#keywords`.
Created `Symbol>>#keywordsStrict` and tests for new usages.

The final goal is to deprecate `String>>#keywords`.

### DIFF
--- a/src/Collections-Strings-Tests/StringTest.class.st
+++ b/src/Collections-Strings-Tests/StringTest.class.st
@@ -1462,6 +1462,14 @@ StringTest >> testIsWideString [
 ]
 
 { #category : #tests }
+StringTest >> testKeywords [
+
+	self assert: (#foo: keywords) equals: #('foo:').
+	self assert: (#foo:bar: keywords) equals: #('foo:' 'bar:').
+	
+]
+
+{ #category : #tests }
 StringTest >> testLastSpacePosition [
 
 	self assert: 'fred the bear' lastSpacePosition equals: 9.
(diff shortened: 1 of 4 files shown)


src/Collections-Strings-Tests/StringTest.class.st
```diff
@@ -1461,6 +1461,14 @@ StringTest >> testIsWideString [
 	self deny: 'éàôüÖ' isWideString
 ]
 
+{ #category : #tests }
+StringTest >> testKeywords [
+
+	self assert: (#foo: keywords) equals: #('foo:').
+	self assert: (#foo:bar: keywords) equals: #('foo:' 'bar:').
```
Ducasse, line 1468: The test should also show the case for unary and binary messages

src/Collections-Strings/Symbol.class.st
```diff
@@ -495,6 +495,15 @@ Symbol >> isUnary [
 	^ self precedence = 1
 ]
 
+{ #category : #sorting }
+Symbol >> keywordsStrict [
+	"Returns the keywords of the provided selector. Does not assume that the reciever is keyword-based."
+
```
Ducasse, line 509: reciever -> receiver

Please add also some executable comments.
I could not understand what is does not assume. So the examples will be more self explanatory

src/Collections-Strings-Tests/StringTest.class.st
Mathilde411: This is the test for the method that should only accept keyword-based selectors, so there shouldn't be any need to test

guillep:
I think the original issue from @Ducasse was to have a method that answers you every selector part that accepts an argument, that for unary/binary returns an empty collection. In that case, it would be expected to have a test asserting that it returns an empty collection (or not in the non strict case), right?

Ducasse:
agreed!

PR #14159, opened 2023-06-30
	fix-deprecated-send-testCreateGroupsForEveryMethodTag

MarcusDenker:
Fixes a deprecated send


Ducasse:
It is a bit obscure why tag are transformed into protocol but ok!

Issue #14056, opened 2023-06-17
double click selection on symbols should include #


Ducasse:
when we double click on #Foo only Foo is selected and this is super annoying. 
It would be better to select #Foo else we have to select the # by hand.

hernanmd:
This issue can be closed now

PR #14160, opened 2023-06-30
double click selection on symbols should include #

hernanmd:
For issue #14056.
Change in RubTextEditor scan characters behavior (selectWord method ) to include the Character numeral (#) on double click selection on symbols.
Add test.

### DIFF
--- a/src/Rubric-Tests/RubTextEditorTest.class.st
+++ b/src/Rubric-Tests/RubTextEditorTest.class.st
@@ -107,6 +107,48 @@ RubTextEditorTest >> testPreviousWord [
 ]
 
 { #category : #tests }
+RubTextEditorTest >> testSelectWord [
+
+	string := '#Lorem #ipsum #dolor #sit #amet'.
+	editor addString: string.
+
+	editor selectWordMark: 0 point: 0.
+	editor selectWord.
+	
+	self assert: editor hasSelection.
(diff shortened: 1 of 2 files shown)


Ducasse:
Oh thanks a lot :)

jecisc:
I have to say that this is annoying me now ahah

Does it makes sens to have a setting on this?

privat:
I'm afraid that a setting add complexity without solving the issue.
Sometimes you want the # and sometimes you don't. What is simpler, adding a missing one or removing an unwanted one?

guillep:
Yes, adding settings for such simple and essential things does not help :)

jecisc:
Then I'll just change that for myself with metalinks because I personally almost never want to select the # because I often just want to copy a selector or class name that was in the form of a symbol and in the few cases were I want a symbol, I prefer to type the # before doing cmd + v. It is much more bothersome to do cmd + v then going to the beginning to remove the # and then going again to the end to keep typing my code.

Ducasse:
I do not understand why some characters are not part of the selection. Because then why selecting :
I always have to double and select the # because I use it a lot of sender and implementors.

MarcusDenker:
But senders and implementors do not need the # to be selected

jecisc:
The # does not need to be selected to be able to browse the sender and implementors.
Just selecting the letters or I think having the cursor in the symbol is enough

PR #14161, opened 2023-06-30
First version of the renameMethod as a driver.

Ducasse:
+ improve RBCondition
+ add tests
+ better MethodNamePresenter
+ deprecated empty -> true emptyCondition -> trueCondition


Ducasse:
Error: Package Monticello-Tests depends on the following classes:
  StDummy
  StDummy2

I do not get why because these classes never got published and I do not understand why they would be in Monticello-Tests?

Ducasse:
I do not know what to do because the classes were never published. Now I removed them from my image and I will push a new version of the package but this is strange.

src/Refactoring-UI/RBRenameMethodDriver.class.st
```diff
@@ -0,0 +1,94 @@
...
+		              selector: originalMessage selector
+		              arguments: originalMessage argumentNames.
+	dialog := StMethodNameEditorPresenter
+		          openOn: methodName
+		          withInvalidArgs: (RBCondition invalidArgumentNamesForSelector: originalMessage selector in: model)
+		          canRenameArgs: self canRenameArgs
+		          canRemoveArgs: self canRemoveArgs
+		          canAddArgs: self canAddArgs
+		          canEditName: self canEditName.
+	dialog cancelled ifTrue: [ CmdCommandAborted signal ].
```
balsa-sarenac, line 104: I've had this before and my builds were failing because using `Commander2` within the `Refactoring-UI` adds another dependency to core baseline. I've changed this a few times, you can check deprecate method driver on how I've done it

Ducasse: Ok I will have a look

Ducasse:
Now we can have this dependency

Ducasse:
Broken tests are unrelated. 
I push it because this is painful to work.

PR #14166, opened 2023-06-30
EFBenchTest Cleanup

MarcusDenker:
EFBenchTest has unused methods, all tests are skipped. It looks like a test class that was abused for benchmarks, but it is not clear how to use it.

I propose to remove it


Ducasse:
It was used to measure EF. May be can simple removed the skipped tests. This is interesting to know the context that were painful.

MarcusDenker:
all tests are skipped in that class

PR #14167, opened 2023-06-30
Fix haltOn tests after debug them

### LINKED ISSUE #505: Random failure: testHaltOnCount
This week we got two times a random failure of this test in Pharo CI:
testHaltOnCount – MacOSX64.Debugging.Utils.Tests.HaltTest

Here is the stack trace:

```
Error
Assertion failed
Stacktrace
TestFailure
Assertion failed
HaltTest(TestAsserter)>>assert:description:resumable:
HaltTest(TestAsserter)>>assert:description:
HaltTest(Object)>>assert:
HaltTest(TestAsserter)>>should:raise:
HaltTest>>shouldHaltWhen:
HaltTest>>shouldHaltAfter:times:
HaltTest>>testHaltOnCount
HaltTest(TestCase)>>performTest
```

PalumboN:
Fix https://github.com/pharo-spec/NewTools/issues/505

The context is this:
1. When a test is debugged, a breakpoint is installed as a _metalink_ before the test execution (so the execution stops at the beginning of the test).
1. During the execution, these tests call `haltOnCount` and related special halts. These halts put a counter **in the AST** model.
1. As the AST is preserved between executions, these tests produce side effects. So, inside the test, the method test itself is recompiled. 
1. When the test finishes, the _metalink_ with the breakpoint is uninstalled from the method AST.

The problem was with installing/uninstalling the breakpoints. At uninstalling it was keeping the old AST in the new CompiledMethod, so the counter was not reset.

That is why only the next execution after debug is failing, it maintains the counter state from the previous execution.

##### Fixed in the Pharo Sprint on 30/06 with @uNouss 🚀

### DIFF
--- a/src/Calypso-SystemPlugins-SUnit-Browser/ClyDebugTestCommand.class.st
+++ b/src/Calypso-SystemPlugins-SUnit-Browser/ClyDebugTestCommand.class.st
@@ -33,10 +33,12 @@ ClyDebugTestCommand >> defaultMenuItemName [
 { #category : #execution }
 ClyDebugTestCommand >> runTest: testSelector of: testClass [
 
-	| breakpoint |
+	| breakpoint node |
+	node := (testClass lookupSelector: testSelector) ast.
 	breakpoint := Breakpoint new
-		              node: (testClass lookupSelector: testSelector) ast;
+		              node: node;
 		              install.
+
 	[ super runTest: testSelector of: testClass ] ensure: [
 		[ breakpoint remove ]
 			on: NotFound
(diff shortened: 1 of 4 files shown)


PalumboN:
Talking with @MarcusDenker: This solution is not good because if we have many _metalinks_ on the same AST this solution will break the relation 👎 

One proposed solution is to clear the halt counter from the nodes at the moment of removing the breakpoint (or before, inside the test, instead of recompiling the method).

PalumboN:
So, I changed the solution to only fixing the failing tests: I moved the code that needs to be recompiled to another method to avoid collisions with metalinks on executed (test) method.

## We can open another issue for
Talking with people, I think the counters for halts should not be attached to the AST (that persists over many executions) but to the current execution.
I propose to change the meaning of "halt when this code was executed N times" to "halt when the current execution passed thru here N times".
With this change, you can put the `haltOn: N` and re-run many times the same test without recompiling the halt. Now if the method with the halt is not recompiled, then you have a _dead halt_ in your image.
Also, recompilation won't be necessary for these tests anymore.

jecisc (approved):

jecisc:
Thank you!

Issue #14173, opened 2023-07-02
Find closest implementors in hierarchy


Ducasse:
Often I'm browsing a self xxxx message and I would like to press on it and get the implementation from a inheritance point of view. Right now the little up/down arrows are only working if the methods xxx is defined in the class.

Ducasse:
for example I would like to know who is implementing confirm: in the hierarchy of SycReplaceMessageCommand

```
SycReplaceMessageCommand >> replaceInAllClasses
	^ self confirm: 'Do you want replace senders of method in the whole classes?'
```

PR #14183, opened 2023-07-02
#14173 Introduce a nice command to spot the closest method overriden …

Ducasse:
…in the receiver superclasses.

So now we can go quite fast to spot the method that is called


Ducasse:
unrelated

PR #14186, opened 2023-07-03
Improve deprecate method flow in driver

Ducasse:
Better flow from driver to refactoring.
The input validity is now simpler at the refactoring and pushed to the driver on user error.


src/Refactoring-Core/RBDeprecateMethodRefactoring.class.st
```diff
@@ -93,17 +103,26 @@ RBDeprecateMethodRefactoring >> callDeprecationMethod [
...
+	class := self classObjectFor: aClass
+]
+
 { #category : #initialization }
 RBDeprecateMethodRefactoring >> deprecateMethod: aSelector in: aClass using: newSel [
+
+	self deprecateMethod: aSelector in: aClass.
 	oldSelector := aSelector asSymbol.
-	newSelector := newSel asSymbol.
 	class := self classObjectFor: aClass
```
balsa-sarenac: I think here `oldSelector` and `class` declaration should be removed and `newSelector` should stay?

Ducasse: Yes I will fix in the next batch!

Ducasse:
Well I have failing tests so I will have a look at them and fix it on the way

PR #14192, opened 2023-07-03
New flow for `MoveMethodsToClassSide`

balsa-sarenac:
Now this has better flow when user supplies multiple methods to move to class side. 
New flow aggregates all warnings and errors and changes. If there are errors, a pop-up is shown. If there are warnings a yes/no pop-up is shown whether user wants to continue. In the end changes are shown to the user.
`StRefactoringPreviewPresenter`s API is changed so that it supports bulk changes.

Right now it shows `RBRefactoringWarning` if there are breaking changes. On `no` it aborts execution, on `yes` continues without breaking changes included. What do you think if we would on `yes` include all changes (that is include breaking changes) and on `no` just perform the safe ones?


Ducasse:
Borken tests are not related so I merge.

PR #14196, opened 2023-07-03
Extend Behavior>>handleFailingBasicNew:

### LINKED ISSUE #14195: Object allocation (very) occasionally fails
**Bug description**
Object allocation can fail if a process preempts the active process during object allocation.

The object allocation design (for variableSubclasses) is to call `#handleFailingBasicNew:` if the initial attempt fails.

handleFailingBasicNew: then tries to garbage collect, and if the largest free segment is big enough tries the allocation again (`#handleFailingFailingBasicNew:`):

```
handleFailingBasicNew: sizeRequested
	"handleFailingBasicNew: gets sent after basicNew: has failed and allowed
	 a scavenging garbage collection to occur.  The scavenging collection
	 will have happened as the VM is activating the (failing) basicNew:.  If
	 handleFailingBasicNew: fails then the scavenge failed to reclaim sufficient
	 space and a global garbage collection is required.  Retry after garbage
	 collecting and growing memory if necessary.

	 Primitive. Answer an instance of this class with the number of indexable
	 variables specified by the argument, sizeRequested.  Fail if this class is not
	 indexable or if the argument is not a positive Integer, or if there is not
	 enough memory available. Essential. See Object documentation whatIsAPrimitive."

	<primitive: 71>
	| bytesRequested |
	bytesRequested := self byteSizeOfInstanceOfSize: sizeRequested.
	Smalltalk garbageCollect < bytesRequested ifTrue:
		[Smalltalk growMemoryByAtLeast: bytesRequested].
	"retry after global garbage collect and possible grow"
	^self handleFailingFailingBasicNew: sizeRequested
```

However the method can be pre-empted (or switched to another process of the same priority via `processPreemptionYields`) between the GC and the second allocation attempt.  If the preempting process allocates an object in the free segment that reduces the free segment size to less than the requested size here, the second allocation attempt will fail.

There are a couple of solutions.  One would be to wrap the GC and second allocation attempt in a block closure and use `#valueUnpreemptively`, however this increases the cost of every call to fix a problem that is extremely rare.

The proposed PR will instead attempt to grow memory if the second allocation attempt fails, and then do a third allocation attempt.  As the vast majority of times the second attempt succeeds, there is no additional overhead in the normal sequence of events.

**To Reproduce**
I haven't been able to reliably reproduce this, however it happens reasonably regularly in a proprietary test run without the proposed PR and not with it.

**Expected behavior**
The object allocation succeeds.

**Version information:**
 - OS: Ubuntu
 - Version: 20.04
 - Pharo Version: 10 (although the code hasn't changed since then).

**Expected development cost**
PR on the way.

akgrant43:
to handle the case where the process is interrupted between garbage collection and the second allocation attempt to grow memory if the second attempt fails.

Fixes: https://github.com/pharo-project/pharo/issues/14195

### DIFF
--- a/src/Kernel/Behavior.class.st
+++ b/src/Kernel/Behavior.class.st
@@ -689,10 +689,22 @@ Behavior >> handleFailingBasicNew: sizeRequested [
 	<primitive: 71>
 	| bytesRequested |
 	bytesRequested := self byteSizeOfInstanceOfSize: sizeRequested.
-	Smalltalk garbageCollect < bytesRequested ifTrue:
-		[Smalltalk growMemoryByAtLeast: bytesRequested].
-	"retry after global garbage collect and possible grow"
-	^self handleFailingFailingBasicNew: sizeRequested
+	^ Smalltalk garbageCollect < bytesRequested ifTrue: 
+		[ Smalltalk growMemoryByAtLeast: bytesRequested. "retry after global garbage collect and possible grow"
+		self handleFailingFailingBasicNew: sizeRequested ]
+	ifFalse: 
+		[ self handleFailingBasicNewWithGC: sizeRequested ]
+]
+
+{ #category : #private }
+Behavior >> handleFailingBasicNewWithGC: sizeRequested [
+	"handleFailingBasicNewWithGC: gets sent after basicNew: has failed, a GC has been	performed and the allocation still fails even though enough memory was reported 	as available. Given that this has happened when there is plenty of virtual memory available, assume that the GC has reported incorrectly and try one more time anyway.	 Primitive. Answer an instance of this class with the number of indexable	 variables specified by the argument, sizeRequested.  Fail if this class is not	 indexable or if the argument is not a positive Integer, or if there is not	 enough memory available. Essential. See Object documentation whatIsAPrimitive."
+
+	<primitive: 71>
+	| bytesRequested |
+	bytesRequested := self byteSizeOfInstanceOfSize: sizeRequested.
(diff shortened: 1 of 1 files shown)


tesonep (approved):

privat:
A possible improvement that could ensure the correction of the code, and was hinted in the issue, could be to add `valueUnpreemptively` but only for the 2nd attempt (or third one if you really do not want to use it)

akgrant43:
I think this is a good idea.  I'd change to:

```
Behavior>>handleFailingBasicNewWithGC: sizeRequested
	"<comment removed for brevity>"

	<primitive: 71>
	| bytesRequested |
	bytesRequested := self byteSizeOfInstanceOfSize: sizeRequested.
	^ [ Smalltalk growMemoryByAtLeast: bytesRequested.
	self handleFailingFailingBasicNew: sizeRequested ]
		valueUnpreemptively.
```

however I don't want to lose @tesonep 's approval. :-)  Pablo, are you OK with this change?

tesonep:
I think it should work.

guillep:
Yes, if the problem comes from a potential stealing from a different thread, avoiding preemption makes sense :).
It's outside of the fast critical path, so worth trying. If we find this is problematic this can be reverted and a separate solution can be tried.

@akgrant43 did you try it on your side? Does it work well?

akgrant43:
It's been running in our image, including at a large customer, since the beginning of July (a few days before the PR was submitted).   We haven't seen an Out-of-memory error since it was applied. :-)

PR #14208, opened 2023-07-04
SmalltalkImage>>#newSpecialObjectsArray remove elements not used by the VM

privat:
Some elements are set but seem used no more.


guillep (approved):
LGTM

privat:
Thanks, now you can review the sibling but independent PR https://github.com/pharo-project/pharo-vm/pull/640

PR #14211, opened 2023-07-04
Fix cannotInterpret: and test many VM callback methods

privat:
Fix and test `ProtoObject>>cannotInterpret:` as the implementation was broken at many levels.

The PR also add systematic tests for the special selectors that the VM might send in specific situations.
These tests try to check the callbacks as they are handled by default in ProtoObject/Object (booooriiiiing), but mostly check custom defined handling trough redefinition in subclasses (with more complex situations).

Tested:

* `#attemptToAssign:withIndex:`
* `#cannotInterpret:`
* `#doesNotUnderstand:`
* `#mustBeBoolean`
* `#run:with:in:`

What is missing:

* Method sends to the context because I am not sure how to provide custom Context class to the VM: `#cannotReturn:`, `#aboutToReturn:through:`, `#unusedBytecode`
* Special selectors that are not implemented `#conditionalBranchCounterTrippedOn:` and `#classTrapFor:` (I did not look at them yet)


src/Kernel/ProtoObject.class.st
```diff
@@ -67,9 +67,11 @@ ProtoObject >> becomeForward: otherObject copyHash: copyHash [
 ProtoObject >> cannotInterpret: aMessage [
 	 "Handle the fact that there was an attempt to send the given message to the receiver but a null methodDictionary was encountered while looking up the message selector.  Hopefully this is the result of encountering a stub for a swapped out class which induces this exception on purpose."
 
-"If this is the result of encountering a swap-out stub, then simulating the lookup in Smalltalk should suffice to install the class properly, and the message may be resent."
+	"If this is the result of encountering a swap-out stub, then simulating the lookup in Smalltalk should suffice to install the class properly, and the message may be resent."
+	
+	"Important: do not send anything to self (here or in a redefinition) as that will likely cause an infinite recursion of `cannotInterpret:`"
 
-	(self class lookupSelector: aMessage selector) ifNotNil:
+	((Behavior classOf: self) lookupSelector: aMessage selector) ifNotNil:
```
guillep, line 78: But this restarts the lookup from the receiver's class. I don't think that  `lookupSelector:` works if the method dictionary is nil. In that case you may have an exception, and the exception may want to print the objects in the stack, and then it will try to print your proxy. And recurse.

src/Kernel-Tests/SpecialSelectorTest.class.st
```diff
@@ -0,0 +1,55 @@
...
+	Is currenlty broken and goes in an infinite recursion of Context>>nextHandlerContext"
+
+	| class object |
+	"Create a new anonymous subclass of ProtoObject with a nil methodDict"
+	class := Behavior new.
+	class basicSuperclass: ProtoObject.
+	class methodDict: nil.
+	object := class new.
+	"Any message send to `object` should callback cannotInterpret: and raise an error"
+	self should: [ object max: 42 ] raise: Error
```
guillep, line 116: Why would you get an exception here?
- object receives max:
- lookup finds nil method dictionary, lookup is restarted from `ProtoObject` looking for `cannotInterpret:`
- The selector is found (redefined below in this PR), the method in ProtoObject executed
- `ProtoObject>>cannotInterpret:` looks up `max:` but that would raise an exception because `nil` method dictionaries are not expected, and recurse (?)
- But if it would be found, it would return `self`. As soon as another message is sent to it, we enter in loop again!

src/Kernel/ProtoObject.class.st
privat: I checked it. ` lookupSelector:` is naive and just call `methodDict at:` without any check.
But `methodDict` is a robust accessor that returns an empty method dictionary if the ivar is nil. (it's documented to be lenient with tools that do not expect nil as an answer)
So the `lookupSelector:` should work and just look up until ProtoObject, find nothing, and return nil.

Then the next step in `ProtoObject>>cannotInterpret:` is just to `Error signal: 'MethodDictionary fault'.`.
That should be caught by `should:raise:` in the test method, but is not according to Jenkins: https://ci.inria.fr/pharo-ci-jenkins2/blue/organizations/jenkins/Test%20pending%20pull%20request%20and%20branch%20Pipeline/detail/PR-14211/3/tests

My hypothesis is that the stack, context, pc or something internal is inconsistent.

src/Kernel-Tests/SpecialSelectorTest.class.st
privat: as written above, the manual look up of `max:` is ok and return nil (no error here).

src/Kernel-Tests/SpecialSelectorTest.class.st
```diff
@@ -0,0 +1,119 @@
...
+SpecialSelectorTest >> testCannotInterpretProtoObject [
+	"Test the implementation of ProtoObject>>cannotInterpret:
+	Is currenlty broken and goes in an infinite recursion of Context>>nextHandlerContext"
+
+	| class object |
+	"Create a new anonymous subclass of ProtoObject with a nil methodDict"
+	class := Behavior new.
+	class basicSuperclass: ProtoObject.
+	class methodDict: nil.
+	object := class new.
```
privat: I got a bug:

* here I call `new`
* that will call `initialize` on the instance
* that cannot be looked up by the VM because `methDict` is nil
* that will call `cannotInterpret:`
* that will do a manual image-side lookup and find `Object>>initialize`
* that will call `Message>>SentTo:`
* that will call `perform:withArguments:inSuperclass:` on the original receiver
* that cannot be looked up by the VM because `methDict` is nil
* same player plays again...

solution: use `basicNew`

src/Kernel/Behavior.class.st
```diff
@@ -17,6 +17,17 @@ Class {
...
+{ #category : #'simulation support' }
+Behavior class >> classOf: anObject [
+	"The class of anObject.
+	This is equivalent to `anObject class` but without a message to be send to `anObject`.
+	This can be very handy if the class of `anObject` is unreliable.
+	e.g. no inheritance from ProtoObject, or having a nil methodDict."
+
+	<primitive: 111>
+	self primitiveFailed
+]
```
guillep, line 31: This method is already available in `MirrorPrimitives` and there was a previous effort to have it only once in that place (otherwise it was repeated in many places). I propose we remove this method and we use the other.

guillep, line 31:
```suggestion
```

src/Kernel/ProtoObject.class.st
```diff
@@ -67,11 +67,13 @@ ProtoObject >> becomeForward: otherObject copyHash: copyHash [
...
 
-"If this is the result of encountering a swap-out stub, then simulating the lookup in Smalltalk should suffice to install the class properly, and the message may be resent."
+	"If this is the result of encountering a swap-out stub, then simulating the lookup in Smalltalk should suffice to install the class properly, and the message may be resent."
+	
+	"Important: do not send anything to self (here or in a redefinition) as that will likely cause an infinite recursion of `cannotInterpret:`"
 
-	(self class lookupSelector: aMessage selector) ifNotNil:
-		["Simulated lookup succeeded -- resend the message."
-		^ aMessage sentTo: self].
+	((Behavior classOf: self) lookupSelector: aMessage selector) ifNotNil:
```
guillep, line 78:
```suggestion
	((MirrorPrimitives classOf: self) lookupSelector: aMessage selector) ifNotNil:
```

src/Kernel/ProtoObject.class.st
```diff
@@ -67,11 +67,13 @@ ProtoObject >> becomeForward: otherObject copyHash: copyHash [
...
+		[ :method | "Simulated lookup succeeded. Evaluate the method!"
+		^ method valueWithReceiver: self arguments: aMessage arguments ].
```
guillep: I need indentation to not bleed, sorry
```suggestion
			^ method valueWithReceiver: self arguments: aMessage arguments ].
```

guillep (changes requested):
Some little comments with committable suggestions.

Tests look good to me :)

About the `cannotReturn:`, why not just testing it with an exception?

```smalltalk
block := self blockWithNonLocalReturn.
self should: [ block value ] raise: CannotReturn.

blockWithNonLocalReturn
  ^ [ ^ self "Should throw always a cannot return exception because the home will be always dead" ]
```

The about to return is probably a bit trickier to test, because it's called when unwinding the stack because of non-local returns. And it's implemented to look for the next method marked as "interested in the unwind", which boils down to ensure blocks (methods marked with primitive 198). Then the hook make sure of calling  the block of the ensure.

However, if you carefully craft a context, you could set your own home context of your own class, and hook the `return:through:` below?

```smalltalk
aboutToReturn: result through: firstUnwindContext

	"Called from VM when an unwindBlock is found between self and its home.
	 Return to home's sender, executing unwind blocks on the way."

	self home return: result through: firstUnwindContext
```

src/Kernel/Behavior.class.st
privat: The class is not in Kernel, so it introduces an unwanted package dependency with ReflectionMirrors-Primitives.
Kernel should try to be minimal and self-contained. Some small code duplication is a lesser concern.

Moreover, MirrorPrimitives is undocumented and almost not used (only 4 references in the whole image outside of its own package `ReflectionMirrors-Primitives`, and those 4 are only for tests), its methods are also undocumented and non-discoverable.


The version in Behavior is not that bad, as the class is meaningful for the intent.

privat:
> About the `cannotReturn:`, why not just testing it with an exception? [...]

I have a work in progress branch with things more or less equivalent with your proposals.
Testing callbacks on real context objects is ok.

I also want to check directly the behaviour of the VM by instrumenting, diverting and hacking the receiver of these callbacks. I cannot really do that with the image implementation of Context.
But catching the callbacks on a made up context-ish class is not easy (and very low-level to debug).
Currently, I have some weird behavior that could be some programmer (me) errors (likely) or some VM limitations (possible).

Anyway, I plan to make a new PR special-methods-for-Context once this one is merged

jecisc:
What is the state of this PR?

Tests are green. Some changes were requested but Jean seems to think the changes requested are not needed.

Can we integrate this change?

PR #14215, opened 2023-07-05
This PR removes dead code related to VTerm output classes

### LINKED ISSUE #5849: Missing implementation for VTermOutputStyleElement #print:withColor:
the method #print:withColor: has sender - but no implementation

MarcusDenker:
- remove method on VTermOutputStyleElement that sent non-existing selectors
- remove #setInDriver: which has no senders, but on one method it sends non-existing selector


jecisc (approved):
VTermOutput seems to need some love but when I asked about it only few people seems to know what is supposed to stay, what is supposed to be removed...

PR #14217, opened 2023-07-05
Rename #allManagers into #allWorkingCopies

jecisc:
Rename MCWrokingCopy>>#allManagers to #allWorkingCopies because it returns all the working copies and I don't see any reason to call them "managers".

After this PR I'll do a pass over the API to finish remove all "managers" and use "working copies" instead.


jecisc:
Failure does not seems to be related

Issue #14221, opened 2023-07-05
ClassInstaller: Improve installation in non default environment


jecisc:
I tried to use the class installer in order to install a class in a non-default environment in the image but that does not seem to be easy. 

In the end I ended up with this code:

```st
	| newEnvironment |
	newEnvironment := SystemDictionary new.
	newClass := ShiftClassInstaller new
		            installingEnvironment: (ShSmalltalkGlobalsEnvironment new
				             environment: newEnvironment;
				             yourself);
		            make: [ :builder |
			            builder
				            name: #SHClassForTest;
				            superclass: Object;
				            category: self generatedClassesPackageName ].
```

I am wondering, can't we do easier? Do we really need ShSmalltalkGlobalsEnvironment? Can't we just add the methods it has on SystemDictionary and give that as an environment directly? Do we need this intermediate class? If yes, can't we add some syntactic suggar to make this easier?

My goal is to be able to have tests that generates classes and that are not installed in the default system and that seems to be painful to do.

jecisc:
On top of this, the class gets added to the wrong package organizer because it is added through the SystemOrganizer and the system organizer is adding everything to the default package organizer of the image...

This will be a mess really difficult to untangle.

privat:
I tried to look at that when I played with the compiler, and also when I played with undeclared variables. A lot of things assume a default environment, other things offers complex, fragile or untested API and lots of unanswered questions.

Related to #13563 maybe...

jecisc:
The problem is that we currently have 2 "package management" in the image. 

One is the historical "SystemOrganizer" (the class is not there anymore, I merged the content in RPackageOrganizer) and the second is RPackageOrganizer. 

When we install a class in the system, it gets install in the SystemOrganizer and through announcements the organizers are trying to do their thing. But we only have one SystemOrganizer in the image even if we can have multiple RPackageOrganizer. Because of this, we have some hacks in RPackage but I'm trying to remove them because they are making the code incredibly complexe for nothing. 

I'm trying to merge the SystemOrganizer into RPackageOrganizer so that we only have one package manager and not two, but this is not an easy task because RPackageOrganizer is too complex for what it does, it is full of bugs, it is missing features and on top of that, some changes do not pass because Espell assume things that I want to change and Espell is currently in a state that makes it hard to modify it. 

So I'm trying to find my way around all those complications ahah

jecisc:
Done!

PR #14222, opened 2023-07-05
Add test on installing a class in a non default environment

jecisc:
Adding a test about adding a class in a non default environment because I didn't see this done in the image and it made it harder for me to find how to do that.

The problem is that we currently can install a package in another environment but the package organizer it ends up added to is the wrong one! It is always the one of the system and not the one we ask :( For now I added assertions that I skipped but it is important we are able to make them pass at some point if we want to be able to have tests generating code in the non default environment to not dirty the image.

I also find the installation in a non default environment too difficult but I opened https://github.com/pharo-project/pharo/issues/14221 for this

### DIFF
--- a/src/Shift-ClassInstaller-Tests/ShClassInstallerTest.class.st
+++ b/src/Shift-ClassInstaller-Tests/ShClassInstallerTest.class.st
@@ -52,7 +52,6 @@ ShClassInstallerTest >> newClass: className superclass: aSuperclass slots: slots
 ShClassInstallerTest >> tearDown [
 
 	(self generatedClassesPackageName asPackageIfAbsent: [ nil ]) ifNotNil: [ :package | package removeFromSystem ].
-
 	super tearDown
 ]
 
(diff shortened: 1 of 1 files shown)


src/Shift-ClassInstaller-Tests/ShClassInstallerTest.class.st
```diff
@@ -288,6 +287,30 @@ ShClassInstallerTest >> testDuplicatedInstanceVariableInSuperclassMetaclass [
...
+		            make: [ :builder |
+			            builder
+				            name: #SHClassForTest;
+				            superclass: Object;
+				            category: self generatedClassesPackageName ].
+
+	self class environment at: #SHClassForTest ifPresent: [ self fail: 'The default environment should not have the class installed.' ].
+	newEnvironment at: #SHClassForTest ifAbsent: [ self fail: 'The default environment should not have the class installed.' ].
+	
+	self skip.
```
privat, line 308: Expect the failure instead of skip, so if is fixed on day it will be easier to remember to update the test.

privat (approved):

src/Shift-ClassInstaller-Tests/ShClassInstallerTest.class.st
jecisc: I'm currently working on RPackage and I hope I'll be able to fix this before Pharo 12 release. So I added a flag for me to find this test in the next few months

src/Shift-ClassInstaller-Tests/ShClassInstallerTest.class.st
```diff
@@ -288,6 +287,30 @@ ShClassInstallerTest >> testDuplicatedInstanceVariableInSuperclassMetaclass [
 	self deny: ((superClass class slots collect: [:e | e name]) includes: 'aSlot')
 ]
 
+{ #category : #tests }
+ShClassInstallerTest >> testInstallInSpecificEnvironment [
+
+	| newEnvironment |
+	newEnvironment := SystemDictionary new.
```
astares, line 294: I would change it into

```Smalltalk
| newEnvironment |
newEnvironment := self class environment class new.
...
```
to be more generic and **not hardcode that we use "SystemDictionary" at the moment**

src/Shift-ClassInstaller-Tests/ShClassInstallerTest.class.st
```diff
@@ -288,6 +287,30 @@ ShClassInstallerTest >> testDuplicatedInstanceVariableInSuperclassMetaclass [
...
+	newClass := ShiftClassInstaller new
+		            installingEnvironment: (ShSmalltalkGlobalsEnvironment new
```
astares, line 296: If this globals environment is a "wrapper" for another environment as class comment in #ShGlobalsEnvironment says, then why not have a 

```Smalltalk
ShSmalltalkGlobalsEnvironment wrapping: newEnvironment
``` 

**instance creation method** #wrapping: in #ShGlobalsEnvironment class like

```Smalltalk
wrapping: anEnvironment

	^ (self new)
			environment: anEnvironment;
			yourself
```
Would make the code easier

PR #14224, opened 2023-07-05
Clean PackageOrganizer methods in testing protocol

jecisc:
Here is a cleaning of the testing protocol of RPackageOrganizer.

- #hasRegistered has no sender and is removed
- #includesPackage: and #includesPackageNamed: are deprecated
- #hasPackage: has been introduce

The renaming is due to the fact that most other parts of the MM are using #hasX: instead of #includesX: and I want to unify things. 

I did other cleanings for example on refactoring preconditions. If I have the courage I'll push more things to this PR such as a test for this method or other simplifications in refactorings related to this.


jecisc:
I added the test. For the other refactoring simplifications I'll need to talk to Steph and Balsa after this is integrated because I am not sure how something is managed.

jecisc:
The error is weird with this PR. I'll try to split it in smaller PRs.

Ducasse:
Build is broken.

jecisc:
I think this PR should now pass. Let's see what the CI says

MarcusDenker (approved):

PR #14225, opened 2023-07-06
Dependency Analyser title management

jordanmontt:
Refactored Dependency Analyser title management. The `title` method will be deprecated. Now one needs to use the `initializeWindow:` method instead.

### DIFF
--- a/src/Tool-DependencyAnalyser-UI/DAAddPackagePresenter.class.st
+++ b/src/Tool-DependencyAnalyser-UI/DAAddPackagePresenter.class.st
@@ -82,6 +82,12 @@ DAAddPackagePresenter >> initializePresenters [
 ]
 
 { #category : #initialization }
+DAAddPackagePresenter >> initializeWindow: aWindowPresenter [
+
+	aWindowPresenter title: 'Choose packages to add'
+]
+
+{ #category : #initialization }
 DAAddPackagePresenter >> selectedItemsFromPackageList [
 	^ packageList chosenItems
 ]
@@ -90,9 +96,4 @@ DAAddPackagePresenter >> selectedItemsFromPackageList [
 DAAddPackagePresenter >> systemPackages [
 
 	^ (self packageOrganizer packages collect: [ :package | package packageName asString ]) asSortedCollection
-]
-
-{ #category : #protocol }
-DAAddPackagePresenter >> title [
-	^ 'Choose packages to add'
 ]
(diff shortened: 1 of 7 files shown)


astares:
@jordanmontt / @estebanlm 

Can we extract the title String into a "self defaultWindowTitle" or similar with #defaultWindowTitle method being in "accessing - defaults" category? 

The #initializeWindow: then could be more generally implemented in a common superclass (of own package or even presolved in Spec2 itself. 

This makes it easier for subclasses to just override the #defaultWindowTitle and provide the title string, no?

jordanmontt:
Hello @astares 

I just checked and in fact one can define the title in the class side of the presenter. So, one is not obliged to redefine the `initializaWindow:` for only setting the title. I changed this PR to use the class side method `title` instead.

astares:
@jordanmontt While it may look like a solution in the first place and works now it does not seem to be the long term solution. I fear this mechanism at the moment only works for compatibility reason. If you check what I guess @estebanlm has written in SpPresenter>>#title 

image

I know this is instance side #title and you are now using the class side #title.

But I think the whole mechanism makes use of the #SpPresenter>>#initializeWindow: as default mechanism still. But this method says:

image

so it will become obsolete and people should provide an own, custom initializeWindow: implementation nonetheless
(where your PR now does not provide one).

It is a little but unclear to me what the mid-term/long term design decision from @estebanlm is - if every subclass of SpPresenter would override and implement its own #initializeWindow: just to provide the title (which is really a common case at least for presenters reflecting a tool window) I would see this a little overkill.

Ideally from my POV we:
A) have a predefined behavior in SpPresenter>>#initializeWindow: that uses more explicit methods like #defaultWindowTitle or #initialExtent and #defaultWindowIcon
B) or (if we do not want to have this for presenters that do not represent a window we could have an "SpWindowPresenter" or "SpRootPresenter" that defines that. 

From the perspective of making no differences between presenters and the universal combination idea of presenters in Spec2 the specific superclass as in B might not make sense (as even window presenters could be reused to be child presenters later additionally). 

So the PR is fine and working - but the question to me is: is it the way it should be to be future proof when SpPresenter>>#initializeWindow: does not set the title later on anymore.

Ducasse:
Let us ask esteban :)

estebanlm:
What is very bad of the `title` method is that it has a whole holder keeping a value that will or will not be used since a presenter is not always meant to become a window (it is a reusable component, isn't?). Moreover, all the window management methods were creating a coupling between the window mechanism (and in particular with morphic), and title was just another way. 
I added the initializeWindow: method with the idea that users could override it and define their own desired behavior there... yes, we could have a `defaultWindowTitle` method to easy this definition (not implemented yet). And users could override it to add a title without needing to override `initializeWindow:`
I am still uncertain about which path to take. 
I am only sure that having things about a window directly inside a presenter is bad :)
Anyway,   putting title in class side and override `initializeWindow:` to use it is a perfectly valid solution ;)

estebanlm (approved):

PR #14227, opened 2023-07-06
14226-nil-cannot-be-pinned-in-memory

### LINKED ISSUE #14226: nil cannot be pinned in memory
Pinning nil leaves the VM in an invalid state.
The same is for true and false.
It should be avoided in the image

tesonep:
Fix #14226
Pinning nil, true or false leaves the VM in an invalid state

### DIFF
--- a/src/Kernel/Boolean.class.st
+++ b/src/Kernel/Boolean.class.st
@@ -188,6 +188,12 @@ Boolean >> or: alternativeBlock [
 	self subclassResponsibility
 ]
 
+{ #category : #pinning }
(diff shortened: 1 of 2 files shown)


privat:
This is weird. Should not pinning those object be a no-op? as they are unmovable in spur.
Do you have more details because now I'm intrigued.

Also, if pinning them cause an invalid VM state, it seems to be a VM issue, not an image issue.

Ducasse:
@tesonep @guillep I like the remark of jean.

Ducasse:
Ok I got it so I will integrate it first and we will see how this can be fixed from the VM side.

PR #14228, opened 2023-07-06
Added comments to the new instrumentation package

jordanmontt:
This PR adds comments as a documentation for the new `Instrumentation` package. @guillep and @BastouP411 can you check that the comments are all right?


Mathilde411:
> This PR adds comments as a documentation for the new `Instrumentation` package. @guillep and @BastouP411 can you check that the comments are all right?

Yeah everything seems to be okay :)

guillep (approved):
Thanks @jordanmontt and @BastouP411 for the review!

PR #14233, opened 2023-07-06
Splitting preconditions

Ducasse:
Started to identify the kind of conditions mainly applicability

### DIFF
--- a/src/Refactoring2-Transformations/RBRemoveParameterTransformation.class.st
+++ b/src/Refactoring2-Transformations/RBRemoveParameterTransformation.class.st
@@ -65,17 +65,29 @@ RBRemoveParameterTransformation >> hasReferencesToTemporaryIn: each [
 ]
 
 { #category : #private }
-RBRemoveParameterTransformation >> myConditions [
+RBRemoveParameterTransformation >> localPreconditions [
 
 	| imps |
 	self getNewSelector.
-	imps := self model allImplementorsOf: oldSelector.	
+	imps := self model allImplementorsOf: oldSelector.
 	^ imps inject: (RBCondition definesSelector: oldSelector in: class)
 		into:
 			[:cond :each |
 			cond & (RBCondition 
 						withBlock: [ (self hasReferencesToTemporaryIn: each) not ]
 						errorString: 'This argument is still referenced in at least one implementor!!')]
+]
+
+{ #category : #private }
+RBRemoveParameterTransformation >> myConditions [
+
+	^ RBCondition true
+]
+
+{ #category : #private }
+RBRemoveParameterTransformation >> preconditions [ 
+	
+	^ super preconditions & self localPreconditions 
 ]
 
 { #category : #private }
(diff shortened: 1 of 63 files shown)


Ducasse:
Strange that they work locally and break the build.

src/Refactoring2-Transformations/RBRemoveParameterTransformation.class.st
```diff
@@ -65,14 +66,15 @@ RBRemoveParameterTransformation >> hasReferencesToTemporaryIn: each [
 
 { #category : #private }
 RBRemoveParameterTransformation >> myConditions [
+
 	| imps |
-	imps := self model allImplementorsOf: oldSelector.
 	self getNewSelector.
```
balsa-sarenac, line 70: Would it make sense to move this to `interactionCondition`?

src/Refactoring2-Transformations/RBTemporaryToInstanceVariableTransformation.class.st
```diff
@@ -100,6 +100,22 @@ RBTemporaryToInstanceVariableTransformation class >> model: aRBSmalltalk class:
...
+RBTemporaryToInstanceVariableTransformation >> applicabilityConditions [ 
+
+	^(RBCondition definesSelector: selector in: class)
+		& (RBCondition definesInstanceVariable: temporaryVariableName asString in: class) not
+		& (RBCondition withBlock: [
+			(class allSubclasses anySatisfy:
+				[ :cls | cls definesInstanceVariable: temporaryVariableName asString ])
+			ifTrue: [ self refactoringWarning:
+				('One or more subclasses of <1p> already defines an<n>instance variable with the same name. Proceed anyway?' expandMacrosWith: class name) ].
+			true ])
```
balsa-sarenac: I think this can be qualified as a breaking change?

balsa-sarenac (commented):
I was catching up with latest changes and while reading the code I left two minor comments.
`interatcionCondition` is nice flag for stuff that should go to driver.
I also like `self flag: #popUp` I'll add those as well so we know exact number of required changes.

src/Refactoring2-Transformations/RBRemoveParameterTransformation.class.st
Ducasse: I do not know :)

src/Refactoring2-Transformations/RBTemporaryToInstanceVariableTransformation.class.st
Ducasse: Yes!

Ducasse: In fact we cannot have a class with an instance variable with the same name it will break the system.

Ducasse: So I do not understand the proceed anyway

Issue #14197, opened 2023-07-03
Object>>asOrderedCollection automatic deprecation transform causes wrong code


jbrichau:
**Bug description**
The deprecation and automatic rewrite of `Object>>asOrderedCollection` causes an issue when there is code in the image where the receiver can be both a `Collection` (subtype) and any other `Object` subtype.

The automatic rewrite is only correct if the `@receiver` of `asOrderedCollection` is never a `Collection`. However, that cannot be guaranteed statically. Since this rewrite is applied by default, you can run into strange situations when ran on a CI server before you understand what is happening.

**To Reproduce**
Implement the following testcase and run it twice while having the setting "Performs automatic transformations on deprecated methods" set to true. The first testrun will succeed and rewrite the code. The second testrun will fail.

```
testAutomaticRewrite

	| objects resultObjects |
	objects := { Object new . OrderedCollection new }.
	resultObjects := objects collect:[ :o | o asOrderedCollection ].
	self assert: resultObjects second isEmpty.
```

**Expected behavior**
While this obviously is a mistake in the code, the problem with automatic rewrite of this rule is that the resulting code is wrong. I expect it not be rewritten automatically.

**Version information:**
 Pharo Version 11

**Expected development cost**
I recommend removing the `transformWith:` in this deprecation declaration. It is not a safe transformation to do and I do not see a way to express a condition either.

**Additional context**
Because Pharo sets the transformation to true by default, the issue arises automatically in SmalltalkCI builds. We have now turned that off as well, but it was the first time we noticed this.

Ducasse:
indeed!

Ducasse:
So we should remove it because we cannot express not a collection. Or we could do a runtime check.
But what is a collection? Because not all the collection inherits from collection.

jbrichau:
Fixed in https://github.com/pharo-project/pharo/pull/14237
Thx!

jbrichau:
This is still present in Pharo12, reopening to also remove it there.

Ducasse:
Sorry so what should we do. Personally I would remove asOrderedCollection from Object. This is a bad idea to have it there.

jbrichau:
The deprecation is fine. It is the automated transform that is not correct (and impossible to do imho, because the method has different polymorphic implementations). An automated transform should always be correct, otherwise it should simply be deprecated and let a human handle it.

Ducasse:
ok I will change it

Ducasse:
[...]

PR #14237, opened 2023-07-07
Disable automated transform for deprecation of Object>>asOrderedCollection

jbrichau:
See #14197 for the reasoning.

I opted to comment out the transform. The method and comment will disappear in Pharo 12 anway.


astares (approved):

Ducasse:
Tx

PR #14245, opened 2023-07-09
Refactoring: reuse `RBAddMethodTransformRefactoring`

balsa-sarenac:
Reused `RBAddMethodTransformRefactoring` where possible.
Renamed `performCompositeRefactoring:` to `generateCompositeChangesFor` to better reflect its behaviour.


Ducasse:
In fact I would rename generateCompositeChangesFor: in generateChangesFor: because the composite for me implies that there is a notion of composition that we do not have here.

PR #14246, opened 2023-07-10
Deprecate #includesPackageNamed:

jecisc:
And use hasPackage:instead. 

Subpart of #14224


jecisc:
I don't understand the error I'm getting :(

jecisc:
I think the problem comes from the fact that the rewrite is been called by iceberg while loading iceberg. 

So we need to merge this first: https://github.com/pharo-vcs/iceberg/pull/1710

jecisc:
Test passed now! This was indeed because of the Iceberg rewrite :)

MarcusDenker (approved):

PR #14247, opened 2023-07-10
Rename some more tags into protocol

jecisc:
Rename some more methods, classes and variables using the tag vocabulary to use the protocol vocabulary


jecisc:
Failing tests seems unrelated

MarcusDenker (approved):

PR #14248, opened 2023-07-10
Clean MCWorkingCopy

jecisc:
This changes finish 3 work that were started on Monticello recently.

The first is to use protocols instead of protocol names were it could simplify the code.

The second was to rename "manager" into "workingCopy" in the code to unify things and make the code more understandable. 

The third was to reduce the "category" dependency of RPackage since we want to remove this in the future. 

The cherry on the cake is that I simplified a lot some parts of the impacted code. Especially #managersForCategory:do: (now called #workingCopiesForExtensionProtocol:do:). And this was already simplified in a recent PR! If you check the code in Pharo 11 and the code now, things are much simpler.

### DIFF
--- a/src/Manifest-Core/ClassDescription.extension.st
+++ b/src/Manifest-Core/ClassDescription.extension.st
@@ -27,5 +27,6 @@ ClassDescription >> manifestBuilderForRuleChecker: aRuleChecker [
 
 { #category : #'*Manifest-Core' }
 ClassDescription >> mcWorkingCopy [
-	MCWorkingCopy managersForClass:  self do: [: package | ^ package ]
+
+	MCWorkingCopy workingCopiesForClass: self do: [ :package | ^ package ]
 ]
--- a/src/Manifest-Core/CompiledMethod.extension.st
+++ b/src/Manifest-Core/CompiledMethod.extension.st
@@ -36,5 +36,6 @@ CompiledMethod >> manifestBuilderForRuleChecker: aRuleChecker [
 
 { #category : #'*Manifest-Core' }
 CompiledMethod >> mcWorkingCopy [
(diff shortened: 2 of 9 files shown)


MarcusDenker (approved):

MarcusDenker:
The only problem is that it touched Zinc, which wants to use the same code in as an old Pharo as possible. I wonder how to deal with that in general

jecisc:
@svenvc 

Hi Sven!
I would like to have your opinion on this change because it touches Zinc in two places.

The first one is an automatic deprecation that was applied while I was doing this change.
The second is a method that was renamed because the name was pretty bad (it was calling MCWorkingCopy "manager" which made the code more complex than it should have). But I see that the method (#frameworkMCVersion) has no sender!

So here are my two questions:

1) Can I commit the automatic rewrite of ZnUtils class >> signalProgress:total:?
2) Can I remove the statically dead #frameworkMCVersion?

svenvc:
Of course we can remove the dead method

But I don't see the change in ZnUtils class >> signalProgress:total: 

It looks just reformatting to me, but I am probably looking over it

jecisc:
humanReadableSIByteSize/humanReadableSISizeString

Subtle change. After I saw that this method got changed, I had to read it 5 times to see why it happened ahah

svenvc:
Hmm, OK.

That is of course a problem: if I do that, Zinc will no longer work on Pharo 7, 8, 9, 10, 11 and that just for a renaming that adds very little value.

You can do ahead and make the change in Pharo 12, we'll see.

jecisc:
I'll update the PR (probably tomorrow?) to remove the dead method and I'll revert the automatic rewrite for now since it is not related to the PR and was just a side effect.

PR #14249, opened 2023-07-10
Update bootstrap scripts

jecisc:
This changes update some API in the bootstrap scripts to not use deprecated calls. 

But I see in multiple of those scripts some code that I think cannot work in the current Pharo image. I guess we never use them. I wonder what we should do? Remove them since we do not use them and they are not working anymore? Keep them? But in that case it would be good that the CI runs them to ensure they are up to date..


src/Pharo30Bootstrap/PBMethodCompiler.class.st
```diff
@@ -13,9 +13,9 @@ PBMethodCompiler >> buildPackagesList [
 	| classes |
 	classes := builder behaviorDefinitions.
 	^((Smalltalk at: #MCWorkingCopy ifAbsent: [ ^#() ] )
-		allManagers
-			select: [ :m | classes anySatisfy: [ :c | m packageInfo includesClassNamed: c name ] ]
-			thenCollect: [ :m | m packageInfo packageName asUppercase ]) asSet asOrderedCollection.
+		allWorkingCopies
+			select: [ :wc | classes anySatisfy: [ :c | wc packageInfo includesClassNamed: c name ] ]
```
jecisc: #packageInfo does not exists on MCWorkingCopy for example in Pharo 12

MarcusDenker (approved):

MarcusDenker:
there is now a conflict

PR #14253, opened 2023-07-10
DeadCodeCleanup-07-10

MarcusDenker:
This PR removes some methods that are not sent, but that do send selectors that do not exist


jecisc:
It's sad that we are losing fast table examples. 

Once the ones with dead code are removed I'll add a smoke test to check that they do not get broken in the future.

jecisc (approved):

PR #14256, opened 2023-07-10
Fix bug in tag creation while creating a category

jecisc:
Still in my effort to kill the categories, I'm doing this PR.

There is an inconsistancy while creating a category and this cause problems in the package model. Before killing the categories, it would be nice that the system behave in the correct way. 

Here is the bug: If we create a category that has no dash in the name, then a package of the same name is created BUT a tag of the same name is also created in the package which is wrong.

I updated the mecanism to not create the tag and added a test about that. All RPackage tests are green but I expect failures in other part of the system. Let's see


jecisc:
Surprisingly, all tests are green :)

This will fix a bug that sometimes appears where the package name end up duplicated as a tag (For example if you create a class in "package-tag" it can later be displayed as "package-tag-package-tag"!)

MarcusDenker (approved):

PR #14257, opened 2023-07-10
Do not merge: System announcer test

jecisc:
In some PR I noticed some weird behavior related to the bootstrap and the system announcer. 

Here is a PR with some logs to see what is happening in a PR that is not failing to find the difference with the other PRs.

I have the impression that at some point we reset the system announcer, it register the announcements via a pragma, but the pragma cache is not good for all methods.

Since I'm not sure, let's see what the logs are saying.


jecisc:
My previous comment was wrong. 

In this version the pragma collection during the reset of the system announcer in the bootstrap collects 4 methods while in the failing PR it collects 2 methods. 

I have no idea what is happening for this to happen.

jecisc:
Not needed anymore, I found the error :)

PR #14258, opened 2023-07-10
Remove dead code in bootstrap

jecisc:
The bootstrap seems to have dead code because it is calling methods that were removed some years ago from Pharo.  Here is a PR to try to remove some of this dead code. Let's see if that works.


src/Pharo30Bootstrap/PBImageBuilder50.class.st
```diff
@@ -357,6 +357,21 @@ PBImageBuilder50 >> createInitialObjects [
 	self checkpoint: 'created_initial_objects'
 ]
 
+{ #category : #'method-compiling' }
+PBImageBuilder50 >> createNewMethod: aMethodDefinition [
+	"	| ownerClass |
+""	ownerClass := self behaviorFrom: aMethodDefinition methodClass instanceSide."
```
privat: ""??

privat (approved):

PR #14259, opened 2023-07-10
Treat Integers and Fractions as having scale 0 in ScaledDecimal arithmetic (fixes #14238)

### LINKED ISSUE #14238: Multiplication with Integer and ScaledDecimal operands has different scale depending on argument order
**Bug description**
Multiplying an integer and ScaledDecimal produces a result with a different scale value depending on argument order:

```
1.12s * 2 "=> 2.24s, i.e. scale = 2"
2 * 1.12s "=> 2.2400s, i.e. scale = 4"
```

While the two results still compare as equal, this is definitely pretty weird for a commutative operation, and it seems wrong in general for multiplication by an integer to increase the scale.

**Expected behavior**
Keep the scale from the ScaledDecimal operand.

**Version information:**
 - OS: Any
 - Version: Any
 - Pharo Version: Any AFAIK, definitely true in 11

**Expected development cost**
I think all that is needed is to coerce Integer to ScaledDecimal with scale 0 instead of that of the ScaledDecimal operand. For other operations, the max of the two is used instead of the sum, so this shouldn't change the result. Pull request inbound shortly.

daniels220:
This takes the approach of treating all non-ScaledDecimals as having an effective scale of 0 when participating in arithmetic with a ScaledDecimal. The underlying precision of the result is always retained in full, but the declared scale will come only from the ScaledDecimal operand. This is already the case when the ScaledDecimal itself is the receiver/left operand, so any other approach would require considerably more thought and broader changes.

### DIFF
--- a/src/Kernel-Tests/ScaledDecimalTest.class.st
+++ b/src/Kernel-Tests/ScaledDecimalTest.class.st
@@ -232,7 +232,22 @@ ScaledDecimalTest >> testLiteral [
 ScaledDecimalTest >> testMultiplicationDoesNotLoosePrecision [
 	"Check that not only the largest scale is considered but the sum of the two scales. See issue #8668"
 
-	self assert: 1.003s3 * 1.006s4 equals: 1.0090180s7
+	self assert: 1.003s3 * 1.006s4 equals: 1.0090180s7.
+	self assert: (1.003s3 * 1.006s4) scale equals: 7.
+]
+
+{ #category : #tests }
+ScaledDecimalTest >> testMultiplicationWithNonScaledDecimalDoesNotGainPrecision [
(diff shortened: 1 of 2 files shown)


daniels220:
Does the "Some checks were not successful" mean I need to do something? Looks like the Windows build failed, in a way that strikes me as probably unrelated to my actual changes—is this known-unreliable?

jecisc:
Our windows CI slave is not feeling well these days. Don't worry about it, the tests passed and I added a label so that we know it is ready for review.

daniels220:
Cool, yeah, that's what I thought based on the label. Thanks!

guillep (approved):
LGTM!

Thanks!

PR #14264, opened 2023-07-11
Remove unclassify hack

jecisc:
One thing that I hated in the protocol management is that classfying with `Protocol unclassified` is used to say "Do not update my protocol". But this leads to really bad behavior such as:

```st
MyClass compile: 'foo ^ 1' classified: #toto.
MyClass compile: 'foo ^ 1' classified: #'as yet unclassified'.
(MyClass>>#foo) protocolName "#toto"
```

This means that if you want to unclassify your method you need to deal with the internal behavior of  ClassDescription.

I propose to use 'nil' as a way to "not update" the protocol and now Protocol unclassified works just like any other protocol! No more hack and if everywhere. This is the result of all the previous cleanings also :)

In order to do this I had to:
- update ClassDescription>>#addAndClassifySelector:withMethod:inProtocol: to remove the conditions
- update ClassDescription>>#compile:notifying: to just use nil as default value
- update ClassDescription>>#classify:under: to use nil as a "do not update protocol" 
- update OpalCompiler>>#install to not cast nil into a Protocol unclassified


jecisc:
Depends on the fix in https://github.com/pharo-project/pharo/pull/14273

jecisc:
This PR is FINALLY green!

MarcusDenker (approved):

PR #14266, opened 2023-07-11
Rename more methods and classes to use "Protocol" vocabulary

jecisc:
Still the same as multiple recent PR. And I still have some more to do (maybe I'll add them to this PR but I need to go home now)


jecisc:
I'll fix the conflict and add some more changes soon

jecisc:
With the last commit I think I removed almost all the "tags" to talk about protocols. I think we can still simplify the code in some places and I have the impression that we still have some bugs around that. But it is already cleaner and I was able to fix multiple bugs during this simplification

PR #14267, opened 2023-07-11
Fix: update changes preview table when changing active scope

balsa-sarenac:
When changing active scope refactoring preview presenter didn't update preview for changes that are going to be applied so it always displayed changes for the whole image. Note that it worked okay just the UI was wrong.


jecisc (approved):

jecisc:
The CI seems to have failed randomly but I cannot relaunch it

MarcusDenker:
I restarted the CI run

PR #14268, opened 2023-07-11
Bugfix: Fix double protocol request in Calypso

jecisc:
When we select methods in Calypso, including extension methods, we can apply the "Move to class defining method" refactoring. In that case, we get asked two times the protocol of the methods to update.

This change fixes this by asking only once by method and ignoring methods that are already defined in their class. I also introduced CompiledMethod>>#isExtensionMethod


Ducasse:
Thanks.

MarcusDenker (approved):

PR #14274, opened 2023-07-12
[Pharo11] This is a backport: 14034 debugger does not evaluate in the context of the receiver

MarcusDenker:
This is a backport: 14034 debugger does not evaluate in the context of the receiver


jecisc:
We have 6 failures but they all seems to be known random failure + two that seems to be due to a unix slave network hiccup

jecisc (approved):

PR #14281, opened 2023-07-12
Simplify package removal

jecisc:
When we remove a method from the system, if it was the last of the protocol, then the protocol get removed. This means that RPackage does not have to take this responsibility


jecisc:
Test failures seems to be due to network hiccups on the unix slave.

MarcusDenker (approved):

Issue #14141, opened 2023-06-28
BlockClosure>>#isClean regression from Pharo 10 to 11


koendehondt:
**Bug description**
`BlockClosure>>#isClean` does not behave the same on Pharo 11 as it did on Pharo 10.

**To Reproduce**

The problem was found in a much larger block, but we managed to strip it down to the bare essentials to reproduce the bug.

In Pharo 10, this holds:
```smalltalk
[ :x | x someMessageTakingABlock: [ x + 1 ] ] isClean >>> true
```

In Pharo 11, it is different:
```smalltalk
[ :x | x someMessageTakingABlock: [ x + 1 ] ] isClean >>> false
```

**Expected behavior**
I expect the same behaviour in Pharo 10 and Pharo 11.

**Screenshots**
N/A

**Version information:**
 - OS: macOS
 - Version: 13.4
 - Pharo Version: 11

**Expected development cost**
I do not know what the development cost would be, as I am not familiar with the code.
I will be glad to help with testing the fix, although the given example can be turned into a unit test easily.

**Additional context**
None.

MarcusDenker:
Interesting, I will check

Ducasse:
Tx for the report. Sorry for the inconvenience.

MarcusDenker:
I now understand the problem. 

We unified the #isClean to use the one that is defined on the AST, while in Pharo10 we have two implementations: one bytecode level, it just does local analysis of the bc of the block, ignoring embedded blocks.

In Pharo11, we use the AST version, which uses the AST and does a global analysis. Which might be too strict: it never denotes blocks as clean that contain non-clean blocks.

Imagine this:

```
[[ self]  ] isClean 
```
In Pharo10, BC analysis: block is clean, embedded block (the one accessing self) is not clean.

But if we would now compile and execute with clean blocks, the self is a problem: the outer ("clean") block would execute with nil receiver, so self would be nil for the embedded block, too, which would be wrong. Accessing "self" in an embedded block makes all the outer blocks non-clean, because we have to get the "self" from it.

But then, your example, when the internal block does not use self/super (or accesses ivars), I think in that case the outer block could indeed be denoted as clean.


So in Summary: 

- the BC level check in Pharo10 actually was *wrong*, 
- the AST level check that we use from Pharo11 is too conservative, it thinks any block that contains non-clean blocks is not clean, too, while we might be able to relax that notion in the cases the receiver is not needed.

MarcusDenker:
[...]

PR #14282, opened 2023-07-12
This fixes #isClean as described in #14141

MarcusDenker:
- #hasEscapingVars should only care about incomming copied vars
- add #isAccessingSelf, similar to #hasNonLocalReturn
- rewrite RBBlockNode>>#isClean in terms of checking for accessing self, return and escaping vars

- add the cases to BlockClosureTest>>#testIsClean

(we keep  #14141 open until this is backported to Pharo11)


jecisc (approved):

MarcusDenker:
I did some experiments
- looked at the difference (153 more clean blocks, look good)
-  recompiled the image with clean blocks enabled, works fine (did run some tests)
- I added the code as a .cs to the issue, for easy backporting

So ok to merge from my side.

Ducasse:
Do we merge it now or when you come back vacation?

MarcusDenker:
As we do not use clean blocks yet, It should have no real impact... but on the other hand, this means that it is ok to wait, too

PR #14283, opened 2023-07-12
Start unification of package removal

jecisc:
I found multiple ways to remove a package in the system. I would like to bring some unification to the package removal because depending on the method used, the cleanings done are not the same.

Here is a first step toward this unification. 
Two methods to remove a package were RPackageOrganizer>>#unregisterPackage: and #unregisterPackageNamed: 

This PR merges both into a method that has a name more coherent with the rest of the system: #removePackage: 

## Next steps 

Next step is to deprecate RPackage>>unregister and to merge the behavior of RPackage>>removeFromSystem and this new method introduced so that #removeFromSystem just has to be `self organizer removePackage: self`.


Ducasse:
Just a remark, could we after rename ensurePackage: and removePackage: as ensurePackageNamed: and removePackageNamed: because we will have first class package and this is super annoying to think we should pass a package and in reality this is string.

PR #14284, opened 2023-07-12
Remove #colorTransform: from BalloonEngine

jecisc:
BalloonEngine has a variable called #colorTransform: and some code managing it, but this code invoke a non existing method in the case the color is not nil. This means that the code cannot work if the variable is set.  I propose to remove it. 

WARNING: My image acts weird when I do this change but I would like to see how a bootstraped image act. Maybe this is just due to the migration of the instances in my image.


jecisc:
I'll close it because I don't know Balloon enough to fix this failure :(

PR #14285, opened 2023-07-13
Deprecate class reorganized

jecisc:
ClassReorganized is an announcement that is fired when we add a protocol, remove a protocol or rename a protocol. But this is useless since we have ProtocolAdded, ProtocolRemoved and ProtocolRenamed events!

I propose to remove it because it is really complex to make sur we do not fire it multiple times, for example when we classify a method that was the last one in its protocol inside a new protocol.

In this PR I made ProtocolAnnouncement fully polymorphic with ClassReorganized and I made "ClassReorganized" a deprecated alias of ProtocolAnnouncement. Like this, the change will not break any code until the deprecated code is removed! :)

This should fix https://github.com/pharo-project/pharo/pull/14264


jecisc:
CI did not start. I'll try to restart it

Ducasse:
Indeed the question of composite or event subsuming multiple more elementary one is painful.

MarcusDenker (approved):

PR #14289, opened 2023-07-13
Do not remove classes through announcements

jecisc:
RPackage currently uses announcements to sync with the system. Since the goal is to integrate it in the MM, this should be removed to let the MM directly sync RPackage. 

Here is a first step: Trying to remove the registration to ClassRemoved to remove a class for the packages.


jecisc:
We have failing tests... But at least the bootstrap did not break :)

jecisc:
I'll need some time to fix this one because I noticed that the tests are in a really bad state and the failures are related to the bad usage of the packages in the tests...

jecisc:
I'll redo it in another PR because the merge is too hard

PR #14291, opened 2023-07-13
Merge addition of class definition in rpackage

jecisc:
One of my goal is to make sure packages use real classes instead of class names. Here is a small step toward that goal: I merged #addClassDefinition: and #addClassDefinitionNamed:

Next step is to merge removeClassDefinition: and #removeClassDefinitionNamed: but this is not as easy since this is used by the system.


Ducasse:
I would like to see how the system will work when a class get obsolete or undeclared. This is for this reason that I used class names and not classes in Package.
Now we will see but I hope that the Obsolete or Undeclared will not break the system.

jecisc:
Marcus already gave me some pointer on thing to take care about and I’ll keep this in mind thanks :)

PR #14292, opened 2023-07-13
Remove brittle method in RPackage

jecisc:
RPackage contains a brittle method to add a selector to a class. Using this method can lead to a state breaking the integrity of a package.

Since it is use only in tests I propose to remove this and use the better API. (It is also brittle but less)


Ducasse:
Ok for me!

PR #14295, opened 2023-07-14
Improve CompiledMethod>>#isExtension

jecisc:
In my RPackage cleanings I want to remove all the dictionaries of names to manipulated the MM directly to get the right informations. This would simplify a lot the code and would remove all the bugs linked to synchronization. In order to do that without losing too much performance I want to speed up some operations first.

In this change I speed up CompiledMethod>>#isExtension and also I fixed a bug in it where every method that was removed from a class was answering true to #isExtension. Now it answers false since the method is unpackaged. I also added a tests to ensure this behavior.

Here are some bench:

```st
[ SystemNavigation default methods do: [ :m | m isExtension ] ] bench

"Before the change: 30 iterations in 5 seconds 13 milliseconds. 5.984 per second"

"After the change: 10 iterations in 5 seconds 436 milliseconds. 1.840 per second"
```

I also removed #isExtensionMethod that I introduced this week and that already had an equivalent.

### DIFF
--- a/src/Kernel-Tests-Extended/CompiledMethodTest.class.st
+++ b/src/Kernel-Tests-Extended/CompiledMethodTest.class.st
@@ -457,6 +457,24 @@ CompiledMethodTest >> testIsDeprecated [
 			ifFalse: [ self deny: (self class >> each) isDeprecated ] ]
 ]
 
+{ #category : #'tests - abstract' }
+CompiledMethodTest >> testIsExtension [
+
+	[
+	| method |
+	self class compile: 'isExtensionTestMethod ^ 1' classified: 'protocol'.
+	self deny: (self class >> #isExtensionTestMethod) isExtension.
+
+	self class compile: 'isExtensionTestMethod ^ 1' classified: '*Kernel-Tests-Generated-Package'.
+	method := self class >> #isExtensionTestMethod.
+	self assert: method isExtension.
+
+	self class removeSelector: #isExtensionTestMethod.
+	self deny: method isExtension ] ensure: [
+		(self packageOrganizer packageNamed: 'Kernel-Tests-Generated-Package' ifAbsent: [ nil ]) ifNotNil: [ :package | package removeFromSystem ].
+		self class compiledMethodAt: #isExtensionTestMethod ifPresent: [ :method | method removeFromSystem ] ]
+]
+
 { #category : #'tests - testing' }
 CompiledMethodTest >> testIsFaulty [
 	|  cm |
--- a/src/Kernel/CompiledMethod.class.st
+++ b/src/Kernel/CompiledMethod.class.st
@@ -438,6 +438,15 @@ CompiledMethod >> isExplicitlyRequired: marker [
 
 { #category : #testing }
 CompiledMethod >> isExtension [
+	"I return true if a method is an extension method. Which means that the methods is not packaged in the package of the class containing the method, but in another package.
+	
+	Note: In the future we want to go away from the usage of protocols to define an extension method. But in the meantime it is the fastest way to know if a method is an extension.
+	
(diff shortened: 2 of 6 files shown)


src/Kernel/CompiledMethod.class.st
```diff
@@ -437,9 +437,11 @@ CompiledMethod >> isExplicitlyRequired: marker [
 ]
 
 { #category : #testing }
-CompiledMethod >> isExtensionMethod [
+CompiledMethod >> isExtension [
 
-	^ self protocol isExtensionProtocol
+	^ self protocol
+		  ifNil: [ false ]
+		  ifNotNil: [ :protocol | protocol isExtensionProtocol ]
```
dionisiydk, line 453: It is shame that packaging is implemented in terms of a special protocol names. It should be a real thing independent from protocols. The original implementation was about this in mind. Now this PR pushes it back.
Imagine that one day the packaging system will be improved - you will need to change isExtension again. 

The "package API" is in place to not rely on protocols even if underhood it is still the foundation. If there are bugs they should be fixed. But any new place where extension protocols are exposed will complicate the future improvement to decouple protocols from packages.

jecisc: Hi Denis,
In the long run I also want to remove the fact that extensions rely on protocols. But currently, RPackage also rely on that everywhere.

My goal is to clean RPackage to make it way simpler because currently I’ve never seen code so brittle as RPackage and I think it would be a considerable effort to not rely on protocol. Once RPackage will be integrated in the MM and the complexity removed, I think it will be easier to remove the usage of extension protocol. But currently, to remove the complexity of RPackage we need to speed up the meta model. And lot of speed problems comes from RPackage itself and the way classes and methods are using it.

I can add a comment explaining that this is temporary and that in the future we want to remove the fact that protocols are used for extensions. But asking for packages to a class or method currently is too long because of the implementation and doing big changes to RPackage is just impossible due to its state.  

Would the comment I proposed to add be enough?

dionisiydk: I am happy that you have plan :)

jecisc: I added a comment with:
- A description of the method
- A explanation about the temporary usage of protocol and how it should be done in the future when RPackage will be faster
- Executable examples

I know it's not nice to use the protocol but it's too hard to clean RPackage and update the way extension methods work at the same time :(

src/Kernel-Tests-Extended/CompiledMethodTest.class.st
```diff
@@ -457,6 +457,24 @@ CompiledMethodTest >> testIsDeprecated [
 			ifFalse: [ self deny: (self class >> each) isDeprecated ] ]
 ]
 
+{ #category : #'tests - abstract' }
+CompiledMethodTest >> testIsExtension [
+
+	[
+	| method |
+	self class compile: 'isExtensionTestMethod ^ 1' classified: 'protocol'.
```
Ducasse, line 465: Hi cyril 

do you know if the compile: adds to the epicea log?
Because I would like to avoid this. 

S.

Ducasse:
> Here are some bench:
> 
> ```smalltalk
> [ SystemNavigation default methods do: [ :m | m isExtension ] ] bench
> 
> "Before the change: 30 iterations in 5 seconds 13 milliseconds. 5.984 per second"
> 
> "After the change: 10 iterations in 5 seconds 436 milliseconds. 1.840 per second"
> ```

Just for the record I got confused because you said it is faster but when I read it it feels the opposite. Not important.

src/Kernel-Tests-Extended/CompiledMethodTest.class.st
jecisc: For now it does but I want to disable that globally for tests

jecisc:
I inverted the after and before sorry :(

PR #14308, opened 2023-07-17
Remove unused package properties

jecisc:
RPackage has an unused class variable. This change removes it.  It had no getter so it is safe to remove.


jecisc:
I didn't notice but there is an accessor on the instance side.

PR #14309, opened 2023-07-17
Fix bug in SystemOrganizer synch

jecisc:
While trying to remove some users of #removeCategory: in the system, I noticed that in some cases, categories were not removed by #removePackage:

This adds a test and improve the category removal.


jecisc:
The failure seems to come from another synchronization problem.

In some tests I have a package that is empty but the system organizer still have classes in this category.

jecisc:
Done in another PR :)

PR #14312, opened 2023-07-17
Improve package cleaning API

jecisc:
The cleaning of the system is mostly done through the SystemOrganizer currently. In order to remove more users of the SystemOrganizer I want to improve the cleaning API of RPackage.

First this PR adds RPackage>>#removeEmptyTags with a test.
It also adds RPAckageOrganizer>>#removeEmptyPackagesAndTags with a test.

It deprecates #removeEmptyPackages and updates the senders for two reasons:
- The implementation was based on the SystemOrganizer and we want to get rid of it
- The name was not fitting since it let the user think we were removing only the packages and not the tags.

### DIFF
--- a/src/BaselineOfMorphic/BaselineOfMorphic.class.st
+++ b/src/BaselineOfMorphic/BaselineOfMorphic.class.st
@@ -149,11 +149,11 @@ BaselineOfMorphic >> cleanUpAfterMorphicInitialization [
 	Symbol rehash.
 
 	"Remove empty packages and protocols"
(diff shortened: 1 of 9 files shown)


jecisc:
I added more things to the PR and update the first comment to reflect all the changes

jecisc:
There is one failing test in document browser with this change

jecisc:
The test is green in the artifact :/

Gabriel-Darbord (approved):

PR #14314, opened 2023-07-17
Try to put category unregistration in a better place

jecisc:
Currently we can remove some tags without removing the categories


jecisc:
This change breaks the image :/
I'll try another approach

PR #14325, opened 2023-07-18
Iterate over a copy of the actionMap in Object>>removeActionsSatisfying: (fixes #14317)

### LINKED ISSUE #14317: Object>>removeActionsSatisfying: modifies the actionMap while iterating
**Bug description**
`Object>>removeActionsSatisfying:` uses a `#keysDo:` over the `actionMap`, but `#removeActionsSatisfying:forEvent:` may remove the event entirely from the map if the last action is removed. `#removeActionsForReceiver:` correctly iterates over a copy (though should really be implemented in terms of `#removeActionsSatisfying:` in the first place).

**To Reproduce**
1. Add actions for a bunch of different events of some object, that have some property in common so they can be selected for removal in one go. At least most of these should have no other actions for that event. (The easiest such criterion is the receiver—indeed that's how I found the bug, I didn't know `#removeActionsForReceiver:` existed so passed my own condition block to `...Satisfying:`.)
2. Remove them all with `#removeActionsSatisfying:`.
3. Trigger one of them and notice that an action that should have been removed is executed (or just examine the actionMap directly).

**Expected behavior**
Actions are correctly removed from the map even when that means removing keys as well.

**Version information:**
 - OS: Windows (applicable to any)
 - Pharo: Found in 11, still present in 12

**Expected development cost**
Pull request incoming soon.

daniels220:
Defer to this implementation rather than reimplementing it for #removeActionsWithReceiver:

Fixes #14317

### DIFF
--- a/src/System-Object Events-Tests/EventManagerTest.class.st
+++ b/src/System-Object Events-Tests/EventManagerTest.class.st
@@ -209,6 +209,35 @@ EventManagerTest >> testRemoveActionsWithReceiver [
 	self assert: ((eventSource hasActionsWithReceiver: self) not)
 ]
 
+{ #category : #'running - remove actions' }
+EventManagerTest >> testRemoveManyActionsSatisfying [
+	"See issue #14317--removing many actions wwould sometimes fail due to modification-under-iteration."
+
+	1 to: 10 do: [ :i |
+		eventSource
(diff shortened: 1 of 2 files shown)


jecisc (approved):
Thanks!

PR #14327, opened 2023-07-18
Adapt check in RBAddMethodChange>>protocols: now that the argument may be a first-class Protocol (fixes #14324)

### LINKED ISSUE #14324: Accept changes from Rewrite Rule Editor gets Protocol DNU #first
**To Reproduce**
1. Open the Rewrite Rule editor and input a transformation rule.
2. Run it and accept some method changes. The methods being changed may have to be in a real protocol (not unclassified), I'm not sure.
3. See walkback/debugger with Protocol DNU #first error.

**Bug description**
Looks like protocols used to be just strings, and at some point became a first-class Protocol object. `RBAddMethodChange.protocols` is, as the name suggests, supposed to be a collection of protocols, but the `#protocols:` setter is meant to accept a single one (not wrapped in a collection) as well. It checks for this using `#isString`, but it no longer _is_ a string, so when handed a single Protocol it assigns it directly and then, when asked for a (definitely singular) `#protocol`, sends it `#first` and gets the error.

The simplest solution would be to flip the conditional so it's `aCollectionOrProtocol isCollection ifTrue: [ aCollectionOrProtocol ] ifFalse: [ { aCollectionOrProtocol } ]`. However this would break if it is still legal to pass a string, which it well might be, and legacy code might do so.

In my own work, I long ago wrote a method `#isTrueCollection`, that is, a collection that is not a string, which is considerably more foolproof in cases like this. In fact this basic logic of "I need a collection but you can hand me just one thing and I'll figure it out" is so common that I also defined `#asTrueCollection`, which wraps the receiver in an array if it is not a "true" collection already.

**Expected development cost**
I can make a PR for the simplest solution (just inlining `isCollection and: [isString not]`), but I'd be open to contributing `isTrueCollection` and `asTrueCollection` as well.

daniels220:
Check if it's already a collection-but-not-a-string, otherwise wrap in an array, rather than checking if it *is* a string and wrapping in an array if so.

Fixes #14324

### DIFF
--- a/src/Refactoring-Changes/RBAddMethodChange.class.st
+++ b/src/Refactoring-Changes/RBAddMethodChange.class.st
@@ -214,11 +214,11 @@ RBAddMethodChange >> protocols [
 ]
 
 { #category : #'initialize-release' }
-RBAddMethodChange >> protocols: aCollectionOrString [
+RBAddMethodChange >> protocols: aCollectionOrProtocol [
 
-	protocols := aCollectionOrString isString
-		             ifTrue: [ { aCollectionOrString } ]
-		             ifFalse: [ aCollectionOrString ]
+	protocols := (aCollectionOrProtocol isCollection and: [ aCollectionOrProtocol isString not])
+		             ifTrue: [ aCollectionOrProtocol ]
+		             ifFalse: [  { aCollectionOrProtocol } ]
 ]
 
 { #category : #accessing }


jecisc (approved):
Oh! This is a good catch thank you :)

I am the one that replaced strings by first class protocols but I missed this case. 
It might be nice in the future to have a test on this

daniels220:
I thought about tests but didn't see any others for RBAddMethodChange, and I'm not familiar enough with the Pharo test suite to know if there's another test class that would be appropriate. If there is, or if you have any guidance for what a hypothetical `RBAddMethodChangeTest` should look like, I might be able to do something, but I'm not that familiar with the RB refactorings and changesets code, so I'm not very confident starting from scratch on my own.

PR #14328, opened 2023-07-18
Update regex doc

Gabriel-Darbord:
Mainly formatting: code blocks, spacing, lines...
Plus a few content changes: spelling, grammar, a missing example...

Time has not been kind to the visuals of this documentation. In particular, it seems that all code blocks are always treated as Smalltalk snippets (maybe a bug?).

image

image

### DIFF
--- a/doc/Regex/1-Introduction.md
+++ b/doc/Regex/1-Introduction.md
@@ -1,31 +1,20 @@
 # Introduction
 
-A regular expression is a template specifying a class of strings. A
-regular expression matcher is a tool that determines whether a string
-belongs to a class specified by a regular expression.  This is a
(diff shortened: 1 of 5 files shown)


guillep (approved):
Thanks Gabriel, this is super nice!

About the formatting, did you try specifying the language in the code block?
I think the code block interpreter in image will assume it's pharo code most of the times.

Anyways, that's for a separate issue.

jecisc (approved):

Gabriel-Darbord:
Thanks Guille, I figured we can make a plain text code block by specifying the language as `text`.
Some formatting didn't work as intended so I'm doing another pass.

PR #14332, opened 2023-07-19
Do not rename category through announcements

jecisc:
During a category renaming, synch RPackage directly instead of using announcements


jecisc:
I updated some tests that were mixing organizers and I also removed a tests that was checking a fallback for a buggy behavior that is not buggy anymore

Gabriel-Darbord (approved):

PR #14335, opened 2023-07-20
Deprecate CategoryAdded announcement

jecisc:
CategoryAdded only has one subscriber currently in the system and it is RPackage.

This change tries to create RPackages without announcements when we add a category to the SystemOrganizer.

Then it removes CategoryAdded announcement and declare it as a deprecated alias of PackageTagAdded.


jecisc:
Redoing this PR

PR #14336, opened 2023-07-20
Clean method compilation of Monticello

jecisc:
Monticello reimplement itself the method installation trying to use the compiler only to compile the method and then managing itself the installation in the class and the change file update.

This seems like some useless duplication. Here is a try to use Opal compiler directly for the method installation to be sure everything is coherent and that there is no bug in what Monticello does. 

This should fix some vanishing protocol creations or other missing announcements.


Ducasse:
Excellent!
This is nice to remove duplicated logic from MC!

Ducasse:
The builds are failing but I do not get how this is related

```
MetacelloNotification: Loading -> Spec2-Transformations-tonel.1 --- github://pharo-spec/Spec:Pharo12/src --- cache
MetacelloNotification: Loaded -> Spec2-Transformations-tonel.1 --- github://pharo-spec/Spec:Pharo12/src --- cache
Instance of FinderUI class did not understand #icon
FinderUI class(Object)>>doesNotUnderstand: #icon
FinderUI class>>finderMenuOn:
```

may be an extension is not installed.

jecisc:
Same :( 
I'll have to investigate because I have no idea what can be the problem here.

The method #icon is not in extension so this is not the problem here.

I might try to do this change in smaller PRs to try to locate the problem.

jecisc:
I'm doing new set f changes that works better than what I did here so I'll close it

PR #14345, opened 2023-07-21
[Backport] rapid click on two different items in a list gets interpreted as a double-click on the first item

ironirc:
This PR solves the problem with 2 subsequent clicks (within double click timing) on 2 different items of FTTableMorph.
Before the fix, this would execute a double click command on the first clicked item.
The fix involves comparing the 2 click positions and resulting item indexes.

A second problem occurs when double clicking on an already selected item.
Before the fix, the item gets de-selected, and nothing further happend.
This PR ensures that item is selected before dispatching the double click command.


jecisc:
Hi, 

Current bug fixes should be done on Pharo 12. If the fix is integrated in Pharo 12 and works fine, we can then backport it to Pharo 11.
I'll try to rebase it and see if we have conflict against P12

jecisc:
It brings other changes so I cannot just rebase it against P11

ironirc:
@jecisc What should I do?

jecisc:
I'll try to cherry pic your fix against P12. 
We can keep this PR for the backport but it should be merge after the P12 one

jecisc:
https://github.com/pharo-project/pharo/pull/14346

MarcusDenker:
(I did a close/reopen so we run the tests against the newest version of Pharo11)

MarcusDenker:
The fix is in Pharo12 since end of July, it seems no problems came up.

We can merge the backport when the tests look ok

MarcusDenker:
The CI run has some tests failing due to a DNU for #asPseudoDoubleClickEvent, for example #testDoubleClickActivatesRowInDoubleClickActivationMode

```
Error
Message not understood: MouseEvent >> #asPseudoDoubleClickEvent
Stacktrace
MessageNotUnderstood
Message not understood: MouseEvent >> #asPseudoDoubleClickEvent
MouseEvent(Object)>>doesNotUnderstand: #asPseudoDoubleClickEvent
SpFTTableMorph(FTTableMorph)>>doubleClick:
SpMorphicBackendForTest>>doubleClickFirstRowAndColumn:
SpListCommonPropertiestTest(SpAbstractListCommonPropertiestTest)>>testDoubleClickActivatesRowInDoubleClickActivationMode
SpListCommonPropertiestTest(TestCase)>>performTest
[ presenter := self classToTest new.
			self initializeTestedInstance.
			super performTest ] in SpListCommonPropertiestTest(SpAbstractAdapterTest)>>performTest
SpMorphicBackendForTest>>runTest:
SpListCommonPropertiestTest(SpAbstractAdapterTest)>>performTest



```

jecisc:
I think this is because a change was integrated to Spec for the Pharo 11 branch but we need to do a new release

jecisc:
BaselineOfPharo class>>#spec needs to be updated to point to v1.2.4 instead of v1.2.3

PR #14346, opened 2023-07-21
rapid click on two different items in a list gets interpreted as a double-click on the first item

### LINKED ISSUE #14168: rapid click on two different items in a list gets interpreted as a double-click on the first item
It's quite inconvenient having to close the window that opens unintentionally due to this issue.

jecisc:
Fixes #14168

This PR solves the problem with 2 subsequent clicks (within double click timing) on 2 different items of FTTableMorph.
Before the fix, this would execute a double click command on the first clicked item.
The fix involves comparing the 2 click positions and resulting item indexes.

A second problem occurs when double clicking on an already selected item.
Before the fix, the item gets de-selected, and nothing further happend.
This PR ensures that item is selected before dispatching the double click command.

### DIFF
--- a/src/Morphic-Core/MouseButtonEvent.class.st
+++ b/src/Morphic-Core/MouseButtonEvent.class.st
@@ -16,11 +16,8 @@ MouseButtonEvent >> asPseudoDoubleClickEvent [
 	| pseudoEvent secondEvent |
 	pseudoEvent := MouseButtonPseudoDoubleClickEvent new copyFrom: self.
 	secondEvent := self copy.
-	secondEvent
-		setType: self type
-		position: self hand cursorPoint
-		buttons: self buttons
-		hand: self hand.
+	secondEvent setPosition:
+		(self hand ifNil: [ self position ] ifNotNil: [ :h | h cursorPoint ]).
 	pseudoEvent secondEvent: secondEvent.
 	^ pseudoEvent
 ]
--- a/src/Morphic-Core/MouseButtonEvent.class.st
+++ b/src/Morphic-Core/MouseButtonEvent.class.st
@@ -11,6 +11,21 @@ Class {
 }
 
 { #category : #accessing }
+MouseButtonEvent >> asPseudoDoubleClickEvent [
+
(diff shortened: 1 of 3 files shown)


ironirc:
@jecisc I see the tests are failing :-(
I think you missed the first commit, which contains `#asPseudoDoubleClickEvent `and a new class `MouseButtonPseudoDoubleClickEvent`

Cause of ending up with 2 commits is: 
When committing with iceberg, it doesn't take the very last actual code, but instead what is cached in the commit overview. Subsequently after the commit, iceberg shows the actualized comparison.
Every now and then I fall into this "trap" ;-)

Just let me know if I you want me do redo the PR(s) the proper way.

Ducasse:
I do not get why the bootstrap is not working so I relaunched.

ironirc:
Things got complicated fast, due to my initial mistake... sorry for that.
My initial PR was targeted at Pharo11.  Then @jecisc cherry picked the changes and created a Pharo12 PR.
I saw tests failed and made corrections. But I could only update the Pharo11 PR.
So somebody needs to cherry pick those last changes and copy them over to this PR.

Moreover, my solution also depends on a minor change to the Spec tests. So that PR needs to done first.
Please see: https://github.com/pharo-spec/Spec/pull/1427.

One more remark: 
To me it feels that the initial dispatched doubleClick event, which is now MouseButtonEvent (mouseDown), should in fact be a new kind of event, i.e. MouseDoubleClickEvent (or something). Potential consumers of the event need information on both clicks to decide what should happen.

Ducasse:
I will wait for cyril to cherry pick your new changes. For now I integrated the spec changes.

jecisc:
Since the spec change was merged I'll restart the CI

ironirc:
What went wrong?

jecisc:
Our windows slave has some troubles. The PR has green tests now and needs a review to be integrated.

Ducasse:
Seems ok for me. So I will integrate it.

ironirc:
Thanks folks!

PR #14347, opened 2023-07-21
SImplify class addition in RPackage

jecisc:
Class addition is RPackage is complexe and here is a first step at simplifying it by removing a method used only in tests

### DIFF
--- a/src/RPackage-Core/RPackage.class.st
+++ b/src/RPackage-Core/RPackage.class.st
@@ -686,9 +686,7 @@ RPackage >> importClass: aClass [
 
 	self basicImportClass: aClass.
 	self basicImportClass: aClass classSide.
-	self
-		addClassDefinition:  aClass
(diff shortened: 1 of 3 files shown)


src/RPackage-Core/RPackage.class.st
```diff
@@ -693,9 +686,7 @@ RPackage >> importClass: aClass [
 
 	self basicImportClass: aClass.
 	self basicImportClass: aClass classSide.
-	self
-		addClassDefinition:  aClass
-		toClassTag: (self toTagName: aClass category)
+	self addClassDefinition: aClass toClassTag: aClass category
```
Ducasse, line 689: Could we remove the category here?

jecisc: My main goal is to remove the categories but I need to simplify RPackage for that as a first step because it is way too complex

PR #14348, opened 2023-07-21
Remove duplicated import logic

jecisc:
In RPackage we have way too many ways to import classes/methods/... Here I'm trying to remove one to use a simpler way. I'm really not sure this will be enough because the way things get imported is really different but let's see what the bootstrap says since this way is used only by the bootstrap


jecisc:
This does not seems to work. I'll try to do that in smaller steps

PR #14349, opened 2023-07-22
Cleanup transformation/refactoring hierarchy

balsa-sarenac:
Clean-up Refactoring-Transformation hierarchy by pushing-up some duplicated methods.
Start removing usages of `preconditionChecking`. `preconditionChecking` is now only difference between base RBTransformation and RBRefactoring.


Ducasse:
Sound good to me. We should keep pushing this.

PR #14350, opened 2023-07-22
Deprectaed #definedClassesDo:

jecisc:
The method #definedClassesDo: is bad in multiples aspects:
- It has duplicated elements and can iterate two times over the same classes
- It iterates over the **names** of the classes while the selector let us think that it iterates over the classes.
- It does not iterate over classes without any selector!
- It used two variables that I want to remove from RPackage :D

It is as simple to do `package definedClasses do:` and the call #name on the class if needed so I propose to just deprecate this one.

The little bonus is that I want to get rid of the variables #classDefinedSelectors and #metaclassDefinedSelectors and this PR remove one users for both

### DIFF
--- a/src/Deprecated12/RPackage.extension.st
+++ b/src/Deprecated12/RPackage.extension.st
@@ -1,0 +1,9 @@
+Extension { #name : #RPackage }
+
+{ #category : #'*Deprecated12' }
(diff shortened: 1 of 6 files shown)


Ducasse:
Thanks for cleaning!

PR #14351, opened 2023-07-23
Backport display scale factor improvements to Pharo 11

Rinzwind:
This pull request backports the display scale factor improvements from earlier pull requests to Pharo 11.


jecisc:
This seems good to me.

But for backport I prefer to wait for other opinion to know if we backport it or not :)

PR #14354, opened 2023-07-24
Forbid to have an extension method in the package owning the class

jecisc:
It is currently possible to have a method that is an extension method that is in the same package as its class. This creates weird behaviors were some tools consider it an extension method and some other consider it not an extension method.

To make thing more coherent I propose that we forbid to have this by classifying such methods under the unclassified protocol instead.

Fixes ##14353


src/Kernel/ClassDescription.class.st
```diff
@@ -264,6 +264,9 @@ ClassDescription >> classify: selector under: aProtocol [
 
 	"In case the method is already classified and we say to classify it under nil, we do not update the protocol."
 	(aProtocol isNil and: [ oldProtocol isNotNil ]) ifTrue: [ ^ self ].
+	
+	"If we try to classify the method as an extension of its own package, just unclassify the method to avoid any weird state where the method could be considered as an extension or not depending on the interpretation."
+	(self isProtocolExtensionFromTheSamePackage: aProtocol) ifTrue: [ ^ self classify: selector under: Protocol unclassified ].
```
tesonep: Why not also signalling a warning to notify this problem? 
The warning can proceed in interactive mode and is logged in headless mode.

src/Kernel/ClassDescription.class.st
```diff
@@ -959,6 +979,9 @@ ClassDescription >> renameProtocol: anOldProtocol as: aNewProtocol [
 	(aNewProtocol isNil or: [ anOldProtocol isNil ]) ifTrue: [ ^ self ].
 
 	(self hasProtocol: anOldProtocol) ifFalse: [ ^ self ].
+	
+	"We should not be able to have a protocol that is an extension of the same class so let's skip the rename."
+	(self isProtocolExtensionFromTheSamePackage: aNewProtocol) ifTrue: [ ^ self ].
```
tesonep: The same as before, it is better to generate a warning.

tesonep: Having something that does nothing without telling the user is really bad behaviour.

tesonep (changes requested):

src/Kernel/ClassDescription.class.st
jecisc: I hesitated about that but I agree that it is better. I'll update the PR

jecisc:
Apparently the warning is preventing the bootstrap to finish. Do you know what I should do so that it just log the problem and proceed?

jecisc:
Checking the code of the compiler it seems that the right way would be to use a Notification or even a SystemNotification instead of a warning

jecisc:
The build now passes and things are working correctly. In the non interactive environment things get logged. 
I just broke some tests that I'll fix now

jecisc:
The last failure is fixed by this: https://github.com/pillar-markup/Microdown/pull/658

But I need someone with merge rights :'(

jecisc:
The requested changes were applied and the tests are now green

PR #14356, opened 2023-07-24
safer morph access in NECMenuMorph>>#mouseDown:

pavel-krivanek:
When the mouse even processing happens, the owning morph may not exist anymore.


tesonep (approved):
An improvement for the ugly hack. It is much better now 👍

astares:
Can we backport this to P11 ?

See https://github.com/pharo-project/pharo/issues/14798

PR #14360, opened 2023-07-24
Clean method removal of RPackage

jecisc:
This is a first step at the cleaning of the method removal of RPackage. 

This is still quite bad but it is a first step a cleaning this. The change make the method removal work as the method addition and reduces the usage of the cache variables of RPackage a little. My mid term goal is to kill some of those caches.

### DIFF
--- a/src/RPackage-Core/RPackage.class.st
+++ b/src/RPackage-Core/RPackage.class.st
@@ -216,36 +216,20 @@ RPackage >> addClassTag: aSymbol [
 { #category : #'add method - compiled method' }
 RPackage >> addMethod: aCompiledMethod [
 	"Add the method to the receiver as a defined method if the class is  defined in it, else as an extension."
-	| methodClass |
 
+	| methodClass |
 	methodClass := aCompiledMethod methodClass.
-	(self includesClass: methodClass )
+	(self includesClass: methodClass)
 		ifTrue: [
 			methodClass isMeta
-				ifTrue: [
-					(metaclassDefinedSelectors
-						at: methodClass instanceSide name
-						ifAbsentPut: [ Set new ])
-						add: aCompiledMethod selector]
-				ifFalse: [
-					(classDefinedSelectors
-						at: methodClass name
-						ifAbsentPut: [ Set new ])
-						add: aCompiledMethod selector ] ]
-		ifFalse:  [
+				ifTrue: [ (metaclassDefinedSelectors at: methodClass instanceSide name ifAbsentPut: [ Set new ]) add: aCompiledMethod selector ]
+				ifFalse: [ (classDefinedSelectors at: methodClass name ifAbsentPut: [ Set new ]) add: aCompiledMethod selector ] ]
+		ifFalse: [
 			methodClass isMeta
-				ifTrue: [
-					(metaclassExtensionSelectors
-						at: methodClass instanceSide name
-						ifAbsentPut: [ Set new ])
-						add: aCompiledMethod selector.]
-				ifFalse: [
-					(classExtensionSelectors
-						at: methodClass name
-						ifAbsentPut: [ Set new ])
-						add: aCompiledMethod selector].
-				"we added a method extension so the receiver is an extending package of the class"
-				self organizer registerExtendingPackage: self forClass: methodClass ].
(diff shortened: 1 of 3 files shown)


src/RPackage-Core/RPackage.class.st
```diff
@@ -1229,18 +1188,34 @@ RPackage >> removeFromSystem [
...
-				ofMetaclassName: methodClass instanceSide originalName ]
+			methodClass isMeta
+				ifTrue: [
+					metaclassDefinedSelectors at: methodClass instanceSide originalName ifPresent: [ :methods |
+						methods remove: aCompiledMethod selector ifAbsent: [  ].
+						methods ifEmpty: [ metaclassDefinedSelectors removeKey: methodClass instanceSide originalName ] ] ]
+				ifFalse: [
+					classDefinedSelectors at: methodClass originalName ifPresent: [ :methods |
+						methods remove: aCompiledMethod selector ifAbsent: [  ].
+						methods ifEmpty: [ classDefinedSelectors removeKey: methodClass originalName ] ] ] ]
```
PalumboN, line 1204: On add is used `name` as key, but here is using `originalName`. What's the difference?

src/RPackage-Core/RPackage.class.st
```diff
@@ -1229,18 +1188,34 @@ RPackage >> removeFromSystem [
...
-				ofClassName: methodClass originalName ].
+			methodClass isMeta
+				ifTrue: [
+					metaclassExtensionSelectors at: methodClass instanceSide originalName ifPresent: [ :methods |
+						methods remove: aCompiledMethod selector ifAbsent: [  ].
+						methods ifEmpty: [ metaclassExtensionSelectors removeKey: methodClass instanceSide originalName ] ] ]
+				ifFalse: [
+					classExtensionSelectors at: methodClass originalName ifPresent: [ :methods |
+						methods remove: aCompiledMethod selector ifAbsent: [  ].
+						methods ifEmpty: [ classExtensionSelectors removeKey: methodClass originalName ] ] ] ].
```
PalumboN, line 1214: Should be nice to hide and reuse the use of the cache in both methods.

Maybe delegating to methods
```
self cacheFor: methodClass ifPresent: [ .. ]
self cacheFor: methodClass ifAbsent: [ .. ]
```

src/RPackage-Core/RPackage.class.st
```diff
@@ -1229,18 +1188,34 @@ RPackage >> removeFromSystem [
...
+
+	((metaclassExtensionSelectors at: methodClass instanceSide originalName ifAbsent: [ #(  ) ]) isEmpty and: [
+		 (classExtensionSelectors at: methodClass instanceSide originalName ifAbsent: [ #(  ) ]) isEmpty ]) ifTrue: [
```
PalumboN, line 1217: `classExtensionSelectors` is using `methodClass originalName` (or name) as key.

Copy-paste issue?

PalumboN (commented):
I left some comments during merging :P

src/RPackage-Core/RPackage.class.st
jecisc: The difference happens with ObsoleteClasses. #name will return the obsolete name while #originalName will return the name of the class when it was not obsolete. 

In the end I would like to have real classes instead of class names in the caches so that we do not have to care about that (and so that we do not have to cast them each time we want a class)

src/RPackage-Core/RPackage.class.st
jecisc: I'm planning to kill some of those caches. For now I'm trying to make the code simpler so that I can do my changes. 

I also want to merge the class and metaclass caches because it's a bother to always check the #isMeta to call two different methods. Once this is done, I'll see if we can improve further the use of the extension selectors :)
This beast is pretty huge and I'm trying to simplify little by little because each change has a lot of impacts ahah

src/RPackage-Core/RPackage.class.st
jecisc: There is no issue because here we are in the #ifFalse: of `methodClass isMeta` so we know we already have the instance side.

But as I said in another comment, I would like to remove the meta/non meta management to use one unified way.

PR #14362, opened 2023-07-25
MethodAddition: let the system manage the announcements

jecisc:
This is a subpart of #14336

This change let the system do the announcements while installing a method instead of preventing announcements and launching its own.


jecisc:
I wonder if the problem is not that now we get all announcements and react to it, but the method #icon is not yet loaded.

jecisc:
I'll close this one and work in the original PR

PR #14363, opened 2023-07-25
Do not remove classes from packages using the name

jecisc:
One of my mid term goal in the simplification of RPackage is to use real classes instead of manipulating class names everywhere.

Here is on step. 

RPackage has two ways of removing a class. #removeClass: and #removeClassNamed: and the version using the class directly call multiple methods in which I found bugs because it was not taking into account obsolete classes. 
Of course, the system currently uses #removeClassNamed: to remove a class. 

I propose in this change to first fix the bugs I found and then to use #removeClass: and to remove all the code based on the names since I want to have real classes in packages in the future.

I also started to remove the name management from the registration.

### DIFF
--- a/src/RPackage-Core/RPackage.class.st
+++ b/src/RPackage-Core/RPackage.class.st
@@ -1074,40 +1074,23 @@ RPackage >> removeAllMethodsFromClass: aClass [
 RPackage >> removeClass: aClass [
 	"I remove the class, methods and potential empty tags from myself."
 
-	| className |
-	className := aClass instanceSide originalName asSymbol.
+	"First I remove all the methods of the class from myself."
 	self removeAllMethodsFromClass: aClass.
-	self removeClassDefinitionNamed: className.
-	self removeClassTagsForClassNamed: className
+
+	"Then I remove the class from myself."
+	(classes remove: aClass originalName asSymbol ifAbsent: [ nil ]) ifNotNil: [ self unregisterClass: aClass ].
+
+	"Lastly I unregister it from the tags"
+	(self classTags select: [ :eachTag | eachTag includesClass: aClass ]) do: [ :eachTag | self removeClassDefinition: aClass fromClassTag: eachTag name ]
 ]
 
 { #category : #'class tags' }
 RPackage >> removeClassDefinition: aClass fromClassTag: aSymbol [
-	"Detags the class aClass with the tag aSymbol"
 
-	self
-		removeClassDefinitionName: aClass name
-		fromClassTag: aSymbol
-]
-
-{ #category : #'class tags' }
-RPackage >> removeClassDefinitionName: aClassName fromClassTag: aSymbol [
-	"Detags the class aClass with the tag aSymbol"
 	| tag |
-
 	tag := self classTagNamed: (self toTagName: aSymbol) ifAbsent: [ ^ self ].
-	tag removeClassNamed: aClassName.
-	tag isEmpty
-		ifTrue: [ self basicRemoveTag: tag ]
-]
-
-{ #category : #'private - register' }
-RPackage >> removeClassDefinitionNamed: aClassNameSymbol [
-	"aClassNameSymbol should be the name of a class and not a metaclass to be coherent with the class invariant that classes are only class names and not metaclass ones.
-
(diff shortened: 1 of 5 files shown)


src/RPackage-Core/RPackage.class.st
```diff
@@ -1045,21 +1045,16 @@ RPackage >> register [
...
 ]
 
 { #category : #'private - register' }
-RPackage >> removeAllMethodsFromClassNamed: aClassNameSymbol [
-	"Remove all the methods (defined and extensions) that are related to class aClassNameSymbol. aClassNameSymbol is the name of a class and not a metaclass - see class comment."
+RPackage >> removeAllMethodsFromClass: aClass [
+	"Remove all the methods (defined and extensions) that are related to the class as parameter. The class should always be instance side."
 
+	| aClassNameSymbol |
+	aClassNameSymbol := aClass originalName asSymbol.
```
Ducasse, line 1057: Here we can do `aClass originalName asSymbol -> aClass instanceSide originalName asSymbol.`

Ducasse: But it was true for the named version but I do not think this is correct for the class version because the metaclass has methods too. So ignore it.

src/RPackage-Core/RPackageOrganizer.class.st
```diff
@@ -744,20 +735,9 @@ RPackageOrganizer >> registerPackage: aPackage [
...
 
-{ #category : #'private - registration' }
-RPackageOrganizer >> registerPackage: aPackage forClassName: aClassNameSymbol [
-	"Register aPackage as the package of the class aClassNameSymbol. The package should contain the class in its class definitions."
-
-	(aPackage includesClassNamed: aClassNameSymbol asSymbol)
-		ifFalse:
-			[self error: aPackage name , ' does not include the class ' , aClassNameSymbol].
-	^classPackageMapping at: aClassNameSymbol put: aPackage
+	(aPackage includesClass: aClass) ifFalse: [ self error: aPackage name , ' does not includes the class ' , aClass name ].
```
Ducasse, line 739: We should probably introduce a specific error because self error: was not nice.

src/RPackage-Core/RPackageOrganizer.class.st
```diff
@@ -803,6 +783,15 @@ RPackageOrganizer >> removeCategory: aCategory [
 	SystemAnnouncer uniqueInstance classCategoryRemoved: category
 ]
 
+{ #category : #'system integration' }
+RPackageOrganizer >> removeClass: class [
+	"Remove the class, the class backpointer, the extensions and the extension backPointer from the receiver and the class involved with the class named: className. className is a class name and should not be a metaclass one. "
+
+	class isClassSide ifTrue: [ self error: 'We should only be able to remove classes and not instances of metaclass.' ].
```
Ducasse, line 790: Does it mean that if we have only extension to a metaclass the package still get the instance side and the metaclass?

Ducasse: I do not remember why I had this constraints about the metaclass.
Now I do not understand the `'We should only be able to remove classes and not instances of metaclass.'` because to me classes are instance of metaclass so I'm really confused.

jecisc: Maybe the comment is not good.
We can do `package removeClass: Object` but not `package removeClass: Object class`.

all of this instance side vs class side is because packages have caches based on instance or class side. I think I’ll try to unify that to remove all the checks in the code (but I’m doing one thing at a time right now)

Ducasse: Ok this is just that you are aware of it.

PalumboN: Another option is that this method works with both (instance and class sides). And get the object that it needed inside.

PalumboN (approved):

jecisc:
@PalumboN I hesitated to do that but this method is only called in one place of the system and there are low probability that anyone will use it in its code. So I think that the error is enough but I'll try to let a message that is more explicit in a future PR. I can also say in the error that if the person really want to do this they can do `package removeClass: Object class instanceSide`

PR #14364, opened 2023-07-25
Simplify class addition assuming classes are created empty

jecisc:
Some code in RPackage assumes that methods can be added to a package before the classes. But we never get in this case since classes are created. empty. I propose to simplify this code by removing tests showing a behavior never happening and by removing this complexe update code.

Three things to note:
- This was one of the reason preventing some changes to improve the tests. 
-  I plan to do another iteration after this one on class addition to make it simpler once again.
- This reduces the usage of the selectors caches I want to remove in RPackage.


Ducasse:
Cyril methods such as extendedMethodsBecomeDefinedForClass: are for changing the status of an extension. So this is not that the method is added in the package before the class but that the status of a method may change.

jecisc:
This method is called only when the class is created in the system, so the extensions are empty. 

I plan also to merge #addClassDefinition: with #basicImportClass: that checks more things (and ensure we manage well the defined selectors and extension selectors), but currently the bootstrap is using this.

We need to check the status of extensions only when we change a class of package and this already deals with the update of extension selectors in another way. 

So this code currently adds complexity but does not bring much value. 

Ps: I’ll take care of the conflict when I’ll have the time

PR #14366, opened 2023-07-26
Speed up in package management

jecisc:
One of my mid term goal is to remove some caches of RPackage that are making RPackage more complex to maintain in a good state with the rest of the system. But in order to do so, the code that will replace the caches needs to be fast enough which is not the case.

This changes speed up divers part of the system. For example:

```st
ms := SystemNavigation default allMethods.

[ ms do: [ :m | m package ] ] bench.

"Before: 7 iterations in 5 seconds 233 milliseconds. 1.338 per second"

"After: 16 iterations in 5 seconds 36 milliseconds. 3.177 per second"
```

List of changes related to speed:
- The way to find the package of a method was speed up and I also reduced the usage of an API I want to deprecate in the future.
- Divers Trait methods were optimized (and I also did some small cleanings such as improving some variables names).
- The collections used in RPackage were updated to be faster (even if I plan to remove some of them in the future)

Other changes:
- I deprecated #packageFromOrganizer: because it was lying on its name and was not using the parameter
- I updated #isExtension to not rely on protocols and use RPackage instead since we want to remove the usage of protocols for extensions in the future. (@dionisiydk I was able to do a faster version relying on packages faster than I though :) )


Ducasse:
Tests of AST-Core are breaking

```
Running tests of AST-Core-Tests
NewUndeclaredWarning: UndefinedObject>>DoIt (a is Undeclared)
NewUndeclaredWarning: MCMockClassE>>foo (a is Undeclared)
NewUndeclaredWarning: UndefinedObject>>DoIt (a is Undeclared)
NewUndeclaredWarning: MCMockClassE>>foo (a is Undeclared)
NewUndeclaredWarning: UndefinedObject>>DoIt (a is Undeclared)
NewUndeclaredWarning: MCMockClassE>>foo (a is Undeclared)
NewUndeclaredWarning: UndefinedObject>>DoIt (b is Undeclared)
NewUndeclaredWarning: MCMockClassE>>foo (b is Undeclared)
NewUndeclaredWarning: UndefinedObject>>DoIt (Δə is Undeclared)
NewUndeclaredWarning: MCMockClassE>>foo (Δə is Undeclared)
NewUndeclaredWarning: UndefinedObject>>DoIt (a is Undeclared)
NewUndeclaredWarning: UndefinedObject>>DoIt (a is Undeclared)
NewUndeclaredWarning: MCMockClassE>>foo (a is Undeclared)
NewUndeclaredWarning: MCMockClassE>>foo (a is Undeclared)
NewUndeclaredWarning: UndefinedObject>>DoIt (a is Undeclared)
NewUndeclaredWarning: MCMockClassE>>foo (a is Undeclared)
KeyNotFound: key #foo:bar: not found in MethodDictionary
MethodDictionary(Dictionary)>>errorKeyNotFound:
[self errorKeyNotFound: key] in MethodDictionary(Dictionary)>>at: in Block: [self errorKeyNotFound: key]
MethodDictionary>>at:ifAbsent:
MethodDictionary(Dictionary)>>at:
MCMockClassE class(Behavior)>>compiledMethodAt:
MCMockClassE class(Behavior)>>>>
[ :sel | methods add: (self environment at: key) >> sel ] in [ :key :val | val do: [ :sel | methods add: (self environment at: key) >> sel ] ] in RPackage>>methods in Block: [ :sel | methods add: (self environment at: ke[..]
IdentitySet(Set)>>do:
[ :key :val | val do: [ :sel | methods add: (self environment at: key) >> sel ] ] in RPackage>>methods in Block: [ :key :val | val do: [ :sel | methods add: (s[..]
[ :assoc |
		  aBlock value: assoc key value: assoc value ] in IdentityDictionary(Dictionary)>>keysAndValuesDo: in Block: [ :assoc |...
[ :each | each ifNotNil: [ aBlock value: each ] ] in IdentityDictionary(Dictionary)>>associationsDo: in Block: [ :each | each ifNotNil: [ aBlock value: each ] ]
IdentityDictionary(Dictionary)>>associationsDo:
IdentityDictionary(Dictionary)>>keysAndValuesDo:
RPackage>>methods
[ :p | p methods ] in RPackageSet>>methods in Block: [ :p | p methods ]
[ :each | stream nextPutAll: (aBlock value: each) ] in [ :stream | self do: [ :each | stream nextPutAll: (aBlock value: each) ] ] in Array(SequenceableCollection)>>flatCollect: in Block: [ :each | stream nextPutAll: (aBlock value: ea[..]
[ :stream | self do: [ :each | stream nextPutAll: (aBlock value: each) ] ] in Array(SequenceableCollection)>>flatCollect: in Block: [ :stream | self do: [ :each | stream nextPutA[..]
Array class(SequenceableCollection class)>>new:streamContents:
Array(SequenceableCollection)>>flatCollect:
RPackageSet>>methods
MCPackage>>snapshot
MCSnapshotResource class>>takeSnapshot
MCSnapshotResource>>setUp
[candidate setUp.
	candidate isAvailable ifTrue: [current := candidate]] in [[candidate setUp.
	candidate isAvailable ifTrue: [current := candidate]]
		ensure: [current == candidate ifFalse: [candidate tearDown]]] in MCSnapshotResource class(TestResource class)>>makeAvailable in Block: [candidate setUp....
[[candidate setUp.
	candidate isAvailable ifTrue: [current := candidate]]
		ensure: [current == candidate ifFalse: [candidate tearDown]]] in MCSnapshotResource class(TestResource class)>>makeAvailable in Block: [[candidate setUp....
CurrentExecutionEnvironment class>>activate:for:
DefaultExecutionEnvironment(ExecutionEnvironment)>>beActiveDuring:

```

src/RPackage-Core/CompiledCode.extension.st
```diff
@@ -1,23 +1,27 @@
 Extension { #name : #CompiledCode }
 
 { #category : #'*RPackage-Core' }
-CompiledCode >> package [
-	^ self packageFromOrganizer: RPackage organizer
+CompiledCode >> extensionPackage [
```
pavel-krivanek, line 4: not really important, but what about the name #extensionPackageOrNil, it sounds more low-level

pavel-krivanek (approved):

src/RPackage-Core/CompiledCode.extension.st
jecisc: One of the next step is to merge RPackage into Kernel when it will be clean and rid of the SystemAnnouncer. When it'll be done I wanted to put the method in a private protocol

PR #14368, opened 2023-07-26
Do not use global environment in RPackage

jecisc:
RPackage needs to know an environment to do some operations such as getting a real class from a class name. For now, the environment used was always the global environment. But in reality the package can be from another environment. 

This change update RPackage and RPackageTag to use the organizer environment to lookup for classes and methods and update the tests so that classes get added to the right environment and remove the hard coded global environment lookup.

This is a big step toward the goal of been able to have multiple enviornment, for example to be able to generate code in tests in special environments


PalumboN (approved):
🚀 🚥

PR #14369, opened 2023-07-26
Save real classes in RPackage

jecisc:
RPackage is currently keeping a list of class names to know the classes it contains. This means that we often need to lookup in the system dictionary to know the real classes.

I propose to save the real classes instead since it is easier to ask to a class its name than to lookup the class in the system.

I also make the collection containing the classes a IdentitySet since instance of classes should stay the same and it is faster. 

Lastly I removed a test that was really not useful since it tests a behavior that should never happen in the system (and if someone does that, then it's his problem, not RPackage problem)


Ducasse:
Do you know if this link will be ok when the class is marked as obsolete?
Because it was not to think about scenario like that that I kept names and not classes.
The other scenario I got in mind was a package with remote classes.

jecisc:
All the tests about obsolete classes are good. 

IIUC when a class become obsolete we remove it from its package and since it is the same instance but with the name updated it works.

I'm looking right now at the failing tests to try and understand the origin of the problem

jecisc:
I have some really weird things. Like an #ensure: that does not execute when there is an error in the initial block or a method that is not called while a log right before the call happens.

jecisc:
I'm redoing it

PR #14370, opened 2023-07-27
Reduce usage of `RPackage organizer`

jecisc:
We have multiple ways to access a RPackageOrganizer and one of them is to use `RPackage organizer`. But this way is acting as a dynamic global variable where we can change the organizer to use.

This was implemented because of the relation between RPackageOrganizer and SystemOrganizer and still cannot be removed because of some things managed through announcements. 

But here is a new step toward the cleaning of this ugly hack so that it is easier to remove in the future. I replaced most of the references to this method by a cleaner reference to the package organizer that does not use this emulated global variable.

I also adapted the code of some tests to make this change work and pushed the right usage of organizer in RPackage. (Some places were always using the global package organizer instead of using their own)


jecisc:
Failing tests are not related (They are due to the new random generator that was merged)

pavel-krivanek (approved):
looks ok

PR #14374, opened 2023-07-27
Refactor: unify refactoring and transformation execution api

balsa-sarenac:
This PR unifies refactoring/transformation API so that they can be used one with another. Changes are mainly in `RBTransformation` and include:

* removing `preconditionChecking` variable
* removing `delegatesPrecondition` for composite transformations
* checking `applicabilityConditions` by default and fixing bunch of tests to support this
* extracting `breakingChangeConditions` out of `applicabilityConditions` NOTE: all breaking change checks are __NOT__ called anymore, that's why so many failing tests need to be fixed. I'll fix them by splitting them on refactoring and transformation tests.

There are a bunch of failing tests, I am pushing this now since I won't be able to work on this until next week. Partial review is welcome.


src/Refactoring-DataForTesting/RBTransformationRuleTestData.class.st
```diff
@@ -280,6 +280,18 @@ RBTransformationRuleTestData >> checkMethod: aSmalllintContext [
 						classified: aSmalllintContext protocols]]
 ]
 
+{ #category : #accessing }
+RBTransformationRuleTestData >> class [
+
+	^ class
+]
```
jecisc: Usually it is a bad idea to have a method called "class" because it overrides the method allowing to get the class of an object.

For example, on compiled method. instead of having a method called "class" it was named "methodClass" so that we can still ask the class of the method.

balsa-sarenac: Yes, I'm aware. I only did it because this is a test fixture method. This will be removed in the next commit.

PR #14377, opened 2023-07-28
Simplify API of RPackage

jecisc:
RPackage API is huge and is exposing a lot of internal behavior about its caches to the users. 

In this PR I am doing an iteration on the cleaning of this API. I removed two low level methods that were used only in a trait test and I used a better RPackage API to do the same.

But the main change here is that I deprecated a lot of other methods that should mostly be used by the system to reduce a lot the API without making the code more complex.

For example:

```st
package includesDefinedSelector: #method1 ofClassName: class name.
package includesDefinedSelector: #method1 ofMetaclassName: class name
```

Can be replaced by

```st
package includesDefinedSelector: #method1 ofClass: class.
package includesDefinedSelector: #method1 ofClass: class class
```

### DIFF
--- a/src/Deprecated12/RPackage.extension.st
+++ b/src/Deprecated12/RPackage.extension.st
@@ -25,6 +25,22 @@ RPackage >> includesDefinedSelector: aSelector ofMetaclassName: aClassName [
 ]
 
 { #category : #'*Deprecated12' }
+RPackage >> includesExtensionSelector: aSelector ofClassName: aClassName [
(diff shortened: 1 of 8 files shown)


jecisc:
I added a new batch of changes.

In this changes I merged some private methods dealing with the class/meta class caches because I plan to merge both in the future.

And I updated some unused API (it was used only in tests) so that they do not rely on the instance variables I want to update in the future, but on the common API of RPackage. And I also deprecated them because they are really specific and are just polluting the API of RPackage.

They can probably be implemented in better ways directly from the client code

PR #14388, opened 2023-07-31
Do not remove classes through announcement

jecisc:
RPackage currently uses announcements to sync with the system. Since the goal is to integrate it in the MM, this should be removed to let the MM directly sync RPackage.

Here is a first step: Trying to remove the registration to ClassRemoved to remove a class for the packages.

Redo of #14289


jecisc:
Tests are finally green! It is possible that this will fix a bug @akevalion is facing (I hope ahah)

akevalion:
I will test it
Update: It fixes the ci
https://github.com/ObjectProfile/Roassal3/actions/runs/5712358000

PR #14389, opened 2023-07-31
Remove unused parameter

jecisc:
Just a small cleaning


Ducasse:
All tested are related to fuel.

PR #14391, opened 2023-07-31
Use classes in the class mapping of RPackageOrganizer

jecisc:
RPackageOrganizer has a class/package mapping currently using the class names for the mapping.

I propose to use classes directly now


jecisc:
It finally passes :)

Gabriel-Darbord (approved):

PR #14393, opened 2023-07-31
Remove the need of ClassFactoryWithOrganization

jecisc:
Looking at ClassFactoryForTestCase and related classes I found multiples problems and a lot of complexity. 

Here is a little list:
- It has a subclass ClassFactoryWithOrganization that does not do only what it has in the name, but it changes the behavior in some ugly ways
- It has an API close to the old class definition instead of delegating to the class builder
- It has a class that is really close in behavior (FLClassFactoryForTestCase) but with multiple changes
- It uses the category API
- It makes no difference between packages and tags

I wanted to do a PR to fix all of that but this is way too much for one PR. So here I propose a first change:

I made ClassFactoryForTestCase customizable with an environment removing the need of ClassFactoryWithOrganization and we can now throw away the hacks it contains.


jecisc:
Failure are not related (they come from another PR and they are already fixed in Pharo 12)

PR #14394, opened 2023-07-31
Push tag API in the class builder

jecisc:
The class builder is currently using the concept of categories to classify the classes in the system. 

We want to get away from the category API so the method #package: was added. But this is not enough because the users of #package: are once again mixing packages and tags. 
I propose to rely on the #category getter for now to build a class, but to use #package: and #tags: in the API to define the class. Like this, we can start to write the code we want to see once the categories will be removed.

To see an example check the changes in ClassFactoryForTestCase that I just did. Once this is integrated it will allow me to clean further ClassFactoryForTestCase in order to deplecate the old class building API and rely more on the #make: way to create classes.


jecisc:
Failures are not related

PR #14397, opened 2023-07-31
#optionInlineNone suppresses #optionInlineX at CompilationContext level (fixes #14323)

### LINKED ISSUE #14323: Compiler optionInlineNone produces incorrect output
**Bug description**

The `optionInlineNone` compiler option is supposed to be equivalent to disabling all the other `optionInlineX` options, but in fact it is only sorta _half_-equivalent—it disables the inlining of the messages themselves but _not the blocks_, resulting in _wildly_ incorrect semantics:

* I discovered this with `and:`/`or:`, which are actually the least affected—they just lose their short-circuiting, both sides are always evaluated. But this is certainly still a problem, e.g. `a notNil and: [a < b]` suddenly isn't actually safe against a `nil DNU #<`.
* `ifTrue:ifFalse:` and `ifNil:ifNotNil:` just...always execute both branches—with the argument to the not-nil block set to nil, as well, in that case. They do return the value from the correct one, but are completely broken for control-flow purposes.
* `repeat`, `timesRepeat:` and `whileTrue[:]` only evaluate the loop body once.
* `to:[by:]do:` evaluates the body once, but with the "argument" set to nil rather than the starting value, almost invariably causing an error.
* I don't want to think about what `caseOf:[otherwise:]` does...probably execute all the cases?

**To Reproduce**

Basically, just compile anything with `<compilerOptions: #(+ optionInlineNone)>`. Even just a DoIt:

```
|i|
<compilerOptions: #(+ optionInlineNone)>
i:= 1.
10 timesRepeat: [i := i + 1].
i. " => yields 2, should be 11"
```

**Expected behavior**
`#(+ optionInlineNone)` should be exactly equivalent to `#(- optionInlineAndOr optionInlineCase optionInlineIf ...etc...)` and produce correct semantics.

**Possible solutions**
The question here is where to intervene. Right now the only method that produces the right answer when called in isolation is `RBMessageNode>>isInlined`—all the various `isInlineX` methods there can still answer true when `optionInlineNone` is set, and in turn, `RBBlockNode>>isInlined` also doesn't check it, which is the source of this bug. The simplest/shortest solution would be to just add a check at the top of `RBBlockNode>>isInlined` (**and** `isInlinedLoop`, which has independent senders), but this seems prone to future breakage—if for instance someone checks `isInlineAndOr` in some third place. As such, I think the _safest_ solution is to change all the `CompilationContext>>optionInlineX` methods to `^self optionInlineNone not and: [options includes: #optionInlineX]`. The specific check for `optionInlineNone` in RBMessageNode>>isInlined could then be removed. This way `optionInlineNone` is very literally equivalent to turning all the others off at a pretty low level, so it shouldn't matter how they're used.

**Expected development cost**
Pull request incoming shortly for the suggested approach above.

daniels220:
Also added an early-out in RBBlockNode>>isInlined similar to the one in RBMessageNode. Either approach--suppression on CompilationContext, or checking on RBBlockNode--is sufficient on its own, but the CompilationContext approach is safer (others who ask #optionInlineX don't need to know about #optionInlineNone to get the correct behavior), while the early-out is more performant when #optionInlineNone is set, and seems like a reasonable thing to do.

Expanded test parameters on OCAST*TranslatorTest to include a sanity-check run using no parameters at all, and one where both the relevant enabling option and #optionInlineNone are both explicitly set. Without the fixes this does cause failures of e.g. OCASTAndOrTranslatorTest>>#testAndWithLeftFalseShortcircuitsRight with optimization #(#optionInlineAndOr #optionInlineNone), which is the situation reported in #14323.

### DIFF
--- a/src/OpalCompiler-Tests/OCASTTranslatorTest.class.st
+++ b/src/OpalCompiler-Tests/OCASTTranslatorTest.class.st
@@ -12,6 +12,17 @@ Class {
 }
 
 { #category : #'building suites' }
+OCASTTranslatorTest class >> inliningOptionsFor: enableOption [
+
+	^ {
+		  #(  ). "No options supplied"
+		  { enableOption }. "Explicitly enable this type of inlining"
+		  #( #optionInlineNone ). "Disable all inlining"
+		  "Disable all inlining, *but* the specific option is also set (and should be ignored). See bug #14323."
+		  { enableOption. #optionInlineNone }}
+]
+
+{ #category : #'building suites' }
 OCASTTranslatorTest class >> testParameters [
 
 	^ super testParameters
@@ -26,7 +37,7 @@ OCASTTranslatorTest >> compilationContext [
 	context := CompilationContext new
 		encoderClass: self encoder;
 		bindings: globals;
-		parseOptions: { #'+' . optimization };
+		parseOptions: #(#+) , optimization;
 		yourself.	
 	
 	^ context
--- a/src/OpalCompiler-Tests/OCASTTranslatorTest.class.st
+++ b/src/OpalCompiler-Tests/OCASTTranslatorTest.class.st
@@ -35,11 +35,10 @@ OCASTTranslatorTest >> compilationContext [
 
 	| context |
 	context := CompilationContext new
-		encoderClass: self encoder;
-		bindings: globals;
-		parseOptions: #(#+) , optimization;
-		yourself.	
-	
+		           encoderClass: self encoder;
+		           bindings: globals;
+		           yourself.
+	self optimization ifNotNil: [ :a | context parseOptions: #( #+ ) , a ].
 	^ context
 ]
 
--- a/src/OpalCompiler-Core/CompilationContext.class.st
+++ b/src/OpalCompiler-Core/CompilationContext.class.st
@@ -496,22 +496,29 @@ CompilationContext >> optionConstantBlockClosure [
 
 { #category : #options }
 CompilationContext >> optionInlineAndOr [
-	^ options includes: #optionInlineAndOr
+
+	^ self optionInlineNone not and: [
+		  options includes: #optionInlineAndOr ]
 ]
 
 { #category : #options }
 CompilationContext >> optionInlineCase [
-	^ options includes: #optionInlineCase
+
+	^ self optionInlineNone not and: [
+		  options includes: #optionInlineCase ]
 ]
 
 { #category : #options }
 CompilationContext >> optionInlineIf [
-	^ options includes: #optionInlineIf
+
+	^ self optionInlineNone not and: [ options includes: #optionInlineIf ]
 ]
 
 { #category : #options }
 CompilationContext >> optionInlineIfNil [
-	^ options includes: #optionInlineIfNil
+
+	^ self optionInlineNone not and: [
+		  options includes: #optionInlineIfNil ]
 ]
 
 { #category : #options }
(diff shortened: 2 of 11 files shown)


daniels220:
Alternatively, I could _remove_ the explicit checking of `optionInlineNone` from `RBMessageNode>>isInlined`, and un-add it in `RBBlockNode>>isInlined`—relying only on the suppression on the `CompilationContext`. This is in some ways cleaner, it's honestly a toss-up from my perspective.

Ducasse:
Thanks for your contribution. Marcus will have a look when he is back from vacation.

src/OpalCompiler-Tests/OCASTTranslatorTest.class.st
```diff
@@ -11,6 +11,17 @@ Class {
...
 }
 
+{ #category : #'building suites' }
+OCASTTranslatorTest class >> inliningOptionsFor: enableOption [
+
+	^ {
+		  #(  ). "No options supplied"
+		  { enableOption }. "Explicitly enable this type of inlining"
+		  #( #optionInlineNone ). "Disable all inlining"
+		  "Disable all inlining, *but* the specific option is also set (and should be ignored). See bug #14323."
```
guillep, line 21: I wouldn't add a bug reference in the source code

guillep (approved):
I think the changes look great! you did a great catch here!

Following up from the bug report, finally you did not add any extra assertion or test, is that right? if so, why (just for documentation purposes)?

jecisc:
We have 153 failing tests with this change. For example:

osx-64 / Tests-osx-64 / testEmptyMethodReturnsReceiver(#encoder->EncoderForSistaV1) – MacOSX64.OpalCompiler.Tests.AST.OCASTBasicTranslatorTest<1s
Error
Instances of UndefinedObject are not indexable
Stacktrace
Error
Instances of UndefinedObject are not indexable
UndefinedObject(Object)>>error:
UndefinedObject(Object)>>errorNotIndexable
UndefinedObject(Object)>>size
Array(SequenceableCollection)>>copyReplaceFrom:to:with:
Array(SequenceableCollection)>>,
OCASTBasicTranslatorTest(OCASTTranslatorTest)>>compilationContext
OCASTBasicTranslatorTest(OCASTTranslatorTest)>>compileSource:
OCASTBasicTranslatorTest(OCASTTranslatorTest)>>testSource:withArguments:
OCASTBasicTranslatorTest(OCASTTranslatorTest)>>testExample:withArguments:
OCASTBasicTranslatorTest(OCASTTranslatorTest)>>testExample:
OCASTBasicTranslatorTest>>testEmptyMethodReturnsReceiver
OCASTBasicTranslatorTest(TestCase)>>performTest

daniels220:
@guillep I didn't add additional test methods or assertions therein, but it turns out I didn't need to—the existing test suite is actually very comprehensive, it just wasn't being run in the specific circumstances where the bug occurred (having both e.g. `optionInlineAndOr` and `optionInlineNone` set—as happens when you add `optionInlineNone` via pragma, since most of the `optionInlineX`es are set by default). So I added some permutations to the `testParameters`, which means there are actually several hundred "new" tests when you run the full suite.

@jecisc Right, I ran all the surrounding tests at several points but apparently not at the very end. Simple matter of adding a nil-check when setting up the compilationContext. Should be good now. Is there something I need to do in general to say, "okay, this needed more work but I think I've done that work now"?

jecisc:
I removed the “need more work” tag, now let’s see what the ci says :)

guillep:
Test failures are flaky, thanks @jecisc for the review, @daniels220 for the good work!

PR #14398, opened 2023-08-01
Make ClassFactoryForTestCase follow changes on class definition

jecisc:
ClassFactoryForTestCase has a big API to create classes and traits and some problems. In this change I propose multiple cleanings.

This change mostly make the definition of classes and traits configurable like we are doing with the new fluid class builder. (In the future I think it would be even better to have a test case running in another environment than the Pharo environment and like that we would remove totally the need of this factory). All the class and trait creation methods are replaced by two methods #make: and #silentlyMake:.

Another change is to go away from the category API to use more the package API.

I removed duplications in tests about this class.

Another change is that the silent removal of class was buggy and did not work well. I fixed that.

Next step is to deprecate #defaultCategory and remove FLClassFactoryForTestCase to use this one instead.

There are still some hacks due to categories present but I cannot remove them before removing the SystemOrganizer.


jecisc:
This is not so easy because the silent class creation is cutting announcements and then RPackage is only able to partioally synch with the SystemOrganizer

Gabriel-Darbord (approved):

PR #14399, opened 2023-08-01
Do not add classes through announcements

jecisc:
A new step toward getting away from the synchronization of RPackageOrganizer through announcements. If this change works well, the new classes will not be added via announcements.

This would also be a step toward fixing some failing tests in Spec that are due to the face that some part of the system are trying to generate silently some code suspending announcements, but since RPackage is updated via announcements, it leaves it in an unsynchronized state.


jecisc:
That does not pass the bootstrap :(

astares:
Just for clarification: if there is no use of announcement(s), how should (custom) tool be aware that something was added?

jecisc:
Announcements are still there but the goal is to package the classes without it.
We have code to silently create classes in some places but because of the packaging relying on announcement, it creates unpackaged classes and also this slow down the system and creates bugs (if you want the package of the created class but you announcement listening got the priority over the RPackage one)

jecisc:
I'll redo this one

PR #14404, opened 2023-08-01
Use ClassFactoryForTestCase instead of FLClassFactoryForTestCase

jecisc:
Fuel implemented its own class factory that started as a total duplication of ClassFactoryForTestCase to run its tests.

This class is at the origin of some weird bugs we have in external projects while running tests.

Instead of keeping this duplication I'm proposing to use ClassFactoryForTestCase directly instead and to add the missing features to it. As you can see, this removes a lot of duplicated code :)


theseion:
It's great that you ported my additions to the builtin class factory, but a heads up would have been nice since I'm still maintaining the code for Pharo 12 ;)

PR #14406, opened 2023-08-02
Simplify category management

jecisc:
Classes have a fallback when they do not know their category or if the category is wrong but I wonder if we even end up in that case. Since I want to remove the categories I'd like to check first if we can simplify its management to ease the migration to packages/tags. 

Let's see if the produced image is working with this simplification


jecisc:
Not so simple apparently

PR #14407, opened 2023-08-02
Package: React to class renaming without announcement

jecisc:
This change tries to make the package react to class renaming without using announcements


jecisc:
Test failure is not related

PR #14410, opened 2023-08-02
Use real classes in #classDefinedSelectors

jecisc:
This change tries to use real classes instead of class names in the variable classDefinedSelectors. If this work I'll apply the same change to metaclassDefinedSelectors then the two extensions dictionaries.

### DIFF
--- a/src/RPackage-Core/RPackage.class.st
+++ b/src/RPackage-Core/RPackage.class.st
@@ -220,30 +220,14 @@ RPackage >> addMethod: aCompiledMethod [
 	(self includesClass: methodClass)
 		ifTrue: [
 			methodClass isMeta
-				ifTrue: [
-					(metaclassDefinedSelectors
-						at: methodClass instanceSide name
-						ifAbsentPut: [ IdentitySet new ])
-						add: aCompiledMethod selector]
-				ifFalse: [
-					(classDefinedSelectors
-						at: methodClass name
-						ifAbsentPut: [ IdentitySet new ])
-						add: aCompiledMethod selector ] ]
-		ifFalse:  [
+				ifTrue: [ (metaclassDefinedSelectors at: methodClass instanceSide name ifAbsentPut: [ IdentitySet new ]) add: aCompiledMethod selector ]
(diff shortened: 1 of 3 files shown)


src/RPackage-Core/RPackage.class.st
```diff
@@ -1072,13 +1056,13 @@ RPackage >> roots [
...
-	classExtensionSelectors valuesDo: [:each | sel addAll: each].
-	metaclassDefinedSelectors valuesDo: [:each | sel addAll: each].
-	classDefinedSelectors valuesDo: [:each | sel addAll: each].
-	^ sel
+	| allSelectors |
+	allSelectors := Set new.
+	metaclassExtensionSelectors valuesDo: [ :each | allSelectors addAll: each ].
+	classExtensionSelectors valuesDo: [ :each | allSelectors addAll: each ].
+	metaclassDefinedSelectors valuesDo: [ :each | allSelectors addAll: each ].
+	classDefinedSelectors valuesDo: [ :selectors | allSelectors addAll: selectors ].
```
Ducasse, line 1064: why only the last one (all the other :each -> selectors)?

src/RPackage-Core/RPackage.class.st
```diff
@@ -875,10 +859,10 @@ RPackage >> removeAllMethodsFromClass: aClass [
 
 	| aClassNameSymbol |
 	aClassNameSymbol := aClass originalName asSymbol.
-	classDefinedSelectors removeKey: aClassNameSymbol ifAbsent: [ nil ].
-	classExtensionSelectors removeKey: aClassNameSymbol ifAbsent: [ nil ].
-	metaclassDefinedSelectors removeKey: aClassNameSymbol ifAbsent: [ nil ].
-	metaclassExtensionSelectors removeKey: aClassNameSymbol ifAbsent: [ nil ].
+	classDefinedSelectors removeKey: aClass ifAbsent: [  ].
```
Ducasse, line 862: I wondered why only the first one is turned into a class and not the others

PR #14412, opened 2023-08-02
fixing deprecancies introduced by miss refactoring

Ducasse:
A driver should be started using the message runRefactoring.


jecisc:
Failures do not seems related but I don't really know where one of them come from (The unpackaged class)

Ducasse:
I have no idea about it either. 
And this is really not related with this changes. 

S

PR #14414, opened 2023-08-02
Creating a package should synch the SystemOrganizer

jecisc:
Creating a package does not create a category in the system organizer currently leading to problems while trying to migrate some parts of the system from the categories to packages


jecisc:
I have no idea how this change can cause this bug in the CI :(

jecisc:
Done in another PR :)

PR #14423, opened 2023-08-03
Epicea: Migrate category addition to package tag addition

jecisc:
Epicea is relying on categories to update the package tags. 

Here is a first step at cleaning this: rely on PackageTagAddition instead and update all the code related to package tag addition to not rely on the categories


jecisc:
Failing tests seems to be flacky tests

jecisc:
Included in #14427

PR #14424, opened 2023-08-03
Use PackageTagAdded instead of CategoryAdded

jecisc:
Make the system use PackageTagAdded instead of CategoryAdded since I want to deprecate CategoryAdded. 

Only two users left:
- Epicea that is been dealt with in another PR
- RPackage that still relies on it to update since changing that is breaking the bootstrap :(


jecisc:
Failing test is flacky

PR #14428, opened 2023-08-04
Epicea: Finish refactoring about packages and tags in Epicea

jecisc:
This PR finish the changes started in Epicea about packages and tags.

Included:
- Do not rely on category announcements
- Do not rely on the category API (We still use some category API but for the class updates and not the package and tags updates)
- Package rename changes were not working
- Package tag rename is now managed via tags
- Clean API
- Add tests


Ducasse:
Hi cyril 

I do not know if these breaking tests are ok. 
And pay attention not to code too late or too early :).

S
```
osx-64 / Tests-osx-64 / testRenameCategoryAlsoRenameAllExtensionProtocols – MacOSX64.Monticello.Tests.RPackage.RPackageCategorySynchronisationTest
<1s
osx-64 / Tests-osx-64 / testIsMessageSentInSystemWithTheSelectorInsideAnArray – MacOSX64.System.Support.Tests.Image.SystemNavigationTest
6s
unix-64 / Tests-unix-64 / testRenameCategoryAlsoRenameAllExtensionProtocols – Unix64.Monticello.Tests.RPackage.RPackageCategorySynchronisationTest
```

jecisc:
One is a flacky tests but the second one is a real failure I think. I’ll investigate :)

jecisc:
I fixed the test!

One nice thing is that it was the test that was wrong :)

There was an implementation of package renaming in RPackage that can be used to rename a package. But if you renamed a package via the SystemOrganizer, instead of using this method, it recorded another way to rename the package. And this other way had flaws. For example it was not announcing the renaming in the system so I removed it and I made the system organizer use the other implementation.

And the test was failing because the other implementation was doing a better job than the old one and the renamed package kept more info

jecisc:
But I see that the way package renaming renames the extension protocols is bad because it does not announce to the system that the protocols were renamed. But I'll take care of that in another PR

Gabriel-Darbord (approved):

PR #14430, opened 2023-08-04
TestRunner should not rely on category announcements

jecisc:
Instead I make it use the tag announcements. 

Tests runner, Epicea and RPackage are the only two last users of the category announcements. I already did a PR to fix Epicea and I am trying to update RPAckage alos even if the bootstrap does not seems to like it. But once those 3 changes are integrated, we can remove those announcements!


jecisc:
Failures are not related

Gabriel-Darbord (approved):

PR #14432, opened 2023-08-04
[Refactor] unified refactoring/transformation execution API

balsa-sarenac:
This PR continues refactoring execution API of refactoring and transformation:

* push up all execution logic to `RBAbstractTransformation`
* rename and deprecate `transform` to `privateTransform`
* deprecate `asRefactoring` since it's no-op and returns self


jecisc (approved):
Changes seems good. Mostly renaming and removing a useless call.

PR #14434, opened 2023-08-04
RPackageTag should store real classes

jecisc:
This change updates RPackageTag to store real classes instead of class names.

A next step could be to remove the #classes variable of RPackage and to just flat collect this variable in the tags. I ran some benches and it is pretty fast to do and could simplify the code quite a lot and remove some sync problems


jecisc:
there is one related tests failure but it is hard to find the origin of the problem

PR #14437, opened 2023-08-05
RPackage: Simplify some code in RPackageOrganizer

jecisc:
The goal of this change is to simplify the overall code of RPackageOrganizer	. The main change is to introduce #ensurePackageMathcing: to factorise some code. I also removed some methods that could just be inlined now that they are simple or unused (except for tests but they could use other methods)
	
I also added some comments.

There is only one change that can break something I have the impression, it is the import of packages used by the bootstrap because the old version used a way that does not announce packages creation and now I am using the normal code to create a package and this way announce the package creation. (This was on my big list of multiple way to create packages that I wanted to reduce and unify). Let's see if the bootstrap accepts this change


jecisc:
Btw, I also removed one test that I added myself recently because now that I know more about RPackage I noticed that it was not a good test

Gabriel-Darbord (approved):

PR #14445, opened 2023-08-07
Refactor: add method refactoring

balsa-sarenac:
Since we unified the execution API add method refactoring can now be composed with add method transformation and temporary class can be removed.


jecisc:
There are tests in failure because of undeclared variables

balsa-sarenac:
Failing tests seems unrelated.

PR #14448, opened 2023-08-07
14141-BlockClosureisClean-regression-from-Pharo-10-to-11 

### LINKED ISSUE #14141: BlockClosure>>#isClean regression from Pharo 10 to 11

MarcusDenker:
Backport from Pharo12

closes #14141

### DIFF
--- a/src/OpalCompiler-Core/OCBlockScope.class.st
+++ b/src/OpalCompiler-Core/OCBlockScope.class.st
@@ -9,7 +9,7 @@ Class {
 
 { #category : #testing }
 OCBlockScope >> hasEscapingVars [
-	^ (copiedVars isEmpty and: [tempVector isEmpty]) not
+	^ self inComingCopiedVarNames isNotEmpty
 ]
 
 { #category : #accessing }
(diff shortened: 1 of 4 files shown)


MarcusDenker:
Failing tests on windows not related, I will open an issue for those

Issue #14451, opened 2023-08-08
[Pahro11] Tests failing on WIN CI


MarcusDenker:
**Bug description**
We have lots of tests failing on WIN only, on Pharo11. 

There seem to be two reasons:

1 Error not supported yet. and ever, e.g. RSBarPlotTest(RSAbstractChartTest)>>testWithOnePlot

```

Error
not supported yet. and ever
Stacktrace
Error
not supported yet. and ever
StrikeFont(Object)>>error:
StrikeFont>>asFreetypeFont
LogicalFont>>asFreetypeFont
RSLabel>>cairoFont
RSCairoMetricsProvider(RSCairoFitMetricsProvider)>>extentsOf:
RSCairoMetricsProvider>>extentsOf:
RSLabel>>computeTextExtents
RSLabel>>textExtents
RSLabel>>textWidth
RSLabel>>resetCache
RSLabel>>text:
RSLegend>>customText:shape:
RSLegend>>title:
RSCanvasController>>renderLegendOn:
RSCanvasController>>onShape:
RSCanvas(RSObjectWithProperty)>>addInteraction:
RSChart>>defaultContainer
RSChart(RSAbstractContainerBuilder)>>container
RSChart(RSAbstractContainerBuilder)>>build
RSBarPlotTest(RSAbstractChartTest)>>testWithOnePlot
RSBarPlotTest(TestCase)>>performTest
```

2 primitive #primLoadSymbol:module: in TFFIBackend failed, e.g. AthensCairoMatrixTest(AthensTransformTest)>>testLoadIdentity

```
Error
primitive #primLoadSymbol:module: in TFFIBackend failed
Stacktrace
PrimitiveFailed
primitive #primLoadSymbol:module: in TFFIBackend failed
TFFIBackend(ProtoObject)>>primitiveFailed:
TFFIBackend(ProtoObject)>>primitiveFailed
TFFIBackend>>primLoadSymbol:module:
TFFIBackend>>loadSymbol:module:
ExternalAddress class>>loadSymbol:module:
TFExternalFunction>>validate
TFSameThreadRunner>>invokeFunction:withArguments:
AthensCairoMatrix>>primLoadIdentity
AthensCairoMatrix>>loadIdentity
AthensCairoMatrix>>initialize
AthensCairoMatrix class(FFIStructure class)>>fromHandle:
AthensCairoMatrix class(ExternalStructure class)>>new
AthensCairoMatrixTest(AthensTransformTest)>>newTransform
[...]

PR #14452, opened 2023-08-08
[P11] Adding a nice error for loadSymbol

tesonep:
We need to print the module and the symbol name to improve readability of the errors
Trying to explain #14451


jecisc:
@tesonep This seems good and only Windows tests are failing, but I think it is what you're trying to debug? (But why is windows running on P11 and not P12?!)

jecisc (approved):

PR #14456, opened 2023-08-08
Make methods in classes in the category ‘Calypso-Browser-Toolbar’ take the display scale factor into account

Rinzwind:
This pull request makes the methods in classes in the category ‘Calypso-Browser-Toolbar’ take the display scale factor into account.


Ducasse:
This is strange the error is 
```
```

jecisc (approved):

PR #14457, opened 2023-08-08
Make methods in classes in the category ‘Calypso-Browser-UI’ take the display scale factor into account

Rinzwind:
This pull request makes the methods in classes in the category ‘Calypso-Browser-UI’ take the display scale factor into account.

### DIFF
--- a/src/Calypso-Browser/ClyActivityAnimationIconMorph.class.st
+++ b/src/Calypso-Browser/ClyActivityAnimationIconMorph.class.st
@@ -49,7 +49,7 @@ ClyActivityAnimationIconMorph >> initialize [
 		hResizing: #shrinkWrap;
 		vResizing: #shrinkWrap;
 		listDirection: #leftToRight;
-		cellInset: 2.
+		cellInset: 2 scaledByDisplayScaleFactor.
 	progress := 1.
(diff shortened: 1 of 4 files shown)


Rinzwind:
There was a mistake in the original commit (`ClyBrowserMorph>>#buildWindow` should not have been modified), I replaced the commit.

Ducasse:
Thanks!

Ducasse:
This is strange this build is also broken with the same error

```

Time class(Object)>>doesNotUnderstand: #millisecondsSince:

```

Rinzwind:
[Build 655 of the ‘Pharo12’ Jenkins job](https://ci.inria.fr/pharo-ci-jenkins2/job/Test%20pending%20pull%20request%20and%20branch%20Pipeline/job/Pharo12/655/) has the same error.

MarcusDenker:
So this means that either https://github.com/pharo-project/pharo/pull/14447 or https://github.com/pharo-project/pharo/pull/14449 introduced the problem. But odd that they did not fail when the CI checked them

MarcusDenker:
I close and re-open to rebuild with the revert applied

MarcusDenker:
works now!

jecisc (approved):

PR #14458, opened 2023-08-08
Make methods in classes in the category ‘Calypso-Browser-Table’ take the display scale factor into account

Rinzwind:
This pull request makes the methods in classes in the category ‘Calypso-Browser-Table’ take the display scale factor into account.


jecisc (approved):

jecisc:
CI crashed early, I'm restarting it

PR #14461, opened 2023-08-09
Revert "RPackage: Clean bootstrap initialization"

MarcusDenker:
Reverts pharo-project/pharo#14449

Check if this is they cause of the broken build.

But this might be harder to fix as the PR merge itself was green, it was the *next* that was failing...


MarcusDenker:
failing test not related

PR #14462, opened 2023-08-09
RPackage: Clean bootstrap initialization

jecisc:
There is an initialization of packages in the bootstrap that is redoing some code in it own way instead of using the RPackage classic API. This make it harder to clean the API. 

In the past weeks I tried to unify some of the behavior. Now let's see if the class import can be unified.

Retry of https://github.com/pharo-project/pharo/pull/14449 after https://github.com/pharo-project/pharo/pull/14461

The goal is to see if I can reproduce the CI failure


jecisc:
Tests passed after merging the latest changes.

Gabriel-Darbord (approved):

MarcusDenker (approved):

MarcusDenker:
This seems to again make problems with losing a method. 

I will do a revert

PR #1461, opened 2018-05-30
22039-Introduce-displayScaleFactor

pavel-krivanek:

PR #14464, opened 2023-08-09
Make methods in the UITheme hierarchy take the display scale factor into account

Rinzwind:
This pull request makes methods in the UITheme hierarchy take the display scale factor into account.

### DIFF
--- a/src/Morphic-Widgets-Taskbar/UITheme.extension.st
+++ b/src/Morphic-Widgets-Taskbar/UITheme.extension.st
@@ -16,7 +16,7 @@ UITheme >> newBasicTaskbarButtonIn: aTaskbar for: aMorph [
 		newRowIn: aTaskbar
 		for: {((self iconNamed: aMorph taskbarIconName) ifNil: [^nil]) asMorph. lab}.
 
-	lm cellInset: 2.
+	lm cellInset: 2 scaledByDisplayScaleFactor.
 
 	button := self
 		newButtonIn: aTaskbar
--- a/src/Polymorph-Widgets/PharoDarkTheme.class.st
+++ b/src/Polymorph-Widgets/PharoDarkTheme.class.st
@@ -47,7 +47,7 @@ PharoDarkTheme >> buttonNormalBorderStyleFor: aButton [
 			  width: 0;
 			  baseColor: Color transparent ].
 	^ SimpleBorderStyle new
-		  width: 1;
+		  width: 1 scaledByDisplayScaleFactor;
 		  baseColor: self buttonColor
 ]
 
@@ -76,7 +76,7 @@ PharoDarkTheme >> configureWindowBorderFor: aWindow [
 	| aStyle |
 	aStyle := SimpleBorderStyle new
 		          color: self borderColor;
-		          width: 1.
+		          width: 1 scaledByDisplayScaleFactor.
 	aWindow borderStyle: aStyle
 ]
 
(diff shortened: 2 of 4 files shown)


Rinzwind:
Pull request for one method that is ‘NewTools’: https://github.com/pharo-spec/NewTools/pull/576

jecisc (approved):

jecisc:
Wouldn't it be simpler to manage the display scale factor in the Morph instead of the UITheme?

Rinzwind:
I’m not sure I understand what you have in mind, but maybe this helps: one problem to avoid is that values get double scaled, as would happen when passing the value of an accessor of one morph to a mutator of another morph if that mutator also scales it, a way to avoid that double scaling problem is to only scale constants.

Rinzwind:
I do think it would make sense to handle the display scale factor somewhere else though. A problem that exists with it now is that things don’t adapt when it is changed. Take this snippet which opens a PolygonMorph:

```smalltalk
(PolygonMorph
	vertices: (((10@10) scaledByDisplayScaleFactor extent: (30@30) scaledByDisplayScaleFactor)
		in: [ :rectangle |
			{ rectangle bottomLeft. rectangle topCenter. rectangle bottomRight } ])
	color: Color transparent
	borderWidth: 1 scaledByDisplayScaleFactor
	borderColor: Color black) openInWorld.
```

In a new Pharo image, after disabling `#showMenubar` and `#showDesktopLogo`, the morph looks like this:

<p></p>

If I change the world scale factor and the display scale factor to ‘Retina display’ scaling:

```smalltalk
World scaleFactor: 1/2.
WorldMorph displayScaleFactor: 2.
```

It does *not* automatically look like this:

<p></p>

For that, I have to do the first snippet again to open a new PolygonMorph. To make it automatic, I think it might make sense to handle the display scale factor in the canvas. I’m not sure if that has been considered before and if so, what problems that was expected to have. Perhaps someone who has worked on it before can answer that, the concept of display scale factor was introduced in commit 0b3ddf114554a8a1 as part of pull request #1461, and there are a few other pull requests and issues that mention [‘displayScaleFactor’](https://github.com/pharo-project/pharo/pulls?q=displayScaleFactor) or [‘display scale factor’](https://github.com/pharo-project/pharo/pulls?q=display+scale+factor), though many of them are the ones I opened over the past few months.

For an ImageMorph:

```smalltalk
(ImageMorph withForm: (ThemeIcons current iconNamed: #pharo))
	position: (10@10) scaledByDisplayScaleFactor;
	openInWorld
```

To automatically go from this:

<p></p>

To this:

<p></p>

I expect an intermediary between the ImageMorph and the Form would need to be introduced, so that a specific Form for each display scale factor can be used, like an object of a class Image which holds a Form for each display scale factor. This would be a bit similar to the separation between [NSImage](https://developer.apple.com/documentation/appkit/nsimage?language=objc) and [NSImageRep](https://developer.apple.com/documentation/appkit/nsimagerep?language=objc) in macOS’s AppKit I think.

For a TextMorph:

```smalltalk
TextMorph new
	contents: (Text string: 'Lorem ipsum'
		attribute: (TextFontReference toFont:
			(LogicalFont familyName: 'Source Sans Pro' pointSize: 15 scaledByDisplayScaleFactor)));
	position: (10@10) scaledByDisplayScaleFactor;
	openInWorld.
```

To automatically go from this:

<p></p>

To this:

<p></p>

I’m not sure whether there’s anything in particular to take into account.

PR #14469, opened 2023-08-10
Fix: make class abstract

balsa-sarenac:
Make class abstract was failing because `model` was `nil`. It's fixed by using model accessor instead of using it directly, since model is lazily initialised.

Besides, I've pushed-up `model` accessor to superclass, since it's the same for refactoring and transformation.


jecisc:
If there was a failure it might be nice to add a regression test :)

Ducasse:
@jecisc can you have a look because most of the broken tests are related to RPackageOrganizer and not related to this changes.

Ducasse:
None of the broken seems related to the changes. 
The changes are internal to RB engine.

PR #14471, opened 2023-08-10
RPackage: Automatically register packages in tests

jecisc:
RPackage tests have some helpers methods to help with the tests. One of them is here to create a new package. 

The problem is that it does not register the new package to the package organizer used for tests. So we have a lot of tests registering the packages by hand and other tests that do not register the package at all. But then the assertions can be wrong because some code is been executed in a state that does not represent the system.

Worst! In a lot of tests, we ended up with two packages of the same name. One registered in the organizer, and one orphan. This was caused by the creation of a class that ensured the requested package name was there and created a new package since the one we saved in a temporary variable was not in the organizer. This bit me a lot while manipulating the tests because I had two packages of the same name and this was hard to debug.

I found one test that really expect an unregistered package so I moved the package registration to the helper methods and adapted the tests. Either to remove the registrations or to fix the assertions that were wrong because of this. 

I also merged two tests and removed two others that had no assertion.

### DIFF
--- a/src/RPackage-Tests/RPackageOrganizerTest.class.st
+++ b/src/RPackage-Tests/RPackageOrganizerTest.class.st
@@ -60,21 +60,15 @@ RPackageOrganizerTest >> testAccessingPackage [
 
 	| p1 |
 	p1 := self createNewPackageNamed: 'P1'.
-	self organizer registerPackage: p1.
-	p1 addClassDefinition: Point.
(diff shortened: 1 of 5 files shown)


jecisc:
As a bonus I reduced the usage of categories in the tests to use first class packages

Gabriel-Darbord (approved):

PR #14475, opened 2023-08-11
Make Remove instance variables works well on a list using a new compo…

Ducasse:
…site.

Requires #14469 to be integrated first.

### DIFF
--- a/src/Calypso-SystemTools-FullBrowser/ManifestCalypsoSystemToolsFullBrowser.class.st
+++ b/src/Calypso-SystemTools-FullBrowser/ManifestCalypsoSystemToolsFullBrowser.class.st
@@ -1,0 +1,8 @@
+"
+Please describe the package using the class comment of the included manifest class. The manifest class also includes other additional metadata for the package. These meta data are used by other tools such as the SmalllintManifestChecker and the critics Browser
+"
+Class {
+	#name : #ManifestCalypsoSystemToolsFullBrowser,
+	#superclass : #PackageManifest,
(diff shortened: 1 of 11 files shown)


Ducasse:
Now that dependency has been introduce let us see.

Ducasse:
I do not get the error because the class it mentioned got rename. SycRemoveVariableCommand does not exist anymore. So to me this is an example of the bug with renaming a class that is extended. 
I do not know how to fix the problem :(
I will add this class again something to see if this is solving the problem else I'm blocked.

Ducasse:
Nothing changed. 

```
tonel:///builds/workspace/est_and_branch_Pipeline_PR-14475/64/src --- cache
Error: Package *Calypso-SystemTools-FullBrowser depends on the following classes:
  SycRemoveVariableCommand
  SycRemoveVariableCommand>>#fullBrowserMenuActivation
  SycRemoveVariableCommand>>#fullBrowserShortcutActivation
```

Ducasse:
Capture 2023-08-13 at 19 06 52

Ducasse:
I have no idea because in my image Calypso-SystemTools-FullBrowser does not extend SycRemoveVariableCommand as shown in screenshot.

I used the dependencyBrowser and look at all the dependent packages of Calypso-SystemTools-FullBrowser and I could not find SycRemoveVariableCommand. 

So I really do not understand.

Ducasse:
Adding back the class extensions did the trick.
We will see later how to remove this unused class.

PR #13906, opened 2023-06-05
Fix doubleClick on tab and on text selection.

kevca87:
This pull request addresses the following issues: #13577 and  #13830

Issue #13830, opened 2023-05-26
Text selection is too strict from the mouse location point of view


Ducasse:
In P11 we got a regression.
In the past we could select complete text by double clicking below the text now the mouse should be on the line of the latest character.

![Capture 2023-05-26 at 14 54 04](https://github.com/pharo-project/pharo/assets/432549/8bc193a5-b0fc-4ba3-a040-3fbf9b63145f)

Ducasse:
In the past the mouse could be below the last line and double click would select everything. 
Now we have to move the mouse on the line 'package: cat'

kevca87:
We found the source of the behavior with @akevalion.
We discover that @guillep change the behavior in [Fix wrong double click when second click outside of original bounds](https://github.com/pharo-project/pharo/commit/7780946856ff1c48bcf0e9ec36eec9b9a33b4bd7)
In the current implementation the `MouseClickState` class handle things outside of its responsibility like the `clickClient containsPoint:`.
It could be solved by changing `containsPoint:` outside of `MouseClickState`
Let us know your thoughts.

Ducasse:
Thanks a lot for your time and investigation.
Can you send a PR?

guillep:
Hi @kevca87, thanks for looking into this!

Could you give more details? I just patched something (actually two times) after a regression, so if you have a better implementation that solves all 2/3 issues, that's great!

akevalion:
Hi, we found that the method `MouseClickState >> #handleEvent:from:` It is too complex, it should be splited into smaller methods. 
About if the MouseClickState should propagate or not click events when they are or not in target morph. I propose that behaviour should be in each client morph, not in MouseClickState.

akevalion:
Hi can we close this issue? I because it was merged here:
- https://github.com/pharo-project/pharo/pull/13906

PR #14476, opened 2023-08-11
Treat a second click within the double-click window but in a different location as a separate click (fixes #14474)

### LINKED ISSUE #14474: Two clicks in rapid succession at different positions are treated as a double-click
**Bug description**
With the default 350ms double-click time, I sometimes make two mouse clicks quite far apart within the double-click window. This results in a double-click at the location of the first mouse-down, when I was expecting two separate clicks.

**To Reproduce**
1. Raise the double-click time to make this easier to see: `HandMorph doubleClickTime: 1000`.
2. Perform two separated clicks in rapid succession. For example, in a text editor, click on a word, then rapidly move the mouse and click on another word.
4. The first word is highlighted as if it had been double-clicked.

**Expected behavior**
The cursor moves to the second word with no highlight—no double-click event is triggered.

**Screenshots**

**Version information:**
 - OS: Any
 - Pharo version: Any

**Expected development cost**
I believe this can be fixed with a one-line change, PR incoming.

daniels220:
I chose to bail only on mouse _down_ outside the drag threshold, such that with a very long double-click time, you could click, move away, move back, and click again and this would still count as a double-click. My thinking is that for most of us it doesn't matter, because this is more-or-less impossible to do within 350ms. However someone with e.g. a hand tremor, who might need to set their double-click time much higher, might do this as part of normal usage. This matches the native Windows behavior, while macOS stops watching for a double-click on any movement even if you move back afterwards. I'd be fine with changing the condition to just `timedOut or: [isDrag]` if you think that's better though—like I said, it doesn't affect me either way.

One related question: Why is a double-click-and-drag detected with zero threshold? Why not use the same threshold as for a normal drag? In other words, why have `isDrag` and `isDragSecond` rather than just using `isDrag` all the time? Seems like a nice simplification that could be made...

### DIFF
--- a/src/Morphic-Base/MouseClickState.class.st
+++ b/src/Morphic-Base/MouseClickState.class.st
@@ -316,9 +316,11 @@ MouseClickState >> handleEvent: evt from: aHand [
 		^false].
 
 	clickState == #firstClickUp ifTrue:[
-		(timedOut or: [localEvt isMouseDown and: [isDrag]]) ifTrue:[
-			"timed out after mouseUp, or second click down far away from first click
-			either way breaks the possibility of a double-click - signal timeout and pass the event"
+		(timedOut or: [localEvt isMouseDown and: [
+				isDrag or: [localEvt buttons ~= firstClickDown buttons]]]) 
+		ifTrue:[
+			"timed out after mouseUp, or the mouse has moved, or this click is of a different button.
+			Can no longer have a double-click - signal timeout and pass the event"
 			aHand resetClickState.
 			self doubleClickTimeout. "***"
 			^true].
--- a/src/Morphic-Base/MouseClickState.class.st
+++ b/src/Morphic-Base/MouseClickState.class.st
@@ -279,6 +279,7 @@ MouseClickState >> handleEvent: evt from: aHand [
 			^true].
 		localEvt isMouseUp ifTrue:[
 
+			"If timedOut or the client's not interested in dbl clicks get outta here"
 			(timedOut or:[doubleClickSelector isNil]) ifTrue:[
 				self click.
 				aHand resetClickState.
@@ -289,7 +290,6 @@ MouseClickState >> handleEvent: evt from: aHand [
 			We will handle a click now."
 			clickState := #firstClickUp.
 			firstClickUp := evt copy.
-			"If timedOut or the client's not interested in dbl clicks get outta here"
 			aHand queuePendingEvent: firstClickUp.
 
 			self click.
@@ -316,8 +316,9 @@ MouseClickState >> handleEvent: evt from: aHand [
 		^false].
 
 	clickState == #firstClickUp ifTrue:[
-		(timedOut) ifTrue:[
-			"timed out after mouseUp - signal timeout and pass the event"
+		(timedOut or: [localEvt isMouseDown and: [isDrag]]) ifTrue:[
+			"timed out after mouseUp, or second click down far away from first click
+			either way breaks the possibility of a double-click - signal timeout and pass the event"
 			aHand resetClickState.
 			self doubleClickTimeout. "***"
 			^true].
--- a/src/Morphic-Base/Object.extension.st
+++ b/src/Morphic-Base/Object.extension.st
@@ -103,18 +103,18 @@ Object >> taskbarIcon [
 ]
 
 { #category : #'*Morphic-Base' }
-Object class >> taskbarIconName [
-	"Answer the icon for an instance of the receiver in a task bar"
-
-	^#smallWindow
-]
-
-{ #category : #'*Morphic-Base' }
 Object >> taskbarIconName [
 	"Answer the icon for the receiver in a task bar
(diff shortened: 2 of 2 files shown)


Ducasse:
Thanks for your contribution. 
We will have a look.
What you should see is that Morphic has not got that much love. 
Recently we are working on Toplo a totally new widget sets. If you are interested please contact me stephane.ducasse@inria.fr

daniels220:
Yeah, I'm aware people've been wanting to replace Morphic for...probably over a decade now. This particular issue is just such a massive usability problem for me, with such a simple fix, I figured I'd share it. I doubt I'll have any time to help with Toplo, but, good luck!

Ducasse:
Hi daniel
Sorry for the time to integrate. I was expecting that someone with a bit more knowledge than me (if any :))
will give us feedback. What I will do is trust you and we integrate it and we see if we get regression. 
Thanks you again for your contribution and interest in improving Pharo. We are all pushing (but we are a small team of highly dedicated people but we are super busy :)).

daniels220:
Thanks—I understand, this is kinda central and esoteric, isn't it! FWIW, the old behavior was bothering me several times an hour, and I've been using this fix since I submitted it and haven't noticed any new problems.

daniels220:
@guillep @Ducasse @kevca87 @akevalion Okay, I just stumbled across #13577, #13830, #13906, which are addressing the same problem—indeed the specific case of clicking on a tab then somewhere else and having the window maximize was the main thing that frustrated me. I initially implemented this fix in Pharo 11, before those were backported, and didn't notice when porting the changes over to Pharo12 to contribute that the issue had already been partially addressed.

That said, I think my solution here is better in that it is simpler, more general, matches the behavior of the native OS (I tested this—two clicks some distance apart but over the same element, within the double-click window, do not trigger a double-click on either macOS or Windows) and avoids morphs having to care. I think the `(self containsPoint: evt position) ifFalse: [ ^self ].` check in `Morph>>doubleClick:` etc. could now be removed, though this is a guess from first principles, not something I've tested.

@akevalion I do agree that `handleEvent:from:` is much too complex, btw, but I imagine this is covered under "Morphic isn't getting much attention these days".

ironirc:
Just to make sure, so you get the full picture, I also already addressed this problem.
Original issue #14168 
I had 2 PR's for this:  
- Pharo12 #14346 
- Pharo11 #14345 (not sure if this was pulled)

I solved it by altering the dispatched doubleClick event, (prior to my change, a MouseButtonEvent - mouseDown), to a  MouseButtonPseudoDoubleClickEvent which contains information about both clicks.
Consumers can then decide based on both clicks how this should be interpreted.
I only changed the behaviour on FTTableMorph>>doubleClick:
Apparently there was very similar problem with tabs, wich I didn't address.

daniels220:
Honestly I'm surprised this hadn't been addressed *long* prior to this, given how much it annoyed me, so in retrospect I'm not surprised that several people have had a go at it.

Honestly—I don't mean this to come across as arrogant, and feel free to disagree—I think my solution obviates both of these, and in doing so demonstrates why it's a cleaner solution overall: it solves the problem as close to the root as possible, so nobody else has to care. In the interests of keeping the codebase clean, I would personally suggest reverting the changes in #13577/#13906 and #14346. Does anyone object? I'll make a PR when I get a chance, which will be a more obvious opportunity for review, but might as well ask before taking the time to put that together.

(Technically there is an edge case where two clicks very close together can nevertheless be on different Morphs—a click at the bottom of one table row followed by one at the top of the next, or the right edge of one tab followed by the left edge of the next—such that the changes I'm suggesting reverting would still have an effect. But, this seems pretty unlikely to occur, and if it does, it's a toss-up whether treating it as a double-click is actually "wrong" from the user's perspective—it's equally likely that they just _almost_ missed their click target in the first place, but did intend to double-click it. It's surprisingly difficult to _intentionally_ perform two clicks like this—it's actually much easier, for me anyway, to perform a proper double-click without moving, or two clicks relatively far apart. So, I think the potential benefit of handling this edge case doesn't outweigh the additional code involved.)

(@ironirc I do like the concept of packaging up full information about a double-click into a single event, but FWIW if I were to do so I would do so by default, rather than requiring the Morph to request it, and drop the "Pseudo" from the class name—it would just be, "oh, yeah, double-clicks carry this extra information since more goes into them". Somewhat a matter of style I guess.)

jecisc:
If we have a general solution I’m in favor of removing the specific ones to keeps a cleaner code!

ironirc:
Do we want a solution that's correct "all the time" as opposed to "most of the time"?
How to decide wether clicks are relatively close or far apart?  Factors like screen resolution come into play, potentially requiring new user settings.
Consider double clicking on 1 row in a table morph, with clicks "relatively far" apart. The double click will fail.
When I contributed my solution, I was aware of at least one other similar issue, specifically concerning the tab bar.
Are there additional known areas in the codebase where this issue has manifested? 
This could help assess the best solution.

PR #14496, opened 2023-08-17
Removing references to text editor

Ducasse:
Trying....


Ducasse:
I tried to open all possible text and widgets to see if it generated errors.

PR #14497, opened 2023-08-17
SortedCollection should not implement #addLast: (fixes #9286)

### LINKED ISSUE #9286: Many SortedCollection methods return potentially mis-sorted results
Basically, lots of inherited methods use `self species` to decide what class of collection to produce, and for SortedCollection this is _sometimes_ appropriate, but very often not. For a very basic example:

```Smalltalk
#(4 1 7) asSortedCollection reversed "=> SortedCollection(7 4 1)"
"BUT:"
#(4 1 7) asSortedCollection reversed reSort "=> SortedCollection(1 4 7)"
```

Seems like `#reversed` should answer an `OrderedCollection`. Other methods affected include pretty much anything with `collect` somewhere in the selector _other than_ basic `#collect:` (which _is_ overridden to produce an `OrderedCollection`). It seems like, rather than override all of these, it would be ideal to dispatch the creation of a thing-we-will-`#collect:`-into to a method and then override just that. Not sure what to call it. `copyEmptyForTransformation` or `...ForCollect` is dumb but at least it's clear, but I'm open to better ideas.

A couple related notes:
* Seems like `SortedCollection` should override `#addLast:`, either with `self shouldNotImplement`, similar to how it disallows `#addFirst:`, or to redirect back to regular `#add:`. Personally I tend towards making it an error, but I'm not familiar with the existing design philosophy or existing uses.
* I see a comment in `#join:` noting that `#addAllLast:` doesn't re-sort—it will naturally follow the choice made re: `#addLast:`, though if we choose to have `#addLast:` re-sort, it would be a worthwhile optimization to re-sort only _once_ (which means calling `super addLast:` initially).
* As for #join: itself, I'm not familiar with that method, but I wonder if it should also fall back to an OrderedCollection?

daniels220:
#addLast: was used internally by the #select: and #collect: families, but it is also part of the public API of OrderedCollection, and in that role, it is not appropriate for a SortedCollection as it may leave the elements out of order.

Avoid near-exact duplication of 10 OrderedCollection methods by dispatching through two new ones:
* #addNoSort: which is explicitly private and expects the caller to maintain sort order (used in #select:, #reject: et al and SortedCollection>>addAll:).
* #copyForTransform[:], which answers a suitable collection for the output of a #collect: (used in, unsurprisingly, anything with 'collect' in the name, plus #reversed).

Also replace (typo'd?) #withIndexSelect:select: with correct implementation of #withIndexSelect:, and remove unreferenced #errorConditionNotSatisfied.

### DIFF
--- a/src/Collections-Sequenceable-Tests/SortedCollectionTest.class.st
+++ b/src/Collections-Sequenceable-Tests/SortedCollectionTest.class.st
@@ -503,6 +503,14 @@ SortedCollectionTest >> testAddAll2 [
 ]
 
 { #category : 'tests - basic' }
+SortedCollectionTest >> testAddLast [
+
+	self
+		should: [ #( 10 9 8 ) asSortedCollection addLast: 11 ]
+		raise: ShouldNotImplement
+]
+
+{ #category : 'tests - basic' }
 SortedCollectionTest >> testCollect [
 
 	|result aSortedCollection|
@@ -804,6 +812,11 @@ SortedCollectionTest >> testRemoveAll [
 
 	c1 add: 13; add: 14.
 	self assert: (c1 first = 14 and: [c1 second = 13]) description: 'the sortBlock has been preserved'
+]
+
+{ #category : 'tests - basic' }
+SortedCollectionTest >> testReversedIsOrdered [
+	self assert: self nonEmpty reversed class equals: OrderedCollection
 ]
 
 { #category : 'tests - basic' }
--- a/src/Collections-Sequenceable/OrderedCollection.class.st
+++ b/src/Collections-Sequenceable/OrderedCollection.class.st
@@ -269,6 +269,13 @@ OrderedCollection >> addLast: newObject [
 	^array at: lastIndex put: newObject
 ]
 
+{ #category : 'private' }
+OrderedCollection >> addNoSort: newObject [
+	"Add newObject to the end of the receiver. If the receiver is sorted, assume that this is the correct position--do not check sort order. Answer newObject."
+
+	^ self addLast: newObject
+]
+
 { #category : 'converting' }
 OrderedCollection >> asArray [
 	"Convert an OrderedCollection into an Array."
@@ -352,7 +359,7 @@ OrderedCollection >> collect: aBlock [
 	"(#(1 2 3) asOrderedCollection collect: [ :v | 10 ]) asArray >>> #(10 10 10)"
 
 	| newCollection |
-	newCollection := self species new: self size.
+	newCollection := self copyForTransform: self size.
 	firstIndex to: lastIndex do:
 		[:index |
 		newCollection addLast: (aBlock value: (array at: index))].
@@ -364,7 +371,7 @@ OrderedCollection >> collect: aBlock from: fromIndex to: toIndex [
 	"Override superclass in order to use addLast:, not at:put:."
 	| result |
 	self ensureBoundsFrom: fromIndex to: toIndex.
-	result := self species new: toIndex - fromIndex + 1.
+	result := self copyForTransform: toIndex - fromIndex + 1.
 	firstIndex + fromIndex - 1 to: firstIndex + toIndex - 1 do:
 		[:index | result addLast: (aBlock value: (array at: index))].
 	^ result
@@ -376,7 +383,7 @@ OrderedCollection >> collect: collectBlock thenSelect: selectBlock [
 
     | newCollection newElement |
 
-    newCollection := self copyEmpty.
+    newCollection := self copyForTransform.
     firstIndex to: lastIndex do: [ :index |
 		newElement := collectBlock value: (array at: index).
 		(selectBlock value: newElement)
@@ -411,6 +418,21 @@ OrderedCollection >> copyEmpty [
 ]
 
 { #category : 'copying' }
+OrderedCollection >> copyForTransform [
+	"As #copyForTransform, but using the default size for a new collection."
+
+	^self species new
+]
+
+{ #category : #copying }
+OrderedCollection >> copyForTransform: copySize [
+	"Answer a 'copy' of the receiver suitable for holding the output of a #collect: or similar, with sufficient capacity for <copySize> elements, but which contains no elements.
+	When transforming the elements of a SortedCollection, the result may no longer be sorted, or indeed sortable, so we provide a hook for it to override and return an OrderedCollection."
+
+	^self species new: copySize
+]
+
+{ #category : #copying }
 OrderedCollection >> copyFrom: startIndex to: endIndex [
 	"Answer a copy of the receiver that contains elements from position
 	startIndex to endIndex."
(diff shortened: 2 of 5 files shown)


daniels220:
I'd appreciate advice on how to add tests—I added the obvious one for `addLast:` itself, and one for `reversed`. I could do the same for `collect:`, but that wasn't actually broken, it was all the _others_, and it feels excessive to go through each and every method in that family and add a test for them, coming up with reasonable dummy conditions etc. The existing tests of those methods should cover _that they have been migrated_ to use `addNoSort:` and/or `copyForTransform[:]`, because they would throw an error if they had not—do you think that's assurance enough?

Ducasse:
In general we like simple tests that covers positive and negative aspects

src/Collections-Sequenceable-Tests/SortedCollectionTest.class.st
```diff
@@ -500,6 +500,14 @@ SortedCollectionTest >> testAddAll2 [
 	sorted2 do: [ :each | self assert: (sorted includes: each)]
 ]
 
+{ #category : #'tests - basic' }
+SortedCollectionTest >> testAddLast [
+
+	self
+		should: [ #( 10 9 8 ) asSortedCollection addLast: 11 ]
+		raise: ShouldNotImplement
```
Ducasse, line 510: Do we also have a positive test? 

| ar |
ar := #( 10 9 8 ) asSortedCollection.
ar addLast: 11.
self assert: ar asArray equals: #( 11 10 9 8)

daniels220: Uh, there isn't really a "positive" case—SortedCollection now `shouldNotImplement` `addLast:`, no exceptions. This is how other Smalltalk dialects do it, and `addFirst:` is already disallowed in Pharo. I suspect the only reason `addLast:` is _not_ already disallowed is that it is used internally by the `select:`/`collect:` family, and `SortedCollection>>addLast:`. The bulk of the changes in this PR are untangling that to use a separate, private API (`addNoSort:`—I'm not attached to the name btw) so that `addLast:` can be safely made `shouldNotImplement`.

My question as far as tests was about all the `collect:`-family methods I had to change that used to return a `SortedCollection` which might or might not be sorted (due to the actions of a `collect:` block), and thus might error later when inserting or re-sorting, which now return an OrderedCollection, analogous to regular `collect:`. I feel like trying to write tests asserting that they all return an OrderedCollection would be a tremendous amount of duplication for little gain, and given the change to `addLast:`, the _existing_ tests would fail if those methods had not been updated (because they would encounter the `shouldNotImplement` error). But I wondered if anyone with more familiarity with the test suite had an idea of something I could do that _would_ be worthwhile.

daniels220:
@jecisc What more work is needed? I don't have a clear path to proceed with tests.

src/Collections-Sequenceable-Tests/SortedCollectionTest.class.st
guillep: Yes, I also think the current PR is okay regarding tests. Cancelling `addLast:` has a positive test: it's cancelled and throws an exception :)

Though, I don't understand why `OrderedCollection` needs to implement `addNoSort:`?

daniels220: `OrderedCollection` implements `addNoSort:` so that methods that need that behavior on both Ordered and SortedCollection—anything with `select:` in the name—can have a common implementation. It is perhaps a little strange to have a method talking about not sorting on a class that doesn't perform any in the first place, but frankly those two are such a ubiquitous part of the standard library that I don't think anyone will have any trouble figuring it out. I chose to implement it this way—and similarly for `copyForTransform:`—despite feeling a little awkward about the names because it just saves _so much code_ that would otherwise be a near-exact copy.

src/Collections-Sequenceable/OrderedCollection.class.st
```diff
@@ -408,6 +415,21 @@ OrderedCollection >> copyEmpty [
 	^self species new
 ]
 
+{ #category : #copying }
+OrderedCollection >> copyForTransform [
```
dionisiydk: The name copy is misleading. It does not return a copy really. Maybe it can be another speciesForTransform (returning a class like a species)?

But maybe it's ok. I dont see the another name

daniels220: Ooooh, I _like_ that idea. You're right, it's not really a copy—I boxed myself in to that way of thinking because it's being used in place of `copyEmpty[:]`. Having a single `speciesForTransform` would eliminate the duplication between known- and unknown-size (`copyForTransform:` with a size argument and unary `copyForTransform` respectively) as well. Updated!

daniels220:
Looks like some "conflicts" got introduced—not actual conflicts but other changes to the file that Git couldn't merge. I resolved those (rebased to latest `Pharo12`) and cleaned up some category-formatting weirdness (did category names in source files recently change from symbol literals to strings, or is something messed up on my end? I made sure they match the rest of the file in what I committed here.) Any chance of getting this merged soon?

jecisc:
Hi,

the format was indeed updated recently to improve the gemstone compatibility. 

Values in tonel are now strings and the class export the package and tag property on top of the category.

Pharo is still able to read symbols as values.

daniels220:
Rebased and took care of the conflict. (Does my doing that ping people who've interacted with the issue? That's really the only purpose of leaving this comment...)

jecisc:
Hi,
Yes it pings people that accepts to receive notifications.

Currently there are 3 tests failing:

```
osx-64 / Tests-osx-64 / testJoinArrayUsingSortedCollection – MacOSX64.Collections.Abstract.Tests.SplitJoin.SplitJoinTest<1s
osx-64 / Tests-osx-64 / testKeysSortedSafely – MacOSX64.Collections.Unordered.Tests.Dictionaries.SmallDictionaryTest<1s
osx-64 / Tests-osx-64 / testKeysSortedSafely – MacOSX64.Collections.Unordered.Tests.Dictionaries.SmallIdentityDictionaryTest<1s
```

daniels220:
Hmmm...the SmallDictionary case is straightforward, just need to use `#addNoSort:`, which makes total sense given it's immediately followed by a `#reSort`. The `#join:` case...I would argue that `#(3) asSortedCollection join: #(1 2)` should answer `OrderedCollection(1 3 2)` not `SortedCollection(1 2 3)`—that the interleaved order of the result is an important part of the intent of `join:`, and this takes precedence over the usual preference to return the same class as the receiver. In this case using `speciesForTransform` would be appropriate. It's also worth noting that, with the current implementation, a SortedCollection with a custom sortBlock won't copy that sortBlock to the output of `join:`, which seems like the worst of both worlds—sorted, but _differently_ from the receiver? Or it's likely to error if the elements don't understand `#<` and the sortBlock was sorting on some property.

daniels220:
Looks like the last CI build broke before getting to tests—I don't _think_ it's related, can someone with permission get it to retry? Would be nice to get this merged.

Ducasse:
This is strange

```
MetacelloNotification: Loading baseline of BaselineOfPharoBootstrapProcess...
IceGenericError: error reading from the zlib stream
IceLibgitErrorVisitor>>visitGenericError:
```

I will relaunch it to check.

daniels220:
@Ducasse Just a quick reminder since I don't see a new CI run yet.

MarcusDenker:
Ci is now running

daniels220:
Huh, looks like the Unix build is basically okay, but macOS and Windows broke for new reasons. Again this doesn't _seem_ related but I can't know that for sure since I don't understand what _is_ going on. Let me know if I can help...

guillep:
Broken tests are not related, I'll integrate, thanks @daniels220 !

PR #14498, opened 2023-08-18
Change `Object>>#printStringLimitedTo:using:` to use ellipsis

Gabriel-Darbord:
By using the [ellipsis unicode character `…`](https://www.compart.com/en/unicode/U+2026), we can display a regular ellipsis instead of two dots, using one less character.

As a side note, `(limitedString truncateTo: limit - 3), '[…]'` could be optimized a bit to `(limitedString first: limit) replaceFrom: limit - 2 to: limit with: '[…]'` which makes one less copy.
However, I don't think this is critical code, so it can stay pretty and more readable.

### DIFF
--- a/src/Kernel/Object.class.st
+++ b/src/Kernel/Object.class.st
@@ -1727,13 +1727,13 @@ Object >> printStringLimitedTo: limit [
 
 { #category : #printing }
 Object >> printStringLimitedTo: limit using: printBlock [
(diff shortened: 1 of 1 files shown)


Gabriel-Darbord:
However, I realize that this turns the String into a WideString, which may not be ideal...

jecisc:
Indeed, wide strings are not as optimized as byte strings. I think it is better to keep byte strings

PR #14500, opened 2023-08-19
Fix: enable push down method for concrete classes

balsa-sarenac:
Push down method refactoring wasn't working on concrete classes. It would just say "Cannot push down method from non-abstract class".

This PR fixes the issue by splitting refactoring preconditions on applicability and breaking changes and also by changing the `generateChanges` so that it raises breaking-change-warning.

Because of this refactoring specific tests are extracted from transformation tests and I've added additional assertions for Driver API (`generateChanges`).


balsa-sarenac:
Failing tests seem unrelated

PR #14503, opened 2023-08-20
Add ‘scale’ parameter for displaying strings using a FreeTypeFont

Rinzwind:
This pull request adds a parameter ‘scale’ for displaying strings using a FreeTypeFont. An example is given by the snippet below. For comparison, for ‘form1’, the example string is displayed normally and then the form is scaled by a factor of two:

For ‘form2’, the string is displayed directly on the larger form with the point size of the font doubled, this provides better detail in the character glyphs, but also changes the width:

For ‘form3’, the string is displayed directly on the larger form using the new ‘scale’ parameter, which provides better detail in the character glyphs while keeping the width:

Before this can be merged: how should loading this into an existing image be handled? I tried it and while it did load, it required closing a debugger and resetting the draw error of some morphs afterwards (so `Morph allSubInstancesDo: #resumeAfterDrawError`).

Some other things to note:
* Only integer scales are supported.
* The FreeTypeCache is not used yet when the scale is not 1.
* The parameter is only available on FreeTypeFont, extending it to the other font classes and the string drawing methods on the canvas classes is left to ‘future work’.

```smalltalk
(baseForm := Form extent: 230@18 depth: Display depth)
	fillColor: Color white.
baseForm getCanvas drawString: String loremIpsum from: 1 to: 40 at: 5@0
	font: (LogicalFont familyName: 'Source Sans Pro' pointSize: 10) color: Color black.
form1 := baseForm scaledToSize: 230@18 * 2.

(form2 := Form extent: 230@18 * 2 depth: Display depth)
	fillColor: Color white.
form2 getCanvas drawString: String loremIpsum from: 1 to: 40 at: 5@0 * 2
	font: (LogicalFont familyName: 'Source Sans Pro' pointSize: 10 * 2) color: Color black.

(form3 := Form extent: 230@18 * 2 depth: Display depth)
	fillColor: Color white.
(port := form3 getCanvas privatePort)
	colorMap: nil;
	combinationRule: Form paint.
(freeTypeFont := (LogicalFont familyName: 'Source Sans Pro' pointSize: 10) realFont)
	installOn: port foregroundColor: Color black
	backgroundColor: Color transparent.
freeTypeFont displayString: String loremIpsum on: port from: 1 to: 40
	at: (5@0) * 2 kern: 0 baselineY: ((5@0) y + freeTypeFont ascent) * 2 scale: 2.

{ form1. form2. form3 } keysAndValuesDo: [ :index :form |
	form getCanvas drawPolygon: { 0@0. 459@0. 459@35. 0@35 } color: Color transparent
		borderWidth: 1 borderColor: Color black.
	PNGReadWriter putForm: form onFileNamed: FileLocator imageDirectory / (index asString , '.png') ]
```

### DIFF
--- a/src/FreeType-Graphics/FreeTypeGlyphRenderer.class.st
+++ b/src/FreeType-Graphics/FreeTypeGlyphRenderer.class.st
@@ -116,14 +116,15 @@ FreeTypeGlyphRenderer >> fixBytesForMono: aGlyphForm [
 ]
 
 { #category : #public }
-FreeTypeGlyphRenderer >> glyphOf: aCharacter colorValue: aColorValue mono: monoBoolean subpixelPosition: sub font: aFreeTypeFont [
+FreeTypeGlyphRenderer >> glyphOf: aCharacter colorValue: aColorValue mono: monoBoolean subpixelPosition: sub font: aFreeTypeFont scale: scale [
 
 	| form |
 	form := self
 		renderGlyph: aCharacter
 		depth: (monoBoolean ifTrue:[1] ifFalse:[8])
 		subpixelPosition: sub
-		font: aFreeTypeFont.
+		font: aFreeTypeFont
+		scale: scale.
 	monoBoolean
 		ifTrue:[
 			form := self fixBytesForMono: form.
@@ -133,14 +134,15 @@ FreeTypeGlyphRenderer >> glyphOf: aCharacter colorValue: aColorValue mono: monoB
 ]
 
 { #category : #public }
-FreeTypeGlyphRenderer >> mode41GlyphOf: aCharacter colorValue: aColorValue mono: monoBoolean subpixelPosition: sub font: aFreeTypeFont [
+FreeTypeGlyphRenderer >> mode41GlyphOf: aCharacter colorValue: aColorValue mono: monoBoolean subpixelPosition: sub font: aFreeTypeFont scale: scale [
 
 	| form |
 	form := self
 		renderGlyph: aCharacter
 		depth: (monoBoolean ifTrue:[1] ifFalse:[8])
 		subpixelPosition: sub
-		font: aFreeTypeFont.
+		font: aFreeTypeFont
+		scale: scale.
 	monoBoolean
 		ifTrue:[
 			form := self fixBytesForMono: form.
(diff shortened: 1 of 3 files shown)


Ducasse:
Thanks Kris. This is good to get your effort there. 
We will discuss to see how we can integrate it.

Ducasse:
May be we should split the PR in two. 
- One with all the added methods without calling any of them 
- Another one and there we may to see what initialize actions should be done.

Rinzwind:
I have spread the changes of the commit over three separate commits ([brach ‘font-scaling-alt’](https://github.com/Rinzwind/pharo/commits/font-scaling-alt)): fe39edfd97741ad0, cf1063019ec80570 and e76787793556482c. Note that that only avoids the problem if they are loaded one by one.

jecisc:
If the problem is just for the loading in an existing image then we should not spend too much time on it. The newly produced bootstraped image will have the changes and people can use this one instead of upgrading their image

Rinzwind:
I’m just closing and reopening to see whether `RubTextSegmentMorphTest>>#testComputeVerticesMultipleLines` fails again on a second build: it failed in the first one in the [‘unix-64’](https://ci.inria.fr/pharo-ci-jenkins2/job/Test%20pending%20pull%20request%20and%20branch%20Pipeline/job/PR-14503/1/testReport/Unix64.Rubric.Tests.Editing.Core/RubTextSegmentMorphTest/unix_64___Tests_unix_64___testComputeVerticesMultipleLines/) case, though not in the [‘osx-64’](https://ci.inria.fr/pharo-ci-jenkins2/job/Test%20pending%20pull%20request%20and%20branch%20Pipeline/job/PR-14503/1/testReport/MacOSX64.Rubric.Tests.Editing.Core/RubTextSegmentMorphTest/osx_64___Tests_osx_64___testComputeVerticesMultipleLines/) case.

Rinzwind:
The test did not fail in the [second build](https://ci.inria.fr/pharo-ci-jenkins2/job/Test%20pending%20pull%20request%20and%20branch%20Pipeline/job/PR-14503/2/testReport/), some other tests in ShClassInstallerTest and RBCopyPackageParametrizedTest did fail though, but I’m guessing that’s not because of the changes in this pull request.

Rinzwind:
I took a look at extending FreeTypeCache to also cache the glyphs for other scales than 1: commit bdf5d73fbfae13e4. I can either add that commit to this pull request or keep it for a next one.

Note that the commit introduces another migration problem for existing Pharo images. While developing, I temporarily added another statement after this one:

https://github.com/pharo-project/pharo/blob/bdf5d73fbfae13e47f4d3c76f9ca1a3830707cca/src/FreeType/FreeTypeCache.class.st#L100

The added statement took into account the previous structure wherein ‘typeTable’ contained the ‘entry’ instead of the ‘scaleTable’:

```smalltalk
scaleTable isDictionary ifFalse: [ typeTable at: typeFlag put: (scaleTable := self dictionaryClass new: 1) ].
```

That way, the method could be changed while not breaking the existing cache, which could then be cleared through `FreeTypeCache clearCurrent`. I could readd the statement if it’s necessary to allow loading in an existing image.

Ducasse:
So I will merge this one because the broken tests are not related.
And we can do another iteration

PR #14504, opened 2023-08-20
Change the #baseFont of the FixedFaceFont set in #fallbackFont in StrikeFont to the StrikeFont itself

Rinzwind:
This pull request changes the `#baseFont` of the FixedFaceFont set in `StrikeFont>>#fallbackFont` to the StrikeFont itself. Before this change, the snippet given below draws:

With this change, the snippet draws:

Note that multiple executions of the snippet use the same StrikeFont instance, it can be sent `#reset` to reset its fallback font.

The send of `#fallbackFont` in the snippet avoids a MessageNotUnderstood in `StrikeFont>>#displayMultiString:on:from:to:at:kern:baselineY:` due to `fallbackFont` being nil. I’m not sure whether that method should use the accessor or whether it should take into account that the variable can be nil in another way: compare with `#widthOf:` which uses the accessor, while `#glyphOf:` and `#glyphInfoOf:into:` use the variable directly with an `#ifNotNil:` test.

A larger question is whether using an instance of StrikeFont still needs to be possible, there don’t seem to be many uses anymore, though there are not none either, the `#realFont` of the `#monthNameFont` and `#weekdayFont` of a CalendarMorph is a StrikeFont.

```smalltalk
strikeFont := (LogicalFont familyName: 'Bitmap DejaVu Sans' pointSize: 9) realFont.
strikeFont fallbackFont.
(form := Form extent: 260@14 depth: Display depth)
	fillColor: Color white.
form getCanvas drawString: (String loremIpsum copyReplaceAll: 'u' with: 'ū') from: 1 to: 40 at: 5@0
	font: strikeFont color: Color black.
form
```

### DIFF
--- a/src/Graphics-Fonts/StrikeFont.class.st
+++ b/src/Graphics-Fonts/StrikeFont.class.st
@@ -1008,7 +1008,7 @@ StrikeFont >> extendMaxAsciiTo: newMax [
 
 { #category : #accessing }
 StrikeFont >> fallbackFont [
-	^ fallbackFont ifNil: [ fallbackFont := FixedFaceFont new errorFont fontSize: self height ]
+	^ fallbackFont ifNil: [ fallbackFont := FixedFaceFont new errorFont baseFont: self ]
 ]
 
 { #category : #accessing }


Ducasse:
I'm dull with fonts. So I cannot really answer but indeed if we can clean and get rid of one kind that we do not use. I'm all in favor.

PR #14508, opened 2023-08-21
Simplify FixedFaceFont by removing ‘displaySelector’

Rinzwind:
This pull request simplifies FixedFaceFont by removing ‘displaySelector’. The two selectors that were used both just displayed a string of the given ‘length’ filled with the ‘substitutionCharacter’.


jecisc:
The class comment still references this variable. Maybe it should be updated

Rinzwind:
I updated the comment.

jecisc (approved):

PR #14509, opened 2023-08-21
Packages: Introduce new API in MCOrganizationDefinition

jecisc:
MCOrganizationDefinition is the Monticello class used to describe packages and tags. 

Currently this class is based on 'categories'. This change adds an alternative that is to give a package name and tags.

Those are used to return the categories afterward so there should be no change in behavior, but it is a first step to use packages and tags instead. 

I started to replace some places using the categories to use the packages names and tag names. In future PR I'll use even more this new API but this will also need to update Tonel and Iceberg.

### DIFF
--- a/src/Monticello/MCOrganizationDefinition.class.st
+++ b/src/Monticello/MCOrganizationDefinition.class.st
@@ -5,7 +5,9 @@ Class {
 	#name : #MCOrganizationDefinition,
 	#superclass : #MCDefinition,
 	#instVars : [
-		'categories'
+		'categories',
+		'packageName',
+		'tagNames'
 	],
 	#category : #'Monticello-Modeling'
 }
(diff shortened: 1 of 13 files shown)


src/Monticello/MCOrganizationDefinition.class.st
```diff
@@ -46,12 +64,14 @@ MCOrganizationDefinition >> basicCommonPrefix [
...
-	^ categories
+
+	^ categories ifNil: [ { self packageName } , (self tagNames collect: [ :tagName | self packageName , '-' , tagName ]) ]
 ]
 
-{ #category : #accessing }
+{ #category : #deprecated }
 MCOrganizationDefinition >> categories: aCollection [
-	"ensure the categories are sorted alphabetically, so the merge don't take it as a conflict"
+	"Do not use. This methods should be replaced by #packageName and #tagNames"
```
Gabriel-Darbord, line 73: Deprecated but still used, even in this PR? In `MCPackage >> snapshot` and `MCStReader >> systemOrganizationFromRecords:`.
I assume `MCOrganizationDefinition class >> categories:` should also be deprecated.

jecisc: Yes, it will take some time to totally remove the usage. It’s the reason I just changed the protocol and I didn’t call the method #deprecated: , I do not want to raise warnings. I just want people not to introduce new senders to this method

Gabriel-Darbord (approved):

PR #14511, opened 2023-08-22
RPackage: Deprecate RPackageOrganizer>>#globalPackageOf:

jecisc:
This method does not do what it says it does and is not used nor useful. I propose to deprecate it


jecisc:
Included in #14517

PR #14512, opened 2023-08-22
Increase usage of new API in MCOrganizationDefinition

jecisc:
I recently introduced a new API on MCOrganizationDefinition to provide packages and tag names instead of categories. Here is an iteration to use this new API more.


jecisc:
Failing tests are related to refactorings and not this change

Ducasse:
Oops there is a conflict.

Gabriel-Darbord (approved):

PR #14513, opened 2023-08-23
RPackage: Remove duplicated classes infos

jecisc:
In RPackage classes are stored in two places: The packages and the tags.  This change proposes to simplify this by storing the classes only in the tags and making the packages ask for the classes to the tags.

I also started to update the class comment


jecisc:
I’m hoping that this PR will reduce the number of randomly failing tests but I’m not sure

src/RPackage-Core/RPackage.class.st
```diff
@@ -1,31 +1,25 @@
...
 ====================
-This class went over 3 internal representations implementation. 
+This class went over 4 internal representations implementation. 
 	- first: 	a list of class + two dictionaries: class * methods
 	This was not good since we add to go all the time over the list of classes.
 	- second: 4 dictionaries class * selectors 
 	This was not good since we want to have defined classes without methods. 
 	- third: 4 dictionaries + definedClasses
+	This was not good because it was way too complex
+	- Thourth: 2 dictionaries + tags including the classes
```
Gabriel-Darbord: Fourth*

src/RPackage-Core/RPackage.class.st
```diff
@@ -514,14 +500,14 @@ RPackage >> importProtocol: aProtocol forClass: aClass [
...
-	^ classes includes: aClass instanceSide
+	^ self definedClasses includes: aClass instanceSide
 ]
 
 { #category : #testing }
 RPackage >> includesClassNamed: aSymbol [
 	"Returns true if the receiver includes class named aSymbol in the classes that are defined within it: only class definition are considered - not class extensions"
 
-	^ classes anySatisfy: [ :class | class name = aSymbol ]
+	^ self definedClassNames includes: aSymbol
```
Gabriel-Darbord: This now requires making 2 intermediate collections. This could be reduced to 1 by only replacing `classes` by `self definedClasses` and keeping the previous `anySatisfy:`.

jecisc:
Thanks for the review. I fixed both

jecisc:
It seems to fix the random failures! :)

jecisc:
All failing tests are related to refactorings and not packages

PR #14516, opened 2023-08-23
Enh/cleaning breaking changes

Ducasse:
Many fixes :)


Ducasse:
I do not get it because it is green on our machines!

Ducasse:
And now there are two tests failing.

Ducasse:
Ah this is bug of RPackage not removing categories so I will push these changes.

PR #14495, opened 2023-08-17
Enh/browse in morphic pane

Ducasse:
- Improvements of browse/implementors in ClyTextEditor

PR #14521, opened 2023-08-24
Revert "Enh/browse in morphic pane"

Ducasse:
Reverts pharo-project/pharo#14495
I will keep these changes for me.

### DIFF
--- a/src/Calypso-SystemTools-FullBrowser/ClyTextEditor.extension.st
+++ b/src/Calypso-SystemTools-FullBrowser/ClyTextEditor.extension.st
@@ -3,24 +3,7 @@ Extension { #name : #ClyTextEditor }
 { #category : #'*Calypso-SystemTools-FullBrowser' }
 ClyTextEditor >> browseFullClass [
 	"self lineSelectAndEmptyCheck: [^ self]."
-	
-	[ 
-	self evaluateSelectionAndDo: [ :result |
-			| classToBrowse |
-			result isClass ifTrue: [ self browser browseClassNamed: result name. ^ self ].
-			
-			result isSymbol and: [ result first isUppercase 
-				ifTrue: [ classToBrowse := self class environment at: result ifAbsent: [ nil ].
-								classToBrowse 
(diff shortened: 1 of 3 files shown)


Ducasse:
No idea if it was running 404 instead

MarcusDenker:
Yes, the CI is somehow broken.

For the revert: yes, I would revert for now. The direction of the change is good, but we need to re-think it deeper.

- What should CMD-B do? It is now the same as "inspect it", just that instead of an inspector, it opens a browser on the class 
- the changes go towards a "browse defintion", which is good, I want that !
- But, the mechanism that implements "browse it" is implemented in a way that it works for "doIt and browse", but not for "browse definition". E.g. when nothing is selected, the editor selects (with an AST based heuristics) the next "wider" expression and executes that. The change tried to analyse the results of the execution to improve navigation, which works sometimes but not for all cases.

Ducasse:
Ok I will do any incremental improvements in the future on that topics. 
The rules are too strange for me Pharo world.

Ducasse:
Does not build because of 

I will do the revert by hand.

MarcusDenker:
it's just that "CMD-B" can not be extended this way to do navigation, as it *executes* code. If the (auto-selected) code is a string or a Symbol, we can do something with the result that is returned. But if the auto-selected code is a hole expression, the result will be the result of the expression, which can not be used for navigation.

e.g. if you have a symbol

```
#sin 
```

and put the cursor at the 'i', the editor will select the whole Symbol, executes it, which will be the symbol. So you can here take the result, analyse it, do senders of.

but imagine

```
1 sin 
```

if you out the cursor in the "i", the editor will select the whole "1 sin", execute it and return a number. There is no way to analyse the number to the browse all implementors of "sin". 

The nicest case are the for global, we have :

```
Object
```

the editor will select the String 'Object', eval it, return the class Object, and then open an browser on "Object class". Here "eval and browse class" is the same as "Browse defintion". But it is really the only case where "doIt and browse" is the same as "browse definition".

jecisc:
The CI passed. Failing tests are unrelated

PR #14522, opened 2023-08-24
Refactor: replace message sends

balsa-sarenac:
* Extract `RBReplaceMessageSendTransformation` outside of `RBChangeMethodNameRefactoring` hierarchy.
* Reuse replace-message-send inside change-method-name
* Remove code duplication around `buildSelectorString:...` methods
* Add tests for `buildSelectorString:withPermuationMap:andArguments`
* Simplified and remove a bunch of hooks from change-method-name hierarchy
* A bunch of small improvements and fixes


jecisc:
testNoShadowedVariablesInMethods seems to be a related failure

Ducasse:
This is a release test so we will fix it just after.
This PR was quite demanding. So I integrate it now and I will check this single problem later.

Issue #4632, opened 2019-09-16
FileLocator for music and video are missing


badetitou:
In the FileLocator, we have direct access to Documents, Downloads, images, etc. folders. 
However, the video and music entries are missing.

It's maybe not that much but it's important when you work with multiple OS.

PR #14532, opened 2023-08-25
Add FileLocator directories for music and video

hernanmd:
Add two locations to FileSystem library to locate the host system music and videos directories, as described in #4632.

Locations can be accessed through:

```smalltalk
FileLocator videos.
FileLocator music.
```

Add tests.


jecisc (approved):
Nice!

astares:
This PR will not work on Windows, there is no environment variable reflecting the music or video directory. 
Also CSIDL_COMMON_VIDEO is an id in C to access it via COM or Win32 API (CSIDL = "Constant Special Item ID List").

For instance once can use it with "SHGetFolderPath"  Win32 API function to find the folder path.

See #getWindowsFontFolderPath where this API is used for the font directory.

astares (changes requested):
The  code

```
 self directoryFromEnvVariableNamed: 'CSIDL_COMMON_VIDEO'
```
can not work as CSIDL_COMMON_VIDEO is not a legal environment variable. So this will **not work on Windows** (see comment above)

In Windows 10/11 the path typically is

C:\Users\<YourUsername>\Music
C:\Users\<YourUsername>\Videos

so you can construct it like

```Smalltalk
(PlatformResolver forCurrentPlatform 
  directoryFromEnvVariableNamed: 'USERPROFILE') / 'Videos'
```

hernanmd:
> The code
> 
> ```
>  self directoryFromEnvVariableNamed: 'CSIDL_COMMON_VIDEO'
> ```
> 
> can not work as CSIDL_COMMON_VIDEO is not a legal environment variable. So this will **not work on Windows** (see comment above)
> 
> In Windows 10/11 the path typically is
> 
> C:\Users<YourUsername>\Music C:\Users<YourUsername>\Videos
> 
> so you can construct it like
> 
> ```smalltalk
> (PlatformResolver forCurrentPlatform 
>   directoryFromEnvVariableNamed: 'USERPROFILE') / 'Videos'
> ```

Thanks for the feedback. I pused an update to use SHGetFolderPathA as you suggested.
I will check with a Windows box.

Ducasse:
So let us know when the PR is ready.

jecisc:
Hi @hernanmd , did you had the time to check by any chance?

hernanmd:
> Hi @hernanmd , did you had the time to check by any chance?

I will check tomorrow with the Windows box in the office and let you know.

tesonep:
This is an improvement but I have two comments:
1- We should use SHGetFolderPathW to handle wide characters in the path (You can see WinPlatform >> #getTempPath to see how it is better implemented)
2- We should use UFFI to perform the call, not using the backend directly. In the case of FT2 that is using it directly it is because it is used during the initialization of the image (maybe it should be pushed later when having UFFI loaded or it is somehting it is old, it does not have a lot of sense because FT2 uses UFFI)

hernanmd:
> This is an improvement but I have two comments: 1- We should use SHGetFolderPathW to handle wide characters in the path (You can see WinPlatform >> #getTempPath to see how it is better implemented) 2- We should use UFFI to perform the call, not using the backend directly. In the case of FT2 that is using it directly it is because it is used during the initialization of the image (maybe it should be pushed later when having UFFI loaded or it is somehting it is old, it does not have a lot of sense because FT2 uses UFFI)

I think by now we should use [SHGetKnownFolderPath](https://learn.microsoft.com/en-us/windows/win32/api/shlobj_core/nf-shlobj_core-shgetknownfolderpath) since SHGetFolderPathW was deprecated. I had a look and I wonder if in this case to pass the rfid as argument, one should use a FFIConstantHandle for the constant REFKNOWNFOLDERID?

guillep:
@tesonep @hernanmd There is no SHGetFolderPathW, the only version of this function accepts already wide strings as an out parameter seemingly. So yes, it needs to be decoded accordingly.

src/FileSystem-Core/WindowsResolver.class.st
```diff
@@ -18,13 +18,109 @@ WindowsResolver >> cache [
...
+				       tfBasicType pointer }
+		       returnType: tfBasicType slong.
+
+	buff := ByteArray new: 1024.
+
+	[ TFSameThreadRunner uniqueInstance invokeFunction: fun withArguments: { 0. 16r000d. 0. 0. buff } ]
+		on: Error
+		do: [ :e | "will get error if ffiplugin is missing" ^ self guessWindowsMusicFolderPath ].
+
+	^ (buff copyFrom: 1 to: (buff indexOf: 0) - 1) asString
```
guillep, line 47: Yes, this is no way to go. We cannot accept this. Please, check the FFI booklet and use the standard way of doing FFI in Pharo please.

src/FileSystem-Core/WindowsResolver.class.st
```diff
@@ -18,13 +18,109 @@ WindowsResolver >> cache [
...
+WindowsResolver >> getWindowsMusicFolderPath [
+	"Answer the windows music folder path.
+	This is obtained through the Windows API if FFI is present, otherwise it is a guess !"
+
+	| tfExternalFunctionClass tfBasicType fun buff |
+	tfExternalFunctionClass := Smalltalk globals at: #TFExternalFunction ifAbsent: [ ^ self guessWindowsMusicFolderPath ].
+	tfBasicType := Smalltalk globals at: #TFBasicType.
+
+	fun := tfExternalFunctionClass
+		       name: 'SHGetFolderPathA'
```
guillep, line 31: We should use the version without A as said in the comment by @hernanmd. And decode the wide string accordingly

src/FileSystem-Core/WindowsResolver.class.st
```diff
@@ -18,13 +18,109 @@ WindowsResolver >> cache [
...
+WindowsResolver >> getWindowsVideosFolderPath [
+	"Answer the windows videos folder path.
+	This is obtained through the Windows API if FFI is present, otherwise it is a guess !"
+
+	| tfExternalFunctionClass tfBasicType fun buff |
+	tfExternalFunctionClass := Smalltalk globals at: #TFExternalFunction ifAbsent: [ ^ self guessWindowsVideosFolderPath ].
+	tfBasicType := Smalltalk globals at: #TFBasicType.
+
+	fun := tfExternalFunctionClass
+		       name: 'SHGetFolderPathA'
```
guillep, line 60: Same as above.

Also, this is repeating all the code above with just an argument that changes.
This should be refactored in a single method with an argument.

src/FileSystem-Core/WindowsResolver.class.st
```diff
@@ -18,13 +18,109 @@ WindowsResolver >> cache [
...
+	^ (buff copyFrom: 1 to: (buff indexOf: 0) - 1) asString
+]
+
+{ #category : #origins }
+WindowsResolver >> guessWindowsMusicFolderPath [
+	"Guess the location of the Windows music folder"
+	| possibilities |
+
+	possibilities := Set new.
+	'cdefghijklmnopqrstuvwxyz' do: [ : drive |
```
guillep, line 85: This does not look well. Is this a standard way of doing this?

src/FileSystem-Core/WindowsResolver.class.st
```diff
@@ -18,13 +18,109 @@ WindowsResolver >> cache [
...
+]
+
+{ #category : #origins }
+WindowsResolver >> guessWindowsVideosFolderPath [
+	"Guess the location of the Windows videos folder"
+
+	| possibilities |
+
+	possibilities := Set new.
+	'cdefghijklmnopqrstuvwxyz' do: [ : drive |
```
guillep, line 101: Same as above

guillep: Also, this code looks repeated.

src/FileSystem-Tests-Core/FileLocatorTest.class.st
```diff
@@ -259,6 +259,13 @@ FileLocatorTest >> testMoveTo [
 	]
 ]
 
+{ #category : #'resolution tests' }
+FileLocatorTest >> testMusicDirectory [
+
+	locator := FileLocator music.
+	self assert: locator resolve equals: FileLocator music resolve
```
guillep, line 266: What is the purpose of this test?

It's testing that x resolve = x resolve...?
The assertion does not make sense to me.

Maybe just a smoke test is enough?

src/FileSystem-Tests-Core/FileLocatorTest.class.st
```diff
@@ -331,6 +338,13 @@ FileLocatorTest >> testSlash [
 	self assert: locator equals: (FileLocator image / 'griffle')
 ]
 
+{ #category : #'resolution tests' }
+FileLocatorTest >> testVideosDirectory [
+
+	locator := FileLocator videos.
+	self assert: locator resolve equals: FileLocator videos resolve
```
guillep, line 345: Same as above

guillep (changes requested):
This PR is far from ready.

Looking at the conflicts, I suggest this is redone from scratch.

src/FileSystem-Core/WindowsResolver.class.st
hernanmd: What's the real problem? (beyond being non-standard)

guillep: Hi, this **is** a real problem. It's nonmaintainable and error-prone. Using the unified FFI framework this code gets 
- closer to the original C call (in the sense that it will be written and read similarly)
- the marshaling is done automatically by a framework that was designed to do so
- same for optimizations, object pinning...

hernanmd: Again, why would I want to get closer to the original C call is there is an easier way to do it? In this case that would be also writing MORE code to get the [KNOWNFOLDERID](https://learn.microsoft.com/en-us/windows/win32/shell/knownfolderid) GUID, which I'm not interested at all. If I would have *another use case* for it, it is reasonable to invest in the translation.

At this stage I don't care if marshalling is done automatically by whom, and the same for optimizations. And asking for this type of stuff sounds like trying to force the same UFFI usage at all costs.

guillep: I think we are not on the same page. Do you want to schedule a call?

hernanmd: We could see it later since it is not urgent, thanks.

Ducasse:
Hi hernan

I should say that I do not like 
- your code
- your attitude

We do have to maintain code when people are moving to other horizons. 
So we CARE that the code is following our rules and frameworks. 
I do not even understand why we are discussing such obvious point. 

Guille is expert on FFI and he is dedicated to the improvement of Pharo. We should follow what he is saying. 
PERIOD. We should not make it lose his time. 

Sorry to be a be abrupt but I do not even understand why I have to make this point. 
So can you either close this PR or fix it?

hernanmd:
Thanks for the feedback. I would re doit from scratch, since I didn't took the right approach.
Let's check it later.

Issue #11236, opened 2022-05-16
Test Runner list selections are very slow


Shaping:
OS: Windows 10
Pharo version: 10 64-bit (stable)

1. Open Test Runner.
2. Select all packages, and note selection is not indicated for a very long time, about 20 seconds.
3. Select all TestCases, and note selection is not indicated for a very long time., about 20 seconds.

![image](https://user-images.githubusercontent.com/12146957/168661703-fa5f784a-fb7b-4c0c-b3b6-a134ef61fee6.png)

The above image shows a run of all tests in the image.  It took about 10 minutes.   The GUI should be more responsive.  There should also be some indication that the testing is ongoing besides the sometimes change test counts, which can be unchanging for long periods.  Would an hourglass cursor work?  The Test Runner window seems effectively modal, and responses happen in windows outside Test Runner.

jordanmontt:
We tried to reproduce this and we saw that this problem is only when you select all with `cmd + a`. If you right click and select "select all" it does not take time to load

hernanmd:
This issue can be closed now as #14534 has been merged into Pharo 12.

jordanmontt:
Yes, this is fixed on Pharo 12. I will close this issue

PR #14534, opened 2023-08-25
Speedup key combination for all items selection in Test Runner

hernanmd:
This PR fix the slow selection time in the Test Runner window, when pressing CMD+A key combination to select all test packages or test classes, as reported in #11236.

It add a key binding combination to the PluggableListMorph, which performs the same action as in the "Select all" menu option, for both packages and test unit classes.


jecisc (approved):

Ducasse:
Do you know that we will remove the old TestRunner UI?

Ducasse:
The problem with Milliseconds: is still there and breaking this build.

hernanmd:
> Do you know that we will remove the old TestRunner UI?

I didn't know, but I opened a new issue to work on this: #14540

PR #14537, opened 2023-08-25
RPackage: Simplify protocol manipulation

jecisc:
The goal of this PR is to simplify all the management of RPackage around protocols. 

It simplifies some logics and pushes the use of real protocols instead of protocol names. I also deprecated a method that had a really bad name and removed some dead code following my changes

I'm marking this PR as a WIP because I had some strange behaviors after doing those changes and I want to manually check the generated image


jecisc:
I'm removing the WIP because I think I found the origin of the weird behaviors

Ducasse:
Really difficult to follow this is why tests are key!

jecisc:
> Really difficult to follow this is why tests are key!

Indeed! A reason why this is difficult is that RPackage is way too complex for what it does. And it is why I'm doing so many PR that do not change the behavior but just simplify the code and try to keep the tests green. And I'm happy that we have so many tests around that part of the system (even if I think that some of them are not good tests. I'm cleaning them little by little)

PR #14538, opened 2023-08-25
RPackage: Do not repackage methods twice

jecisc:
RPackage is listening to the recategorization of methods to check if they need to repackage them. But this mecanism is duplicated and a second place is announcing method repackaging. 

The goal of this change is to try and remove one of those place because I think that the second one is enough for the repackaging


jecisc:
Failing tests are from refactorings and not packages

PR #14542, opened 2023-08-27
Remove StrikeFontSet and StrikeFontSetStoredSetting

Rinzwind:
This pull request removes StrikeFontSet and StrikeFontSetStoredSetting. Both classes have no instances and are not referenced, except for the message `#allSubInstancesDo:` sent to StrikeFontSet from `StrikeFont class>>#shutDown`. The method `#acceptSettings:` of StrikeFontSet also sends a message `#visitStrikeFontSet:` that has no implementors.


jecisc (approved):
Nice cleaning!

jecisc:
Sorry but we integrated this one too fast and we need to revert it because it breaks the bootstrap because of a dependency to the set that is still present in a setting

https://github.com/pharo-project/pharo/pull/14547

Rinzwind:
This seems to be due to a bug. In Pharo Launcher, create a new image using Pharo 12 build 697. Repair the repository ‘pharo’ through ‘Clone again […]’. Then repair it again through ‘Create a new branch […]’. In a System Browser, use ‘Remove’ in the context menu on StrikeFontSet, ignore the remaining reference. Use ‘Commit’ on the repository ‘pharo’: the window shows changes in ‘Graphics-Fonts’ only (1.png). Close the window. In a System Browser, add an empty line to `#settingStoreOn:` in LogicalFont. Use ‘Commit’ on the repository ‘pharo’ again: the window now shows the removal of the extensions of StrikeFontSet from ‘System-Settings-Browser’ (2.png). Note that it still doesn’t show the removal of the extension `#acceptSettings:` in ‘System-Settings-Core’.

**1.png**:


**2.png**:

guillep:
Hi, @Rinzwind I understand that there seems to be a bug in iceberg, but I did not understand what the bug is :)

Anyways, I want this PR in, we need to fix the issue and remove these unused classes!!! :)
Thanks for pushing this!

Rinzwind:
The problem seems to be that removing StrikeFontSet through the ‘Remove’ menu item only marks the package in which the class was defined as dirty, the two packages in which extensions to the class were defined are not marked as dirty, so the following does not hold (tested using Pharo 12 build 697):

```smalltalk
#('Graphics-Fonts' 'System-Settings-Browser' 'System-Settings-Core') allSatisfy: [ :name |
	((IceRepository repositoryNamed: 'pharo') packageNamed: name) isDirty ]
```

I made a new, corrected pull request (#14549).

jecisc:
I’ll add to my todo list to check this problem

Rinzwind:
OK. The error message mentioned in pull request #14547 confused me BTW, I don’t quite get why it refers to ‘Polymorph-Widgets-Traits’ and not ‘System-Settings-Browser’ as that is the package defining the `#settingStoreOn:` method:

```
Error: Package Polymorph-Widgets-Traits depends on the following classes:
  StrikeFontSet
  StrikeFontSet>>#settingStoreOn:
```

PR #14543, opened 2023-08-27
Fix bugs in display of wide characters through StrikeFont

Rinzwind:
This pull request fixes two problems in `#displayMultiString:on:from:to:at:kern:baselineY:` in StrikeFont: wide characters being put at the wrong vertical position, and a branch which is never executed as it would have caused an infinite loop otherwise. Note that there don’t seem to be many uses of instances of StrikeFont anymore, though there are still some, the `#realFont` of the `#monthNameFont` and `#weekdayFont` of a CalendarMorph is a StrikeFont.


jecisc (approved):

jecisc:
Thanks, it is nice to see such cleanings.

Failing tests are related to refactoring so the CI is good

PR #14548, opened 2023-08-28
RPackage: Iteration on tag API in RPackage

jecisc:
This change goal is to simplify the tag API of RPackage and make it more explicit.

* #addClassTag: was a bad name because it does not necessary add a tag. If the tag already exists, it will return the existing one. I propose to rename it #ensureTag: 
* #basicAddClassTag: is inlined in #ensureTag: 
* #classNamesForClassTag is deprecated since it has few utilities
* #basicRemoveTag: and #removeClassTag: were removed to keep only #removeTag:


Ducasse:
I like the ensureTag: idiom :)

PR #14555, opened 2023-08-29
Calypso: Prevent the recategorization of removed methods

### LINKED ISSUE #14553: Remove method doesn't update the protocols dictionary
**Bug description**
When you remove a method from a class it is still present in the protocol dictionary as 'as yet unclassified'.

**To Reproduce**
I've added a new method to `MyClassNonEmptyLeafUnused` and add some code to it. After I remove it the `ProperMethodCategorizationTest>>#testNoUncategorizedMethods` starts failing saying it has uncategorized methods. After debugging I've found that the method is still present in `protocols` dictionary, but is not present in the `methodDict`.

jecisc:
Calypso is listening to method changes to update the method pane. But doing this it forces the recategorization of methods in case the method has no protocol. This is happening when we explicitly remove a method from a protocol during a method removal. 

This change make sure we do not update the protocol of a method if its protocol is nil

Fixes #14553


balsa-sarenac (approved):
LGTM

PR #14557, opened 2023-08-29
RPackage: Remove categories from class removal announcement

jecisc:
This changes bring an update of ClassAnnouncement and especially ClassRemoved to reduce the usage of categories. 

Instead we are basing things on package tags. 

Depends on https://github.com/pharo-project/pharo/pull/14556


jecisc:
Done in another PR :)

PR #14561, opened 2023-08-30
RPackage: #packageTag should work on class side

jecisc:
#packageTag currently only works on instance side which is not coherent with #package that works on instance and class side. This change fixes this problems. 

The way I fixed the problem was to make RPackageTag>>#hasClass: work with class side. 

I also renamed #classTagForClass: into #tagOf: to have a better API. We have methods using 'classTag' and methods using 'tag' in RPackage and I want to align them to use 'tag'

Thank @pavel-krivanek for the bug report!

### DIFF
--- a/src/FluidClassBuilder/FluidBuilder.class.st
+++ b/src/FluidClassBuilder/FluidBuilder.class.st
@@ -92,15 +92,11 @@ FluidBuilder >> fillClassSideFromEnvironment: anEnvironment [
 { #category : #copying }
 FluidBuilder >> fillInstanceSideFromClass: aClass [
 
-	| tag |
-
 	slotsToBuild := aClass slots.
 	uses := aClass traitComposition.
 	self package: aClass package name.
 
-	tag := (aClass package classTagForClass: aClass).
-	tag ifNotNil: [ :s |
-		s isRoot ifFalse: [self tag: (aClass package classTagForClass: aClass) name ]]
+	aClass packageTag ifNotNil: [ :tag | tag isRoot ifFalse: [ self tag: tag name ] ]
 ]
 
 { #category : #building }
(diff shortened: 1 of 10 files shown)


pavel-krivanek:
I'm able to reproduce by creating a class method with a `script` pragma, Like:

```smalltalk
MessageCatcher class>>#msg

	<script>
	
	self yourself
```

and then do `
(MessageCatcher class>>#msg) browse`

Can you confirm?

jecisc:
Yes, and it works with this fix

src/FluidClassBuilder/FluidBuilder.class.st
```diff
@@ -92,15 +92,11 @@ FluidBuilder >> fillClassSideFromEnvironment: anEnvironment [
...
-	| tag |
-
 	slotsToBuild := aClass slots.
 	uses := aClass traitComposition.
 	self package: aClass package name.
 
-	tag := (aClass package classTagForClass: aClass).
-	tag ifNotNil: [ :s |
-		s isRoot ifFalse: [self tag: (aClass package classTagForClass: aClass) name ]]
+	aClass packageTag ifNotNil: [ :tag | tag isRoot ifFalse: [ self tag: tag name ] ]
```
badetitou, line 99: Can be optimized with isNotNil and:

src/Kernel/FluidClassDefinitionPrinter.class.st
```diff
@@ -401,7 +401,7 @@ FluidClassDefinitionPrinter >> slotsOn: s [
 FluidClassDefinitionPrinter >> tagOn: s [
 
 	| tag |
-	(forClass package classTagForClass: forClass) ifNotNil: [ :t |
+	forClass packageTag ifNotNil: [ :t |
 		tag := t name.
 		tag = forClass package name ifFalse: [
 			s crtab.
```
badetitou, line 407: Great job

badetitou (approved):
LGTM

PR #14565, opened 2023-08-31
Monticello: Announce MethodAdded when compiling a method whose prior method is from a different origin

jecisc:
Monticello has a duplication of the kernel code to install methods in the image. I tired to remove it but this is not so trivial.

In this duplication there is one part that is not synchronized: it does not manages traits. The problem arise if you compile a method in a class that already have this method but from a trait. Since there is a prior method we currently raise a MethodModified but in the case that the origin of the method is not the origin of the prior method, we should raise a MethodAdded instead.

This fix a bug in RPackage were all methods are not added to the packages.


badetitou (approved):
LGTM

Also discussed in person
Thanks

PR #14566, opened 2023-08-31
RPackage: Simplify method additions in tests

jecisc:
The tests of RPackage are too complex for what they are doing. I already started to simplify them and here is another step

If you want to add a method to a package currently we are silently compiling methods and since packages add the methods through announcements and the method was silently compiled, it has to add the method by hand in the package.

I replaced this by compiling normally the methods in the tests except for the tests that were explicitly adding method to a package


Ducasse:
is compile logging? Because compileSilently was used to avoid logging.

jecisc:
I hope in the future that it will not log. I need to work on that. But here the tests here are not really good with the silent compiling because it means the tests are not testing how it is done usually in the system.

PR #14567, opened 2023-09-01
Kernel: Introduce environment announcer

jecisc:
SystemAnnouncer is a class used by the system to anounce anything related to the system. It currently uses a singleton.

This prevents to have separated environments that are standalone. And this change is a first step to push in the direction of having standalone environments. I introduced an #announcer variable in SystemDictionary that is currentl the class we use to create an environment. This announcer can be accesed by any object by calling #environmentAnnouncer. For now, this variable points to the unique instance of SystemAnnouncer because the impact of this change is big and I want to do it step by step.

I updated the code announcing anything related to announcing pharo meta model changes. 

The only listener of SystemAnnouncer I changed was RPackage because it is been merged in the meta model. But I didn't update anything else because it will be a big change to know how we will manage all Pharo projects with multiple environments. So for now, the system will only work with the default system announcer. This will still be useful because we will be able to have new environments to do some tests generating code.

I also deprecated BlockClosure>>valueWithoutNotifications since it was not used and the method can easily be inlined and the code stay pretty simple

### DIFF
--- a/src/Deprecated12/BlockClosure.extension.st
+++ b/src/Deprecated12/BlockClosure.extension.st
@@ -1,0 +1,8 @@
+Extension { #name : #BlockClosure }
+
+{ #category : #'*Deprecated12' }
+BlockClosure >> valueWithoutNotifications [
+
+	self deprecated: 'This method will be removed from future version of Pharo since the meta model will not rely on the singleton anymore.'.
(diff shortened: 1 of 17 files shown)


guillep (approved):
LGTM :)

jecisc:
Closed in favor of another PR

PR #14575, opened 2023-09-04
Reduce UIManager references in Presenters + use new API

Ducasse:
- Now that inform: is defined in Application and Presenter use it
- One UIManager less in Presenter code


jecisc:
We have a lot of failing tests with `Message not understood: ByteSymbol >> #orDefaultForUnresolved`

Ducasse:
Tx pablo!

PR #14578, opened 2023-09-04
RPackage: Introduce null design pattern for UnpackagedPackage

jecisc:
When we are asking for the package of an unpackaged class, currently we are returning a package named '_UnpackagedPackage'. 

This change proposes to make UnpackagedPackage a class of its own that would implement the null object pattern to simplify the code. 

It should also introduce a change that is that unpackaged class should have a tag in the unpackaged package instead of returning nil because they are the only classes with a nil tag and this make it more complex to deal with tags. 

WIP


src/RPackage-Core/UnpackagedPackage.class.st
```diff
@@ -0,0 +1,186 @@
+"
+I represent the package of unpackaged classes. (Null object pattern)
+
+I will return empty collection for everything we can ask me and I'll be returned if we ask the package of an unpackaged class.
+"
+Class {
+	#name : #UnpackagedPackage,
```
dionisiydk, line 7: Maybe UndefinedPackage would be a better name. I always found the current name a bit weird

jecisc: For now I wanted to keep this name to make the change easier since the name is used in multiple places in the system

jecisc:
I'll do another one

PR #14579, opened 2023-09-04
RPackage: simplify class removal

jecisc:
This is a little change to simplify the class removal. 

Next step would be to automatically remove the empty tags when we remove a method in RPackageTag>>removeClass: but since this has more impact I'll do it in another PR later

### DIFF
--- a/src/RPackage-Tests/RPackageReadOnlyCompleteSetupTest.class.st
+++ b/src/RPackage-Tests/RPackageReadOnlyCompleteSetupTest.class.st
@@ -322,18 +322,13 @@ RPackageReadOnlyCompleteSetupTest >> testRemoveTaggedClasses [
 	self assert: (((p1 classesForClassTag: #zork) collect: [ :each | each name ]) includes: #B1DefinedInP1).
 
 	"now when we remove a class" "from an existing tags list"
-	p1 removeClassDefinition: a1 fromClassTag: #foo.
+	p1 removeClass: a1.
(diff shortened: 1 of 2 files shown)


src/RPackage-Tests/RPackageReadOnlyCompleteSetupTest.class.st
```diff
@@ -322,18 +322,13 @@ RPackageReadOnlyCompleteSetupTest >> testRemoveTaggedClasses [
...
+	p1 removeClass: a1.
 	self deny: (((p1 classesForClassTag: #foo) collect: [ :each | each name ]) includes: #A1DefinedInP1).
 	self deny: (((p1 classesForClassTag: #foo) collect: [ :each | each name ]) includes: #B1DefinedInP1).
 	self assertEmpty: (p1 classesForClassTag: #foo).
 
-	"from a nonexisting tag list"
-	p1 removeClassDefinition: b1 fromClassTag: #taz.
-	self deny: (((p1 classesForClassTag: #foo) collect: [ :each | each name ]) includes: #B1DefinedInP1).
-	self assert: (((p1 classesForClassTag: #zork) collect: [ :each | each name ]) includes: #B1DefinedInP1).
-
```
jecisc: I removed this because it is not possible to get in that case anymore

Gabriel-Darbord (approved):

Gabriel-Darbord (approved):

PR #14582, opened 2023-09-04
Kernel: Speedup class compilation

jecisc:
Class compilation always has been kinda slow. This speed improved at the beginning of Pharo 12 but reduced again last month. Here a new speed up. 

The first change preducing the speed up is to ask the package of a class by the class and not by the class name in FluidClassDefinitionPrinter. This has a side effect of breaking two tests in System-Changes-Tests but we should be able to bring back those tests to green with some other changes later.

The second change is to improve the way to know if a package includes a class.

It took me 15sec to compile 1000 classes before those changes (without any instance variable or anything special) and new it takes me less than 4sec.


jecisc:
This is not so easy since Epicea seems to rely on it too

jecisc:
I'll come at it again later

PR #14583, opened 2023-09-04
RPackage: Reduce logic in Class>>category

jecisc:
Class>>#category should be removed in the future but it will not be an easy change. This change aim to simplify it and see if it is possible to remove the part checking the system organizer classes


jecisc:
I'm exploring another path so I'll close this one for now

PR #14592, opened 2023-09-05
[WIP] Debug hack in category

jecisc:
Do not merge. This is just here to try and understand how we get into a hack in the categories


jecisc:
This is not needed anymore since I did a PR to remove all hacks and the CI is green :)

PR #14593, opened 2023-09-05
Use separate bind and listen primitives instead of combined listenOn:… family

### LINKED ISSUE #11723: `Socket>>#listenOn:backlogSize:` doesn’t signal an error if another socket is already listening on the same port
In the following example, no error is signaled; shouldn’t the second send of `#listenOn:backlogSize:` signal an error as the first socket is already listening on the same port?

```smalltalk
socket1 := Socket newTCP.
socket1 listenOn: 1701 backlogSize: 1.
socket1 isValid ifFalse: [ Error signal ].
socket2 := Socket newTCP.
socket2 listenOn: 1701 backlogSize: 1.
socket2 isValid ifFalse: [ Error signal ].
```

**Version information:**
- Image: `Pharo-11.0.0+build.224.sha.6fcd6c1aaf965912b63ffc2ef2059d10cf57c744 (64 Bit)`
- VM: `Pharo 9.0.18 built on Sep  2 2022 16:54:02 Compiler: 4.2.1 Compatible Apple LLVM 11.0.3 (clang-1103.0.32.29)`
- OS: macOS 11.7

daniels220:
Fixes https://github.com/pharo-project/pharo/issues/11723

Separate primitives wrapping the bind() and listen() syscalls already exist, and have correct error/return-code handling, unlike the combined primSocket:listenOn:backlogSize: implementation. We can implement the combined version in terms of the separate ones and remove the combined primitive, significantly reducing plugin code.

Also supply default values (interface and backlog size) in the  image, rather than rely on the primitive to accept varargs. This allows a user to better understand the behavior without access to VM source code. Supply reduced forms of the separate bind/listen messages as the home of these default values.

One question: Should there be an explicit socket state added for "bound but not listening or connected"? `listen()` will presumably fail on an unbound socket, so overall I'd prefer to _reduce_ dependency on such extra state not increase it, but I thought I should mention it...

### DIFF
--- a/src/Network-Kernel/Socket.class.st
+++ b/src/Network-Kernel/Socket.class.st
@@ -450,7 +450,18 @@ Socket >> address [
 
 { #category : #'initialize - destroy' }
 Socket >> bindTo: anAddress port: aPort [
-	^self primSocket: socketHandle bindTo: anAddress port: aPort
+	"Bind to the local port <aPort>, on the interface specified by <anAddress>
+	(`SocketAddress zero` specifies all interfaces).
+	Primarily used to prepare to listen for incoming connections with #listen[WithBacklog:]."
+
+	self primSocket: socketHandle bindTo: anAddress port: aPort
+]
+
+{ #category : #'initialize - destroy' }
+Socket >> bindToPort: aPort [
+	"Bind to the local port <aPort>, often in order to listen for incoming connections."
+
+	self bindTo: SocketAddress zero port: aPort
 ]
 
 { #category : #'private - errors' }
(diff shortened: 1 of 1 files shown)


daniels220:
This is intended to go with removing the `primSocket:listenOn:backlogSize:interface:` primitive family from the VM/plugin, perhaps as part of [pharo-vm#604](https://github.com/pharo-project/pharo-vm/pull/604), but it will work without any VM changes, just leaving those primitives unused.

Oh, and I realize I didn't make it clear in the commit message that the error-handling that exists in the separate primitives means this fixes #11723.

svenvc:
Nice!

Ducasse:
Hi thanks for this contribution I flagged it for Pablo and he will have a look.

guillep (approved):
LGTM, Thanks!

jecisc:
Since Guille reviewed it and tests passed I'm merging it

PR #14595, opened 2023-09-05
Fix check already-existing extension in Path>>withExtension: (fixes #14594)

### LINKED ISSUE #14594: `Path>>withExtension:` won't add an extension to filenames that end with those letters
**Bug description**
`Path>>withExtension:` checks if the receiver already has the supplied extension by asking if it `endsWith:` the provided string. But this could be a file whose _name_ ends with those letters, and has no extension at all, e.g. `footxt`. In this case we should produce `footxt.txt`.

**To Reproduce**
1. `'footxt' asFileReference withExtension: 'txt'`
2. Get `footxt`, still no extension

**Expected behavior**
`footxt.txt`

**Version information:**
 - OS: Any
 - Any Pharo Version

**Expected development cost**
Just need to add the extension delimiter to the `endsWith:` check, i.e. check if the name ends with `.txt` not just `txt`. PR incoming.

daniels220:
Check for delimiter followed by the requested extension to distinguish 'footxt' from 'foo.txt'--the former does not have an extension and so `(Path * 'footxt') withExtension: 'txt'` should become `'footxt.txt'` rather than being left with no extension at all.

Also rename some tests to correct "exten*t*ion" typo.

### DIFF
--- a/src/FileSystem-Path/Path.class.st
+++ b/src/FileSystem-Path/Path.class.st
@@ -585,13 +585,12 @@ Path >> segments [
 
 { #category : #navigating }
 Path >> withExtension: extension [
-	| basename name |
-	basename := self basename.
-	^ (basename endsWith: extension)
-		ifTrue: [ self ]
-		ifFalse:
-			[name := basename copyUpToLast: self extensionDelimiter.
-			self withName: name extension: extension]
+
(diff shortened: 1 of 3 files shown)


jecisc (approved):
Great catch! :)

Issue #10136, opened 2021-10-08
Deadlock on Exception with ensure:


demarey:
**Bug description**
Deadlock on Exception with ensure:
It freezes because the UI process is waiting on the semaphore that is never signaled.
However it should be signaled because it is in an **ensure* block.

**To Reproduce**
```smalltalk
s := Semaphore new.
[ [1/0] ensure: [s signal.] ] forkAt: Processor userInterruptPriority.
s wait.
```
WARNING: image freezes!

**Version information:**
 - Pharo Version : 9 and 10

Ducasse:
Ouch!

dionisiydk:
Nothing criminal here :)
UI process waits on semaphore and Cmd+. recovers it. It opens two debugger as expected: from the interrupt and from the ZeroDivide error.

The ensure signal will only happen when the block will exit. But the error is not the condition for this. Smalltalk exceptions do not unwind the context immediately allowing such thing like retry, resume and other interesting things (debugger itself is part of it).
So during ZeroDivide error the unhandledAction of the exception triggers the debugger. It schedules new UI action to open the debugger window. But because UI process is locked on semaphore nothing happens: no such action is processed.

It's not that easy to improve really. In this scenario the debugger needs to spawn new UI process. But it can lead to multiple UI processes. And the current environment will be by this.

guillep:
Hi @dionisiydk,

Hot really.

Here, Cmd+.  does not recover it because the UI process is suspended on the semaphore, and Cmd+. will only break running/ready processes (it's thought to stop infinite recursions and similar). However, being suspended on a semaphore does not count as running nor ready, and since that semaphore is never signaled, then  the process never reaches that state.

Moreover, there is a more deep thing going on here. Reading the code, one would expect that the semaphore will be signaled. If people need to understand how the debugger works to make sense of this problem, then this is not right. Also, is this behaviour different in headless, from the command line?

If the debugger is making programs less reliable and more difficult to reason about locally, then this is not a good behavior.
This is clearly a bug to me.

dionisiydk:
Hi @guillep 

> Here, Cmd+. does not recover it because the UI process is suspended on the semaphore, and Cmd+. will only break running/ready processes (it's thought to stop infinite recursions and similar). However, being suspended on a semaphore does not count as running nor ready, and since that semaphore is never signaled, then the process never reaches that state.

[...]

PR #14598, opened 2023-09-05
When opening a Debugger, spawn a new UI process if the existing one may be blocked

daniels220:
Specifically, if it is waiting on a Semaphore that is not part of a Delay, and thus may be dependent on the operation that is about to be debugged (e.g. due to waiting for an HTTP response from an in-image web server, a background operation in a threadpool, etc).

Addresses the unexpected/problematic part of the behavior mentioned in #10136

I admit, this is a pretty messy/concerning thing to do, but it's a massive improvement in usability for certain cases. Is it at least safe to spawn a new UI process while the existing one is potentially still running (rather than for sure being debugged)?

### DIFF
--- a/src/Debugger-Oups/OupsDebuggerSystem.class.st
+++ b/src/Debugger-Oups/OupsDebuggerSystem.class.st
@@ -134,16 +134,27 @@ OupsDebuggerSystem >> spawnNewUIProcess [
 
 { #category : #helpers }
 OupsDebuggerSystem >> spawnNewUIProcessIfNecessary: aDebugRequest [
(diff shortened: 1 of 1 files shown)


dionisiydk:
Hold on @daniels220 . There is a general fix for the problem waiting for several year now: [#11455](https://github.com/pharo-project/pharo/pull/11455).

daniels220:
Aha, I missed that, probably because it doesn't mention the issue I started from, #10136. Looks like it shares many of the same ideas as far as process state, but yes, takes a more general approach. We can close this, that solution definitely has the potential to be superior.

PR #14602, opened 2023-09-05
Allow formatting the contents of a Playground

daniels220:
Move the fallback formatter for packages without a manifest to exist on UndefinedObject so it is available when even the class is unknown. Delegate parsing for the purpose of formatting to existing RubEditingMode>>parseSource:.


jecisc (approved):
Yes! 
Thank you, I wanted that since a long time :)

PR #14604, opened 2023-09-06
Search strings without symbols using RB pattern matcher

hernanmd:
Add #isString method to RBLiteralValueNode, to enable matching method with strings** but not symbols.
Add tests.

One use of this feature is for instance to browse all the strings in a class which are not symbols.

### DIFF
--- a/src/AST-Core/RBLiteralValueNode.class.st
+++ b/src/AST-Core/RBLiteralValueNode.class.st
@@ -67,6 +67,12 @@ RBLiteralValueNode >> evaluateForReceiver: aReceicer [
 	^ value
 ]
 
+{ #category : #testing }
+RBLiteralValueNode >> isString [
+
+	^ value isString and: [ value isSymbol not ]
+]
+
 { #category : #accessing }
 RBLiteralValueNode >> sourceText [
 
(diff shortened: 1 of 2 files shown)


MarcusDenker (approved):

jecisc (approved):

src/AST-Core/RBLiteralValueNode.class.st
```diff
@@ -67,6 +67,12 @@ RBLiteralValueNode >> evaluateForReceiver: aReceicer [
 	^ value
 ]
 
+{ #category : #testing }
+RBLiteralValueNode >> isString [
```
dionisiydk, line 71: Having isString true for non String classes is a bad idea. It breaks the way to check that the object is a kind of string. I dont mean a kind of String class but the object implementing its protocol. I bet we have many users of isString assuming this semantics

hernanmd: We could rename it as isValueString?

guillep: or isLiteralStringNode?

Issue #14540, opened 2023-08-26
Remove the old TestRunner UI


hernanmd:
**Description**

In Pharo 12 most of the features of the old TestRunner (CMD+U) seem to be covered by the DrTest runner. So the task is to remove the old TestRunner from Pharo 12, and update the references to point to the DrTest runner.

Ducasse:
Yes let us remove it! and improve DrTests

PR #14607, opened 2023-09-06
TestRunner removal

hernanmd:
Completely remove TestRunner tool, as discussed in #14540
Adapt menu items and tool registration


jecisc:
There is an error:

UndeclaredVariableRead: Attempt to read undeclared variable DrTests
UndeclaredVariable>>runtimeUndeclaredRead
BaselineOfBasicTools>>postload:package:

The class is maybe not loaded yet?

hernanmd:
> There is an error:
> 
> UndeclaredVariableRead: Attempt to read undeclared variable DrTests UndeclaredVariable>>runtimeUndeclaredRead BaselineOfBasicTools>>postload📦
> 
> The class is maybe not loaded yet?

True. I moved the registration code to the post-load of DrTests baseline.
Let's wait the CI.

hernanmd:
The build is green though is shown as aborted (?)

jecisc:
The build pass. It has been month that windows is not working well

jecisc (approved):
Im approving and I am supporting this change. We need to push the last missing features of DrTest and it will not happen if we don't use it. And on top of that the old test runner was based on categories and morphs which we want to remove.

I'll not merge it myself because this is a big change, I want more reviews :)

hernanmd:
Sure, this would be a perfect call for @MarcusDenker and @Ducasse ? :)

PR #14609, opened 2023-09-06
Monticello: avoid double loading of class definition

### LINKED ISSUE #8204: MCPackageLoader>>#basicLoadDefinitions loads classes twice

jecisc:
This is a redo of https://github.com/pharo-project/pharo/pull/8282

I tried a more agressive approach to see if this work. If it does not work, I'll try a less agressive one.

### DIFF
--- a/src/Monticello/MCPackageLoader.class.st
+++ b/src/Monticello/MCPackageLoader.class.st
@@ -217,7 +217,13 @@ MCPackageLoader >> sorterForItems: aCollection [
 
 { #category : #private }
 MCPackageLoader >> tryToLoad: aDefinition [
-	[aDefinition addMethodAdditionTo: methodAdditions] on: Error do: [errorDefinitions add: aDefinition].
+	"Here we are only interested by method definition. We want to avoid recompilating classes or traits for example."
+
+	aDefinition isMethodDefinition ifFalse: [ ^ self ].
+
+	[ aDefinition addMethodAdditionTo: methodAdditions ]
+		on: Error
+		do: [ errorDefinitions add: aDefinition ]
 ]
 
 { #category : #public }


MarcusDenker (approved):

chisandrei:
@jecisc  In the end in GT we are using  the version below. One issue we got was when loading classes with traits as sometimes the class was loaded before the trait (or when using traits with a more complicated trait composition). This does not solve the issue, just attempts to reload again class definitions where the class is not in the system. To fully fix it I think we'd need to compute the dependencies between classes/traits and ensure we compile them in the right order.

```
tryToLoad: aDefinition
	(aDefinition isClassDefinition and: [
		aDefinition actualClass notNil ]) ifTrue: [ ^ false ].

	[aDefinition addMethodAdditionTo: methodAdditions] on: Error do: [errorDefinitions add: aDefinition].
```

jecisc:
Thanks, we have Famix that has a lot of complex trait composition so I’ll check next week probably with this project to see if we’re able to find some of those problems

PR #14610, opened 2023-09-06
Make unwind more flexible

guillep:
Stack unwind (correct management of exceptions and non-local returns) management requires the creation of many blocks, which makes it very slow.
This PR introduces a slight change in how unwind is managed that is backwards compatible and allows for custom unwind methods. See for example the code of ensure:

```smalltalk
BlockClosure >> ensure: aBlock
	| complete returnValue |
	<primitive: 198>
	returnValue := self valueNoContextSwitch.
	complete ifNil:[
		complete := true.
		aBlock value.
	].
	^ returnValue
```

The key of the previous method is that when unwind is done, the methods with the marker 198 are found in the stack and then:

-    if the antepenultimate temp (complete) is not nil
       - the first argument is taken (the handler block) and executed via the value message
       - and the antepenultimate temp (complete) is set to true

This PR makes that the handler block is instead put in the first temp, allowing the method to have as many parameters as possible. This simplifies proxy implementations among others. The only impact in the existing code base is as follows:

```smalltalk
ensure: aBlock
	| handler complete returnValue |
	<primitive: 198>
        >> handler := aBlock. <<
	returnValue := self valueNoContextSwitch.
	complete ifNil:[
		complete := true.
		aBlock value.
	].
	^ returnValue
```

This way, trap methods can be written as:

```smalltalk
Proxy >> trapWith: arg1 with: arg2
  | handler |
  handler := "some object that understands value".
  ...
```

This allows

- unwind without extra blocks and fewer contexts
- having a very fast execution path without unwinding in the middle

### DIFF
--- a/src/Kernel/Context.class.st
+++ b/src/Kernel/Context.class.st
@@ -2257,19 +2257,25 @@ Context >> tryPrimitiveFor: aMethod receiver: aReceiver args: arguments [
 { #category : #'special context access' }
 Context >> unwindBlock [
 	"unwindContext only. access temporaries from BlockClosure>>#ensure: and BlockClosure>>#ifCurtailed:"
-	^self tempAt: 1
+
+	"The handler is in the first temporary"
+	^ self tempAt: self numArgs + 1
 ]
 
 { #category : #'special context access' }
 Context >> unwindComplete [
 	"unwindContext only. access temporaries from BlockClosure>>#ensure: and BlockClosure>>#ifCurtailed:"
-	^self tempAt: 2
+
+	"The complete flag is in the antepenultimate temporary"
+	^ self tempAt: self numTemps - 1
 ]
 
 { #category : #'special context access' }
 Context >> unwindComplete: aBoolean [
 	"unwindContext only. access temporaries from BlockClosure>>#ensure: and BlockClosure>>#ifCurtailed:"
-	self tempAt: 2 put: aBoolean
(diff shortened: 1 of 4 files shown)


MarcusDenker (approved):
nice!

src/Kernel/Context.class.st
```diff
@@ -2257,19 +2257,25 @@ Context >> tryPrimitiveFor: aMethod receiver: aReceiver args: arguments [
...
 ]
 
 { #category : #'special context access' }
 Context >> unwindComplete [
 	"unwindContext only. access temporaries from BlockClosure>>#ensure: and BlockClosure>>#ifCurtailed:"
-	^self tempAt: 2
+
+	"The complete flag is in the antepenultimate temporary"
+	^ self tempAt: self numTemps - 1
 ]
```
dionisiydk, line 2271: Would it be better to reserve second temp for complete var? It sounds more intuitive that all reserved temps are in same order. 
Now it's a bit strange to reserve first temp and antepenultimate temp. Proxy with many temps will confuse

guillep: you mean having `^ self tempAt: self numArgs + 2` ?

I'm ok with that :)

dionisiydk:
Yep

чт, 7 сент. 2023 г., 09:42 Guille Polito ***@***.***>:

PR #14616, opened 2023-09-07
Add SixtyFourBitRegister class

hernanmd:
This PR introduces several changes to enhance the existing register implementation in the project:

### New Class: SixtyFourBitRegister
   
I've introduced a new class named SixtyFourBitRegister to support 64-bit registers. This new class is implemented similarly to the existing ThirtyTwoBitRegister class, allowing us to work with 64-bit values.

### Tests

Added comprehensive test coverage for both 32-bit and 64-bit registers. This includes tests for arithmetic operations, bitwise operations, value loading, rotation, and byte access, ensuring the correctness of these registers.

### Refactoring: FixedBitWidthRegister Superclass

To eliminate code duplication and improve maintainability, I've refactored common behavior into a new superclass called FixedBitWidthRegister. Both ThirtyTwoBitRegister and SixtyFourBitRegister now inherit from this superclass.

### Changes Made

- Introduced SixtyFourBitRegister class.
- Added test methods for SixtyFourBitRegister covering various operations.
- Refactored common behavior into the FixedBitWidthRegister superclass.
- Updated existing ThirtyTwoBitRegister tests to align with the changes and added additional tests for completeness.


jecisc:
osx-64 / Tests-osx-64 / testTestCasesEndsWithTestOrTestCase – MacOSX64.ReleaseTests.Release.ReleaseTest<1s
Error

A test class (subclass of TestCase) should have it's name ending with 'Test' or 'TestCase' Violations:  - BitRegisterTestSuite

hernanmd:
Updated

tesonep:
- Why they are not implemented with classes of WordLayout and DoubleWordLayout?
- Why using the two variables implementation, when we can store the value on the objects?
- Why not using the endianness of the machine, the generation of bigendian can be used when generating the bytearray representation.

tesonep (changes requested):
Why they are not implemented with classes of WordLayout and DoubleWordLayout?
Why using the two variables implementation, when we can store the value on the objects?
Why not using the endianness of the machine, the generation of bigendian can be used when generating the bytearray representation.

hernanmd:
> 

I tried using just a value, but partitioning the register seems to be faster so far:

```smalltalk
[ (SixtyFourBitRegister new
	loadFrom: #[18 52 86 120 154 188 222 240]
	 at: 1) asInteger ] bench.  
"22,572,466 iterations in 5 seconds 3 milliseconds. 4511786.128 per second"
"22,634,376 iterations in 5 seconds 1 millisecond. 4525970.006 per second"

[ (SixtyFourBitRegisterValue new
	loadFrom: #[18 52 86 120 154 188 222 240]
	 at: 1) asInteger ] bench.   
"15,834,692 iterations in 5 seconds 1 millisecond. 3166305.139 per second"
"15,727,297 iterations in 5 seconds 1 millisecond. 3144830.434 per second"
``` 

I will keep checking the implementation and let you know

hernanmd:
@tesonep here I attached an alternative implementation which does not use the partitioning into hi and low of the register bytes (if I understood your idea correctly).

Here in the benchmarks I ran most operations tend to be slower but you may check it out.

[BitRegisterValueTest.txt](https://github.com/pharo-project/pharo/files/12561826/BitRegisterValueTest.txt)
[SixtyFourBitRegisterValue.txt](https://github.com/pharo-project/pharo/files/12561820/SixtyFourBitRegisterValue.txt)

tesonep (approved):
Ok, let's integrate this version and then we can make a more performant with the required primitives exposed by the VM

hernanmd:
> Ok, let's integrate this version and then we can make a more performant with the required primitives exposed by the VM

Cool thanks!

PR #14623, opened 2023-09-08
RPackage: Update class category in class builder

jecisc:
When classifying a class in the class installer, the category is not updated. This is done later by RPAckage while listening to the announcements. But this creates a ping pong: Installer classify with the category -> RPackage listen and try to update the tag -> The tag notices a missmatch of categories -> the tag update the category -> the category is updated and an announcement rasied -> rpackage listen to it to update the tag -> since the category matches, the categorixation happens

I would like to simplify this by setting the category between the categorization and the raise of the announcement. 

Also, this will ensure the category of the class and the system organizer are in sync during the class installation which can simplify the state of the system

As a bonus I added an escape in Class>>category to manage in a better way the unclassified classes. Since a recent change I made, classes with unclassified category cannot be in a category of the system organizer. So we do not need to scan it which should speed things up in that case

### DIFF
--- a/src/Kernel/Class.class.st
+++ b/src/Kernel/Class.class.st
@@ -276,7 +276,10 @@ Class >> category [
 	(latter is much more expensive)"
 
 	| result |
-	self basicCategory ifNotNil: [ :symbol | ((self packageOrganizer listAtCategoryNamed: symbol) includes: self name) ifTrue: [ ^ symbol ] ].
+	self basicCategory ifNotNil: [ :symbol |
+		symbol = Class unclassifiedCategory ifTrue: [ ^ symbol ].
+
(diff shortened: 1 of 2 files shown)


Ducasse:
I imagine that you want to remove the category later anyway. No?

jecisc:
Yes I do! 
But in order to do that, I need to simplify the code managing categories. And if you check Class>>category, you can see that we have two hacks there. One checking if the category of the class is right because sometimes a class has a category that is wrong. And also checking if a class with a nil category does not have one in reality. Because sometimes a class has no category but is categorized :(

I want to avoid those hacks to make the system more consistent and make it easier to point something else later

jecisc:
With this change, I am hoping that the first hack is not needed anymore and that we can remove it totally :)

Issue #14533, opened 2023-08-25
Jump to test class should NOT automatically create a class if there is none


Ducasse:
It should ask first.

PR #14625, opened 2023-09-08
Confirmation dialog for "Jump to test class"

hernanmd:
This PR add a confirmation dialog to avoid command "Jump to test class" in the Calypso Browser to create automatically a class if there is none, as reported in #14533

### DIFF
--- a/src/Calypso-SystemPlugins-SUnit-Browser/ClyJumpToTestClassCommand.class.st
+++ b/src/Calypso-SystemPlugins-SUnit-Browser/ClyJumpToTestClassCommand.class.st
@@ -59,7 +59,7 @@ ClyJumpToTestClassCommand >> execute [
 
 	[ browser selectClass: (self testClassFor: targetClass) ]
 		on: ClyInvalidClassForTestClassGeneration
-		do: [ : ex | self inform: 'Cannot generate test class for ' , ex baseClass printString , '.' ]
+		do: [ :ex | self inform: 'Cannot generate test class for ' , ex baseClass printString , '.' ]
 ]
 
 { #category : #execution }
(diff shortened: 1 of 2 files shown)


src/Calypso-SystemPlugins-SUnit-Browser/ClyJumpToTestClassCommand.class.st
```diff
@@ -50,9 +50,16 @@ ClyJumpToTestClassCommand >> defaultMenuItemName [
...
+	self systemEnvironment 
+		classNamed: (self testClassNameFor: targetClass)
+		ifAbsent: [
+			(self confirm: 'Do you want to create a test class for: ' , targetClass asString )
+				ifFalse: [ ^ self ] ].
+
 	[ browser selectClass: (self testClassFor: targetClass) ]
 		on: ClyInvalidClassForTestClassGeneration
-		do: [ :ex | self inform: 'Cannot generate test class for ' , ex baseClass printString , '.' ]
+		do: [ : ex | self inform: 'Cannot generate test class for ' , ex baseClass printString , '.' ]
```
Ducasse: Can you remove this space? : ex -> :ex

Ducasse:
Broken tests are unrelated (take too much time)

src/Calypso-SystemPlugins-SUnit-Browser/ClyJumpToTestClassCommand.class.st
hernanmd: I have never understood this. For method parameters it is accepted to have the argument with spaces but for block parameters it isn't?

Ducasse: Let us not discuss that here. This is not because the syntax allows it that we should have strangely written method. 
99.9 % of the block are written [ :each and not [: each so let us make things simple uniform and easy.

PR #14637, opened 2023-09-11
RPackage: Clean package management in RPackageOrganizer

jecisc:
Here is a cleaning of the package management in RPackage organizer.

We now save an array of packages instead of a dictionary of package names and packages. I also cleaned some methods to rely less on the implementation details.

I did also some speed up


jecisc:
I'll close it for now since this is not so easy

PR #14651, opened 2023-09-13
Monticello: Add package and tag in class definition and start to migrate to it

jecisc:
Add Package and tag info in MCClassDefinition to push out the concept of categories. 

I might migrate more places later. I'd just like to see if the CI is happy with this for now because I found some wild bugs already in the image while doing that.

Here is a funny one: To stay compatible with other smalltalk when possible, if a class has no trait composition, the trait composition and class trait composition are not exported in MCZ format. But the way it is done is by exporting all instance variables except the last 2. Yes, the last 2. So if you add a instance variable or just change the order... You're f***ed :)

Next step is to update the FillTree reader in order to determine the package and the tag from the category depending on the name of the parent forlder (the format seems to always know the parent folder of the file so we can do it like this)

### DIFF
--- a/src/Monticello/MCTraitDefinition.class.st
+++ b/src/Monticello/MCTraitDefinition.class.st
@@ -24,16 +24,16 @@ MCTraitDefinition class >> name: classNameString traitComposition: traitComposit
 
 { #category : #comparing }
 MCTraitDefinition >> = aDefinition [
+
 	self flag: #traits. "Ugly we harcoded the super superclass method.  We will have to refactor the definition hierarchy"
-	
-	^ (aDefinition isKindOf: MCDefinition)
-		and: [(self isRevisionOf: aDefinition)
-		and: [self traitCompositionString = aDefinition traitCompositionString
-		and: [self classTraitCompositionString = aDefinition classTraitCompositionString
-		and: [category = aDefinition category
-		and: [self slotDefinitionString = aDefinition slotDefinitionString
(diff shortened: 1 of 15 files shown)


jecisc:
I added some commit to finish to migrate the users of #category: in the image. Now we also need to update Tonel. Once this is done, we will be able to remove the use of #category to use the package name and the tag name. We will also be able to update Tonel to remove more users of MCOrganizationDefinition>>#category:

src/Monticello/MCTraitDefinition.class.st
```diff
@@ -58,22 +58,21 @@ MCTraitDefinition >> classSlotDefinitionString [
...
+			         package: self category;
+			         beTrait ].
 
 
 	trait ifNotNil: [
 		trait comment: comment stamp: commentStamp.
-		self classInstanceVariables ifNotEmpty: [
-			trait classSide slots: self classInstanceVariables ]].
-	^trait
+		self classInstanceVariables ifNotEmpty: [ trait classSide slots: self classInstanceVariables ] ].
```
Gabriel-Darbord: `MCClassDefinition >> #classInstanceVariables` looks a bit complex, so its result could be given to the block.
```suggestion
		self classInstanceVariables ifNotEmpty: [ :variables | trait classSide slots: variables ] ].
```

Gabriel-Darbord (approved):

PR #14653, opened 2023-09-13
Monticello: update MCClassTraitDefinition to follow changes of MCClassDefinition

jecisc:
I recently updated MCClassDefinition and MCTraitDefinition to push two changes:
- Split the huge constructors to make the instance configurable
- Add #packageName and #tagName variables in order to allow to configure the definition with packages and tags instead of categories even if the code is still exploiting the #category method.

This change will allow me to update Tonel to use packages and tags and once this is done we will be able to remove the depend code of categories to really exploit the #packageName and #tagName variables.

### DIFF
--- a/src/Monticello/MCAddition.class.st
+++ b/src/Monticello/MCAddition.class.st
@@ -7,7 +7,9 @@ Class {
 	#instVars : [
 		'definition'
 	],
-	#category : #'Monticello-Patching'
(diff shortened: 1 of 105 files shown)


jecisc:
Hum... I did this in an image that already has my changes on Tonel which messed up the format. I'll try to fix that

PR #14068, opened 2023-06-19
Make finalization registry entry removal O(1)

guillep:
Turn the finalization registry into a double-linked list.
This turns removal into an O(1) operation while making ephemeron objects a bit bigger (2 inst vars for previous and next).

Issue #14626, opened 2023-09-09
File registry is broken and leaks files


noha:
**Bug description**
The way the File registry (FinalizationRegistry) is handling entries is bogus and leaks all files. 

**To Reproduce**
1. file := File openForWriteFileNamed: 'testfile'.
2. file close
3. testfile entry is still in the File registry

**Expected behavior**
After closing the file the entry should have vanished from the file registry

**Version information:**
 - OS:  MacOS but independent 
 - Pharo Version >= 11

**Expected development cost**
Estimated cost is 2-4 hours because the bug is identified just needs a proper solution

**Additional context**
The bug is there because when a file is opened it is added to the File registry with 

```
add: anObject finalizer: finalizer

	self protected: [ 
		ephemeronCollection add:
			(FinalizationRegistryEntry key: anObject value: finalizer container: self) ].
	^ anObject
```

but on closing of the file it does 

```
remove: anEphemeron ifAbsent: aBlock

	^ self protected: [ ephemeronCollection remove: anEphemeron ifAbsent: aBlock ]
```

where anEphemeron is the file stream object itself. Therefor the added object and the to be removed one are not the same therefor it is a miss in the ephemeronCollection and it returns without unregistering the file. I would have provided a fix directly instead of writing this issue but is unclear how in that context a fast removal should work with the wrapped object. So delegating to guille :)

Ducasse:
Hi Norbert 
Are you in P12? Because we got some changes in finalization so this is an important information.

jecisc:
Reading the description of Norbert, the change will not fix this particular issue I think. 
But the report of Norbert is really nice and it should not be that hard to fix the bug :)

noha:
@Ducasse I'm on p11, I did not check p12 just assumed it hasn't changed there. 

I don't get the rationale of having an IdentitySet for the ephemeron management which contains elements that are subclasses of Association where the key and value are both the stream to be managed. But yes everytime I read about ephemerons and understand it, 5 minutes later it is lost :) So I remember something that the association is needed but then it might be possible to use an IdentityDictionary instead of IdentitySet which would allow fast removal

guillep:
Thanks, I'll check it.

In the previous weeks I was working on some improvements on ephemerons, see https://github.com/pharo-project/pharo/pull/14068

Mostly they were performance improvements, but I remember passing through the file tests when doing that.

Now, I did those changes against Pharo12, with the idea to then backport to P11 if OK.
[...]

PR #14655, opened 2023-09-13
Make finalization registry entry removal O(1) -- backport P11

guillep:
Turn the finalization registry into a double-linked list.
This turns removal into an O(1) operation while making ephemeron objects a bit bigger (2 inst vars for previous and next).

Backport https://github.com/pharo-project/pharo/pull/14068 for Pharo11

### DIFF
--- a/src/Announcements-Core-Tests/WeakAnnouncerTest.class.st
+++ b/src/Announcements-Core-Tests/WeakAnnouncerTest.class.st
@@ -29,32 +29,11 @@ WeakAnnouncerTest >> benchManyWeakGuys [
 	^ time
 ]
 
-{ #category : #benchmarks }
-WeakAnnouncerTest >> benchWeakSubscriptionDynamic [
-	"Run me to see how expensive #becomeForward: is (compared to #benchWeakSubscriptionStatic )"
 
-	| object ann |
+{ #category : #utilities }
+WeakAnnouncerTest >> longTestCase [
 
-	object := Object new.
-	ann := Announcer new.
-
-	^[
-		1000 timesRepeat: [ (ann when: Announcement send: #value to: object) makeWeak ].
-	] timeToRun
-]
-
-{ #category : #benchmarks }
-WeakAnnouncerTest >> benchWeakSubscriptionStatic [
-	"Run me to see how cheap object creation is (compared to #benchWeakSubscriptionDynamic ) "
-
-	| object ann |
-
-	object := Object new.
-	ann := Announcer new.
-
-	^[
-		1000 timesRepeat: [ (ann weak when: Announcement send: #value to: object)  ].
-	] timeToRun
+	self timeLimit: 60 seconds
 ]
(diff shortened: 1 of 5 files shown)


noha (approved):

noha:
So it seems at the moment the case of a leaked file is optimized. Meaning a file that is not closed but just not kept as a reference anymore will be finalized and removed from the FinalizationRegisry in O(1). The normal case for a file being closed and therefor manually removed from the FinalizationRegistrty is O(n) due to the linear scan in the doubled linked list. And the solution to this would be to have FileHandle and BinaryStream have a reference to the finalization entry? Is that about right?

jecisc (approved):

jecisc:
For now I'm adding the "Need more work" tag because we are waiting for th vm 10.0.7 release

guillep:
> So it seems at the moment the case of a leaked file is optimized. Meaning a file that is not closed but just not kept as a reference anymore will be finalized and removed from the FinalizationRegisry in O(1). The normal case for a file being closed and therefor manually removed from the FinalizationRegistrty is O(n) due to the linear scan in the doubled linked list. And the solution to this would be to have FileHandle and BinaryStream have a reference to the finalization entry? Is that about right?

There are two different issues:
 - first, the issue adressed by this PR, is that removing an element from the finalization registry is slow when millions of elements are there. This is because the finalization registry uses an identity dictionary, and identity dictionaries hash objects based on their identity hash, which is limited by 22bits (4M possibilities). This means that at some point there are high chances of having hash collisions. And the current `IdentityDictionary` in Pharo solves collisions by allocating the following next bucket, assuming a linear search could be ok to continue the search. However, this can also easily degenerate. This PR proposes to not use an `IdentityDictionary` because of that. The issue here was particularly visible when using finalization stressing code like Iceberg's history GUI on a large repo such as Pharo. Iceberg then creates thousands of git objects with `autoRelease` and abuses finalization.
 - second, an issue in the VM (the issue with files being "leaked" -- see https://github.com/pharo-project/pharo/issues/14633 and yours in https://github.com/pharo-project/pharo/issues/14626) makes that GC does not trigger the finalization of all possible ephemerons when the ephemerons and keys are in the old space. This makes that a GC will finalize just a couple of ephemerons in some cases, and finalizing all ephemerons requires running the GC multiple (thousands of) times -- for the cases when we have a couple of thousands of ephemerons. This is being fixed in the VM version 10.0.7, that we are releasing today. See https://github.com/pharo-project/pharo-vm/pull/691

guillep:
It's worth noticing that all these issues are performance issues, none of these lead to crashes.

jecisc:
@guillep I think we can now merge?

guillep:
Yes. I've spent now around one hour testing the latest build of P12 with all the latest VM changes and the situation is much better.

I've found a glitch that I'd like to investigate a bit further but it's not a showstopper.

PR #14668, opened 2023-09-15
RPackage: Clean tests and reduce categories usage

jecisc:
This change is cleaning some behaviors in tests and also reducing the number of category manipulations


jecisc:
I'm preparing a better version

PR #14671, opened 2023-09-15
Use basicCategory in Ring instead of category

jecisc:
While creating a Ring definition from a Class or a Trait we are using the method #category. But in the case of classes that got updated or removed from the image, the method #category contains a hack that will try to check the category of the class of the same name in the image. Thus if a class get updated and you ask the ring definition of the old class, it will have the category of the new class.

This change make Ring be based on #basicCategory that is the category of the class even if a newer version of the class is in the image.


jecisc:
I was thinking of adding tests but the categories will soon be removed. I'm fixing this only to help with the removal

PR #14673, opened 2023-09-16
Refactor: rename class refactoring

balsa-sarenac:
Since both rename class refactoring and transformation had the same code one is redundant. This PR removes rename class transformation and migrates everything to rename class refactoring.
Additionally, this PR removes rename and deprecate class refactoring since it was unused.


Ducasse:
We decided to do something else :)

balsa-sarenac:
I've reverted the commit that removed `RenameAndDeprecateClassTransformation` and will expose that refactoring in a separate PR. I'm reopening this PR since we want this changes integrated before continuing.

jecisc (approved):

Ducasse:
Thank balsa!

PR #14676, opened 2023-09-16
Kernel: Make sure category is never nil

jecisc:
In Class>>category we have two hacks:
- One because a class can have a wrong category. I removed most code provoking this problem and the last one is in Fuel but there is a fix incoming
- One because a class might not have a category set but is already in a category

I fixed all the cases where the second case happened already. We just have a few cases were the category is nil instead of #Unclassified during the class installation. In order to be able to remove the nils there, I'm setting the Unclassified symbol at class creation.

### DIFF
--- a/src/Kernel/Class.class.st
+++ b/src/Kernel/Class.class.st
@@ -276,14 +276,14 @@ Class >> category [
 	category name stored in the ivar is still correct and only if this fails look it up
 	(latter is much more expensive)"
 
-	| result |
-	self basicCategory ifNotNil: [ :symbol |
-		symbol = Class unclassifiedCategory ifTrue: [ ^ symbol ].
+	| result cat |
+	cat := self basicCategory.
 
-		((self packageOrganizer listAtCategoryNamed: symbol) includes: self name) ifTrue: [ ^ symbol ] ].
+	cat = Class unclassifiedCategory ifTrue: [ ^ cat ].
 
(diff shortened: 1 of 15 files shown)


src/Kernel/Class.class.st
```diff
@@ -276,14 +276,14 @@ Class >> category [
...
-		symbol = Class unclassifiedCategory ifTrue: [ ^ symbol ].
+	| result cat |
+	cat := self basicCategory.
 
-		((self packageOrganizer listAtCategoryNamed: symbol) includes: self name) ifTrue: [ ^ symbol ] ].
+	cat = Class unclassifiedCategory ifTrue: [ ^ cat ].
 
+	"The next lines are only useful anymore because of a problem in Fuel that should be resolved in not too long. Then we will be able to just return #basicCategory in this whole method"
+	((self packageOrganizer listAtCategoryNamed: cat) includes: self name) ifTrue: [ ^ cat ].
 	result := (self packageOrganizer categoryOfBehavior: self) ifNil: [ #Unclassified ].
```
Gabriel-Darbord, line 286: Fixed since #14677?

Gabriel-Darbord (approved):

src/Kernel/Class.class.st
jecisc: Yes! I’ll do another PR soon to simplify the whole method :)

PR #14681, opened 2023-09-17
Kernel: simplify category management

jecisc:
Class>>#category contained two hacks because of some inconsistencies in the system. The change removes the hacks now that all inconsistencies got removed from the system.

This should speed up #category quite a lot

### DIFF
--- a/src/Kernel/Class.class.st
+++ b/src/Kernel/Class.class.st
@@ -208,11 +208,6 @@ Class >> applyDeprecation [
 ]
 
 { #category : 'accessing' }
-Class >> basicCategory [
-	^category
-]
-
-{ #category : 'accessing' }
 Class >> basicCategory: aSymbol [
 	category := aSymbol
 ]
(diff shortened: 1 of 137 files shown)


src/Kernel/Class.class.st
```diff
@@ -272,21 +267,9 @@ Class >> bindingOf: varName [
...
-
-	cat = Class unclassifiedCategory ifTrue: [ ^ cat ].
-
-	"The next lines are only useful anymore because of a problem in Fuel that should be resolved in not too long. Then we will be able to just return #basicCategory in this whole method"
-	((self packageOrganizer listAtCategoryNamed: cat) includes: self name) ifTrue: [ ^ cat ].
-	result := (self packageOrganizer categoryOfBehavior: self) ifNil: [ #Unclassified ].
-	self basicCategory: result.
+	"Answer the system organization category for the receiver."
 
-	^ result
```
jecisc, line 289: The biggest change is here :)

Gabriel-Darbord (approved):

PR #14686, opened 2023-09-18
ClassDescription is an abstract class 

MarcusDenker:
--> add #isAbstract

This avoids complaining about the subclass responsability that is implemented in both subclasses


jecisc (approved):

jecisc:
testIsAbstract – MacOSX64.Kernel.Tests.Classes.BehaviorTest

PR #14687, opened 2023-09-18
Scopes Browser P12

carolahp:
Creates a browser for creating and editing scopes.
Scopes are used to define an environment when applying code refactorings and while visualizing code in calypso. 
Pharo's main scope contains the code of the full pharo image.


Ducasse:
Caro could you push it to http://gtihub.com/pharo-spec/newTools ? Because like that it loads after spec.

carolahp:
> Caro could you push it to http://gtihub.com/pharo-spec/newTools ? Because like that it loads after spec.

No problem, [here's the new PR](https://github.com/pharo-spec/NewTools/pull/585)

carolahp:
Replaced by [PR](https://github.com/pharo-spec/NewTools/pull/585)

carolahp:
Replaced by [PR](https://github.com/pharo-spec/NewTools/pull/585)

PR #14692, opened 2023-09-19
RPackage: cleanings now that SystemOrganizer is removed

jecisc:
This PR brings a few changes now that the SystemOrganizer is removed.

# Feature

- It is now possible to have a Package + tag that have the same category than a Package name (for example Package #P1 with a tag #T1 in a system with a Package #'P1-T1')
- Implementation of #packageNamed:ifPresent:

# Bug fix

- #packageNamed:(ifAbsent:) now matches name in case insensitive because RPackage does not care about the case. This lead to the removal of #packageNamedIgnoreCase:ifAbsent:
- A bug was fixed when using #ensurePackage: to add a package with the same name than an existing package but a different case
- RPackageOrganizer >> hasPackage:  now knows how to deal with case in package names 

# Cleaning

- RPackage >> removeFromSystem got simplified
- RPackageOrganizer >> basicInitializeFromPackagesList: now uses #ensurePackage: that is a public API instead of the low level #basicRegisterPackage: that is too low level and might be removed later

And missing tests got added

### DIFF
--- a/src/RPackage-Core/RPackageOrganizer.class.st
+++ b/src/RPackage-Core/RPackageOrganizer.class.st
@@ -447,7 +447,9 @@ RPackageOrganizer >> registerInterestToSystemAnnouncement [
 RPackageOrganizer >> registerPackage: aPackage [
 	"A new package is now available and declared in the receiver. "
 
-	self validateCanBeAddedPackageName: aPackage name tagName: nil.
+	(self packageNamed: aPackage name ifAbsent: [ nil ]) ifNotNil: [
+		RPackageConflictError signal: ('Package can not be added because it conflicts with package a package of the same name.' format: { aPackage name }) ].
+
 
 	self basicRegisterPackage: aPackage.
 	aPackage extendedClasses do: [ :extendedClass | self registerExtendingPackage: aPackage forClass: extendedClass ].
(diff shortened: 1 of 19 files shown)


src/RPackage-Core/RPackageOrganizer.class.st
```diff
@@ -232,6 +230,13 @@ RPackageOrganizer >> defineUnpackagedClassesPackage [
...
+	| packageName |
+	packageName := aPackage isString
+		               ifTrue: [ aPackage ]
+		               ifFalse: [ aPackage name ].
+
+	(self hasPackage: aPackage) ifTrue: [ ^ self packageNamed: packageName ].
+
 	^ aPackage isString
 		  ifTrue: [ self packageNamed: aPackage ifAbsent: [ self registerPackage: (RPackage named: aPackage organizer: self) ] ]
 		  ifFalse: [
```
Gabriel-Darbord, line 242: Since there is an `hasPackage:` check that returns if true, it's not necessary to check again in the return statement, assuming `packageNamed:ifAbsent:` detects the absence the same way as `hasPackage:`:
```st
aPackage isString
		  ifTrue: [ self registerPackage: (RPackage named: aPackage organizer: self) ]
		  ifFalse: [
			  self registerPackage: aPackage.
			  aPackage ]
```

Gabriel-Darbord (approved):

src/RPackage-Core/RPackageOrganizer.class.st
jecisc: Indeed! I'll update that in a future PR :)

PR #14698, opened 2023-09-19
Simpler RBNamespace>>#allClassesDo:

MarcusDenker:
Instead of managing a seen Set, just add all the classes to a set and then use do:


MarcusDenker:
Maybe not good: as we work on RBClass, we can not rely on identity... (the RB Model is really not good.... it relies on names too much)

jecisc (approved):

PR #14702, opened 2023-09-20
Shift: Add tests and fix bugs in the change detectors

jecisc:
Change detector has no test right now and I found some bugs in it!

This change brings a bunch of tests to the change detector mostly but it also fixes two bugs.

The first bug fixed is a bug in the slot change detector. In case a class used a trait with slots and had a superclass with slots, change were always detected even if no change happened in the slots. I fixed the way to check the slots if the class has a trait to not always detect a change.

The second bug fixed was cause by the fact that if you add a trait composition to a class that had non, the metaclass was not updated since the builder enhancer was still a default one. Now if we change the trait composition the enhancer get reset

I also renamed ShSuperclassChangedDetector into ShSuperclassChangeDetector to keep the same suffix everywhere

We should do more tests here because I did not tests the kind of changes generated for example. But this is a first step (if someone want to help go ahead because I don't think I'll have the time to add more tests here for now)

### DIFF
--- a/src/BaselineOfShift/BaselineOfShift.class.st
+++ b/src/BaselineOfShift/BaselineOfShift.class.st
@@ -1,10 +1,11 @@
 Class {
-	#name : #BaselineOfShift,
-	#superclass : #BaselineOf,
-	#category : #BaselineOfShift
+	#name : 'BaselineOfShift',
+	#superclass : 'BaselineOf',
+	#category : 'BaselineOfShift',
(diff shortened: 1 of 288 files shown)


jecisc:
Also I did not add tests on the BitLayoutChangeDetector because I am not really sure how this is working :)

MarcusDenker (approved):
Looks good

PR #14714, opened 2023-09-20
RPackage: Remove categories from class removal announcement

jecisc:
This changes bring an update of ClassAnnouncement and especially ClassRemoved to reduce the usage of categories.

Instead we are basing things on package tags.


jecisc:
Redoing this one

PR #14716, opened 2023-09-20
RPackage: #removeTag: should remove contained classes

jecisc:
#removeTag: currently leaves the system in a bastard state because the RPackageOrganizer still contains the class since it was not removed from the system but the package does not.

Checking all senders of this method, they seems to expect the classes contained to be remove from the system so I updated the code to act accordingly.

This should fix some missmatch between package organizer state and unpackaged classes

### DIFF
--- a/src/RPackage-Core/RPackage.class.st
+++ b/src/RPackage-Core/RPackage.class.st
@@ -728,8 +728,8 @@ RPackage >> removeEmptyTags [
 { #category : 'removing' }
 RPackage >> removeFromSystem [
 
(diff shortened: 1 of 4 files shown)


jecisc:
I don't get how this error happens on the CI. Locally this works even after launching the shift tests that are the only ones creating the concerned class

jecisc:
Tests passed now. Seems like there was a random issue. Or that was fixed with my other change :)

MarcusDenker (approved):

PR #14724, opened 2023-09-21
Kernel: Clean method removal (Protocol and Method announcements improvements)

### LINKED ISSUE #14722: Removing an extension method in Calypso does not always remove it from Calypso
When removing extension methods from Calypso it does not necessarily remove it from Calypose.

To reproduce:
- Go on Object
- Select Beacon-Core extensions
- Remove #emit  

The method is still visible and we get a debugger when clicking on it.

I already explored multiple possible cause but did not find the root of the problem. Here are some things I checked:
- The method is not in the method dictionary of Object anymore
- The method is not in the *Beacon-Core protocol of Object anymore
- The method is not in the #extensionSelectors variable of the package Beacon-Core
- The announcement MethodRemoved is announce in the system announcer

It's hard to find the root of this problem but I guess it comes from Calypso reacting badly to the announcement?

jecisc:
This change brings three cleanings.

## Do not announce method recategorization on removal

When we remove a method we also remove it from its protocol. This announce a MethodRecategorized in general. But when we are removing a method, this is not a recategorization. This is a removal. 

Now we do not announce it since we already have MethodRemoved. A test has been added. 

Done here: https://github.com/pharo-project/pharo/compare/Pharo12...jecisc:kernel/clean-method-removal?expand=1#diff-8d01f62f275523f085c8d27fb7282a1359989b887a48f0d2ec2ec47686f8d89dR950

## Factorize TraitChange

Trait change was duplicating the whole method removal code. It made sense in the past because there were specificities in the Trait code. But now things have been simplified and we can just call #removeSelector: on the class.

Done here: https://github.com/pharo-project/pharo/compare/Pharo12...jecisc:kernel/clean-method-removal?expand=1#diff-c88a2ad54aadc6140e5188eeab8f0e5607070bb1ca408929824ac121c317d411R123

## Add the package of the method to MethodRemoved 

When we remove a method, it is removed from its package. In the past this was done through announcements but now it is done directly. This means that some listeners had the package if they were before RPackage (and after the change to remove it without announcement this introduced bugs). And some listened after RPackage and they had to do weird lookups to find the package they wanted. 

Now the announcement knows the package directly. This fix code listening before RPackage and allows us to simplify the code that was listening after RPackage change. 

For example: 
- Simplification in Epicea: https://github.com/pharo-project/pharo/compare/Pharo12...jecisc:kernel/clean-method-removal?expand=1#diff-97416cdad09b4a924b7cf6f9facfb35fdd9fa52c104f0bc1097b738c920d26dfR433
- Simplifications in Monticello: https://github.com/pharo-project/pharo/compare/Pharo12...jecisc:kernel/clean-method-removal?expand=1#diff-4c2c994f70cc8165c54b0e84ab36e2935da4aefb3c562ec1979506a0f40d4d38L286

I also added more tests for all of this

Fixes #14722


jecisc:
I really like the simplification in Monticello :p

jordanmontt (approved):

MarcusDenker (approved):

PR #14734, opened 2023-09-21
CharacterTest in wrong test package

### LINKED ISSUE #14731: CharacterTest in wrong test package
The class **Character** is in package "Kernel" (tagged BasicObjects) - typically one would expect the **CharacterTest** in "Kernel-Tests" package (also tagged under BasicObjects) therefore.

But it is packaged in "Collections-Strings-Tests" which seems like a modularity smell.

If possible we should move this test to package "Kernel-Tests"

astares:
Fix #14731


jecisc:
Because of TonelV3 changes there is now a conflict than cannot be solved via the web interface,

Sorry for the inconvenience

astares:
New PR in https://github.com/pharo-project/pharo/pull/14753

PR #14735, opened 2023-09-21
#species should be in private protocol

### LINKED ISSUE #14732: #species should be in private protocol
- #species typically is used in comparisions and should be in "private" protocol **for all classes**

so the following test should be added as a release test in class #ProperMethodCategorizationTest

```Smalltalk
testSpeciesMethodNeedsToBeInPrivateProtocol
	"The #species methods should be in method protocol 'private'"

	self assureAll: #species areCategorizedIn: #private whenSubclassOf: Object
``` 

and the four violations should be fixed.

image

astares:
Fix #14732


astares:
Windows is green - but Linux/Mac failed. Looks like a CI issue

jecisc:
I think those are random failures, I've seen them on other PR also :/

jecisc (approved):

PR #14738, opened 2023-09-21
Generalize ReClassSideResetMethodProtocolRule

### LINKED ISSUE #14737: Generalize ReClassSideResetMethodProtocolRule
There is an existing rule implemented in **ReClassSideResetMethodProtocolRule** that shows to the user that #reset on the class side should be in 'class initialization' protocol.

image

The rule offers an auto-repair, so the method could be quickly and properly categorized.

We could generalize this existing class into an easily extendable hierarchy of rules (that one can use or not as well as enable/disable depending on needs) that give hints to the user how to categorize the most prominent methods often found in classes like #hash, #=, #printOn:, #initialize, ... and their typical protocol categorization. 

This could also be useful for beginners who are not used to the typical categories like "comparing", "printing", "initialization", ....

astares:
Fix #14737

### DIFF
--- a/src/Renraku-Tests/ManifestRenrakuTests.class.st
+++ b/src/Renraku-Tests/ManifestRenrakuTests.class.st
@@ -2,22 +2,24 @@
 Please describe the package using the class comment of the included manifest class. The manifest class also includes other additional metadata for the package. These meta data are used by other tools such as the SmalllintManifestChecker and the critics Browser
 "
 Class {
-	#name : #ManifestRenrakuTests,
-	#superclass : #PackageManifest,
-	#category : #'Renraku-Tests-Manifest'
+	#name : 'ManifestRenrakuTests',
+	#superclass : 'PackageManifest',
+	#category : 'Renraku-Tests-Manifest',
+	#package : 'Renraku-Tests',
+	#tag : 'Manifest'
 }
 
-{ #category : #'code-critics' }
+{ #category : 'code-critics' }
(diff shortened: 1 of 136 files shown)


MarcusDenker:
Failure to build image:

```
SubclassResponsibility: ReProperClassMethodProtocolRule had the subclass responsibility to implement #selector
ReProperClassMethodProtocolRule(Object)>>subclassResponsibility
ReProperClassMethodProtocolRule(ReProperMethodProtocolRule)>>selector
ReProperClassMethodProtocolRule(ReProperMethodProtocolRule)>>name
```

astares:
- the tests in the image are green
- crash comes from opening the settings browser
- **it reveals a problem in the settings browser / the rule manager when displaying/identifying the rules**
- the class hierarchy is like:

image

with both classes ReProperMethodProtocolRule and ReProperClassMethodProtocolRule being abstract. 
ReProperMethodProtocolRule defines a #subclassResponsibility that ReProperClassMethodProtocolRule must not fulfill (as it is abstract too).

image

astares:
The rule manager collects all visible rules.

image

which is also including ReProperMethodProtocolRule as this simple test proves:

```Smalltalk
ReProperMethodProtocolRule isVisible 
```
But this would be wrong as an abstract class should not be visible.

We can and should fix this by implementing ReAbstractRule class>>#isVisible like this

```Smalltalk
isVisible

	self isAbstract ifTrue: [ ^ false ].
	
	^ self methodOfInteresetSelectors anySatisfy: [ :selector |
		self perform: selector ]
```

This fixes the problem. I will add that here in the PR together with a test

PR #14752, opened 2023-09-23
RPackage: Set class in its package/tag without announcement at catego…

jecisc:
…rization

The goal of this change is to not use announcement to set a class in a package and a tag when we update the category of a class. Instead we do it directly before.


jecisc:
I'll

PR #14753, opened 2023-09-24
CharacterTest in wrong test package (new attempt with second PR)

### LINKED ISSUE #14731: CharacterTest in wrong test package

astares:
Fix #14731

Replaces https://github.com/pharo-project/pharo/pull/14734

### DIFF
--- a/src/Kernel-Tests/CharacterTest.class.st
+++ b/src/Kernel-Tests/CharacterTest.class.st
@@ -7,9 +7,9 @@ This is the unit test for the class Character. Unit tests are a good way to exer
 Class {
 	#name : 'CharacterTest',
 	#superclass : 'ClassTestCase',
-	#category : 'Collections-Strings-Tests-Base',
-	#package : 'Collections-Strings-Tests',
-	#tag : 'Base'
+	#category : 'Kernel-Tests-BaseObjects',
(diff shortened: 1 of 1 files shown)


Ducasse:
I think that we should probably fix first 

DeprecationPerformedNotification: The method RPackageOrganizer>>#categories called from RPackageOrganizerSystemTest>>#testThatExistingPackageNamesDoesNotContainIllegalCharacters has been deprecated. The manipulation of categories got deprecated in Pharo 12. It is better to manipulate directly the packages and tags.
Deprecation: The method RPackageOrganizer>>#categories called from RPackageOrganizerSystemTest>>#testThatExistingPackageNamesDoesNotContainIllegalCharacters has been deprecated. The manipulation of categories got deprecated in Pharo 12. It is better to manipulate directly the packages and tags.

jecisc:
I’ll fix this next week. I missed this one when I removed the users of this method :)

jecisc (approved):

PR #14755, opened 2023-09-24
Rename Driver into InteractionDriver + introduce RBInstanceVariableHa…

Ducasse:
…sReferences


src/Kernel/BlockClosure.class.st
balsa-sarenac, line 1: Was this removed on purpose? Seemed kind of strange, so just wanted to double-check.

src/Refactoring-Tests-Core/RBConditionTest.class.st
```diff
@@ -122,13 +122,26 @@ RBConditionTest >> testHierarchyOfReferencesInstanceVariable [
 
 	| testClass |
 	testClass := rbNamespace classNamed: self class name.
-
+	self halt.
```
balsa-sarenac: This halt was left here, not sure if it is on purpose?

balsa-sarenac (commented):
This looks super nice! Left two minor comments

src/Kernel/BlockClosure.class.st
Ducasse: No this is super strange, I have no idea why. And in addition I removed it from the merge. :(
I have no idea how to fix it so may be I will recommit

src/Refactoring-Tests-Core/RBConditionTest.class.st
Ducasse: I will fix it too.

src/Kernel/BlockClosure.class.st
jecisc: I fixed it for you

Ducasse: Thanks! I was looking for it and I could not find!!!!

jecisc: What I did was to be outside Pharo, revert your commit but then I just commited the reversal of this file and droped the other changes of this revert

Issue #14749, opened 2023-09-22
Missing class side variable after reloading a Trait which used another Trait, from GitHub + image unstability and crashs


labordep:
**Bug description**
Class side variable of a Trait is removed after reloading the Trait.
Due to this problem, my Pharo image is unstable and a lot of crashes can appears.
When I'm reloading the package manually, the var is appears! But the system is not stable 

The Trait on [GitHub](https://github.com/OpenSmock/Molecule/blob/dev-99/Molecule-Tests/MolDirtyType.trait.st) (look at `anInstanceVariable`) : 
```smalltalk
Trait {
	#name : #MolDirtyType,
	#classInstVars : [
		'anInstanceVariable'
	],
	#traits : 'MolComponentType',
	#classTraits : 'MolComponentType classTrait',
	#category : #'Molecule-Tests-Resources - Contracts'
}
```

The Trait after loading the project into Pharo : 
![image](https://github.com/pharo-project/pharo/assets/49183340/896d4451-37ff-4665-acdd-e144ec0399da)
Look at the red methods which are using `anInstanceVariable`.

The Trait after re-loading manually the package : 
![image](https://github.com/pharo-project/pharo/assets/49183340/ceb5470e-b1b5-461e-a826-0e92853f2552)
![image](https://github.com/pharo-project/pharo/assets/49183340/31663a34-6a4c-4282-a2a5-2baab401f50c)
Red methods are always here.
![image](https://github.com/pharo-project/pharo/assets/49183340/15bbd460-6def-4e51-aa1c-f7bfddd2afbf)
![image](https://github.com/pharo-project/pharo/assets/49183340/1a00568c-e7ac-449c-b4eb-2010b19c0f7d)
The image is always unstable, var is here but not considered.

**To Reproduce**
1. Create a Trait1
2. Define another Trait2 which uses Trait1
3. Define a class side variable into Trait2
4. Publish the code into GitHub
5. Create a new image
6. Reload the project : the var in Trait2 is missing

The problem is present when loading manually (Add button) or from a baseline script.

**About crashes and unstability**
The image is unstable at the moment of a Trait1 with a class side var uses a Trait2, or at the moment of we create a class side var in a Trait 1 uses Trait 2.

**Expected behavior**
Variable should be exists, but missing.

**Screenshots**

**Version information:**
 - OS: Windows 11 Pro 64 bits & Windows 10 Pro 64 bits
 - Version: Pharo 12 & Pharo 11 & Pharo 10

[...]

PR #14759, opened 2023-09-25
Monticello: Clean installation of classes and traits

jecisc:
- MCClassDefinition>>createClass  should not hardcode a global access
- MCTraitDefinition>>createClass should not set the class slots and comment outside the class builder
- MCTraitDefinition>>createClass should catch the same warning as the overriden method (I guess? Or it should be removed in both? If someone knows this part of the system better than me let me know what you think :) )

I found this looking at #14749 but this does not fix the problem

### DIFF
--- a/src/Monticello/MCClassDefinition.class.st
+++ b/src/Monticello/MCClassDefinition.class.st
@@ -207,7 +207,7 @@ MCClassDefinition >> createClass [
 
 	superClass := superclassName = #nil ifFalse: [ Smalltalk globals at: superclassName ].
 	^ [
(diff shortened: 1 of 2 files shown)


jecisc:
I added one more thing to this PR which is to set the class trait composition of a MCTraitDefinition in #createClass. 

The goal of this change is to remove the usefulness of MCClassTraitComposition and after some other cleanings we should be able to fix in a clean way issue #14749 after some changes in Tonel

Gabriel-Darbord (approved):

PR #14765, opened 2023-09-25
Remove classes improvements

balsa-sarenac:
Finished RBRemoveClassDriver
Finished migrating conditions to separate classes.
Bunch of small improvements.

### DIFF
--- a/src/Refactoring-Core/RBRemoveClassRefactoring.class.st
+++ b/src/Refactoring-Core/RBRemoveClassRefactoring.class.st
@@ -78,7 +78,7 @@ RBRemoveClassRefactoring >> checkBreakingChangePreconditions [
 	conditions ifEmpty: [ ^ self ].
 	errorStrings := String streamContents: [ :aStream |
 		                conditions do: [ :cond |
-			                cond displayStringOn: aStream ] ].
+			                cond violationMessageOn: aStream ] ].
 	self refactoringWarning: errorStrings
 ]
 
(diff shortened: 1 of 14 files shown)


src/Refactoring-Core/RBRemoveClassRefactoring.class.st
```diff
@@ -75,13 +96,28 @@ RBRemoveClassRefactoring >> classNames: aClassNameCollection [
...
 RBRemoveClassRefactoring >> environmentWithUsersOf: aClassable [
 	^ RBClassEnvironment
 		onEnvironment: RBBrowserEnvironment new
 		classes: (self model classesReferencingClass: aClassable)
 ]
 
+{ #category : 'preconditions' }
+RBRemoveClassRefactoring >> preconditionAreNotMeta [
+
+	^ RBClassesAreNotMetaClassCondition new classes: classesDictionary
```
jecisc, line 118: It could be nice to add #classes: on the class side. And if this is done there is no need to have this method I think. `RBClassesAreNotMetaCondition classes: classesDictionary` is readable enought

jecisc (approved):
I left a comment but that can be for later

src/Refactoring-Core/RBRemoveClassRefactoring.class.st
balsa-sarenac: Agreed, I had it already in my Todo but forgot it :(

PR #14766, opened 2023-09-25
Monticello: do not rely on MCClassTraitDefinition

### LINKED ISSUE #14749: Missing class side variable after reloading a Trait which used another Trait, from GitHub + image unstability and crashs

jecisc:
The infos in MCClassTraitDefinition should go in MCTraitDefinition that already have everything needed about the class trait such as slots. 

Having this duplicated info creates problem because it removes what is already loaded by MCTraitDefinition such as the class slots. 

I propose to just rely on the MCTraitDefinition. This requires also a change in Tonel that I am doing in parallel

Fixes #14749


jecisc:
Needs to be merged with: https://github.com/pharo-vcs/tonel/pull/112

jecisc:
Both PR are green. But we need to merge them both at the same time (The tonel one first so that the build triggered by this PR takes the change). If we don't, we will have some weird diffs in Iceberg since the definitions will not be the same in both sides

MarcusDenker (approved):
ok!

PR #14767, opened 2023-09-25
Fixes  #14740 : '/..' asFileReference canonicalize returns incorrect result

### LINKED ISSUE #14740: `'/..' asFileReference canonicalize` returns incorrect result
**Bug description**
Canonicalize doesn't handle absolute paths that begin with a reference to the folder's parent, e.g. `/..`.

The parent of the root directory is itself (the root directory), so this is a valid path.

**To Reproduce**

demonstrates the issue, the first path is canonicalized correctly, while the second isn't.

**Expected behavior**
`'/..' asFileReference canonicalize >>> '/'`

**Screenshots**
See above

**Version information:**
 - OS: [e.g. iOS] NixOS
 - Version: [e.g. 1.6.1] 23.05
 - Pharo Version [e.g 9] Pharo 12 (Build information: Pharo-11.0.0+build.706.sha.ea9092e5193593025c3d8b59fb1010db149b0431 (64 Bit))

**Expected development cost**
PR on the way.

**Additional context**
N/A

akgrant43:
Any number of parent references at the start of an absolute path should be removed.

- `PathTest>>testCanonicalization`
- `FileReferenceTest>>testCanonicalization`

have been extended to test leading parent references.

### DIFF
--- a/src/FileSystem-Tests-Core/FileReferenceTest.class.st
+++ b/src/FileSystem-Tests-Core/FileReferenceTest.class.st
@@ -216,7 +216,7 @@ FileReferenceTest >> testCanonicalization [
 	self assert: ref path segments equals: #('a' 'c').
 
 	ref := '/..' asFileReference canonicalize.
-	self assertEmtpy: ref path segments.
+	self assertEmpty: ref path segments.
 
 	ref := '/../a/b/../c' asFileReference canonicalize.
 	self assert: ref path segments equals: #('a' 'c')
--- a/src/FileSystem-Tests-Core/FileReferenceTest.class.st
+++ b/src/FileSystem-Tests-Core/FileReferenceTest.class.st
@@ -216,7 +216,7 @@ FileReferenceTest >> testCanonicalization [
 	self assert: ref path segments equals: #('a' 'c').
 
 	ref := '/..' asFileReference canonicalize.
-	self assert: ref path segments isEmpty.
+	self assertEmtpy: ref path segments.
 
 	ref := '/../a/b/../c' asFileReference canonicalize.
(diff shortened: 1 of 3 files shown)


src/FileSystem-Tests-Core/FileReferenceTest.class.st
```diff
@@ -213,6 +213,12 @@ FileReferenceTest >> testCanonicalization [
 	self assert: ref path segments equals: #('a' 'c').
 
 	ref := '/a/b/../c' asFileReference canonicalize.
+	self assert: ref path segments equals: #('a' 'c').
+
+	ref := '/..' asFileReference canonicalize.
+	self assert: ref path segments isEmpty.
```
jecisc:
```suggestion
	self assertEmtpy: ref path segments.
```

src/FileSystem-Tests-Core/PathTest.class.st
```diff
@@ -133,7 +133,14 @@ PathTest >> testCanonicalization [
 	self assert: ref segments equals: #('a' 'c').
 
 	ref := (Path / 'a/b/../c') canonicalize.
+	self assert: ref segments equals: #('a' 'c').
+
+	ref := (Path / '..') canonicalize.
+	self assert: ref segments isEmpty.
```
jecisc:
```suggestion
	self assertEmpty: ref segments.
```

jecisc (commented):
Thank you

I did two small suggestions. It is to use better assertions. In case those assertions will fail in the future, they will print the content of the collection to help debug them instead of just having an assertion failure

jecisc (approved):

akgrant43:
Thanks, @jecisc !

PR #14778, opened 2023-09-26
RPackage: Introduce UndefinedPackage and UndefinedPackageTag

jecisc:
This PR introduces the classes UndefinedPackage and UndefinedPackageTag. 

The goal is to have a null object pattern for the Packages and Tags in the system and simplify future evolution. 

I also deprecated some methods that had equivalents and brought nothing more. #isDefault got renamed into #isUndefined.

I added a bunch of new tests to ensure the behavior is the one intended.

@dionisiydk I took your advice into account and named them Undefined even if for now, the method #name still return the same as before to not break everything.


jecisc:
I did other smaller PR because the changes impacted too many things still

PR #14779, opened 2023-09-26
Compiler: do not use an intermediate stream when compiling BraceArray constructs {}

### LINKED ISSUE #14683: Private WriteStream for the compiler
- with just a little subclass of WriteStream we could avoid to have such method in WriteStream.

```
braceArray: anArray
	"This method is used in compilation of brace constructs.
	It MUST NOT be deleted or altered."

	collection := anArray.
	position := 0.
	readLimit := 0.
	writeLimit := anArray size
```

I always thought that OOP was about incremental definition so why put everything in a single class. 
Especially since a method like that is ONLY used by the compiler.

MarcusDenker:
When compilig {}, we were using a Stream. But that is not needed: we statically know the size of the array, we can just create it and the use a:put: to fill it.

- check decompiler: it was not decompiling to {}, but to the generated code. The same works for at:put:
- fix #examplePushBigArray, the test is green (it checks execution of code, not the generated code)
- remove not needed code int the decompiler,( it was not used before either, as the {} where decompiled to explicit nextPut: sends)
- remove #braceArray, #braceArray: and #braceStream: 

The resulting code is faster, for #defaultColors 1.8 times

fixes #14683


jecisc (approved):
Nice simplification of the code :)

src/OpalCompiler-Core/OCASTTranslator.class.st
```diff
@@ -643,17 +643,17 @@ OCASTTranslator >> visitInlinedBlockNode: anOptimizedBlockNode [
 
 { #category : 'visitor - double dispatching' }
 OCASTTranslator >> visitLargeArrayNode: aRBArrayNode [
-	"Long form: generates (Array braceStream: N) nextPut: a; nextPut: b; ...; braceArray"
+	"Long form: useing at:put:"
```
gcotelli:
```suggestion
	"Long form: using at:put:"
```

gcotelli: Just a typo in the comment

MarcusDenker:
Sadly the bootstrap does not allow to remove the methods...

```
Tue Sep 26 14:32:42 UTC 2023 [Compiler] Initializing Bootstraped Image
+ /builds/workspace/est_and_branch_Pipeline_PR-14779/64/bootstrap-downloads/vmtarget/pharo --headless Pharo12.0-PR-compiler-64bit-2764de3.image
Instance of Array class did not understand #braceStream:
Array class(Object)>>doesNotUnderstand: #braceStream:
```

jecisc:
This is in the part that is outside espell I think so we'll be able to fix it

MarcusDenker:
Maybe doing it in two steps will already be enough, I will later today make a version of the PR that does not remove the methods

PR #14790, opened 2023-09-27
[Refactoring engine] 2023 09 26 enh remove instance variable2

Ducasse:
As a first step it is good and working
Now testing the pop up is difficult.


src/Refactoring-Core/RBAbstractCondition.class.st
```diff
@@ -35,7 +35,7 @@ RBAbstractCondition >> check [
 { #category : 'private' }
 RBAbstractCondition >> errorMacro [
 
-	^ errorMacro ifNil: [ 'unknown' ] ifNotNil: [ errorMacro ]
+	^ errorMacro ifNil: [ self halt. 'unknown' ] ifNotNil: [ errorMacro ]
```
jecisc: There is an halt here
```suggestion
	^ errorMacro ifNil: [ 'unknown' ] ifNotNil: [ errorMacro ]
```

src/Refactoring-Core/RBAndCondition.class.st
```diff
@@ -0,0 +1,50 @@
...
+
+{ #category : 'checking' }
+RBAndCondition >> check [
+	left check
+		ifFalse:
+			[ ^ false].
+	right check
+		ifFalse:
+			[ ^ false].
+	^true
```
jecisc, line 21:
```suggestion
	^ left check and: [ right check ]
```

src/Refactoring-Core/RBCompositeRefactoring.class.st
```diff
@@ -47,6 +47,11 @@ RBCompositeRefactoring >> preconditions [
 	^ self applicabilityPreconditions & self breakingChangePreconditions 
 ]
 
+{ #category : 'preparation' }
+RBCompositeRefactoring >> prepareForInteractiveMode [
+	self
```
jecisc, line 52: Was it intentional to just have a call to self without even a return?

```suggestion
	"Do nothing here"
```

src/Refactoring-Core/RBCondition.class.st
```diff
@@ -250,6 +250,7 @@ RBCondition class >> hierarchyOf: aClass referencesInstanceVariable: aString [
 	^ self new
 		  block: [
 			  aClass withAllSubclasses anySatisfy: [ :each |
+					self halt.
```
jecisc:
```suggestion
```

Halt again

src/Refactoring-Core/RBNewAbstractCondition.class.st
```diff
@@ -0,0 +1,35 @@
...
+
+{ #category : 'checking' }
+RBNewAbstractCondition >> check [
+	self subclassResponsibility
+]
+
+{ #category : 'accessing' }
+RBNewAbstractCondition >> errorString [ 
+
+	^ ' please redefine me '
```
jecisc, line 22: Something seems missing here?

src/Refactoring-Core/RBAndCondition.class.st
Ducasse: Right now I do not want to change this because this is to reflect the old control flow that we simplified.

src/Refactoring-Core/RBNewAbstractCondition.class.st
Ducasse: no.

Ducasse:
I have no idea what changes. 
```
Got initialize
	<ignoreUnusedVariables: #( #foo1 )>
	<ignoreUnusedClassVariables: #( Foo1 )>
	super initialize.
	name := '' 

instead of initialize
	<ignoreUnusedVariables: #( #foo1 )>
	<ignoreUnusedClassVariables: #( Foo1 )>
	super initialize.
	self name: ''.
```

But I did not touch this.

Ducasse:
I do not get why I have now broken tests. 
I will propose to Balsa to work on a side project and do not try to improve Pharo. I think that it will be faster for us to work
and at the end if people wants then may be they will be able to take what we are doing.

Ducasse:
Problems with low level API in RBClass :( fixing this now. 
I hate this code.

PR #14792, opened 2023-09-27
Bootstrap should use P10 VM for small image 

tesonep:
- Boostrapping of small image should use P10 (using Espell)
- Adding a P10 VM download so we can run the build

### DIFF
--- a/bootstrap/scripts/2-download.sh
+++ b/bootstrap/scripts/2-download.sh
@@ -37,6 +37,17 @@ if [ ! -e "${BOOTSTRAP_VMTARGET}" ]; then
 	echo "Target VM: $(${VM} --version)"
 fi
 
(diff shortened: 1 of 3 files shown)


jecisc (approved):
Thank you!

tesonep:
@jecisc can you test if it is working on your machine?

jecisc:
I have a bootstraping happening now and it's able to run the espell image :)

It's currently running so I'll let it run

jecisc:
I'll merge it this afternoon after it finished on my machine

jecisc:
I had a crash but it was a lot later in the bootstrap process so this seems good!

MarcusDenker:
The boostrap is now much faster: 16=17min instead of ~21-24 Minutes

PR #14793, opened 2023-09-27
RPackage: Introduce UndefinedPackageTag

jecisc:
Introduce nil design pattern for the package tag of an unpackaged class


jecisc:
This is weird the tests failing are passing locally and I don't really get what is the origin of the problem

jecisc:
Newer version incoming

PR #14795, opened 2023-09-27
remove #braceArray, #braceArray: and #braceStream:

MarcusDenker:
remove #braceArray, #braceArray: and #braceStream:, no senders in the image (and private methods, so no deprecation)

### DIFF
--- a/src/Collections-Sequenceable/Array.class.st
+++ b/src/Collections-Sequenceable/Array.class.st
@@ -55,14 +55,6 @@ Class {
 	#tag : 'Base'
 }
 
-{ #category : 'brace support' }
-Array class >> braceStream: nElements [
-	"This method is used in compilation of brace constructs.
-	It MUST NOT be deleted or altered."
-
-	^ WriteStream basicNew braceArray: (self new: nElements)
-]
-
 { #category : 'instance creation' }
 Array class >> empty [
 	"A canonicalized empty Array instance."
--- a/src/Collections-Streams/WriteStream.class.st
+++ b/src/Collections-Streams/WriteStream.class.st
@@ -46,25 +46,6 @@ WriteStream >> << anObject [
 		ifFalse: [ anObject putOn: self ]
 ]
 
-{ #category : 'private' }
-WriteStream >> braceArray [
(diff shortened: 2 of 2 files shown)


MarcusDenker:
Fails again with:

```
+ echo Wed Sep 27 12:04:17 UTC 2023 '[Compiler] Initializing Bootstraped Image'
Wed Sep 27 12:04:17 UTC 2023 [Compiler] Initializing Bootstraped Image
+ /builds/workspace/est_and_branch_Pipeline_PR-14795/64/bootstrap-downloads/vmtarget/pharo --headless Pharo12.0-PR-compiler-64bit-2d94a7d.image
Instance of Array class did not understand #braceStream:
Array class(Object)>>doesNotUnderstand: #braceStream:
Color class>>defaultColors
Color class>>initializeColorRegistry
Color class>>initialize
```

So it maybe espell creates an image that uses #braceStream:, we then reload everything so the final image does not use it

tesonep:
Hi Marcus, the problem is that the classes are compiled using the compiler that is in the bootstrapImage, so it is a P10 Opal version. That applies to the classes that are created before the compiler is loaded. 
Cyril is working on updating the bootstrapper image.

MarcusDenker:
Ok, I close the PR for now

jecisc:
> Hi Marcus, the problem is that the classes are compiled using the compiler that is in the bootstrapImage, so it is a P10 Opal version. That applies to the classes that are created before the compiler is loaded. Cyril is working on updating the bootstrapper image.

I would love to be able to update Espell but I don't have the knowledge to be able to build an espell version usable for the bootstrap in Pharo 11 with the latest pharo vm branch. It requires a high knowledge of the vm and espell and I'm really really far from having it. And it is impossible to reproduce the image we are currently using because it requires a specific pharo-vm commit and pharo 10 development image.

With the fix of using the Pharo 10 vm I can debug the bootstrap crashes but not update Espell. I would need help on that subject. I'd be interesting to see the process of getting espell working again, but it would most probably be just me watching. (I'd love to learn more about the vm but it's hard to find the time to work on that subject :( )

PR #14797, opened 2023-09-27
RPackage: Finish to clean the package/tag import logic

jecisc:
With the category management and import logic of RPackage everything was a mess in the import logic. I already cleaned a lot of things and this PR finishes to clean the import logic.

Nothing relies on setting the category anymore inside RPackage (we still rely on the announcement when using Class>>#category: but this is a future work to update). 

- RPackage>>addClass: does not rely on category anymore
- RPackage>>#moveClass:toTag: has been improved to do less useless actions 
- RPackageTag>>#addClass: can now be used to do a full import a of a class in a new package. Before it was adding the class to a tag but not importing anything else (extension methods and compagnie)

I started to add tests on announcements

Future step:

ClassRepackaged is announced when we update the tag of a class. The problem is that it does not contains information about the package tag of the class. I want to add this new info and then to remove all users of ClassRecategorized to deprecate this announcement


jecisc:
To fix this one it would be better to have the undefined package tag integrated. But for that I need to find were the weird bugs I have on the PR comes from

jecisc:
Tests are finally passing!

Gabriel-Darbord (approved):

PR #14801, opened 2023-09-28
RPackage: Drop usage of categories in class installer

jecisc:
This change make the class installer use the package and tag API instead of #category:.

I am really not sure the system is ready for such a change but let's try it


jecisc:
Of course the bootstrap does not like it ahah

jecisc:
Done in another PR

PR #14805, opened 2023-09-28
This removes the class TextEditor which is not used anymore

### LINKED ISSUE #2717: Should deprecate Editor, TextEditor and SmalltalkTextEditor
Editor is referenced in 
- Editor >> editingSettingsOn:  do we add this to RubTextEditor?
- Paragraph >> caretWidth (refer to editor) and Paragraph should be replaced by RubParagraph
- TextMorph >> manageCursor refer to Editor blinkingCursor (but there is nobody else).
- WorldMorph >> dispatchKeyStroke: refer to Editor sending message cmdKeysInText 

We should check the settings and see how they are covered in RubTextEditor

MarcusDenker:
The package has some more code that needs to be moved to other packages, which should be done in another PR


MarcusDenker:
very strange... it seems we can not remove a class anymore if it uses a trait. As the trait claims that it need that class...

jecisc:
I think it might be a bug due to recent changes in RPackage or Refactorings. I'll try to take a look at it

Ducasse:
Package Polymorph-Widgets-Traits depends on the following classes:
  EditorFindReplaceDialogWindow

Ducasse:
This is because of extension methods for example 

```
removeBoxes
	"Remove all label area boxes."

	expandBox ifNotNil: [expandBox delete. expandBox := nil].
	collapseBox ifNotNil: [collapseBox delete. collapseBox := nil]
```

jecisc:
Apparently removing the class did not remove the extensions in the image of Marcus. I need to check if this is something I broke in RPackage.

Ducasse:
What a stupid design.

Ducasse:
What is strange is that this is not related to TextEditor :)

MarcusDenker:
It is simpler to redo this later, for the problem we have the smaller PR (which is a subset)

I close this

PR #14810, opened 2023-09-28
Small cleanup FileSystemStore not needed #subclassResponsabiliy  methods

MarcusDenker:
- #basicModificationTime: has no senders, is #subclassResponsabiliy yet never overridden

- #basicIsWritable:  is #subclassResponsabiliy, has no implementation. The only user isWritable: itself is overriden everywhere


jecisc (approved):
nice catch

PR #14813, opened 2023-09-28
Prevent SystemDictionary key to be something else than symbols 

jecisc:
SystemDictionary keys should always be symbols but it happens that they are objects in some tests. I hope this PR will uncover those tests


jecisc:
Here are the three culprits of messing up the global dictionary :) 

testLookUpInGlobalsObjectKey – MacOSX64.Fuel.Core.Tests.Configuration.FLSerializerTest
testLookUpInGlobalsPassingValue – MacOSX64.Fuel.Core.Tests.Configuration.FLSerializerTest
testLookupAllInGlobals – MacOSX64.Fuel.Core.Tests.Configuration.FLSerializerTest

@theseion Those 3 tests are putting objects as keys of the SystemDictionary. But that should never happen. Do you have a suggestion on what to do here?

theseion:
Let me look at them. I thought I had covered those when I removed the wrapper.

theseion:
AFAICT, the only bad test is `testLookupAllInGlobals`, where I had forgotten to add the globals to the `addedGlobals` collection. How did you determine that the other two tests are bad?

jecisc:
So we added a check to SystemDictionary to raise an error if the key is not a symbol and the three tests listed are failing with this check. 
I’m not on my laptop now but I think I’ve seen that they explicitly set an object as a key in the environment.

theseion:
I see. So that's no longer legal. Good to know. I'll update the tests.

jecisc:
I don’t know if it was in the past, but it has been a while since the system is not able to work with real objects in the keys of the system dictionary because some of the own code of SystemDictionary is sorting the keys for example.

It has been like this since I know this part of the system, but maybe it was different in the past? 

Thank you for your help! :) I was not sure if there was not a mécanisme I did not know about there

theseion:
Should be fixed with #14819.

PR #14816, opened 2023-09-28
Fix announcements during class creation 

jecisc:
If we create a class with traits in the system, the class builder is installing some basic methods in the class. This operation should be silent but in fact we are raising some announcements currently. 

This fixes the problem and adds a tests to ensure the behavior is not reverted.


MarcusDenker (approved):

MarcusDenker:
Speeds up traits test *a lot* (whole test run on linux 3min faster)

PR #14819, opened 2023-09-28
fix: SystemDictionary keys must now be symbols

### LINKED ISSUE #14818: Fuel tests write objects to system globals
As discovered by @jecisc, some Fuel tests use objects as keys for entries in the system globals. This used to be legal but is no longer so (apparently it hasn't been for a while).

See #14813.

theseion:
Some Fuel tests still expected that having objects as keys in the globals dictionary was legal.

Fixes #14818.


jecisc (approved):

jecisc:
Thank you Max! This is really appreciated

PR #14821, opened 2023-09-29
Adding a #reflection: pragma on reflective methods

Inao0:
Currently, it isn't easy to differentiate at a glance a reflective method from a base-level method in Pharo.
I tagged reflective methods with a #reflection: pragma to allow to manipulation of those methods programmatically. 
The parameters of the pragmas are the categories are the ones from IWST paper : [HAL link](https://inria.hal.science/hal-04217271)


MarcusDenker:
There are conflicts

jecisc (approved):

PR #14824, opened 2023-09-29
Allow defining methods from class pane again

MarcusDenker:
add back a check to compile a method from the class pane.
Instead of evaluating and checking for errors, we check for the >> characters in the first line


jecisc (approved):

Ducasse:
I want them :) thx for the coding session.

PR #14827, opened 2023-09-29
Deprecate documentPath in favor of imagePath.

hernanmd:
An update of the previous [PR](https://github.com/pharo-project/pharo/pull/12866/files) to check if it's still failing in Pharo 12


hernanmd:
Close this as it seems it could be resolved through already existing [PR](https://github.com/pharo-project/pharo/pull/12866)

PR #14828, opened 2023-09-29
Refactoring: Remove controllers

jecisc:
Remove the controllers from refactoring since this is not used and we don't need to use that.


Ducasse:
So much less code :)

PR #14831, opened 2023-09-30
fixes #14715: RBNamespace >> classNamed: should convert strings to symbols

### LINKED ISSUE #14715: `RBNamespace >> classNamed:` should convert strings to symbols
Otherwise, we need to convert strings to symbols everywhere and it's not nice...

Ducasse:
just using inforcing asSymbol in the right place


Ducasse:
indeed nil is not a symbol!

```
testRedefineClassChange
	| st |
	st := modelFactory rbNamespace new.
	st defineClass: 'nil << #Object
				slots: { #a};
				sharedVariables: {#A};
				sharedPools: {TextConstants};
				package: #''Refactory-Testing'''.
```

PR #14836, opened 2023-09-30
Remove EditorFindReplaceDialogWindow

hernanmd:
This pull request removes the EditorFindReplaceDialogWindow class as its functionality is now covered by the existing SpRubFindReplaceService. 

The redundant code has been removed to streamline the codebase and reduce maintenance overhead. This change enhances code clarity and ensures consistency in the find and replace functionality across the application.


Ducasse:
Thanks hernan. We will still need a replacement for the find replace because we will move out of Morphic.

Ducasse:
The build did not pass but I do not think that this is related to the issue in itself

```
MetacelloNotification: Loaded -> Fuel-Core-tonel.1 --- tonel:///builds/workspace/est_and_branch_Pipeline_PR-14836/64/src --- cache
Error: Package Polymorph-Widgets-Traits depends on the following classes:
  EditorFindReplaceDialogWindow
  EditorFindReplaceDialogWindow>>#aboutText
  EditorFindReplaceDialogWindow>>#aboutTitle
  EditorFindReplaceDialogWindow>>#removeBoxes
```

Ducasse:
Cyril will check this issue monday.

hernanmd:
After removing the class there are a lot of obsolete instances which `SmalltalkImage current fixObsoleteReferences.` does not actually fix. I've already discarded they come from the ObsoleteClasses and the MC cache:

```smalltalk
Class initialize.
MCMethodDefinition flushMethodCache.
```

But I don't know yet where they come from.

MarcusDenker:
I tried to remove the same class in the PR https://github.com/pharo-project/pharo/pull/14805, there is some more infos there what is happending

hernanmd:
I close this PR as I will re-doit removing extension methods first (which were causing problems due to recent RPackage updates)

jecisc:
I think I might have fix the issue we had here. In fact the methods were removed and RPackage was in the right state but a bug prevented Iceberg to mark the extension package as dirty and thus to commit the change without recalculating the dirty packages. This should now not be a problem anymore! I also added a test on the problematic behavior

PR #14837, opened 2023-09-30
Deprecate old class creation methods

jecisc:
The goal of this change is to deprecate all the old class creation methods. 

The only one left is #subclass: because it still has some sender. One of them in Spec


MarcusDenker:
I would like to discuss this before merging

(this is why workimg over the weekend sometimes is bad... If I want to comment, I can not take time off over the weekend)

jecisc:
This should be good now. Let's see what the CI says. 

It depends on https://github.com/pharo-project/pharo/pull/14878

jecisc:
For now we should not merge this PR because we need to make sure the tooling does not rely at all on this anymore.

jecisc:
Now that we removed the use of old classe definitions in Calypso we are able to merge this one! :)

Ducasse:
So great!

PR #14665, opened 2023-09-14
Add ‘scale’ for display of a RubParagraph using a RubDisplayScanner

Rinzwind:
Pull request #14630 added ‘scale’ for display of a Paragraph using a BitBltDisplayScanner, this pull request extends that to the display of a RubParagraph using a RubDisplayScanner. An example is given by the snippet below. For comparison, for ‘form1’, the example Rubric paragraph is displayed normally and then the form is scaled by a factor of two:

<p align="center"></p>

For ‘form2’, the paragraph is again displayed directly on the larger form so that the character glyphs have better detail:

<p align="center"></p>

```smalltalk
text := Text streamContents: [ :stream |
	stream
		nextPutAll: (Microdown asRichText: '# Lorem Ipsum'); cr;
		nextPutAll: (Microdown asRichText: 'Dolor **sit** amet, _consectetur adipiscing_ elit,');
		nextPutAll: ' ';
		nextPutAll: (Microdown asRichText: 'sed do eiusmod tempor ![](pharo:///Object/iconNamed:/home) incididunt ut');
		nextPutAll: '  ';
		withAttribute: (TextAnchor new anchoredMorph: (CircleMorph new extent: 10@10; yourself); yourself) do: [
			stream nextPut: Character home ];
		nextPutAll: '  ';
		nextPutAll: (Microdown asRichText: '  labore ~et dolore magna~ aliqua.') ].
(editingArea := RubEditingArea new)
	extent: 300@0;
	beWrapped;
	setTextWith: text.
paragraph := editingArea paragraph next next next.
bounds := 0@0 corner: paragraph extent.
scale := 2.

(baseForm := Form extent: paragraph extent depth: Display depth) fillColor: Color white.
baseForm getCanvas
	rubParagraph: paragraph bounds: bounds color: Color black;
	translateBy: bounds topLeft during: [ :translatedCanvas |
		paragraph text embeddedMorphs do: [ :morph |
			translatedCanvas fullDrawMorph: morph ] ].
form1 := baseForm scaledToSize: baseForm extent * scale.

(form2 := Form extent: form1 extent depth: Display depth) fillColor: Color white.
canvas := form2 getCanvas.
canvas setPaintColor: Color black.
port := canvas privatePort.
origin := canvas origin.
scanner := (port clippedBy: ((bounds translateBy: origin) scaleBy: scale)) rubDisplayScannerFor: paragraph
	foreground: Color black.
scanner scale: scale.
[...]

PR #14841, opened 2023-10-01
Factor out methods for saving and restoring patches from HandMorph onto Canvas

Rinzwind:
This pull request factors out methods related to ‘savedPatch’ from HandMorph onto Canvas so that they can be overridden in subclasses of Canvas.

### DIFF
--- a/src/Morphic-Core/Canvas.extension.st
+++ b/src/Morphic-Core/Canvas.extension.st
@@ -1,0 +1,19 @@
+Extension { #name : 'Canvas' }
+
+{ #category : '*Morphic-Core' }
+Canvas >> allocateSavedPatch: extent [
+
+	^ self form allocateForm: extent
+]
+
+{ #category : '*Morphic-Core' }
+Canvas >> contentsOfArea: aRectangle intoSavedPatch: aForm [
+
+	self contentsOfArea: aRectangle into: aForm
+]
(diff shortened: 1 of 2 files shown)


Rinzwind:
I applied this to the ScalingCanvas used in the [‘Retina display’ scaling experiment from earlier pull requests](https://github.com/pharo-project/pharo/pull/14665#issuecomment-1739948590): commits bc516aaadfac9862 and 42f0d80143e851ba (these commits are not included in this pull request). To try this: create an image in Pharo Launcher using Pharo 12 build 833 as the template, clone the ‘pharo’ repository from GitHub using ‘Rinzwind’ as the ‘Owner name’ and ‘pharo’ as the ‘Project name’, then check out the branch ‘scaling-canvas-tryout-6’.

In the extra world opened by `OSWorldRendererScalingCanvas example`, when you drag a window over another window, previously the text and icons in the other window would get blurred because the ‘savedPatch’ of the HandMorph was scaled down and then scaled back up when restored, that has now been fixed. What has not been fixed yet: text and icons in the window that you are dragging still get blurred until you drop the window because the ‘cacheCanvas’ of the HandMorph is not a ScalingCanvas. I will try to fix that next.

MarcusDenker (approved):

PR #14845, opened 2023-10-01
BracedCollection ready for review and integration

Ducasse:
- Added support for broken situation
- now semanticAnalyzer is doing something
- fix the completion to at least not bark
- couple more tests Ready for reviews
- introduced a specific method withAllForBrace: so that other candidates can just redefine this method if needed.
Please review

### DIFF
--- a/src/AST-Core-Tests/RBLiteralCollectionParserTest.class.st
+++ b/src/AST-Core-Tests/RBLiteralCollectionParserTest.class.st
@@ -1,0 +1,106 @@
+Class {
+	#name : 'RBLiteralCollectionParserTest',
+	#superclass : 'RBParserTest',
+	#category : 'AST-Core-Tests-Parser',
+	#package : 'AST-Core-Tests',
+	#tag : 'Parser'
+}
+
+{ #category : 'tests' }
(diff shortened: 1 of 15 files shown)


jecisc:
```
Instance of RBParseErrorNode did not understand #binding
RBParseErrorNode(Object)>>doesNotUnderstand: #binding
OCASTTranslator>>visitLiteralCollectionNode:
RBLiteralCollectionNode>>acceptVisitor:
OCASTTranslator(RBProgramNodeVisitor)>>visitNode:
OCASTTranslator>>visitNode:
[:each |
		self visitNode: each] in OCASTTranslator>>emitMessageNode: in Block: [:each |...
OrderedCollection>>do:
OCASTTranslator>>emitMessageNode:
```

Ducasse:
This is strange that it crashed again on the same because we can now recompile the complete image.

jecisc:
Related pheps: https://github.com/pharo-project/pheps/pull/22

Ducasse:
I prefer not to extend the parser for the sake of it. So I will close my PR and extension proposal. 
I think that the parser extension is only good if we can get a real gain such as not creating the array but I do not see how it would be possible.

PR #14852, opened 2023-10-02
Refactoring: Get away from class definition

jecisc:
This change update the full logic of the add class/trait change in refactorings. Instead of basing them on class definition we now base them on shift class builder.

Other changes includes:
- The merging of the add class, add metaclass, add trait and add class trait classes
- Fix a bug in the remove variable refactoring
- Fix a bug in ShiftClassBuilder where the superclass name was wrong when setting the superclass
- Add #classSlots to ShiftClassBuilder


Ducasse:
Wouahhhhhhhh Super super super cool!

Ducasse:
We should discuss the fold: removal.

PR #14859, opened 2023-10-03
14697-Deprecation-redirection-and-modification-doesnt-work-as-intended 

### LINKED ISSUE #14697: Deprecation redirection and modification doesn't work as intended
**Bug description**
The deprecation redirection (using the transformWith: argument) doesn't work correctly for Pharo 12.
Using it makes some tests don't work the first time, only the times after. It also doesn't modify the old functions for the new ones.

**To Reproduce**
1. Import this [Molecule fork](https://github.com/Eliott-Guevel/Molecule-various-fixes) by executing the following code in a Playground:
```smalltalk
Metacello new
   baseline: 'Molecule';
   repository: 'github://Eliott-Guevel/Molecule-various-fixes';
   load.
```
2. In the Molecule-Tests package, execute the test which is named "testSeveralComponentsLifecycle"
3. The test fails (see screenshot)
4. Relaunching the test works (and subsequent times too, except when relaunching the VM)

**Expected behavior**
The test should not fail the first time. In addition, the deprecated methods aren't changed for their new versions (instan**c**iateAllComponents should become instan**t**iateAllComponents)

**Screenshot**
node parent bug

**Version information:**
 - OS: Windows 11 Family
 - Version: 22621.2134
 - Pharo Version Pharo-12.0.0+build.768.sha.f2c9067e4cc97f23c4625f99b41be9e46197ef46 (64 Bit

**Additional context**
The test passes on Pharo 11 without any problem whatsoever.

MarcusDenker:
fixes the problem of compiled methods not being in sync with the byteocode that a recompile creates. This happens with Undeclared reads where we later load a definition.

fixes #14697

### DIFF
--- a/src/AST-Core/CompiledMethod.extension.st
+++ b/src/AST-Core/CompiledMethod.extension.st
@@ -3,6 +3,13 @@ Extension { #name : 'CompiledMethod' }
 { #category : '*AST-Core' }
 CompiledMethod >> ast [
 	"return an AST for this method. The AST is cached. see class comment of ASTCache"
+	
+	"It is important that the AST is in sync with the CompiledMethod. With Pharo12.
+	 we (for now) might end with a reflective variable access due to repaired Undeclared reads"
+	
+	((self hasLiteral: #runtimeUndeclaredRead) or: [self hasLiteral: #runtimeUndeclaredWrite:]) 
+			ifTrue: [self recompile].
+	
 	^ ASTCache at: self
 ]
 


MarcusDenker:
Hmm... need to think about that more. Now it recompiles, but it puts still  the wrong method in the cache

And my image is no broken.

But it fixes the problem (I guess as it sends #ast multiple times), so we could just merge it as the first workaround

MarcusDenker:
The problem is that if you have a CompiledMethod in hand and you want a mapping *for that method*, this fix will not work, as the method  that you have is the method that we can *never* create a bc-AST mapping for...

Maybe we need a flag in the compiler instead that re-creates exactly this Bytecode even though it is not needed.

(the whole problem comes from using bytecode as the one true meta model... it's wrong).

MarcusDenker:
Lots of tests failing... more work needed...

MarcusDenker:
And as trivial the PR is, if I load it I can not browse the one method it changes. 

I get a DNU "name was send to nil".

This is not relate, but it means that for the one line change I will close the PR and create a new one

MarcusDenker:
I think the problem is that we can not anymore just recompile a method, it somehow leads to RPackage do be not in sync (as it does not know the method)

PR #14865, opened 2023-10-04
14864-Recompiling-a-method-when-a-tool-looks-at-it-DNU-name-is-nil 

### LINKED ISSUE #14849: Calypso: Debugging a test mess up with Calypso
In Pharo 12 it happens often to me that I am debugging a test and then if I click on the name of the test in my calypso, it gets broken saying that #name was sent to nil. 

If this happens all calypso are broken and I can only fix it by closing all Calypsos and launching a GC. Else I get the bug all the time.

I don't know which action exactly during the debugging causes this

### LINKED ISSUE #14864: Recompiling a method when a tool looks at it: DNU "name is nil"
**Bug description**
- open "implementors of" for #yourself. 
- execute (Object>>#yourself) recompile 
- in the senders of browser, browse to another of the three methods, then back to the one of Object

Result: DNU "#name was send to nil".

MarcusDenker:
Even a compiledMethod that is not installed (e.g. because a new version was compiled) should know it's package

fixes #14864

### DIFF
--- a/src/RPackage-Core/CompiledCode.extension.st
+++ b/src/RPackage-Core/CompiledCode.extension.st
@@ -18,6 +18,6 @@ CompiledCode >> extensionPackage [
 CompiledCode >> package [
 	"This method returns the package this method belongs to. It takes into account classes and traits.
 	If the method is in no package, returns nil"
-
+	self isDoIt ifTrue: [ ^ nil ].
 	^ self extensionPackage ifNil: [ self origin package ]
 ]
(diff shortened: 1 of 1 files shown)


MarcusDenker:
testStackTablePackagesLabels is failing, it checks the DoIt case but does not use the correct selector (DoIt)

MarcusDenker:
For the test I did https://github.com/pharo-spec/NewTools/pull/595

jecisc (approved):
Oh thank you for this!

PR #14866, opened 2023-10-04
10094-SourceFileArraychangeRecordsFordo-is-Ring1-specific 

### LINKED ISSUE #10094: SourceFileArray>>#changeRecordsFor:do: is Ring1 specific
**Describe the problem**
SourceFileArray>>#changeRecordsFor:do: and the method it calls expects Ring1 objects, which is very strange.

SourceFileArray should not contain Ring1 specific code

MarcusDenker:
changeRecordsFor:do:  was sending #instanceSideParentName, which is Ring1 specific (the concept of "parent" does not exist in the main meta model).

The only user was creating a Ring Method, just to then extract sourcePointer, name and non-Meta class name (the API if SourcesFile is bad).

This PR fixes the sitation: we now can use CompiledMethods, too, and thus we do not rely on Ring anymore.

fixes #10094

### DIFF
--- a/src/Kernel/CompiledMethod.class.st
+++ b/src/Kernel/CompiledMethod.class.st
@@ -479,11 +479,6 @@ CompiledMethod >> isInstalled [
 ]
 
 { #category : 'testing' }
-CompiledMethod >> isMeta [
-	^self isClassSide
-]
-
-{ #category : 'testing' }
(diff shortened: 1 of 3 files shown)


jecisc (approved):
Nice :3

PR #14873, opened 2023-10-04
speedup-runtimeUndeclaredRead

MarcusDenker:
This adds a recompile when we call #runtimeUndeclaredRead (and write) on a Variable that is not Undeclared. As the Variable exists, the compiler can re-compile without the #runtimeUndeclaredRead send (and using pusLiteralVariable)


MarcusDenker:
Sadly, not that easy...

```
testUndeclaredReparationWithClass – MacOSX64.Kernel.Tests.Extended.Methods.CompiledMethodTest
testUndeclaredReparationWithGlobal – MacOSX64.Kernel.Tests.Extended.Methods.CompiledMethodTest
testUndeclaredReparationWithInstanceVariable – MacOSX64.Kernel.Tests.Extended.Methods.CompiledMethodTest
```

jecisc (approved):

PR #14874, opened 2023-10-04
Simplify Monticello tests

jecisc:
Monticello does a lot of things that seems useless in their setup/teardown. I'm trying to remove most of them. While doing that I found out that some of them are hiding really vicious bugs! 

I'm still fighting to find the root of those bugs but here is a first cleaning of things that seems to not trigger those bugs in order to not lose those improvements


jecisc:
included in #14875

PR #14878, opened 2023-10-04
Do not rely on old class definition in code importer

jecisc:
We have a lot of code that code exported in the past using the old class definition API and CodeImporter is executing this code to load it.

In this change I propose to update the DoItChunk to not execute the code when it includes the old class definition. Instead I not manipulate the AST and I use ShiftClassBuilder. This should allow us to remove all the old API. In case we are not in a class creation, we still evaluate the code.

This might interest @MarcusDenker and @Ducasse 

Future step:

I would like to update the chunk exporter to not export the old class definition in chunk anymore but a script of the class builder.


Ducasse:
Ah I see. 
BTW I think that we can remove the ephemeron:* and also subclass: because 
- where never used
- subclass: alone was never filled out.

jecisc:
Only one test failure left!

jecisc:
Tests are finally passing! :)

src/CodeImport/DoItChunk.class.st
```diff
@@ -149,8 +149,11 @@ DoItChunk >> buildClassFromOldDefinition: ast [
 	(#( CompiledBlock CompiledCode CompiledMethod ) includes: subclassName) ifTrue: [ layoutClass := CompiledMethodLayout ].
 
 	class := self class classInstaller make: [ :aBuilder |
+			| superclass |
+			superclass := self class environment at: superclassName asSymbol.
```
Ducasse: There is something strange that can break the UndefinedClass because the environment of the Chunk may not be the correct one. May be the classBuilder is working on another one. 
If we find a superclass in the environment of the DoItChunk then may be this is the wrong one. 
In any case I think that this is important to pass the name if we do not find the superclass because like that the builder can use the UndefinedClass as a superclass.

Ducasse:
may be we can integrate and check the UndefinedClass in another issue.

src/CodeImport/DoItChunk.class.st
jecisc: Tbh I tried with another implementation before but it wasn't working until I used this environment. We can try to see later what is the problem there and what environment we should use. 

When we execute the code directly, it used the global environment anyway so I guess this currently change nothing to the behavior of code import

Ducasse:
I integrate it now so that we can fix it for the better integration of the UndefinedClass

PR #14887, opened 2023-10-05
Monticello: Repackage classes and methods

jecisc:
In this change I propose to move all classes that are just present to be able to read old exported code of Monticello in "Monticello-BackwardCompatibility".


jecisc:
This should make it good to merge https://github.com/pharo-vcs/tonel/pull/116

jecisc:
I'll need to do in two steps

MarcusDenker (approved):

PR #14897, opened 2023-10-06
Add back #keyword in Pragma for compatibility

### LINKED ISSUE #14303: Pragma>>keyword has been lost... so breaking rules
May be keyword was deprecated. 
If this is the case, I do not really understand the point to remove deprecated message that ease migration and are no brainer!
Could we avoid to damage ourselves more than needed?

MarcusDenker:
add back #keyword as a compatibility method, we are thinking of  adding later a way to tag methods as compatibility similar to deprecation (that then will never be removed)

fixes #14303

### DIFF
--- a/src/Kernel/Pragma.class.st
+++ b/src/Kernel/Pragma.class.st
@@ -223,13 +223,16 @@ Pragma >> hash [
 Pragma >> key [
 	"Answer the keyword of the pragma (the selector of its message pattern).
 	 This accessor provides polymorphism with Associations used for properties."
-	^self selector
+	
+	^ self selector
 ]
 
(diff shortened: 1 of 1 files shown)


Ducasse:
Marcus why not starting to tag such method with <backwardCompatibility> so that we know it.

MarcusDenker:
It is in a "compatibility" category, which is kind of tag so we will find it when we have some real mechanism for managing compatibility methods. 

 I did not want to do more (e.g deciding on a pragma?) as this is not on my active list of things that I work on *now* and we should at least think a little bit before. Sometimes I just can't start yet another project...

PR #14900, opened 2023-10-06
Improving precision in humanReadableByteSizeString

jordanmontt:
The method `humanReadableByteSizeString` returns a string representing the byte string taking the binary value 1024. In the implementation it uses the integer division. The problem is that gives an imprecision of 1. It can be of 1 MB, 1 GB, etc. 

For example,
If one tries`6367000000 humanReadableByteSizeString "5 GB"`. But, the real binary value is `5.92 GB`. So, it can be misleading. This PR adds a precision of 2 decimal points. So now it correctly shows  `6367000000 humanReadableByteSizeString "5.92 GB"`

### DIFF
--- a/src/FileSystem-Core/Integer.extension.st
+++ b/src/FileSystem-Core/Integer.extension.st
@@ -12,8 +12,8 @@ Integer >> humanReadableByteSizeString [
 		inject: self
 		into: [ :value :each |
 			value < 1024
-				ifFalse: [ value // 1024 ]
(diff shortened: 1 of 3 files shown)


jecisc (approved):
This is way better IMO

Maybe we could add a version where we let the user parametrize the number of decimals?

jecisc:
We have 3 executable comments returning the wrong results. You can find them with the executable comments plugin of DrTest

PR #14902, opened 2023-10-06
Merge FluidClassBuilder and ShiftClassBuilder

jecisc:
This change merges FluidClassBuilder and ShiftClassBuilder

Both of those classes have an API that is really similar. FluidClassBuilder is just here to produce instances of ShiftClassBuilder and add the possibility to install it in the system. With this change we now directly create a shift class builder from the Fluid syntax. 

I think it is simpler to have only one builder in the image


jecisc:
Before merging I should check to salvage the comments. I forgot to merge the comments of both projects

jecisc:
I merged the comments. Let's see if the CI accepts this version

jecisc:
This seems to be tricker than I though

jecisc:
In the past, the ShiftClassBuilder was not filling the class slots when using #fillFor: and the ShiftClassInstaller had a hack to copy the class slots all the time except if the user was explicitly specifying `aBuilder classSlots: #()` in the ShiftClassBuilder declaration. 
#fillFor: is now fixed but the copy of class slots was still happening. I now unified the behavior so that class slots work the same as anything else in the class builder. I did that in collaboration with Marcus. 

I updated the tests accordingly.

Now let's see if the CI likes that better

src/Shift-ClassBuilder-Tests/ShiftTraitBuilderTest.class.st
```diff
@@ -134,34 +135,32 @@ FluidTraitBuilderTest >> testExistingTraitWithSlotsArePreservedIfChangingClassSi
...
-		           fillShiftClassBuilder;
 		           yourself.
-	shift := builder shiftClassBuilder.
-	self assert: shift name equals: #TPoint2.
-	self assert: shift slots size equals: 2.
-	self assert: shift category equals: self packageNameForTest , '-Mafia'
+	self assert: builder name equals: #TPoint2.
+	self assert: builder slots size equals: 2.
+	self assert: builder category equals: self packageNameForTest , '-Mafia'.
+	self flag: #askPablo
```
MarcusDenker: I removed all these #flag in a PR last week...

MarcusDenker (approved):

src/Shift-ClassBuilder-Tests/ShiftTraitBuilderTest.class.st
jecisc: I made sure to keep your changes but it seems that a merge did not like the class renaming and excluded again those changes :( I'll fix it asap!

jecisc: I checked and apparently this one was not fixed in your PR. I'll fix it here

Ducasse:
Why do we keep fluidBuilder in Behavior because we could remove it? no?

jecisc:
We probably can but I already changed a lot of things here. This can be a future cleanup :)

Issue #14904, opened 2023-10-06
Unified "status" concept for Processes


daniels220:
**Describe the request**
Process has methods like `isSuspended`, `isTerminated`, `isActiveProcess`, but they do not discriminate between all possible states a process can be in, and it's not always clear how their truth values relate. Dolphin Smalltalk has a method `Process>>status` (I think—might be `#state`) which returns a Symbol from the set:
* `#running`: the active process
* `#ready`: waiting in the ProcessorScheduler priority lists and will run as soon as processor time is available
* `#waiting`: waiting on a Semaphore
* `#suspended`: new process not yet resumed or has been sent `#suspend`
* `#dead`: terminated
* `#debug`: debugger attached

These states are mutually exclusive and clearly enumerate all possible states a process can be in. There are corresponding `isWaiting`, `isReady`, etc. methods as well. Something like this would be very useful in Pharo.

**Expected behavior**
For Pharo, we additionally have `isTerminating`, which still answers true once the process has _finished_ terminating. We could either change this but add an `isTerminatingOrTerminated`, and make `#terminating` a possible status, or just leave it alone as an extra on the side and have `#dead`/`#terminated` status only apply to fully-terminated processes.

`#debug` status carries significant extra complexity so could be omitted for now, though it would be very useful in general.

**Expected development cost**
The implementations of `isWaiting` and `isReady`, the two missing test methods, are straightforward as far as I can tell, and from there building a `#status` implementation isn't too hard. I would want review from someone familiar with the VM to ensure edge cases don't slip through though.

**Version information:** Any OS, any Pharo

daniels220:
Oh—I've been thinking of this for a while, just based on my experience with Dolphin, but what specifically prompted it was #14701/#14905—see comment there.

MarcusDenker:
I like the idea!

Ducasse:
Yes we should revise this part.

Ducasse:
@isCzech are you interested in this issue?

isCzech:
Hi, thanks for reminding me of this issue; I've already saved it for long winter evenings.

I agree having an exhaustive set of five mutually exclusive states a process can be in at any given moment would be helpful when learning (I struggled to internalize the concept of a process without having all five is<state> methods (*) corresponding to the states).

(*) isRunning, isReady, isBlocked, isSuspended, isTerminated

[...]

PR #14905, opened 2023-10-06
Force rescheduling of inactive process when changing its priority (fixes #14701)

### LINKED ISSUE #14701: Boosting the priority of a waiting process does not immediately schedule it to run
**Bug description**
If I have a low-priority process that is waiting in the `quiescentProcessLists`, and I assign it a higher `priority` than the active process, the active process continues to run until preempted by a preexisting higher-priority process.

Additionally, when changing the priority of any non-active process, it is not moved to the correct slot in the `quiescentProcessLists`, which will likely cause it to be scheduled once as if its priority had not been changed, though after that it will end up in the right place.

**To Reproduce**
Evaluate:
```
proc := [ Transcript crShow: 'In boosted process' ] forkAt:
	        Processor userBackgroundPriority.
Transcript crShow: 'In main proc before boosting'.
proc priority: Processor userInterruptPriority.
Transcript crShow: 'In main proc after boosting'.
3 * (10 ** 6) timesRepeat: [ 3 * 5 / 17 * 33 ].
Transcript crShow: 'In main proc after some busywork'.
```

which produces:
```
In main proc before boosting
In main proc after boosting
In main proc after some busywork
In boosted process
```

**Expected behavior**
```
In main proc before boosting
In boosted process
In main proc after boosting
In main proc after some busywork
```

**Version information:** Any OS / any Pharo version

**Expected development cost**
Off the top of my head, the right solution to this seems to be to make `Process>>priority:` a primitive, so it can atomically perform all necessary modifications to `quiescentProcessLists` and manually trigger a scheduling decision. Otherwise there is always the risk of the active process being interrupted by yet a third process while manipulating the processor. If it seems appropriate for a regular `yield` to also check for higher-priority processes this would be a good time to implement that as well. `ProcessorScheduler>>interpriorityYield:` is...concerning...to me, and it would be nice to eliminate it. I don't know enough about the VM to contribute to this, and while I suspect it's not _tremendously_ difficult, I imagine it's not trivial either.

As an alternative, it does seem to work to always suspend/resume the modified process, though it needs a different implementation (no forked process, the active process can handle it just fine) and a guard against resuming a previously-suspended process:

```
interpriorityYield: modifiedProcess
	"The original yield only works within processes of the same priority.
	The primitive does not try to resume a process in a higher priority.
	By suspending and resuming the modified process, we force it to be rescheduled.
	When modifying the active process, we must handle the resume from a forked process.
	We must also be careful not to resume a process that was suspended to begin with.
	(In particular, processes all receive #priority: as part of their creation, before
	it is safe to resume them.)

	This message should not be used directly.
	The only application so far is when changing the priority of a process.
	"

	modifiedProcess isActiveProcess
		ifTrue: [
			[ modifiedProcess resume ] fork.
			modifiedProcess suspend ]
		ifFalse: [
			modifiedProcess isSuspended ifFalse: [
				modifiedProcess
					suspend;
					resume ] ]
```

daniels220:
Need to reschedule only processes that are waiting in the ProcessorScheduler lists, not those that are suspended or waiting on a Semaphore. Seems like there should be a standard way to test for this condition.

### DIFF
--- a/src/Kernel-Tests-Extended/ProcessTest.class.st
+++ b/src/Kernel-Tests-Extended/ProcessTest.class.st
@@ -48,6 +48,81 @@ ProcessTest >> testActiveProcessFromProcesorShouldUseInstalledEffectiveProcess [
 ]
 
 { #category : 'tests' }
+ProcessTest >> testChangingOtherPriorityAffectsScheduling [
+
+	| val sem proc1 proc2 |
+	val := #(  ).
+	sem := Semaphore new.
+	proc1 := [
+	         val := val copyWith: 1.
+	         sem signal ] forkAt: Processor activePriority - 3.
+	proc2 := [
+	         val := val copyWith: 2.
+	         sem signal ] forkAt: Processor activePriority - 2.
+
+	proc1 priority: Processor activePriority - 1.
+	self assert: val equals: #(  ). "Sanity check"
+	sem wait.
+	"proc1 is recognized as higher-priority than proc2.
+	We get the processor back before proc2 runs."
+	self assert: val equals: #( 1 ).
+	sem wait.
+	self assert: val equals: #( 1 2 )
+]
+
+{ #category : #tests }
+ProcessTest >> testChangingOtherPriorityLowerDuringSemaphoreWait [
+
+	| semaphore hasBlockRun backgroundProcess |
+	semaphore := Semaphore new.
+	hasBlockRun := false.
+	backgroundProcess := [
+	                     semaphore wait.
+	                     hasBlockRun := true ] forkAt:
+		                     Processor activePriority + 1.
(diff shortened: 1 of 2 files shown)


daniels220:
BTW writing this is what prompted me to submit #14904 (as you can possibly guess from the consecutive issue numbers). I had initially checked for `modifiedProcess isSuspended ifFalse:` (rather than `(modifiedProcess suspendingList isKindOf: ProcessList) ifTrue:`), before discovering that this would kick loose a process that was previously waiting on a `Semaphore`. In a similar situation in Dolphin I would have said `modifiedProcess isReady ifTrue:`, and it would be nice to have the same option in Pharo.

MarcusDenker (approved):
We checked it at the sprint. To us it looks good. And yes, there should be a better way to check this (to not need isKindOf:), but that can be done in another step.

Issue #14611, opened 2023-09-06
Extract temporary variable doesn't correctly position temp definition


balsa-sarenac:
**Bug description**
When a class has multiple occurrences of a expression that you're extracting to a temp, new temp definition is added just above the expression you selected to inline, instead of the first occurrence.

**To Reproduce**
1. Go to 'RBRemoveMethodParametrizedTest>>#testModelRecursiveMethodThatIsNotReferencedFromOtherMethods'
2. Select on ''someMethodThatIsNotReferenced' asSymbol' 2nd or 3rd occurrence
3. Invoke Extract temp (I've created temp called `selector`)
4. Rule now complains that `selector` might be read before written

**Expected behavior**
Temp is written before first read.

**Additional information**
This is before:
```smalltalk
testModelRecursiveMethodThatIsNotReferencedFromOtherMethods

	| class otherClass |
	class := model classNamed: #Object.
	class
		compile: 'someMethodThatIsNotReferenced ^2'
		classified: #( #accessing ).
	self assert:
		(class definesMethod: 'someMethodThatIsNotReferenced' asSymbol).
	otherClass := model metaclassNamed: self class superclass name.
	otherClass
		compile:
		'someMethodThatIsNotReferenced ^3 someMethodThatIsNotReferenced'
		classified: #( #accessing ).
	model removeClassNamed: self class name.

	self executeRefactorings:
		(self createRefactoringWithModel: model andArguments: {
				 'someMethodThatIsNotReferenced' asSymbol.
				 #Object }).
	self deny:
		(class definesMethod: 'someMethodThatIsNotReferenced' asSymbol)
```

this is after:

```smalltalk
testModelRecursiveMethodThatIsNotReferencedFromOtherMethods

	| class otherClass selector |
	class := model classNamed: #Object.
	class
		compile: 'someMethodThatIsNotReferenced ^2'
		classified: #( #accessing ).
	self assert: (class definesMethod: selector).
	otherClass := model metaclassNamed: self class superclass name.
	otherClass
		compile:
		'someMethodThatIsNotReferenced ^3 someMethodThatIsNotReferenced'
		classified: #( #accessing ).
	model removeClassNamed: self class name.
	selector := 'someMethodThatIsNotReferenced' asSymbol.

	self executeRefactorings:
		(self createRefactoringWithModel: model andArguments: {
				 selector.
				 #Object }).
	self deny: (class definesMethod: selector)
```

[...]

PR #14906, opened 2023-10-06
Replace extracted expression only in containing scope (fixes #14771)

### LINKED ISSUE #14771: Extract temp affects neighboring scopes containing same expression
**Bug description**
If I have, for instance, two iteration blocks, both of which contain a particular expression somewhere, and I select it in one of them and use "extract temporary", the expression is replaced in both but the temp is only declared and initialized in one of them, resulting in a method that doesn't compile and wouldn't behave correctly if it did.

**To Reproduce**
I discovered this while inserting logging code in, of all places, `SourceFile>>tryOpenReadOnly:`—long story—so I'll use that as an example.
1. Select `each asFileReference / basename` in either of the iteration blocks in that method.
2. Extract to a temporary (I chose `loc`).
3. Notice that the other iteration block has also been affected:
```
tryOpenReadOnly: readOnly

	| basename |
	basename := path asFileReference basename.

	"Open a read write stream only if read only access was not requested.
	We need to create the encoding and buffering streams manually because we need a read write stream."
	readOnly ifFalse: [
		potentialLocations do: [ :each |
			[
			stream := SourceFileCharacterReadWriteStream
				          on:
				          (SourceFileBufferedReadWriteStream on:
					           loc unbufferedBinaryWriteStream)
				          encoding: 'utf8'.
			^ self ]
				on: Error
				do: [  ] ] ].

	potentialLocations do: [ :each |
		[
		| loc |
		loc := each asFileReference / basename.
		stream := ZnCharacterReadStream
			          on: loc binaryReadStream
			          encoding: 'utf8'.
		^ self ]
			on: Error
			do: [  ] ]
```

**Expected behavior**
Replace the extracted expression only in the scope in which the temp is declared.

**Screenshots**

**Version information:** Any OS, any Pharo

**Expected development cost**
The simple fix is to do what Dolphin Smalltalk does, and only replace the specific statement that was selected to extract with a reference to the temporary. Strictly speaking any more than this may not be a true refactoring, as the expression may have side-effects. In practice, I think much more commonly the programmer does really want all occurrences of that expression replaced with the temp, or they wouldn't be bothering to extract it to a temp in the first place. Fixing this should be a simple matter of only searching for other occurrences within the immediate enclosing scope. I'll make a PR when I have a chance.

daniels220:
When extracting a temporary, replace only those occurrences of the extracted expression that are in the scope in which the temp is now declared, not others elsewhere in the method.

Also avoid reassigning the parent of a node we may modify later--better to use a copy, I think.

Worth noting that there remains a hole if the expression selected to extract is not the _first_ occurrence within the containing scope—all of them will still be replaced even though, for the earlier ones, the temp will not have been initialized. This is easy enough to fix, the question is, do I fix it by moving the temp assignment ahead of the first occurrence of the expression, or by only replacing occurrences after the selected one?

### DIFF
--- a/src/Refactoring-Core/RBExtractToTemporaryRefactoring.class.st
+++ b/src/Refactoring-Core/RBExtractToTemporaryRefactoring.class.st
@@ -63,7 +63,7 @@ RBExtractToTemporaryRefactoring >> compileNewMethod [
 RBExtractToTemporaryRefactoring >> constructAssignmentFrom: aNode [
 	| valueNode |
 	valueNode := RBVariableNode named: newVariableName.
-	^RBAssignmentNode variable: valueNode value: aNode
+	^RBAssignmentNode variable: valueNode value: aNode copy.
 ]
 
 { #category : 'initialization' }
@@ -76,21 +76,19 @@ RBExtractToTemporaryRefactoring >> extract: anInterval to: aString from: aSelect
 
 { #category : 'transforming' }
 RBExtractToTemporaryRefactoring >> insertTemporary [
-	| node statementNode nodeReferences children |
+
+	| node statementNode nodeReferences |
 	node := self parseTree whichNodeIsContainedBy: sourceInterval.
-	(node notNil and: [ node isValue ])
-		ifFalse: [ self refactoringError: 'Cannot assign to non-value nodes' ].
-	children := self parseTree body allChildren.
-	nodeReferences := children select: [ :each | each = node ] thenCollect: [ :each | each ].
+	(node notNil and: [ node isValue ]) ifFalse: [
+		self refactoringError: 'Cannot assign to non-value nodes' ].
+	nodeReferences := node methodOrBlockNode allChildren select: [ :each |
+		                  each = node ].
 	statementNode := node statementNode.
-	nodeReferences do: [ :each | each replaceWith: (RBVariableNode named: newVariableName) ].
 	statementNode parent
-		addNode: (self constructAssignmentFrom: node)
-			before:
-				(node == statementNode
-						ifTrue: [ RBVariableNode named: newVariableName ]
-						ifFalse: [ statementNode ]);
-		addTemporaryNamed: newVariableName
+		addNode: (self constructAssignmentFrom: node) before: statementNode;
+		addTemporaryNamed: newVariableName.
+	nodeReferences do: [ :each |
+		each replaceWith: (RBVariableNode named: newVariableName) ]
 ]
 
 { #category : 'private - accessing' }
--- a/src/Refactoring-Core/RBExtractToTemporaryRefactoring.class.st
+++ b/src/Refactoring-Core/RBExtractToTemporaryRefactoring.class.st
@@ -63,6 +63,9 @@ RBExtractToTemporaryRefactoring >> compileNewMethod [
 RBExtractToTemporaryRefactoring >> constructAssignmentFrom: aNode [
 	| valueNode |
 	valueNode := RBVariableNode named: newVariableName.
+	"Use a copy so aNode retains its original parent.
+	In this case the caller is done making use of the parent backlink,
+	but it seems sensible to keep it intact whenever possible."
 	^RBAssignmentNode variable: valueNode value: aNode copy.
 ]
 
--- a/src/Refactoring-DataForTesting/RBClassDataForRefactoringTest.class.st
+++ b/src/Refactoring-DataForTesting/RBClassDataForRefactoringTest.class.st
@@ -152,6 +152,14 @@ RBClassDataForRefactoringTest >> demoMethodWithDuplicates [
 ]
 
 { #category : 'tests' }
+RBClassDataForRefactoringTest >> demoMethodWithDuplicatesInBlocks [
+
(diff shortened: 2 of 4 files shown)


Ducasse:
Thanks!

A bit of news from our side. The extract family of refactorings will be our next target for improvement (I imagine that you read the code and it is a bit complex) what we would like to do is to 
- use the same approach as in the ExtractMethodTransformation which is a composition of several more elementary transformations - However, we tried and discovered that the transformation does not extract certain expressions while the refactoring does. So we will have to revisit this.
- Right now we are cleaning the driver implementation that we introduced to be able to avoid uncomprehensible popups when having refactorings with failed preconditions. Basically the conditions must raise pop ups. We hope to finish that soon (we have papers to finish). Then we will focus on extract method because it is super important. Note that the API is bad: we should be able to have a node and not a position. 

S.

src/Refactoring-Core/RBExtractToTemporaryRefactoring.class.st
```diff
@@ -63,7 +63,7 @@ RBExtractToTemporaryRefactoring >> compileNewMethod [
 RBExtractToTemporaryRefactoring >> constructAssignmentFrom: aNode [
 	| valueNode |
 	valueNode := RBVariableNode named: newVariableName.
-	^RBAssignmentNode variable: valueNode value: aNode
+	^RBAssignmentNode variable: valueNode value: aNode copy.
```
balsa-sarenac, line 69: Can you explain why is this necessary? Did you had problems where you needed the original node? If so, which part of the code didn't work?

balsa-sarenac (commented):
Nice work and thanks for contributing!
I have only one question (you can see it below).
As for positioning of the extracted expression, there is already an issue for that (#14611). I'd suggest to address it in a separate PR. Furthermore, for your question where to put the new temp initialization expression, for me it makes the most sense to replace all usages in the scope and put it before first usage. I could not think of an example where I wouldn't want a duplication to be replaced with temp, can you think of a use case?

src/Refactoring-Core/RBExtractToTemporaryRefactoring.class.st
daniels220: That is a case of defensive programming, and perhaps it's overzealous. I like being able to rely on the `parent` backlink of parse nodes to be accurate, and TBH I really wish they were immutable after creation/copy-on-write. In this particular case I'm 99% sure the code will work without it, but note that if (for whatever reason) it were refactored so that `constructAssignmentFrom:` were called prior to retrieving the `statementNode`, it would fail because the parent backlink would be gone. I actually performed a related refactoring with the opposite effect, reordering statements to avoid the need for the `node == statementNode` check. I would be okay with removing the `#copy` though...on reflection I can see the argument that, though it might be nice to be more careful with the parent backlinks, that would be something better done all at once separate from other fixes.

(I was also wondering if there might not be a risk of modifying the One True Parse Tree of the method, the one that's cached and used for things like breakpoints, but I see that that is accessed through a separate selector `#ast` that nobody writing refactoring code is likely to think of using—so that's good.)

daniels220:
@balsa-sarenac I think 99% of the time you're right that replacing all occurrences is the best behavior. I do have some concern simply because it's not guaranteed to be a true refactoring, if the method(s) involved have side-effects. This is already true just by replacing more than one occurrence, but it gets worse to reason about if the assignment may get moved earlier in the method. But, I think the convenience in the common case (reducing duplication of simple chains of accessors) _vastly_ outweighs this concern. The ideal might be to have a checkbox "replace all occurrences", checked by default—I'm not sure how much UI work that would be.

I did leave a comment about the behavior when some occurrences are in _child_ scopes—that is, in-scope, but inside another block.

src/Refactoring-Core/RBExtractToTemporaryRefactoring.class.st
balsa-sarenac: I get your point, and I'm all in for defensive programming. I like the idea that these defensive features are introduced in a separate PR where they would also be explained with a comment saying what they're defending against. Because my first reaction was: "why we need copy here?" and a comment would help a lot.
I'm fine moving `#copy` to another PR, but I'm also fine with staying in this one, and in that case, I would kindly ask you to put a small comment :)

balsa-sarenac:
@daniels220 I agree, we have plans to create configurable UIs in the future. Right now we have higher priorities, so I'll creating an issue out of your comment that we can track and implement in the future.

Ducasse:
argh I wanted to integrate the change but we have a conflict.

daniels220:
@Ducasse Sorry, been super busy. I've rebased the branch—not sure what the conflicts were about, I didn't have to resolve any when rebasing...

src/Refactoring-Core/RBExtractToTemporaryRefactoring.class.st
daniels220: That makes sense, yeah—I'll do it separately in the future, for now to keep things simple I added comments (well, the same one in both locations).

balsa-sarenac:
Failing tests seem unrelated.

Issue #14871, opened 2023-10-04
Re-think Undeclared handling


MarcusDenker:
In Pharo12, Undecareds are now compiled as  a reflective access, for read "push Undeclared, send #runtimeUndeclaredRead"

This is nice, as we can raise an Exception (instead of just returning nil).

When a definition (e.g. a class) is loaded later, the Undeclared fixing will exchange the Undeclared var with the real variable. #runtimeUndeclaredRead there is implemented as a reflective read.

There are two problems with the reflective read / write:

1. It stays there until the method is recompiled, which is slower than needed
2. when asking for the AST, we re-compile from source (as the ast->source and bc->source mapping is what it is used for). Now the recompiled method does not use the reflective read, as the Variable can now be found. This means that the AST describes a mapping to different bytecode, leading to wrong mappings 

The problem of speed we can solve by recompiling in #runtimeUndeclaredRead for non-Undeclared variables

But the AST mapping mismatch is harder to fix. We can not even just add a flag to the compiler to compile all reads as reflective reads, as do not know which Variables used to be Undeclared. We could get that info by either storing it in a property or analysing the bytecode (the later is really ugly, and the former not nice either)

This Issue is here to be references from workarounds to be done

1) implement the recompile ob execution (if the compiler is available)
2) recompile in #ast of CompiledMethod, as in most cases we call it multiple times with the problem happening in later calls, this is the case in https://github.com/pharo-project/pharo/issues/14697

MarcusDenker:
Two PRs implement workaround for now:

- recompile on execution: https://github.com/pharo-project/pharo/pull/14873
- recompile on ast: https://github.com/pharo-project/pharo/pull/14872

MarcusDenker:
- Undeclared repair: Wrong repair if a Class Variable of the same name is added #14982

MarcusDenker:
Some numbers

When do we see lots of Undeclareds? When loading losts of complex code where a good load order is not possible (to always load def before use).

During full image load we see 1356 NewUndeclaredWarning notices in the CI logs. We only see it for globals, as we take care to load superclasses before subclasses (so all ivars are always found) and we made sure to not have Undeclareds outside of two test methods.

This means that we have max 1356 methods with Undeclareds, most likely less (as some have more than one access), in a practical case that might well be a "worst case".

MarcusDenker:
[...]

Issue #11903, opened 2022-11-16
declare:from: is badly named and badly located


Ducasse:
why not moving declare:from: in systemDictionary?
It looks like this is the only candidate 
and I really do not like the name of the selector because it destructively impacts the argument!

So at least it should be `transferKey: key from: aDictionary` 

```
declare: key from: aDictionary 
	"Add key to the receiver. If key already exists, do nothing. If aDictionary 
	includes key, then remove it from aDictionary and use its association as 
	the element of the receiver."

	(self includesKey: key) ifTrue: [^ self].
	(aDictionary includesKey: key)
		ifTrue: 
			[self add: (aDictionary associationAt: key).
			aDictionary removeKey: key]
		ifFalse: 
			[self add: key -> nil]
```

Ducasse:
In addition

```
valueAtNewKey: aKey put: anObject atIndex: index declareFrom: aDictionary 
	"Support for coordinating class variable and global declarations
	with variables that have been put in Undeclared so as to
	redirect all references to the undeclared variable."

	(aDictionary includesKey: aKey)
		ifTrue: 
			[self atNewIndex: index 
				put: ((aDictionary associationAt: aKey) value: anObject).
			aDictionary removeKey: aKey]
		ifFalse: 
			[self atNewIndex: index put: (Association key: aKey value: anObject)]
``` 
should be moved to systemDictionary and not dictionary
This way we can also remove the cancelling in Dictionary subclass Weak...

Ducasse:
And we can remove the same in DictionaryValueHolder

```
valueAtNewKey: aKey put: anObject atIndex: index declareFrom: aDictionary 
	
	value valueAtNewKey: aKey put: anObject atIndex: index declareFrom: aDictionary.
	self valueChanged.
```

MarcusDenker:
This is used with Undeclared, which is a normal Dictionary.

When I looked at this method last time,  I was thinking of adding a dedicated subclass of Dictionary for Undeclared... but it needs a change on the level of the bootstrap so it will take effort to do

jecisc:
Should this be in P11 or P12?

MarcusDenker:
Pharo 12, this is not a simple change

PR #14908, opened 2023-10-06
Optimize Dictionary>>unreferencedKeys

daniels220:
Examine each method, and each literal therein, only once, rather than once for each key in Undeclared. After uninstalling a whole bunch of packages, leaving some 7000 keys in Undeclared, all but 11 of them unreferenced, this takes the runtime of `removeUnreferencedKeys` from upwards of ten minutes IIRC to two seconds.

### DIFF
--- a/src/Collections-Unordered/Dictionary.class.st
+++ b/src/Collections-Unordered/Dictionary.class.st
@@ -933,10 +933,23 @@ Dictionary >> storeOn: aStream [
 
 { #category : 'undeclared' }
 Dictionary >> unreferencedKeys [
-	"This is private code for the Undeclared handling"
+	"This is private code for the Undeclared handling.
+	Optimize for performance--examine each literal only once.
+	Since we know these are variable bindings, they cannot be
+	optimized selectors nor be elements of Arrays, but we do need
+	to traverse nested blocks."
 
-	^self keys select: [ :key |
-			(self systemNavigation allReferencesTo: (self associationAt: key)) isEmpty ]
+	| unreferenced |
+	unreferenced := self keys asSet.
+	self systemNavigation allBehaviorsDo: [ :class |
+		class methodsDo: [ :method |
+			method withAllNestedLiteralsDo: [ :lit |
+				(lit isKindOf: UndeclaredVariable) ifTrue: [
+					unreferenced
+						remove: lit key
+						ifAbsent: [ "Sanity check--a binding with more than one reference may already be gone from the unreferenced set, but should be present in Undeclared itself"
+							self at: lit key ] ] ] ] ].
+	^ unreferenced
 ]
 
 { #category : 'private' }


jecisc:
Oh nice! I reported a few years ago that we had a huge regression on code removal because of this.
I tried to fix it by not removing undeclared and using a weak collection to hold undeclared and let the garbage collector do the work. But I didn’t had the time to finalize this. 

It is good to see improvements on this front.

I think the test failure is not related. I’ve seen it fail randomly and I need to check what causes this

jecisc:
I’ll let Marcus check before merging (probably monday) because he knows way better this part of the system than me

Ducasse:
I have a stupid question. Why do we scan all the behavior?

jecisc:
> I have a stupid question. Why do we scan all the behavior?

IIRC, during the removal it is to find classes that have no reference and can be removed, and to find the classes that are still references and that needs to be added to the undeclared collection

Ducasse:
I do not understand why this is related to a dictionary.
Sorry to be stupid. 
If this is about SystemDictionary then it has nothing to do in Dictionary. 
I do not even understand why dictionary is about classes.

daniels220:
This is a method of Dictionary, but which is only called on the specific dictionary that is `Undeclared`—which is _not_ a SystemDictionary—think of it like the other namespace-like dictionaries, like class and shared pools, but it's not attached to a class, it just lives directly in `Smalltalk globals`. Its contents are not classes, they're `UndeclaredVariable` bindings, whose value is...usually? always?...nil. They do tend to be caused by methods referring to nonexistent classes though. `Undeclared` exists so that two methods which refer to the _same_ undeclared name (again, very common when referring to a nonexistent class) will refer to the identical `UndeclaredVariable` binding, and in turn if that variable is later _declared_, references to the undeclared binding can be replaced by the real one, and the undeclared binding can for sure be discarded.

So, when a method is removed (or edited to no longer refer to a particular undeclared name), that binding _might_ no longer be used anywhere—but we don't know, some other method still might, so it can't be removed from `Undeclared`. So `Undeclared` can accumulate a lot of garbage, especially when installing or uninstalling packages that have circular reference dependencies, uninstalling packages without respecting prerequisite order, etc. This is where `removedUnreferencedKeys` comes in—it is called by, at least, the `ImageCleaner` after it finishes uninstalling packages. At this point we have no choice but to scan all methods of all existing/remaining behaviors, in order to determine which `Undeclared` keys are completely unused, and remove them.

It's basically a naive mark-and-sweep garbage collector for `Undeclared` keys...which is why Cyril's solution of using a weak collection for Undeclared seems like a superior option if it can be made to work—I took the approach I did because I just needed the process to _work_, but faster, and weak collections in Pharo have given me trouble before, never mind trying to use them deep in the guts of the system.

In the meantime, it occurs to me that nothing about this method requires it to live on Dictionary/have `Undeclared` as the receiver—it could as easily be implemented as a method of ImageCleaner or some other appropriate class and just manipulate `Undeclared` externally. So that might be a good approach to reduce confusion of "why is this a method of Dictionary-in-general".

Ducasse:
Thanks for the explanation and yes I knew it but forgot. This is ok to have it on Dictionary but it hurts my object-oriented view :). May be there is not enough specific points to have a specific class for Undeclared. We should discuss with @MarcusDenker

Ducasse:
I will merge it now. It would be good to chew the remark and see if we do something in the future.
I personally do not like to have a dependency to systemNavigation from a core class such as Dictionary.
Since it was already there I will merge. Now it means that we get deadcode when we only load dictionary. 
So I think that ImageCleaner or a friend would be good.

MarcusDenker:
- removeUnreferencedKeys/cleanOutUndeclared is not called on code loading or removal, just in cleanup related code
- we can remove some of the calls / unify  removeUnreferencedKeys/cleanOutUndeclared 
- Yes, we want to not have removeUnreferencedKeys and declare:from: in Dictionary, but use subclasses. But that is not that easy  and needs a bootstrap change. It's a project in itself and needs to be done with care, it can not be done just as part of another fix, nor should it lead to other fixes being delayed. (see https://github.com/pharo-project/pharo/issues/11903 )
- Yes, Undeclared growing is bad and we could e.g try to remove the key when we fix the undeclared or use a weak dict. 
- we need in general to re-think how we handle Undeclared Variables. see https://github.com/pharo-project/pharo/issues/14871

Ducasse:
But marcus we could also just package it in another class. Is it really needed to be in a subclass.

MarcusDenker:
Yes, I did that in this PR:

https://github.com/pharo-project/pharo/pull/14975

PR #14909, opened 2023-10-06
Allow formatting only part of a method or playground (fixes #14618)

### LINKED ISSUE #14618: Allow formatting only selected portion of a method or Playground
**Describe the request**
Sometimes I am editing a long method and want to clean up my new code, without introducing diff noise by reformatting the rest of the method. (Ideally there would be a standard code style enforced by the formatter, but this is a big job of its own and third-party code would still not necessarily conform.) It would be nice to be able to select a subset of the method and have formatting performed on that subset (assuming it parses on its own).

**Expected behavior**
Pressing Ctrl+Shift+F with a selection reformats only that part.

**Expected development cost**
I have a mostly-working solution, but it would be nice to handle indentation better when the selection starts not at the beginning of a line. Would also be nice to auto-expand the selection to find a parseable section if the initial selection doesn't parse.

**Version information:** any OS/any Pharo

daniels220:
First off, I added a basic implementation of notifying the user when the code fails to parse. Ideally I would use the nice pretty floating bubbles that a DoIt in a workspace uses, but I couldn't figure out how to make that work everywhere. In a Playground, saying `self model presenter interactionModel notify:at:in:` does the trick, but (a) law of Demeter, and (b) most other places, IIRC `self model` is a `RubScrolledTextModel` and DNU `#presenter`. I'm not sure the relevant objects are even instantiated, and I certainly don't know how to reach them. This is already an improvement over silently doing nothing, regardless.

Regarding partial formatting, there are some possible further enhancements, but all of them are much harder, yet probably less useful, than getting the basic functionality working:
* If the selection starts at or before the first non-whitespace character of a line, indent the entire formatted output starting with that line's indentation
* If the selection starts or ends in the middle of a token, automatically expand it to include the full token
* If the selection fails to parse, try expanding it to the closest containing parse node, if the method/workspace as a whoel *does* parse and such can be determined

### DIFF
--- a/src/Rubric/RubSmalltalkCodeMode.class.st
+++ b/src/Rubric/RubSmalltalkCodeMode.class.st
@@ -165,14 +165,22 @@ RubSmalltalkCodeMode >> classOrMetaClass: aBehavior [
 
 { #category : 'helper' }
 RubSmalltalkCodeMode >> formatMethodCode [
+
 	| source tree formatted |
-	source := self textArea text asString.
-	tree := self parseSource: source.
-	tree checkFaulty: [ :msg :pos | ^self editor notify: msg at: pos + 1 in: source].
+	self textArea hasSelection
+		ifTrue: [
+			source := self textArea selection.
+			tree := self parseExpression: source ]
+		ifFalse: [
+			source := self textArea text.
+			tree := self parseSource: source ].
+	tree checkFaulty: [ :msg :pos |
+		^ self editor notify: msg at: pos + 1 in: source ].
 	formatted := tree formattedCodeIn: classOrMetaclass.
-	formatted = source
-		ifTrue: [ ^ self ].
-	self textArea updateTextWith: formatted
+	formatted = source ifTrue: [ ^ self ].
+	self textArea hasSelection
+		ifTrue: [ self textArea replaceSelectionWith: formatted ]
+		ifFalse: [ self textArea updateTextWith: formatted ]
 ]
 
 { #category : 'shout' }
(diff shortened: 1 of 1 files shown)


Ducasse:
Hi daniels 
We will have a look and this is great! I love the idea to improve our IDE
Did you see that we added a lot of tests to have remember and spot potential regression on how we extract a selectors based on where the caret is placed? 
I tell you this because what I saw is that if I do not write tests then when we migrate to a new texteditor or code browser
the logic that I nicely designed gets lost. So may be you can encode also your behavior in tests by extending the selector Extractor tests. This way you will force us to maintain it :)

S

Ducasse:
I will merge it and we see :)

daniels220:
To be clear, at this point none of those further enhancements are implemented, just basic functionality, using the selection exactly as it is and replacing it with formatted output. I'm vaguely aware that extract method is smarter about expanding the selection, and yes, that could be a good place to start—though unlike extract method, this needs to work in the Playground and modified-but-unsaved method workspaces, so it may turn out that the same approach wouldn't work.

You're right, though, that I should write a couple tests of this functionality directly. Partly I just wasn't sure _where_ to do so—there's a solid half-dozen classes _involved_ in the process, and just because it was `RubSmalltalkCodeMode` that I modified doesn't mean that's the right place for the tests. Any suggestions?

Ducasse:
Thanks, daniel. This is cool to have you around. We truly appreciate your PR and general attitude. 
We are spending a lot of energy to make Pharo better and well this is a battle :). So this is cool to have people like you who understand that Pharo is a common and shared artifact. Thanks you again!

About my remark. What I learned when we started to improve the selection of selector is that I got totally confused because the extraction was buried in code. With the tests it helped us to think about what we wanted. 
I saw that Jean Privat improved the extraction and added extra tests and this is cool because like that we can all understand  what the system is doing without having to be expert on a given method. So for me it was a key features.

PR #14914, opened 2023-10-07
From isCzech: Add an "exit condition" to the UI loop to allow a former UI process to terminate

dionisiydk:
It is to update #11455 merged with Pharo12 and little change


Ducasse:
Thanks denis!

Ducasse:
I feel sorry about this issue because I'm not good in concurrent programming and changing this ground can introduce subtle bugs so we need to have free cycles to integrate it.

dionisiydk:
and the original PR is updated.

PR #14919, opened 2023-10-07
Fjx and add test for #14918: Behavior>>bindingOf: is different from RBClass >> bindingOf: 

### LINKED ISSUE #14918: Behavior>>bindingOf: is different from RBClass >> bindingOf:
```
testBindingOf

	| root sub |
	rbNamespace := RBNamespace onEnvironment: 	(RBClassEnvironment classes: {MyClassARoot . MySubAccessingSuperclassState }).
		
	root := rbNamespace classNamed: #MyClassARoot.
	sub := rbNamespace classNamed: #MySubAccessingSuperclassState.
	
	self assert: (root bindingOf: #Shared1) equals: (MyClassARoot bindingOf: #Shared1).
	self assert: (sub bindingOf: #Shared1) equals: (MySubAccessingSuperclassState bindingOf: #Shared1).
```

is failing in the second assertion

Ducasse:
Fix #14918 and adds a test

### DIFF
--- a/src/Refactoring-Core/RBAbstractClass.class.st
+++ b/src/Refactoring-Core/RBAbstractClass.class.st
@@ -170,8 +170,8 @@ RBAbstractClass >> binding [
 
 { #category : 'method accessing' }
 RBAbstractClass >> bindingOf: aString [
-	^self realClass classPool associationAt: aString asSymbol
-		ifAbsent: [self realClass classPool associationAt: aString asString ifAbsent: [nil]]
+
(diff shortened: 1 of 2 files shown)


jecisc (approved):
This seems good but I'd like to wait Tuesday to merge it so that marcus can review it too

Ducasse:
there is really nothing there and it is related to RBClass and without it many tests in Refactoring will break.
I will merge it. Because I want to merge the other refactoring PR which is more important in fact.

PR #14921, opened 2023-10-07
2023 10 07 remove shared var driver

### LINKED ISSUE #14800: Mixing selectors and Methods is a pain in RBClass
```
existingMethodsThatReferToInstanceVariable: aString
	| existingMethods |
	existingMethods := self realClass whichSelectorsAccess: aString.
	(newMethods isNil and: [ removedMethods isNil ])
		ifTrue: [ ^ existingMethods ].
	^ existingMethods
		reject: [ :each | (self hasRemoved: each) or: [ self newMethods includesKey: each ] ]
```

if whichSelectorsAccess is returning selectors then existingMethods is a selectors and not methods :(.

```
whichSelectorsReferToInstanceVariable: aString

	| selectors |
	selectors := Set new.
	newMethods
		ifNotNil: [ newMethods
				do: [ :each |
					( each refersToVariable: aString )
						ifTrue: [ selectors add: each selector ]
					]
			].
	self isDefined
		ifTrue: [ selectors addAll: ( self existingMethodsThatReferToInstanceVariable: aString ) ].
	^ selectors
```

if existingMethodsThatReferToInstanceVariable: returns what it says then we are adding methods in selectors :(
Terrible mess.

Ducasse:
Many things
- New RemoveSharedVariableDriver - all tests green
- Added class comments to drivers
- Removed old RemoveSharedVariableDriver
- Removed old Syc command for RemoveClassDriver
- Renamed SycRemoveClass and remove old one
- Removed old tests for the calypso removal strategy of the class removal refactoring
- Better comments
- More tests for definesClassVariable and directlyDefinesCLassVariable in RBClass
- new condition for sharedVariable accesses
- Making sure that RBSharedVariableHasReferences checks both class and instance side for use of sharedVariables!
- Fix #14800
- Add more tests
- Reimplemented fast RBClass >> bindingOf: so that the tests and new functionalities can pass


Ducasse:
I imagine that we may have a conflict. In such case we can merge dev into this PR.

balsa-sarenac (approved):
Looks good a lot of nice work!
I have just one question: we now have `RBRemoveSharedVariableRefactoring` and `RBRemoveSharedVariableRefactoring2`. And it seems to me that the second one is not used. Did I miss something and it is used somewhere, or we can remove it?

Ducasse:
I thought that I remove all the unnecessary code. In fact I removed one version and rename to the name of the previous one so may be it was confused because of that.

Ducasse:
The conflict is normal because this is the same code I needed and I was not patient enough to wait 2 days.

PR #14933, opened 2023-10-09
Cleanup-methodsWithUnboundGlobals 

MarcusDenker:
We have #testMethodsWithUnboundGlobals, which did not fail since years. If we look at "methodsWithUnboundGlobals", we see:

"However, we should investigate why the method gets an incorrect state."

So this is  for a bug that we fixed a long time ago... we can just remove the method and test


Ducasse:
So you do not want to keep it as a regression test? I deduce that it should not happen anymore.

MarcusDenker:
I do not see how this can ever happen with the current compiler

And if we start to add test for anything impossible... where do we stop?

Ducasse:
OK then let us remove it.

PR #14937, opened 2023-10-09
Unsubscribe announcements from finalization registry

guillep:
Announcements are not correctly removed from the finalization registry when explicitly removed.
This causes a bloat of ephemerons related to change sets that are created during the bootstrap.

Also, improve a bit the bootstrap to allow out of source builds.

I'll produce a backport to Pharo11

PR #14938, opened 2023-10-09
[P11 Backport] Unsubscribe announcements from finalization registry

guillep:
Backport of #14937

Announcements are not correctly removed from the finalization registry when explicitly removed.
This causes a bloat of ephemerons related to change sets that are created during the bootstrap.

Also, improve a bit the bootstrap to allow out of source builds.


jecisc:
Maybe for P11 we can just integrate the ephemeron fix and not the bootstrap update?

tesonep (approved):

bootstrap/scripts/4-build.sh
```diff
@@ -219,17 +219,11 @@ ${VM} "${PHARO_IMAGE_NAME}.image" "${IMAGE_FLAGS}" eval --save "Smalltalk vm sav
 ${VM} "${PHARO_IMAGE_NAME}.image" "${IMAGE_FLAGS}" eval --save "Smalltalk vm parameterAt: 43 put: 32"
 
 ${VM} "${PHARO_IMAGE_NAME}.image" "${IMAGE_FLAGS}" eval --save "MCCacheRepository uniqueInstance enable. FFIMethodRegistry resetAll. PharoSourcesCondenser condenseNewSources. Smalltalk garbageCollect"
-${VM} "${PHARO_IMAGE_NAME}.image" "${IMAGE_FLAGS}" clean --release
+#${VM} "${PHARO_IMAGE_NAME}.image" "${IMAGE_FLAGS}" clean --release
```
astares: Why is this commented out?

tesonep: Good catch, I have fixed it, thanks a lot.

Issue #10466, opened 2021-11-26
Calypso Tabs are still not lazy


Ducasse:
May be there is a mistake in the code of BeautifulComments but when I load the package from 

```
Metacello new
 	 baseline: 'BeautifulComments';
 	 repository: 'github://pillar-markup/BeautifulComments:master/src';
 	 load.
```

The code is broken and raise an error (which is good for this bug entry) which happens when you click on a class.

Now the bug is that when we click on a class and the comment pane is NOT activated/visible
the comment renderer is still executed.

Ducasse:
MicRichTextComposer class(Object)>>doesNotUnderstand: #renderComment:of:
ClyRichTextClassCommentEditorToolMorph>>asRenderedText
ClyRichTextClassCommentEditorToolMorph>>editingText
ClyRichTextClassCommentEditorToolMorph(ClyTextEditorToolMorph)>>buildTextMorph
ClyRichTextClassCommentEditorToolMorph>>buildTextMorph
ClyRichTextClassCommentEditorToolMorph(ClyTextEditorToolMorph)>>build
ClyRichTextClassCommentEditorToolMorph(ClyBrowserToolMorph)>>buildPageAndDecorate
[ self buildPageAndDecorate ] in ClyRichTextClassCommentEditorToolMorph(ClyBrowserToolMorph)>>addNotebookPageOn: in Block: [ self buildPageAndDecorate ]
ClyNotebookPageMorph(SpNotebookPageMorph)>>actualPageMorph
ClyNotebookMorph(SpNotebookMorph)>>pageAt:
ClyNotebookMorph(SpNotebookMorph)>>updatePageIndex:
ClyNotebookMorph(TabGroupMorph)>>update:with:
MorphChangedWithArguments>>deliverTo:
ClyNotebookMorph(Morph)>>handleUpdate:
MessageSend>>value:
MessageSend>>cull:
MessageSend>>cull:cull:
[action cull: anAnnouncement cull: announcer] in AnnouncementSubscription>>deliver: in Block: [action cull: anAnnouncement cull: announcer]
FullBlockClosure(BlockClosure)>>on:do:
[ Processor terminateActive ] in BlockClosure>>on:fork: in Block: [ Processor terminateActive ]

Ducasse:
What we discover is that the pane are first added and then removed
just put a self halt into `removeTabIndex:`

```
removeTabIndex: anInteger
	"Remove the tab at the given index and
[...]

PR #14942, opened 2023-10-09
Fix index out of range during Calypso editor tab updates

### LINKED ISSUE #12821: Index out of range using Calypso
Since a while I often get random bugs in Calypso and the error is always the same.

I don’t know how to reproduce exactly but here I got it while saving a method.

This happens on Pharo 11. 

Error message: `Error: index out of range`

Error stack:

```st
PanelMorph(Object)>>error:
PanelMorph(Morph)>>privateAddMorph:atIndex:
PanelMorph(Morph)>>replaceSubmorph:by:
ClyNotebookMorph(SpNotebookMorph)>>updateContentMorph:with:
ClyNotebookMorph(SpNotebookMorph)>>updatePageIndex:
ClyNotebookMorph(TabGroupMorph)>>update:with:
MorphChangedWithArguments>>deliverTo:
ClyNotebookMorph(Morph)>>handleUpdate:
MessageSend>>value:
MessageSend>>cull:
MessageSend>>cull:cull:
[action cull: anAnnouncement cull: announcer] in AnnouncementSubscription>>deliver: in Block: [action cull: anAnnouncement cull: announcer]
[ Processor terminateRealActive ] in [ :ex |
			  | onDoCtx handler bottom thisCtx |
			  onDoCtx := thisContext.
			  thisCtx := onDoCtx home.

			  "find the context on stack for which this method's is sender"

			  [ onDoCtx sender == thisCtx ] whileFalse: [
				  onDoCtx := onDoCtx sender.
				  onDoCtx ifNil: [ "Can't find our home context. seems like we're already forked
				and handling another exception in new thread. In this case, just pass it through handler."
					  ^ handlerAction cull: ex ] ].

			  bottom := [ Processor terminateRealActive ] asContext.
			  onDoCtx privSender: bottom.

			  handler := [ handlerAction cull: ex ] asContext.
			  handler privSender: thisContext sender.

			  (Process forContext: handler priority: Processor activePriority) resume.

			  "cut the stack of current process"
			  thisContext privSender: thisCtx.
			  nil ] in FullBlockClosure(BlockClosure)>>on:fork: in Block: [ Processor terminateRealActive ]

```

In PanelMorph>>#replaceSubmorph:by:, we try to get the index of the oldMorph in the list of submorphs. The problem is that my submorphs only contain a calypso method editor morph, and the old morph is a calypso class definition editor morph. Thus the index returned is 0 and there is a morph.

hernanmd:
This PR prevents Calypso from replacing non-existant morphs in tab group submorphs, which was causing an "index out of range" exception as reported in #12821.

See this use case to reproduce:

https://github.com/pharo-project/pharo/assets/4825959/0a553875-8688-46f3-8116-ed75fafeeaf9

This happens because when Calypso wants to update the its page content, the ClyMethodCreationToolMorph (which is used during the method creation) is no longer present in ClyNotebookMorph(SpNotebookMorph) submorphs - which is manifested by the fact the index of the Calypso tab group PanelMorph submorphs returning 0, meaning it is not found. This means that ClyMethodCreationToolMorph is replaced with a ClyClassDefinitionEditorToolMorph and then with a ClyRichTextClassCommentEditorToolMorph. It was also already reported that replacements are done twice in #10466, so the source of desynchronization is still yet to be found.

And this after the PR is applied:

https://github.com/pharo-project/pharo/assets/4825959/28f105ea-a55f-4983-b87a-cbc605f969c0

### DIFF
--- a/src/Calypso-Browser/ClyNotebookMorph.class.st
+++ b/src/Calypso-Browser/ClyNotebookMorph.class.st
@@ -109,6 +109,18 @@ ClyNotebookMorph >> toolbarMorph [
 	^ toolbarMorph
 ]
 
+{ #category : 'private' }
+ClyNotebookMorph >> updateContentMorph: aMorph with: newMorph [
+	| pageBounds |
+
+	pageBounds := self pageMorph bounds.
+	"We should replace aMorph which comes from pageMorph ONLY if it is present in the contentMorph (a PanelMorph), which is a container in the receiver"
+	(self contentMorph findA: aMorph class)
+		ifNotNil: [ : replaceMorph | self contentMorph replaceSubmorph: aMorph by: newMorph ].
+	self pageMorph bounds: pageBounds.
+	self pageMorph layoutChanged
+]
+
 { #category : 'accessing' }
 ClyNotebookMorph >> useSortedTabsBy: aBlock [
 


jecisc:
Thank you Hernan! This is one of the 2 ui bug that’s bothering me a lot! (The second is the one causing RSOD with a subscript out of bounds coming from rubric selection. It crashes a lot the CI)

hernanmd:
> Thank you Hernan! This is one of the 2 ui bug that’s bothering me a lot! (The second is the one causing RSOD with a subscript out of bounds coming from rubric selection. It crashes a lot the CI)

nice to see I wasn't suffering alone :)

jecisc (approved):

PR #14943, opened 2023-10-09
No need to remove empty protocols after loading morphic

MarcusDenker:
- No need to remove empty protocols after loading morphic (we do it for no other package)
- In addition, remove one explicit #garbageCollect


jecisc (approved):
This is so nice to see all this scripting code go away! :3

PR #14947, opened 2023-10-10
#BaselineOfIDE recompiles some classes in the postload

MarcusDenker:
It is not clear why (there are no comments).

This PR removes the recompile as it seems to be not needed


jecisc (approved):

MarcusDenker:
I will check

PR #14954, opened 2023-10-10
14934-When-prompted-to-create-a-class-the-definition-information-are-ignored

### LINKED ISSUE #14934: When prompted to create a class, the definition information are ignored
- define a method with a reference to an unknown class
- you will get prompted
- change the superclass to a class different than Object

Proceed, the resulting class is inheriting from Object :(

MarcusDenker:
fixes  #14934


Ducasse:
Cool

MarcusDenker:
sadly the test fails and the exception does not provide the needed infos to resume. So this has to be fixed.
(we could maybe split it in two and skip the two tests for now)

MarcusDenker:
I merge this and will fix the tests with another PR

PR #14969, opened 2023-10-11
[WIP] Preserve tree expansion state when (un)checking a class in the method visibility tree (fixes #14559)

### LINKED ISSUE #14559: Checking or unchecking a class in the "show inherited" list collapses the list
**Bug description**
When I open the list of inherited classes to set method visibility (rather than just click the one checkbox for "all except Object"), often I want more than one. But as soon as I check one, the list collapses and I have to reopen it to select another.

**To Reproduce**
1. Open a system browser on a class with a long inheritance chain.
2. Click the disclosure triangle next to "instance side" in the categories pane.
3. Check one of the superclasses.
4. Disclosure triangle collapses and the list of superclasses disappears.

**Expected behavior**
List should remain open/disclosed until I close it. IMO it should stay open when selecting a different class in the same system browser as well, just updating with the new list of superclasses, but that's less important.

**Expected development cost**
I came up with a fix, but it's pretty ugly:

```
ClyQueryViewMorph>>showQuery: aQuery

	self initiateUIChangeBy: [
		| oldDataSource |
		oldDataSource := self dataSource.
		self dataSource: (self dataSourceClass on: aQuery).
		self transferExpansionStateFrom: oldDataSource.
		self ensureSelectedItemIfNeeded ]

transferExpansionStateFrom: oldDataSource

	| itemsSelector actionSelector newItems |
	(oldDataSource isKindOf: ClyExpandedDataSource) ifTrue: [
		itemsSelector := #collapsedItems.
		actionSelector := #collapse ].
	(oldDataSource isKindOf: ClyCollapsedDataSource) ifTrue: [
		itemsSelector := #expandedItems.
		actionSelector := #expand ].
	itemsSelector ifNil: [ ^ self "Impossible?" ].
	newItems := self dataSource itemCursor findItemsSimilarTo:
		            ((oldDataSource perform: itemsSelector) collect: [ :each |
			             each browserItem ]).
	newItems do: [ :each |
		(self dataSource createElementWith: each) perform: actionSelector ]
```

My intent is, I hope, rather obvious: Preserve the expanded/collapsed state of items in the data source when showing a new query. But to do this, I need to choose a *very* particular way of querying. The underlying items are not *equal* because of the change in filtering—I need to match on the name (which `findItemsSimilarTo:` does and `findItemsWith:` does not). But `ClyDataSource>>findItemsSimilarTo:` also checks `self isBasedOnQueryOf: each type`, which I don't know how that's *meant* to work but it answers `false` here and refuses to even look for anything, so the best I've been able to do is a nasty Law of Demeter violation—talking directly to the `itemCursor` instead of the data source itself.

Someone who knows this part of the development system and/or Morphic better might be able to come up with a cleaner solution, but I'm kinda stumped. I also don't know whether there are places this method would be called—other parts of the development tools—where this change wouldn't be appropriate. This Morph seems to be used in the Critic and some kind of find dialog as well?

I'll add that the fact that we're tearing down and recreating the "data source" on toggle seems fundamentally wrong to me, and additionally the fact that collapsed and expanded are different classes of data source rather than just a state flag seems weird. And there's several levels deep of `initiateUIChangeBy:`, `changeStateBy:` etc. So it's...a bit of a mess.

daniels220:
This is obviously much too messy of an approach to integrate—Law of Demeter violations everywhere—but I'm hoping someone with more familiarity with the area can suggest how to clean it up.

### DIFF
--- a/src/Calypso-Browser/ClyQueryViewMorph.class.st
+++ b/src/Calypso-Browser/ClyQueryViewMorph.class.st
@@ -780,11 +780,12 @@ ClyQueryViewMorph >> showQueries: queries as: aQueryResult [
 { #category : 'queries' }
 ClyQueryViewMorph >> showQuery: aQuery [
 
-	| dataSource |
 	self initiateUIChangeBy: [
-		dataSource := self dataSourceClass on: aQuery.
-		self dataSource: dataSource.
-		self ensureSelectedItemIfNeeded]
+		| oldDataSource |
+		oldDataSource := self dataSource.
+		self dataSource: (self dataSourceClass on: aQuery).
+		self transferExpansionStateFrom: oldDataSource.
+		self ensureSelectedItemIfNeeded ]
 ]
 
 { #category : 'testing' }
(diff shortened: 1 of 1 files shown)


MarcusDenker:
Failing test no related

Ducasse:
Well let us first get a nice feature and we do another iteration.

PR #14972, opened 2023-10-11
Better integration of UndefinedClasses in CodeImporter

### LINKED ISSUE #14967: ShiftClassBuilder and UndefinedClass should be better integrated.
Right now the resolver (a strategy responsible to check the superclass does not exist) is only invoked from the ClassDefinitionParser!!!!?

```
ShiftClassBuilder >> classNamed: aName

	^ self buildEnvironment at: aName ifAbsent: [ nil ]
```

should invoke the resolver.

We should pay attention that classNamed: is also invoked on the class.

```
tryToFillOldClass

	oldClass ifNotNil: [ ^ self ].
	name ifNil: [ ^ self ].
	self oldClass: (self classNamed: self name)
```

Ducasse:
@jecisc please have a look

### DIFF
--- a/src/CodeImport/DoItChunk.class.st
+++ b/src/CodeImport/DoItChunk.class.st
@@ -159,7 +159,7 @@ DoItChunk >> buildClassFromOldDefinition: ast [
 				ifFalse: [ aBuilder useStrictSuperclass ].
 		  
 		aBuilder
-			  superclassName: superclassName;
+			  superclassName: superclassName asSymbol;
 			  environment: self class environment; "This is probably bad because it installs all chunks in the global environment but the bootstrap does not work without this for now"
 			  name: subclassName;
 			  layoutClass: layoutClass.
--- a/src/CodeImport/DoItChunk.class.st
+++ b/src/CodeImport/DoItChunk.class.st
@@ -160,7 +160,9 @@ DoItChunk >> buildClassFromOldDefinition: ast [
 		  
 		aBuilder
 			  superclassName: superclassName asSymbol;
-			  environment: self class environment; "This is probably bad because it installs all chunks in the global environment but the bootstrap does not work without this for now"
+			  "This is probably bad because it installs all chunks in the global environment but the bootstrap does not work without this for now"
+			  buildEnvironment: self class environment;
+			  installingEnvironment: self class environment;
 			  name: subclassName;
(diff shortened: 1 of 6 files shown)


MarcusDenker:
When doing the Full Image Build, we get a problem:

```
+ /builds/workspace/est_and_branch_Pipeline_PR-14972/64/bootstrap-downloads/vmtarget/pharo --headless Pharo12.0-PR-monticello_bootstrap-64bit-d0b1a7e.image --no-default-preferences st /builds/workspace/est_and_branch_Pipeline_PR-14972/64/bootstrap-cache/st-cache/Monticello.st --save --quit
Instance of CodeImporter did not understand #enableUndefinedClass
CodeImporter(Object)>>doesNotUnderstand: #enableUndefinedClass
buildClassFromOldDefinition: t1
  self error: 'No decompiler available' in DoItChunk>>buildClassFromOldDefinition: in Block: buildClassFromOldDefinition: t1...
ShiftClassInstaller>>make:
ShiftClassInstaller class>>make:
DoItChunk>>buildClassFromOldDefinition:
DoItChunk>>importFor:logSource:
evaluate
```

jecisc:
@Ducasse What should be the default value of enableUndefinedClass in CodeImporter?

Ducasse:
is it false so that we do not create UndefinedClass

jecisc:
I had this exacte same bug on another PR about the code importer and the problem is that the environment of the installed class is wrong and then during the lookup of the method, the method is not found because it's looked in the wrong class

src/CodeImport/DoItChunk.class.st
```diff
@@ -150,11 +153,14 @@ DoItChunk >> buildClassFromOldDefinition: ast [
...
-		  aBuilder
-			  superclass: superclass;
-			  environment: superclass environment;
+		  
+			importer enableUndefinedClass 
+				ifTrue: [ aBuilder useUndefinedClass ]
+				ifFalse: [ aBuilder useStrictSuperclass ].
+		  
+		aBuilder
+			  superclassName: superclassName;
```
guillep:
```suggestion
			  superclassName: superclassName asSymbol;
```

MarcusDenker:
this was merged in Pharo12. Is that correct?

MarcusDenker:
ah, no, closed

PR #14975, opened 2023-10-12
Deprecate Dictionary #removeUnreferencedKeys 

MarcusDenker:
- make sure nobody uses #removeUnreferencedKeys
- deprecate it
- move #unreferencedKeys implementation to #cleanOutUndeclared


pavel-krivanek (approved):

daniels220:
Nice, thanks! I was too focused on cutting down the runtime to think of moving the implementation, but this is definitely better.

PR #14983, opened 2023-10-12
Make pseudo-variables prioritary in name resolution

guillep:
Pseudo-variables (self, super, thiscontext...) can be overridden by class variables and class pool variables.

With the following code, `Toto new wat` returns nil instead of the instance of `Toto`.

```smalltalk
Object << #Toto
	slots: {};
	sharedVariables: { #self };
	tag: 'Objects-';
	package: 'Kernel'.

Toto >> wat
 ^ self
```

 with tests

### DIFF
--- a/src/OpalCompiler-Core/OCASTSemanticAnalyzer.class.st
+++ b/src/OpalCompiler-Core/OCASTSemanticAnalyzer.class.st
@@ -116,11 +116,14 @@ OCASTSemanticAnalyzer >> outerScope: anObject [
 
 { #category : 'variables' }
 OCASTSemanticAnalyzer >> resolveVariableNode: aVariableNode [
+
 	| var |
-	var := (scope lookupVar: aVariableNode name)
-		ifNil: [ self undeclaredVariable: aVariableNode ].
+	var := (OCPseudoVariableScope new
+		        outerScope: scope;
+		        lookupVar: aVariableNode name) ifNil: [
+		       self undeclaredVariable: aVariableNode ].
 	aVariableNode variable: var.
-	^var
+	^ var
 ]
 
 { #category : 'initialization' }
--- a/src/OpalCompiler-Core/OCPseudoVariableScope.class.st
+++ b/src/OpalCompiler-Core/OCPseudoVariableScope.class.st
@@ -1,0 +1,20 @@
+"
+I model a scope that resolves pharo pseudo-variables such as: `self`, `super` and `thisContext`.
+
+I should by default put at the beginning of the chain when resolving names so these names are resolved with more priority than others.
+"
+Class {
+	#name : 'OCPseudoVariableScope',
+	#superclass : 'OCAbstractScope',
+	#category : 'OpalCompiler-Core-Semantics',
+	#package : 'OpalCompiler-Core',
+	#tag : 'Semantics'
+}
+
+{ #category : 'lookup' }
+OCPseudoVariableScope >> lookupVar: name [
+
+	^ PseudoVariable lookupDictionary
+		  at: name asSymbol
+		  ifAbsent: [ super lookupVar: name ]
+]
--- a/src/OpalCompiler-Tests/OCASTVariableTranslatorTest.class.st
+++ b/src/OpalCompiler-Tests/OCASTVariableTranslatorTest.class.st
@@ -101,6 +101,16 @@ OCASTVariableTranslatorTest >> testPushSelf [
 ]
 
 { #category : 'tests - variables' }
+OCASTVariableTranslatorTest >> testPushSelfRedefined [
+
+	self assert: instance result equals: nil.
+	instance class addClassVarNamed: #self.
+
+	self testExample: #exampleSelf.
+	self assert: instance result equals: instance
+]
+
+{ #category : 'tests - variables' }
 OCASTVariableTranslatorTest >> testPushSuper [
 
 	self assert: instance result equals: nil.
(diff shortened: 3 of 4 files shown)


MarcusDenker:
But is this really a huge problem? You should get a warning that you shadow the variable in the tools.

Kind of "everything should be possible (for experiments), but the user should be guarded one level above".

guillep:
> But is this really a huge problem? You should get a warning that you shadow the variable in the tools.
> 
> Kind of "everything should be possible (for experiments), but the user should be guarded one level above".

Actually, we are working with alternatives to the system dictionary, so having it at that level forces us to duplicate the code of resolving the names of pseudo-variables. Also, this looks a better place to me :)

src/OpalCompiler-Core/OCASTSemanticAnalyzer.class.st
```diff
@@ -116,11 +116,14 @@ OCASTSemanticAnalyzer >> outerScope: anObject [
 
 { #category : 'variables' }
 OCASTSemanticAnalyzer >> resolveVariableNode: aVariableNode [
+
 	| var |
-	var := (scope lookupVar: aVariableNode name)
-		ifNil: [ self undeclaredVariable: aVariableNode ].
+	var := (OCPseudoVariableScope new
+		        outerScope: scope;
```
dionisiydk, line 122: It should be really done at the level of scope construction. Then it can be easily overridden with special compiler configuration.

Ducasse: Thanks we will check.

guillep: Yes, I agree, but that is a feature outside of the scope of this PR.
We could do it afterwards if there is the need.

The good thing about this PR is that the resolution of pseudo-variables becomes part of the scope chain, before it was hardcoded in the global symbol resolution.

src/OpalCompiler-Core/OCPseudoVariableScope.class.st
```diff
@@ -0,0 +1,20 @@
...
+	#superclass : 'OCAbstractScope',
+	#category : 'OpalCompiler-Core-Semantics',
+	#package : 'OpalCompiler-Core',
+	#tag : 'Semantics'
+}
+
+{ #category : 'lookup' }
+OCPseudoVariableScope >> lookupVar: name [
+
+	^ PseudoVariable lookupDictionary
```
MarcusDenker, line 17: A problem here is that you create new PseudoVariables for evert method. This is problematic, as we now have one instance for every definition of a variable, and we use that e.g to for #usingMethods (if not overridden to use the bytecode for speed) and for AST related queries (#readNodes, for example).

Of course this is a question of definition, but this was one of the nice aspects of how it is now.

MarcusDenker: e.g. you can do now 
```
(Smalltalk globals lookupVar: #thisProcess) readNodes
```

(it is slow as it has to create the AST for every method)

src/OpalCompiler-Core/OCASTSemanticAnalyzer.class.st
MarcusDenker: One place where we need to take care is DoIts: in the DoIt, all the PseudoVariables are now wrapped in a DoItVariable which is then added to the scope to shadow the PseudoVariable.  This way when you inspect "thisContext" in the debugger, you get the context that the debugger looks at, not the context of the DoIt evaluation.

This means that we can not look up PseudoVariables as the very first thing, they have to be at least *after* the Scope that we add for the DoIt (this might already be the case, but it has to be checked).

src/OpalCompiler-Core/OCPseudoVariableScope.class.st
MarcusDenker: Another nice aspect of the current way  (and maybe one of the main reasons why it is done like that): you can lookup the one instance of a pseudo variable reflectively outside of the compiler. If the PseudoVariableScope is created just for compilation, this is not possible. 

```

var :=  (Smalltalk globals lookupVar: #self).

"now I can use it reflectively, in the doit it is nil of course"

var readInContext: thisContext.

"like every var it knows it's scope, if I look my name up I get myself"

(var scope lookupVar: var name) == var "true"
```

MarcusDenker: But making the scope of PseudoVariables explicit is nice... one idea: could it not be the parent or immediate child scope of the global scope (depending if we want to allow them to be shadowed from the global or not).

In the current setup, the scopes of Variables that are always allocated (Slots, Class Vars, Globals) are explicit, always allocated scopes. They are not subclass of OCAbstractScope, but they do implement the API and *are* the scopes of the Variables (they form the scope hierarchy directly above the method scope).

```
var := Smalltalk globals lookupVar: #Array.
var class "GlobalVariable"
var scope.  "a SystemDictionary(lots of globals)"

var := SmalltalkImage classVariableNamed: #CompilerClass.
var class  "ClassVariable"
"the scope of a Class Variable is the class"
var scope  "SmalltalkImage" 
```

This for sure we do not want to loose if we get rid of the global scope. E.g. I would assume that the Variables that are defined by a "Module" would have that module as their scope.

In a way, the PseudoVariables are in a not-explicit child scope of the global scope right now, but we could make it explicit and have it as a child or parent scope now of the global scope.  

For"Module" scopes I guess one could have one instance of the PseudoVariableScope per module.

Ducasse: I do not really get why the namespace is responsible for the scope of self. 
I do not understand why self is not looked up first. Because we should not be able to define an instance variable self or a classvariable. Could we make sure that we do not introduce more ways to break the code. So I do not get why having a scope for self by method is a problem.
When we compile a method then the method has a scope and this scope contains self 
I do not understand why something simple like that is not simple.

MarcusDenker: We should discuss.

guillep:
I'll close this PR, it needs to be re-done in a different way.

PR #14984, opened 2023-10-12
Add tests on Fluid class parser and fix bugs

### LINKED ISSUE #14979: buildClassFromAST: and all the users should be migrated to FluidClassDefinition
ASTClassBuilderTest, undefinedClassTests are using the old class format

jecisc:
This changes aim to add tests on the creation of classes from the FuildClassDefinitionParser.  I ported the tests from ASTClassBuilderTest and added a missing one on nil management.

I also fixed a few bugs:
- Shared pools parsing was broken
- nil as a receiver was not working
- Trait as a receiver was not working

For now I keep the tests on the old class parser in ASTOldClassBuilderTest because this is still used by Calypso. We can remove them once we remove the `[v] Fluid`  checkbox from Calypso.

On other subject, I migrated UndefinedClassTest to use the fluid class definition.

Fixes #14979


Ducasse:
Once we will have Fluid only as a syntax I will probably rewrite the fluid parser using Smacc because the manual parsing and pattern matching was a not so good idea. 
But this is good that you fix it. I find strange that we did not spot this bug before.

jecisc:
Tests passed, we can now merge this one

MarcusDenker (approved):

PR #14992, opened 2023-10-13
Adding debug test setup debug menu entry

### LINKED ISSUE #14867: [Feature Request] Debug setup in debug test
I am using a lot the "Debug test" feature of Calypso. But often I want to see the code from the setup also. 

Currently I have to go put a breakpoint in the setup. Maybe we could have a better Debug test feature that allows use to debut the setup, test and tear down?

I don't know yet how to present that to the user. Maybe different menu entries?

StevenCostiou:
I force myself to do small improvements every friday, so: fixes #14867

Looks like that:
Capture d’écran 2023-10-13 à 15 20 38

It actually just moves the breakpoint up into the setup method, and then executes the selected test.

I could make the code better, but I prefer to wait for feedback.

### DIFF
--- a/src/Calypso-SystemPlugins-SUnit-Browser/ClyBrowseCoveringTestCaseCommand.class.st
+++ b/src/Calypso-SystemPlugins-SUnit-Browser/ClyBrowseCoveringTestCaseCommand.class.st
@@ -2,22 +2,24 @@
 I am a command that opens a new browser on the test class corresponding to the selected class.
 "
 Class {
-	#name : #ClyBrowseCoveringTestCaseCommand,
-	#superclass : #ClyBrowserCommand,
+	#name : 'ClyBrowseCoveringTestCaseCommand',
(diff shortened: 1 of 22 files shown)


StevenCostiou:
Comment: might not be the best way to bring this feature to the users, but at least it is there until something better is done ;)

jecisc:
Oh this is cool thanks!

I think that this is good for a first iteration. Later we might group things in a "Debug test" menu with multiples entries: "debug test", "debug setup", "debug tear down"?

I needed that since so long :)

jordanmontt:
It wouldn't be better to put the breakpoint in the setup by default? Normally, one wants to see what happens in there

jecisc:
I'd like to be able to do both personally. Most of the time I want to debut only the test. And sometimes I want to debug the setup also (around 10% of the time?)

jecisc:
I'll merge for now and we can iterate later

jecisc (approved):

PR #14994, opened 2023-10-13
Add `newTenured` allocation methods

jordanmontt:
This PR introduces two new api methods `newTenured` and `newTenured:`. They serve to allocate the requested objects directly in the old space (a.k.a. pre-tenuring).

This is useful when you know that you have an object that will live for a long time. It saves the garbage collector from moving the object from different memory positions when it is in the young space and then moving the object to the old space (a.k.a. tenuring).  It gives the power to the user to do memory optimizations, such as pre-tenuring.

**Before merging**

This PR requires before merging:

- [x] the merging of https://github.com/pharo-project/pharo-vm/pull/701
- [x] merge backport PR https://github.com/pharo-project/pharo-vm/pull/709
- [x] a new vm release of the pharo-10 vm version.


jecisc (approved):
Oh, I really really like this idea! Thank you

Ducasse:
That sounds good but let us see what the vm guys say!

jordanmontt:
Update, PR in the pharo vm merged.
Now it is only missing the new vm release. :)

astares:
Cool idea and nice work - I've seen the VM changes.

Can we please have a better intention revealing selectors than #newOld ? The selector #newOld is confusing and a contradiction somehow if one reads this first time.

What about #allocateNewInOldSpace, #allocateInOldSpace or #newInstanceInOldSpace or someting similar?

jordanmontt:
Thanks @astares 

I agree that `newOld` can be confusing, but I chose that name to be coherent with the other allocator methods such as `new` `newPinned`, etc. `allocateInOldSpace` is too verbose. I propose to name it `newOldSpace` to keep the coherence with the other allocator methods.

jecisc:
Maybe `#newInOldSpace`?

guillep:
I propose `newTenured` which is more aligned with the intent of the method (rather than the implementation).
Of course, people would need to understand what it is to tenure an object, that should be in the method comment, but this would be rather a warning saying that this should be used with caution and that users should be aware of the semantics of tenuring.

guillep:
BTW @jordanmontt this will not be available with a new release because the change was integrated in the Pharo12 branch of the VM which is the unstable branch. Pharo12 is still using the Pharo10 branch of VMs. So if this is to be available, we should back-port it for now (Putting the Pharo12 branch of the VM in production requires a bit more work)

jordanmontt:
Thanks @guillep for your feedback. I have done the things that you requested.

jordanmontt:
Too many merge conflicts. I will close this and open a new one

PR #14996, opened 2023-10-13
Fix Jenkins file uploading stage

Rinzwind:
This pull request tries to fix the Jenkins stage ‘Upload to files.pharo.org-64’ which is broken (see [build 900 of the ‘Pharo12’ job](https://ci.inria.fr/pharo-ci-jenkins2/blue/organizations/jenkins/Test%20pending%20pull%20request%20and%20branch%20Pipeline/detail/Pharo12/900/pipeline/31/)).


jecisc (approved):
Seems good reading it since we are now one folder deeper but the PR CI run does not execute this. I'll merge to see if it fixes the problem really. Thanks

PR #14926, opened 2023-10-08
Fix ‘span’ in BalloonEngine not being as wide as the ‘destForm’

Rinzwind:
This pull request makes `#bitBlt:` on BalloonEngine set a new ‘span’ when ‘bitBlt’ is set so that the width of the ‘span’ is not less than the width of the ‘destForm’ of the new ‘bitBlt’. This fixes issue #14925. The method `#initializeSpan` is removed as `#initialize` also sends `#bitBlt:`.

PR #14998, opened 2023-10-13
Refactor ImageMorph to support more than one Form depicting the same image at different scales

Rinzwind:
This pull request:

* Introduces a class FormSet for representing images for which there can be more than one Form depicting the same image at different scales
* Adds methods to Canvas for drawing a FormSet
* Refactors the ImageMorph hierarchy to replace instance variables that hold a Form by ones that hold a FormSet instead
* Makes ScrollBarMorph use FormSet for the arrows on the buttons at scale 1 and 2

The distinction between FormSet and Form is akin to that between [NSImage](https://developer.apple.com/documentation/appkit/nsimage?language=objc) and [NSImageRep](https://developer.apple.com/documentation/appkit/nsimagerep?language=objc) in macOS’s AppKit. A similar concept is that of an [`` element with a ‘srcset’ attribute in HTML](https://webkit.org/blog/2910/improved-support-for-high-resolution-displays-with-the-srcset-image-attribute/).

I initially used ‘Image’ as the name for the class but found that that got confusing as ‘image’ is already used in many selectors and variable names as a synonym for ‘form’. Once FormSet is used more and ‘image’ becomes more of a synonym for ‘form set’, the class could still be renamed to ‘Image’ as well.

The handling of the ‘none’ case in `FormSet>>#asFormWithExtent:` could be improved by scaling the Form that most closely matches the given extent rather than just the first Form, and the result should perhaps be cached, but there’s no need for either of that at the moment.

Note that loading these changes into an existing Pharo image probably requires using headless mode.

### DIFF
--- a/src/Athens-Morphic/CheckboxButtonMorph.extension.st
+++ b/src/Athens-Morphic/CheckboxButtonMorph.extension.st
@@ -9,12 +9,12 @@ CheckboxButtonMorph >> drawOnAthensCanvas: anAthensCanvas [
 	anAthensCanvas drawShape: self bounds.
 	anAthensCanvas setPaint: self borderStyle.
 	anAthensCanvas drawShape: self bounds.
-	img := self imageToUse.
+	img := self formSetToUse ifNotNil: [ :formSetToUse | formSetToUse asForm ].
 	img
 		ifNotNil: [
 			anAthensCanvas setPaint: img.
 			anAthensCanvas drawShape: (self innerBounds center - (img extent // 2) extent: img extent) ].
-	((self state == #pressed or: [ self state == #repressed ]) and: [ image isNil ])
+	((self state == #pressed or: [ self state == #repressed ]) and: [ formSet isNil ])
 		ifTrue: [
 			anAthensCanvas setPaint: (self paneColor alpha: 0.3).
 			anAthensCanvas drawShape: self innerBounds ].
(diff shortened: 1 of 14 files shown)


Rinzwind:
I applied this to the [‘Retina display’ scaling experiment from earlier pull requests](https://github.com/pharo-project/pharo/pull/14926#issuecomment-1752209059): commits 482bcd7289e596d1 and a7e272b02bb313a9 (these commits are not included in this pull request, the branch is ‘scaling-canvas-tryout-10’).

In the extra world opened by `OSWorldRendererScalingCanvas example`, previously the buttons of scrollbars looked a little blurry as they still used a scale one Form for the arrow, that has now been fixed:

<p align="center">

</p>

ScalingCanvas still uses its ‘FormMapping’ for remapping other forms from the UITheme and ThemeIcons, that should be dropped by using FormSet for those as well.

jecisc (approved):
This seems ok but it's the first time I see most of the code edited so if someone else wants to take a look I would prefer it

Ducasse:
This is a nice design. 
Could you have some tests and comments on FormSet? for example to illustrate how the extent of a set is computed and how a FormSet is created (for example betteer comment in forms: method). 
a test showing how the right form of the formset is selected would be nice too.

Rinzwind:
I added FormSetTest and extended the comment on FormSet.

Ducasse:
Coool!

PR #15001, opened 2023-10-14
Rework baseline of refactorings

### LINKED ISSUE #14826: BaselineOfRefactoring is not well done
Refactoring model is in BasicTools.
Refactoring tests are in BaselineOfRefactoring.
Rules are in BaselineOfRefactoring.

We should have the refactoring model and tests in BaselineOfRefactoring and we should have rules somewhere else

jecisc:
This change merge BaselineOfRefactoring and BaselineOfRefactoring2 + remove what was declared directly in BaselineOfBasicTools.

Depends on #14999 (To avoid the conflict with the removal of the rules from the refactoring baseline)

Fixes #14826


Ducasse:
cool!
Thanks for the time spent!

PR #15004, opened 2023-10-14
RPackage: Change root tag name

jecisc:
Up to now the root tag name (the tag corresponding to untagged classes) had the same name as the package. This is not so great because it means that the name is different for each package (which makes the tooling harder to make) and this caused some bugs if you create a tag that have the name of your package. For example a package Model with a tag Model.

I propose to use the fixed name of #Uncategorized instead of the name of the package


jecisc:
This should now be possible after a few weeks of cleaning :)

PR #15010, opened 2023-10-15
Clean traits tests

jecisc:
This change cleans traits tests by simplifying the cleaning and updating the creation API to mimic the FluidClassBuilder


Ducasse:
Cyril I was wondering why the reason of the change because this is an API internal to the tests.

jecisc:
Some of the cleanings are good but it's true that some other are not so useful. I did them with a script while looking for the origin of a bug. I'll see to update this PR this week to keep only the impactful changes

Ducasse:
I thought so :)

jecisc:
Better fix: https://github.com/pharo-project/pharo/pull/15243

PR #15016, opened 2023-10-16
14945-Class-Variables-should-not-be-sorted-Alphabetically 

### LINKED ISSUE #14945: Class Variables should not be sorted Alphabetically
Class Variables are always shows sorted alphabetically in the class defintion.

This is different to instance Variables, which retain the order that the programmer used.

Why do we sort? Because the names are the keys in a Dictionary, and order of that is undefined.

```
classVarNames
	"Answer a collection of the receiver's class variable names."
	<reflection: 'Class structural inspection - Class variable inspection'>
	^self classPool keys sort
```

The automatic sort is not nice
- Models should never sort anything (this is a view responsability)
- I can not give an order as the developer
- if I add a class variable and accept, it moves to another position when I look at the class again

# What would be better? 
Retain the order that the variables have according to how the developer added them.

# How can we do that?
We could expore the use of an OrderedDictionary for the classPool

MarcusDenker:
fixes #14945:

use an OrderedDictionary for the classPool

### DIFF
--- a/src/Collections-Sequenceable/OrderedDictionary.class.st
+++ b/src/Collections-Sequenceable/OrderedDictionary.class.st
@@ -535,6 +535,11 @@ OrderedDictionary >> printElementsOn: aStream [
 	aStream nextPut: $)
 ]
 
+{ #category : 'private' }
+OrderedDictionary >> rehash [
+	dictionary rehash
+]
+
 { #category : 'removing' }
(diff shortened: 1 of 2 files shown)


MarcusDenker:
this is a deeper change... it breaks in the boostrap. I close it for now

PR #15017, opened 2023-10-16
Check: can we compile Dictionary>>at: block without outer context?

MarcusDenker:
Check if we can compile dictionary>>#at without outer context for the error block


jecisc:
Seems like it passes :)

jecisc (approved):

PR #15022, opened 2023-10-16
Bootstrap: Use Hermes V3

jecisc:
This change is updating hermes to use v3.0.0 of hermes. This bring the v3 of the format that uses package name and trait name instead of categories and is filling the ClassBuilder with those in order to avoid any ambiguity between what is a package and what is a tag in the category.


jecisc:
Tests are now passing! \0/

PR #15023, opened 2023-10-16
Simplify postload of BaselineOfBasicTools

jecisc:
The baseline seems to expect the postload to be executed multiple times but this is not the case (I bootstraped an image that was set to fail if it was going two times there and it did not fail).

I propose to simplify it

### DIFF
--- a/src/BaselineOfBasicTools/BaselineOfBasicTools.class.st
+++ b/src/BaselineOfBasicTools/BaselineOfBasicTools.class.st
@@ -22,9 +22,6 @@ Installs:
 Class {
 	#name : 'BaselineOfBasicTools',
 	#superclass : 'BaselineOf',
-	#classVars : [
-		'Initialized'
-	],
 	#category : 'BaselineOfBasicTools',
 	#package : 'BaselineOfBasicTools'
 }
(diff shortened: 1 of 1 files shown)


guillep (changes requested):
No please, do not remove that :( Later if we re-introduce a diamond we will get bitten.

guillep:
It's like saying "Let's remove the `super initialize` because the superclass method is empty".
But if later somebody defines the initialize method in the superclass we get funny surprises and hours of debugging silly things.

jecisc:
Isn't Metacello suppose to execute only once the post loads even in case of diamonds? Or does it not handle this case?

guillep:
Thanks Cyril, and I'm sorry :(
I would also like that Metacello does it, yes, but it's not the case.

Maybe the fix should be done at that level?

jecisc:
If that's the case maybe we should update Metacello directly no? Because lot of code got extracted in their own baselines and this was not taken into account. For example:
- BaselineOfDrTest
- BaselineOfLibGit
- BaselineOfNewTools
- BaselineOfPharo
- BaselineOfReflectivity

They are not implementing this check

jecisc:
> Thanks Cyril, and I'm sorry :( I would also like that Metacello does it, yes, but it's not the case.
> 
> Maybe the fix should be done at that level?

Thanks for the clarification :)

PR #15026, opened 2023-10-17
Metacello: Simplify repository creation

jecisc:
Metacello has a lot of code present for the compatibility with multiple smalltalk versions but the Pharo version already diverged and is not compatible with other smalltalk anymore. 

Here is a first step at a simplification we can do to not add complexity with this compatibility layer. It moves the repository creation from the Platfrom to the repositories. This is only the first step because Tonel needs to be updated after this PR gets integrated in order to finish this cleaning and remove all the references to the platform.


Ducasse:
Yes simplify!

MarcusDenker (approved):

PR #15028, opened 2023-10-17
Do not recompile methods if slots did not change

jecisc:
When a class changes we recompile all the methods because the slots index might have changed. This change checks that slots changed before doing this because if no slot changed we can just keep the old methods and avid recompilations

### DIFF
--- a/src/OpalCompiler-Core/Behavior.extension.st
+++ b/src/OpalCompiler-Core/Behavior.extension.st
@@ -44,9 +44,13 @@ Behavior >> compileAll [
 { #category : '*OpalCompiler-Core' }
 Behavior >> compileAllFrom: oldClass [
 	"Compile all the methods in the receiver's method dictionary.
-	This validates sourceCode and variable references and forces
-	all methods to use the current bytecode set"
-	oldClass localSelectors do: [:sel | self recompile: sel from: oldClass]
+	This validates sourceCode and variable references and forces all methods to use the current bytecode set"
+
+	"If the slots changed, we need to recompile the methods. Ideally only the methods that changed, but for now we recompile everything. Else we can just use the old methods."
+
+	oldClass allSlots = self allSlots
+		ifTrue: [ oldClass localSelectors do: [ :selector | self addSelectorSilently: selector withMethod: (oldClass compiledMethodAt: selector) ] ]
+		ifFalse: [ oldClass localSelectors do: [ :selector | self recompile: selector from: oldClass ] ]
 ]
 
 { #category : '*OpalCompiler-Core' }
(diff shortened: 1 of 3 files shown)


jecisc:
One test fails but this is probably because the test should be updated. I'll do that later

MarcusDenker:
One thing to check: we need to still be able to force a recompile. E.g when changing the compiler. or when wanting to just clean up when experimenting.

MarcusDenker:
If I call #recompile on a class I want to recompile, allways

MarcusDenker:
I think this is as simple as 

- add this method as compileIfNeededFrom:
- call it from ShiftClassBuilder>>#compileMethods


this way, #recompile send to a class will continue to force a recompile

src/OpalCompiler-Core/Behavior.extension.st
```diff
@@ -44,9 +44,13 @@ Behavior >> compileAll [
...
-	This validates sourceCode and variable references and forces
-	all methods to use the current bytecode set"
-	oldClass localSelectors do: [:sel | self recompile: sel from: oldClass]
+	This validates sourceCode and variable references and forces all methods to use the current bytecode set"
+
+	"If the slots changed, we need to recompile the methods. Ideally only the methods that changed, but for now we recompile everything. Else we can just use the old methods."
+
+	oldClass allSlots = self allSlots
+		ifTrue: [ oldClass localSelectors do: [ :selector | self addSelectorSilently: selector withMethod: (oldClass compiledMethodAt: selector) ] ]
+		ifFalse: [ oldClass localSelectors do: [ :selector | self recompile: selector from: oldClass ] ]
```
dionisiydk, line 53: This is quite dangerous optimization. I agree that if new version of a class does not change the layout we don't need to recompile the method. But how we can safely check that the layout is same?
The approach here is to compare allSlots. Are you sure that it works correctly?
Few points:
- it leads to dependency on the #allSlots implementation. Imagine if tomorrow it will sort variables. Then reordering variables will be broken
- we have a huge hierarchy of various slot classes. Are they all implement equality? Also notice that with this change it will be a hidden requirement to properly implement the equality because otherwise any change of the slot state in the class definition will not be considered as the layout change.
- It's better to introduce a descriptive method #hasSameLayoutAs: (or similar) with this logic to compare allSlots.   

Another observation:
- #addSelectorSilently:withMethod:  modifies the added method. It sets the methodClass to point to new class. 
It means that there will be a time (until the full class change is completed) where the old version of class will contain methods in inconsistent state (pointing to new class). Can it be critical? Imagine that there are live instances where these methods are being executed by VM during the code change.

The current approach with full recompilation is 100% safe and much simpler. So is this optimization really make sense?

MarcusDenker:
I did the cleanups from this PR as part of https://github.com/pharo-project/pharo/pull/15109

jecisc:
I'll close it for now because I'll not have the time to work on this one for now

PR #15032, opened 2023-10-17
Improve disabling of Epicea in tests

jecisc:
In Pharo 12 I disabled Epicea in tests except for Epicea tests. But I am not satisfied with how I did it.

I did it by preventing to add events to the EpMonitor when the execution environment was the test environment. But that has the problem that Epicea is still listening to the events and generating its changes. This can be slow and in some edge cases creates bugs (for example when you test some things on another environment than the default environment). 

I propose to cut Epicea right from the start: when we receive the announcement.

### DIFF
--- a/src/Epicea/EpMonitor.class.st
+++ b/src/Epicea/EpMonitor.class.st
@@ -122,9 +122,6 @@ EpMonitor >> addEvent: anEvent newEntryDo: aBlock [
 EpMonitor >> addEvent: anEvent newEntryDo: aBlock triggerReference: triggerReference [
 
 	| newEntry |
-
-	"During the tests, we should only log with Epicea if the test case declare it wants logging."
-	(CurrentExecutionEnvironment value isTest and: [ CurrentExecutionEnvironment value testCase shouldLogWithEpicea not ]) ifTrue: [ ^ self ].
 	newEntry := log addEntryWith: anEvent tags: [ :tags |
 		            triggerReference isNull ifFalse: [ tags at: EpLog triggererReferenceKey put: triggerReference ].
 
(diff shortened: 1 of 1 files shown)


guillep (commented):
I like the idea, but I wonder if it is better to add this logic here in Epicea or somewhere else. I'm thinking about the test runner for example... But maybe here is ok, dunno :)

Now, I don't think that Epicea tests need a special case here.
If epicea tests need to listen for events, they should do it in a local way, without requiring the global Epicea logger.

jecisc:
I wanted to do it in Epicea because it was the central place for Epicea. I don't want to have the epicea noise if I launch my tests from the little circle from Calypso.

Or maybe we have a central place for all of that somewhere? Also, SUnit is loaded before Epicea and I wanted to contain code of Epicea in Epicea.

For the second part of the comment, I agree that we should not have a special case for epicea here! The problem is that the tests are currently using the global EpMonitor and it takes time (and Epicea knowledge) to update this :( So I went for the simplest solution for now

MarcusDenker:
Nice: this seems to speed up running the tests. 16m 9s on linux is the fastest I have seen for Pharo12 for linux

Ducasse:
I would love not to have package loaded logged. 
We can also ask @hernanmd in the future to revisit the filter and search in Epicea browser because multiple people cannot use it
on simple scenarios.

jecisc:
Package loading is on my todo list! I kept it for now to help me find some bugs in the package management but it should be good to remove it now

Ducasse:
We should document a bit somewhere how a test can declare its logging interest :)

PR #15044, opened 2023-10-18
unskip tests using ProvideAnswerNotification

### LINKED ISSUE #14978: ProvideAnswerNotification: need a way to resume with defaultAnswer
When testing code that uses methods like MorphicUIManager>>#request:initialAnswer:title:, we need a way to resume, but returning the defaultAnswer that was pre-filled in the dialog.

See 
- OCCodeReparatorTest>>#testdefineClass
- OCCodeReparatorTest>>#testdefineTrait

for tests that needs this (they are skipped for now)

MarcusDenker:
the method check for a symbol #default to be returned... so it has a mechanism for it (even though maybe adding the defaultAnswer as state to ProvideAnswerNotification might be an interesting future improvement...)

But this fixes 14978 by using the existing mechanism

### DIFF
--- a/src/OpalCompiler-Tests/OCCodeReparatorTest.class.st
+++ b/src/OpalCompiler-Tests/OCCodeReparatorTest.class.st
@@ -291,15 +291,13 @@ OCCodeReparatorTest >> testUndeclaredVariable [
 
 { #category : 'tests' }
 OCCodeReparatorTest >> testdefineClass [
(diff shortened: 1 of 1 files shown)


jecisc:
The unskipped tests are failing

MarcusDenker:
so it seems this mechanism is only implemented if not headless...

PR #15048, opened 2023-10-18
Turn off Error hander on  #computeFullBounds  for a while to find a bug

MarcusDenker:
we see often a SubscriptOutOfBounds (see #12502) on the CI.

The error handler of #computeFullBounds makes this very gard to track down... this PR turns it off.

I propose to run a bit without this and later make this a preference to turn on in deployment but let it off on the CI?


Ducasse:
Good idea!

PR #15053, opened 2023-10-19
Remove usage of old class definition syntax

### LINKED ISSUE #14950: Check what to do with non fluid class definition
In Pharo 12 we enabled Fluid class definition by default. This leaves a few questions

- Do we ship Pharo 12 with both definitions (Fluid by default) and keep the checkbox in Calypso to go back to the old one?
- Do we deprecated all the old class creation messages in Pharo 12? (There is a PR ready for that)
- Do we remove the old class parser in Pharo 12?
- If no, do we remove the old class parser tests in Pharo 12?

jecisc:
This change removes the usage of the old class definition syntax in the Pharo image, but it keeps a compatibility!

Changes:
- If we save an old class definition syntax we are now using the special parser of DoItChunk to transform it into a ShiftClassBuilder instead of executing it and we raise a warning to the user specifying that this should not be used anymore
- It removes the setting #showFluidClassDefinition
- Remove the checbox in Calypso to go back to old class definition
- Removes the old class parser
- Remove some code that managed the specificities between the two syntaxes 

Depends on #15052

Fixes #14950


jecisc:
Two related failing tests detected

PR #15056, opened 2023-10-19
outer context is not sender context

MarcusDenker:
Context outerContext falls back on #sender. This is not a good idea as the two are really different concepts. A Context of a closure without an outerContext (Clean Blocks, methods) should return nil


Ducasse:
I do not get why all tests pass on windows and one is breaking one mac and linux

```
unix-64 / Tests-unix-64 / testExampleAccessOuterFromCleanBlock – Unix64.OpalCompiler.Tests.Misc.MethodMapTest
```

MarcusDenker:
Win sometimes does not run at all, yet shows green

PR #15063, opened 2023-10-20
remove class TextEditor

MarcusDenker:
this is part of part of  #2717


Ducasse:
So nice that we can remove it :)

MarcusDenker:
aargh... removed a class too much. I will redo

PR #15079, opened 2023-10-23
Deprecate MetacelloPlatfrom>>#authorName:

jecisc:
This mehod was recently removed but it is called by smalltalkCI. I propose to add it back as deprecated.


jordanmontt (approved):
Thanks!

PR #15084, opened 2023-10-23
Calypso: Remove duplicated registration of SycRemoveClassCommand

jecisc:
SycRemoveClassCommand has the same methods in the class and as extension of Calypso-SystemTools-FullBrowser. 

I propose to remove this clash


MarcusDenker (approved):

Ducasse:
I think that we should remove the class and only keep the version 2. 
When I tried the last time I think that iceberg was confused because I remove a class and renamed the version to use the one of version1.

PR #15085, opened 2023-10-23
Rewrite `== nil ifTrue:` to `ifNil:`

jecisc:
Simple changes

### DIFF
--- a/src/Calypso-SystemTools-QueryBrowser/ClyTextEditor.extension.st
+++ b/src/Calypso-SystemTools-QueryBrowser/ClyTextEditor.extension.st
@@ -67,9 +67,10 @@ ClyTextEditor >> referencesTo: aVariableName [
 
 { #category : '*Calypso-SystemTools-QueryBrowser' }
 ClyTextEditor >> referencesToIt [
+
 	| selector |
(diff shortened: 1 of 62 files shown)


Ducasse:
Cyril 
there are not simple and in particular we should not touch kernel methods. 
Why because == is not preempted while ifNil is a preemption point. 
With changes like that we got race conditions on servers after 2 weeks in the past. 
So we decided to pay attention to not touch kernel methods.

MarcusDenker:
"== nil ifTrue:" is save to replace, as it is compiled the same as "ifNil:"

We have to be more careful to introduce new messages. e.g. replacing a "== nil" with isNil would introduce an interrupt check were none was before (e.g. GC can then happen). 

But this PR is save.

Ducasse:
I saw that you did not modify the processList code.

jecisc:
I tried in the past a PR that was replacing both `== nil ifTrue:` and `== nil` and it failed. In between I learned that `== nil ifTrue:` is safe to update as said Marcus so it's the reason I did a PR with only those :)

MarcusDenker (approved):

PR #15087, opened 2023-10-23
15074-Pragmas-lost-when-updating-methods

### LINKED ISSUE #15074: Pragmas lost when updating methods
**Bug description**
Pragma pragmaCache has an inconsistent view of the methods containing which pragmas.

**To Reproduce**
1. Create a method (either class or instance method) with eg: the `<foo: 'yes'>` pragma.
2. Inspect `Pragma allNamed: #foo:.`

**Expected behavior**
an Array with the Pragma instance.

**Steps**
Then when I run:
1. `(MyClass methodNamed: #myMethod) cachePragmas.`
2. `Pragma allNamed: #foo:` I get the pragma
3. but if I update the method code changing the pragma to `<foo: 'no'>`,
4. then when I run Pragma allNamed: #foo:` the Pragma is no longer there.

**Screenshots**
image

**Version information:**
 - OS: [e.g. iOS]: macOS
 - Version: [e.g. 1.6.1]: Sonoma 14.0 running on Apple M1 Pro
 - Pharo Version [e.g 9]: 12

**Expected development cost**
As you know nothing happens magically, so do you have an idea of the development effort required? 
* No idea, I'm not familiar with Pharo internals, and I'm starting in my path towards learning Smalltalk.

In addition, are you willing to help because after all Pharo is yours too?
* Sure, I'll try to help.

**Additional context**

* Could it be that I'm using pragmas wrong and they are not supposed to be used that way?

MarcusDenker:
- add #testRecompile which fails without the fix
- add #copyWithoutNoShallowCopy: that does not copy
- use it when removing properties only

fixes #15074

### DIFF
--- a/src/Kernel-Tests/PragmaTest.class.st
+++ b/src/Kernel-Tests/PragmaTest.class.st
@@ -97,6 +97,15 @@ PragmaTest >> testHash [
 ]
 
 { #category : 'tests - cache' }
+PragmaTest >> testRecompile [
+	self assert: (Pragma allNamed: #testPragmaArg1:arg2:arg3:) notEmpty.
+	self class compile: 'methodWithPragma
+	<testPragmaArg1: #toto arg2: 2 arg3: true>' classified: 'helper'.
+	Smalltalk garbageCollect.
+	self assert: (Pragma allNamed: #testPragmaArg1:arg2:arg3:) notEmpty.
+]
+
+{ #category : 'tests - cache' }
 PragmaTest >> testall [
 	self assert: Pragma all first class equals: Pragma
 ]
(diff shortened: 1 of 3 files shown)


MarcusDenker:
the reason why we see that now is that the compiler itself uses a pragma (#source) to return a method with source, this property is then removed when you install the method, triggering the bug for every compile+install.

Of course, this was there before and could have been triggered when removing a propery on a method with a pragma.

MarcusDenker:
This shows again that complexity will kill us eventually: the idea that AdditionalMethodState contains both associations and Pragmas is just horribly wrong. And of cource the reason for AdditionalMethodState in the first place is the horrible idea of making CompiledMethod very very special (having their own format).

jecisc (approved):

Issue #10027, opened 2021-09-23
[Feature Request] Pragma should gave a #sourceNode method


MarcusDenker:
All Pragmas (<these: #things>) result in instances of Pragma that you can get from the method (e.g. "(Behavior>>#basicNew) pragmas"

What is not easy is to get from those pragma instances to the AST that created them. "(Behavior>>#basicNew) pragmas first sourceNode" does not exist.

Take care: 

- there can be multiple pragmas that are the same
- RBPragmaNode does not store the pragma instance

**Is your feature request related to a problem? Please describe.**
I wanted to highlight the pragma in the Pragma inspector pane, no way to get the text offset now.

**What needs to be done**
RPParseNode should store the Pragma instance, not the information to create one (see senders of #asPragma how they are now created).

After that, we can easily check which RBPragmaNode has the pragma, and thus implement #sourceNode very easily.

MarcusDenker:
I think one idea can be to treat RBPragmaNode a bit the same as a variable: the semantic analysis could annotate the RBPragmaNode with the Pragma instance. 

Looking at #asPragma senders, this would reduce the places where we create pragmas to be just that one place in the semantic analysis and would simplify the code related to reading the compilerOptions Pragma, which we already have to do in Semantic analysis.

see OCASTSemanticAnalyzer>>#visitPragmaNode:

MarcusDenker:
This is sadly right now not possible. Due to the AdditionalMethodState>>#copyWith: being called in the compiler and it copying Pragmas, too, when the compiler finally sets the compiledMethod in the pragma of the method, that pragma is nor the instance that the RBProgramNode references.

We thus need to first do the cleanup we wanted to do in Pharo9: Pragmas should be stored in a property called #pragmas.
This would avoid relying on AdditionalMethodState to store them directly, thus copyWith:  would not copy them.

astares:
Yes - having them as properties would be good.

PR #15090, opened 2023-10-24
AdditionalMethodState: do not copy the Associations/Pragmas when growing or shrinking

MarcusDenker:
This PR makes sure that we never copy the associations and properties when removing an entry from the AdditionalMethodState array.

This is save for the associations as we do not *copy* the method, we just shrink the private AdditionalMethodSate object.

### DIFF
--- a/src/Kernel-Tests/PragmaTest.class.st
+++ b/src/Kernel-Tests/PragmaTest.class.st
@@ -97,6 +97,14 @@ PragmaTest >> testHash [
 ]
 
 { #category : 'tests - cache' }
+PragmaTest >> testPragmaAddingProperty [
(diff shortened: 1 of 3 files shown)


MarcusDenker:
This is a better fix for https://github.com/pharo-project/pharo/issues/15074 and fixes the problem for the case of adding a method propery, too.

Keeping the identity of Pragma now allows us to implement #10027 without having to wait for the simplification to store Pragmas as a property.

MarcusDenker:
CI failure not related:

- SpMorphicBoxAdapter>>#remove: can crash https://github.com/pharo-spec/Spec/issues/1466

jecisc (approved):
Nice cleaning

PR #15098, opened 2023-10-24
Monticello: Unify more method addition

jecisc:
This is a new step at having method addition code closer to the code we have in the system in order to get rid of this duplicated method compilation and installation.


jecisc:
This is included in a better PR. Closing

PR #15100, opened 2023-10-24
Undeclared: keep weak reference to all methods using the Undeclared Variable

MarcusDenker:
When we compile an undeclared, record the compiled methods in the UndeclaredVariabke instance.

This will allow us to fix these methods without having to search all methods.

### DIFF
--- a/src/Kernel/UndeclaredVariable.class.st
+++ b/src/Kernel/UndeclaredVariable.class.st
@@ -8,6 +8,9 @@ In future I can profice logging and user warning when evaluated code accesses un
 Class {
 	#name : 'UndeclaredVariable',
 	#superclass : 'LiteralVariable',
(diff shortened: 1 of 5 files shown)


jecisc:
This prevent the usage of #primitiveChangeClassTo: since the layout is now different.

But since we have properties on this kind of variables maybe we can store the using methods in the properties and clean the properties when we declare it?

MarcusDenker:
Yes, I was thinking the same. If we later switch to recompiling instead of changeClass for the fixing of Undeclareds, we can switch it back to an ivar

jecisc (approved):
And so with that we can remove the cleanings of undeclared? (recompilation of methods and cleanups of the Undeclared collection)
Nice :)

PR #15102, opened 2023-10-24
Introduce and use CategoryConverter

jecisc:
With the removal of the categories we need to handle the code that was using categories in the past. I already handled it in multiple places but this was duplicating the behavior and I found even more places.

In the end I'm proposing to introduce a CetagoryConverter class to centralize this behavior and make all other places rely on it. I also updated MCStReader and MCTraitParser to handle better the categories

### DIFF
--- a/src/CodeImport/OldClassDefinitionBuilder.class.st
+++ b/src/CodeImport/OldClassDefinitionBuilder.class.st
@@ -166,7 +166,7 @@ OldClassDefinitionBuilder >> build [
 	(#( CompiledBlock CompiledCode CompiledMethod ) includes: subclassName) ifTrue: [ layoutClass := CompiledMethodLayout ].
 
 	^ self class classInstaller make: [ :aBuilder |
-		  | superclass realPackageName |
+		  | superclass converter |
 		  superclass := self class environment at: superclassName asSymbol.
 
 		  aBuilder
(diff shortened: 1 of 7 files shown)


MarcusDenker (approved):

Rinzwind:
@jecisc Shouldn’t #tagNames: on MCOrganizationDefinition ensure the tag names are symbols as done in commit fa5708752ab7b2d3 for #tagName: on MCClassDefinition? (*edit:* I misread that commit, the send of #asSymbol was actually added before in commit e8bf7f714781a858)

I tried converting a repository from Pharo 11 to Pharo 12 using the following snippet (based on the one given in pull request #14743). In the converted ‘categories.st’ files the tags are written as arrays of strings, but in subsequent commits they are written as arrays of symbols.

```smalltalk
repository := IceRepository registry detect: [ :registeredRepository | registeredRepository name = repositoryName ].
repository workingCopy packages do: [ :package |
	repository writerClass forInternalStoreFileOut: package latestVersion mcVersion on: repository ]
```

In #processCategory: on CategoryConverter, ‘tagName’ gets set to a string.

Rinzwind:
I opened a pull request to make #tagNames: on MCOrganizationDefinition ensure the tag names are symbols: pull request #18588.

PR #15107, opened 2023-10-25
Make finalization registry a normal strong double linked list

### LINKED ISSUE #15091: Slow finalization for pinned objects
**Bug description**
Starting with the Pharo 12 builds having this PR integrated [1] (I tested builds 720 as pre-integration and 730 as post-integration), we have troubles with the slow GC of our application objects. In the end, it is garbage collected, but it requires 65 GC cycles to do so.

[1] https://github.com/pharo-project/pharo/pull/14068

This is the typical leak structure:

Pharo_iDeuS8gZgv

I have discovered it is related to objects pinned in memory which ExternalAdresses usually are.

**To Reproduce**
1. Create your own `ValueHolder` subclass

```smalltalk
ValueHolder << #MyValueHolder
	slots: {};
	tag: 'Base';
	package: 'NewValueHolder-Core'
```

2. open Transcript, then open Playground and fill it with this code:
```smalltalk
"part 1"
| col oldO size |
col := OrderedCollection new.
oldO := nil.
4 timesRepeat: [ 
	| o |
	o := MyValueHolder new contents: oldO.
	oldO := o.
	o pinInMemory.
	o finalizationRegistry add: o.
	col add: o.
	 ].
Smalltalk garbageCollect.

"part 2"
5 timesRepeat: [ 
	Smalltalk garbageCollect.
	size := MyValueHolder allInstances size.
	size traceCr.
]
```
3. Do-It part 1, then Do-It part 2 (it cannot be executed as one)
4. In Transcript, you will see: 
```
4
3
2
1
0
```
5. Now, try to comment out the line `o pinInMemory` and run part 1 and part 2 again. The value holders are GCed immediately

You may try to run it (with pinning) in the older builds before 720; there the GC is immediate, too.

**Expected behavior**
Objects are not GCed one by one.

guillep:
Fix for https://github.com/pharo-project/pharo/issues/15091

It is what I sketched in #15091.

## Background

The GC run a loop until it converges:
1. scan from the roots finding objects, stop scanning at ephemerons
3. fire the ephemerons we found in this round
4. scan the fire ephemerons (this could lead to discovering new ephemerons)
5. go to 1 if not finished

Important to notice: when an ephemeron is fired it becomes a normal object with strong references, thus holding strongly its key (and its graph). This could prevent the firing of ephemerons in subsequent iterations.

## The problem

Originally (before https://github.com/pharo-project/pharo/pull/14068), step 1 discovered all ephemerons in the registry because it holds strong pointers to all of them
Later (after https://github.com/pharo-project/pharo/pull/14068), when we introduced the regression, step 1 discovered *only* the ephemerons from the head of the registry (the first strong pointer), because they were organized in a linked list of ephemerons.

Thus, both of them found the ephemerons but not in the same iteration of the loop.
Since ephemerons are discovered and fired at different times, it could be that firing ephemeron A at iteration 1 will hold a key preventing the firing of ephemeron B at iteration 2.

In ascii art it looks like

Before (e are ephemerons, k are keys, ==> are strong references,  --> are ephemeral references)

```
Registry
   |====>e1---->k1
   |====>e2---->k1
```

After  (e are ephemerons, k are keys, ==> are strong references,  --> are ephemeral references)
```
Registry
   |====>e1---->k1
         |====>e2---->k1
```

## Solution sketch

So the simple idea that could be implementable now is to change the structure using a normal linked list.
This way the ephemeron could have a backpointer to its link node for O(1) removal, and still, we have a strong path to all ephemerons in the registry to find them in a single pass.
(l are links, e are ephemerons, k are keys, ==> are strong references,  --> are ephemeral references)

```
Registry
   |====> l1 <====> e1---->k1
           |====> l2 <====> e2---->k1
```


guillep:
I've made my check on the generated image, it works as expected performance wise when stressing the infrastructure too. It's good to integrate IMHO.

PR #15113, opened 2023-10-25
Merge OCPragmaTest into  MethodPragmaTest

MarcusDenker:
Future work: move all tests that are about Pragma not involving the compiler to PragmaTest (part of the kernel tests)


jecisc (approved):

src/OpalCompiler-Tests/MethodPragmaTest.class.st
```diff
@@ -240,9 +330,8 @@ MethodPragmaTest >> testMethodSelector [
 
 { #category : 'tests - compiled' }
 MethodPragmaTest >> testNoPragma [
-	| method |
-	method := self compile: '' selector: #foo.
-	self assert: method pragmas equals: #()
+
+	OpalCompiler new parse: self methodNoPragma
```
jecisc: This test has no assertion now. Is that normal?

MarcusDenker: removed

PR #15114, opened 2023-10-25
Monticello: Improve MCZ management

jecisc:
This improves two small things:
- MCDataStream had #readWordLike as backward compatible method but it is still used (for example the bootstrap code sometimes fallback to another importer than MCZ importer because mcz got exported with this selector defined or reading)
- When we cannot read a MCZ, log the failure and give the short stack so that we can find in CIs why we have troubles


jecisc:
Requires #15115

MarcusDenker (approved):

PR #15122, opened 2023-10-26
Do not use categories in ShiftClassInstaller

jecisc:
This one is a big change!

## The change

This change uses packages and tags in the class installer instead of the categories! This means that all importers used by the bootstrap (espell, hermes, Chunk importer, St importer, MCZ importer, FileTree importer, Tonel importer)

It also cleans things we do not need anymore such as exporting the list of kernel packages in a text file.

With that I started to clean what we can drop from the category management but more is to do later!

## The drawback 

I had to do a really ugly hack because of a bug I do not understand tho.. This is in RPackageOrganizer>>#basicBootstrapInitialize. 

The problem: We are storing classes in a variable of RPackageOrganizer and in a variable of RPackageTag. I do not know why, but in some cases the classes present in those variables are equals to the ones in the system dictionary, but are not identical (pointer identity). This leads to a lot of bugs afterward.

As a workaround I scan those variables and replace them by the ones in the system dictionary. But it is really weird that we have this behavior and I do not know why we are having that. 

## Next steps

Multiple next steps are to be done here:
- Deprecated ClassRecategorized
- Deprecate #category: 
- Do multiple cleanings related to categories 
- Remove the usage of announcements that we still have in RPackageOrganizer (We should be able to do that now)
- Update the variable of Class from category to packageTag

Marking as priority medium because it is blocking for a lot of other changes


jecisc:
I think Pablo is in holidays and I hope this can be reviewed before he comes back because I have to do a lot of follow up changes :)

MarcusDenker:
I will review

MarcusDenker (approved):
I have no idea why the classes are not identical... maybe now they get cleaned up later when we later reload all during the full image build? But no idea.

I thin it should be save thought to merge, we can improve that later

jecisc:
I plan to investigate later because I don’t like this hack

PR #15129, opened 2023-10-26
Rewrite `notNil ifTrue:` into `ifNotNil:` in packages starting by M

jecisc:
Simple improvement


MarcusDenker:
this even speeds up things (as it removes a send). It should be save wrt to process switching, as it removes an intererrupt point, it does not add one.

MarcusDenker (approved):

PR #15131, opened 2023-10-26
Create a release test to control the number of senders of #notEmpty

### LINKED ISSUE #15125: Create a release test to control the number of senders of #notEmpty
Currently we have 264 senders of #notEmpty which we should change in favour of using #isNotEmpty (see related issue #15120)
This is distributed to standard Pharo repo but also Spec, Iceberg repo and others

```Smalltalk
SystemNavigation default allSendersOf: #notEmpty
```

Lets create a release test to control the number of senders of #notEmpty so we do not increase the number but can bring the number
down step by step and then finally be able to solve #15120

image

astares:
Fix #15125

### DIFF
--- a/src/ReleaseTests/ProperlyUsedMessagesTest.class.st
+++ b/src/ReleaseTests/ProperlyUsedMessagesTest.class.st
@@ -1,0 +1,17 @@
+Class {
+	#name : 'ProperlyUsedMessagesTest',
+	#superclass : 'TestCase',
+	#category : 'ReleaseTests-CleanCode',
+	#package : 'ReleaseTests',
+	#tag : 'CleanCode'
+}
+
+{ #category : 'tests' }
+ProperlyUsedMessagesTest >> testSendersOfNotEmpty [
+	"Instead of using #notEmpty the method #isNotEmpty should be sent. Check here that 
+	 we do not grow with the number of senders."
+	
+	| violations |
+	violations := SystemNavigation default allSendersOf: #notEmpty.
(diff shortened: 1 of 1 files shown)


MarcusDenker:
I would wait with this.

- there is no deprecation yet
- there is no rule

So this will just lead to people getting angry because that test fails when submitting PRs without any chance to avoid it.

Another thing is that these tests that check numbers are bad as the do not tell us *where*. This leads to this test e.g. failing when running all tests for Spec or Newtools, but no way to find why.

jecisc (changes requested):
Fixing a hardcoded number of violation is bothersome because we don't often update it then we are fixing some places and other introduce the problem again.

I prefer to have a list of violation like I did in ReleaseTest>>#testNoSentButNotImplementedSelectors

astares:
This is a question of discipline. 
Usually I updated after I cleaned. And we can bring it down to zero.

Dont know about your working style ;)

jecisc:
I’ve seen multiple times regressions on the implemented but not sent rule because not every once knows it is there. And we had multiple failures on external projects running Pharo release tests such as New Tools or Spec or Microdown. 

Also when it fails your version give no info on the origine of the failure while giving the list of current violations makes it explicit what is the new violation in the CI logs.

This makes it much easier to maintains the projects

astares:
> * there is no deprecation yet

Yes - see https://github.com/pharo-project/pharo/pull/15137

astares:
Close for now

PR #15132, opened 2023-10-26
RPackage: Clean protocols management

jecisc:
First I removed the usage of announcements to update RPackage.

I simplified some code in Calypso.

And I removed a usage of #ensurePackageOfExtensionProtocol: to use #packageForProtocol: because the package should always be created before and I want the code to fail if it is not the case so that we find the problem


jecisc:
Seems like the problem is caused by: #15134

jecisc:
All tests seems to have pass on this one now! I added a commit to fix a random failure but it should be ready for review

jecisc:
It looks like this PR does not pass anymore since other PRs got merged :(

MarcusDenker (approved):

PR #15133, opened 2023-10-26
Do not announce method recategorization when removing a selector from a protocol

jecisc:
Removal of selector from a protocol is used in 3 cases:
- We have protocols including non existing selectors and we are cleaning
- We are removing a method
- We are rebuilding a method dictionary

IMO we should not announce a method recategorized in those cases because either it's cleaning or removal anc we already announce method removal


jecisc:
Tests failures are unrelated

MarcusDenker (approved):

PR #15134, opened 2023-10-26
MethodAddition: Announce protocol changes

jecisc:
New step at making MethodAddition behave like classic compilation: announce protocol changes as we do usually on installation


jecisc:
This causes a problem loading Iceberg. I need to check if this is intended or not

jecisc:
We cannot just announce the change of protocol because Monticello is doing things in two steps:
- Create the method and install it without the protocol
- Install the method in the right protocol

This means that extension methods get first installed in the class in #as yet unclassified and then they get moved to the extension protocol and extension package. But if we announce the protocol changes this means that RPackage thinks it is a method repackaging instead of an installation and it will mark the package of the class in which the method is installed as dirty. And this will create conflict if we reload the project

jecisc:
I'm closing in favor of another improvement

PR #15136, opened 2023-10-26
Edit package of PositionableStream>>#nextWordsInto:

jecisc:
Move it to Monticello because Monticello is the first thing loaded relying on it.

Maybe this could be improved further later but this could improve the bootstrap

I found this in the bootstrap logs:

```
An error happened while reading MCZ. We will fallback to another format. Short error stack: ReadStream(Object)>>doesNotUnderstand: #nextWordsInto:
WideString class>>newFromStream:
MCDataStream>>readWordLike
MCDataStream>>next
MCClassDefinition(Object)>>readDataFrom:size:
MCDataStream>>readInstance
MCDataStream>>next
MCDataStream>>readArray
MCDataStream>>next
MCSnapshot(Object)>>readDataFrom:size:
```

### DIFF
--- a/src/Monticello/PositionableStream.extension.st
+++ b/src/Monticello/PositionableStream.extension.st
@@ -1,6 +1,6 @@
 Extension { #name : 'PositionableStream' }
 
-{ #category : '*Graphics-Primitives' }
+{ #category : '*Monticello' }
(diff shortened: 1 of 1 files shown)


jecisc:
I'll close that for now because I'll not have the time to fix it

PR #15139, opened 2023-10-27
Split finalization registry from WeakArray and sort dependencies

guillep:
Split the code that manages the finalization process into its own class.
Put it under the System-Finalization package.
Fix all dependencies.
Remove WeakRegistry deprecated in P11


MarcusDenker:
Failing test:

```
the following classes have uncategorized methods: an OrderedCollection(FinalizationProcess class)
```

PR #15146, opened 2023-10-27
Fix moving classes in tags

jecisc:
During a previous refactoring it seems that a super call was left in code and this super is not needed


jordanmontt (approved):
nice catch

PR #15149, opened 2023-10-27
Inlining `select:thenCollect:` and `reject:thenCollect:` to not do two iterations on the same collection

### LINKED ISSUE #15142: Collection>>#select:thenCollect: should do a one pass iteration
This kind of methods are there for optimization to do only one pass instead of two times but this one does not. 

We should update the implementation to do only one pass

jordanmontt:
Fixes #15142

### DIFF
--- a/src/Collections-Abstract/Collection.class.st
+++ b/src/Collections-Abstract/Collection.class.st
@@ -1659,7 +1659,11 @@ Collection >> select: selectBlock thenCollect: collectBlock [
 	"(#(1 2 3 4 5) select: #even thenCollect: #negated) >>> #(-2 -4)"
 	"('Hello, World!' select: #isLetter thenCollect: #uppercase) >>> 'HELLOWORLD'"
 
-	^ (self select: selectBlock) collect: collectBlock
+	| selectedItems |
+	selectedItems := self copyEmpty.
+	self do: [ :e |
+		(selectBlock value: e) ifTrue: [ selectedItems add: (collectBlock value: e) ] ].
+	^ selectedItems
 ]
 
 { #category : 'enumerating' }
--- a/src/Collections-Abstract/Collection.class.st
+++ b/src/Collections-Abstract/Collection.class.st
@@ -1546,11 +1546,11 @@ Collection >> reject: rejectBlock thenCollect: collectBlock [
 
 	"(#(1 2 3 4 5) reject: #even thenCollect: [:x|x*10]) >>> #(10 30 50)"
 
-	| rejectedItems |
-	rejectedItems := self copyEmpty.
+	| newCollection |
+	newCollection := self copyEmpty.
 	self do: [ :e | (rejectBlock value: e) ifFalse: [ 
-			rejectedItems add: (collectBlock value: e) ] ].
-	^ rejectedItems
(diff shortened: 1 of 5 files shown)


Ducasse:
Approved

jecisc:
There was a failing test on both unix and osx. I'm restarting to see if it is related or not (I tried locally and the test is green)

jecisc (approved):

jordanmontt:
Failing tests seem unrelated

jecisc:
Most failing test are unrelated but it is twice now that I see failing tests on CompletionContextTest only in this PR.. Could be random but it could be related also :/

jordanmontt:
Indeed, the test `CompletionContextTest>>#testReceiverArgument` fails because the changes introduced in this PR. 
The problem is that `WeakIdentitySet>>#select:thenCollect: (Collection)` is not returning the expected results.

I'm trying to understand why… I added tests to `WeakSet` and `WeakIdentitySet` but I don't seem to reproduce the issue, the tests are passing as expected.

src/Collections-Abstract/Collection.class.st
```diff
@@ -1542,11 +1542,15 @@ Collection >> reject: aBlock [
...
+	"Optimized implementation"
 
 	"(#(1 2 3 4 5) reject: #even thenCollect: [:x|x*10]) >>> #(10 30 50)"
 
-	^ (self reject: rejectBlock) collect: collectBlock
+	| rejectedItems |
+	rejectedItems := self copyEmpty.
+	self do: [ :e | (rejectBlock value: e) ifFalse: [ 
+			rejectedItems add: (collectBlock value: e) ] ].
+	^ rejectedItems
```
Gabriel-Darbord: A small note about the variable name: the method is called "reject", but it still returns items that were **not** rejected, not the items that were rejected as the name would suggest.
So I think it would make more sense to name the variable `selectedItems` as in `select:thenCollect:`.
```suggestion
	| selectedItems |
	selectedItems := self copyEmpty.
	self do: [ :e | (rejectBlock value: e) ifFalse: [ 
			selectedItems add: (collectBlock value: e) ] ].
	^ selectedItems
```

src/Collections-Unordered/Set.class.st
```diff
@@ -144,6 +144,12 @@ Set >> collect: aBlock [
 	^ newSet
 ]
 
+{ #category : 'copying' }
+Set >> copyEmpty [
+
+	^ self species new: self size
```
Gabriel-Darbord: Setting the size of the copy to the size of the original goes against the way `copyEmpty` is implemented in all other classes, where it is often explicitly set to 0, or implicitly set to the default empty size.

jordanmontt: Indeed, thanks for the comments

jecisc:
Failure should be fixed by #15175

jordanmontt:
I see that `testReceiverArgument` is passing now :)
The other failing tests are relating to the DateParser. Those are related to https://github.com/pharo-project/pharo/issues/15177

I think it is safe to merge now!

jecisc (approved):

PR #15150, opened 2023-10-27
Replacing the use of species new with copyEmpty.

jordanmontt:
Replaces `self species new` with `self copyEmpty` because `copyEmpty` already deals with `species new`. And, it is better to use `copyEmpty` because there are some cases in which you can loose behavior with `self species new`. For example look `SortedCollection>>#copyEmpty`.

### DIFF
--- a/src/Collections-Abstract/Collection.class.st
+++ b/src/Collections-Abstract/Collection.class.st
@@ -609,7 +609,7 @@ Collection >> collect: aBlock [
 	"(#() collect: [:x | x+1]) >>> #()"
 
 	| newCollection |
-	newCollection := self species new.
+	newCollection := self copyEmpty.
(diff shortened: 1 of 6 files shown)


MarcusDenker (approved):

Ducasse:
While I like the idea, I think that copyEmpty is a bad name for 

```
Collection >> copyEmpty

	"Return a new empty collection based on the species"

	"{1@2} copyEmpty >>> #()"
	"'hello' copyEmpty >>> ''"
	"(1 to: 10) copyEmpty >>> #()"

	^ self species new
```

I think that we should rename it (probably in a following PR) as `newEmpty` because copyEmpty gives the wrong impression that it is doing something based on the receiver but it does not.

PR #15152, opened 2023-10-27
Deprecate String>>asDate

ISayarLu:

```smalltalk
```

```smalltalk
```

We deprecate methods: 
- String>>asDate,
- Date>>fromString, and 
- Date>>readFrom.

### DIFF
--- a/src/Collections-Strings-Tests/StringTest.class.st
+++ b/src/Collections-Strings-Tests/StringTest.class.st
@@ -436,17 +436,6 @@ StringTest >> testAsCamelCase [
 ]
 
 { #category : 'tests' }
-StringTest >> testAsDate [
-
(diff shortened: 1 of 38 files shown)


Ducasse:
Good idea!.

MarcusDenker:
This PR as merged even though it has failing tests.

noha:
What is the policy here. I would expect reverting is the obvious choice, no?

guillep:
could we please just revert it instead of making an argument here ? I'm on vacations and not on a computer, otherwise I'd do it myself

guillep:
PR Reverted in https://github.com/pharo-project/pharo/pull/15227 and reopened in https://github.com/pharo-project/pharo/pull/15228.

I'll rework it during the week when I come back from vacations.

PR #15155, opened 2023-10-27
Calypso: Introduce a project view based on baselines in the image

jecisc:
This PR adds a concept of Project to Calypso. 

Until we have Packbot ready, I consider than a project is defined by a baseline and the packages it contains are the packages declared in the baseline. 

Here is a preview of the project view:

image
image

I added a Project abstraction so that it is easier to implement the same kind of view once we have a new browser based on Spec. And also to make it easy to migrate when Packbot is ready.

I also added a setting to make this view the default view. For now the default value is false, but maybe people could be interested to make this true?

Done during the Pharo sprint in collaboration with @koendehondt  

Next step: We could add menus and maybe comments about the project using the comment of the baseline


jecisc:
Oh, I also improved the title to say if we are in a project, package, class or method

Ducasse:
nice

PR #15162, opened 2023-10-28
MethodAddition: Announce protocol changes  and unify more method installation

jecisc:
New step at making MethodAddition behave like classic compilation: announce protocol changes as we do usually on installation + make installation closer to what we have in the kernel/Opal.

This is a new version of a change I tried recently

This includes the changes of #15162 which had this comment:

MethodAddition has its duplicated way of installing methods in the system. I'm trying to make it as close as possible as the system way on installing methods to avoid inconsistant behaviors.

Here is a next step: Classify the method at the same time as we install it.

This will help to reduce the difference in behavior because the current implementation does not announce protocol changes. This is because it creates bugs and I found the root of those bugs! The root is that since we install and classify in two steps, the method get first installed in 'as yet unclassified' and then goes to its right protocol. But in case of extension methods, it would consider that the method changed from non extension to extension method and mark the package of the class as dirty because we removed a method from it. So, to avoid that, Monticello prevent RPackageOrganizer from doing the repackaging by silencing the announcement. But I want to get rid of those announcement listening and also the behavior is bad.

This was really hard to catch


jecisc:
Closing in favor of #15172

PR #15163, opened 2023-10-28
Monticello: MethodAddition should delegate method installation to Opal

jecisc:
The default way to install a method in the system is to use Opal but MethodAddition in Monticello does not use it and a different way of installing methods is implemented.

This change is trying to remove all this duplicated way to delegate everything to Opal.

Some notes:
- The old version does not announce protocol changes, the new version does
- The old version duplicate the whole announcement system, here we just collect them and send them in a batch at the end (this is done so that every methods gets added to the system before we announce they were added to not break mechanisms such as world menu reacting to pragmas)
- The old version compiled the method while creating the MethodAddition, now we compile it when we want to install it

To make this work I had to add a safe guard to Opal Compiler plugins management. If you load a plugin that adds itself in his #compiler method, then you could not load its code except if you instantiated the compiler before loading any method which makes the code more complex. What I did now is that I have a safe guard on plugin application. For now it just logs that a problem happened but we should probably signal a warning later. I just don't know how to raise a warning that should be ignored in non interactive mode.


jecisc:
Blocked by: https://github.com/pharo-project/pharo/issues/15202

jecisc:
So, this is now possible with the recompilation fix on the way in #15263 

Since I got the tests green I decided to do a little more cleanup. I updated the #load method of MCMethodDefinition to try to make it work in any case. I updated the MCPackageLoader to use the #load method and not do some magic. And I removed the now useless MethodAddtion whose code became useless now that MCMethodDefinition should know how to install without problem.

Let's see if this also work, else I'll split the changes

jecisc:
I'll redo one that is cleaner

PR #15166, opened 2023-10-29
Reduce usage of category for protocol

jecisc:
This change rename some "categories" into "protocols". Mostly renaming things in MethodChunk and some users in MethodDefinition


Ducasse:
unrelated

PR #15167, opened 2023-10-29
Add support to Unix Domain Sockets

### LINKED ISSUE #4046: Support Unix domain sockets.
The socket plugin already supports Unix domain sockets. I have code for Pharo 2.0 that I will adapt to current Pharo.

fcr--:
There was already some degree of support, but as seen in this blog post a few more changes were required:

https://samadhiweb.com/blog/2013.07.27.unixdomainsockets.html

fixes #4046

I was testing it with:
```smalltalk
sa := NetNameResolver addressForUnixDomain: '/tmp/sp'.
Socket newIPC ifNotNil: [ :s |
	s connectTo: sa.
	s sendData: 'Hello from Smalltalk!'.
	Transcript show: s receiveData; cr.
	s closeAndDestroy. ].
```
image

### DIFF
--- a/src/Network-Kernel/NetNameResolver.class.st
+++ b/src/Network-Kernel/NetNameResolver.class.st
@@ -58,6 +58,25 @@ NetNameResolver class >> addressForName: hostName timeout: secs [
 	^ result asSocketAddress
 ]
 
+{ #category : 'lookups' }
+NetNameResolver class >> addressForUnixDomain: path [
+	"Create a socket address that can be used on Socket newIPC connectTo:"
+
+	| deadline size result |
+	deadline := Time millisecondClockValue -> (60 * 1000).
+	result := nil.
+
+	self resolverMutex critical: [
+		(self waitForResolverReadyUntil: deadline)
+		ifTrue: [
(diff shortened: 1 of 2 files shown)


svenvc:
Hi Fransisco,

Thanks for your contribution. I have looked at your changes briefly.

Now, you are probably not aware that some of the networking stuff is developed in Zinc HTTP Components in a separate repository. We just recently added some code in this area as well. Now we need to compare the two of them. In any case, we were also able to get the client side working, but got stuck on the server side part which we wanted to be able to write unit tests.

There is a new package https://github.com/svenvc/zinc/tree/master/repository/Zinc-HTTP-UnixSocket.package

You can check some of our commits from Aug 30 https://github.com/svenvc/zinc/commits/master or some of the closed PRs https://github.com/svenvc/zinc/pulls?q=is%3Apr+is%3Aclosed which contain conversations about this subject.

It would be great if we (me and @JurajKubelka) could work together somehow.

Sven

Ducasse:
Thanks. This is super nice to see some collaborations on such topics!
We are going in this iteration over a massive internal cleaning and we are getting there.

PR #15168, opened 2023-10-29
Deprecated `sender` method of block closures

### LINKED ISSUE #15062: Deprecate #sender on BlockClosure
**Bug description**
We have #sender implemented on BlockClosure. It is implemented as:

```
sender
	"Answer the context that sent the message that created the receiver."
	<reflection: 'Message sending and code execution - Message send reification'>
	^outerContext sender
```

and CleanBlockClosure returns 'nil' as it has not outerContext.

I think we should remove it:

- The concepter of "sender" is one of the Stack: it is the stack frame above the current stack frame. BlockClosures do not have "senders".
- using the outerContext is odd. The block is "created" when executing it's home context, not it's outer context
- there are blocks without outerScope

Clients should use the outerContext directly and deal with it being possibly nil.

jordanmontt:
Fixes  #15062

### DIFF
--- a/src/Deprecated12/BlockClosure.extension.st
+++ b/src/Deprecated12/BlockClosure.extension.st
@@ -1,0 +1,10 @@
+Extension { #name : 'BlockClosure' }
+
+{ #category : '*Deprecated12' }
+BlockClosure >> sender [
+	<reflection: 'Message sending and code execution - Message send reification'>
+
+	self deprecated:
+		'The sender is the one of the Stack. BlockClosures do not have senders. The new CleanBlocks introduce blocks withput the outerContext. Now, clients should use the outerContext directly and deal with it being possibly nil.'.
+	^ self subclassResponsibility
+]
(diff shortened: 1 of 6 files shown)


jecisc:
We have some deprecations showing up in the logs such as:

*** Warning: Deprecation: The method FullBlockClosure>>#sender called from MessageTally class>>#tallySendsTo:inBlock:showTree:closeAfter:openResultWindow: has been deprecated. The sender is the one of the Stack. BlockClosures do not have senders. The new CleanBlocks introduce blocks withput the outerContext. Now, clients should use the outerContext directly and deal with it being possibly nil.

We already have a lot of deprecations showing up in the logs after the last sprint, I would like to avoid accumulating too many if possible (because we need to fix all of those before the release)

jordanmontt:
Do you mean that you prefer to wait before merging the PR? Or should I replace the usages of `sender`?

jecisc:
We should remove the users of FullBlockClosure>>#sender before merging this PR I think

PR #15175, opened 2023-10-30
Fix random bug in auto completion

jecisc:
The completion base some entries building on a lookup in the symbols of the system in some places. The current implementation has some bad side effects which are: if a GC happens during the creation of the menus, it is possible that we lose the first entries. (Maybe this is why somtimes we do not get proposed some of the best matches?)

I'm fixing this inconsistency by manipulating only non weak classes

Found with @jordanmontt


jordanmontt (approved):
Good!

MarcusDenker (approved):

PR #15176, opened 2023-10-30
Calypso: Project view should activate class creation tool

jecisc:
So that we can create a class when in project view with a project selected  (or not)


jecisc:
Now: 
image

Before:
image

MarcusDenker (approved):

PR #15180, opened 2023-10-30
RPackage: Do not register to MethodAdded

jecisc:
Do not register to MethodAdded but react directly at the addition of a method

### DIFF
--- a/src/Kernel/CompiledMethod.class.st
+++ b/src/Kernel/CompiledMethod.class.st
@@ -915,6 +915,14 @@ CompiledMethod >> setSourcePointer: srcPointer [
 	start + 1 to: self size do: [:n | self at: n put: (trailerBytes at: n - start) ]
 ]
 
+{ #category : 'testing' }
+CompiledMethod >> shouldBePackaged [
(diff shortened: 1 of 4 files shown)


src/Kernel/CompiledMethod.class.st
```diff
@@ -915,6 +915,14 @@ CompiledMethod >> setSourcePointer: srcPointer [
 	start + 1 to: self size do: [:n | self at: n put: (trailerBytes at: n - start) ]
 ]
 
+{ #category : 'testing' }
+CompiledMethod >> shouldBePackaged [
+	"This method is a hack.. Ideally this code should be inlined in ClassDescription>>#addAndClassifySelector:withMethod:inProtocol: but ObjectsAsMethodsExample needs this in order to work. I don't like this. If someone has a better idea on how to not packages this kind of ObjectAsMethod go ahead please!"
```
MarcusDenker, line 920: We can keep it like that for now... the idea is to remove the need of run:with:in: and replace it with MethodProxies in the future.

MarcusDenker (approved):

PR #15182, opened 2023-10-30
Remove #acceptsLoggingOfCompilation

### LINKED ISSUE #15171: Should we keep #acceptsLoggingOfCompilation?
Maybe @MarcusDenker knows why this is here?

This methods only returns true in the image and the only class implementing it has an implementation returning true (except for its subclasses but it has none). 

Maybe we could just remove that and always log except if we configure to not log things?

jecisc:
This method is a hook that is never used in the image. If someone wants to not log, Opal offers an API for that and if a class want to avoid logging, it can override #compiler on the class side to set the logging to false. It will have the same effect

Fixes #15171

### DIFF
--- a/src/Graphics-Fonts/FontSet.class.st
+++ b/src/Graphics-Fonts/FontSet.class.st
@@ -22,14 +22,6 @@ Class {
 	#package : 'Graphics-Fonts'
 }
 
-{ #category : 'compiling' }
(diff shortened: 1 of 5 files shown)


MarcusDenker (approved):
Good!

We can think in a later step if making compiling without logging should be exposed via the API on Class: ClassDescription>>#compile:classified:withStamp:notifying:logSource:

PR #15194, opened 2023-10-31
Inline date printer

jecisc:
BasicDatePrinter and ExtendedDatePrinter introduce problems of depenedencies. BasicDatePrinter is never used because its #default method is returning ExtendedDatePrinter. On top of this, we have a dependency problem because ExtendedDatePrinter is in a package that is loaded later in the bootstrap. 

I propose to just inline things to make the code simpler. I deprecated those two classes


guillep:
wait no! the basic date pinter is used during the bootstrap. I understand that minimising dependencies is important, but inlining does not necessarily solve the issue as it removes a hook. remember not all hooks are used inside the image. if we only codw for what is inside the image, we will never produce an extensible system -------- Message original --------De : CyrilFerlicot ***@***.***>Date : mar. 31 oct. 2023 à 12:03À : pharo-project/pharo ***@***.***>Cc : Subscribed ***@***.***>Objet : [pharo-project/pharo] Inline date printer (PR #15194)BasicDatePrinter and ExtendedDatePrinter introduce problems of depenedencies. BasicDatePrinter is never used because its #default method is returning ExtendedDatePrinter. On top of this, we have a dependency problem because ExtendedDatePrinter is in a package that is loaded later in the bootstrap.
I propose to just inline things to make the code simpler. I deprecated those two classes



You can view, comment on, or merge this pull request online at:
  https://github.com/pharo-project/pharo/pull/15194

Commit Summary
e097bb9  Inline date printer

File Changes  (8 files)

    A
    src/Deprecated12/BasicDatePrinter.class.st
    (63)
  
    A
    src/Deprecated12/ExtendedDatePrinter.class.st
    (18)
  
    M
    src/Kernel-Chronology-Extras/DateAndTime.extension.st
    (7)
  
    D
    src/Kernel-Chronology-Extras/ExtendedDatePrinter.class.st
    (61)
  
    M
    src/Kernel-Tests/DateTest.class.st
    (12)
  
    D
    src/Kernel/BasicDatePrinter.class.st
    (110)
  
    M
    src/Kernel/Date.class.st
    (49)
  
    M
    src/Kernel/DateAndTime.class.st
    (31)
  

Patch Links:
https://github.com/pharo-project/pharo/pull/15194.patchhttps://github.com/pharo-project/pharo/pull/15194.diff

jecisc:
Do you know where it is used? I have a hard time to find it.

Also, I’m not doing this for the sake of cutting the dependency. I have other changes failing because Kernel is referencing the ExtendingDatePrinter and this one is not present in Kernel

jecisc:
The bootstrap is working with this PR, so I guess the bootstrap does not use it in the end? (Since it’s moved to Deprecated12, it is loaded at the very end)

Ducasse:
Personally I would like to have more printers like objects. 
So we could also move the extended data printer in the kernel if needed.

jecisc:
I'll note on my todo list to merge the printers and I would like to introduce a cached instance to avoid creating a new instance each time we want to print a date because for now it creates a new instance all the time which is not efficient, especially since this class has no state

jecisc:
I'll close this one and when I'll have the time I'll propose a new one based on the discussion. But Idk when that will be :p

PR #15197, opened 2023-10-31
SystemDictionary>>declare:from: should not be an extension

jecisc:
System dictionary really relies on this method, this should be packaged in SystemDictionary

Maybe this is why we have Associations in the SystemDictionary during the bootstrap? Since the method was an extensions it was not loaded at the beginning and it was using the version of Dictionary? To be tested


MarcusDenker (approved):

MarcusDenker:
There is another issue 

- declare:from: is badly named and badly located #11903

Which is true and should be fixed later, this is independent of moving the method

Issue #15198, opened 2023-10-31
ThemeIcons>>#iconNamed:ifNone: often falls back on error handling code


MarcusDenker:
**Describe the problem**
Icon naming was changed many years ago to not have Icon or Form suffixes.

ThemeIcons>>#iconNamed:ifNone: years ago was changed to have a fall back code, which creates a symbol every time...

```
iconNamed: aSymbol ifNone: aBlock

	self icons at: aSymbol asSymbol ifPresent: [ :icon | ^ icon scaledToSize: icon extent * (self currentWorld displayScaleFactor / self scale) ].
	"Trying the old way"

	((aSymbol endsWith: 'Icon') or: [ (aSymbol endsWith: 'Form') ]) ifTrue: [
		self icons
			at: (aSymbol allButLast: 4) asSymbol
			ifPresent: [ :icon | ^ icon scaledToSize: icon extent * (self currentWorld displayScaleFactor / self scale) ]
	].

	^ aBlock value
```

We fixed already many places when reducing calls to #asSymbol at runtime during Pharo11, but there are many more left.

PR #15200, opened 2023-10-31
add trace to #iconNamed:ifNone: to log all old names

MarcusDenker:
For #15198

- add trace to #iconNamed:ifNone:
- fix some


jecisc (approved):

MarcusDenker:
I should fix more of the logs we see when running the tests before merging

Ducasse:
I did not understand the PR because it is no in sync with the title.
I see just fixing names of icons (which is ok).

Ducasse:
So I did not why the work needed status is set. So I will not integrate it.

jecisc:
The change was to add a log when we use the fallback code for icons. But there are still way to many uses of this fallback and this floods the logs. This is the reason Marcus added the need more work tag

jecisc (approved):

jecisc:
I think we can merge and fix the last icons in a next PR?

PR #15208, opened 2023-10-31
RPackage: Remove last listeners to SystemAnnouncer

jecisc:
This change remove the last announcement RPackageOrganizer was listening to! Now everything about RPackage synchronization is done directly in the MM and not via announcement!

The last one was the recategorization of methods. 

After this PR we will also be able to remove RPackage class>>#withOrganizer:do:


jecisc:
And with this we can remove a bootstrap script file :)

PR #15210, opened 2023-11-01
Bootstrap: Simplify management of protocols

jecisc:
This PR simplifies the bootstrap on the management of protocols. This is possible because RPackage got improved recently


jecisc:
Some dependency analyzer tests are failing :( I need to check why

jecisc:
I'll close because it does not seems to work

PR #15212, opened 2023-11-01
Rpackage: Extension methods should know their package

jecisc:
Now extension methods know directly their package which will speed up a lot of things. (See bellow for some examples) Those speed up will allow me to remove multiple caches in RPackage and simplify a lot the code of RPackage! 

Another change this will enable is for the methods to be able to return their package even after they are removed from the system. The next step to enable this is for the class to still know their package after removal (for non extension methods). This is planed for the next few weeks! This will also simplify and speedup more things in the future!

I'm kinda excited by all the speedup and simplifications this will allow me to do and the code to do that is not that complex.

Benches:

```st
allMethods := CompiledMethod allInstances.

[ allMethods collect: #isExtension] timeToRun. 

Before: "0:00:00:00.358"

After: "0:00:00:00.005" 

[ allMethods collect: #package ] timeToRun. 

Before: "0:00:00:00.455"

After: "0:00:00:00.195"  
```

I might add more things to this PR to start removing the caches

Edit: I added the removal of a cache. Here are the benches:

```st
allClasses := Smalltalk allClassesAndTraits.

[ allClasses collect: #extendingPackages ] timeToRun.

Before: "0:00:00:00.009"

After: "0:00:00:00.038"
```

This is slower but I think it is worth it because of the reduction of the complexity of the code and also this code is not called a lot compared to #package or #isExtension

### DIFF
--- a/src/JenkinsTools-Core/HDReport.class.st
+++ b/src/JenkinsTools-Core/HDReport.class.st
@@ -22,14 +22,15 @@ HDReport class >> runPackage: aString [
 HDReport class >> runPackages: aCollectionOfStrings [
 
 	^ aCollectionOfStrings collect: [ :packageName |
-		  | time |
+		  | time result |
 		  time := DateAndTime now.
 		  Transcript << 'Beginning to run tests of ' << packageName << OSPlatform current lineEnding.
 		  "We flush so that if a crash happens during the tests, we print in which package is the naughty test in the logs."
 		  Transcript flush.
-		  self runPackage: packageName.
+		  result := self runPackage: packageName.
 		  Transcript << 'Finished to run tests of ' << packageName << ' in ' << (DateAndTime now - time) humanReadablePrintString << OSPlatform current lineEnding.
-		  Transcript flush ]
+		  Transcript flush.
+		  result ]
 ]
 
 { #category : 'private' }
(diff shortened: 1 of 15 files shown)


jecisc:
I've check the generated image and some methods seems to be considered as defined methods while they are extension methods :( I need to find why

jecisc:
It seems to come from recompilation, let's fix that

astares:
Cool

jecisc:
I can confirm my fix produces an image that seems right

Ducasse:
Cyril now where is stored the package of a method? In the method property I imagine.

Ducasse:
A test I would like to see in particular for the tags is to open calypso on Bloc core package

jecisc:
I save it in the method properties indeed.
I checked and we do not have that much extension methods so it will not add too much memory. And also, with the caches I’ll be able to remove, we will probably gain memory in the end :)

To test the opening on Calypso on Bloc we would need Bloc in the image first

jecisc:
I think I found another bug with this. If I remove an ivar from a class, the property is lost. I think this is because when migrating the methods from the old to the new class there is no prior method present to migrate the property. I'll fix that

jecisc:
I just added a few tests for the problems I found and fix

jecisc:
It seems that we have a test crashing but nothing to help in the logs :(

Ducasse:
Once this is integrated I will load Bloc in P12. Bloc works well in P11 so it should work on P12.

jecisc:
So what was breaking now it the classes that acts like methods with the method #run:with:in: 

I find this system horrible because we have to duplicate part of the compiled method API in a lot of places and each time we change something in the methods management something breaks :(

jecisc:
This is now working! :)

Ducasse:
I agree with you and we should remove the things that are bizarre and unefficient. 
If you have a list of the API that is duplicated please log it so that we can clean it.

PR #15214, opened 2023-11-02
RPackage: Deprecate #includesDefinedSelector:ofClass:

jecisc:
I deprecated #includesDefinedSelector:ofClass: since this is now something we should delegate to the class after the recent changes in RPAckage

I also took the liberty to improve a bunch of tests. RPackage tests still have problems but this will be for other PRs :(


jecisc:
So, all failing tests are about the dependencies.

Either the implementation without the cache does not behave exactly like before. Or the cache had a synchronization problem and we did not find some results in the past. I'll need to investigate

PR #15215, opened 2023-11-02
Merge Metacello-PharoExtensions into Metacello-Base

jecisc:
Metacello-PharoExtensions contains only methods that could make sens to have directly in Metacello and is a package there only for 6 methods. I think it is safe to merge it directly into Metacello.

This also simplifies the bootstrap


Ducasse:
Good move. Now is MCCacheRepository not introducing a dependency?

jecisc:
We load Metacello after Monticello so this should not be a problem

PR #15217, opened 2023-11-02
Improve logs while running tests

jecisc:
The goal of the change is to improve logs while running tests.

It prints when it starts, when it ends and how long it took.

Also it now uses the Transcript instead of Stdout to be sure the logs are in the right order (because transcript is buffered). Also I flush often because if a test crashes the image while the buffer of the transcript is not flushed. we cannot know when the problem occured.

### DIFF
--- a/src/JenkinsTools-Core/HDReport.class.st
+++ b/src/JenkinsTools-Core/HDReport.class.st
@@ -22,14 +22,15 @@ HDReport class >> runPackage: aString [
 HDReport class >> runPackages: aCollectionOfStrings [
 
 	^ aCollectionOfStrings collect: [ :packageName |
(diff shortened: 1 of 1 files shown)


Ducasse:
Why not introducing a transcript log in addition to the Stdout because stdout is for things that potentially do not save the image or when we cannot access the Transcript. Like that you get the transcript version if you are in interactive mode and the Stdout when we do not save nor can access the image.

jecisc:
I wanted to avoid having twice the logs and transcript prints in the stdout anyway. That’s mostly the reason

Ducasse:
Ok I did not know that transcript was printing to stdout.

PR #15219, opened 2023-11-03
Packages: Do some cleanings 

jecisc:
This change cleans some things in the package management

- Remove some #ifNotNil: that cannot have nil now
- Remove hack in MCClassDefinition that was there for tonel
- Simplify #promoteAsPackage
- Unskip a test that does not fail anymore
- Inline #toTagName in ClassDescription>>category: because this method is bad

### DIFF
--- a/src/Kernel/Class.class.st
+++ b/src/Kernel/Class.class.st
@@ -275,16 +275,25 @@ Class >> category: aString [
 	any previous categorization."
 
 	| newPackage |
-
 	self flag: #package. "Should be deprecated"
 	self category = aString ifTrue: [ ^ self ].
 
 	self basicCategory: aString asSymbol.
 
-	self flag: #package. "#toTagName: should be inlined here."
 	newPackage := self packageOrganizer ensurePackageMatching: self category.
-	newPackage moveClass: self toTag: (newPackage toTagName: self category).
 
+	"The category can be
+- A tag name
+- A package name
+- A package name - a tag name
+
+We are trying to find the right one here."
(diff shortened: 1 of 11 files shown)


src/Kernel/Class.class.st
```diff
@@ -275,16 +275,25 @@ Class >> category: aString [
...
 	self flag: #package. "Should be deprecated"
 	self category = aString ifTrue: [ ^ self ].
 
 	self basicCategory: aString asSymbol.
 
-	self flag: #package. "#toTagName: should be inlined here."
 	newPackage := self packageOrganizer ensurePackageMatching: self category.
-	newPackage moveClass: self toTag: (newPackage toTagName: self category).
 
+	"The category can be
```
Ducasse, line 285: I imagine that this is only used by the tonel importer? I mean normally except on such case even the classbuilder should not use this API so we could package this will tonel?

Ducasse:
Remark about category: definition in Class.
I imagine that this is only used by the tonel importer? I mean normally except on such case even the classbuilder should not use this API so we could package this will tonel?

src/Kernel/Class.class.st
Ducasse: I will merge now and may be we could improve the comment later. 
Or push this method to tonel if my hypothesis is correct.

jecisc: This is almost not used anymore. I just need to remove 2-3 senders then I’ll deprecate it because external projects could still be using it

PR #15221, opened 2023-11-03
Calypso: Simplify ClyExtendingPackagesQuery adn clean package API

jecisc:
This change mainly simplifies the code of ClyExtendingPackagesQuery to make it easier to understand and faster (At least it will be faster when #15212 will be integrated).

I also deprecated #isDefinedInPackage: because it the bahavuior was unclear and it is easy to do without this method polluting the API (and it is not used).

Thie behavior was unclear because in some part of the systemit returned true if the entity was in the package as parameter while in other part it returned true if the entity was in the package AND was not an extension (extensions could never return true)

### DIFF
--- a/src/Calypso-Ring/RGMethod.extension.st
+++ b/src/Calypso-Ring/RGMethod.extension.st
@@ -37,12 +37,6 @@ RGMethod >> implementors [
 ]
 
 { #category : '*Calypso-Ring' }
-RGMethod >> isDefinedInPackage: aPackage [
-
-	^ self package = aPackage
-]
-
-{ #category : '*Calypso-Ring' }
 RGMethod >> isInstalled [
 	^ true
 ]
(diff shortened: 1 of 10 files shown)


Ducasse:
I will merge. Now I was wondering why isDefinedIn: was that bad (for me it is more intention revealing that self package = package.

jecisc:
First because it is ambiguous. 

Some places implemented it like `method package = package` and some methods implemented it in a way that it was equivalent to `method isExtension not and: [ method package = package ]`.

And I don't know which version should be the right one. Also, after the code simplification of Calypso that did not use at all this anymore, there was no sender at all in the image. So if a method is ambiguous and has o sender I though maybe it was better to deprecate it

Ducasse:
Thanks for the explanation.

PR #15225, opened 2023-11-03
Feat: Add Override method transformation

balsa-sarenac:
This is a proposal for new refactoring that will be useful in composition.

Regular AddMethodRefactoring does not allow override of a method from superclass or the class.
In composition it's usually only important not to override existing method from the class, and most of the refactorings additionally check this and use AddMethodTransformation, because AddMethodRefactoring breaking change is too restrictive.

This new refactoring allows override from methods from superclass, but it does not allow override of method directly defined in the class.

It is required in multiple places, but I cannot fully refactor it now. I will get back to this when we start working on composition again and have clear solution for checking preconditions. I just put placeholders for the future.

### DIFF
--- a/src/Refactoring-Core/RBAddOverrideMethodRefactoring.class.st
+++ b/src/Refactoring-Core/RBAddOverrideMethodRefactoring.class.st
@@ -1,0 +1,20 @@
+"
+I am a subclass of Add Method Refactoring that loosens preconditions.
+
+Add Method has the breaking change that methods should not be defined in the target class or its superclasses.
+Add Override Method allows Refactoring to create a method in a class that can override a superclass method, but it does not allow overriding of a method that is directly defined in that class.
+
+"
+Class {
+	#name : 'RBAddOverrideMethodRefactoring',
+	#superclass : 'RBAddMethodRefactoring',
+	#category : 'Refactoring-Core-Refactorings',
+	#package : 'Refactoring-Core',
+	#tag : 'Refactorings'
+}
+
+{ #category : 'preconditions' }
+RBAddOverrideMethodRefactoring >> preconditions [
+
+	^ (RBCondition definesSelector: transformation selector in: class) not
+]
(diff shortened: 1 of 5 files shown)


Ducasse:
Ok for me. This is a subtle difference. I would like to know why this is not a transformation since it is changing behavior. But I integrate it now.

balsa-sarenac:
Good point, it's a transformation definitely, I will fix it in this PR since there are some failing tests I need to fix as well.

Ducasse:
Balsa

```
osx-64 / Tests-osx-64 / testExistingClassVariableAccessors(#rbClass->RBAddVariableAccessorWithLazyInitializationTransformation) – MacOSX64.Refactoring.Transformations.Tests.Parametrized.RBAddVariableAccessorsWithLazyInitializationParametrizedTest
<1s
osx-64 / Tests-osx-64 / testExistingInstanceVariableAccessors(#rbClass->RBAddVariableAccessorWithLazyInitializationTransformation) – MacOSX64.Refactoring.Transformations.Tests.Parametrized.RBAddVariableAccessorsWithLazyInitializationParametrizedTest
<1s
osx-64 / Tests-osx-64 / testDeprecateMethodUsingMethodWithSameNumberOfArgs(#rbClass->RBDeprecateMethodTransformation) – MacOSX64.Refactoring.Transformations.Tests.Parametrized.RBDeprecateMethodParametrizedTest
6s
osx-64 / Tests-osx-64 / testDeprecateMethodUsingMethodWithoutArgs(#rbClass->RBDeprecateMethodTransformation) – MacOSX64.Refactoring.Transformations.Tests.Parametrized.RBDeprecateMethodParametrizedTest
```

balsa-sarenac:
@Ducasse I've reverted references and will fix that in the future. These refactorings need to raise warnings, and for that they need driver. So this PR introduces new transformation, in a later one there will be a driver for it.

Ducasse:
Ok Thursday we should have a meeting with caro and I think that she should focus on the drivers. 
Think about a better list than the last time.

Ducasse:
I will merge when build is ready.

PR #15226, opened 2023-11-03
Calypso: Fix taskbar label

Gabriel-Darbord:
After #15220, the taskbar label still had an extraneous leading whitespace.
image
image

### DIFF
--- a/src/Calypso-SystemTools-FullBrowser/ClyFullBrowserMorph.class.st
+++ b/src/Calypso-SystemTools-FullBrowser/ClyFullBrowserMorph.class.st
@@ -873,7 +873,7 @@ ClyFullBrowserMorph >> switchToVariables [
 { #category : 'taskbar-required' }
 ClyFullBrowserMorph >> taskbarLabel [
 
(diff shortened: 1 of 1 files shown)


src/Calypso-SystemTools-FullBrowser/ClyFullBrowserMorph.class.st
```diff
@@ -873,7 +873,7 @@ ClyFullBrowserMorph >> switchToVariables [
 { #category : 'taskbar-required' }
 ClyFullBrowserMorph >> taskbarLabel [
 
-	^ self newWindowTitle copyAfter: $:
+	^ self newWindowTitle copyAfter: Character space
```
jecisc:
```suggestion
	| title |
	title := self newWindowTitle.
	^ title allButFirst: (title findString: ': ')
```

I would prefer to still check for the ':' in case at some point we have a title without the type prefix (I'm not sure that I managed all possible titles of Calypso and this code above has less risk of breaking the taskbar name)

jecisc (changes requested):

PR #15228, opened 2023-11-04
[WIP] Redo "Revert "Deprecate String>>asDate""

guillep:
Redo https://github.com/pharo-project/pharo/pull/15152, originally reverted because there are still broken tests.

```smalltalk
```

```smalltalk
```

We deprecate methods: 
- String>>asDate,
- Date>>fromString, and 
- Date>>readFrom.


guillep:
no! not again! please do not integrate of markedvas WIP!

PR #15229, opened 2023-11-04
Proposal: new message browser layout

guillep:
Screens are wide, not tall.
I often do not have space to browse methods and large lists of methods in the current message browser.
This is aggravated by the critics browser shown in the bottom and the space taken by the tabbing system.

I propose to slightly change the layout by:
 - making the main content left-to-right
 - put the toolbar on the top, as toolbars are

imagen

This is how it looks with critics:

imagen

Just a small refactoring and one new method defining the new layout (only for the message browser).
I think this can make the life easier for many people.


Ducasse:
Nice! With this setup we could even see multiple method definitions side by side :)

PR #15230, opened 2023-11-04
Ring: Simplify Ring-Definitions/Ring-Definitions-Containers

jecisc:
We have a lot of versions of ring in the Pharo ecosystem:
- Ring-Definitions (in Pharo)
- Ring (In Pharo)
- Ring2 (Used by Moose)
- A fork of Ring2 (used by Iceberg)

It is quite hard to maintain all of this and I'm trying to update them to move from categories to packages and tags. 

I don't think it is possible in the long run to keep so many ring versions alive. Ring-Definitions is the oldest and I propose to remove everything that we do not currently use in Pharo and this PR is a first big part of this change.

- Remove RGSlice that is not used
- Remove RGOrganization  that does not seems to be used
- Remove RGNamespace that is not used
- Remove the possibility to have inner packages in RGPackageDefinition
- Simplify RGPackageDefinition (remove dead code, remove checks about dictionaries or other kind of collections when we only use dictionaries|, remove the whole "elements" API...)
- Merge RGAbstractContainer, RGContainer and RGPackageDefinition
- Since Ring-Definitions-Containers contained only one class in the end, I moved it in Ring-Definitions
- Simplify the way to set the package of RGBehaviorDefinition and RGMethodDefinition to not rely on Monticello and have a more simple code
- Remove RGGlobalVariableDescription that is not used


Ducasse:
Good idea!
Iceberg tests are loading RG2. Ring one is limited but we could migrate RG2 and eliminate RG in the future 
if we want to support remote working. I guess that RG is good enough for the package browse when wanted to see the code without loading it.

PR #15232, opened 2023-11-04
Redo "Revert "Deprecate String>>asDate"

guillep:
Redo https://github.com/pharo-project/pharo/pull/15152, originally reverted because there are still broken tests.

```smalltalk
```

```smalltalk
```

We deprecate methods: 
- String>>asDate,
- Date>>fromString, and 
- Date>>readFrom.


src/Calypso-SystemQueries/Project.extension.st
```diff
@@ -1,7 +1,7 @@
 Extension { #name : 'Project' }
 
 { #category : '*Calypso-SystemQueries' }
-Project class >> convertToCalypsoBrowserItem: aProject [
```
guillep, line 4: I got this change because of a merge. @jecisc do you know why?

jecisc: I have no idea where those "classSide" are coming from.. I've see it in the past in some projects but I never discovered why. 

I never do my merges with Pharo, maybe it's the Pharo merger that uses another printer?

guillep:
Strange, I have lots of tests failing that look unrelated to my changes, and look more related to the handling of protocols and extension methods, any idea?

imagen

guillep:
They work in my local image based on build 1020.

Pharo 12.0.0
Build information: Pharo-12.0.0+build.1020.sha.373a7ce1fe48d19844129ab9086329fcf758ce50 (64 Bit)

jecisc:
This is a known failure I'm working on that appears really recently: https://github.com/pharo-project/pharo/issues/15282

The error is really strange because it happens when updating the users of a trait, but the trait is generated in the tests and has no user (: 

I'll try to fix it this week

guillep:
Oh! Not my fault! Then I can remove the WIP from this PR :)

guillep:
Why closing it instead of merging it?

PR #15234, opened 2023-11-05
Simplify Ring-Definitions packaging

jecisc:
- Merge Ring-Definitions-Tests-Containers in Ring-Definitions-Tests since it now has only one class
- Merge Ring-Definitions-Monticello in Monticello because Monticello depends on ring and we have a circular dependency between Monticello and Ring-Definitions-Monticello
- Merge Ring-Definitions-Tests-Monticello in Monticello-Tests


jecisc:
I'll close this one for now

PR #15239, opened 2023-11-06
Make class know its package tag instead of its category

jecisc:
This change has for goal to make classes know their package tag instead of knowing their category.

This is the next huge change to get rid of categories in the system! 

Changes:
- Removed #category ivar from Class 
- Added #packageTag ivar to Class
- Updated Fuel and RPackage to use the new var
- Removed all users of #basicCategory 
- Remove the class and packages cache from RPackage because now we do not need it to know the package of a class! :)
- Deprecate RPackageOrganizer class>>#isPackageDefinedForClass: because this was referring to a global
- Instead of the previous method I deprecated I implemented ClassDescription>>isInstalled


jecisc:
To fix the tests this requires: https://github.com/pharo-project/pharo/pull/15243

PR #15240, opened 2023-11-06
15191-Undeclared-cleanOutUndeclared-should-not-iterate-over-all-methods 

### LINKED ISSUE #11903: declare:from: is badly named and badly located

### LINKED ISSUE #15191: [Undeclared] #cleanOutUndeclared should not iterate over all methods
SmalltalkImage>>#cleanOutUndeclared now iterates over all methods and removes those from the Undeclared dictionary that it does not find:

(this is in Pharo12 already optimized to iterate just once, thanks to https://github.com/pharo-project/pharo/pull/14908 )

```
cleanOutUndeclared
	| unreferenced |
	unreferenced := Undeclared keys asSet.
	self systemNavigation allBehaviorsDo: [ :class |
		class methodsDo: [ :method |
			method withAllNestedLiteralsDo: [ :lit |
				(lit isKindOf: UndeclaredVariable) ifTrue: [
					unreferenced
						remove: lit key
						ifAbsent: [ "Sanity check--a binding with more than one reference may already be gone from the unreferenced set, but should be present in Undeclared itself"
							self at: lit key ] ] ] ] ].

	unreferenced do: [ :key | Undeclared removeKey: key ]
```

But now, with the compiler registering all methods with an  UndeclaredVariable in that object, we can do even better:

```
cleanOutUndeclared2
	| unreferenced |
	unreferenced := Undeclared associations select: [:var | var usingMethods isEmpty].
	unreferenced do: [ :var | Undeclared removeKey: var name ].
```

After PR #15190, #usingMethods only returns the installed methods, making this the same as the iterating over all methods version we had before.

MarcusDenker:
This PR now does:

- Inline and remove the private method #declare:from: from Dictionaries
- This allows to make sure that when we move variables to and from Undeclared, we correctly change the class (this code is not nice, we need to improve the Undeclared fixing in a second step)
- fix tests to not put a Association into the Undeclared Dict
- fix SystemDictionary>>#at:put: to no use Associations when using #at:put: 

And the original starting point:
- do not iterate all methods for #cleanOutUndeclared, but use the recorded #usingMethods

fixes #15191
fixes #11903

### DIFF
--- a/src/Collections-Sequenceable-Tests/OrderedDictionaryTest.class.st
+++ b/src/Collections-Sequenceable-Tests/OrderedDictionaryTest.class.st
@@ -706,22 +706,6 @@ OrderedDictionaryTest >> testCopyEmpty [
 ]
 
 { #category : 'tests' }
-OrderedDictionaryTest >> testDeclareFrom [
-	| dictionary otherDictionary |
-	dictionary := self emptyDictionary.
-	otherDictionary := self dictionaryWithOrderedAssociations.
-	self orderedKeys do: [ :each | self assert: (dictionary declare: each from: otherDictionary) identicalTo: dictionary ].
-	self assertIsDictionary: dictionary withOrderedAssociations: self orderedAssociations.
-	self assertEmpty: otherDictionary.
-
-	self orderedKeys
-		do: [ :each |
-			otherDictionary add: each -> self newValue.
-			self assert: (dictionary declare: each from: otherDictionary) identicalTo: dictionary ].
-	self assertIsDictionary: dictionary withOrderedAssociations: self orderedAssociations
-]
-
-{ #category : 'tests' }
 OrderedDictionaryTest >> testDictionary [
 	| dictionary |
 	dictionary := self emptyDictionary.
--- a/src/Collections-Sequenceable/OrderedDictionary.class.st
+++ b/src/Collections-Sequenceable/OrderedDictionary.class.st
@@ -299,18 +299,6 @@ OrderedDictionary >> collect: aBlock [
 			each key -> (aBlock value: each value)])
 ]
 
-{ #category : 'adding' }
-OrderedDictionary >> declare: aKey from: aDictionary [
-	(self includesKey: aKey)
-		ifTrue: [^ self].
-
-	(aDictionary includesKey: aKey)
-		ifTrue: [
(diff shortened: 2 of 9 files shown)


jecisc (approved):

MarcusDenker:
Failing tests 

```
testClassObsolete – Unix64.ReleaseTests.CleanCode.ObsoleteTest
testFixObsoleteSharedPools – Unix64.ReleaseTests.CleanCode.ObsoleteTest
testTraitObsolete – Unix64.ReleaseTests.CleanCode.ObsoleteTest
testObsoleteClasses – Unix64.ReleaseTests.Release.ReleaseTest
testUndeclared – Unix64.ReleaseTests.Release.ReleaseTest
```

jecisc:
So we still have Associations in the SystemDictionary? This is weird!

jecisc:
Those tests are green in a clean image. This means that we have tests adding associations to the SystemDictionary! It's not the build of the image that is doing this.

MarcusDenker:
The associations might be in the Undeclared Dictionary

jecisc:
```
Message not understood: GlobalVariable >> #unregister
Stacktrace
MessageNotUnderstood
Message not understood: GlobalVariable >> #unregister
GlobalVariable(Object)>>doesNotUnderstand: #unregister
[ :var | var unregister ] in SmalltalkImage>>cleanOutUndeclared
Array(SequenceableCollection)>>select:thenDo:
SmalltalkImage>>cleanOutUndeclared
SmalltalkImage>>fixObsoleteSharedPools
SmalltalkImage>>fixObsoleteReferences
ObsoleteTest>>tearDown
```

MarcusDenker:
will redo

PR #15247, opened 2023-11-07
Stay in calypso project view when selecting a class

jecisc:
This is a big step to have the project view working. When we select a class whose package is not shown in the first pane, before we switched to package view. Now we stay in project view and we expand the item containing the package.

Depends on #15246

### DIFF
--- a/src/Calypso-SystemTools-FullBrowser/ClyFullBrowserMorph.class.st
+++ b/src/Calypso-SystemTools-FullBrowser/ClyFullBrowserMorph.class.st
@@ -657,27 +657,25 @@ ClyFullBrowserMorph >> selectPackage: aPackage [
 
 	self changeStateBy: [
 		packageView selection selectItemsWith: { aPackage }.
-		packageView selection isEmpty ifTrue: [ self switchToPackages ].
-		packageView selection selectItemsWith: { aPackage }.
-	]
+		packageView selection isEmpty ifTrue: [
+			self showInPackageView: aPackage.
+			packageView selection selectItemsWith: { aPackage } ] ]
 ]
 
 { #category : 'navigation' }
 ClyFullBrowserMorph >> selectPackage: aPackage andTag: tagName [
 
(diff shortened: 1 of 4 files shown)


src/Calypso-SystemTools-FullBrowser/ClyFullBrowserMorph.class.st
```diff
@@ -714,6 +712,19 @@ ClyFullBrowserMorph >> showAllMethods [
...
+ClyFullBrowserMorph >> showInPackageView: aPackage [
+
+	"If true we are in the project view and we need to find the package and expand it"
+	(packageView showsItemsFromQuery: ClyAllProjectsQuery)
+		ifTrue: [
+			packageView dataSource allElements
+				detect: [ :element | (element actualObject isKindOf: Project) and: [ element actualObject packages includes: aPackage ] ]
+				ifFound: [ :element | packageView dataSource expand: element ]
+				ifNone: [ "this should not happen but who knows with calypso.." self switchToPackages ] ]
+		ifFalse: [ self switchToPackages ]
```
dionisiydk, line 725: There is API to lookup items and it can be rewritten like: 
```Smalltalk
showInPackageView: aPackage

	(packageView showsItemsFromQuery: ClyAllProjectsQuery)
		ifTrue: [
			self projectSelection selectItemsWhere: [ :each | 
				(each representsItemOfType: Project) and: [ each actualObject includesPackage: aPackage ]].
			self projectSelection items first expand ].
	self packageSelection selectItemsWith: { aPackage }
```

dionisiydk: And this is actually how #selectPackage: should be rewritten

jecisc: I'll probably do another pass soon and integrate your feedback! Thanks

PR #15263, opened 2023-11-07
Fix trait method recompilation and clean

### LINKED ISSUE #15202: #recompile and methods from traits
**Bug description**
When using #recompile, on a method installed from a trait, the orgin method is not recompiled

```
trait := (Trait << #TToto package: 'MyTest') install.
    
trait compile: 'test
    ^ Titi new'.
    
class := (Object << #Toto traits: trait; package: 'MyTest') install.
    
(trait >> #test) bytecodes. "#[32 129 124 92]"
    
(class >> #test) bytecodes. "#[32 129 124 92]"

class2 := (Object << #Titi package: 'MyTest') install.

(trait >> #test) bytecodes.  "#[32 129 124 92]"

(class >> #test) bytecodes.  "#[32 129 124 92]"

"For now, in Pharo12, it recompiles the method with the undeclared on execution, we use that here just as a test for recompile"
class new test.

(trait >> #test) bytecodes.  "#[32 129 124 92]"
    
(class >> #test) bytecodes.  "#[16 124 92]"
```

We think that we should forwad the recompile to the origin:

```
recompile
	^ self origin recompile: self selector
```

and then fix TraitedMetaclass>>#recompile:from: to propagate to the users

jecisc:
The main goal of this PR is to fix the fact that recompiling a method from a trait does not update the method in the users to install the recompiled method. 

I also found some things that seems inconsistant and tried to clean them. All the traits tests are passing so I guess this is good?

Changes:
- CompiledMethod>>#recompile now ask to the origin to recompile the method and not the method class
- Recompiling on TraitedClass and TraitedMetaclass ask the TraitChange class to add the updated method to the users method dictionary
- Unify the recompilation code between TraitedMetaclass and TraitedClass. Here I am not sure about what I did! Maybe we should keep the code on the #traitSource property but add it also to TraitedClass?
- Remove TraitedClass class>>#addAndClassifySelector:withMethod:inProtocol: because I have the impression that the version on TraitedMetaclass is already managing this

Maybe @tesonep wants to check

I also added some tests

Fixes #15202

### DIFF
--- a/src/OpalCompiler-Core/CompiledMethod.extension.st
+++ b/src/OpalCompiler-Core/CompiledMethod.extension.st
@@ -28,7 +28,7 @@ CompiledMethod >> methodNode [
 
 { #category : '*OpalCompiler-Core' }
 CompiledMethod >> recompile [
-	^ self methodClass recompile: self selector
+	^ self origin recompile: self selector
 ]
 
 { #category : '*OpalCompiler-Core' }
--- a/src/TraitsV2-Tests/T2AbstractTest.class.st
+++ b/src/TraitsV2-Tests/T2AbstractTest.class.st
@@ -37,6 +37,11 @@ T2AbstractTest >> newClass: aName with: slots traits: aComposition [
 ]
 
 { #category : 'instance creation' }
+T2AbstractTest >> newTrait: aName [
+	^ self newTrait: aName with: #()
+]
+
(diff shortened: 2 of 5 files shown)


jecisc:
I checked in my image to ask #traitSource to some instance and class side methods from traits before and after recompilation and the property is present. So I think the method I removed could be safely removed :)

PR #15270, opened 2023-11-08
Remove unused Calypso-SystemPlugins-Monticello-Queries

jecisc:
Remove Calypso-SystemPlugins-Monticello-Queries because it does not seems to be used


Ducasse:
Youpi !

PR #15271, opened 2023-11-08
Introduce a specific collection for undeclareds

jecisc:
The goal of this change is to introduce a special collection for Undeclareds. This will allow us to add some constraints to what an undeclared is. For example we can ensure that undeclared contains only UndeclaredVariables and not an association.

This also fix a test that was adding association in the undeclareds

If this works and is integrated maybe @MarcusDenker could improve the class comment with his knowledge?

### DIFF
--- a/src/System-Support/UndeclaredHolder.class.st
+++ b/src/System-Support/UndeclaredHolder.class.st
@@ -1,0 +1,19 @@
+"
+I am a dictionary holding UndeclaredVariables in the system. UndeclaredVariables represent the usage of a class that is not (yet) present in the system.
+
+I am named UndeclaredHolder instead of UndeclaredDictionary because one goal is to transform me into a weak collection in the future so that we do not have to explicitly clean up the undeclareds and delegate the work to the GC.
+"
+Class {
+	#name : 'UndeclaredHolder',
+	#superclass : 'IdentityDictionary',
+	#category : 'System-Support-Utilities',
+	#package : 'System-Support',
+	#tag : 'Utilities'
+}
+
+{ #category : 'private' }
+UndeclaredHolder >> atNewIndex: index put: aUndeclaredVariable [
+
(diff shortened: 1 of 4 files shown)


src/System-Support/UndeclaredHolder.class.st
```diff
@@ -0,0 +1,12 @@
+"
+I am a dictionary holding UndeclaredVariables in the system. UndeclaredVariables represent the usage of a class that is not (yet) present in the system.
+
+I am named UndeclaredHolder instead of UndeclaredDictionary because one goal is to transform me into a weak collection in the future so that we do not have to explicitly clean up the undeclareds and delegate the work to the GC.
+"
+Class {
+	#name : 'UndeclaredHolder',
+	#superclass : 'IdentityDictionary',
```
dionisiydk: It would be better to not inherit dictionary but to wrap it with a dedicated API to register and lookup vars (instead of general collection methods). But it can be achieved later I guess.
Also what about different name? Something like UndeclaredRegistry.

jecisc: This is a first step because a lot of things are initialized in the bootstrap and it is really hard to do a big change there. So we want to cleanup this step by step to not be overwhelmed by a ton a different errors hard to debug (we already tried bigger changes in the past and we got bootstrap crashes that were pretty obscure)

I have to say that I like #UndeclaredRegistry better than #UndeclaredHolder. I'll update it thanks

MarcusDenker (approved):

src/System-Support/UndeclaredHolder.class.st
MarcusDenker: Indeed, just a tiny step of many.. this could even end up being a class variable on UndeclatedVariable (one global less!). But step by step.

daniels220:
Hah, I see I kicked off a flurry of further improvements to this—nice!

PR #15276, opened 2023-11-08
Fix some icon references

jecisc:
Also remove some logs I forgot in dependency analyzer

### DIFF
--- a/src/Athens-Examples/AthensSurfaceExamples.class.st
+++ b/src/Athens-Examples/AthensSurfaceExamples.class.st
@@ -655,7 +655,7 @@ AthensSurfaceExamples class >> exampleDrawForm2 [
 		can draw.
 		can pathTransform scaleBy: 0.5;
 		 translateX: 60 Y: 60.
-		(can setPaint: (self iconNamed: #helpIcon)) repeat.
+		(can setPaint: (self iconNamed: #help)) repeat.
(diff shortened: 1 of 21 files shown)


MarcusDenker (approved):

Ducasse:
Just a question why do we rewrite the icon name? To me the rewrite that @estebanlm did was not syntactical but semantics.
and there it made sense. Now I do not get the point of just rewriting addIcon into add

MarcusDenker:
The code reads like that:

```
iconNamed: aSymbol ifNone: aBlock

	self icons at: aSymbol asSymbol ifPresent: [ :icon | ^ icon scaledToSize: icon extent * (self currentWorld displayScaleFactor / self scale) ].
	"Trying the old way"

	
	((aSymbol endsWith: 'Icon') or: [ (aSymbol endsWith: 'Form') ]) ifTrue: [
		self icons
			at: (aSymbol allButLast: 4) asSymbol
			ifPresent: [ :icon | 
				('Using old icon name, please rename ', aSymbol printString) traceCr.
				^ icon scaledToSize: icon extent * (self currentWorld displayScaleFactor / self scale) ]
	].

	^ aBlock value
```

I understood the "Trying the old way" as being fallback code. And doing an " (aSymbol allButLast: 4) asSymbol" for every icon access seems very strange.

Maybe the Icons should be named "somethingIcon", but they are not... the "icons" dictionary has the names without the Form/Icons part.

PR #15278, opened 2023-11-08
Packages: Remove #definedSelectors cache

jecisc:
#definedSelectors cache is a dictionary of the classes defined in a package and the method selectors defined in those classes. 

This is used because to know the defined selectors of a class you need to reject the methods from traits and the extension methods and #isExtension was way too slow. 

Recently I did a huge speed up of #isExtension so we are now able to remove this cache and simplify the code! I also did a speed if of #includesClass: 

This PR just deal with the removal of the cache but we will be able to also remove more mecanism after this change. I'm just trying to do it step by step to not break the system. This will also avoid any synchronization missmatch between RPackage and Classes

### DIFF
--- a/src/Calypso-SystemQueries/RPackage.extension.st
+++ b/src/Calypso-SystemQueries/RPackage.extension.st
@@ -21,20 +21,13 @@ RPackage class >> defaultSortFunctionForCalypso [
 { #category : '*Calypso-SystemQueries' }
 RPackage >> definesOverridesOf: aMethod [
 
-	(self definesOverridesOf: aMethod in: definedSelectors) ifTrue: [ ^ true ].
-	(self definesOverridesOf: aMethod in: extensionSelectors) ifTrue: [ ^ true ].
-
-	^ false
-]
-
-{ #category : '*Calypso-SystemQueries' }
-RPackage >> definesOverridesOf: aMethod in: classAndSelectors [
-
 	| methodClass selector |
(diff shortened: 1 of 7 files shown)


jecisc:
There must be something not consistant somewhere because the dependencies tests are failing. :/

Is this due to a missmatch between RPackage defined selectors and what we actually have in the image? I need to investigate

jecisc:
Of course, those tests are working locally.. So maybe it's some tests that are messing things?

jecisc:
So, I finally can reproduce. This happens when we run the tests of CoverageCollectorTest before the SystemDependencies tests. I need to find what is happening there

jecisc:
This is CoverageCollectorTest>>#testExample that is messing things up

jecisc:
So, the test is removing the #extensionPackage property of all method of Point and Rectangle so I guess something with reflectivity does not preserve the properties of methods

jecisc:
This one is now ready for a review

jecisc:
Marking as priority high because it blocks multiple other cleanings

Ducasse:
Cyril there are failing tests (23 new ones related to traits-v2).
So I will not integrate it for now.

MarcusDenker (approved):

Ducasse:
@MarcusDenker did you see that there were failing tests?

jecisc:
They were known random failures I am starting to investigate

PR #15279, opened 2023-11-08
OpalCompiler: Add more robustness to the compiler plugins system

jecisc:
I am trying to simplify the method compilation and installation process in Pharo and I found some hacks I tired to remove.

I currently cannot remove them because of the plugin system of Opal. What happens is that I am loading plugins in Pharo, but they override the #compiler method to add themselves as plugin. Since they are not fully loaded this causes a failure. 

I propose to fix this by catching the error that could happen during a plugin execution and just log that an error happened and ignore the plugin. (We should probably use a warning but that will be a future improvement)

### DIFF
--- a/src/OpalCompiler-Core/OCDynamicASTCompilerPlugin.class.st
+++ b/src/OpalCompiler-Core/OCDynamicASTCompilerPlugin.class.st
@@ -61,7 +61,7 @@ OCDynamicASTCompilerPlugin >> priority: anObject [
 { #category : 'transforming' }
 OCDynamicASTCompilerPlugin >> transform: ast [
 
-	^ transformBlock value: ast copy
+	^ transformBlock value: ast
 ]
 
 { #category : 'accessing' }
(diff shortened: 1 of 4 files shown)


dionisiydk:
> I currently cannot remove them because of the plugin system of Opal. What happens is that I am loading plugins in Pharo, but they override the #compiler method to add themselves as plugin. Since they are not fully loaded this causes a failure.

Maybe you can improve the plugins baseline which you are trying to load. Post-load action could safely activate the plugins.
Muting errors is not really useful when you want to debug real issues

jecisc:
You mean something like

```st
MyPluggin class >> compiler
    ^ ClassVariableToSetToTrueInPostLoad ifTrue: [ super compiler addPluggin: self ] ifFalse: [ super compiler ]
```

?

I am not sure because them people will need to know that they have to do that. Tbh I have a hard time to find the best solution here

jecisc:
So we discussed with Marcus and we agreed on this:
- Make the plugin super class a no-op instead of asking subclass responsibilities
- Then if your plugin is simple and have only one method everything is good
- If your plugin is complex and has multiple methods you should package it in a dependent package of the classes using it

I'll do another PR today

PR #15281, opened 2023-11-08
Calypso: Fix refresh of project view

jecisc:
This change implements #= on Project so that calypso can find the right project during its refresh.

Also incorporate feedback from Denis

In the future I would like to improve the project manager to not recreate the projects all the time but this will be a future step


jecisc:
Failing tests are unrelated. This is a random failure of the traits tests

dionisiydk (approved):

PR #15284, opened 2023-11-08
Remove workspace

### LINKED ISSUE #15277: Should we keep the Workspace?
The workspace is still present in the image even if we have the playground.

Do we still need it? Can we remove it? Deprecate it?

jecisc:
Seen comments on #15277 it seems that we should remove the workspace so here we go.

Next step is to rename any reference to workspace left to playground to not confuse people that will not know what is a workspace

Fixes #15277


jecisc:
FallbackMenu>>#fallbackMenuExplanations needs to be updated but I do not know how to provide a new binding to the StPlaygroundPresenter

jecisc:
Tests failures are known random failures

jordanmontt (approved):

MarcusDenker (approved):

PR #15285, opened 2023-11-09
Reflectivity: fix extension method recompilation

jecisc:
ReflectiveMethod can recompile the AST and during this the method properties are lost. Which means we are losing the extension packages. 

I added here a quick fix with a test.

But I think we need something better here. First, can't we share a part of the recompilation process in all the places that recompile methods so that we can persist at least some properties in one point? Also, maybe we could have a mecanism to tag some properties so that we conserve them between recompilation? Maybe we could have a collection of properties to persist on the class side of AdditionalState and when we recompile we transfer those?

To discuss with @MarcusDenker


jecisc:
I did a better fix in #15291

PR #15287, opened 2023-11-09
Cleanup metacello

guillep:
[WIP] This PR will probably need some update because it merges/removes some packages and thus the loading script will need to be updated (and I did not touch it yet).

I spent some time cleaning up old metacello code following what @jecisc was doing.
I was a  bit more aggressive, but we can revert any bad change I did.

imagen

Note that everything removed is code that has not been used for years.
I doubt there are projects out there using these features in Pharo (Tutorial, Toolbox).
Anyways, for most of the removals we can load them separately if needed, since code was just removed and remains untouched.

Metacello tests are passing on my machine.
I also tested manually to load/fetch/record XMLParser from https://github.com/pharo-contributions/XML-XMLParser
With the following scripts + cleanup script

```smalltalk
"Cleanup downloads"
#( 'BitmapCharacterSet' 'OrderPreservingDictionary' 'XML-XMLWriter' 'XML-XMLParser' ) do: [ :rn |
	(IceRepository repositoryNamed: rn) ifNotNil: [ :r |
		r unload; forget ] ]

"Fetch"
Metacello new
  baseline: 'XMLParser';
  repository: 'github://pharo-contributions/XML-XMLParser/src';
  fetch.

"Record"
Metacello new
  baseline: 'XMLParser';
  repository: 'github://pharo-contributions/XML-XMLParser/src';
  record.


"Load"
Metacello new
  baseline: 'XMLParser';
  repository: 'github://pharo-contributions/XML-XMLParser/src';
  load.
```


src/Metacello-Base/Metacello.class.st
```diff
@@ -532,11 +527,6 @@ Metacello >> repository: repositoryDescription [
     self addStatement: #'repositoryArg:' args: {repositoryDescription}
 ]
 
-{ #category : 'api options' }
-Metacello >> repositoryOverrides: aRepositoryDescriptionCollection [
```
demarey, line 536: I think I saw usage of repository overrides. As it is the Metacello API, I would not remove it.

src/Metacello-Core/MetacelloScriptEngine.class.st
```diff
@@ -464,12 +456,6 @@ MetacelloScriptEngine >> repositories [
     ^ self projectSpec repositories
 ]
 
-{ #category : 'options' }
-MetacelloScriptEngine >> repositoryOverrides [
```
demarey, line 468: same here, I would keep it

src/Metacello-Core/MetacelloScriptExecutor.class.st
```diff
@@ -233,11 +228,6 @@ MetacelloScriptExecutor >> repositoryArg: anObject [
     repositoryArg := anObject
 ]
 
-{ #category : 'options api' }
-MetacelloScriptExecutor >> repositoryOverrides: aRepositoryDescriptionList [
```
demarey, line 237: same here

jecisc:
As a first test I'll try to load some projects such as Moose to see if everything is alright. 

Then I'll try to read the code changes even if it's huge. 

If everything is alright I'll merge so that we can check how this behave in Pharo :)

jecisc:
I see some classes such as MetacelloSymbolicVersionSpec got removed. Is it the feature to be able to depend on version such as "v1.?" ?

Some people using small projects might depend on that

jecisc (approved):
I checked and the semantic versioning I was talking about seems to be dealt with in another place. 

I think we can integrate and see how it behave. I tried to load moose and a few of my projects without trouble.

demarey (commented):
There are a lot of code I do not know if it is needed or not but @jecisc did some big loads attempts. It should catch most problems

PR #15289, opened 2023-11-09
Opal: Make plugins no-op to not crash during the loading

jecisc:
It is possible that installing a package with an opal plugin and a class using it in the same package crashes because all methods of the plugin are not loaded. 

This change make Opal pluggin superclass a no-op to avoid crashes on simple cases and add a warning in the class comment to warn about this.

### DIFF
--- a/src/OpalCompiler-Core/OCASTCompilerPlugin.class.st
+++ b/src/OpalCompiler-Core/OCASTCompilerPlugin.class.st
@@ -17,6 +17,11 @@ IR
 These plugins are called <<HERE>>, that is, after semantic analysis before generating the IR.
 They are sorted by #priority and handed the AST without making a copy (as plugins might just analyse the AST). If you modify the AST, you have to make a copy before!
 
+## Note on plugin installation
+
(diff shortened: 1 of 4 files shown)


MarcusDenker (approved):

src/OpalCompiler-Core/OCASTCompilerPlugin.class.st
```diff
@@ -17,6 +17,11 @@ IR
 These plugins are called <<HERE>>, that is, after semantic analysis before generating the IR.
 They are sorted by #priority and handed the AST without making a copy (as plugins might just analyse the AST). If you modify the AST, you have to make a copy before!
 
+## Note on plugin installation
+
+Plugins can usually be used by classes by overriding #compiler method and adding the plugin there. But this can lead to problems if you use a plugin that is defined in the same package than the user during the loading of the package in Pharo.
+This is due to the fact that the plugin can be used before its methods are installed. Simple plugins will not have this problem because I work as a no-op. But if a plugin overrides #transform: and call other methods then it could happen. In that case, make sure to have the users of the pluggin in another package and to get your package dependencies right!
```
MarcusDenker, line 23: the error will not happen but of course the compilation will not take the plugin into account on the compilation where the package is loaded (maybe obvious, and we can merge without fixing the comment I think)

PR #15291, opened 2023-11-09
Opal: Add concept of prior methods 

jecisc:
This change adds to Opal the concept of prior method. This allows us to persist in a cleaner way the persistance of the extension package! No need to do it in multiple places in the system anymore.

This is a step toward another goal that is to be able to persist more properties. If this works, then I'll use this mechanism to also persiste the #traitSource property because adding a meta link to a method from a trait is removing this property and then the method loses this info


jecisc:
This also fix a bug where reflexive methods were removing the #extensionPackage property from methods

MarcusDenker (approved):

PR #15293, opened 2023-11-09
CompileMethod: Introduce the concept of properties to persist

jecisc:
Some places in the system are using compiled method properties to add some state to methods.

Those properties are lost when we recompile the method or when a method of the same name get compiled. 

This change introduces the concept of persisted properties and uses it for extension packages and trait source.

This fixes a bug where meta links were removing the properties of methods coming from traits. 

I added a few tests to ensure we will not have regressions in the future


jecisc:
This is working now

MarcusDenker (approved):

PR #15300, opened 2023-11-10
15299-UFFI-should-release-pointers-allocated-for-packing-for-arity

### LINKED ISSUE #15299: UFFI should release pointers allocated for packing for arity
UFFI should release pointers allocated for packing for arity instead of using autoRelease.
Using autoRelease is not needed, as we know exactly when to release them and it also puts less pressure in the finalisation scheme.

tesonep:
UFFI should release pointers allocated for packing for arity instead of using autoRelease.

Using autoRelease is not needed, as we know exactly when to release them and it also puts less pressure in the finalisation scheme. Fix #15299

### DIFF
--- a/src/UnifiedFFI/ExternalAddress.extension.st
+++ b/src/UnifiedFFI/ExternalAddress.extension.st
@@ -273,7 +273,8 @@ ExternalAddress >> packToArity: arity [
 
 	rolledPointer := self.
 	"Start in 2 because first pointer is self"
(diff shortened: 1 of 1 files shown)


jecisc (approved):
Nice one!

jecisc:
Is it planned for backport?

tesonep:
> Is it planned for backport?

I think it is a good idea to backport, I will wait until we are sure that it does not break everything

PR #15305, opened 2023-11-10
Deprecate TRBProgramNodeVisitor

jecisc:
This Trait is not used in Pharo and it seems overkilled to keep it. We discussed at some point with Stephane and Marcus and decided that it should be deprecated.


Ducasse:
we can just remove it. People can load it from P11.

PR #15306, opened 2023-11-10
Prevent undeclared to be associations

jecisc:
Undeclared should only contain UndeclaredVariables but currently it seems that some places (the bootstrap?) are still using association. Here I am trying to force the usage of the variable to find the missbehaving places

### DIFF
--- a/src/System-Support/UndeclaredRegistry.class.st
+++ b/src/System-Support/UndeclaredRegistry.class.st
@@ -10,3 +10,11 @@ Class {
 	#package : 'System-Support',
 	#tag : 'Utilities'
 }
+
+{ #category : 'private' }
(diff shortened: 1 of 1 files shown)


jecisc:
@MarcusDenker With this PR I found out that we had a case were we do not put an undeclared in the undeclareds and this is in Espell:

https://github.com/guillep/espell/blob/master/src/Espell.package/EPGlobalBinding.class/instance/asLiteralInObjectSpace..st

And

https://github.com/guillep/espell/blob/master/src/Espell.package/EPUndeclaredBinding.class/instance/asLiteralInObjectSpace..st

We put a nil there. Since we cannot patch Espell for now maybe we could do a nil management in UndeclaredRegistry?

PR #15307, opened 2023-11-10
rename smalltalk global variable to pharo

### LINKED ISSUE #15213: Rename `Smalltalk` global variable to `Pharo`
There is the `Smalltalk` global variable that references a unique instance of `SmalltalkImage`. It would make sense that we rename it to `Pharo` and the class to `PharoImage`. I think it will be more natural for people.

jordanmontt:
Fixes #15213 

Things done:
- Created a `PharoImage` class that is a copy of `SmalltalkImage`.
- Make `PharoImage` the super class of `SmalltalkImage`.
- Marked `SmalltalkImage` as deprecated and removed all the methods and slots.
- Replaced the references of `SmalltalkImage` with `PharoImage`.

Modifications of the  `PharoImage` class:

- Added class side `initialize` method in `PharoImage`. The initialize method creates a new global variable named `Pharo` with the same value of `Smaltalk` and it marks as deprecated the `Smalltalk` global variable.
- Replaced all the references of `Smalltalk` global variable inside the instance side of the class `PharoImage` with `self`.
- Replaced all the references of `Smalltalk` global variable inside class side the class `PharoImage` with `Pharo` **except** in the `initialize method`.


src/System-Support/SmalltalkImage.class.st
```diff
@@ -1,1932 +1,17 @@
...
-	
-Startup phases
-During the first stage of start up the UI manager, the default uimanager is switched to a specific non interactive ui manager (`StartupUIManager`). Note that this specific non interactive UIManager kills the system on any attempt to open windows and interaction. So be warned, don't use interaction in the first phase. 
-Then all registered classes execute their start up procedures (which should not imply interactive behavior). 
-After startup list is finished, any deferred startup actions are executed, which you can add using the method
-`addDeferredStartupAction:` method.  
-
+Deprecated class. Now you should use `PharoImage`. Look at its class comment.
 "
 Class {
```
jecisc, line 23: Hi Sebastian,

It would be better to remove this class and add on the initialization of PharoImage this:

```st
self deprecatedAliases: { #SmalltalkImage }
```

Like this, methods such as #isKindOf: or #on:do: will keep working.

The only problem is that we need to ensure in the bootstrap that we call this initialization in case the bootstrap still refers to this class. But in anycase we need to replace the parts of the bootstrap using this

jecisc (changes requested):

jordanmontt:
So, this is breaking the bootstrapping, as expected...

(Edited)

src/System-Support/SmalltalkImage.class.st
jordanmontt: Done

jecisc:
IMO it would be simpler to split this into two changes.

The first one to rename SmalltalkImage and keep it as a deprecated alias. The second one to rename Smalltalk into Pharo.

It would help us find what is the problem in the bootstrap

jecisc:
After discussing with Marcus and Sebastian we will probably delay this

PR #15310, opened 2023-11-10
Introduce a method to get the meta model announcer

jecisc:
This change is the first of a big list of change toward two goals:
- Split system announcer into two parts: One to announce code changes (MethodAdded, ClassAdded, ProtocolAdded, PackageAdded, ...) and tooling changes (breakpoint, ast cache, chunks...)
- Do not have global announcer but environement announcers

I am not sure about the name so if anyone has suggestions go for it! It can be improved later of course. The goal here is just to start reducing the usage of the singleton

### DIFF
--- a/src/Kernel/Behavior.class.st
+++ b/src/Kernel/Behavior.class.st
@@ -543,6 +543,12 @@ Behavior >> cleanUp: aggressive [
 	^self cleanUp
 ]
 
+{ #category : 'accessing' }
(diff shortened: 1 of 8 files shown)


jecisc:
I renamed it #codeChangeAnnouncer after discussing with Pavel

Ducasse:
Sounds a good name.

Ducasse:
Less global direct access is always nice.

PR #15311, opened 2023-11-10
Metacello: MetacelloBaseConfiguration as a deprecated alias

jecisc:
This class is kept only for compatibility. I propose to deprecate it via a deprecated alias


jecisc:
Done in #15312

PR #15312, opened 2023-11-11
Metacello: New cleaning pass

jecisc:
This change is a new cleaning pass on Metacello.

- Merge Metacello-Core and MEtacello-MC since we only have one backend for Metacello and we will not have another one before removing Metacello fron Pharo
- Merge abstract classes with their only concrete subclasses
- Remove a lot of dead code
- Simplify some code (sometimes we had 5 one line utility methods calling each other with only one sender.. A lot)
- Deprecate the fact that we can give blocks for properties such as #author:, #blessing:, #description:, #timestamp:
- Remove a lot of extension methods on some primitive objects of Pharo (We can still remove more but I'm getting tired of this cleaning)
- Remove MetacelloPharoPlatform
- Rewrite things such as `x ~~ nil ifTrue:` by `x ifNotNil:`

As a bonus update ReEqualNilRule to replace by #isNotNil instead of #notNil

I think the size of the stack to load Metacello project will decrease by a lot :) (Since Guille is analyzing stack size ahah)


Ducasse:
ohhhhhh such a massive amount of work. Thanks Cyril.

PR #15313, opened 2023-11-11
Remove TraitsV2-Compatibility

jecisc:
The trait implementation changed since multiple years in Pharo. We discussed with Marcus and we think it is time to drop the compatibility with the old implementation


jecisc:
Fuel is still relying on the compatibility package

jecisc:
And I see also that some tests are using the compatibility classes as test subject. I'll try to clean all of that later

jecisc:
I'll close for a new version

Issue #22, opened 2017-02-21
Build broken for 64 bits


guillep:
Since the introduction of the Unicode tables, the build for 64 bits got broken

https://travis-ci.org/guillep/pharo-core/builds/203456269

pavel-krivanek:
This is strange because till now I only added the tables to the repository but the initialization itself is still not executed.

guillep:
This is apparently happenning not because of the Unicode tables but because a change in the VM.
We were using latest vm because there was no stable version for 64 bits.

If we see the dates of the working and failing jobs, we see that the problem was introduced between the 17 and the 20 of february.

We should pick a fixed VM from here:

http://files.pharo.org/vm/pharo-spur64/linux/

guillep:
I'm doing a bisect. It looks like the broken VM is the one of the 19 of february. The one from 18/02 works fine.

We should probably
 1 - release the one of 18/02 stable
 2 -  report the issue we have to the VM list

I invoke @estebanlm

guillep:
Sorry, I was mistaken the one of 18/02 is already broken. The one that works is definitively the one from 12/02.

http://files.pharo.org/vm/pharo-spur64/linux/pharo-linux-x86_64itimer-201702121843-530863a.zip

guillep:
This is fixed since some time ago

PR #15316, opened 2023-11-12
Add binary selector alias for #newFromArray: for Collection

Rinzwind:
This pull request adds a binary selector as an alias for `#newFromArray:` on Collection’s class as suggested in [pull request #22 in the Pharo Enhancement Proposals repository](https://github.com/pharo-project/pheps/pull/22). I used `#%>` as suggested in a [comment by @Ducasse](https://github.com/pharo-project/pheps/pull/22#issuecomment-1802010966). Other options I considered were `#<-` and `#<=`, which look like leftwards pointing arrows, to reflect that in `Set<-{1. 2}` the elements on the right are put into a collection of the class on the left. I implemented the alias only on Collection’s class, even though `#newFromArray:` is also implemented on Object’s class, as a [comment in pull request #11716](https://github.com/pharo-project/pharo/pull/11716#discussion_r979675710) explains the latter only exists for backwards compatibility.

### DIFF
--- a/src/Collections-Abstract-Tests/TCreationWithTest.trait.st
+++ b/src/Collections-Abstract-Tests/TCreationWithTest.trait.st
@@ -30,6 +30,18 @@ TCreationWithTest >> test0FixtureCreationWithTest [
 ]
 
 { #category : 'tests - creation' }
+TCreationWithTest >> testLessThanMinus [
+
+	| aCol collection |
+	collection := self collectionMoreThan5Elements asArray.
+	aCol := self collectionClass <- collection.
+
+	collection do: [ :each | self assert: (aCol occurrencesOf: each ) equals: ( collection occurrencesOf: each ) ].
+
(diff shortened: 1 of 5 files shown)


Ducasse:
Thanks!

jecisc:
Personally I like <- the best

Ducasse:
<- is for association so I would not use this one. we could use >-
I think that we should use %AnotherChar because it is a space that we do not use often.

jecisc:
Association is using #->

#<- has no sender in the image currently :)

Ducasse:
I ws confused now in old smalltalk it was assignment but it could work telling that the data will flow inside the collection especially if it has no sender.

Rinzwind:
I renamed `#%>` to `#<-`.

Ducasse:
Thanks a lot kris. We are fighting some daemons today :)

PR #15317, opened 2023-11-12
Packages: Cleanup after recent changes

jecisc:
This PR has for goal to cleanup a little after the recent changes.

Let's see if we still need some of the bootstrap hacks.

I also simplified some code by removing methods from old package in #addMethod: and asking the current package of a method directly instead of going through the protocol in a place


jecisc:
This was not passing as I expected. I'll close

PR #15319, opened 2023-11-12
Rename TraitsV2 into Traits

### LINKED ISSUE #15298: Rename TraitsV2 into Traits ?
We often integrate some projects with some "New" or "V2" in the name but over the years it's not new or nobody uses the V1 anymore.

I propose to rename TraitsV2 into Traits because I don't think anybody uses a V1 anymore

jecisc:
Fixes #15298


jecisc:
Failures are not related

Issue #15169, opened 2023-10-29
explicitRequirement causes significant slowdown


JanBliznicenko:
Using `self explicitRequirement` can be quite slow compared to, for example, calling super.

It might not be noticable and barely measurable if it is called sporadically or if the actual implementation of the method is quite slow anyway, but if such method with `explicitRequirement` is called many times in a loop and the actual implementation of the method is quite fast, it can make incredible difference.

I found this out because Roassal started using `explicitRequirement` in many traits, and it caused mouse event handling to take 3 times longer in some cases (https://github.com/pharo-graphics/Roassal/issues/10).

Example:
```Smalltalk
actualExampleCode := 'example ^ ''hello'''.
[
  traitExplicit := Trait named: #TExplicit.
  traitExplicit compile: 'example ^ self explicitRequirement'.
  superclassExplicit := Object subclass: #SuperclassExplicit.
  superclassExplicit compile: actualExampleCode.
  subclassExplicit := superclassExplicit subclass: #SubclassExplicit uses: traitExplicit.
  instanceExplicit := subclassExplicit new.
  
  traitSuper := Trait named: #TSuper.
  traitSuper compile: 'example ^ super example'.
  superclassSuper := Object subclass: #SuperclassSuper.
  superclassSuper compile: actualExampleCode.
  subclassSuper := superclassSuper subclass: #SubclassSuper uses: traitSuper.
  instanceSuper := subclassSuper new.
] on: Deprecation do: #resume.

Smalltalk garbageCollect.
explicitRequirementsTime := [ 1000000 timesRepeat: [ instanceExplicit example ] ] timeToRunWithoutGC.

Smalltalk garbageCollect.
superTime := [ 1000000 timesRepeat: [ instanceSuper example ] ] timeToRunWithoutGC.

{	 
  'explicitRequirements' -> explicitRequirementsTime.
  'super' -> superTime
} inspect.
```

In this example, using explicitRequirements takes, on my machine, over 100x longer. If the code of the `example` method takes more time, like `example ^ DateAndTime now - DateAndTime tomorrow`, the difference is less extreme, but it is still there (2 sec vs 3 sec on my PC).

![obrazek](https://github.com/pharo-project/pharo/assets/9571633/4cfff257-3135-44d9-beb1-d153ae190e0c)

Pharo 11 and 12 (build.973.sha.c0bd11f), measured in 12,
Windows 11,
VM v10.0.0.8

MarcusDenker:
[...]

PR #15322, opened 2023-11-12
Speedup #explicitRequirement

jecisc:
This is a first iteration for the speedup of #explicitRequirement.

It simplifies the execution of the method to use #perform:withArguments:inSuperclass: which is faster.

It also unify the behavior between traits and classes.

This is a work for #15169 but this is still too slow

Here is an example of the perfs:

```st
actualExampleCode := 'example ^ ''hello'''.

traitExplicit := (Trait << #TExplicit) install.
traitExplicit compile: 'example ^ self explicitRequirement'.
superclassExplicit := (Object << #SuperclassExplicit) install.
superclassExplicit compile: actualExampleCode.
subclassExplicit := ((superclassExplicit << #SubclassExplicit)
	                     traits: traitExplicit;
	                     yourself) install.
instanceExplicit := subclassExplicit new.

traitSuper := (Trait << #TSuper) install.
traitSuper compile: 'example ^ super example'.
superclassSuper := (Object << #SuperclassSuper) install.
superclassSuper compile: actualExampleCode.
subclassSuper := ((superclassSuper << #SubclassSuper)
	                  traits: traitSuper;
	                  yourself) install.
instanceSuper := subclassSuper new.

Smalltalk garbageCollect.
explicitRequirementsTime := [ 1000000 timesRepeat: [ instanceExplicit example ] ] timeToRunWithoutGC.

Smalltalk garbageCollect.
superTime := [ 1000000 timesRepeat: [ instanceSuper example ] ] timeToRunWithoutGC.

{
	('explicitRequirements' -> explicitRequirementsTime).
	('super' -> superTime) }.

"Before: {'explicitRequirements'->414. 'super'->2}"
"After: {'explicitRequirements'->293. 'super'->2}"

"And here another one with a method that does not just return a constant."
actualExampleCode := 'example ^ (self packageOrganizer packageNamed: ''Kernel'')'.
"Beofre: {'explicitRequirements'->723. 'super'->198}"
"After: {'explicitRequirements'->540. 'super'->196}"	
```

### DIFF
--- a/src/Kernel/Behavior.class.st
+++ b/src/Kernel/Behavior.class.st
@@ -606,6 +606,7 @@ Behavior >> definedSelectors [
 
 	^ self definedMethods collect: [ :method | method selector ]
 ]
+
 { #category : 'accessing - instances and variables' }
 Behavior >> definedVariables [
 	"return all the Variables defined"
--- a/src/Kernel/Context.class.st
+++ b/src/Kernel/Context.class.st
@@ -294,7 +294,7 @@ Context >> activeOuterContext [
 Context >> arguments [
 	"returns the arguments of a message invocation"
 	<reflection: 'Stack Manipulation - Context'>
-	^(1 to: self numArgs) collect: [:i | self tempAt: i]
+	^(1 to: self numArgs) collect: [:i | self basicAt: i]
 ]
 
 { #category : 'closure support' }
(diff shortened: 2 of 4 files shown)


Ducasse:
As I told you I would remove support for explicitRequirement because it is static and it will be always clunky.

jecisc:
This one is a simple improvement until we do better because we cannot just remove the selector from the image since lot of traits are using it

Ducasse:
They are just declaring something not using it so normally it should not have impact. This is like removing a subclassresponsibility. But step by step let us integrate this guy.

PR #15324, opened 2023-11-12
Manifest builder should not rely on categories

jecisc:
I also deprecated a method to use real packages instead


Ducasse:
The question I asked myseflf was why do we need a different manifestbuilder? Because we could just ask the package of any classes? May be I missed something.

PR #15343, opened 2023-11-13
Fix bug in remake of anonymous classes

jecisc:
In case an anonymous class had to be remake because of a change on a non anonymous class (like updating a trait it uses), the anonymous class was becoming non anonymous :( This fixes it and adds a regression test.

I also added some errors on UndefinedPackage to raise exceptions when we try to create a new tag because it should not have tags other than the undefined tag


Ducasse:
I'm wondering why Class themselves could not know the ano and normal classBuilder. I do not really get why we should ask the namespace. To me it is strange.

PR #15350, opened 2023-11-13
Monticello: Introduce Monticello-Model

jecisc:
I would like to split the beast package Monticello that contains too many things. 

Here is a first step by introducing Monticello-Model and moving the definitions there. 

Long term goal is to load the things that are not needed by the bootstrap such as the exporters later in the bootstrap process.


Ducasse:
I think that it would be good to push location extension to ZnUrl.
Less extension is always better.

PR #15351, opened 2023-11-13
Monticello: Remove dead code

jecisc:
Here is a list of changes:

- Remove unused topological sort of ancestors
- Clean MCCacheRepository (2 unused vars and some methods and removed MCPackageCache with it)
- Remove unused backport feature
- Remove MCConfiguration with its reader, writer and browser because it seems really legacy and an ancestor to Metacello
- Remove unused MCFilteredVersionSorter
- Remove MCD format that seems to be a format to export diffs but this seems unused also
- Remove unused version inspector
- Remove MCSubDicrectoryRepository that seems unused
- Merge some classes together (only one was used)
- Remove MCVersionHistoryBrowser that seems unused

This one goes with this change: https://github.com/pharo-spec/NewTools/pull/610


jecisc:
Lot of code is used only by Gofer.. We'll need to clean this one aswell

MarcusDenker (approved):

PR #15358, opened 2023-11-14
Clean AST-Core

### LINKED ISSUE #15348: AST-Core depends on ClassParser
It seems odd to me that AST-Core depends on ClassParser. For example the method RBCascadeNode>>#handleMetaclassName: or RBMessageNode>>#handleMetaclassName:

We should cleanup this bad dependency

jecisc:
- Do not refere to #globals in #statistic
- Remove dead code #repairCollection and co that are calling non implemented code 
- Move #parsePragmaAndReturn from Refactoring to AST-Core since this is called from code of AST-Core
- Move #handleMetaclassName: to ClassParser
- Update manually resolved dependencies

Fixes #15348


jecisc:
Failures are unrelated and fixed by https://github.com/pharo-project/pharo/pull/15353

MarcusDenker (approved):

PR #15359, opened 2023-11-14
Remove last references to RPackage

jecisc:
- Rename RPackage into Package
- Rename RPackageOrganizer into PackageOrganizer
- Rename RPackageTag into PackageTag
- Rename RPackageConflictError into PackageConflictError

I did not rename RPackageSet because I plan to deprecate this one :)

I also didn't change the tests yet because this PR is already big enough and I did not renamed the packages because I'll update the packaging later

Edit: Now this changes only removes the last references to RPackage


astares:
GREAT !!!

jecisc:
This does not pass :( I'll split it into smaller PR because it's hard to find the origin of the problem with the current logs

jecisc:
I'm trying to rename RPackage into Package and I got a problem.

I am sure that this PR not have any reference to RPackage anymore. In the image, bootstrap scripts and bootstrap pharo code. But when I'm trying to bootstrap, it fails. Right after espell did his magic to load the minimal Pharo code and before Hermes comes to place, we launch one the image without any argument. There is a comment to say that this is because this is needed to enable the command lines. And this step is timing out. I tried to put logs in the startup action but apparently we do not even get there. There is just nothing happening.

I checking in Espell and there is no reference to RPackage. There is also no reference to RPackage in the vm code. So I'm kinda at a loss here. I don't know how renaming this class can break in this way. 

As a workaround I removed the use of a deprecated alias to keep a RPackage deprecated subclass for now : ( But that means we cannot remove it until we find the origin

jecisc:
I think I found the problem :)

jecisc:
Now tests are passing!

PR #15364, opened 2023-11-14
CompiledMethod: Preserve protocol after removal

jecisc:
The protocol is not known by methods, instead it is looked up in the class. This means that when we remove a method we cannot access the protocol anymore. This forced us to keep the protocol in the removed announcements and some code listening to the announcement had to deal with this extra info.

Here I propose an alternative: save the protocol in the properties of a method upon removal and like this we can still access the protocol


jecisc:
To be fair here idk what to do really about the objects as methods. They require to have too many API that should not be there... I don't want to add all those methods : (

Should we really keep those 2 classes?

Ducasse:
When I see how slow run:with:in: is I would deprecate it.

jecisc:
After discussing with Marcus we decided to remove the ObjectsAsMethod tests since they are more a bother than anything else. Since reflectivity uses the method, this is covered by it. When Reflectivity will use the method wrapper of guille, we can get rid of this mecanism

jecisc:
I had to add more exceptions to testNoSentButNotImplementedSelectors because some methods had senders, but not the right one. It was just another random method with the same name but it cannot be called in those methods

MarcusDenker (approved):

PR #15379, opened 2023-11-15
Merge MonticelloGUI and Monticello-GUI-Diff

jecisc:
Because they depend on each other and there is no reason to separate them


guillep:
My fix here seems to solve the issue: https://github.com/pharo-project/pharo/pull/15385

PR #15380, opened 2023-11-15
Some cleanup of CodeExport

jecisc:
- Opal should not depend on CodeExport
- Move export code of ChangeSet to CodeExport
- Move loading of CodeExport to BaselineOfBasicTool instead of the bootstrap (Was loaded by Hermes before)
- Merge CodeExport-Traits into CodeExport because it is now loaded after traits

I think the coolest part is to have two less packages loaded by Hermes :)

### DIFF
--- a/src/CodeExport/ChangeSet.extension.st
+++ b/src/CodeExport/ChangeSet.extension.st
@@ -1,0 +1,182 @@
+Extension { #name : 'ChangeSet' }
+
+{ #category : '*CodeExport' }
+ChangeSet >> fileOut [
+	"File out the receiver, to a file whose name is a function of the
+	change-set name and a unique numeric tag."
+
+	| fileReference dir |
+	dir := self class promptForDefaultChangeSetDirectoryIfNecessary.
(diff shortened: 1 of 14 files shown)


src/CodeExport/ChangeSet.extension.st
```diff
@@ -0,0 +1,182 @@
...
+	(aString isNotNil and: [ aString notEmpty ])
+		ifFalse: [ ^ self ].
+	stream nextChunkPut: aString.	"surroundedBySingleQuotes"
+	stream
+		cr;
+		cr
+]
+
+{ #category : '*CodeExport' }
+ChangeSet class >> promptForDefaultChangeSetDirectoryIfNecessary [
```
Ducasse, line 173: We will have to kill this in the future. No popup.

jecisc: This is just used by the file out feature of Calypso and the Snapshot browser of Monticello. I think we do not care about popups there. But it should use Spec and not the UIManager

Ducasse: Yes

PR #15384, opened 2023-11-16
Workaround FFI crash

jecisc:
Here is a proposition of workaround the FFI crash while we fix it. This should be only temporary


jecisc:
Fixed by Guille

astares:
For the records: fixed in https://github.com/pharo-project/pharo/commit/df14793275f514d0bfe7a73a7e4accfd9d4705cd

PR #15394, opened 2023-11-16
Migrate rename class to driver

Ducasse:
Nice pair programming session with caro.


jecisc:
Bootstrap is failing with:

```
Error: Package Calypso-SystemTools-Core depends on the following classes:
  SycRenameClassCommand
  SycRenameClassCommand>>#browserContextMenuActivation
  SycRenameClassCommand>>#browserShortcutActivation
  SycRenameClassCommand>>#sourceCodeMenuActivation
  SycRenameClassCommand>>#sourceCodeShortcutActivation
```

Ducasse:
Why the bootstrap is involved into that?

Ducasse:
May be we lost this extensions.

jecisc:
> Why the bootstrap is involved into that?

Here this is the full image load. It's possible that the load order of packages is not right

guillep:
What @jecisc calls the bootstrap is the build of the image. Probably in the code you changed you changed some dependencies, so the Baselines should be adjusted

Ducasse:
I just remove a class so the extensions to this class should probably be removed too. I will check.

jecisc:
I have the impression that we again have a bug on the extension removal :( I'll investigate. I added a test about the last thing that broke it, but maybe another mechanism is involved

Ducasse:
What I do not understand is:
- we copied (using the copy class) the class SycRenameClassCommand -> SycRenameClassCommand2
- This copy got the same extensions than the original.
- After working we just remove the class SycRenameClassCommand
- The extensions were removed. 
So everything looks correct to me. 
I checked SycRenameClassCommand2 has the extensions that were needed.

Now SycRenameClassCommand is making a reference to the class RBRenameClassDriver. 
However the dependencies to this package are not new because we are doing it all the time. we have around 13 reference from Syc*Command to Driver classes. 
So I do not really get what could be problem.

Ducasse:
Ah I see and this is the really the problem.
When I removed the class SycRenameClass the extensions where not remove. 

So the package of these extension still wants to added method to the removed class.

```
Error: Package Calypso-SystemTools-Core depends on the following classes:
  SycRenameClassCommand
  SycRenameClassCommand>>#browserContextMenuActivation
```

Ducasse:
I will add back an empty class to see.

Ducasse:
What is really confusing is that I do not see the extending method of SycRenameClassCommand in the package Calypso-SystemTools-Core. I only see the ones of SycRenameClassCommand2.

So I will wait and see. Else I will try to rename SycRenameClassCommand2 into SycRenameClassCommand.

Ducasse:
Now what is super super strange is that the consequence of my browsing of the package Calypso-SystemTools-Core did create probably a notification and now I could see the removal of the extension method of the class SycRenameClassRefactoring. So it means that the the triggering of the removal did not happen when I removed the class
but when I browsed the package :).

jecisc:
I might have an idea why this is breaking. It might be due to the fact that RPackage does not use announcement anymore. Maybe before some code of Monticello happened before the update or RPackage and now it happens after and this changes the package of the updated classes and methods? I’ll check

jecisc:
For the bug I opened this: https://github.com/pharo-project/pharo/issues/15401

jecisc:
I proposed a fix here: https://github.com/pharo-project/pharo/pull/15403

:)

Ducasse:
Thanks!!! I'm happy that you could understand what was happeining

PR #15397, opened 2023-11-16
Bootstrap: Remove pre Spur code

jecisc:
In the Pharo Bootstrap we have classes to manage pre spur code and post spur code. But we only build Spur images now so I propose to simplify the bootstrap by removing the pre spur code. 

This allows us to not have to maintain some code we are never using.


Ducasse:
I have no idea what to say about these changes. So it is certainly ok.

PR #15398, opened 2023-11-16
Do not log recompilation in the source file

jecisc:
If you update your class, adding a slot for example, the methods of the class are migrated to the new instance through recompilation. This action is currently logged in the source file I have the impression and this seems unecessary to me.

Let's try to disable that to see if it works


MarcusDenker (approved):

MarcusDenker (approved):
indeed, especially as we set the sourcePointer to the old one, so the logged sources are never used

PR #15403, opened 2023-11-17
Simplify MCWorkingCopy announcement listening

### LINKED ISSUE #15401: Removing a class does not show removed methods in Iceberg
If you remove a class that has extensions on it, the packages bringing the extensions are not marked as dirty in Iceberg.

I started to investigate and I found that:
- The methods are removed from the extending RPackages
- The working copy of the extension packages in Monticello are getting marked as dirty

So the problem seems to come from the Iceberg code I guess? I need to check how Iceberg is updating its dirty state

jecisc:
Since a while (and checking the code it is probably since a loooong time), if you remove a class with extensions, the extension packages are not marked as dirty in Iceberg.

The origin was that ClassRemoved>>#packagesAffected did not returned the extending packages. I'm fixing this here and adding a regression test. 

I also simplified the code of Monticello. The code to all classes and methods announcement can be factorized into one method now that the announcements got fixed

Fixes #15401


guillep:
oh good catch!

MarcusDenker:
```
MetacelloNotification: Loading -> Iceberg-Libgit-CompatibleUserName.1698954183 --- https://github.com/pharo-vcs/Iceberg.git[dev-2.0] --- cache
Error: You are about to load new versions of the following packages
that have unsaved changes in the image:

  Iceberg-Libgit

```

jecisc:
I think I fixed all bugs preventing this cleaning to work now

MarcusDenker (approved):

PR #15404, opened 2023-11-17
ExternalAddress: Avoid temporary instance of ByteArray

tinchodias:
This change showed 2x speedup on my MacbookPro 2018.

Revealed by illimani-memory-profiler on Bloc benchmarks. See: https://github.com/pharo-graphics/Bloc/issues/322

@tesonep do you agree?

### DIFF
--- a/src/FFI-Kernel/ExternalAddress.class.st
+++ b/src/FFI-Kernel/ExternalAddress.class.st
@@ -194,15 +194,18 @@ ExternalAddress >> fromCString [
 ]
 
 { #category : 'converting' }
-ExternalAddress >> fromInteger: address [
-	"Make me point at address."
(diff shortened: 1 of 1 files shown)


MarcusDenker:
Failing test not related

- [CI] Linux: OSEnvironmentTest fail with "Illegal continuation byte for utf-8 encoding" #13070

tinchodias:
Similar to #15405

tesonep:
It is good, the old one was to support big-endian and little-endian. We don't have big-endian platforms anymore. Maybe putting a comment to say that we only support little-endian

tesonep (changes requested):
I will add a little comment saying that only works in little-endian platforms

tinchodias:
@tesonep please check if 4294a6c is fine

PR #15405, opened 2023-11-17
OSSDL2Driver loop: Avoid creating a new Delay every 5ms

tinchodias:
Revealed by illimani-memory-profiler on Bloc benchmarks. See: https://github.com/pharo-graphics/Bloc/issues/323

### DIFF
--- a/src/OSWindow-SDL2/OSSDL2Driver.class.st
+++ b/src/OSWindow-SDL2/OSSDL2Driver.class.st
@@ -207,16 +207,15 @@ OSSDL2Driver >> eventLoopProcessWithVMWindow [
 
 { #category : 'events-processing' }
 OSSDL2Driver >> eventLoopProcessWithoutPlugin [
-	| event session |
+
+	| event delay |
 	event := SDL_Event new.
-	session := Smalltalk session.
+	delay := Delay forMilliseconds: 5.
 
-	[ session == Smalltalk session]
-	whileTrue: [
+	[ event isNull ] whileFalse: [
 		[ (SDL2 pollEvent: event) > 0 ]
-		whileTrue: [ self processEvent: event ].
-
-		(Delay forMilliseconds: 5) wait ]
+			whileTrue: [ self processEvent: event ].
+		delay wait ]
 ]
 
 { #category : 'initialization' }


jecisc (approved):

tinchodias:
Thanks @jecisc 
BTW I doubt about the number of milliseconds (I just kept same duration). It means checking for mouse moves, etc, at a frequency of 200hz. I suspect 60hz should be enough. I will try locally to lower the rate. It should reduce a bit the CPU% when image is IDLE. @estebanlm opinion or experience?

estebanlm:
1) there were reasons why Delay object needed to be created (maybe they are not valid anymore)
2) 5ms was a random number picked with the incredible and infallible technique known as "the eye" ;) Basically it was a number that was granting a smooth movement without killing too much the CPU. Other numbers may fit better but ultimately, what fits for real is to NOT poll (there is a way to make that, and now that we have threaded and main thread ffi, should be possible to implement it, same way as we do with gtk).

tinchodias:
El vie, 17 nov 2023 a la(s) 12:07, Esteban Lorenzano (
***@***.***) escribió:

>
>    1. there were reasons why Delay object needed to be created (maybe
>    they are not valid anymore)
>
> Ooops @tesonep do you know?

>
>    1. 5ms was a random number picked with the incredible and infallible
>    technique known as "the eye" ;) Basically it was a number that was granting
>    a smooth movement without killing too much the CPU. Other numbers may fit
>    better but ultimately, what fits for real is to NOT poll (there is a way to
>    make that, and now that we have threaded and main thread ffi, should be
>    possible to implement it, same way as we do with gtk).
>
> There is SDL_WaitEvent
https://gamedev.stackexchange.com/questions/189033/should-i-bother-with-sdl-waitevent

Ducasse:
- @estebanlm do you know the reasons? Then in this case we should really add a comment in the method.

Ducasse:
Yes if we can remove polling logic it would be really great.

tinchodias:
I wonder what's the design to implement it. For example, the image has a thread executing Pharo IDE in general, another thread waits on SDL_WaitEvent. There is an event, the SDL thread wakes up, and now in which thread will the event be converted to the OSWindowEvent and dispatched in the World? I guess in the main Pharo IDE thread, how is the new event communicated to it? for sure the threaded ffi has an answer

estebanlm:
In gtk I have this:

```smalltalk
installRunLoop

    'Installing Gtk4 event loop' asBeaconSignal emit.

    context := GMainContext default.
    context acquire
        ifFalse: [ self error: 'I cannot acquire the main context' ].   
    
    runProcess := [
        TFProcessLocalWorker worker: TFMainThreadRunner uniqueInstance.
        [ running ] whileTrue: [
            context iteration: true ].
        context release.   
        'Gtk4 event loop ended' asBeaconSignal emit.
    ]
    forkAt: Processor lowIOPriority
    named: 'Gtk4 event loop'
```

I imagine the usage of `SDL_WaitEvent` would be similar :P 
notice we run the event loop in the main thread (this is mandatory on macs). This forces us to run the vm in worker mode.

Ducasse:
@tesonep can you check that cyril was not too fast integrating this change?
@martin it would be great if you can work on the removal of the poll.

Rinzwind:
I opened a pull request to revert this as it caused an error “This Delay has already been scheduled” to be signaled, and the VM to become unresponsive, when opening an extra world through `OSWindowWorldMorph new open`.

Ducasse:
@tinchodias @tesonep 
Kris found a regression so I will revert it. 
https://github.com/pharo-project/pharo/pull/15553

estebanlm:
Voilà, this was the problem :)On 25 Nov 2023, at 11:53, Kris ***@***.***> wrote:﻿
I opened a pull request to revert this as it caused an error “This Delay has already been scheduled” to be signaled, and the VM to become unresponsive, when opening an extra world through OSWindowWorldMorph new open.

tinchodias:
Oops ok! I want to learn why. I couldn't find the PR or issue entry showing how to see the problem. Could you link it, please?
I suppose it happens randomly when save and quit snapshots in some condition... I don't think it depends on the os platform

Ducasse:
I put the link https://github.com/pharo-project/pharo/pull/15553
You can ask kris @Rinzwind and he will tell you.

Rinzwind:
@tinchodias: as already mentioned, the revert is in pull request #15553; sorry, I forgot to put that in my comment. I get the error “This Delay has already been scheduled” when doing `OSWindowWorldMorph new open` in Pharo 12 build 1123. Build 1124 failed. In Pharo 12 build 1125, the revert was merged, and I don’t get the error anymore. I used VM v10.0.9 on macOS v13.6.1 to check both builds.

PR #15414, opened 2023-11-17
Fix Icerberg extension detection bug

jecisc:
This is a subset of #15403 with just the fix and the test without the cleaning to see if the problem on the PR comes from the cleanings


jecisc:
Failing test is not related, I fixed it in another PR

MarcusDenker (approved):

PR #15417, opened 2023-11-17
Fix flacky testSubClassWithTraitsAfterModificationOfParentSharedPools

jecisc:
TraitSlotScopeTest>>#testSubClassWithTraitsAfterModificationOfParentSharedPools is flacky because it creates a class using a trait and a shared pool from the same packages.

Depending on the order of removal, things can get nasty. This happens when the shared pool get removed first then the trait then the class.

The trait removal triggers a recompilation of the class without the trait. But the sharedpool is still considered in the class even if it got removed. This causes a crash.

I added a workaround to remove the class first but this is just hidding the problem (because I don't have the time for a better fix right now and also I need to discuss with the team to know what to do in this case).

Should we remove the shared pool from its users when we remove it lile we do with traits? Should we order thing when we remove them? Should the trait declare they are removed from their users when we remove them from the system while they still have users? I don't know what to do for now on those subjects


Ducasse:
Cyril I suggest that we log such questions somewhere and that we should organise some meetings to discuss such points.

PR #15424, opened 2023-11-19
Add tests on method announcements and fix a bug in traits announcements

jecisc:
- Add tests on MethodAddition, MethodModified and MethodRemoved announcements
- Add tests on those 3 announcements with traits
- Fix a bug where MethodRemoved was announced for the Trait that had the method but also the for each users of the Trait (This was not consistant with other announcements)
- Remove #isProvidedByTrait because we should never announce anything for methods from traits so it will always be false now that I fixed the bug in traits
- Simplify some code (Some checks were useless and the tests proves it)
- Fix a bug where Traits where announcing changes in method protocol for methods copied in traits users


Ducasse:
```
> Traits-Tests-tonel.1 --- tonel:///builds/workspace/est_and_branch_Pipeline_PR-15424/64/src --- cache
Error: Package Traits-Tests depends on the following classes:
  MethodAnnouncementsTest
  MethodAnnouncementsTest>>#testCompileMethodAnnounceAdditionOnlyInTrait
  MethodAnnouncementsTest>>#testRemoveMethodAnnounceRemovalOnlyInTrait
  MethodAnnouncementsTest>>#testUpdateMethodAnnounceModificationOnlyInTrait
```

jecisc:
So, I found the origin of the problem. #isProvidedByTraits cannot be removed yet because we have more announcements done by Traits that should not be here. I'll try to find them and fix them

Ducasse:
Ok take your laser saber and use the force :)

Ducasse:
I would like to know why a method from a trait should not be announced but we can discuss it.

jecisc:
If you add/update/remove a method from a trait, we announce it because it is actual code update.

But we do not want to announce the same events for each classes or traits using the trait because the fact that we install it in its method dictionary is just an implementation detail. And if you want to react to it, you can react to the original announcement 

In fact, I don’t like this way of doing. This adds a lot of complexity and is not really efficient. Instead of creating copies in each users I would prefer to have it only in the traits and to have a different lookup for traited classes

PR #15438, opened 2023-11-19
Fixes #15437 less debugger open for nothing.

### LINKED ISSUE #15437: Could browsing a mistyped class not show a debugger
- type `StartUpUIManager` in a playground or in a calypso pane and press cmd-B
- We get a debugger with a bad message UndeclaredVariableRead...

Solution catch the error. Showing debuggers all the time is not nice.

Ducasse:
{ #category : 'code evaluation' }
OCDoItSemanticScope >> evaluateDoIt: doItMethod [

	^self receiver withArgs: #() executeMethod: doItMethod
	^ [ self receiver withArgs: #() executeMethod: doItMethod] 
			on: UndeclaredVariableRead
		 	do: [ self receiver ]

### DIFF
--- a/src/OpalCompiler-Core/OCDoItSemanticScope.class.st
+++ b/src/OpalCompiler-Core/OCDoItSemanticScope.class.st
@@ -25,7 +25,9 @@ OCDoItSemanticScope class >> isAbstract [
 { #category : 'code evaluation' }
 OCDoItSemanticScope >> evaluateDoIt: doItMethod [
 
-	^self receiver withArgs: #() executeMethod: doItMethod
+	^ [ self receiver withArgs: #() executeMethod: doItMethod] 
+			on: UndeclaredVariableRead
+		 	do: [ self receiver ]
 ]
 
 { #category : 'testing' }


Ducasse:
There are broken tests and I do not understand why we have tests to check that the user is ANNOYED with a useless debugger.

MarcusDenker:
I think we can do that differently.

There is now an exception raised that has as the default action in interactive mode to open the debugger so people can fix the code.

In the past, this would have just returned "nil".

If we want that, we can change the default action of the Exception to always return nil.

Ducasse:
We cannot "fix the code" where can I fix the code. 
We get a debugger and an ugly stack. 
Let us stop to think that developer can do everything.

MarcusDenker:
I think an idea can be to have an error handling not in evaluateDoIt:, but in the code that is implementing the "browse it". 

We do not want to silence all errors for evaluation in general, as that would lead to very strange effects.

Senders-of and implementors-of can be much more clever in this case as they do not evaluate code but can statically analyse and provide something that might make sense. (they try to get some usable selector from the selected text)

Browse-it with "eval this, then browse class of result" cannot do that, but we can guard for errors better.
The idea is that the error handling that we want here is different than for general execution of code.

Besides Undeclared, you could select code that is syntactically incorrect and do a "browse it", here we would want to then not raise an error, either. 

I close the PR (there is the issue #15437) and will do another PR soon.

Issue #15433, opened 2023-11-19
Bug in Context >> cannotReturn causing VM crash and illegal return


isCzech:
Consider the following examples:

`[[^ 1] on: BlockCannotReturn do: #resume ] fork.  "VM crash"`

`[[true ifTrue: [^ 1]] on: BlockCannotReturn do: #resume ] fork.  "Illegal return from ^1"`

A fix will follow in a PR.

Explanation: In the first example the VM places a #cannotReturn context on top of the [^1] context which at this time has pc past the last bytecode instruction and thus the VM crashes when attempting to return here.
In the second example the pc is past the return, but more bytecodes follow so the return from #cannotReturn context can happily continue, however, this continuing is "illegal" in the sense it should not have happened.
The two examples should consistently cause a walk-back indicating an illegal return attempt.

Ducasse:
THANKS a lot. 
This is nice to have such bombs :)
The VM should not crash in fact

isCzech:
Hi Stephane @Ducasse ,
> The VM should not crash in fact

Indeed! It's been bothering me for a long time.
The fix seems surprisingly simple after all: just nil the pc before returning:
```
cannotReturn: result

	closureOrNil ifNotNil: [ 
		BlockCannotReturn result: result from: self home.
		^self push: pc; pc: nil ].  "backup the pc for the sake of debugging"
	UIManager default
		requestDebuggerOpeningNamed: 'computation has been terminated'
		inContext: thisContext
```
I'll add some comments and send a PR.

Ducasse:
Thanks. This is cool. Any action to crash the vm in a reproducible way is welcomed :)
I got a crash but impossible to reproduce it, sadly.

PR #15443, opened 2023-11-19
Fix inconsistent behavior after BlockCannotReturn exception; fixes issue #15433

isCzech:
Fixes the following examples:

`[[^ 1] on: BlockCannotReturn do: #resume ] fork.  "VM crash"`

`[[true ifTrue: [^ 1]] on: BlockCannotReturn do: #resume ] fork.  "Illegal return from ^1"`

Both examples will consistently cause a walk-back indicating an illegal return attempt.

### DIFF
--- a/src/Kernel-Tests-Extended/ProcessTest.class.st
+++ b/src/Kernel-Tests-Extended/ProcessTest.class.st
@@ -671,6 +671,26 @@ ProcessTest >> testRealActiveProcessFromProcesorShouldIgnoreInstalledEffectivePr
 	self assert: processFromProcessor identicalTo: activeProcess
 ]
 
+{ #category : 'tests - termination' }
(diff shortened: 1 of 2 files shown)


jecisc:
It would be nice to add a test about this problem :)

Ducasse:
Indeed :)

isCzech:
No doubt; I need figure out how, and safely. Aimed for some feedback in the meantime ;)

isCzech:
I guess now the solution is clean and testable. If you remove the `self error` part from #cannotReturn, one assertion fails, if you remove the `pc := nil` bit the other assertion fails and the test won't crash the system.

PS: not sure why two tests failed, seems to be unrelated.

Ducasse:
Thanks! We will check quickly this time :)

MarcusDenker (approved):

PR #15444, opened 2023-11-19
Remove RBFindReplaceDialogWindow and RubFindReplaceService since (1) …

### LINKED ISSUE #15440: RubFindReplaceDialogWindow should be replaced by its Spec version
`RubFindReplaceDialogWindow` is a subclass of Morph!
We should replace with a spec version. 

There is a `SpRubFindReplaceDialog` which is already doing the job. Now `RubFindReplaceDialogWindow` is used by `RubFindReplaceService` which is unused. So we should remove  `RubFindReplaceDialogWindow` and `RubFindReplaceService`.

Ducasse:
…there are unused and (2) there are the counterparts called Sp....

Fix #15440


Ducasse:
```
Finished to run tests of Morphic-Widgets-FastTable-Tests in 63 milliseconds 319 microseconds
Beginning to run tests of ProfStef-Tests
SubscriptOutOfBounds: 1 in an OrderedCollection()
```

DependenciesTest because of different dependency. RubText should be at the level of Spec.
Or this should be an accepted dependency.

Ducasse:
I need help for this one

jecisc:
The problem is that Rubric-SpecFindReplaceDialog is neither in Spec nor Rubric. This is in BaselineOfIDE that is another layer. 

It should be moved somewhere else. Now there is a problem. Currently we have Spec depending on Rubric and Morphic. But with this change, we introduce a dependency from Rubric to Spec which is a cyclic dependency. So this is not so easy to fix.

Ducasse:
I will close it in particular because the SpRubFindReplacePresenter is not really working and it should be improved first. 
Then the dependency can be solved by introducing a hook and setters see https://github.com/pharo-project/pharo/issues/15440

Issue #15395, opened 2023-11-16
RBRemoveClassRefactoring >> preconditionHasNoUsers: should not openBrowserOn:


Ducasse:
Right now and this even in presence of Driver we can get that

``` 
preconditionHasNoUsers: aClassOrTrait

	^ (RBCondition withBlock: [
				(self refactoringAwareUsesOf: aClassOrTrait) isEmpty ])
		  errorMacro:
			  aClassOrTrait name , ' Trait is used.<n>Browse classes using it?';
		  errorBlock: [
			  self openBrowserOn: (self environmentWithUsersOf: aClassOrTrait) ];
		  yourself
```

MarcusDenker:
PR was merged

PR #15479, opened 2023-11-20
15395 - Removes refactoring precondition that checks for classes referenced by their name

carolahp:
## Summary
This PR removes support for checking class referenced by their name before removing a class. Fixes issue #15395 .

## Description
Preconditions in `RBRemoveClassRefactoring` (as stated in the class comment, see screenshot) do not provide support for class referenced by their name.

This PR removes code related to that feature, including tests and classes used for testing.
 
image


Ducasse:
Hi balsa 

to me this is ok. Now what do you think.
It was the idea that the refactorings should not check symbol (the user should do it separately).

balsa-sarenac (approved):
I believe that we already removed this for methods, so it makes sense to be consistent.
Maybe some support for this should exist in the form of post-action that user can perform after the refactoring? I will write it down somewhere for future reference.

jecisc:
I think symbols should not block a refactoring since this is not a direct reference, but I would still like the refactoring to propose me to update the symbols. Because this is what makes refactoring interesting, to be able to do a big change in only a small action

PR #15496, opened 2023-11-21
reduce user of #classPool

MarcusDenker:
reduce low level uses of #classPool, use the class variable API instead (which itself might be improved in the future, the first step is to use the existing API)


jecisc (approved):
Nice! 

I hate when we have 50 names for the same thing :)

PR #15497, opened 2023-11-21
move old class def check to #applyChangesAsClassDefinition

### LINKED ISSUE #15495: Second times we paste a valid method in a class pane we get a syntax error
Second times we paste a method in a class pane we get a syntax error
while the code is totally valid.

MarcusDenker:
fixes #15495


jecisc (changes requested):
I discussed with Marcus and this change is not good because then we will never be able to evaluate the old class definitions.

The probable fix is to use #parseFaultyExpression: instead of #parseExpression

jecisc:
I opened a new PR with another fix. I propose to close this one

Issue #15455, opened 2023-11-19
FileList should be ported to Spec


Ducasse:
Right now it is tighly coupled to Morphic

astares:
Could we have https://github.com/astares/Pharo-Harbour switched from Spec to Spec2 and use
it as a possibly better looking replacement? Ideally in NewTools then,

Screenshots are on the page

astares:
There is also already a new implementation of "NewTools-FileBrowser" by @estebanlm 
which could/should be used as a replacement for FileList in Pharo 12.

For this to work on all platforms we have to get in the PR https://github.com/pharo-project/pharo/pull/15501
to provide a #isHidden method for file references. 

This then could be used for fixing an issue with StFBVisibleFilter to get  "NewTools-FileBrowser-Tests" green for all platforms.
(will provide another PR in NewTools repo for that)

hernanmd:
@astares I am reviewing your implementation, and I think some parts (specially the Wrappers) are really nice and could be merged into the implementation of Esteban. 

It is ok with you if I use the implementation of Harbour for the directory tree?

astares:
@hernanmd yes, license allows it already. Ideally the new File Browser in NewTools will be the future common target/file tool. Would not make sense that anybody builds its own when we can have a single one with all features.

(Harbour is not usable in any recent Pharo anyway as it is based on old Spec)

hernanmd:
Thank you @astares

hernanmd:
I close this issue as the core functionality of the old morphic-based FileList is now ported to Spec 2.
It is accessible by evaluating:

```smalltalk
StFileSystemPresenter open.
```

Feel free to further discuss and open new issues.

PR #15501, opened 2023-11-21
Not possible to ask if a file or folder (via reference) is hidden

### LINKED ISSUE #15499: Not possible to ask if a file or folder (via reference) is hidden
In Pharo 12 we can ask a file reference if it is #isReadable or #isExecutable.

But it is not yet possible to directly ask the file reference if it is hidden, although we already have all 
the support for that in the standard image. The class  DiskDirectoryEntry already uses the necessary stat attributes
and provided a method #isHidden already in a platform independent manner and with tests.

So we should provide a #isHidden for file references using that so that it is possible to directly
write:

```Smalltalk
'/home/user/.bar' asFileReference isHidden 
```

astares:
Fix #15499

### DIFF
--- a/src/FileSystem-Core/AbstractFileReference.class.st
+++ b/src/FileSystem-Core/AbstractFileReference.class.st
@@ -633,6 +633,11 @@ AbstractFileReference >> isFile [
 ]
 
 { #category : 'testing' }
+AbstractFileReference >> isHidden [
+	^ self resolve isHidden
+]
+
+{ #category : 'testing' }
 AbstractFileReference >> isReadable [
 	^ self resolve isReadable
 ]
--- a/src/FileSystem-Core/FileReference.class.st
+++ b/src/FileSystem-Core/FileReference.class.st
@@ -316,6 +316,12 @@ FileReference >> isFile [
 ]
 
 { #category : 'testing' }
+FileReference >> isHidden [
+
+	^ (DiskDirectoryEntry reference: self) isHidden
+]
+
+{ #category : 'testing' }
 FileReference >> isReadable [
 	^ filesystem isReadable: path
 ]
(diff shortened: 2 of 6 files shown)


Ducasse:
Nice

akgrant43 (changes requested):
Looks good, but to be consistent with the other attribute checks I'd suggest changing:

```
FileReference>>isHidden
	^ filesystem isHidden: path attributes: nil
```

astares:
@Ducasse 

Yes, I would like to get this in as it can also help to get the tests green for "NewTools-FileBrowser-Tests" on all platforms. 
At the moment it is green on Windows but failing on Linux/Mac due to an issue in StFBVisibleFilter

This can then be a preparation to fix https://github.com/pharo-project/pharo/issues/15455

akgrant43:
> Looks good, but to be consistent with the other attribute checks I'd suggest changing:
> 
> ```
> FileReference>>isHidden
> 	^ filesystem isHidden: path attributes: nil
> ```

Actually, this change is more than cosmetic, the original suggestion will fail for `FileReference`s that point to a memory file store.

astares:
> filesystem isHidden: path attributes: nil

@akgrant43 
Your suggestion will fail as well for memory file stores - so it is cosmetic only (but nonetheless we can switch to it for consistency reason).

The **file system** in FileSystem>>#isHidden:attributes: **dispatches** the #isHidden:attributes: messages **to** the platform specific **store** and MemoryStore does not provide a suitable implementation for #isHidden:attributes: yet

image

I will fix that and provide additional tests for that. Do not yet integrate.

akgrant43:
Ahh, I should have checked further.  Thanks!

akgrant43 (approved):
Thanks!

PR #15511, opened 2023-11-21
15446 clipboard class should be packaged outside of morphic core

### LINKED ISSUE #15446: Clipboard class should be packaged outside of Morphic-Core
And in addition it should not refer to UIManager

```
chooseRecentClipping

	"Choose by menu from among the recent clippings"

	"Clipboard chooseRecentClipping"

	recent ifNil: [ ^ nil ].
	^ UIManager default
		  chooseFrom: (recent collect: [ :txt |
				   ((txt asString contractTo: 50)
					    copyReplaceAll: String cr
					    with: '\') copyReplaceAll: String tab with: '|' ])
		  values: recent
```

```
clipboardText: aText informing: aString
	UIManager inform: aString.
	^ self clipboardText: aText
```

astares:
- repackage the Clipboard class in a new package "System-Clipboard"
- make it testable with a package "System-Clipboard-Tests" and test class ClipboardTest
- deprecate #chooseRecentClipping on Clipboard class and instance side
- move the UI manager logic to RubTextEditor (as it was the sole sender and is in the UI layer too)
- rename ivar "recent" into "recentClippings" and lazy initialize it
- write tests for the LIFO principle of the remembered clippings as well as for the rotation 
- remove old src files of ExternalClipboard, MacOSClipboard, ... that had been around in src folder although package "System-Clipboard" was not loaded
- now make "System-Clipboard" loadable again in baseline

- this fixes #15446


astares:
Very important note for reviewers of this PR:

 - be aware that the "src" folder of Pharo11 and Pharo12 branch still includes an old subdirectory "System-Clipboard" 
    with legacy code from ancient times (with an ExternalClipboard, MacOSClipboard and other) that is not inside of 
    the Pharo11 and Pharo12 image. 
    
   Looks like it has been forgotten to remove them when Pharo was switched to #OSWindowClipboard.

   These old source files were not loaded, so they stay around and nobody noticed. Now in this PR I us the same 
   package name "System-Clipboard" again to extract Clipboard from Morphic-Core package.
   
   This unfortunately makes the diff in GitHub unusable as it compares the ancient old with the new one and
   not the state in the image with the new one. 
   
   So ideally load the branch into Pharo to review and compare with Iceberg and not the GitHub web diff.

It is also important:
   When manually loading the PR this is not passing bootstrap scripts so you also need to additionally fix the broken class definition of OSWindowClipboard (which points to obsolete class)
 
image

fix it with
```Smalltalk
Clipboard << #OSWindowClipboard
	slots: {};
	tag: 'Base';
	package: 'OSWindow-Core'
```

and then manually evaluate

```Smalltalk
OSWindowClipboard new beDefault 
```
which is typically run during the bootstrap process. Then you have the new code and tests and can review.

-------------------------------------------------------      
image

astares:
BTW: This PR also fixes the "Paste..." menu entry to show the last clipped texts

image

image

astares:
Windows is green, the other two platforms have test trouble on CI only. But this is in StDebuggerTest which is unrelated

image

image

All the debugger tests pass in my Pharo 12 image on LINUX with the PR loaded

image

So this is a CI issue and not related to the PR

jecisc (commented):
Instead of naming the package System-Clipboard I think it might be better to have a package System-ToolingSupport because from the issues of Stephane, other classes will be extracted from Morphic and I don't want to have a lot of new packages with just one class in each package. And most of those classes are used for the tooling support of Pharo.

astares:
@jecisc then lets do that afterwards in another ticket. It was already tough to provide it that way.

Ducasse:
Thanks! this is great to see this happening. Indeed in a second step we can extract from morphic-core the infrastructure that is not Morphic specific (DropHandler and others).

Ducasse:
@astares I know that thales people enhanced a clipboard on windows to be able to pass meta data so that the user can copy an svg from an external tool and get an svg in Pharo for example.

astares:
@Ducasse I do not particulary know about the SVG use case you are talking about.  

What I know and can say is that they somehow build upon my "Pharo-OS-Windows" project and had some Windows specific extensions that initially they managed separately. 

Then they contacted me this year (spring 2023) and asked if I can reintegrate some of these Windows specific stuff as contributions into the "Pharo-OS-Windows" project. This was not done via PR but the code is in there (cleaned, aligned with the project and also covered with tests). 

And yes - this included diffent things but also improvements in Windows Clipboard handling. See https://github.com/astares/Pharo-OS-Windows/commit/fb93d88d23700d1f18f1b32db017dda287dc4171 for details on the Windows clipboard stuff

PR #15514, opened 2023-11-22
Deprecate SnapshotDone announcement

jecisc:
This announcement is useless since we have the SessionManager and just gives duplicated ways to react to the image startup. It is better to keep only one way to react to the image startup


Ducasse:
Is there a registration mechanism in the session manager?

jecisc:
Yes! You can add any class to the registry with a system of priority and then your method can implement the method #startUp: or #shutDown: and it will be called at startup or shut down or image saving

https://github.com/pharo-open-documentation/pharo-wiki/blob/master/General/SessionsManagement.md

Ducasse:
OK I knew and I thought that there was soemthing else.

Ducasse:
I think that the snapshot was probably to  unify and only get announcements.

PR #15518, opened 2023-11-22
Integrate Packages to kernel

jecisc:
Since they are now part of the MM they should be with the other classes of the MM


MarcusDenker (approved):

guillep:
I don't agree with this change.
Otherwise, we should put Array String and Symbol into the kernel package (plus superclasses).

I think this goes in a bad direction.

jecisc:
To me it makes sens that Package lives with Class, Protocol and CompiledMethod. 

If we do not want to have a Kernel that is too big, I would prefer to split it and have for example a package Kernel-CodeModel with all classes I quoted above. We could also have a Kernel-PrimitiveTypes with classes such as Boolean, Number etc (but then should we put String there?)

What do you think about this?

guillep:
It's just we don't have an agreement on what the kernel is.
It's important to notice that we have a reflective kernel, so at some point there are circularities.
And if we aim to remove those circularities *in the kernel* by putting everything in a single package, then we will end up with everything in a kernel package.

I see packages in two ways: as units of loading, and as units of organization.
The first role is useful for loading code and managing dependencies.
The second role is useful to think about the system.
In non-kernel packages (Seaside, Iceberg, Ring, Calypso), packaging has both roles.
In the kernel, the unit of loading is more difficult to achieve due to circularities, and the bootstrap is in charge of dealing with that so we can still have small packages with different concerns in kernel code.

As I said before: If we put Package and protocol, why not Array? I will argue that Array is more kernel than those two.
Array has a strong role at runtime while Package and Protocol are just for organization and completely optional from a runtime point of view.
But then, Array depends on SequenceableCollection, and Collection FTW.
MethodDictionary inherits from Dictionary too.
Should we put them in the kernel?

If we put everything in a single kernel package, we lose the *units of organization* property, and reasoning about the system is more complicated. So let's just have a big image and no packages...

jecisc:
In general I prefer to avoid circular dependencies between packages but there are some exceptions, and the core of Pharo is part of it because grouping everything would create just a huge package impossible to browse.

But I do believe that if Class, Protocol and Method are in the same package, Package should be with them. This is why I think it might be a good idea to organize the kernel depending on the concepts. Or what you call unit of organization. And I think that one good unit of organization could be to groupe what is related to the Pharo code representation such as Behavior, Class, Protocol, CompiledMethod, BlockClosure, Package and the code directly necessary for them such as MethodDictionary. 

What do you think about this unit of organization?

PR #15519, opened 2023-11-22
Rename tests packages

jecisc:
Rename RPackage-Tests into Kernel-Packages-Tests. Also do some renamings in Monticello tests.

I have the impression that some groups and packages are declared in the BaselineOfPharoBootstrap but are not used by the bootstrap.


jecisc:
I'll redo it

PR #15526, opened 2023-11-23
Add String>>#findClosing:startingAt: generalizing #findCloseParenthesisFor:

Gabriel-Darbord:
`findCloseParenthesisFor:` takes nesting into account to find the matching close parenthesis.
This new method generalizes this behavior to any opening and closing characters.
It has the same pitfall: giving an index less than 1 will give an error.
It also has the same return value if no match is found: it returns the size of the collection + 1.
This last point is a bit strange imo, most `find...` methods will return 0 if not found, maybe this should be generalized too?

There is a new edge case: if the same character is both a closing and an opening character, it is currently prioritized as a closing character.
```st
'((' findClosing: $( startingAt: 1. "2"
```
This could be done the other way around, prioritizing as an opening, which would give `3` in this example because the matching closing character is never found.

This new method is faster because:
- it remembers the current character in a variable instead of using `at:` twice
- it uses lazy conditions instead of checking for opening and closing each time
- since we are dealing with characters, we can use identity instead of equality, which is much faster.

```st
string := '(1+(2-3))-3.14159'.
[ string findCloseParenthesisFor: 1 ] benchCompareTo: [ string findClosing: $) startingAt: 1 ] "10784695.783/s * 2.175 = 23454250.950/s"
[ string findCloseParenthesisFor: 4 ] benchCompareTo: [ string findClosing: $) startingAt: 4 ] "21144145.675/s * 1.846 = 39040974.605/s"
```

Note that this method could be in SequenceableCollection instead of String, but then it would be better to use equality.

I also haven't touched `findCloseParenthesisFor:` yet, I'm not sure if it should be deprecated or if it should call this new method.

### DIFF
--- a/src/Collections-Strings-Tests/StringTest.class.st
+++ b/src/Collections-Strings-Tests/StringTest.class.st
@@ -886,15 +886,6 @@ StringTest >> testFindBetweenSubstrings [
 ]
 
 { #category : 'tests' }
-StringTest >> testFindCloseParenthesisFor [
-
-	self assert: ('(1 + 3(2 * 9) - 15)' findCloseParenthesisFor: 1) equals: 19.
-	self assert: ('(1+(2-3))-3.14159' findCloseParenthesisFor: 1) equals: 9.
-	self assert: ('(1+(2-3))-3.14159' findCloseParenthesisFor: 4) equals: 8.
-	self assert: ('()' findCloseParenthesisFor: 10) equals: 3
-]
-
-{ #category : 'tests' }
 StringTest >> testFindClosingStartingAt [
 
 	self assert: ('(1 + 3(2 * 9) - 15)' findClosing: $) startingAt: 1) equals: 19.
--- a/src/Collections-Strings-Tests/StringTest.class.st
+++ b/src/Collections-Strings-Tests/StringTest.class.st
@@ -900,8 +900,8 @@ StringTest >> testFindClosingStartingAt [
 	self assert: ('(1 + 3(2 * 9) - 15)' findClosing: $) startingAt: 1) equals: 19.
 	self assert: ('(1+(2-3))-3.14159' findClosing: $) startingAt: 1) equals: 9.
 	self assert: ('(1+(2-3))-3.14159' findClosing: $) startingAt: 4) equals: 8.
-	self assert: ('()' findClosing: $) startingAt: 10) equals: 3.
-	
+	self assert: ('()' findClosing: $) startingAt: 10) equals: 0.
+
 	self assert: ('List<String>' findClosing: $> startingAt: 5) equals: 12.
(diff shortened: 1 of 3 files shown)


Ducasse:
Hi gabriel

I do not understand your point 

There is a new edge case: if the same character is both a closing and an opening character, it is currently prioritized as a closing character.

```
'((' findClosing: $( startingAt: 1. "2"
```

Ducasse:
Another question would it make sense to remove the less generalized one?

Gabriel-Darbord:
The new edge case is introduced by the fact that the opening character is determined by the character at the given index (the argument given after `startingAt:`).
```st
'((' findClosing: $( startingAt: 1
```
In this example, the first argument is the closing character, and the opening character is the one at the second argument. So they are both the same character: the opening parenthesis `(`.

Currently, when iterating over the string, the implementation first checks if the current character is closing, which decreases the nesting level.
If it matches the closing character, it will not check if it is an opening character, so the nesting will never increase.
Thus, running the example returns `2` because the second parenthesis is considered a closing character.

If it would first check if the current character is an opening character, then the nesting level would only increase and never decrease, and running the example would return `3` (equal to `'((' size + 1`) to indicate that the matching closing character was not found.

Ideally, it should be used so that it never ends up in this particular case, for example:
```st
string := 'Lorem(ipsum)'.
string findClosing: $) startingAt: (string indexOf: $()
```
Nevertheless, it is important to know that it can happen, and to decide how it should behave when it does.

---
> would it make sense to remove the less generalized one?

I don't know for sure if we should remove `findCloseParenthesisFor:`, but there are currently no senders.

jecisc (approved):

oopingEscatologist:
The String class does not speak the language of any particular string, 
so the number of situations in which this method can be expected to
produce correct results is very small. 

The notions of open and close suggest that the string contains not just characters,
but rather a language--one with an assumed alphabet and an assumed but unspecified grammar.
The receiving "string" here is not the correct target--this method belongs in a class
that understands the grammar, so it can find the matching token.

This is true even when the tokens being matched are just characters, as they are here, 
because the other tokens, the tokens not being thought about (which are being skipped over
while looking for opens and closes) might not be represented as single characters--and might even
contain open or close characters, which will accidentally "match" as characters. 

For example, if the "language" of the string is Smalltalk source code, then one needs to know about
the character constant $) which is *not* the same as a closing parenthesis, and about
string constants and/or comments containing parentheses.

And if the language in the string happens to include any markup, say for example HTML,
there will be many other things to consider--tags, and ampersand literals,  and '+' encoded query strings,
and so on.

No matter how this method is written,
it will almost never be the correct thing to use.
And if and when it is correct, it will be in the one case
where you're most likely to write it correctly yourself,
so it didn't really need to be here.

Ducasse:
You are right but this method is already better than the previous one. 
So as an incremental step for Pharo code strings it is better.

PR #15528, opened 2023-11-23
transform timeout: into timeLimit:

Ducasse:


jecisc (approved):
I think we can even remove this one since the default time limit is at 10sec

Ducasse:
OK I did not know so let us close this issue

PR #15530, opened 2023-11-23
Extract Halt from kernel

jecisc:
Halt is a debugging utility. Maybe we can load it later than in the Kernel?


Ducasse:
:)
We should find all the halt left in the system too :)

MarcusDenker:
There is #testMethodsContainNoHalt 

All methods with a halt are tagged with the pragmas #haltOrBreakpointForTesting or #debuggerCompleteToSender

```
testMethodsContainNoHalt

	| methods |
	methods := SystemNavigation new allMethods select: [ :method | method containsHalt ].
	"these methods are using halt for testing something"
	methods := methods reject: [ :method |
		           method hasPragmaNamed: #haltOrBreakpointForTesting ].
	"these methods are implementing halt, we are not interested"
	methods := methods reject: [ :method |
		           method hasPragmaNamed: #debuggerCompleteToSender ].

	"there should be no method left"
	self assert: methods isEmpty description: [
		String streamContents: [ :stream |
			stream
				nextPutAll: 'Found methods with halt';
				print: methods ] ]
```

jecisc:
Tests are failing because some code is checking the methods of a specific protocol

Ducasse:
Looks good to me. I do not get if the calypso failing test is related

jecisc:
Seems like a random failure. I'll restart it

MarcusDenker (approved):

PR #15531, opened 2023-11-23
fix 15529 - Remove unused preconditinos and RBCondition methods

Ducasse:
- remove unused breakingChangePrecondition
- les openBrowser


Ducasse:
One less openBrowserOn: :)

PR #15534, opened 2023-11-23
Remove some SessionManager dependencies from Kernel

jecisc:
This change removes some of the dependencies of SessionManager over the Kernel.

One step is to move out of the Kernel some SessionManager registrations. Another is to move some methods from System-Support to SessionManager package. Another change is to remove the dependency from SystemDictionary to SessionManager by using announcements to update the SessionManager

Next step is to use announcements to update the removal of removed classes and add a way to save and quit without using the session manager


Ducasse:
Strange I cannot see the merge button :(

PR #15541, opened 2023-11-24
Cleanup: reduce users of #classPool

MarcusDenker:
Small refactoring to remove the users of #classPool, use the Reflective API for class variables instead

### DIFF
--- a/src/Gofer-Deprecated/GoferOperationTest.extension.st
+++ b/src/Gofer-Deprecated/GoferOperationTest.extension.st
@@ -103,11 +103,11 @@ GoferOperationTest >> testReinitialize [
 		addClassVarNamed: #ClassSide.
 	class compile: 'initialize InstanceSide := true'.
 	class class compile: 'initialize ClassSide := true'.
-	self assert: (class classPool at: #InstanceSide) isNil.
-	self assert: (class classPool at: #ClassSide) isNil.
+	self assert: (class classVariableNamed: #InstanceSide) read isNil.
+	self assert: (class classVariableNamed: #ClassSide) read isNil.
 	gofer reinitialize.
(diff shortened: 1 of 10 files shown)


src/Gofer-Deprecated/GoferOperationTest.extension.st
```diff
@@ -103,11 +103,11 @@ GoferOperationTest >> testReinitialize [
 		addClassVarNamed: #ClassSide.
 	class compile: 'initialize InstanceSide := true'.
 	class class compile: 'initialize ClassSide := true'.
-	self assert: (class classPool at: #InstanceSide) isNil.
-	self assert: (class classPool at: #ClassSide) isNil.
+	self assert: (class classVariableNamed: #InstanceSide) read isNil.
```
jecisc: Can't #readClassVariableNamed be used instead?

jecisc (approved):

Ducasse:
What is the diff between classVariableNamed and readClassVariable?
This is the first time I hear about readClassVariableNamed: and it is unclear when this one should be used and not classVariableNamed:

jecisc:
From the name I guess one return an object that represents a class variable while the other return the value on the class variable in the class?

MarcusDenker:
- #classVarNamed: is the old API with shortend names, similar to #instVarNamed:. Contrary to the name, it returns the *value*, not the *variable* (as does #instVarNamed:)
- #classVariableNamed: returns the class Variable (subclass of Variable). Else we would need to use "classPool associationAt:", which is bad as it exposes implementation details
- readClassVariableNamed: return the *value* of the variable. (to be used instead of #classVarNamed:).

The idea is to refactor to use the API that we have now (to get rid of all users of #classPool)  and then ask the question if this API we have is good as it is or if we can improve it.

PR #15553, opened 2023-11-25
Revert #15405 to fix error and unresponsive VM when opening extra world

Rinzwind:
This pull request reverts pull request #15405 as it caused an error “This Delay has already been scheduled” to be signaled, and the VM to become unresponsive, when opening an extra world through `OSWindowWorldMorph new open`.


Ducasse:
Thanks. We were wondering indeed if the previous changes was safe and we do not really have tests for it.

PR #15558, opened 2023-11-25
Refactor: remove method driver

balsa-sarenac:
This PR refactors Remove method driver to new driver architecture:

* driver now relies on refactoring to check applicability preconditions
* then, driver gathers breaking changes (methods to be removed shouldn't have senders in this case) and checks them
* if they fail we show user a dialog
* a few preconditions were reified during this transition

This PR is missing reification of last precondition for remove methods, which is `checkSuperMethods`. That one is a bit strange so I've left it for later. I have also created a new task it so we don't forget.


Ducasse:
I looks really nice. This is great to see our architecture getting piece by piece into place.

PR #15560, opened 2023-11-25
Feat: deprecate class driver

balsa-sarenac:
This PR introduces interactive driver for deprecate class family of transformations.
Driver offers user three options:

* Just add `isDeprecated` method to the class side (@MarcusDenker I think you'll like this one :))
* Rename class, add a subclass with old class name and deprecate it
* Deprecate class and migrate references to another class

I tried to clean as much as possible, but there is still some work left. I will get back to this after I clean the API a bit.

New driver can be initalized through:
`Right click on a class -> Refactoring -> Deprecate`

Right now we have duplicate commands that don't use driver. Those are:
`Right click on a class -> Deprecate` (this subclasses and renames the class)
`Right click on a class -> Migrate references` (deprecate and migrate references)
If the driver commands seem good, we can remove these commands.

### DIFF
--- a/src/Refactoring-Core/RBAbstractTransformation.class.st
+++ b/src/Refactoring-Core/RBAbstractTransformation.class.st
@@ -212,6 +212,17 @@ RBAbstractTransformation >> generateChanges [
 	"RBRefactoringManager instance addRefactoring: self"
 ]
 
+{ #category : 'transforming' }
+RBAbstractTransformation >> generateChangesFor: aRefactoring [
+	"I will generate changes and save them in the model, BUT I will not apply them!
+	Use me when a refactorings is composed of multiple other refactorings"
+
+	"Execute the argument but passing the receiver options to that refactoring"
+	aRefactoring copyOptionsFrom: self options.
+	aRefactoring model: self model.
+	aRefactoring generateChanges
+]
+
 { #category : 'accessing' }
 RBAbstractTransformation >> model [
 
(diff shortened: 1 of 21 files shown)


jecisc:
Hi!
Since Pharo 12 it is better to not add deprecated subclasses because they have multiple problems:
- #isKindOf:/on:do: and related methods might not always work the expected ways
- if the class has a special registration mechanism it might not work
- Extension methods will be placed on the subclass and the renamed class will not know them

Instead we can add a sentence to the class initialization you do `self deprecatedAliases: #( #DeprecatedClassName )`


in the future I want to add this to the class definition instead of going through the initialize

balsa-sarenac:
Hey @jecisc,

I now remember that we talked about this back in September. I will look at how `#deprecateAliases:` is used in the image and see if I can easily add it. Thanks for the heads up!

Ducasse:
I would be good to enter an issue to remove a refactoring that does not make sense because it is less work for us. 
And now Balsa worked for nothing.

Ducasse:
So what do we do. 
I'm for removing everything related to class deprecation and if people wants one then they will write one. 
And they will have to follow our architecture.

balsa-sarenac:
To use `deprecatedAliases:` one needs to add a line to the initalize with name of deprecated aliases. I guess this one should be used both for rename and deprecate and for migrate references.
I can work with this and make `RBDeprecateTransformation` can add `deprecatedAliases:` to `initialize`.
Both Rename and Migrate can then be modified to use `RBDeprecateTransformation` to create `deprecatedAliases:` and their own logic?

What do you think of this?

Ducasse:
I can integrate it now and we can add it or not. 
Else we remove them

PR #15573, opened 2023-11-27
remove `CommandLineArguments default`

### LINKED ISSUE #15572: CommandLineArguments are cached and not returning valid value after image save
**Bug description**
When the image accesses the command line arguments, it does it using `CommandLineArguments default`. It is a singleton that caches the value for the lifetime of the image. However, it is never reset! This means that if you run the image with some arguments, save it, and then run the image again with different arguments, the `CommandLineArguments default arguments` will return the obsolete cached array.

`CommandLineArguments new arguments` returns the current value. The `default` message seems to be never actually used and must be removed.

**To Reproduce**
1. run the image with some set of command line arguments
2. evaluate `CommandLineArguments default arguments`
3. save and shut down the image
4. run the image with a different set of command line arguments
5. run `CommandLineArguments default arguments`
6. you will see the old arguments

pavel-krivanek:
fixes #15572

### DIFF
--- a/src/System-BasicCommandLineHandler/CommandLineArguments.class.st
+++ b/src/System-BasicCommandLineHandler/CommandLineArguments.class.st
@@ -12,18 +12,10 @@ Class {
 	#instVars : [
 		'arguments'
 	],
-	#classInstVars : [
-		'singleton'
-	],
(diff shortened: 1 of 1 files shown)


guillep:
Thanks PAvel :)

jecisc (approved):

PR #15577, opened 2023-11-27
remove `Smalltalk vm documentPath`

### LINKED ISSUE #10007: VirtualMachine>>#documentPath is wrong

### LINKED ISSUE #15574: `Smalltalk vm documentPath` is broken
**Bug description**
`Smalltalk vm documentPath` seems to be broken. It looks like it should return the image file, but instead, it usually returns `--interactive` now. 

It is used in `CommandLineArguments>>#initialize` and it seems that the original intention was to pass the image name as the only argument if no other one is available. Even this usage looks quite strange. Those who rely on it should rather use `SmalltalkImage current shortImageName`. I think we should remove the message and the usages.

Related issue:

`CommandLineArguments>>#commandLineArguments` and
`CommandLineArguments>>#initialize` 
do the same thing.

pavel-krivanek:
fixes #15574

### DIFF
--- a/src/System-BasicCommandLineHandler/CommandLineArguments.class.st
+++ b/src/System-BasicCommandLineHandler/CommandLineArguments.class.st
@@ -52,13 +52,7 @@ CommandLineArguments >> arguments [
 CommandLineArguments >> commandLineArguments [
 	"self commandLineArguments"
 
-	| documentPath args |
-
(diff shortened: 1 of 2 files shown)


jecisc:
This is breaking things in the build

PR #15581, opened 2023-11-27
Shuffle tests

guillep:
Add test shuffling during run to guarantee that tests run in any order.
The shuffle seed is configurable in the test suite and as a command line argument when running tests from the command line


tesonep (changes requested):
It will be nice to add the description of the new command line parameter in the Comments of the TestCommandLineHandler class. The comment of the class is used by the help option

tesonep (approved):

jecisc:
I can predict that at least the failing tests of this PR will fail randomly: https://github.com/pharo-project/pharo/pull/13349

:p

tesonep:
The error is related with the change, windows passed because it did not run the tests.... There is another issue there. I think the windows issue is not related to the PR

tesonep:
When running the tests in the small image it fails to initialize the Random seed from the clock. Maybe for the tests in the small image we can provide a seed calculated from outside.

guillep:
> When running the tests in the small image it fails to initialize the Random seed from the clock. Maybe for the tests in the small image we can provide a seed calculated from outside.

Looks like `DoubleWordArray` has a wrong format in the minimal image?

guillep:
imagen

Shuffling tests is here!

jecisc:
If we shuffle test it would be good to generate a way to reproduce a specific order. Like this, if there is a failure that depend on the execution order of tests, we could have a way to reproduce it. Else we might never be able to get in the same situation ever again.

For example by printing the seed and the command line to execute to run tests with this seed?

guillep:
well, that is already in the PR, what do you think is missing?

PR #15583, opened 2023-11-27
Introduce Kernel-CodeModel package

jecisc:
This change splits Kernel package to introduce a Kernel-CodeModel package. 

Kernel is a really big package, the goal is to improve the organisation of Pharo to make it easier to find what we are looking for. Packages, Classes, Methods, Blocks, Protocols and related classes are entities of the same concept and could be packaged together


jecisc:
@guillep I hope this can be a good compromized to not make the kernel package grow too much because of the packages while having the packages close to the classes, methods & co

Ducasse:
Strange I have the same I cannot press on Merge :(

Ducasse:
There are conflicts now.

jecisc:
Solved

jecisc:
@Ducasse Should I merge it if it gets green? (I fixed yet another conflict. It happens a lot since this repackage a lot of classes)

jecisc:
It's hard to handle conflicts every day. I'll fix the current conflict when the PR will be approved/commented or rejected

guillep:
Hi @jecisc I think this change is good, just go on as soon as it is ready ;)

PR #15585, opened 2023-11-27
Move continuation out of the Kernel

jecisc:
Continuations is a class that is not used in the image but might be interesting to have. 

It is currently in Kernel but this is not a right place IMO. I propose to rename System-Clipboard into System-Utilities as it was mentioned in another PR and move the continuations there


Ducasse:
Ok we could also have System-Continuation. Because I would like in System-Support what we use for the System.

Ducasse:
But let us merge it for now.

PR #15586, opened 2023-11-27
Extract InstructionClient from Kernel and Opal

jecisc:
InstructionClient and its subclasses are used for decompilation and debugging. Currently some of the classes are in Kernel and Opal. I propose to move all of this in the Debugging-Core package. This will reduce the amount of code loaded by Espell

### DIFF
--- a/src/BaselineOfGeneralTests/BaselineOfGeneralTests.class.st
+++ b/src/BaselineOfGeneralTests/BaselineOfGeneralTests.class.st
@@ -14,6 +14,7 @@ BaselineOfGeneralTests >> baseline: spec [
 	repository := self packageRepositoryURL.
 	spec for: #'common' do: [
 		spec
+			package: 'Debugging-Core-Tests';
 			package: 'Debugging-Utils-Tests';
 			package: 'NumberParser-Tests';
 			package: 'AST-Core-Tests';
--- a/src/Debugging-Core-Tests/InstructionClientTest.class.st
+++ b/src/Debugging-Core-Tests/InstructionClientTest.class.st
@@ -7,9 +7,8 @@ This is the unit test for the class InstructionClient. Unit tests are a good way
 Class {
 	#name : 'InstructionClientTest',
 	#superclass : 'TestCase',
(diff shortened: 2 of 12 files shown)


MarcusDenker:
But the IRBytecodeDecompiler in Debugging... is that the right place?

jecisc:
I'm not sure. Is this even used? I found only one method other than tests using it and this method has no sender other than test

guillep:
Be careful, there is a dependency in the kernel related to process termination:

imagen

Check the send of `runUntilErrorOrReturnFrom:`

MarcusDenker:
We fall back to the decompiler when there is no source code (something I am not a huge fan of).

The decompiled code then is not only used by the Debugger, but shown in the browser, too, so it is not just a debugging feature.

jecisc:
For now I failed to see the dependency here but I'll check a little more later and also check what those classes are doing more clearly. 

I have the impression that the whole IRBytecodeDecompiler is used only in #decompileIR and this method is used only in 3 tests. Is it worth to keep? I have the impression that it is not but this is really not a part of the system I know well some it's highly possible that I'm missing some background knowledge here.

MarcusDenker:
Ah, FBDDecompiler is a subclass of InstructionClient and does not decompile to IR first.

Decompiling to IR was at some point used to do bytecode level transformations (Javassist style), but I think nobody has used that for a long time.

jecisc:
Maybe we could remove it then? And if an external project uses it, it can be copied outside Pharo?

Ducasse:
And we could have a default that when decompilation is not available just says sorry 'no code visible' because no decompiler. 
Not everything have to be put together.

Now the instructionClient is important for some of the exception functionality. 
Guille cut already a part to decouple it from the Context. 

Now the subclasses such as the IRBytecodeDecompiler is probably something else.

jecisc:
I'll close this one because before checking InstructionClient we discussed about improving the packaging of IRBytecodeDecompiler with Marcus. 

This is probably not good to have this one in Opal, it would have its place into Flashback

PR #15591, opened 2023-11-27
Fixes #15590 always use defined method when extracting a similar body!

### LINKED ISSUE #15590: [RB] option shouldUseExistingMethod: should always be true
So we can remove the option.

S

Ducasse:
For the stupid and useless robot: 

Fixes #15590 always use defined method when extracting a similar body!


Ducasse:
Strange I do not get these tests failing locally :(

Ducasse:
I pushed one package too much

PR #15594, opened 2023-11-27
Fixes 15590: [RB] option shouldUseExistingMethod: should always be true

### LINKED ISSUE #15590: [RB] option shouldUseExistingMethod: should always be true

Ducasse:
For the stupidbot 
Fixes 15590 --- another try


jecisc (approved):

Ducasse:
This is strange because I did not get this change in the image I got this morning.

Ducasse:
It is there is the latest.

PR #15597, opened 2023-11-28
Packages: Add pre and post unload hooks + fix bug in manifest

### LINKED ISSUE #14073: Move package unload hooks from Monticello to RPackage
There exists some package unload hooks in Pharo using the Manifest with #postUnloadAction and #preUnload.

First it would be nice to unify the API renaming the first one into #postUnload. But the main problem is that this feature is declared in the Kernel but the code executing those actions is in Monticello.

IMO this code should be move to RPackage unloading.

See: MCWorkingCopy>>#unload

### LINKED ISSUE #14617: Add pre/post load hooks to package manifest
**Describe the request**
PackageManifest has `preUnload` and `postUnloadAction` methods to allow for cleanup as part of package unload. Sometimes it's useful to have similar actions performed as part of package load/installation. It's possible to handle this by creating a dummy class just for its `initialize` method, but this is ugly.

Ideally I'd like both a `preLoad` and `postLoad`, but `preLoad` would require special consideration to load the package manifest before other classes. Probably best to start with just `postLoad` for now.

**Expected behavior**
Class-side method(s) on `PackageManifest`, `postLoad` and eventually `preLoad`, analogous to `preUnload`. `postLoad` runs at the end of the load process (after other classes and methods), `preLoad` runs immediately after loading the manifest itself, before other classes.

**Expected development cost**
`postLoad` should be pretty easy and would be a good start. `preLoad` requires some thinking about load order to make work that I don't have time to engage with.

**Screenshots**

**Version information:** any OS/any Pharo version.

**Additional context**
Related to #5529 and #14073. Worth noting that Dolphin Smalltalk has "package scripts" for pre/post install/uninstall that are directly properties of the package itself, not attached to a manifest class, and this might be a cleaner approach and better solution to the problem of load order.

jecisc:
This change impacts multiple things:
- Move the concept of pre/post unload hooks from Monticello to Packages. With this, the code will not be executed only if you do #unload on a MCPackage but also if you remove a Package from the system. (with tests)
- Transform #postLoadAction to #postLoad that execute directly the unload code instead of returning a block 
- Fix bug in PackageManifest were the manifest was generated in the global environment even for packages that were in another environment (with test)

Fixes #14617
Fixes #14073

### DIFF
--- a/src/Deprecated12/TheManifestBuilder.extension.st
+++ b/src/Deprecated12/TheManifestBuilder.extension.st
@@ -3,12 +3,6 @@ Extension { #name : 'TheManifestBuilder' }
 { #category : '*Deprecated12' }
 TheManifestBuilder class >> ofPackageNamed: aPackageName [
 
-	| builder |
 	self deprecated: 'Call #ofPackage: instead because the only senders of this method is in NewTools and has the package directly.'.
-	builder := self new.
-	self allManifestClasses
-		detect: [ :each | each package name = aPackageName ]
-		ifFound: [ :manifestClass | builder manifestClass: manifestClass ]
-		ifNone: [ builder createManifestNamed: aPackageName ].
-	^ builder
+	^ self ofPackage: (self packageOrganizer packageNamed: aPackageName)
 ]
--- a/src/Kernel/Package.class.st
+++ b/src/Kernel/Package.class.st
@@ -645,6 +645,7 @@ Package >> removeFromSystem [
 
 	self organizer unregisterPackage: self.
 
+	"Now that the package is removed we can execute the potential postLoad. Since the class is already removed, we saved the method to execute earlier and we execute it now."
 	postUnloadMethod ifNotNil: [ :method | method valueWithReceiver: method methodClass ]
 ]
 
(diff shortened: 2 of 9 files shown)


Ducasse:
I like to get these hooks.
Now instead of doing a lookup (which is more complex) why not defining in the PackageManifest a default empty preload?

Ducasse:
Like that no need for such kind of code 
`postUnloadMethod ifNotNil: [ :method | method valueWithReceiver: method methodClass ]`

jecisc:
The code is there to execute a method that got removed from the system

Ducasse:
May be I missed something obvious like ManifestXXX superclass == Object but this is not the case.

Ducasse:
"The code is there to execute a method that got removed from the system" ?

Ducasse:
Do you mean that the method should be executed after the class Manifest has been removed?
I would then stress that in the method comment.

jecisc:
In case we have a manifest, we want to keep a reference to the method to execute. 

This method can either be:
- The method defined in the manifest if it is present
- the default #postUnload that is empty and defined in the superclass that is PackageManifest

Since the class got removed, we cannot just ask the manifest and execute the method after the removal. 

I'll add a better comment

Ducasse:
This is neat!
Such hooks are really important because we will be able to make sure package unload cleanly and this is super important. 
It will give us a lot of leverage to control the mess.

jecisc:
This was on my TODO list since a while. At first it was not so obvious how to do it but after the cleanings of Packages and Monticello it's becoming really simple :)

PR #15598, opened 2023-11-28
Try testing drivers2

Ducasse:
The idea is to be able to write some stupid and simple tests for drivers.


Ducasse:
Writing more tests for another driver...

PR #15617, opened 2023-11-28
New Refactoring Driver to rename variables

carolahp:
## Summary
Create a refactoring driver to handle the renaming of instance and shared variables.
This driver is called from Calypso's context menu item "rename2"

## Details
Refactoring validations are implemented in subclasses of RBVariableNameCondition (an abstract class created in this PR too).


Ducasse:
Cool I will have a look.

Ducasse:
Hi caro

Several points:
   - using inform: in refactoring is not possible. We should not any pop up in the refactorings.
   - basically the driver should keep asking and ask the refactoring to validate the name and if the name is not correct it should reask the user.
   - You should check how other driver perform their validation the driver will do checkApplicability....

Ducasse:
- if a condition needs to check for a situation e.g. is there a method that accesses a variable the check will set violators so that after we can just ask the condition violators (i.e. who is violating you) This is the condition that knows this information.

src/Refactoring-Core/RBIsNotGlobal.class.st
```diff
@@ -0,0 +1,32 @@
...
+		  model: aNavigationModel;
+		  yourself
+]
+
+{ #category : 'checking' }
+RBIsNotGlobal >> check [
+
+	^ (model includesGlobal: name asSymbol) not
+]
+
```
Ducasse: We could define violators to return an empty array or an array with the faulty one. 
Now sure. 
Now we need some tests for these new conditions.

src/Refactoring-Core/RBIsValidSharedVariableName.class.st
```diff
@@ -0,0 +1,21 @@
...
+	#category : 'Refactoring-Core-Conditions',
+	#package : 'Refactoring-Core',
+	#tag : 'Conditions'
+}
+
+{ #category : 'checking' }
+RBIsValidSharedVariableName >> check [
+	| string |
+	name isString ifFalse: [^false].
+	string := name asString.
```
Ducasse, line 19: I would test first the not uppercase.
Now I wonder why such functionality is not just defined in the superclass RBAbstractTransformation because it does not check that there is already an instance variable in the hierarchy for example.
It is just a textual check independent of any model.

src/Refactoring-Core/RBIsVariableNotDefinedInHierarchy.class.st
```diff
@@ -0,0 +1,31 @@
...
+	#tag : 'Conditions'
+}
+
+{ #category : 'instance creation' }
+RBIsVariableNotDefinedInHierarchy class >> name: aString class: aClass [
+	^ (self name: aString)
+		class: aClass;
+		yourself
+]
+
```
Ducasse, line 21: Here is would be good to get the class of the violation like that from a UI we could say that the class X already defines the variable Y without having the recode it in the UI.

src/Refactoring-Core/RBRenameInstanceVariableRefactoring.class.st
```diff
@@ -92,6 +115,7 @@ RBRenameInstanceVariableRefactoring >> newName: anObject [
 
 { #category : 'preconditions' }
 RBRenameInstanceVariableRefactoring >> preconditions [
+
 	^(RBCondition isValidInstanceVariableName: newName for: class)
 		& (RBCondition definesInstanceVariable: variableName in: class)
 			& (RBCondition hierarchyOf: class definesVariable: newName) not
```
Ducasse: Why not using the new condition (probably except the first one). 
To me isValidInstanceVariableName: should be called in the newName: accessor.

src/Refactoring-UI/RBRenameVariableDriver.class.st
```diff
@@ -0,0 +1,88 @@
...
+		'oldName',
+		'shouldEscape'
+	],
+	#category : 'Refactoring-UI-Drivers',
+	#package : 'Refactoring-UI',
+	#tag : 'Drivers'
+}
+
+{ #category : 'class name' }
+RBRenameVariableDriver class >> rename: aClyVariable [
```
Ducasse: no Cly in variable name.

src/Refactoring-UI/RBRenameVariableDriver.class.st
```diff
@@ -0,0 +1,88 @@
...
+{ #category : 'resources' }
+RBRenameVariableDriver >> configureRefactoring [
+
+	refactoring := variable createVariableRefactoring:
+		               self classRefactoring
+]
+
+{ #category : 'accessing' }
+RBRenameVariableDriver >> model: aRBBrowserEnvironment [ 
+	model := aRBBrowserEnvironment
```
Ducasse: There is no model in the superclass? Looks strange. We will have to pull it up

src/Refactoring-UI/RBRenameVariableDriver.class.st
```diff
@@ -0,0 +1,88 @@
...
+]
+
+{ #category : 'execution' }
+RBRenameVariableDriver >> requestNewNameBasedOn: aName [
+
+	newName := SpRequestDialog new
```
Ducasse, line 46: This should loop. The user may put several times a wrong invalid name.

src/Refactoring-UI/RBRenameVariableDriver.class.st
```diff
@@ -0,0 +1,88 @@
...
+		           title: 'Please provide a new variable name';
+		           label: 'The new name should be unique among the variables in the class hierarchy';
+		           text: aName;
+		           openModal.
+
+	newName ifNil: [
```
Ducasse: We should also check that the name is valid.

src/Refactoring-UI/RBRenameVariableDriver.class.st
```diff
@@ -0,0 +1,88 @@
...
+	self applyChanges
+]
+
+{ #category : 'accessing' }
+RBRenameVariableDriver >> scopes: aCollection [ 
+	scopes := aCollection
+]
+
+{ #category : 'accessing' }
+RBRenameVariableDriver >> variable: aClyVariable [
```
Ducasse: No Cly

Ducasse:
@balsa-sarenac can you have a look at my review

src/Refactoring-Core/RBIsNotGlobal.class.st
```diff
@@ -0,0 +1,32 @@
+"
+I check if no global in the `model` is named the same as `name`
+"
+Class {
+	#name : 'RBIsNotGlobal',
```
balsa-sarenac: I have already implemented similar class in a PR that is already integrated. It's called `RBNameIsGlobalCondition` and has this implementation. So we can avoid creating this and just do `(RBNameIsGlobalCondition new model: model className: name) not`.

src/Refactoring-Core/RBIsVariableNotDefinedInHierarchy.class.st
```diff
@@ -0,0 +1,31 @@
+"
+I check if `name` is not already used by a variable in the `class` hierarchy
+"
+Class {
+	#name : 'RBIsVariableNotDefinedInHierarchy',
```
balsa-sarenac, line 5: I would suggest to make this condition `RBIsVariableDefinedInTheHierarchy` and then use `not` to get the negated version.

src/Refactoring-Core/RBRenameClassVariableRefactoring.class.st
```diff
@@ -34,6 +34,50 @@ RBRenameClassVariableRefactoring class >> rename: aVarName to: aName in: aClass
...
+							 class name.
+							 newName }).
+			^ false ].
+
+	(RBIsNotGlobal name: newName model: model) check ifFalse: [
+		self inform:
+			('There is an existing global named {1}' format: { newName }).
+		^ false ].
+
+	^ true
```
balsa-sarenac: Here it might make sense to create a list of conditions and return it, then you can leverage `checkApplicabilityPreconditions` that will check them.

src/Refactoring-UI/RBRenameVariableDriver.class.st
```diff
@@ -0,0 +1,88 @@
...
+	newName := self requestNewNameBasedOn: oldName.
+	"class := model classNamed: variable definingClass name."
+	shouldEscape ifTrue: [ ^ self ].
+
+	[ refactoring checkInformativePreconditions ] whileFalse: [
+		newName := self requestNewNameBasedOn: newName.
+		shouldEscape ifTrue: [ ^ self ] ].
+
+	self applyChanges
+]
```
balsa-sarenac: Take a look at `RBSharedVariableDriver` and how it implemented this method. That is our target architecture for this with a few tweaks for this use case.

src/Refactoring-Core/RBIsVariableNotDefinedInHierarchy.class.st
balsa-sarenac: I just now realize that will work for `check` but violators will be wrong. I'll see to create not implementation for the violators classes.

carolahp: > I just now realize that will work for `check` but violators will be wrong. I'll see to create not implementation for the violators classes.

Yes, this would be very useful :). For the moment I will keep the `RBIsNotVariableDefinedInTheHierarchy` and I will remove it when we have the violators for `not`

src/Refactoring-UI/RBRenameVariableDriver.class.st
carolahp: The instance variable model exists in `RBInteractionDriver`, but the setter didn't. I just pulled the method model: up

src/Refactoring-UI/RBRenameVariableDriver.class.st
carolahp: Done,  now the name is a string

src/Refactoring-UI/RBRenameVariableDriver.class.st
carolahp: Done

src/Refactoring-Core/RBIsNotGlobal.class.st
carolahp: done in new commit :)

src/Refactoring-UI/RBRenameVariableDriver.class.st
carolahp: according to our last meeting, for the moment we will only validate after setting the newName to the refactoring

src/Refactoring-Core/RBRenameInstanceVariableRefactoring.class.st
carolahp: Same as the next conflict

src/Refactoring-Core/RBIsVariableNotDefinedInHierarchy.class.st
carolahp: This is a good idea, I had done it this way before, but I preferred to keep it simple and set violators just as names :/ ... we should rethink this, violators of a single condition may be of different types

src/Refactoring-Core/RBIsValidSharedVariableName.class.st
carolahp: As we spoke in the last meeting, we are keeping validations inside the conditions. But yes, syntactic validations should be different than validations depending on the model or other objects

src/Refactoring-Core/RBIsNotGlobal.class.st
carolahp: This PR does not include tests yet, I will do one exclusively with tests for the rename variable driver/transformations/conditions

carolahp:
@Ducasse @balsa-sarenac This PR only implements the rename instance variable driver (i.e. the command Rename2 applied to shared variables will halt). The PR is already big enough because I had to merge P12, so I prefer making a new one, after we merge this one, to implement the renaming of shared variables. 

I think the most important code to be reviewed is this: I havecreated "silent" method called `RBRefactoring>>failedApplicabilityPreconditions` to check for applicability conditions without throwing exceptions. This method calls `checkApplicabilityConditions`, captures its exceptions, and returns a collection with all the conditions that failed. The driver uses the information stored in these conditions to display the errors.

Ducasse:
Hi caro I will merge it now and we can continue to improve. 
I would like to go over the driver hierarchy to introduce the hooks I did to write tests.
In addition and on another topic I would like to have a convention to distinguish between 
preconditionsRaisingFailure and those not raising them. 

What about 
```
applicabilityPreconditions
       ...
       self applicabilitySilentPreconditions
```

@balsa-sarenac 

Any suggestions?

Ducasse:
There is an halt in 

Found methods with haltan OrderedCollection(SycRenameVariableCommand2>>#execute) can you remove it?

balsa-sarenac:
@Ducasse @carolahp I propose to create `failedApplicabilityPreconditions` that returns all violators of applicability. `checkApplicabilityPreconditions` would then check if this method returned empty list or not. Driver can then leverage `failedApplicabilityPreconditions` to show nice UI messages to user. Also, with this approach we don't have raising or catching errors.
Caro already has something similar, I just left a few comment on how would I do it.

bootstrap/scripts/4-build.sh
balsa-sarenac, line 1: Is this file mistakenly added?

bootstrap/scripts/runKernelTests.sh
balsa-sarenac, line 1: Is this file mistakenly added?

src/Refactoring-Core/RBIsValidSharedVariableName.class.st
```diff
@@ -0,0 +1,42 @@
+"
+I check if a name is valid for an instance variable
```
balsa-sarenac, line 2: Comment should be updated since this class is used for shared variables.

src/Refactoring-Core/RBRefactoring.class.st
```diff
@@ -66,14 +66,11 @@ RBRefactoring >> canReferenceVariable: aString in: aClass [
...
 	conditions := self applicabilityPreconditions.
-	conditions := conditions reject: [ :cond | cond check ].
-	conditions ifEmpty: [ ^ self ].
-	errorStrings := String streamContents: [ :aStream |
-		                conditions do: [ :cond |
-			                cond violationMessageOn: aStream ] ].
-	self refactoringError: errorStrings
+	failingConditions := conditions reject: [ :cond | cond check ].
+	failingConditions ifEmpty: [ ^ self ].
+	RBRefactoringError signalFor: failingConditions
```
balsa-sarenac: Here you can leverage `failingApplicabilityPreconditions` like:

```
	| failedPreconditions |
	failedPreconditions := self failingApplicabilityPreconditions.
	failedPreconditions ifEmpty: [ ^ self ].
	RBRefactoringError signalFor: failedPreconditions
```

src/Refactoring-Core/RBRefactoring.class.st
```diff
@@ -132,6 +129,18 @@ RBRefactoring >> copyOptionsFrom: aDictionary [
...
 
+{ #category : 'preconditions' }
+RBRefactoring >> failedApplicabilityPreconditions [
+
+	[self checkApplicabilityPreconditions] 
+		on: RBRefactoringError
+		do: [ :err | 
+			^ err failedConditions
+			].
+		^ #()
```
balsa-sarenac: I suggest changing this to:

```suggestion
	^ self applicabilityPreconditions reject: [ :cond | cond check ]
```

This way we avoid error handling, which is something we want to achieve and avoid all errors in the refactorings.

src/Refactoring-UI/RBRenameInstanceVariableDriver.class.st
```diff
@@ -0,0 +1,83 @@
...
+		           title: 'Please provide a new variable name';
+		           text: aName;
+		           openModal.
+
+	newName ifNil: [
+		shouldEscape := true.
+		^ self ].
+	newName = oldName ifTrue: [
+		shouldEscape := true.
+		^ self ].
```
balsa-sarenac, line 46: @Ducasse Should we stop the refactoring if the user doesn't change the name?

src/Refactoring-UI/RBRenameInstanceVariableDriver.class.st
```diff
@@ -0,0 +1,83 @@
...
+	shouldEscape := false.
+
+	[ newName := self requestNewNameBasedOn: lastName.
+	shouldEscape ifTrue: [ ^ self ].
+	refactoring newName: newName.
+	failedConditions := refactoring failedApplicabilityPreconditions.
+	failedConditions isEmpty ]
+		whileFalse: [
+			failedConditions do: [ :cond | 
+				self inform: cond errorString ] ].
```
balsa-sarenac, line 66: Currently, for `(RBNameIsGlobal ... className: name) not` inform message shows `unkonwn`. I think it's because it doesn't have `errorString` implemented. But might be because `not` doesn't work nicely with new violator architecture.
I'll create an issue for this and we can deal with it after this PR.

src/Refactoring-UI/RBRenameInstanceVariableDriver.class.st
```diff
@@ -0,0 +1,83 @@
...
+	refactoring := RBRenameInstanceVariableRefactoring
+		               variable: oldName
+		               class: class
+]
+
+{ #category : 'accessing' }
+RBRenameInstanceVariableDriver >> defaultModel [
+	^ RBNamespace onEnvironment: RBBrowserEnvironment new
+]
+
```
balsa-sarenac: I'd suggest removing this method. See comment below this one

```suggestion

```

src/Refactoring-UI/RBRenameInstanceVariableDriver.class.st
```diff
@@ -0,0 +1,83 @@
...
+			failedConditions do: [ :cond | 
+				self inform: cond errorString ] ].
+
+	self applyChanges
+]
+
+{ #category : 'initialization' }
+RBRenameInstanceVariableDriver >> scopes: aCollection variable: aString for: aClass [ 
+	scopes := aCollection.
+	model :=  self defaultModel.
```
balsa-sarenac: When creating the model, we should use `refactoringScopes` (`scopes` var here). We have `refactoringScopesOn:` defined in the super class. You can check other drivers how they create model.

```suggestion
	scopes := aCollection.
	model := self refactoringScopeOn: scopes first.
```

src/SystemCommands-RefactoringSupport/RBRefactoringError.extension.st
```diff
@@ -1,5 +1,16 @@
...
+RBRefactoringError >> failedConditions [
+	^ failedConditions
+]
+
+{ #category : '*SystemCommands-RefactoringSupport' }
+RBRefactoringError >> faledConditions: aCollection [
+	"aCollection should contain objects of kind RBNewAbstractCondition"
+	failedConditions := aCollection
+]
+
```
balsa-sarenac: Why are these defined as extension methods?  I checked and it works fine if they are not extensions. Would it make sense to put them in `RBRefactoringError`?

balsa-sarenac (commented):
Nice improvements from previous review! Left some additional comments inline.

src/Refactoring-UI/RBRenameInstanceVariableDriver.class.st
```diff
@@ -0,0 +1,83 @@
...
+	],
+	#category : 'Refactoring-UI-Drivers',
+	#package : 'Refactoring-UI',
+	#tag : 'Drivers'
+}
+
+{ #category : 'execution' }
+RBRenameInstanceVariableDriver >> applyChanges [ 
+	refactoring privateTransform.
+	refactoring performChanges
```
balsa-sarenac, line 22: Also, this refactoring doesn't preview the changes. It might be nice to show changes since we're updating references as well, I'm not sure. Old implementation didn't show preview. @Ducasse what do you think?

Take a look at remove class driver:

```suggestion

	self openPreviewWithChanges: refactoring privateTransform
```

Ducasse:
I would say that when we rename variables the preview is not necessary. 
We could have settings for this later.

carolahp:
> Hi caro I will merge it now and we can continue to improve.

Hi @Ducasse , I see you haven't merged it yet. I will take the chance to apply the last comments you and @balsa-sarenac suggested and commit them in a couple of hours

Ducasse:
Oki waiting for you :)

bootstrap/scripts/4-build.sh
carolahp: yes it was, I removed it

carolahp:
> SycRenameVariableCommand2



> @Ducasse @carolahp I propose to create `failedApplicabilityPreconditions` that returns all violators of applicability. `checkApplicabilityPreconditions` would then check if this method returned empty list or not. Driver can then leverage `failedApplicabilityPreconditions` to show nice UI messages to user. Also, with this approach we don't have raising or catching errors. Caro already has something similar, I just left a few comment on how would I do it.

@balsa-sarenac I agree with your suggestions about `failedApplicabilityPreconditions` and I applied them. 
But I feel we are still missing a "silent version" of checkApplicabilityPreconditions as @Ducasse commented above. I think this will be useful for drivers that don't need to retrieve the failed conditions.

bootstrap/scripts/runKernelTests.sh
carolahp: yes it was, I removed it

src/Refactoring-Core/RBRefactoring.class.st
carolahp: done :)

src/Refactoring-UI/RBRenameInstanceVariableDriver.class.st
carolahp: Yes you are right... thanks for creating the issue.
To prevent displaying the word "unknown", only for the moment I avoid the usage of `errorString` and instead I directly use the `violationMessageOn:` to display the error message.

balsa-sarenac:
> But I feel we are still missing a "silent version" of checkApplicabilityPreconditions as @Ducasse commented above. I think this will be useful for drivers that don't need to retrieve the failed conditions.

It makes sense. Let's add it when we see that we need it. Otherwise we might introduce something that we don't need in the future.

src/Refactoring-UI/RBRenameInstanceVariableDriver.class.st
carolahp: Oh no, sorry for the noise, my solution didn't work :c .. I will leave it with errorString for the moment (anyways the GlobalName error will never be displayed alone since all globals start with an uppercase letter

src/SystemCommands-RefactoringSupport/RBRefactoringError.extension.st
carolahp: This was a mistake, thanks for pointing this out. I just fixed it

src/Refactoring-UI/RBRenameInstanceVariableDriver.class.st
carolahp: I didn't implement the change, but just for the record, when I tried the suggested code I got the following error 
image

balsa-sarenac: Okay, thanks for the report. Probably the changes are missing.

carolahp:
> Oki waiting for you :)

@Ducasse we are ready to go, I made all the suggested changes :)
@balsa-sarenac sorry I asked for your review again by mistake and I can't cancel it !

Ducasse:
Cool

Ducasse:
Waiting for the build to finish to build and I will merge!

PR #15621, opened 2023-11-29
Ring-deprecateClassVarNamed 

MarcusDenker:
Ring uses #classVarNamed: to return a *Variable*. But the main model uses it (old style) to return the *Value* of the variable.

This PR deprecates and fixes the senders.


MarcusDenker:
Merge conflict, I created the PR with the last image that I can use Iceberg with... so fixing this has to be wither done textually or I have to wait till I can do a PR from a new Image (that is, Iceberg is fixed)

jecisc (approved):

jecisc:
Fixed

PR #15622, opened 2023-11-29
Fix refactoring AddMethodComment

badetitou:
When looking to RBAddMethodCommentTransformation I discover that it uses `surroundedBy: '"'` instead of `asComment` to convert the string to a SmallTalk comment.

So, I first made the modification. 
But I discovered that in some cases the first `"` is not duplicated, and the last is so

"ABC ABC" becomes: "ABC ABC"""

I saw that in `asComment` there is a special handling of the first `"`. So I decided to do the same for the last `"`.
I added a test for the refactoring and I extended the test case of `asComment` with the edging cases I discovered

> Sorry for the number of edits in `asComment` that appear because of code formatting...


jecisc (approved):
Nice!

PR #15623, opened 2023-11-29
#ShiftClassBuilder>>fillFor: should not use #sharedVariablesFromString:

MarcusDenker:
#ShiftClassBuilder>>fillFor: should not use #sharedVariablesFromString:


jecisc (approved):
Nice catch!

MarcusDenker:
Not that easy

The string variant was used as we have to copy the variables... to not share the instances. But a copy is better...

Issue #15631, opened 2023-11-29
Bug in return simulation


isCzech:
Here's an example:
```
[[self halt. ^ 1] on: BlockCannotReturn do: #resume ] fork
```
Run it, step over halt, step over ^1 and keep stepping over until you get a nonsensical error as a result of decoding an instruction on the pc = nil position.

Or try this example:
```
thisContext pc: nil
```
Debug it, step into #pc: nil and try to step into the return - the return should fail to execute and activate a #cannotReturn: context. Instead, it nonsensically tries to decode a nil pc instruction :)

The root cause is in #return:from: - it allows a simulation to execute an illegal return to a dead or nil context. In addition, #return:from: fails to nil the pc of the context that can't be returned to (i.e. these are two independent bugs in one line of code).

The fix is easy: just identify the illegal return in #return:from:, nil the pc and activate #cannotReturn:.

The fix, however, changes (fixes!) the semantics of #stepToCallee and #step in that currently it allows illegally stepping into a dead or nil context and after the fix it will activate the #cannotReturn: context instead of silently returning nil as a sender context, i.e. mimic the VM behavior. This clearly is an unintended behavior but was taken advantage of in #runUntilErrorOrReturnFrom: and #testSteppingSendsDirectSend.

A PR with the fixed methods will follow.

PR #15632, opened 2023-11-29
Fix a bug in return simulation

isCzech:
please check #15631 for a detailed explanation and examples


isCzech:
Hi, I'd appreciate some help here: A few debugger tests fail because they (mis)used the erroneous return behavior. E.g. #testStepThroughUntilTermination, #testStepIntoUntilTerminationClean and a few more have this line:

`[ session interruptedProcess isTerminated ] whileFalse: [ session stepThrough ].`

Before, stepping through ended up with a nil/dead context and that effectively fulfilled isTerminated because an illegal return to a dead or nil context was "executed" and a dead or nil sender was returned via the simulation (or something in this sense). Now, however, instead of silently executing the illegal return the #cannotReturn is activated and an infinite loop formed.

I'm not a debugger specialist and I'm not sure how to proceed - modify the debugger tests? Or have I missed something?? 

Thanks for your input.

jecisc:
This might interest @StevenCostiou

isCzech:
update: found another bug: the convoluted test in `Context >> isDead` is now obsolete, after recent #terminate changes. Reverting it back to the original simple `pc isNil` check. As a result, #testStepIntoUntilTermination and other StepIntoTest tests are working fine. #testStepThroughUntilTermination still needs further investigation though.

isCzech:
Hi, two Sindarin tests fail due to the fixed return semantics; I'd upload their working versions, but Sindarin packages are somehow not visible by Iceberg (neither in pharo repo nor NewTools repo). I don't know exactly how this works and what the issue with Sindarin is - please advise. 

The two failing test are:
`SindarinDebuggerTest>>#testStepOverFinishedExecution`
`SindarinDebuggerTest>>#testIsExecutionFinished`

All other tests are green so unless someone comes with something, this is pretty much the final version of the fix.
Thanks for any feedback.

src/Kernel/Context.class.st
```diff
@@ -1003,11 +1003,8 @@ Context >> isContext [
 
 { #category : 'testing' }
 Context >> isDead [
-	"Context is dead when
-	- it is self finished (returned to sender)
-	- it is at the last instruction of process termination (it is unwinded and process is suspended)"
 
-	^ pc isNil or: [  self isEndOfProcessTermination ]
```
guillep: Much better yes :)

src/Kernel/Context.class.st
```diff
@@ -1807,9 +1804,9 @@ Context >> return: value from: aSender [
 	"For simulation.  Roll back self to aSender and return value from it.  Execute any unwind blocks on the way.  ASSUMES aSender is a sender of self"
 	<reflection: 'Stack Manipulation - Controlling the stack'>
 	| newTop context |
-	aSender isDead ifTrue: [
-		^ self send: #cannotReturn: to: self with: {value} super: false ].
 	newTop := aSender sender.
+	(aSender isDead or: [newTop isNil or: [newTop isDead]]) ifTrue:
+		[^self pc: nil; send: #cannotReturn: to: self with: {value} super: false ].
```
guillep: I wonder, shouldn't we add a test for this condition? I assume that the existing tests were not covering this, right?

guillep (commented):
I think the changes look good!

Just a comment about tests: is a new test needed or not? Maybe there are other tests covering the situation already, but that feels strange given the description of the issue.

src/Kernel/Context.class.st
```diff
@@ -2411,6 +2419,13 @@ Context >> unwindTo: aContext [
 		ctx := ctx findNextUnwindContextUpTo: aContext]
 ]
 
+{ #category : 'private' }
+Context >> willFailReturn [
```
dionisiydk: Probably the better name would be a simple #canStep. 
There is a feature in the debugger to perform the return from the current pc - kind of from the arbitrary place. So the proposed name sounds like a condition to enable such a function.
And #canStep aligns much better with all senders where we do #step in a loop

dionisiydk: (It will require an inversed logic here)

src/Kernel/Context.class.st
isCzech: True

src/Kernel/Context.class.st
isCzech: The thing is you **can** step at this point but you **don't want to** because the step would initiate #cannotReturn etc which is what the VM would do in this place - to make the step and fail to return etc. What we want here is to "peek ahead" and avoid this situation; it's an analogy of InstructionStream>>willReturn etc. However, this whole #stepToCalleeOrNil method is just a quick fix to replace #stepToCalee in #runUntilErrorOrReturnFrom: while keeping its previous behavior. IMO #stepToCalee should never have been used in #runUntilErrorOrReturnFrom: - #stepToCallee should simulate a step to callee or sender but was used as a cleanup instrument to get rid of those extra guard contexts instead. Much cleaner way would be IMO to use #releaseTo: or something like this but at this point I'd prefer this simple fix to avoid confusion :) (i.e. in the future, if this PR shows no problems, I'd remove this #stepToCalleeOrNil again and replace it with something more conceptual)

isCzech:
Hi,
I'm trying to devise a simple test executing a return to a dead context to cover the fixed functionality of #return:from: but I think I've stumbled upon a weird functionality, possibly a bug, somewhere in simulation or test setup code, I'm unable to discover. This test should work fine (tried in Squeak) but fails here:
```
testBlockCannotReturn
	"Test if #return:from: correctly handles returns to dead or nil contexts."

	| p |
	p := [ thisContext pc: nil ] newProcess.
	[p suspendedContext method selector = #pc:]
		whileFalse: [p step].
	p step step. "stop at #pc: last instruction"
	
	self assert: p suspendedContext method selector equals: #pc:.
	self assert: p suspendedContext pc equals: p suspendedContext method endPC.
	
	p step.  "step into a return to a dead sender"

	self assert: p suspendedContext method selector = #cannotReturn:
```
The reason it fails is the #pc: context activation obtains an #interpretSistaV1ExtendedPush:for: sender instead of the `[ thisContext pc: nil ]` block activation that called the #pc: inside the new process p.

It feels highly illogical.

Please advise. Thanks!

isCzech:
I'm seeing "This branch has conflicts that must be resolved" warning but for some reason don't have write access despite I opened the PR... Anything I can do to check and resolve the conflicts?

isCzech:
@StevenCostiou, FYI this issue is loosely associated with #15433 merged earlier.

isCzech:
> I'm seeing "This branch has conflicts that must be resolved" warning but for some reason don't have write access despite I opened the PR... Anything I can do to check and resolve the conflicts?

How can I resolve the conflicts? Can you please reset the access rights or something preventing me from accessing the Resolve conflicts button? Thanks!

jecisc:
We did not change any setting so this is weird. 

The conflict was not something that the Github UI could handle (file renaming) so that might be the reason. 

I did a merge and pushed it on your fork.

isCzech:
Hi @jecisc, thanks for fixing the branch conflict whatever it was :) 

As I already wrote above, there are two failing Sindarin tests due to the fixed return simulation semantics, but I can't access the Sindarin packages in Iceberg. As I'm not really deeply familiar with Iceberg or the way packages are organized in Pharo, I'm not sure what to do - is Sindarin discontinued or am I doing something wrong that I can't see the packages in Iceberg. I'm really confused now.

> Hi, two Sindarin tests fail due to the fixed return semantics; I'd upload their working versions, but Sindarin packages are somehow not visible by Iceberg (neither in pharo repo nor NewTools repo). I don't know exactly how this works and what the issue with Sindarin is - please advise.
> 
> The two failing test are: 
> `SindarinDebuggerTest>>#testStepOverFinishedExecution`
> `SindarinDebuggerTest>>#testIsExecutionFinished`

jecisc:
Sindarin is not loaded directly by the baseline and because of that it is not present by default in Iceberg. This is something we need to fix.

The project can be found there: https://github.com/pharo-spec/ScriptableDebugger

<blockquote><div> GitHub</div><div><strong><a href="https://github.com/pharo-spec/ScriptableDebugger">GitHub - pharo-spec/ScriptableDebugger</a></strong></div><div>Contribute to pharo-spec/ScriptableDebugger development by creating an account on GitHub.</div></blockquote>

isCzech:
@jecisc thanks for your explanation. Now, however, I'm not sure how you'd like me to proceed: I'd like to place the two failing Sindarin tests to expected failures in the interim because I'm not sure how their authors would like to redesign them with the fixed semantics of return simulation in #return:from:. The question is: would you like me to push this dumb workaround to the project's github (not sure whether the same rules apply as for the main Pharo site) or perhaps you / test authors / someone would rather prefer to review the tests in light of the PR and proceed directly from their end? I've never pushed outside the main Pharo site and I'd hate to make any mess :) Thanks!
```
SindarinDebuggerTest >> expectedFailures
	"temporarily placed here following Pharo PR #15632 Fix a bug in return simulation"

	^ #( testIsExecutionFinished testStepOverFinishedExecution )
```

StevenCostiou:
hi @isCzech I can look at Sindarin during the sprint this week.
Can you make a PR with your expected failures method to the scriptable debugger repo?
I will merge it and open an issue to fix it, hopefully I will manage to find time this week before holidays.
Thanks very much :)

isCzech:
hi @StevenCostiou , I pushed the expected failures method

BUT:  I don't know why all symbols are being replaced with strings in the PR - this is not coming from my image... no idea why this is happening.

This is my only change:
```
SindarinDebuggerTest >> expectedFailures [

	^ #( testIsExecutionFinished testStepOverFinishedExecution )
]

{ #category : 'helpers' }
```
Thanks!

guillep:
Hi @isCzech @StevenCostiou happy new year!

> SindarinDebuggerTest >> expectedFailures [
> 
> 	^ #( testIsExecutionFinished testStepOverFinishedExecution )
> ]

This is the new tonel format I think?
Anyways, the PR changes look clean to me now.

Could you give me some heads up here to see how to continue?
What is missing for integration? (could we have a checklist in the top of the issue?)

Cheers,
G

isCzech:
Hi @guillep , happy new year to you too ! :)
I think nothing is missing, the PR is ready for integration. The last bit, making the two failing Sindarin tests expected failures was not essential for the PR. I don't have any feedback from @StevenCostiou 
Best,
J.

guillep:
Ok, thanks, let's roll this out

PR #15635, opened 2023-11-30
Rpackage some Debugging-Core classes

jecisc:
I have the impression that Debugging-Core was a package created to extract classes and methods from the Kernel but we did not know where to put the stuff.

Here I repackage some classes from it.

Next step could be to repackage the extension methods because some of them are used in the kernel but not present in kernel and most of them are used only in some debugger packages


jecisc:
I think the first steps of the bootstrap are failing on packages that do not contains classes but only extensions :o

PR #15641, opened 2023-11-30
Ensure real superclass setting in RBInsertNewClassRefactoring

### LINKED ISSUE #15602: Cannot insert class anymore because ShiftClassBuilder >> createClass create nil metaclass
Try to insert a superclass the system breaks because newClass below is nil.

```
ShiftClassBuilder >> createClass

	newClass := newMetaclass new.

	newClass setName: self name.

	self builderEnhancer
		configureClass: newClass
		superclass: self superclass
		withLayoutType: self layoutClass
		slots: (self withAdditionalSlots: self slots).

	newClass environment: self installingEnvironment.

	self builderEnhancer classCreated: self
```

hernanmd:
This PR fixes the refactoring browser problem reported in #15602. 

It modifies the RB code to ensure real superclass setting in RBInsertNewClassRefactoring>>privateTransform (this is to avoid setting the RBClass instead of the real superclass).
Also set the real superclass in RBInsertNewClassTransformation instead of the superclass name.
Add also a test reproducing the wrong behavior which now passes for both tests parameters (insert and transformation refactorings).

### DIFF
--- a/src/Refactoring-Core/RBInsertNewClassRefactoring.class.st
+++ b/src/Refactoring-Core/RBInsertNewClassRefactoring.class.st
@@ -71,7 +71,7 @@ RBInsertNewClassRefactoring >> privateTransform [
 	self model
 		defineClass: [ :aBuilder |
 			aBuilder
-				superclass: superclass;
+				superclass: (self model classObjectFor: superclass);
 				name: className;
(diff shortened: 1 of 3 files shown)


jecisc (approved):
Thanks!

PR #15643, opened 2023-11-30
Rename AST-Core classes to start by AST

### LINKED ISSUE #10815: Cleanup: Prefix change from "RB" to "AST" for Abstract syntax tree related classes

jecisc:
Rename classes from AST-Core to start by AST instead of starting by RB.

Since AST-Core is loaded by Hermes I just need to check if the initializes are executed.

Fixes #10815


Ducasse:
there is a conflict

MarcusDenker:
With the aliases, renamings like that are finally possible... but sometimes I wonder if we should not wait till we have a real module system: any prefix is just bad. Whatever it is. But changing it just because it feels nicer is still a change that everyone has to deal with... for what in the end ?

Ducasse:
I like the idea because it is not clear why we have RB for nodes.

Ducasse:
And this is not tomorrow that we will have modules because it raises a lof of questions.

Ducasse:
@jecisc I think that we should redo this PR

jecisc:
I did a smaller PR because this one is too big

PR #15651, opened 2023-12-01
move error handling from CompiledMethod to CompiledCode and rename

MarcusDenker:
... the code was never fixes for the CompiledCode hierachy.

This PR moves the methods up and renames them to not be specific for CompiledMethod


MarcusDenker:
I will add an issue for CompiledCode newFrom:, there is no test and I think it does not work.
(but this PR is nevertheless good to merge independent of that)

PR #15654, opened 2023-12-01
do not use #slotsFromString: / sharedVariablesFromString:

### LINKED ISSUE #15657: Cleanup Class Builder users: only use #sharedVariablesFromString:  if needed
**Describe the problem**
#sharedVariablesFromString: should only be used in cases where it can not be avoided (old style class defs, Hermes)

--> rewrite all the other users to use sharedVariables: instead. (these are mostly tests)

MarcusDenker:
slotsFromString: should only be used by code that is dealing with old style class defs

This PR fixes lots of tests to use #slots: with an array instead.


Ducasse:
Hi marcus did you update the comment of the method to indicate 'slotsFromString: should only be used by code that is dealing with old style class defs'?

MarcusDenker:
Yes, I will do that next (maybe it even makes sense to inline them into the senders where they are used and then deprecate them on the level of the ClassBuilder).

I merge this and do another PR

PR #15656, opened 2023-12-01
Extract ClassDefinitionPrinters from the Kernel

jecisc:
And load it with Opal instead of Kernel


Ducasse:
:) indeed we do not really need them in kernel.

MarcusDenker (approved):

PR #15670, opened 2023-12-02
Fix ephemeron support in Fuel

### LINKED ISSUE #15669: Race condition in Fuel ephemeron tests
Tests on the CI are failing unpredictably because of a race condition in the new ephemeron tests of Fuel.

theseion:
- Raise error if ephemeron implemenation or use are not supported
- Hold strong references to ephemeron keys to prevent the GC from modifying the graph during serialization

Fixes #15669

### DIFF
--- a/src/Fuel-Core/Object.extension.st
+++ b/src/Fuel-Core/Object.extension.st
@@ -13,6 +13,21 @@ Object >> fuelAfterMaterialization [
 ]
 
 { #category : '*Fuel-Core' }
+Object >> fuelCheckEphemeronSupport [
+	FLNotSerializable signal: 'Unsupported ephemeron implementation'
+]
+
+{ #category : '*Fuel-Core' }
+Object >> fuelIsEphemeron [
+	^ self class classLayout class == EphemeronLayout
+]
+
+{ #category : '*Fuel-Core' }
+Object >> fuelWasMourned [
+	^ false
+]
+
+{ #category : '*Fuel-Core' }
 Object >> serializeOn: aStream [
 	FLSerializer
 		serialize: self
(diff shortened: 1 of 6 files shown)


Ducasse:
Thank Max!

src/Fuel-Core/Object.extension.st
```diff
@@ -12,6 +12,21 @@ Object >> fuelAfterMaterialization [
...
 ]
 
+{ #category : '*Fuel-Core' }
+Object >> fuelCheckEphemeronSupport [
+	FLNotSerializable signal: 'Unsupported ephemeron implementation'
+]
+
+{ #category : '*Fuel-Core' }
+Object >> fuelIsEphemeron [
+	^ self class classLayout class == EphemeronLayout
```
tinchodias, line 22: Max, many thanks for the fix. I only a very small comment on this method, for the moment. I browsed `EphemeronLayout` and it seems to be the only layout class without a testing method such as `isEphemeron` (maybe it should exist, but that's not part of this bugfix). 

But: 

Do you think we can replace sending `anObject fuelIsEphemeron` by `anObject class isEphemeronClass`? (`Behavior` understands `isEphemeronClass`). Then we remove this extension.

But2: Maybe `specialAssociationsOf:do:` could delegate to the associations' and the associations to their layout, decide how to serialize without need of any "if". But that's too picky maybe. 

I don't manage to make me an idea of how this would be resolved in general... there is no way to do it? I see from this test that each class with ephemeron layout will have to explicit a wahy to serialize it.

theseion: Thanks for the feedback. I hadn't seen `#isEphemeronClass`, that's a good idea.

As for the second point, I didn't want to restrict ephemerons to subclasses of `Association`. I don't know what the VM implementation is, but on the image side there is no technical reason that ephemerons have to be subclasses of `Association`. The only requirement is that the class has at least one instance variable, which is treated as the key by the GC.

theseion: Made the change.

PR #15674, opened 2023-12-02
Cleanup/extract method first pass

balsa-sarenac:
Started working on extract method. I did a pass on tests and commands:

* All parametrized tests that have single scenario are migrated to classic tests
* Renamed and added comments for ExtractMethodTransformation tests for better understanding
* Give better name to extract method commands
  * extract method - was called extract method 2
  * extract method and occurrences - was called extract method
* A few other small fixes


Ducasse:
Nice :)

PR #15675, opened 2023-12-02
Refactor: extract method transformation cleanup

balsa-sarenac:
I did a second pass on extract method. This time I focused on extract method transformation. This PR includes:

* Extraction of preconditions into separate methods for easier readability (in the next PR I'll start reifying these preconditions)
* Removing non-precondition-checking code from preconditions
* Utilizing new architecture hooks like `prepareForExecution`
* Removing unnecessary temps and instance variables


Ducasse:
Waiting for build validation

Ducasse:
This is strange because I thought that I integrated it yesterday.

PR #15685, opened 2023-12-04
Do not use  #classVarNamed: /  #classVarNamed:put:

MarcusDenker:
do not use #classVarNamed: and classVarNamed:put:, as it is not clear that #classVarNamed:  returns the Value, not the Variable


MarcusDenker:
I should have merged it... now there is a conflict

jecisc:
Fixed

PR #15693, opened 2023-12-04
Refactor: extract method third pass

balsa-sarenac:
This PR continues on previous works and focuses on using new architecture and reifying preconditions.

* remove duplicate calls to methods that weren't cached
* reify all preconditions of the refactoring and add tests for them
* split complex precondition to simpler ones
* utilize `prepareForExecution` to setup all prerequisite code
* add some docs to better explain preconditions


Ducasse:
I will integrate this. 
Now I found strange that the tests RBMax.... do not manipulate and AST or something like that.

balsa-sarenac:
@Ducasse I am aware of that. I tried extracting that logic to precondition, but then it didn't look right, because it was responsible for too many things. Also, a problem was that then the refactoring needed to ask the precondition for the state in the middle of precondition checking, and that didn't seem right to me. Maybe it is, what do you think?

PR #15695, opened 2023-12-05
Classbuilder: rewrite users of #sharedPools: with string arguments

MarcusDenker:
- change some users that use sharedPools: with a string as a paramet.
- add #sharedPoolsFromString: to be used in old style class def implementation


jecisc (approved):
I like the uniformisation!

PR #15699, opened 2023-12-05
15684-ClassBuilder-Enhance-ShSharedVariablesChangeDetector

### LINKED ISSUE #15684: ClassBuilder: Enhance ShSharedVariablesChangeDetector
ShSharedVariablesChangeDetector checks if the SharedVariables changed when updating a class.

The detector now just takes into account the name of the Variable (to ignore the value, which is important!)

```
initialize
	super initialize.
	builderAccessor := [ :e | e layoutDefinition sharedVariables collect:[ :x | x key -> x class ] as: Array ].
	classAccessor := [ :e | e classVariables collect: [ :x | x key -> x class ] as: Array ].
	comparer := [ :a :b | self compareFrom: a to: b ]
```

with:

```
compareFrom: a to: b
	^ a asDictionary = b asDictionary
```

The problem with this is that it does not work with the (experimental) feature that we can define our own shared variables.

Here is an example. We might want to experiment with a variable that defines it's default value as part of the class definition
(instead of relying on class side #initialize)"

```
Object << #MyClass
	slots: {};
	sharedVariables: { #MyTestVar => InitializedClassVariable default: 1 };
	package: 'Slot-Core'
```

This works nicely, the class builder sets the value when the class is created. But if we now change the default:

```
Object << #MyClass
	slots: {};
	sharedVariables: { #MyTestVar => InitializedClassVariable default: 2 };
	package: 'Slot-Core'
```

The change is ignored, as the ShSharedVariablesChangeDetector does not care about anything than the name and thus can not see that the variable was changed and needs to be updated in the new class.

ShSlotChangeDetector does it already better:

```
initialize

	super initialize.
	builderAccessor := [ :e | e allSlots asArray ].
	classAccessor := [ :e | e allSlots asArray ].
	comparer := [ :a :b | self compareSlotCollection: a with: b ]
```

with:

```
compareSlotCollection: a with: b

	(a size = b size) ifFalse: [ ^ false].

	a with: b do: [ :aSlot :bSlot |
			(aSlot hasSameDefinitionAs: bSlot) ifFalse: [ ^ false ] ].

	^ true
```

We should check to do the same here and implement #hasSameDefinitionAs: in the ClassVariable hierarchy.

MarcusDenker:
This is the described change so that ClassVariables are compared with #hasSameDefinitionAs: instead of just the names.

This is *not* yet enough  to fix the problem, will open another PR for that.

fixes #15684

### DIFF
--- a/src/Collections-Support/Association.class.st
+++ b/src/Collections-Support/Association.class.st
@@ -42,6 +42,13 @@ Association >> hasLiteral: aLiteral [
 	^false
 ]
 
+{ #category : 'testing' }
+Association >> hasSameDefinitionAs: otherVariable [
+
+	"As the bootstrap creates class Variables as Associations, we need to be able to compare"
+	^self class == otherVariable class and: [self key = otherVariable key]
+]
+
 { #category : 'variables-toclean' }
 Association >> isAssociation [
 	^ true
--- a/src/Collections-Support/Association.class.st
+++ b/src/Collections-Support/Association.class.st
@@ -46,7 +46,7 @@ Association >> hasLiteral: aLiteral [
 Association >> hasSameDefinitionAs: otherVariable [
 
 	"As the bootstrap creates class Variables as Associations, we need to be able to compare"
-	^self class == otherVariable class and: [self key = otherVariable key]
+	^self class = otherVariable class and: [self key = otherVariable key]
 ]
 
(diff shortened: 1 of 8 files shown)


MarcusDenker:
I added part of it to https://github.com/pharo-project/pharo/pull/15747 to make it easier to experiment to find why this fails

PR #15702, opened 2023-12-05
Fix a ZeroDivide exception when scroll target is 0 calculating scroll deltas

### LINKED ISSUE #15701: ZeroDivide exception setting scroll deltas in GeneralScollPaneMorph
**Bug description**

When setting scroll deltas of a scrollable morph, if the scroll target is 0 then the scroll interval is calculated as the minimum between 0 and 1.0, thus 0 which gives a ZeroDivide exception in method hScrollbarInterval.

**To Reproduce**

```smalltalk
SpPresenter new
	layout: (SpBoxLayout newTopToBottom 
		add: (SpScrollableLayout with: SpBoxLayout newTopToBottom);
		yourself);
	open.
``` 

**Fix**

Thanks to @estebanlm which discovered and fixed this Morphic bug.

hernanmd:
### Issue
ZeroDivide exception occurs when calculating scroll deltas with a scroll target width of 0 in a scrollable morph.

### Solution
Safeguard the scroll width with a default value of 1.0 when the scroll target is 0, to prevent the exception and ensure smooth scrolling behavior.

Closes #15701

### DIFF
--- a/src/Morphic-Widgets-Scrolling/GeneralScrollPaneMorph.class.st
+++ b/src/Morphic-Widgets-Scrolling/GeneralScrollPaneMorph.class.st
@@ -143,7 +143,7 @@ GeneralScrollPaneMorph >> hScrollbar: anObject [
 GeneralScrollPaneMorph >> hScrollbarInterval [
 	"Answer the computed size of the thumb of the horizontal scrollbar."
 
-	^self scrollBounds width asFloat / self scrollTarget width min: 1.0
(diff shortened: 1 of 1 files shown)


estebanlm (changes requested):
you need to do the same with `vScrollbarInterval` :)

hernanmd:
That's true, updated.
Thanks.

estebanlm (approved):

PR #15703, opened 2023-12-05
Tests and enhancements for RenameInstanceVariableDriver

carolahp:
Implements tests for the driver. 
Also implements an alternative way to execute the driver in script mode (not requiring interaction with the user, and signaling errors when the applicability conditions are not met)


Ducasse:
I will integrate it and we will iterate.

PR #15705, opened 2023-12-06
No rule check for non-existing super implementor (Second PR)

### LINKED ISSUE #12766: No rule check for non-existing super implementor
**Bug description**
We do not have a code rule check that will inform about non-existing implementor of a message send if the message is sent by `super`.

**To Reproduce**
1. in any class, create this method
```smalltalk
uniqueMessage 

	self asString.
	^ super uniqueMessage
```
2. no warning

(asString message is here just to prevent warning about a method using super send only)

astares:
Fix #12766


MarcusDenker:
#testNoSentButNotImplementedSelectors is failing for ReSentButNotUnderstoodBySuperRuleTest>>#sampleMethod

Ducasse:
testNoSentButNotImplementedSelectors  is really not self explanatory. I never get what is the No.
We should rename it and provide a better explanation.

astares:
Yes, we could improve the name-

testNoSentButNotImplementedSelectors  is basically "test that we have no selectors remaining that are sent but not  implemented"

or like 

#testThatThereAreNoSelectorsRemainingThatAreSentButNotImplemented

PR #15712, opened 2023-12-06
15596-Cleanup-respondsTo-check-for-doItReceiver

### LINKED ISSUE #15596: Cleanup #respondsTo: check for #doItReceiver
**Describe the problem**
We have now multiple places with checks of the kind:

```
	(self context respondsTo: #doItReceiver)
		ifTrue: [
			doItReceiver := self context doItReceiver.
			doItContext := self context doItContext ]
		ifFalse: [ doItReceiver := doItContext := nil ].
```

This is not nice. We can just implement #doItReceiver and #doItContext (to return nil if not needed) for all the models/context
that evaluate code.

MarcusDenker:
remove #respondsTo:

All models that we want to use for evaluation should implement the required API (they should already)

fixes 15596


Ducasse:
Yes!

PR #15717, opened 2023-12-06
rewrite #anyUserOfClassVarNamed: using the Variable API

MarcusDenker:
To improve the variable API, rewrite #anyUserOfClassVarNamed:

- add ClassVariable #usingClasses and test
- rewrite #anyUserOfClassVarNamed:

### DIFF
--- a/src/Kernel-CodeModel/Class.class.st
+++ b/src/Kernel-CodeModel/Class.class.st
@@ -191,11 +191,10 @@ Class >> allSharedPools [
 
 { #category : 'class variables' }
 Class >> anyUserOfClassVarNamed: aSymbol [
-	self withAllSubclasses do: [ :subclass |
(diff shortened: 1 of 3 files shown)


src/Kernel-CodeModel/Class.class.st
```diff
@@ -191,11 +191,10 @@ Class >> allSharedPools [
 
 { #category : 'class variables' }
 Class >> anyUserOfClassVarNamed: aSymbol [
```
jecisc, line 193: This seems really specific indeed I think we can deprecate it

jecisc (approved):

src/Kernel-CodeModel/Class.class.st
MarcusDenker: is was used to find if a class variable was used in code. For that of course we have a better API on the Variable

PR #15723, opened 2023-12-07
Fix: Syntax highlight of inst vars in inspector eval pane

### LINKED ISSUE #503: Inspector: Syntax highlighting of Variables is not correct in eval pane
The screenshot shows that if I reference "x" in the eval pane, it it highlighted red. This is wrong (and confusing).

Eval work correctly (I can change x by assigning it)

inspector

### LINKED ISSUE #15726: Syntax Highlight in in the inspector
**Bug description**
The syntax hightlight does not take variables into account in the inspector (see https://github.com/pharo-spec/NewTools/issues/503).

If we look at SHRBTextStyler>>#privateStyle:, we see that it ignores the instance and the context of the requestor:

```
privateStyle: aText
	| compiler |

	aText ifEmpty: [ ^ self ].
	compiler := classOrMetaClass compiler
		source: aText asString;
		isScripting: self isScripting;
		requestor: workspace.
        ...
```

For the inspector, it works if we configure the compiler with the doItReceiver:
```
compiler := classOrMetaClass compiler
		source: aText asString;
		receiver: (workspace ifNotNil: [:w | w doItReceiver]);
		context: (workspace ifNotNil: [:w | w doItContext]);
		isScripting: self isScripting;
		requestor: workspace.
```

But even though we pass the doItContext, this now breaks the debugger syntax hightlight, it seems to now parse the methods as "isScripting" (that is, not methods). But adding a traceCr for the context and the isScripting shows that both is correct. Nevertheless, the hightlighting is wrong in the debugger. I have no idea why. 

Why was it even correct before?

MarcusDenker:
#privateStyle: should ask the workspace for a #doItReceiver if not nil.

This fixes syntax highlight in the inspector eval pane

fixes https://github.com/pharo-spec/NewTools/issues/503

### DIFF
--- a/src/OpalCompiler-Core/CompilationContext.class.st
+++ b/src/OpalCompiler-Core/CompilationContext.class.st
@@ -471,7 +471,7 @@ CompilationContext >> initialize [
 
 { #category : 'accessing' }
 CompilationContext >> isScripting [
-	^semanticScope isDoItScope
+	^semanticScope isScripting
 ]
 
 { #category : 'accessing' }
--- a/src/OpalCompiler-Core/CompilationContext.class.st
+++ b/src/OpalCompiler-Core/CompilationContext.class.st
@@ -471,7 +471,7 @@ CompilationContext >> initialize [
 
 { #category : 'accessing' }
 CompilationContext >> isScripting [
-	^semanticScope isScripting
+	^semanticScope isDoItScope
 ]
 
 { #category : 'accessing' }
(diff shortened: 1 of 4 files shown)


MarcusDenker:
But it breaks syntax hightlight in the debugger

MarcusDenker:
I close it for for and open an issue tracker entry, I really do not understand why the debugger parses wrong even after setting the context...

MarcusDenker:
Change isScripting to be false if we eval against a context. But this is not good... more work needed

MarcusDenker:
So I still think we need to be able to model "I parse against an object / a context, but I am not. DoIT"

MarcusDenker:
The solution: set the receiver only if isScripting is true.

PR #15732, opened 2023-12-07
Traits: Fix update of source pointers

jecisc:
It seems that traits were setting the source pointers of the copied methods in system dictionary while adding the method to the class. 

A problem is that the logging of the source code now happens after we install the methods in the system because some info were missing if we logged this before. This result in the fact that currently we have copied methods missing their source pointer. 

I propose an alternative here by updating the source pointer of the copied methods in the users upon the update of the source pointer in the original method instead of doing it while adding the method to the user.

Also I deprecated #setSourcePointer: that seems to be doing the same is #sourcePointer: now (the removal of the source property was duplicated..)

I'm not so sure about this change so if @tesonep has some time to give feedback I'd appreciate it :) 

This bug was detected while I was trying to use Moose in Pharo 12

### DIFF
--- a/src/Kernel-CodeModel/CompiledMethod.class.st
+++ b/src/Kernel-CodeModel/CompiledMethod.class.st
@@ -180,8 +180,8 @@ CompiledMethod >> classBinding: aBinding [
 
 { #category : 'source code management' }
 CompiledMethod >> clearSourcePointer [
-	"we use #setSourcePointer: to not clear the property #source"
-	self setSourcePointer: 0
+
+	self sourcePointer: 0
 ]
 
 { #category : 'accessing' }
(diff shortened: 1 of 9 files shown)


jecisc:
This does not seems to work and I have to say that I have a hard time to think of a good solution.

Here there is a mechanism to set the source pointers when we copy/add methods to the users (dealing with aliasing also). But it does not work in Pharo 12 since the logging of methods happens after installation. 

And setting the source pointers after seems to lead to strange effects and I think it does not deal well with aliases of methods.

So I don't know how to deal with this problem right now... :(

src/Kernel-CodeModel/CompiledMethod.class.st
```diff
@@ -925,9 +913,16 @@ CompiledMethod >> sourcePointer [
...
 	"Drop the #source property if any"
 	self removeProperty: #source.
-	self setSourcePointer: srcPointer
+	trailerSize := self class trailerSize.
+	trailerBytes := srcPointer asByteArrayOfSize: trailerSize.
+	start := self size - trailerSize.
+	start + 1 to: self size do: [ :n | self at: n put: (trailerBytes at: n - start) ].
+
+	"In case the mtehod is definied in a trait, we need to update the source pointers of the copies."
+	self origin = self methodClass ifTrue: [ self origin notifySourcePointerOf: self selector updateTo: self sourcePointer ]
```
MarcusDenker, line 925: Hmm... updating the source pointer of the copies feels wrong. If the original will be changes, it will be re-installed anyway, or not?

MarcusDenker:
What is strange: I do not understand why there where methods with wrong source pointers in the first place. When we copy the traits method, we do set the pointer. So if the source pointer is 0, then the one of the orginal method from the Trait was already wrong. And that is a "normal" method... 

So where does the problem come from? Maybe the condenser is doing something wrong?

jecisc:
The methods have a wrong source pointer because we add the methods to the users right after we install them in the origin class. 

In the past we already logged the source pointer so it worked. 
Now the logging of method happens after the installation. So, when we add the method, we pass the source pointer of the original method, but this one is 0 because it is not set yet. It'll be set after the installation

jecisc:
I reverted in another PR what caused this problem

PR #15647, opened 2023-12-01
Add ScalingCanvas and #canvasScaleFactor for OSWorldRenderer

Rinzwind:
This pull request:

* Adds ScalingCanvas as a subclass of PluggableCanvas. A ScalingCanvas provides scaling on an underlying FormCanvas. The class is taken from the [‘Retina display’ scaling experiment mentioned in previous pull requests](https://github.com/pharo-project/pharo/pull/15561#issuecomment-1826439810).
* Introduces a setting `#canvasScaleFactor` for OSWorldRenderer which determines the scale of the Form used to draw a WorldMorph. The world is drawn on a ScalingCanvas, instead of directly on a FormCanvas, when the setting is set to a value other than 1.

The [screenshots listed in pull request #15561](https://github.com/pharo-project/pharo/pull/15561#issuecomment-1826439810) show the effect of setting the `#canvasScaleFactor` to 2 on Macs with a ‘Retina display’.

The `#canvasScaleFactor` can be set through the Settings Browser so that it can also be saved to disk:

<p align="center">

</p>

Note that pull request #15561 mentions doing `World scaleFactor: 0.5` as a way to use ScalingCanvas, but this no longer applies as the `#canvasScaleFactor` has been separated from the world `#scaleFactor`. The two factors can be set independently. Though on Macs with a ‘Retina display’, it seems best to set them so that the multiplication of the `#canvasScaleFactor` by the reciprocal of the `#scaleFactor` is 2. The following cycles through such combinations:

```smalltalk
OSWorldRenderer canvasScaleFactor: OSWorldRenderer canvasScaleFactor \\ 5 + 1.
World scaleFactor: OSWorldRenderer canvasScaleFactor / 2; changed.
```

A few methods are refactored on FormCanvas to offer a ‘scale’ parameter for use by a ScalingCanvas. As [previously mentioned](https://github.com/pharo-project/pharo/pull/14665#issuecomment-1720133696), it might be better to turn ScalingCanvas into a subclass of FormCanvas. The FormMapping that is used in ScalingCanvas needs to be removed but is still needed because icons are not yet consistently represented as a FormSet with a Form for more than one scale. The `#canvasScaleFactor` is global while it might be better to have a separate factor per world. These and any other necessary improvements can however be left to ‘future work’; merging this now will make it easier for others to give it a try.

I will update the [Pharo 11 backport](https://github.com/pharo-project/pharo/pull/15561#issuecomment-1828476134) when this pull request is merged.

PR #15736, opened 2023-12-07
Rework #nonCachingFullDrawOn: on HandMorph to use a matching ScalingCanvas for the shadow form

Rinzwind:
This pull request reworks the shadow form drawing in `#nonCachingFullDrawOn:` on HandMorph to use a ScalingCanvas with the same scale as the argument when it’s a ScalingCanvas. That avoids the upscaling of the shadow form when the `#canvasScaleFactor` is not 1. The method applies when dragging translucent morphs like:

```smalltalk
Morph new
	color: (Color blue alpha: 0.25);
	extent: 1000 asPoint;
	openInWorld
```

And:

```smalltalk
(Text string: 'Pharo'
	attribute: (TextFontReference toFont:
		(LogicalFont familyName: 'Source Sans Pro' pointSize: 400)))
	asTextMorph openInWorld
```

Note that dragging the TextMorph can be slow due to its large glyphs causing the FreeTypeCache to repeatedly reach its maximum size and be cleared. That can be avoided by setting the maximum size to take the `#canvasScaleFactor` into account:

```smalltalk
FreeTypeCache current maximumSize:
	FreeTypeCache defaultMaximumSize * (OSWorldRenderer canvasScaleFactor raisedTo: 2)
```

### DIFF
--- a/src/Athens-Morphic/AthensCanvasWrapper.class.st
+++ b/src/Athens-Morphic/AthensCanvasWrapper.class.st
@@ -475,6 +475,12 @@ AthensCanvasWrapper >> origin: aPoint [
 ]
 
 { #category : 'canvas drawing - images' }
+AthensCanvasWrapper >> paintFormSet: aFormSet at: aPoint [
+
+	self paintImage: aFormSet asForm at: aPoint
+]
+
+{ #category : 'canvas drawing - images' }
 AthensCanvasWrapper >> paintImage: aForm at: aPoint [
 	self paintImage: aForm at: aPoint sourceRect: aForm boundingBox
 ]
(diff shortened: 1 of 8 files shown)


jecisc:
Maybe when the scaling change we could update the size of the cache at the same time?

jecisc (approved):

MarcusDenker (approved):

Rinzwind:
I’m not sure about updating the cache size automatically when the canvas scale factor changes because I’m not sure what a good size for the cache is. In pull request #15647 I gave a snippet that cycles through canvas scale factors 1 through 5. At 5, the snippet I gave above for setting the maximum size of the FreeTypeCache would set it to 128 000 000 which might be too large.

PR #15739, opened 2023-12-08
ClassInstaller: use the oldClass of the builder

### LINKED ISSUE #15707: Copying a class creates slots with wrong owning class
**Bug description**
Try this:

```
CairoUTF8ConverterTest duplicateClassWithNewName: #TT.

(CairoUTF8ConverterTest slotNamed: #encoder) definingClass "TT"
```

duplicateClassWithNewName: uses #fillFor:, it makes sense to copy the variables here.

But there is ShLayoutDefinition>>#copySlotsIfUsedIn: called during the build which seems related.

MarcusDenker:
remove the use of the oldClass ivar in the ClassInstaller. Always use the oldClass of the builder

The ivar has to be removed with a PR created from source, as it can not be done from the image

### DIFF
--- a/src/Shift-ClassBuilder-Tests/ShClassInstallerTest.class.st
+++ b/src/Shift-ClassBuilder-Tests/ShClassInstallerTest.class.st
@@ -50,7 +50,8 @@ ShClassInstallerTest >> newClass: className superclass: aSuperclass slots: slots
 
 { #category : 'running' }
 ShClassInstallerTest >> tearDown [
-
+	newClass  ifNotNil: [:c | c removeFromSystem].
+	newClass2 ifNotNil: [:c | c removeFromSystem].
 	self packageOrganizer removePackage: self generatedClassesPackageName.
 	super tearDown
 ]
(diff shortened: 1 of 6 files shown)


jecisc (approved):

MarcusDenker:
More complex than I thought... it seems that when we just change the name, the ClassBuilder does not detect a change.

But the code with the oldClass in the installer will use the new name to look up the class, find nothing, and sets oldClass to nil in the builder. oldClass nil will disable the change detector and thus not raise "no change detected".

With oldClass being nil on the level of the installer, it will not call ShiftClassInstaller>>#migrateClassTo:

This is all very very confusing

MarcusDenker:
now TraitTest>>testUsingTraitInAnonymousSubClassAndRedefiningIt is failing, this has to do something with it: as the test never installs the class, the re-lookup was putting oldClass to nil. But now we re-build the class, and here the propagation to the users of the trait fails.

There is the possibility that i was broken since ever but hidden by the re-lookup from the environment which fails for anon classes always.

Issue #15620, opened 2023-11-29
By default, errors are logged twice to the PharoDebug.log


pavel-krivanek:
**Bug description**
By default, errors are logged twice to the `PharoDebug.log` file because `DebugSession logDebuggerStackToFile` is true by default.

Here are the call stacks for both writes:
![image](https://github.com/pharo-project/pharo/assets/572494/bdf0d322-5c9f-4fcc-afa7-010ac9feee43)

**To Reproduce**
1. In a fresh image, do `self halt`in Playground
2. look at `PharoDebug.log`, it contains twice the same content

hernanmd:
From a quick pass on the code, it seems from the code that we could be mixing two concepts: One is dumping stack traces during error exceptions, and another one is dumping stack traces during request opening of a debugger. So I thought about adding a setting to specifically enable stack dump of debugger open requests.

So requesting to open a debugger should log the stack trace?

The entry point to these changes could be `logError:inContext:`

hernanmd:
Closing since it is fixed now.

PR #15746, opened 2023-12-08
Prevent logging exceptions twice when debugger is opened

hernanmd:
Change logDebuggerStackToFile default setting to false since when set to true, it makes the debug session log twice on every request as reported in #15620.

### DIFF
--- a/src/Debugger-Model/DebugSession.class.st
+++ b/src/Debugger-Model/DebugSession.class.st
@@ -60,7 +60,8 @@ DebugSession class >> defaultContextModelClass [
 
 { #category : 'settings' }
 DebugSession class >> logDebuggerStackToFile [
-	^ LogDebuggerStackToFile ifNil: [LogDebuggerStackToFile := true]
+	^ LogDebuggerStackToFile 
+		ifNil: [ LogDebuggerStackToFile := false ]
 ]
 
 { #category : 'settings' }
--- a/src/Debugger-Model/DebuggerSettings.class.st
+++ b/src/Debugger-Model/DebuggerSettings.class.st
@@ -95,7 +95,7 @@ DebuggerSettings class >> generalDebuggingSettingsOn: aBuilder [
 			(aBuilder setting: #logDebuggerStackToFile)
 				label: 'Write message to debug log file when fall into debugger';
 				target: DebugSession;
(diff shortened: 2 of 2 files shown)


guillep:
Maybe you would like to discuss this with @StevenCostiou ?

If Pharo is already logging exceptions, why should the debugger do it also?
My point here is that maybe the code of this setting should just be removed, but I'm not sure :).
Otherwise, changing the setting looks like it's not solving the cause of the problem but its consequence, right? People with a bad setting could also have a bad experience, and not understand where it comes from.
In other words: if the setting has no value, why having it?

StevenCostiou:
After discussion, we propose to merge this PR and open an issue to specifically discuss the need for a second mechanism.
It is not clear right now if we do or not.
However not logging upon debugger opening can improve a bit the speed? - and we need to gain avery possible bit of speed in the debugger ;)

hernanmd:
Ok, I created another issue for this: #15810

Ducasse:
I merge as requeted.

daniels220:
Dang, I've been thinking about saying something about this for a long time, just now getting to it—wish I'd slipped my thoughts in sooner, because:

I think, instead, the system should skip the _unconditional_ logging to the file, in `UIManager>>requestDebuggerOpeningFor:`, allowing the existing setting to do exactly what it says—control whether the stack is logged _at all_ when opening a debugger. In which case I think `true` is a sensible default. As a fallback we should also log in `requestDebuggerOpeningFor:` if `OupsDebuggerSystem` is not present.

guillep:
> Dang, I've been thinking about saying something about this for a long time, just now getting to it—wish I'd slipped my thoughts in sooner, because:
> 
> I think, instead, the system should skip the _unconditional_ logging to the file, in `UIManager>>requestDebuggerOpeningFor:`, allowing the existing setting to do exactly what it says—control whether the stack is logged _at all_ when opening a debugger. In which case I think `true` is a sensible default. As a fallback we should also log in `requestDebuggerOpeningFor:` if `OupsDebuggerSystem` is not present.

I like the idea :)

daniels220:
Actually, I'm going to kinda take that back—but I'm also going to put my actual comment on #15810, since that's the open issue.

PR #15751, opened 2023-12-09
Make FreeTypeFont reuse instances

Rinzwind:
This pull request makes FreeTypeFont reuse instances like FreeTypeFace and LogicalFont do. That allows the FreeTypeCache to be used more efficiently in the following example:

```smalltalk
FreeTypeCache clearCurrent.
2 timesRepeat: [
	(FormCanvas extent: 690@54)
		fillColor: Color white;
		drawString: String loremIpsum from: 1 to: 40 at: 5@0
			font: (LogicalFont familyName: 'Source Sans Pro' pointSize: 30)
			color: Color black.
	Smalltalk garbageCollect ].
(FreeTypeCache current instVarNamed: #fontTable) copy
```

Without the changes of this pull request, the glyphs from the first iteration are not reused in the second iteration: the ‘fontTable’ has two keys. While LogicalFont tries to reuse instances, that does not apply in this example because of the garbage collection. But the corresponding FreeTypeFont does not get collected as it is still in the FreeTypeCache, so it can also be reused when getting the `#realFont` of the new LogicalFont, given that FreeTypeFace already reuses the extant instance. Hence, with the changes of this pull request, the glyphs do get reused: the ‘fontTable’ has only one key.

### DIFF
--- a/src/FreeType/FreeTypeFont.class.st
+++ b/src/FreeType/FreeTypeFont.class.st
@@ -34,29 +34,13 @@ FreeTypeFont class >> forLogicalFont: aLogicalFont fileInfo: aFreeTypeFileInfoAb
 	index := aFreeTypeFileInfoAbstract index.
 	^aFreeTypeFileInfoAbstract isEmbedded
 		ifTrue:[
-			self
-				fromBytes: aFreeTypeFileInfoAbstract fileContents
-				pointSize: pointSize
-				index: index]
+			self new
+				setFace: (FreeTypeFace fromBytes: aFreeTypeFileInfoAbstract fileContents index: index) pointSize: pointSize;
+				yourself]
 		ifFalse:[
-			self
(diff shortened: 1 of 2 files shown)


Ducasse:
@jordanmontt 
This is really interesting I would like to see if we can spot instance creation difference with your tool.

Rinzwind:
I converted this pull request to a draft: the methods that are removed in commits 583907fbbe822467 and db96c6a1f9cc16bb do have uses in this repository, namely [in the documentation](https://github.com/pharo-project/pharo/blob/41431ea28c8e0b2685df9bb0a729d79e7f6fef39/doc/BoltsAndNuts/FreetypeFonts.md?plain=1#L32-L65), so they should be kept.

Rinzwind:
I closed this pull request and opened a new one: https://github.com/pharo-project/pharo/pull/15756

PR #15752, opened 2023-12-10
Fix #removeAllForFont: and #removeAllForType: on FreeTypeCache to support different scales

Rinzwind:
This pull request corrects pull request #14544: `#removeAllForFont:` and `#removeAllForType:` on FreeTypeCache were not adapted to take into account that the ‘type’ no longer maps directly to the ‘entry’ but to an intermediary ‘scaleTable’.

Without the changes of this pull request, the sends of `#removeAllForFont:` and `#removeAllForType:` in the following example each cause an Error to be signaled:

```smalltalk
font := (LogicalFont familyName: 'Source Sans Pro' pointSize: 30) realFont.
colorValue := Color black pixelValueForDepth: 32.
values := {
	[ FreeTypeCache current removeAllForFont: font ].
	[ FreeTypeCache current removeAllForType: (1 << 32) + colorValue ] }
		collect: [ :removeBlock |
			FreeTypeCache clearCurrent.
			1 to: 2 do: [ :scale |
				font glyphOf: $• colorValue: colorValue mono: false subpixelPosition: 0 scale: scale ].
			removeBlock on: Error do: #yourself ].
FreeTypeCache clearCurrent
```

### DIFF
--- a/src/FreeType-Tests/FreeTypeCacheTest.class.st
+++ b/src/FreeType-Tests/FreeTypeCacheTest.class.st
@@ -456,9 +456,11 @@ FreeTypeCacheTest >> testRemoveAllForFont [
 	| fifo |
 	fullCache maximumSize: nil.
 	1 to: 100 do:[:i |
-		fullCache atFont: font1 charCode: i type: 1 scale: 1 put: font1XGlyph].
+		fullCache atFont: font1 charCode: i type: 1 scale: 1 put: font1XGlyph.
+		fullCache atFont: font1 charCode: i type: 1 scale: 2 put: font1XGlyph].
 	1 to: 100 do:[:i |
-		fullCache atFont: font2 charCode: i type: 2 scale: 1 put: font1YGlyph].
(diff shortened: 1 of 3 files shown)


Ducasse:
Thanks Kris. What about have an "integration" test based on your snippet?

Rinzwind:
I’m not sure I understand what you mean. Note that commit fbb0b2a7b includes changes to `#testRemoveAllForFont` and `#testRemoveAllForType` that cause an error in those tests if the changes to `#removeAllForFont:` and `#removeAllForType:` are reverted.

Ducasse:
Yes I saw that you updated the tests.
Often there are tests that are ok but with little meaning because they test a value. 
And I was thinking that your code snippet was a nice addition as it conveys a scenario.

Rinzwind:
I added an extra test that tests the `#releaseCachedState` method of FreeTypeFont and is maybe a bit more meaningful than `#testRemoveAllForFont` (the method sends `#removeAllForFont:` to the cache).

PR #15764, opened 2023-12-11
Repackage layout definitions symbols

jecisc:
Those are used only to print the ldo format definition string so it should be packaged with the old definition printer and not in the kernel


MarcusDenker (approved):
Nice!

PR #15774, opened 2023-12-12
New Refactoring Driver for Renaming Shared Variables

carolahp:
## Summary
This PR creates a new driver, called RenameSharedVariableDriver, to handle the renaming of shared variables.
The new driver is used from Calypso's context menu item "rename2".

Complementary, the driver RenameInstanceVariable was refactored to simplify the implementation.
Finally, tests for both drivers were added.

## Details
In the first place I created an abstract class called RenameVariableDriver from which both drivers inherit.
Tests for the new driver were created, together with a new test for RenameInstanceVariableDriver.

This PR is the continuation of [PR](https://github.com/pharo-project/pharo/pull/15617)


src/Refactoring-Core/RBDirectlyDefinesInstanceVariableCondition.class.st
```diff
@@ -35,3 +35,14 @@ RBDirectlyDefinesInstanceVariableCondition >> instanceVariables: aColOfStrings [
 
 	instanceVariables := aColOfStrings
 ]
+
+{ #category : 'displaying' }
+RBDirectlyDefinesInstanceVariableCondition >> violationMessageOn: aStream [
```
carolahp, line 40: I think we should think about pulling this method up to RBNewAbstractCondition, like below

```
RBNewAbstractCondition 
    self subclassResponsibility
```

carolahp, line 40: I think we should think about pulling this method up to RBNewAbstractCondition, like below

```
RBNewAbstractCondition 
    self subclassResponsibility
```

src/Refactoring-Core/RBDirectlyDefinesSharedVariableCondition.class.st
```diff
@@ -33,3 +33,14 @@ RBDirectlyDefinesSharedVariableCondition >> check [
 RBDirectlyDefinesSharedVariableCondition >> sharedVariables: aCollection [ 
 	sharedVariables := aCollection
 ]
+
+{ #category : 'displaying' }
+RBDirectlyDefinesSharedVariableCondition >> violationMessageOn: aStream [
```
carolahp, line 38: Same as the comment above

src/Refactoring-Core/RBRenameClassVariableRefactoring.class.st
```diff
@@ -34,6 +34,21 @@ RBRenameClassVariableRefactoring class >> rename: aVarName to: aName in: aClass
...
 
+{ #category : 'preconditions' }
+RBRenameClassVariableRefactoring >> applicabilityPreconditions [
+
+	^ {
+		  (RBIsNotAMetaclass class: class).
+		  (RBIsValidSharedVariableName name: newName).
+		  (RBIsVariableNotDefinedInHierarchy name: newName class: class).
+		  (RBNameIsGlobalCondition new model: model className: newName) not.
+		  "For now only rename a shared variable that is locally defined in a class."
```
carolahp, line 45: @Ducasse @balsa-sarenac do you agree on adding this condition here and in the RenameInstanceVariableRefactoring?

carolahp (commented):
@Ducasse @balsa-sarenac I left some comments to discuss

Ducasse:
Hi caro

I'm working on RenameMethod and I have the impression that after we should be able to share some behavior in a common superclass. Right now I'm writing tests for the new conditions I have to convert.

src/Refactoring-Core/RBDirectlyDefinesInstanceVariableCondition.class.st
Ducasse: Yes!

src/Refactoring-Core/RBRenameClassVariableRefactoring.class.st
Ducasse: Ok I checked the existing code and yes this is really an applicabilityPreconditions

src/Refactoring-UI-Tests/RBRenameInstanceVariableDriverTest.class.st
```diff
@@ -22,6 +22,23 @@ RBRenameInstanceVariableDriverTest >> testInvalidName [
...
+				for: MyClassARoot
+				newName: #inexistantVariable1.
+	"Variable `inexistantVariable` is not defined in `MyClassARoot`"
+	self setUpDriver: driver.
+	
+	self should: [driver runRefactoring] raise: RBRefactoringError
+	
+]
+
 { #category : 'tests' }
```
Ducasse, line 42: I'm wondering if the driver should raise an exception or just inform the user. 
Now I'm wondering how by construction this could happen because the UI should only present some valid instance variable names to be renamed?

balsa-sarenac: The flow right now is that driver raises an exception, and then the exception informs the user.
I think it makes sense to remove even that exception handling from it, and only let it do UI notifications and stop with execution once it's done. Is this what you meant?

src/Refactoring-UI-Tests/RBRenameSharedVariableDriverTest.class.st
```diff
@@ -0,0 +1,74 @@
...
+	rbclasses := RBClassEnvironment classes: { MyClassARoot}.	
+	driver := RBRenameSharedVariableDriver new
+				scopes: { rbclasses }
+				variable: #Shared2
+				for: MyClassARoot
+				newName: #Shared2a.
+	self setUpDriver: driver.
+	driver runRefactoring.
+	"Renaming the references to the variable is not encapsulated in change objects, therefore the only change is the renaming of the variable itself."
+	self assert: driver refactoring changes changes size equals: 1
```
balsa-sarenac, line 72: This feels strange a bit. When does the renaming happen?

src/Refactoring-UI/RBRenameVariableDriver.class.st
```diff
@@ -0,0 +1,134 @@
...
+}
+
+{ #category : 'execution' }
+RBRenameVariableDriver >> applyChanges [
+	
+	refactoring privateTransform.
+	self flag: 'The next line is horrible, 
+					I must think of a better way to prevent changes from executing during tests'.
+	(previewPresenterClass isKindOf: StRefactoringPreviewPresenterMock class)
+		ifFalse: [ refactoring performChanges ]
```
balsa-sarenac: Maybe split this into two hooks, one for creating changes one for appying/performing them? Just a suggestion, hope it helps.

carolahp:
> Hi caro
> 
> I'm working on RenameMethod and I have the impression that after we should be able to share some behavior in a common superclass. Right now I'm writing tests for the new conditions I have to convert.

Hi @Ducasse , 
I agree, please let me know when you have the PR to check possible refactors to do

src/Refactoring-UI-Tests/RBRenameSharedVariableDriverTest.class.st
carolahp: The renaming happens in `RBRenameSharedVariableDriver>>applyChanges`. 
Your comment below helps regarding this :)

src/Refactoring-UI/RBRenameVariableDriver.class.st
carolahp: I propose [this](https://github.com/pharo-project/pharo/pull/15774/commits/497c89f41a5fceeb671c20de825f150435ecce75) because I couldn't find a better way to split the changes creation and performing them, while being able to prevent the application of changes during tests and at the same time, the dependency on StRefactoringPreviewPresenterMock. 

I don't like much the idea of using a presenter only to apply changes and never opening it. 
What do you think?

balsa-sarenac: I agree with you, but I think this new way is much better and that we can integrate and iterate later..

src/Refactoring-UI/RBRenameVariableDriver.class.st
```diff
@@ -0,0 +1,131 @@
...
+	],
+	#category : 'Refactoring-UI-Drivers',
+	#package : 'Refactoring-UI',
+	#tag : 'Drivers'
+}
+
+{ #category : 'execution' }
+RBRenameVariableDriver >> applyChanges [
+	"I use the preview presenter to perform changes, but I do not open it"
+	(self previewWithChanges: self changes) performChanges
```
balsa-sarenac: Does it make sense to use refactoring here `refactoring performChanges`?

carolahp: Yes it does, but if I use `refactoring performChanges`, changes will always be performed. I pass by the Preview Presenter because if the presenter is an instance of `StRefactoringPreviewPresenterMock` the changes won't be performed,  whereas if the presenter is an instance of `StRefactoringPreviewPresenter`, the changes will be performed

src/Refactoring-UI-Tests/RBRenameInstanceVariableDriverTest.class.st
Ducasse: Yes

src/Refactoring-UI/RBRenameVariableDriver.class.st
Ducasse: I did not look at the first driver tests I did but I wonder why I did not get the same problems.

Ducasse: I just read it and I do not see why you get this problem. 
Did you check the RemoveClassDriver?
Because call the applyChanges calls the presenter and there is no problem there, since it invokes the presenter
and do not open it and this is not a problem. We are not testing the previewPresenter here but the driver. 
So I do not care that it does not open I care that I can validate that a certain number and kind of changes are done. Does it help you?

```
runRefactoring
	
	self configureRefactoring.
	refactoring checkApplicabilityPreconditions.
	self setBreakingChangesPreconditions.
	haveNoReferences check & emptyClasses check & noSubclasses check
			ifTrue: [ self applyChanges ]
			ifFalse: [ self handleBreakingChanges ]
```

Ducasse: So I do not understand why you have to redefine applyChanges
In the superclass it is defined as 
```
applyChanges

	self openPreviewWithChanges: self changes
```

And to me it does the job.
So why do you redefine it.

Ducasse: @carolahp did you see this comment?

Ducasse:
@carolahp @balsa-sarenac I will integrate this PR even if I do not really understand why we need a preview that auto accept the changes.
From a test point of view we do not need that the changes are applied, we just need to test that the changes are produced. 
This is why in the first driver I wrote the Mock is only doing nothing because I can get access to the changes anyway
so I can write

```

t runRefactoring
and then after self assert: driver changes size equals: ....

```

Do it make sense to you?

PR #15775, opened 2023-12-12
8055-RGClassStrategy-is-the-last-caller-of-definitionWithSlots 

### LINKED ISSUE #8055: RGClassStrategy is the last caller of definitionWithSlots
So we should clean it probably call ClassDefinitionPrinter fluid for:....

MarcusDenker:
no senders anymore after the recent cleanup

fixes #8055


jecisc (approved):
Oh yes I missed this one during my cleaning

PR #15777, opened 2023-12-12
More small cleanups related to class variables / classbuilder

MarcusDenker:
A batch of small improvements related to the class builder cleanups


MarcusDenker:
I should remember to never do two PRs from the same image...

PR #15781, opened 2023-12-12
Position temp definition before first occurrence of extracted expression, and always in outer scope (fixes #14611)

### LINKED ISSUE #14611: Extract temporary variable doesn't correctly position temp definition

daniels220:
If the first occurrence of the expression to be extracted is in a child scope (a block) of the occurrence that was actually *selected*, we need to walk up the tree and put the initialization before the whole statement containing that child scope.

Add RBProgramNode>>statementNodeIn: to answer the statement *at a particular level of nesting/in a particular scope* that contains a node.

This should put #14611 fully to rest—the original issue was already fixed in a previous PR, but I thought of this scoping issue in a comment.

One implementation note—I saw that `RBMethodNode` understands `asSequenceNode` to return its body, so it seemed reasonable for `RBBlockNode` to do the same. It is a _little_ weird given that all nodes understand it and just _wrap_ themselves in a sequence node, but I don't think the two use-cases are likely to be confused.

### DIFF
--- a/src/AST-Core-Tests/RBParserTest.class.st
+++ b/src/AST-Core-Tests/RBParserTest.class.st
@@ -1759,6 +1759,11 @@ RBParserTest >> testQuerying [
 	arg1Node := tree whichNodeIsContainedBy: (102 to: 105).
 	self assert: arg1Node name equals: 'arg1'.
 	self assert: (arg1Node statementNode isMessage and: [ arg1Node statementNode selector value = #+ ]).
+	self assert: (arg1Node statementNodeIn: arg1Node methodOrBlockNode) equals: arg1Node statementNode.
+	self assert: (arg1Node statementNodeIn: arg1Node methodOrBlockNode body) equals: arg1Node statementNode.
+	self assert: (arg1Node statementNodeIn: tree) equals: tree body statements last.
+	self assert: (arg1Node statementNodeIn: tree body) equals: tree body statements last.
+	self should: [ arg1Node statementNodeIn: RBSequenceNode new ] raise: NotFound.
 	self assert: (arg1Node whoDefines: 'arg1') isBlock.
 	self assert: (aNode whoDefines: 'a') isMethod.
 	self assert: (aNode whoDefines: 'b') isSequence.
--- a/src/AST-Core/RBBlockNode.class.st
+++ b/src/AST-Core/RBBlockNode.class.st
@@ -125,6 +125,12 @@ RBBlockNode >> arguments: argCollection [
 	arguments do: [:each | each parent: self ]
 ]
 
+{ #category : 'converting' }
+RBBlockNode >> asSequenceNode [
+
+	^ body
+]
+
 { #category : 'accessing - token' }
 RBBlockNode >> bar [
 	^ bar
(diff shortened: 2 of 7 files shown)


Ducasse:
Thanks Daniel!

jecisc:
This seems to break some tests with undeclared variables reading

daniels220:
I see one failure and ten errors in the Jenkins output, but I can't see which tests those are—how do I see that? Is there something wrong with the CI config? This looks like maybe it tried to parse test results from somewhere and failed:

```
Recording test results
[Checks API] No suitable checks publisher found.
```

daniels220:
Took me a minute to find the list of what tests actually failed in Jenkins, but there we go. Turns out `RBBlockNode>>asSequenceNode` is a bad idea—easy enough to just inline the check I actually care about.

Re: #testThatThereAreNoSelectorsRemainingThatAreSentButNotImplemented, I was able to use existing selectors in the input method for the test...or would it be better to tag with the pragma?

daniels220:
Hmmm...no idea if the unix64 failure is related. I doubt it, since it didn't happen before and it's somewhere unrelated?

jecisc (approved):

PR #15791, opened 2023-12-14
Reduce users of ChangeSet

jecisc:
The change set browser got removed and we currently still have the change set model in the image. 

This PR removes code dealing with the management of the change set so that we can remove it entirely soon. 

It is mostly from Monticello to remove some code that was creating new change sets while loading code. Or also some code that was calling #removeFromChanges


jecisc:
I'm removing the last user here: https://github.com/pharo-spec/NewTools/pull/645

MarcusDenker (approved):

Ducasse:
Another youpi :)

PR #15794, opened 2023-12-14
Remove ChangeSet

jecisc:
This change removes ChangeSet and the classes that are going with it. 

This system is now duplicated with Epicea and all clients got removed. This should stress a little less the system because for each change in the code of the system, this was listening to the announcements to build changes and also registering to finalization. With this removed the compilation of code should be faster and we should stress less the finaliser. 

I am just keeping ChangeRecord because this one has nothing to do with ChangeSet. It is a way to check a method in the source file and gets its history. I'm not sure we should keep it in this particular package but this should be a future cleanup. There are also extensions on positioneable streams that I think can be cleaned up with some work but this is not for right now.


tinchodias:
wow

PR #15799, opened 2023-12-15
Move some debugger related code of Process to debugger packages

jecisc:
This removes some code from the Kernel that is used only in the debugging tools. 

Maybe we should merge Debugging-Core and Debugger-Model? Idk, sometimes I don't know what to put in what between those two


Ducasse:
Hi cyril how do you know that the behavior is not used by the unwind mechanism or ensure:, ifCurtailed:,...

jecisc:
I checked all senders one by one :)

StevenCostiou:
One remark (I will review the PR later I cannot now): I think this kind of change is good in the sense that we can isolate debugger code that should not be that deep embeded in the system core. 

Pushing it further, why do we have debugger code in the system core?
- we should be able to run a system without debugging support
- we should be able to completely isolate the debugger (in the large sense, not only the NewTool) and (un)load it when needed

I want to try to find a few months full time next year to investigate that and possibly start implementing changes.

StevenCostiou:
Implicit message @jecisc : wait next year before merging core and model so that we can discuss the design :P
I think they should be merged but perhaps some code should get its own package.

jecisc:
Ping

jecisc:
I updated the base

jecisc:
Hi, this PR is a green for a while, any chance to get a review? :) (It's not urgent but it was not an easy change to think about what should go where and I'm afraid I'll end up with a conflict that will be hard to resolve at some point :p)

PR #15808, opened 2023-12-17
Enh/refactoring pass over rename driver and tests

Ducasse:
- introduce new hooks for inform, requestDialog.
- remove the non interactive logic
- renamed lastName into previouslyProposedName 
- Fix the problem of renamechange not understanding asText
- Remove the ! of the printing of changes.
- Promote The rename driver as the only possibility (move calypso class method to the version 2 of the command) 

next step: 
- probably remove the newName: argument API and use in the test.
- check how we can bring assertion in some scenario (but isCancelled or other should work).
- factor a bit some test configuration.


src/Refactoring-UI-Tests/RBRenameInstanceVariableDriverTest.class.st
```diff
@@ -6,69 +6,108 @@ Class {
...
+	requestDialog on: #title: respond: 'nothing'.
+	requestDialog on: #text: respond: 'nothing'.
+	driver requestDialog: requestDialog.
+	driver informDialog: StInformDialogMock new.
 	
-	self should: [driver runRefactoring] raise: RBRefactoringError
-	
+
+	"I do not get why and how we can get a refactoring error?"
+	self should: [ driver runRefactoring ] raise: RBRefactoringError
```
balsa-sarenac, line 58: Based on the analysis, here is my assumption on why it's behaving this way:

This code previously was checking applicability preconditions (and they raised Refactoring error).
With your changes the error is never raised (we don't call `checkApplicabilityPreconditions`.

It then loops because we have precondition that variable should exist, which is never met/satisfied in this case. We check preconditions in a loop, and even though we supply good new name, the invalid name precondition is still false and it then asks for a new name and so on...

This is where we come to invariant preconditions (variable must exist) and applicability/input validation (name must be valid). Maybe in the end it makes sense to split these? Or maybe we should use reified preconditions in driver? (driver first checks part of preconditions - the invariant ones; driver then loops and checks other preconditions - input validation ones)

src/Refactoring-UI/RBRenameVariableDriver.class.st
```diff
@@ -32,43 +32,28 @@ RBRenameVariableDriver >> configureRefactoring [
...
-	"The only way to define newName is either using the UI (interactive mode) or using the initialization method >>scopes:variable:for:newName:"
-	self shouldNotImplement 
-]
-
 { #category : 'private' }
-RBRenameVariableDriver >> prepareRefactoring [
-	"The Driver is run in scripting mode, therefore the variable `newName` was defined in the method scopes:variable:for:newName:.
-	We don't need to ask for the name to the user"
-	refactoring newName: newName.
-	refactoring checkApplicabilityPreconditions
```
balsa-sarenac, line 46: (This code here was raising RBRefactoringError)

src/Refactoring-UI-Tests/RBRenameInstanceVariableDriverTest.class.st
Ducasse: Yes I shared this analysis while biking :) about the solution I do not know. 
Now the use of the Mock object helped for the loop.

PR #15811, opened 2023-12-18
Fix a few reflection: pragmas that weren't showing the right category

Inao0:
See the code to get actual category changes


Inao0:
I will redo this pull request from a newer version of Pharo to have fewer conflicts

PR #15819, opened 2023-12-20
Fix Enlumineur Pretty Printer "One line messages" setting not showing options

### LINKED ISSUE #15818: Enlumineur Pretty Printer "One line messages" setting not showing options
**Bug description**
In the Settings Browser there is a setting named "One line messages" which is not showing options.

**To Reproduce**
1. Open the Settings Browser (CMD+O+S)
2. Type One line message (to see missing behavior)
3. Type Multi line messages (to see correct behavior)

**Screenshot**
Screenshot 2023-12-20 at 11 18 57

hernanmd:
Make options visible in the "One line messages" setting, and behave like the "Multi-line messages".
Fixes #15818


Ducasse:
I think that the change is not correct because the messages should be a list of selectors and not a string of an array

hernanmd:
Stef with this fix the setting now is displayed as a list, like the multi-line setting:

Screenshot 2023-12-20 at 13 29 47

hernanmd:
I think the idea of the original creator of Enlumineur was to be able to provide "lists" of selectors, and since there is no easy way to edit multiple lists in the current Settings browser, it was easier to have arrays and then use "compiler evaluate: aStringWithArray" (just guessting why is like this).

What I could do is to open a new issue to use real collection and remove the compiler evaluation (I checked and it required a couple of additional changes).

Ducasse:
Indeed!

PR #15820, opened 2023-12-20
fix some button orders

estebanlm:
button "right order" (which is a convention) is a mess right now, we have decided to put all default actions to right so I am working on that.

### DIFF
--- a/src/Refactoring-UI/StItemsSelectionPresenter.class.st
+++ b/src/Refactoring-UI/StItemsSelectionPresenter.class.st
@@ -88,12 +88,10 @@ StItemsSelectionPresenter >> connectPresenters [
 
 { #category : 'initialization' }
 StItemsSelectionPresenter >> initializeDialogWindow: aModalPresenter [
+	
 	aModalPresenter
-		addButton: 'Refactor' do: [ :presenter | self accept: presenter ];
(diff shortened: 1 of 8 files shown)


Ducasse:
Thanks a lot :)

Ducasse:
Now esteban why we cannot just declare the button and the windowSpec or dialogSpec
decide that the ok button should be on the left or on the right.

estebanlm:
well, to be able to do that we first need to be able to *declare* which one of the buttons will contain the default action, and this is the purpose of this pull request (the replacement of the addButton:do: messages with addDefaultButton:do:).

Now, I also reordered the buttons (other than just change the name) for two reasons:

-  while is in my list to implement the automatic arrangement of default button I still didn't do it because I want to prevent any side effect for the fact none of the dialogs is prepared to have it at the moment, and 
- while automatic reorder is good, it is also good to have a visual confirmation of what you are going to see when the dialog shows (is a general good practice: as close to the visual-mind image you have of what you doing). So, I will eventually force the position of the button (just because then I can also switch the order with a setting), but why not to put the buttons in the right order anyway, specially if I am already modifying the method?

Ducasse:
Ahhhh super cool! I love it :)

estebanlm:
this is ready to merge :)

PR #15821, opened 2023-12-20
UndeclaredVariable read/write: context as argument

MarcusDenker:
We want to be able to re-lookup the name when reading an UndeclaredVariable. For that we need to know the class that we read or write the var.

This PR is a small step: change #runtimeUndeclaredWrite: and runtimeUndeclaredRead to have one arg more, the context where the access is done


jecisc:
Lot of tests are failing in SpCodePresenterTest

I don't know if this is this PR or the spec update of this afternoon

jecisc (approved):

Issue #15783, opened 2023-12-13
Senders in superclass


Ducasse:
SImilarly to implementors about a given method definition, I often need to find the senders of a given method in the classes upper in the inheritance chain above a given method. 
This is super handy when dealing with self send and identifying hooks in complex hierarchy.

PR #15822, opened 2023-12-20
Implements hierarchical senders system command

hernanmd:
Implements (from [this issue request](#15783)) hierarchical senders system command to find the senders of a given method in the classes in the inheritance chain of a that method.
The functionality could be accessed right-clicking a method in Calypso and selecting "Hierarchical senders" menu entry.
Add tests and baseline entry for loading test package.


jecisc:
The problem of this PR is that the new classes are in a "Tests" package tag and not in a package whose name ends with "Tests".

So there is no new package currently.

src/Calypso-SystemTools-Core/SycHierarchicalSendersCommand.extension.st
```diff
@@ -0,0 +1,28 @@
...
+	<classAnnotation>
+
+	^SycSourceCodeMenuActivation byRootGroupItemOrder: 1.2 for: ClyMethodSourceCodeContext
+]
+
+{ #category : '*Calypso-SystemTools-Core' }
+SycHierarchicalSendersCommand class >> sourceCodeShortcutActivation [
+	<classAnnotation>
+
+	^CmdShortcutActivation renamingFor: ClyMethodSourceCodeContext
```
dionisiydk: It will conflict with method rename command (cmd+r)

src/SystemCommands-MessageCommands/SycHierarchicalSendersCommand.class.st
```diff
@@ -0,0 +1,45 @@
+"
+This command find the senders of a given method in the classes upper in the inheritance chain above a given method.
+"
+Class {
+	#name : 'SycHierarchicalSendersCommand',
+	#superclass : 'SycChangeMessageSignatureCommand',
```
dionisiydk: It does look like senders is a kind of change signature command

src/SystemCommands-MessageCommands/SycHierarchicalSendersCommand.class.st
```diff
@@ -0,0 +1,45 @@
...
+{ #category : 'accessing' }
+SycHierarchicalSendersCommand >> defaultMenuItemName [
+
+	^ 'Hierarchical senders'
+]
+
+{ #category : 'execution' }
+SycHierarchicalSendersCommand >> execute [
+
+	Smalltalk tools messageList browse: self searchInTheWholeHierarchy.
```
dionisiydk: You should not implement it from scratch really. 
Calypso already provides a model for such queries. 
Just see how ClyShowLocalImplementorsCommand is implemented. It's really what you want here but for implementors. So for consistency name it ClyShowLocalSendersCommand

dionisiydk: I implemented it this way:

[ClyShowMessageLocalSenderCommand.txt](https://github.com/pharo-project/pharo/files/13733054/ClyShowMessageLocalSenderCommand.txt)

(load into image by renaming file to .st)

Ducasse:
Thanks this is cool.

src/SystemCommands-MessageCommands/SycHierarchicalSendersCommand.class.st
hernanmd: I want to use the command without depending on Calypso, is that possible with your implementation?

src/SystemCommands-MessageCommands/SycHierarchicalSendersCommand.class.st
hernanmd: The class documentation of SycChangeMessageSignatureCommand says: "I am a base class for commands which modify signature of given message". However I don't see how querying the hierarchical senders would change anything. Could you elaborate?

hernanmd:
> The problem of this PR is that the new classes are in a "Tests" package tag and not in a package whose name ends with "Tests".
> 
> So there is no new package currently.

Thanks. I updated the PR with the package.
Failing tests are unrelated. Windows build didn't finished yet.

Ducasse:
I'm not sure that showLocalSenders is better than ShowSendersInHierarchy if it is does show senders in hierarchy :).

I will integrate this issue so that we can continue in another one.

src/Calypso-SystemTools-Core/SycShowLocalSendersCommand.extension.st
```diff
@@ -0,0 +1,31 @@
...
+
+	^SycSourceCodeMenuActivation byRootGroupItemOrder: 1.2 for: ClyMethodSourceCodeContext
+]
+
+{ #category : '*Calypso-SystemTools-Core' }
+SycShowLocalSendersCommand class >> sourceCodeShortcutActivation [
+	<classAnnotation>
+
+	^ CmdShortcutActivation 
+		by: $v meta
```
dionisiydk: I wonder what impact it has. On mac cmd+v is a classic Paste from clipboard command.
So I guess such shortcut here will be fully ignored as it is intercepted inside internals of the text editor. Or the worse case it will break a copy paste behaviour

src/SystemCommands-MessageCommands/SycShowLocalSendersCommand.class.st
```diff
@@ -0,0 +1,81 @@
...
+SycShowLocalSendersCommand >> isComplexRefactoring [
+
+	^ false
+]
+
+{ #category : 'execution' }
+SycShowLocalSendersCommand >> prepareFullExecutionInContext: aToolContext [
+
+	super prepareFullExecutionInContext: aToolContext.
+	selectedMessage := aToolContext lastSelectedMessage.
```
dionisiydk: base class already defines #messages variable retrieved from the context

src/SystemCommands-MessageCommands/SycHierarchicalSendersCommand.class.st
dionisiydk: Well if you don't use query model of Calypso you loose a lot of functionality and you duplicate the query logic which is already implemented.
For example:
- there is no indication that the message browser is in a scoped mode (scope of a hierarchy)
- there is no auto update when new senders appears in the system

src/SystemCommands-MessageCommands/SycShowLocalSendersCommand.class.st
hernanmd: I don't understand, could you explain what you mean?

dionisiydk: Look into the base class SycMessageCommand>>prepareFullExecutionInContext:. It reads all selected #messages from the context (in browser you can select multiple methods and query local senders of all of them).
Thus you don't really need new inst vars in your command as all data is already there:
- #messages are instances of SycMessageDescription which has #methodClass of a method where the message is used.

I see there is a bug in how SycMessageDescription is created from the source code context. But it's easy to fix:
```Smalltalk
SycMessageDescription>>ofMessageNode: astMessageNode
	^(self fromContextOf: astMessageNode methodNode)
		selector: astMessageNode selector;
		argumentNames: (astMessageNode arguments collect: [:each | each formattedCode])
```
The var #contextUser in SycMessageDescription is quite cryptic. The idea was to keep there a kind of a scope where the message is used. It would be nice to rename it to something more meaningful. Maybe it should be just #usingMethod. The constructor #ofMessageNode: can easily extract a full method from given node.

src/SystemCommands-MessageCommands/SycHierarchicalSendersCommand.class.st
dionisiydk: And more interesting is scopes in spawned message browser and tab based navigation when you request local sender from another message browser.
For example try open Inheritance for any method. This is analog of local senders but for implementors. You can see that you can see that the current scope is "Hierarchy" and you can switch it back to full image or other options.
With your approach the message browser just shows given list of methods. No scopes are available.

hernanmd: Thanks for pointing out, I will open another issue to check those issues you mention.

Ducasse:
This is strange because I do not see these changes in the image I just downloaded.

hernanmd:
I just downloaded the latest image and I could see the changes:

Screenshot 2023-12-27 at 22 48 27

So if you could see it now #15783 can be closed

Ducasse:
Yes I will close it.

PR #15827, opened 2023-12-21
Undeclared: inlining of declare:from:

MarcusDenker:
This PR finishes inlining of declare:from:

(see https://github.com/pharo-project/pharo/issues/11903 )


jecisc (approved):
Nice one! I like that this gets removed from dictionaries

PR #15828, opened 2023-12-21
Display trait icon in Calypso class definition editor

### LINKED ISSUE #14840: When a Trait is selected the tab icon not corresponding to a Trait in the source-code browser
**Bug description**
Select a Trait in the source-code browser and see the icon in the tab below.

**Expected behavior**
Same icon green "T" as in the classes list.

**Screenshots**
image

**Version information:**
 - OS: Windows 11 Pro
 - Pharo-11.0.0+build.707.sha.f720787cab0b25f0a8a811d66f2a8d3633ad8ba3 (64 Bit)

ISayarLu:
Fixes #14840

### DIFF
--- a/src/Calypso-SystemTools-Core/ClyClassDefinitionEditorToolMorph.class.st
+++ b/src/Calypso-SystemTools-Core/ClyClassDefinitionEditorToolMorph.class.st
@@ -98,7 +98,9 @@ ClyClassDefinitionEditorToolMorph >> decorateContainerTab [
 
 { #category : 'initialization' }
 ClyClassDefinitionEditorToolMorph >> defaultIconName [
-	^#class
+
+	self editingClass isTrait ifTrue: [ ^ #trait ].
+	^ #class
 ]
 
 { #category : 'to sort' }


jecisc (approved):

jecisc:
Maybe we could do better and ask the class for its icon? So if it's a presenter or a collection it display the icon specific for this kind of class?

astares:
Yes - this would also use custom icons then. Typically this is what #systemIcon is returning

Ducasse:
I integrated it now and we can open another issue.

StevenCostiou:
I would not bother much with Calypso if a spec browser replaces it soon.
We did a quick fix with Imen so that it works as expected, and a better implementation should reproduce the same behavior in the new spec browser.

But for example, we have no guarantee that the expected class cannot be nil in Calypso code (we did not find tests).
So delegating to the class in the current code could produce DNU because of a nil reference and we did not want to do a nil check.

Ducasse:
Agreed!

astares:
@StevenCostiou Is there any WIP going on for a Spec based browser to replace Calypso?

Ducasse:
Yes esteban is working on it.

PR #15829, opened 2023-12-21
Another pass on rename variable driver with caro

Ducasse:


Ducasse:
@balsa-sarenac @carolahp The last commit is important because it fixes the semantics of preconditions (returning a list of condition instead of a condition) in conformance with failed and checkApplicabilityPreconditions. Please have a look and let me know.

Ducasse:
I need help because I do not get at all why the build is failing. I just committed normal code. So may be this is something that happened before but I could not see what is it.

Ducasse:
```
[Pipeline] }
[Pipeline] // stage
[Pipeline] }
[Pipeline] // node
[Pipeline] End of Pipeline
ERROR: No artifacts found that match the file pattern "build/bootstrap-cache/*.zip,build/bootstrap-cache/*.sources". Configuration error?

GitHub has been notified of this commit’s build result

Finished: FAILURE
```

guillep:
Hi Stef, a download failed during the bootstrap, so the build just has to be relaunched (I did it)

Ducasse:
Thanks I did it several times but it was always failing. Now it is passing so this is cool

Ducasse:
None of the broken tests are related to the issue so I merge it. I will try to fix some of the tests in another PR.

src/Refactoring-Core/RBRemoveInstanceVariableRefactoring.class.st
```diff
@@ -58,6 +60,8 @@ RBRemoveInstanceVariableRefactoring >> generateChanges [
 			self breakingChangePreconditions errorString ].
 
 	self privateTransform.
+	self flag: #todo.
+	"the method comment does not indicte this. May be we can add self changes but to all the generateChanges methods. Be consistent. "
 	^ self changes
```
balsa-sarenac, line 65: @Ducasse I agree with this. I was looking of usages of `generateChanges`:
- half of the test are using `refactoring model changes changes` immediately after calling `generateChanges`, so it might be good for this
- another half are not relying on them, but we can ignore the return value in that situation
- in Drivers we almost always need `changes` after calling `generateChanges`
- in Refactorings/Transformation that don't have the driver, we usually don't need `changes` after calling `generateChanges`

So I think it makes sense to return the changes and be consistent everywhere, even though in some places they are not required, but can be safely ignored.

PR #15836, opened 2023-12-22
Cleanups in UIManager dependencies

guillep:
This PR is a step forward in cleaning UIManager dependencies from everywhere.

On the one hand, this implements a configurable exception handler per thread, with a global (configurable) default. Care has been taken to keep it backward compatible: an exception handler distinguishes Exceptions, Notifications, and Warnings.
UIManager methods have been renamed to follow this protocol and be used as error handlers.

On the other hand, I've pulled down dependencies to UI manager to their right users.
It turns out that many UIManager users only really work with a headful morphic UI manager.
In those cases, using the more general `UIManager` API forces all other subclasses to answer messages that do not make sense (e.g., open merge dialog, open an editable text area...).
In those cases, I've pulled down the dependency and make them, when possible, use directly the MorphicUIManager.
MorphicUIManager could be later replaced by a spec related solution if required, while we could shave the CommandLine/NonInteractive UIManagers to the bare minimum responsibilities.
This also clarifies when clients really need a UI.

I've also cleaned dead code around here and there.

In my Image, I have now 
- 45 references to UIManager
- 64 references to MorphicUIManager

Originally there were ~250 references

Extra efforsts in:
- [ ] libgit https://github.com/pharo-vcs/libgit2-pharo-bindings/pull/79


Ducasse:
- Did not build but I like the changes.
- About unhandledError I checked and it was like that in Squeak since decades. (not saying that this is good but this is the UIManager that introduced it).

Ducasse:
```

NewUndeclaredWarning: WorldState class>>quitSession (MorphicUIManager is Undeclared)
NewUndeclaredWarning: WorldState class>>saveAs (MorphicUIManager is Undeclared)
MetacelloNotification: Loaded -> Morphic-Core-tonel.1 --- tonel:///builds/workspace/est_and_branch_Pipeline_PR-15836/64/src --- cache
MetacelloNotification: Evaluating -> baseline [BaselineOfMorphicCore] >> postload:package:
UndeclaredVariableRead: Attempt to read undeclared variable MorphicUIManager
UndeclaredVariable>>runtimeUndeclaredReadInContext:
DisplayScreen>>newDepthNoRestore:
DisplayScreen>>newDepth:
BaselineOfMorphicCore>>postload:package:
[ :aLoader :pkgSpec | self project configuration perform: selector with: aLoader with: pkgSpec ] in MetacelloVersionSpec(MetacelloSpec)>>doItBlock: in Block: [ :aLoader :pkgSpec | self project configurati[..]
```

guillep:
Yes, I'm working on this one right now :)

src/Graphics-Display Objects/DisplayScreen.class.st
```diff
@@ -271,19 +272,6 @@ DisplayScreen >> replacedBy: aForm do: aBlock [
 	ScreenSave := nil
 ]
 
-{ #category : 'other' }
-DisplayScreen >> restore [
```
Ducasse, line 263: Wouahhhhh there were reference from DisplayScreen to MorphicUIManager. O_o

guillep: It's the global callback design pattern :)

Ducasse: Yes. And you can build centralized monolithic application pattern :)

guillep:
CI is good now. Broken tests are not related.

Ducasse:
Cool!

PR #15842, opened 2023-12-24
Add Search filter to Epicea

hernanmd:
This PR add a search text input presenter to filter the epicea entries in the log browser. 
And make it noticeable so people can access fast.

The search feature dynamically update the list entries to match the entered text at each key press.
If there are filters applied, then search in the filtered entries (cleaning the filters will make searching to the complete entries).

Check the following video to see it in action:

https://github.com/pharo-project/pharo/assets/4825959/78c7d725-feee-4051-9163-9f0df6bdb1f5


Ducasse:
Thanks this is cool.
In VW they had a search and when you selected an item you could say click on a combo box 
for selector, class, type (class definition, methods,)
Then you could change the searched text and keep the combo box selections and filter based on them. 

A typical scenario
- look for Po
- Pick up one method with Point>>funkyName and select selector 
- change Point -> Pilou
- and you would see all the methods named funkyName whose class, package contains Pilou

So it was super fast to navigate several spaces.

Ducasse:
I will integrate this PR when the build pass but the scenario I mentioned was quite cool.

src/Epicea/TestCase.extension.st
```diff
@@ -1,14 +1,14 @@
 Extension { #name : 'TestCase' }
 
 { #category : '*Epicea' }
-TestCase >> shouldLogWithEpicea [
+TestCase class >> shouldLogWithEpicea [
+	"When running a test, we disable the loggings of Epicea to run the tests silently. In case a TestCase explicitly wants to Epicea logs, then I can be overriden to return true and enable the logs."
 
-	^ self class shouldLogWithEpicea
+	^ false
```
Ducasse, line 7: Why do you disable this feature?
May be you did it in your image for you?

hernanmd: Mmm I don't remember to changed that code.
I will review the changes now

hernanmd: It could be an automated refactoring?
I've seen a lot of them automatically applied so I have to uncheck changes in Iceberg

jecisc: It’s just the instance and class side methods of the same name that got inverted. That happens sometimes and seems to be a bug in tonel. I’ve it on my todo list to check it but never got the time. So there is no problem here

Ducasse:
I do not think that there are automatic refactorings applied on save.

guillep:
@Ducasse I think he meant deprecations :)

Anyways, probably it was what @jecisc says

Ducasse:
@hernanmd can you do a final pass so that we integrate it?

hernanmd:
> @hernanmd can you do a final pass so that we integrate it?

Done, just fixed a small error when writing to search box if no log was selected.

Ducasse:
Can you handle 

```
stCase class >> shouldLogWithEpicea [
	"When running a test, we disable the loggings of Epicea to run the tests silently. In case a TestCase explicitly wants to Epicea logs, then I can be overriden to return true and enable the logs."

	^ self class shouldLogWithEpicea
	^ false
```

Ducasse:
Tx!

Ducasse:
Oops may be it was a mistake because we have 99 tests failing :(. 
I do not really get this feature.

hernanmd:
Ok, I reverted the commit. 

And I checked and this is to prevent logging of generated methods in TestCases

https://github.com/pharo-project/pharo/assets/4825959/ba0fe0ac-67a2-4c3c-99fa-a464c749bd1f

Ducasse:
Thanks Can you open an issue so that we do not forget to discuss it with Cyril.

hernanmd:
> Thanks Can you open an issue so that we do not forget to discuss it with Cyril.

Ok, done #15875

Issue #15837, opened 2023-12-23
Metacello broken


theseion:
**Bug description**
This PR #15504 (@guillep) has broken SmalltalkCI for Pharo12. In the following method, the `#get` send to a `Metacello` fails:

```smalltalk
basicLoadProjectOn: aSmalltalkCI
	| metacello metacelloPassword metacelloUser |
        "..."
	metacello := (Smalltalk at: #Metacello) new.
        "..."	
	metacello copy get. "----> fails"
```

The failing code is in `MetacelloScriptEngine>>#get`:

```smalltalk
get
  " load a fresh copy from repo"

  | spec projectPackage |
  MetacelloProjectRegistration
    copyRegistryRestoreOnErrorWhile: [ 
      self setDefaultsAndValidateProjectSpec.
      spec := self projectSpec.
      projectPackage := spec projectPackage.
      projectPackage repositorySpecs
        do: [ :repoSpec | repoSpec createRepository flushForScriptGet ].
      projectPackage load. "---> fails"
      "..."
```

**To Reproduce**
This is easily reproducible by cloning SmalltalkCi from https://github.com/hpi-swa/smalltalkCI and running:
```bash
./run.sh -s Pharo64-alpha --headful .smalltalk.ston
```
**Version information:**
 - Pharo Version: 12

jbrichau:
Breaks all builds using Smalltalk-CI.. e.g. https://github.com/SeasideSt/Grease/actions/runs/7308255921/job/19914737832



<blockquote><img src="https://opengraph.githubassets.com/e649d8a5dafdb24b5ff1e544ea5a0133f9f1ee629a351f215054ae43e44b3af0/SeasideSt/Grease" width="48" align="right"><div><img src="https://github.githubassets.com/favicons/favicon.svg" height="14"> GitHub</div><div><strong><a href="https://github.com/SeasideSt/Grease">Add GRPlatform>>sizeOfFile: for older Pharo versions · SeasideSt/Grease@e99e20d</a></strong></div><div>The Grease Portability Library . Contribute to SeasideSt/Grease development by creating an account on GitHub.</div></blockquote>

Ducasse:
Thanks for the feedback.

PR #15504, opened 2023-11-21
Metacello cleanup

guillep:
This PR cleans up several things in Metacello code with the objective of easily changing the way code is loaded.

 - remove ambiguities between specs and directives. specs are loaded, then transformed in directives.
   - a record operation works on specs and generates directives
   - a fetch/load operation works on directives and loads files/code
 - remove duplicated code when loading packages
 - inline methods and remove unnecessary double dispatch to largely simplify code

Also, sneak in some fixes in circular dependencies between unordered and sequenceable collections

I tested that the following projects work
- mutalk
- xml parser
- kbg atomic
- seaside
- geometry
- artefact

PR #15845, opened 2023-12-25
Metacello cleanup

guillep:
Rework of https://github.com/pharo-project/pharo/pull/15504 because of https://github.com/pharo-project/pharo/issues/15837.

This PR cleans up several things in Metacello code with the objective of easily changing the way code is loaded.

 - remove ambiguities between specs and directives. specs are loaded, then transformed in directives.
   - a record operation works on specs and generates directives
   - a fetch/load operation works on directives and loads files/code
 - remove duplicated code when loading packages
 - inline methods and remove unnecessary double dispatch to largely simplify code

Also, sneak in some fixes in circular dependencies between unordered and sequenceable collections

I tested that the following projects work
- mutalk
- xml parser
- kbg atomic
- seaside
- geometry
- artefact

Missing:
 - [ ] test smalltalk CI dependencies


guillep:
Closed in favor of https://github.com/pharo-project/pharo/pull/15846

PR #15854, opened 2023-12-27
Fixed obvious unnecessary users of duration.

Ducasse:
- Next step the valueWithin: onTimeout:


Ducasse:
Hi @guillep
I will wait for your review before merging because it is a bit sensitive

src/Announcements-Core-Tests/AnnouncerTest.class.st
```diff
@@ -93,7 +93,7 @@ AnnouncerTest >> testAnnouncingReentrant [
 		ifTrue: [ ok := true ]
 	] for: self.
 
-	self should: [ announcer announce: Announcement new. ] notTakeMoreThan: 1 second.
+	self should: [ announcer announce: Announcement new. ] notTakeMoreThanMilliseconds: 60.
```
guillep, line 96: This is not equivalent. 1 second is 1000 milliseconds :)

src/DrTests-TestsProfiling/DTTestsProfilerVisitor.class.st
```diff
@@ -33,12 +33,13 @@ DTTestsProfilerVisitor >> visitTestCase: aTestCase [
...
-
+	"We make explicit the dependency to duration"
+	
 	| aTestresult |
 	self profilingData
 		add:
 			(DTTestCaseProfilingData
 				testCase: aTestCase
-				duration: [ aTestresult := aTestCase run ] timeToRun
+				duration: (Duration milliSeconds: ([ aTestresult := aTestCase run ] millisecondsToRun))
```
guillep, line 43: here I don't see why it would be bad to use `timeToRun`. I don't think  `timeToRun` is bad per-se.

Here for example, we already need a duration, so we are not introducing an extra dependency and the code reads nicer...

src/Kernel-CodeModel/BlockClosure.class.st
```diff
@@ -384,6 +384,30 @@ BlockClosure >> method [
 	^ self compiledBlock
 ]
 
+{ #category : 'private timing' }
+BlockClosure >> microsecondClockValue [
+	"Answer the number of microseconds since Epoch"
+
+	^self primUTCMicrosecondsClock
```
guillep, line 391: Why here we don't use `Time`?

src/Kernel-CodeModel/BlockClosure.class.st
```diff
@@ -568,6 +592,17 @@ BlockClosure >> outerContext: ctxt [
...
 
+{ #category : 'private timing' }
+BlockClosure >> primUTCMicrosecondsClock [
+	"Answer the number of micro-seconds ellapsed since epoch.
+	That is since 00:00 on the morning of January 1, 1901 UTC.
+	At least a 60-bit unsigned integer is used internally which is enough for dates up to year 38435.
+	Essential. See Object documentation whatIsAPrimitive. "
+
+	<primitive: 240>
+	self primitiveFailed
```
guillep, line 603: Same, why not having a dependency to Time here?

src/Kernel-CodeModel/BlockClosure.class.st
```diff
@@ -384,6 +384,30 @@ BlockClosure >> method [
...
 
+{ #category : 'private timing' }
+BlockClosure >> microsecondClockValue [
+	"Answer the number of microseconds since Epoch"
+
+	^self primUTCMicrosecondsClock
+]
+
+{ #category : 'private timing' }
+BlockClosure >> microsecondsToRun [
```
guillep, line 395: And maybe this method should belong to the same package as `BlockClosure>>bench` (I'd create maybe later a new package?)

src/Kernel-CodeModel/BlockClosure.class.st
```diff
@@ -993,6 +1028,40 @@ BlockClosure >> valueWithin: aDuration onTimeout: timeoutBlock [
 				ifFalse:[ e pass]]
 ]
 
+{ #category : 'evaluating' }
+BlockClosure >> valueWithinMilliseconds: milliseconds onTimeout: timeoutBlock [
```
guillep, line 1032: I'm not convinced this method should be in Kernel :) But I understand it was before like this already.

src/Kernel-Tests-Extended/ValueWithinFixTest.class.st
```diff
@@ -9,7 +9,7 @@ Class {
 	#tag : 'Processes'
 }
 
-{ #category : 'tests' }
+{ #category : 'tests - to be packaged with duration package' }
```
guillep, line 12: haha :)

src/Kernel-Tests-Extended/ValueWithinFixTest.class.st
```diff
@@ -26,6 +26,22 @@ ValueWithinFixTest >> testValueWithinNonLocalReturnFixReal [
...
+	"The real test for the fix is just as obscure as the original problem"
+
+	| startTime |
+	self valueWithinNonLocalReturn.
+	startTime := [] microsecondClockValue.
+	[[] repeat] valueWithinMilliseconds: 100 onTimeout:[ | deltaTime |
+		"This *should* timeout after 100 msecs but the pending process from
+		the previous invokation will signal timeout after 20 msecs already
+		which will in turn cut this invokation short."
+		deltaTime := [] microsecondClockValue - startTime.
```
guillep, line 39: Super strange to ask the clock to the closure.

src/Kernel/Delay.class.st
```diff
@@ -60,12 +60,26 @@ Delay class >> delaySchedulerClass: aSchedulerClass [
 	self inform: 'Delay scheduler set to ' , aSchedulerClass printString
 ]
 
+{ #category : 'instance creation' }
+Delay class >> forDays: aNumber [
+	"Return a new Delay for the given number of Days"
+	
+	^ self forMilliseconds: aNumber * 1000 * 60 * 60 * 24
```
guillep, line 67: Hmm, shouldn't this use the constants in `ChronologyConstants`? It makes sense that `Delay` depends on that

src/Kernel/Delay.class.st
```diff
@@ -60,12 +60,26 @@ Delay class >> delaySchedulerClass: aSchedulerClass [
...
 Delay class >> forDuration: aDuration [
  	"Return a new Delay for the given duration."
  	^ self forMilliseconds: aDuration asMilliSeconds
 ]
 
+{ #category : 'instance creation' }
+Delay class >> forHours: aNumber [
+	"Return a new Delay for the given number of Hours"
+	
+	^ self forMilliseconds: aNumber * 1000 * 60 * 60
```
guillep, line 80: same here

src/Kernel/Delay.class.st
```diff
@@ -75,6 +89,13 @@ Delay class >> forMilliseconds: aNumber [
 	^ self new setDelay: aNumber forSemaphore: Semaphore new
 ]
 
+{ #category : 'instance creation' }
+Delay class >> forMinutes: aNumber [
+	"Return a new Delay for the given number of Minutes"
+	
+	^ self forMilliseconds: aNumber * 1000 * 60
```
guillep, line 96: and here

src/SUnit-Core/TestAsserter.class.st
```diff
@@ -531,12 +531,12 @@ TestAsserter >> should: aBlock description: aString [
 	self assert: aBlock value description: aString
 ]
 
-{ #category : 'asserting' }
+{ #category : 'to be packaged in duration package' }
```
guillep, line 534: ;)

guillep (approved):
I have some comments to discuss!

src/Announcements-Core-Tests/AnnouncerTest.class.st
Ducasse: :) I will fix this.

src/DrTests-TestsProfiling/DTTestsProfilerVisitor.class.st
Ducasse: I wanted to deprecate timeToRun and to make explicit the dependency from DrTest to the future Duration package.

src/Kernel-CodeModel/BlockClosure.class.st
Ducasse: I thought that we want to get Time out of the kernel.
So I copied the primitive there.

src/Kernel-CodeModel/BlockClosure.class.st
Ducasse: I thought that we do not want to have a dependency to Time from the kernel.

src/Kernel-Tests-Extended/ValueWithinFixTest.class.st
Ducasse: Indeed. Now this is a test of closure and if we want to run these tests without the time package then I do not see how we can do it.

src/Kernel/Delay.class.st
Ducasse: Yes it would be better.  Now we can do it in a second turn. Because it would be good to have it run and then create extra package.

Ducasse:
I feel like I will need help. I tried to repair and pharo disappeared - not sure that I can reproduce it. I think that the fact that I share repo is making things more complex when working on long branches. I have to repair each time pharo changes or that I work on Pharo. I will change this set up because it brings a lot of noise and pain. Tell me if I'm wrong.

src/DrTests-TestsProfiling/DTTestsProfilerVisitor.class.st
guillep: But I like timeToRun!!! Just not in the kernel!

src/Kernel-CodeModel/BlockClosure.class.st
guillep: @Ducasse to me the answer to your comments above is here. I want timeToRun & co out of the kernel too. They should be packaged elsewhere, maybe together with Time in first instance.

src/Kernel-Tests-Extended/ValueWithinFixTest.class.st
guillep: Maybe value with timeout does not need to be kernel? How is it used? I'll check

src/DrTests-TestsProfiling/DTTestsProfilerVisitor.class.st
Ducasse: Here it was really syntactical sugar and make the dependency analysis more difficult. 
This is why I switch it. And it does not remove the dependency of DrTest to Duration.

src/Kernel-CodeModel/BlockClosure.class.st
Ducasse: my goal was to have timeToRun in a Duration Package that is on top of Time's package.

src/Kernel-Tests-Extended/ValueWithinFixTest.class.st
Ducasse: Yes this too. Now having a version with milliseconds was easy and I like to have it this way people can choose what they want and duration is optional.

Ducasse:
On one of the three platforms the tests fully pass.

src/DrTests-TestsProfiling/DTTestsProfilerVisitor.class.st
tinchodias: I agree with guille in timeToRun

tinchodias:
In Bloc "ecosystem" I see 21 methods sending it.

I used `Time millisecondClockValue` in benchmark suites, the pattern in general is:
```
startMS := Time millisecondClockValue.
"do things"
durationMS := Time millisecondClockValue - startMS.
```
but the durationMS often happens in other method.

src/DrTests-TestsProfiling/DTTestsProfilerVisitor.class.st
Ducasse: Me too. Now I repeat it because may be I was unclear. 
In DrTest we can have a dependency to Duration but 
- I wanted to make this explicit hence I refer to the class Duration
- I wanted to make sure that we do not use timeToRun in BlockClosure.

Ducasse:
My point is that often duration is often not useful because it just put a millisecond number into a newly created duration and from that duration object we retrieve the milliseconds!

Ducasse:
@guillep I do not know how to resolve the conflict. 
Now I'm thinking that once the conflict is resolved we can merge and we can address your question in another PR.

guillep:
It must be resolved in iceberg or in the command line... The thing is that you touched Duration while another PR (mine?) moved it to another package. So the github UI will not help there...

Ducasse:
I merge now and we can do another iteration to handle guille's comments.

PR #15859, opened 2023-12-27
introduce a display method in Morph and Morph class to reduce the num…

Ducasse:
…ber of direct reference to one of our last globals

I doubt that it will work but let us see.

### DIFF
--- a/src/Athens-Examples/AthensTigerWindow.class.st
+++ b/src/Athens-Examples/AthensTigerWindow.class.st
@@ -4,6 +4,7 @@ I'm an example that shows a tiger that moves via rotation, translation and zoom 
 Class {
 	#name : 'AthensTigerWindow',
 	#superclass : 'StandardWindow',
+	#classTraits : '{} + TraitedClass',
 	#instVars : [
 		'surface',
 		'scale',
--- a/src/Athens-Morphic/AthensWorldMorph.class.st
+++ b/src/Athens-Morphic/AthensWorldMorph.class.st
@@ -31,8 +31,8 @@ AthensWorldMorph >> restoreMorphicDisplay [
 	DisplayScreen startUp.
 
 	self
-		extent: self display extent;
-		viewBox: self display boundingBox;
+		extent: self displayInstance extent;
+		viewBox: self displayInstance boundingBox;
 		handsDo: [:h | h visible: true; showTemporaryCursor: nil];
 		resizeBackgroundMorph;
 		submorphsDo: [ :each | each displayExtentChanged ];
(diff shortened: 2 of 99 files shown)


Ducasse:
I will rename display in displayInstance because display is already defined to display the receiver in the class DisplayObject.

Ducasse:
And I would love to rename DisplayObject into DisplayableObject

src/Athens-Examples/AthensTigerWindow.class.st
```diff
@@ -4,6 +4,7 @@ I'm an example that shows a tiger that moves via rotation, translation and zoom
 Class {
 	#name : 'AthensTigerWindow',
 	#superclass : 'StandardWindow',
+	#classTraits : '{} + TraitedClass',
```
guillep, line 7: Hmm, we should fix this, this is the default class trait

guillep: (not in this PR)

guillep (approved):
Looks ok, now I really ask myself why are so many users also.

Ducasse:
```
MetacelloNotification: Loaded -> MenuRegistration-tonel.1 --- tonel:///builds/workspace/est_and_branch_Pipeline_PR-15859/64/src --- cache
MetacelloNotification: Loading -> Fuel-Core-tonel.1 --- tonel:///builds/workspace/est_and_branch_Pipeline_PR-15859/64/src --- cache
MetacelloNotification: Loaded -> Fuel-Core-tonel.1 --- tonel:///builds/workspace/est_and_branch_Pipeline_PR-15859/64/src --- cache
Error: The following definitions had errors while loading.  Press Proceed to try to load them again (they may work on a second pass):
  AlphaImageMorph
  ThreePhaseButtonMorph
...
```

I would love to know how we can access to the errors.

src/Athens-Examples/AthensTigerWindow.class.st
Ducasse: Do you have an idea where it comes from because I did not touch this part.

guillep: Yes, it's the fix we did the other day to Traits

Ducasse: Yes I was connecting after...:)

Ducasse:
I think that I will need some help. And I also wonder why so many domain objects (outside of this PR) needs to access Display.

Ducasse:
I close it because we redid it differently.

PR #15865, opened 2023-12-27
Remove some UIManager dependencies from Compression package

hernanmd:
Remove UIManager dependency from GZipReadStream.
Use Spec to open a file viewer of uncompressed contents.

### DIFF
--- a/src/Reflectivity-Tests/MetaLinkAnonymousClassBuilderTest.class.st
+++ b/src/Reflectivity-Tests/MetaLinkAnonymousClassBuilderTest.class.st
@@ -70,17 +70,19 @@ MetaLinkAnonymousClassBuilderTest >> testClassAccessFromClassBuilder [
 
 { #category : 'tests' }
 MetaLinkAnonymousClassBuilderTest >> testCompileClassAccessorForAnonymousClass [
+
 	| class |
 	class := ReflectivityExamples newAnonymousSubclass.
 	builder compileClassAccessorForAnonymousClass: class.
+	
 	self
-		assert: (class methodDict at: #class) sourceCode
+		assert: (class compiledMethodAt: #class) sourceCode
 		equals: builder classAccessorsForAnonymousClasses first.
 	self
-		assert: (class methodDict at: #originalClass) sourceCode
+		assert: (class compiledMethodAt: #originalClass) sourceCode
 		equals: builder classAccessorsForAnonymousClasses second.
 	self
-		assert: (class methodDict at: #realClass) sourceCode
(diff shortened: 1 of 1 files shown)


guillep:
Hi @hernanmd, yes! Moving the dependencies to System-FileRegistry looks like a move in the good direction

src/System-FileRegistry/GZipReadStream.extension.st
```diff
@@ -12,6 +12,20 @@ GZipReadStream class >> fileReaderServicesForFile: fullName suffix: suffix [
 	^ services
 ]
 
+{ #category : '*System-FileRegistry' }
+GZipReadStream class >> openWithContents: contentsString label: titleString [
+	"Open a text viewer on contentsString and window label titleString"
+
+	SpTextPresenter new
```
Ducasse: This should be packaged in a separate package like System-Spec because now we are introducing a dependency from GZip to Spec.

src/System-FileRegistry/ZipArchive.extension.st
```diff
@@ -1,5 +1,15 @@
 Extension { #name : 'ZipArchive' }
 
+{ #category : '*System-FileRegistry' }
+ZipArchive class >> extractAllIn: aFileReferenceOrFileName [
+	"Service method to extract all contents of a zip."
+	| directory |
+	directory := (UIManager default chooseDirectoryFrom: aFileReferenceOrFileName asFileReference) ifNil: [^ self].
```
Ducasse: There is still a ref to UIManager. And I would remove it to keep 

```
^ (self new)
		readFrom: aFileReferenceOrFileName;
		extractAllTo: directory
```
and remove the UI popping behavior.

src/System-FileRegistry/GZipReadStream.extension.st
hernanmd: As it is an extension method the dependency would be from System-FileRegistry to Spec right? 
But I would create System-Spec because it is a nice idea.

I think the dependency browser is displaying extension methods as dependencies.

Ducasse: Yes I meant from the package that contains the extension to Spec, so System-FileRegistry to Spec.

src/System-FileRegistry/ZipArchive.extension.st
hernanmd: Ok, that would remove the service to select where to extract. But I think it is not essential for UI

hernanmd: Done

Issue #15860, opened 2023-12-27
In the message browser cmd B should open a class browser 


Ducasse:
In the attached screenshot, 
when on the right and selecting displayInstance or any selector we should be able to press cmd+b and get to the class browser.
Right now we get a terrible UndefinedObject.
This is a hug break of flow. 
![Capture 2023-12-27 at 17 07 22](https://github.com/pharo-project/pharo/assets/432549/80481a95-3af8-4462-8205-a14d89a3e03b)

hernanmd:
It seems focus is not set once the message browser is opened: 

https://github.com/pharo-project/pharo/assets/4825959/66266193-9ec6-40c9-8a20-3e9547f4e1bc

hernanmd:
Took me a while to discover but problem here. If the selectedIndexes of a FTTableMorph matches with the requested ones, the ifTrue: exits the method and the selection is not refreshed/announced, and then focus is not set.

<img width="977" alt="Screenshot 2023-12-28 at 16 48 25" src="https://github.com/pharo-project/pharo/assets/4825959/33c9c0ab-67a3-4acb-87eb-1f05a962280f">

Ducasse:
Thanks hernan!

PR #15868, opened 2023-12-28
Open class browser on Cmd/Ctrl B in the message brp

hernanmd:
This PR fixes a problem in the message browser, described in #15860

After opened, currently the message browser contains a messages list which must be clicked manually to get the Ctrl+B or Cmd+B key binding working. See this video to visualize the problem:

https://github.com/pharo-project/pharo/assets/4825959/66266193-9ec6-40c9-8a20-3e9547f4e1bc

The PR add a method `selectFirstItem` for explicit request to select the first item in the message browser, and a couple of supporting methods.

See the following video to see how should it work after this PR is applied:

https://github.com/pharo-project/pharo/assets/4825959/89a3d4ee-7111-4919-bfd6-e2e193edb8ac

### DIFF
--- a/src/Calypso-SystemTools-OldToolCompatibillity/ClyOldMessageBrowserAdapter.class.st
+++ b/src/Calypso-SystemTools-OldToolCompatibillity/ClyOldMessageBrowserAdapter.class.st
@@ -132,7 +132,8 @@ ClyOldMessageBrowserAdapter >> open [
 	query
 		criteriaString: autoSelect;
 		criteriaBlock: refreshingBlock.
-	openedBrowser := ClyQueryBrowserMorph openOn: query
+	openedBrowser := ClyQueryBrowserMorph openOn: query.
+	openedBrowser selectFirstItem.
 ]
 
 { #category : 'accessing' }
--- a/src/Calypso-Browser/ClyDataSourceSelection.class.st
+++ b/src/Calypso-Browser/ClyDataSourceSelection.class.st
@@ -156,6 +156,14 @@ ClyDataSourceSelection >> detachedItems [
 ]
 
 { #category : 'controlling' }
+ClyDataSourceSelection >> ensureVisibleFirstItem [
+	
+	rootDataSource table
+		selectIndexes: #();
+	 	selectFirst
+]
+
+{ #category : 'controlling' }
 ClyDataSourceSelection >> ensureVisibleLastItem [
 
 	"in fact first selection index is last selected item"
--- a/src/Calypso-Browser/ClyQueryViewMorph.class.st
+++ b/src/Calypso-Browser/ClyQueryViewMorph.class.st
@@ -685,6 +685,16 @@ ClyQueryViewMorph >> restoreSelectedItems [
 	highlighting restoreTableSelection
 ]
 
+{ #category : 'event handling' }
+ClyQueryViewMorph >> selectFirstItem [
+
+	| rowsCount |
+	
+	rowsCount := self dataSource numberOfRows.
+	rowsCount = 0 ifTrue: [ ^ self ].
+	UIManager default defer: [ self selection ensureVisibleFirstItem ]
+]
+
 { #category : 'controlling' }
 ClyQueryViewMorph >> selectLastItem [
 
--- a/src/Calypso-SystemTools-QueryBrowser/ClyQueryBrowserMorph.class.st
+++ b/src/Calypso-SystemTools-QueryBrowser/ClyQueryBrowserMorph.class.st
@@ -370,6 +370,14 @@ ClyQueryBrowserMorph >> selectClass: aClass [
 ]
 
 { #category : 'navigation' }
+ClyQueryBrowserMorph >> selectFirstItem [
+
+	self changeStateBy: [
+		resultView selectFirstItem
+	]
+]
+
+{ #category : 'navigation' }
 ClyQueryBrowserMorph >> selectLastItem [
 
 	self changeStateBy: [


hernanmd:
Failing tests are not related

Ducasse:
Let us try :)

src/Calypso-SystemTools-OldToolCompatibillity/ClyOldMessageBrowserAdapter.class.st
```diff
@@ -132,7 +132,8 @@ ClyOldMessageBrowserAdapter >> open [
 	query
 		criteriaString: autoSelect;
 		criteriaBlock: refreshingBlock.
-	openedBrowser := ClyQueryBrowserMorph openOn: query
+	openedBrowser := ClyQueryBrowserMorph openOn: query.
+	openedBrowser selectFirstItem.
```
dionisiydk, line 136: This is not a good fix:
- it is not general as it works only from the playground where the adapter is actually used. Try to spawn implementors from the browser like in your first video - it still does not work -  table is not in the focus. To debug the problem you have to understand why focus became lost when window is opened. Notice that in Pharo 7 the initial focus works correctly. So it might be related to the table morph changes or something (not to the Calypso directly).  
- you don't need new selection tooling here like #selectFirstItem method. The first item is really selected when window appears (you can see that It is shown in the table, right?). The initial selection is implemented with the method #ensureSelectedItemIfNeeded.
- for this local workaround you can just enforce the refocus using a defer action: "openedBrowser defer:  [ openedBrowser resultView takeKeyboardFocus ]" instead of #selectFirstItem.

hernanmd: - You have to be more specific. Which steps you mean with "Try to spawn implementors from the browser" ?
- It is true that the first item is selected when window appears, and I already debugged `ensureSelectedItemIfNeeded` (and it took me hours to understand the code for a single stupid change like this). However, did you read [my comment](https://github.com/pharo-project/pharo/issues/15860#issuecomment-1871295480) where this problem comes from?
- I'm not convinced about enforcing the keyboard focus. Why it would be preferable? On the other side, there was already a `selectLastItem` method, so a `selectFirstItem` is compatible with the existing navigation protocol.

dionisiydk: > * You have to be more specific. Which steps you mean with "Try to spawn implementors from the browser" ?

- Open the browser, select any class and method then press Implementors.
- cmd-b will not work from new window

dionisiydk: > * However, did you read [my comment](https://github.com/pharo-project/pharo/issues/15860#issuecomment-1871295480) where this problem comes from?

Yes and it does not make sense to me:

> * Took me a while to discover but problem here. If the selectedIndexes of a FTTableMorph matches with the requested ones, the ifTrue: exits the method and the selection is not refreshed/announced, and then focus is not set

This is correct behaviour to not announce anything if the selection was not changed.

dionisiydk: > * I'm not convinced about enforcing the keyboard focus. Why it would be preferable? 

Your fix actually performs the focus change but as a side effect of a selection change which you enforce on a table. But the selection state is really correct. You have shown it in your issue description.

So instead of triggering a side effect the workaround can be a simple "openedBrowser resultView takeKeyboardFocus" (no need to defer actually as windows is already opened)

dionisiydk: And I am not against for new selection methods. My comment is that they are not needed to this issue.

hernanmd: > > * You have to be more specific. Which steps you mean with "Try to spawn implementors from the browser" ?
> 
>     * Open the browser, select any class and method then press Implementors.
> 
>     * cmd-b will not work from new window

Ok, feel free to open a new issue and/or PR :)

hernanmd: > > * However, did you read [my comment](https://github.com/pharo-project/pharo/issues/15860#issuecomment-1871295480) where this problem comes from?
> 
> Yes and it does not make sense to me:
> 
> > * Took me a while to discover but problem here. If the selectedIndexes of a FTTableMorph matches with the requested ones, the ifTrue: exits the method and the selection is not refreshed/announced, and then focus is not set
> 
> This is correct behaviour to not announce anything if the selection was not changed.

Nope, the correct behavior is not to assume anything about list contents so not to select anything in the first place.

dionisiydk: > > > * However, did you read [my comment](https://github.com/pharo-project/pharo/issues/15860#issuecomment-1871295480) where this problem comes from?
> > 
> > 
> > Yes and it does not make sense to me:
> > > * Took me a while to discover but problem here. If the selectedIndexes of a FTTableMorph matches with the requested ones, the ifTrue: exits the method and the selection is not refreshed/announced, and then focus is not set
> > 
> > 
> > This is correct behaviour to not announce anything if the selection was not changed.
> 
> Nope, the correct behavior is not to assume anything about list contents so not to select anything in the first place.

In your comment the screenshot shows a stack for the method FTTableMorph>>selectIndex:. I don't see the wrong behaviour there, there is no assumption about the list content. Generally the selection has nothing to do with the focus.
The correct fix is to explicitly set the focus to the table once the browser is opened: #15880

hernanmd: No, in the screenshot the method is not selectIndex: but `selectIndexes:andMakeVisibleIf:` which tells me unfortunately you didn't debug it to get it there, or you didn't read carefully the screenshot to arrive to the buggy scenario, even with two videos (otherwise you would have noticed that, as I explained, the table morph is not refreshing because `selectIndexes:andMakeVisibleIf:` receives anArray with the same contents #(1) of `selectedIndexes. Then the condition evaluates to true and now executing the code needed to refresh and get the focus).

Now since I'm not into code golfing or flexing I will not reply anymore.

Issue #14179, opened 2023-07-02
String should not be extended by UIManager... It sucks


Ducasse:
```
String >> displayProgressFrom: minVal to: maxVal during: workBlock
	"Display this string as a caption over a progress bar while workBlock is evaluated.

EXAMPLE (Select next 6 lines and Do It)
'Now here''s some real progress'
	displayProgressFrom: 0 to: 10
	during: [:bar |
	1 to: 10 do: [:x | bar value: x.
			(Delay forMilliseconds: 500) wait]].
"
	^ UIManager default
		displayProgress: self
		from: minVal
		to: maxVal
		during: workBlock
```

PR #15878, opened 2023-12-29
Replace UIManager progress bars with Job

hernanmd:
This PR implements modifications to clean UIManager references related with progress bars.

- Adapt some progress bars which were using the progress method implemented in String (`displayProgressFrom:to:during:`) in this sense `Collection>>#do:displayingProgress:every:` was a central user. This handles #14179 however it moved the extension method from UIManager to Jobs. Senders of this method could be replaced by its job equivalent in a next iteration.
- `EpLogBrowserPresenter>>cachedLogEntries`
- Adapted `ZipArchive>>extractAllTo:`  to use Job based progress bar.
- Adapted `MCHttpRepository>>#displayProgress:during:` to use Job.
- It should be easier now to remove extension method from Jobs package to `UIManager>>#displayProgress:from:to:during: ` in a next iteration. Kept now because it is used by `informUserDuring:` which is used by Spec and Iceberg.

### DIFF
--- a/src/Compression/ZipArchive.class.st
+++ b/src/Compression/ZipArchive.class.st
@@ -138,18 +138,18 @@ ZipArchive >> close [
 ZipArchive >> extractAllTo: aDirectory [
 	"Extract all elements to the given directory; notifying user via UI on progress and if existing files exist"
 
-	UIManager default
-		informUserDuring: [:bar |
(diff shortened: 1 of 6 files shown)


Ducasse:
Yes this is a good move! Cleaning also the String API is nice!

Issue #14891, opened 2023-10-06
Add argument refactoring does not work for class side methods


pavel-krivanek:
**Bug description**
Add argument refactoring does not work for class side methods. When you add an argument to a class side method, it complains that the instance side does not contain the given method.

**To Reproduce**
1. on class side of a class (like `AIGraphAlgorith`) create methods
```smalltalk
msgArg1: arg1 arg2: arg2
```
and 
```smalltalk
msg 

	^ self msgArg1: nil arg2: nil
```
2. on `AIGraphAlgorithm class>>#msgArg1:arg2:`, open context menu and select refactoring **Add argument**
4. in the Method name editor, add an argument named `arg3` and click __Rename__
5. `Alert: AIGraphAlgorithm does not define #msgArg1:arg2:`

If the method of such name exists on the instance side, the refactoring is performed well.

hernanmd:
Fixed in #15879

PR #15879, opened 2023-12-29
Fix Add argument refactoring failing in class side methods

hernanmd:
This PR solve a problem reported in #14891. The System Command `SycChangeMessageSignatureCommand` (applied by selecting _Refactoring -> Add argument_) was always retrieving the instance side class of the last selected method. This was causing the RB to use RBClass in all cases, even when the browser has a class side method selected. In the later, RBMetaclass should be set as the refactored class.

This problem also existed for the "Remove argument" command, in `SycRemoveMessageArgumentCommand`, which cause a dialog window with the message "Method doesn't exist"


Ducasse:
Tx indeed this is a common mistake.

PR #15880, opened 2023-12-30
Explicitly set initial focus to the query browser table during open

dionisiydk:
It is a general fix for #15860.
- cmd+b from just opened query browser and any other shortcuts


Ducasse:
tx

PR #15887, opened 2023-12-31
Start to rename AST classes

jecisc:
This is a subset of the big change I tried not long ago. Here I rename only the classes in the Nodes package tag of AST-Core. (base PR: #15643)

Script used:

```st
oldPrefix := 'RB'.
newPrefix := 'AST'.

('AST-Core' asPackage tagNamed: 'Nodes') classes do: [ :class |
	(class name beginsWith: oldPrefix) ifTrue: [
		| oldName newName |
		oldName := class name.
		newName := (newPrefix , (oldName withoutPrefix: oldPrefix)) asSymbol.
		
		(RBRenameClassRefactoring rename: oldName to: newName) execute.
		
		(Smalltalk at: newName) class compile: 'initialize
	self deprecatedAliases: { #' , oldName , ' }' classified: 'class initialization'
	 ]
 ] displayingProgress: [ :c | c asString ]
```


Ducasse:
@jecisc are the class deprecated? I could not see it because this is important since it will break a lot of code and this is typically a change we will have to keep around for some versions.

jecisc:
I’m adding deprecated aliases so yes they are deprecated! But apparently the application of the deprecation is what is failing currently so I need to investigate why

Ducasse:
Thanks!

jecisc:
I'll close this one because the conflict is too hard to manage

PR #15888, opened 2023-12-31
More Metacello simplifications

guillep:
Cleanups
 - removed lots of unused extensions (or inlined when possible)
 - refactored duplicated code
 - renames clarify concepts
 - remove dependencies to Gofer (now depend only on bare monticello)
 - add composite to avoid manual iteration of collections in many places

Improvements
 - now the directives are linearised in a single list
 - duplicated directives are not loaded/treated twice
 - specs are created only once
 - push dependencies to `Smalltalk>>at:` to a single place
 - make compiler lookup variables in configured environment too


jecisc:
Your work on Metacello is awesome Guille and thank you for the energy you invest on it

Ducasse:
This is cool that we do not depend on Gofer anymore :)
I should say that I really like the baselineOfAST and Opal. 
This makes me think about the baseline of Refactorings. 
In fact I would love to have Refactoring as an external project. In any case this is cool and this gives me energy to continue 
improving the refactoring engine.

jecisc (approved):

PR #15898, opened 2024-01-03
fix: comment typo for #digitAt:base:

jhvhs:
The example in the comment for `Integer#digitAt:base:` was mistyped.

### DIFF
--- a/src/Kernel/Integer.class.st
+++ b/src/Kernel/Integer.class.st
@@ -659,7 +659,7 @@ Integer >> digitAt: anExponent base: base [
 
 	"Return number that represents digit at given position.
 		42 digitAt: 2 base: 10 -> 4
-		42 digitAt: 1 base: 10 -> 1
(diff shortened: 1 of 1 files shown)


jhvhs:
> This issue has either a default title or empty body. We would appreciate it if you could provide more information. Note: I am not a very intelligent bot, I can only react to new comments. Please add a comment for me if you update the body or title.

Updated the body

Ducasse:
Thanks!

Ducasse:
We will integrate it. 
Now if you want to do the next step: 
- turn this comment into executable comments :) like that we are sure that they are correct
How?
Simple

```
42 digitAt: 2 base: 10 -> 4

-> 

"42 digitAt: 2 base: 10 >>> 4"

```

This way we can identify example in comment and execute them to see that they return the correct value, that the method continues to exists....

jhvhs:
Closing in favour of #15911

PR #15908, opened 2024-01-03
Fix #15907 Remove teachable and teachableTest

### LINKED ISSUE #15907: Remove Teachable from SUnit Mooc.
I made the stupid mistake to think that a single tiny class was making more sense in Pharo while in P12 I remove tons of classes.
But this is ok I will juts remove Teachable and in fact my driver refactoring tests are not even using it.
Having multiple external projects is more manageable but I will let people manage it if they want. I do not want Teachable anymore in Pharo.

Ducasse:


Ducasse:
Here it is!

PR #15911, opened 2024-01-03
Feature/Integer documentation

jhvhs:
Examples in the `Integer` class and its extensions are fixed and convereted into executable format.


Ducasse:
Super nice :)
Thanks a lot!

Ducasse:
It is a nice contribution!

PR #15913, opened 2024-01-04
Integrate Teachable into "SUnit-Mock" via LOADING

### LINKED ISSUE #15806: Integrate Teachable into SUnitMock
I misunderstood the difference between Mock and Teachable and I would really like to have Teachable because one some scenarios it is more adapted than Mock.
I asked Torsten about his ok about it.

https://github.com/astares/Pharo-Teachable

### LINKED ISSUE #15912: Integrate Teachable into "SUnit-Mock" via LOADING
Try to integrate Teachable by LOADING not COPYING to satisfy the requirements of
all participants, still allow Teachable to be used independent from SUnit-Mock, not having
any dirty packages when loading a new (external) new versions but also allow to 
properly compose the image by loading (instead of copying code around)

For details see discussion in
- https://github.com/pharo-project/pharo/pull/15873
- https://github.com/pharo-project/pharo/issues/15907

astares:
- properly declare dependencies to external project
- Fix #15912
- Fix #15806
- additionally provides a #reset method on BaselineOfPharo to be able to recompute external projects
- also fixes a lint issue by proper categorization of #baseline: method

### DIFF
--- a/src/BaselineOfPharo/BaselineOfPharo.class.st
+++ b/src/BaselineOfPharo/BaselineOfPharo.class.st
@@ -121,6 +121,13 @@ BaselineOfPharo class >> newToolsRepository [
 	^ (self externalProjectNamed: 'NewTools') repository
 ]
 
+{ #category : 'class initialization' }
+BaselineOfPharo class >> reset [
+	<script>
+
+	ExternalProjects := nil
+]
+
 { #category : 'accessing - external projects' }
 BaselineOfPharo class >> roassal [
 	<externalProject>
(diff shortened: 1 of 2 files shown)


Ducasse:
We do not need teachable integrated like that in Pharo. It will force users to add a dependency to teachable in addition to Sunit. Personally I will not use it. It will just add more complexity for a class with 4 methods. I seriously do not see the need.

Ducasse:
If people wants to use Teachable then they will just use it as an external package. In addition MockObject provides the same functionality even if the API is less obvious. I will do a pass on the comments to make this clearer.
I was stupid to imagein that we can do something simple in this bikeshed world.

astares:
>It will force users to add a dependency to teachable in addition to Sunit.

Wrong. With this PR there is no additional need to add another dependency to Teachable on user side. The users original dependency on SUnit is still valid and enough.

As BaselineOfSUnit has declared the dependency to Teachable the **transitive** dependency mechanism will just kick in and perform its magic.

image

PR #15917, opened 2024-01-04
Doing a pass on comments and protocol to improve understanding the mock

Ducasse:


Ducasse:
What is the consumption of energy of this dull and useless robot?

PR #15935, opened 2024-01-08
Fix class and trait definition printing to consider package tags

jecisc:
This fixes the printing a class, test class and trait template printing depending on the package to also consider the package tag. 

It also removes the printing for the old formats since they are not supportedby Pharo anymore to keep only the fluid format.


Ducasse:
The old printer is not needed in MC or code export?

Ducasse:
Sounds good to me.

Ducasse:
Broken tests are unrelated.

PR #15944, opened 2024-01-10
Cleaning delay and time dependencies.

guillep:
This PR continues the reorganization effort around time and duration.
Code is expected to be backwards compatible, Classes were only moved around and repackaged.
Some tests have been rewritten.

Extract time-related classes to System-Time[-Tests] and System-Benchmark[-Tests].
Rename Kernel-Chronology et al to System-Time. 
Remove direct references to Duration.
Review dependencies to Time and Duration keeping only the necessary ones.


jecisc:
From the impact on the lines of codes I wonder if we're not missing the removal of some renamed packages

guillep:
certainly!

Ducasse:
```
DeprecationPerformedNotification: The method Semaphore>>#wait: called from SemaphoreTest>>#testWaitTimeDuration has been deprecated. Use #waitTimeoutMilliseconds: instead

*** Warning: Deprecation: The method Semaphore>>#wait: called from SemaphoreTest>>#testWaitTimeDuration has been deprecated. Use #waitTimeoutMilliseconds: instead
DeprecationPerformedNotification: The method Semaphore>>#wait: called from SemaphoreTest>>#testWaitTimeDuration has been deprecated. Use #waitTimeoutMilliseconds: instead

*** Warning: Deprecation: The method Semaphore>>#wait: called from SemaphoreTest>>#testWaitTimeDuration has been deprecated. Use #waitTimeoutMilliseconds: instead
DeprecationPerformedNotification: The method Semaphore>>#wait:onCompletion:onTimeout: called from SemaphoreTest>>#testWaitTimeDurationWithCompletionAndTimeoutBlocks has been deprecated. Use #waitTimeoutMilliseconds:onCompletion:onTimeout: instead

*** Warning: Deprecation: The method Semaphore>>#wait:onCompletion:onTimeout: called from SemaphoreTest>>#testWaitTimeDurationWithCompletionAndTimeoutBlocks has been deprecated. Use #waitTimeoutMilliseconds:onCompletion:onTimeout: instead
DeprecationPerformedNotification: The method Semaphore>>#wait:onCompletion:onTimeout: called from SemaphoreTest>>#testWaitTimeDurationWithCompletionAndTimeoutBlocks has been deprecated. Use #waitTimeoutMilliseconds:onCompletion:onTimeout: instead

*** Warning: Deprecation: The method Semaphore>>#wait:onCompletion:onTimeout: called from SemaphoreTest>>#testWaitTimeDurationWithCompletionAndTimeoutBlocks has been deprecated. Use #waitTimeoutMilliseconds:onCompletion:onTimeout: instead
```

guillep:
What is strange is that the timeout killed the entire process instead of killing the test and continuing with the next one... Maybe the test runner does not properly catch deprecations?

guillep:
CI is green (modulo unrelated tests)!

src/System-Time/Semaphore.extension.st
```diff
@@ -0,0 +1,23 @@
+Extension { #name : 'Semaphore' }
+
+{ #category : '*System-Time' }
+Semaphore >> wait: aDuration [
```
dionisiydk, line 4: What is the point of deprecating these two methods? I understand general idea of improving the modularity of the time package but Duration is still here. So fluent human friendly API is defined together with the main required class - Duration. Why we can't keep it? 

Now without this nice API the users will model the timeouts for their objects as primitive numbers (as no alternatives from Pharo libraries, right?). 
It always leads to duplicating comments explaining the units everywhere and duplicating methods providing alternatives for other kind of units. 
The Duration was introduced to address these kind of problems generally. And in Pharo it does the work really well.

I suggest to rethink this decision and to keep the fluent API.

Ducasse:
We do not want to have duration in the core of the system. 
Duration is a dull class only used most of the time to wrap and unwrap a number.

dionisiydk:
> We do not want to have duration in the core of the system. Duration is a dull class only used most of the time to wrap and unwrap a number.

You are not going to remove Duration completelly, right?
Then why its convenient extensions are planed to be removed? 
Notice they are defined in same package. Eventually you will move all them together into new dedicated package. So why remove them now?

Ducasse:
We want to have duration in an optional package (because duration just make the system more verbose for nearly no positive points).  
We are working on kernel design and duration should not be part of it. 
We want a layered system and Semaphore should not depend on Duration.
Semaphore should only know about milliseconds. 
In addition wait: is not  intention revealing. 
Check the API that we proposed waitTimeoutMilliseconds:, waitTimeoutSeconds: 
Now we deprecated the call so I do not see why this is a problem.

dionisiydk:
> We are working on kernel design and duration should not be part of it.
> We want a layered system and Semaphore should not depend on Duration.

In my original comment I said that I understand this goal. No objections here.

> Semaphore should only know about milliseconds.

What is an issue with having extension on the level of Duration package?

> In addition wait: is not intention revealing.

```Smalltalk
semaphore wait: 10 seconds
``` 

How it can be non intuitive? I'm really surprized to hear this. 

> Check the API that we proposed waitTimeoutMilliseconds:, waitTimeoutSeconds:

What about minutes? Will you add another method waitTimeoutMinutes:? Or do you think users don't need this? How will you make such a decision. Also you did not provide same alternative for another method #waitTimeoutMilliseconds:onCompletion:onTimeout:.

Duration elegantly covers all these cases. I would say remove #waitTimeoutSeconds: and keep the duration based method for all non primitive cases (in a dedicated package of course). It will also avoid the duplication of the duration arithmetics (even if it's small).

Thinking about names, #waitTimeoutMilliseconds is a bit confusing: 
- from one side timeout in the name sounds like there will be a timeout exception but it just returns false. 
- from the other side the meaning is like we are asking a semaphore to wait **the** timeout. But we actually waiting for the signal.
IMHO simple form without the timeout word sounds better and simpler (waitMilliseconds: or waitForMilliseconds:).

> Now we deprecated the call so I do not see why this is a problem.

Deprecation means it will be removed. That's why I started this thread. 
Having nice API as extension does not lead to the bad modularity

dionisiydk:
> Duration is a dull class only used most of the time to wrap and unwrap a number.

I feel really sad for the Duration class. It does not deserve it :).
I would say the Duration is one of examples why Pharo/Smalltalk is great. We can model even simple things with objects and benefit from it everywhere.
```Smalltalk
2 hours + 10 minutes.
(3 days + 25 hours) asMinutes.
10 minutes * 3.
DateAndTime now + 34 days.
DateAndTime now - 3 weeks.
10 seconds wait.
```
It's all driven by Duration. 

I am pretty sure your comment was about specific case in Kernel and you did not try to harm Mr Duration :) 
But all these examples are all extensions to the Kernel classes. Maybe they are not packaged in a right way. But please do not remove them in a way to the modular Pharo.

Ducasse:
We will discuss with guille. Now we want to remove these extensions from the kernel. Duration should be optional.

Issue #15664, opened 2023-12-01
Indistinguishable `SortedCollection`s often considered not-equal due to sortBlock


daniels220:
**Bug description**
SortedCollection includes its `sortBlock` as part of equality comparison. This is reasonable enough—worth noting that the two sortBlocks have to result in the same sort order or the collections will already be not-equal just from comparing the elements, but just because the two sortBlocks happen to sort _these_ elements the same, that doesn't imply that they will _always_ act the same (and if the collection has 0 or 1 elements, we know nothing at all about the sorting).

However, `BlockClosure` itself has no implementation of `#=`, so two SortedCollections created by e.g. a factory method that assigns a sort block will always be not-equal, which is problematic.

**To Reproduce**
1. Most blatantly, evaluate: `(SortedCollection sortBlock: [:a :b | b < a]) = (SortedCollection sortBlock: [:a :b | b < a])`
2. More realistically (this is how I discovered this), have a method like:  
```
Person class>>lastNameFirstNameSortBlock
    ^[:a :b | a lastName < b lastName or: [a lastName = b lastName and: [a firstName < b firstName]]]
```  
and then have a codepath that ultimately reduces to `(collA asSortedCollection: Person lastNameFirstNameSortBlock) = (collB asSortedCollection: Person lastNameFirstNameSortBlock)` (and as it happens `collA` and `collB` do have the same *contents*).
3. Observe `false` return

**Expected behavior**
`true` return for same contents and equivalent/indistinguishable `sortBlock`.

**Possible solutions**
The simplest possible solution is to just not compare the sortBlock, assuming that if the contents sort in the same order, the sort blocks are equivalent enough. In most cases this is probably a good assumption—something like a reversed `sortBlock` will basically _always_ produce a different order, and comparing two `SortedCollection`s drawn from the same set of objects (such that there's any _possibility_ of the contents being the same), with different sorting but that might _happen_ to produce the same order in some cases, and _caring_ that we get a not-equal result in those cases...seems pretty strange to me. The one obvious exception is if both collections are empty or have a single element, where the sortBlock doesn't come into play—but even then, especially in the empty case, it would usually be weird to _care_ that the sortBlock is different. Not always though. (This is the behavior in Dolphin Smalltalk.)

[...]

PR #15948, opened 2024-01-10
WIP: Add an implementation of BlockClosure>>=

daniels220:
Blocks are considered equal if they execute the same code and have the same receiver and copied values. Fortunately CompiledCode>>= already exists and seems remarkably robust.

In the vast majority of cases this is straightforward, however I know of two edge cases:
1. Super sends break comparison because they introduce a new copy of the VariableBinding for the class each time.
2. It's not clear what equality should mean for blocks which contain a non-local return. Right now the answer is effectively "use identity" (technically two blocks with the same code from the same execution of the same method will be considered equal, but anyone who actually does that...why???)

This should be considered a little bit WIP pending final decisions about the above two cases, but I believe the basic code is functional.

The original motivation here is fixing #15664—without an implementation of `BlockClosure>>=`, any `SortedCollection` with a custom `sortBlock` is very unlikely to compare equal to another, unless the blocks are actually identical. But VisualWorks and Dolphin Smalltalk both have an implementation of `BlockClosure>>=`, and Pharo already implements `CompiledCode>>=`, so this seems like a reasonable thing to have in general.

### DIFF
--- a/src/Kernel-CodeModel/BlockClosure.class.st
+++ b/src/Kernel-CodeModel/BlockClosure.class.st
@@ -52,6 +52,25 @@ BlockClosure class >> isAbstract [
 	^ self == BlockClosure
 ]
 
+{ #category : 'testing' }
+BlockClosure >> = aBlockClosure [
+
+	self == aBlockClosure ifTrue: [ ^ true ].
+	self class = aBlockClosure class ifFalse: [ ^ false ].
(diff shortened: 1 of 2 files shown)


daniels220:
Huh, CI build is broken due to `Instance of CleanBlockClosure did not understand #asExportedLiteral`. That seems like a separate issue, but I'm also surprised it's coming up here and not on master...? The only place this pull actually adds clean blocks is in the tests, and surely those aren't being loaded in early bootstrap?

PR #15949, opened 2024-01-11
Speedup Float and Integer comparison

jecisc:
I had a strange case while optimizing a clustering algorithme were an implementation was really slow using `x < y` in the code. And it became almost instantaneous using `y >= x`. This is because x is usually a float and y an integer. 

In the first case the comparison primitive fails and the fallback casts the float as fraction. In the second case, the primitive works directly.

I propose to speed up #>, #<, #>= and #<= by inverting the order of the comparison and making the integer the receiver instead of the float. 

This causes a little slow down on #= and #~= in case the fraction part is not 0.0 but this seems relatively small compared to what we gain

```st
[
value := 1.00000001.

10000000 timesRepeat: [
	value < 1.
	value := value + 1 ] ] timeToRun

 "Before: 0:00:00:02.38"
 "After: 0:00:00:00.201"
```


guillep:
Interesting, thanks @jecisc . I'll steal your example for a VM dojo :)

jecisc:
There is a crash and I don't know if it's related or not

guillep:
I dug this a bit. The thing is that integer vs FP comparisons are only defined for boxed floats and not small floats.
And most (if not all) our floats are small floats

```smalltalk
boxedFloat := BoxedFloat64 fromIEEE64Bit: 1.0 asIEEE64BitWord.
unboxedFloat := 1.0.

{[ 1 < boxedFloat ] bench.
[ 1 < unboxedFloat ] bench.

[ boxedFloat < 1 ] bench.
[ unboxedFloat < 1 ] bench.}
"#(
'227070993.003 per second'
'12472895.242 per second'
'239098448.421 per second'
'223868268.746 per second')"
```

Ducasse:
Guille what is a smallFloat vs a Float I thought that we only have "immediate"Floats

guillep:
> Guille what is a smallFloat vs a Float I thought that we only have "immediate"Floats

Boxed floats are still there and supported for those cases where a float cannot be expressed as an immediate because they need the extra bits taken by the tags.

jecisc:
I restarted the CI and the same error happens again. I’m not sure of the origin but this change might not be so simple

guillep:
CI is good!

Ducasse:
What do you mean here

```
The thing is that integer vs FP comparisons are only defined for boxed floats and not small floats.
```

That the methods is not defined so the adapt... is invoked?
You tell me next time we cross each other.

guillep:
I mean the comparison primitives `<` et al.
The primitive handles the case when both are ints.
But if the argument is a boxed float it converts the (int) receiver to float and compares.

However, if the argument is an immediate float, the primitive just fails and executes the (20x slower) fallback code.

PR #15960, opened 2024-01-16
Fix other callers of defineClass: ... builder....

Ducasse:
Fix other callers of defineClass: ... builder....


jecisc (approved):

Ducasse:
It brokes some refactoring tests so I have to check. I really do not like this code.

Ducasse:
So I will close if because it breaks many tests and we will check that if one day we go back to revisit classFor:

PR #15969, opened 2024-01-16
adding back millisecondsToRun: but we will deprecate it.

Ducasse:
adding back millisecondsToRun: but we will deprecate it.


jecisc (approved):

sebastianconcept:
Deprecate it favor of? or feature moved to?

Ducasse:
[ ] millisecondsToRun

PR #15972, opened 2024-01-17
Add release test for ReDoNotSendSuperInitializeInClassSideRule

### LINKED ISSUE #15970: Class side initialize doing `super initialize` reinitialize classes
Users should not do `super initialize` on a class side initialize method.
However, when they do, classes get reinitialized (superclass := Object, empty method dictionary).

Maybe we should introduce an empty initialize method in `Object class` to avoid this problem, but still this could cause (wrong) duplicated initializations.

jecisc:
This change adds a release test to ensure the rule ReDoNotSendSuperInitializeInClassSideRule has no violation in Pharo.

It also fixes the last violations and lastly it speeds up the rule by a lot skipping the check for methods that are not class side initialization. This avoids the creation of a lot of ASTs.

Fixes #15970

### DIFF
--- a/src/General-Rules/ReSuperSendsRule.class.st
+++ b/src/General-Rules/ReSuperSendsRule.class.st
@@ -16,13 +16,6 @@ ReSuperSendsRule class >> uniqueIdentifierName [
 	^'SuperSendsRule'
 ]
 
-{ #category : 'hooks' }
-ReSuperSendsRule >> afterCheck: aNode mappings: mappingDict [
-
-	^aNode methodNode methodClass withAllSubclasses noneSatisfy: [: class |
-		class includesSelector: aNode selector ]
-]
-
 { #category : 'accessing' }
 ReSuperSendsRule >> group [
 	^ 'Design Flaws'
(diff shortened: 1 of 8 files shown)


Ducasse:
Thanks a lot cyril.

src/General-Rules/ReSuperSendsRule.class.st
```diff
@@ -40,3 +33,9 @@ ReSuperSendsRule >> initialize [
 ReSuperSendsRule >> name [
 	^ 'Rewrite super messages to self messages'
 ]
+
+{ #category : 'testing' }
+ReSuperSendsRule >> shouldCheckMethod: aMethod [
+
+	^ aMethod methodClass withAllSubclasses noneSatisfy: [ :class | class includesSelector: aMethod selector ]
```
Ducasse, line 40: why this is not aMethod methodClass class withAllSuclasses?

Ducasse: Argh this is another rule.

Ducasse: I just to not get this code :)

jecisc: It’s another rule and since I optimized the first one I also did this one

Ducasse:
Super nice :)

PR #15973, opened 2024-01-17
Fix bug in integer vs float comparisons

### LINKED ISSUE #15971: Bug in integer vs float comparisons
**Bug description**

```smalltalk
1 < 1.0
 => true
```

which is clearly wrong :)

The issue was introduced accidentally by https://github.com/pharo-project/pharo/commit/ea865c684bf77445e0d5d1474121386477b78c3d

A good fix should add a couple of tests

jecisc:
Fix bug in comparison + add a bunch of tests

There should probably be more tests added on this but this is a start

Fixes #15971

### DIFF
--- a/src/Kernel-Tests/FloatTest.class.st
+++ b/src/Kernel-Tests/FloatTest.class.st
@@ -211,6 +211,46 @@ FloatTest >> testHash [
 	Float infinity hash
 ]
 
+{ #category : 'tests - arithmetic' }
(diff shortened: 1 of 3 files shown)


jordanmontt (approved):
Maybe add some other test just for testing equality? something like

testEquality

self assert: 1.0 = 0

Ducasse:
:)

PR #15976, opened 2024-01-18
Add a guard on class initialization

jecisc:
Add a guard in case we do an unwanted class initialization to warn the users about the risks.

### DIFF
--- a/src/Kernel/Object.class.st
+++ b/src/Kernel/Object.class.st
@@ -109,6 +109,14 @@ Object class >> howToModifyPrimitives [
 	self error: 'comment only'
 ]
 
+{ #category : 'initialization' }
+Object class >> initialize [
+
+	Warning signal:
+		'Proceeding will call `Behavior>>initialize` and this method will update my superclass and my method dictionary. Are you sure you want to proceed?'.
+	^ super initialize
+]
+
 { #category : 'instance creation' }
 Object class >> newFrom: aSimilarObject [
 	"Create an object that has similar contents to aSimilarObject. If the classes have any instance variables with the same names, copy them across. If this is bad for a class, override this method."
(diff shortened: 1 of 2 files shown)


jecisc:
I guess something in the bootstrap is calling initialization on object at some point :/

astares:
Do we really need this as a second warning - there already is a clear lint error:

image

src/Kernel/Object.class.st
```diff
@@ -109,6 +109,14 @@ Object class >> howToModifyPrimitives [
 	self error: 'comment only'
 ]
 
+{ #category : 'initialization' }
+Object class >> initialize [
+
+	Warning signal:
+		'Proceeding will call `Behavior>>initialize` and this method will update my superclass and my method dictionary. Are you sure you want to proceed?'.
+	^ super initialize
```
guillep: Well the point here would be to not do the super initialize at all @jecisc otherwise this breaks everything.

I'm against resumable exceptions just for the sake of it. If we can avoid them, I vouch for that.

jecisc:
@astares If you execute this, in the past it was reinitializing object and was pretty transparent. Now it remove the superclass of the class calling the super and removes the method dictionary. 
We've seen that happen in some projects.

astares (approved):

guillep (changes requested):

Ducasse:
I would not do the super initialize at all in this method.

jecisc:
I'm trying to use an error instead of a warning with an explicit error message but I'm not sure the bootstrap will build with this. Let's try

Ducasse:
Got bitten. So I put an halt in Object class>>initialize 
because it was not a super initialize that kills my class but a setup.

```
setUp
	super setUp.
	sample := MicMicrodownSnippetFactory new.
	parser := Microdown new.
	offlineStatus := Microdown offline.
	Microdown offline: false.
	MicRichTextComposer initialize.
```

Ducasse:
And more like self composerClass initialize!

MarcusDenker:
This was now fixed by adding an empty initialize

Issue #15922, opened 2024-01-06
TraitedMetaclass  testFileOutCategory – MacOSX64.Traits.Tests.TraitFileOutTest is failling


Ducasse:
```
Error

Message not understood: TraitedMetaclass >> #classInstaller
Stacktrace

MessageNotUnderstood
Message not understood: TraitedMetaclass >> #classInstaller
TraitedMetaclass(Object)>>doesNotUnderstand: #classInstaller
[ :user | user classInstaller remake: user basedOn: aBuilder ] in TraitBuilderEnhancer>>propagateChangesToRelatedClasses:builder:
IdentitySet(Set)>>do:
TraitBuilderEnhancer>>propagateChangesToRelatedClasses:builder:
[
		newClass := builder build.
		self installInEnvironment: newClass.
		self installSubclassInSuperclass: newClass.
		builder builderEnhancer beforeMigratingClass: builder installer: self.
		builder builderEnhancer migrateToClass: newClass installer: self.
		builder builderEnhancer afterMigratingClass: builder installer: self.
		builder builderEnhancer propagateChangesToRelatedClasses: newClass builder: builder.
		
	] in ShiftClassInstaller>>make
FullBlockClosure(BlockClosure)>>on:do:
ShiftClassInstaller>>make
ShiftClassInstaller>>make:
ShiftClassInstaller class>>update:to:
MetaclassForTraits(Metaclass)>>trait:slots:
MetaclassForTraits(Metaclass)>>setTraitComposition:
MetaclassForTraits(TraitedMetaclass)>>removeFromComposition:
[ :e | e isObsolete ifFalse: [e removeFromComposition: self class ]] in TD classTrait(Trait)>>removeFromSystem:
IdentitySet(Set)>>do:
TD classTrait(Trait)>>removeFromSystem:
TD classTrait(Class)>>removeFromSystem
[ :class | class removeFromSystem ] in Package>>removeTag:
Array(SequenceableCollection)>>do:
Package>>removeTag:
PackageTag>>removeFromSystem
[ :tag | tag removeFromSystem ] in Package>>removeFromSystem
Set>>do:
Package>>removeFromSystem
[ :package | package removeFromSystem ] in PackageOrganizer>>removePackage:
FullBlockClosure(BlockClosure)>>cull:
[:v | ^ aPresentBlock cull: v] in IdentityDictionary(Dictionary)>>at:ifPresent:ifAbsent:
FullBlockClosure(BlockClosure)>>cull:
[...]

PR #15978, opened 2024-01-18
Fix bug in class trait composition export

guillep:
In the last couple of days we have seen a bug provoking a wrong export of class side trait definitions as follows:

imagen

I chased this bug to the lowest parts of the system, and the cause seems to be twofold:
 - exporters should use `basicTraitComposition` instead of `traitComposition`
 - (meta) classes using traits were being rebuilt even when it was not necessary because of a wrong computation of the metaclass class. This happened only when using the `fillClassSideFromEnvironment:` method, typically when editing a method from Calypso+fluid.

I managed to write 6 tests, only one was broken. All are green after these changes.
However, it is strange to me why we have `fillClassSideFromEnvironment:`, `fillFor:` and others which seem to be duplicated some behavior?

I'd like a second pair of eyes on the tests and my fixes for real here.

Probably related: https://github.com/pharo-project/pharo/issues/15922

### DIFF
--- a/src/Shift-ClassBuilder/ShiftClassBuilder.class.st
+++ b/src/Shift-ClassBuilder/ShiftClassBuilder.class.st
@@ -319,7 +319,7 @@ ShiftClassBuilder >> fillClassSideFromEnvironment: anEnvironment [
 
 	self metaclassClass: old class class.
 	self classSlots: old class slots.
-	self classTraits: old class basicTraitComposition
+	self classTraits: old class traitComposition
 ]
 
 { #category : 'initialization' }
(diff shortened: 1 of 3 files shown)


jecisc (approved):
This seems a legit change after 1089f68d72b9f4cc247dcc01386e9130bbced277

Ducasse:
Apparently basicTraitComposition is not defined.
```
osx-64 / Tests-osx-64 / testRedundantClassDefinition – MacOSX64.Epicea.Tests.EpCodeChangeIntegrationTest
6s
Error

Message not understood: Metaclass >> #basicTraitComposition
Stacktrace

MessageNotUnderstood
Message not understood: Metaclass >> #basicTraitComposition
Metaclass(Object)>>doesNotUnderstand: #basicTraitComposition
ShiftClassBuilder>>fillClassSideFromEnvironment:
Object class(Behavior)>><<
UndefinedObject>>DoIt
OCReceiverDoItSemanticScope(OCDoItSemanticScope)>>evaluateDoIt:
OpalCompiler>>evaluate
OpalCompiler>>evaluate:
EpCodeChangeIntegrationTest>>testRedundantClassDefinition
EpCodeChangeIntegrationTest(TestCase)>>performTest
```

guillep:
CI is now passing!

Issue #15958, opened 2024-01-16
Report class definitions errors when loading code


hernanmd:
When loading class definitions that contain errors, the `loadClassDefinitions` method currently saves the erroneous definitions, but omits any logging of the errors. 

**Expected behavior**
This enhancement proposes that at least the errors can be sent to the Transcript, so that the user can quickly see what loading errors existed and take action.

**Screenshots**
<img width="1237" alt="Screenshot 2024-01-16 at 13 56 53" src="https://github.com/pharo-project/pharo/assets/4825959/6faeecaa-e3be-4dc0-8d54-9d21812be226">

jecisc:
The goal of this variable is to do a retry of those definition because it might be a class using a trait from the same package that gets loaded before the trait. So we retry once the trait is loaded. 

But it's true that if in the end we still have errors they should be reported.

jordanmontt:
Can we close this @hernanmd ? as there is a related PR that has been merged

PR #15979, opened 2024-01-18
Report exception message text when class definition loading fails

hernanmd:
This PR includes a minor change for Pharo 12 so we can see error messages in the Transcript when loading class definitions fail in `MCPackageLoader`.

It addresses the issue #15958.

### DIFF
--- a/src/Monticello/MCPackageLoader.class.st
+++ b/src/Monticello/MCPackageLoader.class.st
@@ -150,6 +150,7 @@ MCPackageLoader >> loadClassDefinitions [
 	errorDefinitions := OrderedCollection new.
 
 	"We first load the class definitions and store the potential errors"
+	[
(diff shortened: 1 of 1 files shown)


jecisc (approved):

jecisc:
I get a 404 while trying to access the CI result. I'll restart it

guillep (changes requested):
Hi @hernanmd , could you explain how this fixes the issue?

Because this is adding an ensure only, but exceptions are caught already inside the loop. So the call to #warnAboutErrors: will happen in any case. What am I missing?

hernanmd:
We checked this with @jecisc some weeks ago and this happened because I had an error in in the whileFalse block, of course errors were recorded, but they were not reported at all because the "Finally we warn about the errors we could not fix." was never reached.

guillep (approved):
ok!

PR #15981, opened 2024-01-18
Fix: cannot move method to class side

### LINKED ISSUE #15790: [RB] Getting failedApplicabilityPreconditions fails
### To reproduce
Try to move an instance side method to a class side method.

### The cause
The following method assumes that `applicabilityPreconditions` returns a collection, but in most of refactorings this method returns an instance of RBCondition.
```
failedApplicabilityPreconditions

	^ self applicabilityPreconditions reject: [ :cond | cond check ]
```

### The problem
The method `applicabilityPreconditions` is sometimes returning a collection and this is useful to handle errors at the level of the driver. 
For example in `RBRenameClassAndDeprecateClassTransformation`, `RBRemoveClassRefactoring` and `RBRenameInstanceVariableRefactoring`

@Ducasse @balsa-sarenac I think we need to find an alternative way to get failed preconditions, but I'm not sure how

balsa-sarenac:
Breaking changes return a list instead of a condition object. Cleanup related code.

Closes: #15790

### DIFF
--- a/src/Refactoring-Core/RBCompositeRefactoring.class.st
+++ b/src/Refactoring-Core/RBCompositeRefactoring.class.st
@@ -31,9 +31,7 @@ RBCompositeRefactoring >> generateChanges [
 	
 	self prepareForExecution.
 	self checkApplicabilityPreconditions.
-	self breakingChangePreconditions check ifFalse: [
-		RBBreakingChangeChecksFailedWarning signal:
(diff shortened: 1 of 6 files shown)


Ducasse:
Nice :)

balsa-sarenac:
Failing tests seem unrelated

jecisc (approved):

PR #15983, opened 2024-01-18
Fix: instance and shared variables driver bugs

### LINKED ISSUE #15982: Rename instance variable and shared variable drivers are broken .
Hi balsa

if you have a couple of minutes could you check the removeInstance/SharedVariable driver?

S

balsa-sarenac:
This PR fixes two small bugs in variables drivers:
- violators list should be flat collected before returning (since we have list of preconditions)
- remove shared variables was missing code to check breaking changes
- 
Closes: #15982


Ducasse:
Sounds good.

PR #15986, opened 2024-01-19
Improve the way to know if a class includes a trait

jecisc:
This change adds #includesTraits: on Behavior so that we do not have to ask the trait composition and is speeding up #includesTraits by having a specific implementation by kind of composition so that we do not collect all traits when the trait is present.

This allows to divide the computation of a moose visualisation by a factor of 3.5.

If this passes I would like to backport it to P11 if possible so that we can make some visualization more usable on bigger models


Ducasse:
cool

Issue #15991, opened 2024-01-20
Instance of RBRenameClassVariableChange did not understand #asText


hernanmd:
**Bug description**
Removing a class (or shared) from Calypso variable raise an Error.

**To Reproduce**
```smalltalk
Object << #MiPrueba
	slots: {};
	sharedVariables: { #Prueba };
	package: 'Prueba1'
```

Then in Calypso select the class variable "Prueba", right click, Rename.

**Screenshots**

<img width="892" alt="Screenshot 2024-01-20 at 15 48 20" src="https://github.com/pharo-project/pharo/assets/4825959/710b55e1-b5cc-4898-aa13-2600e845a572">

PR #15992, opened 2024-01-20
Add missing caret in #textToDisplay method.

hernanmd:
Single byte PR to add missing caret in #textToDisplay method, which was causing an error in renaming class variables, as reported in #15991.


Ducasse:
tx

PR #16007, opened 2024-01-21
Fix ReDoNotSendSuperInitializeInClassSideRule in Calypso

### LINKED ISSUE #16002: Super initialize rule is now checking instance side and this is wrong
Hi cyril

I have the impression that we got a regression with the rule.
I see the following in latest pharo12.

Capture 2024-01-21 at 18 45 33

jecisc:
Calypso duplicates the way to iterate over the AST nodes for rules :(

I'm adapting the code to follow the changes I did in reraku

Fixes #16002


Ducasse:
I got the problem on commit. I do not how the rule engine is called by iceberg. May be this is another problem.

PR #16008, opened 2024-01-22
Remove gofer references

jecisc:
This change remove some dependencies to gofer in Monticello.

It also cleans some dependencies


jecisc:
And coupled with this one, removes the last dependency to gofer in the image: https://github.com/pharo-vcs/iceberg/pull/1795

Issue #15931, opened 2024-01-07
Browse this version is broken


Ducasse:
The external browser uses it.
Did not have the time to check more.

hernanmd:
I could not reproduce a bug with protocolNames but I've found multiple issues using the  "Browse this version in Calypso":

<img width="517" alt="Screenshot 2024-01-08 at 12 30 51" src="https://github.com/pharo-project/pharo/assets/4825959/782e3276-092f-4ab2-a22e-1aa448bb9230">

Trying to browse unloaded packages:

<img width="1186" alt="Screenshot 2024-01-08 at 12 28 37" src="https://github.com/pharo-project/pharo/assets/4825959/cfa47973-faa6-46b8-9f2e-2724a10dc226">

Trying to browse loaded packages:

<img width="955" alt="Screenshot 2024-01-08 at 12 29 10" src="https://github.com/pharo-project/pharo/assets/4825959/e4d169b2-5ca8-44a9-b6a3-8bc9c21fced0">

hernanmd:
Fixed in #16009

PR #16009, opened 2024-01-22
Fixes for browsing a package from a repository using Calypso

hernanmd:
This PR addresses #15931.

It adds fixes for Calypso, which currently crashes when trying to navigate in code from the repository (regardless if the code is in the image or not).

To reproduce these errors, from Iceberg select a project and "Browse this version in Calypso". When trying to browse its Packages, Classes and Methods multiple errors appear because these objects are still instances of Ring objects. These Ring objects are sent messages that are implemented by the new Package implementation, so they will only work if the Package functionality is reified in Ring. 
As a transitional measure and until a single metamodel is unified in Pharo, the methods include checks so that "remote" artifacts can be navigated from Calypso.


jecisc (approved):

src/Calypso-Ring/RGMethod.extension.st
```diff
@@ -87,6 +87,16 @@ RGMethod >> readsSlot: aSlot [
 	^ nodes anySatisfy: [ :node | node binding == aSlot ]
 ]
 
+{ #category : '*Calypso-Ring' }
+RGMethod >> realMethod [
+	"Included to support for browsing unloaded methods, i.e. from the Calypso. 
+	Since Calypso still uses Ring objects, it needs to get the real method so it can understand messages such as #isClassified. The real fix here is to reify Package support (like tags) in Ring (which is planned to be removed). See issue #15931 for details"
+
+	^ (Smalltalk at: self parent resolvedNameOrNil)
```
Ducasse: could we have self class environment instead of Smalltalk

src/Calypso-Ring/RGObject.extension.st
```diff
@@ -0,0 +1,8 @@
+Extension { #name : 'RGObject' }
+
+{ #category : '*Calypso-Ring' }
+RGObject >> realClass [
+	"Answer the <Class> representing the receiver"
+
+	^ Smalltalk at: self resolvedNameOrNil
```
Ducasse: Same here. 
We should not hardcode Smalltalk.

jecisc:
We get `receiver of "selectedClassOrMetaClass" is nil` in a lot of tests

hernanmd:
> We get `receiver of "selectedClassOrMetaClass" is nil` in a lot of tests

I will have a look, thanks for letting me know

hernanmd:
I've sent a fix but the CI is hanging?

jecisc:
image

jecisc:
I see failing tests related to Job but I don't think it comes from this PR?

hernanmd:
> I see failing tests related to Job but I don't think it comes from this PR?

Not related but I haven't see new tests failing.
The windows build is broken because of Epicea failures :(

jecisc:
In fact some changes here are wrong because it does not work to browse code that is not in the image and it breaks Calypso on ring models. 

I'll do a PR to fix this and with Calypso working on Ring models we will be able to remove the snapshot browser!

Issue #15408, opened 2023-11-17
Wrong text selection: selection selects an extra character


Ducasse:
if you look for a piece of text in a method the marked found text is too large.
For example select `does not define`
look for all the methods with it
you will see that `does not define ` is selected and not `does not define`

hernanmd:
I couldn't reproduce the issue:

<img width="1486" alt="Screenshot 2024-01-22 at 13 55 50" src="https://github.com/pharo-project/pharo/assets/4825959/d4693592-bc79-485f-af34-3fa07689558a">

I am missing something?

Ducasse:
When I do it it selects an extra character
Steps to reproduce
- does not define in the method or class pane not quote
- select it and not more than the visible characters
- Source with it in the Search. 
- You get the following

Ducasse:
![Capture 2024-01-22 at 14 04 35](https://github.com/pharo-project/pharo/assets/432549/a4f99524-b70e-4608-ace6-c1bb6d5fbaab)

jecisc:
In a playground menu it is the option "Code search" -> "method source with it..."

hernanmd:
> In a playground menu it is the option "Code search" -> "method source with it..."

Yes, I've got it but the result is what I uploaded above in the screenshot.
I don't see the problem

Ducasse:
I'm trying on Pharo 12.0.0
Build information: Pharo-12.0.0+SNAPSHOT.build.1311.sha.179be510706e561f0f195fe34f76f5f1aedcdf4c (64 Bit)

Ducasse:
And I redid it with the same image without running any preferences.

Ducasse:
@jecisc can you reproduce it because I can systematically

jecisc:
From the playground I cannot but I can from Calypso

hernanmd:
This is because there are three implementations for it

- The one in `methodSourceContainingIt` from Rubric in RubSmalltalkEditor
- The one in `ClyTextEditor`
- The one in `SpBrowseMethodsContainingStringCommand`
 
So I will start removing the Rubric method and add the command to the Source code menu.

Ducasse:
No we should do the inverse. 
This is the responsibility of the text editor to call the correct methods and not the command.

hernanmd:
Found it, seems like a bug in text selection.

[...]

PR #16010, opened 2024-01-22
Fix for Wrong text selection: selection selects an extra character

hernanmd:
This PR handles #15408.

Here the duplicated find in source behavior is cleaned from Rubric, so we have one implementation based in Command avoiding multiple results as reported in the issue.
This PR introduces a Command which can be now found in the method editor in Calypso as follows:

- Select any text to find.
- Right click in the method editor
- Source code -> Method source with it


Ducasse:
Why removing the functionality from RubEditor? I do not like it. 
The command should be as empty as possible and delegates as much as possible. 
In particular because we want to remove Cmd and only use Cm2.0.

hernanmd:
> Why removing the functionality from RubEditor? I do not like it. The command should be as empty as possible and delegates as much as possible. In particular because we want to remove Cmd and only use Cm2.0.

Thanks for the feedback Stef. I tried to re-use the Commander 2 but there were lot of errors. Anyway, I will rework it so we can use 1 implementation instead of 3 (which differ for the same feature).

Ducasse:
Yes thanks hernan. People do not get that duplication is evil.

jordanmontt (approved):
The code looks good. But I don't know too much about Calypso

jecisc (approved):

PR #16021, opened 2024-01-25
Update bootstrap process to refer to Pharo12

jgfoster:
Now that we are doing development on Pharo12, the documentation and bootstrap process should reflect Pharo12.

### DIFF
--- a/bootstrap/scripts/2-download.sh
+++ b/bootstrap/scripts/2-download.sh
@@ -46,7 +46,7 @@ if [ ! -e "${BOOTSTRAP_DOWNLOADS}/vmBootstrap/pharo" ]; then
 	mkdir ${BOOTSTRAP_DOWNLOADS}/vmBootstrap
 	cd ${BOOTSTRAP_DOWNLOADS}/vmBootstrap
 
-	${BOOTSTRAP_REPOSITORY}/bootstrap/scripts/getPharoVM.sh 100 vm $BOOTSTRAP_ARCH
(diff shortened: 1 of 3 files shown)


bootstrap/scripts/2-download.sh
```diff
@@ -46,7 +46,7 @@ if [ ! -e "${BOOTSTRAP_DOWNLOADS}/vmBootstrap/pharo" ]; then
 	mkdir ${BOOTSTRAP_DOWNLOADS}/vmBootstrap
 	cd ${BOOTSTRAP_DOWNLOADS}/vmBootstrap
 
-	${BOOTSTRAP_REPOSITORY}/bootstrap/scripts/getPharoVM.sh 100 vm $BOOTSTRAP_ARCH
+	${BOOTSTRAP_REPOSITORY}/bootstrap/scripts/getPharoVM.sh 120 vm $BOOTSTRAP_ARCH
```
jecisc: All changes are good to me but this one.
The image initializing the boostrap is still a P10 image. We need to invest some times to update this but it is not something fast. So we need to keep P10 here IIRC

jecisc: Also, IIRC using the P12 VM works on unix but not on OSX. So the CI will pass but we cannot build on mac with this update

jgfoster:
@jecisc – I should ask about the process for "resolving" conversations. Do I do that (if I follow your suggestion) or do I let you do that (to verify that you agree)?

jordanmontt (approved):

PR #16022, opened 2024-01-26
Refactor: change method name preconditions

balsa-sarenac:
Removed the `myConditions` and migrated all preconditions to either applicability or breaking changes.

### DIFF
--- a/src/Refactoring-Core/RBParameterRemoval.class.st
+++ b/src/Refactoring-Core/RBParameterRemoval.class.st
@@ -1,0 +1,60 @@
+"
+I am a common superclass for `InlineParameter` and `RemoveParameter` refactoring.
+I have the common state and methods that are shared between the two refactorings.
+"
+Class {
+	#name : 'RBParameterRemoval',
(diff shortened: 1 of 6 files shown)


balsa-sarenac:
Failing tests seem unrelated

src/Refactoring-Core/RBParameterRemoval.class.st
```diff
@@ -0,0 +1,60 @@
+"
+I am a common superclass for `InlineParameter` and `RemoveParameter` refactoring.
+I have the common state and methods that are shared between the two refactorings.
+"
+Class {
+	#name : 'RBParameterRemoval',
```
jecisc: Shouldn't the class ends with 'Refactoring' to be consistent with names?

balsa-sarenac: I was unsure about this. This one is abstract and will never be used as refactorings, it's just a class that holds state and behavior that are used in both subclasses. But, I've renamed it so should be ready to go now. Thanks!

PR #16020, opened 2024-01-25
Refactor refactoring preconditions

balsa-sarenac:
Continue migrating all preconditions to either applicability or breaking changes.

PR #16023, opened 2024-01-26
Refactor: VariableRefactoring doesn't override preconditions

balsa-sarenac:
This PR inverts `VariableRefactoring`s `preconditions` implementation.

Previously it returned both applicability and breaking chagnes, now it removes only applicability (inherited it from super class).
This eases analysis since now only refactorings that have breaking changes actually implement them (or inherit them) and redefine `preconditions`.


balsa-sarenac:
Should be merged after #16020

PR #16028, opened 2024-01-26
Refactor: rename preconditions of pacakge refactoring

balsa-sarenac:
Rename preconditions to applicability preconditions.


balsa-sarenac:
Should be merged after #16020

PR #16030, opened 2024-01-26
Author removal v2

guillep:
Redo https://github.com/pharo-project/pharo/pull/10830

Lots of cleanups

- remove users of the Author class (and with it the useless popup)
- remove dependency from many users
- `Author` gets deprecated, as there seems to be external users. We need a cycle of adapting users


Ducasse:
Probably a reference in the bootstrap

```
+ /builds/workspace/est_and_branch_Pipeline_PR-16030/64/build/bootstrap-downloads/vmtarget/pharo --headless Pharo12.0-PR-64bit-be79724.image --no-default-preferences clean --release

KeyNotFound: key #Author not found in SystemDictionary

SystemDictionary(Dictionary)>>errorKeyNotFound:

[self errorKeyNotFound: key] in SystemDictionary(Dictionary)>>at: in Block: [self errorKeyNotFound: key]

Syst
```

Ducasse:
Do you know who is referencing this class?

guillep:
No but I think it's not a bad strategy to leave it as deprecated (and not use it in the system).
At the end users will be NOOPs but it will provide an easy migration path.

PR #16035, opened 2024-01-26
add back isGlobalClassNameBinding

MarcusDenker:
... not sure it fixes the real problem, but at least we do not get the DNU

### DIFF
--- a/src/ClassParser/CDSharedVariableNode.class.st
+++ b/src/ClassParser/CDSharedVariableNode.class.st
@@ -37,6 +37,14 @@ CDSharedVariableNode >> isClassVariable [
 ]
 
 { #category : 'testing' }
+CDSharedVariableNode >> isGlobalClassNameBinding [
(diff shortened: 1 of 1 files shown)


Ducasse:
```
TestFailure
Assertion failed
RBFormatterTest(TestAsserter)>>assert:description:resumable:
RBFormatterTest(TestAsserter)>>assert:description:
RBFormatterTest(Object)>>assert:
[ :err :pos | self assert: false ] in RBFormatterTest>>formatClass:selector:
FullBlockClosure(BlockClosure)>>cull:cull:
FullBlockClosure(BlockClosure)>>cull:cull:cull:
```

Ducasse:
I would not integrate it because it does not solve the real problem that we lost the AST based menu.

guillep:
Hi @MarcusDenker could you give some more extra details here? The issue is not that self-explicative :)

PR #16040, opened 2024-01-27
Move the call of the Deprecation>>transform to the Deprecation>>defaultAction

jbrichau:
This PR changes how transforming deprecations get executed. 

Before this PR, transforming deprecations were only signaled as exceptions if the transformation was not possible, which then made it fall back to 'normal' deprecation behavior (i.e. transcript log or dialog warning). Because of this difference, a blocking dialog was never raised for transforming deprecations. It was also not possible to capture a transforming deprecation in an error handler, like a 'normal' deprecation.

In this PR, instead of calling `Deprecation>>transform`, we call `Deprecation>>signal` at all times and call `Deprecation>>transform` from `Deprecation>>defaultAction`. We also add a testcase for the transforming deprecation.

For now, the only observable result is that a transforming deprecation now also respects the 'open blocking dialog' setting. A next step could be to use the blocking dialog to ask the user for a more informative action on wether or not to execute the transformation. 

The testcase added in this PR can only work after merging the PR https://github.com/pharo-project/pharo/pull/16036 since it depends on correct behaviour of deprecation warnings in SUnit testcases.

### DIFF
--- a/src/Kernel-Tests/DeprecationTest.class.st
+++ b/src/Kernel-Tests/DeprecationTest.class.st
@@ -1,0 +1,36 @@
+Class {
+	#name : 'DeprecationTest',
+	#superclass : 'TestCase',
+	#category : 'Kernel-Tests-Exceptions',
+	#package : 'Kernel-Tests',
+	#tag : 'Exceptions'
+}
+
+{ #category : 'tests' }
+DeprecationTest >> testTransformingDeprecation [
+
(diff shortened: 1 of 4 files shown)


Ducasse:
Great. Marcus was really happy about that change.

jbrichau:
@Ducasse Thanks! Mind that the included testcase is failing as long as https://github.com/pharo-project/pharo/pull/16036 is not merged also.

Issue #16029, opened 2024-01-26
asUncommentedCode is under specified


Pythonkito:
self assert: '"Hello" ' asUncommentedCode equals: 'Hello'
We don't know what is result if there is a space.
In addition, asUncommentedCode is not used

PR #16047, opened 2024-01-28
Deprecate unused method asUncommentedCode

hernanmd:
Reported in #16029 this method has no users and it is a reminiscent of the old SmalltalkEditor, not present anymore in Pharo 12.


jecisc (commented):
I wonder if this should not be done via deprecation

hernanmd:
> I wonder if this should not be done via deprecation

I checked Github and couldn't find users. Maybe on GitLab but I don't find a critical usage scenario neither.
So I don't know.

astares:
As it is impossible to check all users (all public repos or private projects) using a deprecation is the better practice

hernanmd:
> As it is impossible to check all users (all public repos or private projects) using a deprecation is the better practice

Done

jecisc:
Some class renaming got mixed up in the last commit I think

astares:
@hernanmd 
When you not only deprecate it but also change the package to "Deprecated12" (so it becomes an extension on String from this package) it will also get removed in Pharo 13 iteration when "Deprecated12" package is removed and a new "Deprecated13" package is created. 

This way we typically keep the method as deprecated method within next release (here Pharo 12). This give users a chance to migrate when moving from P11 to P12 to P13. Then in P13 release it will not be there anywmore. So we keep the deprecated for a single iteration before removing.

hernanmd:
> @hernanmd When you not only deprecate it but also change the package to "Deprecated12" (so it becomes an extension on String from this package) it will also get removed in Pharo 13 iteration when "Deprecated12" package is removed and a new "Deprecated13" package is created.
> 
> This way we typically keep the method as deprecated method within next release (here Pharo 12). This give users a chance to migrate when moving from P11 to P12 to P13. Then in P13 release it will not be there anywmore. So we keep the deprecated for a single iteration before removing.

Thank you for the suggestions. If I understood correctly the deprecation process I will move the method to Deprecated12.

astares (approved):

hernanmd:
Replaced by #16081

PR #16058, opened 2024-01-29
Display methods should be in "displaying" protocol

### LINKED ISSUE #16049: Display methods should be in "displaying" protocol
For an object traditionally in Smalltalk there is a clear distinction between 

1. "printing" (like #printString, #printOn:, #fullPrintString, ....) methods 
     printing the object in the debugger or playground 

2. "displaying" methods (like #displayString, #displayStringLimitedTo:, printStringLimitedTo:, ...)
    displaying the object on a user interface

3. "storing" methods (like #storeString, ...)
    To give a representation of the object that is stored on a stream but can be used to recreate the 
    object

which you can find in any decent Smalltalk implementation.
    
Pharo ignores these protocols at the moment and the way we categorize them in the standard image (ist often wrong and) is not consistent.

We should improve on that level.

astares:
Fix #16049


src/Tool-DependencyAnalyser/DAPackageDependency.class.st
```diff
@@ -22,8 +22,9 @@ DAPackageDependency class >> from: aSourcePackage to: aDestinationPackage [
 	^ self new initializeFrom: aSourcePackage to: aDestinationPackage
 ]
 
-{ #category : 'printing' }
+{ #category : 'displaying' }
```
jecisc, line 25: Personally I find that printing and displaying are kind of the same. I'm not sure I'd like two protocols for a usage that is so close. When browsing code I'd prefer to have both grouped and not have a thousand protocols with really few methods in each

astares: When they would be the same - why make a distinction in selector then? Why call one set of variants "displayXXX", while other variants "printXXX" and a third one "storeXXX". Because they serve a different purpose and this should be reflected in the protocol name too.

You can call me one of those grumpy old Smalltalkers (TM) around - but this is useful and more explicit.
To my knowledge in most of the different Smalltalk systems (commercial and non-commercial) such a distinction in name and protocol is made. And it made perfectly sense - especially when you work more extensively with UI where displaying an object is used more often on the "external" (exposed) user interface while printing is used more for "internal" printable representations in inspectors and dev tools like the debugger.

See for instance Discord posts like this:
image

PR #16065, opened 2024-01-29
We should add Fuel as <externalProject>

### LINKED ISSUE #16064: We should add Fuel as <externalProject>
The class BaselineOfPharo lists all external projects - we shoul list Fuel there as well (so it appears in Iceberg list)

astares:
Fix #16064

### DIFF
--- a/src/BaselineOfPharo/BaselineOfPharo.class.st
+++ b/src/BaselineOfPharo/BaselineOfPharo.class.st
@@ -70,6 +70,23 @@ BaselineOfPharo class >> externalProjects [
 ]
 
 { #category : 'accessing - external projects' }
+BaselineOfPharo class >> fuel [
+	<externalProject>
+	
+	^ PharoExternalProject 
+		newName: 'Fuel' 
+		owner: 'theseion' 
+		project: 'Fuel' 
+		version: 'Pharo12'
+]
+
+{ #category : 'repository urls' }
+BaselineOfPharo class >> fuelRepository [
+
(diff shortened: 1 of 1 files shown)


jecisc:
I'm not sure if this is for the projects that exists outside ou if it's for the projects that are loaded in Pharo from outside. And Fuel is copied in Pharo currently and not loaded from outside.

I'm waiting for a review from someone with more knowledge on this before doing anything

astares:
I just wanted to make it visible (thought we use the Pharo12 branch of Fuel)

Unfortunately situation is not the best, and confusing:
- some code is copied/forked
- some projects we use based on a "Pharo12" branch into image build
- for others we create tagged releases that get included (and that require tagging inbetween)
- some code we load via a menu to extend the image
- some old projects on GH are left giving the impression of things are abondened

Do we want/need to check each class/method we change separately on how it gets into the image?

Back in 2008 a strange setup of Squeak release management lead to Pharo as a fork.

astares:
@jecisc yes - Fuel code is copied into Pharo repo and seems synched from time to time. 
Maybe it is more easy to handle that way for Bootstrap process - I do not know the real reason.

Yesterday I contributed https://github.com/theseion/Fuel/pull/292 and wondered why it was not listed as external package.
Now today I was reminded again that it was copied, so now I did the same change in https://github.com/pharo-project/pharo/pull/16071

Anyhow: as we do not require a <externalButCopiedAndSometimesSynchedBackProject> pragma I guess we can just close this one here. But situation is still confusing and not optimal.

PR #16068, opened 2024-01-29
Remove UIManager dependencies from Compression.

hernanmd:
This PR is a continuation of the work started in #15865 (auto-closed because of conflicts (?))
The summary of changes are:

- Moved utility methods to new package System-Spec.
- Add utility methods as extension to Compression classes.
- Use `openDirectoryDialog`.
- Updated Baseline to load System-Spec.

Probably more changes are needed, but let's see the CI first.


guillep (approved):
Ok!

I don't know if `System-Spec` is the best name for the package, as it is not really related to Spec but the tool UIs, right? What do you think?

PR #16069, opened 2024-01-29
Array>>#displayString should be part of Kernel and not an override from NewTools

### LINKED ISSUE #16050: Array>>#displayString should be part of Kernel and not an override from NewTools
We should put that directly into the kernel as Array is a Kernel class and therefore should have its #displayString method there:

image

astares:
Fix #16050


guillep (approved):
Good!

PR #16072, opened 2024-01-29
Implement Process>>state as the single source of truth for #isX methods (implements #14904)

daniels220:
Fill out the full set of testing methods (add #isReady, #isBlocked) and reimplement the existing ones in terms of #state. Use new #isReady in ProcessorScheduler>>interpriorityYield:

### DIFF
--- a/src/Kernel/Process.class.st
+++ b/src/Kernel/Process.class.st
@@ -332,31 +332,31 @@ Process >> installEnvIntoForked: newProcess [
 { #category : 'testing' }
 Process >> isActiveProcess [
 
-	^ self == Processor activeProcess
+	^ self state == #active
+]
+
+{ #category : 'testing' }
+Process >> isBlocked [
+
+	^ self state == #blocked
+]
+
+{ #category : 'testing' }
+Process >> isReady [
+
+	^ self state == #ready
 ]
 
 { #category : 'testing' }
 Process >> isSuspended [
-	"Answer true if I was never scheduled yet (new process, never been sent #resume) or paused (was sent #suspend)"
-	self isActiveProcess ifTrue: [ ^false ].
-	self isTerminated ifTrue: [ ^false ].
-	^myList isNil or: [ myList isEmpty ]
+
+	^ self state == #suspended
 ]
 
 { #category : 'testing' }
 Process >> isTerminated [
 
-	"Answer if the receiver is terminated, i.e. if the receiver is not active and
-	one of the following conditions is met:
-	(1) the receiver is a defunct process (suspendedContext = nil or pc = nil)
-	(2) the receiver is suspended in the endProcess method (We cannot use pragmas... pragmas require the compiler, it generates an evil dependency.
-	It also is not available in minimal images)"
-
-	self isActiveProcess ifTrue: [ ^ false ].
-
-	^suspendedContext isNil or: [
-		suspendedContext isDead or: [
-			(suspendedContext method == (self class >> #endProcess))]]
+	^ self state == #terminated
 ]
 
 { #category : 'testing' }
@@ -602,6 +602,32 @@ Process >> signalException: anException [
 	process is resumed, it will signal the exception"
 
 	oldList ifNotNil: [self resume]
+]
+
+{ #category : 'testing' }
+Process >> state [
+
+	self == Processor activeProcess ifTrue: [ ^ #active ].
+	"A process is #terminated if one of the following conditions is met:
+	(1) the receiver is a defunct process (suspendedContext = nil or pc = nil)
+	(2) the receiver is suspended in the endProcess method (we cannot use pragmas...
+	pragmas require the compiler, it generates an evil dependency.
+	It also is not available in minimal images)"
+	(suspendedContext isNil or: [
+		 suspendedContext isDead or: [
+			 suspendedContext method == (self class >> #endProcess) ] ])
+		ifTrue: [ ^ #terminated ].
+	"Any other process without a suspendingList, or which cannot
+	possibly be present in that list because it is empty,
+	is considered #suspended. This is either a new process that has
+	never been sent #resume, or one that has been sent #suspend."
+	(myList isNil or: [ myList isEmpty ]) ifTrue: [ ^ #suspended ].
+	"We term a process that is waiting in a list on the Processor #ready,
+	because it will run as soon as the Processor is free. Any other type of list,
+	usually a Semaphore, is considered #blocked"
+	^ myList class == ProcessList
+		  ifTrue: [ #ready ]
+		  ifFalse: [ #blocked ]
 ]
 
 { #category : 'changing suspended state' }
--- a/src/Kernel/Process.class.st
+++ b/src/Kernel/Process.class.st
@@ -330,31 +330,51 @@ Process >> installEnvIntoForked: newProcess [
 ]
 
 { #category : 'testing' }
-Process >> isActiveProcess [
+Process >> isActive [
+	"A process is #active if it is currently executing.
+	There can be only one such process, held in a slot on the Processor."
 
 	^ self state == #active
 ]
 
 { #category : 'testing' }
-Process >> isBlocked [
+Process >> isActiveProcess [
+	"A process is #active if it is currently executing.
+	There can be only one such process, held in a slot on the Processor."
 
-	^ self state == #blocked
+	self
+		deprecated: 'Please use #isActive instead'
+		transformWith: '`@receiver isActiveProcess' -> '`@receiver isActive'.
+	^ self isActive
 ]
 
 { #category : 'testing' }
 Process >> isReady [
+	"A #ready process is one that is waiting on one of the priority lists
+	on the Processor, and will run as soon as there are no others ahead of it
+	or at a higher priority (i.e. the Processor is free)."
 
 	^ self state == #ready
 ]
 
 { #category : 'testing' }
 Process >> isSuspended [
+	"A #suspended process will not run until sent #resume. New processes
+	start in this state (though the #fork family sends #resume internally),
+	and can be put back by sending #suspend. Effectively this is the 'default'
+	state, recognized by the absence of any of the others--not active, not
+	terminated, no list on which to be #waiting or #ready."
 
 	^ self state == #suspended
 ]
 
 { #category : 'testing' }
 Process >> isTerminated [
+	"A process is #terminated if one of the following conditions is met:
+	(1) the receiver is a defunct process (suspendedContext = nil or pc = nil)
+	(2) the receiver is suspended in the endProcess method (we cannot use pragmas...
+	pragmas require the compiler, it generates an evil dependency.
+	It also is not available in minimal images)"
 
 	^ self state == #terminated
 ]
@@ -363,6 +383,17 @@ Process >> isTerminated [
 Process >> isTerminating [
 	"lazy initialization is a fallback only for processes that existed before this addition"
 	^ terminating ifNil: [ false ]
+]
+
+{ #category : 'testing' }
+Process >> isWaiting [
+	"A #waiting process is one that is waiting for a Semaphore to be signaled
+	(having previously executed a #wait on that Semaphore, thus the name).
+	This differs from a #ready process, which also is 'waiting' in a list,
+	in that the Semaphore must be explicitly signaled rather than just
+	the Processor being free."
+
+	^ self state == #waiting
 ]
 
 { #category : 'accessing' }
@@ -606,7 +637,8 @@ Process >> signalException: anException [
 
 { #category : 'testing' }
 Process >> state [
-
+	"A process is #active if it is currently executing.
+	There can be only one such process, held in a slot on the Processor."
 	self == Processor activeProcess ifTrue: [ ^ #active ].
 	"A process is #terminated if one of the following conditions is met:
 	(1) the receiver is a defunct process (suspendedContext = nil or pc = nil)
@@ -623,10 +655,11 @@ Process >> state [
 	myList ifNil: [ ^ #suspended ].
 	"We term a process that is waiting in a list on the Processor #ready,
 	because it will run as soon as the Processor is free. Any other type of list,
-	usually a Semaphore, is considered #blocked"
+	usually a Semaphore, is considered #waiting, as the usual way to reach this state
+	is by executing e.g. Semaphore>>wait."
 	^ myList class == ProcessList
 		  ifTrue: [ #ready ]
-		  ifFalse: [ #blocked ]
+		  ifFalse: [ #waiting ]
 ]
 
 { #category : 'changing suspended state' }
@@ -729,11 +762,11 @@ Process >> suspendingList [
 ]
 
 { #category : 'changing process state' }
-Process >> terminate [ 
+Process >> terminate [
 	"Stop the process that the receiver represents forever. Unwind to execute pending ensure: and
 	ifCurtailed: blocks before terminating; allow all unwind blocks to run; if they are currently in
 	progress, let them finish. If the process is in the middle of a #critical: section, release it properly."
-	
+
 	"If terminating the active process, create a parallel stack and run unwinds from there;
 	if terminating a suspended process, create a parallel stack for the process being terminated
 	and resume the suspended process to complete its termination from the parallel stack. Use
@@ -750,8 +783,8 @@ Process >> terminate [
 	so that we leave the ensure: block inside Semaphore>>critical: without signaling the semaphore.
 	Execute termination wrapped in #ensure to ensure it completes even if the terminator
 	process itself gets terminated before it's finished; see testTerminateInTerminate."
-	
-	self isActiveProcess 
-		ifTrue: [self doTerminationFromYourself]
-		ifFalse: [self doTerminationFromAnotherProcess]
+
+	self isActive
+		ifTrue: [ self doTerminationFromYourself ]
+		ifFalse: [ self doTerminationFromAnotherProcess ]
 ]
--- a/src/Kernel/Process.class.st
+++ b/src/Kernel/Process.class.st
@@ -617,11 +617,10 @@ Process >> state [
 		 suspendedContext isDead or: [
 			 suspendedContext method == (self class >> #endProcess) ] ])
 		ifTrue: [ ^ #terminated ].
-	"Any other process without a suspendingList, or which cannot
-	possibly be present in that list because it is empty,
-	is considered #suspended. This is either a new process that has
-	never been sent #resume, or one that has been sent #suspend."
-	(myList isNil or: [ myList isEmpty ]) ifTrue: [ ^ #suspended ].
+	"Any other process without a list is considered #suspended.
+	This is either a new process that has never been sent #resume,
+	or one that has been sent #suspend."
+	myList ifNil: [ ^ #suspended ].
 	"We term a process that is waiting in a list on the Processor #ready,
 	because it will run as soon as the Processor is free. Any other type of list,
 	usually a Semaphore, is considered #blocked"
--- a/src/Kernel/Process.class.st
+++ b/src/Kernel/Process.class.st
@@ -413,14 +413,14 @@ Process >> on: exception do: handlerAction [
 	"This method inject new bottom context into process with exception handler.
 	It uses context jump tricks to achieve it"
 	| currentContext root newRoot |
-	currentContext := self isActiveProcess ifTrue: [ thisContext ] ifFalse: [self suspendedContext].
+	currentContext := self isActive ifTrue: [ thisContext ] ifFalse: [self suspendedContext].
 	root := currentContext bottomContext.
 	newRoot := [
 			[root insertSender: thisContext.
 			currentContext jump] on: exception do: handlerAction.
 		Processor terminateRealActive] asContext.
 
-	self isActiveProcess
+	self isActive
 		ifTrue: [ newRoot jump ]
 		ifFalse: [ self install: newRoot ]
 ]
@@ -527,7 +527,7 @@ Process >> pvtSignal: anException list: aList [
 	have the potential to push an unwanted object on the caller's stack)."
 	<debuggerCompleteToSender>
 	| blocker |
-	self isActiveProcess ifFalse: [^self].
+	self isActive ifFalse: [^self].
 	anException signal.
 	blocker := Semaphore new.
 	[self suspend.
@@ -614,7 +614,7 @@ Process >> signalException: anException [
 	will return to a blocked state unless the blocking Semaphore has excess signals"
 	| oldList |
 	"If we are the active process, go ahead and signal the exception"
-	self isActiveProcess ifTrue: [^anException signal].
+	self isActive ifTrue: [^anException signal].
 
 	"Suspend myself first to ensure that I won't run away in the
 	midst of the following modifications."
--- a/src/Debugger-Oups/Process.extension.st
+++ b/src/Debugger-Oups/Process.extension.st
@@ -10,7 +10,7 @@ Process >> debugWithTitle: title [
 
 	<debuggerCompleteToSender>
 	| context |
-	context := self isActiveProcess
+	context := self isActive
 		           ifTrue: [ thisContext ]
 		           ifFalse: [ self suspendedContext ].
 
--- a/src/Kernel-Extended-Tests/ProcessTest.class.st
+++ b/src/Kernel-Extended-Tests/ProcessTest.class.st
@@ -340,6 +340,7 @@ ProcessTest >> testIsNotSuspendedWhenItIsRunningButNotActiveProcess [
 	semaphore := Semaphore new.
 	process := [ semaphore signal. Processor yield ] fork.
 	semaphore wait.
+	self assert: process isReady.
 	self deny: process isTerminated.
 	self deny: process isSuspended
 ]
@@ -554,6 +555,20 @@ ProcessTest >> testIsTerminatingNormalTermination [
 	self assert: unwound.
 	"A process should terminte itself."
 	self assert: terminator identicalTo: process
+]
+
+{ #category : 'tests' }
+ProcessTest >> testIsWaiting [
+
+	| semaphore process |
+	semaphore := Semaphore new.
+	process := [ semaphore wait. Processor yield.  ] fork.
+	self assert: process isReady.
+	self deny: process isWaiting.
+	Processor yield.
+	self assert: process isWaiting.
+	semaphore signal.
+	self deny: process isWaiting.
 ]
 
 { #category : 'tests - creation' }
--- a/src/Kernel/ProcessorScheduler.class.st
+++ b/src/Kernel/ProcessorScheduler.class.st
@@ -192,7 +192,7 @@ ProcessorScheduler >> interpriorityYield: modifiedProcess [
 			[ modifiedProcess resume ] fork.
 			modifiedProcess suspend ]
 		ifFalse: [
-			(modifiedProcess suspendingList isKindOf: ProcessList) ifTrue: [
+			modifiedProcess isReady ifTrue: [
 				modifiedProcess
 					suspend;
 					resume ] ]
--- a/src/Kernel/ProcessorScheduler.class.st
+++ b/src/Kernel/ProcessorScheduler.class.st
@@ -187,7 +187,7 @@ ProcessorScheduler >> interpriorityYield: modifiedProcess [
 	The only application so far is when changing the priority of a process.
 	"
 
-	modifiedProcess isActiveProcess
+	modifiedProcess isActive
 		ifTrue: [
 			[ modifiedProcess resume ] fork.
 			modifiedProcess suspend ]
--- a/src/UIManager/CommandLineUIManager.class.st
+++ b/src/UIManager/CommandLineUIManager.class.st
@@ -377,7 +377,7 @@ CommandLineUIManager >> quitFrom: aContext withMessage: aString [
 		Process allInstances do: [:each | | ctx |
 			logger nextPutAll: 'Process: '; print: each; cr; nextPutAll: '  stack:'; cr; cr.
 
-			ctx := each isActiveProcess ifTrue: [ thisContext sender ] ifFalse: [ each suspendedContext ].
+			ctx := each isActive ifTrue: [ thisContext sender ] ifFalse: [ each suspendedContext ].
 			ctx ifNotNil: [
 				(ctx stackOfSize: 20) do: [:s | logger print: s; cr ]].
 			logger nextPutAll: '------------------------------'; cr; cr.


jecisc (approved):
Nice!

jecisc:
Failing tests are not related. I'd merge but I think it might be nice to let some time for more people to take a look?

src/Kernel/Process.class.st
```diff
@@ -604,6 +604,32 @@ Process >> signalException: anException [
...
+	It also is not available in minimal images)"
+	(suspendedContext isNil or: [
+		 suspendedContext isDead or: [
+			 suspendedContext method == (self class >> #endProcess) ] ])
+		ifTrue: [ ^ #terminated ].
+	"Any other process without a suspendingList, or which cannot
+	possibly be present in that list because it is empty,
+	is considered #suspended. This is either a new process that has
+	never been sent #resume, or one that has been sent #suspend."
+	(myList isNil or: [ myList isEmpty ]) ifTrue: [ ^ #suspended ].
```
isCzech: Hi, I think here's an opportunity for further simplification: 
`myList ifNil: [ ^ #suspended ].`
should suffice here, because the other part of the condition (`or: [ myList isEmpty ]`) should no longer be necessary. It used to be relevant for older VMs but unfortunately, I can't find the discussion where this was explained (will post when come across it again). So just FYI :)

daniels220: Oh nice! Yeah, I wondered a little about that—notice I updated the comment, thinking "hmm, we check if the list is empty, but what about if there's another process in that list yet the receiver is not? But that clearly shouldn't happen in a well-designed VM, so why the empty check?" Maybe should have said something, except, it seems like PRs get a lot more engagement than comments on old issues, and I didn't want to delay _making_ the PR anyway. Glad you chimed in!

daniels220:
@jecisc Yeah, I'd be happy with some more review—less for equivalency of this to the old implementation, which I'm _pretty_ sure of, more for "is there anything else we can improve while we're here".

src/Kernel/Process.class.st
isCzech: Hi @daniels220 , FYI I've found the discussion regarding the `myList isEmpty` check (see the third post from the top by Vanessa Freudenberg): https://lists.squeakfoundation.org/archives/list/vm-dev@lists.squeakfoundation.org/thread/CGCZH3WM742AGJ7RUGFHGJAHCRGL77ZL/

Ducasse:
Thanks for the proposal cleaning this part is nice. @isCzech thanks for the reference.

isCzech:
> @jecisc Yeah, I'd be happy with some more review—less for equivalency of this to the old implementation, which I'm _pretty_ sure of, more for "is there anything else we can improve while we're here".

@daniels220, another simplification opportunity is #terminated, but unfortunately, I still haven't figured out a small bit in Pharo.  In Cuis the test looks very elegant (IMHO):
```
isTerminated
	"Answer if the receiver is terminated. A healthy process is considered 
	 terminated if the suspendedContext has no unexecuted code left."

	self isRunning ifTrue: [^ false].
	^suspendedContext isNil or: [suspendedContext isDone]
```
where `isDone` checks whether a context chain consists only from contexts standing at their last instruction, i.e. a return :) But at the moment this approach can't be adopted without some modifications around #terminate.

Ducasse:
A method comment for isReady would help to understand. Because I have no idea what is a ready process.

Ducasse:
I wonder where the state names are coming from?
I do not understand what is blocked too.

Here is the previous vocabulary and I would prefer to be close to it. Check the book

We define the states in which a process can be:
• executing - the process is currently executing.
• runnable - the process is scheduled. This process is in one of the prior- ity lists of the scheduler.
• terminated - the process has run and finished its execution. It is not managed anymore by the scheduler. It cannot be executed anymore.
• suspended - the process is not managed by the scheduler: This process is not in one of the scheduler lists or in a semaphore list. The process can become runnable sending it the resume message.
• waiting - the process is waiting on a semaphore waiting list. It is not managed by the scheduler. The process can become runnable when the semaphore releases it.

isCzech:
Hi @Ducasse , 
Yes, I'm aware of the naming from your "Concurrent Programming In Pharo, 2020". 
image

I always wondered why the "executing" process was called "active" in Pharo - historic reasons? I personally like "blocked" because it's a process waiting on a condition variable as opposed to a runnable aka "ready" process also waiting - on one of the priority lists. I suggested the name "blocked" but indeed do not insist and apologise for a confusion I may have caused :) 

For the sake of consistency, shouldn't a test `#isExecuting` synonymous to `#isActiveProcess` be created so that all the states correspond to the enclosed picture? Thanks

Ducasse:
I'm not against changing. I just recorded what people wrote is other books or implementation. 
Now blocked does not mean waiting for new (waiting goes in pair with wait message and we do not send the blocks message. 

For the sake of consistency, shouldn't a test #isExecuting synonymous to #isActiveProcess be created so that all the states correspond to the enclosed picture?

Probably. 

I do not remember exactly but Guille told that a state is missing and it would be good to enhance this.

dionisiydk:
> I always wondered why the "executing" process was called "active" in Pharo - historic reasons? I personally like "blocked" because it's a process waiting on a condition variable as opposed to a runnable aka "ready" process also waiting - on one of the priority lists. I suggested the name "blocked" but indeed do not insist and apologise for a confusion I may have caused :)

I agree with Stef about the #waiting state. And you just wrote here: "blocked is a process waiting for a condition..". 
From the other side I may also think that the process is actually blocked when it waits for the currently running another process. So it's kind of a same meaning but #waiting state is in sync with waiting primitives and therefore it serves better.

dionisiydk:
> We define the states in which a process can be:
> ..
> • runnable - the process is scheduled. This process is in one of the prior- ity lists of the scheduler.

This definition for the runnable state looks like we should just directly use #scheduled word for this state. Runnable is a bit ambiguous. Process waiting for a semaphore is also runnable from the English meaning of this word.
And I agree that #ready is not self explaining at all.

src/Kernel/Process.class.st
```diff
@@ -604,6 +604,31 @@ Process >> signalException: anException [
...
+	self == Processor activeProcess ifTrue: [ ^ #active ].
+	"A process is #terminated if one of the following conditions is met:
+	(1) the receiver is a defunct process (suspendedContext = nil or pc = nil)
+	(2) the receiver is suspended in the endProcess method (we cannot use pragmas...
+	pragmas require the compiler, it generates an evil dependency.
+	It also is not available in minimal images)"
+	(suspendedContext isNil or: [
+		 suspendedContext isDead or: [
+			 suspendedContext method == (self class >> #endProcess) ] ])
+		ifTrue: [ ^ #terminated ].
```
dionisiydk, line 651: Three conditions for a terminated state is definitely a place to improve the termination. But it is for another step.

For example there are several places where suspendedContext is set to nil in the middle of debugger operations (see  suspendedContext writers). Any stepping over these methods eventually will corrupt the process state.

src/Kernel/Process.class.st
```diff
@@ -604,6 +604,31 @@ Process >> signalException: anException [
 	oldList ifNotNil: [self resume]
 ]
 
+{ #category : 'testing' }
+Process >> state [
```
dionisiydk, line 639: I wonder if #status would be a better name instead of #state. I am not insist , just thinking. 
State may have a different meaning like an entire copy of a process stack. Status sounds less ambiguous.

isCzech:
> waiting goes in pair with wait message

I like this argument :)

> process is actually blocked when it waits for the currently running another process.

ha ha, true :)

> Process waiting for a semaphore is also runnable from the English meaning of this word

no argument there :)

I'm influenced by Squeak/Cuis with "blocked". For comparison: Squeak uses "runnable" for Pharo-runnable combined with active (not sure why), Cuis uses "running" for active aka executing, Dolphin and Cuis use "ready" for Pharo-runnable.

> I wonder if #status would be a better name instead of #state.

Good thing is Pharo has a supporting literature so it makes great sense to go along with it - and add `isExecuting" method. #isActiveProcess is an inconsistent name anyway ;) And the same applies to #state - maybe not the best word but already covered in the documentation (which is a blessing in the Smalltalk world!)

src/Kernel/Process.class.st
isCzech: > But it is for another step.

glad to agree :)

Ducasse:
I would indeed prefer isScheduled over runnable if they represent the same state.

guillep:
I like the discussion, I don't have much to add :) Let's just remember to update the doc xD

Ducasse:
So what is the conclusion at the end?

isCzech:
I have two favorite sets of state names:
(A)
• **executing** - the process is currently executing.
• **runnable** - the process is scheduled for execution and is waiting in one of the scheduler priority lists.
• **waiting** - the process is waiting on a condition variable (e.g. a semaphore) list. It is not managed by the scheduler.
• **suspended** - the process is neither executing or in one of the scheduler lists or in a condition variable list; the process can become runnable by sending it the resume message.
• **terminated** - the process has finished its execution. It is not managed by the scheduler.

(B)
• **running** - the process is currently executing.
• **scheduled** - the process is scheduled for execution and is waiting in one of the scheduler priority lists.
• **blocked** - the process is waiting on a condition variable (e.g. a semaphore) list. It is not managed by the scheduler.
• **suspended** - the process is neither executing or in one of the scheduler lists or in a condition variable list; the process can become runnable by sending it the resume message.
• **terminated** - the process has finished its execution. It is not managed by the scheduler.

(A) 
_pros_: 
- it's covered by the existing literature (e.g. Concurrent Programming In Pharo, 2020 - see picture in a previous post)

_cons_:
- runnable is a misleading name (all processes are runnable in the common sense of the word)
- waiting is a too common name; giving it a precise restricted meaning may be misleading

(B)
_pros_:
- running is closer to a common way of talking about active aka executing processes
- scheduled is a better name
- blocked: the purpose of a semaphore is to halt a process as opposed to a scheduler which just controls the order of processes waiting in priority lines until they can run (i.e. until other processes yield, get blocked on a semaphore, get suspended or just terminate); the term "blocked" allows using the word "wait" more freely (unambiguously) for processes waiting in the scheduler priority lines (you can view a condition variable as a form of a scheduler based on a different set of scheduling rules and argue there's no difference between waiting and blocked - which is basically true - but I still prefer the passive form of blocked over the active form of waiting)

_cons_:
- documentation needs to be updated
- more difficult for newcomers to sync with older terminology
- some method comments may need to be updated

Other notes:
- status: the plural is "statuses" (ugh) or sometimes even "status" (ambiguous), or perhaps "stati" (let's rather not); good old state has the same general meaning
- a newly created process is automatically in a suspended or in border situations a terminated state (which is ok and no "new" state is necessary IMHO - in case you wondered)
- it's not entirely true a terminated process cannot be run again:
	(1) it can be assigned a new suspended context and resumed
	(2) currently a terminated process is a suspended process that fulfills some additional conditions; it can be sent the resume message; it's really not "destroyed" irreparably :)

guillep:
> I have two favorite sets of state names: (A) • **executing** - the process is currently executing. • **runnable** - the process is scheduled for execution and is waiting in one of the scheduler priority lists. • **waiting** - the process is waiting on a condition variable (e.g. a semaphore) list. It is not managed by the scheduler. • **suspended** - the process is neither executing or in one of the scheduler lists or in a condition variable list; the process can become runnable by sending it the resume message. • **terminated** - the process has finished its execution. It is not managed by the scheduler.
> 
> (B) • **running** - the process is currently executing. • **scheduled** - the process is scheduled for execution and is waiting in one of the scheduler priority lists. • **blocked** - the process is waiting on a condition variable (e.g. a semaphore) list. It is not managed by the scheduler. • **suspended** - the process is neither executing or in one of the scheduler lists or in a condition variable list; the process can become runnable by sending it the resume message. • **terminated** - the process has finished its execution. It is not managed by the scheduler.
> 
> (A) _pros_:
> 
>     * it's covered by the existing literature (e.g. Concurrent Programming In Pharo, 2020 - see picture in a previous post)
> 
> 
> _cons_:
> 
>     * runnable is a misleading name (all processes are runnable in the common sense of the word)
> 
>     * waiting is a too common name; giving it a precise restricted meaning may be misleading
> 
> 
> (B) _pros_:
> 
>     * running is closer to a common way of talking about active aka executing processes
> 
>     * scheduled is a better name
> 
>     * blocked: the purpose of a semaphore is to halt a process as opposed to a scheduler which just controls the order of processes waiting in priority lines until they can run (i.e. until other processes yield, get blocked on a semaphore, get suspended or just terminate); the term "blocked" allows using the word "wait" more freely (unambiguously) for processes waiting in the scheduler priority lines (you can view a condition variable as a form of a scheduler based on a different set of scheduling rules and argue there's no difference between waiting and blocked - which is basically true - but I still prefer the passive form of blocked over the active form of waiting)
> 
> 
> _cons_:
> 
>     * documentation needs to be updated
> 
>     * more difficult for newcomers to sync with older terminology
> 
>     * some method comments may need to be updated
> 

I think I like better this second one.
Updating the documentation is not so difficult to do.

> 
> Other notes:
> 
>     * status: the plural is "statuses" (ugh) or sometimes even "status" (ambiguous), or perhaps "stati" (let's rather not); good old state has the same general meaning
> 

+1

>     * a newly created process is automatically in a suspended or in border situations a terminated state (which is ok and no "new" state is necessary IMHO - in case you wondered)
> 

Yes

>     * it's not entirely true a terminated process cannot be run again:
>       (1) it can be assigned a new suspended context and resumed
>       (2) currently a terminated process is a suspended process that fulfills some additional conditions; it can be sent the resume message; it's really not "destroyed" irreparably :)

Yes I think that we can agree that this scenario is too borderline. Whoever is doing such kind of manipulations should acknowledge potential hiccups here :)

Ducasse:
I prefer B
- in particular scheduled is good.
- now blocked does not really tell me much. I can understand that waiting is not clear enough compared to scheduled but at least waiting is in par with the semaphore API. So if we get nothing better I prefer isWaiting over isBlocked

src/Kernel/Process.class.st
daniels220: Cool, thanks! I've updated the code to remove the obsolete test.

daniels220:
> @daniels220, another simplification opportunity is #terminated, but unfortunately, I still haven't figured out a small bit in Pharo. In Cuis the test looks very elegant (IMHO): [...] where `isDone` checks whether a context chain consists only from contexts standing at their last instruction, i.e. a return :) But at the moment this approach can't be adopted without some modifications around #terminate.

I definitely feel a bit dirty looking at the Pharo implementation, yes :). The Cuis one is a lot cleaner, but I have to wonder why it's a context _chain_ consisting only of contexts at their last instruction? When would there ever be more than one? I suppose if it's only ever two or three that's fine, but if it's a scenario where they could "build up" due to particular patterns in user code I'd rather have a constant-time test.

Honestly I'm inclined towards a solution where there's simply a primitive that finishes off a process by removing it from the Processor without saving the suspended context (or suspends and then nils the suspended context), such that the test can be just "not active and `suspendedContext isNil`". This is the implementation in Dolphin Smalltalk, with the following comment about the primitive:

> Implementation Note: This primitive is approximately equivalent to:
> ```
> self suspendUnconditionally. 
> suspendedFrame := nil
> ```
> except that the primitive can correctly terminate any process, even the active process. The active process cannot terminate itself correctly in this way in Smalltalk without the assistance of another process, because as soon as it suspends itself, it will obviously stop running, and so will be unable to nil out its suspended context. It is considered that the use of a simple primitive to terminate processes is both less error prone, and consumes less resources, than the alternative of dedicating another process to this task.

On a related note, I have to say I prefer Dolphin's implementation of `#terminate` itself, as well—instead of `doTerminationYourself` and `doTerminationFromAnotherProcess`, termination from another process is handled by interrupting the process-to-terminate in order to have it do the termination itself. Pharo already has a similar mechanism in `Process>>signalException:`—in fact in Dolphin, process termination is handled by throwing a special `ProcessTermination` exception, but it could just as easily be that `#terminate` builds a frame that invokes a new `#pvtTerminate` instead of the `#pvtSignal:list:` that `#signalException:` uses. I know next to nothing about the issues that have existed around process termination in the past, but I know it was mentioned in earlier discussion that `#isTerminating` is fallout from fixing some issues and that it _should_ be able to go away at some point, so I thought I'd throw this out there and see what people think.

Honestly this whole comment should be on another issue, I just don't know what issue that would be—maybe one that doesn't exist yet, and I'm not sure what I would report. I'll end this comment here, and @dionisiydk @isCzech (and anyone else who's interested) let's take this discussion elsewhere, if someone has a suggestion of where.

src/Kernel/Process.class.st
daniels220: See my comment below—this is, indeed, a whoooole other conversation. I would be delighted if one of you wanted to take the initiative to make this into its own issue—which entails first figuring out what, exactly, that issue should _be_.

isCzech:
> I definitely feel a bit dirty looking at the Pharo implementation, yes :). The Cuis one is a lot cleaner, but I have to wonder why it's a context chain consisting only of contexts at their last instruction? When would there ever be more than one? I suppose if it's only ever two or three that's fine, but if it's a scenario where they could "build up" due to particular patterns in user code I'd rather have a constant-time test.

no, no - just two, perhaps three max; but it was easier to write #isDone recursively... The reason behind was that in Squeak/Cuis if #suspend primitive fails than the fallback #suspendAndUnblock gets executed which ends up with two contexts at their last instruction. I can't remember whether there were other cases, (In case you wonder why #suspend might fail, there's a new primitive behind #suspend that fixed this bug: the previous #suspend simply removed a process from a semaphore while the new #suspend primitive backs the process by one instruction, i.e. before the #wait that sent the process to the semaphore - the difference is obviously what happens when you resume such a suspended process.)

> Honestly I'm inclined towards a solution where there's simply a primitive that finishes off a process by removing it from the Processor without saving the suspended context (or suspends and then nils the suspended context), such that the test can be just "not active and suspendedContext isNil".

precisely my thoughts! It's on my todo list to explore this idea :)

> On a related note, I have to say I prefer Dolphin's implementation of #terminate

I disagree :) Dolphin's termination semantics is so much more limited compared to Squeak/Cuis/Pharo - the difference is in treating non local returns inside unwind blocks as normal workflow. Dolphin fails to execute most of such examples (as well as VisualWorks an VisualAge). If you have Cuis installed, check out Process class>>#terminateExamples

> in Dolphin, process termination is handled by throwing a special ProcessTermination exception

... which is precisely what I didn't like; engaging exceptions felt like too much complexity to terminate a process. I chose a straightforward loop to walk the sender chain and execute all unwind blocks. Another highlight is that the process that does the unwinding is the terminating process itself (so we have process faithful termination, can't remember whether that applies to Dolphin too)

daniels220:
Wow, this suddenly got a lot of comments—thanks to everyone who added their thoughts! And, if I fail to respond to you, please feel free to poke me again, this is a lot to go through. It seems like the majority of the discussion is about what to name the states—here's my 2¢ (more like $2, there's a lot to say):

I definitely prefer `#waiting`/`#isWaiting` over blocked—notice that's what I used in the original issue #14904. I didn't have a _strong_ opinion, so I went with the feedback I got there, but I'm more than happy to change it now.

The thing is, I think more-or-less the same argument applies to `#active`/`#isActive`—the test, after all, is <code>self == Processor <b>active</b>Process</code>. This is also the name of the testing method in Dolphin—though it actually uses `#running` as the state name now that I look at it, so :shrug:. One argument in favor of choosing something else: "active" is a pretty common concept—there are 27 implementors and 77 senders of `#isActive`, `#isExecuting` has none, and `#isRunning` has only 4 implementors and 44 senders. This is I suppose the sole advantage of `#isActiveProcess`—it's completely unique and likely to remain that way—but on balance I still dislike the inconsistency. We have to keep it for backcompat, but I would definitely prefer to deprecate it in favor of either `#isActive`, or if we change the state name, `#isRunning` or `#isExecuting`. I've pushed an update assuming `#isActive`, but without rewriting any senders, so it's still easy to pivot.

Regarding `#ready`, I find it intuitive enough—"ready to run"—but by definition, if not everyone does, it's not totally clear. However I definitely _dislike_ both of the proposed alternatives:
* `isRunnable` sounds like a combined testing method—a union of some of the other states, most likely `#active` and `#ready`. FWIW this is the name used in VisualWorks though.
* `isScheduled` makes me first think of a process waiting on a Delay—"scheduled to run at a particular time". I'm guessing no C or other low-level-language programmer would think this way, as the term "scheduler" is pretty clear there, but then Pharo isn't C.

@Ducasse, you're right that better comments would be helpful. I've taken the comments in `#state` and copied and expanded them out to each of the testing methods as well.

And, I see a mention of a missing state—there is one actually! But it's definitely a separate issue. The missing state is `#debug`—a process that is technically `#suspended`, but specifically because it is being debugged, is very different from one that is suspended as part of normal application operation. However this requires the process to have any practical way to _know_ that it's being debugged, which under the current implementation isn't easy. So I've written #16087 for that.

guillep:
Hi @daniels220 I see you did a rename there, do you think it's good to go?

daniels220:
> Hi @daniels220 I see you did a rename there, do you think it's good to go?

Close, but definitely not—doesn't seem like we've settled on the names. The choice of `#waiting` is the one that we seemed to have reached a consensus, or at least strong majority, on, but the final names for what are now `#active` and `#ready` are still up in the air as far as I'm concerned. (I _personally_ am in favor of them as they are, but that's as much due to my history with Dolphin as anything.) I also need to follow through with rewriting references to `isActiveProcess` once that decision is made—I went ahead with the deprecation tag but didn't actually do a "replace senders" or whatever the menu item is called.

daniels220:
@isCzech, okay, this is fascinating but also pretty far off-topic. If there's a better place to move the discussion, let me know, otherwise, everyone else, there's nothing directly relevant to the issue in this comment.

----

> no, no - just two, perhaps three max; but it was easier to write #isDone recursively... The reason behind was that in Squeak/Cuis if #suspend primitive fails [...]

Right, that sort of implementation detail makes sense as a reason.

> I disagree :) Dolphin's termination semantics is so much more limited compared to Squeak/Cuis/Pharo - the difference is in treating non local returns inside unwind blocks as normal workflow. Dolphin fails to execute most of such examples (as well as VisualWorks an VisualAge). If you have Cuis installed, check out Process class>>#terminateExamples

Oh now those are some _neat_ new semantics! Although I think they are somewhat conflating two things that could/should be separate. Effectively what they've done is make the execution of an `ensure:` handler block uninterruptible by `terminate`, or you could say atomic with respect to termination—the terminate in a sense is deferred until the handler block finishes. While this makes sense to apply to `ensure:` handlers, there are times when it would be useful on its own—I just recently encountered an issue where `ProcessMonitorTestService` was terminating one of my processes in the middle of shutting down a web server, and because this was _not_ actually in an `ensure:` block, it did _not_ finish execution, resulting in subsequent tests breaking, and in some cases an image crash. I just fixed my test so the background process would finish execution before the test completed—in fact this revealed what was arguably a bug in my code, so it's not all bad—but being able to say "you cannot terminate this process in the middle of this operation" would be very nice.

...and in fact I believe you can. I mean, trivially you can say `[] ensure: ["termination won't happen here"]`, but that's ugly. But looking at the implementation of `ensure:` and `ifCurtailed:`, I derived the following (the name comes from the fact that this is effectively a critical section, co-critical with `terminate` itself—though for the analogy to be exact, you have to pretend that `terminate` is always being executed in a _different_ process—it's an analogy, not a strictly-correct description):

```smalltalk
terminateCritical
	| handler complete |
	<primitive: 198>
	handler := self.
	complete := true.
	^self valueNoContextSwitch.
```

And sure enough, if I evaluate:

```smalltalk
[ [ Transcript show: 'before'.
Processor activeProcess terminate.
Transcript show: 'after' ] terminateCritical ] fork
```

I see `beforeafter`. Really all I've done is strip out all the bits of `#ensure:` that involve the block that is normally the receiver, and make the parts involving the handler block refer to the receiver...

I'm not sure what exactly you mean by "treating non local returns inside unwind blocks as normal workflow"? As far as I can see, the new Cuis behavior doesn't affect non-local returns at all, only calls to `#terminate` (whether explicit in-proc, explicit from another proc, or implicit by closing a debugger—AFAICT "unhandled exception" is a red herring, it's just another way of causing a `#terminate` call). So for instance Dolphin executes this example:

```smalltalk
[
	[ ] ensure: [
		[self error: #unwind_test] ensure: [
			^Transcript show: #x1]. 
		Transcript show: #x2]
] ensure: [
	Transcript show: #x3].
Transcript show: #x4
```

the same as Pharo/Cuis, but that's because Pharo/Cuis don't do anything special here, none of the new behavior is invoked. Pretty much all the rest of the examples, you're right that Dolphin's semantics are more limited, it will abandon an `ensure:`'d block halfway through on another `#terminate`.

Actually, what's more interesting/concerning is that Dolphin will actually _fail to terminate_ a process due to a nonlocal return in an `ensure:` block. This isn't something you can demonstrate in a workspace, because you need multiple method contexts on the stack to have a valid home for the block to return to, that will then execute more code. I'll have to whip up an example with a dummy class and see how this behaves in Pharo...

> ... which is precisely what I didn't like; engaging exceptions felt like too much complexity to terminate a process. I chose a straightforward loop to walk the sender chain and execute all unwind blocks. Another highlight is that the process that does the unwinding is the terminating process itself (so we have process faithful termination, can't remember whether that applies to Dolphin too)

So I realize that I was under a misconception about Pharo's implementation, which led me to say some misleading things. The part I thought was different about Dolphin is precisely that the process doing the unwinding is the terminating process  itself—I somehow thought `#terminate` used a simulation mechanism like the debugger. I had reason to look at it more closely and I see that it's doing exactly what I thought it should do, and what Dolphin does—building a new frame/context and manually placing it on top of the process-to-terminate's stack, then allowing the code therein to do the work. Dolphin has this as a VM interrupt facility rather than constructing the context by hand, but you actually _could_ do it by hand and it would still work (well, only probably, for reasons not worth explaining). I only brought up `#signalException:` in the first place because I saw that that does "interrupt" the process (by the same manually-add-a-new-top-context approach), and thought that was a point of contrast rather than similarity.

I...mostly agree that the use of an exception is an ugly way to terminate a process. I think I've made use of it a couple times in development code, but only when doing things that I would very much never do in production code. For instance, I wrote a facility that allows for:

```smalltalk
[ guard1 ifTrue: [^false].
guard2 ifFalse: [^false].
obj someActualComputation] withInterceptedReturns
```

which wraps the block such that, when it is sent `value`, the `^`-returns will return to the sender of `value`, not the home context of the block. Very handy in a workspace, not something I'd touch with a 10-foot pole in production. And IIRC, my _original_ implementation looked for `ProcessTermination` so the block wouldn't prevent the process from being terminated, but this caused other problems and I eventually got smarter with my stack introspection and was able to do away with it—so perhaps that's an argument against the use of the exception, as well.

isCzech:
> > Hi @daniels220 I see you did a rename there, do you think it's good to go?
> 
> Close, but definitely not—doesn't seem like we've settled on the names. The choice of `#waiting` is the one that we seemed to have reached a consensus, or at least strong majority, on, but the final names for what are now `#active` and `#ready` are still up in the air as far as I'm concerned. (I _personally_ am in favor of them as they are, but that's as much due to my history with Dolphin as anything.) I also need to follow through with rewriting references to `isActiveProcess` once that decision is made—I went ahead with the deprecation tag but didn't actually do a "replace senders" or whatever the menu item is called.

I'm happy with active/ready/waiting/suspended/terminated if there's consensus.

> isScheduled makes me first think of a process waiting on a Delay—"scheduled to run at a particular time"

probably a good reason for "#scheduled" to go...

> there are 27 implementors and 77 senders of #isActive

in case this is a problem I'm happy with both #running and #executing too

> The choice of `#waiting` is the one that we seemed to have reached a consensus

Yes, #waiting is most common; just out of curiosity I checked some literature about condition variables and both #waiting and #blocked are actually two sides of the same coin: (1) a process is waiting after (2) the process gets blocked on a condition variable.

guillep:
> > > Hi @daniels220 I see you did a rename there, do you think it's good to go?
> > 
> > 
> > Close, but definitely not—doesn't seem like we've settled on the names. The choice of `#waiting` is the one that we seemed to have reached a consensus, or at least strong majority, on, but the final names for what are now `#active` and `#ready` are still up in the air as far as I'm concerned. (I _personally_ am in favor of them as they are, but that's as much due to my history with Dolphin as anything.) I also need to follow through with rewriting references to `isActiveProcess` once that decision is made—I went ahead with the deprecation tag but didn't actually do a "replace senders" or whatever the menu item is called.

haha this is typically an infinite discussion ^^.

> 
> I'm happy with active/ready/waiting/suspended/terminated if there's consensus.

All sets of names chosen are actually good.
I'm good with any of the schemas said above them as soon as they are documented :)
 
> > there are 27 implementors and 77 senders of #isActive
> 
> in case this is a problem I'm happy with both #running and #executing too

I don't think this is a blocking issue as soon as we propose a rewriting + a deprecation ^^

@daniels220 what is your preferred set of names? I think we can settle on those if there are no detractors, because you are pushing the issue :).

isCzech:
Hi @daniels220, 

> @isCzech, okay, this is fascinating but also pretty far off-topic. If there's a better place to move the discussion, let me know, otherwise, everyone else, there's nothing directly relevant to the issue in this comment.

I don't know if it's even possible to move a discussion without breaking the thread... If nobody minds we can perhaps post a few more unrelated comments.

---

> Oh now those are some _neat_ new semantics! Although I think they are somewhat conflating two things that could/should be separate. Effectively what they've done is make the execution of an `ensure:` handler block uninterruptible by `terminate`, or you could say atomic with respect to termination—the terminate in a sense is deferred until the handler block finishes. 

that's an interesting angle; however, by definition termination is _supposed_ to execute whatever has been put into the unwind blocks.

> While this makes sense to apply to `ensure:` handlers, there are times when it would be useful on its own—[...] being able to say "you cannot terminate this process in the middle of this operation" would be very nice.
> 
> ...and in fact I believe you can. I mean, trivially you can say `[] ensure: ["termination won't happen here"]`

that's exactly what we do in #doTerminationFromAnotherProcess; we want to make sure a termination won't get terminated by another termination (from some other process) - see e.g. #testTerminateInTerminate.

> , but that's ugly. 

won't argue :) Cuis wrapped the construct to
```
valueEnsured
	[] ensure: self
```
> But looking at the implementation of `ensure:` and `ifCurtailed:`, I derived the following (the name comes from the fact that this is effectively a critical section, co-critical with `terminate` itself—though for the analogy to be exact, you have to pretend that `terminate` is always being executed in a _different_ process—it's an analogy, not a strictly-correct description):
> 
> ```smalltalk
> terminateCritical
> 	| handler complete |
> 	<primitive: 198>
> 	handler := self.
> 	complete := true.
> 	^self valueNoContextSwitch.
> ```
> 
> And sure enough, if I evaluate:
> 
> ```smalltalk
> [ [ Transcript show: 'before'.
> Processor activeProcess terminate.
> Transcript show: 'after' ] terminateCritical ] fork
> ```
> 
> I see `beforeafter`. Really all I've done is strip out all the bits of `#ensure:` that involve the block that is normally the receiver, and make the parts involving the handler block refer to the receiver...

this is nice; you made `self` a proper unwind block, hence terminating it just resumes execution, as requested :) I have to think about the _critical_ analogy though, don't see it now.

> 
> I'm not sure what exactly you mean by "treating non local returns inside unwind blocks as normal workflow"? 

I tried to say that no longer there are artificial limitations inside the unwind blocks (i.e. #ensure: and ifCurtailed: argument blocks) like "don't use non-local returns insode unwind blocks" or "don't nest the unwind blocks". Similarly, exceptions are often told not to be used as a part of ordinary execution flow but, especially in Smalltalk, I can't see any reason why not use exceptions as a natural workflow tool (*), not just a tool to deal with errors and failures. My plan is to extend the unwind semantics for exceptions and non-local returns too :)

(*) recently I've discovered Barbara Liskov considered exceptions as part of ordinary execution flow - see https://en.wikipedia.org/wiki/CLU_(programming_language).

(Well, a Haskeller would probably give me plenty reasons why that is not a good idea but let's not go there ;) )

>As far as I can see, the new Cuis behavior doesn't affect non-local returns at all, only calls to `#terminate` 

Yes, sorry for confusion, I only meant non-local returns during termination; as I mentioned above, I'd like to extend the same semantics to all non-local returns.

> (whether explicit in-proc, explicit from another proc, or implicit by closing a debugger—AFAICT "unhandled exception" is a red herring, it's just another way of causing a `#terminate` call). So for instance Dolphin executes this example:
> 
> ```smalltalk
> [
> 	[ ] ensure: [
> 		[self error: #unwind_test] ensure: [
> 			^Transcript show: #x1]. 
> 		Transcript show: #x2]
> ] ensure: [
> 	Transcript show: #x3].
> Transcript show: #x4
> ```
> 
> the same as Pharo/Cuis, but that's because Pharo/Cuis don't do anything special here, none of the new behavior is invoked. 

Yes

> Pretty much all the rest of the examples, you're right that Dolphin's semantics are more limited, it will abandon an `ensure:`'d block halfway through on another `#terminate`.
> 
> Actually, what's more interesting/concerning is that Dolphin will actually _fail to terminate_ a process due to a nonlocal return in an `ensure:` block. 

I was so surprised to find that even Instantiations fail these examples spectacularly :) (not that I have much experience with VAST)

> This isn't something you can demonstrate in a workspace, because you need multiple method contexts on the stack to have a valid home for the block to return to, that will then execute more code. I'll have to whip up an example with a dummy class and see how this behaves in Pharo...
> 
> > ... which is precisely what I didn't like; engaging exceptions felt like too much complexity to terminate a process. I chose a straightforward loop to walk the sender chain and execute all unwind blocks. Another highlight is that the process that does the unwinding is the terminating process itself (so we have process faithful termination, can't remember whether that applies to Dolphin too)
> 
> So I realize that I was under a misconception about Pharo's implementation, which led me to say some misleading things. The part I thought was different about Dolphin is precisely that the process doing the unwinding is the terminating process itself—I somehow thought `#terminate` used a simulation mechanism like the debugger. I had reason to look at it more closely and I see that it's doing exactly what I thought it should do, and what Dolphin does—building a new frame/context and manually placing it on top of the process-to-terminate's stack,

to be precise it's not placing the new context/frame _on top_ of the process's stack but it rather creates a _second_ brand new stack; this approach seemed easier than placing it on top of the stack-to-be-unwound.

> then allowing the code therein to do the work. Dolphin has this as a VM interrupt facility rather than constructing the context by hand, but you actually _could_ do it by hand and it would still work (well, only probably, for reasons not worth explaining). I only brought up `#signalException:` in the first place because I saw that that does "interrupt" the process (by the same manually-add-a-new-top-context approach), and thought that was a point of contrast rather than similarity.

btw there a new substantially simplified #signalException: in Squeak

> 
> I...mostly agree that the use of an exception is an ugly way to terminate a process. 

It actually is not a strange idea; an exception return does exactly what termination requires - an unwind. But I found it was hard to take care of all the various border situations. So instead I wrote #unwindTo; and used it for both termination and exceptions (exceptions in Squeak only yet).

> I think I've made use of it a couple times in development code, but only when doing things that I would very much never do in production code. For instance, I wrote a facility that allows for:
> 
> ```smalltalk
> [ guard1 ifTrue: [^false].
> guard2 ifFalse: [^false].
> obj someActualComputation] withInterceptedReturns
> ```
> 
> which wraps the block such that, when it is sent `value`, the `^`-returns will return to the sender of `value`, not the home context of the block. Very handy in a workspace, not something I'd touch with a 10-foot pole in production. And IIRC, my _original_ implementation looked for `ProcessTermination` so the block wouldn't prevent the process from being terminated, but this caused other problems and I eventually got smarter with my stack introspection and was able to do away with it—so perhaps that's an argument against the use of the exception, as well.



<blockquote><div><strong><a href="https://en.wikipedia.org/wiki/CLU_(programming_language)">CLU (programming language) - Wikipedia</a></strong></div></blockquote>

Ducasse:
So I read everything :) and I was thinking that this kind of discussions would be cool to have at ESUG. 
Now about the names, digest and propose something with method with nice comments and we can update the documentation. Now it means that we will have to deprecate current users too. And that we will have this names for the next 10 years at least.

active/ready/waiting/suspended/terminated  are ok for me.

daniels220:
> haha this is typically an infinite discussion ^^.

Fair, but I thought I'd give a little more breather for people to get a word in if they wanted—I know I'd appreciate it if our positions were reversed. Seems like everyone who was already involved has replied though, so I'm good to move forward.

Happily, it seems like we're all content with active/ready/waiting/suspended/terminated, which are my original and preferred names as well, so I've done the rewrite of the base-image senders of `isActiveProcess` so it can truly be considered deprecated.

I also wrote a couple tests of the `ready` and `waiting` states. It's hard to express "these are mutually exclusive" in the tests without an obnoxious amount of repetition, but I think that much I'm comfortable leaving as extremely obvious from the code, in which case it really doesn't take much to cover the new cases—in some cases they're already _there_, it's just a matter of adding assertions in the right places.

In terms of documentation, are the method comments sufficient to work from to modify any other documentation? Is there anything I should help with as far as updating other docs goes?

PR #16073, opened 2024-01-29
Release all waiting processes when destroying a socket (fixes #15975)

### LINKED ISSUE #15975: Destroying a socket does not release processes waiting on that socket
## Bug description
If I call one of the `#waitForData...` family and the other side closes the connection, the `readSemaphore` is signaled reasonably quickly and the method indicates the connection is closed. Similarly for `waitForSendDone`. However if another Smalltalk process calls `close`, and even `destroy`, on _this_ end of the connection, none of the Semaphores are ever signaled and any waiting process will be stuck until the garbage collector reclaims them or their timeouts expire.

## To Reproduce
Note, in all these examples, the other side of the connection is "unresponsive"—otherwise the operation would finish long before the timeout expires. This is just for convenience, there are plenty of cases where this might occur in normal operation—the listening socket of a webserver, or an open WebSocket or similar long-lived connection waiting for traffic that might not occur for many minutes.

### When listening
```smalltalk
Transcript clear.
sock := Socket newTCP
	        listenOn: 1234 backlogSize: 5;
	        yourself.
[
| waitTime |
Transcript crShow: 'Waiting for connection'.
waitTime := Time millisecondsToRun: [
	            sock
		            waitForConnectionFor: 5
		            ifTimedOut: [ Transcript crShow: 'Wait timed out' ] ].
Transcript crShow: 'Wait took ' , waitTime printString , 'ms' ] fork.
Transcript crShow: 'Forked proc started'.
2 seconds asDelay wait.
Transcript crShow: 'Closing socket'.
sock closeAndDestroy.
```
Times out after 5s rather than breaking loose after 2s. Note that the socket-closed case is actually entirely unhandled here and would be treated as a timeout even if the semaphore were signaled.

### When receiving
```smalltalk
Transcript clear.
listen := Socket newTCP
	          listenOn: 1234 backlogSize: 5;
	          yourself.
[
client := Socket newTCP
	          connectTo: (SocketAddress fromDottedString: '127.0.0.1')
	          port: 1234 ] fork.
server := listen waitForAcceptFor: 1.
[
| waitTime |
Transcript crShow: 'Waiting for data'.
waitTime := Time millisecondsToRun: [
	            client
		            waitForDataFor: 5
		            ifClosed: [ Transcript crShow: 'Connection closed' ]
		            ifTimedOut: [ Transcript crShow: 'Wait timed out' ] ].
Transcript crShow: 'Wait took ' , waitTime printString , 'ms' ] fork.
Transcript crShow: 'Forked proc started'.
2 seconds asDelay wait.
Transcript crShow: 'Closing socket'.
client closeAndDestroy.
0.5s seconds asDelay wait.
server closeAndDestroy.
listen closeAndDestroy.
```
In this case the `ifClosed:` branch is taken, but only after the 5-second timeout expires.

### When sending
Note, there isn't a terribly convenient way to send a lot of data with a custom timeout—I'd have to rewrite the outer loop of `sendData:`, so it's easier to just impose a shorter timeout externally—the point is that the process should be unblocked and report the closure much sooner.
```smalltalk
Transcript clear.
listen := Socket newTCP
	          listenOn: 1234 backlogSize: 5;
	          yourself.
[
client := Socket newTCP
	          connectTo: (SocketAddress fromDottedString: '127.0.0.1')
	          port: 1234 ] fork.
server := listen waitForAcceptFor: 1.
[
Transcript crShow: 'Beginning send'.
[
client sendData: (ByteArray new: 2 ** 24).
Transcript crShow: 'Send data completed' ]
	valueWithin: 5 seconds
	onTimeout: [ Transcript crShow: 'Wait timed out' ] ] fork.
Transcript crShow: 'Forked proc started'.
2 seconds asDelay wait.
Transcript crShow: 'Closing socket'.
client closeAndDestroy.
0.5s seconds asDelay wait.
server closeAndDestroy.
listen closeAndDestroy
```

## Expected behavior
When the socket is destroyed, all waiting processes should be unblocked and properly report that the socket has been closed out from under them.

Worth noting that many methods actually break in this circumstance, as they directly refer to the `socketHandle` without a nil-check, and/or call primitives that fail if the socket is in `InvalidSocket` state without checking that first.

## Version information
Any OS, any Pharo

## Expected development cost
It would be a start to send `signalAll` to each Semaphore before niling in it in `destroy`. There is possibly a race condition here as someone could have a reference to the Semaphore on the stack and be about to call `wait` when we call `signalAll`, though this seems very unlikely. Fixing the methods that break if the socket is closed out from under them is a bit more work, but not too difficult—basically any method that refers directly to the `socketHandle`, and also sends any message with `wait` in the name, is suspect. There's probably a way to bring up an actual browser on this, but an inspector will do:
```smalltalk
Socket methodDictionary select: [ :each |
	(each readsSlot: (Socket slotNamed: 'socketHandle')) and: [
		each ast sentMessages anySatisfy: [ :msg | msg beginsWith: #wait ] ] ]
```
In many cases the `wait...` method itself can perform the necessary check and raise an exception if the socket is closed, in which case the caller can remain unchanged. It's probably more in keeping with the rest of the API to add an `ifClosed:` branch to the `wait` methods that already have an `ifTimedOut:`, then implement a version throwing an exception on top of that. The API here is a bit messy and it would be nice to have a "style guide" of sorts for it (that's the wrong phrase—a set of rules/conventions the API follows, basically). I could probably help some with hashing out what questions need to be answered and documenting the way things are and what is decided, but this seems like something that needs input from a number of people to _make_ those decisions.

daniels220:
Send #signalAll to all three semaphores as part of #destroy so any waiting processes are immediately unblocked. Nil the semaphores first to minimize the possibility of a race condition.

Then, improve reporting of this condition by making sure everyone has an #ifClosed: branch--specifically  #waitFor{Connection|Accept}For:, analogous to #waitForData/#waitForSendDone families. Leave #waitForDisconnectionFor: alone as returning a boolean, since the socket being destroyed indeed means it is disconnected, so the only failure condition is a timeout. Plus some cleanup of the loops to reduce duplicated code.

Last, cleanly report attempts to retrieve the last socket error code when the socket is already destroyed

### DIFF
--- a/src/Network-Kernel/Socket.class.st
+++ b/src/Network-Kernel/Socket.class.st
@@ -465,11 +465,18 @@ Socket >> connectTo: hostAddress port: port [
 Socket >> connectTo: hostAddress port: port waitForConnectionFor: timeout [
 	"Initiate a connection to the given port at the given host
 	address. Waits until the connection is established or time outs."
+
 	self connectNonBlockingTo: hostAddress port: port.
 	self
 		waitForConnectionFor: timeout
-		ifTimedOut: [ConnectionTimedOut signal: 'Cannot connect to '
-					, (NetNameResolver stringFromAddress: hostAddress) , ':' , port asString]
+		ifClosed: [
+			ConnectionClosed signal: 'Connection aborted to '
+				, (NetNameResolver stringFromAddress: hostAddress) , ':'
+				, port asString ]
+		ifTimedOut: [
+			ConnectionTimedOut signal: 'Cannot connect to '
+				, (NetNameResolver stringFromAddress: hostAddress) , ':'
+				, port asString ]
 ]
 
 { #category : 'connection open/close' }
@@ -489,18 +496,27 @@ Socket >> dataAvailable [
 
 { #category : 'initialize - destroy' }
 Socket >> destroy [
-	"Destroy this socket. Its connection, if any, is aborted and its resources are freed. Do nothing if the socket has already been destroyed (i.e., if its socketHandle is nil)."
+	"Destroy this socket. Its connection, if any, is aborted and its resources are freed.
+	Any processes waiting on the socket are freed immediately, but it is up to them to
+	recognize that the socket has been destroyed.
+	Do nothing if the socket has already been destroyed (i.e., if its socketHandle is nil)."
 
-	socketHandle
-		ifNotNil: [
-			self isValid
-				ifTrue: [ self primSocketDestroy: socketHandle ].
-			Smalltalk unregisterExternalObject: semaphore.
-			Smalltalk unregisterExternalObject: readSemaphore.
-			Smalltalk unregisterExternalObject: writeSemaphore.
-			socketHandle := nil.
-			readSemaphore := writeSemaphore := semaphore := nil.
-			self unregister ]
+	socketHandle ifNotNil: [
+		| saveSemaphores |
+		self isValid ifTrue: [ self primSocketDestroy: socketHandle ].
+		socketHandle := nil.
+		Smalltalk unregisterExternalObject: semaphore.
+		Smalltalk unregisterExternalObject: readSemaphore.
+		Smalltalk unregisterExternalObject: writeSemaphore.
+		"Stash the semaphores and nil them before signaling to make sure
+		no caller gets a chance to wait on them again and block forever."
+		saveSemaphores := {
+			                  semaphore.
+			                  readSemaphore.
+			                  writeSemaphore }.
+		semaphore := readSemaphore := writeSemaphore := nil.
+		saveSemaphores do: [ :each | each signalAll ].
+		self unregister ]
 ]
 
 { #category : 'receiving' }
@@ -1334,6 +1350,7 @@ Socket >> retryIfWaitingForConnection: aBlock [
 				  ifTrue: [
 					  self
 						  waitForConnectionFor: Socket standardTimeout
+						  ifClosed: nil
 						  ifTimedOut: nil.
 					  aBlock value ]
 				  ifFalse: [ e pass ] ]
@@ -1569,40 +1586,57 @@ Socket >> unregister [
 { #category : 'waiting' }
 Socket >> waitForAcceptFor: timeout [
 	"Wait and accept an incoming connection. Return nil if it fails"
-	^ self waitForAcceptFor: timeout ifTimedOut: [nil]
+
+	^ self waitForAcceptFor: timeout ifClosed: nil ifTimedOut: nil
 ]
 
 { #category : 'waiting' }
-Socket >> waitForAcceptFor: timeout ifTimedOut: timeoutBlock [
+Socket >> waitForAcceptFor: timeout ifClosed: closedBlock ifTimedOut: timeoutBlock [
 	"Wait and accept an incoming connection"
-	self waitForConnectionFor: timeout ifTimedOut: [^timeoutBlock value].
-	^self accept
+
+	self
+		waitForConnectionFor: timeout
+		ifClosed: [ ^ closedBlock value ]
+		ifTimedOut: [ ^ timeoutBlock value ].
+	^ self accept
 ]
 
 { #category : 'waiting' }
 Socket >> waitForConnectionFor: timeout [
 	"Wait up until the given deadline for a connection to be established. Return true if it is established by the deadline, false if not."
 
-	^self
-		waitForConnectionFor: timeout
-		ifTimedOut: [ConnectionTimedOut signal: 'Failed to connect in ', timeout asString, ' seconds']
+	^ self
+		  waitForConnectionFor: timeout
+		  ifClosed: [
+			  ConnectionClosed signal: (socketHandle
+					   ifNil: [ 'Socket destroyed while connecting' ]
+					   ifNotNil: [
+					   'Connection aborted or failed: ' , self socketErrorMessage ]) ]
+		  ifTimedOut: [
+			  ConnectionTimedOut signal:
+				  'Failed to connect in ' , timeout asString , ' seconds' ]
 ]
 
 { #category : 'waiting' }
-Socket >> waitForConnectionFor: timeout ifTimedOut: timeoutBlock [
-	"Wait up until the given deadline for a connection to be established. Return true if it is established by the deadline, false if not."
+Socket >> waitForConnectionFor: timeout ifClosed: closedBlock ifTimedOut: timeoutBlock [
+	"Wait up until the given deadline for a connection to be established.
+	Evaluate closedBlock if the connection is closed locally,
+	or timeoutBlock if the deadline expires.
+	
+	We should separately detect the case of a connection being refused here as well."
 
-	| startTime msecsDelta msecsEllapsed status |
+	| startTime msecsDelta msecsElapsed status |
 	startTime := Time millisecondClockValue.
 	msecsDelta := (timeout * 1000) truncated.
-	status := self primSocketConnectionStatus: socketHandle.
-	[(status = WaitingForConnection) and: [(msecsEllapsed := Time millisecondsSince: startTime) < msecsDelta]]
-		whileTrue: [
-			semaphore waitTimeoutMilliseconds: msecsDelta - msecsEllapsed.
-			status := self primSocketConnectionStatus: socketHandle].
 
-	status = Connected ifFalse: [^timeoutBlock value].
-	^ true
+	[
+	status := self primSocketConnectionStatus: socketHandle.
+	status == WaitingForConnection and: [
+		(msecsElapsed := Time millisecondsSince: startTime) < msecsDelta ] ]
+		whileTrue: [ semaphore waitTimeoutMilliseconds: msecsDelta - msecsElapsed ].
+
+	status == WaitingForConnection ifTrue: [ ^ timeoutBlock value ].
+	status == Connected ifFalse: [ ^ closedBlock value ]
 ]
 
 { #category : 'waiting' }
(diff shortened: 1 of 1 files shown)


guillep (approved):
I like the changes.
Maybe @tesonep wants to give a quick look?

Otherwise it seems good to me.

@daniels220 could you give us for the record some details about how "hard" have you tested this? Or some scenario you foresee that we need to test a bit more?

guillep:
there is a seemingly related issue in the ci, ideas ?
```
Instance of Semaphore did not understand #signalAll

Semaphore(Object)>>doesNotUnderstand: #signalAll

[ :each | each signalAll ] in Socket>>destroy in Block: [ :each | each signalAll ]

Socket>>destroy

Socket>>closeAndDestroy:

Socket>>closeAndDestroy

ZdcSecureSocketStream(ZdcAbstractSocketStream)>>socketClose

[

		socket ifNotNil: [

			self socketClose.

			socket := nil ].

		readBuffer close.

		writeBuffer close ] in ZdcSecureSocketStream(ZdcAbstractSocketStream)>>close in Block: [...

ZdcSecureSocketStream(ZdcAbstractSocketStream)>>close

[ super close ] in ZdcSecureSocketStream>>close in Block: [ super close ]

ZdcSecureSocketStream>>close

ZnClient>>close

ZnClient>>validateConnectionForHost:port:

ZnClient>>url:

ZnClient>>prepareRedirect

ZnClient>>executeWithRedirectsRemaining:

[ self executeWithRedirectsRemaining: self maxNumberOfRedirects ] in ZnClient>>executeWithRetriesRemaining: in Block: [ self executeWithRedirectsRemaining: self max[..]

ZnClient>>executeWithRetriesRemaining:

[ self executeWithRetriesRemaining: self numberOfRetries ] in [

		[ self executeWithRetriesRemaining: self numberOfRetries ]

			on: Error

			do: self ifFailBlock ] in ZnClient>>executeWithTimeout in Block: [ self executeWithRetriesRemaining: self numbe[..]

[

		[ self executeWithRetriesRemaining: self numberOfRetries ]

			on: Error

			do: self ifFailBlock ] in ZnClient>>executeWithTimeout in Block: [...

[ ^ block value ] in ZnClient>>withTimeoutDo: in Block: [ ^ block value ]

[ activeProcess

			psValueAt: index

			put: anObject.

		aBlock value ] in ZnConnectionTimeout(DynamicVariable)>>value:during: in Block: [ activeProcess...

ZnConnectionTimeout(DynamicVariable)>>value:during:

ZnConnectionTimeout class(DynamicVariable class)>>value:during:

ZnClient>>withTimeoutDo:
```

daniels220:
> could you give us for the record some details about how "hard" have you tested this? Or some scenario you foresee that we need to test a bit more?

To be honest, not as hard as I/we probably should, but at the same time I can't put a finger on why. There are just so many different ways to use sockets, and this was driven by finding that a pattern in the code I work with just doesn't work, but I'm always wary when touching code this central to something.

(A little background: I work on a large legacy application being ported to Pharo, and it uses a long-lived socket connection, which has processes waiting on it for data to arrive. In other dialects we found that the most reliable way to close the connection and release those processes was to close the socket "out from under them", rather than explicitly terminate them—trying that would sometimes confuse the FFI layer and cause an image crash, where closing the socket would raise an "error" on the C library side which would cleanly propagate and unwind the FFI call. Jump on over to Pharo, and I find that tests are failing because the connection just never tears down properly as far as the test is concerned (it's checking both the socket and the processes, since it's important that they _do_ wind down so they can be started back _up_ again with a new socket later).)

I find it a bit strange TBH that closing the socket doesn't cause the semaphores to be signaled by the SocketPlugin code—surely closing a socket causes an active `select()` to trigger (or whatever the proper terminology for that is)? But I don't know C well enough to look at that, so I went for the solution I could implement. But that does make me a little uncomfortable, which I think is part of why I feel like yeah, more testing could be good.

Re: the CI failure—I notice that `signalAll` is an extension from `ThreadedFFI`, not part of the kernel. I'd have assumed the tests are all run with the image fully built/all packages loaded, but if not, that would do it—and it may be bad form to introduce a dependency from `Network-Kernel` -> `ThreadedFFI` if it doesn't already exist? Now that, again, you point it out more explicitly, the use of `signalAll` is another point of discomfort actually. In any _safe_ usage AFAICT, there should be at most one process waiting on each semaphore, because otherwise how are they supposed to avoid stepping on each others' toes? So it should be sufficient to merely `signal` each semaphore once. Perhaps on balance that would be better—it opens a theoretical hole, but one that is only likely to matter in already-buggy code. And _eventually_ all processes will be released when the Semaphore is garbage-collected—assuming anyway that nobody else refers to those processes themselves, which would in turn _prevent_ the Semaphore from being GC'd, and that is a possibility.

Lemme sleep on this a bit and see what I think about those two points, and/or if I think of ways to better test it. I notice `SocketTest` is very anemic, most of the real functional testing is happening on `SocketStreamTest`, but this is (unsurprisingly) very coupled to the way `SocketStream` uses `Socket` so it can't test the full API. I can't say I feel like writing out a comprehensive `SocketTest` from scratch, but maybe I can lay out a sketch of what could be done and fill in a couple bits?

guillep:
Haha, actually my question was not to block the issue here but to document it. I know that it's not easy to test/measure the impact of this fix. But I'm glad you have it in mind.

Just FTR, if you think of some tests in the future, I'd like to see them :)

guillep:
Argh, I missed that this was breaking the CI because of a missing `#signalAll` method. I'll revert it, we are going to re-issue the PR, sorry...

guillep:
I see, the `signalAll` method is a threadedFFI extension.

imagen

We  should move the method to the semaphore class directly.

daniels220:
Yes, I saw that `signalAll` is in `ThreadedFFI`, but I had so much else to say it looks like I didn't emphasize that enough in my reply :). If moving it to Semaphore directly makes sense to you, I can do that as part of re-submitting. However it's worth noting that there currently _are_ no other senders of `signalAll` outside of `ThreadedFFI`, so it may be premature to move it. Does my thinking re: not actually _needing_ the `signalAll` make sense to you—that it could be regular `#signal` instead?

guillep:
> Yes, I saw that `signalAll` is in `ThreadedFFI`, but I had so much else to say it looks like I didn't emphasize that enough in my reply :). If moving it to Semaphore directly makes sense to you, I can do that as part of re-submitting.

Yes, it makes sense to me.
My point here is that such a global extension produces silly issues as this one. The extension is globally available, but when the system is building (and loading packages one by one), it may not be available.

> However it's worth noting that there currently _are_ no other senders of `signalAll` outside of `ThreadedFFI`, so it may be premature to move it. 

To me here the extension is adding a missing Semaphore functionality...
- Either it's a ThreadedFFI functionality then it should be moved close to the user to avoid mistakes like this one.
- Or its available for all Semaphore users then we move it to Semaphore class, with no extension.

> Does my thinking re: not actually _needing_ the `signalAll` make sense to you—

This is a third solution (which would be similar to pushing the `signalAll` method as a ThreadedFFI method).
We could re-implement as follows:

```smalltalk
Socket >> signalAll: semaphore
  ...
```

> that it could be regular `#signal` instead?

And yes, if actually we are using semaphores as mutexes, maybe a single `signal` suffices? I agree here :)

Do you want that I revert the revert to continue working on this?

guillep:
Hi @daniels220 any news on this front? How do you want that we proceed?

daniels220:
I've just been busy the past week, thanks for the reminder.

Hmm...I had some concerns about introducing `signalAll` into the core Semaphore API, but thinking about them some more I think it's really usage-dependent. As long as whatever mechanism calls `signalAll` is written with the understanding that more processes might start waiting again immediately, and this is the intended behavior, it's fine.

In this particular case, the "ideal" method would be a `destroy` or something, that not only releases all waiting processes but "breaks" the Semaphore so further calls to `wait` have no effect—or you could think of it as setting `excessSignals` to infinity. But I also think that in practice, not only is this not a concern—the chances of it are laughably miniscule, if it's even possible—but a single `signal` should be sufficient unless the calling code is already buggy. So I went with that approach, but independently I have no problem with `signalAll` being moved to Semaphore.

So, I've resubmitted #16172, which includes the reapply as the first commit before tweaking things. I hope that's convenient for you?

guillep:
> I've just been busy the past week, thanks for the reminder.

Oh, sorry, no intention to bother you, just I needed info to move on.
Thanks for replying fast!

> Hmm...I had some concerns about introducing `signalAll` into the core Semaphore API

I undertand

> , but thinking about them some more I think it's really usage-dependent. As long as whatever mechanism calls `signalAll` is written with the understanding that more processes might start waiting again immediately, and this is the intended behavior, it's fine.

I agree with both your feeling and your final analysis ^^
 
> In this particular case, the "ideal" method would be a `destroy` or something, that not only releases all waiting processes but "breaks" the Semaphore so further calls to `wait` have no effect—or you could think of it as setting `excessSignals` to infinity. 

I like this idea! I'd make it part of another issue though.
One possibility is to break it by setting `excessSignals := nil`, and maybe rework a bit the Semaphore code (both Pharo+vm) to be robust about this case (don't remember if they are from the top of my head).

> But I also think that in practice, not only is this not a concern—the chances of it are laughably miniscule, if it's even possible—but a single `signal` should be sufficient unless the calling code is already buggy. So I went with that approach, but independently I have no problem with `signalAll` being moved to Semaphore.

Yes, this is good as it allows us to move on!
Nothing prevents that we introduce later a proper `signalAll` if we see the use case, it's rather a easy change :)
 
> So, I've resubmitted #16172, which includes the reapply as the first commit before tweaking things. I hope that's convenient for you?

It's great! thanks! I'll pass over it between today and tomorrow!

astares:
@daniels220 

>I work on a large legacy application being ported to Pharo

Can you tell us more (if possible)? Maybe a new story you could describe for https://pharo.org/success ?

<blockquote><div><strong><a href="https://pharo.org/success">Pharo - Pharo Success Stories</a></strong></div></blockquote>

PR #16074, opened 2024-01-30
Expand the dependency analyzer roots by default

jecisc:
When opening the dependency analyzer we get dsiplayed a tree with the packages to analyze (often we only have one) and we can expand them to see their dependencies.

I think that it would be nice to automatically expand the roots so that we see the dependencies directly and that we do not have to do it all the time

Before

image

After

image


astares:
A roassal visualization would be nice and I think it is part of the roassal uml package.

PR #16084, opened 2024-02-01
Remove the old morphic-based FileList

hernanmd:
This is just a first PR to remove the FileList based on Morphic. The tool has already a replacement version based in Spec 2 recently integrated and ready to be used.
The new file system lister or browser is implemented in `StFileSystemPresenter`.


guillep (changes requested):
Hi @hernanmd Thanks!

This PR is just removing the tool from the listed packages, but it's not removing the package. To do that in Iceberg you need to use the "Remove from repository" option and commit that.

hernanmd:
> Hi @hernanmd Thanks!
> 
> This PR is just removing the tool from the listed packages, but it's not removing the package. To do that in Iceberg you need to use the "Remove from repository" option and commit that.

Thanks, I will fix this today as there are a couple of users of UITheme that needs translation.

guillep:
> > Hi @hernanmd Thanks!
> > This PR is just removing the tool from the listed packages, but it's not removing the package. To do that in Iceberg you need to use the "Remove from repository" option and commit that.
> 
> Thanks, I will fix this today as there are a couple of users of UITheme that needs translation.

Oh I see this is a work in progress. I'll mark the PR as draft to avoid accidental merges :)

hernanmd:
Updated. This needs https://github.com/pharo-spec/NewTools/pull/682 prior to be applied.

jecisc:
I resolved the conflict. The PR in NewTools got merged so maybe this PR should not be a draft anymore?

src/Polymorph-Widgets/UITheme.class.st
```diff
@@ -754,12 +754,12 @@ UITheme >> chooseColorIn: aThemedMorph title: aString color: aColor for: aBlock
...
 
-	| dialog |
-	dialog := (FileDialogWindow newWithTheme: aThemedMorph theme)
-		title: title;
-		answerDirectory.
-	path ifNotNil: [ dialog selectPath: path ].
-	^ dialog openModal answer
+	^ StOpenDirectoryDialog new
+		defaultFolder: path;
+		title: (title ifNil: [ 'Choose Directory' translated ]);
```
demarey, line 759: This does not work.
Title if provided is not displayed as the window is not yet open

hernanmd: It looks buggy certainly. I don't know because I never use `translate`  :)
Anyway this could be soon deprecated, @jecisc ?

jecisc: translated should return a string translated in the local language if the translation project is loaded IIRC. In the worst case it should return the original string in any case.

jecisc: I think that some of the API should get deprecated but I'm not yet sure what will get deprecated exactly and right now I'm working on Moose

demarey: The problem here is not the translation. Setting the title before the window is open has no effect.
You can try by yourself:
```smalltalk
StOpenDirectoryPresenter new
		title: 'test set tittle';
		openModal
```

Ducasse:
As discussed for P13

src/Polymorph-Widgets/UITheme.class.st
hernanmd: I've sent a PR in https://github.com/pharo-spec/NewTools/pull/716 to fix this

jecisc: Merged

jecisc:
This has been done in other PRs

PR #16096, opened 2024-02-05
Improving Message list morph UI

jordanmontt:
This PR changes a little the UI of the old Calypso Message Browser.

- It removes the protocol column. This is because when one browses the senders, the protocol information is most of the time not needed. And also, in the new StMessageBrowser there is not a protocol column.
- It changes the default extent to be wider.
- It now uses dynamic width instead of a fixed one.

This is the old UI:
Capture d’écran 2024-02-05 à 16 44 07

And this is the new one:
Capture d’écran 2024-02-05 à 16 43 21

### DIFF
--- a/src/Calypso-SystemTools-QueryBrowser/ClyQueryBrowserMorph.class.st
+++ b/src/Calypso-SystemTools-QueryBrowser/ClyQueryBrowserMorph.class.st
@@ -182,7 +182,7 @@ ClyQueryBrowserMorph >> extraScopesOfSelectedItems [
 { #category : 'initialization' }
 ClyQueryBrowserMorph >> initialExtent [
 
-	^ 1450 @ 620 * self currentWorld displayScaleFactor
+	^ 1350 @ 620 * self currentWorld displayScaleFactor
 ]
 
 { #category : 'initialization' }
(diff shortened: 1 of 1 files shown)


Ducasse:
Thanks now do not invest too much we will remove it. 
It is better to improve the StMessageBrowser. Now I do not know if the one in the image is the latest one. 
@estebanlm

jecisc:
I personally really like the protocol column :(
It helps me a lot for some scenario:
- I want to categorize methods and browse implementors to know where they usually are
- It help me sometime understand the goal of the method 
- It’s really useful to see quickly the extensions and know where methods are packaged when they are

Ducasse:
I can understand cyril. I think that we should invest in StMessageBrowser and we could have an option to show or not the protocol.

astares:
I would also keep the protocol column - it can easily show you if messages are aligned regarding protocol. Also as the filter is applied to this column too you can more easily filter for the interesting methods by keying in the protocol.

Also in most UI (frameworks) the ListView / TableView allows to control which columns are hidden or not. Often this is done via context menu on the column headers. So the user can easily select which column to show or hide or (by dragging) even reorder the columns as required. For sure there is some effort required to implement such a beast but ANY generic table like/list like view in Pharo would profit from that if it could be added to Spec2 somehow in a generic way. 

image

image

estebanlm:
there is a maximum amount of information you can put there without losing functionality because the size of the browser becomes too big (and I also agrees this could be also configurable).
I would accept this PR as it is an improvement and already done, but *knowing* it corresponds to a component that will be replaced soon(tm)... in February for sure :)
ps. no, the one in the image is not the good one.

jordanmontt:
I agree that one should be able to configure which columns one wants to see and also the order. But as said, that will be for the new message browser. I will add back the protocol column as suggested. But, for the new message browser we may need to think in a new layout, because, as Esteban said, the size of the browser becomes too big

hernanmd (approved):
Looks legit for me, it only changes the order of the columns and add dynamic sizing

PR #16106, opened 2024-02-07
Revert "Revert "Remove UIManager dependencies from Compression.""

guillep:
Rework of https://github.com/pharo-project/pharo/pull/16068 by @hernanmd .

The original PR broke the build. But this is going to get fixed :)


hernanmd:
Can you comment what's the core change?
Because I tried to add in BaselineOfBasicTools: 

```smalltalk
		spec package: 'System-Spec' with: [ spec requires: #('NewTools-Core') ].
```

But that raised another error `Error: Name not found: NewTools-Core` while loading the Baseline

hernanmd:
I re-take this PR here: https://github.com/pharo-spec/NewTools/pull/683, they're essentially the same changes, but with reduced dependency to SpApplication, it is simpler to check and does not use the `System-Spec` approach.

So I think this PR can be closed if you agree.

guillep:
> Can you comment what's the core change?

I don't understand the question, could you rephrase?
Do you mean why reverting fixed the build?

> Because I tried to add in BaselineOfBasicTools:
> 
> ```smalltalk
> 		spec package: 'System-Spec' with: [ spec requires: #('NewTools-Core') ].
> ```
> 
> But that raised another error `Error: Name not found: NewTools-Core` while loading the Baseline

Adding the requires clause does not fix the issue, it just adds a new one.
Each baseline only sees what's defined in itself, and in this case BaselineOfBasicTools does not define 'NewTools-Core'

PR #16109, opened 2024-02-07
improving-snapshot-process-P12

tesonep:
- Correctly handling errors during saveAndQuit
    - Creating an object to handle the snapshot process and errors


guillep (approved):
Thanks, same as in P11 :)

PR #16112, opened 2024-02-07
Refactor: make pull up class variable same as pull up instance variable

balsa-sarenac:
Pull up class variable didn't have any breaking changes and it should have.


Ducasse:
oh yes this is much much nicer.

PR #16113, opened 2024-02-08
Cleanup: push down instance variable breaking changes

balsa-sarenac:
Fix push down class variable breaking changes so it is aligned to push down instance variable.
While doing this fix a bunch of bugs and improvements were popping-up, so I fixed them along the way. Hope the PR is not too huge now. Fixes include:

- add breaking changes to push down class variable
- add method that checks for class variable references in a class
- add missing applicability precondition that class variable should exist in order for it to be removed
- thanks to some fixes some skipping tests now pass, so can be unskipped!


Ducasse:
Looks great!

PR #16103, opened 2024-02-07
Revert "Remove UIManager dependencies from Compression."

MarcusDenker:
Reverts pharo-project/pharo#16068 

(this seems to break the build)

PR #16121, opened 2024-02-08
Avoid merged tags for vm download in bootstrap script

hernanmd:
As discussed in #16103, this PR changes the way the Pharo major, minor and patch version numbers are obtained from git describe command.

The update uses a similar way `git describe --tags --first-parent --abbrev=1` which prevents using merge tags. This is useful because in the replaced way, using `git describe --long --tags` could answer unexpected output - this is not starting with `v.[0-9]` and then causing the VM not being downloaded by wget.

### DIFF
--- a/bootstrap/scripts/envversion.sh
+++ b/bootstrap/scripts/envversion.sh
@@ -51,9 +51,9 @@ function set_version_common() {
 	# HACK: Since this may beforehand a PR branch, I do not have all information I need. I assume I will have a tag indicating Pharo version.
 	# Note: do not use `--first-parent` here, since Jenkins is crazy and use the PR as the first parent wheras useful version information in the targeted branch.
     # Set the common env vars extracting version information. Requires that prefix and suffix are set beforehand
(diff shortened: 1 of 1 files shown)


jecisc (approved):

guillep:
No! We should not use abbrev!! :(

jecisc:
Oups sorry, should I revert it?

hernanmd:
> No! We should not use abbrev!! :(

I understood there was no problem using it.
What's the problem? Is there an alternative?

guillep:
Yes, 

> > No! We should not use abbrev!! :(
> 
> I understood there was no problem using it. What's the problem? 

The problem is that we don't know if it will introduce some problem or not...
This may have effects in Iceberg integration, in the file server naming of files, impact the launcher...

I'd prefer not to deal with those bugs if possible now :(

> Is there an alternative?

Yes, not using `abbrev=1` gives the right output, as in the discussion.

hernanmd:
Ok, updated to remove "abbrev=1"

PR #16124, opened 2024-02-08
Avoid merged tags without abbrev parameter for vm download in bootstrap script

hernanmd:
Update to fix `"abbrev=1"` in #16121


guillep:
Thanks @hernanmd !

PR #16125, opened 2024-02-08
Refactor: extract method transformation build transformations

balsa-sarenac:
These are three small changes lightly related to extract method:

- Cleanup extract method transformation's `buildTransformations`
- Clean `isUsed` tests and add new test for assignment


Ducasse:
BTW Marcus told me that he does not know uses: and isUsed.

balsa-sarenac:
I've created this issue #16126 so we remember to get back to them

PR #16130, opened 2024-02-09
[Bug]: Workaround Read-Only Mac /tmp

seandenigris:
On OSX, which symlinks tmp, some versions make the link only-root-writable. Therefore, we check to see if the link target (/private/tmp) exists and if so, reference it directly.

For more details, see https://apple.stackexchange.com/questions/257753/osx-tmp-not-writable-operation-not-permitted


guillep (approved):
Ok to me! Thanks @seandenigris !

Issue #16137, opened 2024-02-11
DAGraphVizGenerator looks like deadcode


Ducasse:
The hierarchy is full of dead code. 
So better remove it.

Ducasse:
fixed

PR #16138, opened 2024-02-11
fix: #16137 remove grapoh viz support from Dependency analysis since …

Ducasse:
…it does not work.


MarcusDenker:
ReleaseTest>>testThatThereAreNoSelectorsRemainingThatAreSentButNotImplemented is failing (there can be more cleaned up).

But as lots of methods related to graphiz are noted as false positives, it would be good to do a pass anyway after.
Thus I merge and then I do the pass next

Issue #16117, opened 2024-02-08
Hierarchical implementors look to have a bug


Ducasse:
I tried to look for hierarchical implementors of printOn: and I do not het the Object one.

Ducasse:
I retried today and I got the same with displayString

hernanmd:
I think what is wrong is the usability. In the sense you have to click the arrow to open the hierarchical implementors.
Is this what you mean?

<img width="868" alt="Screenshot 2024-02-08 at 18 33 21" src="https://github.com/pharo-project/pharo/assets/4825959/48deb898-08f6-426f-8bc1-e581aa348cb7">

Ducasse:
When I click on 
![Capture 2024-02-08 at 20 00 04](https://github.com/pharo-project/pharo/assets/432549/791d6604-c9ff-4bb5-8f79-18bad9ef9fff)
![Capture 2024-02-08 at 20 00 13](https://github.com/pharo-project/pharo/assets/432549/16d595dc-1537-4ca4-bd66-a3a90aa480cc)

Ducasse:
When I select a method with an override I should see the overriden method and right I do not see it.

hernanmd:
Fixed in #16150.

PR #16150, opened 2024-02-13
Fix hierarchical implementors bug

hernanmd:
This PR fixes the reported issue #16117 with the following changes:

- Subclass SycSuperclassImplementorsCommand from SycMessageCommand instead of SycChangeMessageSignatureCommand since this command is only about browsing, not applying refactoring.
- Pull up useful methods to SycMessageCommand.
- Add `searchInTheWholeHierarchy` method to collect hierarchical implementors, in the class and metaclass of selected method.

Video demo of the fix:

https://github.com/pharo-project/pharo/assets/4825959/19dd3f32-e214-47ba-a851-71f9a7d5c933


Ducasse:
Yes!!!!

PR #16154, opened 2024-02-13
Feat: extract method source string api

balsa-sarenac:
Now it's possible to create an instance of extract method refactoring by adding it string of the source code you want to extract, whereas previously we had to use intervals.


Ducasse:
We will review it later

PR #16156, opened 2024-02-14
16051-Correct-FormstoreResourceOn-from-Graphics

### LINKED ISSUE #16051: Correct Form>>#storeResourceOn: from Graphics
Form>>#storeResourceOn: is currently an extension from graphics

We should check if it could be made part of Kernel (or just use storeString)

MarcusDenker:
fixes #16051

deprecate strange unused method

### DIFF
--- a/src/Deprecated12/Form.extension.st
+++ b/src/Deprecated12/Form.extension.st
@@ -1,0 +1,20 @@
+Extension { #name : 'Form' }
+
+{ #category : '*Deprecated12' }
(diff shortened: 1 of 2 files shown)


Ducasse:
I think that being able to save a form in binary form is nice so why do we deprecate it?
I would add a test and make sure that it is working. 
The knowledge of this piece of code is not trivial so I would take care of it.

MarcusDenker:
It looked like dead code, but I am ok with both

PR #16157, opened 2024-02-14
16133 transforming deprecation should not raise a debugger by default but rewrite the code

Ducasse:
16133 transforming deprecation should not raise a debugger by default but rewrite the code #16157


MarcusDenker (approved):

MarcusDenker:
Yes, good.

In the past the setting was just impacting non-rewriting transforms.

Maybe having a dedicated setting could be an idea, but we should try to see if the simple one to have just one setting is better.

PR #16158, opened 2024-02-14
Test: semantics of extract method transformation

balsa-sarenac:
This PR introduces new API to create a transformation, and adds new tests.


Ducasse:
We will do a pass on failing tests if any related

PR #16162, opened 2024-02-14
16161-Clean-up-undeclared-assumes-all-undeclared-are-globals

### LINKED ISSUE #16161: Clean up undeclared assumes all undeclared are globals
**Bug description**
I have a PR (https://ci.inria.fr/pharo-ci-jenkins2/blue/organizations/jenkins/Test%20pending%20pull%20request%20and%20branch%20Pipeline/detail/PR-16030/10/pipeline/25) that is introducing an undeclared (`oldStamp` here).

When `cleanUpUndeclared` runs, it assumes all undeclareds are in Smalltalk (thus are globals or variables).
That is, in this case, wrong.

```
KeyNotFound: key #oldStamp not found in SystemDictionary
SystemDictionary(Dictionary)>>errorKeyNotFound:
[self errorKeyNotFound: key] in SystemDictionary(Dictionary)>>at: in Block: [self errorKeyNotFound: key]
SystemDictionary(Dictionary)>>at:ifAbsent:
SystemDictionary(Dictionary)>>at:
SmalltalkImage>>at:
...
...
MethodDictionary>>valuesDo:
...
EpBehaviorCommentChange class(Behavior)>>methodsDo:
...
OrderedCollection>>do:
SystemDictionary>>allBehaviorsDo:
SystemNavigation>>allBehaviorsDo:
SmalltalkImage>>cleanOutUndeclared
```

# Expected

1) that it does not break like that :)
2) that it properly notifies/logs what is the offending method (I cannot tell from the log)
3) Ideally, that notifies all the offending methods together instead of failing on the first

MarcusDenker:
This is a dumb copy and paste bug, the method used to be on Dictionary...  thus the "self" has to be "Undeclared" now.

I added a test fo that situation (two methods with the same undelared name)

fixes #16161

### DIFF
--- a/src/ReleaseTests/ReleaseTest.class.st
+++ b/src/ReleaseTests/ReleaseTest.class.st
@@ -559,8 +559,10 @@ ReleaseTest >> testThatThereAreNoSelectorsRemainingThatAreSentButNotImplemented 
 
 { #category : 'tests - variables' }
 ReleaseTest >> testUndeclared [
-	| undeclaredVariables validExceptions remaining description |
+	| undeclaredVariables validExceptions remaining description  |
 
+	"we compile a second method with the undeclared #undeclaredStubInstVar1 to trigger the code path of removing twice in #cleanOutUndeclared"
+	self class compile: 'methodForTest ^undeclaredStubInstVar1'.
 	Smalltalk cleanOutUndeclared.
 	undeclaredVariables := Undeclared associations select: [:each |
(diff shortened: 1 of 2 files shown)


guillep:
> This is a dumb copy and paste bug, the method used to be on Dictionary... thus the "self" has to be "Undeclared" now.

Oh! I would never have guessed that xD

> I added a test fo that situation (two methods with the same undelared name)

Great!!

Issue #16127, opened 2024-02-09
Command-R should be for replace in the code pane and not hierarchical implementors


Ducasse:
Command-R should be for replace in the code pane and not hierarchical implementors

jordanmontt:
Fixed in https://github.com/pharo-project/pharo/pull/16168

PR #16168, opened 2024-02-14
Replace $r key combination in method editor pane

hernanmd:
As reported in #16127, this PR unbinds the "Meta + R" key combination in the method browser to the "browse superclass implementor" action. The action was re-bind as "Meta + Y".

Note that this PR does not create a new action or command to rename whatever is under (or left to) the cursor. IMO that would be better to do it in another PR because I checked and the "Meta + R" action depends of the node kind under the cursor and right now is only working for renaming temporaries.


Ducasse:
Thanks Hernan!
This was a real usability issue for people used to this command.

Issue #16123, opened 2024-02-08
asFileName in Tool-FileList used by Fuel


hernanmd:
The method `String>>asFileName` is packaged in Tool-List but used by NewTools-Fuel and CodeExporter.

As FileList is a Morph-based package and it will be removed, we should move this method to another package where it guarantees to survive.

Ducasse:
I would rewrite all the senders of the system to avoid to use it and I would deprecate it but keeping it around. Strings are too much overloaded

PR #16169, opened 2024-02-14
Deprecate asFileName and re-write system senders

hernanmd:
As reported in #16123, this PR should be applied along with the [PR in NewTools-Fuel](694). It is a small PR to clean dependencies which could bring problems when removing old FileList.

Note that FileList senders were not re-written as it is planned to be deprecated too.


guillep:
If you ask me (which you did not :P) I'd simply remove the calls to asFileName.
It's implementation is super weird: it (sometimes!) rewrites the paths! This could hide application bugs and make me developer think I'm opening a file here but actually opening a file there.

Also, for the file too long or invalid paths, we are already validating that (it's the OS when we open the file haha !)

My take
 - I'd not inline the implementation anywhere.
 - I'd just deprecate the method, not use it anymore, remove it next version
 - people should provide correct file paths, and we should provide nice errors to help them find bugs

Thoughts?

hernanmd:
Yes, this PR already moves the method to Deprecated12.

It seems this method comes from the ages of Squeak, I've found a single user (CodeExporter) and it lacks testing.
So next step I will add tests and then re-implement the method to see that without `asFileName` still is green.

src/CodeExport/CodeExporter.class.st
```diff
@@ -12,9 +12,15 @@ Class {
...
-	extension := stOrCsFlag ifTrue: ['st']  ifFalse: ['cs'].
-	targetFile := FileLocator imageDirectory / baseName asFileName, extension.
+	| extension targetFile validFilename targetFilename |
+	
+	extension := stOrCsFlag ifTrue: ['.st'] ifFalse: ['.cs'].
+	
+	validFilename := FilePathEncoder encode: (FileSystem disk 
+		checkName: baseName
+		fixErrors: true).
+	targetFilename := (FilePathEncoder decode: validFilename) , extension.
```
guillep: @hernanmd what I mean is to remove this here too!
There is no need to encode the paths, `FileSystem` and `FileLocator` do it already.

guillep (changes requested):

src/CodeExport/CodeExporter.class.st
Ducasse: Yes I did not have the time to look but this was my point.

jecisc:
Now the result is different:

```
FLFuelOutTests
 ✗ #testFuelOutWriteArrayFile (2ms)
TestFailure: Assertion failed
FLFuelOutTests(TestAsserter)>>assert:description:resumable:
FLFuelOutTests(TestAsserter)>>assert:description:
FLFuelOutTests(Object)>>assert:
FLFuelOutTests>>testFuelOutWriteArrayFile ...assert: fileName asFileReference exists
FLFuelOutTests(TestCase)>>performTest
 ✗ #testFuelOutWriteFileVersioned (6ms)
TestFailure: Assertion failed
FLFuelOutTests(TestAsserter)>>assert:description:resumable:
FLFuelOutTests(TestAsserter)>>assert:description:
FLFuelOutTests(Object)>>assert:
FLFuelOutTests>>testFuelOutWriteFileVersioned ...assert: fileName asFileReference exists
FLFuelOutTests(TestCase)>>performTest
 ✗ #testFuelOutWriteFile (1ms)
TestFailure: Assertion failed
FLFuelOutTests(TestAsserter)>>assert:description:resumable:
FLFuelOutTests(TestAsserter)>>assert:description:
FLFuelOutTests(Object)>>assert:
FLFuelOutTests>>testFuelOutWriteFile ...assert: fileName asFileReference exists
FLFuelOutTests(TestCase)>>performTest
```

hernanmd:
I saw now the failures seems not related (@jecisc can you confirm?)

```
osx-64 / Tests-osx-64 / testExternalIDEDependencies – MacOSX64.System.DependenciesTests.SystemDependenciesTest7s
unix-64 / Tests-unix-64 / testExternalIDEDependencies – Unix64.System.DependenciesTests.SystemDependenciesTest10s
osx-64 / Tests-osx-64 / testFlushOtherEndClosed – MacOSX64.Network.Tests.SocketStreamTest7s
unix-64 / Tests-unix-64 / testFlushLargeMessageOtherEndClosed – Unix64.Network.Tests.SocketStreamTest
```

jecisc:
Yes. But I've seen failures in the NewTools PR I think

Ducasse:
For P13

PR #16176, opened 2024-02-15
Integrate BeautifulComments directly into Pharo

### LINKED ISSUE #16171: BeautifulComment should be managed inside Pharo
Having a separate repository is just overkill for a package with 3 classes. 
I would like to merge it in Pharo.

jecisc:
Fixes #16171


Ducasse:
Oh thanks a LOT!!!

PR #16177, opened 2024-02-15
integrating Debug Points model into Pharo

adri09070:
This PR integrates a new tool to Pharo (at least the model) that will replace the existing breakpoints and watchpoints model.
The original repository is here: https://github.com/adri09070/PharoDebugPoints (which is a fork of another repository: https://github.com/Max-Zurbriggen/PharoDebugPoints from @Max-Zurbriggen ).

This tool allows to install debug points on different type of targets:
- an AST node, which  can be targeted by selecting source code in the Calypso browser
- a variable, which can be targeted by selecting a variable in the variable view in the Calypso browser
- an object, in order to install object-centric debug points. In this case, an object target wraps another target (AST or variable)

Debug points are different types of instrumentation that are used to debug. For now, there are two concret types of debug points: breakpoints and watchpoints.

A debug point can have different types of behaviors that will execute before the debug point is actually hit.

Among these behaviors, there are:
- Once behavior: the debug point becomes disabled after it is hit
- Count behavior: counts the number of times the debug point is hit
- Condition behavior: sets a condition that must evaluate to `true` to hit the debug point
- Chain Link behavior: combines debug points into a chain. In that case, when a debug point is hit in the chain, it disables itself and enables the next debug point in the chain
- Script behavior: executes a script each time a debug point is hit
- Transcript behavior: logs a string to the Transcript each time a debug point is hit

All these behaviors can be set when creating a breakpoint from the Calypso browser in the method editor.

To have more flexibility to configure existing debug points or to create different types of debug points, it is possible to use an API on `DebugPoint` objects or via the `DebugPointManager`.

It will be possible to configure debug points via a UI tool, the Debug Point Browser, which will be the object of another PR in NewTools


adri09070:
There is a bootstrap error, I need to check

Ducasse:
This is strange. Let us know if you need help.

hernanmd:
> There is a bootstrap error, I need to check

Could you point us to the error?
Because in the CI I only see `Instance of MetacelloMCBaselineOfProjectSpec did not understand #requires:`

src/BaselineOfBasicTools/BaselineOfBasicTools.class.st
```diff
@@ -47,10 +47,14 @@ BaselineOfBasicTools >> baseline: spec [
 		spec
 			baseline: 'UI' with: [ spec repository: repository ];
 			baseline: 'Reflectivity' with: [ spec repository: repository ];
-			baseline: 'Athens' with: [
+			baseline: 'DebugPoints' with: [
 				spec
-					loads: 'Cairo-core';
+					requires: #( 'Reflectivity' );
```
guillep: this is the issue. We cannot have requires clauses in baselines, only in packages.
If `DebugPoints` depends on reflectivity, that dependency should be inside `BaselineOfDebugPoints`

guillep (changes requested):

Ducasse:
I learned something :) tx!

src/DebugPoints/DebugPointManager.class.st
```diff
@@ -0,0 +1,374 @@
...
+}
+
+{ #category : 'announcement handling' }
+DebugPointManager class >> handleClassRemoved: anAnnouncement [
+
+	DebugPoint all copy do: [ :debugPoint |
+		debugPoint link methods
+			detect: [ :m | m methodClass = anAnnouncement classRemoved ]
+			ifFound: [ debugPoint remove ] ]
+]
```
StevenCostiou, line 21: For later: because the DebugPoint model the breakpoints at a higher level than metalinks, it would simplify things (like that method) to know more of the underlying model, for instance here it would make sense that each debug points knows the structural elements on which it is installed/active.

In this case the debug point would know the methods it targets.

src/DebugPoints/DebugPointManager.class.st
```diff
@@ -0,0 +1,374 @@
...
+	"adding behaviors"
+	aListOfBehaviorClasses do: [ :bhc | dp addBehavior: bhc new ].
+
+	self notifyDebugPointAdded: dp.
+
+	^ dp
+]
+
+{ #category : 'API' }
+DebugPointManager class >> installNew: aDebugPointClass On: aNode [
```
StevenCostiou: the `On:` should not be capitalized

src/DebugPoints/DebugPointManager.class.st
```diff
@@ -0,0 +1,374 @@
...
+
+	^self installNew: aDebugPointClass On: aNode withBehaviors: {  }.
```
StevenCostiou: `On` should not be capitalized.

src/DebugPoints/DebugPointManager.class.st
```diff
@@ -0,0 +1,374 @@
...
+]
+
+{ #category : 'API' }
+DebugPointManager class >> installNew: aDebugPointClass On: aNode forObject: anObject [
```
StevenCostiou: `On` should not be capitalized.

src/DebugPoints/DebugPointManager.class.st
```diff
@@ -0,0 +1,374 @@
...
+
+	^self installNew: aDebugPointClass On: aNode forObject: anObject withBehaviors: {  }.
+]
+
+{ #category : 'API' }
+DebugPointManager class >> installNew: aDebugPointClass On: aNode forObject: anObject withBehaviors: aListOfBehaviorClasses [
```
StevenCostiou: `On` should not be capitalized.

src/DebugPoints/DebugPointManager.class.st
```diff
@@ -0,0 +1,374 @@
...
+	dp := aDebugPointClass new
+		      node: aNode;
+		      targetInstance: anObject;
+		      yourself.
+
+	^ self installDebugPoint: dp withBehaviors: aListOfBehaviorClasses
+]
+
+{ #category : 'API' }
+DebugPointManager class >> installNew: aDebugPointClass On: aNode withBehaviors: aListOfBehaviorClasses [
```
StevenCostiou: `On` should not be capitalized.

src/DebugPoints/DebugPointManager.class.st
```diff
@@ -0,0 +1,374 @@
...
+
+	"this should be the single point of installation for debug points, removal is in the remove method of debug point instances"
+
+	^ self
+		  installNew: aDebugPointClass
+		  forObject: anObject
+		  onVariableNamed: aSlotNameSymbol
+		  accessStrategy: anAccessStrategySymbol
+		  withBehaviors: {  }
+]
```
StevenCostiou, line 96: We can install a debug point without behaviors?
What should it do in this case?

src/DebugPoints/DebugPointManager.class.st
```diff
@@ -0,0 +1,374 @@
...
+	"access strategy can be #all or #read or #write"
+
+	"this should be the single point of installation for debug points, removal is in the remove method of debug point instances"
+
+	^ self
+		  installNew: aDebugPointClass
+		  forObject: anObject
+		  onVariableNamed: aSlotNameSymbol
+		  accessStrategy: #read
+		  withBehaviors: {  }
```
StevenCostiou, line 126: Same remark about empty behaviors.

src/DebugPoints/DebugPointManager.class.st
```diff
@@ -0,0 +1,374 @@
...
+	"access strategy can be #all or #read or #write"
+
+	"this should be the single point of installation for debug points, removal is in the remove method of debug point instances"
+
+	^ self
+		  installNew: aDebugPointClass
+		  forObject: anObject
+		  onVariableNamed: aSlotNameSymbol
+		  accessStrategy: #write
+		  withBehaviors: {  }
```
StevenCostiou, line 154: Same remark on empty behaviors.

src/DebugPoints/DebugPointManager.class.st
```diff
@@ -0,0 +1,374 @@
...
+	"access strategy can be #all or #read or #write"
+
+	"this should be the single point of installation for debug points, removal is in the remove method of debug point instances"
+
+	^ self
+		  installNew: aDebugPointClass
+		  inClass: aClass
+		  onVariableNamed: aSlotNameSymbol
+		  accessStrategy: #all
+		  withBehaviors: {  }
```
StevenCostiou, line 182: Same remark about empty behaviors.

src/DebugPoints/DebugPointManager.class.st
```diff
@@ -0,0 +1,374 @@
...
+		  withBehaviors: aListOfBehaviorClasses
+]
+
+{ #category : 'announcements' }
+DebugPointManager class >> notifyDebugPointAdded: aDebugPoint [
+
+	| announcement |
+	announcement := DebugPointAdded 
+		on: aDebugPoint
+		nodes: aDebugPoint link nodes.
```
StevenCostiou, line 330: On the question of high/low level concernes, I wonder if the nodes should be known by the model or only by the underlying implementation? @MarcusDenker do you have an opinion on that?

src/DebugPoints/DebugPointManager.class.st
```diff
@@ -0,0 +1,374 @@
...
+]
+
+{ #category : 'removing' }
+DebugPointManager class >> removeFromMethod: aMethod [
+
+	DebugPoint all copy do: [ :debugPoint | 
+		debugPoint link methods
+			detect: [ :m | m == aMethod ]
+			ifFound: [ debugPoint removeFromMethod: aMethod ] ]
+]
```
StevenCostiou, line 374: Same remark, for future work: the debug point model should probably know the methods it affects without having to ask them to the metalink.

StevenCostiou (changes requested):
I did only a partial pass, I will continue later.
In this pass, there are typos that needs to be fixed and comments for future work (*i.e.*, not to be treated in this PR).

src/DebugPoints/DebugPointManager.class.st
adri09070: Well, if it is a basic debug point: nothing. If it is a breakpoint or a watchpoint, it will just be a basic breakpoint or watchpoint.

Anyway, this is not particularly bad. We may want to create a debug point and configure it later

MarcusDenker:
This was merged with failing tests
(the CI was down, all PRs where green because no tests where run)

PR #16186, opened 2024-02-15
Cleanup: Remove <- from collection

### LINKED ISSUE #16185: Cleanup: Remove <- from collection
There is a method called #<- implemented in Collection
- it just has a single sender in AdditionalMethodState class>>#propertiesToPersist (the other two are tests for it) 
- and AdditionalMethodState class>>#propertiesToPersist could be more easily written using #asSet

We should deprecate and remove it.

astares:
Fix #16185

### DIFF
--- a/src/Collections-Abstract-Tests/TCreationWithTest.trait.st
+++ b/src/Collections-Abstract-Tests/TCreationWithTest.trait.st
@@ -30,18 +30,6 @@ TCreationWithTest >> test0FixtureCreationWithTest [
 ]
 
 { #category : 'tests - creation' }
-TCreationWithTest >> testLessThanMinus [
-
-	| aCol collection |
-	collection := self collectionMoreThan5Elements asArray.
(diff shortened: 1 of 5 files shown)


jecisc:
This was a new feature implemented in P12 IIRC

astares:
Feature for what? Confusing people?

jecisc:
To be able to write code such as `Set <- #(1 2 3)`

jecisc:
Originally Stephane wanted to have something like `{ :Set 1 . 2 . 3 }` but some people prefer not to change Pharo's parser and proposed this alternative

astares:
> To be able to write code such as `Set <- #(1 2 3)`

What about:    Set withAll: #(1 2 3) ?

Ducasse:
The problem is that withAll: is a keyword so you have to put () all over the place. This is why I like {:Set 1 . 2 . 3} in the first place. At least it was fun to implement.

PR #16195, opened 2024-02-16
Fix tests in metacello, Spec font chooser and date comments

guillep:
Subject.

Remaining are some tests in refactorings that I did not know how to fix.
Probably @Ducasse or @balsa-sarenac could help with that.


Ducasse:
Sure we will have a look.

PR #16198, opened 2024-02-18
Fix: extract method failing tests

balsa-sarenac:
Extract method has a breaking change precondition now, because of it tests started failing since some code is applicable to be extracted but it breaks the execution flow. To fix it I've modified extract method transformation to raise warning when this scenario happened. So this is a refactoring now, not a real transformation anymore.

Extract method transformation is a refactoring now.
We need to migrate it to its package and rename.
However, this transformation does not work as well as expected, so we should wait a bit with that.


Ducasse:
ok! Waiting for the build to finish

balsa-sarenac:
Failing tests seems like unrelated to this PR.
The #16199 fixes RBParser and Node tests that are failing.

PR #16201, opened 2024-02-18
Refactor MenuItemMorph, PluggableMenuItemSpec and PragmaMenuAndShortcutRegistrationItem to support use of a FormSet and apply that to CmdCommand

Rinzwind:
This pull request refactors MenuItemMorph to support use of a FormSet for the icon and applies that to the CmdCommand hierarchy. In the CmdCommand hierarchy, the overrides of `#defaultMenuIcon` are refactored to override `#defaultMenuIconName` instead.

### DIFF
--- a/src/MenuRegistration/PragmaMenuAndShortcutRegistrationItem.class.st
+++ b/src/MenuRegistration/PragmaMenuAndShortcutRegistrationItem.class.st
@@ -13,6 +13,7 @@ Class {
 		'label',
 		'help',
 		'icon',
+		'iconFormSet',
 		'selector',
 		'arguments',
 		'withSeparatorAfter',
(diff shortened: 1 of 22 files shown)


Rinzwind:
I converted this to a draft. Commit 02f48dcb43a0cc8d introduces a bug in `#setUpIconForMenuItem:` where the argument, despite the variable name ‘aMenuItemMorph’, can also be a PragmaMenuAndShortcutRegistration, which does not understand `#iconFormSet:`.

Rinzwind:
I fixed the bug referred to in my previous message by first refactoring PluggableMenuItemSpec and PragmaMenuAndShortcutRegistrationItem as well.

src/MenuRegistration/PragmaMenuAndShortcutRegistrationItem.class.st
```diff
@@ -13,6 +13,7 @@ Class {
 		'label',
 		'help',
 		'icon',
```
jecisc:
```suggestion
```

jecisc: This instance variable is not used anymore and should be removed

jecisc (changes requested):

src/MenuRegistration/PragmaMenuAndShortcutRegistrationItem.class.st
Rinzwind: Forgot to actually remove it indeed, fixed in commit 49a7a887efa18e50.

Ducasse:
@jecisc can you check it? Tx

Rinzwind:
@jecisc Could I ask for an update on the (re)review? If you’re not available, it might be better to let someone else take over?

jecisc (approved):

PR #16205, opened 2024-02-19
16204-RubPluggableTextFieldMorph-does-not-open-drop-list

### LINKED ISSUE #16204: RubPluggableTextFieldMorph does not open drop list
RubPluggableTextFieldMorph does not open the drop list when clicking the button.
Problem in P12.
Generated by the removal of Display to calculate available space (#15945)

tesonep:
We need to implement the allowed area delegating to the RealStateAgent so it uses the information of the size of the world.
Fix #16204

### DIFF
--- a/src/Polymorph-Widgets/IdentifierChooserMorph.class.st
+++ b/src/Polymorph-Widgets/IdentifierChooserMorph.class.st
@@ -102,7 +102,7 @@ IdentifierChooserMorph >> activate: evt [
 	"Backstop."
 ]
 
-{ #category : 'as yet unclassified' }
+{ #category : 'accessing' }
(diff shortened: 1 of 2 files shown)


src/Polymorph-Widgets/IdentifierChooserMorph.class.st
```diff
@@ -102,6 +102,12 @@ IdentifierChooserMorph >> activate: evt [
 	"Backstop."
 ]
 
+{ #category : 'as yet unclassified' }
```
jecisc:
```suggestion
{ #category : 'accessing' }
```

jecisc (changes requested):

PR #16206, opened 2024-02-19
16132-Caret-return-char-broken-in-Pharo-11-macOS-Keyboard-as-US-International---PC

### LINKED ISSUE #16132: Caret (return char) broken in Pharo 11 macOS Keyboard as U.S. International - PC
**Bug description**
The caret is not working in Pharo 11 with US-International as it used to be in previous Pharo versions.

**To Reproduce**
1. Open a Playground
2. Ensure Keyboard is in `U.S. International - PC`  mode
3. Type ` Shift-6` for the caret
4. It prints something different than the return caret

**Expected behavior**
The return caret should be printed on key up.

**Screenshots**
caret

**Version information:**
 - OS: macOS
 - Version: 12.7.2
 - Pharo Version 11

**Expected development cost**
Any investigation clues? What class controls this behavior?

tesonep:
Fix #16132
We have to convert the  'Modifier Letter Circumflex Accent (U+02C6) without a modifying character to the circunflex (U+005E). Even though it is a valid entry from the user, it makes complicated as it looks like the circunflex but it is not handled as one by the parser.

### DIFF
--- a/src/OSWindow-Core/OSWindowMorphicEventHandler.class.st
+++ b/src/OSWindow-Core/OSWindowMorphicEventHandler.class.st
@@ -157,6 +157,16 @@ OSWindowMorphicEventHandler >> convertPosition: aPosition [
 	^ morphicWorld worldState worldRenderer convertWindowMouseEventPosition: aPosition
 ]
 
+{ #category : 'converting' }
+OSWindowMorphicEventHandler >> convertText: aString [
+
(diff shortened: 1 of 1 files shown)


Ducasse:
Indeed on the student machines we have two different ^ character and one is not working so may be this fix will address this problem too.

jecisc (approved):

PR #16220, opened 2024-02-19
Enhancement check keyword casing via lint rule

### LINKED ISSUE #16207: Enhancement: Check keyword casing via lint rule
In issues like this https://github.com/pharo-project/pharo/pull/16177#pullrequestreview-1888150026
we see that some keyword messages wrongly contain keywords with uppercase while it typically should start lowercase:

image

So we should have the correct form for keyword selectors like:

- **#some: anObject with: anotherObject**

instead of: 

- **#some: anObject With: anotherObject**
- **#Some: anObject with: anotherObject**
- **#Some: anObject With: anotherObject**

checked via a lint rule.

At the moment the system is not lamenting about such cases:

image

We can easily check the compiled method if the selector represents a keyword message and if one of the keyword parts starts with uppercase using the following code:

```Smalltalk
self selector isKeyword 
and: [ 
	(self selector asString substrings: #($:)) 
		anySatisfy: [:each | each first isUppercase ] ]
```

astares:
Fix #16207

### DIFF
--- a/src/General-Rules-Tests/FloatReferencesRuleTest.class.st
+++ b/src/General-Rules-Tests/FloatReferencesRuleTest.class.st
@@ -4,8 +4,9 @@ Class {
 	#instVars : [
 		'testClass'
 	],
-	#category : 'General-Rules-Tests',
-	#package : 'General-Rules-Tests'
+	#category : 'General-Rules-Tests-Base',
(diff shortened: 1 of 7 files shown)


astares:
Unrelated CI issue, one platform green, two yellow

image

jecisc (approved):

PR #16222, opened 2024-02-21
add a way to allow some obsoletes through an environment variable

estebanlm:
to be used on the CI of iceberg, where we need to unload the whole package and after reload it to properly run tests.


jecisc (approved):

MarcusDenker:
yes, this is good... obsoletes are hard to track down.

PR #16227, opened 2024-02-23
Moved #signalAll to Semaphore class instead of being an extension

### LINKED ISSUE #16094: Move `Semaphore>>signalAll` extension to Semaphore class
Pharo12

This extension extends the Semaphore API and makes clients indirectly depend on ThreadedFFI.
This broke PRs such as https://github.com/pharo-project/pharo/pull/16073

This extension should be made a normal method into its class.

jordanmontt:
Fixes #16094


MarcusDenker (approved):

MarcusDenker:
we should wait for the CI

PR #16228, opened 2024-02-23
In TerseGuide fix  documentation

### LINKED ISSUE #16203: Terse Guide "detect:" Collection method documentation
**Bug description**
In the Terse Guide, for Collection, "detect:" is documented as "**find position of** first element that passes test", which is wrong.

**To Reproduce**
Read the Terse Guide.

**Expected behavior**
"detect:" should be documented as "**return** first element that passes test"

uNouss:
Fixes #16203


MarcusDenker (approved):
Thanks!

PR #16232, opened 2024-02-23
Added a new stack class that does not inherit from LinkedList

jordanmontt:
Implemented a new stack class that does not inherit from Collection, but from Object.
The current Stack class inherits from LinkedList which can cause problems. With this new implementation, we have a smaller API.

We deleted the old Stack class, and we created the new one with the same name.

Fixes # #2863 
Worked in the Pharo sprint with @hernanmd


Ducasse:
Hi Seb 
I already did it in Pharo-container. I think that it would be good to use it.

jordanmontt:
Hello Stef,

Ah yes, I just checked your implementation of the Stack. We have basically the same implementation but you use an OrderedCollecion to store the elements while me I use a LinkedList. How do you want to proceed with this?

MarcusDenker:
Build failure:

#findFirst: is missing


```

MetacelloNotification: Loaded -> Calypso-SystemPlugins-Monticello-Browser --- tonel:///builds/workspace/est_and_branch_Pipeline_PR-16232/64/src()
MetacelloNotification: Loaded -> Gofer-Deprecated --- tonel:///builds/workspace/est_and_branch_Pipeline_PR-16232/64/src()
MetacelloNotification: Loaded -> Deprecated12 --- tonel:///builds/workspace/est_and_branch_Pipeline_PR-16232/64/src()
Removing credential.Instance of Stack did not understand #findFirst:
Stack(Object)>>doesNotUnderstand: #findFirst:
MicInlineParser>>findAnyType:
MicInlineParser>>closerOnlyCase
MicInlineParser>>delimiterFoundProcess
```

jordanmontt:
I will close this. In any case we should keep the one that is implemented in pharo-containers.

PR #16236, opened 2024-02-23
fixing reflectivity bug that executes a #before metalink twice if an #after metalink is installed 

### LINKED ISSUE #15283: MetaLink: `before` is executed twice if there is also an `after`
**Bug description**
If you install a MetaLink with control set to `before` **and** another MetaLink set to `after` on the same node, the `before` link will be executed twice.

**To Reproduce**
1. Install the following class in an image or reproduce equivalent methods:
```st
'From Pharo12.0.0 of 30 October 2023 [Build information: Pharo-12.0.0+build.978.sha.d7499e6e3bb54029bc964f77547690a641addaeb (64 Bit)] on 8 November 2023 at 10:56:29.699437 pm'!
Object subclass: #MyClass
	instanceVariableNames: ''
	classVariableNames: ''
	package: 'MyPackage'!

!MyClass methodsFor: 'command line' stamp: 'toto 11/8/2023 22:35'!
hello

	'hello' traceCr! !


!MyClass methodsFor: 'actions' stamp: 'toto 11/8/2023 22:45'!
install

	| beforeMetaLink afterMetaLink method |
	method := self class >> #hello.
	beforeMetaLink := MetaLink new
		                  metaObject: self;
		                  selector: #enterMethod;
		                  control: #before;
		                  yourself.
	afterMetaLink := MetaLink new
		                 metaObject: self;
		                 selector: #exitMethod;
		                 control: #after;
		                 yourself.
	method ast
		link: beforeMetaLink;
		link: afterMetaLink! !

!MyClass methodsFor: 'actions' stamp: 'toto 11/8/2023 22:32'!
enterMethod

	'enter' traceCr! !

!MyClass methodsFor: 'actions' stamp: 'toto 11/8/2023 22:31'!
exitMethod

	'exit' traceCr! !

"-- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- "!

MyClass class
	instanceVariableNames: ''!

!MyClass class methodsFor: 'command line' stamp: 'toto 10/31/2023 10:41'!
hello! !
```
3. Open the Transcript
4. Run `MyClass new install; hello` in the playground

**Expected behavior**
The `before` link should only be executed once.

**Screenshots**
image

**Version information:**
 - OS: macOS
 - Version: 12.6
 - Pharo Version 12

adri09070:
Fixes #15283 

Done with @ValentinBourcier and @RemiDufloer 

When installing two metalinks on the same node, one metalink #before and one metalink #after, the metalink #before was executed twice.

This is because, when executing the instrumented method, it executes a reflective method via `ReflectiveMethod>>#run: aSelector with: anArray in: aReceiver`, which calls:

```Smalltalk
ReflectiveMethod>>#compileAndInstallCompiledMethod
compileAndInstallCompiledMethod
	self wrapperNeeded ifTrue: [ self generatePrimitiveWrapper ].
	self recompileAST.
	self installCompiledMethod
```

When a metalink #after is installed, the method `ReflectiveMethod>>#generatePrimitiveWrapper` is called:

```Smalltalk
generatePrimitiveWrapper
	| wrappedMethod send wrapperMethod assignmentNode |
	wrappedMethod := self compileAST.

	send := RBMessageNode
		receiver: RBVariableNode selfNode
		selector:  #rFwithArgs:executeMethod:
		arguments: {RBArrayNode statements: ast arguments . (RBLiteralNode value: wrappedMethod)}.

     ...

```

This method wraps a method inside a reflective call and this reflective call should ignore metalinks.

However, `wrappedMethod` is obtained by recompiling the AST, which generates a method that contain the #before metalink. Indeed, when inspecting `wrappedMethod`, in the screenshot below, the first 3 bytecodes come from a #before metalink:

image

So the reflective call created by the #after metalink also executes the code of the #before metalink.
As the AST is later recompiled again in `#compileAndInstallCompiledMethod`, the code of the #before metalink is also added at this moment again, resulting the #before metalink to be executed twice: once before the reflective call and once inside the reflective call.

To fix that, we just use the simple compiled method because we don't want to recompile the AST to wrap the reflective method, but we want to wrap the compiled method insted to ensure that the method without metalink is executed inside the reflective call:

```Smalltalk
generatePrimitiveWrapper
	| wrappedMethod send wrapperMethod assignmentNode |
	wrappedMethod := self compileAST.

	send := RBMessageNode
		receiver: RBVariableNode selfNode
		selector:  #rFwithArgs:executeMethod:
		arguments: {RBArrayNode statements: ast arguments . (RBLiteralNode value: self compiledCode)}.

     ...

```

### DIFF
--- a/src/Reflectivity-Tests/ReflectivityControlTest.class.st
+++ b/src/Reflectivity-Tests/ReflectivityControlTest.class.st
@@ -6,7 +6,8 @@ Class {
 	#superclass : 'ReflectivityTestCase',
 	#instVars : [
 		'tag',
-		'link'
+		'link',
+		'link2'
 	],
 	#category : 'Reflectivity-Tests-Base',
 	#package : 'Reflectivity-Tests',
@@ -36,6 +37,7 @@ ReflectivityControlTest >> tagExec: aTag [
 { #category : 'running' }
 ReflectivityControlTest >> tearDown [
 	link ifNotNil: [ link uninstall ].
+	link2 ifNotNil: [ link2 uninstall ].
 	super tearDown
 ]
 
@@ -428,6 +430,41 @@ ReflectivityControlTest >> testAfterVariableNode [
 	self assert: tag equals: 'yes'.
 	self assert: (ReflectivityExamples >> #exampleGlobalRead) class equals: CompiledMethod.
 	self deny: (ReflectivityExamples >> #exampleGlobalRead) isQuick
+]
+
+{ #category : 'tests' }
+ReflectivityControlTest >> testBeforeAndAfterSend [
+
+	| sendNode collection |
+	collection := OrderedCollection new.
+	sendNode := (ReflectivityExamples >> #exampleMethod) sendNodes first.
+	self assert: sendNode isMessage.
+	link := MetaLink new
+		        metaObject: [ collection add: #before ];
+		        selector: #value;
+		        control: #before;
+		        arguments: #(  ).
+	link2 := MetaLink new
+		         metaObject: [ collection add: #after ];
+		         selector: #value;
+		         control: #after;
+		         arguments: #(  ).
+	sendNode link: link.
+	sendNode link: link2.
+
+	self assert: sendNode hasMetalinkAfter.
+	self assertEmpty: collection.
+	self
+		assert: (ReflectivityExamples >> #exampleMethod) class
+		equals: ReflectiveMethod.
+
+	self assert: ReflectivityExamples new exampleMethod equals: 5.
+	self
+		assertCollection: collection
+		equals: #( #before #after ) asOrderedCollection.
+	self
+		assert: (ReflectivityExamples >> #exampleMethod) class
+		equals: CompiledMethod
 ]
 
 { #category : 'tests - before' }
--- a/src/Reflectivity/ReflectiveMethod.class.st
+++ b/src/Reflectivity/ReflectiveMethod.class.st
@@ -98,26 +98,30 @@ ReflectiveMethod >> flushCache [
 
 { #category : 'evaluation' }
 ReflectiveMethod >> generatePrimitiveWrapper [
+
 	| wrappedMethod send wrapperMethod assignmentNode |
 	wrappedMethod := self compileAST.
 
 	send := RBMessageNode
-		receiver: RBVariableNode selfNode
-		selector:  #rFwithArgs:executeMethod:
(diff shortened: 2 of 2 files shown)


Ducasse:
Great! I will wait for marcus review before integrating it.

adri09070:
Looks like unrelated failing tests

adri09070:
@StevenCostiou

MarcusDenker:
I wonder if we, with this change, still take the links into account that are put on the AST "inside" of the method that is wrapped.  e.g. imagine you have a primitive and want to add links on the fail code:



```
~~ anObject
	"Primitive. Answer whether the receiver and the argument are different objects
	(do not have the same object pointer). Do not redefine the message ~~ in
	any other class! Optional (Assuming == is essential). No Lookup. Do not override in any subclass.
	See Object documentation whatIsAPrimitive."

	<primitive: 169>
	self == anObject
		ifTrue: [^ false]
		ifFalse: [^ true]
```


But I think we can merge this change and then continue (the case that you fix is much more common *and* the tests are important).

I will do a test next for the case that I have in mind next.

MarcusDenker:
(re-started a test run, just to make sure before merge)

MarcusDenker (approved):
Good to merge as a first step

adri09070:
> I wonder if we, with this change, still take the links into account that are put on the AST "inside" of the method that is wrapped. e.g. imagine you have a primitive and want to add links on the fail code:
> 
> ```
> ~~ anObject
> 	"Primitive. Answer whether the receiver and the argument are different objects
> 	(do not have the same object pointer). Do not redefine the message ~~ in
> 	any other class! Optional (Assuming == is essential). No Lookup. Do not override in any subclass.
> 	See Object documentation whatIsAPrimitive."
> 
> 	<primitive: 169>
> 	self == anObject
> 		ifTrue: [^ false]
> 		ifFalse: [^ true]
> ```
> 
> But I think we can merge this change and then continue (the case that you fix is much more common _and_ the tests are important).
> 
> I will do a test next for the case that I have in mind next.

That is something I was wondering too and I think you're right but I couldn't find a better fix. 
For example, I thought that with this fix, if we set a metalink `#instead` and a metalink `#after` on the same node, the metalink `#instead` would be ignored...
The thing is we tried and it is not possible, even now, without my fix.
So we didn't search much further.

I am aware that the fix is not perfect, but at least it fixes this bug and if other bugs do appear, we will know where they come from.
The wrapped method was obtained by recompiling the AST and I think the "perfect solution" would be that this recompilation ignore the `#before:` metalink

PR #16244, opened 2024-02-26
use the correct integration branch for iceberg and Pharo

estebanlm:
to be consistent with other external projects


jecisc (approved):
Failing tests all seems to be unrelated

PR #16247, opened 2024-02-26
Fix adding slots in Ring

jecisc:
A problem got introduced in ring with this commit: https://github.com/pharo-project/pharo/commit/d105e1a9715e188eee8f40b7de0120e1c3229600

This commit removed an #asOrderedCollection and then we could have slots in an array. With this, it makes it impossible to add a new slot to a class or a trait. 

This change reverts this change and adds a comment about why we are having the #asOrderedCollection

This is necessary to have Moose working in P12


jecisc:
Failing tests are unrelated

MarcusDenker (approved):
A bit strange that this is only done on #markResolved, but with the comment it is good to add

PR #16250, opened 2024-02-27
Fix: multiple exceptions when executing a rb move class transformation

### LINKED ISSUE #15809: Multiple exceptions when executing a RB move class transformation
**Bug description**

First I get a "NonBooleanReceiver: proceed for truth." exception, originated in method `RBMoveClassTransformation>>applicabilityPreconditions`. Once fixed (1), another MNU is raised with message "Instance of ExampleClass class did not understand #tagName" when trying to repackage a class.

**To Reproduce**
```smalltalk
" Create an example class, required by the transformation "
Smalltalk classInstaller make: [ :aSlotClassBuilder |
		aSlotClassBuilder
			superclass: Object;
			name: #MyClass;
			package: 'My-Package1' ].

" Create target package where MyClass will be moved "
PackageOrganizer default addPackage: 'My-Package2'.

" Execute the transformation "
(RBMoveClassTransformation 
		move: #MyClass
		toPackage: 'My-Package2')
			execute.
```

**Expected behavior**
Class should be repackaged without any issues. Replacing `execute`  by `generateChanges` does not execute any changes.

(1) A fix for this method could be:

```smalltalk
applicabilityPreconditions

	^ self classExist & (RBCondition
		   withBlock: [
			   (self model environment packageAt: self packageName ifAbsent: [ nil ])
				   ifNil: [ false ]
				   ifNotNil: [ :package | self tagName 
						ifNil: [ true ]
						ifNotNil: [ :name | package hasTag: name ] ] ]
		   errorString: 'No package named ' , self packageName , (self tagName
				    ifNil: [ '' ]
				    ifNotNil: [ :name | ' with tag named ' , name ]))
```

balsa-sarenac:
Fixes: #15809

### DIFF
--- a/src/Refactoring-Changes/RBClassRepackagingChange.class.st
+++ b/src/Refactoring-Changes/RBClassRepackagingChange.class.st
@@ -25,7 +25,7 @@ RBClassRepackagingChange class >> repackage: aClass in: aPackage tag: aTag [
 { #category : 'converting' }
 RBClassRepackagingChange >> asUndoOperation [
 
-	^ self class repackage: self changeClass in: self changeClass packageName tag: self changeClass tagName
+	^ self class repackage: self changeClass in: self changeClass packageName tag: self changeClass packageName
 ]
 
 { #category : 'accessing' }
@@ -56,8 +56,8 @@ RBClassRepackagingChange >> changeString [
 RBClassRepackagingChange >> generateChanges [
 
 	self changeClass
-		packageName: self packageName;
-		tagName: self tagName
+		package: packageName
+		tag: tagName 
 ]
 
 { #category : 'accessing' }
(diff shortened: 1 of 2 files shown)


src/Refactoring-Changes/RBClassRepackagingChange.class.st
```diff
@@ -25,7 +25,7 @@ RBClassRepackagingChange class >> repackage: aClass in: aPackage tag: aTag [
 { #category : 'converting' }
 RBClassRepackagingChange >> asUndoOperation [
 
-	^ self class repackage: self changeClass in: self changeClass packageName tag: self changeClass tagName
+	^ self class repackage: self changeClass in: self changeClass packageName tag: self changeClass packageName
```
jecisc: I'm not sure about this change. It seems to me that we are losing an information using the package name as the tag name

balsa-sarenac: Thanks for catching this, I had this in the first version, but I guess I lost it along the way.

jecisc (approved):

PR #16257, opened 2024-02-28
Minor improvement in TF test class

tinchodias:
Rename unixModuleName as unix64LibraryName

Superclass definition is:

FFILibrary>>
unixLibraryName
	"Kept for backward compatibility.
	The system complains that this method is not implemented as it will only exist
	in subclasses that use the old API.
	 Users should use unix32* or unix64*"
	^ self unixModuleName

and:

unixLibraryName
	"Kept for backward compatibility.
	The system complains that this method is not implemented as it will only exist
	in subclasses that use the old API.
	 Users should use unix32* or unix64*"
	^ self unix64LibraryName


tinchodias:
Interesting. As a reference: The Roassal3 polygon supports it (See `RSPolygon>>cornerRadii:`).
A method comment links to https://riptutorial.com/html5-canvas/example/18766/render-a-rounded-polygon-

estebanlm (approved):

PR #16258, opened 2024-02-28
Cleanup: rename method refactoring driver to new architecture with breaking choices

### LINKED ISSUE #14830: Refactoring: Rename refactoring has problems
Rename refactoring has at least two problems if your new method name already exists in subclasses.

First the dialogue displays a "\n" instead of a line return.

But the most annoying is that if this is the case you can browse the methods with the same name, but you cannot apply the renaming. We could do that before :( 

@balsa-sarenac @Ducasse

balsa-sarenac:
Introduces some cleanups to the code, choices to:
- rename method
- browse overrides
- rename and browse overrides

This should fix: #14830 @jecisc could you please test and tell me what do you think of the new flow?


Ducasse:
Looks good to me.

PR #16260, opened 2024-02-28
Fix Bug: `OrderedDictionary>>#at:at:put:` not updating ordered keys

gcotelli:
Fix `OrderedDictionary>>#at:at:put:` and `#at:at:ifAbsentPut:` because it was not registering the key in the `orederedKeys` collection.

So if you send later `keysAndValuesDo:` for example the key is not present, because the dictionary is inconsistent with the orderedKeys collection.

For example, running
```smalltalk
| o |
o := OrderedDictionary new at: 2 at: 3 put: 3;yourself.
o keysAndValuesDo: [ :k :v | ]
```
will raise a debugger


gcotelli:
The Jenkis error looks totally unrelated to the code changes :shrug:

Ducasse:
Thanks a lot. 
This proves that we are not good at testing functionalities.

Issue #15104, opened 2023-10-25
[CI] Windows times out running the tests


MarcusDenker:
**Bug description**
The test run stage is reaching the 35 minute limit.

```
Running tests of Refactoring-UI
Running tests of ProfStef-Tests
Running tests of MonticelloFileServices
Running tests of MonticelloTonel-Core
Running tests of Compression-Tests
script returned exit code 127
```

jecisc:
There seems to be a problem during Epicea's tests. I found this in a log that timed out:

```
Beginning to run tests of Epicea-Tests

























































Finished to run tests of Epicea-Tests in 9 minutes 34 seconds 731 milliseconds
```

More than 9min for Epicea's test is way too long. We have this on the OSX slave:

```
Beginning to run tests of Epicea-Tests

Finished to run tests of Epicea-Tests in 440 milliseconds 849 microseconds
```

Rinzwind:
In [build 4 for pull request #16201](https://ci.inria.fr/pharo-ci-jenkins2/job/Test%20pending%20pull%20request%20and%20branch%20Pipeline/job/PR-16201/4/), the [Windows test run](https://ci.inria.fr/pharo-ci-jenkins2/job/Test%20pending%20pull%20request%20and%20branch%20Pipeline/job/PR-16201/4/execution/node/81/log/) timed out while running the package ‘Epicea-Tests’ (the ellipsis omits 91 empty lines):

```
Beginning to run tests of UnifiedFFI-Tests with random seed 319495693
Finished to run tests of UnifiedFFI-Tests in 1 second 410 milliseconds
Beginning to run tests of Epicea-Tests with random seed 161277837
[…]
Sending interrupt signal to process

script returned exit code -1
```

The [‘progress-Tests-windows-64.log’](https://ci.inria.fr/pharo-ci-jenkins2/job/Test%20pending%20pull%20request%20and%20branch%20Pipeline/job/PR-16201/4/execution/node/156/log/) shows that that was more specifically while running `#testFromDetectIfNotFound ` on EpLogTest:

```
starting testcase: EpLogTest>>testFromDetect ... finished
starting testcase: EpLogTest>>testFromDetectIfNotFound ... 
```

It might be useful for the progress log to show the run times for the individual test methods.

Rinzwind:
[...]

PR #16266, opened 2024-03-01
Make #reportTestCase:runBlock: on HDTestReport print the test run time on the progress log stream

Rinzwind:
As suggested in a [comment on issue #15104](https://github.com/pharo-project/pharo/issues/15104#issuecomment-1972188046), this pull request make `#reportTestCase:runBlock:` on HDTestReport print the test run time on the progress log stream.


Ducasse:
Good idea! Tx

PR #16268, opened 2024-03-01
Adding tests for debug points + integrating debug points to Calypso Browser

adri09070:
- Adding lots of tests that come from previous breakpoint tests
- Adding a system to notify when a debug point has been hit
- Adding possibility to add debug points on class variables
- Integrating DebugPoints to the Calypso browser (the package was there but not loaded by default)


Ducasse:
I'm not expert in SystemAnnouncer. 
I wonder if it would not be better to register to an Announcer local to the tests instead of the systemAnnouncer.
In addition I was wondering if the tests remove the registered objects.

Let us see what @jecisc @tesonep think

adri09070:
I fixed all the failing tests, those left are unrelated.

Regarding the System Announcer, I don't know, the old breakpoint tests were changing the system announcer's unique instance to nil so maybe I can do the same thing ?

Anyway, all registered objects unsubscribe themselves in the tests.

@StevenCostiou

StevenCostiou:
I do not remember why we used the system announcer.
Like this I do not see any problem of using an announcer on the side class of the debug points (or somewhere, but it has to be a sole-instance object).

Perhaps when trying again we will find there was a reason, and this time we should document it.
Or it will work :)

StevenCostiou:
Ah and for the tests in any case we could switch to a local announcer but we need the guarantee that, if the real system uses the system announcer both announcers behave the same.

adri09070:
I did some refactorings:

- removing calls to `#isKindOf:` that are costly and can be avoided
- splitting debug point behaviors into 2 categories: "check" behaviors and "side effect behaviors". All check behaviors determine whether the debug point should be hit, and are executed by design before side-effect behaviors . The API to use all those behaviors, however, does not change.

I won't touch this PR anymore

jecisc:
I checked with Adrien and Steven and Steven will probably do a review soon before we merge it.

estebanlm:
ping @StevenCostiou :P

StevenCostiou:
I will review it tomorrow morning

StevenCostiou (approved):

StevenCostiou:
@jecisc I'm good with it. Should we merge? there is red on the CI.

MarcusDenker:
The red CI is not related to this change (the CI passed for one plattfom with failing test not related)

PR #16269, opened 2024-03-01
Test: rename method driver

balsa-sarenac:
Modifies driver a bit to support testing rename method.

Depends on: #16258


Ducasse:
I resolved the conflict. I selected the Pharo 12 version since the alternative was empty. I hope it will be ok.

PR #16282, opened 2024-03-05
Remove use of ‘set -e’ from script ‘getPharoVM.sh’

Rinzwind:
This pull request removes the use of [‘set -e’](https://pubs.opengroup.org/onlinepubs/9699919799/utilities/V3_chap02.html#set:~:text=an%20individual%20file.-,%2De,an%20exit%20status%20greater%20than%20zero%29%2C%20the%20shell%20immediately%20shall%20exit,-%2C%20as%20if%20by) (added in pull request #14764) from the shell script ‘getPharoVM.sh’ so that when the download of the VM fails, the shell no longer immediately exits but continues to the commands for retrying the download:

https://github.com/pharo-project/pharo/blob/c1b6dac4d3ea2200b5affd5f5063e75dd8eb69a5/bootstrap/scripts/getPharoVM.sh#L25-L28

### DIFF
--- a/bootstrap/scripts/getPharoVM.sh
+++ b/bootstrap/scripts/getPharoVM.sh
@@ -1,7 +1,6 @@
 #!/usr/bin/env bash
 
 set -x
-set -e
 
 if [ $# -lt 1 ]
   then


Rinzwind:
I temporarily included a second commit that I will remove from this pull request: commit 56c05155835e21c6. That change is related to issue #15104 and is just to see whether the Epicea tests also take longer to run in the Windows test run when they are not executed after other tests. I didn’t mark this pull request as a draft for that because I’m not sure Jenkins will still run a build then.

Rinzwind:
I removed commit 56c05155835e21c6, and the ones after, from this pull request. Unfortunately, I still don’t know whether running the ‘Epicea’ tests before other tests makes a difference to how long they take to run in the Windows test run. In all of the builds, the Windows test run seemed to do nothing at all, without any error message, which I assume is related to issue #15105. I also tried making two small changes to the Jenkinsfile to get some more info, thinking Jenkins would pick those up and apply them after at most one build, but that’s apparently not the case.

jecisc (approved):

PR #16301, opened 2024-03-15
Fix: remove extract method transformation from UI

### LINKED ISSUE #16248: [RB] Extract method is broken
In the latest Pharo 12 I had this method:

```st
applyOn: aModel

	(self queryBlock cull: aModel cull: self) ifNotEmpty: [ :entities |
		| tag |
		tag := aModel allTags
			       detect: [ :t | t name = self tagName ]
			       ifNone: [ aModel createTagNamed: self tagName ].
		entities do: [ :e | e tagWith: tag ] ]
```

And I wanted to extract this part:

```st
	aModel allTags
	       detect: [ :t | t name = self tagName ]
	       ifNone: [ aModel createTagNamed: self tagName ]
```

But I got multiple problems:
- The text field was already filled with "fooForNow" that seems to be leftover code
- My piece of code needs `aModel` as argument but I had no argument in the window and I cannot add one
- Accepting ends up in an error message saying the content cannot be parsed

balsa-sarenac:
This removes extract method transformation for the UI menu. We have refactoring implementation that is exposed and works better than this one. This closes: #16248
Besides this fix, extract method refactoring has new shortcut, that makes it easier to execute.

### DIFF
--- a/src/SystemCommands-SourceCodeCommands/SycExtractMethodAndOccurrencesCommand.class.st
+++ b/src/SystemCommands-SourceCodeCommands/SycExtractMethodAndOccurrencesCommand.class.st
@@ -24,7 +24,7 @@ SycExtractMethodAndOccurrencesCommand class >> canBeExecutedInContext: aSourceCo
 SycExtractMethodAndOccurrencesCommand class >> methodEditorShortcutActivation [
 	<classAnnotation>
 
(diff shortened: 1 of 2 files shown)


jecisc (approved):

jecisc:
I tried it and it fixes my issues indeed

Ducasse:
Tx!

PR #16302, opened 2024-03-15
Take the selector into account in #AdditionalMethodState>>#analogousCodeTo:

### LINKED ISSUE #15967: Inconsistency when comparing compiled methods
**Bug description**

When comparing two compiled methods that have the same code but different selectors, they can either be equal of different depending on whether or not they have pragmas.

**To Reproduce**
This shows the issue. We create first two compiled methods with the same code and different selectors. They are not equal.

```
Object subclass: #FooBar2
	instanceVariableNames: ''
	classVariableNames: ''
	package: 'FooBar'.
	
#FooBar2 asClass compile: 'foo1234
	^ 1'.
#FooBar2 asClass compile: 'bar1234
	^ 1'.

(#FooBar2 asClass >> #foo1234) 
	sameLiteralsAs: (#FooBar2 asClass >> #bar1234) "false"
```

Screenshot 2024-01-16 at 15 20 11

But now we just add the same pragma to the two methods and they become equal.

```
Object subclass: #FooBar
	instanceVariableNames: ''
	classVariableNames: ''
	package: 'FooBar'.
	
#FooBar asClass compile: 'foo123
	<haba>
	^ 1'.
#FooBar asClass compile: 'bar123
	<haba>
	^ 1'.
	
(#FooBar asClass >> #foo123) 
	sameLiteralsAs: (#FooBar asClass >> #bar123) "true"
```

Screenshot 2024-01-16 at 15 22 03

**Expected behavior**
The expected result is for the two compiled methods to be different in both cases. Adding a pragma should not make them equal.

**Version information:**
Tested in Pharo 10, 11 and 12

**Expected development cost**
The fix could be to update `AdditionalMethodState>>#analogousCodeTo:` to also test explicitly for the selector:

```
analogousCodeTo: aMethodProperties
	| bs |
	self class == aMethodProperties class ifFalse:
		[^false].
	(bs := self basicSize) = aMethodProperties basicSize ifFalse:
		[^false].
	1 to: bs do:
		[:i|
		((self basicAt: i) analogousCodeTo: (aMethodProperties basicAt: i)) ifFalse:
			[^false]].

       "--------"
       self selector = aMethodProperties selector 
		ifFalse: [ ^ false ].
       "--------"

	^true
```

If the two compiled methods do not have an additional state, then the selector is used during the comparison. But if the method has `AdditionalMethodState` then `AdditionalMethodState>>#analogousCodeTo:` does not check the selector during the comparison.

chisandrei:
This adds an extra check in AdditionalMethodState>>#analogousCodeTo: to explicitly take the selector into account

### DIFF
--- a/src/Kernel-CodeModel/AdditionalMethodState.class.st
+++ b/src/Kernel-CodeModel/AdditionalMethodState.class.st
@@ -81,6 +81,7 @@ AdditionalMethodState >> analogousCodeTo: aMethodProperties [
 		[:i|
 		((self basicAt: i) analogousCodeTo: (aMethodProperties basicAt: i)) ifFalse:
 			[^false]].
+	self selector = aMethodProperties selector ifFalse: [ ^ false ].
 	^true
 ]
 
--- a/src/Kernel-Extended-Tests/AdditionalMethodStateTest.class.st
+++ b/src/Kernel-Extended-Tests/AdditionalMethodStateTest.class.st
@@ -37,6 +37,16 @@ AdditionalMethodStateTest >> testAnalogousCodeTo [
 ]
 
 { #category : 'tests' }
+AdditionalMethodStateTest >> testAnalogousCodeToTakesSelectorIntoAccount [
+	| pragma anotherAmState |
+	pragma := (Object compiledMethodAt: #at:) penultimateLiteral at: #primitive:.
+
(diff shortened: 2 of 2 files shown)


MarcusDenker (approved):

MarcusDenker:
This is good fix as it is now not consistent (difference if extended method state it present or not)

Issue #15178, opened 2023-10-30
Syntax highlighter could activate only if we stop typing for some time


jecisc:
Currently we are launching the syntax highlighter at each keystroke even if we are typing a full method name or string pretty fast. 

Maybe we could highlight only if we do not type for 200ms or something like this?

I know I implemented something like this in FastTable filters. See FTFilterFunction>>#spawnFilterUpdateThread

MarcusDenker:
There used to be some infrastructure that tried to do that in the past, it tried to do the highlighting in another thread and aborting if it took too long.

The problem is that nothing in the editor is thread safe, so this lead to very very strange effects.

So we need a different solution.

What I wanted to fix first is to parse not *twice* per keystroke... if you put a traceCr in 

RBParser>>#parseMethod

you see that it is called twice.

hernanmd:
I checked and there are two `parse` sends as Marcus wrote:

1. One come from SHRBTextStyler
2. Another one come from Calypso `currentEditedAST` 

And the `currentEditedAST` is used by the IconStyler, which only is needed for styling and AST with highlighting.
I don't know enough the Reflectivity breakpoints, but where is this used?

Because I removed `ClyMethodEditorToolMorph>>textChanged:` and all styling works the same in Calypso even when I put some breakpoints.

hernanmd:
Fixed in #16306

PR #16306, opened 2024-03-18
Fix parsing twice per keystroke in Calypso

hernanmd:
This PR removes an unnecessary send to `currentEditedAST` in `ClyMethodEditorToolMorph` textChanged: so we don't parse twice the current method in Calypso.

To easily check the issue, open the Transcript, go to `RBParser` and compile this method:

```smalltalk
parseMethod
	| methodNode |
	
	(String streamContents: [ : s | DateAndTime now asUTC printHMSOn: s ]) traceCr.
	methodNode := self parseMessagePattern.
	self parsePragmas.
	self addCommentsTo: methodNode.
	self parseMethodBodyInto: methodNode.
	^methodNode
```

This was discovered by @jecisc and @MarcusDenker in #15178


jecisc (approved):
Tried it and it seems to work good! :)

PR #16311, opened 2024-03-19
Protect with a mutex the transcript update

guillep:
The transcript is not entirely thread-safe right now.
When writing in the transcript from another thread, a race condition may happen while drawing the UI and this may cause blockage of the UI.

This PR puts the drawing of the Transcript inside the preexisting mutex to avoid such race conditions.

### DIFF
--- a/src/Transcript-Core/ThreadSafeTranscript.class.st
+++ b/src/Transcript-Core/ThreadSafeTranscript.class.st
@@ -177,7 +177,7 @@ ThreadSafeTranscript >> shoutAboutToStyle: aPluggableShoutMorphOrView [
 
 { #category : 'updating' }
 ThreadSafeTranscript >> stepGlobal [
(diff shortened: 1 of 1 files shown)


guillep:
Just to add more info to the issue, we experienced this with @Ducasse when working on the profiler.
We were performing a transcript log on each event, on the profiler thread that is running on a high(er) priority (vs the UI thread).

guillep:
Moreover, the bug was revealed by the new "async" do-it from the playground toolbar button.
- The profiled code is forked into a low-priority process (30)
- letting the UI thread keep running (at priority 40)
- => The UI runs concurrently with the high-priority (>41) profiler process, causing the issue

Before this did not happen because the profiled code was run inside the UI thread, blocking event handling and rendering.
Note that this does not happen either when you do a do-it through a shortcut (Cmd/Ctrl+D/P) because they are synchronous, which would show another issue (that shortcuts and toolbar do not have the same behaviour).

PR #16336, opened 2024-03-23
Announcer: Return early if there isn't any subscriber

tinchodias:
Doing announcements on an empty Announcer showed a 6X speed-up on a MacBookPro 2018:

"20,785,231 iterations in 5 seconds 2 milliseconds. 4155384.046 per second"
"125,964,386 iterations in 5 seconds 1 millisecond. 25187839.632 per second"

Execute before / after the change:
```smalltalk
announcer := Announcer new.
[announcer announce: Announcement new] bench.  
```

### DIFF
--- a/src/Announcements-Core/Announcer.class.st
+++ b/src/Announcements-Core/Announcer.class.st
@@ -27,16 +27,22 @@ Class {
 Announcer >> announce: anAnnouncement [
 
 	^ self isSuspended
-		  ifFalse: [
-			  | announcement |
-			  announcement := anAnnouncement asAnnouncement.
+		ifFalse: [
+			| announcement |
+			announcement := anAnnouncement asAnnouncement.
 
-			  "If the user required to prevent the current announcement, we stop here."
-			  (self preventedAnnouncements handlesAnnouncement: announcement) ifTrue: [ ^ self ].
+			"Stop if there isn't any subscriber"
+			registry isEmpty ifTrue: [ ^ announcement ].
 
-			  registry ifNotNil: [ registry deliver: announcement ].
-			  announcement ]
-		  ifTrue: [ storedAnnouncements ifNotNil: [ storedAnnouncements add: anAnnouncement ] ]
+			"If the user required to prevent the current announcement, we stop here."
+			(self preventedAnnouncements handlesAnnouncement: announcement)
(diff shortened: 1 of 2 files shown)


tinchodias:
I don't think this will have a great general impact, but it's a small contribution. For example in the fresh image where I pushed the PR:

```
Announcer allSubInstances size. "853"
Announcer allSubInstances count: [ :each | each numberOfSubscriptions = 0  ]. "497"
```

In other with Bloc (should check of there is a leak):
```
Announcer allSubInstances size. "1990"
Announcer allSubInstances count: [ :each | each numberOfSubscriptions = 0  ]. "525"
```

tinchodias:
AndreaSystemProfiler says 90% of time is currently spent in this sentence: `self preventedAnnouncements handlesAnnouncement: announcement`. The isEmpty check I placed before that sentence avoids it.

src/Announcements-Core/Announcer.class.st
```diff
@@ -27,16 +27,22 @@ Class {
...
 
 	^ self isSuspended
-		  ifFalse: [
-			  | announcement |
-			  announcement := anAnnouncement asAnnouncement.
-
-			  "If the user required to prevent the current announcement, we stop here."
-			  (self preventedAnnouncements handlesAnnouncement: announcement) ifTrue: [ ^ self ].
-
-			  registry ifNotNil: [ registry deliver: announcement ].
```
tinchodias, line 37: I removed the ifNotNil: check because it's initialized on creation, it's never nil

src/Announcements-Core/Announcer.class.st
```diff
@@ -27,16 +27,22 @@ Class {
...
-			  announcement ]
-		  ifTrue: [ storedAnnouncements ifNotNil: [ storedAnnouncements add: anAnnouncement ] ]
+		ifFalse: [
+			| announcement |
+			announcement := anAnnouncement asAnnouncement.
```
tinchodias, line 32: Tests were broken if the return was placed before this sentence... they expected that if you announcer a class, it's intanced and you get the instance as return.

tinchodias:
Good idea, an optimization like this is better for P13

MarcusDenker (approved):

Issue #16338, opened 2024-03-23
ZnMimePart>>#hasHeaders breaks in Pharo 12


jbrichau:
**Bug description**
The method `ZnMimePart>>#hasHeaders` was changed in commit https://github.com/pharo-project/pharo/commit/031182ec9ea9912a9d5bed6cf47a633dd07e766a 
to replace `self headers isEmpty not` by `self headers isNotEmpty`.

However, the return value of `self headers` is not a collection but an instance of `ZnHeaders` which does not understand `isNotEmpty`.

**To Reproduce**
Execute the `WAUploadFunctionalTest` in Seaside (https://github.com/SeasideSt/Seaside/tree/master/repository/Seaside-Tests-Functional.package/WAUploadFunctionalTest.class/). 
See https://github.com/SeasideSt/Seaside/actions/runs/8402271471/job/23011499505 for a trace of the crash.

PR #15873, opened 2023-12-29
Integrate Teachable objects into SUnitMock

hernanmd:
As requested in #15806 this PR integrates the Teachable package from @emicklei into SUnitMock package.

PR #16339, opened 2024-03-23
ZnHeaders does not understand isNotEmpty

jbrichau:
Revert change made in commit 031182ec9ea9912a9d5bed6cf47a633dd07e766a because ZnHeaders does not understand #isNotEmpty.

Fixes issue #16338

### DIFF
--- a/src/Zinc-HTTP/ManifestZincHTTP.class.st
+++ b/src/Zinc-HTTP/ManifestZincHTTP.class.st
@@ -1,0 +1,17 @@
+"
+Please describe the package using the class comment of the included manifest class. The manifest class also includes other additional metadata for the package. These meta data are used by other tools such as the SmalllintManifestChecker and the critics Browser
+"
+Class {
+	#name : 'ManifestZincHTTP',
+	#superclass : 'PackageManifest',
+	#category : 'Zinc-HTTP-Manifest',
+	#package : 'Zinc-HTTP',
+	#tag : 'Manifest'
+}
+
+{ #category : 'code-critics' }
+ManifestZincHTTP class >> ruleNotEliminationRuleV1FalsePositive [
+
+	<ignoreForCoverage>
+	^ #(#(#(#RGMethodDefinition #(#ZnMimePart #hasHeaders #false)) #'2024-03-23T16:05:34.582387+01:00') )
+]
--- a/src/Zinc-HTTP/ZnMimePart.class.st
+++ b/src/Zinc-HTTP/ZnMimePart.class.st
@@ -196,7 +196,7 @@ ZnMimePart >> hasEntity [
 { #category : 'testing' }
 ZnMimePart >> hasHeaders [
 
-	^ headers isNotNil and: [ self headers isNotEmpty ]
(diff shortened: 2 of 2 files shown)


jbrichau:
Thanks for the fast merger!

noha:
Are these changes pushed upstream to the zinc repo as well?

jbrichau:
@noha The Zinc repo does not have this issue. It was only present in the forked version of Zinc absorbed into the Pharo project.  (see https://github.com/svenvc/zinc/blob/master/repository/Zinc-HTTP.package/ZnMimePart.class/instance/hasHeaders.st)

The class is even in another package... 
This _is_ going to lead to maintenance issues later on, so I don't understand why the codebase for Zinc in Pharo has been forked&absorbed.


<blockquote><div> GitHub</div><div><strong><a href="https://github.com/svenvc/zinc/blob/master/repository/Zinc-HTTP.package/ZnMimePart.class/instance/hasHeaders.st">zinc/repository/Zinc-HTTP.package/ZnMimePart.class/instance/hasHeaders.st at master · svenvc/zinc</a></strong></div><div>Zinc HTTP Components is an open-source Smalltalk framework to deal with the HTTP networking protocol. - svenvc/zinc</div></blockquote>

astares:
You can blame me if you like as had no time left in November (and this time forgot about the additional PR for Zinc repo).
Typically I care about them too. Mostly for all other changes I've opened PRs for @svenvc so changes are folded back into his repos as wide as possible. Still hard to keep in synch and I do not know when the last resynch between Pharo/Zinc/Zodiac had been done as it always requires effort.

Code handling is still confusing and unsolved for parts of the standard image. The way we currently deal with Pharo source code for some external projects is following different strategies:

As a base
a) typically most of the Pharo source code lives in the standard Pharo repo

but additionally

b) some previously external source code is "insourced" (copied) into the same repo as a copy 
    for the standard Pharo repo. Such a  copy requires a second PR work if you want to bring changes
    back to the original external repo). This leads to more excessive forking and wider distance of code streams
    over time if not done properly

c) some code comes from externally versioned projects (with a tag versioned from time to time included via Baseline loading)

d) some code comes from external versioned projects using a branch per version ("Pharo11" or "Pharo12" that is like 
    an included open edition). So you can commit into an external repo in the right branch and it will be included
    in the next image build.

Handling all differently unfortunately still burns additional cycles and gives headaches to new Pharo joiners as 
well as long time contributors.  

There seem to be different ideas/strategy in the core team and community on how we integrate (external party) 
code and manage projects and which model we choose to integrate. 
Last time I tried to raise this more general point was in January - ending up again more in a dispute than 
an agreement on how to solve this better (if interested check #15873, #15907). 

We can also blame the still not yet fully modular Smalltalk approach as Pharo lacks abstractions/tools above the level
of classes in packages embedded into (git) projects loosely assembled via baselines in different distributed git repos.
(for instance ENVY once had Apps, Editions/Versions, Config maps, ... but was dealing with a single repo only).

Iceberg is really a nice and advanced tool - but it is still not easy to handle the "different repo" and "many dependencies" 
cases with multiple baselines. Copying code around makes situation unfortunately things more worse than better.

PR #16341, opened 2024-03-23
Move archival of ‘*.xml’ and ‘*.fuel’ files in ‘Jenkinsfile’ to the ‘finally’ block

Rinzwind:
This pull request moves the archival of the test report `*.xml` and `*.fuel` files in ‘runTests’ in the ‘Jenkinsfile’ from the ‘try’ block to the ‘finally’ block so that any such files are still archived when the test running step fails or times out, which can help in diagnosing that problem.

The ‘junit’ step is not moved as it would presumably throw an error when there’s an incomplete `*.xml` file (that could be avoided by making HDTestReport name the file with a different extension initially and making it rename the file when completed though).

Issue #16280, opened 2024-03-05
[CI] failing test: ReleaseTest>>testThatThereAreNoSelectorsRemainingThatAreSentButNotImplemented


MarcusDenker:
**Bug description**
This tests is failing: ReleaseTest>>testThatThereAreNoSelectorsRemainingThatAreSentButNotImplemented

```
Error

an OrderedCollection('FFILibrary>>#unixLibraryName') should have been empty
Stacktrace

TestFailure
an OrderedCollection('FFILibrary>>#unixLibraryName') should have been empty
ReleaseTest(TestAsserter)>>assert:description:resumable:
ReleaseTest(TestAsserter)>>assert:description:
ReleaseTest(TestAsserter)>>assertEmpty:
ReleaseTest>>testThatThereAreNoSelectorsRemainingThatAreSentButNotImplemented
ReleaseTest(TestCase)>>performTest
```

hernanmd:
I think `unixModuleName` was removed without rewriting senders? 
https://github.com/pharo-project/pharo/pull/16257/files


<blockquote><img src="https://avatars.githubusercontent.com/u/3044265?s=400&v=4" width="48" align="right"><div><img src="https://github.githubassets.com/favicons/favicon.svg" height="14"> GitHub</div><div><strong><a href="https://github.com/pharo-project/pharo/pull/16257">Minor improvement in TF test class by tinchodias · Pull Request #16257 · pharo-project/pharo</a></strong></div><div>Rename unixModuleName as unix64LibraryName
Superclass definition is:
FFILibrary&gt;&gt;
unixLibraryName
&quot;Kept for backward compatibility.
The system complains that this method is not implement...</div></blockquote>

Issue #16119, opened 2024-02-08
[CI] SocketStreamTest failing on CI


MarcusDenker:
**Bug description**
some tests in SocketStreamTest are failing:

on Mac:

- testFlushOtherEndClosed – MacOSX64.Network.Tests.SocketStreamTest
- testNextPutAllFlushOtherEndClosed – MacOSX64.Network.Tests.SocketStreamTest

on Linux:
-  testFlushLargeMessageOtherEndClosed – Unix64.Network.Tests.SocketStreamTest

Rinzwind:
These tests were added in pull request #15984 which was merged in commit 9a4b8d5cd9a4b193.

That commit was used in builds [1336](https://ci.inria.fr/pharo-ci-jenkins2/job/Test%20pending%20pull%20request%20and%20branch%20Pipeline/job/Pharo12/1336/), [1337](https://ci.inria.fr/pharo-ci-jenkins2/job/Test%20pending%20pull%20request%20and%20branch%20Pipeline/job/Pharo12/1337/) and [1338](https://ci.inria.fr/pharo-ci-jenkins2/job/Test%20pending%20pull%20request%20and%20branch%20Pipeline/job/Pharo12/1338/) of the ‘Pharo12’ Jenkins job. Those all had `#testFlushOtherEndClosed` fail on macOS and `#testFlushLargeMessageOtherEndClosed` on Linux.

In build [1442](https://ci.inria.fr/pharo-ci-jenkins2/job/Test%20pending%20pull%20request%20and%20branch%20Pipeline/job/Pharo12/1442/), `#testFlushOtherEndClosed` and `#testNextPutAllFlushOtherEndClosed` failed on macOS, and `#testFlushLargeMessageOtherEndClosed` failed on Linux.

@daniels220: Could you check what the cause of those failures might be and how that can be fixed?

daniels220:
This is fixed in #16172, which is a redo of a close follow-on to #15984. I meant to remind about getting it merged, but I've been unexpectedly busy. It would be great if it could be merged in time for Pharo 12—like I said, it's a really close follow-on to #15984, that just got roadblocked by a mistake on my part with the Semaphore API and then I never got back around to it. But, I understand if this is outside the scope of what's acceptable after code freeze. In that case I can move over the specific change that fixes the test, which is just adding a wait in an appropriate spot.

@guillep sounded like you were going to look at #16172, but it looks like it fell through the cracks?

PR #16342, opened 2024-03-23
Change #getTempPath:buffer: on WinPlatform to use ‘GetTempPathW’ instead of ‘GetTempPath2W’

Rinzwind:
This pull request changes `#getTempPath:buffer:` on WinPlatform to use ‘GetTempPathW’ instead of ‘GetTempPath2W’ for compatibility with versions of Windows older than ‘Windows 10 Build 20348’ or ‘Windows Server Build 20348’.

See the [‘Requirements’ for ‘GetTempPathW’](https://learn.microsoft.com/en-us/windows/win32/api/fileapi/nf-fileapi-gettemppathw#requirements), the [‘Requirements’ for ‘GetTempPath2W’](https://learn.microsoft.com/en-us/windows/win32/api/fileapi/nf-fileapi-gettemppath2w#requirements) and my [earlier comment in pull request #13777](https://github.com/pharo-project/pharo/pull/13777#issuecomment-2016488515).

The [documentation for ‘GetTempPath2W’](https://learn.microsoft.com/en-us/windows/win32/api/fileapi/nf-fileapi-gettemppath2w) indicates it is equivalent to ‘GetTempPathW’ for non-SYSTEM processes: “When calling this function from a process running as SYSTEM it will return the path C:\Windows\SystemTemp, which is inaccessible to non-SYSTEM processes. For non-SYSTEM processes, GetTempPath2 will behave the same as GetTempPath.” As for how ‘GetTempPath2’ is related to ‘GetTempPath2W’: “The fileapi.h header defines GetTempPath2 as an alias which automatically selects the ANSI or Unicode version of this function based on the definition of the UNICODE preprocessor constant. […] For more information, see [Conventions for Function Prototypes](https://learn.microsoft.com/en-us/windows/win32/intl/conventions-for-function-prototypes).”

### DIFF
--- a/src/System-OSEnvironments-Tests/OSEnvironmentTest.class.st
+++ b/src/System-OSEnvironments-Tests/OSEnvironmentTest.class.st
@@ -109,9 +109,11 @@ OSEnvironmentTest >> testSettingEnvValueDuringRevertsValueAfterDuringBlock [
 	envName := self envNameForTest.
 	self instance setEnv: envName value: 'Before'.
 	expected := 'After'.
-	self instance setEnv: envName value: expected during: [  ].
+	self instance setEnv: envName value: expected during: [
+		actual := self instance at: envName.
+		self assert: actual equals: expected ].
 	actual := self instance at: envName.
-	self assert: actual equals: expected
+	self assert: actual equals: 'Before'
 ]
 
 { #category : 'tests' }
--- a/src/System-OSEnvironments/OSEnvironment.class.st
+++ b/src/System-OSEnvironments/OSEnvironment.class.st
@@ -262,7 +262,7 @@ OSEnvironment >> setEnv: nameString value: valueString during: aBlock [
 	aBlock value ] ensure: [
 		oldValue
 			ifNil: [ self removeKey: nameString ]
-			ifNotNil: [ self setEnv: nameString value: valueString ] ]
+			ifNotNil: [ self setEnv: nameString value: oldValue ] ]
 ]
 
 { #category : 'accessing' }
--- a/src/System-OSEnvironments/Win32Environment.class.st
+++ b/src/System-OSEnvironments/Win32Environment.class.st
@@ -121,7 +121,7 @@ Win32Environment >> keysAndValuesDo: aBlock [
 { #category : 'private' }
 Win32Environment >> removeEnvironmentVariable: nameString [
 
-	 ^ self ffiCall: #( int SetEnvironmentVariableW ( Win32WideString nameString, 0 ) )
+	 ^ self ffiCall: #( int SetEnvironmentVariableW ( Win32WideString nameString, Win32WideString 0 ) )
 ]
 
 { #category : 'accessing' }
--- a/src/System-Platforms-Tests/WinPlatformTest.class.st
+++ b/src/System-Platforms-Tests/WinPlatformTest.class.st
@@ -19,5 +19,7 @@ WinPlatformTest >> testGetTempPathFromTMP [
 	OSEnvironment current at: 'TMP' put: expected.
 	actual := OSPlatform current getTempPath.
 	self assert: actual equals: expected ] ensure: [
-		value ifNotNil: [ OSEnvironment current at: 'TMP' put: value ] ]
+		value
+			ifNil: [ OSEnvironment current removeKey: 'TMP' ]
+			ifNotNil: [ OSEnvironment current at: 'TMP' put: value ] ]
 ]
--- a/src/System-Platforms/WinPlatform.class.st
+++ b/src/System-Platforms/WinPlatform.class.st
@@ -86,7 +86,7 @@ WinPlatform >> getTempPath [
 { #category : 'file paths' }
 WinPlatform >> getTempPath: bufferLength buffer: buffer [
 
-    self ffiCall: #(long GetTempPath2W(long bufferLength,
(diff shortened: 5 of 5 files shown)


Rinzwind:
I converted this to a draft. I expect this to fix the issue of the Windows test run timing out (issue #15104), but having a build in which that still times out after pull request #16341 is merged would allow checking that the test run’s `*.xml` and `*.fuel` files are then archived. It remains to be confirmed whether this actually fixes the timeout issue as for all of the builds so far, ‘pharo-ci-jenkins2-win10’ was used, which doesn’t run the tests at all (issue #15105).

Rinzwind:
In build 5, there no longer was a SymbolNotFoundError in the `#setUp` of various tests in [‘Tests-windows-64-Epicea-Tests-Test.xml’](https://ci.inria.fr/pharo-ci-jenkins2/job/Test%20pending%20pull%20request%20and%20branch%20Pipeline/job/PR-16342/5/artifact/Tests-windows-64-Epicea-Tests-Test.xml/*view*/), but a PrimitiveFailed instead. I added a temporary additional commit to get more detailed stack traces to see what the argument to `#createDirectory:` is: commit f197b1a0dc905f6a. I tried opening [‘Tests-windows-64-Epicea-Tests-Test-EpAnnouncementsTest-testMonitorAnnouncesUpdateWhenDisabled.fuel’](https://ci.inria.fr/pharo-ci-jenkins2/job/Test%20pending%20pull%20request%20and%20branch%20Pipeline/job/PR-16342/5/artifact/Tests-windows-64-Epicea-Tests-Test-EpAnnouncementsTest-testMonitorAnnouncesUpdateWhenDisabled.fuel) in the image of [‘Pharo12.0-PR-64bit-6692750.zip’](https://ci.inria.fr/pharo-ci-jenkins2/job/Test%20pending%20pull%20request%20and%20branch%20Pipeline/job/PR-16342/5/artifact/build/bootstrap-cache/Pharo12.0-PR-64bit-6692750.zip) but that signaled an FLMethodChanged, and a MessageNotUnderstood after changing `#defaultMethodChangedWarningDisabled` to answer `true`.

Rinzwind:
As can be seen in the [‘Tests-windows-64-Epicea-Tests-Test.xml’ of build 9](https://ci.inria.fr/pharo-ci-jenkins2/job/Test%20pending%20pull%20request%20and%20branch%20Pipeline/job/PR-16342/9/artifact/Tests-windows-64-Epicea-Tests-Test.xml/*view*/), the PrimitiveFailed was due to the argument to `#createDirectory:` referring to a nonexistent drive ‘X’:

```
<?xml version="1.0" encoding="UTF-8"?>
<testsuite name="Epicea-Tests" tests="55" timestamp="8:45:43.838 pm" seed="377703460"  failures="0" errors="97" time="985.223">                                                            
	<testcase classname="Windows64.Epicea.Tests.EpAnnouncementsTest" name="testMonitorAnnouncesUpdateWhenDisabled" time="20.034">
		<error type="PrimitiveFailed" message="primitive #createDirectory: in WindowsStore failed">
PrimitiveFailed
primitive #createDirectory: in WindowsStore failed
[…]
WindowsStore(DiskStore)>>createDirectory:
	Receiver: a WindowsStore
	Arguments and temporary variables: 
		path: 	Path / 'X:'
[…]
```

That was in turn due to `#testResolveTempPathFromTMP` on WindowsResolverTest having been executed before which left its test value for ‘TMP’ in the environment. Fixed in: commit 2ab718df0205efb0.

The temporary additional commit to get the more detailed stack trace has been removed. This can be merged now.

Rinzwind:
There are some [failing tests in build 15](https://ci.inria.fr/pharo-ci-jenkins2/job/Test%20pending%20pull%20request%20and%20branch%20Pipeline/job/PR-16342/15/testReport/). See issue #16119 for the SocketStreamTest failures, and issue #16280 for the ReleaseTest failures. The other test failures are not caused by the changes in this pull request, but show up because the Windows test run no longer times out.

jecisc:
Is this ready to merge? @Rinzwind 

@MarcusDenker Do we merge with the failing windows tests? They seems linked to the strict FFI that @guillep enabled. (I think, I'm not sure since I never really used FFI :) )

Rinzwind:
It’s ready to be merged yes.

I opened NewTools pull requests [#728](https://github.com/pharo-spec/NewTools/pull/728) and [#729](https://github.com/pharo-spec/NewTools/pull/729) to fix the [‘Windows64.NewTools.FileBrowser.Tests’](https://ci.inria.fr/pharo-ci-jenkins2/job/Test%20pending%20pull%20request%20and%20branch%20Pipeline/job/PR-16342/15/testReport/Windows64.NewTools.FileBrowser.Tests/) failures.

Rinzwind:
The function signature I used in [NewTools pull request #729](https://github.com/pharo-spec/NewTools/pull/729) was not correct, I opened another pull request to (try to) fix it: [NewTools pull request #730](https://github.com/pharo-spec/NewTools/pull/730).

I’m not sure why using `ulong* nil` doesn’t work though, the FFI should perhaps be extended to support that.

Rinzwind:
All three of the [failing tests in build 21](https://ci.inria.fr/pharo-ci-jenkins2/job/Test%20pending%20pull%20request%20and%20branch%20Pipeline/job/PR-16342/21/testReport/) are in SocketStreamTest (issue #16119).

PR #16343, opened 2024-03-24
fixes #16315 dependency browser layout feels wrong

### LINKED ISSUE #16315: Dependency Browser layout feels wrong
The fields in the toolbar do not have enough space.
It would be nice to do a small pass before the release.

imagen

n1toxyz:
This PR proposes a fix for issue #16315. It changes the UI to provide the widgets with their respective space. For this some of the buttons are moved into an SpActionBarPresenter.

See the screenshot below for the new look.
16315-new-ui

### DIFF
--- a/src/Tool-DependencyAnalyser-UI/DADependencyTreePresenter.class.st
+++ b/src/Tool-DependencyAnalyser-UI/DADependencyTreePresenter.class.st
@@ -31,7 +31,7 @@ DADependencyTreePresenter class >> defaultLayout [
 		add: ( SpBoxLayout newHorizontal
 					add: #textPackageField;
 					add: #buttonBrowseCycles;
-					add: #buttonRefresh;
+					add: #buttonRefresh expand: false;
 					add: #buttonDefault;
 					yourself)
 			expand: false;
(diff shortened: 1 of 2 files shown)


src/Tool-DependencyAnalyser-UI/DADependencyTreePresenter.class.st
```diff
@@ -4,6 +4,7 @@ A PDPackageAnalyzerTreeModel shows all dependent packages from a set of packages
 Class {
 	#name : 'DADependencyTreePresenter',
 	#superclass : 'DAPackageTreePresenter',
+	#classTraits : '{} + TraitedClass',
```
jecisc: I think there is a problem with the Tonel export here. 

I've seen messages about this but I did not follow what causes this and what should be done when this happens.

n1toxyz: I see... I didn't notice it. Iceberg didn't show it either. I was using a fresh image when working on the PR. Is there something I can do to fix this?

Ducasse:
Argh again this bug!
Sorry for the inconvenience. 
We often edit the file by end. We are looking for a way to reproduce it.

n1toxyz:
Is it possible for me to edit it by hand or do I need to do a new PR? Do I just have to delete the extra line?

Ducasse:
I imagine just remove the extra line and the preceeding ,

jecisc:
I edited the file

PR #16347, opened 2024-03-25
Make class abstract driver

hernanmd:
This PR implements the "Make Abstract" class refactoring using the new Driver interface.

Make RBAbstractClass match the MOP answering both false by default in #isAbstract.
Migrate RBMakeClassAbstractTransformation.
Add data for testing: #RBClassWithoutSelfClassReferenceTest and #RBWithSelfClassReferenceTest
Update RBMakeClassAbstractParametrizedTest
Add #RBMakeClassAbstractDriver with tests
Update #SycCMakeAbstractCommand


Ducasse:
What is the logic of this refactoring? I mean I do not really understand what is an abstract class in pharo and when i will need it.

hernanmd:
I migrated because it was there and some people may be using it.

However, I never bought the arguments for "improving code clarity" or "preventing improper instantiation" (maybe they're good for educational purposes), so I'll provide an example where I used it some time ago as a dirty experiment: To migrate a legacy system, using automated scripts, some classes couldn't be migrated to the new system. To get a quick diff of which ones weren't migrated, I annotated the already migrated abstract classes with the "isAbstract" method, so I could later see which hierarchies have fewer subclasses than expected.

If there are other use cases I would have to think about them because as you said, there doesn't seem to be too many.

MarcusDenker:
I use it a lot, it is useful for the Browser and the Rules to know that a class is abstract. Imagine having a subclassResponsibility method. The browser nags you to implement it, but the subclass is Abtract, too. So we could add the method again, then it complaints that the method is the same as in the superclass. 

=> add #isAbstract, and the Browser/Rules know that you do not have to implement the #subclassResponsability here and just ask you to implement it where you want it

jecisc:
I also like to know easily when a class is abstract when I'm exploring a project I'm not familiar with. Like that I'm able to understand the structure of the project more easily.

And I know that I should not instantiate those classes but just look at the subclasses.

src/Refactoring-Transformations-Tests/RBAbstractRefactoringTest.class.st
```diff
@@ -9,6 +9,22 @@ Class {
...
+	| env |
+	env := (RBNamespace onEnvironment: (RBPackageEnvironment packageName: 'Refactoring-DataForTesting')).
+	^ env classNamed: self unreferencedClassName 
+]
+
+{ #category : 'helpers' }
+RBAbstractRefactoringTest class >> unreferencedClassName [
+	"Answer a <Symbol> representing a <Class> which should not have any references"
+
+	^ #[82 66 67 108 97 115 115 87 105 116 104 111 117 116 83 101 108 102 67 108 97 115 115 82 101 102 101 114 101 110 99 101 84 101 115 116] asString asSymbol
```
balsa-sarenac: Suggestion: You can also do this with `'Unreferenced', 'ClassName' asSymbol`, that version is a bit more readable then this one since it doesn't require interpretation.

src/Refactoring-UI-Tests/ReMakeClassAbstractDriverTest.class.st
```diff
@@ -0,0 +1,90 @@
...
+	self 
+		deny: rbClass isAbstract
+		description: 'It tests that the RB class is not abstract because it contains its class reference'.
+		
+	self setUpDriver: driver.
+	driver runRefactoring.
+	
+	self 
+		deny: rbClass isAbstract
+		description: 'It tests that the RB class is still abstract after applying the refactoring'.
```
balsa-sarenac: Since this is deny, the description seems wrong, I think it should then say:

```suggestion
	self 
		deny: rbClass isAbstract
		description: 'It tests that the RB class is not abstract after applying the refactoring'.
```

src/Refactoring-UI/RBMakeClassAbstractDriver.class.st
```diff
@@ -0,0 +1,46 @@
...
+		yourself.
+]
+
+{ #category : 'execution' }
+RBMakeClassAbstractDriver >> runRefactoring [
+	"Run the transformation"
+
+	self configureRefactoring.
+	refactoring failedApplicabilityPreconditions
+		ifNotEmpty: [ ^ self inform: 'Preconditions not met' ].
```
balsa-sarenac: Can we here tell the user what exactly went wrong? We know there is only one precontiion, and `refactoring failedApplicabilityPreconditions` returns a list with only that precondition. I'm suggesting that we do something like

```suggestion
	preconditions := refactoring failedApplicabilityPreconditions
	preconditions
		ifNotEmpty: [ ^ self inform: preconditions first errorMessage ].
```
Maybe the code needs to be modified a bit, but you get the gist of it.

balsa-sarenac (approved):

jecisc:
CI got a random failure. I'm restarting it

hernanmd:
Thanks Cyril, it seems that failing tests are not related:

Screenshot 2024-03-28 at 16 37 42

jecisc:
Some tests are refactoring related so it's possible it comes from here no?

hernanmd:
> Some tests are refactoring related so it's possible it comes from here no?

Ok, I fixed the failing tests. Now it seems good to go

https://ci.inria.fr/pharo-ci-jenkins2/blue/organizations/jenkins/Test%20pending%20pull%20request%20and%20branch%20Pipeline/detail/PR-16347/6/tests


<blockquote><div><strong><a href="https://ci.inria.fr/pharo-ci-jenkins2/blue/organizations/jenkins/Test%20pending%20pull%20request%20and%20branch%20Pipeline/detail/PR-16347/6/tests">Jenkins</a></strong></div></blockquote>

balsa-sarenac (approved):

PR #16352, opened 2024-03-28
remaining morphic dialogs: aligning the default option to right

estebanlm:
This is to finish the compatibility adjustment of dialog options: all positive options are default and aligned as last button option. This may be just a setting in the future: once we finish moving all dialogs to spec :)

Still uncertain if the positive (which in some cases is destructive) option needs to be there, that will cause some problems to people (thinking on muscular memory) :/


jecisc (approved):

Ducasse:
Cool!!!!!!

PR #16353, opened 2024-03-28
Add subclass driver

hernanmd:
This PR implements a first version of "Add subclass" refactoring as a driver:

https://github.com/pharo-project/pharo/assets/4825959/663878b5-19e6-4c7b-a0c8-bf097d2bab4a

## Main features

- The Spec UI to add a new subclass (see video above).
- It checks and informs interactively if the class already exists.
- It checks invalid class names.
- Default class comment could be customized implementing `blankCommentTemplate` in the class selected as superclass.

## Implementation notes

- A `RBAddNewClassRefactoring` which differs from the **Insert New Class** refactoring in that does not reparent the subclasses to the new class.
- The driver implementation in `ReAddSubclassDriver` and `ReAddSubclassDriverTest`.

## Known issues

- `CmdCommandAborted signal` and `CmCommandAborted signal` are failing right now and it should be investigated why (debugger keep opening which could mean the exception is not properly handled).
- Next version will include more tests.


src/Refactoring-UI/StRefactoringAddClassPresenter.class.st
```diff
@@ -0,0 +1,234 @@
+"
+It provides a request dialog to add a new class in the system, including presenters to specify its name, package tag and comment.
+
+"
+Class {
```
balsa-sarenac, line 5: `textInput` is not used from superclass, pay attention

src/Refactoring-Core/RBAddNewClassRefactoring.class.st
```diff
@@ -0,0 +1,34 @@
+"
+This refactorings differs from the Insert New Class refactoring in that does not reparent the subclasses to the new class.
+"
+Class {
+	#name : 'RBAddNewClassRefactoring',
+	#superclass : 'RBInsertNewClassRefactoring',
+	#category : 'Refactoring-Core-Refactorings',
+	#package : 'Refactoring-Core',
+	#tag : 'Refactorings'
+}
```
balsa-sarenac: We can reuse `RBInsertNewClassRefactoring` without setting the subclasses like this:

```
	(RBInsertNewClassRefactoring className: newClassName)
		superclass: targetClass asString;
		packageName: targetClass package name;
		tagName: targetClass packageTag name;
		execute
```

src/SystemCommands-ClassCommands/SycCmAddSubclassCommand.class.st
```diff
@@ -34,7 +60,78 @@ SycCmAddSubclassCommand >> name [
...
+	
+	refactoringScopes := context refactoringScopes.
+	targetClass := context lastSelectedClass.
+	(StRefactoringAddClassPresenter on: self)
+		onCancel: [ : dialog |
+			"CmdCommandAborted signal"
+			cancelled := true.
+			dialog close ];
+		onAccept: [ :dialog | dialog close ];
+		openModal
```
balsa-sarenac: Ideally, this should be moved to the driver as well

src/Refactoring-UI/ReAddSubclassDriver.class.st
```diff
@@ -0,0 +1,113 @@
...
+	comment := aString
+]
+
+{ #category : 'accessing' }
+ReAddSubclassDriver >> configureRefactoring [
+
+	refactoring := (RBAddNewClassRefactoring className: self subclass)
+		model: model;
+		superclass: self superclass;
+		subclasses: self superclass subclasses;
```
balsa-sarenac: We can omit this line

balsa-sarenac (changes requested):
Great work! I really like the UI for adding new subclass! Left some comments inline, based on our discussion

src/Refactoring-Core/RBAddNewClassRefactoring.class.st
hernanmd: The purpose of AddNewClassRefactoring is that does not reparent the subclass(es) to the new class.

I understand here you want to avoid having a new class? 

In this sense I'm not sure it is better "to be clever" (it is always a mess to check which message should be avoided to have one feature active or not. I would prefer to have two explicit refactoring classes: One for adding a class with reparent (Insert), and another one without reparent (just Add).

src/Refactoring-UI/StRefactoringAddClassPresenter.class.st
hernanmd: Good catch, I re-used the variable from superclass.

src/SystemCommands-ClassCommands/SycCmAddSubclassCommand.class.st
hernanmd: I'm not sure how to do it. Maybe for a next iteration? (I'm trying to not make this PR bigger)

Ducasse: We should not have any logic in the command. This should be moved to the driver.

Ducasse: Just move it there. 
The length of the PR is only important once we are feature complete :)

src/SystemCommands-ClassCommands/SycCmAddSubclassCommand.class.st
```diff
@@ -5,50 +5,33 @@ Class {
...
-		tagName: self tagName;
-		comment: self comment) runRefactoring.
+	refactoring := (ReAddSubclassDriver 
+		superclass: self targetClass 
+		packageNames: self packageNames)
+			scopes: refactoringScopes.
+	refactoring subclass
+		ifNil: [ ^ self ]. " cancelled "
+	refactoring runRefactoring.
+	newClassName := refactoring subclass.
```
balsa-sarenac: A suggestion: temp `refactoring` might be called `driver`

balsa-sarenac (approved):

src/Refactoring-Core/RBAddNewClassRefactoring.class.st
hernanmd: I will make it as resolved for now, but feel free to open a new issue or we could check it later if it makes sense to merge both classes.

balsa-sarenac (approved):

src/Refactoring-Core/RBAddNewClassRefactoring.class.st
balsa-sarenac: Let's remove this class, since we have InsertNewClass, let's stay on track and follow the API we have so far

balsa-sarenac (changes requested):

MarcusDenker:
Two tests failing due to:

- the following classes have unused instance variables and should be cleaned: SycCmAddSubclassCommand
- the following classes have uncategorized methods: an OrderedCollection(RBInteractionDriver)

Ducasse:
@hernanmd can you take the point of balsa into account so that we can integrate this issue.

hernanmd:
> @hernanmd can you take the point of balsa into account so that we can integrate this issue.

Done

Issue #16373, opened 2024-03-29
Socket>>#waitForSendDoneFor: does not return a boolean (but ZdcSocketStream expects it to)


jbrichau:
**Bug description**
The comment in `Socket>>#waitForSendDoneFor:` says it returns a boolean but it no longer does after changes in https://github.com/pharo-project/pharo/pull/15984/ and https://github.com/pharo-project/pharo/issues/15975

This breaks Zinc, as it still has methods such as `ZdcAbstractSocketStream>>#socketWaitForSendDone` that expect `Socket>>#waitForSendDoneFor:` to still return a boolean.

**To Reproduce**
On a Mac machine only, because it seems Zinc is not using `ZdcSocketStream` on other platforms!
1. Load Seaside (see https://github.com/SeasideSt/Seaside/)
2. Go to `localhost:8080` in a web browser
3. See the error `NonBooleanReceiver` in `ZdcSimpleSocketStream>>#flushBytes:startingAt:count:`
4. Closing the debugger correctly finishes the request, but it pops up on every next request

jbrichau:
The commit that changes the code is https://github.com/pharo-project/pharo/commit/7f99ec49a102f0a60eb8fabbbca7ea9d1cc9d133 
Reverting the change in that commit resolves the error.

Alternatively, `ZdcSimpleSocketStream>>#flushBytes:startingAt:count:` in Zodiac can be changed by removing the `ifFalse:`check since a `ConnectionTimeout` is already signaled by `Socket>>#waitForSendDoneFor:`

jbrichau:
The PR on Zodiac that fixes this issue is https://github.com/svenvc/zodiac/pull/15
However, it is not being merged, probably because it would need to take different Pharo versions into account.

daniels220:
Just in case, @jbrichau did you see my comment on the Zodiac PR?

jbrichau:
@daniels220 Yes, and I think your suggestion is the correct one.

estebanlm:
I just submitted a fix, this is not really a problem to answer  a boolean and it will simplify people's life

daniels220:
[...]

PR #15984, opened 2024-01-18
Improved error handling around sending data in Socket/SocketStream (fixes #15614)

daniels220:
Distinguish timeout and connection-closed, and actually pass errors in SocketStream>>flush rather than always swallowing them.

I believe this addresses the core issues in #15614. Adding a `SocketStream>>waitForSendDone` or similar could be a separate issue if desired.

PR #16376, opened 2024-03-29
Revert comment change from commit f5df5966bd73d05bc04806d2f8735baa4b1783cb

jbrichau:
The method comment is incorrect. This was caused by an (inadvertent?) revert in commit f5df5966bd73d05bc04806d2f8735baa4b1783cb

Context: https://github.com/pharo-project/pharo/issues/16373

The issue itself is not fixed (probably needs to be fixed upstream in Zodiac to take the changes of Socket in Pharo12 into account)

### DIFF
--- a/src/Network-Kernel/Socket.class.st
+++ b/src/Network-Kernel/Socket.class.st
@@ -1677,7 +1677,8 @@ Socket >> waitForDisconnectionFor: timeout [
 
 { #category : 'waiting' }
 Socket >> waitForSendDoneFor: timeout [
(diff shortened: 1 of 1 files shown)


MarcusDenker (approved):

daniels220:
FWIW I think it's not an inadvertent revert, so much as me neglecting to update the comment in #15984, then realizing when working on #16073 and updating it there. Then we ran into the `ThreadedFFI` dependency issue and @guillep reverted the PR, putting things back in the "neglected to update the comment state". But the new comment is already correct at this point, so this is the right thing to do, yes.

Issue #16377, opened 2024-03-29
Make new tools easily accessible


hernanmd:
New tools in Pharo 12 such as File System Browser, Settings Browser and Finder are not accessible from the world menu. 

**Expected behavior**
Add a new menu entry under System menu so we can launch these new tools making more easily accessible for users.

**Expected development cost**
This is no more than a method to group menu entries. New tools should specify this menu entry as parent:

demarey:
I'm not sure we should expose these tools.
We decided to do not make them the default version for P12 because they are not fully ready.
I think it is better to make them the default tools early in P13.

hernanmd:
Two different things.

I understand that the new tools are not configured by default in P12, but simply hiding them from all menus is a lousy way to show progress. In fact, I think it might even be taken the wrong way as it would be deliberately hiding the fact that there are improved versions of the old Squeak-era tools.

I think Pharo users are able to understand when a toolset is in an experimental state.

Ducasse:
We could add (preview) in the menu entry.

jecisc:
PRs got merged.

PR #16379, opened 2024-03-29
New tools parent menu entry

hernanmd:
This PR just add a single method to have a new menu entry item to the System menu so we can group and make new tools easily accessible specifying parent: #NewToolsMenu, as specified in #16377


demarey:
As I explained in the issue, I would prefer not to expose in P12. It would add confusion.
I would prefer to make them the default tools early in P13.

hernanmd:
It would be much better not to disrupt users by assuming that they will get confused, and let them decide for themselves which tools to use. Maybe we could add an description in the menu, something like "experimental" label.

But hiding the tools will only lead to less bugs being reported as they are less exposed.

Ducasse:
We can add 'In preview' as menu item

hernanmd:
> We can add 'In preview' as menu item

Done

demarey (approved):

Issue #16374, opened 2024-03-29
Putting a breakpoint into an unsaved method should force saving it 


LucFabresse:
**Describe the request**
When a breakpoint is put into an unsaved method:
- the method remains unsaved
- the breakpoint works
- it  incorrectly put it into the old method AST
- it does not show up in the current browser

**Expected behavior**
- putting a breakpoint should trigger the  method save

**Screenshots**

**Version information:**
 - OS: All
 - Pharo Version 12

LucFabresse:
fixed in https://github.com/pharo-project/pharo/pull/16382

PR #16382, opened 2024-03-29
quick fix that hide the debug context menus when there are unsaved changes

LucFabresse:


LucFabresse:
[16374](https://github.com/pharo-project/pharo/issues/16374)

demarey (approved):

demarey:
Than you for the fix @LucFabresse

Issue #16313, opened 2024-03-19
Wrong focus in Calypso Find Class Dialog causes exception


guillep:
**Bug description**
 - Open Calypso
 - Right-click in package panel
 - Select "Find class"
 - Type Cmd+W
 => This wrongly closes calypso instead of the find class dialog

Moreover,
 => closing the dialog throws exception sometimes

demarey:
fixed in #16386

guillep:
The attempted fix in https://github.com/pharo-project/pharo/pull/16492 was actually causing other issues such as https://github.com/pharo-project/pharo/issues/16436.

We should properly fix the focus

guillep:
Ok, we chased this issue up to the `#ClyCloseBrowserCommand` that is bound to Cmd+W.
The modal pop up to find classes calls that command that in turn does:

```
browser close
```

That is made to close a tab in a group window, or the (browser)window itself, but it has no knowledge about pop ups that are open by it.

```smalltalk
ClyFullBrowserMorph >> close
	| currentWindow |
	currentWindow := self window.

	(currentWindow ownerThatIsA: GroupWindowMorph) ifNotNil: [:group |
		^self okToChange ifTrue: [
			[group removeWindow: currentWindow]
				on: ProvideAnswerNotification do: [:notification |
					"We already got answer from okToChange check above"
					notification resume: true]]].

	currentWindow close
```

PR #16386, opened 2024-03-30
Fix for Wrong focus in Calypso Find Class Dialog causes exception

hernanmd:
This PR includes a fix for the bug reported in #16313.

The problem was the `WindowShortcuts` keymap category was missing from FTTableMorph during shortcut initialization, this bypassed the keydown event to the next handler, causing the browser window to be closed and subsequent problems in keyboard focus.


jecisc (approved):

demarey (approved):

hernanmd:
Thanks @demarey !

guillep:
Finally with @melkiyasser we propose a different fix: https://github.com/pharo-project/pharo/pull/16493

PR #16392, opened 2024-04-02
16356-Rename-class-refactoring-popup-does-not-have-default-button

### LINKED ISSUE #16356: Rename class refactoring popup does not have default button
When renaming a class, the pop up does not react to the <enter> key. It would be nice if it did.

imagen

I wanted to mark it as "First impression counts" but the tag is not there anymore :(

tesonep:
Improving the request dialog when renaming a class
Fix #16356


demarey:
The change is nice but I do not see the relation with the default button

tesonep:
The default button has been fixed in the PR in Spec.

PR #16400, opened 2024-04-03
Provide a default String for settings without #default: sends.

hernanmd:
This small fix prevents raising an error when a setting does not provide a `default:`. The default value for this case is an empty String, which will provide an input text.

### DIFF
--- a/src/System-Settings-Core/SettingDeclaration.class.st
+++ b/src/System-Settings-Core/SettingDeclaration.class.st
@@ -117,6 +117,7 @@ SettingDeclaration >> addToList: anItem [
 { #category : 'accessing' }
 SettingDeclaration >> default [
 	^ default
+		ifNil: [ default := String empty ]
 ]
 
 { #category : 'accessing' }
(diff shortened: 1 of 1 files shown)


src/System-Settings-Core/SettingDeclaration.class.st
```diff
@@ -181,7 +182,7 @@ SettingDeclaration >> ghostHelp: aString [
 	ghostHelp := aString
 ]
 
-{ #category : 'testing' }
+{ #category : 'accessing' }
```
jecisc: I think testing was a better protocol here

hernanmd: This is weird because I see it already in 'testing' category:

Screenshot 2024-04-04 at 20 36 06

jecisc:
```suggestion
{ #category : 'testing' }
```

Weird indeed. We can just edit the file

jecisc (approved):

hernanmd:
Thanks @jecisc

PR #16401, opened 2024-04-03
Add support for Epicea file out

hernanmd:
This PR: 

- Add support for Epicea file out of class name changes.
- Add support for Epicea file out of class comments.
- Add tests.

### DIFF
--- a/src/Epicea/EpBehaviorCommentChange.class.st
+++ b/src/Epicea/EpBehaviorCommentChange.class.st
@@ -58,7 +58,7 @@ EpBehaviorCommentChange >> behaviorAffectedName [
 EpBehaviorCommentChange >> fileOutEntryOn: aWriteStream [ 
 
 	aWriteStream
-		nextPutAll: '(Smalltalk globals at: #';
+		nextPutAll: '(self class environment at: #';
 		nextPutAll: self behaviorAffectedName;
 		nextPutAll: ' comment: ';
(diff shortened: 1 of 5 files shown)


jecisc (commented):
I am not sure about the hardcoding of Smalltalk here.

For example we have a PR pending to rename Smalltalk into Pharo. And it's not nice to have a global variable.

Maybe `self class environment` is a little better. I don't know if we can do even better right now

hernanmd:
> I am not sure about the hardcoding of Smalltalk here.
> 
> For example we have a PR pending to rename Smalltalk into Pharo. And it's not nice to have a global variable.
> 
> Maybe `self class environment` is a little better. I don't know if we can do even better right now

Ok, changed then

src/Epicea/EpBehaviorCommentChange.class.st
```diff
@@ -54,6 +54,17 @@ EpBehaviorCommentChange >> behaviorAffectedName [
 	^ behavior name
 ]
 
+{ #category : 'accessing' }
+EpBehaviorCommentChange >> fileOutEntryOn: aWriteStream [ 
+
+	aWriteStream
+		nextPutAll: '(self class environment at: #';
```
jecisc, line 61: Aren't you missing the closing parenthesis here? And same for the other file out

jecisc (approved):

PR #16409, opened 2024-04-05
Do not suggest trait when deprecating a class

### LINKED ISSUE #16368: Deprecate class with trait DNU
Select a class
Deprecate menu option
In the popup to choose the "replacement" select a trait

This will show a list of changes.
Accept them => DNU

jecisc:
If we deprecate a class and we select a trait to replace the references, a bug is produced. While it might make sense in some cases to replace a class by a trait, it's something that we do not do often so I propose to just remove the traits from the suggestions for this refactoring.

The alternative would be to update Ring to finish the implementation of the V2 of traits because we are currently in a bastard state of having both V1 and V2 representation but inly using V1 in the system. But that's a bigger change that requires a lot more time.

Fixes #16368

### DIFF
--- a/src/Refactoring-UI/RBDeprecateClassDriver.class.st
+++ b/src/Refactoring-UI/RBDeprecateClassDriver.class.st
@@ -80,12 +80,10 @@ RBDeprecateClassDriver >> requestClassToMigrateReferencesTo [
 
 	newClassName := SpSelectDialog new
 		                title: 'Class to migrate ' , className;
-		                label:
-			                'The new class that will be used to migrate all references of the old class to it.';
-		                items: Smalltalk allClassesAndTraits;
+		                label: 'The new class that will be used to migrate all references of the old class to it.';
+		                items: self class environment allClasses; "This should probably check in the environment of the class to deprecate but I don't know how to get it here"
 		                display: [ :each | each name ];
-		                displayIcon: [ :each |
-			                self iconNamed: each systemIconName ];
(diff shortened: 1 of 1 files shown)


MarcusDenker (approved):

src/Refactoring-UI/RBDeprecateClassDriver.class.st
```diff
@@ -80,12 +80,10 @@ RBDeprecateClassDriver >> requestClassToMigrateReferencesTo [
 
 	newClassName := SpSelectDialog new
 		                title: 'Class to migrate ' , className;
-		                label:
-			                'The new class that will be used to migrate all references of the old class to it.';
-		                items: Smalltalk allClassesAndTraits;
+		                label: 'The new class that will be used to migrate all references of the old class to it.';
+		                items: self class environment allClasses; "This should probably check in the environment of the class to deprecate but I don't know how to get it here"
```
demarey, line 84: You can get it with `RBDeprecateClassRefactoring` instance that inherits `#model` from `RBAbsytractTransformation`.
As `RBDeprecateClassRefactoring` is instanciated via new, you will get the defaul model that is: `RBNamespace onEnvironment: RBBrowserEnvironment new`.
So in the end, I'm not sure there is value to get it. But in the case you want to, just instantiate `RBDeprecateClassRefactoring` before running `requestClassToMigrateReferencesTo:` in `requestClassToMigrateReferencesTo`

PR #16411, opened 2024-04-05
Fix Abstract instance variables command

### LINKED ISSUE #16362: Abstract instance variables DNU
- On a class with slots
- abstract instance variables and OK

imagen

jecisc:
Fixes #16362

### DIFF
--- a/src/SystemCommands-ClassCommands/SycAbstractAllInstVarAccessorsCommand.class.st
+++ b/src/SystemCommands-ClassCommands/SycAbstractAllInstVarAccessorsCommand.class.st
@@ -54,6 +54,6 @@ SycAbstractAllInstVarAccessorsCommand >> selectVariables [
 			withItems: vars
 			selecting: vars.
 		dialog cancelled ifTrue: [ CmdCommandAborted signal ].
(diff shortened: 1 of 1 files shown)


astares:
@jecisc Stupid question: what is the difference between "Abstract instance variable" and "Generate accessors" menu

Both open a (different) dialog, let you choose ivar's and generate accessor methods (while the  "Generate accessors" had an instance preview of the refactoring). Do I miss something obvious?

jecisc:
> @jecisc Stupid question: what is the difference between "Abstract instance variable" and "Generate accessors" menu
> 
> Both open a (different) dialog, let you choose ivar's and generate accessor methods (while the "Generate accessors" had an instance preview of the refactoring). Do I miss something obvious?

To be honest I don't know. 

I opened an issue saying we should add descriptions to refactorings because I don't know what this one does and there is no description.

Ducasse:
Generate an accessor is not a refactoring
just add the methods if needed.
Abstract variable is doing generate accessor + change all the direct accesses to use the accessors.

jecisc:
Someone did the same change as me during the last sprint and it got integrated so I'll close

PR #16422, opened 2024-04-07
Update the Playground shortcut

### LINKED ISSUE #3393: Ctrl + O + W is not a good idea to open the workspace.
As O and W are quite far on the keyboard and CTRL + W is mapped on the window closing, it may happen that the strokes on O and W keys are not fast enough that will result to the closing of the current window. Which is not a good thing..

I will prefer to have something like CTRL + SHIFT + W (or P as it is call Playground). It will avoid further issues...

Let me now if you share my idea!

jecisc:
This change updates the open playground shortcut. Until now it was cmd + o + w because the tool was called workspace in the past. We could not use the p because this was used by the monticello browser. But in Pharo 12 we removed the Monticello browser and we can now have a better shortcut cmd + o + p (Command Open Playground).

I also removed some references to the workspace to unify the image around the term "playground".

Fixes #3393

### DIFF
--- a/src/Deprecated12/ToolShortcutsCategory.extension.st
+++ b/src/Deprecated12/ToolShortcutsCategory.extension.st
@@ -3,6 +3,6 @@ Extension { #name : 'ToolShortcutsCategory' }
 { #category : '*Deprecated12' }
 ToolShortcutsCategory >> openWorkspace [
 
-	self deprecated: 'Use #openPlayground' transformWith: '`@rcv openWorkspace' -> '`@rcv openPlayground'.
-	^ self openPlayground
+	<shortcut>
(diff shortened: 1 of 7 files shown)


jecisc:
This should be merge just after this one: https://github.com/pharo-spec/NewTools/pull/738

tesonep:
@jecisc can we keep both for now? I am thinking in a kind of deprecation

jecisc:
I tried to have 2 shortcuts. I'll see in the produced image if this works

guillep:
Thanks! :)

jecisc:
It works fine :)

MarcusDenker (approved):

PR #16434, opened 2024-04-10
Clean package generation of "Jump to test class"

### LINKED ISSUE #16433: "Jump to test class" is not properly handling tags when creating a test class
In latest Pharo 12:
- using Calypso create a package "MyKillerApp-Core" and within a tag "Model" create a simple class User

image

- Now right click on the class and select *Jump to test class*

image

Now Pharo wrongly generates two packages 

- MyKillerApp-Core-Tests
- MyKillerApp-Core-Tests-Model 

 and puts the generated SUnit test class *UserTest* into the package MyKillerApp-Core-Tests-Model (without a tag)

image

Correct and expected behavior would be to generate only a single package 

- MyKillerApp-Core-Tests

and (if the class to be tested is grouped with a tag) puts the generated SUnit test class *UserTest* into the same tag (here "Model") then.

image

I think this is a bug and might be related to the package / tag work in P12 from @jecisc

jecisc:
Now that packages and tags are well separated we should differenciate them in the test class generation. 

Fixes #16433

### DIFF
--- a/src/Calypso-SystemPlugins-SUnit-Browser/TClyGenerateTestClass.trait.st
+++ b/src/Calypso-SystemPlugins-SUnit-Browser/TClyGenerateTestClass.trait.st
@@ -35,21 +35,6 @@ TClyGenerateTestClass >> isValidClass: inputClass [
 	^ (inputClass isTestCase or: [ inputClass isMeta ]) not
 ]
 
-{ #category : 'action' }
-TClyGenerateTestClass >> newTestClassCategoryFor: aClass [
-
-	| tag |
-	tag := aClass packageTag.
-	^ String streamContents: [ :s |
-		  s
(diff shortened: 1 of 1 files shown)


MarcusDenker (approved):

MarcusDenker:
Full image load fails with

```
NewUndeclaredWarning: SpPresenterSelectorPresenter>>inputModelPort (SpModelPort is Undeclared)
MetacelloNotification: Loaded -> Spec2-Core --- github://pharo-spec/Spec:Pharo12/src(aab7b6b)
Error: The following definitions had errors while loading.  Press Proceed to try to load them again (they may work on a second pass):
  SpRequestDialog
  SpRequestTextDialog


```

estebanlm:
seems to be a lateral problem, I am checking that now.

PR #16439, opened 2024-04-11
Fix: make class abstract was not generating a change

### LINKED ISSUE #16346: [RB] RBAddMethodTransformation does not add changes to the model
**Bug description**
In `RBAddMethodTransformation>>privateTransform` a new method is compiled but the change is not recorded into the model, making `performChanges` later (for example sending `execute` to the transformation) to not find any changes.

**Expected behavior**

This is a possible fix

```smalltalk
privateTransform

	self model changes addChange: (
		self definingClass
			compile: sourceCode
			classified: protocol)
```

balsa-sarenac:
Closes: #16346


hernanmd:
Hi Balsa, I tried this one but it wasn't including the changes in https://github.com/pharo-project/pharo/pull/16347 so I committed it there

PR #16444, opened 2024-04-11
Fix: rename method loop

balsa-sarenac:
Rename method was looping inform dialog when preconditions fail in refactoring, this fixes it.


jecisc:
I see a lot of failing tests but they seems weird when I see the PR.. Maybe we should merge Pharo12 into your branch?

balsa-sarenac:
Indeed it was weird, I've merged now, we'll see how it goes.

balsa-sarenac:
Failing tests seem unrelated now

jecisc (approved):

PR #16449, opened 2024-04-12
Fix #visitBehaviorRepackagedChange: on EpHasImpactVisitor to send #not instead of #y

Rinzwind:
This pull request fixes `#visitBehaviorRepackagedChange:` on EpHasImpactVisitor to send `#not` instead of `#y`.


jecisc (commented):
This was probably introduced by me :o 
Thanks

PR #16461, opened 2024-04-15
Added the compare methods to all the byte array classes so they use the new primitiveCompareBytes

doste:
Now the classes ByteArray, DoubleByteArray, DoubleWordArray and ByteString uses the new primitiveCompareBytes when comparing them. 
 This new primitive was implemented but was never used. Now it is. This would result in a better performance (cf https://github.com/pharo-project/pharo-vm/pull/759).

The comparison methods added have the fallback to use the comparison mechanism as always in case the VM used doesn't support the primitive.

### DIFF
--- a/src/Collections-Native/ByteArray.class.st
+++ b/src/Collections-Native/ByteArray.class.st
@@ -44,6 +44,13 @@ ByteArray class >> readHexFrom: aString [
 	^(self new: aString size // 2) readHexFrom: aString readStream
 ]
 
+{ #category : 'comparing' }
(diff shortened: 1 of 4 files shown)


guillep:
oh so coooool.

Extracted from the VM PR:

> - is 3x faster than doing it directly in Pharo when the compared bytes are few.
> - is hundreds of times faster when comparing thousands of bytes.

jecisc:
The tests are passing. I'll merge so that we can test a little before the release

PR #16466, opened 2024-04-16
16355-Wrong-class-definition-creates-strange-DNUs

### LINKED ISSUE #16355: Wrong class definition creates strange DNUs
Accepting the following class definitions in calypso open debuggers.

```smalltalk
Object << #MyClass
	slots: {};
	IMadeATypo: #x;
	sharedVariables: {};
	package: 'Toto'
```

```smalltalk
Object << #MyClass
	slots: {};
	sharedVariables: {};
	package: 'Toto'. 1
```

The following one does not create the class and browses something else!

```smalltalk
Object << #MyClass
	slots: {};
	sharedVariables: {};
	package: 'Toto'. Object
```

tesonep:
Fix #16355
- Improve the implementation to reduce duplicated code
- Handling issues and showing a dialog


MarcusDenker (approved):
Good ! We should in Pharo13 check if we can not use the classParser here instead of the compiler.

PR #16468, opened 2024-04-16
16436-Using-shortcut-CmdCtrl-W-on-a-window-with-multiple-tabs-closes-the-entire-window-instead-of-focused-tab

### LINKED ISSUE #16436: Using shortcut `Cmd/Ctrl-W` on a window with multiple tabs closes the entire window instead of focused tab
**Bug description**
Using shortcut `Cmd/Ctrl-W` on a window with multiple tabs closes the entire window instead of focused tab

**To Reproduce**
1. Open a senders window for a method and then fill it with tabs (look for implementors and senders of some method)
2. Press `Cmd/Ctrl-W`
3. Entire window is closed

**Expected behavior**
Only the focused tab should be closed.

tesonep:
Fix #16436
- Adding a closeByShortcut so windows can react different
- Renaming closeTopWindow to closeTopWindowByShortcut
- Adding an entry point so submorphs can collaborate with the window
- ClyGroupWindowMorph can handle the event and remove one of the panes

### DIFF
--- a/src/Calypso-Browser/ClyGroupWindowMorph.class.st
+++ b/src/Calypso-Browser/ClyGroupWindowMorph.class.st
@@ -41,6 +41,15 @@ ClyGroupWindowMorph >> dragTab: aSystemWindow event: anEvent in: aTabLabel [
 	anEvent hand grabMorph: aSystemWindow
 ]
 
+{ #category : 'closing' }
+ClyGroupWindowMorph >> handleCloseByShortcut [ 
+	
+	self activeWindow ifNotNil: [ :page | self removeWindow: page ].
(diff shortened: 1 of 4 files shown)


jecisc:
I have a problem with the generated image.

If I have a message browser with 3 tabs. I do CMD + w and it closes the active tab. Then I do it again. I was expecting the second tab to close because it should be the active one. Instead it did nothing.

guillep:
Hi @tesonep !

This was fixed here already before the release
https://github.com/pharo-project/pharo/pull/16493

I think we can close this one, if needed, please tell we will reopen it.

Issue #14916, opened 2023-10-07
Symbol selection: a proposal 


Ducasse:
After working a lot with symbol and classes.
Here is a proposal 

- simple click selects the text without the #
- shift + simple click select the complete symbol

Ducasse:
@jecisc we should introduce this one for the selection symbol

jecisc:
This is not so simple because Rubric is filtering the shift event to extend the selection so the event is not passed down to the click management

Ducasse:
what are the combinations that we have? 
shift command option
we could have option shift click or option cmd click for the # + symbol

hernanmd:
> After working a lot with symbol and classes. Here is a proposal
> 
>     * simple click selects the text without the #
> 
>     * shift + simple click select the complete symbol

I like the idea of Shift + Double click to select the complete symbol.

> This is not so simple because Rubric is filtering the shift event to extend the selection so the event is not passed down to the click management

Indeed is not so simple, and it is in the world loop (meaning the WorldMorph loop is constantly checking `closeTypeIn`).
One problem I've found is that there's no easy way to get the shift key state in `selectWord` (did you found one?).

On the other side it's strange that in `performCmdActionsWith: aKeyboardEvent shifted: aBoolean return: return` aBoolean is never used in the method. So one way is to set a new instance variable there to keep track of the shifted state.

Another thing which could be needed is to handle the command key:

```smalltalk
defaultCommandKeymapping
	| cmdMap |

	cmdMap := Dictionary new.
	{(KeyboardKey home -> #cursorHome:).
	(KeyboardKey end -> #cursorEnd:).
	(KeyboardKey backspace -> #backspace:).
	(KeyboardKey pageUp -> #cursorPageUp:).
	(KeyboardKey pageDown -> #cursorPageDown:).
	(KeyboardKey enter -> #crWithIndent:).
	(KeyboardKey escape -> #escape:).
	(KeyboardKey left -> #cursorLeft:).
	(KeyboardKey right -> #cursorRight:).
	(KeyboardKey up -> #cursorUp:).
	(KeyboardKey down -> #cursorDown:).
	(KeyboardKey keypadLeft -> #cursorLeft:).
	(KeyboardKey keypadRight -> #cursorRight:).
	(KeyboardKey keypadUp -> #cursorUp:).
	(KeyboardKey keypadDown -> #cursorDown:).
	(KeyboardKey keypadEnter -> #crWithIndent:).
[...]

PR #16469, opened 2024-04-16
Do not select the # in symbol selection

jecisc:
See discussion in: https://github.com/pharo-project/pharo/issues/14916


Ducasse:
tx :)

estebanlm:
this is not the same as #14916 ?

jecisc:
This is what we discussed to do before the P12 release in this issue.

MarcusDenker (approved):

estebanlm:
I know, so I can close also the other one?
EDIT: I see the real fix is for P13, better like that :)

PR #16474, opened 2024-04-16
tests for rules from the General-Rules package

AngelHely:
This pull request propose tests for rules from the General-Rules package with the tag migrated


src/General-Rules-Tests/ReAsClassRuleTest.class.st
```diff
@@ -0,0 +1,32 @@
...
+{ #category : 'converting' }
+ReAsClassRuleTest >> myMethodCallingAsClassIfPresents [
+	 ^ #ReAsClassRuleTest asClassIfPresent: [].
+]
+
+{ #category : 'tests' }
+ReAsClassRuleTest >> testRule [ 
+	|critiques|
+	critiques := self myCritiquesOnMethod: self myMethodCallingAsClassIfPresents.
+	self assert: critiques size equals: 1
```
jecisc: This one does not work as we have seen but we'll check together how we can fix that.

src/General-Rules-Tests/ReInconsistentMethodClassificationRuleTest.class.st
```diff
@@ -0,0 +1,25 @@
...
+	#name : 'ReInconsistentMethodClassificationRuleTest',
+	#superclass : 'ReAbstractRuleTestCase',
+	#category : 'General-Rules-Tests-Migrated',
+	#package : 'General-Rules-Tests',
+	#tag : 'Migrated'
+}
+
+{ #category : 'accessing' }
+ReInconsistentMethodClassificationRuleTest >> initialize [ 
+]
```
jecisc: We have a release test that ensure all #initialize methods are in the right protocol and this makes the test fail. 

Instead of having this method here, we could generate the method like this:

```st
self class compile: 'initialize super initialize' classified: 'accessing'.
[ "code of the test" ] ensure: [ (self class >> #initialize) removeFromSystem ]
```

#ensure: will make sure that even if the test crashes, the block deleting the method will be executed.

src/General-Rules-Tests/ReInconsistentMethodClassificationRuleTest.class.st
```diff
@@ -0,0 +1,25 @@
...
+
+{ #category : 'initalize' }
+ReInconsistentMethodClassificationRuleTest >> myCritiques [ 
+^ super myCritiques. 
+]
```
jecisc: Same comment as above

src/General-Rules-Tests/ReClassVariableCapitalizationRuleTest.class.st
```diff
@@ -0,0 +1,34 @@
...
+]
+
+{ #category : 'running' }
+ReClassVariableCapitalizationRuleTest >> tearDown [ 
+	classFactory cleanUp.
+	super tearDown
+]
+
+{ #category : 'tests' }
+ReClassVariableCapitalizationRuleTest >> testPoolVariableNameStartWithLowerCase [
```
jecisc, line 27: It would be nice to add a test in case there is no violation

src/General-Rules-Tests/ReEquivalentSuperclassMethodsRuleTest.class.st
```diff
@@ -0,0 +1,26 @@
...
+	| critiques |
+	critiques := OrderedCollection new.
+	self subjectUnderTest new
+		check: self class >> #sampleMethod
+		forCritiquesDo: [ :critique | critiques add: critique ].
+	^ critiques
+]
+
+{ #category : 'tests' }
+ReEquivalentSuperclassMethodsRuleTest >> testRule [
```
jecisc: It would be nice to add a test in case there is no violation

src/General-Rules-Tests/ReAsClassRuleTest.class.st
```diff
@@ -0,0 +1,32 @@
+Class {
+	#name : 'ReAsClassRuleTest',
+	#superclass : 'ReAbstractRuleTestCase',
+	#classInstVars : [
+		'critiques'
```
jecisc: This instance variable is unused and will make a release test checking that all instance variables are used fail.

jecisc (changes requested):

src/General-Rules-Tests/ReInconsistentMethodClassificationRuleTest.class.st
```diff
@@ -0,0 +1,40 @@
...
+ReInconsistentMethodClassificationRuleTest >> testRuleNotViolated [
+
+	| critiques |
+	self class
+		compile: 'initialize super initialize'
+		classified: 'initialization'.
+	critiques := self myCritiquesOnMethod: self class >> #initialize.
+	self assertEmpty: critiques.
+	[ "code of the test" ] ensure: [
+		(self class >> #initialize) removeFromSystem ]
```
jecisc:
```suggestion
	[ critiques := self myCritiquesOnMethod: self class >> #initialize.
	self assertEmpty: critiques ] ensure: [
		(self class >> #initialize) removeFromSystem ]
```

src/General-Rules-Tests/ReInconsistentMethodClassificationRuleTest.class.st
```diff
@@ -0,0 +1,40 @@
...
+	critiques := self myCritiquesOnMethod: self class >> #myCritiques.
+	self assert: critiques size equals: 1.
+	[ "code of the test" ] ensure: [
+		(self class >> #myCritiques) removeFromSystem ].
+	self class
+		compile: 'initialize super initialize'
+		classified: 'accessing'.
+	critiques := self myCritiquesOnMethod: self class >> #initialize.
+	self assert: critiques size equals: 1.
+	[ "code of the test" ] ensure: [
```
jecisc:
```suggestion
	| critiques |
	self class
		compile: 'myCritiques super myCritiques'
		classified: 'accessing'.
	[ critiques := self myCritiquesOnMethod: self class >> #myCritiques.
	self assert: critiques size equals: 1 ] ensure: [
		(self class >> #myCritiques) removeFromSystem ].
	self class
		compile: 'initialize super initialize'
		classified: 'accessing'.
	[ critiques := self myCritiquesOnMethod: self class >> #initialize.
	self assert: critiques size equals: 1 ] ensure: [
```

The #ensure: method makes sure the argument blocks gets executed after the block receiving the message, even if the block does not finish (eg. a non local return or an exception happens). So the receiving block must not be empty.

jecisc (approved):

PR #16477, opened 2024-04-17
Font-Tests: #testIsAbstractClass should be testIsAbstract like in oth…

### LINKED ISSUE #16476: Font-Tests: #testIsAbstractClass should be testIsAbstract like in other test classes
Only a minor thing but Font tests when testing for #isAbstract should not be called #testIsAbstractClass but #testIsAbstract to align with the other tests in the standard image and provide consistency:

image

astares:
…er test classes

Fix  #16476


MarcusDenker (approved):

jecisc:
I wonder if this is even useful to have as tests..

Ducasse:
To me not.

PR #16485, opened 2024-04-19
P11 - During compilation there are redundant allocation of collections

tesonep:
- Provide an implementation of #next:putAll:startingAt: for AbstractBinaryStream
- Improve the implementation of #duplicateTerminatorMarkOn: to avoid the creation of the stream, and the copies if not needed.


jecisc:
@tesonep Hey Pablo, the equivalent was merged in P12. Should we still backport to P11 or close this PR?

guillep:
I think that if there is no harm, we should integrate. The work is done, there is value in it. Worse case we can always revert it.

PR #16486, opened 2024-04-19
During compilation there are redundant allocation of collections

tesonep:
- Provide an implementation of #next:putAll:startingAt: for AbstractBinaryStream
- Improve the implementation of #duplicateTerminatorMarkOn: to avoid the creation of the stream, and the copies if not needed.


jecisc (approved):

jecisc:
This seems good. Should we merge it @tesonep ?

PR #16383, opened 2024-03-29
Fix serialization of classes with traits in hierarchy

PalumboN:
Try to fix https://github.com/pharo-project/pharo/issues/16345

Probably related to https://github.com/pharo-project/pharo/issues/16014 (TOCHECK)

image

PR #16487, opened 2024-04-19
Fix 16345: serialization of classes with traits in hierarchy

### LINKED ISSUE #16014: Pharo12: Image and sources size too large
**Bug description**
The Pharo12 image was ~51MB, Sources ~40MB (down from Pharo11 58MB image)

Now we suddenly have:

Image size: 62MB
Sources size: 54MB

Sources having grown by more than 10MB looks very very strange

### LINKED ISSUE #16345: Tonel export of class trait can be wrong
Since a few weeks/months in Pharo 12 it happens that the Tonel export of some code generates wrong class trait. It will be `#classTraits : '{} + TraitedClass'`. 

This prevents the code from loading afterward and is not easy to catch. 

This should be fixed before the release of P12.

estebanlm:
this tests should fix #16383

### DIFF
--- a/src/Shift-ClassBuilder-Tests/ShiftClassBuilderTest.class.st
+++ b/src/Shift-ClassBuilder-Tests/ShiftClassBuilderTest.class.st
@@ -54,6 +54,23 @@ ShiftClassBuilderTest >> testBuildClassTraitsAPI [
 ]
 
 { #category : 'tests - gathering' }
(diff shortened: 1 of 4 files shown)


jecisc:
Tests passed. Should we integrate to test it before the release?

estebanlm:
Yes, please On 19 Apr 2024, at 18:25, CyrilFerlicot ***@***.***> wrote:﻿
Tests passed. Should we integrate to test it before the release?

PR #16489, opened 2024-04-19
Reflectivity fixing bytecode to ast mapping

StevenCostiou:
Fixing a bug in syntax highlighting in the debugger because there is a desync between methods on the stack and methods modified with breakpoints in the system.

With @MarcusDenker 

## Step 1
[x] Moving the metalink count from the reflective method to the method node (as a property)
[ ] Ensure that every compiled method has its own reflective method

## Step 2
[ ] Move the bytecode-ast mapping to the method node
[ ] Move the dedicated bytecode-mappings of blocks to a dictionary in their parent method node bytecode-mappings


jecisc:
This PR has one failing test: osx-64 / Tests-osx-64 / testLinkCountTwoLinks – MacOSX64.Reflectivity.Tests.ReflectiveMethodTest

MarcusDenker:
The tests are still buiding using Pharo12 somehow

StevenCostiou:
I'm running the tests again, as I plan to work on it today.

StevenCostiou:
Test pass now. Can someone merge? I will do the other changes in another PR later.

PR #16502, opened 2024-04-21
Athens Demo not working

### LINKED ISSUE #16501: Athens Demo not working
In Pharo 12 check class comment from AthensDemoMoph. It suggests to run:

```Smalltalk
AthensDemoPresenter open
```

which fails with a DNU when you run it .

Screenshot from 2024-04-21 23 10 31

astares:
Fix #16501


jecisc (approved):
It might be nice to have smoke tests also

PR #16503, opened 2024-04-21
Clean System-Announcements-Tests

jecisc:
This change removes the System-Announcements-Tests package and move the test to other packages.

* Some tests are about protocols but we already cover the use cases in ProtocolAnnouncementTest so they are useless
* Some got moved to ReleaseTest because they check the state of the image so they belong to the ReleaseTest package
* Some got moved to AnnouncementTest because what they test is part of Announcement now and not SystemAnnouncement (this got moved during P12 development). Also this does not impact the system announcer anymore, so less impact on a global object.

I also tried to remove a skip on CI in a test because I have the impression we do not need it anymore


jecisc:
I also removed some ref to the singleton of SystemAnnouncer

Issue #16430, opened 2024-04-09
String>>asSymbol no protected against concurrent access


fedemennite:
**Bug description**
Sending String>>asSymbol may fail on inconsistency errors of the symbol table's access, if at least all the following is true:
- the symbols being looked up aren't yet in the symbol table
- such an operation is performed from multiple smalltalk processes

**To Reproduce**
No easy to reproduce.
Sending #asSymbol on sequentially random or enumerated strings from forked smalltalk processes with different priorities

**Expected behavior**
No error should be signaled 

**Version information:**
Pharo 10/11

**Expected development cost**
A simple implementation protecting String>>asSymbol with a mutually exclusive semaphore should be enough and quick to have.
A more refined implementation could try to lookup the symbol without semaphore protection and subsequently go through the semaphore only if the symbol doesn't exist yet.

pavel-krivanek:
Maybe Mutex would be safer to use here but this is such a low-lever speed critical operation that we probably need to accept Semaphore.

pavel-krivanek:
related to https://github.com/pharo-project/pharo/issues/6524

MarcusDenker:
PR was merged: https://github.com/pharo-project/pharo/pull/16565

PR #16506, opened 2024-04-22
Issue 16430 - String>> asSymbol no protected against concurrent access

carolahp:
Proposes solution to issue #16430 

### Summary 
To prevent data races when new symbols are interned into the NewSymbols table, a semaphore (stored as a class variable in the Symbol class) is used to protect the critical code in the method Symbol >> intern

### Dependencies
To fix failures during the bootstrap process due to changes in this PR, espell had to be modified. This PR must be merged only after changes in espell and in PharoBootstrap have been introduced. 
Once tests are passing, the corresponding PR will be created

### Tests
No new tests were added since reproducing dataraces is not straightforward


carolahp:
Replaced by PR [#16565](https://github.com/pharo-project/pharo/pull/16565)

PR #16507, opened 2024-04-22
Adding a Calypso class to display the Debug Points menu group in the method list context menu

### LINKED ISSUE #16488: Debugging context menu is lost with new debug point integration
Before the debug point integration, we had menu items in Calypso's about debugging and watchpoints

imagen

Now those menus are gone in build 1487

adri09070:
Fixes #16488 

The "Debug Points" command group was displayed everywhere in Calypso, except in the method list.

There was just a missing class to describe how it should display in the method list's context menu...

This design is so weird because it's no different than the rest of Calypso so I don't understand why this is necessary... It was a pain because it is not documented


guillep:
Thanks @adri09070 !

Issue #14302, opened 2023-07-16
Add rule for Point x: x y: y -> x@y


Ducasse:
Since x@y is a primitive it goes much faster than Point x: x y: y.
So a rule could help the user to know it.

jecisc:
Done

PR #16508, opened 2024-04-22
Add rule for Point x: x y: y -> x@y

AngelHely:
This pull request propose a new rule for the creation of instance Point  #14302


src/General-Rules/RePointRule.class.st
```diff
@@ -0,0 +1,46 @@
...
+	"This number should be unique and should change only when the rule completely change semantics"
+
+	^ 'PointRule'
+]
+
+{ #category : 'running' }
+RePointRule >> basicCheck: aNode [
+
+	aNode isMessage ifFalse: [ ^ false ].
+	^ #( #x:y: ) includes: aNode selector
```
astares: What if I have my own custom class and send an x:y: message to it - then it would lament as well (although it is not Point where the message is sent to) 

So I think the rule should check additionally for Point being the receiver of the message, no?

jecisc: I agree. And it would be nice to add a test on this exception

jecisc (approved):
Nice thank you.

I'll not integrate it yet since this is a new feature and we are in feature freeze and the CI is down

jecisc:

PR #16509, opened 2024-04-22
P12 external projects  fix repository versions

estebanlm:
subject :)

### DIFF
--- a/src/BaselineOfIDE/BaselineOfIDE.class.st
+++ b/src/BaselineOfIDE/BaselineOfIDE.class.st
@@ -179,7 +179,7 @@ BaselineOfIDE >> loadExtraTools [
 	"Hack: Document Browser is loaded in 2 steps because of the dependency of some packages to Iceberg.
 	The 2nd load is done through Metacello that adds new project registrations in Iceberg (user, not system).
 	We then remove these last registrations."
-	#('NewTools-DocumentBrowser' 'Iceberg') do: [ :repositoryName | | foundRepo |
+	#('NewTools-DocumentBrowser' 'iceberg') do: [ :repositoryName | | foundRepo |
 		foundRepo := IceRepository registry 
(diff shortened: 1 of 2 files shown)


Ducasse:
Esteban what is 2.6.0 for microdown. To me this is not correct. 
the version shipped with Pharo 12 is not the dev version tagged.

Ducasse:
I checked and it should be ok but I would not have name it 2.6.0 because I was with 2.5.2

estebanlm:
> I checked and it should be ok but I would not have name it 2.6.0 because I was with 2.5.2

yeah, I made a mistake. I now re-made the tag against Pharo12 branch (which was what was being integrated)

estebanlm:
> I checked and it should be ok but I would not have name it 2.6.0 because I was with 2.5.2

I always do a minor version when fixing for release :)

MarcusDenker:
The last run https://ci.inria.fr/pharo-ci-jenkins2/job/Test%20pending%20pull%20request%20and%20branch%20Pipeline/job/PR-16509/

shows a problem:

```
[31mMetacelloConflictingProjectError: Load Conflict between existing BaselineOfMicrodown [baseline] from github://pillar-markup/Microdown:v2.6.0/src and BaselineOfMicrodown [baseline] from github://pillar-markup/Microdown:Pharo12/src
[0mMetacelloAllowConflictingProjectUpgrade>>defaultAction
```

MarcusDenker:
I triggered a new check

Ducasse:
Now I'm confused because I do not know which version of microdown (because 2.6.0 does not mean anything to me) is shipped in Pharo. 
Because I do not know on which tagged version 2.6.0 is based. I do not know if it is a correct version and I have no idea how I can check it so what I will do is to not care but I will not move in case of problems.

Sorry but this is not external project user friendly.

PR #16511, opened 2024-04-23
add Open as external window option

pavel-krivanek:
The main purpose is to reveal more bugs :-)


astares (approved):

astares:
Nice one - thanks!

estebanlm:
I am scared about this in P12.
I guess better to do it in P13 and have a full iteration to fix?

jecisc:
I agree that it night be better in P13

pavel-krivanek:
Yes, sure, P12 is frozen. But there is not yet Pharo13 branch available ;)

jecisc:

Issue #5792, opened 2020-02-28
Add a validation rule stating that TestCase should not rely on initialize


Ducasse:
The testing framework nil out all the instance variables but it does not reinvoke the initialize method. It means that you can get a test breaking and when you rerun your test you get the environment of the test broken full of nil while you are thinking that the initialize should have been executed. 
We should define a rule to validate that users do not use initialize but setUp.

PR #16513, opened 2024-04-24
add rule to avoid that TestCase rely on Initialize

AngelHely:
This pull request proposes a rule to validate that users do not use initialize but setUp.
#5792

### DIFF
--- a/src/General-Rules-Tests/ReUseSetUpRuleTest.class.st
+++ b/src/General-Rules-Tests/ReUseSetUpRuleTest.class.st
@@ -21,8 +21,7 @@ ReUseSetUpRuleTest >> testRuleNotViolated [
 
 	| critiques |
 	self class compile: 'setUp'.
-	critiques := self myCritiquesOnClass: self class.
-	self assertEmpty: critiques.
-	[ "code of the test" ] ensure: [
+	[ critiques := self myCritiquesOnClass: self class.
+	self assertEmpty: critiques ] ensure: [
 		(self class >> #setUp) removeFromSystem ]
 ]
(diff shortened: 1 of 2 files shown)


jecisc:

src/General-Rules-Tests/ReUseSetUpRuleTest.class.st
```diff
@@ -0,0 +1,29 @@
...
+
+{ #category : 'tests' }
+ReUseSetUpRuleTest >> testRule [
+
+	| critiques |
+	self class compile: 'initialize' classified: 'initialization'.
+	critiques := self myCritiquesOnMethod: self class >> #initialize.
+	self assert: critiques size equals: 1.
+	[ "code of the test" ] ensure: [
+		(self class >> #initialize) removeFromSystem ]
```
jecisc:
```suggestion
	self class compile: 'initialize' classified: 'initialization'.
	[ critiques := self myCritiquesOnMethod: self class >> #initialize.
	self assert: critiques size equals: 1 ] ensure: [
		(self class >> #initialize) removeFromSystem ]
```

src/General-Rules-Tests/ReUseSetUpRuleTest.class.st
```diff
@@ -0,0 +1,29 @@
...
+
+{ #category : 'tests' }
+ReUseSetUpRuleTest >> testRuleNotViolated [
+
+	| critiques |
+	self class compile: 'setUp'.
+	critiques := self myCritiquesOnClass: self class.
+	self assertEmpty: critiques.
+	[ "code of the test" ] ensure: [
+		(self class >> #setUp) removeFromSystem ]
```
jecisc:
```suggestion
	[ critiques := self myCritiquesOnClass: self class.
	self assertEmpty: critiques ] ensure: [
		(self class >> #setUp) removeFromSystem ]
```

jecisc (changes requested):
The tests need to be improved. I commented in another PR the reason why (and how #ensure: works)

PR #16524, opened 2024-04-26
5 test for rules 

AngelHely:
This pull request propose 5 tests for rules of the package General-Rules

### DIFF
--- a/src/General-Rules-Tests/ReAssignmentInIfTrueRuleTest.class.st
+++ b/src/General-Rules-Tests/ReAssignmentInIfTrueRuleTest.class.st
@@ -14,17 +14,21 @@ ReAssignmentInIfTrueRuleTest >> testRule [
 		compile:
 		'testMethod |var| 1 == 1 ifTrue: [var := 1] ifFalse: [var := 2]'
 		classified: 'test-helper'.
+	[
 	critiques := self myCritiquesOnMethod: self class >> #testMethod.
-	self assert: critiques size equals: 1.
-	[ "code of the test" ] ensure: [ (self class >> #testMethod) removeFromSystem ]
+	self assert: critiques size equals: 1 ] ensure: [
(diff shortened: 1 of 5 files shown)


src/General-Rules-Tests/ReAssignmentInIfTrueRuleTest.class.st
```diff
@@ -0,0 +1,34 @@
+Class {
+	#name : 'ReAssignmentInIfTrueRuleTest',
+	#superclass : 'ReAbstractRuleTestCase',
+	#category : 'General-Rules-Tests-Migrated',
+	#package : 'General-Rules-Tests',
+	#tag : 'Migrated'
+}
+
+{ #category : 'tests' }
+ReAssignmentInIfTrueRuleTest >> testRule [
```
jecisc, line 10: Later I would also like a test with this code:

`testMethod | var | 1 == 1 ifTrue: [ 1 + 1. var := 1 ] ifFalse: [ var := 2 ]`

src/General-Rules-Tests/ReTemporaryNeitherReadNorWrittenRuleTest.class.st
```diff
@@ -0,0 +1,31 @@
...
+	#name : 'ReTemporaryNeitherReadNorWrittenRuleTest',
+	#superclass : 'ReAbstractRuleTestCase',
+	#category : 'General-Rules-Tests-Migrated',
+	#package : 'General-Rules-Tests',
+	#tag : 'Migrated'
+}
+
+{ #category : 'tests' }
+ReTemporaryNeitherReadNorWrittenRuleTest >> testRule [
+
```
jecisc, line 11: I wonder if the rule is triggered when we assign it and do not read it.

We should add a test checking that this is a violation in another PR:

`method |unused| unused := 1`

jecisc (approved):

PR #16525, opened 2024-04-26
Composite extract method refactoring

balsa-sarenac:
This is a bigger PR, introduces composite extract method refactoring and extract set up method. Ping me for any help with reviewing.


Ducasse:
I do not understand how I can see the tests.

MarcusDenker:
Lots of problems due to this PR, especially the fact that linux test runs just crash the VM is not good..

- [CI] [CI] Pharo13 tests on linux crash in Refactoring-Transformations-Tests #16567

This one I fixed already:
- [CI] Pharo13: code with halt was merged #16569 

There are other tests failing, I will open an issue for these, too

PR #16526, opened 2024-04-26
Refactoring of the RefactoryChange manager

balsa-sarenac:
This PR introduces:
- new API for the manager with deprecation of the old api
- fixing previous behavior (changes could get lost if you were doing a bigger refactoring)

Now all changes in undo stack are composite, so no more partial undos.
New API is:
- `performChanges:` with a list of changes to perform
- `performCompositeChange:` with a composite change to perform


jecisc:
There is one failing test but I have the impression it is unrelated

PR #16527, opened 2024-04-26
Refactor: preconditions return array of preconditions by default

balsa-sarenac:
We don't rely on a singe condition with and,or,not (we still use them when needed), but default return value of preconditions is array now.

Also, generateChanges method returns changes by default. A bunch of changes is caused by this, but they are trivial to review.


jecisc:
15 tests are failings with this PR.

Ducasse:
@jecisc where do you see the tests?

jecisc:
> @jecisc where do you see the tests?

I click for example on the #1 at the top and select tests results

jecisc:
image

Ducasse:
OK thanks for the info.

balsa-sarenac:
Failing tests seem unrelated

Ducasse:
nice!

PR #16532, opened 2024-04-27
add tests for testDoWhileFalse: and testDoWhileTrue:

### LINKED ISSUE #10430: Do we need doWhileFalse:
Do we need doWhileFalse: when we have whileFalse: 
and we have one user and the logic look the same.

```
BlockClosure >> doWhileFalse: conditionBlock
	"Evaluate the receiver once, then again as long the value of conditionBlock is false."
 
	| result |
	[result := self value.
	conditionBlock value] whileFalse.

	^ result
```

```
BlockClosure >> whileFalse: aBlock 
	"Ordinarily compiled in-line, and therefore not overridable.
	This is in case the message is sent to other than a literal block.
	Evaluate the argument, aBlock, as long as the value of the receiver is false."

	self value ifFalse: [ aBlock value. self whileFalse: aBlock ].
	^ nil
```

pavel-krivanek:
fixes #10430

### DIFF
--- a/src/Kernel-Tests/BlockClosureTest.class.st
+++ b/src/Kernel-Tests/BlockClosureTest.class.st
@@ -142,6 +142,44 @@ BlockClosureTest >> testCullCullCullCull [
 		equals: 4
 ]
 
+{ #category : 'tests - loops' }
+BlockClosureTest >> testDoWhileFalse [
+
+	| result number |
+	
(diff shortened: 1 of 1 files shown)


jecisc (approved):

jecisc:
Failing tests are unrelated

PR #16535, opened 2024-04-27
replace JPEGReadWriter by PluginBasedJPEGReadWriter and deprecate the…

### LINKED ISSUE #16534: JPEGReadWriter cannot read some files the PluginBasedJPEGReadWriter can
**Bug description**
Some files cannot be read using `JPEGReadWriter` while `PluginBasedJPEGReadWriter ` opens them.

For example the following one:
devices

```smalltalk
'devices.jpg' asFileReference binaryReadStreamDo: [ :stream |
(JPEGReadWriter on: stream) nextImage ].

'devices.jpg' asFileReference binaryReadStreamDo: [ :stream |
(PluginBasedJPEGReadWriter on: stream) nextImage ]
```

The first one fails with error `Error: marker D0 cannot be handled`.
Question is if it makes sense to keep them both in the image or rather rename `PluginBasedJPEGReadWriter` to `JPEGReadWriter` (which should be deleted or externalized) and deprecate the first name.

**Version information:**
 - Pharo Version: 11

pavel-krivanek:
… other name.

fixes #16534


Ducasse:
Can we integrate this?

jecisc:
I was wondering because I don't know this part of the system enough to know if the PR is all good. But I'm not sure we have a lot of people with that knowledge :_)

PR #16537, opened 2024-04-27
Remove empty Manifests from baseline packages, synchronize on the tag…

### LINKED ISSUE #8016: Small cleanup in BaselineOfEnlumineurFormatter and BaselineOfMicrodown
- small cleanup in these two baseline packages: tag the uncategorized baseline packages with tag "Base" 
  (as they have a manifest in that package which is tagged they are uncategorized)

pavel-krivanek:
… name #Baseline for Baseline methods where the Manifest is useful.

fixes #8016


jecisc:
But those manifest are there for the false positive on the not reference rule no?

pavel-krivanek:
yes, but in case of these removed ones, they are completely empty so there is no need to have them at all and there is no need for the baseline packages have more tags. Only in case the manifest is not empty, there will be two tags - Manifest and Baseline.

jecisc:
But I see some content with the ban of the rule: image

pavel-krivanek:
yes, but in this case, it is an obsolete record ;-)

jecisc (approved):

PR #16544, opened 2024-04-28
Adds a simple MoveTo taskbar menu item

### LINKED ISSUE #16543: The window tabs taskbar doesn't have a "Move To" option (only left/right)
**Describe the request**
While the taskbar tabs have move left/rigth options, its painful to move a new tab to a specific location (particularly the front of the tab list for something like a Transcript or Repository window

**Expected behavior**
Provide a menu option to MoveTo X

**Expected development cost**
Small - just needs a new menu item and a method to do this

**Screenshots**

**Version information:**
 - Version: [e.g. 1.6.1]
 - Pharo Version 13

**Additional context**
Have a fix for this that I will submit

macta:
Adds a simple MoveTo menu item for taskbar items (useful to move something like the Transcript or Repository window to the front of the Taskbar)

### DIFF
--- a/src/Morphic-Widgets-Windows/SystemWindow.class.st
+++ b/src/Morphic-Widgets-Windows/SystemWindow.class.st
@@ -1896,6 +1896,33 @@ SystemWindow >> taskbarMoveRight [
 	self moveTaskbarTaskToOffset: 1
 ]
 
+{ #category : 'move' }
+SystemWindow >> taskbarMoveTo [
+	"Move the current task button to the prompted location"
+	
+	| taskBarSize taskLocation desiredLocation answer |
+	
+	taskBarSize := self worldTaskbar orderedTasks size.
+	taskLocation := self worldTaskbar orderedTasks indexOf: self taskbarTask.
+
+	answer := UIManager default
+		          request: 'Enter new task location: 1..', taskBarSize printString
+		          initialAnswer: taskLocation asString
+		          title: 'Move Task'.
+		
+	answer ifNil: [ ^ self ].
+	
+	[
+	desiredLocation := answer asNumber.
+	((1 to: taskBarSize) includes: desiredLocation) ifFalse: [
+		self error: 'incorrect location' ] ] onErrorDo: [
+		UIManager default inform: 'Location must be a number between 1 and ' , taskBarSize printString.
+		^ self ].
+	
+	self moveTaskbarTaskToOffset: (taskLocation - desiredLocation) negated
+
+]
+
 { #category : 'open/close' }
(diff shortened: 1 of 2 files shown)


macta:
I meant to add this to Pharo12 but didn't get to it in time (and I know many people hate the taskbar and lots of windows - but I'm ok with them, and just like to have a few of them at the front and then I clean up the other windows at the end of the list)

Ducasse:
Hi tim

Thanks. I think that in the future we should a pre-release announce :).
BTW I do not really get what is your feature, even reading the code :).

S

macta:
Just to make it more obvious (as per @Ducasse feedback) - its adds a 3rd option to the move tab menu
image

And then lets you pick an absolute location:
image

Entering 1 will put the selected task item at the beginning on the list.

Ducasse:
Ah ok yes now I understand.

src/Morphic-Widgets-Windows/SystemWindow.class.st
```diff
@@ -1896,6 +1896,33 @@ SystemWindow >> taskbarMoveRight [
...
+{ #category : 'move' }
+SystemWindow >> taskbarMoveTo [
+	"Move the current task button to the prompted location"
+	
+	| taskBarSize taskLocation desiredLocation answer |
+	
+	taskBarSize := self worldTaskbar orderedTasks size.
+	taskLocation := self worldTaskbar orderedTasks indexOf: self taskbarTask.
+
+	answer := UIManager default
```
jecisc, line 1908: I know that Stephane wants us to reduce the references to UIManager. But Spec is loaded after SystemWindow so I am not so sure of how to do what you are doing without UIManager. Maybe @Ducasse or @estebanlm know?

macta: Just needs a simple way to ask for input - so presumably there is a recommended way to do this going forward and I can adjust and resubmit.

Worth adding, I've tried to use the taskbar primitives already there to keep the api as small as possible in case there are changes proposed in this area.

Ducasse: :( shitty architecture strikes back. Now SystemWindow is in morphic so I do not care. we should kill it alltogether.
Now this behavior will have to be redone on top of Toplo (and not the spec backend)

macta:
Would it help to have some kind of indirection to ask yes/no numeric questions? At least you could potentially slot something else in place when it's ready?

Ducasse:
I do not know but after we will see because we will have to redo some basic things in Toplo

PR #16547, opened 2024-04-29
proposition rule for nil branch

AngelHely:
This pull request proposes a potential rule that check not have unnecessary null branches. #16515


pavel-krivanek:
Why does it contain the addition of Toplo menu item?

AngelHely:
> Why does it contain the addition of Toplo menu item?

Sorry, I don't know why this file was here.

Ducasse:
Nice :)

PR #16549, opened 2024-04-29
[P12][P13] Fixing bug that raised `NodeNotFound` when an object-centric permalink was installed on a slot defined by a superclass of the object class

### LINKED ISSUE #16548: [P12][P13] MetaLinkInstaller raises `NodeNotFound` when installing an object-centric permalink for a variable slot defined in a superclass
**Bug description**
When installing an object-centric permalink on a variable slot, the method `registerAndInstallPermaLink:forTarget:` takes all nodes that access this slot and it installs the metalink on the equivalent nodes in the anonymous subclass of the object:

```Smalltalk
registerAndInstallPermaLink: aPermaLink forTarget: aSlotOrVar
	| nodes |
	(linksRegistry canReinstallPermaLink: aPermaLink)
		ifFalse: [ ^ self ].

	linksRegistry registerPermaLink: aPermaLink.

	nodes := (aSlotOrVar accessingNodesFor: aPermaLink persistenceType) asIdentitySet.
	aPermaLink targetObjectOrClass link: aPermaLink link toNodes: nodes
```

However, if the target variable slot has been defined by one of the superclasses of the object, it does not exclude the accessing nodes that are not in the object's class hierarchy (so: the nodes from methods that are in brother/cousin classes). As a result, it will try to find equivalent nodes in the anonymous subclass but these nodes do not exist so an exception `NodeNotFound` is raised.

**To Reproduce**

1. open a playground,
2. run `obj := SindarinDebuggerTest new`,
3. Inspect the object `obj`,
4. Click on the object-centric icon in the inspector toolbar (at the top, icon with colored circles),
5. Click on "Break on state access",
6. The exception is raised.

**Expected behavior**
No exception should be raised, the method `registerAndInstallPermaLink:forTarget:` should reject all nodes from methods that are not in the object's class hierarchy

**Version information:**
 - Pharo Version 12, but I wouldn't be surprised if the problem has been there since the beginning

**Expected development cost**
Easy fix, the analysis above was the hardest to do

adri09070:
Fixes #16548 

When installing an object-centric permalink on a variable slot, the method `registerAndInstallPermaLink:forTarget:` takes all nodes that access this slot and it installs the metalink on the equivalent nodes in the anonymous subclass of the object:

```Smalltalk
registerAndInstallPermaLink: aPermaLink forTarget: aSlotOrVar
	| nodes |
	(linksRegistry canReinstallPermaLink: aPermaLink)
		ifFalse: [ ^ self ].

	linksRegistry registerPermaLink: aPermaLink.

	nodes := (aSlotOrVar accessingNodesFor: aPermaLink persistenceType) asIdentitySet.
	aPermaLink targetObjectOrClass link: aPermaLink link toNodes: nodes
```

However, if the target variable slot has been defined by one of the superclasses of the object, it does not exclude the accessing nodes that are not in the object's class hierarchy (so: the nodes from methods that are in brother/cousin classes). As a result, it will try to find equivalent nodes in the anonymous subclass but these nodes do not exist so an exception `NodeNotFound` is raised.

This breaks object-centric debug points, for example

I fixed that method to only keep the accessing nodes that are in the class hierarchy of the target object.

### DIFF
--- a/src/DebugPoints/DebugPoint.class.st
+++ b/src/DebugPoints/DebugPoint.class.st
@@ -204,19 +204,6 @@ DebugPoint >> install [
 	self target install: self link
 ]
 
-{ #category : 'API' }
-DebugPoint >> instanceVariable: aSlot accessStrategy: aSymbol [
-
-	target := DebugPointInstanceVariableTarget new
-		          instanceVariable: aSlot;
-		          accessStrategy: aSymbol;
-		          yourself.
-
-	name := 'var_{1}_{2}' format: {
-			        aSlot name.
-			        aSymbol }
-]
-
 { #category : 'accessing' }
 DebugPoint >> link [
 	^metaLink
@@ -372,6 +359,20 @@ DebugPoint >> target [
 DebugPoint >> targetClass [
 
 	^ self target targetClass
+]
+
+{ #category : 'API' }
+DebugPoint >> targetClass: aClass instanceVariable: aSlot accessStrategy: aSymbol [
+
+	target := DebugPointInstanceVariableTarget new
+					 targetClass: aClass;
+		          instanceVariable: aSlot;
+		          accessStrategy: aSymbol;
+		          yourself.
+
+	name := 'var_{1}_{2}' format: {
+			        aSlot name.
+			        aSymbol }
 ]
 
 { #category : 'scope' }
--- a/src/DebugPoints/DebugPointInstanceVariableTarget.class.st
+++ b/src/DebugPoints/DebugPointInstanceVariableTarget.class.st
@@ -7,7 +7,8 @@ Class {
 	#name : 'DebugPointInstanceVariableTarget',
 	#superclass : 'DebugPointClassTarget',
 	#instVars : [
-		'accessStrategy'
+		'accessStrategy',
+		'targetClass'
 	],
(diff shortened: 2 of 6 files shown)


adri09070:
Some `VariableBreakpoint` tests fail because the fix break the semantics of non-object-centric variable breakpoints:

My fix is necessary if the permalink target is an objec, however the same method is called to install non-object-centric permalinks.
However, when we set a non-oc variable breakpoint on a variable in a class, it considers that the breakpoint should hit if an access is performed in any of the subclasses of the class that defined the target slot.

While the fix, as it is shared with oc variable breakpoints, considers that the variable breakpoint should hit only if the access is in the class hierarchy of the target class (i.e: the class of the object) (which is to me not a bad  semantics, else why let the developer choose a target class? Moreover, my semantics allows to have better control on when the breakpoint should hit)

This should be discussed. Does the semantics need to be changed? Should I change the fix to stick to the old semantics for non-object-centric variable breakpoints? Anyway, variable breakpoints should be completely replaced by variable debug points .

 @StevenCostiou

jecisc:
@StevenCostiou What should we do with this PR?

StevenCostiou:
Please keep it open a bit until I can do a deeper pass

PR #16551, opened 2024-04-29
ProcessorScheduler>>#activeProcess added no instrumentation pragma

jordanmontt:
Added `noInstrumentation` pragma to ProcessorScheduler>>#activeProcess. This is important since this method cannot be instrumented because it's used inside the instrumentation to check for meta-recursions


pavel-krivanek (approved):

Ducasse:
We could port it back to P12.

jordanmontt:
Backporting PR: https://github.com/pharo-project/pharo/pull/16557

PR #16555, opened 2024-04-30
Make it easier to diagnose mock failures by printing message sends properly and storing the send history in the mock

macta:
Probably many more fixes to add but this helps a little bit


macta:
This fix makes the debugger look like this (which is slightly easier to read and diagnose). I think there are more pragmatic and simple fixes we can make to help make mocks helpful without overdoing things.

image

Possibly I should remove the "Expected Message" prefix, so that the messages and failed collections better line up for an easier comparision.
Also - would be tempted to rename messages and failed variables to expected and actual to make it a bit more obvious.

macta:
After previous comment, I tried simplifiying the printOn: and it works better - so have adjusted it:
image

MarcusDenker (approved):

PR #16560, opened 2024-05-01
ExteranlAddress>>fromString: converts string to UTF-8 first

### LINKED ISSUE #8945: Check ExternalAddress >> #fromString: and ByteArray >> #writeString:
These methods are not handling the encoding correctly.
They are unsafe if they are used with wide strings.

pavel-krivanek:
fixes #8945


jecisc (approved):

PR #16562, opened 2024-05-02
Fix a small bug with FFI calls when passing 'NULL'  as argument to the C function

doste:
Now it can infer its type correctly as 'void*'

### DIFF
--- a/src/UnifiedFFI/FFIStrictResolutionMode.class.st
+++ b/src/UnifiedFFI/FFIStrictResolutionMode.class.st
@@ -26,10 +26,8 @@ FFIStrictResolutionMode >> resolveUndeclaredTypeForArgument: aFFIValueArgument w
 	 	 There is no ambiguity.
 	"
 	
-	(aFFIValueArgument value isNil
-		or: [aFFIValueArgument value = 'nil'
-			or: [ aFFIValueArgument value = 'NULL' ]])
-				ifTrue: [ ^ (aResolver resolveType: #'void *') ].
+	(#(nil 'nil' 'NULL') includes: aFFIValueArgument value)
+		 ifTrue: [ ^ (aResolver resolveType: #'void *') ].
 				
 	aFFIValueArgument value == #self ifTrue: [
 		| externalType |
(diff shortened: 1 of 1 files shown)


src/UnifiedFFI/FFIStrictResolutionMode.class.st
```diff
@@ -18,10 +18,19 @@ FFIStrictResolutionMode >> isStrict [
...
+	"Make strict all declarations except those using self and NULL.
 	This is safe because:
 	 - We can infer self from the current class
-	 - the only implementor of prepareAsSelfFromCalloutDeclaration are those classes representing pointers"
+	 - the only implementor of prepareAsSelfFromCalloutDeclaration are those classes representing pointers
+	 - NULL can only be of one specific type: void*
+	 	 There is no ambiguity.
+	"
+	
+	(aFFIValueArgument value isNil
```
astares: This give a lint  issue and to make it easier it could be rewritten as

```Smalltalk
	(#(nil 'nil' 'NULL') includes: aFFIValueArgument value)
		 ifTrue: [ ^ (aResolver resolveType: #'void *') ].
	 ...
```
avoiding the critique to appear

astares:
Also this PR should be against Pharo13 branch (with a possible backport to P12)

tesonep:
I changed the base of the PR, thanks for the catch @astares

src/UnifiedFFI/FFIStrictResolutionMode.class.st
```diff
@@ -18,10 +18,19 @@ FFIStrictResolutionMode >> isStrict [
...
+		or: [aFFIValueArgument value = 'nil'
+			or: [ aFFIValueArgument value = 'NULL' ]])
+				ifTrue: [ ^ (aResolver resolveType: #'void *') ].
```
guillep:
```suggestion
	(#(nil 'nil' 'NULL') includes: aFFIValueArgument value)
		 ifTrue: [ ^ (aResolver resolveType: #'void *') ].
```

src/UnifiedFFI/FFIStrictResolutionMode.class.st
guillep: Thanks @astares, I applied your suggestion using the github review tool.

Let's wait for the CI

guillep:
Thanks @doste for the PR and @astares for the review.
We are going to discuss if this is worth backporting to P12

